PR #1043, opened 2018-05-23
Testing/log

jim-parry:
Flesh out Log module testing, with near 100% code coverage.
Fixes bug in ChromeLoggerHandler re error level conversion.


log-2018-05-22.php
```diff
@@ -0,0 +1,7 @@
+<?php defined('BASEPATH') || exit('No direct script access allowed'); ?>
```
lonnieezell: Don't think you meant to include this one...

jim-parry: Nope - that should have been git ignored. Will fix

jim-parry:
@lonnieezell Merged your testing changes here, & sorted out the namespaces.
Travis-ci runs, though no code coverage, but this still croaks trying to run locally - the same as the comment on your merged PR.
I am not comfortable merging this until I resolve phpunit locally.

lonnieezell:
@jim-parry Is this good to go now?

jim-parry:
It is indeed good to go. I didn't want to merge it directly, in case you had any concerns.

PR #1048, opened 2018-05-25
Testing/validation

jim-parry:
Validation tests refactored & fleshed out.
Code coverage complete except for Rules::is_unique(), which relies on database.


jim-parry:
@lonnieezell This should be good to go if it passes travis-ci & you have no concerns

system/Validation/FileRules.php
```diff
@@ -174,7 +176,9 @@ public function mime_in(string $blank = null, string $params, array $data): bool
 			return false;
 		}
 
-		return in_array($file->getType(), $params);
+//		echo var_dump($file);
```
lonnieezell: Please remove debug code. :)

system/Validation/FormatRules.php
```diff
@@ -334,11 +334,14 @@ public function valid_url(string $str = null): bool
 		}
 		elseif (preg_match('/^(?:([^:]*)\:)?\/\/(.+)$/', $str, $matches))
 		{
-			if (empty($matches[2]))
-			{
-				return false;
-			}
-			elseif ( ! in_array($matches[1], ['http', 'https'], true))
+// removed the test for an empty second match segment. 
+// it should never be found, because of the "(.+)" pattern, which is one or more
```
lonnieezell: Might as well remove the code if it's not needed. you can leave the comment so we don't try to put it back in down the road, though.

jim-parry: Done, but I removed the comment since it looked out of place without the commented code.

lonnieezell:
Looks good. Thanks!

PR #1055, opened 2018-05-30
Refactor/exceptions

jim-parry:
Change **new self** to **new static**, giving the intended result of only having to look for "module" exceptions rather than needing multiple catch clauses.


jim-parry:
oops - I see that I based my branch off the wrong one. Will fix

PR #1058, opened 2018-05-30
Remove encryption classes from autoload classmap

jim-parry:
These mappings were accidentally left in when working on the Encryption class.
They do not belong now, and appropriate replacements will come when encryption is re-examined.


lonnieezell:
At some point I want to do a comparison with most of the items in that classmap removed. I have a feeling it’s not giving us the speed increase we thought it might but haven’t tested that just yet.

jim-parry:
Agreed. There are a few things we need to visit/revisit soon :)

PR #1066, opened 2018-06-06
Use the proper test stream filter

jim-parry:


jim-parry:
@lonnieezell I was surprised that the previous PR passed travis-ci (because I was using the wrong stream filter), and discovered some weirdness in the builds when I investigated. Since a few PRs ago (#1060), travis is reporting an error in the phpunit.xml.dist config, and there is a "Hello World" echoed at the end of the tests. It looks like my last PR didn't pass the unit tests, though it was reported as successful. Hmmm.
Travis claims to be happy with this PR, but the travis details show a ton of errors.

lonnieezell:
Hmm... I've seen that issue with it echoing out Hello World before, but I don't recall the fix. It was a pain to track down, too. I can try to look at things tonight.

PR #1070, opened 2018-06-13
Add \Config\App shared instance, Service:app()

puschie286:
implement shared instance for \Config\App
replace all creation with shared instance usage

-> please check if somewhere its necessary to have a clean/new instance

PS: i know the doc is not updated

### DIFF
--- a/application/Filters/DebugToolbar.php
+++ b/application/Filters/DebugToolbar.php
@@ -37,7 +37,7 @@ class DebugToolbar implements FilterInterface
 		{
 			global $app;
 
-			$toolbar = Services::toolbar(new App());
+			$toolbar = Services::toolbar();
 			$stats   = $app->getPerformanceStats();
 			$data    = $toolbar->run(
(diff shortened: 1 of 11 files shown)


puschie286:
some test results ( 1000 runs ) :
first call takes about 0.2ms for initialize the config and 0.01ms for each access
-> so maybe we should use a similar approach for other configs

lonnieezell:
Interesting approach, and I like a lot of things about it. However, config files are not services, so I don't like it in that file.

It also plays into something bigger I've been pondering for a while now: a `config()` helper function. I wasn't sure how it would work really well until I was reading through your PR, though. 

If you're up for it, I've created [a new Issue](https://github.com/bcit-ci/CodeIgniter4/issues/1071) with the details of how I'd like to see this work. 

Interested?

puschie286:
sure :)
yea it also felt wrong to put it in service, but was the simplest way^^
-> move conversation to #1071

Issue #1071, opened 2018-06-13
New Config helper


lonnieezell:
Inspired by https://github.com/bcit-ci/CodeIgniter4/pull/1070:

- new `CodeIgniter\Config\Config` class with one public static method: `get()`. This takes a class name and returns an instance of the class. 
- By default it returns a shared instance, but a new instance can be returned by setting the second parameter to `false`, just like the Services class.
- If the config name is fully qualified (`Config\App`) it will simply try to instantiate/cache and return.
- If the config name does not include the namespace (`App`), it will attempt to discover the class similar to way [the Services class does](https://github.com/bcit-ci/CodeIgniter4/blob/develop/system/Config/BaseService.php#L179), though don't discover and preload all of them, just as requested.
- new helper method, `config()` that is just a slightly nicer way of calling `CodeIgniter\Config\Config::get()`.

puschie286:
first implementation done
should cover all requested aspects
-> documentation missing

i just copy the getClassname function from BaseService, maybe we should move this function to the filelocator ?

another thing is, we should store instances only by class name to avoid storing two instances for name and namspace + name

lonnieezell:
Sorry for the delay - was on vacation. 

> i just copy the getClassname function from BaseService, maybe we should move this function to the filelocator ?

Good idea.

> another thing is, we should store instances only by class name to avoid storing two instances for name and namespace + name 

sounds good

puschie286:
np, me too^^
changes pushed -> next point would be to replace config creation with sharedInstance requests
PS : "The method getClassname() has a Cyclomatic Complexity of 19. The configured cyclomatic complexity threshold is 10" is this settings correct ?

lonnieezell:
That service hasn't been tweaked to our specific requirements yet. It's just something we're trying out. We don't request changes or block diffs based on that report, so nothing to worry about. 

I'll take a look at the PR later today. Thanks.

PR #1073, opened 2018-06-14
Config helper

puschie286:
config helper implementation for #1071 

### DIFF
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -152,6 +152,54 @@ class FileLocator
 	//--------------------------------------------------------------------
 
 	/**
+	 * Examines a file and returns the fully qualified domain name.
+	 *
+	 * @param string $file
+	 *
+	 * @return string
+	 */
+	public function getClassname(string $file) : string
+	{
+		$php    = file_get_contents($file);
+		$tokens = token_get_all($php);
+		$count  = count($tokens);
+		$dlm    = false;
+		$namespace = '';
+		$class_name = '';
+
+		for ($i = 2; $i < $count; $i++)
+		{
+			if ((isset($tokens[$i-2][1]) && ($tokens[$i-2][1] == "phpnamespace" || $tokens[$i-2][1] == "namespace")) || ($dlm && $tokens[$i-1][0] == T_NS_SEPARATOR && $tokens[$i][0] == T_STRING))
+			{
+				if (! $dlm)
+				{
+					$namespace = 0;
+				}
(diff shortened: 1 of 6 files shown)


lonnieezell:
Looks good. Please add docs and some tests for the class please, and then we are good to go.

jim-parry:
Something is broken in the travis-ci unit tests.
This PR actually did not pass the tests ... I get the following when I run it locally:

> 1) CodeIgniter\Config\ConfigTest::testCreateSingleInstance
> Failed asserting that null is an instance of class "Config\Email".
> 
> /pub7/htdocs/CodeIgniter4/tests/system/Config/ConfigTest.php:14
> 
> 2) CodeIgniter\Config\ConfigTest::testCreateSharedInstance
> Failed asserting that false is true.
> 
> /pub7/htdocs/CodeIgniter4/tests/system/Config/ConfigTest.php:31

Looking at the travis-ci results, it shows an error and four failures, but then echos "Hello World" at the end of the row after 1403/2089 and then exits with "success" !?

jim-parry:
ps - the error and two failures outside of ConfigTest show in other PRs, and were not caused by this PR.

puschie286:
Hmm, the tests passed locally - it’s pretty weird, not sure what’s the different between our installation that can cause this.

jim-parry:
Are you perchance running on WIndows locally?

puschie286:
Yes
  

  
  

  
  
>

jim-parry:
Oops. You will find Config::get('email') because WIndows is case-insensitive. Linux won't find it, hence the test fails.
Lines 10 & 28 in your test will not return a class on Linux.

puschie286:
Oh, so i can’t handle lower case class names at all.
I will push the needed changes tomorrow.

jim-parry:
The alternative is to UCfirst the class name that results, which should work on all platforms.

puschie286:
hmm, the internet says no -> php is **not** case sensitive for classes [SO](https://stackoverflow.com/questions/5260168/capital-letters-in-class-name-php)

i guess the filenames are the problem, because we are loading them directly.

jim-parry:
It is not a classname case sensitivity, it is a filename case sensitivity problem.
If you have a file AbCdEf.php, containing a PurpleDragon class definition, then you can load it as "abcdef.php" on Windows, but you need to use the correct filename case on Linux, or the source file will not be found.
Once loaded, you can then instantiate purpledragon or PURPLEdragon or whatever :-/
The example here is purposeful, to highlight the different issues.

PR #1080, opened 2018-06-22
Aerospike Cache Driver

byazrail:
I've been using it for about 4 months. Aerospike Cache Library.
I also wanted to add it here.
I could also add the Aerospike Session Library if I wanted to.

### DIFF
--- a/system/Cache/Handlers/AerospikeHandler.php
+++ b/system/Cache/Handlers/AerospikeHandler.php
@@ -259,7 +259,7 @@ class AerospikeHandler implements CacheInterface
 			return true;
 		}
 
-		return true;
+		return false;
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jim-parry:
Something like this definitely needs to be GPG-signed for consideration.

The travis-ci config needs to have aerospike added, so that your provided unit testing is run. I don't see it in the list of supported services on travis-ci.org.

byazrail:
Travis service on see : https://github.com/aerospike/aerospike-client-php/blob/master/.travis.yml
After reviewing I will add the necessary.

byazrail:
Travis service on not supported.
Close Requests.

jim-parry:
Bummer. You could always make it into an addin and offer it to the community over the forum (for now).

byazrail:
Oke Thank, you.
Do you have the class library you want to insert?
Oracle FTP etc. I use the libraries because.

jim-parry:
I don't understand your question. Oracle/FTP don't seem to have a connection to aerospike.

jim-parry:
if you are thinking of namespace, then Aerospike would make sense, for an "Aerospike" module, with config & drivers inside it.

PR #1088, opened 2018-06-29
ConfigHelper : case-sensitive

puschie286:

### DIFF
--- a/system/Config/Config.php
+++ b/system/Config/Config.php
@@ -70,7 +70,7 @@ class Config
 			$class = substr($name, $pos + 1);
 		}
 
-		$class = strtolower($class);
+		//$class = strtolower($class);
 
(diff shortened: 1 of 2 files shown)


jim-parry:
This fixed the case sensitivity issue. It looks like there are other issues with the travis-ci build, but not caused by you :)

system/Config/Config.php
```diff
@@ -70,7 +70,7 @@ public static function get(string $name, bool $getShared = true)
 			$class = substr($name, $pos + 1);
 		}
 
-		$class = strtolower($class);
+		//$class = strtolower($class);
```
lonnieezell: Please remove the line instead of commenting it out. If we need something back, we can always look back through the history. Thanks.

PR #1090, opened 2018-06-29
name contains uppercase letters. get file object error.

bangbangda:
With the simplest usage, a single file might be submitted like:
<input type="file" name="userFile" />

Controller code :
$file = $this->request->getFile('userFile');

so. $file is null.

FileCollection file Line 182,
$this->files[$name] = $this->createFileObject($file);

$name not be converted to lowercase.


bangbangda:
With the simplest usage, a single file might be submitted like:
<input type="file" name="userFile" />

Controller code :
$file = $this->request->getFile('userFile');

so. $file is null.

bangbangda:
With the simplest usage, a single file might be submitted like:
`<input type="file" name="userFile" />`

Controller code :
`$file = $this->request->getFile('userFile');`

so. $file is null.

lonnieezell:
Good catch. I had thought that it was standardized to lowercase elsewhere, but doesn't look like it is. However, looks like there's some additional errors in Travis that aren't showing up. Most of those I saw on another request so they aren't an issue with this, but a couple are new. I'll have to pull it down and check it manually later. And hopefully get those Travis errors figured out also. :(

bangbangda:
@lonnieezell 
Hi. the PR not merging?
My blog use the CI4 framework, and this problem has caused me trouble.

lonnieezell:
@bangbangda Still trying to get the Travis errors sorted out. I believe I got all of them figured out except for one. And it's being a pain to track down what's causing it...

bangbangda:
ok Good work!
I am also ready to Travis.
work hard together.
@lonnieezell

jim-parry:
Re travis-ci, where is the "Hello World" coming from? It looks like something is echoing it and then dieing, resulting in a false positive passing of travis.

lonnieezell:
> Re travis-ci, where is the "Hello World" coming from? It looks like something is echoing it and then dieing, resulting in a false positive passing of travis.

@jim-parry That's what I'm trying to pin down. :) It's a pain because it doesn't seem to happen when I test any of the groups in isolation, but have to have run a number of groups of tests prior to it for it to show up. Spent 1.5 hours on it last night, and haven't quite found it. Know that I had encountered this once before and thought we had it taken care of, but no suck luck I suppose. When I use the stock phpunit.dist.xml it shows up for me, but the slightly older version I had copied for local changes didn't let it show up. Couldn't find much difference between the two, though.

jim-parry:
I've spent some time trying to track it down too, without success :(

PR #1091, opened 2018-07-02
change substr with strpos/stripos when possible

samsonasik:


jim-parry:
Why?

samsonasik:
it is faster, based on phpmanual (noted at strstr doc  http://php.net/manual/en/function.strstr.php )

```
If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead.
```

jim-parry:
Hmmm - the reference you link to says that strpos is faster than strstr, but you want to replace substr?
I can see that strpos is faster than strstr, but faster than strpos?? Those functions serve two different purposes.

jim-parry:
Hmm - googling results in lots of anecdotal suggestions that strpos is faster, and one benchmark ...

> "strpos()" was the faster by 0.0041 seconds, (0.445% faster)

Not sure if that would much of a difference. Code clarity seems comparable.
Bottom line: I don't object. Lonnie ?

lonnieezell:
I'm not opposed to the changes, but I'm not a fan of the `0 === x` syntax. I thought we had something against that in our code styles but i'm not seeing it. 

Either way, please change the order of comparisons from `0 === strpos($value,'a:')` to `strpos($value,'a:') === 0`.

Then let's wait to approve until we get whatever is going on with Travis taken care of so we can re-run the tests. I'm going to try to look into that this evening.

samsonasik:
@lonnieezell updated

PR #1092, opened 2018-07-02
count optimization : using truthy check instead when possible

samsonasik:


lonnieezell:
I'll ask what @jim-parry asked last time: Why? 

You can probably save us time if you provide a description with your PR's....

samsonasik:
when the variable is already an array, and we want to check the data is more than 0/data exists, we can just use the truthy check, it will be faster, using `count()` will slower.

PR #1093, opened 2018-07-04
Testing2/helpers

jim-parry:
Enhanced unit testing for the second half of the helpers, with source & docs adjustments as needed.
this doesn't result in 100% code coverage, but significantly improves it.


jim-parry:
codacy flagged issues in the testing code, namely accessing $_POST and $_SESSION. These are on purpose, for testing purposes. How do well codacy they are ok?

jim-parry:
codacy flagged the object() method in html_helper as having an unused parameter, $type.
That is true, and I don't know why it is there (it was there previously).
The user guide makes use of it in its example, though the passed "type" is not part of the expected output.

jim-parry:
is "codacy" an integration for CodeIgniter4-standard?
I don't see meniton of it in composer.json or in a phpcs.xml, nor any config file for the standards it is testing against :(
I could be missing something :-/

lonnieezell:
@jim-parry We could tweak it to be like that, but haven't looked into tweability just yet. Last time we talked we weren't sure if we'd keep it around or not. :)

lonnieezell:
As far as the `object()` method goes it looks to be an issue and type should be included. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object) both a `data` element and a `type` element must be present.

jim-parry:
The tests fail at FeatureTestCaseTest:testCallGet.
If I swap the order of tests, moving that in front of TestCallSimpleGet, then the tests fail siently at testCallSimpleGet.
I do not suspect the tests, but rather a memory limitation. Playing further.

jim-parry:
Alright - I can reproduce the problem on my local system, when the database tests are included. WIll try to solve it here, then push the needed changes.

lonnieezell:
@jim-parry I kinda fixed that issue in https://github.com/bcit-ci/CodeIgniter4/commit/274fcb533e503cf526b485e48da72888196d92d8 but it seemed to break other stuff, or at least make them visible now. Sigh. I've got to crash, though, so I'll look more into it tomorrow night, unless you happen to spot something before then.

jim-parry:
Alright - the build now works locally for me, without errors, and I see that the one travis build (7.1 & mysql) works, but that the others have issues (some a few & some lots). I will see if I can chase down some of these today, but no promises (world cup is on).

On the bright side, the code coverage continues to increase :)

PR #1094, opened 2018-07-06
Fix travis breaking from backwards incompatiblity

jim-parry:
Fixes to address travis breaking.
Some of these are [backwards incompatible PHP7.2 changes](http://php.net/manual/en/migration72.incompatible.php) to work around

### DIFF
--- a/system/Cache/Handlers/MemcachedHandler.php
+++ b/system/Cache/Handlers/MemcachedHandler.php
@@ -287,7 +287,8 @@ class MemcachedHandler implements CacheInterface
 
 		$stored = $this->memcached->get($key);
 
-		if (count($stored) !== 3)
+        // if not an array, don't try to count for PHP7.2
+		if (! is_array($stored) || count($stored) !== 3)
 		{
 			return FALSE;
 		}
--- a/system/Helpers/inflector_helper.php
+++ b/system/Helpers/inflector_helper.php
@@ -60,7 +60,7 @@ if ( ! function_exists('singular'))
 	{
 		$result = strval($string);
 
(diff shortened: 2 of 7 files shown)


jim-parry:
Chipping away at these. WIll continue after the world cup game :-/

jim-parry:
It looks like I have nailed the problems affecting PHP7.1, and the "only" remaining PHP7.2 one has to do with ini_set(). I have found a workaround at https://github.com/yiisoft/yii/commit/0be5fbb5b359065beb2770feb262a4a8f689a309
If that is an acceptable approach, let me know & I shall fix ours similarly.

lonnieezell:
@jim-parry Kind of a pain to have to do that, but makes sense why they're doing it. I guess it makes sense why you shouldn't change session ini settings in the middle of a session, too. :) I guess I don't have an issue with that. 

Bigger question, though, where are we changing session settings after it's started? If it's during session tests, are we sure we're always stopping the session after each test? Or is it something else that's causing it?

jim-parry:
All the remaining errors arise in SessionTest, and it looks like they all stem from ultimately invoking Session:configure() aftera session has started. It could be that just that one method will need beefing up. I can play with that this evening.
It looks like this is one of the backwards incompatible changes for PHP7.2, and we can head it off at the pass if we are careful.

lonnieezell:
This is looking great so far. Nice job! I'm going to go ahead and merge to make it easier to see what errors are actually happening in some of these other PRs people have been waiting on.

jim-parry:
Sounds good. I hope to have the session problem fixed this evening :)

PR #1095, opened 2018-07-08
toolbar cannot be initialized correctly.

bangbangda:
setcookie() is arrays will cause ‘Array to string conversion’ to be wrong.
toolbar cannot be initialized correctly.

Example, Home Controller add code:
`setcookie("cookie[three]", "cookiethree");
setcookie("cookie[two]", "cookietwo");
setcookie("cookie[one]", "cookieone");`

### DIFF
--- a/system/Debug/Toolbar/Views/toolbar.tpl.php
+++ b/system/Debug/Toolbar/Views/toolbar.tpl.php
@@ -237,7 +237,7 @@
 				<?php foreach ($cookies as $name => $value) : ?>
 					<tr>
 						<td><?= $name ?></td>
-						<td><?= $value ?></td>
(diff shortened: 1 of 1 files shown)


lonnieezell:
Well - cookies can’t atore arrays so I don’t think this is ever the case is it? A cookie can only store strings.

bangbangda:
yes, arrays do not appear under standard conditions.
but. setcookie can be an array.
http://php.net/manual/en/function.setcookie.php

CI4 does not need to be processed?

lonnieezell:
Ah. You're absolutely right. I had forgotten that PHP converted that into an actual array when you grabbed it back. Good catch.

bangbangda:
Very efficient.👍

bangbangda:
@lonnieezell Very efficient.👍

PR #1096, opened 2018-07-08
Fix session testing for PHP 7.2+

jim-parry:
Addresses travis-ci breaking with PHP7.2

### DIFF
--- a/system/Session/Session.php
+++ b/system/Session/Session.php
@@ -1,4 +1,6 @@
-<?php namespace CodeIgniter\Session;
+<?php
+
+namespace CodeIgniter\Session;
 
 /**
  * CodeIgniter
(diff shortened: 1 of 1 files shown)


lonnieezell:
@jim-parry I assume you're still working on this? Tests still showing as broken in 7.2.

jim-parry:
I am still working on it. Trying different approaches :-/

jim-parry:
ok - I have been at this for several hours, and am makig a new dent in my wall where I bang my head :-/
The same problem has been reported numerous times, by multiple frameworks or projects, with conflictig solutions, none of which have worked for me.
It sounds like PHP7.2 changes session handling, so that ini_set cannot be executed while an active session is there. We don't call session_start until after ini settings are set in Session:configure(), so this is puzzling.
Will return to it later today.

lonnieezell:
Yikes! Sounds not fun. We are closing the session in teardown(), right? That's the only thing I can think of off the top of my head.

jim-parry:
"Progress" ... the problem is not with session start/stop, it is with headers already sent.
In PHP7.2, if any headers sent, then ini_set reports a warning (Headers already sent) on attempts to change valid session settings (eg session.name) and disallows those changes. 
session_write_close() then session_start() doesn't help. The first succeeds and the second emits a warning "session_start(): Cannot start session when headers already sent in ...".

In PHP7.1, the ini_set produces no warnings, but the last session_start() gives the warning "session_start(): Cannot send session cookie - headers already sent" with a stack trace.

So ... I am pursuing where output might have been commited :-S

lonnieezell:
@jim-parry Hmm... I wonder if it's in the new Feature testing stuff. I can't check right now, but it's possible the response is actually sending out headers and needs to be tweaked to allow for testing. I thought I had taken that into consideration when I did that, but it's possible I didn't...

jim-parry:
Added debugging code in Session, and ob_get_contents() and headers_list() are empty.
I don't see how this is related to feature test stuff, as I don't see that used inside the session testing.
I am stumped, and this PR is not helping to fix the problem :(

lonnieezell:
I was only thinking feature testing might have spit out headers earlier in the testing process which was causing session tests to throw that error. 

Sorry it's being such a PITA.

jim-parry:
Not to worry ... just need to think of a different way to determine the root of the problem, and fix it.
I don't see how the feature tests would cause the problem, but PR1047 is indeed where it started to show :-/
It looks like one of the changes in that PR prevented extraneous output from showing up at the beginning of the unit tests ... I wonder if that is related. As near as I can tell, the session testing doesn't use any of the request/response handling from the feature tests.
I'm going to work on other CI4 stuff for a bit :)

PR #1097, opened 2018-07-09
 auto session start 

bangbangda:
Start automatically when the session is initialized.

Why is it a separate $session->start() ?

Before：
$session = session()->start();
After：
$session = session();

Before：
$session = \Config\Services::session($config);
$session->start();
After：
$session = \Config\Services::session($config);


lonnieezell:
I wasn't involved in the original architecture of the session stuff, I just ported it over. Looking over the main session class I can't see why we shouldn't do this. However, you should put a check within the `start()` method to check if it's running and exit the method at the very beginning of it so that we don't try to start again.

bangbangda:
if sessions are enabled, but none exists. start() method will be executed.
so. will not repeat start().

Please confirm the `system/Config/Services.php` file for details.

lonnieezell:
> if sessions are enabled, but none exists. start() method will be executed.
so. will not repeat start().

True - assuming everyone plays by the rules. :) Someone can still call `start()` if they didn't bother to read the manual closely enough. Also, there are a number of people already building sites on CI4 so it's a simple solution to safeguard their sites a little bit more.

bangbangda:
OK.I understand.
Update the code this afternoon.

bangbangda:
@lonnieezell 
Code update is complete.

bangbangda:
@lonnieezell

bangbangda:
@lonnieezell 
Hi. is there a Issues with this PR?

jim-parry:
@bangbangda Your PR has conflicts that need to be resolved, before travis-ci can be run, and without considering whether or not it is a good idea to auto-start sessions.

lonnieezell:
@bangbangda You merged in changes to the Model class that are already in core. Those should not be part of this, please remove.

bangbangda:
@lonnieezell yes.deletion is complete.

lonnieezell:
Thanks works for me, Thanks @bangbangda

lonnieezell:
Although I just realized this might wreak havoc on testing... but we'll work that out since session tests already broken.

jim-parry:
Hmmm - I remember a similar PR for CI3, and Andrey argued strongly against auto-starting sessions. Don't remember the reasoning.
I don't know that we want to force sessions on evryone, who will then hve to configure them and deal with the hundreds (or much more) of session files resulting from poor or no garbage collection.

I haven't taken a look at the PR, but I trust that any auto-starting is a configurable option.

lonnieezell:
From what I recall of Andrey's reasoning - autostarting the session on every page load could lead to DOS attacks from the sheer number of session being created and some other related security type issues. Completely agree with those. 

What this PR does, though, is to start the session the first time you request it, instead of asking you to do session->start() immediately afterword. I think it should be fine, as I can't think of any additional actions you'd take with the session library before it's started, and none of the methods in the Session class seemed to provide any benefit prior. 

Unless you can think of any issues with that?

jim-parry:
It might work fine, then. A DOS attack could be averted through a Throttle filter :-/

________________________________________

PR #1101, opened 2018-07-15
implements @todo Optimize Escaper instance creation to only load a single instance during page request

samsonasik:
by create a `static variable` of escaper. refs https://3v4l.org/e4gRE

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -276,6 +276,11 @@ if ( ! function_exists('esc'))
 			        $escaper = new \Zend\Escaper\Escaper($encoding);
                        }
 
+                       if ($encoding && $escaper->getEncoding() !== $encoding)
+                       {
+                               $escaper = new \Zend\Escaper\Escaper($encoding);
+                       }
+
 			$data = $escaper->$method($data);
 		}
 
--- a/system/Common.php
+++ b/system/Common.php
@@ -270,8 +270,11 @@ if ( ! function_exists('esc'))
 				$method = 'escape' . ucfirst($context);
 			}
 
-			// @todo Optimize this to only load a single instance during page request.
-			$escaper = new \Zend\Escaper\Escaper($encoding);
+                       static $escaper;
+                       if (! $escaper)
+                       {
+			        $escaper = new \Zend\Escaper\Escaper($encoding);
+                       }
 
 			$data = $escaper->$method($data);
 		}


samsonasik:
the issue that can happen by this is the `$encoding` may be unusable if passed in next `esc()` call.

system/Common.php
```diff
@@ -270,8 +270,11 @@ function esc($data, $context = 'html', $encoding = null)
 				$method = 'escape' . ucfirst($context);
 			}
 
-			// @todo Optimize this to only load a single instance during page request.
-			$escaper = new \Zend\Escaper\Escaper($encoding);
+                       static $escaper;
+                       if (! $escaper)
+                       {
+			        $escaper = new \Zend\Escaper\Escaper($encoding);
```
samsonasik: how about pull `$encoding` (default uses 'utf-8') from `App`configuration ?

lonnieezell: I think that if `$encoding` is null than, yes, pulling it from `App` is appropriate.

system/Common.php
```diff
@@ -270,8 +270,11 @@ function esc($data, $context = 'html', $encoding = null)
 				$method = 'escape' . ucfirst($context);
 			}
 
-			// @todo Optimize this to only load a single instance during page request.
-			$escaper = new \Zend\Escaper\Escaper($encoding);
+                       static $escaper;
```
lonnieezell: I like the idea of optimizing this, since it will get called quite a few times on a page. However, I think we should check the encoding to make sure they match the static version (using `$escaper->getEncoding()`). If they don't it returns a new instance, otherwise returns the static instance. That way we get the benefit of caching it in 99% of the cases, but still have the flexibility if someone needs to do something strange with it.

system/Common.php
```diff
@@ -270,8 +270,16 @@ function esc($data, $context = 'html', $encoding = null)
...
+                       if (! $escaper)
+                       {
+			        $escaper = new \Zend\Escaper\Escaper($encoding);
+                       }
+
+                       if ($encoding && $escaper->getEncoding() !== $encoding)
+                       {
+                               $escaper = new \Zend\Escaper\Escaper($encoding);
+                       }
```
samsonasik: I've updated, when `$encoding` is null, actually, `Zend\Escaper` uses 'utf-8', when user passed non-null of encoding, I added check with `$escaper->getEncoding() !== $encoding`.

PR #1105, opened 2018-07-19
Fixed a bug where you can not change the path freely

ytetsuro:
I wanted to move the `public` directory, but I could not move it.

Since `rewrite.php` before modification depends on the installed directory, it is necessary to create `Serve` class in application and solve it.

This is very troublesome and if spark defines the `FCPATH` constant we thought that you do not have to define $fcpath in rewrite.php.

I think this is the experimentally implemented content at the next PR, but the current situation has arisen.

#996

PR #1108, opened 2018-07-21
Fixed a bug where you can not change the path freely.

ytetsuro:
It is the same intention as the following PR.

#1105 

defined has been DOCUMENT_ROOT by FCPATH in Serve.php.

https://github.com/bcit-ci/CodeIgniter4/blob/develop/system/Commands/Server/Serve.php#L82


kenjis:
It worked fine.
I was able to see Welcome page.

ytetsuro:
Merged Thanks!

Issue #1106, opened 2018-07-20
[TESTS] Session tests for php 7.2


lonnieezell:
Tests are throwing the following error in multiple places, and should not be: 

> ErrorException: ini_set(): Headers already sent. You cannot change the session module's ini settings at this time

lonnieezell:
@jim-parry I believe I found the 2 locations where it's doing this: 

1. the DotEnv tests is writing to the env/putenv/etc and that seems to be causing it
2. the SessionTest files are starting the session. 

Since these are things that we definitely need to test, I'm not sure how best to tackle this. Too bad it's not just a warning, so we could turn warnings off...

jim-parry:
Ok. I will see what I can do refactoring these, perhaps to use VFSStream (the first) and to destroy any session as part of teardown (second one).

jim-parry:
Session testing was already on my list to revisit :)

samsonasik:
I think `session_write_close()` when `session_status() == PHP_SESSION_ACTIVE` can be used for it like I did at zend-session https://github.com/zendframework/zend-session/pull/107

lonnieezell:
Ooo - good thought. We'll try that shortly. Thanks @samsonasik

PR #1110, opened 2018-07-23
WIP Attempt Session fix for PHP7.2 headers already sent error

jim-parry:
Followed suggestions from #1106 and online search results. No joy :(
Doesn't break under PHP7.1, but still does under PHP7.2.
I even isolated the ini_set calls inside a new safeSet method., to try to "play nice".

Further suggestions?

### DIFF
--- a/system/Session/Session.php
+++ b/system/Session/Session.php
@@ -1,4 +1,5 @@
-<?php namespace CodeIgniter\Session;
+<?php
+namespace CodeIgniter\Session;
 
 /**
  * CodeIgniter
@@ -272,7 +273,7 @@ class Session implements SessionInterface
 		}
 		else
 		{
-			ini_set('session.name', $this->sessionCookieName);
+			$this->safeSet('session.name', $this->sessionCookieName);
 		}
 
 		session_set_cookie_params(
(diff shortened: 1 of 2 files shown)


Paradinight:
@jim-parry is it a travis with php7.2 problem or only php7.2?

puschie286:
No, it’s not a bug : https://bugs.php.net/bug.php?id=75227

They added this warning because it doesn’t work this way ( not before 7.2 not now )

Paradinight:
@puschie286 a problem != bug :)

The warning is correct. My question is, if it happend in travis php7.2 or php7.2 too.

puschie286:
Oh, its a raw php change - but it shouldn’t change the behaviour
  

  
  

  
  
>

jim-parry:
It is a "feature" that wasn't enforced before.
BUT, it breaks phpunit locally as well as travis.

jim-parry:
ps the unit tests have worked fine up until php7.2. This is listed in the backwards incompatible changes of 7.2.

samsonasik:
@jim-parry probably `@runInSeparateProcess` in tests can work for it ?

lonnieezell:
@jim-parry is this still valid? I can see it might be good for 7.2 anyway...

samsonasik:
@lonnieezell I think it should be still valid

jim-parry:
I think it is a good idea, setting u up to handle future changes to PHP that might affect us similarly.
It isn't needed though.
I don't have a preference either way. We can always lookup the PR again if needed down the road.

lonnieezell:
@jim-parry I agree. One thing I saw in yii's solution for this is that they saved the session before closing it, then, once it was restarted, they re-applied the session vars. Seems like that would be desirable here.

jim-parry:
I have flagged this as WIP, and can revisit it when I am back from vacation.

jim-parry:
No longer needed - PHP7.2 tests are passing.

PR #1112, opened 2018-07-24
Rework Dotenv testing to use VFS

jim-parry:
In response to #1106


jim-parry:
Doesn't affect PHP7.1 tests; still getting the 28 errors in the PHP7.2 tests.
So, it's arguably better, anr should be merged, but we still have a travis-ci problem :(

PR #1113, opened 2018-07-25
Testing4/http

jim-parry:
Beef up unit testing for HTTP\URI and HTTP\UserAgent.
Remove superfluous code in HTTP\Message.

### DIFF
--- a/system/HTTP/Message.php
+++ b/system/HTTP/Message.php
@@ -1,4 +1,5 @@
-<?php namespace CodeIgniter\HTTP;
+<?php
+namespace CodeIgniter\HTTP;
 
 use CodeIgniter\HTTP\Exceptions\HTTPException;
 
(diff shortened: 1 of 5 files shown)


system/HTTP/Message.php
```diff
@@ -246,13 +240,6 @@ public function getHeaderLine(string $name): string
 			return '';
 		}
 
-		// If there are more than 1 headers with this name,
-		// then return the value of the first.
-		if (is_array($this->headers[$orig_name]))
-		{
-			return $this->headers[$orig_name][0]->getValueLine();
```
lonnieezell: @jim-parry According to the spec you can have more than one header with the same name. Headers like SET_COOKIE and some others I can't think of right now use that ability. It's possible somewhere along the line I simplified, though, and made it so that only 1 could be used, I honestly don't recall at the moment. That was toward the beginning of this 2.5 year journey :) 

What made you decide to remove that capability?

jim-parry: The original code  does not provide for multiple headers. The block at original line 150 will replace an existing header, rather than adding an extra value to it. The original code allowed for a header to have an aeeay value, but not for multiple headers with the same key.
If we are supposed to allow multiple headers with the same key, then Message needs to be modified differently.
At another place, the original comment suggested that with multiple headers for the same key, the original (first) value should be returned, but the original testing asserted that the last value for a key would be returned.
What would you like to see happen?

jim-parry: The block at original line 285 does the same, i.e. line 291 (original) will never be executed.

lonnieezell: Ah - ok, it looks like I decided it was too much trouble. I have vague recollections of running into issues with something when multiple were supported, so it looks like I modified that. I'm fine with removing it for now, then. We can always fix it later if we discovered it was an oops. :)

PR #1116, opened 2018-07-27
Feature/New EntityImmutable class added

ytetsuro:
I think want imutable Entity class.

Therefore, I changed the current Entity class.


lonnieezell:
No thanks. We don't need two Entity classes, and I'm not convinced that has any benefit in this case since the point of an Entity is to hold the current values even after they've been changed so they can be saved back to the database at some point in the future. 

And in removing the set/unset you lose 1/2 of the functionality of the class. 

Feel free to use something like this with your own projects, and share it as it's own OS tool, but not interested in it as part of the core.

ytetsuro:
understanded.
Thank you for reviewing.

PR #1119, opened 2018-07-28
using min($var1, $var2) for compare > check when possible

samsonasik:
ref https://3v4l.org/5TqGI


andreif23:
i dont understand. what's the problem with > ?

samsonasik:
nothing, just to utilize there is function for that

lonnieezell:
No thanks. I don't see any benefit here.

PR #1124, opened 2018-07-31
Controller Filters Bug. 

bangbangda:
If a Response instance is returned,  script execution is not stop.
See application/Filters/Throttle.php for an example. Did not send 429 Code. script execution is not stop.


jim-parry:
Note; contributions should be GPG-signed. See https://bcit-ci.github.io/CodeIgniter4/contributing/signing.html

bangbangda:
@jim-parry OK

PR #1130, opened 2018-08-02
Update View.php

sugenganthos:
Allow views with extension .html, .tpl, empty (assume as .php), etc

### DIFF
--- a/system/View/View.php
+++ b/system/View/View.php
@@ -167,7 +167,9 @@ class View implements RendererInterface
 			$this->saveData = $saveData;
 		}
 
-		$this->renderVars['view'] = str_replace('.php', '', $view) . '.php';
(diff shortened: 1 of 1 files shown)


lonnieezell:
Thanks for this. Looks fine. 

Please add some documentation for this ability. A test wouldn't hurt either.

sugenganthos:
About docs, if you say about inline, i think my inline comment is sufficient. If you say about user documentation, I need some time to find where :)

For test, if you say about PHPUnit and something else, I never use that, and don't know how to use it. Sorry. I just use localhost to test everything.

lonnieezell:
Ok, that's fine about tests. Docs should be added to [this file](https://github.com/bcit-ci/CodeIgniter4/blob/develop/user_guide_src/source/general/views.rst).

Issue #1126, opened 2018-08-01
redirect to route ignore baseurl


puschie286:
there are some redirection problems ( well it only works with absolute paths for me )

e.g redirect()->route( 'ROUTE', [], 302 ) will route to HOST/ROUTE instead of HOST/BaseURL/ROUTE
response->redirect( '/ROUTE' ) have the same result 

`public $baseURL = 'https://*******/WorkDirectory/PHPPlayground/CodeIgniter4/public/';`

reproduce:
1. create a page with multiple uri parts route ( e.g. BaseURL/Test/Redirect/Route )
2. this page redirect to another route ( e.g $this->response->redirect( 'Home' )
3. you will get an two Location headers in the response with `Location: /Home`

* you will be redirected to `https://*******/Home` instead of `https://*******/WorkDirectory/PHPPlayground/CodeIgniter4/public/Home`

puschie286:
currently i use `$this->setHeader('Location', config( App::class )->baseURL.ltrim( $uri, '\\' ) );` as a temporary fix

puschie286:
fixed by #1147

PR #1132, opened 2018-08-03
Fixed problem that redirect does not work when placed in subdirectory.

ytetsuro:
When I set CodeIgniter in the following URL, redirect expects the following behavior.

@link: #1126

### baseURL 

`http://example.com/path/to/codeigniter/`

### requestURL

`http://example.com/path/to/codeigniter/auth/index`

### Redirect expects

`redirect('/auth/login')` -> `http://example.com/path/to/codeigniter/auth/login`
`redirect('login')` -> `http://example.com/path/to/codeigniter/auth/login`

but, it became as follows.

### Redirect actual

`redirect('/auth/login')` -> `http://example.com/auth/login`🙅‍♂️
`redirect('login')` -> `http://example.com/path/to/codeigniter/auth/login`🙆‍♂️

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -568,8 +568,6 @@ class IncomingRequest extends Request
 	 */
 	protected function detectURI($protocol, $baseURL)
 	{
-		$this->uri->setPath($this->detectPath($protocol));
-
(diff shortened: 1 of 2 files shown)


ytetsuro:
sorry this is bad code.

PR #1141, opened 2018-08-06
remove unused use statements

samsonasik:


samsonasik:
travis failure seems nothing to do with it.

PR #1144, opened 2018-08-07
Remove static pages routing

fdjkgh580:
Because it affects normal procedures


lonnieezell:
First - how does this affect normal procedures? 

Second - removing that completely invalidates [this entire page](https://bcit-ci.github.io/CodeIgniter4/tutorial/static_pages.html) of the tutorial.

fdjkgh580:
Hello, this is the problem I encountered, as shown below:

When used Static Pages Routin
image

When removed Static Pages Routin
image

jim-parry:
This looks related to #1240

jim-parry:
I think the solution is to fix the underlying problem, not mask it by dropping the routing rule.

lonnieezell:
This was fixed in da43a39de5808eb5173b848de197afc57c944239

PR #1147, opened 2018-08-07
fix RedirectResponse::route and added test

puschie286:
fixed the default parameter to call the function like docu shows
fixed the missing baseurl in resulting location
added test for route

* relative redirection only works if the baseurl has no sub folder *


sugenganthos:
On CI4.0.0-alpha still failed if CI installed in subfolder. And I didn't see the fix as mentioned on https://github.com/bcit-ci/CodeIgniter4/pull/1132 . Did it canceled?

PR #1154, opened 2018-08-09
Filters rule change: asterisk wildcard

bangbangda:
Rule:
'wechat' => ['except' => ['manage/*', 'admin/*']]
Or
'wechat' => ['except' => ['manage*', 'admin*']]

Before change:
www.bangbangda.me/manage               No
www.bangbangda.me/manage/home    Yes      

After change:
www.bangbangda.me/manage               Yes
www.bangbangda.me/manage/home    Yes


lonnieezell:
First - we don't want `manage` to allow everything after it since that makes it impossible to match a single specific page. 

Second - your change makes it so that a wildcard only matches a single character, which isn't what we want.

PR #1155, opened 2018-08-10
Fixes #1135 : Toolbar oldest file delete and show maximum in the list as App::toolbarMaxHistory

### LINKED ISSUE #1135: Toolbar Oldest files delete bug
system/Debug/Toolbar format function;

			// Oldest files will be deleted
			if ($app->toolbarMaxHistory >= 0 && $i >= $app->toolbarMaxHistory)
			{
				Services::cache()->delete($filenames[$i]);
				continue;
			}

fixed;

			// Oldest files will be deleted
			if ($app->toolbarMaxHistory >= 0 && ($i + 1) >= $app->toolbarMaxHistory)
			{
				Services::cache()->delete($filenames[$i]);
				continue;
			}

Count and file $i start 0.

samsonasik:
Fixes #1135

### DIFF
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -230,6 +230,13 @@ class Toolbar
 
 		for ($i = 0; $i < $total; $i++)
 		{
+                        // Oldest files will be deleted
+			if ($app->toolbarMaxHistory >= 0 && $i+1 > $app->toolbarMaxHistory)
+			{
+				unlink($filenames[$i]);
+				continue;
+                        }
+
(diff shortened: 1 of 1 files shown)


samsonasik:
Before:

before


After:

after

- oldest files deleted first when `$i+1` > `App::toolbarMaxHistory`, which is 20, so, when on the 21th hit, the oldest file will be deleted.
- show list history files maximum as `App::toolbarMaxHistory` property, which is `20`, previously, it shown as `21`.

PR #1159, opened 2018-08-11
reduce str_replace in View/Parser::render()

samsonasik:


system/View/Parser.php
```diff
@@ -129,12 +129,13 @@ public function render(string $view, array $options = null, $saveData = null): s
...
 
-		$view = str_replace('.php', '', $view) . '.php';
+              $template = str_replace('.php', '', $view);
+		$view     = $template . '.php';
 
 		// Was it cached?
 		if (isset($options['cache']))
 		{
-			$cacheName = $options['cache_name'] ?: str_replace('.php', '', $view);
+			$cacheName = $options['cache_name'] ?: $template;
```
lonnieezell: I believe this should be the `$view` variable, shouldn't it? Otherwise you're missing the `.php` extension. At which point, you could have left the previous couple of lines just like they were unless there's a specific use for template and view later?

samsonasik: no, it is the variable that assigned to `$cacheName`, so without ".php", previously, we double call `str_replace` to `$view` to remove ".php". 

$view with ".php" still used in next flow on append to viewPath.

samsonasik:
@lonnieezell I've updated the code to not introduce new variable, so, it uses existing $view variable without `.php` for cacheName

PR #1163, opened 2018-08-16
Create captcha_helper

sugenganthos:
Taken from CI3. Modified to work with CI4

### DIFF
--- a/system/Helpers/captcha_helper
+++ b/system/Helpers/captcha_helper
@@ -1,0 +1,279 @@
+<?php
+/**
+ * CodeIgniter
+ *
+ * An open source application development framework for PHP
+ *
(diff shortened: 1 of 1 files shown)


system/Helpers/captcha_helper
```diff
@@ -0,0 +1,279 @@
...
+			{
+				$$key = $val;
+			}
+			else
+			{
+				$$key = isset($data[$key]) ? $data[$key] : $val;
+			}
+		}
+
+		if ($img_path === '' OR $img_url === ''
```
samsonasik: use `||` instead of `OR`

sugenganthos: Ooppss.. Sorry, forgot about that.

PR #1164, opened 2018-08-17
make name parameter in HTTP\Header mandatory

samsonasik:
if the name is not mandatory, when we create a `HTTP\Header` and call it as string from `__toString()`, we will get the ": " value, which seems incorrect as message header.


lonnieezell:
Hmm. I was thinking there were some headers that only had a single value but I'm not able to find this now, so I"ll merge this for now with the caveat that if I find something along those lines in the future we might have to re-evaluate.

PR #1167, opened 2018-08-18
Don't display debug bar when response body not has "</body>"

samsonasik:
For example: we set non html  in response in controller, e.g.:

```php
	public function index()
	{
		return 'Hello World!';
	}
```

The debug bar should not be displayed.


lonnieezell:
But it should be displayed when returning JSON (since it will be formatted in JSON automatically), which doesn't have a </body> tag. 

Can't accept this solution.

PR #1171, opened 2018-08-21
remove if ($template==forums/categories) check in View/Parser

samsonasik:
I think it was for debugging


lonnieezell:
OH, yeah, oops! Thought I'd gotten that out of there, though I'm still trying to track down why the feature isn't working. Either way, though - this should be out of there. :)

PR #1172, opened 2018-08-21
comparison fix at CIDatabaseTestCase

samsonasik:
previously, it uses assignment which the DBPrefix concated with string "migrations", which if using truthy check, it will always evaluated as `true`, so continue always called at this case. changed `=` to `==`


lonnieezell:
Nice catch. Thanks!

PR #1173, opened 2018-08-22
Update Response.php

sugenganthos:
Remove {csp-style-nonce} from end page when CSP is not enabled. If page send to user contains {csp-style-nonce} or {csp-style-nonce} then it would not pass html check and says 'no value'.

Needs to check about performance because regex will runs to a whole views regardless of CSP setting.


sugenganthos:
Okay then, lets use str_replace instead

PR #1178, opened 2018-08-24
Adjust log level to match RFC 5424

sugenganthos:
Log level threshold is adjusted to match RFC 5424.
Set threshold to 'debug' level practically show 'all' debug messages. This usually the developer wants on debugging sessions

### DIFF
--- a/application/Config/Logger.php
+++ b/application/Config/Logger.php
@@ -18,15 +18,15 @@ class Logger extends BaseConfig
 	|	2 = Alert Messages      - Action Must Be Taken Immediately
 	|   3 = Critical Messages   - Application component unavailable, unexpected exception.
 	|   4 = Runtime Errors      - Don't need immediate action, but should be monitored.
-	|   5 = Debug               - Detailed debug information.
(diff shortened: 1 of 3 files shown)


lonnieezell:
This causes a test to fail. Please adjust the test so that it passes again.

sugenganthos:
I have update the Test. But maybe another test will fail. Sorry, I'm newbie in test unit. This CI4 framework has big test file. I still fail to configure in my local server. Don't know what to do. I have ask in forum but no one answer.

lonnieezell:
I believe the other tests that were failing for you were incorrect. We have a couple of tests that randomly fail due to timestamp issues that we haven't fixed quite yet. All tests are passing now. Thanks.

sugenganthos:
The test on my local server is failed says 'class not found' or something. It is broken path. It resolved if I edit every file and put 'include' on absolute path
Note: I install PHPUnit myself, not using composer included with CI

lonnieezell:
Please ask on the forums for help getting PHPunit working for you. This is for tracking bugs.

PR #1180, opened 2018-08-25
using HTTP\Request instance to pull ip address instead of $_SERVER['REMOTE_ADDR']

samsonasik:


puschie286:
i guess its not necessary to create a new config instance each time -> would use config( App::class ) for the global instance. also thing the same should be possible for HTTP/Request instances or are there reasons not to do so?

samsonasik:
using `config(App::class)` will require to define at `App` class property, while it is calling function inside `HTTP\Request` class. The another solution may add new constant inside `application/Config/Boot/*` or `application/Constants.php`

puschie286:
? its the same as you do, but no new instance ( ok you will have to do config( \Config\App::class ), sry for forgetting namespace )

samsonasik:
What I meant is, using new instance in class property is not doable in php, eg;

```php
class App
{
    public $ip = (new Request())->getIpAddress();
}
```
The alternative is by define as constant

samsonasik:
I've updated to set `ipAddress` property in `__construct()` in each of the class that require it.

jim-parry:
@lonnieezell I am good with this. Ok to merge?

lonnieezell:
That's not exactly what I meant. The preferred way to do this is to grab a copy of the IP address in the services class, [where the driver is instantiated](https://github.com/bcit-ci/CodeIgniter4/blob/develop/system/Config/Services.php#L686) and pass it in the constructor of the driver.

samsonasik:
@lonnieezell implemented

samsonasik:
travis green https://travis-ci.org/bcit-ci/CodeIgniter4/builds/447948785?utm_source=github_status

samsonasik:
as all handlers has ipAddress property, I add ipAddress property to BaseHandler with inject in 2nd parameter.

lonnieezell:
Looks good. Thanks!

Issue #1176, opened 2018-08-22
Feature request. Small but useful. Entity class.


nowackipawel:
There is an feature which allows to cast value of attribute of entity.
Could someone add another options to switch in Entity::castAs($value, string $type)
 i.e. like this:
```
			case 'json-array':
				if (is_string($value) && (strpos($value, '[') === 0 || strpos($value, '{') === 0))
				{
					$value = json_decode($value, true);
				}
				break;
			case 'json-object':
				if (is_string($value) && (strpos($value, '[') === 0 || strpos($value, '{') === 0))
				{
					$value = json_decode($value, false);
				}
				break;
  ```
imho it will be useful.

InsiteFX:
You should ask this in the CodeIgniter CI 4 Forums not here, this is only for reporting bugs etc;

lonnieezell:
I'm not opposed to this. Though I believe there are some security risks when storing json values as an array and not an object, so we might consider not checking for square brackets, just curly braces. 

Care to submit a PR?

jim-parry:
Was this not addressed already, in #1283 ?

lonnieezell:
Yes, it did.

PR #1186, opened 2018-08-30
Small but useful. Entity class. #1176

nowackipawel:
I've changed name which i proposed before (json-object -> json) cuz I use default parameter for json_decode($value, [$assoc = false]) in calling.

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -368,6 +368,7 @@ class Entity
 	 *
 	 * @return mixed
 	 */
+	
 	protected function castAs($value, string $type)
 	{
 		switch($type)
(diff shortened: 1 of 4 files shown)


system/Entity.php
```diff
@@ -388,6 +396,22 @@ protected function castAs($value, string $type)
 					$value = unserialize($value);
 				}
 
+				$value = (array)$value;
+				break;
+			case 'json':
+				if (function_exists('json_decode') && is_string($value) && (strpos($value, '[') === 0 || strpos($value, '{') === 0))
+				{
+					$value = json_decode($value, false);
```
lonnieezell: You should check the json_error_code here, also, to ensure it's valid json, or throw an exception otherwise, so the developer knows what went wrong.

nowackipawel: Firstly I am not convinced about throwing exception when case 'array'  won't throw anything when serialized array is formatted  incorrectly. 

Secondly if you think it is really required to throw exception with json then we have to clarify that
json_error_code() will return error if $value is null; what could be very likely case when data is retrieved from database. Do you reckon we should treat NULL value as exception or just ignore it? i think i should ignore it in that case, as we do in the moment.

--
i will add docs and test in tmrw.

nowackipawel: exception + language  (messages) files added and docs modified

system/Language/en/Cast.php
```diff
@@ -0,0 +1,23 @@
...
+ * @license      https://opensource.org/licenses/MIT	MIT License
+ * @link         https://codeigniter.com
+ * @since        Version 3.0.0
+ * @filesource
+ * 
+ * @codeCoverageIgnore
+ */
+
+return [
+	'Cast.jsonErrorDepth' 		=> 'Maximum stack depth exceeded',
```
lonnieezell: Language strings should not have the filename as part of the key here. So - shouldn't be `Cast.jsonErrorDepth`, should be simply `jsonErrorDepth`.

nowackipawel: done. thx.

nowackipawel:
@lonnieezell  could you check if everything is ok?

lonnieezell:
Merged in PR #1283

PR #1190, opened 2018-08-30
Catch Email Exceptions

puschie286:
catch exceptions that are thrown in the send mail function
-> caused by wrong/invalid mail config

### DIFF
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -188,7 +188,10 @@ class Services extends BaseService
 			$config = new \Config\Email();
 		}
 
-		return new \CodeIgniter\Email\Email($config);
+		$email = new \CodeIgniter\Email\Email($config);
+		$email->setLogger(self::logger(true));
+
+		return $email;
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lonnieezell:
The problem with this is that it hides that actual exception that happened from the developer. I believe that exception will only be thrown if the server isn't setup correctly, right, which is something the developer should know about. 

I haven't had a chance to look through the causes in depth, but if it's something that would only crop up in development, or when testing a new server, it should throw it and fail very loudly so the developer knows. If it's something that can crop up during normal usage, then it should at least be logged so the developer can find the issue later.

puschie286:
Yes, if it is just that it would be fine - but you also have this exception if the Mail Server is busy, failed to establish connection or other problems with the mail server.
  

  
This was the first workaround - so maybe we should check for wrong config exception so we can ignore it ( not catch )
  

  
  

  
  
>

lonnieezell:
Ok, cool. As long as we log it, then, I think it should be fine.

lonnieezell:
Thanks for that addition. I should have been clearer, though. Instead of using Services within the class, I would rather use the LoggerAwareTrait and add the logger instance in the constructor (in system/Config/Services.php)

See the Session class and it's service method for an example.

lonnieezell:
Thanks for doing that. Merging.

PR #1194, opened 2018-09-02
add writable/session directory and set default App::sessionSavePath to it

samsonasik:
to make default file-system based session storage driver work out of the box when `session()` called.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,10 @@ writable/logs/*
 !writable/logs/index.html
 !writable/logs/.htaccess
 
+writable/session/*
+!writable/session/index.html
+!writable/session/.htaccess
+
 writable/uploads/*
 !writable/uploads/index.html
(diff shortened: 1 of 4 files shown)


system/Session/Handlers/FileHandler.php
```diff
@@ -89,7 +89,13 @@ public function __construct($config)
...
 		else
 		{
-			$this->savePath = rtrim(ini_get('session.save_path'), '/\\');
+                       $sessionPath    = rtrim(ini_get('session.save_path'), '/\\');
+                       if (! $sessionPath)
+                       {
+                           $sessionPath = WRITEPATH . 'session';
+                       }
+
+                       $this->savePath = $sessionPath;
```
samsonasik: this is for fallback when `App:: sessionSavePath` is null, but `ini_get('session.save_path')` returns empty string.

PR #1197, opened 2018-09-04
use Validation->getErrors() call instead of Valdation::errors to handle errors that came from session

samsonasik:
on `Validation::getErrors()`, it has handle that errors came from session with key `_ci_validation_errors`

For example, after call:

```php
 return redirect('page')->withInput();
```

On error field, we are calling  at the '/page':

```php
var_dump( $validator->hasError('field'));
```

 it previously won't work as the error is came from the session.


lonnieezell:
Good catch. Thanks!

PR #1198, opened 2018-09-04
add missing function validation_errors() at Common functions

samsonasik:
it is documented at user guide source but the function was not yet defined.

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -1005,3 +1005,14 @@ if (! function_exists('dd'))
 		exit;
 	}
 }
+
+if (! function_exists('validation_errors'))
+{
+	/**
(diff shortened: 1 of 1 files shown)


lonnieezell:
I'm assuming you're talking about in the form helper? If so - this doesn't use the same parameters as the docs show, and isn't in the form helper file. Would also need to tests to be included. 

When you look on the [validation docs](https://bcit-ci.github.io/CodeIgniter4/libraries/validation.html#getting-all-errors) it lists the new way to handle that. 

What should happen, though, is that we remove both that function and the `form_error` function just above it in the docs. I'll do that now. 

Thanks for catching it!

Issue #1201, opened 2018-09-05
multiple rules in validate() for File Upload not working


samsonasik:
Based on documentation, it shown :

```php
$this->validate([
    'avatar' => 'uploaded[avatar]|max_size[avatar,1024]'
]);
```

It will get the following error:

```php
CodeIgniter\Validation\Exceptions\ValidationException: up is not a valid rule.
```

The current working solution is: 

```php
$this->validate([
    'avatar' => [
            'uploaded[avatar]',
            'max_size[avatar,1024]',
        ],
]);
```

is it the documentation issue or the Validation bug ?

samsonasik:
Checking the `Validation::splitRules()`, it got the following: 

```php
/Users/samsonasik/www/CodeIgniter4/system/Validation/Validation.php:738:
array (size=5)
  0 => string 'up' (length=2)
  1 => string 'oa' (length=2)
  2 => string 'ed' (length=2)
  3 => string 'avatar]' (length=7)
  4 => string 'max_size[avatar,1024]' (length=21)
```

The issue seems on it.

osagiesammy:
Thank you @samsonasik I have been on this for more thank three days.

PR #1202, opened 2018-09-05
fix: Bug that can not split correctly when validation rule does not `regex_match`.

ytetsuro:
Based on documentation, it shown :

```php
$this->validate([
    'avatar' => 'uploaded[avatar]|max_size[avatar,1024]'
]);
```

It will get the following error:

```
CodeIgniter\Validation\Exceptions\ValidationException: up is not a valid rule.
```

@see: #1201

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -716,12 +716,14 @@ class Validation implements ValidationInterface
 				$separator = $rules[$open_bracket_pos + 1];
 
 				$regex_end_pos = strpos($rules, $separator . ']');
-				$_rules[] = substr($rules, 0, $regex_end_pos + 2);
+				if ($regex_end_pos !== false) {
+					$_rules[] = substr($rules, 0, $regex_end_pos + 2);
 
-				$rules = substr($rules, $regex_end_pos + 3);
+					$rules = substr($rules, $regex_end_pos + 3);
 
-				$pipe_pos = strpos($rules, '|');
-				continue;
+					$pipe_pos = strpos($rules, '|');
+					continue;
+				}
 			}
 
 			$_rules[] = substr($rules, 0, $pipe_pos);
--- a/tests/system/Validation/ValidationTest.php
+++ b/tests/system/Validation/ValidationTest.php
@@ -503,6 +503,29 @@ class ValidationTest extends \CIUnitTestCase
 
 //--------------------------------------------------------------------
 
+	public function testSplitRuleNotRegexMatchTrue(){
+		$this->validation->setRules([
+			'number' => 'greater_than[10]|less_than[12]',
+		]);
+		$data = [
+			'number' => '11',
+		];
+
+		$result = $this->validation->run($data);
+		$this->assertTrue($result);
+	}
+
+	public function testSplitRuleNotRegexMatchFalse(){
+		$this->validation->setRules([
+			'number' => 'greater_than[10]|less_than[12]',
+		]);
+		$data = [
+			'number' => '16',
+		];
+
+		$result = $this->validation->run($data);
+		$this->assertFalse($result);
+	}
 
 	public function testSplitRulesTrue()
 	{


samsonasik:
I confirm that it fixes the issue 👍

system/Validation/Validation.php
```diff
@@ -716,12 +716,14 @@ protected function splitRules(string $rules): array
 				$separator = $rules[$open_bracket_pos + 1];
 
 				$regex_end_pos = strpos($rules, $separator . ']');
-				$_rules[] = substr($rules, 0, $regex_end_pos + 2);
+				if ($regex_end_pos !== false) {
```
samsonasik: move `{` to next line for consistency

tests/system/Validation/ValidationTest.php
```diff
@@ -503,6 +503,29 @@ public function testHasError()
 
 //--------------------------------------------------------------------
 
+	public function testSplitRuleNotRegexMatchTrue(){
```
samsonasik: move `{` to next line for consistency

ytetsuro:
I think that parsing will fail in the next example.
I do not consider it in this PR.

Should this be fixed with this PR?
Would you like to create a new issue?

```php
$this->validate([
    'date' => 'regex_match[/^[0-9]{4}[\-\.\/][0-9]{2}[\-\.\/][0-9]{2}/]'
]);
```

tests/system/Validation/ValidationTest.php
ytetsuro: Thanks. fixed.

system/Validation/Validation.php
ytetsuro: Thanks. fixed.

samsonasik:
I think this PR should only contains the fix #1201 for easier to review, should be another issue for that.

ytetsuro:
> As for the regex match you're showing: isn't that covered by these tests?

It is not covered.
The code to extract the string of rules is as follows.

https://github.com/bcit-ci/CodeIgniter4/blob/develop/system/Validation/Validation.php#L718

This code will do the following:

1. We are defining the next character of `[` as a delimiter.
1. Get the index of characters up to the first matching `delimiter]`.
1. A character string from the beginning to the index is stored in the rules variable as a validation rule.

The point is,
When writing the following rules, this program works as follows.

`` `php
$ this-> validate ([
    'date' => 'required|regex_match[/^[0-9]{4}[\-\.\/][0-9]{2}[\-\.\/][0-9]{2}/]|max_length[10]'
]);
`` `

Please pay attention to `regex_match` at this time.


1. We are defining the next character of `[` as a delimiter. (`/`)
1. Get the index of characters up to the first matching `/]`.
1. A character string from the beginning to the index is stored in the rules variable as a validation rule.

That is, the stored rules variable is `regex_match[/^[0-9]{4}[\-\.\/]`.

This is not appropriate as a regular expression.
The expected rules variable is `regex_match[/^[0-9]{4}[\-\.\/][0-9]{2}[\-\.\/][0-9]{2}/]`.

Sorry for writing such a long message.

lonnieezell:
Well, sure, but the tests I linked to have multiple brackets inside so I thought it would cover that situation. If it's failing, then by all means - we would love a PR to fix that :) 

For your test, please take a look at how I did the latest test for that. Instead of using some other rule that might fail on its own, I've isolated the test to just the `splitRules` method by creating a method invoker for the protected function.

ytetsuro:
> For your test, please take a look at how I did the latest test for that.

Looked. Thank you tell.

> Well, sure, but the tests I linked to have multiple brackets inside so I thought it would cover that situation. If it's failing, then by all means - we would love a PR to fix that :)
> Instead of using some other rule that might fail on its own, I've isolated the test to just the splitRules method by creating a method invoker for the protected function.

I think it's good.
Likewise, I limited the responsibility range to splitRule method and write the test with #1203.
I am happy if you review it.

PR #1203, opened 2018-09-06
Validation Class - corresponding about the escaped separator.

ytetsuro:
When writing the following rules, this program works as follows.

```php
$ this-> validate ([
    'date' => 'required|regex_match[/^[0-9]{4}[\-\.\/][0-9]{2}[\-\.\/][0-9]{2}/]|max_length[10]'
]);
```

validation rule is `regex_match[/^[0-9]{4}[\-\.\/]`.

but, expected validation rule is `regex_match[/^[0-9]{4}[\-\.\/][0-9]{2}[\-\.\/][0-9]{2}/]`.

Signed-off-by: ytetsuro <phper.0o0@gmail.com>

detail: https://github.com/bcit-ci/CodeIgniter4/pull/1202#issuecomment-418964824

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -720,10 +720,10 @@ class Validation implements ValidationInterface
 			{
 				$separator = $rules[$open_bracket_pos+1];
 
-				$regex_end_pos = strpos($rules, $separator.']');
-
-				if ($regex_end_pos !== false)
+				if (preg_match('/(?<!\\\\)(?:\\\\\\\\)*\\'.$separator.'\]/', $rules, $matches, PREG_OFFSET_CAPTURE))
 				{
+					$regex_end_pos = $matches[0][1];
+
 					$_rules[] = substr($rules, 0, $regex_end_pos+2);
 
 					$rules = substr($rules, $regex_end_pos+3);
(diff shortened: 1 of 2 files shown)


lonnieezell:
I'll be honest - off the top of my head, I couldn't tell you how that particular preg_match regex works lol. All tests are passing, though, so I'll roll with it.

lonnieezell:
@ytetsuro I'm running into an issue when running migrations that seems to be caused by this. Unfortunately, the only error message it gives me is fairly generic: 

> preg_match(): Compilation failed: reference to non-existent subpattern at offset 20
> /Users/kilishan/WebSites/CodeIgniter4/system/Validation/Validation.php - 723

Any ideas?

ytetsuro:
@lonnieezell

Oh... I'm very very very sorry....

It can be solved by changing it as follows.
The intention of this change is to escape anything except those that can not be used as a delimiter.

```diff

+				$separator = (preg_match('/[^a-z0-9\\\s]/i', $separator)) ? '\\'.$separator : $separator;
-				if (preg_match('/(?<!\\\\)(?:\\\\\\\\)*\\'.$separator.'\]/', $rules, $matches, PREG_OFFSET_CAPTURE))
+				if (preg_match('/(?<!\\\\)(?:\\\\\\\\)*'.$separator.'\]/', $rules, $matches, PREG_OFFSET_CAPTURE))
```

However, I noticed when writing this diff, but since the PHP regular expression delimiter has to support cases such as () and [], I think that the processing between these delimiters is not sufficient.

http://php.net/manual/en/regexp.reference.delimiters.php

can i create this diff PR?

lonnieezell:
@ytetsuro yes, please. A PR would be great.

PR #1206, opened 2018-09-07
Port download helper from CI3

jim-parry:
Needed for Compression module.
CI3 had skipped unit tests, reproduced here... will need fleshing out at some point.

### DIFF
--- a/system/Helpers/download_helper.php
+++ b/system/Helpers/download_helper.php
@@ -1,0 +1,167 @@
+<?php
+/**
+ * CodeIgniter
+ *
+ * An open source application development framework for PHP
+ *
+ * This content is released under the MIT License (MIT)
+ *
+ * Copyright (c) 2014-2018 British Columbia Institute of Technology
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


jim-parry:
The unit tests now show a number of risky tests - output buffers not closed!?
Another PR has caused these - I have made no changes to those tests :(

system/Helpers/download_helper.php
```diff
@@ -0,0 +1,167 @@
...
+	 * Generates headers that force a download to happen
+	 *
+	 * @param	mixed	filename (or an array of local file path => destination filename)
+	 * @param	mixed	the data to be downloaded
+	 * @param	bool	whether to try and send the actual file MIME type
+	 * @return	void
+	 */
+	function force_download($filename = '', $data = '', $set_mime = FALSE)
+	{
+		if ($filename === '' OR $data === '')
```
ytetsuro: Is it impossible to realize with the following code?

```php
CodeIgniter\Config\Services::response()->download($filename, $data, $set_mime);
```

https://github.com/bcit-ci/CodeIgniter4/blob/develop/system/HTTP/Response.php#L943

jim-parry: It may very well be - I will check :) Thank you!

jim-parry:
@ytetsuro That does solve my problem, thanks.

PR #1207, opened 2018-09-07
Correct helper tests namespace

jim-parry:
Fix the namespace of the helper unit tests.
Yes, this meant adding some "use" statements, but the namespaces are right :)


jim-parry:
Hmm ... a redis test is failing. Not clear how helper test changes would affect that!

lonnieezell:
Your changes wouldn't. I've started a list that currently has 5 or 6 tests that randomly fail that we need to look into, just hasn't made it to the top of my priority list, yet. :)

PR #1209, opened 2018-09-08
Fix undefined function xml_convert at Database\BaseUtils

samsonasik:
by add xml helper ported from CI3 source


jim-parry:
It looks like this one was missed. Did you want to add the docs page for it, and add it to the helpers/index.rst in the user guide?

samsonasik:
@jim-parry I've added documentation for it, and add `MySQLi` util class for class driver for it when using `MySQL`. It was only for `PostgreSQL`.

PR #1210, opened 2018-09-09
Consolidate honeypot, security & throttle inside filters

jim-parry:
**Description**
The related Honeypot, Security & Throttle folder contents have been added to the Filters folder, to reduce the number of modules.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


jim-parry:
@lonnieezell Filters/HoneypotTest passes when run on its own; not sure why it breaks when run as part of the complete test.

lonnieezell:
I'm not sure what the benefit is here? If other modules have filters, this almost makes it feel like they should be moved out into the Filters namespace, also, which doesn't seem correct to me. 

If we want to keep a possibility down the road of making each module useful on its own, they need to be kept separate, but not if that's still something we're aiming for?

jim-parry:
It seems silly having modules with just the one class (possibly with an exception) - it that class is only going to be used as a filter. There are some other modules that feel funny too, like Typography & Email, though they are not filter-related. 
Just trying to simplify - I don't have strong feelings either way :-/

PR #1217, opened 2018-09-13
Optional parameter for resetSelect() call in Builder's countAll();

nowackipawel:
Identical as is in countAllResult().


lonnieezell:
I'm ok with this addition. Please add to docs.

PR #1218, opened 2018-09-13
Docs/contributing

jim-parry:
Split the "contributing guide" off from the user guide.
This is of interest to framework contributors, and not to developers using the framework.
The *contributing* folder in the repo might be omitted from the release build.


jim-parry:
Apologies for the # of commits, but I had to use the repo to see the effect of .rst changes.

lonnieezell:
I'm not concerned at all with the number of commits. It's all good :)

PR #1224, opened 2018-09-16
try manual install redis extension on php nightly

samsonasik:
Currently, the travis build in php nightly still failures, while we have "allow_failures" for it. This change try to manually install redis extension and check if this is fixes the travis build on php nightly env.


samsonasik:
not working. closing.

PR #1231, opened 2018-09-22
Release build script

jim-parry:
And .... here we go. This is the first crack at the release builder.
It is tested to the "are we there yet?" point, but I am reluctant to proceed with the next level of testing until someone else has taken a peek, to make sure I didn't do somthing stupid.


admin/release
```diff
@@ -0,0 +1,162 @@
...
+echo -e "${BOLD}Updating version dependencies${NORMAL}"
+
+# CI_VERSION definition in system/CodeIgniter.php
+sed -i "/const CI_VERSION/s/'.*'/'${version}${qualifier}'/" system/CodeIgniter.php
+
+# release substitution variable in user_guide_src/source/conf.py
+sed -i "/release =/s/'.*'/'${version}${qualifier}'/" user_guide_src/source/conf.py
+
+# version & date in user_guide_src/source/changelog.rst
+sed -i "/|release|/s/|.*|/${version}${qualifier}/" user_guide_src/source/changelog.rst
```
narfbg: I haven't really checked, but his looks like it might modify older version numbers.

admin/release
```diff
@@ -0,0 +1,162 @@
...
+
+#---------------------------------------------------
+# Are we there yet?
+if [ $action = 'test' ]; then
+    echo -e "${BOLD}Your $branch branch is ready to inspect.${NORMAL}"
+    exit 0
+fi
+echo -e "${BOLD}Are we there yet (yes|no)?${NORMAL}"
+read answer
+if [ $answer != 'yes' ]; then
```
narfbg: I see no exit/return inside this check ... script execution would continue.

admin/release
jim-parry: I edited checked the conf.py, CodeIgniter.php and changelog.rst files, if needed, to make sure there was a unique pattern I could search for. In the case of conf,py and CodeIgniter.php, it was the version's constant declaraiton (only there once). In the case of the changelog,rst, I modified it so that the upcoming release block was driven by the sphinx substitution.
The "fallback" provision is that the new release branch is only built locally, and in the local clone.
I will look into beefing it up a bit to check that there is only the one occurrence of each pattern.

admin/release
jim-parry: Doh - you are right! Thank you for pointing that out :)

jim-parry: Fixed it :)

jim-parry:
I plan to merge this, to get the ball rolling, but to only use the release branch building part for the first release - doing the deployment manually to make sure I didn't forget or mess up anything. The script can be adjusted once we see how that works.

One thing unclear - whether the script will make a pre-release or a full release, from github's perspective. Do I need to do something like the solution [here](https://github.community/t5/How-to-use-Git-and-GitHub/How-to-create-full-release-from-command-line-not-just-a-tag/td-p/6895#), with a curl post inside the script?

lonnieezell:
@jim-parry that sounds like a good idea. Nice work, btw.

PR #1238, opened 2018-09-24
doc fix: replace validation_errors() function with \Config\Services::validation()->listErrors()

samsonasik:
**Checklist:**
- [x] Securely signed commits
- [x] User guide updated

as there is no `validation_errors()` function, and there is a `\Config\Services::validation()->listErrors()` for that.


jim-parry:
Makes sense ... wouldn't you also want to remove the "Note" at the top of the page, about the form helper and validation library not being completed?

samsonasik:
The note removed.

jim-parry:
Thanks :)

PR #1241, opened 2018-09-25
Docs/restructure

jim-parry:
My proposed restructuring of the user guide
- smaller sections
- more task-focused
- easier to digest or browse


lonnieezell:
It's a little hard to tell all of the details from the diff, but looking over the index files, I like the reorganization. Great starting  place.

PR #1242, opened 2018-09-25
workaround for buffer problem

puschie286:
Simple fix for open buffers ( opend by tryToRouteIt )


jim-parry:
That fixes most of them :dagger: Is something similar affecting CLI\ConsoleTest and FeatureTestCaseTest?

Good catch!

puschie286:
yes its the same for ConsoleTest & FeatureTestCaseTest.
commits are coming soon ;)

jim-parry:
Looks good, thanks!

PR #1245, opened 2018-09-26
Fix #1244 (form_hidden declaration)

### LINKED ISSUE #1244: BUG in form_hidden with associative array
There is a bug in the form_hidden function declaration.

In the documentation (https://bcit-ci.github.io/CodeIgniter4/helpers/form_helper.html?highlight=form#form_hidden) we can read :
You can either submit a name/value string to create one field:
```php
form_hidden('username', 'johndoe');
// Would produce: <input type="hidden" name="username" value="johndoe" />
```
... or you can submit an associative array to create multiple fields:
```php
$data = array(
        'name'  => 'John Doe',
        'email' => 'john@example.com',
        'url'   => 'http://example.com'
);

echo form_hidden($data);
```
But the second way doesn't works.

The second parameters ($value) must have a default value.

bvrignaud:
Change the form_hidden function declaration to fix bug #1244

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -144,13 +144,13 @@ if ( ! function_exists('form_hidden'))
 	 * Generates hidden fields. You can pass a simple key/value string or
 	 * an associative array with multiple values.
 	 *
-	 * @param    mixed        $name  Field name
+	 * @param    string|array $name  Field name or associative array to create multiple fields
 	 * @param    string|array $value Field value
 	 * @param    bool         $recursing
 	 *
 	 * @return    string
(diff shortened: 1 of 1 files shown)


lonnieezell:
Thanks for the PR! In the future we ask that contributions be GPG signed. Thanks.

jim-parry:
Just an FYI - I see that the timing test failed, and that the two times were out by 10 seconds. That doesn't sound right at all, and we may have to rethink how we are testing such times. It looks like the timer test is meant to ensure that the formats are the same, regardless of the formatter (Time vs IntlDateFormatter).

PR #1265, opened 2018-09-27
WIP Fix docs re PHP server

### LINKED ISSUE #1240: Route in the News Tutorial Routes are ERROR all over.
Doing the News Tutorial since you changed the routing of the controller.

All the routes are giving a 404 error unless I change them from get and post to add.

Also the create method is throwing a expected array error on the request.
Lonnie something is wrong in the new way the routes and requests are working.

jim-parry:
It looks like the PHP builtin server isn't very useful with CI4, as it doesn't process .htaccess files :(

### DIFF
--- a/user_guide_src/source/installation/index.rst
+++ b/user_guide_src/source/installation/index.rst
@@ -54,6 +54,9 @@ By default, the application will run using the "production" environment. To
 take advantage of the debugging tools provided, you should set the environment
 to "develop".
 
+.. caution:: Using PHP's built-in web server is likely to cause problems,
+	as it does not process the `.htaccess` file used to properly handle requests.
+
 That's it!
 
 If you're new to CodeIgniter, please read the :doc:`Getting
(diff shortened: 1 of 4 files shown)


lonnieezell:
The `php spark serve` command takes care of that limitation by providing a mod_rewrite type functionality...

jim-parry:
Hmm - maybe that should be mentioned in the tutorial :-/
I'm playing with it.

jim-parry:
The tutorial still needs changing. I haven't found a bug in the code yet, but haven't finished playing.

jim-parry:
`php spark serve` helps, but I still get errors trying to serve the default page. /home and /about work fine.

lonnieezell:
Interesting. I haven't pulled down a fresh branch in a while, but I use `serve` for everything when I'm working on CI stuff. Maybe I'll actually have a chance this weekend to look into more.

jim-parry:
Merging this, as it is an improvement though not a solution to the "error" situation going through the tutorial.

Issue #1229, opened 2018-09-21
Time testing in travis-ci wonky


jim-parry:
Travis-ci continues to break intermittently, because one of the tests can be out by 1 second. The error has nothing to do with any changes made in a PR. 

**CodeIgniter 4 version**
develop

**Error message**

1) CodeIgniter\I18n\TimeTest::testTimeWithDateTimeZone
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'2018-09-21 19:56:07'
+'2018-09-21 19:56:08'
/home/travis/build/bcit-ci/CodeIgniter4/tests/system/I18n/TimeTest.php:76

puschie286:
we can reduce the frequency of it occurring by creating the control time right at the beginning

jim-parry:
I'm going to call this one done - I don't see any recurrences of the timing hiccups.
To boot, we now know how to fix them :-/

PR #1273, opened 2018-09-30
Fix/timing

jim-parry:
Re-arrange some of the timing unit tests to avoid fractional/off-by-a-second failures.


jim-parry:
This may not fix absolutely all the timing hiccups - some of the tests can't be broken down or re-arranged further. But, for the most part, it should fix #1229

PR #1277, opened 2018-10-02
Minor improves in system/Test classes (from IDE warnings)

fmertins:
**Description**
1) Some return types like void;
2) phpDoc blocks updates like missing @param, missing @return, etc;
3) Namespace unnecessary usage.

Hi there, this is my first PR to the project, trying to help and improve, thank you.


fmertins:
Hum, ok... now reading how do I sign my commit...

jim-parry:
Making changes to CIUnitTestCase is serious business.
Some of the tests purposefully return something, which may be null, but changing signatures to return void may not be the best idea.
We need to review this carefully.

fmertins:
Hi there, thank you, I can change it if it's necessary. Sorry but right now I'm struggling with sign commits... Ok I have my gpg set, do I need to do a new commit?

jim-parry:
Poissibly even a new PR ... not sure how picky github's branch protection is.
I suggest holding off until Lonnie and/or I can follow the logic chain through.

PR #1278, opened 2018-10-02
A very minimalist commit, just to check signing...

fmertins:
Trying to discover if this commit is signed...


jim-parry:
It isn't signed. If it were, you would see "verified" on the right side of the commit line in this conversation.

fmertins:
Hi, wondering why the heck there are so many steps at GH to do this... :-) Still trying...

PR #1289, opened 2018-10-03
Make it so the trigger arguments are pass by reference

ProjectOrangeBox:
**Description**
Add & so that arguments are passed by references on a trigger call. This allow triggers down the priority "chain" to modify the arguments as needed.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Events/Events.php
+++ b/system/Events/Events.php
@@ -162,12 +162,14 @@ class Events
 	 *  a) All subscribers have finished or
 	 *  b) a method returns false, at which point execution of subscribers stops.
 	 *
+	 * Note: All arguments are passed by reference. This is so the listeners can modify them if necessary.
+	 *
 	 * @param $eventName
 	 * @param $arguments
 	 *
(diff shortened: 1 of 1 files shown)


jim-parry:
1) Your commit isn't GPG-signed ... it would have a "verified" batch on the right side of the conversation above
2) The travis-CI breaks, because the unit testing isn't expecting parameters to be passed by reference - the changes need to be propagated there

lonnieezell:
Yeah, looking at the breaking tests, the syntax we use (`...$args`) doesn't actually allow items to be passed by reference, and I think we need to keep it that way, honestly. This may not be possible without breaking that requirement.

ProjectOrangeBox:
Lonnie, That is understandable, I’ll just make my event library available for everyone once CodeIgniter 4 goes live.

Thank you for your hardworking.

Don Myers

PR #1302, opened 2018-10-09
Exception：No Formatter defined for mime type ''

bangbangda:
If HTTP header Accept is not set, will return the first supported format.

demo:
if "public\js\zepto.js.map" file is  not found, 
log file : Uncaught InvalidArgumentException: No Formatter defined for mime type:''  .........


bangbangda:
Working with Uploaded Files. each file can be save to its new location with the aptly named ``store()`` method.

system/HTTP/Files/UploadedFile.php
```diff
@@ -373,5 +373,26 @@ public function isValid(): bool
...
+     * Save the uploaded file to a new location.
+     *
+     * By default.upload files are saved in writable/uploads directory. the YYYYMMDD folder
+     * and random file name will be created.
+     *
+     * @param string $folderName the folder name to writable/uploads directory.
+     * @param string $fileName   the name to rename the file to.
+     * @return string file full path
+     */
+    public function store($folderName = null, $fileName = null) : string
```
lonnieezell: How is this any different than the move command that you're using in the middle of it?

system/API/ResponseTrait.php
```diff
@@ -346,7 +346,7 @@ protected function format($data = null)
 
 		// Determine correct response type through content negotiation
 		$config = new Format();
-		$format = $this->request->negotiate('media', $config->supportedResponseFormats, true);
+		$format = $this->request->negotiate('media', $config->supportedResponseFormats, false);
```
lonnieezell: What's the reasoning for turning off strict matches here?

system/HTTP/Files/UploadedFile.php
bangbangda: Saving a file is much simpler and requires only one line of code to save the file.

system/API/ResponseTrait.php
lonnieezell: That's fair. I'm good with that change, I think.

Issue #1114, opened 2018-07-25
#1109 breaks route setup/uri parsing


puschie286:
hey, i guess something went wrong with this PR #1109 -> it result in no route will be found
specific: IncommingRequest.php:695-698

im not sure what this change should do (@jim-parry ) but without extracting the URI part the system is unable to find any route -> PageNotFoundException

URI before 695 : /WorkDirectory/PHPPlayground/CodeIgniter4/public/Home
URI after 700 : /WorkDirectory/PHPPlayground/CodeIgniter4/public/Home
excepted : /Home

BaseURL : 192.168.100.111/WorkDirectory/PHPPlayground/CodeIgniter4/public/
$_SERVER[SCRIPT_NAME] : /WorkDirectory/PHPPlayground/CodeIgniter4/public/index.php

jim-parry:
Hmm - it looks like I missed or mis-specified one of the unit tests in tests/system/HTTP/IncomeRequestDetectingTest, or that we need additional test methods.
Could you take a peek and see what looks out of whack (expected resutls for a given URI)?

It looks like I specified SCRIPT_NAME as just /index,php, which is not in the BaseURL you mention above.

puschie286:
still not working - could you explain what your last part is trying to do ?
i dont get why the last check is necessary.
-> you dont handle the fail of your last elseif

i use `else $uri = (string) substr($uri, strlen(dirname($_SERVER['SCRIPT_NAME'])));` after your statement which fixed it for me, but im not sure what is does exactly^^

PR #1304, opened 2018-10-10
Refactor script name stripping in parseRequestURI()

jim-parry:
Even after resolving the earlier problem (#1114), the if/else block wasn't al being tested.
This corrects that, without breaking the unit tests.

@puschie286 Does this still work for your case?
If not, can I have a testcase which breaks it? Thanks


puschie286:
yes every case is handled correctly for me ( sry for late response )

jim-parry:
@puschie286 Thanks for testing :)

PR #1310, opened 2018-10-13
correcting method to get the FCPATH in tests/_support/_bootstrap.php

chris407x:
**Description**

When running tests directly from a different directory (using PHPStorm or phpUnit.phar ) the FCPATH was being set incorrectly: it was using the working directory of the test file instead of the relative path to the _bootstrap.php file.

Changed the method to get the FCPATH from getcwd() to realpath():  
* Matches the format of the other path definitions in this file 
* sets the path relative to _bootstrap.php

The code change is pretty straightforward, the test file proves the change: fail on the old and pass on the new. You may not want to include the test file in the code base because it is fairly complex and would be unnecessary after the code change was made.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/tests/_support/_bootstrap.php
+++ b/tests/_support/_bootstrap.php
@@ -7,8 +7,8 @@ ini_set('display_startup_errors', '1');
 $_SERVER['CI_ENVIRONMENT'] = 'testing';
 define('ENVIRONMENT', 'testing');
 
-// Path to the front controller (this file)
-define('FCPATH', getcwd().'/public'.DIRECTORY_SEPARATOR);
+// path to the directory that holds the front controller (index.php)
+define('FCPATH', realpath(__DIR__.'/../../') . '/public'.DIRECTORY_SEPARATOR);
 
 // The path to the "tests" directory
 define('TESTPATH', realpath(__DIR__.'/../').'/');
@@ -24,5 +24,5 @@ if (! isset($_SERVER['app.baseURL']))
 //--------------------------------------------------------------------
 // Load our TestCase
 //--------------------------------------------------------------------
-
+require_once (TESTPATH . '/../vendor/autoload.php');
 require  __DIR__.'/CIUnitTestCase.php';
--- a/tests/system/Test/BootstrapFCPATHTest.php
+++ b/tests/system/Test/BootstrapFCPATHTest.php
@@ -1,0 +1,90 @@
+<?php
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Class BootstrapFCPATHTest
+ *
+ * This test confirms that the tests/_support/_bootstrap.php
+ * will set the correct FCPATH regardless of the current directory
+ *
+ * It writes a file in the temp directory that loads the bootstrap file
+ * then compares its echo FCPATH; to the correct FCPATH returned
+ * from correctFCPATH();
+ *
+ */
+class BootstrapFCPATHTest extends TestCase
+{
+	private $currentDir = __dir__;
+	private $dir1 = '/tmp/dir1';
+	private $file1 = '/tmp/dir1/testFile.php';
+
+	protected function setUp() {
+		parent::setUp();
+		$this->deleteFiles();
+		$this->deleteDirectories();
+		$this->buildDirectories();
+		$this->writeFiles();
+	}
+
+	protected function tearDown() {
+		parent::tearDown();
+		$this->deleteFiles();
+		$this->deleteDirectories();
+	}
+
+	public function testSetFCPATH(){
+		$result1 = $this->readOutput($this->file1);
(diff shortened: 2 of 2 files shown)


tests/_support/_bootstrap.php
```diff
@@ -7,8 +7,8 @@
 $_SERVER['CI_ENVIRONMENT'] = 'testing';
 define('ENVIRONMENT', 'testing');
 
-// Path to the front controller (this file)
-define('FCPATH', getcwd().'/public'.DIRECTORY_SEPARATOR);
+// path to the directory that holds the front controller (index.php)
+define('FCPATH', realpath(__DIR__.'/../../') . '/public'.DIRECTORY_SEPARATOR);
```
lonnieezell: As long as we're making changes here, we should honor the `public` folder's name that's found in `application/Config/Paths.php`, `$publicDirectory`.

tests/_support/_bootstrap.php
```diff
@@ -24,5 +24,5 @@
 //--------------------------------------------------------------------
 // Load our TestCase
 //--------------------------------------------------------------------
-
+require_once (TESTPATH . '/../vendor/autoload.php');
```
lonnieezell: Were you running into a case where autoload wasn't working here? Should already be included, I believe.

chris407x: Hello Lonnie,

I have been a long time user of CodeIgniter from the 1.x and 2.x days and I'm glad to see version 4 being put together. I would like to give back and volunteer: since the tests are a "help wanted" item I thought that I could start here.

TL:DR:

After further testing I realized that this is not necessary if you run from the composer installed version of phpUnit.  The line is necessary if your are running from a `phpunit.phar`.

If you would rather not change the code here I could add a section to the readme about how to get things to work with `phpunit.phar`.

Details:

This change allows you to run your code with minimal bootstrapping of the system: you aren't running the entire system unless you choose to but you still can extend `CIUnitTestCase`.  If you comment that out, the code below will fail.  This is running from a phpunit.phar file. 

When the tests run from the code igniter root directory as described in `tests/README.md` it picks up the autoloader file by searching in the vendor directory automagically.

PHPStorm has a helper for this and you can also run the code generated from it from the linux console this way (alter for your own path, this is running from phpunit.phar at /home/vagrant/Code/):

The test from a phpunit.phar file (needs the autoloader line):

`/home/vagrant/Code/phpunit --bootstrap /home/vagrant/Code/codeigniter4/tests/_support/_bootstrap.php --no-configuration BootstrapFCPATHTest /home/vagrant/Code/codeigniter4/tests/system/Test/BootstrapFCPATHTest.php --teamcity`

The test uses the composer installed version (does not need the autoloader line):

`/home/vagrant/Code/codeigniter4/vendor/phpunit/phpunit/phpunit --bootstrap /home/vagrant/Code/codeigniter4/tests/_support/_bootstrap.php --no-configuration CodeIgniter\\Cache\\CacheFactoryTest /home/vagrant/Code/codeigniter4/tests/system/Cache/CacheFactoryTest.php --teamcity`

The output looks different but the tests pass.


Cheers,
Chris

tests/_support/_bootstrap.php
chris407x: TL:DR:

- loading the application/Config/Paths.php in this case is not trivial because it has a dependency on FCPATH already being set.
- I would recommend leaving it the string `'public'`. That way if anyone does change it in Config/Paths they will get "file not found" in all their tests.
- `'public'` was a string in the first place, the issues below are probably why
 
Details:

I looked into this and in order to load the `Config\Paths()` the `public/index.php` has to do this:

`define('FCPATH', __DIR__.DIRECTORY_SEPARATOR);
$pathsPath = FCPATH . '../application/Config/Paths.php';
require $pathsPath;
$paths = new Config\Paths()
// then $paths->publicDirectory when you want to see 'public'`

in the test `_bootstrap.php` we are trying to _create_ FCPATH. 

Unfortunately, without FCPATH created no tests will pass (tried this too).

lonnieezell: That's fair. Ignore me on this one. :)

tests/_support/_bootstrap.php
lonnieezell: Hey Chris - definitely appreciate you wanting to jump in and help out! Definitely don't want to discourage that. 

In this case, I think the doc change is best, and leaving that line out. I've found it's generally better to not use a global version and to use a composer-installed version since different versions of frameworks can demand different versions PHPunit to work properly.

chris407x: Ok I'll work on that instead and possibly make a blog post or some other documentation for alternatives/setting up PHPStorm for testing.

jim-parry:
Replaced by #1397

PR #1312, opened 2018-10-14
Add headerEmited (or not) assertions to CUUnitTestCase

jim-parry:
- Enhance system/Test/CUUnitTestCase as above
- update user guide testing/overview page
- update constant definitions in application/Config/Boot since they could be called more than once
- updated server requirements in repo root to match those in the user guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,15 @@ Please read the [*Contributing to CodeIgniter*](https://github.com/bcit-ci/CodeI
 ## Server Requirements
 PHP version 7.1 or higher is required, with the following extensions installed: 
 
-- intl
+- [intl](http://php.net/manual/en/intl.requirements.php)
+- [libcurl](http://php.net/manual/en/curl.requirements.php) if you plan to use the HTTP\CURLRequest library
 
+Additionally, make sure that the following extensions are enabled in your PHP:
+
+- json (enabled by default - don't turn it off)
+- [mbstring](http://php.net/manual/en/mbstring.installation.php)
+- [mysqlnd](http://php.net/manual/en/mysqlnd.install.php)
+- xml (enabled by default - don't turn it off)
 
 ## Running CodeIgniter Tests
 Information on running CodeIgniter test suite can be found in the [README.md](tests/README.md) file in the tests directory.
--- a/tests/system/Test/TestCaseEmissionsTest.php
+++ b/tests/system/Test/TestCaseEmissionsTest.php
@@ -1,0 +1,106 @@
+<?php
+
+namespace CodeIgniter\Test;
+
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -68,8 +68,15 @@ Please read the [*Contributing to CodeIgniter*](https://github.com/bcit-ci/CodeI
 ## Server Requirements
 PHP version 7.1 or higher is required, with the following extensions installed: 
 
-- intl
+- [intl](http://php.net/manual/en/intl.requirements.php)
+- [libcurl](http://php.net/manual/en/curl.requirements.php) if you plan to use the HTTP\CURLRequest library
```
samsonasik: `ext/curl` can be added to `composer.json` in `suggest` part.

jim-parry: Hmm - that package isn't found on packagist.org. Looking at the project homepage, it looks like they expect it to be built from source. It looks like an O/S level thing, rather than a PHP dependency.
I did find a curl/curl, which is an OO wrapper for the curl extension - that still doesn't address installing libcurl.

jim-parry:
ps, on CentOS, libcurl is a standard extension, and can just be yum installed

jim-parry:
@lonnieezell Hoping you can point me in the right direction. When I run phpunit on tests/system/Test, all is well. When they are run against the entire suite, as in travis-ci, the index page is spat out, together with an exception (undefined variable namespace). Sample at https://travis-ci.org/bcit-ci/CodeIgniter4/jobs/441425539. I seem to recall this coming up before, but cannot remember the fix.
Thanks.

tests/system/Test/TestCaseEmissionsTest.php
```diff
@@ -0,0 +1,106 @@
...
+
+		// what do we think we're about to send?
+		$response->setCookie('foo', 'bar');
+		$this->assertTrue($response->hasCookie('foo'));
+		$this->assertTrue($response->hasCookie('foo', 'bar'));
+
+		// send it
+		$response->setBody($body);
+
+		ob_start();
```
lonnieezell: I think this needs to have the `PHP_OUTPUT_HANDLER_CLEANABLE` flag set? According to the docs: 

> The output buffer must be started by ob_start() with PHP_OUTPUT_HANDLER_CLEANABLE flag. Otherwise ob_get_clean() will not work.

README.md
jim-parry: it looks ike ext-curl would be the ticket, for a php extension :) adding that

jim-parry:
ps - It could have been me introducing non-signed commits ... merge commits specifically, synching with latest develop branch. I need to remember to do "git merge -S develop" in my feature branches to synch :-/

PR #1314, opened 2018-10-15
Fix & expand Honeypot & its tests

jim-parry:
- corrected inappropriate method based unit tests
- added unit tests to address properties
- added unit tests to apply honeypot filter
- updated & simplified the Honeypot user guide page

### DIFF
--- a/user_guide_src/source/libraries/honeypot.rst
+++ b/user_guide_src/source/libraries/honeypot.rst
@@ -3,9 +3,9 @@ Honeypot Class
 =====================
 
 The Honeypot Class makes it possible to determine when a Bot makes a request to a CodeIgniter4 application,
-If it's enabled in ``Application\Config\Filters.php`` file. This is done by attaching form fields to any form,
-and this form field is hidden from human but accessible to Bot. When data is entered into the field it's 
-assumed the request is coming from a Bot, then an execption is thrown.
+if it's enabled in ``Application\Config\Filters.php`` file. This is done by attaching form fields to any form,
+and this form field is hidden from a human but accessible to a Bot. When data is entered into the field ,it's 
+assumed the request is coming from a Bot, and you can thro a ``HoneypotException``.
 
 .. contents::
     :local:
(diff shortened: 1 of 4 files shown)


application/Filters/Honeypot.php
```diff
@@ -1,51 +1,45 @@
-<?php namespace App\Filters;
+<?php
+
+namespace App\Filters;
 
 use CodeIgniter\Filters\FilterInterface;
-use CodeIgniter\HTTP\RequestInterface;
-use CodeIgniter\HTTP\ResponseInterface;
+use CodeIgniter\HTTP\IncomingRequest;
```
lonnieezell: These should still be the the Interfaces. Specifying an exact class means that if someone replaces IncomingRequest with their own class, this will throw exceptions, which isn't the desired action.

user_guide_src/source/libraries/honeypot.rst
```diff
@@ -3,9 +3,9 @@ Honeypot Class
 =====================
 
 The Honeypot Class makes it possible to determine when a Bot makes a request to a CodeIgniter4 application,
-If it's enabled in ``Application\Config\Filters.php`` file. This is done by attaching form fields to any form,
-and this form field is hidden from human but accessible to Bot. When data is entered into the field it's 
-assumed the request is coming from a Bot, then an execption is thrown.
+if it's enabled in ``Application\Config\Filters.php`` file. This is done by attaching form fields to any form,
+and this form field is hidden from a human but accessible to a Bot. When data is entered into the field ,it's 
+assumed the request is coming from a Bot, and you can thro a ``HoneypotException``.
```
lonnieezell: I think you missed a `w` on `thro`

application/Filters/Honeypot.php
jim-parry: ok. I changed it to the descendent classes because of methods called that are not in the interfaces. Will fix :)

user_guide_src/source/libraries/honeypot.rst
jim-parry: Oops

PR #1317, opened 2018-10-16
1) Warnings reviews and some phpDoc blocks improvements in system/Tes…

fmertins:
…t files.

2) composer.json, added ext-json as requirement.

### DIFF
--- a/system/Test/CIUnitTestCase.php
+++ b/system/Test/CIUnitTestCase.php
@@ -126,7 +126,7 @@ class CIUnitTestCase extends TestCase
 		$systemPath = realpath(__DIR__.'/../');
 
 		require_once $systemPath.'/'.$this->configPath.'/Paths.php';
-		$this->adjustPaths(new Paths());
+		$paths = $this->adjustPaths(new Paths());
 
 		$app = require $systemPath.'/bootstrap.php';
 		return $app;
--- a/system/Test/CIUnitTestCase.php
+++ b/system/Test/CIUnitTestCase.php
@@ -75,12 +75,13 @@ class CIUnitTestCase extends TestCase
 	 * Custom function to hook into CodeIgniter's Logging mechanism
 	 * to check if certain messages were logged during code execution.
 	 *
-	 * @param string $level
-	 * @param null   $expectedMessage
+	 * @param string      $level
+	 * @param string|null $expectedMessage
 	 *
+	 * @return bool
 	 * @throws \Exception
 	 */
-	public function assertLogged(string $level, $expectedMessage = null)
+	public function assertLogged(string $level, string $expectedMessage = null): bool
 	{
 		$result = TestLogger::didLog($level, $expectedMessage);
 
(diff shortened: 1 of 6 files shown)


fmertins:
Hi folks, sorry but I can't understand why CodeIgniter seems to find the Constants.php file in the public folder... (instead of application folder)

From the Travis CI:
`PHP Fatal error:  require_once(): Failed opening required '/home/travis/build/bcit-ci/CodeIgniter4/public/Config/Constants.php' (include_path='.:/home/travis/.phpenv/versions/7.1.23/share/pear') in /home/travis/build/bcit-ci/CodeIgniter4/system/bootstrap.php on line 98`

system/Test/CIUnitTestCase.php
```diff
@@ -125,7 +126,7 @@ protected function createApplication()
 		$systemPath = realpath(__DIR__.'/../');
 
 		require_once $systemPath.'/'.$this->configPath.'/Paths.php';
-		$paths = $this->adjustPaths(new \Config\Paths());
+		$this->adjustPaths(new Paths());
```
lonnieezell:
```suggestion
		$paths = $this->adjustPaths(new Paths());
```

Removing that return value is why you're seeing the error you're seeing on Travis.

system/Test/ReflectionHelper.php
```diff
@@ -95,8 +89,10 @@ private static function getAccessibleRefProperty($obj, $property)
 	 * @param object|string $obj      object or class name
 	 * @param string        $property property name
 	 * @param mixed         $value    value
+	 * @return void
+	 * @throws \ReflectionException
 	 */
-	public static function setPrivateProperty($obj, $property, $value)
+	public static function setPrivateProperty($obj, string $property, $value): void
```
lonnieezell: Not a fan of enforcing void returns, even though it's semantically correct.

system/Test/CIUnitTestCase.php
lonnieezell: I've found that PHPStorm occasionally lies about random things. :)

fmertins:
Hi, ok I did the correction, but still Travis seems to be broken with SQLite...
Do I need to do something else?

lonnieezell:
> I understood this var is used inside the require but I'm thinking maybe we need another approach to avoid this warning?

I'll be honest, I'm not concerned at all about this warning. And don't listen to Codacy currently as we have never gone in and set it up with our rules. It's an experiment we have had a chance to get back to and decide if it's worth keeping on. 

How would you suggest fixing it? What is here works well, is unobtrusive and solved some situations users ran into in the past with reorganizing folders, etc. If you can show an elegant solution that doesn't change a bunch of things just to avoid a warning, then cool. Otherwise - I'm not bothered by it.

fmertins:
Morning! OK thanks, I was thinking something like "if Codacy or any other tool shows me an error/X I did something wrong and probably won't be accepted". I agree with the actual code and I'll probably move forward looking other files to improve.

lonnieezell:
Awesome. Thanks for your contributions!

PR #1322, opened 2018-10-18
Fix doc error : show_404() doesn't exist any more

bvrignaud:
Fix doc error
show_404() doesn't exist any more


bvrignaud:
Commit not GPG signed

PR #1325, opened 2018-10-18
WIP Add support for non-empty string as a primaryKey in update() 

nowackipawel:
**Description**
In current code every update which contains ID in string format was ignored inside update() method.
Actually in docs (and comments) there is an information that $id passed to update() method can contain an array... - I think it won't work as well; I will try to write a fix in a week but I am not sure about exact date so you can merge this little fix without waiting.

PR #1213, opened 2018-09-11
URI segment as page number in Pagination

nowackipawel:
**Description**
Allows to pass URI segment as the page number to $pager->makeLinks().

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPdocs
- [x] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #1334, opened 2018-10-22
Proposed changelog for alpha.2

jim-parry:
@lonnieezell What do you think? It is complete, in terms of which components were changed, but it doesn't say much about what changes were made. I think it we want that, then we need to update the changelog with each merge.

It is based on the PRs merged since alpha.1 ... anything merged directly, and not through a PR, is not included.
PR titles are not useful for automating this, and commit messages are even less useful.

### DIFF
--- a/user_guide_src/source/changelog.rst
+++ b/user_guide_src/source/changelog.rst
@@ -2,6 +2,133 @@
 Change Log
 ##########
 
+Version |next|
+=================================
+
+Release Date: Not Released
+
+**Second alpha release of CodeIgniter4**
+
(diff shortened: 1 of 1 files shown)


nowackipawel:
@jim-parry  what about: Entity #1213  and  also Model #1325  (which is linked with already merged #1189 )

jim-parry:
@nowackipawel  The first two you mention have not been merged yet, and #1189 was merged Aug 30, i.e. part of alpha.1
This changelog is only for what has been merged as of right now.

PR #1351, opened 2018-10-25
Update View.php

zl59503020:

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345

### DIFF
--- a/system/View/View.php
+++ b/system/View/View.php
@@ -188,7 +188,7 @@ class View implements RendererInterface
 
 		if ( ! file_exists($this->renderVars['file']))
 		{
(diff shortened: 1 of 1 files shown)


jim-parry:
Your change has nothing to do with finding a file - it just reinforces the notion that a file extension for a view would have a php extension by default. That is already inherent in the finction signature.
If you want to use an HTML file as a template, the assumed location would be application/Views, but as HTML it will not benefit from PHP processing.

This is sounding more like a support question (for the forum), and not appropriate for a PR.

zl59503020:
Ok. Yes, but if the template I am using is not a .php file, it will not work properly.

PR #1362, opened 2018-10-26
Validation - is_unique_empty rule

nowackipawel:
New Validation rule is_unique_empty - value should be unique in db (as is in is_unique) or empty (is_null($value) || $value == '')

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -174,6 +174,29 @@ class Rules
 	}
 
 	//--------------------------------------------------------------------
+	
+	/**
+	 * Checks the database to see if the given value is unique or empty.
+	 *
+	 * Example:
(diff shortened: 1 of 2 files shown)


lonnieezell:
Would prefer a more generic rule, `empty`, that stands on it's own. Otherwise we end up with a number of potential rules with empty tacked onto the end of them. 

This would be similar to Laravel's `nullable` rule.

nowackipawel:
That concrete rule (is_unique_empty) will be useful in database when field has has to be unique or null. 
I am not sure if this will work:
``` 'field1' -> 'is_unique[table,field,2]|nullable' ```
, for now i don't think so cuz   is_unique[table,field,2] will return false everytime when there is any null value in  table.field1

PR #1364, opened 2018-10-27
Unique checking while update

kiranrs:
There should be an option for unique check in update stage. Currently there is no option in CI 3. Currentl I edited system/Form_Validation.php in CI 3 to accomplish this.

if (substr_count($field, '.')==3)
      {
         list($table,$field,$id_field,$id_val) = explode('.', $field);
         $query = $this->CI->db->limit(1)->where($field,$str)->where($id_field.' != ',$id_val)->get($table);
      } else {
         list($table, $field)=explode('.', $field);
         $query = $this->CI->db->limit(1)->get_where($table, array($field => $str));
      }
      
      return $query->num_rows() === 0;

    }

Usage Example: $this->form_validation->set_rules('username', 'User Name', 'is_unique['centrename','Centre Name','tbl_login.user_name','user_id.'.$id]'); where $id is the existing id of the updating data.


jim-parry:
Your PR is backwards - you should be merging your feature branch into develop.
I don't think that any of the 723 files changed pertain to what you are trying to accomplish.

Issue #1367, opened 2018-10-27
guessExtension()


nowackipawel:
Am I wrong or this code does not have sense:

```
	public static function guessExtensionFromType(string $type) // $type contains MIME TYPE like 'image/jpeg'
	{
		$type = trim(strtolower($type), '. ');

		foreach (self::$mimes as $ext => $types)
		{
			if (is_string($types) && $types == $type)
			{
				return $ext;
			}
			else if (is_array($types) && in_array($type, $types))
			{
				return $ext;
			}
		}

		return null;
	}
```

for extensions like, i.e.:

```

	public static $mimes = [

		'mid'   => 'audio/midi',
		'midi'  => 'audio/midi',
		'mpga'  => 'audio/mpeg',
		'mp2'   => 'audio/mpeg',

		'jpg'   => ['image/jpeg', 'image/pjpeg'],
		'jpeg'  => ['image/jpeg', 'image/pjpeg'],
		'jpe'   => ['image/jpeg', 'image/pjpeg'],
		'jp2'   => ['image/jp2', 'video/mj2', 'image/jpx', 'image/jpm'],
		'j2k'   => ['image/jp2', 'video/mj2', 'image/jpx', 'image/jpm'],
		'jpf'   => ['image/jp2', 'video/mj2', 'image/jpx', 'image/jpm'],
		'jpg2'  => ['image/jp2', 'video/mj2', 'image/jpx', 'image/jpm'],
		'jpx'   => ['image/jp2', 'video/mj2', 'image/jpx', 'image/jpm'],
		'jpm'   => ['image/jp2', 'video/mj2', 'image/jpx', 'image/jpm'],
		'mj2'   => ['image/jp2', 'video/mj2', 'image/jpx', 'image/jpm'],
		'mjp2'  => ['image/jp2', 'video/mj2', 'image/jpx', 'image/jpm'],
];
```

for image/jpeg it always returns 'jpg' even if real file extension is jpeg or jpe which are also correct;

PR #1368, opened 2018-10-27
UploadedFile->guessExtenstion() & Config/Mimes::guessExtensionFromType() improvements.

nowackipawel:
#1367 
guessExtension() more predictable

### DIFF
--- a/application/Config/Mimes.php
+++ b/application/Config/Mimes.php
@@ -302,12 +302,21 @@ class Mimes
 	 * Attempts to determine the best file extension for a given mime type.
 	 *
 	 * @param string $type
+	 * @param string $proposed_extension - default extension (in case there is more than one with the same mime type)
 	 *
 	 * @return string|null The extension determined, or null if unable to match.
 	 */
-	public static function guessExtensionFromType(string $type)
+	public static function guessExtensionFromType(string $type, ?string $proposed_extension = null)
 	{
+
 		$type = trim(strtolower($type), '. ');
+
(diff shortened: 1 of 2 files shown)


lonnieezell:
The point of that method is to guess the file type without any user-supplied input and your variation uses the client-supplied file extension which is a potential security issue. With your version someone could pass a potentially harmful file but give it an extension that appeared harmless. If the app was poorly programmed it could potentially be something that gets executed, or be javascript that gets displayed in someone's browser when they thought it was an image, etc.

lonnieezell:
I'm sorry. I read the logic wrong the other night. What you're doing there actually does work and still keeps things safe. Sorry about that. Will merge.

nowackipawel:
thx.

daljit3:
Hi - there is an issue with the new update or maybe due to changes by someone else.

**in Config/Mimes.php** this method below returns null.
```
public static function guessExtensionFromType(string $type, ?string $proposed_extension = null) {
return null;
}
```
in **System/HTTP/Files/UploadedFile.php** return type must be string - so the null return from the above throws a critical error.
```
public function guessExtension(): string 
{
		return \Config\Mimes::guessExtensionFromType($this->getMimeType(), $this->getClientExtension());
}

```

daljit3:
Shall I report it as a bug ?

nowackipawel:
Will fix it asap - tmrw

nowackipawel:
My defs of this method was exact the same as was in File . However type hinting should be ?string as is in Config\Mimes. Pr#1399

Issue #1359, opened 2018-10-26
Entity json-array casting not functioning as expected


chistel:
the json-array casting decodes the json data when submitting to db.
an instance is this,` [{"action_desription":"Order Approved by - \t@Staff[1]","action_time":"07-10-2018 14:14:28","action_ip":"127.0.0.1"}]` which is what is supposed to return to the db, instead it attempts submitting the raw array 
```
    ["action_desription"]=>"Order Approved by - @Staff[1]"
    ["action_time"]=> "07-10-2018 14:14:28"
    ["action_ip"]=> "127.0.0.1"
```

its issues revolves around the entity method toArray, which returns the raw data using the __get method.

nowackipawel:
Could you please check if you've got the same problem using 'array' as a casing method? (value should be serialized then).

Could you provide code where you're trying full entity and save data?

jim-parry:
Closed by #1373?

PR #1373, opened 2018-10-28
Turning off casting for db insert/save 

### LINKED ISSUE #1372: boolean cast in entity and validation error for required,  in_list[0,1]
As we know we've got casting in entity class which could cast boolean values to integer (for database) and integers 0,1 to false, true from db.

And I am wondering how to fix that case:

My entity class: 
```
<?php namespace App\Entities\File;

use CodeIgniter\Entity;

class FileConverted extends Entity
{
	protected $fco_id;
	protected $fco_is_active;

	protected $_options = [
		'casts' => [
			'fco_id'			=> 'integer',
			'fco_is_active'		=> 'boolean', //this cast
		],
		'dates'=> ['fco_created_at', 'fco_updated_at'],
		'datamap' => []
	];
```

and model:

```
<?php

class FileConvertedModel extends \CodeIgniter\Model
{
	protected $DBGroup 		= DB_ZTN_GROUP;
	protected $table 		= 't_file_converted';
	protected $primaryKey 	= 'fco_id';
	protected $returnType 	= 'App\Entities\File\FileConverted';

	protected $allowedFields = [
		'fco_is_active',
	];

	protected $validationRules 	= [
		'fco_fil_id'				=> 'required|is_natural_no_zero|less_than_equal_to[16777215]',
		'fco_is_active'				=> 'required|in_list[0,1]',
	];
}
```

and now from other model I am doing sth like this:

```

$file_converted->fco_fct_identifier = $file_conversion_task->fct_identifier;
$file_converted->fco_is_active = 0;
$file_converted->setConvertedName($file->guessExtension());
$file_converted_model->save($file_converted);
```
and I've got error:

```
The fco_is_active field is required.
```
bacause is_required gets FALSE instead of 0 which I assigned to fco_is_active.

For quick fix we could make new rule like 'valid_boolean', use it at one of the first rules and  change $value inside of processRules to false -> 0, true -> 1  for other validation rules like required or in_list[0,1]

What do you think? Is it a bug o proper behavior ?

Or is any way to get casts of saved entity in  Rules class methods without passing every cast through Model::save() , Validation::run() and Validation::processRules()  ?

nowackipawel:
Turning cast for db save/update and validation inside of Model's::classToArray , will fix #1372 and probably #1359;

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -144,7 +144,7 @@ class Entity
 	 *
 	 * @return array
 	 */
-	public function toArray(bool $onlyChanged = falsebool $cast = true): array
+	public function toArray(bool $onlyChanged = false, bool $cast = true): array
 	{
 		$this->_cast = $cast;
 		$return = [];
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -74,6 +74,13 @@ class Entity
 	protected $_original = [];
 
 	/**
+	* Holds info whenever prperties have to be casted
+	*
+	* @var boolean
+	**/
+	private $_cast = true;
+	
+	/**
 	 * Allows filling in Entity parameters during construction.
 	 *
 	 * @param array|null $data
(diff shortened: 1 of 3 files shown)


jim-parry:
Broken unit test... ParserTest::testParseLoopEntityProperties

nowackipawel:
thx, @jim-parry

PR #1376, opened 2018-10-29
Fix cannot use class Honeypot because already in use in App\Filters\Honeypot

samsonasik:
The class named "Honeypot", while it uses `CodeIgniter\Honeypot\Honeypot` class in use statement. Make alias of it as `BaseHoneypot` to avoid conflict.

### DIFF
--- a/application/Config/Filters.php
+++ b/application/Config/Filters.php
@@ -9,7 +9,7 @@ class Filters extends BaseConfig
 	public $aliases = [
 		'csrf' 	  => \App\Filters\CSRF::class,
 		'toolbar' => \App\Filters\DebugToolbar::class,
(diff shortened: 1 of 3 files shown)


jim-parry:
Instead of introducing an alias for Honeypot, why not rename application/Filters/Honeypot to application/Filters/HoneypotFilter ... that seems better suited to the intent of the class.

samsonasik:
@jim-parry I've renamed to HoneypotFilter

samsonasik:
@jim-parry actually, there is already `Services::honeypot()` function in `Config\Services` class, I've updated to : 

```php
$honeypot = Services::honeypot(new \Config\Honeypot());
```

jim-parry:
even better :)

PR #1386, opened 2018-10-30
fix. extend namespace

Instrye:
**Description**
`Pages extends CodeIgniter\Controller` replace to `Pages extends \CodeIgniter\Controller`
`CodeIgniter\Controller` class in root namespace

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/tutorial/static_pages.rst
+++ b/user_guide_src/source/tutorial/static_pages.rst
@@ -30,7 +30,7 @@ code.
 ::
 
 	<?php namespace App\Controllers;
-	class Pages extends CodeIgniter\Controller {
+	class Pages extends \CodeIgniter\Controller {
 
 		public function view($page = 'home')
 		{


jim-parry:
@lonnieezell What convention should we be following?
Some of the examples in the user guide use namespaces (App\Controllers) while others don't.
Some extend "\Codeigniter\..." while others extend "CodeIgniter\...".

lonnieezell:
That's because, in part, I didn't think it was necessary to repeat the full structure every time, but maybe that is a better practice. 

- All controllers should have the `App\Controllers` namespace
- I would always have a `use` statement above the class: `use CodeIgniter\Controller`
- so only ever have to `MyClass extends Controller`

jim-parry:
There are a bunch of instances where this needs to be corrected, and following Lonnie's suggestion, not the path taken by this PR. I am closing this, and will make a sweep of the user guide looking for these changes.
Thank you @Instrye for bringing this to our attention :)

PR #1388, opened 2018-10-30
Phpcs Auto-fix on commit

lonnieezell:
This integrates auto code fixes with PHPCS using [CodeIgniter4-Standard](https://github.com/bcit-ci/CodeIgniter4-Standard) on commit to help ensure code consistency. 

- added CodeIgniter4-Standard to composer.json
- added phpcs to composer.json
- added new post-install script that copies a bash script into the project's .git hooks. 
- new bash script that lints and fixes any fixable errors automatically for each PHP file during commit.

### DIFF
--- a/admin/pre-commit
+++ b/admin/pre-commit
@@ -1,0 +1,40 @@
+#!/bin/sh
+
+PROJECT=`php -r "echo dirname(dirname(dirname(realpath('$0'))));"`
+STAGED_FILES_CMD=`git diff --cached --name-only --diff-filter=ACMR HEAD | grep \\\\.php`
+
+# Determine if a file list is passed
+if [ "$#" -eq 1 ]
+then
+	oIFS=$IFS
+	IFS='
+	'
+	SFILES="$1"
+	IFS=$oIFS
+fi
+SFILES=${SFILES:-$STAGED_FILES_CMD}
+
+echo "Checking PHP Lint..."
+for FILE in $SFILES
+do
+	php -l -d display_errors=0 $PROJECT/$FILE
+	if [ $? != 0 ]
+	then
+		echo "Fix the error before commit."
+		exit 1
+	fi
+	FILES="$FILES $PROJECT/$FILE"
+done
+
+if [ "$FILES" != "" ]
+then
+	echo "Fixing formatting with Code Sniffer..."
+	./vendor/bin/phpcbf --standard=./vendor/codeigniter4/codeigniter4-standard/CodeIgniter4 --encoding=utf-8 -n -p $FILES
+	if [ $? != 0 ]
+	then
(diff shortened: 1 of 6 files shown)


public/index.php
```diff
@@ -4,12 +4,12 @@
 $minPHPVersion = '7.1';
 if (phpversion() < $minPHPVersion)
 {
-	die("You PHP version must be {$minPHPVersion} or higher to run CodeIgniter. Current version: ". phpversion());
+	die("You PHP version must be {$minPHPVersion} or higher to run CodeIgniter. Current version: " . phpversion());
```
jim-parry: Picky, but "your PHP version..."

admin/pre-commit
```diff
@@ -0,0 +1,40 @@
...
+fi
+SFILES=${SFILES:-$STAGED_FILES_CMD}
+
+echo "Checking PHP Lint..."
+for FILE in $SFILES
+do
+	php -l -d display_errors=0 $PROJECT/$FILE
+	if [ $? != 0 ]
+	then
+		echo "Fix the error before commit."
```
jim-parry: CS told me that 160 errors were fixed in 1 file (that I purposely changed a few lines in), and that 82 remained.
No indication of remaining changes expected, but the file was updated.
I re-added it, and re-ran my commit, and it took it.
Is that expected? If so, the error message might be misleading.

admin/pre-commit
```diff
@@ -0,0 +1,40 @@
...
+	FILES="$FILES $PROJECT/$FILE"
+done
+
+if [ "$FILES" != "" ]
+then
+	echo "Running Code Sniffer..."
+	./vendor/bin/phpcbf --standard=./vendor/codeigniter4/codeigniter4-standard/CodeIgniter4 --encoding=utf-8 -n -p $FILES
+	if [ $? != 0 ]
+	then
+		echo "Fix the error before commit."
```
jim-parry: CS told me that 160 errors were fixed in 1 file (that I purposely changed a few lines in), and that 82 remained.
No indication of remaining changes expected, but the file was updated.
I re-added it, and re-ran my commit, and it took it, reporting that no fixable errors were found.
Is that expected? If so, the error message might be misleading.

public/index.php
lonnieezell: Very true -that's what I get for copy pasting and not reviewing in  enough detail. :)

admin/pre-commit
lonnieezell: Oh, no it should not exit like that. The intent of this is just to have it fix the things it can fix, which is mostly code-formatting stuff. The original article had it just checking and exiting if there were any errors. Figured that as overkill and going to put a huge damper on contributions so left changed it to just fix what it could.

PR #1392, opened 2018-10-31
Fix language file

vibbow:
Fix language file `HTTP`, change `cannotSetFilepath` to `cannotSetFilePath`

As is was referenced as `cannotSetFilePath` in `system\Exceptions\DownloadException.php:13`

**Checklist:**
- [ Y ] Securely signed commits
- [ N/A ] Component(s) with PHPdocs
- [ N/A ] Unit testing, with >80% coverage
- [ N/A ] User guide updated
- [ N/A ] Conforms to style guide

### DIFF
--- a/system/Language/en/HTTP.php
+++ b/system/Language/en/HTTP.php
@@ -34,7 +34,7 @@ return [
 
 	// DownloadResponse
 	'cannotSetBinary'            => 'When setting filepath can not set binary.',
-	'cannotSetFilepath'          => 'When setting binary can not set filepath: {0}',
+	'cannotSetFilePath'          => 'When setting binary can not set filepath: {0}',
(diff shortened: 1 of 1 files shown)


jim-parry:
If this is only referenced in DownloadException, it would be better to change that file rather than the language file, as there are multiple translations already that would use the original language key.

system/Language/en/HTTP.php
```diff
@@ -34,7 +34,7 @@
 
 	// DownloadResponse
 	'cannotSetBinary'            => 'When setting filepath can not set binary.',
-	'cannotSetFilepath'          => 'When setting binary can not set filepath: {0}',
+	'cannotSetFilePath'          => 'When setting binary can not set filepath: {0}',
```
jim-parry: Change this in HTTP instead

PR #1394, opened 2018-10-31
add php 7.3 to travis config

samsonasik:
since php: nightly already use php 7.4dev

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ global:
 
 # Recommended by Travis support
 sudo: required
-dist: precise
+dist: trusty
 
 env:
   - DB=mysqli
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ global:
 
 # Recommended by Travis support
 sudo: required
-dist: precise
+dist: trusty
 group: edge
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ global:
 
 # Recommended by Travis support
 sudo: required
-dist: trusty
+dist: precise
+group: edge
 
 env:
   - DB=mysqli
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,13 @@ language: php
 php:
   - 7.1
   - 7.2
+  - 7.3
   - nightly
 
 matrix:
   fast_finish: true
   allow_failures:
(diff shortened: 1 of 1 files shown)


samsonasik:
it seems php 7.4dev on nightly currently only on `dist: trusty`. Current travis config uses `dist: precise` which show as `7.3.0-dev` in nightly. 

The details is as follow:

1. When using `precise`, the `7.3` will be shown as `7.3.0RC3` and the `nightly` will be shown as `7.3.0-dev`.

2. When using `trusty`, the `7.3` will be shown as `7.3.0RC3` and the `nightly` will be shown as `7.4.0-dev`.

samsonasik:
the reason seems by use `group: edge`, it will be shown as `7.3` => `7.3.0RC3`, and `nightly` => `7.4.0-dev`. I also updated to `dist: trusty` as the `Precise` is in the process of being decommissioned.

screen shot 2018-11-01 at 2 20 58 am

- 7.3.0RC3 travis build: https://travis-ci.org/bcit-ci/CodeIgniter4/jobs/449022333
- 7.4.0-dev travis build: https://travis-ci.org/bcit-ci/CodeIgniter4/jobs/449022338

jim-parry:
It looks like memcached is not supported on 7.3 or 7.4 yet ? :(

samsonasik:
yes, it seems not yet supported, but probably, we can manually install something like:

```bash
   - if [[ $TRAVIS_PHP_VERSION = 7.3 ]]; then git clone https://github.com/php-memcached-dev/php-memcached.git && cd php-memcached && phpize && ./configure  && make && make install && echo "zend_extension = memcached.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini && cd ..; fi 
```

worth a try ?

jim-parry:
with memcached not supported yet (per their website), I am not convinced that it makes sense to add 7.3 to the list of versions. We know it will fail.
Using "dist: trusty" makes sense, given that "dist:trusty" is being phased out.
I can't find an explanaiton of "group:edge" on the travis-site, and only some allusions to behaviour when it is used. What does this buy us?

jim-parry:
I see that you specified that the 7.3 build is allowed to fail, so it won't prevent us from moving forward with 7.1 & 7.2 ... will merge

PR #1396, opened 2018-11-01
only register PSR4 "Tests\Support" namespace in "testing" environment

samsonasik:
By this, the "Tests\Support" namespace won't be registered in `development` and `production` env which is not used.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/application/Config/Autoload.php
+++ b/application/Config/Autoload.php
@@ -57,11 +57,7 @@ class Autoload extends \CodeIgniter\Config\AutoloadConfig
 
 		if (defined('ENVIRONMENT') && ENVIRONMENT === 'testing')
 		{
-			$psr4 = array_merge(
-				$psr4,
-				[
-					'Tests\Support' => TESTPATH . '_support',    // So custom migrations can run during testing
-				]);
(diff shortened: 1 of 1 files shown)


application/Config/Autoload.php
```diff
@@ -50,12 +50,20 @@ public function __construct()
...
-			'App'                        => APPPATH,			    // To ensure filters, etc still found,
-			'Tests\Support'              => TESTPATH.'_support',    // So custom migrations can run during testing
+			'Config'      => APPPATH . 'Config',
+			APP_NAMESPACE => APPPATH,                // For custom namespace
+			'App'         => APPPATH,                // To ensure filters, etc still found,
 		];
 
+		if (defined('ENVIRONMENT') && ENVIRONMENT === 'testing')
+		{
+			$psr4 = array_merge(
```
lonnieezell: Why `array_merge` here? Seems like just adding the element directly to the array is more straightforward: 

```
$psr4['Tests\Support'] = TESTPATH .'_support';
```

samsonasik: implemented

PR #1397, opened 2018-11-01
Correcting FCPATH setting in tests/_support/_bootstrap.php

chris407x:
**Description**

Changed the method to get the FCPATH from getcwd() to realpath() to set paths relative to the bootstrap file. This allows tests to be run independently from their directories or using IDEs like PHPStorm.

This is a correction/update to pull request #1310.  It is essentially the same except there is no change to any autoloading behavior.

if you run the test without the change to _bootstrap.php, it will fail and all other tests will fail if you run them from their directory like this:

Before the fix this will fail (run from bash in linux, you may need to adjust your paths):

`/usr/bin/php /home/vagrant/Code/codeigniter4/vendor/phpunit/phpunit/phpunit --bootstrap /home/vagrant/Code/codeigniter4/tests/_support/_bootstrap.php --no-configuration FeatureTestCaseTest /home/vagrant/Code/codeigniter4/tests/system/Test/ReflectionHelperTest.php `

`PHP Warning:  Uncaught require_once(/Config/Constants.php): failed to open stream: No such file or directory`

(FCPATH is wrong)

After the _bootstrap.php fix it will work:

`OK (7 tests, 7 assertions)`

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/tests/_support/_bootstrap.php
+++ b/tests/_support/_bootstrap.php
@@ -7,8 +7,8 @@ ini_set('display_startup_errors', '1');
 $_SERVER['CI_ENVIRONMENT'] = 'testing';
 define('ENVIRONMENT', 'testing');
 
-// Path to the front controller (this file)
-define('FCPATH', getcwd().'/public'.DIRECTORY_SEPARATOR);
+// path to the directory that holds the front controller (index.php)
+define('FCPATH', realpath(__DIR__.'/../../') . '/public'.DIRECTORY_SEPARATOR);
 
 // The path to the "tests" directory
 define('TESTPATH', realpath(__DIR__.'/../').'/');
--- a/tests/system/Test/BootstrapFCPATHTest.php
+++ b/tests/system/Test/BootstrapFCPATHTest.php
@@ -1,0 +1,91 @@
+<?php namespace CodeIgniter\Test;
+
+/**
+ * Class BootstrapFCPATHTest
+ *
+ * This test confirms that the tests/_support/_bootstrap.php
+ * will set the correct FCPATH regardless of the current directory
+ *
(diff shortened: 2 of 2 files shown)


jim-parry:
Is this a replacement for #1310 ?

chris407x:
Hi Lonnie,

As far as I can figure it has to be set here: we are running the application without loading public/index.php. We looked into this in request #1310:

 

> As long as we're making changes here, we should honor the public folder's name that's found in application/Config/Paths.php, $publicDirectory.
> 
>   @chris407x
> chris407x 22 days ago
> TL:DR:
> 
> loading the application/Config/Paths.php in this case is not trivial because it has a dependency on FCPATH already being set.
> I would recommend leaving it the string 'public'. That way if anyone does change it in Config/Paths they will get "file not found" in all their tests.
> 'public' was a string in the first place, the issues below are probably why
> Details:
> 
> I looked into this and in order to load the Config\Paths() the public/index.php has to do this:
> 
> define('FCPATH', __DIR__.DIRECTORY_SEPARATOR); $pathsPath = FCPATH . '../application/Config/Paths.php'; require $pathsPath; $paths = new Config\Paths() // then $paths->publicDirectory when you want to see 'public'
> 
> in the test _bootstrap.php we are trying to create FCPATH.
> 
> Unfortunately, without FCPATH created no tests will pass (tried this too).
> 
>   @lonnieezell
> lonnieezell 21 days ago  Collaborator
> That's fair. Ignore me on this one. :)

lonnieezell:
That's fair. Ignore me on this one. :)

PR #1409, opened 2018-11-04
propose form helper function usage in sample view form and documentation

samsonasik:
There is sample form view at `application/Views/form.php`, I think it is better to show of the usage of 'form' helper. Also, in documentation, we already open the form with `form_open()`, and rest of code sample was using normal html, the form helper usage is for consistency for its usage to the rest.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated

### DIFF
--- a/application/Views/form.php
+++ b/application/Views/form.php
@@ -1,9 +1,12 @@
 <br><br><br>
 
+<?php
+\Config\Services::validation()->listErrors();
 
-<form method="post" action="" enctype="multipart/form-data">
+helper('form');
+echo form_open('form/process',['enctype' => 'multipart/form-data']);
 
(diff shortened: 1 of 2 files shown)


jim-parry:
application/Views/form.php looks like cruft accidentally introduced as part of a non-PR commit in Sept 2016. I don't think it belongs at all. @lonnieezell ?

lonnieezell:
Yes, that looks like it should go away. Also - I'd rather not use any more of the form library in that example than we already do. It just adds another thing to learn for someone already trying to get to grips with the framework. If anything, I'd rather see the `form_open` tag removed and a normal `<form>` tag in its place, followed by a `csrf_field()` called.

PR #1410, opened 2018-11-05
Fix spark serve not working from commit  2d0b325

samsonasik:
In previous latest commit https://github.com/bcit-ci/CodeIgniter4/commit/2d0b32556198aefec7057d99fa3af8daed9a504b , the `HTTPVerb` property of `RouteCollection` changed by call `Services::request()->getMethod()` by @lonnieezell that make command:

```bash
php spark serve
```

Got output:

```bash
ERROR: 404
Controller or its method is not found: App\Controllers\Ciserve::index
```

**Checklist:**
- [x] Securely signed commits

ping @lonnieezell

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -229,7 +229,7 @@ class RouteCollection implements RouteCollectionInterface
 	public function __construct(FileLocator $locator, $moduleConfig)
 	{
 		// Get HTTP verb
-		$this->HTTPVerb = Services::request()->getMethod() ?? 'cli';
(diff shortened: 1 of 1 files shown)


lonnieezell:
While there are definitely some problems with my latest commit that I'll fix tonight, the solution isn't just to revert. There were probably reasons why I changed that. :)

I'll revert for now since I know there's at least one other issue I didn't notice until after. But expect it to change back to something similar tonight. I hadn't thought of it not working from here, though, so I'll keep an eye out for that.

PR #1416, opened 2018-11-06
remove $psr4['Tests\Support'] definition in application\Config\Autoload::__construct()

samsonasik:
Actually, the `Config\Autoload` class in `application\Config` directory is extends the `CodeIgniter\Config\AutoloadConfig` which in the `__construct()` already set `$this->psr4['Tests\Support']`  and it called the:

```php
parent::__construct();
```

early, so it is no longer needed to be redefined as it will be merged with:

```php
$this->psr4     = array_merge($this->psr4, $psr4);
```

which merge parent `psr4` property to its `$psr4` variable. So, I removed its definition in `Config\Autoload::__construct()`.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/application/Config/Autoload.php
+++ b/application/Config/Autoload.php
@@ -55,11 +55,6 @@ class Autoload extends \CodeIgniter\Config\AutoloadConfig
 			'App'         => APPPATH,                // To ensure filters, etc still found,
 		];
 
-		if (defined('ENVIRONMENT') && ENVIRONMENT === 'testing')
-		{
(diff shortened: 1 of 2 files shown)


lonnieezell:
Good catch. Looks like we need to update AutoloadConfig to respect the TESTPATH constant. 

Could you add that this PR?

samsonasik:
there is `SUPPORTPATH` constant, I updated to use `SUPPORTPATH` in AutoloadConfig class.

lonnieezell:
Forgot about that one! Thanks.

PR #1418, opened 2018-11-06
normalize composer.json

samsonasik:
Tools used: https://github.com/localheinz/composer-normalize

**Checklist:**
- [x] Securely signed commits


jim-parry:
I wouldn't worry too much about this - composer.json will be changing with the repo reorg.
That would be a better time to jump in :)

jim-parry:
Going along with this. It makes sense on its own :)

PR #1426, opened 2018-11-07
performance improvement in Database\BaseResult to use truthy check instead of count($var) when possible

samsonasik:
Applied to `resultObject` and `resultArray`:

Before:

```php
if (($c = count($this->resultObject)) > 0)
{
    for ($i = 0; $i < $c; $i ++ )
    {
    }
}
```

After:

```php
if ($this->resultObject)
{
    foreach ($this->resultObject as $row)
    {
    }
}
```

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseResult.php
+++ b/system/Database/BaseResult.php
@@ -222,7 +222,7 @@ abstract class BaseResult implements ResultInterface
 		{
 			foreach ($this->resultObject as $row)
 			{
-				$this->resultArray[$i] = (array) $row;
+				$this->resultArray[] = (array) $row;
 			}
 
 			return $this->resultArray;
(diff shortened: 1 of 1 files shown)


system/Database/BaseResult.php
```diff
@@ -213,16 +213,16 @@ public function getResultArray(): array
...
 		}
 
-		if (($c = count($this->resultObject)) > 0)
+		if ($this->resultObject)
 		{
-			for ($i = 0; $i < $c; $i ++ )
+			foreach ($this->resultObject as $row)
 			{
-				$this->resultArray[$i] = (array) $this->resultObject[$i];
+				$this->resultArray[$i] = (array) $row;
```
puschie286: you forgot $i here

puschie286: i mean there should be no $i in this function

puschie286: dont know why the Travis build success but you forgot to remove the $i on the left hand array in both foreach loops ( l:225, l:268 )

samsonasik: Ooops, fixed now. Thank you @puschie286

PR #1428, opened 2018-11-07
Fixes #1393: add missing sendCookies() after sendHeaders() call on Response::redirect()

### LINKED ISSUE #1393: RedirectResponse does not set cookies
**Describe the bug**

```php
```

It does not works:

```php
```

**CodeIgniter 4 version**

4.0.0-alpha.2

**Affected module(s)**

HTTP

Expected cookies to be set.

**Context**
 - OS: [Debian]
 - Web server [PHP]
 - PHP version [7.2]

samsonasik:
Fixes #1393 . The code of : 

```php
$this->response->redirect() ...
```

is calling `Response::redirect()` which doesn't call `sendCookies()` after `sendHeaders()`, so I added to be as follow:

```php
	public function redirect(string $uri, string $method = 'auto', int $code = null)
	{
                // ...
		$this->sendHeaders();
		$this->sendCookies();
                // ...
         }
```

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/Response.php
+++ b/system/HTTP/Response.php
@@ -27,21 +27,21 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  *
- * @package      CodeIgniter
- * @author       CodeIgniter Dev Team
- * @copyright    2014-2018 British Columbia Institute of Technology (https://bcit.ca/)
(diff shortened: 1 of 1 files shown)


samsonasik:
I think I am missing something, as the `setCookie()` called after `redirect()` which may need to `sendCookies()` after `setCookie()`, closing for now.

PR #1432, opened 2018-11-08
Use mb_strlen to get length of columns

natanfelles:
**Description**

`strlen()` does not consider certain characters. This affected the calculation of spaces between the columns of the `CLI::table()`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


lonnieezell:
Good catch. Thanks!

PR #1440, opened 2018-11-09
Access to model's last inserted ID

nowackipawel:
**Description**
Allows to get insertID after insert($insert_data, false) which returns CodeIgniter\Database\*\Result object instead of insertID.

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -86,7 +86,14 @@ class Model
 	 * @var string
 	 */
 	protected $primaryKey = 'id';
-
+	
+	/**
+	 * Last insert ID
+	 *
+	 * @var int
+	 */
+	protected $insertID = 0;
+	
 	/**
 	 * The Database connection group that
 	 * should be instantiated.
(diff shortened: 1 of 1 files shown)


jim-parry:
There is an insertID() provided in Database\BaseConnection. Can it be used instead of a new method, with a different accessor style (getInsertID vs insertID)?

nowackipawel:
Ofc there is insertID() method in BaseConnection but BaseConnection is not accessible from outside of the model  after insert is called with false as the 2nd parameter.

jim-parry:
Fair enough. It just struck me as odd, but makes sense now.

system/Model.php
```diff
@@ -551,6 +570,8 @@ public static function classToArray($data, string $dateFormat = 'datetime'): arr
 	public function insert($data = null, bool $returnID = true)
```
atefBB, line 570: I think this feature isn't `well` documented! Need to be documented.

PR #1449, opened 2018-11-10
Simplify how to get indexData from mysql/mariadb

natanfelles:
**Description**

Allow to get all mysql/mariadb index types (PRIMARY, INDEX, UNIQUE, FULLTEXT and SPATIAL) without regex. Currently was possible to get PRIMARY, INDEX, UNIQUE types.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


system/Database/MySQLi/Connection.php
```diff
@@ -444,71 +446,76 @@ public function _fieldData(string $table): array
...
 				}
+
+				$keys[$index['Key_name']]->type = $type;
 			}
+
+			$keys[$index['Key_name']]->fields[] = $index['Column_name'];
 		}
 
-		return $retval;
+		return $keys;
```
natanfelles: The [tests](https://travis-ci.org/natanfelles/CodeIgniter4/jobs/453165355#L611) are failing due to return the $keys indexes with the key name. The test [expect to be keys as integers](https://github.com/bcit-ci/CodeIgniter4/blob/53f17e8ff080b1490cd7df3c57ec2924c858e21b/tests/system/Database/Live/ForgeTest.php#L186).

I think about return the indexes with the key names by default, then update the tests - what do you think? Thanks.

natanfelles: With key names will not have issues with ordering, like:

https://github.com/bcit-ci/CodeIgniter4/blob/53f17e8ff080b1490cd7df3c57ec2924c858e21b/tests/system/Database/Live/ForgeTest.php#L184-L207

lonnieezell: That should be fine.

PR #1453, opened 2018-11-10
BugFix  if there extension has only one mime type (string)

nowackipawel:
in_array() expects parameter 2 to be array, string given

APPPATH/Config/Mimes.php at line 316


jim-parry:
The resultig if statement is a bit of a mouthful, but valid :-/

nowackipawel:
You are correct  I was wondering if I should change all extension  values (mimes array) from strings to arrays... but this option was faster and there is literally no change in performance :)

PR #1457, opened 2018-11-11
WIP use $paths->systemDirectory in public/index.php 

samsonasik:
So it reduce manual change when changing `system` directory.

**Checklist:**
- [x] Securely signed commits


jim-parry:
This will work for now, but the solution will be different once we get the the simplified distributions in place.
I would like to defer this until then, and hopefully enlist your help making sure the distrubtions work as intended.

samsonasik:
@jim-parry mergeable?

PR #1459, opened 2018-11-11
Add Vagrantfile

### LINKED ISSUE #1452: WIP Vagrant and/or Docker support
**Direction**

Maybe a Vagrantfile or a Dockerfile, or both, in the development repo be a good idea.

**Describe the bug**

Difficulty to do local tests before to push online.

natanfelles:
**Description**

Closes #1452 

Work in progress. Idea is to be able to just clone the repository and simply run a `vagrant up`.

Adding two vhost - one for the framework, another for PHPUnit Code Coverage HTML.

I did not have time with docker yet - I think is not possible adjust everything in a single file with docker...

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


jim-parry:
If I read the vagrantfile correctly, the two virtual hosts are on the VM itself, right.
And the intent is to take the current folder and sync it to those inside the VM.
Looks good, but I think we will need a bit of a writeup in the user guide, inclugin setting up VB & Vagrant, and getting the appropriate vagrantbox. It feels like a new page at user_guide_src/testing/vagrant.rst would be the most appropriate, with user_guide_src/indexrst updated as well. If you want to take a stab at that, that would be good. If not, leave it for me, as I plan a few more explanations for installation/setup/testing.

natanfelles:
Yes. 

http://localhost:8080 for the framework.
http://localhost:8081 for local code coverage.

Tested MySQL and Postgres with Debian Testing all OK. But in Debian it is more complicated to configure the PHP version. Testing is coming with 7.3 and I noticed some bugs (do not recognize constant values as default values as function params).

Tuning to work with Ubuntu.

Instructions in the manual will be very good. Sorry, my English is not good.

jim-parry:
@lonnieezell I seem to recall you mentioning that you use vagrant. If that is the case, does this seem like a suitably generic approach? Anything to add?

lonnieezell:
I used to use a virtual machine for it but currently just use it in MacOS. The script looks excellent (especially like the 3 different ports!). One thing I would love to see, though, is support for multiple PHP versions, since we sometimes need to debug in different versions to track down issues. 

I use a [great little script, php-version](https://github.com/wilmoore/php-version) which makes that painless. It would be awesome if this was supported out of the box. 

My only other concern is if this might conflict with using the built-in `php spark serve`. I almost never use a full server anymore for development, though I guess that's needed to support the multiple port mapping.

jim-parry:
@lonnieezell I think it is easier to configure the VM to have apache running as a service, than it would be to have it use the spark server. Not sure how easy it would be to configure the spark server as an O/S service, hmmm.
I think that a developer testing their app would find the spark server easier, but for testing the framework and proposed contributions, the VM would seem more "authentic".
Just a thought - even with a VM fired up with Apache, like this PR does, one can always "vagrant ssh" into the box and run the spark server locally. It would not have a port mapped for the host platform, but would still be usable inside the VM, eg REST API testing or unit testing for the app.

jim-parry:
@natanfelles Why the "bionic" version of ubuntu? "trusty" has the lion's share of vagrantbnox downloads. I don't know if the platform version has (or should have) any effect on the testing/unit testing.

natanfelles:
@jim-parry Because it is the most current LTS version: https://www.ubuntu.com/about/release-cycle
And it comes with PHP 7.2 by default.

@lonnieezell I have not had the opportunity to use php-version yet. It would be nice to be able to test in several versions of PHP, in the VM.

In the morning I was studying [phpcov](https://github.com/sebastianbergmann/phpcov) to merge the various tests and show them in the same coverage-html of the VM. To the local coverage be more similar to coverage/coveralls.

The Vagrantfile is working. The VM occupies ~600MB RAM when running PHPUnit. And ~180MB of downloaded packages (~700MB after installation).

Ah, @jim-parry , also with Sphinx ready to be used. I think this will help in editing the user guide.

jim-parry:
@natanfelles Good call, then :)

natanfelles:
**.env** for tests

```ini
app.baseURL = 'http://example.com'
database.tests.hostname = localhost
database.tests.database = codeigniter
database.tests.username = root
database.tests.password = password
database.tests.DBDriver = Postgre # MySQLi, Postgre, SQLite3
database.tests.port = 5432 # MySQLi = 3306 # Postgre = 5432
```

Possible gaps:

#### Apache is down after `vagrant up`

Motive: Unknown
Solution: Login with `vagrant ssh` and run `sudo systemctl restart apache2`

#### PHPUnit error:

```
There was 1 error:

1) CodeIgniter\CodeIgniterTest::testRunDefaultRoute
ErrorException: Undefined variable: value
```

Motive: Unknown - probably is related with vagrant symlinks issues
Solution: Run phpunit directly by `./vendor/phpunit/phpunit/phpunit`

Vagrantfile
```diff
@@ -0,0 +1,161 @@
...
+  # Automatic box update checking
+  config.vm.box_check_update = true
+
+  # CodeIgniter virtual host
+  config.vm.network "forwarded_port", guest: 80, host: 8080, host_ip: "127.0.0.1"
+  # Code Coverage virtual host
+  config.vm.network "forwarded_port", guest: 81, host: 8081, host_ip: "127.0.0.1"
+  # User Guide virtual host
+  config.vm.network "forwarded_port", guest: 82, host: 8082, host_ip: "127.0.0.1"
+  # MySQL server
```
natanfelles: Services can be accessed externally:

Vagrantfile
```diff
@@ -0,0 +1,161 @@
...
+  # but chmod does not work... tests will fail.
+  # Default rsync__args except "--copy-links", to allow phpunit correctly works by symlink
+  config.vm.synced_folder ".", "/var/www/codeigniter", type: "rsync", rsync__args: ["--verbose", "--archive", "--delete", "-z"]
+
+  # Provider-specific configuration
+  config.vm.provider "virtualbox" do |vb|
+    # Display the VirtualBox GUI when booting the machine
+    vb.gui = false
+    # Customize the amount of memory on the VM:
+    vb.memory = "768"
```
natanfelles: VM runs with 400MB - 308MB "stopped". Required ~600MB when running tests

lonnieezell:
@natanfelles I really like what you're building here. Thanks!

jim-parry:
I prefer the virtual host approach myself - more realistic IMHO.
I  presume that 192.168.10.10 is inside the VM? or are you tinking that the alias be defined in the host (developer) platform. If the latter, then not everyone would have 192.168.x.x

natanfelles:
Yes. 192.168.10.10 is the VM IP.

The developer can access directly or by a custom domain (codeignter.test by default, maybe)

natanfelles:
captura de tela de 2018-11-20 23-12-39

natanfelles:
or `example.com` because it is the domain used in the tests

Vagrantfile
jim-parry: I run mine with 1024m. Probably overkill, but hoping/planning for near-complete code coverage eventually.
Is there much of a penalty specifying more memory?

natanfelles: It depends on the user's machine. Thinking about taking up as little resources as possible, I started with 512MB, but PHPUnit broke for not having enough memory to do the coverage. I run again and saw that it occupies about 600MB at most (MySQLi driver). I increased 256MB. 

Testing now with the SQLite3 driver, PHPUnit also could not gererate the code coverage:

```
Type:        ErrorException
Message:     proc_open(): fork failed - Cannot allocate memory
Filename:    /var/www/codeigniter/vendor/sebastian/version/src/Version.php
Line Number: 89
```

I'll go up to 1024 and test again. But this seems to me a very high value, I'm thinking of maybe already letting in the script instructions to create a swap file, because the memory only rises so much when it's generating code coverage.

natanfelles: SQLite3 with `database.tests.database = ':memory:'`...

natanfelles: OK. Increased the default `vb.memory` to 1024 and added script to auto add 1GB of swap if the VM has low RAM.

Tested using SQLite3 with `vb.memory = "512"` and no problem. It used ~120MB of swap during coverage.

jim-parry:
"Vagrantfile" & ".vagrant" are currently in the .gitignore file.
We should rename "Vagrantfile" in this PR to "Vagrantfile.dist", so it doesn't clobber an existing setup.
If you can do that, I will merge this and start on the docs for it :)

PR #1460, opened 2018-11-11
remove unneeded ternary check at HoneyPot::hasContent()

samsonasik:
`! empty()` already return true on not empty, false on empty.

**Checklist:**
- [x] Securely signed commits


samsonasik:
@jim-parry fixed

PR #1462, opened 2018-11-12
using "if" instead of "while" for file_exists() check in File::getDestination()

samsonasik:
**Checklist:**
- [x] Securely signed commits


samsonasik:
or I probably missing something, closing for now

PR #976, opened 2018-03-31
Allow Forge create table ENUM and SET with array constraint

natanfelles:
See: https://forum.codeigniter.com/thread-61466.html

PR #1470, opened 2018-11-12
Allow create table with array field constraints

natanfelles:
**Description**

See #976

Testing only ENUM in MySQLi - but all constraints set as array will be escaped and imploded, works with MySQL SET Data Type too.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -82,7 +82,7 @@ class Forge
 	/**
 	 * List of foreign keys.
 	 *
-	 * @var type
+	 * @var array
 	 */
 	protected $foreignKeys = [];
 
@@ -340,7 +340,7 @@ class Forge
 	/**
 	 * Add Field
 	 *
-	 * @param array $field
+	 * @param array|string $field
 	 *
 	 * @return Forge
 	 */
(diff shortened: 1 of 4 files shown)


natanfelles:
Perhaps. It is intended to work with any driver. 

I am still not aware of Data Types in other DBMS besides MySQL, PostgreSQL and SQLite.

I found only the MySQL ENUM and SET Data Types allowing a list of values so far. So testing only once.

lonnieezell:
Well, Postgres definitely [has Enum](https://www.postgresql.org/docs/9.1/datatype-enum.html). 

sqlite3 has [checks](https://stackoverflow.com/questions/5299267/how-to-create-enum-type-in-sqlite), though it's not quite the same thing.

At the very least the Postgres test should happen. Not too concerned about the sqlite one, though that would be a bonus.

natanfelles:
But these are the queries:

```sql
// crate the type "mood" as ENUM
CREATE TYPE mood AS ENUM ('sad', 'ok', 'happy');
// create table using the "mood" - ENUM is not used inside CRATE TABLE query
CREATE TABLE person (
    name text,
    current_mood mood
);
```
---

SQLite CHECK seems to me to be a hack to emulate MySQL ENUM. 

Thinking about doing something to not escape constraints like this, if the driver is SQLite:

```php
$db_columns = array(
    'status' => array(
        'type'       => 'ENUM',
        'constraint' => 'CHECK( status IN ('locked','unlocked') )',
        'default'    => 'unlocked',
        'comment'    => 'Status of asset (Locked or Unlocked)'
    )
); 
```

natanfelles:
Or, convert the type ENUM to TEXT and the CHECK in the constraint... It allow the same code works in mysql and sqlite

natanfelles:
The postgres "created type" can be used in other  tables/fields too.

lonnieezell:
Oh, good point about Postgres. 

And you're right that Check is a hack. I only mentioned it because it looked like it would have been implemented with an array.

natanfelles:
Can be. But the user also can:

```php
$db_columns = array(
    'status' => array(
        'type'       => 'TEXT CHECK( status IN ("locked","unlocked") )',
    )
); 
```

Or we allow CodeIgniter make the conversion:

```php
// sqlite: status TEXT CHECK( status IN ('unlocked','locked') )
$db_columns = array(
    'status' => array(
        'type'       => 'ENUM',
        'constraint' => ['unlocked', 'locked'],
    )
); 
```
 The same PHP code would works with mysql and sqlite

lonnieezell:
for this, I like letting CI convert it. The most common use case I see for SQLite on the web is running tests, even though the web apps are actually using a different engine, like Mysql or Postgres in production. Converting it on the fly would be one less pain when trying to test your app. And, as you said, they can use check for other things as a text item. 

Would you mind doing that conversion?

lonnieezell:
> Also being able to "create the type" for postgres on the fly. But I fear that this will lead to collisions.

I agree - and since creating the type is done outside of the create table I wouldn't worry about that one. Just sqlite.

PR #1480, opened 2018-11-13
Fix Language Key-File confusion

puschie286:
The Language class currently crash if one key name is the filename of another language file

**Description**
removed the key check on file level

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

*Im not sure why is check was implemented*

### DIFF
--- a/system/Language/Language.php
+++ b/system/Language/Language.php
@@ -152,14 +152,14 @@ class Language
 		$file = substr($line, 0, strpos($line, '.'));
 		$line = substr($line, strlen($file) + 1);
 
-		if (! array_key_exists($line, $this->language))
+		if (! isset( $this->language[$this->locale][$file] ) )
 		{
 			$this->load($file, $this->locale);
 		}
 
 		return [
 			$file,
-			$this->language[$this->locale][$line] ?? $line,
+			$line,
 		];
 	}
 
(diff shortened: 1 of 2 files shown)


system/Language/Language.php
```diff
@@ -152,14 +152,14 @@ protected function parseLine(string $line): array
 		$file = substr($line, 0, strpos($line, '.'));
 		$line = substr($line, strlen($file) + 1);
 
-		if (! array_key_exists($line, $this->language))
+		if (! isset( $this->language[$this->locale][$file] ) )
```
lonnieezell: This needs to stay `array_key_exists` to allow for the possibility of an empty string being the value, since that is technically valid.

puschie286: oh - yes didn't think of this possibility

system/Language/Language.php
```diff
@@ -152,7 +152,7 @@ protected function parseLine(string $line): array
 		$file = substr($line, 0, strpos($line, '.'));
 		$line = substr($line, strlen($file) + 1);
 
-		if (! array_key_exists( $line, $this->language[$this->locale][$file] ) )
+		if (! isset($this->language[$this->locale][$file]) || ! array_key_exists($line, $this->language[$this->locale][$file]))
```
lonnieezell: Why are you checking both ways? All that should be needed is array_key_exists. It's the same as isset, except it works even if the value is empty.

puschie286: e.g "SomeName.en" would array_key_exists( $line, $this->language ) result in true if there is any english file loaded

lonnieezell: Oh, yes, totally mis-read that Sorry about that!

PR #1485, opened 2018-11-15
using truthy check instead of count($data) at Model::doProtectFields()

samsonasik:
For performance improvement.

**Checklist:**
- [x] Securely signed commits


lonnieezell:
While the truthy check might be marginally faster, the difference is negligible and I feel that, in this case, `count()` is clearer. Passing on this one, but thanks.

PR #1486, opened 2018-11-15
Test RedirectResponse problem report

### LINKED ISSUE #1393: RedirectResponse does not set cookies

jim-parry:
Added tests for response->redirect() and redirect(), and both seem ok.
Closes #1393

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -815,7 +815,7 @@ if (! function_exists('redirect'))
 
 		if (! empty($uri))
 		{
-			return $response->to($uri);
+			return $response->route($uri);
 		}
 
 		return $response;
--- a/system/HTTP/Response.php
+++ b/system/HTTP/Response.php
@@ -756,8 +756,6 @@ class Response extends Message implements ResponseInterface
 
 		$this->setStatusCode($code);
 
-		$this->sendHeaders();
-
 		return $this;
 	}
 
--- a/tests/system/CommonFunctionsSendTest.php
+++ b/tests/system/CommonFunctionsSendTest.php
@@ -1,0 +1,54 @@
+<?php
+
+use CodeIgniter\HTTP\Response;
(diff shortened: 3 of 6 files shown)


jim-parry:
@natanfelles Made an adjustment to Common::redirect() that looks like it has fixed the problem.

natanfelles:
Not solved. But I see a light:

https://github.com/codeigniter4/CodeIgniter4/blob/fb60b92f8bacbc2e28c148c4faafe28091147ac4/system/HTTP/Response.php#L759

Headers are sent before the fluent method `->setCookie()` is called

natanfelles:
And is called again in `send()`:

https://github.com/codeigniter4/CodeIgniter4/blob/fb60b92f8bacbc2e28c148c4faafe28091147ac4/system/HTTP/Response.php#L633-L651

Maybe because of this works with the Response but not with RedirectResponse.

Firefox Raw Headers repeat some headers: https://github.com/codeigniter4/CodeIgniter4/issues/1393#issuecomment-439128266 - because of the headers sent in `redirect()` and `send()`, again. 

Something tells me this is there in bootstrap.php

jim-parry:
sendHeader() in redirect() looks out of place, I agree.
I eliminated it, and that does not affect the unit tests - they still pass.
This could very well be something not-testable with unit testing ... headers are a pain at the best of times with phpunit.
In any case, I pushed the change.

jim-parry:
The travis-ci failure is from Redis tests, not related. This is good to test.

PR #1487, opened 2018-11-15
WIP Reviewing some sessions classes

fmertins:
Hi there, I've done some review and minor-simple improvements, as follows below, thanks.

1) BaseHandler, added some simple phpDoc.
2) BaseHandler, improvements at the constructor and added a TODO.
3) DatabaseHandler, just a phpDoc.
4) FileHandler, constructor improvements and indentation correction.
5) Session, added TODO about the return's start() method.

### DIFF
--- a/system/Session/Handlers/BaseHandler.php
+++ b/system/Session/Handlers/BaseHandler.php
@@ -50,7 +50,7 @@ abstract class BaseHandler implements \SessionHandlerInterface
 	/**
 	 * The Data fingerprint.
 	 *
-	 * @var boolean
+	 * @var bool
 	 */
 	protected $fingerprint;
 
@@ -64,49 +64,49 @@ abstract class BaseHandler implements \SessionHandlerInterface
 	/**
 	 * Cookie prefix
 	 *
-	 * @var type
+	 * @var string
 	 */
 	protected $cookiePrefix = '';
 
 	/**
 	 * Cookie domain
 	 *
-	 * @var type
+	 * @var string
 	 */
 	protected $cookieDomain = '';
 
 	/**
 	 * Cookie path
 	 *
-	 * @var type
+	 * @var string
 	 */
 	protected $cookiePath = '/';
 
 	/**
 	 * Cookie secure?
 	 *
-	 * @var type
+	 * @var bool
 	 */
 	protected $cookieSecure = false;
 
 	/**
 	 * Cookie name to use
 	 *
-	 * @var type
+	 * @var string|null
 	 */
 	protected $cookieName;
 
 	/**
 	 * Match IP addresses for cookies?
 	 *
-	 * @var type
+	 * @var bool
 	 */
 	protected $matchIP = false;
 
 	/**
 	 * Current session ID
 	 *
-	 * @var type
+	 * @var string|null
 	 */
 	protected $sessionID;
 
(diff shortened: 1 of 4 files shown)


system/Session/Handlers/BaseHandler.php
```diff
@@ -50,7 +50,7 @@
 	/**
 	 * The Data fingerprint.
 	 *
-	 * @var boolean
+	 * @var bool
```
jim-parry: Are you using the phpcbf pre-commit hook? It changes bool to boolean

system/Session/Session.php
```diff
@@ -184,14 +184,16 @@ public function __construct(\SessionHandlerInterface $driver, $config)
...
+	 * @return mixed
 	 */
 	public function start()
 	{
 		if (is_cli() && ENVIRONMENT !== 'testing')
 		{
 			$this->logger->debug('Session: Initialization under CLI aborted.');
 
-			return;
+			return; // TODO fmertins: should we throw an exception instead return? And then @return Session phpdoc?
```
jim-parry: "return" on its own is on purpose. Throwing an exception would break trying to unit test sessions.
The return returns "void", but more importantly short-circuits any additional logic. So, no it should not be tagged as returning a Session.

system/Session/Handlers/FileHandler.php
```diff
@@ -79,8 +79,9 @@ class FileHandler extends BaseHandler implements \SessionHandlerInterface
 	 * Constructor
 	 *
 	 * @param BaseConfig $config
+	 * @param string $ipAddress
 	 */
-	public function __construct($config, string $ipAddress)
+	public function __construct(BaseConfig $config, string $ipAddress)
```
jim-parry: Same comment as above.

system/Session/Handlers/BaseHandler.php
```diff
@@ -129,9 +129,11 @@
 	 * Constructor
 	 *
 	 * @param BaseConfig $config
+	 * @param string $ipAddress
 	 */
-	public function __construct($config, string $ipAddress)
+	public function __construct(BaseConfig $config, string $ipAddress)
```
jim-parry: The vagueness is on purpose. an App() is actually passed in, with the properties that are not in BaseConfig.
It just needs to be a class with the "proper" properties.
It breaks your build big-time, as you notive from travis-ci.

system/Session/Handlers/BaseHandler.php
fmertins: Hi, no I'm not... Should I? In other words, I'm trying to use bool as a primitive type instead of boolean as eventually a class/interface name. Is this correct? I'm talking about this:
http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

system/Session/Session.php
fmertins: Hum, ok I believe I got. My idea was avoiding return mixed, and because I saw those logs with error or warning I was thinking if it would be better generating an exception right after the log. OK, I'll  change it, tks :-)

system/Session/Handlers/BaseHandler.php
jim-parry: The phpcbf pre-commit hook will "fix" style things. Specifically, if you change "boolean" to "bool", it should change that back to "boolean" before letting you proceed. The codeigniter4/standards repo rule set needs to be fixed to resolve this.
And yes, you should be using the pre-commit hook. That is in our composer.json, and will be installed automatically if you do a "composer update" in your fork's clone.

jim-parry: ps - your reasoning is good, but phpcbf will automatically "undo" your changes.

system/Session/Session.php
fmertins: So, maybe a phpdoc `@return Session|void`? Because we shouldn't return anything different here... :-)

jim-parry: That works

jim-parry: And the todo comment should come off :-/

atishamte:
is this PR really required?

jim-parry:
@atishamte Agreed. Stale, & most changes effected already elsewhere.

PR #1488, opened 2018-11-15
Load Language strings from other locations

### LINKED ISSUE #1262: TODO Language needs improved locating
Language/Language::requireFile(...) has a todo note:
should look into loading from other locations, also probably..

### LINKED ISSUE #1433: Language is not merging with modules
**Describe the bug**

When a Language file with same *name* of an default file of the *system/Language/en/* is in a module folder, it is not merged.

For example:

I have the following folder structure:

```
application/
module-x/
module-x/Commands/
module-x/Language/en/
```

The namespace is proper configured in the Autoload config. I can see the commands when I call the `spark`.

And the `lang()` function also works - BUT, if a module lang file has the same name of a default file, it does not works.

If a file is named *Foo.php*, it works: `lang('Foo.hello')`.

If a file is named *HTTP.php*, it does not works: `lang('HTTP.hello')`.

Same for others, like *Database.php*, etc.

**CodeIgniter 4 version**

4.0.0-alpha.2

**Affected module(s)**

Language

Expected that module Language files be merged.  

**Context**
 - OS: [Debian]
 - Web server [PHP]
 - PHP version [7.2]

natanfelles:
**Description**

Fixes #1433

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Language/Language.php
+++ b/system/Language/Language.php
@@ -249,8 +249,6 @@ class Language
 	 * A simple method for including files that can be
 	 * overridden during testing.
 	 *
-	 * @todo - should look into loading from other locations, also probably...
-	 *
 	 * @param string $path
 	 *
 	 * @return array
@@ -258,6 +256,8 @@ class Language
 	protected function requireFile(string $path): array
 	{
 		$files = service('locator')->search($path);
+
+		$strings = [];
 
 		foreach ($files as $file)
 		{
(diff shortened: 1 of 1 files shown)


system/Language/Language.php
```diff
@@ -270,10 +270,15 @@ protected function requireFile(string $path): array
 			// on this command returning boolean instead
 			// of array during testing, so we've removed
 			// the require_once for now.
-			return require $file;
```
natanfelles: No performance issues here. $files comes with only the matching files.

natanfelles:
The fallback was not added...

PR #1493, opened 2018-11-17
apply declare(strict_types=1);

samsonasik:
**Checklist:**
- [x] Securely signed commits


jim-parry:
This suggestion has come up before, and the consensus, AFAIK, was that we *want* the flexibility of non-strict types.

samsonasik:
how about apply that in `system` and `tests` only? For `application` php files, it will up to user to use strict or not

jim-parry:
It is specifically in system/ that we want flexible typing.
I see flexible typing as a way of voercoming PHP's lack of method overloading, found in other languages.
When the suggestion came up before, Lonnie had a more elegant answer, so I am hoping he can chime in to clarify :)

lonnieezell:
I don't know that it's a more elegant answer, but here goes:

Like it or not, at its core, PHP **is** a flexibly typed language. It is not a typed language like Java or Hack, and that should be embraced. While I understand why people like the idea of it, and why it's helpful, I don't think it's appropriate for CodeIgniter. Type casting is actually a handy thing at times, and the other newer features that came with strict typing, like more argument and return value type hints go a long way toward fixing the issues PHP had by failing loudly for the developer to see. But forcing them to manually cast every number retrieved from an HTML form into an integer, and the like, is an unnecessary burden, I feel. 

In short (too late!) embrace PHP for what it is, and don't try to force it to be a language that it isn't. 

Thanks for the submittal, and for all of the other stuff you've helped out with, but I think we'll pass on this one.

PR #1494, opened 2018-11-17
Method naming

gelviis:
In the class reference of the Response Object in the user guide, statusCode() and reason() are listed. However, it should be getStatusCode() and getReason(), as statusCode() and reason() are not methods in HTTP\Response.


jim-parry:
I will merge this, but expect future contributions to be GPG-signed.

PR #1497, opened 2018-11-17
Remove unused RedirectException and add some PHPDocs

natanfelles:
**Description**

RedirectException never is thrown. Added some possible throwns.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/Response.php
+++ b/system/HTTP/Response.php
@@ -42,14 +42,6 @@ use Config\Format;
 use CodeIgniter\HTTP\Exceptions\HTTPException;
 
 /**
- * Redirect exception
- */
-class RedirectException extends \Exception
-{
-
-}
-
-/**
  * Representation of an outgoing, getServer-side response.
  *
  * Per the HTTP specification, this interface includes properties for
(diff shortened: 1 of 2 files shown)


system/HTTP/Response.php
```diff
@@ -428,13 +420,18 @@ public function setJSON($body)
...
 	{
 		$body = $this->body;
 
 		if ($this->bodyFormat !== 'json')
 		{
+			/**
+			 * @var Format $config
+			 */
 			$config    = config(Format::class);
 			$formatter = $config->getFormatter('application/json');
```
natanfelles: In this file, in the line below, `$formatter->format()` is typed to expect an array.

 https://github.com/codeigniter4/CodeIgniter4/blob/cd8eba1fa72c45c199b5f0d3e91e2fdac91295b6/system/Format/FormatterInterface.php#L49

Removing the type hint in the FormatterInterface allow more flexibility to work with objects, like pass an `\CodeIgniter\Entity` directly.

The conversion could be done inside the method, if necessary:

https://github.com/codeigniter4/CodeIgniter4/blob/cd8eba1fa72c45c199b5f0d3e91e2fdac91295b6/system/Format/XMLFormatter.php#L61

JSONFormatter does not needs.

system/HTTP/Response.php
```diff
@@ -924,12 +931,14 @@ public function getCookie(string $name = null, string $prefix = '')
...
 	 * @param string $prefix
+	 *
+	 * @return $this
 	 */
 	public function deleteCookie($name = '', string $domain = '', string $path = '/', string $prefix = '')
 	{
 		if (empty($name))
 		{
-			return;
+			return $this;
```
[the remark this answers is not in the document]
natanfelles: If the $name is empty. Should not delete all? - removing the default `''`

system/HTTP/Response.php
lonnieezell: Interesting idea, and I'm ok with that. You could easily check if it's an `Entity` and call it's `toArray()` method to get the values.

PR #1498, opened 2018-11-17
Add CURLRequest helper methods

### LINKED ISSUE #1259: TODO CURLRequest needs helpers
HTTP/CURLRequest has a todo note...
Add a few helpers for dealing with JSON, forms, files, etc.

natanfelles:
**Description**

Closes #1259

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -47,8 +47,6 @@ use Config\App;
  * A lightweight HTTP client for sending synchronous HTTP requests
  * via cURL.
  *
- * @todo Add a few helpers for dealing with JSON, forms, files, etc.
- *
  * @package CodeIgniter\HTTP
  */
 class CURLRequest extends Request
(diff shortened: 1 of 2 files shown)


system/HTTP/CURLRequest.php
```diff
@@ -267,6 +265,68 @@ public function put(string $url, array $options = []): ResponseInterface
...
+	/**
+	 * Set JSON data to be sent.
+	 *
+	 * @param mixed $data
+	 *
+	 * @return $this
+	 */
+	public function setJSON($data)
+	{
+		$this->config['json'] = $data;
```
lonnieezell: If we're setting json in the request, shouldn't we also set `Content-Type: application/json`?

natanfelles: We could update this to use the `\Config\Format::getFormatter('application/json')`.

Allow to use a custom JSON Formatter...

jim-parry: I don't think there is any intent to format the JSON data. If anything, it might be stringify'd. The unit testing should make this clearer.

PR #1500, opened 2018-11-18
Ignoring errors suppressed by @

samsonasik:
For example, in application, we have the follwoing code: 

```php
@unlink('file');
```

The error should be ignored if the file not exists as the `unlink()` suppressed with `@`

**Checklist:**
- [x] Securely signed commits


samsonasik:
unit test added.

Issue #1501, opened 2018-11-18
Issues with redirects - had to use exit to make it work and blank page 


daljit3:
---
name: Bug report
about: Help us improve the framework by reporting bugs!

---
**Describe the bug**
There are two issues with the `redirect()` method.
1) I am not able to use `return redirect()->to('/');` so I have to follow it up with an exit to make it work. e.g. 
```
public function hello() {
//this won't work;
return redirect()->to('/');
//this works
redirect()->to('/');
exit;
}
```
2) The other issue is none of the redirects are working now since approximately after yesterday (17 nov 2018) evening. To make sure, I haven't accidently broken it, I just downloaded a new copy of CI4-develop and after setup I could see welcome message to confirm everything working fine. Then I made these modifications below to the Home controller. If I try `mylocalsite/home/daljit` it just shows a blank page.
```
 namespace App\Controllers;

use CodeIgniter\Controller;

class Home extends Controller
{
	public function index()
	{
		return view('welcome_message');
	}

	//--------------------------------------------------------------------
    public function daljit()
    {
        return redirect()->to("/home/hello");
    }

    public function hello()
    {
        echo "Hello!";
    }
}
```
 
**CodeIgniter 4 version**
CodeIgniter 4 develop branch.

jim-parry:
Ok, the oldInput changes have been reverted, and the unit tests show that bug.
They do not show any redirect() problems.
@daljit3 Can you confirm that redirect works again, and that the old() bug is back?

jim-parry:
Hmm - after setting the baseURL & indexPage properties in App, I get a blank page even for .../home,
although the default home apge continues to work. This may not be a redirect problem at all, but something deeper.
I have looked at the commits over the last week, and don't see what would cause this.
daljit3, when did you last pull the develop branch from the repo, before the most recent one that seems to be problematic?

daljit3:
@jim-parry - no sorry redirect still doesn't work and I kind of rely on redirects to get to edit form page to do the testing for old(). Just for clarification, I updated my local copy again just now. 

Also, I seem to have found another issue which wasn't there before but I need to investigate more what this new issue is and why I am getting this exception all of sudden as yesterday it was working. Could it be linked to that redirect issue?

```
CodeIgniter\HTTP\Exceptions\HTTPException 
BASEPATH/HTTP\Exceptions\HTTPException.php at line 106
``

[...]

Issue #1492, opened 2018-11-16
Old() - Seems to have an issue with retrieving array values


daljit3:
---
name: Bug report
about: Help us improve the framework by reporting bugs!

---
**Describe the bug**
My input field is named as "locations[]".  When the form failed validation - I tried to retrieve it as `$locations = old('locations');` Which throws up an error - `strpos() expects parameter 1 to be string, array given - BASEPATH/Common.php at line 786`

The line 786 is in old() method in System/Common.php is 
`
		// If the result was serialized array or string, then unserialize it for use...
		if (strpos($value, 'a:') === 0 || strpos($value, 's:') === 0)
		{
			$value = unserialize($value);
		}
`
Here are the full methods from all the files which I think are related to this issue.
You can see the `$value` is being parsed through `$request->getOldInput($key);` in this old() method. There isn't any where I can see the arrays in POST data being serialised as I looked into WithInput() method too. 

```
	function old(string $key, $default = null, $escape = 'html')
	{
		$request = Services::request();

		$value = $request->getOldInput($key);

		// Return the default value if nothing
		// found in the old input.
		if (is_null($value))
		{
			return $default;
		}

		// If the result was serialized array or string, then unserialize it for use...
		if (strpos($value, 'a:') === 0 || strpos($value, 's:') === 0)
		{
			$value = unserialize($value);
		}

		return $escape === false ? $value : esc($value, $escape);
	}
```
** File system/HTTP/IncomingRequest.php **
```
	public function getOldInput(string $key)
	{
		// If the session hasn't been started, or no
		// data was previously saved, we're done.
		if (empty($_SESSION['_ci_old_input']))
		{
			return;
		}

		// Check for the value in the POST array first.
		if (isset($_SESSION['_ci_old_input']['post'][$key]))
		{
			return $_SESSION['_ci_old_input']['post'][$key];
		}

		// Next check in the GET array.
		if (isset($_SESSION['_ci_old_input']['get'][$key]))
		{
			return $_SESSION['_ci_old_input']['get'][$key];
		}

		helper('array');

		// Check for an array value in POST.
		if (isset($_SESSION['_ci_old_input']['post']))
		{
			$value = dot_array_search($key, $_SESSION['_ci_old_input']['post']);
			if (! is_null($value))
			{
				return $value;
			}
		}

		// Check for an array value in GET.
		if (isset($_SESSION['_ci_old_input']['get']))
		{
[...]

PR #1504, opened 2018-11-18
Revert RedirectResponse changes

jim-parry:
Address issue raised in #1501 #1492

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -783,13 +783,13 @@ if (! function_exists('old'))
 		}
 
 		// If the result was serialized array or string, then unserialize it for use...
-		//      if (is_string($value))
-		//      {
-		if (strpos($value, 'a:') === 0 || strpos($value, 's:') === 0)
+		if (is_string($value))
+		{
+			if (strpos($value, 'a:') === 0 || strpos($value, 's:') === 0)
 			{
-			$value = unserialize($value);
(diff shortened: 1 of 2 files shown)


system/Common.php
```diff
@@ -818,7 +818,7 @@ function redirect(string $uri = null)
 
 		if (! empty($uri))
 		{
-			return $response->route($uri);
```
lonnieezell: This was originally forced to use `route` instead of `to` to make it difficult to redirect to pages outside the site on accident. This was inspired by OWASP's 2013 Top 10 list, [Unvalidated Redirects and Forwards](https://www.owasp.org/index.php/Top_10_2013-A10-Unvalidated_Redirects_and_Forwards). It is no longer in the Top 10 list, but still a decent security precaution, I believe.

jim-parry:
It was originally ->to)_, and I changed it to route(), which made more sense to me.
This got reverted to to() trying to nail down the culprit for the blank pages, but this wasn't the cause.
Are you suggesting this should remain on "route()" when the dust settles?

lonnieezell:
I think it's a good security thing, yes, even though it might be a bit less user-friendly. I'm not 100% convinced it's crucial so if it causes issues with users, i'm fine leaving it at `to` and letting them control their own fate. It's simpler to use that way.

PR #1505, opened 2018-11-19
Do not exit until all Response is completed

### LINKED ISSUE #1393: RedirectResponse does not set cookies

natanfelles:
**Description**

Fixes #1393

CodeIgniter was exiting before send all headers when returning a RedirectResponse.

Now, it acts the same way as when returning a Response redirect.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -326,17 +326,6 @@ class CodeIgniter
 			$this->benchmark->stop('controller');
 		}
 
-		// Handle any redirects
-		if ($returned instanceof RedirectResponse)
-		{
-			if ($returnResponse)
-			{
-				return $returned;
-			}
-
-			$this->callExit(EXIT_SUCCESS);
-		}
-
 		// If $returned is a string, then the controller output something,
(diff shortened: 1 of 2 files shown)


jim-parry:
Good catch.It looks like I caught one of the problems but missed the other.
And you confirm that the fix works, since you filed the original report?

natanfelles:
Yes. Confirmed.

With this the response with `redirect()` helper is similar to `Response::redirect()`.

No headers are repeated.

However removing this exit, the _after filters_, storePreviousURL, and post_system Event will run when returning a RedirectResponse. I have not seen any problem in this, since this is the default process when redirection is done with Response.

lonnieezell:
That looks like that should do it. Good catch.

PR #1506, opened 2018-11-19
Flesh out HTTP/File unit testing

jim-parry:
Included error correction in UploadedFile::guessExtension()


system/HTTP/Files/FileCollection.php
```diff
@@ -92,26 +92,16 @@ public function getFile(string $name)
 			{
 				$name         = explode('.', $name);
 				$uploadedFile = $this->getValueDotNotationSyntax($name, $this->files);
-				if ($uploadedFile instanceof \CodeIgniter\HTTP\Files\UploadedFile)
-				{
-					return $uploadedFile;
-				}
-
-				return null;
+				return ($uploadedFile instanceof \CodeIgniter\HTTP\Files\UploadedFile) ?
```
samsonasik: in same namespace, use direct `instanceof UploadedFile` should be enough:

```suggestion
                              return ($uploadedFile instanceof UploadedFile) ?
```

system/HTTP/Files/FileCollection.php
```diff
@@ -92,26 +92,16 @@ public function getFile(string $name)
...
 			if (array_key_exists($name, $this->files))
 			{
 				$uploadedFile = $this->files[$name];
-				if ($uploadedFile instanceof \CodeIgniter\HTTP\Files\UploadedFile)
-				{
-					return $uploadedFile;
-				}
-
-				return null;
+				return  ($uploadedFile instanceof \CodeIgniter\HTTP\Files\UploadedFile) ?
```
samsonasik: in same namespace, use direct `instanceof UploadedFile` should be enough:

```suggestion
                              return ($uploadedFile instanceof UploadedFile) ?
```

system/HTTP/Files/FileCollection.php
jim-parry: Good suggestion. I don't know why it was fully qualified before.

PR #1507, opened 2018-11-19
Fix multi "empty" string separated by "," marked as valid emails

samsonasik:
In `Validation\FormatRules::valid_emails()`, if we passed a string: 

```php
','
```

It previously checked as valid email, while empty string is not valid email. We need to check if the string separated by comma is an empty string or not. If empty, then return `false` early.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -294,7 +294,13 @@ class FormatRules
 
 		foreach (explode(',', $str) as $email)
 		{
-			if (trim($email) !== '' && $this->valid_email(trim($email)) === false)
+			$email = trim($email);
+			if ($email === '')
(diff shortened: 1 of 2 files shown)


jim-parry:
Testing for ',' makes sense, but the FormatRules change doesn't seem to serve a point. If looking to eliminate a trim(...), then why not just

```
$email = trim($email);
return (empty($email)) ? false : $this->valid_email($email);

```

samsonasik:
It is in the loop, so it doesn't want to return early on non-empty email, so the code sample you gave will return "true" early even the passed email is valid, and next email is not valid.

jim-parry:
I see your point now :)

PR #1510, opened 2018-11-19
handle more fatal errors

samsonasik:
For example, we have the following code:

```php
$class = new class implements \stdClass {}
```

For example, in `Home::index()` function.

Previously, it will get the following output:

before

I updated so it will got the following output:

after

**Checklist:**
- [x] Securely signed commits


jim-parry:
It looks like you are trapping too much output. Did you notice that the code coverage dropped 60%?
I would rather live with a bit of messy debugging output!

samsonasik:
dropping 60% coverage is weird as `ob_start()` should always pass. 

I previously uses `ob_start()` to get buffer error ( https://github.com/samsonasik/ErrorHeroModule/blob/master/src/HeroTrait.php#L25-L41 ) with more conditions and it can be covering by only 3 test examples https://github.com/samsonasik/ErrorHeroModule/blob/master/spec/Middleware/ExpressiveSpec.php#L530-L565 , but with different test framework ( kahlan )

jim-parry:
Check the travis-ci details...
There were 2315 risky tests:
1) CodeIgniter\API\ResponseTraitTest::testNoFormatterJSON

Your ob_start is hijacking the output buffer from phpunit, it looks like, and most of the tests are not running properly.

PR #1511, opened 2018-11-19
apply constants visibility

samsonasik:
Since php 7.1, constant in class can have visibility

**Checklist:**
- [x] Securely signed commits


system/HTTP/URI.php
```diff
@@ -40,20 +40,12 @@
  */
 class URI
 {
-
-	/**
-	 * Sub-delimiters used in query strings and fragments.
-	 *
-	 * @const string
-	 */
-	const CHAR_SUB_DELIMS = '!\$&\'\(\)\*\+,;=';
```
samsonasik: this contant never used (yet), so I removed it.

lonnieezell:
This is redundant. The default visibility of a constant is public. We never use protected or private constants.

PR #1512, opened 2018-11-19
More secure redirection

jim-parry:
Common::redirect() should use ->route() instead of ->to(), at Lonnie's suggestion


jim-parry:
Messed up synching. Will resubmit

PR #1514, opened 2018-11-19
Update Services.php

sugenganthos:
This was will cause wrong file path if using views like 'template.tpl' and makes real path into 'application/Views/template.tpl.php because template.tpl' isn't found and file locator will add '.php' extention.
Giving ROOTPATH makes absolute path to 'template.tpl' and be found.

### DIFF
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -500,7 +500,7 @@ class Services extends BaseService
 		{
 			$paths = config('Paths');
 
-			$viewPath = $paths->viewDirectory;
+			$viewPath = ROOTPATH.$paths->viewDirectory;
 		}
 
 		return new \CodeIgniter\View\View($config, $viewPath, static::locator(true), CI_DEBUG, static::logger(true));


lonnieezell:
It's rare but I have come across applications that shared views, which were placed outside of the application's root folder. 

If the problem is that it's not correctly finding .tpl files, then the fix should be done where it's locating those. I thought someone had already submitted a fix for that in the past.

sugenganthos:
I run that code step by step and found that view path isn't correctly set. Then follows bootstrap.php which has ROOTPATH prefix for setting APPPATH,WRITEPATH,and BASEPATH

sugenganthos:
> It's rare but I have come across applications that shared views, which were placed outside of the application's root folder.

OK, I got it. So that is why Paths.php have that settings now. 
On shared server or hosting services, the developer doesn't know the absolute path, only relative path from their directory. So I can't specify that on Paths.php. 
Also the Paths.php is called by bootstrap.php for specifying APPPATH,BASEPATH,etc. So I think
specifying `public $viewDirectory = APPPATH.'Views';` on Paths.php doesn't work

PR #1515, opened 2018-11-19
Update date_helper.php

sugenganthos:
Add validate date string or time string.
This should be in input filter i think.  But still don't know how to integrate there

### DIFF
--- a/system/Helpers/date_helper.php
+++ b/system/Helpers/date_helper.php
@@ -62,3 +62,78 @@ if (! function_exists('now'))
 		return mktime($hour, $minute, $second, $month, $day, $year);
 	}
 }
+
+if (! function_exists('validate_date'))
+{
+	/**
+	 * Validate date string from <input type="date">.
+	 * Check if the string is valid date string (yyyy-mm-dd)
+	 * 
+	 * @param string $dateString
+	 *
+	 * @return bool
+	 */
+	function validate_date(string $dateString):bool 
+	{
+    		$dateArr = explode('-', $dateString);
+    		if(count($dateArr) === 3)
+		{
+			return checkdate($dateArr[1], $dateArr[2], $dateArr[0]);
+    		}else{
+			return FALSE;
+    		}
+	}
+}
+if (! function_exists('validate_time'))
+{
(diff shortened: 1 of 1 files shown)


system/Helpers/date_helper.php
```diff
@@ -62,3 +62,78 @@ function now(string $timezone = null)
...
+	 * Validate date string from <input type="date">.
+	 * Check if the string is valid date string (yyyy-mm-dd)
+	 * 
+	 * @param string $dateString
+	 *
+	 * @return bool
+	 */
+	function validate_date(string $dateString):bool 
+	{
+    		$dateArr = explode('-', $dateString);
```
lonnieezell: There is no way to know that a date string will be separated by dashes. In the US, I know that's common, but strtotime requires forward slashes to know it's a UK time and know which position the month and the day are. I'm sure this applies to other parts of the world, also, but, the point is you can't guarantee, so you'll need to find another way to get the date.

bvrignaud: I think we must have have another param to specify the date string format expected.
In Ci3, I use this : validate_date(string $dateString, string $dateFormat).
In france, the commonly format is dd/mm/yyyy.

sugenganthos: These function is supposed to validate input from <input type="date"> and <input type="time"> elements. Not from user input. 
[html input on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) says that date value always yyyy-mm-dd. That is what sent to the server. That is different from displayed date which is regards on locale setting.
Also on validate_time.

lonnieezell: If that's the intent, then it should be in a Validation rule, not the date helper. The date helper is for general usage, which this doesn't work for. However, this method [already exists](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Validation/FormatRules.php#L381) as a validation rule, so this appears to not be needed for what you intend.

lonnieezell: The validate_time method should be made a validation rule, as well.

PR #1522, opened 2018-11-21
add property_exists check on BaseBuilder and BaseConnection for $this->$key set value

samsonasik:
To avoid create non-existente property values.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -233,7 +233,10 @@ class BaseBuilder
 		{
 			foreach ($options as $key => $value)
 			{
-				$this->$key = $value;
+				if (property_exists($this, $key))
+				{
+					$this->$key = $value;
+				}
 			}
 		}
 	}
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -338,7 +338,10 @@ abstract class BaseConnection implements ConnectionInterface
 	{
 		foreach ($params as $key => $value)
 		{
-			$this->$key = $value;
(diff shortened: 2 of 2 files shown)


jim-parry:
I'm not sold on this one. The original lets you inject properties into a QB at construction time, while the proposed change only lets you override them.
The block of code is not unit tested, nor can I find anything about it in the user guide, so there is no guidance as to the intent of this parameter.
Maybe it should be eliminated altogether.
@lonnieezell Can you shed light on this?

lonnieezell:
The database stuff is something I never did huge modifications to and so a lot of its inner workings is still a bit unknown to me. That's directly from [CI3](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_driver.php#L365) code, though. That's how the configuration is [passed to the driver](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/Config.php#L107).

From what I can tell the intent was to override, not inject in the first place, but I definitely haven't looked at all of the drivers we don't support. I don't see anything listed in the CI3 user guide about extra params for any of the databases there, so I think it's safe to assume override is all that is needed.

samsonasik:
with assumption, auto create class property field is allowed, it still needs a validation then, eg, when passed options has "numeric" key, ref https://3v4l.org/9L2kE that will make error

lonnieezell:
My assumption was that we were overriding only, not creating new class properties.

samsonasik:
so, the `property_exists()` check is correct?

PR #1525, opened 2018-11-22
remove unneeded try {} catch {} on @fopen at Cache FileHandler::writeFile()

samsonasik:
By providing `@` in front of `fopen()` will make it always return false on failure fopen, so the catch was never reached, `try {} catch {}` is not needed.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Cache/Exceptions/CacheException.php
+++ b/system/Cache/Exceptions/CacheException.php
@@ -5,6 +5,14 @@ class CacheException extends \RuntimeException implements ExceptionInterface
 	/**
 	 * @return \CodeIgniter\Cache\Exceptions\CacheException
 	 */
+	public static function forUnableToWrite(string $path)
+	{
+		return new static(lang('Cache.unableToWrite', [$path]));
+	}
+
+	/**
+	 * @return \CodeIgniter\Cache\Exceptions\CacheException
+	 */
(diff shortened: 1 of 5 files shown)


natanfelles:
I had issues because fopen is suppressed and the catch just returned false.

Testing the Cache Files, some tests return false but is not possible know why if the cache directory has not write permission. Nothing is logged too.

Maybe a note on user_guide about this is important.

natanfelles:
Also, the directories inside `writable/` comes with chmod 755... Coming with 777 in git clone could solve that.

samsonasik:
how about check with `is_really_writable()` function ?

natanfelles:
I see two options at the moment:

1. A note [in the user guide](https://codeigniter4.github.io/CodeIgniter4/libraries/caching.html#file-based-caching).
2. Check with `is_really_writable()` only once in the FileHandler constructor  and if return false throw an exception. This would be the quickest way to detect this problem.

jim-parry:
I like both the last suggestions from natanfelles.

samsonasik:
I've added note in userguide for use 777 to make path writable and use `is_really_writable()` check early in __construct.

samsonasik:
I've added unit test for check against `is_really_writable($path)` in __construct.

PR #1530, opened 2018-11-24
database performance improvement : use foreach() when possible

samsonasik:
Previously, there are `count()` and then loop with `for()`, the foreach actually should already can handle it.

**Checklist:**
- [x] Securely signed commits


lonnieezell:
Honestly, the performance difference you're going to see here is likely so close as to be non-existent. But this is cleaner, easier to read, and doesn't look like it will break anything so I'm down.

PR #1536, opened 2018-11-25
ext-json in composer.json

nowackipawel:
**Description**
ext-json in composer

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,6 +8,7 @@
         "php": ">=7.1",
         "ext-curl": "*",
         "ext-intl": "*",
+        "ext-json": "*",
         "kint-php/kint": "^2.1",
         "zendframework/zend-escaper": "^2.5"
     },


nowackipawel:
Possibly it is but there are many IDEs which shows warning when there is no ext-json specified in composer (ie phpstorm)

lonnieezell:
I believe the ext-json package refers to the old one from PECL, which we definitely do not require. Where does PHP Storm show an error related to that? That's what I use daily and haven't noticed it.

nowackipawel:
Here it is:
screenshot_2018-11-26_22-42-48

lonnieezell:
Ah, just found more information about this error, also: https://medium.com/@luis.barros.nobrega/ext-json-is-missing-in-composer-json-what-is-this-700a1fcad966

In that case - good catch. :) Merging.

PR #1538, opened 2018-11-25
WIP img fix(?) - html_helper

nowackipawel:
there was an issue when alt was defined in attributes array
was:
```empty or from $src array```

is:

```from $src if exists```
or
```from $attributes if there is no alt in $src array```

 Todo:
- parse $attributes if contains string instead of array to check if there is alt attribute in
however this version is better than existing one :)

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345

### DIFF
--- a/system/Helpers/html_helper.php
+++ b/system/Helpers/html_helper.php
@@ -158,15 +158,7 @@ if (! function_exists('img'))
 
 		if (! isset($src['alt']))
 		{
-			if(is_array($attributes) && isset($attributes['alt']))
-			{
-				$src['alt'] = $attributes['alt'];
-			}
-			//If there is no alt attribute defined, set it to an empty string.
-			else
-			{
-				$src['alt'] = '';
-			}
+			$src['alt'] = $attributes['alt'] ?? '';
 		}
 
 
--- a/system/Helpers/html_helper.php
+++ b/system/Helpers/html_helper.php
@@ -156,11 +156,19 @@ if (! function_exists('img'))
 			$src = ['src' => $src];
 		}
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: system/Helpers/html_helper.php [hunk 1]
```diff
@@ -156,12 +156,20 @@ function img($src = '', bool $indexPage = false, $attributes = ''): string
 			$src = ['src' => $src];
 		}
 
-		//If there is no alt attribute defined, set it to an empty string.
 		if (! isset($src['alt']))
 		{
-			$src['alt'] = '';
+			if(is_array($attributes) && isset($attributes['alt']))
```
lonnieezell: This would be cleaner as a ternary.

system/Helpers/html_helper.php
```diff
@@ -184,6 +192,12 @@ function img($src = '', bool $indexPage = false, $attributes = ''): string
 			}
 		}
 
+		// prevent passing "alt" to stringify_attributes
+		if(is_array($attributes) && isset($attributes['alt']))
```
lonnieezell: If we are grabbing this value and shoving it into $src, is there a reason we cannot remove it there, and not do a separate check here?

nowackipawel: Yes there is a reason to do it separately . 
It $src['alt'] exists and $attributes['alt'] exists then funcs will produce: from alt

When I tested it first unset was in line 164... but there is that special case as I mentioned above.

nowackipawel:
@lonnieezell is it still sth to change?

PR #1539, opened 2018-11-26
Validator does not need to be defined as properties

bangbangda:
**Description**
$validator  is only used in function validate(), so no external definition is required.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


lonnieezell:
It's made a class variable so that you can easily grab the errors from it. The way you have it all we get is the success/fail indicator and there's no way to retrieve errors and display them in the views.

PR #1544, opened 2018-11-27
only use of helper(array) on Session::get() instead of Session::__construct

samsonasik:
`helper('array');` on `Session` class only used in `get()` method. So, moved the `helper('array')` usage to `Session::get()` method.

**Checklist:**
- [x] Securely signed commits


natanfelles:
While in the constructor, the helper function is called only once. In the `get()` will be called several times. I do not think it's a good change.

samsonasik:
in other classes, use of helper is only when the helper functions will be called, it mostly for consistency

lonnieezell:
While consistency is a good thing, in this case, I think it should stay in the constructor because @natanfelles is correct that we'll see it called a lot more than needed, and include_once is never the fastest command.

PR #1545, opened 2018-11-27
use app_timezone() in Response and DownloadResponse when app timezone defined

samsonasik:
If Application timezone defined in configuration, it will read the timezone defined instead of 'UTC'.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/HTTP/DownloadResponse.php
+++ b/system/HTTP/DownloadResponse.php
@@ -269,7 +269,7 @@ class DownloadResponse extends Message implements ResponseInterface
 	 */
 	public function setDate(\DateTime $date)
 	{
-		$date->setTimezone(new \DateTimeZone('UTC'));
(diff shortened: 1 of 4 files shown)


system/HTTP/DownloadResponse.php
```diff
@@ -269,7 +269,7 @@ public function getReason(): string
 	 */
 	public function setDate(\DateTime $date)
 	{
-		$date->setTimezone(new \DateTimeZone('UTC'));
+		$date->setTimezone(new \DateTimeZone(app_timezone() ?? 'UTC'));
 
 		$this->setHeader('Date', $date->format('D, d M Y H:i:s') . ' GMT');
```
natanfelles: If the `app_timezone()` is `America/Sao_Paulo` it is not `GMT`. [See also](https://tools.ietf.org/html/rfc7231#section-7.1.1.1).

samsonasik:
ok, i’m closing it.

PR #1548, opened 2018-11-28
allows to set empty html attr

nowackipawel:
**Description**
as for now there was no way to set empty attribute like required or autofocus

### DIFF
--- a/tests/system/Helpers/FormHelperTest.php
+++ b/tests/system/Helpers/FormHelperTest.php
@@ -816,7 +816,7 @@ EOH;
 	public function testFormParseFormAttributesTrue()
 	{
 		$expected = 'readonly ';
-		$this->assertEquals($expected, parse_form_attributes(['readonly' => true]));
+		$this->assertEquals($expected, parse_form_attributes(['readonly' => true], []));
 	}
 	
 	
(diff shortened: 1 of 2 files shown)


lonnieezell:
Looks good, but we need some tests added to test that feature please.

nowackipawel:
@lonnieezell : done.

tests/system/Helpers/FormHelperTest.php
```diff
@@ -810,5 +810,60 @@ public function testSetRadioDefault()
...
+		$expected = 'readonly ';
+		$this->assertEquals($expected, parse_form_attributes(['readonly' => true], []));
+	}
+	
+	
+	// ------------------------------------------------------------------------
+	public function testFormParseFormAttributesFalse()
+	{
+		$expected = 'disabled ';
+		$this->assertEquals($expected, parse_form_attributes(['disabled' => false], []));
```
lonnieezell: Shouldn't a false setting on a value make it so that it does not appear in the attributes list?

nowackipawel: I thought we have more flexibility outside form helper function if we will treat true as well as false in the same way.
However I know what you are thinking so I am open to changes.

lonnieezell: That's fair. I can live with that, and it's still valid HTML, so I'm good. Will merge.

PR #1551, opened 2018-11-28
remove manual define "system/" directory prefix at ComposerScripts

samsonasik:
So in case the "system" directory is renamed, the composer update will install to new renamed directory/ThirdParty directory.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/ComposerScripts.php
+++ b/system/ComposerScripts.php
@@ -49,7 +49,7 @@
  */
 class ComposerScripts
 {
-	protected static $basePath = 'system/ThirdParty/';
+	protected static $basePath = 'ThirdParty/';
 
 	/**
 	 * After composer install/update, this is called to move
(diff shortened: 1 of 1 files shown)


samsonasik:
For example, I do fresh install, and rename system directory. In next month, I do composer update, and in that time, there is update for thirdparty part, eg: Zend\Escaper, so, without this, the updated dependency will not go to the new renamed system directory.

samsonasik:
this should be steps to reproduce:
- rename system directory
- update autoload "system" in composer.json
- remove {new system directory}/Thirdparty/*
- composer update

samsonasik:
without this change, we will get the following error when `system` directory renamed and do composer update:

screen shot 2018-11-29 at 8 56 17 pm

lonnieezell:
I have no issues with this. Looks like needed for those using composer who don't like our naming system :)

PR #1555, opened 2018-11-30
autoload clean up: remove Psr\Log namespace from composer.json

samsonasik:
To be used `system/Config/AutoloadConfig.php` file for its definition. 

**Checklist:**
- [x] Securely signed commits


samsonasik:
updated composer.json at admin/framework/composer.json as well.

PR #1563, opened 2018-11-30
id attribute support added for csrf_field

nowackipawel:
i.e. useful in autocomplete forms (it is faster to reach this field by id than by name or its position in form)

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -688,9 +688,9 @@ if (! function_exists('csrf_field'))
 	 *
 	 * @return string
 	 */
-	function csrf_field()
+	function csrf_field(string $id = null)
 	{
-		return '<input type="hidden" name="' . csrf_token() . '" value="' . csrf_hash() . '">';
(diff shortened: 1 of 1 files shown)


nowackipawel:
@lonnieezell : 
hi there, there is one functionality missing. it is gr8 that we could specify id attribute using this function but .... there is still issue if we want to specify it via "form_open" function. 

What do you think to specify it like this?:

```
function form_open('',  ['csrf_id' => 'new-id'])
```

Honestly I don't like this idea cuz this attribute does not like naturally ... do you have any ideas to achieve it? It will be great to have possibility to define csrf field attribute in form_open as well.

lonnieezell:
There's not an elegant way that I can really think of for that. The bigger question is - how important is it to be able to customize the id for this?

nowackipawel:
Using forms which can be sent by regular post or by javascript call is good to be able to get value of csrf value easily. Additionally getting this value by elements id is the most efficient way .
Thx.

nowackipawel:
PR#1633

nowackipawel:
#1633

PR #1570, opened 2018-12-02
Routes dashes in URI's to be converted. Fixes #1564

### LINKED ISSUE #1564: Maybe routes has problem. [setTranslateURIDashes]
I use this config.
 setTranslateURIDashes(true)

I think that if I connent 'users/login-process' (dash) then,
should be connected to 'login_process' (underscore) method of 'users' controller.
but contrary to my expectation, just I could see 404 page not found error page.
So, I've checked source.
I think that something important source is missing.

so now I'm using below modified code temporarily.

in: system/Router/Router.php :: autoRoute() function

	if ( ! empty($segments))
	{
		$this->method = array_shift($segments);
	}

        /**
         * Added - BEGIN
         */
        if ($this->collection->shouldTranslateURIDashes()) {
            $this->translateURIDashes = true;
            $this->method = str_replace('-', '_', $this->method);
        }
        /**
         * Added - END
         */

So, could you please check this point?

bangbangda:
**Description**
Whether dashes in URI's should be converted to underscores when determining method names.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -700,7 +700,9 @@ class CodeIgniter
 
 		ob_start();
 
-		$this->controller = $this->router->handle($path);
+		// Scans the URI and attempts to match the current URI
+		$this->router->handle($path);
+		$this->controller = $this->router->controllerName();
 		$this->method     = $this->router->methodName();
 
 		// If a {locale} segment was matched in the final route,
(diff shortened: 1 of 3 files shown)


eterv:
sorry. I tested your code, but, no change. and please see below my reply.
https://github.com/codeigniter4/CodeIgniter4/issues/1564#issuecomment-443483480

PR #1572, opened 2018-12-02
Declare some classes as abstract

natanfelles:
**Description**

I think these classes were done to be extended.

If you think there is no such need, please just close it.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


jim-parry:
That feels "funny".
The fact that Model has a constructor tells me that is is meant to be instantiable.
Controller, on the other hand, is not meant to be instantiable. That has come in conversation before, with the intent being that an extending class might have a constructor, but we don't want to rely on that inside our class.
Message feels like a concrete class to me, given its flexibility.

I think it better to keep these as base classes, and not mess with trying to treat them as abstract. Fewer hassles.
It also looks like making the controller abstract messes up some of the output buffering - risky tests in travis-ci.

lonnieezell:
The other question to look at, besides are they meant to be instantiated, is where or not they have any methods the extending class are required to implement. If the answer is no, to that, I don't see any benefit from making them an abstract class. The definition I've always understood for an abstract class was it was one that had abstract methods. None of these do, so I say keep them as non-abstract classes.

jim-parry:
So, not a good class for abstract classes, sorry.

PR #1575, opened 2018-12-02
Adds ?integer, ?double, ?string, etc.  cast types :)

nowackipawel:
Adds new cast types which are casting only if value !== null

i.e. I've got an nullable col in my dbtable which stores strings (empty or not) and I'd like to have possibility to determine if string is null or '';
Now I cannot when I define casting as a string. with this PR if I define ?string in 

```

	protected $_options = [
		'casts' => [
			'usr_id' 			=> 'integer',
			'usr_usp_id' 		=> 'integer',
			'usr_provider_id' 	=> 'string',

			'usr_exc_id'		=> 'integer',
			'usr_drs_id'		=> 'integer',
			'usr_dlc_id'		=> '?string',
			'usr_tzo_id'		=> 'integer',

			'usr_email' 		=> 'string',
			'usr_password' 		=> 'string',

			'usr_first_name' 	=> 'string',
			'usr_last_name' 	=> 'string',
			'usr_nick_name'		=> 'string',

			'usr_is_active' 	=> 'boolean',
			'usr_is_blocked' 	=> 'boolean',
			'usr_is_picture' 	=> 'boolean',
		],
		'dates'=> ['usr_created_at', 'usr_updated_at'],
		'datamap' => []
	];
```
I will get NULL or '' for usr_dlc_id according to database value.

```
if(substr($type, 0, 1) === '?' && $value === null) return null;
```

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -412,6 +412,7 @@ class Entity
 
 	/**
 	 * Provides the ability to cast an item as a specific data type.
+	 * Add ? at the beginning of $type  (i.e. ?string) to get NULL instead of castig $value if $value === null
 	 *
 	 * @param $value
 	 * @param string $type
@@ -426,6 +427,7 @@ class Entity
 		switch($type)
 		{
 			case 'integer':
+			case 'int': //alias for 'integer'
 				$value = (int)$value;
 				break;
 			case 'float':
@@ -438,6 +440,7 @@ class Entity
 				$value = (string)$value;
 				break;
 			case 'boolean':
+			case 'bool': //alias for 'boolean'
 				$value = (bool)$value;
 				break;
 			case 'object':
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -421,6 +421,8 @@ class Entity
 
 	protected function castAs($value, string $type)
 	{
+		if(substr($type,0,1) === '?' && $value === null) return null;
+
 		switch($type)
 		{
 			case 'integer':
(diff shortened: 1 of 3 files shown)


lonnieezell:
Sounds good. Needs documented and tested.

system/Entity.php
```diff
@@ -421,6 +421,8 @@ protected function mutateDate($value)
 
 	protected function castAs($value, string $type)
 	{
+		if(substr($type,0,1) === '?' && $value === null) return null;
+
 		switch($type)
 		{
 			case 'integer':
```
natanfelles: @nowackipawel Since you're around here, could you also add `int` as a case here and `bool` with the boolean below? Thanks.

nowackipawel: I think we shouldn't use aliases but I added it with additional comment that bool/int are aliases.

nowackipawel:
@lonnieezell : docs, tests, aliases for integer and boolean - added

system/Entity.php
natanfelles: The more correct return types are int and bool and not integer and boolean. https://github.com/codeigniter4/coding-standard/issues/22 . Currently,  integer and boolean are the aliases.

Test:

```php
<?php

function foo(?integer $x):int {
	return 2 * $x;
}

echo foo(5);

```

natanfelles: So, the comment is wrong.

nowackipawel: You know... or not I'm lazy and I have almost 60 entities where I use integer and boolean as was in docs :) ... and I don't like aliases and I don't want to change classes.... but ofc you are right tl;dr; I've changed it.

PR #1580, opened 2018-12-03
Update Controller.php

sugenganthos:
I think this is potential of security flaw:

I edit system code:

`public function cachePage(int $time){
    CodeIgniter::cache($time);
    echo 'should not execute';
}
`

And put on url: http://mysite.test/anycontroller/cachePage/10

And it executed (the message appears)

So I have 2 suggestion for this:
1. Have every method on CodeIgniter\Controller and its parent to be declared as 'protected' so it cannot accessed by router
2. Have every method on CodeIgniter\Controller and its parent to be prefixed with _ (single underscore) so it will excluded by router

Thankyou

### DIFF
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -150,7 +150,7 @@ class Controller
 	 *
 	 * @param integer $time
 	 */
-	public function cachePage(int $time)
+	protected function cachePage(int $time)
 	{
 		CodeIgniter::cache($time);
 	}
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -150,7 +150,7 @@ class Controller
 	 *
 	 * @param integer $time
 	 */
-	protected function cachePage(int $time)
+	public function cachePage(int $time)
 	{
 		CodeIgniter::cache($time);
 	}


lonnieezell:
Good catch. I don't use the auto-routing much anymore so didn't think about that. Not every method can be made protected as they are called from outside classes. The underscore method I really am not a fan of, though it might be appropriate here. Will have to consider. 

I have no problem with the change in this PR, but it's causing a test to fail. Please update the tests so that they pass again.

sugenganthos:
My intend to make CI developer team notice this potential. Sorry, didn't know the other way for marking files.
For now, I still find a way to modify test. So for now I revert the function.

I catch it when writing some tracking code (web statistics) in constructor. It work as it should, but not as expected. The IncomingRequest object still NULL at this time. So I went through the router and controller code.

sugenganthos:
Just my curiosity, how you make routing works if turning autoroute off?
Let say, I have about 10 classes for 10 different sections. Do you mention every route with $route->get() one-by-one?

lonnieezell:
Correct. You would specify each available route explicitly. It's not nearly as bad as it might sound at first :)

jim-parry:
Nothing changed when the dust settles, so closing this.

PR #1581, opened 2018-12-03
ContentSecurityPolicy testing & enhancement

jim-parry:
- unit testing substantially complete, pending resolution of a few issues here
- reportOnly method & property were pointless - never referenced. I enhanced these to use the reportOnly property as a default until over-ridden, and adjusted the mutator methds accordingly
- upgradeInsecureRequests was not handled properly. I added a snippet inside buildHeaders() to generate the CSP tag

@lonnieezell Why is the sandbox property treated different from the rest (setX vs addX; boolean first parameter ignored unless empty property and second arg; if the boolean is used, you end with true or false instead of an array of sandbox policy names to enforce, and the test gets an error).


lonnieezell:
It's been a while :)  But the `addX` methods were because they could add multiple settings, where as `setX` methods typically overwrites the previous values. 

Looking back at the sandbox directive and reading over specs - I don't know why I did that, honestly. Can't see a reason for it now. Feel free to alter. :)

system/HTTP/ContentSecurityPolicy.php
```diff
@@ -750,6 +735,16 @@ protected function buildHeaders(ResponseInterface &$response)
 			'report-uri'      => 'reportURI',
 		];
 
+		// inject default base & default URIs if needed
+		if (empty($this->baseURI))
```
lonnieezell: I think your change at https://github.com/codeigniter4/CodeIgniter4/pull/1581/files#diff-340c29575fdbbbc1dd365d4e02e7ae3cR288 makes it so that `$this->baseURI` is never set, doesn't it?

system/HTTP/ContentSecurityPolicy.php
```diff
@@ -274,20 +272,20 @@ public function reportOnly(bool $value = true)
...
 	 *
-	 * @param string  $uri
-	 * @param boolean $reportOnly
+	 * @param string|array $uri
+	 * @param boolean|null $override
 	 *
 	 * @return $this
 	 */
-	public function setBaseURI($uri, bool $reportOnly)
+	public function addBaseURI($uri, ?bool $override = null)
```
lonnieezell: Not a fan of renaming this to `$override`. In IDE's that provide hints on what the variable is, it showing `override` as the hint doesn't give any clue what you're overriding.

system/HTTP/ContentSecurityPolicy.php
jim-parry: Sort of. It was originally set to 'none', and that cannot be over-ridden as that is the most restrictive that a CSP policy can be. It could start out as 'self', which can be added to (the policy allows multiple URLs as base), but none of them should start as 'none'. 
It isn't clear if the element is mandatory - doesn't appear so.
Thinking this through, and based on additional digging, I have changed the defaults for most of the *src directives.

system/HTTP/ContentSecurityPolicy.php
jim-parry: Made some changes to the naming and the explanation. Does this make more sense now?

jim-parry:
Oops - I have some cross-pollination happening, and some broken tests. Fixing!

system/HTTP/ContentSecurityPolicy.php
lonnieezell: Well, right, but when you call `setBaseUri()` it is only adding it to the attributes array, it's not actually setting `$this->baseURI`, so this line will always result in it setting the value to self and the user's option never actually gets used, does it?

jim-parry: Er, doesn't the unit test shows that if you "add" one or more baseURI origins, then it uses them, but if you don't, then it uses 'self'?

system/HTTP/ContentSecurityPolicy.php
lonnieezell: Yes, I think that makes more sense. Though $reportOnly might have been clearer but what you have is cool :)

system/HTTP/ContentSecurityPolicy.php
lonnieezell: Tests do show it working, yes, which means I think there's some dead code that's not actually being used in there regarding the $baseURI property.

jim-parry: Hmmm. I don't see it :(
There are two methods not fully covered by unit testing: generateNonces() and addToHeader().
$baseURI is used in addBaseURI() and generateHeaders().
I don't see what could be "dead".

lonnieezell: Never mind, my bad. I wasn't following the logic all of the way through, and forgot that `addOptions` would set the baseURI property directly. Silly me. 

Then I'm out of objections. Merge away :)

PR #1585, opened 2018-12-04
Type hint MigrationRunner methods 

### LINKED ISSUE #1577: Database Migrations inconsistent info about sequential type
Hello

This only applies to migration type sequential.
After this [commit](https://github.com/codeigniter4/CodeIgniter4/commit/08fbcbb54a28c3506083148f766d52e928471287#diff-1194e9e2013880118067cb327d95c75d) adding type hints to system migration class. The supplied version is parsed as int. 
If I want to migrate to version 002 it will parse this as int and therefore remove the 00. The class will then check for migration file with the name 2.php and not 002.php. Resulting in an error, migration file not found.
This conflicts with the old sequential style and the [docs](https://codeigniter4.github.io/CodeIgniter4/dbmgmt/migration.html?highlight=sequential).

natanfelles:
**Description**

Fixes  #1577

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/MigrationRunner.php
+++ b/system/Database/MigrationRunner.php
@@ -277,12 +277,12 @@ class MigrationRunner
 	/**
 	 * Sets the schema to the latest migration
 	 *
-	 * @param string $namespace
-	 * @param string $group
+	 * @param string|null $namespace
+	 * @param string|null $group
 	 *
 	 * @return mixed    Current version string on success, FALSE on failure
 	 */
-	public function latest($namespace = null, $group = null)
+	public function latest(string $namespace = null, string $group = null)
 	{
 		// Set Namespace if not null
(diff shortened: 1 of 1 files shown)


system/Database/MigrationRunner.php
```diff
@@ -436,7 +436,7 @@ protected function checkMigrations(array $migrations, string $method, int $targe
 		}
 
 		// Check if $targetversion file is found
-		if ($targetversion !== 0 && ! array_key_exists($targetversion, $migrations))
+		if ($targetversion !== '0' && ! array_key_exists($targetversion, $migrations))
```
natanfelles: String needed because the Coding Standard forces strict comparisons `!==`.

PR #1591, opened 2018-12-06
Update routing.rst

sugenganthos:
Help for custom 404 page with custom message
see #1590 for code update


jim-parry:
1) If this user guide change is for #1590, then it should be part of that PR and not a separate one.
2) Need improved grammar ... "Your custom error view is responsible to check if the message is empty or not, if you want to have this functionality"
3) You want to add a unit test for this new behavior, demonstrating its use.

jim-parry:
Add this/these to 1591.

PR #1595, opened 2018-12-08
Retry handle fatal error via pre_system

samsonasik:
Previously, I created PR #1510 which made coverage dropped significantly by placing `ob_start()` in  `system/Debug/Exceptions::initialize()`. 

In this PR, I am trying again this with set in in `pre_system` event.

Before patch:

screen shot 2018-12-08 at 3 41 24 pm

After patch: 

screen shot 2018-12-08 at 3 41 39 pm

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/application/Config/Events.php
+++ b/application/Config/Events.php
@@ -30,9 +30,12 @@ if (ENVIRONMENT !== 'production')
 	Events::on('DBQuery', 'CodeIgniter\Debug\Toolbar\Collectors\Database::collect');
 
 	Events::on('pre_system', function () {
-		\ob_start(function ($buffer) {
-			return $buffer;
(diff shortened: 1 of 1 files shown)


samsonasik:
@jim-parry I've updated again, now with use it if `ENVIRONMENT !== 'testing' check, so it will not influence the testing part on buffering process

samsonasik:
The coverage now remained the same at 72.762%

jim-parry:
It looks like this indeed solves your problem, but I have two concerns:
1) It feels funny to start output buffering, and not to close it. Am I missing something?
2) Is this going to cause problems for other code?

PR #1600, opened 2018-12-10
admin/starter/composer.json clean up

samsonasik:
Since `"codeigniter4/framework": "^4@alpha"` already in `require` part, the other libs that already in `codeigniter4/framework` should no longer needed to be registered in composer.json.

**Checklist:**
- [x] Securely signed commits


jim-parry:
Checks out :)
Sorry for the delay testing - crazy time of year!

PR #1601, opened 2018-12-10
clean up \Config\Services in Common.php

samsonasik:
`Config\Services` already in use statement.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -69,7 +69,7 @@ if (! function_exists('cache'))
 	 */
 	function cache(string $key = null)
 	{
-		$cache = \Config\Services::cache();
+		$cache = Services::cache();
 
 		// No params - return cache object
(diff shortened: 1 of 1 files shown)


jim-parry:
I'm fine with this

lonnieezell:
This should be fine. The only issue I can think of is when people do a copy/paste to app/Config/Services so they can override the method it's possible (though hopefully unlikely) they'll have a different services file used in that class. 

@jim-parry do you think this is worth keeping the full namespace in the class for? It shouldn't come up much, since all of the features expect you to call on Config\Services.

jim-parry:
I think that "synching" app after a new release is a different kettle of fish, and needs to be addressed separately.
This PR as it sits is good to go.

PR #1619, opened 2018-12-18
findColumn

nowackipawel:
**Description**
New model method with allows to get column values. Desired column is specified by name.

You will get array of column values as the result.

i.e.:
```
$items_ids = (new \DSCourseModel)->limit(100)->findColumn();
$items_drs_ids = (new \DSCourseModel)->limit(100)->findColumn(null, 'dsc_drs_id');
```
You could  use it in a other query  like this:
```
(new \DRSModel())->find($items_drs_ids);
or like this:
(new \DRSModel())->find(
   (new \DSCourseModel)->limit(100)->findColumn(null, 'dsc_drs_id')
);
```
if you'd like...


system/Model.php
```diff
@@ -348,6 +348,42 @@ public function find($id = null)
...
+	 *
+	 * @param mixed|array|null $id One primary key or an array of primary keys
+	 * @param mixed|array|null $column_name Column name or null for primaryKey
+	 *
+	 * @return array|null array of table's column values
+	 */
+	public function findColumn($id = null, string $column_name = null) : array
+	{
+
+		if(is_null($column_name))
```
lonnieezell: Prefer braces on conditions like this, each brace on it's own line please.

system/Model.php
```diff
@@ -348,6 +348,42 @@ public function find($id = null)
...
+			$column_name = $this->primaryKey;
+
+		if(!is_string($column_name))
```
lonnieezell: This might be an incorrect check. The type hint enforces it to be either a string or null. So checking if it's a string probably isn't right. Checking if it's null might be better. However, if you don't want to allow a column name parameter at all, then remove the possibility to make it null.

system/Model.php
```diff
@@ -348,6 +348,42 @@ public function find($id = null)
...
+	/**
+	 * Fetches the column of database from $this->table with a primary key
+	 * matching $id.
+	 *
+	 * @param mixed|array|null $id One primary key or an array of primary keys
+	 * @param mixed|array|null $column_name Column name or null for primaryKey
+	 *
+	 * @return array|null array of table's column values
+	 */
+	public function findColumn($id = null, string $column_name = null) : array
```
lonnieezell: variables in CI4 should be camelCased, so `$columnName`

system/Model.php
```diff
@@ -348,6 +348,42 @@ public function find($id = null)
...
+	{
+
+		if(is_null($column_name))
+			$column_name = $this->primaryKey;
+
+		if(!is_string($column_name))
+			throw DataException::forFindColumnIsNotAString();
+
+		$this->select($column_name);
+		$tmp = $this->asArray()->find($id);
```
lonnieezell: Why not chain these two statements together?

system/Model.php
```diff
@@ -348,6 +348,42 @@ public function find($id = null)
...
+
+			foreach ($tmp as $item)
+			{
+				$data[] = $item[$column_name];
+			}
+
+			return $data;
+		}
+
+		return null;
```
lonnieezell: No need to specify `null`. Just `return` does it

system/Model.php
nowackipawel: Will change it in a moment. 
However I found some snake case variables as a function parameters:
 string $end_char = '&#8230;' $end_char $curl_options... so it is not consistent yet :).

system/Model.php
nowackipawel: Sure. Changed. Will update in a while.

system/Model.php
nowackipawel: don't know :). will change it.

system/Model.php
nowackipawel: I extend my CI4 to support arrays for primaryKey that's why I am checking it here for a string. 

However if null is passed as the 2nd arg and primaryKey is not set or is set to null then ... it is worth to throw exception.

system/Model.php
nowackipawel: true..

nowackipawel: But... I like to specify return value which is ?array  for now.... and when I am using return instead of return null I get:
```
A function with return type must return a value (did you mean "return null;" instead of "return;"?) 
```

system/Model.php
```diff
@@ -348,6 +348,41 @@ public function find($id = null)
...
+		$columnName = $columnName ?? $this->primaryKey;
+
+		if(!is_string($columnName))
+		{
+			throw DataException::forFindColumnIsNotAString();
+		}
+
+		$tmp = $this->select($columnName)->asArray()->find($id);
+
+		if(count($tmp))
```
samsonasik: can be faster with truthy check: `if ($tmp)`

nowackipawel: I didn't see any benchmarks for this. thx for letting me know.

I made some tests:

```
<?php //7.2.4
    
    $i1 = 8000000;
    
    $d = $tmp = null;
    $a = 'aa';
    
    var_dump($d);
    
    $t = microtime(true);
    for($i = 0; $i < $i1; $i++)
    {
        if(!$d) { $tmp = $a;  }
    }
    print 'if(!$d) - do not want to use it : ' . (microtime(true) - $t) . PHP_EOL;
    
    
    $t = microtime(true);
    for($i = 0; $i < $i1; $i++)
    {
        if($d == null) { $tmp = $a; }
    }
    print 'd == null - option 1a: ' . (microtime(true) - $t) . PHP_EOL;
        
    $t = microtime(true);
    for($i = 0; $i < $i1; $i++)
    {
        if(null == $d) { $tmp = $a; }
    }
    print 'null == d - option 1b: ' . (microtime(true) - $t) . PHP_EOL;
    
    
        
    $t = microtime(true);
    for($i = 0; $i < $i1; $i++)
    {
        if($d === null) { $tmp = $a; }
    }
    print '$d === null - option 1c: ' . (microtime(true) - $t) . PHP_EOL;
        
    $t = microtime(true);
    for($i = 0; $i < $i1; $i++)
    {
        if(null === $d) { $tmp = $a; }
    }
    print 'null === $d - option 1d: ' . (microtime(true) - $t) . PHP_EOL;
    
    
    $t = microtime(true);
    for($i = 0; $i < $i1; $i++)
    {
        if(is_null($d)) { $tmp = $a; }
    }
    print  'is_null - option 2: ' . (microtime(true) - $t) . PHP_EOL;
    
    $t = microtime(true);
    for($i = 0; $i < $i1; $i++)
    {
        $tmp = $d ?? $a;
    }
    print '?? - option 3: ' . (microtime(true) - $t) . PHP_EOL;
```

```
if(!$d) - do not want to use it : 0.15235209465027
d == null - option 1a: 0.21860098838806
null == d - option 1b: 0.22835683822632
$d === null - option 1c: 0.18482804298401
null === $d - option 1d: 0.15287089347839
is_null - option 2: 0.15867710113525
?? - option 3: 0.15346217155457

if(!$d) - do not want to use it : 0.15974879264832
d == null - option 1a: 0.23204493522644
null == d - option 1b: 0.25987100601196
$d === null - option 1c: 0.15756487846375
null === $d - option 1d: 0.14941382408142
is_null - option 2: 0.14943194389343
?? - option 3: 0.15713381767273

if(!$d) - do not want to use it : 0.14737105369568
d == null - option 1a: 0.20844984054565
null == d - option 1b: 0.21335792541504
$d === null - option 1c: 0.15279507637024
null === $d - option 1d: 0.14113306999207
is_null - option 2: 0.14232611656189
?? - option 3: 0.14513802528381

```

IMHO ?? looks ok, but let's say is_null is slightly faster.

lonnieezell: As a complete aside - there is the new [Iterator](https://codeigniter4.github.io/CodeIgniter4/testing/benchmark.html#using-the-iterator) that can help you make those tests, too....

lonnieezell: For example, to test the truthy vs count, I whipped up the following: 

```
$tests = new Iterator();

$array = [1,2,3,4,5,6,7,8,9,10];

$tests->add('count', function() use($array) {
	if (count($array)){}
});

$tests->add('truthy', function() use($array) {
	if ($array){}
});

echo $tests->run(10000);
```

which gave the following output:

```

Test | Time | Memory
-- | -- | --
count | 0.0018 | 0 Bytes
truthy | 0.0016 | 0 Bytes
```

The time varied a bit but on average they were really damn close, with the truthy check being slightly faster. But let's be a little realistic here: I had to run that check **10,000** times to be able to measure the difference, which makes the difference between the so slight as to not even be noticeable during any application I can think of.

system/Model.php
lonnieezell: That's interesting, and good to know. In the past they've always been treated the same, but I've never tried specifying it with an optional return type either. Cool - let it ride.

system/Model.php
lonnieezell: Agreed - but my point was that you should probably check if it is null, not if it's a string, since PHP enforces that it will be  string if it's present.

nowackipawel:
@lonnieezell  could you write me pls in which file I should write tests for this?

lonnieezell:
Test it in the ModelTest file.

lonnieezell:
Sorry - was on holiday for the last week.

No need to create it. The test file exists at `tests/system/Database/Live/ModelTest.php`. 

If you look at `tests/_support/Database/Seeds/CITestSeeder.php` you'll see data that exists in that database at the time the tests are run. Does that help?

atishamte:
Is this PR completed?

nowackipawel:
I am using findColumn as well as other Model's methods like "parseId()" in my codeand they are working ...  but I am not sure if this PR is up to date.. and I won't be able to write tests in next month or two.

PR #1626, opened 2018-12-25
WIP clean up Paths::$viewDirectory property

samsonasik:
up to `Views` directory can be one up directory and `/Views` from current Config directory, no need to one more up.

```php
__DIR__ . '/../Views';
```

is enough.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Config/Paths.php
+++ b/app/Config/Paths.php
@@ -35,7 +35,7 @@ class Paths
 	 *
 	 * NO TRAILING SLASH!
 	 */
-	public $appDirectory = __DIR__ . '/../../app';
+	public $appDirectory = __DIR__ . '/..';
 
 	/*
 	 * ---------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jim-parry:
This feels odd. if 
public $appDirectory = __DIR__ . '/../../app';
then would it not make sense that the view directory would remain as
public $viewDirectory = __DIR__ . '/../../app/Views';
??
Your change places the Views folder at the same level as app, which is not correct.

samsonasik:
Ok, I've updated `Paths::$appDirectory as well` as well

jim-parry:
My apologies - I think you were right all along.
I incorrectly assumed that  DIR referred to the script being executed initially, namely index.php.
$appDirectory as modified is better, as it does not assume that the folder is called "app".

I'm going to revisit the user guide and shoot for a clearer explanation :)

PR #1627, opened 2018-12-25
if the table is defined by including an alias

titounnes:

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345


lonnieezell:
What problem is this trying to fix? I believe the table stuff should be taken care of by the db connection classes or query builder. If they aren't you shouldn't wrangle it in the model, but fix it deeper in the system.

PR #1633, opened 2018-12-27
Uses csrf_field and form_hidden instead of inline-html in form_open 

nowackipawel:
Allows to pass the ID of CSRF field as a key of form_open $attribute arr.
`form_open(['csrf_id' => 'id-foo-bar']);`

Uses csrf_field for generating CSRF field.
Uses form_hidden instead of inline-html in form_open.


jim-parry:
@lonnieezell The doc change looks good to me.

PR #1634, opened 2018-12-28
When running update return "true" not primary key of data

titounnes:

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -467,18 +467,17 @@ class Model
 
 		if (is_object($data) && isset($data->{$this->primaryKey}))
 		{
-			$response = $this->update($data->{$this->primaryKey}, $data);
+            		$this->update($data->{$this->primaryKey}, $data);
+            		return $data->$this->primaryKey;
 		}
-		elseif (is_array($data) && ! empty($data[$this->primaryKey]))
-		{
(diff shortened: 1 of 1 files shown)


nowackipawel:
Won't be cleaner to change:
`return $data->$this->primaryKey;`
to
`return $data->{$this->primaryKey}` in line 471?

Because you didn't give any details instead of PR title :) could you please write more details what this is about? How this will improve CI?

titounnes:
thanks for the correction. you are right.

lonnieezell:
Can you explain why you think this is needed? 

If you're doing an update you already know the primary key since you have to pass it in, so the only thing you're really interested in is whether it worked or not, right?

titounnes:
the resulting feedback is more informative regarding the data involved. And it can be used to simplify when we need to change data in a table that has a relationship if needed.

lonnieezell:
Sorry - I don't see how it simplifies anything since you've already got the data it returns. I'll pass on this one. Thanks. Look forward to other contributions, though!

PR #1441, opened 2018-11-09
WIP Adds support for non-empty string in delete()

nowackipawel:
** Description **
Allows to pass: int, non-empty string, associative or indexed array as $id
(almost the same as for update() in PR#1325)

I am not sure why someone assumed that primaryKey has to be an int :( .

PR #1560, opened 2018-11-30
update - validation only provided elements

nowackipawel:
According to #1559 I write some code which allows to filter validationRules to that elements which are provided in $data.

----

Allows to filter validation rules to only that elements which are in $data passed to update method
If you'd like to check only fields passed in $data during update just call:
$this->update($your_id, $your_data, true);

if there is 5 rules in your $validation_rules and  only two of them in $your_data array then only that two elements will be checked in validation process if you pass true as a 3rd arg (if you don't provide it or set to false then update will check every rule as is now). 

Useful: i.e. you want to only update user nickname or password instead of passing all user attributes.

Each pull request should address a single issue, and have a meaningful title.

**Description**
Explain what you have changed, and why.

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

---------Remove from here down in your description----------

**Notes**
- Pull requests must be in English
- If the PR solves an issue, reference it with a suitable verb and the issue number
(e.g. fixes <hash>12345
- Unsolicited PRs will be considered, but there is no guarantee of acceptance
- Pull requests should be from a feature branch in the contributor's fork of the repository
  to the develop branch of the project repository

Issue #865, opened 2017-12-14
Model::delete method


daylightstudio:
I'm curious why the Model::delete method is setup to only take an $id and and then there is a separate Model::deleteWhere? If Model::delete was more generic I think it could accommodate more conditions. As it stands now, the Model class needs to be extended to create your own "deleteOrWhere" and "deleteWhereIn" methods.

Thanks for all your hard work!

lonnieezell:
That's a fair point. Most of came simply from a base model I had used for years. I would create multiple small methods to handle specialized cases in each model so it was more readable, so the `delete` and `deleteWhere` were all that I needed. 

With the current model, though, you can access all of the Builder methods in your method chain, as well as the model methods, so it might make more sense to remove `deleteWhere` and rename `delete` to `deleteOne` or something, then allow the builder's normal delete method to be used. But, then that's not consistent with what `find` and `update` do, so not sure.... will have to think on that a bit.

timothymarois:
I think the delete should act like update/where clauses. I've been using the models myself and find it confusing why I can't do `delete(['name'=>'thisname']) `, Should delete whatever equals that query.  That's consistent with the update use, except you don't need to specify a second array/argument. That should be an easy fix, looking first to see if the user put an ID (primary key), or an array of IDs, and then next check if the user puts an array for where clause. This way the delete is used for how everyone would intend it. That's how I would assume it works given the `update` method.

lonnieezell:
@timothymarois I think `delete` is closer to `find`/`findWhere` in usage, honestly. `update` always requires a single id passed to it. 

Looking back over the model methods, it might make sense to make these changes: 

- `delete` should take either one id, or an array of ids. That is consistent with `find`.
- `delete` should also work like the Query Builder version when no parameters are passed in. This allows for `$model->whereIn('status', $statuses)->delete();` flow without the need for a bunch of new custom methods.

I think it would be tricky to reliably determine between an array of ids (which might be strings, also) and would stay consistent with the `find`/`findWhere` combo, which has a closer relation to `delete` than `update` does, I think. 

I'm tempted to say we get rid of `findWhere` and `deleteWhere` and allow both `find` and `delete` to work. Well - that doesn't quite make sense, actually. The `find` method is simply a helper method for `limit->get`.

timothymarois:
@lonnieezell Yeah You're right, I used a version before which reminded me I had `update(array, array)` but in the current state of CI4 it's only `update(id, array)`. 

[...]

PR #1644, opened 2019-01-02
Support for more complex conditions for update(),delete()

nowackipawel:
Replacement for #1441  and #1325  + improvements (outer function for parsing id, and better condition checking).

Allows to pass more complex data for update and delete methods.
Before delete and update supports only numeric $id. and produces: WHERE primaryKey IN(1); or WHERE primaryKey IN(2,3,4) - for arrays.

I've added support for:
- non empty string -> WHERE table.primaryKey = 'value'    `$id = 'value'`  (it is wired it doesn't support it and we are thinking about production version of ci4)
- associative arrays -> WHERE table.field = 'value'  `$id =['field' => 'value']`
- more complex associative arrays -> WHERE table.field = 'value' AND table.fieldin IN(2,3,4) `$id =['field' => 'value', 'fieldin' => [2,3,4]]`

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -680,15 +680,64 @@ class Model
 	}
 
 	//--------------------------------------------------------------------
+	
+	/**
+	 * @param $id integer|string|array|null:
+	 * 		int|string:  WHERE $this->table.$this->primaryKey = $id
+	 * 		indexed array: WHERE $this->table.$this->primaryKey IN ([2,3]) <i.e.: $id = [2,3]>
+	 * 		associative array: WHERE $this->table.key1 = 1 AND $this->table.key2 IN([2,3]) <i.e.: $id = [key1 => 1, key2 => [2,3]]>  
+	 * @return BaseBuilder
+	 */
+	private function parseId($id = null)
+	{
+
+		$builder = $this->builder();
+
+		// is numeric or is non empty string
+		if(!empty($id))
+		{
+			if(is_numeric($id) || !is_string($id))
+			{
+				$builder = $builder->where($this->table.'.'.$this->primaryKey, $id);
+			}
+			elseif(is_array($id))
+			{
+				//$id contains an indexed array
+				if(array_keys($id) === range(0, count($id) -1))
+				{
+					$builder = $builder->whereIn($this->table.'.'.$this->primaryKey, $id);
+				}
+				//$id contains associative array
+				else
+				{
+					foreach ($id as $key => $value)
+					{
+						if(!empty($value) && is_array($value))
+						{
+							$builder = $builder->whereIn($this->table.'.'.$key, $value);
+						}
+						else
+						{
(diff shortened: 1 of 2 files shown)


nowackipawel:
@lonnieezell , @jim-parry:
I'd like you to ask about help with writing unit tests if you are good with this PR. 
I am running out of time , and deadline with my projects are done.

It will be great to include in 4.0.0 release  #1560 as well I am using it in my project and it works like a charm :).

jim-parry:
I can help with the unit tests, in conjunction with 1560, but starting in about a week.

nowackipawel:
@jim-parry : it will be gr8. thx. I've sent you my telegram number as well as messenger in case you will need any information asap.

nowackipawel:
@jim-parry  could you let me know , if you need any help with this tests?

jim-parry:
Thank for for the reminder. I see that I offered to help mid-January,but didn't :(
I will make a new branch for these, so that both you & I can contribute code and PRs, until we are both happy.
That should set you up for unit testing on yor own in future :)

nowackipawel:
Oh. You don't have to create new branch for this :). I think that tests will be peace of cake for you:).  Honestly I have to jump to other framework with my work (rewriting project for ~800k users) for some time, cuz I thought CI beta will be ready at the begining of  March but I was wrong. I cannot risk to launch new one on alpha :(

Btw. I am not sure if last changes in models update method  are  the best companion to idea of $id (in update,delete calls) being flexible. Generally   I'm wondering why CI is going in the direction to limit functionalities in places where it is not necessary :(.
I.e. don't allow to complex primary keys (I had to extended my copy to support it at least partially)

lonnieezell:
> Btw. I am not sure if last changes in models update method are the best companion to idea of $id (in update,delete calls) being flexible. Generally I'm wondering why CI is going in the direction to limit functionalities in places where it is not necessary :(.
> I.e. don't allow to complex primary keys (I had to extended my copy to support it at least partially)

It was not the intention to limit functionality, just an oversight. The new Model is basically a collection of helper methods at it's core. It is patterned off of the myriad of MY_Model classes that people have made for previous versions of CI. Which means it's there to help do common stuff, not provide the 100% solution to everything. 

All of the other MY_Models' that I recall seeing have this same limitation. It's something that especially becomes problematic when digging into more advanced stuff, like ORM features. Which means that even other frameworks like Laravel don't fully support multiple columns for a primary key. And it's not that we're saying "don't implement this" in most cases, but that we haven't gotten around to it once it's been pointed out.

atishamte:
Is any update of this PR?

lonnieezell:
Actually - I think enough time has gone on that we can simply close this one. For one thing, primary keys are now required. For another, we had a [big discussion](https://github.com/codeigniter4/CodeIgniter4/issues/865) where we debated how these methods should work and not really interested in going through that again. Looking at the code, it looks like 1/2 or more of this PR should already be possible in the current codebase.

Issue #1646, opened 2019-01-02
I've noticed strage behaviour in Entity


nowackipawel:
My entity:

```
<?php namespace App\Entities\Question;

use CodeIgniter\Entity;

class QuestionStatMonth extends Entity
{
	protected $qsm_que_id;
	protected $qsm_year;
	protected $qsm_month;
	protected $qsm_ok;
	protected $qsm_nok;

	protected $_options = [
		'casts'   => [
			'qsm_que_id'  	=> 'integer',
			'qsm_year' 		=> 'integer',
			'qsm_month'   	=> 'integer',
			'qsm_ok'    	=> 'integer',
			'qsm_nok'    	=> 'integer',
		],
		'dates'   => [],
		'datamap' => [],
	];

	public function setQsmYear($value)
	{
		$this->qsm_year = ($value - 2000 > 0) ? $value - 2000 : null; 
print $this->qsm_year;
	}

	public function getQsmYear()
	{
		return ! is_null($this->qsm_year) ? $this->qsm_year + 2000 : null;
	}

}
```
when I set my $my_entity->qsm_year = 2017    I've got printed out 17 (what is correct), but when I am trying to save it in database I've got 2017 in DB. I think we should use similar approach as is in #1373 .

I will PR in a while cuz I'd like to see if it won't break any tests.

nowackipawel:
I think we should do this cuz methods like getMyProperty() and setMyProperty() should change property value outside the database. Am I wrong?

nowackipawel:
Already know it is more complicated even every test in travis were passed.


But it works ok as far as we don't have to validation rules like this:

``
		'qsm_year'		=> 'required|greater_than_equal_to[2000]|less_than_equal_to[2100]',
``
According to #1647 database value is as excepted (17 for 2017) but validation is breaking.
We could deal with this in model classToArray method and update/insert/etc. but I am not sure if I should check what we will be able to do with it.

Or we just should assume that in my case this rule should looks like:
``
		'qsm_year'		=> 'required|greater_than_equal_to[0]|less_than_equal_to[100]',
``
because is related to value in database ?

lonnieezell:
I think a better solution might be to create a new method, `toRawArray()` that simply returns all of the values as they currently are, ignoring casts and no using `_get()`. Then that can be used by the `insert/save` methods. 

Should take care of it, right?

PR #1647, opened 2019-01-02
Make entity values _raw for before passing to DB.

nowackipawel:
related to issue 1646

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -78,11 +78,11 @@ class Entity
 	protected $_original = [];
 
 	/**
-	 * Holds info whenever prperties have to be casted
+	 * Holds info whenever properties have to be casted or magic method should be used.
 	 *
 	 * @var boolean
 	 **/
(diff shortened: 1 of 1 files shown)


lonnieezell:
Yeah, not a fan of adding yet another argument to that method. As mentioned in #1646 I think a better solution would be a new method, `toRawArray()`.

nowackipawel:
There was already argument called $cast. I've renamed it cuz it doesn't mean just cast anymore. 

Pls lemme know what  your final decision is. Should I write method what you mentioned above ?

Btw: what about mine concerns about validation rules in this particular case? Should we pass $data converted by getMyVariable methods to validation and raw data to database (in insert/update)?  Won't it look weird ? 
In my opinion to make it fully work we need one type of behavior for raw - database,  second with using getMyVariable - validation, and third with getMyVariable and casting for normal work with entities. :/

PR #1649, opened 2019-01-03
remove php 7.3 from "allow_failures" in travis config

samsonasik:
<strike>Let see if it will make travis build success in php 7.3 env</strike>

**Checklist:**
- [x] Securely signed commits


samsonasik:
Green builds for php 7.1, 7.2, and 7.3:

screen shot 2019-01-04 at 4 12 15 am

PR #1652, opened 2019-01-04
Allows to use subDirs for cache.

nowackipawel:
If you put thousands of files in top of your cache directory it could have impact for performance.
For if you:
- specify cache with "/"  
- using FileHandler
- and  subdirectory in cache folder does not exist 
.... then cache is not saved.

### DIFF
--- a/system/Cache/Handlers/FileHandler.php
+++ b/system/Cache/Handlers/FileHandler.php
@@ -119,7 +119,8 @@ class FileHandler implements CacheInterface
 			'data' => $value,
 		];
 		
-		if(strpos(pathinfo($this->path . $key, PATHINFO_DIRNAME), WRITEPATH . 'cache') === 0 && strpos(!is_dir(pathinfo($this->path . $key, PATHINFO_DIRNAME)) && false === mkdir(pathinfo($this->path . $key, PATHINFO_DIRNAME), 0750, true))
+		$path = pathinfo($this->path . $key, PATHINFO_DIRNAME);
+		if(strpos($path, WRITEPATH . 'cache') === 0 && !is_dir($path) && !is_file($path) && false === mkdir($path, 0750, true))
 		{
 			return false;
 		}
--- a/system/Cache/Handlers/FileHandler.php
+++ b/system/Cache/Handlers/FileHandler.php
@@ -118,7 +118,12 @@ class FileHandler implements CacheInterface
 			'ttl'  => $ttl,
 			'data' => $value,
 		];
-
+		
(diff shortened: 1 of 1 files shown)


lonnieezell:
That's a fair point, but making a new directory for every key probably isn't the best solution either. 

Plus - no tests again.

nowackipawel:
@lonnieezell 
Ofc new directory for every key is not best solution, as long as in one particular directory will be less than up to 1000 files :-) . However if you have to create thousands of cache keys it is better to create subdirs (file system performance).

Additionally why we could NOT create keys with  "/" automatically using File driver  if we could do this easily with other drivers.  Why drivers are not compatible with each others?

lonnieezell:
I have no problem with it supporting a slash in the key name for compatibility reasons. However that doesn't mean allowing arbitrary creation of directories without proper name sanitization, etc, and I don't recall how much of that is done already since those drivers were ports from CI3 that I didn't touch much at all.

Even you admitted creating a directory for each key is not ideal, so why would we choose that particular solution to become part of core? While I have the feeling that the amount of milliseconds it takes to scan that directory on a decent server is minimal, and more than makes up for the savings in processing from caching itself, if we were to optimize this it would be for one of the more traditional methods that's already widely used that creates one or more layers of directories based on the key name itself to determine the folder structure. 

And, honestly, I think that's a good optimization to implement. But this PR doesn't reflect that :)

nowackipawel:
Thank you for creating issue (1658).

> Even you admitted creating a directory for each key is not ideal, so why would we choose that particular solution to become part of core?

The need for such a solution is related to it:
- ordering cache files according to the developer's intention,
- possibility of dividing cache files into different categories,
and then:
- to find files more easily,
- less potential of overwriting files using CI to multisite solutions or if someone is using just ID as a key (ofc there is no guarantee anyway but if we will be able to use / it will be easier to avoid it).

nowackipawel:
Oh and btw, please correct me if I am wrong but I am **almost** sure that my PR has some security improvements over actual version of FileHandler.
```
	$path = pathinfo($this->path . $key, PATHINFO_DIRNAME);
		if(strpos($path, WRITEPATH . 'cache') === 0 && !is_dir($path) && !is_file($path) && false === mkdir($path, 0750, true))
		{
			return false;
		}
```
this `if` which is missing in actual code guarantees that you are not able to overwrite any file outside cache dir. It is possible now if developer is using  values  from uri path (or any other available to user) as a keys. If attacker adds ../../ to this data he will be able to write files outside cache dir.

lonnieezell:
You're right, that is an improvement, but not the full improvement that I'm looking for.

PR #1686, opened 2019-01-24
Refactor/filters

jim-parry:
Move the needed filters into the  CodeIgniter namespace, and relegate Throttle filter to docs only.


jim-parry:
And this has uncovered missing unit tests ... coming :-/

PR #1689, opened 2019-01-25
Model Validation Fix

### LINKED ISSUE #1584: Model's update method fails when validation rules exist
---
name: Bug report
about: Model's update method fails when validation rules exist that refer to record fields that are not going to be updated.

---

**Describe the bug**
Model's update method fails when validation rules exist that refer to record fields that are not going to be updated. The problem is that Model's validation method for update uses all rules instead of only those rules that are related to the updated fields.

**CodeIgniter 4 version**
CodeIgniter 4.0.0 Alpha 2

**Affected module(s)**
Model

The update method below fails because the model's rules state that 'name' is required. But for updating name may not be updated (like in the test case) and therefore the rules related to name should not be part of the validation process.

```
  public function testUpdateWithValidation()
  {
  	$data  = ['description' => 'This is a first test!',
  						'id'          => 42,
  						'name'        => 'valid',
  						'token'       => 42];
  	$model = new ValidModel($this->db);
		$this->assertTrue($model->insert($data));
		$data['description'] = 'This is a second test!';
		unset($data['name']);
		$this->assertTrue($model->update($data)); // assertion fails because of missing 'name'
	}

```

lonnieezell:
Don't run validation rules for data that's not getting set. Fixes #1584

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -727,6 +727,7 @@ class Model
 		{
 			if ($this->validate($data) === false)
 			{
+				dd($this->errors());
 				return false;
 			}
 		}
@@ -1253,18 +1254,28 @@ class Model
 		{
 			$data = (array) $data;
 		}
-		dd($this->validationRules);
+
+		$rules = $this->validationRules;
+		$rules = $this->cleanValidationRules($rules, $data);
+
+		// If no data existed that needs validation
+		// our job is done here.
+		if (empty($rules))
+		{
+			return true;
+		}
+
 		// ValidationRules can be either a string, which is the group name,
 		// or an array of rules.
-		if (is_string($this->validationRules))
+		if (is_string($rules))
 		{
-			$valid = $this->validation->run($data, $this->validationRules, $this->DBGroup);
+			$valid = $this->validation->run($data, $rules, $this->DBGroup);
 		}
 		else
 		{
 			// Replace any placeholders (i.e. {id}) in the rules with
 			// the value found in $data, if exists.
-			$rules = $this->fillPlaceholders($this->validationRules, $data);
+			$rules = $this->fillPlaceholders($rules, $data);
 
 			$this->validation->setRules($rules, $this->validationMessages);
 			$valid = $this->validation->run($data, null, $this->DBGroup);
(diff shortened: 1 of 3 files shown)


nowackipawel:
@lonnieezell 
I think your approach is not perfect. Please correct me if I'm wrong.

In my version there was an option to pass optional parameter to model's update method with information if we'd like to skip rules for non-existing files or not (if we'd like to inform validation that we are know what we are doing and we are  sure skipping part of rules are safe for our app and that's is the reason validation won't be able to find some of fields in our data and should skip it <some of that fields could be defined as required>).

I cannot find that possibility in your code which could cause (in the worst scenario) in security issues when attacker will modify HTML form. Maybe I don't see something but in my opinion your changes will cause removing all of the rules for fields which couldn't be found in $data.?

I.e.
I see a difference between full entity creation, and editing - ALL or ONLY a part of entity properties. IMHO not all of that cases are safe if exposed to the public and we cannot decide when we'd like to filter validation rules or not.

lonnieezell:
What security issue do you envision? If any form comes in and you're simply passing the POST data then the key will exist in the data being checked and the validation will run. If you do any sort of data prep before passing it to the model you will find that it's missing. The only way that field could not be there that I can think of, is if some malicious javascript somehow got on your page and would completely remove the field prior to the form submitting. That sounds like a pretty rare case, and one that I don't think I've heard of happening in an attack before. 

As far as the difference between creation and editing goes, you're correct, but none of the validation process is up to the public unless you allow it to be. The fact that you can take POST data and shove it into validation is a convenience thing, and we've done what seems sensible to prevent mass assignment attacks. That's as close as it gets to being "exposed to the public". You can still easily choose whether those filters are applied by building the object that you pass to the model. If instead of passing POST unchecked, you create an object/array and assign defaults to all values during creation, then all fields will be checked against during validation. Likewise, if you create an Entity from POST fields then use the model's save() or insert() methods, then those fields will also exist and be validated. 

I think the simpler solution works fine here.

nowackipawel:
If you don't be able  (you won't have possibility to define it)  to specify if your  app could  filter and unset rules for non-existing data then attacker could modify your form using mentioned JS or just browser developer tools. That's why I think filtering  should be optional and turned OFF as default.

OFC validation process is not exposed for the public. Even in that case sometimes is easier to pass data directly to the model ... and then option which allows us to filter non-existing data rules or not will be very useful. In that case you don't need to create additionally Entity class at every update() and you don't have to specify all default values of Entity class for insert() <in that case developer  additionaly should take care of compliance of default entity values with database table definitions>;

I can imagine situation:
I've got ITEM record:
item_id = 1 (required)
item_name = 'abc' (required)
item_is_active = 1 (required), but default value is 0

I tried to make a form for editing only item_name ...
correct me if I am wrong but according to your approach I have three options:
1) get all data from db and pass everything to update (with item_is_active value, even if I don't want to edit it),
2) create entity which do not know actual value of is_active state...so I have to get it from DB before saving,
3) do not pass is_active and get validation error instead of  correct value


In other end if we will filter rules for non-existing post data rules:
I'd like to update my ITEMs item_name but in the same moment I'd like to pass hidden field with value of 0. Attacker could remove my hidden item_is_active easily and its value does not change in database. OFC you could and have to make it different way but it is only an example.

What I'd like to say I think option to define if filter should or souldn't work should be specified by developer.

REJack:
@lonnieezell Nice changes but this PR killed my validation rules 😞, now i can't check for `required` in __insert()__ anymore, it would be great to have this feature not in __insert()__.

lonnieezell:
@REJack Please open new issue, then, with examples on how it breaks required on inserts and how we can duplicate. I don't believe the code should have done that but if it is we need to check it out, obviously.

PR #1703, opened 2019-02-04
Update MigrationRunner.php

basic-app:
Without this sorting "$lastMigration = end($migrations)" code block in "latest()" and "latestAll()" functions return not last migration in some cases.


lonnieezell:
This is breaking a couple of tests. Please revise tests to make sure everything is working as expected.

atishamte:
Please close this PR. The suggested code is already there in the repo

PR #1709, opened 2019-02-07
Fix/rc

jim-parry:
Fix resource routing websafe method order checking


jim-parry:
Unrelated travis-ci failure (ThrottleTest)

PR #351, opened 2017-01-01
Fixes #345 form_open escape "/"

samsonasik:
Fixes #345

Issue #345, opened 2016-12-30
bug-form_open_multipart


startbbs:
`<?php echo form_open_multipart('user/login',['class'=>'form']);?>`
when I use form_open_multipart, it would create the follow code
`<form action="http://127.0.0.1/user/login" class="form" enctype="multipart&#x2F;form-data" method="post" accept-charset="utf-8">`

please note the enctype, '/' was changed to `&#x2F;`

samsonasik:
I give a try to fix it at #351

davidgv88:
Hi @samsonasik 

I Think that this is not a issue.

The form works and the file is upload succesfully.

All attributes are being escaped using escapeHtmlAttr function from Zend Escaper Library.

Please look this https://framework.zend.com/manual/2.1/en/modules/zend.escaper.escaping-html-attributes.html

lonnieezell:
Yup, even though it looks strange, it all works as expected in all of my tests.

PR #1714, opened 2019-02-10
form helper enctype unescaped

nowackipawel:


nowackipawel:
```enctype="multipart&#x2F;form-data"``` looks ugly :)

samsonasik:
escaping html attributes is fine, see https://github.com/codeigniter4/CodeIgniter4/issues/345 and https://github.com/codeigniter4/CodeIgniter4/pull/351 that closed.

nowackipawel:
thx for pointing this out. it is just idea... to make html clearer .

lonnieezell:
If you click on the travis Details link you can see what the exact errors that are happening are. 

However, we don't need to complicate the code with special case checks for the sake of making HTML look pretty. It's perfectly functional as is, and safe and accurate.

PR #1715, opened 2019-02-10
form action - should be optional

nowackipawel:
https://www.w3.org/TR/html50/forms.html#attr-fs-action
The action and formaction content attributes, if specified, must have a value that is a valid non-empty URL potentially surrounded by spaces.

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -53,16 +53,12 @@ if (! function_exists('form_open'))
 	 *
 	 * @return string
 	 */
-	function form_open(string $action = '', $attributes = [], array $hidden = []): string
+	function form_open(?string $action = '', $attributes = [], array $hidden = []): string
 	{
 		// If no action is provided then set to the current url
-		if (! $action)
+		if ($action === '')
 		{
 			$action = current_url(true);
-		} // If an action is not a full URL then turn it into one
-		elseif (strpos($action, '://') === false)
-		{
-			$action = site_url($action);
 		}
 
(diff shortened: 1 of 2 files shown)


lonnieezell:
I'm fine with the change with making it optional. However - there is zero need to change all of the URL's. Revert that change and I'll merge.

nowackipawel:
Why there is no need to change URL's? Are you writing about that ones in FormHelperTest.php?  
As for now  `form_open('foo/bar')` returns `action="foo/bar"` instead of `"http://example.com/index.php/foo/bar"` so... if I revert that change travis won't pass anymore.

If you are thinking that `foo/bar` as the first parameter of `form_open($action = 'foo/bar')` should produce `"http://example.com/index.php/foo/bar"` as it was before then we should create solution which allows us to generate just `foo/bar` as the form's action attribute cuz it is correct value for this attr.

lonnieezell:
Actually - I missed that you deleted the portion that generated a full URL from a relative url. That's in there as a security precaution, as a way of allowing coding as relative but the [SEO, etc benefits from full urls](https://moz.com/blog/relative-vs-absolute-urls-whiteboard-friday), and more. And that's how it's been [for a long time](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/helpers/form_helper.php#L74). 

Looking closer I can't accept this. It already allows you to not pass a URL. Whether it displays a URL in the generated code really is nothing anyone should worry about, and the method we have had in places for years is safer.

PR #1719, opened 2019-02-13
Update Cache RedisHandler to support select database.

vibbow:
**Description**
Update Cache RedisHandler to support select database.
As this feature already exist in Session RedisHandler

**Checklist:**
- [ Y ] Securely signed commits
- [ N/A ] Component(s) with PHPdocs
- [ N/A ] Unit testing, with >80% coverage
- [ Y ] User guide updated
- [ Y ] Conforms to style guide

### DIFF
--- a/system/Cache/Handlers/RedisHandler.php
+++ b/system/Cache/Handlers/RedisHandler.php
@@ -122,7 +122,7 @@ class RedisHandler implements CacheInterface
 
 			if (isset($config['database']) && ! $this->redis->select($config['database']))
 			{
-				//              log_message('error', 'Cache: Redis select database failed.');
+				log_message('error', 'Cache: Redis select database failed.');
 			}
 		}
 		catch (\RedisException $e)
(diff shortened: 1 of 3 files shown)


system/Cache/Handlers/RedisHandler.php
```diff
@@ -118,6 +119,11 @@ public function initialize()
 			{
 				//              log_message('error', 'Cache: Redis authentication failed.');
 			}
+
+			if (isset($config['database']) && ! $this->redis->select($config['database']))
+			{
+				//              log_message('error', 'Cache: Redis select database failed.');
```
lonnieezell: This shouldn't be commented out.

lonnieezell: Perfect. Thanks! Merging.

vibbow: I had commented this out to log an error message.

But in the mean time, I notice only `redis->connect` will throw an RedisException is connection is failed.

`redis->auth` will only return a false rather an RedisException if auth is failed.

Should I enable the log for that either?

lonnieezell: Oh, yes, we should definitely be logging all of those situations. Though I've already merged this PR so you'll probably have to make another. Thanks!

PR #1729, opened 2019-02-16
New View Layout functionality for simple template functionality.

lonnieezell:
Provides the ability to use layout views that can then be extended by child views, with content sections.


jim-parry:
There are still time execution timing errors messing with travis-ci. I will fix the ones I can.

PR #1738, opened 2019-02-18
Fix required_with rule bug. Fixes #1728

### LINKED ISSUE #1728: validation rule password never match
**Describe the bug**
The validation error is always triggered when use required_with[] check
The validation error is always triggered when use matches[] check

**CodeIgniter 4 version**
alpha-5

**Affected module(s)**
Form Validator

**Reproduce the behavior**
my rules : 
```
protected $validationRules    = [
        'password'     => 'required|min_length[8]',
        // 'password_confirm' => 'required_with[password]|matches[password]'
        'password_confirm' => 'matches[password]'
    ];
```

```
curl -X POST \
  localhost/users \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -H 'cache-control: no-cache' \
  -d 'password=toto&password_confirm=toto'
```

`"password_confirm": "The password_confirm field is required when password is present."`

remove the require_check[] 

`"password_confirm": "The password_confirm field does not match the password field."`

my validations_rules are in my model. everything seems to work exept this one.

**Context**
 - OS: LMDE2
 - Web server Nginx
 - PHP version 7.1

bangbangda:
**Description**
Fix required_with rule bug.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -304,9 +304,9 @@ class Rules
 		// If the field is present we can safely assume that
 		// the field is here, no matter whether the corresponding
 		// search field is present or not.
-		$present = $this->required($data[$str] ?? null);
+		$present = $this->required($str ?? '');
 
-		if ($present === true)
+		if ($present)
 		{
 			return true;
 		}
@@ -326,8 +326,8 @@ class Rules
 
 		// Remove any keys with empty values since, that means they
 		// weren't truly there, as far as this is concerned.
-		$requiredFields = array_filter($requiredFields, function ($item) use ($data) {
-			return empty($data[$item]);
+		$requiredFields = array_filter($requiredFields, function ($item) use ($data, $str) {
+			return ! empty($data[$item]);
 		});
 
 		return empty($requiredFields);
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -326,7 +326,7 @@ class Rules
 
 		// Remove any keys with empty values since, that means they
 		// weren't truly there, as far as this is concerned.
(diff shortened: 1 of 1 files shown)


system/Validation/Rules.php
```diff
@@ -304,9 +304,9 @@ public function required_with($str = null, string $fields, array $data): bool
 		// If the field is present we can safely assume that
 		// the field is here, no matter whether the corresponding
 		// search field is present or not.
-		$present = $this->required($data[$str] ?? null);
+		$present = $this->required($str ?? '');
```
lonnieezell: Without going back and digging into the logic, are you sure this is correct? Using `$str` instead of `$data[$str]` would check that the current field is required, not that the `with` field is required...

bangbangda: Yes. `$str` is field value. `$data` is Post data. so `$data[$str]` will never get the value.
show demo:
`protected $validationRules    = [
        'password'     => 'required|min_length[8]',
        'password_confirm' => 'matches[password]'
    ];`

`curl -X POST \
  localhost/users \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -H 'cache-control: no-cache' \
  -d 'password=toto&password_confirm=toto'`

`$str` is toto.  `$data['toto']` is null. so `$present` has always been false.

bangbangda: I tested is no problem.

Iamscalla:
I agree with this PR,  mind you,  required_without has same issue. Treated it, but unable to submit a PR.

PR #1742, opened 2019-02-19
Fix required_without rule bug.

bangbangda:
**Description**
- Unable to get the value of the validation field correctly.
- Update validation class  document.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -356,9 +356,9 @@ class Rules
 		// If the field is present we can safely assume that
 		// the field is here, no matter whether the corresponding
 		// search field is present or not.
-		$present = $this->required($data[$str] ?? null);
+		$present = $this->required($str ?? '');
 
-		if ($present === true)
+		if ($present)
 		{
 			return true;
 		}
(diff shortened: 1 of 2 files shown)


niteraven7:
@lonnieezell @bangbangda No, this does NOT fix this rule. I have a form that requires an email address OR phone number OR both. The validation rules are as follows:

`public $recipient = [
	'first_name'		=> 'required|min_length[1]|max_length[50]',
	'last_name'		=> 'required|min_length[1]|max_length[50]',
	'email'			=> 'required_without[number]|valid_email',
	'number'			=> 'required_without[email]|numeric',
];
`
I believe the validation code should look like:

`if (empty($data[$field]))
{
	return false;
}`

The field becomes required when any of the parameter fields are empty (not present or not set). Does that make sense?

bangbangda:
@niteraven7 
Check my [PR](https://github.com/codeigniter4/CodeIgniter4/pull/1747)
Empty is not ok, empty("0") is true.
Trim is ok.

PR #1747, opened 2019-02-20
WIP Fix required_without rule bug. v1.0.1

bangbangda:
**Description**
Verify that the field exists. no validation value is valid.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -364,10 +364,10 @@ class Rules
 		}
 
 		// Still here? Then we fail this test if
-		// any of the fields are not present in $data
+		// any of the fields are not present or value is '' in $data
 		foreach ($fields as $field)
 		{
-			if (! array_key_exists($field, $data))
+			if (! array_key_exists($field, $data) || trim($data[$field]) === '')
 			{
 				return false;
 			}
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -367,7 +367,7 @@ class Rules
 		// any of the fields are not present or value is '' in $data
 		foreach ($fields as $field)
 		{
-			if (! array_key_exists($field, $data) || trim($data[$field]) === '')
+			if (! array_key_exists($field, $data) || ! $this->required($data[$field] ?? ''))
 			{
 				return false;
 			}


niteraven7:
@bangbangda Looks good, I'm glad you acknowledged my logic. I would change code to:

`if (empty($data[$field]) && ($data[$field] != '0')) { return false; }`

Even though this should rarely happen - empty string ' ' should be valid input.

lonnieezell:
The required rules have always required that it is not empty or not 0. Empty strings have always failed the require rule. For cases like that, we do have the [permit_empty](https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html#available-rules) rule.

bangbangda:
Is this better? 
Support for empty arrays.
`if (! array_key_exists($field, $data) || ! $this->required($data[$field] ?? ''))`

@lonnieezell @niteraven7

niteraven7:
Hi @lonnieezell I think you may have confused him :)

I've been thinking about the permit_empty rule and I'd like to propose a change. I don't understand why the rule seems to superseded all other rules if it's in there? Rules are supposed to cascade from one rule to the next and I think it would be helpful if permit_empty functioned the same way. For example, I have a form that requires an email address OR phone number OR both. The validation rules are as follows:

public $recipient = [
'first_name' => 'required|min_length[1]|max_length[50]',
'last_name' => 'required|min_length[1]|max_length[50]',
'email' => 'required_without[number]|permit_empty|valid_email',
'number'	=> 'required_without[email]|permit_empty|numeric'
];

I'd like the rule set to check if email OR number exists and then if one does permit the other to be empty. permit_empty seems to hijack the whole thing. Why can't it just cascade like all the other rules?

Please put some thought into this.

Thank you,
Kyle

lonnieezell:
> I'd like the rule set to check if email OR number exists and then if one does permit the other to be empty. permit_empty seems to hijack the whole thing. Why can't it just cascade like all the other rules?

> Please put some thought into this.

Because of the way required works currently, but it wouldn't hurt for that all to have a revisit. I did that library pretty early on and it's showing some cracks. Anyone want to take a crack at it? :)

atishamte:
There is confusion in the method definition of required_without.
It says, "The field is required when all of the other fields are not present in the data."

I guess the definition should be like,
"The field is required when all of the other fields are present in the data but not required(means, the value can be blank or empty)."

lonnieezell:
Ah, that makes sense then. Yes, it should be modified to only use non-empty fields to check against so that it's functionality matches required_with.

system/Validation/Rules.php
```diff
@@ -364,10 +364,10 @@ public function required_without($str = null, string $fields, array $data): bool
 		}
 
 		// Still here? Then we fail this test if
-		// any of the fields are not present in $data
+		// any of the fields are not present or value is '' in $data
 		foreach ($fields as $field)
 		{
-			if (! array_key_exists($field, $data))
+			if (! array_key_exists($field, $data) || ! $this->required($data[$field] ?? ''))
```
atishamte: This change is not required as we just intend to check here whether the field exists in the data or not. It doesn't matter whether the field is required or not.

atishamte:
I guess we can close this PR as the definitions are changed in the #1862.

PR #1755, opened 2019-02-23
lowercase 'vfsStream' in composer files

MGatner:
**Description**
Composer gives warnings about uppercase letters in paths. It looks like this one was missed at some point - fixing across 3 package locations.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/admin/framework/composer.json
+++ b/admin/framework/composer.json
@@ -14,7 +14,7 @@
     },
     "require-dev": {
         "codeigniter4/codeigniter4-standard": "^1.0",
-        "mikey179/vfsStream": "1.6.*",
+        "mikey179/vfsstream": "1.6.*",
         "phpunit/phpunit": "^7.0",
         "squizlabs/php_codesniffer": "^3.3"
     },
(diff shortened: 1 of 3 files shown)


jim-parry:
Looks good, thanks.
For future reference, commits need to be GPG-signed. You checked that box, but did not securely sign your commit:-/

MGatner:
Yes sorry about that, all my previous ones were and I’m confused why this one wasn’t. I was just researching and found `git config commit.gpgsign true` - I’ll give it a go for next time. Thanks for the fast response!

jim-parry:
just setting the git config isn't enough ... the signing keys need to be registered with your github account, and yout local git user.email needs to match that on your github account too

MGatner:
Yes I have both of those in place already which is why I was confused that it didn't show up as signed - I think this setting was the missing piece. Will test now.

MGatner:
Okay you're right that was not as straightforward as I thought! Got it all set though:
https://github.com/MGatner/CodeIgniter4/commit/961ca68869f840e70af531a370f0dee906e3d7d4
I'll be sure to be verified next time. Cheers!

PR #1762, opened 2019-02-26
decimal rule. shouldn't it accept integers?

nowackipawel:
decimal rule
integers are subset of decimals / floats. so i think  decimal validation rule should pass if value contains decimal OR integer.

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -128,7 +128,7 @@ class FormatRules
 	 */
 	public function decimal(string $str = null): bool
 	{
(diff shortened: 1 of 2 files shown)


lonnieezell:
I'm cool with this change, but it caused a test to fail.

nowackipawel:
done.

atishamte:
I guess It shouldn't pass the test case for integer number as there is separate integer number validator exist. It will create an issue at the time of string integer or decimal matching

nowackipawel:
As long as integers/naturals are subset of decimals/floats I think this test should pass for ints.

nowackipawel:
@lonnieezell : is it ready to merge?

PR #1769, opened 2019-02-28
Correction in Methods and Spellings

atishamte:
Some corrections in,
- Method return type
- Namespace
- Return Type
- Spellings
- Logics

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -106,7 +106,7 @@ class Controller
 	 * @param ResponseInterface        $response
 	 * @param \Psr\Log\LoggerInterface $logger
 	 *
-	 * @throws \CodeIgniter\HTTP\RedirectException
+	 * @throws \CodeIgniter\HTTP\Exceptions\HTTPException
 	 */
 	public function initController(RequestInterface $request, ResponseInterface $response, LoggerInterface $logger)
 	{
@@ -135,7 +135,7 @@ class Controller
 	 *                          considered secure for. Only with HSTS header.
 	 *                          Default value is 1 year.
 	 *
-	 * @throws \CodeIgniter\HTTP\RedirectException
+	 * @throws \CodeIgniter\HTTP\Exceptions\HTTPException
 	 */
 	public function forceHTTPS(int $duration = 31536000)
 	{
@@ -179,14 +179,14 @@ class Controller
 	 * A shortcut to performing validation on input data. If validation
 	 * is not successful, a $errors property will be set on this class.
 	 *
-	 * @param array $rules
-	 * @param array $messages An array of custom error messages
+	 * @param array|string  $rules
+	 * @param array         $messages An array of custom error messages
 	 *
 	 * @return boolean
 	 */
 	public function validate($rules, array $messages = []): bool
 	{
-		$this->validator = Services::validation();
+		$this->validator = \Config\Services::validation();
 
 		// If you replace the $rules array with the name of the group
 		if (is_string($rules))
@@ -211,9 +211,8 @@ class Controller
 		}
 
 		$success = $this->validator
-			->withRequest($this->request)
 			->setRules($rules, $messages)
-			->run();
+			->run($this->request->getPost());
 
 		return $success;
 	}
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -35,7 +35,7 @@
  * @since      Version 3.0.0
  * @filesource
  */
-
+use CodeIgniter\Config\Services;
 use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
 use CodeIgniter\Validation\Validation;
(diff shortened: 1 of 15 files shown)


system/Controller.php
```diff
@@ -179,14 +179,14 @@ protected function loadHelpers()
...
-	 * @param array $messages An array of custom error messages
+	 * @param array|string  $rules
+	 * @param array         $messages An array of custom error messages
 	 *
 	 * @return boolean
 	 */
 	public function validate($rules, array $messages = []): bool
 	{
-		$this->validator = Services::validation();
+		$this->validator = \Config\Services::validation();
```
lonnieezell: Would prefer this was added at the top of the as a use statement. Not sure why it wasn't already.

system/Controller.php
```diff
@@ -211,9 +211,8 @@ public function validate($rules, array $messages = []): bool
 		}
 
 		$success = $this->validator
-			->withRequest($this->request)
 			->setRules($rules, $messages)
-			->run();
+			->run($this->request->getPost());
```
lonnieezell: Were you running into an issue with the withRequest() method? If not, no need to change it.

system/Entity.php
```diff
@@ -566,7 +566,7 @@ private function castAsJson($value, bool $asArray = false)
 
 				if (json_last_error() !== JSON_ERROR_NONE)
 				{
-					throw CastException::forInvalidJsonFormatException(json_last_error());
+					throw \CodeIgniter\Exceptions\CastException::forInvalidJsonFormatException(json_last_error());
```
lonnieezell: Add as a use statement at the top of the file, please.

system/HTTP/IncomingRequest.php
```diff
@@ -152,14 +152,19 @@ class IncomingRequest extends Request
 	 * @param string                      $body
 	 * @param \CodeIgniter\HTTP\UserAgent $userAgent
 	 */
-	public function __construct($config, $uri = null, $body = 'php://input', UserAgent $userAgent)
+	public function __construct($config, $uri = null, $body = 'php://input', UserAgent $userAgent = null)
```
lonnieezell: prefer to require UserAgent in this request. It it already added in the service method, so revert this change.

system/Validation/FormatRules.php
```diff
@@ -196,7 +196,7 @@ public function numeric(string $str = null): bool
 	 *
 	 * @return boolean
 	 */
-	public function regex_match(string $str = null, string $pattern, array $data): bool
+	public function regex_match(string $str, string $pattern, array $data): bool
```
lonnieezell: As with all of the other validation rules, this should accept a null value.

system/Controller.php
atishamte: the run method accepts the $data as a parameter. It was showing the syntactical error if run() is written. Hence changed.

system/HTTP/IncomingRequest.php
atishamte: Changed it because there is a concept, always keep optional parameter the end of parameter listing in function arguments. Hence changed.

system/Controller.php
atishamte: This is an interface function prototype
public function run(array $data, string $group = null): bool;
where $data is not optional

system/HTTP/IncomingRequest.php
lonnieezell: I'm familiar with the concept. In this case, it doesn't work, though. Due to the different types of rules, and some wanting/needing more information than others, the first item should always be the $str. And sometimes that value will be null.

system/Controller.php
lonnieezell: Then the correct solution here is to make $data nullable in the Validator's run method. And adjust tests accordingly. 

it's always possible for PUT or other HTTP verbs than POST to provide data to the application.

atishamte: As per the documentation, I have made the changes. It is written to validate only POST params.
Please refer the attachment and link

https://codeigniter4.github.io/userguide/incoming/controllers.html#validating-post-data

controllers codeigniter4 4 0 0 alpha 5 documentation

Do you want me to make changes for other HTTP verbs?

lonnieezell: Yes, that should be updated. The `withRequest` method handles any of the verbs, which is why that was used in the controller. Looks like I forgot to update the docs. 

Also - if you could break this out into it's own PR that would be great. Simpler to deal with things when it's only one change/type of change in a PR.

atishamte: Sure. I will make a separate PR for the same, where run method $data param is nullable.
I will update the doc as well.

atishamte:
When It will get merged? I am unable to generate new PR

jim-parry:
Patience, young padawan!
You should be creating PR features in separate branches, rather than in your develop branch. Then you would be able to have multiple changes in progress at the same time.

atishamte:
Mistakenly I pushed it to develop, that's why I am asking.
Thanks

atishamte:
All the changes have been done as per @lonnieezell suggested

PR #1776, opened 2019-03-01
test - replace only last operator in field name.

nowackipawel:
want to know what tests this will break.

If all tests will pass I will write what is better in this approach.

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -673,7 +673,7 @@ class BaseBuilder
 
 					if(substr($k,  -1 * strlen($op)) === $op)
 					{
-						$k = strrev(preg_replace(strrev("/" .  $op . "/"),strrev(''),strrev($k),1));
+						$k = rtrim(strrev(preg_replace(strrev("/" .  $op . "/"),strrev(''),strrev($k),1)));
 					}
 
 				}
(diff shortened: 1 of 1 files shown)


lonnieezell:
You know - `composer install` is all it takes to get PHPUnit up and running on your local machine, right?

nowackipawel:
Thank you for letting me know :).

nowackipawel:
So my PR allows us to make sth like:
`(new \ExampleModel)->where('(SELECT COUNT(*) FROM t1 WHERE t1_exa_id = ex_id)',10);`
or its variants like:
`(new \ExampleModel)->where('(SELECT COUNT(*) FROM t1 WHERE t1_exa_id = ex_id) >= ',10);`
`(new \ExampleModel)->where('(SELECT COUNT(*) FROM t1 WHERE t1_exa_id = ex_id)',10);`

what was impossible before.

nowackipawel:
will update code when my solution will work on sth like this:
```
(new \ExampleModel)
->startGroup()
->where('(SELECT COUNT(*) FROM t1 WHERE t1_exa_id = ex_id)','')
->orWhere('(SELECT COUNT(*) FROM t1 WHERE t1_exa_id = ex_id) IS NULL', NULL, false)
->endGroup()

->startGroup()
->where('field_1','')
->orWhere('field_1 IS NULL', NULL, false)
->endGroup()
;
```

PR #1791, opened 2019-03-04
replace mb_strlen() to strlen() in various rules as already return early in ! is_numeric

samsonasik:
Previously, there is a `! is_numeric()` check with return false early, so the `numeric` check is no need mbstring check with `mb_strlen()`, the `strlen()` is enough.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -79,7 +79,7 @@ class Rules
 			return false;
 		}
 
-		return ((int) $val === mb_strlen($str));
+		return ((int) $val === strlen($str));
 	}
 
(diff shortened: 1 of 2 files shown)


lonnieezell:
Actually - I have no idea why it's checking to make sure it's a number first. This should probably be to remove the `! is_numeric` check instead, otherwise we can never actually check the length of the string. At that point, though, we would need to keep `mb_strlen`.

samsonasik:
@lonnieezell Ok, I've removed the `!is_numeric()` check and keep the `mb_strlen()`

PR #1806, opened 2019-03-10
datetime __set support + PHPdoc's

nowackipawel:


nowackipawel:
@lonnieezell  @jim-parry  what do you think about broken test?

IMHO if our app is configured in America/Chicago time, and we are getting  value like '2019-03-12 00:00:00' from database (which is stored as UTC) then after casting to the appTimezone (which is i.e. America/Chicago) we should get sth. like '2019-03-11 19:00:00' . Am I wrong?

```
https://www.timeanddate.com/worldclock/converted.html?p1=64&p2=0
Chicago (USA - Illinois) Monday, 11 March 2019, 04:59:39 CDT UTC-5 hours 
UTC (Time Zone)          Monday, 11 March 2019, 09:59:39 UTC UTC  
```

lonnieezell:
I think it's a good idea for an Entity to set the timezone. A couple of things need to happen to do that well, though. The main one being that there needs to be a timezone class property or some other solution to allow things to be overrode to the current user's value. That's one reason I didn't do that originally, is because i wanted the Entity to be a bit minimal as a first time solution. Then, as CI evolved, we could see how more people used it. 

Even if all of that was implemented well, you've stopped returning a Time instance, which was the whole point of that method, so that's a no-go.

nowackipawel:
@lonnieezell: What about introducing new config option in the App or somewhere which allows user to set lightweight simple inner class solution or allows them to use a sledgehammer to crack a nut with \I18n\Time? 
IMHO it will be win-win solution - simple and nice or just complex.

PR #1657, opened 2019-01-08
Methods for (un)setting / overriding  $validationMessages

nowackipawel:
Instead of doing this if I need to specify custom validationMessage:

```

			$user_model = (new \UserModel());

			$validation->setRules(
				$user_model->getValidationRules(),
				[
					'usr_email' => [
						'is_unique' => 'Twój adres e-mail jest już w naszej bazie. Sprónuj zalogować tak samo jak ostatnim razem, a jeśli Ci sie nie uda skorzystaj z funkcji przypominania hasła.'
					]
				]
			)->run($data, null, $user_model->DBGroup);

			if(!empty($errors = $validation->getErrors()))
			{
				helper('messages');
				set_message(B4_ALERT_DANGER, '<ul><li>' .  implode('</li><li>', $validation->getErrors()) . '</li></ul>');
			}
			else
			{

				$user_model->insert($data);
return redirect()->to(route_to('user_profile'));
			}
```
I could do this:
```

			$user_model = (new \UserModel());
			$user_model->setValidationMessage(
				'usr_email',  
				[
					'is_unique' => 'Twój adres e-mail jest już w naszej bazie. Sprónuj zalogować tak samo jak ostatnim razem, a jeśli Ci sie nie uda skorzystaj z funkcji przypominania hasła.'
				]);
			$user_model->insert($data);
			
			if(!empty($errors = $user_model->errors()))
			{
				helper('messages');
				set_message(B4_ALERT_DANGER, '<ul><li>' .  implode('</li><li>', $errors) . '</li></ul>');
			}
			else
			{
				$user_model->insert($data);
				return redirect()->to(route_to('user_profile'));
			}

```

which looks a  nicer , and  validation is calling only once.

PR #1810, opened 2019-03-11
Model setValidationMessage functions introduced

atishamte:
W.r.t #1657, I have added functionality in Model,
- Added functions for setValidationMessage & setValidationMessages
- Added respective test scenarios
- Added user guide documentation

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1244,6 +1244,35 @@ class Model
 	//--------------------------------------------------------------------
 
 	/**
+	 * Allows to set new validation messages.
+	 * It could be used when you have to change default or override current validate messages.
+	 *
+	 * @param array $validationMessages: new set of validation messages
+	 *
+	 * @return void
+	 */
+	public function setValidationMessages(array $validationMessages)
+	{
+		$this->validationMessages = $validationMessages;
+	}
+	//--------------------------------------------------------------------
+
+	/**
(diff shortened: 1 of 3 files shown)


atishamte:
Test cases failed due to some other functionality committed earlier in Model.
What has to be done now, as the current PR is not relevant to the failed test case scenario

jim-parry:
It looks like your PR 1804 missed changing the return type for Model::insert, and that the Model testing didn't catch it. That need to be fixed here or in a separate PR. 
This points out the need to finish the Model unit testing sooner than later :-/

jim-parry:
You earlier changed the return type of save/updatge etc to boolean, BUT insert returns a boolean or an int, and that is passed into save, which then breaks.
We need to figure how to solve this, or else revert 1804 :(

jim-parry:
I have fixed the cause of the problems. You need to synch your branch with the main repo's develop branch, and then repush your branch. The travis-ci build will restart automatically.

jim-parry:
ps - I have checked out phpDocumentor2, and it is very usable. I will be incorporating the generated API as part of the project docs :)

atishamte:
That's great. So the documentation will get generated automatically based on class and function documentation directly :)

Issue #1807, opened 2019-03-11
\I18n\Time setTimezone does not work as expected


nowackipawel:
I looked into \Time code and.... it does not look good at all.

Additionally setTimezone does not work as excepted .

```

			$t0 = new \CodeIgniter\I18n\Time($transaction->tra_created_at);
			$t1 = $t0->setTimezone('America/Los_Angeles');
			d($t0, $t1);

```
https://codeigniter4.github.io/CodeIgniter4/libraries/time.html?highlight=settimezone

gives me:

https://screenshots.firefox.com/QrGBqc9Is0QPTGKo/dev.zdamyto.com
my $appTimezone ===  'UTC'

I tried to fix it, but this code looks completely messy.

lonnieezell:
Any particular concerns, or just a general thumbing your nose at the code?

lonnieezell:
What exactly were you trying to fix? All times are converted internally to UTC to handle any conversions, and, IIRC, might have been how the DateTime instance that it's extending handles things internally. It's been a while since I've looked at it. 

If you print the time out does it give the correct timezone? The tests seem to indicate that it's working as expected.

lonnieezell:
With the current version of the code, here's what I get: 

```
$t0 = new Time('now', 'UTC');
$t1 = $t0->setTimezone(new \DateTimeZone('America/Chicago'));
d($t0, $t1);

// Gives: 
$t0 CodeIgniter\I18n\Time (6) 2019-03-11 03:00:40.329165+00:00 UTC
$t1 CodeIgniter\I18n\Time (6) 2019-03-11 03:00:40-05:00 CDT
```

and setting $appTimezone to UTC I get: 

```
$t0 = new Time('2016-02-23 12:14:23');
$t1 = $t0->setTimezone(new \DateTimeZone('America/Chicago'));

$t0 CodeIgniter\I18n\Time (6) 2016-02-23 12:14:23+00:00 UTC
$t1 CodeIgniter\I18n\Time (6) 2016-02-23 12:14:23-06:00 CST
```

Both of which act as expected. 

Is your code up to date? Though we haven't made changes to that in quite a while, I don't think. If it is reasonably up to date then what environment is that running on? OS? Server? PHP version? Can you verify what the system thinks `date_default_timezone_get` on your environment when you're trying to run that?

nowackipawel:
> Any particular concerns, or just a general thumbing your nose at the code?

[...]

PR #1815, opened 2019-03-12
Avibility to change mutateDate behaviour

nowackipawel:
Adds possibility to set mutateDate  to return \DateTime instead of \CodeIgniter\I18n\Time instances.
To make it work simpy use this code in your Entity child constructor:
```
	public function __construct(?array $data = null)
	{
		parent::__construct($data, ['dateMutateAs' => 'DateTime']);
	}
```
or
```
	public function __construct(?array $data = null)
	{
		parent::__construct($data);
		$this->setMutateDateAs('DateTime'); // will return boolean - value changed === true.
	}
```

I think it is required to add some code to make  timezone changes work for \CodeIgniter\I18n\Time  instance. For make this work in predictable way  I think we should make some changes in \CodeIgniter\I18n\Time (#1807)

btw/ didn't know why git makes this commit diff so ugly :/

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -1,8 +1,5 @@
 <?php namespace CodeIgniter;
 
-use CodeIgniter\I18n\Time;
-use CodeIgniter\Exceptions\CastException;
-
 /**
  * CodeIgniter
  *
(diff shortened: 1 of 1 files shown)


lonnieezell:
No thanks.

All this does is add additional complexity that has to be maintained over the life of the framework. It's easy enough for someone to extend the Entity class (or replace it completely - it's not that complex of a class) and override the mutateDate functionality if they want something different.

PR #1816, opened 2019-03-12
Pager: add classes required for Bootstrap rendering

MGatner:
**Description**
The user guide claims "The Pager class will render a series of links that are compatible with the Bootstrap CSS framework by default" but by default the resulting HTML is missing the necessary classes to be rendered as a Bootstrap pagination.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Pager/Views/default_full.php
+++ b/system/Pager/Views/default_full.php
@@ -9,34 +9,34 @@ $pager->setSurroundCount(2);
 <nav aria-label="<?= lang('Pager.pageNavigation') ?>">
 	<ul class="pagination">
 		<?php if ($pager->hasPrevious()) : ?>
-			<li>
-				<a href="<?= $pager->getFirst() ?>" aria-label="<?= lang('Pager.first') ?>">
+			<li class="page-item">
+				<a class="page-link" href="<?= $pager->getFirst() ?>" aria-label="<?= lang('Pager.first') ?>">
 					<span aria-hidden="true"><?= lang('Pager.first') ?></span>
 				</a>
 			</li>
-			<li>
-				<a href="<?= $pager->getPrevious() ?>" aria-label="<?= lang('Pager.previous') ?>">
+			<li class="page-item">
+				<a class="page-link" href="<?= $pager->getPrevious() ?>" aria-label="<?= lang('Pager.previous') ?>">
 					<span aria-hidden="true">&laquo;</span>
 				</a>
 			</li>
 		<?php endif ?>
 
 		<?php foreach ($pager->links() as $link) : ?>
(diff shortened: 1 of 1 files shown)


jim-parry:
The user guide is poorly worded. We don't want to add CSS framework attributes.
Way back when, Bootstrap prided itself on styling most HTML elements without additional attributes, and my guess is that is where the original user guide wording came from. I remember using unordered lists for navbars, simply wrapping them in a suitable bootstrap-styled component. That could also be source of the original "claim".
Better is to remove any mention of Bootstrap compatibility, which is misleading.

jim-parry:
This looks like a can of worms! There are several nav styles or pagination support built into Bootstrap, slightly different per bootstrap version - default, links, pills, pager vs plain nav. 

The default_simple and default_head pagination templates seem CSS-framework-free, but not so default_full. The only other place in the user guide which mentions Bootstrap is the validation library. Looking at its views, it adds a couple of CSS attributes to error message divs.

lonnieezell:
Yes, that was originally for Bootstrap 3. Looks like 4 took a step backward in that respect. Too bad.

jim-parry:
So, should the bootstrap styling be removed? 
if we leave it in there, will the community expect it everywhere?
... not to mention "what about bootstrap X or each's favorite CSS framework" :(

MGatner:
They need some classes so users can apply CSS to the out-of-the-box HTML. That said, the “pagination” class on the wrapping NAV ought to be minimum necessary. It is indeed unfortunate that this is no longer enough for Bootstrap, but applying their specific classes to the items definitely seems like a decision to “go with Bootstrap”. 
The alternative is also unfortunate though - requiring a user to create a custom pagination view just because they want classes added to otherwise acceptable HTML. Since (presumably) one would want pagination classes consistent across an app, what about adding some class options to Pager.php, especially since it already exists? That was the first place I went and looked when I realized the items were missing classes.

lonnieezell:
The easy solution for people asking for support of X is to simply say, we only support Y because it seems to be the most widely used and makes a good default. 

However, in this case I say we leave it like it is so that it can be styled without modification, and remove the comment from the docs.

MGatner:
I will send a PR for the user docs.

MGatner:
PR sent. FYI the only remaining reference to Bootstrap styling int he User Docs is in the Validation Library (https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html), the views called by `$validation->listErrors()` and `$validation->showError()`"display in a manner compatible with the Bootstrap CSS framework".

lonnieezell:
I've removed the Bootstrap reference and changed the style to be a bit more generic.

PR #1819, opened 2019-03-12
Add apibot for API docs using phpDocumentor

jim-parry:
New scripts: apibot, alldocs
Tied in to new repo: codeigniter4/api


jim-parry:
@atishamte If you could keep PRs to one module at a time (eg API or Autoloader), that would make it easier to review, thanks.

atishamte:
Ok

PR #1831, opened 2019-03-14
Fix some PHPDoc comments error

vibbow:
Fix some PHPDoc comments error

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


jim-parry:
Good catch. Not sure how the use CodeIgntier\Services made it through unit testing before :-/

PR #1839, opened 2019-03-14
Dont replace slashes with backslashes in the to route

### LINKED ISSUE #1838: Routing to hardcoded ids does not work
**Describe the bug**
The router should allow me to add routes in the following way:
```$routes->get( 'abcd/efgh', 'Pages::id/123' )```
However, in the default Router (Codeigniter\Router), all the `/` in the`$to` parameter get overwritten by `\`. This breaks the functionality of the route since it starts looking at the `Pages` controller for the function `id\123` (which does not exist).

Your own guidelines even have an example that has this kind of route in it.
```$routes->add('blog/joe', 'Blogs::users/34');```
 I have not actually tested this, but I guess that it fails for the same reasons as why my route is failing.

As far as I can tell, the way I want to route corresponds roughly with the routing described in the tutorial.

This problem seems to occur to the following lines of code in the `checkRoutes( string $url )` function (in the Router, lines 471 to 474).
```
elseif (strpos($val, '/') !== false)
	{
		$val = str_replace('/', '\\', $val);
	}
```
These lines translate all the `/` in `$val` to `\`. I don't know why this is done, but commenting those four lines fixes the problem for me (although I don't know what other consequences there are...). 

**CodeIgniter 4 version**
Which version (and branch, if applicable) the bug is in.
The first Codeigniter beta version.

I expect that my routes are routed to the controller with the `123` passed as the `$id` parameter.

**Context**
 - OS: [MacOS]
- Using `php spark serve`
 - PHP version [7.3]

bintzandt:
Fixes #1838 

**Description**
Remove the replacement of `/` with `\` in the `$to` part of routes. This ensures that routes like `$routes->add('blog/joe', 'Blogs::users/34');` will work.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -469,11 +469,7 @@ class Router implements RouterInterface
 				{
 					$val = preg_replace('#^' . $key . '$#', $val, $uri);
 				}
-				elseif (strpos($val, '/') !== false)
-				{
-					$val = str_replace('/', '\\', $val);
-				}
-
+			
 				// Is this route supposed to redirect to another?
 				if ($this->collection->isRedirect($key))
 				{
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -469,7 +469,16 @@ class Router implements RouterInterface
 				{
 					$val = preg_replace('#^' . $key . '$#', $val, $uri);
 				}
-			
+				elseif (strpos($val, '/') !== false)
+				{
+					[ $controller, $method ] = explode( '::', $val );
+					
+					// Only replace slashes in the controller, not in the method.
+					$controller = str_replace('/', '\\', $controller);
+					
+					$val = $controller . '::' . $method;
+				}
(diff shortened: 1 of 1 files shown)


lonnieezell:
That might fix that specific problem, but makes it impossible to find controllers in subdirectories and map them to a namespace. 

It looks like we need to split the $to segment on the double-colons and do that conversion on the left side only. That way it leaves things alone and everything should still work.

bintzandt:
Yeah, I ran into that already.

I have updated the PR so that the slashes are only replaced in the controller part of `$val`. I have used array destructuring, which is only supported in PHP >= 7.1 but I don't think that is a problem since Codeigniter 4 only supports PHP >= 7.2.

This should fix the problem and still allow Codeigniter to find controllers in subdirectories.

Let me know!

PR #1840, opened 2019-03-18
select - option disabled

nowackipawel:
added as last parameter to be sure it won't break anything and to be compatible with current use cases.

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -340,10 +340,11 @@ if (! function_exists('form_dropdown'))
 	 * @param mixed $options
 	 * @param mixed $selected
 	 * @param mixed $extra
-	 *
+	 * @param array $disabled
+	 * 
 	 * @return string
 	 */
-	function form_dropdown($data = '', $options = [], $selected = [], $extra = ''): string
+	function form_dropdown($data = '', $options = [], $selected = [], $extra = '', $disabled = []): string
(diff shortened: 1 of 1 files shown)


lonnieezell:
That would make that the only helper to support a disabled flag. All you have to do is add `disabled` to the `$extra` argument and you're good to go.

nowackipawel:
Hi @loonieezell .
Yes dropdown will be exception , but its semantic/characteristic is different then other fields.  Please correct me if I'm wrong but I am not able to add disabled to particular options for select field with $extra parameter. 
I don't want to make select disabled but only part of the options.

lonnieezell:
That's fair. If you can update tests and provide docs I'm fine.

nowackipawel:
@lonnieezell I'm good with docs but I don't know when I will have time to write tests :/. What do you think? This change can't break anything.

lonnieezell:
> This change can't break anything.

Famous last words. :) The point of tests is also to make sure the code works correctly, which there's no proof of, either. And so that if someone makes a change down the line they can have a higher confidence all of its functionality still works as expected. 

> but I don't know when I will have time to write tests

Honestly, me neither. I'm swamped with the day job at the moment.

nowackipawel:
Should I open the issue about this or we could merge it without tests? :-) .  This feature is working good in my code.

PR #1844, opened 2019-03-19
Model Fix

atishamte:
Model Fix

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -482,7 +482,7 @@ class Forge
 
 		if (($result = $this->db->query($sql)) !== false)
 		{
-			empty($this->db->dataCache['table_names']) || $this->db->dataCache['table_names'][] = $table;
+			empty($this->db->dataCache['table_names']) OR $this->db->dataCache['table_names'][] = $table;
 
 			// Most databases don't support creating indexes from within the CREATE TABLE statement
 			if (! empty($this->keys))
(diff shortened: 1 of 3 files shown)


atishamte:
@jim-parry or @lonnieezell, Please merge this to develop, so build will not fail.

system/Database/Forge.php
```diff
@@ -482,7 +482,7 @@ public function createTable($table, $if_not_exists = false, array $attributes =
 
 		if (($result = $this->db->query($sql)) !== false)
 		{
-			empty($this->db->dataCache['table_names']) || $this->db->dataCache['table_names'][] = $table;
+			empty($this->db->dataCache['table_names']) OR $this->db->dataCache['table_names'][] = $table;
```
lonnieezell: Was this change actually required? Typically we keep to the bitwise || unless needed.

atishamte: If I keep it || then editor gives an error "Cannot use [] for reading". So changed

atishamte: I went through the code flow. The flow should be like if table_names is empty then assign table name to it.
But I run the dummy code,
$a = [];
empty($a) OR $a[] = 5;
print_r($a);

$a = [];
empty($a) || $a[] = 5;
print_r($a);

Both are giving me same output as blank array.

But the code should be,
if(empty($this->db->dataCache['table_names']))
{
        $this->db->dataCache['table_names'][] = $table;
}

Correct me If I am wrong.

lonnieezell: I think you're correct here. This is code ported from CI3 that I hadn't really touched much before. Not a fan of the original coding style there, honestly, but your fix looks correct.

atishamte: Should I generate a PR for correction of this condition like I stated in earlier comment

lonnieezell: Yes, that would be great. Thanks.

PR #1847, opened 2019-03-20
Extend Controller to BaseController by default

MGatner:
**Description**
Create a default extended controller at app/Controllers/BaseController.php and updates Home controller and the User Guide to recommend using this for new Controllers. From a development standpoint, it makes sense to start with this extension even if it doesn't have any modifications since making the change down the road involves going back and editing every controller.
There was a fair amount of discussion on this idea in the forums already. See: https://forum.codeigniter.com/thread-73095.html

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Config/Routes.php
+++ b/system/Config/Routes.php
@@ -53,7 +53,7 @@ $routes->cli('migrations', '\CodeIgniter\Commands\MigrationsCommand::index');
 $routes->cli('ci(:any)', '\CodeIgniter\CLI\CommandRunner::index/$1');
 
 // Prevent access to BaseController
-$routes->add('basecontroller/(:any)', function()
+$routes->add('basecontroller(:any)', function()
 {
     throw \CodeIgniter\Exceptions\PageNotFoundException::forPageNotFound();
 }); 
--- a/system/Config/Routes.php
+++ b/system/Config/Routes.php
@@ -49,5 +49,11 @@ $routes->cli('migrations/(:segment)/(:segment)', '\CodeIgniter\Commands\Migratio
 $routes->cli('migrations/(:segment)', '\CodeIgniter\Commands\MigrationsCommand::$1');
 $routes->cli('migrations', '\CodeIgniter\Commands\MigrationsCommand::index');
 
-// CLI Catchall - uses a _remap to
+// CLI Catchall - uses a _remap to call Commands
 $routes->cli('ci(:any)', '\CodeIgniter\CLI\CommandRunner::index/$1');
+
+// Prevent access to BaseController
+$routes->add('basecontroller/(:any)', function()
+{
+    throw \CodeIgniter\Exceptions\PageNotFoundException::forPageNotFound();
+}); 
--- a/system/Config/Routes.php
+++ b/system/Config/Routes.php
@@ -44,6 +44,12 @@
  * It is called by Config\Routes, and has the $routes RouteCollection
  * already loaded up and ready for us to use.
  */
+// Prevent access to BaseController
+$routes->add('basecontroller(:any)', function()
+{
+    throw \CodeIgniter\Exceptions\PageNotFoundException::forPageNotFound();
+}); 
+
 // Migrations
 $routes->cli('migrations/(:segment)/(:segment)', '\CodeIgniter\Commands\MigrationsCommand::$1/$2');
 $routes->cli('migrations/(:segment)', '\CodeIgniter\Commands\MigrationsCommand::$1');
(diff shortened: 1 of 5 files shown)


system/Config/Routes.php
```diff
@@ -49,5 +49,11 @@
 $routes->cli('migrations/(:segment)', '\CodeIgniter\Commands\MigrationsCommand::$1');
 $routes->cli('migrations', '\CodeIgniter\Commands\MigrationsCommand::index');
 
-// CLI Catchall - uses a _remap to
+// CLI Catchall - uses a _remap to call Commands
 $routes->cli('ci(:any)', '\CodeIgniter\CLI\CommandRunner::index/$1');
+
+// Prevent access to BaseController
+$routes->add('basecontroller(:any)', function()
```
jim-parry: Shouldn't this be basecontroller/(:any)? 
It also seems that it should be in front of the CLI routing

MGatner: When I tested it if I included the slash then "https://example.com/basecontroller" was still a routable address - making it basically "basecontroller*" fixes that and would also cover "BaseController2", etc, if users added more.
As to whether it should appear before or after CLI that I don't know! There were no routes in the system config except the CLI ones so I wasn't sure where to put it.

jim-parry:
(:any) is normally a segment placeholder, not "part" of a segment.
Having said that, the "$routes->cli('ci(:any)',..." seems very odd, and makes no sense to me. WHy would one map ci1 or ci4 to a CommandRunner index method? I could be misinterpreting.
My comment about the slash in the basecontroller rule still stands, and I suugested it be moved above the cli rules in case one of them kicked in before it, sending the request down a deep, dark path that we don't want :-/

I think your rule will fire a 404 for /basecontroller, but not for /basecontroller/anythingelse.

MGatner:
I've tested and it definitely matches as I have said: /basecontroller, /basecontroller/, basecontroller/test, etc.
I will move the rule above CLI - let me know if you want me to changes anything else with it.

jim-parry:
Good to know about the (:any) - I learned something new :)
We still need to determine why the CommandsTest::testRoutesCommand() is failing, across the board.
The base controller and the additional routing rule are the only two things I see different, and I don't offhand see a reason why they would break that test.

MGatner:
Can you tell if it is testing with https://github.com/codeigniter4/CodeIgniter4/pull/1842 in place? If not then `forPageNotFound()` without a parameter will throw a syntax error.

jim-parry:
1842 has been merged, so will be part of the tests.

lonnieezell:
> (:any) is normally a segment placeholder, not "part" of a segment.

Yup. In CI4 `:any` was changed to mean, literally, anything. `:segment` is available if you want just a single segment.

MGatner:
The test is failing because the CLI command `routes` tries to parse the callback as a string. I sent PR https://github.com/codeigniter4/CodeIgniter4/pull/1859 for that command to filter on strings only, though at some point someone might want to figure out a way to have it display something helpful when a function is detected. I initially had it noting the callback but it made for a very messy output in a vanilla setup with no user-defined routes:
```
+----------------------------+---------+------------------------------------------------+
| Route                      | Method  | Command                                        |
+----------------------------+---------+------------------------------------------------+
| /                          | get     | {callback}                                     |
| basecontroller(.*)         | get     | {callback}                                     |
| (.*)/initController        | get     | {callback}                                     |
| basecontroller(.*)         | head    | {callback}                                     |
| (.*)/initController        | head    | {callback}                                     |
| basecontroller(.*)         | post    | {callback}                                     |
| (.*)/initController        | post    | {callback}                                     |
| basecontroller(.*)         | put     | {callback}                                     |
| (.*)/initController        | put     | {callback}                                     |
| basecontroller(.*)         | delete  | {callback}                                     |
| (.*)/initController        | delete  | {callback}                                     |
| basecontroller(.*)         | options | {callback}                                     |
| (.*)/initController        | options | {callback}                                     |
| basecontroller(.*)         | trace   | {callback}                                     |
| (.*)/initController        | trace   | {callback}                                     |
| basecontroller(.*)         | connect | {callback}                                     |
| (.*)/initController        | connect | {callback}                                     |
| migrations/([^/]+)/([^/]+) | cli     | \CodeIgniter\Commands\MigrationsCommand::$1/$2 |
| migrations/([^/]+)         | cli     | \CodeIgniter\Commands\MigrationsCommand::$1    |
| migrations                 | cli     | \CodeIgniter\Commands\MigrationsCommand::index |
| ci(.*)                     | cli     | \CodeIgniter\CLI\CommandRunner::index/$1       |
| basecontroller(.*)         | cli     | {callback}                                     |
| (.*)/initController        | cli     | {callback}                                     |
+----------------------------+---------+------------------------------------------------+
```

MGatner:
Is there a way to rerun the runner now that the other PR is merged?

MGatner:
Is there a way to rerun the runner now that the other PR is merged?

PR #1848, opened 2019-03-20
Test: fix & test Test/ControllerTest, tested

jim-parry:
Minor fixes for DOMParser & ControllerTester
Unit testing for these

DOMParser: 
- blocked two exceptions from code coverage - one because it will never happen in travis-ci, and the other because I don't see a way of inducing it, given how forgiving the DOMDocument is in building a DOM from a string
- fixed the attribute xpath testing

ControllerTester:
- added logger to the mix, since the Controller class expects is
- fixed handling of a controller method returning a string instead of echoing stuff

### DIFF
--- a/tests/system/Test/ControllerTesterTest.php
+++ b/tests/system/Test/ControllerTesterTest.php
@@ -1,0 +1,109 @@
+<?php
+namespace CodeIgniter\Test;
+
+use CodeIgniter\Events\Events;
+use CodeIgniter\Test\Filters\CITestStreamFilter;
+use CodeIgniter\HTTP\Response;
+use Config\App;
+
+class TestCaseTest extends \CIUnitTestCase
+{
+
+	//  protected function tearDown()
+	//  {
+	//      $buffer = ob_clean();
+	//      if (ob_get_level() > 0)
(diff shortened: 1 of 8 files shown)


tests/system/Test/ControllerTesterTest.php
```diff
@@ -0,0 +1,203 @@
...
+
+	public function tearDown()
+	{
+		parent::tearDown();
+	}
+
+	public function testBadController()
+	{
+		$this->expectException(\InvalidArgumentException::class);
+		$logger = new Logger(new LoggerConfig());
```
lonnieezell: Wouldn't using this pollute the user's log files during testing? 

There's a MockLogger that can be used. I believe it's part of the CIUnitTestCase.

jim-parry: Hehe - line 7 of that file ... use Tests\Support\Config\MockLogger as LoggerConfig;
The tests are already using the MockLogger :)

lonnieezell: Doh! Missed that. You were ahead of me. Yay!

PR #1852, opened 2019-03-21
null coalescing changes

atishamte:
As we are isung php version >=7.2, to maintain the consistancy throughout the code, converted ternary to null coalescing.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -245,7 +245,7 @@ class CLI
 		fwrite(STDOUT, $field . $extra_output . ': ');
 
 		// Read the input from keyboard.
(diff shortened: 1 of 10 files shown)


MGatner:
I could be wrong on this, but null coalesce and the ternary shorthand aren't the same thing. For example in the following change:
```
$input = trim(static::input()) ?: $default;
$input = trim(static::input()) ?? $default;
```
`trim()` always returns a string, thus the ?? operator will never return `$default` because the condition will never be explicitly null.

PR #1854, opened 2019-03-22
File::move now returns new file instance for relocated file. Fixes #1782

### LINKED ISSUE #1782: Intention or bug? File::move does not update path
**Describe the bug**
After a call to File::move(...) File::getRealPath() still returns the path of the file's previous location. This does not make sense, I believe, because that location is invalid. Therefore, I tend to call this a bug.

**CodeIgniter 4 version**
CodeIgniter4 Alpha 4

**Affected module(s)**
File

**Expected behaviour, and steps to reproduce if appropriate**

This reproduces the bug(?):
```
log_message('debug',$file->getRealPath());
$file->move($newDirectory,$newName);
log_message('debug',$file->getRealPath()); // expected to be $newDirectory.$newName
```

lonnieezell:


atishamte:
Change the return type of move() method in comment block

PR #1856, opened 2019-03-22
ensure ob_end_flush() when ob_get_level() > 0 on pre_system event

samsonasik:
so the `ENVIRONMENT !== 'testing'` check no longer needed.

**Checklist:**
- [x] Securely signed commits


samsonasik:
I moved `ob_end_flush()` and `ob_start()` outside `ENVIRONMENT !== 'production'` check so it can work in production environment as well.

PR #1860, opened 2019-03-22
Migrationrunner use autoloader

MGatner:
**Description**
Currently MigrationRunner searches app/Config/Autoload.php for specified namespaces to include in migrations when a namespace is specified (or with "all"). This change uses the Autoloader service to search all discovered namespaces, so will include appropriately-namespaced migrations loaded from other sources (e.g. Composer).

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -1,4 +1,4 @@
-<?php namespace CodeIgniter\Autoloader;
+<?php namespace CodeIgniter\Database;
 
 /**
  * CodeIgniter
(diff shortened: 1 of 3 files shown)


system/Database/MigrationRunner.php
```diff
@@ -121,7 +122,7 @@ class MigrationRunner
 	/**
 	 * used to return messages for CLI.
 	 *
-	 * @var array
+	 * @var boolean
```
atishamte: It should be array

system/Database/MigrationRunner.php
```diff
@@ -694,7 +698,7 @@ protected function getVersion()
 	/**
 	 * Retrieves current schema version
 	 *
-	 * @return array    Current migration version
+	 * @return string    Current migration version
```
atishamte: Should be array

lonnieezell:
Thanks for tackling this. However, it makes more sense to use the [FileLocator class](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Autoloader/FileLocator.php) as it already has some methods that might do what you need here, like `listFiles()`.

MGatner:
listFiles() looks awesome and could definitely do a lot of the work for "all", but it seems to be generalized to all namespaces so unable to help with "load migrations from this namespace", unless I am misunderstanding something?
MigrationRunner needs both: load migrations from any namespace, and load migrations from a specified namespace. Right now both use `findMigrations()` with the former calling it for each namespace - maybe splitting those out so "all" methods use listFiles and the methods specific to a namespace get those paths from somewhere (Autoloader?).

lonnieezell:
@MGatner You're right. I thought there were other methods in there that would be appropriate but they don't quite look right. 

I think the ideal solution, then, is to add a new method to FileLocator then, and use it. That way everything's in one central place for those type of features. Would you mind?

system/Database/MigrationRunner.php
```diff
@@ -412,18 +413,21 @@ public function findMigrations()
 		// If $this->path contains a valid directory use it.
 		if (! empty($this->path))
 		{
-			$dir = rtrim($this->path, DIRECTORY_SEPARATOR) . '/';
+			$dirs = [rtrim($this->path, DIRECTORY_SEPARATOR) . '/'];
```
atishamte: Rename this variable to $dir

MGatner:
Yes I’ll update FileLocator then use that rewrite the relevant parts of MigrationRunner.

MGatner:
Updated FileLocator with a new method "listNamespaceFiles" and used that in MigrationRunner's findMigrations to extend namespace migrations to external PSR4 sources.

lonnieezell:
@MGatner Looks like an issue with class names - Travis has errors due to Migration class name being used twice. Guessing the test name.

system/Autoloader/FileLocator.php
```diff
@@ -36,372 +36,757 @@
...
- * Class FileLocator
- *
- * Allows loading non-class files in a namespaced manner.
- * Works with Helpers, Views, etc.
- *
- * @package CodeIgniter
+ * Class MigrationRunner
  */
-class FileLocator
+class MigrationRunner
```
lonnieezell: This file is the FileLocater file - and looks like you've copied MigrationRunner on it?

lonnieezell:
Looks like a big oops in here. I'll do a more thorough review once that's corrected.

system/Autoloader/FileLocator.php
MGatner: Yeah I'm not even sure how I did that... Just sent over a fix. This PR has become kinda a mess, let me know if you'd like a fresh start.

MGatner:
Okay sorry about that. I can't setup GPG on this computer so I've been pasting into GitHub - I must've pasted into the wrong window. Looks good now! Let me know if you'd prefer a fresh PR.

lonnieezell:
No worries. That looks great. And all tests are passing now. Thanks for tackling this!

PR #1861, opened 2019-03-23
New method Find Column w.r.t. #1619

atishamte:
W.r.t #1619, correction in the codes and introduced test cases as well

Its a shorthand query.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -350,18 +350,37 @@ class Model
 	//--------------------------------------------------------------------
 
 	/**
-	 * Fetches the column of database from $this->table with a primary key
-	 * matching $id.
+	 * Fetches the column of database from $this->table
 	 *
-	 * @param string|array                $column_name Column name
-	 * @param integer|string|array        $id          One primary key or an array of primary keys
+	 * @param string        $column_name Column name
 	 *
-	 * @return array|object|null    The resulting row of data, or null.
+	 * @return array|null   The resulting row of data, or null if no data found.
+	 *
+	 * @throws \CodeIgniter\Database\Exceptions\DataException
 	 */
-	public function findColumn($columnName, $id = null)
+	public function findColumn(string $columnName)
 	{
-		return $this->select($columnName)
-		            ->find($id);
+		if (strpos($columnName, ',') !== false)
+		{
+			throw DataException::forFindColumnHaveMultipleColumns();
+		}
+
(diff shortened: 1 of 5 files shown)


lonnieezell:
I believe the purpose of the originally proposed method was to return an array of the values only, at least when a single column was asked for, like: 

```
$job = $model->findColumn('name');
$this->assertEquals('Developer', $job[0]);
$this->assertEquals('Politician', $job[1]);
$this->assertEquals('Accountant', $job[2]);
$this->assertEquals('Musician', $job[3]);
```

That makes it easy to grab a column of values and then immediately use it for a whereIn, or whatever.

atishamte:
I kept the purpose is to shorthand the query, where you can get the response in the format of object or array when you can pass single and multiple columns.
But the response format will remain the same.
If I make any changes w.r.t. only array and someone wants a response in an object, then parsing is the only option to do so.
Also, there would be a confusion of the return type of single and multidimensional array w.r.t. single and multiple columns respectively.

lonnieezell:
My personal feeling is that if all it does is provide a shorthand way to get rid of doing `->select('x')` then it's not a feature I'm interested in. Returning an array of values is something I can see being a time saver for a lot of people so I'm ok with that. 

With that in mind - I think it should always return an array of the actual values. If need be, restrict it to a single column name.

atishamte:
No problem. I will make the necessary changes

atishamte:
Does the second parameter is required in the method? or we should strictly keep it as a result set of all values under the column without where condition.

atishamte:
I have done the changes for single column input along with exception handling

atishamte:
@lonnieezell and @jim-parry, any changes required in this PR?

system/Model.php
```diff
@@ -349,6 +349,42 @@ public function find($id = null)
...
+		                  ->find();
+
+		if (count($resultSet))
+		{
+			$data = [];
+			foreach ($resultSet as $item)
+			{
+				$data[] = $item[$columnName];
+			}
+
```
lonnieezell: Using `array_column` is likely faster here.

atishamte:
Changes done according to array_column

atishamte:
@lonnieezell any changes required here?

lonnieezell:
Looks good! Thanks.

PR #1868, opened 2019-03-25
WIP added db insert ignore support

Tada5hi:

**Description**
Added required insert ignore support for sqlite3 and mysql.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


lonnieezell:
A couple of things here: 

1. Would prefer new `insertIgnore` method added instead of just adding a new argument. The old database layer we've ported has too many arguments already. :) 
2. Any new commands would have to support all database platforms we support, if possible, and throw en error about unsupported feature if it doesn't. You've missed PostgreSQL here. 
3. We could like contributions to be GPG signed please. The user guide has docs for contributors.

Tada5hi:
Hi, thanks your reply.
1. 
Hi, yes of course, but that will nearly occure double function code which is nearly the same. 
But im fine with that.
Should i create also a method insertIgnore(), insertBatchIgnore() in the main BaseBuilder Class or just add them to the mysql/sqlite/postgresql builder class?

I saw that you abstract methods which are the same for most databases in the BaseBuilder Class.  But the insertIgnore statement is a little bit different for mysql and sqlite. Should i still create these methods in the basebuilder class but for example with empty function body or should i just create the methods in the builder class of mysql,sqlite,.. ?

Should i also create new protected methods _insertIgnore(), _insertBatchIgnore() or just new methods for the main public methods?

2. Yes, PostgreSQL is not supporting insert ignore. 
3. I will ;) Im currently trying to sign my already pushed commits. Till now i just was able to sign my latest commit ^^

lonnieezell:
Perhaps a better way is to make a new `ignore()` method that sets a class flag, which is false by default. Then, all of those methods can check the status of that flag to determine if they should be ignored. I think that's cleaner, and definitely better than a lot of new methods, and makes it simple for non-supporting drivers to ignore.

Tada5hi:
Yes think so too ;) 
If you are fine with file changes, i will update the user guide.

lonnieezell:
Yes, I'm fine with the changes. They look good. And you reset $ignoreInsert after the inserts, too - so great! 

Please update docs and add a test for it.

Tada5hi:
Yes, but first i have a little question. I found out that the ignore statement is not just valid for insert statements in case of mysql and sqlite. Here in short which driver supports the ignore statement in select, insert, ...
Select: Mysql
Insert: Mysql, Sqlite
Update: Mysql, Sqlite
Delete: Mysql
So i was thinking about to just modify the _insert(), _update(), ... methods in the BaseBuilder and to add a function called compileIgnore() which checks if the child driver class, has an attribute/variable called insertIgnore for example, which is null/not defined if the ignore feature is not avaialable for that driver and the method will return an empty string, but if its available it will return the value of that variable for example 'IGNORE' or in case of sqlite 'OR IGNORE'. So we keep double code as low as possible. what do you think ?

lonnieezell:
I'm fine with the idea in general, but it has to work in a broader way, where each database can modify the query as needed. For example. INSERT IGNORE is supported in Postgres 9.5+ through [ON CONFLICT DO NOTHING|UPDATE](https://www.postgresql.org/docs/current/sql-insert.html). Other databases may need similar additions when added. So, when coming up with a solution, please keep that in mind, and get it working in Postgres.

Tada5hi:
I think tests are kinda useless in that case @lonnieezell ,because the ignore statement is individual per Driver. 
And the MockConnection Class which is used for tests, just create an instance of MockBuilder instead of the specific Database Builder (f.e MySQLi Builder Class). Thats why the compileIgnore() function always returns an empty string, because the supported ignore statements are declared in the Driver Class. 

This is also why there is no check for the DBDriver in case of the Replace Test, because the specification for the Replace query of sqlite is made in the driver builder class. ('INSERT OR REPLACE' instead of just 'REPLACE). 
So in the test environment all Drivers going to have the same query syntax.

lonnieezell:
We have live database tests that run on Travis in each of the databases, so the tests aren't worthless, and they make sure it works with the actual database in a real-world situation.

Tada5hi:
docs are missing before a merge is possible ?

Tada5hi:
@lonnieezell What do you think ?

jim-parry:
Looks good to me, except for the _insert and _insertBatch methods in Postgre/Builder ... those don't seem related to this PR at all.

MGatner:
I think the content of this one is good, but it has lingered and now needs conflict resolution and maybe some cleanup. @Tada5hi are you still available and willing to touch it up a bit? I think we're really close.

Tada5hi:
sorry @MGatner missed that mention.

MGatner:
No worries! Could you glance at #2446 and make sure it's still all good, in your opinion?

Tada5hi:
loooks good ;)

Issue #1778, opened 2019-03-01
History::setFiles may crash when reading empty file


hwiesmann:
In case one of the files in WRITEPATH.'debugger/debugbar_*.json is empty the method setFiles will crash. The reason for the crash is that History::setFiles does not check if the local variable $contents contains an object before using it in the assignment

`files[] = [....];`

**CodeIgniter 4 version**
CodeIgniter4 Alpha 2.

**Affected module(s)**
History::setFiles.

**Expected behavior, and steps to reproduce if appropriate**
I do not know how it can happen that files in WRITEPATH.'debugbar' do not have any content. Therefore, I cannot give hints how to reproduce the error of having an empty file in the said directory. Though I suspect that a crash occurred while writing to one of the files and therefore the file was created without any content (or garbage content). But if one of the files is empty the above mentioned bug is visible.

atishamte:
Can you close this as well

PR #1869, opened 2019-03-25
History::SetFiles check #1778

atishamte:
History::SetFiles check #1778

As per the issue raised, In case one of the files in WRITEPATH.'debugger/debugbar_*.json is empty the method setFiles will crash. The reason for the crash is that History::setFiles does not check if the local variable $contents contain an object before using it in the assignment.

I have added a check whether the JSON is valid or not. If not then it will skip the iteration else proceed with it.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar/Collectors/History.php
+++ b/system/Debug/Toolbar/Collectors/History.php
@@ -104,22 +104,25 @@ class History extends BaseCollector
 
 			// Get the contents of this specific history request
 			$contents = file_get_contents($filename);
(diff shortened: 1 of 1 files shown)


atishamte:
Travis ci failed only in PHP 7.2, MySQL but passed in all other scenario.
@lonnieezell @jim-parry could you please check as I have changed nothing related to redis

PR #1890, opened 2019-03-28
CURLRequest - ignore HTTP exceptions.

nowackipawel:
Allows to get response content even if HTTP Exception occurs.


lonnieezell:
This feature already exists: [http_errors](https://codeigniter4.github.io/CodeIgniter4/libraries/curlrequest.html#http-errors).

nowackipawel:
Thank you! :).

Issue #1870, opened 2019-03-26
Should unmatched cache() return null?


MGatner:
The cache Service accessory for the Caching Library is incredibly helpful and makes for seamless data sourcing among session, input, etc. However, that it doesn't mimic the other wrappers in returning "null" makes for cumbersome extra checks.
I wonder if there was an explicit reason for cache to return "false" (making it impossible to cache anything to 'false')? And if not would that be okay to change?
Ref: https://codeigniter4.github.io/CodeIgniter4/general/common_functions.html#cache

EDIT: wording

atishamte:
I guess, it should be null if no data found with associated key passed to maintain the consistency throughout the project

MGatner:
Looking into this, it is a bit more complicated as each Cache handler has its own implementation of `get()` that handle the data differently. I still think this is worth changing, but it might take some help form someone who understands each handler (e.g. what does `$this->memcached->get($key)` return for missing keys?).

lonnieezell:
I don't know that there is a reason for it being different honestly. This and the session drivers were all ported over with very little changes, so I didn't analyze them much. I have no problem with them returning NULL, and that makes sense to make for consistency sake. 

Actually, NULL seems to make even more sense given that you could frequently store values of `false` so how do you know that's different then not there? 

But I also don't know the different engines all that well, so would be researching as I went, also.

MGatner:
Needing to store `false` is exactly what brought me to this! I didn’t realize these were direct ports - I’ll do some digging when I get a chance.

PR #1895, opened 2019-03-29
Unmatched Cache Library `get()` return null

MGatner:
**Description**
Currently there is no way to distinguish between returns of a cached value `false` and "no matching key". This changes each handler to return `null` when a key isn't found.
Fixes https://github.com/codeigniter4/CodeIgniter4/issues/1870

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Cache/Handlers/FileHandler.php
+++ b/system/Cache/Handlers/FileHandler.php
@@ -100,7 +100,7 @@ class FileHandler implements CacheInterface
 
 		$data = $this->getItem($key);
 
-		return is_array($data) ? $data['data'] : false;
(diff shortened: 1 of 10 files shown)


MGatner:
Looks like I messed up the reference to the class constant for Memcached result-not-found. I’m on the road today and tomorrow but can fix it after that - if anyone else wants to look into it I think it is a small change.

MGatner:
I think this is finally ready!

lonnieezell:
Looks good. Merging.

PR #1898, opened 2019-03-29
 Updated CodeIgniter version 3.0.0 to 4.0.0

atishamte:
Codeigniter version number was not consistent and updated across the project.
Some files have version 4.0.0 and some are 3.0.0

**Checklist:**
- [x] Securely signed commits


jim-parry:
Er, no - this is not correct. Some of the files are from 4.0 but many are from 3.0 and that is where the copyright should remain until that gets sorted out.

PR #1900, opened 2019-03-30
Update .htaccess for better security and caching

atishamte:
Update .htaccess for better security and caching

Added a few points like,
- Disable directory access
- CORS setting
- Disable image hotlinking
- Disable server signature
- mod expires for caching
- Added GZip and DEFLATE types

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -5,6 +5,9 @@
 # Sets the environment that CodeIgniter runs under.
 # SetEnv CI_ENVIRONMENT development
 
+# Disable directory browsing
+Options All -Indexes
+
 # ----------------------------------------------------------------------
 # UTF-8 encoding
 # ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


msaladna:
This PR needs a rework. htaccess changes should be limited to what's necessary for the application to work as intended. Anything extraneous needs to be the discretion of the person managing the server, including hotlinking - not the application that should be as neutral as possible.

For example, `Options All` implies all options are enabled, including `+FollowSymLinks` and `+SymLinksIfOwnerMatches`. Not all platforms enable `+FollowSymLinks` because in multi-user environments it can lead to some really nasty side-channel attacks with linking `index.foo` to say `config.php` where `index.foo` is the web user, `config.php` is the system user. Enabling such a directive in a secure environments yields a 550 ISE.

`ServerSignature` should be deferred to the admin configuring the server. In reverse proxy environments it would be useful to keep this on to see which node in the chain yielded an error.

`ExpiresDefault` already specifies 1 month as its base. No need to duplicate directives that must be evaluated serially on every request. Keep the htaccess terse if you want maintainability. 

AddOutputFilterByType [DEFLATE](https://httpd.apache.org/docs/2.4/mod/mod_deflate.html) will also take care of compression (and a far better option), but again this should be at the discretion of the global server configuration not the application. No need to also compress with mod_gzip unless you're running < Apache 2.0, but if so you have much bigger problems.

lonnieezell:
I believe hotlinking check has been removed. We'd welcome a PR to take care of all of that.

msaladna:
Jettison the merge. 

This is a framework, not an all-in-one systems management platform. Follow Postel's Law. Do the least that is necessary for CI to work and leave the rest to the user how they want to configure the environment. 

- Do you really want newcomers asking why their svg isn't updating? 
- Do you really need to support IE8?
- Do you really want CI to behave much differently behind Nginx than Apache?

Paradinight:
I would remove the .htaccess file completely. it is not the framework task. 
short: we are not wordpress or an other cms.

PR #1902, opened 2019-03-30
BaseBuilder Corrections

atishamte:
Some BaseBuilder corrections like,
- Undefined variables defined.
- Removed unnecessary declarations
- Typos corrections

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


system/Database/BaseBuilder.php
```diff
@@ -390,13 +397,18 @@ public function selectSum($select = '', $alias = '')
...
 	{
-		if (! is_string($select) || $select === '')
+		if ($select === '')
 		{
 			throw new DatabaseException('The query you submitted is not valid.');
 		}
 
+		if (strpos($select, ',') !== false)
+		{
+			throw new DatabaseException('Only single column allowed.');
```
lonnieezell: Please add this as a new static method in the DatabaseException like the others. Also the message should be localized and stored in a lang file.

system/Database/BaseBuilder.php
```diff
@@ -2343,13 +2360,11 @@ protected function _truncate($table)
...
 	 */
-	public function getCompiledDelete($reset = true)
+	public function getCompiledDelete(bool $reset = true): string
 	{
 		$table = $this->QBFrom[0];
 
-		$this->returnDeleteSQL = true;
-		$sql                   = $this->delete($table, '', null, $reset);
-		$this->returnDeleteSQL = false;
+		$sql = $this->delete($table, '', null, $reset);
```
lonnieezell: While I agree with removing the wrapping lines as not neccessary - the default value for `$returnSQL` in the `delete()` method is false, and this function is expecting it to be returned. 

You need to pass `true` as the 4th argument here.

system/Database/BaseBuilder.php
```diff
@@ -2625,18 +2638,8 @@ protected function compileWhereHaving($qb_key)
 						continue;
 					}
 
-					// $matches = array(
```
lonnieezell: Put these back in please. They serve as a nice reference when working on that code...

system/Database/Postgre/Builder.php
```diff
@@ -351,15 +351,15 @@ protected function _truncate($table)
...
-	 * @param boolean     $insensitiveSearch
+	 * @param string    $prefix
+	 * @param string    $column
+	 * @param string    $not
+	 * @param string    $bind
+	 * @param boolean   $insensitiveSearch
 	 *
 	 * @return string     $like_statement
 	 */
-	public function _like_statement(string $prefix = null, string $column, string $not = null, string $bind, bool $insensitiveSearch = false): string
```
lonnieezell: What's the reason for removing null values here? These have been this way for many, many years as part of CI3 and probably before. Unless you've confirmed that every use case for this method can work when those are not null, they need to stay. 

I get they don't follow best practices of having nullable arguments last, but if that's the only reason you're changing it, it needs to be fixed.

atishamte: This method is called only from only one place, from where all the params are getting passed to it.
So there is no need of declaring arguments as optional. Hence removed

atishamte:
@lonnieezell, Changes are done.

atishamte:
Any update on this @lonnieezell

atishamte:
@lonnieezell @jim-parry, why the last commit didn't build by Travis CI?

jim-parry:
There are conflictrs that need to be resolved, would be my guess.

atishamte:
Thanks for starting the build. I have done with all the changes.

Issue #1767, opened 2019-02-28
Controller Test / Feature Testing output issues


lonnieezell:
The ControllerTester currently only gets a body in the result when the content is `echo()` in the controller. Using a `return` here doesn't work. 

However - the opposite problem exists with FeatureTesting.

jim-parry:
ControllerTester problem fixed by 1848 & 1855. 
Feature testing up shortly.

jim-parry:
@lonnieezell The user guide suggests that withRoutes() takes an array of mappings, but the FeatureTestCase is using an array of triplet arrays (http verb | add, URI pattern, and routing destination.

Which should it be?

lonnieezell:
@jim-parry Looks like docs are wrong in that case. Should be able to specify the HTTP verb to use like implemented.

jim-parry:
ok, thanks. i will update the docs as part of this

jim-parry:
There is a bigger problem underlying this, with testing routing rules not sticking. 
Working on it.

lonnieezell:
Uh-oh! Need any help on that?

jim-parry:
The problem I am experiencing is that the routecollection is not being updated/replaced by second & subsequent tests in the FeatureTestTestCase. I will push what I have as a WIP -maybe you can see something I am missing, thanks.

Issue #1697, opened 2019-01-31
route_to('name'); does not work for other subdomains


nowackipawel:
I've got config like this:

```$routes->get('i/(:any)', 	'App\Controllers\Site\CDoc::item/$1', 	['subdomain' => 'doc', 'as' => 'doc_item']);```

and I'd like to use it in my project on all of subdomains (doc, dev, etc...) which are in my project. 
It is working as far as it is called using `route_to('doc_item','sth');` from doc.domain.com, but when I am trying to do this from dev.domain.com it does not work. 

I think route_to should not filter possible values to one hostname/subdomain as  Router do; that's why we could specify name of the route i.e. `as=>doc_item`. Many sites has many subdomains and use subdomains widely.  IMHO no possibility for simple linking between subdomains is a strong limit.

jim-parry:
You explicitly provided for a subdomain restriction in your routing options, so it would be expected that the routing rule not apply to the other subdomains.
It sounds like what you are suggesting is already doable by not specifying the subdomain option, no?
What am I missing?

Issue #1692, opened 2019-01-27
Router issue - overwriting.


nowackipawel:
I've got two subdomains:
doc and dev.
And my Routes.php is as bellow:
```
$routes->setDefaultNamespace('App\Controllers');
$routes->setDefaultController('Home');
$routes->setDefaultMethod('index');
...
$routes->get('/', 	'App\Controllers\Site\CDoc::index', ['subdomain' => 'doc', 'as' => 'doc_index']);
$routes->add('/', 'Home::index');
```
but after request to https://doc.domain I get Home::index instead of App\Controllers\Site\CDoc::index
I thought the first which is matched is used.... but nope.
I did a test and I changed it to:

```
$routes->add('/', 'Home::index');  // get is working in the same way
$routes->get('/', 	'App\Controllers\Site\CDoc::index', ['subdomain' => 'doc', 'as' => 'doc_index']);
```

effect still the same - doc.domain is opening Home::index .

OFC everything is working as excepted if dev.domain route to Home::index is defined like that : `
$routes->add('/', 'Home::index', ['subdomain' => 'dev']);`

lonnieezell:
Just pushed up a possible fix for this. Can you pull the latest code and give it a try?

nowackipawel:
Will update my code in next few days and I'll let you know. I have to compare my base code with my uncommitted PRs.

nowackipawel:
Ok, I've updated RouteCollection.php only to make quick check.
Unfortunately does not work as expected. 


In this case:
```
$routes->get('/', 			'App\Controllers\Site\CDoc::index', 	['subdomain' => 'doc', 'as' => 'doc_index']);
$routes->add('/', 'Home::index', 	['subdomain' => 'dev']);
``` 
everything is working.

```
$routes->get('/', 			'App\Controllers\Site\CDoc::index', 	['subdomain' => 'doc', 'as' => 'doc_index']);
$routes->add('/', 'Home::index');
 ```
still opens Home:index for https://doc.domain.com.

PR #1904, opened 2019-04-01
WIP Improvements to FeatureTestCaseTest

jim-parry:
- added 'patch' to allowed HTTP verbs in RouteCollection
- added resetRoutes() in RouteCollection, to support FeatureTestCase
- added output capturing to FeatureTestCase::call
- added more methods to tests/_support/Controllers/Popcorn, to generate all the different kinds of responses
- added note to testing/features in user guide

Related to #1767, #1692 & possibly #1697


tests/_support/Controllers/Popcorn.php
```diff
@@ -55,4 +55,19 @@ public function index3()
...
 	}
 
+	public function canyon()
+	{
+		echo 'Hello-o-o';
+	}
+
+	public function json()
+	{
+		$this->responsd(['answer' => 42]);
```
lonnieezell: Typo in `response`. Not sure if intentional for this controller...

jim-parry: Good catch, though it doesn't affect the tests - still messed :(

lonnieezell:
@jim-parry I got it working for me in [this branch](https://github.com/codeigniter4/CodeIgniter4/tree/jim-parry-testing30). Does that work for you also?

jim-parry:
testing

jim-parry:
That does fix things :) I have incorporated the changes into my original PR & pushed. Now I can finish the tests, thanks!

jim-parry:
Hmmm - the tests run locally fine as "./phpunit tests/system/Test/", but not as "./phpunit".
I don't see how that affects things, but will revisit in the morning :-/

jim-parry:
travis-ci is breaking in SQLite3 stuff - not releated to this PR as far as I can tell.

system/CLI/CLI.php
```diff
@@ -147,7 +147,7 @@ class CLI
 	 *
 	 * @var string
 	 */
-	protected static $lastWrite;
+	protected static $lastWrite = 'write';
```
lonnieezell: What were you running into here? Lastwrite should be null before it's used so that an initial call to 'write' will add an empty line.

lonnieezell:
These two might have been caused by these changes: 

> 1) CodeIgniter\CodeIgniterTest::testRunClosureRoute
> Test code or tested code did not (only) close its own output buffers
> 2) CodeIgniter\CodeIgniterTest::testRun404OverrideByClosure
> Test code or tested code did not (only) close its own output buffers

No clue why the sqlite is failing now. Nothing in here should have touched that.

system/CLI/CLI.php
jim-parry: The problem is that the previous unit tests were not looking for that empty line, and hence failed on mmy system. Basically, the previous PR broke that unit test on my system.

jim-parry:
I have been getting some strange unit testing results with CodeIgniterTest and MigrationRunnerTest for some weeks now. I think it might be time to reclone & reconstruct my local repo :-/
I'm going to leave this PR unmerged for now, and deal with it properly when time permits. I'll tag it for the next release instead.

PR #1909, opened 2019-04-03
Update array_helper.php

nowackipawel:
Is array sequential  / associative

### DIFF
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -45,6 +45,37 @@
  * @link       https://codeigniter.com/user_guide/helpers/cookie_helper.html
  */
 
+
+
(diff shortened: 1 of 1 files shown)


atishamte:
Also provide the documentation and test scenarios

atishamte:
With array_key_exists(), I can directly check whether the array is associative or sequential.

nowackipawel:
How about:

[0 => 'a', 'b' => 'c'];

?

atishamte:
This is not an associative array.

lonnieezell:
Sorry. I don't think these have broad enough requirements in day-to-day work, and as @atishamte says - PHP already has tools to help you get the answer you need.

Issue #512, opened 2017-05-20
WIP Improve unit tests


jim-parry:
Our goal is to achieve 80% code coverage before release.
- [x] API 
- [x] Autoloader :)
- [x] CLI :)
- [ ] Cache **
- [ ] Commands **
- [x] Config :)
- [ ] Database **
- [x] Migrations  
- [ ] Debug **
- [x] Events :)
- [x] Exceptions 
- [x] Files :)
- [x] Filters :)
- [x] Format :)
- [x] HTTP :)
- [x] Helpers :)
- [x] Honeypot :)
- [x] i18n --> Time::toFormattedDateString()
- [ ] Images *
- [x] Language :)
- [x] Log :)
- [x] Pager :)
- [x] Router *
- [x] Security :)
- [ ] Session **
- [x] Test :)
- [x] Throttle :)
- [x] Typography :)
- [x] Validation :)
- [x] View :)
- [ ] CodeIgniter.php *
- [x] Common.php --> force_https(), is_really_writable(), function_usable(), dd()
- [x] Controller.php *
- [x] Entity.php 
- [ ] Model.php *

jim-parry:
- [ ] API (almost there) ... in progress JLP
- [ ] Autoloader (almost there) ... in progress KZ
- [ ] CLI
- [ ] Cache
- [ ] Config ... in progress LY
- [ ] Database
- [ ] Debug
- [ ] Events
- [ ] Files
- [ ] Filters
- [ ] Format
- [ ] HTTP
- [ ] Helpers (almost there)
- [ ] Language
- [ ] Log
- [ ] Pager
- [ ] Router
- [ ] Session
- [ ] Test
- [ ] Throttle
- [ ] Typography
- [ ] Validation
- [ ] View
- [ ] CodeIgniter.php
- [ ] Controller.php

Brijesh144:
ff

jaynarayan89:
Is this list up to date?

jim-parry:
@jaynarayan89 It is indeed up-to-date ... there are lots of unit tests needed :-/

paolooo:
@jim-parry Is there any discussion on this on how we can help to improve the tests? I'd like to help on Cache.

nikkolai14:
Hello! Is this still available? I would love to help to improve the Unit Testing. Thanks :)

jim-parry:
@nikkolai14 You bet. Specific needs:
- database testing for postgres
[...]

PR #1911, opened 2019-04-03
JSON Cast exception test cases

atishamte:
W.r.t #512 Adding test scenarios to increase the coverage of Cast Exception
WIP JSON Cast exception test cases.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -313,6 +313,7 @@ class Entity
 	 * @param null   $value
 	 *
 	 * @return $this
+	 * @throws \Exception
 	 */
 	public function __set(string $key, $value = null)
 	{
@@ -349,6 +350,11 @@ class Entity
 			if (($castTo === 'json' || $castTo === 'json-array') && function_exists('json_encode'))
 			{
 				$value = json_encode($value);
+
+				if (json_last_error() !== JSON_ERROR_NONE)
+				{
+					throw CastException::forInvalidJsonFormatException(json_last_error());
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


atishamte:
@lonnieezell @jim-parry, I am trying to write the test case for the private method castAsJson() of Entity Class.
Below is my code for the test scenario.
```
public function testCastAsJSONSyntaxError()
{
	$entity = new Entity();
	$method = $this->getPrivateMethodInvoker($entity,'castAsJson');
		
	$this->expectException(CastException::class);
	$this->expectExceptionMessage('Syntax error, malformed JSON');

	$method("'{ bar: \"baz\" }'", true);
}
```
At the time of build, I am getting error like,
```
1) CodeIgniter\EntityTest::testCastAsJSONSyntaxError
Failed asserting that exception of type "CodeIgniter\Exceptions\CastException" is thrown.
```

Could you please help me out with this. So I will be able to generate test scenarios.

jim-parry:
It looks like you are trying to json_decode a JSON string and expecting that to trigger an exception. That doesn't sound right.
Wouldn't you want to give the method something that can't be json_decoded? eg "( this is bad"

atishamte:
What I am asking, Is this the correct way to implement the test scenarios?

jim-parry:
Yes, it is reasonable.
Normally, however, you would thoroughly test on your system before pushing (and ending up with a bunch of commits), rather than using travis-ci to see if they work. Do you have the unit tests running on your system?

atishamte:
I have tried but the system is not able to configure XDebug. I have tried many times but no success.
Also, I do not have Postgre, so all the test scenarios of it give an error.

PR #1916, opened 2019-04-04
Check if the value is string

daif:
Make sure the value is a string before using preg_match function


jim-parry:
Future PRs need to be GPG-signed, thanks.

Issue #1773, opened 2019-03-01
Ignoring 'required' validation rule for inserts.


nowackipawel:
This is my code:
```

		$test_model = (new \TestModel);

		CLI::write('INSERT_TEST tes_id + tes_txt' , 'red');
		$test_1 = (new \App\Entities\Test());
		$test_1->tes_id = rand(20,200);
		$test_1->tes_txt = '111' . microtime();
		CLI::write(json_encode($test_1->toRawArray()));
		CLI::write($test_model->save($test_1));
		CLI::write('VALIDATION ERRORS: ' . print_r($test_model->errors(true), true));

		CLI::write('INSERT_TEST tes_id + tes_txt' , 'red');
		$test_2 = (new \App\Entities\Test());
		$test_2->tes_id = rand(20,200);
		$test_2->tes_txt = null;
		CLI::write(json_encode($test_2->toRawArray()));
		CLI::write($test_model->save($test_2));  // PRODUCES INSERT INTO `t_test` (`tes_id`) VALUES (138)

		CLI::write('VALIDATION ERRORS: ' . print_r($test_model->errors(), true)); // NO ERRORS!
```

and my model has validation rules like:

```
	protected $validationRules = [
		'tes_txt' 	=> 'required|min_length[2]',
		'tes_json' 	=> 'permit_empty|valid_json'
	];
```

Results for spark code:
```
CodeIgniter CLI Tool - Version 4.0.0-alpha.5 - Server-Time: 2019-03-01 08:55:30am

INSERT_TEST tes_id + tes_txt
{"tes_id":139,"tes_txt":"1110.12893900 1551430530","tes_json":null,"tes_created_at":null,"tes_updated_at":null}
139 INSERT INTO `t_test` (`tes_id`, `tes_txt`) VALUES (139, '1110.12893900 1551430530')
VALIDATION ERRORS: 
INSERT_TEST tes_id + tes_txt
{"tes_id":138,"tes_txt":null,"tes_json":null,"tes_created_at":null,"tes_updated_at":null}
138 INSERT INTO `t_test` (`tes_id`) VALUES (138)
VALIDATION ERRORS: 
```

Unfortunately as long as tes_txt = NULL  $test_2 is passing validation. If I change it to '' or 'x' validation is working (required or min_length[2])

onyxg:
this is the problem: Validation/Rules.php/min_length


```
public function min_length(string $str = null, string $val, array $data): bool
{
[...]

PR #1924, opened 2019-04-08
Model changes w.r.t. #1773

atishamte:
Model changes w.r.t. #1773 

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -533,7 +533,7 @@ class Model
 	{
 		if (method_exists($data, 'toRawArray'))
 		{
-			$properties = $data->toRawArray(true);
+			$properties = $data->toRawArray();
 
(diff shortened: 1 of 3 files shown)


lonnieezell:
Removing the `true` in `toRawArray` will just cause other issues with unintentional overwrites of data.

I don't see how this fixes the `required` rules at all. Can't accept.

atishamte:
In the stated issue #1773, he is trying to save the null value in the database which expected to throw an error of required field.
But when data is passed from insert method to classToArray, it expects to return only changed values.
The previous and current value of the property is null, hence it is not returning the property as changed value.
Hence error is not getting triggered.

PR #1927, opened 2019-04-09
Entity exception for non existed props.

nowackipawel:
As for now if code calls non existed property of entity causing ErrorException with no info at all.
This PR is changing ErrorException to EntityException with translated information like this:
"Trying to access non existent property fle_amount_fromss of App\Entities\Fee\FeeLevel"

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -295,12 +295,12 @@ class Entity
 			$result = $this->castAs($result, $this->_options['casts'][$key]);
 		}
 		
-		if(isset($result) && $result !== null)
+		if(! isset($result) && ! property_exists($this, $key))
 		{
-			return $result;
+			throw EntityException::forTryingToAccessNonExistentProperty($key, __CLASS__);
 		}
 
-		throw EntityException::forTryingToAccessNonExistentProperty($key, __CLASS__);
+		return $result;
 
 	}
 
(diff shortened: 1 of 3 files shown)


lonnieezell:
You're right it's better than current, but that doesn't mean it's the right long-term solution. :)

Still waiting to hear @jim-parry thoughts.

jim-parry:
I favor the idea of declaring entity properties explicitly, and thus support this PR.

lonnieezell:
The part I'm concerned about is when you have a join, and select some fields from the other table. That property was not explicitly set on the object, but was implicitly defined through the query result, though I suppose it would still pass muster as existing at the time someone tried to get a property? It's late and I'm not honestly sure of that answer. Will have to investigate later.

jim-parry:
Hmmm. If you are planning that an entity be the result of a join, would you not then be able to "predict" that it should have a specific property? I haven't used the Entity class, and don't have a good feel for what would work best :-/

nowackipawel:
If entity is a result of joint it will contain all the own properties + new
ones from join. Properties from join are public and should not produce that
exception.

lonnieezell:
@nowackipawel Now that my brain is a little clearer and less tired - you're right. I'm good with this then. 

> If you are planning that an entity be the result of a join, would you not then be able to "predict" that it should have a specific property?

sometimes, but not always since the entity will be used in different situations that might require different sets of data joined in, sometimes only for a single call.

PR #1930, opened 2019-04-09
Model Changes w.r.t. #1773

atishamte:
In the stated issue #1773, he is trying to save the null value in the database which expected to throw an error of required field.
But when data is passed from the insert method to classToArray, it expects to return only changed values.
The previous and current value of the property is null, hence it is not returning the property as changed value.
Hence error is not getting triggered.

So some required changes made here like,
1. Made allowed null for description column of job table in migration
2. Removed if condition from save() method as again update() or save() method performs same operation in the classToArray() method.
3. Introduced $onlyChanged parameter to classToArray() function. So we can escape the only changed value check in case of insert the data but it will return only changed values data in case of the update as the flag value is true by default.
4. Passing false value to classToArray method for variable $onlyChanged in the insert method as we required all the params, not the changed values only.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -484,14 +484,6 @@ class Model
 	 */
 	public function save($data): bool
 	{
-		// If $data is using a custom class with public or protected
-		// properties representing the table elements, we need to grab
-		// them as an array.
-		if (is_object($data) && ! $data instanceof \stdClass)
-		{
-			$data = static::classToArray($data, $this->primaryKey, $this->dateFormat);
-		}
-
 		if (empty($data))
 		{
 			return true;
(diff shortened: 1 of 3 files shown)


lonnieezell:
Much better solution. I'm a little leery about removing the classToArray at the beginning of `save()` but think it should be fine. Thanks.

MGatner:
Point number 4 has introduced a new problem. When a database field is not null and the field on an entity is not set, `classToArray` will still return the null value which tries to get inserted into the database, throwing an error. IMO `save()` should never receive *all* values from `classToArray` and the solution to #1773 should be fixed in `validate()` instead.

MGatner:
After a bit more research I do believe this is an issue and would have failed tests except that the test UserModel.php supplies default values to its fields, a non-mandatory (and probably non-standard) choice:
https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/_support/Models/UserModel.php

I will open a new issue.

PR #1933, opened 2019-04-10
having (Is NULL deletion)

nowackipawel:
Is it breaks anything or will repair having('MAX(sth)', null, false)
before:
```
SELECT `aht_cur_id_2` `cur_id_2`, `aht_cur_id_1` `cur_id_1`, `aht_rate` `rate`
FROM `t_api_trading_trades`
GROUP BY `aht_cur_id_2`, `aht_cur_id_1`
HAVING MAX(aht_api_trade_id) IS NULL
```
now:
```
SELECT `aht_cur_id_2` `cur_id_2`, `aht_cur_id_1` `cur_id_1`, `aht_rate` `rate`
FROM `t_api_trading_trades`
GROUP BY `aht_cur_id_2`, `aht_cur_id_1`
HAVING MAX(aht_api_trade_id)
```

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -706,7 +706,7 @@ class BaseBuilder
 
 				$v = " :$bind:";
 			}
-			elseif (! $this->hasOperator($k))
+			elseif (! $this->hasOperator($k) && $qb_key !== 'QBHaving')
 			{
 				// value appears not to have been set, assign the test to IS NULL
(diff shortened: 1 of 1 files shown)


lonnieezell:
I have no idea if it will break something, since I'm sure it was there for a reason, but a comparison to the CI3 codebase and it looks like it was removed from there at some point since we forked that code. 

Will merge.

PR #1937, opened 2019-04-11
Update Model.php

nowackipawel:
TypeError

Argument 2 passed to CodeIgniter\Database\BaseBuilder::whereIn() must be of the type array or null, string given, called in /home/wanda-exchange/site/system/Model.php on line 802 

on update('string', 'entity');

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -799,7 +799,14 @@ class Model
 
 		if ($id)
 		{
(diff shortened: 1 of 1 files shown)


MGatner:
@lonnieezell this was actually resolved by https://github.com/codeigniter4/CodeIgniter4/pull/1963 and should be closed.

jim-parry:
Resolved

PR #1939, opened 2019-04-11
Remove section that prevents hotlinking

MGatner:
**Description**
The rewrite rule required a hard-coded domain which developers would need to update or their assets would give 403 errors. Better to allow developers to add this themselves than supply an .htaccess file that doesn't work out-of-the-box.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


MGatner:
Discussion on Slack: https://codeigniterchat.slack.com/archives/C47FK3GRY/p1554997719093400

PR #1940, opened 2019-04-12
Fix empty `$data` check since trigger had wrap it into a non-empty array

vibbow:
**Description**
Fix empty `$data` check since trigger had wrap it into a non-empty array

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


vibbow:
This change is no longer needed.

PR #1943, opened 2019-04-13
Model, Entity, Exception & Migration test cases

atishamte:
Model, Entity, Exception & Migration Test Cases to improve code coverage.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Database/Live/ModelTest.php
+++ b/tests/system/Database/Live/ModelTest.php
@@ -2,6 +2,7 @@
 
 use CodeIgniter\Config\Config;
 use CodeIgniter\Database\Exceptions\DataException;
+use CodeIgniter\Entity;
 use CodeIgniter\I18n\Time;
 use CodeIgniter\Model;
 use CodeIgniter\Test\CIDatabaseTestCase;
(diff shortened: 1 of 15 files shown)


tests/system/Database/Live/ModelTest.php
```diff
@@ -709,17 +728,21 @@ public function testCanCreateAndSaveEntityClasses()
 		$this->assertEquals('Developer', $entity->name);
 		$this->assertEquals('Awesome job, but sometimes makes you bored', $entity->description);
 
+		$time = time();
```
lonnieezell: Using seconds can cause false negatives when the test execution crosses the threshold from one second to another, which is why we were using just the date here.

atishamte: But job table has a datetime column datatype as integer, so we cannot save the formatted date in it. Hence changed.

lonnieezell: Then check around the base test classes. I believe @jim-parry create a "close enough" assertion for times like this so tests don't fail randomly because they crossed into a new second during execution.

atishamte: I will just convert this timestamp and database saved timestamp to dates to compare. Will this work?

lonnieezell: that also works.

atishamte: No problem. Will do it.

atishamte:
@lonnieezell, I have done the changes.

Issue #1907, opened 2019-04-02
Wrong controller filter processing because of faulty regular expression generation


hwiesmann:
**Describe the bug**
Filters::processFilters methods replaces internally filter expressions with wrong regular expressions.

**CodeIgniter 4 version**
CodeIgniter4 Beta 0.0.1

**Affected module(s)**
Filters

**Expected behaviour, and steps to reproduce if appropriate**
Assume the following filter statement in Config/Filters.php:

```
public $filters = ['login' => ['before' => ['company/*']]];
```

This filter setting should set a filter for controller Company. Unfortunately, due to the wrong filter processing also controllers like CompanyXYZ will have the filter applied.

The reason for this wrong behaviour is that - for example - 'company/*' is replaced by the regular expression 'company.+' which is basically equivalent with 'company*'. But this is not correct.

The two statements responsible for the replacement are:
```
$path = strtolower(str_replace('/*', '*', $path));
$path = trim(str_replace('*', '.+', $path), '/ ');
```
 
I do not know why `str_replace('/*', '*', $path)` is required?! At least in my example this is the root for the problems.

PR #1946, opened 2019-04-14
Filter changes for path mismatch issue

atishamte:
w.r.t. #1907, changed the replace as it changes the actually mentioned route in before and after if contains '/*'

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -355,9 +355,7 @@ class Filters
 				foreach ($rules as $path)
 				{
 					// Prep it for regex
-					$path = strtolower(str_replace('/*', '*', $path));
-					$path = trim(str_replace('*', '.+', $path), '/ ');
-
+					$path = trim(str_replace('*', '.+', strtolower($path)), '/ ');
 					// Path doesn't match the URI? continue on...
 					if (preg_match('#' . $path . '#', $uri, $match) !== 1)
 					{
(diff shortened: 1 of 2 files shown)


lonnieezell:
I don't think this completely fixes the issues. I had started working on this and ran out of time with the day job hammering me into overtime. But there wasn't any test that actually covered the case they were running into.

Here's the testcase that I was using that was failing: 

```
/**
	 * @see https://github.com/codeigniter4/CodeIgniter4/issues/1907
	 */
	public function testFilterMatching()
	{
		$_SERVER['REQUEST_METHOD'] = 'GET';

		$config  = [
			'globals' => [
				'before' => [
					'foo' => ['except' => 'admin*'],
					'bar'
				],
				'after'  => [
					'foo' => ['except' => 'admin/*'],
					'baz'
				],
			],
		];
		$filters = new Filters((object) $config, $this->request, $this->response);
		$uri     = 'admin';

		$expected = [
			'before' => [
				'bar',
			],
			'after'  => [
				'foo',
				'baz',
			],
		];

		$this->assertEquals($expected, $filters->initialize($uri)->getFilters());
	}
```

When playing with the regex it looked like the had to be `.*` instead of `.+` but that caused other issues, too, so I hadn't quite worked out a proper regex for it yet.

atishamte:
I will check the provided test scenario.

atishamte:
In your test case, the expected should be,
```
$expected = [
	'before' => [
		'bar',
	],
	'after'  => [
		'foo' => ['except' => 'admin/*'],
		'baz',
	],
];
```

PR #1947, opened 2019-04-16
Fix BaseConfig didn't load Registrar files properly

vibbow:
**Description**
In origional `BaseConfig.php` file, in `registrProperties` function, it use `locator` to find all `Config/Registrar.php` files in known namespace, and it returned all file path as an array.

In following code, it use the file path as a callable object, to check if registrars exists (by using method_exists) for current config file, which didn't work in my case.

I'm not sure if method_exists works with file path, but I updated this function to load all registrars files first before future usage.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/BaseConfig.php
+++ b/system/Config/BaseConfig.php
@@ -181,8 +181,15 @@ class BaseConfig
 
 		if (! static::$didDiscovery)
 		{
-			$locator              = \Config\Services::locator();
-			static::$registrars   = $locator->search('Config/Registrar.php');
+			$locator         = \Config\Services::locator();
+			$registrarsFiles = $locator->search('Config/Registrar.php');
+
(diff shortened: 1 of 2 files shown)


vibbow:
Wait a second, the tests is failed and it's related to this change.

It failed because `$properties = $callable::$shortName();` didn't return an array.

And `$properties = $callable::$shortName();` get executed because `method_exists($callable, $shortName)` returned `true`

And I can't see why it returned a `true` because `$registrars` should be empty, because there should be no `'Config/Registrar.php'` file exists in known namespace.

Emm, seemd I need to spend some time to let the test run on my local.

Or if anyone can help to point out where I did wrong.

vibbow:
Get failed test fixed.

PR #1950, opened 2019-04-17
Fix some side issue

vibbow:
**Description**
Fix some side issue

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1024,7 +1024,7 @@ class BaseBuilder
 	 *
 	 * @return string     $like_statement
 	 */
-	private function _like_statement(string $prefix = null, string $column, string $not = null, string $bind, bool $insensitiveSearch = false): string
+	protected function _like_statement(string $prefix = null, string $column, string $not = null, string $bind, bool $insensitiveSearch = false): string
 	{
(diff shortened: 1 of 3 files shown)


system/Database/BaseBuilder.php
```diff
@@ -1024,7 +1024,7 @@ protected function _like($field, string $match = '', string $type = 'AND ', stri
 	 *
 	 * @return string     $like_statement
 	 */
-	public function _like_statement(string $prefix = null, string $column, string $not = null, string $bind, bool $insensitiveSearch = false): string
+	private function _like_statement(string $prefix = null, string $column, string $not = null, string $bind, bool $insensitiveSearch = false): string
```
lonnieezell: Why did you make this private? First off - I'm not a fan of private methods in general since they make it impossible to use in extending classes. Second, I think this _might_ get called by child drivers, but I'm not sure. 

If it really needs to be changed, it should be protected, and there are likely other methods in the same situation.

vibbow: Oh yes, it should be protected. 

I searched entire codebase, `_like_statement` only showed up three times.

Two from `BaseBuilder.php`, one from `Builder.php`, which extend `BaseBuilder`

PR #1952, opened 2019-04-17
Add Validations for `equals()` and `not_equals()`

MGatner:
**Description**
Per conversation in Slack, there is currently no way to check a precise value (or not), such as confirming a user typed "yes". This adds validation rules for `equals` and `not_equals`, along with their language texts, user guide entires, and tests.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


system/Validation/Rules.php
```diff
@@ -65,6 +65,21 @@ public function differs(string $str = null, string $field, array $data): bool
 
 	//--------------------------------------------------------------------
 
+	/**
+	 * Equals the static value provided.
+	 *
+	 * @param string $str
+	 * @param string $str
```
atishamte: Please correct the param here. It should be $val

system/Validation/Rules.php
```diff
@@ -261,6 +276,21 @@ public function min_length(string $str = null, string $val, array $data): bool
 
 	//--------------------------------------------------------------------
 
+	/**
+	 * Does not equal the static value provided.
+	 *
+	 * @param string $str
+	 * @param string $str
```
atishamte: Follow the comment of line 72

atishamte:
Please provide the unit test case related to all the datatype and string case sensitive and insensitive comparison.
Also update the user guide as well.

nowackipawel:
Could we don't use in_list / not_in_list / regex rules ?

MGatner:
@nowackipawel While those supply overlapping functionality, they make for weird error messages such as "The confirmation field must be one of: yes." This allows for clearer messages as well as easier to understand code.

system/Validation/Rules.php
MGatner: Thanks for catching that! Fixed.

MGatner:
> Please provide the unit test case related to all the datatype and string case sensitive and insensitive comparison.
> Also update the user guide as well.

I accidentally sent user guide and tests to my develop branch. I included those and added tests for case sensitivity. Data types are irrelevant because rules "must accept a string as the first parameter" (https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html#creating-custom-rules).

PR #1954, opened 2019-04-19
Fix validation table format

jim-parry:
Fix validation table format mistake introduced in #1952


MGatner:
Sorry about that! I pasted it into GitHub so it would be a verified commit, must have not transferred well. Thanks for cleaning up.

PR #1958, opened 2019-04-20
Code realigned and cleanup of App

atishamte:
Code realigned and cleanup of App
- Remove inline namespace use, moved to the top of the page
- Removed unused code from Controllers
- Realigned and corrected the HTML syntactically
- Updated the user agents

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Cache.php
+++ b/app/Config/Cache.php
@@ -1,5 +1,11 @@
 <?php namespace Config;
 
+use CodeIgniter\Cache\Handlers\DummyHandler;
+use CodeIgniter\Cache\Handlers\FileHandler;
+use CodeIgniter\Cache\Handlers\MemcachedHandler;
+use CodeIgniter\Cache\Handlers\PredisHandler;
+use CodeIgniter\Cache\Handlers\RedisHandler;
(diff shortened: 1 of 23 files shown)


app/Config/Cache.php
```diff
@@ -110,11 +116,11 @@ class Cache extends BaseConfig
 	|
 	*/
 	public $validHandlers = [
-		'dummy'     => \CodeIgniter\Cache\Handlers\DummyHandler::class,
-		'file'      => \CodeIgniter\Cache\Handlers\FileHandler::class,
-		'memcached' => \CodeIgniter\Cache\Handlers\MemcachedHandler::class,
-		'predis'    => \CodeIgniter\Cache\Handlers\PredisHandler::class,
-		'redis'     => \CodeIgniter\Cache\Handlers\RedisHandler::class,
-		'wincache'  => \CodeIgniter\Cache\Handlers\WincacheHandler::class,
+		'dummy'     => DummyHandler::class,
```
lonnieezell: Because these are configuration values that end users are expected to be able to easily change, I prefer these to show the full namespace. Please revert. Also the other similar changes in this PR should be reverted also.

Issue #1953, opened 2019-04-18
redirect($namedRoute) missing helpful exception


MGatner:
Passing an invalid named route to `redirect()` causes a type error:
```
SYSTEMPATH/Router/RouteCollection.php at line 1117
ltrim() expects parameter 1 to be string, object given
...
1115                 // Lose any namespace slash at beginning of strings
1116                 // to ensure more consistent match.
1117                 $to     = ltrim($to, '\\');
```

This should be throwing something helpful, like `HTTPException::forInvalidRedirectRoute`. Tested with and without custom route definitions.

* PHP: 7.2.16
* CodeIgniter: 4.0.0-beta.2

MGatner:
It seems to be choking on Closures, trying to use `ltrim()` on them. Since this section is explicitly looking for a string match to `$search`  I think it would be find to skip anything else:
```
			foreach ($collection as $route)
			{
				$from = key($route['route']);
				$to   = $route['route'][$from];
				if (! is_string($to)
				{
				             continue;
				}
```

PR #1959, opened 2019-04-21
Prevent reverseRoute from searching closures

MGatner:
**Description**
`reverseRoute` searches loaded routes for a string match, but will throw an exception on unmatched routes when it tries to `trim` any routes whose definitions are closures. This ignores non-string destinations in the secondary search phase.
Fix for https://github.com/codeigniter4/CodeIgniter4/issues/1953

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1111,6 +1111,12 @@ class RouteCollection implements RouteCollectionInterface
 			{
 				$from = key($route['route']);
 				$to   = $route['route'][$from];
+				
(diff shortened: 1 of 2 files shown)


MGatner:
I believe this is still relevant, and would be cause `testReverseRoutingReturnsFalseWithNoMatch()` to fail except that it isn't loading default routes from `system/Config/Routes.php` which include Closures. I'm adding a test to demonstrate the failure.

Issue #1928, opened 2019-04-09
Unable to use _remap without default method in controller


krishnan57474:
**Describe the bug**
A controller with method other than _remap
creates error in routes toolbar collector.

**CodeIgniter 4 version**
CodeIgniter4 Beta 0.0.2

**Affected module(s)**
system\Debug\Toolbar\Collectors\Routes.php

**Expected behavior, and steps to reproduce if appropriate**

1. Set CodeIgniter environment to development
2. Rename index method to _remap in app/Controllers/Home.php
3. run

You get method not exists error.

The following statement in system\Debug\Toolbar\Collectors\Routes.php:

        $method    = is_callable($router->controllerName()) ? new \ReflectionFunction($router->controllerName()) : new \ReflectionMethod($router->controllerName(), $router->methodName());

Calling **\ReflectionMethod($router->controllerName(), $router->methodName())** without method_exists check
causes method not exists error.

**Fix**

        $method    = is_callable($router->controllerName()) ? new \ReflectionFunction($router->controllerName()) : new \ReflectionMethod($router->controllerName(), method_exists($router->controllerName(), '_remap') ? '_remap' : $router->methodName());

jim-parry:
Unable to reproduce this specific problem.
Applying the proposed solution breaks ControllerTesterTest, which expects an index() method.
The user guide says that index() *is* the default method, hence it seems reasonable that not having it, whether or not you have a _remap(), would cause problems.

PR #1960, opened 2019-04-21
Fix unable to use _remap without default method

krishnan57474:
**Description**
A controller with method other than _remap creates ReflectionException error in routes collector toolbar.
Fix for #1928

**Checklist:**

- [x]  Securely signed commits

- [x]  Component(s) with PHPdocs

- [x]  Unit testing, with >80% coverage

- [x]  User guide updated

- [x]  Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar/Collectors/Routes.php
+++ b/system/Debug/Toolbar/Collectors/Routes.php
@@ -88,7 +88,7 @@ class Routes extends BaseCollector
 		$route = $router->getMatchedRoute();
 
 		// Get our parameters
-		$method    = is_callable($router->controllerName()) ? new \ReflectionFunction($router->controllerName()) : new \ReflectionMethod($router->controllerName(), $router->methodName());
+		$method    = is_callable($router->controllerName()) ? new \ReflectionFunction($router->controllerName()) : new \ReflectionMethod($router->controllerName(), method_exists($router->controllerName(), '_remap') ? '_remap' : $router->methodName());
 		$rawParams = $method->getParameters();
 
 		$params = [];


system/Debug/Toolbar/Collectors/Routes.php
```diff
@@ -88,7 +88,7 @@ public function display(): array
 		$route = $router->getMatchedRoute();
 
 		// Get our parameters
-		$method    = is_callable($router->controllerName()) ? new \ReflectionFunction($router->controllerName()) : new \ReflectionMethod($router->controllerName(), $router->methodName());
+		$method    = is_callable($router->controllerName()) ? new \ReflectionFunction($router->controllerName()) : new \ReflectionMethod($router->controllerName(), method_exists($router->controllerName(), '_remap') ? '_remap' : $router->methodName());
```
atishamte: I guess this will always go for remap...

krishnan57474: nope it's not always go for remap. only when controller contains remap method. the main issue is, when controller contains remap method only, it creates error. $router->methodName() always return "index" when controller contains remap method only. if controller having both remap and index method, toolbar always choose index instead of remap. but router always prefers remap first. calling ReflectionMethod without checking method exists check creates ReflextionException error.

atishamte: What I am saying here is,
for example, method abc() is present and I am calling same method from route, then also, according to your condition, if both method _remap() and abc() are present, it will always go for _remap().
But ideally, it should go directly to abc()

krishnan57474: yup, you are right. but if you have controller with only one method, that is remap. then router correctly choose remap method, but toolbar always fallback to default method **index**, without checking method existence.

Issue #1971, opened 2019-04-24
Toolbar won't accept custom collectors


MGatner:
The User Guide outlines creating a custom Collector for the Debug Toolbar:
https://codeigniter4.github.io/CodeIgniter4/testing/debugging.html?#creating-custom-collectors

However, the Toolbar service takes a static `$viewsPath` from `app/Config/Toolbar.php` and  then toolbar.tpl.php uses this to render a view template for each Collector:
https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Debug/Toolbar/Views/toolbar.tpl.php#L139

Therein lies the conundrum: `$viewsPath` will either be the default `SYSTEMPATH . 'Debug/Toolbar/Views/'` (where developers should not inject their own `_packagename.tpl`) or some other path which will lack the view templates for core Collectors.

The Guide states "`display()` should return a string of HTML" but it actually needs to return an array: https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Debug/Toolbar/Collectors/BaseCollector.php#L237)

My guess is that the User Guide is reflecting a step in the direction of making it easier to  create custom Collectors (by directly return HTML from `display()`), but that change hasn't been implemented in the Toolbar class yet.

PR #1972, opened 2019-04-24
Toolbar fix for custom collectors

MGatner:
**Description**
Custom collectors fail because toolbar.tpl.php tries to parse their view file from `SYSTEMPATH . 'Debug/Toolbar/Views/'`. See #1971 for complete description of the issue.
This change allows `Collector->display()` to return a string (as stated in the User Guide) and then, when detected in the view template, displays it directly instead of looking for a separate template to parse.

*There's probably a better way to handle this by checking for a namespaced view template for the collector.*

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


lonnieezell:
I think that's fine for now, but does warrant a deeper look into the disconnect between user guide and functionality.

PR #1976, opened 2019-04-30
__set boolean / datetime

nowackipawel:
I dunno why I didn't do this PR earlier. During last merge I've noticed I didn't do it.


nowackipawel:
@lonnieezell @jim-parry  is sth wrong with this PR?

PR #1977, opened 2019-05-01
Test/featuretestcase

### LINKED ISSUE #1767: Controller Test / Feature Testing output issues

jim-parry:
- added 'patch' to allowed HTTP verbs in RouteCollection
- added resetRoutes() in RouteCollection, to support FeatureTestCase
- added output capturing to FeatureTestCase::call
- added more methods to tests/_support/Controllers/Popcorn, to generate all the different kinds of responses
- added note to testing/features in user guide

Related to #1767, #1692 & possibly #1697
Closes #1767

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -619,9 +619,9 @@ class RouteCollection implements RouteCollectionInterface
 	 * Example:
 	 *      $routes->add('news', 'Posts::index');
 	 *
-	 * @param string        $from
-	 * @param array|string  $to
-	 * @param array         $options
+	 * @param string       $from
+	 * @param array|string $to
+	 * @param array        $options
 	 *
 	 * @return RouteCollectionInterface
 	 */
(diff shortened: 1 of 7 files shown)


jim-parry:
Three SQLite3 tests are failing, and there are some warnings for SQLite3 too.
I don't see these being related to the current PR, though they pass travis-ci for other PRs.
This should be ready to go, as far as I can tell.

lonnieezell:
Doesn't look like those tests are affected by anything you did. They seem to be missing a `$refresh = true` or similar on those test classes.

jim-parry:
I think there is something screwy with travis-ci.
I re-ran the test suite that failed, and now three different tests fail - Validation ones.
It seems that the db_user table is sometimes there for SQLite3, and sometimes not.

PR #1985, opened 2019-05-03
Fix filter processing. Fixes #1907

### LINKED ISSUE #1907: Wrong controller filter processing because of faulty regular expression generation

jim-parry:
Reworked & refactored innards of Filters, specifically processGlobals and processFilters.
They had a few issues.

**Checklist:**
- [x] Securely signed commits
- [-] Component(s) with PHPdocs
- [-] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -1,5 +1,4 @@
 <?php
-
 /**
  * CodeIgniter
  *
@@ -317,9 +316,7 @@ class Filters
 	 */
 	public function getArguments(string $key = null)
 	{
-		return is_null($key)
-			? $this->arguments
-			: $this->arguments[$key];
+		return is_null($key) ? $this->arguments : $this->arguments[$key];
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jim-parry:
... and SQLite3 has gotten in travis-ci's way again :(

PR #1989, opened 2019-05-06
argument set() must by type of string - cannot agree

nowackipawel:
set should accept at least NULL | int | string
UPDATE x SET a = NULL ...
UPDATE x SET a = 100
UPDATE x SET a = 'aaa'


lonnieezell:
I'll merge but to be fair, due to the nature of PHP - all of those items you show, null, int, and string get typecast during that method, I believe.

Issue #1992, opened 2019-05-07
Entity null values cause database error


MGatner:
Issue #1773 needed null values to be included in model verification, but by using *all* values from `classToArray` (instead of $onlyChanged) `Model->save()` now attempts to insert `null` entity values into the database, which will throw an exception for columns that cannot be `null`.

I think `save()` should still use only changed values and validation should be updated to check for non-existent required values. Also, I believe #1930 passed tests because tests/system/Database/Live/ModelTest.php (e.g. `testSaveNewEntityWithDate()`) is defining all field values, which is non-mandatory.

lonnieezell:
You're right that `save()` should only used changed fields during updates, but all fields should be used if it's an insert, I would think. I'm not sure off the top of my head how to know which fields are allowed nullable during insert and which should be ignored if they are null, without adding an additional layer of complexity for everyone, which I don't want to do.

lonnieezell:
We're hitting some complex issues here that I think are going to require a refactor of Entities. On the plus side, I think it'll make them more robust. My plan is to make them work a little more like Laravel's Models insofar as you don't have to specify properties anymore. We'll keep two protected properties - `original` and `attributes` that store the original attributes and the current attributes. It should simplify some other usage issues I've been struggling with for a while.

Unfortunately, that means that anyone who's built anything already will have to refactor, but that's the fun of pre-release software I guess.

MGatner:
I love it! Always worth it to stop and redo something the right way. 

Let me know how I can help.

nowackipawel:
PR#1989 ?

lonnieezell:
@MGatner I got through about 80% of the required changes last night, I believe. So shouldn't take too long to get switched over. Thanks, though!

@nowackipawel Unrelated to Entities.

PR #1993, opened 2019-05-07
Add partial Entity save() testing

MGatner:
**Description**
Issue #1992 describes the scenario where an entity without every field set will cause non-null database column to error `{column_name} cannot be null`. This PR will create the necessary tests to catch this scenario prior to providing a fix.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/tests/system/Database/Live/ModelTest.php
+++ b/tests/system/Database/Live/ModelTest.php
@@ -1477,8 +1477,6 @@ class ModelTest extends CIDatabaseTestCase
 		$entity->deleted    = 0;
 		$entity->created_at = new Time('now');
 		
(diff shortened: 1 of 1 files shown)


MGatner:
Testing with the expected exception "worked" (I got the exception type wrong, but it still threw: https://travis-ci.org/codeigniter4/CodeIgniter4/jobs/529399940). I've removed the expectation and this should now cause a failure because the entity doesn't have every field set.

MGatner:
Made irrelevant by Entity refactor, https://github.com/codeigniter4/CodeIgniter4/pull/2002

PR #2001, opened 2019-05-12
Subqueries in BaseBuilder

iRedds:
**Description**
Added subqueries via anonymous function in Query Builder for BaseBuilder::where() and BaseBuilder::whereIn() methods

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -707,9 +707,8 @@ class BaseBuilder
 
 				if ($v instanceof Closure)
 				{
-					$builder = clone $this;
-					$builder->from([], true)->resetQuery();
-					$v = '(' . str_replace("\n", ' ', $v($builder)->getCompiledSelect()) . ')';
+					$builder = $this->cleanClone();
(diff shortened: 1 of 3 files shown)


system/Database/BaseBuilder.php
```diff
@@ -837,13 +847,21 @@ protected function _whereIn(string $key = null, array $values = null, bool $not
 
 		$not = ($not) ? ' NOT' : '';
 
-		$where_in = array_values($values);
-		$ok       = $this->setBind($ok, $where_in, $escape);
+		if ($values instanceof Closure)
+		{
+			$builder = clone $this;
+			$builder->from([], true)->resetQuery();
```
lonnieezell: Since we're using this same code in 2 places, we should refactor those out into their own method to make future maintenance simpler

iRedds: made it

jim-parry:
Looks good to me.

Issue #2006, opened 2019-05-15
validation match[x] don't work anymore... if custom setter is used.


nowackipawel:
In entity:
```
	public function setUsrPassword($value)
	{
//$this->usr_password 
		$this->attributes['usr_password'] = password_hash('my-password', PASSWORD_BCRYPT, ['cost' => 11]);
	}
```
in model:
```
		$this->validationRules['usr_password_1'] = 'required|matches[usr_password]';
```
I think it is because custom set is called now before validation... before last changes it was called before save action.

nowackipawel:
Additionally 

this code in Entity:
```
	public function setUsrPassword($value)
	{
		$this->attributes['usr_password'] = password_hash( trim($value) . config('App')->salt, PASSWORD_BCRYPT, ['cost' => 11]);
	}
```
makes value $user_entity->usr_password  different every time.

lonnieezell:
I'll look into the validation issue, but there's a simple fix for the second part you brought up: use a different field name. For example, in my auth stuff, I've got a field in the database called `password_hash` to be clear about what's being saved. Then in the Entity there is a `setPassword()` method which does the same as yours. However, it's setting the value of the `password_hash` column and that setter is only called if the password has been changed. 

If you save the password field (using your `setUsrPassword` method) then, yes, the hash will be different every time. That's how password_hash works.

lonnieezell:
Custom sets will always be called on an Entity prior to validation if you're doing validation in the model. 

For certain things, like validating two fields that must match each other, that's best done in the controller before you try to save to database.

PR #2007, opened 2019-05-15
Serialize array only

daif:
Serialize the value if it is an array only, the data from database is already serialized.


nowackipawel:
I think is related to #2006  and every other data which were cast before saving to DB.
IMHO now Entity is doing too much casting.

PR #2009, opened 2019-05-16
Bugfix: Routes respect spoofed methods

MGatner:
**Description**
By default the loaded route collection uses REQUEST_METHOD directly from `$_SERVER`:
`$this->HTTPVerb = strtolower($_SERVER['REQUEST_METHOD'] ?? 'cli');`
This means when `Router->checkRoutes()` matches its collection it uses the actual request method instead of the spoofed method from the Request object. This PR injects the proper HTTP verb prior to instantiating the router with `$routes->setHTTPVerb()`, though alternatively RouteCollection's constructor could be changed instead.

To recreate the bug, create a resource route to a non-autorouted controller:
`$routes->resource('foo', ['controller' =>'MyNamespace\Controllers\Foo']);
And then try to access one of the non-POST supported methods by spoofing:
`<input type="hidden" name="_method" value="PUT" />`
Routing ignores the PUT routes and attempts to autoroute:
`Controller or its method is not found: App\Controllers\Foo`

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Commands/Utilities/Routes.php
+++ b/system/Commands/Utilities/Routes.php
@@ -109,6 +109,7 @@ class Routes extends BaseCommand
 			'get',
 			'head',
 			'post',
+			'patch',
 			'put',
 			'delete',
 			'options',


MGatner:
Hmm this is failing for CLI, should've thought of that. Let me think about the best way around this.

PR #2011, opened 2019-05-16
Set the raw data array without any mutations for the Entity

iRedds:
**Description**
When we create, for example, the password hashing method Entity::setPassword(), this method is also called when the value from the database is filled.
As a result, we get a hash of the hash in $entity->password

This PR sets the value of the $attribute property to the unchanged values from the database.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -433,7 +433,7 @@ class Entity
 	 * @param  array $data
 	 * @return $this
 	 */
-	public function setRawArray(array $data)
+	public function setAttributes(array $data)
 	{
 		$this->attributes = $data;
 		$this->syncOriginal();
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -238,7 +238,7 @@ class Entity
 	 * Checks a property to see if it has changed since the entity was created.
 	 * Or, without a parameter, checks if any properties have changed.
 	 *
-	 * @param ?string $key
+	 * @param string $key
 	 *
 	 * @return boolean
 	 */
@@ -247,9 +247,9 @@ class Entity
 		// If no parameter was given then check all attributes
 		if ($key === null)
 		{
-			return 	$this->original !== $this->attributes;
+			return     $this->original !== $this->attributes;
 		}
-		
+
 		// Key doesn't exist in either
 		if (! array_key_exists($key, $this->original) && ! array_key_exists($key, $this->attributes))
 		{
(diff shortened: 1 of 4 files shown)


system/Entity.php
```diff
@@ -427,6 +427,19 @@ public function __isset(string $key): bool
 		return isset($this->attributes[$key]);
 	}
 
+	/**
+	 * Set raw data array without any mutations
+	 *
+	 * @param  array $data
+	 * @return $this
+	 */
+	public function setRawArray(array $data)
```
lonnieezell: Rename to `setAttributes` please.

iRedds: What checks do you mean?

lonnieezell: Not a fan of the name change - it's not an abstract class. Even then not sure i'd like it.

Was referring to the checks in the database layer that you've added where it checks if it's an Entity and then uses `setRawArray`. If we move that to the Entity constructor I don't think any of those checks are needed since things will be set to their raw values the way the database layer currently works.

iRedds: Then we say goodbye to the stdClass class and the custom class.
Because stdClass has no constructor, and the custom class may not have a constructor.

lonnieezell: Nothing's changing there. It works currently. The database engine itself handles that in `$this->resultID->fetch_object($className);`. 

I'm not saying you should remove existing functionality - but remove the portions you added in this PR. They all look something like: 

```
if (is_subclass_of($className, Entity::class))
		{
			$data = $this->fetchAssoc();
			return empty($data) ? false : (new $className())->setRawArray($data);
		}
```

iRedds: I'm confused. 
If I remove this code
```php
if (is_subclass_of($className, Entity::class))
{
    return empty($data = $this->fetchAssoc()) ? false : new $className($data);
}
```
then this PR does not make sense.

lonnieezell: Actually - I might be dense here. I was thinking that it would set through the constructor, but it doesn't. It tries to set the value directly on the object. It's late and I'm tired and distracted with other problems. Sorry - what you've done works fine. Just rename that method, please.

neznaika0:
Any other reason besides setting a password?
I extended the Entity class by adding private properties. abandoned the magic of setters/getters - created get* set*
Now setAttributes cannot call set*
So fill() is the better solution.
Regarding the password hash - this operation is done in the beforeInsert Model or by setting the hash yourself

PR #2012, opened 2019-05-16
Use request->method for HTTP verb

MGatner:
**Description**
By default the loaded route collection uses REQUEST_METHOD directly from $_SERVER:
$this->HTTPVerb = strtolower($_SERVER['REQUEST_METHOD'] ?? 'cli');
This means when Router->checkRoutes() matches its collection it uses the actual request method instead of the spoofed method from the Request object. This PR injects the proper HTTP verb prior using `$request->getMethod()`

To recreate the bug, create a resource route to a non-autorouted controller:
$routes->resource('foo', ['controller' =>'MyNamespace\Controllers\Foo']); And then try to access one of the non-POST supported methods by spoofing: Routing ignores the PUT routes and attempts to autoroute:Controller or its method is not found: App\Controllers\Foo`

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -39,6 +39,7 @@
 
 namespace CodeIgniter\Router;
 
+use Config\Services;
 use CodeIgniter\Autoloader\FileLocator;
 use CodeIgniter\Router\Exceptions\RouterException;
 
@@ -234,7 +235,7 @@ class RouteCollection implements RouteCollectionInterface
 	public function __construct(FileLocator $locator, $moduleConfig)
 	{
 		// Get HTTP verb
-		$this->HTTPVerb = strtolower($_SERVER['REQUEST_METHOD'] ?? 'cli');
+		$this->HTTPVerb = is_cli() ? 'cli' : Services::request()->getMethod();
 
 		$this->fileLocator = $locator;
 
@@ -1115,12 +1116,6 @@ class RouteCollection implements RouteCollectionInterface
 			{
 				$from = key($route['route']);
 				$to   = $route['route'][$from];
-				
-				// ignore closures
-				if (! is_string($to))
-				{
-					continue;
-				}
 
 				// Lose any namespace slash at beginning of strings
 				// to ensure more consistent match.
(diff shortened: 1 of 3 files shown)


MGatner:
Calling backup! @lonnieezell I think this is probably the correct way to handle this - however all the tests presume that routes will be matched off `$_SERVER['REQUEST_METHOD']` instead of `$request->method()` which accounts for spoofing. Before I (or anyone) goes through and updates the tests I want to make sure this is the desired approach.

system/Router/RouteCollection.php
```diff
@@ -1115,12 +1116,6 @@ public function reverseRoute(string $search, ...$params)
 			{
 				$from = key($route['route']);
 				$to   = $route['route'][$from];
-				
-				// ignore closures
-				if (! is_string($to))
```
lonnieezell: Is this not needed? That was recently added by another user to prevent errors. Are you seeing something different?

lonnieezell:
@MGatner Yes, I think that should work. 

Although we should be able to do away with the `is_cli()` check if we handle that during the `CodeIgniter::spoofRequestMethod()` call. Looks like it doesn't set the request method to 'cli' there, and it probably should.

system/Router/RouteCollection.php
MGatner: I added that! It is needed because otherwise a missing reverseRoute will cause an exception. Check the Travis log for what I'm referring to on the test failures:

https://travis-ci.org/codeigniter4/CodeIgniter4/jobs/533482026

MGatner: Ohhh I see what you mean, sorry, yes I guess I didn't have that merged. Will fix.

MGatner: Re-added the tea for closures.

MGatner:
@lonnieezell This changes are in place. It's failing on `testResourcesWithWebsafe`, which when I look at it makes me wonder how that ever succeeded: `resource()` should definitely be returning more than just the 3 routes listed in the test.

MGatner:
Nevermind, it was supposed to be testing POST! Quick fix coming up.

MGatner:
Looks good now!

PR #2015, opened 2019-05-18
Direct user to follow the upgrade steps after installation

agmckee:
As discussed on the community forums, the appstarter doesn't work out of the box after a fresh installation. This was determined to be due to the composer.lock file being out of date, running composer update installs the up to date version of the framework.

**Description**
Updated the installation via Composer documentation to direct users to follow the steps in the "Upgrading" section even after a fresh installation. This ensures the framework is up to date.

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPdocs: N/A
- [ ] Unit testing, with >80% coverage: N/A
- [x] User guide updated
- [ ] Conforms to style guide: N/A

### DIFF
--- a/user_guide_src/source/installation/installing_composer.rst
+++ b/user_guide_src/source/installation/installing_composer.rst
@@ -48,6 +48,8 @@ A sample such installation command, using the default project-root "appstarter":
 
     composer create-project codeigniter4/appstarter -s beta --no-dev
 
+After installation you should follow the steps in the "Upgrading" section.
+
 Upgrading
 -------------------------------------------------------
 


jim-parry:
Oooh - you may have hit the nail on the head ... the appstarter ships with a composer.lock file that it shouldn't! Let me play with that ... this PR might not be needed if the problem can be fixed at the root :)

jim-parry:
composer.lock is git ignored, and I don't know how it got into the repo :(
I have deleted it, and re-created the beta.3 release, but it still shows up with alpha.4 and needs updating :(
I am going to leave your comment in the installation instructions, and try to figure out this puzzle later.
Thanks!

MGatner:
@jim-parry Sorry if this is a dumb question, but did you remove the file or clean it from the history? .gitignore has weird interactions with files already committed and `git rm`.

jim-parry:
Good question. I probably didn't do it right, and need to fix it :(

Issue #2021, opened 2019-05-22
PR#2012 caused 404 exception in spark 


nowackipawel:
This line: 
```
		$this->HTTPVerb = is_cli() ? 'cli' : Services::request()->getMethod();```
has been changed to:
```
		$this->HTTPVerb =  Services::request()->getMethod();
```

and this is the result
``

CodeIgniter CLI Tool - Version 4.0.0-beta.3 - Server-Time: 2019-05-22 14:22:25pm

ERROR: 404
Can't find a route for 'cimigrate'.

```

MGatner:
Just found this myself (https://github.com/codeigniter4/devstarter/issues/2). The issue is that CodeIgniter.php's `spoofRequestMethod()` is actually setting the method on `clirequest` service (see `getRequestObject()`) and RouteCollection is always checking `request` service. I think there is no way around checking for CLI in the RouteCollection constructor.

PR #2022, opened 2019-05-22
Add CLI check to RouteCollection

MGatner:
**Description**
CodeIgniter.php sets the request method on `request` or `clirequest` service, so RouteCollection needs to check the proper service to get the correct HTTPVerb to account for CLI, spoofing, and standard scenarios.
See https://github.com/codeigniter4/CodeIgniter4/issues/2021

FWIW I believe the reason this wasn't caught is that all the request & routing clauses in `CodeIgniter.php` have conditional `! (ENVIRONMENT === 'testing')` to be able to simulate non-CLI requests from CLI. There should probably be a test that spoofs its own environment (if possible) just to check for `spark` integrity.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -234,8 +234,11 @@ class RouteCollection implements RouteCollectionInterface
 	 */
 	public function __construct(FileLocator $locator, $moduleConfig)
 	{
+		// Load the correct request service
+		$request = (is_cli() && ! (ENVIRONMENT === 'testing')) ? Services::clirequest() : Services::request();
+		
 		// Get HTTP verb from current request (accounts for spoofing)
-		$this->HTTPVerb = Services::request()->getMethod();
+		$this->HTTPVerb = $request->getMethod();
 
 		$this->fileLocator = $locator;
 


lonnieezell:
I think a slightly better fix here is to either pass in the request instance either through the constructor or through a setter method. That removes the dependency creation back to CodeIgniter.php, where it already knows which type of request it is.

MGatner:
That's a good idea! I'm out for a few days but I'll take a look when I get a chance. Glad for anyone else to hop in before then.

MGatner:
Looking at this, get/set HTTPVerb and the `$HTTPVerb` property only appear to be used internally to `RouteCollection` and really just for `getRoutes()`. It is also always just a duplicate of `$_SERVER['REQUEST_METHOD']` (old version) or `$request->getMethod()` (current revision).

Do we get rid of `HTTPVerb` altogether, overload the class to `__construct(FileLocator $locator, $moduleConfig, $request = null)` and then just use `$request->getMethod()` for `getRoutes()` fallback?

MGatner:
One more place for HTTPVerb, called from Router but extraneous as it is simply injecting what is loaded by default:

`./Router/Router.php:390:		$routes = $this->collection->getRoutes($this->collection->getHTTPVerb());`

lonnieezell:
@MGatner I've got this about wrapped up, actually. Just have one more test I have to fix up. It was a bigger change than expected. Due to the way routeCollection is used I didn't want to force a new call within route files to that, so I added it as a dependency to the Router class, which then sets the HTTPVerb on the RouteCollection itself.

MGatner:
Oh great! Thanks for addressing it - I look forward to seeing your solution. Any thoughts on testing (re: my original comment "FWIW...")?

PR #2023, opened 2019-05-23
whereIn  $value do not have to be an array

nowackipawel:
i.e.:

```
			$sub_query[] = $this->type === self::API_TRADING_TRADES ? $this->prefix . 'api_trade_id' : $this->prefix . 'api_id';

			$this->select('MAX(' . $subselect[0] . ')', false);

			if($cur_id_from === null)
			{
				$this->groupBy($this->prefix . 'cur_id_1');
			}

			if($cur_id_to === null)
			{
				$this->groupBy($this->prefix . 'cur_id_2');
			}

			if ($type !== null)
			{
				$this->where($this->prefix . 'type', $type);
			}


			$sub_query[] = $this->getCompiledSelect(true);


			$this->whereIn($sub_query[0],'(' . $sub_query[1]  . ')', false);

```
should produce sth like this:
```
SELECT *
FROM `t_api_trading_trades`
WHERE aht_api_trade_id IN (SELECT MAX(aht_api_trade_id)
FROM `t_api_trading_trades`
WHERE `aht_type` = 'B'
GROUP BY `aht_cur_id_2`)
```
(tested, working as excepted)

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -734,12 +734,12 @@ class BaseBuilder
 	 * joined with 'AND' if appropriate.
 	 *
 	 * @param string  $key    The field to search
-	 * @param array   $values The values searched on
+	 * @param array|string   $values The values searched on
 	 * @param boolean $escape
 	 *
 	 * @return BaseBuilder
 	 */
-	public function whereIn(string $key = null, array $values = null, bool $escape = null)
+	public function whereIn(string $key = null, $values = null, bool $escape = null)
 	{
 		return $this->_whereIn($key, $values, false, 'AND ', $escape);
 	}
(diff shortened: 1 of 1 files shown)


nowackipawel:
@lonnieezell @jim-parry  Is sth wrong here?

jim-parry:
The explanation is awkward, but the PR looks reasonable.

lonnieezell:
Well - biggest problem is no tests for this case. Which would be especially helpful now that I've merged another PR that allows closures here. I can try to take a look and make sure they play nice together but we need automated tests in the future @nowackipawel 

While I'm glad you took the time to confirm it worked on your project, that doesn't give us any security that any changes down the road to the framework will continue to work.

PR #2027, opened 2019-05-27
Fix CSRF hash regeneration

Workoverflow:
For CSRF hash regeneration, $this->CSRFHash must be set to NULL when user set 
$this->CSRFRegenerate to TRUE, because in CSRFSetHash() you compare it with NULL, but 
$this->CSRFHash !== NULL after request. I guess you need to create new method in Security class, such as CSRFHashReset() and check my commit.

### DIFF
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -221,6 +221,7 @@ class Security
 		if ($this->CSRFRegenerate)
 		{
 			// Nothing should last forever
+			$this->CSRFHash = null;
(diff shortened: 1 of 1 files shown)


jim-parry:
Looks good to me

Workoverflow:
I’m happy to help you with you difficult way. Thx you for best solution for we apps!01.09.2019, 03:07, "Project lead, CodeIgniter" <notifications@github.com>:Looks good to me

PR #2040, opened 2019-06-05
Update CURLRequest.php

nowackipawel:
empty content-lenght header fix


nowackipawel:
@lonnieezell , @jim-parry : could you merge this pls to fix "Content-length:" header?

PR #2048, opened 2019-06-09
WIP Keep track of Entity exist state

Tada5hi:

**Description**
I needed a good way to keep track of the exist state of an entity, so if i want to save it i can check via the exist state, if an update or an insert is required.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -88,6 +88,14 @@ class Entity
 	protected $original = [];
 
 	/**
+	 * Keep track whether the entity
+	 * exists in the Database Table or not.
+	 *
+	 * @var boolean
+	 */
+	protected $exists = false;
+
+	/**
 	 * Holds info whenever properties have to be casted
 	 *
 	 * @var boolean
@@ -139,6 +147,28 @@ class Entity
 		}
 
 		return $this;
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Set boolean value for entity state.
+	 *
+	 * @param boolean $exists
+	 */
+	public function setExists(bool $exists)
+	{
+		$this->exists = $exists;
+	}
+
+	/**
+	 * Checks if the entity exists in the database.
+	 *
+	 * @return boolean
+	 */
+	public function exists() : bool
+	{
+		return $this->exists;
 	}
 
 	//--------------------------------------------------------------------
--- a/system/Model.php
+++ b/system/Model.php
@@ -672,9 +672,11 @@ class Model
 		// them as an array.
 		if (is_object($rawData) && ! $rawData instanceof stdClass)
 		{
-			$tryUpdate = method_exists($rawData, 'exists') && $rawData->exists() && property_exists($rawData, $this->primaryKey);
+			$tryUpdate = method_exists($rawData, 'exists') && $rawData->exists();
 
 			$rawData = static::classToArray($rawData, $this->primaryKey, $this->dateFormat, false);
+
+			$tryUpdate = $tryUpdate && isset($rawData->{$this->primaryKey});
 
 			if ($tryUpdate)
 			{
@@ -737,7 +739,7 @@ class Model
 		$this->insertID = $this->db->insertID();
 
 		// Update exists state through reference
-		if (is_object($data) && property_exists($data, 'setExists'))
+		if (is_object($data) && method_exists($data, 'setExists'))
 		{
 			$data->setExists(true);
 			$data->{$this->primaryKey} = $this->insertID;
--- a/system/Model.php
+++ b/system/Model.php
@@ -518,7 +518,7 @@ class Model
 	 * @return boolean
 	 * @throws \ReflectionException
 	 */
-	public function save($data): bool
+	public function save(&$data): bool
 	{
 		if (empty($data))
 		{
@@ -540,6 +540,18 @@ class Model
 		}
 
 		$response = $this->insert($data, false);
+
+		// Update exists state through reference
+		if (is_object($data) && method_exists($data, 'setExists'))
+		{
+			$data->setExists(true);
+
+			if (! isset($data->{$this->primaryKey}))
+			{
+				$data->{$this->primaryKey} = $this->insertID;
+			}
+		}
+
 		// call insert directly if you want the ID or the record object
 		if ($response !== false)
 		{
@@ -645,24 +657,20 @@ class Model
 	 * @return integer|string|boolean
 	 * @throws \ReflectionException
 	 */
-	public function insert(&$data = null, bool $returnID = true)
+	public function insert($data = null, bool $returnID = true)
 	{
 		$escape = null;
 
 		$this->insertID = 0;
 
-		// Copy data array, so we don't lose
-		// object reference to $data var in case of updating exist state.
-		$rawData = $data;
-
-		if (empty($rawData))
+		if (empty($data))
 		{
-			$rawData        = $this->tempData['data'] ?? null;
+			$data           = $this->tempData['data'] ?? null;
 			$escape         = $this->tempData['escape'] ?? null;
 			$this->tempData = [];
 		}
 
-		if (empty($rawData))
+		if (empty($data))
 		{
 			throw DataException::forEmptyDataset('insert');
 		}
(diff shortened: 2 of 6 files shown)


Tada5hi:
@jim-parry @lonnieezell would be great if you could merge this to the Developer Branch. Currently im Hard Overriding the BaseResult Class on composer init.

lonnieezell:
Weekend days are typically time with family for me. I'm just stopping by to make a quick comment. I won't be able to merge until sometime later, maybe today, maybe not. 

1. I have no problem with the feature. 
2. You only test success case of the exists, method, not a failure. 
3. You need to provide documentation for the new feature. 
3. That only sets exist flag when an object is retrieved from the database. You also need to set that during an insert to that a call to `insert()` or `save()` on a non-existing entity will now show that it exists (if the save was successful, of course).

system/Model.php
```diff
@@ -642,44 +645,55 @@ public function getInsertID(): int
...
 		}
 
 		// If $data is using a custom class with public or protected
 		// properties representing the table elements, we need to grab
 		// them as an array.
-		if (is_object($data) && ! $data instanceof stdClass)
+		if (is_object($rawData) && ! $rawData instanceof stdClass)
 		{
-			$data = static::classToArray($data, $this->primaryKey, $this->dateFormat, false);
+			$tryUpdate = method_exists($rawData, 'exists') && $rawData->exists();
```
lonnieezell: Why do you feel lit's necessary to check for an update in the middle of the insert method? To me, that just hides errors that a developer needs to address in their app...

tests/system/Database/Live/ModelTest.php
```diff
@@ -1461,6 +1461,44 @@ public function testGetValidationMessagesForReplace()
 
 	//--------------------------------------------------------------------
 
+	public function testSaveNewEntityWithReference()
+	{
+		$entity    = new class extends Entity{
```
lonnieezell: This Entity class is not updated to the [current Entity structure](https://github.com/codeigniter4/CodeIgniter4/pull/2002) and needs to be updated.

system/Model.php
Tada5hi: yeah your complety right, i would just check the exists state in the save method not in the insert method, because only when you use the save method, you want to save an entity and you propaply dont know its state by your own. And when you use insert and update method, your pretty sure you want to update or insert the data set, in case of update by a given key.

Tada5hi:
Yeah no problem ;)
2. fixed it.
3. yeah i will head into this, when checks pass and no changes are required anymore.
4. you'r right.

system/Model.php
```diff
@@ -102,6 +102,13 @@ class Model
 	 */
 	protected $insertID = 0;
 
+	/**
+	 * Indicates if the IDs are auto-incrementing.
+	 *
+	 * @var boolean
+	 */
+	public $incrementing = true;
```
lonnieezell: At the moment the models assume an auto-incrementing id for every table. Please stop trying to revise that. If it's something you truly feel needs addressed again, do so in a separate PR so I don't have to totally kill this one, since they're unrelated. Remove this feature/check.

Tada5hi: you're complety right, i will address this in a seperate pr, im sorry.

Tada5hi:
@lonnieezell hope everything is fine now, i also updated the docs. If anything is missing pls let me know.

user_guide_src/source/models/entities.rst
```diff
@@ -376,3 +376,22 @@ attribute to check::
...
     $user->hasChanged();            // true
+Checking for Existence
+-------------------------------
+
+You can check if an Entity exists before and after its created::
+
+    // Find
+    $user = new User();
+    $user = $userModel->find($id);
+    $user-exists();                 // true
```
MGatner: Dropped part of the arrow! Here (and below) should be `$user->exists();`

tests/system/Database/Live/ModelTest.php
MGatner: @Tada5hi basically drop all the explicit field names and `$_options` and add:
```
protected $dates = ['created_at', 'updated_at', 'deleted_at'];
```

user_guide_src/source/models/entities.rst
Tada5hi: hope its fine now. Please @MGatner let me know if something is missing or needs to be corrected. ^^

MGatner: Looks good!

lonnieezell:
Looks like a test needs an email filled filled out during the test and we should be good to go.

MGatner:
`testSaveNewEntityWithReference()` still has a vestigial `$name` property. There's no need to declare Entity properties anymore, but if you really want to it should be in `$attributes`. 

For example: https://codeigniter4.github.io/CodeIgniter4/models/entities.html#data-mapping

Tada5hi:
everything fine now ?

Tada5hi:
@lonnieezell is here still something to do ? or can this be merged ?

system/Entity.php
```diff
@@ -143,6 +151,28 @@ public function fill(array $data = null)
 
 	//--------------------------------------------------------------------
 
+	/**
+	 * Set boolean value for entity state.
+	 *
+	 * @param boolean $exists
+	 */
+	public function setExists(bool $exists)
```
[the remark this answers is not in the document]
Tada5hi: Yes, because the exist State is controlled outside of the Model object context for example by the model save model if its an Database related entity.

system/Model.php
```diff
@@ -528,22 +528,40 @@ public function save($data): bool
 
 		if (is_object($data) && isset($data->{$this->primaryKey}))
 		{
-			$response = $this->update($data->{$this->primaryKey}, $data);
+			// If primary key is present and the primary key gets auto incremented, than it
```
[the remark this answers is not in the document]
Tada5hi: Indeed. I will resolve that hopefully this evening.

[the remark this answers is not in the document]
Tada5hi: Indeed. I will resolve that hopefully this evening.

jim-parry:
@lonnieezell @MGatner You two have been in the thick of this PR. Should it be merged?

lonnieezell:
Looks like there's a few items that were never resolved. Looking through the logic they've missed the `delete` case, which is the only reason to keep it as class property as far as I can see. If it hasn't been deleted, you can simply check to see whether a value exists for the primaryKey, no need to store. But taking deleted entities into account, you would need to store the state, but no changes were made to the Model to handle the delete state. 

Part of me wonders if this is the wrong place to hold that information with the design patterns we're using here, but I'm not against it. I don't think the implementation is complete, though.

MGatner:
@lonnieezell if Tada5hi is really MIA, do you think this is worth pursuing?

> Part of me wonders if this is the wrong place to hold that information

Tada5hi:
The state change of delete method is missing? Currently the delete method only fulfilled through ids not entity, so should i extend it to also accept entities?

MGatner:
@Tada5hi I think the issue is in the logic of having this field. We can verify if the entity was in the database at the time of insertion or entity creation, but there is no way to guarantee persistence of it since database rows can be removed from a variety of sources, many of which don't use entities. E.g.:
```
$entity = $userModel->find(1);
$db->table('users')->delete(['id' => 1]);
var_dump($entity->exists()); // true
```

Simplistic, but also believable enough that I don't think `exists()` would ever be reliable. I do like the concept though - what about instead of "exists" you change it to "inserted" or "created"? It could have all the same logic and functions, but would be more clear that it had no consistent state value after the database load or insert.

lonnieezell:
What's the consensus on this one? Do we want to change the name? Drop the feature? 

I'm ok with either. It adds a bit of complexity to an already complex set of rules. In my personal work I'd be ok assuming that if an entity has a value for `$primaryKey` than it exists. I think that's probably all confirmation I'd need at that point, honestly.

MGatner:
@lonnieezell I like using the primary key as the indicator for database existence. If @Tada5hi feels strongly about this feature I'm okay with it (assuming the renaming discussed above) but I think this could also be handled with a simple User Guide addition noting the behavior.

Tada5hi:
@lonnieezell @MGatner i would head into this these days,  i was busy over christmas days. I think it's necessary because you can't expect that an object always has a primary which is auto incremented by the dbms and is never set manually or is a composite primary key.

lonnieezell:
No updates forthcoming. Closing.

PR #2049, opened 2019-06-10
Update CURLRequest.php

nowackipawel:
aa362d0995eb08d2d6584f08ff7c0e892fa021b7

prvious fix caused eception if debug has not been specified
Type:        ErrorException
Message:     curl_setopt_array(): supplied argument is not a valid File-Handle resource
Filename:    /home/-exchange/site/system/HTTP/CURLRequest.php

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -674,8 +674,16 @@ class CURLRequest extends Request
 		// Debug
 		if (isset($config['debug']))
 		{
-			$curl_options[CURLOPT_VERBOSE] = $config['debug'] === true ? 1 : 0;
-			$curl_options[CURLOPT_STDERR]  = $config['debug'] === true ? fopen('php://output', 'w+') : $config['debug'];
+			if(is_bool($config['debug']) && $config['debug'] === true)
+			{
+				$curl_options[CURLOPT_VERBOSE] = 1;
+				$curl_options[CURLOPT_STDERR] = fopen('php://output', 'w+');
+			}
+			else if(is_resource($config['debug']))
+			{
+				$curl_options[CURLOPT_VERBOSE] = 1;
+				$curl_options[CURLOPT_STDERR] = $config['debug'];
+			}
 		}
 
 		// Decode Content


nowackipawel:
@lonnieezell : I think test "testDebugOption" is written wrongly. "

CURLOPT_STDERR | An alternative location to output errors to instead of             STDERR.

IMHO CURLOPT_STDERR should be set to file handle or should not exists.

nowackipawel:
@lonnieezell  could you answer?

system/HTTP/CURLRequest.php
```diff
@@ -674,8 +674,16 @@ protected function setCURLOptions(array $curl_options = [], array $config = [])
...
 		if (isset($config['debug']))
 		{
-			$curl_options[CURLOPT_VERBOSE] = $config['debug'] === true ? 1 : 0;
-			$curl_options[CURLOPT_STDERR]  = $config['debug'] === true ? fopen('php://output', 'w+') : $config['debug'];
+			if(is_bool($config['debug']) && $config['debug'] === true)
+			{
+				$curl_options[CURLOPT_VERBOSE] = 1;
+				$curl_options[CURLOPT_STDERR] = fopen('php://output', 'w+');
+			}
+			else if(is_resource($config['debug']))
```
lonnieezell: I'm confused. When is `$config['debug']` going to be a resource? According to the docs it should only ever be a boolean. If true, it turns on verbose and sets the error output, otherwise it leaves verbose to whatever it was otherwise set at.

lonnieezell:
While it may crash if debug hasn't been specified, what you're changing it to doesn't work. See my comment for explanation.

nowackipawel:
CURLOPT_STDERR should be a resource or nothng. in other case this will
cause exceptions.


czw., 20 cze 2019 o 06:47 Lonnie Ezell <notifications@github.com>
napisał(a):

lonnieezell:
@MGatner yes, you're right. My apologies to @nowackipawel 

We should allow a file name to be specified - so convert it to a resource behind the scenes I guess. I don't think it needs to check the path. If it doesn't exist it should create it. If it does exist, and is writable, then it would get overwritten. If it can't be written to (or outside safe_mode boundaries) then it should naturally throw an exception. I would think that exception would be clear enough about what happened and why.

PR #2051, opened 2019-06-10
Soft deletes use deleted_at

MGatner:
**Description**
First attempt. Change all references to `deleted` to use `deleted_at` instead. Direct references to these fields now use dates, and tests check `IS NULL` and `IS NOT NULL` instead of binary values.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -713,16 +713,20 @@ class Model
 		$result = $this->builder()
 				->set($data['data'], '', $escape)
 				->insert();
-
+		
+		// If insertion succeeded then save the insert ID
+		if ($result)
+		{
+			$this->insertID = $this->db->insertID();
+		}
+		
 		$this->trigger('afterInsert', ['data' => $originalData, 'result' => $result]);
 
-		// If insertion failed, get our of here
+		// If insertion failed, get out of here
 		if (! $result)
 		{
 			return $result;
 		}
-
-		$this->insertID = $this->db->insertID();
 
 		// otherwise return the insertID, if requested.
 		return $returnID ? $this->insertID : $result;
--- a/system/Model.php
+++ b/system/Model.php
@@ -120,7 +120,7 @@ class Model
 
 	/**
 	 * If this model should use "softDeletes" and
-	 * simply set a flag when rows are deleted, or
+	 * simply set a date when rows are deleted, or
 	 * do hard deletes.
 	 *
 	 * @var boolean
@@ -182,7 +182,7 @@ class Model
 	 *
 	 * @var string
 	 */
-	protected $deletedField = 'deleted';
+	protected $deletedField = 'deleted_at';
 
 	/**
 	 * Used by asArray and asObject to provide
@@ -356,7 +356,7 @@ class Model
 
 		if ($this->tempUseSoftDeletes === true)
 		{
-			$builder->where($this->table . '.' . $this->deletedField, 0);
+			$builder->where($this->table . '.' . $this->deletedField, null);
 		}
 
 		if (is_array($id))
@@ -427,7 +427,7 @@ class Model
 
 		if ($this->tempUseSoftDeletes === true)
 		{
-			$builder->where($this->table . '.' . $this->deletedField, 0);
+			$builder->where($this->table . '.' . $this->deletedField, null);
 		}
 
 		$row = $builder->limit($limit, $offset)
@@ -457,7 +457,7 @@ class Model
 
 		if ($this->tempUseSoftDeletes === true)
 		{
-			$builder->where($this->table . '.' . $this->deletedField, 0);
+			$builder->where($this->table . '.' . $this->deletedField, null);
 		}
 
 		// Some databases, like PostgreSQL, need order
@@ -912,7 +912,7 @@ class Model
 
 		if ($this->useSoftDeletes && ! $purge)
 		{
-			$set[$this->deletedField] = 1;
+			$set[$this->deletedField] = $this->setDate();
 
 			if ($this->useTimestamps && ! empty($this->updatedField))
 			{
@@ -947,7 +947,7 @@ class Model
 		}
 
 		return $this->builder()
-						->where($this->deletedField, 1)
+						->where($this->deletedField . ' IS NOT NULL')
 						->delete();
 	}
 
@@ -981,7 +981,7 @@ class Model
 		$this->tempUseSoftDeletes = false;
 
 		$this->builder()
-				->where($this->deletedField, 1);
+				->where($this->deletedField . ' IS NOT NULL');
 
 		return $this;
 	}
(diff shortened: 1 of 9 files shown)


MGatner:
@lonnieezell I might need some help. I think everything is correct now but I replaced `'deleted` => 0` with `'deleted_at' IS NULL` (and their opposites) and the PostGres handler doesn't seem to be interpreting the string as a query fragment (https://travis-ci.org/codeigniter4/CodeIgniter4/jobs/543886513). I *think* it is an issue in `seeInDatabase()` and not the framework's `whereHaving()` but I'm not familiar with why PostGres is taking this as an int (or maybe expecting an int?). The `users` table is definitely defining date fields as datetimes, just mock `events` has the int versions of them so I don't think it is related.

MGatner:
Hmm in retrospect I wonder if it would except `'deleted_at' => null` as a valid criterium and translate it? I'll wait for your response but that might be something to test next.

system/Model.php
```diff
@@ -908,7 +912,7 @@ public function delete($id = null, bool $purge = false)
 
 		if ($this->useSoftDeletes && ! $purge)
 		{
-			$set[$this->deletedField] = 1;
+			$set[$this->deletedField] = $this->setDate();
```
lonnieezell: This should be wrapped inside of a `if ($this->useTimestamps)` similar to the `updatedField` just below. You can wrap them together if you need to.

system/Model.php
```diff
@@ -943,7 +947,7 @@ public function purgeDeleted()
 		}
 
 		return $this->builder()
-						->where($this->deletedField, 1)
+						->where($this->deletedField . ' IS NOT NULL')
```
lonnieezell: Boy, I feel like we need some `whereNull()`, `whereNotNull()` methods in the future, but I just checked Mysql, Sqlite, and Postgres and they all share the same syntax so I think we're good here for now.

user_guide_src/source/models/model.rst
```diff
@@ -137,13 +138,15 @@ method.
...
 
-If true, then any delete* method calls will simply set a flag in the database, instead of
+If true, then any delete* method calls will set ``deleted_at`` in the database, instead of
 actually deleting the row. This can preserve data when it might be referenced elsewhere, or
 can maintain a "recycle bin" of objects that can be restored, or even simply preserve it as
 part of a security trail. If true, the find* methods will only return non-deleted rows, unless
 the withDeleted() method is called prior to calling the find* method.
 
-This requires an INT or TINYINT field to be present in the table for storing state. The default field name is  ``deleted`` however this name can be configured to any name of your choice by using $deletedField property.
+This requires a DATETIME field to be present in the table for storing state. The default field
```
lonnieezell: Two things: 

1. requires either a datetime or integer field in the database as per their model's $dateFormat setting.

Okay - only one thing. :)

system/Model.php
MGatner: I think we need to set `deleted_at` regardless of whether `useTimestamps` is toggled. Though that brings up an interesting scenario of what to do if `useSoftDeletes` is true and `useTimestamps` is false, since `setDate` relies on `dateFormat` which is tied to `useTimestamps`.

system/Model.php
MGatner: Real feels a bit awkward having raw SQL in the query field. I'd be in favor of those functions as well.

user_guide_src/source/models/model.rst
MGatner: Added!

system/Model.php
lonnieezell: Good point. Deleting shouldn't be dependent on that, so ignore my comment here. :) 

I think the times when `useSoftDeletes` is true, `useTimestamps` is false, and `dateFormat` is empty should be extremely rare, since `dateFormat` has a default in the Model. It wouldn't hurt to check for that condition and throw an exception with a explantation that `dateFormat` cannot be empty.

MGatner: Thinking about this during dinner, I believe that `dateFormat` should be a requirement (throw if missing) for `useSoftDeletes` and `useTimestamps` but those two should not otherwise be related - that helps keep it clean and only check what is absolutely necessary. If you agree I'll resolve this convo and plan those sanity checks for a separate PR.

lonnieezell: Absolutely agree with that. Sounds good.

MGatner:
IMO this merge would merit a blog entry of some kind, similar to the entity refactor, as it will require developers to intervene with database changes upon upgrading.

user_guide_src/source/models/model.rst
```diff
@@ -144,9 +144,9 @@ can maintain a "recycle bin" of objects that can be restored, or even simply pre
 part of a security trail. If true, the find* methods will only return non-deleted rows, unless
 the withDeleted() method is called prior to calling the find* method.
 
-This requires a DATETIME field to be present in the table for storing state. The default field
-name is  ``deleted_at`` however this name can be configured to any name of your choice by using
-$deletedField property.
+This requires requires either a DATETIME or INTEGER field in the database as per the model's
```
lonnieezell: Double `requires` here. Only need one :)

MGatner: Is that a risk you're willing to take?

lonnieezell: @MGatner I'm not sure I follow - but yes having it say `This requires either` instead of `This requires requires either` is a risk I'm willing to take. :)

MGatner: :) just joking. It's all set, if everything looks good to you!

system/Model.php
lonnieezell: Just noticed - this should have a table prefix before the field.

system/Model.php
```diff
@@ -977,7 +981,7 @@ public function onlyDeleted()
 		$this->tempUseSoftDeletes = false;
 
 		$this->builder()
-				->where($this->deletedField, 1);
+				->where($this->deletedField . ' IS NOT NULL');
```
lonnieezell: Also needs table prefix.

system/Model.php
```diff
@@ -1539,7 +1543,7 @@ public function countAllResults(bool $reset = true, bool $test = false)
 	{
 		if ($this->tempUseSoftDeletes === true)
 		{
-			$this->builder()->where($this->deletedField, 0);
+			$this->builder()->where($this->deletedField, null);
```
lonnieezell: Also needs table prefix.

MGatner: Got those!

lonnieezell:
Well that was quick! :) Looks good. Thanks for all of this!

MGatner:
I can't tell you how glad I will be to remove all my explicit code for `deleted_at` fields on *everything*.

lonnieezell:
> I can't tell you how glad I will be to remove all my explicit code for deleted_at fields on everything.

Well - glad I could help create more work for you and have you be happy about it :)

MGatner:
Haha well more work now for less work later, my kinda work!

PR #2052, opened 2019-06-11
Add various tests for (not) null

MGatner:
**Description**
While working on implementing `deleted_at` I noticed there aren't good explicit tests for `IS NULL` and `IS NOT NULL`. This change provides these tests (hopefully in the right place).

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


lonnieezell:
Nice catch. Thanks!

MGatner:
These can be changed to implement `whereNull` and `whereNotNull` whenever they are brought to life.

PR #2057, opened 2019-06-14
Cache Driver Backups

MohKari:
**Description**

If our application is unable to connect to Redis, a CriticalError exception is now thrown.

CacheFactory catches the CriticalError exceptions that are thrown when Redis, Memcache and Predis fail to connect, it logs the exception and then attempts to initialize the backup cache driver (and if that fails it initializes dummy).

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


MohKari:
Hi, just some additional information regarding the pull request.

It's my first time attempting to contribute to an open source project, so apologies if i have missed anything.

I noticed, what i would call an issue with the current version of CI4, when i purposely mis-configured the connection to my Redis server to simulate what would happen if my Redis server went down, the issue was that the whole application would just stop, due to a RedisException been thrown.

I felt that this should be improved, and if the primary cache doesn't work, we should attempt to use the secondary cache, and if that doesn't work, don't use any cache. I feel like this is a better solution than the application no longer working until someone is able to fix the cache.

lonnieezell:
It sounded like a good addition. Why did you close it? I am out of town and haven’t had a chance to review it yet.

MohKari:
Insecurity of my abilities...

I've been spending the majority of today (4 hours so far) trying to figure out what is going on with the build (I've never used CI before, so I'm learning a lot which I'm quite enjoying).

My current plan is to get a build that passes the tests and also be able to tick off the checklist, then to reopen the pull request.

Issue #2055, opened 2019-06-12
Feature Request: Centralized loggedInUser before RC


MGatner:
The decision not to ship CI4 with an authentication library leaves a big gap for developers wanting to interact with "the current user", as how that is defined may vary drastically between libraries. What they should all have in common is storing a logged-in user in the session.

I proposed (https://forum.codeigniter.com/thread-73257.html) a framework-sanctioned location for this so different libraries could all depend on the same key (e.g. Myth:Auth uses `$_SESSION['logged_in']`). This could be set directly, with a service, or triggered by a listener for a "login" event, and likewise accessed via a service or directly from `session()`.

For the sake of circumventing a myriad of different ways of solving this I would love to see this feature implemented prior to RC1. I am glad to code it if the team decides the policy and guiding principles.

chistel:
This would be a great idea. But year it has to be flexible like laravel's auth to support multiple authentication

lonnieezell:
One problem I see with your code example is that it doesn't work for stateless authentication, like JWT or other common API auth patterns. 

Also - adding a `user()` method to the Session, while convenient, doesn't apply well with the Single Responsibility Pattern since grabbing a current user really doesn't have anything to do with the session. 

Unfortunately, I don't think there's a catch-all solution here due to stateless authentication. Perhaps a better pattern would be to have all libraries accept a user instance, say, though a `setUser()` method. While it's not as convenient, it removes the need for the libraries to know anything about the authentication system used, which is a good thing....

MGatner:
I chose to put it in Session because I believe Session will need to be involved no matter what, being the only place where variables 1) persist across app calls, and 2) are user-specific. Stateless does make it a bit weird to use Session but it would still function, right? I think the alternative (and my original idea) gets more complex: having a User class and service that looks for a local property first, then falls back to Session (conditionally).

The problem with user dependency injection for third-party libraries is it precludes zero-conf packages as it requires the developer to initialize the library explicitly. Defining the "current user" is the only main hurdle I see to enabling zero-conf packages and why I keep pushing for it (alternative to a fully-implemented authentication class, of course).

Paradinight:
I'm against it. One user class is not enought. Most application need more than isLoggedIn :)

I do not see any benefit.

MGatner:
Just to be clear it's more than boolean isLoggedIn, it's *who* is logged in. Applications will definitely still need full auth-entication/-orization but this fills a crucial gap. An example:

[...]

PR #2058, opened 2019-06-17
Centralized user identification

MGatner:
**Description**
This is my example of how I would like to see CI4 have an "official" location for logged-in users so that third-party libraries can consistently and reliably set and access "current user".

*See also* conversation on the forums (https://forum.codeigniter.com/thread-73257.html) and the related issue (https://github.com/codeigniter4/CodeIgniter4/issues/2055)

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/Events.php
+++ b/app/Config/Events.php
@@ -41,3 +41,19 @@ Events::on('pre_system', function () {
 		Services::toolbar()->respond();
 	}
 });
+
+/*
+ * --------------------------------------------------------------------
+ * User Authentication Listeners.
+ * --------------------------------------------------------------------
+ * Hooks provided for external authentication modules. These may be
+ * removed if your application does not authenticate.
+ */
+
+Events::on('login', function ($userId) {
+	Services::session()->user($userId);
+});
+
+Events::on('logout', function () {
+	Services::session()->user(false);
+});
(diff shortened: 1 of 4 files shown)


dafriend:
I don't mean to be harsh, so please don't take this a hostile. This is a bad idea mostly because it is a serious breach of Separation of Concerns. The Session class does not, indeed, should not know anything about Users.

MGatner:
@dafriend no worries, glad to have some feedback! This was one take at a solution to https://github.com/codeigniter4/CodeIgniter4/issues/2055 (where there has been some discussion this morning also), but certainly not the only way it could be solved. And I am way more interested in having a *working solution* than *my solution*. I do expect that the Session class will need to be involved, since most authentication libraries will "login" once and then rely on persistence across page loads - but situating it elsewhere (or even its own class) would be just as suitable.

Since you seem to have your head around framework responsibilities, do you have an idea where some "facilitated information" like this should go?

HieuPT7:
What if my website allows multiple users to log in at once?

MGatner:
This PR is superseded by https://github.com/codeigniter4/CodeIgniter4/pull/2072, but both of these handle user identification on a per-load/-session basis so would be specific to one user and work fine across as many users as you want.

vibbow:
For my understand, you are bind this user feature with session (cookie session).

This will limit this feature to full stack web application only, since most API application didn't use cookie session at all.

PR #2060, opened 2019-06-18
Cache Drive Backups 

MohKari:
**Description**
If our application is unable to connect to Redis, a CriticalError exception is now thrown.

CacheFactory catches the CriticalError exceptions that are thrown when Redis, Memcache and Predis fail to connect, it logs the exception and then attempts to initialize the backup cache driver (and if that fails it initializes dummy).

**Checklist:**
- [ ] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Cache/CacheFactory.php
+++ b/system/Cache/CacheFactory.php
@@ -39,6 +39,7 @@
 namespace CodeIgniter\Cache;
 
 use CodeIgniter\Cache\Exceptions\CacheException;
+use CodeIgniter\Exceptions\CriticalError;
 
 /**
  * Class Cache
@@ -93,7 +94,20 @@ class CacheFactory
 			}
 		}
 
-		$adapter->initialize();
+		// If $adapter->initialization throws a CriticalError exception, we will attempt to
+		// use the $backup handler, if that also fails, we resort to the dummy handler.
+		try
+		{
+			$adapter->initialize();
+		}
+		catch (CriticalError $e)
+		{
+			// log the fact that an exception occurred as well what handler we resorting to
+			log_message('critical', $e->getMessage() . ' Resorting to using ' . $backup . ' handler.');
+
+			// get the next best cache handler (or dummy if the $backup also fails)
+			$adapter = self::getHandler($config, $backup, 'dummy');
+		}
 
 		return $adapter;
 	}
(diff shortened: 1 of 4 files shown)


MohKari:
Was originally pull request https://github.com/codeigniter4/CodeIgniter4/pull/2057, but i had some issues... So here it is again (Working this time hopefully).

system/Cache/CacheFactory.php
```diff
@@ -93,7 +94,20 @@ public static function getHandler($config, string $handler = null, string $backu
...
+		{
+			$adapter->initialize();
+		}
+		catch (CriticalError $e)
+		{
+			// log the fact that an exception occurred as well what handler we resorting to
+			log_message('critical', $e->getMessage() . ' Resorting to using ' . $backup . ' handler.');
+
+			// get the next best cache handler (or dummy if the $backup also fails)
+			$adapter = self::getHandler($config, $backup, 'dummy');
```
lonnieezell: Doesn't the adapter need initialized here once we grab a new one?

lonnieezell:
I like the idea of this a lot. I made one comment on the code, but could you add the same process to the other drivers?

system/Cache/CacheFactory.php
MohKari: Line 109 returns an adapter that has already been initialized.

MohKari:
@lonnieezell I've attempted to implement a similar process for the other drives as requested.

The DummyHander, FileHander and WincacheHandler's initialize() method contain no logic, so I have ignored these.

I modelled my solution on how PredisHandler initialize() method, so I've just added some comments there.

I've changed MemcachedHandler initialize() a fair bit, I've broken it down into two sections contained within the if/elseif block.

I've tested the Redis, Predis, Memcached drivers and I'm happy with how they work (If they can't connect to the cache server, they throw a CriticalError exception). I'm unable to test the Memcache driver (I can't seem to get it to work with PHP 7+).

I'm confident with the code I've written to check if the cache server is contactable when using Memcache, I followed https://www.php.net/manual/en/memcache.connect.php, which suggests that if connect() returns false, the sever can not be contacted, which will trigger our CriticalError exception.

Is it worth performing similar checks on the Handlers I've ignored? Or does the isSupported() method handle it?

lonnieezell:
I believe isSupported should be sufficient, but it's been a while since I've been in those handlers, honestly. 

Thanks for this! 

In the future, though, we would prefer to have signed commits. There's instructions [in the docs](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/signing.rst) to help with that.

Issue #1866, opened 2019-03-24
FileLocator->getNamespaces with parameter


MGatner:
I'm working on an updated MigrationRunner that uses FileLocator to support external PSR4 sources. The `getNamespaces($prefix)` method flattens the return from `$autoloader->getNamespace($prefix)` to a single path, when it is actually an array of paths thus potentially causing `getNamespaces()` to fail to search some valid paths. AFAIK this method is only used by `locateFile()` so is probably safe to rewrite, but seeking advice first.

lonnieezell:
I see 4 uses of it, but all of them are functions within the FileLocator class so you should be fine to modify that as long as all of those other methods don't break: 

- findQualifiedNameFromPath
- listFiles
- locateFiles
- search

atishamte:
@MGatner, any update on this?

MGatner:
I haven't looked into it at all

MGatner:
Looked at this some more, definitely a potential issue. Some packages (like "phpDocumentor\Reflection") define multiple paths for one namespace:

```
["phpDocumentor\Reflection"]=>
  array(3) {
    [0]=>
    string(84) "/var/www/vhosts/.../vendor/composer/../phpdocumentor/reflection-common/src"
    [1]=>
    string(86) "/var/www/vhosts/.../vendor/composer/../phpdocumentor/reflection-docblock/src"
    [2]=>
    string(80) "/var/www/vhosts/.../vendor/composer/../phpdocumentor/type-resolver/src"
  }
```

Using `$locator->locateFile()` will only search the first of these ('reflection-common') and thus fail to match any files in the other paths.

MGatner:
Also, namespaces aren't required to start with a capital ("phpDocumentor" above is a good example again) so this will fail to match sometimes:

https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Autoloader/FileLocator.php#L118

MGatner:
Close by #2059

PR #2064, opened 2019-06-20
Bugfix file locator slash error

MGatner:
**Description**
https://github.com/codeigniter4/CodeIgniter4/pull/2059 introduced a scenario where the combination of `$path` and `$folder` could fail to have a separating slash. This PR moves the trailing slash safety check prior to the folder test.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -144,7 +144,7 @@ class FileLocator
 			// or 'libraries'.
 			if (! empty($folder) && strpos($path . $filename, '/' . $folder . '/') === false)
 			{
-				$path .= $folder;
+				$path .= trim($folder, '/') . '/';
 			}
 
 			$path .= $filename;
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -136,6 +136,9 @@ class FileLocator
 		// Check each path in the namespace
 		foreach ($paths as $path)
 		{
+			// Ensure trailing slash
+			$path = rtrim($path, '/') . '/';
+			
 			// If we have a folder name, then the calling function
 			// expects this file to be within that folder, like 'Views',
 			// or 'libraries'.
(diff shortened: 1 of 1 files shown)


MGatner:
Okay I think this is all set now. So many slash woes!

system/Autoloader/FileLocator.php
```diff
@@ -136,15 +136,18 @@ public function locateFile(string $file, string $folder = null, string $ext = 'p
...
+			// Ensure trailing slash
+			$path = rtrim($path, '/') . '/';
+			
 			// If we have a folder name, then the calling function
 			// expects this file to be within that folder, like 'Views',
 			// or 'libraries'.
 			if (! empty($folder) && strpos($path . $filename, '/' . $folder . '/') === false)
 			{
-				$path .= $folder;
+				$path .= trim($folder, '/') . '/';
```
lonnieezell: Should this be an `rtrim`? We only care about the trailing slash, right?

MGatner: Conceivably `$folder` could have a preceding slash. We could let that be a user input error and go with `rtrim()`, or stick with this. At this point `$path` is guaranteed a trailing slash.

lonnieezell: Hmm... will have to look later and see where all it's being used in the system, but I can imagine possibilities where someone wants to reference a full unix path, which would have a preceding slash that would need to stay.

MGatner: `$folder` is by definition a path within an existing namespace so will always be appended to `$path` (which has the trailing slash) so I think the only thing a preceding slash would get you is doubled slashes, which I'm pretty sure *nix systems treat as a single slash anyways.

MGatner: Also per your comment on the original issue (https://github.com/codeigniter4/CodeIgniter4/issues/1866) `locateFile` is only used internal to FileLocator, so should be fine.

lonnieezell: Ah, thanks for doing the research for me! Wasn't going to have time until tonight. So I'm cool with this then :)

PR #2067, opened 2019-06-24
WIP cleanValidationRules turned off for insert

nowackipawel:
As for now, there was an issue with required files in INSERT
```
		'uvf_dob'					=> 'required|valid_date',
```
if uvf_dob is not exists in data posted to insert then required rule is ignored cuz it is deleted by cleanValidationRules()
the same was happend with uploaded[file] rule;
I think at least for insert cleanValidationRules should not remove rules with required or uploaded

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1390,7 +1390,7 @@ class Model
 			$rules = $this->validation->loadRuleGroup($rules);
 		}
 
-		$rules = $this->cleanValidationRules($rules, $data, $cleanRequired);
+		$rules = $cleanRequired ? $this->cleanValidationRules($rules, $data) : $rules;
 
 		// If no data existed that needs validation
 		// our job is done here.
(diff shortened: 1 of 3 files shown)


nowackipawel:
It could break some tests and it'll require to make changes in test cases, but I am sure for insert cleaning rules should be turned off.
I am not sure about uploaded (FILE) rule if it should be cleared or not...during update/replace

nowackipawel:
@lonnieezell @jim-parry  could you check this useful PR?

jim-parry:
This breaks tests/system/Database/Live/ModelTest :(

nowackipawel:
@jim-parry : Yes, I excepted that. However, IMHO it should work that way. When we are trying to insert data to DB every role should be checked; cleaning validation rules is useful and desirable in update operation only.

MohKari:
I just came across the same issue...

I want to ensure that certain fields always exist during an insert. If they don't a validation error should occur... but the `$this->cleanValidationRules` just removes those rules, I can understand how removing some of the rules makes sense on update, however I still feel like any rule that has a 'required' should be required, rather than getting removed...

Perhaps we could introduce a 'required_on_insert' / 'required_on_update' ?

nowackipawel:
@lonnieezell: tests are passing, I've added  one additional

lonnieezell:
We've had an added parameter to some of the methods it looks like. Revisited this without any additional arguments in https://github.com/codeigniter4/CodeIgniter4/pull/2433

PR #2068, opened 2019-06-24
getExtension() should return string

nowackipawel:
in other case this cause i.e. exception during validation if file is not provided
```
'is_image[upload_' . $i . ']|max_size[upload_' . $i . ',6144]'
```
I thought to declare this method as `public function getExtension(): ?string` but this will be worse option.


MGatner:
This is resolved in a slightly different manner by #2128

lonnieezell:
#2128 already merged.

PR #2070, opened 2019-06-24
WIP Checking file rules for existing files

nowackipawel:
max_size, max_dim, is_image, ext_in, mime_in should be checked only if file is uploaded, in other case... rules require file to be available - it is not necessary as long as we have `uploaded` rule.

### DIFF
--- a/system/Validation/FileRules.php
+++ b/system/Validation/FileRules.php
@@ -122,6 +122,11 @@ class FileRules
 
 		$file = $this->request->getFile($name);
 
+		if($file->getError() === UPLOAD_ERR_NO_FILE)
+		{
+			return true;
+		}
+
 		if (is_null($file))
 		{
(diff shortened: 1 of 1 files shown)


nowackipawel:
Imho test should be corrected as long as we have `uploaded` rule which should be used in conjunction with other methods as stated above.

system/Validation/FileRules.php
```diff
@@ -122,6 +122,11 @@ public function max_size(string $blank = null, string $params, array $data): boo
 
 		$file = $this->request->getFile($name);
 
+		if($file->getError() === UPLOAD_ERR_NO_FILE)
```
jim-parry: Shouldn't this test follow the is_null($file) block?
The order you have it breaks travis-ci, and it looks like rightfully so :(

PR #2072, opened 2019-06-24
WIP Generic User class with service and listeners

MGatner:
**Description**
This is a continuation of the conversation around the interest for a centralized "current user" (https://github.com/codeigniter4/CodeIgniter4/issues/2055). This is a concept implementation of how the framework could function as an arbiter between auth libraries and third-party modules for setting and accessing the "current user".

I'm sure there is a lot more that could be done with this class, but in the interest of leaving CI4 are of authentication material I think keeping it simple and straightforward (including not defining `$user`'s type) leaves the most options.

*See also https://github.com/codeigniter4/CodeIgniter4/pull/2058

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/Events.php
+++ b/app/Config/Events.php
@@ -41,3 +41,20 @@ Events::on('pre_system', function () {
 		Services::toolbar()->respond();
 	}
 });
+
+/*
+ * --------------------------------------------------------------------
+ * User Authentication Listeners.
+ * --------------------------------------------------------------------
+ * Hooks provided for external authentication modules. These may be
+ * removed if your application does not authenticate.
+ */
+
+Events::on('login', function ($user) {
+	Services::user($user);
+});
+
+Events::on('logout', function () {
+	Services::user()->remove();
+});
+
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -814,6 +814,27 @@ class Services extends BaseService
 	//--------------------------------------------------------------------
 
 	/**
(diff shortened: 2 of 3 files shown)


jim-parry:
I think this is a Pandora's box ... while the notion is sensible, I wouldn't want to see this constrain what we might want to do down the road. I think that the issue of authentication & authorization needs to be carefully thought out, with a support mechanism that will play nice with any of the implementations.
It might inconvenience you with 4.0.0, but that can be overcome by adding user() to App\Config]Services, and/or overriding Services::session(), no?
I would be more comfortable hammering this out among the community for 4.1, rather than putting this into 4.0 as a "better than nothing" solution which would likely have to change for 4.1
... My two bits worth

MGatner:
Thanks for the feedback! It is a sticky one. My main interest is for having a predictable way for a module to access a logged in user (of any) rather than any sort of authentication or authorization. I fear that not having something in place for 4.0.0 will miss an opportunity to centralize, as people will all start grabbing or writing auth modules that each store this in their own different way. 

As a module writer, I would name this as my biggest consistent hurdle when developing “addins” that users can Composer plug-and-play.

lonnieezell:
We can talk more later about this, but this is not in the cards at the current time.

MGatner:
Let’s talk. I have some other ideas, and definitely think this PR is not the right direction anymore. This feature is still my number one request for the framework though!

PR #2073, opened 2019-06-24
WIP Issue with casting afer fill.

nowackipawel:
In code like this:
```

					$tmp = (new $this->returnType());
					$tmp->fill($data);

					$this->insert($tmp);
```
i've got an issue related to valid_json cuz array has not been casted to json string.
my change allows to force fill method to cast properties according to declaration from $cast property.

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -113,9 +113,11 @@ class Entity
 	 *
 	 * @param array $data
 	 *
+	 * @param bool $casted
 	 * @return \CodeIgniter\Entity
+	 * @throws \Exception
 	 */
-	public function fill(array $data = null)
+	public function fill(array $data = null, bool $casted = false)
 	{
 		if (! is_array($data))
 		{
@@ -134,7 +136,7 @@ class Entity
 			}
 			else
 			{
-				$this->attributes[$key] = $value;
+				$casted ? $this->__set($key, $value) : $this->attributes[$key] = $value;
 			}
 		}
 


lonnieezell:
I'm not against this change - but at the very least it needs to be documented. Should also have tests.

nowackipawel:
@lonnieezell could so help me with tests/docs?

jim-parry:
Your PR alludes to a JSON casting problem, but I don't see anything JSON in the description or PR changes. Your code change appears to use the undescribed "$casted" parameter to either set object attributes in an entity, or else array elements in an entity's "$attributes" array. The connection to JSON isn't clear.

Apart from the confusing change, it sounds like you still have issues with testing and docs. 
The user guide has a description for [using Vagrant](https://codeigniter4.github.io/userguide/installation/running.html?highlight=vagrant#hosting-with-vagrant) to host your app for testing, and it sets up unit testing too IIRC. Not sure if it is setup for building the documentation, but that could hopefully be added if necessary. Would something like this help you get a test environment working?

MGatner:
This is appears to introduce a way to skip casting yet still map key names and use `setXXXX()` methods. **Entity.php** already has the `setAttributes()` method which allows directly applying attribute values without intervention, so in my opinion this PR offers a very niche and largely unnecessary change. I'm closing it for now but if you really believe there is a case for it please resubmit with User Docs, tests, and a case for this niche functionality.

PR #2074, opened 2019-06-24
Throttler can access bucket for bucket life time

MohKari:
**Description**
I was encountering an issue with the rate limiting which was causing it not to behave as expected.

If i set a rate limit of 1 request a minute, I would be able to make much more than 1 request a minute provided that the requests are a few seconds apart.

This is because the original "bucket"'s life time gets updated to the time between the 1st and 2nd request, meaning the 3rd request typically creates a new "bucket".

**Edit**
Added an additional commit to resolve an additional issue regarding calculating how many seconds must pass before a new token is available. (https://github.com/codeigniter4/CodeIgniter4/pull/2074/commits/8e0ab844be5889a8703b23fcd9dc9caea6cddb70)

Added an additional commit to update ThrottleTest to getTokenTime() to have a value greater than 1 (https://github.com/codeigniter4/CodeIgniter4/pull/2074/commits/1eac5c1ba5e6dabf95fe1eb65cc6fab8628e464f)

**Checklist:**
- [ ] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Throttle/Throttler.php
+++ b/system/Throttle/Throttler.php
@@ -143,7 +143,7 @@ class Throttler implements ThrottlerInterface
 			// If it hasn't been created, then we'll set it to the maximum
 			// capacity - 1, and save it to the cache.
 			$this->cache->save($tokenName, $capacity - $cost, $seconds);
-			$this->cache->save($tokenName . 'Time', time());
+			$this->cache->save($tokenName . 'Time', time(), $seconds);
 
 			return true;
 		}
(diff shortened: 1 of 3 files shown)


MohKari:
There is a `if ($tokens > 0)` check performed. This seems to allow a sort of rolling rate limit.

If rate limiting is set to 1 request a minute, you could perform 2 requests back to back, the 3rd request would trigger the "rate limit reached". 

Is this intended behaviour? It is easily fixed by changing `if ($tokens > 0)` to `if ($tokens >= 1)`.

It just seems a bit odd to me to taken the token pool below 0.

MohKari:
Possibly worth noting that if the rate limit is set to 1. You typically get a new bucket by the time you can make a new request, meaning you can make 2 requests every minute. This could be fixed by doubling the amount of time the "buckets" remain in cache.

I'm unsure what kind of effect this would have on performance, which is why i have not added it in myself, which I'm happy to do.

PR #2083, opened 2019-07-02
WIP Update Validation.php

MohKari:
**Description**
Was experiencing issue where i wanted to display the supplied value within the validation error message. So I've updated the validation class to pass the value through to the error message generation and to replace {value} with the supplied value.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -208,11 +208,11 @@ class Validation implements ValidationInterface
 	 * the error to $this->errors and moves on to the next,
 	 * so that we can collect all of the first errors.
 	 *
-	 * @param string      $field
-	 * @param string|null $label
-	 * @param string      $value
-	 * @param array|null  $rules
-	 * @param array       $data  // All of the fields to check.
+	 * @param string       $field
+	 * @param string|null  $label
+	 * @param string|array $value Value to be validated, can be a string or an array
+	 * @param array|null   $rules
+	 * @param array        $data  // All of the fields to check.
 	 *
 	 * @return boolean
 	 */
@@ -291,7 +291,14 @@ class Validation implements ValidationInterface
 			// Set the error message if we didn't survive.
 			if ($passed === false)
 			{
-				$this->errors[$field] = is_null($error) ? $this->getErrorMessage($rule, $field, $label, $param) : $error;
+				// if the $value is an array, convert it to as string representation
+				if (is_array($value))
+				{
+					$value = '[' . implode(', ', $value) . ']';
+				}
+
+				$this->errors[$field] = is_null($error) ? $this->getErrorMessage($rule, $field, $label, $param, $value)
+					: $error;
 
 				return false;
 			}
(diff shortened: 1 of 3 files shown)


MohKari:
I would like to update the user guide (https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html) to reflect the change i have made, I'm just a bit unsure how to do that...

nowackipawel:
implode(string, array) I think

MohKari:
@nowackipawel Cheers for that, weirdly it seems to work either way, but i have updated the code to use implode(string, array) as that is what it uses in php.net.

jim-parry:
The user guide is built using [restructured text](http://docutils.sourceforge.net/rst.html). They have a [primer](http://docutils.sourceforge.net/docs/user/rst/quickstart.html) and a [quick reference](http://docutils.sourceforge.net/docs/user/rst/quickref.html), and the existing docs would be informative.

Inside the repo, user_guide_src/README.rst describes how to setup the Sphinx tool on your system.

jim-parry:
getErrorMessage is an internal method, and not mentioned in the user guide.This would apply to custom error messages, and maybe that is where it should be mentioned.

MGatner:
@MohKari did you take a look at User Guide updates?

lonnieezell:
I think this change is good. it needs docs updated and a test in place just to make sure it continues working in the future and doesn't break if we make bone-headed changes. :) 

@MohKari Are you able to make these changes in the next couple of weeks?

MohKari:
Things have gotten a little bit busy over here unfortunately, I'll get the tests and document updates done by end of next week so we can put this pull request to bed.

MohKari:
I've updated the branch. 

I've included a test for all of the "tags" that can get replaced on the custom error messages. I'm not sure if the way I have done it is the best way? Should it be testing Validation's getErrorMessage() method instead? As this is where the replacement happens?

I've removed the bit of code that was converting an array into a string, on the value tag replace, as we don't ever validate an actual array (I have some custom rules for validating arrays that I would like to submit an additional pull request for at some point)

I've updated the documentation, I wasn't sure if it would be best to change the way it looks to a table or list, for now I've kind of just hacked it into the existing sentences.

MGatner:
Thanks for the updates! Looks likes there is still a test failing on input types around arrays.

MohKari:
That should do it!

Removed my array to string conversation, because It looked as though we never submit arrays as a value, but one of the tests do. Yay for tests!

Also just realised my last bunch of commits where not verified. Sorry

MohKari:
First time I've ever used git rebase... the experience was **NOT** fun.

MGatner:
Sorry friend! Thanks for the extra effort though. If it’s a personal branch that only I have worked on I usually just `git reset [whatever]`, reapply my commits, and force push.

MohKari:
Haha! That would of been so much easier... I think I got a bit caught up trying to sign previous commits as valid and ended up having like 10+ commits.  In the end I actually did use git reset, and then merged/squashed in a backup branch. Live and learn!

PR #2089, opened 2019-07-07
WIP: Prevent soft deletes without where condition

rino7:

**Description**
There's a bug, or at least a confuse behaviour, when you use soft-delete and execute $model->delete() method with empty value as parameter (0, NULL, FALSE, or simple no parameter).
When soft-delete is false, CI throw a DatabaseException warning about delete() method cannot execute without where or like condition, but, when soft-delete is true no exception nor warning is thrown because CI internally call update() method.
I modified the code in order to throw DatabaseException when soft-delete is  true and $id parameter is empty and either a where condition was set.

In order to maintain consistance between both behaviour (hard and soft delete) I needed to add a getter to QBWhere property so that I be able to eval in Model if there's a condition set.
Additionally, I think that is usefull to be able to access to the compiled 'where' condition from any Model child class in case you want to modify it o just check

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1588,6 +1588,18 @@ class BaseBuilder
 	}
 
 	//--------------------------------------------------------------------
+	/**
+	 * Get compiled 'where' condition string
+	 *
+	 * Compiles the set conditions and returns the sql statement
+	 *
+	 * @return string
+	 */
+	public function getCompiledQBWhere()
+	{
+		return $this->QBWhere;
+	}
+	//--------------------------------------------------------------------
 
 	/**
 	 * Get_Where
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -223,15 +223,14 @@ class BaseBuilder
 	/**
 	 * Constructor
 	 *
-	 * @param  string|array                              $tableName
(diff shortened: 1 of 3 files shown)


jim-parry:
1) It looks like your changes have broken the Postgres unit testing
2) Commits need to be GPG-signed
3) We have our own code sniffer rules, which are not 100% PSR1/2 compatible. I notice that your opening braces have all been moved to the end of the line in front of a block, for instance. If you forked the repo, you should have a git pre-commit hook, which will automatically run your code through our style guide checker and fix code style.

rino7:
> 1. It looks like your changes have broken the Postgres unit testing
> 2. Commits need to be GPG-signed
> 3. We have our own code sniffer rules, which are not 100% PSR1/2 compatible. I notice that your opening braces have all been moved to the end of the line in front of a block, for instance. If you forked the repo, you should have a git pre-commit hook, which will automatically run your code through our style guide checker and fix code style.

OK, sorry, it's my first contribution to this kind of project.
Are you agree if I close this PR and create a new one with signed commit?

jim-parry:
That would be the easiest way, yes.
Have you checked our contributing guidelines in the repository? They explain things, including the workflow, better than the user guide.

PR #2090, opened 2019-07-08
Fix prevent soft delete all without conditions set

rino7:
**Description**
There's a bug, or at least a confuse behaviour, when you use soft-delete and execute $model->delete() method with empty value as parameter (0, NULL, FALSE, or simple no parameter).
When soft-delete is false, CI throw a DatabaseException warning about delete() method cannot execute without where or like condition, but, when soft-delete is true no exception nor warning is thrown because CI internally call update() method.
I modified the code in order to throw DatabaseException when soft-delete is true and $id parameter is empty and either a where condition was set.

In order to maintain consistance between both behaviour (hard and soft delete) I needed to add a getter to QBWhere property so that I be able to eval in Model if there's a condition set.
Additionally, I think that is usefull to be able to access to the compiled 'where' condition from any Model child class in case you want to modify it o just check

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1588,6 +1588,18 @@ class BaseBuilder
 	}
 
 	//--------------------------------------------------------------------
+	/**
+	 * Get compiled 'where' condition string
+	 *
+	 * Compiles the set conditions and returns the sql statement
+	 *
+	 * @return string
+	 */
+	public function getCompiledQBWhere()
+	{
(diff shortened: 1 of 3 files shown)


MGatner:
Nice PR! I commented on the exception language, but the bigger question of "was this a bug or desired behavior" will have to wait on the admins. I'm in favor of this change.

PR #2091, opened 2019-07-09
Timezone select

MGatner:
**Description**
This adds a handy function to the sparse **date_helper** to generate a select field from PHP's available timezones. Takes optional parameters for HTML class, default value, and timezone filters.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


MGatner:
I posted the updated code implementing your format, but I'm not sure that it produces a more desirable global solution. "Timezone abbreviation" doesn't appear to be defined for all valid timezones so falls back to a time offset (+/- #) which itself is inconsistent between ## and #### depending on whole/partial hour and causing #### to get sorted last (probably could be fixed by a sort flag to `ksort()`).

At the end of the day I think any given user will probably only be selecting a timezone once so it isn't a huge UI hit either way, and given the "messiness" of the problem to begin with we might be picking the lesser evil.

lonnieezell:
Oh, well damn. yeah if doesn't provide a timezone abbreviation for everyplace globally that just creates an even nastier solution mixing offset and abbreviation so sounds like we should scrap that and go back to what you had. Sorry about that!

I guess that explains with installing an OS typically shows a little map of the world with a few cities highlighted. Do you want to make that happen :) Kidding.

MGatner:
I could probably have the map done before the Email Library is ready. :O (Sorry Jim)

I will revert to the original version. Thanks for the input though, yours is an elegant solution for zones that support the abbreviation.

lonnieezell:
Actually - the old email lib is back in the repo now so we actually have email again. Yay! So the ball's in my court now to finish up the Migration stuff - which I'll have to get to next week after I get back from a small vacation.

MGatner:
I saw that. I also saw that the rewrite is planned for 4.1, which makes me hesitant to implement it too heavily in modules lest the functions change. A discussion for a different place, for sure...

lonnieezell:
Understandable

jim-parry:
The current "ported Email" is a dog's  breakfast, IMHO.
It is there to say we have it, but I wouldn't flaunt it :-/

MGatner:
:) Well it is better than nothing, or leaving it open to slews of third-party modules post-release (*cough* auth *cough*). I'm looking over the code and CI4 wrappers now and am feeling more hopeful that modules could use the email Service and need not too many changes after a proper CI4 library version of Email hits.

PR #2092, opened 2019-07-11
Original email port

jim-parry:
This is the original Email port from CodeIgniter 3.
It is inelegant and tested by assumption only (it worked it CI3, right?).

It is, however, available "now", as a "better-than-nothing" provision for CodeIgniter 4's initial release.

I will continue to work on a more elegant and tested Email package, targeting 4.1.


jim-parry:
Bit of a hit on code coverage :(

MGatner:
Closed? Could this mean... Email 4.0?

lonnieezell:
I'm cool with this if it was closed by mistake. Feel free to merge if that's the case.

jim-parry:
Dropped by mistake, sorry.

PR #2098, opened 2019-07-20
Fix entity __isset function

vibbow:
**Description**
Fix entity __isset() function, to handle name map and method rewrite.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


lonnieezell:
I don't mind this functionality, but it will need a test and updated docs, also.

PR #2112, opened 2019-07-29
Update `Entity.php` `__isset` method

vibbow:
Update `Entity` -> `__isset()` method, to check if datamap or getter method exists for given variable name.

This functionaly is already declared in method phpdoc, but not been implemenet before.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -424,6 +424,15 @@ class Entity
 	 */
 	public function __isset(string $key): bool
 	{
+		$key = $this->mapProperty($key);
+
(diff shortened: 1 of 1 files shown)


lonnieezell:
Looks good. Please include tests specifically for these cases and we're good.

vibbow:
Test case included.

vibbow:
There are two tests failed related to this change.

Both of them is caused by `toArray` returns original key and mapped key.

I'm try to figure out a way to fix it.

PR #2116, opened 2019-08-01
Add rename option for Response download()

MGatner:
**Description**
Currently an HTTP Response created with `download()` can take either a path or data with a filename, but when the path option is used the filename is set to the actual local file. This allows an additional parameter for a developer to supply a new name for the file to be downloaded.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/HTTP/Response.php
+++ b/system/HTTP/Response.php
@@ -1038,10 +1038,11 @@ class Response extends Message implements ResponseInterface
 	 * @param string      $filename The path to the file to send
 	 * @param string|null $data     The data to be downloaded
 	 * @param boolean     $setMime  Whether to try and send the actual MIME type
+	 * @param string|null $rename   An optional new name for the file when using a path
 	 *
 	 * @return \CodeIgniter\HTTP\DownloadResponse|null
 	 */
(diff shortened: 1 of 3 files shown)


MGatner:
I forgot to update the User Guide on this one. I think with my other PR merged one could actually just load DownloadResponse directly and return it - but since this function exists and is incomplete for standard use scenarios I’ll leave this PR. Let me know if you want anything changed or done differently!

MGatner:
@lonnieezell could you weigh in on this when you have a chance? I have a module close to release and it relies on this and #2126. This one can be worked around using `DownloadResponse` directly but I do think it would be a good addition to `Response`.

lonnieezell:
Sorry - still getting back into the swing after the move. Thanks for the ping. 

I'd rather have a new method, `setFilename` on the `DownloadResponse` instance. I'm not a fan of  adding yet another argument to that method. 

That would allow it to work like: 

```
return response()->download($file)->setFilename('foo.pdf');
```

MGatner:
Elegant! I will implement that.

MGatner:
Closing in favor of #2129

PR #2129, opened 2019-08-12
Add setFileName() to DownloadResponse

MGatner:
**Description**
Currently an HTTP Response created with download() can take either a path or data with a filename, but when the path option is used the filename is set to the actual local file. This adds a setter to the DownloadResponse so a developer can supply a new name for the file on download.

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide


lonnieezell:
Needs doc update and we're good to go.

MGatner:
Supplied!

PR #2131, opened 2019-08-12
Added No Content response to API\ResponseTrait

tangix:
**Description**
Added a method in API\ResponseTrait to generate No Content (204) response to an API request that doesn't provide any meaningful response to the client but the client still needs status 204 back.

NOTE: I hope I got the commit signed properly this time.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/API/ResponseTrait.php
+++ b/system/API/ResponseTrait.php
@@ -65,6 +65,7 @@ trait ResponseTrait
 	protected $codes = [
 		'created'                   => 201,
 		'deleted'                   => 200,
+		'no_content'                => 204,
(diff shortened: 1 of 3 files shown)


lonnieezell:
Unfortunately signing still wasn't there. :( Keep trying. I'll let this one through, though.

tangix:
Thanks. Looks like I managed to get the signing to verify by adding my no-reply github address to my gpg key. I'll keep an eye on this for future PRs :-)

PR #2135, opened 2019-08-13
Encryption

jim-parry:
Barebones encryption module - OpenSSL only.
This should satisy those that want "something" built in, while not offending those who don't want "anything" built-in.
libsodium support can be added once it gets bundled with our minimum version, and indeed the module is architected to make it easy to add any other encryption handlers.


MGatner:
You're crushing it tonight @jim-parry, keep it up :)

system/Config/Services.php
```diff
@@ -185,6 +187,33 @@ public static function curlrequest(array $options = [], ResponseInterface $respo
...
+		if ($getShared === true)
+		{
+			return static::getSharedInstance('encrypter', $config);
+		}
+
+		if ($config !== null && is_object($config))
+		{
+			$config = (array) $config;
+		}
+
```
lonnieezell: We should probably load the default config here if none is provided. That would match how the rest of them handle it.

system/Config/Services.php
```diff
@@ -378,7 +407,7 @@ public static function logger(bool $getShared = true)
 			return static::getSharedInstance('logger');
 		}
 
-		return new \CodeIgniter\Log\Logger(new Logger());
+		return new \CodeIgniter\Log\Logger(new LoggerConfig());
```
lonnieezell: We should probably use `config('Logger')` here so that we take advantage of any caching and don't create any more instances than we need.

system/Encryption/Encryption.php
```diff
@@ -0,0 +1,298 @@
...
+	 * Class constructor
+	 *
+	 * @param  mixed $params Configuration parameters
+	 * @return void
+	 *
+	 * @throws \CodeIgniter\Encryption\Exceptions\EncryptionException
+	 */
+	public function __construct($params = null)
+	{
+		$this->config = array_merge($this->default, (array) new \Config\Encryption());
```
lonnieezell: The config instance should be passed in. Makes it easier to test and customize, instead of forcing the dependency creation inside the constructor.

system/Encryption/Encryption.php
```diff
@@ -0,0 +1,298 @@
...
+
+		if (is_string($params))
+		{
+			$params = ['driver' => $params];
+		}
+
+		$params = $this->properParams($params);
```
lonnieezell: I'm a bit confused why we're keeping an instance of `$params` here since they are set to `$this->config` in `properParams()`? 

Seems cleaner to do `$this->config = $this->properparams($params);` here and not have properParams set it to me.

system/Encryption/Encryption.php
```diff
@@ -0,0 +1,298 @@
...
+
+		// Check for an unknown driver
+		if (isset($this->drivers[$params['driver']]))
+		{
+			throw EncryptionException::forDriverNotAvailable($params['driver']);
+		}
+
+		// determine what is installed
+		$this->handlers = [
```
lonnieezell: This way of checking doesn't support users creating their own handlers, does it? Is that something we should allow for this? Part of me says yes, because if they're on 7.3+ and the framework is still 7.2, they could write one to take advantage of new features.

system/Encryption/Encryption.php
```diff
@@ -0,0 +1,298 @@
...
+	 * HMAC digest to use
+	 */
+	protected $digest = 'SHA512';
+
+	/**
+	 * Map of drivers to handler classes, in preference order
+	 *
+	 * @var array
+	 */
+	protected $drivers = [
```
lonnieezell: If we are going to allow users to make their own handlers, then this should be setup in the config file, I think, right?

system/Encryption/Encryption.php
```diff
@@ -0,0 +1,298 @@
...
+	 * Initialize or re-initialize an encrypter
+	 *
+	 * @param  array $params Configuration parameters
+	 * @return \CodeIgniter\Encryption\EncrypterInterface
+	 *
+	 * @throws \CodeIgniter\Encryption\Exceptions\EncryptionException
+	 */
+	public function initialize(array $params = [])
+	{
+		$params = $this->properParams($params);
```
lonnieezell: Looks like some duplicate code here (with the constructor) that could be refactored to it's own method, or maybe even handled in the `properParams()` method?

system/Encryption/Encryption.php
```diff
@@ -0,0 +1,298 @@
...
+		}
+
+		// Derive a secret key for the encrypter
+		if (isset($params['key']))
+		{
+			$hmacKey          = strcmp(phpversion(), '7.1.2') >= 0 ? \hash_hkdf($this->digest, $params['key']) : self::hkdf($params['key'], $this->digest);
+			$params['secret'] = bin2hex($hmacKey);
+		}
+
+		$handlerName     = 'CodeIgniter\\Encryption\\Handlers\\' . $this->driver . 'Handler';
```
lonnieezell: If we are moving drivers to a config class, we should define these classes there, also. Otherwise we get stuck in the situation we currently have with the database, where there's no easy way to extend it or write our own handlers for it.

system/Encryption/Encryption.php
```diff
@@ -0,0 +1,298 @@
...
+	 *
+	 * @param string  $key    Input key
+	 * @param string  $digest A SHA-2 hashing algorithm
+	 * @param string  $salt   Optional salt
+	 * @param integer $length Output length (defaults to the selected digest size)
+	 * @param string  $info   Optional context/application-specific info
+	 *
+	 * @return string    A pseudo-random key
+	 */
+	public static function hkdf($key, $digest = 'sha512', $salt = null, $length = 64, $info = '')
```
lonnieezell: Since our minimum is PHP 7.2 now, can we kill this?

system/Encryption/Exceptions/EncryptionException.php
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Encryption exception
+ */
+class EncryptionException extends \RuntimeException implements ExceptionInterface
+{
+
+	public static function forDriverNotAvailable(string $driver = null)
+	{
+		return new static(lang('exception.driverNotAvailable', [$driver]));
```
lonnieezell: Doesn't `encryption` as the language file need to be ucfirst to not die on case-sensitive systems?

system/Encryption/Handlers/BaseHandler.php
```diff
@@ -0,0 +1,134 @@
...
+	//--------------------------------------------------------------------
+
+	/**
+	 * Constructor
+	 *
+	 * @param array $config
+	 */
+	public function __construct($config = [])
+	{
+		$this->logger = \Config\Services::logger(true);
```
lonnieezell: Personally, I prefer this to be injected in the services class so we don't hardcode the dependency here.

system/Encryption/Handlers/OpenSSLHandler.php
```diff
@@ -0,0 +1,179 @@
...
+				$this->key = $params;
+			}
+		}
+		if (empty($this->key))
+		{
+			throw new EncryptionException('Encrypter needs a starter key.');
+		}
+
+		// derive a secret key
+		$secret = strcmp(phpversion(), '7.1.2') >= 0 ?
```
lonnieezell: No longer need to compare the version - since we require 7.2.

system/Encryption/Handlers/OpenSSLHandler.php
```diff
@@ -0,0 +1,179 @@
...
+				$this->key = $params['key'];
+			}
+			else
+			{
+				$this->key = $params;
+			}
+		}
+		if (empty($this->key))
+		{
+			throw new EncryptionException('Decrypter needs a starter key.');
```
lonnieezell: This should be brought out into the Exception class and localized.

system/Encryption/Handlers/OpenSSLHandler.php
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		// derive a secret key
+		$secret = strcmp(phpversion(), '7.1.2') >= 0 ?
```
lonnieezell: Check always passes since 7.2 required.

system/Encryption/Handlers/OpenSSLHandler.php
```diff
@@ -0,0 +1,179 @@
...
+				\hash_hkdf($this->digest, $this->key) :
+				\CodeIgniter\Encryption\Encryption::hkdf($this->key, $this->digest);
+
+		$hmacLength = self::substr($this->digest, 3) / 8;
+		$hmacKey    = self::substr($data, 0, $hmacLength);
+		$data       = self::substr($data, $hmacLength);
+		$hmacCalc   = \hash_hmac($this->digest, $data, $this->secret, true);
+		if (! hash_equals($hmacKey, $hmacCalc))
+		{
+			throw new \CodeIgniter\Encryption\EncryptionException('Message authentication failed.');
```
lonnieezell: Should be pulled out into Encryption class and localized.

system/Encryption/Handlers/BaseHandler.php
```diff
@@ -0,0 +1,134 @@
...
+
+	// --------------------------------------------------------------------
+
+	/**
+	 * Byte-safe strlen()
+	 *
+	 * @param  string $str
+	 * @return integer
+	 */
+	protected static function strlen($str)
```
lonnieezell: Why do these methods exist in the different classes? I don't see them being used anywhere.

system/Config/Services.php
jim-parry: ok

system/Config/Services.php
jim-parry: ok

system/Encryption/Encryption.php
jim-parry: It's not so much letting useres making their own handlers, but providing a list of the built-in handlers (there were going to be several if all went well).

system/Encryption/Encryption.php
jim-parry: reworked

system/Encryption/Encryption.php
jim-parry: reworked

system/Encryption/Encryption.php
jim-parry: it wasn't the intent that users create their own drivers for use here, but that we might provide several alternatives

system/Encryption/Encryption.php
jim-parry: reworked

system/Encryption/Encryption.php
jim-parry: done

system/Encryption/Exceptions/EncryptionException.php
jim-parry: oops - fixed

system/Encryption/Handlers/BaseHandler.php
jim-parry: done

system/Encryption/Handlers/BaseHandler.php
jim-parry: they are used in the extending handlers, eg OpenSSLHandler

system/Encryption/Handlers/OpenSSLHandler.php
jim-parry: agreed

system/Encryption/Handlers/OpenSSLHandler.php
jim-parry: agreed

system/Encryption/Handlers/OpenSSLHandler.php
jim-parry: agreed

system/Encryption/Handlers/OpenSSLHandler.php
jim-parry: agreed

jim-parry:
@lonnieezell No apologies needed - appreciate the feedback.
Changes made, though my local build is having issues with a bunch of non-encryption stuff :-/

jim-parry:
I don't know what has gone wrong with the travis build, but definitely something serious :(
I will recreate this branch/PR and see if that helps.

jim-parry:
Found & fixed the root problem with my travis build - messing with Services:logger.
Addressing unit testing & code coverage before updting this PR.

jim-parry:
@lonnieezell Looking better - simplified, tested & hopefully addressing the concerns you raised :)

lonnieezell:
@jim-parry Looks good to me. Merge at your discretion.

PR #2149, opened 2019-08-19
listNamespaceFiles: Ensure trailing slash 

MGatner:
**Description**
The Autoloader merges Composer namespace paths directly from Composer's autoloader, which doesn't include a trailing slash. This means files coming from `autoloader->getNamespace()` may or may not have a trailing slash, which sometimes causes `FileLocator->listNamespaceFiles()` to miss files it should otherwise find.
This change ensures that the namespace path coming from autoloader will always have the trailing slash.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -422,7 +422,7 @@ class Autoloader
 			unset($paths['CodeIgniter\\']);
 		}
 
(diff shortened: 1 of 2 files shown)


lonnieezell:
Oh, nice catch! Thanks.

PR #2154, opened 2019-08-20
Fix email & migrations docs; update changelog

jim-parry:
Correct user guide errors on migration & email pages.
Update changelog, getting ready for RC.1

### DIFF
--- a/user_guide_src/source/changelogs/index.rst
+++ b/user_guide_src/source/changelogs/index.rst
@@ -17,8 +17,23 @@ Version 4.0.0-beta.4
 
 Release Date: July 25, 2019
 
+Highlights:
 
-:doc:`See all the changes. </changelogs/next>`
+There are some breaking changes...
+
+- The Entity class has been refactored;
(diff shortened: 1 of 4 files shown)


fefo-p:
Hello Jim,
I spotted this wrong assumption based on the comment above the variable

/*
	|--------------------------------------------------------------------------
	| Enable/Disable Migrations
	|--------------------------------------------------------------------------
	|
	| Migrations are **disabled** by default for security reasons.
	| You should enable migrations whenever you intend to do a schema migration
	| and disable it back when you're done.
	|
	*/
	public $enabled = **true**;

jim-parry:
@lonnieezell @MGatner Contradictory user guide text pointed out. Which is correct - the text or the setting?

MGatner:
Text is wrong - migrations are *enabled* by default. That was a change with lonnie's refactor and we probably just missed the comment, good catch by fefo-p

jim-parry:
Fixed this with #2157

PR #2155, opened 2019-08-20
Include .gitignore in starters

MGatner:
**Description**
devstarter and appstarter both have very old versions of .gitignore - it isn't getting updated when new deployments are made. This PR adds .gitignore to the list of files copied out during a release.

**Checklist:**
- [X] Securely signed commits

### DIFF
--- a/admin/release-appstarter
+++ b/admin/release-appstarter
@@ -12,7 +12,7 @@ git checkout $branch
 echo -e "${BOLD}Build the framework distributable${NORMAL}"
 
 echo -e "${BOLD}Copy the main files/folders...${NORMAL}"
-releasable='app public writable README.md contributing.md env license.txt spark tests/_support .gitignore'
+releasable='app public writable README.md contributing.md env license.txt spark .gitignore'
 for fff in $releasable ; do
     if [ -d "$fff" ] ; then
         rm -rf $fff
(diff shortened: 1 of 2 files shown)


admin/release-appstarter
```diff
@@ -12,7 +12,7 @@ git checkout $branch
 echo -e "${BOLD}Build the framework distributable${NORMAL}"
 
 echo -e "${BOLD}Copy the main files/folders...${NORMAL}"
-releasable='app public writable README.md contributing.md env license.txt spark tests/_support'
+releasable='app public writable README.md contributing.md env license.txt spark tests/_support .gitignore'
```
jim-parry: tests_support should actually not be in this list (my mistake). If you could remove it, that would save a change later :)

jim-parry:
Good idea. One teeny change requested :)

admin/release-appstarter
MGatner: Sure thing, all set!

jim-parry:
I have merged this, but I don't think that the gitignore in the framework release prep will make any difference. It will end up in vendor/codeigniter4/framework, which will have no effect on a developer's project.

MGatner:
How does https://github.com/codeigniter4/devstarter get seeded? That's what really needs the updated files.

jim-parry:
It doesn't ... any updates are done manually when there are app changes in a release.

MGatner:
Ah thanks for clarifying. I misunderstood that deployment script - sorry for the trouble!

jim-parry:
The appstarter and framework repos are updated as a "normal" part of a release,
That's where the admin/release-appstarter and release-framework scripts are invoked.
This is all done on my system, and then the releases are manually tagged on github.
Not all can be automated, but as much as possible has been :)

jim-parry:
See admin/README.md and admin/workflow.md

PR #2172, opened 2019-08-28
Base FeatureTestCase on CIUnitTestCase

jim-parry:
FeatureTestCase extended CIDatabaseTestCase, even though there were no data calls.
This prevented testing features if the database tests weren't also being run.

### DIFF
--- a/system/Test/FeatureTestCase.php
+++ b/system/Test/FeatureTestCase.php
@@ -53,7 +53,7 @@ use Config\Services;
  *
  * @package CodeIgniter\Test
  */
-class FeatureTestCase extends CIDatabaseTestCase
+class FeatureTestCase extends CIUnitTestCase
 {
 
 	/**


lonnieezell:
@jim-parry This is not the solution for your problem. Feature tests have to extend from the DatabaseTestCase since the entire reason for Feature Testing is to test a URL and check the response, which most of the times requires some kind of database setup first. 

You can always set `$refresh = false` on the test class to turn off running migrations. If there's another issue you're running into, then we need to find a way to turn that feature off in FeatureTestCase.

jim-parry:
@lonnieezell Hmmm - there is nothing in FeatureTestCase or FeatureTestCaseTest that suggests any need for a database, and I am not sure it reasonable to expect that any URL test will of necessity require a database.
Having FeatureTestCase extend CIDatabaseTestCase breaks URL testing with no database connection, hence my NakedFeatureTestCase, which is identical (as you point out) except that it has no database conenction.
Maybe FeatureTestCase should be a trait instead, so that it could used with or without a database test? That might help those who persist data using a document store or XML, rather than a relational database.

I will revert the change, and figure out a different way to test a generic RESTful resource.

lonnieezell:
A trait would be a fine solution.

PR #2174, opened 2019-08-30
Fix _fromTables()

pjsde:
When there are JOINs and more than one table exists, the FROM clause must be enclosed in parentheses, otherwise an error occurs if a JOIN does not refer to the last table.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1994,6 +1994,11 @@ class BaseBuilder
 	 */
 	protected function _fromTables(): string
 	{
+		if ( ! empty($this->QBJoin) && count($this->QBFrom) > 1)
+		{
+			return '('.implode(', ', $this->QBFrom).')';
+		}
+		
 		return implode(', ', $this->QBFrom);
 	}
 
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1994,11 +1994,6 @@ class BaseBuilder
 	 */
 	protected function _fromTables(): string
 	{
-		if ( ! empty($this->QBJoin) && count($this->QBFrom) > 1)
-		{
-			return '('.implode(', ', $this->QBFrom).')';
-		}
-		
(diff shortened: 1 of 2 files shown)


jim-parry:
@lonnieezell This looks good to me, and nothing seems missing. Should this be merged?

lonnieezell:
@pjsde A couple of quick questions: 

1) Can you confirm this syntax works for all database engines we currently support. It will have to if it's in BaseBuilder.
2) Can you point us to existing tests that cover this situation? If not, we will need additional tests written to prove it works on all current engines.

pjsde:
@lonnieezell You are right, this change should not be in BaseBuilder, but in the driver, as it is not valid on all database engines, being only valid for MySQLi.

pjsde:
Can anyone explain to me why despite the merge is done, the changes are not reflected in the develop branch, because if we consult the history of the file  [https://github.com/codeigniter4/CodeIgniter4/commits/develop/system/Database/MySQLi/Builder.php](https://github.com/codeigniter4/CodeIgniter4/commits/develop/system/Database/MySQLi/Builder.php) we can see that the last commit was made in May 4, 2019 by @jim-parry

jim-parry:
This is a mystery to me.
@lonnieezell Any ideas?

lonnieezell:
I have no idea. There was a change to MySQLi/Connection 17 days ago, but everything else was 5 months ago. 

I guess make a new PR? This part seems out of our control and in GitHub's hands.

jim-parry:
I will check some of the other commits around the same time and see what else might have gone astray.

pjsde:
Ok. I'll make a new PR then.

jim-parry:
The commits before & after this one check out, i.e. the commits look right and the changes are in effect.
This commit looks right in the history, but the changes are not in effect.
The only thing I can attribute this to is a hiccup on github's end, possibly synching or replicating something in their database(s).

PR #2196, opened 2019-09-08
WIP Use site_url for route-based RedirectResponse

### LINKED ISSUE #2119: Bug: Redirect to route ignores path set in baseurl
**Describe the bug**
When using redirect to route responses in a controller, paths in the configured baseURL are lost.

This is exactly like #1126, but the code has changed since the fix and the cause is now different. I.e., this is a regression.

**CodeIgniter 4 version**
4.0.0-beta4

**Affected module(s)**
router

With app.baseURL configured as `http://localhost/path/to/app/` and route configured with `$route->get('myController/test', 'MyController::test')`:

`return redirect()->route('MyController::test')` in a controller
* should redirect to `http://localhost/path/to/app/myController/test`, but
* in fact redirects to `http://localhost/myController/test`.

Initial analysis pertaining to the cause:
* Reverse-resolving the route gives `/myController/test` (with preceding slash)
* `RedirectResponse->route` tries to use `base_url` to wrap the path
* `base_url` uses `URI->resolveRelativeURI` to set the path over the `baseURL`
* this removes all path info from the configured base URL since the route URI begins with `/` and therefore is absolute, completely overriding the path

**Context**
 - OS: Linux
 - Web server lighttpd 1.4.54
 - PHP version 7.3.6

yol:
**Description**
base_url does not actually add the base path if the given URL is absolute (i.e., starts with a slash) and does not add the indexPage. Both is necessary for the redirect response to work correctly in case of routes. site_url handles this correctly.

Fixes #2119 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/HTTP/RedirectResponseTest.php
+++ b/tests/system/HTTP/RedirectResponseTest.php
@@ -3,6 +3,7 @@
 namespace CodeIgniter\HTTP;
 
 use Config\App;
+use CodeIgniter\Config\Config;
 use CodeIgniter\Config\Services;
 use CodeIgniter\Validation\Validation;
 use CodeIgniter\Router\RouteCollection;
@@ -24,8 +25,10 @@ class RedirectResponseTest extends \CIUnitTestCase
 
 		$_SERVER['REQUEST_METHOD'] = 'GET';
 
-		$this->config          = new App();
-		$this->config->baseURL = 'http://example.com';
+		$this->config            = new App();
+		$this->config->baseURL   = 'http://example.com';
+		$this->config->indexPage = '';
+		Config::injectMock('App', $this->config);
 
 		$this->routes = new RouteCollection(Services::locator(), new \Config\Modules());
 		Services::injectMock('routes', $this->routes);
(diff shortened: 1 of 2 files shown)


tests/system/HTTP/RedirectResponseTest.php
```diff
@@ -24,8 +25,10 @@ protected function setUp()
 
 		$_SERVER['REQUEST_METHOD'] = 'GET';
 
-		$this->config          = new App();
-		$this->config->baseURL = 'http://example.com';
+		$this->config            = new App();
+		$this->config->baseURL   = 'http://example.com';
+		$this->config->indexPage = '';
```
jim-parry: Setting indexPage to empty breaks the form & HTML helper tests big-time!

pkerling: Why does changing the local `App` instance here influence other tests?

pkerling: And how to resolve this? I need a different `indexPage` for the test

MGatner:
@lonnieezell I believe this was addressed by your URL function rework. Can you confirm?

tests/system/HTTP/RedirectResponseTest.php
MGatner: Don't change it in `setUp()`, set the specific values you need only on the test where they differ.

MGatner: Ohhh I get it now, you were trying to "undo" the injection of your test. Let me think of another way to approach this...

lonnieezell:
@MGatner This was not fixed and i think the change makes sense here, as `base_url` would not include the settings of `$indexPage`, which it definitely should.

MGatner:
Great. I think this is good then but it breaks the default redirect response test because by default the index page is included (which is not reflected in the assertion). Easy fix if @yol is around.

yol:
Hey, sorry I kind of lost track. So the PR is still good? But the "default" redirect response is broken? Can you clarify what you mean by default?

MGatner:
Basically, in **tests/system/HTTP/RedirectResponseTest.php** around line 61, you should change the expected response from "http://example.com/exampleRoute" to "http://example.com/index.php/exampleRoute"

Ref. https://travis-ci.org/codeigniter4/CodeIgniter4/jobs/623289981#L586

lonnieezell:
Fixed in #2496

PR #2197, opened 2019-09-08
Update Session.php

cstechsandesh:
**sessionExpiration issue**
if we set $sessionExpiration to 0 (zero=expire on browser close) in app config file, but due to this condition if (empty($this->sessionExpiration)) it results true for 0 and expiration time sets to $this->sessionExpiration = (int) ini_get('session.gc_maxlifetime');
so i suggest to change this if condition to if(! isset($this->sessionExpiration))

**Description**

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Session/Session.php
+++ b/system/Session/Session.php
@@ -304,7 +304,8 @@ class Session implements SessionInterface
 				$this->sessionExpiration, $this->cookiePath, $this->cookieDomain, $this->cookieSecure, true // HTTP only; Yes, this is intentional and not configurable for security reasons.
 		);
 
-		if (empty($this->sessionExpiration))
+		//if (empty($this->sessionExpiration))
+		if (!isset($this->sessionExpiration))
 		{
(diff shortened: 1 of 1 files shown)


jim-parry:
@lonnieezell Makes sense to me, but I am not sure if there would be any side effects?

cstechsandesh:
Thank you for your response.
Thats true, but due to this, one cannot set $sessionExpiration on browser
exit in config.
As we set 0 (Zero) it's evaluating true (as empty(0) is true) then it sets
expiration from php ini setting.

Unfeigned Regards,
Sandesh Hule.

cstechsandesh:
Sorry Replied from other email.

lonnieezell:
@jim-parry I can't think of any side effects with this, and makes sense to me, also. Merging.

PR #2200, opened 2019-09-09
Notify Kint of dd alias

MGatner:
**Description**
Because `dd()` is calling `Kint::dump` directly, Kint will always display the translated parameter `reset($vars)` as variable name and `dd()`'s source as call location:
```
Called from .../codeigniter4/codeigniter4/system/Common.php:1082 [dd()]
```

This change has `dd()` add itself to Kint's alias list prior to the call so Kint correctly looks back to the originating call. Since the script will always `exit` there's not need to check `Kint::aliases` for duplicates or anything.

**Note**: This is the Kint-approved way to wrap `d()`

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -1079,6 +1079,7 @@ if (! function_exists('dd'))
 	 */
 	function dd(...$vars)
 	{
+		Kint::$aliases[] = 'dd';
 		Kint::dump(...$vars);
 		exit;
 	}


lonnieezell:
Oh - nice! I kept meaning to look into that but never did. Thanks!

MGatner:
Same, finally got my goat :) Sad how long it took me to figure out for how small a change it was...

PR #2201, opened 2019-09-09
Foreign key columns

MGatner:
**Description**
Currently `getForeignKeys()` only returns information on the related tables. While helpful, this doesn't allow any interactions based on those keys as it lacks the column name. This PR adds the locals and foreign column names to the returned object.

Note that SQLite3 does not have this information available, but does have an additional field `sequence` (also added) which is handy for composite keys. The updated User Guide makes this clear.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


lonnieezell:
nice one. thanks.

PR #2204, opened 2019-09-09
fixes #2049 - debug option 

michalsn:
**Description**
CURLRequest class with the default `debug` value reports an error:
```
curl_setopt_array(): supplied argument is not a valid File-Handle resource
```
This pull request fixes this issue and ensures to create resource if we specify a path to the log file. I believe that tests for this are already started in #2168.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -672,10 +672,10 @@ class CURLRequest extends Request
 		}
 
 		// Debug
-		if (isset($config['debug']))
+		if ($config['debug'])
 		{
-			$curl_options[CURLOPT_VERBOSE] = $config['debug'] === false ? 0 : 1;
-			$curl_options[CURLOPT_STDERR]  = is_string($config['debug']) ? fopen($config['debug'], 'w') : fopen('php://output', 'w+');
+			$curl_options[CURLOPT_VERBOSE] = 1;
+			$curl_options[CURLOPT_STDERR]  = is_string($config['debug']) ? fopen($config['debug'], 'a+') : fopen('php://output', 'w+');
 		}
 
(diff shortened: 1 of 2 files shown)


MGatner:
I think @xbotkaj was right and `$config[‘debug’]` will always be set so we should check `if ($config...` instead of `isset`. 

Also can you merge my tests from https://github.com/codeigniter4/CodeIgniter4/pull/2168? Let me know if you want help with a cross-user-fork merge.

michalsn:
Yes, I removed the `isset` option. I also merged your tests - thanks.

Crap... seems like merge portion wasn't signed 😞 oh well...

MGatner:
This is looking good. If there is any way you can setup signing on the device you did the merge from then GitHub will verify it retroactively. Unless @lonnieezell wants to override the merge block.

michalsn:
Well... I made this (not signed) merge on the same device. I tried to fix it somehow yesterday but from what I can see I failed. So, idk... if you can't merge this just close this pull request, copy my code (it's just 3 lines actually) and add it to your branch with tests @MGatner - I don't mind.

MGatner:
:D thanks! I can probably merge your commits and keep you as the author and just do a verified merge myself. I’ll take a look later tonight - we already got another CURL debug report so I’m eager to get this in.

MGatner:
@michalsn I added your commits to https://github.com/codeigniter4/CodeIgniter4/pull/2168 and assuming it passes we should be all set. Feel free to close this PR if you are satisfied.

michalsn:
Ok, thanks @MGatner.

PR #2205, opened 2019-09-09
Fix empty() bug on DBPrefix

MGatner:
**Description**
When Forge runs `dropTable()` it will check for and strip the DBPrefix from a table name if found. It does this check via `(! empty($this->db->DBPrefix) ...` but since `DBPrefix` is acquired through **BaseConnection**'s magic `__get` without a corresponding `__isset()` `empty()` returns false even when it isn't empty. Because of the magic getter and the initial property we can be sure `DBPrefix` will always be there so it is fine to test `($this->db->DBPrefix` directly. Might consider adding `__isset()` to **BaseConnection** at some point.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -605,7 +605,7 @@ class Forge
 		}
 
 		// If the prefix is already starting the table name, remove it...
-		if (! empty($this->db->DBPrefix) && strpos($tableName, $this->db->DBPrefix) === 0)
+		if ($this->db->DBPrefix && strpos($tableName, $this->db->DBPrefix) === 0)
 		{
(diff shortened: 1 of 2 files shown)


MGatner:
Since 
https://github.com/codeigniter4/CodeIgniter4/issues/2206 was partly caused by this Forge issue I'm adding the fix here as well:
* Moved migration rollback before the drop-table loop

MGatner:
@lonnieezell I think this one is a straight bug fix, unless you would like `__isset()` added to **BaseConnection**.

lonnieezell:
It was brought up in another issue, but adding `_isset()` is probably a good idea.

PR #2209, opened 2019-09-10
Add trace()

MGatner:
**Description**
We advertise `trace()` in the User Guide as a debug tool wrapping `Kint::trace()` but it isn't actually there. This PR adds it to **Common.php**.

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -1084,3 +1084,18 @@ if (! function_exists('dd'))
 		exit;
 	}
 }
+
+
+//--------------------------------------------------------------------
+
+if (! function_exists('trace'))
+{
+	/**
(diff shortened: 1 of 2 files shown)


lonnieezell:
Userguide actually shows, `Kint::trace();` as how to use it, but I'm fine adding this feature in. Just need to update the docs.

MGatner:
Ah I see - the subheaders misled me since the other two are global functions. Well it’s fine either way but if this seems good I’ll update that doc.

lonnieezell:
yeah, I think it's worth doing.

MGatner:
@lonnieezell All set!

Issue #2210, opened 2019-09-10
listTables() failing to use correct prefix


MGatner:
Linux, PHP 7.2, Latest develop branch, MySQLi

If I use `$db->listTables(true)` (for `$constrainByPrefix`) with a prefix that includes an `_`I get a SQL statement that doesn't work. Pseudo-code:
```
$db = db_connect(['DBPrefix' => 'dev_']);
$db->listTables(true);
echo (string)$db->getLastQuery();
```

I see that it is trying to escape the underscore with a bang (!):
`SHOW TABLES FROM `database` LIKE 'dev!_%'`

Tracing it back this is indeed what is set on the **BaseConnection**:
```
/**
 * ESCAPE character
 *
 * @var string
 */
public $likeEscapeChar = '!';
```

What am I missing? That doesn't work as an escape, should be `\`, right?

MGatner:
Postrge seems to provide an additional `ESCAPE '!'` to define the escape character:
```
		return $sql . ' AND "table_name" LIKE \''
				. $this->escapeLikeString($this->DBPrefix) . "%' "
				. sprintf($this->likeEscapeStr, $this->likeEscapeChar);
```
But I don't think that works in MySQLi and it's trying to use `!` regardless.

MGatner:
There are actually a couple problems here... The escape character is one, but I'm baffled why it isn't causing failures in #2211. The other issue is that `listTables()` doesn't check `$constrainPrefix` before returning a cached value. That should be a pretty easy fix though.

michalsn:
From what I can see this is a bug inherited from CI3.

The thing is that `SHOW TABLES FROM` query can't have custom escaping character in `LIKE` statement. So something like this won't work:
```sql
SHOW TABLES FROM `database` LIKE 'dev!_%' ESCAPE '!';
```
We can try to use something like this:
```sql
SELECT `table_name` FROM information_schema.tables WHERE `table_schema` = 'database' AND `table_name` LIKE 'dev!_%' ESCAPE '!';
```
But I'm not sure if users always have access to `information_schema`. Any feedback on this would be appreciated. If so, we can do just:

```php
protected function _listTables(bool $prefixLimit = false): string
{
	$sql = 'SHOW TABLES FROM ' . $this->escapeIdentifiers($this->database);

	if ($prefixLimit !== false && $this->DBPrefix !== '')
	{
[...]

PR #2211, opened 2019-09-11
Add listTable() tests

MGatner:
**Description**
Currently there are no tests for the database metadata methods. There are a lot more that should be added but this creates the file and adds tests for `listTables()` that exhibits the error described in https://github.com/codeigniter4/CodeIgniter4/issues/2210.

Someone who knows what's up with the weird `!` escape character business take a look?

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/tests/system/Database/Live/MetadataTest.php
+++ b/tests/system/Database/Live/MetadataTest.php
@@ -1,0 +1,34 @@
+<?php namespace CodeIgniter\Database\Live;
+
+use CodeIgniter\Test\CIDatabaseTestCase;
+
+/**
+ * @group DatabaseLive
+ */
+class MetadataTest extends CIDatabaseTestCase
+{
+	protected $refresh = true;
+
+	protected $seed = 'Tests\Support\Database\Seeds\CITestSeeder';
+
+	public function testCanListTables()
+	{
+		$result = $this->db->listTables();
+
+		// user, job, misc, migrations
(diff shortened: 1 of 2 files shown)


MGatner:
@lonnieezell Sorry I made a mess of this. Good news is `[ci skip]` worked on my commit above!

I believe this is correct now and the failure demonstrates the bug with `ConstrainPrefix`. I'm still not sure on a solution.

tests/system/Database/Live/MetadataTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+		// Restore the prefix and get the tables with the original prefix
+		$this->db->setPrefix($DBPrefix);
+		$result = $this->db->listTables(true);
+
+		$this->assertEquals($this->expectedTables, array_values($result));
+
+		// Clean up temporary table
+		$this->db->setPrefix('tmp_');
+		$forge->dropTable('widgets');
```
michalsn:
```suggestion
		$this->forge->dropTable('widgets');
```

MGatner: Good catch, thanks for that!

MGatner:
How the 🤬 did code coverage go *down* then all I did was add tests?? @lonnieezell this is ready to go unless @michalsn notices anything else.

PR #2212, opened 2019-09-11
Add MySQL-specific character for like escapes

MGatner:
**Description**
Adds `\` as the "like escape character" for MySQLi driver. 

This would fix https://github.com/codeigniter4/CodeIgniter4/issues/2210 but I'm curious if it would break other things.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


jim-parry:
It looks like this does indeed break travis-ci, for Database/Live/LikeTest

MGatner:
Yes, I think this is not the correct way to handle this.

PR #2216, opened 2019-09-13
Update uri.rst

dangereyes88:
Corrected Typo


MGatner:
Thanks @dangereyes88! For future reference if you’re just updating the User Guide (I.e. nothing that requires testing) you can prefix your commit with `[ci skip]` to prevent pipeline testing triggers.

dangereyes88:
ok

PR #2227, opened 2019-09-17
WIP Global "charset" and "collate" settings not working in MySQLi Forge

denis303:
empty($this->db->DBCollat) and empty($this->db->charset) always return true besause "DBCollat" and "charset" properties is protected.

### DIFF
--- a/system/Database/MySQLi/Forge.php
+++ b/system/Database/MySQLi/Forge.php
@@ -139,12 +139,12 @@ class Forge extends \CodeIgniter\Database\Forge
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


jim-parry:
Contributions need to be GPG-signed, thanks.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/CONTRIBUTING.md#signing

MGatner:
This was a symptom of not having a magic `__isset()` method to match the `__get()` but that has since been added via https://github.com/codeigniter4/CodeIgniter4/commit/957d67467ca25b0577bda8205b129125da2e0c1e, making this irrelevant.

PR #2228, opened 2019-09-17
Exclude `sqlite_%` from listTables()

MGatner:
**Description**
SQLite will ad-hoc add various tables prefixed "sqlite_" that function similar to MySQL's `information_schema`. See for example these metadata test results (https://travis-ci.org/codeigniter4/CodeIgniter4/jobs/586068130):
```
1) CodeIgniter\Database\Live\MetadataTest::testListTables
585Failed asserting that two arrays are equal.
586--- Expected
587+++ Actual
588@@ @@
589 Array (
590     0 => 'db_migrations'
591-    1 => 'db_user'
592-    2 => 'db_job'
593-    3 => 'db_misc'
594-    4 => 'db_empty'
595-    5 => 'db_secondary'
596+    1 => 'sqlite_sequence'
597+    2 => 'db_user'
598+    3 => 'db_job'
599+    4 => 'db_misc'
600+    5 => 'db_empty'
601+    6 => 'db_secondary'
602 )
```
... where `sqlite_sequence` is being returned as a table. I think these tables should be excluded as system data to be consistent with the other database drivers. E.g. if I use forge to add two tables to an empty database, `listTables()` should return two tables regardless of driver.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/SQLite3/Connection.php
+++ b/system/Database/SQLite3/Connection.php
@@ -207,6 +207,7 @@ class Connection extends BaseConnection implements ConnectionInterface
 	protected function _listTables(bool $prefixLimit = false): string
 	{
 		return 'SELECT "NAME" FROM "SQLITE_MASTER" WHERE "TYPE" = \'table\''
+			   . ' AND "NAME" NOT LIKE \'sqlite_%\''
 			   . (($prefixLimit !== false && $this->DBPrefix !== '')
 				? ' AND "NAME" LIKE \'' . $this->escapeLikeString($this->DBPrefix) . '%\' ' . sprintf($this->likeEscapeStr,
 					$this->likeEscapeChar)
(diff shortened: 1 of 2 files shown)


system/Database/SQLite3/Connection.php
```diff
@@ -207,6 +207,7 @@ protected function _escapeString(string $str): string
 	protected function _listTables(bool $prefixLimit = false): string
 	{
 		return 'SELECT "NAME" FROM "SQLITE_MASTER" WHERE "TYPE" = \'table\''
+			   . ' AND "NAME" NOT LIKE \'sqlite_%\''
```
michalsn: The current query will match not only `sqlite_` prefix but every name like: `sqlite123`, `sqlitewhatever` etc.

It should be:
```suggestion
			   . ' AND "NAME" NOT LIKE \'sqlite!_%\' ESCAPE \'!\'' 
```

MGatner: Glad you're here checking up... will fix.

PR #2229, opened 2019-09-17
MySQLi escapeLikeStringDirect()

MGatner:
**Description**
As @michalsn points out, some MySQL commands (e.g. "SHOW ____") don't support custom escape characters supplied via "ESCAPE x". Currently `listTables()` is the only framework function suffering from this limitation, but it causes `listTables()` with `$constrainPrefix` to fail to constrain correctly.

This PR defines a new platform-specific function `escapeLikeStringDirect()` that can be used for those rare instances when both "SHOW" and "LIKE" are retried in the same query. This new function applies the default escape character `\` directly the the "LIKE" string for any wildcards. This also has the MySQLi driver's `_listTables()` function use this new method instead to the generic one designed for "SELECT" statements.

Refs:
* https://github.com/codeigniter4/CodeIgniter4/issues/2210
* https://github.com/codeigniter4/CodeIgniter4/pull/2211
* https://github.com/codeigniter4/CodeIgniter4/pull/2212

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/MySQLi/Connection.php
+++ b/system/Database/MySQLi/Connection.php
@@ -390,7 +390,48 @@ class Connection extends BaseConnection implements ConnectionInterface
 	//--------------------------------------------------------------------
 
 	/**
+	 * Escape Like String Direct
+	 * There are a few instances where MySQLi queries cannot take the
+	 * additional "ESCAPE x" parameter for specifying the escape character
+	 * in "LIKE" strings, and this handles those directly with a backslash.
+	 *
+	 * @param  string|string[] $str  Input string
+	 * @return string|string[]
+	 */
+	public function escapeLikeStringDirect($str)
+	{
+		if (is_array($str))
+		{
+			foreach ($str as $key => $val)
+			{
+				$str[$key] = $this->escapeLikeStringDirect($val);
+			}
+
+			return $str;
+		}
+
+		$str = $this->_escapeString($str);
+
+		// Escape LIKE condition wildcards
+		return str_replace([
+			$this->likeEscapeChar,
+			'%',
+			'_',
+		], [
+			'\\' . $this->likeEscapeChar,
+			'\\' . '%',
+			'\\' . '_',
+		], $str
+		);
+
+		return $str;
+	}
+
+	//--------------------------------------------------------------------
+	
+	/**
 	 * Generates the SQL for listing tables in a platform-dependent manner.
+	 * Uses escapeLikeStringDirect().
 	 *
 	 * @param boolean $prefixLimit
 	 *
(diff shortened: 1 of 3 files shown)


MGatner:
Someone tell me how those are valid Postgres strings:
```
1) CodeIgniter\Database\Live\EscapeTest::testEscape
586Failed asserting that two strings are equal.
587--- Expected
588+++ Actual
589@@ @@
590-'SELECT * FROM brands WHERE name = 'O\'Doules''
591+'SELECT * FROM brands WHERE name = 'O''Doules''
592
593/home/travis/build/codeigniter4/CodeIgniter4/tests/system/Database/Live/EscapeTest.php:31
594
5952) CodeIgniter\Database\Live\EscapeTest::testEscapeString
596Failed asserting that two strings are equal.
597--- Expected
598+++ Actual
599@@ @@
600-'SELECT * FROM brands WHERE name = 'O\'Doules''
601+'SELECT * FROM brands WHERE name = 'O''Doules''
602
603/home/travis/build/codeigniter4/CodeIgniter4/tests/system/Database/Live/EscapeTest.php:41
```

jim-parry:
Looks like two diferent escaping expectations, one with \' and one with '' ro end up with a single quote inside a single-quoted string

Can't attest to their validity, but it looks to me like the search is for "O'Doules"

MGatner:
Surely that can't be right? How could Postrgres escape a character within a string wrapped in `'` with another `'`? Maybe I need to do some Postrgres research...

jim-parry:
I have seen that convention before, with other programming languages. I don't know anything about Postgres.

MGatner:
Yeah me neither! I thought I was adding some easy tests, but I suppose it was foolish to assume the same escape character across all drivers.

michalsn:
Escaping single quotes `'` by doubling them `''` is a standard way for PostgreSQL.

https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS

MGatner:
@michalsn Thanks for elucidating! I'm trying a version using each driver's escape char instead.

MGatner:
Finally got it! Someone please review - basically:
* Added `escapeLikeStringDirect()` to handle MySQLi `SHOW ... LIKE ...` query escaping
* Added **EscapeTest.php** for testing `escapeLikeStringDirect()` and a variety of other DB escape types

lonnieezell:
Looks fine to me, but I'm definitely not an expert on it. I'll take yours - and the tests - word for it.

MGatner:
I'm no expert but I feel a lot closer after 7 commits 😳

Put it on the list for 4.5 when we rewrite the Database layer that escape *types* should be differentiated better and *implementations* standardized better.

lonnieezell:
Oh, so now the database rewrite is an actual thing :)

I definitely think it's showing it's age and the fact that it's been updated and worked on for 10+ years. Part of me would really like to rewrite that at some point, but not looking forward to it either lol

MGatner:
Oh yes, most definitely! I'm creating a milestone and assigning only you, with a November deadline. Good luck!

MGatner:
Jokes aside I just looked at my PR history and 11 out of the last 15 were for the database layer, mostly all bug fixes.

PR #2230, opened 2019-09-17
Add escape to SQLite _listTables()

MGatner:
**Description**
Properly escape the SQLite `_listTables()` filter, per https://github.com/codeigniter4/CodeIgniter4/pull/2228#discussion_r325263010

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


MGatner:
I think the coverage test is bugged, and this is a fix to the previously-merge logic error so I'm going to go for it.

Issue #2219, opened 2019-09-14
Add magic __isset to classes with __get


MGatner:
The following classes have a magic `__get()` method without a corresponding `__isset()`:
* CodeIgniter4/system/Database/BaseConnection.php
* CodeIgniter4/system/Model.php
* CodeIgniter4/system/Session/Session.php
* CodeIgniter4/system/I18n/Time.php
* CodeIgniter4/system/CLI/BaseCommand.php
* CodeIgniter4/system/I18n/TimeDifference.php
* CodeIgniter4/system/Encryption/Encryption.php
* CodeIgniter4/system/Encryption/Handlers/BaseHandler.php

In general this is a bad idea because some functions that call language constructs like `isset()` on an instance will fail to return `true` when `__get()` would still return a valid item. See https://github.com/codeigniter4/CodeIgniter4/pull/2205 for one example.

Unless there's some compelling reason not to, I propose each of these classes should have a basic `__isset()` added, modeled after their `__get()`. E.g.:
```
    public function __isset($name)
    {
        return isset($this->attributes[$name]);
    }
```

lonnieezell:
No compelling reason not to do that.

jim-parry:
Fixed by #2231

PR #2231, opened 2019-09-17
Add magic __isset to classes with __get

MGatner:
**Description**
See https://github.com/codeigniter4/CodeIgniter4/issues/2219

This PR adds `isset()` methods to the following classes, modeled after each class's `__get()` method:
* CodeIgniter4/system/Database/BaseConnection.php
* CodeIgniter4/system/Model.php
* CodeIgniter4/system/Session/Session.php
* CodeIgniter4/system/I18n/Time.php
* CodeIgniter4/system/CLI/BaseCommand.php
* CodeIgniter4/system/I18n/TimeDifference.php
* CodeIgniter4/system/Encryption/Encryption.php
* CodeIgniter4/system/Encryption/Handlers/BaseHandler.php

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1634,6 +1634,31 @@ class Model
 		return null;
 	}
 
+	/**
+	 * Checks for the existence of properties across this model, builder, and db connection.
+	 *
+	 * @param string $name
+	 *
+	 * @return bool
+	 */
+	public function __isset(string $name): bool
+	{
+		if (in_array($name, ['primaryKey', 'table', 'returnType', 'DBGroup']))
+		{
+			return true;
+		}
+		elseif (isset($this->db->$name))
+		{
+			return true;
+		}
+		elseif (isset($this->builder()->$name))
+		{
+			return true;
+		}
+
+		return false;
+	}
+
 	//--------------------------------------------------------------------
 
 	/**
--- a/system/Model.php
+++ b/system/Model.php
@@ -1643,7 +1643,7 @@ class Model
 	 */
 	public function __isset(string $name): bool
 	{
-		if (in_array($name, ['primaryKey', 'table', 'returnType', 'DBGroup']))
+		if (property_exists($this, $name))
 		{
 			return true;
 		}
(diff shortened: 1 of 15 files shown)


MGatner:
@lonnieezell or @jim-parry I don't foresee any issues with this but would still appreciate a glance.

jim-parry:
Weird that the same purpose method has to be done so differently across the classes; suggests that it is a good idea.
And code coverage went down because of new code that isn't explicitly tested, I gather :-/

MGatner:
Yeah, it's a bummer but PHP uses `__isset()` natively when `__get()` is defined since it's valid for `__get()` to throw: https://www.php.net/manual/en/language.oop5.overloading.php#object.isset

I suppose we could add tests for each class that received the new magic function, but might make more sense to find a way to exempt them?

MGatner:
Actually something seems to be buggy (or I'm ignorant of something) with coverage tests right now. See for example:
* https://github.com/codeigniter4/CodeIgniter4/pull/2230 - No method changes, just inserted a string concat
* https://github.com/codeigniter4/CodeIgniter4/pull/2211 - Actually *added* tests and had code coverage reduction

jim-parry:
That would make sense, or adding them to existing tests. The troublesome one, IMHO, is the Model, as the property could be interpreted in different ways. Is it a problem if two or more of the possible properties exist?

MGatner:
The model `__isset()` follows the test pattern of `__get()`, so it's possible there's a logic problem but it would be much larger than this new method :). I think though that `isset()` is intentionally ignorant of its content so it's pretty safe to have extra successes and the follow-up `__get()` would be the place for validating.

system/Model.php
```diff
@@ -1634,6 +1634,31 @@ public function __get(string $name)
...
+		}
+		elseif (isset($this->db->$name))
+		{
+			return true;
+		}
+		elseif (isset($this->builder()->$name))
+		{
+			return true;
+		}
+
```
lonnieezell: One thing this doesn't account for is any custom properties a user might have put in their model. I think this needs a `property_exists()` check. Maybe instead of the first check? Unsure.

MGatner: Yes I think you're right. Added via 4157f2965

PR #2232, opened 2019-09-19
Fixes BaseBuilder getWhere() bug

### LINKED ISSUE #2143: Query Builder getWhere Crash
**Describe the bug**
When using the Query Builder, the getWhere method crashes if no offset is given with "Argument 2 passed to CodeIgniter\\Database\\BaseBuilder::limit() must be of the type int, null given"

The default offset should be zero, and if an int is entered, it would use that.

**CodeIgniter 4 version**
BETA4

**Affected module(s)**
/vendor/codeigniter4/framework/system/Database/BaseBuilder.php line 1612

->getWhere([
	'email' => $email,
], 1);

will crash

->getWhere([
	'email' => $email,
], 1, 0);

will run as expected.

Suggest changing line 1603 of /vendor/codeigniter4/framework/system/Database/BaseBuilder.php

from

public function getWhere($where = null, int $limit = null, int $offset = null)

to

public function getWhere($where = null, int $limit = null, int $offset = 0)

**Context**
 - PHP version 7.3.8

michalsn:
**Description**
Fixes #2143 - `getWhere()` bug with offset.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -1090,7 +1090,7 @@ Class Reference
 		:param	int	$limit: The LIMIT clause
 		:param	int	$offset: The OFFSET clause
 		:param 	bool $returnSQL: If true, returns the generate SQL, otherwise executes the query.
-		:param 	bool $reset: Are we want to clear query builder values?
+		:param 	bool $reset: Do we want to clear query builder values?
 		:returns:	\CodeIgniter\Database\ResultInterface instance (method chaining)
 		:rtype:	\CodeIgniter\Database\ResultInterface
 
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -1090,7 +1090,7 @@ Class Reference
 		:param	int	$limit: The LIMIT clause
 		:param	int	$offset: The OFFSET clause
 		:param 	bool $returnSQL: If true, returns the generate SQL, otherwise executes the query.
-	    :param 	bool $reset: Are we want to clear query builder values?
+		:param 	bool $reset: Are we want to clear query builder values?
 		:returns:	\CodeIgniter\Database\ResultInterface instance (method chaining)
 		:rtype:	\CodeIgniter\Database\ResultInterface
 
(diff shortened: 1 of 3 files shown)


jim-parry:
You have added returnSQL and reset parameters to the getWhere() signature.
These don't seem related to the default values bug.
Am I missing something?

michalsn:
1. `returnSQL` - without it this method is not testable.
2. `reset` - I feel it should be consistent with the `get()` method - it just makes more sense to me... so I don't see a reason not to allow reset in this method too.

jim-parry:
@lonnieezell @MGatner Makes sense. Are you good with this?

MGatner:
It looks good to me, really nice tests. It's unfortunate how all the database & builder methods keep growing in number of parameters but I think @michalsn is right that the current implementation needs those. Maybe down the road some way of globally intercepting DB calls to return SQL could replace some of the individual versions?

jim-parry:
If we are adding two parameters to a method signature, shouldn't we be updating the user guide too?

MGatner:
Yes, updates to [user_guide_src/source/database/query_builder.rst](user_guide_src/source/database/query_builder.rst). If @michalsn isn't around I could pass these tonight as a separate PR.

user_guide_src/source/database/query_builder.rst
```diff
@@ -1084,11 +1084,13 @@ Class Reference
...
 		called Query Builder methods.
 
-	.. php:method:: getWhere([$where = NULL[, $limit = NULL[, $offset = NULL]]])
+	.. php:method:: getWhere([$where = NULL[, $limit = NULL[, $offset = NULL[, $returnSQL = FALSE[, $reset = TRUE]]]]])
 
 		:param	string	$where: The WHERE clause
 		:param	int	$limit: The LIMIT clause
 		:param	int	$offset: The OFFSET clause
+		:param 	bool $returnSQL: If true, returns the generate SQL, otherwise executes the query.
+	    :param 	bool $reset: Are we want to clear query builder values?
```
MGatner: Thanks! Small items:
* Indent looks to be 1 tab short
* Should read "**Do** we want to clear..."

michalsn: Done.

MGatner:
Thanks so much @michalsn, looks great.

Issue #2086, opened 2019-07-04
View: $parser->render();


devyr:
View parser is unable to render files with html extensions (in CI3 this is possible). Using "echo view('index.html');" is possible but not with "echo $parser->render('index.html');"

jim-parry:
Closed by #2264

devyr:
Thanks :)

PR #2233, opened 2019-09-20
fix. CodeIgniter/View/Parser::render allow other extension

Instrye:
Fix bug from #2086 
reference `CodeIgniter/View/View::render `


panigrc:
The extension should be configurable instead of hardcoded `php`.
I would expect a constant i.e. VIEW_FILE_EXTENSION

jim-parry:
This should have unit tests to demonstrate that a .html or whatever view file can be successfully rendered.

PR #2240, opened 2019-09-20
Handle JSON POSTs in CSRF

nowackipawel:
Important change to handle JSON POSTS in CSRF.
I wanted to send JSONS instead on normal POST but CSRF filter did not allow me to to this even if token existed in my JSON.
```
{"usr_id":1322189,"ctn":"3a111012714a58b53260ede89380d412","token":"ctn","parent":17031,"mode":{"main":"test","additional":"exam"},"time":{"start":1568992305,"end":1568993826},"result":{"2103":["T"]}}
```
my token name is: ctn

I think this is important, without this we are limiting a lot.

### DIFF
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -207,8 +207,14 @@ class Security
 			return $this->CSRFSetCookie($request);
 		}
 
-		// Do the tokens exist in both the _POST and _COOKIE arrays?
-		if (! isset($_POST[$this->CSRFTokenName], $_COOKIE[$this->CSRFCookieName]) || $_POST[$this->CSRFTokenName] !== $_COOKIE[$this->CSRFCookieName]
+		// Do the token exist in _POST or php://input (json) data?
+		$CSRFTokenValue = $_POST[$this->CSRFTokenName] ??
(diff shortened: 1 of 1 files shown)


jim-parry:
@lonnieezell Are you ok with this? It looks good to me.

MGatner:
It looks solid to me but I’m a CSRF noob. :(

jim-parry:
I spent too much time in the CSRF unit tests, so maybe just a bit better than noob.
Using JSON paykiad for the token is a thing.

michalsn:
Personally, I would rather see a fallback to the request header than decoding whole json. `json_*` methods are really slow in php.

jim-parry:
They may be slow, but they are an allowed thing for CSRF token passing :-/
CSRF looks messy to begin with, and I am sure that there are some subtle things we might have overlooked (outside of performance), that might come up on hacker1 once we add CodeIgntier4 to it.

lonnieezell:
I was just getting ready to say seeing it in a header was likely a better solution, also. I've seen Laravel use `X-CSRF-TOKEN`. Not sure how others handle that.

jim-parry:
I think there are a bunch of ways, and that we don't handle them all yet!

Issue #2234, opened 2019-09-20
# ini_set() expects parameter 2 to be string, integer given


John-Betong:
**ini_set() expects parameter 2 to be string, integer given**

Latest commit 957d674 19 hours ago

file: CodeIgniter4/app/Config/Boot/development.php

Statement conflicts with recommended ini_set(...) parameters
 
```
// line: #11
- ini_set('display_errors', 1);
+ ini_set('display_errors', '1');
```

Issue #2236, opened 2019-09-20
ini_set()  expects parameter 2 to be string, integer given


John-Betong:
Latest commit 957d674 19 hours ago

file: CodeIgniter4/app/Config/Boot/testing.php

Statement conflicts with recommended ini_set(...) parameters
```
// line: #12
- ini_set('display_errors', 1);
+ ini_set('display_errors', '1');
```
### Edit:
Added truncated text

Issue #2235, opened 2019-09-20
ini_set() expects parameter 2 to be string, integer given


John-Betong:
Latest commit 957d674 19 hours ago

file: CodeIgniter4/app/Config/Boot/production.php

Statement conflicts with recommended ini_set(...) parameters
```
// line: #10
- ini_set('display_errors', 1);
+ ini_set('display_errors', '1');
```
### Edit:
Final line was truncated, added missing text.

MGatner:
Probably as simple as `ini_set(‘display_errors’, ‘1’)`

John-Betong:
@MGatner 
Is the submitted issue in the correct format to be easily merged?

If not please supply a sample.

jim-parry:
Sample provided, just above :)

PR #2241, opened 2019-09-20
Fix ini_set parameter type

jim-parry:
init_set expects second parameter to be string.

This is a sample PR, John-Betong, addressing #2234, #2235 and #2236.
One PR, as these are related.

Note: the branch name is misleading, as I thought all your suggestions were for strict typing.

### DIFF
--- a/app/Config/Boot/development.php
+++ b/app/Config/Boot/development.php
@@ -8,7 +8,7 @@
   | painful debugging.
  */
 error_reporting(-1);
-ini_set('display_errors', 1);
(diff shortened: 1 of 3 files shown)


jim-parry:
Note the GPG-signed commit, and we have to wait for travis-ci to complete regression testing, to make sure we didn't break anything.

jim-parry:
Fixes #2234, #2235 and #2236.
Those could have been submitted directly as a PR instead of an issue.

Issue #2202, opened 2019-09-09
CURLRequest - supplied argument is not a valid File-Handle resource #Windows x64


xbotkaj:
**Describe the bug**
Using CURLReqguest class produce error:
CRITICAL - 2019-09-09 21:04:32 --> curl_setopt_array(): supplied argument is not a valid File-Handle resource

**CodeIgniter 4 version**
CI4 - RC1

**Affected module(s)**
CURLRequest

**Expected behaviour, and steps to reproduce if appropriate**
Any attempt where is not set debug option will produce this error since Class initialize $config['debug'] = false and later check if via isset() and not via standard condition (true/false) directly from the option so every time this will pass, and it will pass empty option to CURL options that will produce error.

**Context**
 - OS: Windows 8.1
 - Web server Apache
 - PHP version 7.3.8

MGatner:
Sort fo a duplicate of https://github.com/codeigniter4/CodeIgniter4/pull/2049, which should probably be reopened. This is partly underway (see https://github.com/codeigniter4/CodeIgniter4/pull/2168) but needs an actual PR.

xbotkaj:
I fixed first part, simple by changing condition but anyway after allowing debug it will throw this error:

CRITICAL - 2019-09-10 07:52:14 --> curl_setopt_array(): cannot represent a stream of type Output as a STDIO FILE*

michalsn:
You can try this fix #2204

xbotkaj:
> You can try this fix #2204

Sorry didn't see your pull request :/ I did it a bit another way.

michalsn:
Well... from what I see, you removed the `CURLOPT_STDERR` part so it won't work.
According to the documentation, we should have the ability to write the output to a file or just "echo" it when the debug option is enabled.

xbotkaj:
> Well... from what I see, you removed the `CURLOPT_STDERR` part so it won't work.
> According to the documentation, we should have the ability to write the output to a file or just "echo" it when the debug option is enabled.

Yeah, I just realize it :) You are right...

michalsn:
@xbotkaj this should be fixed by #2168

Please, give it a try if you find a moment.

xbotkaj:
> @xbotkaj this should be fixed by #2168
> 
> Please, give it a try if you find a moment.

Sorry I was few days off... I will look into it more in the next days.

xbotkaj:
> @xbotkaj this should be fixed by #2168
> 
[...]

PR #2244, opened 2019-09-20
CURLRequest ... Fix issue #2202 + Addition header fix

xbotkaj:
**Description**
- Changed condition to check if config['debug'] is a string (only in that case it should allow VERBOSE and STDERR option
- Added check to sendRequest() if debug is 'true' and in that case, it will print error message in output else it will use exception handler like normal it should.
- Edit test case for true option, now it will check if in options array is not set VERBOSE and STDERR.

P.S. Additional fix (sorry I realized it is a new bug too late after the commit) is in the header. It used to send 'host header' that shouldn't be send in request header so I removed it. We should consider why are we sending all response headers in CURLrequest. Should be checked what belongs to request header only. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -489,6 +489,9 @@ class CURLRequest extends Request
 	{
 		$headers = $this->getHeaders();
 
+		// Unset host header since this should be passed only in response
+		unset($headers['Host']);
+
 		if (empty($headers))
 		{
 			return $curl_options;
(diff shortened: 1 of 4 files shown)


xbotkaj:
I checked the documentation for 'Host:' and I misunderstood how it works. In this configuration, we expect only CURL requests to our own host. Either we should mention it in the user guide or set it to null nad let CURL handle request based on URL. The host is not necessary to put there on default.

PR #2249, opened 2019-09-21
CURLrequest fixes debugging

xbotkaj:
**Description**
Issue fix: #2202 

- Changed condition to check if config['debug'] is a string (only in that case it should allow VERBOSE and STDERR option
- Added check to sendRequest() if debug is 'true' and in that case, it will print error message in output else it will use exception handler like normal it should.
- Edit test case for true option, now it will check if in options array is not set VERBOSE and STDERR.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -672,10 +672,10 @@ class CURLRequest extends Request
 		}
 
 		// Debug
-		if ($config['debug'])
+		if (is_String($config['debug']))
 		{
 			$curl_options[CURLOPT_VERBOSE] = 1;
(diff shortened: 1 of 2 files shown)


michalsn:
The proposed changes do not comply with the user guide:
> When debug is passed and set to true, this will enable additional debugging to echo to STDOUT during the script execution. This is done by passing CURLOPT_VERBOSE and echoing the output

xbotkaj:
> The proposed changes do not comply with the user guide:
> 
> > When debug is passed and set to true, this will enable additional debugging to echo to STDOUT during the script execution. This is done by passing CURLOPT_VERBOSE and echoing the output

I see... hmmm I will think about another solution since passing it directly to output is not possible according to error. I will try another way how to do it.

PR #2254, opened 2019-09-22
Fix entity _construct()

pjsde:
When we create an instance of an entity first we must affect the attributes and only after that we synchronize the original otherwise the original never stay with the initial position.

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1994,6 +1994,11 @@ class BaseBuilder
 	 */
 	protected function _fromTables(): string
 	{
+		if ( ! empty($this->QBJoin) && count($this->QBFrom) > 1)
+		{
+			return '('.implode(', ', $this->QBFrom).')';
+		}
+		
 		return implode(', ', $this->QBFrom);
 	}
 
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1994,11 +1994,6 @@ class BaseBuilder
 	 */
 	protected function _fromTables(): string
 	{
-		if ( ! empty($this->QBJoin) && count($this->QBFrom) > 1)
-		{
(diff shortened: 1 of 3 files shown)


jim-parry:
Looks like not all the Entity code is ok with this... travis-ci failing on one of the Entity tests :(

lonnieezell:
I'll need to look closer, but I don't think the logic actually works here. If you are creating a new Entity instance and filling the stuff in the constructor, then the original values are all null. When an entity is created through the database, I am pretty sure that it uses the fill method or sets things individually, then syncs the original values at that point. 

So pretty sure this won't work out but will need to dig into it later.

pjsde:
I assumed that when I create an entity and passed it the data by the constructor I would be providing the initial data that would instantiate the class and not new data.

From the logic you then present when I create an entity manually and if I want to pass the initial data to it then I must use the ->setAttributes () function without passing the data through the constructor.

It does not seem to me to be the correct one because in my view, when passing data through the constructor, this data should be the data to instantiate the class.

lonnieezell:
This is primarily used for determining what gets saved to the database, especially during updates, so is in place to assist the database layer. During saves it compares original data to the current data. For inserts, all data is new which is why original data should be empty at that time. 

I agree that it could cause us to scratch our heads when not thinking about the database, but the database layer, and updating only what has changed (and not overwriting values with nulls, etc). 

I get your view about instantiating a class, and that data still gets saved to the class, but an Entity in this instance, is specifically around as a part of the database layer. It represents a database row, basically.

pjsde:
Ok. From that perspective it makes sense then how it is.

Should I close the PR?

MGatner:
I agree with lonnie on this one. @pjsde thanks for the PR but I think it introduces undesired behavior and an be closed.

On a brighter note I know this question has come up before so hopefully having a searchable history here will help someone else! 🤩

PR #2255, opened 2019-09-22
Choose better types for the ci_sessions table

pstef:
**Description**
Choose better types for the ci_sessions table: 
* for Postgres, use type BYTEA for binary data. Compared to storing base64-encoded text, this takes less space
* update the user guide to suggest using type INET for storing IP addresses in Postgres. This type offers input error checking and specialized operators and functions
* for both MySQL and Postgres, use type timestamp instead of integer. This type is easier to work with when analyzing session activity.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Images/Handlers/BaseHandler.php
+++ b/system/Images/Handlers/BaseHandler.php
@@ -1,5 +1,4 @@
 <?php
-
 /**
  * CodeIgniter
  *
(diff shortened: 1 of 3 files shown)


pstef:
Sorry about that, I've made a new PR for this.

PR #2260, opened 2019-09-23
Explain the whoops page

### LINKED ISSUE #2198: Document the "whoops" error page
Provide an error explanation, instead of just "whoops", if things go wrongs early.

jim-parry:
Simple solution to fix #2198


MGatner:
I like this. I also frequently recommend to people to check/set their log settings in **app/Config/Logger.php** - would a link to the logging section of the User Guide be helpful?

jim-parry:
It woulf br useful, but separate. From what I have seen, if you get a Whoops, it is before the normal error handling & logging has a chance to kick in. Hacig said that, a reference to the logs would be an excellent separate topic. Will add :)

MGatner:
Lovely! I approve.

PR #2261, opened 2019-09-23
Fix lang() signature

### LINKED ISSUE #2075: Return value of lang() must be of the type string, array returned
"Return value of lang() must be of the type string, array returned" this is the error I got when I was trying to access language with list array as per example https://codeigniter4.github.io/CodeIgniter4/outgoing/localization.html#nested-arrays

_Language/en/Login.php_

    `return [
        'list' => [
            // Heading
            'heading_title'         => 'User',
    
            // Text
            'text_success'          => 'Success: You have modified attributes!',
            'text_list'             => 'User List',
            'text_add'              => 'Add User',
            'text_edit'             => 'Edit User',
    
            // Column
            'column_name'            => 'User Name',
            'column_attribute_group' => 'User Group',
            'column_sort_order'      => 'Sort Order',
            'column_action'          => 'Action',
    
            // Entry
            'entry_name'             => 'User Name',
            'entry_attribute_group'  => 'User Group',
            'entry_sort_order'       => 'Sort Order',
    
            // Error
            'error_permission'       => 'Warning: You do not have permission to modify attributes!',
            'error_attribute_group'  => 'User Group Required!',
            'error_name'             => 'User Name must be between 1 and 64 characters!',
            'error_product'          => 'Warning: This attribute cannot be deleted as it is currently assigned to %s products!',
        ]
    ];`

I'm trying access all in one go to avoid call lang function again and again.

In Controller

      $data['lang_list_array'] = lang('Login.list', [], 'en');

The Result I got is "Return value of lang() must be of the type string, array returned". Is it a bug or error on my side, what mistake did I made?

I also tried this from documentation but it also didn't work

    return [
        'list' => [
            'Apples',
            'Bananas',
            'Grapes',
            'Lemons',
            'Oranges',
            'Strawberries'
        ]
    ];

**CodeIgniter 4 version**
v4.0.0-beta-3

**Context**
 - OS: Windows 10
 - Web server Apache 2.4.37
 - PHP version 7.2.14

jim-parry:
lang() indeed returns a string or array of them.
Fixes #2075

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -433,16 +433,16 @@ if (! function_exists('single_service'))
 if (! function_exists('lang'))
 {
 	/**
-	 * A convenience method to translate a string and format it
-	 * with the intl extension's MessageFormatter object.
+	 * A convenience method to translate a string or array of them and format
+	 * the result with the intl extension's MessageFormatter.
 	 *
-	 * @param string $line
-	 * @param array  $args
-	 * @param string $locale
+	 * @param string|[] $line
+	 * @param array     $args
+	 * @param string    $locale
 	 *
 	 * @return string
 	 */
-	function lang(string $line, array $args = [], string $locale = null): string
+	function lang(string $line, array $args = [], string $locale = null)
 	{
 		return Services::language($locale)
 						->getLine($line, $args);
@@ -1084,7 +1084,6 @@ if (! function_exists('dd'))
 		exit;
 	}
 }
-
 
 //--------------------------------------------------------------------
 


klikklak:
Thank you for fixing the signature. There is still a little «bug» in the the comment:
	 * @param string|[] $line
	 * @param array     $args
	 * @param string    $locale
	 *
	 * @return string

=> this should be like that:
	 * @param string   $line
	 * @param array     $args
	 * @param string    $locale
	 *
	 * @return string|[]

PR #2264, opened 2019-09-24
fix. Parser allow other extension

Instrye:
**Description**
- Fix bug from #2086. reference `CodeIgniter/View/View::render ` 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -128,12 +128,13 @@ class Parser extends View
 			$saveData = $this->config->saveData;
 		}
 
-		$view = str_replace('.php', '', $view);
+		$fileExt  = pathinfo($view, PATHINFO_EXTENSION);
+		$view = empty($fileExt) ? $view . '.php' : $view; // allow Views as .html, .tpl, etc (from CI3)
 
 		// Was it cached?
 		if (isset($options['cache']))
 		{
-			$cacheName = $options['cache_name'] ?: $view;
+			$cacheName = $options['cache_name'] ?? str_replace('.php', '', $view);
 
 			if ($output = cache($cacheName))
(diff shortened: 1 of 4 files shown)


MGatner:
The User Guide will need to be updated as well:

"These scripts have a .php extension, but can not contain any PHP." (https://codeigniter4.github.io/CodeIgniter4/outgoing/view_parser.html#what-it-does)

[user_guide_src/source/outgoing/view_parser.rst]

Instrye:
In fact, I just saw that this problem needs to be decided in 4.0.0-rc.2, I don't know what is the reason for fixing this problem. Compatible with CI3? Or is it allowed for any extension? Is this part of the CI4 program?(translate from Google) 
@jim-parry

jim-parry:
It's a defect that makes sense to fix now before the next RC, that's all.
We fix what we can, and defer what we must.

jim-parry:
The view parser was supposed to allow other extensions, and not doing so is an oversight.
I hope to pull together rc.2 on thursday, and hope this can be part of it.

Instrye:
I tried windows / linux / vagrant and there is no way to generate a user guide. Sorry I can't modify the content in the user guide. In fact, English is not my native language, and I don't think I can accurately modify the user guide.(translate from Google)

jim-parry:
You don't have to actually generate the user guide - that is done as part of release building.
The more important thing is to update the appropriate .rst pages.
It is apparent that we need a better process for managing the user guide workflow, but that is something I plan to address in October.

jim-parry:
Thank you :)

PR #2266, opened 2019-09-26
RC.2 release prep

jim-parry:
User guide tweaks, including changelog for RC.2


user_guide_src/source/libraries/uploaded_files.rst
```diff
@@ -126,6 +126,13 @@ In controller::
 	$file1 = $this->request->getFile('images.0');
 	$file2 = $this->request->getFile('images.1');
 
+You can access individual files uploaded at the same time, but with different names,
+using ``getFileMultiple()``::
+
+	$file1 = $this->request->getFileMultiple('somefile');
+	$file2 = $this->request->getFileMultiple('adifferentfile');
+
```
pjsde: This is wrong, because `getFileMultiple()` return an array with the files, so should be something like this:

`$files = $this->request->getFileMultiple('somefile');`
`$file1 = $files[0];`
`$file2 = $files[1];`

user_guide_src/source/incoming/incomingrequest.rst
```diff
@@ -253,10 +253,14 @@ and uses best practices to minimize any security risks.
 		echo $file->getType();          // image/jpg
 	}
 
-You can also retrieve a single file based on the filename given in the HTML file input::
+You can retrieve a single file uploaded on its own, based on the filename given in the HTML file input::
 
 	$file = $request->getFile('uploadedfile');
 
+You can retrieve a single file uploaded as part of multiple ones, based on the filename given in the HTML file input::
+
```
pjsde: As `getFileMultiple()` return an array with the files, the text should be something like this:
You can retrieve the list of the files uploaded as multiple ones, based on the filename given in the HTML file input::
(please correct my english :)  if its wrong)

user_guide_src/source/libraries/uploaded_files.rst
jim-parry: Agreed. I missed that earlier, and should have had more patience and waited for the user guide updates with your PR :(

user_guide_src/source/incoming/incomingrequest.rst
jim-parry: I've adjusted the user guide text for both cases you point out, and that will be in the next push.

PR #2267, opened 2019-09-26
Tests setUp and tearDown: void

MGatner:
**Description**
`setUp()` and `tearDown()` already always return void, but as of PHPUnit 8 the interface requires this in the declaration. This updates the test cases and all tests to add the explicit reference.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


jim-parry:
gmm - a couple of files changed here :-/

MGatner:
Good old recursive `sed` doing work! We'll see how the tests look, but I think it's an innocuous change.

MGatner:
I'm baffled what happened to the test view files. I didn't touch them - maybe some style enforcement injected the newline? It's causing the tests to fail. I'm trying a manual update via GitHub UI but failing that the test might need to be updated.

jim-parry:
I find that the commit pre-hook can mangle the formatting. Maybe it needs to be made smarter or more context-sensitive.

MGatner:
Yeah it's definitely breaking this - I think I will update the test for now because there is no way to commit those files without the newline.

jim-parry:
I have seen this before ... running a group of tests will have \n added to output, whereas running the entire suite doesn't, or vice-versa (can't remember which way it goes).
Anytime I run into this, I make sure the tests work when the lot of them is run, and avoid running that one module's.
This could be a phpunit thing!?

MGatner:
Not sure but I might redo this and exclude those files from the original commit to prevent them changing.

MGatner:
It looks like that fixed it. @jim-parry Do you feel okay about the tests embracing the newlines?

jim-parry:
I am ok with that - it is a different issue to track down at some point.

PR #2268, opened 2019-09-26
Validation session use only if exists

jim-parry:
Modify Validation::getErrors so it doesn't create a session if one isn't in use already

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -1,5 +1,4 @@
 <?php
-
 /**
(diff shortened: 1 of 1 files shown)


jim-parry:
@MGatner This is in response to the slack discussion.
The change doesn't break anything - can you think of a downside?
@lonnieezell ?

MGatner:
Nope! I was watching it for tests myself, looks good and I agree with the change.

PR #2269, opened 2019-09-26
testMode() method for BaseBuilder

michalsn:
**Description**
This PR introduces a new `testMode` variable that is meant to be used in the testing environment. It replaces all occurrences of `$returnSQL`, `$test` and `$testing` - reducing the number of parameters in a couple of methods.

**Important**
This should be considered as a BC change since some parameters from methods are removed. Those parameters weren't well documented but there is still a chance that somebody was using them.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -219,6 +219,13 @@ class BaseBuilder
 	 */
 	protected $canLimitWhereUpdates = true;
 
+	/**
+	 * Builder testing mode status.
+	 *
+	 * @var boolean
+	 */
+	protected $testMode = false;
+
 	//--------------------------------------------------------------------
 
 	/**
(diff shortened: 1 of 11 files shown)


MGatner:
I'm *super* excited to see this happening!

jim-parry:
@MGatner @lonnieezell Is this something we want in RC.2? I figure it's now or after 4.0.0, given the BC

lonnieezell:
I'm all for it happing now.

jim-parry:
Lookihng good, and simpler :)
Is this ready to go?

tangix:
My concern with this is that it won't be possible to return the actual SQL to do last-minute manipulation (for example inserting stuff that is not in the QB natively) but maybe this is OK?
I use the returned SQL string to insert SQL_CALC_FOUND_ROWS and ON DUPLICATE KEY UPDATE for example.

michalsn:
@tangix we have a bunch of methods from the `getCompiled*` family at our disposal.

https://codeigniter4.github.io/userguide/database/query_builder.html?highlight=getcompiled#CodeIgniterDatabaseBaseBuilder::getCompiledSelect

tangix:
> @tangix we have a bunch of methods from the `getCompiled*` family at our disposal.

Yeah, I guess.

PR #2270, opened 2019-09-26
groupStart() refactorization

michalsn:
**Description**
Since we have now `groupStart()` and `havingGroupStart()` it's a good moment to refactor the code responsible for those methods a bit.

In this PR I want to introduce two new protected methods: `groupStartPrepare()` and `groupEndPrepare()`. These will take over the responsibility for handling all other `group*` methods.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1237,25 +1237,11 @@ class BaseBuilder
 	/**
 	 * Starts a query group.
 	 *
-	 * @param string $not  (Internal use only)
-	 * @param string $type (Internal use only)
-	 *
 	 * @return BaseBuilder
 	 */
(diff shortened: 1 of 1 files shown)


jim-parry:
It even looks like the user guide doesn't need updating ... groupStart() there doesn't show parameters :-/
Are you planning unit tests for this?
I don't see this PR as urgent as 2269, as there doesn't appear to be any BC here. Is that right?

lonnieezell:
I like this. As long as we have existing tests in place for the groups then this should be good and doesn't need any additional tests.

michalsn:
The parameters for `groupStart()` were marked in code as "internal use only" so we should be good - no BC changes.

We have tests for existing group methods and I think it should be enough.

PR #2271, opened 2019-09-26
fix ResourcePresenter::setModel()

pjsde:
When we use the `setModel()` function for the first time everything is fine but as we allow you to change the model you have to reset the property that is not updated otherwise you risk having a different name of the created model as we can see in the tests created.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


jim-parry:
Good catch - I thought we had fixed this a while back, but it looks like it got overlooked.
Is this ready to go?

pjsde:
I think so. Passed the tests I created

PR #2276, opened 2019-09-27
Inccorectly formated JSON response , if body is string

nowackipawel:
JSON was incorrectly formatted in case we providing a string (with no " "  around) to setJSON() method.
Of course, this is not the final solution but I am pointing out the place where the issue occurs. 
```
 $this->response
            ->setJSON('my test string without closing " sings around')
            ->send();
```

### DIFF
--- a/system/HTTP/Response.php
+++ b/system/HTTP/Response.php
@@ -447,9 +447,9 @@ class Response extends Message implements ResponseInterface
 	 *
 	 * @return $this
 	 */
-	public function setJSON($body)
+	public function setJSON($body, bool $unencoded = true)
 	{
-		$this->body = $this->formatBody($body, 'json');
+		$this->body = $this->formatBody($body, 'json' . ($unencoded ? '-unencoded' : ''));
 
 		return $this;
 	}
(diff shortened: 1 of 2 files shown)


lonnieezell:
This causes a test to fail. Please fix.

nowackipawel:
@lonnieezell : looks like it is working again.

lonnieezell:
Originally you said:

> Of course, this is not the final solution but I am pointing out the place where the issue occurs.

Do you think this is a final solution now?

nowackipawel:
@lonnieezell : Have to say yes. I thought it is not nice to do it like this, but it is working as it should and there are no performance issues according to it.

jim-parry:
The new parameter to setJSON has a default value of true, which assumes that any string still needs to be encoded/formatted. This behavior is different to the current implementation.
Wouldn't it be better to have a default of false? A developer would then force the formatting of an unacceptable string (as shown in the PR details) by passing the second parameter to setJSON.

The outgoing/response user guide page needs to be updated to show the new parameter.

nowackipawel:
@jim-parry  : Hope it will work ;-)

PR #2278, opened 2019-09-27
Create id primary keys unsigned

MGatner:
**Description**
Changes `$forge->addField('id')` to create an `unsigned` field.

Note that any existing forge foreign key commands pointing to fields created this way will need to be updated to specify the `unsigned` as well, or the `createTable()` command will fail with the generic message "Cannot add foreign key constraint".

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -356,6 +356,7 @@ class Forge
 					'id' => [
 						'type'           => 'INT',
 						'constraint'     => 9,
+						'unsigned'       => true,
 						'auto_increment' => true,
 					],
 				]);
--- a/user_guide_src/source/dbmgmt/forge.rst
+++ b/user_guide_src/source/dbmgmt/forge.rst
@@ -144,7 +144,7 @@ Creating an id field
 --------------------
 
 There is a special exception for creating id fields. A field with type
-id will automatically be assigned as an INT(9) auto_incrementing
+id will automatically be assigned as an unsigned INT(9) auto_incrementing
 Primary Key.
 
 ::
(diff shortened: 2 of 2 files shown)


MGatner:
I think I killed this with a subsequent commit set to `[ci skip]`. I'll push an arbitrary commit to retry tests.

jim-parry:
This sounds like a BC change, for anyone using the forge already.
Is there a reason for the PR? I found conflicting advice when I investigated, but no compelling reason for having unsigned keys.

MGatner:
I haven't researched it, but I was taught that auto-incrementing primary keys should always be unsigned since a) they will never be negative, and b) it increases the range of possible values significantly. 

At hand the issue is that PKs are the most common target for foreign keys, and PKs created by forge's shorthand will require corresponding FKs to be signed to match. My opinion, this is undesirable, but if there isn't compelling evidence for this as a best practice then maybe not worth the hassle? It is easily worked around by not using the shorthand.

jim-parry:
My investigation suggested that either is ok. int gives 2B distinct positive values, whereas unsigned gives 4B. Some devs negate keys on purpose with int, to denote different status, at the risk of that key being reused with an auto_increment insert.
I don't know if CI has been or would be used with an RDB with > 2B records. If not, not an issue. If so, my findinga suggest that the DB would be sharded outside of an app.
Either way, sounds like questionable benefit, especially if a BC.

MGatner:
Sounds good - at this point it seems like a breaking change would be more hassle than any personal benefit to me might merit. :) Retracting.

Issue #2275, opened 2019-09-27
Debug Toolbar error not found tpl error and fix


demortx:
**Describe the bug**
NOT FOUND (/var/www/www-root/data/www/site.com/system/Debug/Toolbar/Views/_database.tpl) -> real name _database.tpl.php

Fix
_database.tpl.php copy and rename _database.tpl and etc files

**CodeIgniter 4 version**
4.0.0-rc.2

**Affected module(s)**
Which package or class is the bug in, if known.

CRITICAL - 2019-09-27 11:33:34 --> Invalid file: {0}
#0 /var/www/www-root/data/www/site.com/system/View/Parser.php(156): CodeIgniter\Exceptions\FrameworkException::forInvalidFile('') -> NOT FOUND (/var/www/www-root/data/www/site.com/system/Debug/Toolbar/Views/_database.tpl) -> real name _database.tpl.php
#1 /var/www/www-root/data/www/site.com/system/Debug/Toolbar/Views/toolbar.tpl.php(139): CodeIgniter\View\Parser->render('_files.tpl')
#2 /var/www/www-root/data/www/site.com/system/Debug/Toolbar.php(495): include('/var/www/www-ro...')
#3 /var/www/www-root/data/www/site.com/system/Debug/Toolbar.php(453): CodeIgniter\Debug\Toolbar->format(Array, 'html')
#4 /var/www/www-root/data/www/site.com/app/Config/Events.php(41): CodeIgniter\Debug\Toolbar->respond()
#5 [internal function]: CodeIgniter\Events\Events::Config\{closure}()
#6 /var/www/www-root/data/www/site.com/system/Events/Events.php(187): call_user_func(Object(Closure))
#7 /var/www/www-root/data/www/site.com/system/CodeIgniter.php(226): CodeIgniter\Events\Events::trigger('pre_system')
#8 /var/www/www-root/data/www/site.com/public_html/index.php(45): CodeIgniter\CodeIgniter->run()
#9 {main}

**Context**
 - OS: [e.g. Windows 99]
 - Web server [apache2handler]
 - PHP version [7.3.0]

dayrui:
/system/View/Parser.php
error, bug

`$fileExt = pathinfo($view, PATHINFO_EXTENSION);
		$view    = empty($fileExt) ? $view . '.php' : $view; // allow Views as .html, .tpl, etc (from CI3)

		// Was it cached?
		if (isset($options['cache']))
		{
[...]

PR #2280, opened 2019-09-27
Rename collector templates

MGatner:
**Description**
Fix for #2275

**Checklist:**
- [X] Securely signed commits
- [X] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [X] Conforms to style guide

PR #2281, opened 2019-09-27
Fix loading templates in toolbar

michalsn:
Ref #2275

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar/Views/toolbar.tpl.php
+++ b/system/Debug/Toolbar/Views/toolbar.tpl.php
@@ -136,7 +136,7 @@
 				<div id="ci-<?= $c['titleSafe'] ?>" class="tab">
 					<h2><?= $c['title'] ?> <span><?= $c['titleDetails'] ?></span></h2>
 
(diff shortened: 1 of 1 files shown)


MGatner:
I took a different approach with https://github.com/codeigniter4/CodeIgniter4/pull/2280. I think either way is fine but I know there are other template references that are error if out than just these two.

michalsn:
Yes, I saw your PR after I submitted mine - sorry 😅 

I didn't close this PR because I thought it's good to stay consistent with the files name convention across this module. So we should remove all `.tpl.php` and use `.tpl` everywhere or stay with what we have now - but it's just me.

I'm quite sure only these two caused problems... If there are still some errors then I can close this PR but I haven't spotted any so far.

MGatner:
I had issues with _database and _files both, so I think whatever we do we’ll need to give it all the same treatment. 

I’m mobile now but I’ll think of a way to add tests for this and that might guide best practice.

PR #2283, opened 2019-09-28
Rename collectors _*.tpl.php to _*.tpl

MGatner:
**Description**
Fix for #2275

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


jim-parry:
Just to be clear, the template filenames ending with .php breaks the toolbar as released? and simply renaming those files fixes this and merits a hotfix?

MGatner:
Sort of. The toolbar uses View Parser for the collector views. Parser used to require a PHO extension, so even though these were templates they were prefixed “.tpl.php”. When they are called by the toolbar they are simply passed as “.tpl” but the Parser uses to apply “.php” to that, which it doesn’t anymore since other formats are acceptable. So it actually looks for the “.tpl” versions - which don’t exist because they are “.tpl.php”.

MGatner:
The hotfix is debatable - the framework is usable but the debug toolbar won’t work on this release as-is.

jim-parry:
I have to do a hotfix for the framework distro ... didn't create the build properly :'(

jim-parry:
Not related to this PR, but might influence hotfix decision. Expecting Lonnie here shortly, or perhaps we should switch to slack to discuss it?
@lonnieezell @MGatner

jim-parry:
I spoke too soon, re framework hotfix. There is a defect in the build script for it, but not affecting the distro as it turns out.

PR #2284, opened 2019-09-28
Fix getErrors() for validation with redirect

michalsn:
**Description**
When performing redirect `_ci validation_errors` session variable is not passed correctly.
Sadly this one is hard to test since we check `! is_cli()` a couple of lines before.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -648,7 +648,7 @@ class Validation implements ValidationInterface
 		{
 			if (isset($_SESSION) && session('_ci_validation_errors'))
 			{
-				$this->errors = unserialize($errors);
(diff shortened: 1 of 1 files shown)


MGatner:
I’ve actually been wondering if is_cli() should exempt testing. I’ve been doing that myself quite a bit.

michalsn:
There is a risk that this may affect tests - but I don't know how much (if any).

MGatner:
Seeing as how this is pretty much a typo I'm going with the change. Tests are good at the moment and if we need to adjust it is likely something that was wrong anyways.

PR #2289, opened 2019-09-30
Don't restrict model's access to properties in a read-only manner

lonnieezell:


MGatner:
This is nice! I assumed there was a reason for this so I was using Reflection to harvest these: https://github.com/tattersoftware/codeigniter4-schemas/blob/65048b2720b52804b7bd30ecbd34d9772dcc558f/src/Handlers/ModelHandler.php#L67

No longer!

Issue #2293, opened 2019-10-01
Lack of test on Travis for PHP 7.4 and tests don't run on upcoming PHP versions


pawelkg:
**Describe the bug**
For now Travis configuration have not set test for PHP 7.4. Other problem is that tests on upcoming PHP version won't run.

**CodeIgniter 4 version**
RC2.1

**Affected module(s)**
none

jim-parry:
This both an issue and a PR? We don't need both.
Why do we need to test PHP7.4 & nightly? The build takes a while as it is, and it is likely to be some time before we look at taking advantage of them.

pawelkg:
For now tests are take for nightly only. I assume that initially at target to PHP 7.4. Now nightly is future version of PHP 8.
Other thing is that now tests for version > 7.3 do not run at all, because composer can not install all packages because of dependencies.
Because PHP7.4 will be released in about month, I think that it's reasonable to make test for this version and prepare to this.

lonnieezell:
We were just discussing this one, actually. The matrix is too unweildy to deal with if we don't have to. I think dropping nightly for now and keeping 7.4 as an `allowed_failure` is workable. 

According to a comment on [this post over at Travis](https://travis-ci.community/t/add-php-7-4-branch-to-test-against/2179/4), we can use `7.4snapshot` might allow us to get that working.

pawelkg:
I update my PR to work with PHP 7.4 (no snapshot) as it's avaliable right now on Travis.
@lonnieezell I wonder if there is a way to change tests to run all databases specified in some file on one pass. That way we will not need to run one travis instance for each driver. As @jim-parry point out, now test take a long time. When there will be more database drivers and more PHP versions to test, there will be easier to test all drivers in one run. In summary that approach should take less time.

MGatner:
7.4 is working from #2514 - a larger pass at the testing service and matrix framework would probably be involved in a refactor.

PR #2294, opened 2019-10-01
Add to travis PHP 7.4nightly and try to run tests on PHP 7.4 and nightly

pawelkg:
Fix issue #2293

**Description**
Travis configuration.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


jim-parry:
ould this be done as two matrices, one for PHP7.2 & 7.3 and one for the others?
Your approach is likely to get unwieldly as we add more database drivers

pawelkg:
I don't think there is much we can do. While I'm not expert in Travis config file, I prepared some simplifies.

.travis.yml
```diff
@@ -1,13 +1,58 @@
...
+        - DB=sqlite
+    - php: 7.4snapshot
+      env:
+        - EXTENDED=false
+        - DB=mysqli
+    - php: 7.4snapshot
+      env:
+        - EXTENDED=false
+        - DB=postgres
+    - php: 7.4snapshot
```
samsonasik: there is already stable `7.4` in travis

pawelkg: I update my PR

lonnieezell:
I'm not a fan of going to a full matrix currently, honestly. I do think we should add 7.4 in right away (I'll take care of that momentarily.) It's always possible we get another solution in the place in the near future. Closing.

PR #2296, opened 2019-10-01
Add counted() to Inflector Helper

MGatner:
**Description**
This adds another inflector function to make easy counted words (e.g. from database results):
```
$users = $userModel->findAll();
echo counted(count($users), 'users'); // "5 users" or "1 user"
```

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Helpers/inflector_helper.php
+++ b/system/Helpers/inflector_helper.php
@@ -165,6 +165,29 @@ if (! function_exists('plural'))
 
 //--------------------------------------------------------------------
 
+if (! function_exists('counted'))
+{
+	/**
+	 * Counted
+	 *
+	 * Takes a number and a word to return the plural or not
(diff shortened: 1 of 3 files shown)


lonnieezell:
This is probably fine, but part of me says - what about non-Arabic numbers for other locales?

MGatner:
No clue. :) How are they handled by `count()`?

MGatner:
This is messy. I think since `counted()`'s dependent functions (`singular`, `plural`, `is_pluralizable`) are hard-coded to English its the best we can do for now. But maybe an international pass at the Inflector helper at some point...

lonnieezell:
Agreed on that. I mean, technically for the numbers we could use the `intl` library, but doesn't help with pluralization, I don't think, so yeah. Stick with English for now on this.

PR #2302, opened 2019-10-04
Fix MySql _fromTables()

pjsde:
When there are JOINs and more than one table exists, the FROM clause must be enclosed in parentheses, otherwise an error occurs if a JOIN does not refer to the last table.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


jim-parry:
I checked the codebase and this was indeed merged this time :)

PR #2307, opened 2019-10-07
Resolve session save handler issue

### LINKED ISSUE #2056: "Cannot call session save handler in a recursive manner"
Screenshot_2019-06-14_at_09_23_07_png

Happens sometimes. Cannot give you a step-by-step guide to reproduce. 
This was also a problem in CodeIgniter 3. Check https://github.com/bcit-ci/CodeIgniter/issues/5545

I think the issue os that `filesize($this->filePath . $sessionID)` in `FileHandler::read` is returning a cached size. Check https://stackoverflow.com/questions/3747982/php-filesize-reporting-old-size 

Adding `clearstatcache()` before `filesize` should fix the issue.

jim-parry:
Closes #2056

### DIFF
--- a/system/Session/Handlers/FileHandler.php
+++ b/system/Session/Handlers/FileHandler.php
@@ -187,7 +187,7 @@ class FileHandler extends BaseHandler implements \SessionHandlerInterface
 			}
 
 			// Needed by write() to detect session_regenerate_id() calls
-			if(is_null($this->sessionID))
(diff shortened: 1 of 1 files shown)


jim-parry:
Yes - it is the solution for several simialr issues, including one in CI3 with a looooong discussion thread. It's not a save handler issue, but a filesize caching problem, according to my findings.
It doesn't break anything, so will merge.

PR #2321, opened 2019-10-10
[ci skip] Update RESTful User Guide

MGatner:
**Description**
I find myself frequently flipping back and forth to the routes expansion for ResourceController and ResourcePresenter. Presenter is currently misaligned, making it hard to read, and the two code sections are not in the same order.

This PR aligns the code and reorganizes both sections to be in the same general order (CRUD-ish). Additionally I added a comparison table to make it easier to view the routes, methods, and functions all at once.

*Note: The reorg and table are two different commits, in case anyone hates my table*

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/incoming/restful.rst
+++ b/user_guide_src/source/incoming/restful.rst
@@ -227,3 +227,26 @@ implement those methods that you want handled.::
 The routing for this would be::
 
     $routes->presenter('photos');
+
+Presenter/Controller Comparison
+=============================================================
+
+This table presents a comparison of the default routes created by `resource()`
+and `presenter()` with their corresponding Controller functions.
+
+=========== ========= ====================== ======================== ==================== ====================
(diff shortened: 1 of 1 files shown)


jim-parry:
I don't object to the different route ordering, but there should be a sentence below it like "The ordering above is for clarity, whereas the actual order the routes are created in, in RouteCollection, ensures proper route resolution.

The table is a good idea - adds clarity.
The Presenter has one extra method, to provide for prompting to delete. That matches the edit & update "style" there, though it doesn't have an equivalent in the Resource methods.

MGatner:
Good point on the ordering - I didn't think to check which order they are actually created. Would you prefer it in the order or actual routing, instead of adding the note? I'm glad to implement either.

Issue #2045, opened 2019-06-07
Model afterInsert return originals?


MGatner:
Right now model event `afterInsert` returns "the original key/value pairs being updated" - which is a little ambiguous, it actually means the values you passed into `insert()` (including fields that didn't match `allowedFields`). While this info is somewhat useful, there is no way to determine what was changed, or even the original data (to perform a diff). I would like to see some way of passing either changed fields or the original fields.

jim-parry:
Is this a feature request or a bug report?

MGatner:
I think that's a question of intent. As a developer, the current behavior threw me off and I was unable to do what I expected with the event. But maybe that was my misunderstanding? It's hard to guess the scope of what model after insert events might need to do, but it makes more sense to me for the event to trigger with the *inserted data* rather than the *submitted data*.

jim-parry:
Makes sense.
This could be treated as a feature request (BC), or as a bug (interpretation). I am not sure how much depends on it, or how many developers are using it as is.

MGatner:
Hard to say! I know @lonnieezell has given Model some attention recently, if he has an opinion.

Either way, if this seems at least partway plausible I can start a PR. Think it's okay to proceed?

jim-parry:
If this is indeed a BC, then I think it should wait. Let's see what Lonnie says.

lonnieezell:
I think it's ok to proceed on this. When that portion of the model was originally created, there were a number of aspects with entities and the like that didn't exist yet, so things have been added on and looks like that was overlooked. But, yes, I believe it should be the data that was actually inserted.

jim-parry:
@MGatner Are you thinking  of proceeding with this? Should it be removed from the rc.3 milestone?

MGatner:
It's a small change, I've been swamped but I'll get it in shortly.

MGatner:
Resolved with https://github.com/codeigniter4/CodeIgniter4/pull/2332

PR #2332, opened 2019-10-15
Change after methods to use actual data

MGatner:
**Description**
Currently model events `afterInsert` and `afterUpdate` pass on the original data they were called with. This severely limits what event triggers can do, not knowing the data that actually reached the database. This PR triggers events with the inserted/updated data instead. Additionally, `afterInsert` now passes the ID of the newly-inserted row (similar to other events), or 0 on failure.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -687,11 +687,6 @@ class Model
 			}
 		}
 
-		// Save the original data so it can be passed to
-		// any Model Event callbacks and not stripped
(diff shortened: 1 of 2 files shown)


MGatner:
Fixes https://github.com/codeigniter4/CodeIgniter4/issues/2045

PR #2335, opened 2019-10-16
Minor grammar fixes

dafriend:
Correct past participle form of verb


MGatner:
Good finds, but not signed!

Issue #2077, opened 2019-06-27
FilterExceptions result in blank page


MGatner:
**Describe the bug**
FilterExceptions are failing to load the exception page in development environment, but rather result in a blank page.

**CodeIgniter 4 version**
Branch `develop`

**Expected behavior, and steps to reproduce if appropriate**
Desired behavior is for any exception to display the callback trace and warning message to the developer. To reproduce add any invalid filter to **app/Config/Filters.php**, e.g.:
```
	public $filters = [
		'foobar' => ['before' => ['*']]
	];
```

lonnieezell:
@MGatner we can close this, right?

PR #2338, opened 2019-10-16
Revert RedirectException change

MGatner:
**Description**
#1769 changed *CodeIgniter.php* to catch `FilterException` instead of `RedirectException`, which breaks redirect routing via exception and introduced the "filter blank page" issue (https://github.com/codeigniter4/CodeIgniter4/issues/2077). 

This PR changes the exception back to the correct exception type.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -38,7 +38,6 @@
 namespace CodeIgniter;
 
 use Closure;
-use CodeIgniter\Filters\Exceptions\FilterException;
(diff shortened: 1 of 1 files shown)


MGatner:
Now also updates `RedirectException` to provide route (as it was intended) instead of the error message.

lonnieezell:
Looks good to me. Feel free to merge as long as tests don't crash.

MGatner:
Here goes!

Issue #2354, opened 2019-10-21
Bug: lang('app.name') should prefer APPPATH.Language/Validation/en/app.php over installed packages


SunilEver:
**Describe the bug**
lang('app.name') should prefer APPPATH.Language/Validation/en/app.php instead of any package like 'vendor/vendorName/package/Language/en/app.php'
In APPPATH.Language/Validation/en/app.php
```
<?php
return [
    'name' => 'New App',
];
```
In 'vendor/vendorName/package/Language/en/app.php'
```
<?php
return [
   'name' => 'App',
];
```
Now `lang('app.name')` should return "New App" instead of "App"

**CodeIgniter 4 version**
v4.0.0-rc.3

**Affected module(s)**
CodeIgniter\Language\Language::class

**Expected behavior, and steps to reproduce if appropriate**
Language files in APPPATH should be given preference

**Context**
- OS: Ubuntu 16
- Web server Apache/2.4.18
- PHP version 7.3.9-1+ubuntu16.04.1+deb.sury.org+1

MGatner:
Glad this issue already exists. See my notes on https://github.com/codeigniter4/CodeIgniter4/pull/2355 - I do think we need to find a way to do this but it needs to be more efficient (and signed, and pass tests).

Essentially it comes down to the return order of `Services::locator()->search($path)` (from **Language.php:302**) being statically: `System`, `Config`, `App`, `modules`. Language reverses this by `array_replace_recursive()` giving priority to the last matched language string.

IMO the correct search order for language strings should be: `App`, `modules`, `System`.

PR #2355, opened 2019-10-21
replaced language setting in APPPATH."Language"

SunilEver:
Fix issue #2354 

**Description**
Updated `CodeIgniter\Language\Language::requireFile()` to prefer APPPATH."Language" Directory

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Language/Language.php
+++ b/system/Language/Language.php
@@ -323,6 +323,12 @@ class Language
 			$strings = $strings[0];
 		}
 
+		# Replace with files in  APPPATH.'Language' directory
+		if( $file = Services::locator()->locateFile($path) )
+		{
+			$strings = array_replace($strings, require $file);
+		}
+
 		return $strings;
 	}
 


MGatner:
I like the idea but this is causing `testPrioritizedLocator()` to fail, which makes me believe that the priority load order still isn't working as expected.

jim-parry:
Some problems with this PR:
- commits not GPG-signed
- unit testing not updated
- user guide not updated

jim-parry:
@MGatner @lonnieezell Neither the user guide or code suggest anything about prioritizing one namespace over another, which appears to be the goal of the bug report / PR. Should App be prioritized?

MGatner:
I think file locating needs to have an explicit hierarchy, which might vary depending on the component. `FileLocator` already uses the generic ordering: `System`, `Config`, `App`, `Modules`. The issue here is that `Language->requireFile()` essentially reverses the order, which could be misleading to anyone familiar with the regular hierarchy:
```
		$files   = Services::locator()->search($path);
...
		$strings = array_replace_recursive(...$strings);
```

As a regular module writer, I do think it is important for developers to be able to supply their own phrasing without modifying the module - so in this case I'm in favor of prioritizing `App` but maybe by pushing this to `FileLocator` instead of the override method here.

MGatner:
I do think what this is addressing constitutes an issue, but I don't think this is the solution. `lang()` is a very high priority function as international-supported projects may call it hundreds of times in a single load, and this solution adds an additional filesystem search to every call:

`Services::locator()->locateFile($path)`

Any language files in `APPPATH` will already be included in `Services::locator()->search($path)`, it might just need to handle prioritizing a bit differently.

Additionally, this is causing tests to fail and the commit is not signed. Closing the PR, but I will open a corresponding issue so it doesn't get lost.

erikkloeze:
> but I will open a corresponding issue so it doesn't get lost

Just had a look at the open issues, and no corresponding issue found. Just a reminder to prevent this issue being unresolved.

MGatner:
The issue is linked directly above yours and has already been addressed. If you’re having a similar issue you could open a support thread on the forums.

erikkloeze:
Noted issue [#2354](https://github.com/codeigniter4/CodeIgniter4/issues/2354) and the accompanying fix in [#2507](https://github.com/codeigniter4/CodeIgniter4/pull/2507). Doesn't the `array_replace_recursive` in **Language:319** (still) overwrite language adjustments made in `App` with the defaults in `vendor` - hence thereby not being able to adjust these?

PR #2358, opened 2019-10-22
Fix Logger config

### LINKED ISSUE #2286: Bug: Logger Path Duplicated
**Direction**
Logger config seems duplicated.

**Describe the bug**
Setting line 45 in  CodeIgniter4/app/Config/Logger.php 

public $path = '';

seems to have no effect.  Double checked with file vendor/codeigniter4/framework/system/Log/Handlers/FileHander line 75 by checking what $config comes in as.  This also uses the value from line 108 and not from 45.

Changing line 108

'path'            => WRITEPATH . 'logs/',

does.  It would seem that line 45 is a holdover from CI3 and should either be removed or do something.

Furthermore, line 118 defaults to php, but in CI4, this folder is no longer stored in a publicly accessible folder, and should default to log, again, seems like a holdover from CI3.

**CodeIgniter 4 version**
RC2

**Affected module(s)**
Logger

jim-parry:
Correct the Config/Logger
- remove second path setting
- change default extension to 'log'
- improve PHPdocs
Closes #2286

### DIFF
--- a/app/Config/Logger.php
+++ b/app/Config/Logger.php
@@ -38,9 +38,8 @@ class Logger extends BaseConfig
 	|--------------------------------------------------------------------------
 	| Error Logging Directory Path
 	|--------------------------------------------------------------------------
-	|
-	|
-	|
+	| By default, logs are written to WRITEPATH . 'logs/'
+	| Specify a different destination here, if desired.
 	*/
 	public $path = '';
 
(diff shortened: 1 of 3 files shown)


MGatner:
I like it. How about https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Log/Handlers/FileHandler.php#L82 also?

jim-parry:
I left tht because the docs say the default extension is "php".
All the "php" extension does is inject a DEFINED||exit scriptlet at the beginning of the log file.
This at least protects the developer who foolishy sticks their log folder somewhere accessible.

Alternate: make "log" the default logfile extension, and change the description in Config/Logger to explain the effect of "php"

lonnieezell:
I'm good with making the default extension .log, honestly.

MGatner:
I think it should be consistent with the Config file, for testing and for odd times when it comes up in testing, etc.

jim-parry:
Should be consistent now

jim-parry:
ohoh - I broke something. WIll return to this later this morning

PR #2364, opened 2019-10-26
Email config doesn't incorporate .env items

dafriend:
A line in .env such as 

    email.SMTPPort = 587

Would not be incorporated into the Email configuration
Caused by Config\Email not extending  CodeIgniter\Config\BaseConfig

### DIFF
--- a/app/Config/Email.php
+++ b/app/Config/Email.php
@@ -1,7 +1,9 @@
 <?php
 namespace Config;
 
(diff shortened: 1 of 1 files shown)


MGatner:
This makes sense to me and is problem just due to porting the current library straight from CI3 but I'd like @jim-parry to take a look too since he handled that.

PR #2368, opened 2019-10-28
Remove log_message from Email::__construct

dafriend:
The practice of logging a "Some Class Initialized" message is a remnant of earlier versions of CI. 

It's not being done in V4.


MGatner:
Agreed. Also, so glad to see this gone! With the nuanced logging levels in CI4 it's not really an issue but it was a bugger in CI3.

PR #2369, opened 2019-10-28
Remove LoggerAwareTrait from Email class

dafriend:
I would like this considered because the `Email` class creates a `Logger` instance when it's instantiated. I'm trying to devise an `EmailHandler` for the `Logger` class which would require an instance of the `Email` class. I sure you can see the endless loop that is formed.

It doesn't seem like removing the `LoggerAwareTrait` would affect `Email` much. It only makes one call to the resulting `logger` object.  Switching that call to using the common function `log_message` seems a good substitute for `$this->logger->log...`

If this PR is merged I've got a working `EmailHandler` ready to go.

### DIFF
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -208,7 +208,6 @@ class Services extends BaseService
 			$config = new \Config\Email();
 		}
 		$email = new \CodeIgniter\Email\Email($config);
-		$email->setLogger(static::logger(true));
 		return $email;
 	}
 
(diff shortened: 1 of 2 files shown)


MGatner:
I think using `log_message()` directly is actually desirable and consistent with other libraries, as we establish logging early as a service and the helper function will always rely on that. I would like @lonnieezell or @jim-parry to take a look as well but I'm good with this change. This library's days are numbered but it would be good to have it feel at least consistent for release.

PR #2381, opened 2019-11-02
Fix: Use of CodeIgniter\Config\Services prevents Service overriding

### LINKED ISSUE #2376: Bug: Usage of `static::methodName` in  CodeIgniter\Config\Services prevents Service overriding
**The Bug**

There are multiple methods in CodeIgniter\Config\Services that make calls to other methods in the class using the form `static::someMethod()`. 

For instance, see CodeIgniter\Config\Services::renderer() where the last line in the method is

    return new \CodeIgniter\View\View(
                $config, 
                $viewPath, 
                static::locator(true), 
                CI_DEBUG, 
                static::logger(true) //  local scope!!!
    );

The problem is that any overriding service defined in APPPATH/Config/Services.php is not in scope so the core class is delivered and not the extended one.

The following services are all called using the `static::someMethod()` form and so cannot be extended.

- logger
- request
- response
- locator
- routes
- cache
- renderer

And I might have missed some. There are quite a few CodeIgniter\Config\Services that make static calls to those services so the ramifications would conceivably multiply.

**CodeIgniter version**
CI 4.0.0-rc.3 at commit 2c04e1cb4bc5a9db49e95cb4cb72e8780ef3ef6b (10/28/19)

**Affected module**
Services

**Reproducing the Problem**

Extend Logger
```
<?php
namespace App\Log;

use Psr\Log\LoggerInterface;
use CodeIgniter\Log\Logger;

class Logger extends Logger implements LoggerInterface
{
    public function myNewMethod($message)
    {
        echo "New Logger Method says: $message";
    }
}

```
In APPPATH/Config/Services.php 
```
public static function logger(bool $getShared = true)
{
    if ($getShared)
    {
        return static::getSharedInstance('logger');
    }

    return new \App\Log\Logger(new \Config\Logger\Logger());
}
```
In a controller (of your choice)

    $this->logger->myNewMethod('There be Dragons!');

Produces the following error

`Call to undefined method CodeIgniter\Log\Logger::myNewMethod() `

**Context**
 - OS: Linux (Debian 9)
 - Web server Apache version 2.4.25
 - PHP version 7.2.17

dafriend:
Fixes #2376

The following statement is used by multiple "core" classes

     use CodeIgniter\Config\Services;

This results in `App\Config\Services` being bypassed which prevents `Service` overrides from being recognized.

All namespace use statements for `Services` in the core classes changed to ` use Config\Services;` so the overrides can be accomplished.

Note: There are a lot of test files that explicitly use `CodeIgniter\Config\Services` that were not changed in the PR. They probably don't need to be and/or should not be changed, but I wanted to point that out.

### DIFF
--- a/system/CLI/CommandRunner.php
+++ b/system/CLI/CommandRunner.php
@@ -40,7 +40,7 @@
 
 namespace CodeIgniter\CLI;
 
-use CodeIgniter\Config\Services;
+use Config\Services;
 use CodeIgniter\Controller;
 
 /**
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -39,7 +39,7 @@
 
 namespace CodeIgniter;
 
-use CodeIgniter\Config\Services;
+use Config\Services;
 use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
 use CodeIgniter\Validation\Validation;
--- a/system/Debug/Toolbar/Collectors/Events.php
+++ b/system/Debug/Toolbar/Collectors/Events.php
@@ -39,7 +39,7 @@
 
 namespace CodeIgniter\Debug\Toolbar\Collectors;
 
-use CodeIgniter\Config\Services;
+use Config\Services;
 use CodeIgniter\View\RendererInterface;
 
 /**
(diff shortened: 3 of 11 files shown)


MGatner:
I think this is good. I also think we need to define a) what can and can't be extended/overridden, and b) what order of precedence various sections give to App/System/Modules/etc. See also https://github.com/codeigniter4/CodeIgniter4/pull/2355#issuecomment-544598709

lonnieezell:
Yes, those were from around from before some of the extension concepts were solidified. So, yes, perfect. Everything should always extend Config\Services. Thanks.

PR #2382, opened 2019-11-02
quick fix postgresql insert id

iam-adty:
`$insertId = $model->insert()`
always give
`$insertId = 0`
on my system when
`pg_version($connID)['server'] = '10.10 (Ubuntu 10.10-0ubuntu0.18.04.1)'`

to get the correct server version, I explode the $v['server'] by space then get the first index of array.

### DIFF
--- a/system/Database/Postgre/Connection.php
+++ b/system/Database/Postgre/Connection.php
@@ -473,7 +473,7 @@ class Connection extends BaseConnection implements ConnectionInterface
 	{
 		$v = pg_version($this->connID);
 		// 'server' key is only available since PostgreSQL 7.4
-		$v = $v['server'] ?? 0;
+		$v = explode(' ', $v['server'])[0] ?? 0;
 
 		$table  = func_num_args() > 0 ? func_get_arg(0) : null;
 		$column = func_num_args() > 1 ? func_get_arg(1) : null;


MGatner:
@iam-adty We'll need signed commits in order to merge.

I have no idea about Postgres is this is typical. Cane someone familiar take a look? @pstef @Tada5hi @hlorehnz @michalsn

pstef:
@MGatner 
Let me answer questions nobody asked:
1. Postgres developers no longer support PG versions before 9.4; 9.4 will lose support soon. Versions 7.4, 8.0, and 8.1 that are mentioned in the method lost support in 2010.
2. internally, we never use lastval() as is can be very misleading whenever a statement invokes nextval() calls on more than one sequence. It becomes even less intuitive when PHP interacts with Postgres via pgBouncer, which happens to be the most popular setup here. We use `UPDATE/INSERT ... RETURNING ...` instead, which we find to be reliably predictable in all cases.

iam-adty:
@MGatner do i need to close this pull request, then create new one with a signed commit?

MGatner:
@iam-adty We will need it signed to merge. Before you go to the effort though could you please converse with @pstef on the best approach for this? I'm woefully ignorant of Postgres but want to make sure what we're putting into the framework represents best practices.

pstef:
For a quick "fix" this seems OK, I suppose.
It would be better to have access to PQserverVersion() from PHP, but we have to rely on parsing server_version.
It would be best to not depend on the PG version at all, but that, I imagine, would be too invasive for CodeIgniter.

MGatner:
@lonnieezell or @jim-parry - seems like this is okay as a PostGres quick fix. Could you give it a look and request a signed PR as necessary?

lonnieezell:
I honestly don't have a clue on best practices here. If there's a more reliable way that's consistent with all of the versions CI users might expect I'm all for that. If not, then I'm fine with this for now. 

And I can always override the merging block on this if we're sure this is acceptable.

MGatner:
I trust @pstef's judgment on PostGres and don't really have any other experience to check against. @lonnieezell I think we got ahead with it.

Issue #2386, opened 2019-11-04
Bug: Mixed migration formats don't order


MGatner:
**Describe the bug**
Migrations support a variety of naming conventions: "The year, month, and date can be separated from each other by dashes, underscores, or not at all." However, in **MigrationRunner** the migrations are keyed by their UID (the version portion) and then `ksort()` orders them. This breaks ordering for mixed version formats:
```
2019-10-22-142810
2019-10-24-132943
20190319121802
```

We already have `getObjectUid()` which generates a UID for each migration so it should be a simple filter_var, but this will likely cause migrations already run to be detected as "not run" and amount to a breaking change, so should be done with caution.

**CodeIgniter 4 version**
Latest develop

**Affected module(s)**
Migrations

MGatner:
Actually the history is run through `getObjectUid()` as well so probably it would be okay to change it there:
```
		// Remove any migrations already in the history
		foreach ($this->getHistory($this->group) as $history)
		{
			unset($migrations[$this->getObjectUid($history)]);
		}
```

PR #2387, opened 2019-11-04
Use only digits for migrations order

MGatner:
**Description**
See #2386 - this PR changes MigrationRunner to use only digits when ordering migrations.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/MigrationRunner.php
+++ b/system/Database/MigrationRunner.php
@@ -682,7 +682,7 @@ class MigrationRunner
 	 */
 	public function getObjectUid($object): string
 	{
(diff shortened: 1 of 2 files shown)


lonnieezell:
I guess this is nicer than just telling people to only use one migration type and that they should be the ones to convince the addon maintainers to change their format :)

MGatner:
This add-on developer may have been caught in the crossfire :ahem: 😳 I definitely plan to update all migrations to the snazzy new format but it's been a lower priority while they have been working.

PR #2389, opened 2019-11-06
Confer silent status to nested seeders

MGatner:
**Description**
Seeds can call other seeds (https://codeigniter4.github.io/userguide/dbmgmt/seeds.html#nesting-seeders) but since seeds are extensions of the seeder themselves the new seeds will start with their own `silent` value. This means any tests using nesting seeders will start displaying seed results for every test call, very undesirable. 

This PR has the parent seeder pass its `silent` value to the child before running it.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/Seeder.php
+++ b/system/Database/Seeder.php
@@ -168,6 +168,7 @@ class Seeder
 			$seeder = new $class($this->config);
 		}
 
+		$seeder->setSilent($this->silent);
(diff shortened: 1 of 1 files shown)


lonnieezell:
Good catch. I approve.

PR #2392, opened 2019-11-08
Add validation function `is_not_unique`

thingNumber1:
Solve this [feature request](https://forum.codeigniter.com/thread-74706.html) at the Codeigniter 4 Forum.

Add a validation to check the database to see if the given value exist. Can ignore records by field/value to filter (currently accept only one filter).


MGatner:
This is looking good! I like your alternate name suggestion on the forums - I'd be in favor of "is_not_unique" since `is_unique` is the other database-specific validation. We will also need this added to [user_guide_src/source/libraries/validation.rst](user_guide_src/source/libraries/validation.rst)

thingNumber1:
Updated the User Guide and changed the name to "is_not_unique"

PR #2399, opened 2019-11-13
Revise Controllers Documentation  [ci skip]

dafriend:
Primarily adds details to the "Hello World" demo. In particular, it addresses file naming and uses a ``Helloworld`` class instead of ``Blog``.

Revise "Defining a Default Controller" to include commenting out ``$routes->get('/', 'Home::index');`` so the demo will work.

Some general word-crafting too.

### DIFF
--- a/user_guide_src/source/incoming/controllers.rst
+++ b/user_guide_src/source/incoming/controllers.rst
@@ -234,7 +234,7 @@ Private methods
 In some cases, you may want certain methods hidden from public access.
 To achieve this, simply declare the method as private or protected. 
 That will prevent it from being served by a URL request. For example,
-if you were to define a method like this::
+if you were to define a method like this for the `Helloworld` controller::
 
 	protected function utility()
 	{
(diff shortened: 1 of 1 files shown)


user_guide_src/source/incoming/controllers.rst
```diff
@@ -213,16 +232,16 @@ Private methods
...
 
 	protected function utility()
 	{
 		// some code
 	}
 
-Trying to access it via the URL, like this, will not work::
+then trying to access it using the following URL will not work::
 
 	example.com/index.php/blog/utility/
```
MGatner: Maybe make this match the new example controller?
`example.com/index.php/helloworld/utility/`

dafriend: Good idea. Done.

MGatner: Thanks! Looks great, merging.

PR #2400, opened 2019-11-14
Remove redundant check in conditionals

dafriend:
Because `empty()` is essentially the concise equivalent to `!isset($var) || $var == false` using both `isset() && !empty()` in a conditional serves no purpose.

The `isset()` check has been removed.

(OK, I know this is trivial. It's just my OCD kicking in.)

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -313,7 +313,7 @@ class Entity
 			$result = $this->mutateDate($result);
 		}
 		// Or cast it as something?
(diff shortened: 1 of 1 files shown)


MGatner:
`empty()` checks a few other things (like `[ ]`) but you're totally right that it makes `isset()` irrelevant - actually I'm pretty sure that is its intent. My OCD appreciates this as well :) If tests pass it's golden.

PR #2403, opened 2019-11-14
Swap order of operations in conditional

dafriend:
Want to be sure `$reportOnly` is there before checking if it is a string.

### DIFF
--- a/system/HTTP/ContentSecurityPolicy.php
+++ b/system/HTTP/ContentSecurityPolicy.php
@@ -811,7 +811,7 @@ class ContentSecurityPolicy
 
 		foreach ($values as $value => $reportOnly)
 		{
-			if (is_numeric($value) && is_string($reportOnly) && ! empty($reportOnly))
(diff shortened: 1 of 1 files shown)


dafriend:
Not a performance seeking change. Trying to avoid `is_string($reportOnly)` from throwing an exception in the case where `$reportOnly` it is not set. Checking for `! empty($reportOnly)` first eliminates that possibility.

MGatner:
I understand the intent, I'm saying that since `$reportOnly` is the recipient of `foreach`'s stuffing it will *always* be set so there is no worry of an exception.
```
foreach ($values as $value => $reportOnly)
```
I don't think the CI failure was related, but if this isn't accomplishing anything else it might not be worth the trouble. Thoughts?

dafriend:
While examining this in the overall context of the call to ``addToHeader()` and that method itself, you are correct. This PR is not worth the trouble.

PR #2407, opened 2019-11-15
Fix url helper functions to work when site hosted in subfolders. 

### LINKED ISSUE #2347: Bug: URL, Redirect and Pagination misbehave
I've similar problems as #2119 + some bonuses :)

I've deployed project to `test.com/e` and set `baseURL` from .env file. Issues are:

- Redirect:
1. `redirect()->to('uri')` function redirects to `test.com/uri` not `test.com/e/uri`
2. `redirect()->to('uri')` at localhost(spark serve) duplicates url when tries to redirect. But at server works as expected but with issue N1. At localhost I'm using with trailing slash to bypass the issue `->to('/uri')` 

- base_url misbehave: 
1. `base_url('uri')` -> `test.com/e/uri`.
2. `base_url('/uri')` -> `test.com/uri`. Should be `test.com/e/uri` I guess
3. `base_url() . 'uri'` -> `test.com/e/uri` (works but not a pretty solution)

- Pagination: `$pager->links()` creates `test.com/?page=2` instead of `test.com/e?page=2`

### LINKED ISSUE #2365: Bug: current_url function not working as expected.
**Describe the bug**
Using current_url function produces an unexpected result. 
Expecting accurate url, server root with uri string provided instead.
Screenshot attached.
codeigniter-currenturl-bug

**CodeIgniter 4 version**
Commit # a424b3e38d0da46ff86c56b2298d8d3f94fb1a26 of the `develop` branch.

**Affected module(s)**
URL Helper.

Expected behavior:
 - Function should return URL as set in the App\Config\App.
EDIT: Actual expected behavior: the full URL (including segments) of the page where the function is executed. As pointed out by @dafriend .

Steps to reproduce: 
 - Call current_url in a view.

**Context**
 - OS: Mac OS X 10.13.6
 - Web server: Apache 2.4.33
 - PHP version: 7.2.14

lonnieezell:
- base_url fixed to ensure a trailing slash when it needs it to property resolve relative URIs
- current_url fixed to merge any folders in the request URI's path. 

Fixes #2365 
Fixes #2347

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -21,7 +21,7 @@ class App extends BaseConfig
 	| environments.
 	|
 	*/
-	public $baseURL = 'http://localhost:8080';
+	public $baseURL = 'http://localhost:8080/';
 
 	/*
 	|--------------------------------------------------------------------------
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -174,6 +174,11 @@ class IncomingRequest extends Request
 
 		$this->populateHeaders();
 
+		// Get our current URI.
+		// NOTE: This WILL NOT match the actual URL in the browser since for
+		// everything this cares about (and the router, etc) is the portion
+		// AFTER the script name. So, if hosted in a sub-folder this will
+		// appear different than actual URL. If you need that, use current_url().
 		$this->uri = $uri;
 
 		$this->detectURI($config->uriProtocol, $config->baseURL);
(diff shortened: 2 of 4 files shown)


MGatner:
Wow that was a mess. Who knew it would be so complicated? Thanks for the digging!

lonnieezell:
@MGatner Yeah, it was a bit of a pain to track down. I got the base_url fix relatively easy, but the other took me down a few wrong turns.  Hopefully this fixes it.

PR #2408, opened 2019-11-15
Ensure previous_url() gets accurate URI.

### LINKED ISSUE #2378: previous_url() not loading the base path together
Hello I had issue with the previous_url() helper when I try to use 
```php
redirect()->back()
```
It does not load the previous url with the base path. For example I was at http://localhost/scriptfolder/admin/auth/login
So when error trigger on form validation redirect back to the previous url as http://localhost/admin/auth/login. So I discovered that the complete url does not session with the scriptfolder together so I did some customization to the Core script which work for me by adding
```php
$uri = str_replace('https://'. $_SERVER['HTTP_HOST'].'/', '', $uri);
$uri = str_replace('http://'. $_SERVER['HTTP_HOST'].'/', '', $uri);
$uri = (string) site_url($uri);
```
to the
```php
vendor/codeigniter4/framework/system/CodeIgniter.php
```
to this

```php
        /**
	 * If we have a session object to use, store the current URI
	 * as the previous URI. This is called just prior to sending the
	 * response to the client, and will make it available next request.
	 *
	 * This helps provider safer, more reliable previous_url() detection.
	 *
	 * @param \CodeIgniter\HTTP\URI $uri
	 */
	{
		// Ignore CLI requests
		if (is_cli())
		{
			return;
		}
		// Ignore AJAX requests
		if (method_exists($this->request, 'isAJAX') && $this->request->isAJAX())
		{
			return;
		}

		// This is mainly needed during testing...
		if (is_string($uri))
		{
			$uri = new URI($uri);
		}
		
                // Here is where I edited to mine
		$uri = str_replace('https://'. $_SERVER['HTTP_HOST'].'/', '', $uri);
		$uri = str_replace('http://'. $_SERVER['HTTP_HOST'].'/', '', $uri);
		
		$uri = (string) site_url($uri);

                // End of my edit

		if (isset($_SESSION))
		{
			$_SESSION['_ci_previous_url'] = (string) $uri;
		}
	}

```
which I know is not good practice so 
I don't know if I should say I am suggesting touching the Codeigniter.php to this my edit or any other best method you wish

lonnieezell:
When storing previous url we should use current_url() to ensure we ge…t baseURL paths and any query vars. 

Fixes #2378

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -373,7 +373,7 @@ class CodeIgniter
 
 		// Save our current URI as the previous URI in the session
 		// for safer, more accurate use with `previous_url()` helper function.
-		$this->storePreviousURL($this->request->uri ?? $uri);
+		$this->storePreviousURL((string)current_url(true));
 
 		unset($uri);
 


lonnieezell:
Thanks! Was letting tests run last night as I crashed. So I'm glad you commented to remind me :)

Issue #2318, opened 2019-10-10
Bug: Cannot pass a string param with a space to a custom parser plugin


SteeveDroz:
**Describe the bug**
When passing parameters to a custom made plugin in the parser, if a space is present, each word is a different value in the array.

**CodeIgniter 4 version**
v4.0.0-rc.2.1

**Affected module(s)**
Parser

**Expected behavior, and steps to reproduce if appropriate**

Here is my code:

### app/Views/pages/home.html
```html
{+header title="My website"+}
<h1>Welcome to m website!</h1>
{+footer+}
```
### in app/Config/View.php
```php
public $plugins = [
    'header' => 'App\Controllers\BaseController::header',
    'footer' => 'App\Controllers\BaseController::footer'
];
```

### in app/Controllers/BaseController.php
```php
public static function header($params = [])
{
    var_dump($params);
    …
}
```
### Value of the `var_dump()`

```
array(2) {
    ["title"]=>
    string(4) ""My"
    [0]=>
    string(8) "website""
}
```
### Expected value of the `var_dump()`

```
array(1) {
    ["title"]=>
    string(10) "My website"
}
```

**Context**
 - OS: Linux Fedora 30
 - Web server Apache 2.4.41
 - PHP version 7.3.10

Instrye:
I am trying to fix

lonnieezell:
Hmm. `str_getcsv` should use the double quote as a default enclosure symbol, so it should keep items within quotes grouped already.

lonnieezell:
According to [this repl](https://repl.it/@lonnieezell/DutifulSerpentineActivemovie), `str_getcsv` is not the problem.

PR #2411, opened 2019-11-17
#2318 - fix parse params of plugin

oleg1540:
**Description**
Fix bug #2318 
Replace **str_getcsv()** function by regexp.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -762,28 +762,15 @@ class Parser extends View
 			{
 				$params = [];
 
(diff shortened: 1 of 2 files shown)


lonnieezell:
I hesitate on this one only because that's a pretty gnarly regex pattern that takes a bit to decipher what it's doing. I wish I knew why `str_getcsv()` wasn't working in this condition. Will likely merge anyway, but want to dig a little deeper first.

PR #2427, opened 2019-11-22
Add $recipients property to Config\Email

dafriend:
Needed so that a ``recipient`` can be defined in .env and will be picked up in BaseConfig.

Most useful for automated testing situations.


MGatner:
Makes sense. It's a little weird the way this class handles config, probably because of the CI3 port, but something to address with the refactor.

PR #2435, opened 2019-12-04
Fix XSS

MashinaMashina:
Add esc

Fix XSS in session and variables. Now I see this: https://prnt.sc/q69wnx

### DIFF
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -136,7 +136,7 @@ class Toolbar
 			{
 				foreach ($items as $key => $value)
 				{
-					$varData[esc($key)] = is_string($value) ? esc($value) : print_r($value, true);
+					$varData[esc($key)] = is_string($value) ? esc($value) : esc(print_r($value, true));
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


lonnieezell:
I don't think this works. Esc only works on strings I and it should be handled for that strings based on the line you changed. 

Not sure exactly what code situation this is fixing for, but doesn't initially seem like a correct implementation. Can you provide code you're using that has made that present in the toolbar to begin with?

MashinaMashina:
esc support arrays https://codeigniter4.github.io/userguide/general/common_functions.html

> If $data is an array, then it loops over it, escaping each ‘value’ of the key/value pairs.

View data has HTML code https://prnt.sc/q6ac9r

$_GET data using esc function, but varData dont using

MashinaMashina:
Or we can use like that print_r(esc($value), true), but with using esc(print_r($value, true))

MashinaMashina:
Use $_SESSION[1] = "<H1>XSS</H1>";

Or use setVar with html code

MashinaMashina:
@lonnieezell merge pls this commit to CI

lonnieezell:
Already fixed in 68026f8b68d0a81f67520eb9ba9154ed8f471aac

PR #2451, opened 2019-12-11
Extending Route config / more flexibility for route_to and other functions / routung to different domains/proto/ports

nowackipawel:
Allows to add route which is not in the same domain/subdomain.
Allows to generate correct domain with route_to('doc_item','tx') -> ie. https://doc.domain.tld/i/tx 
which is defined like i.e.:

		$routes->get('', '\App\Controllers\Site\CDoc::index', ['subdomain' => 'doc', 'as' => 'doc_index']);
		$routes->get('c/(:any)', '\App\Controllers\Site\CDoc::index/$1', ['subdomain' => 'doc', 'as' => 'doc_category']);
		$routes->get('i/(:any)', '\App\Controllers\Site\CDoc::item/$1', ['subdomain' => 'doc', 'as' => 'doc_item']);

 or

	$routes->group(
	'/',
	['hostname' => 'doc.domain.tld', 'enabled' => true],
	function ($routes) {
		$routes->get('', '\App\Controllers\Site\CDoc::index', ['as' => 'doc_index']);
		$routes->get('c/(:any)', '\App\Controllers\Site\CDoc::index/$1', ['as' => 'doc_category']);
		$routes->get('i/(:any)', '\App\Controllers\Site\CDoc::item/$1', ['as' => 'doc_item']);
	}
	);

Adding support for 
$options['proto' => 'protocol for route' , 'port' => 'custom port', 'enabled' => 'enabling route even if it is in different domain/subdomain than requested page']

I will not be able to write the tests :/

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1251,7 +1251,7 @@ class RouteCollection implements RouteCollectionInterface
 				    $prefix = $this->getProto($collection[$search]['options']) .
 					$collection[$search]['options']['subdomain'] .
 					str_replace($old, '', $_SERVER['HTTP_HOST']) .
-					$this->getPort($collection[$search]['options']) . '/';
+					$this->getPort($collection[$search]['options']);
 				}
 				else if(!empty($collection[$search]['options']['hostname']))
 				{
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1246,13 +1246,12 @@ class RouteCollection implements RouteCollectionInterface
 			if (array_key_exists($search, $collection))
 			{
 				$prefix = '';
-				if(!empty($collection[$search]['options']['subdomain']))
+				if(!empty($collection[$search]['options']['subdomain']) && $collection[$search]['options']['subdomain'] !== ($old = $this->determineCurrentSubdomain()))
 				{
-				    $old = $this->determineCurrentSubdomain();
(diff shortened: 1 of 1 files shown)


MGatner:
@nowackipawel I see you’ve been working hard at this one! I haven’t had a chance to look closely at it, but just a reminder that a PR like this will definitely need test and updates it the User Guide. It also helps the maintainers if you tag a PR “WIP” until you are confident it is ready for review.

lonnieezell:
In general I like the idea, but have some issues with the implementation. 

From a quick glance: 

1. Functions like `getProto` and `getPath` don't belong in the RouteCollection class. You can already get that information from the Request's URI object. We don't want to maintain the same functionality in more than one place. 
2. No need to abbreviate `protocol` as `proto`. Spell the whole word out, it's not that long. 
3. I don't like the `enabled` option, and specifically don't like that it only applies to hostname settings. 

However, unless someone is willing to champion this one, we can't accept the PR without docs and a test and I see that you said you won't be able to write tests, so I'm going to have to close this one.

nowackipawel:
@lonnieezell : please take a look, I think prepared it according to your advices from PR#2451 , if you will find it is ok I will prepare version for merging with main RouteCollection class.

PR #2452, opened 2019-12-12
[ci skip] Error correction in reference to Query Builder emptyTable m…

jlamim:
**Description**
Fixed an error referencing the emptyTable method in the Query Builder documentation, where the method name was empty_table, when the correct one in the new version is emptyTable.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -1142,7 +1142,7 @@ the data to the first parameter of the function::
 	// WHERE id = $id
 
 If you want to delete all data from a table, you can use the truncate()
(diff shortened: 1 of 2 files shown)


MGatner:
@jlamim This looks good but you need to sign your commits! I cannot merge a PR while it has unverified commits.

jlamim:
@MGatner, I added a new commit with a documentation update in the migrations part where in the reference table at the end of the documentation it referred to the storage path configuration, but it is not in the "/app/Config/Migrations.php" configuration file and this information could be confusing to readers.

PR #2462, opened 2019-12-24
 Fix impossible length for migration table id.

ytetsuro:

**Description**

id column is too long.

https://dev.mysql.com/doc/refman/5.6/en/integer-types.html

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)


MGatner:
I agree the constraint is a mistake (probably copy-paste error) but BIGINT seems overkill. I suppose migrations will never take much storage either way though.

Issue #2467, opened 2019-12-28
Bug: Move work with $_GET['page'] from Model to Pager


MashinaMashina:
required move from Model::paginate() line
`$page = $page >= 1 ? $page : (ctype_digit($_GET['page'] ?? '') && $_GET['page'] > 1 ? $_GET['page'] : 1);` to Pager, because it pager business logic.
To Model::paginate() we must use $pager->getCurrentPage()

MashinaMashina:
I will try to create PR

MashinaMashina:
https://github.com/codeigniter4/CodeIgniter4/pull/2468

MashinaMashina:
Already fixed

PR #2468, opened 2019-12-28
Fix that Pager::getCurrentPage() try to return string type

MashinaMashina:
https://github.com/codeigniter4/CodeIgniter4/issues/2467

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1122,14 +1122,13 @@ class Model
 	 */
 	public function paginate(int $perPage = 20, string $group = 'default', int $page = 0)
 	{
-		// Get the necessary parts.
-		$page = $page >= 1 ? $page : (ctype_digit($_GET['page'] ?? '') && $_GET['page'] > 1 ? $_GET['page'] : 1);
+		$pager = \Config\Services::pager();
+		$page = $page >= 1 ? $page : $pager->getCurrentPage($group);
 
 		$total = $this->countAllResults(false);
 
 		// Store it in the Pager library so it can be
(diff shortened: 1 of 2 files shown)


MGatner:
I haven't dug into this yet, but isn't using `getCurrentPage()` in the Model circular since we are using it to store the page? Sorry if I have misunderstood.

Also I'd recommend casting instead of using `intval()` if all you're doing is ensuring a string is an int: `(int) $page`

lonnieezell:
Looking at this and I think there's a bigger issue here, actually. With the Model always accessing a shared instance of the pager, if you have multiple models doing pagination on a single page it is going to have issues. 

First step, then, is to get a new instance of the pager: 

```
$pager = \Config\Services::pager(null, null, true);
```

Then an automated way to handle a unique pager name is needed, with the ability for a developer to set the $_GET variable name to differentiate between multiple pagers.

lonnieezell:
@MashinaMashina are you able to make changes to this?

MashinaMashina:
@lonnieezell okay, soon I'll send this.

But instead $pager = \Config\Services::pager(null, null, true); it must be $pager = \Config\Services::pager(null, null, false);

lonnieezell:
@MashinaMashina any opportunity to make these changes in the next week or two?

MGatner:
@MashinaMashina it looks to me like this was covered by #2489 (which is now showing conflicts here). Can this PR be closed?

lonnieezell:
Yes, pretty sure this was fixed by that commit. Closing.

PR #2482, opened 2020-01-10
Services request add URI Core System extend support

byazrail:
URI Core System extending

**Description**
Services::request function changed.
static::uri() was used instead of the new URI() library.
The reason is;
If we want to extend the 'URI' library, we need to extend the 'request' service.

**Checklist:**
- [ ] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


lonnieezell:
Please have future pull request signed.

PR #2487, opened 2020-01-14
Feature Add Oracle driver

ytetsuro:
**Description**
I wanted to use CodeIgniter4 in an Oracle environment.
Also, it was determined that support would be provided because ORACLE was described in the document.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

The design of the program to get the insertId of this `PR` is not so good.

Please good solution.

### DIFF
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -91,7 +91,7 @@ class Builder extends BaseBuilder
             $sql .= '	INTO ' . $table . ' (' . $insertKeys . ') VALUES ' . $value . "\n";
         }
 
-        return $sql . 'SELECT * FROM dual';
+        return $sql . 'SELECT * FROM DUAL';
     }
 
     /**
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -63,13 +63,7 @@ class Builder extends BaseBuilder
     protected $db;
 
     /**
-     * Insert batch statement
-     *
      * Generates a platform-specific insert string from the supplied data.
-     *
-     * @param string $table  Table name
-     * @param array  $keys   INSERT keys
-     * @param array  $values INSERT values
      */
     protected function _insertBatch(string $table, array $keys, array $values): string
     {
@@ -101,13 +95,7 @@ class Builder extends BaseBuilder
     }
 
     /**
-     * Replace statement
-     *
      * Generates a platform-specific replace string from the supplied data
-     *
-     * @param string $table  The table name
-     * @param array  $keys   The insert keys
-     * @param array  $values The insert values
      */
     protected function _replace(string $table, array $keys, array $values): string
     {
@@ -162,14 +150,10 @@ class Builder extends BaseBuilder
     }
 
     /**
-     * Truncate statement
-     *
      * Generates a platform-specific truncate string from the supplied data
      *
      * If the database does not support the truncate() command,
      * then this method maps to 'DELETE FROM table'
-     *
-     * @param string $table The table name
      */
     protected function _truncate(string $table): string
     {
@@ -177,12 +161,7 @@ class Builder extends BaseBuilder
     }
 
     /**
-     * Delete
-     *
      * Compiles a delete string and runs the query
-     *
-     * @param mixed $where The where clause
-     * @param int   $limit The limit clause
      *
      * @throws DatabaseException
      *
@@ -198,11 +177,7 @@ class Builder extends BaseBuilder
     }
 
     /**
-     * Delete statement
-     *
      * Generates a platform-specific delete string from the supplied data
-     *
-     * @param string $table The table name
      */
     protected function _delete(string $table): string
     {
@@ -215,12 +190,7 @@ class Builder extends BaseBuilder
     }
 
     /**
-     * Update statement
-     *
      * Generates a platform-specific update string from the supplied data
-     *
-     * @param string $table  the Table name
-     * @param array  $values the Update data
      */
     protected function _update(string $table, array $values): string
     {
@@ -240,11 +210,7 @@ class Builder extends BaseBuilder
     }
 
     /**
-     * LIMIT string
-     *
      * Generates a platform-specific LIMIT clause.
-     *
-     * @param string $sql SQL Query
      */
     protected function _limit(string $sql, bool $offsetIgnore = false): string
     {
@@ -274,13 +240,7 @@ class Builder extends BaseBuilder
     }
 
     /**
-     * Insert statement
-     *
      * Generates a platform-specific insert string from the supplied data
-     *
-     * @param string $table         The table name
-     * @param array  $keys          The insert keys
-     * @param array  $unescapedKeys The insert values
      */
     protected function _insert(string $table, array $keys, array $unescapedKeys): string
     {
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -240,21 +240,4 @@ class Builder extends BaseBuilder
         $this->limitUsed = false;
         parent::resetSelect();
     }
-
-    /**
-     * Generates a platform-specific insert string from the supplied data
-     */
-    protected function _insert(string $table, array $keys, array $unescapedKeys): string
-    {
-        $this->db->lastInsertedTableName = $table;
-        $sql                             = 'INSERT %sINTO %s (%s) VALUES (%s) RETURNING ROWID INTO :CI_OCI8_ROWID';
-
-        return sprintf(
-            $sql,
-            $this->compileIgnore('insert'),
-            $table,
-            implode(', ', $keys),
-            implode(', ', $unescapedKeys)
-        );
-    }
 }
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -76,7 +76,7 @@ class Builder extends BaseBuilder
             $selectQueryValues = [];
 
             foreach ($values as $value) {
-                $selectValues = implode(',', array_map(static fn($value, $key) => $value . ' as ' . $key, explode(',', substr(substr($value, 1), 0, -1)), $keys));
+                $selectValues        = implode(',', array_map(static fn ($value, $key) => $value . ' as ' . $key, explode(',', substr(substr($value, 1), 0, -1)), $keys));
                 $selectQueryValues[] = 'SELECT ' . $selectValues . ' FROM DUAL';
             }
 
@@ -97,7 +97,7 @@ class Builder extends BaseBuilder
      */
     protected function _replace(string $table, array $keys, array $values): string
     {
-        $fieldNames = array_map(static fn($columnName) => trim($columnName, '"'), $keys);
+        $fieldNames = array_map(static fn ($columnName) => trim($columnName, '"'), $keys);
 
         $uniqueIndexes = array_filter($this->db->getIndexData($table), static function ($index) use ($fieldNames) {
             $hasAllFields = count(array_intersect($index->fields, $fieldNames)) === count($index->fields);
@@ -116,7 +116,7 @@ class Builder extends BaseBuilder
 
         $sql = 'MERGE INTO ' . $table . "\n USING (SELECT ";
 
-        $sql .= implode(', ', array_map(static fn($columnName, $value) => $value . ' ' . $columnName, $keys, $values));
+        $sql .= implode(', ', array_map(static fn ($columnName, $value) => $value . ' ' . $columnName, $keys, $values));
 
         $sql .= ' FROM DUAL) "_replace" ON ( ';
 
@@ -124,15 +124,15 @@ class Builder extends BaseBuilder
         $onList[] = '1 != 1';
 
         foreach ($uniqueIndexes as $index) {
-            $onList[] = '(' . implode(' AND ', array_map(static fn($columnName) => $table . '."' . $columnName . '" = "_replace"."' . $columnName . '"', $index->fields)) . ')';
+            $onList[] = '(' . implode(' AND ', array_map(static fn ($columnName) => $table . '."' . $columnName . '" = "_replace"."' . $columnName . '"', $index->fields)) . ')';
         }
 
         $sql .= implode(' OR ', $onList) . ') WHEN MATCHED THEN UPDATE SET ';
 
-        $sql .= implode(', ', array_map(static fn($columnName) => $columnName . ' = "_replace".' . $columnName, $replaceableFields));
+        $sql .= implode(', ', array_map(static fn ($columnName) => $columnName . ' = "_replace".' . $columnName, $replaceableFields));
 
         $sql .= ' WHEN NOT MATCHED THEN INSERT (' . implode(', ', $replaceableFields) . ') VALUES ';
-        $sql .= ' (' . implode(', ', array_map(static fn($columnName) => '"_replace".' . $columnName, $replaceableFields)) . ')';
+        $sql .= ' (' . implode(', ', array_map(static fn ($columnName) => '"_replace".' . $columnName, $replaceableFields)) . ')';
 
         return $sql;
     }
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -163,6 +163,8 @@ class Builder extends BaseBuilder
     /**
      * Compiles a delete string and runs the query
      *
+     * @param mixed $where
+     *
      * @throws DatabaseException
      *
      * @return mixed
@@ -224,7 +226,7 @@ class Builder extends BaseBuilder
             return $sql . ' OFFSET ' . (int) $offset . ' ROWS FETCH NEXT ' . $this->QBLimit . ' ROWS ONLY';
         }
 
-        $this->limitUsed = true;
+        $this->limitUsed    = true;
         $limitTemplateQuery = 'SELECT * FROM (SELECT INNER_QUERY.*, ROWNUM RNUM FROM (%s) INNER_QUERY WHERE ROWNUM < %d)' . ($offset ? ' WHERE RNUM >= %d' : '');
 
         return sprintf($limitTemplateQuery, $sql, $offset + $this->QBLimit + 1, $offset);
@@ -245,7 +247,7 @@ class Builder extends BaseBuilder
     protected function _insert(string $table, array $keys, array $unescapedKeys): string
     {
         $this->db->lastInsertedTableName = $table;
-        $sql = 'INSERT %sINTO %s (%s) VALUES (%s) RETURNING ROWID INTO :CI_OCI8_ROWID';
+        $sql                             = 'INSERT %sINTO %s (%s) VALUES (%s) RETURNING ROWID INTO :CI_OCI8_ROWID';
 
         return sprintf(
             $sql,
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -244,7 +244,7 @@ class Builder extends BaseBuilder
      */
     protected function _insert(string $table, array $keys, array $unescapedKeys): string
     {
-        $this->db->latestInsertedTableName = $table;
+        $this->db->lastInsertedTableName = $table;
         $sql = 'INSERT %sINTO %s (%s) VALUES (%s) RETURNING ROWID INTO :CI_OCI8_ROWID';
 
         return sprintf(
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -76,9 +76,7 @@ class Builder extends BaseBuilder
             $selectQueryValues = [];
 
             foreach ($values as $value) {
-                $selectValues = implode(',', array_map(static function ($value, $key) {
-                    return $value . ' as ' . $key;
-                }, explode(',', substr(substr($value, 1), 0, -1)), $keys));
+                $selectValues = implode(',', array_map(static fn($value, $key) => $value . ' as ' . $key, explode(',', substr(substr($value, 1), 0, -1)), $keys));
                 $selectQueryValues[] = 'SELECT ' . $selectValues . ' FROM DUAL';
             }
 
@@ -99,9 +97,7 @@ class Builder extends BaseBuilder
      */
     protected function _replace(string $table, array $keys, array $values): string
     {
-        $fieldNames = array_map(static function ($columnName) {
-            return trim($columnName, '"');
-        }, $keys);
+        $fieldNames = array_map(static fn($columnName) => trim($columnName, '"'), $keys);
 
         $uniqueIndexes = array_filter($this->db->getIndexData($table), static function ($index) use ($fieldNames) {
             $hasAllFields = count(array_intersect($index->fields, $fieldNames)) === count($index->fields);
@@ -120,9 +116,7 @@ class Builder extends BaseBuilder
 
         $sql = 'MERGE INTO ' . $table . "\n USING (SELECT ";
 
-        $sql .= implode(', ', array_map(static function ($columnName, $value) {
-            return $value . ' ' . $columnName;
-        }, $keys, $values));
+        $sql .= implode(', ', array_map(static fn($columnName, $value) => $value . ' ' . $columnName, $keys, $values));
 
         $sql .= ' FROM DUAL) "_replace" ON ( ';
 
@@ -130,21 +124,15 @@ class Builder extends BaseBuilder
         $onList[] = '1 != 1';
 
         foreach ($uniqueIndexes as $index) {
-            $onList[] = '(' . implode(' AND ', array_map(static function ($columnName) use ($table) {
-                return $table . '."' . $columnName . '" = "_replace"."' . $columnName . '"';
-            }, $index->fields)) . ')';
+            $onList[] = '(' . implode(' AND ', array_map(static fn($columnName) => $table . '."' . $columnName . '" = "_replace"."' . $columnName . '"', $index->fields)) . ')';
         }
 
         $sql .= implode(' OR ', $onList) . ') WHEN MATCHED THEN UPDATE SET ';
 
-        $sql .= implode(', ', array_map(static function ($columnName) {
-            return $columnName . ' = "_replace".' . $columnName;
-        }, $replaceableFields));
+        $sql .= implode(', ', array_map(static fn($columnName) => $columnName . ' = "_replace".' . $columnName, $replaceableFields));
 
         $sql .= ' WHEN NOT MATCHED THEN INSERT (' . implode(', ', $replaceableFields) . ') VALUES ';
-        $sql .= ' (' . implode(', ', array_map(static function ($columnName) {
-            return '"_replace".' . $columnName;
-        }, $replaceableFields)) . ')';
+        $sql .= ' (' . implode(', ', array_map(static fn($columnName) => '"_replace".' . $columnName, $replaceableFields)) . ')';
 
         return $sql;
     }
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -223,7 +223,7 @@ class Builder extends BaseBuilder
                 $sql .= ' ORDER BY 1';
             }
 
-            return $sql . ' OFFSET ' . (int) $offset . ' ROWS FETCH NEXT ' . $this->QBLimit . ' ROWS ONLY';
+            return $sql . ' OFFSET ' . $offset . ' ROWS FETCH NEXT ' . $this->QBLimit . ' ROWS ONLY';
         }
 
         $this->limitUsed    = true;
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -225,9 +225,9 @@ class Builder extends BaseBuilder
         }
 
         $this->limitUsed = true;
+        $limitTemplateQuery = 'SELECT * FROM (SELECT INNER_QUERY.*, ROWNUM RNUM FROM (%s) INNER_QUERY WHERE ROWNUM < %d)' . ($offset ? ' WHERE RNUM >= %d' : '');
 
-        return 'SELECT * FROM (SELECT inner_query.*, rownum rnum FROM (' . $sql . ') inner_query WHERE rownum < ' . ($offset + $this->QBLimit + 1) . ')'
-            . ($offset ? ' WHERE rnum >= ' . ($offset + 1) : '');
+        return sprintf($limitTemplateQuery, $sql, $offset + $this->QBLimit + 1, $offset);
     }
 
     /**
@@ -244,10 +244,15 @@ class Builder extends BaseBuilder
      */
     protected function _insert(string $table, array $keys, array $unescapedKeys): string
     {
-        // Has a strange design.
-        // Processing to get the table where the last insert was performed for insertId method.
         $this->db->latestInsertedTableName = $table;
+        $sql = 'INSERT %sINTO %s (%s) VALUES (%s) RETURNING ROWID INTO :CI_OCI8_ROWID';
 
-        return 'INSERT ' . $this->compileIgnore('insert') . 'INTO ' . $table . ' (' . implode(', ', $keys) . ') VALUES (' . implode(', ', $unescapedKeys) . ') RETURNING ROWID INTO :CI_OCI8_ROWID';
+        return sprintf(
+            $sql,
+            $this->compileIgnore('insert'),
+            $table,
+            implode(', ', $keys),
+            implode(', ', $unescapedKeys)
+        );
     }
 }
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -205,8 +205,8 @@ class Builder extends BaseBuilder
         }
 
         return 'UPDATE ' . $this->compileIgnore('update') . $table . ' SET ' . implode(', ', $valStr)
-                . $this->compileWhereHaving('QBWhere')
-                . $this->compileOrderBy();
+            . $this->compileWhereHaving('QBWhere')
+            . $this->compileOrderBy();
     }
 
     /**
--- a/system/Database/OCI8/Builder.php
+++ b/system/Database/OCI8/Builder.php
@@ -214,7 +214,7 @@ class Builder extends BaseBuilder
      */
     protected function _limit(string $sql, bool $offsetIgnore = false): string
     {
-        $offset = (int) ($offsetIgnore === false) ? $this->QBOffset : 0;
+        $offset = (int) ($offsetIgnore === false ? $this->QBOffset : 0);
         if (version_compare($this->db->getVersion(), '12.1', '>=')) {
             // OFFSET-FETCH can be used only with the ORDER BY clause
             if (empty($this->QBOrderBy)) {
--- a/system/Database/OCI8/Connection.php
+++ b/system/Database/OCI8/Connection.php
@@ -303,7 +303,7 @@ class Connection extends BaseConnection implements ConnectionInterface
             $retval[$i]->type = $query[$i]->DATA_TYPE;
 
             $length = $query[$i]->CHAR_LENGTH > 0 ? $query[$i]->CHAR_LENGTH : $query[$i]->DATA_PRECISION;
-            $length = $length ?? $query[$i]->DATA_LENGTH;
+            $length ??= $query[$i]->DATA_LENGTH;
 
             $retval[$i]->max_length = $length;
 
--- a/system/Database/OCI8/Connection.php
+++ b/system/Database/OCI8/Connection.php
@@ -635,9 +635,7 @@ class Connection extends BaseConnection implements ConnectionInterface
              * Easy Connect string from these 3 settings, assuming
              * that the database field is a service name.
              */
-            $this->DSN = $this->hostname
-                . $easyConnectablePort
-                . $easyConnectableDatabase;
+            $this->DSN = $this->hostname . $easyConnectablePort . $easyConnectableDatabase;
 
             if (preg_match($this->validDSNs['ec'], $this->DSN)) {
                 return;
--- a/system/Database/OCI8/Connection.php
+++ b/system/Database/OCI8/Connection.php
@@ -624,8 +624,8 @@ class Connection extends BaseConnection implements ConnectionInterface
         }
 
         $isEasyConnectableHostName = $this->hostname !== '' && strpos($this->hostname, '/') === false && strpos($this->hostname, ':') === false;
-        $easyConnectablePort       = ((! empty($this->port) && ctype_digit($this->port)) ? ':' . $this->port : '');
-        $easyConnectableDatabase   = ($this->database !== '' ? '/' . ltrim($this->database, '/') : '');
+        $easyConnectablePort       = ! empty($this->port) && ctype_digit($this->port) ? ':' . $this->port : '';
+        $easyConnectableDatabase   = $this->database !== '' ? '/' . ltrim($this->database, '/') : '';
 
         if ($isEasyConnectableHostName && ($easyConnectablePort !== '' || $easyConnectableDatabase !== '')) {
             /* If the hostname field isn't empty, doesn't contain
--- a/system/Database/OCI8/Connection.php
+++ b/system/Database/OCI8/Connection.php
@@ -594,6 +594,7 @@ class Connection extends BaseConnection implements ConnectionInterface
 
             if ($primaryColumnType !== 'NUMBER') {
                 $primaryColumnName = '';
+
                 continue;
             }
         }
--- a/system/Database/OCI8/Connection.php
+++ b/system/Database/OCI8/Connection.php
@@ -198,7 +198,7 @@ class Connection extends BaseConnection implements ConnectionInterface
 
             oci_set_prefetch($this->stmtId, 1000);
 
-            $result = oci_execute($this->stmtId, $this->commitMode) ? $this->stmtId : false;
+            $result          = oci_execute($this->stmtId, $this->commitMode) ? $this->stmtId : false;
             $insertTableName = $this->parseInsertTableName($sql);
 
             if ($result && $insertTableName !== '') {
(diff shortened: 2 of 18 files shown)


lonnieezell:
Thanks so much for submitting this! You rock for that. I won't be able to get to this until after the next RC release, honestly. Besides reviewing this, we need to get Travis setup to be able to test it if possible. I'm honestly not familiar enough with Oracle to know if there's a simple to setup community version. Looks like there's Express Edition but would have to look deeper to see if we could integrate.

ytetsuro:
@lonnieezell 
Thanks feedback.

> Looks like there's Express Edition but would have to look deeper to see if we could integrate.

Humm...
it may be difficult.

https://github.com/oracle/docker-images/issues/1156

Is there anything I can do to help with the this review?

yogahilmi:
I tried using this driver with Oracle 18C, it works fine.

Thank you @ytetsuro

kenjis:
@ytetsuro @lonnieezell 
This repo has GitHub workflows with Oracle image.
https://github.com/yiisoft/db-oracle

michalsn:
I was working on tests for this at some point and the main problem I had, was how to set an empty database because by default oracle comes with a lot of additional tables that break our tests. Also, I think we need at least Oracle 12 to support "auto-increment" values.

Maybe this thing with a fresh database is trivial but my DevOps skills are low 😅

jkashwin:
Any update on the availability of OCI8 driver support for CI4 ?

javawebmedia:
I am really waiting for this Codeigniter 4 Oracle driver. Hopefully, it can be released soon. Thank you

ytetsuro:
@kenjis 

> This repo has GitHub workflows with Oracle images.

Very thanks.

Using the same idea, I checked the containers used in version 12c and above and found that `quillbuilduser/oracle-18-xe` was used.

The following repositories use this container.

- https://github.com/getquill/quill
- https://github.com/knex/knex/blob/01cfa98c2cdd9d2f8c91320d4551efdaa1768d21/scripts/docker-compose.yml
- https://github.com/JetBrains/Exposed/blob/b6d00e51062c91c6b23a89de55f900e2e4f24225/buildScripts/docker/docker-compose-oracle.yml

@lonnieezell 
@kenjis 

> we need to get Travis setup to be able to test it if possible

Changed GitHub Actions so that the OCI8 tests work with the above container.
Please review.

system/Database/OCI8/Builder.php
```diff
@@ -0,0 +1,293 @@
...
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Database\OCI8;
+
+use CodeIgniter\Database\BaseBuilder;
+use CodeIgniter\Database\Exceptions\DatabaseException;
+
+/**
+ * Builder for MySQLi
```
kenjis, line 18:
```suggestion
 * Builder for OCI8
```

system/Database/OCI8/Forge.php
```diff
@@ -0,0 +1,310 @@
...
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Database\OCI8;
+
+/**
+ * Forge for MySQLi
```
kenjis, line 15:
```suggestion
 * Forge for OCI8
```

system/Database/OCI8/PreparedQuery.php
```diff
@@ -0,0 +1,134 @@
...
+ */
+
+namespace CodeIgniter\Database\OCI8;
+
+use BadMethodCallException;
+use CodeIgniter\Database\BasePreparedQuery;
+use CodeIgniter\Database\PreparedQueryInterface;
+
+/**
+ * Prepared query for MySQLi
```
kenjis, line 19:
```suggestion
 * Prepared query for OCI8
```

system/Database/OCI8/Result.php
```diff
@@ -0,0 +1,122 @@
...
+
+namespace CodeIgniter\Database\OCI8;
+
+use CodeIgniter\Database\BaseResult;
+use CodeIgniter\Database\ResultInterface;
+use CodeIgniter\Entity;
+use stdClass;
+
+/**
+ * Result for OCI
```
kenjis, line 20:
```suggestion
 * Result for OCI8
```

system/Database/OCI8/Utils.php
```diff
@@ -0,0 +1,38 @@
...
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Database\OCI8;
+
+use CodeIgniter\Database\BaseUtils;
+use CodeIgniter\Database\Exceptions\DatabaseException;
+
+/**
+ * Utils for MySQLi
```
kenjis, line 18:
```suggestion
 * Utils for OCI8
```

kenjis:
@ytetsuro @lonnieezell @michalsn @jkashwin @javawebmedia
What version of Oracle Database should this driver support?
12.1+?
12.2+?
18c+?
other?

And we need to add about this driver in
https://codeigniter4.github.io/CodeIgniter4/intro/requirements.html

ytetsuro:
@kenjis 

Thanks for the quick feedback.

> What version of Oracle Database should this driver support?

I think `12.1+`.
I used the following version of this driver when I created it.

- 12.1
- 18c

The driver I created this time doesn't seem to use any of the features that were deprecated in 18c, so I think 12.1+ is fine.

https://docs.oracle.com/en/database/oracle/oracle-database/18/upgrd/behavior-change-deprecate-desupport-oracle-12c-rel-2.html#GUID-3BAFD95E-4D00-4F0F-BC80-6064F497F878

tests/system/Models/PaginateModelTest.php
```diff
@@ -46,7 +46,7 @@ public function testPaginatePassPerPageParameter(): void
     public function testPaginateForQueryWithGroupBy(): void
     {
         $this->createModel(ValidModel::class);
-        $this->model->groupBy('id');
+        $this->model->select('id')->groupBy('id');
```
kenjis, line 49: Why do you need `->select('id')`?

ytetsuro: @kenjis 

Thanks, response.

> Why do you need ->select('id')?

The reason is that it is grouped by id.
I think you will get `ERROR 1055 (42000)` with `only_full_group_by` in MySQL, for the same reason.
Similar to the above, Oracle cannot select items that are not group by when using group by.

I think it strange that you can select anything but id.
This is because it is illegal in standard SQL.

https://dev.mysql.com/doc/refman/5.6/en/group-by-handling.html

> For example, this query is illegal in standard SQL because the nonaggregated name column in the select list does not appear in the GROUP BY:

tests/system/Database/Live/GroupTest.php
```diff
@@ -43,7 +43,7 @@ public function testHavingBy()
         $result = $this->db->table('job')
             ->select('name')
             ->groupBy('name')
-            ->having('SUM(id) > 2')
+            ->having('SUM(' . $this->db->protectIdentifiers('id') . ') > 2')
```
kenjis, line 46: I want to remove `$this->db->protectIdentifiers()`, because it is Query Builder.
But how?

system/Database/BaseBuilder.php
```diff
@@ -1358,9 +1359,9 @@ public function set($key, $value = '', ?bool $escape = null)
             if ($escape) {
                 $bind = $this->setBind($k, $v, $escape);
 
-                $this->QBSet[$this->db->protectIdentifiers($k, false, $escape)] = ":{$bind}:";
+                $this->QBSet[$this->db->protectIdentifiers($k, false)] = ":{$bind}:";
```
kenjis, line 1362: This change changes the behavior.
Even if `$escape` is false, `protectIdentifiers()` escapes the identifier.
https://github.com/codeigniter4/CodeIgniter4/blob/c05f5a6dae2b08e9ebf82d5f7888dd8fcbff5813/system/Database/BaseConnection.php#L954-L958

Users can't stop escaping.

system/Database/BaseBuilder.php
```diff
@@ -1358,9 +1359,9 @@ public function set($key, $value = '', ?bool $escape = null)
...
             } else {
-                $this->QBSet[$this->db->protectIdentifiers($k, false, $escape)] = $v;
+                $this->QBSet[$this->db->protectIdentifiers($k, false)] = $v;
```
kenjis, line 1364: The same as above.

tests/system/Database/Live/PreparedQueryTest.php
```diff
@@ -114,8 +123,8 @@ public function testExecuteRunsQueryAndReturnsManualResultObject()
             return (new Query($db))->setQuery($sql);
         });
 
-        $query->execute('foo', 'foo@example.com', '');
-        $query->execute('bar', 'bar@example.com', '');
+        $query->execute('foo', 'foo@example.com', 'US');
+        $query->execute('bar', 'bar@example.com', 'GB');
```
kenjis, line 127: Why do you need to change the empty strings?

system/Database/BaseBuilder.php
ytetsuro: I decided that there was no case where the column name in the set clause did not need to be escaped, so I turned off the unescaping process.

Are there any use cases where the column names in the set clause do not need escaping?

tests/system/Database/Live/GroupTest.php
ytetsuro: @kenjis 

Thanks, response.

You're right.

> But how?

Hmmm…

If you are using the OCI8 driver, what do you think about doing an ALTER TABLE to rename the id column to an uppercase column name?

This method does not essentially solve the QueryBuiler problem.
However, it may be a good solution in that it does not include this problem in the scope of this PR.

It looks just like this.

https://github.com/codeigniter4/CodeIgniter4/pull/2487/commits/729acdc7df399432bd2f5398c302b367cf019553#diff-c2d356f0d870b5a4a5bf49fad5e365c8ec6b3ee9b0f5a6bcb1f879801cd372d9L206-L215

tests/system/Database/Live/PreparedQueryTest.php
ytetsuro: Thanks, good comment.

This is because Oracle treats an empty string as null and the value of the country column is NotNull, so the query will fail.
oci8 also treats empty strings as null.

https://github.com/php/php-src/blob/master/ext/oci8/tests/bind_sqltafc.phpt#L149-L175

system/Database/BaseBuilder.php
kenjis: Good point. But it seems it has nothing to do with this Oracle driver.
Can you separate the changes that are not specific to the Oracle driver as another PR?
This PR is too big and takes too much time to review.

system/Database/Forge.php
```diff
@@ -466,6 +466,8 @@ public function createTable(string $table, bool $ifNotExists = false, array $att
 
                 return false;
             }
+
+            return true;
```
kenjis, line 470: Is this a required change only for Oracle drivers?

tests/system/Database/Live/GroupTest.php
kenjis: How about converting to uppercase when creating tables?

kenjis: No. Query Builder basically quotes identifier like `"id"`
when `$builder->where('id', 3);`

system/Database/BaseBuilder.php
ytetsuro: Sorry to make your review so difficult.
I will split it up.💪

system/Database/Forge.php
ytetsuro: You're right!
I created a PR.
However, the only pattern where the `createTableIfStr` property is false is for SQLite less than 3.3, except for OCI8.
Therefore, it may be easier to maintain if you fix it in this PR as the test code is a little difficult to convey the intention as shown below.
Maybe I'm just a bad test code.

https://github.com/codeigniter4/CodeIgniter4/blob/f564f6b7eca2504ac894bbaf7c3b6d18ce9b7b85/tests/system/Database/Forge/CreateTableTest.php#L23-L40

kenjis: #5133 was merged. This change is no longer necessary.

tests/system/Database/Live/PreparedQueryTest.php
kenjis: You're correct.
Oracle treats an empty string as null.

tests/system/Database/Live/GroupTest.php
ytetsuro: @kenjis 

Thanks, response.

> No. Query Builder basically quotes identifiers like "id".

No. can’t escape identifier in parenthesis.

https://github.com/codeigniter4/CodeIgniter4/blob/4649204306276af6c06bfbd229fb1895fb5b0e4e/system/Database/BaseConnection.php#L970-L979

> How about converting to uppercase when creating tables?
> If you are using the OCI8 driver, what do you think about doing an ALTER TABLE to rename the id column to an uppercase column name?

I don't think I can solve either of these problems.
Oracle can handle lowercase letters as long as they are escaped.
However, if they are not escaped, they are treated as uppercase.
In other words, it will fail in the following cases.

https://github.com/codeigniter4/CodeIgniter4/pull/2487/files#diff-4d07390c9284f69a1dca0fd6dd95b0ba949997b35d1266828f0619bef65c7c22R58-R59

In the first having, the id is escaped and treated as lowercase.

The implementation I have in mind now is the following.

1. parse AST from SQL using a library like `PhpMyAdmin\SqlParser`.
2. escape identifiers from the AST.

In this case, we have the following concerns
I don't think it is desirable to run a heavy process many times for issuing a single query, what do you think?

kenjis: This is a issue in the Query Builder. There is no complete SQL statement to parse.
The current implementation of `protectIdentifiers()` may receive the string like a sub query `(SELECT MAX(advance_amount) FROM "orders" WHERE "id" > 2`, which does not have `)` at the end.

It seems to me that it is better to keep the specifications simple, maintain speed, and not do anything too complicated.

If we accept the comment as the specification of the Query Builder:
> If a parenthesis is found we know that we do not need to escape the data or add a prefix. 
> Added exception for single quotes as well, we don't want to alter literal strings.

The Query Builder never escapes items which contain `(`, `)`, and `'`.

In this case, it is recommended to write code like `->having('SUM("id") > 2')`.
No, `->having('SUM("id") >',  2)` is better.

ytetsuro: I think you are right.
However, MySQL uses the escape character "`".
Then my code will look like the following.

```php
$result = $this->db->table('job')
            ->select('name')
            ->groupBy('name')
            ->having('SUM(' . $this->db->protectIdentifiers('id') . ') >',  2);
```

In short, I wanted to erase `$this->db->protectIdentifiers()` the same way you did.
However, I can't think of a way to get rid of it.😭😭😭

kenjis: Okay. We can't solve this problem without making parser for parts of SQL (like `'SUM(id) > 2'`), 
and quotes identifiers automatically.

1. Use Quoted Identifiers
- When we create tables with *Forge* in a normal way, identifiers will be quoted.
- In this case, on Oracle, we always need to use quoted identifiers like `"name"`. 
- But MySQL uses `` ` `` to quote identifiers like `` `name` ``.

2. Use Unquoted Identifiers
- When we create tables with unquoted identifiers, Oracle considers them to be uppercase.
- Query Builder quotes identifiers in most cases.
- If we use a lowercase identifier like `name`, QB converts it to `"name"`, and there is no such identifier.

---

Oracle is a special case. I don't show code like `->having('SUM(' . $this->db->protectIdentifiers('id') . ') >',  2)`
only for Oracle.

How about using if statements to divide cases?

```php
    public function testHavingBy()
    {
        if ($this->db->DBDriver === 'OCI8') {
            $result = $this->db->table('job')
                ->select('name')
                ->groupBy('name')
                ->having('SUM("id") > 2')
                ->get()
                ->getResultArray();
        } else {
            $result = $this->db->table('job')
                ->select('name')
                ->groupBy('name')
                ->having('SUM(id) > 2')
                ->get()
                ->getResultArray();
        }

        $this->assertCount(2, $result);
    }
```

ytetsuro: @kenjis 

> How about using if statements to divide cases?

Very thanks, response.
I agree with this.
I'll modify it this way.

> only for Oracle.

I did not have enough knowledge about this.
I am sorry.

To be precise, it seems to be a specification of ANSI-SQL, not Oracle.
That is, it is ANSI-SQL compliant, and a similar implementation may be needed when implementing `Firebird`, which CodeIgniter will support.

tests/system/Database/Live/GroupTest.php
```diff
@@ -40,10 +40,13 @@ public function testGroupBy()
...
     public function testHavingBy()
     {
+        $isANSISQL   = in_array($this->db->DBDriver, ['OCI8'], true);
+        $sumIdColumn = $isANSISQL ? 'SUM("id")' : 'SUM(id)';
+
         $result = $this->db->table('job')
             ->select('name')
             ->groupBy('name')
-            ->having('SUM(id) > 2')
+            ->having($sumIdColumn . ' >', 2)
```
kenjis, line 49: I think it is not good practice that using variable in the key of *having* or *where* method of QB.

tests/system/Database/Live/GroupTest.php
kenjis: It seems you are correct. It comes from ANSI-SQL.

https://stackoverflow.com/questions/153944/is-sql-syntax-case-sensitive/19933159#19933159

tests/system/Database/Live/GroupTest.php
ytetsuro: Thanks, response.

Fixed it.

Sure!
Looking at this code makes me want to look for SQL Injection!😂

system/Database/BaseBuilder.php
ytetsuro: This change was merged into CI, so I rebased it.

system/Database/BaseBuilder.php
ytetsuro: This change was merged into CI, so I rebased it.

system/Database/Forge.php
ytetsuro: This change was merged into CI, so I rebased it.

system/Database/BaseBuilder.php
```diff
@@ -1003,7 +1003,7 @@ protected function _like($field, string $match = '', string $type = 'AND ', stri
                 $bind = $this->setBind($k, "%{$v}%", $escape);
             }
 
-            $likeStatement = $this->_like_statement($prefix, $k, $not, $bind, $insensitiveSearch);
+            $likeStatement = $this->_like_statement($prefix, $this->db->protectIdentifiers($k, false, $escape), $not, $bind, $insensitiveSearch);
```
ytetsuro, line 1006: splited #5170

system/Database/BaseBuilder.php
```diff
@@ -1260,6 +1260,7 @@ public function orderBy(string $orderBy, string $direction = '', ?bool $escape =
         if ($direction === 'RANDOM') {
             $direction = '';
             $orderBy   = ctype_digit($orderBy) ? sprintf($this->randomKeyword[1], $orderBy) : $this->randomKeyword[0];
+            $escape    = false;
```
ytetsuro, line 1263: @kenjis 

Should we split the code here?

The intention of this fix is because escaping is not necessary for RANDOM.

In ORACLE, random numbers are `DBMS_RANDOM.RANDOM`. That is, it refers to the `RANDOM` column of the `DBMS_RANDOM` table.
In MySQL, it is `RAND()`, it is function.

The third argument `$escape` of orderBy is null if it is omitted.
It depends on the `protectIdentifiers` property of Connection to toggle whether escaping is performed or not.
Also, the escaping process is to add the prefix to the table name.

In other words, in the case of ORACLE, it will add the prefix to  `DBMS_RANDOM.RANDOM` and execute a query for a non-existent table, which will fail.

It is not a good idea for users to specify the value of the third argument escape with the driver in mind.

Therefore, I am making this fix because I thought it would be better to not escape for RANDOM and add the escaped column name to the `randomKeyword` property of each driver.

This is not limited to Oracle, but in the Database supported by CodeIgniter4 at this time, RANDOM is all function calls and cannot be reproduced.

system/Database/BaseConnection.php
```diff
@@ -1269,8 +1269,12 @@ protected function _escapeString(string $str): string
...
     public function callFunction(string $functionName, ...$params): bool
     {
-        $driver = strtolower($this->DBDriver);
-        $driver = ($driver === 'postgre' ? 'pg' : $driver) . '_';
+        $driver = strtolower($this->DBDriver) . '_';
+        if ($this->DBDriver === 'postgre') {
+            $driver = 'pg_';
+        } elseif ($this->DBDriver === 'oci8') {
+            $driver = 'oci_';
+        }
```
ytetsuro, line 1277: This is the reason why the coverages are treated as being lowered.
What should I do?

It may be possible to fudge the coverage by doing the following, but it is a tricky code.

```php
$driver = strtolower($this->DBDriver);
$driver = (['postgre' => 'pg', 'oci8' => 'oci'][$driver] ?? $driver) . '_';
```

system/Database/BaseBuilder.php
kenjis: The more you split this PR, the more likely it is to be merged.
I'm not sure, how long it will take it to be merged.
Because this PR is too big for anyone to fully review.

system/Database/BaseConnection.php
kenjis: Something like this?
```diff
@@ -1269,12 +1273,7 @@ abstract class BaseConnection implements ConnectionInterface
      */
     public function callFunction(string $functionName, ...$params): bool
     {
-        $driver = strtolower($this->DBDriver) . '_';
-        if ($this->DBDriver === 'postgre') {
-            $driver = 'pg_';
-        } elseif ($this->DBDriver === 'oci8') {
-            $driver = 'oci_';
-        }
+        $driver = $this->getDriverFunctionPrefix();
 
         if (strpos($driver, $functionName) === false) {
             $functionName = $driver . $functionName;
@@ -1291,6 +1290,11 @@ abstract class BaseConnection implements ConnectionInterface
         return $functionName(...$params);
     }
 
+    protected function getDriverFunctionPrefix()
+    {
+        return strtolower($this->DBDriver) . '_';
+    }
+
     //--------------------------------------------------------------------
     // META Methods
     //--------------------------------------------------------------------
```

ytetsuro: @kenjis 

Thanks, response.
I think it's really great.
You're a god.

system/Database/BaseBuilder.php
ytetsuro: @kenjis 

I understand.
Thank you.
I will do my best to remove all diffs from this PR except `system/Database/OCI8`.💪

system/Database/BaseConnection.php
ytetsuro: splited #5178

system/Database/BaseBuilder.php
ytetsuro: splited #5179

kenjis:
@ytetsuro 
> The design of the program to get the insertId of this PR is not so good.

What's wrong with it?

ytetsuro:
@kenjis 

> What's wrong with it? 

My approach is to allow the last insert id to be retrieved only when a row is added with QueryBuilder::insert().

kenjis:
> For developers, I think it would be better if the API asked for the table name and explored the SEQUENCE internally, what do you think?

I think passing the table name is better for developers.

> My concern is that insertID() is defined as abstract in the parent class.

If we add a parameter, it is a breaking change.
https://3v4l.org/JTMSP
So we can't include the change in the next release.

We will add that feature in the future and I recommend that it not be included in this PR.

Incidentally, CI3 has the three type of `insert_id()`:
```
public function insert_id()
public function insert_id($generator_name, $inc_by = 0) ... ibase
public function insert_id($name = NULL) ... pdo
```

kenjis:
> Would it still be a disruptive change if it does not affect existing users by using the default argument feature as follows?

It is still a breaking change. If a user extends BaseConnection, it breaks.

https://3v4l.org/lnu0u

ytetsuro:
@kenjis 

Thanks, good response.

> If a SQL statement is INSERT, we get the table name. To do so, we need to parse the statement.

Good!
If it's an INSERT, the subquery won't be the target, so it'll be easy to parse the table name!

But, I have discovered a new problem.😭
The way we are thinking now, we will get an unintended id.
SEQUENCE is not unique for a connection.
This means that if you and I execute a write operation at the same time, and I get the last id, my process may get the id that you wrote.🤮

MySQL's last_insert_id is also set on a per-connection basis.

> The ID that was generated is maintained in the server on a per-connection basis.

https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_last-insert-id

With this in mind, at the moment I can't think of a way to get the last id with the query() method.😭😭

tests/system/Database/Live/OCI8/LastInsertIDTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $this->assertSame($actual, 5);
+    }
+
+    public function testGetInsertIDWithHasCommentQuery()
+    {
+        $sql = <<<SQL
+-- INSERT INTO "db_misc" ("key", "value") VALUES ('key', 'value')
+/* INSERT INTO "db_misc" ("key", "value") VALUES ('key', 'value') */
+INSERT /* INTO "db_misc" */ INTO "db_job"  ("name", "description") VALUES (' INTO "abc"', ?)
+SQL;
```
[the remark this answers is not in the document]
ytetsuro: @kenjis 

Thanks, good response.

Fixed it!

system/Database/OCI8/Connection.php
```diff
@@ -0,0 +1,730 @@
...
+
+    /**
+     * Get the table name for the insert statement from sql.
+     *
+     * @param string $sql
+     *
+     * @return string
+     */
+    public function parseInsertTableName(string $sql): string {
+        $commentStrippedSql = preg_replace(['/\/\*(.|\n)*?\*\//m', '/^-- .*/'], '', $sql);
```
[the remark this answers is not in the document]
ytetsuro: Thanks, response.

I fixed it by referring to the attached link.

kenjis:
@ytetsuro Thanks!

It seems all the problems except the [bug on storedProcedure()](https://forum.codeigniter.com/thread-76381-post-390991.html#pid390991) are solved.

kenjis:
```
There were 104 errors:

1) CodeIgniter\Database\Live\OCI8\CallStoredProcedureTest::testCallPackageProcedure
ErrorException: oci_execute(): ORA-06550: line 1, column 7:
PLS-00201: identifier 'CALCULATOR.PLUS' must be declared
ORA-06550: line 1, column 7:
PL/SQL: Statement ignored

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/OCI8/Connection.php:201
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:670
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:598
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/OCI8/Connection.php:520
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Database/Live/OCI8/CallStoredProcedureTest.php:51
```
https://github.com/codeigniter4/CodeIgniter4/runs/4568866943?check_suite_focus=true

ytetsuro:
@kenjis 

Hi.
Very thanks response and long-time review.

> It seems all the problems except the bug on storedProcedure() are solved.

Fixed it.

Please help me.
I want to create stored procedure in migration.
But, happen following error when using oci_parse.
Is it possible to create a stored procedure using the oci8 extension?

`ErrorException: oci_execute(): OCI_SUCCESS_WITH_INFO: ORA-24344: success with compilation error in`

The documentation for oci_parse says to create it in advance.

https://www.php.net/manual/en/function.oci-parse.php#refsect1-function.oci-parse-examples

In the documentation of oci_bind_array_by_name, there is a sample to create a stored procedure, but this sample fails.
The error in case of failure is `ORA-24344`, which is the same error I saw this time.

https://www.php.net/manual/en/function.oci-bind-array-by-name.php#refsect1-function.oci-bind-array-by-name-examples

Current commit, the stored procedure is created in the [workflow of github actions](https://github.com/codeigniter4/CodeIgniter4/pull/2487/files#diff-117de92b2c602fd6341909de2b67f7d0a0cb47a8193c7573c1f6752c38492f68R121).
However, it is not a very satisfactory fix.
Do you know how to create a stored procedure using oci8?

kenjis:
Did you see https://www.oracle.com/technetwork/database/database-technologies/php/201212-ug-php-oracle-1884760.pdf ?

*PL/SQL Success With Information Warnings* (P.210 in the PDF) shows code to check for informational errors and warning.

ytetsuro:
@kenjis 

Thank you.
I hadn't seen that.

When I checked it, there was a bug in the Connection class that removed the semicolon.
Fixed it.

https://github.com/codeigniter4/CodeIgniter4/pull/2487/commits/c3d82cb59234f07445c7671a07742d33740743cd#diff-dcb48cf393b982829be79ab49001306c5f06420ded11fe7c462462922ec92191L192-L195

This fix requires the removal of the semicolon for code outside of the OCI8 directory.
This is a fix only for OCI8.
Do you think this fix should be put out as a separate PR?

https://github.com/codeigniter4/CodeIgniter4/pull/2487/commits/c3d82cb59234f07445c7671a07742d33740743cd#diff-a07677ecb7c1a5d29e0d0a8e72f00b5b77b577508602ada4fb8c5c1e0c7c5916L1045-L1052

kenjis:
> Do you think this fix should be put out as a separate PR?

Yes.
I don't get what the problem is at a sight.
Send another PR and explain it, please.

ytetsuro:
@kenjis 

Thanks, response.
I made PR.
Please review.

https://github.com/codeigniter4/CodeIgniter4/pull/5513

ytetsuro:
> It seems all the problems except the [bug on storedProcedure()](https://forum.codeigniter.com/thread-76381-post-390991.html#pid390991) are solved.

Fixed it.
Please review.

I think this solves the problem of storedProcedure.
I couldn't confirm the problem 1 he was talking about, but I could confirm the problem 2, so I fixed it.
I also added a test to show that problem 1 could not be confirmed.

lonnieezell:
@kenjis @paulbalandan  I think we've all reviewed this in the past. Does anyone see anything we need to do in order to get this merged?

kenjis:
@lonnieezell I think this is completed. This feature will be included in v4.2.0.
I'm not sure Ci4 branching, so I hold off on merging.
At least I'm going to merge when `develop` will become a branch for 4.2.0.

javawebmedia:
> @lonnieezell I think this is completed. This feature will be included in v4.2.0. I'm not sure Ci4 branching, so I hold off on merging. At least I'm going to merge when `develop` will become a branch for 4.2.0.

waiting for the new features

kenjis:
@ytetsuro CI 4.2 requires PHP 7.4 or later, and coding style has been changed for 7.4.
Can you rebase and run rector to fix the GitHub Action errors?

ytetsuro:
@kenjis 

Very thanks long time review.
Of cource, i can!

lonnieezell:
@ytetsuro thanks for sticking with us on this one! Once you get those changes taken care of we'll be able to finally merge!

javawebmedia:
Thank you so much. Hopefully it is going to be release soon.

kenjis:
@ytetsuro This is one of the new big features in v4.2.0.
Can you add a short description to the changelog?
https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/changelogs/v4.2.0.rst#enhancements

user_guide_src/source/changelogs/v4.2.0.rst
```diff
@@ -23,6 +23,8 @@ Enhancements
     - Added the functions ``csp_script_nonce()`` and ``csp_style_nonce()`` to get nonce attributes
     - See :ref:`content-security-policy` for details.
 - New View Decorators allow modifying the generated HTML prior to caching. 
+- Added new OCI8 driver for database.
+    - You to access Oracle Database. They support SQL and PL/SQL statements.
```
kenjis, line 27:
```suggestion
    - It can access Oracle Database and supports SQL and PL/SQL statements.
```

ytetsuro:
@kenjis 

Thanks for, feedback!

> Can you rebase and run rector to fix the GitHub Action errors?

Fixed it!💪

> Can you add a short description to the changelog?

Added it!
In addition, Added a document to the query method about the presence of semicolons for OCI8.
Please review.

https://github.com/codeigniter4/CodeIgniter4/pull/2487/commits/4698fc6b1b2a4ab36e2dff3910d060407d51f663

user_guide_src/source/database/queries.rst
```diff
@@ -27,6 +27,9 @@ this::
 
     $query = $db->query('YOUR QUERY HERE');
 
+.. note:: If you are using OCI8 Driver, SQL statements should not end with a semi-colon (`;`).
+    PL/SQL statements should end with a semi-colon (`;`).
```
kenjis, line 31:
```suggestion
.. note:: If you are using OCI8 Driver, SQL statements should not end with a semi-colon (``;``).
    PL/SQL statements should end with a semi-colon (``;``).
```

kenjis (changes requested):
I added an inline comment.
And can you resolve the conflict?

I would like to merge as soon as possible.

ytetsuro:
@kenjis 

Thanks for, review!

I Fixed it!!💪

I think should add storedProcedure method in the document.
But where to write it I do not know.😭

https://forum.codeigniter.com/thread-76381-post-391902.html#pid391902

Another question is that the storedProcedure method seems to have been removed in CI3.🤔
If the reason is reasonable, I would like to remove the storedProcedure method in this PR as well.

https://github.com/bcit-ci/CodeIgniter/issues/3981#issuecomment-1006056197

What do you think?

kenjis:
@ytetsuro Please fix coding style.

kenjis:
I don't know why `stored_procedure()` was removed in v3.
Since v3 and v4 are developed separately, there is no need to make them the same.

I think deleting it will cause problems for users who are already using it.

ytetsuro:
> Please fix coding style.

Fixed following!💪

https://github.com/codeigniter4/CodeIgniter4/pull/5581

system/Database/OCI8/Connection.php
```diff
@@ -0,0 +1,720 @@
...
+     * @return mixed
+     *
+     * params array keys
+     *
+     * KEY      OPTIONAL  NOTES
+     * name     no        the name of the parameter should be in :<param_name> format
+     * value    no        the value of the parameter.  If this is an OUT or IN OUT parameter,
+     *                    this should be a reference to a variable
+     * type     yes       the type of the parameter
+     * length   yes       the max size of the parameter
```
kenjis, line 491: Since this description is about parameters, please modify it to the `@param` format.

system/Database/OCI8/Connection.php
```diff
@@ -0,0 +1,720 @@
...
+    public function storedProcedure(string $procedureName, array $params)
+    {
+        if ($procedureName === '') {
+            throw new DatabaseException(lang('Database.invalidArgument', [$procedureName]));
+        }
+
+        // Build the query string
+        $sql = 'BEGIN ' . $procedureName . '(';
+
+        $haveCursor = false;
```
kenjis, line 502: `$haveCursor` is not used. Is it okay?

ytetsuro:
> I think deleting it will cause problems for users who are already using it.

Thanks for, feedback!!

I sure.

system/Database/OCI8/Connection.php
```diff
@@ -0,0 +1,720 @@
...
+    {
+        return $this->cursorId = oci_new_cursor($this->connID);
+    }
+
+    /**
+     * Executes a stored procedure
+     *
+     * @param string $procedureName procedure name to execute
+     *
+     * @return mixed
```
kenjis, line 482: Could you make it more specific types?

kenjis:
> Random ordering is not currently supported in Oracle and will default to ASC instead.
https://codeigniter4.github.io/CodeIgniter4/database/query_builder.html?highlight=oracle

It this correct?

ytetsuro:
@kenjis 

> I don't know why stored_procedure() was removed in v3.
> Since v3 and v4 are developed separately, there is no need to make them the same.
> I think deleting it will cause problems for users who are already using it.

I am tempted to delete storedProcedure and getCursor after seeing this reply and narfbg's reply.
https://github.com/bcit-ci/CodeIgniter/issues/3981#issuecomment-1013785678

I understand that CodeIgniter3 and CodeIgniter4 are different, but as narfbg said, Database Driver abstracts Database operations, and this is the same for CodeIgniter4.

The return value of db_connect is also BaseConnection, and the user's IDE cannot determine that those methods exist.

I also agree that it would be annoying to users who are already using it.

To that end, here are the solutions I'm currently considering.
Which do you think is better?

1. create a file like Helpers/oci8_helper to realize storedProcedure in a different module from Database abstraction.
1. stop using original prepared statements when using OCI, and use a class that uses oci_bind_by_name to create queries.
1. Or if you have a better solution, please let me know.

system/Database/OCI8/Connection.php
ytetsuro: Thanks, the feedback.

Fixed it!.

system/Database/OCI8/Connection.php
ytetsuro: Thanks, for feedback.

Fixed it!.

system/Database/OCI8/Connection.php
ytetsuro: No problem.

Added test.🤟

ytetsuro:
> It this correct?

Thanks for feedback.

Removed it.

kenjis:
> Database Driver abstracts Database operations, and this is the same for CodeIgniter4.

I'm not sure CI4 aims to Database Driver abstraction.
If we aim to it, we cannot use any Database specific functionality,
and cannot add any Database Driver specific functionality to a specific driver.
We should remove `callFunction()`, because it does not work if a user changes the database.

Personally I think abstraction is Query Builder (and CodeIgniter\Model) and Forge.
And Stored Procedure is not so rare functionality, but not all databases have it.
It might be possible that BaseConnection has `storedProcedure()`.

> 2. stop using original prepared statements when using OCI, and use a class that uses oci_bind_by_name to create queries.

I don't get what you mean. What is *original prepared statements* ?

ytetsuro:
@kenjis 

Thanks for feedback.

I'm learning a lot. Thank you.

> I'm not sure CI4 aims to Database Driver abstraction.
> If we aim to it, we cannot use any Database specific functionality,
> and cannot add any Database Driver specific functionality to a specific driver.

Indeed, I don't think we should add it!

 I thought that inheriting from BaseConnection meant that I was abstracting the Database connection.
I don't think we should do that even as a Liskov substitution principle.
If it is to be a subclass of BaseConnection.

However, the idea that you should use PDO if you want abstraction may also exist...

> We should remove callFunction(), because it does not work if a user changes the database.
> Personally I think abstraction is Query Builder (and CodeIgniter\Model) and Forge.

I don't know why there is a `callFunction` that can only return boolean.
However, as you said, abstraction may not be the goal when there are callFunction and query methods.

Thank you. I learned a lot and understand.

> And Stored Procedure is not so rare functionality, but not all databases have it.
> It might be possible that BaseConnection has storedProcedure().

It's true that storedProcedure is not an ORACLE-specific feature.
MySQL can call storedProcedure in a CALL statement via the query method.
In OCI, I think we need to prepare a separate storedProcedure method because we need to allocate variables of reference and resource types.
I think because the binding process for prepared statements is a proprietary implementation.
This is because references and resource types can only be bound using functions such as oci_bind_by_name.

From the above, I feel that it would be better to implement it in 2.
What do you think?

> I don't get what you mean. What are original prepared statements?

I'm sorry.
My explanation was not good.
I meant to stop the original binding implementation for prepared statement.
Specifically, we are considering whether it would be possible to design the binding process in the Query class so that the PreparedQuery class is used for binding.
By doing so, we believe that users will be able to execute storedProcedure in the query method, just like in MySQL.

kenjis:
```
$ composer cs-fix
$ vendor/bin/rector process file1 file2 ...
```

kenjis:
I would like to release Oracle driver with functionality that can upgrade CI 3.1 code to 4.x as soon as possible.
If we can execute stored procedures by the `query()` method, Does it provide value to the Oracle users?

ytetsuro:
> If we can execute stored procedures by the query() method, Does it provide value to the Oracle users?

No.

I can't offer much value in being able to run storedProcedure in the query method.
However, I think I can provide value in stopping proprietary bindings.
For example, it will be able to store binary data.
Being able to execute storedProcedure from the query method is one such example.

https://www.php.net/manual/en/function.oci-bind-by-name.php#example-1789

kenjis:
@ytetsuro I don't fully understand your 2. plan, but it seems a good way to go.

But this PR is too long living and I really want to finish it.
Refactoring or improvements can be added later.

Can you fix the coding style error?

system/Database/OCI8/PreparedQuery.php
```diff
@@ -0,0 +1,134 @@
...
+     * Replaces the ? placeholders with :1, :2, etc parameters for use
+     * within the prepared query.
+     */
+    public function parameterize(string $sql): string
+    {
+        // Track our current value
+        $count = 0;
+
+        return preg_replace_callback('/\?/', static function ($matches) use (&$count) {
+            return ':' . ($count++);
```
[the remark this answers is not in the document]
kenjis: The PHPDoc was fixed, starting with `:0`.

kenjis (approved):
LGTM

ytetsuro:
@kenjis 

Thank you very much for supporting this PR for a very long time.

> I don't fully understand your 2. plan, but it seems a good way to go.

I understood.
I would like to issue an issue with more details on this content when this PR is merged.

Thank you very much!!

samsonasik (approved):

kenjis:
The one of the oldest, and most commented PR has been merged!

PR #2489, opened 2020-01-14
Fix pager

MashinaMashina:
All models have single pager.
Opportunity to change parameter whith we get current page.
Models now dont work with $_GET['page']


MashinaMashina:
After tests, setPageSelector() dont normal work.

I get after ->setPageSelector('poger') this:
?poger=1&page=3

I dont know how to do this function good.

PR #2490, opened 2020-01-15
fix. Toolbar init view Error

### LINKED ISSUE #2464: Issue passing data to views
I am running v4.0.0-rc.3, and I was generating some excel output from a view.  Because of the toolbar debugging, I kept getting the "<!-- DEBUG-VIEW (yadda yadda)" in the output (which is fine since I saw that the toolbar was generating that via the View collector).  I disabled (commented out), the Views collector in app\config\Toolbar.php, and everything worked great.  However, other stuff started to break.  

I have my app set up so there is a primary layout file in views, with many "partial" views that I use.  As long as the Views collector is enabled, I can reference variables passed to the layout view within the sub-views, but once I disable it, they are no longer available.  I don't really care which way it is, as long as it stays consistent.  I'm afraid when I go to production and have the Toolbar disabled, my app will stop working.

Here is an example:

In a controller, I call:

return view('layouts/primary', ['foo'=>'bar']);

in the primary.php view, I have:

[CODE]
echo $foo; 
echo view('layouts/blocks/navigation');
[/CODE]

Within the navigation view file, I have:
[CODE]
echo $foo; 
[/CODE]

When I execute this with the TOOLBAR VIEWS FILTER ENABLED, I get "barbar", but when I disable the views collector,  I get an error indicating that the variable $foo is not defined in the navigation view.

Instrye:
fix #2464. `Services::renderer` parameter passing error.

### DIFF
--- a/user_guide_src/source/testing/debugging.rst
+++ b/user_guide_src/source/testing/debugging.rst
@@ -75,13 +75,14 @@ can easily make your own to customize the toolbar. To determine which collectors
 the **app/Config/Toolbar.php** configuration file::
 
 	public $collectors = [
-		'CodeIgniter\Debug\Toolbar\Collectors\Timers::class',
-		'CodeIgniter\Debug\Toolbar\Collectors\Database::class',
-		'CodeIgniter\Debug\Toolbar\Collectors\Logs::class',
-		'CodeIgniter\Debug\Toolbar\Collectors\Views::class',
- 		'CodeIgniter\Debug\Toolbar\Collectors\Cache::class',
-		'CodeIgniter\Debug\Toolbar\Collectors\Files::class',
-		'CodeIgniter\Debug\Toolbar\Collectors\Routes::class',
+		'\CodeIgniter\Debug\Toolbar\Collectors\Timers::class',
+		'\CodeIgniter\Debug\Toolbar\Collectors\Database::class',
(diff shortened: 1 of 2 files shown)


user_guide_src/source/testing/debugging.rst
```diff
@@ -72,16 +72,17 @@ Choosing What to Show
...
-		'CodeIgniter\Debug\Toolbar\Collectors\Timers',
-		'CodeIgniter\Debug\Toolbar\Collectors\Database',
-		'CodeIgniter\Debug\Toolbar\Collectors\Logs',
-		'CodeIgniter\Debug\Toolbar\Collectors\Views',
- 		'CodeIgniter\Debug\Toolbar\Collectors\Cache',
-		'CodeIgniter\Debug\Toolbar\Collectors\Files',
-		'CodeIgniter\Debug\Toolbar\Collectors\Routes',
+the **app/Config/Toolbar.php** configuration file::
+
+	public $collectors = [
```
natanfelles: This should not have the `::class` part.

or `'\CodeIgniter\Debug\Toolbar\Collectors\Timers::class'` should be changed to `\CodeIgniter\Debug\Toolbar\Collectors\Timers::class`

Instrye: thanks. you is right.

PR #2497, opened 2020-01-17
Fix splitQueryPart()

MashinaMashina:
Bug:
redirect()->to('/?param') returns url without param

If param has not value, it removes
/?param=1 => /?param=1
/?param => /

In php if you open page with ?param in $_GET array you will see this:
array (
	[param] => '',
)

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -811,11 +811,9 @@ class URI
 	{
 		$parts = explode('=', $part, 2);
 
-		// If there's only a single element, no pair,
-		// then we return null
 		if (count($parts) === 1)
 		{
-			$parts = null;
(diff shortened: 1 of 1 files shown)


MashinaMashina:
@lonnieezell why not just use parse_str() instead splitQueryPart()+decode()+other code in setQuery() ?

lonnieezell:
> why not just use parse_str() instead splitQueryPart()+decode()+other code in setQuery() ?

Honestly - because that function slipped my mind at the time and i was up to my eyes reading  HTTP specs at the time :) 

That's a perfectly good solution if you want to update to that.

MashinaMashina:
@lonnieezell now code via parse_str

PR #2502, opened 2020-01-20
View Renderer - Reset sections after generating the ouput

### LINKED ISSUE #2491: Issue with layouts renderer and sections => merge sections in one view
Hello,

It seems that there is a problem with the layouts and I think that this is due to "ob_start / ob_flush / ..." in the renderer.
It will be easier to explain my problem with an example...

```
public function send(): string
{
    $email = view('email');
   
    // Send email script here...

    return view('confirmation');
} 
```

View "email" extends a layout "x" with a section called "body".
View "confirmation" extends another layout "y" with a section called "body".

Sometimes (but not always), "view('confirmation')" displays the "body" section of "email" and "confirmation". The renderer merges both sections even if the layout is not the same.
This is really annoying. For the moment, I try to prefix all my sections in each layouts but it would be better if I didn't have to do this kind of things.

najdanovicivan:
**Description**
Sections array is left populated after generating output. This creates issues when the same shared View rendered instance is called for the second time.

Fixes #2491

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/View.php
+++ b/system/View/View.php
@@ -247,6 +247,9 @@ class View implements RendererInterface
 			$output       = $this->render($layoutView, $options, $saveData);
 		}
 
+		// Reset sections
+		$this->sections = [];
+
 		$this->logPerformance($this->renderVars['start'], microtime(true), $this->renderVars['view']);
 
 		if ($this->debug && (! isset($options['debug']) || $options['debug'] === true))


MGatner:
This is actually breaking views nested in layouts, causing `renderSection()` not to return its output. I'm going to revert this while we look for a different solution.

PR #2506, opened 2020-01-22
Proposal: HTTP Response - Fix crash on CSP methods CSP is disabled

### LINKED ISSUE #2456: Bug: if not CSPEnaled but i have some
**Describe the bug**
if for test change app.CSPEnabled from true to false
BUT if 
into initController i used $response->CSP->addStyleSrc('https://fonts.googleapis.com');

Return an error "Call to a member function addStyleSrc() on null "

**CodeIgniter 4 version**
RC-3

**Context**
 - OS: Win
 - Web server Apache
 - PHP version 7.3

najdanovicivan:
**Description**
If CSP is disabled property $CSP in HTTP/Response is not initialized.

If we try to access the CSP methods on the request object anywhere in code with CSP disabled it will crash the framework with "Call to a member function …. on null "

In order to avoid this CSP object can be initiated regardless of CSP config.

I’m aware that this is not the most efficient way to bypass the issue but some mechanism for disabling CSP should exist without having to do modifications everywhere in code.

Maybe better idea will be to create mock class to be loaded instead which will respond with catchall  magic methods like __call __set __get ….. But I don’t know if it is worth doing it as it will require adding additional class in framework.

Fixes #2456 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/Response.php
+++ b/system/HTTP/Response.php
@@ -239,13 +239,10 @@ class Response extends Message implements ResponseInterface
 		// Also ensures that a Cache-control header exists.
 		$this->noCache();
 
-		// Are we enforcing a Content Security Policy?
-		if ($config->CSPEnabled === true)
-		{
-			$this->CSP        = new ContentSecurityPolicy(new \Config\ContentSecurityPolicy());
-			$this->CSPEnabled = true;
-		}
+		// We need CSP object even if not enabled to avoid calls to non existing methods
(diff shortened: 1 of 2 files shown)


najdanovicivan:
I've additionally added Unit Test to cover such case.

lonnieezell:
I think this is a fine solution. Using a mock class would still have the performance cost of locating and loading the file so we wouldn't gain anything there. 

Thanks!

PR #2508, opened 2020-01-23
First page extension - PagerRenderer

nowackipawel:
optional parameter in 
`$pager->getFirst(true);` and `$pager->links(true)` makes PAGE part optional for the first page .
1st:  domain.tld/path
any other: domain.tld/path/number or domain.tld/path?page=number

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -508,7 +508,15 @@ class URI
 			throw HTTPException::forURISegmentOutOfRange($number);
 		}
 
-		$this->segments[$number] = $value;
+		if($number + 1 === count($this->segments) && empty($value))
+		{
+		    unset($this->segments[$number]);
+		}
+		else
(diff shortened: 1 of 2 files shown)


MGatner:
I like the concept but I'm uncomfortable with the change to `setSegment()`. Is that necessary to the new functionality of Pager? If not, please resubmit it as a separate PR.

nowackipawel:
If I remember setSegment method is my method introduced when I PR another changes to pager (segment instead of ?page query) in which I forgot to unset empty segments . In case the first page should not have any number at the end there was slash added which is not necessary and looks a little weird .

Do you think it will be better to introduce another method removeSegment instead ?

lonnieezell:
Two biggest problems are 1) no docs for the new feature and 2) no new tests for the feature. Cannot accept without those.

PR #2517, opened 2020-01-27
Fix error_exception: threw exception

benrogmans:
Tiny fix. `$value` of `$row['args']` was triggering an exception:

> Uncaught ErrorException: print_r(): Property access is not allowed yet

Updated to use $params[$key].


benrogmans:
Sorry, this is not a proper solution. Throws errors in other cases.

PR #2526, opened 2020-01-30
Page in the official documentation on ajax requests with iSAJAX() fixes #2454

### LINKED ISSUE #2454: Bug: IsAJAX() relies on inconsistent headers
Newer JavaScript implementations (i.e. **fetch**) don't set the `HTTP_X_REQUESTED_WITH` header, making the current implementation of `IncomingRequest::isAJAX()` unreliable.

So far I have not found a reliable way to distinguish **fetch** requests, and if this is ultimately the case then `isAJAX()` should be removed or reworked so it is clear that it is not definitive.

Ref. https://stackoverflow.com/questions/44202593/detect-a-fetch-request-in-php

jlamim:
**Description**
After the conversations on issue #2454 , I created a page in the documentation, within the "General Topics" section to explain the issue of using `isAJAX()`, giving examples of how to force the `X-Requested-With` header to be sent both with the Fetch API as well as with jQuery, VueJS and React, thus facilitating the understanding and use of the resource by CI4 users.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/concepts/http.rst
+++ b/user_guide_src/source/concepts/http.rst
@@ -97,6 +97,10 @@ is an object-oriented representation of the HTTP request. It provides everything
 The request class does a lot of work in the background for you, that you never need to worry about.
 The ``isAJAX()`` and ``isSecure()`` methods check several different methods to determine the correct answer.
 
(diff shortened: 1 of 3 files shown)


MGatner:
That's a good writeup. Are there places in the existing docs that should link to this page? A couple places at **incoming/incomingrequest.html** look likely.

jlamim:
Yes @MGatner, there are and I will make this link. I didn't do it before because I was waiting for a sign (comment, acceptance of PR) in relation to this page that I created. I'll do that today.

PR #2538, opened 2020-02-10
Parser function just like view function

mostafakhudair:
I'm wondering why CodeIgniter doesn't provide parser function to grab Parser Class Huh , so we can use it directly anywhere without calling it as instance.  Idea

for example:
before "as documentation":
```
/* loading parser class through its service. */
/* or loading it by instantiate it directly. */ 
$parser = new \CodeIgniter\View\Parser();

/* Then we can use it like this. */
$parser->setData($data)->render('index'); 
```

I wrote this function and placed it in app/Common.php, "actually just copy and paste" Big Grin  system/Common.php view function and do some few changes to it.
```
function parser(string $name, array $data = [], array $options = []): string
{
    /**
     * @var CodeIgniter\View\Parser $renderer
     */

    $saveData = null;
    if (array_key_exists('saveData', $options) && $options['saveData'] === true)
    {
        $saveData = (bool) $options['saveData'];
        unset($options['saveData']);
    }

    return $parser->setData($data, 'raw')->render($name, $options, $saveData);
} 
```

after:
```
/* now we can use it in one short line. */
parser('index', $data); 
```

it's works as expected, so can it added officially to CodeIgniter 4?  Exclamation

btw I couldn't pull a request on github it's so difficult there.  Confused


lonnieezell:
I would be willing to accept a clean PR for this. However, what you have currently is coming from your `gh_pages`  branch into our develop branch. Not workable. Take a look at the "files changed" up above and you'll see over 900. I don't think you changed 900+ files for this simple addition. 

Instead, checkout the develop branch. Make sure it's up to date. Create a new branch off of that in your fork. Make your changes. then submit a PR.

PR #2540, opened 2020-02-10
valid_ip removed $data which was causing exception

nowackipawel:


MGatner:
That data parameter is odd and appears that it should be removed, as you noted. Please also remove its reference from the inline docs (@param). 

What is the `inet_ntop`? You should explain additions and ideally add them in a separate PR.

nowackipawel:
@MGatner inet_ntop and pton are converting the IPv4 or v6 string to its binary/human readable form.
I am just thinking about extending Entity as well to store it bydefault like this... and this extending will be preety easy:
in castAs:
``

            case 'ip':
                $value = inet_ntop($value);
                break;
``
in __set
``

            if($castTo === 'ip')
            {
                $value = inet_pton($value);
            }
``
in fact I am using this in my extended entity class to do not use setters every time when I need to stroe IP in that efficient way.

MGatner:
Since the tests are all passing this doesn't seem to have affected the outcomes, while potentially improving the service. It looks good to me, will merge - thanks!

Re: Entity auto-cast, I like the idea. Since casts are only used when a developer requests them I think it is a fine addition. If you want to add it please start a new PR and be sure to update the User Guide and have tests with the new cast.

Issue #2373, opened 2019-10-29
Update package dependency version for kint-php/kint


Martin-4Spaces:
It would be nice if the `composer.json` file could be updated with kint-php/kint 3.2 instead of 
2.1.
2.1 is kind of old (2017) and I am having conflicts with other composer packages depending on kint 3.2 :)
Ty!

MGatner:
Good suggestion. We'll have to test it first but I think this would be good to do before release.

lonnieezell:
The new version of Kint does not make our requirements any easier. lol 

Since we have a vocal segment that doesn't like Composer and just wants to download the zip of the framework we've previously copied the file to ThirdParty and used that file so that it's always available. Well, in 3.x Kint is more mature and complex. The worst offender currently is that it tells composer to load it's init file automatically so getting a working system that works both ways is proving a little tricky. I'll work on it some more tomorrow, hopefully.

musmanikram:
@lonnieezell i don't know what you are meant by vocal segment.
But i tried to update kint version, please have a look on this PR: https://github.com/codeigniter4/CodeIgniter4/pull/2542

najdanovicivan:
@lonnieezell  I've added two possible options for handling this one properly.

Option one is to obtain kint from git repo requiring modification of commit hash for each update.

The second one builds kint locally. It requires changing the ini configuration but it avoids versioning issues with commit hashes 

IMO I like the second one better.

MGatner:
@najdanovicivan that's some nice work on those! Thanks for taking the time.

lonnieezell:
@najdanovicivan what type of "building" with Kint is required? We don't necessarily need the phar file, I don't think. Of the top of my head, that's likely my preferred way to, but need a little more info, please.

lonnieezell:
@najdanovicivan @MGatner I would love it if you guys could pull down PR #2564  and give it a spin. 

I think it works, and makes it possible to use without a phar file, since I really don't like the idea of requiring every dev to modify the php.ini to allow phars to not be read-only. That sounds like a dangerous setting I'd rather not force.

najdanovicivan:
@lonnieezell I gave it some thought over the weekend how to avoid compiling phar. I see you've already came to the similar solution but my idea is bit different. Check out #2565 I've submitted this one to your kint3 branch. 

[...]

PR #2542, opened 2020-02-12
2373 Update kint-php version to 3.3

musmanikram:
Update kint version to 3.3 #2373

**Description**
Update package dependency version for kint-php/kint from 2.1 to 3.3

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         "ext-curl": "*",
         "ext-intl": "*",
         "ext-json": "*",
-        "kint-php/kint": "^2.1",
+        "kint-php/kint": "^3.3",
         "psr/log": "^1.1",
         "laminas/laminas-escaper": "^2.6"
     },
--- a/user_guide_src/source/testing/debugging.rst
+++ b/user_guide_src/source/testing/debugging.rst
@@ -19,10 +19,7 @@ and much, much more.
 Enabling Kint
 =============
 
-By default, Kint is enabled in **development** and **testing** environments only. This can be altered by modifying
-the ``$useKint`` value in the environment configuration section of the main **index.php** file::
-
-    $useKint = true;
+By default, Kint is enabled in **development** and **testing** environments only.
 
 Using Kint
 ==========


lonnieezell:
Unfortunately, there's much more to it than this. While your fix works fine for composer-based installs, it will fail completely for those that want to simply download a release and not use composer. 

For those, the post-install script has to be modified to copy the files into the appropriate place, and then everything has to work smoothly. I had attempted this at one point and ran into issues there, as described in the issue you referenced. Haven't had an opportunity to get back to it just yet.

musmanikram:
> Unfortunately, there's much more to it than this. While your fix works fine for composer-based installs, it will fail completely for those that want to simply download a release and not use composer.
> 
> For those, the post-install script has to be modified to copy the files into the appropriate place, and then everything has to work smoothly. I had attempted this at one point and ran into issues there, as described in the issue you referenced. Haven't had an opportunity to get back to it just yet.

I will try to fix zip issue. But is it very important to give support for zip download? I mean we can enforce to use only composer. Otherwise, we have to support both and will face these issues in future,

lonnieezell:
> But is it very important to give support for zip download?

Yes. That is the CodeIgniter legacy. And, at least with v4, we are keeping the ability to download and go, just like we're keeping the ability to auto-route based on the URL. These are things that helped make CI what it is, and we are trying to make an improved version of an existing framework, not creating a brand new framework, so we do have to keep the feel where possible, and the things many many have loved about it, whether or not we all want to use that style of programming.

musmanikram:
> > But is it very important to give support for zip download?
> 
> Yes. That is the CodeIgniter legacy. And, at least with v4, we are keeping the ability to download and go, just like we're keeping the ability to auto-route based on the URL. These are things that helped make CI what it is, and we are trying to make an improved version of an existing framework, not creating a brand new framework, so we do have to keep the feel where possible, and the things many many have loved about it, whether or not we all want to use that style of programming.

ahh that make sense. I will try to spend some time it.

musmanikram:
I noticed it's a bit tricky now. I am closing this PR and we can see this issue in future

PR #2548, opened 2020-02-14
Add ajax.rst to user guide /general/index.rst

dafriend:
This was missed when ajax.rst documentation was added during merge #2526
Without being added to index.rst the page won't be included in the build.

### DIFF
--- a/user_guide_src/source/general/index.rst
+++ b/user_guide_src/source/general/index.rst
@@ -12,6 +12,7 @@ General Topics
     logging
     errors
     caching
+    ajax
     modules
     managing_apps
     environments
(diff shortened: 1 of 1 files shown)


MGatner:
I see that it is missing, but the PR appears to revert all its own changes. Is it just adding "ajax" to **user_guide_src/source/general/index.rst**, like your first commit?

dafriend:
It is doing only that. The first PR was not signed.

dafriend:
Geez, what a mess. I reverted so I could redo the commit properly signed. Clearly missed the fact that the desired change didn't get included. I seems to be good now.

MGatner:
Looks good now, thanks a lot!

MGatner:
Your original unsigned commit is still there so I cannot merge this. @lonnieezell could override or you could resubmit a new PR. “Revert” undoes the changes of a commit but leaves the actual commit; to delete a commit you would need to “reset” and force push.

PR #2551, opened 2020-02-14
disable buffer check on "testing" environment

samsonasik:
**Description**
To clean up the buffer notice in "testing" environment which previously shows:

```bash
There were 3 risky tests:

1) CodeIgniter\CodeIgniterTest::testRunClosureRoute

2) CodeIgniter\CodeIgniterTest::testRun404OverrideByClosure

3) FeatureTestCaseTest::testEchoes
```

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Config/Events.php
+++ b/app/Config/Events.php
@@ -20,14 +20,17 @@ use CodeIgniter\Events\Events;
  */
 
 Events::on('pre_system', function () {
-	while (\ob_get_level() > 0)
+	if (ENVIRONMENT !== 'testing')
 	{
-		\ob_end_flush();
-	}
+		while (\ob_get_level() > 0)
+		{
+			\ob_end_flush();
+		}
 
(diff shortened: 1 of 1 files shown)


samsonasik:
From 3 down to 1:

```bash
There was 1 risky test:

1) FeatureTestCaseTest::testEchoes
```

MGatner:
Wow thanks for this, such an annoyance. I looked into it at one point but was baffled where the issue was. Unfortunate that it is in **app/** but this should affect TravisCI more than most users.

MGatner:
We’re still getting one of these. @samsonasik any chance you could look at this last one since you seem to have dug into this issue?
```
1) FeatureTestCaseTest::testEchoes
366 Test code or tested code did not (only) close its own output buffers
```

samsonasik:
I will check the last one when I have a chance ;)

PR #2553, opened 2020-02-14
Update to Kint 3.3

najdanovicivan:
**Description**
The only way to satisfy the requirement for CI and Kint is to use compiled kint.phar. Fortunately phar version exists but it's only available on github repo and composer package does not include it. 

There is an issue of not allowing the kint to load multiple time which is fortunately behavior of the Kint itself as it checks and defines the KINT_DIR on load which prevents it from loading multiple times.

The composer.json file has been updated to pull the kint from github repo using the commit hash of latest production version 3.3 which contains the phar file. This is not the best way to do it as kint classes will exist in multiple places and with hardcoded hash it will require to manually update version in composer.json file on each release but it's the only way to do it at the moment.

It will be nice if we can get kint-php/kint to provide phar version via composer repositories so taht in future it can easily be updated

Ref #2373 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


najdanovicivan:
Another way to achieve the same functionality will be to completely get rid of Kint composer package and provide the kint.phar file in the framework. This may solve some issues related to dependencies but this will make updating kint completely manual.

Another possible option will be to include Kint build scripts in CI4 and build Kint from composer script on each update.

PR #2561, opened 2020-02-15
Fix - Add ajax to docs "General " page [ci skip]

dafriend:
Because it wasn't on index the new "ajax" page wasn't included in the build.


MGatner:
Much better - thanks!

PR #2563, opened 2020-02-16
Extending entity [bool support in __set]

nowackipawel:
we are doing with bool the same what with array and json ;-)... converting to store it in database.

This should also stop to brake validation rule required if property is bool and we are setting it using true/false instead of 0/1


lonnieezell:
Closing as appears abandoned.

PR #2564, opened 2020-02-17
Updated loader and composer script to use Kint 3.3 for #2373

lonnieezell:
Updated the system to use Kint 3.3. 

Due to new, more mature architecture of Kint, we have to copy over all files and ensure our file gets used instead of Composers.


MGatner:
I’ll try this out and look at the other two PRs more in depth. Just to be clear the requirements are:
* Kint as a module that gets updated with new releases
* A working solution for both Composer and download installation methods

lonnieezell:
@MGatner Yes, those are the correct use cases. 

I've just merged some nice updates from @najdanovicivan so you'll need to pull down the latest before giving it a spin. Thanks.

PR #2565, opened 2020-02-17
Updates To Kint Loading

najdanovicivan:
**Description**
Here's another way of loading Kint. This makes it use composer version if it's available and only loads Kint from ThirdParty folder if there is no composer version already loaded. This makes it possible to properly update Kint to latest version from composer when when ComposerScripts.php is not being executed fo example when composer's been added after getting zipped framework files and script is left out.

Additionally I've added some configurations for Kint to be obtained from Config\App. We might  add additional Kint configurations there as well.

One important config to be set is 
```
\Kint\Renderer\RichRenderer::$folder = false;
```

This make Kint to be displayed in place where it's called. Otherwise it goes to bottom as toolbar which creates conflicts with debug toolbar.

This also handles the situation where there are Kint calls left out in production environment and Kint is not loaded via composer by always loading Kint classes and setting Kint as disabled. Otherwise the scripts will crash if there are Kint calls and environment is set to anything but production. With this it only skips Kint calls and the app does not crash anymore.

Ref #2373 and #2564 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -266,4 +266,15 @@ class App extends BaseConfig
 	*/
 	public $CSPEnabled = false;
 
+	/*
+	|--------------------------------------------------------------------------
+	| Kint
+	|--------------------------------------------------------------------------
+	|
+	| Kint Related configuration
+	|
+	*/
+	public $kintRendererTheme  = 'aante-light.css';
+	public $kintRendererFolder = false;
+
 }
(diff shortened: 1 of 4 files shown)


lonnieezell:
> \Kint\Renderer\RichRenderer::$folder = false;

I was searching for this unsuccessfully last night. Thanks. 

I think your solution works well. I initially balked at not using our autoloader, but pushing it to its own makes sense as it bypasses the ugly hack I was doing by removing it from our autoloader.

And I do like the idea of config options for Kint. Nice touch.

Merging. Still want a couple of people to bang on the kint3 branch under different scenarios before I merge that into develop just yet, though.

PR #2569, opened 2020-02-17
Use lowercase service names

### LINKED ISSUE #2534: Bug: system/Config/BaseService.php getSharedInstance() will not return mocks with non-lowercase key
**Direction**

**Describe the bug**
system/Config/BaseService.php injectMock() method converts argument $name to lowercase,
but getSharedInstance() does not, which leads to BaseService not returning mocks with name containing uppercase letter.

**CodeIgniter 4 version**
4.0.0-rc.3,
Develop [checkout @ 2020.02.09 09:05]

**Affected module(s)**
CodeIgniter\Config\BaseService

Steps to reproduce:

1. create a pair of classes, e.g MyClass and MockMyClass
2. add method to Config\Services, e.g.:

        public static function myClass(bool $getShared = true) {
            if ($getShared){ 
                return static::getSharedInstance('myClass')
            }
            return new MyClass();
        }

3. Create a test case that injects MockMyClass

       class TestClass extends CIUnitTestCase {
            $mock = MockMyClass();
            Services::injectMock('myClass', $mock);
            $testObject = Services::myClass();
            $this-assertEquals(spl_object_id($mock), spl_object_id($testObject));
       }

Expected behaviour: Services should return instance of MockMyClass
Actual behaviour: Services returns new (shared) instance of MyClass

MGatner:
**Description**
Shared and mocked instance names are all stored in lowercase, but not necessarily looked up by the same. This PR forces the lowercase version to be used during `getSharedInstance()` as well.

Fixes #2534 

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -66,6 +66,7 @@ class BaseService
 	/**
 	 * Cache for instance of any services that
 	 * have been requested as a "shared" instance.
+	 * Keys should be lowercase service names.
 	 *
 	 * @var array
 	 */
@@ -106,6 +107,8 @@ class BaseService
 	 */
 	protected static function getSharedInstance(string $key, ...$params)
 	{
+		$key = strtolower($key);
+
 		// Returns mock if exists
 		if (isset(static::$mocks[$key]))
 		{


MGatner:
Just a note... This causes static function calls to use the lowercase version as well. Because PHP is case insensitive this works fine, but could cause problems if that ever were to change.

PR #2570, opened 2020-02-17
Model's set method should accept not only string

nowackipawel:
I think someone adds casting here by mistake

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -498,12 +498,12 @@ class Model
 	 * builder methods and still get validated data, like replace.
 	 *
 	 * @param mixed        $key
-	 * @param string       $value
+	 * @param string|int       $value
 	 * @param boolean|null $escape
 	 *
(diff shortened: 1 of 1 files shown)


MGatner:
`Model::set()` should share the same signature with `Builder::set()`, which also only takes string values for input. The handling of the value type happens in a few different places, but ultimately the value is concatenated so safer to remain a string. Actually though, it should be of type `?string` since you can pass an array of values for `$key` and leave `$value` out altogether.

nowackipawel:
@MGatner  ;-) thank you for pointing this out. updated. and yes you are right about the rest.

MGatner:
You still have `string|int` for the docs. I don't know if inline docs care about things like `"4"` but this will always need to take a `string` for input.

MGatner:
Looks correct, thank you!

PR #2573, opened 2020-02-18
Image verification

### LINKED ISSUE #2421: Bug: ImageHandler has no reality checks
**Describe the bug**
ImageHandler does not do any verification that it has a valid image loaded, which can lead to some errors with unhelpful descriptions. E.g. `fit()` will try to calculate the aspect ration with a null value, so `$xRatio = $width / $origWidth;` results in "Division by zero".

I don't know where validation should happen - `withFile()` is maybe the best bet but doesn't catch instances where developers forget to use it before calling other methods.

**CodeIgniter 4 version**
develop

**Affected module(s)**
Image Manipulation Class

MGatner:
**Description**
Currently the framework fails with unhelpful messages if any of the Image handler functions are used without first calling `withFile()` on a valid image file path. This PR adds some checks that `withFile()` has been invoked and its source is an actual image.

Fixes #2421.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Images/Exceptions/ImageException.php
+++ b/system/Images/Exceptions/ImageException.php
@@ -5,6 +5,11 @@ use CodeIgniter\Exceptions\FrameworkException;
 
 class ImageException extends FrameworkException implements ExceptionInterface
 {
+	public static function forMissingImage()
+	{
+		return new static(lang('Images.sourceImageRequired'));
+	}
+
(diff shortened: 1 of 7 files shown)


MGatner:
@lonnieezell I think this one is set, give it a quick glance?

Issue #2445, opened 2019-12-09
trailing slash cause redirect to root


hoceyn:
when I using this url :
`<?= base_url('dashboard/admin'); ?>`
result will this :
http://localhost:8888/mysite.com/public_html/dashboard/admin
it work properly.

but when I add trailing slash :
`<?= base_url('dashboard/admin/'); ?>`
it redirect to :
http://localhost:8888/dashboard/admin

- I didnt change .htaccess
- Im using mamp pro in mac
- apache server

**and this example work in CI3 .**

lonnieezell:
Are you using the latest develop branch, or rc-3? We've fixed some URL issues since rc-3 that sounds like it might be related to this.

StudPro:
I have the same issue. I used rc-3 and develop

Config
Apache on windows
htaccess: default from rc-3
env: 
app.forceGlobalSecureRequests = true
app.baseURL = 'https://localhost/ci4/'
Route: $route->add('landingpage','landingpage');

Controller
Into controller demo page do
redirect('landingpage) 
or redirect()->to('/landingpage');
or redirect()->to(route_to('landingpage'));

go to https://localhost/landingpage and not in https://localhost/ci4/landingpage

What am I doing wrong?

tangix:
Like you I have the same problem. To workaround my problem I have patched the return from fillRouteParams() in system/Router/RouteCollection.php at line 1320 to:
`return ltrim($from, '/');`
In the original code in the develop branch a slash is added:
`return '/' . ltrim($from, '/');`
I have tried to whip up a PR but I can't get the tests to properly work with redirect().

MGatner:
@hoceyn @StudPro Can you try again since Lonnie pushed some changes? https://github.com/codeigniter4/CodeIgniter4/issues/2505#issuecomment-578588461

MGatner:
I haven't been able to replicate this, but I did find that preceding slashes get interpreted as relative to the domain root, rather than the subfolder. E.g.:
```
app.baseURL = 'https://localhost/ci4/'
echo base_url('/foo/bar');
```

The above will return "https://localhost/foo/bar" instead of "https://localhost/ci4/foo/bar". This seems to be an issue specific to `base_url()` since `HTTP\URI` should correctly return the former. I have a fix coming for that - maybe it will help these other issues?

DeadOnTime:
For me this fixed the problem: https://stackoverflow.com/a/27264788

[...]

PR #2574, opened 2020-02-18
Subfolder base_url() with parameter

MGatner:
**Description**
Currently when providing a URI parameter to `base_url()` if it has a preceding slash it will cause the resulting URL to be relative to the root domain, rather than a subfolder defined in `app.baseUrl`. This PR trims preceding and trailing slashes in the helper to ensure the result from `HTTP\URI` is still relative to the base URL.

May fix some of #2445, hard to say without being able to replicate.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -106,6 +106,7 @@ if (! function_exists('base_url'))
 		{
 			$uri = implode('/', $uri);
 		}
+		$uri = trim($uri, '/');
 
 		// We should be using the configured baseURL that the user set;
 		// otherwise get rid of the path, because we have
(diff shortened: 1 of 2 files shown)


MGatner:
@lonnieezell Need your ruling on this. IMO the test has a logic fallacy, because `base_url()` should _always_ be relative to, well, the base URL (which might include a subfolder) whereas `uri_string()`
 should _always_ be absolute to the domain root. If that's the case let me know and I'll remove that portion of the test; otherwise I think we have some other changes to make...

lonnieezell:
I would agree with your logic here. I think it might cause unexpected results for some people that develop in a subfolder then push it to a site without, or vice-versa, but I don't think there's any good way to make that work for them.

MGatner:
Sounds good. I’ll remove those test scenarios and try it again.

PR #2585, opened 2020-02-21
Update - triming / from the path 

nowackipawel:
I think someone remove it by mistake and because of that I when I was d($this->segment) in setSegment() methd i've got "" on the first place of the array.

Do not know who and when but removing trim here make segment pagination not work anymore.

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -705,7 +705,7 @@ class URI
 	{
 		$this->path = $this->filterPath($path);
 
-		$this->segments = explode('/', $this->path);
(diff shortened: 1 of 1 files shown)


MGatner:
All tests pass so I assume this is still kosher with the recent URL changes. I'd still like @lonnieezell to take a look though.

lonnieezell:
Looking at blame for this: https://github.com/codeigniter4/CodeIgniter4/blame/develop/system/HTTP/URI.php that hasn't been changed in the last 4 years. I think it's likely something else you're running into.

Issue #2554, opened 2020-02-14
[Docs] Loading Environment into Configuration documentation described wrong


JanZelenka:
**Describe the bug**
The documentation at [Docs » General Topics » Working With Configuration Files](https://codeigniter4.github.io/userguide/general/configuration.html) states the following:
_If the prefix of a namespaced variable matches the configuration class name exactly, case-sensitive, then the trailing part of the variable name (after the dot) is treated as a configuration property name._ 

This doesn't seem to work. What does work, though, is specifying the prefix in all lowercase, regardless of the class name capitalization. And since the examples in the **env** file are actually specified this way, I believe this is a documentation issue rather than any other.

**CodeIgniter 4 version**
4.0.0-rc.3

**Affected module(s)**
Config documentation

MGatner:
It looks like it checks both the class name (which maybe requires namespacing?) and the "shortPrefix" which is indeed lowercased: https://github.com/codeigniter4/CodeIgniter4/blob/8d92dd0ef121ccacf3ea9811fee9e7676a25bb0f/system/Config/BaseConfig.php#L160

IMO you are right and this is a docs issue, the User guide should be updated to reflect the current behavior.

dafriend:
The help file also shows `.env` examples with PascalCase prefixes.
Speaking of the _.env_ file, the docs could use a little section on where it is, how to use the example that's included, more clarity no how config classes and .env entries relate.

Shall I take a whack at all that?

MGatner:
That would be tremendous! It is a frequent confusion on the forums so I’ve been meaning to see what we actually have in the UG and how we can beef it up. Anything you’d like to send over would be swell.

dafriend:
I'm on it.

dafriend:
While working through the code of ``BaseConfig::getEnvValue()`` I figured out how it actually works but would like some feedback (probably from @lonnieezell)  on if what I learned is the intended behavior.

With regard to [Incorporating Environment Variables Into a Configuration ](https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#incorporating-environment-variables-into-a-configuration), the namespace prefix must be the whole namespace (case-sensitive) of the configuration class. For example, to set the ``SMTPPort`` value for the Email configuration from `.env` you have to use the following.

    Config\Email.SMTPPort = 587

And we know that the "short-prefix" form works too.

    email.SMTPPort = 587

[...]

PR #2612, opened 2020-02-24
[ci skip] Revamp Configuration docs page.

dafriend:
Revised configuration.rst as discussed in issue #2554

### DIFF
--- a/user_guide_src/source/general/configuration.rst
+++ b/user_guide_src/source/general/configuration.rst
@@ -181,7 +181,7 @@ the full (case-sensitive) namespace of the class.
 
 The same holds for a *short prefix*, which is a namespace using only the lowercase version of 
 the configuration class name. If the short prefix matches the class name, 
-the value will from **.env** replaces the configuration file value.
+the value from **.env** replaces the configuration file value.
 ::
 
     app.CSRFProtection  = true    
--- a/user_guide_src/source/general/configuration.rst
+++ b/user_guide_src/source/general/configuration.rst
@@ -1,125 +1,151 @@
-################################
-Working With Configuration Files
-################################
+#############
+Configuration
+#############
 
-Every application needs a way to define various settings that affect the application.
-These are handled through configuration files. Configuration files simply
-hold a class that contains its settings as public properties. Unlike in many other frameworks,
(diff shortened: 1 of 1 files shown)


user_guide_src/source/general/configuration.rst
```diff
@@ -1,125 +1,151 @@
...
 	// Get shared instance with config function
-	$config = config( 'Pager' );
+	$config = config('Pager');
 
 	// Access config class with namespace
 	$config = config( 'Config\\Pager' );
 
-	// Access settings as class properties
+- By using ``\Config\Services::class`` which will apply an instance of the class's config for you::
+
```
lonnieezell: Services are different than config files. Please remove the sections about services here.

user_guide_src/source/general/configuration.rst
```diff
@@ -132,35 +158,53 @@ prefix, followed by a dot (.), and then the variable name itself::
...
+    Config\App.CSPEnabled = true
+
+
+.. note:: Both the namespace prefix and the property name are case-sensitive. They must exactly match the full namespace and property names as defined in the configuration class file.
 
-When you instantiate a configuration file, any namespaced environment variables
-are considered for merging into the configuration objects' properties.
+The same holds for a *short prefix*, which is a namespace using only the lowercase version of 
+the configuration class name. If the short prefix matches the class name, 
+the value will from **.env** replaces the configuration file value.
```
lonnieezell: I'm guessing `will` should be removed from this sentence?

dafriend: Done.

user_guide_src/source/general/configuration.rst
dafriend: Yes, but when you use Services the associated config is (always?) instantiated and applied for you. That was the point I was trying to make. 
Not germane?

MGatner: I think it is convenient but I had to read it a few times to realize what it meant. I think it’s better left out of this part but could maybe have a life as a bonus “note” in Services?

MGatner:
Very nice pass at this whole section, thanks for the work!

dafriend:
Associated issue #2554 was addressed by this merge and should be closed

PR #2615, opened 2020-02-25
Simplified the condition as both clauses returning same output

atishamte:
Simplified the condition as both clauses returning same output

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/app/Config/Mimes.php
+++ b/app/Config/Mimes.php
@@ -516,11 +516,7 @@ class Mimes
 
 		foreach (static::$mimes as $ext => $types)
 		{
-			if (is_string($types) && $types === $type)
-			{
-				return $ext;
-			}
-			else if (is_array($types) && in_array($type, $types))
+			if ((is_string($types) && $types === $type) || (is_array($types) && in_array($type, $types)))
 			{
 				return $ext;
 			}


app/Config/Mimes.php
```diff
@@ -516,11 +516,7 @@ public static function guessExtensionFromType(string $type, ?string $proposed_ex
 
 		foreach (static::$mimes as $ext => $types)
 		{
-			if (is_string($types) && $types === $type)
-			{
-				return $ext;
-			}
-			else if (is_array($types) && in_array($type, $types))
+			if ((is_string($types) && $types === $type) || (is_array($types) && in_array($type, $types)))
```
rahul05ranjan: Please format code. Code looks ugly!

atishamte: Ugly? The code is optimisation of different if conditions. It's not meant to showcase the multiline beautiful reordering. Optimised code is real beauty

MGatner:
Looks better. @lonnieezell This is essentially a formatting change to **app** so should be noted on release but is non-essential for patching.

Issue #2617, opened 2020-02-25
Bug: ResourceController json response always empty


benrogmans:
When creating a ResourceController, [following the documentation](https://codeigniter4.github.io/userguide/incoming/restful.html#resourcecontroller) any json response will be empty. This is because json_encode() can not handle the \CodeIgniter\Entity class automatically.

**Proposed solution**
\CodeIgniter\Entity to implement \JsonSerializable and the method `jsonSerialize()` to return its attributes.

**CodeIgniter 4 version**
4.0.2

**Affected module(s)**
system/Entity.php

**Expected behavior, and steps to reproduce if appropriate**
Following the user guide I would expect to get a json result instead of an empty response.

**Context**
 - macOS
 - php spark serve
 - PHP version 7.2

PR #2619, opened 2020-02-25
Implement JsonSerializable on Entity class

benrogmans:
Allow json_encode() to be called on an Entity class, to have its attributes returned, json_encoded. This is not just a nice to have but also solves an issue for the ResourceController (#2617).

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated (N/A)
- [x] Conforms to style guide

PS. I created a PR for this contribution before but it was messy so I closed it and did it again. First contribution, hope it meets your standards. Congrats on version 4.0!

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -46,7 +46,7 @@ use CodeIgniter\Exceptions\CastException;
 /**
  * Entity encapsulation, for use with CodeIgniter\Model
  */
-class Entity
+class Entity implements \JsonSerializable
 {
 	/**
 	 * Maps names used in sets and gets against unique
@@ -613,5 +613,16 @@ class Entity
 			}
 		}
 		return $tmp;
+	}
+
+	/**
+	 * Support for json_encode()
+	 *
+	 * @return array|mixed
+	 * @throws \Exception
+	 */
+	public function jsonSerialize()
+	{
+		return $this->toArray();
 	}
 }
--- a/tests/system/EntityTest.php
+++ b/tests/system/EntityTest.php
@@ -685,6 +685,13 @@ class EntityTest extends \CodeIgniter\Test\CIUnitTestCase
 		$this->assertTrue(isset($entity->FakeBar));
 	}
 
+	public function testJsonSerializableEntity()
+	{
+		$entity = $this->getEntity();
+		$entity->setBar('foo');
+		$this->assertEquals(json_encode($entity->toArray()), json_encode($entity));
+	}
+
 	protected function getEntity()
 	{
 		return new class extends Entity


benrogmans:
I did amend the commit with my signature and I pushed it to my branch; still merging is blocked. Anyone can help?

`$ git log --show-signature`
```
commit 71eb4ccf014308310e3fca6967c95df11026766a (HEAD -> entity-jsonserializable, origin/entity-jsonserializable)
gpg: Signature made Tue Feb 25 20:53:19 2020 CET
gpg:                using RSA key A8994E83D326366D3D90A6A378198037B77190A5
gpg: Good signature from "Benjamin Rogmans <ben@rogmansmedia.nl>" [ultimate]
Author: Ben Rogmans <ben@rogmansmedia.nl>
Date:   Tue Feb 25 20:45:34 2020 +0100

    Entity to implement JsonSerializable
```

iRedds:
It's a simple way.
But in this case, you send all the data. Raw data (no datamap, no casts). 
With password hash, financial or personal data, etc.

benrogmans:
Yes that’s right... the method could be overridden of course to return only certain attributes but the developer would need to think of that. Any ideas to do this in a better way? 🤔

iRedds:
> Yes that’s right... the method could be overridden of course to return only certain attributes but the developer would need to think of that. Any ideas to do this in a better way?

In the project that I am currently supporting, I have added a proxy class that implements the JsonSerializable interface. The constructor take the entity. The class property contains fields for publication. The Сlass::jsonSerialize() method in a loop gets fields from an entity.
[AbstractProxy.zip](https://github.com/codeigniter4/CodeIgniter4/files/4255957/AbstractProxy.zip)

MGatner:
I like this change but I agree with iRedd that returning raw data values seems like it could be misleading. We already have `toArray()` which handles casts and filters - maybe fetch the data from there?

benrogmans:
Yes I agree as well. Calling __get() like in iRedds AbstractProxy does seem to make sense.

But for the issue I referenced #2617, the ResourceController is not calling toArray(). Maybe that should change then?

MGatner:
I still think your implementation is good, and will satisfy the ResourceController - I just want to see the data coming from `toArray()` instead of `$this->attributes`. Especially for an API or other user-facing controller we want to be showing the cast & filtered version of an entity, not its wrong attributes.

benrogmans:
Of course 🤦‍♂ Thanks for that.

MGatner:
Thanks for that update! I think the test failure is a fluke - I’ve seen it once or twice before. It does beg the question of tests though. Could you add something basic to demonstrate expected behavior of using this new feature? Probably showing that calling JSON functions on it automatically use the new method and return expected attributes & casts.

MGatner:
@benrogmans Any update on getting tests added?

benrogmans:
Yes, will provide those next week

MGatner:
@benrogmans Any word on tests?

benrogmans:
Sorry for the delay. Test pushed. Using assertEquals to make sure that the json string content actually represents the attributes. I hope this is what you were looking for?

benrogmans:
I'm sorry about the messy history. I had trouble signing an older merge commit. My contribution including its test are now in a single signed commit.

lonnieezell:
No problem. Nice addition. Thanks.

PR #2624, opened 2020-02-26
Colored table in CLI

enix-app:
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner:
Good change, looks like the static command is suppressing a newline character perhaps? Tests need to be updated for the slightly different output since it is an exact match. Alternatively you could run the result through `trim()` or always append a newline to $table manually.

lonnieezell:
@enix-app We like this change. Are you able to address the failing tests?

PR #2625, opened 2020-02-26
Fixing buffer FeaturetestCaseTest::testEchoes()

samsonasik:
**Description**

Fixing buffer in test case:

```
There was 1 risky test:

1) FeatureTestCaseTest::testEchoes
```

**Checklist:**
- [x] Securely signed commits


samsonasik:
@MGatner current solution is the `ob_start()` needs to be placed in test itself because already echoed when call : `$response = $this->get('home')` on echoed controller action.

PR #2626, opened 2020-02-26
using pcov instead of xdebug

samsonasik:
**Description**
To make faster build.

**Checklist:**
- [x] Securely signed commits


samsonasik:
just realized xdebug is required in CIUnitTestCase

PR #2640, opened 2020-02-28
[Docs] Fix Composer Installation Page - existing Project

t12ung:
**Description**
- Fix unpaired backquotes and missing '4' in vendor path
- Also, added directory separator at front of vendor path, since
CI3 needs it = \_\_DIR__ . '/../../system'


t12ung:
I've just discovered it won't work having just tried it. It should be this:

`__DIR__ . '/../../vendor/codeigniter4/framework/system'`

I'll close and re-open a new PR after making the changes.

PR #2644, opened 2020-02-29
Create pull_request.yml

lonnieezell:
Trying to get docs auto-building on PRs.

### DIFF
--- a/pull_request.yml
+++ b/pull_request.yml
@@ -1,0 +1,13 @@
+name: "Pull Request Docs Check"
+
+on: 
+- pull_request
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: ammaraskar/sphinx-action@master
+      with:
+        docs-folder: "user_guide_src/"


pull_request.yml
```diff
@@ -0,0 +1,13 @@
+name: "Pull Request Docs Check"
+
+on: 
+- pull_request
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
```
MGatner: Use `actions/checkout@v2` now.

pull_request.yml
```diff
@@ -0,0 +1,13 @@
...
+    - uses: ammaraskar/sphinx-action@master
```
MGatner: I haven’t looked more but when you showed me this I thought it was a build script. Is there a syntax check we could run for PRs, and then run the build action on merge?

jreklund:
Looks like a great idea, but with the current user guide we won't get any syntax highlight. As it depending upon `cilexer` styling. Either we figure out how to include and build it**. Or we update to the latest Sphinx and `source/conf.py` settings to default into PHP instead. Future proofing the project. As it refers to `easy_install` that's not being used anymore.

** I tried building it today on my server to check that my latest user guide PR looked good. But it never found `ci` styling. No idea what went wrong.

MGatner:
@jreklund this is next on my list to look at after #2668. However I have zero Sphinx experience so I’m glad to see you do. I’ll reach out on Slack but feel free to share any resources here too.

jreklund:
I have made PR #2671, removing internal package (cilexer) as a requirement.

For this we will need to run `pre-build-command: apt install -y python-pip` in case it's not already included in that action.

At first glance they use pip (Python 2) and not pip3 (Python 3), so we should be all good there. As the Sphinx version we require only works with Python 2.

paulbalandan:
I think this has already been addressed by the `userguide_ci.yml` workflow?

lonnieezell:
Oh, wow. Forgot about this one. But, yes, this is already handled. Thanks!

Issue #2633, opened 2020-02-27
Bug: forceHTTPS method ignores baseURL configuration when redirecting


jlamim:
**Describe the bug**
When doing a redirect using `$this->forceHTTPS()`, the URL to which the redirection is made does not match the URL of the application.

**URL accessed:** `http://codigos-livroci4.lsd/blog_ci4/public/blog/https`

**Redirect URL:** `https://codigos-livroci4.lsd/blog/https`

**You should redirect to:** `https://codigos-livroci4.lsd/blog_ci4/public/blog/https`

**baseURL configured:** `http://codigos-livroci4.lsd/blog_ci4/public`

**CodeIgniter 4 version**
4.0.2

**Affected module(s)**
Controller

**Expected behavior, and steps to reproduce if appropriate**
You should redirect to `https://codigos-livroci4.lsd/blog_ci4/public/blog/https`

**Context**
 - OS: Windows 10
 - Apache 2.4.41
 - PHP 7.4.0

michalsn:
Right now URI class has no knowledge about a folder in which the application works. I would expect to have this information available via `getPath()`, but we actually override this value because we rely on it heavily to rewrite segments and load controller from the appropriate namespace.

The quick fix would be to use `baseURL` as a hostname but I'm not sure about it... sounds really wrong since it would return an unexpected result for `getHost()` method when running application in a subfolder.

We could also have something like `$uri->getFolder()`? Idk... looking for other ideas.

MGatner:
The `url_helper` is loaded universally and has some functions for helping identify these segments just for this reason. Totally okay to call functions from `url_helper` in core classes.

michalsn:
I'm not sure how `url_helper` can be helpful here. Can you elaborate?
I can also be missing something really simple/obvious here.

MGatner:
Well I haven't actually looked into the code behind this issue at all, so it might not be relevant, but it sounds like you need either a modified version of `current_url()` or some parameter-ized version of `base_url()`, right?

michalsn:
Sadly no. We're working with the URI class here. It's a whole different problem: https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Common.php#L392

MGatner:
It does use `URI` but there's really not a compelling reason for that. Ultimately all it does is construct a string to pass to `Response::redirect()`: https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Common.php#L419

[...]

PR #2647, opened 2020-02-29
fix forceHTTPS bug #2633

michalsn:
**Description**
When CI was installed in a subfolder, the redirection to https did not include this folder.
See: #2633

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -413,12 +413,16 @@ if (! function_exists('force_https'))
 				->regenerate();
 		}
 
-		$uri = $request->uri;
-		$uri->setScheme('https');
+		$baseURL = config(App::class)->baseURL;
+
+		if (strpos($baseURL, 'http://') === 0)
+		{
+			$baseURL = (string) substr(rtrim($baseURL, '/'), strlen('http://'));
+		}
 
 		$uri = URI::createURIString(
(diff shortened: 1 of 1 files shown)


michalsn:
Tests are failing just because I didn't have much luck with some time object 😅

MGatner:
Looks like the test failure was a fluke timing incident. Maybe worth looking into as a separate bug. I think this is ready but I’d like to take a closer look when not on mobile - thanks for taking it on!

MGatner:
Above concern aside I think this is a fine solution. I probably would have done `preg_replace()` on `current_url()` but I don’t know that either solution does better or faster work. Let me know your thoughts on the `rtrim` then we can merge.

michalsn:
My thoughts on `rtrim` - deliver a clear hostname - without a forward slash at the end. But after your comment, I actually checked one more time how the code looks like in `URI::createURIString` and if there is no slash at the end we add it anyway... so `rtrim` has no sense here after all. Just pushed a change.

The thing with using `current_url` is that besides changing the protocol to `https`, we also have to change `baseURL` (in case app is running in a folder), so... string manipulation would be messy. We have a nice class that can handle this for us, so why not use it - that was my thinking when choosing this approach.

lonnieezell:
Looks good to me. Thanks! 

Merging.

PR #2648, opened 2020-02-29
Remove call to resolveRelativeURI in detectURI method

michalsn:
**Description**
I noticed this when looking at `IncomingRequest` class. It seems like this method call does nothing here and can be safely removed.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


lonnieezell:
Hm. Without digging into it I believe that line should handle adding the path portion of the baseURI (for subfolders) onto the main URL. 

Could you double-check on that? I see tests passing but this one makes me nervous.

michalsn:
Yes, I'm quite confident about this one. Inside the `resolveRelativeURI` method we don't change any value for the main request. The final result of this method is returned and since we do nothing with it - this call is not necessary.

Double checked in the project with subfolders.

Issue #2645, opened 2020-02-29
Bug: spark no longer lists function when used by without any parameters


lonnieezell:
in 4.0.2 when issuing the command: `php spark` it should run the `list` command. It currently shows the help instead.

MGatner:
I feel pretty confident this is collateral from https://github.com/codeigniter4/CodeIgniter4/pull/2571

lonnieezell:
@MGatner that was my guess as well. Just had to log it so we could get it fixed since I saw it in the forums.

PR #2650, opened 2020-02-29
Fix default CLI command #2645

michalsn:
**Description**
Update default CLI command to run `list` instead of `help`.
See: #2645

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/CLI/CommandRunner.php
+++ b/system/CLI/CommandRunner.php
@@ -104,7 +104,7 @@ class CommandRunner extends Controller
 
 		if (is_null($command))
 		{
-			$command = 'help';
+			$command = 'list';
 		}
 
 		return $this->runCommand($command, $params);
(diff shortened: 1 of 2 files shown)


MGatner:
Hmm that test actually suggests the default command should indeed have been `help`. @lonnieezell?

michalsn:
Documentation states it should be a list of available commands: https://codeigniter4.github.io/userguide/cli/cli_commands.html?highlight=cli#running-commands

> When called without specifying a command, a simple help page is displayed that also provides a list of available commands

It's a bit confusing at first because we use the term "help page" but in the end, we should get a list of available commands. I guess this is what we really want.

lonnieezell:
Yes - looking for a list of commands there. Sorry for any confusion.

PR #2652, opened 2020-02-29
Updateing logging level restriction for logging.rst

FlyingWings:
If the level is 5 or less, debug message will also be ignored, thus there should be warning(level 5)，any levels higher than 5 would be ignored.


MGatner:
Good catch! I’ll have to wait for the pipeline before I can merge. In the future if you’re just making a change to docs you can prefix you commit message with “[ci skip]” to prevent that from running.

FlyingWings:
> Good catch! I’ll have to wait for the pipeline before I can merge. In the future if you’re just making a change to docs you can prefix you commit message with “[ci skip]” to prevent that from running.

Thanks for replying! As I've undergoing translation docs to simplified chinese, it's also helpful for me to learn from codes and docs. And much thanks for your effort : )

Issue #2586, opened 2020-02-22
Bug: CI cant display error with API Response Trait


MashinaMashina:
If we uses https://codeigniter4.github.io/userguide/outgoing/api_responses.html and we get error, in example from db, CI cant display this error and die with this:
`Fatal error</b>:  Uncaught CodeIgniter\Format\Exceptions\FormatException: Failed to parse json string, error: &quot;Type is not supported&quot;. in /var/www/system/Format/Exceptions/FormatException.php:9`

It because error contains error bug trace with a lot of elements, and json_encode cant handle it.

https://prnt.sc/r62612
https://prnt.sc/r626jq

https://prnt.sc/r626qf
https://prnt.sc/r627gl

MGatner:
Probably related to #2434

Issue #2434, opened 2019-12-04
Bug: JSONFormatter->format() cannot handle errordata, only outputs it's own error


hcbd:
**Describe the bug**
CI only outputs the error thrown in JSONFormatter->format() function, not the one that preceded it.
CI Outputs:
```
<br />
<font size='1'><table class='xdebug-error xe-uncaught-exception' dir='ltr' border='1' cellspacing='0' cellpadding='1'>
<tr><th align='left' bgcolor='#f57900' colspan="5"><span style='background-color: #cc0000; color: #fce94f; font-size: x-large;'>( ! )</span> Fatal error: Uncaught CodeIgniter\Format\Exceptions\FormatException: Failed to parse json string, error: "{0}". in /source/system/Format/Exceptions/FormatException.php on line <i>9</i></th></tr>
<tr><th align='left' bgcolor='#f57900' colspan="5"><span style='background-color: #cc0000; color: #fce94f; font-size: x-large;'>( ! )</span> CodeIgniter\Format\Exceptions\FormatException: Failed to parse json string, error: "{0}". in /source/system/Format/Exceptions/FormatException.php on line <i>9</i></th></tr>
<tr><th align='left' bgcolor='#e9b96e' colspan='5'>Call Stack</th></tr>
<tr><th align='center' bgcolor='#eeeeec'>#</th><th align='left' bgcolor='#eeeeec'>Time</th><th align='left' bgcolor='#eeeeec'>Memory</th><th align='left' bgcolor='#eeeeec'>Function</th><th align='left' bgcolor='#eeeeec'>Location</th></tr>
<tr><td bgcolor='#eeeeec' align='center'>1</td><td bgcolor='#eeeeec' align='center'>0.0133</td><td bgcolor='#eeeeec' align='right'>2283712</td><td bgcolor='#eeeeec'>CodeIgniter\Debug\Exceptions->exceptionHandler(  )</td><td title='/source/system/Debug/Exceptions.php' bgcolor='#eeeeec'>.../Exceptions.php<b>:</b>0</td></tr>
[...]

PR #2655, opened 2020-03-01
Fix JSON error on recursion in data

MashinaMashina:
Bugs:
https://github.com/codeigniter4/CodeIgniter4/issues/2434 and https://github.com/codeigniter4/CodeIgniter4/issues/2586

 _! $result &&_ needs because json_encode generating result, all is ok, but json_last_error() anyway eqally error "Type not supported"

### DIFF
--- a/system/Format/JSONFormatter.php
+++ b/system/Format/JSONFormatter.php
@@ -56,7 +56,7 @@ class JSONFormatter implements FormatterInterface
 	 */
 	public function format($data)
 	{
-		$options = JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES;
+		$options = JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_PARTIAL_OUTPUT_ON_ERROR;
 
 		$options = ENVIRONMENT === 'production' ? $options : $options | JSON_PRETTY_PRINT;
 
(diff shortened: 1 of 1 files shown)


MashinaMashina:
I do not know what to do with tests. 
In https://github.com/codeigniter4/CodeIgniter4/issues/2586 I want get error from database, not json formatter

MashinaMashina:
I do not know what to do with tests.
In https://github.com/codeigniter4/CodeIgniter4/issues/2586 I want get error from database, not json formatter

system/Format/JSONFormatter.php
```diff
@@ -56,13 +56,13 @@ class JSONFormatter implements FormatterInterface
...
 	{
-		$options = JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES;
+		$options = JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_PARTIAL_OUTPUT_ON_ERROR;
 
 		$options = ENVIRONMENT === 'production' ? $options : $options | JSON_PRETTY_PRINT;
 
 		$result = json_encode($data, $options, 512);
 
-		if (json_last_error() !== JSON_ERROR_NONE)
+		if ( ! $result && json_last_error() !== JSON_ERROR_NONE)
```
michalsn: I think this is a condition we want to have here. This should also fix the tests.
```suggestion
		if (! in_array(json_last_error(), [JSON_ERROR_NONE, JSON_ERROR_RECURSION]))
```

Issue #2616, opened 2020-02-25
Bug: Chrome logger does not work.


dafriend:
**Describe the bug**
Plain and simple, doesn't work. I think the framework changed significantly since the code was first written. The event that was to send the headers doesn't exist any longer.

I discovered this some time ago and devised a fix but never made a PR on it.

Feel free to assign this one to me. It might take a little while to get time for it. I mostly need to devise some unit tests.

MGatner:
Appreciate that you offered to look into it. I don’t use Chrome so I’m pretty unfamiliar with the structure but... any ideas why the tests are passing even if this has been broken for a while?

dafriend:
I knew at one point, but have forgotten now.

dafriend:
I refreshed my memory about the issue because I know y'all are curious why it passes the unit test but fails in the real world.

The unit test method calls ``ChromeLoggerHandler::sendLogs()`` which puts the required header into a response object. The test then checks the current response object for the head. It's there all right. Assertion true and we're in the green.

However.  here is how ``ChromeLoggerHandler`` is set up to call ``sendLogs()``

`Events::on('post_controller', [$this, 'sendLogs'], EVENT_PRIORITY_HIGH);`

Except there isn't an event named _'post_controller'_, and a non-existent event won't be triggered. Obviously.

MGatner:
You are so right! Good catch. I’d say just try switching to `post_controller_constructor` and see if that magically fixes it?

dafriend:
I actually have... er, that is, had a solution. I've made a mess of my local repo and cannot find the final working commit. It's in there somewhere. 

It would be nice to find since I tackled the @todo and worked out some fancy object formatting which users should appreciate. I also added logic so that the ChromeLogger cannot be instantiated in the production environment. We would be quite negligent to allow that to happen.

I'm on it though.

MGatner:
Fabulous. ;) I trust it to you

PR #2656, opened 2020-03-01
Logging Refactor

dafriend:
**Description**
Fix Chrome Logger per #2616 
But wait! There's more...
- It greatly enhances how the logger parses and displays PHP variables.
- Prevents ChromeLogger from running in a production environment.
- The documentation page was redone to accommodate the configuration changes and describe the logging of variables, etc.  

**Checklist:**
- [ x] Securely signed commits
- [ x] Component(s) with PHPdocs
- [ x] Unit testing, with >80% coverage
- [ x] User guide updated
- [?] Conforms to style guide

This might be a breaking change in that ``\Config\Logger`` has changed. Some new properties were added and others were effectively renamed.

However, the overall API is not changed and ``Logger`` still conforms to PSR-3.


app/Config/Logger.php
```diff
@@ -1,133 +1,186 @@
...
-	| By default, logs are written to WRITEPATH . 'logs/'
-	| Specify a different destination here, if desired.
-	*/
-	public $path = '';
+	/**
+	 * The $fileLevelsHandled property defines the message severity levels that
+	 * should be written to the log file.
+	 *
+	 * Here are the values and the message type
+	 *     0 - off
```
lonnieezell: Why remove the examples from this list?

app/Config/Logger.php
```diff
@@ -1,133 +1,186 @@
...
+	 * turns file logging off.
+	 *
+	 * If you put a single value in an array then only that level will be logged.
+	 * For instance, using
+	 *     $levelsHandled = [8];
+	 * then only 'debug' messages get logged.
+	 *
+	 * For a live site you'll usually want critical (3) or lower to be logged, otherwise
+	 * your log files will fill up very fast.
+	 *
```
lonnieezell: This amount of text is more appropriate for user guide than a doc block. The list of levels was provided as a quick reference only.

app/Config/Logger.php
```diff
@@ -1,133 +1,186 @@
...
+	 * @var integer|array
+	 */
+	public $fileLevelsHandled = 8;
```
lonnieezell: Please change the name back as this is definitely a BC break since this requires a change in every application that has been started on the framework so far.

PR #2665, opened 2020-03-06
fix. getYear can't week based year

### LINKED ISSUE #2663: Bug: \I18n\Time object displaying +1 year when object date is set to 2021-12-31
**Describe the bug**
Got the following object:
`CodeIgniter\I18n\Time (6) 2021-12-31 19:34:42-03:00 -03` retrieved from DB

I use `$oc->periodoFin->toLocalizedString("dd'-'MM'-'YYYY")` to echo it to screen and what I get is `31-12-2022`. I tried using `$oc->periodoFin->getYear()` and It still returns `2022`. (actually, I tried every method I could think of, same result).
On the other hand, when I post that variable with the correct date, it's saved successfully into the DB.
Also, if I d() the object onto the screen (same where the printing error occurs), I get the correct result (`CodeIgniter\I18n\Time (6) 2021-12-31 19:34:42-03:00 -03`).

This happens with dates ranging from `2021-12-26` till `2021-12-31` (both included).
If I use `2021-12-25` or `2022-01-01` (or any other date for that matter) things work as expected.

**UPDATE**
If I just echo $oc->periodoFin then date gets printed correctly, though not in the format I need it to be showed in.

**CodeIgniter 4 version**
4.0.0 final
**Affected module(s)**
CodeIgniter\I18n\Time

**Context**
 - OS: macOS 13.0.5
 - Web server:  Apache/2.4.41 (Unix)
 - PHP version: 7.3.8

Instrye:
Fix #2663 `CodeIgniter\I18n\Time` default format is `yyyy-MM-dd HH:mm:ss`, but `getYear` use `Y`,
the method return based week year. Last days of the some year, the method reutn current year +1.It's confusing 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -423,7 +423,7 @@ class Time extends DateTime
 	 */
 	public function getYear(): string
 	{
-		return $this->toLocalizedString('Y');
+		return $this->toLocalizedString('y');
 	}
 
 	//--------------------------------------------------------------------
--- a/tests/system/I18n/TimeTest.php
+++ b/tests/system/I18n/TimeTest.php
@@ -240,8 +240,10 @@ class TimeTest extends \CodeIgniter\Test\CIUnitTestCase
 	public function testGetYear()
 	{
 		$time = Time::parse('January 1, 2016');
+		$time2 = Time::parse('December 31, 2019');
 
 		$this->assertEquals(2016, $time->year);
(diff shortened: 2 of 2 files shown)


fefo-p:
Thanks!

PR #2666, opened 2020-03-06
Fix ResourceController Issue, including Test Unit

### LINKED ISSUE #2617: Bug: ResourceController json response always empty

arif-rh:
This is an alternative way to fix #2617 

I found that variable `$format` should be moved to `ResponseTrait` , so that it can be used inside `format` method.

I have add two unit testing for this change

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/API/ResponseTrait.php
+++ b/system/API/ResponseTrait.php
@@ -373,11 +373,15 @@ trait ResponseTrait
 		// Determine correct response type through content negotiation
 		$config = new Format();
 
-		if (!in_array($this->format, ['json', 'xml']))
+		if (! in_array($this->format, ['json', 'xml']))
+		{
 			$format = $this->request->negotiate('media', $config->supportedResponseFormats, false);
-		else 
+		}
+		else
+		{
 			$format = "application/$this->format";
-			
+		}
+
 		$this->response->setContentType($format);
 
 		// if we don't have a formatter, make one
--- a/system/API/ResponseTrait.php
+++ b/system/API/ResponseTrait.php
@@ -93,6 +93,12 @@ trait ResponseTrait
 		'not_implemented'           => 501,
 	];
 
+	/**
+	 *
+	 * @var string the representation format to return resource data in (json/xml)
+	 */
+	protected $format = 'json';
+
 	//--------------------------------------------------------------------
 
 	/**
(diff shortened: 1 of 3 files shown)


arif-rh:
Hi @MGatner , I have updated to follow the style guide

lonnieezell:
Unfortunately I don't know exactly what Jim had planned here. Doing it as an after filter definitely makes sense at first glance.

system/API/ResponseTrait.php
```diff
@@ -366,7 +372,15 @@ protected function format($data = null)
 
 		// Determine correct response type through content negotiation
 		$config = new Format();
-		$format = $this->request->negotiate('media', $config->supportedResponseFormats, false);
+
+		if (! in_array($this->format, ['json', 'xml']))
```
MGatner: Does hard-coding the available formats prevent meaningful extension? I wonder if this should be checking again the application-configured formatters instead (**app/Config/Format.php**) in case a developer has defined their own.

arif-rh: what I understand from `ResourceController` is this feature can set default format for an API. For example, all API by default using json format, but one API will have xml format, so we can overide this by `setFormat` function.

setFormat

Currently, setFormat only validate two kind of format, either json or xml, that why it need to be re-checked, by hard-coding the available format. 

In case a developer has defined their own format, he can set it up with custom format then system will check again using request negotiate.

@MGatner , tell me if my understanding is wrong.

MGatner: No, you are very right! I didn’t realize that was hard-codes into `setFormat` as well. Might be worth changing these down the road but we’ll leave that for the next pass at the RESTful, maybe with the After Filter.

Issue #2606, opened 2020-02-24
Unnecessary files in the download installation


jlamim:
Analyzing the set of files downloaded in one installation via composer and another via manual download I realized that in manual installation there are many unnecessary files, which can end up leaving the user confused.

The following is a list of files and folders that are not part of the system core and could be removed:

- / contributing
- / admin
- /.github
- PULL_REQUEST_TEMPLATE.md
- stale.yaml
- .nojekyll
- Vagrant.dist
- DCO.md
- CHANGELOG.md
- .editorconfig

fefo-p:
admin files are needed (at least **admin/setup.sh**) after running `composer update`

zahhar:
I would not name this a bug, rather enchancement.

MGatner:
Some of these files indeed should not be there, this is a bug. The download version also assumes composer is not being used so I think excludes the admin scripts, but will check on that.

fefo-p:
Though, whoever downloads CI4 will be developing over it, so maybe including a simple text file, stating which files are needed to run vs which are needed to develop, would be helpful.
Just a thought

MGatner:
I am close to a rework of the release logic and need a verdict on which files go where. @lonnieezell may just want to make that call but this is a good place for community feedback.

***

Here is the current behavior (all repo references are in the `codeigniter4/` org):

Manual downloads and AppStarter's vendor directory both contain `framework`. This repo [receives the following](develop/admin/release-framework) from `codeigniter4`:
```
app
docs
public
system
writable
contributing.md
env
license.txt
spark
.gitignore
tests/_support
admin/framework/README.md (to root)
admin/framework/composer.json (to root)
admin/framework/phpunit.xml.dist (to root)
```

AppStarter contains `appstarter`. This repo [receives the following](develop/admin/release-appstarter) from `codeigniter4`:
```
app
public
writable
contributing.md
env
license.txt
spark
tests/_support
admin/starter/.github/workflows/phpunit.yml (to root)
admin/starter/.gitignore (to root)
admin/starter/README.md (to root)
admin/starter/builds (to root)
admin/starter/composer.json (to root)
admin/starter/phpunit.xml.dist (to root)
admin/starter/app/Config/Paths.php (to root)
admin/starter/tests (to root)
```

[...]

PR #2668, opened 2020-03-07
Releases Refactor

MGatner:
**Description**
This PR is "phase 1" of automating releases, targeting the distribution repos.
* A new GitHub Action is defined, triggered by release publication
* Release scripts are retired in favor of a new deploy scripts in **.github/scripts**
* Repo files are differentiated by intended project use (see also https://github.com/codeigniter4/CodeIgniter4/issues/2606)
* Docs updated for the release workflow
* A number of bugs fixed that either included wrong files or failed to remove legacy files

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


MGatner:
Due to the nature of the trigger, this change is not easily tested. I created forks of the three repos `codeigniter4`, `framework`, and `appstarter` and temporarily pointed the Action at those. You can review the successful test:
* New release created: https://github.com/MGatner/CodeIgniter4/releases/tag/v6.0.0
* Action triggered, jobs successful: https://github.com/MGatner/CodeIgniter4/actions/runs/51138279
* Distribution repos updated with corrected files: https://github.com/MGatner/framework/commit/7a7fc5cefb97e704fa52583d97bb9d20ce9e2c80
* Corresponding distribution releases generated: https://github.com/MGatner/framework/releases/tag/v6.0.0, https://github.com/MGatner/appstarter/releases/tag/v6.0.0

.github/scripts/deploy-appstarter
```diff
@@ -0,0 +1,34 @@
...
+echo "Merging files from $1 to $2"
+
+# Prepare the source
+cd $SOURCE
+git checkout master
+
+# Prepare the target
+cd $TARGET
+git checkout master
+rm -rf *
```
lonnieezell: This command scares me :) Can you explain it is doing. Seems like you checkout master branch then delete everything?

lonnieezell:
Other than the delete the entire server command making me nervous it all looks good. :)

.github/scripts/deploy-appstarter
MGatner: The way the old script worked it would copy in the chosen files (like we do below) pre-wiping any directories (this was the bugged part from the `cprm` typo). Where that fails is when there are files that are renamed or removed (notice we have license.txt and a duplicate legacy LICENSE because of this). Using the modified version above it checks out the target repo (AppStarter or framework) and rebuilds the repo from scratch (* ignores files with leading periods so it doesn’t affect the `.git` directory). Also, it’s just a new commit so we could always back up one and nothing is lost!

MGatner:
Come on @lonnieezell, you know you want to.....

lonnieezell:
Oh, all right @MGatner :)

PR #2675, opened 2020-03-09
WhereIn() not supporting non array variables 🐛

jmcelreavey:

**Description**
- When using WhereIn if the variable being passed in is not an array or closure it is simply returned with no user feedback. 
    - However we have logic in place which allows regular values.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -987,12 +987,22 @@ class BaseBuilder
 	{
 		if (empty($key) || ! is_string($key))
 		{
-			throw new InvalidArgumentException(sprintf('%s() expects $key to be a non-empty string', debug_backtrace(0, 2)[1]['function']));
+			if (CI_DEBUG)
+			{
+				throw new InvalidArgumentException(sprintf('%s() expects $key to be a non-empty string', debug_backtrace(0, 2)[1]['function']));
+			} 
+			
+			return this;
 		}
 
 		if ($values === null || (! is_array($values) && ! ($values instanceof Closure)))
 		{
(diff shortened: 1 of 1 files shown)


michalsn:
Technically speaking this is not a bug. Ref: https://codeigniter4.github.io/userguide/database/query_builder.html?highlight=wherein#whereIn

jmcelreavey:
> Technically speaking this is not a bug. Ref: https://codeigniter4.github.io/userguide/database/query_builder.html?highlight=wherein#whereIn

True, maybe if it threw an exception instead. Just caught me out since in CodeIgniter 3 this worked without issue and failing that it would have thrown an exception if it didn't like the type

michalsn:
You're right. It worked, but it was more like an undocumented feature. The CI3 documentation also states that we should provide an array, in the second parameter.

If you ask me, I would go back to the code we had and just throw an exception.

jmcelreavey:
> You're right. It worked, but it was more like an undocumented feature. The CI3 documentation also states that we should provide an array, in the second parameter.
> 
> If you ask me, I would go back to the code we had and just throw an exception.

Fair enough, I'm happy either way. I've updated my pull request to reflect this.

michalsn:
One last suggestion. We should throw an exception only when CI_DEBUG is enabled., Simple check: `if (CI_DEBUG)` will be enough, and in other case, we should just return `$this` as before.

MGatner:
Looking good. @jmcelreavey could you implement the `CI_DEBUG` suggestion to keep it in line with other Builder exceptions?

jmcelreavey:
@michalsn - good point! @MGatner - Done! 👍

MGatner:
Thanks! I’ll watch for the tests

Issue #2678, opened 2020-03-09
Bug:   The custom error config of validation is not working


iguojun:
**Describe the bug**
According to the document description(/docs/libraries/validation.html#id13), 
I set custom error message as the labeled style in "\Config\Validation",  
```
// App\Config\Validation.php

public $login = [
    'username' => [
        'label'  => 'Username',
        'rules'  => 'required',
        'errors' => [
            'required' => 'custom username required error msg.',
        ],
    ],
    'password' => [
        'label'  => 'Password',
        'rules'  => 'required',
        'errors' => [
            'required' => 'custom password required error msg.',
        ],
    ]
];
```

Run the script below,
```
// App\Controllers\SomController.php

$validation =  \Config\Services::validation();
$validation->reset()->setRuleGroup('login')->run();
$errors = $validation->getErrors();
print_r($errors);
```

And I lost the custom error and got the default error message instead.
```
Array
(
    [username] => The Username field is required
    [password] => The Password field is required.
)
```
**CodeIgniter 4 version**
CodeIgniter 4.0.2

**Affected module(s)**
\App\Controller

**Context**
 - OS: [CentOS7]
 - Web server [Apache/2.4.41 (codeit) OpenSSL/1.1.1d PHP/7.4.3]
 - PHP version [7.4.3]

PR #2679, opened 2020-03-09
Fixed issue of  phrase custom validation error message config

iguojun:
After getRuleGroup($group),
Instead asign to $this->rule,
We should call $this->setRules($rules) at least once.

**Description**
See: (https://github.com/codeigniter4/CodeIgniter4/issues/2678)

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -475,7 +475,7 @@ class Validation implements ValidationInterface
 	public function setRuleGroup(string $group)
 	{
 		$rules       = $this->getRuleGroup($group);
-		$this->rules = $rules;
+		$this->setRules($rules);
 
 		$errorName = $group . '_errors';
 		if (isset($this->config->$errorName))


musmanikram:
@iguojun Adding some more tests would be better

lonnieezell:
Yes, this needs a test or two to test the specific issue you're trying to fix.

iguojun:
> @iguojun Adding some more tests would be better

Thanks for your response!

As a new php learner, I've no idea how to write a test script now, and I'll learn and do my best.

musmanikram:
@iguojun that's k no worries mate.
Will help you in this.

MGatner:
IMO this is the correct way to handle this, and a method should never set `$this->rules` directly. Some tests would be great but I think this should be merged even if we can't come up with them.

@iguojun Tests aren't as scary as they seem! Check out **tests/system/Validation/ValidationTest.php** and likely just add a couple methods there, modeled after the existing ones.

musmanikram:
@MGatner I will push some tests to @iguojun PR. Working on it

musmanikram:
@MGatner i can't push my changes to @iguojun PR.
@iguojun Please have a look on my PR and you can get some idea how to write some tests. I used the same input you added in your issue.

MGatner:
@musmanikram I think only maintainers can push to PR branches on approved forks. You can always fork someone else’s branch (or add their fork as a remote) and merge their work. I think since this PR is a one-line change I will close it in favor of yours, but I would still like it if @iguojun took a look at the new one, #2699

PR #2686, opened 2020-03-11
Fix #2676 : add ability to test redirect()->route() via ControllerTester

### LINKED ISSUE #2676: Bug: Can't test redirect()->route('routename'), while redirect()->to('path') is working with ControllerTester
**Describe the bug**

For example, I have a controller function `Home::index` that redirect to route "product-index":

```php

{
	public function index()
	{
		return redirect()->route('product-index');
	}
}
```

On Test part, I create test with: 

```php
<?php namespace AppTest\Controller;

use App\Controllers\Home;
use CodeIgniter\Test\CIDatabaseTestCase;
use CodeIgniter\Test\ControllerTester;

class HomeTest extends CIDatabaseTestCase
{
	use ControllerTester;

	public function testIndexRedirectToProduct()
	{
		$result = $this->controller(Home::class)
					   ->execute('index');

		$this->assertTrue($result->isRedirect());
	}
}
```

and it got error: 

```bash
vendor/bin/phpunit tests/Controller/HomeTest.php
PHPUnit 8.5.2 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 284 ms, Memory: 12.00 MB

There was 1 error:

1) AppTest\Controller\HomeTest::testIndexRedirectToProduct
CodeIgniter\HTTP\Exceptions\HTTPException: 

/Users/samsonasik/www/ci4/vendor/codeigniter4/framework/system/HTTP/Exceptions/HTTPException.php:168
/Users/samsonasik/www/ci4/vendor/codeigniter4/framework/system/HTTP/Response.php:314
/Users/samsonasik/www/ci4/vendor/codeigniter4/framework/system/Test/ControllerTester.php:191
/Users/samsonasik/www/ci4/tests/Controller/HomeTest.php:14
```

If I change the code with `redirect()->to(base_url('product'))`.

```php 

{
	public function index()
	{
		return redirect()->to(base_url('product'));
	}
}
```

It working: 

```bash
vendor/bin/phpunit tests/Controller/HomeTest.php
PHPUnit 8.5.2 by Sebastian Bergmann and contributors.

Time: 258 ms, Memory: 12.00 MB

OK (1 test, 1 assertion)
```

**CodeIgniter 4 version**
4.0.2

Working with `redirect()->route()`

**Context**
 - OS: OSX
 - Web server Apache
 - PHP version 7.4

samsonasik:
Fixes #2676 to add ability to test redirect()->route() in test bootstrap so the existing route both in `app` and `tests` will be discovered.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/admin/starter/composer.json
+++ b/admin/starter/composer.json
@@ -12,6 +12,12 @@
 		"mikey179/vfsstream": "1.6.*",
 		"phpunit/phpunit": "8.5.*"
 	},
+	"autoload": {
+		"psr-4": {
+			"App\\": "app",
+			"Config\\": "app/Config"
+		}
+	},
 	"autoload-dev": {
 		"psr-4": {
 			"Tests\\Support\\": "tests/_support"
--- a/admin/starter/composer.json
+++ b/admin/starter/composer.json
@@ -12,12 +12,6 @@
 		"mikey179/vfsstream": "1.6.*",
 		"phpunit/phpunit": "8.5.*"
 	},
-	"autoload": {
-		"psr-4": {
-			"App\\": "app",
-			"Config\\": "app/Config"
-		}
-	},
 	"autoload-dev": {
 		"psr-4": {
 			"Tests\\Support\\": "tests/_support"
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -59,7 +59,5 @@ $loader->initialize(new Config\Autoload(), new Config\Modules());
 // Register the loader with the SPL autoloader stack.
 $loader->register();
 
-$routes = \Config\Services::routes();
-$routes->getRoutes('*');
-
 require_once APPPATH . 'Config/Routes.php';
+$routes->getRoutes('*');
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -61,3 +61,5 @@ $loader->register();
 
 $routes = \Config\Services::routes();
 $routes->getRoutes('*');
+
+require_once APPPATH . 'Config/Routes.php';
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -13,14 +13,14 @@ require CONFIGPATH . 'Paths.php';
 $paths = new Config\Paths();
 
 // Define necessary framework path constants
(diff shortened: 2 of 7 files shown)


samsonasik:
I've added unit test for it.

samsonasik:
It seems this method only check routes in `test/_support` part, I will check to add namespace `Config` and `App` as well.

samsonasik:
I got the solution for autoload in test doesn't detect `App` and `Config` namespace in project structure in bootstrap test! 

I register "App" and "Config" to `admin/starter`'s composer.json so when running test from the appstarter project in the next release, it will always work as it discoverable from composer as running test means developer using composer as autoloader.

samsonasik:
travis build on php72:postgres is unrelated with it.

samsonasik:
I found a better solution. Instead of add the autoloader, I added a 1 line after setup routes, which is a require_once of `APPPATH . 'Config/Routes.php'`, so the solution is add the following to `system/Test/bootstrap.php`:

```php 
$routes = \Config\Services::routes();
$routes->getRoutes('*');

require_once APPPATH . 'Config/Routes.php';
```
That's it! So developer don't need to update their composer.json.

samsonasik:
ready for review. travis build php73:mysqli is unrelated.

samsonasik:
actually, the `APPPATH . 'Config/Routes.php` already define `Services::routes(true)` and assign to `$routes` variable, so, I updated to use existing `$routes` variable and call `getRoutes(*)` on it:

```php
require_once APPPATH . 'Config/Routes.php';
$routes->getRoutes('*');
```

samsonasik:
travis build green 🎉

MGatner:
This looks good and well written. I would like to test it with a couple existing projects and modules to make sure loading the routes that early doesn't effect other services, like mocking.

samsonasik:
@MGatner merge-able ?

MGatner:
I still haven't tried it! Let me do a couple right now.

MGatner:
Okay, looks good! Tested two projects and two modules and this change didn't seem to affect them.

MGatner:
@samsonasik Thanks for the ping BTW. I've got a ton going on and the repo has been quite busy. If you see other things that merit my attention don't hesitate to mention me.

PR #2690, opened 2020-03-11
Bugfix: session->get() need return null if data not exist

belousovsw:
The result is not as expected if the key is not found

I think, more be correctly if the function get() always return null if key not found in session.

### DIFF
--- a/system/Session/Session.php
+++ b/system/Session/Session.php
@@ -487,13 +487,13 @@ class Session implements SessionInterface
 	 * Get user data that has been set in the session.
 	 *
 	 * If the property exists as "normal", returns it.
-	 * Otherwise, returns an array of any temp or flash data values with the
+	 * Otherwise, returns an null of any temp or flash data values with the
 	 * property key.
 	 *
 	 * Replaces the legacy method $session->userdata();
 	 *
 	 * @param  string $key Identifier of the session property to retrieve
-	 * @return array|null	The property value(s)
+	 * @return string|array|null	The property value(s)
 	 */
 	public function get(string $key = null)
 	{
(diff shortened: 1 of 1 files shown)


belousovsw:
We make definitely function
- If is set data in session, we receive data.
- If no data we get 'null'
If we modify the function as I suggested, we get clear and understanding data.
```
$s1->set([
    'string' => 'hello', 
    'array1' => [1,2,3],
    'array2' => [],
]);
$x1 = $s->get('string')??''; // hello
$x2 = $s->get('stringX')??'by'; // by
$x3 = $s->get('array1')??'[]; // [1,2,3]
$x4 = $s->get('array2')??[1,2,3]; // [1,2,3] 
$x5 = $s->get('arrayX)??[]; // []

foreach($s->get()??'[] AS $x) {
   `echo` $x; // helo | [1,2,3] | []
}
foreach($sX->get()??'[] AS $x) {
   `echo` $x; // not executed
}

```
In result we get clear code.

But if we don't change function, we need code like this:
```
$s1->set([
    'string' => 'hello', 
]);
$x1 = $s->get('string')??''; // hello
$x2 = $s->get('stringX')??'by'; // [] *** this is not correct, I wait string in this, not array
if (empty($x2)) {
    $x2 = 'by';
} 
```

michalsn:
This is the only place I see we can do any improvements here (it would be a very rare case):
```php
elseif (empty($_SESSION))
{
	return $key === null ? [] : null;
}
```
This way if no `$_SESSION` is available we can return `null` when someone calls `get($key)` and `[]` when `get(null)` is called. 

Other than that, the original code in the framework is working just as we would expect it to work. No changes are needed. Calling `get(null)` should return an empty array if there is no data, and calling `get(key)` should return null when this key is not set - and that's how it works now.

I consider the examples given to justify changes in the code to be not valid.

PR #2698, opened 2020-03-12
[ci skip] redis config in cache configuration

Instrye:
`app/Config/redis.php` not work


lonnieezell:
Thanks for the PR. I think it's worth keeping the list of options (though updated for our config file). Could you add those back in?

Instrye:
CI4 has been released, Unless necessary, the contents of the App directory should not be modified. Composer update can't replace `App\Config`.

If really need it, I will delete the redis configuration in session and cache so that they use `App\Config\Redis.php` uniformly. same `App\Config\Database`

Instrye:
Sorry, I understand you wrong.

PR #2703, opened 2020-03-14
Fix ChromeLogger

dafriend:
**Description**
Fixes Issue #2616


MGatner:
Thanks for taking a pass at this! Looks like a test just needs to be updated for the new format?

You did a lot of work for #2656 and I know there were some issues but I'd hate to see the baby go with the bath water. Is there something we can do to help recover that?

dafriend:
> Thanks for taking a pass at this! Looks like a test just needs to be updated for the new format?

Here's the fix for the failing unit test.

tests/system/Log/ChromeLoggerHandlerTest.php
```diff
@@ -56,23 +55,26 @@ public function testSetDateFormat()
...
 		$data->code        = 123;
 		$data->explanation = "That's no moon, it's a pumpkin";
 		$result            = $logger->setDateFormat('F j, Y');
 
-		$logger->handle('debug', $data);
-		$peek = $logger->peekaboo();
+		$logger->handle('warning', $data);
 
-		$this->assertEquals($data->explanation, $peek[0]['explanation']);
+		$this->assertTrue($response->hasHeader('x-chromelogger-data'));
```
lonnieezell: Maybe I'm mis-reading something here, but this should actually be false, shouldn't it? If `TestHandler` is only set to log `critical` errors, and we tell the logger to handle a `warning`, then TestHandler (and therefor `ChromeLoggerHandler`) shouldn't send any data, right?

dafriend: The decision to actually log (or not) is made in ``\CodeIgniter\Log\Logger::log()`` and despite the object being called ``$logger`` it is not an instance of ``Logger`` it is an instance of ``ChromeLoggerHandler``.  The severity level of the message has no bearing on this test.

lonnieezell: Doh! You're right. Thanks for clarifying. I'm good with this then.

lonnieezell:
Thanks @dafriend !

PR #2706, opened 2020-03-15
fix #2705 : countAllResults after find

### LINKED ISSUE #2705: Bug: calling countAllResults after find($id) produce wrong result
**Describe the bug** 

calling `countAllResults` after `find($id)` produce wrong result. for example, I made below tests:

````php
    public function testFindIdCountAllResults()
	{
		$model = new UserModel($this->db);

		$model->find(1);

		$this->assertEquals(1, $model->countAllResults());
	}

	public function testFindArrayCountAllResults()
	{
		$model = new UserModel($this->db);

		$model->find([1,2]);

		$this->assertEquals(2, $model->countAllResults());
	}
````
and both produce an error, because all `countAllResults` is 4

find countAllResults

This issue only happen when we passing `$id` to `find` method. If we omit it, it will work fine, because without passing `$id` it will call `findAll` instead.

**CodeIgniter 4 version** 

4.0.2

**Affected module(s)** 

CodeIgniter\Model

Make test as description above to reproduce the issue

**Context**

- OS: Windows 10
- Web server Apache 2.4.37
- PHP version 7.3.1

arif-rh:
**Description**
This PR fixes #2705 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -356,11 +356,12 @@ class Model
 	 * Fetches the row of database from $this->table with a primary key
 	 * matching $id.
 	 *
-	 * @param mixed|array|null $id One primary key or an array of primary keys
+	 * @param mixed|array|null $id    One primary key or an array of primary keys
+	 * @param boolean          $reset Are we want to clear query builder values after find?
 	 *
 	 * @return array|object|null    The resulting row of data, or null.
 	 */
-	public function find($id = null)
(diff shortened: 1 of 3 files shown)


lonnieezell:
Thanks for the good work, but see comment on #2705

PR #2707, opened 2020-03-15
Fix #2704: ensure route registered via $routes->cli() not accessible via web browser even autoroute is true

### LINKED ISSUE #2704: Bug: $routes->cli() accessible via web browser if autoroute is true
For example, I have the following  `cli` route: 

```php 
// set auto route to true
$routes->setAutoRoute(true); 

// set "hello" cli route
$routes->cli('hello', 'Commands\Hello::index');
```

and I have the controller: 

```php 
<?php
namespace App\Controllers\Commands;

class Hello extends BaseController
{
    public function index()
    {
        echo 'Hello';
    }
}
```

then, I open it via web browser, it still accesible via : `http://ci4.local/commands/hello` , it still accessible.

**CodeIgniter 4 version**
4.0.2

By define `cli()`, it should decline the http access into it. The documentation doesn't say it require autoroute to be false: 

```
You can create routes that work only from the command-line, and are inaccessible from the web browser, with the cli() method
```

**Context**
 - OS: OSX
 - Web server Apache
 - PHP version 7.4

samsonasik:
Fix #2704: ensure route registered via $routes->cli() not accessible via web browser even autoroute is true.

**Checklist:**
- [x] Securely signed commits


samsonasik:
**Before** : 
Screen Shot 2020-03-15 at 5 58 15 PM

**After**: 
Screen Shot 2020-03-15 at 5 58 25 PM

samsonasik:
Not sure how to write unit test for it. I included screenshot before and after changed above.

samsonasik:
@MGatner I've added feature test case for demoing 404 page thrown when open cli routes from http

samsonasik:
it seems need some autoload tweak for testing "App" namespace for feature test demo, I will re-check again.

samsonasik:
The test tweak implemented.

samsonasik:
@@MGatner travis build green 🎉

system/Router/Router.php
```diff
@@ -542,6 +542,25 @@ public function autoRoute(string $uri)
...
+			$controller  = '\\' . $this->collection->getDefaultNamespace();
+			$controller .= $this->directory ? str_replace('/', '\\', $this->directory) : '';
+			$controller .= $this->controllerName();
+			$methodName  = $this->methodName();
+
+			foreach ($this->collection->getRoutes('cli') as $route)
+			{
+				if (is_string($route))
+				{
+					if (strpos($route, $controller . '::' . $methodName) === 0)
```
samsonasik: using `strpos` is on purpose to handle parameter, eg: controller::method/$1: 

```php
$routes->cli('hello/(:any)', 'Commands\Hello::index/$1');
```

samsonasik:
I've added feature test case for demoing 404 page thrown when open parameterized cli routes from http

samsonasik:
I've added condition check for default method index when routing defined in cli as controller name only, which fallback to method index, eg:

```php 
$routes->cli('hello', 'Hello');
```

samsonasik:
I've added condition for multi case, eg: capitalized controller or method in URI with compare with strtolower of the controller and method.

samsonasik:
travis build green 🎉

tests/system/Test/FeatureTestCaseTest.php
```diff
@@ -204,4 +204,54 @@ public function testCallZeroAsPathGot404()
...
+			],
+		];
+	}
+
+	/**
+	 * @dataProvider provideRoutesData
+	 */
+	public function testOpenCliRoutesFromHttpGot404($from, $to, $httpGet)
+	{
+		$this->expectException(PageNotFoundException::class);
```
kenjis, line 239: These tests pass because there is no `App\Controllers\Hello` class.

kenjis, line 239: These tests pass because there is no `App\Controllers\Hello` class.

PR #2712, opened 2020-03-17
Bug fix: fix missing Migrations.migWriteError lang data

samsonasik:
The "migWriteError" language data is not registered at `system/Language/en/Migrations.php` file, so, when write migration failure, it was only get text "Migrations.migWriteError".

I've ~~added the "migWriteError" language~~ fixed it with use existing "Migrations.writeError" data to `system/Language/en/Migrations.php` file. I provided unit test to it at tests/system/Commands/SessionsCommandsTest.php.

**Checklist:**
- [x] Securely signed commits


samsonasik:
I checked that there is another file that uses "Migrations.writeError" language data at `system/Command/Database/CreateMigration.php`.

For consistency, I updated to use existing "Migrations.writeError", added path to parameter so the error is more verbose.

system/Commands/Sessions/CreateMigration.php
```diff
@@ -126,7 +126,7 @@ public function run(array $params = [])
 		helper('filesystem');
 		if (! write_file($path, $template))
 		{
-			CLI::error(lang('Migrations.migWriteError'));
+			CLI::error(lang('Migrations.writeError', [$path]));
```
samsonasik: I updated with existing `Migrations.writeError` language data.

samsonasik:
@MGatner travis build at 7.2:sqlite seems unrelated as due to redis

MGatner:
This has been failing on a few tests. Unfortunately I cannot override failed CI results so @lonnieezell will need to get involved.

samsonasik:
I will close and re-open the PR to trigger build

samsonasik:
@MGatner travis build green 🎉

PR #2717, opened 2020-03-17
Bug Fix: Fix #2716 open path /0 got error on development and 200OK on production, which should be 404

### LINKED ISSUE #2716: Bug: open path /0 at uri got error "Class Home does not exist" on development environment
When on development environment, open page eg: http://ci4.local/0 got "Class Home does not exist" and when open at production environment it doesn't has error, but show the Home page, I think it should be 404 instead.

**CodeIgniter 4 version**
4.0.2

On development and production env, it should be 404 page.

**Context**
 - OS: MacOS
 - Web server Apache
 - PHP version 7.4

samsonasik:
Fixes #2716 .

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -616,6 +616,7 @@ class IncomingRequest extends Request
 			$this->uri->setScheme(parse_url($baseURL, PHP_URL_SCHEME));
 			$this->uri->setHost(parse_url($baseURL, PHP_URL_HOST));
 			$this->uri->setPort(parse_url($baseURL, PHP_URL_PORT));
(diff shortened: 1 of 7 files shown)


samsonasik:
Implemented the fix. Here is the result in autoroute = true.

1. If defined in routes configuration, the `/0` path is allowed.
2. If not defined in routes configuration, it will find controller named "0" which will be fallback to 404 page.

I've added unit test in URI, Router, and RouteCollection test classes.

samsonasik:
I've added feature test case as well for demoing 404 page thrown when access `/0` if it is not registered in the route.

samsonasik:
travis build green 🎉

PR #2721, opened 2020-03-18
Fix validation for array input fields

### LINKED ISSUE #2714: Array validation has a problem
Hello!

I am trying to validate an array structure that comes from my form:
```html
<input type="checkbox" name="category[]" value="id2">
<input type="checkbox" name="category[]" value="id3">
<input type="checkbox" name="category[]" value="id4">
```

I use the validation library dynamically building the rules for each field which would arrive as `category.0`, `category.1` ... 

```php
foreach($data['categorie'] as $key => $value){
    $rules[ 'category.' . $key ] = 'required|alpha|max_length[2]';
}  
```
validation only works well for all values that have key 1 forwards, but does not correctly identify the field that is in the `category.0` position validation fails with "field is required"

```php
 ["category.0"]=>
  string(34) "The categorie.0 field is required.
```

I have debugged that all the fields get to my controller correctly but it seems to be a validation library error.

thank you very much for your amazing work!

musmanikram:
Fix #2714 

**Description**
In function `_array_search_dot` empty check for `$currentIndex` was failing for index 0 in array.
`empty()` will return `true` for value `0`. 

 Details: https://www.php.net/manual/en/function.empty.php

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -79,7 +79,7 @@ if (! function_exists('_array_search_dot'))
 			? array_shift($indexes)
 			: null;
 
-		if (empty($currentIndex) || (! isset($array[$currentIndex]) && $currentIndex !== '*'))
+		if ((empty($currentIndex)  && intval($currentIndex) !== 0) || (! isset($array[$currentIndex]) && $currentIndex !== '*'))
 		{
 			return null;
 		}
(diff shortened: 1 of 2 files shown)


musmanikram:
@MGatner can you please merge this also? It's issue has been closed by the reporter.
Because i gave him a solution

PR #2724, opened 2020-03-19
BaseBuilder - syncOriginal called twice on data fetch for Entity Class

najdanovicivan:
**Description**
BaseResult sets data on Entity object using setAttributes method. syncOriginal is called from setAttributes method so there is no need to call it again from the BaseResult.

I don't know if there is even a good reason to call it from BaseResult. I might be useful if there is a need for data processing after it's been loaded from DB but in such case syncOriginal might not be a good name for the method.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseResult.php
+++ b/system/Database/BaseResult.php
@@ -39,6 +39,8 @@
 
 namespace CodeIgniter\Database;
 
+use CodeIgniter\Entity;
+
 /**
  * Class BaseResult
  */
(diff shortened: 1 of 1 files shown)


lonnieezell:
I believe the point of this was simply to record the state of the entity as it existed after being pulled from the database. That way someone can make changes to the entity and pass it back to the model in a `save()` or `update()` and it can check to see what has actually been changed. So it is necessary to call it from BaseResult. Not necessary to call it twice, though.

PR #2726, opened 2020-03-20
Fixes #2725 : move logging directory path config to CodeIgniter\Log\Handlers\FileHandler config

### LINKED ISSUE #2725: Bug: Logger `path` property ignored
CI V 4.0.2 on develop at commit 84b9cc

**Affected module(s)**
``CodeIgniter\Log\Logger``

``Config\Logger::path`` setting is not used as the value for where to save log files.

Reproduce by setting some folder other than the default in ``Config\Logger::path``  e.g.

    public $path = WRITEPATH . 'uploads/';

Then log something somewhere, e.g.

```

{
    public function index()
    {
	$this->logger->emergency("This will not be written to a file in the WRITEPATH/uploads folder");
    }
}
```

File will be written to  `WRITEPATH/logs` instead of `WRITEPATH/luploads`.

**Context**
 - Linux Debian 10
 - Apache/2.4.38 (Debian)
 - PHP version  7.3.14

samsonasik:
Fixes #2725 . The Logger::$path config never get as it tried to access the path config in "CodeIgniter\Log\Handlers\FileHandler" config under Logger::$handlers property. Also, the `??` parameter make test for nullable/not isset, while it check the empty string value. Now, to avoid BC break, I use the following flow:

> note: $config is a configuration under Logger::$handlers['CodeIgniter\Log\Handlers\FileHandler'], not Logger class properties

- set $config['path'] to $config['path'] if isset, fallback to WRITEPATH . 'logs/';
- $this->path to $config['path'] if not empty string, fallback to WRITEPATH . 'logs/';

**Checklist:**
- [x] Securely signed commits


samsonasik:
Got error in test : "fopen(vfs://root/log-2020-03-20.log): failed to open stream: "org\bovigo\vfs\vfsStreamWrapper::stream_open" call failed". It seems the vfsStream() put -> read requires tweak.

samsonasik:
The test has been tweaked for writing file with vfsStream

samsonasik:
travis build green :tada:

MGatner:
This looks good to me. There should be no need for `path` in the base config since it is handled-specific. Since it’s a change to **app** I’d like @lonnieezell to take a look before merging. Also @dafriend has spent a lot of time with logging recently and may have some input. 
Thanks for the work, these have been very solid PRs.

app/Config/Logger.php
```diff
@@ -109,7 +109,7 @@ class Logger extends BaseConfig
 			'filePermissions' => 0644,
 
 			/*
-			 * Logging Directory Path
+			 * Error Logging Directory Path
```
dafriend: Except sometimes the message isn't an error.

system/Log/Handlers/FileHandler.php
```diff
@@ -77,7 +77,8 @@ public function __construct(array $config = [])
 	{
 		parent::__construct($config);
 
-		$this->path = $config['path'] ?: WRITEPATH . 'logs/';
+		$config['path'] = $config['path'] ?? WRITEPATH . 'logs/';
+		$this->path     = $config['path'] ?: WRITEPATH . 'logs/';
```
dafriend: I suggest this as it's less code and uses the same pattern as the other property assignments.

```
    $this->path = empty($config['filePath']) ? WRITEPATH . 'logs/' : $config['filePath'];
```

app/Config/Logger.php
samsonasik: resolved.

system/Log/Handlers/FileHandler.php
samsonasik: @dafriend implemented.

samsonasik: @dafriend I rolled back the changes to use the `??` first, and `?:` later as it can make for whatever reason user create folder named "0", it will still working.

samsonasik:
I close re-open the PR to re-trigger travis build

system/Log/Handlers/FileHandler.php
samsonasik: @dafriend I just realized that the directory path requires `/` suffix, so "0" is not valid, "0/" is valid. I applied your suggestion.

samsonasik:
@dafriend I applied your suggestion, travis build green 🎉

lonnieezell:
I think this change makes sense. My initial reaction was that it was a BC change that would cause a new MAJOR release, which we don't want to do. But I think another PR just removed that anyway, since it wasn't being used, and this doesn't affect the API so I think we're OK. Will merge.

PR #2729, opened 2020-03-20
Fix issue set_radio when value is 0

### LINKED ISSUE #2728: Bug: set_radio() in Form Helper does not work when radio button value equals "0"
If your radio button has the value "0" the set_radio() function will not output "checked='checked'" if that value had been clicked. 

I believe this is because empty("0") returns true in the line at https://github.com/codeigniter4/CodeIgniter4/blob/54d370f5658583cab58be3c34fb8d46ec45fc85f/system/Helpers/form_helper.php#L894

If you add `<input type="radio" name="myradio" value="0" <?php echo  set_radio('myradio', '0'); ?> />` to a form you can see that it will not stay checked on submission.

musmanikram:
Fix #2728 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -891,7 +891,7 @@ if (! function_exists('set_radio'))
 
 		// Unchecked checkbox and radio inputs are not even submitted by browsers ...
 		$result = '';
-		if (! empty($input = $request->getPost($field)) || ! empty($input = old($field)))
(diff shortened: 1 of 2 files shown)


MGatner:
This looks great! Thanks for the fast response.

PR #2734, opened 2020-03-21
Fixes #2520 : ensure service(router)->controllername() returns fqcn

### LINKED ISSUE #2520: Bug: router service adds backslash to controllername if route is configured
**Describe the bug**
When a route isn't defined, the controllerName() function in service('router') returns an expected string like 'App\Controllers\{controllername}'. But when a route is defined in the routes.php it also returns an extra \ like '\App\Controllers\{controllername}'.

**CodeIgniter 4 version**
CodeIgniter 4 RC 3 zip version
CodeIgniter 4 Dev starter composer
CodeIgniter 4 App starter composer

**Affected module(s)**
Probably the router service

Let the namespace define the route, and return the controller name using
`$router = service("router");
echo $router->controllerName();`
It should return the namespace of the controller with the name.

Now define the route in routes.php, and the same code will return an \ with the namespace of the controller with the controllername.

I expected them to be the same.

samsonasik:
Fixes #2520 : ensure service(router)->controllername() returns fqcn

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -579,7 +579,7 @@ class Router implements RouterInterface
 		// We have to check for a length over 1, since by default it will be '\'
 		if (strpos($this->controller, '\\') === false && strlen($this->collection->getDefaultNamespace()) > 1)
 		{
-			$this->controller = '\\' . str_replace('/', '\\', $this->collection->getDefaultNamespace() . $this->directory . $this->controllerName());
+			$this->controller = '\\' . ltrim(str_replace('/', '\\', $this->collection->getDefaultNamespace() . $this->directory . $this->controllerName()), '\\');
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


samsonasik:
I've added unit test for it.

samsonasik:
I've added `ltrim` check first to ensure clearance `\\` prefix before prepend with the `\\` prefix. I've added data provider in the unit test to test it.

samsonasik:
I've added test for both autoroute and not autoroute to returns always fqcn.

MGatner:
Awesome, I was going to ask about tests when I looked at it earlier! Looks great

MGatner:
I can’t see tests results for this one - maybe an issue with the new GitHub mobile app? I’ll check it out on desktop in a bit.

samsonasik:
rebased

samsonasik:
@MGatner travis build green :tada:

PR #2741, opened 2020-03-21
Fix #2740 regex decimal format rules

### LINKED ISSUE #2740: Bug: Decimal validation fails without leading digit
**Describe the bug**
Numbers like `.25` fail validation because the regex is requiring a leading digit:

`'/^[\-+]?[0-9]+(|\.[0-9]+)$/'`

There should also be another case in the decimal provider at **tests/system/Validation/FormatRulesTest.php**.

This should be a quick fix but I'm so bad at regex and Stack Overflow doesn't agree on how to handle this. :O

**CodeIgniter 4 version**
Develop latest

**Affected module(s)**
Validation

`setRules` with `decimal` on `.11` should be `true`

samsonasik:
Fix #2740 to allow  number `.25` as a valid number.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -152,7 +152,7 @@ class FormatRules
 	 */
 	public function decimal(string $str = null): bool
 	{
-		return (bool) preg_match('/^[\-+]?[0-9]+(|\.[0-9]+)$/', $str);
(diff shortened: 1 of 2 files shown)


samsonasik:
@MGatner travis build green 🎉

Issue #2743, opened 2020-03-22
Bug: Logger context placeholders {file} and {line} are wrong


dafriend:
**Describe the bug**
The values for {file} and {line} should be for the location where the call to log something occurred. 
Instead, the file sometimes evaluates to the Logger class name or the class file. 
The line number is usually, but not always, correct.

**CodeIgniter 4 version**
v 4.0.2 

Given this controller

```
<?php namespace App\Controllers;

class Home extends BaseController
{
	public function index()
	{
		$this->logger->critical("FILE: {file} LINE:{line}");  // line 7
	}

}
```
The log entry produced is

> CRITICAL - 2020-03-22 01:03:47 --> FILE: CodeIgniter\Log\Logger LINE:7

But it should be

> CRITICAL - 2020-03-22 00:19:03 --> FILE: APPPATH/Controllers/Home.php LINE:7

PR #2744, opened 2020-03-22
Logger {file} {line} fix

dafriend:
Fixes issue #2743 
Revise ``CodeIgniter\Log\Logger::determineFile()`` to produce the expected output.
This change required a unit test change.

### DIFF
--- a/system/Log/Logger.php
+++ b/system/Log/Logger.php
@@ -467,37 +467,52 @@ class Logger implements LoggerInterface
 		return strtr($message, $replace);
 	}
 
-	//--------------------------------------------------------------------
 
 	/**
-	 * Determines the current file/line that the log method was called from.
-	 * by analyzing the backtrace.
+	 * Determines the file and line that the logging call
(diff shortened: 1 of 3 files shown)


dafriend:
> @dafriend Just out of curiosity, what scenario would elicit an 'unknown' response? Can we recreate a test case for that?

Really good question. I dimly remember a situation where the result was 'unknown', but I don't remember why. I'm sure I saw the need while working on the solution, but that was last October and the exact reason escapes me now.

A little research indicates the reasons that the 'file' or 'line' items would not be set are edge-case items, for instance, ``debug_backtrace()`` called during a registred shutdown function or xdebug's settings aren't robust enough that the stackframe we're looking is generated. 

We could run ``determineFile()`` without calling any of the "log" functions, but why? All we would be doing is testing that ``in_array()`` and ``isset()`` work which is already proven.

MGatner:
I thought so, just checking!

PR #2748, opened 2020-03-23
clean up dead code detected by  rectorphp/rector

samsonasik:
clean up dead code detected by rector php https://github.com/rectorphp/rector

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -437,7 +437,7 @@ class CLI
 		// Do it once or more, write with empty string gives us a new line
 		for ($i = 0; $i < $num; $i ++)
 		{
-			static::write('');
+			static::write();
 		}
 	}
 
@@ -504,9 +504,7 @@ class CLI
 			$string .= "\033[4m";
 		}
 
-		$string .= $text . "\033[0m";
-
-		return $string;
+		return $string . ($text . "\033[0m");
 	}
 
 	//--------------------------------------------------------------------
@@ -714,11 +712,6 @@ class CLI
 				static::$segments[] = $_SERVER['argv'][$i];
 				continue;
 			}
-
-			// We set $optionsFound here so that we know to
-			// skip the next argument since it's likely the
-			// value belonging to this option.
-			$optionsFound = true;
 
 			$arg   = str_replace('-', '', $_SERVER['argv'][$i]);
 			$value = null;
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -700,14 +700,12 @@ class CLI
 	 */
 	protected static function parseCommandLine()
 	{
-		$optionsFound = false;
-
 		// start picking segments off from #1, ignoring the invoking program
 		for ($i = 1; $i < $_SERVER['argc']; $i ++)
 		{
 			// If there's no '-' at the beginning of the argument
 			// then add it to our segments.
-			if (! $optionsFound && mb_strpos($_SERVER['argv'][$i], '-') === false)
+			if (mb_strpos($_SERVER['argv'][$i], '-') === false)
 			{
 				static::$segments[] = $_SERVER['argv'][$i];
 				continue;
@@ -724,10 +722,6 @@ class CLI
 			}
 
 			static::$options[$arg] = $value;
-
-			// Reset $optionsFound so it can collect segments
-			// past any options.
-			$optionsFound = false;
 		}
 	}
 
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -140,7 +140,6 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
 	 * The database-dependent portion of the prepare statement.
 	 *
 	 * @param string $sql
-	 * @param array  $options Passed to the connection's prepare statement.
 	 *
 	 * @return mixed
 	 */
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -144,7 +144,7 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
 	 *
 	 * @return mixed
 	 */
-	abstract public function _prepare(string $sql, array $options = []);
+	abstract public function _prepare(string $sql);
 
 	//--------------------------------------------------------------------
 
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -140,10 +140,11 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
 	 * The database-dependent portion of the prepare statement.
 	 *
 	 * @param string $sql
+	 * @param array  $options Passed to the connection's prepare statement.
 	 *
 	 * @return mixed
 	 */
-	abstract public function _prepare(string $sql);
+	abstract public function _prepare(string $sql, array $options = []);
 
 	//--------------------------------------------------------------------
 
--- a/system/Database/BaseUtils.php
+++ b/system/Database/BaseUtils.php
@@ -417,7 +417,7 @@ abstract class BaseUtils
 	 *
 	 * @return mixed
 	 */
-	abstract public function _backup(array $prefs = null);
+	public abstract function _backup();
 
 	//--------------------------------------------------------------------
 }
--- a/system/Database/BaseUtils.php
+++ b/system/Database/BaseUtils.php
@@ -413,8 +413,6 @@ abstract class BaseUtils
 	/**
 	 * Platform dependent version of the backup function.
 	 *
-	 * @param array|null $prefs
-	 *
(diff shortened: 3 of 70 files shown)


samsonasik:
@MGatner Yes, I added a [wip] in the title.

samsonasik:
@MGatner I updated some parts that false positive and manually clean up on CLI class. Travis build green :tada: . I removed [wip] in the title.

system/Session/Handlers/RedisHandler.php
```diff
@@ -189,8 +189,6 @@ public function read($sessionID): string
 			{
 				$this->sessionID = $sessionID;
 			}
-
-			$session_data                               = $this->redis->get($this->keyPrefix . $sessionID);
```
michalsn: This should definitely not be removed.

michalsn:
If we remove some parameters from methods, we should also remove the docblock part. It will be a nightmare to spot these later.

system/Session/Handlers/RedisHandler.php
samsonasik: fixed.

samsonasik:
@michalsn I've applied:
-  rolled back false positive in redishandler 
- removed unused param docblocks parts.

samsonasik:
rebased.

system/CLI/CLI.php
```diff
@@ -437,7 +437,7 @@ public static function newLine(int $num = 1)
 		// Do it once or more, write with empty string gives us a new line
 		for ($i = 0; $i < $num; $i ++)
 		{
-			static::write('');
+			static::write();
```
lonnieezell: This needs the empty string left in order to get the new line on the CLI.

samsonasik: the parameter is default to empty string `''`, so it no need a passing an empty string: 

```php
public static function write(string $text = '', string $foreground = null, string $background = null)
```

samsonasik: proof https://3v4l.org/VHqUI

system/Database/BasePreparedQuery.php
```diff
@@ -140,11 +140,10 @@ public function prepare(string $sql, array $options = [], string $queryClass = '
 	 * The database-dependent portion of the prepare statement.
 	 *
 	 * @param string $sql
-	 * @param array  $options Passed to the connection's prepare statement.
 	 *
 	 * @return mixed
 	 */
-	abstract public function _prepare(string $sql, array $options = []);
```
lonnieezell: We should leave this one in for now. I don't think it's used currently, but IIRC will be necessary for other drivers in the future.

system/Database/BaseUtils.php
```diff
@@ -413,11 +413,9 @@ public function backup($params = [])
 	/**
 	 * Platform dependent version of the backup function.
 	 *
-	 * @param array|null $prefs
-	 *
 	 * @return mixed
 	 */
-	abstract public function _backup(array $prefs = null);
+	public abstract function _backup();
```
lonnieezell: While it might be "proper" to call this abstract, I would prefer not to in this case.

system/Database/MySQLi/PreparedQuery.php
```diff
@@ -56,12 +56,10 @@ class PreparedQuery extends BasePreparedQuery implements PreparedQueryInterface
 	 * override this method.
 	 *
 	 * @param string $sql
-	 * @param array  $options Passed to the connection's prepare statement.
-	 *                        Unused in the MySQLi driver.
 	 *
 	 * @return mixed
 	 */
-	public function _prepare(string $sql, array $options = [])
```
lonnieezell: Please leave in for now.

system/Database/MySQLi/Utils.php
```diff
@@ -67,11 +67,9 @@ class Utils extends BaseUtils
 	/**
 	 * Platform dependent version of the backup function.
 	 *
-	 * @param array|null $prefs
-	 *
 	 * @return mixed
 	 */
-	public function _backup(array $prefs = null)
```
lonnieezell: Please leave in.

system/Database/Postgre/PreparedQuery.php
```diff
@@ -74,13 +74,11 @@ class PreparedQuery extends BasePreparedQuery implements PreparedQueryInterface
 	 * override this method.
 	 *
 	 * @param string $sql
-	 * @param array  $options Passed to the connection's prepare statement.
-	 *                        Unused in the MySQLi driver.
 	 *
 	 * @return mixed
 	 * @throws \Exception
 	 */
-	public function _prepare(string $sql, array $options = [])
```
lonnieezell: Since we are leaving it in for the base class we need it here to keep the signature the same.

system/Database/Postgre/Utils.php
```diff
@@ -66,11 +66,9 @@ class Utils extends BaseUtils
 	/**
 	 * Platform dependent version of the backup function.
 	 *
-	 * @param array|null $prefs
-	 *
 	 * @return mixed
 	 */
-	public function _backup(array $prefs = null)
```
lonnieezell: leave in

system/Database/SQLite3/PreparedQuery.php
```diff
@@ -66,12 +66,10 @@ class PreparedQuery extends BasePreparedQuery implements PreparedQueryInterface
 	 * override this method.
 	 *
 	 * @param string $sql
-	 * @param array  $options Passed to the connection's prepare statement.
-	 *                        Unused in the MySQLi driver.
 	 *
 	 * @return mixed
 	 */
-	public function _prepare(string $sql, array $options = [])
```
lonnieezell: leave in

system/Database/SQLite3/Utils.php
```diff
@@ -60,11 +60,9 @@ class Utils extends BaseUtils
 	/**
 	 * Platform dependent version of the backup function.
 	 *
-	 * @param array|null $prefs
-	 *
 	 * @return mixed
 	 */
-	public function _backup(array $prefs = null)
```
lonnieezell: leave in

system/Email/Email.php
```diff
@@ -820,7 +820,6 @@ protected function getMessageID()
 	 */
 	protected function getProtocol()
 	{
-		$this->protocol                                                      = strtolower($this->protocol);
```
lonnieezell: Verify that this is set lowercase elsewhere. Otherwise, we do need to keep this in.

system/Database/BaseUtils.php
samsonasik: @lonnieezell do you want to remove this `_backup()` method ?

samsonasik: I reverted Baseutils class change.

samsonasik:
@lonnieezell  please review again

system/RESTful/ResourceController.php
```diff
@@ -93,7 +93,7 @@ public function index()
 	 *
 	 * @return array	an array
 	 */
-	public function show($id = null)
```
lonnieezell: All methods in ResourceController with the `$id=null` arguments must stay. It's intended to be extended and doesn't work if they're missing.

system/RESTful/ResourcePresenter.php
```diff
@@ -88,18 +88,16 @@ public function index()
 	/**
 	 * Present a view to present a specific resource object
 	 *
-	 * @param  type $id
 	 * @return string
 	 */
-	public function show($id = null)
```
lonnieezell: All methods in ResourcePresenter with the `$id=null` arguments must stay. It's intended to be extended and doesn't work if they're missing.

system/View/Parser.php
```diff
@@ -256,12 +256,12 @@ public function setData(array $data = [], string $context = null): RendererInter
...
 	 *
-	 * @param  string $template
-	 * @param  array  $data
-	 * @param  array  $options  Future options
+	 * @param string $template
+	 * @param array  $data
+	 *
 	 * @return string
 	 */
-	protected function parse(string $template, array $data = [], array $options = null): string
```
lonnieezell: Please leave the $options array in.

system/CLI/CLI.php
lonnieezell: Oh, you're right. Ignore :)

samsonasik:
@lonnieezell I've reverted removal id parameter in resource controller and resource presenter, also revert the removal of options parameter at view parser.

lonnieezell:
Awesome. Thanks so much! Then I believe we're good to go as soon as Travis comes back with the test results again.

samsonasik:
@lonnieezell travis build green :tada:

MGatner:
I still approve, but no way I'm merging that ;)

lonnieezell:
@MGatner YOLO, right? lol

PR #2754, opened 2020-03-25
replace for to foreach detected by rectorphp/rector

samsonasik:
remove count for loop with for when possible to use foreach with command:

```bash
rector process --set code-quality --rule "Rector\CodeQuality\Rector\For_\ForToForeachRector" system
```

**Checklist:**
- [x] Securely signed commits


samsonasik:
needs many manual checks.

PR #2756, opened 2020-03-26
add "prefer-stable" => true to admin/starter/builds when ./builds development called

samsonasik:
@MGatner This is to handle composer update try to catch all "dev" version of dependencies in `vendor` directory when running composer update after : 

```bash
php builds development
composer update
```

With make `"prefer-stable" => true`, it will only get dev version when required, eg: in : 

```
"codeigniter4/codeigniter4": "dev-develop"
```

Other packages will try get stable version. The result of composer.json will have: 

```json
    "minimum-stability": "dev",
    "prefer-stable": true,
```

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/admin/starter/builds
+++ b/admin/starter/builds
@@ -41,6 +41,7 @@ if (is_file($file))
 			{
 				// Set 'minimum-stability'
 				$array['minimum-stability'] = 'dev';
+				$array['prefer-stable']     = true;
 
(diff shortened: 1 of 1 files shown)


MGatner:
Yes I think this is a good change - I do this myself manually in most projects.

PR #2759, opened 2020-03-26
Parser use LoggerInterface

dafriend:
Change type hint for ``$logger`` argument of ``Parser::__construct()`` to use `Psr\Log\LoggerInterface` instead of `CodeIgniter\Log\Logger` which will allow alternative PSR-3 loggers to be used in ``Parser``.

This brings ``Parser`` in line with the other core classes that have a "logger" dependency.

### DIFF
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -39,7 +39,7 @@
 
 namespace CodeIgniter\View;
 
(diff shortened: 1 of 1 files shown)


MGatner:
Good one!

PR #2768, opened 2020-03-28
Added "Shared Hosting with cPanel" to "Running Your App" [ci skip]

jreklund:
**Description**
I wrote instructions to use CodeIgniter on **Shared Hosting with cPanel**, but it can also be applied on Shared Hosting in general.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/installation/running.rst
+++ b/user_guide_src/source/installation/running.rst
@@ -208,3 +208,56 @@ Once set up, you can then launch your webapp inside a VM, with the command::
 Your webapp will be accessible at ``http://localhost:8080``, with the code coverage
 report for your build at ``http://localhost:8081`` and the user guide for
 it at ``http://localhost:8082``.
+
+Shared Hosting with cPanel
+=================================================
+
+cPanel is a Linux-based web hosting control panel that provides a powerful set of automation tools in an easy and
+straightforward interface, allowing a simplified process for hosting websites.
+
+The contents of CodeIgniter's ``public`` folder should be placed in your domain's **Document Root**, in cPanel's case it's
+called ``public_html`` for your **Main Domain**. Here is an example of a directory structure that are created by cPanel:
+
+::
+
+    /bin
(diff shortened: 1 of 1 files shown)


lonnieezell:
I'm also torn on this one. I do think many people have an issue with the Primary Domain on CPanel, but there are other ways that work that will keep the files together, like mentioned in [this article](https://www.rosehosting.com/blog/how-to-install-laravel-on-a-cpanel-server/). There's probably a few thousand other articles out there like it since all of the other frameworks use a similar setup. 

I think including a basic "server setup" page might be preferable that gives bare bones setup instructions for Apache, NGinx and on a tricky hosting platform like CPanel. But I'd like use to find the simplest way to host on a primary domain on CPanel that we can.

jreklund:
Hi, thanks for the feedback. At my Shared hosting I can't connect to the server with SSH, and haven't figured out any other ways to change it. It's a bummer that you can't change the document root from the UI like you can with addon (and sub-) domains.

@MGatner What if you want to use your primary domain you bought? Not everyone have one domain to waste. (Don't want to delete my primary domain to test it out myself, as it will wipe my current website.)

I wrote this part as it have come up a few times in the forum and people don't know what to do. I have seen people make .htaccess and redirect into /public/ folder just so their website worked. So that's why this where born. This aren't that hard to figure out if you know what you are doing, but as this is hard enough, I think the SSH part will be scary as hell.

We had a similar topic (if I remember correctly) in CI3 manual, on how to move index.php into public_html instead. But maybe this suites better for the "Learn" portal on the new website?

lonnieezell:
We [still have docs](https://codeigniter.com/user_guide/general/managing_apps.html#renaming-or-relocating-the-application-directory) on moving the main index.php file. But you bring up a good point - this would be a great thing for the new Learn section on the upcoming website. Let's leave if for that.

PR #2770, opened 2020-03-28
fix json function clean useless variables

mostafakhudair:
some fixes


MGatner:
`$this->responsd`

How did this even exist this long?? Is this controller function just never called? Baffling. Thanks for the fix though!

PR #2775, opened 2020-03-29
Update on "Build Your First Application"

jreklund:
**Description**
- Reference the view() function (for rendering templates) as a function instead of method.
- Removed unused code (that got copied from CI3)
- Added references to other part of the userguide
- Code examples are up to CI4 coding standards
- Using esc() to prevent XSS, and give hints about it.
- Use view() method consistent in the tutorial. First pages had showme, and later on used view() for News. Give user a warning about the different view() method and function.
- Improved rendering on some elements
- Always reference a method by index() and not index. 

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/tutorial/index.rst
+++ b/user_guide_src/source/tutorial/index.rst
@@ -59,7 +59,7 @@ From your command line type the following:
 
 ::
 
-    ➜ composer create-project codeigniter4/appstarter ci-blog -s rc
+    composer create-project codeigniter4/appstarter ci-blog -s rc
 
 This creates a new folder, ci-blog, which contains your application code, with
 CodeIgniter installed in the vendor folder.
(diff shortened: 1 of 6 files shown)


user_guide_src/source/tutorial/index.rst
```diff
@@ -59,7 +59,7 @@ From your command line type the following:
 
 ::
 
-    ➜ composer create-project codeigniter4/appstarter ci-blog -s rc
+    composer create-project codeigniter4/appstarter ci-blog -s rc
```
MGatner: Do we want to drop `-s rc` now that we're a legit stable release? I think so, and then we have the `./builds` commands for switching to source as needed.

jreklund: Oh, so that's what it's for. Will remove it. :-)

lonnieezell:
Looks great to me! I'll give @MGatner  a chance to resolve his change request.

PR #2778, opened 2020-03-30
Rework get_filenames

MGatner:
**Description**
https://forum.codeigniter.com/thread-75898.html

`get_filenames()` current has two different return types, contingent on its second parameter. The first (default) is a flat list of all files in all subdirectories:
> document.txt
> cat.jpg
> rainbow.exe

And the second is the same files by their absolute paths:
> /var/www/html/mysite/document.txt
> /var/www/html/mysite/assets/cat.jpg
> /var/www/html/mysite/downloads/cached/rainbow.exe

I am of the opinion that the first is pretty much useless, and that a relative listing of file paths would be much more helpful. See the forum discussion for more details.

This PR changes the function significantly to allow for the new return types: relative to source, and full path. It also no longer uses `realpath()` since that makes the function incompatible with virtual paths (e.g. `vfsStream` for testing) - if this is an issue an alternative would be to apply `realpath()` to the final file name before adding it to the return array.

Tests have been updated accordingly.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Helpers/filesystem_helper.php
+++ b/system/Helpers/filesystem_helper.php
@@ -209,12 +209,12 @@ if (! function_exists('get_filenames'))
 	 * Reads the specified directory and builds an array containing the filenames.
 	 * Any sub-folders contained within the specified path are read as well.
 	 *
-	 * @param string $source_dir   Path to source
-	 * @param mixed  $include_path Whether to include the path as part of the filename; empty for no path, 'relative' for a relative path, not empty for full path
+	 * @param string       $source_dir   Path to source
+	 * @param boolean|null $include_path Whether to include the path as part of the filename; false for no path, null for a relative path, true for full path
 	 *
 	 * @return array
 	 */
-	function get_filenames(string $source_dir, $include_path = ''): array
+	function get_filenames(string $source_dir, ?bool $include_path = false): array
 	{
 		$files = [];
 
@@ -228,11 +228,11 @@ if (! function_exists('get_filenames'))
 					RecursiveIteratorIterator::SELF_FIRST
 				) as $name => $object)
 			{
-				if (empty($include_path))
+				if ($include_path === false)
 				{
 					$files[] = pathinfo($name, PATHINFO_BASENAME);
 				}
-				elseif ($include_path === 'relative')
+				elseif (is_null($include_path))
 				{
 					$files[] = str_replace($source_dir, '', $name);
 				}
--- a/system/Helpers/filesystem_helper.php
+++ b/system/Helpers/filesystem_helper.php
@@ -209,12 +209,12 @@ if (! function_exists('get_filenames'))
 	 * Reads the specified directory and builds an array containing the filenames.
 	 * Any sub-folders contained within the specified path are read as well.
 	 *
-	 * @param string  $source_dir   Path to source
-	 * @param boolean $include_path Whether to include the path as part of the filename
+	 * @param string $source_dir   Path to source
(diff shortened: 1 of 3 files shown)


MGatner:
UPDATE: After thinking about the `realpath()` change I decided to have it use the function whenever it resolves correctly. This way it should have cleaner output but still work with virtual paths and streams.

lonnieezell:
One concern is that this would be a BC breaking change. Which, if we kept with semantic versioning, means a 5.0 release, I think. Way too soon for that. :)

The only way I can think to do this in a non BC-breaking way is to add another method that is a convenience wrapper for the current one.

MGatner:
My first attempt as actually to vary the second parameter to allow all three return types, which I can still look at doing. Can you specify what you would consider BC? Is it just the change from flat file to relative path without `$include_path`?

I think the current implementation is actually a problem with its use of `realpath()`, and while that will result in a change (especially for test cases) I'd consider it a bug fix. Virtual paths current resolve to "false" and then get suffixed with `/` creating paths like "/foo" (see the test case) which will not resolve to their corresponding virtual path.

The only other change is that I'm no longer ignoring `.xxxx` files. This was partly out of the opinion that those files are useful and `SKIP_DOTS` handles "." and ".." automatically. The User Guide doesn't mention hidden file behavior but I could revert that (or add another parameter) as well.

MGatner:
@lonnieezell See above but one more question... is differently-ordered output going to be a BC? `RecursiveDirectoryIterator` uses the OS's ordering, so I am sorting results to make them consistent across uses but output will differ from a recursive `readdir()`.

lonnieezell:
@MGatner I have no issues with ordering. In your PR here, though, you're removing a parameter and functionality. Granted, it might be functionality we no longer desire, but it's still removing functionality. 

We could deprecate that functionality and take it out in the future.

MGatner:
Actually, this thing is a mess. I think we need to start over, maybe begin with a clearer definition in the User Guide. Currently it states:

> Takes a server path as input and returns an array containing the names of all files contained within it. The file path can optionally be added to the file names by setting the second parameter to TRUE.

Unclear whether it is intended to be recursive. The docblock states:

> Reads the specified directory and builds an array containing the filenames. Any sub-folders contained within the specified path are read as well.

Seems likely that it is supposed to be recursive? Definitely the implementation is:
```
...
if (is_dir($source_dir . $file) && $file[0] !== '.')
{
	get_filenames($source_dir . $file . DIRECTORY_SEPARATOR, $include_path, true);
```

But then look at the current test cases, this is not right. The test structure paths are:
```
vfs://root/.hidden
vfs://root/AnEmptyFolder
vfs://root/boo
vfs://root/boo/far
vfs://root/boo/faz
vfs://root/foo
vfs://root/foo/bar
vfs://root/foo/baz
vfs://root/simpleFile
```

And yet the expected result without `$include_path` is:
```
'foo',
'boo',
'AnEmptyFolder',
'simpleFile',
```

How does that work? Friendly note in the test:

> Not sure the directory names should actually show up here but this matches v3.x results.

Is this a bugged implementation held over from CI3? If so I'm fine with a BC to clean this up.

MGatner:
> you're removing a parameter 

The parameter isn't so much an issue, as the docblock states:

`@param boolean $recursion    Internal variable to determine recursion status - do not use in calls`

But I get the call on functionality. Process my above comment while I push my changes to add back the "flat file" option.

MGatner:
@lonnieezell re-review on this updated, non-breaking version. If this still is not acceptable I think my approach will be to fix the `realpath()` bug and then add a separate function for relative listings.

Issue #2780, opened 2020-03-30
Database model error when limiting delete()


pietrobe03:
When trying to delete data from the database with a limit using a model, the LIMIT block is appended twice by the query builder.

This example code
```
$gameCardStackModel->where("gameId", $game["id"])
->orderBy("id", "asc")
->limit(8)
->delete();
```
results in this Query
```
DELETE FROM  `GameCardStacks`\nWHERE `gameId` = '1' LIMIT 8 LIMIT 8
```
Stacktrace:
[stacktrace.json.txt](https://github.com/codeigniter4/CodeIgniter4/files/4404813/stacktrace.json.txt)

samsonasik:
should can be closed in favor of https://github.com/codeigniter4/CodeIgniter4/pull/2781

PR #2781, opened 2020-03-30
#2780 - LIMIT.

nowackipawel:
#2780 
_limit() is called inside delete(), no need to call it again in _delete();

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1706,9 +1706,9 @@ class BaseBuilder
 	 *
 	 * @return string
 	 */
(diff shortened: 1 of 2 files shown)


nowackipawel:
@lonnieezell : could you please rerun travis again? It failed because of: "The command "composer install --prefer-source" failed and exited with 1 during ."

MGatner:
@nowackipawel I'll admit I don't totally understand this. Could you perhaps supply a test case for the scenario this fixes? It would be good to have anyways.

nowackipawel:
Ofc
https://github.com/codeigniter4/CodeIgniter4/issues/2780

PR #2782, opened 2020-03-30
make Model::paginate() use default perPage from Config\Pager->perPage if $perPage parameter not passed

samsonasik:
Currently, the `$perPage` is default to `20` in `Model::paginate()` function. I think it should read from `config('Pager')->perPage` instead. I apply make default to null, and check if it is not passed, it will use the default `config('Pager')->perPage`

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -500,7 +500,7 @@ class Services extends BaseService
 
 		if (empty($config))
 		{
-			$config = new \Config\Pager();
+			$config = config('Pager');
 		}
 
 		if (! $view instanceof RendererInterface)
--- a/system/Model.php
+++ b/system/Model.php
@@ -1123,17 +1123,16 @@ class Model
 	 */
 	public function paginate(int $perPage = null, string $group = 'default', int $page = 0)
 	{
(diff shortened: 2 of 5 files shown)


samsonasik:
I've added unit test for use case change config('Pager')->perPage

samsonasik:
travis build green 🎉

samsonasik:
I've added check in `Pager` class to read `$perPage` from `Config('Pager')->perPage` to handle it when used in outside Model/Db related pagination.

samsonasik:
I moved the handle per page null check to Pager class, use `$this->pager->getPerPage()` pulled from Pager instance to get final perPage value that will be used to make a limit query in the `Model::paginate()`

samsonasik:
I updated `CodeIgniter\Config\Services::pager()` to use `config('Pager')` as well if no $config passed (null), instead of manual "new Config\Pager()" like in other services which uses `config()` function. It also make it testable in paginate in model test.

samsonasik:
travis build green 🎉

MGatner:
Definitely a good change, and a solid PR. Thanks as always @samsonasik

PR #2785, opened 2020-04-01
test for CodeIgniter\Config\Services::email()

samsonasik:
Test for `CodeIgniter\Config\Services::email()`. Make `CodeIgniter\Config\Services` class 100% tested.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/tests/system/Config/ServicesTest.php
+++ b/tests/system/Config/ServicesTest.php
@@ -55,6 +55,18 @@ class ServicesTest extends \CodeIgniter\Test\CIUnitTestCase
 		$this->assertInstanceOf(\CodeIgniter\HTTP\CURLRequest::class, $actual);
 	}
 
+	public function testNewEmail()
+	{
+		$actual = Services::email();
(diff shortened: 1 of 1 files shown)


MGatner:
You probably thought this would be a quick PR, but I have a request for you to consider. It’s been on my list for a while to write a `MockEmail` class. It would essentially be an extension of the Email class but instead of sending it would store an array of the parameters that could be retrieved. Or something along those lines. Would you be interested in taking this on? Could make this part of a larger update to Email before the class gets rewritten.

samsonasik:
Do you think the class will be `CodeIgniter\Test\Mock\MockEmail` ? mock 28 public methods ? or do you have list of methods that needs to be mocked?

MGatner:
Yes that’s the class. I haven’t looked through the methods but my guess is only the actual send would need to be mocked. I can look when I’m not on mobile and specify, but I also trust your judgement.

samsonasik:
I created separate PR #2788 for new MockEmail cass.

PR #2788, opened 2020-04-02
add MockEmail class

samsonasik:
Based on @MGatner comment at https://github.com/codeigniter4/CodeIgniter4/pull/2785#issuecomment-607180488

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Test/Mock/MockEmail.php
+++ b/system/Test/Mock/MockEmail.php
@@ -6,7 +6,10 @@ class MockEmail extends Email
 {
 	public function send($autoClear = true)
 	{
-		$this->clear();
+		if ($autoClear)
+		{
+			$this->clear();
+		}
 
 		return true;
 	}
--- a/system/Test/Mock/MockEmail.php
+++ b/system/Test/Mock/MockEmail.php
@@ -13,13 +13,12 @@ class MockEmail extends Email
 
 	public function send($autoClear = true)
 	{
-		$this->archive = get_object_vars($this);
-
 		if ($autoClear)
 		{
 			$this->clear();
 		}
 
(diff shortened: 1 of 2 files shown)


samsonasik:
I've added unit test for `$autoClear` parameter in `send()` method.

lonnieezell:
This looks good. We do need a way to incorporate testing failures, though, too. Not sure what the preferred way here would be, whether it's another method that specifies the result, or another parameter I'm not too concerned. 

@MGatner thoughts?

MGatner:
Yes I think this is a great start. I was thinking of something like this:
```
	/**
	 * Record of mock emails sent.
	 * @var string
	 */
	public $archive;
...
	public function send($autoClear = true)
	{
...
		$this->archive[] = get_object_vars($this);

		return true;
	}
```

That way a test could check `$mockEmail->archive` to see the actual emails that were "sent". I don't think that code will work because of the scope but you get the idea.

samsonasik:
@MGatner I've added `$archive` property for save properties on `send()` method called.

samsonasik:
I realized that `clear()` methods clear the properties, so, I placed the `archive` saving after `autoClear` check.

MGatner:
Thanks @samsonasik this is precisely what I'd hoped for! Glad to see it wasn't too complicated either.

MGatner:
> I realized that `clear()` methods clear the properties, so, I placed the `archive` saving after `autoClear` check.

@samsonasik I'm working on some tweaks to Email testing... Isn't the precisely what we *don't* want? Setting `$this->archive = get_object_vars($this);` after `clear()` means the values will always be blank. Should be before like you originally had it, right?

PR #2790, opened 2020-04-03
add ability to replace {locale} to request->getLocale() in form_open('action')

samsonasik:
So no need to manually add locale from variable when call `form_open()` in view.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated


samsonasik:
I updated the userguide for its addition ability.

samsonasik:
merge-able?

PR #2793, opened 2020-04-03
test for API\ResponseTrait::format() with format is not json or xml

samsonasik:
By this, `CodeIgniter\API\ResponseTrait` will have 100% test coverage.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/tests/system/API/ResponseTraitTest.php
+++ b/tests/system/API/ResponseTraitTest.php
@@ -452,4 +452,44 @@ EOH;
 		$this->assertEquals($expected, $this->response->getBody());
 	}
 
+	public function testFormatByRequestNegotiateIfFormatIsNotJsonOrXML()
+	{
+		$config = [
+			'baseURL'          => 'http://example.com',
+			'uriProtocol'      => 'REQUEST_URI',
+			'defaultLocale'    => 'en',
+			'negotiateLocale'  => false,
+			'supportedLocales' => ['en'],
+			'CSPEnabled'       => false,
+			'cookiePrefix'     => '',
+			'cookieDomain'     => '',
+			'cookiePath'       => '/',
+			'cookieSecure'     => false,
+			'cookieHTTPOnly'   => false,
+			'proxyIPs'         => [],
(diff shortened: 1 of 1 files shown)


MGatner:
Using `MockIncomingRequest` means that `header('accept')` won't ever be set, so that means the format will always be set to the first formatter available (per https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/Negotiate.php#L213). I expect RESTful features like this to have a major pass soon, so my preference would be to change the test to check that it is matching the first formatter, instead of explicitly assuming `application/json`. Does that make sense, or am I being crazy?

samsonasik:
the application/json is based on detection of data, if the data is a "xml like", it will marked as application/xml. See that I changed format to "txt" in controller __construct that it means formatting will find best match of negotiation with check supported format

samsonasik:
this test the following lines https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/API/ResponseTrait.php#L376-L379

MGatner:
I understand the test, my concern is that the function call in question:
```
$this->request->negotiate('media', $config->supportedResponseFormats, false);
```

... ends up calling `Negotiator::media()` which uses `getBestMatch()`:
```
return $this->getBestMatch($supported, $this->request->getHeaderLine('accept'), true, $strictMatch);
```

And in our test scenario `getHeaderLine('accept')` will always be empty. Without a header `getBestMatch()` simply falls back to the first format on the list:
```
return $strictMatch ? '' : $supported[0];
```
... so instead of checking for 'application/json; charset=UTF-8' should we check against something like `config('App')->supportedResponseFormats[0]`?

samsonasik:
Ok, I updated with check against : `config('Format')->supportedResponseFormats[0]`

samsonasik:
merge-able?

lonnieezell:
@MGatner?

MGatner:
Sorry! My email is a total train wreck. Thanks for the mention, they help!

PR #2797, opened 2020-04-04
[ci skip] Fix download badge total shows

samsonasik:
Previously, it only shows "4.4k", while it already "24.73k" downloads, seems because of path url location to `codeigniter4/codeigniter4` instead of `codeigniter4/framework`. 

I updated the badge to use `poser.pugx.org` as the `img.shields.io` only shows 0 downloads.

**Checklist:**
- [x] Securely signed commits


samsonasik:
**Before**

Screen Shot 2020-04-04 at 8 00 30 AM

**After**

Screen Shot 2020-04-04 at 8 01 37 AM

TomasVotruba:
Nice job :clap:

Issue #2194, opened 2019-09-06
spark route issue


hatsat32:
**description**

Normally if codeigniter 4 can't find a apropriate route, it shows 404 page. But when i started server with `php spark serve` command and enter a url that includes dot `.` and does not exist in routes, it show default `welcome_message` page.

For example
* `http://localhost:8080/page/doesnotexist` returns ci4's 404 page as usual
* `http://localhost:8080/page/does.notexist` returns welcome_message page
* `http://localhost:8080/notexist` return ci4's 404 page as as usual
* `http://localhost:8080/notexist.` return welcome_message page

when i use php's default server as fallow
```
cd public/
php -S localhost:8080
```
Results:
* `http://localhost:8080/page/doesnotexist` returns ci4's 404 page as usual
* `http://localhost:8080/page/does.notexist` returns php's not found page
* `http://localhost:8080/notexist` returns ci4's 404 page as usual
* `http://localhost:8080/notexist.` returns php's not found page

Probably this issue about CodeIgniter 4 CLI Tool but i I couldn't find where exactly it is :(

CodeIgniter 4 Version 4.0.0-rc.1
OS: Linux ubuntu 19.04
PHP version 7.2.19-0ubuntu0.19.04.2

jim-parry:
Cannot reproduce this.
All the examples you show, with php spark serve, return a CI 404

hatsat32:
sorry but I still get the same orror. I pulled ci4 github repo and tested it again.

* `http://localhost:8080/pagenotexist` CI4's 404 not found page
* `http://localhost:8080/pagenot.exist` default page (welcome_message.php)

I tested in different platforms again and get same error :(

* ubuntu 18.04.02. My remote server. with this command `php spark serve --host 0.0.0.0`
	* PHP 7.2.19-0ubuntu0.18.04.2
	* pulled ci4 github repo
* ubuntu 19.04. command: `php spark serve`
	* PHP 7.2.19-0ubuntu0.19.04.2
	* pulled ci4 github repo
* kali linux 2019.3 command: `php spark serve`
	* PHP 7.3.8-1
	* pulled ci4 github repo

I dont think this error about my environment or system. I tested 3 different systems. If you relly dont get this error, maybe more people should test it. Than we can learn what is going on here.

lonnieezell:
I can recreate it on Mac running php 7.2

lonnieezell:
[...]

PR #2810, opened 2020-04-07
Ensure $_SERVER['SCRIPT_NAME'] ends with PHP

willnode:
**Description**
When doing spark serve, `$_SERVER['SCRIPT_NAME']` has a bug that leads to URI like `/uploads/avatar/1.jpg` being treated as `/`. This solves the problem by making sure that `$_SERVER['SCRIPT_NAME']` is indeed a PHP file, otherwise don't process it futher.
This solves issue #2194 

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -721,7 +721,7 @@ class IncomingRequest extends Request
 		$query = $parts['query'] ?? '';
 		$uri   = $parts['path'] ?? '';
 
-		if (isset($_SERVER['SCRIPT_NAME'][0]) AND pathinfo($_SERVER['SCRIPT_NAME'], PATHINFO_EXTENSION) === 'php')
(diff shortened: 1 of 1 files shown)


system/HTTP/IncomingRequest.php
```diff
@@ -721,7 +721,7 @@ protected function parseRequestURI(): string
 		$query = $parts['query'] ?? '';
 		$uri   = $parts['path'] ?? '';
 
-		if (isset($_SERVER['SCRIPT_NAME'][0]))
+		if (isset($_SERVER['SCRIPT_NAME'][0]) AND pathinfo($_SERVER['SCRIPT_NAME'], PATHINFO_EXTENSION) === 'php')
```
samsonasik, line 724: use `&&` instead of 'AND'

```suggestion
              if (isset($_SERVER['SCRIPT_NAME'][0]) && pathinfo($_SERVER['SCRIPT_NAME'], PATHINFO_EXTENSION) === 'php')
```

willnode: done

lonnieezell:
Good catch. Thanks for this!

PR #2811, opened 2020-04-07
add Parser helper

mostafakhudair:
add new helper to return instance of CodeIgniter\View\Parser


MGatner:
A few tests would also be superb. :)

lonnieezell:
@mostafakhudair Are you going to be able to make these changes?

PR #2814, opened 2020-04-09
ignore coverage on !CI_DEBUG

samsonasik:
Ignore coverage when CI_DEBUG is false as never hit in tests.

**Checklist:**
- [x] Securely signed commits


samsonasik:
travis build green 🎉

PR #2815, opened 2020-04-09
Cleanup Model

mostafakhudair:
remove useless variables and do some cleanup, minimize lines.

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -675,7 +675,7 @@ class Model
 		// Validate data before saving.
 		if ($this->skipValidation === false)
 		{
-			if ($this->cleanRules(false)->validate($data) === false)
+			if ($this->cleanRules()->validate($data) === false)
 			{
 				return false;
 			}
@@ -741,7 +741,7 @@ class Model
 		{
 			foreach ($set as $row)
 			{
-				if ($this->cleanRules(false)->validate($row) === false)
+				if ($this->cleanRules()->validate($row) === false)
 				{
 					return false;
 				}
@@ -809,7 +809,7 @@ class Model
 		// Validate data before saving.
 		if ($this->skipValidation === false)
 		{
-			if ($this->cleanRules()->validate($data) === false)
+			if ($this->cleanRules(true)->validate($data) === false)
 			{
 				return false;
 			}
@@ -862,7 +862,7 @@ class Model
 		{
 			foreach ($set as $row)
 			{
-				if ($this->cleanRules()->validate($row) === false)
+				if ($this->cleanRules(true)->validate($row) === false)
 				{
 					return false;
 				}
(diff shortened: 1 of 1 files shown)


system/Model.php
```diff
@@ -570,7 +554,7 @@ public function save($data): bool
 	 * @return array
 	 * @throws \ReflectionException
 	 */
-	public static function classToArray($data, $primaryKey = null, string $dateFormat = 'datetime', bool $onlyChanged = true): array
+	protected static function classToArray($data, $primaryKey = null, string $dateFormat = 'datetime', bool $onlyChanged = true): array
```
samsonasik, line 557: this is bc break

mostafakhudair: my fault, revert done thanks

system/Model.php
```diff
@@ -1374,13 +1346,13 @@ public function setValidationMessage(string $field, array $fieldMessages)
 	 * Should validation rules be removed before saving?
 	 * Most handy when doing updates.
 	 *
-	 * @param boolean $choice
+	 * @param boolean $clean
 	 *
 	 * @return $this
 	 */
-	public function cleanRules(bool $choice = false)
+	public function cleanRules(bool $clean = true)
```
samsonasik, line 1353: change default value of true/false is bc break, please revert.

mostafakhudair: please check
insert and insertBatch methods $this->cleanRules(false), 
update and updateBatch methods $this->cleanRules()
replace method $this->cleanRules()

I think it doesn't bc break

methods its

samsonasik: it is public method, and public method can be used in extended class. If the class that extends Model call cleanRules without parameter, it will make different result.

mostafakhudair: I thought it will return the default propriety protected $cleanValidationRules = true;

samsonasik: no, with default value in method parameter like this, it will replace it with assignment. see: https://3v4l.org/0hmeY for the proof

mostafakhudair: please check this https://3v4l.org/qL7Ml and then i will revert

samsonasik: The problem is the existing user that already call the method **without** passing any parameter in his/her application in the extended class(es), and he/she run composer update, he/she will got different result as it have different default value.

mostafakhudair: reverted and thank you for your patience

system/Model.php
```diff
@@ -1418,9 +1390,7 @@ public function validate($data): bool
 			$rules = $this->validation->loadRuleGroup($rules);
 		}
 
-		$rules = $this->cleanValidationRules
-			? $this->cleanValidationRules($rules, $data)
-			: $rules;
+		$rules = $this->cleanValidationRules ? $this->cleanValidationRules($rules, $data): $rules;
```
lonnieezell, line 1393: I find these easier to read when on multiple lines. Curious - why change?

Issue #2812, opened 2020-04-09
Bug:  &quot;Type is not supported&quot; in Postgresql POST restful


samsonasik:
I tried to create a restful api with POST data for create products and call:

```bash
 curl -i -X POST -d "product_code=A1114&product_name=Corn" http://ci4.local/products
```

and got the following errors:

```bash
HTTP/1.1 500 Internal Server Error
Date: Thu, 09 Apr 2020 01:13:21 GMT
Server: Apache/2.4.41 (Unix) PHP/7.4.4
X-Powered-By: PHP/7.4.4
Cache-control: no-store, max-age=0, no-cache
Content-Length: 2568
Connection: close
Content-Type: application/json; charset=UTF-8

<br />
<b>Fatal error</b>:  Uncaught CodeIgniter\Format\Exceptions\FormatException: Failed to parse json string, error: &quot;Type is not supported&quot;. in /Users/samsonasik/www/ci4/vendor/codeigniter4/codeigniter4/system/Format/Exceptions/FormatException.php:9
Stack trace:
#0 /Users/samsonasik/www/ci4/vendor/codeigniter4/codeigniter4/system/Format/JSONFormatter.php(67): CodeIgniter\Format\Exceptions\FormatException::forInvalidJSON('Type is not sup...')
#1 /Users/samsonasik/www/ci4/vendor/codeigniter4/codeigniter4/system/API/ResponseTrait.php(401): CodeIgniter\Format\JSONFormatter-&gt;format(Array)
#2 /Users/samsonasik/www/ci4/vendor/codeigniter4/codeigniter4/system/API/ResponseTrait.php(131): CodeIgniter\Debug\Exceptions-&gt;format(Array)
#3 /Users/samsonasik/www/ci4/vendor/codeigniter4/codeigniter4/system/Debug/Exceptions.php(165): CodeIgniter\Debug\Exceptions-&gt;respond(Array, 500)
#4 [internal function]: CodeIgniter\Debug\Exceptions-&gt;exceptionHandler(Object(ErrorException))
#5 {main}
  thrown in <b>/Users/samsonasik/www/ci4/vendor/codeigniter4/codeigniter4/system/Format/Exceptions/FormatException.php</b> on line <b>9</b><br />
{
    "title": "ErrorException",
    "type": "ErrorException",
    "code": 500,
[...]

Issue #2817, opened 2020-04-09
Bug: Validation placeholder not being replaced


erikkloeze:
**Describe the bug**
When using validation placeholders in a model and validating data, the placeholder is not being replaced with the actual data. The validation therefore returns false.

**CodeIgniter 4 version**
4.0.2

**Expected behavior, and steps to reproduce if appropriate**
Using the [myth-auth](https://github.com/lonnieezell/myth-auth) package and the GroupModel. Validation rules are defined in the model as:
```
protected $validationRules = [
	'name' => 'required|max_length[255]|is_unique[auth_groups.name,name,{name}]',
	'description' => 'max_length[255]',
];
```

When trying to update a group in the controller using an already existing `name` in the database:
```
$groupModel = new GroupModel();

if (! $this->validate($rules = $groupModel->getValidationRules())) {
	return redirect()->back()->withInput()->with('errors', $groupModel->errors());
}
```
The validation returns false. Looking at the DB (MySQL) query ran it can be seen why:
```SELECT 1 FROM `auth_groups` WHERE `name` = 'admin' AND `name` != '{name}' LIMIT 1```

A manual replacement of `{name}` with the actual value resolves the issue.

**Comment by samsonasik** [comment]
i guess you are using postgresql, seems similar issue with my comment at https://github.com/codeigniter4/CodeIgniter4/issues/2812#issuecomment-611283231

**Comment by erikkloeze** [comment]
After digging through the code I might have found the solution. This might also solve #2812. 

In the `Model` class the `validate()` function includes a `fillPlaceholders()` before it sets the rules and runs the validation:
https://github.com/codeigniter4/CodeIgniter4/blob/895f1e67eb895eacff204e5baba73e5e1b2629af/system/Model.php#L1434-L1439

In the `Controller` class this call to `fillPlaceholders()` is not present:
https://github.com/codeigniter4/CodeIgniter4/blob/895f1e67eb895eacff204e5baba73e5e1b2629af/system/Controller.php#L216-L219

When the `fillPlaceholders()` is added to the `Controller` class and called before the validator runs, the issue is solved:

```
// Replace any placeholders (i.e. {id}) in the rules with
// the value found in $data, if exists.
$rules = $this->fillPlaceholders($rules, $this->request->getVar());

$success = $this->validator
	->withRequest($this->request)
	->setRules($rules, $messages)
	->run();
```

PR #2819, opened 2020-04-10
Validation placeholders in Controller not replaced

erikkloeze:
This pull request should fix issues #2812 and #2817.

**Description**
Duplicated the `fillPlaceholders()` function from Model.php to Controller.php. Called the function before the validator runs.

**Checklist:**
TODO

### DIFF
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -212,11 +212,71 @@ class Controller
 
 			$rules = $validation->$rules;
 		}
+		
+        // Replace any placeholders (i.e. {id}) in the rules with
+        // the value found in $data, if exists.
+       	$rules = $this->fillPlaceholders($rules, $this->request->getVar());
 
 		return $this->validator
 			->withRequest($this->request)
 			->setRules($rules, $messages)
 			->run();
+	}
+	
+	/**
+	 * Replace any placeholders within the rules with the values that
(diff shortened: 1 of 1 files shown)


samsonasik:
seems requires 2 things:

1. cs check - fix, you can enable cs by run:

```
./admin/setup.sh
```

and try do some changes in the file, and commit. First commit should fail as it act to fix the cs. Retry commit will apply the commit.

2. New tests if possible.

samsonasik:
also, I think placeholder replacement check rules should not in controller as validation may be in model.

njsx:
This is not universal a solve! 

This fix will be work ONLY with data from REQUEST (GET or POST). What about if I want to check data from array in my Controllers's method?

```
$arr['id'] = $this->request->getVar('id');
$arr['login'] = $this->request->getVar('login');

if ($this->validation->run($arr, 'myRuleSet') ) { 
.....
```

In my app\Config\Validation.php:

```
public $myRuleSet = [
		'login' => "is_unique[admins.login,id,{id}]",
	];
```
For is_unique placeholder {id} will never be recognized!

samsonasik:
can be closed in favor of https://github.com/codeigniter4/CodeIgniter4/pull/2897

PR #2820, opened 2020-04-11
ignore coverage on exit and die

samsonasik:
as never hit in tests.

**Checklist:**
- [x] Securely signed commits


samsonasik:
travis build green, on php73:sqlite build is unrelated.

Issue #2822, opened 2020-04-12
Bug: getSegments() returns an array with 2 empty strings when accessing / 


includebeer:
I have created a filter named 'Localize' in my CI4 app. In the before() function I call getSegments() on the request object. If I go to a URL containing some segments, like domain.com/fr or domain.com/fr/blog it work as expected. getTotalSegments will return 1 or 2, and getSegments will return an array with 'fr' or with 'fr and 'blog'.

The problem is when I go to the root of the website (domain.com), getTotalSegments will return 2 and getSegments will return an array containing 2 empty string:

```
DEBUG - 2020-04-12 16:27:23 --> ---- Filter Localize ----
DEBUG - 2020-04-12 16:27:23 --> getPath() = /
DEBUG - 2020-04-12 16:27:23 --> getTotalSegments() = 2
DEBUG - 2020-04-12 16:27:23 --> getSegments() = Array
(
    [0] => 
    [1] => 
)

DEBUG - 2020-04-12 16:27:25 --> ---- Filter Localize ----
DEBUG - 2020-04-12 16:27:25 --> getPath() = fr
DEBUG - 2020-04-12 16:27:25 --> getTotalSegments() = 1
DEBUG - 2020-04-12 16:27:25 --> getSegments() = Array
(
    [0] => fr
)

DEBUG - 2020-04-12 16:27:33 --> ---- Filter Localize ----
DEBUG - 2020-04-12 16:27:33 --> getPath() = fr/blog
DEBUG - 2020-04-12 16:27:33 --> getTotalSegments() = 2
DEBUG - 2020-04-12 16:27:33 --> getSegments() = Array
(
    [0] => fr
    [1] => blog
)
```

This is with CodeIgniter 4.0.2, PHP 7.4.2, Apache server (MAMP) on macOS.

Expected behavior would be 0 segments for URI / 
```
DEBUG - 2020-04-12 16:27:23 --> ---- Filter Localize ----
DEBUG - 2020-04-12 16:27:23 --> getPath() = /
DEBUG - 2020-04-12 16:27:23 --> getTotalSegments() = 0
DEBUG - 2020-04-12 16:27:23 --> getSegments() = Array
(
)
```

samsonasik:
can be closed in favor of #2824

PR #2824, opened 2020-04-15
fix. URI path is empty

Instrye:
fix. #2822. path is empty and '', URI can't normal work

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -710,7 +710,15 @@ class URI
 	{
 		$this->path = $this->filterPath($path);
 
-		$this->segments = explode('/', $this->path);
+		$tempPath = trim($this->path, '/');
+		if ($tempPath === '')
+		{
+			$this->segments = [];
+		}
+		else
+		{
+			$this->segments = explode('/', $tempPath);
+		}
 
 		return $this;
 	}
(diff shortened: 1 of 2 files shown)


system/HTTP/URI.php
```diff
@@ -726,7 +734,16 @@ public function refreshPath()
...
+		$tempPath = trim($this->path, '/');
+		if ($tempPath === '')
+		{
+			$this->segments = [];
+		}
+		else
+		{
+			$this->segments = explode('/', $tempPath);
+		}
+		$this->segments = explode('/', trim($this->path, '/'));
```
atishamte, line 746: conflicted line with above logic

system/HTTP/URI.php
```diff
@@ -726,7 +734,16 @@ public function refreshPath()
 	{
 		$this->path = $this->filterPath(implode('/', $this->segments));
 
-		$this->segments = explode('/', $this->path);
+		$tempPath = trim($this->path, '/');
```
atishamte, line 737: Can we write this in ternary

system/HTTP/URI.php
Instrye: thanks.

system/HTTP/URI.php
Instrye: thanks.

PR #2833, opened 2020-04-16
add ability for nested language definition

samsonasik:
So, the language file can have something like:

```php
<?php
// Nested.php
return [
   'a' => [
	   'b' => [
		   'c' => [
			   'd' => 'e',
		   ],
	   ],
   ],
];
```

and can be called with `lang('Nested.a.b.c.d');`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated


samsonasik:
travis build green 🎉

lonnieezell:
This needs a mention in the docs and then it's good to go.

samsonasik:
@lonnieezell  I've updated documentation for support nested language definition addition.

lonnieezell:
Looks good. Thanks.

PR #2671, opened 2020-03-07
Re-write userguide to support Python 3 and future proofing Sphinx

jreklund:
This are a complete re-write of the Sphinx support for the manual, changes included and not limited too:
- Changed the requirement to Python 3 and changed conf.py to reflect those changes
- Contribution documentation and readme
- Minimal makefile (for building html/epub...)
- Removed "cilexer" dependency
- Added requirements.txt for automatic building userguide
- Removed bundle theme and load it externally with pip
- Custom Javascript for adding chapter and subject classes to all pages
- Changed some code styling in the userguide that broke

As discussed in #2644.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

PR #2842, opened 2020-04-18
Improve flash of unstyled content in userguide

jreklund:
sphinx_rtd_theme v0.4.3 have [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) (FOUC) problems.

- Removed + next to the navigation to mitigate left to right jumping in navigation
- Changed rendering depth to 2, like userguide v3
- Included a complete replacement of layout.html to be able to mitigate FOUC, by moving CSS above JS.

We still have problems with top to bottom jumping in navigation, due to the length of our navigation. That's not something that can be easily fixed as that's bundled JS with the theme. As per v3 userguide you need to scroll down to the current menu item. In v4 it jumps down to the selected one, giving better UX. That's what causing that jump.

This FOUC are included in the development branch of sphinx_rtd_theme, but as per previous discussions no one knows when they will release a new version (with Sphinx 2 support as well). And due to the nature of the problem, I choose to bundle this fix. As it's to critical for the UX. 

PS. Forgot to include [ci skip] ...

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/_static/css/citheme.css
+++ b/user_guide_src/source/_static/css/citheme.css
@@ -246,6 +246,10 @@ div#pulldown-menu {
 	margin-right: 0.65em;
 }
 
+.wy-menu-vertical li span.toctree-expand::before {
+	display: none !important;
+}
+
 /* Elements ----------------------------------------------------------------- */
 
 .highlighted {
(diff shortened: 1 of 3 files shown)


jreklund:
@lonnieezell Here are the fix for [UX problems](https://forum.codeigniter.com/thread-76146.html) reported in the forum.

lonnieezell:
@jreklund I like limiting it that way and bringing out the grey also. Good changes. Thanks for tracking these down! 

At least if we want to make changes it's simpler now since the layout is in one place. :)

lonnieezell:
@jreklund I'm trying to find our conversation about the javascript solution for the ads on the UG but not having any luck. Do you remember which thread that exists on? Thanks!

jreklund:
@lonnieezell Hi, it's in PR #2671.

PR #2843, opened 2020-04-18
add $segment parameter in pager call by Model.php

paul45:
i don't know why the $segment parameter is not present when the Pager library is call by the Model paginate function

add the $segment option when the Pager library is call by the Model paginate function

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1138,7 +1138,7 @@ class Model
 
 		// Store it in the Pager library so it can be
 		// paginated in the views.
-		$this->pager = $pager->store($group, $page, $perPage, $total,$segment);
+		$this->pager = $pager->store($group, $page, $perPage, $total, $segment);
 		$perPage     = $this->pager->getPerPage($group);
 		$offset      = ($page - 1) * $perPage;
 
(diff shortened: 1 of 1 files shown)


system/Model.php
```diff
@@ -1138,7 +1138,7 @@ public function paginate(int $perPage = null, string $group = 'default', int $pa
 
 		// Store it in the Pager library so it can be
 		// paginated in the views.
-		$this->pager = $pager->store($group, $page, $perPage, $total);
+		$this->pager = $pager->store($group, $page, $perPage, $total,$segment);
```
samsonasik, line 1141: space after `,`

paul45: yes

lonnieezell:
Please update the docs to reflect this change and then I think we're good.

paul45:
I don't know what to write in the docs and the other optional param "$page" is not in the docs 
my english is bad, maybe somebody can rewrite the documentation for me

lonnieezell:
No worries, I'll update it. Thanks. FYI - from your comments here - your English is fine. :)

paul45:
Good thx you

PR #2846, opened 2020-04-19
Improve readability in the userguide with a fixed size

jreklund:
- I have now made the userguide fixed to 900px, as we previously had it extended all the way, making it harder for people to read.
- The CI flame are now positioned to the bottom right corner in the "content" box. It made it harder to read, specifically on mobile devices.
- Making tables responsive again, but with "pre" rendering instead of the default nowrap. Rendering as we see it in .rst, real "enter" in templates results in enter in the userguide itself. Or we will end up with "endless" scrolling in [Validation - Available Rules](https://codeigniter.johaneklund.com/userguide/master/libraries/validation.html#available-rules).
- Fixed incorrect rendering in "Localization". 

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/_static/css/citheme.css
+++ b/user_guide_src/source/_static/css/citheme.css
@@ -256,6 +256,11 @@ div#pulldown-menu {
 	background-color: #f1d40f !important;
 }
 
+/* override table width restrictions */
+.wy-table-responsive table td, .wy-table-responsive table th {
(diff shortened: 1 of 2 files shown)


lonnieezell:
Looks good. Thanks!

PR #2847, opened 2020-04-19
Normalize dir separator of Exceptions::cleanPath and added more paths to clean

paulbalandan:
**Description**
Change the directory separator in cleaning paths to the platform-specific DIRECTORY_SEPARATOR for nicer look, especially for Windows. Also added new paths to clean.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -372,17 +372,23 @@ class Exceptions
 	 */
 	public static function cleanPath(string $file): string
 	{
-		if (strpos($file, APPPATH) === 0)
+		switch (true)
 		{
-			$file = 'APPPATH/' . substr($file, strlen(APPPATH));
-		}
-		elseif (strpos($file, SYSTEMPATH) === 0)
-		{
-			$file = 'SYSTEMPATH/' . substr($file, strlen(SYSTEMPATH));
-		}
-		elseif (strpos($file, FCPATH) === 0)
-		{
-			$file = 'FCPATH/' . substr($file, strlen(FCPATH));
+			case strpos($file, VIEWPATH) === 0:
+				$file = 'VIEWPATH' . DIRECTORY_SEPARATOR . substr($file, strlen(VIEWPATH));
+				break;
+			case strpos($file, APPPATH) === 0:
+				$file = 'APPPATH' . DIRECTORY_SEPARATOR . substr($file, strlen(APPPATH));
+				break;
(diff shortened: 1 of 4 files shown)


system/Debug/Exceptions.php
```diff
@@ -372,17 +372,23 @@ protected function determineCodes(Throwable $exception): array
...
-		}
-		elseif (strpos($file, SYSTEMPATH) === 0)
-		{
-			$file = 'SYSTEMPATH/' . substr($file, strlen(SYSTEMPATH));
-		}
-		elseif (strpos($file, FCPATH) === 0)
-		{
-			$file = 'FCPATH/' . substr($file, strlen(FCPATH));
+			case strpos($file, VIEWPATH) === 0:
+				$file = 'VIEWPATH' . DIRECTORY_SEPARATOR . substr($file, strlen(VIEWPATH));
```
lonnieezell, line 378: Unfortunately, we cannot assume the view path will always be what is set in the config file. When using themes, it's standard to grab a new instance of the renderer while setting a different base path. So no good way to clean this one up. Please remove.

system/bootstrap.php
```diff
@@ -87,6 +87,14 @@
 	define('TESTPATH', realpath($paths->testsDirectory) . DIRECTORY_SEPARATOR);
 }
 
+/**
+ * The path to the views directory
+ */
+if (! defined('VIEWPATH'))
```
lonnieezell, line 93: VIEWPATH wasn't provided on purpose, since views can move around much more often than others. Standard location can be easily accessed via APPPATH.'Views', so not necessary. Please remove.

system/Test/bootstrap.php
```diff
@@ -16,12 +16,14 @@
 defined('APPPATH')       || define('APPPATH', realpath($paths->appDirectory) . DIRECTORY_SEPARATOR);
 defined('WRITEPATH')     || define('WRITEPATH', realpath($paths->writableDirectory) . DIRECTORY_SEPARATOR);
 defined('SYSTEMPATH')    || define('SYSTEMPATH', realpath($paths->systemDirectory) . DIRECTORY_SEPARATOR);
+defined('VIEWPATH')      || define('VIEWPATH', realpath($paths->viewDirectory) . DIRECTORY_SEPARATOR);
```
lonnieezell, line 19: VIEWPATH isn't defined in `system/bootstrap` so don't define it here. See other comments.

PR #2851, opened 2020-04-20
Strip directory separators from auto-generated cell cache name. Fixes…

### LINKED ISSUE #2821: Bug: Cell Caching in View Cells and SOLVE!
Quote from UserGuide in section about View Cells:

> You can cache the results of the view cell call by passing the number of seconds to cache the data for as the third parameter.

And example:

`<?= view_cell('\App\Libraries\Blog::recentPosts', 'limit=5', 300) ?>`

This is variant is **does not work**! Cache file **does not creating**!

BUT! This variant **is WORK** if I pass the **fourth argument** with the file name! Thus, in version 4.0.2 I must always pass the file name!

The problem in file system/Views/Cell.php in line:

```
list($class, $method) = $this->determineClass($library);

// Is it cached?

$cacheName = ! empty($cacheName) ? $cacheName : $class . $method . md5(serialize($params));
```

if I change on:

`$cacheName = ! empty($cacheName) ? $cacheName : md5(serialize($params));`

It's WORK!

lonnieezell:
Strip directory separators from auto-generated cell cache name. Fixes #2821

### DIFF
--- a/system/View/Cell.php
+++ b/system/View/Cell.php
@@ -110,7 +110,9 @@ class Cell
 		list($class, $method) = $this->determineClass($library);
 
 		// Is it cached?
-		$cacheName = ! empty($cacheName) ? $cacheName : $class . $method . md5(serialize($params));
+		$cacheName = ! empty($cacheName)
+			? $cacheName
+			: str_replace(['\\', '/'], '', $class) . $method . md5(serialize($params));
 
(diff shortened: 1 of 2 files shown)


paul45:
Hello 
i think there are the same issue for view caching 
on windows it work only if i set the cache name

PR #2852, opened 2020-04-20
Update Language.php

zl59503020:
Update Language Class getLine method 

**Description**
Compatible language packs

### DIFF
--- a/system/Language/Language.php
+++ b/system/Language/Language.php
@@ -149,6 +149,7 @@ class Language
 			$parsedLine,
 		] = $this->parseLine($line, $this->locale);
 
+		$lines = [];
 		foreach (explode('.', $parsedLine) as $row)
 		{
 			if (! isset($current))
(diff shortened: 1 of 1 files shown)


lonnieezell:
Please describe problem this is trying to fix...

zl59503020:
For example, my language pack looks like this:
file: language/en/App.php
`
return [

    'init' => '',
    
    'base.serverError' => 'The server is abnormal, please contact the manager',
    'base.needLogin' => 'Please Login',
];
`
use
`
echo lang('App.base.needLogin');
`
This will prompt that the corresponding language pack content cannot be found

lonnieezell:
Sorry. I feel this adds unneeded complexity to the code with minimal gains. There's many other ways you could name those language strings and have it work. The dot is treated as a separator and overriding the meaning of that here would only add to confusion down the road. Pass.

Thanks for the contribution, though!

PR #2854, opened 2020-04-20
Support for [option disabled] n form_dropdown.

nowackipawel:
`
form_dropdown(['name'=>'foo', 'disabled_options' => ['bar']], [...options...]);
`
will produce:
`<option value="bar" disabled>....</option>`

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -357,7 +357,7 @@ if (! function_exists('form_dropdown'))
 	 *
 	 * @return string
 	 */
-	function form_dropdown($data = '', $options = [], $selected = [], $extra = ''): string
+		function form_dropdown($data = '', $options = [], $selected = [], $extra = ''): string
 	{
 		$defaults = [];
 		if (is_array($data))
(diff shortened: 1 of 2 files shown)


jreklund:
Personally I see this as a hack more than actually solving the issue.

```function form_dropdown($data = '', $options = [], $selected = [], $extra = '', $disabled = [])```

Not sure how this project are about creating breaking changes, but this will be cleaner:

```function form_dropdown($data = '', $options = [], $selected = [],  $disabled = [], $extra = '')```

lonnieezell:
> Not sure how this project are about creating breaking changes, but this will be cleaner:

We try to follow semantic versioning, which means any BC change would require us to put out a v5. Not ready for that.

lonnieezell:
Not a fan of the change, sorry. And there's no tests.

nowackipawel:
@lonnieezell : am I wrong or there is no way to mark options as disabled right now?
Mine version is not breaking the compatibility of  form_dropdown but  because of it extending it with no to mix html code with form_helper functions.

PR #2858, opened 2020-04-21
getRawInput support json

Instrye:
`getRawInput` get `php://input` and convert it to an array.
the method usualy parse `PUT`,`PATCH`,`DELETE` raw data.
RESTful usualy use `application/json`,
so, i think. the method should parse json data

 ( sorry, My English is poor  )
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -374,8 +374,14 @@ class IncomingRequest extends Request
 	 */
 	public function getRawInput()
 	{
-		parse_str($this->body, $output);
-
+		if ((($ContentType = $this->getHeader('Content-Type')) !== null) && $ContentType->getValue() === 'application/json')
+		{
+			$output = $this->getJSON(true);
+		}
+		else
+		{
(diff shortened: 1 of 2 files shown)


lonnieezell:
We already have the `getJSON()` method which does what you're trying to do here. I appreciate you taking the time to submit a PR, but since this is duplicate functionality, I'm going to have to pass. 

I look forward to your next PR, though.

Instrye:
> We already have the `getJSON()` method which does what you're trying to do here. I appreciate you taking the time to submit a PR, but since this is duplicate functionality, I'm going to have to pass.
> 
> I look forward to your next PR, though.

`getRawInput` just use in `Validation::withRequest`
if `getRawInput` not parse json, Validation is not work.
Or we have a better way,Make Validation support json

lonnieezell:
I would rather have` Validation::withRequest` updated and leave `getRawInput` to actually return the raw input and not modify it in any way.

Issue #2818, opened 2020-04-10
Bug: saveData option doesn't work in Views


njsx:
I have a two views header.php and footer.php with a same content:

`<h3><?= $title ?></h3>`

In my Controller I have this code:
```

public function index() {

      $data['title'] = "Welcome!";

      echo view('header', $data);

      echo view('footer');


```
When I call this Controller I can see on my screen:
`
Welcome!
Welcome!
`

In App\Config\View.php I set:
`public $saveData = false;` (or true doesn't matter)

In the UserManual 

> If you call the function multiple times in a single request, you will have to **pass the desired data to each view**.

Regardless of the transmitted values, this parameter does not work!!! I always see the same thing!

diolan12:
Never try using double view call like yours.
I use once in my controller, just for load the main view.
Because I'm using return, not echo.
Then I load bunch of header or footer etc inside view.

**app/Controllers/Main.php**
```
public function index() {
  $data = array(
    'title' => 'Hello World!',
    'header' => array(
      'greet' = 'Welcome!'
    ),
    'footer' => array(
      'copyright' => 2020
    )
  );
  return view('welcome_message', $data);
}
```

**app/Views/welcome_message.php**
```
<html>
  <head>
    <title><?= $title?></title>
  </head>
  <body>
    <?= view('header', $header)?>
    <?= view('footer', $footer)?>
  </body>
</html>
```

**app/Views/header.php**
```
<head>
  <h3><?= $greet?></h3>
</head>
```

**app/Views/footer.php**
```
<footer>
  <p>Copyright <?= $copyright?></p>
</footer>
```

Notice how I load another view:
`<?= view('some_view', $array)?>`

It is easier to read by human, not just a bunch of hundreds or even thousands of code in one file.

njsx:
Sorry for my bad English...But I try to answer.

I do any projects also as you and I never use multiple times call the function in a single request, but the problem is not in this! The problem is that the stated feature in the user guide DOES not WORK! I am translating official documentation into Russian and I need to tell the community about how this function works. To illustrate by examples, but the examples don'T WORK! That's why I report the problem to the developers.

iRedds:
I confirm a bug in system\Common.php function view()

```php		
$saveData = true;
if (array_key_exists('saveData', $options) && $options['saveData'] === true)
{
	$saveData = (bool) $options['saveData'];
	unset($options['saveData']);
}
```
[...]

PR #2861, opened 2020-04-21
fix. saveData not work

Instrye:
fix. #2818
- fix saveData
- use tempData render view
- saveData is false, Do not delete saved data.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


tests/system/View/ViewTest.php
```diff
@@ -231,13 +231,16 @@ public function testCachedRender()
 	public function testRenderStringSavingData()
 	{
 		$view = new View($this->config, $this->viewsDir, $this->loader);
+		$expected = '<h1>Hello World</h1>';
 
+		//I think saveData is sava current data, is not clean already set data.
 		$view->setVar('testString', 'Hello World');
-		$expected = '<h1>Hello World</h1>';
```
Instrye, line 236: the saveData is true, data is saved. becuse first set false, then set true.

tests/system/View/ParserTest.php
```diff
@@ -957,10 +957,10 @@ public function testRenderStringSavingData()
 
 		$expected = '<h1>Hello World</h1>';
 		$pattern  = '<h1>{testString}</h1>';
-		$this->assertEquals($expected, $parser->renderString($pattern, [], true));
```
Instrye, line 960: the saveData is true, data is saved. becuse first set false, then set true.

PR #2871, opened 2020-04-23
Fixes session active detection on force_https function and add more test CodeIgniter::forceSecureAccess() run force_https()

samsonasik:
- Fixes session active detection on force_https function
- Add more test CodeIgniter::forceSecureAccess() run force_https() when `$config->forceGlobalSecureRequests` is true.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -403,7 +403,7 @@ if (! function_exists('force_https'))
 			$response = Services::response(null, true);
 		}
 
-		if (is_cli() || $request->isSecure())
+		if (ENVIRONMENT !== 'testing' && (is_cli() || $request->isSecure()))
 		{
 			return;
 		}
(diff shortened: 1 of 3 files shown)


system/Common.php
```diff
@@ -403,14 +400,14 @@ function force_https(int $duration = 31536000, RequestInterface $request = null,
...
 		{
 			return;
 		}
-		// @codeCoverageIgnoreStart
-		// If the session library is loaded, we should regenerate
+
+		// If the session status is active, we should regenerate
 		// the session ID for safety sake.
-		if (class_exists('Session', false))
+		if (session_status() === PHP_SESSION_ACTIVE)
```
samsonasik, line 410: Also fix the session check as well. Previosly, the `class_exists('Session')` will always return false as no class named `Session`. Changed with session active check `session_status() === PHP_SESSION_ACTIVE` above.

samsonasik:
I updated PR title description for the new fix.

samsonasik:
I've added test for `force_https()` with null request and response as parameters.

samsonasik:
travis build green 🎉

PR #2872, opened 2020-04-23
Better locale matching against broad groups. Fixes #2774

### LINKED ISSUE #2774: Bug: negotiateLocale bug in Safari with fr-ca locale
I have set the parameter **negotiateLocale = true** in app/Config/app.php
The problem is **request->getLocale()** doesn't always return the right locale.

In Chrome the language is set to french first, then english, and it's working as expected, getLocale return **"fr"**:
```
request->getUserAgent() = Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.149 Safari/537.36
request->getHeaderLine('accept-language') = fr-FR,fr;q=0.9,en-US;q=0.8,en;q=0.7
request->getLocale()                      = fr
```

But in Safari, it's only set to french and getLocale return **"en"**:
```
request->getUserAgent() = Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.5 Safari/605.1.15
request->getHeaderLine('accept-language') = fr-ca
request->getLocale()                      = en
```

This is on macOS 13.15.3 (Catalina) with CodeIgniter 4.0.2, PHP 7.4.2.
The supported languages are set to ['en', 'fr']. 

My guess is there's a bug in the class Negotiate when parsing the accept-language header. The expected behavior would be that getLocale() return "fr" when accept-language is set to "fr-ca". But it looks like it doesn't recognize "fr-ca" as valid, and fall back to "en".

lonnieezell:
Fixes #2774 by implementing a match for locales against the broad locale group.

### DIFF
--- a/tests/system/HTTP/NegotiateTest.php
+++ b/tests/system/HTTP/NegotiateTest.php
@@ -138,6 +138,17 @@ class NegotiateTest extends \CodeIgniter\Test\CIUnitTestCase
 	}
 
 	//--------------------------------------------------------------------
+
+	/**
+	 * @see
+	 */
+	public function testAcceptLanguageMatchesBroadly()
+	{
+		$this->request->setHeader('Accept-Language', 'fr-FR,fr;q=0.9,en-US;q=0.8,en;q=0.7');
+
+		$this->assertEquals('fr', $this->negotiate->language(['fr', 'en']));
+	}
+
 	public function testBestMatchEmpty()
 	{
 		$this->expectException(Exceptions\HTTPException::class);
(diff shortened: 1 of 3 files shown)


tests/system/HTTP/NegotiateTest.php
```diff
@@ -138,6 +138,17 @@ public function testAcceptLanguageBasics()
 	}
 
 	//--------------------------------------------------------------------
+
+	/**
+	 * @see
```
samsonasik, line 143: `@see need link

lonnieezell: Good catch. Thanks!

PR #2873, opened 2020-04-23
Bug fix in Throttler class check validation

jlamim:
**Description**
While I was structuring an example to compose my book, I found that even if I correctly set the parameters for the `check()` method, a request was always executed more than the defined limit.

This was because when checking for the existence of the cache (`($tokens = $this->cache->get($tokenName)) === null`) if it was null, `true` was returned, and just created the cache, without start the count, which always allowed one more request, because only the next time the counter would start.

The `return true` was removed in this check and in checking the amount of tokens it was adjusted to `$tokens> = 0`.

After these changes, the limit defined in `check` started to be applied successfully.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Throttle/Throttler.php
+++ b/system/Throttle/Throttler.php
@@ -145,8 +145,6 @@ class Throttler implements ThrottlerInterface
 			// capacity - 1, and save it to the cache.
 			$this->cache->save($tokenName, $capacity - $cost, $seconds);
 			$this->cache->save($tokenName . 'Time', time(), $seconds);
-
-			return true;
 		}
 
 		// If $tokens > 0, then we need to replenish the bucket
@@ -171,7 +169,7 @@ class Throttler implements ThrottlerInterface
 
 		// If $tokens > 0, then we are safe to perform the action, but
 		// we need to decrement the number of available tokens.
-		if ($tokens > 0)
+		if ($tokens >= 0)
 		{
 			$this->cache->save($tokenName, $tokens - $cost, $seconds);
 			$this->cache->save($tokenName . 'Time', time(), $seconds);
--- a/system/Throttle/Throttler.php
+++ b/system/Throttle/Throttler.php
@@ -167,7 +167,7 @@ class Throttler implements ThrottlerInterface
 		$tokens += $rate * $elapsed;
 		$tokens  = $tokens > $capacity ? $capacity : $tokens;
 
-		// If $tokens > 0, then we are safe to perform the action, but
+		// If $tokens >= 0, then we are safe to perform the action, but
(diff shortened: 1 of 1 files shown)


system/Throttle/Throttler.php
```diff
@@ -171,7 +169,7 @@ public function check(string $key, int $capacity, int $seconds, int $cost = 1):
 
 		// If $tokens > 0, then we are safe to perform the action, but
 		// we need to decrement the number of available tokens.
-		if ($tokens > 0)
+		if ($tokens >= 0)
```
samsonasik, line 172: comment need to be updated

jlamim: Yes, I am updating both comments and checking errors in unit tests.

system/Throttle/Throttler.php
```diff
@@ -139,8 +139,7 @@ public function check(string $key, int $capacity, int $seconds, int $cost = 1):
 		$tokenName = $this->prefix . $key;
 
 		// Check to see if the bucket has even been created yet.
-		if (($tokens = $this->cache->get($tokenName)) === null)
-		{
+		if (($tokens = $this->cache->get($tokenName)) === null) {
```
jreklund, line 142: This should be reverted to match CodeIgniters coding standard.
{ = should be on it's own line.

system/Throttle/Throttler.php
```diff
@@ -169,10 +168,9 @@ public function check(string $key, int $capacity, int $seconds, int $cost = 1):
 		$tokens += $rate * $elapsed;
 		$tokens  = $tokens > $capacity ? $capacity : $tokens;
 
-		// If $tokens > 0, then we are safe to perform the action, but
+		// If $tokens >= 1, then we are safe to perform the action, but
 		// we need to decrement the number of available tokens.
-		if ($tokens > 0)
-		{
```
jreklund, line 175: This should be reverted to match CodeIgniters coding standard.
{ = should be on it's own line.

system/Throttle/Throttler.php
```diff
@@ -169,10 +168,9 @@ public function check(string $key, int $capacity, int $seconds, int $cost = 1):
...
+		if ($tokens >= 1) {
```
ChoiBU, line 173: how about
if ($token)

MGatner:
I think I understand this, and it is a good move. @jlamim there are some good suggestions here, any chance you could take a pass at cleaning it up?

michalsn:
It would be nice to have it fixed, but I think @jlamim is busy, so I'll merge it and then fix the code style myself.

Issue #2875, opened 2020-04-24
Bug: Can't override already defined language in pager


pixobit:
**Describe the bug**
Not completely sure if this was intended, or it's an issue, but I personally find it inconvenient.
I noticed that it's impossible to override the pager translations, unless you change the translation key.

**CodeIgniter 4 version**
4.0.2

**Affected module(s)**
Pager

**Expected behavior, and steps to reproduce if appropriate**
Add pager.php to your languages folder, and insert any of the pager translations, for example:
return [
    'first' => '«',
    'previous' => '‹',
    'next' => '›',
    'last' => '»',
];
Create a pager object with or without a custom template, and use makeLinks to generate the links. You will notice that the expected translations from your file won't be applied, instead the default will be used.

jlamim:
@pixobit  There was really a bug in the paging structure. I already sent a PR with the correction, soon to be released and you can update the code of your application.

PR #2879, opened 2020-04-25
Bug fixes - Pager

jlamim:
**Description**
This PR corrects the following problems:

- issue #2875, where the next and previous links did not use language information, always maintaining a standard format
- next and previous links returning the same values as first and last link

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Pager/PagerRenderer.php
+++ b/system/Pager/PagerRenderer.php
@@ -146,7 +146,7 @@ class PagerRenderer
 	 */
 	public function hasPrevious(): bool
 	{
-		return $this->first > 1;
+		return $this->current > 1;
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lonnieezell:
The current pager system sees the Next and Previous as the next group of available pages to display and not the next/previous page of results. It's unfortunate that made it through to the release but that's they way things are now and can't be changed without BC breaks. We could add new features for `nextPage()` and `previousPage()`, and that should probably be a thing. 

All that to say - your changes to the next and previous functionality breaks the way it's currently intended to work and cannot be accepted. 

The language change is fine, but I can't merge them separately the from the other (rejected) changes.

jlamim:
@lonnieezell I fully understand the fact that I can't accept PR.

I will make a new PR to correct the dictionary issue and another one with the addition of new features for "next" and "prev" based on pages, so as not to break the current structure.

I will also review the documentation of the paging part to make it clearer about the issue of getNext and getPrev not corresponding to pages, but to groups of pages.

lonnieezell:
@jlamim That sounds great. Thanks! I've merged the first PR.

PR #2882, opened 2020-04-25
Make validation placeholders always available

jreklund:
This are another take on #2817 and the PR #2819 to fix it.

There have been a couple of posts in the forum wanting/trying to use placeholders, but it's limited to Models. It's such a nice feature that it should be available in any form of validation.

Now people can e.g. store a rule in their config file and pass it along from the controller.

- Moved fillPlaceholders from Model into Validation
- Added test for is_unique with placeholder. I think one are enough, it either works or it dosen't with `$this->validation->run()`, as all function depending on validation are ending up in the the run() function.
- Moved "Validation Placeholders" from the Model userguide into Validation.
- "Translation Of Messages And Validation Labels" never got rendered in Validation userguide.
- Fixed some indention in Validation userguide. "tabs" gives 8 spaces, instead of 4.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1432,10 +1432,6 @@ class Model
 			return true;
 		}
 
-		// Replace any placeholders (i.e. {id}) in the rules with
-		// the value found in $data, if exists.
-		$rules = $this->fillPlaceholders($rules, $data);
-
 		$this->validation->setRules($rules, $this->validationMessages);
 		$valid = $this->validation->run($data, null, $this->DBGroup);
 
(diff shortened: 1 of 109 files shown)


system/Model.php
```diff
@@ -1473,62 +1469,6 @@ protected function cleanValidationRules(array $rules, array $data = null): array
...
-	 * The value of {id} would be replaced with the actual id in the form data:
-	 *
-	 *  'required|is_unique[users,email,id,13]'
-	 *
-	 * @param array $rules
-	 * @param array $data
-	 *
-	 * @return array
-	 */
-	protected function fillPlaceholders(array $rules, array $data): array
```
samsonasik, line 1496: I think direct remove protected method is BC break. To keep it, it need to be kept with with change like the following:

```php
    /*
     * @deprecated use fillPlaceholders($data) from Validation instead
     */
    protected function fillPlaceholders(array $rules, array $data): array
    {
        return (function ($validation, $data) {
            $validation->fillPlaceholders($data);
            return $validation->getRules();
        })->bindTo($validation = $this->validation, get_class($this->validation))($validation, $data);
    }
```

jreklund: No other functions in CodeIgniter call fillPlaceholders. You wan't this in case someone extended Model?

samsonasik: That's the cost of `protected` method instead of `private`. Ok, let's the maintainer decide to keep or remove it ;)

lonnieezell: Yeah, I have to agree with @samsonasik here. Anyone extending could use it directly. Even though not too many probably are, we should leave it and add the deprecated note, I think.

lonnieezell:
@jreklund After the deprecation notice mentioned above, I think we're good to include this one. Nice change. Thanks.

jreklund:
Hi, thanks for the feedback. I will have the changes ready at the end of the week ready for CI 4.0.3 release.

samsonasik:
@jreklund also, please rebas against latest develop.

jreklund:
Damn... that looks so weird doing a rebase... Dosen't look right at all!

PR #2894, opened 2020-04-27
Added ability to delete row with string primary key via Model::delete($id)

samsonasik:
To make enable  to delete row with string primary key via Model::delete($id) 

**Checklist:**
- [x] Securely signed commits


samsonasik:
travis build green 🎉

PR #2895, opened 2020-04-27
[ci skip] Add `make.bat` for Windows users

paulbalandan:
Enable Windows users to run `make html` without having to install [GnuWin32's make](https://sourceforge.net/projects/gnuwin32/files/make/).

**Description**
This batch file is auto-generated by `sphinx-quickstart` and I think should be committed for the benefit of contributors using Windows.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/user_guide_src/make.bat
+++ b/user_guide_src/make.bat
@@ -1,0 +1,35 @@
+@ECHO OFF
+
+pushd %~dp0
+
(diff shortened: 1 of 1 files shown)


jreklund:
I can back this change, it's the official make file.

PR #2896, opened 2020-04-27
Update user guide README

paulbalandan:
**Description**
Update user guide's README to include instructions for using Python Launcher for Windows.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated

### DIFF
--- a/user_guide_src/README.rst
+++ b/user_guide_src/README.rst
@@ -28,9 +28,13 @@ or ``python3``.
 	python3 --version
 	Python 3.6.9
 
+	# For Windows
+	py --version
+	Python 3.8.1
+
 If you're not on 3.5+, go ahead and install the latest 3.x version from
 `Python.org <https://www.python.org/downloads/>`_. Linux users should use their
-operative systems built in Package Managers to update.
+operating systems' built-in Package Managers to update.
 
 pip
 ---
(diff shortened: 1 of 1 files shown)


jreklund:
Hi, are this a special installation that can't use `python`? Mine are downloaded from Python.org.

```
C:\Users\jreklund>py --version
Python 3.8.2

C:\Users\jreklund>python --version
Python 3.8.2

C:\Users\jreklund>pip --version
pip 20.0.2 from c:\program files (x86)\python38-32\lib\site-packages\pip (python 3.8)

C:\Users\jreklund>pip3 --version
pip 20.0.2 from c:\program files (x86)\python38-32\lib\site-packages\pip (python 3.8)
```

paulbalandan:
Although use of `python` in Windows can be the more straightforward way, there may be instances of multiple Python installations in the system (e.g. installed Python 2.x, 3.x). This is where the Python launcher `py` comes into play.

As put by [PEP 397](https://www.python.org/dev/peps/pep-0397/), "A Python launcher is a single executable which uses a number of heuristics to locate a Python executable and launch it with a specified command line." Simply put, this launcher acts as a resolver that "allows the launcher to be the executable associated with '.py' files to support multiple Python versions concurrently."

You may read more on the Rationale part of the said PEP. Thank you.

jreklund:
Okey, I have now read up on this, and based on provide information and Google about the whole shebang. And for py to actually work as intended we need to specify a shebang in conf.py, to fail safe everything.

`#!/usr/bin/env python3`

Personally I'm more for forcing stuff, as with just `py` it can be `python 2`, if you only have that installed. So my take on the whole thing would be to use `py -3` as we soon™ can't build the user guide with Python 2. 

`py -3 --version`
`py -3 -m pip --version`
`py -3 -m pip install -r user_guide_src/requirements.txt`

jreklund:
I took another look of it, we don't need to modify our conf.py, as the sphinx-build that get's installed already require python 3 (if installed with python 3). So if we just force `py -3` everywhere I think we are gold.

paulbalandan:
I am not fond of forcing stuffs but your suggestion does no harm to me. So I'm in.

paulbalandan:
Closing this PR to update my fork. The rebase looks messy.

jreklund:
If we don't force it. You will be installing Sphinx with Python2 (if that's the primary one), and that will fail. That's why I suggested we should force it. :-)

PR #2897, opened 2020-04-27
Make validation placeholders always available

jreklund:
Same as #2882 but without rebase... this are based on develop instead.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/libraries/validation.rst
+++ b/user_guide_src/source/libraries/validation.rst
@@ -64,11 +64,11 @@ The Form
 Using a text editor, create a form called **Signup.php**. In it, place this
 code and save it to your **app/Views/** folder::
 
-	<html>
-	<head>
-	    <title>My Form</title>
-	</head>
-	<body>
+    <html>
+    <head>
+        <title>My Form</title>
+    </head>
+    <body>
 
         <?= $validation->listErrors() ?>
 
@@ -90,8 +90,8 @@ code and save it to your **app/Views/** folder::
 
         </form>
 
-	</body>
-	</html>
+    </body>
+    </html>
 
 The Success Page
 ================================================
@@ -99,18 +99,18 @@ The Success Page
 Using a text editor, create a form called **Success.php**. In it, place
 this code and save it to your **app/Views/** folder::
 
-	<html>
-	<head>
-	    <title>My Form</title>
-	</head>
-	<body>
+    <html>
+    <head>
+        <title>My Form</title>
+    </head>
+    <body>
 
         <h3>Your form was successfully submitted!</h3>
 
         <p><?= anchor('form', 'Try it again!') ?></p>
 
-	</body>
-	</html>
+    </body>
+    </html>
 
 The Controller
 ================================================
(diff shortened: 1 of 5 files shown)


user_guide_src/source/libraries/validation.rst
```diff
@@ -417,6 +415,37 @@ you previously set, so ``setRules()``, ``setRuleGroup()`` etc. need to be repeat
...
+=======================================================
+
+The Validation class provides a simple method to replace parts of your rules based on data that's being passed into it. This
+sounds fairly obscure but can be especially handy with the ``is_unique`` validation rule. Placeholders are simply
+the name of the field (or array key) that was passed in as $data surrounded by curly brackets. It will be
+replaced by the **value** of the matched incoming field. An example should clarify this::
+
+    $validation->setRules([
+        'email' => 'required|valid_email|is_unique[users.email,id,{id}]'
+    ];
```
samsonasik, line 428: missing close parenthesis

user_guide_src/source/libraries/validation.rst
```diff
@@ -417,6 +415,37 @@ you previously set, so ``setRules()``, ``setRuleGroup()`` etc. need to be repeat
...
+    $_POST = [
+        'id' => 4,
+        'email' => 'foo@example.com'
+    ]
+
+then the ``{id}`` placeholder would be replaced with the number **4**, giving this revised rule::
+
+    $validation->setRules([
+        'email' => 'required|valid_email|is_unique[users.email,id,4]'
+    ];
```
samsonasik, line 442: missing close parenthesis

user_guide_src/source/models/model.rst
```diff
@@ -562,37 +562,6 @@ value an array of fieldnames of interest.::
     // get the rules for only the "city" and "state" fields
     $rules = $model->getValidationRules(['only' => ['city', 'state']]);
 
-Validation Placeholders
------------------------
```
samsonasik, line 566: wait, are you sure about removing `Validation Placeholders` documentation from Model ? It seems still valid?

samsonasik: @jreklund I tried your changes, and `validationRules` definition in Model still working, please revert this removal.

jreklund: I did, the code are only in the Model due to BC. I moved it into the Validation part of the user guide as it's not unique to the Model any more.

It will work from Models, Controller, Libraries etc. Every time you execute $validation->run(), so it's not bound to Model.

samsonasik: Yes, but that's still work if we define `validationRules` in the model, I think the _documentation_ of defining `validationRules` needs to be kept as it will still works if we define there.

jreklund:
The "Validation Placeholders" have been reverted. I also took some time to fix a couple of other things like:
- Spaces instead of tabs.
-  `new \App\Models\UserModel();` instead of ` new App\Models\UserModel();`
- [] instead of `array()`
- setValidationMessage and setValidationMessages didn't render correctly
- 

You can view it here if you fint it easier:
https://codeigniter.johaneklund.com/userguide/master/models/model.html

lonnieezell:
It is in 4.0.3! Thanks.

PR #2899, opened 2020-04-28
New features for pagination

jlamim:
**Description**
The new features comprise a set of methods that will allow the user to build a pagination structure following the traditional way, where "next" and "prev" advance or return to a page and not to a group of pagination links as is the default.

Added methods:

- `hasPreviousPage:` checks if there is a page before the current page
- `getPreviousPage:` retrieves the URI of the page prior to the current page
- `hasNextPage:` checks if there is a page after the current page
- `getNextPage:` retrieves the page's URI after the current page

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Pager/PagerRenderer.php
+++ b/system/Pager/PagerRenderer.php
@@ -343,4 +343,78 @@ class PagerRenderer
 	}
 
 	//--------------------------------------------------------------------
+
+	/**
+	 * Checks to see if there is a "previous" page before our "first" page.
+	 *
+	 * @return boolean
+	 */
+	public function hasPreviousPage(): bool
+	{
+		return $this->current > 1;
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Returns a URL to the "previous" page.
+	 *
+	 * You MUST call hasPreviousPage() first, or this value may be invalid.
+	 *
+	 * @return string|null
+	 */
+	public function getPreviousPage()
+	{
+		if (!$this->hasPreviousPage()) {
+			return null;
+		}
+
+		$uri = clone $this->uri;
+
+		if ($this->segment === 0) {
+			$uri->addQuery($this->pageSelector, $this->current - 1);
+		} else {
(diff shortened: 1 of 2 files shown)


lonnieezell:
Looks good. Thanks! 

It does need new tests added for the new methods and something added to the docs about this.

jlamim:
@lonnieezell I'm already working on testing and updating the documentation. Soon I will give the commits.

system/Pager/PagerRenderer.php
```diff
@@ -343,4 +343,78 @@ protected function updatePages(int $count = null)
...
+	 * Returns a URL to the "next" page.
+	 *
+	 * You MUST call hasNextPage() first, or this value may be invalid.
+	 *
+	 * @return string|null
+	 */
+	public function getNextPage()
+	{
+		if (!$this->hasNextPage()) {
+			return null;
```
samsonasik, line 407: this returns null needs test

system/Pager/PagerRenderer.php
```diff
@@ -343,4 +343,78 @@ protected function updatePages(int $count = null)
...
+	 * Returns a URL to the "previous" page.
+	 *
+	 * You MUST call hasPreviousPage() first, or this value may be invalid.
+	 *
+	 * @return string|null
+	 */
+	public function getPreviousPage()
+	{
+		if (!$this->hasPreviousPage()) {
+			return null;
```
samsonasik, line 369: this return null need test

system/Pager/PagerRenderer.php
jlamim: OK! I will write the tests and update.

system/Pager/PagerRenderer.php
jlamim: OK! I will write the tests and update.

lonnieezell:
Looks good @jlamim Thanks for tackling this!

PR #2907, opened 2020-04-30
Fix Model::first() only use orderBy() when group by is not empty

samsonasik:
This fix the condition, eg: in Postgresql when we can supply sql like:

```sql
SELECT SUM(id) as total FROM "user" WHERE 1 = 1 LIMIT 1 ;
```

without group by, and followed order by. So, make order by only when there is a group by.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -103,7 +103,7 @@ class BaseBuilder
 	 *
 	 * @var array
 	 */
-	protected $QBGroupBy = [];
+	public $QBGroupBy = [];
 
 	/**
 	 * QB HAVING data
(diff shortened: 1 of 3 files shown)


samsonasik:
It seems "group by" resetted on query builder limit call. it require a deep check.

samsonasik:
Fixed. On `withDeleted()->first()`, it requires to group by primary key when there is no group by clause.

samsonasik:
It seems not really working if `Model::useSoftDeletes` is false.

samsonasik:
added check force group by only when Model::useSoftDeletes is true

samsonasik:
travis build green 🎉

PR #2914, opened 2020-05-01
[ci skip] User Guide: Update welcome image to version 4.0.3

jlamim:
**Description**
Update of the page image and welcome to display the structure of version 4.0.3 instead of version 4.0.0-rc.4, avoiding confusion on the part of beginners.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


lonnieezell:
Actually - to save us from having to do this every time, would you mind just removing the version number from the graphic altogether, so it just says "Welcome to CodeIgniter"? That would be helpful.

jlamim:
@lonnieezell Updated image without specifying the version.

samsonasik:
The view at `app/Views/welcome_message.php` needs to be updated.

PR #2916, opened 2020-05-02
Fix is_unique/is_not_unique validation called on POST/PUT via API in Postgresql

samsonasik:
Fixing #2812 . In postgresql, somehow, call via API with Resource Controller and is_unique/is_not_unique validation make query `id != {id}` not work. While in mysql work. This check if ignoreValue doesn't has `{` prefix and `}` suffix.

**Checklist:**
- [x] Securely signed commits


samsonasik:
travis build green. It doesn't broke existing tests.

samsonasik:
I've added unit test for testing run `(new Rules())->is_unique()` to prove the condition in postgresql environment

samsonasik:
travis build green 🎉

michalsn:
I don't like this solution since it will only "patch" this particular validation method and the real problem is in the `fillPlaceholders()` method. Also, it's not related exclusively to the Resource Controller.

This problem occurs when there is no value to replace the placeholder variable in the validation rule. It's not a huge problem when working with MySQL (because we end up with a check like: `$row->where("id !=", {id})`), but with PostgreSQL... well, the same query will produce the type mismatch since this DB engine actually checks the field types.

Anyway, in both cases, I would call it a bug.

We should get rid of the placeholders from validation rules that have no match in the provided variables. This way, we won't fire the additional check: `$row = $row->where("{$ignoreField} !=", $ignoreValue);` in the `is_unique` validation method. Also, any custom validation method that will be created by the developer, will be working as expected - because this way we can actually check if the variable was provided or not and react.

Please see #2917 as an alternative solution.

samsonasik:
@michalsn Thank you, I tried your changes and that's working. I'm closing it.

samsonasik:
re-opened it

jreklund:
Hi, if we decide to with this route `is_not_unique` should behave the same.

As per #2917 please not that these tables will not work anymore. So that we have that in mind.
| Key | Translation|
| ------------- |:-------------:|
| {pages.nextPage}      | Next page |
| {pages.previousPage}      | Previous page      |

Personally I'm a fan of early returns:
```php
if (preg_match('/(?:{)(.*)(?:})/', $ignoreValue) > 0)
{
    return true; // If true are correct, haven't checked...
}
```

michalsn:
My 2 cents. Maybe this regex would work better here since it wouldn't match @jreklund examples with translation string keys.

```php
preg_match('/\{(\w+)\}/', $ignoreValue)
```
Also, early returns are great but here we probably should make a query even if there is no value to ignore.

jreklund:
Yeah, maybe you are right. Better to semi-fail, you can create stuff but don't update stuff if you forget `id`.

I think it's a good compromise, until we have found a better solution to the problem. It's more likely you have a dot syntax for strings.
```php
if (preg_match('/\{(\w+)\}/', $ignoreValue) > 0)
{
    $row = $row->where("{$ignoreField} !=", $ignoreValue);
}
```


I have something similar to this [Localization - Replacing Parameters](https://codeigniter.com/user_guide/outgoing/localization.html#replacing-parameters)
```
{key, type, default, params}
{pages.nextPage, int, 0}
```

samsonasik:
I've updated preg_match check {} on is_unique rule, and also updated is_not_unique with preg_match {} check and add tests for is_not_unique. I removed redundant tests that in separate file  (UniqueRuleTest file)

samsonasik:
travis build green :tada:

samsonasik:
I think it needs is_numeric check for inside "{}" so string like "{abc}" or "{}" itself can be passed.

samsonasik:
It seems numeric check is not possible without getting the `$row->get()->getFieldData()` but that may be take too much check and db compatibility.

samsonasik:
I've  added check field != matches[1] of {(\w+)\} if found .

samsonasik:
hm.., or not, because curl like : 
```
curl -i -X PUT -d "product_code=A1104&product_name=Corn11" http://ci4.local/products/11k
```

will make error in postgresql because `11k` is not a numeric if we check `field != matches[1] of {(\w+)} `.

samsonasik:
Oh, using `preg_match('/\{(\w+)\}/', $ignoreValue)` only still cause an error for the following curl call:

```bash
curl -i -X PUT -d "product_code=A1104&product_name=Corn11" http://ci4.local/products/11k
```

need a better solution.

samsonasik:
I've found a solution for it. I added check against real column type. First, I define the numeric types in mysql, postgres, and sqlite:

```php
public const NUMERIC_TYPES = [
    'bit',
    'int2',
    'int4',
    'int8',
    'float4',
    'float8',
    'numeric',
    \MYSQLI_TYPE_DECIMAL,
    \MYSQLI_TYPE_NEWDECIMAL,
    \MYSQLI_TYPE_FLOAT,
    \MYSQLI_TYPE_DOUBLE,
    \MYSQLI_TYPE_TINY,
    \MYSQLI_TYPE_SHORT,
    \MYSQLI_TYPE_LONG,
    \MYSQLI_TYPE_LONGLONG,
    \MYSQLI_TYPE_INT24,
    \SQLITE3_INTEGER,
    \SQLITE3_FLOAT,
];
```
Then, I check if data found, we check if the field is non-numeric or (numeric && is_numeric($value)): 

```php
$result = $row->get(null, 0, false);
if ($result->getRow() !== null)
{
    $isNumeric = in_array($result->getFieldData()[0]->type, self::NUMERIC_TYPES);
    if (! $isNumeric || $isNumeric && is_numeric($where_value))
    {
        $row = $row->where($where_field, $where_value);
    }
}
```

samsonasik:
the preg_match seems now no longer needed as before filtering, it check the field type.

samsonasik:
I've removed unneeded preg_match() check for {} as it now check the field type first before filtering.

samsonasik:
sqlite seems may detect numeric types as falsy when run manual check `is_not_unique` test. I've added false to list of allowed a numeric types check detection.

samsonasik:
travis build green 🎉 . Ready for review ;)

jreklund:
Correct me if I'm wrong but we are doing the following now right?
- Making up to two extra queries. `$row->get(null, 0, false)` and `$result->getFieldData()`. Or are this cached data already?

michalsn:
We're doing one additional query. `getFieldData()` operates on a result object in this case.

Also, we shouldn't always use `select('1')` - only when `$ignoreField` is empty.

I don't want to be a malcontent, but personally I preferred the initial regex solution, which was already quite effective. This is becoming quite complicated and not easy to maintain, but maybe it's just me.

samsonasik:
As I commented before, the regex won't work to handle api call like the following:

```
curl -i -X PUT -d "product_code=A1104&product_name=Corn11" http://ci4.local/products/11k
```

As `11k` is not a numeric. So, it needs a real DB field type check.

`$row->get(null, 0, false)` ensure to get 1 record but doesn't reset existing select. So, the query can be continued when next condition met.
`$result->getFieldData()` will only get 1 field of selected column if the query only select one column.

For me, additional check to get field type is fine as far as it works.

michalsn:
> As I commented before, the regex won't work to handle api call like the following:
> 
> ```
> curl -i -X PUT -d "product_code=A1104&product_name=Corn11" http://ci4.local/products/11k
> ```
> 
> As `11k` is not a numeric. So, it needs a real DB field type check.

For me, this type of check is a little out of the scope for this rule and should be handled by a different rule specially prepared for `id` field itself - for example, `is_numeric` or something.

The only problem the `is_unique` rule had was running the additional `where` condition when we weren't able to replace the placeholder with data. So, I think that assuming the presence of the pattern `{variable}` is a reasonable compromise.

So... IMO this type of regex would be good enough in this case.
```php
if (! preg_match('/^\{(\w+)\}$/', $ignoreValue))
{
    $row = $row->where("{$ignoreField} !=", $ignoreValue);
}
```

samsonasik:
@michalsn Ok, I've updated with regex again, with `^` and `$`.

samsonasik:
travis build failure at mysql is unrelated with changes

samsonasik:
close-reopen to restart build

samsonasik:
travis build green :tada:

MGatner:
This one has a long history, most of which I’m ignorant about. Do @michalsn and @jreklund both approve? Is this ready?

jreklund:
@MGatner It looks good to me. Consider using squash on this one. So you only get one commit message instead. :-)

samsonasik:
@jreklund commit has been squashed.

samsonasik:
travis build green :tada:

samsonasik:
@MGatner this is ready. Any chance to get it merged? Thank you.

MGatner:
Sorry, yes all done and merged in! Thank you, all three of you, for the communication and collaboration.

Issue #2921, opened 2020-05-03
Translation file not get correct locale in 4.0.3 


nicojmb:
Hi, i just update to 4.0.3 and get translation problems with multiple locales.

I debug the code and i see in RouterCollection Class a new method called "LocalizeRoute", this method check current locale in "Language" class and must be check in "Request".

The problem is in this line:
https://github.com/codeigniter4/CodeIgniter4/blob/7edf31bd3b461384a7b5f50602f879165859d9ec/system/Router/RouteCollection.php#L1280

If change to this, works fine:

`return strtr($route, ['{locale}' => Services::request()->getLocale()]);`

Regards!

PR #2927, opened 2020-05-04
Localization bug fix in localizeRoute method

jlamim:
**Description**
The `CodeIgniter\Router\RouteCollection::localizeRoute` method was in error in the routine that adjusts the URL for the isioma selected by the user through `locale`.

The bug was reported in issue #2921.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1277,7 +1277,7 @@ class RouteCollection implements RouteCollectionInterface
 	 */
 	protected function localizeRoute(string $route) :string
 	{
-		return strtr($route, ['{locale}' => Services::language()->getLocale()]);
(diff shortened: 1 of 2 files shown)


lonnieezell:
This makes sense to me. Since that wasn't caught by a test previously, could you work up a test or two to go with this?

jlamim:
@lonnieezell  I performed functional tests from the original code, without correction, and in fact there was no substitution for {locale} inserted in the route.

I checked the code and realized that in fact there was an error in the call to the method, and with this call it only occurred in the case of a reverse route, identifying this error could really take time.

I will work on unit tests and update soon.

jlamim:
@lonnieezell The requested tests have been added.

PR #2928, opened 2020-05-04
Set validation rules for model dynamically

michalsn:
**Description**
This PR adds the ability to set validation rules for model dynamically via methods: `$model->setValidationRules($rules)` or `$model->setValidationRule($field, $rules)`. This can be very useful when working on applications that have many configuration options.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1381,6 +1381,37 @@ class Model
 	//--------------------------------------------------------------------
 
 	/**
+	 * Allows to set validation rules.
+	 * It could be used when you have to change default or override current validate rules.
+	 *
+	 * @param array $validationRules
+	 *
+	 * @return void
+	 */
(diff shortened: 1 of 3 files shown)


jreklund:
I think we should keep the same syntax as the rest of the user guide.

Like this (copied from user guide). Personally at first glance it didn't look right, but travis said OK so it works. xD
```
        'email'    => [
            'rules'  => 'required|valid_email',
            'errors' => [
                'valid_email' => 'Please check the Email field. It does not appear to be valid.'
            ]
        ]
```

michalsn:
Yeah... now I realized that it's not really documented anywhere :) 
Anyway, consistency is good - fixed.

PR #2931, opened 2020-05-05
make model object via model() function in restful controllers

samsonasik:
Use shortcut to create model, also by using `model()` function, uses shared instance.

**Checklist:**
- [x] Securely signed commits


jreklund:
Shouldn't `model()` just be used on the application side?
This will just increase the loading time a bit.

samsonasik:
Yes and no. If the model already created, eg: in filter, it will uses existing instance.

PR #2938, opened 2020-05-05
Fixes #2932 : Fix Translation key separated by dot

### LINKED ISSUE #2932: Bug: Translation key separated by dot
**Describe the bug**
In version 4.0.2 translation keys may be separated by a dot. E.g

```
return [
     'quantity' => 'Quantity',
     'logs.quantity' => 'The quantity of the product has been changed to <b>{value}</b>',
];
```

However, in version 4.0.3 it no longer works as before. The `logs.quantity` key is not found, but strangely, the value from the` quantity` key is returned. If the `quantity` key is deleted, we get nothing.

After changing the split into nested arrays, everything works as it should

```
return [
    'quantity' => 'Quantity',
    'logs' => [
        'quantity' => 'The quantity of the product has been changed to <b>{value}</b>',
    ],
];

```

**CodeIgniter 4 version**
4.0.3

**Affected module(s)**
Language

Probably after changes in this pull request https://github.com/codeigniter4/CodeIgniter4/pull/2833

If we don't use nested arrays in translations

`lang ('Product.quantity')` should return `Quantity`

`lang ('Product.logs.quantity')` should return `The quantity of the product has been changed to <b>{value}</b>`

**Context**
 - OS: Ubuntu
 - Web server cli-server
 - PHP version 7.3

samsonasik:
Fixes #2932 : Fix Translation key may separated by dot.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Language/Language.php
+++ b/system/Language/Language.php
@@ -163,6 +163,13 @@ class Language
 			}
 		}
 
+		if ($output === null)
+		{
+			$row    = current(explode('.', $parsedLine));
+			$key    = substr($parsedLine, strlen($row) + 1);
+			$output = $this->language[$this->locale][$file][$row][$key] ?? null;
+		}
+
 		if ($output === null && strpos($this->locale, '-'))
 		{
 			[$locale] = explode('-', $this->locale, 2);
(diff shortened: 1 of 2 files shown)


samsonasik:
travis build green 🎉

PR #2940, opened 2020-05-06
Allow Exceptions config to extend BaseConfig

paulbalandan:
**Description**
Allow `Config\Exceptions` to extend the `BaseConfig` so that user-defined modules or external modules may supply their own registrars to populate the properties to their specific needs. Or users may use `.env` for the same purpose.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Exceptions.php
+++ b/app/Config/Exceptions.php
@@ -1,12 +1,13 @@
 <?php namespace Config;
 
+use CodeIgniter\Config\BaseConfig;
+
 /**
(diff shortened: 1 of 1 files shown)


lonnieezell:
I believe exception handling was one of the earlier elements ported over from CI3, so the BaseConfig didn't exist then, and it was missed since then. 

Merging.

PR #2944, opened 2020-05-06
Fix FeatureTestCase paramater

ChoiBU:
if $params is NULL, then it should work with default values


ChoiBU:
It should work with default value

PR #2946, opened 2020-05-06
Fix FeatureTestCase paramater

ChoiBU:
It should work by default value

### DIFF
--- a/system/Test/FeatureTestCase.php
+++ b/system/Test/FeatureTestCase.php
@@ -298,7 +298,7 @@ class FeatureTestCase extends CIDatabaseTestCase
 		$config = config(App::class);
 		$uri    = new URI($config->baseURL . '/' . trim($path, '/ '));
 
-		$request      = new IncomingRequest($config, clone($uri), $params, new UserAgent());
+		$request      = new IncomingRequest($config, clone($uri), $params ?? 'php://input', new UserAgent());
 		$request->uri = $uri;
 
 		$request->setMethod($method);


MGatner:
I think this is good but it also reveals another underlying issue. `IncomingRequest::__construct` has for the inline docs on its third parameter:

` * @param string|null  body`

But we are passing it either `array $params` or default `'php://input'`. Given these functions of `IncomingRequest` I think that `$body` *has* to be a string:
```
	public function getJSON(bool $assoc = false, int $depth = 512, int $options = 0)
	{
		return json_decode($this->body, $assoc, $depth, $options);
	}

	public function getRawInput()
	{
		parse_str($this->body, $output);

		return $output;
	}
```

Which means we can't pass in `$params` as an array. I'm confused how this is working currently, or maybe I am missing something, but shouldn't `$params` be used with `IncomingRequest::setGlobal(string $method, $value)` instead?

I haven't used feature tests yet, maybe some help from @michalsn or @samsonasik  or @jreklund?

michalsn:
I think that `$params` as an array is there to make our life easier. But to actually work, we need a check before assigning this variable to `IncomingRequest`:

```php
if (is_array($params))
{
    $params = http_build_query($params);
}
```

MGatner:
I don't think `$params` should be used for `$body` at all, unless I'm really missing something. I'm confused because it seems like: a) this is blatantly wrong, even the types don't match up, but b) clearly someone has used a feature test before. @lonnieezell any ideas?

jreklund:
This are a semi-dupe of #2947, as the same change are made there. With a quick look we never pass in a second parameter, that's why it never dies. So the $params should be removed and only `'php://input'` should be providedd.

No idea if setupRequest should do anything with $params thought.

jreklund:
Incorrectly setting `string|null` on 14 April 2019; [diff](https://github.com/codeigniter4/CodeIgniter4/commit/928448ea4ae24eae01144dd556db07e930f56a2c#diff-37133b5847b8672dad0a1a3870708d09R153)

lonnieezell:
@MGatner You're correct - we don't need to pass `$params` to the body. It gets handled by [populateGlobals()](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Test/FeatureTestCase.php#L168) call directly after creating the request object. So, it's basically getting ignored currently. I'm working on a couple of additional tests to flesh that out since it doesn't look like tests really ever cover that situation very much.

lonnieezell:
I am still concerned about the use of `php://input` during testing. Not sure how that will translate when running multiple tests that have nothing to do with PHP's actual input stream. Not sure how to test that just yet.

MGatner:
This was handled by #3001

PR #2947, opened 2020-05-06
Feature response constructor

ChoiBU:
How about using the interface as a constructor parameter

### DIFF
--- a/system/Test/FeatureResponse.php
+++ b/system/Test/FeatureResponse.php
@@ -40,6 +40,7 @@ namespace CodeIgniter\Test;
 
 use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\HTTP\Response;
+use CodeIgniter\HTTP\ResponseInterface;
 use Config\Format;
 use PHPUnit\Framework\TestCase;
 
--- a/system/Test/FeatureResponse.php
+++ b/system/Test/FeatureResponse.php
@@ -68,7 +68,7 @@ class FeatureResponse extends TestCase
 	 *
 	 * @param Response $response
 	 */
-	public function __construct(Response $response = null)
+	public function __construct(ResponseInterface $response = null)
 	{
 		$this->response = $response;
 
--- a/system/Test/FeatureResponse.php
+++ b/system/Test/FeatureResponse.php
@@ -67,7 +67,7 @@ class FeatureResponse extends TestCase
 	/**
 	 * Constructor.
 	 *
-	 * @param Response $response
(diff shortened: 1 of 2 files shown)


system/Test/FeatureResponse.php
```diff
@@ -68,7 +69,7 @@ class FeatureResponse extends TestCase
 	 *
 	 * @param Response $response
```
samsonasik, line 70: remove unneded @param as already type hinted, and the type hint already changed

samsonasik: `@var` in response property need to be updated

system/Test/FeatureResponse.php
```diff
@@ -40,6 +40,7 @@
 
 use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\HTTP\Response;
```
samsonasik, line 42: should no longer needed as no longer used

samsonasik:
`@var` in response property need to be updated

system/Test/FeatureResponse.php
ChoiBU: Fix it

system/Test/FeatureResponse.php
ChoiBU: update complete

system/Test/FeatureTestCase.php
```diff
@@ -298,7 +298,7 @@ protected function setupRequest(string $method, string $path = null, array $para
 		$config = config(App::class);
 		$uri    = new URI($config->baseURL . '/' . trim($path, '/ '));
 
-		$request      = new IncomingRequest($config, clone($uri), $params, new UserAgent());
+		$request      = new IncomingRequest($config, clone($uri), $params ?? 'php://input', new UserAgent());
```
lonnieezell, line 301: Since the FeatureTestCase class is always calling IncomingRequest from within code, is there any point to using `php://input` here?

ChoiBU: Needed when sending a request body to json

MGatner:
@ChoiBU I like the change to using the interface, but the changes to `FeatureTestCase` don't belong in here and are already addressed by #3001. Could you remove that from the PR and resubmit for testing?

ChoiBU:
> @ChoiBU I like the change to using the interface, but the changes to `FeatureTestCase` don't belong in here and are already addressed by #3001. Could you remove that from the PR and resubmit for testing?

Here it is #3060

PR #2948, opened 2020-05-06
Compress Timeline for Database Queries

mostafakhudair:
Simplified database queries component shown in timeline to decrease timeline for a hundred queries
Before:
before
After:
after

### DIFF
--- a/system/Debug/Toolbar/Collectors/Database.php
+++ b/system/Debug/Toolbar/Collectors/Database.php
@@ -145,15 +145,21 @@ class Database extends BaseCollector
 			];
 		}
 
+		$start = 0;
+		$duration = 0;
+
 		foreach (static::$queries as $query)
 		{
-			$data[] = [
-				'name'      => 'Query',
-				'component' => 'Database',
-				'start'     => $query->getStartTime(true),
-				'duration'  => $query->getDuration(),
-			];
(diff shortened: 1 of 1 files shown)


lonnieezell:
@mostafakhudair  I like the idea in general. I can see instances where it might be misleading though. Say a query was done to check authorization at the very beginning, then an external API was called that took a little time, then more queries were done to update information before returning a response. In this case, it would show as starting at the first time and lasting for maybe 5ms, even if there was 1 second of API calling in between the first and the second. 

I think that looking at the gap between calls and splitting it off into a separate row if the gap is sufficiently big (a configurable setting) would be ideal here. That would give a more accurate representation if someone is looking at the bar to try to figure out when things were happening.

system/Debug/Toolbar/Collectors/Database.php
```diff
@@ -145,13 +145,22 @@ protected function formatTimelineData(): array
 			];
 		}
 
-		foreach (static::$queries as $query)
+		if (count(static::$queries) > 0)
```
samsonasik, line 148: use truthy check instead (faster):

```suggestion
              if (static::$queries)
```

MGatner:
@mostafakhudair Have you had a chance to read Lonnie's comment above? Any chance you're up for another pass at this?

PR #2950, opened 2020-05-06
Ensure get translation output logic work on selected locale, dashed locale, and fallback "en"

samsonasik:
By move the logic to get translation output to separate method.

**Checklist:**
- [x] Securely signed commits


samsonasik:
I also added new test  to verify that dot can be inside key language for nested array.

samsonasik:
travis build green :tada:

Issue #2949, opened 2020-05-06
Bug: Request URI segment is our of range CI 4.0.3


desttools:
**Describe the bug**
I have code in my app structured like this:
`if ($request->uri->getSegment(2)){
...
}`

It evaluates fine in 4.0.2 and earlier, but breaks in 4.0.3, throwing this error:

`
CodeIgniter\HTTP\Exceptions\HTTPException
Request URI segment is our of range: 1 
`

**CodeIgniter 4 version**
4.0.3 master

**Affected module(s)**
Which package or class is the bug in, if known.

**Expected behavior, and steps to reproduce if appropriate**
Previously this returned NULL. Maybe my code needs to be updated to restructure this conditional? 

**Context**
 - ubuntu0.18.04.4
 - Apache/2.4.29 (Ubuntu)
 - PHP: 7.2.24-0

samsonasik:
it seems you need to update your code, something like this:

```php
if ($request->uri->getTotalSegments() >= 2 && $request->uri->getSegment(2)) 
{

}

```
Just an example, not in front of pc ;)

michalsn:
Oh, dear... I guess it's not gonna be very friendly to use in the long run... I haven't noticed this earlier but I would love to see this method to work similarly to one known from CI3. I guess it's a bit too late now.

Anyway... it's working that way from the beginning. No changes were made. You are experiencing this exception because it will always occur when you ask for a segment which number is greater than your total number of segments right now + 2.

crustamet:
Yeah i had this issue i resolved like how **samsonasik** said.

Of course if you use this how the old CI3 worked, you will have many problems to fix.
That is why CI4 is better, they force you to fix stupid stuff happened to CI3.

All nice and dandy. I think it is better to verify the segment that you are not out of range to know exactly the route on which you on.

But there is another issue i had to workaround that i really don`t like and understand.
When you get segments and you have no segment in the url
it returns an array of two empty elements

```
$request->uri->getSegment(0);

returns [0=>'', 1=>''];
```
It is cool not that much of a deal i think they return like this because something from the core needs to be empty to verify the controller and the method.

I am sure they can fix this but it will change a lot of files.

michalsn:
@crustamet I have to kindly disagree that in CI3 it was working "stupid". If you asked for a segment that wasn't there you always received a `null` in response. Personally, I would say the current implementation makes it more complicated to use.

jreklund:
[...]

PR #2952, opened 2020-05-08
Improvements for getSegment() method from URI class

michalsn:
**Description**
This PR introduces some improvements for `URI::getSegment()` method, adding two optional parameters:

* `$default` - default value to use if given segment is empty
* `$silent` - if set to `true` it prevents throwing an exception

```php
// this will return 'foo'
$uri->getSegment(11, 'foo', true);
// and this will throw an exception - as usual
$uri->getSegment(11);
```

This is related to the issue #2949

I think that this might be very handy in some situations. Handling the scenario where we want to check if given parameter is out there is not very friendly for developers right now. These changes will save some of us a lot of code.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -482,19 +482,12 @@ class URI
 		// but we still have to deal with a zero-based array.
 		$number -= 1;
 
-		if ($number > count($this->segments))
+		if (! $silent && $number > count($this->segments))
 		{
-			if ($silent)
-			{
-				return $default;
-			}
-
 			throw HTTPException::forURISegmentOutOfRange($number);
 		}
 
-		return (isset($this->segments[$number]) && $this->segments[$number] !== '')
-		 ? $this->segments[$number]
-		 : $default;
+		return $this->segments[$number] ?? $default;
 	}
 
 	/**
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -469,12 +469,14 @@ class URI
 	/**
 	 * Returns the value of a specific segment of the URI path.
 	 *
-	 * @param integer $number
+	 * @param integer     $number  Segment number
+	 * @param string|null $default Default value when segment is empty (equal to '')
+	 * @param boolean     $silent  Throw an exception or not
 	 *
-	 * @return string     The value of the segment. If no segment is found,
-	 *                    throws InvalidArgumentError
+	 * @return string|null The value of the segment. If no segment is found,
+	 *                     throws InvalidArgumentError
 	 */
-	public function getSegment(int $number): string
+	public function getSegment(int $number, ?string $default = '', bool $silent = false): ?string
 	{
 		// The segment should treat the array as 1-based for the user
 		// but we still have to deal with a zero-based array.
@@ -482,10 +484,17 @@ class URI
 
 		if ($number > count($this->segments))
 		{
+			if ($silent)
+			{
+				return $default;
+			}
+
 			throw HTTPException::forURISegmentOutOfRange($number);
 		}
 
-		return $this->segments[$number] ?? '';
+		return (isset($this->segments[$number]) && $this->segments[$number] !== '')
+		 ? $this->segments[$number]
+		 : $default;
 	}
 
 	/**
--- a/tests/system/HTTP/URITest.php
+++ b/tests/system/HTTP/URITest.php
@@ -65,6 +65,45 @@ class URITest extends \CodeIgniter\Test\CIUnitTestCase
 
 	//--------------------------------------------------------------------
 
+	public function testSegmentOutOfRangeWithDefaultValue()
+	{
+		$this->expectException(HTTPException::class);
+		$url = 'http://abc.com/a123/b/c';
+		$uri = new URI($url);
+		$uri->getSegment(22, 'something');
+	}
+
+	//--------------------------------------------------------------------
+
+	public function testSegmentsWithDefaultValue()
+	{
+		$uri = new URI('http://hostname/path/to');
+
+		$this->assertEquals(['path', 'to'], $uri->getSegments());
+		$this->assertEquals('path', $uri->getSegment(1));
+		$this->assertEquals('to', $uri->getSegment(2, 'different'));
+		$this->assertEquals('script', $uri->getSegment(3, 'script'));
+		$this->assertEquals('', $uri->getSegment(3));
+
+		$this->assertEquals(2, $uri->getTotalSegments());
+		$this->assertEquals(['path', 'to'], $uri->getSegments());
+	}
+
+	//--------------------------------------------------------------------
+
+	public function testSegmentOutOfRangeSilent()
+	{
+		$uri = new URI('http://hostname/path/to/script');
+
+		$this->assertEquals('', $uri->getSegment(22, '', true));
+		$this->assertEquals('something', $uri->getSegment(33, 'something', true));
+
+		$this->assertEquals(3, $uri->getTotalSegments());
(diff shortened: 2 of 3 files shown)


jreklund:
Don't validate anything:
```
testSegmentOutOfRangeWithDefaultValue
```

How can this even be true? You don't suppress the message, so it should throw an error.
```
$this->assertEquals('script', $uri->getSegment(3, 'script'));
$this->assertEquals('', $uri->getSegment(3));
```

You don't need the silent part. If it existed, it would be populated or give you the default value.
```php
public function getSegment(int $number, string $default = ''): string
{
	// The segment should treat the array as 1-based for the user
	// but we still have to deal with a zero-based array.
	$number -= 1;

	if ($default !== '' && $number > count($this->segments))
	{
		throw HTTPException::forURISegmentOutOfRange($number);
	}

	return $this->segments[$number] ?? $default;
}
```

michalsn:
> Don't validate anything:
> 
> ```php
> testSegmentOutOfRangeWithDefaultValue
> ```
> 
> How can this even be true? You don't suppress the message, so it should throw an error.
> 
> ```php
> $this->assertEquals('script', $uri->getSegment(3, 'script'));
> $this->assertEquals('', $uri->getSegment(3));
> ```

Well... you have to look at the code and check how it is working now. The Exception is thrown only if you're trying to get a segment that is greater than `total segments + 2`. So this line of code `$this->assertEquals('', $uri->getSegment(3))` would be perfectly valid even without merging my changes.

About the part that proposes to skip the `$silent` parameter.

This implementation below wouldn't allow us to return an empty string for segments that don't exist - rather unacceptable. 
> ```php
> public function getSegment(int $number, string $default = ''): string
> {
> 	// The segment should treat the array as 1-based for the user
> 	// but we still have to deal with a zero-based array.
> 	$number -= 1;
> 
> 	if ($default !== '' && $number > count($this->segments))
> 	{
> 		throw HTTPException::forURISegmentOutOfRange($number);
> 	}
> 
> 	return $this->segments[$number] ?? $default;
> }
> ```

And this one, as you pointed out, would return `null` in some cases even when we don't really want it.
> ```php
> public function getSegment(int $number, string $default = null): ?string
> {
> 	// The segment should treat the array as 1-based for the user
> 	// but we still have to deal with a zero-based array.
> 	$number -= 1;
> 
> 	if (! is_null($default) && $number > count($this->segments))
> 	{
> 		throw HTTPException::forURISegmentOutOfRange($number);
> 	}
> 
> 	return $this->segments[$number] ?? $default;
> }
> ```

So, that's why the `$silent` parameter is required. Proposed implementation changes nothing in the way this method work when you don't use these new parameters. I think this is a reasonable compromise to simplify the work of developers and don't introduce BC changes.

jreklund:
Yeah, sorry about that. That's even more stupid in my opinion. Why would you allow fetching something that's not there. So I didn't think about +1 thing. 

setSegment() works the same now that I look at it. You can set it +2 more than there are segments. It still gets appended as +1.

Did some actual testing with this one now, instead of just writing from memory. I can't get any segment with '', unless I have manually used setSegment(). Are there any special cases I'm missing? As I can't trigger this one.
```
isset($this->segments[$number]) && $this->segments[$number] !== ''
```

I get the same result with the one that we had before.
```
$this->segments[$number] ?? $default
```

Too bad we have made a real release, because "fixing" the problem will require all programmers to use three parameters. As we can't do BC changes. Personally I get the same result with this one. Less code yay!
```php
public function getSegment(int $number, ?string $default = '', bool $silent = false): ?string
{
	// The segment should treat the array as 1-based for the user
	// but we still have to deal with a zero-based array.
	$number -= 1;

	if (! $silent && $number > count($this->segments))
	{
		throw HTTPException::forURISegmentOutOfRange($number);
	}

	return $this->segments[$number] ?? $default;
}
```

Seeing how this works, I'm just going to replace it, if I will ever need to use it. I have already replaced some. :-(
```php
public function getSegment(int $number, string $default = null): ?string
{
    return $this->segments[$number - 1] ?? $default;
}
```

michalsn:
The thing with `setSegment()` is actually reasonable and I understand it since you can't have segments that don't exist, like ie. there are no segments 4 and 5 but then all of sudden you have segment 6? That would have no sense and would be not valid.

> Did some actual testing with this one now, instead of just writing from memory. I can't get any segment with '', unless I have manually used setSegment(). Are there any special cases I'm missing? As I can't trigger this one.
> 
> ```
> isset($this->segments[$number]) && $this->segments[$number] !== ''
> ```
> 
> I get the same result with the one that we had before.
> 
> ```
> $this->segments[$number] ?? $default
> ```

Yeah... you are right. I have no idea why I changed it - I will send an update. Thanks.

> Seeing how this works, I'm just going to replace it, if I will ever need to use it. I have already replaced some, as they differ to much from they "should" work, based on how other frameworks have solved the same thing.

I guess every framework has its own "things". And "this thing" isn't something really terrible - as long as this change will be merged ;P

We all have some habits and expectations about how something "should" works. And the truth is that before the v4 release there were not so many real-world applications built, so there are probably still a few things to polish up.

jreklund:
If you only want valid positions with `setSegment()`, you should not be able to set 5 in case there aren't one in segment 4. Don't see the use of jumping one over, same with `getSegment()`. This may just be me thought.

How's your take on adding a `setSilent()` like the MigrationRunner? So you can permanently disable the throws as well (need to silently apply to all functions in URI). Because always setting three parameters to `getSegment()` will get tiresome. 

Of course it will be differences, but CI4 are the only major player that throws an error message for this. Sure you can set a standard.

```php
// You are more likely to use an if statement than use it an try block.
if(($id = $this->request->uri->getSegment(3)) === null)
{
    throw new Exception('No id found...');
}

// Instead of 
try
{
    $id = $this->request->uri->getSegment(3);
}
catch(Exception $e)
{
   throw new Exception('No id found...');
}

// Or this to be able to use it in an if statement later.
try
{
    $id = $this->request->uri->getSegment(3);
}
catch(Exception $e)
{
   $id = null
}

if($id === null)
{
    // Load error to user
}
else
{
   // Load real content
}

```

michalsn:
> If you only want valid positions with setSegment(), you should not be able to set 5 in case there aren't one in segment 4.

I haven't looked at `setSegment()` before but now I checked it and you're right. I would call it a bug. I think I will open an issue to address these problems with `getSegment()` and `setSegment()` because besides the proposal from this PR, there is a bigger problem in my opinion.

> Don't see the use of jumping one over, same with getSegment(). This may just be me thought.

In some applications, I used it heavily - ie. when we had many categories with different options. We didn't always have a full URL, so support for this and default values were a real saver there.

`setSilent()` - at first I was like "no"... but actually it makes sense. But I guess this would have to be addressed in a different PR. For now, I will probably wait for the opinion of someone from the core team regarding all of this.

Personally, I would prefer one-liner with three parameters than catching the exception every single time. The other advantage is that you can set a default value really easily.

MGatner:
I was thinking of a `setSilent()` method before I even got that far in your conversation. I like that it is consistent with other components and that it keeps parameters to a minimum. This PR would be adding two parameters to one method, but if we want a silent option down the road it would add more parameters to another method. I'm open to counter-arguments but I'd rather see this resolved now before we create the new public method and would need a BC to adjust it later.

MGatner:
Should have also specified that I think this is a good change and a very nice PR! Thanks for the work, and for the review from @jreklund

jreklund:
I think after the setSilent() method we could implement a default value as well (but without $silent of course), as it's a nice feature. :-)

PR #2958, opened 2020-05-08
Add altPath to choose in Exceptions::render

paulbalandan:
**Description**
In `Exceptions::render()`, added `$altPath` and `$altView` so that if the path configured in `Config\Exceptions` does not exist or the view does not exist either, then it will look for the error view file provided as default by the framework.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -104,7 +104,7 @@ class Exceptions
 	{
 		$this->ob_level = ob_get_level();
 
-		$this->viewPath = realpath(trim($config->errorViewPath)) . DIRECTORY_SEPARATOR;
+		$this->viewPath = rtrim($config->errorViewPath, '\\/ ') . DIRECTORY_SEPARATOR;
 
 		$this->config = $config;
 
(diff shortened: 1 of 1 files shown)


system/Debug/Exceptions.php
```diff
@@ -330,7 +339,7 @@ protected function collectVars(Throwable $exception, int $statusCode): array
 	 *
 	 * @param \Throwable $exception
 	 *
-	 * @return array
+	 * @return integer[]
```
MGatner, line 342: Is this actually a valid docblock return type?

MGatner:
This looks good! See my comment on inline docs, but ready to merge after that.

system/Debug/Exceptions.php
paulbalandan: This was actually `int[]` but phpcbf changed it to the long form because `codeigniter4/coding-standard` changes `bool` and `int` to their respective long forms.

paulbalandan: Fixed docblock

MGatner:
Looks good! Thanks for addressing the docblock.

PR #2964, opened 2020-05-09
Bug fix Pager Segment parameter out of bound fallback set current page = 1

samsonasik:
Handle when segment parameter out of bound fallback set current page = 1 at https://github.com/codeigniter4/CodeIgniter4/blob/a063f14e62b36531f52afa0bf71b8e54c00f6399/system/Pager/Pager.php#L536-L543

**Checklist:**
- [x] Securely signed commits

By this, the `Pager` also will have 100% test coverage.


samsonasik:
travis build green :tada:

PR #2973, opened 2020-05-11
Fix variable name

ChoiBU:
Property name is wrong


MGatner:
It's odd that this hasn't caused issues or test failures, but I think you are correct that this would have prevented reusing stored results.

PR #2976, opened 2020-05-11
clean up: remove unneeded require_once in app/Config/Services.php

samsonasik:
It is unneded since the Autoloading already handle it in `system/bootstrap.php`. 

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Config/Services.php
+++ b/app/Config/Services.php
@@ -2,8 +2,6 @@
 
 use CodeIgniter\Config\Services as CoreServices;
 
-require_once SYSTEMPATH . 'Config/Services.php';
-
 /**
  * Services Configuration file.
  *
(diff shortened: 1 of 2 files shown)


samsonasik:
close-reopen to restart travis build

samsonasik:
travis build green :tada:

kloukas:
This change caused the following error

```
PHP Fatal error:  Uncaught Error: Class 'CodeIgniter\Config\Services' not found in /var/www/test.biglle.com/app/Config/Services.php

```

This is a fresh installation, installed with git clone.
Using nginx and php-fpm if it matters.

Re-adding the line solves the issue.

samsonasik:
the test shows that it should work, tested with apache in my local dev, the test in travis work as well with simulation rename autoload, I will check again tomorrow

kloukas:
Same issue reported [here](https://github.com/codeigniter4/CodeIgniter4/commit/d2d0d77ed7f886bc13ae881dd553e869d9f77d4d#commitcomment-39097525)

samsonasik:
seems related with class_exists in system/bootstrap to be autoloaded true in 2nd param, I will check it

samsonasik:
@kloukas please check https://github.com/codeigniter4/CodeIgniter4/pull/2977 if that works

PR #2977, opened 2020-05-11
ensure load "SYSTEMPATH . Config/Services.php" in system/bootstrap.php

samsonasik:
This should fix the issue reported by @kloukas  https://github.com/codeigniter4/CodeIgniter4/pull/2976#issuecomment-626948222

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Config/Autoload.php
+++ b/app/Config/Autoload.php
@@ -1,6 +1,6 @@
 <?php namespace Config;
 
-require_once SYSTEMPATH . 'Config/AutoloadConfig.php';
+use CodeIgniter\Config\AutoloadConfig;
 
 /**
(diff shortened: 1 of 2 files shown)


samsonasik:
applied  move out require_once of AutoloaderConfig from Config\Autoload class to bootstrap as well

samsonasik:
I tested and it works in with and without composer autoload

kloukas:
Yup, that fixed it. Thanks!

samsonasik:
@kloukas than you for the check. @lonnieezell could you verify ? Thank you.

Instrye:
@lonnieezell please check this PR. current develop can't work.
image

Issue #2965, opened 2020-05-10
Bug: 'method' is an empty string in PATH_INFO causes Unhandled Exception


willnode:
**Describe the bug**

When running `php spark serve`, I noticed that if I set this to `PATH_INFO`

https://github.com/codeigniter4/CodeIgniter4/blob/a063f14e62b36531f52afa0bf71b8e54c00f6399/app/Config/App.php#L53

And navigate to URL `localhost:8080/home/`, causes `ErrorException: Unitialized string offset: 0` in 

https://github.com/codeigniter4/CodeIgniter4/blob/a063f14e62b36531f52afa0bf71b8e54c00f6399/system/CodeIgniter.php#L865

Because `$this->method` is `""`.

**CodeIgniter 4 version**
Version 4.0.3 release (~in latest develop, it's just shows 404 page, skipping the actual error and throws `PageNotFoundException::forMethodNotFound`, probably because recent changes in error reporting level?~ sorry I forgot that by default CI is in production mode) 

**Affected module(s)**
None, I think.

**Expected behavior, and steps to reproduce if appropriate**
It should render the same page as `localhost:8080/` or `localhost:8080/home/index`

**Context**
 - OS: Windows 10 64 bit
 - PHP version 7.3.9

willnode:
There's a workaround that works for me:

https://github.com/codeigniter4/CodeIgniter4/blob/a063f14e62b36531f52afa0bf71b8e54c00f6399/system/Router/Router.php#L231-L236

I changed them into:

```
	public function methodName(): string
	{
		return ($this->translateURIDashes
			? str_replace('-', '_', $this->method)
			: $this->method) ?: 'index';
	}
```

However I'm not sure if this a good bugfix as a PR.

willnode:
Additional info: I'm using `PATH_INFO` because I found it plays nice and well to remove `/public` in my local apache server using [.htaccess](https://github.com/willnode/crm-toolkit/blob/master/api/.htaccess) file to root project.

Instrye:
Default, method is 'index'. 
![image](https://user-images.githubusercontent.com/9927289/81626924-9016ab80-942f-11ea-9f0e-030c305caa06.png)

willnode:
@Instrye sorry I forgot to mention that it happens when you add a trailing slash

![image](https://user-images.githubusercontent.com/20214420/81627233-ec75cd00-9427-11ea-9459-37c6b81c7dc2.png)

PR #2978, opened 2020-05-12
Fix. the parsing method is empty

Instrye:
Fix: #2965 
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -534,11 +534,7 @@ class Router implements RouterInterface
 		// has already been set.
 		if (! empty($segments))
 		{
-			$tempMethod = array_shift($segments);
-			if ($tempMethod !== '')
-			{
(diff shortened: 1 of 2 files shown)


system/Router/Router.php
```diff
@@ -534,7 +534,11 @@ public function autoRoute(string $uri)
 		// has already been set.
 		if (! empty($segments))
 		{
-			$this->method = array_shift($segments);
+			$tempMethod = array_shift($segments);
+			if ($tempMethod !== '')
+			{
+				$this->method = $tempMethod;
+			}
```
samsonasik, line 541: I think it can use one line as `$tempMethod` is not going to be re-use:

```php
$this->method = array_shift($segments) ?: $this->method;
```

PR #2984, opened 2020-05-13
Making the Framework Compatible with PHP 7.10 standards

John-Betong:
Adding the following script to index.php and to every single ./system PHP file makes the script compatible with the PHP Strict 7.?? source.

If and only if 'declare(strict_types=1)' is declared in the parent file, strict_types have absolutely no effect. 

The following three files fail PHP strict validation because they "type juggle" to change the numeric to a bool value:

Files:
./app/Config/Boot/development.php
./app/Config/Boot/production.php
./app/Config/Boot/testing.php

// FAIL
// defined('CI_DEBUG') || define('CI_DEBUG', 1);

// PASS:
defined('CI_DEBUG') || define('CI_DEBUG', TRUE);

These three small changes make it possible to render the Welcome page using PHP Strict Mode.


samsonasik:
please change target to develop branch instead of gh-pages.

lonnieezell:
Hey John - looks like you're doing a PR against the wrong branch. You're targeting `gh_pages` instead of develop.

MGatner:
Please resubmit as a PR against `develop`.

PR #2987, opened 2020-05-13
Update DotEnv.php

lazharous:
**Description**
Bug Fixing: Nested variables are now parsed correctedly, the previous version called the setVariable() function in a loop after all the parsing happened, returning always null for nested vars.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/DotEnv.php
+++ b/system/Config/DotEnv.php
@@ -76,19 +76,9 @@ class DotEnv
 	 */
 	public function load(): bool
 	{
-		$vars = $this->parse();
+		$loaded = $this->parse();
 
-		if ($vars === null)
-		{
-			return false;
-		}
-
-		foreach ($vars as $name => $value)
-		{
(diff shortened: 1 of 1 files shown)


system/Config/DotEnv.php
```diff
@@ -76,34 +76,24 @@ public function __construct(string $path, string $file = '.env')
...
 	//--------------------------------------------------------------------
 
 	/**
 	 * Parse the .env file into an array of key => value
 	 *
-	 * @return array|null
+	 * @return boolean
 	 */
-	public function parse(): ?array
+	public function parse(): bool
```
samsonasik, line 91: change return type of public method is always bc break

MGatner: Agree with this, we need a way to make this fix work without changing the return type.

PR #2990, opened 2020-05-14
fix. Time::setTimezone can't work

### LINKED ISSUE #2989: Bug: Difference in behaviour of native DateTime::setTimezone() and CI's Time:setTimezone()
**Describe the bug**
The native php DateTime::setTimezone() functions converts the time of object to the specified timezone, CI's Time Class which extends DateTime simply changes the Timezone without doing any conversion.

**CodeIgniter 4 version**
Current develop branch

Consider the following code

```php
$time = new \DateTime('13 May 2020 10:00 GMT');
$time->setTimeZone(new \DateTimeZone('GMT+1'));
```
This creates a new DateTime object with time `2020-05-13T10:00:00+00:00` and after conversion it becomes `2020-05-13T11:00:00+01:00`, the timezone changes and the time is converted to the new timezone.

Doing the same using CI's Time class

```php
$time = new Time('13 May 2020 10:00 GMT');
$time = $time->setTimezone(new \DateTimeZone('GMT+1'));
```

Initially, this creates an object with time `2020-05-13T10:00:00+00:00` as well, after conversion however the object has time `2020-05-13T10:00:00+01:00`, the timezone changes but there is not conversion done. The time is 10:00AM in both instances

If this is the intended behaviour and not a bug should the docs be changed to reflect that? They currently say "Converts the time from it’s current timezone into the new one" which does not exactly describe what the function is doing.

Instrye:
Fix #2989
`Time::setTimezone` shuold same `DateTime::setTimezone`
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -773,7 +773,7 @@ class Time extends DateTime
 	/**
 	 * Returns a new instance with the revised timezone.
 	 *
-	 * @param \DateTimeZone $timezone
+	 * @param string|\DateTimeZone $timezone
 	 *
 	 * @return \CodeIgniter\I18n\Time
 	 * @throws \Exception
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -780,7 +780,8 @@ class Time extends DateTime
 	 */
 	public function setTimezone($timezone)
 	{
-		return Time::parse($this->toDateTimeString(), $timezone, $this->locale);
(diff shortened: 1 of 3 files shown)


system/I18n/Time.php
```diff
@@ -780,7 +780,8 @@ protected function setValue(string $name, $value)
 	 */
 	public function setTimezone($timezone)
```
samsonasik, line 781: the `@param` can be updated with : 

```php
@param string|\DateTimeZone $timezone
```

PR #2992, opened 2020-05-14
Update BaseConnection.php

zl59503020:
Configuration file app\Config\Database item "failover" Effective

**Description**
When using DBDriver = MySQLi, if I configure the failover option, but if the main configuration database connection times out, an exception will be thrown directly and execution will stop. It does not attempt to use the configuration of failover items.

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -380,7 +380,7 @@ abstract class BaseConnection implements ConnectionInterface
 		} 
 		catch (\Throwable $e) 
 		{
-			log_message('error', 'connect database error: ' . $e->getMessage());
+			log_message('error', 'Error connecting to the database: ' . $e->getMessage());
 		}
 
 		// No connection resource? Check if there is a failover else throw an error
@@ -408,7 +408,7 @@ abstract class BaseConnection implements ConnectionInterface
 					} 
 					catch (\Throwable $e) 
 					{
-						log_message('error', 'connect database error: '.$e->getMessage());
+						log_message('error', 'Error connecting to the database: '.$e->getMessage());
 					}
 
 					// If a connection is made break the foreach loop
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -373,8 +373,12 @@ abstract class BaseConnection implements ConnectionInterface
 
 		$this->connectTime = microtime(true);
 
-		// Connect to the database and set the connection ID
-		$this->connID = $this->connect($this->pConnect);
+		try {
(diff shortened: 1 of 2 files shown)


MGatner:
I think this is a good change! Any chance we could get a test to demonstrate the issue? Also I think if you're trying for an all-inclusive catch statement it is better to use `catch (\Throwable $e)`

zl59503020:
file /app/Config/Database.php
`
	public $default = [
		'DSN'      => '',
		'hostname' => 'mydb.hostname.url',
		'username' => 'username',
		'password' => 'password',
		'database' => 'database',
		'DBDriver' => 'MySQLi',
		'DBPrefix' => '',
		'pConnect' => false,
		'cacheOn'  => false,
		'cacheDir' => '',
		'charset'  => 'utf8',
		'swapPre'  => '',
		'encrypt'  => false,
		'compress' => false,
		'strictOn' => false,
		'failover' => [
			[
				'hostname' => 'mydb.hostname.backup.url',
			]
		],
		'port'     => 3306,
	];
`

If I fail to connect to the database for the first time, I can also try to connect to the database using an alternate address.

system/Database/BaseConnection.php
```diff
@@ -373,8 +373,15 @@ public function initialize()
...
-		// Connect to the database and set the connection ID
-		$this->connID = $this->connect($this->pConnect);
+		try 
+		{
+			// Connect to the database and set the connection ID
+			$this->connID = $this->connect($this->pConnect);
+		} 
+		catch (\Throwable $e) 
+		{
+			log_message('error', 'connect database error: ' . $e->getMessage());
```
MGatner, line 383: Slight change to the wording, here and below:
`log_message('error', 'Error connecting to the database: ' . $e->getMessage());`

zl59503020: Okay thank you. Sorry, English is not my mother tongue, so some descriptions are wrong, I translated through Google.

zl59503020:
> This looks great - way simpler than my approach. Only one thing is missing - tests.
> 
> @zl59503020 you can simply copy tests from my PR: https://github.com/codeigniter4/CodeIgniter4/pull/3074/files#diff-9bf06a87308bcfd15e908922d54dcf29 - this way, we will make sure that the failover actually works.



> This looks great - way simpler than my approach. Only one thing is missing - tests.
> 
> @zl59503020 you can simply copy tests from my PR: https://github.com/codeigniter4/CodeIgniter4/pull/3074/files#diff-9bf06a87308bcfd15e908922d54dcf29 - this way, we will make sure that the failover actually works.


Ok, thank you

PR #2997, opened 2020-05-15
bug fix: ensure "BIGINT" to correctly applied to Postgresql Forge

samsonasik:
Previously, On calling Forge at PostgreSQL, like the following:

```php
<?php
$this->forge->addField([
	'id'     => [
	    'type'           => 'BIGINT',
	    'unsigned'       => true,
	    'auto_increment' => true,
	],
]);
```

It previously converted to integer instead of bigint. By this chage, it ensure correctly make bigint when we define "BIGINT".

**Checklist:**
- [x] Securely signed commits


samsonasik:
I've added unit test for it.

samsonasik:
travis build green :tada:

PR #3000, opened 2020-05-15
Add setSilent() method to URI class

michalsn:
**Description**
Using `$uri->setSilent()` method will disable throwing exceptions. May be handy for using with methods like `getSegment()`.

Ref: #2952, #2949

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/libraries/uri.rst
+++ b/user_guide_src/source/libraries/uri.rst
@@ -248,3 +248,17 @@ Finally, you can retrieve an array of all of the segments::
 		1 => '15',
 		2 => 'profile'
 	]
+
+===========================
+Disable Throwing Exceptions
+===========================
+
+By default, some methods of this class may throw an exception. If you want to disable it, you can set a special flag
+that will prevent throwing exceptions.
+::
+
+	// Disable throwing exceptions
+	$request->setSilent();
+
+	// Enable throwing exceptions
+	$request->setSilent(false);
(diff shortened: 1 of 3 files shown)


user_guide_src/source/libraries/uri.rst
```diff
@@ -248,3 +248,17 @@ Finally, you can retrieve an array of all of the segments::
...
+===========================
+Disable Throwing Exceptions
+===========================
+
+By default, some methods of this class may throw an exception. If you want to disable it, you can set a special flag
+that will prevent throwing exceptions.
+::
+
+	// Disable throwing exceptions
+	$request->setSilent();
```
jreklund, line 261: Should be $uri->setSilent();

user_guide_src/source/libraries/uri.rst
```diff
@@ -248,3 +248,17 @@ Finally, you can retrieve an array of all of the segments::
...
+
+	// Enable throwing exceptions
+	$request->setSilent(false);
```
jreklund, line 264: Should be $uri->setSilent(false);

jreklund (changes requested):
Looks good, but I think we should reference it as $uri instead.

user_guide_src/source/libraries/uri.rst
michalsn: Thanks.

user_guide_src/source/libraries/uri.rst
michalsn: Done.

michalsn:
Oh great... now it not mergeable... I wonder what went wrong.

michalsn:
I don't know how to resolve this, so I'm closing this PR in favor of a new one: #3005

PR #3001, opened 2020-05-16
Don't add params to IncomingRequest in FeatureTestCase, per discussion in #2946

lonnieezell:
`$params` from GET and POST requests were being passed as the body to the `IncomingRequest`, which doesn't match up with how things actually work. Since `$params` were already being set in the globals correctly,this tweaks it all to work better and work with `$params` for GET requests in either `$params` or in the URI itself (i.e. `/home?foo=bar`).

### DIFF
--- a/system/Test/FeatureTestCase.php
+++ b/system/Test/FeatureTestCase.php
@@ -164,7 +164,7 @@ class FeatureTestCase extends CIDatabaseTestCase
 		$_SESSION                  = [];
 		$_SERVER['REQUEST_METHOD'] = $method;
 
-		$request = $this->setupRequest($method, $path, $params);
+		$request = $this->setupRequest($method, $path);
 		$request = $this->populateGlobals($method, $request, $params);
 
 		// Make sure the RouteCollection knows what method we're using...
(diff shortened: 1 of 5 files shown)


lonnieezell:
@MGatner good catch! Will do. Thanks.

system/Test/FeatureTestCase.php
```diff
@@ -289,16 +289,15 @@ public function options(string $path, array $params = null)
...
 	 * @return \CodeIgniter\HTTP\IncomingRequest
 	 */
-	protected function setupRequest(string $method, string $path = null, array $params = null): IncomingRequest
+	protected function setupRequest(string $method, string $path = null): IncomingRequest
 	{
 		$config = config(App::class);
 		$uri    = new URI($config->baseURL . '/' . trim($path, '/ '));
 
-		$request      = new IncomingRequest($config, clone($uri), $params, new UserAgent());
+		$request      = new IncomingRequest($config, clone($uri), null, new UserAgent());
```
ChoiBU, line 300: How can I send body?
ex) featureTest
// GET /test-url?type=a
$this->get('/test-url', ['type' => 'a']);
or
// POST /test-url
// request body (type=a)
$this->post('/test-url' ['type => 'a']);

lonnieezell: Those functions still work like they always did. The $params array wasn't doing anything when sent to IncomingRequest in the constructor anyway. It's handled via a different method.

ByeongUkChoi: How can I send JSON body?

Issue #3006, opened 2020-05-16
Image reorient with exif


paul45:
hello
my code is
`Services::image('imagick')
                    ->withFile(xxxxxx)
                    ->reorient()
                    ->save(xxx,85);`

when i upload an image from my android the image appears upside down

i think it is a issue on CodeIgniter\Images\Handlers\BaseHandler in methode reorient()
i think the rotation is not on the good direction

PR #3007, opened 2020-05-16
Fixe rotation direction on reorient()

paul45:
**Description**
Fixe #3006
I think the rotation from exif is not in the good direction 

Change the rotation direction on image reorient()

Maybe reorient methode should be add in the doc

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Images/Handlers/BaseHandler.php
+++ b/system/Images/Handlers/BaseHandler.php
@@ -533,15 +533,15 @@ abstract class BaseHandler implements ImageHandlerInterface
 				return $this->rotate(180)
 								->flip('horizontal');
 			case 5:
-				return $this->rotate(270)
-								->flip('horizontal');
-			case 6:
-				return $this->rotate(270);
-			case 7:
 				return $this->rotate(90)
 								->flip('horizontal');
-			case 8:
+			case 6:
 				return $this->rotate(90);
+			case 7:
+				return $this->rotate(270)
+								->flip('horizontal');
+			case 8:
+				return $this->rotate(270);
 			default:
 				return $this;
 		}
--- a/system/Images/Handlers/BaseHandler.php
+++ b/system/Images/Handlers/BaseHandler.php
@@ -533,15 +533,15 @@ abstract class BaseHandler implements ImageHandlerInterface
 				return $this->rotate(180)
 								->flip('horizontal');
 			case 5:
-				return $this->rotate(90)
-								->flip('horizontal');
-			case 6:
-				return $this->rotate(90);
-			case 7:
 				return $this->rotate(270)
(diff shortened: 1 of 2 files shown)


MGatner:
I have no idea how this is supposed to work. Can you share any resources that indicate what the correct process should be?

jreklund:
Haven't checked the back-bone of rotate(), but GD and ImageMagick uses to different values for the same rotation. So I guess @paul45 are using ImageMagick? Or it's wrong in GD. xD

paul45:
@jreklund i use ImageMagick but i think it is the same in GD because the reorient() method is in the basehandler.php

@MGatner 
https://www.impulseadventure.com/photo/exif-orientation.html

for exemple if the orientation code from exif is 6 we need rotate 90° in CW

jreklund:
Okay, then we can't accept this as is. As GD and ImageMagick uses different values for the same thing. I wrote a similar rotation function but based off Imagick::rotateImage, and had to apply a conversion based on [this post](https://www.php.net/manual/en/imagick.rotateimage.php#119184), in another project.

I can't see any conversion being made in CodeIgniter at the moment, so that function needs to be as well. Personally I applied the conversion in ImageMagick, as the GD are the main one used.

paul45:
ok you are right @jreklund the actual reorient() is good for GD but not for ImageMagick 

maybe the reorient() method will be put in GDHandler.php and in ImageMagickHandler.php(with this change) instead of BaseHandler.php

MGatner:
@paul45 I think that is the right way to go. Would you be willing to tackle this?

paul45:
an other way is to add a optional direction parameter to rotate() in both class for select CW or CCW
and use this parameter in reorient()

jreklund:
Removing reorient from BaseHandler would be a breaking change if I'm not wrong, as it's a public function. So it would need to be converted in either GD or ImageMagick. 

Hard to tell if someone are depending on the current orientation values at the moment, by extending the BaseHandler. So the safe bet are to converting it in ImageMagick I'm afraid.

MGatner:
@paul45 consensus seems to be that we move this to `ImageMagick` and leave `BaseHandler`'s version in place. I think a test to demonstrate proper rotation would help me a lot, since I only look at this every 3 weeks 🙄

paul45:
maybe a solution without duplicate code, with protected $rotateAngle like @jreklund said should be more elegant ?
i don't know how to make test sorry

jreklund:
It's less duplicated code at least, if everyone are fine with adding two new variables. :-)

michalsn:
I will merge this at the current state and add some tests later today.

PR #3008, opened 2020-05-17
Add default value to getSegment() method

michalsn:
**Description**
This PR adds a second parameter to the `getSegment()` method, so we can change the default value that will be returned when there is no segment set. Usually, this feature will be used along with `setSilent()`, so that the exception will be not thrown.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/HTTP/URITest.php
+++ b/tests/system/HTTP/URITest.php
@@ -105,7 +105,6 @@ class URITest extends \CodeIgniter\Test\CIUnitTestCase
 		$this->assertEquals('', $uri->getSegment(3));
 
 		$this->assertEquals(2, $uri->getTotalSegments());
-		$this->assertEquals(['path', 'to'], $uri->getSegments());
 	}
 
 	//--------------------------------------------------------------------
--- a/tests/system/HTTP/URITest.php
+++ b/tests/system/HTTP/URITest.php
@@ -74,6 +74,56 @@ class URITest extends \CodeIgniter\Test\CIUnitTestCase
 
 	//--------------------------------------------------------------------
 
+	public function testSegmentOutOfRangeWithDefaultValue()
+	{
+		$this->expectException(HTTPException::class);
+		$url = 'http://abc.com/a123/b/c';
+		$uri = new URI($url);
+		$uri->getSegment(22, 'something');
+	}
+
+	//--------------------------------------------------------------------
+
+	public function testSegmentOutOfRangeWithSilentAndDefaultValue()
+	{
+		$url = 'http://abc.com/a123/b/c';
+		$uri = new URI($url);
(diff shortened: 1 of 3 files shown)


tests/system/HTTP/URITest.php
```diff
@@ -74,6 +74,56 @@ public function testSegmentOutOfRangeWithSilent()
...
+		$uri->setSilent();
+
+		$this->assertEquals(['path', 'to'], $uri->getSegments());
+		$this->assertEquals('path', $uri->getSegment(1));
+		$this->assertEquals('to', $uri->getSegment(2, 'different'));
+		$this->assertEquals('script', $uri->getSegment(3, 'script'));
+		$this->assertEquals('', $uri->getSegment(3));
+
+		$this->assertEquals(2, $uri->getTotalSegments());
+		$this->assertEquals(['path', 'to'], $uri->getSegments());
```
jreklund, line 108: Duplicated assertEquals.

user_guide_src/source/libraries/uri.rst
```diff
@@ -234,6 +234,19 @@ what the values of the segments are. The segments start at 1 being the furthest
 		echo $request->uri->getSegment(2);
 	}
 
+You can also get the different default value for a given segment. Default value is set by using the second parameter of the ``getSegment()`` method.
```
jreklund, line 237: "another default" (instead of the different) or "You can also set a different default value for a particular segment by using the second parameter of the ``getSegment()`` method."

Should we add a note about what the actual default value are? Can't seem to find it in this document.

user_guide_src/source/libraries/uri.rst
```diff
@@ -234,6 +234,19 @@ what the values of the segments are. The segments start at 1 being the furthest
...
+::
+	// URI = http://example.com/users/15/profile
+
+	// will print 'profile'
+	echo $request->uri->getSegment(3, 'foo');
```
jreklund, line 242: Change into $uri->, also change the rest of the code (that's not yours in URI Segments) into the same layout for consistency.

michalsn:
Changes are ready.

user_guide_src/source/libraries/uri.rst
michalsn: I mentioned the default value.

user_guide_src/source/libraries/uri.rst
michalsn: Done.

tests/system/HTTP/URITest.php
michalsn: Fixed.

MGatner:
Thanks you two. I know this was a much bigger process than this PR lets on!

PR #3009, opened 2020-05-18
Fix. toodebug disabled, view debug comment is show.

### LINKED ISSUE #3002: Bug: Filters 'except' option not removing DebugToolbar comment for view
**Describe the bug**
[`except`](https://codeigniter4.github.io/userguide/incoming/filters.html?highlight=except#globals) option in filters disables the [Debug Toolbar](https://codeigniter4.github.io/userguide/testing/debugging.html?highlight=toolbar#the-debug-toolbar) but adds the comment `<!-- DEBUG-VIEW START -->`. This comment breaks the css when I m trying to load the view as `CSS` with `$this->response->setContentType('text/css')`. May causes the problem for other content types also.

**CodeIgniter 4 version**
v4.0.3

**Affected module(s)**
CodeIgniter\View

Expected that the comment should not be added when `URI` available in `except` filter.

**Filters.php**
```
public $globals = [
	'before' => [],
	'after'  => [
		'toolbar' => ['except' => 'style'],
		//'honeypot'
	],
];
```
https://github.com/codeigniter4/CodeIgniter4/blob/9bbc99136dce5b57baf8d27064ce85b9c2ee619c/system/View/View.php#L262-L282

**Context**
 - OS: Windows 7
 - Web server: Apache/2.4.41
 - PHP version: 7.3.9

Instrye:
- fix #3002 tooldebug except, but view debug comment is show.
- add filters except example

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -15,11 +15,16 @@ class Filters extends BaseConfig
 	// Always applied before every request
 	public $globals = [
 		'before' => [
+			'toolbar' => [
+				'except' => 'api/*',
+			],
 			//'honeypot'
 			// 'csrf',
 		],
 		'after'  => [
-			'toolbar',
+			'toolbar' => [
+				'except' => 'api/*',
+			],
 			//'honeypot'
 		],
 	];
--- a/app/Config/Toolbar.php
+++ b/app/Config/Toolbar.php
@@ -16,6 +16,9 @@ class Toolbar extends BaseConfig
 	| toolbarMaxHistory = Number of history files, 0 for none or -1 for unlimited
 	|
 	*/
+
+	public $enabled = false;
+
 	public $collectors = [
 		\CodeIgniter\Debug\Toolbar\Collectors\Timers::class,
 		\CodeIgniter\Debug\Toolbar\Collectors\Database::class,
--- a/system/Filters/DebugToolbar.php
+++ b/system/Filters/DebugToolbar.php
@@ -57,6 +57,7 @@ class DebugToolbar implements FilterInterface
 	 */
 	public function before(RequestInterface $request)
 	{
+		config('toolbar')->enabled = true;
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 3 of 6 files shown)


app/Config/Toolbar.php
```diff
@@ -16,6 +16,9 @@ class Toolbar extends BaseConfig
 	| toolbarMaxHistory = Number of history files, 0 for none or -1 for unlimited
 	|
 	*/
+
+	public $enabled = false;
```
lonnieezell, line 20: Should be enabled by default so there's one less step to do to get it working.

app/Config/Filters.php
```diff
@@ -15,11 +15,16 @@ class Filters extends BaseConfig
 	// Always applied before every request
 	public $globals = [
 		'before' => [
+			'toolbar' => [
+				'except' => 'api/*',
```
lonnieezell, line 19: Please remove the exceptions. They make an assumption about projects that the framework shouldn't make.

system/Filters/DebugToolbar.php
```diff
@@ -57,6 +57,7 @@ class DebugToolbar implements FilterInterface
 	 */
 	public function before(RequestInterface $request)
 	{
+		config('toolbar')->enabled = true;
```
lonnieezell, line 60: Setting it here makes the setting pointless since it will always be set when you get into the View class, doesn't it?

tests/system/Filters/DebugToolbarTest.php
```diff
@@ -23,6 +23,16 @@ protected function setUp(): void
...
 	//--------------------------------------------------------------------
 
+	public function testDebugToolbarFilterExcept()
+	{
+		$data     = [
+			'testString' => 'bar',
+			'bar'        => 'baz',
+		];
+		$expected = '<h1>bar</h1>';
+		$this->assertEquals($expected, view('\Tests\Support\View\Views\simple', $data, []));
```
lonnieezell, line 33: This test won't actually run it through the controller and fire the filters themselves, so doesn't seem to test the new stuff you've added.

Instrye:
> Good catch! Unfortunately, i don't think your fix will work for reasons listed in the other comments.
> 
> It seems that a better option would be to add a new method to the Filters class that allows arbitrary checking of whether a filter `shouldRun($filterName)` would be a better fix here. That can then be used in the View class to determine whether or not to insert the comments.

This is a good way, I will try as soon as possible

PR #3010, opened 2020-05-18
fix. toolbar disable,view comment not remove

### LINKED ISSUE #3002: Bug: Filters 'except' option not removing DebugToolbar comment for view

Instrye:
- fix #3002. toolbar disable,but view comment not remove.
- ~~delete filter argumets support. because, filter muset `implements FilterInterface`. `before` only receive request. so, if filter argumetns send to `before`, will get an error.~~
- Alias resolution is no longer done in the `run` method, `Filters::$filtersClass` save alias to class
**Checklist:**
- FiltersTest. append `aliases` to config
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner:
Thank you @Instrye - this looks like quite a bit of work and some good changes. I don't have my head around it, I'll have to try another day. In the meantime, anybody else willing to give it a review?

paulbalandan:
This needs to be revised to incorporate changes in `FilterInterface`.

samsonasik:
@MGatner could you re-check it?

system/View/View.php
```diff
@@ -269,7 +269,7 @@ public function render(string $view, array $options = null, bool $saveData = nul
 
 		$this->logPerformance($this->renderVars['start'], microtime(true), $this->renderVars['view']);
 
-		if ($this->debug && (! isset($options['debug']) || $options['debug'] === true))
+		if (($this->debug && (! isset($options['debug']) || $options['debug'] === true)) && in_array('CodeIgniter\Filters\DebugToolbar', service('filters')->getFiltersClass()['after']))
```
samsonasik, line 272: can use strict in_array check: 

```suggestion
              if (($this->debug && (! isset($options['debug']) || $options['debug'] === true)) && in_array('CodeIgniter\Filters\DebugToolbar', service('filters')->getFiltersClass()['after'], true))
```

tests/system/Filters/FiltersTest.php
```diff
@@ -913,4 +998,33 @@ public function testFilterAlitasMultiple()
 		$this->assertEquals('http://exampleMultipleCSP.com', $request->csp);
 	}
 
+	public function testFilterClass()
+	{
+		$config  = [
+			'aliases' => [
+				'multipeTest' => [
```
MGatner, line 1005:
```suggestion
				'multipleTest' => [
```

tests/system/Filters/FiltersTest.php
```diff
@@ -913,4 +998,33 @@ public function testFilterAlitasMultiple()
...
+		$config  = [
+			'aliases' => [
+				'multipleTest' => [
+					'CodeIgniter\Filters\fixtures\Multiple1',
+					'CodeIgniter\Filters\fixtures\Multiple2',
+				],
+			],
+			'globals' => [
+				'after' => [
+					'multipeTest',
```
MGatner, line 1012:
```suggestion
					'multipleTest',
```

MGatner:
A few issues failing the static analysis check, but otherwise this looks good. Waiting on Travis. @samsonasik should we merge despite the CS failures? Or resolve those first?

samsonasik:
@MGatner I think code static analysis check need to incorporated first: 

```
  Line   system/Filters/Filters.php                                             
  157    Access to an undefined property                                        
         CodeIgniter\Filters\Filters::$argumentsClass.                          
  183    Access to an undefined property                                        
         CodeIgniter\Filters\Filters::$argumentsClass.                          
  338    Access to an undefined property                                        
         CodeIgniter\Filters\Filters::$argumentsClass.                          
  495    Return typehint of method                                              
         CodeIgniter\Filters\Filters::processAliasesToClass() has invalid type  
         CodeIgniter\Filters\type.                                              
  497    Method CodeIgniter\Filters\Filters::processAliasesToClass() should     
         return CodeIgniter\Filters\type but return statement is missing.      
```

system/Filters/Filters.php
```diff
@@ -60,6 +60,11 @@ class Filters
 		'after'  => [],
 	];
 
+	protected $filtersClass = [
+		'before' => [],
+		'after'  => [],
+	];
+
```
samsonasik, line 67: new property can be added:

```suggestion

           /** @var array */
          protected $argumentsClass;

```

system/Filters/Filters.php
```diff
@@ -474,6 +477,36 @@ protected function processFilters(string $uri = null)
 		}
 	}
 
+	/**
+	 * filter alias to class
+	 *
+	 * @return type
```
samsonasik, line 483: docblock can be updated to:

```suggestion
       * @return void
       * @throws FilterException
```

michalsn (changes requested):

MGatner:
@Instrye It looks like one more round of tweaks, if you're available for those.

Instrye:
😂I want to provide quality phpdoc, But my English level hinders me.

system/Filters/Filters.php
```diff
@@ -334,16 +343,29 @@ public function enableFilter(string $name, string $when = 'before')
...
 		}
 
+		if (is_array($this->config->aliases[$name]))
+		{
+			$classNames = $this->config->aliases[$name];
+		}
+		else
+		{
+			$classNames = [$this->config->aliases[$name]];
+		}
```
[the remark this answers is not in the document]
Instrye: thanks

michalsn (approved):
Looks good, thanks! 

You only have to resolve some conflicts with the develop branch and I think we're good to go.

Instrye:
@michalsn Everything's ready.

michalsn:
Thanks!

PR #3012, opened 2020-05-18
Replace Logger::cleanFileNames for DRY code

paulbalandan:
**Description**
For DRY code, replace logic of `Logger::cleanFileNames` with that of `Exceptions::cleanPath`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Log/Logger.php
+++ b/system/Log/Logger.php
@@ -39,6 +39,7 @@
 
 namespace CodeIgniter\Log;
 
+use CodeIgniter\Debug\Exceptions;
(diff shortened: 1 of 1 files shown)


lonnieezell:
I appreciate this, but was working through some large error handling improvements a couple of nights ago, and the Exceptions::cleanPath method is going to go away. It is being moved into system/Common.php as a new helper `clean_path`. I could try to split push that up late tonight, unless you want to implement that in a PR before I have a chance to.

PR #3027, opened 2020-05-21
Use fully-qualified namespace for PHP in Session

dafriend:
**Description**

Code better optimized by calling with fully-qualified PHP function names. 
For instance, the PHP function time in the following

     $_SESSION['__ci_last_regenerate'] = \time();

It's not a giant optimization, but it's more than a micro-optimization. 

Existing unit test `SessionTest.php` unchanged and all tests pass.
User guide not affected by these changes.

### DIFF
--- a/system/Session/Handlers/DatabaseHandler.php
+++ b/system/Session/Handlers/DatabaseHandler.php
@@ -111,12 +111,12 @@ class DatabaseHandler extends BaseHandler implements \SessionHandlerInterface
 		$this->db = Database::connect($this->DBGroup);
 
 		// Determine Database type
-		$driver = strtolower(get_class($this->db));
-		if (strpos($driver, 'mysql') !== false)
+		$driver = \strtolower(\get_class($this->db));
+		if (\strpos($driver, 'mysql') !== false)
 		{
 			$this->platform = 'mysql';
 		}
-		elseif (strpos($driver, 'postgre') !== false)
+		elseif (\strpos($driver, 'postgre') !== false)
 		{
 			$this->platform = 'postgre';
 		}
@@ -160,12 +160,12 @@ class DatabaseHandler extends BaseHandler implements \SessionHandlerInterface
 	{
 		if ($this->lockSession($sessionID) === false)
 		{
-			$this->fingerprint = md5('');
+			$this->fingerprint = \md5('');
 			return '';
 		}
 
 		// Needed by write() to detect session_regenerate_id() calls
-		if (is_null($this->sessionID))
+		if (\is_null($this->sessionID))
 		{
 			$this->sessionID = $sessionID;
 		}
@@ -187,7 +187,7 @@ class DatabaseHandler extends BaseHandler implements \SessionHandlerInterface
 			// ID regeneration, so we need to explicitly set this to
 			// FALSE instead of relying on the default ...
 			$this->rowExists   = false;
-			$this->fingerprint = md5('');
+			$this->fingerprint = \md5('');
 
 			return '';
 		}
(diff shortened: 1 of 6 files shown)


samsonasik:
there is php-cs-fixer rules for that (native_function_invocation) to apply to entire code base I think

lonnieezell:
Can you point to something showing this actually has a benefit? Honestly, it looks horrid and I would hate to have to prefix every native PHP command.... especially if it's not worth it.

dafriend:
Right, it's more important to look good than to be good. 
OK, that's snarky and I apologize. But really?

If you want reasons beyond removing any ambiguity about which function to use and preventing name collision issues consider:

 - The big kids do it (Laravel, Symphony, Laminas)

  - For the following to happen the functions need to use a fully qualified namespace. Since PHP 7.0, some PHP functions are replaced directly by opcode (as opposed to function calls via opcode). This optimization is, more or less, the equivalent to the c inline function. For PHP functions that cannot be optimized that way, the NS will be resolved immediately eliminating the need to check the current NS for the function.

 - See benchmarks at http://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/ 

 - Documented enhancement via fqns https://github.com/scssphp/scssphp/issues/96

IMHO, this should have been part of the coding standard the minute you decided to utilize namespace.

As @samsonasik points out there is a php-cs-fixer rule that could be added workflow that would eventually get the job done.

lonnieezell:
@dafriend Thank you for actually providing additional info as requested instead of just dropping a line of snark and leaving. I was asking because I was not aware of the amount and, if the performance increase was <1.0% I couldn't see it as being worth it, honestly. With the results you've pointed out being between 2%-8% that makes it worth it. 

Though it needs to be automated to be worthwhile. I don't suppose one of you two is willing to submit a PR to https://github.com/codeigniter4/coding-standard to add that?

MGatner:
I'm approaching obsessive about my code so take it with a grain of salt, but I really don't like the preceding backslash. The results have really big ranges, which is understandable since this is hard to test, but would an alternative be to identify high-density function-call classes and implement one of the alternate solutions that article mentions? `use function foo;` or `declare(no_dynamic_functions=1)`? Probably not worth the time spent, but I feel better having my opinion recorded. :)

dafriend:
@MGatner What is it about a preceding backslash that you "really don't like"?

I don't understand what your statement, "The results have really big ranges," refers to.

michalsn:
It looks like we can only gain something with this change and it's easy to achieve as @samsonasik mentioned. If this can be added to our coding standard, then I think this change should take place.

samsonasik:
in laminas, we use `use function x` instead of adding backslash to the native functions.

lonnieezell:
I think the `use function x` solution is good too. Then we can address it per file where it makes sense. @dafriend would you mind updating these changes to that style?

dafriend:
Can do.

dafriend:
I have started to make the conversion to `use function fn` and find that, in some cases, it is causing me a **lot** of cognitive dissonance. It seems really silly to add `use function strtolower;` so that we don't have to put a backslash before the one and only call to `strtolower()` in the script.

It is happening quite a lot that only one or two calls are made to a PHP "root" namespace function in a class file.

How about a compromise where we implement `use function fn` only when there is more than one call to `fn()` in the script?

lonnieezell:
I think that's fine for now.

system/Session/Handlers/DatabaseHandler.php
```diff
@@ -43,6 +43,10 @@
 use CodeIgniter\Database\BaseConnection;
 use CodeIgniter\Session\Exceptions\SessionException;
 use Config\Database;
+use function strpos;
+use function md5;
+use function time;
+use function base64_encode;
```
samsonasik, line 49: I think sorted will be better.

MGatner:
@dafriend I still think this is worth doing. Do you have a separate PR in the works? Can we resurrect this one?

dafriend:
Thanks for your appreciation. However, the PR was sitting here for months and the "deciders" apparently didn't find it worth merging. It's hard not to feel like you've proposed a dumb-a%% idea and rather than hang there as a badge of dumbness I withdraw the contribution - meager as it might have been.

MGatner:
Sorry about the delays. We’ve gotten pretty backlogged on PRs and nobody is reviewing ones further down the stack. I’ve been far too busy at my day job, getting university computer labs ready for students amid COVID. I would like to do a pass at all the backlogged PRs next time I have some CI4 time so if you want to reopen this I think it is a good thing for the framework.

PR #3031, opened 2020-05-23
Fixes #3028 add ability to manipulate string result of view->renderSection()

### LINKED ISSUE #3028: Bug:  Not possible manipulate results of renderSection() on View Layouts
**Describe the bug**
It is not possible to manipulate the string result of a View Layout `$this->renderSection()` without using ob_* functions due the void function works with `echo` and not `return`.

https://github.com/codeigniter4/CodeIgniter4/blob/6d019e5354629fa7c1e293c2f463d91004c6e26c/system/View/View.php#L496-L510

This example https://codeigniter4.github.io/userguide/outgoing/view_layouts.html#creating-a-layout show a echo but it is not necessary.

**CodeIgniter 4 version**
4.0.3

**Affected module(s)**
View Layouts

Expect to can save result of `renderSection()` in a variable.

**Context**
 - OS: Ubuntu
 - Web server [PHP built in]
 - PHP version [7.4]

samsonasik:
Fixes #3028 . add ability `View::renderSection()` to returns string.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/View/View.php
+++ b/system/View/View.php
@@ -493,20 +493,21 @@ class View implements RendererInterface
 	 *
 	 * @param string $sectionName
 	 */
-	public function renderSection(string $sectionName)
+	public function renderSection(string $sectionName) : string
 	{
 		if (! isset($this->sections[$sectionName]))
 		{
-			echo '';
-
-			return;
+			return '';
 		}
 
+		$result = '';
 		foreach ($this->sections[$sectionName] as $key => $contents)
 		{
-			echo $contents;
+			$result .= $contents;
 			unset($this->sections[$sectionName][$key]);
 		}
+
+		return $result;
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


samsonasik:
travis build green :tada:

system/View/View.php
```diff
@@ -493,20 +493,21 @@ public function endSection()
 	 *
 	 * @param string $sectionName
 	 */
-	public function renderSection(string $sectionName)
+	public function renderSection(string $sectionName) : string
```
samsonasik, line 496: while in documentation, the `renderSection()` is called as : 

```php
    <?= $this->renderSection('content') ?>
```

It may cause a bc break to user when user just use as is:

```php
 <?php $this->renderSection('content') ?>
```

It may require a 2nd parameter `$echoed` = true I think. //cc @natanfelles

samsonasik: I just added 2nd parameter `$echoed` default to true to avoid bc break.

PR #3033, opened 2020-05-24
Fix default type of UserAgent::$agent

### LINKED ISSUE #3029: Bug: Fatal error when no user-agent is available
**Describe the bug**
If there is no user-agent available, an error will happen.

**CodeIgniter 4 version**
v4.0.3

**Affected module(s)**
HTTP\UserAgent

1- Add a method to the Home controller
2- Put `var_dump((string)\Config\Services::request()->getUserAgent());` in the method you just created.
3- call that method from the command line(for instance)
4- I expect to see `string(0) ""` or `NULL` without seeing any exceptions

The actual result:
`PHP Fatal error:  Method CodeIgniter\HTTP\UserAgent::__toString() must not throw an exception, caught TypeError: Return value of CodeIgniter\HTTP\UserAgent::getAgentString() must be of the type string, null returned
`

**Context**
 - OS: Fedora 31
 - Web server -
 - PHP version 7.3.17

paulbalandan:
**Description**
Fixes #3029 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/UserAgent.php
+++ b/system/HTTP/UserAgent.php
@@ -132,7 +132,7 @@ class UserAgent
 	 *
 	 * @param null|\Config\UserAgents $config
 	 */
-	public function __construct(?UserAgents $config = null)
+	public function __construct(UserAgents $config = null)
 	{
 		$this->config = $config ?? new UserAgents();
 
--- a/system/HTTP/UserAgent.php
+++ b/system/HTTP/UserAgent.php
@@ -51,7 +51,7 @@ class UserAgent
 	 *
 	 * @var string
 	 */
-	protected $agent;
+	protected $agent = '';
 
 	/**
 	 * Flag for if the user-agent belongs to a browser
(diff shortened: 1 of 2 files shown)


system/HTTP/UserAgent.php
```diff
@@ -130,14 +130,11 @@ class UserAgent
 	 *
 	 * Sets the User Agent and runs the compilation routine
 	 *
-	 * @param null $config
+	 * @param null|\Config\UserAgents $config
 	 */
-	public function __construct($config = null)
+	public function __construct(?UserAgents $config = null)
```
lonnieezell, line 135: I don't think the `?` is needed here since the default value is null.

paulbalandan: I know that `= null` here already implies nullability of the type but I have read somewhere that it is perfectly valid.

paulbalandan: This is from the original [RFC](https://wiki.php.net/rfc/nullable_types):

image

But if you're against it, I can still remove that. :)

element-code: I would prefer to use the explicit `?` in case the default changes. On the other hand its redundant.

paulbalandan: Basing on other code within the codebase, the `?` is indeed not used. Changing..

PR #3035, opened 2020-05-26
Fix CLI output colors

mostafakhudair:
Remove dark_blue it uses light_blue color effect
correct light_yellow and yellow colors


mostafakhudair:
I got your point, for now i just revert changes and correct dark_blue color with right effect

PR #3037, opened 2020-05-27
Added SameSite attribute to CSRF cookie

erikkloeze:
**Description**
In order to provide an additional layer of defense against cross-site request forgery (CSRF) attacks, the SameSite attribute is added to the CSRF cookie.

**Checklist:**
- [ ] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -239,7 +239,7 @@ class App extends BaseConfig
 	| CSRFTokenName   = The token name
 	| CSRFHeaderName  = The header name
 	| CSRFCookieName  = The cookie name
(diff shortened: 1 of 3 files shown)


app/Config/App.php
```diff
@@ -239,13 +239,15 @@ class App extends BaseConfig
 	| CSRFTokenName   = The token name
 	| CSRFHeaderName  = The header name
 	| CSRFCookieName  = The cookie name
+	| CSRFCookieSameSite  = The cookie SameSite attribute (none, Lax or Strict)
```
lonnieezell, line 242: Case for these should match what a developer is expected to use. You're using all lowercase below, but these are listed out as Initial Caps.

Issue #3041, opened 2020-05-27
Bug: Route redirection not working


MichalPB1:
**Describe the bug**
After using 
```
$routes->addRedirect('example', 'profile/dashboard');
```
I get an error 
```
key() expects parameter 1 to be array, string given
```
but if the second param I set as named route then everything is working properly. The problem exists only when the second param is a URI

**CodeIgniter 4 version**
4.0.2

**Affected module(s)**
```
Router/Router.php at line 430
```
I not sure but here I found some changes for this line https://github.com/codeigniter4/CodeIgniter4/pull/2481/files#diff-d2374d65fd714c18dbee9f5adaa8dd6cL488

**Expected behavior, and steps to reproduce if appropriate**
 Redirection working properly for the second param as a URI .

**Context**
 - OS: Ubuntu
 - PHP version 7.3

Instrye:
This problem is because `profile/dashboard` not set in `$routes`. I will fix.

MichalPB1:
@Instrye thx for your answer :) but it not true. That route is already defined and if I change it to route name everything works.

I think that redirect should work even that URI is not defined - you don't think so? We could have a situation that redirection to not existed URI redirect it to the next URI.

MichalPB1:
It's more complicated :) It looks like redirection doesn't work for addresses with / eg `checkout/shoppingCart`, `profile/orders` when route name is defined

Now the redirection works
```
$routes->addRedirect('asdasd', 'checkout/shoppingCart', 301);

$routes->get('shoppingCart', 'CheckoutController::shoppingCart');
```
Now the redirection does not work:
```
$routes->addRedirect('asdasd', 'checkout/shoppingCart', 301);

$routes->get('shoppingCart', 'CheckoutController::shoppingCart', ['as' => 'checkout.shoppingCart']);
```

Instrye:
I just use `$routes->add` to test, This looks bad.

Regarding your second question, I think you are right, I have made a change

Instrye:
> It's more complicated :) It looks like redirection doesn't work for addresses with / eg `checkout/shoppingCart`, `profile/orders` when route name is defined
> 
> Now the redirection works
> 
> ```
> $routes->addRedirect('asdasd', 'checkout/shoppingCart', 301);
> 
> $routes->get('shoppingCart', 'CheckoutController::shoppingCart');
> ```
> 
> Now the redirection does not work:
> 
> ```
[...]

PR #3044, opened 2020-05-29
Fix. router no set, Redirect can't work

Instrye:
Fix. #3041

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Router/RouteCollectionTest.php
+++ b/tests/system/Router/RouteCollectionTest.php
@@ -970,10 +970,11 @@ class RouteCollectionTest extends \CodeIgniter\Test\CIUnitTestCase
 	{
 		$routes = $this->getCollector();
 
-		$routes->addRedirect('users', 'Users::index', 307);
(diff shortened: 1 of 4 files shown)


tests/system/Router/RouteCollectionTest.php
```diff
@@ -970,10 +970,11 @@ public function testAddRedirect()
 	{
 		$routes = $this->getCollector();
 
-		$routes->addRedirect('users', 'Users::index', 307);
+		//The second parameter is either the new URI to redirect to, or the name of a named route.
+		$routes->addRedirect('users', 'users/index', 307);
```
samsonasik, line 974: I think instead of modify existing test, it needs new tests for new allowed usage?

Instrye: This not a new allowed usage. The test is a mistak,It should not use Class::method.This is not what we accept

PR #3049, opened 2020-05-30
Fixed $schema default attribute

lazharous:
**Description**
The default search_path for PostgreSQL is "$user", public.
Putting the default schema to public removed the access to $user.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


lonnieezell:
I don't know enough about Postgres to know if this is better honestly. I do know it breaks a few tests because the migrations table doesn't show up in a few for some reason. 

@michalsn Was it you that had more Postgres knowledge?

lazharous:
> I don't know enough about Postgres to know if this is better honestly. I do know it breaks a few tests because the migrations table doesn't show up in a few for some reason.
> 
> @michalsn Was it you that had more Postgres knowledge?

I am using CodeIgniter since version 2, on different production websites, always with Postgres as the database. For what it counts, I can say that setting the schema variable to "public" overwrites the default Postgres setting that enables access to 2 schema, the "public" and the one with the same name as the user logged in.

For example I am currently using a Postgres database with 3 schemas, with 3 different users and I can only connect to "public" if that variable is set to "public".

Waiting for someone to shed light over the testing errors, I am not used to Travis.

MGatner:
If @pstef is around he would probably know.

pstef:
I don't know if it's better, either - that depends on the use case and I haven't looked at how this is used at all. These two random statements are true regardless: you can always change search_path to something else during the connection session, and schema public is not guaranteed to exist (in fact, I often drop it).

michalsn:
Historically speaking, "public" was always the default value for the `schema` in CodeIgniter. 

* Changing this now could possibly break someone's application. 
* If we want to use different schema we can always override default value via config file or ENV.

lonnieezell:
Sounds like this should be dropped then as it's a breaking change and, as pointed out, it can always be set via config.

PR #3057, opened 2020-06-02
Fix. redirect custom HTTP Code not work

Instrye:
From #3041
Thanks @MichalPB1 's suggest

Redirect HTTP Code can't work. when method is get, it always set as 307.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/CodeIgniterTest.php
+++ b/tests/system/CodeIgniterTest.php
@@ -360,6 +360,56 @@ class CodeIgniterTest extends \CodeIgniter\Test\CIUnitTestCase
 		$this->assertEquals('http://example.com/pages/notset', $response->getHeader('Location')->getValue());
 	}
 
+	public function testRunRedirectionWithHTTPCode303()
+	{
+		$_SERVER['argv']            = [
+			'index.php',
+			'example',
+		];
+		$_SERVER['argc']            = 2;
+		$_SERVER['REQUEST_URI']     = '/example';
+		$_SERVER['SERVER_PROTOCOL'] = 'HTTP/1.1';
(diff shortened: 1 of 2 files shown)


MGatner:
@Instrye it looks like there was more discussion on the issue since this PR. Is this ready for review?

Instrye:
> 
> 
> @Instrye it looks like there was more discussion on the issue since this PR. Is this ready for review?

Yes,It ready.

tests/system/CodeIgniterTest.php
```diff
@@ -360,6 +360,56 @@ public function testRunRedirectionWithURINotSet()
...
+		$routes->addRedirect('example', 'pages/notset', 301);
+
+		$router = Services::router($routes, Services::request());
+		Services::injectMock('router', $router);
+
+		ob_start();
+		$this->codeigniter->useSafeOutput(true)->run();
+		ob_get_clean();
+		$response = $this->getPrivateProperty($this->codeigniter, 'response');
+		$this->assertEquals('303', $response->getStatusCode());
```
kenjis, line 385: Why is this code changed to 303?
The route sets the status code 301.

kenjis: I sent a PR to fix the behavior: #7445

kenjis: I mean, it should be 301 ?

iRedds: Essentially it should be 301, but the handler is written so that the custom response code is ignored.

PR #3061, opened 2020-06-03
Move FeatureTest methods to trait

MGatner:
**Description**
`FeatureTestCase` is a superclass of `CIDatabaseTestCase` with methods specific to HTTP feature testing (and a few properties). Since developers will often want to configure specific database settings in their tests, currently that code must be duplicated across classes that extend `CIDatabaseTestCase` and `FeatureTestCase` each, or include feature test methods not necessary for database-only tests.

This PR moves methods to a new trait so they may be included optionally in test cases "on the fly". Additionally it changes a few of the methods so the properties are optional, making `FeatureTestCase` a convenience class identical to its current iteration but not needed for developers who want to implement their own.

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide


MGatner:
I’m confused what happened here. Are traits maybe handled differently when it comes to unit testing docblocks? If anyone has ideas I would love some help.

MGatner:
I see it, I started with a really old version of **FeatureTestCase.php**. Update coming.

PR #3062, opened 2020-06-04
Feature Tests & forceSecure

MGatner:
**Description**
Currently `FeatureTest` initiates an insecure request, which means if `forceGlobalSecureRequests` is activated then every status will be `302`. This PR has `FeatureTestTrait::setupRequest` check for that configuration and notify the app to make a special exemption so requests will be both a) secure and b) not redirected.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -400,7 +400,7 @@ if (! function_exists('force_https'))
 			$response = Services::response(null, true);
 		}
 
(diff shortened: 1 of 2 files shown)


lonnieezell:
Woo hoo! Tests passed. :)

PR #3065, opened 2020-06-05
Move shouldDiscover in a new system/Modules class

mostafakhudair:
Creating new system/Modules/Modules.php file and moves shouldDiscover function into it away from user hand like before app/Config/Modules.php
to help avoid tampering with the core functions

### DIFF
--- a/system/bootstrap.php
+++ b/system/bootstrap.php
@@ -120,6 +120,7 @@ if (! class_exists(Config\Autoload::class, false))
 {
 	require_once SYSTEMPATH . 'Config/AutoloadConfig.php';
 	require_once APPPATH . 'Config/Autoload.php';
+	require_once SYSTEMPATH . 'Modules/Modules.php';
 	require_once APPPATH . 'Config/Modules.php';
 }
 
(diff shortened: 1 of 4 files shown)


system/bootstrap.php
```diff
@@ -120,6 +120,7 @@
 {
 	require_once SYSTEMPATH . 'Config/AutoloadConfig.php';
 	require_once APPPATH . 'Config/Autoload.php';
+	require_once SYSTEMPATH . 'Modules/Modules.php';
```
samsonasik, line 123: should be added to `system/Test/bootstrap.php` as well.

lonnieezell:
@mostafakhudair Are you going to be able to fix up these tests and address the comment that @samsonasik made?

samsonasik:
@lonnieezell @mostafakhudair already applied to `system/Test/bootstrap.php`

samsonasik:
The tests still needs update indeed

mostafakhudair:
@lonnieezell everthing is wroking well now

PR #3069, opened 2020-06-05
Add fake() function

MGatner:
**Description**
Provides a quick-and-easy helper function for Fabricator the leverages models pre-configured for faking data.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


MGatner:
@lonnieezell Let me know your thoughts on this. Originally it was a static method to `Fabricator` but then I got thinking there would be more functions to justify a test helper. It does constitute test-specific code in a production-available helper though, so could revert to the static method if you prefer.

lonnieezell:
Hm. Thought I had given this one the thumbs up the other day while waiting for tests to run. Must have been a different one. 

But, yes, I like this. Fits in with the rest of the helpers and is nicer to use, I would think.

PR #3077, opened 2020-06-09
FeatureTest use current session

MGatner:
**Description**
Currently, running a Feature Test will start with a fresh Session when you make the call(). While this helps ensure that each test is contained it means you cannot prep anything in the Session, such as logging a user in.

This PR is solution "Plan A" which uses the current Session (unless overriding values are passed to `withSession()`) and blanks the session *after* the call, returning the resulting session in `FeatureResponse` for examination.

Since this may entail a breaking change I'm considering this a draft PR and will submit a non-BC option as well, but note that I think this is a better solution for feature testing.

Ref. https://forum.codeigniter.com/thread-76673.html

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/FeatureResponse.php
+++ b/system/Test/FeatureResponse.php
@@ -57,6 +57,13 @@ class FeatureResponse extends TestCase
 	public $response;
 
 	/**
+	 * State of $_SESSION after the call.
+	 *
+	 * @var array|null
+	 */
+	public $session;
+
+	/**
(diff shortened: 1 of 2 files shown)


lonnieezell:
Personally - I iike Plan B the best. And not just because the tests pass on that one lol.

PR #3078, opened 2020-06-09
Add withSession null case

MGatner:
**Description**
Currently, running a Feature Test will start with a fresh Session when you make the call(). While this helps ensure that each test is contained it means you cannot prep anything in the Session, such as logging a user in.

This PR is solution "Plan B" which overloads `withSession()` to default to use the current Session values. This still requires calling `withSession()` on every feature test, but allows the session to prepared by all the usual means by the test case.

See also #3077.

Ref. https://forum.codeigniter.com/thread-76673.html

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -90,13 +90,13 @@ trait FeatureTestTrait
 	/**
 	 * Sets any values that should exist during this session.
 	 *
-	 * @param array $values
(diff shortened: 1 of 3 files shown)


lonnieezell:
I like this one the best, honestly. I think the explicitness might get devs from getting hit with unintended consequences when they forget to clear the SESSION during a test and have to spend an hour trying to figure out why....

PR #3080, opened 2020-06-09
Fix class collision "PageNotFound" error

MGatner:
**Description**
 #2958 allowed more robust testing for fallback paths on exception display, but the use of `extract()` caused a collision in some cases between the actual file to include and the variable `$file` to display. This PR sets a more unique name of the include path to avoid collision.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -287,11 +287,11 @@ class Exceptions
 		// Check if the view exists
 		if (is_file($path . $view))
 		{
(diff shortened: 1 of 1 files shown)


paulbalandan:
@MGatner I've made the PR at #3076

MGatner:
Ah geez you completely solved this already and I didn’t notice it! Thanks for the PR, will merge.

PR #3081, opened 2020-06-09
Added innerJoin, leftJoin and rightJoin functions

pera-programer:
Added innerJoin, leftJoin and rightJoin functions to facilitate making table joins

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -704,6 +704,25 @@ class BaseBuilder
 
 		return $this;
 	}
+        
+        
+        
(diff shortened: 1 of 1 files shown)


lonnieezell:
I'm all for these changes, as they make the code a bit more readable, but we need tests in place for each function and docs for the functions so people know they exist :) 

Also, please ensure PR's are [signed](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/signing.rst).

samsonasik:
please create another branch instead of same `develop` and do rebase

Issue #3063, opened 2020-06-04
Bug: Risky feature tests with Controller returns


MGatner:
**Describe the bug**
`FeatureTestCase` allows simulating an app run and gathering the output for testing purposes. If a controller returns the output instead of echoing it (e.g. `return view('Home');`) it isn't cleaning the output buffer correctly causing risky tests with the following warning:

> Test code or tested code did not (only) close its own output buffers

**CodeIgniter 4 version**
`develop` latest

**Affected module(s)**
`FeatureTestTrait`

**Expected behavior, and steps to reproduce if appropriate**
Controller:
```
class Home extends BaseController
{
	public function index()
	{
		return view('home');
	}
```

Test:
```
class HomeTest extends FeatureTestCase
{
	public function testGetHome()
	{
		$result = $this->get('/');

		$result->assertStatus(200);
```

**Context**
 - OS: Linux
 - Web server: Apache
 - PHP version: 7.3

MGatner:
@samsonasik any chance this is an easy fix similar to #2551 or #2625? I honestly have dug into it yet but won't have a chance to today.

samsonasik:
I tried do buffering check in the FeatureTestCase class if i remember correctly, but not work. I hope I can re-check again if I have a chance.

samsonasik:
@MGatner should can be closed in favor of #3083 ?

MGatner:
Yes! Thanks for catching that. I had a “fixes ____” commit but I must have overwritten it when I force pushed.

PR #3083, opened 2020-06-10
Clean buffers after feature test

MGatner:
**Description**
This might be a long shot at #3063 but I can't run feature tests locally so I'm giving it a try here.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -127,6 +127,8 @@ trait FeatureTestTrait
 	 */
 	public function call(string $method, string $path, array $params = null)
 	{
+		$buffer = \ob_get_level();
+
 		// Clean up any open output buffers
 		// not relevant to unit testing
 		// @codeCoverageIgnoreStart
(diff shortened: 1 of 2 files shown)


MGatner:
I think this is the correct implementation now. This could sometimes "hide" issues with views but explicit testing should uncover those, I don't believe that's the job of the `FeatureTestCase`.

@samsonasik would you review please, since you took the last pass at this?

samsonasik:
@MGatner can you remove ob_start() in the tests at:

1. https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Test/FeatureTestCaseTest.php#L230 

2. https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Test/FeatureTestCaseTest.php#L244

to ensure it works?

samsonasik:
that still make risky tests

MGatner:
@samsonasik I don't totally understand it but I *think* cases that echo directly from an empty buffer needs to recreate the original buffer. I just added a commit so that the buffer is returned to its pre-test value regardless of state (up or down). It's kind of weird but it will prevent risky tests. If there is some deeper issue with buffering in the framework itself then it alludes me.

MGatner:
Travis build green, no risky tests!

PR #3084, opened 2020-06-10
Add webp support to Image Manipulation Class

nicojmb:

**Description**
Support webp image in Image Manipulation Class

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated

**Notes**
(e.g. fixes <hash>12345)

### DIFF
--- a/system/Images/Handlers/GDHandler.php
+++ b/system/Images/Handlers/GDHandler.php
@@ -365,6 +365,17 @@ class GDHandler extends BaseHandler
 					throw ImageException::forSaveFailed();
 				}
 				break;
+			case IMAGETYPE_WEBP:
+				if (! function_exists('imagewebp'))
+				{
+					throw ImageException::forInvalidImageCreate(lang('images.webpNotSupported'));
+				}
+
+				if (! @imagewebp($this->resource, $target))
+				{
+					throw ImageException::forSaveFailed();
+				}
+				break;	
 			default:
 				throw ImageException::forInvalidImageCreate();
 		}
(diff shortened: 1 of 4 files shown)


michalsn:
@nicojmb we have two things to fix here: 

* The first - tests. If you look at Travis, we have 3 tests that are failing now, we have to change constant from `IMAGETYPE_PNG` to `IMG_PNG` since this is the correct one returned by `getimagesize()` function.
* The second thing that I didn't notice earlier is that your commits are not signed and we can't merge your code. Please check this page to learn more about code signing: https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/signing.rst

@MGatner is there a way to sign the code right now, to make it mergeable? I think you showed me once some instructions but I didn't succeed make them work for me (it was probably my fault). So, I'm just wondering is there a way to "fix" this PR in this state.

MGatner:
@michalsn you can add credentials to GitHub post-push and it will verify precious commits, but if the commits weren’t signed already that won’t work. Anyone could pull the fork, `git reset develop`, then re-add and re-commit a signed version and make a new PR. Or Lonnie could just override the verification requirement 😇

MichalPB1:
Can it be merged? This is something that is useful to me :)

michalsn:
Right now we can't merge it since it's not signed and tests are failing. If nothing will change, I will probably wait till the weekend and try to fork this branch and re-commit.

lonnieezell:
Yeah, the biggest problem is the tests. I am happy to override. And am thinking about removing that requirement, anyway, actually.

michalsn:
Yes, removing the code signing requirement would probably be a good idea.

PR #3094, opened 2020-06-13
Fixed FeatureTest isOK to return true for redirects. Fixes #3072

### LINKED ISSUE #3072: Bug: FeatureResponse::isOk and redirects
**Describe the bug**
Am I missing something obvious? `FeatureResponse::isOk` looks like this:
```
		// Only 200 and 300 range status codes
		// are considered valid.
		if ($this->response->getStatusCode() >= 400 || $this->response->getStatusCode() < 200)
		{
			return false;
		}

		// Empty bodies are not considered valid.
		if (empty($this->response->getBody()))
		{
			return false;
		}
```

... but won't redirects (300 statuses) *always* have an empty body? Regardless of the always, if you use the standard `return redirect()->to(...` in a controller and then test it with `FeatureTestCase` then the resulting `assertOk()` will fail, which seems like a mistake to me.

**CodeIgniter 4 version**
`develop`

**Affected module(s)**
`FeatureResponse`

*See above*

**Context**
 - OS: Linux
 - Web server: Apache
 - PHP version: 7.3

lonnieezell:


MGatner:
Thanks! Just what I wanted.

PR #3099, opened 2020-06-14
Fix typehints of $request in Validation

paulbalandan:
**Description**
Fixes typehints of the `$request` object in `Validation` and `FileRules`.

Currently, some of the methods used on `$request` do not belong to `RequestInterface` but are found in `IncomingRequest`, like `getFile`, `getFileMultiple`, `getRawInput`, etc. This causes typehinting errors in IDEs like VSCode.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FileRules.php
+++ b/system/Validation/FileRules.php
@@ -39,7 +39,7 @@
 
 namespace CodeIgniter\Validation;
 
-use CodeIgniter\HTTP\RequestInterface;
+use CodeIgniter\HTTP\IncomingRequest;
 use Config\Services;
 
 /**
@@ -51,7 +51,7 @@ class FileRules
 	/**
 	 * Request instance. So we can get access to the files.
 	 *
-	 * @var \CodeIgniter\HTTP\RequestInterface
+	 * @var \CodeIgniter\HTTP\IncomingRequest
 	 */
 	protected $request;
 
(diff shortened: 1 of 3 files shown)


michalsn:
Hmm... I think this would be a BC change.

Using `RequestInterface` means that every class that implements this interface may be used. Now we are limiting this to only one class: `IncomingRequest` which not necessarily have to be used by the developer. Anybody can write his own implementation of "IncomingRequest" and load it via service.

paulbalandan:
Yes, I have also thought that this may be a BC change.

But my point is most of the methods used on `$request` in the validation do not belong to `RequestInterface`. Say, I made my own class purely implementing `RequestInterface` out-of-the-box and passing this manually as the `$request` object to `FileRules`, the constructor won't complain. But when I execute a method, like `max_size`, this will likely cause an error here:
```php
if (! ($files = $this->request->getFileMultiple($name)))
    {
        $files = [$this->request->getFile($name)];
    }
``` 
This is because `getFile` and `getFileMultiple` are methods of `IncomingRequest`.

paulbalandan:
My second take on this is to put those "missing" methods to `RequestInterface`, but classes already implementing this need to be reworked which I think is a bit of messy.

michalsn:
Well... it's not about implementing only methods that are declared in the interface class, it's about the ability to use other classes than `IncomingRequest`.

paulbalandan:
Hmm.. fair point. Thanks.

PR #3103, opened 2020-06-15
Allow array file uploads to be validated also by 'uploaded'. Fixes #3018

### LINKED ISSUE #3018: Bug: Multiple File Uploads validation rule "uploaded[inputName]" does not work when the input name is an Array
**Describe the bug**
Multiple File Uploads validation rule "uploaded[inputName]" does not work when the input name is an Array 
`<input type="file" multiple name="theFiles[]">`

All other file related rules work with array field names

**CodeIgniter 4 version**
4.0.3

**Affected module(s)**
Validation\FileRules.php

Looks like all of the methods (except uploaded()) inside of FileRules.php use this lines of code

`if (! ($files = $this->request->getFileMultiple($name)))
		{
			$files = [$this->request->getFile($name)];
		}`
before performing the required validations using a `foreach` loop

`function uploaded()` does not have this lines and only checks for single file. Therefore, if input is an array of files, it fails and returns false.

For now the workaround is to use as the rule ['theFiles' => 'uploaded[theFiles.0 ]']

I don't know if it is an intended behavior but in documentation it says:

> The file validation rules apply for both single and multiple file uploads.

paulbalandan:
**Description**
Currently, all validation checks of `FileRules` allows validating array file uploads, except for `uploaded`. This PR adds that functionality. The documentation had already said that **all** file validation rules check for both single and multiple file uploads.

Fixes #3018 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FileRules.php
+++ b/system/Validation/FileRules.php
@@ -84,22 +84,38 @@ class FileRules
 	 */
 	public function uploaded(string $blank = null, string $name): bool
 	{
-		$file = $this->request->getFile($name);
-
-		if (is_null($file))
+		if (! ($files = $this->request->getFileMultiple($name)))
 		{
-			return false;
+			$files = [$this->request->getFile($name)];
 		}
 
-		if (ENVIRONMENT === 'testing')
+		foreach ($files as $file)
 		{
-			return $file->getError() === 0;
+			if (is_null($file))
+			{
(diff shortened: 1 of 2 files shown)


michalsn:
Thanks :)

PR #3104, opened 2020-06-15
Fix to ImageMagickHandler

colethorsen:
-Saving an image back to the same location with a lower quality would result in a failure attempting to copy.
-Saving an image to a new spot with a lower quality would result in the image not having a new quality set.

**Description**

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Images/Handlers/ImageMagickHandler.php
+++ b/system/Images/Handlers/ImageMagickHandler.php
@@ -288,23 +288,34 @@ class ImageMagickHandler extends BaseHandler
 	{
 		$target = empty($target) ? $this->image() : $target;
 
-		// If no new resource has been created, then we're
-		// simply copy the existing one.
-		if (empty($this->resource))
+		// If no new resource has been created,  and we
+		// aren't adjusting the quality then we're simply
+		// copy the existing one.
+		if (empty($this->resource) && $quality == 100)
 		{
+			// we're moving it to the same spot so don't do anything.
+			if($target === null)
+			{
+				return true;
+			}
+
 			$name = basename($target);
 			$path = pathinfo($target, PATHINFO_DIRNAME);
 
 			return $this->image()->copy($path, $name);
 		}
 
+		$source = $this->image()->getPathname();
+
 		// Copy the file through ImageMagick so that it has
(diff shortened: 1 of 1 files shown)


system/Images/Handlers/ImageMagickHandler.php
```diff
@@ -288,23 +288,34 @@ public function save(string $target = null, int $quality = 90): bool
...
-		// If no new resource has been created, then we're
-		// simply copy the existing one.
-		if (empty($this->resource))
+		// If no new resource has been created,  and we
+		// aren't adjusting the quality then we're simply
+		// copy the existing one.
+		if (empty($this->resource) && $quality == 100)
 		{
+			// we're moving it to the same spot so don't do anything.
+			if($target === null)
```
michalsn, line 297: I can't see the scenario when this condition would be used.

system/Images/Handlers/ImageMagickHandler.php
```diff
@@ -288,23 +288,34 @@ public function save(string $target = null, int $quality = 90): bool
...
+				return true;
+			}
+
 			$name = basename($target);
 			$path = pathinfo($target, PATHINFO_DIRNAME);
 
 			return $this->image()->copy($path, $name);
 		}
 
+		$source = $this->image()->getPathname();
```
michalsn, line 308: What if we already made some operations on the image? Not using `$this->resource` means that we will lose everything.

colethorsen: you're right, missed copying the resource over if it exists.

system/Images/Handlers/ImageMagickHandler.php
colethorsen: This would fail for instance if someone set it to save to the same location and set a quality to 100. I'd think it would be easier for this to just leave it alone, than to fail trying to copy it to the same spot

michalsn: I understand but look at line 289. `$target` will never be a `null` value.

michalsn:
Fixed via #3268

PR #3107, opened 2020-06-16
Update CLI getHeight and getWidth

paulbalandan:
**Description**
At present, actual checking of terminal height and width happens only for *nix systems. For Windows, only default values are given.

This PR uses [`mode CON`](https://www.computerhope.com/modehlp.htm) to get the actual lines and columns for Windows systems. However, there were reports that Cygwin and GitBash returns incorrect values when using `mode CON` so, as fallback, `tput` was used.

As for Unix systems, added `stty size` as primary check before fallback to `tput`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


lonnieezell:
Thanks!

Issue #3105, opened 2020-06-15
Bug: Implicit controller methods failing


MGatner:
**Describe the bug**
I don't have time to dig but my project feature tests work fine on 2901fdd3af7066b43db99074fe997211c70bd437 but fail on latest `develop` whenever a route uses an implicit controller method. I suspect #3096 simply because it deals with routing but not sure of anything.

E.g.:
```
3) HomeTest::testGetHome
CodeIgniter\Exceptions\PageNotFoundException: Controller method is not found: index

/var/www/project/vendor/codeigniter4/codeigniter4/system/Exceptions/PageNotFoundException.php:14
/var/www/project/vendor/codeigniter4/codeigniter4/system/CodeIgniter.php:984
/var/www/project/vendor/codeigniter4/codeigniter4/system/CodeIgniter.php:327
/var/www/project/vendor/codeigniter4/codeigniter4/system/Test/FeatureTestTrait.php:158
/var/www/project/vendor/codeigniter4/codeigniter4/system/Test/FeatureTestTrait.php:184
/var/www/project/tests/features/HomeTest.php:9
```

**CodeIgniter 4 version**
`develop`

**Affected module(s)**
Routing, maybe Feature Tests.

**Expected behavior, and steps to reproduce if appropriate**
Running `get('/')` should load my home controller as mapped in **app/Config/Routes.php**

**Context**
 - OS: Ubuntu 18.04
 - Web server: Apache
 - PHP version: 7.3

MGatner:
Now I can't replicate this. I'm going to leave it overnight until I get back to the original project but might be a false alarm.

MGatner:
Having the issue on GitHub Actions now.

MGatner:
I've narrowed this down to a test that is causing the problem. Running the suite without it completes without issue, but with the test somehow messes with the Feature Test routes. There's nothing different about this feature test than the others, currently baffled but I will keep digging.

PR #3108, opened 2020-06-16
Use explicit Routes in feature testing

MGatner:
**Description**
Currently two separate test cases will cause a conflict in the resulting `$routes` used in the simulated application. I think this ultimately has to do with re-requiring **Routes.php** in `tryToRouteIt()`:
```
if (empty($routes) || ! $routes instanceof RouteCollectionInterface)
{
	require APPPATH . 'Config/Routes.php';
}
```

This PR forces `FeatureTestTrait::call()` to load `$routes` from Services (if none were specified) to prevent the corruption. See also #3105 for symptoms.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -146,10 +146,8 @@ trait FeatureTestTrait
 		$request = $this->populateGlobals($method, $request, $params);
 
 		// Make sure the RouteCollection knows what method we're using...
-		if (! empty($this->routes))
-		{
-			$this->routes->setHTTPVerb($method);
-		}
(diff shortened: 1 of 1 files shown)


MGatner:
The problem with replicating it is that it needs to be two different test cases and they have to be run in the same process (I think). I can think about it a little bit more but it does definitely solve the issue. 

Generally I think Feature Tests probably need a little bit more “reset”, like the other issue about Filters suggests. That could be an alternate way to deal with this route bleed-over.

Issue #579, opened 2017-07-01
Propose Moving Model::classToArray() method to Entity class as public method


daylightstudio:
I've found it useful to be able for a model's entity class to return an array of the values being saved (e.g. for a toJson method on the Entity class). Currently the method that extracts that data is in the Model class but protected. I would like to suggest moving that method to the Entity class as public and have the Model class call it instead.

```
protected $exclude = [];

public function values()
{
	$mirror = new \ReflectionClass($this);
	$props  = $mirror->getProperties(\ReflectionProperty::IS_PUBLIC | \ReflectionProperty::IS_PROTECTED);

	$properties = [];

	$exclude = array_merge(['datamap', 'exclude'], $this->exclude);

	// Loop over each property,
	// saving the name/value in a new array we can return.
	foreach ($props as $prop)
	{
		if (in_array($prop->getName(), $exclude))
		{
			continue;
		}

		// Must make protected values accessible.
		$prop->setAccessible(true);
		$properties[$prop->getName()] = $prop->getValue($this);
	}

	return $properties;
}
```

lonnieezell:
I need to think on this one a little more. Here's my concerns: 

1. This wouldn't support anyone who didn't want to use the Entity class at the basis of a class-based value-object. Which may not be a concern, because we probably want to encourage people to use Entity in these instances. However, feels like we might be reducing user choice here. 

2. (and this is very much a personal thing) I'm not sure I really want to encourage taking the current class and jamming it back into a JSON object. My view is that you should explicitly decide what to return so that you don't screw things up when you have to version an API, or add new db-columns that are intended for internal usage and you don't want it thrown back in the API.

daylightstudio:
What about making classToArray a static public method?

lonnieezell:
> What about making classToArray a static public method?

That, I think is doable.

lonnieezell:
Fixed in 2d05e6470ea00cbcc1d5e032fca7ca17d72f06f3

PR #3117, opened 2020-06-18
fixed missing ) after statement

ertanyildiz:
There was no closing parenthesis on #579 in `user_guide_src/source/incoming/incomingrequest.rst:243`. I've added the missing `)` to the line.


michalsn:
Thanks, but there are two things. 

* We can only merge commits that are signed. You can read more about it here: https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/signing.rst
* Also, for the future - if your changes don't involve any actual code from framework, you can use `[ci skip]` in your commit description to just skip Travis build.

ertanyildiz:
@michalsn Thanks for your feedback. I have reopened the pull request as requested.

PR #3119, opened 2020-06-18
Email Event

MGatner:
**Description**
Defines a new Event, "email", triggered on successful email send. The event receives the Email class's current set of properties so listeners can respond to email specifics.

@samsonasik Please review the move of `archive` on the MockEmail class per my other comment. If you want to review this whole PR I would appreciate it :)

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -1548,13 +1548,13 @@ class Email
 		}
 		$this->buildMessage();
 		$result = $this->spoolEmail();
-		if ($result && $autoClear)
-		{
-			$this->clear();
-		}
-
 		if ($result)
 		{
+			if ($autoClear)
+			{
+				$this->clear();
+			}
+
 			Events::trigger('email', get_object_vars($this));
 		}
 
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -39,6 +39,7 @@
 
 namespace CodeIgniter\Email;
 
+use CodeIgniter\Events\Events;
 use Config\Mimes;
 
 /**
@@ -1530,8 +1531,7 @@ class Email
 		{
 			$this->setReplyTo($this->headers['From']);
 		}
-		if (empty($this->recipients) && ! isset($this->headers['To']) && empty($this->BCCArray) && ! isset($this->headers['Bcc']) && ! isset($this->headers['Cc'])
-		)
+		if (empty($this->recipients) && ! isset($this->headers['To']) && empty($this->BCCArray) && ! isset($this->headers['Bcc']) && ! isset($this->headers['Cc']))
 		{
 			$this->setErrorMessage(lang('Email.noRecipients'));
 			return false;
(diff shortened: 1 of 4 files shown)


system/Email/Email.php
```diff
@@ -1552,6 +1552,12 @@ public function send($autoClear = true)
 		{
 			$this->clear();
 		}
+
+		if ($result)
```
samsonasik, line 1556: if can be combined with previous if :

```php
if ($result)
{
         if ($autoClear)
         {
                   $this->clear();
         }
        
       Events::trigger('email', get_object_vars($this));

}
```

system/Test/Mock/MockEmail.php
```diff
@@ -11,14 +12,27 @@ class MockEmail extends Email
...
+	 *
+	 * @var boolean
+	 */
+	public $returnValue = true;
+
 	public function send($autoClear = true)
 	{
+		$this->archive = get_object_vars($this);
+
 		if ($autoClear)
```
samsonasik, line 26: I think it need to be : 

```suggestion
              if ($autoClear && ! $this->returnValue)
```

So the `archive` property will not be cleared when `retrunValue` is true.

tests/system/Email/EmailTest.php
```diff
@@ -36,4 +38,42 @@ public function testEmailSendWithClearance($autoClear)
...
+			$result = $arg;
+		});
+
+		$this->assertTrue($email->send());
+
+		$this->assertIsArray($result);
+		$this->assertEquals(['foo@foo.com'], $result['recipients']);
+	}
+
+	public function testFailreDoesNotTriggerEvent()
```
samsonasik, line 61: typo fix: 

```suggestion
      public function testFailureDoesNotTriggerEvent()
```

tests/system/Email/EmailTest.php
```diff
@@ -36,4 +38,42 @@ public function testEmailSendWithClearance($autoClear)
 			$this->assertEquals('foo@foo.com', $email->archive['recipients'][0]);
 		}
 	}
+
+	public function testSuccessTriggersEvent()
```
samsonasik, line 42: need to have consistent grammar, trigger or triggers

system/Email/Email.php
MGatner: Updated.

system/Test/Mock/MockEmail.php
MGatner: `clear()` does not affect `$archive` but rather a pre-defined list of properties, so I think this is fine. Your comment did make me realize though that we should only be clearing on successful returns, so I'm giving this the same consolidation treatment as `Email` above.

tests/system/Email/EmailTest.php
MGatner: Oops! Thanks.

tests/system/Email/EmailTest.php
MGatner: Done.

PR #3136, opened 2020-06-20
New Model Method `last()`

mostafakhudair:
as first() method, last() method to determine last row with respecting useSoftDeleted property

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -498,6 +498,47 @@ class Model
 		return $eventData['data'];
 	}
 
+    	//--------------------------------------------------------------------
+
+    	/**
+     	* Returns the last row of the result set. Will take any previous
+     	* Query Builder calls into account when determining the result set.
+     	*
+     	* @return array|object|null
+     	*/
+    	public function last()
+    	{
+		$builder = $this->builder();
+
+		if ($this->tempUseSoftDeletes === true)
+		{
+	    		$builder->where($this->table . '.' . $this->deletedField, null);
+		}
+		else
+		{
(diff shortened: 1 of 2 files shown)


michalsn:
Thank you for your PR. Friendly reminder - new features always need some tests.

Unfortunately, I don't think that the performance of the proposed solution is sufficient. If we had 100k records in the database, they would all be loaded into memory.

The easiest solution for this would probably be to use a subquery for the FROM part of the query, but the Builder class doesn't currently support this.

system/Model.php
```diff
@@ -498,6 +498,47 @@ public function first()
...
+
+		// Some databases, like PostgreSQL, need order
+		// information to consistently return correct results.
+		if (! empty($builder->QBGroupBy) && empty($builder->QBOrderBy) && ! empty($this->primaryKey))
+		{
+	    		$builder->orderBy($this->table . '.' . $this->primaryKey, 'asc');
+		}
+
+		$eventData = $this->trigger('afterFind', [
+	    		'data' => $builder->get()->getLastRow($this->tempReturnType)
```
michalsn, line 533: Actually, I just have a thought that we can make it better this way:

```suggestion
		$subQuery   = $builder->getCompiledSelect();
		$subBuilder = $this->db->table('(' . $subQuery . ') subQuery');
		$row = $subBuilder->orderBy('subQuery.' . $this->primaryKey, 'desc')->limit(1, 0)->get()->getFirstRow($this->tempReturnType);
                
		$eventData = $this->trigger('afterFind', [
	    		'data' => $row
```
Any other ideas?

Issue #3113, opened 2020-06-17
Bug: SQLite3 database file created in /public folder


ramiresviana:
**Describe the bug**
When using database with SQLite3 as DBDriver parameter, the database file is created in the /public application folder.

**CodeIgniter 4 version**
4.0.3

**Steps to reproduce**
1. Set database and DBDriver parameters in database configuration file.
2. Create new table with CodeIgniter Forge;
3. Database file is created in /public folder.

**Expected behavior**
Database files can contain private information and should not be created in the public folder by default.

**Context**
 - OS: Ubuntu 18.04.2
 - Web server: CodeIgniter Spark
 - PHP version: 7.2.24

lonnieezell:
You can change the location of the file using the `$database` value in the configuration for that connection. Give it a path and filename.

ramiresviana:
@lonnieezell the problem is not being able to change the location of the file, but rather the default path being the public folder. This is a undocumented behavior and can add a security flaw on applications that use CodeIgniter, not every developer is going to check where is the database is going to be write.

If a framework offers me sqlite support, i would expect that it will take some measures to protect the file from being leaked on the internet.

lonnieezell:
That's a fair point. And I guess it is default behavior inherited from CI3. I don't recall it having ever been changed, honestly. 

We should set the default location in the `writable` folder somewhere that makes sense. 

Thanks.

samsonasik:
I think the directory can be named "writable/data"

MGatner:
Cross-referencing my opinion: https://github.com/codeigniter4/CodeIgniter4/pull/3151#issuecomment-648232171

PR #3142, opened 2020-06-21
Fix default location for SQLite3 database

michalsn:
**Description**
We should make it very clear in the release notes that the default location has changed. I know this is more a bugfix, but it also may break "things" for some people.

See: #3113 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,9 @@ writable/session/*
 writable/uploads/*
 !writable/uploads/index.html
 
+writable/data/*
+!writable/data/index.html
+
 writable/debugbar/*
 
 php_errors.log
--- a/system/Database/SQLite3/Connection.php
+++ b/system/Database/SQLite3/Connection.php
@@ -85,6 +85,12 @@ class Connection extends BaseConnection implements ConnectionInterface
 		}
 		try
 		{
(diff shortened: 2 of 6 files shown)


lonnieezell:
@michalsn Could you start a changelog for the 4.0.4 and add a note about this change? Don't want to forget that add that.

michalsn:
@lonnieezell Done.

lonnieezell:
Looks good. Thanks! Merging.

ivantcholakov:
I wonder whether writable/ directory is enough from maintenance point of view. Here is an illustration of the problem:

writable/
    keep/
    disposable/

There is data that is temporary (cache files or similar) and there is data that is valuable and should not be lost by a mistake. If all these two categories of data are placed under a single directory, it would be harder to distingush them, mistakes might happen, somebody might delete precious directories/files.

samsonasik:
@ivantcholakov I created new PR #3151 to handle that with dedicated `./data` directory in root project.

MGatner:
Cross-referencing my opinion: https://github.com/codeigniter4/CodeIgniter4/pull/3151#issuecomment-648232171

PR #3143, opened 2020-06-21
more test for PagerRenderer

samsonasik:
Added more test to PagerRenderer for testing `getPreviousPage()` and `getNextPage()` when segment > 0. By this,  `Pager\PagerRenderer` class will have 100% test coverage.

**Checklist:**
- [x] Securely signed commits


samsonasik:
I think segment need to be forced to 2 when the current uri already have a path, eg : `/foo`.

samsonasik:
travis build green :tada: the PagerRenderer now has 100% test coverage https://coveralls.io/jobs/64150325/source_files/951769722

Issue #3138, opened 2020-06-20
Execute bootstrapEnvironment() & detectEnvironment() before Services::exceptions() 


arma7x:
I did override current Services::exceptions() and my new Services::exceptions() need to access specific value of CONSTANT loaded from custom boot files based upon the current environment.
Posibble to update [Codeigniter.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/CodeIgniter.php) by moving [Line 190](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/CodeIgniter.php#L190) below [Line 194](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/CodeIgniter.php#L194)

arma7x:
missmatch label, Refactor

michalsn:
I think it shouldn't cause problems. And your reasoning to have access to the environment variables make sense to me.

You should try to make a PR (if you want to this change happen sooner rather than later 😉).

PR #3147, opened 2020-06-22
define environment variables

arma7x:
https://github.com/codeigniter4/CodeIgniter4/issues/3138#issue-642431631


michalsn:
Sorry, but we can only merge commits that are signed. You can read more about it here: https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/signing.rst

PR #3151, opened 2020-06-23
use writable directory for sqlite default location

samsonasik:
<strike>To avoid removed by incident as @ivantcholakov comment at https://github.com/codeigniter4/CodeIgniter4/pull/3142#issuecomment-647875975</strike>

**Checklist:**
- [x] Securely signed commits


michalsn:
Honestly speaking `data` folder placed in a root folder looks a bit "random" to me. Also, I don't like the idea of having a "special" folder in the project root just in case someone will use SQLite3 database.

Overall I'm not sure if placing database in the `writable` folder is really an issue... but if we must move it somewhere else, I would choose `app/database` folder - it seems more logical and usually, we don't delete anything from there.

Anyway, this change will stay with us for some time, so I can't be the one to decide about this. 

@lonnieezell, @MGatner any thoughts?

MGatner:
I think that for permissions handling we should keep everything that the framework needs write access to in the **writable** folder. I understand the concerns from @ivantcholakov that this groups temp data together with important and sensitive data but I think the configuration and security issues are more important. 

Personally I always put my SQLite database at **writable/database.db** as it is super clear what it is (this is your database) and skips the extra folder. We also have precedence for this from the Playground: https://github.com/codeigniter4projects/playground/blob/develop/.env.example

samsonasik:
ok, closing it

michalsn:
Well... now I'm wondering if placing database at **writable/database.db** isn't actually a better idea 😅

MGatner:
🤷‍♂️ I think it's open. Since #3142 hasn't been released I think it is safe to change, but I also think your solution is fine. The only additional advantage I see to **writable/database.db** is that it doesn't require a change to **.gitignore** and the new **data** folder, which amount to required updates to existing projects.

michalsn:
It's definitely not too late for any changes in this matter.

Another advantage is that when we access **writable** directory, we immediately see the database file, so it's less likely to be deleted by accident.

samsonasik:
Ok, I will update to point to writable dir

MGatner:
Sounds good. Thanks you two! Good progress. It will be nice not having to set that variable in every SQLite test scenario now too :)

samsonasik:
@michalsn @MGatner updated

MGatner:
Looking better! Do we know what `$this->database` will be by default? I think it would be good to add the file to **.gitignore** (e.g. `writable/database.db`) but I can't immediately tell if that's the correct name.

samsonasik:
updated with simplified check database !== 'memory' && strpos($this->database, DIRECTORY_SEPARATOR) === false for adding WRITEPATH prefix

michalsn:
I would try to add `writable/*.db` and `writable/*.sqlite` to the **.gitignore**. `$this->database` is empty by default, so it can be anything. In fact, we can't sure if the file extension will be used at all.

samsonasik:
I've added  `writable/*.db` and `writable/*.sqlite` to .gitignore

samsonasik:
I updated tests config to use "database.db" to prove that the patch is working

samsonasik:
@MGatner @michalsn travis build green :tada:

PR #3152, opened 2020-06-23
Revamp Email archive

MGatner:
**Description**
Real sorry to do this fam - more issues with the new Email archive functionality. I added a test this time that will fail under the current conditions - basically repeated email sends with the same instance will overwrite archive values with blanks sometimes. This PR addresses it by using the `$tmpArchive` property instead of trying to bundle it all onto `$archive`. I've also consolidated the functionality to a new method to make it cleaner.

Hopefully the last on Email from me for a bit. :)

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -58,9 +58,15 @@ class Email
 	/**
 	 * Properties from the last successful send.
 	 *
+	 * @var array|null
+	 */
+	public $archive;
+	/**
+	 * Properties to be added to the next archive.
+	 *
 	 * @var array
 	 */
-	public $archive = [];
+	protected $tmpArchive = [];
 	/**
 	 * @var string
 	 */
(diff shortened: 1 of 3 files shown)


MGatner:
@samsonasik @michalsn

system/Email/Email.php
```diff
@@ -58,9 +58,15 @@ class Email
 	/**
 	 * Properties from the last successful send.
 	 *
+	 * @var array|null
+	 */
+	public $archive;
```
michalsn, line 63: How about making this variable protected and adding `getArchive()` method with a parameter that accepts `null|string` value?

michalsn: Oh... I get it. We should leave it as it is.

MGatner:
Cool, thank you for the feedback!

PR #3153, opened 2020-06-23
Refresh Composer files

MGatner:
**Description**
This repo holds the master copies of **composer.json** for the other CodeIgniter 4 repositories and project launchers. A few fo them have become out of sync. This PR cleans them up and updates versions and names to be consistent.

Issues addressed:
* Incorrect project name for **framework**
* Missing `dev` requirements added since last release (`faker`, `predis`)
* Missing script command "test"
* Inconsistent PHPUnit version scheme (`8.5.*` standardized to `^8.5`)
* Inconsistent tabs/spaces standardized to tabs

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/admin/framework/composer.json
+++ b/admin/framework/composer.json
@@ -11,16 +11,16 @@
 		"ext-json": "*",
 		"ext-mbstring": "*",
 		"kint-php/kint": "^3.3",
-		"psr/log": "^1.1",
-		"laminas/laminas-escaper": "^2.6"
+		"laminas/laminas-escaper": "^2.6",
+		"psr/log": "^1.1"
 	},
 	"require-dev": {
 		"codeigniter4/codeigniter4-standard": "^1.0",
 		"fzaninotto/faker": "^1.9@dev",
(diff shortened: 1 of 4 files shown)


samsonasik:
@MGatner could you normalize composer.json please, you can use `composer-normalize` for it https://github.com/ergebnis/composer-normalize

jreklund:
The only thing I can think of are `require-dev` block in framework. We don't include anything to my knowledge that require them. The user guide shows how to add phpunit manually, if you want to write tests for your application.

michalsn:
I think that the user guide mentions a `--no-dev` flag a couple of times, so I would go with a full list for `require-dev`. It's easier to have a full list than to search for the components you need later.

MGatner:
@samsonasik I ran them through the normalize tool - it looks like it just reordered alphabetically.

Re: `require-dev` on **framework** I would rather keep them because even if people don't use them it helps with package trackers to know related entries.

Issue #3156, opened 2020-06-23
valid_url does not work correctly


luispastendev:
Hello!

Url validation doesn't work, if I pass a string like this `asdf` it passes the validation which should be wrong

my validation is as follows:
```php
'rules' => 'permit_empty|valid_url|max_length[512]'
```

Hugs!

crustamet:
Tell us what is the expected result bro.
From my point of view the rule works correctly
asdf can be a valid url if you use localhost with some proxy or what not.
Just like localhost is a valid url, asdf is the same.

If you want to prepend the HTTP or HTTPS just use prep_url on the input.

michalsn:
Not a bug, sorry - `asdf` is a valid hostname.

Please take a look at [RFC 2396](http://www.faqs.org/rfcs/rfc2396) - these are the rules behind `FILTER_VALIDATE_URL`.

But we're open to adding a new rule that will be more "restrictive".

MGatner:
I'm reopening this issue because I think there are some things that need to be addressed.

1. `RFC 2396` is obsolete. The new doc is [RFC 3986](https://tools.ietf.org/html/rfc3986)
2. The spec differentiates between "URI" (Identifier) and "URL" (Location). A URL:

> The  term "Uniform Resource Locator" (URL) refers to the subset of URIs that, in addition to identifying a resource, provide a means of locating the resource by describing its primary access mechanism (e.g., its network "location").

3. The section ["Syntax Components"](https://tools.ietf.org/html/rfc3986#section-3) describes the parts of the URI. Of note, it *requires* a scheme (e.g. "https") but not a path:

> The scheme and path components are required, though the path may be empty (no characters)

4. Our current validation rule very clearly allows the opposite in order to validate "common" URLs like "CodeIgniter.com". See also the validation rules tests for some wonky passes (https://github.com/codeigniter4/CodeIgniter4/blob/d9864e33758aca9aca9dc44c121e402e2d51136f/tests/system/Validation/FormatRulesTest.php#L88)

Here's my opinion:

a) We need `valid_url_strict` or something that has a lower tolerance for actual URLs.
b) I think the current validator should *not* pass simple strings like "asdf", but in order to prevent a breaking change we should still accept "CodeIgniter.com". Probably could check for a scheme and pass to a) if found, use a simple regex otherwise.

[...]

PR #3157, opened 2020-06-24
fix. valid_urlnot work

Instrye:
Fix. #3156

**Description**
`valid_url` will no longer support separate strings, This conflicts with the previous implementation.
But, If this is not done, any string can be verified.

The PR need discuss and decide.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -368,11 +368,11 @@ class FormatRules
 			{
 				return false;
 			}
-
-			$str = $matches[2];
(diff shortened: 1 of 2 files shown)


michalsn:
Sorry, we can't accept this, but we're open to new validation rules.

See: https://github.com/codeigniter4/CodeIgniter4/issues/3156#issuecomment-648735197

PR #3161, opened 2020-06-24
Fix ternary operator for $path in Namespaces Command

paulbalandan:
**Description**
Currently, the `$path` is determined by using a `realpath` on it, or itself on failure. **BUT** this uses the `null coalescing operator`. Since `realpath` returns `false` on failure, the right-hand side of the statement will never be reached even on failure. Thus, this PR addresses this by using an `Elvis operator` instead.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


paulbalandan:
Travis failure is on FabricatorLiveTest. Seems unrelated..

michalsn:
Good catch, thanks!

PR #3162, opened 2020-06-24
Update delete_files() helper function

### LINKED ISSUE #3015: Bug: delete_files() and hidden directories
**Describe the bug**
`delete_files()` in the Filesystem Helper currently ignores hidden directories. I'm not sure if this is a "bug" since the code seems to do it deliberately (notice the first line):
```
if (is_dir($path . DIRECTORY_SEPARATOR . $filename) && $filename[0] !== '.')
{
	delete_files($path . DIRECTORY_SEPARATOR . $filename, $del_dir, $htdocs, $_level + 1);
}
elseif ($htdocs !== true || ! preg_match('/^(\.htaccess|index\.(html|htm|php)|web\.config)$/i', $filename))
{
	@unlink($path . DIRECTORY_SEPARATOR . $filename);
}
```
But this isn't documented anywhere and seems an odd exclusion. For example, cleaning up an example project `delete_files()` fails to remove the root directory because it contains **.github/**.

If leaving hidden directories is intentional, then I would be in favor of giving this the same treatment as `get_filenames()`: remove the recursive call, use a `RecursiveIteratorIterator` instead, and add a parameter for `bool $hidden = false`.

**CodeIgniter 4 version**
`develop`

**Affected module(s)**
Helpers > Filesystem

Given:
ROOT/app/[files]
ROOT/writable/[files]
ROOT/.github/[files]

I would expect calling `delete_files('ROOT', true)` to wipe the whole project, but it does not.

**Context**
 - OS: Linux
 - Web server: Apache
 - PHP version: 7.3

michalsn:
**Description**
Support for removing hidden folders. See: #3015

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/filesystem_helper.php
+++ b/system/Helpers/filesystem_helper.php
@@ -201,7 +201,7 @@ if (! function_exists('delete_files'))
 
 			return true;
 		}
-		catch (\Exception $e)
+		catch (\Throwable $e)
 		{
 			return false;
 		}
--- a/system/Helpers/filesystem_helper.php
+++ b/system/Helpers/filesystem_helper.php
@@ -160,39 +160,48 @@ if (! function_exists('delete_files'))
 	 * @param string  $path    File path
 	 * @param boolean $del_dir Whether to delete any directories found in the path
 	 * @param boolean $htdocs  Whether to skip deleting .htaccess and index page files
-	 * @param integer $_level  Current directory depth level (default: 0; internal use only)
+	 * @param boolean $hidden  Whether to include hidden files (files beginning with a period)
 	 *
 	 * @return boolean
 	 */
-	function delete_files(string $path, bool $del_dir = false, bool $htdocs = false, int $_level = 0): bool
+	function delete_files(string $path, bool $del_dir = false, bool $htdocs = false, bool $hidden = false): bool
 	{
-		// Trim the trailing slash
-		$path = rtrim($path, '/\\');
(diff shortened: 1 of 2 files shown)


system/Helpers/filesystem_helper.php
```diff
@@ -160,39 +160,48 @@ function write_file(string $path, string $data, string $mode = 'wb'): bool
...
 				}
 			}
 
-			closedir($current_dir);
-
-			return ($del_dir === true && $_level > 0) ? @rmdir($path) : true;
+			return true;
 		}
-		catch (\Exception $fe)
+		catch (\Exception $e)
```
paulbalandan, line 204: Just asking. Should we use `\Throwable` here instead?

michalsn: Yes, thank you.

michalsn:
The thing is that we need our files to be in a certain order so we can delete them and `get_filenames()` function doesn't give us that. 

Of course, we can work on the result returned by `get_filenames()` but it won't be very pretty. I would stay with the current approach, even if these two functions are a bit similar.

michalsn:
I'm not sure why Coveralls show decreased test coverage. I will say it's a mistake and merge.

MGatner:
I’ve found the Coveralls CI to be unreliable on small PRs. Not worried about it.

PR #3168, opened 2020-06-25
public function getQBFrom(): array

georgewar85:
I currently need to get the tables loaded by the command $builder = $this->db->table (...) and there is no public method for this.

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -298,6 +298,16 @@ class BaseBuilder
 			}
 		}
 	}
+	
+	/**
+	 * Return the protected QBFrom[];
+	 *
+	 * @return Array
+	 */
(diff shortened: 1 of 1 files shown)


lonnieezell:
Wow. This one slipped by all of us. Sorry about that. 

If approved I would ask for a different name that didn't mimic the internals of the class, but was more pleasant from a dev perspective, like `getFrom`. However, I think it would be odd to have a single thing that you could get and not be able to get others. Also not sure how truly useful this is to the general public. 

Since it's been so long, I'm closing this for now. It can always be revisited if desired, but would need tests, docs, etc.

Issue #3169, opened 2020-06-26
Router RegEx not working. 


mmrtonmoybd:
RegEx not working in router. I use this
 [a-z0-9\x{0980}-\x{09ff}-]+$/u 
RegEx but Codeigniter throw a error out over of range system/Router/Router.php on line 425

michalsn:
Sorry, but we don't allow to set regex flags or delimiters. But I guess we should support `u` by default.

mmrtonmoybd:
> Sorry, but we don't allow to set regex flags or delimiters. But I guess we should support `u` by default.

Thank you now it's working and router RegEx now support Unicode.

michalsn:
I'm reopening this issue since it's not resolved yet.

@mmrtonmoybd it would be great if you could check #3172 and see if it works as it should.

mmrtonmoybd:
Unicode router not working with place holder.
Like this :
$routes->addPlaceholder(':category', '([a-z\x{0980}-\x{09ff}-]+)');
$routes->addPlaceholder(':pslug', '([a-zA-Z0-9\x{0980}-\x{09ff}-]+)');

$routes->add('(:category)/(:pslug)', 'Single::index/$1/$2');

But it is working on segment like this

$routes->add('([a-z\x{0980}-\x{09ff}-]+)/([a-zA-Z0-9\x{0980}-\x{09ff}-]+)', 'Single::index/$1/$2');

michalsn:
I just checked it and it works fine. I added a test for this to make sure.

I believe your definition for placeholder is incorrect. Try this one:

```php
$routes->addPlaceholder(':category', '[a-z\x{0980}-\x{09ff}-]+');
$routes->addPlaceholder(':pslug', '[a-zA-Z0-9\x{0980}-\x{09ff}-]+');

$routes->add('(:category)/(:pslug)', 'Single::index/$1/$2');
```

PR #3172, opened 2020-06-26
Add Unicode support for regular expressions in router

michalsn:
**Description**
This PR adds Unicode support for regular expressions in the router.

Ref: #3169

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


samsonasik:
probably add fix for https://github.com/codeigniter4/CodeIgniter4/issues/3180 for url_title helper as well?

michalsn:
Thanks, @samsonasik. I think I will address this in a separate PR and apply the changes to the `mb_url_title()`, since original helper should probably stay as it is.

PR #3179, opened 2020-06-28
Use explicit 'PHP_EOL' in wordwrap

paulbalandan:
**Description**
Running `CLITest` on Windows will likely fail on the `testWrap` method giving this error message:

image

Upon investigation, this was because `CLI::wrap` uses the `wordwrap` function and sets the third parameter into its default value of `\n` instead of the platform-specific `PHP_EOL`. This PR adds that.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -785,7 +785,7 @@ class CLI
 
 		$max = $max - $pad_left;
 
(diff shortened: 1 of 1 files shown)


michalsn:
Thanks!

PR #3183, opened 2020-06-29
Add apidocs action

paulbalandan:
**Description**
Adds the action to update API docs. ~~I have made this barely working up to `Make the new API docs` step. The deploy step fails for the username. Perhaps it is because I don't have credentials to push to `codeigniter4/api` in my testing.~~

~~Another is the use of the phar file. I'm more inclined to download the phar file per run rather than maintain the phar in the codebase so as to save disk space. But, I cannot make the download through curl successful. But if you think otherwise, it's fine.~~

After a series of runs, had finally made this turn green.
[Workflow run](https://github.com/paulbalandan/CodeIgniter4/actions/runs/157498787)
[API push](https://github.com/paulbalandan/api)

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.github/workflows/apidocs-action.yml
+++ b/.github/workflows/apidocs-action.yml
@@ -13,7 +13,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout
+      - name: Identify
+        run: |
+          git config --global user.email "action@github.com"
+          git config --global user.name "${GITHUB_ACTOR}"
+
+      - name: Checkout repo
         uses: actions/checkout@v2
 
       - name: Setup PHP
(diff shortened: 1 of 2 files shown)


MGatner:
@paulbalandan when I was creating the deployment action I forked all the affected repos and modified the organization section of the action so I can test it local to my GitHub account first. Might be a solution if you think there is a permissions issue.

paulbalandan:
Wow. Thanks! I'll try this later this evening.

paulbalandan:
I think this is ready. Requesting review from @lonnieezell or @MGatner .

MGatner:
Were you able to test it local to your org? If so could you link action results please?

paulbalandan:
Kindly see the link above.

MGatner:
“Process exited with code 23”

https://github.com/codeigniter4/CodeIgniter4/actions/runs/157699136

paulbalandan:
Didn't see that coming. I'll check later in the morning what is missing still.

MGatner:
It could be an issue for @lonnieezell, like ensuring the secret token has access to the API repo.

paulbalandan:
Visually inspecting this action's syntax against the green built run in mine, I think the problem is the missing `cd ./source` before the curl call. Might have overlooked that while copying. I'll send a PR for that later.

Issue #3180, opened 2020-06-28
Bug: Url Helper have a bug


mmrtonmoybd:
In url helper have a method url_title(). It can makes a slug from string. It also support Unicode. But it not totally support Unicode. 
I tried to do a string to slug but codeigniter gives output:
$cname = "মশিউর রহমান";
echo url_title($cname, '-', TRUE);
Output: মশউর-রহমন

Expect: মশিউর-রহমান

I do few changes in system/Helpers/url_helper.php 

In before 
///////////////////////////////
```
function url_title(string $str, string $separator = '-', bool $lowercase = false): string
	{
		$q_separator = preg_quote($separator, '#');

		$trans = [
			'&.+?;'                   => '',
			'[^\w\d _-]'              => '',
			'\s+'                     => $separator,
			'(' . $q_separator . ')+' => $separator,
		];

		$str = strip_tags($str);
		foreach ($trans as $key => $val)
		{
			$str = preg_replace('#' . $key . '#iu', $val, $str);
		}

		if ($lowercase === true)
		{
			$str = mb_strtolower($str);
		}

		return trim(trim($str, $separator));
	}
}
```
/////////////////////////////
I do few changes
In after
```
function url_title(string $str, string $separator = '-', bool $lowercase = false, string $patran = ''): string
	{
		$q_separator = preg_quote($separator, '#');

		if ($patran != '') {
		  $trans = [
			'&.+?;'                   => '',
			$patran              => '',
			'\s+'                     => $separator,
			'(' . $q_separator . ')+' => $separator,
		];
		} else {
		   $trans = [
			'&.+?;'                   => '',
			'[^\w\d _-]'              => '',
			'\s+'                     => $separator,
			'(' . $q_separator . ')+' => $separator,
		];
		}

		$str = strip_tags($str);
		foreach ($trans as $key => $val)
		{
			$str = preg_replace('#' . $key . '#iu', $val, $str);
		}

		if ($lowercase === true)
		{
			$str = mb_strtolower($str);
		}

		return trim(trim($str, $separator));
	}
} 
```
////////////////

$cname = "মশিউর রহমান";
url_title($cname, '-', TRUE, '[^a-z0-9\x{0980}-\x{09ff}]');
Output: মশিউর-রহমান
It works and output same as my expected.
I add a pattern option that anyone can replace there string as they want.
I hope the changes will be add next version. If I do any worng. Please comment on this issue.

PR #3185, opened 2020-06-29
Improvements to the mb_url_title() helper

michalsn:
**Description**
This PR adds Unicode support to the `mb_url_title()` helper. See: #3180

I don't want to sound totally ignorant, but it would be great if someone with better knowledge of the non-Latin alphabet could check if this is working as it should. Right now this is purely based on feedback from @mmrtonmoybd.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -618,7 +618,7 @@ if (! function_exists('url_title'))
 if (! function_exists('mb_url_title'))
 {
 	/**
-	 * Create URL Title that takes into account accented characters
+	 * Create URL Title that takes into account Unicode and accented characters
 	 *
 	 * Takes a "title" string as input and creates a
 	 * human-friendly URL string with a "separator" string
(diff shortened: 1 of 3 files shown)


ivantcholakov:
1. I suppose you mean UTF-8.

2. The function convert_accented_characters() has its old name that is understandable for people with Latin-origin language. But generally speaking it is transliteration of whatever text you have (even Chinese glyphs) into ASCII. And such a transliteration from cyrillic alphabet to ASCII is standartized, it is language-dependent - there is a Bulgarian way, there is a Russin way, etc. So, even without immediate implementation the signature of this function should be convert_accented_characters($string, $language).

3. Maybe it would easier for supporting UTF-8, not such fragmented efforts to be done. Maybe it would be better a set of UTF-8-aware functions to be implemented first for being maintained easily, something like Kohana did many years ago with their UTF8 class. And then this set to be used where it is appropriate. At the moment, I think that hardly readable regular expressions in many places might bring difficulties.

ivantcholakov:
Here is a collection of my patches on CI3 that worked well for me for years: https://github.com/ivantcholakov/starter-public-edition-4/blob/v4.4.6/platform/common/helpers/MY_text_helper.php

MGatner:
I will let @michalsn take a look at those first.

mmrtonmoybd:
@MGatner now mb_url_title() is not work perfectly as before. I test it when @michalsn  add mb_url_title() and it works as expect, now it is not working as expect.

michalsn:
Thanks for the feedback @ivantcholakov!
1. Yes, I meant UTF-8
2. You're right, I guess everything should be "translated" to ASCII with the relation to a given language.
3. I have to think about how to address this correctly but a separate class is probably a good idea.

Obviously changes in this PR are not a good idea.

PR #3192, opened 2020-07-01
ArgType 'String' is not valid in MessageFormatter class (fixes #3191)

### LINKED ISSUE #3191: Bug: MessageFormatter argType of "String" isn't valid
**Direction**
System/Language/Language.php uses PHP's MessageFormatter class for international support. In some cases, an argType of "string" is used, which is not a valid argType. An example is in System/Language/en/Email.php, line 26. This came to light for me when trying to simply send an email, which did send, but threw an error trying to log the success message. Having done some testing, I found this error, simply removing the "string" from the type fixes it. Looking through the language files, there are very few instances where this appears

**CodeIgniter 4 version**
4.0.2

**Affected module(s)**
System/Language/en/*.php

**Context**
 - OS: Ubuntu 20.04
 - Web server: Apache 2.4
 - PHP version: 7.4.7

HughieW:
First noticed when sending an email using the Email class, a few of the language strings in System/Language/en/*.php use an argtype of "string" which isn't valid in the PHP MessageFormatter class. This causes an error to be logged, or in a production environment, an error to be shown and the script halted. Removing this argType (leaving it undefined) fixes this completely.

There was also a typo in System/Language/en/Images.php in the same place (closing bracket instead of a closing brace) which I fixed at the same time.

### DIFF
--- a/system/Language/en/Email.php
+++ b/system/Language/en/Email.php
@@ -23,7 +23,7 @@ return [
    'sendFailurePHPMail'   => 'Unable to send email using PHP mail(). Your server might not be configured to send mail using this method.',
    'sendFailureSendmail'  => 'Unable to send email using PHP Sendmail. Your server might not be configured to send mail using this method.',
    'sendFailureSmtp'      => 'Unable to send email using PHP SMTP. Your server might not be configured to send mail using this method.',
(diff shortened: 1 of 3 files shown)


michalsn:
Thanks!

Issue #3188, opened 2020-06-30
Bug: pagination not working with segment


mmrtonmoybd:
Pagination not working with segment, but the page query parameter pagination is still working.
https://codeigniter.com/user_guide/libraries/pagination.html#specifying-the-uri-segment-for-page
 In user guide says
 “It is also possible to use a URI segment for the page number, instead of the page query parameter. Simply specify the segment number to use as the fourth argument. URIs generated by the pager would then look like https://domain.tld/model/[pageNumber] instead of https://domain.tld/model?page=[pageNumber].:”
In user guide given an example code 
```
$users = $userModel->paginate(10, ‘group1’, null, 3);
```
I am doing same thing. 
I do 
```
$data = [
         'get' => $post->where('status', 1)->where('admin_status', 1)->where('category', $category)->orderBy('pid', 'desc')->paginate(4, 'default', null, 4),
         'pager' => $post->pager
         ];
```
But it is not working. It's show an error 

TypeError
Argument 3 passed to CodeIgniter\Model::paginate() must be of the type int, null given, called in /storage/emulated/0/htdocs/code/app/Controllers/Category.php on line 14

SYSTEMPATH/Model.php at line 1140

Maybe it's a bug.

PR #3194, opened 2020-07-01
Fix default value for page in Model::paginate()

michalsn:
**Description**
This PR fixes the default value for the `$page` parameter in Model::paginate().

Ref: #3188

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide


mmrtonmoybd:
@michalsn pagination not working now.
It shows same data on every page , data isn't change. 
I have 16 rows. I want to show 4 data per page. But it shows same data every page. Maybe it's not fixed yet.

michalsn:
Apparently there is another bug with `segment` support. I will look into it tomorrow.

mmrtonmoybd:
Thank you @michalsn .
I am making a magazine on codeigniter. So I need pagination to display per page posts by category. I already doing pagination with page query parameter, but my boss rejected it because it is not seo friendly url. 
Almost all works are done in this project. I already two times rejected for not seo friendly pagination link.

Pagination links() method also have a problem. 
If I didn't need to use pagination group, but I need to use my own pagination template. 
So I basically do like 
```
<?= $pager->links(null, 'moshiur') ?>
```
It gives me a error, because links first Argument must be a string null given. 

Is it possible to give null value on links method both Arguments ?

michalsn:
@mmrtonmoybd  instead of `null` you should use `default` for group name.

mmrtonmoybd:
@michalsn do you fixed paginate show results with segment bug ?

michalsn:
@mmrtonmoybd Yes, I hope it's fixed. The tests fail because of a random test failure.

mmrtonmoybd:
Well done @michalsn
I am seeing that you fixed many issue in codeigniter like Super Man. You are the coding hero. Thank you 😊❤️❤️
Now it is working. But codeigniter ignore your pull request.

michalsn:
@mmrtonmoybd I was only waiting for you to confirm that it's working fine. Thanks for your feedback!

mmrtonmoybd:
@michalsn 
Codeigniter $pager->links() array contain the active boolean not working.
It return false for every page.
My current page is 5, but it return active=false for the page 5.


array(3) { [0]=> array(3) { ["uri"]=> string(42) "http://localhost:8080/category/exclusive/3" ["title"]=> int(3) ["active"]=> bool(false) } [1]=> array(3) { ["uri"]=> string(42) "http://localhost:8080/category/exclusive/4" ["title"]=> int(4) ["active"]=> bool(false) } [2]=> array(3) { ["uri"]=> string(42) "http://localhost:8080/category/exclusive/5" ["title"]=> int(5) ["active"]=> bool(false) } }

PR #3197, opened 2020-07-02
Hide in bootstrap4

Sosko:
If the CSP policy is on, the style wouldn't work. So if the project use Bootstrap 4 or Bootstrap 3, it will be better to add thes classes.

### DIFF
--- a/app/Config/Honeypot.php
+++ b/app/Config/Honeypot.php
@@ -13,19 +13,10 @@ class Honeypot extends BaseConfig
 	public $hidden = true;
 	
 	/**
-         * Hide the Honeypot. If not set or null, the style parameter will be use. Otherwise it will be set the class parametr by inserted string.
-         * Example:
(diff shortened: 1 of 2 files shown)


michalsn:
Good point, but we can't accept hardcoded classes which work only with certain CSS framework. The proper way to solve this would be adding a new config variable to the `Config/Honeypot.php` e.g. `$container`. It would contain HTML which can be edited easily. 

Will you be able to make these changes?

Sosko:
Ok, I tried, and make it simple as possible.

Sosko:
This small thing, and how long it takes :-D
By the way, this thing with style, bother me too, when I use the debug bar. I know, this thing is only for debugging on local machine, but create CI own style file, which will be by default in the public folder, sound for me like the best strategy.

michalsn:
Thank you for taking care of this.

I don't think we want to have a special CSS file shipped with the framework.

PR #3203, opened 2020-07-02
Fixed #3199

### LINKED ISSUE #3199: Bug: Error exception layout when its an error inside a view where code can go outside the limits
It's normal for view code to get long lines of HTML. Today I had a simple error of an "Undefined index" with the this output.

Screenshot from 2020-07-02 15-35-20

The problem was in that line, however I can't see the complete line there because it goes outside the dark background.

When I make a text selection this is what I can see:

Screenshot from 2020-07-02 15-37-24

Maybe the dark container could have a horizontal scroll?

I'm reporting this here, but later I'll try and add a solution with a pull request.

mpmont:
Added just a overflow-x: scroll to mimic the expected result reported in the issue #3199

### DIFF
--- a/app/Views/errors/html/debug.css
+++ b/app/Views/errors/html/debug.css
@@ -64,6 +64,7 @@ p.lead {
     border-radius: 5px;
     font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
     margin: 0;
+    overflow-x: scroll;
 }
 .source span.line {
     line-height: 1.4;


MGatner:
Good fix! This might need to be added to the SCSS. @LittleJ are you around to weigh in?

MGatner:
I'll give @LittleJ another day or so, but I think this is okay since right now only the Debug Toolbar is handled by SASS (https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/css.rst)

mpmont:
Yes it does look like SCSS its being used just there.

LittleJ:
Hey ! Yes, the toolbar is managed by the SCSS files as you pointed :-) But no link with the work on Trello regarding the website though :-D

LittleJ:
I will be able to have a look at this on this the incoming week. But @mpmont, you can do the changes yourself if you want. It would even be better 😄 

Instructions have been left in the file that @MGatner mentioned: https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/css.rst

SASS files are in the folder: admin/css/debug-toolbar

And I am always available on Github or Slack for any question 😉

MGatner:
Thanks LittleJ. Since this is not related to the debug toolbar I think it is fine as-is.

PR #3207, opened 2020-07-03
Fix for getting database error

### LINKED ISSUE #3204: Bug: MySQL Errors are not reported
**Describe the bug**
When I insert the data to database using a model, If any MySQL errors exits, Its not being reported.

```php
$userModel = new \App\Models\UserModel();

$newUser = [
	'name'	 	=> $this->request->getVar('name'),
	'username'	=> $this->request->getVar('username'),
	'email'		=> $this->request->getVar('email'),
	'password'	=> $this->request->getVar('password'),
];
 
$user_id = $userModel->insert($newUser);

// here the name column does not exists in the database 
$user_id returns '0'

//But when I try 

$userModel->errors();  //Returns nothing 

```

The only way I can see there is an error is var_dump($userModel) // and go through the whole array to see the error
  
and here is my model

```php 
namespace App\Models;

use CodeIgniter\Model;

class UserModel extends Model
{
    protected $table      = 'users';
    protected $primaryKey = 'id';

    protected $returnType     = 'array';
    protected $useSoftDeletes = true;

    protected $allowedFields = ['name', 'email', 'username','password'];

    protected $useTimestamps = true;
    protected $createdField  = 'created_at';
    protected $updatedField  = 'updated_at';
    protected $deletedField  = 'deleted_at';

    protected $validationRules    = [];
    protected $validationMessages = [];
    protected $skipValidation     = false;
 
} 

```

**CodeIgniter 4 version**
CodeIgniter 4.0.3

**Affected module(s)**
Models

**Context**
 - OS: macOS Mojave
 - Web server  Apache
 - PHP version : 7.4

michalsn:
**Description**
This PR fixes a bug regarding the method used to get the database error. All DB drivers use `error()` method but in our interface, we had `getError()` which wasn't really used anywhere except for the `Model` class. 

See: #3204

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -373,12 +373,12 @@ abstract class BaseConnection implements ConnectionInterface
 
 		$this->connectTime = microtime(true);
 
-		try 
+		try
 		{
 			// Connect to the database and set the connection ID
 			$this->connID = $this->connect($this->pConnect);
-		} 
-		catch (\Throwable $e) 
+		}
+		catch (\Throwable $e)
 		{
 			log_message('error', 'Error connecting to the database: ' . $e->getMessage());
 		}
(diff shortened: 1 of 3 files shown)


MGatner:
Yikes, that was an odd oversight! I think this is okay as we don't officially support any other services using that interface but I'd like @lonnieezell to make the call.

lonnieezell:
Oh, wow. Looks good to me. Thanks!

PR #3209, opened 2020-07-03
Cast currentPage value to integer in Pager class

michalsn:
**Description**
This PR fixes lack of casting for `currentPage` when using segments.

See: https://github.com/codeigniter4/CodeIgniter4/pull/3194#issuecomment-653206930

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Pager/Pager.php
+++ b/system/Pager/Pager.php
@@ -559,7 +559,7 @@ class Pager implements PagerInterface
 		{
 			try
 			{
-				$this->groups[$group]['currentPage'] = (int) $this->groups[$group]['uri']->getSegment($this->segment[$group]);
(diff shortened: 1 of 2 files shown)


MGatner:
This might be far-fetched, but casting the return of `getSegment()` to `int` would cause any accidental non-numeric string as `1`, or `''` (the default) as `0`. Do we need to be concerned for cases where the current `URI` is set to `$silent = true` and thus returns an unexpected value? Might be overkill, but the fix is easy by adding `$uri->setSilent(false);`

michalsn:
As for type casting - yes, I know, but in most cases, this would be a wrong configuration. The "funny" part is that we actually need `currentPage` to be `0` sometimes (when the segment is empty `""`). This is a bit strange but in that case, we use this value only internally.

I've added the `setSilent(false)` to the URI class call.

Issue #3134, opened 2020-06-19
Make the Model work without auto_increment primary key


MashinaMashina:
**Describe the bug**
1. Table:
```sql
CREATE TABLE test (
key varchar(20),
value integer,
primary key (key)
)
```
2. Code:
```php
$testModel = new TestModel;
$testModel->insert($data);
```
3. Error:  pg_query(): Query failed: ERROR: lastval is not yet defined in this session 

I can create PR, but without description. I think to create property CodeIgniter\Model::$has_autoincrement and add checking it in \system\Model.php:731, but I don't know if is a good idea
```php
if ($result and $this->has_autoincrement)
{
	$this->insertID = $this->db->insertID();
}
```

michalsn:
It's not really a bug - more like a feature request.

Contributions are always welcomed - but please be sure to add some tests, to make sure this change would not break anything. Also, take into consideration that we have to support `$this->insertID` in a model (it has to work) even if there is no `auto_increment` in the database set.

MashinaMashina:
For get inserted primary key value, the best way is using "RETURNING primaryKey" in query, but database driver dont know primary key, model can't change SQL string.

I dont know a way

michalsn:
@MashinaMashina I think you should create a PR. This way we can easier discuss your code. 

In general, if the "auto increment" option is disabled and there is no `data[primary_key]` available, we should throw an exception.

Additionally, we will need some tests and docs updates. If you're willing to work on this I will be glad to help with your questions if you will have any.

crustamet:
I wanted this feature before, this going to be commited ? or it will just remain like it is right now ?

michalsn:
@crustamet There is nothing we can merge yet. I'm not sure if @MashinaMashina will try to work on this though. If you're motivated enough, you can prepare a PR for this, but we will need some tests too.

MashinaMashina:
@michalsn Do we have tests for Model.php?

michalsn:
@MashinaMashina Yes, we have: https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/Live/ModelTest.php

MashinaMashina:
@michalsn  I try to create tests, but it runs on sqllite, not postgres https://github.com/MashinaMashina/CodeIgniter4/tree/Model_without_autoincrement_key

Is good this?

michalsn:
[...]

PR #3211, opened 2020-07-03
Model without autoincrement key

MashinaMashina:
https://github.com/codeigniter4/CodeIgniter4/issues/3134

PR needs description to docs of CI4

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -736,8 +736,7 @@ class Model
 			}
 			else
 			{
-				// ??
-				// INSERT INTO {$this->table} %data% RETURNING {$this->primaryKey}
+				throw new \Exception('Can\'t return primary key');
 			}
 		}
 		elseif ($result)
--- a/system/Model.php
+++ b/system/Model.php
@@ -102,7 +102,7 @@ class Model
 	 *
 	 * @var boolean
 	 */
-	protected $hasAutoincrement = true;
+	protected $useAutoIncrement = true;
 
 	/**
 	 * Last insert ID
@@ -643,7 +643,7 @@ class Model
 	 *
 	 * @return integer
 	 */
-	public function getInsertID(): int
+	public function getInsertID()
 	{
 		return $this->insertID;
 	}
@@ -730,20 +730,17 @@ class Model
 		// If insertion succeeded then save the insert ID
 		if ($result)
 		{
-			if ( ! $this->hasAutoincrement)
+			if (! $this->useAutoIncrement)
 			{
-				if (isset($data[$this->primaryKey]))
+				if (empty($data[$this->primaryKey]))
 				{
-					$this->insertID = $data[$this->primaryKey];
+					throw DataException::forEmptyDataset('pk');
 				}
-				else
-				{
-					throw new \Exception('Can\'t return primary key');
-				}
+				$this->insertID = $data[$this->primaryKey];
 			}
 			else
 			{
-				$this->insertID = $this->db->insertID();
+				$this->insertID = (int) $this->db->insertID();
 			}
 		}
 
--- a/system/Model.php
+++ b/system/Model.php
@@ -641,7 +641,7 @@ class Model
 	/**
 	 * Returns last insert ID or 0.
 	 *
-	 * @return integer
+	 * @return integer|mixed
 	 */
 	public function getInsertID()
 	{
--- a/system/Model.php
+++ b/system/Model.php
@@ -728,20 +728,23 @@ class Model
 				->insert();
 
 		// If insertion succeeded then save the insert ID
-		if ( ! $this->hasAutoincrement)
+		if ($result)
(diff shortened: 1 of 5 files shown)


MashinaMashina:
@michalsn why did not start migration from https://github.com/codeigniter4/CodeIgniter4/blob/90b76f20182ff96a53f6bf2273af6b2f5a0c08f5/tests/_support/DatabaseTestMigrations/Database/Migrations/2020-07-02-165706_model_without_autoincrement.php ?

michalsn:
@MashinaMashina Wrong `namespace`. You can fix it or add code responsible for creating a new table to the migration file that already exists. 

I'll try to look into your code later tonight.

MashinaMashina:
@michalsn  then it bug in spark. I created with "php stark migrate:create"

michalsn:
@MashinaMashina This should help with tests:
* please, move: `tests/_support/DatabaseTestMigrations/Database/Migrations/2020-07-02-165706_model_without_autoincrement.php` 
* to: `tests/_support/Database/Migrations/2020-07-02-165706_model_without_autoincrement.php`
* and change namespace to: `Tests\Support\Database\Migrations`

MashinaMashina:
I dont know what I write right comment about function getInsertID() about returns types

system/Model.php
```diff
@@ -730,20 +730,17 @@ public function insert($data = null, bool $returnID = true)
...
 			{
-				if (isset($data[$this->primaryKey]))
+				if (empty($data[$this->primaryKey]))
 				{
-					$this->insertID = $data[$this->primaryKey];
-				}
-				else
-				{
-					throw new \Exception('Can\'t return primary key');
+					throw DataException::forEmptyDataset('pk');
```
michalsn, line 737: I know I suggested it, but now I just checked the message behind this exception and I think we should create a new exception.
```suggestion
					throw DataException::forEmptyPrimaryKey();
```
This would require to create a new method in `system/Database/Exceptions/DataException.php` file and also an entry to the language file - something like "You must provide a Primary Key value".

system/Model.php
```diff
@@ -641,7 +641,7 @@ public static function classToArray($data, $primaryKey = null, string $dateForma
 	/**
 	 * Returns last insert ID or 0.
 	 *
-	 * @return integer
+	 * @return integer|mixed
```
michalsn, line 644: Please, make this `integer|string`

system/Model.php
MashinaMashina: I think the primary key cannot always be transmitted and this should not be a mistake.

Now I don't know this case, in example - primary key is random value from database. We cant got it from LASTVAL() and it isn't in $data array.

Best way edit insert SQL string to
INSERT INTO table (v) VALUES('val') RETURNING {primary key}

MashinaMashina: I don't know how to do it

tests/_support/Database/Migrations/2020-07-02-165706_model_without_autoincrement.php
```diff
@@ -1,4 +1,4 @@
-<?php namespace Tests\Support\DatabaseTestMigrations\Database\Migrations;
+<?php namespace Tests\Support\Database\Migrations;
```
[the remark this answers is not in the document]
MashinaMashina: It is already there

system/Model.php
michalsn: I understand your reasoning behind mentioning the `RETURNING` in the `INSERT` query, but this is more like a separate problem/idea that should be addressed in a different PR.

Here we are trying to deal only with the scenario where a developer is responsible for providing the primary key.

system/Model.php
```diff
@@ -723,7 +730,21 @@ public function insert($data = null, bool $returnID = true)
...
+			{
+				if (empty($data[$this->primaryKey]))
+				{
+					$this->insertID = null;
+				}
+				else
+				{
+					$this->insertID = $data[$this->primaryKey];
+				}
+			}
```
michalsn, line 743: Instead of using `$data` array we should use `$eventData`. This way developers can take advantage of model filers.
```suggestion
			if (! $this->useAutoIncrement)
			{
				$this->insertID = $eventData['data'][$this->primaryKey];
			}
```
Also, we need to check if there is a primary key set when `$useAutoIncrement === false`, so right after `beforeInsert` (line 723), we should add a check, like:

```php
$eventData = $this->trigger('beforeInsert', ['data' => $data]);

if (! $this->useAutoIncrement && empty($eventData['data'][$this->primaryKey]))
{
        throw DataException::forEmptyPrimaryKey();
}
```
We need to create a new method in `system/Database/Exceptions/DataException.php` file and also an entry to the language file - something like "You must provide a Primary Key value". If you're not sure how to handle this, just see how other exceptions in this class look like.

PR #3212, opened 2020-07-03
getSegment() fixed with conditions changed

### LINKED ISSUE #2962: Bug: URI class - working with segments is really strange
**Describe the bug**
I think there is a problem in the way that some methods are handling segments right now.
Here is an example:
```php
$uri = new \CodeIgniter\HTTP\URI('http://something.com/hello/there');

// will return ['hello', 'there']
d($uri->getSegments());
// will return 'hello' - ok
d($uri->getSegment(1));
// will return 'there' - ok
d($uri->getSegment(2));
// will return '' - why? it's not there
d($uri->getSegment(3));
// will throw an exception - ok
// not really easy to work with but this is a different story
d($uri->getSegment(4));
```
Here we go with a really strange part:
```php
$uri = new \CodeIgniter\HTTP\URI('http://something.com/hello/there');

// will return '' - why? it's not there
d($uri->getSegment(3));
// will NOT throw an exception (even if segment 3 is not set)
d($uri->setSegment(4, 'honey'));
// will return '' - that's surprising considering no error when
// we set it a line before
d($uri->getSegment(4));
// will return 'honey'
d($uri->getSegment(3));
```
Actually I don't think it's very complicated to fix, but the main problem are BC changes. People can already rely on the empty string that is returned in some cases. So the question is how should it be handled.

Also working with `getSegment()` and throwing an exception when segment is not there is not very easy to work with. So it would be good to have something to surpass this without the need of catching the exception every time. One idea was proposed here: #2952

**CodeIgniter 4 version**
All versions

**Affected module(s)**
URI class

**Expected behavior**
```php
$uri = new \CodeIgniter\HTTP\URI('http://something.com/hello/there');

// should throw an exception
d($uri->getSegment(3));

```

```php
$uri = new \CodeIgniter\HTTP\URI('http://something.com/hello/there');

// should throw an exception
d($uri->getSegment(3));
// should throw an exception
d($uri->setSegment(4, 'honey'));
// should throw an exception
d($uri->getSegment(4));
```

rajat315315:
Fixes #2962 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -508,13 +508,14 @@ class URI
 	{
 		// The segment should treat the array as 1-based for the user
 		// but we still have to deal with a zero-based array.
-		$number -= 1;
 
-		if ($number > count($this->segments) && ! $this->silent)
+		if (($number <= 0 || $number > count($this->segments)) && ! $this->silent)
 		{
 			throw HTTPException::forURISegmentOutOfRange($number);
 		}
 
+		$number -= 1;
+		
 		return $this->segments[$number] ?? $default;
 	}
 
--- a/system/Pager/Pager.php
+++ b/system/Pager/Pager.php
@@ -211,7 +211,12 @@ class Pager implements PagerInterface
 
 		if ($segment > 0 && $this->groups[$group]['currentPage'] > 0)
 		{
-			$page = $this->groups[$group]['currentPage'];
+				try
+				{
+					$this->groups[$group]['currentPage'] = $this->groups[$group]['uri']->getSegment($this->segment[$group]);
+					$page = $this->groups[$group]['currentPage'];
+				}
+				catch (\CodeIgniter\HTTP\Exceptions\HTTPException $e){}
 		}
 
 		$perPage                             = $perPage ?? $this->config->perPage;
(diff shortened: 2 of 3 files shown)


lonnieezell:
Can you fix up the failing test please? 

https://travis-ci.org/github/codeigniter4/CodeIgniter4/jobs/704736451

rajat315315:
Can somebody clarify the Unit test at:
[[tests/system/Pager\PagerTest.php#L=428]](tests/system/Pager\PagerTest.php#L=428)
Is it correct?

I think it should be:
$this->assertEquals(10, $this->pager->getCurrentPage());

michalsn:
This would be a BC change. Surprisingly this is not covered by the URITest class.

Let's say we have this address: `http://example.com/one/two`. Now, calling `$uri->getSegment(3)` originally would return `""` (empty string) but after this change, it will throw an exception.

In URI class we have a "rule" that restrains throwing exception until the segment isn't greater than `n+1` (where `n` i actually defined segment of URI).

Anyway, I think that #2962 is pretty much covered by introducing `setSilent()` method. Other things can't be changed now without introducing BC changes.

ivantcholakov:
At the transition CI2 --> CI3 a convention was adopted. If a method does not "know" what to return, it returns null, not an empty string. Is this valid for CI4?

michalsn:
@ivantcholakov No, CI4 has quite different rules - at least by default.

paulbalandan:
@michalsn are we still keeping this? or are we deferring this for a BC breaking version jump? If no, perhaps we can close this?

michalsn:
Personally, I think people rely too much on the current implementation to change it in 4.1. We can revisit this at some point but I'll close it for now.

kenjis:
See #7249

Issue #3216, opened 2020-07-04
Bug: no Filter arguments passed


tangix:
**Describe the bug**
When running Filters, before() and after() doesn't get arguments passed to them. This is contrary to the information in the docs:
https://codeigniter.com/user_guide/incoming/routing.html#applying-filters
Changing the method signature is not possible since this would break FilterInterface validation.

**CodeIgniter 4 version**
4.0.3

**Affected module(s)**
\CodeIgniter\Filters\Filter
\CodeIgniter\Filters\FilterInterface

**Expected behavior, and steps to reproduce if appropriate**
I should be able to get the arguments from the route into my Filter. With a route specified as:
`$routes->add('users/delete/(:segment)', 'AdminController::index', ['filter' => 'admin-auth:dual,noreturn']);`
I'd expect to get access to `['dual', 'noreturn']` in my Filter implementation.

**Context**
- PHP 7.3.19

hermitprime:
Same issue here. The user guide does not give any method to retrieve the parameter. So the bug is fixed?

tangix:
> Same issue here. The user guide does not give any method to retrieve the parameter on the filters. So the bug is fixed?

To the best of my knowledge there is no easy way to fix this without breaking old code implementing the FilterInterface. My PR created a new interface with the correct function signature and changing the logic in the library to check for the new interface as well as the old. (Unfortunately my PR doesn't pass the tests on github but does locally, so I don't know if the PR will be considered at all.)
The core devs will have to look at this I guess and possibly find a better way forward. Currently I have applied my patch to the framework, but that's not the correct way to do it.

tangix:
Fixed in 4.0.4

PR #3217, opened 2020-07-04
Added new interface FilterInterfaceWithArguments

tangix:
**Description**
Fixing #3216 by implementing a new interface to avoid breaking already implemented Filters.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Filters/FilterInterfaceWithArguments.php
+++ b/system/Filters/FilterInterfaceWithArguments.php
@@ -1,0 +1,85 @@
+<?php
+
+/**
+ * CodeIgniter
+ *
+ * An open source application development framework for PHP
+ *
+ * This content is released under the MIT License (MIT)
+ *
+ * Copyright (c) 2014-2019 British Columbia Institute of Technology
+ * Copyright (c) 2019-2020 CodeIgniter Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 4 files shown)


tangix:
All tests pass locally, so I am not sure what I'm seeing...

michalsn:
@tangix Can you pull the latest changes from the dev branch and add `Services::reset()` to the `FiltersWithArgumentsTest::setUp()` just after the call to the `parent::setUp()`?

tangix:
Done.

michalsn:
@lonnieezell @MGatner

We probably need to update the user guide for this, but besides that - are you good with these changes?

tangix:
> @lonnieezell @MGatner
> 
> We probably need to update the user guide for this, but besides that - are you good with these changes?

Also, please consider if there is a better way to do this without break backward compatibility.

lonnieezell:
So basically all we're doing is creating a new interface that accepts `$args`, correct? In my mind, this is a bug since it is expected to be working currently. So all we need to do is add the argument to the interface and method. And maybe add some better docs, right? 

While I realize this is technically a BC break, it won't break any existing code. And should be working actually since Myth:Auth uses those filters, though I haven't tried it in a while.

tangix:
@lonnieezell 
Yeah, I was torn between changing the current interface (breaking code) or adding a new interface (possibly cluttering down the framework)...

lonnieezell:
@tangix And I totally respect that you were trying to avoid the BC break here and really appreciate it! I just think this qualifies as a bug so we can fix it the right way.

tangix:
> @tangix And I totally respect that you were trying to avoid the BC break here and really appreciate it! I just think this qualifies as a bug so we can fix it the right way.

I can live with that and I think that would be the correct way forward to avoid another interface and added complexity to CI4.

tangix:
@lonnieezell 
I see 4.0.4 is around the corner but with no fix for this issue. Do you want me to create a new PR based on develop branch, change the FilterInterface (introduce a BC breaking fix) and also check over the docs?

MGatner:
@tangix the goal for 4.0.4 is within the next 24 hours. If you can get this reworked and submitted today I think it would be a great addition given that developers using it will require manual intervention, and a point release is a great time for that.

tangix:
OK, created #3316 fixing the actual bug instead of adding a new interface.

PR #3167, opened 2020-06-25
Add the upload artifact action

paulbalandan:
**Description**
Docs action still fails because it cannot `cp` the user guide folder. This adds the `actions/upload-artifact`. This is also referenced in [`ammaraskar/sphinx-action-test` default workflow](https://github.com/ammaraskar/sphinx-action-test/blob/master/.github/workflows/default.yml).

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #3218, opened 2020-07-04
Remove underscore in UG build path

paulbalandan:
**Description**
In [comment to #3167 ](https://github.com/codeigniter4/CodeIgniter4/pull/3167#issuecomment-653716546), docs still fails the build process. Done a little checking and found that we used the build path as `_build` instead of `build` used in the `Makefile`.

Checked this locally for a green build: https://github.com/paulbalandan/CodeIgniter4/actions/runs/157521660

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.github/workflows/docs_nightly.yml
+++ b/.github/workflows/docs_nightly.yml
@@ -29,13 +29,13 @@ jobs:
     - uses: actions/upload-artifact@v2
       with:
         name: HTML Documentation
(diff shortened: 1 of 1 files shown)


MGatner:
Ooo that would be great if this got the Action going! This one has been a pain. Good catch!

PR #3220, opened 2020-07-04
Allow ignoring *.db inside folders

paulbalandan:
**Description**
Since we are ignoring sqlite DBs inside the `writable` folder, might as well ignore db files inside folders within the `writable`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


michalsn:
Good idea, thanks!

PR #3222, opened 2020-07-04
Fix insert Entity object on postgres

MashinaMashina:
I add check that $data->{$primaryKey} is set, before add it to $data. Otherwise we get error on insert:
> pg_query(): Query failed: ERROR: null value in column "id" violates not-null constraint

because sql:
> INSERT INTO table (key, id) VALUES('value', NULL)


michalsn:
Good catch, thanks!

PR #3234, opened 2020-07-06
Feature seed command

abilioposada:
**Description**
Add seed command generation

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Database/CreateSeeder.php
+++ b/system/Commands/Database/CreateSeeder.php
@@ -130,7 +130,7 @@ class CreateSeeder extends BaseCommand
 			{
 				if ($namespace === $ns)
 				{
-					$homepath = realpath(reset($path));
+					$homepath = realpath(reset($path)) . DIRECTORY_SEPARATOR;
 					break;
 				}
 			}
--- a/system/Commands/Database/CreateSeeder.php
+++ b/system/Commands/Database/CreateSeeder.php
@@ -62,7 +62,7 @@ class CreateSeeder extends BaseCommand
 	 *
 	 * @var string
 	 */
-	protected $name = 'db:seed:create';
+	protected $name = 'make:seeder';
 
 	/**
 	 * the Command's short description
@@ -76,7 +76,7 @@ class CreateSeeder extends BaseCommand
 	 *
 	 * @var string
 	 */
-	protected $usage = 'db:seed:create [seeder_name] [options]';
+	protected $usage = 'make:seeder [seeder_name] [options]';
 
 	/**
 	 * the Command's Arguments
--- a/system/Commands/Database/CreateSeeder.php
+++ b/system/Commands/Database/CreateSeeder.php
@@ -143,7 +143,7 @@ class CreateSeeder extends BaseCommand
 		$template = <<<EOD
 <?php namespace $ns\Database\Seeds;
 
-use CodeIgniter\DataBase\Seeder;
+use CodeIgniter\Database\Seeder;
 
 class {name} extends Seeder
 {
--- a/system/Commands/Database/CreateSeeder.php
+++ b/system/Commands/Database/CreateSeeder.php
@@ -55,7 +55,7 @@ class CreateSeeder extends BaseCommand
 	 *
 	 * @var string
 	 */
-	protected $group = 'Database';
+	protected $group = 'Generators';
 
 	/**
 	 * The Command's name
--- a/system/Commands/Database/CreateSeeder.php
+++ b/system/Commands/Database/CreateSeeder.php
@@ -1,0 +1,174 @@
+<?php
+/**
+ * CodeIgniter
+ *
+ * An open source application development framework for PHP
+ *
+ * This content is released under the MIT License (MIT)
+ *
+ * Copyright (c) 2014-2019 British Columbia Institute of Technology
+ * Copyright (c) 2019-2020 CodeIgniter Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


system/Commands/Database/CreateSeeder.php
```diff
@@ -0,0 +1,174 @@
...
+		helper('inflector');
+		
+		$name = array_shift($params);
+
+		if (empty($name))
+		{
+			$name = CLI::prompt(lang('Seed.nameFile'));
+		}
+
+		if (empty($name))
```
paulbalandan, line 115: Instead of adding another `if` here, we can just add the `required` validation rule in the prompt above.

system/Commands/Database/CreateSeeder.php
```diff
@@ -0,0 +1,174 @@
...
+		if (! empty($ns))
+		{
+			// Get all namespaces
+			$namespaces = Services::autoloader()->getNamespace();
+
+			foreach ($namespaces as $namespace => $path)
+			{
+				if ($namespace === $ns)
+				{
+					$homepath = realpath(reset($path));
```
paulbalandan, line 133: Since we used `APPPATH` as the initial value and it ends with the DIRECTORY_SEPARATOR, this one also needs to end in a DIRECTORY_SEPARATOR.

system/Commands/Database/CreateSeeder.php
```diff
@@ -0,0 +1,174 @@
...
+				}
+			}
+		}
+		else
+		{
+			$ns = 'App';
+		}
+
+		// full path
+		$path = $homepath . '/Database/Seeds/' . $name . '.php';
```
paulbalandan, line 144: And the beginning forward slash in `/Database/Seeds/` needs to be removed.

system/Commands/Database/CreateSeeder.php
```diff
@@ -0,0 +1,174 @@
...
+
+		// Class name should be pascal case now (camel case with upper first letter)
+		$name = pascalize($name);
+
+		$template = <<<EOD
+<?php namespace $ns\DataBase\Seeds;
+
```
paulbalandan, line 151: I believe `DataBase` will not work with PSR-4 autoloading. Especially in case-sensitive filesystems.

lonnieezell:
Thanks for this. Please change the command name to `make:seeder`, though. I'd like to start getting more CLI generators in place eventually, so let's make a place for them to live.

abilioposada:
@paulbalandan I was following [system/Commands/Database/CreateMigration.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Commands/Database/CreateMigration.php) file, but I can do all those changes

system/Commands/Database/CreateSeeder.php
abilioposada: It shows the full path to the file
Screenshot from 2020-07-06 12-42-32

system/Commands/Database/CreateSeeder.php
```diff
@@ -0,0 +1,174 @@
...
+ */
+class CreateSeeder extends BaseCommand
+{
+	/**
+	 * The group the command is lumped under
+	 * when listing commands.
+	 *
+	 * @var string
+	 */
+	protected $group = 'Database';
```
lonnieezell, line 58: Sorry, we should probably change this also, come to think of it. I think "Generators" might work.

system/Commands/Database/CreateSeeder.php
lonnieezell: Correct, the 'B' should be lowercase.

lonnieezell:
@abilioposada Yes, please follow @paulbalandan suggestions. Things have been formalized more since the CreateMigration command was originally written. It will be revisited at some point in the future. :)

And thanks for this!

system/Commands/Database/CreateSeeder.php
paulbalandan: Yes. That's because of the `str_replace` at the last line. But if you'll `var_dump` the `$path` assuming we don't supply the namespace option, we will get this.
image

Notice the `app\/Database`.

system/Commands/Database/CreateSeeder.php
paulbalandan: The `B` in `DataBase` below in line 346 should be lowercase too.

system/Commands/Database/CreateSeeder.php
```diff
@@ -0,0 +1,168 @@
...
+		$path = $homepath . 'Database/Seeds/' . $name . '.php';
+
+		// Class name should be pascal case now (camel case with upper first letter)
+		$name = pascalize($name);
+
+		$template = <<<EOD
+<?php namespace $ns\Database\Seeds;
+
+use CodeIgniter\DataBase\Seeder;
+
```
paulbalandan, line 147: Sorry, the B here also should be lowercase.

system/Commands/Database/CreateSeeder.php
```diff
@@ -0,0 +1,168 @@
...
+				if ($namespace === $ns)
+				{
+					$homepath = realpath(reset($path)) . DIRECTORY_SEPARATOR;
+					break;
+				}
+			}
+		}
+		else
+		{
+			$ns = 'App';
```
paulbalandan, line 134: I was thinking if we can use `APP_NAMESPACE`  if it is defined and use `App` otherwise. Something like this:

`$ns = defined('APP_NAMESPACE') ? APP_NAMESPACE : 'App';`

system/Commands/Database/CreateSeeder.php
```diff
@@ -0,0 +1,168 @@
...
+		$template = str_replace('{name}', $name, $template);
+
+		helper('filesystem');
+		if (! write_file($path, $template))
+		{
+			CLI::error(lang('Seed.writeError', [$path]));
+			return;
+		}
+
+		CLI::write('Created file: ' . CLI::color(str_replace($homepath, $ns, $path), 'green'));
```
paulbalandan, line 166: One last thing. I tried this with `Myth\Auth` as the namespace and I've got a rather nasty terminal output: `Myth\AuthDatabase/Seeds/TestSeeder.php`. I would suggest adding this line before line 166:

`$ns = rtrim(str_replace('\\', DIRECTORY_SEPARATOR, $ns), '\\') . DIRECTORY_SEPARATOR;`

BTW, thank you for doing my suggestions. 😁

michalsn:
Nice work! Thanks, @abilioposada, and @paulbalandan.

PR #3238, opened 2020-07-07
Add ability to call commands programatically.

lonnieezell:
- Refactored CommandRunner to bring all logic out to its own class.
- added new `command` common helper function to allow calling other commands programatically.
- Updated CLI class to be able to handle being ran when NOT on the cli. Will echo out the content instead.

### DIFF
--- a/system/CLI/BaseCommand.php
+++ b/system/CLI/BaseCommand.php
@@ -115,10 +115,10 @@ abstract class BaseCommand
 	/**
 	 * BaseCommand constructor.
 	 *
-	 * @param \Psr\Log\LoggerInterface       $logger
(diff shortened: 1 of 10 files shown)


lonnieezell:
@MGatner thanks for the review! Will definitely come in handy for cron stuff, I believe, which is part of why I built it. Also - was trying to write tests for a `cache:clear` command, and realized what a giant pain it was since the command runner was a controller. :)

lonnieezell:
Oh - and don't forget the already started [codeigniter4-tasks](https://github.com/lonnieezell/codeigniter4-tasks/) if you're going to be working on Tasks. :)

MGatner:
Oh believe me I haven’t forgotten! I’m just holding off as long as I can before unleashing on it 😊

PR #3241, opened 2020-07-07
Run apidocs action only when system files are changed

paulbalandan:
**Description**
Currently, `apidocs-action` runs on every push to `develop` regardless if there are changes to system files. So this PR adds a `paths` to the action's syntax. This is consistent with our `phpdoc.dist.xml` which transforms only files in the `system` directory.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


michalsn:
Really nice addition, thanks!

PR #3252, opened 2020-07-08
Implement model callback overrides

MGatner:
**Description**
This PR adds additional functionality to `Model` to allow/deny callbacks via a property (`$allowCallbacks`) or a per-execution method (`allowCallbacks(bool)`). This will help for testing, for extending models, or for unique situations where a developer needs a "safe mode" of model execution.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1681,7 +1681,7 @@ class Model
 	 *
 	 * @return Model
 	 */
-	public function allowCallbacks($val = true)
+	public function allowCallbacks(bool $val = true)
 	{
 		$this->tempAllowCallbacks = $val;
 
--- a/system/Model.php
+++ b/system/Model.php
@@ -267,45 +267,60 @@ class Model
 	 */
 
 	/**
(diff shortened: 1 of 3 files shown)


system/Model.php
```diff
@@ -1657,6 +1673,21 @@ public function countAllResults(bool $reset = true, bool $test = false)
...
 
+	/**
+	 * Sets $tempAllowCallbacks value so that we can temporarily override
+	 * the setting. Resets after the next trigger.
+	 *
+	 * @param boolean $val
+	 *
+	 * @return Model
+	 */
+	public function allowCallbacks($val = true)
```
michalsn, line 1684:
```suggestion
	public function allowCallbacks(bool $val = true)
```

MGatner:
This was a teeny change and has passed some tests already so I’m merging. Thanks @michalsn !

PR #3126, opened 2020-06-19
Prefer app translations

element-code:
This closes #3125

**Description**
Since the namespaces are sorted "most preferred first" and the locator searches and returns files in these order we can simply rotate the array to get the most preferred as overrides.

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

PR #3254, opened 2020-07-09
Fixes #3125 : add ability to override existing translation en in system language from App

### LINKED ISSUE #3125: Bug:  Can't override Translations Package Files / Keys
**Describe the bug**
When using the translations package, you can't override the validation messages, for example.

**CodeIgniter 4 version**
`4.0.3`

**Affected module(s)**
`CodeIgniter\Language`

require `codeigniter4/translations`, try override any key in `App/Translations/[lang]/Validation.php`.

samsonasik:
Alternative of #3126 to allow override existing translation validation.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Language/en/Validation.php
+++ b/app/Language/en/Validation.php
@@ -1,0 +1,4 @@
+<?php
+
+// override core en language system validation or define your own en language validation message
+return [];
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -244,15 +244,19 @@ class FileLocator
 			if (isset($namespace['path']) && is_file($namespace['path'] . $path))
 			{
 				$fullPath = $namespace['path'] . $path;
-				if ($prioritizeApp || strpos($fullPath, APPPATH) !== 0)
+				if ($prioritizeApp)
 				{
 					$foundPaths[] = $fullPath;
 				}
 				else
 				{
-					if (! in_array($fullPath, $appPaths, true) && strpos($fullPath, APPPATH) === 0)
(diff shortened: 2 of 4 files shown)


samsonasik:
I've added unit test for it.

samsonasik:
Previously, I change the `FileLocator::search()` autoloader. After some thinking, I think the change need to be in the `Language::requireFile()` to avoid BC break.

samsonasik:
I think there is a better way, add a 3rd argument in `FileLocator` autoloader, eg: `bool $prioritizeSystem = true` so it can be tested properly, and re-usable.

samsonasik:
I've added 3rd parameter `bool $prioritizeApp = true` in `Autoloader\FileLocator::search()` with set to false in `Language\Language::requireFile()`. Added unit test for it.

MGatner:
I like the $prioritizeApp option. Does this still weight the remaining files by their alphabetical order? This should follow the approved priority hierarchy, see https://github.com/codeigniter4/CodeIgniter4/pull/2507. But basically: App, Config, modules, System. E.g. developers should be able to include a third-party language pack and have it override system defaults automatically.

samsonasik:
Yes. And the 3rd parameter is default to true which uses existing functionality. Only in usage in `Language\Language::requireFile()` it uses false to make `App` language loaded in the last entry.

samsonasik:
travis build green :tada:

samsonasik:
updated to run `array_unique` once after array merged only.

samsonasik:
travis build green 🎉

PR #3257, opened 2020-07-09
\Config\Services::response()->deleteCookie Cannot run，because class R…

wangyupeng:
\Config\Services::response()->deleteCookie Cannot run，because class Response:$this->cookies Not initialized，so deleteCookie、getCookie、hasCookie Cannot run。

Solution:
delete cookie，also is set cookie timeout。


michalsn:
This clearly doesn't work. Please, create an issue with proper description to recreate the problem.

Also, please don't push not related commits to one PR.

PR #3273, opened 2020-07-11
Change getFieldData -> Add more readible field type names

devorama:

**Description**
Change getFieldData -> Add more readible field type names
Added more readible field type names as is the norm with the other dirvers (Postrges/sqllite)
to enable better understanding of what field type actually is.

Added Lenth property also as max_length is that of current data and not of what field can hold.
Will apply to others drivers also making it = max_length if driver dont support it.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/MySQLi/Result.php
+++ b/system/Database/MySQLi/Result.php
@@ -87,6 +87,41 @@ class Result extends BaseResult implements ResultInterface
 	 */
 	public function getFieldData(): array
 	{
+		static $data_types = [
+			MYSQLI_TYPE_DECIMAL		=> 'decimal',
+			MYSQLI_TYPE_NEWDECIMAL	=> 'newdecimal',
+			MYSQLI_TYPE_FLOAT		=> 'float',
+			MYSQLI_TYPE_DOUBLE		=> 'double',
+			
+			MYSQLI_TYPE_BIT			=> 'bit',
+			MYSQLI_TYPE_TINY		=> 'tiny',
+			MYSQLI_TYPE_SHORT		=> 'short',
+			MYSQLI_TYPE_LONG		=> 'long',
+			MYSQLI_TYPE_LONGLONG	=> 'longlong',
+			MYSQLI_TYPE_INT24		=> 'int24',
+			
+			MYSQLI_TYPE_YEAR		=> 'year',
+			
+			MYSQLI_TYPE_TIMESTAMP	=> 'timestamp',
+			MYSQLI_TYPE_DATE		=> 'date',
+			MYSQLI_TYPE_TIME		=> 'time',
+			MYSQLI_TYPE_DATETIME	=> 'datetime',
+			MYSQLI_TYPE_NEWDATE		=> 'newdate',
+			
+			MYSQLI_TYPE_INTERVAL	=> 'interval',
+			MYSQLI_TYPE_SET			=> 'set',
+			MYSQLI_TYPE_ENUM		=> 'enum',
+			
+			MYSQLI_TYPE_VAR_STRING	=> 'var_string',
+			MYSQLI_TYPE_STRING		=> 'string',
+			MYSQLI_TYPE_CHAR		=> 'char',
+			
+			MYSQLI_TYPE_GEOMETRY	=> 'geometry',
+			MYSQLI_TYPE_TINY_BLOB	=> 'tiny_blob',
+			MYSQLI_TYPE_MEDIUM_BLOB	=> 'medium_blob',
+			MYSQLI_TYPE_LONG_BLOB	=> 'long_blob',
+			MYSQLI_TYPE_BLOB		=> 'blob',
+		];
 		$retVal    = [];
 		$fieldData = $this->resultID->fetch_fields();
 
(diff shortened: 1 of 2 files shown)


system/Database/MySQLi/Result.php
```diff
@@ -87,16 +87,52 @@ public function getFieldNames(): array
...
+			MYSQLI_TYPE_VAR_STRING	=> 'var_string',
+			MYSQLI_TYPE_STRING		=> 'string',
+			MYSQLI_TYPE_CHAR		=> 'char',
+			
+			MYSQLI_TYPE_GEOMETRY	=> 'geometry',
+			MYSQLI_TYPE_TINY_BLOB	=> 'tiny_blob',
+			MYSQLI_TYPE_MEDIUM_BLOB	=> 'medium_blob',
+			MYSQLI_TYPE_LONG_BLOB	=> 'long_blob',
+			MYSQLI_TYPE_BLOB		=> 'blob',
+		];
```
samsonasik, line 124: It will make a maintenance burden in the future when a new type added/removed.

devorama: The same was done with SQL Lite driver, that why I followed the same logic. 
Cant think of a alternative for if the idea is the data returned by fielddata should be standard but blind to what database is in the background then something like a map will be needed as php dont seem to give alternative functions for it.

MGatner:
I like the change but agree that we can’t change the format of `type`. How about adding an additional field, something like `typeName`? It could be `null` if I matched but that way we still return the same value in `type`. 

I use `getFieldData()` in my database schema drafter, and although I don’t use `type` explicitly it is stored in the schema and could affect things downstream. https://github.com/tattersoftware/codeigniter4-schemas/blob/6ae885846f0ab08da4afadbbc178f6ee224e1575/src/Drafter/Handlers/DatabaseHandler.php#L115

michalsn:
Well, it's a shame that we discover such inconsistencies only after some time. 

I guess that an additional field is the only way to make this change happen.

system/Database/MySQLi/Result.php
```diff
@@ -87,16 +87,52 @@ public function getFieldNames(): array
...
 		$retVal    = [];
 		$fieldData = $this->resultID->fetch_fields();
 
 		foreach ($fieldData as $i => $data)
 		{
 			$retVal[$i]              = new \stdClass();
 			$retVal[$i]->name        = $data->name;
-			$retVal[$i]->type        = $data->type;
+			$retVal[$i]->type        = $data_types[$data->type];
```
michalsn, line 132: Please, leave the `type` as it is and add additional variable `typeName`.
```suggestion
			$retVal[$i]->type        = $data->type;
			$retVal[$i]->typeName    = $data_types[$data->type] ?? null;
```

system/Database/MySQLi/Result.php
devorama: I can change the line to have this logic.

$retVal[$i]->type        = isset($data_types[$data->type]) ? $data_types[$data->type] : $data->type;

then if their are future unmapped types, it wont break framework, but return the numeric value and one can then later on with update add the new type. 

New database field types is not added easily thus it wont be something that will need to be maintained for every single new mysql version I believe

system/Database/MySQLi/Result.php
```diff
@@ -87,16 +87,52 @@ public function getFieldNames(): array
...
+			
+			MYSQLI_TYPE_VAR_STRING	=> 'var_string',
+			MYSQLI_TYPE_STRING		=> 'string',
+			MYSQLI_TYPE_CHAR		=> 'char',
+			
+			MYSQLI_TYPE_GEOMETRY	=> 'geometry',
+			MYSQLI_TYPE_TINY_BLOB	=> 'tiny_blob',
+			MYSQLI_TYPE_MEDIUM_BLOB	=> 'medium_blob',
+			MYSQLI_TYPE_LONG_BLOB	=> 'long_blob',
+			MYSQLI_TYPE_BLOB		=> 'blob',
```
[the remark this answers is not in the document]
devorama: 100% will attend to this

system/Database/MySQLi/Result.php
devorama: Make sense, will look at

devorama:
@MGatner  & @michalsn  I have made the change   to add a field typeName
But what about SQLite3 driver ? Its getFieldData must it also be changed to follow the same standard where type represent the true value returned by the database driver and typeName the human readible version of it ?

devorama:
Sorry I merged a change on Formatter into this branch, Dunno how to undo that, still new to using github interface

MGatner:
> I have made the change to add a field typeName
> But what about SQLite3 driver ? Its getFieldData must it also be changed to follow the same standard where type represent the true value returned by the database driver and typeName the human readible version of it ?

@devorama I misunderstood the current state of things. Sorry for the extra work, but if MySQLi and SQLite3 are currently presenting different behavior then that is an issue and addressing it is not a breaking change. I will trust @michalsn to see this one through since he seems to understand it better than me, but I appreciate the work you have done.

Re: accidental commit... Probably easiest just to revert the changes you don't want: `git revert a32d3bf`. It will create a new commit with the reversion but it's easier & less risky than force pushes.

devorama:
> Re: accidental commit... Probably easiest just to revert the changes you don't want: `git revert a32d3bf`. It will create a new commit with the reversion but it's easier & less risky than force pushes.
@MGatner  Thanks will try to revert feel l so stupid doing it all via browser, cant use the github desktop. thanks for help

devorama:
@MGatner  and @michalsn and all other reviewers.

I noticed now that I did these changes to my **development** branch in my fork. Please **reject** this change request I will create a new branch and make all the changes in that branch,very sorry for all this I am still learning how the web interface works, but think I have figured it out now.

michalsn:
@devorama I will close all your PR since they seem a bit "messy". The code addressing different issues is mixed. You should try GitHub Desktop or SourceTree - it probably will be much easier for you to work with it.

PR #3280, opened 2020-07-12
optimize FileLocator autoloader : use foreach instead of for with count when possible

samsonasik:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -171,34 +171,38 @@ class FileLocator
 	{
 		$php        = file_get_contents($file);
 		$tokens     = token_get_all($php);
-		$count      = count($tokens);
 		$dlm        = false;
 		$namespace  = '';
 		$class_name = '';
 
-		for ($i = 2; $i < $count; $i++)
+		foreach ($tokens as $i => $token)
 		{
-			if ((isset($tokens[$i - 2][1]) && ($tokens[$i - 2][1] === 'phpnamespace' || $tokens[$i - 2][1] === 'namespace')) || ($dlm && $tokens[$i - 1][0] === T_NS_SEPARATOR && $tokens[$i][0] === T_STRING))
+			if ($i < 2)
+			{
+				continue;
+			}
+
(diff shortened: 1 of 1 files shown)


michalsn:
Sorry for the silly question but is `foreach` faster than `for`? Or this is just for readability purposes?

samsonasik:
previously, with `for`, it need to run `count()` first, while direct foreach is enough.

michalsn:
Yes, I understand this part, but isn't `foreach` slower than `for`?

samsonasik:
It faster. Example benchmark about `for` with `count` vs direct `foreach` : https://coderwall.com/p/il1tog/php-for-vs-foreach-benchmark

samsonasik:
```
➜  ~ curl https://gist.githubusercontent.com/nodesocket/3909074/raw/80e1bd3612b2b1680d6d519c8fd53bb79435e9de/gistfile1.php -o benchmark-for-vs-foreach.php

➜  ~ php benchmark-for-vs-foreach.php 
For took: 2.862ms
For with count() took: 3.643ms
Foreach took: 1.489ms
```

michalsn:
I get quite different results to be honest
```
For took: 0.128ms
For with count() took: 0.175ms
Foreach took: 0.162ms
```
Also checked it on sandbox: https://phpsandbox.io/n/red-king-pnbr

samsonasik:
The code that I updated previosly uses count+for, so foreach still win

michalsn:
Not really. "For with count()" is when we use `count()` inside `for()` loop, like:
```php
for($i = 0; $i < count($elements); $i++)
```
and it's not the case here.

samsonasik:
The benchmark code need update to associative array and run the code inside loop, I updated the benchmark code at https://phpsandbox.io/n/super-fog-72g7 with place assign `count()` to variable first, but execute it. I got it:

```bash
For took: 0.066ms
For with count() took: 0.167ms
Foreach took: 0.099ms
```

foreach wins on this case.

michalsn:
Okay, now it looks good. Thanks.

PR #3282, opened 2020-07-12
optimize Database BaseBuilder : use foreach instead of for with count when possible

samsonasik:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -685,14 +685,12 @@ class BaseBuilder
 			}
 
 			$cond = ' ON ';
-			$i    = 0;
-			foreach ($conditions as $condition)
+			ksort($conditions);
+			foreach ($conditions as $i => $condition)
 			{
-				$operator = $this->getOperator($conditions[$i]);
+				$operator = $this->getOperator($condition);
 				$cond    .= $joints[$i];
-				$cond    .= preg_match("/(\(*)?([\[\]\w\.'-]+)" . preg_quote($operator) . '(.*)/i', $conditions[$i], $match) ? $match[1] . $this->db->protectIdentifiers($match[2]) . $operator . $this->db->protectIdentifiers($match[3]) : $conditions[$i];
-
-				$i++;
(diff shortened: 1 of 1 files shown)


system/Database/BaseBuilder.php
```diff
@@ -685,11 +685,14 @@ public function join(string $table, string $cond, string $type = '', bool $escap
 			}
 
 			$cond = ' ON ';
-			for ($i = 0, $c = count($conditions); $i < $c; $i ++)
+			$i    = 0;
```
michalsn, line 688: Maybe it's something obvious (but I can't see it). Why do we still need a counter here?

samsonasik: I don't quite understand the logic, but without numeric increment from 0, it got error. On my first commit, I set it without loop numeric key ( https://github.com/codeigniter4/CodeIgniter4/pull/3282/commits/0c61267b0d749d15d21e7392e834d46a313dd7d7 ), and got error:

```bash
1) Builder\JoinTest::testJoinMultipleConditions
--- Expected
+++ Actual
@@ @@
-'SELECT * FROM "table1" LEFT JOIN "table2" ON "table1"."field1" = "table2"."field2" AND "table1"."field1" = 'foo' AND "table2"."field2" = 0'
+'SELECT * FROM "table1" LEFT JOIN "table2" ON  AND "table2"."field2" = 0 AND "table1"."field1" = 'foo'"table1"."field1" = "table2"."field2"'

/Users/samsonasik/www/CodeIgniter4/tests/system/Database/Builder/JoinTest.php:69
```

Make increment with `$i` fix that.

michalsn: OK, now I see it. You can use `ksort($conditions)` before `foreach` loop. That should fix it.

samsonasik: ok, updated. Thank you 👍

michalsn: Oh... you know what? I'm really sorry but can you place `ksort` after line 679? Because if there are no multiple conditions we don't really need this.

samsonasik: I've updated to place `ksort` after line 679, thank you ;)

michalsn:
Thank you @samsonasik

PR #3286, opened 2020-07-12
Add JSON_NUMERIC_CHECK to JSON encoding

devorama:
This allows encoding to make numeric data more true to its type (not enclose them with "", making the receiving system identify
the types easier.

Note making the options variable configurable might be handy in the future depending on the needs of the users
people they interface with they might need additional options added here

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

PR #3287, opened 2020-07-12
FieldData -> add typeName field and length field

devorama:
Pull request was done wrong previously where I did a merge into my development branch and send pull request from there.

https://github.com/codeigniter4/CodeIgniter4/pull/3273

This is to correct that mistake

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/MySQLi/Result.php
+++ b/system/Database/MySQLi/Result.php
@@ -87,6 +87,42 @@ class Result extends BaseResult implements ResultInterface
 	 */
 	public function getFieldData(): array
 	{
+		static $data_types = [
+			MYSQLI_TYPE_DECIMAL     => 'decimal',
+			MYSQLI_TYPE_NEWDECIMAL  => 'newdecimal',
+			MYSQLI_TYPE_FLOAT       => 'float',
+			MYSQLI_TYPE_DOUBLE      => 'double',
+			
+			MYSQLI_TYPE_BIT         => 'bit',
+			MYSQLI_TYPE_TINY        => 'tiny',
+			MYSQLI_TYPE_SHORT       => 'short',
+			MYSQLI_TYPE_LONG        => 'long',
+			MYSQLI_TYPE_LONGLONG    => 'longlong',
+			MYSQLI_TYPE_INT24       => 'int24',
+			
+			MYSQLI_TYPE_YEAR        => 'year',
+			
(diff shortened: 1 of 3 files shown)


michalsn:
Can you make changes to other drivers in this PR too?

system/Database/MySQLi/Result.php
```diff
@@ -95,11 +131,12 @@ public function getFieldData(): array
 			$retVal[$i]              = new \stdClass();
 			$retVal[$i]->name        = $data->name;
 			$retVal[$i]->type        = $data->type;
+			$retVal[$i]->typeName    = isset($data_types[$data->type]) ? $data_types[$data->type] : $data->type;
```
michalsn, line 134: This should be `null` if key is missing.

devorama: Will change to be null then, but to be honest I believe the developer might more quickly find the reason for it returning a number instead of the text if he gets a number as n reply and realize it is not mapped. But will make the change as requested.

devorama:
> Can you make changes to other drivers in this PR too?

@michalsn will do so yes. Will also follow same concept to return null if not found

devorama:
Changes made to other drivers are requested. 
Note I notice not all set the expected values for example 
Postgres and SQLlite dont set primary key and default. 
I know it 10/1 not big concern, but coding wise will it not be wise to now put in these standard
expected fields and set them to null then if the drivers do allow these type of data to be retrieved in the future one 
wont miss it as when reviewing the code one would see it is a expected result ?

michalsn:
No, it's okay. We mention in the user guide that not all metadata are available for every driver. Also, these types of things don't change often so I think we're good.

devorama:
@michalsn  Feel like I am waisting your guys time, but I cannot get my git repo fixed for  
https://github.com/codeigniter4/CodeIgniter4/pull/3286
But I dunno if this pull request wont work. If it does not then let me know and I will redo this and we can just close this one

michalsn:
We need one more change. Variable `typeName` should be renamed to `type_name` since we use snake case for all other variables.

devorama:
@michalsn 
Good day, I see my re-fork has broken thus pull request, I will redo it in the new fork
hopefully from now on it will be fixed. 
Will add all the changes as above in it as requested.

PR #3288, opened 2020-07-12
add JSON_NUMERIC_CHECK to json encode options

devorama:
linked to 
https://github.com/codeigniter4/CodeIgniter4/pull/3286

This allows encoding to make numeric data more true to its type (not enclose them with "", making the receiving system identify
the types easier.

Note making the options variable configurable might be handy in the future depending on the needs of the users
people they interface with they might need additional options added here

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/app/Config/Format.php
+++ b/app/Config/Format.php
@@ -52,7 +52,7 @@ class Format extends BaseConfig
 	|
 	*/
 	public $formatterOptions  = [
-		'application/json' => JSON_NUMERIC_CHECK,
+		'application/json' => null,
 		'application/xml'  => null,
 		'text/xml'         => null,
 	];	
--- a/app/Config/Format.php
+++ b/app/Config/Format.php
@@ -52,9 +52,9 @@ class Format extends BaseConfig
 	|
 	*/
 	public $formatterOptions  = [
-		'application/json' => JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES ,
-		'application/xml'  => null,
-		'text/xml'         => null,
+		'application/json' => JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES,
+		'application/xml'  => 0,
+		'text/xml'         => 0,
 	];	
 	//--------------------------------------------------------------------
 
--- a/app/Config/Format.php
+++ b/app/Config/Format.php
@@ -52,7 +52,7 @@ class Format extends BaseConfig
 	|
 	*/
 	public $formatterOptions  = [
-		'application/json' => null,
+		'application/json' => JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES ,
 		'application/xml'  => null,
 		'text/xml'         => null,
 	];	
--- a/app/Config/Format.php
+++ b/app/Config/Format.php
@@ -47,8 +47,6 @@ class Format extends BaseConfig
 	|
 	| Additional Options to adjust default formatters behaviour.
 	| For each mime type, list the additional options that should be used. 
-	| Set to null by default
-	| Note this will  add to the default options not replace it.
 	|
 	*/
 	public $formatterOptions  = [
--- a/app/Config/Format.php
+++ b/app/Config/Format.php
@@ -39,7 +39,23 @@ class Format extends BaseConfig
 		'application/xml'  => \CodeIgniter\Format\XMLFormatter::class,
 		'text/xml'         => \CodeIgniter\Format\XMLFormatter::class,
 	];
-
+	
+	/*
+	|--------------------------------------------------------------------------
+	| Formatters Options
+	|--------------------------------------------------------------------------
+	|
+	| Additional Options to adjust default formatters behaviour.
+	| For each mime type, list the additional options that should be used. 
+	| Set to null by default
+	| Note this will  add to the default options not replace it.
+	|
+	*/
+	public $formatterOptions  = [
+		'application/json' => JSON_NUMERIC_CHECK,
+		'application/xml'  => null,
+		'text/xml'         => null,
+	];	
 	//--------------------------------------------------------------------
 
 	/**
--- a/system/Format/JSONFormatter.php
+++ b/system/Format/JSONFormatter.php
@@ -57,7 +57,7 @@ class JSONFormatter implements FormatterInterface
 	public function format($data)
 	{
 		$options = $config->formatterOptions['application/json'] ?? JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES;
-		$options = $options | JSON_PARTIAL_OUTPUT_ON_ERROR ;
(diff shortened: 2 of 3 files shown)


michalsn:
Adding a numeric check just like that may potentially break someone's code. Developers may already use code like: 
```php
if ($foo->bar === '1')
``` 

But I like your idea of making this configurable. Would you like to work on this PR and add something like `$formatterOptions` array to the `Config/Format.php` file? We would use it like:
```php
$config = new Filter();
$options = $config->formatterOptions['application/json'] ?? JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_PARTIAL_OUTPUT_ON_ERROR
```
I'm not really sure about naming the keys for this new variable so if anybody has a different idea, I'm all ears.

devorama:
Hi there @michalsn  , I will do this asap. Will look at the current names being used and see if I can add something that appropiate.

> But I like your idea of making this configurable. Would you like to work on this PR and add something like `$formatterOptions` array to the `Config/Format.php` file? We would use it like:

devorama:
@michalsn  I added the config value and also the code for XML and JSON Formatter, for XML I added the additional option to be added as per the php docs or default to the phpdocs default.

One thing to note is I did not make it to overwrite the default options but to add to it, the reason for this is if you look at
JSONFormatter the JSON_PARTIAL_OUTPUT_ON_ERROR is actually important for error checking afterwards in the same function, thus allowing a total override of default options might break actual logic for that reason I believed it saver to make it a add to current option only. The default for the settings are all null.

michalsn:
The whole sense of making it configurable is that we should have the ability to change default flags. So please add all current flags to the config file as default.

You can check later in the code if `JSON_PARTIAL_OUTPUT_ON_ERROR` is set via bitwise operator and add it if needed.
```php
if (! ($options & JSON_PARTIAL_OUTPUT_ON_ERROR))
// add it to options
```

michalsn:
Actually I don't think that we have to check it at all (options) - we can just add it and it should work fine (even if it's already present).

johpromed:
Hi there. The logic I have is as follow, basically it will do what you suggest in your last post. But based on wanting to give control to the user to select the encoding options (basically if they want JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES )
I will move all but JSON_PARTIAL_OUTPUT_ON_ERROR  to be in the Config/Format.php , then  the result will be achieved what you are aiming for.

```
		$options = $config->formatterOptions['application/json'] ?? 0;
		$options = $options | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_PARTIAL_OUTPUT_ON_ERROR ;

```
> Actually I don't think that we have to check it at all (options) - we can just add it and it should work fine (even if it's already present).

michalsn:
Just update your code. If I will have any requests for changes I will let you know :)

devorama:
Oops used my company account there :), will do the chagne now, just finished testing it and it seems to work :)

> Just update your code. If I will have any requests for changes I will let you know :)

devorama:
@michalsn  K changes made, should work now.

app/Config/Format.php
```diff
@@ -39,7 +39,23 @@ class Format extends BaseConfig
...
+	|--------------------------------------------------------------------------
+	|
+	| Additional Options to adjust default formatters behaviour.
+	| For each mime type, list the additional options that should be used. 
+	| Set to null by default
+	| Note this will  add to the default options not replace it.
+	|
+	*/
+	public $formatterOptions  = [
+		'application/json' => JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES ,
```
michalsn, line 55:
```suggestion
		'application/json' => JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES,
```

system/Format/JSONFormatter.php
```diff
@@ -56,7 +56,8 @@ class JSONFormatter implements FormatterInterface
 	 */
 	public function format($data)
 	{
-		$options = JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_PARTIAL_OUTPUT_ON_ERROR;
+		$options = $config->formatterOptions['application/json'] ?? 0;
```
michalsn, line 59:
```suggestion
		$options = $config->formatterOptions['application/json'] ?? JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES;
```
This way we can introduce this change with new configuration variable really easy.

app/Config/Format.php
```diff
@@ -39,7 +39,23 @@ class Format extends BaseConfig
...
+		'application/xml'  => null,
+		'text/xml'         => null,
```
michalsn, line 57: Please make it default to `0`.

system/Format/JSONFormatter.php
```diff
@@ -56,7 +56,8 @@ class JSONFormatter implements FormatterInterface
...
+		$options = $options | JSON_PARTIAL_OUTPUT_ON_ERROR ;
```
michalsn, line 60:
```suggestion
		$options = $options | JSON_PARTIAL_OUTPUT_ON_ERROR;
```

devorama:
@michalsn changes made,will try and enforce the coding standard more with next changes.

system/Format/XMLFormatter.php
```diff
@@ -66,7 +66,8 @@ public function format($data)
 			// @codeCoverageIgnoreEnd
 		}
 
-		$output = new \SimpleXMLElement('<?xml version="1.0"?><response></response>');
+		$options = $config->formatterOptions['application/xml'] ?? 0;
+		$output = new \SimpleXMLElement('<?xml version="1.0"?><response></response>',$options);
```
michalsn, line 70:
```suggestion
		$output = new \SimpleXMLElement('<?xml version="1.0"?><response></response>', $options);
```
Sorry, I missed this one.

michalsn:
@devorama No worries. Just one last change and I think we're good.

devorama:
@michalsn  all done :)

app/Config/Format.php
```diff
@@ -39,7 +39,23 @@ class Format extends BaseConfig
...
+	/*
+	|--------------------------------------------------------------------------
+	| Formatters Options
+	|--------------------------------------------------------------------------
+	|
+	| Additional Options to adjust default formatters behaviour.
+	| For each mime type, list the additional options that should be used. 
+	| Set to null by default
+	| Note this will  add to the default options not replace it.
+	|
```
paulbalandan, line 52: Please change the comment. The $options are no longer additions but rather the defaults already.

devorama:
@michalsn  and @paulbalandan  Made the changes as requested , sorry for the config slip up.

PR #3298, opened 2020-07-13
reduce repetitive getDefaultNamespace() and controllerName() function call in Router

samsonasik:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -548,11 +548,13 @@ class Router implements RouterInterface
 			$this->params = $segments;
 		}
 
+		$defaultNameSpace = $this->collection->getDefaultNamespace();
+		$controllerName   = $this->controllerName();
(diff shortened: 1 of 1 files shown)


system/Router/Router.php
```diff
@@ -548,11 +548,13 @@ public function autoRoute(string $uri)
 			$this->params = $segments;
 		}
 
+		$defaultNameSpace = $this->collection->getDefaultNamespace();
```
michalsn, line 551: I know this is a bit picky, but can we make it `defaultNamespace`?

MGatner: I, also, think this is picky and would like to see it changed 😅

samsonasik: I've updated to `defaultNamespace` 👍

michalsn:
Thank you!

PR #3299, opened 2020-07-13
FieldData -> add typeName field and length field 

devorama:
As per pull request
https://github.com/codeigniter4/CodeIgniter4/pull/3287
added field type name.
had to redo it as I had to re-fork the repo and that broke the previous pull request.

Will return NULL if mapped field type name not found

the DB version of the type returned by driver is stored in the type column

Where applicable length is stored or made equal to max_length as in for example MySQL the max_length is based on
the data returned and not the DB schema where length is closer to the truth but still not 100%
cant think of a alternative that wont be affect the performance of the procedure. Is a need if want to build schema
from this data in the future

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/MySQLi/Result.php
+++ b/system/Database/MySQLi/Result.php
@@ -87,6 +87,42 @@ class Result extends BaseResult implements ResultInterface
 	 */
 	public function getFieldData(): array
 	{
+		static $data_types = [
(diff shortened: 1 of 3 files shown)


michalsn:
Yay! We made it finally 🥳

Thank you @devorama!

PR #3302, opened 2020-07-13
force_https didn’t force https

colethorsen:
when https:// is present in a defined base URL and a user accesses the site on on an insecure protocol, it would not force them onto secure.

**Description**
just because the baseURL defines the protocol as https the user isn't necessarily forced onto https. This forces the user onto https.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -451,7 +451,7 @@ if (! function_exists('force_https'))
 		{
 			$baseURL = (string) substr($baseURL, strlen('https://'));
 		}
-		else if (strpos($baseURL, 'http://') === 0)
+		elseif (strpos($baseURL, 'http://') === 0)
 		{
 			$baseURL = (string) substr($baseURL, strlen('http://'));
 		}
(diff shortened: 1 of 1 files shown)


system/Common.php
```diff
@@ -447,13 +447,11 @@ function force_https(int $duration = 31536000, RequestInterface $request = null,
...
 
-		// If we already use 'https' then return immediately
 		if (strpos($baseURL, 'https://') === 0)
 		{
-			return;
+			$baseURL = (string) substr($baseURL, strlen('https://'));
 		}
-
-		if (strpos($baseURL, 'http://') === 0)
+		else if (strpos($baseURL, 'http://') === 0)
```
michalsn, line 454: One little change - just to be compliant with our code standard:
```suggestion
		elseif (strpos($baseURL, 'http://') === 0)
```

michalsn:
Thank you!

Issue #3297, opened 2020-07-12
Bug: Setting Encryption Service Key in app/Config/Encyption.php


durbintl:
**Describe the bug**
Using `hex2bin` within app/Config/Encryption.php throws the error **Constant expression contains invalid operations**. Currently a `__constructor` function must be used to set the key using the method as stated in the documentation at [Encryption Service](https://codeigniter.com/user_guide/libraries/encryption.html#encoding-keys-or-results)

**Example**
`public function __contstruct(){
    parent::__construct();
    $this->key = hex2bin('example');
}`

It may also be worth mentioning that setting your encryption key in a version tracked file is not a best security practice. It would be far more secure to set it in .env file as a Hex string and convert to bin on load, since the .env is not version tracked.

**CodeIgniter 4 version**
4.0.3

**Affected module(s)**
Encryption Service

**Expected behavior, and steps to reproduce if appropriate**
See description above.

**Context**
 - OS: macOS Catalina 10.15.5
 - Web server: Apache 2.2.34
 - PHP version: 7.4.2

paulbalandan:
`Config\Encryption` extends `BaseConfig`. So technically we can add the encryption key in a gitignored `.env` file. It will be parsed on run time. You just need to add an entry in your env file.

durbintl:
@paulbalandan True, however it is not a good idea to store a binary string in a file along with plain text. Opening the file with an editor will corrupt the binary on save. With the changes that @michalsn has made, it will allow the use of a hex string with a `hex2bin:` modifier which will be decrypted by CI at runtime. I have already implemented the changes and it's running perfect. @michalsn thanks for all your hard work. The CI community would be lost without you!!

michalsn:
@durbintl Thanks for the kind words, but I'm not the only person who contributes to this project

PR #3307, opened 2020-07-14
Add hex2bin prefix handling for encryption key

michalsn:
**Description**
This PR adds support for 'hex2bin' prefix for `$key` variable from Encryption class. Both in PHP config and .env file.
See: #3297

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


system/Config/BaseConfig.php
```diff
@@ -92,6 +92,12 @@ public function __construct()
 		foreach ($properties as $property)
 		{
 			$this->initEnvValue($this->$property, $property, $prefix, $shortPrefix);
+
+			// Handle hex2bin prefix
+			if ($shortPrefix === 'encryption' && $property === 'key' && substr($this->$property, 0, 8) === 'hex2bin:')
```
samsonasik, line 97: can use strpos instead of substr: 

```suggestion
                      if ($shortPrefix === 'encryption' && $property === 'key' && strpos($this->$property, 'hex2bin:') === 0)
```

system/Config/DotEnv.php
```diff
@@ -182,6 +182,12 @@ public function normaliseVariable(string $name, string $value = ''): array
 
 		$value = $this->resolveNestedVariables($value);
 
+		// Handle hex2bin prefix
+		if ($name === 'encryption.key' && substr($value, 0, 8) === 'hex2bin:')
```
samsonasik, line 186: can use strpos instead of substr: 

```suggestion
              if ($name === 'encryption.key' && strpos($value, 'hex2bin:') === 0)
```

PR #3311, opened 2020-07-15
Make Redirect trailing slashes redirect with base

Moutard3:
**Description**
When using /$1 it doesn't prepend RewriteBase to the uri, so going to /sub/controller/ will redirect to /controller instead of /sub/controller

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -18,7 +18,8 @@ Options All -Indexes
 
 	# Redirect Trailing Slashes...
 	RewriteCond %{REQUEST_FILENAME} !-d
-	RewriteRule ^(.*)/$ $1 [L,R=301]
+	RewriteCond %{REQUEST_URI} (.+)/$
+      RewriteRule ^ %1 [L,R=301]
 
 	# Rewrite "www.example.com -> example.com"
 	RewriteCond %{HTTPS} !=on
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -18,7 +18,7 @@ Options All -Indexes
 
 	# Redirect Trailing Slashes...
 	RewriteCond %{REQUEST_FILENAME} !-d
-	RewriteRule ^(.*)/$ /$1 [L,R=301]
+	RewriteRule ^(.*)/$ $1 [L,R=301]
 
 	# Rewrite "www.example.com -> example.com"
 	RewriteCond %{HTTPS} !=on


MGatner:
I’m really not very good with `.htaccess` rules - maybe somebody else could review this?

MGatner:
@paulbalandan you took the last pass at **.htaccess** - interested?

michalsn:
I believe that on Windows this changed rule may not work as expected. I faced a similar problem a few years ago. Maybe something like this? 
```
RewriteCond %{REQUEST_FILENAME} !-d
RewriteCond %{REQUEST_URI} (.+)/$
RewriteRule ^ %1 [L,R=301]
```

It's from my old project and it worked fine.

michalsn:
Anyway...  I wonder shouldn't we comment out these lines responsible for redirecting trailing slashes by default? They already seem to lead to some confusion #3330. The only benefits from these rules are related to the SEO I guess. Which shouldn't be something we force on people. 

You know... my thinking is... you test your code locally via `spark serve` and everything is great but then you upload your project to the server and suddenly "things are broken".

Any thoughts?

paulbalandan:
Forgive me for my ignorance. But is `sub` there referring to a subfolder in the document root, or a subfolder for the controller?

paulbalandan:
@Moutard3 , have you tried uncommenting line 17 in `.htaccess`?

`# RewriteBase /`

michalsn:
@paulbalandan From what I understand `RewriteBase` is already uncommented and set to `/sub`

Moutard3:
I already set the rewritebase to /sub (tried /sub/ too). I had this problem on windows, and changing this line fixed it.
If you'd rather delete it, it would be okay too.

michalsn:
@Moutard3 Yes, the trailing slash redirection rule we currently have does not work with `RewriteBase`.

Maybe @paulbalandan will find some time later and will be able to check the proposed changes on Windows?

paulbalandan:
@Moutard3 I'm having problems installing CI4 in a subfolder. Can you tell me how did you setup yours so that I can check your proposed changes? The installation is taking me so much hours now. Thanks.

michalsn:
Okay, I verified that the proposed change doesn't work when we're using a "clear" domain name (without RewriteBase).

public/.htaccess
```diff
@@ -18,7 +18,7 @@ Options All -Indexes
 
 	# Redirect Trailing Slashes...
 	RewriteCond %{REQUEST_FILENAME} !-d
-	RewriteRule ^(.*)/$ /$1 [L,R=301]
+	RewriteRule ^(.*)/$ $1 [L,R=301]
```
michalsn, line 21:
```suggestion
	RewriteCond %{REQUEST_URI} (.+)/$
      RewriteRule ^ %1 [L,R=301]
```

Moutard3: Yeah it's working with your change, I'm commiting this

Moutard3:
I checked your suggestion, and it's still working for me

PR #3312, opened 2020-07-15
[ci skip] documentation : response->download() require getBody() to be called after send() to force download

samsonasik:
Tried in controller, without `getBody()` call after send, it will only shows the file content, not force download.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated

**Without `getBody()` call after `send()`:**

Screen Shot 2020-07-15 at 5 29 11 PM

**With `getBody()` call after `send()`:**

Screen Shot 2020-07-15 at 5 26 36 PM


samsonasik:
I create alternative PR for it at #3313 which update the code.

PR #3316, opened 2020-07-15
Pass filter arguments to after() and before() methods

tangix:
**Description**
Fixing #3216 by changing the FilterInterface::after() and before() to contain the arguments from the routing setup.
Updating core classes implementing FilterInterface, tests and docs.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Filters/CSRF.php
+++ b/system/Filters/CSRF.php
@@ -57,6 +57,7 @@ use Config\Services;
  */
 class CSRF implements FilterInterface
 {
+
 	/**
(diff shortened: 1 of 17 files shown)


tangix:
Note that code implementing FilterInterface needs to be updated!

lonnieezell:
Could you add a note to update the implementations on the Changelog? https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/changelogs/v4.0.4.rst

Issue #3303, opened 2020-07-13
Bug: URI::setQuery replaces dots in URL parameters with underscores


JanZelenka:
When performing Curl calls, the method mentioned in the title uses parse_str to make an array of parameters.
As per the PHP documentation, parse_str replaces characters illegal in variable names with underscore.
The outgoing query parser must be able to work with the request without modifying it.

*CodeIgniter 4 version:* 4.0.3
*Affected module(s)*: Those using the URI class
*Expected behavior*: Performing a CURL call shouldn't end up sending a different request than the one specified by the calling code.
*Context*: MacOS MAMP 4.5 with PHP 7.2.1

devorama:
If I might ask, what type of query variables are you trying to send  for the document state
"Because variables in PHP can't have dots and spaces in their names, those are converted to underscores"
This can I assume that this will be good values to test with  ?

`?test me=4&test.another=23`

JanZelenka:
Yes, I believe that would be a good test. My parameter is this: ?script.param=123

devorama:
Hi there. 
On a subnote you look for those parameters in a php script via the $_GET,$_POST or $_REQUEST they will also 
be converted as with parse_str. Thus it maybe not best practice, but will give it a look to see if there is a solution if somebody else 
dont come up with a solution beforehand

https://www.php.net/manual/en/language.variables.external.php

devorama:
This just for information sake. Tested with the following
```
 		  $uri = new URI();  
		  $arr['script.param'] = 1234;
		  $arr['another test'] = 'hallo test';
		  $arr['another[bracket]'] = 'hallo bracket';
          $arr['another-minus'] = 'hallo minus';		  
		  $uri = $uri->setQueryArray($arr);
		  
		  echo 'setQueryArray  : '.$uri->getQuery() . '</br>';
		  
		  $uri = new URI();  
		  $uri = $uri->setQuery('?script.param=1234&another test=hallo test&another[bracket]=hallo bracket&another-minus=hallo minus');
		  echo 'setQuery  : '.$uri->getQuery() . '</br>';
```
and the output is the following
```
setQueryArray : script_param=1234&another_test=hallo+test&another%5Bbracket%5D=hallo+bracket&another-minus=hallo+minus
setQuery : script_param=1234&another_test=hallo+test&another%5Bbracket%5D=hallo+bracket&another-minus=hallo+minus
```

devorama:
@michalsn 
[...]

PR #3326, opened 2020-07-16
Add support for using raw query string

michalsn:
**Description**
This PR adds the ability to use raw query string data (key names with dots are allowed) when a special flag `URI::useRawQueryString()` is set. 

By default `URI::setQuery()` and `URI::setQueryArray()` methods uses native `parse_str()` function to prepare data which may cause problems when using "non standard" key names when working with `CURLRequest` class.

From now on `CURLRequest` class will always use `URI::useRawQueryString()` as default when working with a query string.

See: #3303

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -130,7 +130,7 @@ class CURLRequest extends Request
 		parent::__construct($config);
 
 		$this->response = $response;
-		$this->baseURI  = $uri;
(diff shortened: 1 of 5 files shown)


MGatner (approved):
Your regex is way too involved for me but the tests look good so you have my 👍

PR #3329, opened 2020-07-16
Bug: CURLRequest lib doesn't set correct headers with digest auth

ert-vivien:
**Description**
CURLRequest library makes a curl request and parse the result to set headers and body content to the correct objects. This parsing checks the first double carriage return and put the first part of the curl result as headers and the rest as body.
Problem is when using digest auth we get 2 headers in the curl response causing response headers being falsy and body containing a header.
This pull request handle this case by skipping the first headers generated by the first digest http call and start splitting for body and headers at the second double carriage return when having a successful digest auth.

Forum threads here : https://forum.codeigniter.com/thread-77046.html

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -459,7 +459,7 @@ class CURLRequest extends Request
 		 // If request and response have Digest
 		if (isset($this->config['auth'][2]) && $this->config['auth'][2] === 'digest' && strpos($output, 'WWW-Authenticate: Digest') !== false)
 		{
-				$output = substr($output, strpos($output, $breakString) + 5);
+				$output = substr($output, strpos($output, $breakString) + 4);
 		}
 
 
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -471,7 +471,8 @@ class CURLRequest extends Request
 		{	
 			$break = strpos($output, $break_string, $first_carriage_return + strlen($break_string));
 
-		} else 
+		} 
+		else 
 		{
 			$break = strpos($output, "\r\n\r\n");
 
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -447,37 +447,24 @@ class CURLRequest extends Request
 		}
 
 		$output = $this->sendRequest($curl_options);
+		
+		// Set the string we want to break our response from
+		$breakString = "\r\n\r\n";
 
 		if (strpos($output, 'HTTP/1.1 100 Continue') === 0)
 		{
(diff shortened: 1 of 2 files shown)


system/HTTP/CURLRequest.php
```diff
@@ -452,9 +452,31 @@ public function send(string $method, string $url)
 		{
 			$output = substr($output, strpos($output, "\r\n\r\n") + 4);
 		}
+		
+		// Set the string we want to break our response from
+		$break_string = "\r\n\r\n";
```
samsonasik, line 457: variable need to be camel case for consistency

michalsn, line 457: Please, change it to `$breakString = "\r\n\r\n";` then place it right after line 449 and replace all occurrences of `"\r\n\r\n"` with this variable.

system/HTTP/CURLRequest.php
```diff
@@ -452,9 +452,31 @@ public function send(string $method, string $url)
...
+			$break = strpos($output, $break_string, $first_carriage_return + strlen($break_string));
 
-		// Split out our headers and body
-		$break = strpos($output, "\r\n\r\n");
+		} 
+		else 
+		{
+			$break = strpos($output, "\r\n\r\n");
+
+		}
```
michalsn, line 479: Please simplify the above code to this:
```php
                // If request and response have Digest
		if (isset($this->config['auth'][2]) && $this->config['auth'][2] === 'digest' && strpos($output, 'WWW-Authenticate: Digest') !== false)
		{
				$output = substr($output, strpos($output, $breakString) + 4);
		}

		// Split out our headers and body
		$break = strpos($output, $breakString);
```

ert-vivien:
How can one test a digest auth response? What endpoint should we hit? Or would you like to only have the current tests running clean?

michalsn:
We already have tests that have to be updated. Please update `CURLRequestTest::testAuthDigestOption()` and `CURLRequestTest::testSetAuthDigest()` test. You can take a look at `CURLRequestTest::testSendContinuedWithManyHeaders()` as an example on how to test output headers.

In general, we have to prepare an output string first and then apply it to the request:

```php
$output = "headers go here";
$this->request->setOutput($output)->request('get', 'http://example.com', [
...
```
And then we check `$response->getBody()`, header keys and maybe response code. That's it. It will be really similar to the `testSendContinuedWithManyHeaders()` test.

ert-vivien:
Thanks for the nice feedbacks. As a new developper in the open source world I learned a few things about unit tests in PHP :)

michalsn:
Looking good. Thank you!

PR #3331, opened 2020-07-16
Add 'base64:' prefix handling in encryption key

paulbalandan:
**Description**
This PR adds the `base64:` prefix for use in encryption keys, alongside `hex2bin:` implemented in #3307 .

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


paulbalandan:
Help please for the DotEnvTest. I don't know where went wrong.

paulbalandan:
I got this weird behavior of `getenv` on Windows where `getenv()` does not give the key-value pair of the binary string, but `$_ENV` does. This adds the `$_ENV` check first before defaulting to the `getenv` call. Tests pass locally on Windows.

paulbalandan:
And on Travis!! Yey

tests/system/Config/DotEnvTest.php
```diff
@@ -71,6 +72,20 @@ public function testLoadsHex2Bin()
 
 	//--------------------------------------------------------------------
 
+	public function testLoadsBase64()
+	{
+		$dotenv = new DotEnv($this->fixturesFolder, 'base64encryption.env');
+		$dotenv->load();
+
+		$value = $_ENV['encryption.key'] ?? getenv('encryption.key');
```
michalsn, line 80: Don't check `$_ENV['encryption.key']`.
```suggestion
		$value = getenv('encryption.key');
```
We have to cleanup variable `encryption.key`. Put this call before instantiating `DotEnv` object.
```php
putenv('encryption.key')
```

tests/system/Config/DotEnvTest.php
```diff
@@ -61,7 +62,7 @@ public function testLoadsHex2Bin()
 		$dotenv = new DotEnv($this->fixturesFolder, 'encryption.env');
 		$dotenv->load();
 
-		$value = getenv('encryption.key');
+		$value = $_ENV['encryption.key'] ?? getenv('encryption.key');
```
michalsn, line 65:
```suggestion
		$value = getenv('encryption.key');
```
and then `putenv('encryption.key')` before the line where we create new `DotEnv` object (if needed).

paulbalandan:
@michalsn , I did your suggestions but tests fail on Windows, which was my first predicament I solved with the `$_ENV` check.

```

Runtime:       PHP 7.4.7 with Xdebug 2.9.4

..FF...........                                                   15 / 15 (100%)

Time: 57.36 seconds, Memory: 50.00 MB

There were 2 failures:

1) CodeIgniter\Config\DotEnvTest::testLoadsHex2Bin
--- Expected
+++ Actual
@@ @@
-'f699c7fd18a8e082d0228932f3acd40e1ef5ef92efcedda32842a211d62f0aa6'
+'c3b6e284a2c387c3bd18c2a8c3a0e2809ac39022e280b032c3b3c2acc3940e1ec3b5c3afe28099c3afc38ec39dc2a32842c2a211c3962f0ac2a6'

C:\Users\P\Desktop\Web Dev\CodeIgniter4\tests\system\Config\DotEnvTest.php:70

2) CodeIgniter\Config\DotEnvTest::testLoadsBase64
--- Expected
+++ Actual
@@ @@
-'L40bKo6b8Nu541LeVeZ1i5RXfGgnkar42CPTfukhGhw='
+'L8KNGyrFveKAusOww5vCucOjUsOeVcOmdeKAueKAnVd8aCfigJjCqsO4w5gjw5N+w6khGhw='

C:\Users\P\Desktop\Web Dev\CodeIgniter4\tests\system\Config\DotEnvTest.php:87
```

michalsn:
@paulbalandan hmm... okay. Windows is tricky 😅 Maybe we can try something like this: remove `putenv('encryption.key')` before creating `DotEnv` object from both tests. And add docblock:
```
* @runInSeparateProcess
* @preserveGlobalState  disabled
```
to the `testLoadsBase64()` test only.

If this will fail too then I'm good with your solution. Also sorry, I know it's "just" tests, but such things bother me.

paulbalandan:
@michalsn , unfortunately that doesn't work either. I have tried also using those docblocks in both tests but to no avail. It is interesting to note that the `getenv` call only fails for the binary strings, but succeeded on others, so I filed a bug report [here](https://bugs.php.net/bug.php?id=79875).

I have noticed that we have `DotEnv::getVariable` method and used that to check the values using `ReflectionMethod` since it's protected, and miraculously I get green builds. Perhaps, we can make that method public to aid in testing?

michalsn:
@paulbalandan I feel like I shouldn't have asked for these changes at all... I'm just wasting your time. Merging.

PR #3334, opened 2020-07-17
Fix method incompatibility of BaseCommand::run()

paulbalandan:
**Description**
The abstract function `BaseCommand::run` has the following method signature: `run(array $params)`. Meanwhile, most of the native commands extending BaseCommand use the method signature `run(array $params = [])` instead. For this I assume that the latter is the intended usage. In IDEs with syntax highlighting, I get method incompatibility errors due to this.

While I don't see it as a BC change, please advise if otherwise. Thanks.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/BaseCommand.php
+++ b/system/CLI/BaseCommand.php
@@ -133,7 +133,7 @@ abstract class BaseCommand
 	 *
 	 * @param array $params
 	 */
-	abstract public function run(array $params = []);
+	abstract public function run(array $params);
 
 	//--------------------------------------------------------------------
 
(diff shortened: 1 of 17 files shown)


michalsn:
Sorry, but this is a BC change.

If someone had command declared like:
```php
public function run(array $params)
```
he will get an error.

MGatner:
@lonnieezell took a pass at this recently with the new commands component. Probably nothing to be done but maybe he can weigh in.

paulbalandan:
If this is BC break, then perhaps the opposite should be done. Normalize all native commands to use `run(array $params)`.

michalsn:
Yes, the opposite shouldn't cause any problems.

paulbalandan:
Understood. I will revise this PR to do the reverse.

MGatner:
Looks good! Glad to get those copy-paste comments changed too 😅

PR #3338, opened 2020-07-17
optimize excerpt text helper : no need strtolower on strpos, replace with stripos

samsonasik:
**Checklist:**
- [x] Securely signed commits


MGatner:
Good catch!

PR #3344, opened 2020-07-17
Model beforeFind

MGatner:
**Description**
This one grew a bit. Original intent was to define the `beforeModel` callback event. That's done but I also discovered I had a logic error with `allowCallbacks` resetting during `trigger()`, because most (now all) methods that use triggers actually call `trigger()` twice - fixed. I discovered a Model test that should not have been passing (`db_job` is not a valid table) - fixed. And also there were no tests for `beforeDelete` - added.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -386,13 +386,15 @@ class Model
 	 */
 	public function find($id = null)
 	{
-		// Call the before event and check for a return
-		$eventData = $this->trigger('beforeFind', ['id' => $id, 'method' => 'find']);
-		if (! empty($eventData['returnData']))
+		if ($this->tempAllowCallbacks)
 		{
-			return $eventData['data'];
+			// Call the before event and check for a return
+			$eventData = $this->trigger('beforeFind', ['id' => $id, 'method' => 'find']);
+			if (! empty($eventData['returnData']))
+			{
+				return $eventData['data'];
+			}
 		}
-
 		$builder = $this->builder();
 
 		if ($this->tempUseSoftDeletes === true)
@@ -420,10 +422,18 @@ class Model
 			$row = $row->getResult($this->tempReturnType);
 		}
 
-		$eventData = $this->trigger('afterFind', ['id' => $id, 'data' => $row]);
+		$eventData = [
+			'id'   => $id,
+			'data' => $row,
+		];
+		if ($this->tempAllowCallbacks)
+		{
+			$eventData = $this->trigger('afterFind', $eventData);
+		}
 
 		$this->tempReturnType     = $this->returnType;
 		$this->tempUseSoftDeletes = $this->useSoftDeletes;
+		$this->tempAllowCallbacks = $this->allowCallbacks;
 
 		return $eventData['data'];
 	}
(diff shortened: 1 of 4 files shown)


tangix:
Just so I understand this. With this, would I be able to define a beforeFind event along the following lines:
```php
public function setPaging()
{
    $builder = $this->builder();

    // Handle paging
    // -------------
    $start = $request->getGet('start');
    $limit = $request->getGet('limit');
    if (!is_null($limit) && !is_null($start)) {
        $builder->limit($limit, $start);
    }
    return ['returnData' => [] ]; // return empty to continue
}
```

separate functions for sorting and finding would be possible to chain.

MGatner:
Yes; during `beforeFind` you can access and influence `builder()` and any other properties of the current model, and these will be taken into effect when the query is actually executed. Additionally you have the options to return `['returnData' = true]` (along with some data) to send back some "intercepted" data

tangix:
Hmmm... some second thoughts here. For my front-end UI to work correctly with paging, I need the total number of records in the result (i.e. without paging applied). I do the following in my BaseModel:

```php

class BaseModel extends \CodeIgniter\Model
{

    protected $useTimestamps = true;
    protected $useSoftDeletes = false;


    /**
     * Handle pagination with Ext.data.virtual.Store
     *
     * @return JsonStoreResponse
     */
    public function getVirtualStoreData()
    {

        $builder = $this->builder();

        // Handle paging
        // -------------
        $start = $request->getGet('start');
        $limit = $request->getGet('limit');
        if (!is_null($limit) && !is_null($start)) {
            $builder->limit($limit, $start);
        }

        // Handle sorting
        // --------------
        $sort = json_decode($request->getGet('sort'));
        if ( ! is_null($sort)) {
            foreach ($sort as $f) {
                if (property_exists($f, 'property') && property_exists($f, 'direction')) {
                    $builder->orderBy($f->property, $f->direction);
                }
            }
        }

        // Handle filtering
        // ----------------
        $filter = json_decode($request->getGet('filter'));
        if ( ! is_null($filter)) {
            foreach ($filter as $f) {
                $builder->orGroupStart();
                if (property_exists($f, 'property') && property_exists($f, 'operator') && property_exists($f,
                        'value')) {
                    switch ($f->operator) {
                        case 'like':
                            $builder->like($f->property, $f->value);
                            break;
                        case '>':
                        case '<':
                            if ( ! in_array($f->property, $this->allowedFields)) {
                                $builder->having($f->property . $f->operator, $f->value);
                            } else {
                                $builder->where($f->property . $f->operator, $f->value);
                            }
                            break;
                        case '==':
                            if ( ! in_array($f->property, $this->allowedFields)) {
                                $builder->having($f->property, $f->value);
                            } else {
                                $builder->where($f->property, $f->value);
                            }
                            break;
                    }
                }
                $builder->groupEnd();
            }
        }

        // Do the stuff
        // ------------
        $sql = $builder->getCompiledSelect(false);
        $query = $this->db->query($sql);
        $row = $query->getResult($this->tempReturnType);
        $query->freeResult();

        $total = $builder->countAllResults();

        $result = new JsonStoreResponse(true, $total);
        $result->setData($row);

        return $result;
    }
}
```

The trick I do is to get the sql from the compiled select and not reset the builder. I then do a `$builder->countAllResults()` to get the total count.

From the Controllers I call it like this:
```php
    public function index()
    {
        $credits = new ExamTypeModel();
        $reply = $credits->select('*')
            ->select('(SELECT SUM(balance) FROM user_credits WHERE typeLink=examTypeID) AS user_balance',
                false)
            ->select('(SELECT SUM(balance) FROM credits WHERE typeLink=examTypeID) AS company_balance',
                false)
            ->getVirtualStoreData();

        return $this->extRespond($reply);
    }
```

I am not 100% sure I see how I would apply the beforeFind hooks to allow me to use `$credits->find()` from the Controller. 
I will think about this during the day.

PR #3348, opened 2020-07-18
Added routes example with multiple segments to docs

tangix:
**Description**
Added an example describing multi-segment routes since this may be good to know.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


paulbalandan:
Maybe you can add if that is not the intended behavior they can use `(:segment)`.

michalsn:
@tangix If you are submitting changes that don't include actual changes to the framework code, please use `[ci skip]` in the commit description. This way Travis build will be skipped. Thanks!

PR #3354, opened 2020-07-19
SQLite3 - Fix to invalid column type results

devorama:
**Description**

If fielddata is called at least one row must be pulled to get the types, Note that if the dataset
has no rows then it will still behave as in the old way where all will return type NULL.

In the bast as now result was pulled or the resultset was at the last row all types returned the default NULL

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


devorama:
Note as SQLite3 only return column types of 
 SQLITE3_INTEGER
SQLITE3_FLOAT
SQLITE3_TEXT 
SQLITE3_BLOB
SQLITE3_NULL

I really wanted to add a additional check when it is of type string to see if column is of type datetime with a test with strtotime, but I realized that it will give false positives as Yesterday etc will also be seen as dates.
For example this table structure
prent

Will now return
SQLITE3_INTEGER
SQLITE3_TEXT   -> actually a date but php driver returns it as TEXT
SQLITE3_TEXT 
SQLITE3_TEXT 
SQLITE3_TEXT 
SQLITE3_TEXT  -> actually a date but php driver returns it as TEXT

devorama:
I added blob and text types also 
to test the return of type 
SQLITE3_BLOB

prent

But for them it returned type NULL for each one of them

system/Database/SQLite3/Result.php
```diff
@@ -95,7 +95,11 @@ public function getFieldData(): array
 		];
 
 		$retVal = [];
-
+		//To insure correct column types are returned a fetch must be done at least one
+		//this does a fetch regardless as all CI result queries do dataSeek calls before
+		//calling fetch arrays, thus this will be save to not mess with any other result calls
```
michalsn, line 100:
```suggestion
		// To ensure that the correct column types are returned, we must perform a fetch
```

system/Database/SQLite3/Result.php
```diff
@@ -95,7 +95,11 @@ public function getFieldData(): array
...
+                $tmpArr = $this->resultID->fetchArray(SQLITE3_NUM );
```
michalsn, line 101:
```suggestion
              $this->resultID->fetchArray(SQLITE3_NUM);  
```

system/Database/SQLite3/Result.php
```diff
@@ -106,7 +110,9 @@ public function getFieldData(): array
 			$retVal[$i]->max_length = null;
 			$retVal[$i]->length     = null;
 		}
-
+                unset($tmpArr);
```
michalsn, line 113:
```suggestion
```

devorama:
> We will also need some tests for this. You can update `Database\Live\GetTest::testGetFieldData`

Still learning this, but will check it out and so add as requered thank you, something new to learn :)

devorama:
Sorry wanted to mention, might it not be better to add on the test table other common field types like float/numeric/decimal and blob's for the other db types etc to test with ?

michalsn:
@devorama Yes, you can even add a dedicated table for this (if this will be helpful). 

It would be great if you could also add tests for other drivers. I know this is a little bit out of scope for this PR but it would make our life easier.

devorama:
> It would be great if you could also add tests for other drivers. I

Will do so yes. I geuss it best to test allt ypes that each DB can provide. I personally would have perferred a more central  list of type names but that wont make sense for all the different types that mysql provide for instance.

devorama:
Will create a seperate table for this now. Will look something like this
in mySQL
prent

and like this in SQLLite
prent

I dont have Postgress (my pc wont be able to handle anymore) Thus when I book in if somebody will be able to test if the creation works on it also

devorama:
@michalsn I just realized now via the testing that the FieldData call used for the Builder is not the one used for the result. Builder uses
_fieldData
and this does a total different type of field data than that of result for result does it based on the data coming back while builder tries todo a fielddata on the tables schema without a result. 
Thus this specific test to use builder wont work, I will have to  add a entry in table and select from it to test the results fielddata and then also have the normal builders fielddata test (currently the builders field data fails totally)

devorama:
Scrap my last comment as it does seem to call the results fielddata, I will make a sepeate test for the connections fielddata method, But what is a concern here via the builder the results result for SQLite3 columnType call si all just null no value and it does register 19 columns, this something weird going on here, will debug and try to fix will first test against mysql also

michalsn:
We're interested only in tests for `Result::getFieldData()` here. Please revert any changes related to the `Connection` classes.

Also, you can test only fields that are supported by the given driver.

devorama:
K, I will revert connection changes.  SQLlite dont have a mapping of fields issue it seems, but I will make the table creation and test for mysql and sqlite3 for now as I dont have postgress.
Note the mysql on works at least , will see what is going on with sql lite

prent

devorama:
SQLlite is officially a problem. If there is no record in the table then the PHP columnType  call returns nothing
If there is a record ONLY the columns that has data in  them will have their types returned else it returns NULL type.
I did a test by inserting just a ID and this is the result
prent

I will emulate after table creation a data insert to get some more correct return results, but this does show that the driver
cant be trusted on its results as it depends on the current rows data in it.

Note the Field types I used for mysql and sqlite is valid for both as sql lite maps these internally as needed.
can see under  Affinity Name Examples, there is a small list as a example that it maps
https://www.sqlite.org/datatype3.html

devorama:
Data returned with newly inserted data will be for SQL lite as follow
prent

devorama:
@michalsn  The connection changes was reversed, Seeding of data was added, creation is done for SQLite and MysqLI at moment as I cannot test Postgres if we can have that tested we can remove the IF clauses on creation and seeding of  data (in theory it should work in postgres)
Note I have added all the type testing specifically to allow us to pick up of there ar changes in the PHP driver return of field types going forward that might require new mapping,this way we cant miss it.

I understand what you mean I must add those that is supported by the driver, I can look to add some of those left out (in my comment of the create table in test) but it depends also on the mySQL version used for JSON is a new field to mySQL, thus cant really add that yet. Thus would suggest we make this the start off point to use and then adjust when we have Postgres on board and iron out any issue that we might run into.

tests/_support/Database/Migrations/20160428212500_Create_test_tables.php
```diff
@@ -91,6 +91,101 @@ public function up()
...
+			     'null'             => true,
+			],
+			'type_timestamp'   => [
+			     'type'             => 'TIMESTAMP',
+			     'null'             => true,
+			],
+			'type_blob'   => [
+			     'type'             => 'BLOB',
+			     'null'             => true,
+			],
```
michalsn, line 184: Please align this entire segment inside `if` correctly.

system/Database/MySQLi/Connection.php
```diff
@@ -495,8 +495,8 @@ public function _fieldData(string $table): array
 		$retVal = [];
 		for ($i = 0, $c = count($query); $i < $c; $i++)
 		{
-			$retVal[$i]       = new \stdClass();
-			$retVal[$i]->name = $query[$i]->Field;
+			$retVal[$i]              = new \stdClass();
+			$retVal[$i]->name        = $query[$i]->Field;
```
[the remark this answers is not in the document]
devorama: I actually made the indent inline with the lines below so that they are uniform,but will correct to old indent.

tests/_support/Database/Seeds/CITestSeeder.php
```diff
@@ -67,6 +67,39 @@ public function run()
 				],
 			],
 		];
+                if (($this->db->DBDriver === 'SQLite3') || ($this->db->DBDriver === 'MySQLi'))
```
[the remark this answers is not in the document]
devorama: Will be removing the DB specific ones as I am installing and testing postgres now also

tests/_support/Database/Seeds/CITestSeeder.php
```diff
@@ -67,6 +67,39 @@ public function run()
...
+					 'type_blob' 		=> 'test blob',
+			];
+			if ($this->db->DBDriver === 'SQLite3') 
+			{
+				 $data['type_test'][0]['type_date']      = '2020/01/11';
+				 $data['type_test'][0]['type_time']      = '15:22:00';
+				 $data['type_test'][0]['type_datetime']  = '2020/06/18 05:12:24';
+				 $data['type_test'][0]['type_timestamp'] = '2019/07/18 21:53:21';
+			 
+			} else  if ($this->db->DBDriver === 'MySQLi') {
```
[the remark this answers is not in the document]
devorama: Will be removing the db specific ones as I am installing and testing postgress now also

tests/system/Database/Live/GetTest.php
```diff
@@ -95,6 +95,55 @@ public function testGetFieldData()
 
 		$this->assertEquals('id', $jobs[0]->name);
 		$this->assertEquals('name', $jobs[1]->name);
+
+		$type_test = $this->db->table('type_test')
```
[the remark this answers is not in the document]
devorama: Have removed driver specific creation and seeding, installing nad testing postgres now

tests/_support/Database/Migrations/20160428212500_Create_test_tables.php
devorama: Will copy and paste to notepad wokring in the browser is bit of hassle.

michalsn:
I'm glad you will be working on Postgres tests too. Please remember about updating `Live\MetadataTest::testListTables` since you added one more table to the database. Thanks.

tests/_support/Database/Migrations/20160428212500_Create_test_tables.php
devorama: Copied and tried to algin values etc as above and below

system/Database/SQLite3/Result.php
```diff
@@ -95,7 +95,9 @@ public function getFieldData(): array
 		];
 
 		$retVal = [];
-
+		// To ensure that the correct column types are returned, we must perform a fetch
+                $this->resultID->fetchArray(SQLITE3_NUM);
```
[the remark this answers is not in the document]
devorama: Sorry I feel dump as I dont seem to understand what change you want from above,sorry about this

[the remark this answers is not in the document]
michalsn: It's about code alignment - following the coding standard makes it easier to keep the code clean.

paulbalandan:
A bit of advice. You may install the `precommit` hook in the `admin` folder into your `.git/hooks` folder so that it will automatically check if your code adheres to the coding standard before you do a commit. Just run `bash admin/setup.sh` in your terminal and you're done.

devorama:
@paulbalandan  Thank you I will try this, still learning ,but all of this helps alot thank you.

devorama:
I will have to put in seperate pull for Postgres as it cannot handle all the types. Will have to make it seperate for postgres at moment the table create, insert and check. Then create a seperate pull request to fix the types that dont map correctly so that one can use the same types in all 3 databes for create etc.
prent

devorama:
@paulbalandan  I have done the following

>  Just run `bash admin/setup.sh` in your terminal and you're done.

But now it seems it got its directories confused, will try and see what I can figure out on it

prent

paulbalandan:
> @paulbalandan I have done the following
> 
> > Just run `bash admin/setup.sh` in your terminal and you're done.
> 
> But now it seems it got its directories confused, will try and see what I can figure out on it
> 
> prent
> 
> Running directly from bash in the hooks folder creates this.
> prent
> 
> But on the bright side I got my system to now work with the 2 git users, that was one of the main problems I had with it I had todo all the code igninter coding in browser as I could not update and push from my drive.

Looking at the errors, you must ensure that you have run the command in your project's root, i.e. ` C:\development_CI4\CI4dev`. Then it seems you have not installed the dev dependency PHP_Codesniffer?

system/Database/SQLite3/Result.php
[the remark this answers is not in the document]
devorama: Hi @michalsn  if you can please view I belive I did it correct now when I pulled the source down to my PC I saw it was not aligned and re-align it and booked in. Will try to get the tools working on my pc to help with source format checking going forward

devorama:
@paulbalandan  Thanks for info, yes I dont have it installed I saw that later on, will try and get it installed my PC is just so old I hope it can take it :) Thanks for the help.

system/Database/SQLite3/Result.php
```diff
@@ -95,7 +95,8 @@ public function getFieldData(): array
 		];
 
 		$retVal = [];
-
+        $this->resultID->fetchArray(SQLITE3_NUM);
```
michalsn, line 98: Please align the indent with line 97.

tests/system/Database/Live/GetTest.php
```diff
@@ -95,6 +95,73 @@ public function testGetFieldData()
...
+						 ->get()
+						 ->getFieldData();
+		
+				if ($this->db->DBDriver === 'SQLite3')
```
michalsn, line 103: Please align the indent with line 104.

system/Database/SQLite3/Result.php
devorama: I figured out why, it had spaces infront that why I keep on beying confused as it looks correct for me and then it diff.
Before

prent

After
prent

tests/system/Database/Live/GetTest.php
devorama: I hope it correct now, I had the special chars on now to see tabs and spaces, sorry you have to babysit me like this.

PR #3357, opened 2020-07-19
optimize Router::checkRoutes() : remove unneded uri check when routes is empty

samsonasik:
**Checklist:**
- [x] Securely signed commits


michalsn:
Now we really don't waste any time 👍

PR #3358, opened 2020-07-19
optimize Model::doProtectFields(): remove unneded count($data)

samsonasik:
use truthy check is enough

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1265,14 +1265,11 @@ class Model
 			throw DataException::forInvalidAllowedFields(get_class($this));
 		}
 
-		if (is_array($data) && $data)
+		foreach ($data as $key => $val)
 		{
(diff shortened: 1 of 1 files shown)


system/Model.php
```diff
@@ -1265,7 +1265,7 @@ protected function doProtectFields(array $data): array
 			throw DataException::forInvalidAllowedFields(get_class($this));
 		}
 
-		if (is_array($data) && count($data))
+		if (is_array($data) && $data)
```
paulbalandan, line 1268: Maybe we can dispense with the use of the ` && $data` check? The loop won't execute if we would supply an empty array.

samsonasik: updated,  remove unneeed if (is_array($data)), it is always an array as a field parameter

PR #3361, opened 2020-07-19
DotEnv: No binary strings to environment variables

paulbalandan:
**Description**
Per comment by the php maintainer in my bug report [here](https://bugs.php.net/bug.php?id=79875), they presented two advice in properly dealing with binary strings in the environment. This applies only for Windows. So I'll be presenting two mutually-exclusive PRs for the CI4 maintainers to decide the appropriate approach.

This is the **first** alternative: **Do not pass binary strings to the environment variables.** `BaseConfig` already has a parser for the prefix handling so `DotEnv` can be relieved from that responsibility.

Result for DotEnvTest
```
PS C:\Users\P\Desktop\Web Dev\CodeIgniter4> vendor/bin/phpunit -v --filter DotEnvTest

Runtime:       PHP 7.4.8 with Xdebug 2.9.4

...............                                                   15 / 15 (100%)

Time: 54.49 seconds, Memory: 50.00 MB

OK (15 tests, 49 assertions)
```

Result for BaseConfigTest
```
PS C:\Users\P\Desktop\Web Dev\CodeIgniter4> vendor/bin/phpunit -v --filter BaseConfigTest

Runtime:       PHP 7.4.8 with Xdebug 2.9.4

................                                                  16 / 16 (100%)

Time: 48.13 seconds, Memory: 50.00 MB

OK (16 tests, 45 assertions)
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/DotEnv.php
+++ b/system/Config/DotEnv.php
@@ -184,20 +184,6 @@ class DotEnv
 
 		$value = $this->resolveNestedVariables($value);
 
-		if ($name === 'encryption.key')
-		{
-			// Handle hex2bin prefix
-			if (strpos($value, 'hex2bin:') === 0)
-			{
-				$value = hex2bin(substr($value, 8));
-			}
-			// Handle base64 prefix
-			elseif (strpos($value, 'base64:') === 0)
-			{
-				$value = base64_decode(substr($value, 7), true);
-			}
-		}
-
 		return [
 			$name,
 			$value,
--- a/tests/system/Config/DotEnvTest.php
+++ b/tests/system/Config/DotEnvTest.php
@@ -57,36 +57,32 @@ class DotEnvTest extends \CodeIgniter\Test\CIUnitTestCase
 
 	//--------------------------------------------------------------------
 
+	/**
+	 * @runInSeparateProcess
+	 * @preserveGlobalState  disabled
+	 */
 	public function testLoadsHex2Bin()
 	{
 		$dotenv = new DotEnv($this->fixturesFolder, 'encryption.env');
 		$dotenv->load();
 
-		$m = new \ReflectionMethod($dotenv, 'getVariable');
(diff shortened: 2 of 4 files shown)


michalsn:
I had no idea that Windows handles this in a such way. Thank you for digging into it @paulbalandan. I really appreciate it.

I think we should go with **this** approach. We should mention this change in the release notes but there is really no BC as no one will probably try to use `getenv('encryption.key')` alone.

If we would choose #3362 then on Windows, we would need to use all this conversion to make it work. It would mean a lot of additional explanation and code for developers. I believe it's not worth it since even if people develop their apps on Windows, at the end of the day they deploy their code to the Linux machine (most of the time).

Any thoughts @lonnieezell, @MGatner?

michalsn:
@paulbalandan I see no feedback so most likely I will go with this PR. Can you merge develop branch and add a note to changelog for v4.0.5 that will say something like:
> Handling prefix `hex2bin:` in .env file for variable `encryption.key` has changed. If you used to explicit functions calls like `getenv('encryption.key')` or `env('encryption.key')` with a special prefix inside your variable, then these functions will no longer return a binary string. This change is due to incompatibility with handling binary strings in environment variables on a Windows platform. Accessing `encryption.key` via class config variable remains unchanged and is automatically converted to a binary string when using a `hex2bin:` prefix.

paulbalandan:
@michalsn, for me I prefer this PR more as it is much cleaner and will work well with `key:generate` spark command I'm currently developing. So, I'll do that changelog update later.

Wait. Just to confirm, this changelog you are referring is the `rst` version for the user guide, right?

michalsn:
@paulbalandan Yes, changelog in user guide.

michalsn:
Thank you.

PR #3362, opened 2020-07-19
DotEnv: Convert binary string to UTF-8 for Windows

paulbalandan:
**Description**
Per discussion in #3361 , this is **second** alternative to binary string handling. **Convert binary strings to UTF-8 codepage and convert them back on retrieval by `BaseConfig`.**

Basically, we're aiming for interoperability as what is Windows is doing right now. So for proper handling, we convert first the binary string parsed by `DotEnv` into UTF-8 using the `sapi_windows_cp_conv` function available since PHP/7.1. Afterwards, when these values will be used to populate the configuration properties, we instruct `BaseConfig` to convert it back. Conversion is thru `ANSI <=> UTF-8`.

Result for DotEnvTest
```
PS C:\Users\P\Desktop\Web Dev\CodeIgniter4> vendor/bin/phpunit -v --filter DotEnvTest

Runtime:       PHP 7.4.8 with Xdebug 2.9.4

...............                                                   15 / 15 (100%)

Time: 42.4 seconds, Memory: 50.00 MB

OK (15 tests, 51 assertions)
```

Result for BaseConfigTest
```
PS C:\Users\P\Desktop\Web Dev\CodeIgniter4> vendor/bin/phpunit -v --filter BaseConfigTest

Runtime:       PHP 7.4.8 with Xdebug 2.9.4

................                                                  16 / 16 (100%)

Time: 47.68 seconds, Memory: 50.00 MB

OK (16 tests, 45 assertions)
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/BaseConfig.php
+++ b/system/Config/BaseConfig.php
@@ -94,8 +94,27 @@ class BaseConfig
 
 			if ($shortPrefix === 'encryption' && $property === 'key')
 			{
+				// Handle values from .env already parsed by DotEnv
+				if (strpos($this->$property, 'hex2bin:') === false && strpos($this->$property, 'base64:') === false)
+				{
+					if ('\\' === DIRECTORY_SEPARATOR)
+					{
+						/**
+						 * Windows requires binary strings to be converted to UTF-8 codepage
+						 * before being stored in environment. Since DotEnv already did the conversion
+						 * and we need the original binary string, we'll just reverse the process here.
+						 *
+						 * @see https://bugs.php.net/bug.php?id=79875
+						 */
+						// @codeCoverageIgnoreStart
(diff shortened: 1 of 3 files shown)


ivantcholakov:
Just a side note about:

```php
if ('\\' === DIRECTORY_SEPARATOR)
```

It seems that Windows detection would often enough. I think, early at bootstrapping process a constant IS_WINDOWS (or whatever you name it) should be created. People pick up various expressions to detect Windows OS, if a such constant is defined, detection would be the same everywhere and code readability would be better.

michalsn:
Problem resolved via #3361

PR #3365, opened 2020-07-20
New common function: clean_path

paulbalandan:
**Description**
Use a common function `clean_path` as a means for nicer looking output instead of defining it as a method for class that needs it, e.g. `Exceptions`, `Logger`, `BaseCollector`, etc. With this, we can dispense use of redundant methods like `Exceptions::cleanPath` in favor of a global function.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -112,6 +112,40 @@ if (! function_exists('cache'))
 	}
 }
 
+if (! function_exists('clean_path'))
+{
+	/**
+	 * A convenience method to clean paths for
+	 * a nicer looking output. Useful for exception
+	 * handling, error logging, etc.
+	 *
+	 * @param string $path
+	 *
+	 * @return string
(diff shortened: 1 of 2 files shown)


lonnieezell:
When I saw this I was thinking, "Wait - this is already in there. I remember writing it!" Only to realize that I did that in some exception class refactoring I haven't had a chance to get back to. :) 

Thanks for this.

paulbalandan:
I remember you telling me about that refactoring with the exception class when I made a PR similar to this on reducing redundant code.

MGatner:
Thanks for handling `realpath()` correctly. 🤗

paulbalandan:
Well, all thanks to your review comment before. 😁

PR #3366, opened 2020-07-20
Optimize Autoloader::loadInNamespace in getting path

paulbalandan:
**Description**
Currently, directory `rtrim` accounts only for forward slash but not for backward slash, which may result to the following:
```
$this->requireFile('C:\\Users\\P\\Desktop\\Web Dev\\CodeIgniter4\\system\\/Database/MySQLi/Connection.php')
```

With the new changes from this PR, the path will look like this:
```
$this->requireFile('C:\\Users\\P\\Desktop\\Web Dev\\CodeIgniter4\\system\\Database\\MySQLi\\Connection.php')
```

I know this seems for fancy output but I'm creating a custom exception handling output and the `\\/` and inconsistent directory separators looks unaesthetic.

PS. Changed the tests to change `/` to the `DIRECTORY_SEPARATOR` for it to pass on Windows.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -287,12 +287,11 @@ class Autoloader
 		{
 			foreach ($directories as $directory)
 			{
-				$directory = rtrim($directory, '/');
+				$directory = rtrim($directory, '\\/');
 
 				if (strpos($class, $namespace) === 0)
 				{
-					$filePath = $directory . str_replace('\\', '/',
-							substr($class, strlen($namespace))) . '.php';
+					$filePath = $directory . str_replace('\\', DIRECTORY_SEPARATOR, substr($class, strlen($namespace))) . '.php';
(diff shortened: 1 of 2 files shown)


lonnieezell:
@paulbalandan Custom exception handling? Can you hit me up on Slack? That was something that I had in the roadmap for 4.1 so would love to make sure we're heading in the same direction. Thanks!

lonnieezell:
Oh, and this looks good. Merging.

paulbalandan:
@lonnieezell , that would be great and I'd love to. But I do not have a Slack account. Do you have any other means?

lonnieezell:
@paulbalandan sorry - this notification got lost in all of the other notifications from the great work everyone is doing! :) 

Maybe just email me and we'll go that way for now? lonnieje@gmail.com

PR #3375, opened 2020-07-21
Remove close brace, add missing colon and text format

turtlix:
Remove excess close brace in optional parameters.
Add missing colon for return type.
Add formatted note.
Change directory name config into Config.
Format file location as bold.


MGatner:
Nice clean up! Thank you

Issue #3368, opened 2020-07-20
Bug: Model can not insert Entity


natanfelles:
**Describe the bug**

Model is not inserting entity values.

The following code:

```php
<?php namespace App\Controllers;

use App\Entities\User;
use App\Models\Users;

class Home extends BaseController
{
	public function index()
	{
		$user = new User([
			'name' => 'John Doe',
			'email' => 'jd@mail.com',
			'password' => 'secret',
		]);

		$users = new Users();
		$users->insert($user);
	}
}

```

throws the exception:

**CodeIgniter\Database\Exceptions\DataException**
_There is no data to insert._ 

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
Model/Entity

**Expected behavior, and steps to reproduce if appropriate**
Expect the data to be inserted.

**Context**
 - OS: Ubuntu
 - Web server: PHP
 - PHP version: 7.4.8

natanfelles:
See also: https://forum.codeigniter.com/thread-77096.html

michalsn:
I couldn't reproduce your error. In the linked forum thread you're saying something about custom Entity class. Could you please provide us your Model and Entity?

natanfelles:
Hello @michalsn !

Here is the model:

```php
<?php namespace App\Models;

use CodeIgniter\Model;
use Faker\Generator;

class Users extends Model
{
    protected $table = 'users';
    protected $allowedFields = [
        'name',
        'email',
        'password',
    ];
    protected $useTimestamps = true;
    protected $returnType = User::class;

    public function fake(Generator &$faker)
    {
        return [
            'name' => $faker->name,
            'email' => $faker->email,
            'password' => password_hash('password', PASSWORD_DEFAULT),
        ];
    }
} 
```

And here is the entity, return type of the model:

```php
<?php namespace App\Models;

use CodeIgniter\Entity;

class User extends Entity
{
    protected $id;
    protected $name;
    protected $email;
    protected $password;
} 
```

And the controller, where the `fake` is called to test:

```php
<?php namespace App\Controllers;

use App\Models\Users as UsersModel;

class Home extends BaseController
{
    public function index()
    {
        helper('test');
        $user = fake(UsersModel::class);
        var_dump($user);
    }
} 
```

lonnieezell:
Actually, the following code example is not how it should work for quite a while: 

```
<?php namespace App\Models;

use CodeIgniter\Entity;

class User extends Entity
{
[...]

PR #3377, opened 2020-07-21
Fix fill() method in Entity class

michalsn:
**Description**
When we declare class properties for Entity class then `__set()` method won't be called.

> __set() is run when writing data to inaccessible (protected or private) or non-existing properties

See: #3368

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -124,7 +124,7 @@ class Entity implements \JsonSerializable
 
 		foreach ($data as $key => $value)
 		{
-			$this->$key = $value;
+			$this->__set($key, $value);
 		}
 
 		return $this;
--- a/user_guide_src/source/changelogs/v4.0.5.rst
+++ b/user_guide_src/source/changelogs/v4.0.5.rst
@@ -1,0 +1,11 @@
+Version 4.0.5
+====================================================
+
+Release Date: Not released
+
+**4.0.5 release of CodeIgniter4**
+
+Enhancements
+------------
+
+- Fixed a bug in ``Entity`` class where declaring class parameters was preventing data propagation to the ``attributes`` array.


natanfelles:
I confirm that this resolved issue #3368 . Thank you!

MGatner:
This is a breaking change and IMO it is more of a philosophical change than a bug fix. There are legit reasons a developer may want to declare properties and have them filled but not included in `toArray`, etc. @natanfelles' issue seems to have been in misunderstanding how Entity properties are declared - you should add them as fields in `$attributes` rather than class properties.

michalsn:
@MGatner I can't agree with you as we should take a look at how this method looked like before: https://github.com/codeigniter4/CodeIgniter4/pull/2968/files

From my perspective we simply forgot that parameters may be declared in the class and then `__set()` won't be triggered.

MGatner:
I think you're right and that was an unintentional BC, but I don't know if it merits another BC to "fix" it. 

My personal opinion is I think this is a good change. As is, if someone had a field named "dates" and tried to fill it into an Entity it would really screw the class: `$entity = new Entity(['dates' => 2674862738])`. This would overwrite the date property, so maybe that constitutes a bug and is worth another BC? Let's keep this discussion going.

MGatner:
Yeah the more I think about this the more I think you're right. We will need to note this as a minor breaking change - unlikely that many people will be affected but it could still mess something up. Can we start a `4.0.5` changelog with this PR and include a note on this interaction? Also I'd like to be sure @lonnieezell sees this.

lonnieezell:
I agree this should be fixed. The original intent was that you do exactly what @natanfelles did in his example. so, yes, make a note of it in the the Upgrade Guide that we should start maintaining and I think it's good. :)

lonnieezell:
After reading back through the original post, I need to clarify. You should be able to do this: 

```
$user = new User([
  'email' => 'joe@example.com',
  'name' => 'Joe Smith'
]);
$users = new UserModel();
$users->save($user);
```

Double-checking against the Model and the change is good.

michalsn:
I've added info to the changelog.

PR #3382, opened 2020-07-22
Add UTC offset to spark header

paulbalandan:
**Description**
I know it is a bit of nitpicking but I always get confused why my local time is different from server time. I'm suggesting putting a UTC offset to the server time printed by spark.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


lonnieezell:
I like this. Merging.

Issue #3251, opened 2020-07-08
Bug: Empty Entity date attributes receive current timestamp when accessed


JanZelenka:
I have the following scenario:

* Using the Entity class
* Setting the $dates property to an array of attributes that are supposed to be of type DateTime
* On loading the Entity data from the database, the dates have no value
* When trying to get the value, *__get()* doesn't check for an empty value, calls the mutateDate method, which in turn end up calling *Time::parse()* resulting in the current timestamp.

*CodeIgniter 4 version:* 4.0.3
*Affected module:* Entity class
*PHP version:* 7.2

Instrye:
This is expected, because `new DateTime('')` return current time. I think this is a problem with your data rather than a problem with the framework.

JanZelenka:
Thanks for the feedback but I still believe this is a framework issue. Imagine this scenario:

I have a table _Process_ with column _startedOn_ in my database. When this column is empty, it indicates a process that hasn't started yet.
In CodeIgniter I create an Entity class _Process_. I load data of a specific process into _$CurrentProcess_ of type _Process_ and try this:
`if ( empty( $CurrentProcess->starterdOn ) ) { $CurrentProcess->start() }`

My processes never get started because even though in the database they have no startup date specified, CodeIgniter substitutes my empty value with the current timestamp.

Bottom line, if a column is empty in the database, the respective attribute in your Entity must be empty as well.

Instrye:
I will submit a PR based on your suggestion, but whether to merge or not depends on the judgment of the administrator.

MGatner:
I think the expectation is that you would use `null` in your database, so the result wouldn't be a string. While I don't know what one would expect from mutating `""` I agree getting back the current time stamp is unexpected.

JanZelenka:
Yeah, that would indeed be the expectation - except my database engine (FileMaker) doesn't recognize `null`, `""` is its equivalent of `null`

michalsn:
Well... I believe that in the Entity class we have a special use for the empty string and it should not be changed.

@JanZelenka If your database is specific and there is no support for `null` values then you should probably extend the Entity class and make changes just for your own use.

An empty string should still be supported and return the current date - this is expected behavior.

PR #3395, opened 2020-07-24
fix. Entity::mutateDate set empty return current time

Instrye:
Fix: #3251 Entity::mutateDate set empty return current time
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -495,7 +495,7 @@ class Entity implements \JsonSerializable
 			return Time::createFromTimestamp($value);
 		}
 
-		if (is_string($value))
+		if (is_string($value) && ! empty($value))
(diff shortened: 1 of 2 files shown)


michalsn:
I'm afraid this can break someone's code. Please, consider this scenario:
```php
        $project = new Project([
            'name'        => 'Dan',
            'custom_date' => '',
        ]);

        $model = new ProjectModel();
        $model->insert($project);
```
Normally this would set a current date to the `custom_date` field (of course if this field would be present in `$dates` array), but with this change, it won't be possible.

An empty string can be actually used in many cases/scenarios.

PR #3397, opened 2020-07-24
Command "key:generate"

paulbalandan:
**Description**
Added new command for encryption key generation to `.env` file.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Encryption/GenerateKey.php
+++ b/system/Commands/Encryption/GenerateKey.php
@@ -1,0 +1,225 @@
+<?php
+
+/**
+ * CodeIgniter
+ *
+ * An open source application development framework for PHP
+ *
+ * This content is released under the MIT License (MIT)
+ *
+ * Copyright (c) 2014-2019 British Columbia Institute of Technology
+ * Copyright (c) 2019-2020 CodeIgniter Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


paulbalandan:
Tests are failing. 😢

michalsn:
How about making one of the methods (hex or base64) a default one? There is no much difference and it would be just easier for developers to type `key:generate`.

I would add support for the key length (with default value).

I'm also wondering should we support the case when there is no `env` file at all (even this default one shipped with the framework). Maybe it's a good idea to mention in the description that we will place a key in the `env` file.

paulbalandan:
Ok. I will do the default encoding and description update. For the support for key length, I don't pretty much understood what you mean. Can you explain a bit more?

michalsn:
The length of the encryption key should vary depending on the encryption algorithm we use. 
https://codeigniter.com/user_guide/libraries/encryption.html#setting-your-encryption-key

michalsn:
What do you think about checking if `env` file actually exists before trying to making a copy?

Also, the last thing - maybe we should replace `hex` with `hext2bin` - since you made it a default method it won't be a big problem to type it and we would keep the names consistency.

paulbalandan:
If the default shipped `env` file is not actually existing before copy, what should be our course of action then? Should we throw an exception, or just output the `$newKey` and an error message to `STDERR`?

michalsn:
Yes, probably just a nice error message that says the `env` file is not available and then we should print the key on the screen.

michalsn:
This is really nice feature. Thank you!

PR #3398, opened 2020-07-24
Allow dashes within CLI options

### LINKED ISSUE #3205: Bug: CLI: Smell on reading parameters
I am testing a command line like this:

```
php spark assets:compile test-less-min
```

The name test-less-min does not appear within the parameters BaseCommand::run(array $params) (at the successor class).

```
php spark assets:compile test
```

This works.

If there are dashes inside string parameters, they (the parameters) are wiped out.

I briefly tried to locate what might cause this effect, here is a visual check, I am not sure whether it is enough:

https://github.com/codeigniter4/CodeIgniter4/blob/d9864e33758aca9aca9dc44c121e402e2d51136f/system/CLI/CLI.php#L871 - what comment says as an intention and what the following line does are different.

https://github.com/codeigniter4/CodeIgniter4/blob/d9864e33758aca9aca9dc44c121e402e2d51136f/system/CLI/CLI.php#L877 - I wonder here why ALL the dashes are to disappear. Maybe using ltrim() would do the job?

paulbalandan:
**Description**
#3208 fixed #3205 with regard to resolving dashes within CLI segments. However, if we have an option like `-opt-in` then `CLI::getOptions` will give `optin` as the key, removing the dash in between.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -899,7 +899,7 @@ class CLI
 				continue;
 			}
 
-			$arg   = str_replace('-', '', $_SERVER['argv'][$i]);
+			$arg   = ltrim($_SERVER['argv'][$i], '-');
 			$value = null;
 
 			// if there is a following segment, and it doesn't start with a dash, it's a value.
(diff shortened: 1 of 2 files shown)


michalsn:
Yes, good catch. Thank you!

PR #3399, opened 2020-07-24
return format hours

bagimuka:
line 1209
$phrase = lang('Time.hours', [abs($hours)]);
$before = $hours < 0;

**Description**
return format in Hours.

### DIFF
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -1205,7 +1205,8 @@ class Time extends DateTime
 		else if ($hours !== 0)
 		{
 			// Display the actual time instead of a regular phrase.
-			return $this->format('g:i a');
+			$phrase = lang('Time.hours', [abs($hours)]);
+			$before = $hours < 0;
 		}
 		else if ($minutes !== 0)
 		{


michalsn:
Just trying to start Travis build.

michalsn:
I just realized that the current behavior is clearly desired since it's documented in the user guide.

From the other side... the proposed change also has a lot of sense, even if it is not precise in minutes. I'm not sure what to do with it. @MGatner what do you think about this one?

bagimuka:
> I just realized that the current behavior is clearly desired since it's documented in the user guide.
> 
> From the other side... the proposed change also has a lot of sense, even if it is not precise in minutes. I'm not sure what to do with it. @MGatner what do you think about this one?

this code will appear 1 Hour - 23 Hours. if no do that code.. he will bug appear in time.. trust me. and test code that in your localhost.

MGatner:
The current behavior is odd, since `hours` is the only non-relative format. Someone clearly made this call a while ago. 

Here's my opinion: I think the current PR represents an opinion, but even if we agree with it I don't know that it merits a change. However, I've been interested in a "smart" Date display that isn't relative; e.g. something from today would be "10:34am", from this year would be "June 16", from 2014 would be "January 4, 2014". Get it? It only displays the date part needed relative to the current time. 

So what if we add a `$relative` parameter, changed `hours` to this PR and moved the above description to `$relative = false`?

bagimuka:
> The current behavior is odd, since `hours` is the only non-relative format. Someone clearly made this call a while ago.
> 
> Here's my opinion: I think the current PR represents an opinion, but even if we agree with it I don't know that it merits a change. However, I've been interested in a "smart" Date display that isn't relative; e.g. something from today would be "10:34am", from this year would be "June 16", from 2014 would be "January 4, 2014". Get it? It only displays the date part needed relative to the current time.
> 
> So what if we add a `$relative` parameter, changed `hours` to this PR and moved the above description to `$relative = false`?

yeah. use set true or false . if user want use format in return hours.

michalsn:
I was looking at other libraries that utilize its own "humanize" functions for dates and this change seems a good idea IMO. Also, I believe that proposed behavior would be something anyone would expect.

I'm not sure about adding a parameter to this method though. It seems that returning formatted date doesn't belong here and we should rather create an additional method to handle this - something like `formattedDifference()` maybe?

@bagimuka If you want this PR to be merged, we need to fix [tests](https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/I18n/TimeTest.php#L913) and update the [user guide](https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/libraries/time.rst).

michalsn:
Okay, I just checked that the code initially worked as in this PR but was changed. I have no choice but to ask @lonnieezell why are we handling it this way? Do you mind if we change it back?

lonnieezell:
If I approved that change before, I missed that it changed the date. There shouldn't have been a change in results, but just having it localized, IIRC.

MGatner:
That's encouraging. I'm still interested in the non-relative version but I think @michalsn is right it could be it's own method.

bagimuka:
so how now @lonnieezell ? is that need fix or not?

michalsn:
@bagimuka As I understand your changes are fine. We just need an update for tests and user guide.

PR #3401, opened 2020-07-24
Command() parse args correctly

paulbalandan:
**Description**
The `command()` common function presently parses the arguments list by using a call to `explode` only. While this is not a problem when list consists only of arguments, e.g. `test param1 param2`. This may cause problems, however, when we supplied options to the list, e.g. `test param1 -option1 value1 param2`. The `explode` call will give a one-dimensional array where the option keys themselves are the values. So instead of an option `option1 => $value1` we get instead `[1] => option1`.

This PR attempts to emulate `CLI::parseCommandline` in terms of parsing arguments.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


system/Common.php
```diff
@@ -163,8 +163,31 @@ function command(string $command)
 	{
 		$runner = service('commands');
 
-		$params  = explode(' ', $command);
-		$command = array_shift($params);
+		$args    = explode(' ', $command);
+		$command = array_shift($args);
+
+		$params = [];
+		for ($i = 0, $c = count($args); $i < $c; $i++)
```
samsonasik, line 170: can use foreach which is faster, ref https://phpsandbox.io/n/super-fog-72g7 

```bash
For took: 0.069ms
For with count() took: 0.442ms
Foreach took: 0.132ms
```

paulbalandan: Ok. But how will I check the next item `[$i + 1]` using `foreach`? I tried foreach first but stumbled here.

samsonasik: well, you can do: 

```php
foreach ($args as $i => $arg)
{
    if ($args[$i+1] === 'foo') 
   {

   }
}
```

paulbalandan: Alright. I forgot that `foreach` has another syntax. Will change.

PR #3408, opened 2020-07-26
Allow to customize the request's headers in FeatureTestCase

lefuentes:
**Description**
When you develop a protected api, sometimes you need to send tokens in the request header to authenticate that you can use this api. So you can create a filter that verifies that field in the header of the request.

When you want to test this API, with FeatureTestCase, you can't because it doesn't have the option to add headers to the order.

So I added this method, `withHeaders`, which before submitting the order, inserts custom headers.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -102,6 +102,25 @@ trait FeatureTestTrait
 	}
 
 	/**
+	 * Set request's headers
+	 * 
+	 * Example of use 
+	 * withHeaders([
+	 *  'Authorization' => 'Token'
+	 * ])
+	 * 
+	 * @param array|null Array of headers
+	 * 
+	 * @return $this
+	 */
+	public function withHeaders(array $headers = null)
+	{
+		$this->headers = $headers;
+
+		return $this;
+	}
+
+	/**
 	 * Don't run any events while running this test.
 	 *
 	 * @return $this
@@ -143,6 +162,7 @@ trait FeatureTestTrait
 		$_SERVER['REQUEST_METHOD'] = $method;
 
 		$request = $this->setupRequest($method, $path);
+		$request = $this->setupHeaders($request);
 		$request = $this->populateGlobals($method, $request, $params);
 
 		// Make sure the RouteCollection knows what method we're using...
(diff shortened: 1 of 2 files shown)


michalsn:
This looks good to me, but I'd like @MGatner to approve it first.

system/Test/FeatureTestTrait.php
```diff
@@ -101,6 +101,25 @@ public function withSession(array $values = null)
...
+	 * Example of use 
+	 * withHeaders([
+	 *  'Authorization' => 'Token'
+	 * ])
+	 * 
+	 * @param array|null Array of headers
+	 * 
+	 * @return $this
+	 */
+	public function withHeaders(array $headers = null)
```
paulbalandan, line 116: Maybe we can have the default value of `$headers` to an empty array `[]`, instead? With this change, we can remove the need for the not empty check in your foreach later.

system/Test/FeatureTestTrait.php
```diff
@@ -302,6 +322,26 @@ protected function setupRequest(string $method, string $path = null): IncomingRe
...
+	/**
+	 * Setup the custom request's headers
+	 * 
+	 * @param \CodeIgniter\HTTP\Request $request
+	 * 
+	 * @return \CodeIgniter\HTTP\IncomingRequest
+	 */
+	protected function setupHeaders(Request $request)
+	{
+		if(!empty($this->headers))
```
paulbalandan, line 334: This conditional can be removed if we have the default value set to `[]` instead of `null`.

MGatner: We actually need to keep this since this is a Trait and there is no guarantee that a developer has remembered to add the class property.

system/Test/FeatureTestTrait.php
```diff
@@ -302,6 +322,26 @@ protected function setupRequest(string $method, string $path = null): IncomingRe
...
+		{
+			foreach($this->headers as $name => $value)
```
MGatner, line 336: Leave a space here: `foreach (`

system/Test/FeatureTestTrait.php
MGatner: I like using empty array as the default since that is the class property default.

MGatner:
Looks great! Just waiting on tests - shouldn't be a problem with these minor changes though.

lefuentes:
Thanks to all!

MGatner:
Looks good. Thanks so much @lefuentes - very nice first PR. I look forward to your next one!

PR #3409, opened 2020-07-26
Add url_to() URL helper function

ImMaax:
This PR adds a URL helper function called `url_to()`.

**Description**
I added a URL helper called `url_to()`. It creates absolute URLs based on controllers. I'm commonly using this function in basically all of my CodeIgniter projects. Here is an example:

```php
service('routes')->get('/welcome', 'Pages::welcome');
echo url_to('Pages::welcome'); # => http://example.com/welcome

service('routes')->get('/page/(:segment)', 'Pages::view/$1');
echo url_to('Pages::view', 'home'); # => http://example.com/page/home
```

It is meant to be used inside of views to generate links. The main advantage of this function is that you can change your routes later on, but not having to change your links across all your views.

This is the first time I'm contributing code to CodeIgniter, I hope I didn't miss something.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -648,7 +648,9 @@ if (! function_exists('url_to'))
 	 */
 	function url_to(string $controller, ...$args): string
 	{
-		if (! class_exists($controller))
+		$class = explode('::', $controller)[0];
+
+		if (! class_exists($class))
 		{
 			$controller = service('routes')->getDefaultNamespace() . $controller;
 		}
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -650,18 +650,23 @@ if (! function_exists('url_to'))
 	 */
 	function url_to(string $controller, ...$args): string
 	{
-		if (strpos($controller, '::') !== false)
-		{
-			$class = explode('::', $controller)[0];
-
-			if (! class_exists($class))
-			{
-				$controller = service('routes')->getDefaultNamespace() . $controller;
-			}
-		}
-
 		if (! $route = route_to($controller, ...$args))
 		{
+			if (strpos($controller, '::') !== false)
+			{
+				$class = explode('::', $controller)[0];
+
+				if (! class_exists($class))
+				{
+					$controller = service('routes')->getDefaultNamespace() . $controller;
+				}
+			}
+
+			if ($route = route_to($controller, ...$args))
+			{
+				return site_url($route);
+			}
+
 			$explode = explode('::', $controller);
 
 			if (isset($explode[1]))
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -650,11 +650,6 @@ if (! function_exists('url_to'))
 	 */
 	function url_to(string $controller, ...$args): string
 	{
-		if (! $route = route_to($controller, ...$args))
-		{
-			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
-		}
-
 		if (strpos($controller, '::') !== false)
 		{
 			$class = explode('::', $controller)[0];
@@ -663,6 +658,11 @@ if (! function_exists('url_to'))
 			{
 				$controller = service('routes')->getDefaultNamespace() . $controller;
 			}
+		}
+
+		if (! $route = route_to($controller, ...$args))
+		{
+			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
 		}
 
 		return site_url($route);
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -648,11 +648,14 @@ if (! function_exists('url_to'))
 	 */
 	function url_to(string $controller, ...$args): string
 	{
-		$class = explode('::', $controller)[0];
-
-		if (! class_exists($class))
+		if (strpos($controller, '::') !== false)
 		{
-			$controller = service('routes')->getDefaultNamespace() . $controller;
+			$class = explode('::', $controller)[0];
+
+			if (! class_exists($class))
+			{
+				$controller = service('routes')->getDefaultNamespace() . $controller;
+			}
 		}
 
 		return site_url(route_to($controller, ...$args));
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -644,13 +644,15 @@ if (! function_exists('url_to'))
 	 * @param string $controller
 	 * @param mixed  ...$args
 	 *
+	 * @throws \CodeIgniter\Router\Exceptions\RouterException
+	 *
 	 * @return string
 	 */
 	function url_to(string $controller, ...$args): string
 	{
-		if (empty($controller))
+		if (! $route = route_to($controller, ...$args))
 		{
-			return site_url();
+			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
 		}
 
 		if (strpos($controller, '::') !== false)
@@ -663,6 +665,6 @@ if (! function_exists('url_to'))
 			}
 		}
 
-		return site_url(route_to($controller, ...$args));
+		return site_url($route);
 	}
 }
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -664,7 +664,14 @@ if (! function_exists('url_to'))
 		{
 			$explode = explode('::', $controller);
 
-			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.controllerNotFound', [$explode[0], $explode[1]]));
+			if (isset($explode[1]))
+			{
+				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.controllerNotFound', [$explode[0], $explode[1]]));
+			}
+			else
+			{
+				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
+			}
 		}
 
 		return site_url($route);
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -648,6 +648,11 @@ if (! function_exists('url_to'))
 	 */
 	function url_to(string $controller, ...$args): string
 	{
+		if (empty($controller))
+		{
+			return site_url();
+		}
+
 		if (strpos($controller, '::') !== false)
 		{
 			$class = explode('::', $controller)[0];
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -655,6 +655,6 @@ if (! function_exists('url_to'))
 			$controller = service('routes')->getDefaultNamespace() . $controller;
 		}
 
-		return base_url(route_to($controller, ...$args));
+		return site_url(route_to($controller, ...$args));
 	}
 }
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -652,16 +652,7 @@ if (! function_exists('url_to'))
 	{
 		if (! $route = route_to($controller, ...$args))
 		{
-			$explode = explode('::', $controller);
-
-			if (isset($explode[1]))
-			{
-				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.controllerNotFound', [$explode[0], $explode[1]]));
-			}
-			else
-			{
-				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
-			}
+			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
 		}
 
 		return site_url($route);
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -652,6 +652,13 @@ if (! function_exists('url_to'))
 	{
 		if (! $route = route_to($controller, ...$args))
 		{
+			$explode = explode('::', $controller);
+
+			if (isset($explode[1]))
+			{
+				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.controllerNotFound', [$explode[0], $explode[1]]));
+			}
+
 			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
 		}
 
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -662,7 +662,9 @@ if (! function_exists('url_to'))
 
 		if (! $route = route_to($controller, ...$args))
 		{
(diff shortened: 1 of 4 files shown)


system/Helpers/url_helper.php
```diff
@@ -634,3 +634,27 @@ function mb_url_title(string $str, string $separator = '-', bool $lowercase = fa
...
+	function url_to(string $controller, ...$args): string
+	{
+		$class = explode('::', $controller)[0];
+
+		if (! class_exists($class))
+		{
+			$controller = service('routes')->getDefaultNamespace() . $controller;
+		}
+
+		return base_url(route_to($controller, ...$args));
```
MGatner, line 658: I *think* this needs to be `site_url` to account for subdirectories - can confirm on desktop later.

ImMaax: If I change `base_url` to `site_url`, it will prepend `index.php` (so `http://example.com/xyz` would become `http://example.com/index.php/xyz`). Otherwise, it's still fine, so changed that. Will commit soon.

ImMaax:
> This is awesome! What a great addition, thank you for the idea and PR. See my note on the URL function of choice. I also would like to see a few more tests, maybe a provider would be in order? Things like passing an empty string and `/`, and testing that names routes work as expected. If you can add those that would be great, I'm excited for this one - it will also be very handy with `FeatureResponse::getRedirectUrl()`.

So first off, thank you! For both the feedback and the appreciation. I changed `base_url` to `site_url` and added tests for empty strings as well as passing a slash. What exactly are providers and named routes, though? 😬

@MGatner

MGatner:
Thanks for that! So quick too. A data provider is just a way to specify a bunch of test criteria and results for a single test. The way you are doing it is fine, but you could also provide an array of 'route' => 'url' and run it through the same test. Search for `@dataProvider` amping the tests for some examples. 

Named routes are a way of providing an overriding name for a route: https://codeigniter4.github.io/CodeIgniter4/incoming/routing.html#using-named-routes

I think the Router call is handling this for you, it's just hard for me to tel on mobile.

ImMaax:
> Thanks for that! So quick too. A data provider is just a way to specify a bunch of test criteria and results for a single test. The way you are doing it is fine, but you could also provide an array of 'route' => 'url' and run it through the same test. Search for `@dataProvider` amping the tests for some examples.
> 
> Named routes are a way of providing an overriding name for a route: https://codeigniter4.github.io/CodeIgniter4/incoming/routing.html#using-named-routes
> 
> I think the Router call is handling this for you, it's just hard for me to tel on mobile.

TIL about named routes in CI and data providers in PHPUnit, thanks for telling me about them! `url_to` is now compatible with named routes and I reworked the tests for it a little. Hope this fits now :)

ImMaax:
> Welcome! This is looking great. I can't wait for your next PRs 🤗
> 
> I should have asked for this before, but how about some tests (and docs) for failure handling? Like what happens if I call `url_to('Foo::bar')` and `Foo` doesn't exist, or `bar` isn't a method?
> 
> Your data provider is great, and if you wanted you could expand it to all your tests. Make your toes have two values (class, URL) and your test take two parameters and then you can test them all in one go.

On failure, the function will just return the base URL of `site_url()`, so for example `http://example.com/index.php`. Should this be changed or is it fine as is? I'll also rework the tests once again.

system/Helpers/url_helper.php
MGatner: The `index.php` should appear or not depending on the user's app setting, which I think is good. It could be another test scenario though, to make sure `url_to()` is respecting `config('App')->indexPage`.

MGatner:
IMO a failure should not return a valid URL, because that would imply that there is an existing route for the query. I would return either `null` or even throw an exemption. Given that this should almost always be a developer's internal use I'd be in favor of throwing. Something like:
```
if (! $route = route_to($controller, ...$args))
{
    throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', $controller));
}

return site_url($route);
```

People can always catch it if they don't care, but I would want to know if I had typoed my controller name or something instead of providing links to base. Might be a good time for @michalsn or @samsonasik or someone else to give an opinion as well.

system/Helpers/url_helper.php
ImMaax: No matter what I try, doesn't seem to work, if I try to respect that indexPage setting

ImMaax:
> IMO a failure should not return a valid URL, because that would imply that there is an existing route for the query. I would return either `null` or even throw an exemption. Given that this should almost always be a developer's internal use I'd be in favor of throwing. Something like:
> 
> ```
> if (! $route = route_to($controller, ...$args))
> {
>     throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', $controller));
> }
> 
> return site_url($route);
> ```
> 
> People can always catch it if they don't care, but I would want to know if I had typoed my controller name or something instead of providing links to base. Might be a good time for @michalsn or @samsonasik or someone else to give an opinion as well.

Yeah, a failure should not do that. However, that's the default behavior of doing `site_url(route_to())`, so I thought it might fit in. But you're right. Changed it, exception is now thrown.

system/Helpers/url_helper.php
```diff
@@ -634,3 +634,37 @@ function mb_url_title(string $str, string $separator = '-', bool $lowercase = fa
...
+	 *
+	 * @throws \CodeIgniter\Router\Exceptions\RouterException
+	 *
+	 * @return string
+	 */
+	function url_to(string $controller, ...$args): string
+	{
+		if (! $route = route_to($controller, ...$args))
+		{
+			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
```
samsonasik, line 655: `HTTP.invalidRoute` seems for redirect response, probably use `HTTP.controllerNotFound` with uses explode `::` from controller and method variable definition.

ImMaax: Will change that, thank you

michalsn:
Personally, I would rather follow the behavior present in the `route_to()` function and return `false` on failure. Throwing an exception is fine for development mode but not really for production (we basically get a white page with error). It's a helper function - nobody will probably bother to catch en exception in a view or every controller (besides it wouldn't be very handy). But it's just my preference.

MGatner:
I'm fine with it either way. Generally I would like to see the framework throw more exceptions when things go wrong - throwing and catching is the best way to deal with problems and in cases like this amounts to mostly the same thing as gathering the result and then testing for `false`. I think you're right that helper functions are maybe the exception (haha) to this rule since they are designed for quick-and-easy access to existing class methods, but it still comes down to whether you want a "Whoops" screen or some malformed link on your production page if you've screwed up your code. Difference is, the Whoops will be logged and reported whereas the broken link will need to be noticed.

ImMaax:
> I'm fine with it either way. Generally I would like to see the framework throw more exceptions when things go wrong - throwing and catching is the best way to deal with problems and in cases like this amounts to mostly the same thing as gathering the result and then testing for `false`. I think you're right that helper functions are maybe the exception (haha) to this rule since they are designed for quick-and-easy access to existing class methods, but it still comes down to whether you want a "Whoops" screen or some malformed link on your production page if you've screwed up your code. Difference is, the Whoops will be logged and reported whereas the broken link will need to be noticed.

I personally feel like throwing is the better choice in that case and totally agree with you.

user_guide_src/source/helpers/url_helper.rst
```diff
@@ -360,3 +360,25 @@ The following functions are available:
     Pass the URL string to the function like this::
 
         $url = prep_url('example.com');
+
+.. php:function:: url_to([$controller[, ...$args]])
```
paulbalandan, line 364: Based on the function signature, `$controller` is not an optional argument, so probably you can remove the square brackets around it.

tests/system/Helpers/URLHelperTest.php
```diff
@@ -1250,4 +1250,64 @@ public function testBasedWithOtherIndex()
...
+	public function testUrlToThrowsOnEmptyOrMissingRoute(string $route)
+	{
+		$this->expectException(\CodeIgniter\Router\Exceptions\RouterException::class);
+
+		url_to($route);
+	}
+
+	public function urlToProvider()
+	{
+		if (config('App')->indexPage != '') {
```
paulbalandan, line 1279: The opening braces should be on its own line.

system/Helpers/url_helper.php
```diff
@@ -634,3 +634,37 @@ function mb_url_title(string $str, string $separator = '-', bool $lowercase = fa
...
+
+			if (! class_exists($class))
+			{
+				$controller = service('routes')->getDefaultNamespace() . $controller;
+			}
+		}
+
+		if (! $route = route_to($controller, ...$args))
+		{
+			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
```
paulbalandan, line 665: I think you have missed doing @samsonasik comment here.

user_guide_src/source/helpers/url_helper.rst
ImMaax: Oh yea, that was a mistake, fixing that!

tests/system/Helpers/URLHelperTest.php
ImMaax: Fixed it

michalsn:
I think we need some changes here. We have to make sure that this new helper will be able to support not only [reverse routing](https://codeigniter.com/user_guide/incoming/routing.html#reverse-routing) but also [named routes](https://codeigniter.com/user_guide/incoming/routing.html#using-named-routes). I'm 100% sure people will expect that because it's basically the same feature.

ImMaax:
> I think we need some changes here. We have to make sure that this new helper will be able to support not only [reverse routing](https://codeigniter.com/user_guide/incoming/routing.html#reverse-routing) but also [named routes](https://codeigniter.com/user_guide/incoming/routing.html#using-named-routes). I'm 100% sure people will expect that because it's basically the same feature.

Uhh, it does support named routes 😁

michalsn:
Looking at the code, I'm pretty sure it does not.

ImMaax:
> Looking at the code, I'm pretty sure it does not.

I'm pretty sure it does. Just checked it:

```php
service('routes')->add('/test', 'Test::index', ['as' => 'main']);
echo url_to('main'); # => http://localhost:8080/index.php/test
```

Am I missing something?

michalsn:
Yes, try something that is not defined.

ImMaax:
> Yes, try something that is not defined.

Yup, you're right, in case of failure, the string is expected to be a class::method notation, changed that.

system/Helpers/url_helper.php
```diff
@@ -634,3 +634,46 @@ function mb_url_title(string $str, string $separator = '-', bool $lowercase = fa
...
+	{
+		if (strpos($controller, '::') !== false)
+		{
+			$class = explode('::', $controller)[0];
+
+			if (! class_exists($class))
+			{
+				$controller = service('routes')->getDefaultNamespace() . $controller;
+			}
+		}
```
michalsn, line 661: We should run this code only if `route_to()` fail. In other case, it's just a waste of time.

MGatner: @ImMaax Looks like we're almost there! Could you move this around and run tests again?

ImMaax: @MGatner sure, sure. I'm currently busy, but will work on this later today!

system/Helpers/url_helper.php
```diff
@@ -634,3 +634,51 @@ function mb_url_title(string $str, string $separator = '-', bool $lowercase = fa
...
+				if (! class_exists($class))
+				{
+					$controller = service('routes')->getDefaultNamespace() . $controller;
+				}
+			}
+
+			if ($route = route_to($controller, ...$args))
+			{
+				return site_url($route);
+			}
```
michalsn, line 668: Almost there. These lines are unnecessary.

system/Helpers/url_helper.php
```diff
@@ -634,3 +634,51 @@ function mb_url_title(string $str, string $separator = '-', bool $lowercase = fa
...
+
+			$explode = explode('::', $controller);
+
+			if (isset($explode[1]))
+			{
+				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.controllerNotFound', [$explode[0], $explode[1]]));
+			}
+			else
```
samsonasik, line 676: no need else as already throw early.

system/Helpers/url_helper.php
ImMaax: @michalsn Sorry, but I don't quite get how they're unneccesary. What should it look like instead?

michalsn: We've already tried using `route_to()` on line 653 - so we know that will fail. Adding the default namespace will not change this.

ImMaax: So now, I've got this:

```php
	function url_to(string $controller, ...$args): string
	{
		if (! $route = route_to($controller, ...$args))
		{
			$explode = explode('::', $controller);

			if (isset($explode[1]))
			{
				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.controllerNotFound', [$explode[0], $explode[1]]));
			}
			else
			{
				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));
			}
		}

		return site_url($route);
	}
```

Works well with named routes and if you pass in a namespaced controller, but if I pass in a non-namespaced controller, it will throw an exception. That's why I added the default namespace. Am I still overlooking something?

michalsn: If we want to add the ability to omit the default namespace, this change should happen in the `route_to()` function. I just want to keep the consistency between functions. Of course, this is only my opinion.

ImMaax: Alright, this makes perfect sense and consistency is important to me too. However, I'd like to know what others think about this (@MGatner @samsonasik @paulbalandan just to mention who's been involved in this thread). I'm perfectly fine with both having it in there and not having it in there.

system/Helpers/url_helper.php
paulbalandan: Yes. Useless else here.

michalsn: @ImMaax It would be great if you could fix this one too, but if not then I will merge it tomorrow.

ImMaax: @michalsn Removed the if/else and explode :)

michalsn: @ImMaax We have a little misunderstanding :) Everything was good except this "else", so:
```php
			if (isset($explode[1]))
			{
				throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.controllerNotFound', [$explode[0], $explode[1]]));
			}

			throw new \CodeIgniter\Router\Exceptions\RouterException(lang('HTTP.invalidRoute', [$controller]));

```

ImMaax: Ohh, haha, alright, will change this real quick

michalsn:
@ImMaax Sorry it took so long. We appreciate your contribution!

ImMaax:
> @ImMaax Sorry it took so long. We appreciate your contribution!

No problem! Really glad I could help! Looking forward to more contributions! :)

PR #3413, opened 2020-07-27
Fixes Autoloader::loadLegacy() always called

samsonasik:
The `loadLegacy()` function was always called in `Autoloader::loadClass()` function for just normal `Home` Controller accessing `/` page, while there is no legacy class. /cc @paulbalandan if you can verify.

Tested in MacOS.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -280,14 +280,14 @@ class Autoloader
 	{
 		if (strpos($class, '\\') === false)
 		{
-			return false;
+			$class = 'Config\\' . $class;
 		}
 
 		foreach ($this->prefixes as $namespace => $directories)
 		{
 			foreach ($directories as $directory)
 			{
-				$directory = rtrim($directory, DIRECTORY_SEPARATOR);
+				$directory = rtrim($directory, '\\/');
 
 				if (strpos($class, $namespace) === 0)
 				{
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -280,7 +280,16 @@ class Autoloader
 	{
 		if (strpos($class, '\\') === false)
 		{
-			$class = 'Config\\' . $class;
+			$class    = 'Config\\' . $class;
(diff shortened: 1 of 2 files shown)


samsonasik:
seems still not working, I will check more.

samsonasik:
Ok, so, on debugging, got class `Modules` that fallback in `loadLegacy()`

samsonasik:
It seems it happen on calling `config()` function, eg: `config('Modules')`

samsonasik:
call `config('Modules)` -> loadLegacy() called.
call `config('Config\Modules)` -> loadLegacy() not called.

samsonasik:
I added `Config\\` prefix when class provided is not namespaced. If it still not found in the `Config\\` namespace, then find `loadLegacy()` as fallback. By this, both `config('Modules)` won't call `loadLegacy()` function.

samsonasik:
I've added unit test for it.

samsonasik:
updated with faster find assumption of Config class without looping.

system/Autoloader/Autoloader.php
```diff
@@ -280,6 +280,15 @@ protected function loadInNamespace(string $class)
 	{
 		if (strpos($class, '\\') === false)
 		{
+			$class    = 'Config\\' . $class;
+			$filePath = APPPATH . str_replace('\\', '/', $class) . '.php';
```
paulbalandan, line 284: Use `DIRECTORY_SEPARATOR` as replacement for pretty printing shutdown handler backtrace.

samsonasik: updated.

MGatner:
@samsonasik Are all non-namespaced calls from `config()` triggering `loadLegacy()`? That might merit reworking.

samsonasik:
travis build green 🎉

samsonasik:
@MGatner Yes, it was down there if no namespace specified. Probably, the `loadlegacy()` can be removed for `4.1.0` ?

Issue #3404, opened 2020-07-25
Encryption Class does not allow to change digest parameter from SHA512 to SHA256 or others


crustamet:
I was trying to encrypt a message with a lower security hmac.

With the SHA512 it just occupy so much space and i want to occupy less space for the message to be encrypted.

For a 6 char message i get a big encrypted message while this is to big for my database to handle.

I just want to use hmac SHA256 for this.

Is this the intended functionality for the encryption library ?
Maybe i`m using something that i don't need to use.

crustamet:
Or can we please have a variable that returns the hex instead of binary for the config encryption ?

michalsn:
Just add `public $digest = 'SHA256';` to your Encryption config file.

Hex for encryption config key - there is a PR underway that will solve this: https://github.com/codeigniter4/CodeIgniter4/pull/3397

PR #3416, opened 2020-07-27
Make encryption HMAC digest configurable

paulbalandan:
**Description**
Make `HMAC digest` configurable through `Config\Encryption` or through `.env` file. Also optimized some code in the Encryption library.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Encryption/Encryption.php
+++ b/system/Encryption/Encryption.php
@@ -103,7 +103,7 @@ class Encryption
 
 		$this->key    = $config->key;
 		$this->driver = $config->driver;
-		$this->digest = $config->digest;
+              $this->digest = $config->digest ?? 'SHA512';
 
 		// determine what is installed
 		$this->handlers = [
--- a/system/Encryption/Encryption.php
+++ b/system/Encryption/Encryption.php
@@ -99,12 +99,11 @@ class Encryption
 	 */
 	public function __construct(BaseConfig $config = null)
 	{
-		if (empty($config))
-		{
-			$config = new \Config\Encryption();
-		}
-		$this->driver = $config->driver;
+		$config = $config ?? new \Config\Encryption();
+
 		$this->key    = $config->key;
+		$this->driver = $config->driver;
+		$this->digest = $config->digest;
 
 		// determine what is installed
 		$this->handlers = [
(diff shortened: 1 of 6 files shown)


michalsn:
Well... we could add `cipher` as well, but... Personally, I have a little problem with that. Because every time we make these configurable things user-friendly... people without proper knowledge show up and make changes that they shouldn't be doing ;)

So... basically I think that "hiding" these config variables was intentional.

system/Encryption/Encryption.php
```diff
@@ -132,10 +131,11 @@ public function __construct(BaseConfig $config = null)
 	public function initialize(BaseConfig $config = null)
 	{
 		// override config?
-		if (! empty($config))
+		if ($config)
 		{
-			$this->driver = $config->driver;
 			$this->key    = $config->key;
+			$this->driver = $config->driver;
+			$this->digest = $config->digest;
```
samsonasik, line 138: probably add default value:

```suggestion
                      $this->digest = $config->digest ?? 'SHA512';
```

system/Encryption/Encryption.php
```diff
@@ -99,12 +99,11 @@ class Encryption
...
-		if (empty($config))
-		{
-			$config = new \Config\Encryption();
-		}
-		$this->driver = $config->driver;
+		$config = $config ?? new \Config\Encryption();
+
 		$this->key    = $config->key;
+		$this->driver = $config->driver;
+		$this->digest = $config->digest;
```
samsonasik, line 106: probably add default value:

```suggestion
              $this->digest = $config->digest ?? 'SHA512';
```

paulbalandan:
> Well... we could add `cipher` as well, but... Personally, I have a little problem with that. Because every time we make these configurable things user-friendly... people without proper knowledge show up and make changes that they shouldn't be doing ;)
> 
> So... basically I think that "hiding" these config variables was intentional.

I totally agree with your point here. It is quite annoying someone would mess the configs and blaming the framework for it as a bug. Adding `cipher` as well is a bit too much, I think. My concern here is like in #3404 where hosting providers don't provide databases that handle big encrypted messages nor modern charset handling.

paulbalandan:
@MGatner, yes this is ready.

PR #3419, opened 2020-07-29
Master

vocaloidat:
-en
due to
$routes->resource('api/photo');
$routes->presenter('admin/photos');
This will cause the photo.php as photos.php in the admin and api folders to fail to run.
The reason is because: the essential path resource, presenter. ‘/’, ‘\’. It’s the opposite. The former parameter needs to be ‘/’,
But the latter requires ‘\’.
So I added code to prevent this kind of error.

--cn
由于
$routes->resource('api/photo');
$routes->presenter('admin/photos');
会导致，在admin，api文件夹的photo.php as photos.php 无法运行。
原因是因为：本质的路径resource，presenter。 ‘/’，‘\’。是相反的，传入的参数前者需要‘/’，
但是后者则需要‘\’。
所以我增添了代码，防止出现这种错误。


MGatner:
This appears to be a duplicate. Closing in favor of #3421

PR #3423, opened 2020-07-29
add app and app/Config to autoload composer.json for app starter

samsonasik:
The starter app is mostly installed via composer create-project. I think it is better to use composer autoload for it.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/admin/starter/composer.json
+++ b/admin/starter/composer.json
@@ -13,6 +13,12 @@
 		"mikey179/vfsstream": "1.6.*",
 		"phpunit/phpunit": "^8.5"
 	},
+	"autoload": {
+		"psr-4": {
+			"App\\": "app",
+			"Config\\": "app/Config"
+		}
+	},
 	"autoload-dev": {
 		"psr-4": {
 			"Tests\\Support\\": "tests/_support"


paulbalandan:
I think we should exclude `App\` from here as we support custom namespace via `APP_NAMESPACE`.

samsonasik:
That change may less happen, also, composer autoload doesn't break codeigniter 4 autoloader, that will be a fallback namespace.

kenjis:
We cannot change `app` folder name. Because the composer's path overwrite the config in `Config\Autoload`.
https://forum.codeigniter.com/thread-81153.html

@samsonasik What's the benefit to use composer autoload?

samsonasik:
As far as i remember, with prioritize composer autoload when possible, it will got benefit for composer itself (eg: classmap on `dump-autoload -o`).

Probably in case of changing `app` directory, documentation need to be updated to change `composer.json` as well.

iRedds:
It's time to abandon the internal autoloader and use only the composer.

kenjis:
@samsonasik Thank you. I'm not sure that composer's autoloader has priority. I'll check.

kenjis:
@iRedds Do you mean the zip installation will be discontinued?

lonnieezell:
One core tenet of CI  has always been the ability to download a zip and not worry about anything. From my experience on the forums since v4 has been released this is still the case for a large chunk of users.

We must always remember the broader user-base and CodeIgniter's history. While we don't want to be completely beholden to old ways, we do need to pay attention to the users, and not alienate a broad group of them for no discernible gain.

paulbalandan:
This could have been avoided if we did not have the `autoload.psr4` section. I tried with `codeigniter4/codeigniter4` renaming **app/** to **myapp/**, and the only thing I need to change is the FCPATH in **spark** and **public/index.php**.

iRedds:
@kenjis Do you need a zip installation?
@lonnieezell The CI history says use an array for configuration files.
Do you have zip download data to say that this is the majority?
For example, Composer generated 37,735 installs in the last 30 days, of which 1,311 were made today alone.
https://packagist.org/packages/codeigniter4/framework/stats#major/4

kenjis:
@iRedds Personally I never use zip installation, but I don't say drop it soon.
CI is the last major framework that can be installed without Composer.
I think drop of the Zip installation should be carefully considered.

I see users using Zip who report GitHub issues sometime.

iRedds:
@kenjis Composer.phar and the installed framework can be added to the zip.

kenjis:
The path is from Composer, but 
it seems CI's autoloader loads `App\Controllers\Home`. CI 4.1.8 appstarter.

Screenshot 2022-01-29 18 33 58

kenjis:
No, @samsonasik is correct.
If I run `composer dump-autoload -o`, `'App\\Controllers\\Home'` is included in Composer classmap file,
and CI4 autoloader use it.

kenjis:
@iRedds 
> Composer.phar and the installed framework can be added to the zip.

I don't think it is good idea.
It forces users to use Composer without permission.

PR #3425, opened 2020-07-29
GeneratorCommand feature

paulbalandan:
**Description**
I know that CI4 will not host generator commands, but I believe it will be nice of us to provide a base class to help devs in creating their own generators. This base class will do the heavy lifting of verifying the class name input, getting the file path, checking if the file already exists and wants to overwrite it, and finally the actual writing of the generated class to the file.

For demonstration, I have converted our two existing generators, `CreateMigration` and `CreateSeeder`, to extend from this `GeneratorCommand` thus removing duplicated code. Also added tests for these two commands.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


lonnieezell:
I don't have time to do a full review at the moment - but I will say "never say never" :) 

Things might have changed over the last few years, and I fully expect code generators to be a thing. They're too much of a time saver and pragmatic option not to. I just don't have time at the moment for much of anything at the moment. Not sure if there's any code in the Libraries section of [Vulcan](https://github.com/lonnieezell/vulcan/) that you feel might be helpful here or not, but feel free to use and abuse anything you find there.

paulbalandan:
Ok. So let me take back my statement that "CI4 *will not* host generator commands". 😅

I have a question, though. Presently, `migrate:create`, `make:seeder`, and `session:migration` have this option `-n => Set namespace`. I don't know why it is even there or the logic why it is there. For me, setting up a namespace that is not defined will just mess up with PSR4 autoloading or with `FileLocator`'s searching. For this, I took the liberty of removing that option. Also, I think user generated files should be saved on `APPPATH`' subdirectories, not anywhere else.

lonnieezell:
The `-n` option shouldn't create a new namespace. It allows you to create those files within that namespace/folder structure. This can be handy if you use a lot of modules in your app that might not fall within the App namespace. While I didn't create those additions, I can see it being handy. I had several CI apps that I've built in the past where all of the core functionality was split into modules for different reasons at different times.

paulbalandan:
If I get it right, if I have these commands `php spark  migrate:create -n Myth\Auth` or `php spark migrate:create -n CodeIgniter`, I am expecting my generated files to be residing in the `Myth\Auth` subdirectory or `CodeIgniter` subdirectory.

lonnieezell:
> If I get it right, if I have these commands `php spark migrate:create -n Myth\Auth` or `php spark migrate:create -n CodeIgniter`, I am expecting my generated files to be residing in the `Myth\Auth` subdirectory or `CodeIgniter` subdirectory.

Yes, that's correct.

PR #3426, opened 2020-07-29
Generators commands

abilioposada:
Generators commands

**Description**
Create generators commands to scaffold files and avoid boilerplate

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/BaseCommand.php
+++ b/system/CLI/BaseCommand.php
@@ -150,7 +150,7 @@ abstract class BaseCommand
 	{
 		// The CommandRunner will grab the first element
 		// for the command name.
-		array_unshift($params, $command);
+		array_push($params, $command);
 
 		return $this->commands->run($command, $params);
 	}
--- a/system/Commands/Generators/CreateAll.php
+++ b/system/Commands/Generators/CreateAll.php
@@ -1,0 +1,113 @@
+<?php
+/**
+ * CodeIgniter
+ *
+ * An open source application development framework for PHP
+ *
+ * This content is released under the MIT License (MIT)
+ *
+ * Copyright (c) 2014-2019 British Columbia Institute of Technology
+ * Copyright (c) 2019-2020 CodeIgniter Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @package    CodeIgniter
+ * @author     CodeIgniter Dev Team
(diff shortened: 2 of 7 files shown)


system/Commands/Generators/CreateAll.php
```diff
@@ -0,0 +1,113 @@
...
+	 * @var String
+	 */
+	protected $name = 'make:all';
+
+	/**
+	 * The Command's short description
+	 *
+	 * @var String
+	 */
+	protected $description = 'Creates the scaffolding structure for an specific entity.';
```
MGatner, line 70:
```suggestion
	protected $description = 'Creates the scaffolding structure for a specific entity.';
```

system/CLI/BaseCommand.php
```diff
@@ -150,7 +150,7 @@ protected function call(string $command, array $params = [])
 	{
 		// The CommandRunner will grab the first element
 		// for the command name.
-		array_unshift($params, $command);
+		array_push($params, $command);
```
MGatner, line 153: This seems like a pretty major change with potential issues. Can you elaborate why?

system/Commands/Generators/CreateController.php
```diff
@@ -0,0 +1,168 @@
...
+
+		// Full path
+		$path = $homepath . 'Controllers/' . $name . '.php';
+
+		// Class name should be pascal case now (camel case with upper first letter)
+		$name = pascalize($name);
+
+		$template = <<<EOD
+		<?php namespace $ns\Controllers;
+
```
MGatner, line 144: Maybe add `use App\Controllers\BaseController` in case their namespace isn't the default, otherwise they will have to have a `BaseController` class in their namespace, which seems counter-intuitive.

system/Commands/Generators/CreateModel.php
```diff
@@ -0,0 +1,188 @@
...
+
+	protected $returnType         = "{entity}";
+	protected $useSoftDeletes     = true;
+
+	protected $allowedFields      = [];
+
+	protected $useTimestamps      = false;
+	protected $createdField       = "created_at";
+	protected $updatedField       = "updated_at";
+	protected $deletedField       = "deleted_at";
```
MGatner, line 165: These strings should use single quotes `'`, probably need to use `EOD` like you did above.

MGatner, line 165: These are the default fields and are probably unlikely to change for most models, or be changed in a `BaseModel` file. I recommend leaving them out, unless you have a compelling reason for including them.

system/Commands/Generators/CreateModel.php
```diff
@@ -0,0 +1,188 @@
...
+
+'use CodeIgniter\Model;
+
+class {name} extends Model
+{
+	protected $table              = "{table}";
+	protected $primaryKey         = "id";
+
+	protected $returnType         = "{entity}";
+	protected $useSoftDeletes     = true;
```
MGatner, line 158: `$useSoftDeletes` and `$useTimestamps` seem to be chosen arbitrarily here. These should either match their defaults in `Model` or have optional arguments to set them from CLI.

system/Commands/Generators/CreateModel.php
```diff
@@ -0,0 +1,188 @@
...
+	protected $skipValidation     = false;
+}
+';
+
+		$template = str_replace('{name}', $name, $template);
+		$template = str_replace('{table}', empty( $table ) ? '' : $table, $template);
+		$template = str_replace('{entity}', empty( $entity ) ? 'array' : "App\\Entities\\" . pascalize($entity), $template);
+
+		helper('filesystem');
+
```
MGatner, line 178: Since you support other namespaces I think you need the same `mkdir` when the path does not exists like you did for `Entities`.

michalsn:
I'm not sure if this class was available when you created this PR but now we have a new class `GeneratorCommand` which should be probably used instead of `BaseCommand`. It has many cool features.

abilioposada:
Like @michalsn said when I created this Pull Request there weren't `GeneratorCommand` merged yet, I was about to update it but nobody had reviewed mine. @MGatner I could do all of those changes and more in a gratefully way, but, what is going to happen if there's one more robust scaffolder commands #3480 mentioned?

system/CLI/BaseCommand.php
abilioposada: When I tried to call another commands on `CreateAll.php` like `$this->call('make:entity', [$name]);` then `array_unshift` put the command itself first, creating at the end an unexpected `make:entity.php` file.

I searched where it was previously used, and it was in `CreateMigration.php` in which an associative array was the parameter (No matter the order). So I just pushed the command at the end in case you would need it in a future.

system/Commands/Generators/CreateModel.php
abilioposada: Maybe I could escape them `\'` because with `EOD` it recognize the properties as variables and it breaks because are undefined

michalsn:
@abilioposada Thank you for your work on this one but I'm closing it in favor of #3449

Looking forward to your other contributions to the framework!

PR #3430, opened 2020-07-31
Clearing up confusion with TLS and SSL in Email Library, fixes #3429

### LINKED ISSUE #3429: Bug: SMTP to port 465 should use TLS from the start
**Describe the bug**
As per this thread on the forum: https://forum.codeigniter.com/thread-77192.html I'd say that the logic when connecting to the SMTP server is flawed. The `SMTPCrypto` setting complicates matters further since it controls the channel *or* the upgrade of the channel.
Connection to port 465 should use implicit TLS as per RFC8314 (i.e. the `SMTPCrypto` should be `"ssl"` but this is confusing as the server should require TLS). A STARTTLS command may be issued. However, some servers (Amazon SES for example) does not allow STARTTLS on port 465, but requires it on port 587.
Setting `$SMTPCrypto = 'ssl'` allows connecting to servers on port 587 but doesn't run the STARTTLS command so email submission will fail.

Furthermore, the PHP constant `STREAM_CRYPTO_METHOD_TLS_CLIENT` only enables TLS version 1.0 so `STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT` should be used in the call to stream_socket_enable_crypto if `$SMTPCrypto = 'tls'` is used. (Possibly `STREAM_CRYPTO_METHOD_ANY_CLIENT` may be used, but this would also enable SSL ciphers and I cannot say I understand the security implications of this).

I believe 

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
Email Library

When connecting to 465, the Email Library should always use TLS when connecting. `SMTPCrypto` should be set to blank.
When connecting to 587 `SMTPCrypto` should be set to `tls` to enable SMTP command STARTTLS to run.
The setting "ssl" doesn't make any sense but is required to get connection to 465 working.

**Context**
 - PHP version 7.3.20

tangix:
**Description**
Making sure connection for email submissions to port 465 is always using implicit TLS as per RFC8314.
Updating docs to describe the differences between port 465 and 587 and why/when STARTTLS is required and the setting `$SMTPCrypto`.
Current implementation only enables TLSv1.0 for the upgrade after STARTTLS, this PR enables versions 1.0, 1.1 and 1.2.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -1863,6 +1863,7 @@ class Email
 			return true;
 		}
 
+		$ssl = '';
 		if ($this->SMTPPort === 465)
 		{
 			$ssl = 'tls://';
@@ -1870,10 +1871,6 @@ class Email
 		elseif ($this->SMTPCrypto === 'ssl')
 		{
 			$ssl = 'ssl://';
-		}
-		else
-		{
-			$ssl = '';
 		}
 
 		$this->SMTPConnect = fsockopen(
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -1862,7 +1862,20 @@ class Email
 		{
 			return true;
 		}
-		$ssl               = ($this->SMTPCrypto === 'ssl') ? 'ssl://' : '';
+
+		if ($this->SMTPPort === 465)
+		{
+			$ssl = 'tls://';
+		}
+		elseif ($this->SMTPCrypto === 'ssl')
+		{
+			$ssl = 'ssl://';
+		}
+		else
+		{
+			$ssl = '';
+		}
+
 		$this->SMTPConnect = fsockopen(
 				$ssl . $this->SMTPHost, $this->SMTPPort, $errno, $errstr, $this->SMTPTimeout
 		);
@@ -1877,7 +1890,8 @@ class Email
 		{
 			$this->sendCommand('hello');
 			$this->sendCommand('starttls');
-			$crypto = stream_socket_enable_crypto($this->SMTPConnect, true, STREAM_CRYPTO_METHOD_TLS_CLIENT);
+			$crypto = stream_socket_enable_crypto($this->SMTPConnect, true, STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT |
+				STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT);
 			if ($crypto !== true)
 			{
 				$this->setErrorMessage(lang('Email.SMTPError', $this->getSMTPData()));
--- a/user_guide_src/source/libraries/email.rst
+++ b/user_guide_src/source/libraries/email.rst
@@ -87,6 +87,32 @@ The settings used for the last successful send are available from the
 instance property ``$archive``. This is helpful for testing and debugging
 to determine that actual values at the time of the ``send()`` call.
 
+SSL versus TLS for SMTP Protocol
+================================
+
+To protect the username, password and email content while communicating with the SMTP server,
+encryption on the channel should be used used. Two different standards are widely deployed and
+it is important to understand the differences when trying to troubleshoot email sending
+issues.
+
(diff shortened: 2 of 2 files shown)


user_guide_src/source/libraries/email.rst
```diff
@@ -87,6 +87,32 @@ The settings used for the last successful send are available from the
 instance property ``$archive``. This is helpful for testing and debugging
 to determine that actual values at the time of the ``send()`` call.
 
+SSL versus TLS for SMTP Protocol
+================================
+
+To protect the username, password and email content while communicating with the SMTP server,
+encryption on the channel should be used used. Two different standards are widely deployed and
```
paulbalandan, line 94: `used` was repeated.

user_guide_src/source/libraries/email.rst
```diff
@@ -87,6 +87,32 @@ The settings used for the last successful send are available from the
...
+it is important to understand the differences when trying to troubleshoot email sending
+issues.
+
+Most SMTP servers allow connections on ports 465 or 587 when submitting emails. (The
+original port 25 is seldom used because of many ISPs have blocking rules in place and
+since the communication is entirely in clear-text).
+
+The key difference is that port 465 expects the communication channel to be secured using TLS
+from the start as per RFC8314. A connection to port 587 allows clear-text connection and later
```
paulbalandan, line 103: probably add an external link to this rfc for cross reference.

tangix: Thanks!

system/Email/Email.php
```diff
@@ -1862,7 +1862,20 @@ protected function SMTPConnect()
...
+			$ssl = 'tls://';
+		}
+		elseif ($this->SMTPCrypto === 'ssl')
+		{
+			$ssl = 'ssl://';
+		}
+		else
+		{
+			$ssl = '';
+		}
```
michalsn, line 1877:
```suggestion
		$ssl = '';
		if ($this->SMTPPort === 465)
		{
			$ssl = 'tls://';
		}
		elseif ($this->SMTPCrypto === 'ssl')
		{
			$ssl = 'ssl://';
		}
```

system/Email/Email.php
```diff
@@ -1877,7 +1890,8 @@ protected function SMTPConnect()
 		{
 			$this->sendCommand('hello');
 			$this->sendCommand('starttls');
-			$crypto = stream_socket_enable_crypto($this->SMTPConnect, true, STREAM_CRYPTO_METHOD_TLS_CLIENT);
+			$crypto = stream_socket_enable_crypto($this->SMTPConnect, true, STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT |
+				STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT);
```
michalsn, line 1894: Would it make any difference if we used `STREAM_CRYPTO_METHOD_ANY_CLIENT` here instead?

tangix: That's an excellent question! `STREAM_CRYPTO_METHOD_ANY_CLIENT` would also enable the deprecated SSL 2.0 and 3.0 ciphers (https://tools.ietf.org/html/rfc7568) and I am not sure what the framework's policy on that is.
Deprecation of TLS 1.0 and 1.1 seems to be underway with the major browsers phasing out support during 2020 (https://www.entrustdatacard.com/blog/2018/november/deprecating-tls)
TLS 1.3 seems to be unsupported by PHP even though openssl supports it, but I haven't been able to understand the implicit support (i.e. is openssl trying to negotiate the highest possible version etc?)

michalsn: OK, thanks. We can leave it as it is.

michalsn:
Good changes! Thank you for working on it.

lonnieezell:
Very nice, thanks!

PR #3435, opened 2020-07-31
CLI Print Pretty

mostafakhudair:
Remove unless before/after newLine()
Fix Windows cmd colors
Remove useless methods

Here's some examples:
https://i.ibb.co/r3JnQPs/Annotation-2020-07-31-210349.png
https://i.ibb.co/SnwwdJV/Annotation-2020-07-31-213208.png

### DIFF
--- a/system/CLI/Commands.php
+++ b/system/CLI/Commands.php
@@ -41,6 +41,8 @@ namespace CodeIgniter\CLI;
 
 use CodeIgniter\Log\Logger;
 use Config\Services;
+use ReflectionClass;
+use ReflectionException;
 
 /**
  * Class Commands
@@ -72,7 +74,7 @@ class Commands
 	 */
 	public function __construct($logger = null)
 	{
-		$this->logger = $logger ?? service('logger');
+		$this->logger = $logger ?? Services::logger();
 	}
 
 	/**
@@ -163,7 +165,7 @@ class Commands
 
 			try
 			{
-				$class = new \ReflectionClass($className);
+				$class = new ReflectionClass($className);
 
 				if (! $class->isInstantiable() || ! $class->isSubclassOf(BaseCommand::class))
 				{
@@ -173,12 +175,13 @@ class Commands
 				$class = new $className($this->logger, $this);
 
 				// Store it!
-				if ($class->group !== null)
+				if (! is_null($class->group))
 				{
 					$this->commands[$class->name] = [
 						'class'       => $className,
 						'file'        => $file,
 						'group'       => $class->group,
+						'name'	      => $class->name,
 						'description' => $class->description,
 					];
 				}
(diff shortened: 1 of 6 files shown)


system/CLI/Console.php
```diff
@@ -93,14 +93,7 @@ public function run(bool $useSafeOutput = false)
 	 */
 	public function showHeader()
 	{
-		CLI::newLine();
-
-		CLI::write(CLI::color('CodeIgniter CLI Tool', 'green')
-				. ' - Version ' . CodeIgniter::CI_VERSION
-				. sprintf(' - Server Time: %s UTC%s', date('Y-m-d H:i:sa'), date('P')));
-
+		CLI::write('CodeIgniter v' . CodeIgniter::CI_VERSION . ' Command Line Tool - Server-Time: ' . date('Y-m-d H:i:s'), 'light_blue');
```
paulbalandan, line 96: Why remove the UTC offset? I just made a PR days ago to include that, though.

system/Commands/ListCommands.php
```diff
@@ -115,82 +107,43 @@ public function run(array $params)
...
-
-				$out = CLI::color($title, 'yellow');
-
-				if (isset($item['description']))
-				{
-					$out .= CLI::wrap($item['description'], 125, strlen($title));
-				}
+				CLI::write(
+					CLI::color($this->setPad($command['name'], $length, 2, 2), 'green')
+					 . $command['description']
```
paulbalandan, line 137: If we'll remove the wrap here then long descriptions would span across the terminal width and overflow.

system/CLI/Commands.php
```diff
@@ -173,20 +175,21 @@ public function discoverCommands()
...
 
 				// Store it!
-				if ($class->group !== null)
+				if (! is_null($class->group))
 				{
 					$this->commands[$class->name] = [
 						'class'       => $className,
 						'file'        => $file,
 						'group'       => $class->group,
+						'name'	      => $class->name,
```
paulbalandan, line 184: Is there a benefit of adding this line? As I see it, the command's name is already used as the array key and can be accessed later by foreach using `foreach ($commands as $name => $command)`, which is the current way.

system/Commands/ListCommands.php
```diff
@@ -73,7 +73,7 @@ class ListCommands extends BaseCommand
 	 *
 	 * @var string
 	 */
-	protected $description = 'Lists the available commands.';
+	protected $description = 'List of available commands.';
```
paulbalandan, line 76: I think the former is more apt as it describes the command's function, i.e. what it is doing.

michalsn:
Changes look good, but I would like to hear from @paulbalandan first. He made a lot of quality changes to the `CLI` lately, so he knows this class better than me right now :)

michalsn:
Thank you both!

PR #3438, opened 2020-08-01
add .gitattributes to admin/framework to allow --prefer-dist install only core framework

samsonasik:
With `.gitattributes`, when running `--prefer-dist` (mostly default implementation), we can get the something like this:

```
└─- vendor/codeigniter4
    ├── framework
         ├── license.txt
         ├── README.md
         ├── composer.json
         ├── license.txt
         ├── system
```

We can still get the full of framework structure in `vendor` directory with `--prefer-source`, documentation updated in `Upgrading` section in `Composer Installation` page.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated

### DIFF
--- a/admin/framework/.gitattributes
+++ b/admin/framework/.gitattributes
@@ -1,0 +1,8 @@
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/env export-ignore
+/phpunit.xml.dist export-ignore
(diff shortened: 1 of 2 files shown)


michalsn:
Good change. Thank you!

We have more choices and we can choose a more minimalist version if we want.

PR #3441, opened 2020-08-03
Enforce platform requirements

paulbalandan:
**Description**
Reading through the forums, many users, especially those manual installation users, are complaining undefined functions, such as `locale_set_default`, which can be resolved by installing the required PHP extensions. I think if we can warn them beforehand that their systems are missing these extensions and enforce them to install these then we'll have fewer problems along the road. I think this is much better than suggesting to comment out those undefined functions.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -189,7 +189,13 @@ class CodeIgniter
 		// Setup Exception Handling
 		Services::exceptions()->initialize();
 
-		$this->resolvePlatformExtensions();
+		// Run this check for manual installations
+		if (! is_file(COMPOSER_PATH))
+		{
+			// @codeCoverageIgnoreStart
+			$this->resolvePlatformExtensions();
+			// @codeCoverageIgnoreEnd
+		}
 
 		// Set default locale on the server
 		locale_set_default($this->config->defaultLocale ?? 'en');
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * CodeIgniter
  *
@@ -41,6 +42,7 @@ namespace CodeIgniter;
 use Closure;
 use CodeIgniter\Debug\Timer;
 use CodeIgniter\Events\Events;
+use CodeIgniter\Exceptions\FrameworkException;
 use CodeIgniter\Exceptions\PageNotFoundException;
 use CodeIgniter\HTTP\CLIRequest;
 use CodeIgniter\HTTP\DownloadResponse;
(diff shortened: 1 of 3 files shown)


lonnieezell:
I like what you're doing here, but wonder if we shouldn't hide that exception on production? We are bootstrapped at that point so we should be able to log it and give a generic error.

paulbalandan:
The resolving happens after the exceptions class is initialized so the thrown exception can be logged.

paulbalandan:
On production, it will just give the generic Whoops page and logs the error in the logs. Here is my logs test:
```
CRITICAL - 2020-08-03 11:24:58 --> The framework needs the following extension(s) installed and loaded: sodium.
#0 C:\Users\P\Desktop\Web Dev\CodeIgniter4\system\CodeIgniter.php(242): CodeIgniter\Exceptions\FrameworkException::forMissingExtension('sodium')
#1 C:\Users\P\Desktop\Web Dev\CodeIgniter4\system\CodeIgniter.php(192): CodeIgniter\CodeIgniter->resolvePlatformExtensions()
#2 C:\Users\P\Desktop\Web Dev\CodeIgniter4\system\bootstrap.php(181): CodeIgniter\CodeIgniter->initialize()
#3 C:\Users\P\Desktop\Web Dev\CodeIgniter4\public\index.php(36): require('C:\\Users\\P\\Desk...')
#4 {main}
```

lonnieezell:
Good point. Looks good, then!

MGatner:
I think that Composer handles this via `required` - is there any sense in restricting this call to manual installations? I don't know what kind of load the extension check represents but this would be added them to every page load.

paulbalandan:
For Composer installed applications this method is superfluous as the missing extensions are flagged early. But we also cater to manual users where system requirements can be bypassed or overlooked easily. If we can have a check to run this method only on manual installations then it may be better, but I cannot think of one. Also, I do not think this will cause considerable overhead even if this will be called on every page load.

michalsn:
I agree with @MGatner here. 

If we really want this feature to be included, maybe we should check for the `vendor` folder and make a check only if it's not there? I believe it's fair to assume that if someone uses a composer then CodeIgniter is installed that way as well.

lonnieezell:
These checks could also be added to the debug bar. When you click on the version number it brings up some system info, and was always intended to provide info about missing stuff on their setup. While it wouldn't solve all of the issues you're trying for here, it will provide helpful info for debugging. Maybe with an alert icon if things are missing? 

I like the idea presented about checking for an existing Composer installation before running the checks.

paulbalandan:
I have thought of checking for the `vendor` folder for Composer installations but suddenly it came to me that someone can easily hack the check by just putting a vendor folder.

michalsn:
Yes, `COMPOSER_PATH` sounds good to me.

paulbalandan:
Maybe adding the checks to the debug bar will be covered by another PR. :)

lonnieezell:
> Maybe adding the checks to the debug bar will be covered by another PR. :)

Agreed. Was going to comment the same. :) Merging.

Thanks!

PR #3446, opened 2020-08-04
Fix CLI option error in user guide

learning:
Fix CLI option error in user guide

**Description**
It should be a `<space>` instead of `=`

**Checklist:**
- [ ] Securely signed commits
- [X] User guide updated
- [ ] Conforms to style guide


lonnieezell:
Good catch, thanks. 

For future reference, if you are only making changes to documentation, you can add `[ci skip]` to the front of the PR name so it doesn't use TravisCI time when it doesn't need it.

PR #3447, opened 2020-08-04
Add support for SameSite cookies

### LINKED ISSUE #3442: Bug: Support for SameSite cookie setting missing
**Describe the bug**
Not possible to configure SameSite setting for cookies (session and CSRF).
The differences between PHP 7.2 and 7.3 makes this more difficult.
I now find myself in the situation that I need to be able to configure this setting and after migration to PHP 7.3 the hack with `cookiePath = '/;samesite=none';` no longer works.

I could possible do some work on this, but I'd like to know if there are anything planned before I start working on a PR.

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
Sessions + ?

It should be possible to configure SameSite settings for cookies.

tangix:
First version for review

**Description**
Fixes #3442 

Added support for setting SameSite on cookies. https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05
Allows setting the SameSite parameter for cookies and supports both PHP 7.2 and 7.3+ version.
Includes hacky support for PHP 7.2 for setting SameSite in the path - see SO https://stackoverflow.com/questions/39750906/php-setcookie-samesite-strict (for example)
Code contains version checks if less than 7.3.0 and the tests also differs.

I cannot get the unit tests to pass locally, so this is to check if travis is able to run them through.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


tangix:
Seems like session mock wasn't correct. Now all tests pass.

tangix:
I have done **some** initial testing in both 7.2 and 7.3 and see that Session is working and the cookie's SameSite value is set correctly. Also the CSRF cookie is set with the correct SameSite. Implications of setting SameSite on CSRF is beyond my understanding though.

system/HTTP/Response.php
```diff
@@ -247,6 +253,7 @@ public function __construct($config)
 		$this->cookiePath     = $config->cookiePath;
 		$this->cookieSecure   = $config->cookieSecure;
 		$this->cookieHTTPOnly = $config->cookieHTTPOnly;
+		$this->cookieSameSite = $config->cookieSameSite;
```
samsonasik, line 256: need default value fallback:

```suggestion
              $this->cookieSameSite = $config->cookieSameSite ?? '';
```

system/Security/Security.php
```diff
@@ -184,6 +191,7 @@ public function __construct($config)
 		$this->CSRFHeaderName = $config->CSRFHeaderName;
 		$this->CSRFCookieName = $config->CSRFCookieName;
 		$this->CSRFRegenerate = $config->CSRFRegenerate;
+		$this->CSRFSameSite   = $config->CSRFSameSite;
```
samsonasik, line 194: need default fallback as well:

```suggestion
              $this->CSRFSameSite   = $config->CSRFSameSite ?? 'Lax';
```

tangix: I am not sure I understand the reasoning behind this as no other config has this check but instead is set as default values in the classes?

samsonasik: the `public $CSRFSameSite` is a new config, so need a default value in case user do composer update without comparing new config in app/

tangix: Fair point.

app/Config/App.php
```diff
@@ -199,6 +199,13 @@ class App extends BaseConfig
 	| 'cookiePath'     = Typically will be a forward slash
 	| 'cookieSecure'   = Cookie will only be set if a secure HTTPS connection exists.
 	| 'cookieHTTPOnly' = Cookie will only be accessible via HTTP(S) (no javascript)
+	| 'cookieSameSite' = Configure cookie SameSite setting. Defaults to '' to not set
```
paulbalandan, line 202: Per RFC, if no samesite attribute is given or it is invalid, it shall default to `None`. Should we do that too here?

app/Controllers/Home.php
```diff
@@ -1,5 +1,7 @@
 <?php namespace App\Controllers;
 
+use CodeIgniter\Services;
```
paulbalandan, line 3: Unused `use` statement.

system/HTTP/Response.php
```diff
@@ -1036,10 +1060,33 @@ protected function sendCookies()
...
-			// PHP cannot unpack array with string keys
-			$params = array_values($params);
+			if (PHP_VERSION_ID < 70300)
+			{
+				// For PHP 7.2 we need to use the hacky method of setting SameSite in the path
+				if (isset($params['samesite']))
+				{
+					if (in_array(strtolower($params['samesite']), ['none', 'lax', 'strict']))
+					{
+						$params['path'] = '; samesite=' . $params['samesite'];
```
paulbalandan, line 1070: I think you mean here `.=`

tests/system/Session/SessionTest.php
```diff
@@ -542,4 +543,93 @@ public function testGetDotKey()
...
+		{
+			$cookies = $session->cookies;
+			$this->assertCount(1, $cookies);
+			$this->assertCount(3, $cookies[0]);
+			$this->assertIsArray($cookies[0][2]);
+			$this->assertArrayNotHasKey('samesite', $cookies[0][2]);
+		}
+	}
+
+	public function testInvalidSameSite()
```
paulbalandan, line 613: Should default to `None`?

system/HTTP/Response.php
tangix: Correct! Will add a better test for this as well.

app/Config/App.php
paulbalandan: OK. I think I have misread things.

For the invalid samesite, Sec. 4.1.2.7 says `If the "SameSite" attribute's value is something other than these three known keywords, the attribute's value will be treated as "None".`

tangix: Maybe setting CSRFSameSite also to '' would be a good idea? I don't know when browsers will start enforcing SameSite on cookies.

paulbalandan: Per my reading the RFC, `Lax` is the suggested value for CSRF.

tangix: Yeah, so 'Lax' should be standard for CSRF it seems.

system/HTTP/Response.php
```diff
@@ -241,12 +247,13 @@ public function __construct($config)
...
-		$this->cookiePath     = $config->cookiePath;
-		$this->cookieSecure   = $config->cookieSecure;
-		$this->cookieHTTPOnly = $config->cookieHTTPOnly;
+		$this->CSPEnabled           = $config->CSPEnabled;
+		$this->cookiePrefix         = $config->cookiePrefix;
+		$this->cookieDomain         = $config->cookieDomain;
+		$this->cookiePath           = $config->cookiePath;
+		$this->cookieSecure         = $config->cookieSecure;
+		$this->cookieHTTPOnly       = $config->cookieHTTPOnly;
+			  $this->cookieSameSite = $config->cookieSameSite ?? '';
```
michalsn, line 256: Please align with the lines above.

system/HTTP/Response.php
```diff
@@ -888,6 +902,16 @@ public function setCookie(
 			'httponly' => $httponly,
 		];
 
+		if ($samesite !== '')
+		{
+			if (in_array(strtolower($samesite), ['none', 'lax', 'strict']))
+			{
+				$cookie['samesite'] = $samesite;
+			}
+		}
```
michalsn, line 911:
```suggestion
		if ($samesite !== '' && in_array(strtolower($samesite), ['none', 'lax', 'strict']))
		{
			$cookie['samesite'] = $samesite;
		}
```

system/Security/Security.php
```diff
@@ -283,9 +291,48 @@ public function CSRFSetCookie(RequestInterface $request)
...
+			if (in_array(strtolower($this->CSRFSameSite), ['none', 'lax', 'strict']))
+			{
+				$params['samesite'] = $this->CSRFSameSite;
+			}
+
+			setcookie(
+				$this->CSRFCookieName,
+				$this->CSRFHash,
+				$params
+					);
```
michalsn, line 334: Please align.

system/Session/Session.php
```diff
@@ -1014,9 +1057,48 @@ protected function startSession()
...
+			if (in_array(strtolower($this->cookieSameSite), ['none', 'lax', 'strict']))
+			{
+				$params['samesite'] = $this->cookieSameSite;
+			}
+
+			setcookie(
+				$this->sessionCookieName,
+				session_id(),
+				$params
+					);
```
michalsn, line 1100: Please align.

system/HTTP/Response.php
```diff
@@ -1036,10 +1060,33 @@ protected function sendCookies()
...
+			{
+				// For PHP 7.2 we need to use the hacky method of setting SameSite in the path
+				if (isset($params['samesite']))
+				{
+					if (in_array(strtolower($params['samesite']), ['none', 'lax', 'strict']))
+					{
+						$params['path'] .= '; samesite=' . $params['samesite'];
+					}
+					unset($params['samesite']);
+				}
```
michalsn, line 1073:
```suggestion
				if (isset($params['samesite']) && in_array(strtolower($params['samesite']), ['none', 'lax', 'strict']))
				{
					$params['path'] .= '; samesite=' . $params['samesite'];
					unset($params['samesite']);
				}
```

system/HTTP/Response.php
```diff
@@ -1036,10 +1060,33 @@ protected function sendCookies()
...
+				setcookie(...$params);
+			}
+			else
+			{
+				// PHP 7.3 and later have a signature for setcookie() with options array as third argument
+				// and SameSite is possible to set there
+				$name  = $params['name'];
+				$value = $params['value'];
+				unset($params['name']);
+				unset($params['value']);
```
michalsn, line 1086:
```suggestion
				unset($params['name'], $params['value']);
```

tangix:
@michalsn 
Lax is the new default, makes sense. PHP 7.2 will still be supported for a couple of months....

system/HTTP/Response.php
```diff
@@ -247,6 +253,7 @@ public function __construct($config)
 		$this->cookiePath     = $config->cookiePath;
 		$this->cookieSecure   = $config->cookieSecure;
 		$this->cookieHTTPOnly = $config->cookieHTTPOnly;
+		$this->cookieSameSite = $config->cookieSameSite ?? 'Lax';
```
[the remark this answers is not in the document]
michalsn: I think we should also add some check here:
```php
if (! in_array(strtolower($this->cookieSameSite), ['none', 'lax', 'strict'])
{
// throw an exception
}
```

system/HTTP/Response.php
```diff
@@ -888,6 +902,13 @@ public function setCookie(
 			'httponly' => $httponly,
 		];
 
+		if ($samesite !== '' && in_array(strtolower($samesite), ['none', 'lax', 'strict']))
+		{
+			$cookie['samesite'] = $samesite;
+		}
+
```
michalsn, line 909: `samesite` should be always added.
```suggestion
```

system/HTTP/Response.php
```diff
@@ -1036,10 +1057,29 @@ protected function sendCookies()
 
 		foreach ($this->cookies as $params)
 		{
-			// PHP cannot unpack array with string keys
-			$params = array_values($params);
+			if (PHP_VERSION_ID < 70300)
+			{
+				// For PHP 7.2 we need to use the hacky method of setting SameSite in the path
+				if (isset($params['samesite']) && in_array(strtolower($params['samesite']), ['none', 'lax', 'strict']))
```
michalsn, line 1063: I think we can manage without this check as we already handle it through the constructor.

system/Security/Security.php
```diff
@@ -179,11 +186,12 @@ class Security
...
-		$this->CSRFTokenName  = $config->CSRFTokenName;
-		$this->CSRFHeaderName = $config->CSRFHeaderName;
-		$this->CSRFCookieName = $config->CSRFCookieName;
-		$this->CSRFRegenerate = $config->CSRFRegenerate;
+		$this->CSRFExpire         = $config->CSRFExpire;
+		$this->CSRFTokenName      = $config->CSRFTokenName;
+		$this->CSRFHeaderName     = $config->CSRFHeaderName;
+		$this->CSRFCookieName     = $config->CSRFCookieName;
+		$this->CSRFRegenerate     = $config->CSRFRegenerate;
+			  $this->CSRFSameSite = $config->CSRFSameSite ?? 'Lax';
```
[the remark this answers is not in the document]
michalsn: We should probably check if `$this->CSRFSameSite` has allowed value - just like in `Response::__constuct()`

system/Session/Session.php
```diff
@@ -303,9 +312,43 @@ protected function configure()
...
 
-		session_set_cookie_params(
-				$this->sessionExpiration, $this->cookiePath, $this->cookieDomain, $this->cookieSecure, true // HTTP only; Yes, this is intentional and not configurable for security reasons.
-		);
+		if (PHP_VERSION_ID < 70300)
+		{
+			// In PHP < 7.3.0, there is a "hacky" way to set the SameSite parameter
+
+			$sameSite = '';
+			if (in_array(strtolower($this->cookieSameSite), ['none', 'lax', 'strict']))
```
michalsn, line 320: This check won't be needed anymore since we have `__constructor()`

system/Session/Session.php
```diff
@@ -303,9 +312,43 @@ protected function configure()
...
+			// PHP 7.3 adds another function signature allowing setting of samesite
+			$params = [
+				'lifetime' => $this->sessionExpiration,
+				'path'     => $this->cookiePath,
+				'domain'   => $this->cookieDomain,
+				'secure'   => $this->cookieSecure,
+				'httponly' => true, // HTTP only; Yes, this is intentional and not configurable for security reasons.
+			];
+
+			if (in_array(strtolower($this->cookieSameSite), ['none', 'lax', 'strict']))
```
michalsn, line 344: This check won't be needed anymore since we have `__constructor()`

system/Session/Session.php
```diff
@@ -1014,9 +1057,48 @@ protected function startSession()
...
 	{
-		setcookie(
-				$this->sessionCookieName, session_id(), (empty($this->sessionExpiration) ? 0 : time() + $this->sessionExpiration), $this->cookiePath, $this->cookieDomain, $this->cookieSecure, true
-		);
+		if (PHP_VERSION_ID < 70300)
+		{
+			// In PHP < 7.3.0, there is a "hacky" way to set the samesite parameter
+
+			$sameSite = '';
+			if (in_array(strtolower($this->cookieSameSite), ['none', 'lax', 'strict']))
```
michalsn, line 1065: This check won't be needed anymore since we have `__constructor()`

system/Session/Session.php
```diff
@@ -1014,9 +1057,48 @@ protected function startSession()
...
+			// PHP 7.3 adds another function signature allowing setting of samesite
+			$params = [
+				'expires'  => $this->sessionExpiration,
+				'path'     => $this->cookiePath,
+				'domain'   => $this->cookieDomain,
+				'secure'   => $this->cookieSecure,
+				'httponly' => true,
+			];
+
+			if (in_array(strtolower($this->cookieSameSite), ['none', 'lax', 'strict']))
```
michalsn, line 1091: This check won't be needed anymore since we have `__constructor()`

system/Test/Mock/MockAppConfig.php
```diff
@@ -11,6 +11,7 @@ class MockAppConfig
 	public $cookiePath     = '/';
 	public $cookieSecure   = false;
 	public $cookieHTTPOnly = false;
+	public $cookieSameSite = '';
```
michalsn, line 14:
```suggestion
	public $cookieSameSite = 'Lax';
```

system/Test/Mock/MockCLIConfig.php
```diff
@@ -11,6 +11,7 @@ class MockCLIConfig extends \Config\App
 	public $cookiePath     = '/';
 	public $cookieSecure   = false;
 	public $cookieHTTPOnly = false;
+	public $cookieSameSite = '';
```
michalsn, line 14: Just for consistency
```suggestion
	public $cookieSameSite = 'Lax';
```

system/Test/Mock/MockSession.php
```diff
@@ -49,15 +50,48 @@ protected function startSession()
...
-			$this->cookieDomain,
-			$this->cookieSecure,
-			true,
-		];
+		if (PHP_VERSION_ID < 70300)
+		{
+			// In PHP < 7.3.0, there is a "hacky" way to set the samesite parameter
+
+			$sameSite = '';
+			if (in_array(strtolower($this->cookieSameSite), ['none', 'lax', 'strict']))
```
michalsn, line 58: This check won't be needed anymore since we have `__constructor()`

system/Test/Mock/MockSession.php
```diff
@@ -49,15 +50,48 @@ protected function startSession()
...
+			// PHP 7.3 adds another function signature allowing setting of samesite
+			$params = [
+				'lifetime' => $this->sessionExpiration,
+				'path'     => $this->cookiePath,
+				'domain'   => $this->cookieDomain,
+				'secure'   => $this->cookieSecure,
+				'httponly' => true,
+			];
+
+			if (in_array(strtolower($this->cookieSameSite), ['none', 'lax', 'strict']))
```
michalsn, line 84: This check won't be needed anymore since we have `__constructor()`

tangix:
@michalsn 
I have also spent the last days thinking about this. Since I am not 100% OK with having a feature added without any way to turn it off, I initially defaulted to `$cookieSameSite = '';` to indicate that the SameSite attribute should not be set at all. After som back and forth with @paulbalandan we decided with `'Lax'` as the best default value. This makes my initial idea with `''` as way to avoid setting the attribute at all more complicated to implement.

I wonder if it would be ok to allow explicitly setting `$cookieSameSite =null;` (or possibly `false`) to completely disable setting the SameSite attribute in case this is required for some reason?

system/HTTP/Response.php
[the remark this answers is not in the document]
tangix: The point here made by @samsonasik (https://github.com/codeigniter4/CodeIgniter4/pull/3447#discussion_r465781515) is that since this is a new setting, we need to handle that gracefully when updating using composer. Or adding a note to docs and have a BC discussion ;-)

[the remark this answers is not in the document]
paulbalandan: Per RFC, if the value given to samesite does not fall within the three values, it must be treated as `None`.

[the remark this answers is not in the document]
tangix: But that's how the browser should handle an illegal attribute. If it is missing, the browser should do... what? I still think CI should be able to not set the attribute at all.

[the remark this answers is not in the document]
michalsn: > The point here made by @samsonasik (#3447 (comment)) is that since this is a new setting, we need to handle that gracefully when updating using composer.

I believe he was referring to the variable from the config file - and that's true. Here we go with `this->cookieSameSite` as a fallback which is the variable from this class - so we can be sure it's there.

> Per RFC, if the value given to samesite does not fall within the three values, it must be treated as `None`.

Very good point - I missed this note when I was reading RFC yesterday. I guess we should do the same... however I like the idea with  "exception" - it would let developers avoid some simple mistakes.

> But that's how the browser should handle an illegal attribute. If it is missing, the browser should do... what? I still think CI should be able to not set the attribute at all.

All modern browsers have some default value for it. So if it's not set - the browser will add it. IMO there is no point for implementing this feature. This could also potentially leads to certain problems - when a cookie is set in one browser and not in the other.

michalsn:
@tangix Honestly, I see no point giving the ability to not set `SameSite` policy. All modern browsers go with some default value for it, if not `Lax` then other. Also if `SameSite` policy is not specified explicitly then we got warnings.

We should promote building a secure code and from what I see (looking at how browsers support it), it's already a "new standard".

In general, I would like to go with the default value `Lax` for every cookie and developers can change it if they want.

tangix:
> In general, I would like to go with the default value `Lax` for every cookie and developers can change it if they want.

Lax it is then. I will update the PR with this, add a throw for bad setting and fix the tests.

system/HTTP/Response.php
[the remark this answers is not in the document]
tangix: Added exception

tangix:
Based on https://github.com/codeigniter4/CodeIgniter4/issues/3442#issuecomment-671283004 :
May we should let this functionally be handled by core PHP and skip support for 7.2 (that is EOL in 4 months)? I've put this RP as Draft again so it won't get merged by accident...

michalsn:
Let's hope that next week someone else will have enough time to look at this PR from a different perspective and give us feedback.

michalsn:
@tangix Seems like you were right about the option to not set the `SameSite` at all. I missed that PHP docs state that. I think it should be handled as you proposed earlier - so that `null` value means `SameSite` shouldn't be set.

Personally I still think it has not much sense and may cause problems in some cases but since this is the supported option in the native function then I guess we have no choice but to follow :)

tangix:
> @tangix Seems like you were right about the option to not set the `SameSite` at all. I missed that PHP docs state that. I think it should be handled as you proposed earlier - so that `null` value means `SameSite` shouldn't be set.

`null` is not be the easiest setting to deal with throughout the code. To keep consistency with the other settings, an empty string `''` may be OK?

michalsn:
I'm not sure if `null` can cause any problems in implementation but you convinced me with a statement about consistency. An empty string seems to be the right choice 👍

system/HTTP/Response.php
```diff
@@ -246,6 +253,12 @@ public function __construct($config)
 		$this->cookiePath     = $config->cookiePath;
 		$this->cookieSecure   = $config->cookieSecure;
 		$this->cookieHTTPOnly = $config->cookieHTTPOnly;
+		$this->cookieSameSite = $config->cookieSameSite ?? $this->cookieSameSite;
+
+		if (! in_array(strtolower($this->cookieSameSite), ['', 'none', 'lax', 'strict']))
```
michalsn, line 258:
```suggestion
		if (! in_array(strtolower($this->cookieSameSite), ['', 'none', 'lax', 'strict'], true))
```

system/HTTP/Response.php
```diff
@@ -868,6 +883,16 @@ public function setCookie(
 			$httponly = $this->cookieHTTPOnly;
 		}
 
+		if (is_null($samesite))
+		{
+			$samesite = $this->cookieSameSite ?? '';
+		}
+
+		if (! in_array(strtolower($samesite), ['', 'none', 'lax', 'strict']))
```
michalsn, line 891:
```suggestion
		if (! in_array(strtolower($samesite), ['', 'none', 'lax', 'strict'], true))
```

system/Security/Security.php
```diff
@@ -184,12 +191,18 @@ public function __construct($config)
...
 		$this->CSRFCookieName = $config->CSRFCookieName;
 		$this->CSRFRegenerate = $config->CSRFRegenerate;
+		$this->CSRFSameSite   = $config->CSRFSameSite ?? $this->CSRFSameSite;
 
 		if (isset($config->cookiePrefix))
 		{
 			$this->CSRFCookieName = $config->cookiePrefix . $this->CSRFCookieName;
 		}
 
+		if (! in_array(strtolower($this->CSRFSameSite), ['', 'none', 'lax', 'strict']))
```
michalsn, line 201:
```suggestion
		if (! in_array(strtolower($this->CSRFSameSite), ['', 'none', 'lax', 'strict'], true))
```

system/Session/Session.php
```diff
@@ -181,9 +190,15 @@ public function __construct(\SessionHandlerInterface $driver, $config)
...
 
-		$this->cookieDomain = $config->cookieDomain;
-		$this->cookiePath   = $config->cookiePath;
-		$this->cookieSecure = $config->cookieSecure;
+		$this->cookieDomain   = $config->cookieDomain;
+		$this->cookiePath     = $config->cookiePath;
+		$this->cookieSecure   = $config->cookieSecure;
+		$this->cookieSameSite = $config->cookieSameSite;
+
+		if (! in_array(strtolower($this->cookieSameSite), ['', 'none', 'lax', 'strict']))
```
michalsn, line 198:
```suggestion
		if (! in_array(strtolower($this->cookieSameSite), ['', 'none', 'lax', 'strict'], true))
```

system/HTTP/Response.php
```diff
@@ -1035,10 +1067,35 @@ protected function sendCookies()
 
 		foreach ($this->cookies as $params)
 		{
-			// PHP cannot unpack array with string keys
-			$params = array_values($params);
+			if (PHP_VERSION_ID < 70300)
+			{
+				// For PHP 7.2 we need to use the hacky method of setting SameSite in the path
+				if (isset($params['samesite']) && in_array(strtolower($params['samesite']), ['none', 'lax', 'strict']))
```
michalsn, line 1073:
```suggestion
				if (isset($params['samesite']) && in_array(strtolower($params['samesite']), ['none', 'lax', 'strict'], true))
```

system/HTTP/Response.php
```diff
@@ -1035,10 +1067,35 @@ protected function sendCookies()
...
+			else
+			{
+				// PHP 7.3 and later have a signature for setcookie() with options array as third argument
+				// and SameSite is possible to set there
+				$name  = $params['name'];
+				$value = $params['value'];
+				unset($params['name'], $params['value']);
+
+				// If samesite is blank string, skip setting the attribute on the cookie
+				if (isset($params['samesite']) && ($params['samesite'] === ''))
```
michalsn, line 1092:
```suggestion
				if (isset($params['samesite']) && $params['samesite'] === '')
```

system/Language/en/HTTP.php
```diff
@@ -74,4 +74,7 @@
    'uploadErrNoTmpDir'          => 'File could not be uploaded: missing temporary directory.',
    'uploadErrExtension'         => 'File upload was stopped by a PHP extension.',
    'uploadErrUnknown'           => 'The file "%s" was not uploaded due to an unknown error.',
+
+	// SameSite setting
+   'invalidSameSiteSetting'     => 'SameSite setting must None, Lax, Strict or blank string. Given: {0}',
```
michalsn, line 79:
```suggestion
   'invalidSameSiteSetting'     => 'The SameSite setting must be None, Lax, Strict, or a blank string. Given: {0}',
```

system/Language/en/Session.php
```diff
@@ -20,4 +20,5 @@
    'writeProtectedSavePath' => 'Session: Configured save path "{0}" is not writable by the PHP process.',
    'emptySavePath'          => 'Session: No save path configured.',
    'invalidSavePathFormat'  => 'Session: Invalid Redis save path format: {0}',
+   'invalidSameSiteSetting' => 'SameSite setting must None, Lax, Strict or blank string. Given: {0}',
```
michalsn, line 23:
```suggestion
   'invalidSameSiteSetting' => 'Session: The SameSite setting must be None, Lax, Strict, or a blank string. Given: {0}',
```

michalsn (changes requested):
Just a few small things and there is some syntax issue in `response.rst` file. 

We will need to mention these changes in the `changelogs/v4.0.5.rst` too.
Other than that, I think this is quite ready 👍

system/Session/Session.php
```diff
@@ -181,9 +190,15 @@ public function __construct(\SessionHandlerInterface $driver, $config)
 		$this->sessionTimeToUpdate      = $config->sessionTimeToUpdate;
 		$this->sessionRegenerateDestroy = $config->sessionRegenerateDestroy;
 
-		$this->cookieDomain = $config->cookieDomain;
-		$this->cookiePath   = $config->cookiePath;
-		$this->cookieSecure = $config->cookieSecure;
+		$this->cookieDomain   = $config->cookieDomain;
+		$this->cookiePath     = $config->cookiePath;
+		$this->cookieSecure   = $config->cookieSecure;
+		$this->cookieSameSite = $config->cookieSameSite;
```
samsonasik, line 196: need default fallback value:

```suggestion
              $this->cookieSameSite = $config->cookieSameSite ?? $this->cookieSameSite;
```

system/HTTP/Response.php
```diff
@@ -817,6 +830,7 @@ public function redirect(string $uri, string $method = 'auto', int $code = null)
 	 * @param string       $prefix   Cookie name prefix
 	 * @param boolean      $secure   Whether to only transfer cookies via SSL
 	 * @param boolean      $httponly Whether only make the cookie accessible via HTTP (no javascript)
+	 * @param null         $samesite
```
samsonasik, line 833: need include `string` in `@param`:

```suggestion
       * @param string|null  $samesite
```

system/Helpers/cookie_helper.php
```diff
@@ -62,16 +62,27 @@
 	 * @param boolean      $secure   True makes the cookie secure
 	 * @param boolean      $httpOnly True makes the cookie accessible via
 	 *                                 http(s) only (no javascript)
+	 * @param string       $sameSite The cookie SameSite value
```
samsonasik, line 65:
```suggestion
       * @param string|null  $sameSite The cookie SameSite value
```

michalsn:
@tangix Thank you for taking your time and working on this! I think this is a really important feature. Thanks!

tangix:
> @tangix Thank you for taking your time and working on this! I think this is a really important feature. Thanks!

You are welcome! CodeIgniter is an integral part of my business and I spend all of my PHP-development in either 3.0 or 4.0. I absolutely hate having to tinker with the core-files so giving back to the community what I fix gives a warm cozy feeling.

PR #3449, opened 2020-08-05
Move built-in generators to Generators folder

paulbalandan:
**Description**
Move generator commands to `Generators` folder for a bit of consistency.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


paulbalandan:
I'm not sure what happened but changing the namespace in the `view()` made phpunit spew out the contents of the view?
```
Runtime:       PHP 7.2.15 with Xdebug 2.6.1
Configuration: /home/travis/build/codeigniter4/CodeIgniter4/phpunit.xml.dist
<@php
namespace {namespace};
class {class} extends Migration
{
	public function up()
	{
		$this->forge->addField([
			'id'         => [
				'type'       => 'VARCHAR',
				'constraint' => 128,
				'null'       => false,
			],
			'ip_address' => [
				'type'       => 'VARCHAR',
				'constraint' => 45,
				'null'       => false,
			],
			'timestamp'  => [
				'type'       => 'INT',
				'constraint' => 10,
				'unsigned'   => true,
				'null'       => false,
				'default'    => 0,
			],
			'data'       => [
				'type'       => 'TEXT',
				'null'       => false,
				'default'    => '',
			],
		]);
		$this->forge->addKey('id', true);
		$this->forge->addKey('timestamp');
		$this->forge->createTable('', true);
	}
	public function down()
	{
		$this->forge->dropTable('', true);
	}
}
.............................................................   61 / 3385 (  1%)
```

michalsn:
Well, that's strange... maybe tomorrow I will be able to dig into it more.

paulbalandan:
OK. I have found that PHPUnit will `include` all PHP files it will find as defined in the path in `phpunit.xml`. To circumvent this, make all views end with `.tpl`. This way it will not be included and its contents will be spewed out to the terminal.

paulbalandan:
Another Nirvana. Just found out that we can actually exclude files and directories in the XML file. So, no need to worry about file extensions.

PR #3384, opened 2020-07-23
Allow Common Datatypes for Table creation between Drivers

devorama:
**Description**
Allow common datatypes between Drivers for table creation and alters.
This allows Postgress to have ENUM,SET types (will create the type inside it for use in table).
Test cases where updated now Testing of datatypes is not driver dependant on creation of tables anymore

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

PR #3457, opened 2020-08-06
Make PostgreSQL DBDriver consistent : using Postgre instead of postgre

samsonasik:
In the documentation, it is named "Postgre", while in the code, it named "postgre" lower case. This change ensure to make it consistent to use "Postgre" naming in PostgreSQL Connection's DBDriver property. 

Postgre

Also, on **non case sensitive** operating system, it caused an issue when we can configure `.env` as lower "postgre" and when moving to **case sensitive** operating system, it got error class not found.

This patch fixed it.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1493,7 +1493,8 @@ abstract class BaseConnection implements ConnectionInterface
 	 */
 	public function callFunction(string $functionName, ...$params): bool
 	{
-		$driver = ($this->DBDriver === 'postgre' ? 'pg' : strtolower($this->DBDriver)) . '_';
+		$driver = strtolower($this->DBDriver);
+		$driver = ($driver === 'postgre' ? 'pg' : $driver) . '_';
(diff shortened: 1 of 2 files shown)


samsonasik:
travis build green 🎉

devorama:
Hi there.
Note this might make that testing for Postgre is ignored if you check the files
tests\_support\Database\Seeds\CITestSeeder.php
prent

and

tests\_support\Database\Seeds\20160428212500_Create_test_tables.php
prent

Ticket https://github.com/codeigniter4/CodeIgniter4/pull/3384 will help minimize the dependincy of  this checks

samsonasik:
No, I made a change the value to "Postgre" so it still works in the tests, the `strtolower()` check only in assignment to local `$driver` variable.

devorama:
I see now yes, sorry read the the read and green lines and swapped in my head.
Thanks, have a great day

PR #3459, opened 2020-08-07
Custom Filters for Code Modules

devorama:
Add ability to add custom namespace filters to allow 
config of seperate code not to require change of core systems config
files. This will allow easier updates of core system without
concern of overwriting custom filters that was added.

I dont know how todo a Testcase for this if anybody has ideas I will do it

Also If accepted the Code Modules doc should be update, I have not done before but would like to
if it is accepted

Signed-off-by: devorama <33496470+devorama@users.noreply.github.com>

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -4,6 +4,9 @@ use CodeIgniter\Config\BaseConfig;
 
 class Filters extends BaseConfig
 {
+	//Enable or Disable Auto Discovery of custom Filters
+	public $discoverFilters = false;	
+	
 	// Makes reading things below nicer,
 	// and simpler to change out script that's used.
 	public $aliases = [
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -4,7 +4,6 @@ use CodeIgniter\Config\BaseConfig;
 
 class Filters extends BaseConfig
 {
-
 	// Makes reading things below nicer,
 	// and simpler to change out script that's used.
 	public $aliases = [
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -4,9 +4,7 @@ use CodeIgniter\Config\BaseConfig;
 
 class Filters extends BaseConfig
 {
-	//Enable or Disable Auto Discovery of custom Filters
-	public $discoverFilters = false;	
-	
+
 	// Makes reading things below nicer,
 	// and simpler to change out script that's used.
 	public $aliases = [
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -112,7 +112,7 @@ class Filters
 	 * @param \Config\Filters   $config
 	 * @param RequestInterface  $request
 	 * @param ResponseInterface $response
-	 * @param $moduleConfig
+	 * @param \Config\Modules	$moduleConfig
 	 */
 	public function __construct($config, RequestInterface $request, ResponseInterface $response,$moduleConfig = null)
 	{
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -42,6 +42,7 @@ namespace CodeIgniter\Filters;
 use CodeIgniter\Filters\Exceptions\FilterException;
 use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
+use Config\Services;
 
 /**
  * Filters
@@ -110,7 +111,40 @@ class Filters
 		$this->config  = $config;
 		$this->request = &$request;
 		$this->setResponse($response);
+		if ($this->config->discoverFilters)
+		{
+			$this->discoverFilters();
+		}		
 	}
+	
+	//--------------------------------------------------------------------
+
+	/**
+	 * If discoverFilters is enabled in Config then system will try to auto
+	 * Discovery custom filters files in Namespaces and allow access to 
+	 * The config object via the variable $customfilters as with the routes file
+	 * Sample : 
+	 * $customfilters->aliases['custom-auth'] = \Acme\Blob\Filters\BlobAuth::class;
+	 */	
+	private function discoverFilters()
+	{
+		$locater = Services::locator();
+		
+		$customfilters = $this->config;
+		
+		$files = $locater->search('Config/Filters.php');
+
+		foreach ($files as $file)
+		{
+			// Don't include our main file again...
+			if ($file === APPPATH . 'Config/Filters.php')
+			{
+				continue;
+			}
+
+			include $file;
+		}
+	}	
 
 	/**
 	 * Set the response explicity.
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -111,7 +111,7 @@ class Filters
 		$this->config  = $config;
 		$this->request = &$request;
 		$this->setResponse($response);
-		if ($this->config->discoverFilters)
+		if ((!is_array($this->config)) && ($this->config->discoverFilters))
 		{
 			$this->discoverFilters();
 		}		
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -119,11 +119,10 @@ class Filters
 		$this->config  = $config;
 		$this->request = &$request;
 		$this->setResponse($response);
+
 		$this->moduleConfig = $moduleConfig;
-		if ($this->moduleConfig == null)
-		{
-			$this->moduleConfig = config('Modules');
-		}
+		$this->moduleConfig = $moduleConfig ?? config('Modules');
+
 		if ($this->moduleConfig->shouldDiscover('filters'))
 		{
 			$this->discoverFilters();
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -123,11 +123,6 @@ class Filters
 
 		$this->moduleConfig = $moduleConfig;
 		$this->moduleConfig = $moduleConfig ?? config('Modules');
-
-		if ($this->moduleConfig->shouldDiscover('filters'))
-		{
-			$this->discoverFilters();
-		}		
 	}
 	
 	//--------------------------------------------------------------------
@@ -143,7 +138,7 @@ class Filters
 	{
 		$locater = Services::locator();
 		
-		$customfilters = $this->config;
+		$filters = $this->config;
 		
 		$files = $locater->search('Config/Filters.php');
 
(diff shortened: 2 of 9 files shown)


avegacms:
I think it's a good feature for modules

system/Filters/Filters.php
```diff
@@ -110,7 +111,40 @@ public function __construct($config, RequestInterface $request, ResponseInterfac
 		$this->config  = $config;
 		$this->request = &$request;
 		$this->setResponse($response);
+		if ((!is_array($this->config)) && ($this->config->discoverFilters))
```
samsonasik, line 114: `$this->config` is a `\Config\Filters` instance, so, can use the following check:

```suggestion
              if (isset($this->config->discoverFilters) && $this->config->discoverFilters)
```

devorama: Thanks you, have made changes this makes alot more sense

devorama:
> Great addition! PR looks pretty good, please be sure to run it through the code sniffer for style guide compliance. I would also 
like a test case to confirm discovery is working as intended.
@MGatner  Hi there, If you can give suggestions on the test case for that is the part that I dunno as all the test cases are based around hardcoded config being passed and not a actual Code module system running inside the actual whole CI framework, I will try and look at other test cases on how the  load but it seems all are more directly related to specific modules areas and not loading the whole code module, thus really stummped at writing a test for it. Personally have tested it in my own enviroment and knows it work but understand theneed of the test case for this.

paulbalandan:
You should probably use `Config\Modules` in setting up auto-discovery just like the others, like events, etc.

devorama:
@MGatner  and @paulbalandan  I think this solution for testing the custom filter is valid and will work. Basically I have added the custom filter under tests/_support/Config/Filters.php and it creates a class for it under tests/_support/Filters/Customfilter.php I then updated the routes to add a new route that uses this filter.  tests/_support/Config/Routes.php to look like this $routes->add('testingfilter', 'TestController::index',  ['filter' => 'test-customfilter']); normally as I remember this creates a error if the filter was not found, I will look to maybe add another test controller to be called regardless to make sure it gets called but I think this is the best way one could emulate a code module and test the actual use of the functionality in the system, hope that will be okay

devorama:
trying to get routes and filters to load as a code module correctly when doing the codeignitertest  as CI's AutoloaderConfig does load the _support as a code module
prent

Found that routes file did not have namespace added that hopedully it will solve the probelm and route will point to hello controller directly for testing, hope this works

michalsn:
Maybe I am missing something but is there any reason why this shouldn't be handled via `Config/Modules` as @paulbalandan pointed out?

devorama:
> 
> 
> Maybe I am missing something but is there any reason why this shouldn't be handled via `Config/Modules` as @paulbalandan pointed out?

I basically dont know enough of CI inner workings yet to write a test for this, for looking at how events test work with Config/Modules I dont see how one can do that for Filters with it. I will look again but test wise I wanted to achieve a test to load it and use it, but I think the closes I can come maybe is to test the autodiscovery but it feels wrong as need to confirm it will work inside the workings between loading custom filter from code module and using in a router and I cant achieve that , but will re-look

devorama:
I checked code again and Modules work for Events and routes for they actually have code for it in their classes and a module config get passed to them and then inside they basically use the module as the config flag I used in the filters config file. I did not want to change teh Filters constructor and dont think it is wise to just create a default module object (config('Modules') ) . Basically still stumped how to test this. If anybody can explain what I am missing will appreciate for the actual code works I know I am using it, but to write a test is confusing as I actually dont know how the test system works behind the scenes when and how is certain objects loaded, is the whole CI system loaded at a point etc, basically a few stuff for me still to learn before I can fully grasp how to get the test modules in place.

devorama:
@MGatner @michalsn @paulbalandan  Thank you I figured out how the testing system now loads and how to manipulate it for testing purposes, the testing will now be successfull and test the loading and calling of a custom filter from a code module.  I was considering to introduce the Modules unit into filters also as with Events and Routes to allow a person to enable and disable the module as with those, this will require me to change the contructor of Filters by adding the module config at the end , I will default it to null and if it is null then load the default config('Modules') for example, this will then remove the need of the configuration variable in Filters config itself to set discovering of filters on or off. Must I rather do that for this instead of a new pull request ?

devorama:
I made the change as it is more inline with the design of CI to use the modules config to enable disable the discovery of custom files from there, thus removed the new config value in filters config and added the logic to use module config as required

system/Filters/Filters.php
```diff
@@ -104,13 +112,52 @@ class Filters
 	 * @param \Config\Filters   $config
 	 * @param RequestInterface  $request
 	 * @param ResponseInterface $response
+	 * @param \Config\Modules	$moduleConfig
```
samsonasik, line 115: @param \Config\Modules|null

system/Filters/Filters.php
```diff
@@ -104,13 +112,52 @@ class Filters
...
+	public function __construct($config, RequestInterface $request, ResponseInterface $response,$moduleConfig = null)
 	{
 		$this->config  = $config;
 		$this->request = &$request;
 		$this->setResponse($response);
+		$this->moduleConfig = $moduleConfig;
+		if ($this->moduleConfig == null)
+		{
+			$this->moduleConfig = config('Modules');
+		}
```
samsonasik, line 126: can use $this->moduleConfig = $moduleConfig ?? config('Modules');

system/Filters/Filters.php
```diff
@@ -104,13 +112,52 @@ class Filters
 	 * @param \Config\Filters   $config
 	 * @param RequestInterface  $request
 	 * @param ResponseInterface $response
+	 * @param \Config\Modules	$moduleConfig
 	 */
-	public function __construct($config, RequestInterface $request, ResponseInterface $response)
+	public function __construct($config, RequestInterface $request, ResponseInterface $response,$moduleConfig = null)
```
samsonasik, line 117: need \Config\Modules type hint

samsonasik:
please rebase latest develop, phpstan level increased to level 3 that can do more static analysis check

system/Filters/Filters.php
devorama: Hi ther, I am little confused how todo this, copied from RouterCollection so that I try and follow the standard and it looks like this , how do I do that Config type hint ?
prent

devorama:
> 
> 
> please rebase latest develop, phpstan level increased to level 3 that can do more static analysis check

Hi there, I think have it correct  now ? it seems to be happy with the statis analysis check in all and looks good I think, no idea what it does.
prent

system/Filters/Filters.php
devorama: K I think I have booked a solution in that will work, I looked at the FileLocator in RouteCollection and copied the same concept from it for Modules.

devorama:
> 
> 
> please rebase latest develop, phpstan level increased to level 3 that can do more static analysis check

Followed your tutorial thank you and rebased and forced update back in as per tutorial. should all be happy now

system/Filters/Filters.php
```diff
@@ -104,13 +113,46 @@ class Filters
 	 * @param \Config\Filters   $config
 	 * @param RequestInterface  $request
 	 * @param ResponseInterface $response
+	 * @param \Config\Modules|null $moduleConfig
 	 */
-	public function __construct($config, RequestInterface $request, ResponseInterface $response)
+	public function __construct($config, RequestInterface $request, ResponseInterface $response,Modules $moduleConfig = null)
```
michalsn, line 118:
```suggestion
	public function __construct($config, RequestInterface $request, ResponseInterface $response, Modules $moduleConfig = null)
```

system/Filters/Filters.php
```diff
@@ -104,13 +113,46 @@ class Filters
...
 	{
 		$this->config  = $config;
 		$this->request = &$request;
 		$this->setResponse($response);
+
+		$this->moduleConfig = $moduleConfig;
```
michalsn, line 124:
```suggestion
```

system/Filters/Filters.php
```diff
@@ -104,13 +113,46 @@ class Filters
...
 	}
+	
+	//--------------------------------------------------------------------
+
+	/**
+	 * If discoverFilters is enabled in Config then system will try to auto
+	 * Discovery custom filters files in Namespaces and allow access to 
+	 * The config object via the variable $customfilters as with the routes file
+	 * Sample : 
+	 * $customfilters->aliases['custom-auth'] = \Acme\Blob\Filters\BlobAuth::class;
```
michalsn, line 135:
```suggestion
	 * $filters->aliases['custom-auth'] = \Acme\Blob\Filters\BlobAuth::class;
```

michalsn (changes requested):
I think this is really the last round of changes from me. Thanks!

michalsn (approved):
It looks good to me, but I'll let @MGatner review this before merging as he was already involved here.

app/Config/Filters.php
```diff
@@ -4,6 +4,7 @@
 
 class Filters extends BaseConfig
 {
+
```
MGatner, line 7: Please remove this newline, as it constitutes a change to a file in **app** that otherwise does not need to be touched.
```suggestion
```

tests/_support/Config/Routes.php
```diff
@@ -1,7 +1,5 @@
-<?php
-
+<?php namespace  Tests\Support\Config;
```
MGatner, line 1: Extra space here.
```suggestion
<?php namespace Tests\Support\Config;
```

tests/system/CodeIgniterTest.php
```diff
@@ -56,7 +56,7 @@ public function testRunEmptyDefaultRoute()
 
 		$this->assertStringContainsString('Welcome to CodeIgniter', $output);
 	}
-
+
```
MGatner, line 59: This file does not need any changes.
```suggestion
```

devorama: Sorry was one of the places I tried todo a test for the filter from will correct

devorama:
No idea why Static Analysis Check is giving this error, it ons files I did not ouch ?
prent

samsonasik:
It seems new phpstan released https://github.com/phpstan/phpstan/releases/tag/0.12.40

You seems can remove `// @phpstan-ignore-line` at:

-  `system/HTTP/Response.php` line 902
- `system/I18n/Time.php` line 560

samsonasik:
I created PR https://github.com/codeigniter4/CodeIgniter4/pull/3547 for phpstan notice

PR #3465, opened 2020-08-08
Added rm_string in text_helper

### LINKED ISSUE #3462: Form validation does not have a rule to validate if field value differs from a string
Hello i am trying to validate an input that must have another string besides what i set

So for example
i have an input type text with value of "Test"

I want to validate this field that must be changed so it cannot have string Test inside the field

I have no rule for this
Something like 
is_not_string[test]

mmrtonmoybd:
Allows to removed specific keyword or fields from string.
I added a new function rm_string  in text helper. It will removed specific keywords or fields from given string.
I am not good at English, so I did not give huge description about it.
Thank You All.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/text_helper.php
+++ b/system/Helpers/text_helper.php
@@ -880,4 +880,34 @@ if (! function_exists('excerpt'))
 	}
 
 	//--------------------------------------------------------------------
+	if (! function_exists('is_not_string'))
+{
+	/**
+	 * Is Not String.
+	 *
+	 * Allows to removed specific keyword or fields from string
+	 *
+	 * @param string  $str      Given String
+	 * @param array  $fields    Given value that will be replace
+	 * @param integer $replace   Matches any of the patterns are replaced with the replacement string.
+	 *
+	 * @return string
+	 *
+	 */
+	function is_not_string(string $str, array $fields, string $replace = null): string
+	{
+		$string = $str;
+		if ($replace !== null) {
+			$replace = $replace;
+		} else {
+			$replace = "";
+		}
+		foreach ($fields as $value) {
+		$string = preg_replace("~{$value}~s", $replace, $string);
+		}
+		return $string;
+	}
+}
+
+	//--------------------------------------------------------------------
 }
--- a/system/Helpers/text_helper.php
+++ b/system/Helpers/text_helper.php
@@ -880,34 +880,29 @@ if (! function_exists('excerpt'))
 	}
 
 	//--------------------------------------------------------------------
-	if (! function_exists('rm_string'))
+	if (! function_exists('replace_string'))
 {
 	/**
 	 * Remove String.
 	 *
 	 * Allows to removed specific keyword or fields from string
 	 *
-	 * @param string  $str      Given String
-	 * @param array  $fields    Given value that will be replace
+	 * @param string  $string Given String
+	 * @param array  $values    Given value that will be replace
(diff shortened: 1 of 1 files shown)


mmrtonmoybd:
@crustamet may be your problem are fixed now.
@michalsn  am I do right ?

michalsn:
The initial issue was related to the validation, but it seems it can be solved with an existing rule.

samsonasik:
please follow coding standard, you can run composer update and it will auto-fix in next commits when you make new changes.

michalsn:
I'm not really sure if this functionality fits as a helper function. Basically it's a one-liner in a loop... Also, why `preg_replace()` and not `str_replace()`?

Personally I'm not sold yet, but maybe you have some good use cases?

Other things:
* Code should comply with our code style
* We need tests
* We need updates to the user guide

system/Helpers/text_helper.php
```diff
@@ -879,5 +879,35 @@ function excerpt(string $text, string $phrase = null, int $radius = 100, string
...
+		if ($replace !== null) {
+			$replace = $replace;
+		} else {
+			$replace = "";
+		}
+		foreach ($fields as $value) {
+		$string = preg_replace("~{$value}~s", $replace, $string);
+		}
+		return $string;
+	}
```
michalsn, line 909:
```suggestion
	function replace_string(string $string, array $fields, string $replace = ''): string
	{
		foreach ($fields as $value)
		{
			$string = preg_replace("~{$value}~s", $replace, $string);
		}

		return $string;
	}
```

system/Helpers/text_helper.php
```diff
@@ -880,4 +880,29 @@ function excerpt(string $text, string $phrase = null, int $radius = 100, string
...
+	 */
+	function replace_string(string $string, array $values, string $replace = ''): string
+	{
+		foreach ($values as $value)
+		{
+			$string = preg_replace("~{$value}~s", $replace, $string);
+		}
+		return $string;
+	}
+
```
samsonasik, line 905: missing close `}`, also, again, please provide unit tests and follow user guide, do composer update and it will automatically fix it on next commit.

mmrtonmoybd:
@samsonasik 
I will fixed all problem soon, also give a unit test. 
I close this pull request now, again reopen it soon. 
This replace_string are helpful . when the developer didn't show some words/keywords in blog post /comment box.
Example : I want to push drugs in her body.
If I use this function the drugs will be replaced.
Output: I want to push in her body.

Soon I give a new PR .
Thanks to all Codeigniter Teams

michalsn:
I just realized that we have `word_censor()` function already. So this would be redundant functionality.

https://codeigniter4.github.io/userguide/helpers/text_helper.html#word_censor

mmrtonmoybd:
@michalsn  Yes the function word_censor is same as my expected. I have not see this function at before. 
Sorry for my mistake 😣😣😣.

michalsn:
No worries, I forgot about this too. I just remembered it because you gave me your use case example.

As for route thing. We have `route_to()` and `url_to()` functions: 
https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Common.php#L953 
https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Helpers/url_helper.php#L639

PR #3467, opened 2020-08-08
Allow your own primary key to be used with the Model class

michalsn:
**Description**
This PR adds a new option to the Model class: `$useAutoIncrement` which when set to `false` allows us to provide our own primary key for each record in the table. This is really handy e.g. when we want to implement 1:1 relation or use UUIDs in our table.

This new feature doesn't introduce any BC changes.

Ref: #3134

Closes #3211

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


system/Database/Exceptions/DataException.php
```diff
@@ -29,6 +29,20 @@ public static function forEmptyDataset(string $mode)
 		return new static(lang('Database.emptyDataset', [$mode]));
 	}
 
+	/**
+	 * Used by Model's insert/update methods when there isn't
```
paulbalandan, line 33:
```suggestion
	 * Used by Model's insert/update methods when there is no
```

system/Database/Exceptions/DataException.php
```diff
@@ -29,6 +29,20 @@ public static function forEmptyDataset(string $mode)
...
+	 * primaryKey defined and Model has option useAutoIncrement
```
paulbalandan, line 34:
```suggestion
	 * primary key defined and Model has option `useAutoIncrement`
```

system/Language/en/Database.php
```diff
@@ -19,6 +19,7 @@
    'invalidArgument'                  => 'You must provide a valid {0}.',
    'invalidAllowedFields'             => 'Allowed fields must be specified for model: {0}',
    'emptyDataset'                     => 'There is no data to {0}.',
+   'emptyPrimaryKey'                  => 'There is no primaryKey defined when trying to make {0}.',
```
paulbalandan, line 22:
```suggestion
   'emptyPrimaryKey'                  => 'There is no primary key defined when trying to make {0}.',
```

system/Model.php
```diff
@@ -98,6 +98,13 @@ class Model
 	 */
 	protected $primaryKey = 'id';
 
+	/**
+	 * Primary key uses auto increment.
```
paulbalandan, line 102:
```suggestion
	 * Whether primary key uses auto increment.
```

system/Model.php
```diff
@@ -619,11 +626,34 @@ public function save($data): bool
 			return true;
 		}
 
-		if (is_object($data) && isset($data->{$this->primaryKey}))
+		$makeUpdate = $this->useAutoIncrement;
+
+		// When useAutoIncrement feature is disabled check
+		// in the database if given record already exists
+		if (! $this->useAutoIncrement)
```
paulbalandan, line 633:
```suggestion
		if (! $makeUpdate)
```

tests/system/Database/Live/ModelTest.php
```diff
@@ -2303,6 +2304,75 @@ public function testSoftDeleteWithTableJoinsFirst()
 
 	//--------------------------------------------------------------------
 
+	public function testUseAutoIncrementSetToFalseInsertException()
+	{
+		$this->expectException(DataException::class);
+		$this->expectExceptionMessage('There is no primaryKey defined when trying to make insert');
```
paulbalandan, line 2310:
```suggestion
		$this->expectExceptionMessage('There is no primary key defined when trying to make insert');
```

user_guide_src/source/models/model.rst
```diff
@@ -139,6 +141,16 @@ is used with methods like ``find()`` to know what column to match the specified
...
     as expected.
 
+**$useAutoIncrement**
+
+Specifies if the table uses an auto-increment feature for ``$primaryKey``. If set to ``false``
+then you are responsible for providing primary key value for every record in the table. This 
+feature may be handy when we want to implement 1:1 relation or use UUIDs for our model.
+
+.. note:: If you set ``$useAutoIncrement`` to ``false`` then make sure to set your primary
+    key in the database to ``unique``. This way you will make sure that all Model's features
```
paulbalandan, line 151:
```suggestion
    key in the database to ``unique``. This way you will make sure that all of Model's features
```

user_guide_src/source/models/model.rst
```diff
@@ -139,6 +141,16 @@ is used with methods like ``find()`` to know what column to match the specified
...
+    will work the same as before.
```
paulbalandan, line 152:
```suggestion
    will still work the same as before.
```

michalsn:
Thank you :)

MGatner:
Let me think about this, and/or have @lonnieezell take a look. While it doesn't break anything from the framework, some of these changes (like `insertID()` can now return a string) could break people's integrations. For example, a module that assumes numeric indexes and attempts arithmetic operations or sorting.

MGatner:
I don't think it is a breaking change sort of issue. I'm wondering how many projects and modules have been made on the assumption that models and entities use `int` values for primary keys, and if we allow non-auto-incrementing strings are we responsible for that philosophy shift in a minor patch.

My guess is that my own answer is that this is an improvement and since there is no BC we move ahead. But I want to make sure that seems right to all. I currently have a Firebase project that use's Google's UIDs as primary key, so I get the interest for this but I don't want to be hasty about it.

michalsn:
Well... philosophy doesn't really change - until you disable auto-increment functionality. If this happens then it's the developer's responsibility to adjust their code. We don't change anything by default.

But I get it - I appreciate your concern and we can postpone this to 4.1.

lonnieezell:
I have no problems with the way this is implemented. I think we can add this now. Unless you're using a strict version of CI4 I think type conversion would handle 99% of any cases I can think of. I'm fine to include it now.

PR #3468, opened 2020-08-09
Turn on toolbar if CI_DEBUG equals true

MashinaMashina:
Set constant CI_DEBUG into TRUE must be enable debug toolbar

We can define in file https://github.com/codeigniter4/CodeIgniter4/blob/develop/app/Config/Boot/production.php

### DIFF
--- a/app/Config/Events.php
+++ b/app/Config/Events.php
@@ -44,7 +44,7 @@ Events::on('pre_system', function () {
 	 * --------------------------------------------------------------------
 	 * If you delete, they will no longer be collected.
 	 */
-	if (ENVIRONMENT !== 'production')
+	if (CI_DEBUG)
 	{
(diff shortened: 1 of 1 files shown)


michalsn:
And the reason for this change is... ? 

IMO this may introduce security issues. I think the current behavior is more appropriate.

MashinaMashina:
The documentation says that the toolbar is enabled by CI_DEBUG, but in reality the toolbar is enabled by ENVIRONMENT constant.

https://codeigniter4.github.io/userguide/testing/debugging.html?highlight=ci_debug

If ENVIRONMENT equals production we can't enable toolbar.
If ENVIRONMENT equals development we can't totally disable toolbar.

michalsn:
Oh... okay, that's a good point. You should mention this at the beginning - merging :)

Issue #3470, opened 2020-08-09
before() Filters that are executed are missing from codeigniter debug toolbar.


crustamet:
I have some filters that does some inserts into a database after and before.

Those queries i don't see them into the toolbar from after() Filter Function only from before().

I would like to see those queries if possible.

PR #3472, opened 2020-08-10
Run the toolbar filter as the last one

michalsn:
**Description**
The `toolbar` filter should be run as the last one since in other case it can't register all things are happening in other filters.

See: #3470

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -240,6 +240,16 @@ class Filters
 		$this->processMethods();
 		$this->processFilters($uri);
 
+		// Set the toolbar filter to the last position to be executed
+		if (in_array('toolbar', $this->filters['after']) &&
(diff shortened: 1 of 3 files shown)


system/Filters/Filters.php
```diff
@@ -240,6 +240,16 @@ public function initialize(string $uri = null)
 		$this->processMethods();
 		$this->processFilters($uri);
 
+		// Set the toolbar filter to the last position to be executed
+		if (in_array('toolbar', $this->filters['after']) &&
```
samsonasik, line 244: can use strict in array: 

```suggestion
              if (in_array('toolbar', $this->filters['after'], true) &&
```

michalsn: Thanks!

PR #3476, opened 2020-08-10
Fix help command on inexistent commands

paulbalandan:
**Description**
We have made spark give suggestions when possible for inexistent commands but failed to account for cases when we use the `help` command to get info on a command, which can be inexistent too.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/Commands.php
+++ b/system/CLI/Commands.php
@@ -87,26 +87,8 @@ class Commands
 	{
 		$this->discoverCommands();
 
-		if (! isset($this->commands[$command]))
+		if (! $this->verifyCommand($command, $this->commands))
 		{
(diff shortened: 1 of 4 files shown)


system/CLI/Commands.php
```diff
@@ -197,6 +179,43 @@ public function discoverCommands()
...
+
+				$message .= implode("\n    ", $alternatives);
+			}
+
+			CLI::error($message);
+			CLI::newLine();
+			return false;
+		}
+
+		return true;
```
michalsn, line 216: Can we make it less "nested"? I mean something like:
```php
if (isset($commands[$command]))
{
	return true;
}
```

PR #3480, opened 2020-08-11
[Feature] Robust code scaffolder commands

paulbalandan:
**Description**
Add code generators that leverages the power of `GeneratorCommand`.

- `make:controller <name> [options]`
- `make:model <name> [options]`
- `make:entity <name> [options]`
- `make:scaffold <name> [options]`
- `make:command <name> [options]`
- `make:filter <name> [options]`

Tests are underway, but I would like to have a few feedback of additions or removals from current proposed.
Default property values are lifted from the respective parent class.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide
- [x] User guide updated

### DIFF
--- a/system/Commands/Generators/CreateController.php
+++ b/system/Commands/Generators/CreateController.php
@@ -1,0 +1,254 @@
+<?php
+
+/**
+ * CodeIgniter
+ *
+ * An open source application development framework for PHP
+ *
+ * This content is released under the MIT License (MIT)
+ *
+ * Copyright (c) 2014-2019 British Columbia Institute of Technology
+ * Copyright (c) 2019-2020 CodeIgniter Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @package    CodeIgniter
+ * @author     CodeIgniter Dev Team
+ * @copyright  2019-2020 CodeIgniter Foundation
+ * @license    https://opensource.org/licenses/MIT	MIT License
+ * @link       https://codeigniter.com
+ * @since      Version 4.0.0
+ * @filesource
+ */
+
+namespace CodeIgniter\Commands\Generators;
+
+use CodeIgniter\CLI\CLI;
+use CodeIgniter\CLI\GeneratorCommand;
+
+/**
+ * Creates a skeleton controller file.
+ */
+class CreateController extends GeneratorCommand
+{
+	/**
+	 * The Command's name
+	 *
+	 * @var string
+	 */
+	protected $name = 'make:controller';
+
+	/**
+	 * The Command's short description
+	 *
+	 * @var string
+	 */
+	protected $description = 'Creates a new controller file.';
+
+	/**
+	 * The Command's usage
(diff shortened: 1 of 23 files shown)


truelineinfotech:
@paulbalandan thanks for this awesome Feature. can we have one more command for module something like
`make:module <name> [options]`

michalsn:
I like this. You added many features here.

Although we already have one attempt for generators #3426. This PR seems more advanced and feature-rich.

I only have one suggestion. I guess developers may want to extend these classes to prepare their own small changes for generated files. I believe it would be better if they have access to all methods. So maybe we should switch from `private` to `protected`? We use `private` in a few places.

system/Commands/Generators/CreateController.php
```diff
@@ -0,0 +1,254 @@
...
+		'name' => 'The controller class name',
+	];
+
+	/**
+	 * The Command's options
+	 *
+	 * @var array
+	 */
+	protected $options = [
+		'-base'    => 'Extends from BaseController instead of CodeIgniter\\Controller.',
```
lonnieezell, line 86: Shouldn't extending from BaseController be the default?

paulbalandan:
> I would love to see us snag one thing on this from Laravel: if I give the name of the Controller/Model/whatever as `Folder/ControllerName` it creates the folder with the `Controllers` directory and modifies the namespace. This is super handy for creating controllers in, say, Admin folders, etc.

We have this functionality baked in already. It is a default option inherited from `GeneratorCommand`. Using the `-n` option will direct the generation to that specific namespace.
`make:controller user -n CodeIgniter` will create a controller file in SYSTEMPATH/Controllers/User.php with namespace `CodeIgniter\Controllers`.

paulbalandan:
> @paulbalandan thanks for this awesome Feature. can we have one more command for module? something like
> `make:module <name> [options]`

Im not sure what you want here. In my understanding a module is a complete library by itself, and generating that will be complicated.

paulbalandan:
> I only have one suggestion. I guess developers may want to extend these classes to prepare their own small changes for generated files. I believe it would be better if they have access to all methods. So maybe we should switch from `private` to `protected`? We use `private` in a few places.
Fair point. I'll change that later.

truelineinfotech:
> > @paulbalandan thanks for this awesome Feature. can we have one more command for module? something like
> > `make:module <name> [options]`
> 
> Im not sure what you want here. In my understanding a module is a complete library by itself, and generating that will be complicated.

https://github.com/enmanuel97/cli-create
https://github.com/enmanuel97/cli-create/blob/master/src/Commands/Module.php

paulbalandan:
> > > @paulbalandan thanks for this awesome Feature. can we have one more command for module? something like
> > > `make:module <name> [options]`
> > 
> > 
> > Im not sure what you want here. In my understanding a module is a complete library by itself, and generating that will be complicated.
> 
> here is simple example,
> https://github.com/enmanuel97/cli-create
> https://github.com/enmanuel97/cli-create/blob/master/src/Commands/Module.php
> https://github.com/monkenWu/CodeIgniter4_CLI-Create

OK. Looking at enmanuel97/cli-create you are meaning to say that modules refer to a set. So this is handled already by `make:scaffold`. :)

lonnieezell:
> EDIT: I think I misunderstood that, creation in subfolders is also supported. That is why I created the method GeneratorCommand::sanitizeClassName. By this when I enter the class name as api/user the generated class will be App\Controllers\Api\User in the Api subfolder.

Perfect. I missed that. In that case thanks for indulging my request and making it work so quickly :)

paulbalandan:
Added some tests. Will add requested commands later.

paulbalandan:
> This looks good to me. Was there more to do here? @paulbalandan Does this cover the content and intent of the other Generators PR?

Yes, unless you have some requested generators to add. :) Also, I'm asking if the generators deserve their own section in the user guide?

MGatner:
I think that's a good idea. While they are CLI commands they have an entirely separate function and people who want them will not be looking for them nested under a particular component implementation.

paulbalandan:
Working on the user guide.

paulbalandan:
This is essentially ready.
@lonnieezell , @MGatner , @michalsn

system/Commands/Generators/Views/model.tpl.php
```diff
@@ -0,0 +1,43 @@
+<@php
+
+namespace {namespace};
+
+use CodeIgniter\Model;
+
+class {class} extends Model
```
michalsn, line 7: If we're listing every option here then I would add recently introduced variable 
```php
protected $useAutoIncrement = true;
```

paulbalandan: Wow. I almost forgot that one! Thanks.

mostafakhudair:
Thank you for this powerful add..
I would like to suggest to rename Generators commands folder to Maker or just Generator.. also rename CreateCommand CreateController etc... to MakeCommand MakeController... like its usage make:controller .. or generate:controller if you prefer to keep GeneratorCommand as it

paulbalandan:
The `Generators` namespace was suggested by @lonnieezell early on when the `make:seeder` was first created. It was not me. For the naming part of the classes, I think it doesn't make much a difference as both have same meanings. I'll have the others say what they think on this.

mostafakhudair:
I think renaming commands on the same way will be better than renaming them on the same meaning but different way

Current way:
Generator\CreateController make:controller

Suggestions:
Generator\GenerateController generate:controller
Maker\MakeController make:controller
Creator\CreateController create:controller

Its make more sense, any way i would like to hear others thoughts about that

michalsn:
I don't see any issues with the currently used namespace. Command itself is short, so everything is fine.

michalsn:
Thanks for this!

PR #3483, opened 2020-08-12
4.1.0: remove Autoloader::loadLegacy() method

samsonasik:
`Autoloader::loadLegacy()` method was previously used for migration non-namespaced classes transition to CodeIgniter v4. Since `4.1.0`, I think this can be removed.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/user_guide_src/source/changelogs/v4.1.0.rst
+++ b/user_guide_src/source/changelogs/v4.1.0.rst
@@ -1,0 +1,10 @@
+Version 4.1.0
+====================================================
+
+Release Date: Not released
+
+**4.1.0 release of CodeIgniter4**
+
+Removed:
+
+- `Autoloader::loadLegacy()` method was previously used for migration non-namespaced classes transition to CodeIgniter v4. Since `4.1.0`, this support removed.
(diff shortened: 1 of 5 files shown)


user_guide_src/source/changelogs/v4.1.0.rst
```diff
@@ -0,0 +1,10 @@
+Version 4.1.0
+====================================================
+
+Release Date: Not released
+
+**4.1.0 release of CodeIgniter4**
+
+Removed:
+
+- `Autoloader::loadLegacy()` method was previously used for migration non-namespaced classes transition to CodeIgniter v4. Since `4.1.0`, this support removed.
```
paulbalandan, line 10:
```suggestion
- `Autoloader::loadLegacy()` method was previously used for migration of non-namespaced classes in transition to CodeIgniter v4. Since `4.1.0`, this support was removed.
```

samsonasik: thank you

kenjis:
Why was this feature dropped in v4.1?

MGatner:
@kenjis See #develop threads Nov 28-30, 2020. Basically: this was to allow classes from CI3 to be used without namespaces, but changes in 4.1 required that all classes follow PSR-4 so this feature was no longer necessary.

PR #3484, opened 2020-08-12
New helper: url_is

lonnieezell:
New helper method, `url_is()` to compare the current URL with one of your choice. With (*) wildcard.

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -590,7 +590,7 @@ if (! function_exists('url_title'))
 
 		$trans = [
 			'&.+?;'                   => '',
-			'[^\w\d\pL\pM _-]'              => '',
+			'[^\w\d\pL\pM _-]'        => '',
 			'\s+'                     => $separator,
 			'(' . $q_separator . ')+' => $separator,
 		];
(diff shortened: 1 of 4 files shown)


lonnieezell:
@MGatner So someone visits `foo/`. Should `foo` work? I'm assuming yes, but that also means it should be a folder and the server probably wants to look for an index file or something. Just double-checking my assumption here.

MGatner:
I think probably so, but most importantly I think it is important to have it be consistent and documented. I field so many issues about people confused why the filter `admin/(:any)` doesn't apply to "https://example.com/admin"...

nowackipawel:
> @MGatner So someone visits `foo/`. Should `foo` work? I'm assuming yes, but that also means it should be a folder and the server probably wants to look for an index file or something. Just double-checking my assumption here.

I found implementations (i.e. bitkub.com or satang.pro api where adresses like   /foo and /foo/ are directing to different places. 
I guess if someone visits foo/  then foo should not work but foo/ or foo* should.

Maybe it will be a good idea to add an additional (optional) parameter which declare behavior of this helper?

michalsn:
> I found implementations (i.e. bitkub.com or satang.pro api where adresses like /foo and /foo/ are directing to different places.

I don't think this is possible in CodeIgniter. We treat both URLs as the same. Even if trailing slashes redirect option in the `.htaccess` file is disabled.

lonnieezell:
@MGatner How about now?

PR #3486, opened 2020-08-12
4.1.0 remove FileLocator::legacyLocate() method

samsonasik:
`FileLocator::legacyLocate()` method was previously used for ability to locate PSR-4 namespaced files and non-namespaced files. Since `4.1.0`, I think only `psr-4` namespace locate file should be supported.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -90,10 +90,10 @@ class FileLocator
 			$file = substr($file, strlen($folder . '/'));
 		}
 
-		// Is not namespaced? Try the application folder.
+		// Is not namespaced? Returns false immediately
 		if (strpos($file, '\\') === false)
 		{
-			return $this->legacyLocate($file, $folder);
+			return false;
 		}
 
 		// Standardize slashes to handle nested directories.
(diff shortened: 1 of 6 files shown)


samsonasik:
After some check, it seems used when called `config()` function without namespacing, I am closing it.

samsonasik:
I reopen it, I applied to use `return $this->autoloader->loadClass($file)` when no namespace found in passed file, assume a Config class.

samsonasik:
To reduce unnecessary `is_file()` check when Config class called repetitively, I modified `Autoloader::requireFile()` to cache to loaded files to static variable.

system/Autoloader/FileLocator.php
```diff
@@ -90,10 +90,10 @@ public function locateFile(string $file, string $folder = null, string $ext = 'p
 			$file = substr($file, strlen($folder . '/'));
 		}
 
-		// Is not namespaced? Try the application folder.
+		// Is not namespaced? assume a Config class
 		if (strpos($file, '\\') === false)
 		{
-			return $this->legacyLocate($file, $folder);
+			return $this->autoloader->loadClass($file);
```
samsonasik, line 96: the `Autoloader::loadClass()` actually do require_once file, not sure if it is ok or we can just check is_file Config class here.

lonnieezell:
Big question is why do we feel this needs to be removed? It's a huge break for those that are using it which I would think would be more of a 5.0 change than a 4.1. To be honest, I don't use it, but many do and find it easier to get into CI with. What's your thoughts?

samsonasik:
I only think about the "legacy" removal. Ok, probably 5.0 is better for this change. I am closing it.

PR #3487, opened 2020-08-13
[wip] start adding github actions for cs check

samsonasik:
**Checklist:**
- [x] Securely signed commits


.github/workflows/cs_check.yml
```diff
@@ -0,0 +1,31 @@
...
+  build:
+    name: PHP ${{ matrix.php-versions }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.4']
+    steps:
+      - name: Setup PHP Action
+        uses: shivammathur/setup-php@1.8.2
```
MGatner, line 22: We should use `setup-php@v2`, unless you have some compelling reason to stick at that version.

.github/workflows/cs_check.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches:
+      - 'master'
+    paths:
+      - 'app'
+      - 'system'
+
+jobs:
+  build:
+    name: PHP ${{ matrix.php-versions }}
```
MGatner, line 14: Since you have just specified 7.4 we don't need to use a matrix here. Use something meaningful for the name, like "Run Check"

.github/workflows/cs_check.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
```
MGatner, line 18: Ditch the matrix

.github/workflows/cs_check.yml
```diff
@@ -0,0 +1,31 @@
...
+        php-versions: ['7.4']
+    steps:
+      - name: Setup PHP Action
+        uses: shivammathur/setup-php@1.8.2
+        with:
+          extensions: intl
+          php-version: "${{ matrix.php-versions }}"
```
MGatner, line 25: `php-version: '7.4'`

.github/workflows/cs_check.yml
```diff
@@ -0,0 +1,31 @@
...
+    paths:
+      - 'app'
+      - 'system'
+
+jobs:
+  build:
+    name: PHP ${{ matrix.php-versions }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
MGatner, line 17: `fail-fast` won't apply since there is only one job to run, can remove.

.github/workflows/cs_check.yml
samsonasik: fixed.

.github/workflows/cs_check.yml
samsonasik: fixed.

.github/workflows/cs_check.yml
samsonasik: fixed.

.github/workflows/cs_check.yml
samsonasik: fixed.

.github/workflows/cs_check.yml
samsonasik: fixed.

MGatner:
The workflow looks much better! One thought... What is the intention of this workflow? Does it actually change code? Or confirm that code matches the style guide? Currently the workflow is set to trigger on `master` which should only receive `push` during release and should never receive a `pull_request`. The triggers should probably be changed to match the desired scenario.

samsonasik:
It checks if the code follow the standard, showing that new push/pull request follow the standard, like `--dry`. We probably need to use `develop` or add `4.1` as well if there is a direct push to develop/4.1. Any opinion?

MGatner:
I think if the goal is to notify developers and contributors of style issues then we should trigger it on pull requests - pushes only trigger on develop when the “damage is already done”. Maybe `on: pull_request: branches: develop 4.1`?

samsonasik:
ok, updated to use `develop` and `4.1` for direct push check.

samsonasik:
continue adding system to cs check

MGatner:
Came across this today, might offer some features over running `phpcs` ourselves: https://github.com/marketplace/actions/phpcs-code-review

samsonasik:
that's require `GH_BOT_TOKEN`, do we have it?

MGatner:
I think that is just an access token for the repo, which we already have in **deploy.yml**:
`token: ${{ secrets.ACCESS_TOKEN }}`

I'm fine with it either way. If running the command manually is working and the marketplace Action isn't adding anything useful then probably not worth the effort.

michalsn:
From what I see CS check gives us some errors right now... though I'm not sure we want to "fix" them all (or even should they be fixed). 

I believe it would be a good idea to run CS check only for files that are added to the current PR and not against all codebase. Not sure how hard it would be to set but suggested by @MGatner Action has it covered.

samsonasik:
I am closing this for now.

PR #3488, opened 2020-08-13
apply code static analysis check with phpstan

samsonasik:
apply code static analysis check with phpstan. start with level 0.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/API/ResponseTrait.php
+++ b/system/API/ResponseTrait.php
@@ -102,6 +102,13 @@ trait ResponseTrait
 	 */
 	protected $format = 'json';
 
+	/**
+	 * Formatter instance
+	 *
+	 * @var null|\CodeIgniter\Format\FormatterInterface
+	 */
+	protected $formatter = null;
+
 	//--------------------------------------------------------------------
 
 	/**
--- a/system/API/ResponseTrait.php
+++ b/system/API/ResponseTrait.php
@@ -102,13 +102,6 @@ trait ResponseTrait
 	 */
 	protected $format = 'json';
 
-	/**
-	 * Formatter instance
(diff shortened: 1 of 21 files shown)


MGatner:
My comments from the `phpcs` Actions workflow should apply here as well. 

I know nothing about phpstan so I won't review that portion yet.

samsonasik:
Found a bug when running phpstan in `system/Database/MySQLi/Result.php`: 

```
 ------ ------------------------------------------------------------------------------------- 
  97     Array has 2 duplicate keys with value 1 (MYSQLI_TYPE_TINY, MYSQLI_TYPE_CHAR).        
  111    Array has 2 duplicate keys with value 247 (MYSQLI_TYPE_INTERVAL, MYSQLI_TYPE_ENUM).  
 ------ ------------------------------------------------------------------------------------- 
```

As same index key, it may cause a wrong name on getting data type name.

samsonasik:
@MGatner phpstan run green for level 0 🎉 , ready for review ;)

system/API/ResponseTrait.php
```diff
@@ -102,6 +102,13 @@ trait ResponseTrait
 	 */
 	protected $format = 'json';
 
+	/**
+	 * Formatter instance
+	 *
+	 * @var null|\CodeIgniter\Format\FormatterInterface
+	 */
+	protected $formatter = null;
```
samsonasik, line 110: this may be a false positive ref https://3v4l.org/pv1c7

samsonasik: reverted as false positive for this trait undefined property.

PR #3489, opened 2020-08-13
Make command() mimic getopt parsing

paulbalandan:
**Description**
While making tests for my generators PR, I've noticed that we cannot pass in a multi-worded option value to `command()` simply because we are making an `explode` on the string input using the spaces. With this, `command('command -d "some value"')` will be broken into `array(0 => 'command', '-d' => '"some', 1 => 'value"')`, instead of the expected `array(0 => 'command', '-d' => 'some value')` just like how getopt will give it.

After searching the net, I came across this solution from Symfony's `StringInput::tokenize` which perfectly did the job right.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -169,6 +169,11 @@ if (! function_exists('command'))
 		$length = strlen($command);
 		$cursor = 0;
 
+		/**
+		 * Adopted from Symfony's `StringInput::tokenize()` with few changes.
+		 *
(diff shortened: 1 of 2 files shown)


paulbalandan:
Add reference comment, for proper attribution.

MGatner:
Thanks! Looks great.

PR #3493, opened 2020-08-14
Upgrading Sphinx and remove bundled layout.html template

jreklund:
Hi, just saw PR #3460 updating `sphinx-rtd-theme` into 5.0 and that means that we no longer need to bundle our own layout.html, as we only did it to be get rid of flashing content/design. And that have been fixed in this version.

I also bumped sphinx itself into 2.4.4 as it's the same version that our Github action dockerfile are using. That means that we no longer need to force `docutils` to 0.16 as Sphinx 2 supports iterator.
```
c:\program files (x86)\python38-32\lib\site-packages\sphinx\util\nodes.py:94: FutureWarning:
The iterable returned by Node.traverse()
will become an iterator instead of a list in Docutils > 0.16.
for classifier in reversed(node.parent.traverse(nodes.classifier)):
```

Other good news are that Python 3 have been forced, as you no longer can build it using Python 2. The future is now!

**Checklist:**
- [X] Securely signed commits


MGatner:
This is exciting! I remember when all these stopgaps were put in place. I'm still rather ignorant of Sphinx - do you feel this is ready to merge?

jreklund:
Indeed it is, the day have finally come! I got a notification that automatic built script for the user guide had been merged, so that's when I started to look for the script and found the theme version had been bumped. The build script itself works great (with these changes), tested it under Windows, Ubuntu 18 and the Github Action gave us a green light.

But as we bump the Sphinx version up, we get a more modern html output, and our citheme.css needs updating. I jumped the gun, only looked that it built successfully and didn't navigate past the first couple of pages. So that's something I need to correct before this can be merged.

Sphinx 2.4.4 - CI - Develop
https://codeigniter.jreklund.dev/userguide/master/outgoing/view_parser.html#setDelimiters
https://codeigniter.jreklund.dev/userguide/master/general/common_functions.html#cache
Sphinx 1.8.5 - CI 4.0.4
https://codeigniter.jreklund.dev/userguide/latest/outgoing/view_parser.html#setDelimiters
https://codeigniter.jreklund.dev/userguide/latest/general/common_functions.html#cache

MGatner:
Excellent! Thanks so much for being on top of this. I'm excited to have these in place. I will mark this `WIP` until it is ready, just remove the tag and mention me when it's set.

jreklund:
@MGatner Glad I can be of service. :-) I have now updated `citheme.css` to be compatible with Sphinx 2. Removing `navigation` completely as that is only being used in CI3 user guide apparently and made some quirky looking CSS targets to fix the rest. Now it looks like before, with some slightly smaller margins for class methods. 

PS. You may need to do a force refresh (CTRL+F5) for the demo links to show correctly. As .css files are being cached on my server.

MGatner:
Great work! Glad to see this going live.

lonnieezell:
Looking at the changes I think we're good but just verifying that the Carbon ads still get inserted correctly for production, right? @jreklund

jreklund:
@lonnieezell Yes, carbon.js are still getting bundled and included on all pages. Have run it manually to see that it gets bundled on the right location (under the menu).

PR #3495, opened 2020-08-15
Codeigniter\Entity - Added $deep parameter in toArray and toRawArray

najdanovicivan:
**Description**

Added boolean $deep parameter in toArray and toRawArray methods. This parameters allows casting inner custom and entity objects to array as well

Additionally added :
  - Unit Tests
  - Return Type Declaration for Entity Getters to avoid IDE errors
  - Some Code Style Fixes

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -140,11 +140,12 @@ class Entity implements \JsonSerializable
 	 *
 	 * @param boolean $onlyChanged If true, only return values that have changed since object creation
 	 * @param boolean $cast        If true, properties will be casted.
+	 * @param boolean $recursive   If true, inner entities will be casted as array as well.
 	 *
 	 * @return array
 	 * @throws \Exception
 	 */
-	public function toArray(bool $onlyChanged = false, bool $cast = true): array
+	public function toArray(bool $onlyChanged = false, bool $cast = true, bool $recursive = false): array
 	{
 		$this->_cast = $cast;
 		$return      = [];
@@ -164,6 +165,18 @@ class Entity implements \JsonSerializable
 			}
 
 			$return[$key] = $this->__get($key);
+
+			if ($recursive)
+			{
+				if ($return[$key] instanceof Entity)
+				{
+					$return[$key] = $return[$key]->toArray($onlyChanged, $cast, $recursive);
+				}
+				elseif (is_callable([$return[$key], 'toArray']))
+				{
+					$return[$key] = $return[$key]->toArray();
+				}
+			}
 		}
 
 		// Loop over our mapped properties and add them to the list...
@@ -174,6 +187,18 @@ class Entity implements \JsonSerializable
 				if (array_key_exists($to, $return))
 				{
 					$return[$from] = $this->__get($to);
+
+					if ($recursive)
+					{
+						if ($return[$from] instanceof Entity)
+						{
+							$return[$from] = $return[$from]->toArray($onlyChanged, $cast, $recursive);
+						}
+						elseif (is_callable([$return[$from], 'toArray']))
+						{
+							$return[$from] = $return[$from]->toArray();
+						}
+					}
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


system/Entity.php
```diff
@@ -140,11 +140,12 @@ public function fill(array $data = null)
 	 *
 	 * @param boolean $onlyChanged If true, only return values that have changed since object creation
 	 * @param boolean $cast        If true, properties will be casted.
+	 * @param boolean $deep		   If true, inner entities will be casted as array as well.
```
michalsn, line 143: Just fixing alignment.
```suggestion
	 * @param boolean $deep        If true, inner entities will be casted as array as well.
```

system/Entity.php
```diff
@@ -163,7 +164,21 @@ public function toArray(bool $onlyChanged = false, bool $cast = true): array
 				continue;
 			}
 
-			$return[$key] = $this->__get($key);
+			$value = $this->__get($key);
```
michalsn, line 167: Is it necessary to use an additional variable here? IMO `$return[$key]` would work just fine.

system/Entity.php
```diff
@@ -187,17 +216,35 @@ public function toArray(bool $onlyChanged = false, bool $cast = true): array
...
+	public function toRawArray(bool $onlyChanged = false, bool $deep = false): array
 	{
 		$return = [];
 
 		if (! $onlyChanged)
 		{
-			return $this->attributes;
+			if ($deep)
+			{
+				return array_map (function ($value) use ($onlyChanged, $deep) {
```
michalsn, line 232:
```suggestion
				return array_map(function ($value) use ($onlyChanged, $deep) {
```

system/Entity.php
```diff
@@ -187,17 +216,35 @@ public function toArray(bool $onlyChanged = false, bool $cast = true): array
...
+					if ($value instanceof Entity)
+					{
+						$value = $value->toRawArray($onlyChanged, $deep);
+					}
+					else if (is_callable([$value, 'toRawArray']))
```
michalsn, line 237:
```suggestion
					elseif (is_callable([$value, 'toRawArray']))
```

system/Entity.php
```diff
@@ -187,17 +216,35 @@ public function toArray(bool $onlyChanged = false, bool $cast = true): array
...
+					{
+						$value = $value->toRawArray();
+					}
+					return $value;
+				}, $this->attributes);
+			}
+			else
+			{
+				return $this->attributes;
+			}
```
michalsn, line 247:
```suggestion

			return $this->attributes;
```

system/Entity.php
```diff
@@ -173,7 +188,21 @@ public function toArray(bool $onlyChanged = false, bool $cast = true): array
...
-					$return[$from] = $this->__get($to);
+					$value = $this->__get($to);
+
+					if ($deep)
+					{
+						if ($value instanceof Entity)
+						{
+							$value = $value->toArray($onlyChanged, $cast, $deep);
+						}
+						else if (is_callable([$value, 'toArray']))
```
michalsn, line 199:
```suggestion
						elseif (is_callable([$value, 'toArray']))
```

system/Entity.php
```diff
@@ -163,7 +164,21 @@ public function toArray(bool $onlyChanged = false, bool $cast = true): array
...
-			$return[$key] = $this->__get($key);
+			$value = $this->__get($key);
+
+			if ($deep)
+			{
+				if ($value instanceof Entity)
+				{
+					$value = $value->toArray($onlyChanged, $cast, $deep);
+				}
+				else if (is_callable([$value, 'toArray']))
```
michalsn, line 175:
```suggestion
				elseif (is_callable([$value, 'toArray']))
```

system/Entity.php
```diff
@@ -173,7 +188,21 @@ public function toArray(bool $onlyChanged = false, bool $cast = true): array
 			{
 				if (array_key_exists($to, $return))
 				{
-					$return[$from] = $this->__get($to);
+					$value = $this->__get($to);
```
michalsn, line 191: IMO using `$return[$key]` would be better. No need to introduce an additional variable.

system/Entity.php
```diff
@@ -207,7 +254,19 @@ public function toRawArray(bool $onlyChanged = false): array
...
 			}
 
-			$return[$key] = $this->attributes[$key];
+			if ($deep)
+			{
+				if ($value instanceof Entity)
+				{
+					$value = $value->toRawArray($onlyChanged, $deep);
+				}
+				else if (is_callable([$value, 'toRawArray']))
```
michalsn, line 263:
```suggestion
				elseif (is_callable([$value, 'toRawArray']))
```

paulbalandan:
How about renaming the `$deep` parameter to `$recursive`? IMO it is more descriptive of what it does.

najdanovicivan:
> How about renaming the `$deep` parameter to `$recursive`? IMO it is more descriptive of what it does.

@paulbalandan It makes sense to me using either of those names let's see what other think about it.

MGatner:
`$recursive` would be more consistent with other variable names in the framework already.

system/Entity.php
```diff
@@ -207,7 +248,19 @@ public function toRawArray(bool $onlyChanged = false): array
...
```
michalsn, line 257: You missed one spot :)
```suggestion
				elseif (is_callable([$value, 'toRawArray']))
```

najdanovicivan:
@michalsn regarding the `else if` and `elseif` can we see to get this rule added in CodeSniffer standard ?

PR #3500, opened 2020-08-16
Fixes EncryptionException::forNoHandlerAvailable() handler parameter call

samsonasik:
Two things:

- `EncryptionException::forNoHandlerAvailable` is for to pass handler name, it was passed driver in `Encryption::__construct()`.
- `EncryptionException::forNoHandlerAvailable()` require `$handler` parameter, it was no parameter.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Encryption/Encryption.php
+++ b/system/Encryption/Encryption.php
@@ -87,6 +87,13 @@ class Encryption
 		'OpenSSL',
 	];
 
+	/**
+	 * Handlers that are to be installed
+	 *
+	 * @var array
+	 */
+	protected $handlers = [];
+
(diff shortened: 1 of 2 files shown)


samsonasik:
For note, it catched by phpstan level 1 that I run locally (currently level 0 that in github). For phpstan level 1, it may require separate PR as there will be a false positives to registered.

samsonasik:
Actually, the `$this->handlers` is not really needed ?

samsonasik:
I updated to remove unneeded `$this->handlers` definitions as never used, use `$this->driver` as the handler definition instead.

PR #3502, opened 2020-08-16
Set default logging threshold to 4

paulbalandan:
**Description**
Looking through the forums, I've realized that developers' dilemma of `Unable to connect to database` exceptions can be known its root cause by logging the errors also. The default is set to `3` which logs only critical errors and higher. This PR suggests increasing the threshold to `4` to make errors logged too.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Logger.php
+++ b/app/Config/Logger.php
@@ -32,7 +32,7 @@ class Logger extends BaseConfig
 	| your log files will fill up very fast.
 	|
 	*/
-	public $threshold = 3;
+	public $threshold = 4;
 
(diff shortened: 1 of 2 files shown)


lonnieezell:
I agree with this change. Good to go once that test gets fixed.

paulbalandan:
Thanks. The failing test asserts that there are no logged errors for the day. This would be now failing mainly because errors are now being recorded during the test run. I'll probably change the assertion for this.

PR #3508, opened 2020-08-17
Codeigniter\Entity - Fix ToArray datamap

najdanovicivan:
**Description**

There are couple of issues with the way datamap works 

- It keeps doubles the values original and mapped where only mapped should be kept
- If datamap is used on existing attribute it will not work properly as value is already processed through data map in the getter
- When unsetting the key we need to double check if there is not another map that uses the key in $to map as if unset it won't be accessible in the next loop

I've added the getSwappedEntity() method which returns the entity which is affected by all those issues. 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -150,15 +150,21 @@ class Entity implements \JsonSerializable
 		$this->_cast = $cast;
 		$return      = [];
 
+		$keys = array_keys($this->attributes);
+		$keys = array_filter($keys, function ($key) {
+			return strpos($key, '_') !== 0;
+		});
+
+		if (is_array($this->datamap))
+		{
+			$keys = array_diff($keys, $this->datamap);
+			$keys = array_unique(array_merge($keys, array_keys($this->datamap)));
+		}
+
 		// we need to loop over our properties so that we
 		// allow our magic methods a chance to do their thing.
-		foreach ($this->attributes as $key => $value)
+		foreach ($keys as $key)
 		{
-			if (strpos($key, '_') === 0)
-			{
-				continue;
-			}
-
 			if ($onlyChanged && ! $this->hasChanged($key))
 			{
 				continue;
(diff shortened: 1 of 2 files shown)


najdanovicivan:
This PR is not final yet. This is just to understand the issue. I'm thinking about more elegant solution to avoid having all those checks and do everything in single foreach loop.

As datamap already works on magic getter the idea is to properly prepare keys and then run foreach loop to get values using those keys. 

The way to prepare those keys will be done like this

1. get all keys of $this->attributes let those be named $keys
2. filter out keys starting with _underscore from $keys
3. get unique values fro the $this->datamap let those be named as $mapped_keys
4. filter out the $mapped_keys from the $key 
5. Append all keys from $this->datamap to $keys 
6. Run single foreach $keys as $key and pouplate $results with $this->__get($key)
 
Let me know what you think about doing it this way

najdanovicivan:
I've submitted another commit which will handle this by processing keys first

michalsn:
I like this. Nice and clean solution.

system/Entity.php
```diff
@@ -150,15 +150,21 @@ public function toArray(bool $onlyChanged = false, bool $cast = true, bool $recu
 		$this->_cast = $cast;
 		$return      = [];
 
+		$keys = array_keys($this->attributes);
+		$keys = array_filter($keys, function ($key) {
```
MGatner, line 154: Isn't this just `preg_filter()`?

tests/system/EntityTest.php
```diff
@@ -916,6 +924,28 @@ protected function getSimple()
...
+
+			protected $_original = [
+				'foo' => 'foo',
+				'bar' => 'bar',
+			];
+
+			protected $datamap = [
+				'bar'          => 'foo',
+				'foo'          => 'bar',
+				'original_bar' => 'bar',
```
kenjis, line 944: @najdanovicivan 
Each property (array key) in the Entity has a value, so if the column `bar` has two properties, 
we need to discard one value when inserting to database.

The `$datamap` seems to be invalid. Am I something wrong?

kenjis:
@najdanovicivan 
> I've added the getSwappedEntity() method which returns the entity which is affected by all those issues.

You added the swapped case, but do you really need the use case?
It seems it gives users a lot of complexity.

If we prohibit it, the Entity would be simpler.

PR #3509, opened 2020-08-18
Apply phpstan level 2

samsonasik:
Fixes various things:

- docblocks 
- Config type hint 
- too many parameter in method call
- clean up unnecessary check driver instance of SessionHandlerInterface in Session::start() since it type hinted in constructor
- possible concatenation between array and string handling in `Forge::_processIndexes($table)` that can be returns string in overriden method

**Checklist:**
- [x] Securely signed commits


samsonasik:
All green 🎉

samsonasik:
There are 8 levels https://phpstan.org/

PR #3511, opened 2020-08-18
Fix handling returnSQL parameter in Model::replace()

### LINKED ISSUE #3510: Bug: Inconsistency in replace() method
Not sure if this is a bug or not I was looking at the builder replace and the model replace.

BaseBuilder line 2318: public function replace(array $set = null)

Model line 1225: public function replace($data = null, bool $returnSQL = false)

return $this->builder()->replace($data, $returnSQL);

Method call is provided 2 parameters but the method signature uses 1 parameter.

BaseBuilder takes one but the model is passing 2 to it.

michalsn:
**Description**
Fixes #3510

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -1282,7 +1282,7 @@ class Model
 			}
 		}
 
-		return $this->builder()->replace($data, $returnSQL);
+		return $this->builder()->testMode($returnSQL)->replace($data);
 	}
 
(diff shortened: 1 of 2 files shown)


samsonasik:
@michalsn please rebase latest develop

PR #3515, opened 2020-08-19
Optional pretty SQL syntax in toolbar

enix-app:
**Description**
I think this is hidden feature. Works (only) if [highlightjs.php](https://github.com/scrivo/highlight.php) installed in vendor dir.

Light mode:

before_light

after_light

Dark mode:

before_dark

after_dark

Tested with ``myth/auth`` and ``tatter/relations``.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar/Collectors/Database.php
+++ b/system/Debug/Toolbar/Collectors/Database.php
@@ -210,16 +210,21 @@ class Database extends BaseCollector
 		{
 			$sql = $query->getQuery();
 
-			foreach ($highlight as $term)
+			if( !class_exists('Highlight\Highlighter') )
 			{
-				$sql = str_replace($term, "<strong>{$term}</strong>", $sql);
+				foreach ($highlight as $term)
+				{
+					$sql = str_replace($term, "<strong>{$term}</strong>", $sql);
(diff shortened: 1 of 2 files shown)


michalsn:
There are some code style tweaks that should be fixed.

Other than that... I'm not sure if support for this specific external library should be included in the core (but this is only my opinion). I would say: "no", but maybe others have a different opinion. So... let's wait and see.

MGatner:
I agree that we cannot include a specific third-party class. Some ideas:

1. It looks like this functionality would not be hard to reproduce. We could host our own version, pared down to just what we need. 

2. Alternatively this would make a great module, e.g. `codeigniter4-toolbar-highlights`. It could include the third party package and reformatted toolbar collectors just as you have here.

enix-app:
I can't comment more because I have problem with my English.

So simply. If you have working with composer, the toolbar works fine if you have installed hljsphp. This works only in development mode. I think this is not affected in your production. So, if the package not maintained we can change the future :).

Sadly, I can't explore more the toolbar because view parser has limitations.

PR #3516, opened 2020-08-20
Apply phpstan level 3

samsonasik:
Fixes the following things:

- type declaration docblocks
- possible return type declaration in methods
- ensure key to be string in `system/Encryption/Handlers/OpenSSLHandler.php`

**Checklist:**
- [x] Securely signed commits


samsonasik:
All green 🎉

MGatner:
I'm flipping through these but gotta be honest, I'm not really looking into all the changes. I'm trusting you, the standards, and the tests that this is ultimately good for the framework (everything I see I like).

PR #3517, opened 2020-08-21
Add param type to constructor to force param type hint

devorama:
Add Config Modules param type to constructor for param type linked
picked this up when a change request was asked in
https://github.com/codeigniter4/CodeIgniter4/pull/3459
and I based my code on the routecollection constructor, thus applying the same code change as requested there
here.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -43,6 +43,7 @@ use CodeIgniter\Autoloader\FileLocator;
 use CodeIgniter\HTTP\Request;
 use CodeIgniter\Router\Exceptions\RouterException;
 use Config\Services;
+use Config\Modules;
 
 /**
  * Class RouteCollection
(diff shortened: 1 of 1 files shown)


system/Router/RouteCollection.php
```diff
@@ -233,7 +234,7 @@ class RouteCollection implements RouteCollectionInterface
 	 * @param \CodeIgniter\Autoloader\FileLocator $locator
 	 * @param \Config\Modules                     $moduleConfig
 	 */
-	public function __construct(FileLocator $locator, $moduleConfig)
+	public function __construct(FileLocator $locator,Modules $moduleConfig)
```
samsonasik, line 237: For coding standard, it need space between `,` and `Modules`, you can run `composer update` locally, so the `bash admin/setup.sh` will be executed so next change will validate coding standard first before commit.

PR #3533, opened 2020-08-24
using predefined Session property values when config->propertyName null in Session  __construct

samsonasik:
Use predefined `Session` class property values when it defined as fallback when passed config->propertyName value is null/not exists.

**Checklist:**
- [x] Securely signed commits


samsonasik:
close re-open to restart build

PR #3534, opened 2020-08-25
Add addTimestamps function to Forge

rezamoradix:
Add new method called `addTimestamps` in Forge.

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -1360,6 +1360,7 @@ class Forge
 	protected function _addTimestamps()
 	{
 		// Usually overridden by drivers
+		return $this;
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


MGatner:
There are a couple return type failures (see the Action results for code analysis). If we're going to add this it really needs to be implemented in all the supported drivers. We will also need some test cases.

michalsn:
I'm not sure about support for this in SQLite3. Proposed implementation will require a table to have a unique, single primary key to work properly. Not sure if this is a big issue but for sure has to be mentioned. 

Also, I'm not a fan of hardcoding the field names - having default values probably would work better.

paulbalandan:
Final output of PR doesn't allow flexibility of name of time fields, and can rather be implemented in the userland level.

PR #3535, opened 2020-08-25
Apply phpstan level 4

samsonasik:
Fixes the following things:

- remove unneeded `is_*` check on exact types (by type hinted or returned functions)
- remove unreacheable statements
- missing docblock declaration as possible return value in methods
- fix return declaration when there is possible multiple return types
- fix wrong exact type comparison
- remove unnecessary break after return statement

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,7 +1,5 @@
-includes:
-        - vendor/phpstan/phpstan/conf/bleedingEdge.neon
-
 parameters:
+    treatPhpDocTypesAsCertain: false
     bootstrapFiles:
         - qa-bootstrap.php
     excludes_analyse:
@@ -37,5 +35,10 @@ parameters:
         - '#Method CodeIgniter\\Validation\\ValidationInterface::run\(\) invoked with 3 parameters, 0-2 required#'
         - '#Return type \(bool\) of method CodeIgniter\\Log\\Logger::(emergency|alert|critical|error|warning|notice|info|debug|log)\(\) should be compatible with return type \(null\) of method Psr\\Log\\LoggerInterface::(emergency|alert|critical|error|warning|notice|info|debug|log)\(\)#'
         - '#Return type \(bool\) of method CodeIgniter\\HTTP\\Files\\UploadedFile::move\(\) should be compatible with return type \(CodeIgniter\\Files\\File\) of method CodeIgniter\\Files\\File::move\(\)#'
+        - '#Call to function is_null\(\) with mysqli_stmt\|resource will always evaluate to false#'
+        - '#Negated boolean expression is always (true|false)#'
     scanDirectories:
       - system/Helpers
+    dynamicConstantNames:
+        - ENVIRONMENT
+        - CI_DEBUG
--- a/system/Commands/Database/MigrateRefresh.php
+++ b/system/Commands/Database/MigrateRefresh.php
@@ -111,7 +111,7 @@ class MigrateRefresh extends BaseCommand
 
 		if (ENVIRONMENT === 'production')
 		{
-			$force = array_key_exists('f', $params) || CLI::getOption('f');
+			$force = CLI::getOption('f');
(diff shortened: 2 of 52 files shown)


phpstan.neon
```diff
@@ -1,7 +1,5 @@
-includes:
-        - vendor/phpstan/phpstan/conf/bleedingEdge.neon
```
MGatner, line 2: Curious about the removal of this configuration?

samsonasik: this was caused an error in level 4 in my local dev even the memory allocation is 2GB:

```
Reached internal errors count limit of 50, exiting...                                                                                                         
     Internal error: Child process timed out after 60.0 seconds. Try making it longer with parallel.processTimeout setting.                                        
     Child process error (exit code 255):                                                                                                                          
     Fatal error: Allowed memory size of 2147483648 bytes exhausted (tried to allocate 20480 bytes) in                                                             
     phar:///Users/samsonasik/www/CodeIgniter4/vendor/phpstan/phpstan/phpstan/vendor/nette/utils/src/Utils/Callback.php on line 58                                 
                                                                                                                                                                   
                                                                                                                                                                   
     [ErrorException]                                                                                                                                              
                                                                                                                                                                   
     Allowed memory size of 2147483648 bytes exhausted (tried to allocate 20480 bytes)                                                                             
                                                                                                                                                                   
                                                                                                                                                                   
     Fatal error: Uncaught ErrorException: Use of undefined constant FCPATH - assumed 'FCPATH' (this will throw an Error in a future version of PHP) in            
     /Users/samsonasik/www/CodeIgniter4/system/Common.php:137                                                                                                      
     Stack trace:                                                                                                                                                  
     #0 /Users/samsonasik/www/CodeIgniter4/system/Common.php(137): CodeIgniter\Debug\Exceptions->errorHandler(2, 'Use of undefine...', '/Users/samsonas...', 137,  
     Array)                                                                                                                                                        
     #1 /Users/samsonasik/www/CodeIgniter4/app/Views/errors/cli/error_exception.php(11): clean_path('phar:///Users/s...')                                          
     #2 /Users/samsonasik/www/CodeIgniter4/system/Debug/Exceptions.php(308): include('/Users/samsonas...')                                                         
     #3 /Users/samsonasik/www/CodeIgniter4/system/Debug/Exceptions.php(174): CodeIgniter\Debug\Exceptions->render(Object(ErrorException), 500)                     
     #4 /Users/samsonasik/www/CodeIgniter4/system/Debug/Exceptions.php(224): CodeIgniter\Debug\Exceptions->exceptionHandler(Object(ErrorException))                
     #5 [internal function]: CodeIgniter\Debug\Exceptions->shutdownHandler()                                                                                       
     #6 {main}                                                                                                                                                     
       thrown in /Users/samsonasik/www/CodeIgniter4/system/Common.php on line 137     
```

samsonasik:
All green 🎉

system/Commands/Database/MigrateRefresh.php
```diff
@@ -111,7 +111,7 @@ public function run(array $params)
 
 		if (ENVIRONMENT === 'production')
 		{
-			$force = array_key_exists('f', $params) || CLI::getOption('f');
+			$force = CLI::getOption('f');
```
paulbalandan, line 114: Why remove the `array_key_exists`? If this is removed, using `command('migrate:refresh -f')` will no longer be forced and will always have a prompt.

samsonasik: before if, `$params` always replaced with ['b' => 0], so it never reached.

samsonasik: except, before if, `$params` is going to be changed to : 

```php
$params['b'] = 0;
```

paulbalandan: Anyway, database commands aren't tested yet. So perhaps, the outright replacement of `$params` is a bug.

samsonasik: yes, that's probably a possible bug, this change only ensure no unreacheable code removed.

PR #3536, opened 2020-08-25
Fix the bug of getValidationRules method

phpvod:
Fix the bug of getValidationRules method


michalsn:
Can you elaborate on what bug this PR fixes?

Also, I'm not sure why phpstan raised an error for `ValidationInterface`. Hopefully, @samsonasik can help us here.

samsonasik:
as Model class uses ValidationInterface type, and the caller method is not in the interface, phpstan.neon need to be modified for that, see https://github.com/codeigniter4/CodeIgniter4/blob/29270019ff91a457b34ca32393069b2c0f06b5a0/phpstan.neon#L34

phpvod:
The fix is not perfect, please ignore.
This BUG will lose the error message defined in config\validation.php.

paulbalandan:
If I am not mistaken from the comment of OP, this can be closed?

michalsn:
I believe OP is saying that the error message will be lost when we define it in the `app/Config/Validation.php`. I have to verify if it's really a bug - maybe tomorrow. Unless someone will be faster :)

michalsn:
Nope, sorry. I can't see any errors related to the validation message stored in a config file. If you're still experiencing this bug, feel free to open an issue with details/example steps that will allow us to reproduce it.

PR #3537, opened 2020-08-25
Index.php optimization

paulbalandan:
**Description**
Some optimizations:
- replace function calls to equivalent constants to prevent overriding of those functions
- using `version_compare` to handle comparison of pre-release versions
```
var_dump('7.2-rc.1' < '7.2'); // false
var_dump(version_compare('7.2-rc.1', '7.2', '<')); // true
```
- Using unary `++` instead of `+= 1`
- and others

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -2,9 +2,9 @@
 
 // Valid PHP Version?
 $minPHPVersion = '7.2';
-if (phpversion() < $minPHPVersion)
+if (version_compare(PHP_VERSION, $minPHPVersion, '<'))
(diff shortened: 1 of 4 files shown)


samsonasik:
spark file may need to be updated as well

paulbalandan:
Thanks. I missed the spark file.

samsonasik (approved):

PR #3541, opened 2020-08-26
Apply phpstan level 5

samsonasik:
Fixes the following things:

- incorrect set error_reporting via ini_set, should be via `error_reporting` function
- ensure ini_set values to be string
- fix 4th parameter value to be 0 for preg_match instead of null
- ensure to use correct type in function values
- fix incorrect 2nd passed parameter of `lang()` function to be array.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Commands/Generators/CreateScaffold.php
+++ b/system/Commands/Generators/CreateScaffold.php
@@ -129,7 +129,7 @@ class CreateScaffold extends BaseCommand
 		];
 
 		// Call those commands!
-		$class = $params[0] ?? CLI::getOption(0);
+		$class = $params[0] ?? CLI::getOption('0');
 		$this->call('make:controller', array_merge([$class], $controllerOpts, $genOptions));
 		$this->call('make:model', array_merge([$class], $modelOpts, $genOptions));
 		$this->call('make:entity', array_merge([$class], $genOptions));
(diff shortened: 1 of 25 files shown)


samsonasik:
All green 🎉

system/Commands/Generators/CreateScaffold.php
```diff
@@ -129,7 +129,7 @@ public function run(array $params)
 		];
 
 		// Call those commands!
-		$class = $params[0] ?? CLI::getOption(0);
+		$class = $params[0] ?? CLI::getOption('0');
```
paulbalandan, line 132:
```suggestion
		$class = $params[0];
```
Good catch for this. But I think I made a bug here. this should be CLI::getArgument(0). Anyway, the params check is enough in this case.

system/Test/bootstrap.php
```diff
@@ -1,5 +1,5 @@
 <?php
-ini_set('error_reporting', E_ALL);
+error_reporting (E_ALL);
```
paulbalandan, line 2:
```suggestion
error_reporting(E_ALL);
```

system/Commands/Generators/CreateScaffold.php
samsonasik: fixed.

system/Test/bootstrap.php
samsonasik: fixed.

PR #3542, opened 2020-08-26
Standardize Action workflows

MGatner:
**Description**
GitHub web, Desktop, and mobile all display Action results differently, but in some places the name of the `job` is displayed instead of the workflow, making it confusing what the problem was during failures (see screenshots). This PR provides names to every workflow, job, and step to help make it more clear when something has gone wrong.

While working with the files, there were quite a few inconsistencies between the YAML definitions. This PR also applies a number of standardizations:
* Categorizes workflows with prefixes (deploy-, test-)
* Standardizes version to use latests major release
* Makes deployment jobs conditional on the repository
* Applies tests to all version branches (`'4.*'`)
* Removes redundant tests (e.g. on push after PR)
* Standardizes quotes: remove unnecessary, default single, double only for shell

Additionally this adds Dependabot for Composer and Action version tracking.

@paulbalandan @samsonasik @lonnieezell Please make sure I did not butcher the authorship of your workflows in the process 😮

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [x] Conforms to style guide

IMG_3719
Screen Shot 2020-08-26 at 9 56 19 AM


.github/workflows/test-phpstan.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  pull_request:
+    branches:
+      - 'develop'
+      - '4.*'
+    paths:
+      - 'app/**'
+      - 'system/**'
+
```
samsonasik, line 16: `push` need to be added in case of direct push to develop/4.*

.github/workflows/test-phpstan.yml
```diff
@@ -0,0 +1,35 @@
+# When a Pull Request is opened, perform a static
```
samsonasik, line 1: When a Pull Request is opened or direct push executed

Issue #3085, opened 2020-06-10
Bug: Running Feature Tests on multiple endpoints using filters


caswell-wc:
**Describe the bug**
I have a Feature Test class with multiple tests on the same endpoint. This endpoint uses a filter to handle authentication which then adds the authenticated user to the request so it can be used in the controller. When I run the tests individually, it works. However, if I try to run all of the tests in the test class, it gives me an error when trying to access the authenticated user that has been added to the request. 

After doing A LOT of digging, I think I have found the issue. The filters functionality alters the request object by altering `$this->request` in the `Filters` object. This works most of the time because that property points to the same object as the request that later gets passed to the controller. However, in the use case that I described above, they do not. This is because in `Codeigniter::handleRequest()` when it gets the filters object using `Services::filters()`, it is getting a shared instance of `Filters` which already exists when the second test runs. So it gets back a Filters instance whose `request` parameter no longer points to the same request as what gets passed to the controller.

A simple solution to this would be to alter Codeigniter::handleRequest() so that when it finishes running the filters, after checking to see if $possibleRedirect is a response, you add this:

    if ($possibleRedirect instanceof RequestInterface)
    {
        $this->request = $possibleRedirect;
    }    

**CodeIgniter 4 version**
v4.0.3

**Affected module(s)**
 - `CodeIgniter\CodeIgniter`
 - `CodeIgniter\Filters\Filters`

**Expected behavior, and steps to reproduce if appropriate**
1. Create a Feature Test class with 2 tests that are testing the same endpoint
2. Create a filter and register it to run on both endpoints. (In mine the filter is in general so it runs everywhere)
3. In your before() on the filter, add a new property to the request and return the request.
4. In your controller endpoint, do something to access the property that you added.
5. Run all the tests in the Feature Test class.
6. This should work but currently it will throw an error saying that the new property doesn't exist on the request object.
 
**Context**
 - OS: CentOs7
 - Web server: Apache 2.4.6
 - PHP version: 7.2.24

caswell-wc:
I did discover this morning that if you add this to your test class, it will resolve this issue for now.
    
    public function tearDown():void {
        parent::tearDown();
        Services::reset();
    }

I would still consider this a bug but this at least gives a way to get around it until it can be fixed. I will try to create a PR for this if I can find time.

MGatner:
[...]

PR #3544, opened 2020-08-26
Handling requests sent back from filters

caswell-wc:
Expands the fix of #3085 

**Description**
Filters have the ability to alter the request like adding a user object from an auth filter for example. When it does that the request gets altered implicitly by the filter because the filter's request is pointing to the same object as the request property of the Codeigniter object. If the request gets altered, the filters also  return the request into the $possibleRedirect variable. This change makes the update to the request property more explicit by setting the property with what gets returned from the filters. This ensures that if for some reason the object pointers get messed up, the request used by the controllers is the altered one.


caswell-wc:
This change worked in my local environment but it is causing a bunch of issues in the static analysis check so I will see if I can figure those out later. If anyone has an idea of what is going on there, I would appreciate the help.

michalsn (changes requested):
We would need some tests for this.

As for static analysis check - rebase should help.

system/CodeIgniter.php
```diff
@@ -101,7 +102,7 @@
 	/**
 	 * Current request.
 	 *
-	 * @var HTTP\Request|HTTP\IncomingRequest|CLIRequest
+	 * @var HTTP\Request|HTTP\IncomingRequest|CLIRequest|RequestInterface
```
MGatner, line 105: I think this should all just be `ReqiestInterface`. @samsonasik ant compelling reason why the other hints would be needed?

system/CodeIgniter.php
```diff
@@ -434,6 +435,11 @@
 			{
 				return $possibleRedirect->send();
```
MGatner, line 436: While we're here, I don't think there is any need to check for `RedirectResponse` *and* `ResponseInterface` since they are handled the same. Let's remove the first block.

system/CodeIgniter.php
samsonasik: if method caller use CLIRequest or IncomingRequest only method, that needed, if not, it can only use  RequestInterface.

samsonasik:
@caswell-wc It seems you messed up the rebase, better to make backup branch, back to this branch, reset all, rebase again, and cherry-pick your own commits.

paulbalandan:
What's the status of this PR?

MGatner:
@caswell-wc just wrapped up another outstanding PR so hopefully he is looking at this soon. This is possibly resolved by #3864.

caswell-wc:
This request got pretty messed up and was a really small change so I'm closing it and will recreate it with a fresh fork.

Issue #2185, opened 2019-09-03
Feature: RESTful Filter


jim-parry:
Add a RESTful filter to handle data representation transformations for REST requests

MGatner:
According to Jim, the intent for this was "to automatically translate between JSON & XML data in & out." (https://github.com/codeigniter4/CodeIgniter4/pull/2165). Also "once i am done, we shouldn't need the formatter explicitly. The filter will be a before & after filter for a ResourceController" (Slack).

He doesn't appear to have done any work on it but it doesn't seem to complicated.

lonnieezell:
I'm curious how much people still use XML for their RESTful work. Seems like JSON has kind of won out. Tempted to close this since nothing is happening on it.

MGatner:
I think the Filter would still be desirable outside of the XML bit. I believe the intent was for it to handle all the negotiation and formatting so that wouldn't be a responsibility of the controller/trait. 

That said, anecdotally I have yet to come across an XML API instance where I would have needed that feature.

lonnieezell:
@MGatner now that time has marched on, what are your thoughts on this? 

Reading back through it, it strikes me as more magical than CI typically tries to be. Again, tempted to close. If people decide it's  a worthy feature then a discussion or PR will show up, but I haven't seen anyone mention a need for this....

MGatner:
@lonnieezell The problem has always been the lack of expertise in this area, since Jim passed. We had a forum visitor with a lot of great advice but who couldn't implement the code - I offered to do the coding but it just hasn't happened yet. If anyone else wants to do the work, this thread and user seem to have what we need to proceed:

https://forum.codeigniter.com/thread-79278.html

lonnieezell:
I don't see what he was talking about having anything to do with this particular issue, though. 

Regarding his post, my understanding was the PUT didn't allow for multipart/form inputs, but heading over to MDN tells me I seem to be wrong. Maybe that was just a typical use-case that I had learned in the day? Not sure, but his point has merit - we shouldn't restrict to POST for form validation/files, I guess. Needs more research, though.

iRedds:
RESTful cannot be a filter. This should be the default work of the framework core.
But core is not ready for implementation right now.

MGatner:
[...]

PR #3545, opened 2020-08-26
Set response correctly for json API responses

### LINKED ISSUE #3079: Bug: Using assertJSONFragment with respond() in ResponseTrait
**Describe the bug**
I am creating a feature test with this code:
    
    $attributes = ['test'];
        
    $result = $this->get('/users');

    $result->assertOK();

    $result->assertJSONFragment($attributes);

My Users Controller looks like:

    class Users extends BaseController {

        use ResponseTrait;

        public function index() {
            $data = ['test'];

            return $this->respond($data, 200);
        }
    }

The assertJSONFragment call is throwing an error saying: `Argument #2 of PHPUnit\Framework\Assert::assertArraySubset() must be an array or ArrayAccess`. 
After A LOT of digging, I determined that this is happening because the respond function in the ResponseTrait is never setting the bodyFormat in the Response class to "json". Because of this, the data ends up being json_encoded twice. This makes it so that when the assertion decodes the data, it is just getting the string from the original encoding and is passing that string into assertArraySubset.

**CodeIgniter 4 version**
v4.0.3

**Affected module(s)**
`CodeIgniter\API\ResponseTrait` and `\CodeIgniter\Test\FeatureTestCase`

I would have expected my tests to come back as passing since the data being passed in matches what is being asserted against. 

**Context**
 - OS: CentOS7
 - Web server: Apache 2.4.6
 - PHP version 7.2.24

caswell-wc:
fixes #3079 

**Description**
In the CodeIgniter\API\ResponseTrait it defaults the format to json and everything is treated as a json but when it goes to send the response, it uses the setBody function in the response object. The problem with this is that it never sets the bodyFormat correctly in the response object and that causes issues when running tests and trying to assert different things about the json because the response json is not encoded correctly. This change makes sure that the bodyFormat will be set correctly by using setJSON function when the format is json.


caswell-wc:
The errors being given by the Static Analysis Check are not very descriptive. If anyone can help with that it would be appreciated.

samsonasik:
@caswell-wc you need to fix travis build first https://travis-ci.org/github/codeigniter4/CodeIgniter4/jobs/721516405 because your change made different tests expectation. That's the first priority.

```
There were 3 failures:

1) CodeIgniter\API\ResponseTraitTest::testRespondSets404WithNoData

Failed asserting that 'null' is null.

/home/travis/build/codeigniter4/CodeIgniter4/tests/system/API/ResponseTraitTest.php:174

2) CodeIgniter\API\ResponseTraitTest::testRespondSetsStatusWithEmptyData

Failed asserting that 'null' is null.

/home/travis/build/codeigniter4/CodeIgniter4/tests/system/API/ResponseTraitTest.php:183

3) CodeIgniter\API\ResponseTraitTest::testRespondSetsCorrectBodyAndStatus

Failed asserting that 'application/json; charset=UTF-8' starts with "text/html".

/home/travis/build/codeigniter4/CodeIgniter4/tests/system/API/ResponseTraitTest.php:193
```

For phpstan, it seems your change cause ignored error can be removed:

1. you can remove `// @phpstan-ignore-line` at `system/HTTP/Response.php` line 902
2. you can remove `// @phpstan-ignore-line` at `system/I18n/Time.php` line 560

samsonasik:
I created PR https://github.com/codeigniter4/CodeIgniter4/pull/3547 for phpstan notice. You can focus on tests.

samsonasik:
please rebase latest develop

system/API/ResponseTrait.php
```diff
@@ -134,8 +134,13 @@ public function respond($data = null, int $status = null, string $message = '')
 			$output = $this->format($data);
 		}
 
-		return $this->response->setBody($output)
-						->setStatusCode($status, $message);
+		if ($this->format === 'json') {
```
michalsn, line 137:
```suggestion
		if ($this->format === 'json')
		{
```

michalsn (changes requested):
We need some tests for this.

MGatner:
@jim-parry 's original intent was for this to be handled by a Format Filter, see https://github.com/codeigniter4/CodeIgniter4/issues/2185. I will leave it up to the team if that still seems worth doing, but I'm afraid this PR would preclude that functionality for the `ReaponseTrait`.

MGatner:
Actually looking at this change I'm afraid this would constitute a breaking change. I realize there is some expectation that `ResponseTrait` would handle the formatting for you, hence considering this a bug, but Jim was clear that was not the intent. Changing this would break current REST implementations using their own formatters, unless they were already checking for JSON/XML.

caswell-wc:
@MGatner I think I understand what you are saying about the breaking change. The real issue that I am trying to resolve here, as explained in #3079 is that if the body is formatted as json, it never actually sets Response::bodyFormat to 'json'. It stays as html. Because it never gets changed it ends up json_encoding the body of the response twice before the assertJSONFragment can json_decode it. I researched this when I first created the issue and found that if the bodyFormat gets updated correctly, it solves the problem which is why I used the setJSON function because it also sets the bodyFormat. The other option that I thought of was to create a setter on the Response object for bodyFormat and then call that in the ResponseTrait::format method when it is setting the content type on the response. Is that a more viable solution?

samsonasik:
You need rebase instead of merge. Your commits history seems already ruined now, you can do like the following:

```bash
git checkout -b temp
git checkout json-api-response
git reset --hard HEAD~52
git checkout develop

# pull latest develop via git url
git pull git@github.com:codeigniter4/CodeIgniter4.git develop

# or pull latest develop via https url
git pull https://github.com/codeigniter4/CodeIgniter4.git develop

git checkout json-api-response
git  cherry-pick 8b4e5dcbf4028f0e07d2eafe7ace92d97cd943a1
git  cherry-pick d4e307cff53eb81fc887e9761c6c61585c15713e
git  cherry-pick 3ea8a08de3e599dee38b54ea35627a85d0a39b3d
git  cherry-pick 811c17d86ee30fcf36e12a1488a4f6960ff31fd4
```

Next time, you need to rebase instead, I created a tutorial for clean rebase at https://samsonasik.wordpress.com/2015/09/16/practical-git-4-rebasing-conflicted-task-branch-against-primary-branch/

MGatner:
*Definitely implement the rebase samsonasik outlined above.*

I think I see your point now, the issue is not the formatting of the body but rather notifying the response of the format type. I will have to look at this a bit more, or hear feedback from others. Jim had this all figured out but I'm not nearly as adept at RESTful implementations as he was. I think his goal was to have the actual Controller to be "format agnostic" and have the Filter handle the formatting (unless overridden). This does seem to be in tension with the current `ResponseTrait` as it handles formatting currently. Maybe @lonnieezell can help.

caswell-wc:
I actually thought I had rebased it but I ran into issues along the way and when I look back at it, I think I know where I went wrong and it ended up merging instead. Once we figure out the best path to accomplish what we are trying to do, I will either fix that or I may just close this PR and create a new one.

MGatner:
Sounds good! Thank you for bringing this up - we'll get a little more discussion going and figure it out.

lonnieezell:
I originally had a vision of being able to have the controller negotiate the correct content type with the client in a little utopian vision once myself. IIRC, playing with some API generation proved it didn't work out as well as hoped, as utopias often don't. :) 

I prefer to allow the developer to explicitly declare the output format. The ResponseTrait seems like a good place to do that. However, we would have to provide for all supported output formats (which is only XML and JSON at this time) to be set, not just JSON. 

My other concern with the filter Jim recommended was the possibility of conversion errors across the different formats and back. It might not be a big deal but I'm concerned the number of edge cases would be a pain. Especially for a feature where 90% of the use case is going to be to spit out JSON anyway lol.

caswell-wc:
@lonnieezell So are you thinking that we stick with the change in this PR but also add logic to use the Response::setXML when the format is 'xml'?

caswell-wc:
@lonnieezell @MGatner Do either of you have anymore thoughts about this. I can see 2 options. 

1. We could adjust the respond method as I have done here so that it uses Response::setJson() if the ResponseTrait::format is 'json' or Response::setXml() if the ResponseTrait::format is 'xml'.

2. Create a setter on Response that would allow for setting Response::bodyFormat. Then either in ResponseTrait::format() or ResponseTrait::respond(), we use that setter to make sure the Response::bodyFormat is the same as ResponseTrait::format.

I like the second option better because it is allowing the ResponseTrait to format the response instead of needlessly sending the formatted response through another format method which of course gets ignored because it's already formatted. The only downside I see to the second option is that it does expose to the developer the ability to set the bodyFormat. I am not aware of a problem with this but you guys know the system better then I do.

MGatner:
@caswell-wc Could you rebase first? I'm trying to look back at my notes on this and the proposed changes but they are hard to see with all the intermediary cruft. Let me know if you need help with that command sequence - I think @samsonasik also has a handy blog on the subject.

MGatner:
Bump!

caswell-wc:
@MGatner Sorry about the confusion on this. I have been having a lot of issues with this rebase stuff. I'm pretty new to it. Anyway, I think I got it now. Please let me know if I still need to do something with it.

MGatner:
Yes looks much better! Thank you. 

I think with 4.1 around the corner we do take the easy route to try to get this done. 

> We could adjust the respond method as I have done here so that it uses Response::setJson() if the ResponseTrait::format is 'json' or Response::setXml() if the ResponseTrait::format is 'xml'.

Let's do this for now (add XML support) and then if we need another pass at the handling later we can revisit.

MGatner (approved):
Yay! Looks good. Thanks so much. Anyone else? @samsonasik ?

samsonasik (approved):

system/API/ResponseTrait.php
```diff
@@ -114,7 +114,23 @@ public function respond($data = null, int $status = null, string $message = '')
 			$output = $this->format($data);
 		}
 
-		return $this->response->setBody($output)->setStatusCode($status, $message);
+		if (! is_null($output))
+		{
+			if ($this->format === 'json')
+			{
+				return $this->response->setJSON($output)
+					->setStatusCode($status, $message);
```
paulbalandan, line 122: Can we make this as one line?

system/API/ResponseTrait.php
```diff
@@ -114,7 +114,23 @@ public function respond($data = null, int $status = null, string $message = '')
...
+			}
+
+			if ($this->format === 'xml')
+			{
+				return $this->response->setXML($output)
+					->setStatusCode($status, $message);
```
paulbalandan, line 128: One liner?

system/API/ResponseTrait.php
```diff
@@ -114,7 +114,23 @@ public function respond($data = null, int $status = null, string $message = '')
...
+			}
+		}
+
+		return $this->response->setBody($output)
+			->setStatusCode($status, $message);
```
paulbalandan, line 133: One liner?

paulbalandan (commented):
Can we make these returns as one liners?

system/API/ResponseTrait.php
caswell-wc: Fixed

system/API/ResponseTrait.php
caswell-wc: Fixed

system/API/ResponseTrait.php
caswell-wc: Fixed

paulbalandan (approved):

PR #3547, opened 2020-08-27
Fix phpstan error notice after update to v0.12.40

samsonasik:
It seems new phpstan released https://github.com/phpstan/phpstan/releases/tag/0.12.40 that make we can clean up unneeded `// @phpstan-ignore-line` at 

-  `system/HTTP/Response.php` line 902
- `system/I18n/Time.php` line 560

after update to phpstan v0.12.40

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
 		"codeigniter4/codeigniter4-standard": "^1.0",
 		"fzaninotto/faker": "^1.9@dev",
 		"mikey179/vfsstream": "1.6.*",
-		"phpstan/phpstan": "^0.12.37",
+		"phpstan/phpstan": "^0.12.40",
 		"phpunit/phpunit": "^8.5",
 		"predis/predis": "^1.1",
 		"squizlabs/php_codesniffer": "^3.3"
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -18,7 +18,7 @@
 		"codeigniter4/codeigniter4-standard": "^1.0",
 		"fzaninotto/faker": "^1.9@dev",
 		"mikey179/vfsstream": "1.6.*",
-		"phpstan/phpstan": "^0.12.37",
+		"phpstan/phpstan": "^0.12.40",
```
paulbalandan, line 21:
```suggestion
		"phpstan/phpstan": "^0.12",
```
Can we omit the patch part since we're using the caret here?

samsonasik: updated.

MGatner:
I think we'll do this PR, but in the bigger picture I'm a bit concerned that a minor point update to PHPStan appears to cause workflow failures. @samsonasik from your experience is this normal? We may need to consider using a lockfile for CI if so.

samsonasik:
@MGatner I think that's normal.

PR #3550, opened 2020-08-27
use OSKAR-phpstan for run phpstan

samsonasik:
Use OSKAR-phpstan for run phpstan . Added `.github/workflows/test-phpstan.yml` to `paths` of `pull_request` to ensure run phpstan when workflow configuration changed.
 
**Checklist:**
- [x] Securely signed commits


samsonasik:
intl extension seems need to be enabled inside the docker oskar-phpstan, not easily doable. I'm closing for now.

MGatner:
Okay, thanks for looking into it! I think the way you have it is fine, my only thought had been to save on the dependent files locally.

PR #3563, opened 2020-08-29
Refactor Format class

paulbalandan:
**Description**
Closes #3149 .

I tried to pull that PR and continuing from that but when I rebased to latest develop I got merge conflicts. I am not yet adept in solving merge conflicts, even IDE-aided, so I opted for the long road. This PR also adds tests for the new class addition and new inserted function to `Services`.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Format.php
+++ b/app/Config/Format.php
@@ -1,54 +1,59 @@
-<?php namespace Config;
+<?php
+
+namespace Config;
 
 use CodeIgniter\Config\BaseConfig;
 
 class Format extends BaseConfig
 {
-	/*
-	|--------------------------------------------------------------------------
-	| Available Response Formats
-	|--------------------------------------------------------------------------
-	|
-	| When you perform content negotiation with the request, these are the
-	| available formats that your application supports. This is currently
-	| only used with the API\ResponseTrait. A valid Formatter must exist
-	| for the specified format.
-	|
-	| These formats are only checked when the data passed to the respond()
-	| method is an array.
-	|
-	*/
+	/**
+	 * --------------------------------------------------------------------------
+	 * Available Response Formats
+	 * --------------------------------------------------------------------------
+	 *
+	 * When you perform content negotiation with the request, these are the
+	 * available formats that your application supports. This is currently
+	 * only used with the API\ResponseTrait. A valid Formatter must exist
+	 * for the specified format.
+	 *
+	 * These formats are only checked when the data passed to the respond()
+	 * method is an array.
+	 *
+	 * @var string[]
+	 */
 	public $supportedResponseFormats = [
 		'application/json',
 		'application/xml', // machine-readable XML
 		'text/xml', // human-readable XML
 	];
 
-	/*
-	|--------------------------------------------------------------------------
-	| Formatters
-	|--------------------------------------------------------------------------
-	|
-	| Lists the class to use to format responses with of a particular type.
-	| For each mime type, list the class that should be used. Formatters
-	| can be retrieved through the getFormatter() method.
-	|
-	*/
+	/**
+	 * --------------------------------------------------------------------------
+	 * Formatters
+	 * --------------------------------------------------------------------------
+	 *
+	 * Lists the class to use to format responses with of a particular type.
+	 * For each mime type, list the class that should be used. Formatters
+	 * can be retrieved through the getFormatter() method.
+	 *
+	 * @var array<string, string>
+	 */
 	public $formatters = [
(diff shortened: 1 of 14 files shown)


mostafakhudair:
Good job, my PR just missed tests, I asked for help with it but you know. ..

michalsn (approved):

app/Config/Format.php
```diff
@@ -1,54 +1,59 @@
-<?php namespace Config;
```
MGatner, line 1: We're currently all over the place on these. Let's see if @lonnieezell wants to make a call on the styling of namespaces before we change them.

system/API/ResponseTrait.php
```diff
@@ -102,6 +102,13 @@ trait ResponseTrait
 	 */
 	protected $format = 'json';
 
+	/**
+	 * Current Formatter instance. This is usually set by ResponseTrait::format
```
MGatner, line 106: We should make sure this plays nicely with the current JSON API pull request, in case that one goes through. #3545

app/Config/Format.php
paulbalandan: I agree with this. Currently it's a mix, and we need a definitive style guide for this one.

system/API/ResponseTrait.php
paulbalandan: I don't get how this will conflict with that PR. It tackles the `respond` method which does not use `$this->formatter` directly. Also, the property is uninitialized so classes using the trait will not have problems assigning a value to that.

MGatner: Okay great. That was my hunch but just wanted to make sure!

app/Config/Format.php
lonnieezell: My personal preference has always been on line 1, but it seems my IDE and lots of other programmers prefer it with a blank space after the opening php tag. I'm fine with either of those options. Not of fan of having the CI docblock above it, though. I think namespace should be the first item in the file.

So do either of you have preferences on first line vs third line?

MGatner: I've been doing it on first line but I think I picked that up from you ;) I don't use an IDE so I'm not sure how much work it is to fight against one. Are there any standards? Does our CS have options that might indicate a "default"?

paulbalandan: I am following [PSR-12](https://www.php-fig.org/psr/psr-12/).

> When the opening <?php tag is on the first line of the file, it MUST be on its own line with no other statements unless it > is a file containing markup outside of PHP opening and closing tags.

system/Format/Format.php
```diff
@@ -0,0 +1,107 @@
...
+	public function getConfig()
+	{
+		return $this->config;
+	}
+
+	/**
+	 * A Factory method to return the appropriate formatter for the given mime type.
+	 *
+	 * @param string $mime
+	 *
```
samsonasik, line 81: `@throws FormatException` can be added.

app/Config/Format.php
lonnieezell: I'm good with following the PSR on this. I had forgotten 12 got released, actually, so thanks for the reminder.

system/Format/Format.php
paulbalandan: Thanks!

app/Config/Format.php
MGatner: Oh that's great! Yes we should definitely move towards that then. The consensus is third line for namespace then?

paulbalandan: Yes. Third line for namespace if no file-level doc block. Namespace after doc block if present. This is following PSR12.

I'll take note on this on my upcoming PR on php-cs-fixer fixes.

lonnieezell: Bummer - namespace after file docblock is in PSR. Oh  well. Can't have it all :)

MGatner:
@paulbalandan is this ready?

paulbalandan:
> @paulbalandan is this ready?

I dunno. Is there more to add here? It's ready regardless. Sorry, I'm weak at colloquial English so I don't quite get what Lonnie last meant. 😂

MGatner:
I think it is good! That conversation resolved. 

***

"Bummer" = that is unfortunate

"Can't have it all" = you cannot get everything you want

MGatner:
Thanks to @paulbalandan and @mostafakhudair for this one!

PR #3565, opened 2020-08-29
apply strict in_array with pass true to 3rd parameter

samsonasik:
Ensure to make strict `in_array`.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Images/Handlers/BaseHandler.php
+++ b/system/Images/Handlers/BaseHandler.php
@@ -361,9 +361,9 @@ abstract class BaseHandler implements ImageHandlerInterface
 	{
 		// Allowed rotation values
 		$degs = [
-			(float) 90,
-			(float) 180,
-			(float) 270,
+			90.0,
+			180.0,
+			270.0,
 		];
 
 		if (! in_array($angle, $degs, true))
--- a/system/Images/Handlers/BaseHandler.php
+++ b/system/Images/Handlers/BaseHandler.php
@@ -361,12 +361,12 @@ abstract class BaseHandler implements ImageHandlerInterface
 	{
 		// Allowed rotation values
 		$degs = [
-			90,
-			180,
-			270,
+			(float) 90,
+			(float) 180,
+			(float) 270,
 		];
 
-		if (! in_array($angle, $degs))
+		if (! in_array($angle, $degs, true))
 		{
 			throw ImageException::forMissingAngle();
 		}
--- a/system/Validation/CreditCardRules.php
+++ b/system/Validation/CreditCardRules.php
@@ -245,9 +245,11 @@ class CreditCardRules
 		}
 
 		// Make sure it's a valid length for this card
-		$lengths = explode(',', $info['length']);
+		$lengths = array_map(function ($value) {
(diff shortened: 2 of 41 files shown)


system/Validation/CreditCardRules.php
```diff
@@ -245,9 +245,11 @@ public function valid_cc_number(string $ccNumber = null, string $type): bool
 		}
 
 		// Make sure it's a valid length for this card
-		$lengths = explode(',', $info['length']);
+		$lengths = array_map(function ($value) {
```
MGatner, line 248: Is this one going to be worth the extra effort to ensure strict types?

samsonasik: phpstan complain about that, I may add `@phpstan-ignore-line` ignore it if you want.

samsonasik: It seems casting to int is required:

```php
$info = [
    'length' => '1,2,3',  
];

$lengths = explode(',', $info['length']);
var_dump(in_array(1, $lengths, true));
var_dump(in_array('1', $lengths, true));

$lengths = array_map(function ($value) {
    return (int) $value;
}, explode(',', $info['length']));
var_dump(in_array(1, $lengths, true));
```

Got the following output:

```
bool(false)
bool(true)
bool(true)
```

see https://3v4l.org/lPubnY

samsonasik: another solution is cast the `strlen()` itself to string.

```php
if (! in_array((string) strlen($ccNumber), $lengths, true))
```

samsonasik: Updated to cast the `strlen()` value itself to string. It should be ok now.

system/Images/Handlers/BaseHandler.php
```diff
@@ -361,12 +361,12 @@ public function rotate(float $angle)
 	{
 		// Allowed rotation values
 		$degs = [
-			90,
-			180,
-			270,
+			(float) 90,
+			(float) 180,
+			(float) 270,
```
paulbalandan, line 366: Why not just write these as `90.0`, `180.0`, `270.0`?

samsonasik: because we can pass "90" as float ref https://3v4l.org/BFb9u

MGatner: I don't understand the difference. Isn't `(float) 90 === 90.0`?

system/Validation/CreditCardRules.php
MGatner: I meant if the strict version of `in_array` was worth the trouble of a custom array mapping, for performance and code clarity.

system/Images/Handlers/BaseHandler.php
samsonasik: oh, you're correct.  I will update to `90.0`, not in front of pc right now

samsonasik: fixed.

samsonasik:
All green :tada:

PR #3567, opened 2020-08-30
Refactor RedirectResponse

paulbalandan:
**Description**
Continuing on and closes #3279 
Additional:
- Removed unneeded `empty` check before `foreach` for arrays. `foreach` won't execute if passed array is empty.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/RedirectResponse.php
+++ b/system/HTTP/RedirectResponse.php
@@ -51,24 +51,24 @@ class RedirectResponse extends Response
 	 * Sets the URI to redirect to and, optionally, the HTTP status code to use.
 	 * If no code is provided it will be automatically determined.
 	 *
-	 * @param string       $uri    The URI to redirect to
-	 * @param integer|null $code   HTTP status code
+	 * @param string       $uri
+	 * @param integer|null $code
 	 * @param string       $method
 	 *
 	 * @return $this
 	 */
 	public function to(string $uri, int $code = null, string $method = 'auto')
 	{
-		// If it appears to be a relative URL, then convert to full URL
-		// for better security.
+		// If its a relative URL, convert to full URL for a better security.
 		if (strpos($uri, 'http') !== 0)
 		{
-			$url = current_url(true)->resolveRelativeURI($uri);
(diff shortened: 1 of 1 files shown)


system/HTTP/RedirectResponse.php
```diff
@@ -170,12 +177,7 @@ public function with(string $key, $message)
 	 */
 	public function withCookies()
 	{
-		$cookies = service('response')->getCookies();
-
-		if (empty($cookies))
-		{
-			return $this;
-		}
+		$cookies = Services::response()->getCookies();
```
MGatner, line 180: Why not put this right in the `foreach`? We don't need the variable again.

system/HTTP/RedirectResponse.php
```diff
@@ -195,12 +199,7 @@ public function withCookies()
 	 */
 	public function withHeaders()
 	{
-		$headers = service('response')->getHeaders();
-
-		if (empty($headers))
-		{
-			return $this;
-		}
+		$headers = Services::response()->getHeaders();
```
MGatner, line 202: Same.

system/HTTP/RedirectResponse.php
```diff
@@ -63,13 +63,14 @@ public function to(string $uri, int $code = null, string $method = 'auto')
...
 		{
-			$url = current_url(true)->resolveRelativeURI($uri);
-			$uri = (string)$url;
+			$uri = (string) current_url(true)->resolveRelativeURI($uri);
 		}
 
 		return $this->redirect($uri, $method, $code);
 	}
 
+	//--------------------------------------------------------------------
```
MGatner, line 72: I'm not a fan of the separators between all functions. I think they serve some visual clarity as logical dividers between types of methods in large classes. We probably need to decide how we want to handle them framework-wide for consistency, but for now I think not adding them is better IMO.

paulbalandan: This is not my addition. But I'll remove them. Personally not a fan either.

PR #3570, opened 2020-08-30
Use null coalescing to routes resolution in FeatureTestTrait::call

paulbalandan:
**Description**
This change will allow us to not be constrained in using `FeatureTestTrait` within the context only of `FeatureResponse`. We can also use this on other test cases, like `CIUnitTestCase`. This also resolves phpstan's error "Ternary operation always return true."

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -167,7 +167,7 @@ trait FeatureTestTrait
 		$request = $this->populateGlobals($method, $request, $params);
 
 		// Make sure the RouteCollection knows what method we're using...
-		$routes = $this->routes ?: Services::routes(); // @phpstan-ignore-line
+		$routes = $this->routes ?? Services::routes();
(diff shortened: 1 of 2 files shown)


paulbalandan:
Sure do.

paulbalandan:
I'm confused. `HomeTest` does not fail locally.

```
PS C:\Users\P\Desktop\Web Dev\CodeIgniter4> vendor/bin/phpunit -v --filter HomeTest

Runtime:       PHP 7.4.9 with Xdebug 2.9.6

Time: 58.65 seconds, Memory: 52.00 MB

OK (1 test, 1 assertion)
```

michalsn:
@paulbalandan It was failing because we use `Services::reset()` in other tests. And it wipes out the default routes config.

michalsn (approved):

PR #3571, opened 2020-08-30
Allow string for parameter $obj in ReflectionHelper::getAccessibleRefProperty()

samsonasik:
There is `is_object($obj)` check and its otherwise condition using `ReflectionClass`, that make it possible to pass a string parameter.

**Checklist:**
- [x] Securely signed commits


MGatner:
I think API deploy failed because the other deployment interlaces with this one. Should self-correct on next deployment.

PR #3575, opened 2020-08-31
fix getNext and getPrevious links

sherifsalah:
**Description**
changed getPrevious to current - 1 to get previous page instead of first -1 that gets first page. same in getNext.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Pager/PagerRenderer.php
+++ b/system/Pager/PagerRenderer.php
@@ -171,11 +171,11 @@ class PagerRenderer
 
 		if ($this->segment === 0)
 		{
-			$uri->addQuery($this->pageSelector, $this->first - 1);
+			$uri->addQuery($this->pageSelector, $this->current - 1);
 		}
 		else
 		{
-			$uri->setSegment($this->segment, $this->first - 1);
+			$uri->setSegment($this->segment, $this->current - 1);
 		}
 
 		return (string) $uri;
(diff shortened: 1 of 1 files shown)


paulbalandan (commented):
Have you tried calling first `$pager->setSurroundCount()` before calling `$pager->links()`? Current implementation is pretty much desired.

sherifsalah:
when i tried to create a custom template already played with `$pager->setSurroundCount()` and it has no effect, i found out it was not the problem and it was actually always giving my the first or last page anyway .. when i modified the code to the suggested one it works flawlessly (and i guess it should be logically curr+ or curr-. maybe i'm missing something here? can you please elaborate on how current implementation is required?

paulbalandan:
@sherifsalah For example, I have the current page set at `5` and I want to have the links before and after (the `surroundCount`) to be `2` each, that will give me something like this:
```
3   |  4  |  5  |  6  |  7
```

1. If you are looking for going to page 3, use `getPrevious`
2. If you are looking for going to page 4, use `getPreviousPage`
3. If you are looking for going to page 6, use `getNextPage`
4. If you are looking for going to page 7, use `getNext`

michalsn:
@sherifsalah I think you're looking for `getPreviousPage()` and `getNextPage()` as mentioned above. I know this might be confusing but we had to do this way because of the compatibility between RC and final release. Please read more about it in the user guide https://codeigniter4.github.io/userguide/libraries/pagination.html (at the bottom of the page).

paulbalandan:
@michalsn, I think this can be closed?

michalsn:
Yes, I think it's sorted out, thanks.

PR #3576, opened 2020-08-31
remove `else` on return/throw/continue early

samsonasik:
When there is already `return` or `throw`  or `continue` early in previous `if`, we can remove `else`.

**Checklist:**
- [x] Securely signed commits


michalsn (approved):
It simplify the code. I like the idea.

PR #3583, opened 2020-09-02
Allow fallback when make:scaffold's argument is missing

paulbalandan:
**Description**
Without the fallback, we'll get the Error `Undefined offset: 0`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Generators/CreateScaffold.php
+++ b/system/Commands/Generators/CreateScaffold.php
@@ -129,7 +129,7 @@ class CreateScaffold extends BaseCommand
 		];
 
 		// Call those commands!
-		$class = $params[0];
+		$class = $params[0] ?? CLI::getSegment(2);
 		$this->call('make:controller', array_merge([$class], $controllerOpts, $genOptions));
 		$this->call('make:model', array_merge([$class], $modelOpts, $genOptions));
 		$this->call('make:entity', array_merge([$class], $genOptions));


michalsn:
In what situation this will happen? Shouldn't this index (`$params[0]`) always be set?

paulbalandan:
In the terminal, try this: `make:scaffold` (without the class name argument). It will throw an ErrorException

michalsn:
I believe we always need a file/class name when we use generators and because of that, if `$params[0]` is empty we should produce a "nice" error and then just quit. Fallback to `getSegment()` gives us nothing.

paulbalandan:
Yes, a class name is required for generators. In this setup, since we have the fallback `getSegment` in place, in case no class name is given, `null` would be passed as the class name. Then, the "real" generators will handle the `null` class name.

paulbalandan:
If the generators down the line needs a class name, then they will prompt for one. If name is optional, they will not complain.

michalsn:
Well, running this code gives me a result:
```cli
Created file: APPPATH/Controllers/Make:scaffold.php
...
```
I don't think this is something we want.

paulbalandan:
Ah yes. The segment counting again. 😂 It should be the 2nd segment.

michalsn (approved):

PR #3584, opened 2020-09-02
Use view files for generator commands

paulbalandan:
**Description**
Use view stubs for the templates required by generator commands for future maintainability.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Generators.php
+++ b/app/Config/Generators.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Config;
+
+use CodeIgniter\Config\BaseConfig;
+
+class Generators extends BaseConfig
+{
+	/**
+	 * --------------------------------------------------------------------------
+	 * Generator Commands' Views
+	 * --------------------------------------------------------------------------
+	 *
+	 * This array defines the mapping of generator commands to the view files
+	 * they are using. If you need to customize them for your own, copy these
+	 * view files in your own folder and indicate the location here.
+	 *
+	 * You will notice that the views have special placeholders enclosed in
+	 * curly braces `{...}`. These placeholders are used internally by the
+	 * generator commands in processing replacements, thus you are warned
+	 * not to delete them or modify the names. If you will do so, you may
+	 * end up disrupting the scaffolding process and throw errors.
+	 *
+	 * YOU HAVE BEEN WARNED!
+	 *
+	 * @var array<string, string>
+	 */
+	public $views = [
+		'make:command'      => 'CodeIgniter\\Commands\\Generators\\Views\\command.tpl.php',
+		'make:controller'   => 'CodeIgniter\\Commands\\Generators\\Views\\controller.tpl.php',
+		'make:entity'       => 'CodeIgniter\\Commands\\Generators\\Views\\entity.tpl.php',
+		'make:filter'       => 'CodeIgniter\\Commands\\Generators\\Views\\filter.tpl.php',
+		'migrate:create'    => 'CodeIgniter\\Commands\\Generators\\Views\\migration.tpl.php',
+		'make:model'        => 'CodeIgniter\\Commands\\Generators\\Views\\model.tpl.php',
+		'make:seeder'       => 'CodeIgniter\\Commands\\Generators\\Views\\seed.tpl.php',
+		'session:migration' => 'CodeIgniter\\Commands\\Generators\\Views\\session_migration.tpl.php',
+	];
+}
--- a/app/Config/Generators.php
+++ b/app/Config/Generators.php
@@ -30,7 +30,7 @@ class Generators extends BaseConfig
 		'make:controller'   => 'CodeIgniter\\Commands\\Generators\\Views\\controller.tpl.php',
 		'make:entity'       => 'CodeIgniter\\Commands\\Generators\\Views\\entity.tpl.php',
 		'make:filter'       => 'CodeIgniter\\Commands\\Generators\\Views\\filter.tpl.php',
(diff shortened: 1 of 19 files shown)


michalsn:
I like this. 

What do you think about adding an array variable to the configuration file that would define the used views? This way making "small" changes to the generated files would become very simple. We would just "tell" the generator to use our view file.

paulbalandan:
Okay, but in what config file will that be put.?

michalsn:
I guess `Config/Migrations.php` would be the right place. But we would also have to hard-code the view names in our generator classes - for an easy upgrade path (fallback).

paulbalandan:
OK. So I guess you mean that the user-configurable views are for the migrations and seeders only.

michalsn:
Yeah... I just realized we don't have any config files related to commands. Maybe @MGatner will have an idea of how to handle this regarding an easy upgrade path? And maybe he will validate if this is a good idea in the first place :)

MGatner:
I think this is great! Just to be clear on the Config file... something like the `$views` to property we use in Myth:Auth Config? https://github.com/lonnieezell/myth-auth/blob/develop/src/Config/Auth.php

`Generators` have grown a lot since their inception - I think it would be fine to make a new Config file for them, just so long as we have a fallback for when it is missing from **app/**.

michalsn:
Yes, I was thinking exactly about the way Myth:Auth is doing it 😄

And yes, we need a fallback for the view name for every generator.

MGatner:
My beef with the way Myth:Auth does it (with the disclaimer that I'm pretty sure I wrote that) is that there is no way to overwrite one view without supplying the whole array. Likewise, if a view were added down the road any existing extensions in **app/** would have to add that. I wonder if we can come up with a "smart" Config file that does some merging, kind of like Routes?

paulbalandan:
My question is... should ALL the generators be given a configuration? My concern is the other generators have complicated code and have in-set text replacements in the templates.

MGatner:
> should ALL the generators be given a configuration

Do you mean a configuration, or a view? I don't see a problem with treating Generators like a Controller: most of the time they will have a view, maybe a view with some variables passed in. But occasionally there is output to display that still makes sense to handle in the Controller.

paulbalandan:
> > should ALL the generators be given a configuration
> 
> Do you mean a configuration, or a view? I don't see a problem with treating Generators like a Controller: most of the time they will have a view, maybe a view with some variables passed in. But occasionally there is output to display that still makes sense to handle in the Controller.

Yes, yes. I'm referring to views.

MGatner:
This is looking good. Can we get an alias of `make:migration` to `migrate:create`? I guess just a new class that calls the other. 

Also is `{@inheritDoc}` a legit PHPDocs block?

MGatner (changes requested):
*See comment*

paulbalandan:
> This is looking good. Can we get an alias of `make:migration` to `migrate:create`? I guess just a new class that calls the other.

Are we leaning to deprecating use of `migrate:create` in favor of `make:migration`??

> Also is `{@inheritDoc}` a legit PHPDocs block?

Yes, it signals PHPDoc to copy the parent method's doc block into the inherited method.

MGatner:
Nothing we've discussed officially but I think it makes sense for consistency to move toward the other. Maybe it would make sense to move the code and leave `migrate:create` as a deprecated alias. What do you think?

paulbalandan:
Yes, thinking the same.

michalsn:
As for `{@inheritDoc}` I believe that docblock is automatically inherited.
> Currently some applications have DocBlocks containing just the {@inheritDoc} inline tag to indicate that their complete contents should be inherited. This usage breaks with the PHPDoc Standard as summaries cannot contain inline tags and inheritance is automatic; you do not need to define a special tag for it.
However, it does make clear that an element has been explicitly documented (and thus not forgotten). As such we are working to include a new (normal) tag in the PHPDoc Standard @inheritDoc that will serve that purpose.

paulbalandan:
> As for `{@inheritDoc}` I believe that docblock is automatically inherited.
> 
> > Currently some applications have DocBlocks containing just the {@inheritdoc} inline tag to indicate that their complete contents should be inherited. This usage breaks with the PHPDoc Standard as summaries cannot contain inline tags and inheritance is automatic; you do not need to define a special tag for it.
> > However, it does make clear that an element has been explicitly documented (and thus not forgotten). As such we are working to include a new (normal) tag in the PHPDoc Standard @inheritdoc that will serve that purpose.

However, our API generator complains.
> 
**Commands/Generators/CreateModel.php**
Type | Line | Description
-- | -- | --
error | 45 | No summary for class \CodeIgniter\Commands\Generators\CreateModel
error | 88 | No summary for method getClassName()
error | 100 | Argument $rootNamespace is missing from the Docblock of getNamespacedClass
error | 100 | Argument $class is missing from the Docblock of getNamespacedClass
error | 100 | No summary for method getNamespacedClass()
error | 105 | No summary for method getTemplate()
error | 119 | Argument $template is missing from the Docblock of setReplacements
error | 119 | Argument $class is missing from the Docblock of setReplacements
error | 119 | No summary for method setReplacements()

michalsn:
Oh... okay - I didn't check to be honest. In that case, all is good.

This is documented very strangely in their website though.

MGatner (approved):
Very nice deprecation. Is this all set? Looks good to me.

paulbalandan:
> Very nice deprecation. Is this all set? Looks good to me.

Feel free. 😁

PR #3585, opened 2020-09-02
Use CLI::getSegment(2) to refer to the first argument of generator command

paulbalandan:
**Description**
I've noticed that `getSegment`'s code deducts 1 from the index parameter before checking which made me realize that it is one-based instead of zero-based. This PR updates `GeneratorCommand::getClassName()` to use that.

In retrospect, why was it that way?

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -947,6 +947,8 @@ class CLI
 	 *  // segment(3) is 'three', not '-f' or 'anOption'
 	 *  > php spark one two -f anOption three
 	 *
+	 * **IMPORTANT:** The index here is one-based instead of zero-based.
+	 *
 	 * @param integer $index
 	 *
 	 * @return mixed|null
--- a/system/CLI/GeneratorCommand.php
+++ b/system/CLI/GeneratorCommand.php
@@ -152,7 +152,8 @@ abstract class GeneratorCommand extends BaseCommand
 	 */
 	protected function getClassName(): string
 	{
-		$name = $this->params[0] ?? CLI::getSegment(0);
+		$name = $this->params[0] ?? CLI::getSegment(2);
+
 		return $name ?? '';
 	}
 
@@ -203,6 +204,7 @@ abstract class GeneratorCommand extends BaseCommand
 	protected function getRootNamespace(): string
 	{
 		$rootNamespace = $this->params['n'] ?? CLI::getOption('n') ?? APP_NAMESPACE;
+
 		return trim(str_replace('/', '\\', $rootNamespace), '\\');
 	}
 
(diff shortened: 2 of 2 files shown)


paulbalandan:
I think this should be getSegment(2).

michalsn (approved):
Just waiting for tests to pass before merging.

MGatner (changes requested):
Could we get a test case showing what this fixes? How has this been working so far?

paulbalandan:
> Could we get a test case showing what this fixes? How has this been working so far?

Unfortunately, I think we cannot setup an automated test for this. The concrete classes extending GeneratorCommand all require class names as argument, so the empty string given here will trigger a `CLI::prompt` which is currently untestable. You may test them manually.

Analyzing how CLI parses the arguments and stores them, here is an illustration:
```php
php spark make:controller blog -bare

CLI::getSegment(0) =>> undefined // since we are accessing the -1 index
CLI::getSegment(1) =>> 'make:command' // first segment after the name of invoking program is the name of the command
CLI::getSegment(2) =>> 'blog' // next segments would be the arguments passed to the command
CLI::getOption('bare') =>> true // since it is set as a flag
```

paulbalandan:
> Could we get a test case showing what this fixes? How has this been working so far?

Unfortunately, I think we cannot setup an automated test for this. The concrete classes extending GeneratorCommand all require class names as argument, so the empty string given here will trigger a `CLI::prompt` which is currently untestable. You may test them manually.

Analyzing how CLI parses the arguments and stores them, here is an illustration:
```php
php spark make:controller blog -bare

CLI::getSegment(0) =>> undefined // since we are accessing the -1 index
CLI::getSegment(1) =>> 'make:command' // first segment after the name of invoking program is the name of the command
CLI::getSegment(2) =>> 'blog' // next segments would be the arguments passed to the command
CLI::getOption('bare') =>> true // since it is set as a flag
```

John-Betong:
I have just been working on my dynamic menu system and had to count($uri->getSegments()) to prevent the following error being thrown:
#### // $tmpSub = $uri->getSegment(2) // MAYBE THROW ERROR  
Request URI segment is out of range:

michalsn:
@John-Betong You can use `setSilent()` instead: https://codeigniter4.github.io/userguide/libraries/uri.html#uri-segments

John-Betong:
@michalsn,
> @John-Betong You can use `setSilent()` instead: https://codeigniter4.github.io/userguide/libraries/uri.html#uri-segments

That's new to me and could come in handy some time in the future.

I don' think I like hiding exceptions, far prefer to know what is happening.

I prefer testing for results and using if/else/endif statements.

PR #3587, opened 2020-09-02
[DEV] Enhancements to dev workflows

paulbalandan:
**Description**
- Run actions only on main repo
- Add phpstan analysis to `precommit`
- Strictly check only for PHP files ending in `.php`. Will exclude `.phpcs`, `.php_cs`, etc.
- Run linting if there are PHP files to check

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/admin/pre-commit
+++ b/admin/pre-commit
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 PROJECT=`php -r "echo dirname(dirname(dirname(realpath('$0'))));"`
-STAGED_FILES_CMD=`git diff --cached --name-only --diff-filter=ACMR HEAD | grep \\\\.php`
+STAGED_FILES_CMD=`git diff --cached --name-only --diff-filter=ACMR HEAD | grep \\\\.php$`
 
 # Determine if a file list is passed
 if [ "$#" -eq 1 ]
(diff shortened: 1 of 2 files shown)


admin/pre-commit
```diff
@@ -14,17 +14,35 @@ then
...
+if [ "$FILES" != "" ]
+then
+	echo "Running PHPStan..."
+	# Run on whole codebase
+	./vendor/bin/phpstan analyse
+
 	if [ $? != 0 ]
 	then
-		echo "Fix the error before commit."
+		echo "Fix the phpstan error(s) before commit."
```
samsonasik, line 42: phpstan doesn't fix the file, probably change wording with "Analyze code"

paulbalandan: I don't get it. Yes, phpstan doesn't fix the file. But if there are violations, it will echo them out as errors. So it must be fixed or resolved.

samsonasik: oh, you're correct :+1:

michalsn (approved):

MGatner:
I deliberately left the syntax-check workflows in place for non-organization repos, since they don't require any special permissions and may catch some of these errors before they run against our Action. I think it is fine to leave them there, unless you have a compelling reason not to?

MGatner:
Nice work on the PHPStan precommit! I think that will help catch a lot of issues before they hit us.

paulbalandan:
> I deliberately left the syntax-check workflows in place for non-organization repos, since they don't require any special permissions and may catch some of these errors before they run against our Action. I think it is fine to leave them there, unless you have a compelling reason not to?

My reason I think is not so compelling -- just for the bit of consistency with other actions.

MGatner:
Let's leave them running then. I think it is probably not common that they trigger on an individual's repo but in the rare cases it does it could catch something helpful. 

Idea, what if we update the `branches` restriction to be "not master" (have to research this, maybe `!master`?) so when people are pushing to their feature branches it will trigger the syntax checks?

MGatner:
Looking at **test-phpstan.yml** I notice that we are missing the Composer cache in that one. It should use the same build as **test-phpunit.yml** so we want them to use the same cache. Could you please add that section to this PR? I think it is a logical place to do it.

MGatner (approved):
Thanks!

PR #3589, opened 2020-09-02
Refinements to Autoloader

paulbalandan:
**Description**
Mini refinements
- Ensure paths have correct ending directory separators.
- When removing namespaces, ensure the namespace is set.
- Add more tests to Autoloader and remove redundant test

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -39,6 +39,9 @@
 
 namespace CodeIgniter\Autoloader;
 
+use Config\Autoload;
+use Config\Modules;
+
 /**
  * CodeIgniter Autoloader
  *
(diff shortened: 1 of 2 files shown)


system/Autoloader/Autoloader.php
```diff
@@ -100,11 +103,11 @@ class Autoloader
 	 * the valid parts that we'll need.
 	 *
 	 * @param \Config\Autoload $config
-	 * @param \Config\Modules  $moduleConfig
+	 * @param \Config\Modules  $modules
```
samsonasik, line 106: `@param Modules $modules` can be used, no longer require fqcn

paulbalandan: My preference has always been FQCN for phpdocs since that will help my IDE in resolving names. If others would want short class names, I'll oblige.

michalsn: I have no preferences in that matter. But maybe we should standardize it somehow in the codebase.

michalsn (approved):

system/Autoloader/Autoloader.php
MGatner: I prefer to use the shortname for anything that has a `use` statement, but I don't use an IDE so it doesn't actually have negative consequences. I will say that PHPStan accepts the shortname.

MGatner (approved):

PR #3602, opened 2020-09-04
Put database driver in separate folder

mostafakhudair:


mostafakhudair:
@samsonasik could you help to solve this phpstan issue

michalsn:
Sorry that I didn't comment on this earlier - I see you have done some heavy work here. Personally I'm not sure if we want to move all of the drivers to a separate folder. 

We have 2 new DB drivers coming and I would like to avoid a situation where creators of these new drivers would have to update everything "just because".

Changing the namespace may be a problem. Also, you have renamed the SQLite3 driver to SQLite. It would be a BC change and besides that, there was something called SQLite (v2), so that's the reason why we should use the explicit name `SQLite3`.

mostafakhudair:
Thanks for your feedback, My opinion is to make one folder for database drivers for more organize as you said there is 2 db drivers under development after that we will have 5 driver folders under systems/Database/ folder it will be mess right there thats why I think it will be nice to put all drivers into one folder just like ci3, maybe 2 db drivers developers can change their work and fix conflicts and maybe i should wait until they finished their work and then make changes.

about SQLite3 I thought there is no need to use driver version (3) but i also may revert changes of renaming it if it will bc change,

qury:
Hi guys, i have very little to be done for sqlsrv before i make a PR, i think it would be best to merge the Oracle and Sqlsrv driver before we refactor the code.

PR #3605, opened 2020-09-05
using EXIT_ERROR constant in CodeIgniter::bootstrapEnvironment()

samsonasik:
There is `EXIT_ERROR` constant that can be used.

**Checklist:**
- [x] Securely signed commits


MGatner:
API workflow failure is likely from interlaced merged during workflow execution.

PR #3606, opened 2020-09-05
URI class update for better handling of URLs with sub-folders

### LINKED ISSUE #3603: Bug: `current_url()` helper returns wrong url with slash between host and port
**Describe the bug**
The `current_url` helper returns wrong url when using port (adds a slash between hostname and port).
When having a baseUrl with a port (usually on dev environment):

In `.env`:
```ini
app.baseURL="http://localhost:8080/"
```
```php
echo current_url();
// http://localhost/:8080/
```

**CodeIgniter 4 version**
Latest `develop` (https://github.com/codeigniter4/CodeIgniter4/commit/87dbd2963555be88ce1576be992e23f23ecf323d)

**Affected module(s)**
`url_hepler` function in `/system/Helpers/url_helper.php`

When setting a port in baseURL, `current_url()` should return `http://localhost:8080/`

**Context**
 - OS: Debian GNU/Linux 10 (buster)
 - Local dev server with spark
 - PHP version 7.2

michalsn:
**Description**
This PR fixes some issues with `URI` class when `baseURL` contains a sub-folder.

Now we get the same results as when we would get when calling `Services::request()`. Before, the results were polluted with `baseURI` path.

Also fixes #3603

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -596,11 +596,11 @@ class URI
 		// If hosted in a sub-folder, we will have additional
 		// segments that show up prior to the URI path we just
 		// grabbed from the request, so add it on if necessary.
-		$baseUri  = new URI(config(\Config\App::class)->baseURL);
-		$basePath = trim($baseUri->getPath(), '/') . '/';
+		$baseUri  = new self(config(\Config\App::class)->baseURL);
+		$basePath = rtrim($baseUri->getPath(), '/') . '/';
 		$path     = $this->getPath();
 
-		if ($basePath !== '/' && strpos($path, $basePath) !== 0)
+		if (! empty($basePath) && $basePath !== '/' && strpos($path, $basePath) !== 0)
 		{
 			$path = $basePath . $path;
 		}
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -600,7 +600,7 @@ class URI
 		$basePath = rtrim($baseUri->getPath(), '/') . '/';
 		$path     = $this->getPath();
 
-		if (! empty($basePath) && $basePath !== '/' && strpos($path, $basePath) !== 0)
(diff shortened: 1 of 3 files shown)


system/HTTP/URI.php
```diff
@@ -593,8 +593,20 @@ public function getTotalSegments(): int
 	 */
 	public function __toString(): string
 	{
+		// If hosted in a sub-folder, we will have additional
+		// segments that show up prior to the URI path we just
+		// grabbed from the request, so add it on if necessary.
+		$baseUri  = new URI(config(\Config\App::class)->baseURL);
```
samsonasik, line 599: should can use `new self(...)`:

```suggestion
              $baseUri  = new self(config(\Config\App::class)->baseURL);
```

michalsn: Thanks!

system/HTTP/URI.php
```diff
@@ -593,8 +593,20 @@ public function getTotalSegments(): int
...
 	public function __toString(): string
 	{
+		// If hosted in a sub-folder, we will have additional
+		// segments that show up prior to the URI path we just
+		// grabbed from the request, so add it on if necessary.
+		$baseUri  = new self(config(\Config\App::class)->baseURL);
+		$basePath = rtrim($baseUri->getPath(), '/') . '/';
+		$path     = $this->getPath();
+
+		if (! empty($basePath) && $basePath !== '/' && strpos($path, $basePath) !== 0)
```
samsonasik, line 603: `$basePath` never be empty as always added with `/` so the empty check can be removed.

```suggestion
              if ($basePath !== '/' && strpos($path, $basePath) !== 0)
```

MGatner (approved):
I haven't tested this myself but the tests and code looks good. If you think it is ready please proceed!

PR #3607, opened 2020-09-05
Use DocBlocks to document Config classes' properties

paulbalandan:
**Description**
In CI3, we used arrays to manage configurations. That's why we only used multiline comments (`T_COMMENT`) to document each config. However, in CI4 we have shifted to config classes. Class properties are structural elements that needs to have its documentation in DocBlocks (`T_DOC_COMMENT`). In here, we can add more information on each config variable by defining its type, adding linkable references, etc. This also helps static code analysis tools detect errors in usage by knowing the types used. Plus side also is that in IDEs the definition of the config class' properties will be shown when you hover your mouse over the property. These benefits are deprived on the use of regular comments.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -1,278 +1,434 @@
-<?php namespace Config;
+<?php
+
+namespace Config;
 
 use CodeIgniter\Config\BaseConfig;
 
 class App extends BaseConfig
 {
-
-	/*
-	|--------------------------------------------------------------------------
-	| Base Site URL
-	|--------------------------------------------------------------------------
-	|
-	| URL to your CodeIgniter root. Typically this will be your base URL,
-	| WITH a trailing slash:
-	|
-	|	http://example.com/
-	|
-	| If this is not set then CodeIgniter will try guess the protocol, domain
-	| and path to your installation. However, you should always configure this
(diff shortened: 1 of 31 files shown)


paulbalandan:
Latest test failure is on Predis, which is unrelated here. I think we should somehow modify or remove the metadata test for both Redis and Predis, as current tests results are very volatile.

MGatner:
This was a lot of work! I hope you had some help from Rector or some automation IDE. I think this is great but I would like to look through them a bit more before I merge, or unless someone else can complete a review. 

Re: P/Redis tests I agree. It has been on my list to stand up a GitHub Action for PHPUnit testing and then bring up the idea of dropping Travis. Not sure if that would change consistency of results but it would be a good time to assess things like that.

paulbalandan:
@MGatner any feedback on this PR?

Changes here are 98% changing the pipes `|` to asterisks `*`, 1% adding `@var` type hints, and 1% importing classes used instead of using FQCN.

MGatner (approved):
Yes I think this is ready to go!

paulbalandan:
Thanks for that! 👌

PR #3612, opened 2020-09-06
Port MSSql Database driver from CI3

qury:
**Description**
Port of Sqlsrv driver.  #503 

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


MGatner:
This is awesome! Thank you so much @qury - a huge first step. 

Next thing to do is work with PHPStan to resolve and code analysis issues. You can run this locally from the repo like this: `./vendor/bin/phpstan analyze`. When I get a chance I will look through things myself, as will the other members of the dev team (hopefully someone else has solid MSSQL experience).

paulbalandan (commented):
I have no experience in MSSql but I have some reviews to share.

Importing `CI_DEBUG` in `use const CI_DEBUG;` is I think won't give us a performance boost as it is not optimized by the PHP compiler? It is not a native PHP constant.

Also, there is the inconsistent use of global namespaces and functions import. There are classes where you used `use stdClass;` eliminating the use of `new \stdClass` but there are some with the leading namespace prefix, like `throw new \BadMethodCallException`.

qury:
Closing in favor of #3714

PR #3613, opened 2020-09-07
read composer.json to get minimal php version required in public/index.php

samsonasik:
Read composer.json to get minimal php version required in `public/index.php`. It can reduce path to check when upgrade php requirement.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -1,12 +1,13 @@
 <?php
 
 // Valid PHP Version?
-$minPHPVersion = '7.2';
+preg_match('#(\d+(.\d+)?(.\d+)?)#', json_decode(file_get_contents(__DIR__ . '/../composer.json'))->require->php, $match);
+$minPHPVersion = current($match);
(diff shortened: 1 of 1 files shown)


MGatner:
I like the idea but I have two concerns:

1. This adds a file read and decode to every single call, could this impact performance?

2. I know some people still move **index.php** and assuming that **composer.json** is a directory up will break those scenarios or add one more place to change directory paths. 

Others' thoughts?

MGatner:
A couple more thoughts...

3. When existing projects upgrade to a new version of the framework with updated PHP requirements the minimum version won't be at composer.json or index.php but rather in vendor/. Composer installations will throw system requirement errors but manual installations won't know. We might need to think about an additional version check somewhere in system/

4. Related, this change seems most relevant to manual installations, since Composer will already be validating PHP for installations, but composer.json is not otherwise required for manual installations.

samsonasik:
Yes, performance will be slower, just reduce manual change. I am closing it then.

PR #3618, opened 2020-09-07
Allow GeneratorCommand's sortImports to be configurable

paulbalandan:
**Description**
I am making a spark command that publishes a module's language files to userland or to other language folders. The problem is the method `GeneratorCommand::sortImports` is interfering with the file created.

Initially, the idea for the sort imports method is to automatically sort the class imports a class template is using before outputting the final product. This is not a problem if the file contains classes that uses class imports. For language files, however, we are using a return array with translations defined as index-value pairs. The `sortImports` algorithm checks for word matches starting with `use blah blah`, collects those matches, sorts them, then output them. If in the translation array I have the pairs:
```php
return [
    'foo' => 'This one will use {0} on foo.',
    'bar' => 'This will use bar on {0}.',
];
```
`sortImports` will collect the `use` phrases and sorts them, destroying the array on output, something like:

```php
return [
    'foo' => 'This one will
    'bar' => 'This will
use {0} on foo.',
use bar on {0}.',
];
```

This PR tries to resolve this issue by setting an internal `$sortImports` flag variable (default to `true`). Then child generator classes not wanting to have their imports sorted or will affect the file generation (such as my case) will just have to turn off this flag and `GeneratorCommand` will return the template unscathed.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/GeneratorCommand.php
+++ b/system/CLI/GeneratorCommand.php
@@ -78,6 +78,15 @@ abstract class GeneratorCommand extends BaseCommand
 	];
 
 	/**
+	 * Whether to sort class imports.
+	 *
+	 * @internal
+	 *
+	 * @var boolean
+	 */
+	private $sortImports = true;
+
+	/**
 	 * The params array for easy access by other methods.
 	 *
 	 * @var array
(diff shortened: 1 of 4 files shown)


MGatner (commented):
I think this is a good change regardless, but couldn't we also change the regex only to match "use" at the beginning of a line? That should eliminate any errant "use" words unrelated to class references.

paulbalandan:
> I think this is a good change regardless, but couldn't we also change the regex only to match "use" at the beginning of a line? That should eliminate any errant "use" words unrelated to class references.

Hmm. Yes, we should also do that, too.

paulbalandan:
Just realized that matching `use` at beginning of phrases will solve my use case on language files. Nice one!

OK then. The sort import flag will just be used to allow sorting to be optional on part of child generator commands.

MGatner:
Sounds good! Only place I could see an issue is with Traits, but so far we don't generate anything with traits so not an immediate issue.

PR #3621, opened 2020-09-08
Write console long options using '--'

paulbalandan:
**Description**
I just realized that we are unknowingly accepting options to be written either starting with `-` or `--`. This is because on trimming the options we are using `ltrim` which greedily trims the dashes. So:
```php
echo ltrim('-allow', '-'); // allow
echo ltrim('--allow', '-'); // allow
```

This PR suggests using the "longopts-style" for writing long options, i.e. using `--`. This change will not affect previous usages because of the `ltrim` above and `mb_strpos($arg, '-') === 0` will match for both `-` and `--`. Additionally, I intentionally did not change the unit tests for commands that uses `-` to show that it will still work as intended.

For `getOptionString`, I added optional parameters that will change the existing parsing to use the longopts style. It is default to `false` to match with current behavior. Then, I noticed that `CLI::getOptionString` and `CLIRequest::getOptionString` behaves differently in terms of trimming the final output string. The former gives it raw while the latter outputs it trimmed. Oh well.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -1017,9 +1017,12 @@ class CLI
 	 * Returns the options as a string, suitable for passing along on
 	 * the CLI to other commands.
 	 *
+	 * @param boolean $useLongOpts Use '--' for long options?
+	 * @param boolean $trim        Trim final string output?
+	 *
 	 * @return string
 	 */
-	public static function getOptionString(): string
+	public static function getOptionString(bool $useLongOpts = false, bool $trim = false): string
 	{
 		if (empty(static::$options))
 		{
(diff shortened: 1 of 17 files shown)


MGatner (approved):
Looks good! People will still be able to use the short version on existing commands, correct? `db:migrate -all`?

paulbalandan:
> Looks good! People will still be able to use the short version on existing commands, correct? `db:migrate -all`?

Yes. Either usage will work. `migrate -all` , `migrate --all`

michalsn (approved):

PR #3629, opened 2020-09-09
Split Session Configuration

mostafakhudair:
Move session configuration into new Config\Session Class


MGatner:
What happened here? This looked like a good approach, though probably something we would have to assess for breaking changes. I think adding configuration files to **app/** is generally okay though.

mostafakhudair:
I don't figure out where is the issue, Travis build doesn't clear to me

paulbalandan:
Try to close reopen your PR to restart Travis build.

paulbalandan:
I do not know what exit code 9 for phpunit means so I guess you should try to test your changes locally first. If this is php's exit code channeled through phpunit, then this could be a `SIGKILL` signal that resulted from multiple fatal errors. Not sure though.

mostafakhudair:
Anything new? I'm thinking of pulling a new request

mostafakhudair:
Anything new? I'm thinking of pulling a new request

mostafakhudair:
Anything new? I'm thinking of pulling a new request

paulbalandan:
Please do as you will. 😁

PR #3631, opened 2020-09-10
In method find returns uniform data

Sosko:
If find is used like: `$model->where("column", "5")->find()`, the ci4 returned array like [0->["id"=>1, "column"=>5]]. But mine opinion is, that it will be better to return only the first row

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -426,7 +426,7 @@ class Model
 		{
 			$row = $builder->get();
 
-			$row = $row->getResult($this->tempReturnType);
+			$row = $row->getFirstRow($this->tempReturnType);
 		}
 
 		$eventData = [


MGatner:
I like the concept but this is not something we can do in the framework. It would make a good addition to your `BaseModel`, where you detect certain kinds of singleton results based on database keys and flatten the array.

Sosko:
I don't get it
In your documentation:

> find()
> Returns a single row where the primary key matches the value passed in as the first parameter:

So why not returns directly first row?

michalsn:
@Sosko Because it's designed to work this way. Please read also other lines of the user guide regarding the `find()` method.
> If no parameters are passed in, will return all rows in that model’s table, effectively acting like findAll(), though less explicit.

You're not passing any parameters. The proposed change would break functionality of the Model for many people.

In general, we're always open for improvements to the Model class (and CodeIgniter) but sadly this particular change cannot be accepted - sorry.

Sosko:
I'm really apologise to you, I'm mist that line.
But my opinion, that the `find` method, should always return only one row.
Thank you for your time

PR #3632, opened 2020-09-10
Fix Filters' discoverFilters on class redeclaration

paulbalandan:
In a new project I have defined `Config => app/Config/` in my `composer.json`'s `autoload.psr-4` so that phpstan will not complain later of undiscovered symbols. However, I am getting class redeclaration error for `Config\Filters`, particularly on the `discoverFilters` method of `CodeIgniter\Filters\Filters`.

image

This is because `Services::locator` will locate two `Config\Filters` in my case, 1) as defined by `Config\Autoload` and 2) as defined by Composer. Now, the discoverFilters method will just check if the path to the file is not equal to `APPPATH . 'Config/Filters.php`. Since the Composer one is a relative route usually having the `vendor` in its path, it will fail the check and will be `include`d. Since it is really identical to the main filter config, the PHP error will be thrown for same class name declared.

This PR modifies the conditional include to check the fully qualified class name of the files instead. Other changes are just aesthetic fixes.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -121,9 +121,7 @@ class Filters
 	 *
 	 * @var \Config\Modules
 	 */
-	protected $moduleConfig;
-
-	//--------------------------------------------------------------------
+	protected $modules;
 
 	/**
 	 * Constructor.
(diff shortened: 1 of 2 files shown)


michalsn (approved):

paulbalandan:
Requesting review from @MGatner

MGatner (approved):

PR #3634, opened 2020-09-10
Multiple Environment Manage

ajmeireles:

**Description**
Created a capacity to manipulate multiple environment for projets who use a single core system for a lot of applications, suck as: api/, app/ and want to manipulate the environment.

**Checklist:**
- [ ] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


app/Config/Constants.php
```diff
@@ -25,6 +25,16 @@
...
 
+/*
+ | --------------------------------------------------------------------
+ | Environment File
+ | --------------------------------------------------------------------
+ |
+ | This define allows you to control the name of the environment file so that 
+ | you can manipulate more than one environment attached to a single core system.
+ */
+defined('ENVIRONMENT_FILE') || define('ENVIRONMENT_FILE', '.env');
```
michalsn, line 36:
```suggestion
 | This allows you to control the name of the env file so that you can handle 
 | more than one file when dealing with multiple environments in one installation.
 */
defined('ENV_FILE') || define('ENV_FILE', '.env');
```

system/Config/DotEnv.php
```diff
@@ -58,9 +58,9 @@ class DotEnv
...
 	 *
 	 * @param string $path
-	 * @param string $file
+	 * @param string $file The environment file name
 	 */
-	public function __construct(string $path, string $file = '.env')
+	public function __construct(string $path, string $file = ENVIRONMENT_FILE)
 	{
 		$this->path = rtrim($path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $file;
 	}
```
michalsn, line 66:
```suggestion
	public function __construct(string $path, string $file = null)
	{
		if ($file === null)
		{
			$file = defined('ENV_FILE') ? ENV_FILE : '.env';
		}
		$this->path = rtrim($path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $file;
	}
```

user_guide_src/source/general/configuration.rst
```diff
@@ -88,7 +88,9 @@ Environment Variables and CodeIgniter
 CodeIgniter makes it simple and painless to set Environment Variables by using a “dotenv” file. The term comes from the file name, which starts with a dot before the text “env”.
 
 CodeIgniter expects **.env** to be at the root of your project alongside the ``system``
-and ``app`` directories. There is a template file distributed with CodeIgniter that’s
+and ``app`` directories and you is able to change the name of .env file if you need manipulate
+a multiples applications connected with a single core system but with different environments at 
+``app/Config/Constants.php``. There is a template file distributed with CodeIgniter that’s 
 located at the project root named **env** (Notice there’s no dot (**.**) at the start?).
```
michalsn, line 94:
```suggestion
and ``app`` directories. If you want to handle multiple environments in one installation you can change
the default file name via the **ENV_FILE** constant located in ``app/Config/Constants.php``. There is a template
file distributed with CodeIgniter that’s located at the project root named **env** (Notice there’s no dot (**.**) at the start?).
```

michalsn (changes requested):
I think this is a good addition but we need some small changes and also the `env` file has to be restored.

user_guide_src/source/general/configuration.rst
ajmeireles: > I think this is a good addition but we need some small changes and also the `env` file has to be restored.

Yes! Thank you. It's my first contribuiton to the CI4. I will follow your code types from now.

ajmeireles (commented):
Changes for a better code types.

John-Betong:
May I suggest this simpler version if it is not too late:

file: index.php
` define('LOCALHOST', 'localhost'===$_SERVER['SERVER_NAME']);

# SWAP .env file ==> DEFAULT TO ONLINE and NO CHANGE
   $tmpEnvFile = '../envLOCAL_SPECIAL';
   if(LOCALHOST && file_exists($tmpEnvFile) ):
      $ok = copy($tmpEnvFile, '../.env');
   endif; 

`

michalsn:
@John-Betong As I understand, the point of this PR is to be able to use different `.env` files when working on multiple applications (https://codeigniter4.github.io/userguide/general/managing_apps.html) with one installation of CI.

michalsn:
Hmm... actually I think it should work out of the box. @ajmeireles Did you follow the user guide for setting the correct directory structure to support multiple applications with one installation (https://codeigniter4.github.io/userguide/general/managing_apps.html)? Or maybe I'm missing the point of this PR?

John-Betong:
> @John-Betong As I understand, the point of this PR is to be able to use different `.env` files when working on multiple applications (https://codeigniter4.github.io/userguide/general/managing_apps.html) with one installation of CI.

#### If I understand how the problem of setting different /emv file:
`
### SET APPLICATION PATHS
$tmpAPP  = 'dick'; // [ tom. dick, harry, codeigniter ]
  switch($tmpAPP):
    case 'tom':
      $tmpEnvFile = '../envdick';
    break;

    case 'dick':
      $tmpEnvFile = '../envdick';
    break;

    case 'harry':
      $tmpEnvFile = '../envharry';
    break;

    default: 'codeigniter':
      $tmpEnvFile = '../envLOCAL';
    break;
  endswitch;  
$pathsPath = FCPATH . '../' .$tmpAPP .'/Config/Paths.php';
  
# DEFAULT TO ONLINE and NO CHANGE
  if(LOCALHOST && file_exists($tmpEnvFile) ):
    $ok = copy($tmpEnvFile, '../.env');
  endif; 
`

paulbalandan:
> Hmm... actually I think it should work out of the box. @ajmeireles Did you follow the user guide for setting the correct directory structure to support multiple applications with one installation (https://codeigniter4.github.io/userguide/general/managing_apps.html)? Or maybe I'm missing the point of this PR?

After much contemplation on this, I believe the user guide pretty much tackled the handling of multiple apps, which covers the changes being sought by this PR. Since one would be maintaining multiple "root" for their multiple apps, then placing the `.env` file to each root would mean it will still work because `DotEnv` will look in ROOTPATH which must have been correctly set already by each application. Also, I think the changes made here was only allowing the parseable env file to be named whatever one wants and not sticking to default `.env`.

paulbalandan:
> > @John-Betong As I understand, the point of this PR is to be able to use different `.env` files when working on multiple applications (https://codeigniter4.github.io/userguide/general/managing_apps.html) with one installation of CI.
> 
> #### If I understand how the problem of setting different /emv file:
> `
> 
> #### SET APPLICATION PATHS
> $tmpAPP = 'dick'; // [ tom. dick, harry, codeigniter ]
> switch($tmpAPP):
> case 'tom':
> $tmpEnvFile = '../envdick';
> break;
> 
> ```
> case 'dick':
>   $tmpEnvFile = '../envdick';
> break;
> 
> case 'harry':
>   $tmpEnvFile = '../envharry';
> break;
> 
> default: 'codeigniter':
>   $tmpEnvFile = '../envLOCAL';
> break;
> ```
> 
> endswitch;
> $pathsPath = FCPATH . '../' .$tmpAPP .'/Config/Paths.php';
> 
> #### DEFAULT TO ONLINE and NO CHANGE
> if(LOCALHOST && file_exists($tmpEnvFile) ):
> $ok = copy($tmpEnvFile, '../.env');
> endif;
> `

As I understand this solution, I believe it will not be scalable. I mean this solution means that for every app that I would add to my multi-CI4 app, then it is imperative for me to add a new `case` statement that would handle that addition. So if it the long run I got 20 apps using a shared system, then there must be 20 `case`s plus a default.

michalsn:
> > Hmm... actually I think it should work out of the box. @ajmeireles Did you follow the user guide for setting the correct directory structure to support multiple applications with one installation (https://codeigniter4.github.io/userguide/general/managing_apps.html)? Or maybe I'm missing the point of this PR?
> 
> After much contemplation on this, I believe the user guide pretty much tackled the handling of multiple apps, which covers the changes being sought by this PR. Since one would be maintaining multiple "root" for their multiple apps, then placing the `.env` file to each root would mean it will still work because `DotEnv` will look in ROOTPATH which must have been correctly set already by each application. Also, I think the changes made here was only allowing the parseable env file to be named whatever one wants and not sticking to default `.env`.

Yes, I agree with you. The path described in the user guide works correct and should be used to handle multiple apps with a single CI installation.

PR #3642, opened 2020-09-14
Remove the autoinc logic as it dont

devorama:
seem to work, it creates a constraint that is not autoinc
each time. By doing it the same way as the other databases
one can use it now as a autoinc and not get
not null errors when inserting data

This is how definition looks after change
prent

Data is inserted with seeder still correctly

prent

The table without auto inc looks like all the others did before change
prent
prent

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


devorama:
Note not totally related but this doc does confirm that we can use the auto inc as is
from sqllites website

https://sqlite.org/autoinc.html

michalsn (approved):
Good catch. I think the current implementation would be accurate for SQLite2 only.

PR #3643, opened 2020-09-14
Unify QA bootstrap files

MGatner:
**Description**
This PR adjusts our PHPUnit bootstrap file to work with PHPStan, and hopefully any other external code processor. Not only does this allow us to unify the process, but modules and projects should now be able to use the same one without needing to define their own.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -11,7 +11,11 @@ defined('CI_DEBUG') || define('CI_DEBUG', true);
 
 // Often these constants are pre-defined, but query the current directory structure as a fallback
 defined('HOMEPATH') || define('HOMEPATH', realpath(rtrim(getcwd(), '\\/ ')) . DIRECTORY_SEPARATOR);
-$source = is_dir(HOMEPATH . 'app') ? HOMEPATH : 'vendor/codeigniter4/codeigniter4/';
+$source = is_dir(HOMEPATH . 'app') 
+       ? HOMEPATH 
+       : (is_dir('vendor/codeigniter4/framework/') 
+               ? 'vendor/codeigniter4/framework/' 
+               : 'vendor/codeigniter4/codeigniter4/');
 defined('CONFIGPATH') || define('CONFIGPATH', realpath($source . 'app/Config') . DIRECTORY_SEPARATOR);
 defined('PUBLICPATH') || define('PUBLICPATH', realpath($source . 'app/Config') . DIRECTORY_SEPARATOR);
 unset($source);
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -11,13 +11,13 @@ defined('CI_DEBUG') || define('CI_DEBUG', true);
 
 // Often these constants are pre-defined, but query the current directory structure as a fallback
 defined('HOMEPATH') || define('HOMEPATH', realpath(rtrim(getcwd(), '\\/ ')) . DIRECTORY_SEPARATOR);
-$source = is_dir(HOMEPATH . 'app') 
-       ? HOMEPATH 
-       : (is_dir('vendor/codeigniter4/framework/') 
-               ? 'vendor/codeigniter4/framework/' 
-               : 'vendor/codeigniter4/codeigniter4/');
+$source = is_dir(HOMEPATH . 'app')
+	   ? HOMEPATH
+	   : (is_dir('vendor/codeigniter4/framework/')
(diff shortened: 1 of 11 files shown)


MGatner:
@samsonasik would you take a look at the PHPStan implementation?

@paulbalandan would you verify this is acceptable for Windows?

system/Test/bootstrap.php
```diff
@@ -7,6 +7,14 @@
 // Make sure it recognizes that we're testing.
 $_SERVER['CI_ENVIRONMENT'] = 'testing';
 define('ENVIRONMENT', 'testing');
+defined('CI_DEBUG') || define('CI_DEBUG', true);
+
+// Often these constants are pre-defined, but query the current directory structure as a fallback
+defined('HOMEPATH') || define('HOMEPATH', realpath(rtrim(getcwd(), '\\/ ')) . DIRECTORY_SEPARATOR);
+$source = is_dir(HOMEPATH . 'app') ? HOMEPATH : 'vendor/codeigniter4/codeigniter4/';
```
samsonasik, line 14: should be `vendor/codeigniter4/framework` instead for release path, not development path:

```suggestion
$source = is_dir(HOMEPATH . 'app') ? HOMEPATH : 'vendor/codeigniter4/framework/';
```

Need note addition, if using next version ( 4.0.5+), require `--prefer-source` for testing purpose if rely to `vendor/codeigniter4/framework/` directory.

samsonasik: or better like this:

```php
$source = is_dir(HOMEPATH . 'app') 
       ? HOMEPATH 
       : (is_dir('vendor/codeigniter4/framework/') 
               ? 'vendor/codeigniter4/framework/' 
               : 'vendor/codeigniter4/codeigniter4/');
```

samsonasik: For `--prefer-source` on testing purpose reason is on PR https://github.com/codeigniter4/CodeIgniter4/pull/3438

MGatner: I forgot it could be the release version, good catch. I like that second solution, covers all bases.

samsonasik:
needs note additon on `Test/bootstrap.php`:

```php
// when rely on vendor/codeigniter4 directory, you need to use `--prefer-source`,
// @see https://github.com/codeigniter4/CodeIgniter4/pull/3438 ,
// eg: on getting `app` directory from 'vendor/codeigniter4/framework` when you create distributable module
```

MGatner:
@samsonasik I would be more inclined to add a "bootstrap" section to the User Guide. Because these constants are defined in **phpunit.xml.dist** this really only applies to PHPStan and any scenario where a develop deliberately removes the variables. In other words, I think this is mostly for us in developer the framework and modules. That said I do like have more transparency but I don't know that putting it in the bootstrap file itself makes sense.

Anyone else have thoughts?

samsonasik:
In the user guide is ok

system/Test/bootstrap.php
```diff
@@ -7,6 +7,18 @@
...
+
+// Often these constants are pre-defined, but query the current directory structure as a fallback
+defined('HOMEPATH') || define('HOMEPATH', realpath(rtrim(getcwd(), '\\/ ')) . DIRECTORY_SEPARATOR);
+$source = is_dir(HOMEPATH . 'app') 
+       ? HOMEPATH 
+       : (is_dir('vendor/codeigniter4/framework/') 
+               ? 'vendor/codeigniter4/framework/' 
+               : 'vendor/codeigniter4/codeigniter4/');
+defined('CONFIGPATH') || define('CONFIGPATH', realpath($source . 'app/Config') . DIRECTORY_SEPARATOR);
+defined('PUBLICPATH') || define('PUBLICPATH', realpath($source . 'app/Config') . DIRECTORY_SEPARATOR);
```
paulbalandan, line 20:
```suggestion
defined('PUBLICPATH') || define('PUBLICPATH', realpath($source . 'public') . DIRECTORY_SEPARATOR);
```

paulbalandan:
Putting in the user guide would be nicer.

michalsn (approved):
Looks good. +1 for adding the user guide section.

system/Test/bootstrap.php
MGatner: Wooooowww glad you caught that! The module I tested with must not have used PUBLICPATH at all. Copy-paste fail. 🤦‍♂️

PR #3645, opened 2020-09-15
Implement getTotal item on pager

agungsugiarto:
**Description**
This PR implement getTotal() item on pager.
Example:
```php
$model = new SampleModel();

$data = $model->paginate();
$total = $model->pager->getTotal();
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Pager/PagerInterface.php
+++ b/system/Pager/PagerInterface.php
@@ -112,6 +112,17 @@ interface PagerInterface
 	//--------------------------------------------------------------------
 
 	/**
+	 * Returns the total number of items in data store.
+	 *
+	 * @param string $group
+	 *
+	 * @return integer
+	 */
+	public function getTotal(string $group = 'default'): int;
+
+	//--------------------------------------------------------------------
+
+	/**
 	 * Returns the total number of pages.
 	 *
 	 * @param string $group
--- a/system/Pager/PagerInterface.php
+++ b/system/Pager/PagerInterface.php
@@ -112,17 +112,6 @@ interface PagerInterface
 	//--------------------------------------------------------------------
 
 	/**
-	 * Returns the total number of items in data store.
-	 *
-	 * @param string $group
-	 *
-	 * @return integer
-	 */
-	public function getTotal(string $group = 'default'): int;
-
-	//--------------------------------------------------------------------
-
-	/**
 	 * Returns the total number of pages.
 	 *
 	 * @param string $group
--- a/system/Pager/Pager.php
+++ b/system/Pager/Pager.php
@@ -263,6 +263,22 @@ class Pager implements PagerInterface
 	//--------------------------------------------------------------------
 
 	/**
+	 * Returns the total number of items in data store.
+	 *
+	 * @param string $group
+	 *
+	 * @return integer
+	 */
+	public function getTotal(string $group = 'default'): int
+	{
+		$this->ensureGroup($group);
+
+		return $this->groups[$group]['total'];
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
 	 * Returns the total number of pages.
 	 *
 	 * @param string $group
--- a/tests/system/Pager/PagerTest.php
+++ b/tests/system/Pager/PagerTest.php
@@ -205,6 +205,13 @@ class PagerTest extends \CodeIgniter\Test\CIUnitTestCase
 		$this->assertEquals(1, $this->pager->getPageCount());
 	}
 
+	public function testGetTotalCorrectValue()
+	{
+		$this->pager->store('foo', 3, 12, 70);
+
+		$this->assertEquals(70, $this->pager->getTotal('foo'));
+	}
+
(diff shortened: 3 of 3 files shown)


system/Pager/PagerInterface.php
```diff
@@ -111,6 +111,17 @@ public function setPath(string $path, string $group = 'default');
...
 	//--------------------------------------------------------------------
 
+	/**
+	 * Returns the total number of items in data store.
+	 *
+	 * @param string $group
+	 *
+	 * @return integer
+	 */
+	public function getTotal(string $group = 'default'): int;
```
paulbalandan, line 121: Adding this to the interface would be breaking to those already implementing their own Pager via this interface.

agungsugiarto: Can you give me example related what is breaking change for this. Thanks

michalsn:
I think previously I was encouraging people to use `getDetails()` method instead but since this is "popping up" again, then maybe it's a good idea to add this method after all.

system/Pager/PagerInterface.php
paulbalandan: For example, in v4.0.4 I have implemented my own Pager.

```php
class MyPager implements PagerInterface
{
    // my code logic where I have implemented the interface's existing methods
}
```

Now, on upgrade to a later version, I would still be expecting this to work. Instead, I would get a `Fatal error` since I have not implemented the new method `getTotal`.

michalsn: Adding a new method to the interface means that this method has to be implemented in every class that uses this interface. If someone wrote his own pager class then after update this change will break his code.

agungsugiarto: @paulbalandan @michalsn yes, will be breaking change if implement own Pager, but getTotal() is nice for spesific to get total instead of using getDetails()['total']

Example im implement [Fractal PaginatorInterface](https://github.com/thephpleague/fractal/blob/c1042fd31fd33d86fea7a3a54f643ad08905dc96/src/Pagination/PaginatorInterface.php#L19) will be need spesific get total. Here is my own implement:
```php
<?php

namespace App\Transformers;

use CodeIgniter\Pager\PagerInterface;
use League\Fractal\Pagination\PaginatorInterface;

class CodeIgniterPaginatorAdapter implements PaginatorInterface
{
    /**
     * The pagiantor.
     * 
     * @var \CodeIgniter\Pager\PagerInterface
     */
    protected $paginator;
    
    /**
     * Create a new illuminate pagination adapter.
     *
     * @param \CodeIgniter\Pager\PagerInterface $paginator
     * @return void
     */
    public function __construct(PagerInterface $paginator)
    {
        $this->paginator = $paginator;
    }

    /**
     * @inheritdoc
     */
    public function getCurrentPage()
    {
        return $this->paginator->getCurrentPage();
    }

    /**
     * @inheritdoc
     */
    public function getLastPage()
    {
        return $this->paginator->getLastPage();
    }

    /**
     * @inheritdoc
     */
    public function getTotal()
    {
        // instead of using
        // return $this->paginator->getDetails()['total'];
        
        return $this->paginator->getTotal();
    }

    /**
     * @inheritdoc
     */
    public function getCount()
    {
        return $this->paginator->getPageCount();
    }

    /**
     * @inheritdoc
     */
    public function getPerPage()
    {
        return $this->paginator->getPerPage();
    }

    /**
     * @inheritdoc
     */
    public function getUrl($page)
    {
        return $this->paginator->getPageURI($page);
    }

    /**
     * Get the paginator instance.
     * 
     * @return \CodeIgniter\Pager\Pager
     */
    public function getPaginator()
    {
        return $this->paginator;
    }
}
```

paulbalandan: Yes. I know what you are trying to achieve here. It would be a shortcut to what you have said. But we cannot discount the possibility that there is a developer out there who have implemented their own pager class. Why? Because it is part of the public API. *That developer* would be cursing on us once his perfectly working application will get errors after he made an update, considering that this is not a BC-breaking version update.

michalsn: I'm not against adding `getTotal()` to the `Pager` class but we should skip adding this method to the `PagerInterface` simply because this type of change is not worth introducing a BC.

michalsn (changes requested):
Please remove `getTotal()` method from `PagerInterace` and I'll be happy to merge this.

agungsugiarto:
> Please remove `getTotal()` method from `PagerInterace` and I'll be happy to merge this.

Wait, before this let me hear mr @samsonasik  opinion to add interface getTotal().

samsonasik:
Yes, please remove `getTotal()` method from the `PagerInterface`.

michalsn (approved):
Thank you for your first contribution to this project!

Looking forward to more :)

samsonasik (approved):

PR #3652, opened 2020-09-16
Clarify model find* event data

MGatner:
**Description**
Currently the `beforeFind` and `afterFind` events can be triggered by three different methods, each of which has different data conditions which are difficult for the event recipient to parse. For example, if `$eventData['data']` contains an array is it an array of results or a single result array?

This PR defines "singletons" - model `find*` methods intended to return a single result array - and adds the boolean value to the event data. Additionally this adds the calling method to `trigger()`'s event data for `afterFind`, same as is currently available for `beforeFind`.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -393,10 +393,17 @@ class Model
 	 */
 	public function find($id = null)
 	{
+		$singleton = is_numeric($id) || is_string($id);
+
 		if ($this->tempAllowCallbacks)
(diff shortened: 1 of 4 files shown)


michalsn (approved):
Useful changes.

Issue #3655, opened 2020-09-18
Bug: The lifetime of the CSRF check cookie cannot be set to 0 (Session).


lf-uraku-yuki:
**Describe the bug**
The lifetime of the CSRF check cookie cannot be set to 0 (Session).
Probably because the current time and the set value are added.

\Config\App.php
```
public $CSRFExpire = 0;
public $CSRFRegenerate = false;
```

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
CodeIgniter\Security\Security::CSRFSetCookie

**Expected behavior, and steps to reproduce if appropriate**
If 0 is specified, it will not be added to time () .

**Context**
 - OS: Windows 10Pro 2004 64bit
 - Web server: php spark serve
 - PHP version: 7.4.10 Thread Safe 64bit

PR #3656, opened 2020-09-18
If CSRFExpire specifies 0, set 0 as it is

lf-uraku-yuki:
**Description**
If CSRFExpire specifies 0, set 0 as it is.

fix: #3655

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -288,7 +288,11 @@ class Security
 	 */
 	public function CSRFSetCookie(RequestInterface $request)
 	{
-		$expire        = time() + $this->CSRFExpire;
+		if ($this->CSRFExpire !== 0) {
+			$expire        = time() + $this->CSRFExpire;
+		} else {
+			$expire        = $this->CSRFExpire;
+		}
 		$secure_cookie = (bool) $this->cookieSecure;
 
 		if ($secure_cookie && ! $request->isSecure())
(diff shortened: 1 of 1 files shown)


system/Security/Security.php
```diff
@@ -288,7 +288,11 @@ public function CSRFVerify(RequestInterface $request)
 	 */
 	public function CSRFSetCookie(RequestInterface $request)
 	{
-		$expire        = time() + $this->CSRFExpire;
+		if ($this->CSRFExpire !== 0) {
+			$expire        = time() + $this->CSRFExpire;
+		} else {
+			$expire        = $this->CSRFExpire;
+		}
```
michalsn, line 295:
```suggestion
		$expire        = $this->CSRFExpire === 0 ? $this->CSRFExpire : time() + $this->CSRFExpire;
```

michalsn (changes requested):
Good catch, but can we simplify this a bit?

system/Security/Security.php
lf-uraku-yuki: Certainly that is more appropriate. to fix and commit.

michalsn:
Thank you and congratulations on your first contribution to this project!

We looking forward to more :)

PR #3658, opened 2020-09-19
Fix checking for redirect in FeatureResponse

### LINKED ISSUE #3654: Bug: isRedirect does not respond depending on how redirects are set up.
**Describe the bug**
HTTP Testing ->isRedirect() returns true for RedirectReseponse returned by 
`return redirect()->to(url)`
but false for Response returned by 
`return $this->response->redirect(url, method, status_code)`
Returns.

If this is the intended specification, there is no problem, but it is a little difficult to understand.

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
CodeIgniter\Test\FeatureResponse

The controller returns a Response that has executed the method redirect.

**Context**
 - OS: Windows 10 Pro 2004 64bit
 - Web server: php spark serve
 - PHP version: 7.4.10 Thread Safe 64bit

michalsn:
**Description**
Fixes #3654

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/ControllerResponse.php
+++ b/system/Test/ControllerResponse.php
@@ -196,13 +196,15 @@ class ControllerResponse {
 	}
 
 	/**
-	 * Returns whether or not the Response was a redirect response
+	 * Returns whether or not the Response was a redirect or RedirectResponse
 	 *
 	 * @return boolean
 	 */
 	public function isRedirect(): bool
 	{
-		return $this->response instanceof RedirectResponse;
+		return $this->response instanceof RedirectResponse
+			|| $this->response->hasHeader('Location')
+			|| $this->response->hasHeader('Refresh');
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


MGatner (approved):
I think this is a good change, but I've also been under the assumption that a Response which is a redirect will always be a RedirectResponse. Should we start enforcing that, e.g. for this issue have `Response::redirect()` actually return a new `RedirectResponse`?

michalsn:
At first, I wanted to do this that way but then I realized that it won't work if we want to keep the current functionality. `Response::redirect()` method is not the "final" one.

But maybe I'm missing something so I'm open to any ideas on how to handle this using `RedirectResponse`.

MGatner:
I see what you mean, because `redirect()` is just updating the instance and could return whatever class implemented it. I am still hesitant about a response being a redirect without being a `RedirectResponse` but since this is already the current behavior it is beyond the scope of this issue/PR anyway.

MGatner (approved):

PR #3662, opened 2020-09-20
Fix Message return types

MGatner:
**Description**
A number of `Message` methods have docblocks for `@return Message` which makes static analysis angry when a child class is returned. This PR switches to `$this` and specifies a `self` return type on the function. I also cleaned up the file a bit but nothing that would affect code.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/HTTP/Message.php
+++ b/system/HTTP/Message.php
@@ -46,7 +46,6 @@ use CodeIgniter\HTTP\Exceptions\HTTPException;
  */
 class Message
 {
-
 	/**
 	 * List of all HTTP request headers.
 	 *
@@ -89,7 +88,6 @@ class Message
 	protected $body;
 
 	//--------------------------------------------------------------------
-	//--------------------------------------------------------------------
 	// Body
 	//--------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


MGatner:
Apparently the updated return types resolved this PHPStan error we were ignoring:
> #Call to an undefined method CodeIgniter\\HTTP\\Message::setStatusCode\(\)#

I removed that line and updated the file format to tabs (recommendation from https://ne-on.org) and alphabetized the sequences to be easier to find.

MGatner:
@samsonasik If you would please check over the PHPStan changes for me!

samsonasik:
it seems ok 👍

system/HTTP/Message.php
```diff
@@ -351,6 +328,21 @@ public function prependHeader(string $name, string $value)
...
+	 * Takes a header name in any case, and returns the
+	 * normal-case version of the header.
+	 *
+	 * @param string $name
+	 *
+	 * @return string
+	 */
+	protected function getHeaderName(string $name): string
+	{
+		$lower_name = strtolower($name);
```
michalsn, line 341: I know you're just moving around this method but can we take this opportunity and change the variable name from `$lower_name` to `$lowerName`?

Also, there is one code divider (//------) left after this method.

MGatner: Removed the divider and the use of the interim variable.

michalsn (approved):

MGatner:
I'm fine not finishing the tests again if you are.

system/HTTP/Message.php
samsonasik: it seems we can use is rector rule for that, that can be to update entire codebase

https://github.com/rectorphp/rector/blob/master/rules/coding-style/src/Rector/Variable/UnderscoreToCamelCaseVariableNameRector.php

PR #3669, opened 2020-09-22
Update view cell doc [ci skip]

### LINKED ISSUE #3668: Documentation Error
Under View Cells guide, under the section "Cell Parameters"
In the last example, the parameters passed to recentPosts() method wasn't accessed correctly.

paulbalandan:
**Description**
Fixes #3668 .

~Along the way, I found a teeny error in `intro/index` so I fixed that too.~

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/outgoing/view_cells.rst
+++ b/user_guide_src/source/outgoing/view_cells.rst
@@ -16,7 +16,7 @@ must return the generated HTML as a string. The method can be either a static me
 Cell Parameters
 ---------------
 
-You can further refine the call by passing a list of parameters in the second parameter to the method. The values passed 
+You can further refine the call by passing a list of parameters in the second parameter to the method. The values passed
 can be an array of key/value pairs, or a comma-separated string of key/value pairs::
 
     // Passing Parameter Array
@@ -25,7 +25,7 @@ can be an array of key/value pairs, or a comma-separated string of key/value pai
     // Passing Parameter String
     <?= view_cell('\App\Libraries\Blog::recentPosts', 'category=codeigniter, limit=5') ?>
 
-    public function recentPosts(array $params=[])
+    public function recentPosts(array $params = [])
     {
         $posts = $this->blogModel->where('category', $params['category'])
                                  ->orderBy('published_on', 'desc')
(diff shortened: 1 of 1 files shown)


user_guide_src/source/intro/index.rst
```diff
@@ -17,9 +17,9 @@ or completely replaced to make the system work the way you need it to.
 In short, CodeIgniter is the malleable framework that tries to provide
 the tools you need while staying out of the way.
 
-***********************
-Who is CodeIgniter For?
-***********************
+************************
+For Whom is CodeIgniter?
```
michalsn, line 21: English is not my first language but that sounds really formal and a bit unnatural to me. Can we go back to the initial form?

michalsn (changes requested):

user_guide_src/source/intro/index.rst
paulbalandan: Considering the word order, the correction should have been: "Whom is CodeIgniter For?" But this construction is too informal.

The original construction of "Who ... for" is grammatically incorrect as "who" would be referring to CodeIgniter and "for" is misplaced. The intended usage of the title is asking for the intended users so it should have been "whom".

michalsn: I will have to leave this decision to @MGatner since I'm not qualified enough 😅

MGatner: The emendation is grammatically correct... and very stiff. But using "who" where one should use "whom" is a bit egregious (in my book). What if we ditch this and go with something else altogether, like "Is CodeIgniter For You?" or "CodeIgniter's Target Audience" or "Who Uses CodeIgniter?" - something both keeping with the User Guide's colloquial tone without butchering the already poor state of the English language.

paulbalandan:
Extracted the intro/index changes to a separate PR, #3681 . Never expected this to grow that much. 😂

Also, sorry for two topics on one PR. I've learned my lesson there. :)

MGatner (approved):

PR #3672, opened 2020-09-23
WIP URI slashes

MGatner:
**Description**
Fixes some issues related to inconsistencies with slashes in URLs.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -41,7 +41,7 @@
 	</logging>
 
 	<php>
-		<server name="app.baseURL" value="http://example.com"/>
+		<server name="app.baseURL" value="http://example.com/"/>
 
 		<!-- Directory containing phpunit.xml -->
 		<const name="HOMEPATH" value="./"/>
--- a/tests/system/HTTP/URITest.php
+++ b/tests/system/HTTP/URITest.php
@@ -343,6 +343,14 @@ class URITest extends \CodeIgniter\Test\CIUnitTestCase
 		$this->assertEquals($expected, (string) $uri);
 	}
 
+	public function testSetPathEquivalisesConstructor()
+	{
+		$uri1 = new URI('http://example.com/some/path');
+		$uri2 = (new URI())->setPath('some/path');
+
+		$this->assertEquals($uri1->getPath(), $uri2->getPath());
+	}
+
 	//--------------------------------------------------------------------
 
 	public function invalidPaths()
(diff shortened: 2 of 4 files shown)


michalsn (commented):
Since you changed the `baseURL` (but it's good) some tests have to be adjusted.

I think the best solution would be to use `ltrim()` for the `path` part if its value is different than `/` - since absolute URIs should use a `/` for an empty path.

MGatner:
This code (in `URI::filterPath()`) has me scratching my head:
```
		// Fix up some leading slash edge cases...
		if (strpos($orig, './') === 0)
		{
			$path = '/' . $path;
		}
		if (strpos($orig, '../') === 0)
		{
			$path = '/' . $path;
		}
```

Paths like "./foo/bar" become "/./foo/bar"? But the rest of the code makes me think the intent was that `getPath()` would always return a preceding slash.

MGatner:
Consider this test too:
```
public function testConstructorSetsAllParts()
	{
		$uri = new URI('http://username:password@hostname:9090/path?arg=value#anchor');
...
		$this->assertEquals('/path', $uri->getPath());
```

It seems to me that the intent was for this method to return a preceding slash. I wouldn't have made that choice myself (maybe it is to ensure redirects, etc., aren't relative?) but I think our solution should aim for that.

MGatner:
Actually, the tests are all over the place. It is hard to tell what the actual intent is. Either way (enforcing a preceding slash or removing all preceding slashes) will result in test failures as it is now.

@lonnieezell Any chance you could weigh in?

MGatner:
Alright, well, I feel like a heel. Here's the comment on `getPath()` which pretty much debunks everything I've said so far:

> The path can either be empty or absolute (starting with a slash) or
> rootless (not starting with a slash). Implementations MUST support all
> three syntaxes.
>
> Normally, the empty path "" and absolute path "/" are considered equal as
> defined in RFC 7230 Section 2.7.3. But this method MUST NOT automatically
> do this normalization because in contexts with a trimmed base path, e.g.
> the front controller, this difference becomes significant. It's the task
> of the user to handle both "" and "/".
>
> The value returned MUST be percent-encoded, but MUST NOT double-encode
> any characters. To determine what characters to encode, please refer to
> RFC 3986, Sections 2 and 3.3.
>
> As an example, if the value should include a slash ("/") not intended as
> delimiter between path segments, that value MUST be passed in encoded
> form (e.g., "%2F") to the instance.
>
> @see    https://tools.ietf.org/html/rfc3986#section-2
> @see    https://tools.ietf.org/html/rfc3986#section-3.3

I will have to take a fresh look at this tomorrow.

MGatner:
Given my previous comment, I don't think `setPath()` is the problem. I do still think we need to encourage a trailing slash on `baseURL` and our current handling does not do a good job handling scenarios where those differ. I will open a new PR to start on this.

PR #3673, opened 2020-09-23
Normalize url_is test path

MGatner:
**Description**
Addresses an issue where the tested path for `url_is()` could have inconsistent results based on preceding slashes.

The underlying issue should be addressed in #3672 but this will ensure `url_is()` works in the meantime, and should function as a general precaution against `URI` return values.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -675,7 +675,7 @@ if (! function_exists('url_is'))
 	{
 		// Setup our regex to allow wildcards
 		$path        = '/' . trim(str_replace('*', '(\S)*', $path), '/ ');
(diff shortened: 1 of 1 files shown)


michalsn (approved):
Seems good to me.

MGatner:
Despite having changed my mind about #3672 I still think this PR is the correct normalization. `url_is()` represents a sort of "convenience shorthand" that should not care whether `getPath()` is returning a relative path (to `baseURL`) or an absolute. For example, when called with no other context one would expect "admin*" should match "https://example.com/admin/users" without needing to specify the preceding `/` to note its absoluteness.

PR #3678, opened 2020-09-24
Update controllers.rst [ci skip]

InsiteFX:
Updated the controllers to use the BaseController and added references to the original controller.

**Checklist:**
- [ x] Securely signed commits
- [x ] User guide updated

### DIFF
--- a/user_guide_src/source/incoming/controllers.rst
+++ b/user_guide_src/source/incoming/controllers.rst
@@ -60,10 +60,6 @@ If you did it right you should see::
 
 This is valid::
 
-    <?php namespace App\Controllers;
-
-    use CodeIgniter\BaseController;
-
     class Helloworld extends BaseController
     {
 
@@ -71,20 +67,12 @@ This is valid::
 
 This is **not** valid::
 
-    <?php namespace App\Controllers;
-
-    use CodeIgniter\BaseController;
-
     class helloworld extends BaseController
     {
 
     }
 
 This is **not** valid::
-
-    <?php namespace App\Controllers;
-
-    use CodeIgniter\BaseController;
 
     class HelloWorld extends BaseController
     {
--- a/user_guide_src/source/incoming/controllers.rst
+++ b/user_guide_src/source/incoming/controllers.rst
@@ -36,6 +36,8 @@ For security reasons be sure to declare any new methods as protected or private.
 
     <?php
 
+    namespace App\Controllers;
+
     class Helloworld extends BaseController
     {
         public function index()
@@ -60,6 +62,11 @@ If you did it right you should see::
 
 This is valid::
 
+    <?php
+
+    namespace App\Controllers;
+
+
     class Helloworld extends BaseController
     {
 
@@ -67,12 +74,20 @@ This is valid::
 
 This is **not** valid::
 
+    <?php
+
+    namespace App\Controllers;
+
     class helloworld extends BaseController
     {
 
     }
 
 This is **not** valid::
+
+    <?php
+
+    namespace App\Controllers;
 
     class HelloWorld extends BaseController
     {
@@ -109,9 +124,9 @@ controller gets called.**
 
 Let's try it. Add a new method to your controller::
 
-    <?php namespace App\Controllers;
+    <?php
 
-    use CodeIgniter\BaseController;
+    namespace App\Controllers;
 
     class Helloworld extends BaseController
     {
@@ -144,9 +159,9 @@ For example, let's say you have a URI like this::
 
 Your method will be passed URI segments 3 and 4 ("sandals" and "123")::
 
-    <?php namespace App\Controllers;
+    <?php
 
-    use CodeIgniter\BaseController;
+    namespace App\Controllers;
 
     class Products extends BaseController
     {
--- a/user_guide_src/source/incoming/controllers.rst
+++ b/user_guide_src/source/incoming/controllers.rst
@@ -27,11 +27,17 @@ Let's try it: Hello World!
 
 Let's create a simple controller so you can see it in action. Using your text editor, create a file called Helloworld.php,
 and put the following code in it. You will notice that the Helloworld Controller is extending the BaseController you can
-also extend the Controller if you do not need the functionality of the BaseController.::
+also extend the Controller if you do not need the functionality of the BaseController.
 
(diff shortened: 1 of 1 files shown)


user_guide_src/source/incoming/controllers.rst
```diff
@@ -26,13 +26,14 @@ Let's try it: Hello World!
 ==========================
 
 Let's create a simple controller so you can see it in action. Using your text editor, create a file called Helloworld.php,
-and put the following code in it::
+and put the following code in it. You will notice that the Helloworld Controller is extending the BaseController you can
+also extend the Controller if you do not need the functionality of the BaseController.::
```
michalsn, line 30: I think it would be nice to explain here the functionality of the `BaseController` - since we mention it and advise users to use it. Maybe just copy-paste the description from it?

user_guide_src/source/incoming/controllers.rst
```diff
@@ -26,13 +26,14 @@ Let's try it: Hello World!
...
 
     <?php namespace App\Controllers;
 
-    use CodeIgniter\Controller;
+    use CodeIgniter\BaseController;
```
michalsn, line 34: The `BaseController` is in the `App\Controllers` namespace so there is no need for `use ...` here.

Maybe it would make sense to leave one example that uses `Controller` so users could understand the difference between using `Controller` and `BaseController` when it comes to the "loading" part? I don't want the newbies to feel lost.

michalsn (changes requested):
In general, I think this change is a good idea. But... I believe the difference between `Controller` and `BaseController` should be described more.

user_guide_src/source/incoming/controllers.rst
InsiteFX: Ok, I'll layout better for everyone to understand I changed it because the frameworks Home controller is using.

user_guide_src/source/incoming/controllers.rst
```diff
@@ -58,9 +65,9 @@ This is valid::
...
```
michalsn, line 68: Sorry, I probably should mention that all occurrences of this are not needed.

Also, there is no `BaseController` in the `CodeIgniter` namespace :)

user_guide_src/source/incoming/controllers.rst
```diff
@@ -69,9 +76,9 @@ This is **not** valid::
...
```
michalsn, line 79: Should be removed.

user_guide_src/source/incoming/controllers.rst
```diff
@@ -80,9 +87,9 @@ This is **not** valid::
...
```
michalsn, line 90: Should be removed.

user_guide_src/source/incoming/controllers.rst
```diff
@@ -119,9 +126,9 @@ Let's try it. Add a new method to your controller::
...
```
michalsn, line 129: Should be removed.

user_guide_src/source/incoming/controllers.rst
```diff
@@ -154,9 +161,9 @@ Your method will be passed URI segments 3 and 4 ("sandals" and "123")::
...
```
michalsn, line 164: Should be removed.

user_guide_src/source/incoming/controllers.rst
```diff
@@ -338,9 +345,9 @@ these helper files will be automatically loaded into memory so that you can use
 inside the controller::
 
     namespace App\Controllers;
-    use CodeIgniter\Controller;
+    use CodeIgniter\BaseController;
```
michalsn, line 348: Should be removed.

michalsn (changes requested):
Almost there :)

user_guide_src/source/incoming/controllers.rst
michalsn: Yes, I get it.

user_guide_src/source/incoming/controllers.rst
```diff
@@ -26,13 +26,20 @@ Let's try it: Hello World!
 ==========================
 
 Let's create a simple controller so you can see it in action. Using your text editor, create a file called Helloworld.php,
-and put the following code in it::
+and put the following code in it. You will notice that the Helloworld Controller is extending the BaseController you can
```
paulbalandan, line 29:
```suggestion
and put the following code in it. You will notice that the Helloworld Controller is extending the BaseController. You can
```

user_guide_src/source/incoming/controllers.rst
```diff
@@ -26,13 +26,20 @@ Let's try it: Hello World!
...
+also extend the Controller if you do not need the functionality of the BaseController.
 
-    <?php namespace App\Controllers;
+The BaseController provides a convenient place for loading components and performing functions that are needed by all your
+controllers. Extend this class in any new controllers:
```
paulbalandan, line 33:
```suggestion
controllers. Extend this class in any new controller:
```

user_guide_src/source/incoming/controllers.rst
```diff
@@ -26,13 +26,20 @@ Let's try it: Hello World!
 ==========================
 
 Let's create a simple controller so you can see it in action. Using your text editor, create a file called Helloworld.php,
-and put the following code in it::
+and put the following code in it. You will notice that the Helloworld Controller is extending the BaseController you can
+also extend the Controller if you do not need the functionality of the BaseController.
```
paulbalandan, line 30:
```suggestion
also extend the CodeIgniter\\Controller if you do not need the functionality of the BaseController.
```

InsiteFX:
Paul, are you saying to remove the namespace and use cause on all of the example controlers in the section?

michalsn:
@InsiteFX Only the `use CodeIgniter\BaseController;` part should be removed. Please restore the `<?php namespace App\Controllers;`.

michalsn (approved):

user_guide_src/source/incoming/controllers.rst
paulbalandan, line 348: Last one to be removed.

paulbalandan: I'll do this in a PR shortly.

paulbalandan (approved):

PR #3680, opened 2020-09-25
apply rector: set variable name to camel case

samsonasik:
@MGatner here I try apply variable name to camel case checker with Rector.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide


michalsn:
From the first look - I think we would need a rule that leaves the initial underscore, e.g. replacing `$_str` variable to `$str` is not desired.

samsonasik:
we can extends the rule with skip variable with _ in first char if needed.

system/Database/BaseBuilder.php
```diff
@@ -3300,14 +3300,14 @@ protected function isLiteral(string $str): bool
...
 
-		if (empty($_str))
+		if (empty($str))
 		{
-			$_str = ($this->db->escapeChar !== '"') ? ['"', "'"] : ["'"];
+			$str = ($this->db->escapeChar !== '"') ? ['"', "'"] : ["'"];
 		}
 
-		return in_array($str[0], $_str, true);
+		return in_array($str[0], $str, true);
```
michalsn, line 3310: Great, it seems to be a problem here for example.

samsonasik: fixed.

samsonasik:
@michalsn I've added `utils/Rector/UnderscoreToCamelCaseLocalVariableNameRector.php` for skip '_' in first character.

MGatner:
This is looking great! I'm going to apply this to a few projects as well.

michalsn:
This looks really good. Only Travis seem to have some hiccup - all tests hanged at the same time. We have to try to rerun this.

system/Helpers/filesystem_helper.php
```diff
@@ -290,7 +290,7 @@ function get_filenames(string $source_dir, ?bool $include_path = false, bool $hi
 	function get_dir_file_info(string $source_dir, bool $top_level_only = true, bool $recursion = false): array
 	{
 		static $fileData = [];
-		$relative_path   = $source_dir;
+		$relativePath    = $source_dir;
```
michalsn, line 293: I wonder if there is a rule that will allow us to rename parameters for functions as well?

system/Database/BaseBuilder.php
michalsn: Great, thanks!

system/Helpers/filesystem_helper.php
samsonasik: there is https://github.com/rectorphp/rector/blob/master/rules/coding-style/src/Rector/Variable/UnderscoreToCamelCaseVariableNameRector.php for that, but we will need to update documentation: doc block and user guide manually

michalsn: Ok, great. I think we should add this too.

Updating the user guide probably will take some time but it's not that it will change much for users... hopefully, people can figure out the difference quite easy :)

samsonasik: Updating docblock `@param` manually in the code will take time as well, as phpstan will report that. I will check if I can add auto-update `@param` docblock rule when updating parameter value if I have a chance

michalsn: Yes, you're right. I forgot about phpstan.

samsonasik:
close reopen to restart travis build

samsonasik:
it seems the dependency cause a very long running phpunit

samsonasik:
trying with distributed rector-prefixed phar with require-dev nette/utils

samsonasik:
Travis build green 🎉

michalsn:
Yay, you made the tests work!

I think this is a good start and this can be merged.

It would be great if we could check the possibility of handling the parameters for functions as well in the separate PR. @MGatner do you think that this type of change would be a good idea?

Of course there is no hurry.

michalsn (approved):

samsonasik:
@michalsn I've updated to update both parameter and `@param` docblock

system/Helpers/filesystem_helper.php
samsonasik: implemented in latest commit.

samsonasik:
All green 🎉 🎉 🎉

michalsn:
It looks great to me! Let's wait if others have anything to add.

lonnieezell (approved):

.github/workflows/test-rector.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          php-version: '7.4'
+          extensions: intl
+
+      - name: Use latest Composer
+        run: composer self-update
+
+      - name: Validate composer.json
+        run: composer validate --strict
+
```
MGatner, line 40: This should use the Composer cache so it can benefit from all the other running workflows. E.g from **test-phpstan.yml**:
```
      - name: Get composer cache directory
        id: composer-cache
        run: echo "::set-output name=dir::$(composer config cache-files-dir)"

      - name: Create composer cache directory
        run: mkdir -p ${{ steps.composer-cache.outputs.dir }}

      - name: Cache composer dependencies
        uses: actions/cache@v2
        with:
          path: ${{ steps.composer-cache.outputs.dir }}
          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
          restore-keys: ${{ runner.os }}-composer-
```

MGatner (changes requested):

.github/workflows/test-rector.yml
samsonasik: implemented.

samsonasik:
I think it is ready. We can try use `rector/rector` instead of current alternative `rector/rector-prefixed` for more stable dependencies check later after #3689 as if using `rector/rector`, travis build is too slow.

MGatner (approved):

PR #3683, opened 2020-09-25
CI Licences and Docblock for Exceptions

mostafakhudair:
Include CodeIgniter Licences in all Exceptions classes and add Docblock for most Exceptions methods.


samsonasik:
remove interface is a bc break

samsonasik:
Beside of bc break I mentioned before, please only create one specific purpose PR to ease review and rollback if something bad happen in the future. 

I am closing this.

PR #3687, opened 2020-09-26
Make travis build faster

samsonasik:
`laminas/laminas-zendframework-bridge` is only needed when there is package prefixed with `Zend`, `Expressive`, or `Apigility`. Since package like `laminas-escaper` already uses `Laminas` prefix,  it can be removed via removing it in travis process, so hopefully phpunit run will be faster.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,8 @@ before_install:
 before_script:
   - echo 'extension = memcached.so' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
   - echo 'extension = redis.so' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
-  - composer install --prefer-source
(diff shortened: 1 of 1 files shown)


samsonasik:
using composer remove command got this : 

```
laminas/laminas-zendframework-bridge is not required in your composer.json and has not been removed
```

using replace seems can cause an issue when developer want to use the repo as is and want to install dependency.

samsonasik:
I think the safest way to re-fill `vendor/laminas/laminas-zendframework-bridge/src/autoload.php` to empty string in travis before_script:

```yaml
   - php -r 'file_put_contents("vendor/laminas/laminas-zendframework-bridge/src/autoload.php", "");'
```

samsonasik:
It seems not really make faster. I am closing it.

PR #3689, opened 2020-09-27
Test framework with github actions

michalsn:
**Description**
This PR is an attempt to replace TravisCI with Github Actions.

**Checklist:**
- [x] Securely signed commits


samsonasik:
probably the following tool can be used https://github.com/migrify/ci-to-github-actions ?

michalsn:
Well... the drop in coverage is interesting. I have no idea what went wrong 😅

.github/workflows/test-phpunit.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Test with PHPUnit
+        run: script -e -c "vendor/bin/phpunit -v"
+        env:
+          DB: ${{ matrix.db-platforms }}
+          TERM: xterm-256color
+
+      - name: Run Coveralls
+        if: ${{ success() }}
+        run: |
+          composer global require twinh/php-coveralls
```
samsonasik, line 92: should can use `php-coveralls/php-coveralls` instead?

michalsn: I wanted to, but as I understand it's not yet fully supported https://github.com/php-coveralls/php-coveralls/issues/273

samsonasik:
I think coverage decreased is fine as far as the the test still works.

.github/workflows/test-phpunit.yml
```diff
@@ -0,0 +1,95 @@
+name: Test Framework
+
+on:
+  push:
+    branches: 
+      - master
+  pull_request:
+    branches:
+      - develop
```
samsonasik, line 9: 4.1 need to be updated as wel

paulbalandan:
My guess for the decreased coverage is that coveralls only considered the lines covered by the first run using SQLite3. Runs for MySQLi and Postgres were not considered anymore. Is it even possible to consolidate the runs?

.github/workflows/test-phpunit.yml
michalsn: Done, thanks.

michalsn:
Thanks, I'm testing this right now - will see :)

paulbalandan:
I think we need to set up `COVERALLS_PARALLEL: true` for multiple runs. Then setup another job for `coveralls_finish`.

https://github.com/twinh/php-coveralls#github-actions

app/Config/Database.php
```diff
@@ -89,13 +89,13 @@ public function __construct()
 		{
 			$this->defaultGroup = 'tests';
 
-			// Under Travis-CI, we can set an ENV var named 'DB_GROUP'
+			// Under Github Actions, we can set an ENV var named 'DB'
```
MGatner, line 92: I would like to see this whole thing find a better home than **app/Config/Database.php**. I'm not sure where that is right now, but I will think on it.

michalsn:
@paulbalandan Oh... it was there and I missed it - thanks :)

.github/workflows/test-phpunit.yml
michalsn: I see there is now a pending PR for this issue, so of course, if they merge it we will switch to `php-coveralls`.

app/Config/Database.php
MGatner: I think the way to do this will be with a Database Registrar in `Tests\Support` specific to this repo. Problem is that registrars are currently disabled during testing (I have no idea why - @lonnieezell?). I have opened https://github.com/codeigniter4/CodeIgniter4/pull/3693 to see about changing that.

lonnieezell: I don't have any recollection on why those were turned off for testing, and I can't think of a horrible reason to run them - so sounds go for it! :)

MGatner:
The new DB Config content looks good! Probably need to rebase.

michalsn:
Great, I broke something and I didn't even notice it. Will try to fix it.

samsonasik:
The badge build in readme.md need to be updated to github action badge

MGatner:
Could we name the workflow "PHPUnit" to be consistent with our other test workflows?

tests/_support/Config/Registrar.php
```diff
@@ -0,0 +1,34 @@
...
+	{
+		$config = [];
+
+		// Under Github Actions, we can set an ENV var named 'DB'
+		// so that we can test against multiple databases.
+		if ($group = getenv('DB'))
+		{
+			if (is_file(TESTPATH . '_github/Database.php'))
+			{
+				require TESTPATH . '_github/Database.php';
```
MGatner, line 22: Sorry, just thinking of this... but let's get rid of the extra file. Now that we have an established Registrar we can work with the settings directly here. It is a better example of how to use a Registrar properly, so could be helpful to developers looking through the code.

.github/workflows/test-phpunit.yml
```diff
@@ -0,0 +1,108 @@
+name: PHPUnit
+
+on:
+  push:
```
MGatner, line 4: I'm not a fan of the `push` triggers, since anything pushed will already have been tested during PRs. I'm fine if we want to leave them as a "just in case" but it will likely just end up being debris in the Actions tab.

MGatner (changes requested):
Looking very good! A few things I should have mentioned earlier. But I also want @lonnieezell to approve the removal of TravisCI before we merge, since that is a pretty big organizational decision. We could always run tests on both in the interim, but that has disadvantages.

lonnieezell:
I'm all for doing this. Travis has become really slow, whereas Actions seems to be much faster, and having everything in one place is great. 

Is there a way to do an "allowed failure" and test against 8 yet with Actions? I know it will fail currently, but it's something we need to get in place sometime soon.

MGatner:
@lonnieezell There's a lot of noise about "allowed failures" but basically it looks like there is not currently an equivalent feature to the way Travis lets us test against nightlies. Check out this issue if you want to contribute to said noise: https://github.com/actions/toolkit/issues/399

MGatner:
For now, since the plan is to release `4.1` alongside the PHP changes maybe we should add PHP 8 to the matrix for `4.*` branches?

lonnieezell:
@MGatner That's too bad. The continue on failure action doesn't seem to help here, either, I don't think. Oh well.

But yes, I think adding 8 to 4.1 would be good.

MGatner:
Oooo the `experimental` flag is a nice find!

michalsn:
Well... idk if this will work out as we want it to but fingers crossed.

michalsn:
Ok... this is not working. I guess we would have to make all checks successful 😕 Probably it's possible but have no much sense.

michalsn:
Ok, this is a bit redundant but I have no other ideas on how to handle this. Right now PHP 8.0 has its own job which is only run for 4.1 branch.

Too bad that `matrix` is not available for the `if` condition on the "main" level. It would simplify our file a lot.

If someone has any other ideas on how to handle this, I'm all ears. Hopefully, github will introduce something similar to `allow-failure` soon.

MGatner:
Thanks for your work on this! Your experience is very valuable for the big upcoming November stuff. 

I think for now I'd rather see a different YML file pushed to the `4.1` branch and `develop` not deal with 8 at all. IMO 8 is not worth testing right now in any definitive way, and maybe GitHub will have some better solution available by November. 🤞

michalsn:
Yeah... I'm totally fine with excluding PHP8. We can address this later in 4.1 branch.

MGatner (approved):
Agreed. Looks great! Travis is taking 2 hours for a run - excited to see if this speeds is up!

.github/workflows/test-phpunit.yml
keradus: @michalsn , the change was incorporated into upstream of php-coveralls , with big thanks for @twinh

PR #3692, opened 2020-09-28
Fix transparency handling for webp

### LINKED ISSUE #3690: Image format webp is not preserve transparent
Image format webp is not preserve transparent during resize. 

I found the problem in the code `/system/Images/Handlers/GDHandler.php`

` 
Line 249

...
if ($this->image()->imageType === IMAGETYPE_PNG) // png we can actually preserve transparency
{
	imagealphablending($dest, false);
	imagesavealpha($dest, true);
}
...
`

fixing the code on this one solves the problem

`
...
if ($this->image()->imageType === IMAGETYPE_PNG || $this->image()->imageType === IMAGETYPE_WEBP)
...
`

Please keep this issue in mind in future releases. Thanks!

michalsn:
**Description**
Fixes #3690

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Images/Handlers/BaseHandler.php
+++ b/system/Images/Handlers/BaseHandler.php
@@ -133,6 +133,16 @@ abstract class BaseHandler implements ImageHandlerInterface
 	];
 
 	/**
+	 * Image types with support for transparency.
+	 *
+	 * @var array
+	 */
+	protected $supportTransparency = [
+		IMAGETYPE_PNG,
+		IMAGETYPE_WEBP,
+	];
+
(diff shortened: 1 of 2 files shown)


MGatner:
Maybe we should make the list of `IMAGE_TYPE` that supports transparency a class variable or something, so we don't have to keep updating this function (e.g. JPEG2000 supports it, might need to add down the road).

MGatner (approved):
Thanks! Looks good.

PR #3698, opened 2020-09-30
Builder Get Connection

MGatner:
**Description**
It is a source of frequent confusion that the Query Builder cannot return the last insert ID, errors, or other data related to the database connection. Because $db is protected there is no way to access the database instance once you have instantiated a Builder. This adds a method to return the `ConnectionInterface` in use on a builder.

Supersedes #3697 because I realized that `ConnectionInterface` does not specific a `insertID()` method. Not sure if that is an oversight but this is a little more flexible anyways.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -271,9 +271,9 @@ class BaseBuilder
 	/**
 	 * Constructor
 	 *
-	 * @param  string|array                              $tableName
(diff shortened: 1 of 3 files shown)


michalsn (approved):
It would be nice to mention about this in the user guide since this may be handy for some people - this way it won't become a "secret" feature :)

PR #3684, opened 2020-09-26
New Array Helper Function

savasdersimcelik:
Returns the value of an element with a key value in an array of uncertain depth.

Each pull request should address a single issue and have a meaningful title.

**Description**
I wrote a new method for Array Helper. With this method, we can search an array of uncertain depth.

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

PR #3701, opened 2020-09-30
New Array Helper Function

MGatner:
**Description**
Supersedes #3684 with updated provider tests.

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -120,3 +120,34 @@ if (! function_exists('_array_search_dot'))
 		return $array[$currentIndex];
 	}
 }
+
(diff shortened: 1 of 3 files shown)


MGatner:
Added User Guide reference

MGatner:
@savasdersimcelik while adding the User Guide info I noticed that `dot_array_search()` already accepts the parameters in the order "search, array" so I updated your function to be consistent. I also removed the restriction that `$key` be a string (since it could really be any valid index type) and added another test for that.

Please review and make sure I kept the intent of your PR.

michalsn (approved):

PR #3702, opened 2020-09-30
Factory class for Entities

paulbalandan:
**Description**
This PR exposes two new API in creating new or shared instances of `Entity` classes.

Procedural: Using the `entity()` common function in creating shared/single instances of Entities.
OOP: Using `CodeIgniter\EntityFactory` in creating shared/single instances of Entities.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -39,6 +39,8 @@
 
 use CodeIgniter\Config\Config;
 use CodeIgniter\Database\ConnectionInterface;
+use CodeIgniter\Entity;
+use CodeIgniter\EntityFactory;
 use CodeIgniter\Files\Exceptions\FileNotFoundException;
 use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\HTTP\RequestInterface;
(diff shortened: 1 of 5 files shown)


MGatner:
I'm not opposed, but I don't understand. Unlike Models where you reuse the same model over and over to generate different results... each instance of an Entity should have a unique set of `$attributes` and thus explicitly *not* be reused. Could you describe the theoretical use case for this?

paulbalandan:
I understand that you can set the return type of a model query to a specific entity class and use that later to save changes into the database. My use case is that this entity instance needs to pass through hoops and loops with different external libraries for processing before being actually saved into the db. One possible solution I have thought is to extend the library and pass the instance to the constructor for proper dependency injection. But this becomes burdensome if the entity instance needs to pass *several* libraries. My other solution is this: getting the same instance from wherever I accessed it. Then after I have saved this, I can just unset this instance for another session with a different row.

michalsn:
This can be closed in favor of #3722

PR #3705, opened 2020-10-01
Autoloader include

MGatner:
**Description**
This issue arose with the Autoloader. I don't understand all the nuances of the difference but I thought I'd at least run it through the tests and see if anyone else here with more Autoloader experience had ideas.

See: https://github.com/phpstan/phpstan/issues/3907

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -377,7 +377,7 @@ class Autoloader
 
 		if (is_file($file))
 		{
-			require_once $file;
+			include_once $file;
 
 			return $file;
 		}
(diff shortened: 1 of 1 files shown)


mostafakhudair:
Good solution but i think you forgot renaming requireFile method to includeFile also editing DocBlock.

MGatner:
Both good suggestions! But I need to make sure this is even theoretically viable first.

michalsn (approved):
I love when problems can be solved just by simple changes like this 😄 

From what I remember `include_once` only throws a warning but looking at the tests it isn't causing any issues, so 👍

PR #3706, opened 2020-10-01
adding solution to CORS error

rattones:
Adding solution to CORS error, applying direct CORS header to the response object by setting CORS config parameters

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -431,4 +431,17 @@ class App extends BaseConfig
 	 * @var boolean
 	 */
 	public $CSPEnabled = false;
+
+	/*
+	|--------------------------------------------------------------------------
+	| CORS - Cross Origin Resource Sharing
+	|--------------------------------------------------------------------------
+	| Enables the CORS solution for API to accept requests for outside domains
+	| on PHP backend application
+	|
+	| For a better understanding of CORS, see these documents:
+	|   - https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Controle_Acesso_CORS
+	|   - https://www.php.net/manual/en/function.header.php
+	*/
+	public $CORSEnabled = true;
 }
--- a/app/Config/CORS.php
+++ b/app/Config/CORS.php
@@ -1,0 +1,36 @@
+<?php namespace Config;
+
+use CodeIgniter\Config\BaseConfig;
+
+/**
+ * Class CORS
+ *
+ * Stores the default settings for the CORS, if you choose to use it. The
+ * values here will be read in and set as defaults for the site. If needed,
+ * they can be overridden on a page-by-page basis.
+ *
+ * Suggested reference for explanations:
+ *   - https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Controle_Acesso_CORS
+ *   - https://www.php.net/manual/en/function.header.php
+ *
+ * @author Marcelo Ratton <www.MarceloRatton.com>
+ * @package Config
+ */
+class CORS extends BaseConfig
+{
+  /**
+   * Header to allow CORS for any domain request
+   * Access-Control-Allow-Origin  = * for all domains, if you need to set one
+   *    change this value
+   * Access-Control-Allow-Methods = POST, GET, OPTIONS, PUT, DELETE - default
+   *    methods, for all methods in HTTP: POST, GET, OPTIONS, PUT, DELETE,
+   *    PATCH, MATCH, CONNECT, TRACE
+   */
+  public $headerList = [
+    'Access-Control-Allow-Origin'      => '*',
+    'Access-Control-Allow-Methods'     => 'POST, GET, OPTIONS, PUT, DELETE',
+    'Access-Control-Allow-Credentials' => 'true',
+    'Access-Control-Max-Age'           => '86400',
+    'Access-Control-Allow-Headers'     => 'Content-Type, Authorization, X-Requested-With'
+];
+}
--- a/app/Config/CORS.php
+++ b/app/Config/CORS.php
@@ -10,7 +10,7 @@ use CodeIgniter\Config\BaseConfig;
  * they can be overridden on a page-by-page basis.
  *
  * Suggested reference for explanations:
- *   - https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Controle_Acesso_CORS
+ *   - https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
  *   - https://www.php.net/manual/en/function.header.php
  *
  * @author Marcelo Ratton <www.MarceloRatton.com>
--- a/system/HTTP/CORS.php
+++ b/system/HTTP/CORS.php
@@ -50,7 +50,7 @@ class CORS
 					throw new HTTPException("Method not allowed");
 				}
 			}
-			array_push($this->allowedMethods, [$name=>$value]);
+			$this->headerList[$name]= $value;
 		}
 
 	}
--- a/system/HTTP/CORS.php
+++ b/system/HTTP/CORS.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace CodeIgniter\HTTP;
+
+use CodeIgniter\HTTP\Exceptions\HTTPException;
+
+/**
+ * Class CORS - Cross-Origin Resource sharing
+ *
+ * Provides tools for working with the CORS header to help defeat XSS attacks.
+ *
+ * @see     https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+ * @see     https://www.html5rocks.com/en/tutorials/file/xhr2/#toc-cors
+ * @package CodeIgniter\HTTP
+ */
+
+class CORS
+{
+
+	/**
+	 * Allowed methods
+	 * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+	 *
+	 * @var array
+	 */
+	protected $allowedMethods= [
+		'GET',
+		'HEAD',
+		'POST',
+		'PUT',
+		'DELETE',
+		'CONNECT',
+		'OPTIONS',
+		'TRACE',
+		'PATCH'
+	];
+
+	/**
+	 * Headers to be mounted
+	 *
+	 * @var array
+	 */
+	public $headerList= [];
+
+	public function __construct(\Config\CORS $config)
+	{
+		foreach ($config->headerList as $name=>$value) {
+			if ($name === 'Access-Control-Allow-Methods') {
+				if (!$this->methods($value)) {
+					throw new HTTPException("Method not allowed");
+				}
+			}
+			array_push($this->allowedMethods, [$name=>$value]);
+		}
+
+	}
+
+	/**
+	 * Verify if the method is allowed
+	 *
+	 * @param string $methods
+	 * @return bool
+	 */
+	private function methods(string $methods) : bool
+	{
+		// gerating array of methods in user config
+		$arrayMethods= explode(',', $methods);
+
+		// for each user method test if it is allowed
+		foreach ($arrayMethods as $method) {
+			// removing white spaces
+			$method= trim($method);
+			if (!in_array($method, $this->allowedMethods)) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+}
--- a/system/HTTP/Response.php
+++ b/system/HTTP/Response.php
@@ -271,8 +271,8 @@ class Response extends Message implements ResponseInterface
 		$this->cookieSameSite = $config->cookieSameSite ?? $this->cookieSameSite;
 
     // CORS
-    $this->CORSEnabled     = $config->CORSEnabled;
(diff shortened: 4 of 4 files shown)


rattones:
adding a solution to CORS error

rattones:
Adding solution to CORS error, applying direct CORS header to the response object by setting CORS config parameters

app/Config/App.php
```diff
@@ -431,4 +431,17 @@ class App extends BaseConfig
...
+	|--------------------------------------------------------------------------
+	| CORS - Cross Origin Resource Sharing
+	|--------------------------------------------------------------------------
+	| Enables the CORS solution for API to accept requests for outside domains
+	| on PHP backend application
+	|
+	| For a better understanding of CORS, see these documents:
+	|   - https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Controle_Acesso_CORS
+	|   - https://www.php.net/manual/en/function.header.php
+	*/
```
paulbalandan, line 445:
```suggestion
	/*
	 * --------------------------------------------------------------------------
	 *  CORS - Cross Origin Resource Sharing
	 * --------------------------------------------------------------------------
	 * Enables the CORS solution for API to accept requests for outside domains
	 * on PHP backend application
	 *
	 * For a better understanding of CORS, see these documents:
	 * - https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Controle_Acesso_CORS
	 * - https://www.php.net/manual/en/function.header.php
	 *
	 * @var boolean
	 */
```

app/Config/CORS.php
```diff
@@ -0,0 +1,36 @@
...
+ * Stores the default settings for the CORS, if you choose to use it. The
+ * values here will be read in and set as defaults for the site. If needed,
+ * they can be overridden on a page-by-page basis.
+ *
+ * Suggested reference for explanations:
+ *   - https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+ *   - https://www.php.net/manual/en/function.header.php
+ *
+ * @author Marcelo Ratton <www.MarceloRatton.com>
+ * @package Config
```
paulbalandan, line 17:
```suggestion
 *
```
While the intention is noble, the coding style guide does not accept @author tags. Also, @package tags are unnecessary since the file is namespaced.

system/HTTP/CORS.php
```diff
@@ -0,0 +1,81 @@
...
+use CodeIgniter\HTTP\Exceptions\HTTPException;
+
+/**
+ * Class CORS - Cross-Origin Resource sharing
+ *
+ * Provides tools for working with the CORS header to help defeat XSS attacks.
+ *
+ * @see     https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+ * @see     https://www.html5rocks.com/en/tutorials/file/xhr2/#toc-cors
+ * @package CodeIgniter\HTTP
```
paulbalandan, line 14:
```suggestion
```

system/HTTP/CORS.php
```diff
@@ -0,0 +1,81 @@
...
+	private function methods(string $methods) : bool
+	{
+		// gerating array of methods in user config
+		$arrayMethods= explode(',', $methods);
+
+		// for each user method test if it is allowed
+		foreach ($arrayMethods as $method) {
+			// removing white spaces
+			$method= trim($method);
+			if (!in_array($method, $this->allowedMethods)) {
```
paulbalandan, line 73:
```suggestion
			if (!in_array($method, $this->allowedMethods, true)) {
```

system/HTTP/Response.php
```diff
@@ -720,6 +737,14 @@ public function sendHeaders()
 			header($name . ': ' . $this->getHeaderLine($name), true, $this->statusCode);
 		}
 
+    // Verifying if CORS is not already sent and if is enabled
+    if (!in_array('Access-Control-Allow-Origin', headers_list()) && $this->CORSEnabled) {
```
paulbalandan, line 741:
```suggestion
    if (! in_array('Access-Control-Allow-Origin', headers_list(), true) && $this->CORSEnabled) {
```

paulbalandan (changes requested):
Please add tests to show that this is working. Also, failing tests should be addressed, too.

Aside from few review notes, please run the code style checker against your code.

app/Config/CORS.php
rattones: other system files have these tags, so keep it

rattones:
> Please add tests to show that this is working. Also, failing tests should be addressed, too.
> 
> Aside from few review notes, please run the code style checker against your code.

I need to study how to correct these tests cause I any change for the database in my suggestion for the CORS problem ..

system/HTTP/Response.php
```diff
@@ -257,6 +270,10 @@ public function __construct($config)
 		$this->cookieHTTPOnly = $config->cookieHTTPOnly;
 		$this->cookieSameSite = $config->cookieSameSite ?? $this->cookieSameSite;
 
+    // CORS
+    $this->CORSEnabled     = $config->CORSEnabled;
```
samsonasik, line 274: this is bc break without `??` check, look at previous config set:

```php
$this->CORSEnabled = $config->CORSEnabled ?? $this->CORSEnabled;
```

app/Config/CORS.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+class CORS extends BaseConfig
```
samsonasik, line 19: Add new config class in `app` directory is bc break without handling at `system`.

system/HTTP/Response.php
rattones: this is not form my code ...

app/Config/CORS.php
rattones: I don't think so, cause at the default configuration is **false** so it will not be used by default

system/HTTP/Response.php
rattones: It's not necessary, all other config params are mandatories in **App/Config/App.php** file

samsonasik: when user run `composer update` with existing their appstarter, that new config will make application break. That's why need check if that exists.

app/Config/CORS.php
samsonasik: It instantiated in `system/HTTP/Response.php`, that's why it make break, see my comment at https://github.com/codeigniter4/CodeIgniter4/pull/3706#discussion_r508553713

samsonasik:
Please read https://github.com/codeigniter4/CodeIgniter4/blob/develop/CONTRIBUTING.md#breaking-changes

system/HTTP/Response.php
rattones: ok, I understand and I'll fix it in my PR but just looking in **Systema/HTTP/Response.php** file we have this:
>		$this->CSPEnabled     = $config->CSPEnabled;
>		$this->cookiePrefix   = $config->cookiePrefix;
>		$this->cookieDomain   = $config->cookieDomain;
>		$this->cookiePath     = $config->cookiePath;
>		$this->cookieSecure   = $config->cookieSecure;
>		$this->cookieHTTPOnly = $config->cookieHTTPOnly;
>		$this->cookieSameSite = $config->cookieSameSite ?? $this->cookieSameSite;

just one are using this and it's in "*production*"

app/Config/CORS.php
rattones: ok I agree,
but this will make almost a copy of the ""Config/CORS.php"" file into **System/HTTP** folder ... or I can avoid the file in **Config** folder? just having a empty file/class there?
Cause I'm trying to keep as simple as possible

that is it?

system/HTTP/Response.php
paulbalandan: The thing is the last setting is just new after v4.0.4, so users using v4.0.4 and below do not have the sameSite setting in their App config. Adding the `??` will ensure fallback in case the App config is not updated to have the sameSite setting.

This case also apply to your PR. This is a new addition so existing users will be lacking that.

app/Config/CORS.php
paulbalandan: According to the PHPDoc documentation, @package tags are only required for **non-namespaced** classes. The use of package tags was carried over from CI3. Since CI4 is actively using namespaces, the tag is really not necessary. Those lingering package tags will be addressed down the road, so please do not add more.

system/HTTP/Response.php
```diff
@@ -257,6 +270,10 @@ public function __construct($config)
 		$this->cookieHTTPOnly = $config->cookieHTTPOnly;
 		$this->cookieSameSite = $config->cookieSameSite ?? $this->cookieSameSite;
 
+    // CORS
+    $this->CORSEnabled    = $config->CORSEnabled ?? $config->CORSEnabled;
+    $this->CORS 					= new CORS(new \Config\CORS());
```
samsonasik, line 275: It is bc break, no `\Config\CORS()` yet in current 4.0.4 version.

system/HTTP/Response.php
rattones: it's done ..

agungsugiarto:
I thinks cors need to separate to the filter and implement preflight request.
Take the look my own implementation cors for CodeIgniter4 https://github.com/agungsugiarto/codeigniter4-cors

rattones:
> I think cors need to separate to the filter and implement preflight request.
> Take the look my own implementation cors for CodeIgniter4 https://github.com/agungsugiarto/codeigniter4-cors

I liked that solution but you create a `config/cors.php` like me, and that is the point they are saying that cause bc break

agungsugiarto:
> I liked that solution but you create a `config/cors.php` like me, and that is the point they are saying that cause bc break

My config its separate outside, this config will be fine and not be bc the system.

system/HTTP/Response.php
paulbalandan: OK I will demonstrate with a scenario what will happen if this new config file is not handled properly.

Let us say I have a project using `v4.0.4` as the framework version. In this version there are no `CORS` class in `system/` yet and `Config\CORS` in `app/`. Let us assume further that I am using Composer so the files are in the `vendor/` directory. Now let us say the framework releases the new version `v4.0.5` this November. I got excited and updated my project to use the new version on that date. On my update, in the `vendor/` directory the `CORS` class from this PR is now found in `vendor/codeigniter4/framework/system/HTTP/`. The `Config\CORS` file also from this PR is now found in `vendor/codeigniter4/framework/app/Config/`. OK, good then. **BUT** take note that your `CORS` class **will not** look for the config file in `vendor/codeigniter4/framework/app/Config/`. It will **instead** look for the file in **my project's `app/Config/` directory**. But is that `Config\CORS` file in my project's `app/Config/` folder? Definitely no. So in the instantiation of the `CORS` class it will throw an error for class not found, effectively breaking my project.

So instead of being happy on upgrading to v4.0.5 for the new features I instead got a broken project. A completely failing project which I have no recourse but to revert to v4.0.4 just for my project to be alive.

I hope now you get what @samsonasik is trying to tell that this is a BC break without proper handling.

paulbalandan:
> > I liked that solution but you create a `config/cors.php` like me, and that is the point they are saying that cause bc break
> 
> My config its separate outside, this config will be fine and not be bc the system.

Agreed. That config is from an external library so it will not break the system.

system/HTTP/Response.php
rattones: Let me understand what you said. If someone updates a project with this new file `Config/CORS.PHP`the composer will put this file at `vendor` folder, not in `App/Config` of the project? OK, I agree, but how can this be added new config feature without creating an external library for this? Cause the intention here is adding this as a solution to the CORS problem when creating RESTful APIs and this need a configuration params as part of this solution, not as an external library but a core solution.

paulbalandan: Here you need to come up with a fallback solution to ensure compatibility with earlier versions. So instead of creating a new `Config\CORS` class you can put those settings as a separate section to `Config\App` class. Then in your main `CORS` class, you will have **default values** provided for the equivalent settings in config. So this is basically like the implementation in samesite: `$this->cookieSamesite = $config->cookieSamesite ?? $this->cookieSamesite;`.

No need for you to create an external lib for that. What I meant in my comment below was @agungsugiarto 's library will not cause BC breaks because it is external. His config is consumed by his library's main class which in turn is passed down to the core.

In your case, the config will be integrated to the core. So without proper precautions and handling, the disadvantages may outweigh the benefits so there's a need to have fallbacks. For earlier CI versions, these fallbacks will be their defense against BC before they can actually update the affected config files. Of course, this is not a problem with newer releases.

rattones: I made some tests here and it's really will cause a BC BREAK cause `composer` will put this file in the `vendor` folder.

So, the `namespace` is correctly set, the `Config/CORS.php` it's really necessary to this solution of the core framework CORS error. 

So it can be rejected, as you want, or try to add this as a new version that will need a manual upgrade as is the versions 3 for 4.


For now, I'll up this solution to the CORS problem in my GitHub as a library that will need to manually and try to add this to the `composer`

paulbalandan:
Closing this for now until we arrive at a better solution.

PR #3707, opened 2020-10-02
Update Function Resize - Image Manipulation

adigamara24:
Make optional for developer, if want to use function resize image just set width and make height set auto ..
So make set height is optional

**Description**

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)


michalsn:
I think what you want to do has already been implemented and can be achieved with a call like this:

```php
$handler->resize(100, 0, true, 'width');
```

Can you confirm that such a call has the same effect that you want to achieve?

adigamara24:
Wow, it's working and the same effect .. thanks @michalsn

PR #3712, opened 2020-10-02
Factory instance basenames

MGatner:
**Description**
This addresses a bug I've been working around for a while. Consider this scenario: a `WidgetModel` in the `App` namespace, and an unrelated third-party module with its own `WidgetModel`.

Currently, `ModelFactory` stores shared instances solely by their "base name" (class without the namespace). This creates a problem for our scenario, because even supplying the full namespace will only compare the base names:
```
$appModel = model('App\Models\WidgetModel');
$modModel = model('Module\Org\Models\WidgetModel');
get_class($modModel); // App\Models\WidgetModel
```
This PR changes `ModelFactory` to tracks base names separately from the full class name so a developer can choose between `model($fullClassName)` (which will always return the intended class) or `model($baseClassName)` (which lets the `FileLocator` determine the best match).
***
If these changes seem appropriate to all, the same structure should be applied to `Config`, or maybe even abstracted to a `Components` class that functions similar to `Services` but for distinct components (models, config, entities(?), some thing?).

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,10 +30,10 @@
 		}
 	},
 	"autoload-dev": {
-                "psr-4": {
-                        "Utils\\": "utils"
-                }
-        },
+		"psr-4": {
+			"Utils\\": "utils"
+		}
+	},
 	"scripts": {
 		"post-update-cmd": [
 			"@composer dump-autoload",
--- a/system/Database/ModelFactory.php
+++ b/system/Database/ModelFactory.php
@@ -13,6 +13,13 @@ class ModelFactory
 	static private $instances = [];
 
 	/**
+	 * Mapping of class basenames (no namespace) to instances.
+	 *
+	 * @var string[]
+	 */
+	static private $basenames = [];
+
+	/**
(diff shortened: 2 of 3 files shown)


michalsn (approved):
I agree that this should be implemented for other components as well.

paulbalandan:
Agreed. This should be extracted into an abstract class where all factory classes will just extend from.

MGatner:
I actually had something a little more Services-like in mind. Would one (or both) of you check out my Handlers module? https://github.com/tattersoftware/codeigniter4-handlers

It dynamically loads components based on their path within a namespace. I was thinking of merging some of that functionality into this new Factory implementation.

michalsn:
Sure, I will look into it tomorrow.

MGatner:
Actually, feel free to look at it still but I've gone a different direction. I'm almost done with a first draft, I will post it when I have the class done for input before I work on tests and implementation.

PR #3713, opened 2020-10-02
Controller Helpers

MGatner:
**Description**
`helper()` already prefers an array, so unpacking the names with `foreach` is just a performance cost.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -170,10 +170,7 @@ class Controller
 			return;
 		}
 
-		foreach ($this->helpers as $helper)
-		{
-			helper($helper);
-		}
+		helper($helper);
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


mostafakhudair:
I suggest that we should call it directly inside initController while loadHelpers method will not be used somewhere else

```
public function initController(RequestInterface $request, ResponseInterface $response, LoggerInterface $logger)
{
	$this->request  = $request;
	$this->response = $response;
	$this->logger   = $logger;

	if ($this->forceHTTPS > 0)
	{
		$this->forceHTTPS($this->forceHTTPS);
	}

	// Autoload helper files .
	if (! empty($this->helpers))
	{
		helper($this->helpers);
	}
} 
```

michalsn (approved):

PR #3717, opened 2020-10-03
Fix PHPUnit status badge in readme

samsonasik:
no need `?branch` in uri.

**Checklist:**
- [x] Securely signed commits


michalsn (approved):
I was pretty sure it was there, thanks.

PR #3720, opened 2020-10-03
Load a JSON request into Validator 

### LINKED ISSUE #3719: Bug: Validating Json Requests
**Describe the bug**
When building a json API, the Controller::validate function does not work because the data from the json sent to the request is not being loaded correctly.

**CodeIgniter 4 version**
CodeIgniter 4.0.4

**Affected module(s)**
Validation

1. Create a controller endpoint for accepting the request
2. In the endpoint use this function call `$this->validate(['username' => 'required|alpha_dash')` to validate your request
3. Send `{"username": "test"}` to the endpoint
4. I would expect this to pass the validation but it is saying that username is missing.

**Context**
 - OS: CentOS7
 - Web server: Apache 2.4.6
 - PHP version: 7.2.24

caswell-wc:
fixes #3719 

**Description**
Adjusted Validation::withRequest so that it recognizes when the request is a json request and loads the data appropriately.
This allows someone who is building a JSON API to be able to validate their incoming JSON's using the Controller::validate function.

**Checklist:**
- [x] Securely signed commits
- [ ] (N/A) Component(s) with PHPdocs
- [ ] (N/A) User guide updated n/a
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -364,7 +364,7 @@ class Validation implements ValidationInterface
 	 */
 	public function withRequest(RequestInterface $request): ValidationInterface
 	{
-		if ($request->hasHeader('Content-Type') && $request->getHeader('Content-Type')->getValue() === 'application/json')
+		if ($request->isJson())
 		{
 			$this->data = $request->getJSON(true);
 			return $this;
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -364,6 +364,12 @@ class Validation implements ValidationInterface
 	 */
 	public function withRequest(RequestInterface $request): ValidationInterface
 	{
+		if ($request->hasHeader('Content-Type') && $request->getHeader('Content-Type')->getValue() === 'application/json')
+		{
+			$this->data = $request->getJSON(true);
+			return $this;
+		}
+
 		if (in_array($request->getMethod(), ['put', 'patch', 'delete'], true))
 		{
 			$this->data = $request->getRawInput();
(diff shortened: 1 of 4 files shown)


michalsn (approved):
It seems to be a really useful addition. I think it's good to merge, but let's wait for a moment and see if others have anything to add.

system/Validation/Validation.php
```diff
@@ -364,6 +364,12 @@ protected function processRules(string $field, string $label = null, $value, $ru
 	 */
 	public function withRequest(RequestInterface $request): ValidationInterface
 	{
+		if ($request->hasHeader('Content-Type') && $request->getHeader('Content-Type')->getValue() === 'application/json')
```
MGatner, line 367: I see this code or similar a lot. Maybe it is time for `Request::isJSON()`? Maybe even on `Message`? @caswell-wc you appear to work with JSON a lot in the HTTP/API layer - what do you think?

caswell-wc: I have been thinking about suggesting something like that. I think it's a good idea. I have some other ideas on how we could improve the request and having that method would be helpful.

caswell-wc: I went ahead and created that method and pushed it into this PR.

MGatner: I like it! Since we already have `getJSON()` could you please make the capitalization consistent? `isJSON()`

caswell-wc: That's fixed

paulbalandan (approved):

PR #3721, opened 2020-10-03
Removing unused --dev and autoload on PHPUnit workflow

samsonasik:
Try make PHPUnit workflow faster.

**Checklist:**
- [x] Securely signed commits


samsonasik:
it seems faster now @michalsn

michalsn (approved):
Yey! 🎉 

I think the slowdown may have been due to the switching from `rector-prefixed` to `rector`. Thanks for diving into it!

PR #3722, opened 2020-10-04
Factories

MGatner:
**Description**
Defines a new dynamic lookup class, `Factories`, for returning shared instances of named components.

Supersedes #3712 (but still read about the `ModelFactory` bug there, which also applies to the previous `Config`).

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [Unreleased](https://github.com/codeigniter4/CodeIgniter4/tree/HEAD)
+
+[Full Changelog](https://github.com/codeigniter4/CodeIgniter4/compare/v4.0.4...HEAD)
+
+**Deprecations:**
+
+- `CodeIgniter\Database\ModelFactory` is now deprecated in favor of `CodeIgniter\Config\Factories::models()`
+- `Config\Config` is now deprecated in favor of `CodeIgniter\Config\Factories::config()`
+
 ## [v4.0.4](https://github.com/codeigniter4/CodeIgniter4/tree/v4.0.4) (2020-07-15)
 
 [Full Changelog](https://github.com/codeigniter4/CodeIgniter4/compare/v4.0.3...v4.0.4)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 **Deprecations:**
 
 - `CodeIgniter\Database\ModelFactory` is now deprecated in favor of `CodeIgniter\Config\Factories::models()`
-- `Config\Config` is now deprecated in favor of `CodeIgniter\Config\Factories::config()`
+- `CodeIgniter\Config\Config` is now deprecated in favor of `CodeIgniter\Config\Factories::config()`
 
 ## [v4.0.4](https://github.com/codeigniter4/CodeIgniter4/tree/v4.0.4) (2020-07-15)
 
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -27,6 +27,7 @@ parameters:
 		- '#Call to an undefined method CodeIgniter\\HTTP\\Request::(getPath|getSegments|getMethod|setLocale|getPost)\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Router\\RouteCollectionInterface::(getDefaultNamespace|isFiltered|getFilterForRoute|getRoutesOptions)\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Validation\\ValidationInterface::loadRuleGroup\(\)#'
+		- '#Call to an undefined static method CodeIgniter\\Config\\Factory::[a-z]+\(\)#'
 		- '#Call to function is_null\(\) with mysqli_stmt\|resource will always evaluate to false#'
 		- '#Cannot access property [\$a-z_]+ on ((bool\|)?object\|resource)#'
 		- '#Cannot call method [a-zA-Z_]+\(\) on ((bool\|)?object\|resource)#'
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -27,7 +27,7 @@ parameters:
 		- '#Call to an undefined method CodeIgniter\\HTTP\\Request::(getPath|getSegments|getMethod|setLocale|getPost)\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Router\\RouteCollectionInterface::(getDefaultNamespace|isFiltered|getFilterForRoute|getRoutesOptions)\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Validation\\ValidationInterface::loadRuleGroup\(\)#'
-		- '#Call to an undefined static method CodeIgniter\\Config\\Factory::[a-z]+\(\)#'
+		- '#Call to an undefined static method CodeIgniter\\Config\\Factories::[a-z]+\(\)#'
 		- '#Call to function is_null\(\) with mysqli_stmt\|resource will always evaluate to false#'
 		- '#Cannot access property [\$a-z_]+ on ((bool\|)?object\|resource)#'
 		- '#Cannot call method [a-zA-Z_]+\(\) on ((bool\|)?object\|resource)#'
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -27,7 +27,6 @@ parameters:
 		- '#Call to an undefined method CodeIgniter\\HTTP\\Request::(getPath|getSegments|getMethod|setLocale|getPost)\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Router\\RouteCollectionInterface::(getDefaultNamespace|isFiltered|getFilterForRoute|getRoutesOptions)\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Validation\\ValidationInterface::loadRuleGroup\(\)#'
-		- '#Call to an undefined static method CodeIgniter\\Config\\Factories::[a-z]+\(\)#'
 		- '#Call to function is_null\(\) with mysqli_stmt\|resource will always evaluate to false#'
 		- '#Cannot access property [\$a-z_]+ on ((bool\|)?object\|resource)#'
 		- '#Cannot call method [a-zA-Z_]+\(\) on ((bool\|)?object\|resource)#'
--- a/system/Config/Factories.php
+++ b/system/Config/Factories.php
@@ -107,7 +107,7 @@ class Factories
 		$config = self::getConfig(strtolower($component));
 
 		// First argument is the name, second is whether to use a shared instance
-		$name      = array_shift($arguments);
+		$name      = trim(array_shift($arguments), '\\ ');
 		$getShared = array_shift($arguments) ?? true;
 
 		if (! $getShared)
@@ -153,10 +153,7 @@ class Factories
 	protected static function locateClass(array $config, string $name): ?string
 	{
 		// Check for low-hanging fruit
-		if (class_exists($name)
-			&& (! $config['prefersApp'] || strpos($name, APP_NAMESPACE) === 0)
-			&& (! $config['instanceOf'] || is_a($name, $config['instanceOf'], true))
-		)
+		if (class_exists($name) && self::verifyPrefersApp($config, $name) && self::verifyInstanceOf($config, $name))
 		{
 			return $name;
 		}
@@ -164,7 +161,7 @@ class Factories
 		$basename = self::getBasename($name);
 
 		// Look for an App version if requested
-		$appname = APP_NAMESPACE . '\\' . $config['path'] . '\\' . $basename;
+		$appname = $config['component'] === 'config' ? 'Config\\' . $basename : APP_NAMESPACE . '\\' . $config['path'] . '\\' . $basename;
 		if ($config['prefersApp'] && class_exists($appname))
 		{
 			return $appname;
@@ -198,13 +195,59 @@ class Factories
 		{
 			$class = $locator->getClassname($file);
 
-			if ($class && (! $config['instanceOf'] || is_a($class, $config['instanceOf'], true)))
+			if ($class && self::verifyInstanceOf($config, $class))
 			{
 				return $class;
 			}
 		}
 
 		return null;
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Verifies that a class & config satisfy the "prefersApp" option
+	 *
+	 * @param array  $config The array of component-specific directives
+	 * @param string $name   Class name, namespace optional
+	 *
+	 * @return boolean
+	 */
+	protected static function verifyPrefersApp(array $config, string $name): bool
+	{
+		// Anything without that restriction passes
+		if (! $config['prefersApp'])
+		{
+			return true;
+		}
+
+		// Special case for Config since its App namespace is actually \Config
+		if ($config['component'] === 'config')
+		{
+			return strpos($name, 'Config') === 0;
+		}
+
+		return strpos($name, APP_NAMESPACE) === 0;
+	}
+
+	/**
+	 * Verifies that a class & config satisfy the "instanceOf" option
+	 *
+	 * @param array  $config The array of component-specific directives
+	 * @param string $name   Class name, namespace optional
+	 *
+	 * @return boolean
+	 */
+	protected static function verifyInstanceOf(array $config, string $name): bool
+	{
+		// Anything without that restriction passes
+		if (! $config['instanceOf'])
+		{
+			return true;
+		}
+
+		return is_a($name, $config['instanceOf'], true);
 	}
 
 	//--------------------------------------------------------------------
--- a/system/Config/Factories.php
+++ b/system/Config/Factories.php
@@ -48,6 +48,9 @@ use Config\Services;
  * and name. The "shared instance" implementation provides a
  * large performance boost and helps keep code clean of lengthy
  * instantiation checks.
+ *
+ * @method static \CodeIgniter\Model models(...$arguments)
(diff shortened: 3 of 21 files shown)


MGatner:
See what you all think of this. I will obviously add tests and docs but I want us to iron out the core class first. I would also like to see about implementing it for `Config`, though might need to tweak how it loads its own configuration if we do so they don't loop.

Also, should this move to **system/** root instead of being part of **system/Config**? Or even be part of **system/Autoload**?

michalsn:
I like how it looks. You added some nice config options so it's even better. I am very much into using this wherever it will bring some benefits.

I think it fits good to **system/Config** - alongside the `Services` class.

MGatner:
@michalsn Thanks for the feedback! I have some improvements in the works I'm excited about - I think this will becomes a very versatile core aspect of the framework when it is ready.

MGatner:
Thanks! I keep going back and forth on the names... @michalsn @lonnieezell have any input?

1. (current) `Factory` class to generate instances that uses a `Factories` config file.
2. `Factories` class that uses a `Factory` config file

IMO (1) is easier to type and makes more sense logically, but (2) is consistent with `Services` and sets up a nice parallel between the two.

michalsn:
@MGatner I think I like option 2 more, but I don't mind 1 either. Glad I could help 😅

lonnieezell:
@MGatner I'm ok with either, but do agree that option 2 parallels existing stuff in the framework better, keeping a nice consistency. And a decent IDE makes typing lots of things even easier than the naming itself. :)

MGatner:
Glad to hear support from both of you because I already changed it :) Got some test failures to work through and working on the User Guide, then this should be set for review.

system/Config/Factory.php
```diff
@@ -0,0 +1,80 @@
...
+ * @since      Version 4.0.0
+ * @filesource
+ */
+
+namespace CodeIgniter\Config;
+
+use CodeIgniter\Autoloader\Autoloader;
+use CodeIgniter\Autoloader\FileLocator;
+use Config\Autoload;
+use Config\Modules;
```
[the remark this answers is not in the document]
MGatner: Ooo right, thanks! Copy-paste mistake there, good catch.

system/Config/Factories.php
```diff
@@ -0,0 +1,333 @@
...
+	 * @param array  $arguments
+	 *
+	 * @return mixed
+	 */
+	public static function __callStatic(string $component, array $arguments)
+	{
+		$component = strtolower($component);
+
+		// Load the component-specific configuration
+		$config = self::getConfig($component);
```
mostafakhudair, line 110:
```suggestion
		// Load the component-specific configuration
		$config = self::getConfig(strtolower($component));
```

system/Config/Factories.php
```diff
@@ -0,0 +1,333 @@
...
+ * @link       https://codeigniter.com
+ * @since      Version 4.0.0
+ * @filesource
+ */
+
+namespace CodeIgniter\Config;
+
+use CodeIgniter\Autoloader\FileLocator;
+use CodeIgniter\Config\Factory;
+
```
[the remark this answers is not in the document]
MGatner: I got those with the other one, pushing commit soon. `Factory` is covered because it is the same namespace. `Services` technically is covered but I think it should be `Config\Services` to be thorough - will add.

MGatner:
Failing tests again :/ I don't have more time to work on this now - if anyone can figure out why the failures and post back I can take another shot at it later in the week.

MGatner:
Okay I see the problem: `App\Config\Whatever` is not a namespace path because we use `Config\Whatever` - I will have to add some special handling for that.

MGatner:
Down to what appears to be a discrepancy with `ReflectionClass::getStaticPropertyValue()` between PHP versions. I might need to find a way around using that.

paulbalandan (commented):
I think instead of putting the skip annotation *per test*, you can instead put it in the `setUp`.

MGatner:
> I think instead of putting the skip annotation _per test_, you can instead put it in the `setUp`.

But I don't want to skip all the tests, just those 3. Did I misunderstand?

paulbalandan:
> > I think instead of putting the skip annotation _per test_, you can instead put it in the `setUp`.
> 
> But I don't want to skip all the tests, just those 3. Did I misunderstand?

I did misunderstand. 😂 I just looked at the changes of the latest commit. Silly me.

MGatner:
I believe this is ready! It's possible that `prefersApp` will break some existing uses of `model()` and `config()`, but with the bug it's also possible that it replicates accidental desired behavior from the bug already.

system/Config/Factories.php
```diff
@@ -0,0 +1,382 @@
...
+		if (class_exists($name) && self::verifyPrefersApp($config, $name) && self::verifyInstanceOf($config, $name))
+		{
+			return $name;
+		}
+
+		// Determine the relative class names we need
+		$basename = self::getBasename($name);
+		$appname  = $config['component'] === 'config' ?
+			'Config\\' . $basename :
+			APP_NAMESPACE . '\\' . $config['path'] . '\\' . $basename;
```
paulbalandan, line 165:
```suggestion
			rtrim(APP_NAMESPACE, '\\') . '\\' . $config['path'] . '\\' . $basename;
```
Just to be sure.

system/Config/Factories.php
```diff
@@ -0,0 +1,382 @@
...
+	}
+
+	/**
+	 * Helper method for injecting mock instances
+	 *
+	 * @param string $component Lowercase, plural component name
+	 * @param string $name      The name of the instance
+	 * @param object $instance
+	 */
+	public static function injectMock(string $component, string $name, $instance)
```
paulbalandan, line 352:
```suggestion
	public static function injectMock(string $component, string $name, object $instance)
```

user_guide_src/source/concepts/factories.rst
```diff
@@ -0,0 +1,117 @@
...
+.. contents::
+    :local:
+    :depth: 2
+
+Introduction
+============
+
+Like ``Services``, ``Factories`` are an extension of autoloading that helps keep your code
+concise yet optimal, without having to pass around object instances between classes. At its
+simplest, ``Factories`` provides a common way to create a class instance and access it from
```
paulbalandan, line 14:
```suggestion
simplest, ``Factories`` provide a common way to create a class instance and access it from
```

user_guide_src/source/concepts/factories.rst
```diff
@@ -0,0 +1,117 @@
...
+anywhere. This is a great way to reuse object states and reduce memory load from keeping
+multiple instances loaded across your app.
+
+Anything can be loaded from a Factory, but the best examples are those classes that are used
+to work on or transmit common data. The framework itself makes heavy use of the ``Config``
+Factory to make sure the correct configuration is in use at any time.
```
paulbalandan, line 20:
```suggestion
Anything can be loaded from Factories, but the best examples are those classes that are used
to work on or transmit common data. The framework itself makes heavy use of the ``Factory``
config to make sure the correct configuration is in use at any time. 
```

user_guide_src/source/concepts/factories.rst
```diff
@@ -0,0 +1,117 @@
...
+multiple instances loaded across your app.
+
+Anything can be loaded from a Factory, but the best examples are those classes that are used
+to work on or transmit common data. The framework itself makes heavy use of the ``Config``
+Factory to make sure the correct configuration is in use at any time. 
+
+Take a look at ``Models`` as an example. You can access the Factory specific to ``Models``
+by using the magic static method of the Factories class, ``Factories::models()``. Because
+of the common path structure for namespaces and folders, ``Factories`` knows to look for
+files and classes within ``Models``, so you can request a model by its shorthand base name::
```
paulbalandan, line 25:
```suggestion
Take a look at ``Models`` as an example. You can access the Factory specific to ``Models``
by using the magic static method of the Factories class, ``Factories::models()``. Because
of the common path structure for namespaces and folders, ``Factories`` know where to look for
files and classes within ``Models``, so you can request a model by its shorthand base name::
```

user_guide_src/source/concepts/factories.rst
```diff
@@ -0,0 +1,117 @@
...
+
+	$widgets = Factories::models('Some\Namespace\Models\WidgetModel');
+
+Next time you ask for the same class anywhere in your code, ``Factories`` will be sure
+you get back the instance as before::
+
+	class SomeOtherClass
+	{
+		$widgets = Factories::models('WidgetModel');
+
```
paulbalandan, line 41:
```suggestion
	}
```

user_guide_src/source/concepts/factories.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Factories Behavior
+==================
+
+The default behavior of ``Factories`` might not always work for ever component. For example,
+say your component name and its path do not align, or you need to be sure that only a certain
+type of class instances are returned, or your component classes require additional parameters
+for their constructor. Learn how to use ``Factories`` and its configuration to meet your needs.
```
paulbalandan, line 49:
```suggestion
The default behavior of ``Factories`` might not always work for every component. For example,
say your component name and its path do not align, or you need to be sure that only a certain
type of class instance is returned, or your component classes require additional parameters
for their constructor. Learn how to use ``Factories`` and its configuration to meet your needs.
```


> or your component classes require additional parameters for their constructor.

Isn't the additional arguments for constructor handled by the splat operator (`...`)? I mean after the two `array_shift`, what is left of the `$arguments` is unpacked to the constructor.

user_guide_src/source/concepts/factories.rst
```diff
@@ -0,0 +1,117 @@
...
+Factory Configuration
+---------------------
+
+What works for one component may not work for all. ``Factories`` supports alternate behavior
+at the component level via configurations. A configuration consists of any of the settings,
+which will fall back on the default if not provided.
+
+======================= ============== ==================================================================================================================== ===================================================
+Key                     Type           Description                                                                                     Default
+======================= ============== ==================================================================================================================== ===================================================
```
paulbalandan, line 83: Python syntax check screams the table is malformed. I think this is because the dividers for `Description` and `Default` are continuous instead of broken. Also, I think the upper and bottom dividers are not of same length.

user_guide_src/source/testing/overview.rst
```diff
@@ -309,6 +309,24 @@ Removes all mocked classes from the Services class, bringing it back to its orig
 
 .. note:: The ``Email`` and ``Session`` services are mocked by default to prevent intrusive testing behavior. To prevent these from mocking remove their method callback from the class property: ``$setUpMethods = ['mockEmail', 'mockSession'];``
 
+Mocking Factory Instances
+=========================
+
+Similar to with Services, you may find yourself needing to supply a pre-configured class instance
```
paulbalandan, line 315:
```suggestion
Similar to Services, you may find yourself needing to supply a pre-configured class instance
```

MGatner:
Thanks for the review @paulbalandan! I will get to these shortly.

user_guide_src/source/concepts/factories.rst
MGatner: That's correct. The header section alludes to the Parameters section that covers this. I reworded it a bit so hopefully it is more clear.

MGatner:
Test failure is unrelated.

phpstan.neon.dist
```diff
@@ -27,6 +27,7 @@ parameters:
 		- '#Call to an undefined method CodeIgniter\\HTTP\\Request::(getPath|getSegments|getMethod|setLocale|getPost)\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Router\\RouteCollectionInterface::(getDefaultNamespace|isFiltered|getFilterForRoute|getRoutesOptions)\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Validation\\ValidationInterface::loadRuleGroup\(\)#'
+		- '#Call to an undefined static method CodeIgniter\\Config\\Factories::[a-z]+\(\)#'
```
paulbalandan, line 30: I'm curious what the actual error message is. Maybe we can add `Factories` as a [universal object crate class](https://phpstan.org/config-reference#universal-object-crates) or adding the `@property type $name` to the expected properties of this one.

paulbalandan (commented):
User guide syntax check still fails. Other than the phpstan ignored message, I approve. ;)

tests/system/Config/FactoriesTest.php
```diff
@@ -0,0 +1,264 @@
...
+		Factories::reset();
+	}
+
+	protected function getFactoriesStaticProperty(...$params)
+	{
+		// First parameter is the actual property
+		$name = array_shift($params);
+
+		$factory  = new ReflectionClass(Factories::class);
+		$property = $factory->getStaticPropertyValue($name);
```
michalsn, line 27: This is a permission issue on lower PHP versions. This way, we should be able to get rid of test skipping.
```suggestion
		$factory  = new ReflectionClass(Factories::class);
		$properties = $factory->getStaticProperties();
		$property = $properties[$name] ?? [];
```

michalsn (approved):
I think it looks really good.

We should only mention about `ModelFactory` depreciation in the changelog.

phpstan.neon.dist
MGatner: Ooo this is great! I didn't know about the crate objects. Now I have to update like 15 repos 😢

MGatner: Actually, crate classes are only concerned with properties, not methods. Same with the `@property` tag. I poked around the PHPStan repo and there was some recent movement on this issue (https://github.com/phpstan/phpstan/issues/3641) but the solutions are either to `mixin`  another class (not applicable) or declare the `@method`, which might work for our explicit calls but will not cover the generic uses like `Factories::foo('bar')` where you don't know "foo" ahead of time.

Here are the actual errors without the ignore:
```
 ------ ----------------------------------------- 
  Line   system/Common.php                        
 ------ ----------------------------------------- 
  256    Call to an undefined static method       
         CodeIgniter\Config\Factories::config().  
  872    Call to an undefined static method       
         CodeIgniter\Config\Factories::models().  
 ------ ----------------------------------------- 

 ------ ----------------------------------------- 
  Line   system/Config/Config.php                 
 ------ ----------------------------------------- 
  60     Call to an undefined static method       
         CodeIgniter\Config\Factories::config().  
 ------ ----------------------------------------- 

 ------ ----------------------------------------- 
  Line   system/Database/ModelFactory.php         
 ------ ----------------------------------------- 
  24     Call to an undefined static method       
         CodeIgniter\Config\Factories::models().  
 ------ ----------------------------------------- 
```

MGatner: Confirmed, adding the following to the class docblock does prevent the errors:
```
 * @method static \CodeIgniter\Model models(...$arguments)
 * @method static \Config\BaseConfig config(...$arguments)
```
We'll go with that for now but it won't solve the issue for projects downstream.

MGatner:
> User guide syntax check still fails.

@paulbalandan Any ideas on what is still wrong?

MGatner:
> We should only mention about `ModelFactory` depreciation in the changelog.

I added a note for both `ModelFactory` and `Config`

michalsn (changes requested):
Could you also add this to the changelog in the user guide? I think this is the main one, and the `changelog.md` is always generated semi-automatically before release.

MGatner:
Oops! Sorry, will do. I haven't touched the changelogs before, good for me to learn this :D

paulbalandan:
> > User guide syntax check still fails.
> 
> @paulbalandan Any ideas on what is still wrong?

Found the fault. After the preferApp row, you should add the bottom table borders that matches above borders.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## [Unreleased](https://github.com/codeigniter4/CodeIgniter4/tree/HEAD)
+
+[Full Changelog](https://github.com/codeigniter4/CodeIgniter4/compare/v4.0.4...HEAD)
+
+**Deprecations:**
+
+- `CodeIgniter\Database\ModelFactory` is now deprecated in favor of `CodeIgniter\Config\Factories::models()`
+- `Config\Config` is now deprecated in favor of `CodeIgniter\Config\Factories::config()`
```
paulbalandan, line 10:
```suggestion
- `CodeIgniter\Config\Config` is now deprecated in favor of `CodeIgniter\Config\Factories::config()`
```

user_guide_src/source/changelogs/v4.0.5.rst
```diff
@@ -23,3 +23,5 @@ Deprecations:
 - Deprecated ``BaseCommand::getPad`` in favor of ``BaseCommand::setPad``.
 - Deprecated ``Config\Format::getFormatter`` in favor of ``CodeIgniter\Format\Format::getFormatter``
 - Deprecated ``migrate:create`` command in favor of ``make:migration`` command.
+- Deprecated ``CodeIgniter\Database\ModelFactory`` in favor of ``CodeIgniter\Config\Factories::models()``
+- Deprecated ``Config\Config`` in favor of ``CodeIgniter\Config\Factories::config()``
```
paulbalandan, line 27:
```suggestion
- Deprecated ``CodeIgniter\Config\Config`` in favor of ``CodeIgniter\Config\Factories::config()``
```

paulbalandan (changes requested):
Last two. :)

paulbalandan (approved):

michalsn (approved):

paulbalandan:
I believe this is mergeable already. @MGatner any last additions?

MGatner:
I *think* it's ready, I'm just scared. 😳

paulbalandan:
I see this as a useful addition. 🎉

PR #3724, opened 2020-10-04
Test coverage once per PHP version

michalsn:
**Description**
This PR aims to speed up the tests by setting the `coverage_enable` flag only once per the PHP version. This should free up the available runners faster.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -19,7 +19,18 @@ jobs:
       fail-fast: false
       matrix:
         php-versions: ['7.2', '7.3', '7.4']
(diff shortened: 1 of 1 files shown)


michalsn:
Apparently Coveralls is smarter than I thought 😅 It looks like it takes into account all the test rounds - that's why we get the decrease (~2%) in coverage when I try to run them only once per PHP version.

The idea itself is great because tests without coverage are super fast and we can "release" runners really fast so they can be avaliable for other tasks.

But... since the test times are pretty good now, I don't think we need to sacrifice test coverage for that. However, it is a good option if we need it in the future.

PR #3725, opened 2020-10-04
Feature Tests with a Formatted Request Body

caswell-wc:
**Description**
I added the ability to set a body format on Feature Tests so that the request body can be set according to that format. This allows the Feature Tests to work on controllers that are using functions such as `$request->getJson()` or are expecting the body to contain a JSON or XML.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -403,23 +403,26 @@ trait FeatureTestTrait
 	 */
 	protected function setRequestBody(Request $request, $params = null)
 	{
-		if (empty($params))
+		if (! empty($this->bodyFormat))
 		{
-			$params = $request->fetchGlobal('request');
-		}
-		$formatMime = '';
-		if ($this->bodyFormat === 'json')
-		{
-			$formatMime = 'application/json';
-		}
-		elseif ($this->bodyFormat === 'xml')
-		{
-			$formatMime = 'application/xml';
-		}
-		if (! empty($formatMime))
-		{
-			$formatted = Services::format()->getFormatter($formatMime)->format($params);
-			$request->setBody($formatted);
+			if (empty($params))
+			{
+				$params = $request->fetchGlobal('request');
+			}
+			$formatMime = '';
+			if ($this->bodyFormat === 'json')
+			{
+				$formatMime = 'application/json';
+			}
+			elseif ($this->bodyFormat === 'xml')
+			{
+				$formatMime = 'application/xml';
+			}
+			if (! empty($formatMime))
+			{
+				$formatted = Services::format()->getFormatter($formatMime)->format($params);
+				$request->setBody($formatted);
+			}
 		}
 
 		return $request;
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -122,6 +122,19 @@ trait FeatureTestTrait
 	}
 
 	/**
+	 * Set the format the request's body should have.
+	 *
+	 * @param  string $format The desired format. Currently supported formats: xml, json
+	 * @return $this
+	 */
+	public function withBodyFormat(string $format)
+	{
+		$this->bodyFormat = $format;
+
+		return $this;
+	}
+
+	/**
 	 * Don't run any events while running this test.
 	 *
 	 * @return $this
@@ -165,6 +178,7 @@ trait FeatureTestTrait
 		$request = $this->setupRequest($method, $path);
 		$request = $this->setupHeaders($request);
 		$request = $this->populateGlobals($method, $request, $params);
+		$request = $this->setRequestBody($request);
 
 		// Make sure the RouteCollection knows what method we're using...
 		$routes = $this->routes ?? Services::routes();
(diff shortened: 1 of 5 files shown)


MGatner:
I like this a lot! Any chance we could get https://github.com/codeigniter4/CodeIgniter4/pull/3545 finished first so we could make sure this runs correctly with it?

system/Test/FeatureTestTrait.php
```diff
@@ -165,6 +178,7 @@ public function call(string $method, string $path, array $params = null)
 		$request = $this->setupRequest($method, $path);
 		$request = $this->setupHeaders($request);
 		$request = $this->populateGlobals($method, $request, $params);
+		$request = $this->setRequestBody($request);
```
MGatner, line 181: I'm curious about this addition. Why was this necessary? What happens without it?

caswell-wc: That's the method that actually does the work of formatting the parameters and putting them on the body.

caswell-wc:
@MGatner This shouldn't affect #3545 because that is working with the response while this is working with the request. I am still planning on working on that one as well but I was waiting to hear back on how we should proceed with it.

system/Test/FeatureTestTrait.php
```diff
@@ -376,4 +390,41 @@ protected function populateGlobals(string $method, Request $request, array $para
...
+	 * Set the request's body formatted according to the value in $this->bodyFormat.
+	 * This allows the body to be formatted in a way that the controller is going to
+	 * expect as in the case of testing a JSON or XML API.
+	 *
+	 * @param  Request    $request
+	 * @param  null|array $params  The parameters to be formatted and put in the body. If this is empty, it will get the
+	 *                               what has been loaded into the request global of the request class.
+	 * @return mixed
+	 */
+	protected function setRequestBody(Request $request, $params = null)
```
samsonasik, line 404: I think `array` type hint can be added for `$params` with returns `Request`

```suggestion
      protected function setRequestBody(Request $request, array $params = null): Request
```

system/Test/FeatureTestTrait.php
```diff
@@ -376,4 +390,41 @@ protected function populateGlobals(string $method, Request $request, array $para
...
+
+	/**
+	 * Set the request's body formatted according to the value in $this->bodyFormat.
+	 * This allows the body to be formatted in a way that the controller is going to
+	 * expect as in the case of testing a JSON or XML API.
+	 *
+	 * @param  Request    $request
+	 * @param  null|array $params  The parameters to be formatted and put in the body. If this is empty, it will get the
+	 *                               what has been loaded into the request global of the request class.
+	 * @return mixed
```
samsonasik, line 402: can use return Request: 

```suggestion
       * @return Request

**Review by michalsn — COMMENTED** [review]
It would be nice if `withBodyFormat()` method would actually set a `Content-Type` for reaquest.

Also, maybe it would be a good idea to add `withBody()` method as well? When it comes to working with XML as input, there is a big chance that this document will be really complicated and not really something that can be handled by `Formatter` class.

File: system/Test/FeatureTestTrait.php (hunk 2 above)
Comment (caswell-wc) [reply system/Test/FeatureTestTrait.php]: I made this change in my latest commit.

File: system/Test/FeatureTestTrait.php (hunk 3 above)
Comment (caswell-wc) [reply system/Test/FeatureTestTrait.php]: I made this change in my latest commit.

**Comment by caswell-wc** [comment]
@michalsn Those are some good suggestions. I have included them in the latest commit to this PR.

**Review by paulbalandan — APPROVED** [review]

**Comment by caswell-wc** [comment]
@michalsn @MGatner @samsonasik Is there anything else that I need to do to get this approved and merged?

File: system/Test/FeatureTestTrait.php [hunk 4]
```diff
@@ -376,4 +403,48 @@ protected function populateGlobals(string $method, Request $request, array $para
...
+	 * expect as in the case of testing a JSON or XML API.
+	 *
+	 * @param  Request    $request
+	 * @param  null|array $params  The parameters to be formatted and put in the body. If this is empty, it will get the
+	 *                               what has been loaded into the request global of the request class.
+	 * @return Request
+	 */
+	protected function setRequestBody(Request $request, array $params = null): Request
+	{
+		if (! empty($this->requestBody))
```
samsonasik, line 419: empty can be "0", probably ensure if it is a string, compare with ""

system/Test/FeatureTestTrait.php
```diff
@@ -376,4 +403,48 @@ protected function populateGlobals(string $method, Request $request, array $para
...
+		{
+			$request->setBody($this->requestBody);
+			return $request;
+		}
+
+		if (! empty($this->bodyFormat))
```
samsonasik, line 425: compare with empty string as well

MGatner:
@caswell-wc Two small changes form samsonasik- can we wrap this up?

caswell-wc:
@MGatner @samsonasik I have made those changes. Sorry about the long wait. Life has been super crazy lately.

samsonasik (approved):

MGatner (approved):
No worries! I am as well, totally understandable. I'm glad to be getting this wrapped up before 4.1 though - thank you!

PR #3730, opened 2020-10-06
Load helpers directly from the initController

mostafakhudair:
remove loadHelpers  method while it isn't used somewhere else, since we can use helper() method anywhere

### DIFF
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -31,7 +31,7 @@
  * @package    CodeIgniter
  * @author     CodeIgniter Dev Team
  * @copyright  2019-2020 CodeIgniter Foundation
- * @license    https://opensource.org/licenses/MIT	MIT License
+ * @license    https://opensource.org/licenses/MIT - MIT License
  * @link       https://codeigniter.com
  * @since      Version 4.0.0
  * @filesource
@@ -53,7 +53,6 @@ use Psr\Log\LoggerInterface;
  */
 class Controller
 {
-
 	/**
 	 * An array of helpers to be automatically loaded
 	 * upon class instantiation.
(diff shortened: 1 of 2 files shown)


samsonasik:
While it is probably reasonable, protected method can be used in userland. I think it is better to mark it as `@deprecated` and warn that it will be removed in 4.1 for example.

system/Controller.php
```diff
@@ -123,7 +120,11 @@ public function initController(RequestInterface $request, ResponseInterface $res
 			$this->forceHTTPS($this->forceHTTPS);
 		}
 
-		$this->loadHelpers();
+		// Autoloading helper files.
+		if (! empty($this->helpers))
+		{
+			helper($this->helpers);
+		}
```
paulbalandan, line 127: I agree on deprecating it first before outright remove. I may be using it in a controller code elsewhere, then suddenly my application will break for unknown method.

Furthermore, the `empty` check can be removed, since `helper` already manages loading via arrays thru `foreach`. Executing `foreach` on empty arrays will just do nothing.

```suggestion
		helper($this->helpers);
```

system/Controller.php
```diff
@@ -160,21 +161,6 @@ protected function cachePage(int $time)
...
-	 */
-	protected function loadHelpers()
-	{
-		if (empty($this->helpers))
-		{
-			return;
-		}
-
-		helper($this->helpers);
-	}
```
paulbalandan, line 174:
```suggestion
	/**
	 * Handles "auto-loading" helper files.
	 *
	 * @deprecated This is just a helper method for `helper`. Use `helper` directly.
	 */
	protected function loadHelpers()
	{
		if (empty($this->helpers))
		{
			return;
		}

		helper($this->helpers);
	}
```

system/Controller.php
mostafakhudair: I think we need to keep check if helpers property not empty to avoid running helper method script and do unnecessary loads.

michalsn: The `helper()` function has no default value so checking is required.

michalsn:
Info in the code is not enough. We would have to add info about deprecation to the [changelog](https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/changelogs/v4.0.5.rst) as well.

system/Controller.php
paulbalandan: I don't follow. `$this->helpers` is already initialized with an empty array. So passing this directly to `helper` function will just do nothing. Am I missing something?

michalsn: Oh... you're right, this time I missed something 😅

paulbalandan: > I think we need to keep check if helpers property not empty to avoid running helper method script and do unnecessary loads.

I don't think it is an unnecessary load as the common functions are already loaded long before the controller is instantiated.

michalsn (approved):

paulbalandan (approved):

PR #3734, opened 2020-10-07
Remove dump-autoload in post-update-cmd

paulbalandan:
**Description**
I don't know the original intent for this but `composer dump-autoload` is already called after every `install`, `update`, `remove` operation, so there's no need to call it again. If there's a reasoning behind this, please correct me. 😊

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/admin/framework/composer.json
+++ b/admin/framework/composer.json
@@ -29,7 +29,6 @@
 	},
 	"scripts": {
 		"post-update-cmd": [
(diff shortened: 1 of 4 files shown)


MGatner:
I'm not sure where I picked up that habit but that was certainly my doing.

MGatner (approved):

michalsn (approved):

PR #3735, opened 2020-10-07
Run phpunit tests on changed PHP files only

paulbalandan:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -5,6 +5,12 @@ on:
     branches:
       - develop
       - '4.*'
+    paths:
+      - 'app/**'
+      - 'public/**'
+      - 'system/**'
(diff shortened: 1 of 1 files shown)


michalsn:
Should we add `composer.json` to this?

paulbalandan:
Added more files.

michalsn:
I guess the `test-phpunit.yml` file should also be added.

michalsn (approved):

MGatner:
**utils/** just has the Rector parameters and I don't think should be included. @samsonasik?

samsonasik:
@MGatner yes, `utils` can be excluded. If we want to add `utils`, it should be added to `test-rector.yml`.

samsonasik (approved):

PR #3736, opened 2020-10-07
Streamline gitattributes

paulbalandan:
**Description**
When using the development branch as dependency for a module, or just even using in `appstarter` for catching up with the latest changes, the downloaded package is always "bloated" with regard to unnecessary files and folders. This PR attempts to clean up the distributable package (when using `--prefer-dist`) by putting in `.gitattributes` (through `export-ignore`) the "unusable" files.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,19 +5,27 @@
 
 # git files
 .gitattributes export-ignore
-# .gitignore
+.gitignore export-ignore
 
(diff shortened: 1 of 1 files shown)


lonnieezell (approved):
Nice changes. Thanks!

PR #3695, opened 2020-09-29
Add `db:create` Command to Create Database

26Piedrosa:
**Description**
Create database using `php spark db:create <db_name>`

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPdocs
- [x] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

PR #3746, opened 2020-10-09
Create database command

paulbalandan:
**Description**
Supersedes and closes #3695 . 

Added special handling for SQLite3 connections.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Database/CreateDatabase.php
+++ b/system/Commands/Database/CreateDatabase.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * CodeIgniter
  *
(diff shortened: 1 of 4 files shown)


paulbalandan:
I don't know why it went wrong for postgres.

michalsn:
Seem like a general issue with `createDatabase()`. The produced query should be like: `CREATE DATABASE "user"`. The escape character should be applied around the database name.

michalsn:
You know what... forget my last comment. It doesn't work because `user` is a restricted word in PostgreSQL. Change the database name for something else and we should be good.

paulbalandan:
> You know what... forget my last comment. It doesn't work because `user` is a restricted word in PostgreSQL. Change the database name for something else and we should be good.

Okay. Didn't know that. Changing. Thanks.

paulbalandan:
Off-topic to this PR but related to database creation.

It took me a long time pondering why SQLite's Forge didn't create databases using `createDatabase` without going throughout the code. Why was it set to return `true` instead of throwing a `DatabaseException`?

I mean in the parent `Forge` there's a check if the db creation is disabled for the driver by setting the `createDatabaseStr` to `false`. So sqlite's createDatabase can be something like:
```php
	public function createDatabase(string $dbName, bool $ifNotExists = false): bool
	{
		$this->createDatabaseStr = false;

		return parent::createDatabase($dbName, $ifNotExists);
	}
```

michalsn:
Regarding SQLite, I can't tell really. Your idea seems good to me though.

michalsn (approved):

PR #3749, opened 2020-10-09
Fix bug in Parser with similar variable names

### LINKED ISSUE #3726: Bug: Parser will overwrite the value if the initial variable is same
**Direction**
when using parser to render view, the parser can't differentiate between {expired_discount} with {expired_discount_date}

**Describe the bug**
Code example
Controller:
```
<?php

namespace App\Controllers;

class Test extends BaseController
{
	public function index()
	{
		$data = [
			'expired_discount'           => '1212312423470',
			'expired_discount_date'  => '1999-09-09',
		];
		echo $this->parser->setData($data)->render('test');
	}
}
```

View:
```
<p>Expired Discount Serial = {expired_discount}</p>
<p>Expired Discount Date = {expired_discount_date}</p>
```

Result:
```
Expired Discount Serial = 1212312423470

Expired Discount Date = 1212312423470
```

expired_discount_date will have the same value as expired_discount

**CodeIgniter 4 version**
branch: develop

value of expired_discount should not be the same with expired_discount_date

**Context**
 - OS: Window 10
 - Apache 2.4.46
 - 7.4.10

michalsn:
**Description**
This PR fixes error in `Parser` when similar variable names would be treated as the same.

Fixes #3726

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -333,7 +333,7 @@ class Parser extends View
 	 */
 	protected function parseSingle(string $key, string $val): array
 	{
-		$pattern = '#' . $this->leftDelimiter . '!?\s*' . preg_quote($key) . '\s*\|*\s*([|\w<>=\(\),:.\-\s\+\\\\/]+)*\s*!?' . $this->rightDelimiter . '#ms';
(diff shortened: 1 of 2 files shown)


hurtsky:
Looks like this PR fixed the issue. Thanks

samsonasik (approved):

samsonasik:
Thank you @michalsn

PR #3751, opened 2020-10-10
Fix routes options to be recognizable via used HTTP request method

### LINKED ISSUE #3700: Bug: getRoutesOptions return wrong when I have >2 routes has same name, but different method (HTTPVerb)
//my routes

`$routes->get('product/add', 	'App\Product::add', ['as' => 'app_product_add']);`
`$routes->post('product/add', 'App\Product::add_ajax', ['as' => 'app_product_add_ajax']);`

//my function 
`$router  = \Config\Services::router();`
`$match  = $router->getMatchedRoute(); // => IT'S OK`
`$options = $router->getMatchedRouteOptions(); //=> WRONG BY METHOD`

$options alway is: `['as' => 'app_product_add_ajax']`
I want setMethod() and default method by current request in here for return right options

How can't I hot fix this bug?

Help me.
Thanks

### LINKED ISSUE #3733: Bug: route filters don't apply for different methods
**Direction**
same route filter is applied for different method

**Describe the bug**
for e.g.
~~~~
$routes->get('', AdminController@index, ['filter' => 'permission:index']);
$routes->delete('', AdminController@index, ['filter' => 'permission:delete]);
~~~~
If I call either method I will get filter params as 'delete' not 'index' even for get requests.

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
Route filter

I index request is called route parameters should be index.

**Context**
 - OS:Ubuntu
 - Web server: apache2
 - PHP version : 7.3

michalsn:
**Description**
This PR fixes routes options to be recognizable via used HTTP request method. Right now routes options are shared - one instance per route. It leads to problems like: wrong named route name returned or firing incorrect filter.

I don't think this will introduce any BC, rather will fix buggy behavior, but if anyone notices the problem please let me know.

Fixes #3700 
Fixes #3733

**Checklist:**
- [x] Securely signed commits
- [ ] N/A User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -558,7 +558,7 @@ class RouteCollection implements RouteCollectionInterface
 	 * Returns one or all routes options
 	 *
 	 * @param string $from
-	 * @param string $verb
+       * @param string|null $verb
 	 *
 	 * @return array
 	 */
@@ -1288,7 +1288,7 @@ class RouteCollection implements RouteCollectionInterface
 	 * Checks a route (using the "from") to see if it's filtered or not.
 	 *
 	 * @param string $search
-	 * @param string $verb
+       * @param string|null $verb
 	 *
 	 * @return boolean
 	 */
@@ -1312,7 +1312,7 @@ class RouteCollection implements RouteCollectionInterface
 	 * has a filter of "role", with parameters of ['admin', 'manager'].
 	 *
 	 * @param string $search
-	 * @param string $verb
+       * @param string|null $verb
 	 *
 	 * @return string
 	 */
@@ -1591,16 +1591,13 @@ class RouteCollection implements RouteCollectionInterface
 	/**
 	 * Load routes options based on verb
 	 *
-	 * @param string $verb
+       * @param string|null $verb
 	 *
 	 * @return array
 	 */
 	protected function loadRoutesOptions(string $verb = null): array
 	{
-		if (empty($verb))
-		{
-			$verb = $this->getHTTPVerb();
-		}
+		$verb = $verb ?: $this->getHTTPVerb();
 
 		$options = $this->routesOptions[$verb] ?? [];
 
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -514,7 +514,7 @@ class RouteCollection implements RouteCollectionInterface
 	/**
 	 * Returns the raw array of available routes.
 	 *
-	 * @param mixed $verb
+	 * @param string|null $verb
 	 *
 	 * @return array
 	 */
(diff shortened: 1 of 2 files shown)


system/Router/RouteCollection.php
```diff
@@ -558,12 +558,15 @@ public function getRoutes($verb = null): array
 	 * Returns one or all routes options
 	 *
 	 * @param string $from
+	 * @param string $verb
```
samsonasik, line 561:
```suggestion
       * @param string|null $verb
```

system/Router/RouteCollection.php
```diff
@@ -1285,12 +1288,15 @@ protected function localizeRoute(string $route) :string
 	 * Checks a route (using the "from") to see if it's filtered or not.
 	 *
 	 * @param string $search
+	 * @param string $verb
```
samsonasik, line 1291:
```suggestion
       * @param string|null $verb
```

system/Router/RouteCollection.php
```diff
@@ -1306,17 +1312,15 @@ public function isFiltered(string $search): bool
 	 * has a filter of "role", with parameters of ['admin', 'manager'].
 	 *
 	 * @param string $search
+	 * @param string $verb
```
samsonasik, line 1315:
```suggestion
       * @param string|null $verb
```

system/Router/RouteCollection.php
```diff
@@ -1582,4 +1586,35 @@ public function resetRoutes()
 		}
 	}
 
+	//--------------------------------------------------------------------
+
+	/**
+	 * Load routes options based on verb
+	 *
+	 * @param string $verb
```
samsonasik, line 1594:
```suggestion
       * @param string|null $verb
```

MGatner (changes requested):
I like this. I agree with @samsonasik's changes, also... I know `getRoutes()` has an unspecified variable type on `$verb` but other things will break if it is not `string|null` so I think we should go ahead and update the function declaration to be consistent.

system/Router/RouteCollection.php
```diff
@@ -1582,4 +1586,35 @@ public function resetRoutes()
...
+	 *
+	 * @return array
+	 */
+	protected function loadRoutesOptions(string $verb = null): array
+	{
+		if (empty($verb))
+		{
+			$verb = $this->getHTTPVerb();
+		}
```
paulbalandan, line 1603:
```suggestion
	$verb = $verb ?: $this->getHTTPVerb();
```

michalsn:
Thanks for your code suggestions guys.

MGatner (approved):

element-code:
👍 i was writing a bug report just at this moment and saw the merge while digging for the buggy code 🎆

PR #3754, opened 2020-10-10
Optimize Model

paulbalandan:
**Description**
Optimized `Model`.
- Removes unneeded variables.
- Uses truthy checks and null coalescing operators where possible
- Make setter methods return `$this` for easy method chaining.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -73,12 +73,10 @@ use stdClass;
  *      - allow specifying the return type (array, object, etc) with each call
  *      - ensure validation is run against objects when saving items
  *
- * @package CodeIgniter
- * @mixin   BaseBuilder
+ * @mixin BaseBuilder
  */
 class Model
 {
-
 	/**
 	 * Pager instance.
 	 * Populated after calling $this->paginate()
@@ -165,8 +163,6 @@ class Model
 	 * @var string
 	 */
 	protected $dateFormat = 'datetime';
-
-	//--------------------------------------------------------------------
 
 	/**
 	 * The column used for insert timestamps
@@ -268,11 +264,14 @@ class Model
 	protected $validation;
 
 	/*
-	 * Callbacks. Each array should contain the method
-	 * names (within the model) that should be called
-	 * when those events are triggered. "Update" and "delete"
-	 * methods are passed the same items that are given to
-	 * their respecitve method.
+	 * Callbacks.
+	 *
+	 * Each array should contain the method names (within the model)
+	 * that should be called when those events are triggered.
+	 *
+	 * "Update" and "delete" methods are passed the same items that
+	 * are given to their respective method.
+	 *
 	 * "Find" methods receive the ID searched for (if present), and
 	 * 'afterFind' additionally receives the results that were found.
 	 */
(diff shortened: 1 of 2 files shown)


samsonasik (approved):

system/Model.php
```diff
@@ -2001,14 +1911,14 @@ public function __get(string $name)
 			return $this->{$name};
 		}
 
-		if (isset($this->db->$name))
```
MGatner, line 2004: Do the `{` braces add something here? I never use them unless they are syntactically required.

system/Model.php
```diff
@@ -1247,66 +1209,55 @@ public function purgeDeleted()
...
 	 * Sets $useSoftDeletes value so that we can temporarily override
 	 * the softdeletes settings. Can be used for all find* methods.
 	 *
 	 * @param boolean $val
 	 *
-	 * @return Model
+	 * @return $this
 	 */
-	public function withDeleted($val = true)
+	public function withDeleted(bool $val = true)
```
MGatner, line 1224: This is the only potential break I see, for anyone passing in non-booleans , but since we have the doc block I think it is safe.

system/Model.php
michalsn: Same here. Maybe `{` make it more readable for people? Either way, we should remove them or keep them for all cases.

michalsn (approved):

system/Model.php
paulbalandan: Since we are accessing the properties of either `$this->db` and `$this->builder` dynamically using the value of `$name`, it is recommended that this dynamic access be made *explicit* by surrounding with curly braces. This helps to disambiguate usage by making it clear that the enclosed is dynamic. This apparently helps also in refactoring.

Also, migration from PHP 5 to PHP 7 has some changes in semantics in how variable variables are interpreted by the parser. This [guide](https://www.php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling.indirect) shows the differences between PHP 5 and PHP 7 and how an explicit `{ }` can help in disambiguation.

`php-cs-fixer` rule: `explicit_indirect_variable`

Readings:
https://stackoverflow.com/questions/35598187/when-to-wrap-curly-braces-around-a-variable
https://www.php.net/manual/en/language.variables.variable.php

system/Model.php
paulbalandan: I don't see it as breaking since the user guide did not mention it to be accepting a required argument, albeit having a default value. If someone is using a reputable IDE the doc block will surely inform that it is accepting booleans as param.

system/Model.php
MGatner: I get the disambiguation, so I definitely agree with uses like below where you made this change:
`$result = $this->db->{$name}(...$params);`

But in cases where it is a simple variable property and there is no other way to interpret `$this->db->$name` it seems redundant and definitely not something I want to go through and apply everywhere in the framework nor in my code. If for some reason it helps static analysis or Rector or something I can get that, but if it is just to be consistent with the disambiguation cases then I would be in favor of dropping them.

paulbalandan: Ok, granted I would revert the changes, there are still places in the framework where this is applied. For example, old line 2011 of this `Model`, or in the constructor of `CodeIgniter\HTTP\ContentSecurityPolicy`. If these would be left out, then it would be inconsistent. Besides, you do not need to manually go over the framework to change these lines. You can use code style checkers like what I've said earlier, `php-cs-fixer` has this rule `explicit_indirect_variable` to make it consistent across the framework.

MGatner: It's true, we definitely aren't consistent. I say just leave it how it is and if we want to make a pass sometime down the road we can.

Overall this is a great move for Model! I love the clarity it brings to the code.

paulbalandan:
`composer remove --dev phpstan/phpstan` is failing

samsonasik:
@paulbalandan that fixed at https://github.com/codeigniter4/CodeIgniter4/pull/3772 , I will merge that first.

samsonasik:
@paulbalandan please rebase, should be ok now.

PR #3758, opened 2020-10-11
Don't instantiate Faker on Seeder constructor

paulbalandan:
**Description**
This attempts to hasten database live tests by extracting `Faker` instantiation from `Seeder`'s constructor to a new static method and using a static property so that it will persist across instances. I've added tests for uncovered lines.

My only worry for this is that the `use` statement for `Faker\Factory` may break if the library is not installed for manual installations. @samsonasik I've tried using `@noRector` but still it imports them. Any ideas?

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Seeder.php
+++ b/system/Database/Seeder.php
@@ -42,6 +42,7 @@ namespace CodeIgniter\Database;
 use CodeIgniter\CLI\CLI;
 use Config\Database;
 use Faker\Factory;
+use Faker\Generator;
 use InvalidArgumentException;
 
 /**
@@ -85,18 +86,18 @@ class Seeder
 	protected $forge;
 
 	/**
-	 * Faker Generator instance.
-	 *
-	 * @var \Faker\Generator|null
-	 */
-	protected $faker;
-
-	/**
 	 * If true, will not display CLI messages.
 	 *
 	 * @var boolean
 	 */
 	protected $silent = false;
+
+	/**
+	 * Faker Generator instance.
+	 *
+	 * @var \Faker\Generator|null
+	 */
+	private static $faker;
 
 	/**
 	 * Seeder constructor.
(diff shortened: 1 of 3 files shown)


samsonasik:
`use` statement won't break existing manual installation as far as there is a `class_exists()` check, it is just statement.

samsonasik:
For `@noRector` usage, you can do like the following:

```php
/** @noRector \Utils\Rector\UnderscoreToCamelCaseVariableNameRector */
$a_b_c;
```

paulbalandan:
Thanks for clarifying. So I'm just fearing for nothing. 🤣

system/Database/Seeder.php
```diff
@@ -126,7 +126,21 @@ public function __construct(Database $config, BaseConnection $db = null)
...
 		$this->forge = Database::forge($this->DBGroup);
-		$this->faker = class_exists(Factory::class) ? Factory::create() : null;
+	}
+
+	/**
+	 * Gets the Faker Generator instance.
+	 *
+	 * @return \Faker\Generator|null
+	 */
+	public static function faker()
```
samsonasik, line 136: You change with this:

```suggestion
      public static function faker(): ?\Faker\Generator
```

and run :

```bash
vendor/bin/rector process system/Database/Seeder.php
```

so it should will automatically add use statement for `use Faker\Generator` to use statements. It doesn't sort use statements so probably need manual check or use tools like php-cs-fixer to sort it, eg:

```php
php-cs-fixer fix system/Database/Seeder.php --rules=ordered_imports
```

michalsn:
Can we add info about `Faker` being available (and example how to access it) to the user guide? I think I missed it last time.

michalsn (approved):

PR #3759, opened 2020-10-11
add rootDirectory property

mostafakhudair:
Add \Config\Paths::rootDirectory holds project main folders
Optimize bootstrap


MGatner:
I'm generally not a fan of defining more directories to track. Can you explain what this allows that isn't available by relative paths to existing directory definitions?

system/Test/bootstrap.php
```diff
@@ -33,7 +33,7 @@
 defined('APPPATH')       || define('APPPATH', realpath(rtrim($paths->appDirectory, '\\/ ')) . DIRECTORY_SEPARATOR);
 defined('WRITEPATH')     || define('WRITEPATH', realpath(rtrim($paths->writableDirectory, '\\/ ')) . DIRECTORY_SEPARATOR);
 defined('SYSTEMPATH')    || define('SYSTEMPATH', realpath(rtrim($paths->systemDirectory, '\\/')) . DIRECTORY_SEPARATOR);
-defined('ROOTPATH')      || define('ROOTPATH', realpath(APPPATH . '../') . DIRECTORY_SEPARATOR);
+defined('ROOTPATH')      || define('ROOTPATH', realpath(rtrim($paths->rootDirectory, '\\/ ')) . DIRECTORY_SEPARATOR);
```
samsonasik, line 36: it is bc break without property_exists check

system/bootstrap.php
```diff
@@ -37,63 +31,45 @@
...
  * @filesource
  */
 
-/*
+use CodeIgniter\CodeIgniter;
+use CodeIgniter\Config\DotEnv;
+use CodeIgniter\Services;
+use Config\App;
+use Config\Autoload;
+use Config\Modules;
```
samsonasik, line 45: I created PHPStan rule for it at https://github.com/codeigniter4/CodeIgniter4/pull/3786, no need to check manually ;)

samsonasik:
please revert unrelated change to PR stuff, also it is BC break as I mentioned before without property_exists/isset check.

PR #3772, opened 2020-10-15
Bug fix: ensure to use autoload = false in class_exists() in low-hanging fruit in Factories::locateClass()

samsonasik:
I tried latest `dev-develop` when running test in a codeigniter4 module and without set `autoload=false` in 2nd parameter of `class_exists()` in `Factories::locateClass()`, I got the error:

```bash
Cannot declare class Config\App, because the name is already in use

at APPPATH/Config/App.php:7

Backtrace:
  1    [internal function]
       CodeIgniter\Debug\Exceptions()->shutdownHandler()


Fatal error: Cannot declare class Config\App, because the name is already in use in /Users/samsonasik/www/codeigniter4-files/vendor/codeigniter4/codeigniter4/app/Config/App.php on line 7
```

This patch fixed the issue.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -98,10 +98,10 @@ jobs:
       - name: Install dependencies
         run: |
           composer install --no-progress --no-suggest --no-interaction --prefer-dist --optimize-autoloader
-          composer remove --dev phpstan/phpstan
-          composer remove --dev rector/rector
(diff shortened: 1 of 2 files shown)


samsonasik:
@MGatner above error happen when I tried to test your `codeigniter4-files` module in php 7.4.9. Update the `autoload` to false in 2nd `class_exists` above fix the issue.

PR #3775, opened 2020-10-15
report coveralls only in specific php version

samsonasik:
@michalsn I think report coveralls can be only in specific php version to handle false positive decreased coverage.

As I see, it seems the stable report in php <strike>7.3</strike> 7.4. I added if php version check to handle it. Please verify if it is ok.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -124,8 +124,7 @@ jobs:
     needs: [tests]
     runs-on: ubuntu-latest
     steps:
-      - if: matrix.php-versions == '7.4'
-        name: Coveralls Finished
+      - name: Coveralls Finished
         uses: coverallsapp/github-action@master
         with:
           github-token: ${{ secrets.github_token }}
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -119,3 +119,14 @@ jobs:
           COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           COVERALLS_PARALLEL: true
           COVERALLS_FLAG_NAME: PHP ${{ matrix.php-versions }} ${{matrix.db-plaftorms}}
+
+  coveralls-finish:
+    needs: [tests]
+    runs-on: ubuntu-latest
+    steps:
+      - if: matrix.php-versions == '7.4'
+        name: Coveralls Finished
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.github_token }}
+          parallel-finished: true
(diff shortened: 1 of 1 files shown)


samsonasik:
It seems `coveralls-finish` job actually not needed, as the `php-coveralls --coverage_clover=build/logs/clover.xml -v` already upload to coveralls site

samsonasik:
@michalsn All green 🎉

.github/workflows/test-phpunit.yml
```diff
@@ -110,21 +110,12 @@ jobs:
...
-      - name: Run Coveralls
+      - if: matrix.php-versions == '7.4'
+        name: Run Coveralls
         run: |
           composer global require php-coveralls/php-coveralls:^2.4
           php-coveralls --coverage_clover=build/logs/clover.xml -v
         env:
           COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           COVERALLS_PARALLEL: true
           COVERALLS_FLAG_NAME: PHP ${{ matrix.php-versions }}
```
michalsn, line 121: Can we try to add DB here? I'm curious if this will give us usable info.
```suggestion
          COVERALLS_FLAG_NAME: PHP ${{ matrix.php-versions }} ${{matrix.db-plaftorms}}
```

.github/workflows/test-phpunit.yml
```diff
@@ -110,21 +110,12 @@ jobs:
...
-
-  coveralls-finish:
-    needs: [tests]
-    runs-on: ubuntu-latest
-    steps:
-      - name: Coveralls Finished
-        uses: coverallsapp/github-action@master
-        with:
-          github-token: ${{ secrets.github_token }}
-          parallel-finished: true
```
michalsn, line 130: We have to restore it. Since our build run in parallel Coveralls needs to know when the job is finished. Now the page with results says: "Pending completion".

michalsn (changes requested):
Hmm... I think we can try. We don't really have a "special" code written for a given PHP version (except for one place I know of), so if it will help handle false decreasing coverage reports it should be good.

.github/workflows/test-phpunit.yml
samsonasik: implemented.

.github/workflows/test-phpunit.yml
samsonasik: implemented with add `- if: matrix.php-versions == '7.4'` check

michalsn: I don't think there is a `matrix` variable available here. Also, it's not needed since it runs only once for all tests.

samsonasik: ok, removed if check.

michalsn: Thanks.

samsonasik:
@michalsn All green 🎉

PR #3776, opened 2020-10-15
[RFC][Dev] Use php-cs-fixer as code style checker

paulbalandan:
**Description**
This PR is a bid to use `php-cs-fixer` in lieu of `php_codesniffer` as the main code style checker for the framework.

***Changes to workflow***
`php-cs-fixer`, by default, uses `.php_cs.dist` as its configuration file. This will contain all the rules to be applied across the entire codebase. However, according to the authors themselves, there is no way to disable certain rule(s) for each file, like `@noRector`, `phpcs:disable`, or `@codeCoverage` for other static code analysis libraries. I am mentioning this because of the specific rule `header_comment` which includes a file-level document block to `system` files, while other files (`app`, `public`, `tests`) have no file-level docblocks.

To achieve this, I made two configuration files:
- `system.php_cs.dist` - Applies the CodeIgniter4 rules plus the `header_comment` rule to system files
- `no_header.php_cs.dist` - Applies to other PHP files (app, public, tests) without the `header_comment` rule

To manually apply the rules to PHP files, you can use:
`vendor/bin/php-cs-fixer fix --verbose --config=system.php_cs.dist` for system files
`vendor/bin/php-cs-fixer fix --verbose --config=no_header.php_cs.dist` for other files

This can be done automatically with the changes in the `pre-commit` hook. The script is changed wherein after the staged PHP files are scanned by PHPStan, these are separated whether a `system` file or other file, then applied using the syntax above.

***Notable changes to codebase***
- PHPDoc for `integer` and `boolean` are now using the short form, `int` and `bool`
- ~string functions (`substr`, `strlen`, `strpos`, etc.) are now using the more byte-safe `mb_str` variants which is backed by `ext-mbstring` being a required PHP extension~
-  `else if` is changed to `elseif`
- useless `else` and useless variables before `return` are removed
- phpunit assertions are using the more direct equivalents, for example: `assertEquals(true, condition) => assertTrue(condition)`
- file-level docblocks are enforced on system files. This move also ensures that incrementing the date of copyright can be done automatically by just changing the configuration file's default header comment
- consecutive calls to `isset` or `unset` are merged to single calls
- and many more...

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide


michalsn:
Do I understand it correctly and this aims to replace the codeigniter4-standard? 

I haven't checked everything yet, but I have some thoughts after viewing a couple of the first files.

* I'm not a fan of the changes made to the config files - when it comes to alignments. Many times the alignment is broken when there is a comment line between lines.
* I don't like the idea of using multibyte functions everywhere. I don't really see benefits - quite opposite, they're slower. There is nothing in the core of the framework that will need to use multibyte functions. Helper functions designed to be used by users are a different story of course (and maybe router/uri feature).
* I see strange changes being made that change the way the code works: https://github.com/codeigniter4/CodeIgniter4/pull/3776/files#diff-6a1aac03cd08c7e1e25d9f99bae314fb296093c9bb9e70571e79a66c7d9230a6L510

paulbalandan:
> Do I understand it correctly and this aims to replace the codeigniter4-standard?

Yes, because `codeigniter4-standard` is relying on `php_codesniffer`

> I haven't checked everything yet, but I have some thoughts after viewing a couple of the first files.
> 
> * I'm not a fan of the changes made to the config files - when it comes to alignments. Many times the alignment is broken when there is a comment line between lines.

I'll look into this.

> * I don't like the idea of using multibyte functions everywhere. I don't really see benefits - quite opposite, they're slower. There is nothing in the core of the framework that will need to use multibyte functions. Helper functions designed to be used by users are a different story of course (and maybe router/uri feature).

OK, reverted.

> * I see strange changes being made that change the way the code works: https://github.com/codeigniter4/CodeIgniter4/pull/3776/files#diff-6a1aac03cd08c7e1e25d9f99bae314fb296093c9bb9e70571e79a66c7d9230a6L510

That's a manual mistake, already fixed.

samsonasik:
@paulbalandan how about keep using `codeigniter4-standard` but make `codeigniter4-standard` using php-cs-fixer, is that possible? probably make v2.0 for `codeigniter4-standard`?

paulbalandan:
I'll address this properly next time when my free time permits me. I'm closing it for now.

paulbalandan:
> @paulbalandan how about keep using `codeigniter4-standard` but make `codeigniter4-standard` using php-cs-fixer, is that possible? probably make v2.0 for `codeigniter4-standard`?

I like the idea but I don't know if others would be OK with that.

paulbalandan:
What I am doing for now is one-by-one using php-cs-fixer rules and check those which will not conflict with codesniffer's changes.

louisl:
Hello all, just been made aware of this. I don't see why the two cant co-exist as long as the results are the same and they're kept up to date. I did spend a fair amount of time building the `codeigniter4-standard` especially on the fixing aspects formatting arrays and such. I'm aware it's out of date, particularly the installation instructions (things are a little easer now), I built it 3 years ago ;) however I haven't had much feedback or instruction on altering it. 

Although `codeigniter4-standard` was intended to be used for the core - running it through the current `CodeIgniter4` code base it appears it hasn't been used by all contributors. It also can be used and may be in use by people in their own projects and CI pipelines.

PR #3777, opened 2020-10-15
getSize() does not accept parameters like getSizeByUnit()

eSkiSo:

**Description**
getSize() does not accept the unit parameter like getSizeByUnit() as stated on the user guide.
Not sure if that is suposed to be like that or it was a "copy/paste" issue on this document, but does induce into error.

**Checklist:**
- [ ] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


michalsn (approved):
Good catch, thanks!

PR #3782, opened 2020-10-16
Factory parameter options

MGatner:
**Description**
After using `Factories` for a bit I've found some of the functionality to be a bit awkward. The classes that `Factories` replaced had a single option, `getShared`, which I gave undue prominence to in the abstracted class. But with the limit on passable parameters this forced other directives to come from the configuration.

This PR renames factory "configurations" to "options" to help distinguish them from actual `Config` classes. It moves `getShared` to be another option and changes `__callStatic` to take a set of options as the second parameter, which allows full customization by each call.

An example of what this enables... Say an `Authentication` services needs to look up a user via its `UserModel`. By using `Factories`'s `preferApp`, it can allow a developer to provide a different model to use simply by creating `App\Models\UserModel`. However, if the Service wants to enforce certain restrictions it could provide an `instanceOf` directive to make sure it was finding something it could use:
```
$model = Factories::models('UserModel', [
	'preferApp'  => true,
	'instanceOf' => 'Auth\Interfaces\UserModelInterface'
]);
```

Another example, if a developer wants to leverage the shared instance functionality but wants to be sure to get a specific class that can be handled without changing the entire component's options:
```
$widget = Factories::widgets('Org\Widgets\ColorWidget', [
	'getShared' => true,
	'preferApp' => false,
]);
```

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/concepts/factories.rst
+++ b/user_guide_src/source/concepts/factories.rst
@@ -41,43 +41,29 @@ you get back the instance as before::
 		...
 	}
 
-Factories Behavior
+Factory Parameters
 ==================
 
-The default behavior might not work for every component. For example, say your component
-name and its path do not align, or you need limit returns to a certain type of class, or
-your component classes require additional parameters to their constructor. This section
-covers configuring behavior for a specific Factories component.
+``Factories`` takes as a second parameter an array of option values (described below).
+These directives will override the default options configured for each component.
 
-Factory Parameters
-------------------
-
-By default, ``Factories`` assumes that you want to locate a shared instance of a component
-with the provided name. The convenience of ``Factories`` for locating classes is sometimes
-desirable without using its shared instance storage. By adding a second parameter to the
-magic static call, you can control whether ``Factories`` will return a new or shared instance
-each time::
-
-	$users = Factories::models('UserModel', true); // Default; will always be the same instance
-	$other = Factories::models('UserModel', false); // Will always create a new instance
-
-Additionally, any more parameters passed at the same time will be forwarded on to the class
-constructor, making it easy to configure your shared instance on-the-fly. For example, say
+Any more parameters passed at the same time will be forwarded on to the class
+constructor, making it easy to configure your class instance on-the-fly. For example, say
 your app uses a separate database for authentication and you want to be sure that any attempts
 to access user records always go through that connection::
 
 	$conn  = db_connect('AuthDatabase');
-	$users = Factories::models('UserModel', true, $conn);
+	$users = Factories::models('UserModel', [], $conn);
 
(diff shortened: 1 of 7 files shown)


user_guide_src/source/concepts/factories.rst
```diff
@@ -109,11 +107,29 @@ file might look like this::
...
 unrelated "Filters" path in its namespace.
 
-Runtime configuration is even easier: simply supply the desired configuration values to
-``Factories`` using the ``setConfig()`` method and they will be merged with the default
-values and stored for the next call::
+setOptions Method
+-----------------
+
+The ``Factories`` class has a static method to allow runtime option configuration: simply
+supply the desired array of options using the ``setConfig()`` method and they will be
```
samsonasik, line 114: renamed to `setOptions()` ?

user_guide_src/source/concepts/factories.rst
```diff
@@ -109,11 +107,29 @@ file might look like this::
...
+merged with the default values and stored for the next call::
 
 	Factories::setConfig('filters', [
```
samsonasik, line 117: renamed to setOptions() ?

michalsn (approved):
Really good changes. 

Besides the small things in the docs that were already spotted... I wonder if we can allow a `bool|array` as a second parameter, like:
```php
// calling
Factories::config($name, false);
// would be the same as
Factories::config($name, ['getShared' => false]);
```
I guess it would be the most used parameter. I'm not sure about it though.

Overall, this is good to merge.

samsonasik:
@michalsn still need update doc, see my prev comment

michalsn:
> Besides the small things in the docs that were already spotted..

@samsonasik Yes, I know 😉

user_guide_src/source/concepts/factories.rst
MGatner: Good catch! I did a global search-and-replace but it must have applied to `.php` files only.

samsonasik:
@MGatner you forgot to git push before mark comment as resolved?

MGatner:
> I wonder if we can allow a `bool|array` as a second parameter

That was how I started the refactor, but then after working with it I started to believe that giving `getShared` that level of prominence was usually a mistake. `Factories` is most useful for accessing shared instances, because most of the time it will be easier to instantiate a class directly if you don't want its shared version. And half the time when you want a fresh instance you will be specifying other options as well (particularly `instanceOf` and/or `preferApp`). I'm open to pushback, but this is my opinion from writing and using the class so far.

MGatner:
@samsonasik Yes sorry, I was responding to the other review as well before I pushed.

nc03061981:
Thanks CI Teams

PR #3786, opened 2020-10-18
Add Custom PHPStan rule to check Use statement after license docblock

samsonasik:
I created custom PHPStan rule to check Use statement after license docblock like in `system/boostrap.php` which previously use statement is missplaced before license.

**Checklist:**
- [x] Securely signed commits


samsonasik:
All green 🎉

michalsn (approved):

samsonasik:
Thank you @michalsn for the review.

PR #3787, opened 2020-10-18
Add CSV casting

MGatner:
**Description**
Adds support for Entity's `castAs` to use comma-separated values, a common "sub-collection" database storage format.
```
class Widget extends Entity
{
	protected $casts = [
		'colors' => 'csv',
	];
}

$widget->colors = ['red', 'yellow', 'green']; // Stored in the database as "red,yellow,green"
...
$widget = model(WidgetModel::class)->find(1);
var_dump($widget->colors); // ['red', 'yellow', 'green']
```

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/models/entities.rst
+++ b/user_guide_src/source/models/entities.rst
@@ -356,9 +356,9 @@ the value whenever the property is set::
     class User extends Entity
     {
         protected $casts = [
-            'options' => 'array',
-	    'options_object' => 'json',
-	    'options_array' => 'json-array'
+            'options'        => 'array',
+	        'options_object' => 'json',
+	        'options_array'  => 'json-array'
         ];
     }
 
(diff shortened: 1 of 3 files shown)


user_guide_src/source/models/entities.rst
```diff
@@ -370,6 +370,25 @@ the value whenever the property is set::
     $user->options = $options;
     $userModel->save($user);
 
+CSV Casting
+-----------
+
+If you know you have a array of simple values, encoding them as a serialized or JSON string
```
michalsn, line 376:
```suggestion
If you know you have a flat array of simple values, encoding them as a serialized or JSON string
```

michalsn (approved):
I think it would be nice to add a specific keyword to the user guide, but other than that I think it looks great. A very useful feature.

MGatner:
What does that mean, "add a specific keyword"? Is there User Guide metadata or something I am unaware of?

michalsn:
Maybe I put it wrong in words, but I meant "flat array".

nc03061981:
Thanks CI Teams

MGatner:
@michalsn I missed your review comment! Makes much more sense now 😂 Good change, I will add it.

PR #3790, opened 2020-10-20
Add BC Break Policy

samsonasik:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,11 +3,11 @@
 
 ## Contributions
 
-We expect all contributions to conform to our [style guide](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/styleguide.rst), be commented (inside the PHP source files), 
-be documented (in the [user guide](https://codeigniter4.github.io/userguide/)), and unit tested (in the [test folder](https://github.com/codeigniter4/CodeIgniter4/tree/develop/tests)). 
+We expect all contributions to conform to our [style guide](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/styleguide.rst), be commented (inside the PHP source files),
+be documented (in the [user guide](https://codeigniter4.github.io/userguide/)), and unit tested (in the [test folder](https://github.com/codeigniter4/CodeIgniter4/tree/develop/tests)).
 There is a [Contributing to CodeIgniter](./contributing/README.rst) section in the repository which describes the contribution process; this page is an overview.
 
-Note, we expect all code changes or bug-fixes to be accompanied by one or more tests added to our test suite to prove the code works. If pull requests are not accompanied by relevant tests, they will likely be closed. Since we are a team of volunteers, we don't have any more time to work on the framework than you do. Please make it as painless for your contributions to be included as possible. If you need help with getting tests running on your local machines, ask for help on the forums. We would be happy to help out. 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -63,6 +63,11 @@ each other and from whatever else you are doing with your repository!
 
 You must [GPG-sign](./contributing/signing.rst) your work, certifying that you either wrote the work or otherwise have the right to pass it on to an open-source project. This is *not* just a "signed-off-by" commit, but instead, a digitally signed one.
 
+### BC Break Policy
+
+1. New class/property/constant in `system`, that is OK.
+2. New class/property/constant in `app` directory that will be consumed in `system` is a big no without handling in `system` itself if the new class/property/constant not exists.
+
```
MGatner:
```suggestion
### Breaking Changes

In general any changes that would disrupt existing uses of the framework are considered "breaking changes" and are not allowed. A few specific examples to pay attention to:

1. New classes/properties/constants in `system` are acceptable, but anything in the `app` directory that will be used in `system` should be backwards-compatible
2. Any changes to non-private methods must be backwards-compatible with the original definition

```

MGatner (changes requested):
I suggested some more verbose and different wording

CONTRIBUTING.md
samsonasik: implemented.

CONTRIBUTING.md
```diff
@@ -63,6 +63,13 @@ each other and from whatever else you are doing with your repository!
 
 You must [GPG-sign](./contributing/signing.rst) your work, certifying that you either wrote the work or otherwise have the right to pass it on to an open-source project. This is *not* just a "signed-off-by" commit, but instead, a digitally signed one.
 
+### Breaking Changes
+
+In general any changes that would disrupt existing uses of the framework are considered "breaking changes" and are not allowed. A few specific examples to pay attention to:
```
paulbalandan, line 68:
```suggestion
In general, any change that would disrupt existing uses of the framework is considered a "breaking change" and will not be favorably considered. A few specific examples to pay attention to:
```

CONTRIBUTING.md
```diff
@@ -63,6 +63,13 @@ each other and from whatever else you are doing with your repository!
...
+
+1. New classes/properties/constants in `system` are acceptable, but anything in the `app` directory that will be used in `system` should be backwards-compatible.
+2. Any changes to non-private methods must be backwards-compatible with the original definition.
+
```
paulbalandan, line 72: More additions:
3. Deleting non-private properties or methods without prior deprecation notices is frowned upon and will likely be closed.
4. Deleting or renaming public classes and interfaces, as well as those not marked as `@internal`, without prior deprecation notices or not providing fallback solutions will also not be favorably considered.

paulbalandan (changes requested):

CONTRIBUTING.md
samsonasik: applied.

CONTRIBUTING.md
samsonasik: applied.

michalsn (approved):
This is badly needed. Though I doubt that people will read this before sending a PR, at least we'll have a place to send them to for an explanation.

lonnieezell (approved):

PR #3796, opened 2020-10-22
Prevent events duplication on initialisation

paulbalandan:
**Description**
I have always been wondering why on `appstarter` the `pre_system` event is called twice, as shown in the toolbar. Then, this [forum thread](https://forum.codeigniter.com/thread-77765.html) emerged also discussing this *duplication*. It turns out, as pointed by OP, that on events discovery the `Config\Events` file is included twice even though they are referencing the same file. This PR attempts to filter `Events::$files` to have only unique found files to `include`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Events/Events.php
+++ b/system/Events/Events.php
@@ -39,6 +39,7 @@
 
 namespace CodeIgniter\Events;
 
+use Config\Modules;
 use Config\Services;
 
 define('EVENT_PRIORITY_LOW', 200);
@@ -50,7 +51,6 @@ define('EVENT_PRIORITY_HIGH', 10);
  */
 class Events
 {
-
 	/**
 	 * The list of listeners.
 	 *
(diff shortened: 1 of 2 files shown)


MGatner (approved):
I had no idea this was happening! Good catch. I suppose we could also switch to `include_once` to be super sure but I hear that has a slight performance impact.

michalsn (approved):

paulbalandan:
> I had no idea this was happening! Good catch. I suppose we could also switch to `include_once` to be super sure but I hear that has a slight performance impact.

Is a couple of thousand milliseconds a performance hit? I just tried changing `include` to `include_once` and the Bootstrap event in toolbar went up tremendously on first time load. It even maxed the 30 seconds execution time. Relevant times were min of 1000 ms to max of 6000 ms on Windows. But returned to average 30 ms on subsequent loads. I guess it is not worth it as I see it in my pseudo-benchmarking. 🤣

MGatner:
Wow! If that's the case, we should probably change out Autoloader to use `include` and se what performance improvements we get. By definition, it should only be called on to include a file that has not already bene included.

paulbalandan:
I guess so. I've read that the performance time increase was due to the internal tracking of what have already been included.

PR #3799, opened 2020-10-23
Add new Custom Rector Rule to Pass true to 3rd parameter of in_array when no value provided

samsonasik:
Added new Custom Rector Rule to Pass `true` to 3rd parameter of `in_array` when no value provided.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/utils/Rector/PassStrictParameterToFunctionParameterRector.php
+++ b/utils/Rector/PassStrictParameterToFunctionParameterRector.php
@@ -14,13 +14,18 @@ use Rector\Core\RectorDefinition\CodeSample;
 use Rector\Core\RectorDefinition\RectorDefinition;
 
 /**
- * Pass true to 3rd parameter of in_array when no value provided
+ * Pass strict to function parameter on specific position argument when no value provided
  */
-final class PassTrueToThirdParameterInArrayRector extends AbstractRector
+final class PassStrictParameterToFunctionParameterRector extends AbstractRector
 {
+	private const FUNCTION_WITH_ARG_POSITION = [
+		// position start from 0
+		'in_array' => 2,
+	];
+
 	public function getDefinition(): RectorDefinition
 	{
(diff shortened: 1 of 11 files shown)


samsonasik:
All green :tada:

paulbalandan (approved):
Stricter means less prone to errors. 🥳

Can we have also, maybe in a separate PR, other functions which have a `$strict` parameter also set to true, like in `base64_decode` and other functions?

samsonasik:
@paulbalandan Yes, next, I can create new rule for other functions, or refactor it to accept register new function with expected arg position to be true.

utils/Rector/PassStrictParameterToFunctionParameterRector.php
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Pass strict to function parameter on specific position argument when no value provided
+ */
+final class PassStrictParameterToFunctionParameterRector extends AbstractRector
+{
+	private const FUNCTION_WITH_ARG_POSITION = [
+		// position start from 0
+		'in_array' => 2,
+	];
```
samsonasik, line 24: @paulbalandan I've refactored to allow register more functions to specific position to be filled with strict parameter in the above constant.

samsonasik:
latest phpstan release seems cause an error https://github.com/phpstan/phpstan/releases/tag/0.12.51

samsonasik:
I currently ignore new unrelated phpstan error due new release of phpstan

samsonasik:
All green 🎉

utils/Rector/PassStrictParameterToFunctionParameterRector.php
paulbalandan: Thank you for indulging my request that fast.

michalsn (approved):

samsonasik:
@paulbalandan @michalsn Thank you for the review.

PR #3804, opened 2020-10-23
Limit toolbar to initial head tag

MGatner:
**Description**
I have a form to edit email content, the body of which includes a `<head>` tag. The toolbar filter is injecting itself into the form to edit this content because it matches `<head>`. This PR limits the toolbar replacement to the first instance of the head tag.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -395,10 +395,7 @@ class Toolbar
 
 			if (strpos($response->getBody(), '<head>') !== false)
 			{
-				$response->setBody(
-						str_replace('<head>', '<head>' . $script, $response->getBody())
-				);
-
+				$response->setBody(preg_replace('<head>', '<head>' . $script, $response->getBody(), 1));
 				return;
 			}
 


MGatner:
Apparently I don't know how `str_replace` works...

sfadschm:
Does this work I know?
I think the `$count` parameter of `str_replace` is where the number of replacements gets stored after the function (hence it is a reference).
To set a limit to the number of replacements, `preg_replace` might be the way to go.

MGatner:
Oooohhhhhhhhhhhhh. Crap. I will have to redo later, I'm on mobile for ~24 hours. 

Thanks for explaining that, I was so confused why they would pass that by reference... Somebody needs to take away my developer license. 😳

sfadschm:
Happens to the best ;-)

system/Debug/Toolbar.php
```diff
@@ -395,8 +395,9 @@ public function prepare(RequestInterface $request = null, ResponseInterface $res
 
 			if (strpos($response->getBody(), '<head>') !== false)
 			{
+				$count = 1;
```
paulbalandan, line 398: I'm pretty sure this variable is unused. Do we need this somewhere else?

paulbalandan, line 398: I'm pretty sure this variable is unused. Do we need this somewhere else?

MGatner: Ayup. Leftover from my debacle above. I will remove next time I'm on desktop.

paulbalandan (approved):
Weird. Live MetadataTest is failing for Postgre. Rerunning, just to be sure. 😄

michalsn (approved):

samsonasik:
@MGatner this seems cause a `<>` character in template ? Tried latest develop: 

Screen Shot 2020-10-25 at 12 35 10 PM

samsonasik:
@MGatner yes, it seems cause a bug I mentioned in my previuos comment:

Screen Shot 2020-10-25 at 12 35 10 PM

samsonasik:
Viewing source: it got:

```diff
<!-- DEBUG-VIEW START 1 APPPATH/Views/welcome_message.php -->
<!DOCTYPE html>
<html lang="en">
++<<head>
```

with double `<`

samsonasik:
I created PR https://github.com/codeigniter4/CodeIgniter4/pull/3813 for that.

paulbalandan:
I think the problem is with the use of `preg_replace` the `<>` is treated as regex delimiters. I believe the input should be enclosed properly, like `/<head>/`.

samsonasik:
@paulbalandan thank you, should be fixed at #3813

MGatner:
Ah yes, sorry guys. I checked that `<>` weren't special characters but totally forgot about delimiters. 

Aside, do we not have any toolbar injection tests? How did this ever get through?

samsonasik:
It seems not test for that yet https://coveralls.io/builds/34438714/source?filename=system/Debug/Toolbar.php

PR #3805, opened 2020-10-24
Inflector dashizer function additional whitespaces replace

mostafakhudair:
Allow helper dashizer function replaces whitespaces and underscores with dashes

### DIFF
--- a/system/Helpers/inflector_helper.php
+++ b/system/Helpers/inflector_helper.php
@@ -31,7 +31,7 @@
  * @package    CodeIgniter
  * @author     CodeIgniter Dev Team
  * @copyright  2019-2020 CodeIgniter Foundation
- * @license    https://opensource.org/licenses/MIT    MIT License
+ * @license    https://opensource.org/licenses/MIT - MIT License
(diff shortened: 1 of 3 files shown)


michalsn:
Sorry but this could be a breaking change for some people. Probably the same result can be achieved by using `url_title()` helper function.

MGatner:
How about a second parameter that is a list of characters to replace, similar to `trim()`, that defaults to underscore?

michalsn:
That would be an acceptable solution. However, personally I would rather avoid situations where we add parameters to existing functions - especially when it can be handled by the other function (I think it can).

MGatner:
Generally I agree, but since this one is such a parallel to existing string manipulation functions I'm not opposed to it. But yes, this largely just recreates str_replace with array input 😋

paulbalandan:
The desired output of the PR can be achieved by a mix of `dasherize` and `underscore`.
```php
dasherize(underscore('my dog_spot')); // my-dog-spot
```
So leveling up the capability of `dasherize` is too much.

PR #3808, opened 2020-10-24
Fix undefined property BaseConfig::$key

paulbalandan:
**Description**
Before PHPStan v0.12.51, initialisation of `Config\Encryption`'s `$key` property via `BaseConfig` is not marked as error. This was due to, as the author states, PHPStan not really doing the check at all. However, this is now being marked as erroneous.

See: [phpstan/phpstan@3998](https://github.com/phpstan/phpstan/issues/3998)

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/BaseConfig.php
+++ b/system/Config/BaseConfig.php
@@ -39,6 +39,7 @@
 
 namespace CodeIgniter\Config;
 
+use Config\Encryption;
 use Config\Modules;
 use Config\Services;
 use ReflectionClass;
@@ -56,7 +57,6 @@ use RuntimeException;
  */
 class BaseConfig
 {
-
 	/**
 	 * An optional array of classes that will act as Registrars
 	 * for rapidly setting config class properties.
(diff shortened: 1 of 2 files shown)


samsonasik (approved):

MGatner (approved):

samsonasik:
Thank you @paulbalandan

system/Config/BaseConfig.php
```diff
@@ -90,15 +90,15 @@ public function __construct()
...
+		$prefix      = static::class;
 		$slashAt     = strrpos($prefix, '\\');
 		$shortPrefix = strtolower(substr($prefix, $slashAt === false ? 0 : $slashAt + 1));
 
 		foreach ($properties as $property)
 		{
 			$this->initEnvValue($this->$property, $property, $prefix, $shortPrefix);
 
-			if ($shortPrefix === 'encryption' && $property === 'key')
+			if ($this instanceof Encryption && $property === 'key')
```
samsonasik, line 101: actually, parent should not know about the child implementation, but probably that next refactor, like have abstract method that called and overridden in child.

PR #3810, opened 2020-10-25
Use include for autoloading

MGatner:
**Description**
Try using `include` instead of `include_once` during Autoloading, since by definition files should only be autoloader when their class does not yet exist. May grant a performance increase.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -163,7 +163,7 @@ class Autoloader
 				return false;
 			}
 
-			include $config[$class];
+			include_once $config[$class];
 		}, true, // Throw exception
 			true // Prepend
 		);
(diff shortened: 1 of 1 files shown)


paulbalandan:
I checked Composer's own class autoloader and it also uses `include`. So far I think we can solve the errors by adjusting the tests.

MGatner:
I totally agree, I just haven't had a chance to dig into it. If you're looking into it feel free to take over the PR or start a new one.

paulbalandan:
I am looking into refactor Autoloader but I'm afraid I might introduce BC. 😂

paulbalandan:
@MGatner I'll probably close this one first and look into refactoring `Autoloader` once the changes to it in `4.1` got merged.

PR #3811, opened 2020-10-25
Move Encryption class check from BaseConfig to Encryption class

samsonasik:
@paulbalandan this is what I meant at comment https://github.com/codeigniter4/CodeIgniter4/pull/3808#discussion_r511521137 

**Checklist:**
- [x] Securely signed commits


samsonasik:
For note: this may be a bc break if user want to use `config('Encryption')` before first call : `Services::encrypter($config, true)`

samsonasik:
After some thinking, I think it is only can happen in major change to ensure no bc break. Closing.

PR #3821, opened 2020-10-26
Fix doc return type

MGatner:
Missed one.


MGatner:
I'm going to be naughty and just merge this.

PR #3831, opened 2020-10-28
Remove package tags

paulbalandan:
**Description**
This PR removes unnecessary `@package` tags ported from earlier CI versions.

According to the [PHPDoc website](https://docs.phpdoc.org/latest/guide/references/phpdoc/tags/package.html#package):

> The @package tag can be used as a counterpart or supplement to Namespaces. Namespaces provide a functional subdivision of Structural Elements where the @package tag can provide a logical subdivision in which way the elements can be grouped with a different hierarchy.
>
> **If, across the board, both logical and functional subdivisions are equal is it NOT RECOMMENDED to use the @package tag to prevent maintenance overhead.**

In earlier CI versions the package tags were used as in place replacement for namespaces. Since CI4 is actively using namespaces, then usage of package tags is no longer necessary.

This PR also removes other superfluous tags ported from earlier CI versions.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/API/ResponseTrait.php
+++ b/system/API/ResponseTrait.php
@@ -25,8 +25,6 @@ use Config\Services;
  *
  * @property IncomingRequest $request
  * @property Response        $response
- *
- * @package CodeIgniter\API
  */
 trait ResponseTrait
 {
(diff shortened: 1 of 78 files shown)


samsonasik (approved):

samsonasik:
Thank you @paulbalandan

samsonasik:
@paulbalandan there is `@package` left at BaseController in app directory, could you remove that as well? Thank you.

PR #3833, opened 2020-10-29
Feature/entity set cast

mikouaji:
**Description**
I have added a public method `setCast` to the `CodeIgniter\Entity` class.
As I was hashing entity ids with the casting functionality there were problems when I wanted to update such entity.
After some searching I have found out that the `$_cast` variable is messing me up, so I made this method.
There would be no problem but as this var is private I had do change the `CodeIgniter\Entity` class, and this means that my app will break after every framework update.
There is a way around it, and right now I am using it, but its too much for my taste and adding a simple setter to the class is way better.
Current workaround:
```
use CodeIgniter\Entity;
use CodeIgniter\Model;
use ReflectionProperty;

class BaseModel extends Model
{
    public function save($data): bool{
        $entityReflection = new ReflectionProperty(Entity::class, '_cast');
        $entityReflection->setAccessible(true);
        $entityReflection->setValue($data, false);
        $result = parent::save($data);
        $entityReflection->setValue($data, true);
        return $result;
    }
}
``` 
And here we have the code after adding the method to `CodeIgniter\Entity`:
```
use CodeIgniter\Model;

class BaseModel extends Model
{
    public function save($data): bool{
        $data->setCast(false);
        $result = parent::save($data);
        $data->setCast(true);
        return $result;
    }
}
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


paulbalandan (changes requested):
I'm curious. How are you hashing the IDs that casting is hindering the process?

Also, we need tests, one or two tests very simple that shows it can change the behavior of casting. This also ensures the new lines are covered by testing to make code coverage happy.

mikouaji:
I've extended the `Entity::castAs` function and have something like `protected $casts=['id'=>'hash']`, so in `Model::save` (line 549) `$response = $this->update($data->{$this->primaryKey}, $data);` the `id` is cast by the magic getter.

I'll try to write tests next week, I'm not good at those not gonna lie. If you can provide one that does a similar task that would be helpful :)

Commits squashed ;)

paulbalandan:
I don't think if there are existing tests for the casting. The general idea for the test is since you will be changing the behavior of `$_casts` to false, then test is expected that casting of variables will not happen. So, if you have set the `$casts` property to include `'id' => 'string'`, then I would expect that if my original ID is an integer then no changes will happen. It will remain as `int` instead of the casted `string`.

MGatner:
We can't use `setCast()` because this would trigger anytime someone wanted to use their own "cast" attribute, due to the magic `__set()`:
```
$entity->cast = 'Philip';
```

I suggest using simply `cast()`, and make it return the current value of $_cast when no parameter is supplied.

mikouaji:
I've changed the function name to `cast` as suggested, also test were added :)

paulbalandan (approved):

paulbalandan:
@mikouaji thank you for your first contribution :)

PR #3835, opened 2020-10-30
Clarify before filters behavior

### LINKED ISSUE #3579: Bug: When have multi filter class, and "before" function return true，other filters after this filter not working
**CodeIgniter 4 version**
4.0.4

Sometimes I want to use "return true" in filter's "before" function to termination function continues to running. But if a filter use 'return true", and this filter is not the last, the filters after this are not working.

**Here some test code:**
/app/Config/Filters.php
``` 
public $aliases = [
        't1' => \App\Filters\Test1::class,
        't2' => \App\Filters\Test2::class
	];
public $globals = [
	'before' => ['t1', 't2'],
	'after'  => ['toolbar'],
];
```
/app/Filters/Test1.php
```
    public function before(RequestInterface $request, $arguments = null)
    {
        var_dump('test1111');
        return true;
    }
```
/app/Filters/Test2.php
```
    public function before(RequestInterface $request, $arguments = null)
    {
        var_dump('test222');
    }
```

Test1 is working but Test2 not working. I read the code, maybe the reason is here:
/system/Filters/Filters.php :190
```
if (empty($result))
{
	continue;
}

return $result;
```
When "before" return true, "continue" not working.  **so i try to return null or return false in "before" function, after filter is working.**
This may not be a bug, I hope this situation can be explained in the document.

paulbalandan:
**Description**
Fixes #3579 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated

### DIFF
--- a/user_guide_src/source/incoming/filters.rst
+++ b/user_guide_src/source/incoming/filters.rst
@@ -7,7 +7,7 @@ Controller Filters
     :depth: 2
 
 Controller Filters allow you to perform actions either before or after the controllers execute. Unlike :doc:`events </extending/events>`,
-you can very simply choose which URI's in your application have the filters applied to them. Incoming filters may
+you can very simply choose which URIs in your application have the filters applied to them. Incoming filters may
 modify the Request, while after filters can act on and even modify the Response, allowing for a lot of flexibility
 and power. Some common examples of tasks that might be performed with filters are:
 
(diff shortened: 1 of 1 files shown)


user_guide_src/source/incoming/filters.rst
```diff
@@ -7,7 +7,7 @@ Controller Filters
     :depth: 2
 
 Controller Filters allow you to perform actions either before or after the controllers execute. Unlike :doc:`events </extending/events>`,
-you can very simply choose which URI's in your application have the filters applied to them. Incoming filters may
+you can very simply choose which URIs in your application have the filters applied to them. Incoming filters may
```
samsonasik, line 10: "very simply" wording seems need to be removed.

MGatner (approved):

PR #3839, opened 2020-10-30
Apply Rector: CountArrayToEmptyArrayComparisonRector

samsonasik:
Change count array comparison to empty array comparison to improve performance

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -2,6 +2,7 @@
 
 use Rector\Core\Configuration\Option;
 use Rector\CodeQuality\Rector\Return_\SimplifyUselessVariableRector;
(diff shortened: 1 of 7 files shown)


samsonasik:
All green :tada:

samsonasik:
Checked benchmark with 1000000 at https://phpsandbox.io/n/cool-limit-q2mg seems sometime faster, sometime slower, or same. 

Tested locally, compare with `$array !== []` is 2x faster.

paulbalandan (approved):
I've seen this trend in some repo. Is this check also more performant than truthy checks? I mean if `if ($array !== []` is faster than `if ($array)`?

samsonasik:

PR #3846, opened 2020-11-01
Parser: Fixes #3824 (html colors throwing errors)

### LINKED ISSUE #3824: Bug: Parser preg_replace_callback exception if template has #
**Describe the bug**
If the template has for example:
```html
<p style="color:#999999">{test}</p>
``` 
the parser will trow the exception **preg_replace_callback(): Unknown modifier '9'**
*Notice the 9 modifier comes from the first char after #* 
This only happens if you use **setData**, if not the error does not happen.
 
**CodeIgniter 4 version**
4.0.4 (live)

**Affected module(s)**
Parser

- No error:
```php
$template = '<p style="color:red; background-color:#993399;">Price $: 12.50</p>';
echo $parser->renderString($template);
```

- Error
```php
$data   = [
	'books' => [
		['price' => '12.50'],
	],
];
$template = '{books}<p style="color:red; background-color:#993399;">Price $: {price}</p>{/books}';
$parser->setData($data);
echo $parser->renderString($template);
```

**Context**
 - OS: Ubuntu: 18.04.5 LTS
 - Web server: Apache/2.4.29
 - PHP version: PHP 7.4.10

eSkiSo:
Fixes issue where trying to parse an html template that contains colors such as "#fff" or  "#000000" results in **exception preg_replace_callback(): Unknown modifier 'f'**

**Description**
Changed **parse** function to call 2 new functions (at start and end) that handle html color to avoid parsing errors due to the use of **#** 

**Checklist:**
- [ ] Securely signed commits

### DIFF
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -258,7 +258,10 @@ class Parser extends View
 		// Remove any possible PHP tags since we don't support it
 		// and parseConditionals needs it clean anyway...
 		$template = str_replace(['<?', '?>'], ['&lt;?', '?&gt;'], $template);
-
+		
+		// Replace color coding with temporary values
+		$template = $this->parseHtmlColors($template);
+		
 		$template = $this->parseComments($template);
 		$template = $this->extractNoparse($template);
 
@@ -290,6 +293,8 @@ class Parser extends View
 				$template = $this->replaceSingle($pattern, $content, $template, $escape);
 			}
 		}
+		
+		$template = $this->unParseHtmlColors($template);
 
 		return $this->insertNoparse($template);
 	}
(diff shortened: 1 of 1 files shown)


system/View/Parser.php
```diff
@@ -425,6 +430,58 @@ protected function parseComments(string $template): string
...
+	 * Replaces html colors with temporary value so it does not affect parsing regex
+	 * 
+	 * Takes ..color:#fff, background-color:#993399 and color=#0284D0> 
+	 * Returns ..color{{color:fff}}, background-color{{color:993399}}, and color{{color=0284D0}}
+	 * 
+	 * @param string $template
+	 * 
+	 * @return string
+	 */
+	function parseHtmlColors(string $template) : string
```
samsonasik, line 444: please add modifier, you can use private for it: 

```suggestion
	private function parseHtmlColors(string $template): string
```

system/View/Parser.php
```diff
@@ -425,6 +430,58 @@ protected function parseComments(string $template): string
...
+		return $template;
+	}
+
+	/**
+	 * Restores html colors to their original state
+	 *
+	 * @param string $parsedTemplate
+	 * @return string
+	 */
+	function unParseHtmlColors(string $parsedTemplate) : string
```
samsonasik, line 463: please add modifier, you can use private for it: 

```suggestion
	private function unParseHtmlColors(string $parsedTemplate): string
```

samsonasik:
@eSkiSo you can run `composer update && vendor/bin/rector process` to fix the rector notice message. Also, please add unit test. Thank you.

paulbalandan:
We should be providing a long term solution to this. The `#` may not only be used for hex colors but on other instances as well which needs to be addressed too.

eSkiSo:
Closing PR, issue seems resolved on #3848

PR #3848, opened 2020-11-02
Parser: Escape hashes in main pattern

### LINKED ISSUE #3824: Bug: Parser preg_replace_callback exception if template has #

paulbalandan:
**Description**
Fixes #3824 and alternative to #3846 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -17,13 +17,10 @@ use ParseError;
 use Psr\Log\LoggerInterface;
 
 /**
- * Class Parser
- *
- *  ClassFormerlyKnownAsTemplateParser
+ * Class for parsing pseudo-vars
  */
 class Parser extends View
 {
-
 	/**
 	 * Left delimiter character for pseudo vars
 	 *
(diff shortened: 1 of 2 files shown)


system/View/Parser.php
```diff
@@ -572,12 +569,18 @@ public function setDelimiters($leftDelimiter = '{', $rightDelimiter = '}'): Rend
 	 */
 	protected function replaceSingle($pattern, $content, $template, bool $escape = false): string
 	{
-		// Any dollar signs in the pattern will be mis-interpreted, so slash them
+		// Any dollar signs in the pattern will be misinterpreted, so slash them
 		$pattern = addcslashes($pattern, '$');
 
+		// Flesh out the main pattern from the delimiters and escape the hash
+		if (preg_match('/^(#)(.*)(#(m?s)?)$/', $pattern, $parts))
```
samsonasik, line 576: I think it need note with `@see https://regex101.com/path` to show the regex demo, eg: 

```php
// @see https://regex101.com/r/gV1eVk/1
```

so in the future, when there is a bug fix on regex, it will be easier to check/update.

samsonasik (approved):

PR #3853, opened 2020-11-03
Optimize bootstrap

mostafakhudair:


system/bootstrap.php
```diff
@@ -10,8 +10,9 @@
  */
 
 use CodeIgniter\CodeIgniter;
+use CodeIgniter\Config\BaseService;
```
paulbalandan, line 13: Using `BaseService` instead of `Config\Services` deprives users to use their own implementation of `autoloader`.

system/bootstrap.php
```diff
@@ -106,18 +107,9 @@
 
 require_once SYSTEMPATH . 'Autoloader/Autoloader.php';
 require_once SYSTEMPATH . 'Config/BaseService.php';
-require_once SYSTEMPATH . 'Config/Services.php';
-require_once APPPATH . 'Config/Services.php';
```
paulbalandan, line 110: See comment above.

system/bootstrap.php
```diff
@@ -106,18 +107,9 @@
...
-if (! class_exists('CodeIgniter\Services', false))
-{
-	class_alias('Config\Services', 'CodeIgniter\Services');
-}
-
-$loader = Services::autoloader();
-$loader->initialize(new Autoload(), new Modules());
-$loader->register(); // Register the loader with the SPL autoloader stack.
+// Initialize and register the loader with the SPL autoloader stack.
+BaseService::autoloader()->initialize(new Autoload(), new Modules())->register();
```
paulbalandan, line 112: See comment above for `BaseService`.

I don't know for the class aliasing if it is ever used in the wild but it should stay unless there is a negligible chance that users are using the aliased version.

system/bootstrap.php
mostafakhudair: I see that, but i think the framework's Autoloader shouldn't be extended or replaced with custom autoloader.
is there any benefit in using a custom autoloader?

paulbalandan: As the class is not declared "final", it can be extended to change the behavior of some of its methods. Users may have their own peculiar use case for extending the autoloader, like fitting for their specific needs or doing a workaround. For starters, I have extended Autoloader in one of my projects way back to work around a behavior so other users might be doing the same.

mostafakhudair: Thank you, I revert some changes, I think our framework autoloader is strong enough to not be replaced or extended and should be declare as final class, maybe later we shall do this change.

system/bootstrap.php
```diff
@@ -77,7 +78,7 @@
 	require_once APPPATH . 'Config/Constants.php';
 }
 
-// Let's see if an app/Common.php file exists
+// Require app/Common.php file if exists.
 if (file_exists(APPPATH . 'Common.php'))
```
MGatner, line 81: While you're here, I've read that `is_file()` is way faster than `file_exists` - and also technically more correct in this case.

MGatner (changes requested):

paulbalandan (approved):

samsonasik:
@mostafakhudair thank you for the PR. 
@MGatner @paulbalandan thank you for the review.

PR #3856, opened 2020-11-05
Dynamically fill arguments for single service

### LINKED ISSUE #3854: Bug: single_service does not set service's arguments correctly
**Describe the bug**
The global `single_service` method sets the getShared attribute to `false` and then calls the service like this:

    // Ensure it's NOT a shared instance
    array_push($params, false);
    
    return Services::$name(...$params);

This works fine for services which only have the `getShared` argument passed to them (e.g. "timer" which is used both in the docs and in the doc block of `single_service` as an example and which is also used in the CommonFunctionTest). Services accepting multiple arguments however will trigger a type error (e.g., the "pager" service):

> Argument 1 passed to CodeIgniter\Config\Services::pager() must be an instance of Config\Pager or null, bool given, called in C:\XAMPP...

It seems that the `...$params` three dot operator (or whatever you want to call it) does not work as intended here.

Calling the same service with `service()` works, as no arguments at all are passed to `pager()`, such that all arguments fallback to their default values.

I see two solutions here:
1. The documentation should be updated to explain, that all arguments of a service called via `single_service` have to be included in the call, e.g. `single_service('pager', [null, null])`. However, this would make the function itself useless I think, as I can just directly invoke the service then.
2. We need to find a way to insert the `$params` array into the service function argument list "from the back". Honestly, I couldn't find any starting point when looking for something like that. Maybe there is a way to determine the number of arguments a function expects, so we could pad the `$params` array with the corresponding number of 'null' values (as all service function arguments seem to default to 'null' as far as I see).

Any ideas on this?

**CodeIgniter 4 version**
latest develope

**Affected module(s)**
services / common methods

The getShared attribute should be unpacked to the last attribute of the requested service function.

**Context**
 - Windows 10
 - Apache 2.4
 - PHP 7.4

paulbalandan:
**Description**
Fixes #3854 .

This PR deprecates the protected `BaseService::discoverServices()` method and splitting its services discovery function into a new protected `buildServicesCache()`. The resolution of where the service might be defined is now defined in a new public static method `serviceExists()` that takes the name of service as lone argument. This returns an array of `bool $result, ?string $service`.

For the `single_service` function, this takes use of `ReflectionMethod` and `ReflectionParameter` to introspect the signature of the method requested and fills up the `$params` by the default values but also ensures that the last argument is always `false` to ensure no shared instance.

@sfadschm, please stress test this solution to find any possible regressions.

So far, the tests checks the function when:
- the service is not defined
- the params are not provided
- at least one param is given
- all possible params are provided, with the last set to `true`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -1036,7 +1036,6 @@ if (! function_exists('service'))
 if (! function_exists('single_service'))
 {
 	/**
-	 * Allow cleaner access to a Service.
 	 * Always returns a new instance of the class.
 	 *
 	 * @param string     $name
@@ -1046,10 +1045,36 @@ if (! function_exists('single_service'))
 	 */
 	function single_service(string $name, ...$params)
 	{
-		// Ensure it's NOT a shared instance
-		array_push($params, false);
+		$service = Services::serviceExists($name);
 
-		return Services::$name(...$params);
+		if ($service === null)
+		{
+			// The service is not defined anywhere so just return.
+			return null;
+		}
+
+		$method = new ReflectionMethod($service, $name);
+		$count  = $method->getNumberOfParameters();
+		$mParam = $method->getParameters();
+		$params = $params ?? [];
+
+		if ($count === 1)
+		{
+			// This service needs only one argument, which is the shared
+			// instance flag, so let's wrap up and pass false here.
+			return $service::$name(false);
+		}
+
+		// Fill in the params with the defaults, but stop before the last
+		for ($startIndex = count($params); $startIndex <= $count - 2; $startIndex++)
+		{
+			$params[$startIndex] = $mParam[$startIndex]->getDefaultValue();
+		}
+
+		// Ensure the last argument will not create a shared instance
+		$params[$count - 1] = false;
+
+		return $service::$name(...$params);
 	}
 }
 
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -65,7 +65,12 @@ class BaseService
 	 */
 	protected static $services = [];
 
-	//--------------------------------------------------------------------
+	/**
+	 * A cache of the names of services classes found.
+	 *
+	 * @var array<string>
+	 */
+	private static $serviceNames = [];
 
 	/**
 	 * Returns a shared instance of any of the class' services.
@@ -98,8 +103,6 @@ class BaseService
 		return static::$instances[$key];
 	}
 
-	//--------------------------------------------------------------------
-
 	/**
 	 * The Autoloader class is the central class that handles our
 	 * spl_autoload_register method, and helper methods.
@@ -122,8 +125,6 @@ class BaseService
 
 		return new Autoloader();
 	}
-
-	//--------------------------------------------------------------------
 
 	/**
 	 * The file locator provides utility methods for looking for non-classes
@@ -149,8 +150,6 @@ class BaseService
 		return new FileLocator(static::autoloader());
 	}
 
-	//--------------------------------------------------------------------
-
 	/**
 	 * Provides the ability to perform case-insensitive calling of service
 	 * names.
@@ -162,17 +161,40 @@ class BaseService
 	 */
 	public static function __callStatic(string $name, array $arguments)
 	{
-		$name = strtolower($name);
+		$service = static::serviceExists($name);
 
-		if (method_exists(Services::class, $name))
+		if ($service === null)
 		{
-			return Services::$name(...$arguments);
+			return null;
 		}
 
-		return static::discoverServices($name, $arguments);
+		return $service::$name(...$arguments);
 	}
 
-	//--------------------------------------------------------------------
+	/**
+	 * Check if the requested service is defined and return the declaring
+	 * class. Return null if not found.
+	 *
+	 * @param string $name
+	 *
+	 * @return string|null
+	 */
+	public static function serviceExists(string $name): ?string
+	{
+		static::buildServicesCache();
+		$services = array_merge([Services::class], self::$serviceNames);
+		$name     = strtolower($name);
+
+		foreach ($services as $service)
+		{
+			if (method_exists($service, $name))
+			{
+				return $service;
+			}
+		}
+
+		return null;
+	}
 
 	/**
 	 * Reset shared instances and mocks for testing.
@@ -181,8 +203,7 @@ class BaseService
 	 */
 	public static function reset(bool $initAutoloader = false)
 	{
-		static::$mocks = [];
-
+		static::$mocks     = [];
 		static::$instances = [];
 
 		if ($initAutoloader)
@@ -190,8 +211,6 @@ class BaseService
 			static::autoloader()->initialize(new Autoload(), new Modules());
 		}
 	}
-
-	//--------------------------------------------------------------------
 
 	/**
 	 * Inject mock object for testing.
@@ -204,8 +223,6 @@ class BaseService
 		static::$mocks[strtolower($name)] = $mock;
 	}
 
-	//--------------------------------------------------------------------
-
 	/**
 	 * Will scan all psr4 namespaces registered with system to look
 	 * for new Config\Services files. Caches a copy of each one, then
@@ -216,6 +233,10 @@ class BaseService
 	 * @param array  $arguments
 	 *
 	 * @return mixed
+	 *
+	 * @deprecated
+	 *
+	 * @codeCoverageIgnore
 	 */
 	protected static function discoverServices(string $name, array $arguments)
 	{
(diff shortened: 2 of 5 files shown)


sfadschm:
Thanks! First of all, this solves the specific issue from the bug report.

So far, I have tested by invoking any of the available services with every possible combination of arguments.
The following services failed:

`clirequest`
>   ErrorException
>   array_shift() expects parameter 1 to be array, null given
>   SYSTEMPATH\HTTP\CLIRequest.php at line 213
This likely happens because the command is not executed from CLI I guess. If so, maybe we should setup protection against invoking the service like I tried and throw a corresponding exception? 
  
`encrypter`
>  throws `CodeIgniter\Encryption\Exceptions\EncryptionException`
This is expected I think, as with the default value 'null' no Config is passed and hence no passphrase is set.
I wonder if it is really useful to have the null fallback for $config (however, this cannot be changed now).

  
`session`
> ini_set(): A session is active. You cannot change the session module's ini settings at this time 
> SYSTEMPATH\Session\Handlers\FileHandler.php at line 110

I actually don't know where this is comming from. Will take a look later.

All in all, I think the PR is fine, errors are likely not related to these changes or not even to the `single_service `method at all.
Personally, I also like the new implementation with `serviceExists` :)

system/Common.php
```diff
@@ -1046,10 +1045,45 @@ function service(string $name, ...$params)
...
+		$finalCount = count($params);
+
+		if ($finalCount === $count)
+		{
+			$params[$finalCount - 1] = false;
+		}
+		else
+		{
+			$params[] = false;
+		}
```
sfadschm, line 1084: Do we really need this check? 
`$count` already holds the total number of method parameters so I think we can just `$params[$Count - 1] = false;`.

paulbalandan: The `$finalCount` holds the count of `$params` after it has been filled by ReflectionParameter. My original code for that was simply push in the last value as `false`, but realized what if a user defined **all** the params including the last one? This check will ensure that if the last param was passed as true or any other else, it will be turned to false for no shared instance.

paulbalandan:
> `clirequest`
> 
> > ErrorException
> > array_shift() expects parameter 1 to be array, null given
> > SYSTEMPATH\HTTP\CLIRequest.php at line 213
> > This likely happens because the command is not executed from CLI I guess. If so, maybe we should setup protection against invoking the service like I tried and throw a corresponding exception?

I don't encounter this problem. Did you call it from the browser?

> `encrypter`
> 
> > throws `CodeIgniter\Encryption\Exceptions\EncryptionException`
> > This is expected I think, as with the default value 'null' no Config is passed and hence no passphrase is set.
> > I wonder if it is really useful to have the null fallback for $config (however, this cannot be changed now).
> 
> `session`
> 
> > ini_set(): A session is active. You cannot change the session module's ini settings at this time
> > SYSTEMPATH\Session\Handlers\FileHandler.php at line 110
> 
> This is failing because we try to initiate a second session instance here (with file storage in this case). Maybe invoking a second session object should be forbidden or before ini_set it should be checked whether this is still possible.

The encrypter is failing because the unaltered `Config\Encryption` has no defined starter key. Since this will be passed to the Encryption library because `null` was passed, we cannot do anything for the test.

So far, I excluded these two from the provided services because of the errors you just mentioned. For the session, I think it's beyond this PR to check.

system/Common.php
sfadschm: But even if I pass in all arguments including `getShared` as true , then `$params[$count - 1] = false;` will still replace this value or am I overseeing something?
Lets take `single_service('exceptions', null, null, null, true);`:
`count($params)` will be `4`
`$count` will be `4` 
Hence we can directly set `$params[3]` to `false`.

sfadschm:
> So far, I excluded these two from the provided services because of the errors you just mentioned. For the session, I think it's beyond this PR to check.
I agree, this should be a new issue.

> I don't encounter this problem. Did you call it from the browser?
Yes this was called via browser. There, `getServer('argv')` seems to be null.
Does it work in the browser for you?

system/Common.php
paulbalandan: Consider this example:

`single_service('curlrequest');`
CurlRequest has 4 defined parameters.

The `$params` count after filling up is 3 which is `[[], null, null]`. Then the last param for shared instance will be pushed.

`single_service('curlrequest', [], null, null, true);`

The `$params` count will be 4 and there would be no filling up since the user has defined the params. But we need to change the last param to false to ensure no shared instance. So `$params[4 - 1] = false;`

sfadschm: Yes, but still `$count  = $method->getNumberOfParameters();` in both cases holds the value `4` and can directly be used I think.

paulbalandan:
> > I don't encounter this problem. Did you call it from the browser?
> 
> Yes this was called via browser. There, `getServer('argv')` seems to be null.
> Does it work in the browser for you?

Just checked `CLIRequest`'s code. `getServer('argv')` is actually calling `$_SERVER['argv']` which is only set for CLI. Since you are on the browser, getServer will just return `null` and `array_shift` will complain. After all, CLIRequest is meant to be called from the CLI so it should not pose a problem.

sfadschm:
> > > I don't encounter this problem. Did you call it from the browser?
> > 
> > 
> > Yes this was called via browser. There, `getServer('argv')` seems to be null.
> > Does it work in the browser for you?
> 
> Just checked `CLIRequest`'s code. `getServer('argv')` is actually calling `$_SERVER['argv']` which is only set for CLI. Since you are on the browser, getServer will just return `null` and `array_shift` will complain. After all, CLIRequest is meant to be called from the CLI so it should not pose a problem.

That's what I was trying to say :-)
Just thought it might be useful to prohibit calling such CLI only services from the browser or at least throw a meaningful error message.

system/Common.php
paulbalandan: OK. I get what you meant. The extra variable can be dispensed. Thanks.

paulbalandan:
> > > > I don't encounter this problem. Did you call it from the browser?
> > > 
> > > 
> > > Yes this was called via browser. There, `getServer('argv')` seems to be null.
> > > Does it work in the browser for you?
> > 
> > 
> > Just checked `CLIRequest`'s code. `getServer('argv')` is actually calling `$_SERVER['argv']` which is only set for CLI. Since you are on the browser, getServer will just return `null` and `array_shift` will complain. After all, CLIRequest is meant to be called from the CLI so it should not pose a problem.
> 
> That's what I was trying to say :-)
> Just thought it might be useful to prohibit calling such CLI only services from the browser or at least throw a meaningful error message.

Oh. That deserves another PR. :)

system/Common.php
sfadschm: Thats still not what I meant :D

After the for loop ( line 1069)
  `for ($startIndex = count($params); $startIndex <= $count - 2; $startIndex++)`
there is only two possible outcomes. Either, `$params` now hat the size `$count` (i.e., if the user also passed in the `$getShared` argument) or  `$params` now has the size `$count - 1` (because the last loop value for `$startIndex` is `$count - 2`).

In both cases, we can just do `$params[$count - 1] = false;` without further check.
In the first case, this will overwrite the `$getShared` value passed by the user. In the second case, this will create the index for `$getShared` and set it to false.

Both have the same outcome.

So, suggestion is: Remove lines 1075-1076 and 1078-1082.

I hope I was more precise this time :D

paulbalandan: Ok. That's clever. I didn't notice that. Changing.

sfadschm: One more thing: 
In the current implementation, if someone mistakenly passes too many arguments (lets say 5 arguments for 'curlrequest), then on line 1075, `count($params)` will be 5 and `$count` will be 4, so the else clause will be called and create a sixth argument.
This is also solved by the suggestion above.
Btw.: As we are already using reflection here, we might want to check if the user overloads the service, like check if `count($params)` is bigger than `$count` at the beginning and throw an exception if so.

paulbalandan: I was thinking earlier of how to check for many args passed and that surely solves it.

I don't know for the exception for param overloading as php itself is not throwing an error. But for best practice maybe we can do that. I'll take a second look on it tomorrow.

MGatner:
I will have to look through this more on desktop, but initial question: what does returning `[bool, string]` for the check offer over simply `?string`?

paulbalandan:
> I will have to look through this more on desktop, but initial question: what does returning `[bool, string]` for the check offer over simply `?string`?

Good catch. The method is originally named `isServiceExists` and so expected to return bool but then needed to get the matched service. Checking against null would be a much simpler alternative I think.

paulbalandan:
@MGatner do you have time for a review on this?

MGatner:
Yes, looking now. I went through it last tome so just a glance

MGatner (approved):
Very nice. I wish I'd had this a couple weeks ago 😭

MGatner:
This caused PHPStan to fail - probably due to a newer version. We will need to fix that and probably lock in a version.

paulbalandan:
> This caused PHPStan to fail - probably due to a newer version. We will need to fix that and probably lock in a version.

I'll look into this.

PR #3857, opened 2020-11-05
Fix latest phpstan error on Forge

paulbalandan:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -331,7 +331,7 @@ class Forge
 	 */
 	public function addKey($key, bool $primary = false, bool $unique = false)
 	{
-		if ($primary === true)
+		if ($primary)
 		{
 			foreach ((array) $key as $one)
 			{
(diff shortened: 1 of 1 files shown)


system/Database/Forge.php
```diff
@@ -341,9 +341,10 @@ public function addKey($key, bool $primary = false, bool $unique = false)
 		else
 		{
 			$this->keys[] = $key;
-			if ($unique === true)
+
+			if ($unique)
 			{
-				$this->uniqueKeys[] = ($c = count($this->keys)) ? $c - 1 : 0;
+				$this->uniqueKeys[] = count($this->keys) - 1;
```
sfadschm, line 347: This change seems to be right according to the phpstan interpretation.
However, looking at the old implementation, I think something like this was intended by the original contributor:
`$this->uniqueKeys[] = count($this->keys) ? (count($this->keys) - 1) : 0;`
Can't think of a shorter syntax right now.
Maybe also stricter checks should be added here.

paulbalandan: But if you will look at the line before that, `$this->keys` is pushed with the value of `$key`, so the count will always return true and the falsy condition will never be reached, just as phpstan says.

sfadschm: You're right, didn't think about that.

sfadschm (approved):

samsonasik (approved):

samsonasik:
@paulbalandan thank you.

PR #3858, opened 2020-11-05
Expressly forbid CLIRequest from running in non-CLI env

paulbalandan:
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CLIRequest.php
+++ b/system/HTTP/CLIRequest.php
@@ -12,6 +12,7 @@
 namespace CodeIgniter\HTTP;
 
 use Config\App;
+use RuntimeException;
 
 /**
  * Class CLIRequest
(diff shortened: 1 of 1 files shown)


sfadschm (approved):

MGatner:
Should this throw system/CLI/Exceptions/CLIException.php ?

sfadschm:
I would think that the CLIExceptions should rather be used when a user executes a CLI command in a wrong manner.
In the case here, a user tries to call a CLI-only service from browser (or not-CLI). Thats a different situation if you ask me.
However, I'm not sure if this distinction is a little picky...

Maybe this would make sense in the ConfigException, as services are located in Config?

paulbalandan:
IMO `CLIException` should be thrown when it is running already in CLI and something went wrong. But here it is attempting to run in a non-CLI. I opted for the more generic `RuntimeException` as it is the safest. If I were to use a framework specific exception, then I would choose one under the HTTP namespace but I don't like creating a new static method just for that.

MGatner (approved):

samsonasik:
Thank you @paulbalandan

PR #3860, opened 2020-11-06
DX: Normalise whitespaces

paulbalandan:
**Description**
Normalizes whitespaces by removing unnecessary ones and adding where it would be needed.

EDIT: Some of rules applied by php-cs-fixer are reversed by php-codesniffer, and I cannot do anything as it is in the pre-commit guard.
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Views/errors/html/error_exception.php
+++ b/app/Views/errors/html/error_exception.php
@@ -64,12 +64,12 @@
 							<!-- Trace info -->
 							<?php if (isset($row['file']) && is_file($row['file'])) :?>
 								<?php
-								if (isset($row['function']) && in_array($row['function'], ['include', 'include_once', 'require', 'require_once']))
-									{
+								if (isset($row['function']) && in_array($row['function'], ['include', 'include_once', 'require', 'require_once'], true))
+								{
 									echo $row['function'] . ' ' . static::cleanPath($row['file']);
 								}
 								else
-									{
+								{
 									echo static::cleanPath($row['file']) . ' : ' . $row['line'];
 								}
 								?>
@@ -89,9 +89,9 @@
 										<?php
 										$params = null;
 										// Reflection by name is not available for closure function
-										if (substr( $row['function'], -1 ) !== '}')
+										if (substr($row['function'], -1) !== '}')
 										{
-											$mirror = isset( $row['class'] ) ? new \ReflectionMethod( $row['class'], $row['function'] ) : new \ReflectionFunction( $row['function'] );
+											$mirror = isset($row['class']) ? new \ReflectionMethod($row['class'], $row['function']) : new \ReflectionFunction($row['function']);
 											$params = $mirror->getParameters();
 										}
 										foreach ($row['args'] as $key => $value) : ?>
@@ -114,7 +114,7 @@
 						</p>
 
 						<!-- Source? -->
-						<?php if (isset($row['file']) && is_file($row['file']) &&  isset($row['class'])) : ?>
+						<?php if (isset($row['file']) && is_file($row['file']) && isset($row['class'])) : ?>
 							<div class="source">
 								<?= static::highlightFile($row['file'], $row['line']) ?>
 							</div>
@@ -354,7 +354,7 @@
 
 				<ol>
 				<?php foreach ($files as $file) :?>
-					<li><?= htmlspecialchars( static::cleanPath($file), ENT_SUBSTITUTE, 'UTF-8') ?></li>
+					<li><?= htmlspecialchars(static::cleanPath($file), ENT_SUBSTITUTE, 'UTF-8') ?></li>
 				<?php endforeach ?>
 				</ol>
 			</div>
--- a/system/View/Cell.php
+++ b/system/View/Cell.php
@@ -44,7 +44,6 @@ use ReflectionMethod;
  */
 class Cell
 {
-
 	/**
 	 * Instance of the current Cache Instance
 	 *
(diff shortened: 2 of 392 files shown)


system/View/Cell.php
```diff
@@ -120,10 +119,8 @@ public function render(string $library, $params = null, int $ttl = 0, string $ca
 
 			$output = $instance->{$method}();
 		}
-		elseif (($paramCount === 1) && (
-				(! array_key_exists($refParams[0]->name, $paramArray)) ||
-				(array_key_exists($refParams[0]->name, $paramArray) && count($paramArray) !== 1))
-		)
+		elseif (($paramCount === 1)
+			&& ((! array_key_exists($refParams[0]->name, $paramArray)) || (array_key_exists($refParams[0]->name, $paramArray) && count($paramArray) !== 1)))
```
MGatner, line 123: This should either be three lines or one line.

tests/_support/Language/en/Core.php
```diff
@@ -1,17 +1,4 @@
...
- *
- * @package    CodeIgniter
- * @author     CodeIgniter Dev Team
- * @copyright  2014-2018 British Columbia Institute of Technology (https://bcit.ca/)
- * @license    https://opensource.org/licenses/MIT	MIT License
- * @link       https://codeigniter.com
- * @since      Version 4.0.0
- * @filesource
- *
- * @codeCoverageIgnore
```
MGatner, line 13: Was this intentionally removed?

MGatner (changes requested):
Mostly good changes here. I made a few specific notes, and a list of the files I think were better before. Generally I don't like adding newlines between a bunch of single-line properties - I think that covers a lot of these, so if there is a rule we could tweak around that it might save some time.

I don't like the changes to:
* Constants
* ContentSecurityPolicy
* Logger
* ResponseInterface
* MockAppConfig
* MockCLIConfig
* bootstrap.php
* all the tests/_support/Commands headers
* tests/EventModel
* tests/CommandRunnerTest
* all the tests/Cache/* headers
* tests/SImpleConfig
* tests/ModelTest "new class extends Entity"
* tests/HoneypotTest
* tests/CookieHelperTest

I need to look at Email and the large diffs at the end separately.

paulbalandan:
> Mostly good changes here. I made a few specific notes, and a list of the files I think were better before. Generally I don't like adding newlines between a bunch of single-line properties - I think that covers a lot of these, so if there is a rule we could tweak around that it might save some time.
> 
> I don't like the changes to:
> 
> * Constants
Added rule '||' to 'align_single_space_minimal'

> * ContentSecurityPolicy
My intention for enabling class_attributes_separation is to properly put docblocks there afterwards

> * Logger
Reverted

> * ResponseInterface
Reverted

> * MockAppConfig
Reverted

> * MockCLIConfig
Reverted

> * bootstrap.php
Added rule '||' to 'align_single_space_minimal'

> * all the tests/_support/Commands headers
This is consistent with PSR12

> * tests/EventModel
Reverted

> * tests/CommandRunnerTest
Reverted

> * all the tests/Cache/* headers
The headers are consistent with PSR12

> * tests/SImpleConfig
Reverted

> * tests/ModelTest "new class extends Entity"
class_attribute_separation is now disabled

> * tests/HoneypotTest
Reverted

> * tests/CookieHelperTest
Reverted

> I need to look at Email and the large diffs at the end separately.

app/Views/errors/html/error_exception.php
```diff
@@ -354,7 +354,7 @@
 
 				<ol>
 				<?php foreach ($files as $file) :?>
-					<li><?= htmlspecialchars( static::cleanPath($file), ENT_SUBSTITUTE, 'UTF-8') ?></li>
+					<li><?= htmlspecialchars(clean_path($file), ENT_SUBSTITUTE, 'UTF-8') ?></li>
```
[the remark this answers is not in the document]
paulbalandan: Ah yes. To be removed.

app/Views/errors/html/error_exception.php
```diff
@@ -389,9 +389,9 @@
 		<div class="container">
 
 			<p>
-				Displayed at <?= date('H:i:sa') ?> &mdash;
-				PHP: <?= phpversion() ?>  &mdash;
-				CodeIgniter: <?= \CodeIgniter\CodeIgniter::CI_VERSION ?>
+				Displayed at <?= date('H:i:sa'); ?> &mdash;
+				PHP: <?= PHP_VERSION; ?>  &mdash;
+				CodeIgniter: <?= \CodeIgniter\CodeIgniter::CI_VERSION; ?>
```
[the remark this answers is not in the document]
paulbalandan: Same. Three rules applied here.

tests/_support/Language/en/Core.php
paulbalandan: Ah yes. This was a stray header.

MGatner (changes requested):
This one is pretty painful to review, it might help to break it up by subfolders or something.

Still some of the split properties I don't like, e.g.:
* tests/_support/Commands/ParamsReveal.php
* tests/system/Commands/GenerateKeyTest.php
* tests/system/Config/fixtures/Encryption.php 
* tests/system/Database/Migrations/MigrationRunnerTest.php

That last one (Migration Runner Test) has some weird spaces in the object array cast too.

paulbalandan:
> This one is pretty painful to review, it might help to break it up by subfolders or something.

Will do.

PR #3863, opened 2020-11-06
Add MockFilters for feature testing

MGatner:
**Description**
Filters can return a `ResponseInterface` directly, which bypasses feature testing's buffer handling and causes CLI output. This PR adds a new `MockFilters` which intercepts responses and starts a new buffer predictively to prevent output (the buffer will be closed shortly after by `FeatureTestTrait`.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -17,6 +17,7 @@ use CodeIgniter\HTTP\Request;
 use CodeIgniter\HTTP\URI;
 use CodeIgniter\HTTP\UserAgent;
 use CodeIgniter\Router\Exceptions\RedirectException;
+use CodeIgniter\Test\Mock\MockFilters;
 use Config\App;
 use Config\Services;
 use Exception;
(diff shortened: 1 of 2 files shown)


MGatner:
Probably finding a way to incorporate `MockResponse` into feature tests would be better but our whole `FeatureTestTrait` revolves around gathering buffered output, so mark it as "to do" for somewhere down the road...

paulbalandan (commented):
We need a few tests for this.

Quick question. Since this is preventing output to be printed in the CLI, does the `CITestStreamFilter` can't do the same? In the spark commands tests we use that class to capture output printed in the CLI so I am thinking maybe it can do the same here.

MGatner:
That's a good question: I'm not sure. The problem isn't so much that there is output, but rathe that feature tests may have endpoints which use Filters that return a Response directly. In CodeIgniter.php if this is detected it will send the response immediately, causing it to be written out.

I think we would still need some way to interrupt that process, since plenty of feature tests write output, just not directly. But I will look into it more.

MGatner:
Actually digging through the "why" right now I think that the filter response in CodeIgniter.PHP should be amended to respect `$returnResponse`.

MGatner:
Superseded by a PR with a more appropriate solution.

PR #3866, opened 2020-11-08
Fabricator failure

MGatner:
**Description**
Fixes a bug in `Fabricator` where `create()` could fail to insert anything causing `reset($ids)` to be null and `model->find()` would then return all rows. This PR has Fabricator throw a helpfully-descriptive error message when the create call fails.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/Fabricator.php
+++ b/system/Test/Fabricator.php
@@ -551,9 +551,9 @@ class Fabricator
 			else
 			{
 				throw new FrameworkException(lang('Fabricator.createFailed', [
-													  $this->model->table,
-													  implode(' ', $this->model->errors()),
-												  ]));
+					$this->model->table,
+					implode(' ', $this->model->errors()),
+				]));
 			}
 		}
 
--- a/system/Test/Fabricator.php
+++ b/system/Test/Fabricator.php
@@ -547,14 +547,10 @@ class Fabricator
 			{
 				$ids[] = $id;
 				self::upCount($this->model->table);
+				continue;
 			}
-			else
-			{
-				throw new FrameworkException(lang('Fabricator.createFailed', [
-					$this->model->table,
-					implode(' ', $this->model->errors()),
-				]));
-			}
+
+			throw FrameworkException::forFabricatorCreateFailed($this->model->table, implode(' ', $this->model->errors()));
 		}
 
 		// If the model defines a "withDeleted" method for handling soft deletes then use it
--- a/system/Test/Fabricator.php
+++ b/system/Test/Fabricator.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter\Test;
 
+use CodeIgniter\Exceptions\FrameworkException;
 use CodeIgniter\Model;
 use Faker\Factory;
 use Faker\Generator;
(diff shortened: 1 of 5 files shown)


system/Test/Fabricator.php
```diff
@@ -545,6 +548,13 @@ public function create(int $count = null, bool $mock = false)
 				$ids[] = $id;
 				self::upCount($this->model->table);
 			}
```
samsonasik, line 550: probably bettter to use `continue` statement in previous `if`, and remove `else`

paulbalandan: Yes, the `continue` would be better here.

system/Test/Fabricator.php
```diff
@@ -545,6 +548,13 @@ public function create(int $count = null, bool $mock = false)
...
+			else
+			{
+				throw new FrameworkException(lang('Fabricator.createFailed', [
+					$this->model->table,
+					implode(' ', $this->model->errors()),
+				]));
```
paulbalandan, line 556: The formatting issue is one of the reasons I don't like codesniffer as the CS here.

On another note, the current pattern for framework-specific exceptions is to use public static methods and then use the localization there. I am not aware of the reasons behind but I think it should be the same here. Or perhaps this is the start and let's use the `new` keyword from now on?

system/Test/Fabricator.php
MGatner: Will do!

system/Test/Fabricator.php
samsonasik: @paulbalandan I created PHPStan rule for that at PR https://github.com/codeigniter4/CodeIgniter4/pull/3868

MGatner: I didn't do this initially because I didn't want to add more test code to the production section, but I think it still makes sense anyway and helps clean up that throw. Changed.

utils/PHPStan/CheckUseStatementsAfterLicenseRule.php
```diff
@@ -35,7 +35,7 @@ public function processNode(Node $node, Scope $scope): array
 	{
 		$comments = $node->getAttribute('comments');
 
-		if ($comments === [])
+		if ($comments === [] || ! is_array($comments))
```
samsonasik, line 38: `$node->getAttribute('comments')` will always returns `array`, if PHPStan complain about it, we can add `@var` doc before assignment:

```php
                /** @var array $comments */
 		$comments = $node->getAttribute('comments');
```

MGatner: I didn't mean to commit that, but that's a local fix I had to make because the PHPStan pre-commit hook would fail with "Invalid use of foreach on non-array". I was planning to add it at some point so maybe now is the time. I'm not sure what is causing the issue higher but it is an actual type issue, not a static analysis issue.

samsonasik: Ok, you can change above assignment to:

```php
$comments = $node->getComments();
```

Above will always returns array. Ref : 

https://github.com/nikic/PHP-Parser/blob/1d1bc8a36459d67acef508b952dad097960b110e/lib/PhpParser/NodeAbstract.php#L104-L106

that `getComments()` has empty `[]` fallback.

MGatner: Yep! That worked. Incoming.

samsonasik (approved):

MGatner:
Latest changes do not affect unit tests; merging.

PR #3868, opened 2020-11-08
[PHPStan] Add PHPStan rule to check FrameworkException Instantiation Via Static Call

samsonasik:
Ref https://github.com/codeigniter4/CodeIgniter4/pull/3866/files#r519355043 /cc @paulbalandan 

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/utils/PHPStan/CheckFrameworkExceptionInstantiationViaStaticCallRule.php
+++ b/utils/PHPStan/CheckFrameworkExceptionInstantiationViaStaticCallRule.php
@@ -34,8 +34,7 @@ final class CheckFrameworkExceptionInstantiationViaStaticCallRule implements Rul
 			return [];
 		}
 
-		$className = (string) $class;
-		if (! is_a($className, FrameworkException::class, true))
+		if (! is_a((string) $class, FrameworkException::class, true))
 		{
 			return [];
 		}
--- a/utils/PHPStan/CheckFrameworkExceptionInstantiationViaStaticCallRule.php
+++ b/utils/PHPStan/CheckFrameworkExceptionInstantiationViaStaticCallRule.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Utils\PHPStan;
+
+use CodeIgniter\Exceptions\FrameworkException;
+use PhpParser\Node;
+use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Name\FullyQualified;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+
+final class CheckFrameworkExceptionInstantiationViaStaticCallRule implements Rule
+{
+	/**
+	 * @var string
+	 */
+	private const ERROR_MESSAGE = 'FrameworkException instance creation via new expression is not allowed, use static call instead';
+
+	public function getNodeType(): string
+	{
+		return New_::class;
+	}
+
+	/**
+	 * @param New_ $node
+	 */
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$class = $node->class;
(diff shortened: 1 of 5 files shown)


samsonasik:
All green 🎉

utils/PHPStan/CheckFrameworkExceptionInstantiationViaStaticCallRule.php
```diff
@@ -0,0 +1,45 @@
...
+	public function processNode(Node $node, Scope $scope): array
+	{
+		$class = $node->class;
+		if (! $class instanceof FullyQualified)
+		{
+			return [];
+		}
+
+		$className = (string) $class;
+		if (! is_a($className, FrameworkException::class, true))
```
paulbalandan, line 38:
```suggestion
		if (! is_a((string) $class, FrameworkException::class, true))
```

My concern is not all exceptions extends `FrameworkException` but also use the static call for throwing, like in Cache and CLI. I think these should be covered by this rule as well. I think we can make all CI4 defined exceptions implement `CodeIgniter\Exceptions\ExceptionInterface` and set this rule to check if this interface was implemented to see usage of `new`.

samsonasik: I tried using check a `CodeIgniter\Exceptions\ExceptionInterface`, got 72 errors, see https://github.com/codeigniter4/CodeIgniter4/pull/3868/checks?check_run_id=1370576084 . 

It seems many places need to be updated if we are going to check of `CodeIgniter\Exceptions\ExceptionInterface` to force to use static call for its intantiation.

paulbalandan: OK. Those errors are no joke. It seems there's a lot of contradictions and inconsistencies here.

If I'm to decide, I would go to using `new` for throwing exceptions as it is more natural and cleaner. I think the static methods are unnecessary and would just add to maintenance overhead over time. But as I said earlier, I'm not aware of the original intent behind this pattern.

@MGatner what do you think?

samsonasik: I personally think that we can enforce for instanceof `FrameworkException` in previous commit, as it already in many places used with static functions avaiable as well.

MGatner: Yikes, yes that is a lot of inconsistencies. If the existing implementation for `FrameworkException` is already working with just the one Router exception change let's go with that. This is the looser version so leaves open changes down the road, and I think this is a minor issue compared to the upcoming 4.1

samsonasik: Ok, back to use `FrameworkException` check

utils/PHPStan/CheckFrameworkExceptionInstantiationViaStaticCallRule.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Utils\PHPStan;
+
+use CodeIgniter\Exceptions\FrameworkException;
+use PhpParser\Node;
+use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Name\FullyQualified;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+
+final class CheckFrameworkExceptionInstantiationViaStaticCallRule implements Rule
```
samsonasik, line 14: I think it should be named "CheckFrameworkExceptionInstantiationViaNamedConstructorRule": 

```suggestion
final class CheckFrameworkExceptionInstantiationViaNamedConstructorRule implements Rule
```

PR #3872, opened 2020-11-09
Handle errors() null return

MGatner:
**Description**
An addendum to #3866 where I failed to account for `Model::errors()` returning `null` instead of an empty array.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide


MGatner:
I'm going to pull the trigger on this since it is so small and passing, and I need it in another library's CI.

PR #3070, opened 2020-06-06
Small Feature: debugbar link

enix-app:
I think this useful when we would test or check an URI by click the link without write or copy/paste to address bar. Only GET method.
Support custom placeholder.

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

PR #3873, opened 2020-11-10
Update debug toolbar

totoprayogo1916:
**Description**
Improves the appearance of the Debug Toolbar.
The previous toolbar.css file is more up-to-date than toolbar.scss, (since #3070)
and fixes a cluttered menu (align not centered), compresses the Codeigniter logo.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide
- [ ] User guide updated


paulbalandan (commented):
Can you show a visual representation of the changes here? I tried pulling your PR locally but I did not see noticeable changes. If you can guide me visually on what you changed I would appreciate it.

MGatner (approved):
I approve the code, but please address @paulbalandan 's request. Also curious if @LittleJ is around to take a look?

LittleJ:
@MGatner Thanks for the heads up. All has been done respecting the rules. Just wondering why to supply uncompressed CSS. I think the more we gain weight (and speed) for the final user (which will probably not tweak this toolbar), the best it is. But it is only a (my) point of view :-)

totoprayogo1916:
The one above is the old version.
while the bottom is the current change.

In the old version the menu icon appears to point more upwards.

comparison
cc: @paulbalandan

paulbalandan:
> @MGatner Thanks for the heads up. All has been done respecting the rules. Just wondering why to supply uncompressed CSS. I think the more we gain weight (and speed) for the final user (which will probably not tweak this toolbar), the best it is. But it is only a (my) point of view :-)

I'm in favor of distributing a minified version of the CSS, and probably the JS too, in the next releases. But this might be for another PR. :)

paulbalandan:
> The one above is the old version.
> while the bottom is the current change.
> 
> In the old version the menu icon appears to point more upwards.
> 
> comparison
> cc: @paulbalandan

Ah I see. The changes are subtle but I like it!

paulbalandan (approved):

paulbalandan:
@totoprayogo1916 thank you.

MGatner:
Thanks all! The icon alignment has always bothered me. This is a good change.

PR #3875, opened 2020-11-11
Update Negotiate.php

zl59503020:
Fix automatic language recognition error

**Description**
Fix automatic language recognition error

if request header [Accept-Language] => en-us;en

`
$this->request->setLocale($this->request->negotiate('language', $this->request->config->supportedLocales));
`

### DIFF
--- a/system/HTTP/Negotiate.php
+++ b/system/HTTP/Negotiate.php
@@ -244,7 +244,9 @@ class Negotiate
 			{
 				$param = [];
 				if (preg_match(
-						'/^(?P<name>.+?)=(?P<quoted>"|\')?(?P<value>.*?)(?:\k<quoted>)?$/', $pair, $param
+					'/^(?P<name>.+?)=(?P<quoted>"|\')?(?P<value>.*?)(?:\k<quoted>)?$/',
+					$pair,
+					$param
 				))
 				{
 					$parameters[trim($param['name'])] = trim($param['value']);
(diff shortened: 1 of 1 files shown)


paulbalandan (changes requested):
Instead of checking the resulting `$param` array if empty using loose checking, it would be better to rely on the return value of `preg_match` which returns false when no match is made.

system/HTTP/Negotiate.php
```diff
@@ -243,10 +243,12 @@ public function parseHeader(string $header): array
 			foreach ($pairs as $pair)
 			{
 				$param = [];
```
paulbalandan, line 245:
```suggestion
```

system/HTTP/Negotiate.php
```diff
@@ -243,10 +243,12 @@ public function parseHeader(string $header): array
...
-				preg_match(
+				if (preg_match(
 						'/^(?P<name>.+?)=(?P<quoted>"|\')?(?P<value>.*?)(?:\k<quoted>)?$/', $pair, $param
```
paulbalandan, line 247:
```suggestion
					'/^(?P<name>.+?)=(?P<quoted>"|\')?(?P<value>.*?)(?:\k<quoted>)?$/',
					$pair,
					$param
```

paulbalandan (changes requested):
Last fixup.

paulbalandan (approved):

Issue #3838, opened 2020-10-30
Bug: insertBatch not generating createdField


wiedhodho:
when I use insertBatch, the field createdField not generating current date/timestamp

database: mariadb
feature: insertbatch

paulbalandan:
Did you use the `useTimestamps` set to `true`?

wiedhodho:
Yes, I have set $useTimestamps to true, but it still null on database.

paulbalandan:
Please provide your Model code where you are doing the insertBatch.

paulbalandan:
Please confirm if you set the following correctly:

1. $useTimestamps is set to true
2. the $createdField is included in the array of allowedFields
3. the $createdField is set in the database

wiedhodho:
This is my model code
`
protected $table = 'dokumentasi';
	protected $primaryKey = 'dokumentasi_id';

	protected $returnType = 'object';
	protected $useSoftDeletes = true;

	protected $allowedFields = [
		'dokumentasi_program',
		'dokumentasi_user',
		'dokumentasi_foto1',
		'dokumentasi_foto2',
		'dokumentasi_foto3',
	];

	protected $useTimestamps = true;
	protected $createdField = 'dokumentasi_created';
	protected $updatedField = 'dokumentasi_lastupdate';
	protected $deletedField = 'dokumentasi_deleted';

	protected $validationRules = [
		'dokumentasi_program' => 'required',
		'dokumentasi_user' => 'required',
		'dokumentasi_foto1' => 'required',
		'dokumentasi_foto2' => 'required',
		'dokumentasi_foto3' => 'required',
	];
`

wiedhodho:
oh sorry, this is my fault, I forgot to add createdfield into allowedfields

wiedhodho:
this is my model code:

`protected $table = 'data_miskin';
	protected $primaryKey = 'data_miskin_id';

	protected $returnType = 'object';
	protected $useSoftDeletes = true;

	protected $allowedFields = [
		'data_miskin_penerima',
		'data_miskin_kategori',
		'data_miskin_lastupdate',
		'data_miskin_user',
		'data_miskin_program',
	];

	protected $useTimestamps = true;
	protected $createdField = 'data_miskin_created';
	protected $updatedField = 'data_miskin_lastupdate';
	protected $deletedField = 'data_miskin_deleted';

	protected $validationRules = [
		'data_miskin_penerima' => 'required',
[...]

PR #3884, opened 2020-11-14
Add insertBatch test using timestamp data type

paulbalandan:
**Description**
In addition to the current tests to insertBatch particular on time functions, this PR adds a test to check the behavior using `TIMESTAMP` data type in migration and `datetime` format in `Model::$dateFormat`.

Ancillarily this reduces some lines to make code more compact.

Ref: #3838 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner:
I'm a fan of these cleaner and normalized file formats but they make reviews really difficult, especially since I do many of my reviews on mobile. If it's possible to separate the commits into content and format that would make it a lot easier.

PR #3886, opened 2020-11-15
Fix phpstan error on database

paulbalandan:
**Description**
This one needs arbitration.

PHPStan flags the requirement of docblock using `BaseConnection` but constructor accepts `ConnectionInterface`. `BaseConnection` implements `ConnectionInterface` so it should not pose a problem but maybe phpstan flags this as error because there may be utility methods in `BaseConnection` that are not present in the interface.

This PR changes requirement to `BaseConnection` altogether as it meets the `ConnectionInterface` requirement and may cover the methods unique to BaseConnection. I think this should not be a BC as users would be using the configured DB drivers which already extends `BaseConnection`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -12,20 +12,17 @@
 namespace CodeIgniter\Database;
 
 use BadMethodCallException;
-use CodeIgniter\Database\MySQLi\Connection;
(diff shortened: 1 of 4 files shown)


samsonasik:
@paulbalandan thank you.

PR #3888, opened 2020-11-15
Fix PHPStan notice in latest develop

samsonasik:
`BaseConnection` is already implementing `ConnectionInterface` and the `$db` used is an instanceof `BaseConnection`

**Checklist:**
- [x] Securely signed commits


samsonasik:
All green 🎉

PR #3890, opened 2020-11-16
Rename tests to conform with PSR4

paulbalandan:
**Description**
I am including `tests/system/` to composer's psr4 autoload-dev key so that new subfolders and classes will be read. However, during `composer update` there were test classes marked by Composer as not conforming to PSR4 standard which will be skipped in autoload. These were changed in this PR.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide


samsonasik (approved):

samsonasik:
@paulbalandan thank you.

PR #3891, opened 2020-11-16
DX: Split Model testing into several subunits

### LINKED ISSUE #3792: Dev: Split ModelTest
Model testing happens entirely in one huge file: **tests/system/Database/Live/ModelTest.php**. Not only is this file huge and difficult to navigate, it is grouped under the live database tests which have additional restrictions when quite a few of the tests do not require an actual database.

paulbalandan:
**Description**
Closes #3792 .

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide


paulbalandan:
Test failure is on MigrationRunner and is unrelated.

composer.json
```diff
@@ -36,7 +36,8 @@
 	},
 	"autoload-dev": {
 		"psr-4": {
-			"Utils\\": "utils"
+			"CodeIgniter\\": "tests/system/",
```
MGatner, line 39: This makes me uneasy. Why is it needed?

MGatner, line 39: This makes me uneasy. Why is it needed?

system/Database/MigrationRunner.php
```diff
@@ -956,8 +956,9 @@ public function ensureTable()
 				'null'       => false,
 			],
 			'class'     => [
-				'type' => 'TEXT',
-				'null' => false,
+				'type'       => 'VARCHAR',
+				'constraint' => 255,
+				'null'       => false,
```
MGatner, line 961: While I think this is a good change, are we absolutely sure this is backwards-compatible if anyone has built something on top of the migrations table?

MGatner (changes requested):
Generally looks great! I assume all the test stayed the same, and were just moved to new files, correct? Also a couple comments inline that may or may not need changing.

composer.json
paulbalandan: If I will not include this then `CodeIgniter\Models` namespace will not be found and generally all test classes will not be found.

system/Database/MigrationRunner.php
paulbalandan: I don't know what happened but when I moved the Models tests out of Database Live, I get errors on MigrationRunnerTest::testGetHistory for Sqlsrv. It says that I cannot make equal to comparisons for TEXT. This is what I found to be the solution.

composer.json
paulbalandan: Since Composer's and PHPUnit's autoloaders rely on PSR4, the new namespace `CodeIgniter\Models` will be looked on `system/Models/` but that directory doesn't exist. Including `CodeIgniter\\ => tests/system` as a dev namespace will ensure that autoloaders will find the test classes.

paulbalandan:
> Generally looks great! I assume all the test stayed the same, and were just moved to new files, correct?

Yes, and I just made a central method to generate a model class and tidied up things.

composer.json
MGatner: I guess since it is only on `dev` it should not affect anything else.

system/Database/MigrationRunner.php
MGatner: Okay, it kind of baffles me. Let's see if anyone else has thoughts but I'm conditional okay with it. This had no place being a `TEXT` to being with.

MGatner:
@paulbalandan Can we get this rebased?

MGatner (approved):
PHPStan failure allowed. Looks good, thanks for taking this one on! Pet peeve for a long time

PR #3892, opened 2020-11-16
Update Content Security Policy

paulbalandan:
**Description**
- Add missing docblocks.
- Fix phpstan ignored line.
- Fix formatting
- Fix IDE errors
 
**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/app/Config/ContentSecurityPolicy.php
+++ b/app/Config/ContentSecurityPolicy.php
@@ -120,7 +120,7 @@ class ContentSecurityPolicy extends BaseConfig
 	 * and `<applet>` tags. This directive can't be used in
 	 * `<meta>` tags and applies only to non-HTML resources.
 	 *
-	 * @var string|string[]
+	 * @var string|string[]|null
 	 */
 	public $frameAncestors = null;
 
--- a/app/Config/ContentSecurityPolicy.php
+++ b/app/Config/ContentSecurityPolicy.php
@@ -5,44 +5,155 @@ namespace Config;
 use CodeIgniter\Config\BaseConfig;
 
 /**
- * Class ContentSecurityPolicyConfig
- *
  * Stores the default settings for the ContentSecurityPolicy, if you
  * choose to use it. The values here will be read in and set as defaults
  * for the site. If needed, they can be overridden on a page-by-page basis.
  *
  * Suggested reference for explanations:
- *    https://www.html5rocks.com/en/tutorials/security/content-security-policy/
+ *
+ * @see https://www.html5rocks.com/en/tutorials/security/content-security-policy/
  */
 class ContentSecurityPolicy extends BaseConfig
 {
-	// broadbrush CSP management
+	//-------------------------------------------------------------------------
+	// Broadbrush CSP management
+	//-------------------------------------------------------------------------
 
-	public $reportOnly              = false; // default CSP report context
-	public $reportURI               = null; // URL to send violation reports to
-	public $upgradeInsecureRequests = false; // toggle for forcing https
+	/**
+	 * Default CSP report context
+	 *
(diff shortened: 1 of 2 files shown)


paulbalandan:
Anyone to review? @samsonasik @MGatner ?

system/HTTP/ContentSecurityPolicy.php
```diff
@@ -188,52 +185,47 @@ class ContentSecurityPolicy
...
 	 *
 	 * @param ContentSecurityPolicyConfig $config
 	 */
 	public function __construct(ContentSecurityPolicyConfig $config)
 	{
-		foreach ($config as $setting => $value) // @phpstan-ignore-line
+		foreach (get_object_vars($config) as $setting => $value)
 		{
-			if (isset($this->{$setting}))
+			if (property_exists($this, $setting))
```
samsonasik, line 199: property_exists with isset is different, property_exists got data even value is null. ref https://3v4l.org/m99JJ

paulbalandan: Yes. But isn't it a bug to use `isset`? For example, this class' `$reportURI` property is uninitialized so `isset` to it will evaluate to `false` and I cannot set it in the constructor.

samsonasik: probably, @MGatner any opinion?

MGatner: Yes let's use `property_exists` so we can override the null values.

system/HTTP/ContentSecurityPolicy.php
```diff
@@ -624,17 +584,20 @@ protected function addOption($options, string $target, bool $explicitReporting =
...
-	 * @param ResponseInterface|Response $response
+	 * @param ResponseInterface $response
+	 *
+	 * @return void
 	 */
 	protected function generateNonces(ResponseInterface &$response)
 	{
+		/**
+		 * @var Response $response
+		 */
```
MGatner, line 600: Why is this necessary over ResponseInterface? If someone is using their own Response class then this still needs to work.

MGatner (approved):
Very nice cleanup! Just the one incidental comment.

app/Config/ContentSecurityPolicy.php
```diff
@@ -5,44 +5,155 @@
...
+	 */
+	public $formAction = 'self';
+
+	/**
+	 * Specifies the sources that can embed the current page.
+	 * This directive applies to `<frame>`, `<iframe>`, `<embed>`,
+	 * and `<applet>` tags. This directive can't be used in
+	 * `<meta>` tags and applies only to non-HTML resources.
+	 *
+	 * @var string|string[]
```
samsonasik, line 123: @var string|string[]|null

system/HTTP/ContentSecurityPolicy.php
paulbalandan: I just put it there so that vscode will not turn red everytime I open this file. The current `ResponseInterface` does not have the method used there but is found in `Response`.

MGatner: Hmm that seems to me a larger issue, that code framework code could rely on an incomplete interface. I would be in favor of considering the current ResponseInterface bugged by not accounting for `getBody()` and thus adding it. Thoughts for another Issue perhaps.

paulbalandan: OK, I'll revert that for now.

samsonasik (approved):

paulbalandan:
@samsonasik @MGatner thank you for the reviews

PR #3893, opened 2020-11-16
form_helper - Allow arrays as default and return values in set_value method

najdanovicivan:
**Description**
When array type form fields are used like multiple select it is needed to pass back multiple values as array

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -713,13 +713,13 @@ if (! function_exists('set_value'))
 	 * Grabs a value from the POST array for the specified field so you can
 	 * re-populate an input field or textarea
 	 *
-	 * @param string  $field      Field name
-	 * @param string  $default    Default value
-	 * @param boolean $htmlEscape Whether to escape HTML special characters or not
+	 * @param string          $field      Field name
+	 * @param string|string[] $default    Default value
+	 * @param boolean         $htmlEscape Whether to escape HTML special characters or not
 	 *
-	 * @return string
+	 * @return string|string[]
 	 */
-	function set_value(string $field, string $default = '', bool $htmlEscape = true): string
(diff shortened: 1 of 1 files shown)


MGatner (approved):
Looks good to me. I will leave it in case anyone else wants to set eyes on it. A test case for array input would be awesome, but our `set_value()` tests are already a little on the weak side.

system/Helpers/form_helper.php
```diff
@@ -713,13 +713,13 @@ function form_close(string $extra = ''): string
...
-	 * @param string  $default    Default value
-	 * @param boolean $htmlEscape Whether to escape HTML special characters or not
+	 * @param string       $field      Field name
+	 * @param string|array $default    Default value
+	 * @param boolean      $htmlEscape Whether to escape HTML special characters or not
 	 *
 	 * @return string
 	 */
-	function set_value(string $field, string $default = '', bool $htmlEscape = true): string
+	function set_value(string $field, $default = '', bool $htmlEscape = true)
```
paulbalandan, line 722: Why remove the `: string` return typehint?

najdanovicivan: Becasue the method return value can be string array. Having type hint in such case will throw error.

I see that I missed to update the return type in phpdoc. It should be string|array

najdanovicivan:
I've updated the code to be more clear. Forms in HTML can return string or string[] so instead of array I've updated the PhpDoc to use string|string[] for both default parameter and return type

system/Helpers/form_helper.php
paulbalandan: That is what I'm thinking earlier. Good to see the changes.

paulbalandan (approved):

PR #3895, opened 2020-11-16
Model: support returning array indexed by a specified column

rmilecki:
```
This allows Model users to get array or results indexed with a specific
column, $primaryKey by default. It can be simply used as:

$model->indexed()->findAll()

It may be useful e.g. for generating array to be used with
form_dropdown().

Signed-off-by: Rafał Miłecki <rafal@milecki.pl>
```

**Description**
I personally use it for `form_dropdown()` and found someone else requesting for it as well in forum thread [ Query result as an indexed array or object by index of choice ](https://forum.codeigniter.com/thread-76215.html). I was told to submit a Pull Request, so I share my code.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -195,6 +195,11 @@ class Model
 	protected $tempReturnType;
 
 	/**
+	 * Used by indexed() to return array indexed by a specified key
+	 */
+	protected $tempReturnIndexed = false;
+
+	/**
 	 * Whether we should limit fields in inserts
 	 * and updates to those available in $allowedFields or not.
 	 *
@@ -446,10 +451,16 @@ class Model
 
 		$row = $row->getResult($this->tempReturnType);
 
+		if ($this->tempReturnIndexed)
+		{
+			$row = array_column($row, null, $this->tempReturnIndexed);
+		}
+
 		$eventData = $this->trigger('afterFind', ['data' => $row, 'limit' => $limit, 'offset' => $offset]);
 
 		$this->tempReturnType     = $this->returnType;
 		$this->tempUseSoftDeletes = $this->useSoftDeletes;
+		$this->tempReturnIndexed  = false;
 
 		return $eventData['data'];
 	}
@@ -487,6 +498,7 @@ class Model
 
 		$this->tempReturnType     = $this->returnType;
 		$this->tempUseSoftDeletes = $this->useSoftDeletes;
+		$this->tempReturnIndexed  = false;
 
 		return $eventData['data'];
 	}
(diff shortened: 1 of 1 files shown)


system/Model.php
```diff
@@ -1068,6 +1080,22 @@ public function asObject(string $class = 'object')
...
+	/**
+	 * Sets the return to be array indexed by a specific key (column table).
+	 *
+	 * @param string $key Column to use for indexing (defaults to $primaryKey)
+	 *
+	 * @return Model
+	 */
+	public function indexed($key = null)
+	{
+		$this->tempReturnIndexed = $key === null ? $this->primaryKey : $key;
```
OJMichael, line 1092: This could also be written as:

```php
$this->tempReturnIndexed = $key ?? $this->primaryKey;
```

paulbalandan (changes requested):
Please update your fork first then rebase the changes to your `model-indexed` branch. You can use these steps, assuming `origin` is your fork's remote and `upstream` is the remote of codeigniter4 repo.

```
git checkout develop
git pull upstream develop
git push
git checkout model-indexed
git rebase develop
git push --force
```

craigrodway:
The Yii framework has this functionality in their Array helper, which I think is a better place to put it rather than only in the model - allowing for more re-use.

https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseArrayHelper.php#L367-L500

sfadschm:
> The Yii framework has this functionality in their Array helper, which I think is a better place to put it rather than only in the model - allowing for more re-use.
> 
> https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseArrayHelper.php#L367-L500

I also implemented such functionality in my CI4 array helper (although not as advanced).
Maybe a combination would be nice, so that we have the Array helper available for all occasions, but still have the proposed `indexed` method in the model for convenience? (Which would then of course use the helper method.)

sfadschm:
Just some more notes on this, as I am thinking about something similar atm:
1. The database column which is used to reindex the array might not have the `unique` attribute, so handling of duplicate values should be considered. (This could, e.g., be canceling the reindexing totally, keeping only the first/last occurence of a key value or even grouping array elements with the same key value into subarrays.)
2. Model queries might return arrays of entities. In this case, a simple `array_column` will likely not work.
3. Just an idea right now, but maybe the `array_helper` functions, such as `dot_array_search` could be used to add more flexibility, but might also increase load.

paulbalandan:
@rmilecki are you able to do the changes and fork update?? If we'll not hear from you soon, we'll close this as abandoned.

PR #3897, opened 2020-11-17
pre-commit hook does not work with alias

najdanovicivan:
**Description**
In some workflows like  MAMP on OSX default php version is set using alias in bash profile

```
alias php='/Applications/MAMP/bin/php/php7.4.2/bin/php -c "/Library/Application Support/appsolute/MAMP PRO/conf/php7.4.2.ini"'
```

In such case when phpstan and codesniffer are executed via git hoot it will run them using the default system php instead of using the alias.

Prefixing the method call with php triggers the alias and solves the issue.
**Checklist:**
- [x ] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/admin/pre-commit
+++ b/admin/pre-commit
@@ -35,7 +35,7 @@ if [ "$FILES" != "" ]
 then
 	echo "Running PHPStan..."
 	# Run on whole codebase
-	./vendor/bin/phpstan analyse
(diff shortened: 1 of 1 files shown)


paulbalandan (approved):

samsonasik:
thank you @najdanovicivan

PR #3900, opened 2020-11-18
Handling requests sent back from filters

caswell-wc:
Expands the fix of #3085

**Description**
This is a recreation of #3544 to eliminate issues with the rebase messing up that request. 

Filters have the ability to alter the request like adding a user object from an auth filter for example. When it does that the request gets altered implicitly by the filter because the filter's request is pointing to the same object as the request property of the Codeigniter object. If the request gets altered, the filters also return the request into the $possibleResponse variable. This change makes the update to the request property more explicit by setting the property with what gets returned from the filters. This ensures that if for some reason the object pointers get messed up, the request used by the controllers is the altered one.

**Checklist:**
- [x] Securely signed commits
- [ ] N/A Component(s) with PHPdocs
- [ ] N/A User guide updated
- [x] Conforms to style guide


caswell-wc:
@MGatner @michalsn I have tested this in my project to make sure it works but I have not found a way to create a PHPUnit test for this. If you have any suggestions on how to create a test for it, I would appreciate it.

system/CodeIgniter.php
```diff
@@ -407,6 +407,8 @@ protected function handleRequest(RouteCollectionInterface $routes = null, Cache
 			{
 				return $returnResponse ? $possibleResponse : $possibleResponse->pretend($this->useSafeOutput)->send();
 			}
+
+			$this->request = $possibleResponse;
```
MGatner, line 411: We need to check the actual return type. `Filters::run()` has the following behavior:
1. If a Filter returns a `RequestInterface` then store it and continue
2. If a Filter returns a `ResponseInterface` then return it and quit (handled just above)
3. If a Filter returns something `empty()` then keep running filters
4. If a Filter returns anything else return it

Case 4 could cause us issues, since `$this->request` needs to be "the current Request object". So something like:

```suggestion
			if ($possibleResponse instanceof RequestInterface)
			{
				$this->request = $possibleResponse;
			}
```

MGatner (changes requested):

system/CodeIgniter.php
caswell-wc: I missed that 4th case so that makes sense. I have pushed up that change.

caswell-wc:
@MGatner I am bumping into PHPStan issues with this that I'm not real sure how to resolve. Could you take a look?

paulbalandan:
```
  590    No error to ignore is reported on line 590.                       
  851    No error to ignore is reported on line 851.                       
```
There is a comment before those lines `@phpstan-ignore-line` or `@phpstan-ignore-next-line`. You just need to remove those comments.

```
  813    Parameter #2 $request of static method                            
         CodeIgniter\Config\Services::router() expects                     
         CodeIgniter\HTTP\Request|null, CodeIgniter\HTTP\RequestInterface  
         given.                                                            
```
Just revert your changes to this line and this will go away. Currently you changed:
```diff
-        * @var HTTP\Request|HTTP\IncomingRequest|CLIRequest
+	 * @var HTTP\Request|HTTP\IncomingRequest|CLIRequest|RequestInterface
 	 */
 	protected $request;
```


```
     Ignored error pattern #Access to an undefined property                     
     CodeIgniter\\HTTP\\Request::\$uri# was not matched in reported errors.     
     Ignored error pattern #Call to an undefined method                         
     CodeIgniter\\HTTP\\Request::(getPath|getSegments|getMethod|setLocale|getP  
     ost)\(\)# was not matched in reported errors.
```
These two are in the `ignoredErrors` section of `phpstan.neon.dist` file. Just delete those two from there.

caswell-wc:
@paulbalandan I made the changes that you suggested and that seemed to cause new errors. Any ideas?

MGatner:
This is part of the mess of the HTTP layer that I referenced I have been dealing with. I don't think there is a clean fix, but my recommendation is to check if `$possibleResponse instance of Request` instead and that will ensure the correct type is in place. 

The other SA issues I will have to look at not on mobile, but it is possible that PHPStan version has changed and some errors are now detected or not different than before.

system/CodeIgniter.php
```diff
@@ -407,6 +407,11 @@ protected function handleRequest(RouteCollectionInterface $routes = null, Cache
 			{
 				return $returnResponse ? $possibleResponse : $possibleResponse->pretend($this->useSafeOutput)->send();
 			}
+
+			if ($possibleResponse instanceof RequestInterface)
```
paulbalandan, line 411: Change this to this will solve two of phpstan errors.
```suggestion
			if ($possibleResponse instanceof Request)
```

phpstan.neon.dist
```diff
@@ -30,12 +30,10 @@ parameters:
 		- '#Access to an undefined property CodeIgniter\\Database\\Forge::\$dropConstraintStr#'
 		- '#Access to an undefined property CodeIgniter\\Database\\BaseConnection::\$mysqli|\$schema#'
 		- '#Access to an undefined property CodeIgniter\\Database\\ConnectionInterface::(\$DBDriver|\$connID|\$likeEscapeStr|\$likeEscapeChar|\$escapeChar|\$protectIdentifiers|\$schema)#'
-		- '#Access to an undefined property CodeIgniter\\HTTP\\Request::\$uri#'
```
paulbalandan, line 33: Reinstate this line and the other one below will solve your other errors.

phpstan.neon.dist
```diff
@@ -30,12 +30,10 @@ parameters:
...
 		- '#Access to protected property CodeIgniter\\Database\\BaseConnection::(\$DBDebug|\$DBPrefix|\$swapPre|\$charset|\$DBCollat|\$database)#'
 		- '#Call to an undefined method CodeIgniter\\Database\\BaseConnection::_(disable|enable)ForeignKeyChecks\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Database\\BaseConnection::supportsForeignKeys\(\)#'
 		- '#Call to an undefined method CodeIgniter\\Database\\ConnectionInterface::(tableExists|protectIdentifiers|setAliasedTables|escapeIdentifiers|affectedRows|addTableAlias|getIndexData)\(\)#'
-		- '#Call to an undefined method CodeIgniter\\HTTP\\Request::(getPath|getSegments|getMethod|setLocale|getPost)\(\)#'
```
paulbalandan, line 38: Reinstate.

MGatner (approved):

PR #3901, opened 2020-11-18
Array_helper: Add function array_sort_by_multiple_keys.

sfadschm:
**Description**
This PR adds a convenience method to CI's `array_helper` to allow sorting a multidimensional array by a list of columns contained in the array. As I find myself using this regularly, I thought maybe others would appreciate this functionality, too.

The helper function takes the reference to an array `&$array` that is to be sorted and an associative array `$sortColumns` of column names and SORT_* flags.
The referenced array is looped and a onedimensional array of values for each column defined in the `$sortColumns` array is generated. Then all of these arrays are passed to the array_multisort function together.

This results in `&$array` being sorted by all of the defined columns and flags hierarchically, which might be useful, e.g., for sorting arrays of entities returned from models.

The method works on both arrays containing sub-arrays and arrays containing objects (e.g., `Entities`).
By using the `.` separator in the column names in `$sortColumns`, deeper levels of `&$array` can be accessed for sorting (e.g., when sorting a list of entities loaded with their relations from Tatter\Relations).

Opinions and improvements very welcome :)

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

Not sure about the style guide, but I hope it does conform.
Will update user guide if feature is considered useful.

### DIFF
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -123,7 +123,7 @@ if (! function_exists('array_deep_search'))
 	}
 }
 
-if (!function_exists('array_sort_by_multiple_keys'))
+if (! function_exists('array_sort_by_multiple_keys'))
 {
 	/**
 	 * Sorts a multidimensional array by its elements values. The array
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -196,11 +196,7 @@ if (!function_exists('array_sort_by_multiple_keys'))
 		// Append the array as reference
 		$tempArray[] = &$array;
 
-		// The array_multisort method accepts all sorting arrays and
-		// flags as an argument list. As we have those stored in a
-		// dynamically sized array, we use the call_user_func_array
-		// function, which parses the array to a list and then passes
-		// it to array_multisort.
-		return call_user_func_array('array_multisort', $tempArray);
+		// Pass sorting arrays and flags as an argument list.
+		return array_multisort(...$tempArray);
 	}
 }
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -122,3 +122,82 @@ if (! function_exists('array_deep_search'))
 		return null;
 	}
 }
+
+if (!function_exists('sort_by_multiple_keys'))
+{
+
+	/**
+	 * Sorts an associative multidimensional array by the values of the
+	 * passed columns names. The sorting params represent an associative
+	 * array of column names and sorting flags.
+	 * Both arrays of objects and arrays of array can be sorted.
+	 *
+	 * Example:
+	 * 	sort_by_multiple_keys($players,
+	 * 			[
+	 * 				'team.hierarchy' => SORT_ASC,
+	 * 				'position'       => SORT_ASC,
+	 * 				'name'           => SORT_STRING,
+	 * 			]
+	 * 		);
+	 * The '.' dot operator in the column name indicates a deeper array or
+	 * object level. In principle, any number of sublevels could be used,
+	 * as long as the level and column exist in every array element.
+	 *
+	 * @param array &$array the reference of the array to be sorted
+	 * @param array $sortColumns an associative array of columns to sort
+	 *              after and their sorting flags
+	 *
+	 * @return boolean
+	 */
+	function array_sort_by_multiple_keys(array &$array, array $sortColumns): bool
+	{
+		// Check if there really are columns to sort after
+		if (empty($sortColumns) || empty($array))
+		{
+			return $array;
+		}
+
+		// Group sorting indexes and data
+		$tempArray = [];
+		foreach ($sortColumns as $key => $sortFlag)
+		{
+			// Get sorting values
+			$carry = $array;
+
+			// The '.' operator separates nested elements
+			foreach (explode('.', $key) as $keySegment)
+			{
+				// Distinguish array of objects and arrays of arrays
+				if (is_object(reset($carry)) ?? false)
+				{
+					// Extract the object attribute
+					foreach ($carry as $index => $object)
+					{
+						$carry[$index] = $object->$keySegment;
+					}
+				}
+				else
+				{
+					// Extract
+					$carry = array_column($carry, $keySegment);
+				}
+			}
+
+			// Store this iterations sorting paramters
+			$tempArray[$key][0] = $carry;
+			$tempArray[$key][1] = $sortFlag;
+		}
+
+		// Prepare sorting array
+		$sortArray = [];
+		foreach ($sortColumns as $index => $sortFlag)
+		{
+			$sortArray[] = &$tempArray[$index][0];
+			$sortArray[] = &$tempArray[$index][1];
+		}
+		$sortArray[] = &$array;
+
+		return call_user_func_array('array_multisort', $sortArray);
+	}
+}
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -143,7 +143,7 @@ if (!function_exists('sort_by_multiple_keys'))
 	 * object level. In principle, any number of sublevels could be used,
 	 * as long as the level and column exist in every array element.
 	 *
-	 * @param array &$array the reference of the array to be sorted
+	 * @param array $array the reference of the array to be sorted
 	 * @param array $sortColumns an associative array of columns to sort
 	 *              after and their sorting flags
 	 *
@@ -154,7 +154,7 @@ if (!function_exists('sort_by_multiple_keys'))
 		// Check if there really are columns to sort after
 		if (empty($sortColumns) || empty($array))
 		{
-			return $array;
+			return false;
 		}
 
 		// Group sorting indexes and data
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -134,15 +134,19 @@ if (!function_exists('array_sort_by_multiple_keys'))
 	 *
 	 * Example:
 	 * 	array_sort_by_multiple_keys($players,
-	 * 			[
-	 * 				'team.hierarchy' => SORT_ASC,
-	 * 				'position'       => SORT_ASC,
-	 * 				'name'           => SORT_STRING,
-	 * 			]
-	 * 		);
+	 * 		[
+	 * 			'team.hierarchy' => SORT_ASC,
+	 * 			'position'       => SORT_ASC,
+	 * 			'name'           => SORT_STRING,
+	 * 		]
+	 * 	);
+	 *
 	 * The '.' dot operator in the column name indicates a deeper array or
 	 * object level. In principle, any number of sublevels could be used,
 	 * as long as the level and column exist in every array element.
+	 *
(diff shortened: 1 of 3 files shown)


paulbalandan (changes requested):
We need some tests for this and mentions in the user guide.

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,81 @@ function array_deep_search($key, array $array)
...
+			foreach (explode('.', $key) as $keySegment)
+			{
+				// Distinguish array of objects and arrays of arrays
+				if (is_object(reset($carry)) ?? false)
+				{
+					// Extract the object attribute
+					foreach ($carry as $index => $object)
+					{
+						$carry[$index] = $object->$keySegment;
+					}
```
samsonasik, line 177: use `continue` after foreach, and `else` no longer needed.

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,81 @@ function array_deep_search($key, array $array)
...
+		foreach ($sortColumns as $key => $sortFlag)
+		{
+			// Get sorting values
+			$carry = $array;
+
+			// The '.' operator separates nested elements
+			foreach (explode('.', $key) as $keySegment)
+			{
+				// Distinguish array of objects and arrays of arrays
+				if (is_object(reset($carry)) ?? false)
```
paulbalandan, line 171: Why is there a need for the null coalescing here? `is_object` will always return a boolean so no chance the right side would be reached.

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,81 @@ function array_deep_search($key, array $array)
...
+		// Prepare sorting array
+		$sortArray = [];
+		foreach ($sortColumns as $index => $sortFlag)
+		{
+			$sortArray[] = &$tempArray[$index][0];
+			$sortArray[] = &$tempArray[$index][1];
+		}
+		$sortArray[] = &$array;
+
+		return call_user_func_array('array_multisort', $sortArray);
```
paulbalandan, line 200: Why did you use `call_user_func_array` instead of calling `array_multisort` directly?

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,81 @@ function array_deep_search($key, array $array)
...
+			}
+
+			// Store this iterations sorting paramters
+			$tempArray[$key][0] = $carry;
+			$tempArray[$key][1] = $sortFlag;
+		}
+
+		// Prepare sorting array
+		$sortArray = [];
+		foreach ($sortColumns as $index => $sortFlag)
```
paulbalandan, line 193: I think this could be combined with the first similar `foreach`. Just remove or simplify the intermediary steps.

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,81 @@ function array_deep_search($key, array $array)
 		return null;
 	}
 }
+
+if (!function_exists('sort_by_multiple_keys'))
```
paulbalandan, line 126:
```suggestion
if (! function_exists('array_sort_by_multiple_keys'))
```

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,81 @@ function array_deep_search($key, array $array)
...
+{
+	/**
+	 * Sorts an associative multidimensional array by the values of the
+	 * passed columns names. The sorting params represent an associative
+	 * array of column names and sorting flags.
+	 * Both arrays of objects and arrays of array can be sorted.
```
paulbalandan, line 132:
```suggestion
	 * Sorts an associative multidimensional array by the values of the
	 * passed columns' names. The sorting params represent an associative
	 * array of column names and sorting flags.
	 *
	 * Both array of objects and array of arrays can be sorted.
```

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,81 @@ function array_deep_search($key, array $array)
...
+	 *
+	 * Example:
+	 * 	sort_by_multiple_keys($players,
```
paulbalandan, line 135:
```suggestion
	 * 	array_sort_by_multiple_keys($players,
```

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,81 @@ function array_deep_search($key, array $array)
...
+	 * 				'name'           => SORT_STRING,
+	 * 			]
+	 * 		);
+	 * The '.' dot operator in the column name indicates a deeper array or
+	 * object level. In principle, any number of sublevels could be used,
+	 * as long as the level and column exist in every array element.
+	 *
+	 * @param array $array the reference of the array to be sorted
+	 * @param array $sortColumns an associative array of columns to sort
+	 *              after and their sorting flags
```
paulbalandan, line 148:
```suggestion
	 * @param array $array             the reference of the array to be sorted
	 * @param array $sortColumns an associative array of columns to sort
	 *                                                after and their sorting flags
```

The suggestion preview looks confusing. I suggested here that the explanations for the params should be aligned too.

system/Helpers/array_helper.php
sfadschm: Correct. Fixed.

system/Helpers/array_helper.php
sfadschm: True, thats a remainder from an old implementation, unfortunately...

system/Helpers/array_helper.php
sfadschm: You're right, thanks!.
I always thought that all array parameters of `array_multisort` need to be references. Turns out they do not.
Fixed.

system/Helpers/array_helper.php
sfadschm: Fixed.

system/Helpers/array_helper.php
sfadschm: Fixed.

system/Helpers/array_helper.php
sfadschm: Fixed.

system/Helpers/array_helper.php
sfadschm: I rewrote this section, sounded a little confusing reading it again.

system/Helpers/array_helper.php
sfadschm: To enable sorting with multiple priority levels  `array_multisort` needs to receive each of the value arrays followed by the respective sort flag(s) in a single call. As it only accepts a list, this was the solution I came up with.
If we just call `array_multisort` in each of the `foreach` loops with a single 'value array' and flag, then in the end the `&$array` will only be sorted by the last entry in `$sortColumns`.

sfadschm:
Working on user guide and tests (have to figure out how test work first :D).

sfadschm:
Added documentation. Is there any way I can see the parsed output of the `.rst`?

system/Helpers/array_helper.php
paulbalandan: > If we just call `array_multisort` in each of the `foreach` loops with a single 'value array' and flag, then in the end the `&$array` will only be sorted by the last entry in `$sortColumns`.

But I see it called *after* the `foreach` loop, not during. Am I missing something?

sfadschm: Its called after `foreach`, because we want it sorted by all of the data we collected in the `foreach `at once.

See `Example #3` here:
https://www.php.net/manual/de/function.array-multisort.php

We need to feed a list of onedimensional arrays and sorting flags to `array_multisort`. In our case, we don't know the length of this list, as it depends on the `$sortColumns` that the user passes. Therefore, we collect all arguments in a single array and feed it to `array_multisort `via `call_user_func_array`, which translates the `$tempArray` into a list of function arguments.
I hope this makes it clearer :)

paulbalandan: Ah I get it now. You are unpacking the arguments to `array_multisort`. Maybe we can have a small comment there explaining the need for the `call_user_func_array` so that future refactors won't mess with that.

Thanks for the thorough explanation, though. 😁

sfadschm: Alright :D Will push that with the tests to save on phpunit time.

sfadschm: Its in.

sfadschm:
> 
> 
> We need some tests for this and mentions in the user guide.

Added some test. Not sure if they way I did is legit though (espacially the `sortByMultipleKeysSeeder` method)...

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,85 @@ function array_deep_search($key, array $array)
 		return null;
 	}
 }
+
+if (!function_exists('array_sort_by_multiple_keys'))
```
paulbalandan, line 126:
```suggestion
if (! function_exists('array_sort_by_multiple_keys'))
```

tests/system/Helpers/ArrayHelperTest.php
```diff
@@ -181,4 +251,58 @@ public function deepSearchProvider()
 			],
 		];
 	}
-}
+
+	public function sortByMultipleKeysProvider()
```
paulbalandan, line 255:
```suggestion
	public static function sortByMultipleKeysProvider(): iterable
```

You can combine this with your seeder below, so that when you feed the data to your test methods they can accept three arguments: `public function testSomething($input, $sortColumns, $expected)`.

tests/system/Helpers/ArrayHelperTest.php
```diff
@@ -181,4 +251,58 @@ public function deepSearchProvider()
...
+			2 => [
+				'name'     => 'Frank',
+				'position' => 1,
+				'team'     => [
+					'order' => 1,
+				],
+			],
+		];
+	}
+}
```
paulbalandan, line 308: Missing final new line.

system/Helpers/array_helper.php
```diff
@@ -122,3 +122,85 @@ function array_deep_search($key, array $array)
...
+
+		// Append the array as reference
+		$tempArray[] = &$array;
+
+		// The array_multisort method accepts all sorting arrays and
+		// flags as an argument list. As we have those stored in a
+		// dynamically sized array, we use the call_user_func_array
+		// function, which parses the array to a list and then passes
+		// it to array_multisort.
+		return call_user_func_array('array_multisort', $tempArray);
```
paulbalandan, line 204: On second thought, I think we can use here the splat operator.

```suggestion
		return array_multisort(...$tempArray);
```

sfadschm: You are very right! Guess this wasn't available yet when I first wrote the method.

tests/system/Helpers/ArrayHelperTest.php
sfadschm: Fixed.

tests/system/Helpers/ArrayHelperTest.php
sfadschm: Done.

system/Helpers/array_helper.php
sfadschm: Fixed.

paulbalandan (approved):

PR #3907, opened 2020-11-19
Throttle: allow removal of throttler buckets

element-code:
This allows to manual remove throttler buckets manually.
E.g. after paying a plan or a successful login.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Throttle/Throttler.php
+++ b/system/Throttle/Throttler.php
@@ -150,12 +150,19 @@ class Throttler implements ThrottlerInterface
 		return false;
 	}
 
-	public function remove(string $key) : void
+	/**
+	 * @param string $key The name of the bucket
+	 *
+	 * @return $this
+	 */
+	public function remove(string $key): self
 	{
 		$tokenName = $this->prefix . $key;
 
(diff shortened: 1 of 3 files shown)


system/Throttle/Throttler.php
```diff
@@ -151,6 +151,14 @@ public function check(string $key, int $capacity, int $seconds, int $cost = 1):
 		return false;
 	}
 
+	public function remove(string $key) : void
```
paulbalandan, line 154:
```suggestion
	public function remove(string $key): void
```

paulbalandan (changes requested):
This needs some tests and user guide mention.

lonnieezell:
@element-code are you able to add tests to this?

element-code:
Yes, but it will take some time, see https://github.com/codeigniter4/CodeIgniter4/pull/3911#issuecomment-740486515
Help is appreciated :)

paulbalandan (approved):

PR #3909, opened 2020-11-19
Array_helper: Add missing test for _array_search_dot.

sfadschm:
**Description**
With this PR, `array_helper` tests will have 100% coverage.

However, I am not sure, if the current behavior of `_array_search_dot` is desired.

https://github.com/codeigniter4/CodeIgniter4/blob/e6bfd328b3353e178c8b3fe852dbfcccd940501c/system/Helpers/array_helper.php#L90

This line can only be reached, if either the current index is not the last in the dot-notation search string (i.e., we want to search in a deeper level) and at the same time `$array[$currentIndex]` is not an array (i.e., there is no deeper level to search in).

So instead of failing because the search depth is smaller than the array depth, we just break the search and return the deepest level found. Is that desired?

**Checklist:**
- [ ] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Helpers/ArrayHelperTest.php
+++ b/tests/system/Helpers/ArrayHelperTest.php
@@ -167,7 +167,7 @@ class ArrayHelperTest extends \CodeIgniter\Test\CIUnitTestCase
 
 	//--------------------------------------------------------------------
 
-	public function deepSearchProvider()
+	public static function deepSearchProvider()
 	{
 		return [
 			[
--- a/tests/system/Helpers/ArrayHelperTest.php
+++ b/tests/system/Helpers/ArrayHelperTest.php
@@ -19,6 +19,17 @@ class ArrayHelperTest extends \CodeIgniter\Test\CIUnitTestCase
 		$this->assertEquals(23, dot_array_search('foo.bar', $data));
 	}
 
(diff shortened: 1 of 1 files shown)


paulbalandan (approved):

kenjis:
@sfadschm @paulbalandan 

Why not `null`?
```php
$data = [
    'foo' => [
        'bar' => 23,
     ],
];
dot_array_search('foo.bar.baz', $data); // returns 23
```

sfadschm:
Yep, thats what I meant in the description.
This should either be null or throw an exception, as the data we pass to the method should match the search patterns normally.

However, this PR was just to document the current behaviour ☺

I believe changing the behavior here would be a BC break, but I do not think that people actually rely on the fallback, as it only activates if the search pattern has more levels than the array.

sfadschm:
Yep, thats what I meant in the description.
This should either be null or throw an exception, as the data we pass to the method should match the search patterns normally.

However, this PR was just to document the current behaviour ☺

I believe changing the behavior here would be a BC break, but I do not think that people actually rely on the fallback, as it only activates if the search pattern has more levels than the array.

kenjis:
@sfadschm Thank you for quick feedback!

I think this is a bug. I opened an issue #5939

PR #3911, opened 2020-11-19
CLI: Prompt: allow array-based validation rules

### LINKED ISSUE #3766: CLI: method prompt should accept array validation rules
When using validation for `promt` it should accept either a
- validation rule set name
- validation rules, as string, separated by a pipe
- validation rules, as array

element-code:
This fixes #3766

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


element-code:

lonnieezell:
Changes seem fine, but a) we need tests, b) we need user guide updates, and c) you've left the PR in draft mode which indicates you're still working on it. Are you able to add this?

element-code:
@lonnieezell answering for all open PR's on my side on general questions:
- I can only work on CI in small time-slots so kinda slow progress here -- and the draft mode on.
- I have literally zero experience with tests so i would enjoy some help on this end
- I will provide updates to the user guide

MGatner:
@element-code If you can get User Guide updated I can do some tests for this.

element-code:
Hey @MGatner or @paulbalandan  if you can give me some basic input on the following things, I would try writing them myself.
- which tests are required for this method or recommended
- how to run tests in my local environment (steps to take, no need for detailed explanations)

MGatner:
@element-code im on mobile so can't do much for you on specifics... but you should be able to look at the existing greats for Validation rule in tests/system/Valdiation to figure out some similar ones. Locally, run `composer install` and `composer test` to launch the test suite - that should do it!

paulbalandan:
Currently, `CLI::prompt` is not testable because it requires keyboard input which makes PHPUnit hanging. @element-code maybe I can pull locally to test it myself but it is currently a draft so I believe you still have more to add?

element-code:
@paulbalandan it was in draft since the tests and user guide update were missing (so that you guys don't have to review it and tell me that there are tests missing 😀)
Would be nice if you will do that.

paulbalandan:
@element-code ok I can do that, but can you resolve the conflicts first?

sfadschm:
> 
> 
> Currently, `CLI::prompt` is not testable because it requires keyboard input which makes PHPUnit hanging. @element-code maybe I can pull locally to test it myself but it is currently a draft so I believe you still have more to add?

I have not tested this yet, but can we do something like in the answer here with our `CLI::prompt`?
https://stackoverflow.com/questions/50857595/test-php-file-with-console-input-with-phpunit/50857931

paulbalandan:
> > Currently, `CLI::prompt` is not testable because it requires keyboard input which makes PHPUnit hanging. @element-code maybe I can pull locally to test it myself but it is currently a draft so I believe you still have more to add?
> 
> I have not tested this yet, but can we do something like in the answer here with our `CLI::prompt`?
> https://stackoverflow.com/questions/50857595/test-php-file-with-console-input-with-phpunit/50857931

I think that will not work 😂. I'm thinking this past week of introducing an optional `$noInteraction` param to `CLI::prompt` so that if it is `true`, it will directly use the default value if given.

sfadschm:
😆  I was thinking about something similar.
Using the default value will help many cases, but in some cases we might want  to try to willingly insert wrong values to also test command abortion/failure.
In that case we would rather need to provide an optional array like `promptSubstitutions` where we constantly `shift` the first value each time we would normally call `prompt`.
But not so beautiful either.

paulbalandan:
The current build of the `CLI` library makes us constrained to expand its functionalities. I was thinking months back that prompts should have something similar to Symfony's defaulting when `--no-interaction` flag is passed. And other things..

With my proposed `$noInteraction` param, then in a spark command we can just pass in the value of `CLI::getOption('no-interaction')` to that parameter. For manual testing, we can just type there a boolean.

sfadschm:
Oh I see, that seems reasonable. I thought `$noInteraction` was intended to always be boolean only. But the way you explained seems useful.

system/CLI/CLI.php
```diff
@@ -264,7 +272,7 @@ public static function prompt(string $field, $options = null, string $validation
 		// Read the input from keyboard.
 		$input = trim(static::input()) ?: $default;
 
-		if (isset($validation))
+		if ($validation)
 		{
 			while (! static::validate($field, $input, $validation))
```
paulbalandan, line 277: $validation->setRules() which is used by validate expects string rules.

```suggestion
			$validation = trim(implode('|', $validation), '|');

			while (! static::validate($field, $input, $validation))
```

system/CLI/CLI.php
```diff
@@ -14,6 +14,7 @@
 use CodeIgniter\CLI\Exceptions\CLIException;
 use Config\Services;
 use Throwable;
+use InvalidArgumentException;
```
paulbalandan, line 17: Put this before use Throwable

system/CLI/CLI.php
```diff
@@ -220,17 +221,26 @@ public static function input(string $prefix = null): string
...
-	public static function prompt(string $field, $options = null, string $validation = null): string
+	public static function prompt(string $field, $options = null, $validation = null): string
 	{
 		$extraOutput = '';
 		$default     = '';
 
+		if ($validation && ! is_array($validation) && ! is_string($validation))
+		{
+			throw new InvalidArgumentException('$rules can only be of type string|array');
+		}
```
paulbalandan, line 238: Add new line

paulbalandan (commented):
These are my comments while locally testing this.

system/CLI/CLI.php
element-code: It must have been a mistake on my part as I tested this with several rules.
I would not drop support of array-based validation rules since it is required for features like anonymous functions as rules. Instead i'd chanche `validate` to internaly use `setRules` which supports array-based rules?

system/CLI/CLI.php
paulbalandan: @element-code Rector is complaining. You must've added trailing whitespace while adding the new line.

system/CLI/CLI.php
paulbalandan: That would not be possible at the moment without a breaking change. Validation::setRule has this signature
```
public function setRule(string $field, string $label = null, string $rules, array $errors = [])
```
Removing the string typehint would be breaker for existing user implementations overriding the method since this is declared public.

element-code: Sorry, i wasnt clear about that. I'd wanted to change to use setRule**s** which can handle that.
```php
$validation->setRules([$field => [
    'label' => $label,
    'rules' => $rules,
]]);
```

system/CLI/CLI.php
element-code: Yeah, thanks to the Github Code-Editor 🙃

system/CLI/CLI.php
paulbalandan: Ah. So you'd be swapping `setRule` with `setRules`. Fair enough. That's ok with me.

paulbalandan (approved):

PR #3913, opened 2020-11-19
Feature: Validation: allow anonymous functions as rules

element-code:
Allows to use an anonymous function with a name as a rule:
```php
$rules = [
  'max_length[191]',
  [
    'foobar',
    function(mixed $value, array $data) {
      return (bool) $value;
    },
  ],
];
```
Anonymous-function rules also require to be wrapped in an array and have a name for errors as first array item

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Language/en/Validation.php
+++ b/system/Language/en/Validation.php
@@ -17,6 +17,8 @@ return [
    'groupNotFound'         => '{0} is not a validation rules group.',
    'groupNotArray'         => '{0} rule group must be an array.',
    'invalidTemplate'       => '{0} is not a valid Validation template.',
+   'invalidRule'           => '{0} can\'t be used as a rule.',
+   'unnamedRule'           => '{0} has to have a name.',
 
 	// Rule Messages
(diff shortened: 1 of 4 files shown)


element-code:

paulbalandan (changes requested):
Aside from the user guide update, this would need some testing.

lonnieezell:
I have no problems with this addition, but will need tests and user guide updates or we'll have to close. @element-code are you able to add those?

element-code:
Yes, but it will take some time, see https://github.com/codeigniter4/CodeIgniter4/pull/3911#issuecomment-740486515
Help is appreciated.

MGatner:
I think this is maybe related to your other Validation PR? I will wait to hear back on the one first.

element-code:
@MGatner no, this is completely detatched (and doesnt support placeholders anyway rn) :)

lonnieezell:
This has gone over a year without the tests being added. Closing.

PR #3916, opened 2020-11-23
Group routes without a prefix.

francis94c:

**Description**
I ran into bit of an issue while grouping my routes and wanted to achieve something like the below.
```
$routes->group('v1', ['namespace' => 'App\Controllers\V1'], function ($routes) {
    $routes->post('verify/begin', 'VerificationController::generateVerificationCode');
    $routes->post('verify/code', 'VerificationController::verifyCode');
    $routes->post('users', 'UserController::createUser');
    $routes->post('auth', 'AuthController::authenticateUser');
    
    $routes->group(null, ['filter' => 'api-auth', 'namespace' => 'App\Controllers\V1'], function($routes) {
        $routes->put('users', 'UserController::updateUser');
    });
});
```
In essence i wanted to group some routes without a prefix but have them run under certain filters. I realized that when i hit the below end point
```
PUT /v1/users
```
I basically get a 404.

So  i made some changes to make this possible.

These changes could help one group routes under a subdomain for instance without the need for a route prefix, apply security filters on certain routes like i did above, without the need for a prefix, and so on.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -713,7 +713,7 @@ class RouteCollection implements RouteCollectionInterface
 
 		// To register a route, we'll set a flag so that our router
 		// so it will see the group name.
-		// If the group name is null, we go on using the previously built group name.
+		// If the group name is empty, we go on using the previously built group name.
 		$this->group = $name ? ltrim($oldGroup . '/' . $name, '/') : $oldGroup;
 
 		$callback = array_pop($params);
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -701,12 +701,12 @@ class RouteCollection implements RouteCollectionInterface
 	 *            $route->resource('users');
 	 *     });
 	 *
-	 * @param string|null    $name      The name to group/prefix the routes with.
+	 * @param string         $name      The name to group/prefix the routes with.
 	 * @param array|callable ...$params
 	 *
 	 * @return void
 	 */
-	public function group(?string $name, ...$params)
+	public function group(string $name, ...$params)
 	{
 		$oldGroup   = $this->group;
 		$oldOptions = $this->currentOptions;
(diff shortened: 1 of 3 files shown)


paulbalandan:
Is the same not achieved by using an empty string instead of `null`?

```php
    $routes->group('', ['filter' => 'api-auth', 'namespace' => 'App\Controllers\V1'], function($routes) {
        $routes->put('users', 'UserController::updateUser');
    });
```

francis94c:
That's what i thought at first, I tried it but it didn't work.

On investigating the `RouteCollection` class.

on line 739, i see 

```
$this->group = ltrim($oldGroup . '/' . $name, '/');
```

With an empty string, we are left with '{oldGroup}//' which won't match.

I replaced the line with
```
$this->group = $name ? ltrim($oldGroup . '/' . $name, '/') : $oldGroup;
```
to make it work...

paulbalandan:
I think it is a bug. Your added line would also work for empty strings as it will evaluate to `false`.

> I replaced the line with
> 
> ```
> $this->group = $name ? ltrim($oldGroup . '/' . $name, '/') : $oldGroup;
> ```

So maybe instead of going for a nullable string typehint, we stick to string typehint and just add that line?

I'll have the others say what they think on this. @MGatner ?

francis94c:
I'll update and push

system/Router/RouteCollection.php
```diff
@@ -713,7 +713,8 @@ public function group(string $name, ...$params)
 
 		// To register a route, we'll set a flag so that our router
 		// so it will see the group name.
-		$this->group = ltrim($oldGroup . '/' . $name, '/');
+		// If the group name is null, we go on using the previously built group name.
```
paulbalandan, line 716:
```suggestion
		// If the group name is empty, we go on using the previously built group name.
```

paulbalandan (approved):

paulbalandan:
@francis94c thank you for your first contribution

francis94c:
My pleasure @paulbalandan

PR #3919, opened 2020-11-24
PHP 8 Support

MGatner:
**Description**
This is a first pass at some of the changes that need to happen for PHP 8.0 release. One significant thing this PR does not address is the change to GD return types to use `GDImage` (see https://php.watch/versions/8.0/gdimage), with will need to be addressed in **images/Handler/GDhandler.php**.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/.github/workflows/test-phpstan.yml
+++ b/.github/workflows/test-phpstan.yml
@@ -20,8 +20,12 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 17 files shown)


MGatner:
Splitting this out into 7.2-compatible changes and other. See #3931

PR #3930, opened 2020-11-27
Split Model to have BaseModel without database related stuff

najdanovicivan:
**Description**
The goal with this change is to have BaseModel which does not contain any database related code so that it can easily be extended to support other kinds of databases like NoSQL document based

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -51,13 +51,6 @@ abstract class BaseModel
 	public $pager;
 
 	/**
-	 * Name of database table
-	 *
-	 * @var string
-	 */
-	protected $table;
-
-	/**
 	 * Last insert ID
 	 *
 	 * @var integer|string
@@ -1361,20 +1354,6 @@ abstract class BaseModel
 			default:
 				throw ModelException::forNoDateFormat(static::class);
 		}
-	}
-
-	/**
-	 * Specify the table associated with a model
-	 *
-	 * @param string $table Table
-	 *
-	 * @return $this
-	 */
-	public function setTable(string $table)
-	{
-		$this->table = $table;
-
-		return $this;
 	}
 
 	/**
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -28,17 +28,18 @@ use stdClass;
 /**
  * Class Model
  *
- * The Model class provides a number of convenient features that
- * makes working with a database table less painful.
+ * The BaseModel class provides a number of convenient features that
+ * makes working with a databases less painful. Extending this class
+ * provide means of implementing various database systems
  *
  * It will:
- *      - automatically connect to database
- *      - allow intermingling calls between db connection, the builder,
- *          and methods in this class.
  *      - simplifies pagination
- *      - removes the need to use Result object directly in most cases
  *      - allow specifying the return type (array, object, etc) with each call
+ *      - automatically set and update timestamps
+ *      - handle soft deletes
  *      - ensure validation is run against objects when saving items
+ *      - process various callbacks
+ *      - allow intermingling calls to the db connection
  */
 abstract class BaseModel
 {
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -13,7 +13,6 @@ namespace CodeIgniter;
 
 use Closure;
 use CodeIgniter\Database\BaseResult;
-use CodeIgniter\Database\ConnectionInterface;
 use CodeIgniter\Database\Exceptions\DatabaseException;
 use CodeIgniter\Database\Exceptions\DataException;
 use CodeIgniter\Exceptions\ModelException;
@@ -43,13 +42,6 @@ use stdClass;
  */
 abstract class BaseModel
 {
-	/**
-	 * The table's primary key.
-	 *
-	 * @var string
-	 */
-	protected $primaryKey = 'id';
-
 	/**
 	 * Pager instance.
 	 * Populated after calling $this->paginate()
@@ -120,8 +112,6 @@ abstract class BaseModel
 	 * Allowed: 'datetime', 'date', 'int'
 	 *
 	 * @var string
-	 *
-	 * @todo make it more generic to be used by other models
 	 */
 	protected $dateFormat = 'datetime';
 
@@ -173,9 +163,9 @@ abstract class BaseModel
 	/**
 	 * Database Connection
 	 *
-	 * @var ConnectionInterface
+	 * @var object
 	 *
-	 * @todo check if Interface can be used for NO-SQL
+	 * @todo check if ConnectionInterface can be used for NO-SQL
 	 */
 	protected $db;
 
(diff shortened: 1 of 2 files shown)


najdanovicivan:
It's almost completely split the only issue so far is how to get rid of $primaryKey property

The problematic method is classToArray as it is defined as static method and $this->primaryKey is passed on each call. 

I’d suggest we change this method not to be static. Then it can be overridden inside Model class to handle primary while we still keep the conversion in base class.  

If we still need it for backwards compatibility reasons it can always be implemented via __callStatic in the Model.

najdanovicivan:
Another thought. I've used `db` to prefix methods in Model Class. When I take a look at BaseBuilder I see that `_` is used so it might be a good idea to use it in model class as well

system/BaseModel.php
```diff
@@ -0,0 +1,1847 @@
...
+		];
+
+		if ($this->tempAllowCallbacks)
+		{
+			$this->trigger('afterUpdate', $eventData);
+		}
+
+		$this->tempAllowCallbacks = $this->allowCallbacks;
+
+		return $result;
```
mostafakhudair, line 969:
```suggestion
		$eventData = [
			'id'     => $id,
			'data'   => $eventData['data'],
			'result' => $this->dbUpdate($id, $eventData['data']),
		];

		if ($this->tempAllowCallbacks)
		{
			$this->trigger('afterUpdate', $eventData);
		}

		$this->tempAllowCallbacks = $this->allowCallbacks;

		return $eventData['result'];
```

system/BaseModel.php
```diff
@@ -0,0 +1,1847 @@
...
+		];
+
+		if ($this->tempAllowCallbacks)
+		{
+			$this->trigger('afterDelete', $eventData);
+		}
+
+		$this->tempAllowCallbacks = $this->allowCallbacks;
+
+		return $result;
```
mostafakhudair, line 1114:
```suggestion
		$eventData = [
			'id'     => $id,
			'data'   => null,
			'purge'  => $purge,
			'result' => $this->dbDelete($id, $purge),
		];

		if ($this->tempAllowCallbacks)
		{
			$this->trigger('afterDelete', $eventData);
		}

		$this->tempAllowCallbacks = $this->allowCallbacks;

		return $eventData['result'];
```

najdanovicivan:
@mostafakhudair Thanks on these

system/BaseModel.php
```diff
@@ -0,0 +1,1832 @@
...
+			]);
+
+			if (! empty($eventData['returnData']))
+			{
+				return $eventData['data'];
+			}
+		}
+
+		$eventData = [
+			'data'      => $this->dbFirst();,
```
mostafakhudair, line 503:
```suggestion
			'data'      => $this->dbFirst(),
```
My fault copy paste failure

najdanovicivan:
I went ahead and removed ClassToArray method in favour of new object method objectToArray() as this method beside conversion to array also converts the Time instances to strings. 

So I've split the method in two methods objectToRawArray which does only the conversion and then objectToArray() internally call ObjectToRawArray

This makes it easy to override it it SQL Model to always add the PrimaryKey. This makes it possible to remove the primary key now from the BaseModel.

paulbalandan (changes requested):
Can we change the abstract protected functions to not start with underscore? It was corrected with variables a while back and it was too late to do the same with methods as it will break apps. Moving forward I hope this would not be seen in new code as it adds no value.

Also, in accord with PSR12 and PSR2 please reorder `protected abstract function` to `abstract protected function`.

najdanovicivan:
@paulbalandan I've intentionally put the underscores to follow the same methodology from BaseBuilder. I don't know what's the best way to go with that. I'm completely fine with any of these. I just wanted to be more consistent with the existing framework code.

There are some CodeSniffer rules I don't like but those should be separate PR but let me just state them. I'd like to here opinions before I'm even to start working with the PR

- Private methods are fine to start with _ but protected are not. But we have existing protected methods starting with _ in the code
- Line Length I believe it's 120 characters. It's just to low limit I'd rather got with 200 or 250 
- @throws require Message which ends with dot (.) I don't think we even need forcing for the message for these
- This I'm not sure about but maybe no forcing @param message as well. There are a lot of methods missing those

paulbalandan:
> * Private methods are fine to start with _ but protected are not. But we have existing protected methods starting with _ in the code

I think it was mentioned before that no methods should start with `_` except for `_remap` to retain BC with CI3. The practice of prepending underscores to imply non-public visibility was a remnant from CI3. Since it is too late to rename such other methods, those were left as-is.

> * Line Length I believe it's 120 characters. It's just to low limit I'd rather got with 200 or 250

I'm not aware that codesniffer enforces line length but I do not object with yours either. After all, PSR12 does not enforce a hard limit on line length.

> * @throws require Message which ends with dot (.) I don't think we even need forcing for the message for these

I'm not aware of this either. But grammatically speaking this should be the case since error messages are sentences.

> * This I'm not sure about but maybe no forcing @param message as well. There are a lot of methods missing those

I'm not sure what you meant by this if either `@param string $name` or `@param string $name the name`. Either way `@param` and `@return` docblocks are needed even though appearing as unneeded because the API generator needs them.

paulbalandan:
My suggestion for replacement of the non-public methods is this:

From:
```php
public function find();
protected function _find();
```

To:
```php
public function find();
protected function doFind();
```

najdanovicivan:
@paulbalandan regarding CodeSniffer rules

@param string $name **comment**

This is required and have to start with capital letter @param does not enforce the dot at the end

@throws Exception **comment** 

This is required and have to start with capital letter also have to end with dot

What does not make sense is that param comment does not require dot while throws comment requires it

najdanovicivan:
So the issue is due to these 

We have these in ruleset.xml
```
    <rule ref="Squiz.Commenting.FunctionComment.ParamCommentFullStop">
        <severity>0</severity>
    </rule>
```

But we're missing
```
    <rule ref="Squiz.Commenting.FunctionComment.ThrowsNoFullStop">
        <severity>0</severity>
    </rule>
```

So we should either require full stop at the end for both or not require it anywhere

paulbalandan:
I'm not really that well-versed with codesniffer but here are my suggestions:

1. `@param string $name comment` - `comment` should not start with a capital and neither stop with a period
2. `@throws Exception comment` - comment is optional and need not to be enforced for all. But if ever it appears there, it should follow the same rules with `@param`
3. `@throws` as well as other phpdocs should be explicitly provided although it is superfluous because PHPDocumentor requires it to generate our API documentation. Without these, the compilation will error.

If there are corresponding rulesets for what I mentioned, please open a PR to the `coding-standards` repo.

najdanovicivan:
> * `@param string $name comment` - `comment` should not start with a capital and neither stop with a period

@paulbalandan there are 2 different options here

By should not does it mean if it starts with capital letter and ends with dot it should be thrown as a warning or it does not matter and the rule enforcing this should be disabled ?

paulbalandan:
> > * `@param string $name comment` - `comment` should not start with a capital and neither stop with a period
> 
> @paulbalandan there are 2 different options here
> 
> By should not does it mean if it starts with capital letter and ends with dot it should be thrown as a warning or it does not matter and the rule enforcing this should be disabled ?

If there are rules that enforce all small letters and no period then those should be used instead. If none, then current rules should be disabled.

system/BaseModel.php
```diff
@@ -0,0 +1,1825 @@
...
+ * makes working with a database table less painful.
+ *
+ * It will:
+ *      - automatically connect to database
+ *      - allow intermingling calls between db connection, the builder,
+ *          and methods in this class.
+ *      - simplifies pagination
+ *      - removes the need to use Result object directly in most cases
+ *      - allow specifying the return type (array, object, etc) with each call
+ *      - ensure validation is run against objects when saving items
```
paulbalandan, line 41: I think this docblock needs to be revised since this `BaseModel` is DB-agnostic?

najdanovicivan: Sure. A lot of comments have to be reworked.
 
I also noticed something else. 
$this->table is not used anywhere in BaseModel now so it should be moved to Model as well. And remove $this->table from all the comments

Also with all this. I'm noticing now how simple it will be to add support for Composite Keys. I'm not sure if this is to be implemented with this PR or should I make additional one for such purpose 

```
	protected $primaryKey = ['composite1','composite2'];
```

Then find can be called like this

```
      $model->find(['composite1_value', 'composite2_value'])
```

paulbalandan: Composite primary keys should be for another PR and discussion. 👍

najdanovicivan:
https://github.com/codeigniter4/coding-standard/pull/31 I've submitted PR for the CodeSniffer Rules

MGatner:
I do think this is really good and important but I will need to look at it more closely. I have a NoSQL model stub I would like to try putting on `BaseModel` to test it out. https://github.com/tattersoftware/codeigniter4-firebase/blob/develop/src/Model.php

paulbalandan (commented):
I'm okay with this but I'll wait first for @MGatner 's review.

MGatner:
I have reviewed the code but I still have not had a chance to test this with a NoSQL instance or my own custom Model class. We want these deprecations to post to `4.1` and tests are all passing so I will approve this PR, but please note that this split may require some more work down the road.

MGatner (approved):

MGatner:
@paulbalandan I think you've finished reviewing but your Review has not been updated. I will give it a day.

paulbalandan (approved):

najdanovicivan:
@MGatner I'm mostly certain that it will require some additional work down the road. But this should be enough to get it going. I'm already having idea of doing a further split for BaseBuilderModel which will come useful when additional drivers are implemented using builder class

So in the end we'll have something like  BaseModel->BaseBuilderModel->Model but everything is still raw with all of these. I already have some PRs planed for BaseConnection, QueryInterface and Database toolbar collector but I intend on submitting those one we have MongoDB driver ready to be publicly released so that everything makes sense and it is possible to do unit tests

PR #3931, opened 2020-11-27
Prep for PHP 8

MGatner:
**Description**
Taking a different approach than https://github.com/codeigniter4/CodeIgniter4/pull/3919, this PR includes changes that are generally good for the framework on any version of PHP, but will be required with PHP 8. One addendum: I've excluded `GDHandler` from static analysis since PHP 8 uses a new class type instead of resources which works fine with the current code but not fully-typed analysis.

Once this is all passing it should be merged and used as the new base for `4.1` which includes anything non-7.2 friendly.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -24,6 +24,7 @@ parameters:
 		- system/Commands/Generators/Views/*
 		- system/Config/Routes.php
 		- system/Debug/Toolbar/Views/toolbar.tpl.php
+		- system/Images/Handlers/GDHandler.php
 		- system/ThirdParty/*
 		- system/Validation/Views/single.php
 	ignoreErrors:
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -416,6 +416,11 @@ class CodeIgniter
 		{
 			$controller = $this->createController();
 
+			if (! method_exists($controller, '_remap') && ! is_callable([$controller, $this->method], false))
+			{
+				throw PageNotFoundException::forMethodNotFound($this->method);
+			}
+
 			// Is there a "post_controller_constructor" event?
 			Events::trigger('post_controller_constructor');
 
@@ -889,11 +894,6 @@ class CodeIgniter
 		if (! class_exists($this->controller, true) || $this->method[0] === '_')
 		{
 			throw PageNotFoundException::forControllerNotFound($this->controller, $this->method);
-		}
-		if (! method_exists($this->controller, '_remap') &&
-				! is_callable([$this->controller, $this->method], false))
-		{
-			throw PageNotFoundException::forMethodNotFound($this->method);
 		}
 	}
 
--- a/system/Database/Database.php
+++ b/system/Database/Database.php
@@ -41,7 +41,7 @@ class Database
 	 * @return   mixed
 	 * @internal param bool $useBuilder
 	 */
-	public function load(array $params = [], string $alias)
+	public function load(array $params, string $alias)
 	{
 		// Handle universal DSN connection string
 		if (! empty($params['DSN']) && strpos($params['DSN'], '://') !== false)
--- a/system/Database/Database.php
+++ b/system/Database/Database.php
@@ -41,8 +41,13 @@ class Database
 	 * @return   mixed
 	 * @internal param bool $useBuilder
 	 */
-	public function load(array $params, string $alias)
+	public function load(array $params = [], string $alias = '')
 	{
+		if (empty($alias))
+		{
+			throw new InvalidArgumentException('You must supply the parameter: alias.');
+		}
+
 		// Handle universal DSN connection string
 		if (! empty($params['DSN']) && strpos($params['DSN'], '://') !== false)
 		{
--- a/system/Database/Database.php
+++ b/system/Database/Database.php
@@ -41,13 +41,8 @@ class Database
 	 * @return   mixed
 	 * @internal param bool $useBuilder
 	 */
-	public function load(array $params = [], string $alias = '')
+	public function load(array $params = [], string $alias)
 	{
-		if (empty($alias))
-		{
-			throw new InvalidArgumentException('You must supply the parameter: alias.');
-		}
-
 		// Handle universal DSN connection string
 		if (! empty($params['DSN']) && strpos($params['DSN'], '://') !== false)
 		{
--- a/system/Database/MySQLi/Connection.php
+++ b/system/Database/MySQLi/Connection.php
@@ -134,7 +134,7 @@ class Connection extends BaseConnection
 					if ($this->encrypt['ssl_verify'])
 					{
 						defined('MYSQLI_OPT_SSL_VERIFY_SERVER_CERT') &&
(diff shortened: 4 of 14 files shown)


system/CodeIgniter.php
```diff
@@ -890,11 +895,6 @@ protected function startController()
 		{
 			throw PageNotFoundException::forControllerNotFound($this->controller, $this->method);
 		}
-		if (! method_exists($this->controller, '_remap') &&
-				! is_callable([$this->controller, $this->method], false))
```
MGatner, line 894: PHP 8 changes the behavior of `is_callable()` such that we need to pass it an instantiated object. Since we are about to get the controller instance anyways I moved this check after.

> The ability to call non-static methods statically has been removed. Thus is_callable() will fail when checking for a non-static method with a classname (must check with an object instance).

https://www.php.net/manual/en/migration80.incompatible.php

phpstan.neon.dist
```diff
@@ -24,6 +24,7 @@ parameters:
 		- system/Commands/Generators/Views/*
 		- system/Config/Routes.php
 		- system/Debug/Toolbar/Views/toolbar.tpl.php
+		- system/Images/Handlers/GDHandler.php
```
MGatner, line 27: PHP 8 uses `GDImage` class type for the `image*` functions, which works fine with our current code but breaks the analysis. Ignoring the only file with these references until it can be reworked to support all versions or until PHPStan adds version-specific exemption handling.

system/Database/Database.php
```diff
@@ -41,7 +41,7 @@ class Database
 	 * @return   mixed
 	 * @internal param bool $useBuilder
 	 */
-	public function load(array $params = [], string $alias)
```
MGatner, line 44: This (and the majority of these changes) are to correct the improper syntax of a required variable following an optimal variable, a big "no no" in PHP 8.

system/Database/MySQLi/Connection.php
```diff
@@ -134,7 +134,7 @@ public function connect(bool $persistent = false)
 					if ($this->encrypt['ssl_verify'])
 					{
 						defined('MYSQLI_OPT_SSL_VERIFY_SERVER_CERT') &&
-						$this->mysqli->options(MYSQLI_OPT_SSL_VERIFY_SERVER_CERT, true);
```
MGatner, line 137: Another subset of changes are for using booleans for `bool|int` in places where PHP 8 expects only `int`.

system/Test/ReflectionHelper.php
```diff
@@ -97,6 +97,6 @@ public static function setPrivateProperty($obj, $property, $value)
 	public static function getPrivateProperty($obj, $property)
 	{
 		$refProperty = self::getAccessibleRefProperty($obj, $property);
-		return $refProperty->getValue($obj);
+		return is_string($obj) ? $refProperty->getValue() : $refProperty->getValue($obj);
```
MGatner, line 100: I'm not sure how this was working previously, but in PHP 8 `ReflectionProperty::getValue()` can *only* take an object parameter. Maybe an undocumented "feature" in 7.x that you could pass a string?

system/Database/Database.php
lonnieezell: They've actually started enforcing that at the language level? Wow. I need to find time to read up on PHP8 changes very soon!

MGatner: It's a depreciation notice:
```
PHP Deprecated:  Required parameter $from follows optional parameter $verbs in /home/runner/work/CodeIgniter4/CodeIgniter4/system/Router/RouteCollection.php on line 991
9
```

lonnieezell (approved):
As long as these are the basis for 4.1 I have no problems with these.

paulbalandan (commented):
I'm all for these changes if these will be necessary for the framework to be PHP8 compatible.

My question is this: I'm not that too familiar with LSP. Will the addition/removal of defaults in optional params be a method signature violation if someone extends these methods?

MGatner:
> Will the addition/removal of defaults in optional params be a method signature violation if someone extends these methods?

Yes, if someone has extended the method and duplicated the deprecated format it will break. I don't think there is a way around this but breaking class extensions seems much better than implementations.

lonnieezell (commented):
I think the revision works well. As you said on Slack, some unfortunate extra code, but it solves the problem and doesn't break BC. Thumbs up from me.

system/Validation/Validation.php
```diff
@@ -206,8 +206,13 @@ public function check($value, string $rule, array $errors = []): bool
 	 *
 	 * @return boolean
 	 */
-	protected function processRules(string $field, string $label = null, $value, $rules = null, array $data): bool
+	protected function processRules(string $field, string $label = null, $value, $rules = null, array $data = null): bool
 	{
+		if (is_null($data))
```
mostafakhudair, line 211:
```suggestion
	protected function processRules(string $field, string $label = null, $value, $rules = null, array $data = []): bool
	{
		if (empty($data))
```

phpstan.neon.dist
paulbalandan: Maybe we can use stub files for this? I'm not sure if there is a capability in phpstan to conditionally include stub files based on the current PHP version.

system/CodeIgniter.php
paulbalandan: Related to testing, I think we should change how we invoked phpunit's `assert*` methods. It is by design a static method and phpunit allows calling it non statically. However when I used phpstan's extra strict rules, I get 'Dynamic calls to static method assert** is not allowed'.

system/Validation/Validation.php
MGatner: It has to accept an empty array which is why I added the `null` type. I would definitely have preferred `empty()`.

MGatner:
To @paulbalandan 's point: should the parameter changes be moved to the section of work only applying to 4.1, since it is potentially breaking for class extensions and not entirely necessary for 4.0.5?

system/CodeIgniter.php
MGatner: We're doing it "the PHPUnit way": https://phpunit.readthedocs.io/en/9.3/assertions.html

I didn't even realize they were static. Is there some localized forwarder or something?

MGatner:
> should the parameter changes be moved to the section of work only applying to 4.1

I decided this was the case. If `4.0.5` is only ever used for 7.2 then there will be no problem with these parameters. This PR will retain changes that are good to have for both releases, and I created #3938 to handle the PHP 8 deprecation.

system/CodeIgniter.php
paulbalandan: I've looked into the code of TestCase and Assert and didn't find some wrapper to dynamic calls. Maybe phpunit is taking advantage of the way static methods can be called dynamically in php lt 8.

paulbalandan (approved):

MGatner:
Yet another Postgres failure. We'll need to look at that soon. 

Rest looks good, as expected since this is now a subset. Thanks for the approval @paulbalandan

PR #3935, opened 2020-11-28
Deprecate Message::getHeader(s)

MGatner:
**Description**
In the interest of moving towards partial PSR-7 compliance this PR adds the `header()` and `headers()` methods to `Message` and deprecates the current `getHeader()` and `getHeaders()` methods which conflict with the PSR's interface.

Note that I've left tests pointing to the original methods since they call the new ones and thus should not adversely affect code coverage.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -657,7 +657,7 @@ class CodeIgniter
 			$output  = $cachedResponse['output'];
 
 			// Clear all default headers
-			foreach ($this->response->getHeaders() as $key => $val)
+			foreach ($this->response->headers() as $key => $val)
 			{
 				$this->response->removeHeader($key);
 			}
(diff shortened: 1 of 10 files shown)


MGatner:
Postgres fails are unrelated. IMO this is ready, requesting reviews.

lonnieezell (approved):
Looks good. My only concern is if we should make a bigger deal somewhere of it being deprecated, but I'm not sure where. And looking at the current 4.0.5 changelog we already have several deprecations there.

paulbalandan (approved):

MGatner:
I used to love CI3's "Upgrade Guide". We have the stub for it in 4 (https://codeigniter.com/user_guide/installation/upgrading.html) and we should probably start filling it in for 4.1 and on, noting deprecations etc. I will start something before merging this.

PR #3936, opened 2020-11-28
Alternative HTTP (PSR-7)

MGatner:
**Description**
This PR supplies PSR-7 compatible variants of the framework HTTP layer. Along with #3935 this lays the groundwork for moving the HTTP layer to full PSR-7 compliance which would allow CodeIgniter to integrate with other message-dependent libraries without the need to install additional packages (Guzzle).

This is an alternate implementation route to #3934, but see additional notes there. 

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


lonnieezell:
I think this is the better of the two options. My biggest concern is the namespace. I feel it still needs to be within the HTTP module in the `system`, if only to keep things organized cleanly and not have part of the layer somewhere else, adding confusion. We could have `app\HTTP \PSR7`, but I wonder how confusing that might make the confusion? Probably not too bad, I don't think. Thoughts?

MGatner:
I realize `app` in your response was an accident, but it got me thinking: what if we solved this in App so upgrading was just a user toggle that could (later) be removed? Here's my idea: I've almost submitted a PR a few times for a Class Aliaser on boot. We go ahead and add aliasing to the boot process and make these new classes dependent on some condition. That way we can stick them wherever (system/HTTP/PSR7 is fine) and a) still use the same namespace b) not be committed to nested, duplicate code long-term.

composer.json
```diff
@@ -12,6 +12,7 @@
 		"ext-mbstring": "*",
 		"kint-php/kint": "^3.3",
 		"laminas/laminas-escaper": "^2.6",
+		"psr/http-message": "^1.0",
```
paulbalandan, line 15: With this PSR included in the main `require`, we should also include the library in `system/ThirdParty` for non-Composer users. Then have `ComposerScripts` updated to update this package as well,

paulbalandan (commented):
I agree this is the cleaner alternative. My objection only is the same with Lonnie on the use of the namespace. I believe too it should be under `HTTP`.

MGatner:
@samsonasik or @paulbalandan If either of you have any PHPStan genius, it does not like the new Autoloader class aliasing:
```
------ ----------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   system/HTTP/CLIRequest.php                                                                                                                                 
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------- 
         Internal error: Internal error: Expected to find an ancestor with class name CodeIgniter\HTTP\Request on CodeIgniter\HTTP\CLIRequest, but none was found.  
         Run PHPStan with --debug option and post the stack trace to:                                                                                               
         https://github.com/phpstan/phpstan/issues/new                                                                                                              
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

It can't seem to find the aliased class even though a) the Autoloader creates the alias, b) the original class file is still in place.

I tried providing it with the Autoloader directly (https://phpstan.org/user-guide/command-line-usage#--autoload-file%7C-a) even though we bootstrap it, and I tried creating an addition bootstrap file to alias the classes explicitly prior to analysis (https://phpstan.org/user-guide/discovering-symbols#class-aliases) but neither affected these errors.

Any ideas?

system/CodeIgniter.php
```diff
@@ -747,10 +747,10 @@ protected function generateCacheName(Cache $config): string
 	{
 		if (get_class($this->request) === CLIRequest::class)
```
samsonasik, line 748: how about : 

```suggestion
              if ($this->request instanceof CLIRequest)
```

composer.json
MGatner: I injected the files and added them to the internal class map, but I don't see any `Psr\Log` handling in **ComposerScripts** so I didn't do anything there. If I missed something please let me know.

MGatner:
I reported the PHPStan issue: https://github.com/phpstan/phpstan/issues/4144

paulbalandan (approved):

lonnieezell:
@MGatner should we close this?

MGatner:
Yes this one was based on the Autoloader aliasing, can be gonzo.

PR #3938, opened 2020-11-28
Backfill Parameters

MGatner:
**Description**
This PR addresses the now-deprecated(PHP 8) use of mandatory parameters following optional parameters. Split off from #3931.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/Database.php
+++ b/system/Database/Database.php
@@ -41,8 +41,13 @@ class Database
 	 * @return   mixed
 	 * @internal param bool $useBuilder
 	 */
-	public function load(array $params = [], string $alias)
+	public function load(array $params = [], string $alias = '')
 	{
+		if (empty($alias))
+		{
+			throw new InvalidArgumentException('You must supply the parameter: alias.');
+		}
+
 		// Handle universal DSN connection string
 		if (! empty($params['DSN']) && strpos($params['DSN'], '://') !== false)
 		{
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -16,6 +16,7 @@ use CodeIgniter\HTTP\Files\FileCollection;
 use CodeIgniter\HTTP\Files\UploadedFile;
 use Config\App;
 use Config\Services;
+use InvalidArgumentException;
 use Locale;
 
 /**
@@ -128,8 +129,13 @@ class IncomingRequest extends Request
 	 * @param string|null $body
 	 * @param UserAgent   $userAgent
 	 */
-	public function __construct($config, URI $uri = null, $body = 'php://input', UserAgent $userAgent)
+	public function __construct($config, URI $uri = null, $body = 'php://input', UserAgent $userAgent = null)
 	{
+		if (empty($uri) || empty($userAgent))
+		{
+			throw new InvalidArgumentException('You must supply the parameters: uri, userAgent.');
+		}
+
 		// Get our body from php://input
 		if ($body === 'php://input')
 		{
--- a/system/Images/Handlers/BaseHandler.php
+++ b/system/Images/Handlers/BaseHandler.php
@@ -15,6 +15,7 @@ use CodeIgniter\Images\Exceptions\ImageException;
 use CodeIgniter\Images\Image;
 use CodeIgniter\Images\ImageHandlerInterface;
 use Config\Images;
+use InvalidArgumentException;
 
 /**
  * Base image handling implementation
@@ -662,8 +663,13 @@ abstract class BaseHandler implements ImageHandlerInterface
 	 *
 	 * @return array
 	 */
-	protected function calcAspectRatio($width, $height = null, $origWidth, $origHeight): array
+	protected function calcAspectRatio($width, $height = null, $origWidth = 0, $origHeight = 0): array
 	{
+		if (empty($origWidth) || empty($origHeight))
+		{
+			throw new InvalidArgumentException('You must supply the parameters: origWidth, origHeight.');
+		}
+
 		// If $height is null, then we have it easy.
 		// Calc based on full image size and be done.
 		if (is_null($height))
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -12,6 +12,7 @@
 namespace CodeIgniter\Validation;
 
 use Config\Database;
+use InvalidArgumentException;
 
 /**
  * Validation Rules.
@@ -352,13 +353,18 @@ class Rules
 	 *     required_with[password]
 	 *
 	 * @param string|null $str
-	 * @param string      $fields List of fields that we should check if present
+	 * @param string|null $fields List of fields that we should check if present
 	 * @param array       $data   Complete list of fields from the form
 	 *
 	 * @return boolean
 	 */
-	public function required_with($str = null, string $fields, array $data): bool
+	public function required_with($str = null, string $fields = null, array $data = []): bool
 	{
+		if (is_null($fields) || empty($data))
+		{
+			throw new InvalidArgumentException('You must supply the parameters: fields, data.');
+		}
+
 		$fields = explode(',', $fields);
 
 		// If the field is present we can safely assume that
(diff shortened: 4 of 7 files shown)


MGatner:
Rebasing 4.1 and trying again.

MGatner:
Looks good! @paulbalandan you still around by chance?

system/Database/Database.php
```diff
@@ -41,8 +41,13 @@ class Database
 	 * @return   mixed
 	 * @internal param bool $useBuilder
 	 */
-	public function load(array $params = [], string $alias)
+	public function load(array $params = [], string $alias = '')
 	{
+		if (empty($alias))
```
paulbalandan, line 46:
```suggestion
		if ($alias === '')
```

system/Database/Database.php
```diff
@@ -41,8 +41,13 @@ class Database
...
+		{
+			throw new InvalidArgumentException('You must supply the parameter: alias.');
```
paulbalandan, line 48:
```suggestion
			throw new InvalidArgumentException('You must supply the parameter: alias.'); // @codeCoverageIgnore
```

system/HTTP/IncomingRequest.php
```diff
@@ -128,8 +129,13 @@ class IncomingRequest extends Request
 	 * @param string|null $body
 	 * @param UserAgent   $userAgent
 	 */
-	public function __construct($config, URI $uri = null, $body = 'php://input', UserAgent $userAgent)
+	public function __construct($config, URI $uri = null, $body = 'php://input', UserAgent $userAgent = null)
 	{
+		if (empty($uri) || empty($userAgent))
+		{
+			throw new InvalidArgumentException('You must supply the parameters: uri, userAgent.');
+		}
```
paulbalandan, line 137:
```suggestion
		if ($uri === null || $userAgent === null)
		{
			throw new InvalidArgumentException('You must supply the parameters: uri, userAgent.'); // @codeCoverageIgnore
		}
```

MGatner:
Since this was approved in the other PR I am going to proceed with the merge.

system/Images/Handlers/BaseHandler.php
```diff
@@ -662,8 +663,13 @@ public function fit(int $width, int $height = null, string $position = 'center')
 	 *
 	 * @return array
 	 */
-	protected function calcAspectRatio($width, $height = null, $origWidth, $origHeight): array
+	protected function calcAspectRatio($width, $height = null, $origWidth = 0, $origHeight = 0): array
 	{
+		if (empty($origWidth) || empty($origHeight))
+		{
+			throw new InvalidArgumentException('You must supply the parameters: origWidth, origHeight.');
```
paulbalandan, line 670:
```suggestion
			throw new InvalidArgumentException('You must supply the parameters: origWidth, origHeight.'); // @codeCoverageIgnore
```

system/Validation/Rules.php
```diff
@@ -352,13 +353,18 @@ public function required($str = null): bool
 	 *     required_with[password]
 	 *
 	 * @param string|null $str
-	 * @param string      $fields List of fields that we should check if present
+	 * @param string|null $fields List of fields that we should check if present
```
paulbalandan, line 356:
```suggestion
	 * @param string $fields List of fields that we should check if present
```

system/Validation/Rules.php
```diff
@@ -352,13 +353,18 @@ public function required($str = null): bool
...
 	 *
 	 * @return boolean
 	 */
-	public function required_with($str = null, string $fields, array $data): bool
+	public function required_with($str = null, string $fields = null, array $data = []): bool
 	{
+		if (is_null($fields) || empty($data))
+		{
+			throw new InvalidArgumentException('You must supply the parameters: fields, data.');
+		}
```
paulbalandan, line 366:
```suggestion
	public function required_with($str = null, string $fields = '', array $data = []): bool
	{
		if ($fields === '' || $data === [])
		{
			throw new InvalidArgumentException('You must supply the parameters: fields, data.'); // @codeCoverageIgnore
		}
```

system/Validation/Rules.php
```diff
@@ -404,13 +410,18 @@ public function required_with($str = null, string $fields, array $data): bool
...
 	 * @return boolean
 	 */
-	public function required_without($str = null, string $fields, array $data): bool
+	public function required_without($str = null, string $fields = null, array $data = []): bool
 	{
+		if (is_null($fields) || empty($data))
+		{
+			throw new InvalidArgumentException('You must supply the parameters: fields, data.');
+		}
+
```
paulbalandan, line 424: Same comment with `required_with`

system/Validation/Validation.php
```diff
@@ -206,8 +206,13 @@ public function check($value, string $rule, array $errors = []): bool
...
 	 * @return boolean
 	 */
-	protected function processRules(string $field, string $label = null, $value, $rules = null, array $data): bool
+	protected function processRules(string $field, string $label = null, $value, $rules = null, array $data = null): bool
 	{
+		if (is_null($data))
+		{
+			throw new InvalidArgumentException('You must supply the parameter: data.');
+		}
+
```
paulbalandan, line 215:
```suggestion
	protected function processRules(string $field, string $label = null, $value, $rules = null, array $data = []): bool
	{
		if ($data === [])
		{
			throw new InvalidArgumentException('You must supply the parameter: data.'); // @codeCoverageIgnore
		}

```

paulbalandan (changes requested):
I was influenced by PHPStan's extra strict rules in that `empty()` implies weak comparison and we should strive to be more specific what 'emptiness' we are checking. We already have a rector rule that changes `empty` checks on arrays to `=== []` and I believe we should do the same for scalar types.

paulbalandan:
> Since this was approved in the other PR I am going to proceed with the merge.

Oh man! I was a little too late. 😂

MGatner:
Ah sorry! Those are good changes. I am working on the class aliases but I'll get these done up later today or tomorrow.

system/Validation/Validation.php
MGatner: This one has to add `null` as an input option because `[]` is actually valid input and we're trying to catch cases where the parameter was not supplied at all.

system/Validation/Rules.php
MGatner: Same as below, `$fields` is allowed to be an empty string so we need to add the `null` option. This could probably be addressed on calling classes but it is also possible that someone is calling this directly (though I really hope not).

sfadschm:
Did this PR also cover nullable optional parameters?
The link supplied in #3957 indicates that such definitions (while still being wrong) do not trigger the PHP8.0 deprecation message.

MGatner:
What is wrong with nullable optional parameters?

This PR dealt with the deprecated use of mandatory parameters following optional ones.

sfadschm:
Sorry for the confusion, I will try to elaborate:
Required arguments should never be put after optional arguments, or PHP8 will throw a deprecation message. This has been dealt with in this PR.

However, according to the link provided in the referenced issue, PHP8 will not show such message in the following case (typed optional argument defaulting to null):
`function test(string $param1 = null, string $param2)`

As far as I understand the deprecation, this type of argument list would still be incorrect, although it does not trigger the deprecation message.

Therefore I was asking whether the selection of functions targeted by this PR was based on the PHP8 deprecation messages or if all functions with optional parameters before required ones have been changed. 
In the first case, this PR might have missed some of the functions with argument lists like shown in the example.

Hope this makes it clearer :-)

MGatner:
While I would like to see that category changed as well, the focus of this PR was *actual* deprecation notices as they arose during PHP 8 testing. I am unhappy that these parameter inputs exist to begin with, but without breaking the methods for all users we can't really change that in a clean way.

If you identify any of the above and want to submit a non-breaking PR we could certainly resolve. I'll be honest I didn't know that was a separate category!

PR #3940, opened 2020-11-28
Autoloading Aliases

MGatner:
**Description**
This PR adds a new `Autoloader` property to the config file: `$aliases`. Aliasing will allow the autoloader to intercept a class request and return an aliased version instead on-the-fly. This will make it easier to transition framework classes with user intervention as well as to replace core classes with user-defined versions.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -29,15 +29,18 @@ use InvalidArgumentException;
  *                  Quux.php    # Foo\Bar\Qux\Quux
  *
  * you can add the path to the configuration array that is passed in the constructor.
- * The Config array consists of 2 primary keys, both of which are associative arrays:
- * 'psr4', and 'classmap'.
+ * The Config array consists of up to 3 primary keys, all of which are associative arrays:
+ * 'psr4', 'classmap', and 'aliases'.
  *
  *      $Config = [
  *          'psr4' => [
- *              'Foo\Bar'   => '/path/to/packages/foo-bar'
+ *              'Foo\Bar'   => '/path/to/packages/foo-bar',
  *          ],
  *          'classmap' => [
- *              'MyClass'   => '/path/to/class/file.php'
+ *              'MyClass'   => '/path/to/class/file.php',
+ *          ]
+ *          'aliases' => [
+ *              'Original'  => 'Alias',
  *          ]
  *      ];
  *
@@ -67,6 +70,13 @@ class Autoloader
 	 */
 	protected $classmap = [];
 
+	/**
+	 * Stores alias as key, original as value.
+	 *
+	 * @var array<string, string>
+	 */
+	protected $aliases = [];
+
 	//--------------------------------------------------------------------
 
 	/**
@@ -95,6 +105,11 @@ class Autoloader
 		if (isset($config->classmap))
 		{
 			$this->classmap = $config->classmap;
+		}
+
+		if (isset($config->aliases))
+		{
+			$this->aliases = $config->aliases;
 		}
 
 		// Should we load through Composer's namespaces, also?
@@ -221,7 +236,7 @@ class Autoloader
 	//--------------------------------------------------------------------
 
 	/**
-	 * Loads the class file for a given class name.
+	 * Loads the class file for a given class name or its alias.
 	 *
 	 * @param string $class The fully qualified class name.
 	 *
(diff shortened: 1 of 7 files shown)


system/Autoloader/Autoloader.php
```diff
@@ -29,15 +29,18 @@
...
  *
  *      $Config = [
  *          'psr4' => [
- *              'Foo\Bar'   => '/path/to/packages/foo-bar'
+ *              'Foo\Bar'   => '/path/to/packages/foo-bar',
  *          ],
  *          'classmap' => [
- *              'MyClass'   => '/path/to/class/file.php'
+ *              'MyClass'   => '/path/to/class/file.php',
+ *          ]
```
paulbalandan, line 41:
```suggestion
 *          ],
```

system/Autoloader/Autoloader.php
```diff
@@ -295,45 +316,6 @@ protected function loadInNamespace(string $class)
 
 	//--------------------------------------------------------------------
 
-	/**
```
paulbalandan, line 298: Removing loadLegacy already has a commit in the 4.1 branch.

MGatner: Hmm maybe this should go to 4.1 then? What do you think?

paulbalandan: Yes. Retain this method for now.

MGatner: I see it now. I'm confused why that went to `4.1` because I think we want it in `4.0.5` as well. maybe that was before we decided to release both? I will try cherry-picking it along with this PR.

MGatner: Actually that's going to make a mess of merging the release notes. I will remove the test for now since it is failing and unnecessary.

MGatner: I think I see what happened. I must have opened **Autoloader.php** while I was still on `4.1` and accidentally started working off that. I will restore it and apply to the commit to the correct base version, which should rebase without too much effort into `4.1`.

MGatner:
Okay got the branch issue cleared up, ready for re-review!

lonnieezell:
I understand why you want to do this, especially with the PSR7 stuff coming up. I guess I'm a little concerned this muddies up when to use this vs using a service. This also wouldn't make services nearly as explicit since you may call one class in the service, but it gets implicitly aliased in a config file. To me this adds a bit of confusion to the framework. 

I'm curious to hear other's thoughts.

MGatner:
I would like to hear others' thoughts as well. I wouldn't consider this ideal but it provides two things that we don't have currently:
1. A way to transition classes that need breaking changes (frequently needed)
2. A way to replace core classes (frequently requested)

Services can kind of solve the first problem but then you're stuck with the new version in App, which works if you've transitioned to a new namespace but not really if you've tried to do an in-place swap.

MGatner:
Two alternatives that I see:
1. We control aliases internally and make them conditional on a Config setting. This way it isn't "any class can be any class" but there are a few preset replacements we know are a possibility
2. We loosen our major version restrictions and start building 5. I know it sucks but the number of BC issues that come up every week could really start piling up to a new version quickly.

paulbalandan:
My thoughts on this:

1) Mark this feature as `@internal` and to be used only for the transitory updating of classes in compliance with PSR7. After the transition is complete, remove this feature with no promise of BC for those who used it.

--or--

2) Have a set of classes allowed to use this feature so that we control misuse and abuse. Yes, a Config setting would be nice but something that is unreachable from users' tampering.

MGatner:
One more thing to think about: currently there is no way to replace system-defined services. I assume this is a mistake/bug, but it adds another layer of difficulty solved by aliasing.

tangix:
I second that aliasing would add a layer of confusion. Personally I use Services all over the place and love the simplicity.

Regarding overriding framework classes, in my 4.0.4 project I have "fixed" the Redis TTL bug by creating my own RedisHandler and adding my own class to ``$validHandlers`` in ``app/Config/Cache.php``. This allowed me to fix the problem without messing in ``vendor`` directory. Happy camper :-)

MGatner:
Thanks all for the feedback. It sounds like we want to stick with our current tools, and avoid adding layers of complexity that something like this aliasing introduces. Closing.

Issue #3230, opened 2020-07-05
HTTP message interfaces & factories (PSR-7/17)


MGatner:
At some point we should make sure we are implementing these compliance definitions:

* https://www.php-fig.org/psr/psr-7/
* https://www.php-fig.org/psr/psr-17/

agungsugiarto:
Suggested using
• https://github.com/laminas/laminas-diactoros
• https://github.com/slimphp/Slim-Psr7

Tada5hi:
What is the status about this @MGatner ? Are you working on it or should i do a pr? 😇✌️

agungsugiarto:
ping @MGatner

MGatner:
I'm not actively working on this. I started a couple weeks back and determined it should not be too much work.

Tada5hi:
so @MGatner you will catch it up later or should i just go ahead?

MGatner:
If you can implement it then please go ahead and do so! I will not be working on this for a while.

Tada5hi:
@MGatner thx, i think and hope so 😄

MGatner:
@Tada5hi No pressure but I'm gathering some info for 4.1 - do you think this will be done this week?

asrofie:
Hi Team,
Anyone still working on this ?
If no one, I want to try handle it.

MGatner:
Nobody is working on it actively, please do! Familiarize yourself with the existing PRs and the conversations on those. The tricky part, if I recall, is that our version of Message has `getHeader()` and `getHeaders()` which do not comply with PSR-7 (because they return `Header`). However since then I *think* we figured out that we could make `Header` stringable and still comply. 

Regardless pay attention that no breaking changes are made, and leverage those HTTP Integration tests (so you shouldn't have to write many!).

MGatner:
Also, if you determine that this won't fit in the core framework you can compare what I did with PSR Cache and Simple Cache: https://github.com/codeigniter4/cache

We could definitely add HTTP as a separate module

asrofie:
Oke, I will create implementation first.

lonnieezell:
Very stale. Closing.

PR #3941, opened 2020-11-29
PSR-7 Implementation

MGatner:
**Description**
This is a PR for the future. I have been trying to force the PSR-7 transition but I think it is not time yet. The deprecation of the current `Message::getHeader()` and `Message::getHeaders()` is the important piece to have ready by `4.1` and those are merged.

This still has some work to be done but should be a decent basis. Pay attention to the new **tests/system/HTTP/PSR7** which integrates `Http\Psr7Test\Tests`, a third-party development library for PSR-7 compliance. These test results will outline the remaining work to be done to make the current classes compliant. There are additional classes that haven't started which should refer to https://github.com/php-fig/http-message/tree/master/src.

One more note: I started with the current HTTP classes as a basis, but whoever works on this down the road might consider starting with an exiting PSR-7 implementation and work on getting it CI4-compliant instead. In the short-term I will be working on tightening the definitions for our HTTP interfaces so this might become a much more realistic process.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


lonnieezell:
@MGatner should we close this?

MGatner:
I would like this one to stay because even though it won't work "as is" with the updated HTTP layer it contains the code and tests to take our current HTTP classes to(wards) PSR-7. The other PR really just makes deprecations and splits classes into traits.

agungsugiarto:
Hello @MGatner, this pr will be merge on future? Im will be happy, because codeigniter step closer with compliance psr. Thanks

MGatner:
@agungsugiarto It is one of my goals to have the framework provide a PSR-7 compliant HTTP layer, but there are a few shifts that need to happen before that is possible.

kenjis:
Ref #3230

kenjis:
@MGatner Because of coding style changes, it is difficult to rebase the latest develop.
But it seems we could merge changes like this PR into v4.2.
What do you think?

MGatner:
I think I've changed my mind on this. Our current HTTP layer is unequivocally at odds with PSR-7, and in order to move it to compliance we would either need a breaking change or a new namespace. 

Because of this I think the best short-term solution is to do what we did with Cache: create a separate, optional repo with wrapper classes that are PSR compliant yet use the framework's HTTP layer.

lonnieezell:
@MGatner can this be closed then?

PR #3942, opened 2020-11-29
Update php version constraints

paulbalandan:
**Description**
Avoid open ended PHP version constraints.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/admin/framework/composer.json
+++ b/admin/framework/composer.json
@@ -5,7 +5,7 @@
 	"homepage": "https://codeigniter.com",
 	"license": "MIT",
 	"require": {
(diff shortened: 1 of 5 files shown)


MGatner:
Are we certain that `4.0.5` will be compatible with PHP 8? I don't know off the top of my head why not but it would be good to be sure...

paulbalandan:
I think users of `4.0.5` on PHP8 will be getting a lot of deprecation notices on the ordering of optional params before required params. Other than that, I don't think there will be problems as you have prep the code before.

MGatner (approved):

PR #3945, opened 2020-11-30
Fix CS for Email

paulbalandan:
**Description**
This only does the fixing of code style within the Email class to make it less stuffy to read and not the big refactor as originally envisioned.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -28,86 +28,101 @@ class Email
 	 * @var array|null
 	 */
 	public $archive;
+
 	/**
 	 * Properties to be added to the next archive.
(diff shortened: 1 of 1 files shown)


MGatner:
Nice cleanup! I always wondered what to do with spacing on lines with `=` followed by `.=` `+=` etc.... This seems to suggest you ignore their indentation

paulbalandan:
Because the Email class heavily used short circuiting expressions, some are out of view in my IDE due to line length so I have to add an intervening line to re-align those.

PR #3956, opened 2020-12-02
Fix mime types for .srt and .vtt.

### LINKED ISSUE #3921: Bug: Uploaded SRT files are saved as CSV
**Describe the bug**
Uploading SRT files does not work, SRT files are saved as `CSV`.

**CodeIgniter 4 version**
dev-develop

**Affected module(s)**
CodeIgniter\HTTP\Files\UploadedFile

Create a form.
Upload a file name `test.srt`.
Uploaded file is renamed `test.csv` instead of `test.srt`.

**Context**
 - OS: Ubuntu 20.04
 - Web server nginX
 - PHP version 7.3

**Proposed Solution**
When a `srt` file is uploaded, its mime type is recognized as `application/octet-stream`.
Then `CodeIgniter\Files\File->guessExtension()` calls `Mimes::guessExtensionFromType($this->getMimeType());` to get the file extension.
Since `'srt'` only has `['text/srt', 'text/plain']` the first type matching  `application/octet-stream` is returned : `CSV`.

So in order to make it work, in `Config\Mimes.php` we would just need to update `srt` to:
`'srt' => ['text/srt', 'text/plain', 'application/octet-stream'],`
Et voilà !

sfadschm:
Fixes #3921.
Came across this when trying to upload `.vtt` files.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Config/Mimes.php
+++ b/app/Config/Mimes.php
@@ -462,10 +462,12 @@ class Mimes
 		'srt'   => [
 			'text/srt',
 			'text/plain',
+			'application/octet-stream',
 		],
 		'vtt'   => [
 			'text/vtt',
 			'text/plain',
+			'application/octet-stream',
 		],
 		'ico'   => [
 			'image/x-icon',


paulbalandan (commented):
I don't the fix for that is as simple as this. 

Currently, csv is the first match for application/octet-stream. Even if adding the type to srt and vtt, STILL the first matching array key is csv. In line 517 of your edited file you will see it is using foreach, so it is always csv that will match first.

sfadschm:
This fix is actually working, because in #2128, a fallback to the clients extension was added to `UploadedFile->guessExtension()`.

This passes the client extension to the `Mimes` method and then matches before the mentioned foreach loop here:
https://github.com/codeigniter4/CodeIgniter4/blob/2b65052534c5849570d18304a314a22c13c63b52/app/Config/Mimes.php#L510-L515

paulbalandan (approved):
OK. I've missed looking at `UploadedFile`.

Thanks.

Issue #3957, opened 2020-12-02
PHP8: Deprecate required param after optional


enix-app:
I've found a lot of code bugs in PHP v8. Generally the use of parameters in functions.
Reference: https://php.watch/versions/8.0/deprecate-required-param-after-optional

Why don't the required parameters take precedence?
This should not be the writing syntax.

sfadschm:
Are you using the latest develop version?
This was done already in #3938.

enix-app:
Okay, thanks!

PR #3958, opened 2020-12-03
Avoid required arguments after typed nullable arguments.

sfadschm:
Extends #3938 and related to #3957.

**Description**
This PR adds default values for all function arguments that follow an optional argument.
Interestingly, while such argument arrangement triggers a depracation notice in PHP8 if the first optional argument is not typed, it does not as long as the optional parameter is typed and its default value is `null`.

E.g., the following function definitions are incorrect and will trigger the PHP8 warning:
`function test(string $param1 = 'hi' string $param2)`
`function test($param1 = null, string $param2)`

This definition does not trigger the warning, but is still incorrect:
`function test(string $param1 = null, string $param2)`

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -374,13 +374,13 @@ class CodeIgniter
 	 * Handles the main request logic and fires the controller.
 	 *
 	 * @param RouteCollectionInterface|null $routes
-	 * @param Cache                         $cacheConfig
+	 * @param Cache|null                    $cacheConfig
 	 * @param boolean                       $returnResponse
 	 *
 	 * @return RequestInterface|ResponseInterface|mixed
 	 * @throws RedirectException
 	 */
-	protected function handleRequest(RouteCollectionInterface $routes = null, Cache $cacheConfig, bool $returnResponse = false)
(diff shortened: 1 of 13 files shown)


sfadschm:
Not sure about the PHPStan error.
The test passes `null` to the `$body` argument of `IncomingRequest`. Does this intend `$body` to fallback to its default value or should it become `''` by this?

MGatner:
Unfortunately I don't think it is as easy as providing default null inputs. Because the original methods were "valid" the current parameters function as required with their defined type. This means `foo(string $arg, $bar = null)` is not the same as `foo(string $arg = null, $bar = null)`. The former throws an error of $arg is not passed and expects it to be string only. 

Take a closer look at the original PR and add exceptions if you want to implement this the same way.

sfadschm:
You are right, these changes were not carefully thought through.

I just read through the deprecation discussion at PHP again and I think this whole PR is actually meaningless.
As most of the cases covered here are actually just workaround for nullable typed arguments, these could be left the way they are.
The only thing that might be done is rather cosmetic and would be to change `string $param = null` to `?string $param`.

I will be closing this for that reason.
Should the mentioned cosmetic changes be done?

PR #3836, opened 2020-10-30
Separate Security Configuration

mostafakhudair:
Creates new Config\Security class and make \CodeIgniter\Security\Security class depends on it

removes useless properties
rename some properties

rename some methods
```
CSRFVerify() to verify()
getCSRFHash() to getHash()
getCSRFTokenName() to getTokenName()
setHash() to generateHash()
```

add new public methods:
```
getHeaderName()
getCookieName()
isExpired()
shouldRedirect()
```

edits changes in user guide

- [x] Securely signed commits
- [x] Component(s) with PHPdocs
- [x] User guide updated
- [x] Conforms to style guide

Closes #3703

PR #3961, opened 2020-12-04
Refactor Security Class

mostafakhudair:
Creates new Config\Security class and make \CodeIgniter\Security\Security class depends on it

removes useless properties
rename some properties

rename some methods
```
CSRFVerify() to verify()
getCSRFHash() to getHash()
getCSRFTokenName() to getTokenName()
setHash() to generateHash()
```

add new public methods:
```
getHeaderName()
getCookieName()
isExpired()
shouldRedirect()
```

edits changes in user guide

- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

Closes #3836


system/Config/Services.php
```diff
@@ -677,7 +677,7 @@ public static function security(App $config = null, bool $getShared = true)
 			return static::getSharedInstance('security', $config);
 		}
 
-		$config = $config ?? config('App');
+		$config = $config ?? config('App') ?? config('Security');
```
paulbalandan, line 680: I think the check for Security config should come first before App.
```suggestion
		$config = $config ?? config('Security')  ?? config('App');
```

system/Security/Security.php
```diff
@@ -411,4 +376,89 @@ public function sanitizeFilename(string $str, bool $relativePath = false): strin
...
+	protected function generateHash(): string
+	{
+		if (is_null($this->hash))
+		{
+			// If the cookie exists we will use its value.
+			// We don't necessarily want to regenerate it with
+			// each page load since a page could contain embedded
+			// sub-pages causing this feature to fail
+			if (isset($_COOKIE[$this->cookieName]) && is_string($_COOKIE[$this->cookieName]) && preg_match('#^[0-9a-f]{32}$#iS', $_COOKIE[$this->cookieName]) === 1
+			)
```
paulbalandan, line 394: Can we break this in viewable line breaks?

paulbalandan (commented):
The refactor is good to me but we need the review of others since this is a major component of the framework. Some notes:

- Fix the user guide syntax check
- In the main Security class, you cannot rename or remove the public and protected methods as it would mean a BC break for people using those. The deprecations and detours for the renamed Security methods are good though.
- See other comments below.

MGatner:
This looks great! I'm a fan of splitting these out, you've been doing a great job with it. I'd like to give this one more look when I'm not on mobile (or have someone else's review) but let's plan to include this in 4.1 so these deprecations can get entered.

MGatner:
Is there any merit to merging this and ContentSecurityPolicy config? I'm not terribly knowledgeable of the various security components but as a novice having all "security" under one Config seems logical?

mostafakhudair:
> Is there any merit to merging this and ContentSecurityPolicy config? I'm not terribly knowledgeable of the various security components but as a novice having all "security" under one Config seems logical?

While ContentSecurityPolicy config serve HTTP\ContentSecurityPolicy class, and Security config serve Security\Security class, I think there is no point in merging them together at least for now.

lonnieezell:
I think this looks good also. @MGatner are we good to merge this into 4.0.5 or should this be a 4.1 change? It looks like everything is deprecated and BC safe.

MGatner:
@lonnieezell I think this is fine for 4.0.5. Since we decided to support PHP 8 in 4.0.5 the only real rubric for 4.1 is support for PHP 7.2.

PR #3964, opened 2020-12-04
Optimize Autoloader

mostafakhudair:
Optimize Autoloader Filelocator classes

add new Autoloader method `loadClassmap()` that registers classmap files with autoloader

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -233,18 +233,18 @@ class Autoloader
 	 *
 	 * @param string $class The class name.
 	 *
-	 * @return string|false The mapped file on success, or false on failure.
+	 * @return void
 	 */
-	public function loadClassmap(string $class)
+	public function loadClassmap(string $class): void
 	{
 		$classmap = is_array($this->classmap) ? $this->classmap : [];
 
-		if (empty($classmap[$class]))
+		if (! empty($classmap[$class]))
 		{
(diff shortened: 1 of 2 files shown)


lonnieezell:
Have you been able to measure this and see if it helps at all? I only ask because I tried some other optimizations around this early on and found that PHP's file location cache is actually really good, and the gains were extremely minimal and just added complexity to the framework.

system/Autoloader/Autoloader.php
```diff
@@ -221,45 +205,57 @@ public function removeNamespace(string $namespace)
...
+	 * @return string|false The mapped file on success, or false on failure.
 	 */
 	public function loadClass(string $class)
 	{
-		$class = trim($class, '\\');
-		$class = str_ireplace('.php', '', $class);
+		$class = str_ireplace('.php', '', trim($class, '\\'));
 
-		$mapped_file = $this->loadInNamespace($class);
+		$mappedFile = $this->loadInNamespace($class);
```
samsonasik, line 218: use underscore on purpose as removed in 4.1, not worth updated for conflict, see rector.php config to skip this file for check underscore to camelcase check

MGatner:
I think this has some good content but at this point I'd rather see `4.1` released first to resolve all the various discrepancies and then rebase this against it.

PR #3966, opened 2020-12-05
fix: add dot_array_search to required_with and required_without

### LINKED ISSUE #3965: Bug: required_with validation rule does not work with arrays
**Describe the bug**
When using array as key in a form, `required_with` never finds the field.

**CodeIgniter 4 version**
dev-develop

**Affected module(s)**
`CodeIgniter\Validation\Rules::required_with()`

With these data:
```
'contacts' => [
    'name' => 'Joe Smith',
    'friend' => 'John Doe'
]

$validation->setRules([
    'contacts.name' => 'required_with[contacts.friend]'
]);
```
`required_with` never finds `contacts.friend` because no `dot_array_search` is performed.

**Context**
 - OS: Ubuntu
 - Web server: Nginx
 - PHP version: 7.4

**Suggested correction**
in `codeigniter4/codeigniter4/system/Validation/Rules.php` (`CodeIgniter\Validation\Rules::required_with()`):
```php
	/**
	 * The field is required when any of the other required fields are present
	 * in the data.
	 *
	 * Example (field is required when the password field is present):
	 *
	 *     required_with[password]
	 *
	 * @param string|null $str
	 * @param string      $fields List of fields that we should check if present
	 * @param array       $data   Complete list of fields from the form
	 *
	 * @return boolean
	 */
	public function required_with(
		$str = null,
		string $fields,
		array $data
	): bool {
		$fields = explode(',', $fields);

		// If the field is present we can safely assume that
		// the field is here, no matter whether the corresponding
		// search field is present or not.
		$present = $this->required($str ?? '');

		if ($present) {
			return true;
		}

		// Still here? Then we fail this test if
		// any of the fields are present in $data
		// as $fields is the lis
		$requiredFields = [];

		foreach ($fields as $field) {
			if (
				(strpos($field, '.') !== false &&
					!empty(dot_array_search($field, $data))) ||
				(array_key_exists($field, $data) && !empty($data[$field]))
			) {
				$requiredFields[] = $field;
			}
		}

		return empty($requiredFields);
	}
```

benjaminbellamy:
Closes #3965

**Description**
Added dot_array_search to required_with and required_without so that they work when using array as key in a form.

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -422,13 +422,13 @@ class Rules
 
 		// Still here? Then we fail this test if
 		// any of the fields are not present in $data
-		foreach ($fields as $field) {
+		foreach ($fields as $field)
+		{
 			if (
-				(strpos($field, '.') !== false &&
-					empty(dot_array_search($field, $data))) ||
-				(strpos($field, '.') === false &&
-					(!array_key_exists($field, $data) || empty($data[$field])))
-			) {
+				(strpos($field, '.') !== false && empty(dot_array_search($field, $data))) ||
+				(strpos($field, '.') === false && (! array_key_exists($field, $data) || empty($data[$field])))
+			)
+			{
 				return false;
 			}
 		}
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -379,8 +379,8 @@ class Rules
 		foreach ($fields as $field)
 		{
 			if (
-				(strpos($field, '.') !== false && ! empty(dot_array_search($field, $data))) ||
-				(array_key_exists($field, $data) && ! empty($data[$field]))
+				(array_key_exists($field, $data) && ! empty($data[$field])) ||
+				(strpos($field, '.') !== false && ! empty(dot_array_search($field, $data)))	
 			)
 			{
 				$requiredFields[] = $field;
@@ -425,8 +425,8 @@ class Rules
 		foreach ($fields as $field)
 		{
 			if (
-				(strpos($field, '.') !== false && empty(dot_array_search($field, $data))) ||
-				(strpos($field, '.') === false && (! array_key_exists($field, $data) || empty($data[$field])))
+				(strpos($field, '.') === false && (! array_key_exists($field, $data) || empty($data[$field]))) ||
+				(strpos($field, '.') !== false && empty(dot_array_search($field, $data)))
 			)
 			{
 				return false;
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -376,19 +376,15 @@ class Rules
 		// as $fields is the lis
 		$requiredFields = [];
 
-		foreach ($fields as $field)
-		{
-			if (array_key_exists($field, $data))
-			{
+		foreach ($fields as $field) {
+			if (
+				(strpos($field, '.') !== false &&
+					!empty(dot_array_search($field, $data))) ||
+				(array_key_exists($field, $data) && !empty($data[$field]))
+			) {
 				$requiredFields[] = $field;
 			}
 		}
-
-		// Remove any keys with empty values since, that means they
-		// weren't truly there, as far as this is concerned.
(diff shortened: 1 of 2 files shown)


system/Validation/Rules.php
```diff
@@ -376,20 +376,16 @@ public function required_with($str = null, string $fields, array $data): bool
...
-		foreach ($fields as $field)
-		{
-			if (array_key_exists($field, $data))
-			{
+		foreach ($fields as $field) {
+			if (
+				(strpos($field, '.') !== false &&
+					!empty(dot_array_search($field, $data))) ||
+				(array_key_exists($field, $data) && !empty($data[$field]))
+			) {
```
mostafakhudair, line 384:
```suggestion
		foreach ($fields as $field)
		{
			if (
				(strpos($field, '.') !== false && ! empty(dot_array_search($field, $data))) ||
				(array_key_exists($field, $data) && ! empty($data[$field]))
			)
			{
```

system/Validation/Rules.php
```diff
@@ -425,10 +421,13 @@ public function required_without($str = null, string $fields, array $data): bool
...
-		{
-			if (! array_key_exists($field, $data))
-			{
+		foreach ($fields as $field) {
+			if (
+				(strpos($field, '.') !== false &&
+					empty(dot_array_search($field, $data))) ||
+				(strpos($field, '.') === false &&
+					(!array_key_exists($field, $data) || empty($data[$field])))
+			) {
```
mostafakhudair, line 430:
```suggestion
		foreach ($fields as $field)
		{
			if (
				(strpos($field, '.') !== false && empty(dot_array_search($field, $data))) ||
				(strpos($field, '.') === false && (! array_key_exists($field, $data) || empty($data[$field])))
			)
			{
```

mostafakhudair (changes requested):

system/Validation/Rules.php
```diff
@@ -378,18 +378,15 @@ public function required_with($str = null, string $fields, array $data): bool
 
 		foreach ($fields as $field)
 		{
-			if (array_key_exists($field, $data))
+			if (
+				(strpos($field, '.') !== false && ! empty(dot_array_search($field, $data))) ||
+				(array_key_exists($field, $data) && ! empty($data[$field]))
```
lonnieezell, line 383: I think flipping the order of these two will eek out a little more performance since `empty` and `array_key_exists` and almost assuredly faster than `dot_array_search`.

system/Validation/Rules.php
```diff
@@ -427,7 +424,10 @@ public function required_without($str = null, string $fields, array $data): bool
 		// any of the fields are not present in $data
 		foreach ($fields as $field)
 		{
-			if (! array_key_exists($field, $data))
+			if (
+				(strpos($field, '.') !== false && empty(dot_array_search($field, $data))) ||
+				(strpos($field, '.') === false && (! array_key_exists($field, $data) || empty($data[$field])))
```
lonnieezell, line 429: Same thing here. Please reverse the order.

lonnieezell (changes requested):
I left a very nit-picky comment lol. 

Other than that, though, we do need tests in place to cover this feature before we can merge.

benjaminbellamy (commented):
I switched them.

system/Validation/Rules.php
benjaminbellamy: Done! (And for what it’s worth, I tested it.)

system/Validation/Rules.php
benjaminbellamy: Done!

paulbalandan:
@benjaminbellamy the tests Lonnie referred here are not local tests but the PHPUnit tests. You should add new tests to the existing tests to see that this fix is working as intended.

benjaminbellamy:
> @benjaminbellamy the tests Lonnie referred here are not local tests but the PHPUnit tests. You should add new tests to the existing tests to see that this fix is working as intended.

Done!

paulbalandan (approved):
@benjaminbellamy  thank you

PR #3968, opened 2020-12-06
Use nullable typehints for non-optional typed function arguments.

sfadschm:
This PR addresses the case when a typed function argument has a defined default of `null` and is followed by a required argument.
In this case, the assigned default value is only used to enable a nullable type before PHP 7.1.

In this PR, such assignements are changed to real nullable types.
Also fixed some DocBlocks that I came across.

Possible extension: Use nullable type also for typed optional arguments with a default `null` value.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -380,7 +380,7 @@ class CodeIgniter
 	 * @return RequestInterface|ResponseInterface|mixed
 	 * @throws RedirectException
 	 */
-	protected function handleRequest(RouteCollectionInterface $routes = null, Cache $cacheConfig, bool $returnResponse = false)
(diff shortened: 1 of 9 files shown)


samsonasik (approved):

MGatner:
I'm sorry I missed this until now. Most of these were addressed very cautiously to `4.1` branch in order to prevent breaking changes. I have just now cherry-picked those changes into `develop` so `4.0.5` can support PHP 8. If you would still like to migrate `null` defaults into true nullable types please rebase and fix the conflicts.

sfadschm:
Conflicts are resolved. I like the implementation with the `InvalidParameterException`.
I still think this true nullable types are a useful PHP feature, so we should use it.

samsonasik:
Thank you @sfadschm for the PR.
Thank you @MGatner for the review.

PR #3972, opened 2020-12-08
Shortags are deprecated in php7.4

gnovaro:
Large version of php echo always have backward compatibility
Shortags will be removed in php 8.0
https://wiki.php.net/rfc/deprecate_php_short_tags

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)

### DIFF
--- a/app/Views/welcome_message.php
+++ b/app/Views/welcome_message.php
@@ -219,7 +219,7 @@
 
 	<div class="heroe">
 
-		<h1>Welcome to CodeIgniter <?= CodeIgniter\CodeIgniter::CI_VERSION ?></h1>
+		<h1>Welcome to CodeIgniter <?php echo CodeIgniter\CodeIgniter::CI_VERSION ?></h1>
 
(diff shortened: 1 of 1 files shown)


paulbalandan (commented):
The `<?=` is not covered by the short php opening deprecation. This RFC covers `<?` used in place of `<?php`.

@see https://wiki.php.net/rfc/deprecate_php_short_tags_v2

> ## Current state
> Currently the <? short open tag is controlled by the short_open_tag ini setting. This ini setting is enabled by default (if no ini files is used), but disabled in both php.ini-development and php.ini-production.
>
> If short_open_tag is enabled <? is interpreted as the start of a PHP context. If short_open_tag is disabled, <? is interpreted as literal text without special meaning.
>
> The <?= combined opening and echo tag is not controlled by short_open_tag. It is always available, and will continue to be always available after this RFC.

PR #3973, opened 2020-12-08
Deprecate redundant HTTP keys

paulbalandan:
**Description**
With the merge of the refactored Security class, the redundant `HTTP.disallowedAction` and `HTTP.invalidSamSiteSetting` can be removed in favor of the `Security` counterpart.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/HTTP/Exceptions/HTTPException.php
+++ b/system/HTTP/Exceptions/HTTPException.php
@@ -241,6 +241,6 @@ class HTTPException extends FrameworkException
 	 */
 	public static function forInvalidSameSiteSetting(string $samesite)
 	{
-		return new static(lang('HTTP.invalidSameSiteSetting', [$samesite]));
+		return new static(lang('Security.invalidSameSiteSetting', [$samesite]));
 	}
 }
--- a/system/Language/en/HTTP.php
+++ b/system/Language/en/HTTP.php
@@ -54,6 +54,7 @@ return [
 	'methodNotFound'             => 'Controller method is not found: {0}',
 
 	// CSRF
+	// @deprecated use `Security.disallowedAction`
 	'disallowedAction'           => 'The action you requested is not allowed.',
 
 	// Uploaded file moving
@@ -72,5 +73,6 @@ return [
 	'uploadErrUnknown'           => 'The file "%s" was not uploaded due to an unknown error.',
 
 	// SameSite setting
+	// @deprecated use `Security.invalidSameSiteSetting`
 	'invalidSameSiteSetting'     => 'The SameSite setting must be None, Lax, Strict, or a blank string. Given: {0}',
 ];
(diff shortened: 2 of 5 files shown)


sfadschm:
Actually, I just checked and they still seem to be in use, e.g., here:
https://github.com/codeigniter4/CodeIgniter4/blob/a2de1f6f966fb1e7f62f1038f517244a8460ab04/system/HTTP/Exceptions/HTTPException.php#L242-L245
and in the corresponding tests.
I'm all for dropping as many duplicates as possible but not sure if it makes sense to change `HTTP` to `Security` in the above example.

sfadschm:
Can we maybe add some testing for all language strings called in the framework to see if they are really defined?

MGatner:
I like @sfadschm's idea, it could even be implemented as an Action for PRs. It should check both directions: for language calls missing a key and for unused keys. 

For this PR, and generally, I think that we need to deprecate language strings instead of removing them because any module or project could be using the framework strings. While it technically would not crash the code the output could suddenly change to something unhelpful like "System.description"

sfadschm:
Checking both directions sounds reasonable. I though about such tests some days ago, but the only thing I could come up with was to load all language keys, then read the contents of all framework files and `preg_match_all` all the contents with `lang('...')`.
The results could then be generated with two simple `array_diffs` but I still feel this puts some significant extra load on the tests.

I agree on the deprecation, but I think we can not really have both tests and deprecation, as the tests would have no way to tell a language string has alreaddy been deprecated and should not throw an error, or am I missing out on something?

paulbalandan:
I believe system messages should be strictly for internal use and should not be used in users' code and covered by BC promise. The usage should remain strictly in the framework alone and users should implement their own keys. Since these messages are typically used within thrown exceptions, users wanting to use the messages should use it rather indirectly, i.e. using the `throw Exception::something();`

sfadschm:
> I believe system messages should be strictly for internal use and should not be used in users' code and covered by BC promise. The usage should remain strictly in the framework alone and users should implement their own keys. Since these messages are typically used within thrown exceptions, users wanting to use the messages should use it rather indirectly, i.e. using the `throw Exception::something();`

I'm with you on this, but I still also think that there is projects out there using those messages, which will face some unexpected messages like MGatner said.

How about applying this to 4.1 with a note in the update changelog and maybe a small notice in the user manual, that these keys are not BC covered?

MGatner:
@paulbalandan What do you think of @sfadschm's idea? I think that is a good compromise: we add a comment that these are deprecated but leave them, then note in the UG and changelog for 4.1 the system language strings are all considered *internal* and can be changed without warning.

@lonnieezell I think you should weigh in on this as well.

lonnieezell:
I agree system messages are intended to be for internal use. I like the suggested compromise.

MGatner:
Key deprecation looks good. The PHP 8 SA is allowed (https://github.com/codeigniter4/CodeIgniter4/pull/4065). What about a line in the UG and changelog about language files being internal? or were you thinking that would be a separate PR?

paulbalandan:
> Key deprecation looks good. The PHP 8 SA is allowed (#4065). What about a line in the UG and changelog about language files being internal? or were you thinking that would be a separate PR?

Yes, I'm thinking of a separate PR for them.

PR #3975, opened 2020-12-09
spark: validate the PHP version

enix-app:
**Description**

As ``index.php`` have this, ``spark`` must have it too. I just copied and pasted it.

### DIFF
--- a/spark
+++ b/spark
@@ -1,6 +1,14 @@
 #!/usr/bin/env php
 <?php
 
+// Valid PHP Version?
+$minPHPVersion = '7.2';
+if (version_compare(PHP_VERSION, $minPHPVersion, '<'))
+{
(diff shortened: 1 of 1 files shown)


paulbalandan (commented):
This check is already done in `ServeCommand`.

Maybe what we can do is remove the check in ServeCommand and have the check here instead.

enix-app:
It's better to put it in front, not in command.
Spark will not show a list of commands if the PHP version does not match, 

```bash
$ php8 spark
```

paulbalandan:
I agree an early terminating call to unsupported PHP version check is more preferable.

MGatner:
@enix-app Can you remove the check from ServeCommand?

enix-app:
Sorry there was an accident. My memory is short term :)

I opened a new PR. I'll close this.

PR #3983, opened 2020-12-12
Build Cookie Class

mostafakhudair:
Separated cookie class will help us to use cookie in whole  framework such as. Security::sendCookie(), Session::setCookie(), Response::sendCookies(), now we are able to do this from one main class \CodeIgniter\Cookie\Cookie

after merge: i'll update other classes with new cookie class methods
and will deprecate some methods and properties

userguide not finished yet

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


mostafakhudair:
Fails i'll fix when i back to home

MGatner:
Initially looks good. I don't have time to do a proper review but I approve the approach. Thank you!

mostafakhudair:
Some of tests were failed I'll fix it soon

paulbalandan:
Before you progress further, please rename your subsequent commits to be more descriptive of the change. If it will just be the same, I assume that it is only an amended commit and should be forced pushed instead.

system/Cookie/BaseCookie.php
```diff
@@ -0,0 +1,306 @@
...
+	 *
+	 * @param CookieConfig $config
+	 * 
+	 * @throws CookieException
+	 */
+	public function __construct(CookieConfig $config)
+	{
+		$this->properties = get_object_vars($config);
+
+		foreach (get_class_vars(get_class($this)) as $key => $value)
```
paulbalandan, line 98: Is there a reason on using `get_class_vars(get_class($this))` instead of using `get_object_vars($this)`?

system/Cookie/BaseCookie.php
```diff
@@ -0,0 +1,306 @@
...
+	 * @return $this
+	 */
+	public function setSameSite(string $samesite)
+	{
+		if (! in_array(strtolower($samesite), ['none', 'lax', 'strict', ''], true))
+		{
+			$samesite = $this->properties['samesite'];
+		}
+        
+        $this->samesite = ucfirst($samesite);
```
paulbalandan, line 270: Kindly align this one.

system/Cookie/BaseCookie.php
```diff
@@ -0,0 +1,306 @@
...
+	//--------------------------------------------------------------------
+
+	/**
+	 * Reset configuration to default.
+	 *
+	 * @return $this
+	 */
+	public function reset()
+	{
+		foreach (get_class_vars(get_class($this)) as $key => $value)
```
paulbalandan, line 296: Same comment with `get_class_vars` above.

system/Cookie/BaseCookie.php
```diff
@@ -0,0 +1,306 @@
...
+	 * 
+	 * @throws CookieException
+	 */
+	public function __construct(CookieConfig $config)
+	{
+		$this->properties = get_object_vars($config);
+
+		foreach (get_class_vars(get_class($this)) as $key => $value)
+		{
+			if (property_exists($this, $key) && isset($this->properties[$key]))
```
paulbalandan, line 100:
```suggestion
			if (isset($this->properties[$key]))
```
I don't think `property_exists` is needed as you are checking already on the child cookie's properties.

system/Cookie/BaseCookie.php
```diff
@@ -0,0 +1,306 @@
...
+	/**
+	 * Reset configuration to default.
+	 *
+	 * @return $this
+	 */
+	public function reset()
+	{
+		foreach (get_class_vars(get_class($this)) as $key => $value)
+		{
+			if (property_exists($this, $key) && isset($this->properties[$key]))
```
paulbalandan, line 298:
```suggestion
			if (isset($this->properties[$key]))
```

system/Cookie/Cookie.php
```diff
@@ -0,0 +1,303 @@
...
+	 *
+	 * @param CookieConfig $config
+	 */
+	public function __construct(CookieConfig $config)
+	{
+		parent::__construct($config);
+	}
+
+	//--------------------------------------------------------------------
+
```
paulbalandan, line 39: This constructor can be removed if no special initializations would be done by the implementing child cookie.

system/Cookie/Cookie.php
```diff
@@ -0,0 +1,303 @@
...
+use Config\Services;
+
+class Cookie extends BaseCookie implements CookieInterface
+{
+	/**
+	 * Stores all cookies.
+	 *
+	 * @var array
+	 */
+	protected $cookies = [];
```
paulbalandan, line 24: I thought the `Cookie` class will be representing a single web cookie, so I'm surprised to see this was actually a "jar of cookies" class. I would expect, and others might too, that this is a single cookie object representation. My suggestion is to split the unique Cookie class from the Cookie collection capacity.

In Laravel, they have this `CookieJar` class which is the "jar" for all cookies, and calling a cookie would be done through this class. In CakePHP I think it is named `CookieCollection`.

So basically, this class would be split:
- `Cookie` representing a single web cookie which properties like `name` and `prefix` would be accessed.
- `CookieCollection` or `CookieJar` (or whatever you like) creating and storing all `Cookie` objects like some sort of a factory class.

paulbalandan (changes requested):
The initial implementation is good for me. I have some comments though. My major concern, however, is I don't like the `Cookie` class performing two functions here: representing a single cookie, through extending `BaseCookie`, and representing at the same time a collection of cookies, through implementing `CookieInterface`. These two functions should be on two distinct classes. Please read through my comments for my explanations.

MGatner:
This looks like a real quality PR. @paulbalandan it looks to me like your requests have been addressed - okay with merging? I don't do a lot with cookies so would like someone else to confirm the merge.

MGatner (approved):

paulbalandan:
Oh, missed this one. I'll do another review this evening after work. 😄

system/Cookie/BaseCookie.php
```diff
@@ -0,0 +1,306 @@
...
+	 * @return $this
+	 */
+	public function setSameSite(string $samesite)
+	{
+		if (! in_array(strtolower($samesite), ['none', 'lax', 'strict', ''], true))
+		{
+			$samesite = $this->properties['samesite'];
+		}
+
+		$this->samesite = ucfirst(strtolower($samesite));
```
paulbalandan, line 270: I think we should be consistent in throwing exceptions here if wrong samesite is given. That way, developers won't get confused why there are setting this value but another one gets assigned. It's like telling them directly the problem they're doing.

system/Cookie/Exceptions/CookieException.php
```diff
@@ -0,0 +1,23 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Cookie\Exceptions;
+
+use CodeIgniter\Exceptions\ExceptionInterface;
+use CodeIgniter\Exceptions\FrameworkException;
+
+class CookieException extends FrameworkException implements ExceptionInterface
```
paulbalandan, line 17: No need to reimplement the interface. `FrameworkException` already implements it.

paulbalandan (commented):
I see that you now have a separate `CookiePresenter` class. Thanks for indulging my reviews. Allow me to make some more.

In your proposed implementation, `CookiePresenter` is the single web cookie while `Cookie` is the cookie collection. I believe it would be more natural, or more fluent, to have the opposite. `Cookie` as the single web cookie class, while `CookiePresenter` as the collection of cookies.

Two interfaces:
- `CookieInterface` -  interface for single web cookie tokens. This would contain the methods defined in the proposed `BaseCookie` class.
- `CookiePresenterInterface` - interface for cookie collection objects. This would contain the methods in your proposed `CookieInterface`.

Abstraction
- `BaseCookie` would be implementing `CookieInterface`, providing concrete classes to do little changes
- `Cookie` would be extending `BaseCookie`. In the proposed implementation, BaseCookie accepts a `CookieConfig` instance while `CookiePresenter` expects several arguments. Maybe we can do is the `Cookie` will inherit the `CookieConfig` instance then just override the values using the `set()` method.

Services
- `Services::cookie()` will now return a `CookiePresenter` if no args supplied, but an instance of `Cookie` if some args are supplied, or `null` if inexistent
- Getting cookies from services will return either null or instance of `Cookie`

mostafakhudair:
Thanks for the review with many good points, but about current names  I think`Cookie` for cookie collection and `CookiePresenter`for present single web cookie is much cooler regardless of the logic,, I belive that you are right but i would like to hear others opinions

sfadschm:
Personally, I would go for `Cookie` to present a single cookie. 
How is this handled with other collections throughout the framework?
E.g., for files, we have `FileCollection` and `File`.

mostafakhudair:
@MGatner what about this PR :)

mostafakhudair:
@michalsn would you take a look here please?

PR #3988, opened 2020-12-14
Fix testing docs indentation

kenjis:
**Description**
A tab is converted to 8 spaces. So it is better to use only spaces for indentation.

screenshot

**Checklist:**
- [x] Securely signed commits
- [n/a] Component(s) with PHPdocs
- [n/a] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide


paulbalandan (commented):
If that's the case then almost all RST docs should be changed too, not only these.

kenjis:
@paulbalandan Yes.

PR #3995, opened 2020-12-17
Changes in Autoloader::register() method

iRedds:
Removed spl_autoload_extensions() function
This function is used with spl_autoload() or spl_autoload_register() (no arguments). 
But CI4 does not use such methods for class autoloading.  

The anonymous function defined in a class has access to the class context. 
This means that we can use the $this variable inside an anonymous function.

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -113,27 +113,19 @@ class Autoloader
 	 */
 	public function register()
 	{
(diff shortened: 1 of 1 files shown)


paulbalandan (approved):

samsonasik (approved):

samsonasik:
Thank you @iRedds

PR #4006, opened 2020-12-20
Fix Access BaseController Methods via url

mostafakhudair:
While the route methods is sensitive to letters we still can access BaseController::index via localhost/BaseController or localhost/BaseController/index, because route configured to prevent access localhost/basecontroller, since autoRoute method ucfirst('basecontroller') :: output > Basecontroller that is not actually found.

- [x] Securely signed commits


samsonasik (approved):

samsonasik:
Thank you @mostafakhudair

PR #4007, opened 2020-12-21
ignore files when using composer

totoprayogo1916:
**Description**
Exclude some files when users run `create-project`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/admin/starter/.gitattributes
+++ b/admin/starter/.gitattributes
@@ -1,0 +1,2 @@
+/.gitattributes export-ignore
+/.github export-ignore


paulbalandan (changes requested):
`builds` is needed to quickly switch between stable and unstable versions such as to test snapshot builds or test new unreleased features. Do not put it here.

admin/starter/.gitattributes
```diff
@@ -0,0 +1,3 @@
+.gitattributes export-ignore
+
```
samsonasik, line 2: no need multiple lines, single line is ok

admin/starter/.gitattributes
```diff
@@ -0,0 +1,2 @@
+.gitattributes export-ignore
+.github/ export-ignore
```
samsonasik, line 2: `/` prefix need to be added to ensure only ignore root dir/file, see https://github.com/codeigniter4/CodeIgniter4/blob/develop/admin/framework/.gitattributes

paulbalandan (approved):

samsonasik (approved):

samsonasik:
@totoprayogo1916 thank you for the PR.

PR #3932, opened 2020-11-27
PHP 8 Action workflows

MGatner:
**Description**
Enables PHP 8 in the relevant CI workflows. This PR will require #3938 and a rebase before being close to passing.

**Checklist:**
- [X] Securely signed commits
- [X] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [X] Conforms to style guide

Issue #3974, opened 2020-12-09
Bug: PHP8 session flush data is not deleted


katsuhidem:
Due to a change in the string-number comparison specification, 'old' is always a false and will not be removed

https://github.com/codeigniter4/CodeIgniter4/blob/a2de1f6f966fb1e7f62f1038f517244a8460ab04/system/Session/Session.php#L430

**Context**
 - OS: Windows 10
 - Web server Apache 
 - PHP version 8.0.0

PR #4012, opened 2020-12-21
PHP 8 Actions

MGatner:
*Supercedes #3932*

**Description**
Adds PHP 8 to the testing and analysis workflows. This will give some idea of the work left to do for `4.0.5` to support PHP 8, and any conflicts among dependencies for supporting both `7.2` and `8.0`

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/.github/workflows/test-phpstan.yml
+++ b/.github/workflows/test-phpstan.yml
@@ -20,8 +20,12 @@ on:
 
 jobs:
   build:
-    name: Analyze code (PHPStan)
+    name: PHP ${{ matrix.php-versions }} Static Analysis
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.4', '8.0']
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -29,7 +33,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.4'
+          php-version: ${{ matrix.php-versions }}
           extensions: intl
 
       - name: Use latest Composer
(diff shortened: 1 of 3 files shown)


MGatner:
This needs #4014 before re-testing.

MGatner:
MSSQL database does not seem to be working under PHP 8. @michalsn I don't know if you're around and have any ideas, or can at least point us in the right direction? Or anyone else with MSSQL experience?

Also one cryptic issue that I have not been able to figure out yet:
```
1) CodeIgniter\Session\SessionTest::testCanFlashData
Failed asserting that true is false.
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Session/SessionTest.php:328
```

My hunch is that we should never be calling `Session::start()` consecutively, even for testing, but I have not had a chance to troubleshoot.

michalsn:
@MGatner We probably should switch the [extension](https://github.com/MGatner/CodeIgniter4/blob/php8-workflows/.github/workflows/test-phpunit.yml#L91) to `sqlsrv-beta` since it's the only version that supports PHP 8 right now. But I'm not sure if this gonna work.

paulbalandan:
> Also one cryptic issue that I have not been able to figure out yet:
> 
> ```
> 1) CodeIgniter\Session\SessionTest::testCanFlashData
> Failed asserting that true is false.
> /home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Session/SessionTest.php:328
> ```

I believe this is somehow related to #3974 with regard to the new string number comparison PHP8 now makes internally.

michalsn:
@MGatner I just ran the tests and using `sqlsrv-beta` will do the trick.

Fix for session flash data is ready in #4053.

MGatner:
Okay that's great! I will make the switch after rebasing for your session fix. Good to see you back @michalsn - thanks for the help.

MGatner:
Yay, all passing! Thanks @michalsn. Anyone want to do the honors?

paulbalandan (approved):

PR #4013, opened 2020-12-21
Add Interface to Security Component

mostafakhudair:
While Security Class is one of core system classes and it might be replaced by user Security class, it should have it's own interface for more flexibility.

- [x] Securely signed commits

### DIFF
--- a/system/Security/SecurityInterface.php
+++ b/system/Security/SecurityInterface.php
@@ -1,0 +1,92 @@
+<?php
+
+/**
+ * This file is part of the CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Security;
+
+use CodeIgniter\HTTP\RequestInterface;
+use CodeIgniter\Security\Exceptions\SecurityException;
+
+/**
+ * Expected behavior of a Security.
+ */
+interface SecurityInterface
+{
+	/**
+	 * CSRF Verify
+	 *
+	 * @param RequestInterface $request
+	 *
+	 * @return $this|false
+	 * 
+	 * @throws SecurityException
+	 */
+	public function verify(RequestInterface $request);
+
+	/**
+	 * Returns the CSRF Hash.
+	 *
+	 * @return string|null
+	 */
+	public function getHash(): ?string;
+
+	/**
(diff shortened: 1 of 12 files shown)


system/Security/Exceptions/SecurityException.php
```diff
@@ -11,9 +11,10 @@
 
 namespace CodeIgniter\Security\Exceptions;
 
+use CodeIgniter\Exceptions\ExceptionInterface;
 use CodeIgniter\Exceptions\FrameworkException;
 
-class SecurityException extends FrameworkException
+class SecurityException extends FrameworkException implements ExceptionInterface
```
paulbalandan, line 17: I'm sure FrameworkException already implements ExceptionInterface, so there is absolutely no need to reimplement it here. Please revert changes made to this file.

system/Security/SecurityInterface.php
```diff
@@ -0,0 +1,106 @@
...
+	 *
+	 * @param RequestInterface $request
+	 *
+	 * @return $this|false
+	 * 
+	 * @throws SecurityException
+	 */
+	public function verify(RequestInterface $request);
+
+	//--------------------------------------------------------------------
```
paulbalandan, line 33: Please remove ALL code separator comments used in this file.

paulbalandan (changes requested):

system/Security/SecurityInterface.php
mostafakhudair: Why? all methods in the framework interfaces are separated by comment line,

paulbalandan: _Not all_ interfaces, actually.

Up until now, code separator comments serve no real purpose than to delineate perfectly readable code. These just add unnecessary lines of non functional code.

Although these sort of comments are used frequently does not mean it is empirically right when it adds no value itself. We're just being economic here. Besides, there is an automated fix to remove all existing code separators if the need be, but it is just not the priority now.

lonnieezell: This is a style holdover from CI3 that we're slowly removing as we process files. While it provides a nice visual cue, most decent editors/IDEs today provide a way to do something similar. And if not, the comment block does a good job of it on its own.

samsonasik:
please rebase instead of merge

samsonasik:
I wrote a blog post how to rebase against feature branch https://samsonasik.wordpress.com/2015/09/16/practical-git-4-rebasing-conflicted-task-branch-against-primary-branch/

PR #4014, opened 2020-12-21
Backfill 4.1

MGatner:
**Description**
Applies PHP 8 changes done to `4.1` back into `develop` now that we want to support 8 in `4.0.5`.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Database/Database.php
+++ b/system/Database/Database.php
@@ -41,8 +41,13 @@ class Database
 	 * @return   mixed
 	 * @internal param bool $useBuilder
 	 */
-	public function load(array $params = [], string $alias)
+	public function load(array $params = [], string $alias = '')
 	{
+		if (empty($alias))
+		{
+			throw new InvalidArgumentException('You must supply the parameter: alias.');
+		}
+
 		// Handle universal DSN connection string
 		if (! empty($params['DSN']) && strpos($params['DSN'], '://') !== false)
 		{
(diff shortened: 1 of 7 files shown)


system/Database/Database.php
```diff
@@ -41,8 +41,13 @@ class Database
 	 * @return   mixed
 	 * @internal param bool $useBuilder
 	 */
-	public function load(array $params = [], string $alias)
+	public function load(array $params = [], string $alias = '')
 	{
+		if (empty($alias))
```
samsonasik, line 46: alias === '' is better since empty can check with "0"

system/Router/RouteCollection.php
```diff
@@ -988,8 +989,13 @@ public function presenter(string $name, array $options = null): RouteCollectionI
 	 *
 	 * @return RouteCollectionInterface
 	 */
-	public function match(array $verbs = [], string $from, $to, array $options = null): RouteCollectionInterface
+	public function match(array $verbs = [], string $from = '', $to = '', array $options = null): RouteCollectionInterface
 	{
+		if (empty($from) || empty($to))
```
samsonasik, line 994: use empty string comparison as well?

MGatner: These are all good suggestions, but belonged in the original PR that was cherry-picked. I don't have time to wait for tests again but you (or anyone) could apply these in a new PR after this. Please review for the content at hand.

paulbalandan (approved):

PR #4015, opened 2020-12-21
Improve Postgres testing

MGatner:
**Description**
This is my lazy attempt at circumventing the Postrgres database failures during testing. They seem to be related to a test table hanging around (or maybe concurrency issue?) so I'm dropping the table much sooner.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/tests/system/Database/Live/ForgeTest.php
+++ b/tests/system/Database/Live/ForgeTest.php
@@ -601,8 +601,8 @@ class ForgeTest extends CIDatabaseTestCase
 		elseif ($this->db->DBDriver === 'Postgre')
 		{
 			// Check types
-			$this->assertEquals($fieldsData[0]->type, 'integer');
-			$this->assertEquals($fieldsData[1]->type, 'character varying');
+			$this->assertEquals($fieldsData[0]->type, 'integer', print_r($fieldsData, true));
+			$this->assertEquals($fieldsData[1]->type, 'character varying', print_r($fieldsData, true));
 
 			$this->assertEquals($fieldsData[0]->max_length, 32);
 			$this->assertNull($fieldsData[1]->default);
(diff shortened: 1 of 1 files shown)


paulbalandan (approved):

MGatner:
Thanks for the quick responses @paulbalandan !

kenjis:
@MGatner See <https://github.com/codeigniter4/CodeIgniter4/pull/3993#issuecomment-749271506>.
The order of the column is not the same as the argument of `addField()`.

```php
		$this->forge->addField([
			'id'       => [
				'type'           => 'INTEGER',
				'constraint'     => 11,
				'unsigned'       => false,
				'auto_increment' => true,
			],
			'username' => [
				'type'       => 'VARCHAR',
				'constraint' => 255,
				'unique'     => false,
			],
			'name'     => [
				'type'       => 'VARCHAR',
				'constraint' => 255,
			],
			'active'   => [
				'type'       => 'INTEGER',
				'constraint' => 11,
				'default'    => 0,
			],
		]);
```

MGatner:
Interesting. Sometimes it passes though, so there must be some inconsistency to unspecified returns. The easy solution may be just to order test results before checking - I'm not Postgres expert so I'm unclear if there is a deeper solution to be had here.

kenjis:
@MGatner 
[The SQL](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/Postgre/Connection.php#L294-L297) does not have `order by`, so the output order is changeable.

Should add `ORDER BY "ordinal_position"`?

kenjis:
@MGatner All tests passed! #4016

MGatner:
@kenjis I see `ORDER BY "ordinal_position"` in the file you linked already. Do we need something else?

PR #4018, opened 2020-12-22
$links as private variable

totoprayogo1916:
**Description**
Convert `$links` as private variable

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Pager/PagerRenderer.php
+++ b/system/Pager/PagerRenderer.php
@@ -79,9 +79,9 @@ class PagerRenderer
 	/**
 	 * Links to display
 	 *
-	 * @var array
+	 * @var array|null
 	 */
-	private $links = [];
+	private $links;
 
 	//--------------------------------------------------------------------
 
@@ -289,11 +289,12 @@ class PagerRenderer
 	 */
 	public function links(): array
 	{
-		if ($this->links !== [])
+		if (isset($this->links))
 		{
 			return $this->links;
 		}
 
+		$this->links = [];
 		$uri = clone $this->uri;
 
 		for ($i = $this->first; $i <= $this->last; $i++)
--- a/system/Pager/PagerRenderer.php
+++ b/system/Pager/PagerRenderer.php
@@ -76,6 +76,13 @@ class PagerRenderer
 	 */
 	protected $pageSelector;
 
+	/**
+	 * Links to display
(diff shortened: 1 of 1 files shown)


system/Pager/PagerRenderer.php
```diff
@@ -282,20 +289,18 @@ public function getCurrent(): string
 	 */
 	public function links(): array
 	{
-		$links = [];
-
```
samsonasik, line 286: if it going to happen, it needs return early like this:

if $this->links !== [] 
{
    return $this->links; 
}

samsonasik:
could you provide the use case for make it property?

system/Pager/PagerRenderer.php
```diff
@@ -76,6 +76,13 @@ class PagerRenderer
 	 */
 	protected $pageSelector;
 
+	/**
+	 * Links to dipsplay
```
MGatner, line 80: Typo
```suggestion
	 * Links to display
```

system/Pager/PagerRenderer.php
```diff
@@ -76,6 +76,13 @@ class PagerRenderer
...
+	 *
+	 * @var array
+	 */
+	private $links = [];
```
MGatner, line 84: `$last` technically can be `< 1` so one empty array is a valid return for `links()` already. This should start as `null` and be checked as such for processing.
```suggestion
	 * @var array|null
	 */
	private $links;
```

system/Pager/PagerRenderer.php
```diff
@@ -282,20 +289,23 @@ public function getCurrent(): string
 	 */
 	public function links(): array
 	{
-		$links = [];
+		if ($this->links !== [])
+		{
+			return $this->links;
+		}
```
MGatner, line 296:
```suggestion
		if (isset($this->links))
		{
			return $this->links;
		}

		$this->links = [];
```

MGatner (changes requested):

MGatner:
@totoprayogo1916 did you abandon this?

totoprayogo1916:
> @totoprayogo1916 did you abandon this?

I think yes @MGatner . It might even be complicated.

PR #4019, opened 2020-12-22
Add Interface to Security Component

mostafakhudair:
While Security Class is one of core system classes and it might be replaced by user Security class, it should have it's own interface for more flexibility.

closes #4013 

- [x] Securely signed commits
- [x] Component(s)


paulbalandan (approved):

paulbalandan:
Thank you.

PR #4021, opened 2020-12-22
Move db drivers

mostafakhudair:
Move all database drivers' classes inside system\Database\Drivers folder

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -41,7 +41,7 @@ jobs:
       fail-fast: false
       matrix:
         php-versions: ['7.2', '7.3', '7.4']
-        db-platforms: ['MySQLi', 'Postgre', 'SQLite3', 'Sqlsrv']
+        db-platforms: ['MySQLi', 'Postgre', 'SQLite3', 'SQLSRV']
 
     services:
       mysql:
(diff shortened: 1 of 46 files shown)


samsonasik:
moving class namespace is BC break.

mostafakhudair:
How can we solve this? because Database folder looks messy beside those drivers are used by database configuration and no thing has been changed there

samsonasik:
unfortunately, that will likely only happen in next major version if that still relevant, maybe 5.0

samsonasik:
This is on "Breaking Changes" part in CONTRIBUTING.md:

```
4. Deleting or renaming public classes and interfaces, as well as those not marked as @internal, without prior deprecation notices or not providing fallback solutions will also not be favorably considered.
```

https://github.com/codeigniter4/CodeIgniter4/blob/develop/CONTRIBUTING.md#breaking-changes

mostafakhudair:
I see that, but my thought was those drivers aren't used by users directly and it's handled by framework itself throw driver name 'MySQLi, Postgre, etc...' 
who will use full qualified class name to load the driver?

or maybe we could merge this to 4.1 branch?

samsonasik:
The action is not by guessing, keeping backward compatibility on stable version is always needed.

MGatner:
We could alias and deprecate the old ones but thus far we've tried to keep aliases out of the core as much as possible to prevent confusion / keep it clean. I don't know that the benefits of these nested drivers are worth breaking that.

MGatner:
That said it looks like there is more work in this PR than just nesting the drivers. Could you split it out so we could deliberate over the changes separately?

mostafakhudair:
@MGatner I'll work on separate changes in this PR today, and YES i think it's worth :D I would like to see every line code of this framework clean and organized.

PR #4023, opened 2020-12-23
Rename Sqlsrv driver

mostafakhudair:
Rename `Sqlsrv` driver to `SQLSRV` to compatible with PHP extension name, like other drivers MySQLi, SQLite3 etc...

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -41,7 +41,7 @@ jobs:
       fail-fast: false
       matrix:
         php-versions: ['7.2', '7.3', '7.4']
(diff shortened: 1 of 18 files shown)


michalsn (approved):
Since there was no official release that would include `Sqlsrv`, I'm okay with this change.

I just want to point out that our naming convention doesn't follow the naming convention used in PHP database extensions. Because of this, any other attempts of unification shouldn't happen before v5 (because of BC).

MGatner (approved):

paulbalandan (approved):

samsonasik (approved):

PR #4024, opened 2020-12-23
Add initDriver Method

mostafakhudair:
Add new method `initDriver()` to `Database\Database:class` to initialize database driver

### DIFF
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -31,6 +31,10 @@ class BaseController extends Controller
 
 	/**
 	 * Constructor.
+	 *
+	 * @param RequestInterface  $request
+	 * @param ResponseInterface $response
+	 * @param LoggerInterface   $logger
 	 */
 	public function initController(RequestInterface $request, ResponseInterface $response, LoggerInterface $logger)
 	{
@@ -40,7 +44,6 @@ class BaseController extends Controller
 		//--------------------------------------------------------------------
 		// Preload any models, libraries, etc, here.
 		//--------------------------------------------------------------------
-		// E.g.:
-		// $this->session = \Config\Services::session();
+		// E.g.: $this->session = \Config\Services::session();
 	}
 }


paulbalandan (approved):

iRedds:
Sorry, but this looks ugly. Like php 5.
We know all the drivers. Why not do it like this
```php
public function getDriverManager(array $config) : DriverManagerInterface
{
    switch (strtolower($config['DBDriver'])) {
        case 'mysqli' :
            return new MySQLiDriverManager();
        case 'postgre' :
            return new PostgreDriverManager();
        // etc
        default:
            throw new DatabaseException('invalid driver');
    }
}

public function getUtils(ConnectionInterface $db) : BaseUtils
{
    return $this->driverManager->getUtuls($db);
}

public function getForge(ConnectionInterface $db) : Forge
{
    return $this->driverManager->getForge($db);
}


class MySQLiDriverManager implements DriverManagerInterface
{
    public function getUtils(ConnectionInterface $db) : Utils
    {
        return new \CodeIgniter\Database\MySQLi\Utils($db);
    }
    
    public function getForge(ConnectionInterface $db) : Forge
    {
        return new \CodeIgniter\Database\MySQLi\Forge($db);
    }
    // etc
}

```

mostafakhudair:
Could you explain why would we do this?

I mean why we should create full class with child classes for each driver we have, instead of loading drivers forge/utils directly via one simple function.

iRedds:
It looks sexy ) 
Getters have a more correct return type hint.
Direct strict dependency between database driver and his classes Util, Forge, Result, etc.

Your code is a universal solution to replace existing class name generator.
But I think the whole concept of loading database driver classes is wrong (no only Forge and Utils).

```php
//Dude, we need a result class.
//Challenge accepted!
$resultClass = str_replace('Connection', 'Result', get_class($this));
return new $resultClass($this->connID, $this->resultID);
````

Perhaps my opinion is wrong

**Comment by paulbalandan** [comment]
@mostafakhudair can you remove the merge commit from PR 4029 from your PR's commit?

**Comment by paulbalandan** [comment]
@mostafakhudair your revert also reverted the commit from PR 4029. What I mean earlier is to have the commit removed thru a clean rebase. You can use this flow (assuming you are on the `db-locate-driver-method` branch in your local machine):

```
git reset --hard HEAD~1
git reset HEAD~4
git stash
git rebase develop
git stash pop
git commit -m 'Add initDriver method' -S
git push --force
```

mostafakhudair:
@paulbalandan thank you very much for this steps, but it's look like i messed it up, so I closed this PR with new PR #4040

PR #4044, opened 2020-12-30
Use baseURL to calculate in fillRouteParams() and parseRequestURI()

eakkew:
**Description**
If we deploy ci4 into subfolder,, route function will be broker because path create and parse from relative path which is `/`

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -678,9 +678,10 @@ class IncomingRequest extends Request
 		// parse_url() returns false if no host is present, but the path or query string
 		// contains a colon followed by a number. So we attach a dummy host since
 		// REQUEST_URI does not include the host. This allows us to parse out the query string and path.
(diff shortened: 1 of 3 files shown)


MGatner:
You will need to namespace the App Config class. Preferable do this with a `use` statement at the top of the file, and the same for `URI` (instead of using the full namespace in the method).

eakkew:
> You will need to namespace the App Config class. Preferable do this with a `use` statement at the top of the file, and the same for `URI` (instead of using the full namespace in the method).

fixed it at 683e18be1d6f8f39c1db8bd50c487e71daafa4b9

MGatner:
You did not. You are using `App:class` from the Router namespace but that class does not exist. You need to add below line 18, so it looks like this:
```
use CodeIgniter\Router\Exceptions\RouterException;
use Config\App;
```

paulbalandan:
With the import use statement fixed, it seems proposed changes will not work. Please open another PR if you can look into other options.

PR #4046, opened 2020-12-31
feat: add GROUP_CONCAT to Aggregate Functions

benjaminbellamy:
GROUP_CONCAT(): Return a concatenated string
Super useful!
https://dev.mysql.com/doc/refman/8.0/en/aggregate-functions.html

**Description**
I added GROUP_CONCAT to Aggregate Functions

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -485,6 +485,23 @@ class BaseBuilder
 	{
 		return $this->maxMinAvgSum($select, $alias, 'COUNT');
 	}
+	
+	//--------------------------------------------------------------------
+
+	/**
+	 * Select Group_Concat
+	 *
+	 * Generates a SELECT GROUP_CONCAT(field) portion of a query
+	 *
+	 * @param string $select The field
+	 * @param string $alias  An alias
+	 *
+	 * @return $this
+	 */
+	public function selectGroupConcat(string $select = '', string $alias = '')
+	{
+		return $this->maxMinAvgSum($select, $alias, 'GROUP_CONCAT');
+	}
 
 	//--------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


paulbalandan (changes requested):
We'd need some unit tests to see it in action and user guide updates. Also, please remove the extra whitespace reported by Rector.

samsonasik:
For rector notice, you can fix by run:

vendor/bin/rector process

michalsn:
This function is nice but I don't think it should be included as we only support those that can be used in all databases. As far as I know, this particular aggregate function is only supported in MySQL.

samsonasik:
maybe add to https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/MySQLi/Builder.php , make value in in_array to a property that the property values added in mysql builder.

michalsn:
That seems like a solution but I don't really know what is our general preference for supporting custom database features?

My personal preference is to avoid it because it can cause a lot of pain with maintaining/refactoring code later. Also, the benefits are doubtful for me.

benjaminbellamy:
OK so I guess this was a bad good idea…

najdanovicivan:
GROUP_CONCAT is very useful especially when you need to be very careful with DB performance. 

One of the simplest example is having blog posts with tags table and you want to display all tags for the single blog post. It can be done with 2 queries one to get the post and other to get the post tags

By using GROUP_CONCAT you can get all tags as coma separated values in the same query when post is being loaded.  Imagine now having categories, comments ..... 

Instead of doing multiple queries you get all the data in single query which will be a way faster. 

I'd very much like to have selectGroupConcat() but it should be implemented only in the MySQLi Builder class

michalsn:
Yes, I agree that `GROUP_CONCAT` is useful, but as I said - I'm not sure if we should start adding methods that will work exclusively for only one driver.

And, it's not that there is something that stops us from using `GROUP_CONCAT`, we just don't have method that would do it for us.

But in general I believe we should have some discussion about this, because this is just my personal preference about drivers.

PR #4053, opened 2021-01-02
Fix for deleting session flash data in php8

### LINKED ISSUE #3974: Bug: PHP8 session flush data is not deleted

michalsn:
**Description**
Due to the string-number comparison changes in PHP 8, we need a fix for deleting session flash data.

Closes #3974

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Session/Session.php
+++ b/system/Session/Session.php
@@ -425,9 +425,8 @@ class Session implements SessionInterface
 			{
 				$_SESSION['__ci_vars'][$key] = 'old';
 			}
-			// Hacky, but 'old' will (implicitly) always be less than time() ;)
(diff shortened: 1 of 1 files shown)


MGatner (approved):
Awesome! Glad this wasn't a big problem.

PR #4055, opened 2021-01-02
Refactor Generators

mostafakhudair:
Refactor generators commands to match with standard commands with additional useful options

All generators extends `BaseCommand` using `GeneratorTrait`
Fix Controller extends `PresenterController` methods
Auto generate Entity class on `make:model` with option `-return entity`
for example:
`php spark make:model users -table users -return entity`
will result
`App\Models\Users`
`App\Entities\Users`

Add suffix class name by prepend component name, with `--suffix` option
for example:
`php spark make:controller users`
will result
`App\Controllers\Users`

`php spark make:controller users`
will result
`App\Controllers\UsersController`

Scaffold command removed for now and will be added in a separate PR later

Simplify code and clean unnecessary vars

userguide will be updated in a separated PR

- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Mimes.php
+++ b/app/Config/Mimes.php
@@ -149,6 +149,7 @@ class Mimes
 			'multipart/x-zip',
 		],
 		'rar'   => [
+			'application/vnd.rar',
 			'application/x-rar',
 			'application/rar',
 			'application/x-rar-compressed',


MGatner:
This looks like a good move, but I will need to review it more thoroughly on desktop. Also looks like some static analysis failures, so maybe WIP? We would need to have this done before release in order to avoid breaking changes on the generator class names.

mostafakhudair:
Thanks @MGatner , and I'm working on it, I hope if i finish it today

paulbalandan (commented):
Generally, I like the refactors made. I'll do a full review tomorrow but my initial concern is on programmatic calls. `CLI::getOption()` will return nothing when the command is not called via cli, that is why we need to check array_key_exists on the $params. This way, the options can be tested via unit tests. So, please do not remove those array_key_exists checks already in place.

Another one is on the `--force` option handling. This option is meant to allow overwriting files but you made it impossible to overwrite even if `--force` is given. It is like you are offering something and then when they grab that you refuse to give that right away. I find it quite illogical.

mostafakhudair:
file overwritten fine with ``--force``  option
I just put force check just after file is replaced 
Screenshot from 2021-01-02 17-02-54

paulbalandan:
Alright, then. I maybe got quite confused with that block.

Please then re-include the access to `$params` for the options so it will be possible to use `command('make:command foo --bare)` and others programmatically. Also, please check the changes you've made to the unit tests and inquire why it runs "forever" and stuck at 4%. Maybe you inadvertently made a recursion error or something.

system/Commands/Generators/SessionMigrationGenerator.php
```diff
@@ -0,0 +1,101 @@
...
+	 * @var string
+	 */
+	protected $name = 'session:migration';
+
+	/**
+	 * the Command's short description
+	 *
+	 * @var string
+	 */
+	protected $description = '[DEPRECATED] Generates the migration file for database sessions, use "make:migration" instead.';
```
paulbalandan, line 46: The deprecation is uncalled for. Calling `session:migration` is much faster than calling `make:migration` plus options to make it compatible to the current result. Plus the options of `make:migration` are not compatible with `session:migration`. The latter does not have the options `--table` and `--dbgroup` so directing users to use that is a deadend.

system/Commands/Generators/SessionMigrationGenerator.php
```diff
@@ -0,0 +1,101 @@
...
+	 * Performs the necessary replacements.
+	 *
+	 * @param string $class
+	 *
+	 * @return string
+	 */
+	protected function setReplacements(string $class): string
+	{
+		$table           = CLI::getOption('table');
+		$data['DBGroup'] = is_string(CLI::getOption('dbgroup')) ? CLI::getOption('dbgroup'): 'default';
```
paulbalandan, line 91: Handling for programmatic calls?

system/Commands/Generators/CommandGenerator.php
```diff
@@ -0,0 +1,121 @@
...
+	 *
+	 * @var array
+	 */
+	protected $options = [
+		'--command'   => 'The command name, Defaults: "command:name"',
+		'--group'     => 'The command group, Defaults: [basic -> "CodeIgniter", generator -> "Generators"].',
+		'--type'      => 'The command type. Options [basic, generator], Defaults: "basic".',
+		'--namespace' => 'Set root namespace. Defaults: "APP_NAMESPACE".',
+		'--force'     => 'Force overwrite existing file.',
+		'--prevent'   => 'Prevent renaming class as framework standards format (e.g. UserCommand => User).',
```
paulbalandan, line 73: I don't think we have that convention strictly followed with the framework itself. Take a look at current CLI commands. Those do not mostly have `Command` suffixes. For others (controllers, migrations, etc.), there is no convention either for suffixing the component name. For controllers if you are using `$routes->setAutoRoute(true);` you would definitely not suffix `Controller` in the class name.

For this and other commands, I propose that suffixing the component name is default to `false`. If users would want their classes suffixed, then add the option to the command call. I guess from `--prevent` to `--suffix` to add the suffix.

paulbalandan (commented):
These review comments are not yet exhaustive but my initial comments only. Most of the comments are on the proper programmatic call handling.

paulbalandan:
I guess the 360 minute testing was caused by an executed call to `CLI::prompt` which keeps PHPUnit waiting for a keyboard input.

mostafakhudair:
@paulbalandan  Thanks for this helpful review, give me few hours to be in home and i'll fix all of this mistakes

system/Commands/Generators/CommandGenerator.php
mostafakhudair: --suffix brilliant suggest, Thank you

MGatner:
Nice work team! I came to see where this was at, I like seeing all the activity.

system/CLI/GeneratorTrait.php
```diff
@@ -0,0 +1,285 @@
...
+	public function execute(array $params): void
+	{
+		if (CLI::getOption('namespace') === 'CodeIgniter')
+		{
+			CLI::error(lang('CLI.generator.invalidNamespace'));
+			CLI::newLine();
+
+			return;
+		}
+
```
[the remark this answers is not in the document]
mostafakhudair: Using one line warning alongside to confirmation  or separate it in two lines?

Screenshot from 2021-01-04 21-37-08

system/Commands/Generators/SessionMigrationGenerator.php
mostafakhudair: I have added option `-session` to `make:migration` to build `ci_session_migration_file` with two question from table name and database group with default values: 'ci_session', 'default'

mostafakhudair:
`CLI::promot` is ignored now but phpunit still stuck, any help on this? @paulbalandan

tests/system/Commands/GeneratorsTest.php
```diff
@@ -0,0 +1,98 @@
...
+		$this->assertStringContainsString('Error while creating file: ', CITestStreamFilter::$buffer);
+
+		chmod(APPPATH . 'Filters', 0755);
+	}
+
+	public function testGenerateFailsOnUndefinedNamespace()
+	{
+		try
+		{
+			command('make:model cars -n CodeIgnite');
```
sfadschm, line 77: This is where looping starts when executing test on Windows locally one by one.

tests/system/Commands/ModelGeneratorTest.php
```diff
@@ -0,0 +1,113 @@
...
+		$file = APPPATH . 'Entities/User.php';
+		$this->assertFileExists($file);
+		$dir = dirname($file);
+		is_file($file) && unlink($file);
+		is_dir($dir) && rmdir($dir);
+	}
+
+	public function testGenerateModelWithOptionSuffix()
+	{
+		command('make:model user -suffix');
```
sfadschm, line 107: Here it also starts to loop infinitely.

sfadschm:
@mostafakhudair I ran the test locally one by one and indicated where looping happens. Seems to be related to the `make:model` command. Hope this helps.

tests/system/Commands/ScaffoldGeneratorTest.php
```diff
@@ -41,11 +37,11 @@ protected function getFileContents(string $filepath): string
 	public function testCreateScaffoldProducesManyFiles()
 	{
 		command('make:scaffold people');
```
sfadschm, line 39: This also starts looping.

mostafakhudair:
@sfadschm Thanks for your time to test this, but could you tell me why specially make:model looping

system/Commands/Generators/ModelGenerator.php
```diff
@@ -0,0 +1,137 @@
...
+	 */
+	protected function prepare(string $class): string
+	{
+		$table   = $this->getOption('table');
+		$DBGroup = $this->getOption('dbgroup') ?? 'default';
+		$return  = $this->getOption('return') ?? 'array';
+
+		if (! is_string($table))
+		{
+			$table = CLI::prompt(lang('CLI.generator.tableName'), null, 'required'); // @codeCoverageIgnore
```
sfadschm, line 105: @mostafakhudair This is where it stops as it expects an input, like paul suspected.

mostafakhudair: and how to avoid testing this? does codeCoverageIgnore isn't enough?

sfadschm: `codeCoverageIgnore ` only tells the coverage script that the marked line should not count as "untested".
If you want to stay that way, you should update the tests such that the `make:model` command always gets passed a table name directly.

sfadschm: Maybe you could also add an `echo` command to the tests to simulate user input, but I'm not familiar with that.
Maybe @paulbalandan knows more?

sfadschm: You need to add the default table name to all calls to `make:model` in `ModelGeneratorTest.php`.
Maybe try running the tests locally before pushing your changes here.
You can use `phpunit --no-coverage --filter ModelGeneratorTest`.
You can likely ignore all failures that get displayed (at least on Windows) but you will see whether it gets stuck or not.

sfadschm:
Btw. what happened to the model table defaulting to pluralized class name?
https://github.com/codeigniter4/CodeIgniter4/blob/453af979ad2e026559312899fc5fcbfe93646b8d/system/Commands/Generators/CreateModel.php#L127-L142

system/Commands/Generators/MigrationGenerator.php
```diff
@@ -0,0 +1,134 @@
...
+	protected function prepare(string $class): string
+	{
+		$data = [
+			'session' => false,
+			'matchIP' => false,
+		];
+
+		if ($this->getOption('session'))
+		{
+			$table = CLI::prompt(lang('CLI.generator.tableName')); // @codeCoverageIgnore
```
sfadschm, line 108: This is the same problem as with `make:model`. Also on line 111.

system/Commands/Generators/ModelGenerator.php
mostafakhudair: Thank you

system/Commands/Generators/MigrationGenerator.php
mostafakhudair: I think table name should be equal to model name, most of tables name aren't plural

sfadschm: I guess thats kind of personal preference, but personally, I'd go with pluralized forms.
Imagine having a database for your car shop, then you would have a `cars` model and a `car` entity. As every row in the database represents a single `car`, the table itself holds multiple _cars_, hence I would use that as name, too.
Thinking about it, this is also more consistent with the principle, that each model should represent a single database table.

mostafakhudair:
Test ran well on local machine, but still stucked at 6% in here

sfadschm:
Thats still `MigrationGeneratorTest`. See comment somewhere above (kind of hard to follow the changes with all the force pushes :D).
So check the MigrationGenerator for required `prompts` and you should be fine.

mostafakhudair:
I already fixed model and migrate prompt before i push changes

system/Commands/Generators/MigrationGenerator.php
```diff
@@ -0,0 +1,125 @@
...
+	public function run(array $params)
+	{
+		$this->component = 'Migration';
+		$this->directory = 'Database\Migrations';
+		$this->template  = 'migration.tpl.php';
+
+		if ($this->getOption('session'))
+		{
+			$params[0] = '_create_ci_session_table';
+		}
```
sfadschm, line 87: Not sure what the exact problem is, but in `testGenerateMigrationWithOptionSession` this condition is not fulfilled. 
If you do `make:migration test -session` it works (but the test fails of course).
So there might be a proplem with recognizing the `--session` option with `getOption`.

For some reason `$this->getOption('session')` is `bool(true)` in the `prepare` method but it is `null` in the `run` method.

sfadschm: The issue seems to be that when calling `$this->getOptions`, the `$params` array is not yet copied to `$this->params` as that only happens in the `execute` function. I feel like that call should thus move to `prepare` or you just directly check the `$params` array in the run function for the `session` index.

mostafakhudair: Oh what a suffering from that problem, looks like everything is working fine now, I appreciate your review and help Thank you @sfadschm

system/Commands/Generators/ModelGenerator.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @return string
+	 */
+	protected function prepare(string $class): string
+	{
+		$table   = $this->getOption('table');
+		$DBGroup = $this->getOption('dbgroup');
+		$return  = $this->getOption('return') ?? 'array';
+
+		$table   = is_string($table) ? $table : strtolower($this->params[0] ?? CLI::getSegment(2));
+		$DBGroup = is_string($DBGroup) ? $DBGroup : 'default';
```
paulbalandan, line 104:
```suggestion
		$namespace = trim(implode('\\', array_slice(explode('\\', $class), 0, -1)), '\\');
		$baseClass   = str_replace($namespace . '\\', '', $class);

		$table   = $this->getOption('table');
		$DBGroup = $this->getOption('dbgroup');
		$return  = $this->getOption('return') ?? 'array';

		$table   = is_string($table) ? $table : strtolower(plural($baseClass));
		$DBGroup = is_string($DBGroup) ? $DBGroup : 'default';
```

system/Commands/Generators/ModelGenerator.php
```diff
@@ -0,0 +1,136 @@
...
+
+		if (! in_array($return, ['array', 'object', 'entity'], true))
+		{
+			$return = CLI::prompt(lang('CLI.generator.returnType'), ['array', 'object', 'entity'], 'required'); // @codeCoverageIgnore
+			CLI::newLine();
```
paulbalandan, line 109:
```suggestion
			// @codeCoverageIgnoreStart
			$return = CLI::prompt(lang('CLI.generator.returnType'), ['array', 'object', 'entity'], 'required');
			CLI::newLine();
			// @codeCoverageIgnoreEnd
```

system/Commands/Generators/ModelGenerator.php
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		if ($return === 'entity')
+		{
+			$return = str_replace('Models', 'Entities', $class);
+
+			if (strpos($return, 'Model'))
+			{
+				$return = str_replace('Model', 'Entity', $return);
+			}
```
paulbalandan, line 119: You are forcing here the suffix option. At least use the default setting of no suffixes.
```suggestion
			if ($pos = strrpos($return, 'Model'))
			{
				// Strip 'Model' from name
				$return = substr($return, 0, $pos);
			}
```

system/Commands/Generators/ModelGenerator.php
```diff
@@ -0,0 +1,136 @@
...
+
+			$this->call('make:entity', array_merge(
+				[$this->params[0] ?? CLI::getSegment(2)], $this->params ?? CLI::getOptions()
+			));
```
paulbalandan, line 123:
```suggestion
			$this->call('make:entity', array_merge(
				[$baseClass],
				$this->params
			));
```

system/Commands/Generators/SeederGenerator.php
```diff
@@ -0,0 +1,86 @@
...
+		'name' => 'The seeder class name.',
+	];
+
+	/**
+	 * The Command's Options
+	 *
+	 * @var array
+	 */
+	protected $options = [
+		'--namespace' => 'Set root namespace. Defaults: "APP_NAMESPACE".',
```
paulbalandan, line 68:
```suggestion
		'--namespace' => 'Set root namespace. Default: "APP_NAMESPACE".',
```

system/Commands/Generators/SessionMigrationGenerator.php
```diff
@@ -0,0 +1,110 @@
...
+	 * @var string
+	 */
+	protected $name = 'session:migration';
+
+	/**
+	 * The Command's Description
+	 *
+	 * @var string
+	 */
+	protected $description = '[DEPRECATED] Generates the migration file for database sessions, use "make:migration -session" instead.';
```
paulbalandan, line 46:
```suggestion
	protected $description = '[DEPRECATED] Generates the migration file for database sessions, use "make:migration --session" instead.';
```

system/Commands/Generators/SessionMigrationGenerator.php
```diff
@@ -0,0 +1,110 @@
...
+	protected $usage = 'session:migration [options]';
+
+	/**
+	 * The Command's Options
+	 *
+	 * @var array
+	 */
+	protected $options = [
+		'-t' => 'Supply a table name.',
+		'-g' => 'Database group to use. Defaults: "default".',
```
paulbalandan, line 62: Add these options to `make:migration`

system/Commands/Generators/SessionMigrationGenerator.php
```diff
@@ -0,0 +1,110 @@
...
+	 *
+	 * @return string
+	 */
+	protected function prepare(string $class): string
+	{
+		$table           = $this->getOption('t');
+		$data['DBGroup'] = $this->getOption('g');
+		$data['matchIP'] = config('App')->sessionMatchIP ?? false;
+
+		$table           = is_string($table) ? $table : 'ci_session';
```
paulbalandan, line 93:
```suggestion
		$table           = is_string($table) ? $table : 'ci_sessions';
```

system/Commands/Generators/SessionMigrationGenerator.php
```diff
@@ -0,0 +1,110 @@
...
+namespace CodeIgniter\Commands\Generators;
+
+use CodeIgniter\CLI\BaseCommand;
+use CodeIgniter\CLI\CLI;
+use CodeIgniter\CLI\GeneratorTrait;
+
+/**
+ * Generates a migration file for database sessions.
+ *
+ * @deprecated Use make:migration instead.
```
paulbalandan, line 21:
```suggestion
 * @deprecated Use make:migration instead.
 * @codeCoverageIgnore
```

tests/_support/Commands/Foobar.php
```diff
@@ -1,11 +0,0 @@
...
-
-use Config\App;
-use CodeIgniter\CLI\CLI;
-
-return [
-   'foo' => 'The command will use this as foo.',
-   'bar' => 'The command will use this as bar.',
-   'baz' => 'The baz is here.',
-   'bas' => CLI::color('bas', 'green') . (new App())->baseURL,
-];
```
paulbalandan, line 11: Reinstate this test. This is for the configurable sort imports testing.

tests/_support/Commands/LanguageCommand.php
```diff
@@ -1,37 +0,0 @@
-<?php
-
-namespace Tests\Support\Commands;
-
-use CodeIgniter\CLI\GeneratorCommand;
-
-class LanguageCommand extends GeneratorCommand
```
paulbalandan, line 7: Reinstate this test file for the configurable sort imports test. You cannot just delete a test just to make the tests succeed.

tests/system/Commands/ConfigurableSortImportsTest.php
```diff
@@ -1,53 +0,0 @@
-<?php
-
-namespace CodeIgniter\Commands;
-
-use CodeIgniter\Test\CIUnitTestCase;
-use CodeIgniter\Test\Filters\CITestStreamFilter;
-
-class ConfigurableSortImportsTest extends CIUnitTestCase
```
paulbalandan, line 8: Reinstate this test. I am using this feature in my library.

tests/system/Commands/SessionsCommandsTest.php
```diff
@@ -29,23 +31,11 @@ public function tearDown(): void
...
-
-	public function testOverriddenCreateMigrationCommand()
-	{
-		command('session:migration -t mygoodies');
-		$result = CITestStreamFilter::$buffer;
-
-		// make sure we end up with a migration class in the right place
-		$this->assertStringContainsString('Created file:', $result);
-		$this->assertStringContainsString('_CreateMygoodiesTable.php', $result);
+		$this->assertStringContainsString('_CreateCiSessionTable.php', CITestStreamFilter::$buffer);
```
paulbalandan, line 38: Reinstate the ability to override the sessions table and also this test. That feature was not introduced by Generator Command. It was before so removing that is a BC break.

system/CLI/GeneratorTrait.php
```diff
@@ -0,0 +1,318 @@
...
+	 */
+	protected function getOption(string $name)
+	{
+		if (! array_key_exists($name, $this->params))
+		{
+			return CLI::getOption($name);
+		}
+
+		// If the param doesn't have a value, simply return TRUE
+		// so they know it was set, otherwise return the actual value.
```
paulbalandan, line 315:
```suggestion
```

paulbalandan (changes requested):
I saw a lot of comments here. Some are on the deviation from the original logic of `GeneratorCommand`, others on the design, while others is removing existing functionalities. I saw some changes removing existing functions even before `GeneratorCommand` was existing, so technically that removal is a BC break for existing users. I'll have to put the BC break tag until those BCs are resolved.

Also, don't dismiss right away a comment without providing an explanation. If you do not agree, tell us so.

system/CLI/GeneratorTrait.php
```diff
@@ -0,0 +1,318 @@
...
+		{
+			$class = CLI::prompt(lang('CLI.generator.className'), null, 'required'); // @codeCoverageIgnore
+			CLI::newLine();
+		}
+
+		helper('inflector');
+
+		$component = strtolower(singular($this->component));
+		$class     = strtolower($class);
+		$class     = strpos($class, $component) ? str_replace($component, ucfirst($component), $class) : $class;
```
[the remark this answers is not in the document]
mostafakhudair: This needed because without it, if someone append component name to class name `make:controller blogcontroller` it will return `Blogcontroller`, i used this trick to check if the class name includes component name then capitalize component name `BlogController`

system/CLI/GeneratorTrait.php
```diff
@@ -0,0 +1,318 @@
...
+	private function qualifyClassName(): string
+	{
+		// Gets the class name from input.
+		$class = $this->params[0] ?? CLI::getSegment(2) ?? '';
+
+		if (empty($class))
+		{
+			$class = CLI::prompt(lang('CLI.generator.className'), null, 'required'); // @codeCoverageIgnore
+			CLI::newLine();
+		}
```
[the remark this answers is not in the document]
mostafakhudair: I'm not forcing it! I'm using your old code getClassName but just in 1 line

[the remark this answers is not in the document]
paulbalandan: Yes, I'm aware of that. `GeneratorCommand::getClassName()` does not have any prompts because class name was conceived to be optional. If name is needed, then child classes can just provide a prompt. For other generators, for example language, class names are not required. So it is better to have the trait treat class names as optional at first and have classes using it to make them required when needed.

system/CLI/GeneratorTrait.php
```diff
@@ -0,0 +1,318 @@
...
+	/**
+	 * Builds the contents for class being generated, doing all
+	 * the replacements necessary, and alphabetically sorts the
+	 * imports for a given template.
+	 *
+	 * @param string $class
+	 *
+	 * @return string
+	 */
+	private function buildContent(string $class): string
```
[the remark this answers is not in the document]
mostafakhudair: Turned it on back

system/CLI/GeneratorTrait.php
```diff
@@ -0,0 +1,318 @@
...
+	{
+		// Retrieves the namespace part from the fully qualified class name.
+		$namespace = trim(implode('\\', array_slice(explode('\\', $class), 0, -1)), '\\');
+
+		array_push($search, '<@php', '{namespace}', '{class}');
+		array_push($replace, '<?php', $namespace, str_replace($namespace . '\\', '', $class));
+
+		return str_replace($search, $replace, $this->renderTemplate($data));
+	}
+
```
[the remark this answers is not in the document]
mostafakhudair: Added

system/CLI/GeneratorTrait.php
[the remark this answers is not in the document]
paulbalandan: Actually, your solution can be acceptable too as most generators require class names. For the sake of other generators soon to be made which do not require class names, this prompt will become a problem. To get rid of this problem, I suggest extracting this part to its method getClassName(). This way if a future generator does not need class names with prompts it would just need to override that method, instead of overriding the whole qualifyClassName.

system/Commands/Generators/CommandGenerator.php
```diff
@@ -0,0 +1,126 @@
...
+			$group = 'Generators';
+		}
+
+		return $this->parseTemplate($class, [
+			'{group}',
+			'{command}',
+		], [
+			$group,
+			$command,
+		], ['type' => $type]);
```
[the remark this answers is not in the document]
mostafakhudair: What about this for one line code?

```suggestion
		return $this->parseTemplate(
			$class, ['{group}', '{command}'], [$group, $command], ['type' => $type]
		);
```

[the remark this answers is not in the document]
paulbalandan: As long as it is readable and CodeSniffer won't mess with array formatting, I'm good with one liners. CodeSniffer is really a pain in formatting arrays nowadays.

system/Commands/Generators/ModelGenerator.php
```diff
@@ -0,0 +1,136 @@
...
+		'name' => 'The model class name.',
+	];
+
+	/**
+	 * The Command's Options
+	 *
+	 * @var array
+	 */
+	protected $options = [
+		'--table'     => 'Supply a table name.',
```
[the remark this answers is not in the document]
mostafakhudair: I had a point on this, but it doesn't matter,, pluralized...

system/CLI/GeneratorTrait.php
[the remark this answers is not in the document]
mostafakhudair: Can we use a property something like . boolean $hasClassName = true, and change it status with method in a child class whether we need that?

system/CLI/GeneratorTrait.php
```diff
@@ -0,0 +1,318 @@
...
+		// Gets the root namespace from input.
+		$root = trim(str_replace('/', '\\', $this->getOption('namespace') ?? APP_NAMESPACE), '\\');
+
+		if (strncmp($class, $root, strlen($root)) === 0)
+		{
+			return $class;
+		}
+
+		return $root . '\\' . $this->directory . '\\' . str_replace('/', '\\', $class);
```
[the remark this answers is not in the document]
mostafakhudair: I created `make:config` to generate config file, but i'm wondering what is language generator and what it should do?

system/CLI/GeneratorTrait.php
```diff
@@ -0,0 +1,318 @@
...
+
+		// Check if the namespace is actually defined and we are not just typing gibberish.
+		$base = Services::autoloader()->getNamespace($root);
+
+		if (! $base = reset($base))
+		{
+			CLI::error(lang('CLI.namespaceNotDefined', [$root]));
+			CLI::newLine();
+			
+			return '';
```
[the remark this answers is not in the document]
mostafakhudair: throwing exception in command line looks very messy, but one line error code looks better, just it

system/Commands/Generators/ModelGenerator.php
mostafakhudair: your suggestion will not append `Entity` string to $returnType property, but the entity name will be suffixed with Entity, i will try to find another way to avoid this issue

system/CLI/GeneratorTrait.php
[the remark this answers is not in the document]
paulbalandan: Ok. We can go with that. Is that method something like `setHasClassName(bool $hasClassName)`? If the primary purpose is to turn on/off the `$hasClassName` property, it can live with this trait. If you'll be implementing that boolean flag, I'll be expecting it in the `if` statement of the prompt.
```php
if ($this->hasClassName && is_null($class))
{
    // ...
}
```

system/CLI/GeneratorTrait.php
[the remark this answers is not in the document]
paulbalandan: This `make:config` and `make:language` are most useful for CodeIgniter4 modules. Let's say `myth/auth`. It has it own config and language files. A dev using myth/auth may want to copy those files into his `app` folder to make changes or own variation. He can do so by typing in the spark command.
Check out this config and language generators I made using `GeneratorCommand` previously.

[PublishConfigCommand](https://github.com/paulbalandan/liaison-revision/blob/develop/src/Commands/PublishConfigCommand.php)
[PublishLanguageCommand](https://github.com/paulbalandan/liaison-revision/blob/develop/src/Commands/PublishLanguageCommand.php)

system/CLI/GeneratorTrait.php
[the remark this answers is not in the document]
paulbalandan: OK. Just note this change in the user guide.

system/Commands/Generators/ModelGenerator.php
paulbalandan: The suffixing of the Entity via the model generator should be configured by the `--suffix` option too.

system/CLI/GeneratorTrait.php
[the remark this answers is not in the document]
mostafakhudair: I think `make:config` and `make:language` are kind of copy files and doesn't belongs to `Generators`, and it fit more 
 with `Publisher` `publich:config` and `publish:language`, so can we merge this for now? and then maybe we create publisher command version?

PR #4056, opened 2021-01-03
Solidate mime type guessing.

sfadschm:
This PR continues the discussion in #4054.

**Description**
- Removed the `application/octet-stream` mime type from non-binary file extensions (e.g., `csv`, `srt`, `vtt`, `pdf`).
- Add comment to hint towards enabling the `fileinfo` extension.
- Modify the `guessExtensionFromType` method to
  - return `null` if an extension is proposed but the mime type does not match the extension.
  - reverse search the `$mimes` array only if no extension was proposed.

**Things to do**
- Add not to Changelog/Update Guide to alert developers to check their mime type handling (and possibly also to enable `fileinfo` again, as several recent bug reports are potentially solved by this).

**Note**
Symphony is using this repo for their mime type list:
https://github.com/jshttp/mime-db
https://cdn.jsdelivr.net/gh/jshttp/mime-db@v1.44.0/db.json

This might be worth a look and might offer a consistent source for mime type detection.
Symphony automatically updates its mime type list via this script:
https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/Mime/Resources/bin/update_mime_types.php

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
 - `CodeIgniter\Config\Config` is now deprecated in favor of `CodeIgniter\Config\Factories::config()`
 - HTTP Layer Refactor: Numerous deprecations have been made towards a transition to a PSR-compliant HTTP layer. [See the User Guide](user_guide_src/source/installation/upgrade_405.rst)
 
+**Mime Type Detection**
+
+- `Config\Mimes::guessExtensionFromType` now only reverse searches the `$mimes` array if no extension is proposed (i.e., usually not for uploaded files).
+- The fallback values of `UploadedFile->getExtension()` and `UploadedFile->guessExtension()` have been changed. `UploadedFile->getExtension()` now returns `$this->getClientExtension()` instead of `''`; `UploadedFile->guessExtension()` now returns `''` instead of `$this->getClientExtension()`.
+These changes increase security when handling uploaded files as the client can no longer force a wrong mime type on the application. However, these might affect how file extensions are detected in your application.
+
 ## [v4.0.4](https://github.com/codeigniter4/CodeIgniter4/tree/v4.0.4) (2020-07-15)
 
 [Full Changelog](https://github.com/codeigniter4/CodeIgniter4/compare/v4.0.3...v4.0.4)
--- a/app/Config/Mimes.php
+++ b/app/Config/Mimes.php
@@ -12,6 +12,9 @@ namespace Config;
  * the most common one should be first in the array to aid the guess*
  * methods. The same applies when more than one mime-type exists for a
  * single extension.
+ *
+ * When working with mime types, please make sure you have the ´fileinfo´
+ * extension enabled to reliably detect the media types.
  */
 
 class Mimes
@@ -33,7 +36,6 @@ class Mimes
 			'text/csv',
 			'text/x-comma-separated-values',
 			'text/comma-separated-values',
-			'application/octet-stream',
 			'application/vnd.ms-excel',
 			'application/x-csv',
 			'text/x-csv',
@@ -69,7 +71,6 @@ class Mimes
 			'application/pdf',
 			'application/force-download',
 			'application/x-download',
-			'binary/octet-stream',
 		],
 		'ai'    => [
 			'application/pdf',
@@ -462,12 +463,10 @@ class Mimes
 		'srt'   => [
 			'text/srt',
 			'text/plain',
-			'application/octet-stream',
 		],
 		'vtt'   => [
 			'text/vtt',
 			'text/plain',
-			'application/octet-stream',
 		],
 		'ico'   => [
 			'image/x-icon',
@@ -509,11 +508,20 @@ class Mimes
 
 		$proposedExtension = trim(strtolower($proposedExtension));
 
-		if ($proposedExtension !== '' && array_key_exists($proposedExtension, static::$mimes) && in_array($type, is_string(static::$mimes[$proposedExtension]) ? [static::$mimes[$proposedExtension]] : static::$mimes[$proposedExtension], true))
+		if ($proposedExtension !== '')
 		{
-			return $proposedExtension;
+			if(array_key_exists($proposedExtension, static::$mimes) && in_array($type, is_string(static::$mimes[$proposedExtension]) ? [static::$mimes[$proposedExtension]] : static::$mimes[$proposedExtension], true))
+			{
+				// The detected mime type matches with the proposed extension.
+				return $proposedExtension;
+			}
+
+			// An extension was proposed, but the media type does not match the mime type list.
+			return null;
 		}
 
+		// Reverse check the mime type list if no extension was proposed.
+		// This search is order sensitive!
 		foreach (static::$mimes as $ext => $types)
 		{
 			if ((is_string($types) && $types === $type) || (is_array($types) && in_array($type, $types, true)))
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,9 @@
 		"rector/rector": "^0.8",
 		"squizlabs/php_codesniffer": "^3.3"
 	},
+	"suggest": {
+		"ext-fileinfo": "Improves mime type detection for files"
+	},
 	"config": {
 		"optimize-autoloader": true,
 		"preferred-install": "dist",
--- a/system/HTTP/Files/UploadedFile.php
+++ b/system/HTTP/Files/UploadedFile.php
@@ -308,24 +308,29 @@ class UploadedFile extends File implements UploadedFileInterface
 	 * Overrides SPLFileInfo's to work with uploaded files, since
 	 * the temp file that's been uploaded doesn't have an extension.
 	 *
-	 * Is simply an alias for guessExtension for a safer method
-	 * than simply relying on the provided extension.
-	 * Additionally it will return clientExtension in case if there are
-	 * other extensions with the same mime type.
+	 * This method tries to guess the extension from the files mime
+	 * type but will return the clientExtension if it fails to do so.
+	 *
+	 * This method will always return a more or less helpfull extension
+	 * but might be insecure if the mime type is not machted. Consider
+	 * using guessExtension for a more safe version.
 	 */
 	public function getExtension(): string
 	{
-		return $this->guessExtension();
+		return $this->guessExtension() ?: $this->getClientExtension();
 	}
 
 	/**
-	 * Attempts to determine the best file extension.
+	 * Attempts to determine the best file extension from the file's
+	 * mime type. In contrast to getExtension, this method will return
+	 * an empty string if it fails to determine an extension instead of
+	 * falling back to the unsecure clientExtension.
 	 *
 	 * @return string
 	 */
 	public function guessExtension(): string
 	{
-		return Mimes::guessExtensionFromType($this->getMimeType(), $this->getClientExtension()) ?? $this->getClientExtension();
+		return Mimes::guessExtensionFromType($this->getMimeType(), $this->getClientExtension()) ?? '';
 	}
 
 	//--------------------------------------------------------------------
--- a/user_guide_src/source/changelogs/v4.0.5.rst
+++ b/user_guide_src/source/changelogs/v4.0.5.rst
@@ -12,6 +12,8 @@ Enhancements:
 - New URL helper function ``url_is()`` which allows you to check the current URL to see if matches the given string.
 - Services now have their config parameters strictly typehinted. This will ensure no one will pass a different config instance. If you need to pass a new config with additional properties, you need to extend that particular config.
 - Support for setting SameSite attribute on Session and CSRF cookies has been added. For security and compatibility with latest browser versions, the default setting is ``Lax``.
+- Guessing file extensions from mime type in ``Config\Mimes::guessExtensionFromType`` now only reverse searches the ``$mimes`` array if no extension is proposed (i.e., usually not for uploaded files).
+- The getter functions for file extensions of uploaded files now have different fallback values (``$this->getClientExtension()`` for ``UploadedFile->getExtension()`` and ``''`` for ``UploadedFile->guessExtension()``). This is a security fix and makes the process less dependent on the client.
 
 Changes:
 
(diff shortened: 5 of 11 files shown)


sfadschm:
I will leave the tests failing for now until further discussion.

The current issue is the following:
- A file is uploaded with filename `xyz.txt` and mime type `text/csv`.
- The new implementation checks the proposed extension (`txt`) in the `$mimes` array and finds that `text/csv` does not match the configured mime type for the `txt` extension. Thus it returns `null` and `UploadedFile->guessExtension()` fallsback to `$this->getClientExtension()`.

For my opinion, I'm not sure if the current result is desired. The extension fallback implemented in #2128 seems useful, but right now it will still allow us to upload a file whose extension does not match the configured mime types, but the file will still keep its extension.

michalsn:
Thanks, @sfadschm! I will try to look into it this week.

michalsn:
I like the changes you made here. Now the problems you mentioned:

1. This gonna be fixed by #4085 I believe
2. I think that `UploadedFile->guessExtension()` should fallback to empty string. And `UploadedFile->getExtension()` should fallback to the `$this->getClientExtension()`. This way `guessExtension()` can return an empty string when there is no match and `getExtension()` most likely always gonna return something more or less useful.

Also, there is one more thing that should be fixed. `Validation\FileRules::ext_in()` - this one should use `$file->guessExtension()` instead of `$file->getExtension()`. Simply for security reasons.

Some work on the mime types list can be done later, as it may be more time-consuming and more problematic to handle. So, if you would be up for it, we can for sure makes some change in this area, but in a different PR.

Of course, we should mention these changes made in `guessExtension()` and `getExtension()` in the changelog, but this is more like a bug/security fix for me.

sfadschm:
I think youre right with the first point. Exprected test results will be adapted when the other PR is through.
I also agree with the second point. It sounds right to me, that guessing the extension can just return nothing (`null` would be nicer, but we can't change the method signature) and getting the extension returns the best match.
This should definitely be mentioned in the update changelog. Can I do this in this PR or is this up tp the mainteners after merge?

With the above changes I also implemented your suggestion as it fits the target of this PR to make mime guessing a little more "secure" by not relying on the client unless people intentionally use the client functions.

sfadschm:
I just saw that the return types of extension methods in `File` and `UploadedFile` differ.

For `File` guessing can return `null`:
https://github.com/codeigniter4/CodeIgniter4/blob/d09a546284462fc39db8dabf9a5e42be6458f8fe/system/Files/File.php#L101
For `UploadedFile` it can only return an `empty string` on failure:
https://github.com/codeigniter4/CodeIgniter4/blob/d09a546284462fc39db8dabf9a5e42be6458f8fe/system/HTTP/Files/UploadedFile.php#L326

The `getExtension` methods of both classes (in case of `File`, this is the `SplFileInfo->getExtension` method) strictly return strings.
Does this make sense and could it be changed without BC?

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
 - `CodeIgniter\Config\Config` is now deprecated in favor of `CodeIgniter\Config\Factories::config()`
 - HTTP Layer Refactor: Numerous deprecations have been made towards a transition to a PSR-compliant HTTP layer. [See the User Guide](user_guide_src/source/installation/upgrade_405.rst)
 
+**Mime Type Detection**
+
+- `Config\Mimes::guessExtensionFromType` now only reverse searches the `$mimes` array if no extension is proposed (i.e., usually not for uploaded files).
+- The fallback values of `UploadedFile->getExtension()` (now: `''`) and `UploadedFile->guessExtension()` (now: `$this->getClientExtension()`) have been changed.
```
michalsn, line 16: Fallbacks should be flipped

app/Config/Mimes.php
```diff
@@ -509,11 +508,20 @@ public static function guessExtensionFromType(string $type, string $proposedExte
...
 		{
-			return $proposedExtension;
+				if(array_key_exists($proposedExtension, static::$mimes) && in_array($type, is_string(static::$mimes[$proposedExtension]) ? [static::$mimes[$proposedExtension]] : static::$mimes[$proposedExtension], true))
+				{
+					// The detected mime type matches with the proposed extension.
+					return $proposedExtension;
+				}
+			
+				// An extension was proposed, but the media type does not match the mime type list.
+				return null;
```
michalsn, line 520: You indent this a little too much.

system/HTTP/Files/UploadedFile.php
```diff
@@ -315,7 +315,7 @@ public function getTempName(): string
 	 */
 	public function getExtension(): string
 	{
-		return $this->guessExtension();
+		return  $this->guessExtension() ?: $this->getClientExtension();
```
michalsn, line 318: Please remove extra space after `return`.

michalsn (commented):
As for the changelog, please add changes to the user guide: https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/changelogs/v4.0.5.rst

`guessExtension()` - well... this change potentially can make some mess, this is a public method, and developers don't expect to receive a `null` value, but an empty string at the worst-case scenario.

app/Config/Mimes.php
sfadschm: Thanks, I don't really get along well with Atom yet :D

CHANGELOG.md
sfadschm: Maybe its time to go to bed already....

sfadschm:
> `guessExtension()` - well... this change potentially can make some mess, this is a public method, and developers don't expect to receive a `null` value, but an empty string at the worst-case scenario.

What I meant was rather replacing the nullable type hint by a simple `string`, as I think the basic method structure should be based on `SplFileInfo` and the nullable typehint was rather unintended.
But some people might already use the possible `null` return as well.

michalsn:
> What I meant was rather replacing the nullable type hint by a simple string

We can't. `File::guessExtension()` can actually return a `null` value.

sfadschm:
Sorry, I wasn't being precise. Proposed change which might break BC:

https://github.com/codeigniter4/CodeIgniter4/blob/6861a58c36f1b4f2cb24dc15da95002288b64bc9/system/Files/File.php#L101-L104
to

	public function guessExtension(): string
	{
		return Mimes::guessExtensionFromType($this->getMimeType()) ?? '';
	}

That would make the function compatible with `UploadedFile->guessExtension` and `getExtension`.

sfadschm:
Btw. while I think this technically is a BC break, the cases will be very few. The method doe not pass a proposed extension to `Mimes` and thus uses the reverse search of mime types. As it gets its mime type from `fileinfo` the only case where `File->guessExtension` will currently return `null` is when the detected mime type is not in the `$mimes` array at all.

michalsn:
> Btw. while I think this technically is a BC break, the cases will be very few. The method does not pass a proposed extension to Mimes and thus uses the reverse search of mime types. As it gets its mime type from fileinfo the only case where File->guessExtension will currently return null is when the detected mime type is not in the $mimes array at all.

Yes, it's true. But I don't want to make any changes to the return types.

I know that there is a small difference between return types between `File::guessExtension()` and `UploadFie::guessExtension()` but the current state represents what was released as an official version of v4. So, if there is no really good reason to change, the return type should stay as it is.

I don't even know when `File::guessExtension()` is called honestly (if it's called anywhere).

sfadschm:
Okay, agreed.

sfadschm:
I think this is ready for review and/or opinions now.
Don't know what is going on with PHPStan right now, but I would think unrelated 😅 

@michalsn I thought about it again and implementing something like automatic mime type updates does not seem important enough to justify the workload and maintenance. After we have the mime handling refactored I will see if there is any important differences between our `$mimes` and the `mimes-db` and if so just update our library for once.

user_guide_src/source/changelogs/v4.0.5.rst
```diff
@@ -12,6 +12,8 @@ Enhancements:
 - New URL helper function ``url_is()`` which allows you to check the current URL to see if matches the given string.
 - Services now have their config parameters strictly typehinted. This will ensure no one will pass a different config instance. If you need to pass a new config with additional properties, you need to extend that particular config.
 - Support for setting SameSite attribute on Session and CSRF cookies has been added. For security and compatibility with latest browser versions, the default setting is ``Lax``.
+- Guessing file extensions from mime type in ``Config\Mimes::guessExtensionFromType`` now only reverse searches the ``$mimes`` array if no extension is proposed (i.e., usually not for uploaded files).
+- The getter functions for file extensions of uploaded files now have different fallback values (``$this->getClientExtension()`` for ``UploadedFile->getExtension()`` and ``''`` for ``UploadedFile->guessExtension()``). This makes the process less dependent on the client.
```
michalsn, line 16:
```suggestion
- The getter functions for file extensions of uploaded files now have different fallback values (``$this->getClientExtension()`` for ``UploadedFile->getExtension()`` and ``''`` for ``UploadedFile->guessExtension()``). This is a security fix and makes the process less dependent on the client.
```

michalsn (changes requested):
This looks good. I have only one small request.

I'm okay with leaving the current mime list as it is. And yeah - don't worry about PHPStan.

sfadschm:
Okay thanks.
Should I squash and regroup the commits for a more readable history? I think paul mentioned something like that some time ago.

michalsn:
Yes, you can but if you don't know how to do it, we will squash it during a merge.

michalsn (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
 - `CodeIgniter\Config\Config` is now deprecated in favor of `CodeIgniter\Config\Factories::config()`
 - HTTP Layer Refactor: Numerous deprecations have been made towards a transition to a PSR-compliant HTTP layer. [See the User Guide](user_guide_src/source/installation/upgrade_405.rst)
 
+**Mime Type Detection**
+
+- `Config\Mimes::guessExtensionFromType` now only reverse searches the `$mimes` array if no extension is proposed (i.e., usually not for uploaded files).
+- The fallback values of `UploadedFile->getExtension()` (now: `$this->getClientExtension()`) and `UploadedFile->guessExtension()` (now: `''`) have been changed.
```
paulbalandan, line 16: I don't know if it is just me but can we clarify the `now:` here? I mean does now means the previous behavior, or now means the new behavior made by this change.

Or you can break this into sentences. "The fallback values of `UploadedFile->getExtension()` and `UploadedFile->guessExtension()` have been changed. `UploadedFile->getExtension` now returns blabla from blabla while `UploadedFile->guessExtension` now returns blabla from blabla."

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
...
+These changes increase security when handling uploaded files, as the client can no longer force wrong mime type on the application. However, they might affect how file extensions are detected in your application.
```
paulbalandan, line 17:
```suggestion
These changes increase security when handling uploaded files as the client can no longer force wrong mime type on the application. However, these might affect how file extensions are detected in your application.
```

app/Config/Mimes.php
```diff
@@ -12,6 +12,9 @@
  * the most common one should be first in the array to aid the guess*
  * methods. The same applies when more than one mime-type exists for a
  * single extension.
+ *
+ * When working with mime types, please make sure you have the ´fileinfo´
+ * extension enabled to reliably detect the media types.
```
paulbalandan, line 17: For composer installed apps, maybe we can add `ext-fileinfo: For better mime detection` in the suggests section of composer.json

CHANGELOG.md
sfadschm: Agreed.

app/Config/Mimes.php
sfadschm: I added the suggest section and a notice. Please check, I'm not really comfortable with composer yet 😅

sfadschm: Talking about `composer` 😆 : 
Whenever I run `composer update` in my local repo, GitHub finds changes for all files in `system/ThirdParty/Kint`. Is that desired behavior?

paulbalandan: I am not aware of any updates to kint. Perhaps the copy of kint in your system/ThirdParty is not updated?

composer.json
```diff
@@ -52,5 +52,8 @@
 		"forum": "http://forum.codeigniter.com/",
 		"source": "https://github.com/codeigniter4/CodeIgniter4",
 		"slack": "https://codeigniterchat.slack.com"
-	}
+	},
+	"suggest": {
+ 		"ext-fileinfo": "Improves mime type detection for files"
+  }
```
paulbalandan, line 58: Please align the closing brace here. Then move the suggests section after require-dev. Add this also to admin/starter/composer.json

sfadschm: What about `admin/module` and `admin/framework?`

app/Config/Mimes.php
sfadschm: There is actually no update, the files do not have content changes, but the modification date was changed.
I think its because of the `post-update` script which copies the files for `kint` and `laminas-escaper` even if there was no update to them.

sfadschm: Btw. automatically installing the pre-commit hook after `composer-update` does not seem to work on Win, too, because `bash` is not a valid command.

paulbalandan: Yes, that date modification is intended as `ComposerUpdate` has no way of telling if the copied files are actually changed. It simply copies from `vendor` to `system/ThirdParty`.

I installed my Git executable on Win10 manually using the available Git for Windows installer. Choose from their the option to include install the optional Linux tools and commands. This way I can run `bash -c 'cp env .env'` directly on powershell or cmd.

composer.json
paulbalandan: Ah yes, forgot those two. Kindly change them too.

app/Config/Mimes.php
sfadschm: Okay.
Thanks, I think I remember I saw that option, but opted out...

paulbalandan (approved):

michalsn:
Thanks @sfadschm!

Thanks @paulbalandan!

PR #4061, opened 2021-01-04
Allow route filter as array (Multiple filters)

cambagulung:
**Description**
This afternoon I needed two filters with arguments for my controller, and that's not possible, combine multiple filters didn't help,

['filter' => ['auth:with_token', 'can:do_this']

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -129,11 +129,13 @@ class Filters
 
 		$files = $locator->search('Config/Filters.php');
 
-		foreach ($files as $file) {
+		foreach ($files as $file)
+		{
 			$className = $locator->getClassname($file);
 
 			// Don't include our main Filter config again...
-			if ($className === 'Config\\Filters') {
+			if ($className === 'Config\\Filters')
+			{
 				continue;
 			}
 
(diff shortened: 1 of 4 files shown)


system/Filters/Filters.php
```diff
@@ -129,13 +129,11 @@ private function discoverFilters()
 
 		$files = $locator->search('Config/Filters.php');
 
-		foreach ($files as $file)
-		{
+		foreach ($files as $file) {
```
samsonasik, line 132: revert this, you can :

```
 vendor/bin/phpcbf --standard=./vendor/codeigniter4/codeigniter4-standard/CodeIgniter4 --encoding=utf-8 system
 ```

**Comment by samsonasik** [comment]
besides to revert cosmetic changes, please provide unit test

File: system/Filters/Filters.php [hunk 2]
```diff
@@ -329,42 +315,50 @@ public function addFilter(string $class, string $alias = null, string $when = 'b
...
 	 * are passed to the filter when executed.
 	 *
-	 * @param string $name
+	 * @param string|array $filter
 	 * @param string $when
 	 *
 	 * @return Filters
 	 */
-	public function enableFilter(string $name, string $when = 'before')
+	public function enableFilter($filter, string $when = 'before')
```
samsonasik, line 323: this is BC break ref https://3v4l.org/7b0GA

cambagulung: > besides to revert cosmetic changes, please provide unit test

I don't know how to make a unit test for this change

sfadschm:
Rather than trying to change the variable type to an array, how about something like introducing another delimeter, lets say `;` to be able to apply multiple filters from a single string?

paulbalandan:
Changing `enableFilter()`'s signature is a sure BC break. I would say the delimiter solution of @sfadschm is feasible here.

paulbalandan:
@cambagulung Are you still interested in finishing up this PR? If not, I'll close in a week if no further activity by then.

PR #4064, opened 2021-01-05
[Static Analysis] Use Rector 0.9 for 4.1 branch with php version feature 7.3 

samsonasik:
This is the start over. Update rector dependency to 0.9 for 4.1 branch with php version feature 7.3

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
 		"phpstan/phpstan": "^0.12",
 		"phpunit/phpunit": "^9.1",
 		"predis/predis": "^1.1",
-		"rector/rector": "^0.8",
+		"rector/rector": "^0.9",
 		"squizlabs/php_codesniffer": "^3.3"
 	},
 	"config": {
--- a/rector.php
+++ b/rector.php
@@ -30,7 +30,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
 	// auto import fully qualified class names
 	$parameters->set(Option::AUTO_IMPORT_NAMES, true);
 	$parameters->set(Option::ENABLE_CACHE, true);
-	$parameters->set(Option::PHP_VERSION_FEATURES, '7.2');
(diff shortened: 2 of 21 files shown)


samsonasik:
This PHPStan notice only happen in php 8: 

```bash
 ------ -------------------------------------------- 
  Line   system/Test/Filters/CITestStreamFilter.php  
 ------ -------------------------------------------- 
  50     No error to ignore is reported on line 50.  
 ------ -------------------------------------------- 
 ```

**Comment by samsonasik** [comment]
I think we can run PHPStan only in PHP 8

**Review by paulbalandan — APPROVED** [review]
PHPStan is now unpredictable with the results on PHP8. If we want to run static analysis also on PHP8, we may need to have a separate baseline file for PHP8-only errors.

**Comment by MGatner** [comment]
I addressed PHPStan in https://github.com/codeigniter4/CodeIgniter4/pull/4065. Let's hold off on a decision and see fi anything comes from Ondre's comments upstream. If you want to follow the links in my PR and add your support it may help with popular opinion.

For now we'll consider this an allowable failure until we have a solution (my PR or otherwise).

File: .github/workflows/test-phpstan.yml [hunk 1]
```diff
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.4', '8.0']
+        php-versions: ['8.0']
```
MGatner, line 28: Please revert this after #4065 and we'll try it again.

Come to think of it, let's add `7.3` as well. As long as we are supporting all three versions they should all pass static analysis.

MGatner (changes requested):

.github/workflows/test-phpstan.yml
samsonasik: resetted.

MGatner (approved):

PR #4065, opened 2021-01-05
PHPStan Ignore File

MGatner:
**Description**
We’ve run into our first PHP 7/8 PHPStan issue: https://github.com/codeigniter4/CodeIgniter4/runs/1650630303

For whatever reason this line in **CITestStreamFilter** is no longer an issue in PHP 8, but persists in PHP 7:
```
	public function filter($in, $out, &$consumed, $closing)
	{
		while ($bucket = stream_bucket_make_writeable($in))
		{
			static::$buffer .= $bucket->data;

			$consumed += $bucket->datalen;
		}

		// @phpstan-ignore-next-line
		return PSFS_PASS_ON;
	}
```
Ondre is considering work on how to improve this split handling (https://github.com/phpstan/phpstan/issues/4223 by our own @paulbalandan, https://github.com/phpstan/phpstan/issues/4060#issuecomment-723559902) but the current solution is to have different config files for each version.

Given that the problem currently is just single, small, test file I propose we ignore it for all versions and postpone dealing with this issue until we see what Ondre does upstream.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/.github/workflows/test-phpstan.yml
+++ b/.github/workflows/test-phpstan.yml
@@ -10,6 +10,7 @@ on:
     paths:
       - 'app/**'
       - 'system/**'
+      - phpstan.neon.dist
   push:
     branches:
       - 'develop'
@@ -17,6 +18,7 @@ on:
     paths:
       - 'app/**'
       - 'system/**'
+      - phpstan.neon.dist
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


samsonasik:
The `// @phpstan-ignore-next-line` can be removed in `CITestStreamFilter.php`

paulbalandan (commented):
I guess this is our best resort for now pending Ondrej's decision. I would like to have the `// @phpstan-ignore-next-line` removed to be a bit cleaner since we are now ignoring this file altogether.

MGatner:
Oh right! Thanks. Removed!

samsonasik (approved):

paulbalandan (approved):

PR #4066, opened 2021-01-05
Validation: Capability to rewrite rules function in custom validation

abydahana:
In this case, i would to rewrite "is_unique" rules function to my own validation by extending it to core rules since it limited only 4 parameter:

`is_unique[some_table.some_field,primary_key,{primary_key}]`

I need to rewrite it because there will be a possibility to use more primary key validation, e.g: The code number must be unique but based on its parent category (the code number can be equal but the category id must be different):

`is_unique[some_table.some_field,primary_key,{primary_key},category_id,{category_id}]`

Tried to overriding the rules just like in CI3 (sorry for mention) but it's ignored since the core rules has that function exists.
```
<?php namespace Aksara\Laboratory;

class Customrules extends \CodeIgniter\Validation\Rules
{
	/**
	 * Trying to override the core rules
	 */
	public function is_unique(string $str = null, string $field, array $data) : bool
	{
		 // some validation
	}
}
```


paulbalandan:
I'm sorry but you've made the wrong PR. You want to merge `develop` into `master`. Please use your fork for PRs.

abydahana:
> I'm sorry but you've made the wrong PR. You want to merge `develop` into `master`. Please use your fork for PRs.

Oh i'm sorry. Didn't know that. Thanks for telling.

PR #4068, opened 2021-01-05
Warn users on system messages being for internal use

paulbalandan:
**Checklist:**
- [x] Securely signed commits


sfadschm (approved):
Message looks good.
I would have liked to see this in the UG, too. However, I read through the language section and there is no suitable place for such message as the system language strings are never mentioned. So I think adding such note might only cause confusion.

michalsn (approved):

PR #4071, opened 2021-01-06
Model/BaseModel - Fix primary key and  add @throws for builder method

najdanovicivan:
**Description**
This fixes the issue with primaryKey property which was unintentionally left inside base model. This also ads @throws for builder() method

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1675,13 +1675,6 @@ abstract class BaseModel
 		if (method_exists($data, 'toRawArray'))
 		{
 			$properties = $data->toRawArray($onlyChanged, $recursive);
-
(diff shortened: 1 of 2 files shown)


michalsn (commented):
Tests should be adjusted, but other than that is looks good.

najdanovicivan:
> Tests should be adjusted, but other than that is looks good.

I was looking at the code from master branch. But the builder() method was made public before I did the BaseModel split so it should be left as public

michalsn (approved):

paulbalandan (approved):

Issue #2969, opened 2020-05-10
Relax the requirement for SQL to be a string


JanZelenka:
Hi!
I'm building a Database driver. The difficulty is that the database engine is queried using an API, not SQL. As a consequence the table (FROM clause) is a part of the API URL, not the query itself. I would find it helpful if the DB methods that work with the $sql parameter could be relaxed not to require strictly the string data type and allow an array as well.
Thanks for considering

**CodeIgniter 4 version**
4.0.3

**Affected module(s)**
Database

MGatner:
Could you provide an example of what you're talking about it?

JanZelenka:
It goes like this:
I have a Builder class that creates the queries for me - which need to be in a JSON format. Also, since it is API based, the table name is a part of the URL but the where clause is in the body.
To properly use the Connection class I need to pass these array based queries as string, in my override of the execute() method convert them back to arrays, extract the different pieces of information and then convert back to string to use in the API call.

It's lot of converting back and forth that could be skipped if methods like simpleQuery(), execute() and so would allow for arrays to passed in.

lonnieezell:
Unfortunately I don't think we can do that unless it is in the next major release. All of the current drivers have type-hinting of `string`. If anyone is using that to cast variables into a string for them it would be completely broken.

MGatner:
Stringables are accepted for string typehints, so you can accomplish this by passing a class with `__toString()` and you could use your arrays as an Iterable.

PR #4074, opened 2021-01-07
Better support for NoSQL drivers

najdanovicivan:
**Description**
 - BaseConnection - Query class determined automatically if Driver specific Query class exist
 - Removed strong type of string for $sql param in methods since NoSQL mostly use arrays 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -593,11 +593,11 @@ abstract class BaseConnection implements ConnectionInterface
 	/**
 	 * Executes the query against the database.
 	 *
-	 * @param string $sql
+	 * @param mixed $sql
 	 *
 	 * @return mixed
 	 */
-	abstract protected function execute(string $sql);
+	abstract protected function execute($sql);
 
 	/**
 	 * Orchestrates a query against the database. Queries must use
@@ -607,16 +607,16 @@ abstract class BaseConnection implements ConnectionInterface
 	 * Should automatically handle different connections for read/write
 	 * queries if needed.
 	 *
-	 * @param string  $sql
-	 * @param mixed   ...$binds
-	 * @param boolean $setEscapeFlags
-	 * @param string  $queryClass
+	 * @param mixed       $sql
+	 * @param mixed       ...$binds
+	 * @param boolean     $setEscapeFlags
+	 * @param string|null $queryClass
 	 *
 	 * @return BaseResult|Query|boolean
 	 *
 	 * @todo BC set $queryClass default as null in 4.1
 	 */
-	public function query(string $sql, $binds = null, bool $setEscapeFlags = true, string $queryClass = '')
+	public function query($sql, $binds = null, bool $setEscapeFlags = true, string $queryClass = null)
 	{
 		$queryClass = $queryClass ?: $this->queryClass;
 
@@ -718,11 +718,11 @@ abstract class BaseConnection implements ConnectionInterface
 	 * is performed, nor are transactions handled. Simply takes a raw
 	 * query string and returns the database-specific result id.
 	 *
-	 * @param string $sql
+	 * @param mixed $sql
 	 *
 	 * @return mixed
 	 */
-	public function simpleQuery(string $sql)
+	public function simpleQuery($sql)
 	{
 		if (empty($this->connID))
 		{
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -97,13 +97,13 @@ class Query implements QueryInterface
 	/**
 	 * Sets the raw query string to use for this statement.
 	 *
-	 * @param string  $sql
+	 * @param mixed   $sql
 	 * @param mixed   $binds
 	 * @param boolean $setEscape
 	 *
 	 * @return $this
 	 */
-	public function setQuery(string $sql, $binds = null, bool $setEscape = true)
+	public function setQuery($sql, $binds = null, bool $setEscape = true)
 	{
 		$this->originalQueryString = $sql;
 
@@ -158,9 +158,9 @@ class Query implements QueryInterface
 	 * Returns the final, processed query string after binding, etal
 	 * has been performed.
 	 *
-	 * @return string
+	 * @return mixed
 	 */
-	public function getQuery(): string
+	public function getQuery()
 	{
 		if (empty($this->finalQueryString))
 		{
--- a/system/Test/Mock/MockConnection.php
+++ b/system/Test/Mock/MockConnection.php
@@ -43,16 +43,16 @@ class MockConnection extends BaseConnection
 	 * Should automatically handle different connections for read/write
 	 * queries if needed.
 	 *
-	 * @param string  $sql
-	 * @param mixed   ...$binds
-	 * @param boolean $setEscapeFlags
-	 * @param string  $queryClass
+	 * @param mixed       $sql
+	 * @param mixed       ...$binds
+	 * @param boolean     $setEscapeFlags
+	 * @param string|null $queryClass
 	 *
 	 * @return BaseResult|Query|boolean
 	 *
 	 * @todo BC set $queryClass default as null in 4.1
 	 */
-	public function query(string $sql, $binds = null, bool $setEscapeFlags = true, string $queryClass = '')
+	public function query($sql, $binds = null, bool $setEscapeFlags = true, string $queryClass = null)
 	{
 		$queryClass = str_replace('Connection', 'Query', static::class);
 
@@ -161,11 +161,11 @@ class MockConnection extends BaseConnection
 	/**
 	 * Executes the query against the database.
 	 *
-	 * @param string $sql
+	 * @param mixed $sql
 	 *
 	 * @return mixed
 	 */
-	protected function execute(string $sql)
+	protected function execute($sql)
 	{
 		return $this->returnValues['execute'];
 	}
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -75,13 +75,13 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
 	 * NOTE: This version is based on SQL code. Child classes should
 	 * override this method.
 	 *
-	 * @param string $sql
+	 * @param mixed  $sql
 	 * @param array  $options    Passed to the connection's prepare statement.
 	 * @param string $queryClass
 	 *
 	 * @return mixed
 	 */
-	public function prepare(string $sql, array $options = [], string $queryClass = 'CodeIgniter\\Database\\Query')
+	public function prepare($sql, array $options = [], string $queryClass = 'CodeIgniter\\Database\\Query')
 	{
 		// We only supports positional placeholders (?)
 		// in order to work with the execute method below, so we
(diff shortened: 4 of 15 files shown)


system/Test/Mock/MockConnection.php
```diff
@@ -43,14 +43,14 @@ public function shouldReturn(string $method, $return)
...
-	 * @param string  $queryClass
+	 * @param mixed       $sql
+	 * @param mixed       ...$binds
+	 * @param boolean     $setEscapeFlags
+	 * @param string|null $queryClass
 	 *
 	 * @return BaseResult|Query|false
 	 */
-	public function query(string $sql, $binds = null, bool $setEscapeFlags = true, string $queryClass = 'CodeIgniter\\Database\\Query')
+	public function query($sql, $binds = null, bool $setEscapeFlags = true, ?string $queryClass = null)
```
samsonasik, line 53: while the case is probably minimal, this is bc break,  ref https://3v4l.org/QES73

system/Database/BaseConnection.php
```diff
@@ -302,6 +309,10 @@ public function __construct(array $params)
 		{
 			$this->$key = $value;
 		}
+
+		$queryClass = str_replace('Connection', 'Query', get_class($this));
```
paulbalandan, line 313:
```suggestion
		$queryClass = str_replace('Connection', 'Query', static::class);
```

system/Database/BaseConnection.php
```diff
@@ -302,6 +309,10 @@ public function __construct(array $params)
...
+
+		class_exists($queryClass) && $this->queryClass = $queryClass;
```
paulbalandan, line 315: While short circuiting is not wrong, these instances may pose problems with static analysis tools on undiscovered symbols. I suggest using the braces here:
```php
if (class_exists($queryClass))
{
    $this->queryClass = $queryClass;
}
```

system/Database/BaseConnection.php
```diff
@@ -572,11 +583,11 @@ public function addTableAlias(string $table)
 	/**
 	 * Executes the query against the database.
 	 *
-	 * @param string $sql
+	 * @param mixed $sql
 	 *
 	 * @return mixed
 	 */
-	abstract protected function execute(string $sql);
+	abstract protected function execute($sql);
```
paulbalandan, line 590: As stated, changing the signature may be breaking for existing implementations.

system/Database/BaseConnection.php
```diff
@@ -588,15 +599,17 @@ public function addTableAlias(string $table)
...
+	 * @param mixed       ...$binds
+	 * @param boolean     $setEscapeFlags
+	 * @param string|null $queryClass
 	 *
 	 * @return BaseResult|Query|false
 	 */
-	public function query(string $sql, $binds = null, bool $setEscapeFlags = true, string $queryClass = 'CodeIgniter\\Database\\Query')
+	public function query($sql, $binds = null, bool $setEscapeFlags = true, ?string $queryClass = null)
 	{
+		is_null($queryClass) && $queryClass = $this->queryClass;
```
paulbalandan, line 611: Aside from the possible BC break in the method signature, this line can use the Elvis operator.
```suggestion
		$queryClass = $queryClass ?: $this->queryClass;
```

system/Database/Query.php
```diff
@@ -473,6 +473,58 @@ protected function matchSimpleBinds(string $sql, array $binds, int $bindCount, i
...
 
+	/**
+	 * Returns string to display in debug toolbar
+	 *
+	 * @return string
+	 */
+	public function debugToolbarDisplay() : string
+	{
+		// Key words we want bolded
+		$highlight = [
```
paulbalandan, line 484: I don't know if there is a performance gain but maybe we can set this as `static` variable?
```suggestion
		static $highlight = [
```

system/Debug/Toolbar/Collectors/Database.php
```diff
@@ -138,59 +138,13 @@ protected function formatTimelineData(): array
...
-			foreach ($highlight as $term)
-			{
-				$sql = str_replace($term, "<strong>{$term}</strong>", $sql);
-			}
-
-			$data['queries'][] = [
+		$data['queries'] = array_map(function ($query) {
+			return [
+				//@phpstan-ignore-next-line
 				'duration' => ($query->getDuration(5) * 1000) . ' ms',
```
paulbalandan, line 144: I'm not sure of the phpstan error you're suppressing but if we do it like this will it be gone?
```suggestion
		$data['queries'] = array_map(function (Query $query) {
			return [
				'duration' => (string) ($query->getDuration(5) * 1000) . ' ms',
```

najdanovicivan: PhpStan error happens at
$query->getDuration(5) * 1000

since getDuration() returns string and it warns about doing multiplication on the string so it will be more appropriate to do
floatval($query->getDuration(5))*1000

najdanovicivan:
As both QueryClass and $sql string type are BC I'm gonna move the debug toolbar stuff in separate PR

najdanovicivan:
DebugToolbar changes are moved to separate pr #4074

paulbalandan:
To preserve BC, how about using a delimiter-based solution to the `$sql` string param? Then use `explode`. It may not be optimal but that's the least our option. For NoSQL implementations with array SQL, we can use `implode` first then this one will explode it.

paulbalandan:
Existing issue: #2969 

So it is either preserve BC or push with this one with proper upgrading instructions.

najdanovicivan:
@paulbalandan I've tried doing it using jsonEncoded array but this all makes it very annoying to write queries as you have to write query as array then encode it each time query is called.

The solution here is much more elegant. The only issue is how to get around the BC.

IMO it's completely fine merging this one in 4.1 and adding a section in docs on how to do the update. I don't have any clue how many custom driver implementation exist but I don't think at this point there will be a lot of them. So changing this early on and giving users a clear instructions about the change so that we can have better code makes it worthy of a BC change than doing a lot of nasty patches in the future.

This will also completely resolve the #2969

najdanovicivan:
Also Query Class handling can be resolved without doing a BC by making the query class empty string instead of the hardcoded default value. 

I can split the Query Class stuff into separate PR and revisit this one once it's merged

michalsn:
Personally, I'm good with these changes as long as they will be merged to the 4.1 branch with proper instructions for migration and changelog.

I think that the alternative of having some helper function to encode/decode the strings every time would be really bad in a long run.

najdanovicivan:
Ok. I'll go ahead and split the Query Class stuff outside of this one so it can be merged

najdanovicivan:
Here's the PR for QueryClass #4089 Handling once this one's merged I'll rebase this one on the development

MGatner:
Now that the Query class is merged, and before you go through the rebase trouble, I had an idea that does not circumvent the breaking nature but might help.

Can we define a database-specific `Statement` class? At its simplest, `BaseStatement` could be a protected property with `__toString()` to be compatible with the current SQL drivers, but this would free up NoSQL drivers (or PDO, or whatever) to define more complex statement directives to be handled by `Query`. Does that get too fancy?

najdanovicivan:
@MGatner This is not a bad idea but it will require updating all the existing drivers. I'll take a look into it to see what can be done.

paulbalandan:
@najdanovicivan the `Stringable` class concept of @MGatner is allowed on `string`  typehints. Maybe we can use that instead to avoid the BC. See https://3v4l.org/LnIRE

MGatner:
Wow for real, I thought that was one PHP 8 but your 3v4l seems to suggest it was previous versions as well. In that case we should be in awesome shape, just need a well-thought `Statement` class/base/interface/something.

najdanovicivan:
I'll try doing it with BaseStatement class. But in such case we should replace all the exiting Implementations with it.  I don't like mixing objects with string. In such case all methods should have type of BaseStatement which will again be BC. But having class for query statements will be very useful especially for NoSQL databases.

najdanovicivan:
In any case the type hint adds problems. Additionally if type of the query is changed to be `Stringable` it will not accept regular strings.  And TypeError will be thrown

```
Argument 1 passed to CodeIgniter\Database\BaseConnection::query() must implement interface Stringable, string given
```

MGatner:
Any class with a `__toString()` method will be accepted for functions that require a `string` parameter - `Stringable` is actually more restrictive than `string` in this regard.

system/Database/BaseConnection.php
```diff
@@ -593,11 +593,11 @@ public function addTableAlias(string $table)
 	/**
 	 * Executes the query against the database.
 	 *
-	 * @param string $sql
+	 * @param mixed $sql
```
kenjis, line 596: `mixed` means anything. It does not help for code reading.

Can you change it to `string|array`?

system/Database/BaseConnection.php
kenjis: Does the var name `$sql` make sense in the context of NoSQL?

system/Database/BaseConnection.php
najdanovicivan: It may be even needed to use object for such so it can be string|array|object but the main problem is type hint in the method definition. Since merging this one will break all the exiting user implementation where any of these methods have override.

system/Database/BaseConnection.php
najdanovicivan: The better name for the context will be `$statement`

lonnieezell:
This has been 3 months with no activity. Is this still a desired feature @MGatner @samsonasik @paulbalandan ?

MGatner:
It's kind of a deadlock. CI4 works very poorly with NoSQL databases, but the changes we would need to make to add support are all breaking (though small). 

For my NoSQL project (Firestore) I ended up ditching the database layer altogether. My hunch is that because NoSQL implementations don't share a common language that they will be much harder to make "driver-agnostic" and may not be worth integrating into a unified database layer. If that is that case then the focus should be on making Model and Entity classes to maximize support. 

But I've also worked with @najdanovicivan on some of these NoSQL implementations and he has done a lot of work and had some success, I respect his opinions on the matter.

lonnieezell:
I've only used a NoSQL db once, though it was on a pretty large application. My experience then led me to believe the workflow is too different in all but the simplest of cases, to be standardized into a driver-agnostic solution. There are portions that could be included in a NoSQL layer, though I think things would get pretty complex and convoluted if you were to attempt to rearchitect that. Personally, I'm all for simplifying and keeping each in their own lane, so to speak.

najdanovicivan:
I believe this one can be closed. I found a way to get around this limitation it's not the nice looking solution but at least completely avoids the need to modify all of those files.

What I dis was made a custom statement class which implements Stringable where __toString() basically returns serialize($this).  So the object can be passed as string to the execute method where it is unserialized to get the query parameters.

Maybe better option will be to have some sort of BaseStatement class but it will still be a BC so there is no way to make this one without introducing a BC.

paulbalandan:
Closing this for now. Until we can arrive at a solution that meets all of our expectations and reservations, this PR is obviously at a standstill. Thanks for the PR, anyway, @najdanovicivan !

PR #4080, opened 2021-01-08
Get JSON Variable

caswell-wc:
**Description**
This pull request gives more options when building a JSON API for accessing data in the JSON of a request. This was done by making getVar() smart enough to know when it is a JSON request and grab the data accordingly. This also allows for using "dot" notation for getting data outside of the root level. For more control this PR also gives a new function called getJsonVar() which allows the developer to control the format of the returned data.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -288,7 +288,7 @@ class IncomingRequest extends Request
 	//--------------------------------------------------------------------
 
 	/**
-	 * Fetch an item from the $_REQUEST object. This is the simplest way
+	 * Fetch an item from the $_REQUEST object or a JSON input stream. This is the simplest way
 	 * to grab data from the request object and can be used in lieu of the
 	 * other get* methods in most cases.
 	 *
@@ -300,6 +300,26 @@ class IncomingRequest extends Request
 	 */
 	public function getVar($index = null, $filter = null, $flags = null)
 	{
+		if ($this->isJSON())
+		{
+			if (is_null($index))
+			{
+				return $this->getJSON();
+			}
+
+			if (is_array($index))
+			{
+				$output = [];
+				foreach ($index as $key)
+				{
+					$output[$key] = $this->getJsonVar($key);
+				}
+				return $output;
+			}
+
+			return $this->getJsonVar($index);
+		}
+
 		return $this->fetchGlobal('request', $index, $filter, $flags);
 	}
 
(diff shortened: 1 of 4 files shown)


samsonasik:
please rebase

caswell-wc:
@samsonasik Rebase done, thanks

system/HTTP/IncomingRequest.php
```diff
@@ -325,6 +345,25 @@ public function getJSON(bool $assoc = false, int $depth = 512, int $options = 0)
 		return json_decode($this->body, $assoc, $depth, $options);
 	}
 
+	/**
+	 * Get a specific variable from a JSON input stream
+	 *
+	 * @param  string  $index The variable that you want which can use dot syntax for getting specific values.
+	 * @param  boolean $assoc If TRUE return the result as an associative array.
+	 * @return array|mixed|null|object
```
paulbalandan, line 353:
```suggestion
	 * @param string  $index The variable that you want which can use dot syntax for getting specific values.
	 * @param boolean $assoc If true, return the result as an associative array.
	 *
	 * @return mixed
```

samsonasik (approved):

system/HTTP/Message.php
```diff
@@ -140,7 +140,7 @@ public function getProtocolVersion(): string
 	 *
 	 * @deprecated Use header calls directly
 	 */
-	public function isJSON()
+	public function isJSON(): bool
```
samsonasik, line 143: revert this
```suggestion
      public function isJSON()
```

paulbalandan (approved):

system/HTTP/Message.php
caswell-wc: @samsonasik Just so I know in the future, can you tell me why this needed to be reverted? There are plenty of places where return types are being used in the framework and this function is meant to only return a boolean.

samsonasik: change return type declaration is BC break https://3v4l.org/YHmT2

iRedds: @samsonasik The method must explicitly return a boolean value.
The method was added to the development branch between 4.0.4 and 4.1.0 (4.0.5) releases. So it cannot be BC. 
Your example is not a BC, but a bad practice implementation. 

What you did was a slap in the face. "I'm the boss. You're a fool "(с)

samsonasik: The method marked as `@deprecated` so changing return type is not make sense.It may considered to be removed entirely instead.

The example is BC break, once we define public/protected method on non-final class, we need to maintain that.

iRedds: This method has been marked as ```@deprecated``` but has never been included in any release. It could just as easily have been removed. There is no reason to mark it as ```@depratacted```.

The framework core does not override this method, so BC cannot be.
I repeat. This method was not available until release 4.1.
But suddenly he began to break something. 

Stop justifying insanity!

samsonasik: @iRedds could you provide a PR to remove this? Thank you.

PR #4081, opened 2021-01-08
Fixing a bug in Message::isJSON

caswell-wc:
Fixes the bug pointed out in a comment on #3719.

**Description**
Fixed an issue in isJSON that was breaking it if the content type header had the encoding in it as well.

**Checklist:**
- [x] Securely signed commits
- [ ]N/A Component(s) with PHPdocs
- [ ]N/A User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/Message.php
+++ b/system/HTTP/Message.php
@@ -142,7 +142,14 @@ class Message implements MessageInterface
 	 */
 	public function isJSON()
 	{
-		return $this->hasHeader('Content-Type')
-			&& $this->header('Content-Type')->getValue() === 'application/json';
+		if (! $this->hasHeader('Content-Type'))
+		{
+			return false;
+		}
+
+		$header = $this->header('Content-Type')->getValue();
+		$parts  = explode(';', $header);
+
+		return in_array('application/json', $parts);
 	}
 }
--- a/system/HTTP/Message.php
+++ b/system/HTTP/Message.php
@@ -150,6 +150,6 @@ class Message implements MessageInterface
 		$header = $this->header('Content-Type')->getValue();
 		$parts  = explode(';', $header);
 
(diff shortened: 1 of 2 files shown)


caswell-wc:
@MGatner  I did notice that you put a deprecated flag on this function. I'm wondering why that is when you were the one that suggested the creation of the function. Given the change needed in this PR, I am wondering if it might be a better idea to keep it?

MGatner:
I'm okay with this PR since it is a bugfix, but we did decide to deprecate that. I know I recommended it but since I have spent a lot of time with the HTTP layer. One of the goals is to incorporate a PSR-7 HTTP layer into the framework, and a corollary to that is relocating functionality to their proper locations.

MGatner (approved):

paulbalandan (approved):

system/HTTP/Message.php
```diff
@@ -142,7 +142,14 @@ public function getProtocolVersion(): string
...
-			&& $this->header('Content-Type')->getValue() === 'application/json';
+		if (! $this->hasHeader('Content-Type'))
+		{
+			return false;
+		}
+
+		$header = $this->header('Content-Type')->getValue();
+		$parts  = explode(';', $header);
+
+		return in_array('application/json', $parts, true);
```
mostafakhudair, line 153:
```suggestion
		return in_array('application/json', explode(';', $this->header('Content-Type')->getValue()), true);
```

use one line

samsonasik: @mostafakhudair that's harder to read/understand the flow.

samsonasik (approved):

samsonasik:
Thank you @caswell-wc

caswell-wc:
You're welcome @samsonasik.

Issue #4084, opened 2021-01-08
Bug: getRandomName Return file extension .csv when upload docx or xlsx becouse mimetype application/octet-stream are set 


sandrocantagallo:
**Describe the bug**
if i try to upload a docx file with this code:

``` PHP
  helper(['form', 'url']);
        if (isset($_FILES['file-1'])) {
            $strInputFileName = "file-1";
        } else {
            $strInputFileName = "file-2";
        }
        $input = $this->validate([
            'file' => [
                'uploaded['.$strInputFileName.']',
                'mime_in['.$strInputFileName.',image/jpg,image/jpeg,image/png,application/pdf,application/msword,application/msword,application/excel,application/x-excel,application/vnd.openxmlformats-officedocument.wordprocessingm,application/octet-stream,application/vnd.openxmlformats-officedocument.wordprocessingml.document,application/vnd.openxmlformats-officedocument.spreadsheetml.sh,application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet]',
                'max_size['.$strInputFileName.',25048000]',
            ]
        ]);

        if (!$input) {
            $data = array('error' => "Errore: Non è possibile salvare il file nel sistema. Controllare dimensioni massime e estensione dello stesso: ".print_r($_FILES['file-1'], true));
            echo json_encode($data);
        } else {

            if($files = $this->request->getFiles())
            {
            foreach($files[$strInputFileName] as $file)
            {
                if ($file->isValid() && ! $file->hasMoved())
                {

                    $mime_file = $file->getMimeType();

                    $newName = $file->getRandomName();

                    $file->move(WRITEPATH.'uploads', $newName);
                    $name = $file->getName();
                    echo json_encode(array("data" =>  $name));

                } else {
                    $data = array('error' => "Errore: Tipo file non valido, il file è stato rimosso.");
                    echo json_encode($data); 
                }
            }
            } else {
                $data = array('error' => "Errore: Nessun file inviato. Riprovare nuovamente.");
                echo json_encode($data);
            }
            
        }
```
the extection of the new filename generated with $file->getRandomName() is **.csv**

if i try to get mimetype with  $file->getMimeType(); it return right: application/vnd.openxmlformats-officedocument.wordprocessingml.document but in getRandomName() the mimetype detected is: application/octet-stream

i think that the problem is here: vendor\codeigniter4\framework\system\HTTP\Files\UploadedFile.php

```PHP
/**
[...]

PR #4085, opened 2021-01-08
Use getMimeType instead of getClientMimeType.

### LINKED ISSUE #3979: Bug: RAR file detected as CSV

sfadschm:
Possibly fixes #3979 and #4084.

As stated in the PHPDoc, `getClientMimeType` is not trustable. As `File->getMimeType()` defaults to the same value that `getClientMimeType` instantly returns if `fileinfo` is not available, we can use this method to more reliably guess extensions.

I feel that this is also related to most of the `octet-stream` related issues that keep popping up recently, as the mime type provided by the client (browser) often defaults to that (e.g. #3921).

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide


sfadschm:
Tests are passing locally. Not really sure why `fileA.txt` gets guessed as `csv`.

michalsn:
Are the tests are really passing locally for you? From what I see calling `$file->getExtension()` is the same as calling `$file->guessExtension()`.

sfadschm:
:-D I will go through the test steps and see where this happens. I feel the guessing enters the reverse search loop for some reason and thus returns `csv` as it always does.
Is `fileinfo` enabled in the Github tests?

michalsn:
I just sat down at my computer to check it and tests are failing for me also. The mime type returned by the OS for this file is `application/octet-stream`. But... when I added some content to this file, it started to be recognized correctly.

sfadschm:
Interesting, but would match the failing test upstream. What OS are you on? I'm on Win10.

michalsn:
macOS and a virtual machine with Ubuntu, so the environment is similar to GitHub actions.
I got the same results on both.

sfadschm:
Okay. I will try to add some content to it. 
Originally, `fileA` and `fileB` only contain a single line break. I found that in Windows if I remove the line break, they are detected as `application/empty`.
So maybe this is some issues with special character only files or such?

michalsn:
> Originally, fileA and fileB only contain a single line break. I found that in Windows if I remove the line break, they are detected as application/empty.

Yes, here it acts similar, I get `inode/x-empty`. But I'm not gonna pretend I have an idea why it is happening 😅 I guess empty files just fallback to the default mime type automatically?

sfadschm:
Seems like that. O would think the same for the original files. Maybe on Mac/Ubuntu a file with only line break is not detected as `text` because it does not consider special characters as text characters. If we aplly the other MIME type PR (or similar), these would return `null` as guessed extension, which I feel to be the most right, as no content type can be deduced from empty/special char only files.
How do you feel about this?

michalsn:
Yes, I think it's correct and the `null` value would be desired in that case. Following what OS returns is the best thing we can do here.

michalsn:
Oh.. and as for this PR I think we should modify those files (add content to them), so they can be recognized correctly.

sfadschm:
Alright.
Can I add a `rar` file to the files also or is this considered unsafe?
Would be nice to also text with `non-text` mime types.

michalsn:
I don't see a problem. Just keep the size really small, and the number of files you want to test sane 😄

sfadschm:
Seems to work. Thanks @michalsn!

michalsn (approved):
It looks good. Seems like a big improvement to mime checking and also for testing it.

I have one thing I would change but maybe it's better to address this in a different PR (maybe this one which is in draft right now).

sfadschm:
Alright, feel free to add to the draft or redraft :-)

paulbalandan (approved):

michalsn:
Thank you @sfadschm!

PR #4092, opened 2021-01-10
Alternative cookie implementation

paulbalandan:
**Description**
This is an alternative Cookie object implementation for the HTTP component and others (Security, Session).

- A `CookieInterface` defines the basic methods for all HTTP cookies.
- A `ClonableCookieInterface`, which extends `CookieInterface` defines methods of an immutable HTTP cookie.
- A `Cookie` implementing `ClonableCookieInterface` representing an immutable HTTP cookie object.
- A `CookieStore` represents an immutable collection of `Cookie` objects.

This tried to retain BC as much as possible by retaining previous logic. However, inevitable changes include `ResponseTrait::getCookies()` will return an array of `Cookies` instead of array of arrays; `ResponseTrait::getCookie()` will return `Cookie[]|Cookie|null` instead of array of arrays, array, or null; 

**Checklist:**
- [x] Securely signed commits
- [x] Integration with Security and Session components
- [x] Conforms to style guide

**To-do:**
- [ ] User guide updated
- [ ] Upgrading instructions


system/HTTP/Cookie/Cookie.php
```diff
@@ -0,0 +1,811 @@
...
+		{
+			$newDefaults = [
+				'prefix'   => $config->cookiePrefix,
+				'expires'  => $config->cookieExpires ?? 0,
+				'domain'   => $config->cookieDomain,
+				'path'     => $config->cookiePath,
+				'secure'   => $config->cookieSecure,
+				'httponly' => $config->cookieHTTPOnly,
+				'raw'      => $config->cookieRaw ?? false,
+				'samesite' => $config->cookieSameSite,
```
paulbalandan, line 139:
```suggestion
				'samesite' => $config->cookieSameSite ?? self::SAMESITE_LAX,
```

michalsn:
It looks solid. I hope to look at this code during the week.

But I have a general question. Is this rewrite adds any missing features or resolve underlying problems that the previous implementation had? I'm asking about it because of BC. Cookies are most likely used quite often, so any change would be painful to developers.

app/Config/App.php
```diff
@@ -279,14 +280,14 @@ class App extends BaseConfig
...
-	 * Cookie HTTP Only
+	 * Cookie HttpOnly
 	 * --------------------------------------------------------------------------
 	 *
 	 * Cookie will only be accessible via HTTP(S) (no JavaScript).
 	 *
 	 * @var boolean
 	 */
-	public $cookieHTTPOnly = false;
+	public $cookieHTTPOnly = true;
```
iRedds, line 290: I think that the default property value should be the same as the setcookie() function, equal to "false"

mostafakhudair:
This is inconsistent with Build Cookie Class PR #3983 !!

app/Config/App.php
paulbalandan: I set it to true for better security. I think the framework should lean on secure practices for apps.

paulbalandan:
> It looks solid. I hope to look at this code during the week.
> 
> But I have a general question. Is this rewrite adds any missing features or resolve underlying problems that the previous implementation had? I'm asking about it because of BC. Cookies are most likely used quite often, so any change would be painful to developers.

This adds new features on top of current implementation to retain BC so I'm not certain if the current has some bugs. Honestly, when I integrated the initial state of the Cookie to the HTTP layer, I found IMO "absurd" defaults so I need to modify a lot to retain BC. I have annotated places where BC was retained. There were some defaulting, like in path, that I feel buggy that I opt not to include.

Most of the references here are MDN and RFCs on cookies.

paulbalandan:
> This is inconsistent with Build Cookie Class PR #3983 !!

Yes, because this is an _alternative_.

system/HTTP/Cookie/Cookie.php
```diff
@@ -0,0 +1,811 @@
...
+ * $cookie->getName(); // test_cookie
+ *
+ * $cookie->withName('prod_cookie');
+ * $cookie->getName(); // test_cookie
+ *
+ * $cookie2 = $cookie->withName('prod_cookie');
+ * $cookie2->getName(); // prod_cookie
+ * ```
+ */
+class Cookie implements ClonableCookieInterface
```
paulbalandan, line 36: Implement the `ArrayAccess` interface for BC retention.

mostafakhudair:
> s is an _alternative_.

I already started PR 30days ago and still under progress why we don't finish it ? and then make improvements to it?

michalsn:
I think I like how it works, but we would need clear guidance on how to migrate our existing code to the new Cookie class. Also, this definitely can't go to the 4.0.x branch - the change is just too big.

4.1 feels more proper for such a change but I would like to hear what others think about it.

system/HTTP/ResponseTrait.php
```diff
@@ -762,39 +724,40 @@ public function getCookie(string $name = null, string $prefix = '')
...
+
+				$this->cookieStore = $store->put($cookie);
 				break;
 			}
 		}
 
-		if (! $cookieHasFlag)
+		if (! $found)
 		{
 			$this->setCookie($name, '', '', $domain, $path, $prefix);
```
paulbalandan, line 762:
```suggestion
			$this->setCookie($name, '', 0, $domain, $path, $prefix);
```

PR #4093, opened 2021-01-11
Fix RendererInterface default params

### LINKED ISSUE #4086: Bug: classes overwrites parameter defaults, phpDoc's defined in interfaces
Please fix difference(phpDoc, parameter defaults) between classes and their interfaces.

For example:
in `system/View/RendererInterface.php` `$saveData = false`:
~~~php
interface RendererInterface {
public function render(string $view, array $options = null, bool $saveData = false): string;
public function renderString(string $view, array $options = null, bool $saveData = false): string;
}
~~~

in `system/View/View.php`  `$saveData = null`:
~~~php
class View implements RendererInterface {
public function render(string $view, array $options = null, bool $saveData = null): string
public function renderString(string $view, array $options = null, bool $saveData = null): string
}
~~~

kenjis:
**Description**
Fixes #4086

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/RendererInterface.php
+++ b/system/View/RendererInterface.php
@@ -22,17 +22,17 @@ interface RendererInterface
 	 * Builds the output based upon a file name and any
 	 * data that has already been set.
 	 *
-	 * @param string  $view
-	 * @param array   $options  Reserved for 3rd-party uses since
-	 *                          it might be needed to pass additional info
-	 *                          to other template engines.
-	 * @param boolean $saveData If true, will save data for use with any other calls,
+	 * @param string       $view
+	 * @param array        $options  Reserved for 3rd-party uses since
+	 *                               it might be needed to pass additional info
+	 *                               to other template engines.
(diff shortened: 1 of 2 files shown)


MGatner:
Let's discuss this at the linked issue first. I do think the discrepancy is a mistake so might merit this change, but altering the interface definitely counts as a breaking change.

WinterSilence:
@kenjis @MGatner it's break backward compatibility

kenjis:
@WaldemarStanislawski What's the difference between `@param bool|null` and `?bool`?

WinterSilence:
@kenjis
You can pass NULL:
~~~php
/**
 * @param bool|null $value
 */
function f1(?bool $value = null)
{
}
~~~
You can't pass NULL:
~~~php
/**
 * @param bool $value
 */
function f2(bool $value = null)
{
}
~~~

PR #4094, opened 2021-01-11
Fixed a bug where a newline was treated as a valid value even if it was included at the end.

ytetsuro:
**Description**
I want it to be false when

```php
<?php
$validation->check("0\n", 'decimal');
```

However, the regular expression validation implemented in CodeIgniter4 allowed a trailing newline code.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -44,6 +44,7 @@ class FormatRules
 			return true;
 		}
 
+		// @see https://regex101.com/r/LhqHPO/1
 		return (bool) preg_match('/\A[A-Z ]+\z/i', $value);
 	}
 
@@ -56,7 +57,8 @@ class FormatRules
 	 */
 	public function alpha_dash(?string $str = null): bool
 	{
-			return (bool) preg_match('/\A[a-z0-9_-]+\z/i', $str);
+		// @see https://regex101.com/r/XfVY3d/1
+		return (bool) preg_match('/\A[a-z0-9_-]+\z/i', $str);
 	}
 
 	/**
@@ -72,6 +74,7 @@ class FormatRules
 	 */
 	public function alpha_numeric_punct($str)
 	{
+		// @see https://regex101.com/r/6N8dDY/1
 		return (bool) preg_match('/\A[A-Z0-9 ~!#$%\&\*\-_+=|:.]+\z/i', $str);
 	}
 
@@ -96,6 +99,7 @@ class FormatRules
 	 */
 	public function alpha_numeric_space(?string $str = null): bool
 	{
+		// @see https://regex101.com/r/0AZDME/1
 		return (bool) preg_match('/\A[A-Z0-9 ]+\z/i', $str);
 	}
 
@@ -123,6 +127,7 @@ class FormatRules
 	 */
 	public function decimal(?string $str = null): bool
 	{
+		// @see https://regex101.com/r/HULifl/1/
 		return (bool) preg_match('/\A[-+]?[0-9]{0,}\.?[0-9]+\z/', $str);
 	}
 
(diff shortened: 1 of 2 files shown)


MGatner:
It took me a minute but I think I understand this now. Basically, because we used `^` and `$` to check "beginning/end of *line*" the newlines were not included in the validation. Switching to `\A` and `\z` (https://www.php.net/manual/en/regexp.reference.anchors.php) matches based on the entire *string* which will handle newline as a character to be validated as well.

Would implementing the `PCRE_MULTILINE` modifier (https://www.php.net/manual/en/reference.pcre.pattern.modifiers.php) be a better solution? I'm not great with regex.

MGatner:
@ytetsuro Please squash this to a single commit. Alternatively, whoever merges it please be sure to squash.

ytetsuro:
> Please squash this to a single commit.

fixed.

ytetsuro:
> better solution?

It is I think a bad solution.
If you enable PCRE_MULTILINE modifier, it will be enabled when following.

```php
<?php
$foo = <<<EOT
0123
234
567
EOT;
preg_match('/^[0-9]+$/m', $foo); // true
```

MGatner:
Gotcha, thanks for clarifying! I'm good with this solution then, we'll see if anyone else weighs in.

MGatner (approved):

paulbalandan (approved):

system/Validation/FormatRules.php
```diff
@@ -44,7 +44,7 @@ public function alpha_space(?string $value = null): bool
 			return true;
 		}
 
-		return (bool) preg_match('/^[A-Z ]+$/i', $value);
+		return (bool) preg_match('/\A[A-Z ]+\z/i', $value);
```
samsonasik, line 47: please add `@see https://regex101.com/` link comment so it will be easier to verify in the future if bug/improvement needed

michalsn (approved):

system/Validation/FormatRules.php
ytetsuro: @samsonasik 

> please add @see https://regex101.com/ link comment so it will be easier to verify in the future if bug/improvement needed

Do we write this link in the code where all the regular expressions are written?
I don't see the need to add a link.
I think the attached URL is maintained by an individual, but how long will it be valid?

samsonasik: You can check these:

https://github.com/codeigniter4/CodeIgniter4/blob/3119fc8d0b8790cc408c292bcab218931f6d32d2/system/View/Parser.php#L576

https://github.com/codeigniter4/CodeIgniter4/blob/3119fc8d0b8790cc408c292bcab218931f6d32d2/system/View/Parser.php#L742

https://github.com/codeigniter4/CodeIgniter4/blob/3119fc8d0b8790cc408c292bcab218931f6d32d2/utils/Rector/UnderscoreToCamelCaseVariableNameRector.php#L28

Yes, not everwhere yet, but I think it will be better to add it whenever possible for better future.

I've no information about how long it will be valid, but it seems it long enough ,for example, this is from 5 years ago https://regex101.com/library/wL5mP7 .

ytetsuro: Will this link really be easy to maintain?
Every time a regular expression is edited, does the URL also need to be maintained, and will all the maintainers review and manage it?

The system also has the ability to delete the URLs that are created.
After the merge, if the link is deleted, the information will be lost.
I think test code is more constructive than a link that may or may not work.

What do other MEMBERS think about this?
I don't want to add this URL as my commit, but I will reluctantly add it if many members agree with this.

sfadschm: I vote for adding the link.
As it is positioned directly above the code I would think it is implicit to maintain it (or at least see what it does) when changing the regex itself.
Also, while of course the link might point to a dead end some future day, I consider it to be better to provide information that might get lost instead of not providing info at all in the first place.

michalsn: It would be nice to have a link. It will make our life easier in the future.

ytetsuro: Thanks for giving me your input.
I have added your comment.

PR #4097, opened 2021-01-12
Catch DateTime failure

MGatner:
**Description**
If `DateTime::createFromFormat` fails then it causes unexpected behavior in our own `createFromFormat`. This checks for failure (return `false`) and throws an exception.

See: https://www.php.net/manual/en/datetime.createfromformat.php

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
 		"codeigniter4/codeigniter4-standard": "^1.0",
 		"fakerphp/faker": "^1.9",
 		"mikey179/vfsstream": "^1.6",
-		"phpstan/phpstan": "^0.12",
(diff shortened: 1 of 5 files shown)


MGatner:
I'm locking PHPStan since minor updates keep causing CI issues. FYI @samsonasik we will want to do periodic updates but this has been needed for a while.

samsonasik:
@MGatner yes, locking version is ok 👍

michalsn (approved):

samsonasik (approved):

samsonasik:
Thank you @MGatner

PR #4102, opened 2021-01-14
Clarify Renderer discrepancy

MGatner:
**Description**
Fixes a discrepancy between the interface docblock and the actual method definitions. Clarifies the usage in `View`.

* Fixes https://github.com/codeigniter4/CodeIgniter4/issues/4086
* Replaces https://github.com/codeigniter4/CodeIgniter4/pull/4093

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Config/View.php
+++ b/system/Config/View.php
@@ -17,6 +17,14 @@ namespace CodeIgniter\Config;
 class View extends BaseConfig
 {
 	/**
+	 * When false, the view method will clear the data between each
+	 * call.
+	 *
+	 * @var boolean
+	 */
+	public $saveData = true;
+
+	/**
 	 * Parser Filters map a filter name with any PHP callable. When the
 	 * Parser prepares a variable for display, it will chain it
 	 * through the filters in the order defined, inserting any parameters.
--- a/system/View/View.php
+++ b/system/View/View.php
@@ -155,7 +155,7 @@ class View implements RendererInterface
 	 *  - cache      Number of seconds to cache for
 	 *  - cache_name Name to use for cache
 	 *
-	 * @param string       $view     The view contents
+	 * @param string       $view     File name of the view source
 	 * @param array        $options  Reserved for 3rd-party uses since
 	 *                               it might be needed to pass additional info
 	 *                               to other template engines.
--- a/system/View/View.php
+++ b/system/View/View.php
@@ -152,12 +152,16 @@ class View implements RendererInterface
 	 * data that has already been set.
 	 *
 	 * Valid $options:
-	 *     - cache 		number of seconds to cache for
-	 *  - cache_name	Name to use for cache
+	 *  - cache      Number of seconds to cache for
+	 *  - cache_name Name to use for cache
 	 *
-	 * @param string       $view
-	 * @param array|null   $options
-	 * @param boolean|null $saveData
+	 * @param string       $view     The view contents
+	 * @param array        $options  Reserved for 3rd-party uses since
+	 *                               it might be needed to pass additional info
+	 *                               to other template engines.
+	 * @param boolean|null $saveData If true, saves data for subsequent calls,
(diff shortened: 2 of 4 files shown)


samsonasik (approved):

MGatner:
@WinterSilence @kenjis anything to add?

WinterSilence:
@MGatner 
Define property `saveData` in `ViewConfig`:
~~~php
class ViewConfig extends BaseConfig
{
     /**
      * @var boolean ...
      */
     public $saveData = false;
     ....
}
~~~
https://github.com/codeigniter4/CodeIgniter4/blob/0489678a54d0a4dab534cb1aaaece3faac459ec3/system/View/View.php#L145 replace to `$this->saveData = (bool) $config->saveData;`
Also, need fix `@param array|null $options ` and `$this->renderVars['options'] = $options ?? [];`

MGatner:
@WinterSilence Thank you, all good fixes. Please re-review.

system/View/View.php
```diff
@@ -142,7 +142,7 @@ public function __construct(ViewConfig $config, string $viewPath = null, FileLoc
 		$this->loader   = $loader ?? Services::locator();
 		$this->logger   = $logger ?? Services::logger();
 		$this->debug    = $debug ?? CI_DEBUG;
-		$this->saveData = $config->saveData ?? null;
+		$this->saveData = (bool) $config->saveData;
```
paulbalandan, line 145: Correct me if I'm wrong. But in `View` config there is really no `$saveData` property so it is now added and yet in here it is used all the while? I think we still need the fallback to null here.

`$this->saveData = (bool) ($config->saveData ?? null);`

MGatner: We added it in **system/Config/View.php** so it will for sure always exist. Casting it to `bool` is there as a safeguard in case someone set it to `null` or `''` or `'monkeySoup'` or something in their **app/Config/View.php**.

system/Config/View.php
```diff
@@ -16,6 +16,14 @@
  */
 class View extends BaseConfig
 {
+	/**
+	 * When false, the view method will clear the data between each
+	 * call.
+	 *
+	 * @var boolean
+	 */
+	public $saveData = true;
```
WinterSilence, line 25: it's can be `false` by default - `RendererInterface::render(string $view, array $options = null, bool $saveData = false)`

MGatner: It is `true` in **app/Config/View.php** which extends this, so I think it makes more sense to mirror that.

WinterSilence: @MGatner why not same to `RendererInterface::render()`?

MGatner: I have no clue, I didn't write any of this. Probably a mistake or miscommunication among earlier devs. At this point though we have a public interface which we can't change and an App config file which we could change but would need to note in docs - I think keeping the config files consistent even though they don't match the interface makes the most sense given the mess it already is.

system/View/View.php
paulbalandan: Hmm so this was a mishap from before. I think the `??` should still remain because if a user doesn't update his view config, then `$saveData` will still not be there and PHP will complain of an undefined property. Since we're casting to bool anyway, why not change the fallback to false, then cast to bool for safety.
```suggestion
		$this->saveData = (bool) ($config->saveData ?? false);
```

MGatner: I think we're talking past each other. `$saveData` has always been in **app/Config/View.php**, typed as `boolean`. That file extends **system/Config/View.php** where `$saveData` was missing (this PR adds it). Either way, it will always be present and always be a boolean so the null coalescence is unnecessary.

paulbalandan: Ah. See it now. Thanks for the clarification.

paulbalandan (approved):

PR #4103, opened 2021-01-14
Add Cache File mode

MGatner:
**Description**
* Adds support for configurable file modes when using the `FileHandler` for caching.
* Deprecates `Config\Cache::$storePath` in favor of a new driver-specific array of settings (similar to other drivers)
* Fixes a bug where `'0'` was not allowed as a cache prefix

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Cache/Handlers/FileHandler.php
+++ b/system/Cache/Handlers/FileHandler.php
@@ -38,7 +38,7 @@ class FileHandler implements CacheInterface
 	 * Mode for the stored files.
 	 * Must be chmod-safe (octal).
 	 *
-	 * @var int
+	 * @var integer
 	 *
 	 * @see https://www.php.net/manual/en/function.chmod.php
 	 */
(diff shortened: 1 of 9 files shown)


MGatner:
Failure is unrelated.

michalsn (approved):

user_guide_src/source/libraries/caching.rst
```diff
@@ -59,9 +59,9 @@ more complex, multi-server setups.
 If you have more than one application using the same cache storage, you can add a custom prefix
 here that is prepended to all key names.
 
-**$path**
+**$file**
 
-This is used by the ``file`` handler to show where it should save the cache files to.
+This is an array of settings specific to the  ``File`` handler determine how it should save the cache files.
```
sfadschm, line 64: Some word is missing here.
Maybe ..._handler **to** determine_...?

MGatner: Yep! Good catch, thank you.

system/Cache/Handlers/FileHandler.php
```diff
@@ -44,14 +54,23 @@ class FileHandler implements CacheInterface
 	 */
 	public function __construct(Cache $config)
 	{
-		$path = ! empty($config->storePath) ? $config->storePath : WRITEPATH . 'cache';
-		if (! is_really_writable($path))
+		if (isset($config->file))
```
paulbalandan, line 57: If this returns `false`, then `$this->mode` will not be initialized. Either, we add that to the `else` or add a default value to the property `$mode`.

samsonasik: there is already `??` for that as fallback

paulbalandan: If only `isset($config->file)` resolves to true then there's is fallback. If the Cache config is not updated to use the new property, `$mode` is uninitialized.

https://3v4l.org/qFvvV

samsonasik: I think we can do with property_exisrts then https://3v4l.org/uFhnG

paulbalandan: Hmm. I like that solution as it is cleaner.

MGatner: Good catch and elegant solution! I will implement this later on desktop.

MGatner: I had to modify this a bit to keep the backwards-compatible support for `$config->storePath` but I think the spirit of it is there.

MGatner:
While moving things around I noticed that we use `$this->prefix = $config->prefix ?: ''` in all the handlers. Since `Config\Cache::$prefix` is already defined as a `string` all this essentially does is disables using falsey `'0'` as a prefix, so I've changed it out (still cast as a string in case someone has it incorrectly typed).

paulbalandan (approved):

PR #4106, opened 2021-01-15
docs: change sample code of redirect() to be more common

kenjis:
**Description**
Change sample code of `redirect()` to be more common.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/incoming/filters.rst
+++ b/user_guide_src/source/incoming/filters.rst
@@ -69,7 +69,7 @@ This is typically used to perform redirects, like in this example::
 
         if (! $auth->isLoggedIn())
         {
-            return redirect()->to('/login');
+            return redirect()->to(site_url('login'));
         }
     }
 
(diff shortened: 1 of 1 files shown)


user_guide_src/source/incoming/filters.rst
```diff
@@ -69,7 +69,7 @@ This is typically used to perform redirects, like in this example::
 
         if (! $auth->isLoggedIn())
         {
-            return redirect('login');
+            return redirect()->to('/login');
```
MGatner, line 72: Please encourage using `site_url()` since we have so many issues with URLs in subfolders:
```suggestion
            return redirect()->to(site_url('login'));
```

MGatner (changes requested):

user_guide_src/source/incoming/filters.rst
kenjis: @MGatner Thank you for your advice. Done.

MGatner (approved):

PR #4107, opened 2021-01-15
Cache remember

agungsugiarto:
**Description**
Retrieve & Store cache
Sometimes we wish to retrieve an item from the cache, but also store a default value if the requested item doesn't exist.
Usage:
```php
$value = Service::cache->remember('sample', 3600, function () {
    return [
       'sample' => 'data',
    ];
});
``` 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Cache/CacheInterface.php
+++ b/system/Cache/CacheInterface.php
@@ -11,6 +11,8 @@
 
 namespace CodeIgniter\Cache;
 
+use Closure;
+
 /**
  * Cache interface
  */
@@ -31,6 +33,19 @@ interface CacheInterface
 	 * @return mixed
 	 */
 	public function get(string $key);
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Get an item from the cache, or execute the given Closure and store the result.
+	 *
+	 * @param string  $key      Cache item name
+	 * @param integer $ttl      Time to live
+	 * @param Closure $callback Callback return value
+	 *
+	 * @return mixed
+	 */
+	public function remember(string $key, int $ttl, Closure $callback);
 
 	//--------------------------------------------------------------------
 
--- a/system/Cache/CacheInterface.php
+++ b/system/Cache/CacheInterface.php
@@ -37,19 +37,6 @@ interface CacheInterface
 	//--------------------------------------------------------------------
 
 	/**
-	 * Get an item from the cache, or execute the given Closure and store the result.
-	 *
-	 * @param string  $key      Cache item name
-	 * @param integer $ttl      Time to live
-	 * @param Closure $callback Callback return value
-	 *
-	 * @return mixed
-	 */
-	public function remember(string $key, int $ttl, Closure $callback);
-
-	//--------------------------------------------------------------------
-
-	/**
 	 * Saves an item to the cache store.
 	 *
 	 * @param string  $key   Cache item name
--- a/system/Cache/CacheInterface.php
+++ b/system/Cache/CacheInterface.php
@@ -11,8 +11,6 @@
 
 namespace CodeIgniter\Cache;
 
-use Closure;
-
 /**
  * Cache interface
  */
--- a/system/Cache/Handlers/FileHandler.php
+++ b/system/Cache/Handlers/FileHandler.php
@@ -11,15 +11,13 @@
 
 namespace CodeIgniter\Cache\Handlers;
 
-use CodeIgniter\Cache\CacheInterface;
 use CodeIgniter\Cache\Exceptions\CacheException;
 use Config\Cache;
-use Closure;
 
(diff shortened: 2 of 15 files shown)


system/Cache/CacheInterface.php
```diff
@@ -11,6 +11,8 @@
 
 namespace CodeIgniter\Cache;
 
+use Closure;
+
```
paulbalandan, line 15: I think this is unused.

system/Cache/Handlers/FileHandler.php
```diff
@@ -84,6 +85,31 @@ public function get(string $key)
...
+
+		if (! is_null($value))
+		{
+			return $value;
+		}
+
+		$this->save($key, $value = $callback(), $ttl);
+
+		return $value;
+	}
```
paulbalandan, line 109: Except for DummyHandler, it seems all cache handlers have the same implementation logic for this method. I do not see any differences. Perhaps we can extract `$this->remember` into an abstract parent `BaseHandler` which these classes would extend. Then DummyHandler would just override this method to return null.

```php
abstract class BaseHandler implements CacheInterface
{
	/**
	 * Get an item from the cache, or execute the given Closure and store the result.
	 *
	 * @param string  $key      Cache item name
	 * @param integer $ttl      Time to live
	 * @param Closure $callback Callback return value
	 *
	 * @return mixed
	 */
	public function remember(string $key, int $ttl, Closure $callback)
	{
		$value = $this->get($key);

		if (! is_null($value))
		{
			return $value;
		}

		$this->save($key, $value = $callback(), $ttl);

		return $value;
	}
}

class FileHandler extends BaseHandler
{
//...
}

class DummyHandler extends BaseHandler
{
    public function remember(string $key, int $ttl, Closure $callback)
    {
        return null;
    }
}
```

paulbalandan (commented):
This needs to be discussed in the user guide.

system/Cache/Handlers/FileHandler.php
MGatner: Yes definitely consolidate the code to an abstract base class.

system/Cache/CacheInterface.php
MGatner: It is not used, but I'm torn on whether it should be. Adding a method to the handlers without indicating it in the interface makes it unusable by anything type-hinted `CacheInterface`. However, adding it to the interface is a breaking change. This is why I tend to favor abstract base classes over interfaces.

Maybe since we are adding `BaseHandler` to consolidate `remember()` we should move these definitions there as abstract methods and deprecate the whole interface?

MGatner (changes requested):
Please participate in and resolve conversations.

system/Cache/CacheInterface.php
paulbalandan: I am against deprecating the interface. An abstract class, as an incomplete implementation of a contract, can provide the default implementations of the contract or add new features on top of it. So an abstract class is a supplement to an interface, not its replacement.

Another way I'm thinking is extending the CacheInterface by a new interface with this method `remember`. Then we can either change the phpdocs to use this new interface or use `instanceof` checks.
`($cache instanceof NewInterface) && $cache->remember();`

system/Cache/Handlers/FileHandler.php
agungsugiarto: Agree with @paulbalandan, i will refactor with `BaseHandler`.

paulbalandan (commented):
Please add a discussion in the user guide.

paulbalandan:
bump @agungsugiarto for the UG update

agungsugiarto:
Forgive me @paulbalandan, im not good to write some documentation.

paulbalandan:
ping @samsonasik , there seems to be a problem with rector. It says it cannot find the `ReflectionUnionType` class. I think it's internally used by rector.

samsonasik:
@paulbalandan Yes, I reverted phpstan update https://github.com/codeigniter4/CodeIgniter4/pull/4173 . That issue is came from phpstan.

paulbalandan (approved):

system/Cache/CacheInterface.php
datamweb: Are you open to accepting a breaking change in v4.7 by officially adding the `remember()` method to `CacheInterface`? 
Or would you prefer a non-breaking PHPDoc update like the one below?
```php
// In Common.php and Services.php
/**
 * @return BaseHandler|CacheInterface
 */
```

paulbalandan: Adding it to the interface in 4.7 is fine.

Issue #1991, opened 2019-05-07
Compound primary-key not supported


diegoldev:
**Describe the bug**
Trying to set a model with compound primary-key (two or more columns) is not supported by Codeigniter Model and DB Builder

**CodeIgniter 4 version**
4.0.0.beta.3

**Affected module(s)**
Codeigniter Model and DB Builder

**Expected behavior, and steps to reproduce if appropriate**
1) Define a compound primary-key table. (E.g. book_author which pk be: book_id,author_id)
2) Set $this->primaryKey as array ['book_id', 'author_id']
3) Try to get a row with $this->find()
4) An Exception is thrown because 'array cannot be parsed as a string'

**Context**
 - OS: Debian 9
 - Web server: Apache 2
 - PHP version 7.3

nowackipawel:
I did that in my copy of CI4 cuz I needed it.
In fact, I did it by extending the Model class but only for insert / update method with a little dirty code.

As for now, CI4 has been implemented (as I understood from PRs, issues comments) with no support for compound PKs. :(

rino7:
@nowacki sure, I could do by myself but I think that should be supported in a future since other way the framework would force you to use not normalized database/table.

nowackipawel:
@rino7 - that's true.

rino7:
@nowackipawel Will appreciate if you could share your code with me

nowackipawel:
Pls, send me remainder to my email (you will find mine on my profile page). I cannot find yours on your profile page. I will share it in next few days (maybe you made me that way to improve sth, but I don't think so;-)), but please do not except hi-quality code I wrote it to get basic support for that kinda PKs.

nowackipawel:
I have it done with:
```

	/**
	 * @param  $id integer|string|array|null:
	 *         int|string:  WHERE $this->table.$this->primaryKey = $id
	 *         indexed array: WHERE $this->table.$this->primaryKey IN ([2,3]) <i.e.: $id = [2,3]>
	 *         associative array: WHERE $this->table.key1 = 1 AND $this->table.key2 IN([2,3]) <i.e.: $id = [key1 => 1, key2 => [2,3]]>
	 * @return BaseBuilder
	 */
	private function parseId($id = null)
	{
		$builder = $this->builder();

		// is numeric or is non empty string
		if (! empty($id))
		{
			if (is_numeric($id) || is_string($id))
			{
[...]

PR #4115, opened 2021-01-17
Add multi column primary key in model

rakoitde:
**Description**
Codeigniter4 currently only supports one column of primary keys. In some cases, however, it is helpful to be able to use a primary key with several columns.
Known issue "Compound primary-key not supported" #1991

**Usage**
`class myModel extends Model`
`{`
`	protected $table      = 'myTable';`
`	protected $primaryKey = [ 'first_id' , 'second_id' ];`
...

`$mymodel->find(['1','1']);`
`$mymodel->delete(['first_id' => '1', 'second_id' => '1']);`

**Checklist:**
- [ ] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -157,11 +157,35 @@ class Model extends BaseModel
 			$builder->where($this->table . '.' . $this->deletedField, null);
 		}
 
-		if (is_array($id))
+		if (is_array($id) && !is_array($this->primaryKey))
 		{
 			$row = $builder->whereIn($this->table . '.' . $this->primaryKey, $id)
 				->get()
 				->getResult($this->tempReturnType);
+		}
+		if (is_array($id) && is_array($this->primaryKey))
+		{
+
+			if (!is_array($id[0])) {
+				$id = array($id);
+			}
+
+			$row = $builder;
+
+			foreach ($id as $values) {
+
+				$row->orGroupStart();
+
+				for ($i=0; $i < count($this->primaryKey); $i++) { 
+					
+					$row = $row->where($this->table . '.' . $this->primaryKey[$i], $values[$i]);
+				}
+
+				$row = $row->groupEnd();
+			}
+
+			$row = $row->get()
+				       ->getResult($this->tempReturnType);
 		}
 		elseif ($singleton)
 		{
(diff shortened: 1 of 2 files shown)


MGatner:
We cannot change the type on `$primaryKey` without introducing breaking changes. What you could probably do is introduce an alternate model (like "CompoundKeyModel") which was an extension of `Model` with these changes. I'm not sure if proliferating Models is something we would want core to the framework or not, but since you've written the code might as well.

nowackipawel:
I added some code to #1991  I am using it and there are no BC.

MGatner:
@nowackipawel `Model` has since split out `BaseModel` so your code will need some adjustments.

The bigger issue around breaking changes is not whether the modified code works, but whether it will break existing use of the class. For example, I develop `Tatter\Relations` which automates table joins based off Models. Pretend I knew nothing about this PR, now this line in my code becomes a syntax error (https://github.com/tattersoftware/codeigniter4-relations/blob/d033806af70e3eade9baf89ac4ea5c5995b92d59/src/Traits/ModelTrait.php#L276):
```
$ids = array_column($rows, $this->primaryKey);
```

In order to make breaking changes like this we either need to make a major release (version 5), deprecate the current behavior and "phase in" the new (version 4.1+), or find a way to work around it (like my `CompoundModel` example).

nowackipawel:
@MGatner  please check #4120. It seams that it is working in every test case ;-) this could be a good starting point.

MGatner:
I replied to the PR, but basically: it is a good implementation but it is still a breaking change. You will either need to find a way to make a separate class or some way to work with `$primaryKey` as a string (like explode a special character or something).

rakoitde:
Thanks for the evaluation of the PR and suggested solutions.

MGatner:
Yes I think that is a solution. I wonder if we should have some bigger picture conversations about Model in general? I already get confused frequently between BaseModel and Model - maybe there is a way we can do this better in a dedicated namespace?

lonnieezell:
Yes, I think we need to be careful here. The database layer is already complex enough, maybe even more complex than needed currently.

najdanovicivan:
I'm not sure what you mean by model being very complex. The BaseModel is very important for me as I relay on it to use Models/Entities with custom DB drivers for ElasticSearch and MongoDB. The Class naming may not be optimal but so far I only see the benefits having DB layer separated form everything else in the model.

I'll start working on putting up the CompositeModel class. We can decide on actual class naming later on. It should be very similar with the Model class. I'll make this one extend the BaseModel directly so it's completely independent  from the current Model class.  Will work on it today and submit a PR.

Issue #4116, opened 2021-01-17
current_url() global method returning URLs without the index.php part. 


npwsamarasinghe:
CI VERSIONS: CI4 4.04 and develop branch.

I saw someone talking about this issue on the CI forum and thought to give it a try. And yes that method not returning URLs with the index.php part even it's in the indexPage configuration.

MGatner:
Thanks for taking a look at this. I've responded on the PR, but this seems to be intended (according to the User Guide and tests). I acknowledge it does not make much sense, but if it is not a bug then it would be a breaking change. 

We will see what the rest of the team thinks, but an alternative would be to define a new function like "this_url()" that does it the way you would expect.

MGatner:
Per the conversation on the PR this is not a bug, but an opportunity for an enhancement.

MGatner:
I'm running into this issue again trying to verify "active" links on menus. This closer experience is making me realize that the current function is a problem: it generates a URL that will not work in most configurations with an `indexPage`. That said we probably still can't change `current_url()` but I am looking into alternatives and would like to deprecate that function so as not to promulgate the issue.

lonnieezell:
To me this is a bug. Current url should return the current url whether it has `index.php` in it or not. I haven't looked at the code for that in a while, but my guess is it should be using `site_url` instead of `base_url`.

PR #4117, opened 2021-01-17
Fix for current_url() global method not returning index.php part.

npwsamarasinghe:
This PR address this issue: #4116


MGatner (commented):
If `current_url()` is supposed to be equivalent to `site_url(uri_string())` wouldn't it be easier and more consistent to reflect that in the code? I already see a fair amount of duplicate between `site_url()` and your new version of `current_url()`.

MGatner:
Actually looking at your test failures it appears this was the intended behavior. I'm not saying I agree with it, but if the tests and the User Guide both agree that `current_url()` should *not* include the index page it is going to be harder to justify changing.

npwsamarasinghe:
Actually `site_url(uri_string())` is _not_ 100% identical. `current_url()` is support for _full_ version of _current url_. For example, we can pass `boolean true` to this method and it will return _current url_ with the _query string_. Also with this method we can easily access the URI object if we need. So this method is very flexible than `site_url(uri_string())`. And I agree docs should be more descriptive about this.

I think this should not be the default behaviour. This method should respect the `indexPage` configuration. This change should not introduce any issues and Its just only adding `indexPage` to the urls.

CI3 also has this method and its returning urls with the `index.php` part.
[https://codeigniter.com/userguide3/helpers/url_helper.html#current_url](https://codeigniter.com/userguide3/helpers/url_helper.html#current_url)

I highly appreciate if someone can explain how this should not be the default behaviour of this method?

Thanks.

michalsn:
This is definitely a BC. The best we can do now is to add a separate method that will do what you want.

The reason behind this is simple - we can't be sure how users were using this helper. And this was working this way (along with documentation) from the initial release. In the worst-case scenario adding `index.php` to the result string will break someone's code. I can imagine people using this function to dynamically link images for example.

MGatner:
> I highly appreciate if someone can explain how this should not be the default behaviour of this method?

It isn't a matter of "should" but of what is already in place. See @michalsn 's comment above. IMO just make a new function (I proposed `this_url()` but anything appropriate) and build it off `site_url()` to reduce code repetition.

PR #4118, opened 2021-01-17
Precise exception trace

### LINKED ISSUE #4110: Bug: Wrong file/line in exceptions created by "factory" methods
**Direction**

Wrong `$file`, `$line` properties in exceptions created by "factory" methods.

**Describe the bug**

`$file`/`$line` where the exception **was created**.

For example, class [CodeIgniter\Exceptions\FrameworkException](https://github.com/codeigniter4/CodeIgniter4/blob/6ee89af0f8e31bf30a0298d7f558b5acdd447207/system/Exceptions/FrameworkException.php#L24)

~~~php
// file "index.php", line 10:
$e =  FrameworkException::forEnabledZlibOutputCompression();
var_dump($e->getFile(), $e->getLine()); // file '.../FrameworkException.php', line 26
~~~
because  exception created in https://github.com/codeigniter4/CodeIgniter4/blob/6ee89af0f8e31bf30a0298d7f558b5acdd447207/system/Exceptions/FrameworkException.php#L26

~~~php
// file "index.php", line 10:
$e =  new FrameworkException(lang('Core.enabledZlibOutputCompression'));
var_dump($e->getFile(), $e->getLine()); // file 'index.php', line 10
~~~

**CodeIgniter 4 version**
4.04

**Fixes**

1. Convert factory methods to Exception classes:
~~~php
abstract class FrameworkException extends RuntimeException implements ExceptionInterface
{
}

class EnabledZlibOutputCompressionException extends FrameworkException
{
    public function __construct(string $message = 'Core.enabledZlibOutputCompression', int $code = 0, Throwable $previous = null)
    {
          parent::__construct(lang($message), $code, $previous);
    }
}
~~~
2. Set `$file`/`$line` by `debug_backtrace()`, for example:
~~~php
public static function forEnabledZlibOutputCompression()
{
	 $e = new static(lang('Core.enabledZlibOutputCompression'));
         ['file' => $e->file, 'line' => $e->line] = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1)[0];
        return $e;
}
~~~

paulbalandan:
**Description**
When using the normal way of instantiating exceptions, the `getLine` and `getFile` will show where it is raised:
```
// file1.php
$e = new Exception('');
var_dump($e->getFile()); // file1.php
```

However, when using the framework's way of calling exceptions, `getLine` and `getFile` will show where it is instantiated, i.e., inside the exception's factory static methods.
```
// file2.php
$e = FrameworkException::forCopyError();
var_dump($e->getFile()); // FrameworkException.php
```

This PR resolves this by assigning to the `$line` and `$file` property of the exception the `debug_backtrace` function's second array, which is the place where the exception is raised. For this purpose, this creates the `DebugTraceableTrait::trace()` which does that.

Fixes #4110 

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/Exceptions/CLIException.php
+++ b/system/CLI/Exceptions/CLIException.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter\CLI\Exceptions;
 
+use CodeIgniter\Exceptions\DebugTraceableTrait;
 use RuntimeException;
 
 /**
@@ -18,6 +19,8 @@ use RuntimeException;
  */
 class CLIException extends RuntimeException
 {
+	use DebugTraceableTrait;
+
 	/**
 	 * Thrown when `$color` specified for `$type` is not within the
 	 * allowed list of colors.
--- a/system/Cache/Exceptions/CacheException.php
+++ b/system/Cache/Exceptions/CacheException.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter\Cache\Exceptions;
 
+use CodeIgniter\Exceptions\DebugTraceableTrait;
 use RuntimeException;
 
 /**
@@ -18,6 +19,8 @@ use RuntimeException;
  */
 class CacheException extends RuntimeException implements ExceptionInterface
 {
+	use DebugTraceableTrait;
+
 	/**
 	 * Thrown when handler has no permission to write cache.
 	 *
--- a/system/Database/Exceptions/DataException.php
+++ b/system/Database/Exceptions/DataException.php
@@ -11,10 +11,13 @@
 
 namespace CodeIgniter\Database\Exceptions;
 
+use CodeIgniter\Exceptions\DebugTraceableTrait;
 use RuntimeException;
 
 class DataException extends RuntimeException implements ExceptionInterface
 {
+	use DebugTraceableTrait;
+
 	/**
 	 * Used by the Model's trigger() method when the callback cannot be found.
 	 *
--- a/system/Encryption/Exceptions/EncryptionException.php
+++ b/system/Encryption/Exceptions/EncryptionException.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter\Encryption\Exceptions;
 
+use CodeIgniter\Exceptions\DebugTraceableTrait;
 use CodeIgniter\Exceptions\ExceptionInterface;
 use RuntimeException;
 
@@ -19,6 +20,8 @@ use RuntimeException;
  */
 class EncryptionException extends RuntimeException implements ExceptionInterface
 {
+	use DebugTraceableTrait;
+
 	/**
 	 * Thrown when no driver is present in the active encryption session.
 	 *
--- a/system/Exceptions/CastException.php
+++ b/system/Exceptions/CastException.php
@@ -16,6 +16,8 @@ namespace CodeIgniter\Exceptions;
  */
 class CastException extends CriticalError
 {
+	use DebugTraceableTrait;
+
 	/**
 	 * Error code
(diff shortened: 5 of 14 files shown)


MGatner (approved):
Looks good to me. @WinterSilence does this address your issue?

WinterSilence:
@MGatner I thought about it again and find better solution:
~~~php
class FrameworkException extends RuntimeException
{
    public function __construct(string $message, int $code = 0, Throwable $previous = null)
    {
        parent::__construct($message, $code, $previous);
        $trace = $this->getTrace()[0];
        if (isset($trace['class']) && $trace['class'] === get_class($this)) {
            $this->file = $trace['file'];
            $this->line = $trace['line'];
        }
    }
    // factory methods not changed
}
~~~

WinterSilence:
I created issue #4110 to discuss before any PRs. For example, need to fix properties `trace` and `string` or not?

paulbalandan:
That could be another solution but not only FrameworkException is using the factory pattern. There are other exceptions merely extending \RuntimeException. If we'd be using this, then we have to change all other constructors with the same code, which is against DRY.

WinterSilence:
@paulbalandan then we can move it in trait

paulbalandan:
> @paulbalandan then we can move it in trait

If I get it right, the trait would look like this?
```php
protected static function trace(string $message, ...$args)
{
    $e = new static($message, ...$args);
    $trace = $e->getTrace()[0];
    if (isset($trace['class']) && $trace['class'] === static::class)
    {
            ['file' => $e->file, 'line' => $e->line] = $trace;
    }
}
```

If not, I'm all ears for the alternative solution.

WinterSilence:
@paulbalandan nope
~~~php
trait ExceptionTrait
{
    public function __construct(string $message, int $code = 0, Throwable $previous = null)
    {
        parent::__construct($message, $code, $previous);

        // fix instance created by "factory" method
        $trace = $this->getTrace()[0];
        if (isset($trace['class']) && $trace['class'] === get_class($this)) {
            ['file' => $this->file, 'line' => $this->line] = $trace;
        }
    }
}

class FrameworkException extends RuntimeException
{
    use ExceptionTrait;
}
~~~

paulbalandan:
Wow. I'm not even aware you can use `__construct` inside traits. Thanks. I'll implement this.

WinterSilence:
@paulbalandan hmm... it's obvious - trait is similar to class

WinterSilence:
@paulbalandan but why you don't want create separated exceptions? #4110 fix 1

paulbalandan:
> @paulbalandan hmm... it's obvious - trait is similar to class

Yes, but thought traits have no access to these methods.

> @paulbalandan but why you don't want create separated exceptions? #4110 fix 1

If you'll ask me, I prefer fix 1. Or even calling the exception directly `throw new FrameworkException('Your zlib.output_compression...');`.  But the pattern for throwing exceptions throughout the framework has always been using factory methods. It was set long time before so I cannot change that behavior.

WinterSilence:
@paulbalandan what is **these** methods?

>  It was set long time before so I cannot change that behavior.

BC troubles are inevitable at refactoring legacy code. append migration sniffs for fix code

paulbalandan:
> @paulbalandan what is **these** methods?

those `public static function forWhatever()` functions. That has been the way ever since.

> 
> > It was set long time before so I cannot change that behavior.
> 
> BC troubles are inevitable at refactoring legacy code. append migration sniffs for fix code

Agree with you on this, but changing those now here will be out of scope of this PR.

PR #4120, opened 2021-01-17
4115 - Add multi column primary key in model

nowackipawel:
Alternative implementation of #4115 

I was using compound primary keys in 4.0.3 with no issues and no braking changes.
This is 1st try to implement it with new features like BaseModel that probably has been introduced in 4.0.4.

// I am wondering how many tests this will break cuz I wrote it in the fly on github editor ;-)

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -849,11 +849,7 @@ abstract class BaseModel
 	 */
 	public function update($id = null, $data = null): bool
 	{
-		if (is_numeric($id) || is_string($id))
-		{
-			$id = [$id];
-		}
-
+		
 		$data = $this->transformDataToArray($data, 'update');
 
 		// Validate data before saving.
@@ -973,11 +969,6 @@ abstract class BaseModel
 	 */
 	public function delete($id = null, bool $purge = false)
 	{
-		if ($id && (is_numeric($id) || is_string($id)))
-		{
-			$id = [$id];
-		}
-
 		$eventData = [
 			'id'    => $id,
 			'purge' => $purge,
@@ -1096,6 +1087,55 @@ abstract class BaseModel
 		return $error['message'] ?? null;
 	}
 
+	
+	
+	/**
+	 * @param  $id integer|string|array|null:
+	 *         int|string:  WHERE $this->table.$this->primaryKey = $id
+	 *         indexed array: WHERE $this->table.$this->primaryKey IN ([2,3]) <i.e.: $id = [2,3]>
+	 *         associative array: WHERE $this->table.key1 = 1 AND $this->table.key2 IN([2,3]) <i.e.: $id = [key1 => 1, key2 => [2,3]]>
+	 * @return BaseBuilder
+	 */
+	private function parseId($id = null)
+	{
+		$builder = $this->builder();
+
+		// is numeric or is non empty string
+		if (! empty($id))
+		{
+			if (is_numeric($id) || is_string($id))
+			{
+				$builder = $builder->where($this->table . '.' . $this->primaryKey, $id);
+			}
+			elseif (is_array($id))
+			{
+				//$id contains an indexed array
+				if (array_keys($id) === range(0, count($id) - 1))
+				{
+					$builder = $builder->whereIn($this->table . '.' . $this->primaryKey, $id);
+				}
+				//$id contains associative array
+				else
+				{
+					foreach ($id as $key => $value)
+					{
+						if (! empty($value) && is_array($value))
+						{
+							$builder = $builder->whereIn($this->table . '.' . $key, $value);
+						}
+						else
+						{
+							$builder = $builder->where($this->table . '.' . $key, $value);
+						}
+					}
+				}
+			}
+		}
+
+		return $builder;
+	}
+
+	
 	// endregion
 
 	// region Pager
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1086,55 +1086,6 @@ abstract class BaseModel
 
 		return $error['message'] ?? null;
 	}
-
-	
-	
-	/**
-	 * @param  $id integer|string|array|null:
-	 *         int|string:  WHERE $this->table.$this->primaryKey = $id
-	 *         indexed array: WHERE $this->table.$this->primaryKey IN ([2,3]) <i.e.: $id = [2,3]>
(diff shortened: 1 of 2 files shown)


MGatner:
@nowackipawel It is a good start on something, but it cannot be a modification to the existing `Model` or `BaseModel`. "Breaking change" does not refer to the tests, it refers to any changes that could cause code somebody's CodeIgniter project to break. Please review here: https://github.com/codeigniter4/CodeIgniter4/blob/b3a72d296062dfc96f02cc1c0a91e35a61871eb6/CONTRIBUTING.md#breaking-changes

(Re-)implementing methods around the possibility of `Model::$primarykey` being an array will break anything that assumes the current `string` type.

michalsn:
The only way I see that this may work is by adding a new variable `Model::primaryKeys`. But for now, this is a BC.

nowackipawel:
@michalsn : gretings from Lusówko - 15 km from Poznan ;-)

najdanovicivan:
This can be done completely without BreakingChange. I had this in plan when I did the BaseModel split. And the implementation is completely doable in model class.

```
protected $primaryKey = ['key1', 'key2'];
```

then each method will take $id as array value containig ['key1val', 'key2val'] note that it can also be array of arrays
[['key1val', 'key2val'], ['key1val', 'key2val'], ['key1val', 'key2val']]


The only methods which need updating are 

doFind()
doFirst()
doInsert()
doInsertBatch()
doUpdate()
doDelete()
idValue()
shouldUpdate()
objectToRawArray()

MGatner:
Changing the type of `$primaryKey` would be a breaking change. This one comes up a lot, I think we should put it on the roadmap for version 5.

najdanovicivan:
@MGatner it wont be a BC since primary key will still take regular string value for single column keys.  And there is no type hint for it in any methods. Type is only stated in PHPDoc

nowackipawel:
> @MGatner it wont be a BC since primary key will still take regular string value for single column keys. And there is no type hint for it in any methods. Type is only stated in PHPDoc

I think the same. That's why I introduced parseId method. In the same time when we add complex key support we reduced lines of code ;-)

MGatner:
The Model classes are tricky because a) they will *always* be extended, probably the most extended class and b) they have magic accessors making all their properties publicly readable. So consider this:
```
<p>This is my model. It's key is <?= $model->primaryKey =?></p>
```
This now breaks with an array. Or any of our Model Events that expect a certain structure to `$data` - I know this array isn't typed but it's contents could be in any dependencies:
```
		$eventData = [
			'id'        => $id,
			'data'      => $this->doFind($singleton, $id),
			'method'    => 'find',
			'singleton' => $singleton,
		];
```

najdanovicivan:
@MGatner yes it can break things if $primaryKey is defined as array. But if this change is done the way I mention $primaryKey will still be regular string.  So all the existing code will not be affected until the end user decide to use it as array in which case the user will be responsible for updating exiting code to work properly with the new implementation.

MGatner:
> if this change is done the way I mention $primaryKey will still be regular string

@najdanovicivan I was looking at [your comment above](https://github.com/codeigniter4/CodeIgniter4/pull/4120#issuecomment-778589265) where `$primaryKey` is most definitely not a string:
```
/**
* The table's primary key.
*
* @var string|string[]
*/
protected $primaryKey = ['key1', 'key2'];
```

If you were referencing something else please specify. And actually, there are a number of theoretical discussions here that I think are confusing things. Let's keep this PR to the proposed changes (which are not acceptable because of BC) and any theory or alternate solutions can move to the original issue: https://github.com/codeigniter4/CodeIgniter4/issues/1991

@nowackipawel if you want to keep this PR open you will need to re-implement it. For example this is not acceptable:
```
if(is_array($this->primaryKey)) {
```

najdanovicivan:
@MGatner the idea behind all of this is that primaryKey can be either string or array of strings

So for the exiting models we still keep the $primaryKey as string. So all the existing functionality works as is.
```
/**
* The table's primary key.
*
* @var string|string[]
*/
protected $primaryKey = 'id' ;
```

But when there is a need to deal with composite keys the primary key will be defined as array
```
/**
* The table's primary key.
*
* @var string|string[]
*/
protected $primaryKey = ['key1', 'key2'];
```

So this check in the model code completely make sense to me

```
is_array($this->primaryKey); //means that composite key is being used 

//Then we have some constrains on $id value if primary key is array

if(is_array($this->primaryKey))
{
  is_array($id); // Must be true
  if(!is_array($id[0])  //Handle special case where there are multiple $id like in find() method
  {
     count($id) === count($this->primaryKey) //Must be true
  }
  else //Means we're handling multiple ids
  {
    foreach($id as $idValue) 
    {
         count(idValue) === count($this->primaryKey) //Must be true
    }
  }
  //So here $id value is defined as array containing the values in same order as defined in $this->primaryKey
  //Which in case of the previous example will be [$key1Value, $key2Value];
}
else 
{
   //Means we have regular key
}
```

I don't know if the explanation is clear enough. I can do some initial implementation of the functionality. I am 100% certain that it will successfully pass al the unit tests. Then it can be tested to see if it will break anything.

MGatner:
The explanation is clear, but just because things pass tests does not mean they are acceptable for breaking changes. Changing the type of a property and all the methods that use it can break a lot of things built on top of the current code.

Another example, I have a module to handle relationship loading automatically. One of the methods gathers the primary keys from `find()` results using `array_column()` which cannot accept an array for the second parameter: https://github.com/tattersoftware/codeigniter4-relations/blob/459d721bcf87566a02ede96f10ceb681653bd415/src/Traits/ModelTrait.php#L276

I happen to be here and know about this proposed change, so if it went through I would know to make that update. But if I didn't and somebody using my library added a Model with a compound key their app would throw.

lonnieezell:
Sounds like survey says this is breaking change, but this never got closed. Closing.

kenjis:
> Another example, I have a module to handle relationship loading automatically. One of the methods gathers the primary keys from `find()` results using `array_column($rows, $this->primaryKey)` which cannot accept an array for the second parameter: https://github.com/tattersoftware/codeigniter4-relations/blob/459d721bcf87566a02ede96f10ceb681653bd415/src/Traits/ModelTrait.php#L276

Adding an enhancement such that `$primaryKey` accepts arrays will not break the existing code that `$primaryKey` is a string.

But your module does not support the enhancement, so if a user writes new code with an array `$primaryKey`, your module will not work.

It seems the enhancement is not a breaking change.

PR #4122, opened 2021-01-17
Introducing new URL helper method called this_url().

npwsamarasinghe:
This PR address #4116 issue.

A new method that mostly same as the current_url(). But based on the site_url() instead of base_url().

Thanks.

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -130,7 +130,7 @@ if (! function_exists('current_url'))
 	 * Returns the full URL (including segments) of the page where this
 	 * function is placed
 	 *
-	 * @param boolean $returnObject True to return an object instead of a strong
+	 * @param boolean $returnObject True to return an object instead of a string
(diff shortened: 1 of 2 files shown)


michalsn:
We need tests, a user guide, and a [changelog](https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/changelogs/v4.0.5.rst) update before we can merge this.

paulbalandan:
@npwsamarasinghe Do not make merge commits in your fork when you update your copy with the latest develop. Such commits will show up in your PRs here.

MGatner:
@npwsamarasinghe I copied the tests from `current_url()` and update them for my expectations from your new method. Some of the results do not meet this expectations. Please either correct the test cases for your intention or fix your method.

paulbalandan:
@MGatner I think this is closable too.

MGatner:
Superseded by #4675

PR #4125, opened 2021-01-18
NoSQL Entity Methods

MGatner:
**Description**
The Entity class takes advantage of some features of SQL databases that do not extend to NoSQL types. In particular, a SQL table has a set schema and thus predictable fields and data types coming out and requirements for writing. With NoSQL this often isn't the case, and anything can go into or come out of a document.

This PR adds some modified version of Entity methods that can help enforce some expected values on an Entity so a NoSQL model could choose to enact a "strict" data model. If these methods are well received I will write tests and User Guide entries for them.

Note bene: These are methods I have been using for months in a Firestore-based project.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -17,6 +17,7 @@ use DateTime;
 use Exception;
 use JsonSerializable;
 use ReflectionException;
+use UnexpectedValueException;
 use stdClass;
 
 /**
@@ -88,13 +89,13 @@ class Entity implements JsonSerializable
 	 * class properties, using any `setCamelCasedProperty()` methods
 	 * that may or may not exist.
 	 *
-	 * @param array $data
+	 * @param array|null $data
 	 *
 	 * @return $this
 	 */
 	public function fill(array $data = null)
 	{
-		if (! is_array($data))
+		if (is_null($data))
 		{
 			return $this;
 		}
@@ -102,6 +103,37 @@ class Entity implements JsonSerializable
 		foreach ($data as $key => $value)
 		{
 			$this->__set($key, $value);
+		}
+
+		return $this;
+	}
+
+	/**
+	 * Extension of `fill()` that verifies all expected
+	 * fields are present. Allows `null` input for new
+	 * empty Entities.
+	 *
+	 * @param array|null $data
+	 *
+	 * @return $this
+	 */
+	public function fillStrict(array $data = null)
+	{
+		if (is_null($data))
+		{
+			return $this;
+		}
+
+		$this->fill($data);
+
+		// Check each $casts field
+		foreach ($this->casts as $field => $type)
+		{
+			// If the type is not nullable then make sure it is set
+			if (strpos($type, '?') === false && ! isset($this->attributes[$field]))
+			{
+				throw new UnexpectedValueException(static::class . ' missing required field ' . $field);
+			}
 		}
 
 		return $this;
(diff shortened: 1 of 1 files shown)


MGatner:
@najdanovicivan I'd particularly like to hear from you on this. I know Mongo *can* have schemas, but I think some of the same issues might be possible?

MGatner:
I've been thinking about this and I have a better solution. Closing this for now.

najdanovicivan:
@MGatner for now we're using setters to handle the data to be of correct type in the $attributes property.  And we're overriding the setAttributes method to properly initialize the data.

I have a set of ideas to make additional improvements to Entity class but still haven't got time to work on it.

What I don't like about casts is that those only work in getters. Also when datamap is being used the getter methods naming do not follow the datamap.

So we'll need some improvements for the data initialization to make sure it's attributes are filled with the correct variable type. 

There are multiple methods which are there to fill the data setAttributes, fill() ... we need to make sure that only one of them is used when data is loaded from db.  I followed the logic from BaseResult which always uses the setAttributes method when I did other driver implementations.

I also had issue setting data from $_POST in the entity as fill() method will only call the magic setter which makes it impossible to fill some properties which are outside the $attributes property.

So I ended up filling the data this way
```
foreach ($data as $key => $value)
		{
			$entity->$key = $value;
		}
```

The issue with datamap is on the other hand very confusing. Here's the example of using it with ElasticSearch where we have fields prefixed with data type. You have to put everything in same index in ElasticSearch in order to be able to use joins
```
protected $datamap = [
    'live' => 'site_live'
];

//Setter method still use the old name as setter doe not follow the datamap it should work with setLive() method
public function setSiteLive($value): self
{
		$this->attributes['site_live'] = filter_var($value, FILTER_VALIDATE_BOOLEAN);
		return $this;
}
```

So we can either modify the $datamap but to avoid the BC we can add separate property $dbmap which will just do the conversion between field names in db. So in the end we get different names in $attributes arrays eg. site_live value get's stored as $this->attributes['live'] 

And one very useful feature will be auto initialization of nested object/object arrays. I'd like to be able to have something like this
```
protected $initTypes = [
    'categories' = '?Articles\Entities\Category[]'
    'user' = 'Users\Entities\User'
    'live' = 'boolean'
];
```

What will this do in the end is init the object having $this->attributes['categories'] as array of Article\Entities\Category objects and $this->attributes['users'] initialized as Users\Entities\User object.  $this->attributes['live'] will be converted to bool at initialization.

There will probably be more changes needed along the way. So i propose we make some sort of ticket with voting where anyone will be able to suggest feature request so that we can vote on what is to be implemented.

MGatner:
This is all good stuff. I haven't had to wrestle with the datamap but it seems like that might need a pass, maybe even bug fixes or clarification about what it should and should not do. 

For attribute validation (part of this PR) I think I decided this belongs in the Model. I would like to expand Model Validation to be able to apply to items coming *out* of the database as well - this will help a lot with "field expectation" in unstructured NoSQL documents. 

I should also add, with Firestore I use `$attributes` on Entity very heavily to provide initial values. I'm not sure we would need a separate "init" property/method since you can provide them manually, unless I misunderstood you?

najdanovicivan:
There are some issues having validation in model eg. nested values in arrays are not being validated properly. I don't like it at all as you might end up changing the model eg using different db. I believe data validation should be done completely inside entity having the method validate() there but that's completely outside the scope for this one.

Init I've mentioned does not have anything to do with default values. I set those in entity constructor where needed.

$initTypes is just a cast which will happen to the $attributes when data is loaded from db. It only extends the default types to be able to cast to custom class or array of items of that class

MGatner:
The advantage of validating database results in the model is that the rules are all present there already (assuming they are the same, which I can't imagine a scenario where they would not be?). It's also a logical pairing since Model already has the whole scaffold for interactive with Validation class. I'm not sure what to think about nested values - I think that's a limitation of the current `Model` since SQL databases (mostly) cannot return them. 

I was not opposed to Entities doing some validation (that's how I am doing on my current Firestore project) until I decide it is a bit of a data handling violation. Currently an Entity can receive *absolutely anything*, then has some tools for defining various ways of presenting that info. Introducing validation creates the new concept of an "invalid Entity", whereas keeping it at the Model gives us the opportunity to decide what to do with "invalid" data (throw exception, skip it / not create the Entity).

PR #4128, opened 2021-01-18
Fix errors/html/error_exception.php

kenjis:
**Description**
`errors/html/error_exception.php` throws error when constants contain resource and can't see the real exception.

```
ErrorException #1

Uncaught ErrorException: htmlspecialchars() expects parameter 1 to be string, resource given in .../app/Views/errors/html/error_exception.php:182 Stack trace: #0 [internal function]: CodeIgniter\Debug\Exceptions->errorHandler(2, 'htmlspecialchar...', '/Users/kenji/wo...', 182, Array) #1 .../app/Views/errors/html/error_exception.php(182): htmlspecialchars(Resource id #4, 8, 'UTF-8') #2 .../vendor/codeigniter4/codeigniter4/system/Debug/Exceptions.php(281): include('/Users/kenji/wo...') #3 .../vendor/codeigniter4/codeigniter4/system/Debug/Exceptions.php(147): CodeIgniter\Debug\Exceptions->render(Object(...\Exception\NotSupportedException), 500) #4 [internal function]: CodeIgniter\Debug\Exceptions->exceptionHandler(Object(...\Exception\NotSupportedException)) #5 {main} thrown 
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Views/errors/html/error_exception.php
+++ b/app/Views/errors/html/error_exception.php
@@ -178,7 +178,7 @@
 							<tr>
 								<td><?= htmlspecialchars($key, ENT_IGNORE, 'UTF-8') ?></td>
 								<td>
-									<?php if (! is_array($value) && ! is_object($value) && ! is_resource($value)) : ?>
+									<?php if (is_string($value)) : ?>
 										<?= htmlspecialchars($value, ENT_SUBSTITUTE, 'UTF-8') ?>
 									<?php else: ?>
 										<?= '<pre>' . print_r($value, true) ?>
@@ -253,7 +253,7 @@
 							<tr>
 								<td><?= htmlspecialchars($key, ENT_IGNORE, 'UTF-8') ?></td>
 								<td>
-									<?php if (! is_array($value) && ! is_object($value)) : ?>
+									<?php if (is_string($value)) : ?>
 										<?= htmlspecialchars($value, ENT_SUBSTITUTE, 'UTF-8') ?>
 									<?php else: ?>
 										<?= '<pre>' . print_r($value, true) ?>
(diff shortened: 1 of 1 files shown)


app/Views/errors/html/error_exception.php
```diff
@@ -178,7 +178,7 @@
 							<tr>
 								<td><?= htmlspecialchars($key, ENT_IGNORE, 'UTF-8') ?></td>
 								<td>
-									<?php if (! is_array($value) && ! is_object($value)) : ?>
+									<?php if (! is_array($value) && ! is_object($value) && ! is_resource($value)) : ?>
```
samsonasik, line 181: 1st parameter on htmlspecialchas must be string, how about use `is_string($value)` ?

app/Views/errors/html/error_exception.php
```diff
@@ -178,7 +178,7 @@
...
 										<?= htmlspecialchars($value, ENT_SUBSTITUTE, 'UTF-8') ?>
 									<?php else: ?>
 										<?= '<pre>' . print_r($value, true) ?>
```
paulbalandan, line 184: Though unrelated, I am just wondering. Should this line have a closing `</pre>` tag? Or it is unnecessary?

Also, there's another `htmlspecialchars` of `$value` at the bottom. Might as well include that.

app/Views/errors/html/error_exception.php
kenjis: Good point!

app/Views/errors/html/error_exception.php
kenjis: `</pre>` should be.

kenjis:
@samsonasik @paulbalandan 
Fixed:
- use `is_string($value)` in all places
- add `</pre>`
- escape `print_r()` output

paulbalandan (approved):

samsonasik (approved):

samsonasik:
@kenjis thank you for the PR. 

PR #4131, opened 2021-01-18
remove useless "raw html"

totoprayogo1916:
**Description**
I think it's useless,
```
.. raw:: html	

  <div class="custom-index container"></div>
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/user_guide_src/source/dbmgmt/migration.rst
+++ b/user_guide_src/source/dbmgmt/migration.rst
@@ -17,10 +17,6 @@ include migrations from all namespaces.
 .. contents::
   :local:
 
-.. raw:: html
-
-  <div class="custom-index container"></div>
-
 ********************
 Migration file names
 ********************
--- a/user_guide_src/source/helpers/cookie_helper.rst
+++ b/user_guide_src/source/helpers/cookie_helper.rst
@@ -8,10 +8,6 @@ cookies.
 .. contents::
   :local:
 
-.. raw:: html
-
-  <div class="custom-index container"></div>
-
 Loading this Helper
 ===================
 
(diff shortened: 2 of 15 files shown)


MGatner:
I don't think these have any purpose. @paulbalandan do you know?

paulbalandan:
The `raw:: html` actually embeds the following HTML into the generated HTML file instead of displaying it. I'm not sure if the div classes are actually used or defined in the user guide's CSS.

paulbalandan:
@totoprayogo1916 can you show a before and after render of the user guide from this PR?

totoprayogo1916:
@paulbalandan : its found at [contributing/documentation](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/documentation.rst)

but when render is not effect ,
after render,  it just create empty div tag like:
image

paulbalandan:
@totoprayogo1916 yes it will just be put there. Can you show me what the page looks like with this div and after removing it. I am thinking maybe this div is controlling the `clear: both` css property. But if there's no noticeable change, I am fine removing those blocks.

totoprayogo1916:
Sample from https://github.com/codeigniter4/CodeIgniter4/pull/4131/files#diff-438c1b41228e87925336e3c0af80780be95bf8bea65ee11efa4ec91fbc89c930

before: 
image

after: 
image

totoprayogo1916:
and no specific CSS here
image

paulbalandan:
Then, it is really useless. 😂

paulbalandan (approved):

PR #4133, opened 2021-01-19
Fix: Removed type hint from documentation for closures from database queries

iRedds:
Based on http://forum.codeigniter.com/thread-74714.html
Using an example from the documentation, the developer might not understand that the BaseBuilder type hint in the closure is a reference to the Codeigniter\Database\BaseBuilder class.
By specifying a BaseBuilder type hint without using "use Codeigniter\Database\BaseBuilder", a developer will receive a type error.

**Checklist:**
- [x] User guide updated

### DIFF
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -300,7 +300,7 @@ methods:
 
     ::
 
-        $builder->where('advance_amount <', function(BaseBuilder $builder) {
+        $builder->where('advance_amount <', function($builder) {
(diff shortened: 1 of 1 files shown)


samsonasik:
I think it is better to mention the use statement instead of removing type hint.

iRedds:
@samsonasik In fact, this type hint does not affect anything. Only those who use the IDE can feel its benefits. Such developers will be able to add a type hint to their code themselves.

michalsn:
Using type hint doesn't change much here, it's true - but I feel like we should promote good practices, and using type hint is one of them.

PR #4135, opened 2021-01-19
prep_url() with https://

totoprayogo1916:
**Description**
Add parameter to force https in `prep_url()`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -522,26 +522,31 @@ if (! function_exists('auto_link'))
 if (! function_exists('prep_url'))
 {
 	/**
-	 * Prep URL - Simply adds the http:// part if no scheme is included.
+	 * Prep URL - Simply adds the http:// or https:// part if no scheme is included.
 	 *
 	 * Formerly used URI, but that does not play nicely with URIs missing
 	 * the scheme.
 	 *
-	 * @param  string $str the URL
+	 * @param  string  $str    the URL
+	 * @param  boolean $secure set true if you want to force https://
 	 * @return string
 	 */
-	function prep_url(string $str = ''): string
+	function prep_url(string $str = '', bool $secure = false): string
 	{
-		if ($str === 'http://' || $str === '')
+		if (in_array($str, ['http://', 'https://', '//', ''], true))
 		{
 			return '';
 		}
 
-		$url = parse_url($str);
-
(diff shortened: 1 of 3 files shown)


system/Helpers/url_helper.php
```diff
@@ -522,26 +522,31 @@ function auto_link(string $str, string $type = 'both', bool $popup = false): str
...
-	 * @param  string $str the URL
+	 * @param  string  $str    the URL
+	 * @param  boolean $secure set true if you want to force https://
 	 * @return string
 	 */
-	function prep_url(string $str = ''): string
+	function prep_url(string $str = '', bool $secure = false): string
 	{
-		if ($str === 'http://' || $str === '')
+		if ($str === 'http://' || $str === 'https://' || $str === '//' || $str === '')
```
paulbalandan, line 536: How about use strict `in_array`?

```suggestion
		if (in_array($str, ['http://', 'https://', '//', '']), true)
```

tests/system/Helpers/URLHelperTest.php
```diff
@@ -1151,10 +1151,37 @@ public function testAutoLinkPopup($in, $out)
 
 	public function testPrepUrl()
```
paulbalandan, line 1152: You can actually combine this with your secure prep url testing by using a data provider and tweaking this test's signature.
```suggestion
	public function testPrepUrl(string $input, string $expected, string $secure)
```

Then your test would just just have 2 assertions. Test the input on both secure and unsecure option. Since we're using data providers, future cases can be flexibly added.

Another, as much as possible, please use the stricter `assertSame` instead of `assertEquals`.

user_guide_src/source/helpers/url_helper.rst
```diff
@@ -356,13 +356,14 @@ The following functions are available:
     This function works the same as :php:func:`url_title()` but it converts all
     accented characters automatically.
 
-.. php:function:: prep_url($str = '')
+.. php:function:: prep_url($str = '', $secure = false)
```
paulbalandan, line 359: Since these are optional arguments you should use the brackets.
```suggestion
.. php:function:: prep_url([$str = ''[, $secure = false]])
```

paulbalandan (approved):

michalsn (approved):

PR #4136, opened 2021-01-19
PHPStan Fixes

MGatner:
**Description**
* Adds **composer.json** to the path list for PHPStan Action
* Specifies the Same Site Cookie string values for PHPStan compliance
* Removes the ignore line for a now-passing check

See https://github.com/codeigniter4/CodeIgniter4/pull/4134

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -304,7 +304,7 @@ class App extends BaseConfig
 	 * set to `None`, `$cookieSecure` must also be set.
 	 *
 	 * @var string
-	 * @phpstan-var 'Lax'|'None'|'Strict'
+       * @var string 'Lax'|'None'|'Strict'
 	 */
 	public $cookieSameSite = 'Lax';
 
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -303,7 +303,7 @@ class App extends BaseConfig
 	 * (empty string) means no SameSite attribute will be set on cookies. If
 	 * set to `None`, `$cookieSecure` must also be set.
 	 *
-	 * @var string
+	 * @var 'Lax'|'None'|'Strict'
 	 */
 	public $cookieSameSite = 'Lax';
 
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -303,7 +303,6 @@ class App extends BaseConfig
 	 * (empty string) means no SameSite attribute will be set on cookies. If
 	 * set to `None`, `$cookieSecure` must also be set.
 	 *
-	 * @var string
        * @var string 'Lax'|'None'|'Strict'
 	 */
 	public $cookieSameSite = 'Lax';
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -303,7 +303,8 @@ class App extends BaseConfig
 	 * (empty string) means no SameSite attribute will be set on cookies. If
 	 * set to `None`, `$cookieSecure` must also be set.
 	 *
-	 * @var 'Lax'|'None'|'Strict'
+	 * @var string
+	 * @phpstan-var 'Lax'|'None'|'Strict'
 	 */
 	public $cookieSameSite = 'Lax';
 
--- a/app/Config/Security.php
+++ b/app/Config/Security.php
@@ -86,7 +86,7 @@ class Security extends BaseConfig
 	 * Defaults to `Lax` as recommended in this link:
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
-	 * @var string
+	 * @var 'Lax'|'None'|'Strict'
 	 */
 	public $samesite = 'Lax';
 }
--- a/app/Config/Security.php
+++ b/app/Config/Security.php
@@ -87,7 +87,7 @@ class Security extends BaseConfig
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
 	 * @var string
-	 * @phpstan-var 'Lax'|'None'|'Strict'
+       * @var string 'Lax'|'None'|'Strict'
 	 */
 	public $samesite = 'Lax';
 }
--- a/app/Config/Security.php
+++ b/app/Config/Security.php
@@ -86,7 +86,6 @@ class Security extends BaseConfig
 	 * Defaults to `Lax` as recommended in this link:
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
-	 * @var string
        * @var string 'Lax'|'None'|'Strict'
 	 */
 	public $samesite = 'Lax';
--- a/app/Config/Security.php
+++ b/app/Config/Security.php
@@ -86,7 +86,8 @@ class Security extends BaseConfig
 	 * Defaults to `Lax` as recommended in this link:
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
-	 * @var 'Lax'|'None'|'Strict'
+	 * @var string
+	 * @phpstan-var 'Lax'|'None'|'Strict'
 	 */
 	public $samesite = 'Lax';
 }
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -99,7 +99,7 @@ class Security implements SecurityInterface
 	 * Defaults to `Lax` as recommended in this link:
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
-	 * @var string
+	 * @var 'Lax'|'None'|'Strict'
 	 */
 	protected $samesite = 'Lax';
 
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -100,7 +100,7 @@ class Security implements SecurityInterface
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
 	 * @var string
-	 * @phpstan-var 'Lax'|'None'|'Strict'
+       * @var string 'Lax'|'None'|'Strict'
 	 */
 	protected $samesite = 'Lax';
 
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -99,7 +99,8 @@ class Security implements SecurityInterface
 	 * Defaults to `Lax` as recommended in this link:
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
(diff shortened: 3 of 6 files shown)


MGatner:
It appears that Rector does not understand explicit strings for PHPDoc types, which PHPStan requires. @samsonasik or @paulbalandan have any suggestions?

MGatner:
E.g.
```
2) app/Config/Security.php
28

29
30
--- Original
31
+++ New
32
@@ -86,7 +86,7 @@
33
 	 * Defaults to `Lax` as recommended in this link:
34
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
35
 	 *
36
-	 * @var 'Lax'|'None'|'Strict'
37
+	 * @var Lax|None|Strict
38
 	 */
39
 	public $samesite = 'Lax';
40
 }
41
```

samsonasik:
That seems related to Format preserving doc, ref :

- https://github.com/nikic/PHP-Parser/issues/344
- https://github.com/rectorphp/rector/issues/4334

I may create a workaround patch for that in rector side, eg in the ContentPatcher class:

https://github.com/rectorphp/rector/blob/133322cc658fcd8d5ba162b4a18c7d9d88eba714/src/PhpParser/Printer/ContentPatcher.php#L142-L147

For now, I think we can do with:

```php
/** @var string 'Lax'|'None'|'Strict' */
```

ref :
- https://getrector.org/demo/5aea2ecd-f2db-4f60-b61d-ccfa06d9c457 
- https://phpstan.org/r/03e63c8d-6d30-4bed-bb76-1ca36a0b679c

paulbalandan:
Try this if you need to be explicit of the type in phpstan.

```php
// for vars
 * @var string
 * @phpstan-var 'Lax'|'None'|'Strict'

// for returns
 * @return object
 * @phpstan-return TObject
```

Seems phpstan understand those pseudo-docblocks just like in psalm.

samsonasik:
@paulbalandan the `@phpstan-var 'Lax'|'None'|'Strict'` is  not work for rector post -processing right now ref https://getrector.org/demo/1d208398-63da-4f18-8e44-00e16b72bb3e

Need a workaround patch in rector side while wait for a real fix in nikic/php-parser side.

paulbalandan:
It actually works for me though. You need to have both `@var` and `@phpstan-var`.

https://getrector.org/demo/7ed7b50f-7193-4f45-bade-322d372d7343

samsonasik:
Ok, that seems a solution, but need different `/* */` block to make work in phpstan ref 

https://phpstan.org/r/15ddc4ad-57a1-4d3b-ac04-bbb2075dda73

vs

https://phpstan.org/r/421f43cf-2eac-4f5f-98c2-929fbca30543

paulbalandan:
No need. You just have a typo in your second phpstan ref

samsonasik:
@paulbalandan oh yeah, you're correct 👍

MGatner:
Thanks for taking a look guys. Can someone either push the final version you agreed on or suggest it in a review?

app/Config/App.php
```diff
@@ -303,7 +303,7 @@ class App extends BaseConfig
 	 * (empty string) means no SameSite attribute will be set on cookies. If
 	 * set to `None`, `$cookieSecure` must also be set.
 	 *
-	 * @var string
+	 * @var 'Lax'|'None'|'Strict'
```
paulbalandan, line 306:
```suggestion
	 * @var string
	 * @phpstan-var 'Lax'|'None'|'Strict'
```

app/Config/Security.php
```diff
@@ -86,7 +86,7 @@ class Security extends BaseConfig
 	 * Defaults to `Lax` as recommended in this link:
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
-	 * @var string
+	 * @var 'Lax'|'None'|'Strict'
```
paulbalandan, line 89:
```suggestion
	 * @var string
	 * @phpstan-var 'Lax'|'None'|'Strict'
```

system/Security/Security.php
```diff
@@ -99,7 +99,7 @@ class Security implements SecurityInterface
 	 * Defaults to `Lax` as recommended in this link:
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
-	 * @var string
+	 * @var 'Lax'|'None'|'Strict'
```
paulbalandan, line 102:
```suggestion
	 * @var string
	 * @phpstan-var 'Lax'|'None'|'Strict'
```

system/Session/Session.php
```diff
@@ -125,7 +125,7 @@ class Session implements SessionInterface
 	 * Cookie SameSite setting as described in RFC6265
 	 * Must be 'None', 'Lax' or 'Strict'.
 	 *
-	 * @var string
+	 * @var 'Lax'|'None'|'Strict'
```
paulbalandan, line 128:
```suggestion
	 * @var string
	 * @phpstan-var 'Lax'|'None'|'Strict'
```

app/Config/App.php
totoprayogo1916: and for empty string?

app/Config/Security.php
samsonasik, line 89:
```suggestion
       * @var string
       * @phpstan-var 'Lax'|'None'|'Strict'
 ```

File: system/Security/Security.php (hunk 3 above)
Comment (samsonasik) [inline system/Security/Security.php:102]:
```suggestion
       * @var string
       * @phpstan-var 'Lax'|'None'|'Strict'
 ```

File: system/Session/Session.php [hunk 5]
```diff
@@ -125,7 +125,7 @@ class Session implements SessionInterface
 	 * Cookie SameSite setting as described in RFC6265
 	 * Must be 'None', 'Lax' or 'Strict'.
```
samsonasik, line 126:
```suggestion
       * @var string
       * @phpstan-var 'Lax'|'None'|'Strict'
 ```

File: app/Config/App.php (hunk 1 above)
Comment (paulbalandan) [reply app/Config/App.php:306]: empty string is really not allowed either in the cookie spec or in the php code. we only use it to define that _no samesite_ will be set. But this behavior is now "obsolete" as MDN reports that modern browsers requires cookies to have `SameSite=Lax` if no samesite is given. This error is thrown in the console.

**Comment by MGatner** [comment]
Thank you both. GitHub is not allowing me to commit revisions for some reason, looks like a bug, but I will get these applied later when I can get back to my server with this branch.

**Comment by totoprayogo1916** [comment]
Is it possible to use this : `@var string[Lax|None|Strict]` ?

**Comment by samsonasik** [comment]
@MGatner @paulbalandan the `@phpstan-var 'Lax'|'None'|'Strict'` seems only working in rector 0.9. I think for now, we can use:

```php
/** @var string 'Lax'|'None'|'Strict' */
```

as I previously suggested. 

@totoprayogo1916 this patch want a quoted doc.

app/Config/App.php
```diff
@@ -304,6 +304,7 @@ class App extends BaseConfig
 	 * set to `None`, `$cookieSecure` must also be set.
 	 *
 	 * @var string
+	 * @phpstan-var 'Lax'|'None'|'Strict'
```
samsonasik, line 307:
```suggestion
       * @var string 'Lax'|'None'|'Strict'
```

app/Config/Security.php
```diff
@@ -87,6 +87,7 @@ class Security extends BaseConfig
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
 	 * @var string
+	 * @phpstan-var 'Lax'|'None'|'Strict'
```
samsonasik, line 90:
```suggestion
       * @var string 'Lax'|'None'|'Strict'
```

system/Security/Security.php
```diff
@@ -100,6 +100,7 @@ class Security implements SecurityInterface
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
 	 * @var string
+	 * @phpstan-var 'Lax'|'None'|'Strict'
```
samsonasik, line 103:
```suggestion
       * @var string 'Lax'|'None'|'Strict'
```

system/Session/Session.php
```diff
@@ -126,6 +126,7 @@ class Session implements SessionInterface
...
 	 *
 	 * @var string
+	 * @phpstan-var 'Lax'|'None'|'Strict'
```
samsonasik, line 129:
```suggestion
       * @var string 'Lax'|'None'|'Strict'
```

app/Config/App.php
```diff
@@ -304,6 +304,7 @@ class App extends BaseConfig
 	 * set to `None`, `$cookieSecure` must also be set.
 	 *
 	 * @var string
```
samsonasik, line 306:
```suggestion
```

app/Config/Security.php
```diff
@@ -87,6 +87,7 @@ class Security extends BaseConfig
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
 	 * @var string
```
samsonasik, line 89:
```suggestion
```

system/Security/Security.php
```diff
@@ -100,6 +100,7 @@ class Security implements SecurityInterface
 	 * @see https://portswigger.net/web-security/csrf/samesite-cookies
 	 *
 	 * @var string
```
samsonasik, line 102:
```suggestion
```

system/Session/Session.php
```diff
@@ -126,6 +126,7 @@ class Session implements SessionInterface
 	 * Must be 'None', 'Lax' or 'Strict'.
 	 *
 	 * @var string
```
samsonasik, line 128:
```suggestion
```

samsonasik:
I am going to fix it right now, I will try to fix the error affected in system/Security/Security.php and system/Session/Session.php    in separate PR

samsonasik:
Thank you all, I created new PR for temporary ignore the phpstan notice in Security and Session class, with `@todo` to clean up in 4.1 for rector 0.9 https://github.com/codeigniter4/CodeIgniter4/pull/4146

MGatner:
Thanks all. Sorry to be MIA finishing this up - special thanks to @samsonasik

PR #4141, opened 2021-01-21
feat: add property accessor to PagerRenderer

kenjis:
**Description**
Add property accessor to PagerRenderer.
To generate Pager HTML that is the same as CI3's.
The template would be like this:
```php
    <?php if ($pager->hasPreviousPage()) : ?>
    <a href="<?= $pager->getFirst() ?>" data-ci-pagination-page="<?= $pager->first ?>" rel="start">&laquo;First</a>
    <a href="<?= $pager->getPreviousPage() ?>" data-ci-pagination-page="<?= $pager->current - 1 ?>" rel="prev">&lt;</a>
    <?php endif ?>
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Pager/PagerRenderer.php
+++ b/system/Pager/PagerRenderer.php
@@ -113,7 +113,7 @@ class PagerRenderer
 			return $this->$key;
 		}
 
-		throw new InvalidArgumentException('No such a property: ' . $key);
+		throw new InvalidArgumentException('No such property: ' . $key);
 	}
 
 	//--------------------------------------------------------------------
--- a/system/Pager/PagerRenderer.php
+++ b/system/Pager/PagerRenderer.php
@@ -11,6 +11,8 @@
 
 namespace CodeIgniter\Pager;
 
+use InvalidArgumentException;
+
 /**
  * Class PagerRenderer
  *
@@ -93,6 +95,25 @@ class PagerRenderer
 		$this->pageCount    = $details['pageCount'];
 		$this->segment      = $details['segment'] ?? 0;
 		$this->pageSelector = $details['pageSelector'] ?? 'page';
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Accessor for properties if they exist.
+	 *
+	 * @param string $key
+	 *
+	 * @return mixed
+	 */
+	public function __get(string $key)
+	{
+		if (property_exists($this, $key))
+		{
+			return $this->$key;
+		}
+
+		throw new InvalidArgumentException('No such a property: ' . $key);
 	}
 
 	//--------------------------------------------------------------------
--- a/tests/system/Pager/PagerRendererTest.php
+++ b/tests/system/Pager/PagerRendererTest.php
@@ -582,7 +582,7 @@ class PagerRendererTest extends \CodeIgniter\Test\CIUnitTestCase
 	public function testPropertyAccessorGetPropertyThatDoesNotExists()
 	{
 		$this->expectException(InvalidArgumentException::class);
-		$this->expectExceptionMessage('No such a property: properyNeverEverExists');
+		$this->expectExceptionMessage('No such property: properyNeverEverExists');
 
 		$uri     = $this->uri;
 		$details = [
--- a/tests/system/Pager/PagerRendererTest.php
+++ b/tests/system/Pager/PagerRendererTest.php
@@ -3,6 +3,7 @@
 namespace CodeIgniter\Pager;
 
 use CodeIgniter\HTTP\URI;
+use InvalidArgumentException;
 
 class PagerRendererTest extends \CodeIgniter\Test\CIUnitTestCase
 {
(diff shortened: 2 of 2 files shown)


system/Pager/PagerRenderer.php
```diff
@@ -97,6 +99,25 @@ public function __construct(array $details)
...
+	 * @return mixed
+	 */
+	public function __get(string $key)
+	{
+		if (property_exists($this, $key))
+		{
+			return $this->$key;
+		}
+
+		throw new InvalidArgumentException('No such a property: ' . $key);
```
paulbalandan, line 116:
```suggestion
		throw new InvalidArgumentException('No such property: ' . $key);
```

paulbalandan (commented):
Please rebase.

tests/system/Pager/PagerRendererTest.php
```diff
@@ -560,4 +561,39 @@ public function testGetNextPageWithSegmentHigherThanZero()
...
+
+		$this->assertEquals(1, $pager->first);
+		$this->assertEquals(3, $pager->current);
+		$this->assertEquals(10, $pager->last);
+	}
+
+	public function testPropertyAccessorGetPropertyThatDoesNotExists()
+	{
+		$this->expectException(InvalidArgumentException::class);
+		$this->expectExceptionMessage('No such a property: properyNeverEverExists');
```
MGatner, line 585:
```suggestion
		$this->expectExceptionMessage('No such property: properyNeverEverExists');
```
Need to update the test case too!

MGatner (changes requested):

iRedds:
Maybe change all properties to public?

kenjis:
@iRedds I don't need public properties. I just need to get (read) them, not need to write them.

iRedds:
@kenjis Who cares. Properties are defined in the constructor.

tests/system/Pager/PagerRendererTest.php
kenjis: @MGatner Thank you. I fixed.

kenjis:
@iRedds I don't need public properties. Do you need public properties?

paulbalandan (approved):

iRedds:
If you don't need public properties, don't change their values.
I don't see any point in this pager as part of the framework. I understand that you are doing them read-on.

But why not just make them public? They only work with a template anyway.

MGatner:
Public properties should be avoided unless the class is intentionally a mutable one (like Entity). I would like less public properties in the framework overall, not more. An accessory is the correct way to handle this.

michalsn:
Since we are working with these properties inside the class, we shouldn't allow them to be set from outside the class as this can lead to problems.

michalsn (approved):

iRedds:
It may be better to use getters in this case than the magic method?

MGatner:
Getters would be the standard way but I believe the point of the PR was to allow backwards-compatibility with the "CI3 way". I don't have a particular issue with the magic method as long as it doesn't mess up the static analysis.

kenjis:
@iRedds 
I just implemented getters with the magic method.
Do you mean having 6 getter methods like `getCurrent()` is better than having `__get()`?

iRedds:
@kenjis In this case, any property added to the class in the future, but not intended for public use, will be automatically accessed using the magic method.
I think this is not entirely correct.

kenjis:
@iRedds 
When a property will be added to the class in the future, but not intended for public use, 
we can restrict accessing the property.

MGatner:
In general I think using getter methods is better. @kenjis I would like to hear how much of the goal of this PR is supporting the CodeIgniter 3 direct access method versus just getting at the data.

michalsn:
The only downside is that "natural" getters for these properties are in most cases already taken, so we would have to come up with something more "complicated". And this is probably the only reason in favor of using the magic method for me.

Having something like `$pager->getPropertyPrevious()` would be more elegant than writing `$pager->current - 1`. However, I'm not sure about the naming convention.

kenjis:
@MGatner What do you mean by `supporting the CodeIgniter 3 direct access method`?

I thought PagerRenderer has data only for pagination rendering, so it is more flexible that the view can access all the properties inf the PagerRenderer.

But in my use case, I don't need all the properties.
I'm migrating CI3 app to CI4, and `default_full.php` has became like this:
```php
<?php

/**
 * @var \CodeIgniter\Pager\PagerRenderer $pager
 */

$pager->setSurroundCount(2);
?>

<p class="pagination">
<?php if ($pager->hasPreviousPage()) : ?>
    <?php if ($pager->current !== 2) : ?>
    <a href="<?= $pager->getFirst() ?>" data-ci-pagination-page="<?= $pager->first ?> rel="start">&laquo;<?= lang('Pager.first') ?></a>
    <?php endif ?>

    <a href="<?= $pager->getPreviousPage() ?>" data-ci-pagination-page="<?= $pager->current - 1 ?>" rel="prev">&lt;</a>
<?php endif ?>

<?php foreach ($pager->links() as $link) : ?>
<?php if ($link['active']) : ?>
    <strong><?= $link['title'] ?></strong>
<?php else : ?>
    <a href="<?= $link['uri'] ?>" data-ci-pagination-page="<?= $link['title'] ?>">
        <?= $link['title'] ?>
    </a>
<?php endif ?>
<?php endforeach ?>

<?php if ($pager->hasNextPage()) : ?>
    <a href="<?= $pager->getNextPage() ?>" data-ci-pagination-page="<?= $pager->current + 1 ?>" rel="next">&gt;</a>

    <?php if ($pager->current !== ($pager->last - 1)) : ?>
    <a href="<?= $pager->getLast() ?>" data-ci-pagination-page="<?= $pager->last ?>"><?= lang('Pager.last') ?>&raquo;</a>
    <?php endif ?>
<?php endif ?>
</p>
```

MGatner:
This class is very confusing because "first", "last", "current" etc are used interchangeably for page numbers and for URIs. I think we might do better to differentiate those, so instead of having `$pager->first` (number) and `$pager->getFirst()` (URL) we have some clearer accessors.

kenjis:
How about these?
```
getFirstPageNumber()
getCurrentPageNumber()
getLastPageNumber()
```

kenjis:
I take down this PR.
To make all properties readable is too open. 
https://github.com/codeigniter4/CodeIgniter4/blob/49520c7862ab7ce42bb7da54df89a04e33d49fa1/system/Pager/PagerRenderer.php#L45

I sent another PR: #4188

PR #4145, opened 2021-01-21
Update docs for Entity 

### LINKED ISSUE #4143: Calling update() with an Entity throws an error
**Describe the bug**
suppose I have the following example:

```php
<input type="hidden" name="id" value=<?=$data->id?>>
<input type="checkbox" name="active" <?=$data->active ? 'checked' : ''?> >
```

If the checkbox is not activated, no key arrives in the request so I assign a default value in an entity

```php
class Test extends Entity
{
    protected $attributes = [
        'active' => 0,
    ];

	protected function setActive(string $val){
        if($val == 'on'){
            $this->attributes['active'] = 1;
        }
        return $this;
    }
}
```

Now I fill the entity and proceed to save the change, if the value is set active it works perfectly but when no value is sent and it is about saving the default value 0 of the entity it sends the following error

```php 
// controller
$test = new Test($this->request->getPost());
$model = model('TestModel');
$model->update($test->id, $test);
```

imagen

I have solved it using it in the following way but with the **update()** method as shown in the documentation that problem is found

```php
$model->where('id', $test->id)
			->set(['active' => $test->active])
			->update();
```

Thank you very much team!

**CodeIgniter 4 version**
dev develop

**Affected module(s)**
model update()

The registry must be updated when you use entities with default values

**Context**
 - OS: Windows 10
 - Web server: apache 2.4.35
 - PHP version 7.2

michalsn:
**Description**
This PR adds some more clarity to the way that Entities are handled when using them with model's methods: `insert()` and `update()`.

Closes #4143

**Checklist:**
- [x] User guide updated

### DIFF
--- a/user_guide_src/source/models/entities.rst
+++ b/user_guide_src/source/models/entities.rst
@@ -114,6 +114,9 @@ When the User is passed to the model's **save()** method, it automatically takes
 and saving any changes to columns listed in the model's **$allowedFields** property. It also knows whether to create
 a new row, or update an existing one.
 
+.. note:: When we are making a call to the ``insert()`` all the values from Entity are passed to the method, but when we
+    call the ``update()``, then only values that have changed are passed.
+
 Filling Properties Quickly
 --------------------------
 


paulbalandan (approved):

samsonasik:
Thank you @michalsn

Issue #3359, opened 2020-07-19
Bug: php spark migrate throws wrongful CodeIgniter\Database\Exceptions\DatabaseException Unable to connect to the database


gosocial2:
**Bug Description**
Impossible to perform migrations through spark (experienced in versions 4.0.2 to 4.0.4) 

Although DB credentials are set in .env and/orin  app/Config/Database.php and successful query results obtained through Controller classes,

`php spark migrate` causes `CodeIgniter\Database\Exceptions\DatabaseException`

Exact command line output:

```

CodeIgniter CLI Tool - Version 4.0.4 - Server-Time: 2020-07-19 05:22:57am

Running all new migrations...
An uncaught Exception was encountered

Type:        CodeIgniter\Database\Exceptions\DatabaseException
Message:     Unable to connect to the database.
Filename:    /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/Database/BaseConnection.php
Line Number: 425

	Backtrace:
						-646 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/Database/BaseConnection.php::initialize
								-1548 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/Database/BaseConnection.php::query
								-1591 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/Database/BaseConnection.php::listTables
								-945 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/Database/MigrationRunner.php::tableExists
								-193 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/Database/MigrationRunner.php::ensureTable
								-128 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/Commands/Database/Migrate.php::latest
								-100 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/CLI/Commands.php::run
								-107 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/CLI/CommandRunner.php::run
								-85 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/CLI/CommandRunner.php::index
								-914 - /Applications/MAMP/ci4projs/deneme/vendor/codeigniter4/framework/system/CodeIgniter.php::_remap
[...]

PR #4147, opened 2021-01-23
Fix for retrieving migration history

### LINKED ISSUE #4142: Bug: PostgreSQL driver issues
I have had this bug (well, I think it is a bug anyways - I was wrong before, plenty of times) for a while now, but was to lazy to even report it. I was just kinda working around it.

I am facing a constant issue with migrations when using postgresql.

The setup is:
Apache 2.4.46
PHP 7.4.14
PostgreSQL 13.1

My config files setup:
```
public $prod = [
	'DSN'      => '',
	'hostname' => '',
	'username' => '',
	'password' => '',
	'database' => '',
	'DBDriver' => '',
	'DBPrefix' => '',
	'pConnect' => false,
	'cacheOn'  => false,
	'cacheDir' => '',
	'charset'  => 'utf8',
	'swapPre'  => '',
	'encrypt'  => false,
	'compress' => false,
	'strictOn' => false,
	'failover' => [],
];
```

My .env setup:
```
database.prod.hostname = localhost
database.prod.database = playground
database.prod.username = postgres
database.prod.password = MyPasswordGoesHere
database.prod.DBDriver = Postgre
database.prod.DBDebug = TRUE
database.prod.schema = public
database.prod.port = 5432
database.prod.DSN = 'pgsql:dbname=playground;port=5432;host=localhost'
```

The above setup works perfectly fine when running the app. Fetching/inserting/updating data using models or query builder is smooth and fast - no issues.

However when I try to do migration here is where the problems start:
`php spark migrate` results in the below:
image

The workaround for this problem is to comment out the DSN part of the .env file, like so:
`#database.prod.DSN = 'pgsql:dbname=playground;port=5432;host=localhost'`

When I do the above, the migrations run almost perfectly. Obviously, this kills the app itself, so once I am done with migration I need to un-comment the DSN line in order to have the app working.

Now, the reason I have written almost over there, is because this work fine the first time I run migration. However when I add new migration file I get the below:
image

Here's how my migration files look like:
image

and here is the _migration_ table in DB:
image

As you can see, from the above screenshots, all of my previous migration files are in the database, all that is missing is the last file in the list _2021-01-20 170310_approval_workflow.php_

So, why on earth is the CI trying to run the migration again for those that have clearly already been run?

michalsn:
**Description**
This PR fixes the way we loading the history of migrations so that all migrations can be checked and not only these from the `default`  group.

Fixes #4142

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Database/MigrateStatus.php
+++ b/system/Commands/Database/MigrateStatus.php
@@ -84,11 +84,6 @@ class MigrateStatus extends BaseCommand
 		$runner = Services::migrations();
 		$group  = $params['g'] ?? CLI::getOption('g');
 
-		if (is_string($group))
-		{
-			$runner->setGroup($group);
-		}
-
 		// Get all namespaces
 		$namespaces = Services::autoloader()->getNamespace();
 
@@ -120,7 +115,7 @@ class MigrateStatus extends BaseCommand
 				continue;
 			}
 
-			$history = $runner->getHistory();
+			$history = $runner->getHistory((string) $group);
 			ksort($migrations);
 
 			foreach ($migrations as $uid => $migration)
--- a/system/Database/MigrationRunner.php
+++ b/system/Database/MigrationRunner.php
@@ -190,7 +190,7 @@ class MigrationRunner
 		}
 
 		// Remove any migrations already in the history
-		foreach ($this->getHistory($this->group) as $history)
+		foreach ($this->getHistory((string) $group) as $history)
 		{
 			unset($migrations[$this->getObjectUid($history)]);
 		}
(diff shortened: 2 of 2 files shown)


MGatner (approved):
Looks good! Does this close that issue though, or just the group discrepancy?

michalsn:
The group discrepancy is probably the only valid issue there. But I will wait for the author's response.

MGatner:
@michalsn Do you think this is related to #3359? And thus to https://github.com/lonnieezell/myth-auth/issues/298 ?

michalsn:
@MGatner Not really. The issues you mentioned seem to be related to the lack of a database extension enabled in the CLI environment. 

In the original issue, I have linked, the problem with the connection is most likely related to some problems with passing environment variables.

MGatner:
The weird thing in these cases is that the migration would actually succeed (create the table) even with the error. I suspect the issues stemmed from multiple sources and somehow got convoluted. Oh well. 

This one all set? Good by me.

michalsn:
Yes... I'm not sure how accurate are some of these reported issues. It seems impossible to have a connection error and successfully processed migrations at the same time 🤷‍♂️ 

This is ready - I will just wait for the confirmation from the original issue creator.

PR #4166, opened 2021-01-27
improve validation if_exist for dot array syntax

jahmedk:
I had a problem with if_exists on dot array syntax, I tried to improve it.


paulbalandan:
Can you provide your use case for this change?

jahmedk:
image

jahmedk:
image

array_key_exists will always return false in this case with dot array validation syntax.

PR #4167, opened 2021-01-27
feat(router): add alternate-content option to route

yassinedoghri:
feat(router): add alternate-content option to route

**Description**

Hey, I discussed this last week with @MGatner on slack. I was wondering if we could have the same route point to different controller methods depending on the user request. For example, a user might request a route with "application/json" to get json data or else be shown html.
Apparently, the router does not support this yet. I was invited to submit a PR, so here it is :)

With many hours trying to understand and tweak the Router, I've come up with a (debatable) solution to extend the route options. Basically, one would have to add an "alternate-content" option to any route with this format:

```php
$routes->add('bar', 'TestController::default', [
    'alternate-content' => [
        'application/json' => 'AnotherTestController::json',
        'application/activity+json' => 'ActivityPubTest::actor',
        'application/ld+json; profile="https://www.w3.org/ns/activitystreams"' => 'ActivityPubTest::actor',
    ],
]);
```
Then, when handling the route, the router would check if there is any alternate content type to match to the request's accept header. If so, before parsing out the controller and method names + params, the "TestController::default" string would be replaced by the controller method string associated with the content-type. I've injected the code were I thought it would make sense and reused some of the negotiator service logic to match the content-types.

To finish, it's my first PR here and I do not have much experience with the CodeIgniter4 code base, so any feedback would be much appreciated, whether it is about the overall solution, tests, naming, style...

Thanks!

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/HTTP/Negotiate.php
+++ b/system/HTTP/Negotiate.php
@@ -317,7 +317,7 @@ class Negotiate
 	 *
 	 * @return boolean
 	 */
-	protected function match(array $acceptable, string $supported, bool $enforceTypes = false, $matchLocales = false): bool
+	public function match(array $acceptable, string $supported, bool $enforceTypes = false, $matchLocales = false): bool
 	{
 		$supported = $this->parseHeader($supported);
 		if (is_array($supported) && count($supported) === 1)
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter\Router;
 
+use Config\Services;
 use CodeIgniter\Exceptions\PageNotFoundException;
 use CodeIgniter\HTTP\Request;
 use CodeIgniter\Router\Exceptions\RedirectException;
@@ -408,6 +409,10 @@ class Router implements RouterInterface
 			// Does the RegEx match?
 			if (preg_match('#^' . $key . '$#u', $uri, $matches))
 			{
+				$this->matchedRouteOptions = $this->collection->getRoutesOptions(
+					$matchedKey
+				);
+
 				// Is this route supposed to redirect to another?
 				if ($this->collection->isRedirect($key))
 				{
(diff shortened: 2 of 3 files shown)


system/HTTP/Negotiate.php
```diff
@@ -317,7 +317,7 @@ public function parseHeader(string $header): array
 	 *
 	 * @return boolean
 	 */
-	protected function match(array $acceptable, string $supported, bool $enforceTypes = false, $matchLocales = false): bool
+	public function match(array $acceptable, string $supported, bool $enforceTypes = false, $matchLocales = false): bool
```
MGatner, line 320: You cannot change the scope because it will break any classes that extend this: https://3v4l.org/13KBX

Need to find another way around.

MGatner (changes requested):
Thank you for the PR. My initial impression is that it feels a little strange to "override" the `$to` portion in an option. Having not looked into this deeply, is there no way to create Routes with multiple criteria? (In this case it would be "verb - contentType")

One breaking change noted in-line. This would also need coverage in the User Guide but let's get the community to weigh in and figure out the best approach so you don't end up redoing/wasting work.

iRedds:
It seems to me that header level routing is overkill. 
Header-based business logic can also be implemented in a controller.

yassinedoghri:
> Thank you for the PR. My initial impression is that it feels a little strange to "override" the `$to` portion in an option. Having not looked into this deeply, is there no way to create Routes with multiple criteria? (In this case it would be "verb - contentType")
> 
> One breaking change noted in-line. This would also need coverage in the User Guide but let's get the community to weigh in and figure out the best approach so you don't end up redoing/wasting work.

I thought of changing the routes array structure at first as you mentioned but was reluctant to it because of the breaking change. That's why I went with extending the options as it was a good way of introducing the change without too much hassle.

> It seems to me that header level routing is overkill.
> Header-based business logic can also be implemented in a controller.

That's what I actually did before looking for another solution. I'm working on a project that requires many routes to have different content based on the accept header, I was wondering if the router provided something as such. Now that I've dug a bit more into the code, I guess it could be done with a filter like @MGatner initially directed me to. I'll try it.

Thanks for the feedback!

system/HTTP/Negotiate.php
yassinedoghri: Noted, thanks!

MGatner:
I think a Filter is the way to go. It can return an altered Request to impose a new "route" based on the headers. I think it would be awesome to be able to do something like:
```
$routes->post('users/update', 'ApiController::user', ['content' => 'XMLHttpRequest']);
$routes->post('users/update', 'WebController::users/update');
```

But I don't think Routes can support this. 
@yassinedoghri This is good work though, if you think you would still use this in your own project I would recommend releasing it as a module. Now that core Services can be overrides by third-party libraries it shouldn't be too difficult to support.

yassinedoghri:
I played a bit with the filters, correct me if I'm wrong but there's no way to change the controller / method that will be called after the "before" filter, right?
AFAIK, there would be currently no way (apart from doing it in each controller method somehow) of having one route serve multiple content types :/

> ```php
> $routes->post('users/update', 'ApiController::user', ['content' => 'XMLHttpRequest']);
> $routes->post('users/update', 'WebController::users/update');
> ```

@MGatner This would be great, I might override some core services to make this work as you suggested.

MGatner:
@yassinedoghri You're right, the controller and method are set during routing and changing the request does not re-process these. This is something that should likely be changed, but for now it feels like you are backed into a corner and will either need to replace core services or work with redirects or something else less ideal.

paulbalandan:
@MGatner is this PR to be considered in the future or we can close this for now?

Issue #4069, opened 2021-01-05
Bug: db->query returning ResultInterface object for failed queries, doesn't return boolean for write queries


sneakyimp:
**Describe the bug**
If you put your CI4 project in production mode (i.e., CI_ENVIRONMENT=production) then the `BaseConnection::query()` function will return a `ResultInterface` object with `resultID`=`false` when you run a bad query. This appears to be due to [line 662 in system/Database/BaseConnection.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseConnection.php#L662). It is perhaps easy to have overlooked this problem because an exception is usually thrown if you have DBDebug on.

The comments in the source code do not provide much insight into how the `BaseConnection::query()` function should behave. It's also confusing that it can return numerous different data types: `BaseResult|Query|false`. According to the [db query documentation](https://codeigniter.com/user_guide/database/queries.html), this return value is not correct:

> The query() function returns a database result object when “read” type queries are run which you can use to show your results. When “write” type queries are run it simply returns TRUE or FALSE depending on success or failure...

I also don't see any way this function could return `true` under any circumstances. I would also ask if it's wise to be constructing `ResultInterface` classes when `resultID` is a boolean value? The var name `resultID` suggests a resource variable. Perhaps we should revisit the various result classes and make sure we take into account the possibility of a boolean value for resultID?

There should probably include a comprehensive class reference for Database\ConnectionInterface in the documentation somewhere in the documentation. If there is, I haven't been able to find it.

Finally, another minor complication is that mysqli_query actually doesn't return a boolean value for an OPTIMIZE TABLE query, despite what the [PHP documentation](https://php.net/mysqli_query) says:

>Returns false on failure. For successful SELECT, SHOW, DESCRIBE or EXPLAIN queries mysqli_query() will return a mysqli_result object. For other successful queries mysqli_query() will return true.

For an OPTIMIZE TABLE query, mysqli_query returns a mysqli_result object -- whether the optimization succeeds or fails. See #4059. 

**CodeIgniter 4 version**
The current develop branch of CodeIgniter 4 has this problem (as of this writing)

**Affected module(s)**
Database\BaseConnection.php
and anything that utilizes it

**Expected behavior, and steps to reproduce if appropriate**
According to the docs `$query` should be `false` here:
```php
$query = $db->query("SELECT * FROM table_does_not_exist");
```
[...]

PR #4176, opened 2021-01-28
Bug 4069 - query returns Result even when query result is FALSE

sneakyimp:
Initial commit fixing wrong return value in **Database/BaseConnection:query.** When underlying resultID if false, this function should return false as described in #4069. This is a preliminary (and trivial) correction. In order for this change to be run, the db must be configured with DBDebug=false (e.g., in production mode).

NOTE that this code change is based on the [CI 3 behavior](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_driver.php#L659). This may seem like a fundamental change but I truly believe that returning a Result object when the query failed was a mistake because these cast to `true`. I have been unable to detect any change in the behavior of the phpunit tests from this one change, so it's my guess that the phpunit tests do not ever run with DBDebug=false. I've constructed a phpunit test, BadQueryTest which seems to work on my machine, but I have added this line to my .env file:
```
database.tests.DBDebug = false
```

I suspect my BadQueryTest might have trouble with the automate tests that will run, and would ask help (@kenjis ? @michalsn ?) in massaging things here to achieve the right test and make sure this isn't a breaking change. I've only made this change to try and make sure CI4 behaves as the [documentation](https://codeigniter.com/user_guide/database/queries.html#regular-queries) says it will. There are no changes required to the user docs. On the contrary, there is probably still more work to be done to the code to make sure it behaves as the documentation says it does. In particular, the code will not currently return `true` under any circumstances.

**Description**
I changed one line ([line 680](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseConnection.php#L680)) to return `false` instead of a `Result` object. This conforms both to CI3 behavior but also to the behavior described in the CI4 documentation. I added a BadQueryTest which probably needs work.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


sneakyimp:
Looking at the test errors, it would appear that some of the tests expect a `ResultInterface` object to be returned when, according to the docs, no such object is specified -- and, given that the [docs on the db->query functon](https://codeigniter.com/user_guide/database/queries.html#regular-queries) say write queries return a boolean, we should be expecting a boolean:

>The query() function returns a database result object when “read” type queries are run which you can use to show your results. When “write” type queries are run it simply returns TRUE or FALSE depending on success or failure.

The errors from my last commit:
```
There were 7 errors:

1) CodeIgniter\Models\DeleteModelTest::testDeleteFail
ErrorException: Trying to get property 'resultID' of non-object

/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Models/DeleteModelTest.php:30

2) CodeIgniter\Models\DeleteModelTest::testDeleteWithSoftDeleteFail
ErrorException: Trying to get property 'resultID' of non-object

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseBuilder.php:2474
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:418
/home/runner/work/CodeIgniter4/CodeIgniter4/system/BaseModel.php:995
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Models/DeleteModelTest.php:59

3) CodeIgniter\Models\InsertModelTest::testInsertResultFail
ErrorException: Trying to get property 'resultID' of non-object

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:281
/home/runner/work/CodeIgniter4/CodeIgniter4/system/BaseModel.php:751
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:705
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Models/InsertModelTest.php:138

4) CodeIgniter\Models\SaveModelTest::testSaveNewRecordArrayFail
ErrorException: Trying to get property 'resultID' of non-object

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:281
/home/runner/work/CodeIgniter4/CodeIgniter4/system/BaseModel.php:751
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:705
/home/runner/work/CodeIgniter4/CodeIgniter4/system/BaseModel.php:667
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Models/SaveModelTest.php:54

5) CodeIgniter\Models\SaveModelTest::testSaveUpdateRecordArrayFail
ErrorException: Trying to get property 'resultID' of non-object

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseBuilder.php:2474
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:353
/home/runner/work/CodeIgniter4/CodeIgniter4/system/BaseModel.php:887
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:737
/home/runner/work/CodeIgniter4/CodeIgniter4/system/BaseModel.php:663
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Models/SaveModelTest.php:84

6) CodeIgniter\Models\UpdateModelTest::testUpdateResultFail
ErrorException: Trying to get property 'resultID' of non-object

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseBuilder.php:2474
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:353
/home/runner/work/CodeIgniter4/CodeIgniter4/system/BaseModel.php:887
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:737
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Models/UpdateModelTest.php:100

7) CodeIgniter\Database\Live\BadQueryTest::testBadQuery
ErrorException: pg_query(): Query failed: ERROR:  relation "table_does_not_exist" does not exist
LINE 1: SELECT * FROM table_does_not_exist
                      ^

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/Postgre/Connection.php:170
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:715
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:643
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Database/Live/BadQueryTest.php:18
```

For example, in DeleteModelTest, we see this:
```php
	public function testDeleteFail(): void
	{
		$this->setPrivateProperty($this->db, 'DBDebug', false);
		$this->createModel(JobModel::class);
		$this->seeInDatabase('job', ['name' => 'Developer']);

		$result = $this->model->where('name123', 'Developer')->delete();
		$this->assertFalse($result->resultID);
		$this->seeInDatabase('job', ['name' => 'Developer']);
	}
```
Neither [model docs](https://codeigniter.com/user_guide/models/model.html#deleting-data) nor the [query builder docs](https://codeigniter.com/user_guide/database/query_builder.html?highlight=builder#delete) give any details about the return values of their respective delete methods. We can see by reading [the source](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Model.php#L386) that the Model delete method utilizes either the builder::update or builder::delete functions. We can see from the [update](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseBuilder.php#L2439) and [delete](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseBuilder.php#L2827) source code that those functions return the result of `db->query`. As mentioned above, the `db->query` docs should be returning boolean values.

kenjis:
```diff
--- a/tests/system/Database/Live/BadQueryTest.php
+++ b/tests/system/Database/Live/BadQueryTest.php
@@ -15,6 +15,7 @@ class BadQueryTest extends CIDatabaseTestCase
    {
        // this throws an exception in this testing environment, but in production it'll return FALSE
        // perhaps check $this->db->DBDebug for different test?
+       $this->setPrivateProperty($this->db, 'DBDebug', false);
        $query = $this->db->query('SELECT * FROM table_does_not_exist');
        $this->assertEquals(false, $query);
    }
```

```
$ vendor/bin/phpunit tests/system/Database/Live/BadQueryTest.php --no-coverage
PHPUnit 8.5.4 by Sebastian Bergmann and contributors.

Time: 138 ms, Memory: 12.00 MB

OK (1 test, 1 assertion)
```

sneakyimp:
Thanks @paulbalandan for requesting a review from @michalsn

This commit is clearly not ready -- aside from the nearly trivial change to  BaseConnection::query, some classes and tests that use this function will need to be examined and the function should probably be further altered to return `true` for successful write-type queries. This is not an outrageous amount of work, but a requires a fair effort. I'd like to know how folks feel about this change I'm hoping to make. I think it's necessary because

1) The CI4 documentation, presumably published with the 4.0 release, says that's how the function should work, and developers of CI4 projects have likely already written code expecting boolean results for write-type queries.
2) This is how the [CI3 query function](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_driver.php#L574) behaves, so migrations from CI3 to CI4 will be easier (which is my primary concern).

It might be a bit of a chore to return an actual boolean `true` value for all successful write-type queries because if differences in behavior between the various DBMS query functions. We'd probably have to do regex on the query being run to determine whether or not the query is a write-type query or read-type query. For everyone's reference i've constructed two queries to run on the test db:
```sql
-- bad query
UPDATE table_does_not_exist SET name='Developers' WHERE id=1
-- good query
UPDATE db_job SET name='Developer' WHERE id=1
```

The results of the various native DBMS functions are as follows:
| function        | bad           | good  |
| ------------- |-------------| -----|
| mysqli_query| `false` | `true` |
| pg_query | `false` | resource of type pgsql result |
| SQLite3::query | `false` | `true` |
| sqlsrv_query | `false` | resource of type SQL Server Statement |

I was thinking of mimicking the old CI3 function which returns TRUE for successful write-type queries (using a function to detect them) or which can be forced through an optional parameter to return a result object.

Any comments welcome.

kenjis:
I think CI3's behavior is reasonable.
https://codeigniter.com/userguide3/database/db_driver_reference.html#CI_DB_driver::query

The CI4 User Guide says like the following, and no contradictions with it.

> The query() function returns a database result object when “read” type queries are run which you can use to show your results. When “write” type queries are run it simply returns TRUE or FALSE depending on success or failure.
https://codeigniter4.github.io/CodeIgniter4/database/queries.html#regular-queries

michalsn:
Thanks, @sneakyimp for your research on the subject! Sorry that I'm answering so late, but I haven't had time to properly sit and read about this problem earlier.


I guess we can do something like this:
```php
if ($query->isWriteType())
{
    return false;
}

return new $resultClass($this->connID, $this->resultID);
```
and handle the scenario similarly when the query is successful. 

I wouldn't worry about BC, since we're fixing a bug here.

sneakyimp:
To start, I was hoping to define one centralized isWriteType function but the existing CI4 codebase has a bit of a code redundancy issue. It defines isWriteType in three different places, none of which is accessible to the other:
* [SQLite3 Connection class.](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/SQLite3/Connection.php#L498)
* [SQLSRV Connection class](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/SQLSRV/Connection.php#L512)
* [base Query class](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/Query.php#L305)

The SQLite3 and Query classes use an identical regex.  The SQLSRV is almost identical but adds **MERGE** queries as well like the [old CI3 function](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_driver.php#L996). I've taken the liberty of adding a static `Query::sqlIsWriteType()` method to the base Query class -- it seems logical to associate such functionality with this class. I have made all other isWriteType functions refer to this one function.

It looks like there's a phpunit test method defined in tests/system/Database/BaseQueryTest, `testIsWriteType` but I can't seem to find anywhere that this function might actually be called. My knowledge of how the phpunit tests work is somewhat limited, but the definition of that function is the *only* time any PHP source contains the string *testIsWriteType*.

I don't think we need to check `isWriteType()` if the underlying DBMS function has returned false. [CI3 query function](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_driver.php#L574) does not do so, and I can't imagine what we should be returning if a read-type query has resulted in `false` from the underlying DBMS function. Please correct me if I'm overlooking something. 

I'm looking through the source code for references to resultID to determine which should change and am not entirely sure in some cases:

**Mock/MockConnection** 
I'm *guessing* this class is intended to mimic the behavior of an actual db connection so I've add a check to see if the sql is writable and, if so, return a boolean type. Please advise.

I made a couple of changes to the unit tests impacted by the changes. I've also tried to improve the return types specified for the various insert/delete/query functions to reflect the actual possibilities.

sneakyimp:
So it looks like the badQueryTest I've added needs some massage. This behaves very differently depending on whether DBDebug is on or off or which DBMS I'm running. With DBDebug = true, we get a variety of different exceptions thrown when we run this code:
```php
$query = $this->db->query('SELECT * FROM table_does_not_exist');
```
|engine|exception class| mesg |
|---|---|---|
| mysqli | mysqli_sql_exception | Table 'ci4_test.table_does_not_exist' doesn't exist |
| postgres | ErrorException | pg_query(): Query failed: ERROR:  relation "table_does_not_exist" does not exist |
| sqlite3 | ErrorException | SQLite3::query(): Unable to prepare statement: 1, no such table: table_does_not_exist |
| sqlsrv | Exception | [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Invalid object name 'table_does_not_exist'. |

If `DBDebug` is `false` (i.e., ENVIRONMENT is **production**) then all 4 DBMS options return a `$query` value of `false` with no exception thrown. I'm not really sure how to go about construction this test. Any advice would be most welcome.

sneakyimp:
I have modified my BadQueryTest in an attempt to make it check for the `DBDebug` setting and, if it is `true`, just to expect a generic exception:
```php
	public function testBadQuery()
	{
		if ($this->db->DBDebug) {
			// TODO: expect an exception, class and message varies by DBMS
		} else {
			// this throws an exception in this testing environment, but in production it'll return FALSE
			// perhaps check $this->db->DBDebug for different test?
			$query = $this->db->query('SELECT * FROM table_does_not_exist');
			$this->assertEquals(false, $query);
		}
	}
```
I've run this locally and, unfortunately, it gives a warning:
```
There was 1 risky test:

1) CodeIgniter\Database\Live\BadQueryTest::testBadQuery
This test did not perform any assertions

/var/www/ci/tests/system/Database/Live/BadQueryTest.php:14

OK, but incomplete, skipped, or risky tests!
Tests: 1, Assertions: 0, Risky: 1.
```
I wonder if the code should maybe have a try/catch block somewhere to normalize the exception thrown into a single DatabaseException type? Please advise.

sneakyimp:
The commit I just pushed is resulting in a [Rector failure](https://github.com/codeigniter4/CodeIgniter4/pull/4176/checks?check_run_id=1865815598). It looks like Rector is removing a single line? I am fairly mystified at the change it's making, which doesn't seem consistent with the rest of the file -- why would it delete that one blank line when there are so many other identical ones? Also, I've tried running Rector on the system folder locally but it hangs on test 231 of 1148:
```
$ vendor/bin/rector process tests
Rector 0.8.56
Config file: rector.php

  231/1148 [▓▓▓▓▓░░░░░░░░░░░░░░░░░░░░░░░]  20%
```

kenjis:
Thank you for your investigation.

It seems CI4 is not properly designed for exceptions. 
I think database exceptions should be standardized regardless of the driver.

sneakyimp:
> It seems CI4 is not properly designed for exceptions.
> I think database exceptions should be standardized regardless of the driver.

Should we perhaps start a separate issue/bug report for the lack of exception standardization? I don't imagine it would be too difficult to roll that change in with the ones I've already made here.

Also, @kenjis, could you suggest any way to deal with the Rector test that is failing? It's not clear to me what change it wants to make -- if it's just trying to delete a single blank line, that's pretty confusing because there are many other blank lines just like that one already in the file which do not get changed. Also, I'm curious why I cannot get Rector to complete locally. It freezes when I try to run it and will not complete.

kenjis:
```sh-session
$ git fetch codeigniter
$ git checkout develop
$ git merge codeigniter/develop
$ composer update
$ vendor/bin/rector process tests
Rector 0.9.21
Config file: rector.php

  235/1164 [▓▓▓▓▓░░░░░░░░░░░░░░░░░░░░░░░]  20%
```
The progress rate doesn't move from there.

kenjis:
The command was not the same in GitHub Actions.

```sh-session
$ vendor/bin/rector process --dry-run
Rector 0.9.21
Config file: rector.php

 1500/1500 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

                                                                                             
 [OK] Rector is done!                                                                        
 ```

No problem.

@sneakyimp Why don't you rebase this branch?

**Comment by sneakyimp** [comment]
@kenjis I'm confused by your posts. I hope you might clarify some things for me:

> The command was not the same in GitHub Actions.
> 
> ```
> $ vendor/bin/rector process --dry-run
> Rector 0.9.21
> Config file: rector.php
> 
>  1500/1500 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
> 
>                                                                                              
>  [OK] Rector is done!                                                                        
> ```

1) the `--dry-run` flag just tells rector to output the changes rather than altering files. Using this flag in my working directory still hangs.
2) My version of rector is 0.8.56. I've tried `composer update` and `composer upgrade` but neither seem to alter the rector version.


> @sneakyimp Why don't you rebase this branch?

3) As I mentioned before, I am not very familiar with rebasing. Why would that help? Also, is it strictly necessary? I deleted my prior fork and only just created this fork 13 days ago.

Any additional detail would be much appreciated.

**Comment by kenjis** [comment]
@sneakyimp Okay.

Your base CI4 is very old. So your rector's version is old even you run `composer update`.

```diff
Author: Abdul Malik Ikhsan <samsonasik@gmail.com>
Date:   Tue Jan 5 02:10:45 2021 +0700

    [Static Analysis] Use Rector 0.9 for 4.1 branch with php version feature 7.3

diff --git a/composer.json b/composer.json
index 9516cd390..a845905ef 100644
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
        "phpstan/phpstan": "^0.12",
        "phpunit/phpunit": "^9.1",
        "predis/predis": "^1.1",
-       "rector/rector": "^0.8",
+       "rector/rector": "^0.9",
        "squizlabs/php_codesniffer": "^3.3"
    },
    "suggest": {

```

So rebasing helps you. I'm not sure what the rector error is. But it was the past error.
No problem in the current CI4.

And you are not familiar with git. I would show you how to do it.

kenjis:
Clone your repos:
```
git clone git@github.com:sneakyimp/CodeIgniter4.git
cd CodeIgniter4/
```

Add remote branch `codeigniter`:
```
git remote add codeigniter git://github.com/codeigniter4/CodeIgniter4.git
```

Fetch the latest CI4 (remote branch):
```
git fetch codeigniter
```

Checkout your PR branch:
```
git checkout bugs/issue-4069
```

Rebase your branch:
```
git rebase codeigniter/develop
```

sneakyimp:
Thank you! 

> Your base CI4 is very old. So your rector's version is old even you run `composer update`.

I guess 13 days is pretty old. I see that the latest commit to the develop branch is a mere 8 hours ago.

> And you are not familiar with git. I would show you how to do it.

You've been quite helpful. Let me read up on my own and I think I can figure it out.

sneakyimp:
@kenjis I have rebased as you so kindly suggested. Sadly, I cannot run `composer update` because someone changed the composer.json file to require PHP 7.3 or greater. This is problematic for anyone running Ubuntu 18.04 (which will receive maintenance updates until 2023). Its default php 7 installation is 7.2. There is no apt option for php 7.3, 7.4 or 8.

```
$ ./composer.phar update
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires php ^7.3||^8.0 but your php version (7.2.24) does not satisfy that requirement.
  Problem 2
    - phpunit/phpunit[9.1.0, ..., 9.2.6] require php ^7.3 -> your php version (7.2.24) does not satisfy that requirement.
    - phpunit/phpunit[9.3.0, ..., 9.3.8] require php ^7.3 || ^8.0 -> your php version (7.2.24) does not satisfy that requirement.
    - phpunit/phpunit[9.3.9, ..., 9.5.2] require php >=7.3 -> your php version (7.2.24) does not satisfy that requirement.
    - Root composer.json requires phpunit/phpunit ^9.1 -> satisfiable by phpunit/phpunit[9.1.0, ..., 9.5.2].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
```

I think it's a mistake to remove support for php 7.2.

kenjis:
Unfortunately CI 4.1 requires PHP 7.3. And it was decided in the past.

```diff
commit 22e34b190f24ff21735f6dcf7bb2671181234165
Author: MGatner <mgatner@icloud.com>
Date:   Fri Oct 30 18:13:27 2020 +0000

    Bump PHP req, PHPUnit version

diff --git a/admin/framework/composer.json b/admin/framework/composer.json
index df3468dba..e5e88487c 100644
--- a/admin/framework/composer.json
+++ b/admin/framework/composer.json
@@ -5,7 +5,7 @@
    "homepage": "https://codeigniter.com",
    "license": "MIT",
    "require": {
-       "php": "^7.2 || ^8.0",
+       "php": "^7.3||^8.0",
        "ext-curl": "*",
        "ext-intl": "*",
        "ext-json": "*",
@@ -18,7 +18,7 @@
        "codeigniter4/codeigniter4-standard": "^1.0",
        "fakerphp/faker": "^1.9",
        "mikey179/vfsstream": "^1.6",
-       "phpunit/phpunit": "^8.5",
+       "phpunit/phpunit": "^9.1",
        "predis/predis": "^1.1",
        "squizlabs/php_codesniffer": "^3.3"
    },
```

kenjis:
@sneakyimp Can you use `ppa:ondrej/php`?
https://launchpad.net/~ondrej/+archive/ubuntu/php

sneakyimp:
@kenjis thank you very much for the kind suggestion. I'm always reluctant to use non-official repos for my workstation. I ended up upgrading to Ubuntu 20, which has an apt package for PHP 7.4. I still think it's a poor decision to abandon PHP 7.2. Ubuntu 18.04 LTS, which only offers PHP 7.2 as an apt package will provide maintenance updates for two more years and extended security maintenance until 2028.

I have rebased as advised and committed the latest changes. That was *a lot* of work just so rector could remove a single blank line. I still don't understand why rector would take issue with that one empty line when there are at least a dozen others just like it.

Before:
```php
	public function getVersion(): string;

	/**
	 * Orchestrates a query against the database. Queries must use
	 * Database\Statement objects to store the query and build it.
	 * This method works with the cache.
	 *
	 * Should automatically handle different connections for read/write
	 * queries if needed.
	 *
	 * @param string $sql
	 * @param mixed  ...$binds
	 *
	 * @return BaseResult|Query|boolean
	 */
	public function query(string $sql, $binds = null);
```

after:
```php
	public function getVersion(): string;

	/**
	 * Orchestrates a query against the database. Queries must use
	 * Database\Statement objects to store the query and build it.
	 * This method works with the cache.
	 *
	 * Should automatically handle different connections for read/write
	 * queries if needed.
	 *
	 * @param string $sql
	 * @param mixed  ...$binds
	 *
	 * @return BaseResult|Query|boolean
	 */
	public function query(string $sql, $binds = null);

```

kenjis:
It seems your branch is not to be expected status, because it has 168 commits and Files changed 195.

You have the merge commoit: https://github.com/codeigniter4/CodeIgniter4/pull/4176/commits/0cf7b27729b94f2d8339f7ceb3146e375a502688
I'm not sure, but it makes the PR too big?

sneakyimp:
> It seems your branch is not to be expected status, because it has 168 commits and Files changed 195.
> 
> You have the unneeded merge commit: [0cf7b27](https://github.com/codeigniter4/CodeIgniter4/commit/0cf7b27729b94f2d8339f7ceb3146e375a502688)
> I'm not sure, but it makes the PR too big?

I tried rebasing as you suggested -- although I didn't re-clone the repo because I already had it checked out. I haven't the foggiest idea why this PR is so big.

kenjis:
@sneakyimp Do you understand the difference between `merge` and `rebase`?
If not, see https://stackoverflow.com/questions/16666089/whats-the-difference-between-git-merge-and-git-rebase

In this situation, you don't have to use `merge` at all.

After all, you need to fix this PR commits. I would explain it.

kenjis:
It seems you need the commits 3e5558885, 094d0032a, f5994929a, d9f568d8e.

```
$ git log -b bugs/issue-4069 --oneline
d9f568d8e (HEAD -> bugs/issue-4069, sneakyimp/bugs/issue-4069) tiny change made by record...remove of one line in ConnectionInterface
0cf7b2772 Merge branch 'bugs/issue-4069' of https://github.com/sneakyimp/CodeIgniter4 into bugs/issue-4069
f5994929a improved BadQueryTest, simplified return value in BaseUtils::optimizeTable
094d0032a changes to make query() function return boolean for write-type queries. centralize isWriteType functionality in one place. Change tests impacted by this change
3e5558885 Initial commit fixing wrong return value in Database/BaseConnection::query. When underlying resultID if false, this should return false. Added BadQueryTest to tests.
86402521e (origin/develop, origin/HEAD, codeigniter/develop, develop) Merge pull request #4262 from kenjis/fix-github/workflows/test-phpunit
```

sneakyimp:
> @sneakyimp Do you understand the difference between `merge` and `rebase`?
> If not, see https://stackoverflow.com/questions/16666089/whats-the-difference-between-git-merge-and-git-rebase

I am dimly aware of the difference, having read [the documentation here](https://git-scm.com/book/en/v2/Git-Branching-Rebasing). I confess I didn't understand why you suggested that I rebase. I did not understand why i should re-clone a repo that I already had locally, so I attempted to rebase my local repo.

> In this situation, you don't have to use `merge` at all.

I did not use any merge command. I tried to follow your example above except I chose to operate on my local repository instead of cloning again. I forget the actual sequence of commands, but the merge happened automatically at one point. To be clear, I did not manually type any merge commands at all.

> After all, you need to fix this PR commits. I would explain it.

I will certainly need help to sort this out. As I mentioned before, I am not familiar with rebasing.

kenjis:
> I did not understand why i should re-clone a repo that I already had locally

It seems I wrote it wrong. You don't have to re-clone a repo at all. 

I just showed all the flow of the process. You have to clone a repo only once.

kenjis:
Update your `develop` branch:
```
$ git fetch codeigniter
$ git checkout develop
$ git merge codeigniter/develop
```

Create new branch from the latest `develop`:
```
$ git checkout -b new-bugs/issue-4069
```

Cherry-pick needed commits:
```
$ git cherry-pick 3e55588 094d003 f599492 d9f568d
```

Back up the original PR branch: 
```
$ git checkout bugs/issue-4069 
$ git checkout -b bk-bugs/issue-4069
```

Remove the original PR branch:
```
$ git branch -D bugs/issue-4069 
```

Rename the new branch:
```
$ git checkout new-bugs/issue-4069 
$ git branch -M bugs/issue-4069 
```

Force push the new branch:
```
$ git push --force-with-lease origin bugs/issue-4069
```

sneakyimp:
@kenjis Thank you **so much** for your git instructions. It appears to me that this has [solved the issue](https://github.com/codeigniter4/CodeIgniter4/compare/develop...sneakyimp:bugs/issue-4069) and we now have a nice clean commit? I look forward to any comments, critique, or suggestions.

sneakyimp:
@michalsn I believe this PR is possibly ready. If you could review, I would be most grateful. Note that I changed some of the return types declared in the javadoc blocks to better reflect the actual return values.

sneakyimp:
Any word on this PR? I suspect it's getting more stale by the hour, but it's quite fundamental to the database functionality. I wonder if perhaps we should add more unit tests? Any thoughts @michalsn @samsonasik @paulbalandan @kenjis ?

kenjis:
These database related code is difficult to read, because there are so many return types in a function.
My head is confused.

kenjis:
@sneakyimp Could you add explanation in `@return` something like this? 

```
@return BaseResult|Query|boolean When ... returns BaseResult. When ... returns Query. When ... returns true
```

system/Database/SQLSRV/Connection.php
```diff
@@ -511,7 +512,7 @@ public function execute(string $sql)
 	 */
 	public function isWriteType($sql)
 	{
-		return (bool) preg_match('/^\s*"?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX|MERGE)\s/i', $sql);
+		return Query::sqlIsWriteType($sql);
```
kenjis, line 515: It seems it is a bad idea that moving a driver specific logic to common  `Query::sqlIsWriteType()`.
The regex is specific for the driver, not all the same.

kenjis:
> Should we perhaps start a separate issue/bug report for the lack of exception standardization? I don't imagine it would be too difficult to roll that change in with the ones I've already made here.

I think we should. But after this PR is finished.

sneakyimp:
@kenjis thank you, again. you are extremely helpful.

> These database related code is difficult to read, because there are so many return types in a function.
> My head is confused.

I could not agree more! My changes to these return types were an attempt to simplify them from their prior state, which is very complicated and confusing. In particular, the use of [pretend](https://github.com/bcit-ci/CodeIgniter/blob/324628c27ca82e89d5e3a85034127835d29dd9fc/system/database/DB_driver.php#L996) in the query function seems confusing. I don't think the query function should be returning a query object under any circumstances, but I'm not familiar enough with the role or use of this property to gut this functionality.

>Could you add explanation in @return something like this?

I could certainly try. Perhaps someone can explain to me what [pretend](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseConnection.php#L1719) is all about?

>It seems it is a bad idea that moving a driver specific logic to common Query::sqlIsWriteType().
>The regex is specific for the driver, not all the same.

I took a pretty hard look at the function and it seemed safe enough to me. It's basically just slightly modified version of the [CI3 function CI_DB_Driver::is_write_type method](https://github.com/bcit-ci/CodeIgniter/blob/develop/system/database/DB_driver.php#L996). In CI3, some of the specific DBMS drivers override this definition, e.g., [postgre_driver](https://github.com/bcit-ci/CodeIgniter/blob/324628c27ca82e89d5e3a85034127835d29dd9fc/system/database/drivers/postgre/postgre_driver.php#L274). If you think it's wise, I can emulate the CI3 behavior -- although I do not know personally why PostGreSQL would return anything but a boolean value for an INSERT/UPDATE function. I guess it's the RETURNING part there which alters the behavior.

kenjis:
> Perhaps someone can explain to me what pretend is all about?

It is for prepared statements.
https://codeigniter4.github.io/CodeIgniter4/database/queries.html#prepared-queries

kenjis:
> although I do not know personally why PostGreSQL would return anything but a boolean value for an INSERT/UPDATE function.

What do you mean? It seems always returns boolean.

system/Database/Query.php
```diff
@@ -297,15 +297,27 @@ public function getErrorMessage(): string
...
+	/**
+	 * Determines if the statement is a write-type query or not.
+	 *
+	 * @param string $sql
+	 * @return boolean
+	 */
+	public static function sqlIsWriteType($sql): bool
+	{
+		return (bool) preg_match('/^\s*"?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX|OPTIMIZE|MERGE)\s/i', $sql);
+	}
```
kenjis, line 309: I think this static function is not needed. Because there is `isWriteType()`.

system/Database/Query.php
```diff
@@ -297,15 +297,27 @@ public function getErrorMessage(): string
...
 	 * Determines if the statement is a write-type query or not.
 	 *
 	 * @return boolean
 	 */
 	public function isWriteType(): bool
 	{
-		return (bool) preg_match(
-						'/^\s*"?(SET|INSERT|UPDATE|DELETE|REPLACE|CREATE|DROP|TRUNCATE|LOAD|COPY|ALTER|RENAME|GRANT|REVOKE|LOCK|UNLOCK|REINDEX)\s/i', $this->originalQueryString);
+		return self::sqlIsWriteType($this->originalQueryString);
 	}
```
kenjis, line 321: It is better to delegate the logic to the DB driver.
Because SQL is DBMS specific, so we should not make it be common.

system/Database/Query.php
sneakyimp: Please review my [comment above](https://github.com/codeigniter4/CodeIgniter4/pull/4176#issuecomment-775565181). The current CI4 code already has numerous different places where it needs to check whether an SQL expression reflects a writable type -- not always in a query. I think the check for writable query needs to be static so it can be invoked elsewhere.

sneakyimp:
> > although I do not know personally why PostGreSQL would return anything but a boolean value for an INSERT/UPDATE function.
> 
> What do you mean? It seems always returns boolean.

Apparently not. Please refer to the [CI3 postgre_driver's is_write_type function](https://github.com/bcit-ci/CodeIgniter/blob/324628c27ca82e89d5e3a85034127835d29dd9fc/system/database/drivers/postgre/postgre_driver.php#L274) that I mentioned in my prior post.

kenjis:
>  In CI3, some of the specific DBMS drivers override this definition, e.g., postgre_driver. If you think it's wise, I can emulate the CI3 behavior 

I think CI3 way is better. It should be that a specific DBMS driver defines the definition.
This is because SQL varies a bit from DBMS to DBMS.

kenjis:
> Apparently not. Please refer to the CI3 postgre_driver's is_write_type function that I mentioned in my prior post.

https://github.com/bcit-ci/CodeIgniter/blob/324628c27ca82e89d5e3a85034127835d29dd9fc/system/database/drivers/postgre/postgre_driver.php#L274-L279

When the `if` condition is true, it returns false.

When it is false, it returns `parent::is_write_type($sql)`.

https://github.com/bcit-ci/CodeIgniter/blob/324628c27ca82e89d5e3a85034127835d29dd9fc/system/database/DB_driver.php#L996-L999

`parent::is_write_type($sql)` also return bool.

Am I missing something?

system/Database/Query.php
sneakyimp: As I posted before the CI3 codebase has an [is_write_type function in the base driver class](https://github.com/bcit-ci/CodeIgniter/blob/18718545e0faa98d62df3f65d045d84494ec29ca/system/database/DB_driver.php) and only overrides it in these places:
* [odbc driver](https://github.com/bcit-ci/CodeIgniter/blob/18718545e0faa98d62df3f65d045d84494ec29ca/system/database/drivers/odbc/odbc_driver.php#L299)
* [PDO odbc driver](https://github.com/bcit-ci/CodeIgniter/blob/18718545e0faa98d62df3f65d045d84494ec29ca/system/database/drivers/pdo/subdrivers/pdo_odbc_driver.php#L182)
* [PDO pgsql driver](https://github.com/bcit-ci/CodeIgniter/blob/18718545e0faa98d62df3f65d045d84494ec29ca/system/database/drivers/pdo/subdrivers/pdo_pgsql_driver.php)
* [pgsql driver](https://github.com/bcit-ci/CodeIgniter/blob/18718545e0faa98d62df3f65d045d84494ec29ca/system/database/drivers/postgre/postgre_driver.php)

So we would need to override the baseline function I have now only in the case of Postgres for CI4.

sneakyimp:
> > Apparently not. Please refer to the CI3 postgre_driver's is_write_type function that I mentioned in my prior post.
> 
> https://github.com/bcit-ci/CodeIgniter/blob/324628c27ca82e89d5e3a85034127835d29dd9fc/system/database/drivers/postgre/postgre_driver.php#L274-L279
> 
> When the `if` condition is true, it returns false.
> 
> When it is false, it returns `parent::is_write_type($sql)`.
> 
> https://github.com/bcit-ci/CodeIgniter/blob/324628c27ca82e89d5e3a85034127835d29dd9fc/system/database/DB_driver.php#L996-L999
> 
> `parent::is_write_type($sql)` also return bool.
> 
> Am I missing something?

It looks like we've misunderstood each other. I was expressing surprise that an INSERT or UPDATE query would ever return anything but a boolean value indicating success or failure -- apparently adding RETURNING to the query causes Postgres to return a record set. NOTE: this is the *only* DBMS in CI4 that behaves differently, if we are planning to copy CI3's is_write_type behavior.

kenjis:
I got it!

> The optional RETURNING clause causes INSERT to compute and return value(s) based on each row actually inserted (or updated, if an ON CONFLICT DO UPDATE clause was used). This is primarily useful for obtaining values that were supplied by defaults, such as a serial sequence number.
https://www.postgresql.org/docs/9.6/sql-insert.html

If a user uses `RETURNING`, he/she probably expects the values.

system/Database/Query.php
kenjis: Generally speaking, *static* is not good. It might imply bad design.

How about BaseConnection have `isWriteType()`?

sneakyimp: Thanks for your helpful commentary, @kenjis 

> Generally speaking, _static_ is not good. It might imply bad design.

I'm not sure I agree that static functions are "not good" or "imply bad design." I've never seen any scholarly articles discouraging the use of static functions in a class and would be interested in reading more if you'd like to suggest a link. My reason for using a static function was because I searched the code and found existing isWriteType functions in both QueryInterface and ConnectionInterface classes. I would draw your attention to [my post above](https://github.com/codeigniter4/CodeIgniter4/pull/4176#issuecomment-775565181) where I point out that the current CI4 code currently defines 3 nearly identical isWriteType functions, two in the SQLite3 and SQLSRV connection classes, and once in the base Query class.

> How about `BaseConnection` has `isWriteType()`?

Your suggestion to define this function in the Connection classes seems reasonable, but these classes have no SQL property of their own, so this function would need to receive an `$sql` parameter to check. The reason I thought it should go in the Query class is because queries have SQL associated with them. The QueryInterface class declares this function and it is defined to operate on `$this->originalQueryString`. On the other hand, I just searched and apparently that function is *never* called anywhere in the entire system folder. It is, however, called in [$this->originalQueryString](https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/BaseQueryTest.php#L190). I don't feel like I should necessarily be changing the QueryInterface class, which requires an isWriteType function. If we define the isWriteType function as static in the Connection classes, we could call it from the Query classes without having to instantiate the other class.

I'm going to have to look at this tomorrow. It's late here. In the meantime, I'll need to figure out how to update/merge/rebase my bugs/issue-4069 branch with the latest source from CI4. I had added CI4 to my repo as a remote as a basis for PR #4307 so things are getting pretty complicated in my fork. I expect I'll be confused about the correct git approach.

kenjis:
Is my understanding correct?
class dia

http://www.plantuml.com/plantuml/uml/VP9DIyGm48Rlyoi2xc7tiBru4v5b1K4GeIXuyJBDZZAGZvfCmKNyxwODjg5RsakXpFkyqoSvHC4WdJK0x8J21sfIzzuvqiBUFOxVVa2bvuYHZXsR4mLremGRGzlDy7f6IxjXvgkZSDxku1S0coW1jIXjCCOXFR4LszQG9ITnfV5gaw9IzymvVOQhgZirpANDT20K0iYzB_KJ2rs_5yFqBUHE5olbpiPxao0StqBAlPvRoeMZeTnJhheIoqlaIF7FCmQiCtBQx_TBgluhojF9kAIVtMSlnDkM3JsmGrEN2gV4RLKjmIURvV2Ix4Cosj-aFm00

sneakyimp:
> Is my understanding correct?

To be honest, I'm not entirely sure. CI4 [makes use of reflection to instantiate query classes](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseConnection.php#L614). To be honest, I find this bit of code, in particular the str_replace bit, quite awkward:
```php
		$queryClass = $queryClass ?: $this->queryClass;

		if (empty($this->connID))
		{
			$this->initialize();
		}

		$resultClass = str_replace('Connection', 'Result', get_class($this));
		/**
		 * @var Query $query
		 */
		$query = new $queryClass($this);
```

If you are trying to find out where these classes are instantiated, a quick grep search suggests that we also need to take MockConnection and BasePreparedQuery classes:
```
$ schphp -i 'new .*query' system
system/Test/Mock/MockConnection.php:		$query = new $queryClass($this);
system/Database/BaseConnection.php:		$query = new $queryClass($this);
system/Database/BaseConnection.php:			throw new DatabaseException('You must set the database table to be used with your query.');
system/Database/BaseBuilder.php:			throw new DatabaseException('A table must be specified when creating a new Query Builder.');
system/Database/BaseBuilder.php:		$query = new Query($this->db);
system/Database/BasePreparedQuery.php:		$query = new $queryClass($this->db);
system/Database/BasePreparedQuery.php:			throw new BadMethodCallException('Cannot call getQueryString on a prepared query until after the query has been prepared.');
system/HTTP/Exceptions/HTTPException.php:		return new static(lang('HTTP.malformedQueryString'));
system/HTTP/URI.php:	 * Adds a single new element to the query vars.
```
Of all these classes, only MockConnection lacks a BaseConnection object. I believe all the others receive a db object when they are constructed.

If you mean to suggest that we always use $this->db->isWriteType in all those classes, this might not work so well with MockConnection -- I don't know. It seems like a lot of work to make sure we inject a Connection object in all these cases just to make sure we can check isWriteType when a static function would serve perfectly well. It's my understanding that static *variables* are problematic when you have a multithreading/multiprocessing situation, but a static *function* which operates on a *single variable received as a parameter* hardly seems like 'bad design' as you suggested. On the contrary, the various Connection classes can make this function available to our entire codebase without threading some connection object through all the necessary constructors -- which are hard to identify due to the use of Reflection.

kenjis:
In the `__construct()`:
https://github.com/codeigniter4/CodeIgniter4/blob/6e1163f57bc65413f215ef0a34c1cedf4f127952/system/Database/BaseConnection.php#L313-L318
It means `$this->queryClass` is `CodeIgniter\Database\Query` or DBMS specific `Query` if exists.

And
https://github.com/codeigniter4/CodeIgniter4/blob/6e1163f57bc65413f215ef0a34c1cedf4f127952/system/Database/BaseConnection.php#L612-L614
`query()` can change the Query class at the run time. For what?

kenjis:
DBMS specific Query classes are assumed. But `BaseBuilder` create an instance of `Query`.

https://github.com/codeigniter4/CodeIgniter4/blob/6e1163f57bc65413f215ef0a34c1cedf4f127952/system/Database/BaseBuilder.php#L1834

Is this a bug?

sneakyimp:
> `query()` can change the Query class at the run time. For what?

Good question! I did a quick grep search and I don't *think* that `$queryClass` parameter is ever used:
```
$ schphp "\->query([^)]*," .
./system/Database/SQLite3/Connection.php:		if (($query = $this->query('PRAGMA TABLE_INFO(' . $this->protectIdentifiers($table, true, null,
./system/Database/BaseBuilder.php:		$query = $this->db->query($sql, null, false);
./system/Database/BaseBuilder.php:		$result = $this->db->query($sql, $this->binds, false);
./system/Database/BaseBuilder.php:				$this->db->query($sql, $this->binds, false);
./system/Database/BaseBuilder.php:			$result = $this->db->query($sql, $this->binds, false);
./system/Database/BaseBuilder.php:		return $this->testMode ? $sql : $this->db->query($sql, $this->binds, false);
./system/Database/BaseBuilder.php:			$result = $this->db->query($sql, $this->binds, false);
./system/Database/BaseBuilder.php:				$this->db->query($sql, $this->binds, false);
./system/Database/BaseBuilder.php:		return $this->db->query($sql, null, false);
./system/Database/BaseBuilder.php:		return $this->db->query($sql, null, false);
./system/Database/BaseBuilder.php:		return $this->testMode ? $sql : $this->db->query($sql, $this->binds, false);
./system/Database/BaseBuilder.php:		return $this->db->query($sql, $this->binds, false);
./system/Database/BaseBuilder.php:		return $this->db->query($sql, $this->binds, false);
./system/Database/SQLSRV/Builder.php:		return $this->db->query($sql, $this->binds, false);
./system/Database/SQLSRV/Builder.php:		return $this->db->query($sql, $this->binds, false);
./system/Database/SQLSRV/Builder.php:		$result = $this->db->query($sql, $this->binds, false);
./system/Database/SQLSRV/Builder.php:		return $this->testMode ? $sql : $this->db->query($sql, $this->binds, false);
./system/Database/Forge.php:			if (! $this->db->query(sprintf($ifNotExists ? $this->createDatabaseIfStr : $this->createDatabaseStr, $dbName, $this->db->charset, $this->db->DBCollat)))
./system/Database/Forge.php:		return $this->db->query($this->checkDatabaseExistStr, $dbName)->getRow() !== null;
./system/Database/Forge.php:		if (! $this->db->query(sprintf($this->dropDatabaseStr, $dbName)))
./system/Database/Forge.php:		$result = $this->db->query(sprintf($this->renameTableStr,
./system/Database/BaseUtils.php:		$query = $this->db->query(sprintf($this->optimizeTable, $this->db->escapeIdentifiers($tableName)));
./system/Database/BaseUtils.php:			$res = $this->db->query(sprintf($this->optimizeTable, $this->db->escapeIdentifiers($tableName)));
./system/Database/BaseUtils.php:		$query = $this->db->query(sprintf($this->repairTable, $this->db->escapeIdentifiers($tableName)));
./system/Database/Postgre/Builder.php:		return $this->db->query($sql, $this->binds, false);
./system/Database/Postgre/Builder.php:		return $this->db->query($sql, $this->binds, false);
./system/Session/Handlers/DatabaseHandler.php:			if ($this->db->query("SELECT GET_LOCK('{$arg}', 300) AS ci_session_lock")->getRow()->ci_session_lock)
```

Should we remove it?

system/Database/Query.php
kenjis: @sneakyimp I appreciate your sincere attitude.

I've also never seen any scholarly articles, but I saw blog articles. You can see them by googling "static method harmful".

The main reason is it breaks OOP princeples.
I just googled and found one: 
https://medium.com/@karmanno/why-are-static-embers-and-methods-bad-for-object-oriented-design-b4d0c141512c

I understand that you say static methods with no side-effects is safe.
I don't insist never use static method.

I think where the method should be.

sneakyimp:
> DBMS specific Query classes are assumed. But `BaseBuilder` create an instance of `Query`.
> 
> https://github.com/codeigniter4/CodeIgniter4/blob/6e1163f57bc65413f215ef0a34c1cedf4f127952/system/Database/BaseBuilder.php#L1834
> 
> Is this a bug?

Perhaps, perhaps not? The QueryInterface class declares a `getQuery` function which is used in that function you've linked,  `BaseBuilder::compileFinalQuery`. I don't see one of the DBMS-specific query classes overriding this function:

```
$ egrep -r "ion(\s)+getQuery" system
system/Database/Query.php:	public function getQuery(): string
system/Database/BasePreparedQuery.php:	public function getQueryString(): string
system/Database/PreparedQueryInterface.php:	public function getQueryString(): string;
system/Database/QueryInterface.php:	public function getQuery();
system/HTTP/URI.php:	public function getQuery(array $options = []): string
```
The only place the `getQuery` function is defined in the db-related classes is in the Query class.

kenjis:
> I don't see one of the DBMS-specific query classes overriding this function:

Yes, there are no DBMS-specific query classes yet.

kenjis:
I updated the class diagram.



http://www.plantuml.com/plantuml/uml/VLDTQy8m57tFhxYmmXGqBtkJCSJ1G1XZCc4llgJngXbfeiajM8R_VMb2NIYbfL2QtFDnJjEb9-6ehZHZoX2wWv08hzOObAIiMT_NVXc4QoKyhcgbz-XW0IHAZPCYFZv4XTCuSwdHDHDFJfaZ5FwYPr2KokpzbDqO4sNO4P90Qk5z9FoN0rMTDLPeo8-OU2W2b3QjpBZW33ZdquJ1AygQC1-BVde6E7-1o1pYIOU2aB7az6lphWYVTjamNWc3J3z4SbbQgmE9yjykuBRD6IV9oh2JhFzae9zwaE_MkOkGhBuU1nldI6Wxdm-rtabyMaz7Xxiqt6LFRwFqnXA2KySJWJtqQ1QmFIaFuVwfFO5eInLj2hupQuAzHMyUAIogiDiWySvDVXAJUkepC4fRdPN6DsM4ZYefP8vuvdmewlsZvyE3Ps8PmhT_pHy0

kenjis:
`BaseConnection::execute()` is protected.
Why is `SQLite3\Connection::execute()` is public?

system/Database/Query.php
sneakyimp: > The main reason is it breaks OOP princeples.
> I just googled and found one:
> https://medium.com/@karmanno/why-are-static-embers-and-methods-bad-for-object-oriented-design-b4d0c141512c

I have tried to understand his points in this article, and I think we satisfy his guidelines for use of static methods:

>Use static methods, when:
>* You need to separate extremely reusable code, which is tightly linked with usage of some tookit or set of classes
>* You need to create a functionality that is tightly coupled with the implementation but isn't linked with any instance. For example data-creation methods or enum converting methods

I would add that the static methods I propose receive as parameter a simple string and return a boolean value. These functions will make no reference at all to objects and will make no assumption about object properties or array keys which may or may not be defined.

I agree that the BaseConnection and DBMS-specific Connection classes would be a good place for this function. I would point out that these various classes do not have any internal SQL property which we might check. Our method would need to receive the `$sql` to be checked as a parameter. Because this function makes no reference to any class properties, it seems an ideal candidate for a static method.

kenjis:
I am still not sure where the knowledge of whether a query is write-type or not should be.

But CI4 also should handle the PostgreSQL `RETRUNING` in CI3: 
https://github.com/bcit-ci/CodeIgniter/blob/324628c27ca82e89d5e3a85034127835d29dd9fc/system/database/drivers/postgre/postgre_driver.php#L276

After all, a DBMS-specific class should have the knowledge of whether a query is write-type or not.
`RETRUNING` is PostgreSQL only. On SQLite we can make a query like:
`INSERT INTO RETRUNING (id) values (1);`
It is necessary that DBMS-specific classes can override the method to judge write-type or not.

sneakyimp:
> I updated the class diagram.

You left out the system/Test/Mock/MockConnection class, which instantiates a Query object but has no Connection object to which it might refer. I'm not sure what your reason is for this diagram? I'm guessing you want to see where we call the `query()` method or instantiate `Query` objects to make sure they have a Connection instance so they can call Connection->isWriteType? I would point out that we needn't worry about that if isWriteType is a static method. As I posted above, I think this is a good use case for a static method because the method refers only to the parameter that is passed in, which is a primitive data type. It also returns a primitive data type.

>BaseConnection::execute() is protected.
>Why is SQLite3\Connection::execute() is public?

I do not know, but I do see some stuff that *might* (or might not) be of concern in the Prepared Query classes:
```
system/Database/SQLite3/PreparedQuery.php:		$this->result = $this->statement->execute();
system/Database/MySQLi/PreparedQuery.php:		return $this->statement->execute();
```

kenjis:
> I'm not sure what your reason is for this diagram?

I just want to understand the current implementation/design. It is difficult to me.

`MockConnection` is a mock just for testing. So it does not matter.

`PreparedQuery` is not for testing, but It is for prepared statements.
We should consider it, but I think we can do it later.

Anyway, I don't know much about it, so I don't include anything in the diagram that can be excluded.

kenjis:
> I do not know, but I do see some stuff that might (or might not) be of concern in the Prepared Query classes:

It seems I must look into Prepared Query classes.

sneakyimp:
@kenjis I am very grateful for your participation. I don't feel that I have any authority to make sweeping changes, and appreciate your expertise and experience. I hope I am not irritating you.

> I just want to understand the current implementation/design. It is difficult to me.

And for me. I'm gradually coming to grips with it.

> `MockConnection` is a mock just for testing. So it does not matter.

It doesn't look like MockConnection uses isWriteType so I agree we can probably ignore MockConnection. Hopefully the unit tests will complain if we introduce a problem.

> `PreparedQuery` is not for testing, but It is for prepared statements.
> We should consider it, but I think we can do it later.

I don't think the changes I'm making will affect PreparedQuery stuff at all. I've taken great are to avoid changing any functions or interfaces used by other classes.

> Anyway, I don't know much about it, so I don't include anything in the diagram that can be excluded.

The diagram is helpful to visualize the interplay of the classes. My approach has been to use `grep` search to detect use of existing functions.

sneakyimp:
> It seems I must look into Prepared Query classes.

Looking at the PreparedQuery classes (only SQLite3 and MySQLi refer to `$this->statement->execute`), it looks like the $this->statement is not a Connection object but rather a primitive DBMS object.

There do appear to be some instances of `$query->execute` in `tests/system/Database/Live/PreparedQueryTest.php` -- but I don't know if these would be broken if we made `QLite3\Connection::execute()` protected.

sneakyimp:
@kenjis, I have made the changes as you suggested -- avoiding any static methods. I had a closer look and the MockConnection, because it extends the `BaseConnection` class, inherits its `isWriteType` method which alleviated my concerns about that method being available wherever needed.

I have added MERGE type queries to the `BaseConnection::isWriteType` function even though it's apparently not supported by MySQLi because this is what CI3 did.

I added an `isWriteType()` declaration to `ConnectionInterface` because the automated tests were complaining about return types in the Query class.

I added a new unit test, Database/Live/WriteTypeQueryTest where I have made an initial attempt at unit tests. This test could probably be enhanced, and it overlaps with Database/BaseQueryTest.php somewhat, but it does test a couple of DBMS-specific isWriteType functions and it tests actual queries generated by the query builder.

In running the unit tests locally, I noticed that it seems like the code assumes the tables/data structures don't exist before the test runs but then doesn't eliminate the tables when tests are complete, so you get some tests failing if you run these tests repeatedly. `SQLITE/AlterTableTest::testModifyColumnSuccess` should probably destroy and recreate the `janky` table to make this test repeatable. `testModifyColumnSuccess` and other such functions also should probably reset the table to make this test repeatable.

tests/system/Database/Live/BadQueryTest.php
```diff
@@ -0,0 +1,28 @@
...
+		if ($this->db->DBDebug) {
+			// expect an exception, class and message varies by DBMS
+			$this->expectException(\Exception::class);
+			$query = $this->db->query('SELECT * FROM table_does_not_exist');
+		} else {
+			// this throws an exception in this testing environment, but in production it'll return FALSE
+			// perhaps check $this->db->DBDebug for different test?
+			$query = $this->db->query('SELECT * FROM table_does_not_exist');
+			$this->assertEquals(false, $query);
+		}
```
kenjis, line 25: In test execution, the `if` condition is true or false, not both.
You should test both cases.

tests/system/Database/Live/DbUtilsTest.php
```diff
@@ -115,8 +115,14 @@ public function testUtilsOptimizeTableFalseOptimizeDatabase()
...
-		$this->expectException(DatabaseException::class);
-		$this->expectExceptionMessage('Unsupported feature of the database platform you are using.');
+		if ($this->db->DBDebug) {
+			$this->expectException(DatabaseException::class);
+			$this->expectExceptionMessage('Unsupported feature of the database platform you are using.');
+			$util->optimizeDatabase();
+		} else {
+			$result = $util->optimizeDatabase();
+			$this->assertFalse($result);
+		}
```
kenjis, line 125: You should test both cases.

kenjis:
> SQLITE/AlterTableTest::testModifyColumnSuccess should probably destroy and recreate the janky table to make this test repeatable. testModifyColumnSuccess and other such functions also should probably reset the table to make this test repeatable.

I sent PR: #4334

tests/system/Database/Live/BadQueryTest.php
sneakyimp: I don't know what you mean?

tests/system/Database/Live/DbUtilsTest.php
sneakyimp: I don't understand your comment here.

tests/system/Database/Live/BadQueryTest.php
kenjis, line 25: In test execution, the if condition is true or false, not both.
You should test both cases, `$this->db->DBDebug` is true and false.

tests/system/Database/Live/DbUtilsTest.php
kenjis, line 125: You should test both cases, `$this->db->DBDebug` is true and false.

kenjis:
@sneakyimp I deleted the previous comments and I re-wrote them.

kenjis:
This is another issue, because CI3 does not have prepared statement functionality.

When we use prepared statements, we prepare and execute a query.
https://codeigniter4.github.io/CodeIgniter4/database/queries.html#executing-the-query

When a prepared query fails, what value should be returned?
The current implementation always returns a Result object.

class dia

http://www.plantuml.com/plantuml/uml/ZPH1Yzim48Nl-XK3CJGBMLrw2wKikr1OA6MN1dh9PMnD4XLPSgGnn9JyzyeMhXt7DHi28Ph9cs-UNlmK63tNfH52MoP_m8BWnLbB1MjdN_-T_H4GNyyOwBdMHf67BJ3cXjPPz_43ItheUiur-MOTs6jxX2oSpGQoN5iL7iHL2CnZ1Gk6mc08dU0m3dHP6Ih9SbY0M6K60x-tOvO7Ra1AcQ1MTA6YPXgtTJmJ02c_GgSSroiy8PCGYVJd-tVDz7a_QgOBarLtIoJAt3aJHNJuvJNJhgbedL3cIKR-9u1XwmMzQq-Nc0Vx9XfYMIFzsZq-phdUZtXpWO-UzgcvtprzMvItZWcyFfuOt64YiuNTIGU8xzzrOC3ML6ota7lxog0S1Vk9uw6ErOPOzZJZ9FQnK3bivpReoidkzbfTMMb3txH5qq4a3yQ2NwISqnaoCMgUZSnmbs-UAlIaPk-qIiNRxABts43cgdPskCsCb39JoFY_lr8_RHg7POnLTaxyQO-HcvDeVnHwqxkTX5yYtz8YT-IpdZwHLUs3wYy0

tests/system/Database/Live/BadQueryTest.php
sneakyimp: I still don't understand. If `$this->db-DBDEbug` casts as `true`, code execution starts at line 17 -- we must expect the exception before executing the query. Otherwise code execution goes to line 23, and we must run the query before we assert that its result is false.

tests/system/Database/Live/DbUtilsTest.php
sneakyimp: I still don't understand what you mean.

sneakyimp:
> This is another issue, because CI3 does not have prepared statement functionality.
> 
> When we use prepared statements, we prepare and execute a query.
> https://codeigniter4.github.io/CodeIgniter4/database/queries.html#executing-the-query
> 
> When a prepared query fails, what value should be returned?
> The current implementation always returns a Result object.

Thank you for digging this deeply into the code, and thank you for your very kind help. I would not be able to contribute without your guidance.

I have not looked into the prepared statement functionality. Would it be acceptable to create a separate issue/bug report for that? I very badly need to get back to work on my CI3 to CI4 site migration which has been stymied by these bugs I've been working on. I could probably look at it after March 1st.

system/Database/Query.php
sneakyimp: As you suggested, the latest changes have moved `isWriteType($sql)` into the BaseConnection class, and it is in some cases overridden by the DMBS-specific Connection classes (see Postgre\Connection and SQLSRV\Connection). The Query classes also have `isWriteType()` functions now, and these use the Connection objects passed into their constructors.

tests/system/Database/Live/BadQueryTest.php
kenjis: When we run the test case, we really test either of below:
- $this->expectException(\Exception::class);
- $this->assertEquals(false, $query);

But we should test both cases.
In other words, we should do:
1. set `$this->db->DBDebug` true and test
2. set `$this->db->DBDebug` false and test

kenjis:
Yeah, I think prepared statement issue is another separate issue/bug.

To be honest, I'm tired of this big PR.
I hope it will be merged as soon as possible.

michalsn (changes requested):
Thanks, both of you @sneakyimp and @kenjis for taking your time and working on this.

Besides small tweaks in tests, it looks like it's almost ready. 

The last thing we need is a changelog update and migration instructions. 

Please make an entry here: https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/changelogs/v4.1.2.rst under `Bugs fixed` - when you will be referring to the fixed return status.

We need to make the migration instructions when it comes to the changed `ConnectionInterface`, so please copy this file: https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/installation/upgrade_410.rst and based on it create `upgrade_412.rst` and add instructions for adding a new method.

tests/system/Database/Live/BadQueryTest.php
sneakyimp: Aha! I think I understand now -- sorry.

I should remove the `if` statement and instead explicitly set `DBDebug=true` and test that and then set it to false and run that test also so that the test performs both tests every time. I was under the mistaken impression that the tests that ran would try it both ways.

sneakyimp: If I understand @kenjis and @michalsn correctly, you are saying I must construct these tests to *actually change the value of DBDebug* to force the test to run under both conditions. Unfortunately, **any change in this value appears to persist between tests and will affect other tests.** I.e., the unit tests start with whatever DBDebug default (default value, from Config, from .env, etc) and if you change DBDebug in one unit test method, that changed value will still be in effect for the following test.

Is there any function to restore the original DBDebug value? I suppose I might alter these unit tests to make note of the original value, but I need some additional information: Are all the tests in `DbUtilsTest` guaranteed to run in the order they are defined in the PHP file? Must I take care to restore the DBDebug value to its default before running each test?

NOTE: I tried defining a class property `$origDebug` and also defining a test to run first, `testFirst()` like so:
```php
	protected $origDebug;

	/**
	 * This test must run first to store the inital debug value before we tinker with it below
	 */
	public function testFirst() {
		$this->origDebug = $this->getPrivateProperty($this->db, 'DBDebug');
		echo "orig in testFirst\n";
		var_dump($this->origDebug);
		
		$this->assertIsBool($this->origDebug);
	}
```
**HOWEVER** it would appear that phpunit instantiates a new instance of `DbUtilsTest` for each and every test method it contains so `$this->origDebug` will be `null` inside every other test function inside it even while any changes to `$this->db->DBDebug` persist between tests.

Please advise! Do I understand your instructions correctly? Is there a best practice for changing DBDebug values? Why are no other live db tests except DEbugTest manipulating this value?

sneakyimp: I have had some success if I make $origDebug a static variable because it then persists between different instances of `DBUtilsTest`:
```php
	protected static $origDebug;

	/**
	 * This test must run first to store the inital debug value before we tinker with it below
	 */
	public function testFirst() {
		$this::$origDebug = $this->getPrivateProperty($this->db, 'DBDebug');
		
		$this->assertIsBool($this::$origDebug);
	}
```
and then if I can be sure these tests will always run in the order they are defined in the PHP file, this should avoid changing DBDebug from whatever default it may have except for the scope of these tests:
```php
	public function testUtilsOptimizeTableFalseOptimizeDatabaseDebugTrue()
	{
		$util = (new Database())->loadUtils($this->db);
		$this->setPrivateProperty($util, 'optimizeTable', false);
		
		// set debug to true -- WARNING this change will persist!
		$this->setPrivateProperty($this->db, 'DBDebug', true);
		
		$this->expectException(DatabaseException::class);
		$this->expectExceptionMessage('Unsupported feature of the database platform you are using.');
		$util->optimizeDatabase();
		
		// this point in code execution will never be reached
	}

	public function testUtilsOptimizeTableFalseOptimizeDatabaseDebugFalse()
	{
		$util = (new Database())->loadUtils($this->db);
		$this->setPrivateProperty($util, 'optimizeTable', false);

		// set debug to false -- WARNING this change will persist!
		$this->setPrivateProperty($this->db, 'DBDebug', false);
		
		$result = $util->optimizeDatabase();
		$this->assertFalse($result);
		
		// restore original value grabbed from testFirst -- WARNING this change will persist!
		$this->setPrivateProperty($this->db, 'DBDebug', $this::$origDebug);
	}

```

system/Database/ConnectionInterface.php
```diff
@@ -193,4 +192,12 @@ public function escape($str);
...
 
 	//--------------------------------------------------------------------
+
+	/**
+	 * Determines if the statement is a write-type query or not.
+	 *
+	 * @param  string $sql
+	 * @return boolean
+	 */
+	public function isWriteType($sql): bool;
```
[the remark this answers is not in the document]
sneakyimp: @michalsn :
I have changed the two tests you pointed out in an attempt to run the tests for `DBDebug=true` and `DBDebug=false`. As I mentioned in this [prior](https://github.com/codeigniter4/CodeIgniter4/pull/4176#discussion_r582159687) [comments](https://github.com/codeigniter4/CodeIgniter4/pull/4176#discussion_r582171805) I had some concern about what effect this might have on other unit tests. My solution assumes the tests are executed in the order they are defined in a given unit test file.

I have modified the changelog and upgrade documents as you requested. I hope that these changes satisfy your change request.

tests/system/Database/Live/BadQueryTest.php
kenjis: >  I.e., the unit tests start with whatever DBDebug default (default value, from Config, from .env, etc) and if you change DBDebug in one unit test method, that changed value will still be in effect for the following test.

Yes. Ideally, all the state before a test run should be completely identical. 

But currently, CI4 tests have a lot of hidden dependencies, and it will be difficult to control them perfectly,
so the best approach is to save the DBDebug value before execution and restore it.

kenjis: > HOWEVER it would appear that phpunit instantiates a new instance of DbUtilsTest for each and every test method it contains so $this->origDebug will be null inside every other test function inside it even while any changes to $this->db->DBDebug persist between tests.

PHPUnit resets properties in a test case, because the state before each test should be identical
and ideally a test should not depend on another test.

If you want to keep a value between test methods, you can use static properties or Test Dependencies
https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#test-dependencies

system/Database/ConnectionInterface.php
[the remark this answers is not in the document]
kenjis: @michalsn It looks good to me.
If you have no problems, merge this, please.

tests/system/Database/Live/BadQueryTest.php
sneakyimp: > PHPUnit resets properties in a test case, because the state before each test should be identical
> and ideally a test should not depend on another test.

I did test whether a change to the DBDebug value in one test method would persist when PHPUnit moves on to the next test and it did ⚠️⚠️⚠️.

I tried adding a class property that was non-static, `public $origDebug` and I set this value in a test method but, to my surprise, it was NULL in every subsequent test method in that file. This is what made me think it was creating a new instance of the test class for each method. I added a constructor function and this was called repeatedly for a single test file -- presumably once for each test. I surmised from these experiments that the db object would retain any changes made to DBDebug until *all* the subsequent tests had run. PHP passes objects by reference, so if you modify them inside a function, the changes can persist:
```php
class foo {
  public $prop = "default";
}

function myFunc($v) {
  $v->prop = "inside";
  echo $v->prop . "\n";
}

$v = new foo;
echo $v->prop . "\n";

myFunc($v);

echo $v->prop . "\n";
```
output:
```
default
inside
inside
```
 
> If you want to keep a value between test methods, you can use static properties or Test Dependencies
> https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#test-dependencies

I did indeed end up using a `public static $origDebug` var to hold the original/default value for DBDebug, and this seems to work, but I would point out that *this assumes the tests will be executed in the order they are defined in the file*.

kenjis: > I did test whether a change to the DBDebug value in one test method would persist when PHPUnit moves on to the next test and it did ⚠️⚠️⚠️.

`$this->db` comes from the static method (it was moved to trait recently):

CodeIgniter4/DatabaseTestTrait.php at develop · codeigniter4/CodeIgniter4
https://github.com/codeigniter4/CodeIgniter4/blob/e4661270a2b4ce12bbc2ec456ea984e3f149ec36/system/Test/DatabaseTestTrait.php#L72-L76

> this assumes the tests will be executed in the order they are defined in the file.

If I remember correctly, phpunit runs tests in the order.

michalsn (approved):
Looks good. Thanks!

kenjis (approved):

michalsn:
@sneakyimp Thank you for this.

sneakyimp:
> @sneakyimp Thank you for this.

I'm happy to contribute, and feel this is important. I think it was probably overlooked because the phpunit tests get run with `DBDebug=true` ?? We should probably consider a test run with that set to `false` because that's what it will be in production mode, right?

Also, I hope you guys will review & accept PR #4307 . @paulbalandan took a look but I haven't heard anything in five days. I think autorouting is a very serious consideration for code organization, to reduce complicated routes files, for seo optimization, and for security to prevent malicious code execution issues.

PR #4184, opened 2021-01-29
new array helper: array_flatten_with_dots

paulbalandan:

**Description**
Supersedes and closes #4166 .

This PR introduces a new array helper, `array_flatten_with_dots`, to flatten multidimensional arrays using the dot syntax. This is useful for validations involving multi-keys just like the predicament in the linked PR, `if_exist` validation.

to @jahmedk, can you test this one to see if it will solve your issue. I've tested this locally with your test script and the error message is now showing.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -202,7 +202,7 @@ class Validation implements ValidationInterface
 	 * @param string|null  $label
 	 * @param string|array $value Value to be validated, can be a string or an array
 	 * @param array|null   $rules
-	 * @param array        $data  // All of the fields to check.
+	 * @param array        $data  All of the fields to check.
 	 *
 	 * @return boolean
 	 */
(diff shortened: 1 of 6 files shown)


sfadschm:
I like this.
How about a little recursion for arrays with more than two levels?

system/Validation/Validation.php
```diff
@@ -213,15 +213,16 @@ protected function processRules(string $field, string $label = null, $value, $ru
 			throw new InvalidArgumentException('You must supply the parameter: data.');
 		}
 
-		// If the if_exist rule is defined...
 		if (in_array('if_exist', $rules, true))
 		{
+			// If the if_exist rule is defined
 			// and the current field does not exists in the input data
```
MGatner, line 219:
```suggestion
			// and the current field does not exist in the input data
```

MGatner (changes requested):
Handy function! I would like to see a data provider for the test with some more fringe cases, for example an empty string key.

MGatner (approved):

jahmedk:
Thanks it's working :)

PR #4195, opened 2021-01-31
Add additional empty checks after field protection for update/insert.

### LINKED ISSUE #3896: Bug: Premature empty check in Model->update function.
**Describe the bug**
Not sure if I'm overseeing something here...

In the `Model->update()` function several checks for empty data are implemented.
The last one happens here:
https://github.com/codeigniter4/CodeIgniter4/blob/d63020b30977707dac997b8c3505c7636defea11/system/Model.php#L970-L974
Later on, field protection is done:
https://github.com/codeigniter4/CodeIgniter4/blob/d63020b30977707dac997b8c3505c7636defea11/system/Model.php#L984
This can be an issue, if we try to update a database row using an entity (`$model->save($entity)`) and no changes have been submitted. In this case, the entity object will be cast to array here:
https://github.com/codeigniter4/CodeIgniter4/blob/d63020b30977707dac997b8c3505c7636defea11/system/Model.php#L959
Hereby, unchanged fields are removed by the `Entity->toRawArray()` function. However, some fields will remain in the `$data` array (e.g., the primary key column and the 'honeypot').

Therefore, the last empty check is not triggered as `$data` is not empty. After `doProtectFields`, the `$data` array is empty and consequently triggers an error ('You must use the "set" method to update an entry. ').

**CodeIgniter 4 version**
latest develop

**Affected module(s)**
System\Model

see description

**Context**
 - Windows 10
 - Apache 2.4
 - PHP 7.4

sfadschm:
Fixes #3896.

**Description**
This PR adds additional empty data checks to the `Model` `update` and `insert` methods.
These are needed because `doProtectFields` is currently called after the last empty dataset check, but can further remove elements from the `data` array.
See linked issue for details.

Will add tests later today.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Models/UpdateModelTest.php
+++ b/tests/system/Models/UpdateModelTest.php
@@ -260,11 +260,11 @@ final class UpdateModelTest extends LiveModelTestCase
 		$this->createModel(EventModel::class);
 
 		$data = (object) [
-							 'name'    => 'Foo',
-							 'email'   => 'foo@example.com',
-							 'country' => 'US',
-							 'deleted' => 0,
-						 ];
+			'name'    => 'Foo',
+			'email'   => 'foo@example.com',
+			'country' => 'US',
+			'deleted' => 0,
+		];
 
 		$id = $this->model->insert($data);
 
--- a/tests/system/Models/UpdateModelTest.php
+++ b/tests/system/Models/UpdateModelTest.php
@@ -260,11 +260,11 @@ final class UpdateModelTest extends LiveModelTestCase
 		$this->createModel(EventModel::class);
 
 		$data = (object) [
-			'name'    => 'Foo',
-			'email'   => 'foo@example.com',
-			'country' => 'US',
-			'deleted' => 0,
-		];
+							 'name'    => 'Foo',
+							 'email'   => 'foo@example.com',
+							 'country' => 'US',
+							 'deleted' => 0,
+						 ];
 
 		$id = $this->model->insert($data);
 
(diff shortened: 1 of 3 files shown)


michalsn:
Yes, it makes sense to throw an exception here instead of a Builder class.

Waiting for tests, thanks.

sfadschm:
Is there a way that I can do the database live test on my local setup? Like it seems like I only need to add credentials somewhere but I can't seem to find where.

michalsn:
The simplest way would be to add/edit `.env` file and fill the `database.tests.*` part.

sfadschm:
Well yeah, let's just pretend I never asked that....

tests/system/Models/UpdateModelTest.php
```diff
@@ -260,6 +260,26 @@ public function testUpdateObjectWithDataException(): void
 		$this->createModel(EventModel::class);
 
 		$data = (object) [
+							 'name'    => 'Foo',
+							 'email'   => 'foo@example.com',
+							 'country' => 'US',
+							 'deleted' => 0,
+						 ];
```
sfadschm, line 267: @paulbalandan Is this indentation intended? It is appplied by the pre-commit hook with cs-fixer.

paulbalandan: codesniffer is stubbornly wrong on fixing array indentations, especially these types when the array has something before the opening bracket.

I suggest fixing all other things codesniffer fixes, but do not follow on array indentations using `--no-verify` on commit

sfadschm: Alright thanks. I will need to go back to command line committing then, got used to Github Desktop quite too much 😆

michalsn (approved):

paulbalandan (approved):

michalsn:
Thanks, @sfadschm!

PR #4203, opened 2021-02-01
[Rector] Apply Rector: ForToForeachRector

samsonasik:
Apply `ForToForeachRector` rector rule to use foreach instead of for when possible.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -447,9 +447,9 @@ if (! function_exists('safe_mailto'))
 		$output = '<script type="text/javascript">'
 				. 'var l=new Array();';
 
-		foreach ($x as $i => $x)
+		foreach ($x as $i => $value)
 		{
-			$output .= 'l[' . $i . "] = '" . $x . "';";
+			$output .= 'l[' . $i . "] = '" . $value . "';";
 		}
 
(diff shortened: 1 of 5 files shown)


system/Helpers/url_helper.php
```diff
@@ -447,9 +447,9 @@ function safe_mailto(string $email, string $title = '', $attributes = ''): strin
 		$output = '<script type="text/javascript">'
 				. 'var l=new Array();';
 
-		for ($i = 0, $c = count($x); $i < $c; $i ++)
+		foreach ($x as $i => $x)
```
paulbalandan, line 450: Is this not good practice, overwriting the original?

samsonasik: The singularizing process only works in rector if the loop variable is a plural. I can manually change if that needed. Actually, the `$x` still works. Do you have suggestion of new variable of value var ?

paulbalandan: Ok. I always thought this type of assignment will overwrite the original. But maybe we can use another letter for the assignment, to make it less confusing.

samsonasik: I updated to `$value`

paulbalandan (approved):

samsonasik:
Thank you @paulbalandan  for the review.

PR #4222, opened 2021-02-03
Bug #4174 FIX. Exclude CURRENT_TIMESTAMP from being escaped

### LINKED ISSUE #4174: Bug: forge cant set default value = CURRENT_TIMESTAMP
Forge for create table cant set default value `CURRENT_TIMESTAMP`
```
'dt' => [
    'type'    => 'TIMESTAMP',
    'default' => CURRENT_TIMESTAMP
],
```

**CodeIgniter 4 version**
v4.0.4

**Context**
 - OS: Windows 10
 - PHP version 7.4.11

xbotkaj:
Signed-off-by: Jozef Botka <52993808+xbotkaj@users.noreply.github.com>

Fixes #4174 

**Description**
Excluded CURRENT_TIMESTAMP from being escaped. This is only HotFix. It must be considered to rework this part of forge. Default value can contain complex values and cannot be escapes by defualt. It will make breaking change in the future after rework.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -1144,6 +1144,11 @@ class Forge
 				$attributes['NULL'] = true;
 				$field['null']      = empty($this->null) ? '' : ' ' . $this->null;
 			}
+			// CURRENT_TIMESTAMP can not be escaped
+			else if ($attributes['DEFAULT'] === 'CURRENT_TIMESTAMP')
+			{
+				$field['default'] = $this->default . $attributes['DEFAULT'];
+			}
 			else
 			{
 				$field['default'] = $this->default . $this->db->escape($attributes['DEFAULT']);


paulbalandan:
@totoprayogo1916 can you confirm if this fixes your issue?

michalsn:
We would have to support all kinds of these field values. I think the better idea is to pass a string as a field definition: https://codeigniter4.github.io/CodeIgniter4/dbmgmt/forge.html#passing-strings-as-fields

xbotkaj:
> We would have to support all kinds of these field values. I think the better idea is to pass a string as a field definition: https://codeigniter4.github.io/CodeIgniter4/dbmgmt/forge.html#passing-strings-as-fields

Hi @michalsn best solution would be simply not escaping DEFAULT field, but that would make breaking change. Other option would be add field name e.g. DEFAULT_RAW with raw data that would not be escaped. This would solve everyhting without breaking change and keeping forge working withount need what you meantioned, composing the whole column definition by yourself in raw SQL.

michalsn:
Personally, I think that writing raw SQL is the correct way to handle this kind of scenario. Adding `default_raw` has no added value since these types of queries are not transferable between database engines.

But if others think that `default_raw` is a good idea, I won't be against it.

xbotkaj:
> Personally, I think that writing raw SQL is the correct way to handle this kind of scenario. Adding `default_raw` has no added value since these types of queries are not transferable between database engines.
> 
> But if others think that `default_raw` is a good idea, I won't be against it.

I would be preferable for simple RAW but up tou others if we want to have this breaking change, but it is most reasonable way. :)

totoprayogo1916:
> @totoprayogo1916 can you confirm if this fixes your issue?

Sorry, I haven't tried it.
But will there be any more changes @xbotkaj ?

xbotkaj:
> > @totoprayogo1916 can you confirm if this fixes your issue?
> 
> Sorry, I haven't tried it.
> But will there be any more changes @xbotkaj ?

I believe yes, it denpends what exactly will be in favour if sending raw default field or add another field for raw data. Now it is more about if we want to go to breaking change or not.

If you need you can use this hotfix temporary until it will be solved completely for this thing :)

MGatner:
After reading through the (MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html) I agree with @michalsn that this should be handled with existing raw query strings. There are too many aliases that would need supporting and all of it specific to MySQL (with variants for MariaDB and SQLite3). 

Introducing `default_raw` does not seem to me to be worth the duplication when raw query strings can cover all keywords, not just "default".

xbotkaj:
Hi @MGatner then why to use fields at all when we can simply write raw queries all the time? :) No, really that is not an solution, it is workaround. CI support "default" field so it should be able use things like CURRENT_TIMESTAMP (very common) without needs to write whole field in raw query. If you write valid DEFAULT statment it should pass it to DB. Now it is very confusing that you are able use just escaped values. No it is not intuitívne and self-explanatory. 

As per documentation -> default/value : to generate a default value in the field definition. (There is no reason not assume I cannot use CURRENT_TIMESTAMP as my default value or something else.) :)

MGatner:
Well to be clear, I think the original solution could have been better. But we aren't ready to move to version 5.0 yet so we need to work around what's there. Maybe `addField()` could take an additional parameter whether to escape the value? Similar to other database methods?

xbotkaj:
Sure I undrestand, becuase of that I mentioned, this is breaking change if we will make it not escaped as default in the first place. If we make addField with parameter not escaped I am not sure how it would look. since we pass array of fields and it would mean that applies for all. There is another option, to put prefix in string? Something like 
`
addField(
    'field' => ['type' => 'DATETIME', 'default' => '#no_esc#CURRENT_TIMESTAMP']
)
`

This could work until 5.0 where we would make it default no escaping. 

But up to you, there is list of options in this thread :)

kenjis:
Hi, everyone.

How about this?

```php
use CodeIgniter\Database\RawString;

$fields = [
    'id'      => [
        'type'           => 'INT',
        'constraint'     => 5,
        'unsigned'       => true,
        'auto_increment' => true
    ],
    'ts'      => [
        'type'       => 'TIMESTAMP',
        'default'    => new RawString('CURRENT_TIMESTAMP'),
    ],
];
$forge->addField($fields);
```

If we add an additional parameter to `addField()` whether to escape the value,
most use cases only need it for one or few columns.
We have to write `addField()` more than one time for a table.
The code will be a bit difficult to read.

```php
$fields1 = [
    'id'      => [
        'type'           => 'INT',
        'constraint'     => 5,
        'unsigned'       => true,
        'auto_increment' => true
    ],
];
$fields2 = [
    'ts'      => [
        'type'       => 'TIMESTAMP',
        'default'    => 'CURRENT_TIMESTAMP',
    ],
];
$forge->addField($fields1);
$forge->addField($fields2, false);
```

paulbalandan:
Superseded by #5957

PR #4223, opened 2021-02-03
Bug #4035 FIX. Close session write stream when targeting the same server

### LINKED ISSUE #4035: Bug: CURLrequest deadlock on cookie
**Describe the bug**
Using cookies and making CURL request to the same machine makes deadlock. CURL request should not populate Cookie to header (for local development) because it will wait until the first request will finish to unlock cookie and then it will have access to cookie resource. 

Example:
`$client = Services::curlrequest();
$response = $client->request('GET', 'http://localhost');`

This will result in deadlock and waiting for the response. (Sometimes it will make run once and makes deadlock on the second run)

**CodeIgniter 4 version**
4.0.4 develop - latest

**Affected module(s)**
CURLrequest

Processing CURLrequest without deadlock.

**Context**
 - OS: Windows 8.1
 - Apache 2.4
 - PHP version 7.4

P.S. i also tried 'cookie' option like this and with static name of file:
`$tempCode = substr(bin2hex(random_bytes(10)),0,10);
		$curlresponse = $client->request('GET', 'http://localhost/',['cookie' => WRITEPATH . $tempCode . '_CookieSaver.txt']);`

xbotkaj:
Signed-off-by: Jozef Botka <52993808+xbotkaj@users.noreply.github.com>

Fixes #4035 

**Description**
If we use sessions and in the same time making CURL request for the same machine, it will make deadlock. We need close session write stream, execute CURL and resume session.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -797,8 +797,20 @@ class CURLRequest extends Request
 
 		curl_setopt_array($ch, $curlOptions);
 
+		// Stop session to avoid deadlock if target is the same server
+		if(isset($_SESSION) && strstr($curlOptions[CURLOPT_URL], base_url()))
+		{
+			session_write_close();
+		}
+
 		// Send the request and wait for a response.
 		$output = curl_exec($ch);
+
+		// Resume session
+		if(isset($_SESSION) && strstr($curlOptions[CURLOPT_URL], base_url()))
+		{
+			session_start();
+		}
 
 		if ($output === false)
 		{


MGatner:
I'm not sure that I am comfortable with this relying on `base_url()` to determine the session lock. Is there any reason not to release the session on every call, since theoretically a CURLRequest could be a long delay and cause other bottlenecks?

michalsn:
I don't like the idea to have a code that deals with a session at all in this class.

@xbotkaj Please check my answer to your bug report: https://github.com/codeigniter4/CodeIgniter4/issues/4035#issuecomment-773824433

paulbalandan:
Closing based on https://github.com/codeigniter4/CodeIgniter4/issues/4035#issuecomment-774762469

PR #4231, opened 2021-02-05
[Rector] Apply Rector: CombineIfRector

samsonasik:
Merges nested if statements with `CombineIfRector` rector rule.

**Checklist:**
- [x] Securely signed commits


paulbalandan (approved):

MGatner (commented):
Wow that was a surprising number of those.

PR #4232, opened 2021-02-05
no more ( -_ | -_- | _- ) in the result of url_title()

totoprayogo1916:
**Description**
Result of `"url_title()"` has a dash and an underscore even though it is set as underscore, and appears on the other hand as well.
here the example:

```
url_title('make file read-only', '_')
// result: make_file_read-only

// now: make_file_read_only
```
or

```
url_title('underscore_ and dash-')
// result: underscore_-and-dash

// now: underscore-and-dash
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -571,19 +571,27 @@ if (! function_exists('url_title'))
 	 */
 	function url_title(string $str, string $separator = '-', bool $lowercase = false): string
 	{
-		$qSeparator = preg_quote($separator, '#');
+		$qSeparator = preg_quote($separator);
 
 		$trans = [
-			'&.+?;'                  => '',
-			'[^\w\d\pL\pM _-]'       => '',
-			'\s+'                    => $separator,
-			'(' . $qSeparator . ')+' => $separator,
+			'![^' . $qSeparator . '\pL\pM\pN\s]+!u' => '', // remove all chars are not the separator, letters, numbers, whitespace or another character (e.g. accents, umlauts, enclosing boxes, etc.)
+			'![' . $qSeparator . '\s]+!u'        => $separator, // replace all separator chars with $separator
 		];
+
+		// make all dashes/underscores into separator
+		if ($separator === '-')
+		{
+			$str = str_replace('_', '-', $str);
+		}
+		else if ($separator === '_')
+		{
+			$str = str_replace('-', '_', $str);
+		}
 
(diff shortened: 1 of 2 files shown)


MGatner:
I'm not convinced this is a "bug" and thus something we can just change. Are there RFCs or precedents that would guide this? Also any behavior changes in `url_title()` should probably be applied to `mb_url_title()` as well

michalsn:
This function works as it supposed to work. If you want to replace the `-` or `_` occurrences, you have to use `str_replace()` separately.

This would be a big breaking change.

totoprayogo1916:
updated,
any suggestion @michalsn @MGatner ?

michalsn:
My point was, that you can create your own function which will use `url_title()` and `str_replace()`.

`url_title()` is good in its current form - I see no reason to change it. The proposed changes will affect the current behavior, which is not buggy.

kenjis:
The `url_title()` in this PR seems good to me.
But it is BC break.

I don't know CI4's BC policy.

paulbalandan:
I think the confusion in `url_title`'s work had arisen from its phpdoc. It currently says that this has the separator string between word characters, which is what this PR tries to achieve. On the other hand, our code tells otherwise. It replaces spaces and duplicated instances of the separator with a single separator and removes forbidden chars. There is no telling that this "normalizes" separators between words. The phpdoc and code should be consistent with each other.

For the PR, I think we should not restrict the separator replacements to only `_` and `-` but can be any other as long as it is not forbidden in the URL.

michalsn:
I'm okay with adding custom separator choices, but I'm strongly against changes in the behavior.

The thing is that people use this function a lot without saving the result to the database. I saw many use cases when this function wasn't used only to generate URL slug but also to remove whitespaces from the file names before saving them and then linking to these dynamically created files. Another reason is compatibility with CI3, which is also important.

element-code:
> Another reason is compatibility with CI3, which is also important.

Im strongly against this view. We should take CI4 as a chanche to decouple from the slow CI3 development which was one of it's biggest disadvantages when compared to other frameworks!

michalsn:
Backward compatibility for some functions is very important and IMO this is one of these cases. I'm not against adding a new helper function though (like `slug()` or something) but it would have to be a "complete" solution with solid support for all languages.

PR #4239, opened 2021-02-07
Fix HealthTest::testBaseUrlHasBeenSet()

### LINKED ISSUE #3977: Bug: appstarter HealthTest::testBaseUrlHasBeenSet fails
**Describe the bug**
HealthTest::testBaseUrlHasBeenSet fails.

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
tests

~~All tests should pass.~~
HealthTest is supposed to check for various "common mistakes" in a user's app, including if App::$baseURL is set.
So fails only when a user does not set baseURL.

```
$ composer create-project codeigniter4/appstarter ci4
$ cd ci4/
$ composer test
...
There was 1 failure:

1) HealthTest::testBaseUrlHasBeenSet
Failed asserting that false is true.

/Users/kenji/tmp/ci4/tests/unit/HealthTest.php:31

FAILURES!
Tests: 5, Assertions: 6, Failures: 1.
Script phpunit handling the test event returned with error code 1
```

**Context**
 - OS: [macOS 10.15.7]
 - Web server [n/a]
 - PHP version [7.4.13]

kenjis:
**Description**
Fixes #3977

Tests about `baseURL`:
- `baseURL` is a valid URL
- Set it in `.env` or `app/Config/App.php`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/admin/starter/tests/unit/HealthTest.php
+++ b/admin/starter/tests/unit/HealthTest.php
@@ -21,7 +21,7 @@ class HealthTest extends \CodeIgniter\Test\CIUnitTestCase
 		// First check in .env
 		if (is_file(HOMEPATH . '.env'))
 		{
-			$env = (bool) preg_grep("/^app\.baseURL = './", file(HOMEPATH . '.env'));
+			$env = (bool) preg_grep('/^app\.baseURL = ./', file(HOMEPATH . '.env'));
 		}
 
 		// Then check the actual config file
--- a/admin/starter/tests/unit/HealthTest.php
+++ b/admin/starter/tests/unit/HealthTest.php
@@ -1,5 +1,7 @@
 <?php
 
+use Config\Services;
+
 class HealthTest extends \CodeIgniter\Test\CIUnitTestCase
 {
 	public function setUp(): void
@@ -16,18 +18,35 @@ class HealthTest extends \CodeIgniter\Test\CIUnitTestCase
 
 	public function testBaseUrlHasBeenSet()
 	{
-		$env = $config = false;
+		$validation = Services::validation();
+		$env = false;
 
-		// First check in .env
+		// Check the baseURL in .env
 		if (is_file(HOMEPATH . '.env'))
 		{
 			$env = (bool) preg_grep('/^app\.baseURL = ./', file(HOMEPATH . '.env'));
 		}
 
-		// Then check the actual config file
-		$reader = new \Tests\Support\Libraries\ConfigReader();
-		$config = ! empty($reader->baseURL);
+		if ($env)
+		{
+			// BaseURL in .env is a valid URL?
+			// phpunit.xml.dist sets app.baseURL in $_SERVER
+			// So if you set app.baseURL in .env, it takes precedence
+			$config = new Config\App();
+			$this->assertTrue(
+				$validation->check($config->baseURL, 'valid_url'),
+				'baseURL "' . $config->baseURL . '" in .env is not valid URL'
+			);
+		}
 
-		$this->assertTrue($env || $config);
+		// Get the baseURL in app/Config/App.php
+		// Don't forget phpunit.xml.dist sets app.baseURL
+		$reader = new \Tests\Support\Libraries\ConfigReader();
+
+		// BaseURL in app/Config/App.php is a valid URL?
+		$this->assertTrue(
+			$validation->check($reader->baseURL, 'valid_url'),
+			'baseURL "' . $reader->baseURL . '" in app/Config/App.php is not valid URL'
+		);
 	}
 }
--- a/admin/starter/tests/unit/HealthTest.php
+++ b/admin/starter/tests/unit/HealthTest.php
@@ -19,7 +19,7 @@ class HealthTest extends \CodeIgniter\Test\CIUnitTestCase
 	public function testBaseUrlHasBeenSet()
 	{
 		$validation = Services::validation();
-		$env = false;
+		$env        = false;
 
 		// Check the baseURL in .env
 		if (is_file(HOMEPATH . '.env'))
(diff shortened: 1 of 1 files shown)


kenjis:
```sh-session
$ composer create-project codeigniter4/appstarter ci4
$ cd ci4/
$ vendor/bin/phpunit --no-coverage
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

....F                                                               5 / 5 (100%)

Time: 00:00.085, Memory: 12.00 MB

There was 1 failure:

1) HealthTest::testBaseUrlHasBeenSet
baseURL in app/Config/App.php is not set
Failed asserting that false is true.

/Users/kenji/tmp/ci4/tests/unit/HealthTest.php:38

FAILURES!
Tests: 5, Assertions: 7, Failures: 1.
```

kenjis:
```diff
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -23,7 +23,7 @@ class App extends BaseConfig
     *
     * @var string
     */
-   public $baseURL = 'http://localhost:8080/';
+   public $baseURL = 'http://abc.example.com/';
 
    /**
     * --------------------------------------------------------------------------

```

```sh-session
$ vendor/bin/phpunit --no-coverage
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

.....                                                               5 / 5 (100%)

Time: 00:00.080, Memory: 12.00 MB

OK (5 tests, 8 assertions)
```

kenjis:
```sh-session
$ git reset --hard HEAD
$ cp env .env
$ vi .env
```

```diff
--- env	2021-02-01 11:56:38.000000000 +0900
+++ .env	2021-02-07 11:09:59.000000000 +0900
@@ -20,7 +20,7 @@
 # APP
 #--------------------------------------------------------------------
 
-# app.baseURL = ''
+app.baseURL = http://xyz.example.com/
 # app.forceGlobalSecureRequests = false
 
 # app.sessionDriver = 'CodeIgniter\Session\Handlers\FileHandler'
```

```sh-session
$ vendor/bin/phpunit --no-coverage
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

.....                                                               5 / 5 (100%)

Time: 00:00.078, Memory: 12.00 MB

OK (5 tests, 8 assertions)
```

kenjis:
I think this is probably OK. But a bit complicated...

MGatner:
Per the issue conversation this should be updated only to fix the regex. But please make sure the regex allows single quotes still, the version in this PR does not.

kenjis:
URL with single quotes passes the test. `'` is included in `.`.

```diff
--- env	2021-02-01 11:56:38.000000000 +0900
+++ .env	2021-02-10 10:57:00.000000000 +0900
@@ -20,7 +20,7 @@
 # APP
 #--------------------------------------------------------------------
 
-# app.baseURL = ''
+app.baseURL = 'http://localhost:8080/'
 # app.forceGlobalSecureRequests = false
 
 # app.sessionDriver = 'CodeIgniter\Session\Handlers\FileHandler'
```

```sh-session
$ vendor/bin/phpunit --no-coverage
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

.....                                                               5 / 5 (100%)

Time: 00:00.075, Memory: 12.00 MB

OK (5 tests, 8 assertions)
```

paulbalandan:
why are we disallowing `localhost:8080` for the baseURL?

kenjis:
The user guide says:
> Open the app/Config/App.php file with a text editor and set your base URL. If you need more flexibility, the baseURL may be set within the .env file as app.baseURL=”http://example.com/”. (Always use a trailing slash on your base URL!)
https://codeigniter4.github.io/CodeIgniter4/installation/running.html#initial-configuration-set-up

So I think users are supposed to set their base URL, and `localhost:8080` is the default, they did not set it.

MGatner:
Localhost is fine for a base URL! See my comment above, but the issue is when a user has explicitly created a scenario where the base URL is blank - out of the box it should work fine.

MGatner:
I think the regex currently fails to catch `app.baseUrl = ''` which should fail.

kenjis:
How about this? I've cleaned up.

I removed disallowing `localhost:8080`, and changed the validation to use Validator.

`app.baseURL = ''` is not a valid URL, so it fails.

admin/starter/tests/unit/HealthTest.php
```diff
@@ -16,18 +18,35 @@ public function testIsDefinedAppPath()
...
+			$config = new Config\App();
+			$this->assertTrue(
+				$validation->check($config->baseURL, 'valid_url'),
+				'baseURL "' . $config->baseURL . '" in .env is not valid URL'
+			);
 		}
 
-		// Then check the actual config file
+		// Get the baseURL in app/Config/App.php
+		// Don't forget phpunit.xml.dist sets app.baseURL
```
paulbalandan, line 43: I think this is not needed. As `ConfigReader` does not call `Config\App`'s constructor, its properties would not be populated by environment values but only by the hardcoded values.

kenjis: @paulbalandan I fixed the comment. My intention is why we need to use ConfigReader.
I forgot about phpunit.xml.dist at the first time.

paulbalandan (approved):

bereldhuin:
I noticed another problem with the regexp (using CI 4.1.1) : VisualStudio code tends to remove spaces in .env files.

So my baseUrl looks like this : 
`app.baseURL='http://192.168.1.11:81/'`
Instead of
`app.baseURL = 'http://192.168.1.11:81/'`

Which makes the phpunit fail on testBaseUrlHasBeenSet

PR #4241, opened 2021-02-07
Develop sqlsrv schema

hdrodriguez:
**Description**

In codeigniter 3 using sqlsrv handler we could use a different scheme than "dbo" for example "adm", "tools", the full name being configured as follows: [DBNAME]. [SCHEMA]. [TABLENAME] however in the version 4.1 to obtain a record the full name looks like this: [DBNAME]. [TABLENAME] and to insert or update then if you take [DBNAME]. [SCHEMA]. [TABLENAME] from the key database configuration file " schema ", then it occurs to me that the table name" schema.tableName "is used in the model for sqlsrv. If the schema is not set then it takes "dbo" by default.

`private function getFullName(string $table): string
	{
		$tableExploded = explode('.', $table);
		if (count($tableExploded) === 2)
		{
			$this->db->schema = str_replace('"', '', $tableExploded[0]);
			$tableName        = $tableExploded[1];
		}
		else
		{
			$tableName = $table;
		}

		if ($this->db->escapeChar === '"')
		{
			return '"' . $this->db->getDatabase() . '"."' . $this->db->schema . '"."' . str_replace('"', '', $tableName) . '"';
		}
		return '[' . $this->db->getDatabase() . '].[' . $this->db->schema . '].[' . str_replace('"', '', $tableName) . ']';
	}
`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/SQLSRV/Builder.php
+++ b/system/Database/SQLSRV/Builder.php
@@ -95,6 +95,8 @@ class Builder extends BaseBuilder
 	 */
 	protected function _insert(string $table, array $keys, array $unescapedKeys): string
 	{
+
+
 		$fullTableName = $this->getFullName($table);
 
 		// insert statement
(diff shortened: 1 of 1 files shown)


MGatner:
@hdrodriguez this is over my MSSQL head but your description makes me think it would be worthwhile. Could you add a test case demonstrating the difference?

michalsn:
Thanks, @hdrodriguez this is a bug indeed. But as @MGatner said we have to fix it "globally". Please check this PR #4246.

I would appreciate your feedback.

hdrodriguez:
Thank you very much for your comments @michalsn and @MGatner  I think that PR 4246 covers more things and takes into account the use of aliases, which is very important when working with relationships, however the change that I have proposed only affected getFullName and focused in the passage of a schema through the model >>> protected $ table = "schema.table", at PR 4246 could we include what was my code trying to achieve?

I could add my tests but they would conflict with PR 4246.

should I close this PR?

michalsn:
This PR will be automatically closed after solving the issue.

PR #4264, opened 2021-02-11
Ability to create class instances inside entity.

nowackipawel:
With this you could create in your entities inner instances of other entities.

I.e.:

you have entity that you wish to contain other entity in it.... lets say you need POSCommon entity in your RequestRate class...

```
<?php

namespace App\Entities\API;

use App\Libraries\Ext\Entity;

class RequestRate extends Entity
{
	protected $casts = [
		'commission_buy'						=> '?decimal',
		'commission_sell'						=> '?decimal',
		'pos_common'							=> 'class:App\Entities\API\POSCommon'
	];


	protected $attributes = [
		'commission_buy'						=> null,
		'commission_sell'						=> null,
		'pos_common'							=> null,
	];

	protected $datamap = [
		'commissionBuy' => 'commission_buy',
		'commissionSell' => 'commission_sell',
		'r' => 'pos_common'
	];

}

```

with that change it is peace of cake ;-)

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -368,9 +368,18 @@ class Entity implements JsonSerializable
 			// Array casting requires that we serialize the value
 			// when setting it so that it can easily be stored
 			// back to the database.
-			if ($castTo === 'array' || strpos($castTo, 'class:') === 0)
+			if ($castTo === 'array')
 			{
 				$value = serialize($value);
+			}
+
+			if (strpos($castTo, 'class:') === 0)
+			{
+				$type = substr($castTo,6);
+
+				if(class_exists($type)) {
+					$value = serialize(new $type($value));
+				}
 			}
 
 			// JSON casting requires that we JSONize the value
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -368,7 +368,7 @@ class Entity implements JsonSerializable
 			// Array casting requires that we serialize the value
 			// when setting it so that it can easily be stored
 			// back to the database.
-			if ($castTo === 'array')
+			if ($castTo === 'array' || strpos($castTo, 'class:') === 0)
 			{
 				$value = serialize($value);
 			}
@@ -542,6 +542,17 @@ class Entity implements JsonSerializable
 				return null;
 			}
 			$type = substr($type, 1);
+		}
+		
+		
+		if(strpos($type, 'class:') === 0)
+		{
+			$type = substr($type,6);
+
+			if(class_exists($type)) {
+				$value = is_array($value) ? new $type($value) : unserialize($value);
+			}
+			return $value;
 		}
 
 		switch($type)


nowackipawel:
```
		$post = new RequestRate($this->request->getJSON(true));

		var_dump($post->pos_common);
```

Screenshot from 2021-02-11 15-36-00

MGatner:
I've been using something similar with my NoSQL database, since entities can actually have other entities "inside" them. I think a cast as a class would be a beneficial feature but we would need tests, User Guide, changelog, etc.

MGatner:
Mentioning @najdanovicivan here too.

najdanovicivan:
Regarding the situation with this we have multiple things to consider here 

1. The way data is stored in the db as we have 3 ways to do that: 
             Serialized String 
             Json String
             Raw Array

All of those can also represent array of class objects. For example with one to many relations

2. Second is the kind of class. 

- If the class is instance of Entity is should be initiated as new entity object and data should be passed set using setAttributes method on the Entity Object and the value should be revert back to array for writing using toRawArray() method. This Affect all the data formats serialized, json, raw. We first convert it back to regular array then we initialize the Entity object from such array.

- If data is stored in json format we first need to check if class is instance of JsonSerializable and then decode object directly from json.

- Only in the end we can call unserialize. But at this point we need to make sure that result is instance of the needed class and if not trow and exception for this.

3. Data writing - We need to make sure that what type of data we get from attributes we write back the same time of data for the DB update


So I'll propose the format for those casts:

```
class-serialized:App\Entities\Entity - Single Class Object from serialized value
class-json:App\Entities\Entity - Single Class Object from JSON encoded value
class-raw:App\Entities\Entity - Single Class Object from raw array value

class-serialized:App\Entities\Entity[] - Array of Class Object from serialized value
class-json:App\Entities\Entity[] - Array of  Class Object from JSON encoded value
class-raw:App\Entities\Entity[] - Array of  Class Object from raw array value
```

@MGatner, @paulbalandan Feel free to add your thought and opinions

nowackipawel:
> ```
> class-serialized:App\Entities\Entity - Single Class Object from serialized value
> class-json:App\Entities\Entity - Single Class Object from JSON encoded value
> class-raw:App\Entities\Entity - Single Class Object from raw array value
> 
> class-serialized:App\Entities\Entity[] - Array of Class Object from serialized value
> class-json:App\Entities\Entity[] - Array of  Class Object from JSON encoded value
> class-raw:App\Entities\Entity[] - Array of  Class Object from raw array value
> ```
> 

In my Entity class I am using also php-decimal.io casting and I found pretty clear that kinda format:
```
decimal:4,2   (for defining precision)
```
in this case I would propose:
```
class:App\Entities\Entity,serialized
class:App\Entities\Entity,json
class:App\Entities\Entity,raw

and for array:

class:App\Entities\Entity,serialized,[]
class:App\Entities\Entity,json,[]
class:App\Entities\Entity,raw,[]
or
class-array:App\Entities\Entity,serialized
class-array:App\Entities\Entity,json
class-array:App\Entities\Entity,raw
```
I understand your format is related to the way how to define arrays in some languages, but requires more parsing.

najdanovicivan:
Arrays in PHPdoc are defined as App\Entities\Entity[] so it makes sense to use [] suffix then we can easily check if it has to be array by doing 

```
$cast = explode(':', $cast);
$class = $cast[1]
$cast = $cast[0];
$array = substr($class,-2) === '[]';
$class =  rtrim($cast,'[]');
```

nowackipawel:
> Arrays in PHPdoc are defined as App\Entities\Entity[] so it makes sense to use [] suffix then we can easily check if it has to be array by doing
> 
> ```
> $cast = explode(':', $cast);
> $class = $cast[1]
> $cast = $cast[0];
> $array = substr($class,-2) === '[]';
> $class =  rtrim($cast,'[]');
> ```
> 
> Also note that $cast should accept nullable prefix ?

It looks good to me. 
btw/ nullability is already supported - that feature is important for me - the best proof for that is I was the one who introduced it :D


It would be great if someone will declare that he/she/they will take care about documentation and tests and format proposed by you is acceptable then I could continue to work on it.

najdanovicivan:
Regarding the format we should wait to hear from any of the project members

@lonnieezell  @MGatner @samsonasik @michalsn  @paulbalandan

lonnieezell:
Does this support entities within entities? How large does that get? Maybe it would make sense to do something more robust that only stores the attribute values and the primary key and serializes that so it's not serializing the entire entity? That uses less db space, and I think would bypass some potential edge case issues we are bound to run into. 

Additionally - unless you can find a "sponsor" willing to write tests and docs, please take the time to do it for your PR's. The handful of volunteers that work on this framework are also very busy, just like you, and we don't have time to finish everyone's PR's. You'll have a better time getting something accepted if you can flesh the PR out yourself.

najdanovicivan:
@lonnieezell Entity objects are not serialized php objects. It serialized array obtained from toRawArray() method and initialization passed those data to new entity via setAttributes method. This is same way as db model does it

lonnieezell:
You're right. My bad. 

As I'm thinking about this more I'm wondering what the practical use is? Basically this seems to be a way to store a relationship, but that relationship doesn't get updated when the underlying entity does. And if it's a static class it seems a factory that could provide an updated version might make more sense so things don't have to get updated in all entities. It seems @MGatner sees a use for it that I'm not, though. So maybe I'll just stay out of this one lol. 

Either way, someone would need to finish out the PR for it to have a chance of being accepted.

najdanovicivan:
@lonnieezell It does not make much sense when SQL is being used but it makes total sense with NoSQL where objects can be stored inside of objects.

It also make sense for smaller objects to be stored as serialized or json values inside sql table

I’m often storing configs as json in sql table as I dont need those to be available for querying

lonnieezell:
> @lonnieezell It does not make much sense when SQL is being used but it makes total sense with NoSQL where objects can be stored inside of objects.
> 
> It also make sense for smaller objects to be stored as serialized or json values inside sql table
> 
> I’m often storing configs as json in sql table as I dont need those to be available for querying

That's fair.

MGatner:
@lonnieezell The use case where I've actually thought about writing something like this myself is with Firestore. I'm not sure how other "NoSQL" databases work but you can't do joins in Firestore so you have nested "tables" and "row references". So in this case if I had `users` table it could have a `widgets` table as a child of one user, and that `widget` could have a reference to a top-level `colors` row. In this instance I might want to to get a `UserEntity` and if I say `$user->getWidgets()` that might return a `WidgetModel`. Then a `WidgetEntity` might have `$widget->getColor()` which would fetch and unpack a `ColorEntity`. Just as an example:

```
$user = model(UserModel::class)->find(1);
foreach ($user->widgets->findAll() as $widget)
{
	echo $widget->color->asHtml();
}
```

lonnieezell:
That makes sense. I've only worked with NoSQL on one project and that was at least 7 years ago so my mind is definitely not there currently :) Thinking back and looking at your description I can definitely see how it could be handy. This project was an automotive aggregator and this could have cleaned up the code a bit.

najdanovicivan:
I'd also like to have casting in load time. So that setAttrubites method will store already processed values. As there's an issue with for example ElasticSearch where all bool values will be returned as true/false strings. But when trying to save those back in the module it will throw an exception as string values are not accepted. But that is completely different thing for which a completely separate PR should be made

paulbalandan:
I think this can be closed in favor of #4308 ?

najdanovicivan:
@paulbalandan #4308  won't solve the issue. As getter for the cast will always return the new entity. So if you do any modification to the inner entity it won't be reflected back to the main entity event worse all the changes will be lost when the getter is called for the second time.

I'have an idea of extending the casts bit further. I was waiting on #4308 to get merged first. Now I can put up another PR which will make this possible. I already mentioned something on the PR there about adding serialize and unserialize methods

lonnieezell:
This hasn't been touched in 9 months. Additionally, the reasons it was given as being a good thing is for NoSQL databases, which we currently do not directly support. Instead of adding features that are handy for a feature that doesn't exist yet, I think that should be added when/after the other feature is implemented. Seems like someone can fairly easily handle this within their entity should they need it. 

Closing.

PR #4266, opened 2021-02-12
New assertRedirectedTo assertion for HTTP testing.

lonnieezell:
Adds a new assertion for HTTP testing: `assertRedirectedTo()` that combines `assertRedirect()`while comparing the passed in URI with the `getRedirectURL()` value.


system/Test/FeatureResponse.php
```diff
@@ -98,6 +98,27 @@ public function assertRedirect()
 		$this->assertTrue($this->isRedirect(), 'Response is not a redirect or RedirectResponse.');
 	}
 
+	/**
+	 * Assert that a given response was a redirect
+	 * and it was redirect to a specific URI.
+	 *
+	 * @param string $uri
+	 */
+	public function assertRedirectedTo(string $uri)
```
MGatner, line 107: > and it was redirect

> assertRedirectedTo

I think we should choose a case and stick with it. I would prefer "redirect" to be consistent with the other methods (so `assertRedirectTo()`) but as long as it is consistent and clear.

system/Test/FeatureResponse.php
```diff
@@ -98,6 +98,27 @@ public function assertRedirect()
...
+	{
+		if (! $this->isRedirect())
+		{
+			return false;
+		}
+
+		$uri         = trim(strtolower($uri));
```
MGatner, line 114: Why lower case? I don't think the return from `getRedirectUrl()` will necessarily be lower.

tests/system/Test/FeatureResponseTest.php
```diff
@@ -151,6 +151,15 @@ public function testGetRedirectUrlReturnsNull()
...
 
+	public function testRedirectedToSuccess()
+	{
+		$this->getFeatureResponse('<h1>Hello World</h1>');
+		$this->feature->response = new RedirectResponse(new App());
+		$this->feature->response->redirect('foo/bar');
+
+		$this->feature->assertRedirectedTo('foo/bar');
+	}
+
```
MGatner, line 162: I think a test using a full URL is in order since your method supports it.

MGatner (changes requested):
This is a great addition! I basically end up doing this myself in almost every project 🙄
I added some thoughts.

system/Test/FeatureResponse.php
lonnieezell: might be true. I know headers are case-insensitive, which obviously isn't the same. Will turn both to lowercase then.

lonnieezell:
Changes made.

user_guide_src/source/testing/feature.rst
```diff
@@ -198,6 +198,14 @@ Asserts that the Response is an instance of RedirectResponse.
 
     $result->assertRedirect();
 
+**assertRedirectedTo()**
```
paulbalandan, line 201: docs is not yet updated

kenjis: @lonnieezell Could you update and fix the conflict in changelog?

lonnieezell:
@MGatner changes ready for re-review

MGatner (changes requested):

lonnieezell:
@MGatner updated

MGatner:
I think this is ready but needs a rebase after the recent Rector fixes.

MGatner (approved):

MGatner:
Glad to see this make it in! Precisely the area I am working on.

PR #4270, opened 2021-02-12
Store entity property as GZIPPED binary strings

nowackipawel:
With that you could store all of data types in database (BLOB) type as gzipped binnary strings.
It could save a lot of data in case of longer strings / logs. Tested with string but it should work perfectly with other types as well as nullable. 

uncompressed: 2267 
compressed with gzip: 1079

gzip is fast and should work everywhere.

If you are interested guys with it , it would be great to write docs and tests ;-)

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -348,6 +348,7 @@ class Entity implements JsonSerializable
 			$value = $this->mutateDate($value);
 		}
 
+		$castTo = false;
 		$isNullable = false;
 		$isGzipped = false;
 
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -338,7 +338,7 @@ class Entity implements JsonSerializable
 	 * @return $this
 	 * @throws Exception
 	 */
-	public function __set(string $key, $value = null)
+	public function __set(string $key, $value = null, $castTo = false)
 	{
 		$key = $this->mapProperty($key);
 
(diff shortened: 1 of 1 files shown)


MGatner:
Seems like a cool idea, but we would definitely needs tests and User Guide docs and adherence to the Style Guide. I know everyone wants to write the code and not the "boring stuff" but the maintainers here are too strapped to field partial PRs.

Feel free to re-open if you would like to complete the PR yourself.

nowackipawel:
I am deep in project that require from me a lot of coding in Java and PHP right now.  I am just sharing what I wrote and  found possibly interesting for others. I do not think writing tests/docs are boring stuff I just do not have time for this lately.
Because some of my ideas were rejected (because of lacks of docs/test) for now I have rewritten pretty much parts of model and entity classes that are working with current CI codebase  so it is not as easy to me to jump into this code to integrate it :/.

I will try to write some docs test in the future but I dunno when. Many my PR are already marked as abandoned.

MGatner:
I understand, and you write good stuff but this isn't a mild preference: it is absolutely necessary for the foundation of the framework. If you write useful stuff that you would like to share but don't have the time to follow PR guidelines I would suggest publishing them as a separate library (this is what I do). 

Your contributions are appreciated here but the maintainers are just as busy and get no compensation for doing this so we really cannot be expected to complete PRs. Any time you want to come back and complete a PR we will gladly review.

PR #4272, opened 2021-02-12
Fabricator: Update model errors on implode

natanfelles:
**Description**
For a strange reason the method `$this->model->errors()` returns a string.

BEFORE (implode error):
```
/home/natanfelles/CodeIgniter/se/vendor/codeigniter4/framework/system/Test/Fabricator.php:553:
string(52) "Duplicate entry '16-14' for key 'product_id_user_id'"
An uncaught Exception was encountered

Type:        ErrorException
Message:     implode(): Invalid arguments passed
Filename:    /home/natanfelles/CodeIgniter/se/vendor/codeigniter4/framework/system/Test/Fabricator.php
Line Number: 554
```
Note the error in the `implode` function.

AFTER (implode without error):
```
/home/natanfelles/CodeIgniter/se/vendor/codeigniter4/framework/system/Test/Fabricator.php:553:
string(48) "Duplicate entry '44-10' for key 'post_id_tag_id'"
An uncaught Exception was encountered

Type:        CodeIgniter\Exceptions\FrameworkException
Message:     Fabricator falhou ao inserir na tabela blog_posts_with_tags: Duplicate entry '44-10' for key 'post_id_tag_id'.
Filename:    /home/natanfelles/CodeIgniter/se/vendor/codeigniter4/framework/system/Test/Fabricator.php
Line Number: 554
```

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

  
- [ ] User guide updated

### DIFF
--- a/system/Test/Fabricator.php
+++ b/system/Test/Fabricator.php
@@ -43,7 +43,7 @@ class Fabricator
 	/**
 	 * Model instance (can be non-framework if it follows framework design)
 	 *
-	 * @var CodeIgniter\Model|object
+	 * @var \CodeIgniter\Model|object
 	 */
 	protected $model;
 
@@ -549,7 +549,7 @@ class Fabricator
 				continue;
 			}
 
-			throw FrameworkException::forFabricatorCreateFailed($this->model->table, implode(' ', $this->model->errors() ?? []));
+			throw FrameworkException::forFabricatorCreateFailed($this->model->table, implode(' ', (array)$this->model->errors()));
 		}
 
 		// If the model defines a "withDeleted" method for handling soft deletes then use it


MGatner:
If `BaseModel::errors()` is returning a string then there is something much more seriously wrong than the Fabricator. This method can only call `Validation::getErrors(): array` (https://github.com/codeigniter4/CodeIgniter4/blob/b1d4ffce3693b080b21489a2fc0b0b43324317d4/system/Validation/Validation.php#L743) or `BaseConnection::error()` (https://github.com/codeigniter4/CodeIgniter4/blob/b1d4ffce3693b080b21489a2fc0b0b43324317d4/system/Database/BaseConnection.php#L1760).

I think the problem might be with the latter, that we return only a portion of the database error: 
```
return $error['message'] ?? null;
```
even though `BaseModel::errors()` by definition must return an `array`.

natanfelles:
Yes. The issue is in the `$error['message']`.

Giving a var_dump in the `$error` variable we get the  `$error['message']` as a string (should be array):

```
/home/natanfelles/CodeIgniter/se/vendor/codeigniter4/framework/system/BaseModel.php:1109:
array(2) {
  'code' =>
  int(1062)
  'message' =>
  string(50) "Duplicate entry '44-2' for key 'product_id_tag_id'"
}

```

natanfelles:
What should `BaseModel::errors()` return since `$error['message']` is a string?

https://github.com/codeigniter4/CodeIgniter4/blob/ae833c1b21cae7ec65ba1b96a6588cd7b6c5b504/system/BaseModel.php#L1095-L1111

1) `0 => 'message'`

or

2) `'message' => 'message'`

or

3) `'code' => 'code',
'message' => 'message'`

or

4) other?

MGatner:
Let's be consistent with `Validation::getErrors()` which returns `array<string,string>` with the key as the "field" and value as the message. So maybe `['database' => $error['message']]` - how does that sound?

MGatner:
Actually, let's think bigger. I think it is a mistake for `BaseModel::errors()` to assume an array format from `Model::doErrors()` - that defeats the whole point of splitting these methods out. Also, since this has been a bug forever I would like to consider reworking this because returning `array|null` is completely unhelpful. I will make a PR

natanfelles:
Fixed. Thank you @MGatner !

PR #4276, opened 2021-02-13
Use verbose DB output for errors

paulbalandan:
**Description**
This PR changes the error output for DB connections to be more verbose than the generic `Unable to connect to the database.`. This is an attempt to reduce questions on the forum regarding this despite raising the logging threshold to `4`.

`php spark migrate --all`

image

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/app/Views/errors/html/error_404.php
+++ b/app/Views/errors/html/error_404.php
@@ -74,7 +74,7 @@
 
 		<p>
 			<?php if (! empty($message) && $message !== '(null)') : ?>
-				<?= esc($message) ?>
+				<?= esc(nl2br($message)) ?>
 			<?php else : ?>
 				Sorry! Cannot seem to find the page you were looking for.
 			<?php endif ?>
--- a/app/Views/errors/html/error_exception.php
+++ b/app/Views/errors/html/error_exception.php
@@ -21,7 +21,7 @@
 		<div class="container">
 			<h1><?= esc($title), esc($exception->getCode() ? ' #' . $exception->getCode() : '') ?></h1>
 			<p>
-				<?= esc($exception->getMessage()) ?>
+				<?= esc(nl2br($exception->getMessage())) ?>
 				<a href="https://www.duckduckgo.com/?q=<?= urlencode($title . ' ' . preg_replace('#\'.*\'|".*"#Us', '', $exception->getMessage())) ?>"
 				   rel="noreferrer" target="_blank">search &rarr;</a>
 			</p>
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -342,6 +342,7 @@ abstract class BaseConnection implements ConnectionInterface
 		//--------------------------------------------------------------------
 
 		$this->connectTime = microtime(true);
+		$connectionErrors  = [];
 
 		try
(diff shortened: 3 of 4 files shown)


system/Database/BaseConnection.php
```diff
@@ -392,7 +395,13 @@ public function initialize()
 			// We still don't have a connection?
 			if (! $this->connID)
 			{
-				throw new DatabaseException('Unable to connect to the database.');
+				$nl2br = is_cli() ? "\n" : '<br/>';
```
samsonasik, line 398: `\n` can be PHP_EOL constant?

system/Database/BaseConnection.php
```diff
@@ -392,7 +395,13 @@ public function initialize()
 			// We still don't have a connection?
 			if (! $this->connID)
 			{
-				throw new DatabaseException('Unable to connect to the database.');
+				$nl2br = is_cli() ? PHP_EOL : '<br/>';
```
MGatner, line 398: I don't like this class caring about CLI versus web. I understand this is an exception in the making but can we find another way to do this?

MGatner (changes requested):

system/Database/BaseConnection.php
lonnieezell: Exceptions are already caught and have templates specific to html or cli, right? So shouldn't be a need for it here unless I'm missing something. Seems a better choice is to use line breaks here, and make sure the exception templates deal with it correctly. That's the proper place for this, I believe. Otherwise, it's just a band-aid on a different problem and we try to keep band-aids away from the core framework. :)

app/Views/errors/html/error_404.php
```diff
@@ -74,7 +74,7 @@
 
 		<p>
 			<?php if (! empty($message) && $message !== '(null)') : ?>
-				<?= esc($message) ?>
+				<?= esc(str_replace(PHP_EOL, '<br/>', $message)) ?>
```
MGatner, line 77: Why not [nl2br](https://www.php.net/manual/en/function.nl2br.php)?

app/Views/errors/html/error_exception.php
```diff
@@ -21,7 +21,7 @@
 		<div class="container">
 			<h1><?= esc($title), esc($exception->getCode() ? ' #' . $exception->getCode() : '') ?></h1>
 			<p>
-				<?= esc($exception->getMessage()) ?>
+				<?= esc(str_replace(PHP_EOL, '<br/>', $exception->getMessage())) ?>
```
MGatner, line 24: `nl2br()`

MGatner (changes requested):
One question, but overall this looks like a vast improvement. I like the differentiation between primary and failover errors, nice touch!

app/Views/errors/html/error_404.php
paulbalandan: Silly me. I forgot that. Sorry for reinventing the wheel. 🤣

paulbalandan: Fixed

MGatner (approved):
Looks good pending tests.

PR #4280, opened 2021-02-13
[DocBlock] Add property $request with @var IncomingRequest in BaseController

samsonasik:
Controller that extends `BaseController` mostly on http, I think it will be better for IDE detection with mark as `@var IncomingRequest` for its `$request` property.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -3,6 +3,7 @@
 namespace App\Controllers;
 
 use CodeIgniter\Controller;
+use CodeIgniter\HTTP\IncomingRequest;
 use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
 use Psr\Log\LoggerInterface;
@@ -20,6 +21,13 @@ use Psr\Log\LoggerInterface;
 
 class BaseController extends Controller
 {
+	/**
+	 * Instance of the main Request object.
+	 *
+	 * @var IncomingRequest
+	 */
+	protected $request;
+
 	/**
 	 * An array of helpers to be loaded automatically upon
 	 * class instantiation. These helpers will be available


paulbalandan (approved):
Agreed. My IDE turns red on incomplete interface reliance.

samsonasik:
@paulbalandan the another solution is probably add `&` instead of `|`  in the `CodeIgniter\Controller`'s request docblock:

```bash

	/**
	 * Instance of the main Request object.
	 *
	 * @var RequestInterface&\CodeIgniter\HTTP\IncomingRequest&\CodeIgniter\HTTP\CLIRequest
	 */
	protected $request;
```

that will always work, what do you think?

paulbalandan:
But I have an experience with `phpstan` before, using `phpstan-strict-rules` I think, using the union `&` in the docblock then passing in a lower type will result in an error. For example here, `Property $request expects IncomingRequest, RequestInterface given`.

samsonasik:
PHPStan seems can't verify `&`  for the following: 

```bash
@var RequestInterface&\CodeIgniter\HTTP\IncomingRequest&\CodeIgniter\HTTP\CLIRequest
```

```bash
------ ---------------------------------------------------------------- 
  44     PHPDoc tag @var for property CodeIgniter\Controller::$request   
         contains unresolvable type. 
```

The mark as `IncomingRequest` in BaseController seems the possible solution for now.

MGatner:
This makes me a little uncomfortable with the PSR-7 stuff up in the air, but I think the changes there will require a major version so we could adjust this then anyways. For now I think this will fix a number of other issues for places uses `Request` and `RequestInterface`.

samsonasik:
@MGatner so this PR or the alternative https://github.com/codeigniter4/CodeIgniter4/pull/4282 ?

MGatner:
I don't really have any IDE experience, and you've taught me everything I know about PHPStan. I trust you two to figure this out. 😊

paulbalandan:
I'm uncomfortable with a `phpstan-ignore-next-line` in the phpdoc. Seems like we are suppressing or hiding something. I'd vote for this one instead. And probably we'd do `CLIRequest` for `CommandRunner::$request`.

samsonasik:
@MGatner thank you, I prefer the alternative https://github.com/codeigniter4/CodeIgniter4/pull/4282 but will wait if anyone have another opinion.

samsonasik:
@paulbalandan Ok, I will merge this 🎉

MGatner:
Actually, sorry to think about this late, but I think if you run `php public/index.php` then `BaseController::$request` will actually be a `CLIRequest`: https://github.com/codeigniter4/CodeIgniter4/blob/363b09094b5a37ce4796781c933091057e6ae8cb/system/CodeIgniter.php#L590

We need to test for this and either update the docblock or revert this change until we can figure this out.

MGatner:
https://github.com/codeigniter4/CodeIgniter4/blob/69038f652fdd63950dd355d87526b00a8fee0cbf/system/Common.php#L756-L759

MGatner:
@paulbalandan @samsonasik Please don't let this one slip by

samsonasik:
probably `@var IncomingRequest&CliRequest` is the way to go?

paulbalandan:
> probably `@var IncomingRequest&CliRequest` is the way to go?

Hmm. I think this is the way.

samsonasik:
@MGatner fixed at https://github.com/codeigniter4/CodeIgniter4/pull/4285

PR #4285, opened 2021-02-13
Add CLIRequest to request property in BaseController

samsonasik:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -3,6 +3,7 @@
 namespace App\Controllers;
 
 use CodeIgniter\Controller;
+use CodeIgniter\HTTP\CLIRequest;
 use CodeIgniter\HTTP\IncomingRequest;
 use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
@@ -24,7 +25,7 @@ class BaseController extends Controller
 	/**
 	 * Instance of the main Request object.
 	 *
-	 * @var IncomingRequest
+	 * @var IncomingRequest&CLIRequest
 	 */
 	protected $request;
 
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -25,7 +25,8 @@ class BaseController extends Controller
 	/**
 	 * Instance of the main Request object.
 	 *
-	 * @var IncomingRequest&CLIRequest
+	 * @phpstan-ignore-next-line
+         * @var IncomingRequest&CLIRequest
 	 */
 	protected $request;
 
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -25,8 +25,7 @@ class BaseController extends Controller
 	/**
 	 * Instance of the main Request object.
 	 *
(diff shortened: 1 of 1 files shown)


app/Controllers/BaseController.php
```diff
@@ -24,7 +25,8 @@ class BaseController extends Controller
 	/**
 	 * Instance of the main Request object.
 	 *
-	 * @var IncomingRequest
+	 * @phpstan-ignore-next-line
+         * @var IncomingRequest&CLIRequest
```
paulbalandan, line 29: Ah my bad. Should this be `IncomingRequest|CLIRequest` instead? Because `&` means the property type can be `IncomingRequest` and `CLIRequest` at the same time.

paulbalandan: Just like in PHPUnit's mocking objects. It uses `&` between the `MockObject` and the mocked class to mean both types co-exist in the object mocked.

samsonasik: that seems not detected in IDE? I tried with include RequestInteface and sems intelephense only detect RequestInterface, I use mobile interface so I can't check yet for only these 2 with pipe. could you verify, thank you.

paulbalandan: I'll check

paulbalandan: This one uses `IncomingRequest|CLIRequest` and testing on VSCode.

image

paulbalandan: phpstan is no errors.

samsonasik: thank you, I think that should go to `CodeIgniter\Controller` class instead.

paulbalandan: Yes, I think so.

samsonasik: or not, as in `CodeIgniter\Controller` will got error, placing here seems fine.

samsonasik: Just back to computer, trying in `CodeIgniter\Controller` got the following error:

```bash
  84     Property CodeIgniter\Controller::$request                            
         (CodeIgniter\HTTP\CLIRequest|CodeIgniter\HTTP\IncomingRequest) does  
         not accept CodeIgniter\HTTP\RequestInterface.
```

seems placing in this seems fine

paulbalandan: If you'll combine the three in pipes in `CodeIgniter\Controller` it will work `@var RequestInterface|IncomingRequest|CLIRequest`

samsonasik: updated.

samsonasik: @paulbalandan using `@var RequestInterface|IncomingRequest|CLIRequest` unfortunatelly not working on IDE ( I use VsCode):

Screen Shot 2021-02-14 at 00 05 37

paulbalandan: Ok. I see. I used `getPath()` and it turned red. Seems here is the right place.

MGatner (commented):
Looks good, thanks!

PR #4287, opened 2021-02-14
Fix Query Builder limit for SQL Server 2008

abydahana:
The Query Builder will throw an error on  SQL Server 2008 (Version 10.x) when using limit since it doesn't yet support offset. This commit is part of worked limitation from CodeIgniter 3 and has been tested on SQL Server 2008.

### DIFF
--- a/system/Database/SQLSRV/Builder.php
+++ b/system/Database/SQLSRV/Builder.php
@@ -369,19 +369,19 @@ class Builder extends BaseBuilder
 			// Get the fields to select from our subquery, so that we can avoid CI_rownum appearing in the actual results
 			if (count($this->QBSelect) === 0 OR strpos(implode(',', $this->QBSelect), '*') !== FALSE)
 			{
-				$select = '*'; // Inevitable
+				$select			= '*'; // Inevitable
 			}
 			else
 			{
 				// Use only field names and their aliases, everything else is out of our scope.
-				$select = array();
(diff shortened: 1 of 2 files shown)


samsonasik:
please run:

```bash
vendor/bin/rector process
```
and commit to fix rector notice.

abydahana:
> please run:
> 
> ```shell
> vendor/bin/rector process
> ```
> 
> and commit to fix rector notice.

Any updates or another suggestion of this commit?

samsonasik:
could you provide unit test for it? Thank you.

abydahana:
Sure. How could i present the unit testing since it require the 3rd party DBMS such SQL Server? Please give me advice.

michalsn (changes requested):
We can't use Live tests, but we can test it using Builder tests.

Please take a look here to see how we use `SQLSRVBuilder` to make test exclusively run with SQLSRV: https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/Builder/JoinTest.php#L88

Based on this example it should be easy to do - you will be able to set desired DB version via `MockConnection`, I think.

paulbalandan:
This may be useful for users of sql server but no activity has been made for testing this change. The last 2 commits are out of scope of the PR. Closing this for now. Please open a new PR if you can add tests for this.

PR #4292, opened 2021-02-16
[Rector] Apply Rector: InlineIfToExplicitIfRector

samsonasik:
Change inline if to explicit if with `InlineIfToExplicitIfRector` rector rule.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -2367,7 +2367,10 @@ class Email
 
 		is_array($include) || $include = [$include]; // @phpstan-ignore-line
 
-		in_array('headers', $include, true) && $rawData  = htmlspecialchars($this->headerStr) . "\n";
+		if (in_array('headers', $include, true))
+		{
+			$rawData = htmlspecialchars($this->headerStr) . "\n";
+		}
 		in_array('subject', $include, true) && $rawData .= htmlspecialchars($this->subject) . "\n";
 		in_array('body', $include, true) && $rawData    .= htmlspecialchars($this->finalBody);
 
--- a/.github/workflows/test-rector.yml
+++ b/.github/workflows/test-rector.yml
@@ -26,8 +26,12 @@ on:
 
 jobs:
   build:
-    name: Analyze code (Rector)
+    name: PHP ${{ matrix.php-versions }} Analyze code (Rector)
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.4', '8.0']
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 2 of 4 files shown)


samsonasik:
I included phpstan stub `ReflectionUnionType` at https://github.com/codeigniter4/CodeIgniter4/pull/4292/commits/5e71bf9329acf48ea17b0db6230cf4589c1b2f4d for handle  "Class 'ReflectionUnionType' not found" in php < 8.0

samsonasik:
I updated github action config to use php 7.4 and 8.0 for rector check to ensure work in both php 7 and php 8

samsonasik:
All green 🎉

system/Email/Email.php
```diff
@@ -2367,7 +2367,10 @@ public function printDebugger($include = ['headers', 'subject', 'body'])
 
 		is_array($include) || $include = [$include]; // @phpstan-ignore-line
```
paulbalandan, line 2368: I know it is out-of-scope, but should `InlineIfToExplicitIfRector` also refactor boolean OR and concatenations? Looking at the code it only deals with `BooleanAnd` and `Assign` types.

system/Email/Email.php
```diff
@@ -2367,7 +2367,10 @@ public function printDebugger($include = ['headers', 'subject', 'body'])
...
 
-		in_array('headers', $include, true) && $rawData  = htmlspecialchars($this->headerStr) . "\n";
+		if (in_array('headers', $include, true))
+		{
+			$rawData = htmlspecialchars($this->headerStr) . "\n";
+		}
 		in_array('subject', $include, true) && $rawData .= htmlspecialchars($this->subject) . "\n";
 		in_array('body', $include, true) && $rawData    .= htmlspecialchars($this->finalBody);
```
paulbalandan, line 2375: If we could have the Rector rule cover these two also that would be great.

paulbalandan (approved):
I approve, but I have some comments that may be better addressed at `rectorphp/rector` for improvement of the rector rule.

system/Email/Email.php
samsonasik: Yes, BooleanOr should be covered, that maybe next improvement 👍

system/Email/Email.php
samsonasik: Yes, hopefully next improvement 👍

system/Email/Email.php
samsonasik: implemented in rector dev-master, wait for next release ;)

system/Email/Email.php
samsonasik: implemented in rector dev-master, wait for next release ;)

MGatner (commented):
I'm a fan of this change!

PR #4293, opened 2021-02-16
[libraries/validation] make userguide same as source

totoprayogo1916:
**Description**
adjust the guide the same as the source.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide


paulbalandan (commented):
This can be a bit misleading to other users if they do not know of namespaces and importing symbols using `use`. I suggest adding `use CodeIgniter\Validation\Rules;` etc. so that new users can still follow.

paulbalandan (approved):

PR #4295, opened 2021-02-17
Friendlier error page to match default home style.

lonnieezell:
Matches the `error_exception` page to the default home page styling for more consistency in the product and a friendlier error message (no more in-your-face red bar).


MGatner (commented):
I love the DuckDuckGo change ;) Nice touch.

Requisite "will this play nicely with the CSS work @LittleJ has done"?

lonnieezell:
@MGatner Good question, but it is about the same as what existed previously, just with colors to match home page.

MGatner (approved):

PR #4296, opened 2021-02-17
Add more info to seeInDatabase failures.

lonnieezell:
Refactored the `seeInDatabase` assertion into a Constraint so we can provide context when it fails by showing potentially similar rows within the database.


system/Test/Constraints/SeeInDatabase.php
```diff
@@ -0,0 +1,126 @@
...
+	{
+		$builder = $this->db->table($table);
+
+		$similar = $builder->where(
+				array_key_first($this->data),
+				$this->data[array_key_first($this->data)]
+			)->limit($this->show)
+			->get()->getResultArray();
+
+		if (count($similar))
```
samsonasik, line 86: using `$similar !== []` should be faster: 

```suggestion
              if ($similar !== [])
```

 I think it is possibly next improvement for `CountArrayToEmptyArrayComparisonRector` rector rule:

https://github.com/codeigniter4/CodeIgniter4/blob/ae833c1b21cae7ec65ba1b96a6588cd7b6c5b504/rector.php#L54

which currently can't detect direct truthy `count($var)` cond.

lonnieezell: Changed.

system/Test/Constraints/SeeInDatabase.php
```diff
@@ -0,0 +1,126 @@
...
+		}
+		else
+		{
+			// Does the table have any results at all?
+			$results = $this->db->table($table)
+				->limit($this->show)
+				->get()
+				->getResultArray();
+
+			if (count($results) === 0)
```
samsonasik, line 98: Can use `$results !== []` as well. This seems doesn't detected by rector due to `Builder->get()` can return both `ResultInterface|false`.

https://github.com/codeigniter4/CodeIgniter4/blob/b833a002f0eb1185fbf30a52c7dadc6632b2900b/system/Database/BaseBuilder.php#L1855

If the `@return` docblock is correct, I think it is better to check if `Builder->where()->get()` result `ResultInterface` first in line 83 to avoid error: undefined method `getResultArray` on false.

lonnieezell: Looks like it will only return `false` if `query()` is run within a transaction, which this isn't. So I think we should be fine as is.

lonnieezell: Change pushed.

MGatner:
This looks good but will need to be rebased for the `DatabaseTestTrait` PRs.

MGatner (changes requested):
Please rebase after #4337

Issue #4235, opened 2021-02-05
Bug: Wrong HTML code in output of "form_input" helper function


lbreme:
**Describe the bug**
When using the "form_input" helper function with these example parameters:
<pre>
&lt;?php
$identity = [
	'name' =>  'identity',
	'id'   =>  'identity',
	'type' =>  'email',
	'value' => ''
];
echo form_input($identity, '', array('class' => 'form-control form-control-lg'));
?&gt;
</pre>
I get the following incorrect HTML code:

<pre>
&lt;input type=&quot;email&quot; name=&quot;identity&quot;value=&quot;&quot; id=&quot;identity&quot; class=&quot;form-control&amp;#x20;form-control-lg&quot; /&gt;
</pre>

There are two errors:

1) the space between <i>name=&quot;identity&quot;</i> and <i>value=&quot;&quot;</i> is missing.
2) the value of the attribute <i>class</i> is encoded in its entirety (therefore the space becomes erroneously &amp;#x20;), when instead it should be encoded for each single value separated by a space (i.e. in this case <i>form-control</i> and <i>form-control-lg</i>)

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
The first error is in the <i>parse_form_attributes</i> function (line 947) in the helper file called <i>/system/Helpers/form_helper.php</i>
The second error is in the <i>stringify_attributes</i> function (line 1136) in the file <i>/system/Common.php</i>

**Expected behavior, and steps to reproduce if appropriate**
I expect the <i>form_input</i> function to return the following correct HTML code:

<pre>
&lt;input type=&quot;email&quot; name=&quot;identity&quot; value=&quot;&quot; id=&quot;identity&quot; class=&quot;form-control form-control-lg&quot; /&gt;
</pre>

If I may, I would propose the following solutions to the two problems:
1) replace line number 947 in the <i>form_helper.php</i> file with:
<pre>
$att .= $key . '=&quot;' . $val . '&quot;' . ($key == key(array_slice($default, -1, 1, true)) ? '' : ' ');

or for PHP >= 7.3.0:

$att .= $key . '=&quot;' . $val . '&quot;' . ($key === array_key_last($default)) ? '' : ' ');
</pre>
2) Replace line number 1136 in the <i>Common.php</i> file with:
<pre>
$atts .= ($js) ? $key . '=' . $val. ',' : ' ' . $key . '=&quot;' . $val . '&quot;'; // without any attribute escaping
</pre>

[...]

PR #4298, opened 2021-02-17
Bug #4235 fix. Wrong HTML code in output of "form_input" helper function

lbreme:
**Description**
I fixed error number 1 as documented in my bug #4235 (file _/system/Helpers/form_helper.php_). On the other hand, error number 2 I did not fix as documented in my bug #4235, because the problem is actually in the _/system/ThirdParty/Escaper/Escaper.php_ library file, where spaces in HTML, JAVASCRIPT and CSS attributes are also incorrectly encoded.

**Checklist:**
- [*] Securely signed commits
- [ ] User guide updated
- [*] Conforms to style guide

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -944,7 +944,7 @@ if (! function_exists('parse_form_attributes'))
 				{
 					continue;
 				}
-				$att .= $key . '="' . $val . '"' . ($val === end($default) ? '' : ' ');
+				$att .= $key . '="' . $val . '"' . ($key === key(array_slice($default, -1, 1, true)) ? '' : ' ');
 			}
 			else
 			{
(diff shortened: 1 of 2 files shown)


samsonasik:
please don't change ThirdParty content as that will be replaced when there is new release. You can send PR to Laminas Escaper repository instead for that.

https://github.com/laminas/laminas-escaper

lbreme:
> 
> 
> please don't change ThirdParty content as that will be replaced when there is new release. You can send PR to Laminas Escaper repository instead for that.
> 
> https://github.com/laminas/laminas-escaper

Ok, thank you so much. In fact, it's not a Codeigniter problem.

totoprayogo1916:
Is accepted to change  `esc($val, 'attr')` to `esc($val, 'html')` @samsonasik ?

https://github.com/codeigniter4/CodeIgniter4/blob/e4661270a2b4ce12bbc2ec456ea984e3f149ec36/system/Common.php#L1176

samsonasik:
@totoprayogo1916 `html` is default parameter value, `esc($val)` is equal. That need additional/update unit test.

PR #4307, opened 2021-02-19
updated Router to properly translate uri dashes that map to controller subdirs

### LINKED ISSUE #4294: Bug: Cannot add route to controller in filename with dash/hyphen
**Describe the bug**
CodeIgniter 4's "translate uri dashes" functionality does not work under certain circumstances. If you want dashes in your uri that map into dashes in some controller subdirectory name, then you cannot add a route targeting controllers in that subdirectory.

I have an [earlier post on the Codeigniter forum](https://forum.codeigniter.com/thread-78215.html) relating to this where I am trying to support URLs with dashes in them. Long story short, I must use dashes in my controller subdirectory names if I want to have dashes in my url. For whatever reason, the CI4 auto-routing is unable to recognize when a url with a dash in the path is referring to a controller subdirectory with an underscore.

E.g., if I want this url to connect to the default **Home** controller in my **Nested-directory** subdirectory:

`https://www.example.com/subdir/nested-directory`

Then I have to use a dash in the file system directory name. That url works if I put my controller in this location:

`app/Controllers/Subdir/Nested-directory/Home.php`

*Note* that the properly working Home controller uses an underscore in its namespace:
```php
namespace App\Controllers\Subdir\Nested_directory;

class Home extends BaseController {

	public function index()
	{
		die("this is " . __METHOD__);
	}
}
```

That same url does *not* work if I put it in a directory with an underscore:

`app/Controllers/Subdir/Nested_directory/Home.php`

Curiously, it the dash in the url I request `does` get properly routed to an underscore while CI4 is calculating the controller path. *However*, CI4 doesn't bother to check if the url is requesting the default Home controller in a subdirectory. The complaint it gives is:

> Controller or its method is not found: \App\Controllers\Subdir\Nested_directory::index

So it *almost* works, is just doesn't bother checking for the default **Home** controller in a subdir.  Additionally, if I change my url to have an underscore in it, then CI4 can locate the Home controller in the subdirectory named with an underscore. This url:

`https://www.example.com/subdir/nested_directory`

Connects correctly to the controller here:

`app/Controllers/Subdir/Nested_directory/Home.php`

Previously, it was enough to simply make sure I name my subdirectories with dashes so they match the urls, however **this does not work when i must add a $routes->add directive mapping urls to a controller that happens to live in a subdirectory with a dash.** This route directive works just fine in my Config/Routes file if **Nested_directory** contains an underscore:

```php
$routes->add("foobar/([0-9]+)", "Subdir\Nested_directory\Home::index/$1");
```

This serves a request for https://www.example.com/foobar/123 with the correct Home controller.

But if if the **Nested-directory** subdirectory containing my Home controller contains a dash, meaning my Home controller file is

`app/Controllers/Subdir/Nested-directory/Home.php`

then this will not work:

```php
$routes->add("foobar/([0-9]+)", "Subdir\Nested_directory\Home::index/$1");
```

This will also not work:

```php
$routes->add("foobar/([0-9]+)", "Subdir\Nested-directory\Home::index/$1");
```

Please note that I need to have these dashes instead of underscores because that is [recommended by Google](https://developers.google.com/search/docs/advanced/guidelines/url-structure). I also very much do not want to create a big, complicated Routes file where I must route all of my many controllers. The auto-routing of CI3 was working just fine. I think this problem could be remedied with a small code change.

Can anyone point me to a solution?

For reference, here is my Config/Routes.php file

```php
// Create a new instance of our RouteCollection class.
$routes = Services::routes();

// Load the system's routing file first, so that the app and ENVIRONMENT
// can override as needed.
if (file_exists(SYSTEMPATH . 'Config/Routes.php'))
{
 require SYSTEMPATH . 'Config/Routes.php';
}

/**
 * Router Setup
 */
$routes->setDefaultNamespace('App\Controllers');
$routes->setDefaultController('Home');
$routes->setDefaultMethod('index');
$routes->setTranslateURIDashes(true);
$routes->set404Override();
$routes->setAutoRoute(true);

/**
 * Route Definitions
 */

// We get a performance increase by specifying the default
// route since we don't have to scan directories.
$routes->get('/', 'Home::index');

// neither of these work if the controller is in Nested-directory
//$routes->add("foobar/([0-9]+)", "Subdir\Nested-directory\Home::index/$1");
$routes->add("foobar/([0-9]+)", "Subdir\Nested_directory\Home::index/$1");
```
**CodeIgniter 4 version**
4.1.1

**Affected module(s)**
system/Router ? not entirely sure.

1) I would expect CI4 to auto route a request for example.com/subdir/nested-directory to any of these controllers (if they exist), ideally in the sequence here:

Controllers/Subdir/Nested_directory::index
Controllers/Subdir/Nested-directory/Home::index
Controllers/Subdir/Nested_directory/Home::index

2) I might also expect adding a route like this:
```php
$routes->add("foobar/([0-9]+)", "Subdir\Nested-directory\Home::index/$1");
```
to locate the controller at `Controllers/Subdir/Nested-directory/Home::index`. Note this dash translation for added routes may not be necessary if the autorouting feature can properly translate uri dashes.

**Context**
 - OS: Ubuntu 20.04 LTS
 - Web server: Apache 2.4.41
 - PHP version: 7.4

sneakyimp:
Fixes #4294 

**Description**
I have made a minor tweak to CodeIgniter\Router::validateRequest so that dashes in a subdirectory are mapped to underscores if `translateURIDashes`= `true`. This appears to correctly auto-route requests when you have a dash in the URI that corresponds to a controller directory.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -594,11 +594,13 @@ class Router implements RouterInterface
 		// is found or when such a directory doesn't exist
 		while ($c-- > 0)
 		{
-			$test = $this->directory . ucfirst($this->translateURIDashes === true ? str_replace('-', '_', $segments[0]) : $segments[0]);
+			$segmentConvert = ucfirst($this->translateURIDashes === true ? str_replace('-', '_', $segments[0]) : $segments[0]);
+			$test           = APPPATH . 'Controllers/' . $this->directory . $segmentConvert;
 
-			if (! is_file(APPPATH . 'Controllers/' . $test . '.php') && $directoryOverride === false && is_dir(APPPATH . 'Controllers/' . $this->directory . ucfirst($segments[0])))
+			if (! is_file($test . '.php') && $directoryOverride === false && is_dir($test))
 			{
-				$this->setDirectory(array_shift($segments), true);
+				$this->setDirectory($segmentConvert, true);
+				array_shift($segments);
 				continue;
 			}
 
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -692,7 +692,7 @@ class Router implements RouterInterface
 	 */
 	private function isValidSegment(string $segment): bool
 	{
-		return (bool)preg_match('/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/', $segment);
+		return (bool) preg_match('/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/', $segment);
 	}
 
 	//--------------------------------------------------------------------
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -498,7 +498,7 @@ class Router implements RouterInterface
 	{
 		$segments = explode('/', $uri);
 
-		$segments = $this->validateRequest($segments);
+		$segments = $this->scanControllers($segments);
 
 		// If we don't have any segments left - try the default controller;
 		// WARNING: Directories get shifted out of the segments array.
@@ -510,6 +510,12 @@ class Router implements RouterInterface
 		else
 		{
 			$this->controller = ucfirst(array_shift($segments));
+		}
+
+		$controllerName   = $this->controllerName();
+		if (!$this->isValidSegment($controllerName))
+		{
+			throw new PageNotFoundException($this->controller . " is not a valid controller name");
 		}
 
 		// Use the method name if it exists.
@@ -526,7 +532,6 @@ class Router implements RouterInterface
 		}
 
 		$defaultNamespace = $this->collection->getDefaultNamespace();
-		$controllerName   = $this->controllerName();
 		if ($this->collection->getHTTPVerb() !== 'cli')
 		{
 			$controller = '\\' . $defaultNamespace;
@@ -573,33 +578,45 @@ class Router implements RouterInterface
 	//--------------------------------------------------------------------
 
 	/**
-	 * Attempts to validate the URI request and determine the controller path.
+	 * Scans the controller directory, attempting to locate a controller matching the supplied uri $segments
 	 *
 	 * @param array $segments URI segments
 	 *
-	 * @return array URI segments
+	 * @return array returns an array of remaining uri segments that don't map onto a directory
 	 */
-	protected function validateRequest(array $segments): array
+	protected function scanControllers(array $segments): array
 	{
 		$segments = array_filter($segments, function ($segment) {
-			// @phpstan-ignore-next-line
-			return ! empty($segment) || ($segment !== '0' || $segment !== 0);
+			return $segment !== '';
 		});
+		// numerically reindex the array, removing gaps
 		$segments = array_values($segments);
 
-		$c                 = count($segments);
-		$directoryOverride = isset($this->directory);
+
+		// if a prior directory value has been set, just return segments and get out of here
+		if (isset($this->directory))
+		{
+			return $segments;
+		}
+
 
 		// Loop through our segments and return as soon as a controller
 		// is found or when such a directory doesn't exist
+		$c                 = count($segments);
 		while ($c-- > 0)
 		{
 			$segmentConvert = ucfirst($this->translateURIDashes === true ? str_replace('-', '_', $segments[0]) : $segments[0]);
+			// as soon as we encounter any segment that is not PSR-4 compliant, stop searching
+			if (!$this->isValidSegment($segmentConvert)) {
+				return $segments;
+			}
+
 			$test           = APPPATH . 'Controllers/' . $this->directory . $segmentConvert;
 
-			if (! is_file($test . '.php') && $directoryOverride === false && is_dir($test))
+			// as long as each segment is *not* a controller file but does match a directory, add it to $this->directory
+			if (! is_file($test . '.php') && is_dir($test))
 			{
-				$this->setDirectory($segmentConvert, true);
+				$this->setDirectory($segmentConvert, true, false);
 				array_shift($segments);
 				continue;
 			}
@@ -617,26 +634,50 @@ class Router implements RouterInterface
 	 * Sets the sub-directory that the controller is in.
 	 *
 	 * @param string|null   $dir
-	 * @param boolean|false $append
+	 * @param boolean $append
+	 * @param boolean $validate if true, checks to make sure $dir consists of only PSR4 compliant segments
 	 */
-	public function setDirectory(string $dir = null, bool $append = false)
+	public function setDirectory(string $dir = null, bool $append = false, bool $validate = true)
 	{
 		if (empty($dir))
 		{
 			$this->directory = null;
-			return;
+			return true;
 		}
 
-		$dir = ucfirst($dir);
+		if ($validate)
+		{
+			$segments = explode("/", trim($dir, "/"));
+			foreach($segments as $segment)
+			{
+				if (!$this->isValidSegment($segment))
+				{
+					return;
+				}
+			}
+		}
 
 		if ($append !== true || empty($this->directory))
 		{
-			$this->directory = str_replace('.', '', trim($dir, '/')) . '/';
+			$this->directory = trim($dir, "/") . "/";
 		}
 		else
 		{
-			$this->directory .= str_replace('.', '', trim($dir, '/')) . '/';
+			$this->directory .= trim($dir, "/") . "/";
 		}
+
+	}
+
+	/**
+	 * Returns true if the supplied $segment string represents a valid PSR-4 compliant namespace/directory segment
+	 *
+	 * regex comes from https://www.php.net/manual/en/language.variables.basics.php
+	 *
+	 * @param string $segment
+	 * @return bool
+	 */
+	private function isValidSegment(string $segment): bool {
+		return (boolean)preg_match('/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/', $segment);
 	}
 
 	//--------------------------------------------------------------------
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -512,10 +512,10 @@ class Router implements RouterInterface
 			$this->controller = ucfirst(array_shift($segments));
 		}
 
-		$controllerName   = $this->controllerName();
-		if (!$this->isValidSegment($controllerName))
+		$controllerName = $this->controllerName();
+		if (! $this->isValidSegment($controllerName))
 		{
-			throw new PageNotFoundException($this->controller . " is not a valid controller name");
+			throw new PageNotFoundException($this->controller . ' is not a valid controller name');
 		}
 
 		// Use the method name if it exists.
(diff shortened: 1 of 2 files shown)


sneakyimp:
@paulbalandan @kenjis @michalsn Could someone review this PR to fix #4294?

paulbalandan (commented):
I've looked at the changes and it looks good to me. I agree with the changes you've made. I have a minor cosmetic request. Change `$segmentConvert` to simply `$segment`. :)

I'll check first if this method is covered by tests or not before approving this for merge.

paulbalandan:
Can you add a test case for this one similar to here https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Router/RouterTest.php#L215-L227 but using the underscored subfolder as in your case. It should assert that it will still get the correct controller and method.

sneakyimp:
Thank you, @paulbalandan!

> I've looked at the changes and it looks good to me. I agree with the changes you've made. I have a minor cosmetic request. Change `$segmentConvert` to simply `$segment`. :)

Please note that I chose a different variable name to distinguish this value from the raw segments referred to elsewhere. It was my feeling that this more clearly indicated that the segment had been changed with ucfirst and with the dash-to-underscore replacement. I'd like to keep it because I think it makes the intent of the code much clearer. It was pretty difficult to understand the intent of this code because there are so few comments and the variables and functions are poorly named. In particular, the function `validateRequest` is very poorly named. I'd very much like to give that function a different name but have not done so in order to expedite matters. Please let me know if I may keep this other var name for greater clarity?

I have another question: may I remove the ucfirst call in [setDirectory](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Router/Router.php#L620)? This function is only called in the validateRequest function, where I have already applied ucfirst to the supplied parameter, and in [system/Test/FeatureTestTrait.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Test/FeatureTestTrait.php#L191) where the supplied parameter is `null`.


> 
> I'll check first if this method is covered by tests or not before approving this for merge.

I have run the phpunit test and it passed all the previous tests, but please do run any additional tests:

`vendor/phpunit/phpunit/phpunit tests/system/Router`

I'll work on adding another test as you directed.

paulbalandan:
The tests run pretty smoothly and in fact the coverage increased for `Router` so that's good. But since we're fixing a bug here we'll want to document it that the fix has been well tested.

sneakyimp:
@paulbalandan I have committed another set of changes. NOTE that it implements more stringent enforcement of PSR-4 compliance in that it will throw PageNotFoundExceptions for controller names which are not acceptable PHP class names. This has the benefit of rejecting any `..` or `.` attempts which may sneak into the Router but it will also reject controller names with dashes or periods or puncutation, etc. This should make the Router a bit more efficient because it eliminates a fair amount of additional logic that was getting executed.

Note that I did change `RouterTest::testZeroAsURIPath()` to expect an Exception instead of passing. A zero is not a valid Controller name -- you can't declare a `class 0 {};` in PHP nor can you use a zero in any namespace declaration.

system/Router/Router.php
```diff
@@ -614,29 +632,53 @@ protected function validateRequest(array $segments): array
...
+	 * @param boolean     $validate if true, checks to make sure $dir consists of only PSR4 compliant segments
 	 */
-	public function setDirectory(string $dir = null, bool $append = false)
+	public function setDirectory(string $dir = null, bool $append = false, bool $validate = true)
 	{
 		if (empty($dir))
 		{
 			$this->directory = null;
-			return;
+			return true;
```
kenjis, line 644: Why returns `true`?

system/Router/Router.php
```diff
@@ -614,29 +632,53 @@ protected function validateRequest(array $segments): array
...
+	 * Returns true if the supplied $segment string represents a valid PSR-4 compliant namespace/directory segment
+	 *
+	 * regex comes from https://www.php.net/manual/en/language.variables.basics.php
+	 *
+	 * @param  string $segment
+	 * @return boolean
+	 */
+	private function isValidSegment(string $segment): bool
+	{
+		return (boolean)preg_match('/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/', $segment);
```
kenjis, line 679: `(boolean)` is not used.

```suggestion
		return (bool) preg_match('/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/', $segment);
```

system/Router/Router.php
```diff
@@ -573,32 +578,45 @@ public function autoRoute(string $uri)
...
-	 * Attempts to validate the URI request and determine the controller path.
+	 * Scans the controller directory, attempting to locate a controller matching the supplied uri $segments
 	 *
 	 * @param array $segments URI segments
 	 *
-	 * @return array URI segments
+	 * @return array returns an array of remaining uri segments that don't map onto a directory
 	 */
-	protected function validateRequest(array $segments): array
+	protected function scanControllers(array $segments): array
```
paulbalandan, line 587: As this method is not flagged as `@internal`, changing the name is technically BC breaking.

system/Router/Router.php
```diff
@@ -614,29 +632,53 @@ protected function validateRequest(array $segments): array
...
 
-		$dir = ucfirst($dir);
+		if ($validate)
+		{
+			$segments = explode('/', trim($dir, '/'));
+			foreach ($segments as $segment)
+			{
+				if (! $this->isValidSegment($segment))
+				{
+					return;
```
paulbalandan, line 654: What is the real return type of this method? This line returns `void`, but line 644 returns `bool`.

system/Router/Router.php
```diff
@@ -573,32 +578,45 @@ public function autoRoute(string $uri)
...
+		// numerically reindex the array, removing gaps
 		$segments = array_values($segments);
 
-		$c                 = count($segments);
-		$directoryOverride = isset($this->directory);
+		// if a prior directory value has been set, just return segments and get out of here
+		if (isset($this->directory))
+		{
+			return $segments;
+		}
```
paulbalandan, line 599: Why is this necessary? As this is untested, can you add a test case covering this possibility.

system/Router/Router.php
```diff
@@ -614,29 +632,53 @@ protected function validateRequest(array $segments): array
...
+		{
+			$segments = explode('/', trim($dir, '/'));
+			foreach ($segments as $segment)
+			{
+				if (! $this->isValidSegment($segment))
+				{
+					return;
+				}
+			}
+		}
```
paulbalandan, line 657: This addition is not tested.

system/Router/Router.php
paulbalandan: Or can be `preg_match('/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/', $segment) === 1`.

system/Router/Router.php
sneakyimp: Thank you so much for reviewing the code closely!

> Why returns `true`?

Oh no that's a mistake. I had considered having setDirectory return a boolean value if the directory was valid. I changed my mind but accidentally left this in there. I will remove the return value of true.

system/Router/Router.php
sneakyimp: I'll change the cast to (bool).

system/Router/Router.php
sneakyimp: > As this method is not flagged as `@internal`, changing the name is technically BC breaking.

In order for this change to break any code, someone would have to have extended the router class. I did a grep search of the source and could find no place where it extends this class. This returns no results.
```
grep -r "extends.*Router" CodeIgniter4
```
If we are really concerned about it being a breaking change, perhaps we can define `validateRequest()` to simply call `scanControllers` and deprecate it:
```php
/**
 * @deprecated
 */
protected function validateRequest(array $segments): array
{
    return $this->scanControllers($segments);
}
```

If you want me to change the function name back to `validateRequest` I will do it, but it is a really bad name that does NOT describe what that function is doing.

system/Router/Router.php
sneakyimp: `setDirectory` should not return anything. The return value of `true` is a mistake I made and I will remove it.

system/Router/Router.php
sneakyimp: I'm not entirely sure and will dig deeper. I definitely think we could use some more test cases and will try to add them.

system/Router/Router.php
paulbalandan: But users may be extending the Router class for their own use case and we cannot be sure that they did not use this method. I'll go with the deprecation instead and reroute calls to scanController.

sneakyimp:
@paulbalandan I've pushed another commit and I think it resolves the issues you mentioned in your review -- for some reason github only offers me a "resolve conversation" button rather than continuing our discussion. I hope you will please take a look and let me know if you have further concerns.

system/Router/Router.php
```diff
@@ -614,29 +648,53 @@ protected function validateRequest(array $segments): array
...
+	 * Returns true if the supplied $segment string represents a valid PSR-4 compliant namespace/directory segment
+	 *
+	 * regex comes from https://www.php.net/manual/en/language.variables.basics.php
+	 *
+	 * @param  string $segment
+	 * @return boolean
+	 */
+	private function isValidSegment(string $segment): bool
+	{
+		return (bool)preg_match('/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/', $segment);
```
paulbalandan, line 695: CS: Add space after the cast. `(bool) preg_match`

system/Router/Router.php
```diff
@@ -573,32 +578,61 @@ public function autoRoute(string $uri)
...
+			}
+
+			$test = APPPATH . 'Controllers/' . $this->directory . $segmentConvert;
 
-			if (! is_file(APPPATH . 'Controllers/' . $test . '.php') && $directoryOverride === false && is_dir(APPPATH . 'Controllers/' . $this->directory . ucfirst($segments[0])))
+			// as long as each segment is *not* a controller file but does match a directory, add it to $this->directory
+			if (! is_file($test . '.php') && is_dir($test))
 			{
-				$this->setDirectory(array_shift($segments), true);
+				$this->setDirectory($segmentConvert, true, false);
```
paulbalandan, line 634: I'm wondering. You added a third param to `setDirectory` optionally set to `true` but purposely set it to `false` here. As you have promoted a stricter check on the segments, no validation here seems a bit off. Or am I missing something?

sneakyimp:
@paulbalandan sorry I'm slow to respond. For some reason github isn't letting me respond directly under your review comment so I'm commenting here.

>CS: Add space after the cast. (bool) preg_match

I have added the space, committed to my local repo, rebased on the latest develop branch of CI4, and pushed. Please let me know if you have additional requests, questions, or issues.

>I'm wondering. You added a third param to setDirectory optionally set to true but purposely set it to false here. As you have promoted a stricter check on the segments, no validation here seems a bit off. Or am I missing something?

By default, setDirectory has `$validate=true`. The only reason I set it to `$validate=false` [when I call it in scanController](https://github.com/sneakyimp/CodeIgniter4/blob/bugs/issue-4294/system/Router/Router.php#L634) is because I have already [validated $segmentConvert a few lines before](https://github.com/sneakyimp/CodeIgniter4/blob/bugs/issue-4294/system/Router/Router.php#L624). My thinking was that it would be inefficient to validate the segment twice. There might be some better way to do this -- I considered having setDirectory return a boolean depending on whether the validation succeded or not, but setDirectory is a public method, and this would probably have introduced a Breaking Change.

sneakyimp:
I have no idea why these checks failed. They seem totally unrelated to my PR.

kenjis:
It is probably because ubuntu is now updated to 20.04.
But I don't know what's wrong.

paulbalandan:
Ubuntu latest is now 20.04 so only 7.4+ is built in in the image. We need to specify the extensions we need for lower php versions. I'll send a PR.

paulbalandan:
Please rebase,

sneakyimp (commented):
> Please rebase,

I have rebased and force-pushed. Please let me know if you have any additional thoughts.

paulbalandan (approved):

tests/system/Router/RouterTest.php
```diff
@@ -575,4 +723,35 @@ public function testRegularExpressionPlaceholderWithUnicode()
...
 	}
+
+	public function testRouterPriorDirectory()
+	{
+		$router = new Router($this->collection, $this->request);
+
+		$router->setDirectory('foo/bar/baz', false, true);
+		$router->handle('Some_controller/some_method/param1/param2/param3');
+
+		$this->assertEquals('foo/bar/baz/', $router->directory());
```
kenjis, line 734: @sneakyimp Why is this test necessary?
Why does `setDirectory()` take precedence over URI?

kenjis: I've found your answer.
https://github.com/codeigniter4/CodeIgniter4/pull/4307/files#r584375184

PR #4308, opened 2021-02-19
Alternative Entity type casting

iRedds:
**Description**
I propose an alternative implementation for type casting. 

- All type casts that existed by default have been moved to separate classes. 
- ~~Reworked magic methods __get and __set~~.
- ~~The setter method now has call priority.~~
- ~~If a getter/setter and type castings are defined, then only the getter/setter will be called.~~
- Added functionality for defining custom type casting. Custom type handlers have get and set methods. 
- Types can now have additional parameters.
- ~~All default type casts cannot be overridden.~~

Example
```php
    // Custom type registration 
    class MyEntity extends Entity
    {
        // Specifying the type for the field
        protected $casts = [
            'key' => 'base64',
        ];

        //Bind the type to the handler
        protected $castHandlers = [
            'base64' => 'App\Entity\Cast\CastBase64',
        ];
    }

    // Handler 
    class CastBase64 extends \CodeIgniter\EntityCast\AbstractCast
    {
        public static function get($value, array $params = [])
        {
            return base64_decode($value);
        }

        public static function set($value, array $params = [])
        {
            return base64_encode($value);
        }
    }
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


michalsn:
I like this. It adds a lot of flexibility but I think we have to make some of these features optional:

> * The setter method now has call priority.
> * If a getter/setter and type castings are defined, then only the getter/setter will be called.

If I'm not mistaken, those two will break the compatibility with the current version. So we have two options here: we can either get back to what we have now or make this an optional feature.

I can imagine that we can set a special flag in Entity that will make it work, so we can choose the type of functionality we expect from our entity. I know that it adds some complexity, but backward compatibility for Entities is a must-have.

system/Entity.php
```diff
@@ -527,69 +474,74 @@ protected function mutateDate($value)
...
+			$type   = $matches[1];
+			$params = array_map(function ($param) {
+				return trim($param);
+			}, explode(',', $matches[2]));
 		}
 
-		return $value;
+		$type = trim($type, '[]');
+
+		$handlers = array_merge($this->castHandlers, $this->defaultCastHandlers);
```
najdanovicivan, line 532: Should be inverted so that castHandles can override defaultCastHandlers

najdanovicivan (commented):
That's great.  Is there an ETA when this can be merged ?

Additionally I'd add property $initCastHandlers = [] and $defaultInitCastHandlers which would trigger casting in setAttributes method.

Then I'd introduce Config\Entity where defaultInitCastHandlers and defaultCastHandlers can be defined. Preferable it should be possible to Override this per module.

Another thing I'd like is to be able to define those casts to work with reference values. For example I have some item entities which have categories. And I want to display a lot of those on single page. I have slug for category in item. Then I load all the categories in some array. What I do next is only set a pointer to the category object inside item instead of copying those into items. That way memory usage can be reduced by a bit.

Also note the issue below

system/Entity.php
iRedds: I had doubts about replacing the default handlers.
Your opinion helped me decide

iRedds:
@michalsn My point of view is this. Setters and Getters are needed for special cases when you need to bypass type casting and implement other logic. 
I believed the implementation was the wrong implementation at the time of design. Therefore, I took responsibility for making changes in the behavior of Getters and Setters. 
At the moment, I don't know how best to implement the ability to switch to the priority mode of Getters and Setters without the headache, duplicate and ugly code.
Maybe someone will succeed.
But now I decided to return the old behavior.

iRedds:
@najdanovicivan I think the merge will be possible when the maintainers have no questions about the current implementation.

Can't imagine what the cast is for in the setAttributes method. 
Perhaps you mean some special case. 

If you like my PR, I think it would be better to wait for it to merge. And then you can submit your implementation based on my changes. It will be more clear with the code and examples.

najdanovicivan:
@iRedds cast in setAttributes is useful for nosql when you have nested objects. Let's take this one for example. We have a ecommerce site which support product variations which different pricing options. So you have something like this in MongoDB


```
{
    "varations" [
          {
              "name" : "item-only",
              "price": 1000
          },
          {
              "name" : "item-with-case",
              "price": 1500
          }
          {
              "name" : "item-with-accessories",
              "price": 2000
          }
     ]
}
```

We have 2 classes. Product and ProductVariation so we need to be able to add/modify those variations

```
$variation = new ProductVariation()
$variation->name = 'new'
$variation->price = 3000

//This one is not possible with magic for the time being
$product->variations[] = $variation;

//So we have to do
$variations = $product->variations; 
$variations[] = $variation;
$product->variations = $variations;

//This will work with setter/getter casting but will do the object construction each time getter is called

//Now this one will do completely nothing on the entity as setter will never be triggered
$product->variations[0]->name = 'New name';
```

Regarding the DB reading/writing we have toRawArray method for which if $deep parameter is passed as true it will handle the conversion of inner entities and we'll get the complete data to write in the db.

Problem with setAttributes is that it never goes through the setters so we'll have array in attributes instead of having an entity object stored in attributes.

iRedds:
@najdanovicivan Now I understand what you were talking about.
I have a persistent feeling that earlier on CI3 using Entity I faced a similar problem, but I cannot find where.
As far as I understand the concept of Entity applied during development is that the data should exist "AS IS". That is, having received a string from the database, we only transform it to receive it, but the data itself does not change.

I don't think it would be correct to change the behavior of the setAttributes method.
What is now done in the toRawArray method is a crutch, which, moreover, does not work correctly. 
Another approach must be found.

najdanovicivan:
https://github.com/codeigniter4/CodeIgniter4/blob/4cbfe2cd25ff8b08eee323cc734a2113cead8eb1/system/Database/MySQLi/Result.php#L159-L175

Looking at these the only thing setting the values inside entity form db is the setAttributes method. 

By adding defaultInitCastHandlers/InitCastHandlers we can allow for such behavior. Those will be empty by default so the existing functionality won't be affected but it will allow the feature to be used if needed. 

Having defaultInitCastHandlers in Config/Entity just provide means to setup those handlers globally. In the same manner it's done with validation rules. So that there is no need to define those in every Entity class

system/Entity.php
```diff
@@ -527,69 +508,67 @@ protected function mutateDate($value)
...
+		//Attempt to retrieve additional parameters if specified
+		// type[param, param2,param3]
+		if (preg_match('/^(.+)\[(.+)\]$/', $type, $matches))
+		{
+			$type   = $matches[1];
+			$params = array_map(function ($param) {
+				return trim($param);
+			}, explode(',', $matches[2]));
+		}
+
```
najdanovicivan, line 556: We should have `nullable` added in parameters `if (strpos($type, '?') === 0)`

There might be a situation where `null` might have to be return instead of $value. For example casting empty string to `?int`. The cast method should return null in such case but having `int` as cast should trow an exception or maybe in some custom cast class return 0.

najdanovicivan (changes requested):

system/Entity.php
iRedds: The nullable type is handled above.
I see no reason to apply nullable to parameters. They are not computed and are passed to the handler as a string

iRedds:
I understand the reason why you want to change the setAttributes method. But at the moment I am what the correct implementation should be. And I think that it is outside the scope of the current PR. 

There are 2 places in the setAttributes method where you can convert. 
1. Before $this->attributes = $data;
2. After $this->syncOriginal(); 

In the first case, when converting to an object, we will receive the same object passed by reference in the $attributes and $original arrays. And when the object changes, it will not be processed if only the changed attributes need to be retrieved. 
```php 
$a = ['a' => new stdClass];
$b = $a;
$b['a']->a = 1;
var_dump($a === $b); // true
```
In the second case, the attribute will be in the "changed" status all the time, even if it has not changed.
Moreover, the toRawArray (true, true) method does not work correctly.

Do you have a solution for this?

system/Entity.php
najdanovicivan: There are situations where it might be useful. There is no way to submit null via form. So in most cases I have custom setters which will set the empty string got via post to null. 

Having nullable as parameter in custom setter object will make it possible to completely get rid of such setters as it will be possible to make custom cast which will handle empty string to null conversion.

najdanovicivan:
We can modify hasChanged() method to iterate attributes check if value is of Entity class and if so it will forward the call to the inner Entity object 

I already did something very similar for to array method with introduction of the $recursive parameter. 

Then it makes more sense of doing conversion in `setAttributes()` instead of `syncOriginal()`

system/Entity.php
iRedds: How do you imagine that? I don't understand.
'field' => 'my_type[?param]'  Like this?

najdanovicivan: set $isNullable inside if (strpos($type, '?') === 0)

And then here after pregmatch append it to the params

if($isNullable)
{
  $params[] = ‘nullable’;
}

najdanovicivan:
I've given some thought into this and got an interesting idea. 

@iRedds what do you thing about extending the CastInterface with 2 additional methods

```
```

system/Entity.php
najdanovicivan: That's a good question. It can be done that way but it makes things very confusing as there will be 3 cases which allow null values.

`?bool`
`?bool[nullable]`
`bool[nullable]`

iRedds:
At the moment the question is whether this PR will be merged )

system/Entity.php
iRedds: OK. Commit with your suggestion added

system/Entity.php
```diff
@@ -527,69 +508,76 @@ protected function mutateDate($value)
...
-	 * @param string $type
+	 * @param mixed  $value     Attribute value
+	 * @param string $attribute Attribute name
+	 * @param string $method    Allowed to "get" and "set"
 	 *
 	 * @return mixed
 	 * @throws Exception
 	 */
-	protected function castAs($value, string $type)
+	protected function castAs($value, string $attribute, string $method)
```
paulbalandan, line 518: Add a default value to `$method` so it won't break user's implementations using this method.

system/Entity.php
```diff
@@ -527,69 +508,76 @@ protected function mutateDate($value)
...
-			case 'timestamp':
-				return strtotime($value);
+		//Attempt to retrieve additional parameters if specified
+		// type[param, param2,param3]
+		if (preg_match('/^(.+)\[(.+)\]$/', $type, $matches))
+		{
+			$type   = $matches[1];
+			$params = array_map(function ($param) {
+				return trim($param);
+			}, explode(',', $matches[2]));
```
paulbalandan, line 558:
```suggestion
			$params = array_map('trim', explode(',', $matches[2]));
```

system/EntityCast/AbstractCast.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+
+namespace CodeIgniter\EntityCast;
+
+/**
+ * Class AbstractCast
+ *
+ * @package CodeIgniter\EntityCast
```
paulbalandan, line 9: Remove the package tags

system/EntityCast/AbstractCast.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
```
paulbalandan, line 2: Add the license headers

system/EntityCast/CastAsArray.php
```diff
@@ -0,0 +1,35 @@
...
+ * @package CodeIgniter\Entity\Cast
+ */
+
+class CastAsArray extends AbstractCast
+{
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function get($value, array $params = []) : array
```
paulbalandan, line 18: Do not add space before the colon.
```suggestion
	public static function get($value, array $params = []): array
```

system/EntityCast/CastAsDouble.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace CodeIgniter\EntityCast;
+
+/**
+ * Class CastAsDouble
+ *
+ * @package CodeIgniter\Entity\Cast
+ */
+
+class CastAsDouble extends AbstractCast
```
paulbalandan, line 12: This should be the same with `CastAsFloat` so no longer necessary.

system/EntityCast/CastAsJson.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+
+namespace CodeIgniter\EntityCast;
+
+use stdClass;
+use CodeIgniter\Exceptions\CastException;
```
paulbalandan, line 7: Alphabetically order these two.

system/EntityCast/CastAsJson.php
```diff
@@ -0,0 +1,58 @@
...
+		$tmp = ! is_null($value) ? (in_array('array', $params, true) ? [] : new stdClass) : null;
+		if (function_exists('json_decode')
+			&& ((is_string($value)
+					&& strlen($value) > 1
+					&& in_array($value[0], ['[', '{', '"'], true))
+				|| is_numeric($value)
+			)
+		)
+		{
+			$tmp = json_decode($value, (bool) $params);
```
paulbalandan, line 31: You can use here the `JSON_THROW_ON_ERROR` flag and catch any `JsonException`

system/EntityCast/CastAsJson.php
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function set($value, array $params = []) : string
+	{
+		if (function_exists('json_encode'))
+		{
+			$value = json_encode($value, JSON_UNESCAPED_UNICODE);
```
paulbalandan, line 48: Use `JSON_THROW_ON_ERROR`

system/EntityCast/CastAsTimestamp.php
```diff
@@ -0,0 +1,22 @@
...
+
+class CastAsTimestamp extends AbstractCast
+{
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function get($value, array $params = [])
+	{
+		return strtotime($value);
```
paulbalandan, line 20: Maybe you can additionally check the return value if not `false`. If `false`, then throw an error for invalid timestamp.

user_guide_src/source/models/entities.rst
```diff
@@ -418,8 +418,112 @@ Stored in the database as "red,yellow,green"::
...
+            //   [1]=>
+            //   string(6) "param2"
+            //   [2]=>
+            //   string(6) "param3"
+            // }
+        }
+    }
+
+.. note:: If the casting type is marked as nullable ``?bool`` and the passed value is not null, then the parameter with
+the value ``nullable`` will be passed to the casting type handler.
```
paulbalandan, line 521: Indent

iRedds: I try to follow the PSR-12 style, for example 120 characters per line.

paulbalandan: Ah. I mean the `the value nullable...`. This causes parsing errors in sphinx.

iRedds:
* License comment added to all files in PR .
* In the Entity :: castAs () method, the $ method parameter is set to its default value. 
* Removed @package in docBlock.
* JSON casting now throws an exception instead of json_last_error(). 
* Timestamp casting now throws an exception if the value cannot be converted to a number.
* Double casting is now an alias for float.

user_guide_src/source/models/entities.rst
```diff
@@ -418,8 +418,111 @@ Stored in the database as "red,yellow,green"::
...
+            //   [1]=>
+            //   string(6) "param2"
+            //   [2]=>
+            //   string(6) "param3"
+            // }
+        }
+    }
+
+.. note:: If the casting type is marked as nullable ``?bool`` and the passed value is not null, then the parameter with the value ``nullable`` will be passed to the casting type handler.
+If casting type has predefined parameters, then ``nullable`` will be added to the end of the list.
```
paulbalandan, line 521: This needs to be also indented. Sphinx warns `Explicit markup ends without a blank line; unexpected unindent.`. Align this with `note` above.

system/Exceptions/CastException.php
```diff
@@ -43,4 +43,20 @@ public static function forInvalidJsonFormatException(int $error)
...
+
+	public static function invalidCastMethod() : self
+	{
+		return new static(lang('Cast.wrongCastMethod'));
+	}
+
+	public static function invalidTimestamp() : self
+	{
+		return new static(lang('Cast.invalidTimestamp'));
+	}
```
paulbalandan, line 60: Framework's way of declaring new static methods is `for` + the statement. So in this case it would be `forInvalidTimestamp` etc,

user_guide_src/source/models/entities.rst
iRedds: I'm sorry. I was stupid.

system/Exceptions/CastException.php
paulbalandan: I'm sorry I wasn't clear enough. What I meant is that all new methods would begin with `for`, not only `invalidTimestamp`.

iRedds: Fixed

paulbalandan (approved):
Last thing, please add an entry to the changelog in `user_guide_src`.

najdanovicivan (approved):

michalsn (approved):

paulbalandan:
Thank you, @iRedds

iRedds:
Thank you all for the help with PR

PR #4312, opened 2021-02-20
Enforce errors format

MGatner:
**Description**
There is currently a bug where `BaseModel::errors()` returns a `string` error message despite specifying `array|null` as the return type. This is causing runtime errors in `Fabricator` when trying to `implode()` error values (see discussion at #4272).

This PR traces errors back through their two potential sources: `Validation::errors()` and `ConnectionInterface::error()`. These return very different formats (which I have tightened in their docblocks) which `Model` should be handling.

Due to this bug and the preexisting formats from `Validation` and `Connection` I am updating the docblock for `BaseModel::errors` and `Model::errors` no longer to allow `null` because a) null values are never inherited from their sources, they were only set in the model, and b) they introduce an unnecessary extra type check from any calling method. This is already consistent with the User Guide and the function definition but I noted the docblock update in the changelings.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1087,27 +1087,25 @@ abstract class BaseModel
 	 * Grabs the last error(s) that occurred. If data was validated,
 	 * it will first check for errors there, otherwise will try to
 	 * grab the last error from the Database connection.
+	 * The return array should be in the following format:
+	 *  ['source' => 'message']
 	 *
 	 * @param boolean $forceDB Always grab the db error, not validation
 	 *
-	 * @return array|null
+	 * @return array<string,string>
 	 */
 	public function errors(bool $forceDB = false)
 	{
 		// Do we have validation errors?
 		if (! $forceDB && ! $this->skipValidation)
 		{
-			$errors = $this->validation->getErrors();
-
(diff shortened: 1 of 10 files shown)


MGatner:
@najdanovicivan Could you please take a look at the `Model`/`BaseModel` changes?

@natanfelles Can you please verify this fixes your issue?

Review from anyone else appreciated too.

MGatner:
@samsonasik It would be handy to get Rector results annotated inline - do you know if they support that yet? E.g. Infection's [--logger-github](https://infection.github.io/guide/usage.html#Configuration-settings)

samsonasik:
@MGatner it seems not yet as far as I know.

najdanovicivan:
This is all good. I see that I have nasty patch for this in my controller

```
! is_array($errors) && $errors = [$errors];
```

This should make those refundable. I like it.

samsonasik (approved):

paulbalandan (approved):

MGatner:
Thanks all!

PR #4324, opened 2021-02-22
[Rector] Apply Rector: PreparedValueToEarlyReturnRector

samsonasik:
Return early prepared value in ifs with `PreparedValueToEarlyReturnRector` rector rule.

**Checklist:**
- [x] Securely signed commits


MGatner (approved):
Very nice! This one is a bit more complex but a definite clarity improvement.

PR #4325, opened 2021-02-22
Bugfix empty Model errors

MGatner:
**Description**
While standardizing error formats (#4312) test and I both failed to account for the fact that `ConnectionInterface::errors()` returns content even when there is no error. This extends the work from that PR to define a "no error" value. This is a little awkward because the "no error" value of `code` for each driver varies:
* MySQL: `0`
* SQLite3: `0`
* Postgres: `''`
* SQLSRV: `'00000'`

I settled on `interval(code) === 0` which is consistent with our current drivers and hopefully works with any third-party extensions.

FYI this was yet another bug in the code previous to #4312 because `BaseModel::errors()` would return `null` only when `message` was empty (`return $error['message'] ?? null;`) which was not always the case (e.g. SQLite3 returns `'not an error`').

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -471,7 +471,13 @@ class Model extends BaseModel
 	 */
 	protected function doErrors()
 	{
+		// $error is always ['code' => string|int, 'message' => string]
 		$error = $this->db->error();
+
+		if (intval($error['code']) === 0)
+		{
+			return [];
+		}
 
 		return [get_class($this->db) => $error['message']];
 	}
--- a/system/Model.php
+++ b/system/Model.php
@@ -474,7 +474,7 @@ class Model extends BaseModel
 		// $error is always ['code' => string|int, 'message' => string]
 		$error = $this->db->error();
 
-		if (intval($error['code']) === 0)
+		if ((int) $error['code'] === 0)
 		{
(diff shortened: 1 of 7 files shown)


system/Model.php
```diff
@@ -471,8 +471,14 @@ protected function doReplace(array $data = null, bool $returnSQL = false)
 	 */
 	protected function doErrors()
 	{
+		// $error is always ['code' => string|int, 'message' => string]
 		$error = $this->db->error();
 
+		if (intval($error['code']) === 0)
```
paulbalandan, line 477:
```suggestion
		if ((int) $error['code'] === 0)
```

MGatner: I went with `intval()` because PHP 8 changed a lot of string conversions and comparisons. Why do you recommend the cast over the function?

paulbalandan: My argument is because casting is faster. For php 8 changes, as I understand it php 8 changes behavior on loose checking (`==`) only. Others are unaffected. https://3v4l.org/0EG4N

paulbalandan (approved):

PR #4327, opened 2021-02-23
[Rector] Apply Rector: RenameParamToMatchTypeRector

samsonasik:
Rename parameter variable to match typed parameter in methods with `RenameParamToMatchTypeRector` rector rule.

**Checklist:**
- [x] Securely signed commits


paulbalandan:
This will break in php8 with regard to named parameter usage. https://3v4l.org/qnTGX

samsonasik:
Ok 👍 , closing then

PR #4335, opened 2021-02-24
Refactor CIDatabaseTestCase to use a trait instead.

lonnieezell:
- Refactored all database refresh/seeding/etc code from `CIDatabaseTestCase` to a new `DatabaseTestTrait`.
- Moved database assertions to main `CIUnitTestCase``


MGatner (approved):

MGatner:
Actually in retrospect the trait should not include the properties because people will want to overwrite those in their test case. I will make a new PR to fix this.

lonnieezell:
They can still override them once the trait is include. Been a while but pretty sure that works normally.

MGatner:
You can override them but not in the same class with the trait. So for example this is not allowed:
```
class MyProjectTestCase extends CIUnitTestCase
{
	use DatabaseTestTrait;

	protected $seed = MySeeder::class;
}
```

... because it will be a conflict with the property trait. I have a fix for this almost ready, I will request your review.

Issue #4339, opened 2021-02-24
Implement default fields escaping configuration in Model


reinaldorauch:
Edit: I thought that it was an undocumented feature to define the default for controlling the escaping of the fields in the model but now understanding more about the code i that it was not the case. So im proposing this to implement this feature to set the defaults  for escaping the fields.

An example of a situation that this is desired if you have a table

```SQL
CREATE TABLE a_test (
  id INT UNSIGNED AUTO_INCREMENT,
  bit_value BIT(2) NOT NULL
) engine = innodb;
```

If you escape the value going to bit_value, lets say it must be 0, it will become the string '0', that, in MySQL it converts to the binary value of the '0' that is not b'0' and therefore borking the insert.

If i could define the default for the escape in the model as below:

```php
<?php
// ...
class TestModel extends Model {
    // ...
    protected $escape = [
        'bit_value' => false
    ];
    // ...
}
```

It would help me to use the default save, insert and update methods defined in Model.

I opened the #4340 PR to implement this issue. 

Any thoughts on this will be welcome

kenjis:
It seems nobody needs this.
Close.

We use GitHub issues to track BUGS and to track approved DEVELOPMENT work packages. We use our [forum](http://forum.codeigniter.com/) to provide SUPPORT and to discuss FEATURE REQUESTS.

PR #4340, opened 2021-02-24
[WIP] Implement escape defaults in the Model

reinaldorauch:
Currently, if you have in your model defined the $escape property, it will be ignored when doing a insert or update operation (see [this](https://github.com/codeigniter4/CodeIgniter4/issues/4339) issue).

I'm attempting to fix this by moving those statements inside the conditional above because they don't seem to make sense outsite thus ocasioning the issue.

**Checklist:**
- [X] Securely signed commits
- [X] Component(s) with PHPdocs (do not apply)
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -693,10 +693,11 @@ class Model extends BaseModel
 				$data = $this->transformDataToArray($data, 'insert');
 				$data = array_merge($this->tempData['data'], $data);
 			}
+
+			$this->escape   = $this->tempData['escape'] ?? [];
+			$this->tempData = [];
 		}
 
-		$this->escape   = $this->tempData['escape'] ?? [];
-		$this->tempData = [];
 
 		return parent::insert($data, $returnID);
 	}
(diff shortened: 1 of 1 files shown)


MGatner:
I don't totally understand how this fixes the issue you reported, but initially it makes sense to me that we only escape data within the conditional. Could you come up with a test case that would have failed prior to your PR but is now fixed? Probably one for both insert and update are in order.

MGatner:
Actually looking at it again I think we need to add the escape regardless of temp data. Regardless I think the test cases will elucidate this for me.

michalsn:
The `$escape` property in the Model is used only internally to determine if the value should be escaped or not. Please see the test example here: https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Models/InsertModelTest.php#L264

To set this value you have to use `set()` method with your model. It works the same as in the Builder class.

reinaldorauch:
Sorry,  I didn't updated this PR with my recent findings. Only after analyzing better the code that i realized what @michalsn pointed.

I opened an [issue](https://github.com/codeigniter4/CodeIgniter4/issues/4339) before to explain the issue but i've updated that to suggest a feature to pre-configure the escaped properties of the model.

I will update this PR to implement this feature

paulbalandan:
2 months have passed and no updates to this PR. @reinaldorauch is this PR a WIP or are you still interested tackling this?

reinaldorauch:
@paulbalandan Sorry, i have no more interest in pushing this forward. I will close it for now

PR #4350, opened 2021-02-25
Fix setBinds of BaseBuilder

### LINKED ISSUE #4345: Bug: insertBatch generates an incorrect SQL query if the fields differ only in number at the end
Example:
```sql
CREATE TABLE `ip_list` (`ip` VARCHAR(100),`ip2` VARCHAR(100))
```
Code:
```php
$data = [
  [ 'ip' => '1.1.1.0', 'ip2' => '1.1.1.2' ],
  [ 'ip' => '2.2.2.0', 'ip2' => '2.2.2.2' ],
  [ 'ip' => '3.3.3.0', 'ip2' => '3.3.3.2' ],
  [ 'ip' => '4.4.4.0', 'ip2' => '4.4.4.2' ]
];
$db->table('ip_list')->insertBatch($data);
```
Generating incorrect SQL query:
```sql
INSERT INTO `ip_list` (`ip`,`ip2`) VALUES ('1.1.1.0','4.4.4.0'), ('2.2.2.0','2.2.2.2'), ('3.3.3.0','3.3.3.2'), ('4.4.4.0','4.4.4.2')
```
Seems error in setBind `$this->binds[$key . $count] ...`

**Context**
 - CodeIgniter: 4.0.4
 - PHP: 7.2.24

paulbalandan:
**Description**
Fixes #4345.

This PR changes how `BaseBuilder::setBind()` stores the named binds in `$this->binds` and the counting of `$this->bindsKeyCount`.

- **Changes to `$this->binds`**

Currently, using the bug report's table structure, table fields are incremented in the binds property this way:

Assuming `ip` and `ip2` fields in an insert batch of size `4`:
```
Iteration         |      Named Bind for `ip`        |          Named Bind for `ip2`    |
---------------------------------------------------------------------------------
     1                               `ip`                                     `ip2`
     2                              `ip0`                                    `ip20`
     3                              `ip1`                                    `ip21`
     4                              `ip2`                                    `ip22`
```

You'll see that on the 4th iteration, the named bind of `ip` is `ip2` which is the same with the named bind of `ip2` on its 1st iteration. Since this is saved as an array, PHP will just silently replace the value of the original `ip2` key, causing the bug reported.

This PR proposes to use a dot array notation in saving binds. Such as the first named bind is ~`ip.0`~ `ip` and the next named bind is `ip.1`. In essence, the new binds will be:
```
1 => ['ip', 'ip2'],
2 => ['ip.1, 'ip2.1'],
3 => ['ip.2', 'ip2.2'],
4 => ['ip.3', 'ip2.3'],
```

- **Changes to `$this->bindsKeyCount`**

This change is minimal but I think still deserves to be noticed. `$this->bindsKeyCount` is described as `Collects the key count for named parameters in the Query object.`. Since this is a count, we should start counting at 1 not 0. I am not aware if this is used elsewhere outside `$this->setBind` so please advise if this will break something.

~PS. I'll fix later the failing tests which uses the contents of the `$binds` property.~

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


paulbalandan:
The Postgre failure seems to be a bug.

michalsn (commented):
Honestly, I would stay with the current naming convention and just add a dot before the index. It will cause fewer problems if someone (for some reason) is manipulating the string obtained from `getBinds()`.

Most of the time our queries don't run in "batch" mode, so adding `0` index to every binding seems a bit unnatural to me.

paulbalandan:
Returned first key to original key name instead of `key.0`. For the succeeding binds, it would be `key.1`, `key.2`, ...

michalsn (approved):
Thank you for working on this one.

lonnieezell:
I'm not sure if the issue existed before this, but I have some code that works on 4.1.x and doesn't on the latest dev-develop anymore and seems to be tied to setBinds. I haven't had a chance to dig into yet but this seems to be one of the only PRs to deal with setBinds after 4.1.... Here's some code to replicate it if you have a chance to look into before I do: 

```
public function testWhereWithLower()
	{
		$this->db->table('job')
		         ->insert([
			         'name'        => 'Brewmaster',
			         'description' => null,
		         ]);

		$jobs = $this->db->table('job')
		                 ->where('LOWER(job.name)', 'brewmaster')
		                 ->get()
		                 ->getResult();

		$this->assertCount(1, $jobs);
	}
```

PR #4352, opened 2021-02-25
Entity - FilterVar Casting in magic setter

najdanovicivan:
**Description**
I've noticed that only csv, array and json are casted on magic setter. This one adds additional types in those casts. The most needed here is cast for bool. Especially when ElasticSearch is used as database for which any updates with int as bool value will fail.

Additionally some unit tests should be added

- Use switch instead of multiple if calls
- Added boolean, integer, float, double and string casts

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -359,31 +359,49 @@ class Entity implements JsonSerializable
 
 		if (! $isNullable || ! is_null($value))
 		{
-			// CSV casts need to be imploded.
-			if ($castTo === 'csv')
+			switch ($castTo)
 			{
-				$value = implode(',', $value);
-			}
-
-			// Array casting requires that we serialize the value
-			// when setting it so that it can easily be stored
-			// back to the database.
-			if ($castTo === 'array')
-			{
-				$value = serialize($value);
-			}
-
-			// JSON casting requires that we JSONize the value
-			// when setting it so that it can easily be stored
-			// back to the database.
-			if (($castTo === 'json' || $castTo === 'json-array') && function_exists('json_encode'))
-			{
-				$value = json_encode($value, JSON_UNESCAPED_UNICODE);
-
-				if (json_last_error() !== JSON_ERROR_NONE)
-				{
-					throw CastException::forInvalidJsonFormatException(json_last_error());
-				}
+				case 'bool':
+				case 'boolean':
+					$value = filter_var($value, FILTER_VALIDATE_BOOLEAN, $isNullable ? FILTER_NULL_ON_FAILURE : 0);
+					break;
+				case 'int':
+				case 'integer':
+					$value = (int) $value;
+					break;
+				case 'float':
+					$value = (float)$value;
+					break;
+				case 'double':
+					$value = (double)$value;
+					break;
+				case 'string':
+					$value = (string)$value;
+					break;
+				// CSV casts need to be imploded.
+				case 'csv':
+					$value = implode(',', $value);
+					break;
+				// Array casting requires that we serialize the value
+				// when setting it so that it can easily be stored
+				// back to the database.
+				case 'array' :
+					$value = serialize($value);
+					break;
+				// JSON casting requires that we JSONize the value
+				// when setting it so that it can easily be stored
+				// back to the database.
+				case 'json':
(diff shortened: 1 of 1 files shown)


najdanovicivan:
What I noticed about casts is that only csv,json,json-array and array work on setter. This is a bit confusing so I’ve tried adding other types as well. But I see now with test failing that it might be impossible to implement it that way.

The problematic test `testCastSetter()` set the entity value as string then tries reading it with both casting on and off.  

As I'm looking into this I see it might not be possible to do conversions in setter using casts. But at leas we should have boolean values converted to avoid the mentioned issues with some of the NoSQL databases.

iRedds:
Take a look at my implementation. https://github.com/codeigniter4/CodeIgniter4/pull/4308 
This does not change the default casting behavior, but it adds custom handlers for the casting.

najdanovicivan:
@iRedds I like it. I've put some comments in #4308

system/Entity.php
```diff
@@ -359,31 +359,49 @@ public function __set(string $key, $value = null)
...
+				case 'int':
+				case 'integer':
+					$value = (int) $value;
+					break;
+				case 'float':
+					$value = (float)$value;
+					break;
+				case 'double':
+					$value = (double)$value;
+					break;
```
MGatner, line 377: Float and double are the same in PHP
```suggestion
				case 'float':
				case 'double':
					$value = (float) $value;
					break;
```

system/Entity.php
```diff
@@ -359,31 +359,49 @@ public function __set(string $key, $value = null)
...
+				case 'string':
+					$value = (string)$value;
```
MGatner, line 379: I'm pretty sure our style guide requires a space after casts.
```suggestion
					$value = (string) $value;
```

system/Entity.php
```diff
@@ -359,31 +359,49 @@ public function __set(string $key, $value = null)
...
-			{
-				$value = json_encode($value, JSON_UNESCAPED_UNICODE);
-
-				if (json_last_error() !== JSON_ERROR_NONE)
-				{
-					throw CastException::forInvalidJsonFormatException(json_last_error());
-				}
+				case 'bool':
+				case 'boolean':
+					$value = filter_var($value, FILTER_VALIDATE_BOOLEAN, $isNullable ? FILTER_NULL_ON_FAILURE : 0);
```
MGatner, line 366: I don't understand this one. If `__set()` casting is supposed to "mirror" `castAs()` why are we converting back to boolean? Is this to prevent `null` when a cast is not nullable?

It also seems like we are introducing some additional values, "on off yes no":

>Returns true for "1", "true", "on" and "yes". Returns false otherwise.
> If FILTER_NULL_ON_FAILURE is set, false is returned only for "0", "false", "off", "no", and "", and null is returned for all non-boolean values.

MGatner (changes requested):
What about casts for `object`, `datetime`, and `timestamp`? It seems like this should be comprehensive to `castAs()`. Also we will definitely need some tests demonstrating these new casts.

system/Entity.php
najdanovicivan: The problem with not having setter is that it will allow setting any sort of string as value for bool property. Trying to write such strings in many NoSQL databases will result in error

The problem with current casts functionality it that there is no way to limit what is allowed to be written in attributes array.

And the bool is just example how can it produce sorts of issues. Currently I have a custom setter for each bool value in my entities where I'm doing those conversions. 

And the problem gets even more annoying when there's a datamap in play as those setters will have completely different name

system/Entity.php
najdanovicivan: Yes they are. But there is the same logic in castAs method so I wanted to follow it here

system/Entity.php
najdanovicivan: I like what's suggested in #4308 

It will make this PR completely redundant. And it will give a lot of possibilities with casting so I'd rather #4308 to be merged than this one.

najdanovicivan:
This one is not needed anymore

Duplicate of #4308

Issue #4360, opened 2021-02-27
Bug: Entity::toRawArray incorrect work. 


iRedds:
I believe that the behavior of Entity::toRawArray is not correct and not fully tested. 
For example, I'll take a test 
https://github.com/codeigniter4/CodeIgniter4/blob/4cbfe2cd25ff8b08eee323cc734a2113cead8eb1/tests/system/EntityTest.php#L773-L792

If the toRawArray() method is called with arguments $onlyChanged = true and $recursive = true 
Then the result of execution will be 
```php
[
    'entity' => []
]
```
instead of
```php
[
    'entity' => [
 	'foo'        => null, 
 	'bar'        => null, 
 	'default'    => 'sumfin', 
 	'created_at' => null, 
   ]
]
```

najdanovicivan:
I can confirm this is bug. When I initially added the $recursive parameter I did not take into consideration that hasChanged method have to be updated as well.

We need deffiniton on how this is supposed to work.

If there is a single change inside entity object.  ToRawArray() method on the inner entity might have to be called with $onlyChanged = false

Most of NoSQL are going to replace the complete value with the new object instead of updating just modified values. There are means of updating the nested values but those differ a lot per DB.

So I'd go with toRawArray doing replace of complete inner object. If there is a need for implementing the update on the inner model those can always be handled in the Model by doing non recursive toRawArray() and manually handling the inner entities.

I can go ahead and start working on fixing this one.

lonnieezell:
Very stale. Closing.

PR #4368, opened 2021-02-28
WIP: Entity - Properly handle only changed in recursive toArray()/toRawArray()

najdanovicivan:
**Description**
Related to #4360 pointed out by @iRedds 

When recursive parameter was introduced in those methods the intended purpose was to be able to generate json to be returned in for example Ajax responses.

It was not planned that at some point those will be used to write data in the db as it is impossible to write arrays directly in SQL. But now with the introduction on BaseModel we have means to write entities in NoSQL database as nested objects.

The issue with NoSQL and nested object is that methods for updating the nested objects are very different for each database so the only feasible solution is not to update the inner values but instead do complete replace of the nested object in the root document. 

So the use of recursive option in those methods should always return the complete data array even if only single attribute of the nested object is changed. 

Models which are going to implement methods for partially updating the nested objects should be developed using non recursive version of such methods and handling the inner object updates manually inside the model logic.

For the time being this ads support for properly handling nested Entity objects in hasChanges() methods and converts single objects to arrays in both toArray() and toRawArray() methods.

But in the meantime I noticed another issue with the recursive parameter. In the current state it's not possible to do a recursive conversion on attributes which are arrays of entity objects and such cases have to be handled properly. So the work on this continues. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -151,9 +151,10 @@ class Entity implements JsonSerializable
 
 			if ($recursive)
 			{
+				//@todo: handle arrays of entities
 				if ($return[$key] instanceof Entity)
 				{
-					$return[$key] = $return[$key]->toArray($onlyChanged, $cast, $recursive);
+					$return[$key] = $return[$key]->toArray(false, $cast, $recursive);
 				}
 				elseif (is_callable([$return[$key], 'toArray']))
 				{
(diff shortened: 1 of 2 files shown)


paulbalandan:
What is the status of this?

najdanovicivan:
This is still WIP. This will currently work for single entity attributes. But it is still needed to handle changed properly for attributes which contain arrays of entity objects.

lonnieezell:
Hasn't been actively worked on in almost a year. Closing. We would still love to see a fix for that bug, but this one isn't ready.

PR #4370, opened 2021-03-01
Add ControllerFactory

kenjis:
**Description**
Add `ControllerFactory` to override the instantiation of the controller.

This PR enables users to use dependency injection (container) when instantiation of a controller.

See <https://forum.codeigniter.com/thread-78699.html>.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -907,8 +907,9 @@ class CodeIgniter
 	 */
 	protected function createController()
 	{
-		$class = new $this->controller();
-		$class->initController($this->request, $this->response, Services::logger());
+		$controllerFactory = Services::controllerfactory($this->request, $this->response, Services::logger());
+
+		$class = $controllerFactory->create($this->controller);
 
 		$this->benchmark->stop('controller_constructor');
 
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -13,6 +13,7 @@ namespace CodeIgniter\Config;
 
 use CodeIgniter\Autoloader\Autoloader;
 use CodeIgniter\Autoloader\FileLocator;
+use CodeIgniter\ControllerFactory;
 use Config\Autoload;
 use Config\Modules;
 use CodeIgniter\Cache\CacheInterface;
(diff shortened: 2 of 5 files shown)


system/CodeIgniter.php
```diff
@@ -907,8 +907,9 @@ protected function startController()
 	 */
 	protected function createController()
 	{
-		$class = new $this->controller();
-		$class->initController($this->request, $this->response, Services::logger());
+		$controllerFactory = Services::controllerfactory($this->request, $this->response, Services::logger());
```
paulbalandan, line 910: can you make this as camelCase?

kenjis: Do you mean `Services::controllerFactory()`?

paulbalandan: Ah nevermind. I forgot services are checked using `strtolower`

lonnieezell:
I'll be honest - I'm struggling with the need for this one. You've replaced what is accomplished in 2 lines with a new service and a new class. Only a single controller is ever loaded at a time, so this doesn't do anything for caching. It doesn't help performance and likely hinders it a touch. I cannot come up with a compelling reason to make this change. it just seems to add complexity for complexity's sake. 

Closing for now but feel free to make you case.

lonnieezell:
Reading back over the referenced thread I get it. This is an attempt to make it more like what people are used to. We purposefully chose to not use a DI container. See the comment [here](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Config/BaseService.php#L60) and give Maxwell's blog a read to understand why.

kenjis:
@lonnieezell 

> We purposefully chose to not use a DI container.

This PR never introduces a DI container to CI4.
This provides a way to override a controller creation.

What should a user do if he/she wants to inject dependencies into a controller?

kenjis:
@lonnieezell I came up with another way. #4376

lonnieezell:
> This PR never introduces a DI container to CI4.
> This provides a way to override a controller creation.

I understand that. But it's adding an additional layer of complexity to make it possible to support one person's desire for something we already made a decision not to support. 

Your new solution has more promise. I'll look it over in more detail later.

MGatner:
I guess this will accomplish the desired result, but I was thinking more along the lines of `Config\Services::controller()` as a way for this particular user (and anyone) to supply their own controller class.

PR #4376, opened 2021-03-01
Add Services::codeigniter()

kenjis:
**Description**
In the current implementation, there is no way to extend the class `CodeIgniter`.
This PR enables users to extend `CodeIgniter`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -13,6 +13,7 @@ namespace CodeIgniter\Config;
 
 use CodeIgniter\Autoloader\Autoloader;
 use CodeIgniter\Autoloader\FileLocator;
+use CodeIgniter\CodeIgniter;
 use Config\Autoload;
 use Config\Modules;
 use CodeIgniter\Cache\CacheInterface;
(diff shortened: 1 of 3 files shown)


lonnieezell:
While I think the reason this PR was created is unneeded, I have no problems with people extending this class on their own. As long as they don't come back and complain things don't work after they do lol.

Maybe there should be a comment "Extending this class voids your warranty?" :) 

Just kidding. I'm fine with this. Merging.

kenjis:
@lonnieezell Thanks for merging!
Yes, the reason this PR was created is just one use case for this.
This is more flexible than my thought.

The problem that users who want to use DI container for their application 
must modify the framework code has been the one of the biggest problem for me since CI3 was released. Now the problem was solved.

Personally I respect your decision that CI4 does not use DI container.
And I like the bonus side-effect.

>  A bonus side-effect is that IDEs are able to determine what class you are calling whereas with DI Containers there usually isn't a way for them to do this.

So I do not like the fact that this side effect is no longer present in the current Services.

PR #4379, opened 2021-03-02
Bugfix invalid HTTP codes

MGatner:
**Description**
Fixes a bug where `ControllerTester` would set any Exception code as an HTTP code which could cause a lot of debugging confusion:

> CodeIgniter\HTTP\Exceptions\HTTPException: 0 is not a valid HTTP return status code

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/ControllerTester.php
+++ b/system/Test/ControllerTester.php
@@ -164,9 +164,11 @@ trait ControllerTester
 		}
 		catch (Throwable $e)
 		{
-			if ($code >= 100 || $code < 600)
+			$code = $e->getCode();
+
+			if ($code >= 100 && $code < 600)
 			{
-				$result->response()->setStatusCode($e->getCode());
+				$result->response()->setStatusCode($code);
 			}
 		}
 		finally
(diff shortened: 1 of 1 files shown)


lonnieezell (approved):
This looks fine to me once tests pass.

system/Test/ControllerTester.php
```diff
@@ -164,7 +164,10 @@ public function execute(string $method, ...$params)
 		}
 		catch (Throwable $e)
 		{
-			$result->response()->setStatusCode($e->getCode());
+			if ($code >= 100 || $code < 600)
+			{
+				$result->response()->setStatusCode($e->getCode());
+			}
```
paulbalandan, line 170:
```suggestion
			$code = $e->getCode();

			if ($code >= 100 || $code < 600)
			{
				$result->response()->setStatusCode($code);
			}
```

MGatner: I realized this mistake lying in bed last night and thought to myself "I wonder if anyone will notice before I can fix it?" 😋

I pasted this from ResponseTrait. The OR needs to be AND as well. I had a computer science professor who insisted you should never paste code, maybe she was right...

paulbalandan (approved):

PR #4384, opened 2021-03-02
Fix errant regex not considering new lines

### LINKED ISSUE #4367: ErrorException preg_replace_callback(): Unknown modifier '{' SYSTEMPATH/View/Parser.php at line 584
image

if the braces are preceded by a hash or other characters or are also in the middle of () the parser does not just go into the loop outside the loop same situation everything works.
Are there any solutions?

paulbalandan:
**Description**
Fixes #4367 

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -574,8 +574,8 @@ class Parser extends View
 		$pattern = addcslashes($pattern, '$');
 
 		// Flesh out the main pattern from the delimiters and escape the hash
-		// See https://regex101.com/r/1GIHTa/1
-		if (preg_match('/^(#)(.*)(#(m?s)?)$/', $pattern, $parts))
+		// See https://regex101.com/r/IKdUlk/1
+		if (preg_match('/^(#)(.+)(#(m?s)?)$/s', $pattern, $parts))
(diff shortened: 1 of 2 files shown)


MGatner (approved):

system/View/Parser.php
```diff
@@ -574,8 +574,8 @@ protected function replaceSingle($pattern, $content, $template, bool $escape = f
 		$pattern = addcslashes($pattern, '$');
 
 		// Flesh out the main pattern from the delimiters and escape the hash
-		// See https://regex101.com/r/1GIHTa/1
-		if (preg_match('/^(#)(.*)(#(m?s)?)$/', $pattern, $parts))
+		// See https://regex101.com/r/IKdUlk/1
```
samsonasik, line 577: @paulbalandan to update regex101 version, you can use "Update Regex" button instead of "Save Regex" so the new url generated will be increased number /2 /3, etc, eg: https://regex101.com/r/1GIHTa/2

ref https://twitter.com/samsonasik/status/1355829852138430468

PR #4390, opened 2021-03-05
Fix: filter_var usage

iRedds:
**Description**
When using the flag parameter in the filter_var function, the default value was incorrectly passed. 
By default, null is passed, but an array or integer is expected.
Soft type casting may mask future problems. 

Changes:
* Updated variable types in docBlock 
* Added nullable type for filter parameter 
* Defined a default value for the filter and flag arguments if passed null.
* The filter and flag arguments are cast to the expected types. 
* Adjusted the data validation condition based on the changes above. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -356,7 +356,7 @@ class IncomingRequest extends Request
 
 		if (! is_array($data))
 		{
-			$filter = is_numeric($filter) ? (int) $filter : FILTER_DEFAULT;
+			$filter = $filter ?? FILTER_DEFAULT;
 			$flags  = is_array($flags) ? $flags : (is_numeric($flags) ? (int) $flags : 0);
 
 			return filter_var($data, $filter, $flags);
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -342,13 +342,13 @@ class IncomingRequest extends Request
 	/**
 	 * Get a specific variable from a JSON input stream
 	 *
-	 * @param  string       $index  The variable that you want which can use dot syntax for getting specific values.
-	 * @param  boolean      $assoc  If true, return the result as an associative array.
-	 * @param  integer|null $filter Filter Constant
-	 * @param  mixed        $flags
+	 * @param  string             $index  The variable that you want which can use dot syntax for getting specific values.
+	 * @param  boolean            $assoc  If true, return the result as an associative array.
+	 * @param  integer|null       $filter Filter Constant
+	 * @param  array|integer|null $flags  Option
 	 * @return mixed
 	 */
-	public function getJsonVar(string $index, bool $assoc = false, $filter = null, $flags = null)
+	public function getJsonVar(string $index, bool $assoc = false, ?int $filter = null, $flags = null)
 	{
 		helper('array');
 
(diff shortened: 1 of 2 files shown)


system/HTTP/IncomingRequest.php
```diff
@@ -342,21 +342,23 @@ public function getJSON(bool $assoc = false, int $depth = 512, int $options = 0)
...
-	 * @param  integer|null $filter Filter Constant
-	 * @param  mixed        $flags
+	 * @param  string             $index  The variable that you want which can use dot syntax for getting specific values.
+	 * @param  boolean            $assoc  If true, return the result as an associative array.
+	 * @param  integer|null       $filter Filter Constant
+	 * @param  array|integer|null $flags  Option
 	 * @return mixed
 	 */
-	public function getJsonVar(string $index, bool $assoc = false, $filter = null, $flags = null)
+	public function getJsonVar(string $index, bool $assoc = false, ?int $filter = null, $flags = null)
```
mostafakhudair, line 351:
```suggestion
	public function getJsonVar(string $index, bool $assoc = false, ?int $filter, $flags = null)
```

system/HTTP/RequestTrait.php
```diff
@@ -240,14 +240,14 @@ public function setGlobal(string $method, $value)
...
-	 * @param mixed             $flags
+	 * @param string             $method Input filter constant
+	 * @param string|array|null  $index
+	 * @param integer|null       $filter Filter constant
+	 * @param array|integer|null $flags  Options
 	 *
 	 * @return mixed
 	 */
-	public function fetchGlobal($method, $index = null, $filter = null, $flags = null)
+	public function fetchGlobal(string $method, $index = null, ?int $filter = null, $flags = null)
```
mostafakhudair, line 250:
```suggestion
	public function fetchGlobal(string $method, $index = null, ?int $filter, $flags = null)
```

mostafakhudair (approved):

system/HTTP/IncomingRequest.php
iRedds: 1. A nullable type assumes NULL, but does not define a default.
2. Violating php 8 compatibility

system/HTTP/RequestTrait.php
iRedds: 1. A nullable type assumes NULL, but does not define a default.
2. Violating php 8 compatibility

paulbalandan:
Please rebase.

iRedds:
@paulbalandan  Looks like I did something wrong.

paulbalandan:
I'm not sure how to fix this but I guess you need to cherry pick your PR commit `adfa1ec`

iRedds:
I think I did it. 
Thx @totoprayogo1916

system/HTTP/IncomingRequest.php
```diff
@@ -342,21 +342,23 @@ public function getJSON(bool $assoc = false, int $depth = 512, int $options = 0)
...
 	{
 		helper('array');
 
 		$data = dot_array_search($index, $this->getJSON(true));
 
 		if (! is_array($data))
 		{
-			$filter = $filter ?? FILTER_DEFAULT;
+			$filter = is_numeric($filter) ? (int) $filter : FILTER_DEFAULT;
```
paulbalandan, line 359: should the typecasting be unnecessary here because you have already typehinted it already as `?int`, so the null coalesce should remain

system/HTTP/RequestTrait.php
```diff
@@ -257,10 +257,8 @@ public function fetchGlobal($method, $index = null, $filter = null, $flags = nul
 		}
 
 		// Null filters cause null values to return.
-		if (is_null($filter))
-		{
-			$filter = FILTER_DEFAULT;
-		}
+		$filter = is_numeric($filter) ? (int) $filter : FILTER_DEFAULT;
```
paulbalandan, line 260: same comment with above

system/HTTP/IncomingRequest.php
iRedds: Hm... I agree.

paulbalandan (approved):

PR #4392, opened 2021-03-06
Fix missing extensions in CI/CD

paulbalandan:
**Checklist:**
- [x] Securely signed commits


kenjis:
It seems there are still some fonts missing.

> convert: unable to read font `helvetica' @ error/annotate.c/RenderFreetype/1338.

paulbalandan:
I've reinstalled `ghostscript` but still imagick's `convert` still fails.

kenjis:
I'm not sure, but how about installing `gsfonts`?

paulbalandan:
> I'm not sure, but how about installing `gsfonts`?

Thanks!

PR #4401, opened 2021-03-10
[IncomingRequest] set return null on IncomingRequest::getOldInput() when session has not been started and session key not found

samsonasik:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -490,7 +490,7 @@ class IncomingRequest extends Request
 	 *
 	 * @param string $key
 	 *
-	 * @return mixed
+	 * @return mixed|null
 	 */
 	public function getOldInput(string $key)
 	{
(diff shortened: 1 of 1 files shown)


system/HTTP/IncomingRequest.php
```diff
@@ -490,15 +490,15 @@ public function getUserAgent()
 	 *
 	 * @param string $key
 	 *
-	 * @return mixed
+	 * @return mixed|null
```
paulbalandan, line 493: `mixed` already implies `null` included

paulbalandan (commented):
I think the current return type of `mixed|void` is a bug, so approving this.

system/HTTP/IncomingRequest.php
samsonasik: removed.

paulbalandan (approved):

samsonasik:

PR #4404, opened 2021-03-10
Helper form_multiselect $data param as array

eafarooqi:
**Description**
Helper function form_multiselect $data parameter type should be mixed as form_dropdown. 
So when passed array it should handle the array instead of error.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -313,14 +313,14 @@ if (! function_exists('form_multiselect'))
 	/**
 	 * Multi-select menu
 	 *
-	 * @param string $name
+	 * @param mixed  $name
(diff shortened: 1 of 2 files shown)


samsonasik:
could you add test case for it? it may be at after the following test:

https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Helpers/FormHelperTest.php#L540

or you can use data provider for single test name with multi data.

eafarooqi:
I have added a new test case

samsonasik (approved):

paulbalandan (approved):

PR #4419, opened 2021-03-12
Test Traits

MGatner:
**Description**
I've been enjoying the new `DatabaseTestTrait`, but it reveals how awkward the `$setUpMethods` and `$tearDownMethods` properties are for actual use. There is no easy way to extend them, traits cannot use them without conflict errors, adjusting them in the framework would be a breaking change...

This PR moves toward a new approach: dynamic trait staging methods. I've borrow some helper functions from Laravel to pull all the traits from a test case and then check for a corresponding method name. For example:
```
trait AuthTrait
{
	protected setUpAuthTrait()
	{
		$user = $this->createFakeUser();
		$this->logInUser($user);
	}
...

class AuthenticationFeatureTest
{
	use AuthTrait;
...
```

This allows each trait to define its own set up and tear down methods without the test case needing to know about them. If no methods are present then nothing happens.

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide


lonnieezell (approved):
I like it. Makes me wonder if we need to adjust the FeatureTestTrait at all. Haven't had a chance to look at that though.

MGatner:
@lonnieezell I had the same thought but when I looked into it `FeatureTestTrait` doesn't actually do any staging at all, it handles everything during `call()`.

I'm doing some massive test optimization on a project right now, leveraging traits instead of my previous class extension methods. I've bought my test time down to about 1/5 of what it used to be. When I tackle the feature tests part I will see if anything sticks out to me as needing improvements.

lonnieezell:
> I've bought my test time down to about 1/5 of what it used to be.

@MGatner Wow. That's sweet. I'm always for faster tests. That means I'll run the full suite more often :)

MGatner:
Failures were because I forgot namespaces :O Fix incoming.

paulbalandan:
Would it be better to add those functions in common? I think usefulness of those will go beyond testing.

MGatner:
I'm open to it. At this point they weren't needed elsewhere and since they can be moved any time I figured I'd throw them in testing, but I'm glad to hear otherwise.

PR #4420, opened 2021-03-12
[Rector] Apply Rector: UnusedForeachValueToArrayKeysRector

samsonasik:
Change foreach with unused `$value` but only `$key`, to `array_keys()` with `UnusedForeachValueToArrayKeysRector` rule.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -2,6 +2,7 @@
 
 use Rector\CodeQuality\Rector\Expression\InlineIfToExplicitIfRector;
 use Rector\CodeQuality\Rector\For_\ForToForeachRector;
+use Rector\CodeQuality\Rector\Foreach_\UnusedForeachValueToArrayKeysRector;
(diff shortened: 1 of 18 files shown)


iRedds:
I may be wrong, but it seems to me that the array_keys function contains a loop inside.
https://github.com/php/php-src/blob/fe3dbe52944af66a61281987147c2c1d2ca59afe/ext/standard/array.c#L3823

[php]
// I mean 
foreach(array_keys($array) as $value) {
    //some work 
}
// It looks like: 
$arrayKeys = [];
foreach ($array as $key => $value) {
    $arrayKeys[] = $key;
}

foreach ($arrayKeys as $value) { 
        //some work 
}
[/php]

samsonasik:
All green 🎉

paulbalandan (approved):

MGatner (commented):
Wow I'm guilty of this. I like the solution.

PR #4423, opened 2021-03-13
Add more info to `seeInDatabase` assertions

lonnieezell:
Refactored the seeInDatabase assertion into a Constraint so we can provide context when it fails by showing potentially similar rows within the database.


MGatner:
This is awesome! Glad you got it in. Now I want to make some failures so I can use it 😂

PR #4424, opened 2021-03-13
[Helper] Fixes base_url("0") and site_url("0") to correctly point to /0

samsonasik:
Handle `base_url("0")` to ensure point to `/0` with compare to empty string instead of using `empty($uri)` check.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -47,7 +47,7 @@ if (! function_exists('site_url'))
 		{
 			$fullPath .= rtrim($config->indexPage, '/');
 		}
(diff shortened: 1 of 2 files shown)


samsonasik:
fixes site_url() with 0 as well

samsonasik:
All green 🎉

MGatner (changes requested):
> $uri URI string or array of URI segments

I think you've eliminated the check against `[]`

MGatner (approved):
Oops! Nevermind I just looked at the whole file and we've imploded by that point. Looks good!

PR #4425, opened 2021-03-13
Prioritize database trait methods

MGatner:
**Description**
An alternative to #4422 (see the description of the problem there). I think this simpler solution is the better choice. I do think `class_uses_recursive()` is probably costly so I took advantage still of storing the results for a little performance boost.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/CIUnitTestCase.php
+++ b/system/Test/CIUnitTestCase.php
@@ -162,6 +162,13 @@ abstract class CIUnitTestCase extends TestCase
 	//--------------------------------------------------------------------
 
 	/**
+	 * Store of identified traits.
+	 *
+	 * @var string[]|null
+	 */
+	private $traits;
+
+	/**
 	 * Load the helpers.
 	 */
 	public static function setUpBeforeClass(): void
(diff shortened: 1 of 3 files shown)


user_guide_src/source/testing/overview.rst
```diff
@@ -155,11 +155,9 @@ Traits
 
 A common way to enhance your tests is by using traits to consolidate staging across different
 test cases. ``CIUintTestCase`` will detect any class traits and look for staging methods to
```
paulbalandan, line 157:
```suggestion
test cases. ``CIUnitTestCase`` will detect any class traits and look for staging methods
```

paulbalandan (approved):

system/Test/CIUnitTestCase.php
```diff
@@ -185,16 +192,14 @@ protected function setUp(): void
...
-
-			if (method_exists($this, $method))
-			{
-				$this->$method();
-			}
+			$this->setUpDatabase();
 		}
+
+		// Check for other trait methods
+		$this->callTraitMethods('setUp');
```
lonnieezell, line 202: Wouldn't this end up running the database methods twice?

MGatner: Nope, because I renamed its method to `setUpDatabase()` which no longer matches the paradigm `setUp{TraitName}()`. Good catch though, I almost made that mistake myself!

PR #4427, opened 2021-03-14
Enable to use extened CodeIgniter class in testing

kenjis:
**Description**
#4376 makes the `CodeIgniter` class extensible.
But `CIUnitTestCase` has hard coded code `new MockCodeIgniter()`.

When we extend the `CodeIgniter` class, we need to use the extended `MockCodeIgniter`
in testing. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -168,8 +168,7 @@ class CodeIgniter
 	 */
 	public function initialize()
 	{
-		// Define environment variables
-		$this->detectEnvironment();
+		// Load boot file
 		$this->bootstrapEnvironment();
 
 		// Setup Exception Handling
@@ -494,36 +493,6 @@ class CodeIgniter
 		Events::trigger('post_system');
 
 		return $this->response;
-	}
-
-	//--------------------------------------------------------------------
-
-	/**
-	 * You can load different configurations depending on your
-	 * current environment. Setting the environment also influences
-	 * things like logging and error reporting.
-	 *
-	 * This can be set to anything, but default usage is:
-	 *
-	 *     development
-	 *     testing
-	 *     production
-	 */
-	protected function detectEnvironment()
-	{
-		// Make sure ENVIRONMENT isn't already set by other means.
-		if (! defined('ENVIRONMENT'))
-		{
-			// running under Continuous Integration server?
-			if (getenv('CI') !== false)
-			{
-				define('ENVIRONMENT', 'testing');
-			}
-			else
-			{
-				define('ENVIRONMENT', $_SERVER['CI_ENVIRONMENT'] ?? 'production');
-			}
-		}
 	}
 
 	//--------------------------------------------------------------------
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -47,6 +47,7 @@ use CodeIgniter\Router\RouteCollectionInterface;
 use CodeIgniter\Router\Router;
 use CodeIgniter\Security\Security;
 use CodeIgniter\Session\Session;
+use CodeIgniter\Test\Mock\MockCodeIgniter;
 use CodeIgniter\Throttle\Throttler;
 use CodeIgniter\Typography\Typography;
 use CodeIgniter\Validation\Validation;
@@ -150,6 +151,11 @@ class Services extends BaseService
 		}
 
 		$config = $config ?? config('App');
+
+		if (ENVIRONMENT === 'testing')
+		{
+			return new MockCodeIgniter($config);
+		}
 
 		return new CodeIgniter($config);
 	}
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -152,11 +152,6 @@ class Services extends BaseService
 
 		$config = $config ?? config('App');
 
-		if (ENVIRONMENT === 'testing')
-		{
-			return new MockCodeIgniter($config);
-		}
-
 		return new CodeIgniter($config);
 	}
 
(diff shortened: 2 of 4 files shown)


kenjis:
```sh-session
$ php spark
PHP Warning:  Use of undefined constant ENVIRONMENT - assumed 'ENVIRONMENT' (this will throw an Error in a future version of PHP) in /Users/kenji/work/codeigniter/CodeIgniter4/system/Config/Services.php on line 155
PHP Stack trace:
PHP   1. {main}() /Users/kenji/work/codeigniter/CodeIgniter4/spark:0
PHP   2. require() /Users/kenji/work/codeigniter/CodeIgniter4/spark:53
PHP   3. CodeIgniter\Config\BaseService::codeigniter() /Users/kenji/work/codeigniter/CodeIgniter4/system/bootstrap.php:156
PHP   4. CodeIgniter\Config\BaseService::__callStatic($name = 'codeigniter', $arguments = []) /Users/kenji/work/codeigniter/CodeIgniter4/system/bootstrap.php:156
PHP   5. CodeIgniter\Config\Services::codeigniter($config = *uninitialized*, $getShared = *uninitialized*) /Users/kenji/work/codeigniter/CodeIgniter4/system/Config/BaseService.php:245
PHP   6. CodeIgniter\Config\BaseService::getSharedInstance($key = 'codeigniter', ...$params = variadic(NULL)) /Users/kenji/work/codeigniter/CodeIgniter4/system/Config/Services.php:150
PHP   7. CodeIgniter\Config\Services::codeigniter($config = NULL, $getShared = FALSE) /Users/kenji/work/codeigniter/CodeIgniter4/system/Config/BaseService.php:174
```

MGatner:
We can't remove `detectEnvironment()` from `CodeIgniter`. I definitely think there are some improvements to be made to the whole entry point & app instantiation, but I don't like the Service being environment-dependent. Probably a better short-term solution would be for `CIUnitTestCase` to inject `MockCodeIgniter` into the service.

kenjis:
> I don't like the Service being environment-dependent.

Can you tell me what's wrong or your concern?

> Probably a better short-term solution would be for CIUnitTestCase to inject MockCodeIgniter into the service.

How and where do we inject extended MockCodeIgniter into the service?

MGatner:
It's opposite of how Services usually function. The framework comes with a default adapter for every service, and no matter the context this will be the same. When services need to be mocked either so they can be tested or so they don't get in the way of other tests, we inject the alternate version during test set up. 

Here I would like to know why there is a desire to use the service instead of MockCodeIgniter directly. I do think injecting it makes sense in case anything else calls the service, but I don't know why our tests would need to rely on the service to begin with.

kenjis:
I extended `CodeIgniter` to use a DI container to instantiate controllers.
The DI container gives all dependent objects to the controlle constructor (constructor injection) to instantiate it.

A controller is like this:
```php
class CustomerInfo extends MyController
{
   ...

    public function __construct(
        CartRepository $cartRepository,
        CustomerInfoRepository $customerInfoRepository,
        Twig $twig
    ) {
        parent::__construct();

        $this->cartRepository = $cartRepository;
        $this->customerInfoRepository = $customerInfoRepository;
        $this->twig = $twig;
    }

   ...
```

When I ran feature tests, all tests failed. Because the `MockCodeIgniter` can't instantiate
controllers correctly.

I've never used `ControllerTester` trait yet, but the same problem is in it.

MGatner:
I would say that in your project you have overridden a core service and this you will need to handle that in your tests - this isn't something that the framework needs to be accounting for as long as it provides the tools to handle it (Services::injectMock)

paulbalandan:
Initially looking at the code I agree on this but hearing @MGatner 's reasons makes me think that, yes, services should not be dependent on the environment it runs. So my suggestion would be to use the injectMock feature of services. Besides, we are in testing mode so we can also take advantage of the mocking feature.

It goes something like this:
```diff
-$app = new MockCodeIgniter(new App);
+$app = Services::injectMock('codeigniter', new MockCodeIgniter(new App));
```

I'm not sure if I followed the correct syntax for mocking services. But something along that we can achieve this.

MGatner:
That's the right idea. I think this is what we want in the framework:
```
$app = new MockCodeIgniter(new App);
Services::injectMock('codeigniter', $app);
```

This won't fix @kenjis problem but he can do the same in his project's test case:
```
$app = new KenjisCoolInjectorCodeIgniter(new App);
Services::injectMock('codeigniter', $app);
```

kenjis:
Thank you for thinking about my PR.
I agree with @MGatner , services should not be dependent on the environment it runs.
In the first place, code like `(ENVIRONMENT === 'testing')` is not a good design.

I've changed the code in `CIUnitTestCase`:
https://github.com/codeigniter4/CodeIgniter4/blob/128931375279ce0f10e3664cfc9133b60d338c63/system/Test/CIUnitTestCase.php#L530-L542

But after all I need to override the method in my use case.
Apparently, this PR's way is not very fruitful.
It makes no big difference whether I extract the following code into a method and override only that, or override the entire  `createApplication` method.

```php
 	$app = new MockCodeIgniter(new App()); 
 	Services::injectMock('codeigniter', $app); 
```

Traitify FeatureTestCase made my life little easier, and
it seems my case is not common, I take down this PR.

MGatner:
Thanks for the feedback @kenjis. After understanding you use case, I believe the following:
1. It is easily accomplished with existing tools (the PR you mentioned helps a lot)
2. It represents an overriding of core services so should *not* be supported natively by the framework

I am glad you have found a solution. If you want to talk more about testing solutions I have been working with them *a lot* recently, feel free to hop on Slack and ping me.

PR #4436, opened 2021-03-15
Repeat App Routes

MGatner:
**Description**
Presumably for performance reasons, `RouteCollection` skips re-loading routes from `App`. This works until feature tests reset services and reload routes. Then the collection fails to discover App routes during `discoverRoutes()`.

I think the root of this issue is because the original Routes service is initiated during bootstrap (see **system/Test/bootstrap.php** at the bottom) so there may be a better solution by moving that into the test classes somewhere, but I spent so long tracking this down I have to get it written all out or I will forget it.

EDIT: My thought was correct. I've replaced my original solution with this variant. Feature Tests will now use the App initialized version of `RouteCollection` instead of the raw service version.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -17,6 +17,7 @@ use CodeIgniter\HTTP\Request;
 use CodeIgniter\HTTP\URI;
 use CodeIgniter\HTTP\UserAgent;
 use CodeIgniter\Router\Exceptions\RedirectException;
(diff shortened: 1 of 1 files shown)


MGatner:
Edit: Forget the above, I think I found the better solution. Feature Tests will now use the App initialized version of `RouteCollection` instead of the raw service version.

lonnieezell (approved):

PR #4442, opened 2021-03-16
[WIP] Include "LIMIT 0" in compiled SELECT statements

brickcamp:
**Description**
The following function of the BaseBuilder class permits setting LIMIT to zero.

```
limit(?int $value = null, ?int $offset = 0)
```

I think this is by design, as the default value here is `null`, in contrast to `$offset`. But setting it to zero has no effect on the resulting SQL statement (to my awareness). 

This is caused by the `compileSelect` method. Adding "LIMIT 0" is skipped there because zero is interpreted as `false` - the empty default value. So, instead of an empty result, all table lines (fitting the other conditions) are returned.

So I propose changing the IF statement in `compileSelect` to differentiate between `0` and `false` when checking `$this->QBLimit`.

**Checklist:**
Sorry, I don't know how to fulfill the points on the checklist. But I searched all open and closed issues / PRs and haven't noticed this topic anywhere. So, I hope this small change is helpful nevertheless :)

**EDIT**: Oh, sorry, I see the tests failing one after the other. So that's not the solution.

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2042,7 +2042,7 @@ class BaseBuilder
 			$this->where($where);
 		}
 
-		if (! empty($limit))
+		if (! is_null($limit))
 		{
 			$this->limit($limit, $offset);
 		}
@@ -2466,7 +2466,7 @@ class BaseBuilder
 			$this->where($where);
 		}
 
-		if (! empty($limit))
+		if (! is_null($limit))
 		{
 			if (! $this->canLimitWhereUpdates)
 			{
@@ -2522,7 +2522,7 @@ class BaseBuilder
 		return 'UPDATE ' . $this->compileIgnore('update') . $table . ' SET ' . implode(', ', $valStr)
 				. $this->compileWhereHaving('QBWhere')
 				. $this->compileOrderBy()
-				. ($this->QBLimit ? $this->_limit(' ', true) : '');
+				. ($this->QBLimit !== false ? $this->_limit(' ', true) : '');
 	}
 
 	//--------------------------------------------------------------------
@@ -2861,12 +2861,12 @@ class BaseBuilder
 
 		$sql = $this->_delete($table);
 
-		if (! empty($limit))
+		if (! is_null($limit))
 		{
 			$this->QBLimit = $limit;
 		}
 
-		if (! empty($this->QBLimit))
+		if ($this->QBLimit !== false)
 		{
 			if (! $this->canLimitDeletes)
 			{
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -3040,7 +3040,7 @@ class BaseBuilder
 				. $this->compileWhereHaving('QBHaving')
 				. $this->compileOrderBy(); // ORDER BY
 		// LIMIT
-		if ($this->QBLimit)
+		if ($this->QBLimit !== false)
 		{
 			return $this->_limit($sql . "\n");
 		}
--- a/system/Database/Postgre/Builder.php
+++ b/system/Database/Postgre/Builder.php
@@ -245,7 +245,7 @@ class Builder extends BaseBuilder
 	 */
 	public function delete($where = '', int $limit = null, bool $resetData = true)
 	{
-		if (! empty($limit) || ! empty($this->QBLimit))
+		if (! is_null($limit) || $this->QBLimit !== false)
 		{
 			throw new DatabaseException('PostgreSQL does not allow LIMITs on DELETE queries.');
 		}
@@ -286,7 +286,7 @@ class Builder extends BaseBuilder
 	 */
 	protected function _update(string $table, array $values): string
 	{
-		if (! empty($this->QBLimit))
+		if ($this->QBLimit !== false)
 		{
 			throw new DatabaseException('Postgres does not support LIMITs with UPDATE queries.');
 		}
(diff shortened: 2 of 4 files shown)


sfadschm:
I am not sure whether the current behaviour is desired. What is the use of setting `LIMIT 0`?
However, your tests are failing because `null` (the default value) passes your condition. I think it should be `!== NULL`.

brickcamp:
Concerning the why: The documentation of for example [MySQL](https://dev.mysql.com/doc/refman/5.7/en/limit-optimization.html) explicitly allows it - saying:

> `LIMIT 0` quickly returns an empty set. This can be useful for checking the validity of a query. It can also be employed to obtain the types of the result columns within applications that use a MySQL API that makes result set metadata available...

The need for this is debatable of course - agreed. For me, the design of the `limit` function implied, that it should work. Meaning the null-default on `value` versus the zero-default on `offset`. The behaviour of the API surprised me so I thought I bring it up.

sfadschm:
> For me, the design of the `limit` function implied, that it should work.

I would think so, too. Maybe a good starting ppint would be to apply the strict `false` test to all other truthy checks of `qbLimit` in BaseBuilder, too (e.g., line 2525). Not sure if that would fix the tests though. 
I will take a look tomorrow.

sfadschm (changes requested):
I think all these occurences need to be changed.

Edit: Yeah this went wrong....

system/Database/BaseBuilder.php
```diff
@@ -3040,7 +3040,7 @@ protected function compileSelect($selectOverride = false): string
 				. $this->compileWhereHaving('QBHaving')
 				. $this->compileOrderBy(); // ORDER BY
 		// LIMIT
-		if ($this->QBLimit)
+		if ($this->QBLimit !== false)
```
sfadschm, line 3043:
```suggestion
		if (false !== $this->QBLimit)
```

sfadschm: I think we do this all over the framework in that order now.

sfadschm (changes requested):
Also change
https://github.com/codeigniter4/CodeIgniter4/blob/e52293644ee39ed07efd6d61d30878436e646d95/system/Database/BaseBuilder.php#L2525
https://github.com/codeigniter4/CodeIgniter4/blob/e52293644ee39ed07efd6d61d30878436e646d95/system/Database/BaseBuilder.php#L2869

sfadschm:
The problem with the test is a deeper one.
In `Model.php`, we have
https://github.com/codeigniter4/CodeIgniter4/blob/97c73b3c55323381ff17e4912290d3b687847ad7/system/Model.php#L186
Where the default value for `$limit` is `0` for some reason.
I think someone from the dev team should take a look if this can be changed with BC.

brickcamp:
Oh okay, the `doFindAll` definitely makes this harder. The default value must stay as part of the API for BC I think. Maybe one could change it to act like false/null was passed if `$limit` is set to zero there. Haven't looked at it yet. The difference in default values could also be pointed out in the documentation. 

I will have a look at it, might take time up to the weekend though. But this really intrigued me now. I would like to make a complete proposal - not one, that fails all the tests 😅 Thanks for pointing out the other lines in the code!

system/Database/BaseBuilder.php
brickcamp: Are you sure? I searched the codebase and it's mostly done with false at the end (140+ vs. 9). Same for "=== false" (150+ vs 17). But I can switch it around of course! :slightly_smiling_face: Will do so when the tests pass as I just added more of this :sweat_smile:

sfadschm: Oh okay 😆 
Maybe just wait with that then. If it is desired, someone from the team will mention it after everything works.

brickcamp:
Okay, now [this assert](https://github.com/codeigniter4/CodeIgniter4/blob/2abeb6ea6d6524052608453bce06387e2e9d1552/tests/system/Database/Builder/GetTest.php#L43) is the only one failing in all the checks. But this really brings up the backwards compatibility issue we feared :sweat_smile: 

The default value of the corresponding method parameter is defined as `null`. But of course, users can pass in zero nevertheless and depend on this not resulting in `LIMIT 0` in the SQL statement. Just like this test does.

I guess making `LIMIT 0` work isn't worth breaking running installations. One could implement, that only the `limit` method of the query builder can result in a `LIMIT 0`, but this seems incoherent. Also, current users could have done the same thing there.

So, I guess the best way would be to define that CodeIgniter shall never pass `LIMIT 0` (or similiar statements like `TOP(0)`) to the database. That's in fact what the current implementation does. It should just be explicitly documented (e.g. [here](https://codeigniter.com/user_guide/database/query_builder.html?highlight=get#limit)) to avoid confusion (which is why I brought it up :slightly_smiling_face:).

If this is the way to go, maybe the default values of `limit` parameters could then be changed to `0` to make this behaviour more obvious. For existing redefinitions of methods, `null` and `false` should still be handled though.

system/Database/BaseBuilder.php
paulbalandan: We don't do yoda conditions. :)

sfadschm: Okay my bad. Thought I saw it in a recent PR.

pixobit:
Perhaps this could be merged into the next major version, where BC isn't such a strict rule

paulbalandan:
It seems the user who opened this PR got deleted so there's no chance this will get completed. If someone wants to take over please do so, but I'm closing this first.

Issue #4438, opened 2021-03-16
Request: Feature Test Optimization


MGatner:
Feature tests are really slow. It is not unusual for a single test to take up to 2 seconds. I have optimized my set up as best I can, running database migrations & seeds once but the tests are still magnitudes slower than other tests.

If anyone has experience analyzing PHP runtimes to figure out where our bottlenecks are, improving execution in the framework would be a huge resource savings trickling down to all projects.

kenjis:
You can profile with Xdebug.
See <https://xdebug.org/docs/profiler>.

lonnieezell:
I've never had the chance but at some point would like to dive deep into the database layer as that seems the slowest link. And I don't think it's just the database connection. 

I'm assuming you're using an in-memory SQLite db for your tests already?

MGatner:
Yea this is with SQLite, memory, and `$migrateOnce` and `$seedOnce`. Similar non-feature tests run very quickly - it is definitely something else happening. 

@kenjis Thanks for the tip! I already use Xdebug a lot so I will look into this profiler.

MGatner:
I believe the suggestions here are sufficient tools for anyone looking to optimize their projects.

PR #4445, opened 2021-03-17
Use speedtrap to document slow tests

paulbalandan:
**Description**
Use `johnkary/phpunit-speedtrap` to document slow tests in the framework's test suites. This is the starting point to optimize the speed of our tests.

Ref: #4438 

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,20 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		bootstrap="system/Test/bootstrap.php"
-		backupGlobals="false"
-		colors="true"
-		convertErrorsToExceptions="true"
-		convertNoticesToExceptions="true"
-		convertWarningsToExceptions="true"
-		stopOnError="false"
-		stopOnFailure="false"
-		stopOnIncomplete="false"
-		stopOnSkipped="false"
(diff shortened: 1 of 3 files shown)


phpunit.xml.dist
```diff
@@ -39,7 +42,28 @@
...
 	</testsuites>
-	<logging/>
+
+	<listeners>
+		<listener class="JohnKary\PHPUnit\Listener\SpeedTrapListener">
+			<arguments>
+				<array>
+					<!-- Number of milliseconds when a test is considered "slow" (Default: 500ms) -->
+					<element key="slowThreshold">
+						<integer>500</integer>
```
paulbalandan, line 52: We should set our minimum slowness threshold here.

MGatner: I think starting at 500 (the default) makes sense. We can inline different thresholds or specific tests to ignore once we identify ones that "can't" be fixed.

This might be way beyond the scope of this PR but... I would love annotations from this.😅 Could you look into if anyone has found a solution for that yet?

paulbalandan: Annotations like this?
```php
/** @slowThreshold 1000 */
public function testAnythingHere(): void
{
    // ...
}
```

MGatner: No, some Action services will actually annotate the file under Review. PHPStan does it. I can't find an example right now but I can create one real quick.

MGatner: See here for how the Action result gets overlaid with its context: https://github.com/codeigniter4/CodeIgniter4/pull/4451/files

MGatner (approved):

PR #4446, opened 2021-03-17
Filter Test Trait

MGatner:
**Description**
This adds a new test trait to assist with testing filters in an efficient and concise manner.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/FilterTestTrait.php
+++ b/system/Test/FilterTestTrait.php
@@ -126,8 +126,6 @@ trait FilterTestTrait
 	 * @param string                 $position "before" or "after"
 	 *
 	 * @return callable
-	 *
-	 * @noRector \Rector\EarlyReturn\Rector\If_\ChangeIfElseValueAssignToEarlyReturnRector
 	 */
 	protected function getFilterCaller($filter, string $position): callable
 	{
@@ -162,21 +160,17 @@ trait FilterTestTrait
 		{
 			$request = clone $this->request;
 
-			$fn = function (array $params = null) use ($filter, $request) {
+			return function (array $params = null) use ($filter, $request) {
 				return $filter->before($request, $params);
 			};
 		}
-		else
-		{
-			$request  = clone $this->request;
-			$response = clone $this->response;
 
-			$fn = function (array $params = null) use ($filter, $request, $response) {
-				return $filter->after($request, $response, $params);
-			};
-		}
+		$request  = clone $this->request;
+		$response = clone $this->response;
 
-		return $fn;
+		return function (array $params = null) use ($filter, $request, $response) {
+			return $filter->after($request, $response, $params);
+		};
 	}
 
 	/**
--- a/system/Test/FilterTestTrait.php
+++ b/system/Test/FilterTestTrait.php
@@ -156,16 +156,15 @@ trait FilterTestTrait
 			throw FilterException::forIncorrectInterface(get_class($filter));
 		}
 
+		$request = clone $this->request;
+
 		if ($position === 'before')
 		{
-			$request = clone $this->request;
-
 			return function (array $params = null) use ($filter, $request) {
 				return $filter->before($request, $params);
 			};
 		}
 
-		$request  = clone $this->request;
 		$response = clone $this->response;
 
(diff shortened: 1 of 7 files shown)


MGatner:
Rector complaining about my `if..else` but I would like to keep the format since it makes clear the "before" and "after" functions. @samsonasik is there a way to direct Rector to ignore a line, similar to PHPStan?

samsonasik:
`@noRector` seems can be used https://github.com/rectorphp/rector/blob/main/docs/how_to_ignore_rule_or_paths.md#in-a-file

MGatner:
Thanks @samsonasik, trying that now. Squashing some of the commits as well, then I would appreciate a review if you're still around.

MGatner:
All green 🎉

tests/system/Filters/FiltersTest.php
```diff
@@ -1069,4 +1069,26 @@ public function testFilterClass()
...
+					'before' => ['admin*'],
+				],
+			],
+		];
+
+		$filters = new Filters((object) $config, $this->request, $this->response);
+		$uri     = 'admin';
+
+		$this->assertEquals(['foo'], $filters->initialize($uri)->getFilters()['before']);
+		$this->assertEquals([], $filters->reset()->getFilters()['before']);
```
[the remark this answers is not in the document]
MGatner: Is that preferable? That tests for anything falsey whereas `assertEquals` is more specific. I actually usually use `assertSame` for empty arrays so it includes typing - I probably pasted this one in.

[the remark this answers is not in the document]
samsonasik: ok, assertSame() should be ok then for consistency

system/Test/FilterTestTrait.php
```diff
@@ -0,0 +1,289 @@
...
+		}
+
+		if (! $filter instanceof FilterInterface)
+		{
+			throw FilterException::forIncorrectInterface(get_class($filter));
+		}
+
+		if ($position === 'before')
+		{
+			$request = clone $this->request;
```
paulbalandan, line 161: put this before the if so line 168 would be not needed

system/Test/FilterTestTrait.php
```diff
@@ -0,0 +1,289 @@
...
+	 *
+	 * @param string $route    The route to test
+	 * @param string $position "before" or "after"
+	 * @param string $alias    Alias for the anticipated filter
+	 *
+	 * @return void
+	 */
+	protected function assertFilter(string $route, string $position, string $alias): void
+	{
+		if (! in_array($position, ['before', 'after'], true))
```
paulbalandan, line 216: move this check to getFiltersForRoute so that other assertions can check the value of position

system/Test/FilterTestTrait.php
MGatner: Good suggestion, implementing.

system/Test/FilterTestTrait.php
MGatner: Ah thanks I thought I had done this. I seem to have crossed commits at one point because my Rector changes reverted so maybe I reverted that too? Fixing.

system/Test/FilterTestTrait.php
```diff
@@ -0,0 +1,288 @@
...
+	//--------------------------------------------------------------------
+
+	/**
+	 * Initializes dependencies once.
+	 *
+	 * @return void
+	 */
+	protected function setUpFilterTestTrait(): void
+	{
+		if ($this->doneFilterSetUp === false)
```
samsonasik, line 91: how about flip the if and `return; `early?

```php
if ($this->doneFilterSetUp === true)
{
    return;
}

// Create our own Request and Response so we can
```
by that, the deep if can be reduced.

MGatner: Yes good point. I used to have some stuff outside that `if` but now it controls the whole method. Will change.

MGatner:
I'm off to bed. Someone please merge if it is deemed ready.

paulbalandan (approved):

samsonasik (approved):

PR #4444, opened 2021-03-17
Fix Query named bind recognition

paulbalandan:
**Description**
Fixes #4443 by changing the regex used by `Query` on how it recognizes a named bind by including parentheses.

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPdocs
- [x] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #4448, opened 2021-03-17
Fix expressions in Database/Query

WinterSilence:
post-fix for #4444

**Description**
Fix regexp's and optimize "if" statement. 

https://regex101.com/r/pMoYqY/1

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -321,11 +321,11 @@ class Query implements QueryInterface
 	{
 		$sql = $this->finalQueryString;
 
-		$hasNamedBinds = preg_match('/\:(?:[\w\.\(\)]+)\:/', $sql) === 1;
+		$hasNamedBinds = preg_match('/\:(?:[a-z\d.)_(]+)\:/i', $sql) === 1;
 
 		if (empty($this->binds)
 			|| empty($this->bindMarker)
-			|| (strpos($sql, $this->bindMarker) === false && $hasNamedBinds === false)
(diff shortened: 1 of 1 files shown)


system/Database/Query.php
```diff
@@ -321,11 +321,11 @@ protected function compileBinds()
 	{
 		$sql = $this->finalQueryString;
 
-		$hasNamedBinds = preg_match('/\:(?:[\w\.\(\)]+)\:/', $sql) === 1;
+		$hasNamedBinds = preg_match('/\:(?:[a-z\d.)_(]+)\:/i', $sql) === 1;
```
paulbalandan, line 324: Please update the link of the regex101 in the phpdoc.

WinterSilence:
@paulbalandan i rewrite it shortest, test link updated

paulbalandan (approved):

system/Database/Query.php
```diff
@@ -315,17 +315,17 @@ public function getOriginalQuery(): string
 	 *
 	 * @return void
 	 *
-	 * @see https://regex101.com/r/jgAWNQ/2
+	 * @see https://regex101.com/r/EUEhay/1 Test
```
samsonasik, line 318: you can use "Update Regex" instead of "Save Regex" in regex101.com so the uri path will continue /2 /3 etc instead of new uri start over from /1

WinterSilence: @samsonasik didn't know about it

PR #4455, opened 2021-03-19
[Rector] Apply Rector: RemoveUnusedPrivateMethodRector

samsonasik:
Remove unused private method with `RemoveUnusedPrivateMethodRector` rector rule.

Also currently pin `symplify/rule-doc-generator` to v9.2.11 due removed `Symplify\RuleDocGenerator\Contract\DocumentedRuleInterface` see https://github.com/rectorphp/rector/issues/5908

The interface moved to `symplify/rule-doc-generator-contracts` that will be included in next rector release, but for now, pin to `symplify/rule-doc-generator` to v9.2.11.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Entity.php
+++ b/system/Entity.php
@@ -577,22 +577,6 @@ class Entity implements JsonSerializable
 		return $handlers[$type]::$method($value, $params);
 	}
 
-	//--------------------------------------------------------------------
-
-	/**
-	 * Cast as JSON
-	 *
-	 * @param mixed   $value
-	 * @param boolean $asArray
-	 *
-	 * @return mixed
-	 * @throws CastException
-	 */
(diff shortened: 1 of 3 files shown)


system/Entity.php
```diff
@@ -577,22 +577,6 @@ protected function castAs($value, string $attribute, string $method = 'get')
...
-	 * @param mixed   $value
-	 * @param boolean $asArray
-	 *
-	 * @return mixed
-	 * @throws CastException
-	 */
-	private function castAsJson($value, bool $asArray = false)
-	{
-		return CastAsJson::get($value, $asArray ? ['array'] : []);
-	}
```
samsonasik, line 594: oh, seems it is actually used magically.

PR #4462, opened 2021-03-21
Switch to tachycardia for slow tests profiling and reporting

paulbalandan:
**Description**
Enhancement to `johnkary/phpunit-speedtrap` that can have inline annotations for Github Actions.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -122,6 +122,10 @@ jobs:
         env:
           COMPOSER_AUTH: ${{ secrets.COMPOSER_AUTH }}
 
+      - name: Profile slow tests in PHP 7.4 MySQLi for now
+        if: matrix.php-versions == '7.4' && matrix.db-platforms == 'MySQLi'
+        run: echo "TACHYCARDIA_MONITOR_GA=enabled" >> $GITHUB_ENV
+
       - name: Test with PHPUnit
         run: script -e -c "vendor/bin/phpunit --color=always"
         env:
--- a/composer.json
+++ b/composer.json
@@ -17,8 +17,8 @@
 	"require-dev": {
 		"codeigniter4/codeigniter4-standard": "^1.0",
 		"fakerphp/faker": "^1.9",
-		"johnkary/phpunit-speedtrap": "^3.3",
 		"mikey179/vfsstream": "^1.6",
+		"nexusphp/tachycardia": "^1.0",
 		"phpstan/phpstan": "0.12.82",
 		"phpunit/phpunit": "^9.1",
 		"predis/predis": "^1.1",
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -43,26 +43,26 @@
 		</testsuite>
 	</testsuites>
 
-	<listeners>
-		<listener class="JohnKary\PHPUnit\Listener\SpeedTrapListener">
+	<extensions>
+		<extension class="Nexus\PHPUnit\Extension\Tachycardia">
 			<arguments>
 				<array>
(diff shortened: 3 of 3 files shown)


kenjis:
> Nexus\PHPUnit\Extension\Tachycardia identified these 30 slow tests:
⚠  Took 13.06s from 0.50s limit to run CodeIgniter\\CLI\\CommandRunnerTest::testGoodCommand

I don't know why the test takes so long.

On my local environment:
```sh-session
$ vendor/bin/phpunit tests/system/CLI/ --no-coverage 
PHPUnit 9.5.3 by Sebastian Bergmann and contributors.

................................................                  48 / 48 (100%)

Time: 00:02.875, Memory: 14.00 MB

OK (48 tests, 96 assertions)
```

MGatner:
This looks awesome! Thanks for whipping this up - I hope it was a fun project. This made me laugh:

> but injected with anabolic steroids

Looks like there is a merge conflict, not sure what we changed in composer recently?

.github/workflows/test-phpunit.yml
```diff
@@ -122,6 +122,10 @@ jobs:
         env:
           COMPOSER_AUTH: ${{ secrets.COMPOSER_AUTH }}
 
+      - name: Profile slow tests in PHP 7.4 MySQLi for now
+        if: matrix.php-versions == '7.4' && matrix.db-platforms == 'MySQLi'
+        run: echo "TACHYCARDIA_MONITOR_GA=enabled" >> $GITHUB_ENV
```
MGatner, line 127: It looks like the extension is enabled by default and I don't see your configuration disabling it anywhere. Why this explicit `enabled`?

paulbalandan: This env variable is for enabling the reporting in Github Actions as it is disabled by default.

MGatner: Is that just the `tabulate` or the entire extension? That was not clear to me from the docs. I added this to my toolkit (https://github.com/tattersoftware/codeigniter4-tools/commit/42124c8e18c7ba0499aa6539a760d67ad7c57242) but skipped the env Config because I thought it would run by default (with tabulate true).

MGatner (approved):

.github/workflows/test-phpunit.yml
paulbalandan: `tabulate: true` would just turn the console report from plain listing to tabular format.

The extension is handled by two env variables: `TACHYCARDIA_MONITOR` for console reporting (enabled by default) and `TACHYCARDIA_MONITOR_GA` for Github Actions reporting (disabled by default). The `tabulate` option is used only by the console reporting to modify the report's format. For GA, we use the workflow command syntax for annotating.

paulbalandan:
> > Nexus\PHPUnit\Extension\Tachycardia identified these 30 slow tests:
> > ⚠  Took 13.06s from 0.50s limit to run CodeIgniter\CLI\CommandRunnerTest::testGoodCommand
> 
> I don't know why the test takes so long.
> 
> On my local environment:
> 
> ```
> $ vendor/bin/phpunit tests/system/CLI/ --no-coverage 
> PHPUnit 9.5.3 by Sebastian Bergmann and contributors.
> 
> ................................................                  48 / 48 (100%)
> 
> Time: 00:02.875, Memory: 14.00 MB
> 
> OK (48 tests, 96 assertions)
> ```

I think this depends on the local machine and within PHPUnit itself on its bootstrap.

.github/workflows/test-phpunit.yml
MGatner: Thanks for explaining. I see you added docs too! I will update my toolkit. For CI4, I think this PR is ready, unless you have more changes to commit?

PR #4463, opened 2021-03-22
Fix wrong FROM provided after reset query

### LINKED ISSUE #4461: Bug: shared BaseBuilder instance and Model
**Describe the bug**
Wrong approach to implementing a generic BaseBuilder instance in the model and post processing queries to the database. 
When using a subquery, the work of the model / builder breaks. 

```php
$t = new Test(); //table test

$t->from('(SELECT * FROM test_1) t', true)->paginate(1);
d($t->db->getLastQuery()); // "SELECT * FROM (SELECT * FROM test_1) t LIMIT 1"

$t->findAll();
d($t->db->getLastQuery()); // "SELECT * FROM (SELECT * FROM test_1) t"  expected SELECT * FROM test

$t->from('(SELECT id, name FROM test_1) t', true)->paginate(1);
d($t->db->getLastQuery()); // "SELECT * FROM (SELECT id, name FROM test_1) t LIMIT 1"

$t->findAll(); // SQL error -  SELECT * FROM (SELECT id
```

The fact is that after the query is executed, the table is reassigned. 
https://github.com/codeigniter4/CodeIgniter4/blob/de8d7f40d799d584d86af2945deac113e1a7c840/system/Database/BaseBuilder.php#L3396-L3400

For the first case, the value will be: ``(SELECT * FROM test_1) t`` 
For the second case: ``(SELECT id``, because the from() method can accept as a table name a string consisting of several tables separated by commas: from('table_1, table_2, table_3)

**Affected module(s)**
Model, BaseBuilder

**You can fix this with crutches, but I think it's time to change the concept of the model and query builder.**

kenjis:
**Description**
Fixes #4461

I'm not sure this can be perfect. But the first attempt to fix it.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -3381,7 +3381,6 @@ class BaseBuilder
 	{
 		$this->resetRun([
 			'QBSelect'   => [],
-			'QBFrom'     => [],
 			'QBJoin'     => [],
 			'QBWhere'    => [],
 			'QBGroupBy'  => [],
@@ -3399,7 +3398,7 @@ class BaseBuilder
 		}
 
 		// Reset QBFrom part
-		if (is_string($this->tableName))
+		if (! empty($this->QBFrom) && is_string($this->tableName))
 		{
 			$this->from($this->tableName, true);
 		}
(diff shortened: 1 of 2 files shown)


kenjis:
It seems that how to reset FROM is not what I had in mind.

iRedds:
Maybe it would be more correct to use $this->getTable() instead of $this->tableName?
I dont know. 
I like this solution if we consider it within the current concept.

PR #4467, opened 2021-03-22
Language: getTranslationOutput: make method overridable by child classes

element-code:
Makes `getTranslationOutput` overridable by child classes.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Language/Language.php
+++ b/system/Language/Language.php
@@ -146,7 +146,7 @@ class Language
 	/**
 	 * @return array|string|null
 	 */
-	private function getTranslationOutput(string $locale, string $file, string $parsedLine)
+	protected function getTranslationOutput(string $locale, string $file, string $parsedLine)
 	{
 		$output = $this->language[$locale][$file][$parsedLine] ?? null;
 		if ($output !== null)


MGatner:
I don't see any problem with this. I believe @samsonasik wrote this part so I will defer to him.

element-code:
Okay @MGatner,

long story short, i want to do a low level implementation change in my CI4 based application.
In this case its only that the language resolve for `foo.bar.title` should not fall back to `foo.title`. Easy enough, just override & change the behavior of `getTranslationOutput`, any use case inside the language service is covered, updateability is perfectly fine.
But since `getTranslationOutput` is private, i can't. Now I have to overwrite all methods inside the language service with their original code just to be able to change the behavior of `getTranslationOutput`. The updateability is also bad bc now we have duplicated code which overrides the original code.

This is just a small use case, consider someone wants to implement caching or has a complete differing language string resolver.

MGatner:
Understandable! Since `Language` is offered as an existing Service I think we want to make sure it is as extensible as possible, so unless @samsonasik has a compelling reason why this should remain private I am in favor of this change.

samsonasik:
I am ok with this 👍

samsonasik (approved):

MGatner (approved):

element-code:
Thanks 👍

PR #4470, opened 2021-03-22
Add Model::initialize()

MGatner:
**Description**
I frequently find myself extending the constructor on module models to do some additional configuration. That makes for some awkward repeating of parameters, like:
```
	/**
	 * @param ConnectionInterface $db
	 * @param ValidationInterface $validation
	 */
	public function __construct(ConnectionInterface &$db = null, ValidationInterface $validation = null)
	{
		// Call the module constructor
		parent::__construct($db, $validation);
		
		// Do whatever I needed to do
	}
```

With the split into `BaseModel` and `Model` this gets even more uncomfortable, as additional extensions of `BaseModel` likely won't have the same constructor parameters.

This PR defines a new "event" during `BaseModel::__construct()` that checks for an `initModel()` method which will allow developers to handle model extensions without messing with the constructor itself. The inspiration came from `Controller` (though it's actually the opposite there since the contrsuctor is left for the developer).

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -97,14 +97,14 @@ class Model extends BaseModel
 	 */
 	public function __construct(ConnectionInterface &$db = null, ValidationInterface $validation = null)
 	{
-		parent::__construct($validation);
-
 		/**
 		 * @var BaseConnection $db
 		 */
 		$db = $db ?? Database::connect($this->DBGroup);
 
 		$this->db = &$db;
+
+		parent::__construct($validation);
 	}
 
 	/**
--- a/user_guide_src/source/models/model.rst
+++ b/user_guide_src/source/models/model.rst
@@ -73,6 +73,28 @@ that extends ``CodeIgniter\Model``::
 This empty class provides convenient access to the database connection, the Query Builder,
 and a number of additional convenience methods.
 
+Should you need additional setup in your model you may define an ``initModel()`` function
+which will be run at the very beginning of the Model's constructor. This allows you to perform
+extra steps without repeating the constructor parameters, for example extending other models::
+
+    <?php
+
+    namespace App\Models;
+
+    use Modules\Authentication\Models\UserAuthModel;
+
+    class UserModel extends UserAuthModel
+    {
+    	/**
+    	 * Called during initialization. Appends
+    	 * our custom field to the module's model.
+    	 */
+        protected function initModel()
+        {
+        	$this->allowedFields[] = 'middlename';
+        }
+    }
+
 Connecting to the Database
 --------------------------
 
--- a/user_guide_src/source/models/model.rst
+++ b/user_guide_src/source/models/model.rst
@@ -89,7 +89,7 @@ extra steps without repeating the constructor parameters, for example extending 
     	 * Called during initialization. Appends
     	 * our custom field to the module's model.
     	 */
-        protected function inititialize()
+        protected function initialize()
         {
         	$this->allowedFields[] = 'middlename';
         }
--- a/user_guide_src/source/models/model.rst
+++ b/user_guide_src/source/models/model.rst
@@ -73,8 +73,8 @@ that extends ``CodeIgniter\Model``::
 This empty class provides convenient access to the database connection, the Query Builder,
 and a number of additional convenience methods.
 
-Should you need additional setup in your model you may define an ``initModel()`` function
-which will be run at the very beginning of the Model's constructor. This allows you to perform
+Should you need additional setup in your model you may extend the ``inititialize()`` function
+which will be run immediately after the Model's constructor. This allows you to perform
 extra steps without repeating the constructor parameters, for example extending other models::
 
     <?php
@@ -89,7 +89,7 @@ extra steps without repeating the constructor parameters, for example extending 
     	 * Called during initialization. Appends
     	 * our custom field to the module's model.
     	 */
-        protected function initModel()
+        protected function inititialize()
         {
         	$this->allowedFields[] = 'middlename';
         }
--- a/user_guide_src/source/models/model.rst
+++ b/user_guide_src/source/models/model.rst
@@ -73,7 +73,7 @@ that extends ``CodeIgniter\Model``::
 This empty class provides convenient access to the database connection, the Query Builder,
 and a number of additional convenience methods.
 
-Should you need additional setup in your model you may extend the ``inititialize()`` function
+Should you need additional setup in your model you may extend the ``initialize()`` function
 which will be run immediately after the Model's constructor. This allows you to perform
 extra steps without repeating the constructor parameters, for example extending other models::
 
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -295,6 +295,11 @@ abstract class BaseModel
 	 */
 	public function __construct(ValidationInterface $validation = null)
 	{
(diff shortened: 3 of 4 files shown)


iRedds:
hm... magic?
```php
class MySuperModel extends Model
{
    public function __construct()
    {
        $this->initModel();
    }
   
    public function initModel(){}
}

class MyCasualModel extends MySuperModel
{
    public function initModel()
    {
    }
}
```

MGatner:
@iRedds I'm not sure I understand the point you are trying to make. You have removed the ability to pass parameters to your child model classes. Are you suggesting I add `initModel()` to the `BaseModel` class to avoid the call to `method_exists()`?

iRedds:
You can extend the default Model class with your base model class, which will call the initialization method in the constructor.

You are trying to replace the call to parent::__construct(); on $this->initModel(); 
Sorry, but I don't see any profit in this feature.

MGatner:
> You are trying to replace the call to parent::__construct(); on $this->initModel();

That's precisely what I am trying to do. As we move the framework towards multiple models (e.g. "SQL" and "NoSQL", but maybe others) I want to allow developers to disassociate their actual model implementation from the underlying framework structure. Already we have that "standard" Model with this constructor:
```
public function __construct(ConnectionInterface &$db = null, ValidationInterface $validation = null)
```
... and yet another simple extension of `BaseModel` may not have a constructor and thus use `BaseModel`'s signature:
```
public function __construct(ValidationInterface $validation = null)
```

This means any extending class will be reliant on the underlying implementation in order to do something very basic like initialize a configuration:
```
class MyModel extends Model
{
	public function __construct(ConnectionInterface &$db = null, ValidationInterface $validation = null)
	{
		parent::__contruct($db, $validation); // Won't work with other implementations of BaseModel

		$this->config = config('MyModelConfigurator');
	}
}
```

> You can extend the default Model class with your base model class

My current use case is the example in my original post, which is explicitly for extending module Models which precludes using an application base model.

I'm not set on this, my current method works fine, but I do think it is a useful addition and would like to hear more feedback. @najdanovicivan maybe you've run into situations where this would be helpful?

najdanovicivan:
@MGatner I like this. It will be usefull especially for setting validationRules, alowedField when there is a need to append additional rules/files to the parent value.

najdanovicivan:
@MGatner why indeed use method_exist. I'd rather go with adding empty initModel() method inside the BaseModel class as having it defined as empty there will provide hints for code completion and also PhpDoc can be written for the method so that documentation generation can pick it up as well.  

That way it can also be defined as protected method. Otherwise someone might not see it exist and is called in constructor so one can define it as public for some other use which may lead to various sorts of issues.

MGatner:
That would be fine. I was leaving it out so as not to suggest a signature or scope, leave those up to the developer, but I don't see any reason for it to be public and an empty signature would allow anything on child classes. I will change this later today.

MGatner:
@najdanovicivan Implemented.

paulbalandan (commented):
I would like to have the method called after all assignments and initializations in the constructor. This way child classes can still change the state of some properties already initialized by the constructor.

Also, can we make it simply as `initialize`?

MGatner:
> after all assignments and initializations in the constructor

As in, you would prefer this?
```
	public function __construct(ValidationInterface $validation = null)
	{
		$this->tempReturnType     = $this->returnType;
		$this->tempUseSoftDeletes = $this->useSoftDeletes;
		$this->tempAllowCallbacks = $this->allowCallbacks;

		/**
		 * @var Validation $validation
		 */
		$validation = $validation ?? Services::validation(null, false);

		$this->validation = $validation;

		$this->initModel();
	}
```

The only issue with that (and why I decided to go with it first) is that have a precedent for extended constructors calling their parent first. For example `Model::__construct()`:
```
	public function __construct(ConnectionInterface &$db = null, ValidationInterface $validation = null)
	{
		parent::__construct($validation);

		/**
		 * @var BaseConnection $db
		 */
		$db = $db ?? Database::connect($this->DBGroup);

		$this->db = &$db;
	}
```

We could certainly move `Model`'s call to parent last for consistency and leave it up to other extensions to do the same - I do like the idea of giving the developer "final dibs".

> Also, can we make it simply as `initialize`?

I have no connection to the name `initModel()`, I chose it for consistency with `initController()` but as Is stated above that is actually the mirror opposite approach (`__construc(t()` is left for the developer). If there are no other preferences I will change it to `initialize()`.

MGatner:
Test failure is unrelated.

paulbalandan:
My contention for calling `$this->initialize()` as last in the queue is because this serves as a "post" processor for the internal state of the child model, somehow like a hook. We call this after all the works in the parent constructor to do some finishing touches. The call to `parent::__construct` is another case where we want to initialize first the parent because it may doing some initializations that we will need later in our own initialization. For example if the parent initializes the `$db` property and I'll gonna be using that later for setting up a db prefix, then I would be needing to call first `parent::__construct`.

user_guide_src/source/models/model.rst
```diff
@@ -73,6 +73,28 @@ that extends ``CodeIgniter\Model``::
 This empty class provides convenient access to the database connection, the Query Builder,
 and a number of additional convenience methods.
 
+Should you need additional setup in your model you may extend the ``inititialize()`` function
```
paulbalandan, line 76:
```suggestion
Should you need additional setup in your model you may extend the ``initialize()`` function
```

system/Model.php
```diff
@@ -97,14 +97,14 @@ class Model extends BaseModel
...
-		parent::__construct($validation);
-
 		/**
 		 * @var BaseConnection $db
 		 */
 		$db = $db ?? Database::connect($this->DBGroup);
 
 		$this->db = &$db;
+
+		parent::__construct($validation);
```
paulbalandan, line 107: This can be reverted to its original place.

user_guide_src/source/models/model.rst
```diff
@@ -73,6 +73,28 @@ that extends ``CodeIgniter\Model``::
...
+
+    use Modules\Authentication\Models\UserAuthModel;
+
+    class UserModel extends UserAuthModel
+    {
+    	/**
+    	 * Called during initialization. Appends
+    	 * our custom field to the module's model.
+    	 */
+        protected function inititialize()
```
paulbalandan, line 92:
```suggestion
        protected function initialize()
```

paulbalandan (commented):
Some typo errors only.

system/Model.php
MGatner: I think this needs to stay last. Even your example above requires it here:

> For example if the parent initializes the $db property and I'll gonna be using that later for setting up a db prefix

Otherwise `initialize()` won't have access to `$this->db`. Call stack should look like:
1. `Model::__construct()`
2. `BaseModel::__construct()`
3. `MyModel::initialize()`

That way the developer gets the very last line in the stack so has the most properties available.

MGatner:
Oops! Thanks for catching the typos. See my comment on the placement of `parent::__construct()`, but otherwise I think this is ready to go.

paulbalandan (approved):

samsonasik (approved):

PR #4476, opened 2021-03-23
Implementing subqueries in FROM section

iRedds:
**Description**
I am waiting feedback of the community, so I see no reason to add tests and descriptions to UG right now.

**Added BaseConnection::raw() method.** 
It expects a raw SQL subquery as an argument and returns an instance of the SqlExpression class. 
This is to prevent processing subqueries as a simple string.

**BaseConnection::table() and BaseBuilder::from() methods**.
Now takes as an argument Closure, BaseBuilder, SqlExpression, array, string, null.

```php
//Closure 
$this->db->table(function(BaseBuilder $builder) { /* subquery */  });

// BaseBuilder 
$subquery = $this->db->table('user');
$this->db->table($subquery);

// SqlExpression
$this->db->table($this->db->raw('SELECT * FROM user'));

// new instance of BaseBuilder without dafeult table 
// Thanks to the designers. Because it is impossible to correctly create a clean instance of the class without BCs. 
$this->db->table(null);
```
This solution makes it possible to use a wrapper for the main query. 
```sql
SELECT * FROM (main query)
```
and will allow better implementation of support for UNION queries. 

Query builder methods that previously took Closure as an argument now take BaseBuilder and SqlExpression.
```php
$builder->where('id', $this->db->raw('SELECT id FROM user'));
```

**Help wanted**
I also want to ask about this "problem".
The problem starts with tests (SQLSRV). We get an instance of the query builder. 
```php
$this->db = new MockConnection(['DBDriver' => 'SQLSRV', 'database' => 'test', 'schema' => 'dbo']);
$builder = new SQLSRVBuilder('user', $this->db);
```
If ``$this->db->table(null)`` is called inside the query builder, then the ``MockBuilder`` instance will be retrieved, not the ``SQLSRVBuilder`` instance.
And the subquery will not be generated correctly.

There are several solutions. 
Use ``new static(null, $this->db)`` instead of ``$this->db->table(null)`` , but phpstan says: ``Unsafe usage of new static()``.
Or use the SQLSRV driver's Coonnection class instead of the MockConnection class for testing. 
```php
$this->db = new Connection(['DBDriver' => 'SQLSRV', 'database' => 'test', 'schema' => 'dbo']);
```
But I'm not sure if replacing the connection class is correct. 

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -251,14 +251,14 @@ class BaseBuilder
 	/**
 	 * Constructor
 	 *
-	 * @param  string|array        $tableName
-	 * @param  ConnectionInterface $db
-	 * @param  array               $options
+	 * @param  Closure|BaseBuilder|SqlExpression|array|string|null $tableName
+	 * @param  ConnectionInterface 							       $db
+	 * @param  array               							       $options
 	 * @throws DatabaseException
 	 */
 	public function __construct($tableName, ConnectionInterface &$db, array $options = null)
 	{
-		if (empty($tableName))
+		if (empty($tableName) && (is_array($tableName) || is_string($tableName)))
 		{
 			throw new DatabaseException('A table must be specified when creating a new Query Builder.');
 		}
@@ -266,7 +266,11 @@ class BaseBuilder
 		/** @var BaseConnection $db */
 		$this->db = $db;
 
-		$this->tableName = $tableName;
+		if (is_string($tableName))
+		{
+			$this->tableName = $tableName;
+		}
+
 		$this->from($tableName);
 
 		if (! empty($options))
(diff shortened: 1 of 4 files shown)


kenjis:
> If $this->db->table(null) is called inside the query builder, then the MockBuilder instance will be retrieved, not the SQLSRVBuilder instance.

It seems the `$this->db` should be an object to return `SQLSRVBuilder`.
And `MockConnection` does not behave like that. It is not a correct mock.

If you can write tests with the real SQLSRV `Connection` without a real SQL Server, it is no problem.
If not, it is better to create a new MockConnection class like `SQLSRV\MockConnection`.
Or you could use PHPUnit mock functionality to create a mock:
https://phpunit.readthedocs.io/en/9.5/test-doubles.html

https://github.com/codeigniter4/CodeIgniter4/blob/fed0254a06e5f104b2ff642df4f78632d70bd171/tests/system/HTTP/Cookie/CookieStoreTest.php#L107-L110

lonnieezell:
Seems to be overridden by #5510 

Closing.

Issue #4297, opened 2021-02-17
Bug: can't change and override valid locales


davislasis:
**Problem**
System does not change incoming request locale with overriding it from outer source (not App config, like database) using filter approach

**Describe the bug**
- all my routes has {locale} in front as first url segment 'en/dashboard', 'lv/dashboard' etc
- supported locales are stored in local mysql database and managed through admin area
- my project will populate / manage languages only through database and admin area, no need of App config $supportedLocales as client will not change and re-commit files
- to avoid caching locale change in frontend, i have setup mainFilter which caches potential locale change at any point (route) within the system
- IncomingRequest class is initialised before filters, so if my filter override supportedLocales - it does not change in further system loding

**CodeIgniter 4 version**
4.1.1

Solution 1:
- IncomingRequest class - variable `$validLocales` change from `protected` to `public`
- this way it is possible to override values if needed

**Affected module(s)**
/system/HTTP/IncomingRequest.php

[https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L99](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L99)

Solution 2:
- as App config does not have `$validLocales` variable
- as IncomingRequest class - variable `$validLocales` is not heavily used elsewhere
- as IncomingRequest class controller mandatory requests App config as first method param
- remove variable `$validLocales` from IncomingRequest class
- as IncomingRequest class method setLocale() uses `$validLocales`, replace it with already loaded `$this->config->supportedLocales`

**Affected module(s)**
/system/HTTP/IncomingRequest.php

remove: [https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L99](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L99)

remove: [https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L162](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L162)

update: [https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L225](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L225)

```php

	/**
	 * Sets the locale string for this request.
[...]

PR #4478, opened 2021-03-23
Bug: can't change and override valid locales #4297

davislasis:
[Bug: can't change and override valid locales #4297](https://github.com/codeigniter4/CodeIgniter4/issues/4297)

**Description**
Fix IncomingRequest ValidLocales

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -221,7 +221,7 @@ class IncomingRequest extends Request
 	{
 		// If it's not a valid locale, set it
 		// to the default locale for the site.
-		if (! in_array($locale, $this->validLocales, true))
+		if (! in_array($locale, $this->config->supportedLocales, true))
 		{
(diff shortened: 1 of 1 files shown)


MGatner:
`IncomingRequest::$validLocales` is already defined as `\Config\App::$supportedLocales`: https://github.com/codeigniter4/CodeIgniter4/blob/b3ebd3da92b9da3971fb02d912f3923939b9c008/system/HTTP/IncomingRequest.php#L161

If this fix is working it is because your `\Config\App` instance is being updated between the constructor and the actual call, but that creates an inconsistent external dependency we don't want to introduce. I will look at the rest of your content on the Issue, but you may just need to make sure `\Config\App` is initialized correctly with a [Registrar](https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#registrars) or something prior to `IncomingRequest::__construct()`.

PR #4479, opened 2021-03-23
Suppress PhpStorm warning to `db` in `Model`

### LINKED ISSUE #4474: Bug: PHPStorm anlysis fault
**Direction**

**Describe the bug**
phpStorm is flagging these to lines in the Model as ERRORS.
$this->db says to make it public there is no $db in the top values

**CodeIgniter 4 version**
Newest Development version today.

**Affected module(s)**
system/Model

Should have no errors

**Context**
 - OS: Windows 10 x64 20H2
 - Web server Apache 2.3.46
 - PHP version 8.0.3

`	public function __construct(ConnectionInterface &$db = null, ValidationInterface $validation = null)
	{
		parent::__construct($validation);

		/**
		 * @var BaseConnection $db
		 */
		$db = $db ?? Database::connect($this->DBGroup);

		$this->db = &$db; <-- ERROR
	}
`
And here.

`	public function builder(?string $table = null)
	{
		// Check for an existing Builder
		if ($this->builder instanceof BaseBuilder)
		{
			// Make sure the requested table matches the builder
			if ($table && $this->builder->getTable() !== $table)
			{
				return $this->db->table($table);
			}

			return $this->builder;
		}

		// We're going to force a primary key to exist
		// so we don't have overly convoluted code,
		// and future features are likely to require them.
		if (empty($this->primaryKey))
		{
			throw ModelException::forNoPrimaryKey(static::class);
		}

		$table = empty($table) ? $this->table : $table;

		// Ensure we have a good db connection
		if (! $this->db instanceof BaseConnection)
		{
			$this->db = Database::connect($this->DBGroup); <-- ERROR
		}

		$builder = $this->db->table($table);

		// Only consider it "shared" if the table is correct
		if ($table === $this->table)
		{
			$this->builder = $builder;
		}

		return $builder;
	}
`

Says to make them public but if you look at the top of the Model there is no $db defined.

kenjis:
**Description**
Fixes #4474

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -90,6 +90,11 @@ class Model extends BaseModel
 	protected $escape = [];
 
 	/**
+	 * @var BaseConnection
+	 */
+	protected $db;
+
+	/**
 	 * Model constructor.
 	 *
 	 * @param ConnectionInterface|null $db         DB Connection
--- a/system/Model.php
+++ b/system/Model.php
@@ -90,11 +90,6 @@ class Model extends BaseModel
 	protected $escape = [];
 
 	/**
-	 * @var BaseConnection
-	 */
-	protected $db;
-
-	/**
 	 * Model constructor.
 	 *
 	 * @param ConnectionInterface|null $db         DB Connection
--- a/system/Model.php
+++ b/system/Model.php
@@ -40,7 +40,8 @@ use ReflectionProperty;
  *      - allow intermingling calls to the builder
  *      - removes the need to use Result object directly in most cases
  *
- * @mixin BaseBuilder
+ * @mixin    BaseBuilder
+ * @property BaseConnection $db
  */
 class Model extends BaseModel
 {


samsonasik (approved):

iRedds:
But the property is already defined in BaseModel.

InsiteFX:
Yes it is defined in the BaseModel, but phpStorm is flagging it as two errors and flagging  Model.php line with red error.

phpStorm is saying to make the property public instead of protected.

kenjis:
@iRedds @InsiteFX Thanks.
It seems the warning is caused by `@mixin`.

MGatner:
I'm open to the communal opinion, but I don't like this. This seems to be a bug with PHPStorm - the `protected $db` property is available on `BaseModel` and publicly accessible via `__get()`, but for some reason it is giving the `@mixin` of `BaseBuilder` precedence, where `protected $db` also exists but has no magic getter.

IMO people passionate about PHPStorm should take this as a bug report to their product. It should be easy to recreate the bug with a barebones repo demonstrating the problem.

kenjis:
@MGatner You have a point. But I don't know the definition of `@mixin`.
So I don't know it is a bug or not.

It seems `@mixin` assumes two classes have no parent-child relationship.

MGatner:
> When a class delegates unknown method calls and property accesses to a different class using `__call` and `__get`/`__set`, we can describe the relationship using @mixin PHPDoc tag

https://phpstan.org/writing-php-code/phpdocs-basics#mixins

MGatner:
Can anyone else here who had confirmed the issue please check against the latest build of PHPStorm? InsiteFX believes the issue may be fixed and this PR no longer needed.

iRedds:
PhpStorm 2020.3.3
Build #PS-203.7717.64, built on March 16, 2021

MGatner:
Okay, not sure this is worth all the effort. Can someone confirm that this PR solves the problem? I believe `@property` automatically includes `@property-read` so should fix this problem.

InsiteFX:
MGatner, I opened a new issue because I thought this one was closed, please delete the other one.

I have contacted phpStorm support and I' am waiting to here from them. The ERRORS came back in today's  newest build it's strange because yesterday's build it went away but now it's back.

lonnieezell:
I have had similar issues with PHPStorm in my day job with Laravel projects. No rhyme or reason but it will occasionally decide it doesn't like something in the file even though the same thing is done many other times throughout the app. I just ignore it these days....

MGatner:
To be fair: we are hard on static analysis. We have Model with its own magic getter and mixins to BaseBuilder and BaseConnection (from BaseModel), and BaseBuilder has its own magic getter and mixin to BaseConnection which... guess what? has a magic getter. 

I think when Lonnie rewrites the database layer before the end of this month 🙃 he should remove most if not all of the mixins and stick to local accessors.

MGatner:
We can remove this any time so I'm just going to try it.

PR #4484, opened 2021-03-24
Feature Response null

MGatner:
**Description**
Our `FeatureResponse` constructor allows `null` input but very clearly relies on having a valid `ResponseInterface`. ~~For now this PR provides a fallback when `null` is provided to prevent breaking changes but IMO the parameter should be made mandatory at some point.~~ This PR removes the `null` option.

Along those lines, this is part of some bigger test response changes I have in the works but I didn't want to bundle this bugfix with that PR.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/FeatureResponse.php
+++ b/system/Test/FeatureResponse.php
@@ -42,11 +42,11 @@ class FeatureResponse extends TestCase
 	 *
 	 * @param ResponseInterface $response
 	 */
-	public function __construct(ResponseInterface $response = null)
+	public function __construct(ResponseInterface $response)
 	{
 		$this->response = $response;
 
-		$body = $response->getBody();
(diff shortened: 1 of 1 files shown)


samsonasik:
It is bug, `ResponseInterface` must be mandatory or `$response->getbody()` will error.

MGatner:
@samsonasik I take your feedback as support for updating the signature, so I've reimplemented for that. I doubt anyone is extending this class and, if so, unlikely for it to be with a duplicate constructor. Given that this is all the test namespace I'm okay with some more leniency with bug fixes like this.

samsonasik (approved):

PR #4486, opened 2021-03-24
Test Response

MGatner:
**Description**
`ControllerResponse` is quite lacking compared to `FeatureResponse`, yet both have a number of identical methods and properties. This PR consolidates them into a new class `TestResponse`, which adds a number of features to both classes as well as deprecating others that are no longer needed.

I left the previous classes in place so no tests should be affected. A future PR should have `ControllerTester` and `FeatureTestTrait` switch to returning `TestResponse` instead, but likely that will be version 5.

Note: This PR relies on #4484

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide


MGatner:
Failure is unrelated - another "close timing" error.

lonnieezell (approved):

Issue #4381, opened 2021-03-02
Bug: IncomingRequest.php getVar() 


lenuse:
**Direction**
The getVar method cannot get $_REQUEST under a GET request 

**Describe the bug**
When the http content-type is json, send a get request, the request parameters cannot be obtained, and the program issues an exception 
![C24864D6-8178-4fe4-AA76-933DCB50A2A9](https://user-images.githubusercontent.com/12166648/109593164-71fb7280-7b4b-11eb-9a68-d7906a05aded.png)

**CodeIgniter 4 version**
4.1.1

**Affected module(s)**
getVar method

**Expected behavior, and steps to reproduce if appropriate**
When $this->body is null, you should get the value in $_REQUEST 

**Context**
 - OS:  Windows 10
 - Web server  nginx all
 - PHP version 7.4.3

luispastendev:
When you receive json content you can get it with `$this->request->getJson(true)` which will get the content and convert it into an associative array with that you could access the values of your request 

```php
$body = $this->request->getJSON(true);
print_r($body['param']);
```

lenuse:
```js
$.ajax({
   url: "/api/products?type=1&page=1&limit=20",
    headers: {
        Accept: "application/json; charset=utf-8"
    },
    type: "get",
    success: function (data) {
    }
});
```
```php
class products extends \CodeIgniter\Controller
{
    // url is /api/products
    public function products() {
        $this->request->getVar('limit');
    }
}

```
I want to use `$this->request->getVar('limit')` but but the program throws an exception。

luispastendev:
@lenuse what I see what you want to get is actually a query string not a json, `getVar` is not going to serve you for that, what you should do is do something like the following:

```php
parse_str($this->request->uri->getQuery(),$output);
echo $output['limit'];
```

see: 
- https://codeigniter.com/user_guide/libraries/uri.html#query
- https://www.php.net/manual/es/function.parse-str.php

lenuse:
In version 4.0.4, using the getVar() method to obtain query string parameters is allowed. But after adding the content-type judgment, the method is incompatible. And 4.0.5 is a minor version update, this should not happen. I have now updated the code of my production environment.

lenuse:
If it is not compatible, please close this bug

luispastendev:
you can close the issue or some admin to do it @paulbalandan

luispastendev:
[...]

PR #4493, opened 2021-03-26
The getVar method is compatible with version 4.0.4 

lenuse:
The getVar method is compatible with version 4.0.4 #4381 

**Description**
In the current version, when the request is the get method and the content-type is json, the $_REQUEST parameter cannot be obtained 

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

**Notes**
- Increase the length of the judgment body 
- Add unit tests under json and get methods

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -281,7 +281,7 @@ class IncomingRequest extends Request
 	//--------------------------------------------------------------------
 
 	/**
-	 * Fetch an item from the $_REQUEST object or a JSON input stream. This is the simplest way
+	 * Fetch an item from JSON input stream with fallback to $_REQUEST object. This is the simplest way
 	 * to grab data from the request object and can be used in lieu of the
 	 * other get* methods in most cases.
 	 *
@@ -312,8 +312,7 @@ class IncomingRequest extends Request
 
 			return $this->getJsonVar($index, false, $filter, $flags);
 		}
-		$output = $this->fetchGlobal('request', $index, $filter, $flags);
-		return is_null($index) ? (object)$output : $output;
+		return $this->fetchGlobal('request', $index, $filter, $flags);
 	}
 
 	//--------------------------------------------------------------------
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -293,7 +293,7 @@ class IncomingRequest extends Request
 	 */
 	public function getVar($index = null, $filter = null, $flags = null)
 	{
-		if (strpos($this->getHeaderLine('Content-Type'), 'application/json') !== false)
+		if (strpos($this->getHeaderLine('Content-Type'), 'application/json') !== false && !is_null($this->body))
 		{
 			if (is_null($index))
 			{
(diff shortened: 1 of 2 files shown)


system/HTTP/IncomingRequest.php
```diff
@@ -293,7 +293,7 @@ public function isSecure(): bool
 	 */
 	public function getVar($index = null, $filter = null, $flags = null)
 	{
-		if (strpos($this->getHeaderLine('Content-Type'), 'application/json') !== false)
+		if (strpos($this->getHeaderLine('Content-Type'), 'application/json') !== false && !is_null($this->body))
```
MGatner, line 296:
```suggestion
		if (strpos($this->getHeaderLine('Content-Type'), 'application/json') !== false && ! is_null($this->body))
```

tests/system/HTTP/IncomingRequestTest.php
```diff
@@ -383,6 +383,33 @@ public function testGetVarWorksWithJson()
...
+		$config = new App();
+		$config->baseURL = 'http://example.com/';
+		// get method
+		$uri = new URI('http://example.com/path?foo=bar&fizz=buzz');
+		$_REQUEST['foo'] = 'bar';
+		$_REQUEST['fizz'] = 'buzz';
+		$request = new IncomingRequest($config, $uri, 'php://input', new UserAgent());
+		$request = $request->withMethod('GET');
+		// json type
+		$request->setHeader('Content-Type', 'application/json');
```
MGatner, line 397:
```suggestion
		$config          = new App();
		$config->baseURL = 'http://example.com/';

		// GET method
		$_REQUEST['foo']  = 'bar';
		$_REQUEST['fizz'] = 'buzz';

		$request = new IncomingRequest($config, new URI('http://example.com/path?foo=bar&fizz=buzz'), 'php://input', new UserAgent());
		$request = $request->withMethod('GET');

		// JSON type
		$request->setHeader('Content-Type', 'application/json');
```

MGatner (changes requested):
This looks good. A few style changes. I would also like @luispastendev to take a look since he was involved in the original discussion, as well as anyone else on the team who might know more about this than me.

luispastendev:
@MGatner , take a look at a previous installation of codeigniter and observe that the getVar code was touched to support JSON requests, in this case I suppose it is fine because if the empty body was sent it generated an error.

On the other hand browsing the internet to the RFCs 7230-7237 specification nothing is mentioned about not sending body by get, in my opinion I suppose it is fine.

See the highest rated answer:
https://stackoverflow.com/questions/978061/http-get-with-request-body

samsonasik (approved):

paulbalandan (approved):

samsonasik:
Thank you @lenuse

iRedds:
I believe that if an empty body is sent with a Content-Type: application/json header then an exception should be thrown.

lenuse:
> I believe that if an empty body is sent with a Content-Type: application/json header then an exception should be thrown.

The getVar method is to get the json body or $_RESQUEST parameter. If the body is empty, an exception is returned and $_REQUEST will not be obtained. If the request API does not use the resetful style, the value of $_REQUEST will be used.

shantiwebsolution:
is this merged with master ? do I get updated code via composer update ?

kenjis:
@shantiwebsolution 
Yes.
Please ask support questions on our [forum](https://forum.codeigniter.com/forum-30.html).

PR #4497, opened 2021-03-26
Tweaks to TestResponse

MGatner:
**Description**
This is a slight "addendum" to the `TestResponse` PR. Fixes a few things:
1. `assertSessionHas()` uses a nice display for failed assertions, however `$value` is not always a string which can lead to errors like "array to string conversion". This PR limits the value display to when it is actually a string. You still get the PHPUnit comparison, e.g.:
```
The value of 'session-var' does not match expected value.
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Array (
-        'class' => 'foobar'
+        'class' => 'success'
         'text' => 'New factory created successfully.'
     )
 )
```
2. `assertJSONExact()` accepts an `array` which it will automatically convert to JSON which is very handy for testing RESTful responses. This PR expands that to accept objects (which will be converted to arrays) for endpoints that may return a singleton. Previously this would cause an error:
> TypeError: Argument 1 passed to PHPUnit\Framework\Assert::assertJsonStringEqualsJsonString() must be of the type string, object given, called in /opt/CodeIgniter4/system/Test/TestResponse.php on line 439
3. `isRedirect()` was incorrectly grouped with the statuses - this PR moves it to the Redirects section.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/TestResponse.php
+++ b/system/Test/TestResponse.php
@@ -142,18 +142,6 @@ class TestResponse extends TestCase
 	}
 
 	/**
-	 * Returns whether or not the Response was a redirect or RedirectResponse
-	 *
-	 * @return boolean
-	 */
-	public function isRedirect(): bool
-	{
-		return $this->response instanceof RedirectResponse
-			|| $this->response->hasHeader('Location')
-			|| $this->response->hasHeader('Refresh');
-	}
-
-	/**
 	 * Asserts that the status is a specific value.
 	 *
 	 * @param integer $code
@@ -188,6 +176,18 @@ class TestResponse extends TestCase
 	//--------------------------------------------------------------------
 	// Redirection
 	//--------------------------------------------------------------------
+
+	/**
+	 * Returns whether or not the Response was a redirect or RedirectResponse
+	 *
+	 * @return boolean
+	 */
+	public function isRedirect(): bool
+	{
+		return $this->response instanceof RedirectResponse
+			|| $this->response->hasHeader('Location')
+			|| $this->response->hasHeader('Refresh');
+	}
 
 	/**
 	 * Assert that the given response was a redirect.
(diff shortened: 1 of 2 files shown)


system/Test/TestResponse.php
```diff
@@ -272,10 +272,19 @@ public function assertSessionHas(string $key, $value = null)
 	{
 		$this->assertTrue(array_key_exists($key, $_SESSION), "'{$key}' is not in the current \$_SESSION");
 
-		if ($value !== null)
+		if (is_null($value))
+		{
+			return;
+		}
+
+		if (is_string($value))
```
paulbalandan, line 280: Why not use `var_export` on`$value` to make it usable on the message?

MGatner: The first iteration had that but it just ends up duplicating the display and looks messy. I will paste my example result from that iteration.

MGatner: I found this made it difficult to determine what the actual message was:
```
There was 1 failure:

1) PresenterWriteTest::testCreate
The value of 'alerts-queue' (Array ( [0] => Array ( [class] => foobar [text] => New factory created successfully. ) ) ) does not match expected value.
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Array (
-        'class' => 'foobar'
+        'class' => 'success'
         'text' => 'New factory created successfully.'
     )
 )
```

paulbalandan: Okay. Then how about using is_scalar to accommodate int, floats?

MGatner: That's a great idea! I'm mobile for the next 12 hours but if someone can suggest it I can commit the change (or you can).

MGatner: Need to confirm as well whether an object with `__toString()` will pass `is_scalar()`. If not we might want to check both scalar and string.

MGatner: Nevermind, `is_string` and `is_scalar` both always fail on objects - you have to check `is_object` and `method_exists('__toString')` to support that.

paulbalandan (approved):

PR #4499, opened 2021-03-27
Fixes generator not preserving letter case

### LINKED ISSUE #4495: Bug: Differences in file names created with  CLI command
**Describe the bug**
Hi, @paulbalandan
I use CLI for create, modal, seed and ...
When I give the name as **TestModal** but the corresponding file with the name **Testmodal** is created in the directories.

**CodeIgniter 4 version**
4.1.1

**Affected module(s)**
CLI

step 1 : php spark make:modal
Step 2 : **TestModal**
Step 3 : Files are created  but named **Testmodal**.
This problem exists for other commands(make:seeder ,make:filter and ...) as well. 

**Context**
 - OS: [Windows 10]
 - Web server [Apache 2.0 Handler]
 - PHP version [8.0.0]

paulbalandan:
**Description**
Fixes #4495 

@datamweb can you test this solution?

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/GeneratorTrait.php
+++ b/system/CLI/GeneratorTrait.php
@@ -212,17 +212,24 @@ trait GeneratorTrait
 		{
 			// @codeCoverageIgnoreStart
 			$nameLang = $this->classNameLang ?: 'CLI.generator.className.default';
-
-			$class = CLI::prompt(lang($nameLang), null, 'required');
+			$class    = CLI::prompt(lang($nameLang), null, 'required');
 			CLI::newLine();
 			// @codeCoverageIgnoreEnd
 		}
 
 		helper('inflector');
 
-		$component = strtolower(singular($this->component));
-		$class     = strtolower($class);
-		$class     = strpos($class, $component) !== false ? str_replace($component, ucfirst($component), $class) : $class;
+		$component = singular($this->component);
+
+		/**
+		 * @see https://regex101.com/r/a5KNCR/1
+		 */
+		$pattern = sprintf('/([a-z][a-z0-9_\/\\\\]+)(%s)/i', $component);
+
+		if (preg_match($pattern, $class, $matches) === 1)
+		{
+			$class = $matches[1] . ucfirst($matches[2]);
+		}
 
 		if ($this->enabledSuffixing && $this->getOption('suffix') && ! strripos($class, $component))
 		{
--- a/tests/system/Commands/CommandGeneratorTest.php
+++ b/tests/system/Commands/CommandGeneratorTest.php
@@ -5,7 +5,10 @@ namespace CodeIgniter\Commands;
 use CodeIgniter\Test\CIUnitTestCase;
 use CodeIgniter\Test\Filters\CITestStreamFilter;
 
-class CommandGeneratorTest extends CIUnitTestCase
+/**
+ * @internal
+ */
+final class CommandGeneratorTest extends CIUnitTestCase
 {
 	protected $streamFilter;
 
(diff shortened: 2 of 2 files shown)


system/CLI/GeneratorTrait.php
```diff
@@ -212,17 +212,20 @@ protected function qualifyClassName(): string
...
 			// @codeCoverageIgnoreEnd
 		}
 
 		helper('inflector');
 
-		$component = strtolower(singular($this->component));
-		$class     = strtolower($class);
-		$class     = strpos($class, $component) !== false ? str_replace($component, ucfirst($component), $class) : $class;
+		$component = singular($this->component);
+		$pattern   = sprintf('/([a-z][a-z0-9_\/\\\\]+)(%s)/i', singular($this->component));
```
samsonasik, line 223: better to have https://regex101.com link as usual.

system/CLI/GeneratorTrait.php
```diff
@@ -212,17 +212,20 @@ protected function qualifyClassName(): string
 		{
 			// @codeCoverageIgnoreStart
 			$nameLang = $this->classNameLang ?: 'CLI.generator.className.default';
```
datamweb, line 214: why **CLI.generator.className.default** !! Do you mean  **CLI.generator.className** ??

paulbalandan: No. The generator class names have been modified to be specific to the component they are creating. For example, in `make:model` if you are missing the class input you will be prompted with `Model class name:` instead of the generic `Class name`. `CLI.generator.className.default` is the lang string for the `Class name` input.

datamweb:
It works, but I have a question?
I wrote a custom command to create a helper file.
System helpers are created with the names array_helper or cookie_helper and ...
So I tend to create a personal helper in the same way(myhelper_test).

### Custom command created by me: make:helper
step1 : php spark make: helper
step2: and then enter a new name. For example **myhelper_test**
Result : my helper is created **But** Helper is called **MyhelperTest**.
I think it is better to stop this problem now.
what is your opinion?

helperName

paulbalandan:
This behavior is intended. This uses the `pascalize` function used internally by `GeneratorTrait` to change snake_case to PascalCase. This is because generators are designed to be used with "class files", those files containing custom PHP classes. It is pascalized to conform with PSR-2 or PSR-12, I think.

If you will be creating your own generator on "non-class files", I suggest not to use the class name input to provide the name of your helper as it will be pascalized. A way to do this is to supply the helper name as an option value. So you can be doing something like this:
```
php spark make:helper --name my_custom_helper

File created: APPPATH\Helpers\my_custom_helper.php
```

A little side note: custom helper files should be ending in `_helper` in order to be loaded when called via `helper('my_custom_helper')`

system/CLI/GeneratorTrait.php
datamweb: Ok.
Do not need to create a string className.default in the system \ Language \ en \CLI.php file?

datamweb:
> This behavior is intended. This uses the `pascalize` function used internally by `GeneratorTrait` to change snake_case to PascalCase. This is because generators are designed to be used with "class files", those files containing custom PHP classes. It is pascalized to conform with PSR-2 or PSR-12, I think.
> 
> If you will be creating your own generator on "non-class files", I suggest not to use the class name input to provide the name of your helper as it will be pascalized. A way to do this is to supply the helper name as an option value. So you can be doing something like this:
> 
> ```
> php spark make:helper --name my_custom_helper
> 
> File created: APPPATH\Helpers\my_custom_helper.php
> ```
> 
> A little side note: custom helper files should be ending in `_helper` in order to be loaded when called via `helper('my_custom_helper')`

Thank you very much for your comprehensive answer. I was fully justified.

PR #4502, opened 2021-03-28
Relocate Cookie

mostafakhudair:
Moving cookie class to main system folder, while its a single component and doesn't depends on http class

- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Cookie/Cookie.php
+++ b/system/Cookie/Cookie.php
@@ -9,7 +9,7 @@
  * file that was distributed with this source code.
  */
 
-namespace CodeIgniter\HTTP\Cookie;
+namespace CodeIgniter\Cookie;
 
 use ArrayAccess;
 use CodeIgniter\HTTP\Exceptions\CookieException;
--- a/system/Common.php
+++ b/system/Common.php
@@ -11,12 +11,12 @@
 
 use CodeIgniter\Cache\CacheInterface;
 use CodeIgniter\Config\Factories;
+use CodeIgniter\Cookie\Cookie;
+use CodeIgniter\Cookie\CookieStore;
 use CodeIgniter\Database\BaseConnection;
 use CodeIgniter\Database\ConnectionInterface;
 use CodeIgniter\Debug\Timer;
 use CodeIgniter\Files\Exceptions\FileNotFoundException;
-use CodeIgniter\HTTP\Cookie\Cookie;
-use CodeIgniter\HTTP\Cookie\CookieStore;
 use CodeIgniter\HTTP\Exceptions\CookieException;
 use CodeIgniter\HTTP\Exceptions\HTTPException;
 use CodeIgniter\HTTP\RedirectResponse;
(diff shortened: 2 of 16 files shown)


system/Cookie/CloneableInterface.php
```diff
@@ -9,15 +9,15 @@
...
+namespace CodeIgniter\Cookie;
 
 use DateTimeInterface;
 
 /**
  * Interface for a fresh Cookie instance with selected attribute(s)
  * only changed from the original instance.
  */
-interface CloneableCookieInterface extends CookieInterface
+interface CloneableInterface extends CookieInterface
```
paulbalandan, line 20: Please revert the name. `CloneableInterface` feels incomplete. Cloneable what? The current name is desired.

system/Cookie/Cookie.php
```diff
@@ -36,7 +36,7 @@
  * $cookie2->getName(); // prod_cookie
  * ```
  */
-class Cookie implements ArrayAccess, CloneableCookieInterface
+class Cookie implements ArrayAccess, CloneableInterface
```
paulbalandan, line 39: See comment above

paulbalandan (commented):
My contention for grouping the Cookie under the HTTP is simply because it is an HTTP concept. It is the representation of an _HTTP_ cookie. But if others would like to see the cookie in a separate folder, I'll gladly oblige.

Other comment is on renaming of `CloneableCookieInterface`. See below.

MGatner (commented):
I like this! HTTP layer is bloated, and cleaning it up as much as possible is a good thing. I agree with the interface name but otherwise his looks great.

MGatner (changes requested):
In addition to the changes already noted please recreate the current classes as deprecated extensions of the new ones.

paulbalandan:
> I like this! HTTP layer is bloated, and cleaning it up as much as possible is a good thing. I agree with the interface name but otherwise his looks great.
> 
> > simply because it is an HTTP concept
> 
> @paulbalandan I understand the interest but I think we have precedence in this with a number of other components (like Honeypot, Security, Throttle...). With the goal of moving towards PSR-7 compliance I think having HTTP stick to PSR components would be a logical separation.

Hmm. Ok then. We'll have it as a separate component. Aside from that, there's no need to deprecate as this is in `>4.1.1`. Just change the changelog entry. I think the namespace there is written as `HTTP\Cookie`

system/Cookie/CloneableInterface.php
mostafakhudair: Cloneable just like Countable Interface, But i agree it looks incomplete

mostafakhudair:
@MGatner CloneableInterface or CloneableCookieInterface?

MGatner:
I vote for `CloneableCookieInterface` as it is clearer what the interface is for.

paulbalandan (approved):

PR #4508, opened 2021-03-30
Create Config::Cookie Class

mostafakhudair:
While we are taking a step forward to make config folder separated as much as possible, Cookie class should have it's own config file

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -3,7 +3,6 @@
 namespace Config;
 
 use CodeIgniter\Config\BaseConfig;
-use DateTimeInterface;
 
 class App extends BaseConfig
 {
@@ -242,6 +241,8 @@ class App extends BaseConfig
 	 * Set a cookie name prefix if you need to avoid collisions.
 	 *
 	 * @var string
+	 * 
+	 * @deprecated use Config\Cookie::$prefix property instead.
 	 */
 	public $cookiePrefix = '';
 
@@ -253,6 +254,8 @@ class App extends BaseConfig
 	 * Set to `.your-domain.com` for site-wide cookies.
 	 *
 	 * @var string
+	 * 
+	 * @deprecated use Config\Cookie::$domain property instead.
 	 */
 	public $cookieDomain = '';
 
@@ -264,6 +267,8 @@ class App extends BaseConfig
 	 * Typically will be a forward slash.
 	 *
 	 * @var string
+	 * 
+	 * @deprecated use Config\Cookie::$path property instead.
 	 */
 	public $cookiePath = '/';
 
@@ -275,6 +280,8 @@ class App extends BaseConfig
 	 * Cookie will only be set if a secure HTTPS connection exists.
 	 *
 	 * @var boolean
+	 * 
+	 * @deprecated use Config\Cookie::$secure property instead.
 	 */
 	public $cookieSecure = false;
 
@@ -286,6 +293,8 @@ class App extends BaseConfig
 	 * Cookie will only be accessible via HTTP(S) (no JavaScript).
 	 *
 	 * @var boolean
+	 * 
+	 * @deprecated use Config\Cookie::$httponly property instead.
 	 */
 	public $cookieHTTPOnly = true;
 
@@ -310,39 +319,10 @@ class App extends BaseConfig
 	 * will be set on cookies. If set to `None`, `$cookieSecure` must also be set.
 	 *
 	 * @var string
+	 * 
+	 * @deprecated use Config\Cookie::$samesite property instead.
 	 */
 	public $cookieSameSite = 'Lax';
-
-	/**
-	 * --------------------------------------------------------------------------
-	 * Cookie Raw
-	 * --------------------------------------------------------------------------
-	 *
-	 * This flag allows setting a "raw" cookie, i.e., its name and value are
-	 * not URL encoded using `rawurlencode()`.
-	 *
-	 * If this is set to `true`, cookie names should be compliant of RFC 2616's
-	 * list of allowed characters.
-	 *
-	 * @var boolean
-	 *
-	 * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes
-	 * @see https://tools.ietf.org/html/rfc2616#section-2.2
-	 */
-	public $cookieRaw = false;
-
-	/**
-	 * --------------------------------------------------------------------------
-	 * Cookie Expires Timestamp
-	 * --------------------------------------------------------------------------
-	 *
-	 * Default expires timestamp for cookies. Setting this to `0` will mean the
-	 * cookie will not have the `Expires` attribute and will behave as a session
-	 * cookie.
-	 *
-	 * @var DateTimeInterface|integer|string
-	 */
-	public $cookieExpires = 0;
 
 	/**
 	 * --------------------------------------------------------------------------
--- a/app/Config/Cookie.php
+++ b/app/Config/Cookie.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace Config;
+
+use CodeIgniter\Config\BaseConfig;
+use DateTimeInterface;
+
+class Cookie extends BaseConfig
+{
+	/**
+	 * --------------------------------------------------------------------------
+	 * Cookie Prefix
+	 * --------------------------------------------------------------------------
+	 *
+	 * Set a cookie name prefix if you need to avoid collisions.
+	 *
+	 * @var string
+	 */
+	public $prefix = '';
+
+	/**
+	 * --------------------------------------------------------------------------
+	 * Cookie Expires Timestamp
+	 * --------------------------------------------------------------------------
+	 *
+	 * Default expires timestamp for cookies. Setting this to `0` will mean the
+	 * cookie will not have the `Expires` attribute and will behave as a session
+	 * cookie.
+	 *
+	 * @var DateTimeInterface|integer|string
+	 */
+	public $expires = 0;
+
+	/**
+	 * --------------------------------------------------------------------------
+	 * Cookie Path
+	 * --------------------------------------------------------------------------
+	 *
+	 * Typically will be a forward slash.
+	 *
+	 * @var string
+	 */
+	public $path = '/';
+
+	/**
+	 * --------------------------------------------------------------------------
+	 * Cookie Domain
+	 * --------------------------------------------------------------------------
+	 *
+	 * Set to `.your-domain.com` for site-wide cookies.
+	 *
+	 * @var string
+	 */
+	public $domain = '';
+
+	/**
+	 * --------------------------------------------------------------------------
+	 * Cookie Secure
+	 * --------------------------------------------------------------------------
+	 *
+	 * Cookie will only be set if a secure HTTPS connection exists.
+	 *
+	 * @var boolean
+	 */
+	public $secure = false;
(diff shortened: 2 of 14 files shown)


MGatner:
This is a good change. I would like to make a release this weekend, which would solidify all the Cookie changes since 4.1 - let's be sure to get this finished so it can be included.

mostafakhudair:
@MGatner I think next release should be done without any of new Cookie Class, because of cookie class is bloated with alot usless methods and need to be refactored, also to avoid BC

MGatner:
@mostafakhudair I thought there had been significant improvements to our Cookie handling since the last release, work that you and @paulbalandan both did. What do you think should be held back? Is it something that can easily be reverted and re-applied for the following release?

Paul, what are your thoughts?

paulbalandan:
Bloated? What aspects of the `Cookie` are bloated? Cookie is just implementing its contract interfaces' methods, which are actually getters and cloners of a web cookie's attributes. Plus, it has bonus methods to help parse a `Cookie-Header` string into a functional `Cookie` object, and some validation methods to make sure the passed attributes are valid against the multitude of cookie specifications. Are these useless to you? The bloat, I would say IMO, is the `ArrayAccess` implementation and some `?:` logic, but that is a compromise to retain backward compatibility to the time cookie handling was just using plain arrays.

If you can pinpoint the places that needs refactor, we'd consider excluding cookie in the release and perhaps I'd do the refactor myself. Otherwise, aside from the config reassignment, the cookie class is stable as-is.

system/Cookie/Cookie.php
```diff
@@ -629,7 +631,7 @@ public function withSameSite(string $sameSite)
 	 */
 	public function offsetExists($offset)
 	{
-		if (in_array($offset, ['expire', 'httponly', 'samesite'], true))
+		if (in_array($offset, ['expires', 'httponly', 'samesite'], true))
```
paulbalandan, line 634: This is a BC break.

system/Cookie/Cookie.php
```diff
@@ -653,19 +655,19 @@ public function offsetGet($offset)
 			throw new InvalidArgumentException(sprintf('Undefined offset "%s".', $offset));
 		}
 
-		if ($offset === 'expire')
+		if ($offset === 'expires')
```
paulbalandan, line 658: BC break.

paulbalandan (changes requested):

app/Config/App.php
```diff
@@ -310,40 +319,11 @@ class App extends BaseConfig
...
-	 *
-	 * If this is set to `true`, cookie names should be compliant of RFC 2616's
-	 * list of allowed characters.
-	 *
-	 * @var boolean
-	 *
-	 * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes
-	 * @see https://tools.ietf.org/html/rfc2616#section-2.2
-	 */
-	public $cookieRaw = false;
```
samsonasik, line 332: this is BC break, existing user no longer can't use `(new Config\App())->cookieRaw;`

mostafakhudair: (new Config\App())->cookieRaw; wasn't released yet,

samsonasik: oh, ok 👍

samsonasik:
@mostafakhudair please incorporate @paulbalandan review

app/Config/Cookie.php
```diff
@@ -0,0 +1,119 @@
...
+	 * - ''
+	 *
+	 * Alternatively, you can use the constant names:
+	 * - `Cookie::SAMESITE_NONE`
+	 * - `Cookie::SAMESITE_LAX`
+	 * - `Cookie::SAMESITE_STRICT`
+	 *
+	 * Defaults to `Lax` for compatibility with modern browsers. Setting `''`
+	 * (empty string) means default SameSite attribute set by browsers (`Lax`)
+	 * will be set on cookies. If set to `None`, `$cookieSecure` must also be set.
```
paulbalandan, line 96: $cookieSecure change to $secure

system/Cookie/Cookie.php
```diff
@@ -118,28 +119,29 @@ class Cookie implements ArrayAccess, CloneableCookieInterface
...
-				'raw'      => $config->cookieRaw ?? false,
-				'samesite' => $config->cookieSameSite ?? self::SAMESITE_LAX,
+				'prefix'   => $config->prefix,
+				'expires'  => $config->expires ?? 0,
+				'path'     => $config->path,
+				'domain'   => $config->domain,
+				'secure'   => $config->secure,
+				'httponly' => $config->httponly,
+				'samesite' => $config->samesite ?? self::SAMESITE_LAX,
+				'raw'      => $config->raw ?? false,
```
paulbalandan, line 144: No need for the fall back `??` values since you are using the new config.

system/Cookie/Cookie.php
```diff
@@ -660,12 +662,12 @@ public function offsetGet($offset)
...
 		{
-			return $this->httpOnly;
+			return $this->httponly;
 		}
 
 		if ($offset === 'samesite')
 		{
-			return $this->sameSite;
+			return $this->samesite;
 		}
```
paulbalandan, line 671: Since you made all property names as lowercase, then there's no  need for these 2 guards here. Also in offsetExists method, no need to check for the two. Only check for `expire`.

system/Session/Session.php
```diff
@@ -178,22 +179,27 @@ public function __construct(SessionHandlerInterface $driver, App $config)
...
-			'domain'   => $config->cookieDomain ?? '',
-			'path'     => $config->cookiePath ?? '/',
-			'secure'   => $config->cookieSecure ?? false,
+			'path'     => $this->cookiePath,
+			'domain'   => $this->cookieDomain,
+			'secure'   => $this->cookieSecure,
 			'httponly' => true, // for security
-			'samesite' => $config->cookieSameSite ?? Cookie::SAMESITE_LAX,
+			'samesite' => $this->cookieSameSite,
+			'raw'      => $cookie->raw ?? false,
```
paulbalandan, line 202: Do not add dependence to the deprecated cookie properties. It is deprecated so that it will no longer be used in active code and can be safely removed in the future but still preserving for those users who depended on such. By deprecating but still actively depending on those properties would defeat the purpose of the deprecation.

system/Test/Mock/MockCookieConfig.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Test\Mock;
+
+use Config\Cookie;
+
+class MockCookieConfig extends Cookie
```
paulbalandan, line 16: Do not add useless mocks. Mocks can be easily made natively with PHPUnit's mocking system.

tests/system/Cookie/CookieTest.php
```diff
@@ -173,24 +174,40 @@ public function testSetCookieHeaderCreation(string $header, array $changed): voi
...
-		];
+				  'test=value; Secure; HttpOnly; SameSite=Lax',
+				  [
+					  'name'     => 'test',
+					  'value'    => 'value',
+					  'secure'   => true,
+					  'httponly' => true,
+					  'samesite' => 'Lax',
+				  ],
+			  ];
```
paulbalandan, line 210: Revert the changes here. our codesniffer implementation cannot still handle gracefully these nested arrays but can become unsightly.

tests/system/Security/SecurityTest.php
```diff
@@ -209,12 +209,12 @@ public function testSendingCookiesFalse(): void
 		$request = $this->createMock(IncomingRequest::class);
 		$request->method('isSecure')->willReturn(false);
 
-		$config = new MockAppConfig();
+		$config = config('Cookie');
```
paulbalandan, line 212: I'm not sure if we have proper cleanup of shared instances in the tests, but please refrain from using shared instances. Instead, use the `injectMock` facility of `Factories` with a subsequent `reset` to clean up.

The thing is using `config(...)` calls in the tests and altering the state will persist and affect the other tests depending on another `config(..)` calls. As much as possible, we want to make sure each and every test receives a fresh, unadulterated instance of a config it needs.

paulbalandan (changes requested):

app/Config/Cookie.php
mostafakhudair: shoud we force set secure to true when developer uses None a samesite attribute?

tests/system/Cookie/CookieTest.php
mostafakhudair: Actually I didn't change this, maybe my IDE changed it, reverted

app/Config/Cookie.php
paulbalandan: That is in the cookie spec. None samesite should have the Secure attribute. Otherwise, the cookie is rejected.

mostafakhudair: what is this happend?

if (strtolower($samesite) === self::SAMESITE_NONE && ! $secure)
{
	//throw CookieException::forInvalidSameSiteNone();
	self::setDefaults(['secure' => true];
}

paulbalandan: That would be condoning the developer's mistake. What sets apart CI4 from CI3 is that we throw exceptions for coding mistakes rather than the framework correcting them and have the dev with no idea that they have a mistake.

system/Session/Session.php
```diff
@@ -181,19 +182,24 @@ public function __construct(SessionHandlerInterface $driver, App $config)
...
-			'domain'   => $config->cookieDomain ?? '',
-			'path'     => $config->cookiePath ?? '/',
-			'secure'   => $config->cookieSecure ?? false,
+			'path'     => $config->path,
+			'domain'   => $config->domain,
+			'secure'   => $config->secure,
 			'httponly' => true, // for security
-			'samesite' => $config->cookieSameSite ?? Cookie::SAMESITE_LAX,
+			'samesite' => $config->samesite,
+			'raw'      => $config->raw ?? false,
```
paulbalandan, line 202:
```suggestion
			'raw'      => $config->raw,
```
```suggestion
			'raw'      => $config->raw ?? false,
```

tests/system/Cookie/CookieTest.php
mostafakhudair: Reverted more than once, but when I commit and pull, it auto reverse my changes

paulbalandan: Are you using VS Code?

paulbalandan: You can use the Git Staged Amend No Verify option. Then Push Force

paulbalandan: if not. directly in terminal
```
git commit --amend --no-verify
git push --force-with-lease
```

mostafakhudair: Thank you

mostafakhudair:
What is wrong with tests?

samsonasik:
That seems race condition issue in SQLSRV action. It seems unrelated.

MGatner:
Have tests been rerun since #4539?

mostafakhudair:
Yes, twice

paulbalandan (approved):

paulbalandan:
Thank you, @mostafakhudair

MGatner:
🥳

PR #4512, opened 2021-04-01
Cease catching non-HTTP errors

MGatner:
**Description**
`ControllerTester` catches any exceptions and wraps them into a `TestResponse`. If the error is a valid HTTP status code then it will set it on the response, otherwise it ignores them. This ignoring means that any regular exception (i.e. developer mistake) will return a `TestResponse` with `200` status and a `null` body - very misleading.

This PR re-throws exceptions that are non-HTTP statuses to prevent masking underlying issues.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/ControllerTester.php
+++ b/system/Test/ControllerTester.php
@@ -179,10 +179,13 @@ trait ControllerTester
 		{
 			$code = $e->getCode();
 
-			if ($code >= 100 && $code < 600)
+			// If code is not a valid HTTP status then assume there is an error
+			if ($code < 100 || $code >= 600)
 			{
-				$result->response()->setStatusCode($code);
+				throw $e;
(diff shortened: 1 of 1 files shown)


paulbalandan (changes requested):
I'm in for this change to promote better debugging. We need an entry to the changelog user guide to announce this change which can be a BC to users' tests.

MGatner:
This should actually be considered an addendum to https://github.com/codeigniter4/CodeIgniter4/pull/4379, which hasn't been released yet so I think it should not break anything. Tests that error out with "invalid status code" on 4.1 would now throw the original exception instead.

samsonasik (approved):

PR #4513, opened 2021-04-03
Bug: can't change and override valid locales #4297

davislasis:
Bug: can't change and override valid locales [#4297](https://github.com/codeigniter4/CodeIgniter4/issues/4297)

**Description**
You can override setLocale() with 2nd parameter if you need to do that. For example before controller in filter files as detailed described in issue link provided above.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -214,14 +214,20 @@ class IncomingRequest extends Request
 	 * Sets the locale string for this request.
 	 *
 	 * @param string $locale
+	 * @param array $locales
 	 *
 	 * @return IncomingRequest
 	 */
-	public function setLocale(string $locale)
+	public function setLocale(string $locale, array $locales = [])
 	{
+		if (! empty($locales))
+		{
+			$this->validLocales = array_merge($this->validLocales, $locales);
+		}
+
 		// If it's not a valid locale, set it
 		// to the default locale for the site.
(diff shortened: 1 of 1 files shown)


system/HTTP/IncomingRequest.php
```diff
@@ -214,11 +214,17 @@ public function getLocale(): string
...
 	 *
 	 * @param string $locale
+	 * @param array $locales
 	 *
 	 * @return IncomingRequest
 	 */
-	public function setLocale(string $locale)
+	public function setLocale(string $locale, array $locales = [])
 	{
+		if (! empty($locales))
```
paulbalandan, line 223:
```suggestion
		if ($locales !== [])
```

WinterSilence:
@davislasis
1. can you post useful example
2. why not separate method?
3. you method only merge - can't delete locales.

WinterSilence:
@davislasis you need replace https://github.com/codeigniter4/CodeIgniter4/blob/7d6116344a22657c365cfcb374a77b9801a1f9ee/system/HTTP/IncomingRequest.php#L161 to `$this->validLocales =& $config->supportedLocales;` and change supported locales by config

lonnieezell:
Judging from discussion on the forums currently, locales is something we will be looking at in more detail in the coming months. Since this is a breaking change, closing. If we can get a PR that fixes this without a BC break we'd definitely consider it.

PR #4523, opened 2021-04-06
Shortcircuit toolbar on CLI

MGatner:
**Description**
`Toolbar` is already disabled under CLI, but the listener events still trigger. This PR excludes those Events from triggering as well, which could prevent some performance hits (especially if developers add their own collectors).

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/Events.php
+++ b/app/Config/Events.php
@@ -46,7 +46,7 @@ Events::on('pre_system', function () {
 	 * --------------------------------------------------------------------
 	 * If you delete, they will no longer be collected.
 	 */
-	if (CI_DEBUG)
+	if (CI_DEBUG && ! is_cli())
 	{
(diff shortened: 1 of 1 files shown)


app/Config/Events.php
```diff
@@ -46,7 +46,7 @@
 	 * --------------------------------------------------------------------
 	 * If you delete, they will no longer be collected.
 	 */
-	if (CI_DEBUG)
+	if (CI_DEBUG && ! is_cli())
```
paulbalandan, line 49: if I will use `php spark serve` then is `is_cli()` result to `true`? If so, does this mean we'll specifically disallow toolbar on development server?

MGatner: While I don't think so, more importantly this is the exact boolean construct already used in `Toolbar` itself, so we aren't changing any scenarios just disabling it sooner.

paulbalandan (approved):

PR #4532, opened 2021-04-08
Table alias is not working with softDeletes 

jahmedk:
Created property (tableAlias) for handling alias in model


paulbalandan (changes requested):
Please clean first your copy of your fork. There's a lot of unnecessary changes here.

PR #4534, opened 2021-04-08
[Query] Do not recompile binds if already compiled

### LINKED ISSUE #4518: Bug: Query::matchSimpleBinds index problem only toolbar.
**Describe the bug**

**CodeIgniter 4 version**
4.1.1

Screen Shot 2021-04-05 at 13 42 00
Screen Shot 2021-04-05 at 13 42 20

**Context**
 - OS: [Centos 7]
 - Web server [nginx]
 - PHP version [7.3.27]

paulbalandan:
**Description**
Fixes #4518 

The thing is the Database Collector call `Query::debugToolbarDisplay` to return a nicely formatted SQL string. However, that method will erroneously call `$this->compileBinds()` everytime. The problem with this approach is that the DB operation had run before the toolbar display collection so all queries have been compiled. Calling it again will give false positives for bind markers when special syntaxes uses the bind marker (such as Postgres' JSON operations).

What we really need is the compiled final query string. Since the compiling was done before collection, we can be sure that it is already set. However, for the sake of prudence, we'll put up a guard if the requested string is not yet initialized and then we'll call `compileBinds()` right away.

ping @byazrail kindly check this if it will work on your issue

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -472,7 +472,12 @@ class Query implements QueryInterface
 			')',
 		];
 
-		$sql = $this->getQuery();
+		if (empty($this->finalQueryString))
+		{
+			$this->compileBinds(); // @codeCoverageIgnore
+		}
+
+		$sql = $this->finalQueryString;
 
 		foreach ($highlight as $term)
 		{


byazrail:
@paulbalandan,

Thanks problem solved 👍

vlakoff:
Refs #5127.

vlakoff:
You mentioned Postgres' JSON operations. Could you provide an example of SQL string that gets modified again if the compiles are run on it a second time? This would help writing tests for #5127.

kenjis:
@vlakoff This? https://github.com/codeigniter4/CodeIgniter4/issues/4518#issuecomment-813532971
```php
$builder = $db->table('search p');
$sql = $builder
	->select('p.*')
	->where('p.location ?| ARRAY[\'639\']', null, false)
	->where('p.status', 1)
	->orderBy('p.id', 'ASC')
	->limit(10)
	->get();
```

https://www.postgresql.org/docs/9.5/functions-json.html

Issue #4527, opened 2021-04-07
Table alias is not working with softDeletes


jahmedk:
![image](https://user-images.githubusercontent.com/22505794/113851306-cfbb5400-97b4-11eb-9542-e854ee044ffa.png)
![image](https://user-images.githubusercontent.com/22505794/113851343-d944bc00-97b4-11eb-9415-0b3b6fdf9fab.png)
![image](https://user-images.githubusercontent.com/22505794/113851575-127d2c00-97b5-11eb-89e6-d0d0c68861fa.png)

paulbalandan:
I don't recall `Model` directly supporting table aliasing in its `$table` property. It is erroring because `$model->table` is also used in forming the complete field name, such as table.id. In your case, since you aliased it directly in the model, your id would be queried as `cities AS ct.id` which is obviously wrong since you don't have a `cities` field.

You should do the aliasing the table like this:
```php
class CitiesModel extends Model
{
    protected $table = 'cities';
    // ...
}

$builder = model('App\Models\CitiesModel')->builder('cities AS ct');
$builder->select(['ct.name']);
// ... 
```

jahmedk:
Thanks @paulbalandan

jahmedk:
When i set the alias according to your solution, findAll is undefined
![image](https://user-images.githubusercontent.com/22505794/113973541-ecf42f00-9855-11eb-9f9c-4a6020cf68eb.png)

jahmedk:
When i set the alias according to your solution, findAll is undefined
![image](https://user-images.githubusercontent.com/22505794/113973857-6be96780-9856-11eb-97a2-b8fc0a23f2ea.png)

paulbalandan:
`findAll` is a method of Model, while `$builder` is an instance of `BaseBuilder`. Do this:
```php
$model = model('App\Models\CitiesModel');

// initialize the builder's table as aliased ct
$model->builder('cities AS ct'); // notice we don't save the builder instance as compared before.

// call the model using the aliased table using builder mixin
$model->select(['ct.name']);
$model->join(...);
$model->findAll(100, 0);
```

iRedds:
@paulbalandan In your example, if the property of the table is "cities AS ct", it will raise the error described in the issue. If the value of the table property is different, then the call to the builder ``$model->builder('cities AS ct'); `` will not change anything, since the method will return a new instance of the builder, not a shared.

jahmedk:
Now in select alias isn't working
[...]

PR #4536, opened 2021-04-08
Fixed table alias is not working with softDeletes

jahmedk:
Created property (tableAlias) for handling alias in model

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -53,6 +53,13 @@ class Model extends BaseModel
 	protected $table;
 
 	/**
+	 * The table's alias
+	 *
+	 * @var string
+	 */
+	protected $tableAlias;
+
+	/**
 	 * The table's primary key.
 	 *
 	 * @var string
@@ -106,6 +113,19 @@ class Model extends BaseModel
 		$this->db = &$db;
 
 		parent::__construct($validation);
+		if (strpos($this->table, ' ') !== false)
+		{
+			// if the alias is written with the AS keyword, remove it
+			$this->tableAlias = preg_replace('/\s+AS\s+/i', ' ', $this->table);
+
+			// Grab the alias
+			$this->tableAlias = trim(strrchr($this->tableAlias, ' '));
+
+			// Store the alias, if it doesn't already exist
+			$this->db->addTableAlias($this->tableAlias);
+		} else {
+			$this->tableAlias = $this->table;
+		}
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


WinterSilence:
table name can contain whitespaces: 
~~~sql
CREATE TABLE `to do` (`id` INT(11) NULL DEFAULT NULL) COLLATE='utf8_general_ci' ENGINE=InnoDB;
~~~

WinterSilence:
@jahmedk I told you in issue - you must check by `Connection::tableExists($table)`. It's can be work as in other cases: `to as do` or `to do`.  do you try add only `$this->db->addTableAlias($table);`? why not fix it in `setTable($table)`?

jahmedk:
Yes @WinterSilence you're right we should handle all possible cases

jahmedk:
@paulbalandan according to @WinterSilence test case we must check Connection::tableExists($table) before creating alias in BaseBuilder and BaseModel

WinterSilence:
@jahmedk what's about 2 last questions?

jahmedk:
> table name can contain whitespaces:
> 
> ```sql
> CREATE TABLE `to do` (`id` INT(11) NULL DEFAULT NULL) COLLATE='utf8_general_ci' ENGINE=InnoDB;
> ```
?

jahmedk:
@WinterSilence ??

WinterSilence:
@jahmedk 
> do you try add only $this->db->addTableAlias($table);? why not fix it in setTable($table)?

jahmedk:
> @jahmedk
> 
> > do you try add only $this->db->addTableAlias($table);? why not fix it in setTable($table)?

Because setTable is just setting the table property of CodeIgniter\Model and in the CodeIgniter\Database\BaseBuilder they are checking alias with whitespace i am thinking alias should be handle in CodeIgniter\Database\BaseBuilder.

jahmedk:
> > @jahmedk
> > > do you try add only $this->db->addTableAlias($table);? why not fix it in setTable($table)?
> 
> Because setTable is just setting the table property of CodeIgniter\Model and in the CodeIgniter\Database\BaseBuilder they are checking alias with whitespace i am thinking alias should be handle in CodeIgniter\Database\BaseBuilder.

Your suggestion is good https://github.com/codeigniter4/CodeIgniter4/issues/4527#issuecomment-816260255

WinterSilence:
@jahmedk don't understand you:

>  i am thinking alias should be handle in CodeIgniter\Database\BaseBuilder.

but your PR changed `Model`

jahmedk:
> @jahmedk don't understand you:
> 
> > i am thinking alias should be handle in CodeIgniter\Database\BaseBuilder.
> 
> but your PR changed `Model`

Yes i did
But it is not complete

PR #4538, opened 2021-04-09
[Rector] Fixing rector action hang

samsonasik:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/test-rector.yml
+++ b/.github/workflows/test-rector.yml
@@ -68,4 +68,4 @@ jobs:
         run: composer update --ansi --no-interaction
 
       - name: Run static analysis
-        run: vendor/bin/rector process --dry-run --no-progress-bar
+        run: vendor/bin/rector process --dry-run --debug
--- a/.github/workflows/test-rector.yml
+++ b/.github/workflows/test-rector.yml
@@ -68,8 +68,4 @@ jobs:
         run: composer update --ansi --no-interaction
 
       - name: Run static analysis
-        run: |
-          vendor/bin/rector process app --dry-run --debug
-          vendor/bin/rector process system --dry-run --debug
-          vendor/bin/rector process tests --dry-run --debug
-          vendor/bin/rector process utils/Rector --dry-run --debug
+        run: vendor/bin/rector process --dry-run --no-progress-bar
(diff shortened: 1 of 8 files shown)


samsonasik:
Github actions seems very slow

samsonasik:
It seems `__DIR__ . '/app/Config/App.php',` re-autoload somewhere, so it hang, I register it to skipped option.

.github/workflows/test-rector.yml
```diff
@@ -68,4 +68,4 @@ jobs:
         run: composer update --ansi --no-interaction
 
       - name: Run static analysis
-        run: vendor/bin/rector process --dry-run --no-progress-bar
```
MGatner, line 71: I added the "no-progress-bar" because it doesn't display correctly during actions and ends up adding a lot of debris to the output. If "debug" is mutually exclusive with that option that is fine but I would prefer not to have it.

MGatner (changes requested):

.github/workflows/test-rector.yml
samsonasik: I see, I just want to debug in github action to see what last output that cause hang. I will revert back when everything green again.

samsonasik:
It is very strange, it is possibly due dependencies of phpstan - phpstan/phpdoc-parser - nikic/phpparser - rector on latest update. I tried some pin versions on previous version but no luck yet.

.github/workflows/test-rector.yml
samsonasik: using latest rector/rector:dev-main seems fixed it, reverted the command.

samsonasik:
`@param` space should be ignored and not reported in `rector`:

```diff
-	 * @throws             Exception
+	 * @throws Exception
```

samsonasik:
temporary disable auto import as cause spaces on `@param`/`@throws` trimmed which always replaced by CodeIgniter 4 coding standard.

composer.json
```diff
@@ -22,7 +22,7 @@
 		"phpstan/phpstan": "0.12.83",
 		"phpunit/phpunit": "^9.1",
 		"predis/predis": "^1.1",
-		"rector/rector": "^0.10",
+		"rector/rector": "dev-main",
```
samsonasik, line 25: should use pinned version on next release.

PR #4541, opened 2021-04-09
Entity Cast Organize

mostafakhudair:


iRedds:
I don't think it is correct to rename AbstractCast to BaseCast.
You should also make changes to the UserGuide.

iRedds:
Since you have created a separate Entity namespace, it would be logical to move the exception and the Entity class itself there.
But the latter will lead to BC.

mostafakhudair:
I just realized that all abstracted classes named Base...., that's why I renamed it to BaseCast

mostafakhudair:
I'm planing to move Entity class to the new namespace, but yet.

MGatner:
I think this is a better structure. I'm on mobile so can't confirm but... these haven't been released yet, right?

mostafakhudair:
Yes, That's why I did it now to avoid BC later

mostafakhudair:
@MGatner  we need to make sure is will be merged before upcoming release

MGatner:
*sigh* I mean I'm glad we're getting these taken care while we can, but now this is delaying release. Let's do what we can to get this wrapped up ASAP

system/Exceptions/CastException.php
```diff
@@ -43,20 +43,4 @@ public static function forInvalidJsonFormatException(int $error)
 				return new static(lang('Cast.jsonErrorUnknown'));
 		}
 	}
```
paulbalandan, line 45: Shouldn't this exception be gone in favor of the other?

paulbalandan (commented):
I'd like to get @iRedds' review comments on this as he's the original author of the addition.

iRedds:
@paulbalandan, In principle, I have already expressed my opinion above. 
This PR does not change the functionality that I proposed.

paulbalandan:
> @paulbalandan, In principle, I have already expressed my opinion above.
> This PR does not change the functionality that I proposed.

I am just wondering if you have more to add.

MGatner:
I think we should go ahead and move Entity. You can leave `CodeIgniter\Entity` as a deprecated class extension of `CodeIgniter\Entity\Entity`. I'm open to other ideas though.

paulbalandan:
I would like to see `CodeIgniter\Entity` as an empty file with only `class_exists(...)` and `class_alias(...)` to the new namespace. I just thinking if this would be autoloaded by our `Autoloader` class.

system/Exceptions/CastException.php
mostafakhudair: This would be BC, this exception were added before version 4.1

paulbalandan: If I'm not mistaken, system exceptions are strictly for internal use and is not covered by BC. Wait, I'll check.

paulbalandan: OK. That was for language strings. Do add `@deprecated` tag for the class instead.

mostafakhudair:
What if we cleaned current Entity class from all methods and extends it from CodeIgniter\Entity\Entity, and also mark at as deprecated class?

MGatner:
Can we get this squashed? Also happy to squash it on merge, but we don't need 30 commits of the same thing :O

mostafakhudair:
sorry for that, but i cannot push from my local to this PR any more, but once test is fixed, i'll create new clean PR with one commit :D

PR #4543, opened 2021-04-10
[Rector] Apply RemoveDefaultArgumentValueRector

samsonasik:
Remove argument value, if it is the same as default value with `RemoveDefaultArgumentValueRector` rector rule.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/tests/system/View/ParserTest.php
+++ b/tests/system/View/ParserTest.php
@@ -419,7 +419,7 @@ class ParserTest extends CIUnitTestCase
 		{
 			$expected = $value;
 		}
-		$this->assertEquals($expected, \esc($value, 'html'));
+		$this->assertEquals($expected, \esc($value));
 	}
 
 	//------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


tests/system/View/ParserTest.php
```diff
@@ -764,7 +764,7 @@ public function testParserPluginClosure()
 			return 'Hello, ' . trim($params[0]);
 		};
 
-		$this->parser = new Parser($config, $this->viewsDir, $this->loader);
+		$this->parser             = new Parser($config, $this->viewsDir, $this->loader);
```
paulbalandan, line 767: this shouldn't align

MGatner (approved):
After Paul's style note, this looks good. 

Aside, I feel like I'm fighting our styling more than finding it helpful these days. So many messed up indentations and type names.

tests/system/View/ParserTest.php
samsonasik: it is changed by coding standard, I need to manually disable cs : 

```bash
Running Code Sniffer...
F 1 / 1 (100%)



PHPCBF RESULT SUMMARY
----------------------------------------------------------------------
FILE                                                  FIXED  REMAINING
----------------------------------------------------------------------
...www/CodeIgniter4/tests/system/View/ParserTest.php  1      96
----------------------------------------------------------------------
A TOTAL OF 1 ERROR WERE FIXED IN 1 FILE
```

I've updated it.

PR #4545, opened 2021-04-10
Fix dot notation for if_exist

### LINKED ISSUE #4521: Bug: if_exist not working with ".*" notation
**Describe the bug**
When building validation rules for an optional array, the if_exist doesn't work correctly.

**CodeIgniter 4 version**
4.1.1

**Affected module(s)**
Validation

Here is the code that I am using to validate my status array that may or may not come into the endpoint in my controller:
```
if (!$this->validate(
    [
        'status.*' => 'if_exist|in_list[status1,status2]'
    ]
)) {
    return $this->fail($this->validator->getErrors());
}
```
When I pass something like:
```
{ "status": ["bad-status"]}
```
It doesn't fail the validation because it thinks the array doesn't exist and skips the "in_list" rule.

**Context**
 - OS: CentOS 8
 - Web server: Apache 2.4.37 
 - PHP version: 8.0.2

paulbalandan:
**Description**
Fixes #4521 

This is a patch fix only for the `if_exist` rule and DOES NOT yet cover other rules.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -215,11 +215,30 @@ class Validation implements ValidationInterface
 
 		if (in_array('if_exist', $rules, true))
 		{
-			// If the if_exist rule is defined
-			// and the current field does not exist in the input data
-			// we can return true, ignoring all other rules to this field.
-			if (! array_key_exists($field, array_flatten_with_dots($data)))
+			$flattenedData = array_flatten_with_dots($data);
+			$ifExistField  = $field;
+
+			if (strpos($field, '.*') !== false)
 			{
+				// We'll change the dot notation into a PCRE pattern
+				// that can be used later
+				$ifExistField = str_replace('\.\*', '\.(?:[^\.]+)', preg_quote($field, '/'));
+
+				$dataIsExisting = array_reduce(array_keys($flattenedData), static function ($carry, $item) use ($ifExistField) {
+					$pattern = sprintf('/%s/u', $ifExistField);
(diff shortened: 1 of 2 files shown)


system/Validation/Validation.php
```diff
@@ -812,24 +831,21 @@ protected function splitRules(string $rules): array
...
 		);
 
-		$_rules = preg_split(
-				$pipeNotInBracket,
-				$rules
-		);
+		$rules = preg_split($pipeNotInBracket, $rules);
 
-		return array_unique($_rules);
+		return array_unique($rules);
```
samsonasik, line 842: iirc, it was @michalsn 's request to keep `$_` naming, see https://github.com/codeigniter4/CodeIgniter4/blob/63d0b27420e7506f3ac84147a3d7bc6c0adc6e39/utils/Rector/UnderscoreToCamelCaseVariableNameRector.php#L107

MGatner:
Other than Abdul's comment this looks good!

Out of curiosity... does `@internal` really do anything on a `final` class?

paulbalandan:
> Other than Abdul's comment this looks good!
> 
> Out of curiosity... does `@internal` really do anything on a `final` class?

Actually, I copied this rule from php-cs-fixer. `@internal` means the class should not be used in users' code. This can be applied to anything where BC promise is not guaranteed. `final` classes may or may not be for `@internal` use. `final` classes only limit their inheritance.

PR #4551, opened 2021-04-10
Cookie final touch

mostafakhudair:


paulbalandan (commented):
I'm sorry but I do not agree to consolidating the expires methods. The current implementation wants to achieve  a "fluent" interface in interacting with the Cookie and the proposed changes makes this "less fluent". On "strict" static analysis, (i.e. using phpstan-strict-rules), this change will fail because it cannot determine whether the result is string or int based on implementation alone. Another is on the 3 withExpire** methods. While the two is much more of an addon, it is much more user friendly to use "withExpired" to quickly get an expired cookie. We're doing our users a good UX with this.

samsonasik:
Your commit seems mixed with develop branch which make harder to review. Please rebase.

paulbalandan:
I'm sorry but still I cannot accept this one. Thanks for your time, though.

mostafakhudair:
isn't Cookie::withExpiresAt() gives the same result of Cookie::withExpired()?

PR #4560, opened 2021-04-12
[Rector] Apply Rector: ChangeArrayPushToArrayAssignRector

samsonasik:
Change array_push() to direct variable assign with `ChangeArrayPushToArrayAssignRector` rector rule.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Files/File.php
+++ b/system/Files/File.php
@@ -200,7 +200,7 @@ class File extends SplFileInfo
 				{
 					$i = end($parts);
 					array_pop($parts);
-					array_push($parts, ++ $i);
+					$parts[]     = ++ $i;
 					$destination = $info['dirname'] . '/' . implode($delimiter, $parts) . $extension;
 				}
 				else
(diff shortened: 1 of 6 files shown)


system/Files/File.php
```diff
@@ -200,7 +200,7 @@ public function getDestination(string $destination, string $delimiter = '_', int
 				{
 					$i = end($parts);
 					array_pop($parts);
-					array_push($parts, ++ $i);
+					$parts[]     = ++ $i;
```
paulbalandan, line 203: Do we have a rector rule that makes the increment/decrement operators close to their subjects?
```diff
- ++ $i;
- $j ++
+ ++$i
+ $j++
```

paulbalandan (approved):

system/Files/File.php
samsonasik: It seems not yet for now.

Issue #4558, opened 2021-04-12
Bug: Mysql connection issue with MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT 


fedeburo:
https://github.com/codeigniter4/CodeIgniter4/blob/1d40479705a531a3f98371bd12588a20f601268b/system/Database/MySQLi/Connection.php#L162-L165

Hi, i am facing a problem making a mysqli ssl connection to database. 
I am using CodeIgniter version 4, 
mysqli->client_info = "mysqlnd 7.4.3"
Database driver: "MySQLi"

The problem is that when i have the ssl_verify = false, it never gets into:
`elseif (defined('MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT') && version_compare($this->mysqli->client_info, '5.6', '>='))
					{
						$client_flags += MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
					}`.

The version_compare is comparing the mysqli->client_info which in my case is "mysqlnd 7.4.3" to a "5.6" and this will always return false.

Something similar is happening on line 166.

A possible fix it could be replace that part of the version_compare to:
`elseif (defined('MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT') && version_compare(preg_replace("/[a-zA-Z\s]/", "", $this->mysqli->client_info), '5.6', '>='))` 
Where there i am replacing any string and whitespace before the version number.
Would you be able to change this? If not, which could be the solution to this problem?

paulbalandan:
Can you open a PR for this?

fedeburo:
> Can you open a PR for this?

Ok. I am not sure if i should create a new branch or make the pull request in an existing one?

paulbalandan:
Please create a new branch for your PR.

fedeburo:
#4566 pull request done

PR #4566, opened 2021-04-13
Mysql connection issue with MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT

fedeburo:
Changes on file MySQLi/Connection.php
Comparing versions numbers without the string when connecting through ssl with a ssl_verify = false

**Description**
Fixes Issue #4558 . Added a regex in the if condition when comparing mysqli_client_info with version_compare(). Now, it is comparing the version number without the previous string and whitespace.

**Checklist:**
- [ -] Securely signed commits
- [ - ] Component(s) with PHPdocs
- [  ] Unit testing, with >80% coverage
- [-  ] User guide updated
- [- ] Conforms to style guide


MGatner:
How about `preg_filter` with digits and `.` instead? Seems like that's what we are actually after.

cmacccaba:
Hi @MGatner, 
I've been working with @fedeburo on this, and running some tests against PHP version_compare() method, and also looking into the preg_filter() approach, but believe there is simpler solution, and one that is more developer friendly for future. 

We've been using version_compare() to compare apples to engines, and it will always fail. Lets compare apples with apples :) 

**A proposed solution:** prefix 'mysqlnd ' to the 2nd string param in the version_compare(), E.g. In the elseif() and if() conditions, Change the parts: 

```php
version_compare($this->mysqli->client_info, '5.6', '>=') 
// TO: version_compare($this->mysqli->client_info, 'mysqlnd 5.6', '>=')

version_compare($this->mysqli->client_info, '5.7.3', '<=') 
// TO: version_compare($this->mysqli->client_info, 'mysqlnd 5.7.3', '<=')
```

- No need for preg_filter() and regex interfering with the inner workings of the php method 'version_compare()', and potentially causing future bugs/issues that would be later difficult to find and debug. 
- The tests I've run confirm version compare will work as expected when fed comparable strings.(See code below). 

**Why?** 
By adding preg_filter() in the inputs to the version_compare() method makes things very messy indeed, and potentially interfere with the inner workings of version_compare() eg. the development state parts (eg '4.1.2-dev' etc.). Here is an extract from the PHP Manual: "*...not only versions with different levels like '4.1' and '4.1.2' can be compared but also any PHP specific version containing development state.*". 

- If we truly believe there is an issue with version_compare() then a bug should be reported with https://bugs.php.net/report.php , but tests I've run show that we have the wrong string for the 2nd param, and is not an issue with version_compare(). 
- If we believe the native php methods mysqli_get_client_info() has an issue and shouldn't have the String Prefix 'mysqlnd ' before the version number, again a bug should be reported to php.net. 
- Those 2 methods are probably not things we should interfere with inside CIv4 logic. 
- We probably shouldn't be modifying/filtering the end users real mysqli client_info string, but rather set our hard-coded comparable correctly. 

By only modifying the 2nd hard-coded string (currently '5.6') to 'mysqlnd 5.6' it will be very easily identified and changed later IF PHP.net decide that mysqli_get_client_info() shouldn't have the string prefix and update the method in a future version. 

I've checked the output of mysqli_get_client_info() AKA mysqli->client_info in CIv4, that provides the version info with a few different machines and apache servers and all have the prefix 'mysqlnd ' before the version numbers. (Have checked with devices running php v5 and v7, I have not checked with anything running php v8). 

Also note, I read through the PHP bug report referenced in the developer comment above the elseif() Line 160, and the term 'mysqlnd' appears 55 times, so I don't expect there will be any changes coming to the output of mysqli_get_client_info() anytime soon.

If in doubt about how version_compare() works, check the Manual doc's, its designed to compare two "PHP-standardized" version number strings. If they are not standardized, then they are not comparable. Link: https://www.php.net/manual/en/function.version-compare

**CODE AND TESTS:** 
```php
$versionList = array(
		'mysqlnd 5.0.12-dev'=> "false.", 
		'mysqlnd 5.6' 		=> "true.",
		'mysqlnd 5.6.0-dev' => "true.", //Should be seen as TRUE, here to show things are working as they should be.
		'mysqlnd 5.7.0-dev' => "true.", //Should be seen as TRUE, here to show things are working as they should be. 
		'mysqlnd 7.4.3' 	=> "true."
	); 
### TEST CURRENT LOGIC IN CIv4 - Compare to '5.6':
#------------------------------------------------------------
	foreach ($versionList as $sqli_client_info => $expectedResult) 
	{
		echo "<br>For case '".$sqli_client_info."': <b> \tExpected: ".$expectedResult."</b>  \tActual: "; var_dump( version_compare($sqli_client_info, '5.6', '>=') );
	}
# OUTPUTS: 
//For case 'mysqlnd 5.0.12-dev': Expected: false. 	Actual: bool(false)
//For case 'mysqlnd 5.6': 	Expected: true. 	Actual: bool(false)
//For case 'mysqlnd 5.6.0-dev': Expected: true. 	Actual: bool(false)
//For case 'mysqlnd 5.7.0-dev': Expected: true. 	Actual: bool(false)
//For case 'mysqlnd 7.4.3': 	Expected: true. 	Actual: bool(false)
### SUMMARY: 
# Seems like when string and/or whitespace before the number, we always get false. BECAUSE the '5.6' in the if conditions are mising a matching prefix 'mysqlnd '.
# Note, that the '-dev' postfix is not effected, and things work as expected and as per PHP Manual docs. 


### PROPOSED NEW LOGIC IN CIv4  - Compare to 'mysqlnd 5.6':
#------------------------------------------------------------
foreach ($versionList as $sqli_client_info => $expectedResult) 
	{ 
    	echo "<br>For case '".$sqli_client_info."': <b> \tExpected: ".$expectedResult."</b>  \tActual: "; var_dump( version_compare($sqli_client_info, 'mysqlnd 5.6', '>=') );
	}
# OUTPUTS: 
//For case 'mysqlnd 5.0.12-dev': Expected: false. 	Actual: bool(false)
//For case 'mysqlnd 5.6': 	Expected: true. 	Actual: bool(true)
//For case 'mysqlnd 5.6.0-dev': Expected: true. 	Actual: bool(true)
//For case 'mysqlnd 5.7.0-dev': Expected: true. 	Actual: bool(true)
//For case 'mysqlnd 7.4.3': 	Expected: true. 	Actual: bool(true)

```

MGatner:
I'm very happy to review an alternate solution. This is not my area of expertise so I leave the approach decisions up to you all.

fedeburo:
New pull request sent with the new fix 
`version_compare($this->mysqli->client_info, '5.6', '>=') 
// TO: version_compare($this->mysqli->client_info, 'mysqlnd 5.6', '>=')

version_compare($this->mysqli->client_info, '5.7.3', '<=') 
// TO: version_compare($this->mysqli->client_info, 'mysqlnd 5.7.3', '<=')`

cmacccaba:
Found this discussion around the same issues (back in 2016), adding it here for reference: 
https://forum.codeigniter.com/printthread.php?tid=64389

and it mentions this resource: https://www.php.net/manual/en/mysqli.installation.php
where we can see 'mysqlnd ' listed in the table as the default for PHP version 5.4.x and above. Note there is a 2nd one listed 'libmysqlclient' for PHP versions 3.0.x and lower, which may need to be supported somewhere.

fedeburo:
@MGatner i notice that most of the test are failing in:
Failed asserting that File exists: APPPATH/Language/en/Foobar.php\n or File exists: APPPATH/Language/es/Foobar.php\n. 

Im not sure what is this.. Should i make a new fork and new pull request?

app/Language/en/Foobar.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+use Config\App;
```
paulbalandan, line 3: I am not sure why you have these two test language files in your PR. And causes the failing tests. This is part of the Generator commands test and should in no way end up here. Is your fork not updated?

fedeburo:
Hi @paulbalandan, the fork was updated and now another pull request was made without those languages files. I dont know why and how those files were there...
Wait for your acception.
Another question: once the pull request is done and accepted, how long does this change will have been released in the new version? I need to know an approximately date in order to know if i have to do a manual modification in the different environments of the webapp that I am developing.

system/Database/MySQLi/Connection.php
```diff
@@ -157,8 +157,7 @@ public function connect(bool $persistent = false)
 					//
 					// https://secure.php.net/ChangeLog-5.php#5.6.16
 					// https://bugs.php.net/bug.php?id=68344
-					elseif (defined('MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT') && version_compare($this->mysqli->client_info, '5.6', '>='))
-					{
+					elseif (defined('MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT') && version_compare($this->mysqli->client_info, 'mysqlnd 5.6', '>=')){
```
paulbalandan, line 160:
```suggestion
					elseif (defined('MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT') && version_compare($this->mysqli->client_info, 'mysqlnd 5.6', '>='))
					{
```

system/Database/MySQLi/Connection.php
```diff
@@ -178,7 +177,7 @@ public function connect(bool $persistent = false)
 			)
 			{
 				// Prior to version 5.7.3, MySQL silently downgrades to an unencrypted connection if SSL setup fails
-				if (($clientFlags & MYSQLI_CLIENT_SSL) && version_compare($this->mysqli->client_info, '5.7.3', '<=')
+				if (($clientFlags & MYSQLI_CLIENT_SSL) && version_compare($this->mysqli->client_info, 'mysqlnd 5.7.3', '<=')
```
paulbalandan, line 180:
```suggestion
				if (($clientFlags & MYSQLI_CLIENT_SSL) && version_compare($this->mysqli->client_info, 'mysqlnd 5.7.3', '<=')
```

paulbalandan (commented):
Just minor cosmetic changes.

fedeburo:
> Just minor cosmetic changes.

Ok @paulbalandan , do i have to do a new pull request with those changes?

paulbalandan:
> Hi @paulbalandan, the fork was updated and now another pull request was made without those languages files. I dont know why and how those files were there...
> Wait for your acception.
> Another question: once the pull request is done and accepted, how long does this change will have been released in the new version? I need to know an approximately date in order to know if i have to do a manual modification in the different environments of the webapp that I am developing.

We're supposed to do a release two weeks ago but was delayed because of last minute refactors that need to be finished. Once all tidied up we can make a release. I cannot commit to an exact date but hopefully by next week or before the month ends.

paulbalandan (approved):
Just waiting for tests to pass. 😁

fedeburo:
> Just waiting for tests to pass. 😁

Me too... 😁

@paulbalandan @MGatner 
I wonder why the SQLSRV driver tests are failing if the only file that was modify is .../Database/MySQLi/Connection.php (nothing was modified under (SQLSRV) folder). 

It seems that some constants and functions are undefined on .../Database/SQLSRV/Connection.php file. 
image
I notice that these constants and functions are also undefined on the other app that I am working on.
How can I solve this?
Are the SQLSRV tests necessary in this pull request?

paulbalandan:
I have been trying in the past week to debug this sqlsrv problem. You can see my PR on that. Ideally we need all tests to pass to know that this fix is working. But since sqlsrv is failing in itself, your changes should not fail all tests other than for sqlsrv.

fedeburo:
> I have been trying in the past week to debug this sqlsrv problem. You can see my PR on that. Ideally we need all tests to pass to know that this fix is working. But since sqlsrv is failing in itself, your changes should not fail all tests other than for sqlsrv.

Thanks @paulbalandan. 
It seems that no other test have failed apart from the SQLSRV. 
Is this pull request accepted?

PR #4567, opened 2021-04-13
feat(cache): add deleteMatching method to remove multiple cache items

yassinedoghri:
### Add deleteMatching() method to cache service

**Description**

Maintaining a big number of cache keys with dependencies to each other can be a hassle with the current existing methods in the cache service.
This feature allows for deleting multiple cache items at once using a [glob pattern](https://en.wikipedia.org/wiki/Glob_(programming)):

```php
cache()->deleteMatching('some_prefix*');
cache()->deleteMatching('*some_suffix');
// etc.
```

To handle cache dependencies, CI4 users would be able to structure their cache items with the same key prefix and delete the whole "pack" in one go.

Initially discussed in this post: https://forum.codeigniter.com/thread-77789-post-380976.html and on [Slack](https://codeigniterchat.slack.com/archives/C47FK3GRY/p1617614168038800).

**Limitations**

The implementation is only effective for **file** and **redis/predis** handlers.

Because of certain limitations, memcached and wincache handlers do not implement the method.
Both would require to go through the whole cache store (if it ever could) and match each keys
against the glob pattern. That would inevitably result in performance issues.

Also, correct me if I'm wrong, I believe memcached and wincache are not used as much as file and redis handlers. That being said, knowing the limitations and documenting them well would allow users to have access to this feature. That could remove some of the heavy load of having to maintain cache items and dependencies between them.

Any feedback is welcome!

**Checklist:**

- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Cache/CacheInterface.php
+++ b/system/Cache/CacheInterface.php
@@ -59,17 +59,6 @@ interface CacheInterface
 	//--------------------------------------------------------------------
 
 	/**
-	 * Deletes items from the cache store matching a given pattern.
-	 *
-	 * @param string $pattern Cache items glob-style pattern
-	 *
-	 * @return mixed
-	 */
-	public function deleteMatching(string $pattern);
-
-	//--------------------------------------------------------------------
-
-	/**
 	 * Performs atomic incrementation of a raw stored value.
 	 *
 	 * @param string  $key    Cache ID
--- a/system/Cache/CacheInterface.php
+++ b/system/Cache/CacheInterface.php
@@ -59,6 +59,17 @@ interface CacheInterface
 	//--------------------------------------------------------------------
 
 	/**
+	 * Deletes items from the cache store matching a given pattern.
+	 *
+	 * @param string $pattern Cache items glob like pattern
+	 *
+	 * @return mixed
+	 */
+	public function deleteMatching(string $pattern);
+
+	//--------------------------------------------------------------------
+
+	/**
 	 * Performs atomic incrementation of a raw stored value.
 	 *
 	 * @param string  $key    Cache ID
--- a/system/Cache/CacheInterface.php
+++ b/system/Cache/CacheInterface.php
@@ -61,7 +61,7 @@ interface CacheInterface
 	/**
 	 * Deletes items from the cache store matching a given pattern.
 	 *
-	 * @param string $pattern Cache items glob like pattern
+	 * @param string $pattern Cache items glob-style pattern
 	 *
 	 * @return mixed
 	 */
--- a/system/Cache/Handlers/DummyHandler.php
+++ b/system/Cache/Handlers/DummyHandler.php
@@ -89,6 +89,20 @@ class DummyHandler extends BaseHandler
 	//--------------------------------------------------------------------
 
 	/**
+	 * Deletes items from the cache store matching a given pattern.
+	 *
+	 * @param string $pattern Cache items glob like pattern
+	 *
+	 * @return boolean
+	 */
+	public function deleteMatching(string $pattern)
+	{
+		return true;
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
 	 * Performs atomic incrementation of a raw stored value.
 	 *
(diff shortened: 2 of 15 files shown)


system/Cache/CacheInterface.php
```diff
@@ -58,6 +58,17 @@ public function delete(string $key);
...
+	 * Deletes items from the cache store matching a given pattern.
+	 *
+	 * @param string $pattern Cache items glob like pattern
+	 *
+	 * @return mixed
+	 */
+	public function deleteMatching(string $pattern);
+
+	//--------------------------------------------------------------------
+
```
paulbalandan, line 71: You cannot add this to the interface as it would be breaking for others using this interface. Perhaps adding a default implementation of this method in the base class, instead?

system/Cache/Handlers/FileHandler.php
```diff
@@ -160,6 +160,30 @@ public function delete(string $key)
...
+	 *
+	 * @return boolean
+	 */
+	public function deleteMatching(string $pattern)
+	{
+		$success = true;
+
+		foreach (glob($this->path . $pattern) as $filename)
+		{
+			if (! is_file($filename) || ! unlink($filename))
```
paulbalandan, line 176: Perhaps suppress the error here
```suggestion
			if (! is_file($filename) || ! @unlink($filename))
```

system/Cache/Handlers/PredisHandler.php
```diff
@@ -184,6 +185,30 @@ public function delete(string $key)
...
+	 *
+	 * @param string $pattern Cache items glob like pattern
+	 *
+	 * @return boolean
+	 */
+	public function deleteMatching(string $pattern)
+	{
+		$success = true;
+
+		foreach (new Iterator\Keyspace($this->redis, $pattern) as $key)
```
paulbalandan, line 199: Please import the whole class name. Rector will complain on this style.

system/Cache/Handlers/DummyHandler.php
```diff
@@ -88,6 +88,20 @@ public function delete(string $key)
...
+	/**
+	 * Deletes items from the cache store matching a given pattern.
+	 *
+	 * @param string $pattern Cache items glob like pattern
+	 *
+	 * @return boolean
+	 */
+	public function deleteMatching(string $pattern)
+	{
+		return true;
```
paulbalandan, line 100: Why is this returning true while others are false?

yassinedoghri: The method returns true for success. The dummy handler always returns true so it doesn't break the code.

For memcached and wincache, as there are limitations, it doesn't do anything but would break if used in a condition.
It is so that users wouldn't use it. But, now that I think of it, maybe it is wiser to throw an exception stating that the method isn't implemented for those specific handlers.

paulbalandan: I think we should go for throwing exceptions if not implemented.

system/Cache/CacheInterface.php
yassinedoghri: I understand the concern though I think it should be in the interface defined with the other cache methods. Having it anywhere else would be kind of a hack.
As for others using it, and correct me if I'm wrong, if this feature is merged, it would be at first in the develop branch that is meant for this kind of "unstable" change. On the next release, this would have to be communicated as potential breaking change.

Anyways, that's just my two cents, and I am fine with changing as requested!

paulbalandan (changes requested):
I maintain my stance on not adding this to the interface. Following semver, minor releases should have no breaking changes.

Furthermore, please adjust your failing tests.

yassinedoghri:
@paulbalandan I've made the requested changes. Sorry for the failing tests, didn't pay attention... I think all should be ok now, could you please take a look?

yassinedoghri:
Apparently SQLSRV tests are failing because no connection could be established to the database...

paulbalandan (approved):

PR #4570, opened 2021-04-14
[Rector] Apply Full PHP 7.3 Rector set list

samsonasik:
This try apply full PHP 7.3 Rector set list.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -246,7 +246,7 @@ class FormatRules
 	 */
 	public function valid_json(string $str = null): bool
 	{
-		json_decode($str);
+		json_decode($str, false, 512, JSON_THROW_ON_ERROR);
 		return json_last_error() === JSON_ERROR_NONE;
 	}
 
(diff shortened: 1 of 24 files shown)


system/Validation/FormatRules.php
```diff
@@ -246,7 +246,7 @@ public function valid_base64(string $str = null): bool
 	 */
 	public function valid_json(string $str = null): bool
 	{
-		json_decode($str);
+		json_decode($str, false, 512, JSON_THROW_ON_ERROR);
 		return json_last_error() === JSON_ERROR_NONE;
```
paulbalandan, line 250: Should this be handled by a try-catch now?

samsonasik: Yes, probably need to be handled in the `rector` itself when next is `json_last_error()` check, try with Exception with message `json_last_error_msg()`

samsonasik:
closing for now as possibly break existing functionality.

PR #4573, opened 2021-04-15
Refactor Entity Class

mostafakhudair:


samsonasik:
What is the benefit of new namespace? next we need to move "Model" classes to Model namespace? The current usage is easier to access (no need deep namespace extends) for mostly db based apps.

mostafakhudair:
@samsonasik whats happend, i just followed your tut https://samsonasik.wordpress.com/2015/09/16/practical-git-4-rebasing-conflicted-task-branch-against-primary-branch/

sfadschm:
I had the same issue with MGatner's Schemas repository yesterday. I did a rebase with Github Desktop which has always worked before, but this time, the main branch commits appeared again. Still don't know why...

mostafakhudair:
I fixed it by deleting the repo directory in the local machine and then reclone it then rebase and it's worked fine

PR #4577, opened 2021-04-16
Change Entity Namespace

mostafakhudair:
Change Entity class namespace
Refactor Entity casts

- [x] User guide updated
- [x] Conforms to style guide


mostafakhudair:
@paulbalandan i'm done where, would you like to review it?

samsonasik:
What is the benefit of new namespace? next we need to move "Model" classes to Model namespace? The current usage is easier to access (no need deep namespace extends) for mostly db based apps.

mostafakhudair:
The benefit is not about the namespace, it's about organize system directories/sub-directories

paulbalandan (approved):

paulbalandan:
I'm done here. I would like @MGatner to take  final look.

system/Entity.php
```diff
@@ -11,612 +11,22 @@
...
-	 */
-	public function cast(bool $cast = null)
-	{
-		if (null === $cast)
-		{
-			return $this->_cast;
-		}
-		$this->_cast = $cast;
-		return $this;
+		parent::__construct($data);
```
MGatner, line 30: What is the point of this?

mostafakhudair: Nothing,, removed :D

MGatner (approved):
Looks good to me! Any final requests or concerns?

mostafakhudair:
should i add Services::entity() in a separated PR? or push it in this PR?

MGatner:
Let's hold off on a service for now. I'd like some resolution to our interface discussion before we start adding more services for classes that don't use interfaces.

MGatner (approved):

PR #4586, opened 2021-04-19
extra </li> in tabs area

jbrahy:
There was just an extra end tag in the HTML.


jbrahy:
Only should have been the removing the extra </li> I committed the other lang="en" and then reverted it.

samsonasik (approved):

samsonasik:
Thank you @jbrahy

PR #4588, opened 2021-04-20
Feature: Escaping array dot notation

iRedds:
**Description**
Sometimes an array key can contain a dot. In this case, you cannot access the value of this key using dot notation.
This PR makes available keys containing a dot, with dot notation.

```php
$data = [
    'foo' => [
        'bar.baz' => 23
    ],
];

// Returns: 23
$barBaz = dot_array_search('foo.bar\.baz', $data);
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -28,7 +28,7 @@ if (! function_exists('dot_array_search'))
 	{
 		$segments = preg_split('/(?<!\\\)\./', rtrim($index, '* '), 0, PREG_SPLIT_NO_EMPTY);
 
-		$segments = array_map(function ($key) {
+		$segments = array_map(static function ($key) {
 			return str_replace('\.', '.', $key);
 		}, $segments);
 
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -26,7 +26,11 @@ if (! function_exists('dot_array_search'))
 	 */
 	function dot_array_search(string $index, array $array)
 	{
-		$segments = explode('.', rtrim(rtrim($index, '* '), '.'));
+		$segments = preg_split('/(?<!\\\)\./', rtrim($index, '* '), 0, PREG_SPLIT_NO_EMPTY);
+
+		$segments = array_map(function ($key) {
+			return str_replace('\.', '.', $key);
+		}, $segments);
 
 		return _array_search_dot($segments, $array);
 	}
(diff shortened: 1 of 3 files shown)


system/Helpers/array_helper.php
```diff
@@ -26,7 +26,11 @@
 	 */
 	function dot_array_search(string $index, array $array)
 	{
-		$segments = explode('.', rtrim(rtrim($index, '* '), '.'));
+		$segments = preg_split('/(?<!\\\)\./', rtrim($index, '* '), 0, PREG_SPLIT_NO_EMPTY);
```
paulbalandan, line 29: Should this be four backslashes?

system/Helpers/array_helper.php
```diff
@@ -26,7 +26,11 @@
...
+
+		$segments = array_map(function ($key) {
```
paulbalandan, line 31:
```suggestion
		$segments = array_map(static function ($key) {
```

user_guide_src/source/helpers/array_helper.rst
```diff
@@ -57,6 +57,23 @@ The following functions are available:
         // Returns: 23
         $baz = dot_array_search('foo.*.baz', $data);
 
+    If the array key contains a dot, then the key can be escaped with a backslash::
+        $data = [
```
paulbalandan, line 61: Please fix the check warnings in the user guide.

system/Helpers/array_helper.php
iRedds: It doesn't matter in this case. 4 and 3 backslashes in a string will be converted to 2 backslashes.

system/Helpers/array_helper.php
iRedds: Why did you suggest using ``static``?

user_guide_src/source/helpers/array_helper.rst
iRedds: Thanks. Fixed.

system/Helpers/array_helper.php
paulbalandan: Using static here is for micro-optimization.

system/Helpers/array_helper.php
paulbalandan: Hmm., ok.

paulbalandan (approved):

MGatner (approved):

PR #4589, opened 2021-04-20
BaseModel - Add public getIdValue() method

najdanovicivan:
**Description**
This makes it possible to use protected idValue() method from outside of the model to get the id value of the entity. 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -463,8 +463,25 @@ abstract class BaseModel
 	 * @param array|object $data Data
 	 *
 	 * @return integer|array|string|null
+	 *
+	 * @deprecated Add an override on getIdValue() instead. Will be removed in version 5.0.
 	 */
 	abstract protected function idValue($data);
+
+	/**
+	 * Public getter to return the id value using the idValue() method
+	 * For example with SQL this will return $data->$this->primaryKey
+	 *
+	 * @param array|object $data
+	 *
+	 * @return array|integer|string|null
+	 *
+	 * @todo: Make abstract in version 5.0
+	 */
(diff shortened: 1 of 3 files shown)


MGatner:
It would be great if we could find a way to deprecate the protected version and move towards this instead, but I don't know how to handle that with an abstract method.

najdanovicivan:
The only way we can introduce deprecation is to make the abstract idValue() method deprecated. Then we change in all places to use getIdValue() instead. 

After that we change getIdValue() to also be abstract function and implement in the Model class. the old idValue() method will just pass the call to getIdValue() so for some time we'll have both methods available. In any case it will be breaking change as introducing new abstract method in BaseModel will require the method to be implemented in all child classes

MGatner:
Understood! Please add a note to the User Guide changelog about this new function - we can handle deduplicating in version 5.

najdanovicivan:
@MGatner I'll add the deprecation and modify the existing model to use the new method

najdanovicivan:
If I am to add a new method in the interface it will be BC in any case.  Since all the custom models will have to be updated anyway

najdanovicivan:
I've found a way on how to do deprecation without introducing a BC. The only issue that I have to define both methods in base model. 

The only downside of this approach is that any of those methods won't be forced to be defined in the model class. And if there is no single of them defined the infinitive loop will occur.

MGatner:
I'd rather have the duplicate method than the side effects you notes. Can you revert to abstract and resolve merge conflicts? Then I think this is good.

Issue #4137, opened 2021-01-19
Bug: Debug Toolbar - Memory Leak - Allocation Exception


najdanovicivan:
**Describe the bug**
In CodeIgniter\Debug\Toolbar  run() method we have the following code to print variables
```
$varData[esc($key)] = is_string($value) ? esc($value) : '<pre>' . esc(print_r($value, true)) . '</pre>';
```
The issue here is that print_r() on object may lead to very high memory usage when $value is object or an array of objects

**CodeIgniter 4 version**
github-develop 

**Affected module(s)**
Debug\Toolbar

**Expected behavior, and steps to reproduce if appropriate**

I don't know if this is a good practice but in some of my controllers I'm passing the controller to the view by reference

```
$data['controller'] = &this;
```

I also usually pass array of entity object as $data['items']

In most cases this will work. But if there are a lot of public properties inside the controller or the entity. print_r() will try printing out everything and completely fill the available memory. In the worst possible example when there is for example entity object which have a reference pointer to child entity object and the child entity object has a reference pointer to parent object it will lead to infinitive loop. 

I propose doing a check on $value if it is an object. In case of array check if array items are objects. So this will be a recursive method processing the array. If $value is an object we can then check if method toArray() exist and render the value returned from toArray method otherwise we just return string `object (className)`

paulbalandan:
I am not well-versed in memory leaks issue but is changing to `var_export` not feasible? Also, in terms of infinite looping, I think I saw an example in a dumping function (cannot remember if it is `print_r` or `var_export` or other else) that if the loop exists, it will just print `RECURSION` on that looping part.

najdanovicivan:
var_export might be a solution as it ignores the recursion. I'll check it out to see if how it will perform

najdanovicivan:
It won't be possible to use var_export as it throws an ErrorException with message `var_export does not handle circular references`

kenjis:
How about this?
https://github.com/pradosoft/prado/blob/master/framework/Util/TVarDumper.php

najdanovicivan:
I’ve looked at TVarDumper but I don like adding additional dependency as it bit tricky to manage them.

CI4 needs to be able to run without composer so it will need scripts for copying those in framework

najdanovicivan:
Doctrine debug might be used as an idea how to do it in framework implementation

najdanovicivan:
[...]

PR #4590, opened 2021-04-20
Debug/Toolbar - Memory issue fix

najdanovicivan:
**Description**
This is the very simple method to avoid memory leak issue #4137 when entity objects with circular references are passed to the view. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -109,7 +109,7 @@ class Toolbar
 			{
 				foreach ($items as $key => $value)
 				{
-					$varData[esc($key)] = is_string($value) ? esc($value) : '<pre>' . esc(print_r($value, true)) . '</pre>';
+					$varData[esc($key)] = is_string($value) ? esc($value) : '<pre>' . esc($this->processVar($value)) . '</pre>';
 				}
 			}
 
@@ -467,5 +467,28 @@ class Toolbar
 		}
 
 		return $output;
+	}
+
+	/**
+	 * Process the varable to string for display
+	 *
+	 * @param mixed $var Variable
+	 *
+	 * @return boolean|string
+	 */
+	protected function processVar($var)
+	{
+		if (is_object($var))
+		{
+			$var = is_callable([$var, 'toArray']) ? $var->toArray() : get_class($var);
+		}
+		if (is_array($var))
+		{
+			foreach ($var as &$aVar)
+			{
+				$aVar = $this->processVar($aVar);
+			}
+		}
+		return print_r($var, true);
 	}
 }
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -22,6 +22,7 @@ use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
 use Config\Services;
 use Config\Toolbar as ToolbarConfig;
+use Kint\Kint;
 
 /**
  * Debug Toolbar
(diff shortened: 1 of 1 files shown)


system/Debug/Toolbar.php
```diff
@@ -468,4 +468,27 @@ protected function format(string $data, string $format = 'html'): string
...
 	}
+
+	/**
+	 * Process the varable to string for display
+	 *
+	 * @param mixed $var Variable
+	 *
+	 * @return boolean|string
+	 */
+	protected function processVar($var)
```
MGatner, line 479: Let's make this private for now, unless there are compelling reasons to allow access to this method.

system/Debug/Toolbar.php
```diff
@@ -468,4 +468,27 @@ protected function format(string $data, string $format = 'html'): string
...
+	{
+		if (is_object($var))
+		{
+			$var = is_callable([$var, 'toArray']) ? $var->toArray() : get_class($var);
```
MGatner, line 483: This no longer a) tells us the object class for array-ables nor b) allows displaying valid, non-recursive objects that are not array-able.

MGatner (changes requested):
I like the concept but not the end result. We need to find some way to catch as much as we can. Does Kint (already included) not have some solution for this in place?

najdanovicivan:
@MGatner  I managed to make Kint work inside debug bar. The issue I had is that Kint render will return JS and CSS in as the string and since those are set as text via JS. The javascript from Kint will never be executed. So I had to play a bit with it to add the JS and CSS into the page head instead and then drop those when it's being rendered in the debug bar

MGatner:
I'll take a look - if it is working then that's probably good. Kint also supports prefix modifiers to adjust the behavior - I think we could probably use `~` for "text output", or maybe `@` to "return output". Play around with those and see what you think. 

https://kint-php.github.io/kint/advanced/

system/Debug/Toolbar.php
```diff
@@ -343,12 +363,19 @@ public function prepare(RequestInterface $request = null, ResponseInterface $res
 				return;
 			}
 
+			$oldKintMode        = Kint::$mode_default;
+			Kint::$mode_default = Kint::MODE_RICH;
+			$kintScirpt         = @Kint::dump('');
+			Kint::$mode_default = $oldKintMode;
+			$kintScirpt         = substr($kintScirpt, 0, strpos($kintScirpt, '</style>') + 8 );
```
MGatner, line 370:
```suggestion
			$kintScript         = substr($kintScript, 0, strpos($kintScirpt, '</style>') + 8 );
```
Typo in the variable name.

system/Debug/Toolbar.php
```diff
@@ -343,12 +363,19 @@ public function prepare(RequestInterface $request = null, ResponseInterface $res
 				return;
 			}
 
+			$oldKintMode        = Kint::$mode_default;
+			Kint::$mode_default = Kint::MODE_RICH;
+			$kintScirpt         = @Kint::dump('');
```
MGatner, line 368: Ah, I see you already found the `@`.

system/Debug/Toolbar.php
```diff
@@ -343,12 +363,19 @@ public function prepare(RequestInterface $request = null, ResponseInterface $res
...
+			Kint::$mode_default = $oldKintMode;
+			$kintScirpt         = substr($kintScirpt, 0, strpos($kintScirpt, '</style>') + 8 );
+
 			$script = PHP_EOL
 					. '<script type="text/javascript" {csp-script-nonce} id="debugbar_loader" '
 					. 'data-time="' . $time . '" '
 					. 'src="' . site_url() . '?debugbar"></script>'
 					. '<script type="text/javascript" {csp-script-nonce} id="debugbar_dynamic_script"></script>'
 					. '<style type="text/css" {csp-style-nonce} id="debugbar_dynamic_style"></style>'
+					. $kintScirpt
```
MGatner, line 378:
```suggestion
					. $kintScript
```

MGatner (changes requested):
I think this should be fine. It definitely adds some complexity. See what you think of the text output option and if that is a worthy trade-off.

najdanovicivan:
I am using @ prefix to return output. I tried text output as well but it looks nasty.  

I was unable to make it work by chaining prefixes eg @~dd() will not work for me. 

Another problem with text output is it can get very long especially when controller is being passed to the view and if any of the variables contains HTML it may break the <pre> tag

MGatner (approved):

PR #4591, opened 2021-04-20
EntityCast - Boolean - Use Filter Var

najdanovicivan:
**Description**
We have bool cast just cast the type to (bool) this changes it to use filter_var instead so values like `true` , `false`, `yes`, `no` can be casted properly to boolean.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/EntityCast/CastAsBoolean.php
+++ b/system/EntityCast/CastAsBoolean.php
@@ -22,6 +22,6 @@ class CastAsBoolean extends AbstractCast
 	 */
 	public static function get($value, array $params = []): bool
 	{
-		return (bool) $value;
+		return filter_var($value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) ?? true;
 	}
 }


system/EntityCast/CastAsBoolean.php
```diff
@@ -22,6 +22,6 @@ class CastAsBoolean extends AbstractCast
 	 */
 	public static function get($value, array $params = []): bool
 	{
-		return (bool) $value;
+		return filter_var($value, FILTER_VALIDATE_BOOLEAN);
```
samsonasik, line 25: Probably use `is_string($value) && in_array($value, ['true', 'false', 'yes', 'no'], true)` check first for the filter_var(), and fallback to `(bool) $value` otherwise ?

```suggestion
              if (is_string($value) && in_array($value, ['true', 'false', 'yes', 'no'], true))
              {
                  return filter_var($value, FILTER_VALIDATE_BOOLEAN);
              }

             return (bool) $value;
```

najdanovicivan: but it will also require strtolower($value) to handle it case insensitive

I suppose this is due to handling integers as true

What about doing 

```
filter_var($value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) ?? true;
```

From PHP doc

If FILTER_NULL_ON_FAILURE is set, false is returned only for "0", "false", "off", "no", and "", and null is returned for all non-boolean values.

najdanovicivan: I just did test with FILTER_NULL_ON_FAILURE it will return false only for the following

```
null
0
'0'
'no' (case insensitive)
'false' (case insensitive)
```

najdanovicivan: The only difference I see with this approach is 

```
$value = [];
var_dump((bool)$value); //false
var_dump(filter_var($value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) ?? true); //true
````

I'm not sure which one should be correct behavior but can be easily solved with:
```
var_dump(filter_var($value ?: false, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) ?? true);
```

MGatner:
I would prefer to leave it as it is. PHP booleans are confusing (as this thread demonstrates) so I think sticking to the docs will help developers know what to expect: https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting

If any other cast of `(bool) 'no'` would be `true` then why should our Entities return `false`?

najdanovicivan:
> I would prefer to leave it as it is. PHP booleans are confusing (as this thread demonstrates) so I think sticking to the docs will help developers know what to expect: https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting
> 
> If any other cast of `(bool) 'no'` would be `true` then why should our Entities return `false`?

The issue happens with nosql databases like ElasticSearch or Mongo where db will return string `false` instead of bool value

MGatner:
See also https://www.php.net/manual/en/function.boolval.php

najdanovicivan:
What about adding a new cast 'strict-bool' which will do the casting on both setter and getter with filter var ?

MGatner:
> The issue happens with nosql databases like ElasticSearch or Mongo where db will return string false instead of bool value

Hmm that is unfortunate. Where is that coming from? Native PHP drivers or extensions?

Would we want to "fix" this for all Results at the driver level? E.g. `$result = $query->get()->getResultArray()` would still have `'false'` which could cause unexpected issues mixing Entities and raw results.

najdanovicivan:
The 'false' can come form ElasticSearch which accepts both 'false' and false in json and it index the value properly as bool but reading it back it can return it either as string or as bool

MGatner:
I really don't think this is an Entity issue but a problem we are inheriting from PHP generally and these database drivers specifically. I would like to see a little research on how other frameworks/libraries handle this, but my instinct is still the same: we either should not "fix" this (let the developers handle it) or we should handle it universally before Entity ever comes into play.

paulbalandan:
Closing, as this seems to be needing to be seen from a different angle. Feel free to open a new PR with a new implementation.

PR #4596, opened 2021-04-21
Simplify Cookie Class

mostafakhudair:
Since we use php min version `7.3`, why not passing all cookie options to the constructor itself directly.
Removed `Cookie::create` while `new Cookie` do the same thing directly.
Renamed `Cookie::getExpiresTimestamp` to `Cookie::getExpires` to be short and refer to it's property,
Renamed `Cookie::withExpiresAt` to `Cookie::withExpires` to fit it's property and also it may return expired cookie.

I think this is last change to new Cookie class should be done before the upcoming release

@paulbalandan what do you think about this changes? is it keeps the interface "fluent"? Thanks in advance

### DIFF
--- a/system/Cookie/Cookie.php
+++ b/system/Cookie/Cookie.php
@@ -27,7 +27,7 @@ use LogicException;
  * reassign this new instance to a new variable to capture it.
  *
  * ```php
- * $cookie = Cookie::create('test_cookie', 'test_value');
+ * $cookie = new Cookie('test_cookie', 'test_value');
  * $cookie->getName(); // test_cookie
  *
  * $cookie->withName('prod_cookie');
@@ -149,8 +149,8 @@ class Cookie implements ArrayAccess, CloneableCookieInterface
 			$newDefaults = $config;
 		}
 
-		// This array union ensures that even if passed `$config`
-		// is not `CookieConfig` or `array`, no empty defaults will occur.
+		// This array union ensures that even if passed `$config` is not
+		// `CookieConfig` or `array`, no empty defaults will occur.
 		self::$defaults = $newDefaults + $oldDefaults;
 
 		return $oldDefaults;
@@ -197,21 +197,19 @@ class Cookie implements ArrayAccess, CloneableCookieInterface
 			$data[strtolower($attr)] = $val;
 		}
 
-		return static::create($name, $value, $data);
+		return new static($name, $value, $data);
 	}
 
 	/**
-	 * Create Cookie objects on the fly.
+	 * Construct a new Cookie instance.
 	 *
-	 * @param string               $name
-	 * @param string               $value
-	 * @param array<string, mixed> $options
+	 * @param string               $name    The cookie's name
+	 * @param string               $value   The cookie's value
+	 * @param array<string, mixed> $options The cookie's options
 	 *
 	 * @throws CookieException
-	 *
-	 * @return static
 	 */
-	public static function create(string $name, string $value = '', array $options = [])
+	final public function __construct(string $name, string $value = '', array $options = [])
 	{
 		$options += self::$defaults;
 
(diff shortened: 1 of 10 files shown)


system/Cookie/Cookie.php
```diff
@@ -445,7 +401,7 @@ public function getOptions(): array
 			'domain'   => $this->domain,
 			'secure'   => $this->secure,
 			'httponly' => $this->httponly,
-			'samesite' => $this->samesite ?: ucfirst(self::SAMESITE_LAX),
+			'samesite' => $this->samesite,
```
paulbalandan, line 404: Please revert. This is for fallback catch for the blank samesite setting.

system/Cookie/Cookie.php
```diff
@@ -197,25 +197,23 @@ public static function fromHeaderString(string $cookie, bool $raw = false)
...
-	 *
-	 * @return static
 	 */
-	public static function create(string $name, string $value = '', array $options = [])
+	final public function __construct(string $name, string $value = '', array $options = [])
 	{
 		$options += self::$defaults;
 
-		$options['expires'] = static::convertExpiresTimestamp($options['expires']);
+		$options['expires'] = $this->convertExpiresTimestamp($options['expires']);
```
paulbalandan, line 216: Any reason why making the method as not static and available only for the object instance? My intention here is before someone instantiates a `Cookie` object and when given a "non-uniform" timestamp, a dev can simply use `Cookie::convertExpiresTimestamp('blah')` before passing this to the constructor in order to lessen errors.

system/Cookie/Cookie.php
```diff
@@ -197,25 +197,23 @@ public static function fromHeaderString(string $cookie, bool $raw = false)
...
+		return new static($name, $value, $data);
 	}
 
 	/**
-	 * Create Cookie objects on the fly.
+	 * Construct a new Cookie instance.
 	 *
-	 * @param string               $name
-	 * @param string               $value
-	 * @param array<string, mixed> $options
```
paulbalandan, line 208: Why remove the array properties?

paulbalandan (changes requested):
Your PR title is for changing the Cookie constructor yet you clearly have two PRs here.

1. I approve the simplification of the constructor call for the Cookie as that is leaning towards 7.3.
2. Unless you can give me reason why `convertExpiresTimestamp` is no longer accessible via class, I cannot accept the removal of `static`.
3. For the `expires` changes (which clearly misplaced here in this PR), I understand that you want to make things simple here but I'm on the side of making things "fluent" on the side of the developer using this API. For example, if I want to make an "expired" Cookie object, which would I choose as more fluent -- `$cookie->withExpires(0)` or `$cookie->withExpired()`. I believe even non-devs would choose the 2nd option as that is clearer. Another one is I want to get the expires string suitable for `Set-Cookie` headers. I would definitely want to use `$cookie->getExpiresString()` over `$cookie->getExpires(false)`. The `getExpiresTimestamp()` on the other hand is a side effect of having `getExpiresString()` so as to differentiate the two expires getters. It will be clear that one gets the string expires while the other gets the integer value.
4. I don't get why the tests on Array Access were removed when there's no changes to that in the class. By removing those tests I'm assuming that the `Cookie`, after this PR change, now allows mutability which is in contrast to its immutable nature.

system/Cookie/Cookie.php
paulbalandan: On `setrawcookie()` and `setcookie()` calls, empty samesites would cause the cookie to be rejected for creation.

system/Cookie/Cookie.php
mostafakhudair: Thanks for explaining this one, and everything

mostafakhudair:
`Cookie::getExpiresString` is still here and has not been removed.

mostafakhudair:
cannot set property via array access on class constructor this case syntax error `new Cookie('cookie', 'monster')['expires'] = 7200;`
`syntax error, unexpected token "["`

system/Cookie/Cookie.php
mostafakhudair: by wrong, reverted

user_guide_src/source/libraries/cookies.rst
```diff
@@ -65,7 +51,7 @@ There are currently five (5) ways to create a new ``Cookie`` value object.
 
     // Supplying a Set-Cookie header string
     $cookie = Cookie::fromHeaderString(
-        'remember_token=f699c7fd18a8e082d0228932f3acd40e1ef5ef92efcedda32842a211d62f0aa6; Path=/; Secure; HttpOnly; SameSite=Lax',
+        'remember_token=f699c7fd18a8e082d0228932f3acd40e1ef5ef92efcedda32842a211d62f0aa6; Path=/; Secure; HTTPOnly; SameSite=Lax',
```
paulbalandan, line 54: This is format of a `Set-Cookie` header.

paulbalandan (changes requested):

MGatner:
I assume this is yet another PR that will need to be resolved before release. Please prioritize this.

paulbalandan (changes requested):
Unless you'll revert all changes to `expires` we can get this merged. Otherwise, I'm afraid that is a blocker for merge.

On a side note, I think I made a mistake with `withExpiresAt($expires = 0)` by setting a default value. This should not be the case. It should be only `withExpiresAt($expires)`.

user_guide_src/source/libraries/cookies.rst
```diff
@@ -581,7 +549,7 @@ Class Reference
         :rtype: ``Cookie``
         :returns: new ``Cookie`` instance
 
-        Creates a new Cookie that will expire from the browser.
+        Creates a new Cookie that will expire the cookie.
```
paulbalandan, line 552: Please revert

user_guide_src/source/libraries/cookies.rst
```diff
@@ -615,17 +583,17 @@ Class Reference
...
-    .. php:method:: withHttpOnly([bool $httpOnly = true])
+    .. php:method:: withHTTPOnly([bool $httponly = true])
 
-        :param bool $httpOnly:
+        :param bool $httponly:
         :rtype: ``Cookie``
         :returns: new ``Cookie`` instance
 
-        Creates a new Cookie with new "HttpOnly" attribute.
+        Creates a new Cookie with new "HTTPOnly" attribute.
```
paulbalandan, line 592: Please revert

system/Cookie/CookieInterface.php
```diff
@@ -106,7 +106,7 @@ public function getValue(): string;
 	 *
 	 * @return integer
 	 */
-	public function getExpiresTimestamp(): int;
+	public function getExpires(): int;
```
paulbalandan, line 109: I'll accept your change for `withExpires($expires)`. But for here, can we reach a compromise and meet halfway?

I want this to be at least `getExpiresTime()` to differentiate it clearly from `getExpiresString()`.

mostafakhudair: done

user_guide_src/source/libraries/cookies.rst
```diff
@@ -486,29 +472,11 @@ Class Reference
 
         Create a new Cookie instance from a ``Set-Cookie`` header.
 
-    .. php:staticmethod:: create(string $name[, string $value = ''[, array $options = []]])
+    .. php:staticmethod:: __construct(string $name[, string $value = ''[, array $options = []]])
```
paulbalandan, line 475:
```suggestion
    .. php:method:: __construct(string $name[, string $value = ''[, array $options = []]])
```

paulbalandan (approved):

paulbalandan:
Thank you, @mostafakhudair

mostafakhudair:
@paulbalandan Many thanks to you for you'r professional review and your patience

MGatner:
Thanks all around! I think we are now finally cleared for release.

PR #4600, opened 2021-04-23
[Rector] Apply Full PHP 7.3 Rector Set List

samsonasik:
Using : 

```php
	$parameters->set(Option::SETS, [
		SetList::PHP_73,
	]);
```

to apply full php 7.3 rector set list.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -46,6 +46,7 @@ parameters:
 		- '#Return type \(bool\) of method CodeIgniter\\HTTP\\Files\\UploadedFile::move\(\) should be compatible with return type \(CodeIgniter\\Files\\File\) of method CodeIgniter\\Files\\File::move\(\)#'
 		- '#Return type \(bool\) of method CodeIgniter\\Test\\TestLogger::log\(\) should be compatible with return type \(null\) of method Psr\\Log\\LoggerInterface::log\(\)#'
 		- '#Parameter \#2 \$assoc of function json_decode expects bool, null given#'
+	reportUnmatchedIgnoredErrors: false
 	parallel:
 		processTimeout: 300.0
 	scanDirectories:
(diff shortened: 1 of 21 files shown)


phpstan.neon.dist
```diff
@@ -45,6 +45,8 @@ parameters:
 		- '#Return type \(bool\) of method CodeIgniter\\Log\\Logger::(emergency|alert|critical|error|warning|notice|info|debug|log)\(\) should be compatible with return type \(null\) of method Psr\\Log\\LoggerInterface::(emergency|alert|critical|error|warning|notice|info|debug|log)\(\)#'
 		- '#Return type \(bool\) of method CodeIgniter\\HTTP\\Files\\UploadedFile::move\(\) should be compatible with return type \(CodeIgniter\\Files\\File\) of method CodeIgniter\\Files\\File::move\(\)#'
 		- '#Return type \(bool\) of method CodeIgniter\\Test\\TestLogger::log\(\) should be compatible with return type \(null\) of method Psr\\Log\\LoggerInterface::log\(\)#'
+		- '#Parameter \#2 \$assoc of function json_decode expects bool, null given#'
+	reportUnmatchedIgnoredErrors: false
```
paulbalandan, line 49: I think this should be set as true so that we can remove errors no longer applicable.

paulbalandan (commented):
I like the JSON_THROW_ON_ERROR flag addition but I'm worrying someone's code will now break and throw the blame on us. I think this should be on the changelog to say that json operations now throw errors to force users to rectify their code.

phpstan.neon.dist
samsonasik: yes, it currently shows notice in php < 8

samsonasik:
I can create another alternative PR for allow `JsonThrowOnError`, with skip `StringifyStrNeedlesRector`. Let see if @MGatner have other opinion.

MGatner:
Generally I agree about throwing new errors but looking through these specific instances I think a failed operation will already cause other errors that will just be harder to track down. The exception may be the Toolbar changes but that's only for dev instances anyways when (I believe) we want to be throwing more errors generally.

samsonasik:
@MGatner skipping with negation seems not possible as I see right now, eg:

```php
^((?!system\/Debug\/Toolbar\.php).)*$
```

Manually skip many files seems not the right way.

I think the safest solution is the alternative PR https://github.com/codeigniter4/CodeIgniter4/pull/4601 which skip both `JsonThrowOnError` and `StringifyStrNeedlesRector`.

samsonasik:
Closing in favor of #4601

PR #4601, opened 2021-04-23
[Rector] Apply Full PHP 7.3 Rector Set List (Skip JsonThrowOnErrorRector & StringifyStrNeedlesRector)

samsonasik:
Alternative of https://github.com/codeigniter4/CodeIgniter4/pull/4600 with skip JsonThrowOnError & StringifyStrNeedlesRector.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -21,7 +21,9 @@ use Rector\EarlyReturn\Rector\Foreach_\ChangeNestedForeachIfsToEarlyContinueRect
 use Rector\EarlyReturn\Rector\If_\ChangeIfElseValueAssignToEarlyReturnRector;
 use Rector\EarlyReturn\Rector\If_\RemoveAlwaysElseRector;
 use Rector\EarlyReturn\Rector\Return_\PreparedValueToEarlyReturnRector;
-use Rector\Php73\Rector\FuncCall\ArrayKeyFirstLastRector;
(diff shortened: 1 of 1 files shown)


paulbalandan (commented):
I don't understand why the string cast is needed here.

samsonasik:
That's based on the following https://wiki.php.net/rfc/deprecations_php_7_3#string_search_functions_with_integer_needle 

For function that return string from helper, eg, `singular()`, the helpers directory seems need to be registered to `Option::BOOTSTRAP_FILES`   to make skip function that already return string to no need re-cast.

samsonasik:
We can skip `StringifyStrNeedlesRector` if it is unneeded check.

paulbalandan:
Looking at the previous lines where the string casted variables arise, it seems these are already string so casting is unnecessary.

samsonasik:
Ok, I skipped the `StringifyStrNeedlesRector` as well.

paulbalandan (approved):
Seems this is the better choice for now.

Just wondering, what are in the Set List for 7.3?

samsonasik:
That's collection of rector rules for php 7.3, you can check at:
https://github.com/rectorphp/rector/blob/main/config/set/php73.php

samsonasik:
I am merging it then, thank you @paulbalandan @MGatner for the review.

PR #4604, opened 2021-04-23
Optimizations for Autoloader

paulbalandan:
**Description**
Optimizations include:
- Make loading classmaps as a separate method
- Load classmaps from Composer, if available

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter\Autoloader;
 
+use Composer\Autoload\ClassLoader;
 use Config\Autoload;
 use Config\Modules;
 use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


system/Autoloader/Autoloader.php
```diff
@@ -210,7 +191,24 @@ public function removeNamespace(string $namespace)
...
+	 *
+	 * @param string $class
+	 *
+	 * @return string|false
+	 */
+	public function loadClassmap(string $class)
+	{
+		$file = $this->classmap[$class] ?? '';
+
+		if ($file !== '')
```
samsonasik, line 205: While it should never happen, I think `is_string($file)` check need to be added to avoid falsy `$file` for saver

system/Autoloader/Autoloader.php
```diff
@@ -210,7 +191,24 @@ public function removeNamespace(string $namespace)
...
+	 *
+	 * @param string $class
+	 *
+	 * @return string|false
+	 */
+	public function loadClassmap(string $class)
+	{
+		$file = $this->classmap[$class] ?? '';
+
+		if (is_string($file) && $file !== '')
```
MGatner, line 205: This seems like a whole lot of effort not to use `empty()`.

paulbalandan: I'm using `phpstan/phpstan-strict-rules` in my workflows a lot, and one of the strict rules I'm following is the disallowance of `empty()` which implies loose comparison.

MGatner: Understood. I'm not against this verbose, strict approach but I personally prefer `empty()` when it saves multiple conditions.

MGatner (approved):

PR #4606, opened 2021-04-23
Fix validation of array data

### LINKED ISSUE #4510: Array Validation Fails
hi there!

validation with more than 1 level only validates the first match. 

Example: 
```php
// rule
'bar.*.foo' => [
    'rules' => 'required'
]

// data
"bar" => [
  [
    'foo' => 'baz'
  ],
  [
    'foo' => ''
  ]
]
```

the above validation should fails, however it passes.  https://codeigniter4.github.io/userguide/libraries/validation.html#validating-keys-that-are-arrays

paulbalandan:
**Description**
Fixes #4510 

This changes the behavior of `_array_search_dot`. Currently, the function returns early when it matched a non-null element in its value traversal, leaving the remaining elements in the loop as untouched. This PR accumulates the finds in an array, filters out the nulls. If resulting array contains only one element, then we'll return that element instead of an array in order to preserve BC. Otherwise, an array containing matches will be returned.

@luispastendev, can you check this? This passes locally on your use case .

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -39,8 +39,10 @@ if (! function_exists('dot_array_search'))
 if (! function_exists('_array_search_dot'))
 {
 	/**
-	 * Used by dot_array_search to recursively search the
+	 * Used by `dot_array_search` to recursively search the
 	 * array with wildcards.
+	 *
+	 * @internal This should not be used on its own.
 	 *
 	 * @param array $indexes
 	 * @param array $array
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -22,7 +22,7 @@ if (! function_exists('dot_array_search'))
 	 * @param string $index
 	 * @param array  $array
 	 *
-	 * @return mixed|null
+	 * @return mixed
 	 */
 	function dot_array_search(string $index, array $array)
 	{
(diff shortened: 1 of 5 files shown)


luispastendev:
Thank you very much Paul! provides the changes and now it works fine, only one detail @WinterSilence detect another use case when an empty array is sent, it evaluates it as true, what do you think about it? 

https://github.com/codeigniter4/CodeIgniter4/issues/4510#issuecomment-813532899

paulbalandan:
> Thank you very much Paul! I tested the changes and now it works fine, only one detail @WinterSilence detect another use case when an empty array is sent, it evaluates it as true, what do you think about it?
> 
> [#4510 (comment)](https://github.com/codeigniter4/CodeIgniter4/issues/4510#issuecomment-813532899)

My latest push contains some fix for empty arrays. Can you check also?

luispastendev:
Perfect I was solved! thank you very much for your time paul!

system/Helpers/array_helper.php
```diff
@@ -45,14 +45,12 @@ function dot_array_search(string $index, array $array)
 	 * @param array $indexes
 	 * @param array $array
 	 *
-	 * @return mixed|null
+	 * @return mixed
 	 */
 	function _array_search_dot(array $indexes, array $array)
```
MGatner, line 50: If this has always been a "private" function can we go ahead and mark it `@internal`?

MGatner (changes requested):
See my one comment, everything else looks great.

system/Helpers/array_helper.php
paulbalandan: Now marked `@internal`

MGatner (approved):

PR #4609, opened 2021-04-24
failValidationError can take an array of errors

caswell-wc:
**Description**
Adjusted the ResponseTrait::failValidationError so it can take an array of validation errors

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


MGatner (changes requested):
Unfortunately we cannot change the function signature like this because it will break any developer classes extending this. I do see how this functionality could be appealing though - how about adding a new method `failValidationErrors` with this functionality? We could even deprecate the old one so it could be removed down the road. See what the rest of the team thinks too...

paulbalandan (commented):
Aside from the point of @MGatner , the method's name itself implies using a single error so allowing an array of errors here seems unnatural. `failValidationErrors` is not a bad idea though. Maybe we can go that route?

caswell-wc:
@MGatner @paulbalandan Yeah, I was debating between the same approaches. I went the way I did because the body of both functions are exactly the same but I agree with both of your points so I pushed an update.

paulbalandan (approved):

MGatner:
I would like both of your opinions on whether we should deprecate the `string` version.

paulbalandan:
> I would like both of your opinions on whether we should deprecate the `string` version.

I don't have a strong opinion against deprecating it. Seems to me that having one authoritative version will be one less of a maintenance issue, so I'm good with deprecating the old one if you think so.

caswell-wc:
I could be swayed either way. I agree with @paulbalandan about having one less function means less to maintain. If we are going to do that, I would think we would want to change the signature for failValidationErrors() so it can accept an array or a string.

paulbalandan:
> I could be swayed either way. I agree with @paulbalandan about having one less function means less to maintain. If we are going to do that, I would think we would want to change the signature for failValidationErrors() so it can accept an array or a string.

Hmm okay. I think that's our way to go. Please do what's necessary.

MGatner:
@caswell-wc please also add this to **user_guide_src/source/changelogs/v4.1.2.rst**

paulbalandan (approved):

MGatner (approved):

MGatner:
Thanks @caswell-wc !

Issue #4617, opened 2021-04-26
HUGE BUG: update() function updates all records if id is empty


thewhistler1:
Example code:

```
$params = array(
     'guid' => 12345
);
$employerModel->update($employer_id,$params);
```

The variable $employer_id was null for whatever reason and so the resulting query was this: 

`update employers set guid = '12345', updated_at = '2021-04-26 06:34:13'; `

This resulted in the function updating **ALL records** instead of no records.  

**CodeIgniter version**: 4.1.1

**Expected behavior -** 
I would expect the same behavior that results when you write the query manually:
`update employers set guid = '12345', updated_at = '2021-04-26 06:34:13' WHERE employer_id = ''; `

The above query would not update any records unless you indeed had a employer with a null id. So I think the appropriate outcome would be to not update any records.

**Context**
 - OS: Ubuntu 20.04
 - Web server Apache/2.4.41 (Ubuntu)
 - PHP version PHP 7.4.3

musmanikram:
Indeed it's a HUGE bug and by mistake, if someone passes a variable with a null value can mess u the whole table.
I think it should throw an exception. Will create a PR and let see what others think about this.

iRedds:
This is not a bug. You want the framework to read your mind.

The update($id, $data) method of the model is equivalent to code
```php
->whereIn('primary_key', [$id])->set($data)->update();
```
The code was made universal. Therefore, when $id = null, the condition is not added so that all records can be updated.

Your expected behavior will break working code for other developers.

You can rewrite the update method to achieve the behavior you want.
Or add conditions to the request like 
```php
->whereIn('employer_id', [$employer_id])->update(null, $params);
```
Or add a condition to the code 
```php
if ($employer_id !== null)
{
    ->update($employer_id, $params);
}
```

musmanikram:
@thewhistler1 I think @paulbalandan have a pretty much summarised why it's not a  bug.
Please read [this](https://github.com/codeigniter4/CodeIgniter4/pull/4618#issuecomment-831353492) .

@paulbalandan Thanks again mate

lonnieezell:
I would also point out that is what regression tests are for - keeping us from shooting ourselves in the foot with dangerous code :) 

This is not a code as pointed out, and is functioning as intended. Closing.

PR #4618, opened 2021-04-27
Update will throw exception for null id

musmanikram:
#4617 
**Description**
Throwing an exception when null is provided instead of id to update. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2463,6 +2463,17 @@ class BaseBuilder
 			$this->where($where);
 		}
 
+		if (empty($this->QBWhere))
+		{
+			if (CI_DEBUG)
+			{
+				throw new DatabaseException('Updates are not allowed unless they contain a "where" or "like" clause.');
+			}
+			// @codeCoverageIgnoreStart
+			return false;
+			// @codeCoverageIgnoreEnd
+		}
+
 		if (! empty($limit))
 		{
 			if (! $this->canLimitWhereUpdates)
--- a/tests/system/Database/Live/UpdateTest.php
+++ b/tests/system/Database/Live/UpdateTest.php
@@ -17,15 +17,11 @@ class UpdateTest extends CIUnitTestCase
 
 	public function testUpdateSetsAllWithoutWhere()
 	{
+		$this->expectException(DatabaseException::class);
+		$this->expectExceptionMessage('Updates are not allowed unless they contain a "where" or "like" clause.');
+
 		$this->db->table('user')
 					->update(['name' => 'Bobby']);
-
-		$result = $this->db->table('user')->get()->getResult();
-
-		$this->assertEquals('Bobby', $result[0]->name);
-		$this->assertEquals('Bobby', $result[1]->name);
-		$this->assertEquals('Bobby', $result[2]->name);
-		$this->assertEquals('Bobby', $result[3]->name);
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


lonnieezell:
If this is put in place, how would you be able to perform a global update across all rows? It's rare, but is something that's needed occasionally.

musmanikram:
> If this is put in place, how would you be able to perform a global update across all rows? It's rare, but is something that's needed occasionally.

I agree but if someone use a variable with a null value by mistake or forget to add id can mess up the whole table. I think we should be more defensive in this way.
If someone wants to update all rows they can pass put a condition like `id > 0`

sfadschm:
I could imagibe some type of placeholder for triggering updated on all rows, e.g., "*". However, whatever placeholder would be chosen could as well be a valid database value for the 'where' clause.
I do not really see any option here without BC break right now.

MGatner:
How about a configuration option at the driver level? So if a developer knows she doesn't want sweeping updates and wants to activate that safeguard.

iRedds:
> How about a configuration option at the driver level? So if a developer knows she doesn't want sweeping updates and wants to activate that safeguard.
This is not a solution, it is a crutch. 

When done too convenient or versatile, it leads to stepping on a rake in some cases.

The behavior of the method when passed as a null ID is understandable. If the programmer does not check the value and then complains that it does not work as he expects, then this is not a problem of the framework, but of the programmer.

paulbalandan:
I agree with @iRedds here. Retrofitting the framework to what is expected by the programmer when it is otherwise is a bad idea. The method has been functioning that way since the first pre-release I believe and changing the behavior to what is "expected" appears to me as a complicit act to cover up for a programmer's lack of prudence in checking his own code.

musmanikram:
What about delete? If we pass null or empty string it throws an exception `Deletes are not allowed unless they contain a "where" or "like" clause.`. 
Then it should delete all the rows to make it consistent?

paulbalandan:
1. We have tests in place that suggests global updates are allowed when passing a `null` id. This means this behavior is quite intended.

2. The method signature of `Model::update` is https://github.com/codeigniter4/CodeIgniter4/blob/8b2e7c29043977fac378c37690cc951a715c2bd5/system/Model.php#L686. As you can see, the first parameter `id` is nullable. The second parameter is the data itself to be updated.

If `$params` is the data to be updated, passing `$model->update($params)` is different from `$model->update(null, $params)`. Why? In `$model->update($params)` you are not passing `$params` as your data BUT as your ID instead.

You cannot reach the second parameter without supplying something in the first parameter (well unless you are using PHP8 you can use named arguments like `$model->update(data: $params)`). If you really want to supply only the second parameter, then you have no choice but to write `$model->update(null, $params)`. This is how PHP7 and below works.

If you'll dive into the code, it is clear that it allows global updates by checking for non-empty ID. If a non-empty ID is supplied, a WHERE clause is added to the query. So passing an empty ID intentionally or not, the framework will assume you want a global update because it is designed to do such.

If you do not want the framework's default behavior, then you should yourself add the necessary safeguards in your code.
```php
if (! empty($id))
{
    $model->update($id, $params);
}

```

musmanikram:
> 1. We have tests in place that suggests global updates are allowed when passing a `null` id. This means this behavior is quite intended.
> 2. The method signature of `Model::update` is https://github.com/codeigniter4/CodeIgniter4/blob/8b2e7c29043977fac378c37690cc951a715c2bd5/system/Model.php#L686
>    . As you can see, the first parameter `id` is nullable. The second parameter is the data itself to be updated.
> 
> If `$params` is the data to be updated, passing `$model->update($params)` is different from `$model->update(null, $params)`. Why? In `$model->update($params)` you are not passing `$params` as your data BUT as your ID instead.
> 
> You cannot reach the second parameter without supplying something in the first parameter (well unless you are using PHP8 you can use named arguments like `$model->update(data: $params)`). If you really want to supply only the second parameter, then you have no choice but to write `$model->update(null, $params)`. This is how PHP7 and below works.
> 
> If you'll dive into the code, it is clear that it allows global updates by checking for non-empty ID. If a non-empty ID is supplied, a WHERE clause is added to the query. So passing an empty ID intentionally or not, the framework will assume you want a global update because it is designed to do such.
> 
> If you do not want the framework's default behavior, then you should yourself add the necessary safeguards in your code.
> 
> ```
> if (! empty($id))
> {
>     $model->update($id, $params);
> }
> ```

@paulbalandan this points totally makes sense. Thank you so much for such detailed answer.

I will close the PR :)

PR #4622, opened 2021-04-29
Fix: Nested sections rendering

iRedds:
**Description**
Based on http://forum.codeigniter.com/thread-79132.html

This fix allows the use of nested sections in templates. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/View.php
+++ b/system/View/View.php
@@ -109,12 +109,22 @@ class View implements RendererInterface
 	 */
 	protected $layout;
 
+
 	/**
 	 * Holds the sections and their data.
 	 *
 	 * @var array
 	 */
 	protected $sections = [];
+
+	/**
+	 * The name of the current section being rendered,
+	 * if any.
+	 *
+	 * @var string|null
+	 * @deprecated
+	 */
+	protected $currentSection;
 
 	/**
 	 * The name of the current section being rendered,
@@ -405,9 +415,12 @@ class View implements RendererInterface
 	 * @param string $name Section name
 	 *
 	 * @return void
+	 *
 	 */
-	public function section(string $name): void
+	public function section(string $name)
 	{
+		//Saved to prevent BC.
+		$this->currentSection = $name;
 		$this->sectionStack[] = $name;
 
 		ob_start();
@@ -419,7 +432,7 @@ class View implements RendererInterface
 	 * @return void
 	 * @throws RuntimeException
 	 */
-	public function endSection(): void
+	public function endSection()
 	{
 		$contents = ob_get_clean();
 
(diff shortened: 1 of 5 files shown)


system/View/View.php
```diff
@@ -120,9 +120,9 @@ class View implements RendererInterface
 	 * The name of the current section being rendered,
 	 * if any.
 	 *
-	 * @var string|null
+	 * @var array<string>
 	 */
-	protected $currentSection;
+	protected $sectionStack = [];
```
paulbalandan, line 125: We cannot just replace this as this may be used in others' code. Maybe a deprecation would be okay.

system/View/View.php
```diff
@@ -402,35 +402,41 @@ public function extend(string $layout)
 	/**
 	 * Starts holds content for a section within the layout.
 	 *
-	 * @param string $name
+	 * @param string $name Section name
+	 *
+	 * @return void
 	 */
-	public function section(string $name)
+	public function section(string $name): void
```
paulbalandan, line 409: Modifying this would break extensions.
```suggestion
	public function section(string $name)
```

system/View/View.php
```diff
@@ -402,35 +402,41 @@ public function extend(string $layout)
...
 	{
-		$this->currentSection = $name;
+		$this->sectionStack[] = $name;
```
paulbalandan, line 411: I know it may sound absurd but maybe someone is using the value of `$this->currentSection` after this method is called. I'm thinking we leave it as-is for preserving BC for now then we remove this line in the future with a `@todo` comment.

system/View/View.php
```diff
@@ -402,35 +402,41 @@ public function extend(string $layout)
...
 	}
 
 	/**
+	 * Captures the last section
+	 *
+	 * @return void
 	 * @throws RuntimeException
 	 */
-	public function endSection()
+	public function endSection(): void
```
paulbalandan, line 422:
```suggestion
	public function endSection()
```

paulbalandan (commented):
I like this. It opens new doors for view layouts. Just some small comments and a mention in the user guide and changelog and then we're set.

lonnieezell:
@iRedds I like these. Any chance you'll be able to tackle these changes?

iRedds:
Sorry. Missed notifications from github

system/View/View.php
iRedds: Hope I implemented your suggestion correctly.

paulbalandan (approved):

PR #4623, opened 2021-04-29
[Test] Use @codeCoverageIgnore for deprecated class/method

samsonasik:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,7 +9,8 @@
 		 colors="true"
 		 failOnRisky="true"
 		 failOnWarning="true"
-		 verbose="true">
+		 verbose="true"
+		 ignoreDeprecatedCodeUnits="true">
 
 	<coverage processUncoveredFiles="true">
 		<include>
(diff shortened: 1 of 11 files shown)


paulbalandan (approved):
I think we can also use the ignoreDeprecatedCodeUnits key of the coverage key in phpunit.xml.dist in addition to these annotations.

samsonasik:
I've added ignoreDeprecatedCodeUnits="true" to phpunit config https://github.com/codeigniter4/CodeIgniter4/pull/4623/commits/0eeb9f6e0611315199c4c2df2fd82aee9723c863

phpunit.xml.dist
```diff
@@ -9,7 +9,8 @@
 		 colors="true"
 		 failOnRisky="true"
 		 failOnWarning="true"
-		 verbose="true">
+		 verbose="true"
+		 ignoreDeprecatedCodeUnits="true">
 
 	<coverage processUncoveredFiles="true">
```
paulbalandan, line 15:
```suggestion
		 verbose="true">

	<coverage processUncoveredFiles="true" ignoreDeprecatedCodeUnits="true">
```
This should be on the `coverage` element per phpunit's XSD

samsonasik: updated

paulbalandan (approved):

samsonasik:

PR #4625, opened 2021-04-29
Add fallback for Config\Cookie

### LINKED ISSUE #4619: Bug: Missing Cookie Config
After the latest Cookie PR I am getting the following trying to run tests using `dev-develop`:
```
1) LocalizeFilterTest::testRemapsSelectedRoutes with data set #2 ('docs')
ErrorException: Trying to get property 'path' of non-object

/var/www/site/vendor/codeigniter4/codeigniter4/system/Session/Session.php:197
/var/www/site/vendor/codeigniter4/codeigniter4/system/Test/CIUnitTestCase.php:342
/var/www/site/vendor/codeigniter4/codeigniter4/system/Test/CIUnitTestCase.php:240
```

No time to dig now but this needs attention ASAP, if you are available @mostafakhudair @paulbalandan 
Context: A `4.1.1` project freshly switched to `dev-develop` and run `composer update`

paulbalandan:
**Description**
Fixes #4619 

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Cookie/Cookie.php
+++ b/system/Cookie/Cookie.php
@@ -13,7 +13,6 @@ namespace CodeIgniter\Cookie;
 
 use ArrayAccess;
 use CodeIgniter\Cookie\Exceptions\CookieException;
-use Config\App;
 use Config\Cookie as CookieConfig;
 use DateTimeInterface;
 use InvalidArgumentException;
@@ -128,7 +127,6 @@ class Cookie implements ArrayAccess, CloneableCookieInterface
 	public static function setDefaults($config = [])
 	{
 		$oldDefaults = self::$defaults;
-
 		$newDefaults = [];
 
 		if ($config instanceof CookieConfig)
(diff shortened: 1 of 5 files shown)


MGatner (approved):

system/Cookie/Cookie.php
```diff
@@ -289,9 +286,7 @@ public function getPrefixedName(): string
 		else
 		{
 			$search  = str_split(self::$reservedCharsList);
-			$replace = array_map(static function (string $char): string {
-				return rawurlencode($char);
-			}, $search);
+			$replace = array_map('rawurlencode', $search);
 
 			$name .= str_replace($search, $replace, $this->getName());
```
mostafakhudair, line 291:
```suggestion
			$name .= str_replace($search, array_map('rawurlencode', $search), $this->getName());
```

system/Cookie/Cookie.php
```diff
@@ -289,9 +286,7 @@ public function getPrefixedName(): string
...
 		}
```
mostafakhudair, line 292: Short if condition? 
```suggestion
		$name   = $this->getPrefix();
		$search = str_split(self::$reservedCharsList);
		$name  .= $this->isRaw()
			? $this->getName()
			: str_replace($search, array_map('rawurlencode', $search), $this->getName());

		return $name;
```

paulbalandan: I tried using the short if but it does not look too much understandable.
```php
return $this->getPrefix() . $this->isRaw()
    ? $this->getName()
    : str_replace(
        $search = str_split(self::$reservedCharsList),
        array_map('rawurlencode', $search),
        $this->getName()
    );
```
I would still prefer understandable/readable code rather than concise one.

sfadschm:
Dumb question which might be related:
If someone sees this in `Config\App` :
https://github.com/codeigniter4/CodeIgniter4/blob/aff1f3af17919e06e5cec264a4c83043ea619ab4/app/Config/App.php#L245
and decides to use `Config\Cookie` instead as suggested then goes on and deletes the property from `App\Config`, he will get an error from `HTTP\Response`:
https://github.com/codeigniter4/CodeIgniter4/blob/aff1f3af17919e06e5cec264a4c83043ea619ab4/system/HTTP/Response.php#L161-L164
Should these assignements not try to read the (deprecated) property from `Config\App` and if it is not there, get it from `Config\Cookie`?

paulbalandan:
IMO, it is not the framework's problem. Those properties are only deprecated, so technically they are still supported until the next major. If anyone deletes them intentionally then they are inflicting upon themselves their own breaking changes. The cost of providing workarounds for _possible_ removal of deprecated members in userland then removing those workarounds in next major version is not worth the benefit.

paulbalandan:
Those assignments in Response are practically useless as those aren't used anymore. Those are there for the sake of BC preservation for users still using them.

sfadschm:
Alright, I guess I misunderstood the principle of deprecations. However, still feels weird that this seems to be handled differently across the framework.

PR #4628, opened 2021-04-30
fix(cache): add check for redis empty results in deleteMatching

yassinedoghri:
**Description**

The deleteMatching implementation for the RedisHandler is not deleting all matching cache items as intended.
This PR is meant to fix that bug and fix the tests that were supposed to fail.

Also, the deleteMatching method now returns the number of deleted items as it is offers better feedback than true/false (mainly for testing purposes).

Thanks to @benjaminbellamy for the heads up!

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Cache/Handlers/RedisHandler.php
+++ b/system/Cache/Handlers/RedisHandler.php
@@ -228,11 +228,11 @@ class RedisHandler extends BaseHandler
 	 *
 	 * @param string $pattern Cache items glob-style pattern
 	 *
-	 * @return integer $deleted The number of deleted items
+	 * @return integer The number of deleted items
 	 */
 	public function deleteMatching(string $pattern)
 	{
-		$deleted  = 0;
+		$matchedKeys = [];
 		$iterator = null;
 
 		do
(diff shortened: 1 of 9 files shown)


system/Cache/Handlers/RedisHandler.php
```diff
@@ -228,25 +228,33 @@ public function delete(string $key)
...
+			$keys = $this->redis->scan($iterator, $pattern);
+
+			// Redis may return empty results, so protect against that
+			if ($keys !== false)
 			{
-				if ($this->redis->del($key) !== 1)
+				foreach ($keys as $key)
 				{
-					$success = false;
+					if ($this->redis->del($key) === 1)
```
paulbalandan, line 248: `del` seems to accept an array of keys as well as its arguments. Why did you not use that?

yassinedoghri: Because I didn't see that 😅 Thanks for the heads up!
I also optimized the FileHandler implementation by adding a `GLOB_NOSORT` flag to the `glob()` function: no need to sort the files before deleting them... It should be faster now!

paulbalandan (approved):

paulbalandan:
Thank you, @yassinedoghri

PR #4629, opened 2021-04-30
Cache getMetadata() Format

MGatner:
**Description**
`CacheInterface::getMetadata()` thus far returns a generic array of info "dependent on each adapter", as well as `false` or `null` for missing items (also dependent on each adapter). This PR attempts to standardize the expected return values while still leaving the `false` in place for now. Additionally:
* Fixes a bug where "expire" was not always an absolute time value
* Fixes a discrepancy between relative and absolute TTL values
* Adds support for `null` values to "expire" when a TTL was specified as "never expire"

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/changelogs/v4.1.2.rst
+++ b/user_guide_src/source/changelogs/v4.1.2.rst
@@ -26,6 +26,7 @@ Changes:
 - ``Entity::mutateDate`` uses external cast handler ``DatetimeCast::get``.
 - In order for ``Config\**`` classes to get their respective properties' values from the ``.env``, it is now necessary to namespace the property with the name of the class. Previously, the property names are enough but now disallowed because it can get system environment variables, like ``PATH``.
 - The array helper ``_array_search_dot`` is now marked for ``@internal`` use. As this is used by ``dot_array_search``, users should not use ``_array_search_dot`` directly in their code.
+- ``CacheInterface::getMetaData()`` returns have been specified to be null or and array with at least the "expires" key (absolute epoch expiration); on misses the File, Memcached, and Wincache Handlers still return ``false`` which will become ``null` in a future release.
 
 Deprecations:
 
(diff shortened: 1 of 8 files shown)


user_guide_src/source/changelogs/v4.1.2.rst
```diff
@@ -26,6 +26,7 @@ Changes:
 - ``Entity::mutateDate`` uses external cast handler ``DatetimeCast::get``.
 - In order for ``Config\**`` classes to get their respective properties' values from the ``.env``, it is now necessary to namespace the property with the name of the class. Previously, the property names are enough but now disallowed because it can get system environment variables, like ``PATH``.
 - The array helper ``_array_search_dot`` is now marked for ``@internal`` use. As this is used by ``dot_array_search``, users should not use ``_array_search_dot`` directly in their code.
+- ``CacheInterface::getMetaData()`` returns have been specified to be null or and array with at least the "expires" key (absolute epoch expiration); on misses the File, Memcached, and Wincache Handlers still return ``false`` which will become ``null` in a future release.
```
paulbalandan, line 29: UG check fails here

MGatner: I don't see it on mobile. "User Guide syntax check: successful in 1m 14s". Do you know what the problem is?

MGatner: Ohhh I see I missed a backtick

MGatner, line 29:
```suggestion
- ``CacheInterface::getMetaData()`` returns ``null`` for misses, or an array with at least the "expires" key (absolute epoch expiration). The File, Memcached, and Wincache Handlers still return ``false`` which will become ``null`` in a future release.
```

paulbalandan (approved):

PR #4631, opened 2021-05-01
Patch Log code

MGatner:
**Description**
The `Psr\Log` code in **ThirdParty** is a few versions old. This PR updates to the latest code from the PSR repo.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/ThirdParty/PSR/Log/AbstractLogger.php
+++ b/system/ThirdParty/PSR/Log/AbstractLogger.php
@@ -15,10 +15,11 @@ abstract class AbstractLogger implements LoggerInterface
      * System is unusable.
      *
      * @param string $message
-     * @param array $context
-     * @return null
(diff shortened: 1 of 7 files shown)


iRedds:
I have a question.
Why not move this in the composer dependency?

MGatner:
It is already in the Composer dependency, and that version will be used by Composer installations. For download users though they will access this pre-injected version, which requires occasional updates.

paulbalandan (approved):

paulbalandan:
I believe though this should be handled by `ComposerScripts`.

MGatner:
I think those are only for Kint and Escaper. My guess is that Log has barely changed in many years so didn't seem to need updates. These changes are really just docblock tweaks.

PR #4633, opened 2021-05-01
Expand Time for interface

MGatner:
**Description**
Many libraries use `DateTimeInterface` to allow `DateTime` or `DateTimeImmutable`. In a few cases in `Time` our functions are already compatible with either yet the docblocks limit it to `DateTime`. This PR expands the use for the interface, with one deprecation for an explicit method type.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -1253,19 +1253,18 @@ class Time extends DateTime
 	{
 		if ($time instanceof Time)
 		{
-			$time = $time->toDateTime()
-					->setTimezone(new DateTimeZone('UTC'));
-		}
-		elseif ($time instanceof DateTime || $time instanceof DateTimeImmutable)
-		{
-			$time = $time->setTimezone(new DateTimeZone('UTC'));
+			$time = $time->toDateTime();
 		}
 		elseif (is_string($time))
 		{
 			$timezone = $timezone ?: $this->timezone;
 			$timezone = $timezone instanceof DateTimeZone ? $timezone : new DateTimeZone($timezone);
 			$time     = new DateTime($time, $timezone);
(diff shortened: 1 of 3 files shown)


iRedds:
I suggest changes within this PR. 
Replace this code:
https://github.com/codeigniter4/CodeIgniter4/blob/9896c24315509e5a31694d5b660f4ab2ce562473/system/I18n/Time.php#L1233-L1248
On cleaned of duplication.
```php
 if ($time instanceof Time) 
 { 
 	$time = $time->toDateTime(); 
 } 
 elseif (is_string($time)) 
 { 
 	$timezone = $timezone ?: $this->timezone; 
 	$timezone = $timezone instanceof DateTimeZone ? $timezone : new DateTimeZone($timezone); 
 	$time     = new DateTime($time, $timezone); 
  } 

 if ($time instanceof DateTimeInterface) 
 { 
 	$time = $time->setTimezone(new DateTimeZone('UTC')); 
 } 

```

MGatner:
I did it like this because `DateTimeInterface` doesn't actually have a `setTimeZone()` method. It might be okay to assume it since the interface is internal:

> It is not possible to implement this interface with userland classes.

I'm not sure if PHPStan would complain though.

iRedds:
This is just a suggestion and I'm not really against using your condition.

```php 
$time instanceof DateTime || $time instanceof DateTimeImmutable
```
The main thing is to get rid of duplication.

MGatner:
Got it! If you can add it as an inline suggestion I can accept it on mobile, otherwise it will need to wait until I'm back on desktop.

MGatner:
@iRedds I think I got that taken care of.

paulbalandan (approved):

PR #4634, opened 2021-05-01
Refactor ComposerScripts

paulbalandan:
**Description**
Refactor `CodeIgniter\ComposerScripts` to make it a little performant while allowing new direct dependencies to be added with ease.

This class is technically a "development" class so this can be changed at will without worrying for BC breaks.

**Checklist:**
- [x] Securely signed commits
- [x] Tested locally. Git status is clean.

### DIFF
--- a/system/ComposerScripts.php
+++ b/system/ComposerScripts.php
@@ -11,210 +11,170 @@
 
 namespace CodeIgniter;
 
-use ReflectionClass;
-use ReflectionException;
+use FilesystemIterator;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use SplFileInfo;
 
 /**
- * ComposerScripts
- *
- * These scripts are used by Composer during installs and updates
+ * This class is used by Composer during installs and updates
  * to move files to locations within the system folder so that end-users
  * do not need to use Composer to install a package, but can simply
- * download
+ * download.
  *
  * @codeCoverageIgnore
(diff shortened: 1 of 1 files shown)


MGatner:
This is definitely an improvement. I have been wanting to add a `Publisher` component to the framework for some time. It would mostly handle moving files from `vendor` into various parts of a project. This could be to facilitate patching the **app/** folder (like `Patches` and `Liaison Revision`) or to push necessary assets into **public/**; I could also see it being useful here.

If this is something you'd have interest in taking on then it could supersede these changes, but if that's beyond your current scope we can proceed with this.

paulbalandan:
> This is definitely an improvement. I have been wanting to add a `Publisher` component to the framework for some time. It would mostly handle moving files from `vendor` into various parts of a project. This could be to facilitate patching the **app/** folder (like `Patches` and `Liaison Revision`) or to push necessary assets into **public/**; I could also see it being useful here.
> 
> If this is something you'd have interest in taking on then it could supersede these changes, but if that's beyond your current scope we can proceed with this.

It appears to be a good addition but I cannot commit myself to be working on this in the coming days/weeks. Maybe we can add this to the "Nice to Have" and look back into it when time permits.

MGatner (approved):
No worries! I'm planning to work on it when I get a chance, I need it for a few projects and modules. Thanks for this PR, some really good improvements in here.

PR #4635, opened 2021-05-01
[Cache] Allow covariant returns and optimize code

paulbalandan:
**Description**
- Allow covariant return types in docblocks of Cache handlers
- Simplify other code

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide


MGatner (approved):
Great cleanup. It has come to my attention that I missed `MockCache` in my recent round of changes. However that class needs a bit more work to be an appropriate Mock, so if you don't get to it for this PR I will mimic your changes there when I do it.

PR #4637, opened 2021-05-02
Cache Key Validation

MGatner:
**Description**
Cache keys are currently a free-for-all, though the specification is much tighter than that. Additionally, some handlers have limits on the length of key that can be used. So far the only one I know for sure is `FileHandler`, because using too long of a key fails to create the file since it is invalid in the filesystem.

This PR adds cache key validation and hashing to address the above issues. Note that I am *not* a cache expert, so if someone with more knowledge or experience wants to chime in I would be grateful. For example, the restricted characters were based on PSR-6 but some handlers may actually allow broader key usage than that.

**Checklist:**
- [X] Securely signed commits
- [x] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Cache/Handlers/BaseHandler.php
+++ b/system/Cache/Handlers/BaseHandler.php
@@ -14,12 +14,65 @@ namespace CodeIgniter\Cache\Handlers;
 use Closure;
 use CodeIgniter\Cache\CacheInterface;
 use Exception;
+use InvalidArgumentException;
 
 /**
  * Base class for cache handling
  */
 abstract class BaseHandler implements CacheInterface
 {
+	/**
+	 * Reserved characters that cannot be used in a key or tag.
+	 *
+	 * @see https://github.com/symfony/cache-contracts/blob/c0446463729b89dd4fa62e9aeecc80287323615d/ItemInterface.php#L43
+	 */
+	public const RESERVED_CHARACTERS = '{}()/\@:';
+
+	/**
+	 * Maximum key length.
+	 */
+	public const MAX_KEY_LENGTH = PHP_INT_MAX;
+
+	/**
+	 * Prefix to apply to cache keys.
+	 * May not be used by all handlers.
+	 *
+	 * @var string
+	 */
+	protected $prefix;
+
+	/**
+	 * Validates a cache key according to PSR-6.
+	 * Keys that exceed MAX_KEY_LENGTH are hashed.
+	 *
+	 * @param string $key    The key to validate
+	 * @param string $prefix Optional prefix to include in length calculations
+	 *
+	 * @throws InvalidArgumentException When $key is not valid
+	 *
+	 * @see https://github.com/symfony/cache/blob/7b024c6726af21fd4984ac8d1eae2b9f3d90de88/CacheItem.php#L158
+	 */
+	public static function validateKey($key, $prefix = ''): string
+	{
+		if (! is_string($key))
+		{
+			throw new InvalidArgumentException('Cache key must be a string');
+		}
+		if ($key === '')
+		{
+			throw new InvalidArgumentException('Cache key cannot be empty.');
+		}
+		if (strpbrk($key, self::RESERVED_CHARACTERS) !== false)
+		{
+			throw new InvalidArgumentException('Cache key contains reserved characters ' . self::RESERVED_CHARACTERS);
+		}
+
+		// If the key with prefix exceeds the length then return the hashed version
+		return strlen($prefix . $key) > static::MAX_KEY_LENGTH ? $prefix . md5($key) : $prefix . $key;
+	}
+
+	//--------------------------------------------------------------------
+
 	/**
 	 * Get an item from the cache, or execute the given Closure and store the result.
 	 *
--- a/system/Cache/Handlers/BaseHandler.php
+++ b/system/Cache/Handlers/BaseHandler.php
@@ -23,8 +23,7 @@ abstract class BaseHandler implements CacheInterface
 {
 	/**
 	 * Reserved characters that cannot be used in a key or tag.
-	 *
-	 * @see https://github.com/symfony/cache-contracts/blob/c0446463729b89dd4fa62e9aeecc80287323615d/ItemInterface.php#L43
+	 * From https://github.com/symfony/cache-contracts/blob/c0446463729b89dd4fa62e9aeecc80287323615d/ItemInterface.php#L43
 	 */
 	public const RESERVED_CHARACTERS = '{}()/\@:';
 
@@ -44,13 +43,12 @@ abstract class BaseHandler implements CacheInterface
 	/**
 	 * Validates a cache key according to PSR-6.
 	 * Keys that exceed MAX_KEY_LENGTH are hashed.
+	 * From https://github.com/symfony/cache/blob/7b024c6726af21fd4984ac8d1eae2b9f3d90de88/CacheItem.php#L158
 	 *
 	 * @param string $key    The key to validate
 	 * @param string $prefix Optional prefix to include in length calculations
 	 *
 	 * @throws InvalidArgumentException When $key is not valid
-	 *
-	 * @see https://github.com/symfony/cache/blob/7b024c6726af21fd4984ac8d1eae2b9f3d90de88/CacheItem.php#L158
 	 */
 	public static function validateKey($key, $prefix = ''): string
 	{
--- a/system/Cache/Handlers/FileHandler.php
+++ b/system/Cache/Handlers/FileHandler.php
@@ -21,11 +21,9 @@ use Throwable;
 class FileHandler extends BaseHandler
 {
 	/**
-	 * Prefixed to all cache names.
-	 *
-	 * @var string
+	 * Maximum key length.
 	 */
-	protected $prefix;
+	public const MAX_KEY_LENGTH = 255;
 
 	/**
 	 * Where to store cached files on the disk.
@@ -94,8 +92,7 @@ class FileHandler extends BaseHandler
 	 */
 	public function get(string $key)
 	{
-		$key = $this->prefix . $key;
-
+		$key  = static::validateKey($key, $this->prefix);
 		$data = $this->getItem($key);
 
 		return is_array($data) ? $data['data'] : null;
@@ -114,7 +111,7 @@ class FileHandler extends BaseHandler
 	 */
 	public function save(string $key, $value, int $ttl = 60)
 	{
-		$key = $this->prefix . $key;
+		$key = static::validateKey($key, $this->prefix);
 
 		$contents = [
 			'time' => time(),
@@ -152,7 +149,7 @@ class FileHandler extends BaseHandler
 	 */
 	public function delete(string $key)
 	{
-		$key = $this->prefix . $key;
+		$key = static::validateKey($key, $this->prefix);
 
 		return is_file($this->path . $key) && unlink($this->path . $key);
 	}
@@ -170,7 +167,7 @@ class FileHandler extends BaseHandler
 	{
 		$deleted = 0;
 
-		foreach (glob($this->path . $pattern,  GLOB_NOSORT) as $filename)
+		foreach (glob($this->path . $pattern, GLOB_NOSORT) as $filename)
 		{
 			if (is_file($filename) && @unlink($filename))
 			{
@@ -193,8 +190,7 @@ class FileHandler extends BaseHandler
 	 */
 	public function increment(string $key, int $offset = 1)
 	{
-		$key = $this->prefix . $key;
-
+		$key  = static::validateKey($key, $this->prefix);
 		$data = $this->getItem($key);
 
 		if ($data === false)
@@ -222,12 +218,11 @@ class FileHandler extends BaseHandler
 	 * @param string  $key    Cache ID
 	 * @param integer $offset Step/value to increase by
 	 *
-	 * @return bool
+	 * @return boolean
 	 */
 	public function decrement(string $key, int $offset = 1)
 	{
-		$key = $this->prefix . $key;
-
+		$key  = static::validateKey($key, $this->prefix);
 		$data = $this->getItem($key);
 
 		if ($data === false)
@@ -288,7 +283,7 @@ class FileHandler extends BaseHandler
 	 */
 	public function getMetaData(string $key)
 	{
-		$key = $this->prefix . $key;
+		$key = static::validateKey($key, $this->prefix);
 
 		if (! is_file($this->path . $key))
 		{
@@ -342,26 +337,26 @@ class FileHandler extends BaseHandler
 	 * Does the heavy lifting of actually retrieving the file and
 	 * verifying it's age.
 	 *
-	 * @param string $key
+	 * @param string $filename
 	 *
 	 * @return boolean|mixed
 	 */
-	protected function getItem(string $key)
+	protected function getItem(string $filename)
 	{
-		if (! is_file($this->path . $key))
+		if (! is_file($this->path . $filename))
 		{
 			return false;
 		}
 
-		$data = unserialize(file_get_contents($this->path . $key));
+		$data = unserialize(file_get_contents($this->path . $filename));
 
 		// @phpstan-ignore-next-line
 		if ($data['ttl'] > 0 && time() > $data['time'] + $data['ttl'])
 		{
 			// If the file is still there then remove it
-			if (is_file($this->path . $key))
+			if (is_file($this->path . $filename))
 			{
-				unlink($this->path . $key);
+				unlink($this->path . $filename);
 			}
 
 			return false;
--- a/system/Cache/Handlers/FileHandler.php
+++ b/system/Cache/Handlers/FileHandler.php
@@ -23,7 +23,7 @@ class FileHandler extends BaseHandler
 	/**
 	 * Maximum key length.
 	 */
-	public const MAX_KEY_LENGTH = 255;
+	public const MAX_KEY_LENGTH = PHP_MAXPATHLEN;
 
 	/**
 	 * Where to store cached files on the disk.
--- a/system/Cache/Handlers/WincacheHandler.php
+++ b/system/Cache/Handlers/WincacheHandler.php
@@ -22,15 +22,6 @@ use Exception;
 class WincacheHandler extends BaseHandler
 {
 	/**
-	 * Prefixed to all cache names.
-	 *
-	 * @var string
-	 */
-	protected $prefix;
-
-	//--------------------------------------------------------------------
-
-	/**
 	 * Constructor.
 	 *
 	 * @param Cache $config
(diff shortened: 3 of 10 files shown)


system/Cache/Handlers/BaseHandler.php
```diff
@@ -14,12 +14,65 @@
...
+	 * @throws InvalidArgumentException When $key is not valid
+	 *
+	 * @see https://github.com/symfony/cache/blob/7b024c6726af21fd4984ac8d1eae2b9f3d90de88/CacheItem.php#L158
+	 */
+	public static function validateKey($key, $prefix = ''): string
+	{
+		if (! is_string($key))
+		{
+			throw new InvalidArgumentException('Cache key must be a string');
+		}
```
paulbalandan, line 60: Why not typehint the `string` so we get a `TypeError` instead?
`public static function validateKey(string $key, string $prefix = ''): string`

system/Cache/Handlers/BaseHandler.php
```diff
@@ -14,12 +14,65 @@
...
+		{
+			throw new InvalidArgumentException('Cache key cannot be empty.');
+		}
+		if (strpbrk($key, self::RESERVED_CHARACTERS) !== false)
+		{
+			throw new InvalidArgumentException('Cache key contains reserved characters ' . self::RESERVED_CHARACTERS);
+		}
+
+		// If the key with prefix exceeds the length then return the hashed version
+		return strlen($prefix . $key) > static::MAX_KEY_LENGTH ? md5($key) : $key;
```
paulbalandan, line 71: Why is the `$prefix` not included in the return?

system/Cache/Handlers/FileHandler.php
```diff
@@ -94,8 +92,7 @@ public function initialize()
 	 */
 	public function get(string $key)
 	{
-		$key = $this->prefix . $key;
-
+		$key  = static::validateKey($key, $this->prefix);
```
paulbalandan, line 95: If the `$this->prefix` is really not included in the return of `validateKey`, then `$key` should be prepended with `$this->prefix` before it is used.

Applies to other instances where `static::validateKey` is used.

system/Cache/Handlers/BaseHandler.php
MGatner: It's for consistency with PSR-6 (I'm working on an adapter for the framework). If we handle it as a type here then the PSR-6 version will cause errors instead of expected exceptions. The root of the problem is that the PSR-6 interfaces are compatible with PHP 5 so don't include typehints until the newest versions which require PHP 8: https://www.php-fig.org/psr/psr-6/

system/Cache/Handlers/BaseHandler.php
MGatner: Two-fold: 1) We have no specification around prefix handling, so there is no assumption that handlers will use them even when set (even our internal handlers are inconsistent), and 2) one of the goals of this method is to keep it "generic" enough to be used by the PSR-6 adapters as well, which will probably ignore this return value (so as not to double-process keys) but definitely won't deal in prefixes at all.

system/Cache/Handlers/FileHandler.php
MGatner: You're right! My mistake. If you agree with my other comments on prefixing I will update these.

MGatner:
Thanks for the review @paulbalandan. I'd like your feedback on my responses before I implement any changes.

system/Cache/Handlers/BaseHandler.php
paulbalandan: Ok. Understood.

system/Cache/Handlers/BaseHandler.php
paulbalandan: What's your take on making this a `void` return type, i.e., when the length of the key exceeds the max length it throws an `InvalidArgumentException` too just like the previous ifs?

I am thinking why is there a need to use MD5 hashes of keys. Since an excessively long key is an "invalid" key, then should it follow this:
> * Any time an invalid argument is passed into a method it must throw an
> * exception class which implements Psr\Cache\InvalidArgumentException.

MGatner: I'm taking my lead from the PHP-Cache tests since that is what I plan on using to test the PSR-6 & 16 adapters. According to the test case (which is not FIG-official mind you) excessively long keys *should* be acceptable. It's the last dialing test on my implementation which is what instigated this PR.

MGatner:
@paulbalandan Go to my fork and select the `psr-cache` branch and that will give you an idea of what I'm working towards. This is part of a much bigger project to supply adapters for most PSR components that we don't supply natively. In this case, I am working on a project at my day job that uses a package which requires a PSR-6 implementation; currently I have to include an external cache library to make this work, even though we have a very healthy cache layer.

system/Cache/Handlers/BaseHandler.php
paulbalandan: Okay, let's proceed then. I've looked at your branch and the code looks good so I'll trust you on this.

MGatner: Awesome, thanks for checking it out! I'll fix the prefixes and run through tests again.

MGatner: Actually after working with this some more I think not returning the prefix is pretty confusing and outweighs the benefits. Since the PSR-6 adapter will ignore the return value I think I will go with your suggestion.

system/Cache/Handlers/BaseHandler.php
```diff
@@ -14,12 +14,65 @@
...
+		{
+			throw new InvalidArgumentException('Cache key cannot be empty.');
+		}
+		if (strpbrk($key, self::RESERVED_CHARACTERS) !== false)
+		{
+			throw new InvalidArgumentException('Cache key contains reserved characters ' . self::RESERVED_CHARACTERS);
+		}
+
+		// If the key with prefix exceeds the length then return the hashed version
+		return strlen($prefix . $key) > static::MAX_KEY_LENGTH ? $prefix . md5($key) : $prefix . $key;
```
paulbalandan, line 71: Since we're checking against the length of `$prefix . $key`, should the hash include the prefix also?
```suggestion
		return strlen($prefix . $key) > static::MAX_KEY_LENGTH ? md5($prefix . $key) : $prefix . $key;
```

system/Cache/Handlers/BaseHandler.php
```diff
@@ -14,12 +14,65 @@
...
+	 *
+	 * @throws InvalidArgumentException When $key is not valid
+	 *
+	 * @see https://github.com/symfony/cache/blob/7b024c6726af21fd4984ac8d1eae2b9f3d90de88/CacheItem.php#L158
+	 */
+	public static function validateKey($key, $prefix = ''): string
+	{
+		if (! is_string($key))
+		{
+			throw new InvalidArgumentException('Cache key must be a string');
```
lonnieezell, line 59: Should these errors be localized?

system/Cache/Handlers/WincacheHandler.php
```diff
@@ -81,6 +73,8 @@ public function get(string $key)
 	 */
 	public function save(string $key, $value, int $ttl = 60)
 	{
+		$key = static::validateKey($key);
+
 		return wincache_ucache_set($this->prefix . $key, $value, $ttl);
```
paulbalandan, line 78: Remove the duplicate prefix
```suggestion
		return wincache_ucache_set($key, $value, $ttl);
```

system/Cache/Handlers/WincacheHandler.php
```diff
@@ -95,6 +89,8 @@ public function save(string $key, $value, int $ttl = 60)
 	 */
 	public function delete(string $key)
 	{
+		$key = static::validateKey($key);
+
 		return wincache_ucache_delete($this->prefix . $key);
```
paulbalandan, line 94:
```suggestion
		return wincache_ucache_delete($key);
```

system/Cache/Handlers/WincacheHandler.php
```diff
@@ -124,6 +120,8 @@ public function deleteMatching(string $pattern)
 	 */
 	public function increment(string $key, int $offset = 1)
 	{
+		$key = static::validateKey($key);
+
 		return wincache_ucache_inc($this->prefix . $key, $offset);
```
paulbalandan, line 125:
```suggestion
		return wincache_ucache_inc($key, $offset);
```

system/Cache/Handlers/WincacheHandler.php
```diff
@@ -139,6 +137,8 @@ public function increment(string $key, int $offset = 1)
 	 */
 	public function decrement(string $key, int $offset = 1)
 	{
+		$key = static::validateKey($key);
+
 		return wincache_ucache_dec($this->prefix . $key, $offset);
```
paulbalandan, line 142:
```suggestion
		return wincache_ucache_dec($key, $offset);
```

system/Cache/Handlers/WincacheHandler.php
```diff
@@ -183,6 +183,8 @@ public function getCacheInfo()
 	 */
 	public function getMetaData(string $key)
 	{
+		$key = static::validateKey($key);
+
 		if ($stored = wincache_ucache_info(false, $this->prefix . $key))
```
paulbalandan, line 188:
```suggestion
		if ($stored = wincache_ucache_info(false, $key))
```

system/Cache/Handlers/BaseHandler.php
```diff
@@ -14,12 +14,65 @@
...
+		}
+		if ($key === '')
+		{
+			throw new InvalidArgumentException('Cache key cannot be empty.');
```
lonnieezell, line 63: I'm not necessarily opposed to throwing the exception directly here, but most of the rest of the framework uses external exception classes with static methods. Does it make sense to do that here for consistency or is that overkill?

tests/system/Cache/Handlers/BaseHandlerTest.php
```diff
@@ -0,0 +1,66 @@
...
+{
+	/**
+	 * @dataProvider invalidTypeProvider
+	 */
+	public function testValidateKeyInvalidType($input)
+	{
+		$this->expectException('InvalidArgumentException');
+		$this->expectExceptionMessage('Cache key must be a string');
+
+		BaseHandler::validateKey(false);
```
paulbalandan, line 17:
```suggestion
		BaseHandler::validateKey($input);
```

system/Cache/Handlers/BaseHandler.php
```diff
@@ -14,12 +14,65 @@
...
+	 * Keys that exceed MAX_KEY_LENGTH are hashed.
+	 *
+	 * @param string $key    The key to validate
+	 * @param string $prefix Optional prefix to include in length calculations
+	 *
+	 * @throws InvalidArgumentException When $key is not valid
+	 *
+	 * @see https://github.com/symfony/cache/blob/7b024c6726af21fd4984ac8d1eae2b9f3d90de88/CacheItem.php#L158
+	 */
+	public static function validateKey($key, $prefix = ''): string
```
lonnieezell, line 55: If all of the handlers are based on this class, why make this static? It's already being instantiated so there's no savings there.

system/Cache/Handlers/BaseHandler.php
MGatner: No, the prefix may be crucial to some handlers for the end result (i.e. an optimized `FileHandler` may organize folder structures using the prefix as a starting point) so we need to be sure it ends up as the beginning of the final key. It is included in the length calculation to mitigate accidental overages: e.g. `FileHandler` has a 255 limit because of the filesystem, so if you pass a 250 key with a 10 prefix it needs to be hashed.

system/Cache/Handlers/BaseHandler.php
MGatner: Your two comments are intimately related. Throughout the framework we don't localize `InvalidArgumentException` content precisely because they are not runtime exceptions (and thus should use the `new static FrameworkException` approach).

That said, I think these are up for grabs as to whether they should be considered invalid arguments or runtime issues. I went with `InvalidArgumentException` because these will be re-used with the PSR-6 adapter which requires a cache-specific `InvalidArgumentException` for precisely these failures: https://www.php-fig.org/psr/psr-6/

system/Cache/Handlers/BaseHandler.php
MGatner: *See below*

system/Cache/Handlers/WincacheHandler.php
MGatner: Whoops! Missed that - incoming.

tests/system/Cache/Handlers/BaseHandlerTest.php
MGatner: 🙄 Thanks Paul. I woodwork as well and have a rule that I never use saws when I'm tired - looks like Is would adopt the same for coding 😂

system/Cache/Handlers/BaseHandler.php
MGatner: Because this will be used by the PSR-6 adapter, which was the inspiration for this to begin with. I originally had this as a static method on the Cache Item class but I thought it would be a benefit to the framework Cache core - then I realized we had this bug with `FIleHandler` and long keys and it made sense to kill two issues with one static method.

system/Cache/Handlers/FileHandler.php
```diff
@@ -21,11 +21,9 @@
 class FileHandler extends BaseHandler
 {
 	/**
-	 * Prefixed to all cache names.
-	 *
-	 * @var string
+	 * Maximum key length.
 	 */
-	protected $prefix;
+	public const MAX_KEY_LENGTH = 255;
```
paulbalandan, line 26: I've just learned from Symfony Filesystem of the `PHP_MAXPATHLEN` constant available since PHP5.3. Why did you not use it here instead?

system/Cache/Handlers/WincacheHandler.php
```diff
@@ -81,7 +73,9 @@ public function get(string $key)
 	 */
 	public function save(string $key, $value, int $ttl = 60)
 	{
-		return wincache_ucache_set($this->prefix . $key, $value, $ttl);
+		$key = static::validateKey($this->prefix . $key);
```
paulbalandan, line 76:
```suggestion
		$key = static::validateKey($key, $this->prefix);
```

system/Cache/Handlers/WincacheHandler.php
```diff
@@ -124,7 +120,9 @@ public function deleteMatching(string $pattern)
 	 */
 	public function increment(string $key, int $offset = 1)
 	{
-		return wincache_ucache_inc($this->prefix . $key, $offset);
+		$key = static::validateKey($this->prefix . $key);
```
paulbalandan, line 123:
```suggestion
		$key = static::validateKey($key, $this->prefix);
```

system/Cache/Handlers/WincacheHandler.php
```diff
@@ -183,7 +183,9 @@ public function getCacheInfo()
 	 */
 	public function getMetaData(string $key)
 	{
-		if ($stored = wincache_ucache_info(false, $this->prefix . $key))
+		$key = static::validateKey($this->prefix . $key);
```
paulbalandan, line 186:
```suggestion
		$key = static::validateKey($key, $this->prefix);
```

system/Cache/Handlers/BaseHandler.php
lonnieezell: That's fair. I'm fine with this then.

system/Cache/Handlers/FileHandler.php
MGatner: Ooo that's great! Will change.

system/Cache/Handlers/BaseHandler.php
```diff
@@ -14,12 +14,65 @@
...
 
 /**
  * Base class for cache handling
  */
 abstract class BaseHandler implements CacheInterface
 {
+	/**
+	 * Reserved characters that cannot be used in a key or tag.
+	 *
+	 * @see https://github.com/symfony/cache-contracts/blob/c0446463729b89dd4fa62e9aeecc80287323615d/ItemInterface.php#L43
```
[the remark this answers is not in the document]
paulbalandan: Just read this line: `but MUST be able to return the original unmodified key string.` Since MD5 hashing is a one-way process, then aren't we deviating again from PSR-6?

system/Cache/Handlers/BaseHandler.php
lonnieezell: I thought it might have had to do with PSR6 so I checked that last night but didn't see where the interface required it be static? Or is it just that you need it usable by the other class? If that's the case, I'm cool.

system/Cache/Handlers/BaseHandler.php
```diff
@@ -14,12 +14,65 @@
...
+	 */
+	public const RESERVED_CHARACTERS = '{}()/\@:';
+
+	/**
+	 * Maximum key length.
+	 */
+	public const MAX_KEY_LENGTH = PHP_INT_MAX;
```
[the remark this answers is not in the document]
MGatner: It's basically meant to be "unlimited". I couldn't find limits for all the handlers, but P/Redis stores the keys as binary strings with a limit of 512 MB - huge, in other words: https://stackoverflow.com/questions/5606106/what-is-the-maximum-value-size-you-can-store-in-redis

I'm open to other suggestions, or varying the specific handlers more, but that's how I got there.

system/Cache/Handlers/BaseHandler.php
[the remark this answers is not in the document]
MGatner: That link is because I copied this code from Symfony, more of an attribution than a reference. Maybe I should remove the `@see` and just have the link?

`BaseHandler::validateKey()` is not trying to be PSR-6 compliant - we'd have bigger issues to address if that were the case. The PSR-6 adapter will use this but handle its own keys and exceptions.

system/Cache/Handlers/BaseHandler.php
MGatner: It's not that it is required to be static, but rather that the adapter will want to handle verification as well. We could have this be a class protected method and let the PSR-6 adapter have it own but it would be almost entirely duplicate code. Making this static here lets us re-use it outside of a specific handler's context.

MGatner:
Test failures are unrelated:
```

55
Err:20 https://packages.microsoft.com/ubuntu/20.04/prod focal InRelease
56
  Could not connect to packages.microsoft.com:443 (13.90.56.68), connection timed out [IP: 13.90.56.68 443]
57
Err:21 https://packages.microsoft.com/repos/azure-cli focal Release
58
  Unable to connect to packages.microsoft.com:https: [IP: 13.90.56.68 443]
59
Reading package lists...
60
E: The repository 'https://packages.microsoft.com/repos/azure-cli focal Release' does not have a Release file.
61
Error: Process completed with exit code 100.
```

Any more thoughts on this one?

paulbalandan (approved):
I'm good with this, unless Lonnie has some more to add. 😁

lonnieezell (approved):

PR #4638, opened 2021-05-02
Update MockCache

MGatner:
**Description**
`MockCache` was left out of the recent updates to the Cache handlers. This PR brings it in line with the expectations and fixes a few bugs.

Note that this PR requires the validation from #4637 and will need to rerun tests after that is resolved.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide


system/Test/Mock/MockCache.php
```diff
@@ -137,6 +147,7 @@ public function deleteMatching(string $pattern)
 			if (fnmatch($pattern, $key))
 			{
 				unset($this->cache[$key]);
+				unset($this->expirations[$key]);
```
paulbalandan, line 150: `deleteMatching` should now return `int`.

PR #4644, opened 2021-05-04
URI::removeDotSegments()

MGatner:
**Description**
This is "stage one" for some bugfixes in URI resolution. This changes the internal `removeDotSegments()` method to static and fixes a logic bug (`ltrim()` could never leave a single slash).

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -143,6 +143,116 @@ class URI
 	//--------------------------------------------------------------------
 
 	/**
+	 * Builds a representation of the string from the component parts.
+	 *
+	 * @param string $scheme
+	 * @param string $authority
+	 * @param string $path
+	 * @param string $query
+	 * @param string $fragment
+	 *
+	 * @return string
+	 */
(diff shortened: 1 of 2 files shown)


lonnieezell:
What's the purpose of making `removeDotSegments()` static? That's a BC break since it was a public method.

MGatner:
It's marked as an `@internal` method, so we should be safe to change it. I haven't actually looked to see if it was mentioned in the User Guide. I made it static so `IncomingRequest` could use it (see #4646) because the version of this same method there was a) inconsistent and b) duplicating functionality.

lonnieezell (approved):
I missed that it was marked internal. While technically still a BC break, when so marked it's a "use at your own risk" feature.

MGatner:
Agreed! I did check and we don't mention it anywhere in the User Guide, also it looks like the API Docs are smart enough to leave off `@internal` methods so it does not appear here either: https://codeigniter4.github.io/api/classes/CodeIgniter.HTTP.URI.html

PR #4645, opened 2021-05-04
fixes(useSoftDeletes)make model

marcos-queiroz:
I noticed that when using spark make: model the variable useSoftDeletes is being generated in the singular, useSoftDelete, which causes an error when setting to TRUE, when executing $model->delete (1), the record is being permanently deleted.

**Description**
A small change in the variable 'useSoftDelete' to 'useSoftDeletes' as shown in the documentation.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


iRedds:
This has already been fixed.
It's just that due to the long release cycle, simple fixes are not available.

marcos-queiroz:
Thank you, congratulations on your work.

PR #4646, opened 2021-05-04
Internal URI handling

MGatner:
**Description**
"Stage two" of URI changes, this PR backs off some of the heavy-handed coercing that `URI::__toString()` does to create "project URLs". IMO `URI` should not be in the business of dealing with `baseURL` at all, but since this is currently the case we can't remove it entirely. Also improves `IncomingRequest`'s interactions with `URI` to detect the "current" URL, and fixes a bug where scheme inconsistencies could create looping redirects.

Note: This PR relies on #4644

**Checklist:**
- [X] Securely signed commits
- [X] Component(s) with PHPdocs
- [X] Unit testing, with >80% coverage
- [X] User guide updated
- [X] Conforms to style guide

PR #4647, opened 2021-05-05
URL Class

MGatner:
**Description**
"Stage Three", this PR introduces a companion class to `URI`: `URL`. The `URL` class handles internal URIs (i.e. specific to the project) in the following manner:
* A `URL` is always constructed with a relative path
* A `URL` will always use the App config values from its construction (`baseURL`, `indexPage`, `forceGlobalSecureRequest`)
* A `URL` is immutable, not affected by changes to `$_SERVER` or services once it has been created

I would like `URL` to become the definitive source for internal URLs within the framework (e.g. replace the logic in `base_url()` and `current_url()`, be used by `RedirectResponse`, etc) but for now this PR is just introducing the concept and demonstrating the tests. Please be thorough with any reviews, since URLs have been a long pain point (particularly around subfolders).

Note: This PR relies on #4644 and #4646.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/HTTP/URL.php
+++ b/system/HTTP/URL.php
@@ -39,26 +39,34 @@ class URL
 	//--------------------------------------------------------------------
 
 	/**
-	 * Returns an instance representing the current URL.
+	 * Returns an instance representing the base URL.
+	 *
+	 * @param string $uri Additional URI string to include
 	 *
 	 * @return static
 	 */
-	final public static function base()
+	final public static function base(string $uri = '')
 	{
-		return new static('');
+		// Base URLs never include the index page
+		$config            = clone config('App');
+		$config->indexPage = '';
+
+		return new static($uri, $config);
 	}
 
 	/**
-	 * Returns an instance representing the current URL.
+	 * Returns an instance representing a routed URL.
 	 *
-	 * @param string $uri
+	 * @param string $uri Named route, reverse route, or URI string
 	 *
 	 * @return static
 	 */
 	final public static function to(string $uri)
 	{
+		$uri = rtrim($uri, '/ ');
+
 		// Check for a named or reverse-route
-		if ($route = Services::routes()->reverseRoute($uri))
+		if ($uri !== '' && $route = Services::routes()->reverseRoute($uri))
 		{
 			return new static($route);
 		}
--- a/system/HTTP/URL.php
+++ b/system/HTTP/URL.php
@@ -12,6 +12,7 @@
 namespace CodeIgniter\HTTP;
 
 use CodeIgniter\HTTP\IncomingRequest;
+use CodeIgniter\Router\Exceptions\RouterException;
 use Config\App;
 use Config\Services;
 use InvalidArgumentException;
@@ -20,7 +21,7 @@ use InvalidArgumentException;
  * URI wrapper to work with complete project URLs.
  * This class creates immutable instances.
  */
-class URL
+final class URL
 {
 	/**
 	 * Underlying URI instance
@@ -34,18 +35,29 @@ class URL
 	 *
 	 * @var string
 	 */
-	protected $relativePath;
+	private $relativePath;
 
 	//--------------------------------------------------------------------
 
 	/**
-	 * Returns an instance representing the base URL.
+	 * Creates the base URL.
+	 *
+	 * @return static
+	 */
+	public static function base()
+	{
+		return static::public('');
+	}
+
+	/**
+	 * Creates a URL to unrouted public files (typically assets).
+	 * Similar to base_url('path/to/file')
 	 *
 	 * @param string $uri Additional URI string to include
 	 *
 	 * @return static
 	 */
-	final public static function base(string $uri = '')
+	public static function public(string $uri)
 	{
 		// Base URLs never include the index page
 		$config            = clone config('App');
@@ -55,33 +67,45 @@ class URL
 	}
 
 	/**
-	 * Returns an instance representing a routed URL.
-	 *
-	 * @param string $uri Named route, reverse route, or URI string
-	 *
-	 * @return static
-	 */
-	final public static function to(string $uri)
-	{
-		$uri = rtrim($uri, '/ ');
-
-		// Check for a named or reverse-route
-		if ($uri !== '' && $route = Services::routes()->reverseRoute($uri))
-		{
-			return new static($route);
-		}
-
-		return new static($uri);
-	}
-
-	/**
 	 * Returns an instance representing the current URL.
 	 *
 	 * @return static
 	 */
-	final public static function current()
+	public static function current()
 	{
 		return static::fromRequest(Services::request());
+	}
+
+	/**
+	 * Creates a framework URL.
+	 *
+	 * @param string $uri
+	 *
+	 * @return static
+	 */
+	public static function to(string $uri)
+	{
+		return new static(rtrim($uri, '/ '));
+	}
+
+	/**
+	 * Creates a URL to a named or reverse route.
+	 *
+	 * @param string $uri Named or reverse route
+	 *
+	 * @return static
+	 *
+	 * @throws RouterException
+	 */
+	public static function route(string $uri)
+	{
+		// Check for a named or reverse-route
+		if ($route = Services::routes()->reverseRoute($uri))
+		{
+			return new static($route);
+		}
+
+		throw RouterException::forInvalidRoute($uri);
 	}
 
 	/**
@@ -92,7 +116,7 @@ class URL
 	 *
 	 * @return static
 	 */
-	final public static function fromRequest(IncomingRequest $request)
+	public static function fromRequest(IncomingRequest $request)
 	{
 		$path  = $request->detectPath($request->config->uriProtocol);
 		$query = isset($_SERVER['QUERY_STRING']) ? '?' . $_SERVER['QUERY_STRING'] : '';
@@ -109,7 +133,7 @@ class URL
 	 * @param string   $relativePath
 	 * @param App|null $config
 	 */
-	final public function __construct(string $relativePath = '', App $config = null)
+	public function __construct(string $relativePath = '', App $config = null)
 	{
 		$config = $config ?? config('App');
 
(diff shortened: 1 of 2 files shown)


lonnieezell:
Haven't looked at the code yet but my initial thought is - this is making the framework more complex and confusing than it needs to be. Can we just start prepping for a bigger release and fix this in the existing URI class? Have two classes so closely related to each other are going to be super confusing for any devs that dig into the inner workings. Or just to those who's IDE starts to suggest both classes for them. 

Additionally - isn't the purpose of `site_url` to be `base_url` but taking `index.php` settings into account?

MGatner:
We really need both. It doesn't need to be implemented this way, but we need to be able to do two things:
1. Build out a URI representing any valid URI for use anywhere one would want a URI
2. Build "project-specific" URIs, based off the configuration in `App` and (potentially) any special routes

Currently this is a real problem: https://github.com/codeigniter4/CodeIgniter4/blob/4cc14520e29787ac3441f20a4aec0017a3678600/system/HTTP/URI.php#L566-L584

I've tried to mitigate it in #4646 but this "bleeds" point 2 into point 1. `URI` should not know or care about the framework settings because it makes it impossible to create any URI one might want because the framework settings take over.

Currently we handle "project-specific" URIs with the URL Helper functions (`base_url()`, `site_url()`, `current_url()`) - the problem with these is that the final project has no concept of "what is project" versus "what is path" (e.g. http://example.com/foo/bar - is the `baseURL` http://example.com/ with "foo/bar" path? or is it http://example.com/foo/ with "bar" path?) so I decided to go with a class solution so we have some way of storing and accessing that info.

I originally had `URL` as `FrameworkURI` that was an extension of `URI`, but there are too many setters in `URI` that could "violate" an internal URI so I split it out; at that point I realized it was basically a full URL so went with the name, but I'm open to changing it.

MGatner:
> isn't the purpose of site_url to be base_url but taking index.php settings into account

Exactly. `site_url()` amounts to "take `base_url()`, add `App::$indexPage`, add any given path, be smart about slashes." `URL::to()` tries to encompass this behavior but with a little more utility of resolving named and reverse routes.

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	//--------------------------------------------------------------------
+
+	/**
+	 * Creates the base URL.
+	 *
+	 * @return static
+	 */
+	public static function base()
+	{
+		return static::public('');
```
paulbalandan, line 56: Since the class is `final`, then inheritance is no more possible. So, there's no need to use the `static` keyword. Instead use `self`.
```suggestion
		return self::public('');
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	 *
+	 * @return static
+	 */
+	public static function public(string $uri)
+	{
+		// Base URLs never include the index page
+		$config            = clone config('App');
+		$config->indexPage = '';
+
+		return new static($uri, $config);
```
paulbalandan, line 73:
```suggestion
		return new self($uri, $config);
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	 *
+	 * @param string $uri Additional URI string to include
+	 *
+	 * @return static
+	 */
+	public static function public(string $uri)
+	{
+		// Base URLs never include the index page
+		$config            = clone config('App');
+		$config->indexPage = '';
```
paulbalandan, line 71:
```suggestion
		$config = clone config('App');

		$config->indexPage = '';
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	 * @return static
+	 */
+	public static function current()
+	{
+		if (isset(self::$current))
+		{
+			return self::$current;
+		}
+
+		return static::fromRequest(Services::request());
```
paulbalandan, line 88:
```suggestion
		return self::fromRequest(Services::request());
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	/**
+	 * Creates a framework URL.
+	 *
+	 * @param string $uri
+	 *
+	 * @return static
+	 */
+	public static function to(string $uri)
+	{
+		return new static(rtrim($uri, '/ '));
```
paulbalandan, line 100:
```suggestion
		return new self(rtrim($uri, '/ '));
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	 * @return static
+	 *
+	 * @throws RouterException
+	 */
+	public static function route(string $uri)
+	{
+		// Check for a named or reverse-route
+		if ($route = Services::routes()->reverseRoute($uri))
+		{
+			return new static($route);
```
paulbalandan, line 117:
```suggestion
			return new self($route);
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	 * @param IncomingRequest $request
+	 *
+	 * @return static
+	 */
+	public static function fromRequest(IncomingRequest $request)
+	{
+		$path  = $request->detectPath($request->config->uriProtocol);
+		$query = isset($_SERVER['QUERY_STRING']) ? '?' . $_SERVER['QUERY_STRING'] : '';
+
+		return new static($path . $query, $request->config);
```
paulbalandan, line 136:
```suggestion
		return new self($path . $query, $request->config);
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * URI wrapper to work with complete project URLs.
+ * This class creates immutable instances.
+ */
+final class URL
+{
+	/**
+	 * Explicit URL to use for current.
+	 *
+	 * @var static|null
```
paulbalandan, line 29:
```suggestion
	 * @var self|null
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	 * @var string
+	 */
+	private $relativePath;
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Creates the base URL.
+	 *
+	 * @return static
```
paulbalandan, line 52:
```suggestion
	 * @return self
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+		return static::public('');
+	}
+
+	/**
+	 * Creates a URL to unrouted public files (typically assets).
+	 * Similar to base_url('path/to/file')
+	 *
+	 * @param string $uri Additional URI string to include
+	 *
+	 * @return static
```
paulbalandan, line 65:
```suggestion
	 * @return self
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+		$config            = clone config('App');
+		$config->indexPage = '';
+
+		return new static($uri, $config);
+	}
+
+	/**
+	 * Returns an instance representing the current URL.
+	 *
+	 * @return static
```
paulbalandan, line 79:
```suggestion
	 * @return self
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+
+		return static::fromRequest(Services::request());
+	}
+
+	/**
+	 * Creates a framework URL.
+	 *
+	 * @param string $uri
+	 *
+	 * @return static
```
paulbalandan, line 96:
```suggestion
	 * @return self
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+	{
+		return new static(rtrim($uri, '/ '));
+	}
+
+	/**
+	 * Creates a URL to a named or reverse route.
+	 *
+	 * @param string $uri Named or reverse route
+	 *
+	 * @return static
```
paulbalandan, line 108:
```suggestion
	 * @return self
```

system/HTTP/URL.php
```diff
@@ -0,0 +1,236 @@
...
+		throw RouterException::forInvalidRoute($uri);
+	}
+
+	/**
+	 * Returns an instance representing the URL from
+	 * an Incoming Request (as defined by Config\App).
+	 *
+	 * @param IncomingRequest $request
+	 *
+	 * @return static
```
paulbalandan, line 129:
```suggestion
	 * @return self
```

system/HTTP/URL.php
MGatner: On the topic, how do you think the class should be scoped? [Originally](https://github.com/codeigniter4/CodeIgniter4/pull/4647/commits/42a60ce50059ec4f492fd72078b6b3a972853ee2) I had a normal class with `final` static and constructor methods (hence the leftover `static` references here) but I decided that nobody would go through the hassle of extending for the few allowed methods so made the whole thing `final`.

lonnieezell:
I'll be honest. I'm not a fan of having both. I understand your reasoning but think it will make things more complex/confusing for the developers. I wish we could find a way to refactor the current class to work in conjunction with the url helpers. 

That being said, if the rest of the team is fine with it then I'll shut up :)

MGatner:
@lonnieezell I understand. Let's think of ways we could make this better. I do think the helpers could be "beefed" up to cover this class but we would lose the ability to distinguish path parts (see my comment about "what is project" versus "what is path"). That might be an acceptable tradeoff? But there are some real issues with the current setup that need addressing. Just an example, `IncomingRequest` is clear that it only builds out the URI it needs for itself: https://github.com/codeigniter4/CodeIgniter4/blob/2e0e020920844f46ea78178fed297edda03f27d7/system/HTTP/IncomingRequest.php#L153-L160

> If you need that, use current_url().

... but then this is the entirety of `current_url()`:
```
$uri = clone Services::request()->uri;
```

I believe that a support class (in this case `URL`) gives us a lot more flexibility than the helper functions for working with framework URIs - maybe we could make it an internal class? or move it out of `HTTP`?

PR #4648, opened 2021-05-05
[ci skip] Remove ajax filter reference from docs. Fixes #2314

### LINKED ISSUE #2314: Feature: AJAX filters don't work
**Describe the bug**
According to the User Guide on `$methods` in **app/Config/Filters.php**:
> In addition to the standard HTTP methods, this also supports two special cases: ‘cli’, and ‘ajax’.

However, I cannot find anywhere in the code that would run AJAX filters. Method is set in **system/Filters/Filters** with no clause for AJAX:
```
$method = strtolower($_SERVER['REQUEST_METHOD'] ?? 'cli');
if (array_key_exists($method, $this->config->methods))
```

Unless I've missed something, I think this isn't implemented, and my test Filter is definitely not firing on AJAX calls.

**CodeIgniter 4 version**
Latest develop

**Affected module(s)**
Controller Filters

lonnieezell:
Removes the reference to `ajax` filters as an available method since it is not implemented. 

Long time coming :)

### DIFF
--- a/user_guide_src/source/incoming/filters.rst
+++ b/user_guide_src/source/incoming/filters.rst
@@ -167,11 +167,8 @@ specify the method name in lowercase. It's value would be an array of filters to
         'get'  => ['baz'],
     ]
 
-In addition to the standard HTTP methods, this also supports two special cases: 'cli', and 'ajax'. The names are
-self-explanatory here, but 'cli' would apply to all requests that were run from the command line, while 'ajax'
-would apply to every AJAX request.
-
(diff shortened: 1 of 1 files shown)


MGatner:
Bravo! That was a long time 😂 I would still like to see something like this down the road but it clearly has not been a priority.

MGatner (approved):

lonnieezell:
Agreed. Trying to go through the back log and start pruning all of these issues. Some have been around way too long :)

MGatner:
I know, it's a disgrace how long our lists have gotten. I've been trying to pack in the last few things for a release and then join you on the "pruning".

PR #4652, opened 2021-05-07
Add branch alias for develop branch

paulbalandan:
**Description**
Add a branch alias to composer.json to alias `develop` branch as `4.x-dev`. This way libraries can just write `codeigniter4/codeigniter4: 4.x-dev` in their `require` section.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,8 @@
 	},
 	"extra": {
 		"branch-alias": {
-			"dev-develop": "4.x-dev"
+			"dev-develop": "4.x-dev",
+			"dev-master": "4.x"
 		}
 	},
 	"autoload": {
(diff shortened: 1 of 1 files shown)


MGatner:
Does this keep them from having to add the repository as well?

paulbalandan:
> Does this keep them from having to add the repository as well?

No. Since the repo is not a composer package, the entry to `repositories` section is still needed.

[Alias](https://getcomposer.org/doc/articles/aliases.md)

MGatner:
Ah thanks that's helpful. According to this I think we should also add `"dev-master": "4.x"`

MGatner (approved):

PR #4655, opened 2021-05-09
[Scripts] Make sure bash script still works in Windows

paulbalandan:
#4650 does not work on Windows.

```console
> CodeIgniter\ComposerScripts::postUpdate
> if [ -f "admin/setup.sh" ]; then bash admin/setup.sh; fi
Script if [ -f "admin/setup.sh" ]; then bash admin/setup.sh; fi handling the post-update-cmd event returned with error code 1
```

Fixed:
```console
> CodeIgniter\ComposerScripts::postUpdate
> bash -c "if [ -f admin/setup.sh ]; then bash admin/setup.sh; fi"
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
 	"scripts": {
 		"post-update-cmd": [
 			"CodeIgniter\\ComposerScripts::postUpdate",
-			"if [ -f \"admin/setup.sh\" ]; then bash admin/setup.sh; fi"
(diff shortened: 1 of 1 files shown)


samsonasik (approved):

samsonasik:
Thank you @paulbalandan

MGatner:
Nice work.

PR #4657, opened 2021-05-09
Reset Single Service

MGatner:
**Description**
Frequently I find myself doing acrobatics to accomplish this. It is past time this was added: the ability to reset a single Service by its name.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -307,8 +307,7 @@ class BaseService
 	 */
 	public static function resetSingle(string $name)
 	{
-		unset(static::$mocks[$name]);
-		unset(static::$instances[$name]);
+		unset(static::$mocks[$name], static::$instances[$name]);
 	}
 
 	/**
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -301,6 +301,17 @@ class BaseService
 	}
 
 	/**
+	 * Resets any mock and shared instances for a single service.
+	 *
+	 * @param string $name
+	 */
+	public static function resetSingle(string $name)
+	{
+		unset(static::$mocks[$name]);
+		unset(static::$instances[$name]);
+	}
+
+	/**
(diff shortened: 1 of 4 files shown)


paulbalandan (commented):
To clarify, what you mean here is to reset a single "shared" service, right? Not reset a service given by `single_service`.

You can fix the failing tests by adding the `resetSingle` method in the `$excl` of `CommonSingleServiceTest::serviceNamesProvider()`

MGatner:
`single_service()` ignores shared services, and mocks only apply to shared services, so this should not affect that function.

Thanks for the test fix, I will change that when I switch back to that branch.

paulbalandan:
Thanks. I got confused while reading the title.

system/Config/BaseService.php
```diff
@@ -300,6 +300,17 @@ public static function reset(bool $initAutoloader = false)
...
 
+	/**
+	 * Resets any mock and shared instances for a single service.
+	 *
+	 * @param string $name
+	 */
+	public static function resetSingle(string $name)
+	{
+		unset(static::$mocks[$name]);
+		unset(static::$instances[$name]);
```
paulbalandan, line 311: Combine as one call to `unset`.

Can we also check if the name is set first before unsetting.

MGatner: Will combine!

> Can we also check if the name is set first before unsetting.

What is the benefit to this? `unset` does not care so it adds unnecessary steps.

MGatner, line 311:
```suggestion
		unset(static::$mocks[$name], static::$instances[$name]);
```

paulbalandan: Hmm. Ok. I was today years old when I learned unsetting an inexistent variable raises no error. 🤔🤪

paulbalandan (approved):

PR #4658, opened 2021-05-09
Add stricter validation for URLs

### LINKED ISSUE #3156: valid_url does not work correctly

MGatner:
**Description**
Fixes #3156.

This one was a bit of a contentious conversation. I ended up going with a separate rule that borrows the validation regex from Laravel and still using `filter_var()`. I don't know that there is an ideal way to do this but this seems to me a solid middle ground between expectations and implementation.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -328,7 +328,9 @@ class FormatRules
 	}
 
 	/**
-	 * Checks a URL to ensure it's formed correctly.
+	 * Checks a string to ensure it is (loosely) a URL.
+	 * Warning: this rule will pass basic strings like
+	 * "banana"; use valid_url_strict for a stricter rule.
 	 *
 	 * @param string $str
 	 *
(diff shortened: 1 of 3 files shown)


MGatner:
@lonnieezell Please read through the linked issue (sorry, I know it is long) and see if you still feel the same. I was also in the camp that "muffins" should not count as a valid URL, but there were lots of differing opinions that I want to respect.

lonnieezell:
Ok, I went back and read through the comments again. Here's the thing - the current `valid_url` is worthless. It's no different then validating it's a string, which we have other rules for. And, yes, I probably wrote this function thinking "it matches the spec so it's good, right?" not realizing how it doesn't help. 

I think about the only difference between the current method and the new method is that the new method a) doesn't add on a scheme and b) validates the scheme is one of `http`, `https`, `mailto`, `tel`, or `sms`. If we limit the schemes like that, then we can't validate against App store links that take users to the correct app in Apple or Android stores, ssh addresses, etc. Is that something that's desired? 

Honestly, I'm wondering if simply removing the automatic adding of scheme from the current function is all we need to do. PHP will take care of the rest. We would need to made a big deal of it in our upgrade guide and our announcement, but I think that's the right thing to do. As you and Instyre pointed out - we weren't validating URL's we were validating domains....

MGatner:
I'm good with that. I'm not sure if @michalsn is around to have a last word (or if he cares).

I came across this: https://gist.github.com/kpobococ/92f120c6c4a9a52b84e3. It is pretty much a file to itself, but if we want to offer strict RFC3986 validation that could be an option (it's public domain).

michalsn:
Personally, at this point, I'm okay with changing how `valid_url()` method works entirely since now it brings a lot of confusion.

But to make it works we probably should restrict some "allowed" schemes (with the possibility to override it) in other case we gonna end up with people complaining that `whatever://codeigniter.com` is treated as a valid URL.

Maybe something like:
```php
public function valid_url(string $str, string $validSchemes = 'http,https,mailto,tel,sms')
{
    $scheme       = parse_url($str, PHP_URL_SCHEME);
    $validSchemes = explode(',', $validSchemes);
    
    if (! in_array($scheme, $validSchemes))
    {
        return false;
    }
    
    return filter_var($str, FILTER_VALIDATE_URL) !== false;
}
```
This way we can always change the valid schemes if we need to.

Does anybody have a quick comparison for `RFC2396` vs `RFC3986`? Honestly, I have no idea what are the differences.

paulbalandan:
Superseded by #5268

Issue #4356, opened 2021-02-26
Discussion: Interface Discrepancies


MGatner:
# Introduction
I want to begin a "big picture" discussion about something that has caused numerous issues in the past: framework interfaces. Let me also point out up front that *this is currently a problem*. We have done a handful of patch jobs to "make things work" but they have not addressed the underlying issues. A few references for example:
* #4329
* #4280
* #4285
* #3900

# Interfaces
An interface is a contract between the code and our developers:
1. A class that implements an interface must have all the defined methods.
2. An interface cannot be changed without a major release.
3. Anything that uses an interface in a typehint or docblock must use only that interface's methods.

# Problem
Number three is where we get in trouble. Due to the nature of frameworks we need to move classes around a lot. In order to allow extending/replacing core components we use interfaces. In our current code base, sometimes this has been done well and maintained and sometimes it has not. Two examples...

### `CacheInterface`
The `CacheInterface` is an example of interfacing handled well. The interface has 10 public methods, all the handlers implement those methods without embellishment, and our Cache Service returns the interface: https://github.com/codeigniter4/CodeIgniter4/blob/ecb250777679fa0d2c8638bd3d4012c67cf5a706/system/Config/Services.php#L90-L99

This means that other components like `Throttler` can use `CacheInterface` safely:
https://github.com/codeigniter4/CodeIgniter4/blob/ecb250777679fa0d2c8638bd3d4012c67cf5a706/system/Throttle/Throttler.php#L30-L35

... and any developer who wished to use custom caching can [define their own](https://codeigniter4.github.io/CodeIgniter4/concepts/services.html#defining-services).

### `ConnectionInterface`
I'm going to pick on the database layer because of this recent issue (#4329) and because it is a bit of a mess generally. `ConnectionInterface` has 16 public methods but they are not nearly comprehensive enough to support all our required database methods so we use `BaseConnection implements ConnectionInterface` and then have handlers extend `BaseConnection`. This way when `MigrationRunner` needs to check for the "migrations" table is can use the embellished method provided by `BaseConnection`: https://github.com/codeigniter4/CodeIgniter4/blob/ecb250777679fa0d2c8638bd3d4012c67cf5a706/system/Database/MigrationRunner.php#L935-L944

[...]

PR #4660, opened 2021-05-09
[Feature] Introduce the `Debug\Deprecation` utility component

paulbalandan:
**Description**
This PR introduces the `CodeIgniter\Debug\Deprecation` namespace containing a set of classes and traits to be used in deprecating certain elements in the framework. This can be used in the future for debugging deprecated elements that needs eventual removal.

**1. CodeIgniter\Debug\Deprecation\Deprecation**
This final static class is the "trigger" for the deprecation proper. It hosts a variety of methods to call in a deprecation message for a particular element (e.g., a class, method, property, or method parameter).

This has two modes on how to handle a deprecation. If mode is `Deprecation::LOG_MESSAGE`, which is the default mode, the deprecation call will just log the message to our registered loggers with a default log level of `error` (to conform with the default logging threshold and increase the chance the message is logged and read by the user). If mode is set to `Deprecation::THROW_EXCEPTION`, a `DeprecationException` will be thrown containing the deprecation message thus halting the script execution. Mode can be set by `Deprecation::setMode()` and retrieved by `Deprecation::mode()`.

**2. CodeIgniter\Debug\Deprecation\DeprecationException**
This exception class is a child of `ErrorException` class. The original approach was to use the regular mechanic of `trigger_error($message, E_USER_DEPRECATED)` but I found out that our error handler catches this error and throws an `ErrorException` instead. So, to make a short cut, this class is thrown instead directly. This class adds a minor tweak to the error message and ensures `$severity` level is `E_USER_DEPRECATED`.

**3. CodeIgniter\Debug\Deprecation\DeprecatedClassMethodTrait**
This support trait is intended to be imported by classes with deprecated methods and doesn't want those to be used outside the class. This trait can also be used by parent classes to prevent child classes to use their deprecated methods. The trait intercepts calls to methods via the magic `__call` and `__callStatic` methods. *(More info on the class definition)*

**4. CodeIgniter\Debug\Deprecation\DeprecatedClassPropertyTrait**
This is similar to the `DeprecatedClassMethodTrait` except that this trait intercept calls to properties instead using the magic `__get` and `__set` methods. This trait is available only for properties in object context because PHP do not currently have `__getStatic` and `__setStatic` methods. *(More info on the class definition)*

Localized deprecation messages are stored in `system/Language/en/Deprecation.php`.

When using the traits, please be aware of the notion of [overloading in PHP](https://www.php.net/manual/en/language.oop5.overloading.php). If the inaccessible property/method is declared in the current scope, the traits' specially tweaked magic methods won't be invoked. That is, accessing a deprecated private property within the class will not emit a deprecation message since this is accessible.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

**Pending:**
- [ ] User guide updated
- [ ] Integration into the framework internals

### DIFF
--- a/system/Debug/Deprecation/DeprecatedClassMethodTrait.php
+++ b/system/Debug/Deprecation/DeprecatedClassMethodTrait.php
@@ -1,0 +1,160 @@
+<?php
+
+/**
+ * This file is part of the CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Debug\Deprecation;
+
+use BadMethodCallException;
+use Error;
+use ReflectionMethod;
+
+/**
+ * This support trait can be used in two flavors.
+ * - Used to prevent child classes from using deprecated parent methods;
+ * - Used to prevent method calls outside the class.
+ *
+ * __(1) Used to prevent child classes from using deprecated parent methods__
+ *
+ * If intention is to encapsulate code within the parent class, this trait
+ * should be imported into the parent. The deprecated method should be set
+ * as `private` and the name and its replacement will be setup in a private
+ * `$deprecatedMethods` property array. The property is set private to prevent
+ * overrides by child classes. If the deprecated method is static, the setup
+ * will be made in a private static `$deprecatedStaticMethods` property array.
+ *
+ * __(2) Used to prevent method calls outside the class__
+ *
+ * To prevent out-of-class method calls of the deprecated methods, the method
+ * should be set as either `protected` or `private` (i.e., preventing access).
+ * After that, add the name of the method and its replacement into the respective
+ * array property, depending if the method is static or not.
+ *
+ * __NOTE:__
+ *
+ * This trait will give access to previously inaccessible methods of a class. If this
+ * behavior is not desired, the names of methods to be excluded and to remain as
+ * not accessible should be added to the private static `$methodAccessExclusions`
+ * property.
+ *
+ * @property-read array<string, string>  $deprecatedMethods
+ * @property-read array<string, string>  $deprecatedStaticMethods
+ * @property-read array<integer, string> $methodAccessExclusions
+ */
+trait DeprecatedClassMethodTrait
+{
+	/**
+	 * Intercept calls to inaccessible and/or deprecated non-static methods.
+	 *
+	 * @param string  $name      Name of the static method
+	 * @param mixed[] $arguments Arguments to be passed
+	 *
+	 * @throws Error                  When the method is not deprecated but is not allowed for access
+	 * @throws BadMethodCallException When the method is not defined
+	 * @throws DeprecationException   When the method is deprecated and handling mode is `Deprecation::THROW_EXCEPTION`
+	 *
+	 * @return mixed
+	 */
+	public function __call(string $name, array $arguments)
+	{
+		if (! method_exists($this, $name))
+		{
+			throw new BadMethodCallException(sprintf('Trying to access undefined method "%s::%s()".', static::class, $name));
+		}
+
+		if (array_key_exists($name, $this->deprecatedMethods))
+		{
+			// a deprecated method is always resolved to the class importing
+			// this trait, so this is always `self::class`
+			$classMethod = self::class . '::' . $name;
+
+			Deprecation::triggerForMethod($classMethod, $this->deprecatedMethods[$name]);
+		}
+
+		// did we accidentally grant access to methods we should not let everyone to use?
+		if (in_array($name, self::$methodAccessExclusions, true))
+		{
+			$visibility = self::determineMethodVisibility($name);
+			$backtrace  = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 3);
+
+			throw new Error(sprintf(
+				'Call to %s method "%s::%s()" from scope "%s".',
+				$visibility,
+				self::class,
+				$name,
+				$backtrace[1]['class'] ?? $backtrace[2]['class'] ?? 'unknown'
+			));
+		}
+
+		return $this->{$name}(...$arguments);
+	}
+
+	/**
+	 * Intercept calls to inaccessible and/or deprecated static methods.
+	 *
+	 * @param string  $name      Name of the static method
+	 * @param mixed[] $arguments Arguments to be passed
+	 *
+	 * @throws Error                  When the method is not deprecated but is not allowed for access
+	 * @throws BadMethodCallException When the method is not defined
+	 * @throws DeprecationException   When the method is deprecated and handling mode is `Deprecation::THROW_EXCEPTION`
+	 *
+	 * @return mixed
+	 */
+	public static function __callStatic(string $name, array $arguments)
+	{
+		if (! method_exists(self::class, $name))
+		{
+			throw new BadMethodCallException(sprintf('Trying to access undefined static method "%s::%s()".', self::class, $name));
+		}
+
+		if (array_key_exists($name, self::$deprecatedStaticMethods))
+		{
+			// a deprecated static method is always resolved to the class
+			// importing this trait, so this is always `self::class`
+			$classMethod = self::class . '::' . $name;
+
+			Deprecation::triggerForStaticMethod($classMethod, self::$deprecatedStaticMethods[$name]);
+		}
+
+		// did we accidentally grant access to methods we should not let everyone to use?
+		if (in_array($name, self::$methodAccessExclusions, true))
+		{
+			$visibility = self::determineMethodVisibility($name);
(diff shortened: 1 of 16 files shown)


MGatner:
I like what I'm seeing! It looks like you have this under control, but I expect @WinterSilence would be helpful for a review (https://github.com/codeigniter4/CodeIgniter4/issues/4356#issuecomment-787547781).

paulbalandan:
> I like what I'm seeing! It looks like you have this under control, but I expect @WinterSilence would be helpful for a review ([#4356 (comment)](https://github.com/codeigniter4/CodeIgniter4/issues/4356#issuecomment-787547781)).

I have fairly covered all possible instances where a deprecation may appear. What baffles me is how to detect usage of static properties and constants.

WinterSilence:
class DeprecatedClassPropertyTrait - 100% bad idea, because trait reset magic methods.
That's all too "heavy". Why not use `@deprecated` tag or easy helper class like as:
~~~php
class Deprecated
{
     public static function method(string $name, string $description = null): void;
     public static function property(string $name, string $description = null): void;
     /**
      * Common method called in `method()`/`property()`.
      */
     protected static function log(string $name, string $description = null): void;
}
~~~
~~~php
class BCClass
{
    /**
     * @var int ...
     * @deprecated Use ... instead
     */
    protected $deprecatedProperty = 1;

    /**
     * @deprecated Use ... instead
     */
    public function deprecatedMethod()
    {
          Deprecated::method(__METHOD__, 'Use ... instead');
           .... code ...
    }
}
~~~

paulbalandan:
I'm aware using magic methods has side effects. But this is the only way I can think of to intercept calls to deprecated properties. This idea is implemented by and borrowed from TYPO3 CMS.

Use of `@deprecated` is not disregarded. But the benefit of that is only for those using IDEs that parses this docblock. Those who do not use IDEs will not know immediately if they are using a deprecated element or not, unless they read the documentation or changelog.

If you will read the PR description, I have discussed there that I initially went on using the `trigger_error(..., E_USER_DEPRECATED)` mechanic. But, since our error handler catches this and throws an `ErrorException` instead, I went on throwing the exception directly for shortcut. I created a custom class extending this to make the error message a bit customized.

My goal for this is to be comprehensive enough to cover all possible instances where a deprecation might arise. I have added some validation rules in case this class is used in userland code. This class is intended primarily for internal use within the framework to mark those deprecated elements but I realized it would be a shame if extensions cannot use this in their codebase. So I allowed the use but ensured this will "not be used for fun" and adding gibberish and complaining this does not work as intended. Besides, I consider the component to be advanced that only a few would use this in their code but we should always be prudent and add safeguards along the way.

paulbalandan:
> @paulbalandan
> 
> > I'm aware using magic methods has side effects.
> 
> I meant other thing: many classes define magic methods, that's trait(s) replaces them.

Ah yes. I have thought this might be the case for some of our classes. In such cases I am thinking of aliasing the method during import.
```php
class Foo
{
    use DeprecatedClassPropertyTrait {
        __set as otherName;
    }

    public function __set(...)
    {
        $value = self::otherName(...);
        ...
    }
}
```

paulbalandan:
> @paulbalandan
> 
> > Use of `@deprecated` is not disregarded. But the benefit of that is only for those using IDEs that parses this docblock.
> 
> I mean parse phpdoc's and generate "deprecated" list, it's can be display on package update or/and at autoload at "debug mode".

I am also currently contemplating on an efficient way to do this. My initial step is to define a mode for the deprecation. The default mode is to log the message. Then, we can set the mode to "scream" the deprecation message when used.

My next step, which I am looking for the best way, is to dynamically change the mode. On userland, I am looking at adding the setup to the `php spark serve` call, like adding a flag. Something like `php spark serve --deprecation-mode=throw` will make encountered deprecations to throw. But I am not yet sold on this.

For framework use, I am thinking of an environment flag to make this strict and throw deprecations.

WinterSilence:
~~~json
{
   ...
  "scripts": {
      ...
      "post-package-update": "CodeIgniter\\ComposerScripts::postPackageUpdate"
  },
  ...
}
~~~
~~~php
namespace CodeIgniter;

...
use ReflectionClass;
use ReflectionProperty;
use Composer\Installer\PackageEvent;
use Composer\Autoload\ClassMapGenerator;
// package 'phpdocumentor/reflection-docblock'
use phpDocumentor\Reflection\DocBlockFactory;

class ComposerScripts
{
    ...

    public static function postPackageUpdate(PackageEvent $event)
    {
        $deprecated = [];
        $bocBlockFactory  = DocBlockFactory::createInstance();

        $classes = ClassMapGenerator::createMap(__DIR__, null, $event->getIO());
        $classes = array_keys($classes);
        foreach ($classes as $class) {
           $reflectionClass = new ReflectionClass($class);
           foreach (array_merge($reflectionClass->getProperties(), $reflectionClass->getMethods()) as $reflection) {
               if (!$reflection->getDocComment() || $reflection->getDeclaringClass() !== $reflection->class) {
                   continue;
               }
               $tags = $bocBlockFactory->create($reflection)->getTagsByName('deprecated');
               if ($tags) {
                   if ($reflection instanceof ReflectionProperty::class) {
                        $name = $reflection->class . '::$' . $reflection->name;
                        $deprecated[$name] = $name . ' ' . implode(' ', $tags);
                   } else {
                        $name = $reflection->class . '::' . $reflection->name;
                        $deprecated[$name] = $name . '() ' . implode(' ', $tags);
                   }

               }
           }
           
        }

        $event->getIO()->write('<warning>Deprecated:' . PHP_EOL . implode(PHP_EOL, $deprecated) .'</warning>');
    }
}
~~~

paulbalandan:
> ```json
> {
>    ...
>   "scripts": {
>       ...
>       "post-package-update": "CodeIgniter\\ComposerScripts::postPackageUpdate"
>   },
>   ...
> }
> ```

I like this. I'll include this in the integrations. However, I think it is quite intrusive to always show the deprecation message whether after package update or app running. Users would want this lengthy message to show if they are actually using those deprecated elements. This script however will be useful for the internal updating of the framework to locate all deprecated elements at once.

paulbalandan:
> @paulbalandan
> 
> > In such cases I am thinking of aliasing the method during import.
> 
> it's magic methods :(

Apparently, it works. https://3v4l.org/IBA3o

WinterSilence:
@paulbalandan yes, that's work, but why you think what's better than something like [`Deprecated`/`BCClass`](https://github.com/codeigniter4/CodeIgniter4/pull/4660#issuecomment-837539214)?

WinterSilence:
@paulbalandan
> Apparently, it works. https://3v4l.org/IBA3o

I try say: it's not work as magic method, it's can't work as "class using trait use trait's methods, but that's methods call class methods if they defined".

WinterSilence:
@paulbalandan
> The original approach was to use the regular mechanic of trigger_error($message, E_USER_DEPRECATED) but I found out that our error handler catches this error and throws an ErrorException instead. 

fixed in #4680

paulbalandan:
Since this is being reviewed anyway, I am requesting review now to know others' opinions on this before I continue work on this.

WinterSilence:
@paulbalandan ok, I vote to use simple helper like as https://github.com/koseven/koseven/blob/devel/system/classes/KO7/Core.php#L1050

michalsn:
It looks like @paulbalandan has put a lot of work into this PR, but if the main goal here is to "mark" deprecated methods then I would be in favor of doing it in a very simple way, as suggested by @WinterSilence (helper function + detection of the depreciation tag). 

This is just my preference. I like to keep things simple if it's possible - less code is usually better. Both from the perspective of maintaining the code and from the perspective of those who will use it later.

I'm still curious about others' opinions.

MGatner:
I'd like to get this resolved. I agree @paulbalandan has done lots of work, and it is quality code. It is definitely more complex than the simple function. I'm still open to hearing more opinions, but here is where I am at right now:
* If this utility is something we can offer to developers as a framework feature to enhance their own project & library workflow, I think it is worthwhile pursuing it as a development component (which the framework uses as well, of course)
* If this ends up mostly being an internal tool to keep us from forgetting to put `deprecate('Use something else')` into our classes then the complexity may be too daunting, especially in the interest of encouraging more contributors

paulbalandan:
Thanks, @MGatner !  🎉 I originally intended this to be mostly internal while it is on its early stage but while working on it I conclude we can offer this to "advanced" users, especially CI4 module creators.

From a bit of research long ago in support of my side, I came across [`doctrine/deprecations`](https://github.com/doctrine/deprecations) which is somehow similar to this PR's architecture. So I think this could be useful internally and _externally_.

lonnieezell:
Late to the party, but here's my $0.02. I also prefer things kept simpler when possible. 

In this case, I see a lot of value from having deprecated methods show up in the logs. Having the framework add a call to trigger that seems simple enough and very helpful. 

I also like the two different modes. Nice touch. 

I'm not convinced at all about the traits to intercept calls. To me, those feel like more complexity than necessary, and potential gotchas waiting to happen.

PR #4667, opened 2021-05-10
PSR: Cache

MGatner:
**Description**
The core framework intentionally does not aim to meet all PSR interfaces, which for the most part works fine. Some PSR components allowing provisions for libraries and projects that complete the requirements. In this case, having adapters available in the framework itself can prevent CodeIgniter users from having to install additional third-party libraries that duplicate functionality.

This is the first PR to add adapters for PSR components, targeting Cache. There are two Cache PSR's from FIG:
* [Caching Interface](https://www.php-fig.org/psr/psr-6/)
* [Common Interface for Caching Libraries](https://www.php-fig.org/psr/psr-16/)

These adapters sit on top of `CodeIgniter\Cache`'s handlers. Because of the additional layer the PSR adapters are not recommended for direct use except when the PSR interfaces are required.

Tests are provided by php-cache who offers complete integration testing for PSR-6 and PSR-16 implementations.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


system/Psr/Cache/Pool.php
```diff
@@ -0,0 +1,339 @@
...
+	 *
+	 * @param object|null $object
+	 *
+	 * @throws CacheArgumentException
+	 */
+	public function __construct($object = null)
+	{
+		if (is_null($object))
+		{
+			$this->adapter = Services::cache();
```
lonnieezell, line 50: I think standardizing on using the `service()` helper is a good thing for us. I think it keeps us safer if we make changes down the line to ensure proper service detection ordering since it doesn't call a specific class.

lonnieezell (approved):
I have no problem with this and the implementation looks fine. I just left one comment.

MGatner:
@samsonasik any idea about this Rector error causing test failures?
```
Class                                                                  
         'Symplify\ComposerJsonManipulator\Bundle\ComposerJsonManipulatorBundle' not found
```

PR #4672, opened 2021-05-11
Split URL Helper tests

MGatner:
**Description**
In preparation for "Stage Two" of URI rework (targeting URL Helper) this PR splits apart the massive URL Helper test into 4 different files to make it easier to control `$_SERVER` manipulations and minimize overlap. These tests will have significant changes in the next PR so I wanted to split them out first to make it clear what actually is changed.

The only content updates were to `testCurrentURLReturnsBasicURL()` and `testCurrentURLReturnsObject()` which were actually bugged but didn't show it because `$this->config` was not injected.

Note: I realize there is significant code overlap now in the staging methods but these also are likely to change in the next PR so I have not consolidated them on purpose.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide


paulbalandan (commented):
Just nitpicking on my end. Can you make the name of the tests with Url in it? Like BaseTest to BaseUrlTest.

MGatner:
> Just nitpicking on my end. Can you make the name of the tests with Url in it? Like BaseTest to BaseUrlTest.

Glad to! Any idea why test are flipping out?

MGatner:
I'm baffled about these tests. I reran them three times and they keep having these same weird errors, even though all I did was split one file into four. Other PRs have run successful tests since so I'm not sure what to think of it.

lonnieezell:
I've seen cases where tests relied on $_SERVER or $_SESSION changes that were happening in a previous file. When that file was ran on it's own its fine, but ran in sequence and it didn't work. In those cases I was able to fix it clearing the session (or whatever) during the test setUp. Not saying that's the case here, but something is definitely going on. 

The autoloader isn't getting reset somewhere, though I don't know if it matters: 

```
'loader' => CodeIgniter\Autoloader\FileLocator Object &000000005899509a000000007780000a (
                'autoloader' => CodeIgniter\Autoloader\Autoloader Object &000000005898fd0a000000007780000a (
                    'prefixes' => Array &23 (
                        'CodeIgniter' => Array &24 (
                            0 => '/home/runner/work/CodeIgniter4/CodeIgniter4/system/'
                            1 => '/home/runner/work/CodeIgniter4/CodeIgniter4/system/'
                            2 => '/home/runner/work/CodeIgniter4/CodeIgniter4/system/'
                            // and a few hundred more of these...
```

paulbalandan:
I don't know if it'll make sense, but try resetting the mocked configs after each test?

MGatner:
Okay good suggestions - I will look at what changed and try mimicking.

MGatner:
That wasn't it - I didn't change the staging at all in this phase. They all reset `$_SERVER = []` in `tearDown()` and `Services::reset(true)` in `setUp()`, same as live.

MGatner:
Try closing and reopening I guess.

paulbalandan:
try backupGlobals enabled for all tests. I see the original test but the split tests do not except for one

MGatner:
Ooo good catch! I bet that is it, that would make total sense.

paulbalandan (approved):

PR #4673, opened 2021-05-11
 Copy Psr\Cache and Psr\SimpleCache to ThirdParty

paulbalandan:


paulbalandan:
> Since these adapters are explicitly for integrating external libraries, what if we made the PSR packages as dev requirements instead? Download users would already need to install the libraries they are referencing so will already be in charge of adding dependencies. And the packages will necessarily all `require` the PSR interfaces already, so Composer users will be fine.

That's fine to me. But I think we need to add a guard like `if (! class_exists(Psr\Cache\CacheInterface::class)` to prompt users to download the package. That should go to the entrypoint of the adapters. Is that in SupportTrait?

MGatner:
It won't work because the classes themselves already need to implement the interfaces. We need to decide whether we are going to provide these or not, and if not I think the best we can do is provide instructions for download users in the User Guide.

paulbalandan:
Looping @lonnieezell for his take on this.

I can take either way depending on what majority wants.

WinterSilence:
@MGatner https://github.com/codeigniter4/CodeIgniter4/blob/3509f60fa86c5c8a472ceb1d448722d5c4d959da/system/Psr/Cache/SimpleCache.php#L15 `Psr\SimpleCache` must be required

MGatner:
@WinterSilence It only needs to be required for people using that class. The way the autoloader lazy-loads there should be no problem having that file in someone's project without `psr-simplecache`. We are covered here by `require-dev` so all our analysis tools will be fine. For people actually planning to use it with a third-party package that package will already `require` it so the class will be present.

WinterSilence:
@MGatner it's added for users not uses composer

WinterSilence:
@MGatner it's not user friendly and not documented in code.  I think need move it in separated package.

MGatner:
I could see moving it to a separate library. We would need to decide this ASAP before a release (or just leave this part out would work). 

@paulbalandan @lonnieezell what do you think?

lonnieezell:
Making the PSR compatibility layer to a separate package is appealing to me until such a time as we actually implement the interfaces.

MGatner:
After sleeping on this I definitely think moving to a separate package is the way to go. Thank you for the suggestion @WinterSilence.

@paulbalandan Thank you for your work on this PR - sorry it will go unused but I enjoyed seeing how efficient your new `ComposerScripts` are for expansion.

PR #4676, opened 2021-05-11
Refactor CommandRunnerTest to make it a little faster

paulbalandan:
~This is the best refactor that I can do. Time went down to at most 5 seconds faster.~
~Apparently, nothing changed running in Ubuntu.~
Verdict: The test itself is really sluggish.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -64,6 +64,9 @@
 					<element key="tabulate">
 						<boolean>false</boolean>
 					</element>
+					<element key="collectBare">
+						<boolean>true</boolean>
+					</element>
 				</array>
 			</arguments>
 		</extension>
--- a/tests/system/CLI/CommandRunnerTest.php
+++ b/tests/system/CLI/CommandRunnerTest.php
@@ -1,66 +1,47 @@
 <?php
+
 namespace CodeIgniter\CLI;
 
-use CodeIgniter\Config\DotEnv;
-use CodeIgniter\HTTP\IncomingRequest;
-use CodeIgniter\HTTP\Response;
-use CodeIgniter\HTTP\URI;
-use CodeIgniter\HTTP\UserAgent;
+use CodeIgniter\Log\Logger;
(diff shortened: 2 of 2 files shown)


paulbalandan:
Why is the ListCommand an expensive command?? 🤔

MGatner:
Are you certain `Expeditable` is functioning correctly? It's fishy that `testGoodCommand()` is marked slow but its almost-duplicate `testDefaultCommand` is not.

paulbalandan:
> Are you certain `Expeditable` is functioning correctly? It's fishy that `testGoodCommand()` is marked slow but its almost-duplicate `testDefaultCommand` is not.

I have thought of that too. So I've run the test with Xdebug listening on the background. I've ascertained the bare times are being recorded in the $GLOBALS array and used in the output. So it means the actual test is taking long.

paulbalandan:
What's interesting though is when I switched the two tests' places running testDefaultCommand first then testGoodCommand, the slow time gets attached to testDefaultCommand. I've also added breakpoints and noted that the CommandRunner keeps on compiling all available commands **per test run** instead of once only during constructor initialization.

MGatner:
> the slow time gets attached to testDefaultCommand

That's baffling, because all that amounts to in the code is:
```
		$command = array_shift($params);

		if (is_null($command))
		{
			$command = 'list';
		}
```
... which should be a fraction of a nano-second 😳

paulbalandan:
I think the culprit is in the Commands::discoverCommands when running $this->commands->run

MGatner:
This brings up the question: is there any reason not to use `Expeditable` for the entire test suite? IMO it is the desired reporting behavior.

paulbalandan:
IMO I would preempt first usage of `Expeditable` until we can track down all slow tests and optimize what we can do to make them faster. After exhausting all code refactors available and are still slow, then I would add the trait into the test suite.

PR #4683, opened 2021-05-13
Optimize CommandRunner and Commands

paulbalandan:
**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/CommandRunner.php
+++ b/system/CLI/CommandRunner.php
@@ -21,13 +21,11 @@ use ReflectionException;
 class CommandRunner extends Controller
 {
 	/**
-	 * The Command Manager
+	 * Instance of class managing the collection of commands
 	 *
 	 * @var Commands
 	 */
 	protected $commands;
-
-	//--------------------------------------------------------------------
 
 	/**
 	 * Constructor
@@ -58,8 +56,6 @@ class CommandRunner extends Controller
 		return $this->index($params);
 	}
 
-	//--------------------------------------------------------------------
-
 	/**
 	 * Default command.
 	 *
@@ -70,14 +66,9 @@ class CommandRunner extends Controller
 	 */
 	public function index(array $params)
 	{
-		$command = array_shift($params);
+		$command = array_shift($params) ?? 'list';
 
-		if (is_null($command))
-		{
-			$command = 'list';
-		}
-
-		return service('commands')->run($command, $params);
+		return $this->commands->run($command, $params);
 	}
 
 	/**
--- a/system/CLI/Commands.php
+++ b/system/CLI/Commands.php
@@ -11,14 +11,12 @@
 
 namespace CodeIgniter\CLI;
 
+use CodeIgniter\Autoloader\FileLocator;
 use CodeIgniter\Log\Logger;
-use Config\Services;
 use ReflectionClass;
 use ReflectionException;
 
 /**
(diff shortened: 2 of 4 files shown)


paulbalandan:
@MGatner this optimization made all test cases of `CommandRunnerTest` to be all under 0.50 s.

system/CLI/CommandRunner.php
```diff
@@ -71,13 +67,9 @@ public function _remap($method, ...$params)
 	public function index(array $params)
 	{
 		$command = array_shift($params);
+		$command = $command ?? 'list';
```
MGatner, line 70:
```suggestion
 		$command = array_shift($params) ?? 'list';
```

system/CLI/Commands.php
```diff
@@ -77,39 +74,38 @@ public function run(string $command, array $params)
...
 	public function discoverCommands()
 	{
-		if (! empty($this->commands))
+		if ($this->commands !== [])
 		{
 			return;
 		}
 
-		$files = service('locator')->listFiles('Commands/');
+		$locator = Services::locator();
```
MGatner, line 93: Lonnie wants us to start using the helper function. 
```suggestion
		$locator = service('locator');
```

tests/system/CLI/CommandRunnerTest.php
```diff
@@ -1,66 +1,49 @@
...
-	private $stream_filter;
-	protected $env;
-	protected $config;
-	protected $request;
-	protected $response;
-	protected $logger;
-	protected $runner;
+	public static function setUpBeforeClass(): void
+	{
+		self::$logger = Services::logger();
```
MGatner, line 23:
```suggestion
		self::$logger = service('logger');
```

tests/system/CLI/CommandRunnerTest.php
```diff
@@ -1,66 +1,49 @@
...
+		self::$runner = new CommandRunner();
+
+		self::$runner->initController(Services::request(), Services::response(), self::$logger);
```
MGatner, line 26: Service helper

MGatner (changes requested):
Glad you cracked this egg! Moving the discovery to the constructor could cause some people problems, but generally I think it is a better design. Maybe see if we can get another review from @lonnieezell or @samsonasik on that point? Otherwise just a couple of tweaks.

lonnieezell:
I typically prefer not to do things like that in a constructor but in this case I don't see a problem with it. Commands aren't something that are called all of the time, and testing is where that has the tendency to be the biggest issue. But that's why we're moving it now so it looks good to me!

paulbalandan:
> Glad you cracked this egg! Moving the discovery to the constructor could cause some people problems, but generally I think it is a better design. Maybe see if we can get another review from @lonnieezell or @samsonasik on that point? Otherwise just a couple of tweaks.

I have pondered on this and I don't think moving discovery to the constructor would give us problems. Discovery essentially fills up the $commands array prop with the discovered commands. Current design shows that every invoke of run we call discoverCommands which in turn returns early if the $commands prop is already populated. So this prop is never overwritten. Moving the discovery at constructor initialization would be a big gain, as the tests demonstrate.

samsonasik (approved):

MGatner (approved):

PR #4684, opened 2021-05-13
Update 'updated_at' when enabled in replace()

### LINKED ISSUE #4624: Bug: $model->replace() not generated timestamp
**Describe the bug**
When using $model->replace($data), the timestamp field not generating 

**CodeIgniter 4 version**
Codeigniter 4.1.1

**Context**
 - OS: Debian Linux
 - Web server Nginx
 - PHP version 7.4

paulbalandan:
**Description**
Fixes #4624 
Supersedes and closes #4662 

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1031,6 +1031,10 @@ abstract class BaseModel
             return false;
         }
 
+        if ($this->useTimestamps && $this->updatedField && ! array_key_exists($this->updatedField, (array) $data)) {
+            $data[$this->updatedField] = $this->setDate();
+        }
+
         return $this->doReplace($data, $returnSQL);
     }
 
--- a/system/Database/Postgre/Builder.php
+++ b/system/Database/Postgre/Builder.php
@@ -39,8 +39,6 @@ class Builder extends BaseBuilder
     ];
 
     /**
-     * Compile Ignore Statement
-     *
      * Checks if the ignore option is supported by
      * the Database Driver for the specific statement.
      *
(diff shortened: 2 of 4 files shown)


MGatner:
What's happening with this? Why are we all failing at this issue... 🤦‍♂️

paulbalandan:
The Postgre builder actually has a bug which I discovered that testMode will make replace still execute the SQL. For Sqlsrv, I am not sure where went wrong. I am still to setup sql server locally so I can see this thru xdebug.

MGatner:
😭

MGatner (approved):
I don't understand what happened there but it's working now and I'm happy.

sfadschm (commented):
Is the check for `$data` still needed?

paulbalandan:
> Is the check for `$data` in the `if` statement still needed?

You mean the validation part? I think so because replace is like update, isn't it? Why do you think it's not needed anymore?

system/BaseModel.php
```diff
@@ -1090,12 +1090,19 @@ public function onlyDeleted()
      */
     public function replace(array $data = null, bool $returnSQL = false)
     {
+        $data = $data ?? [];
```
sfadschm, line 1093: Sorry, I messed up the reference on mobile...
I was thinking about this line. I do not really understand why it is needed here.

If I'm correct, the null coalescing operator on `$data` will only reset it to `[]` if `$data` is `null``, as it is typehinted.
In line 1096, we do `if($data && ...  && validate($data))`, both `[]` and `null` would evaluate `false` in the first condition, so we do not need to enure `$data` to be array here.

In the second `if` block, `array_key_exists`  requires an array, so I think we can pull the null coalescence directly before the second `if`. It is not needed in the first `if`.

I hope you can understand what I mean 😆

paulbalandan: Ah yes. It's misplaced. I'll fix that when I'm not on mobile. Thanks!

MGatner (approved):

PR #4688, opened 2021-05-15
Remove Psr\Cache

MGatner:
**Description**
Per the previous discussion, this removes `Psr\Cache` so it can be created as a separate module. I've started work on this module (https://github.com/MGatner/cache) - we will need to decide where it lives. I lean towards https://github.com/codeigniter4 since it is explicitly an add-on component that should be in the same namespace, but I could also see arguments for https://github.com/codeigniter4projects

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide


lonnieezell:
I'm all for the main `codeigniter4` group myself.

MGatner:
Test failure is due to the inevitable code coverage reduction.

PR #4689, opened 2021-05-15
MockCache::getCacheInfo()

MGatner:
**Description**
When testing with `MockCache` using an unknown cache agent it is impossible to determine what happens in the cache without using private property inspectors. This PR expands `MockCache::cacehInfo()` to return the current keys so it is easy to make assertions like "did use cache".

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Test/Mock/MockCache.php
+++ b/system/Test/Mock/MockCache.php
@@ -235,7 +235,7 @@ class MockCache extends BaseHandler implements CacheInterface
 	 * The information returned and the structure of the data
 	 * varies depending on the handler.
 	 *
-	 * @return array Keys currently present in the store
+	 * @return string[] Keys currently present in the store
 	 */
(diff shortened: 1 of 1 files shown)


system/Test/Mock/MockCache.php
```diff
@@ -235,11 +235,11 @@ public function clean()
 	 * The information returned and the structure of the data
 	 * varies depending on the handler.
 	 *
-	 * @return mixed
+	 * @return array Keys currently present in the store
```
paulbalandan, line 238:
```suggestion
	 * @return string[] Keys currently present in the store
```

paulbalandan (approved):

paulbalandan (approved):

PR #4691, opened 2021-05-15
Add URI cast

MGatner:
**Description**
Adds `uri` as a cast option for Entities.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Entity/Cast/URICast.php
+++ b/system/Entity/Cast/URICast.php
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * This file is part of the CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Entity\Cast;
+
+use CodeIgniter\HTTP\URI;
+
+/**
+ * Class URICast
+ */
+class URICast extends BaseCast
+{
+	/**
+	 * @inheritDoc
+	 */
+	public static function get($value, array $params = []): object
+	{
+		return $value instanceof URI ? $value : new URI($value);
+	}
+}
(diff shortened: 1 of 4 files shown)


system/Entity/Cast/URICast.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Class URICast
+ */
+class URICast extends BaseCast
+{
+	/**
+	 * @inheritDoc
+	 */
+	public static function get($value, array $params = []): object
```
samsonasik, line 24:
```suggestion
	public static function get($value, array $params = []): URI
```

MGatner: Thanks 🙏

user_guide_src/source/models/entities.rst
```diff
@@ -330,7 +330,7 @@ You can specify that properties in your Entity should be converted to common dat
 This option should be an array where the key is the name of the class property, and the value is the data type it
 should be cast to. Casting only affects when values are read. No conversions happen that affect the permanent value in
 either the entity or the database. Properties can be cast to any of the following data types:
-**integer**, **float**, **double**, **string**, **boolean**, **object**, **array**, **datetime**, and **timestamp**.
+**integer**, **float**, **double**, **string**, **boolean**, **object**, **array**, **datetime**, **timestamp**, and **URI**.
```
paulbalandan, line 333: Should this one be lowercased `uri` or it does not matter?

MGatner: I read these as the types they are cast into, not the value to use in `$casts`. But maybe it should be the latter, since we don't seem to have that anywhere else?

MGatner: Yeah seeing datetime lowercase I believe you are right.

MGatner, line 333:
```suggestion
**integer**, **float**, **double**, **string**, **boolean**, **object**, **array**, **datetime**, **timestamp**, and **uri**.
```

paulbalandan (approved):

PR #4695, opened 2021-05-17
Some fixes to the Pager Class

najdanovicivan:
**Description**
- Use view() method for rendering which can easily be overridden - I have an override in common.php to load the appropriate site template which is configurable by the user. The issue with pager is that is uses rendered directly instead of just loading the view and it makes it very hard to do the override

- Added setter for total pages to be used with NoSQL databases like ElasticSearch can return total count in the same query which is used for getting the data from the DB avoiding the need for another query to get count. This method makes it possible to set the total counts after initialing the pager and obtaining data results.

- Added public getter for RawCurrentPage. I have a need to show 404 error if greater page than total pages is requested. With current methods the pager will always return the last available page instead of actually requested page. This methods makes it possible to obtain the raw page value from the URL segment/ query string parameter 

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/system/Pager/Pager.php
+++ b/system/Pager/Pager.php
@@ -67,10 +67,10 @@ class Pager implements PagerInterface
 	/**
 	 * Constructor.
 	 *
-	 * @param PagerConfig       $config
-	 * @param RendererInterface $view
+	 * @param PagerConfig            $config
+	 * @param RendererInterface|null $view
 	 */
-	public function __construct(PagerConfig $config, RendererInterface $view)
+	public function __construct(PagerConfig $config, ?RendererInterface $view = null)
 	{
 		$this->config = $config;
 		$this->view   = $view;
@@ -152,6 +152,12 @@ class Pager implements PagerInterface
 		if (! array_key_exists($template, $this->config->templates))
 		{
 			throw PagerException::forInvalidTemplate($template);
+		}
+
+		if (isset($this->view))
+		{
+			return $this->view->setVar('pager', $pager)
+				->render($this->config->templates[$template]);
 		}
 
 		return view($this->config->templates[$template], ['pager' => $pager]);
--- a/system/Pager/Pager.php
+++ b/system/Pager/Pager.php
@@ -154,8 +154,7 @@ class Pager implements PagerInterface
 			throw PagerException::forInvalidTemplate($template);
 		}
 
-		return $this->view->setVar('pager', $pager)
-						->render($this->config->templates[$template]);
+		return view($this->config->templates[$template], ['pager' => $pager]);
 	}
 
 	//--------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


system/Pager/Pager.php
```diff
@@ -154,8 +154,7 @@ protected function displayLinks(string $group, string $template): string
 			throw PagerException::forInvalidTemplate($template);
 		}
 
-		return $this->view->setVar('pager', $pager)
-						->render($this->config->templates[$template]);
+		return view($this->config->templates[$template], ['pager' => $pager]);
```
MGatner, line 157: Why remove the call to the local instance? I think this needs to use the injected renderer.

MGatner (changes requested):

system/Pager/Pager.php
najdanovicivan: I've tested the code using the view() method and it works properly. The issue with local rendered is that you cannot that easily override the path of the view. 

What I'm doing to load different template is that I prepend the template namespace to the path for example

'default_full'   => 'CodeIgniter\Pager\Views\default_full',

I end up loading the view

\Templates\{ConfiguredTemplateName}\CodeIgniter\Pager\Views\default_full 

And I'm doing the prepend inside the view() method. 

I don't see why we need to use the injected view renderer. 

view is passed as AppServices::renderer() in the pager service;
common view() method will call $renderer = Services::renderer(); 

The issue here is that in order to modify the view path for pager with the current setup it is needed to either create a custom rendered class or override the display method with custom pager class. Having it call view() method makes it a lot easier to achieve this as you'll need to override a single method only. IMO RendererInterface should be completely removed from pager and have it rely on the view() method

MGatner: There's nothing magic about `view()`, it just loads and configures the Renderer Service then calls `setData()`: https://github.com/codeigniter4/CodeIgniter4/blob/4ac34c323cd6f074719a0ac713fb31038dec6448/system/Common.php#L1234-L1250

We already have an instance of `RenderInterface` injected so we need to use that. There is no guarantee that this is the same Renderer that will be accessed by `view()` so making this switch introduces a potential breaking change. I am not disagreeing with your point that the dependency *should not* be this way - just that it is and we're stuck with it for now regardless.

najdanovicivan:
@MGatner I've pushed another change. This will make the view option parameter and will be used only if it's set. If it's not passed the view() function will be called.

This is the attempt to avoid the BC while still making it use view() method by default

MGatner:
I'm afraid this won't work either because anyone using their own `Pager` will no longer be guaranteed `$view`. 

I feel fairly certain there is another way for you to accomplish what you need. Since (by default) `Pager::$view` is the same instance as `service('renderer')`, why not apply any changes to it prior to calling the Pager method?
```
service('renderer')->doSomething();
echo $pager->links();
```

paulbalandan:
Closing this for now. Please open a new PR that is non-BC, based on discussion.

PR #4696, opened 2021-05-17
Fix nullable type not showing in SQL string

### LINKED ISSUE #4693: Bug: NULL Fields definition not working for MSQLi Forge in Migration due property $_null and $null names difference
**Describe the bug**

Using Migrations, when you define NULL fields, they are not created as NULLable in MariaDB database.

Table definition with 3 fields set up as NULLable
image

Look at the result table in database, where the 3 fields do not accept NULL 
image

Looking at Database\Forge.php I noticed there is a differente between a protected property $_null in MySQLi\Forge and $null in Database\Forge 

Here it is possible to see when the decision depends on the $null property wich is empty
image

Unfortunately in MySQLi\Forge the property is defined as $_null so it is ignored by Database\Forge
image

**Suggested Solution**
 
As far as I could work on it,  it is a simply change $_null to $null in MySQLi\Forge line 92 solves the problem

After the change NULLable filds passed to work as expected.

image

It still need some additional tests to understand why $_null was written like that, as well as search for any colateral behavior.

**CodeIgniter 4 version**
Master Branch

**Affected module(s)**
vendor\codeigniter4\framework\system\Database\MySQLi\Forge.php

paulbalandan:
**Description**
Fixes #4693 .

It appears the Base Forge provides use of `$null` property but all DB extensions uses the underscored version `$_null`. Then the Base Forge uses `$this->null` so I don't know which to follow. For now, I followed the extensions' work.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -150,6 +150,8 @@ class Forge
 	 * NULL value representation in CREATE/ALTER TABLE statements
 	 *
 	 * @var string
+	 *
+	 * @internal Used for marking nullable fields. Not covered by BC promise.
 	 */
 	protected $null = '';
 
--- a/system/Database/MySQLi/Forge.php
+++ b/system/Database/MySQLi/Forge.php
@@ -88,8 +88,10 @@ class Forge extends BaseForge
 	 * NULL value representation in CREATE/ALTER TABLE statements
 	 *
 	 * @var string
+	 *
+	 * @internal
 	 */
-	protected $_null = 'NULL';
+	protected $null = 'NULL';
 
 	//--------------------------------------------------------------------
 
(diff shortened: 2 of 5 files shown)


system/Database/Forge.php
```diff
@@ -151,7 +151,7 @@ class Forge
 	 *
 	 * @var string
 	 */
-	protected $null = '';
+	protected $_null = '';
```
paulbalandan, line 154: I'm torn here if I should deprecate this, or rename since this is the actual prop name used by extensions.

michalsn: It's more work, but I think we should follow the naming convention from the "base" Forge class. Besides, from what I remember we're trying to eliminate underscores from the variable names, so it would be a step forward as well. 

This is an internal variable, so I don't think there will be any side effects.

We should simply treat this as a bug - no need for depreciation.

MGatner: Let's go ahead and mark this `@internal` if we're going to treat it that way.

michalsn (approved):

PR #4703, opened 2021-05-18
Create .git/hooks directory if not already present

jeromegamez:
Greetings 👋 ,

the Composer `post-update-cmd` failed on my machine because the `admin/setup.sh` script tried to copy the pre-commit hook into `.git/hooks`, which didn't exist yet.

:octocat:


MGatner (approved):
Thanks for your first contribution!

PR #4719, opened 2021-05-21
Try...Suppress

MGatner:
**Description**
The `@` suppression was preventing errors from being thrown & caught in the `try...catch` block, causing `delete_files()` to return `true` even when it failed. The "failure" test was still passing because `RecursiveIteratorIterator` still failed on the non-existent directory.

@samsonasik Is there a Rector rule for `@` suppression in `try...catch`? Seems like it should be an easy one.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Helpers/filesystem_helper.php
+++ b/system/Helpers/filesystem_helper.php
@@ -81,9 +81,9 @@ if (! function_exists('directory_mirror'))
 	 * Recursively copies the files and directories of the origin directory
 	 * into the target directory, i.e. "mirror" its contents.
 	 *
-	 * @param string $originDir
-	 * @param string $targetDir
-	 * @param bool $overwrite   Whether individual files overwrite on collision
+	 * @param string  $originDir
+	 * @param string  $targetDir
+	 * @param boolean $overwrite Whether individual files overwrite on collision
 	 *
 	 * @return void
 	 *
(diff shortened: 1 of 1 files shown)


samsonasik:
Do you mean a rector rule that can remove `@` error suppress? It seems not yet. 

There is `ErrorSuppress` node for that that we can use on create custom rector rule for it, like this:

Screen Shot 2021-05-22 at 09 00 27

@MGatner do you want me to create PR for it to add custom rector rule to https://github.com/codeigniter4/CodeIgniter4/tree/develop/utils/Rector for it?

MGatner:
In my opinion, there should never be an error suppression (e.g. `@unlink($file)`) inside a `try {}` block, because it prevents the error from being thrown and caught. I assume any instance of these would be a logic error. 

If that's easy to write it would be a nice addition, but don't spend a lot of time on it. I thought there might be something already since this seems like perhaps not an uncommon mistake.

system/Helpers/filesystem_helper.php
```diff
@@ -103,7 +103,7 @@ function directory_mirror(string $originDir, string $targetDir, bool $overwrite
 
 		$dirLen = strlen($originDir);
 
-		/** @var SplFileInfo $file */
+		// @var SplFileInfo $file
```
paulbalandan, line 106: I don't think it is correct to change to simple comment. IDEs will not be able to parse the docblock

paulbalandan:
I agree. The `@` operator should not be present inside a try catch as that would be contradicting.

As there is an `ErrorSuppress` node, maybe there's an equivalent `Node` object for try catch in the php-parser library. We should check the presence of both node types before applying the rector rule.

samsonasik:
Ok, so it should be RemoveErrorSuppressInsideTryRector

samsonasik:
@MGatner @paulbalandan here the PR https://github.com/codeigniter4/CodeIgniter4/pull/4724

PR #4721, opened 2021-05-21
Test with MySQL 8

jeromegamez:
Greetings 👋 ,

I'm running PHP 8 and MySQL 8 on my machine and noticed one failing test (`ForgeTest::testAddFields()`, it seems that the `max_length` cannot be retrieved for `BIGINT` fields).

https://github.com/codeigniter4/CodeIgniter4/blob/6da0e5bdc6213edc40ef040de5cbfdcc45a03ead/tests/system/Database/Live/ForgeTest.php#L617-L623

Here's the failing test on my fork: https://github.com/jeromegamez/CodeIgniter4/runs/2643301715?check_suite_focus=true

In order to be able to show the error, I started by adding it to the test matrix (but am not sure how to change the workflow name to reflect it, since all matrix combinations start all types of databases).

<s>The added `mysql8` branch in the triggers is only so that I could run the action, this would of course be removed before merging, if you wish to proceed, that is 😅 </s>

I don't know CodeIgniter at all and am only using this repo, so, if you have an idea why this test might fail, I'm all ears👂, otherwise, I will try to dive deeper into it in the hopes to find a good explanation or even a fix 😅

:octocat:

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -138,7 +138,7 @@ jobs:
           DB: ${{ matrix.db-platforms }}
           TERM: xterm-256color
 
-      - if: matrix.php-versions == '7.4'
+      - if: github.repository_owner == 'codeigniter4' && matrix.php-versions == '7.4'
         name: Run Coveralls
         run: |
           composer global require php-coveralls/php-coveralls:^2.4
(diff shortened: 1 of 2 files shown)


paulbalandan:
According to the MySQL documentation, this feature is already deprecated as of v8.0.17.

> As of MySQL 8.0.17, the display width attribute is deprecated for integer data types; you should expect support for it to be removed in a future version of MySQL.

jeromegamez:
Thanks for the research, @paulbalandan 🤗! I added a commit so that the `max_length` check is skipped when tested with MySQL >= 8.0.17 

(The formatting change in the `testCreateTableWithArrayFieldConstraints()` method was automatically applied by the project's pre-commit hook)

MGatner (approved):
Nice sleuthing! I can see I need to spend some more time on MySQL 8 😯

paulbalandan (approved):

PR #4726, opened 2021-05-23
Fix truncated SCRIPT_NAME

MGatner:
**Description**
It appears that some PHP web servers do not guarantee the `SCRIPT_NAME` as a substring of `REQUEST_URI`. From consulting other frameworks, it is still safe to determine the relative path this way, but we need to ensure the segments before checking.

**Checklist:**
- [X] Securely signed commits
- n/a User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -270,7 +270,7 @@ class IncomingRequest extends Request
 			foreach (explode('/', $_SERVER['SCRIPT_NAME']) as $i => $segment)
 			{
 				// If these segments are not the same then we're done
-				if ($segment !== $segments[$i])
+				if (! isset($segments[$i]) || $segment !== $segments[$i])
 				{
 					break;
 				}


system/Language/en/Publisher.php
```diff
@@ -0,0 +1,17 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+// Publisher language settings
+return [
+	'expectedFile'      => 'Publisher::{0} expects a valid file.',
+	'expectedDirectory' => 'Publisher::{0} expects a valid directory.',
+	'collision'         => 'Publisher encountered an unexpected {0} while copying {1} to {2}.',
+];
```
samsonasik, line 17: @MGatner it seems the new files are unrelated to the PR?

MGatner: Wow yeah sorry about that! I think I forgot to `-u` my stash so the new files carried over. I will force back to the correct file.

samsonasik (approved):

paulbalandan (approved):

PR #4727, opened 2021-05-24
Publisher Class

MGatner:
**Description**
This PR introduces a new component: the Publisher Class. This combines a handful of different solutions I've been using in my modules to solve a variety of problems:
1. How do I maintain project assets with version dependencies?
2. How do I manage uploads and other "dynamic" files that need to be web accessible?
3. How can I update my project when the framework or modules change?
4. How can the framework/modules publish their modified content to a project?

`Publisher` has two functional modes: it can be instantiated and used dynamically with runtime input, or can use predefined workflows staged for automatic discovery. With either approach developers use fluent-style methods to build a payload manifest and then one of the "write" commands to deliver the files. For example, to inject the latest Bootstrap minimized JS & CSS files:
```
$publisher = new Publisher();
$publisher
	->addPath('vendor/twbs/bootstrap/dist')
	->retainPattern('*.min.*)
	->merge();
```

By default `Publisher` assumes the source is `ROOTPATH` and the destination is `FCPATH` but this can be changed with the constructor:
```
$publisher = new Publisher('vendor', FCPATH . 'assets/vendor/');
```
... or set directly on extensions:
```
class BootstrapPublisher extends Publisher
{
	protected $source = ROOTPATH . 'vendor/twbs/bootstrap';
	protected $destination = FCPATH . 'bootstrap/';
}
```

I will have lots more explanations and examples in the User Guide but I wanted to get this PR up for reviews since it has been a while since we added an entirely new component.

*Note: This will need to be re-styled after the coding standards release.*

*A few motivations for this library:*
* https://github.com/tattersoftware/codeigniter4-assets#publishing
* https://github.com/tattersoftware/codeigniter4-patches#description
* https://github.com/lonnieezell/myth-auth/blob/develop/docs/commands.md#publish
* https://github.com/paulbalandan/liaison-revision
* https://github.com/codeigniter4/CodeIgniter4/pull/4634

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/Publisher.php
+++ b/app/Config/Publisher.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Config;
+
+use CodeIgniter\Config\Publisher as BasePublisher;
+
+/**
+ * Publisher Configuration
+ *
+ * Defines basic security restrictions for the Publisher class
+ * to prevent abuse by injecting malicious files into a project.
+ */
+class Publisher extends BasePublisher
+{
+	/**
+	 * A list of allowed destinations with a (pseudo-)regex
+	 * of allowed files for each destination.
+	 * Attempts to publish to directories not in this list will
+	 * result in a PublisherException. Files that do no fit the
+	 * pattern will cause copy/merge to fail.
+	 *
+	 * @var array<string,string>
+	 */
+	public $restrictions = [
+		ROOTPATH => '*',
+		FCPATH   => '#\.(?css|js|map|htm?|xml|json|webmanifest|tff|eot|woff?|gif|jpe?g|tiff?|png|webp|bmp|ico|svg)$#i',
+	];
+}
--- a/system/Publisher/Exceptions/PublisherException.php
+++ b/system/Publisher/Exceptions/PublisherException.php
@@ -13,18 +13,39 @@ namespace CodeIgniter\Publisher\Exceptions;
 
 use CodeIgniter\Exceptions\FrameworkException;
 
+/**
+ * Publisher Exception Class
+ *
+ * Handles exceptions related to actions taken by a Publisher.
+ */
 class PublisherException extends FrameworkException
 {
+	/**
+	 * Throws when a file should be overwritten yet cannot.
+	 *
+	 * @param string $from The source file
+	 * @param string $to   The destination file
+	 */
 	public static function forCollision(string $from, string $to)
 	{
 		return new static(lang('Publisher.collision', [filetype($to), $from, $to]));
 	}
 
+	/**
+	 * Throws when an object is expected to be a directory but is not or is missing.
+	 *
+	 * @param string $caller The method causing the exception
+	 */
 	public static function forExpectedDirectory(string $caller)
 	{
 		return new static(lang('Publisher.expectedDirectory', [$caller]));
 	}
 
+	/**
+	 * Throws when an object is expected to be a file but is not or is missing.
+	 *
+	 * @param string $caller The method causing the exception
+	 */
 	public static function forExpectedFile(string $caller)
 	{
 		return new static(lang('Publisher.expectedFile', [$caller]));
--- a/system/Publisher/Exceptions/PublisherException.php
+++ b/system/Publisher/Exceptions/PublisherException.php
@@ -50,4 +50,26 @@ class PublisherException extends FrameworkException
 	{
 		return new static(lang('Publisher.expectedFile', [$caller]));
 	}
+
+	/**
+	 * Throws when given a destination that is not in the list of allowed directories.
+	 *
+	 * @param string $destination
+	 */
+	public static function forDestinationNotAllowed(string $destination)
+	{
+		return new static(lang('Publisher.destinationNotAllowed', [$destination]));
+	}
+
+	/**
+	 * Throws when a file fails to match the allowed pattern for its destination.
+	 *
+	 * @param string $file
+	 * @param string $directory
+	 * @param string $pattern
+	 */
+	public static function forFileNotAllowed(string $file, string $directory, string $pattern)
+	{
+		return new static(lang('Publisher.fileNotAllowed', [$file, $directory, $pattern]));
+	}
 }
--- a/system/Publisher/Exceptions/PublisherException.php
+++ b/system/Publisher/Exceptions/PublisherException.php
@@ -15,6 +15,11 @@ use CodeIgniter\Exceptions\FrameworkException;
 
 class PublisherException extends FrameworkException
 {
+	public static function forCollision(string $from, string $to)
+	{
+		return new static(lang('Publisher.collision', [filetype($to), $from, $to]));
+	}
+
 	public static function forExpectedDirectory(string $caller)
 	{
 		return new static(lang('Publisher.expectedDirectory', [$caller]));
@@ -23,10 +28,5 @@ class PublisherException extends FrameworkException
 	public static function forExpectedFile(string $caller)
 	{
 		return new static(lang('Publisher.expectedFile', [$caller]));
-	}
-
-	public static function forCollision(string $from, string $to)
-	{
-		return new static(lang('Publisher.collision', [filetype($to), $from, $to]));
 	}
 }
--- a/system/Publisher/Exceptions/PublisherException.php
+++ b/system/Publisher/Exceptions/PublisherException.php
@@ -1,0 +1,32 @@
+<?php
+
+/**
+ * This file is part of the CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Publisher\Exceptions;
+
+use CodeIgniter\Exceptions\FrameworkException;
+
+class PublisherException extends FrameworkException
+{
+	public static function forExpectedDirectory(string $caller)
+	{
+		return new static(lang('Publisher.expectedDirectory', [$caller]));
+	}
+
+	public static function forExpectedFile(string $caller)
+	{
+		return new static(lang('Publisher.expectedFile', [$caller]));
+	}
+
+	public static function forCollision(string $from, string $to)
+	{
+		return new static(lang('Publisher.collision', [filetype($to), $from, $to]));
+	}
+}
--- a/system/Publisher/Publisher.php
+++ b/system/Publisher/Publisher.php
@@ -108,8 +108,8 @@ class Publisher
 			return [];
 		}
 
-		// Loop over each file checking to see if it is a Primer
-		foreach ($files as $file)
+		// Loop over each file checking to see if it is a Publisher
+		foreach (array_unique($files) as $file)
 		{
 			$className = $locator->findQualifiedNameFromPath($file);
 
@@ -228,8 +228,10 @@ class Publisher
 			$attempts = 10;
 			while ((bool) $attempts && ! delete_files($directory, true, false, true))
 			{
+				// @codeCoverageIgnoreStart
 				$attempts--;
 				usleep(100000); // .1s
+				// @codeCoverageIgnoreEnd
 			}
 
 			@rmdir($directory);
@@ -268,6 +270,13 @@ class Publisher
 
 			// Try to remove anything else
 			unlink($to);
+		}
+
+		// Make sure the directory exists
+		$directory = pathinfo($to, PATHINFO_DIRNAME);
+		if (! is_dir($directory))
+		{
+			mkdir($directory, 0775, true);
 		}
 
 		// Allow copy() to throw errors
@@ -311,10 +320,11 @@ class Publisher
 	 * This method should be reimplemented by
 	 * child classes intended for discovery.
 	 *
-	 * @return void
+	 * @return bool
 	 */
 	public function publish()
 	{
+		return $this->addPath('/')->merge(true);
 	}
 
 	//--------------------------------------------------------------------
@@ -686,7 +696,7 @@ class Publisher
 		$this->errors = [];
 
 		// Get the file from source for special handling
-		$sourced = self::matchFiles($this->getFiles(), $this->source);
+		$sourced = self::filterFiles($this->getFiles(), $this->source);
 
 		// Handle everything else with a flat copy
 		$this->files = array_diff($this->files, $sourced);
--- a/system/Publisher/Publisher.php
+++ b/system/Publisher/Publisher.php
@@ -73,6 +73,14 @@ class Publisher
 	private $published = [];
 
 	/**
+	 * List of allowed directories and their allowed files regex.
+	 * Restrictions are intentionally private to prevent overriding.
+	 *
+	 * @var array<string,string>
+	 */
+	private $restrictions;
+
+	/**
 	 * Base path to use for the source.
 	 *
 	 * @var string
@@ -134,6 +142,8 @@ class Publisher
 	 * @param string $directory
 	 *
 	 * @return string
+	 *
+	 * @throws PublisherException
 	 */
 	private static function resolveDirectory(string $directory): string
 	{
@@ -152,6 +162,8 @@ class Publisher
 	 * @param string $file
 	 *
 	 * @return string
+	 *
+	 * @throws PublisherException
 	 */
 	private static function resolveFile(string $file): string
 	{
@@ -191,7 +203,7 @@ class Publisher
 	 *
 	 * @return string[]
 	 */
-	private static function matchFiles(array $files, string $pattern)
+	private static function matchFiles(array $files, string $pattern): array
 	{
 		// Convert pseudo-regex into their true form
 		if (@preg_match($pattern, null) === false) // @phpstan-ignore-line
@@ -236,49 +248,6 @@ class Publisher
 		}
 	}
 
-	/**
-	 * Copies a file with directory creation and identical file awareness.
-	 * Intentionally allows errors.
-	 *
-	 * @param string  $from
-	 * @param string  $to
-	 * @param boolean $replace
-	 *
-	 * @return void
-	 *
-	 * @throws PublisherException For unresolvable collisions
-	 */
-	private static function safeCopyFile(string $from, string $to, bool $replace): void
-	{
-		// Check for an existing file
-		if (file_exists($to))
-		{
-			// If not replacing or if files are identical then consider successful
-			if (! $replace || same_file($from, $to))
-			{
-				return;
-			}
-
-			// If it is a directory then do not try to remove it
-			if (is_dir($to))
-			{
-				throw PublisherException::forCollision($from, $to);
-			}
-
-			// Try to remove anything else
-			unlink($to);
-		}
-
-		// Make sure the directory exists
-		if (! is_dir($directory = pathinfo($to, PATHINFO_DIRNAME)))
-		{
-			mkdir($directory, 0775, true);
-		}
-
-		// Allow copy() to throw errors
-		copy($from, $to);
-	}
-
 	//--------------------------------------------------------------------
 
 	/**
@@ -293,6 +262,20 @@ class Publisher
 
 		$this->source      = self::resolveDirectory($source ?? $this->source);
 		$this->destination = self::resolveDirectory($destination ?? $this->destination);
+
+		// Restrictions are intentionally not injected to prevent overriding
+		$this->restrictions = config('Publisher')->restrictions;
+
+		// Make sure the destination is allowed
+		foreach (array_keys($this->restrictions) as $directory)
+		{
+			if (strpos($this->destination, $directory) === 0)
+			{
+				return;
+			}
+		}
+
+		throw PublisherException::forDestinationNotAllowed($this->destination);
 	}
 
 	/**
@@ -314,6 +297,8 @@ class Publisher
 	 * discovery.
 	 *
 	 * @return boolean
+	 *
+	 * @throws RuntimeException
 	 */
 	public function publish(): bool
 	{
@@ -683,7 +668,7 @@ class Publisher
 
 			try
 			{
-				self::safeCopyFile($file, $to, $replace);
+				$this->safeCopyFile($file, $to, $replace);
 				$this->published[] = $to;
 			}
 			catch (Throwable $e)
@@ -707,7 +692,7 @@ class Publisher
 	{
 		$this->errors = $this->published = [];
 
-		// Get the file from source for special handling
+		// Get the files from source for special handling
 		$sourced = self::filterFiles($this->getFiles(), $this->source);
 
 		// Handle everything else with a flat copy
@@ -722,7 +707,7 @@ class Publisher
 
 			try
 			{
-				self::safeCopyFile($file, $to, $replace);
+				$this->safeCopyFile($file, $to, $replace);
 				$this->published[] = $to;
 			}
 			catch (Throwable $e)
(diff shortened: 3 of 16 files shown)


system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+	 * @return $this
+	 */
+	public function addUri(string $uri)
+	{
+		// Figure out a good filename (using URI strips queries and fragments)
+		$file = $this->getScratch() . basename((new URI($uri))->getPath());
+
+		// Get the content and write it to the scratch space
+		$response = service('curlrequest')->get($uri);
+		write_file($file, $response->getBody());
```
mostafakhudair, line 580:
```suggestion
		write_file($file, service('curlrequest')->get($uri)->getBody());
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+		{
+			$this->addUri($uri);
+		}
+
+		return $this;
+	}
+
+	/**
+	 * Downloads a file from the URI
+	 * and adds it to the file list.
```
mostafakhudair, line 567:
```suggestion
	 * Downloads a file from the URI, and adds it to the file list.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Removes any files from the list that match
+	 * the supplied pattern (within the optional scope).
+	 *
+	 * @param string $pattern    Regex or pseudo-regex string
+	 * @param string|null $scope A directory to limit the scope
```
mostafakhudair, line 592:
```suggestion
	/**
	 * Removes any files from the list that match the supplied pattern
	 * (within the optional scope).
	 *
	 * @param string      $pattern Regex or pseudo-regex string
	 * @param string|null $scope A directory to limit the scope
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+		}
+
+		return $this->addDirectory($full, $recursive);
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Downloads and stages files from
+	 * an array of URIs.
```
mostafakhudair, line 549:
```suggestion
	/**
	 * Downloads and stages files from an array of URIs.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+		}
+
+		return $this;
+	}
+
+	/**
+	 * Adds a single path to the file list.
+	 *
+	 * @param string $path
+	 * @param bool $recursive
```
mostafakhudair, line 524:
```suggestion
	 * @param string  $path
	 * @param boolean $recursive
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+		return $this;
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Verifies and adds paths to the list.
+	 *
+	 * @param string[] $paths
+	 * @param bool $recursive
```
mostafakhudair, line 506:
```suggestion
	/**
	 * Verifies and adds paths to the list.
	 *
	 * @param string[] $paths
	 * @param boolean  $recursive
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+
+		return $this;
+	}
+
+	/**
+	 * Verifies and adds all files
+	 * from a directory.
+	 *
+	 * @param string $directory
+	 * @param bool $recursive
```
mostafakhudair, line 476:
```suggestion
	 * Verifies and adds all files from a directory.
	 *
	 * @param string  $directory
	 * @param boolean $recursive
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Verifies and adds files from each
+	 * directory to the list.
+	 *
+	 * @param string[] $directories
+	 * @param bool $recursive
```
mostafakhudair, line 457:
```suggestion
	/**
	 * Verifies and adds files from each directory to the list.
	 *
	 * @param string[] $directories
	 * @param boolean  $recursive
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+		{
+			$this->addFile($file);
+		}
+
+		return $this;
+	}
+
+	/**
+	 * Verifies and adds a single file
+	 * to the file list.
```
mostafakhudair, line 411:
```suggestion
	 * Verifies and adds a single file to the file list.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+
+		return $this->files;
+	}
+
+	/**
+	 * Sets the file list directly.
+	 * Files are still subject to verification.
+	 * This works as a "reset" method with [].
+	 *
+	 * @param string[] $files  A new file list to use
```
mostafakhudair, line 379:
```suggestion
	 * Sets the file list directly, files are still subject to verification.
	 * This works as a "reset" method with [].
	 *
	 * @param string[] $files The new file list to use
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+	 * @return array<string,Throwable>
+	 */
+	public function getErrors(): array
+	{
+		return $this->errors;
+	}
+
+	/**
+	 * Optimizes and returns the
+	 * current file list.
```
mostafakhudair, line 362:
```suggestion
	 * Optimizes and returns the current file list.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+ * to its source or destination and may indicate
+ * either a file or directory fo unconfirmed
+ * existence.
+ * Class failures throw a PublisherException,
+ * but some underlying methods may percolate
+ * different exceptions, like FileException,
+ * FileNotFoundException, InvalidArgumentException.
+ * Write operations will catch all errors in the
+ * file-specific $errors property to minimize
+ * impact of partial batch operations.
```
mostafakhudair, line 40:
```suggestion
 * Publishers read in file paths from a variety of sources and copy
 * the files out to different destinations. This class acts both as
 * a base for individual publication directives as well as the mode
 * of discovery for said instances. In this class a "file" is a full
 * path to a verified file while a "path" is relative to its source
 * or destination and may indicate either a file or directory of
 * unconfirmed existence.
 * class failures throw the PublisherException, but some underlying
 * methods may percolate different exceptions, like FileException,
 * FileNotFoundException or InvalidArgumentException.
 * Write operations will catch all errors in the file-specific
 * $errors property to minimize impact of partial batch operations.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+		sort(self::$discovered[$directory]);
+
+		return self::$discovered[$directory];
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Resolves a full path and verifies
+	 * it is an actual directory.
```
mostafakhudair, line 130:
```suggestion
	/**
	 * Resolves a full path and verifies it is an actual directory.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+			$caller = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1];
+			throw PublisherException::forExpectedDirectory($caller['function']);
+		}
+
+		return $directory;
+	}
+
+	/**
+	 * Resolves a full path and verifies
+	 * it is an actual file.
```
mostafakhudair, line 149:
```suggestion
	 * Resolves a full path and verifies it is an actual file.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+
+	/**
+	 * Removes a directory and all its files
+	 * and subdirectories.
+	 *
+	 * @param string $directory
+	 *
+	 * @return void
+	 */
+	private static function wipeDirectory(string $directory)
```
mostafakhudair, line 223:
```suggestion
	 * Removes a directory and all its files and subdirectories.
	 *
	 * @param string $directory
	 *
	 * @return void
	 */
	private static function wipeDirectory(string $directory): void
```

mostafakhudair (changes requested):

system/Publisher/Exceptions/PublisherException.php
```diff
@@ -0,0 +1,32 @@
...
+
+	public static function forExpectedFile(string $caller)
+	{
+		return new static(lang('Publisher.expectedFile', [$caller]));
+	}
+
+	public static function forCollision(string $from, string $to)
+	{
+		return new static(lang('Publisher.collision', [filetype($to), $from, $to]));
+	}
```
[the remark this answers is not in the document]
MGatner: We don't usually docblock our exception static methods. I don't know why, but I was following that convention!

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+	}
+
+	//--------------------------------------------------------------------
+
+	/**
+	 * Filters an array of files, removing files not
+	 * part of the given directory (recursive).
+	 *
+	 * @param string[] $files
+	 * @param string $directory
```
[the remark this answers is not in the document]
MGatner: I like the change in wording but I want to keep the separators. They offer good visual distinctions between sections for large files like this.

system/Publisher/Publisher.php
MGatner: Separator

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,723 @@
...
+			: self::filterFiles($this->files, $scope);
+
+		// Match the pattern within the scoped files
+		$matched = self::matchFiles($files, $pattern);
+
+		// ... and remove their inverse
+		return $this->removeFiles(array_diff($files, $matched));
+	}
+
+	//--------------------------------------------------------------------
```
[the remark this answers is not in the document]
MGatner: Separator

MGatner:
Thanks @mostafakhudair, great clean up!

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,704 @@
...
+	 * Array of discovered Publishers.
+	 *
+	 * @var array<string,self[]|null>
+	 */
+	private static $discovered = [];
+
+	/**
+	 * Directory to use for methods
+	 * that need temporary storage.
+	 * Created on-the-fly as needed.
```
mostafakhudair, line 48:
```suggestion
	 * Directory to use for methods that need temporary storage.
	 * Created on-the-fly as needed.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,704 @@
...
+	/**
+	 * The current list of files.
+	 *
+	 * @var string[]
+	 */
+	private $files = [];
+
+	/**
+	 * Exceptions for specific files
+	 * from the last write operation.
```
mostafakhudair, line 63:
```suggestion
	 * Exceptions for specific files from the last write operation.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,704 @@
...
+	/**
+	 * Base path to use for the destination.
+	 *
+	 * @var string
+	 */
+	protected $destination = FCPATH;
+
+	/**
+	 * Discovers and returns all Publishers
+	 * in the specified namespace directory.
```
mostafakhudair, line 85:
```suggestion
	 * Discovers and returns all Publishers in the specified namespace directory.
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,704 @@
...
+		foreach (array_unique($files) as $file)
+		{
+			$className = $locator->findQualifiedNameFromPath($file);
+
+			if (is_string($className) && class_exists($className) && is_a($className, self::class, true))
+			{
+				self::$discovered[$directory][] = new $className();
+			}
+		}
+		sort(self::$discovered[$directory]);
```
mostafakhudair, line 118:
```suggestion
		}

		sort(self::$discovered[$directory]);
```

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,704 @@
...
+				throw PublisherException::forCollision($from, $to);
+			}
+
+			// Try to remove anything else
+			unlink($to);
+		}
+
+		// Make sure the directory exists
+		$directory = pathinfo($to, PATHINFO_DIRNAME);
+		if (! is_dir($directory))
```
mostafakhudair, line 271:
```suggestion
		if (! is_dir($directory = pathinfo($to, PATHINFO_DIRNAME)))
```

tests/_support/Publishers/TestPublisher.php
```diff
@@ -0,0 +1,16 @@
...
+namespace Tests\Support\Publishers;
+
+use CodeIgniter\Publisher\Publisher;
+
+class TestPublisher extends Publisher
+{
+	/**
+	 * Runs the defined Operations.
+	 */
+	public function publish()
```
mostafakhudair, line 12:
```suggestion
	public function publish(): bool
```

mostafakhudair (changes requested):

paulbalandan (commented):
Can we get some initial docs for this? I can't visualize this in action. 😅

MGatner:
Oh for sure! Lots more coming, just other priorities at the moment. I'll mark this Draft so it isn't confusing.

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,736 @@
...
+		try
+		{
+			$directory = self::resolveDirectory($full);
+		}
+		catch (PublisherException $e)
+		{
+			return $this->addFile($full);
+		}
+
+		return $this->addDirectory($full, $recursive);
```
paulbalandan, line 565: This can be placed inside the `try`

system/Publisher/Publisher.php
```diff
@@ -0,0 +1,736 @@
...
+		$full = $this->source . $path;
+
+		// Test for a directory
+		try
+		{
+			$directory = self::resolveDirectory($full);
+		}
+		catch (PublisherException $e)
+		{
+			return $this->addFile($full);
```
paulbalandan, line 562: Should we log also the error?

tests/_support/Publishers/TestPublisher.php
```diff
@@ -0,0 +1,50 @@
...
+
+use CodeIgniter\Publisher\Publisher;
+use RuntimeException;
+
+final class TestPublisher extends Publisher
+{
+	/**
+	 * Fakes an error on the given file.
+	 *
+	 * @return $this
```
paulbalandan, line 13:
```suggestion
```

user_guide_src/source/libraries/publisher.rst
```diff
@@ -0,0 +1,440 @@
...
+				$publisher->addPath('daily_photo.jpg')->copy($replace = true);
+			}
+			catch (Throwable $e)
+			{
+				$this->showError($e);
+			}
+		}
+	}
+
+Now running ``spark publish:daily`` will keep your homepage's image up-to-date. What if the photo is
```
paulbalandan, line 151: I do not see a `publish:daily` command. Is this still to be made?

paulbalandan (commented):
I think I grasp the concept and process flow. Just some comments.

user_guide_src/source/libraries/publisher.rst
```diff
@@ -0,0 +1,440 @@
...
+		protected $name        = 'publish:daily';
+		protected $description = 'Publishes the latest daily photo to the homepage.';
+
+		public function run(array $params)
+		{
+			$publisher = new Publisher('/path/to/photos/', FCPATH . 'assets/images');
+
+			try
+			{
+				$publisher->addPath('daily_photo.jpg')->copy($replace = true);
```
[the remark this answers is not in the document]
MGatner: I was trying to make it clear what the `true` parameter did. Maybe switch it to a comment?

system/Publisher/Publisher.php
MGatner: We want this to throw if `addDirectory()` or `addFile()` fail. The above `try..catch` is to determine whether the `$path` is really really a directory or not.

tests/_support/Publishers/TestPublisher.php
MGatner: Oops! this used to be an instance method 😅

user_guide_src/source/libraries/publisher.rst
MGatner: It's the example command immediately above 🙃

user_guide_src/source/libraries/publisher.rst
```diff
@@ -0,0 +1,440 @@
...
+	$publisher->addUri('https://example.com/feeds/daily_photo.jpg')->copy($replace = true);
+
+Asset Dependencies Example
+==========================
+
+You want to integrate the frontend library "Bootstrap" into your project, but the frequent updates makes it a hassle
+to keep up with. You can create a publication definition in your project to sync frontend assets by adding extending
+``Publisher`` in your project. So **app/Publishers/BootstrapPublisher.php** might look like this::
+
+	<?php
```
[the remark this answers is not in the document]
MGatner: I'm not sure why. I tweaked the formatting - hopefully that fixed it.

user_guide_src/source/libraries/publisher.rst
```diff
@@ -0,0 +1,442 @@
...
+	$publisher->addUri('https://example.com/feeds/daily_photo.jpg')->copy(true);
+
+Asset Dependencies Example
+==========================
+
+You want to integrate the frontend library "Bootstrap" into your project, but the frequent updates makes it a hassle
+to keep up with. You can create a publication definition in your project to sync frontend assets by adding extending
+``Publisher`` in your project. So **app/Publishers/BootstrapPublisher.php** might look like this::
+
+	namespace App\Publishers;
```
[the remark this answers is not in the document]
MGatner: Can you tell what is wrong with it?

user_guide_src/source/libraries/publisher.rst
```diff
@@ -0,0 +1,442 @@
...
+resource and publish it out instead::
+
+	$publisher->addUri('https://example.com/feeds/daily_photo.jpg')->copy(true);
+
+Asset Dependencies Example
+==========================
+
+You want to integrate the frontend library "Bootstrap" into your project, but the frequent updates makes it a hassle
+to keep up with. You can create a publication definition in your project to sync frontend assets by adding extending
+``Publisher`` in your project. So **app/Publishers/BootstrapPublisher.php** might look like this ::
```
paulbalandan, line 163: Can you try removing the space between `::`
```suggestion
``Publisher`` in your project. So **app/Publishers/BootstrapPublisher.php** might look like this::
```

MGatner: I just added that this last round.

MGatner: I think I might see it, missing a single quote

paulbalandan (approved):

lonnieezell (changes requested):
In general I like this a lot. My biggest concern is from a security standpoint. Maybe I missed it but is there anything preventing a bad package from copying files from elsewhere in the system into the `public` folder? Or similar dastardly deeds? 

I think there has to be some way to limit the available scope of where they can work. In old versions of Bonfire I limited it to anything within the working directory for both reading and writing. 

I have seen instances in other languages where a trusted package that has been around for years is taken over by another maintainer and malevolent code gets inserted which is then potentially added to many projects and developers were completely unaware. Even if they had done their homework and inspected the package nicely to begin with, after trusting it for years, you would have no reason to expect. 

That's why I think we need to minimize the possibility of this at the framework level here.

MGatner:
I had some security initially, especially around what files you could acquire from URIs, but I scrapped it after working on the discovery. We may want to think about some framework-level security or something, because the truth is at this point with our autodetection and execution this would be really easy in dozens of places (like sticking arbitrary code in a module's Config/Routes.php). 

@lonnieezell like the extra attention to public/ though since anything there is web-accessible. What about denying PHP files sourced from vendor/ heading to public/? Or some similar extension-level filtering?

lonnieezell:
Hm. This one's tricky. I can imagine typical scenarios where multiple sites are built on on instance of CI and they may have shared resources, meaning restricting this to just the web root would be problematic for those cases. 

Maybe we make a white list of directories it can operate within? By default it's just the project root and any directories within that folder? Granted that makes public available but it needs to be for your example of copying bootstrap files to `public`. Would it be overkill to have a separate whitelist for readable and writable paths?

MGatner:
> your example of copying bootstrap files

I think this is actually the number one target use of Publisher. That's my guess at least. A lot of this was built on my Assets module whose purpose is to publish JS/CSS to web root and assist loading them. I wouldn't want to hinder that, but putting in restrictions (like filetypes) seems wise.

> Would it be overkill to have a separate whitelist for readable and writable paths?

Actually I think this is a pretty good idea. Since we already have `retainPattern()` and `removePattern()` this could be very easy to implement. I'll take a shot and ask for another review.

MGatner:
@lonnieezell I went through a few iterations of this. It's a bit of a tricky business because code executed in an auto-discovered class already technically has access to do anything it wants. I decided to focus on "what ends up where": https://github.com/codeigniter4/CodeIgniter4/pull/4727/commits/f33d8809fa96de9fcdcd30042fc4435410abb799

This introduces `restrictions`, which is a way of controlling the allowed destinations and types of files that can go to each destination. I've added some additional measures to make sure child classes of `Publisher` cannot tamper with their own restrictions to make it even more unlikely that someone's **var/mail** folder ends up copied into their web-accessible root.

app/Config/Publisher.php
```diff
@@ -0,0 +1,28 @@
...
+	/**
+	 * A list of allowed destinations with a (pseudo-)regex
+	 * of allowed files for each destination.
+	 * Attempts to publish to directories not in this list will
+	 * result in a PublisherException. Files that do no fit the
+	 * pattern will cause copy/merge to fail.
+	 *
+	 * @var array<string,string>
+	 */
+	public $restrictions = [
```
paulbalandan, line 24: How about not default exposing this to userland? Since you have set `Publisher::$restrictions` as private, I think changing should be available only to child Publishers.

lonnieezell (approved):

mostafakhudair (commented):
With PSR-12, I think this should be reverted to bool.

mostafakhudair:
Yeah also brackets, my fault, I didn't know that it will be auto fix, review deleted, thanks for let me know <3

MGatner:
No worries! I appreciate your attention to detail. I noted it in the description, but it was a later edit so you probably never saw it:
> Note: This will need to be re-styled after the coding standards release.

app/Config/Publisher.php
MGatner: I thought of making this `protected` with a `__get()` method but ultimately I decided it did not matter. Restrictions are stored during Publisher construction, so an errant module Publisher would not be able to modify the Config value prior to initialization. Any other instances where a module has access to the Config prior to discovery it can already do whatever it wanted anyway.

paulbalandan: Hmm. Fair enough.

paulbalandan (approved):

app/Config/Publisher.php
MGatner: I definitely get the suggestion. I went through a bunch of iterations for how to handle security, and basically decided that trying to limit arbitrary module code is useless (without larger framework buy-in) so focused on "what can we do to prevent automated publications from running malicious code?" I believe the current implementation addresses that, with the understanding that there are numerous holes elsewhere in the framework.

MGatner:
I have a plan to refactor this a bit but the current PR has gotten rather cumbersome so I am going to merge it and then open a new one.

PR #4729, opened 2021-05-24
Replace `isset()` with the `??` null coalesce operator

jeromegamez:
Since the minimum PHP version is PHP 7.3, we can make use of the language features it provides 🥳, in this case:

```php
$x = $y ?? null
```
instead of
```php
$x = isset($y) ? $y : null
``` 

:octocat:

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -22,6 +22,7 @@ use Rector\EarlyReturn\Rector\Foreach_\ChangeNestedForeachIfsToEarlyContinueRect
 use Rector\EarlyReturn\Rector\If_\ChangeIfElseValueAssignToEarlyReturnRector;
 use Rector\EarlyReturn\Rector\If_\RemoveAlwaysElseRector;
 use Rector\EarlyReturn\Rector\Return_\PreparedValueToEarlyReturnRector;
+use Rector\Php70\Rector\Ternary\TernaryToNullCoalescingRector;
 use Rector\Php73\Rector\FuncCall\JsonThrowOnErrorRector;
 use Rector\Php73\Rector\FuncCall\StringifyStrNeedlesRector;
 use Rector\Set\ValueObject\SetList;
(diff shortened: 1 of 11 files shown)


paulbalandan (changes requested):
Files under system/ThirdParty should not be touched, as these are outside dependencies.

jeromegamez:
Ah, I didn't know, thanks for the heads up! The changes in system/ThirdParty have been reverted 💪

paulbalandan (approved):

jeromegamez:
@paulbalandan Speaking of the ThirdParty dependencies… I'm not familiar with the reasons, but assuming this was needed in older CI versions, do you think it would be possible to replace the copied dependencies with "native" composer dependencies? The `psr/log`, for example, is already present in the `require` section of the composer.json 🤔

paulbalandan:
> @paulbalandan Speaking of the ThirdParty dependencies… I'm not familiar with the reasons, but assuming this was needed in older CI versions, do you think it would be possible to replace the copied dependencies with "native" composer dependencies? The `psr/log`, for example, is already present in the `require` section of the composer.json 🤔

The ThirdParty dependencies are there to support non-Composer users, particularly those who prefer directly downloading the archive file from the Releases section or from the website. We have the `ComposerScripts` which regularly checks for these deps for updates and copies them over.

jeromegamez:
Understood, thanks!

samsonasik:
@jeromegamez please add `Rector\Php70\Rector\Ternary\TernaryToNullCoalescingRector` to `rector.php` and run `vendor/bin/rector process --clear-cache` in your local dev:

```php
// rector.php
use Rector\Php70\Rector\Ternary\TernaryToNullCoalescingRector;

// ... 
$services->set(TernaryToNullCoalescingRector::class);
```

That should take care in the entire codebase.

jeromegamez:
Done!

samsonasik (approved):

samsonasik:
Thank you @jeromegamez

PR #4731, opened 2021-05-24
Remove explicit condition that is always true

jeromegamez:
The relevant part is in the `excerpt()` function, the rest was automatically applied by phpcbf.

The condition previously was

```php
if (isset($phrase)) {
    // ...
} elseif (!isset($phrase)) {
    // ...
}
```

Replacing the second `elseif` with just `else` also removes the need to ignore the following lines with PHPStan 🥳

:octocat:


paulbalandan (approved):

samsonasik (approved):

samsonasik:
Thank you @jeromegamez

PR #4733, opened 2021-05-24
Entity getCastHandlers

MGatner:
**Description**
I think it is important for classes that interact with `Entity` (mostly models) to know what data types any given Entity can handle. It's unfortunate these properties were not static to begin with, but this PR provides a deprecation and static getter to access it in a backwards-compatible approach.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Entity/Entity.php
+++ b/system/Entity/Entity.php
@@ -34,6 +34,35 @@ use JsonSerializable;
 class Entity implements JsonSerializable
 {
 	/**
+	 * Default cast handlers
+	 *
+	 * @var array<string, string>
+	 */
+	private static $defaultCastHandlers = [
+		'array'     => ArrayCast::class,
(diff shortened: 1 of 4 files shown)


lonnieezell (commented):
Would it make sense to implement some magic method that can read from the new static `$defaultCastHandlers` to ease transition away from it?

MGatner:
I'm not sure what that would be, but I'm down for suggestions!

MGatner:
I sat down and looked at this again and I'm still not sure what you mean @lonnieezell. Could you give an example or some pseudo-code?

MGatner:
While I still think this feature is desirable the Entity class has grown more and we have even higher priority issues with it. I am closing this, with possible version 5 reconsideration.

PR #4734, opened 2021-05-24
Add environment spark command

paulbalandan:
**Description**
Adds a spark command to show/change the current environment. Changes will be effected in the next script execution.
```console
$ php spark env
# Your current environment is set as testing.

$ php spark env development
# Environment successfully changed to "development".

```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Utilities/Environment.php
+++ b/system/Commands/Utilities/Environment.php
@@ -79,7 +79,7 @@ final class Environment extends BaseCommand
 	{
 		if ($params === [])
 		{
-			CLI::write(sprintf('Your environment is currently set as %s.', CLI::color(ENVIRONMENT, 'green')));
+			CLI::write(sprintf('Your environment is currently set as %s.', CLI::color($_SERVER['CI_ENVIRONMENT'] ?? ENVIRONMENT, 'green')));
 			CLI::newLine();
 
 			return;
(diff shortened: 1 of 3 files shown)


system/Commands/Utilities/Environment.php
```diff
@@ -0,0 +1,153 @@
...
+				CLI::newLine();
+
+				return false;
+			}
+
+			copy($baseEnv, $envFile);
+		}
+
+		$pattern = preg_quote($_SERVER['CI_ENVIRONMENT'] ?? ENVIRONMENT, '/');
+		$pattern = sprintf('/^[#\s]*CI_ENVIRONMENT[=\s]+%s$/m', $pattern);
```
samsonasik, line 146: please add https://regex101.com comment for this for future reference

paulbalandan: Added

MGatner (approved):
I like this, very helpful and easy to use. This commits us to a baseline **env** that is entirely comments - this is currently the case and I believe this was always the intention, but is there some place we could make note of this "for future generations"? (i.e. for me when I forget next week) Maybe somewhere in **contributing/** we need some more "policies" or something?

paulbalandan:
> I like this, very helpful and easy to use. This commits us to a baseline **env** that is entirely comments - this is currently the case and I believe this was always the intention, but is there some place we could make note of this "for future generations"? (i.e. for me when I forget next week) Maybe somewhere in **contributing/** we need some more "policies" or something?

I'm not sure I get what you mean. Are you referring to the env file?

MGatner:
Right, since this command will publish the template (**env**) to a live **.env** we need to be certain that [our repo version of the template](https://github.com/codeigniter4/CodeIgniter4/blob/develop/env) is all comments, otherwise using this command could cause accidental environment settings.

I was looking for ideas of where we could track this - maybe add it to https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/internals.rst ?

paulbalandan:
Understood. Will look into this.

MGatner:
Thanks! Looks great. Re-approve!

PR #4742, opened 2021-05-25
Add default TTL

MGatner:
**Description**
I am constantly adding a default Cache Time To Live value to modules and projects that interact with the cache. It's quite frustrating that the framework interface and handlers have this value hard-coded, so we cannot change that in version 4, but supplying this in the config file gives an opportunity for developers to implement it themselves (and to replace the hard-coded interface in version 5).

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/Cache.php
+++ b/app/Config/Cache.php
@@ -84,6 +84,21 @@ class Cache extends BaseConfig
 
 	/**
 	 * --------------------------------------------------------------------------
+	 * Default TTL
+	 * --------------------------------------------------------------------------
+	 *
+	 * The default number of seconds to save items when none is specified.
+	 *
+	 * WARNING: This is not used by framework handlers where 60 seconds is
+	 * hard-coded, but may be useful to projects and modules. This will replace
+	 * the hard-coded value in a future release.
+	 *
+	 * @var integer
+	 */
+	public $ttl = 60;
+
+	/**
+	 * --------------------------------------------------------------------------
 	 * File settings
 	 * --------------------------------------------------------------------------
 	 * Your file storage preferences can be specified below, if you are using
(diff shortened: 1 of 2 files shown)


paulbalandan (commented):
If this is not used by the framework, then how do projects benefit from this?

MGatner:
Two points...

1. Numerous components rely on the cache. For example, if I want my `WidgetFactory` to be able to store/save a `Widget` to cache I can call `cache()->save('widget', $widget)` but then developers using my `WidgetFactory` are locked into 60 seconds (which sometimes is wildly inappropriate). If I want to let developers set their own TTL I either need to add parameters to every method that might save to cache or supply a configuration (I typically choose the latter). But then when you make a `FruitFactory` now we have `WidgetConfig::$cacheTTL` and `FruitConfig::$cacheTTL`. If we don't provide a means of managing this in the framework then every component that uses cache is stuck with this choice.
2. Introducing the configuration value now lets people start using it so there are no surprises when it is added in version 5. This includes us - we cannot change the Cache handler interface but anywhere in the framework we want to pass in `$cacheConfig->ttl` (with default value 60) we will be maintaining backwards-compatibility while also offering configurability.

MGatner:
I'll also add, TTL is starting to push me towards the cache edge. The handling is very inconsistent (e.g. sometimes `0` means "don't cache", others it means "cache indefinitely"). We've already had some small breaking changes to Cache to fix bugs, and still haven't released our PSR adapters (https://github.com/codeigniter4/cache/pull/2). We might need to think about some larger Cache changes soon.

paulbalandan (approved):

Issue #4382, opened 2021-03-02
Discussion: Coding Standard


kenjis:
CodeIgniter4 code does not follow the coding standard.

When I checked CI4 code under `system`, phpcs reported 6,336 errors  .

```sh-session
$ vendor/bin/phpcs --standard=./vendor/codeigniter4/codeigniter4-standard/CodeIgniter4 --encoding=utf-8 system/ | grep ' | ERROR ' | wc -l
    6336
```

This means that the coding standard does not match the current situation.

Since we have the coding standard, I think we should remove unnecessary rules and bring errors and warnings to zero.

What are your thoughts?

paulbalandan:
I have proposed a few PR before changing our coding standard from codesniffer to cs-fixer, because our current coding standard is not updated and pretty unmaintained now. I have started before a big overhaul but resulted in several diffs so I opted for small batches.

paulbalandan:
Now that cs fixer supports php8, I'll try to revive my old PR to replace our cs.

paulbalandan:
I have talked with Louis before on slack and he suggested opening a new repo instead of replacing the current repo.

kenjis:
Do you mean we will stop maintaining [the current coding standard](https://github.com/codeigniter4/coding-standard) and move to php-cs-fixer?
What is the reason?

WinterSilence:
@paulbalandan @kenjis you can merge configs by [symplify/easy-coding-standard](https://github.com/symplify/easy-coding-standard)

kenjis:
@WinterSilence  Thanks!
Since we have already the coding standard, if there is something wrong with it,
it is better not throw it all away.
easy-coding-standard might be a better solution?

WinterSilence:
@kenjis ESC is aggregator for CsFixer and CodeSniffer rules + additional rules for them. package contains validator and normalizer. It's easy solution for use that's packages together. You can, step by step, replace CodeSniffer rules to CsFixer aliases.

sfadschm:
> When I checked CI4 code under `system`, phpcs reported 6,336 errors.

I know CI4 does not follow the standard everywhere, but just for my understanding: Did you exclude the `ThirdParty` folder when running the scan?

kenjis:
It seems there is no error in `system/ThirdParty/`.

```sh-session
[...]

PR #4743, opened 2021-05-26
[Rector] Apply Rector : ListToArrayDestructRector

samsonasik:
Change `list()` to array destruct with `ListToArrayDestructRector` rector rule.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -23,6 +23,7 @@ use Rector\EarlyReturn\Rector\If_\ChangeIfElseValueAssignToEarlyReturnRector;
 use Rector\EarlyReturn\Rector\If_\RemoveAlwaysElseRector;
 use Rector\EarlyReturn\Rector\Return_\PreparedValueToEarlyReturnRector;
 use Rector\Php70\Rector\Ternary\TernaryToNullCoalescingRector;
+use Rector\Php71\Rector\List_\ListToArrayDestructRector;
 use Rector\Php73\Rector\FuncCall\JsonThrowOnErrorRector;
 use Rector\Php73\Rector\FuncCall\StringifyStrNeedlesRector;
 use Rector\Set\ValueObject\SetList;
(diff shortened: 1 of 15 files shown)


paulbalandan (commented):
I have no questions for this except that can we make the array destructuring in a single line?

The current sniff for this makes the list call as multiline and it looks ugly when combined with aligned `=`.

samsonasik:
Yes, that seems part of CodeIgniter Coding Standard that make it on before commit, run rector just use one line for it:

Screen Shot 2021-05-26 at 13 25 57

samsonasik:
@paulbalandan implemented, I updated to use single line.

paulbalandan (approved):
Yes, the multilining should be reported in the `coding-standards` repo.

MGatner (approved):
I haven't looked at every change but I like this conceptually.

MGatner:
We really need to address the code sniffing. I find myself using `--no-verify` on most of my commits because it screws up spacing, or insists `int` should be `integer`. We have another discussion going on whether to switch vendors - can we resurrect that thread and get this solved?

paulbalandan:
> We really need to address the code sniffing. I find myself using `--no-verify` on most of my commits because it screws up spacing, or insists `int` should be `integer`. We have another discussion going on whether to switch vendors - can we resurrect that thread and get this solved?

We have this thread #4382 .

There's a suggestion to use ecs there, but I find it not easy to use IMO. For the `translations` repo, I have already switched that to php-cs-fixer.

samsonasik:
is this merge-able for now or need to wait for the coding-standard to detect it ?

paulbalandan:
I think we can merge this for now. Let us tackle the cs separately.

samsonasik:
@paulbalandan @MGatner thank you for the review.

PR #4744, opened 2021-05-26
Fix GC issue when session lifetime is set to 0

### LINKED ISSUE #4169: Bug: All database handler sessions lost during garbage collection, if $sessionExpiration set to 0
**Describe the bug**
When using the Database Session handler and setting $sessionExpiration to 0 in App.php, all sessions are effectively lost when the garbage collector runs.  

**CodeIgniter 4 version**
Current version (4.0.4)

**Affected module(s)**
DatabaseHandler
https://github.com/codeigniter4/CodeIgniter4/blob/6d56d6b5b723b72ec9a0485dc5e9ef1a40aca101/system/Session/Handlers/DatabaseHandler.php#L323

This can be seen more easily by setting session.gc_probability to 0 in php.ini after setting $sessionExpiration to 0.  Any stored sessions in the ci_session table are lost after navigating to another CI controller.

**Context**
 - OS: Windows 10
 - Apache 2.4.46
 - PHP version 7.3.21

lf-uraku-yuki:
**Description**
Fixes #4169 

If the session validity period is larger than 1, it operates as usual, and if it is 0, the setting value on the PHP side is used for GC timing.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/libraries/sessions.rst
+++ b/user_guide_src/source/libraries/sessions.rst
@@ -458,10 +458,10 @@ Preference                     Default                                      Opti
     unexpected results or be changed in the future. Please configure
     everything properly.
 
-.. note:: If ``sessionExpiration`` is set to ``0`` , the session.gc_maxlifetime
-    setting set by PHP in session management will be used as is
-    (often the default value of ``1440`` ). This needs to be changed using
-    php.ini etc. as needed.
+.. note:: If ``sessionExpiration`` is set to ``0``, the ``session.gc_maxlifetime``
+    setting set by PHP in session management will be used as-is
+    (often the default value of ``1440``). This needs to be changed in
+    ``php.ini`` or via ``ini_set()`` as needed.
 
 
 In addition to the values above, the cookie and native drivers apply the


MGatner:
I don't really know what this does, but the code does solve the bug. Any other maintainers familiar? Or @lf-uraku-yuki can you offer more explanation/documentation on what is happening here?

MGatner:
Is there any scenario where someone would be setting this to `0` intentionally?

lf-uraku-yuki:
The session expiration time is set to the same value in session.gc_maxlifetime in the current implementation. Therefore, session.gc_maxlifetime will be set to 0 even if you want to set the session expiration date to 0 (= While the browser is open). This side effect occurs with the gc method in each SessionHandler and destroys the existing session.

lf-uraku-yuki:
Note that sessionHandler gc is called every time a session is started and is executed with the probability calculated from session.gc_divisor and session.gc_probability.

https://www.php.net/manual/en/sessionhandler.gc.php

MGatner (approved):
Okay I think I understand now. And to answer my own question: no, you would never intentionally set this to `0` *while using the Session class*. I'll leave the PR in case anyone else wants a look, but I'm good with it.

michalsn (commented):
As far as I remember setting `0` will work just fine on Debian and will stop PHP gc entirely. This PR can cause problems for some people since their apps will start using the default value which is `1440`.

We probably should mention this at least in the changelog.

MGatner:
Really hard to find info on this, but StackOverflow does seem to agree:

> A temporary fix might be to do something like change your session.gc_probability to 0 so the session garbage collecting NEVER happens.

> But watch out, on most xampp/ampp/...-setups and some linux destributions it's 0, which means the file will never get deleted until you do it within your script (or dirty via shell)

I think we need some more research before going with this.

nmolinos:
I'm not too keen on this fix as it leaves all stale sessions in the table forever (if $sessionExpiration is set to 0).  What I did as a quick fix in my environment was to modify the garbage collection method in the DatabaseHandler class to clean out sessions with timestamps older than 2 days.

```
public function gc($maxlifetime): bool
{
	//deletes all sessions if $sessionExpiration is set to 0 (on browser close)
	//https://github.com/codeigniter4/CodeIgniter4/issues/4169
	//return ($this->db->table($this->table)->delete('timestamp < ' . (time() - $maxlifetime))) ? true : $this->fail();

	//Instead, clean out stale sessions older than 2 days (ending sessions on browser close or 2 days whichever comes first)
	return ($this->db->table($this->table)->delete('timestamp < UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 2 day))')) ? true : $this->fail();
}
```

I think the ideal solution involves having two separate values for these settings.  One that holds the client-side session expiration, and another for the server-side session expiration used during garbage collection (in place of my hard-coded 2 days in the above example).  It would certainly be difficult to clearly explain to someone in the documentation or configuration file though.

Thoughts?

kenjis:
@nmolinos 
I mean if this PR gets merged, 
when `$sessionExpiration` is set to 0, `session.gc_maxlifetime` would be 1440 (PHP default).
And some stale sessions would be deleted.

```php
		elseif ($this->sessionExpiration > 0)
		{
			ini_set('session.gc_maxlifetime', (string) $this->sessionExpiration);
		}
```

kenjis:
@lf-uraku-yuki Thank you for your cooperation!

nmolinos:
> Are we sure this is safe across different OSes? See e.g. Michal's comment [above](https://github.com/codeigniter4/CodeIgniter4/pull/4744#pullrequestreview-669024151).

Since this change would effectively cause sessions to be removed after 24 minutes of inactivity (for the PHP.ini default of 1440 in some cases -- when sessionExpiration is set to 0), this should definitely be made more apparent.  Possibly adding a note to the configuration file where sessionExpiration is set, in addition the ChangeLog as mentioned by Michal.

kenjis:
When you set `$sessionExpiration` `0`.

Before:
```
session.gc_maxlifetime = 0
session.cookie_lifetime = 0
```

After:
```
session.gc_maxlifetime = 1440 (PHP default)
session.cookie_lifetime = 0
```

kenjis:
@MGatner @michalsn 

> As far as I remember setting 0 will work just fine on Debian and will stop PHP gc entirely. 

I've confirmed Debian 10.11's behavior.
`$sessionExpiration = 0` does not stop PHP gc.

### Procedure

Build docker environment and installs CI 4.1.4.
```
$ git clone git@github.com:kenjis/docker-codeigniter-apache.git
$ cd docker-codeigniter-apache
$ make create-project
```

Check Debian version.
```
$ make web
docker-compose exec web bash
root@df110741382e:/work/backend# cat /etc/debian_version 
10.11
```

Update Home controller.
```php
<?php

namespace App\Controllers;

use Config\Services;

{
    public function index()
    {
        //ini_set('session.gc_divisor', (string) 1);

        $session = Services::session();
    }
}
```

Create session files.
```
$ curl http://localhost/
$ curl http://localhost/
$ curl http://localhost/
$ curl http://localhost/
$ curl http://localhost/
```

```
$ ls -l writable/session/
total 56
-rw-------  1 kenji  staff   91 10 13 11:09 ci_session3318f2d7835906ce4fbbdea234eafaef8bc6877a
-rw-------  1 kenji  staff   91 10 13 11:09 ci_session5716d271e8d185a3fbc919f106cf3b0b4ef4abd5
-rw-------  1 kenji  staff   91 10 13 11:09 ci_session675da4983f476cf258b5b3ba9f8fe7d77443ff04
-rw-------  1 kenji  staff   91 10 13 11:09 ci_sessionaaba50ceee3c236fa456359cc97950a67d7f899d
-rw-------  1 kenji  staff   91 10 13 11:09 ci_sessionbabc6451df73ca53c87d496b49b287aa3db3a0d9
-rwxr-xr-x  1 kenji  staff  131 10 13 10:54 index.html
```

Set `session.gc_divisor` 1.
```php
<?php

namespace App\Controllers;

use Config\Services;

{
    public function index()
    {
        ini_set('session.gc_divisor', (string) 1);

        $session = Services::session();
    }
}
```

Set `$sessionExpiration` 0.
```diff
--- a/backend/app/Config/App.php
+++ b/backend/app/Config/App.php
@@ -174,7 +174,7 @@ class App extends BaseConfig
      *
      * @var int
      */
-    public $sessionExpiration = 7200;
+    public $sessionExpiration = 0;
 
     /**
      * --------------------------------------------------------------------------
```

Run Home controller.
```
$ curl http://localhost/
```

```
$ ls -l writable/session/
total 8
-rwxr-xr-x  1 kenji  staff  131 10 13 10:54 index.html
```

kenjis:
@MGatner @michalsn @nmolinos 
`session.gc_maxlifetime = 0` does not stop the GC. It sees all sessions as 'garbage' and potentially cleaned up.
So this PR does not shortens the lifetime of Sessions, just lengthens them.
See https://github.com/codeigniter4/CodeIgniter4/pull/4744#issuecomment-941810278

But when a user set `$sessionExpiration` 0 (it means until the browser is closed),
the PHP default 1440 seconds may not be enough.

So it is better to add a note that when you set `$sessionExpiration` 0, you may want to change
`session.gc_maxlifetime` more than 1440.

MGatner:
Yes, stellar investigation! Puts my mind at ease. Is the comment you wrote above above appropriate for our User Guide? We would also need an entry in the Change Log it sounds like.

lf-uraku-yuki:
The signature and committer email address do not match because the user specified at the time of commit was incorrect. I'm sorry .

kenjis:
@lf-uraku-yuki Thank your for your contribution!

Next time, if possible, please use `git rebase` instead of `git merge` to update your branch.

PR #4748, opened 2021-05-27
Don't load PHP input body, if the request is too big

Sosko:
When I the client sending big http request (upload big file/s), there is potential risk, that the php haven't enough free memory:

`Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 98566176 bytes) in /var/www/ci4/system/HTTP/IncomingRequest.php on line 158`

Maybe there is somewhere config for disable this functionality, but I was unable to find it in the documentation.

This will prevent load the content to the memory, if the content is too big

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)

### DIFF
--- a/app/Config/Exceptions.php
+++ b/app/Config/Exceptions.php
@@ -57,4 +57,16 @@ class Exceptions extends BaseConfig
 	 * @var array
 	 */
 	public $sensitiveDataInTrace = [];
+        
+        
+        /**
+         * --------------------------------------------------------------------------
+         * Throw error on big request
+         * --------------------------------------------------------------------------
+         * If will be send big request on CI, throw an exception.
+         * 
+         * @var bool
+         */
+        
+        public $throwExceptionOnBigRequest = false;
 }
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -150,15 +150,24 @@ class IncomingRequest extends Request
 		{
 			throw new InvalidArgumentException('You must supply the parameters: uri, userAgent.');
 		}
-
-		// Get Maximum avaible memory
-		$memoryLimit = preg_split('/\d+\K/', ini_get('memory_limit'));
-		// Translate memory limit from human readible to bytes and substract already used memory
-		$freeMemory = ((int) $memoryLimit[0] ?? 0 ) * pow(1024, (array_search(strtolower($memoryLimit[1] ?? ""), ["k", "m", "g", "t"]) ?: -1) + 1) - memory_get_peak_usage();
-		// Get our body from php://input - if the send content is too big, it wouldn't be load to the memory
-		if ($body === 'php://input' && ((int) $_SERVER['CONTENT_LENGTH'] ?: 0) < $freeMemory)
+		// Get our body from php://input
+		if ($body === 'php://input')
 		{
+                    // Get Maximum avaible memory
+                    $memoryLimit = preg_split('/\d+\K/', ini_get('memory_limit'));
+                    // Translate memory limit from human readible to bytes
+                    $memoryLimitBytes = ((int) $memoryLimit[0] ?? 0 ) * pow(1024, ((int) array_search(strtolower($memoryLimit[1] ?? ""), ["k", "m", "g", "t"]) ?: -1) + 1);
+                    // Get free space in memory by substracting already used memory
+                    $freeMemory = $memoryLimitBytes - memory_get_peak_usage();
+                    // If the send content is too big, it wouldn't be load to the memory
+                    if(((int) $this->fetchGlobal("server", "CONTENT_LENGTH") ?? 0) > $freeMemory)
+                    {
+                        log_message(7, "The 'php://input' is too big for loading it to the \$body");
+                    } 
+                    else 
+                    {
 			$body = file_get_contents('php://input');
+                    }
 		}
 
 		$this->config       = $config;
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -162,7 +162,7 @@ class IncomingRequest extends Request
                     // If the send content is too big, it wouldn't be load to the memory
                     if(((int) $this->fetchGlobal("server", "CONTENT_LENGTH") ?? 0) > $freeMemory)
                     {
-                        log_message(7, "The 'php://input' is too big for loading it to the \$body");
+                        log_message("debug", "The 'php://input' is too big for loading it into \$body");
                     } 
                     else 
                     {
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -154,7 +154,7 @@ class IncomingRequest extends Request
 		// Get Maximum avaible memory
 		$memory_limit = preg_split('/\d+\K/', ini_get('memory_limit'));
 		// Translate memory limit from human readible to bytes and substract already used memory
-		$free_memory = $memory_limit[0] * pow(1024, (array_search(strtolower($memory_limit[1]), ["k", "m", "g", "t", "p"]) ?: -1) + 1) - memory_get_peak_usage();
+		$free_memory = ((int) $memory_limit[0]) * pow(1024, (array_search(strtolower($memory_limit[1]), ["k", "m", "g", "t", "p"]) ?: -1) + 1) - memory_get_peak_usage();
 		// Get our body from php://input - if the send content is too big, it wouldn't be load to the memory
 		if ($body === 'php://input' && ((int) $_SERVER['CONTENT_LENGTH'] ?: 0) < $free_memory)
 		{
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -158,7 +158,8 @@ class IncomingRequest extends Request
                     // Get unit
                     $unit = strtolower($memoryLimit[1] ?? "");
                     // Get the exponent
-                    $exponent = ((int) array_search($unit, ["k", "m", "g", "t"]) ?: - 1) + 1;
+                    $possibleExponents = ["", "k", "m", "g", "t"];
+                    $exponent = (int) array_search($unit, $possibleExponents) ?? 0;
                     // Translate memory limit from human readible to bytes
                     $memoryLimitBytes = ((int) $memoryLimit[0] ?? 0 ) * pow(1024, $exponent);
                     // Get free space in memory by substracting already used memory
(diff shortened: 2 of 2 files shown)


iRedds:
I think the implementation is not completely correct.

I believe that when the limit is exceeded, it is necessary to write to the log and, for example, throw an exception with an HTTP 500 error or define a handler.
Otherwise, the error cannot be detected, which will complicate the debugging.

Sosko:
The problem is, that the PHP throw the error, but it's not necessary always  parse the body. For example, when you uploading big file/s.
I agree, that it will be good to make notification about this state...

MGatner:
I agree with @iRedds - this should be a fatal condition that should throw an exception and quit.

I'm also not sure that we should be trying to guess the free memory - aren't there all sorts of memory management layers happening in every different environment that could make this misleading? In my opinion probably the safest is to error out when `CONTENT_LENGTH` exceeds `memory_limit` - this could still leave open a "sweet spot" for errors but I think that is up to the web server and OS, not the framework.

Sosko:
I really don't understand, why you need only HTTP error (by CI)....

image

So if you help me and write how I can avoid to load big HTTP request body, I will be very happy

Just try it by yourself upload big files.

MGatner:
@Sosko If your files are so large that they are exceeding the PHP memory limit then you have a server configuration issue. You should set your `upload_max_filesize` and `post_max_size` to something manageably under `memory_limit`. It is not the framework's responsibility to manage this. What @iRedds and I are proposing is to have a more helpful error when this does occur, but (unless I see some compelling research otherwise) this needs to be a hard stop.

Sosko:
Ok, I will try to explain my problem in more details.

When I started uploading files up to 125MB (just for testing the framework), the CI works fine.
`<form method="POST" enctype="multipart/form-data"><input type="file" name="file" /><input type="submit" /></form>`
In my PHP ini I have set the `memory_limit` to size `128M` and the `post_max_size` to default `8M`. For security reason I have set via nginx config only for this controller (Upload) `client_max_body_size 5000G;`. Because I don't want to set globally the POST limit for this project on this big number.

So when I starting uploading files bigger that the memory limit the CI still want load the full PHP body request (including the uploading files) into the memory (into the property $body). 

When the PHP error accoutre my first thought was: Why the CI loading the uploaded file (php://input) into RAM? This isn't necessary. Because the file is in the `/tmp` folder and the variable `$_FILES` have the necessary information's about the uploaded file. And I start this pull request.

And my opinion is still the same, stop loading big HTTP request into RAM even if will not be accessible all features in CI in this very specific case.

And I don't want to create single PHP file, I want be secured by CI, have access to DB via CI and have accessible the CI features...

MGatner:
Just to be sure I understand you: you have `post_max_size` set to `8M` but uploading a file larger than 128M still causes this crash? If this is true then I have not been understanding how `php://input` interacts with these configurations.

iRedds:
@Sosko, as far as I know, using a POST request with enctype="multipart/form-data" you cannot get data via php://input 
https://www.php.net/manual/en/wrappers.php.php#wrappers.php.input

Sosko:
@iRedds I now, but you can clearly see the error in my screenshot that the error occur when the request was send. When I comment the line which loading content from `php://input`, everything works ok. You can tryit by yourself via swagger editor:
```
openapi: "3.0.0"
info:
  description: ""
  version: "V1"
servers:
  - url: http://127.0.0.1/ci4
paths:
  /upload:
    post:
      tags:
      - "upload"
      description: "Upload file"
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                type:
                  type: "string"
                  description: "Type of media: disk"
                id:
                  type: "integer"
                  description: "id of object for it"
                file:
                  type: string
                  format: binary
      responses:
        200:
          description: "ok"
```

MGatner:
Actually, if [this thread](https://bugs.php.net/bug.php?id=74611) is still accurate then this does appear to be an issue: if `post_max_size` or `upload_max_filesize` is exceeded then the content is still available in `php://input`. I think we need some other way of differentiating this situation though, like maybe checking headers. I will think on this some more.

MGatner:
I've looked at 5 other frameworks. None of them seem to be doing any special handling around potential `memory_limit` violations due to `php://input`. I will say that almost all of them create a Stream context, rather than reading the data right away. It may be that in these sort of cases the framework will delay the data read, check the headers later and identify it as a failed POST, then never read in from `"php://input"`. It would take more research to be sure, but I'm inclined to think that is how we should handle this as well.

feryardiant:
> Just try it by yourself upload big files.

@Sosko you might want to try implementing Upload chunk instead.

> I was unable to find the answer in the documentation. 

Also, [the forum](https://forum.codeigniter.com/thread-65547-post-369032.html) might give you the answer.

Sosko:
> > Just try it by yourself upload big files.
> 
> @Sosko you might want to try implementing Upload chunk instead.

@feryardiant This is not great for this applications, the upload speed significantly decreased. And trust me, you don't want wait longer when you uploading (hundreds or thousands GB) big file.


I'll be so happy, if the CI just don't load the content of `php://input` if the input is too big. That will be enough for me. Nothing will break. I don't except that this problem will have more people, so if it will be in config file the option for disabling it, it will be great.

feryardiant:
> @feryardiant This is not great for this applications, the upload speed significantly decreased. And trust me, you don't want wait longer when you uploading (hundreds or thousands GB) big file.

@Sosko Somehow I wondering what kind of use case you were working on, I personally never had any experience with uploading more than 3GB file in single run. And you said _hundreds or thousands gb_ ? 😮 

Even for uploading 100 MB file, I always use chunk and never had any slow down issue, also implement resumable upload whenever possible. In that case, I rather choose reliability over speed. It's pointless to have good upload speed if it causing some issues.

That said, I believe that you should re-check/configure your application stack, not only the back-end framework you use. As stated by @MGatner above _"It is not the framework's responsibility to manage this."_

feryardiant:
> you tell me to set the memory_limit to 800GB for correct working of CI? Or more when the file will be 1TB?

Nope, I'm telling you to _"re-check/configure your application stack"_ not just the back-end framework you were using, or the language, but the whole stack.

> And can you please explain me, why the CI must load all `php://input` to RAM all the time?

I am not proficient enough to give an advice in this particular case.

> I will do the upload by single PHP script (which I don't want, because I have to save some information in DB with validation and authentication).

You certainly need to spend some times to research what's works best to your scenario. Perhaps _try upload over FTP instead of HTTP_, or _try another client-side upload library_, or even _setup another (microservice-ish) back-end using another framework specifically to handle that functionality and connected to the main app through API_ or whatever.

On the other hand, you might need to read these :
- > If configuring the webserver to accept that large POST requests is not an option, or when you want to go beyond the webserver's limit, then you have no other option than splitting the file in the client side and reassembling the parts in the server side.
  > Said that, FTP is a much better choice than HTTP for transferring (large) files over network. I'd reconsider the choice of using HTTP for that.
  https://stackoverflow.com/a/5053380/881743
- > Browsers are very good at that and the overhead for a big file is really near nothing. Don't use websockets for that task.
  https://stackoverflow.com/a/11081396/881743

MGatner:
I've done a bit more research, and it is actually simpler than all this. `CONTENT_LENGTH` is *only* set on POST requests and refers specifically to the size of the POST content. The POST content is **always** available from `php://input`, whether it is valid or not. The problem is that we read the input stream regardless of POST validity - we should instead be doing something like checking `if (isset($_SERVER['CONTENT_LENGTH']) && $_SERVER['CONTENT_LENGTH'] > get_post_max_value()) ...`. This will prevent reading invalid POST content into the request body. We will also need to check if there are other locations where `php://input` is accessed, and if those need similar checks or if we should somehow invalidate the resource. 

In summary, I believe @Sosko is correct that we have a problem here, but the solution is much simpler than comparing available memory.

feryardiant:
> why the CI must load all `php://input` to RAM all the time?

Your question drive me to spend some times to dig in. 😆
And end up with No, CI does not. Not _all the time_

As my understanding `php://input` will only be used when we send other than `GET` & `POST` request. While file upload will always in `$_FILES` global variable no matter how big the file is. That's handled by PHP it-self, not the framework.

To prove it, I tried to create simple project

```
composer create-project codeigniter4/appstarter ci4 && cd ci4
```
Create vscode `launch.json`
```
touch .vscode/launch.json
```
paste this config which is the default
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Listen for Xdebug",
            "type": "php",
            "request": "launch",
            "port": 9003
        }
    ]
}
```
add this line to `app/Config/Routes.php` file
```php
$routes->get('/', 'Home::index');
$routes->post('/', 'Home::upload');
```
add update `app/Controllers/Home.php` file
```php
{
	public function index()
	{
		return view('upload'); // => create view with simple POST multipart form
	}

	public function upload()
	{
		$file = $this->request->getFile('file'); // => $_FILES

		if ($file->isValid() && $file->move(WRITEPATH . 'uploads', $file->getRandomName())) {
			var_dump(
				$this->request->getPost(), // => $_POST
				$this->request->getRawInput() // => $request->body = null
			);
			echo $file->getClientName() . ' have been uploaded';
		} else {
			echo "No file";
		}
	}
}
```
Make sure you have `xdebug.mode=debug` in your `php.ini` file, add break point to the `if` statement and start vscode debug.

ze9thSdi0Z

_correct me if I'm wrong_ 😁

feryardiant:
> We will also need to check if there are other locations where `php://input` is accessed

@MGatner sent request using other than POST & GET method, (PUT, PATCH, DELETE, etc)

image

feryardiant:
Today I just read about https://uppy.io, they said

> Large uploads survive network hiccups thanks to resumable file uploads via the open [tus](https://tus.io) standard

and somewhat bring me back to this issue and I thought it might be useful to share it with you guys, _just in case_

MGatner:
Thanks @feryardiant! I discovered `tus` not too long ago, it is pretty awesome. I don't think it is really relevant to this issue or the framework (at this point) but it is good to raise awareness of it.

feryardiant:
@MGatner 

> I don't think it is really relevant to this issue

Not really, since @Sosko said _"When I the client sending big http request (upload big file/s)"_ in the very beginning of this issue and the words _"Large uploads survive network hiccups"_ on the uppy.io feature page, might help him to solve his problem.

> or the framework (at this point)

Yes, Exactly!

MGatner:
This PR is not the way to handle the issue. I have created a new bug report to track it (https://github.com/codeigniter4/CodeIgniter4/issues/6246) for now until a solution can be coded. 

@Sosko thank you for bringing this to our attention and attempting a solution.

kenjis:
Why nobody implements this https://github.com/codeigniter4/CodeIgniter4/pull/4748#issuecomment-859526582 ?

MGatner:
Dunno! I never made a PR for it.

PR #4756, opened 2021-05-31
[Autoloader] include_once is not needed on Autoloader::loadClass() with no namespace

samsonasik:
include_once is not needed on get class with no namespace for `Autoloader::loadClass()` call as it try to get `app\Config` that autoloaded.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -22,17 +22,17 @@ use InvalidArgumentException;
  * An autoloader that uses both PSR4 autoloading, and traditional classmaps.
  *
  * Given a foo-bar package of classes in the file system at the following paths:
- *```
+ * ```
  *      /path/to/packages/foo-bar/
  *          /src
  *              Baz.php         # Foo\Bar\Baz
  *              Qux/
  *                  Quux.php    # Foo\Bar\Qux\Quux
- *```
+ * ```
  * you can add the path to the configuration array that is passed in the constructor.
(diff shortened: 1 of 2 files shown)


samsonasik:
The test for it is already in : 

https://github.com/codeigniter4/CodeIgniter4/blob/5fc58f91c306c27c51728d74c2f92b236788a0fc/tests/system/Autoloader/AutoloaderTest.php#L175-L189

MGatner (approved):

paulbalandan (commented):
I think I'm missing something. If a non-namespaced file is found, this will now just return the file path without `include_once` it. So where will this get loaded? Will this be delegated in the next autoload function registered?

samsonasik:
Actually, it should just return false as it somehow handled by Fabricator, iirc, it previously was needed as cause class redeclare, @paulbalandan what do you think of non - namespace return false?

samsonasik:
@paulbalandan looking at the previous change on this, this was introduced for faster load due to `loadLegacy()` at https://github.com/codeigniter4/CodeIgniter4/pull/3413 . 

As the `loadLegacy()` no longer exists, it is safe to just return false early. On looking on Config class, eg: 'Modules' config, there is `config('Modules')` usage for that.

I updated at https://github.com/codeigniter4/CodeIgniter4/pull/4756/commits/81c54543b106d83276eab2dcd761b04100beaf88

paulbalandan (approved):

samsonasik:
@MGatner @paulbalandan thank you for the review.

PR #4759, opened 2021-06-01
[Rector] Apply Rector: VarConstantCommentRector

samsonasik:
Add `@var` to Class Constants with `VarConstantCommentRector` rector rule.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -22,16 +22,14 @@ class URI
 	/**
 	 * Sub-delimiters used in query strings and fragments.
 	 *
-	 * @const string
-	 * @var   string
+	 * @var string
 	 */
 	const CHAR_SUB_DELIMS = '!\$&\'\(\)\*\+,;=';
 
 	/**
 	 * Unreserved characters used in paths, query strings, and fragments.
 	 *
-	 * @const string
-	 * @var   string
+	 * @var string
 	 */
 	const CHAR_UNRESERVED = 'a-zA-Z0-9_\-\.~';
 
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -23,6 +23,7 @@ class URI
 	 * Sub-delimiters used in query strings and fragments.
 	 *
 	 * @const string
+	 * @var   string
 	 */
 	const CHAR_SUB_DELIMS = '!\$&\'\(\)\*\+,;=';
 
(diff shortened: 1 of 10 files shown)


system/HTTP/URI.php
```diff
@@ -23,13 +23,15 @@ class URI
 	 * Sub-delimiters used in query strings and fragments.
 	 *
 	 * @const string
```
paulbalandan, line 25: Not related but is `@const` a valid phpdoc?

paulbalandan (approved):

system/HTTP/URI.php
samsonasik: I can't find the reference for `@const` in php documentor manual, remove it?

paulbalandan: That'd be nice. 😁

samsonasik: removed https://github.com/codeigniter4/CodeIgniter4/pull/4759/commits/52a1823e589f6a9852af4efad180e8489e09403c 👍

paulbalandan: Thanks

samsonasik:

MGatner:
I realize I missed the boat on this, but this seems unnecessary. Class constants are constant, so their type is always apparent from their declaration. What does adding docblocks accomplish?

samsonasik:
@MGatner that mostly for consistency with other constants that has `@var`

MGatner:
I would rather see a rule that pulled all `@var` tags from constants. "var" is shorthand for "variable" (literally "values that change") and is decidedly not "constant". In addition to being unnecessary I suspect this is probably a docblock violation (though I haven't researched that).

samsonasik:
This is example in phpDocumentor  for using `@var` for const.

https://github.com/phpDocumentor/phpDocumentor/blob/afecb1227c943f31633ef4587769fbdfa7e47cc7/tests/data/ApiTagTest.php#L65-L74

I will try create new custom rule for pull `@var` for Class Constants if we have decision is to remove them.

MGatner:
I do see that constants can have `@var` tags (also found this https://github.com/phpDocumentor/phpDocumentor/blob/dbbe6c48e3553870f3077d8dcc2eda57a1df2c3b/src/phpDocumentor/Descriptor/ConstantDescriptor.php#L109-L123). But does that mean they should? All I can find on that topic is debates on StackOverflow 😅

I'd say at this point it is up to us. I won't fight it if you all want them there, doesn't hurt anything.

samsonasik:
@MGatner ok, here PR to remove it https://github.com/codeigniter4/CodeIgniter4/pull/4766

PR #4763, opened 2021-06-01
Revised coding standards

paulbalandan:
**Description**
This is yet another attempt to migrate to `php-cs-fixer` as our cs tool.

`utils/PHPCSFixer/CodeIgniter4.php` contains the rules, while `.php-cs-fixer.dist.php` is the configuration file as per v3.0 of `php-cs-fixer`

I have not yet added all available rules as diffs will become astronomically large. My goal is to add the rules in batches until we can move this in its own repo so that other repos (like the `translations` etc.) can benefit from the same ruleset.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Database/SQLSRV/Connection.php
+++ b/system/Database/SQLSRV/Connection.php
@@ -129,12 +129,11 @@ class Connection extends BaseConnection
 			unset($connection['UID'], $connection['PWD']);
 		}
 
+		sqlsrv_configure('WarningsReturnAsErrors', 0);
 		$this->connID = sqlsrv_connect($this->hostname, $connection);
 
 		if ($this->connID !== false)
 		{
-			sqlsrv_configure('WarningsReturnAsErrors', 0);
-
 			// Determine how identifiers are escaped
 			$query = $this->query('SELECT CASE WHEN (@@OPTIONS | 256) = @@OPTIONS THEN 1 ELSE 0 END AS qi');
 			$query = $query->getResultObject();


MGatner:
I've been following your work this morning on NexusPHP CS Config. I like the project and the approach, but is there any way we can integrate this with `codeigniter4/codeigniter4-standard` so we retain an "official" coding standard for the framework?

MGatner:
For this PR it appears that some spacing rules have been applied. I like your plan of keeping rules separated - could this PR just be the change to the utility, without applying it at all?

paulbalandan:
> I've been following your work this morning on NexusPHP CS Config. I like the project and the approach, but is there any way we can integrate this with `codeigniter4/codeigniter4-standard` so we retain an "official" coding standard for the framework?

Actually, I have talked this with Louis before (some time in late last year) and he's not in favor of integrating this as involved are two different cs tools. He believes changing CS on a major version bump would make users "caught off guard" and suggested opening a new repo instead in the `codeigniter4` namespace. So, if ever this got finished I'm planning to create a `codeigniter4/revised-coding-standards` (or some name else) .

paulbalandan:
The spacing changes happened for the array contents. our current cs maintains a 0.75 tab space and aligns the array closing brace with the nearest element.

>  could this PR just be the change to the utility, without applying it at all?

Sorry, I do not understand this. 😅

MGatner:
> if ever this got finished I'm planning to create a codeigniter4/revised-coding-standards

Okay that sounds great. I'm not so much after keeping it on the current repo as having it a part of the CodeIgniter organization so we can publish the expectation that all projects use it as the official styler.

> Sorry, I do not understand this

Oops! I just meant: can this PR update the workflows, Composer, config files, etc without changing any code in **system/** or **tests/**?

.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,30 @@
...
+use Nexus\CsConfig\Factory;
+use PhpCsFixer\Finder;
+use Utils\PHPCSFixer\CodeIgniter4;
+
+$finder = Finder::create()
+	->files()
+	->in([
+		__DIR__ . '/system',
+		__DIR__ . '/tests',
+		__DIR__ . '/utils',
```
samsonasik, line 14: `app` as well?

paulbalandan:
> > Sorry, I do not understand this
> 
> Oops! I just meant: can this PR update the workflows, Composer, config files, etc without changing any code in **system/** or **tests/**?

So should the changes be in a separate PR?

.php-cs-fixer.dist.php
paulbalandan: I intentionally left `app/` because when we reach the `header_comment` rule I don't want to add the headers there. So, I'm planning a separate configuration file for `app` similar to this one except that `header_comment` is specifically set to `false`.

MGatner:
@paulbalandan Can you come weigh in on the PSR conversation at `#develop` before going any further?

paulbalandan:
I'll break the actual code changes in a separate PR.

paulbalandan:
OK. I don't know what happened. 😂

PR #4764, opened 2021-06-01
Update Query.php

randallmoraes:
With the regular expression '/:[a-z\d.)_()+:/i' the criteria are not met when there is a function in the where, for example :
MD5 (field) = $ value
MD5 (concat(field1,field2)) = $ value

**Description**

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -321,7 +321,7 @@ class Query implements QueryInterface
 	{
 		$sql = $this->finalQueryString;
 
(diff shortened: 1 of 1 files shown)


paulbalandan (changes requested):
You just reverted this to the old condition which does not work with postgres JSON operators. I would be keen to accept this for a more inclusive regex.

And please add tests.

PR #4765, opened 2021-06-01
allow for folders for file caching

colethorsen:

**Description**
Fix the previous breaking change that prevents folders from being created for file based caching allow folders for file caching increases performance and organization when there are significant numbers of cached files.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

**Notes**
(e.g. fixes <hash>12345)

### DIFF
--- a/system/Cache/Handlers/BaseHandler.php
+++ b/system/Cache/Handlers/BaseHandler.php
@@ -64,9 +64,9 @@ abstract class BaseHandler implements CacheInterface
 		{
 			throw new InvalidArgumentException('Cache key cannot be empty.');
 		}
-		if (strpbrk($key, self::RESERVED_CHARACTERS) !== false)
+		if (strpbrk($key, static::RESERVED_CHARACTERS) !== false)
 		{
-			throw new InvalidArgumentException('Cache key contains reserved characters ' . self::RESERVED_CHARACTERS);
+			throw new InvalidArgumentException('Cache key contains reserved characters ' . static::RESERVED_CHARACTERS);
 		}
 
 		// If the key with prefix exceeds the length then return the hashed version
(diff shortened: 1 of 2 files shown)


colethorsen:
You also can't just remove a feature/bug/whatever you want to call it that production applications are relying on without a proper deprecation cycle, and without a proper solution to the problem that subdirectories fix. i.e. it could just be built to automatically subdirectory based on the first x characters in the string or something similar.

MGatner:
> You also can't just remove ... that production applications are relying

Agreed. Like I said this was totally unintentional, but I believe that is because the feature was unintentional to begin with (or at least entirely undocumented/commented). We can make this a priority but I'm not comfortable making this call by myself. If you have time and capacity to flesh out this PR to be a proper "allow subdirectories" emendation it would save time if we go that route.

colethorsen:
I’m not suggesting you put it back permenantly but by implementing this hotfix solves the problem short term so people can upgrade without breaking entire systems (the other functionality would have always been broken for them anyways) and then working on a long term solution that solves the problem would be secondary.

On Wed, Jun 2, 2021 at 8:50 AM, MGatner ***@***.***> wrote:


You also can't just remove ... that production applications are relying

Agreed. Like I said this was totally unintentional, but I believe that is because the feature was unintentional to begin with (or at least entirely undocumented/commented). We can make this a priority but I'm not comfortable making this call by myself. If you have time and capacity to flesh out this PR to be a proper "allow subdirectories" emendation it would save time if we go that route.

MGatner:
We for sure won't be hotfixing (i.e. pushing to master and re-releasing). If you mean providing this solution in the repo you can do this with the content you already made by pointing at your branch. Update **composer.json** as follows:
```
	"require": {
		"php": "^7.3 || ^8.0",
		"codeigniter4/codeigniter4": "dev-feature/fix-file-cache",
	},
	"repositories": [
		{
			"type": "vcs",
			"url": "https://github.com/colethorsen/codeigniter4"
		}
	]
```

The forums are back up. If you PM me your Slack email I will invite you directly.

paulbalandan:
Superseded by #5008

PR #4766, opened 2021-06-02
[Rector] Remove @var from class constant

samsonasik:
I created new custom rector rule `RemoveVarTagFromClassConstantRector` to remove `@var` from class const per this discussion https://github.com/codeigniter4/CodeIgniter4/pull/4759#issuecomment-852125720

**Checklist:**
- [x] Securely signed commits


michalsn (approved):

MGatner (approved):

kenjis:
Why don't we remove this Rector?
I don't see the need to delete `@var` that someone wrote.

> The `@var` tag defines which type of data is represented by a value of a Constant, Property or Variable.
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#518-var

MGatner:
I don't have a strong opinion. My original point was that the `@var` tag is unnecessary on constants because their type is apparent at runtime. If it's helpful for devs to indicate that explicitly on the code then a comment could always be used without the tag. 

I suppose an argument could be made for non-native types, like:
```php
/**
 * @var class-string<Model>
 */
public const $model = Widget::class;
```

PR #4769, opened 2021-06-03
Expand Query named binds recognition

### LINKED ISSUE #4760: Bug: Error using SQLITE3 strftime in CodeIgniter 4.1.2
**Describe the bug**
This line was working perfectly before the update to CodeIgniter 4.1.2
`->where('strftime("%Y-%m", release)', date('Y-m'))`
Now I have the following error:
`SQLite3::query(): Unable to prepare statement: 1, unrecognized token: &quot;:strftime('%Y-%m',&quot;`

**CodeIgniter 4 version**
4.1.2

**Context**
 - OS: Fedora 34
 - Web server: Running directly on Spark server not in production
 - PHP version: 7.4.19

paulbalandan:
**Description**
Another pass to fixing named binds recognition by adding a negative lookahead for `:=` types.

Fixes #4760 .
Supersedes and closes #4764 

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -315,13 +315,13 @@ class Query implements QueryInterface
 	 *
 	 * @return void
 	 *
-	 * @see https://regex101.com/r/EUEhay/1 Test
+	 * @see https://regex101.com/r/EUEhay/4
 	 */
 	protected function compileBinds()
 	{
 		$sql = $this->finalQueryString;
 
-		$hasNamedBinds = preg_match('/:[a-z\d.)_(]+:/i', $sql) === 1;
+		$hasNamedBinds = preg_match('/:((?!=).+):/', $sql) === 1;
 
 		if (empty($this->binds)
 			|| empty($this->bindMarker)


MGatner (approved):
I don't feel qualified to assess the underlying issue and solution, but the code is fine.

system/Database/Query.php
```diff
@@ -315,13 +315,13 @@ public function getOriginalQuery(): string
 	 *
 	 * @return void
 	 *
-	 * @see https://regex101.com/r/EUEhay/1 Test
+	 * @see https://regex101.com/r/qvzPaZ/1
```
samsonasik, line 318: can use versioned regex101 uri: 

```suggestion
       * @see https://regex101.com/r/EUEhay/3
```

samsonasik (approved):

vlakoff:
The outer parentheses are useless, so the code could be slightly simplified to:

```php
preg_match('/:(?!=).+:/', $sql)
```

paulbalandan:
That's true, as we're not saving the matches. If you like, you can send a PR to simplify it.

kenjis:
This fix is very difficult to understand. Because the bug is related to Query Builder.

When `$builder->where('id', 1)`, Query Builder creates a query like `WHERE "id" = :id:` inside.

When `$builder->where('LOWER(jobs.name)', 'accountant')`,
it creates `WHERE LOWER(jobs.name) = :LOWER(jobs.name):`.

And after all `Query::compileBinds()` binds the value.

Query Builder uses Named Bindings.

PR #4771, opened 2021-06-03
Convert tabs to spaces

paulbalandan:
**Description**
Depends on #4770 .

This enables the `indentation_type` rule to convert usages of tabs to 4 spaces for indentations.

Note: Changes are purely conversion of tabs to spaces as generated by php-cs-fixer. No other manual changes done except for the errant `\App\Config\App` used in `Config\Reader` in `admin/module/`.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide


sfadschm:
So CI4 will now be using spaces for both indentation and alignment?

.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,34 @@
...
+use Utils\PhpCsFixer\CodeIgniter4;
+
+$finder = Finder::create()
+    ->files()
+    ->in([
+        __DIR__ . '/app', // @todo relocate when `header_comment` is enabled
+        __DIR__ . '/public', // @todo relocate when `header_comment` is enabled
+        __DIR__ . '/system',
+        __DIR__ . '/tests',
+        __DIR__ . '/utils',
```
samsonasik, line 16: The https://github.com/codeigniter4/CodeIgniter4/blob/develop/admin/starter/app/Config/Paths.php need to be added as well?

paulbalandan: Done.

paulbalandan:
> So CI4 will now be using spaces for both indentation and alignment?

Per discussion, we'll start adopting PSR-12.

lonnieezell (approved):
Looks good from what I can tell. My knowledge of Nexus is slim, but everything else seems appropriate.

samsonasik (approved):

samsonasik:
composer.json seems conflict with latest change by dependabot.

totoprayogo1916:
This should be update too @paulbalandan ?
https://github.com/codeigniter4/CodeIgniter4/blob/develop/admin/pre-commit

totoprayogo1916:
in CI4.1.3, framework & sppstarter still use tabs.

paulbalandan:
> in CI4.1.3, framework & sppstarter still use tabs.

This is intended. These changes will have a dedicated release.

lighth7015:
Honestly this doesn't seem like a good idea. For example accessibility will suffer (tabs allow users to set their preferred indentation according to the settings that make sense for them. Forcing spaces may have a detrimental effect for those people.

paulbalandan:
Hi! Thanks for your insight. However, this conversion to spaces applies only to the framework code. End users are **not forced** to convert their existing code to use spaces. They can still use their preference here. As such, I see no detrimental effect of this to end users as they are not directly affected. Again, this change is internal.

gnat:
Long looong time Code Igniter user here going back to 2008 with multiple major sites on Code Igniter. Got the EllisLab tatoo and all... This was a dumb change.

> Hi! Thanks for your insight. However, this conversion to spaces applies only to the framework code. End users are **not forced** to convert their existing code to use spaces. They can still use their preference here. As such, I see no detrimental effect of this to end users as they are not directly affected. Again, this change is internal.

Why even do it then? Seriously restrain yourself from making changes to just "pad out" your github man.

gnat:
On that note I am disliking this recent trend of people joining a project just to add/remove white space. It ends up being quite disruptive to those who use the projects.

kenjis:
@gnat You can see Paul's contributions to CodeIgniter4:
https://github.com/codeigniter4/CodeIgniter4/pulls?q=is%3Amerged+is%3Apr+author%3Apaulbalandan+

kenjis:
> Why even do it then?

It is because all maintainers agreed with switching to PSR-12.

lonnieezell:
I have also used CodeIgniter since pretty close to that same time, version 1.6 IIRC. Glad to see you still around! 

> I would highly advise a revert as it forces a style change of everything built on Code Igniter going back to V1.

This doesn't force a style change on anything except the internal code for CodeIgniter. Developers not contributing code to the framework itself are not affected at all. I understand tabs vs spaces is a hot topic for a lot of people. I used to be on the tabs side, myself. Over the years I've found that using a decent IDE makes it not matter anymore. 

> Why even do it then? Seriously restrain yourself from making changes to just "pad out" your github man.

We chose to adopt PSR-12 to take advantage of the many bits of tooling and automation that can come with that. We understand the PSR's are not for everyone, and they were never meant to be. They were specifically developed to aid in framework interoperability. 

> Nobody here modifies Code Igniter or writes add-ons? Your "contribution" just wonked up the entire Code Igniter ecosystem. What was the win?

Again - unless you are contributing code to the core of the framework, this really doesn't affect you. If you write addons you can format it however you choose. If you edit the core files, then there might be a small irritation and that's understandable. But there's usually better solutions than modifying the core for your project. 

You don't have to like spaces, which obviously you don't, and that's totally cool. Continuing using tabs in your own code and enjoy.

gnat:
Thank you for the chill responses. It's reassuring to hear this was by consensus of the current team of maintainers.

Yeah, Code Igniter is still one of my all time favorites. Glad to see people still passionate about it. Was pretty shocked to do a pull and see this major changeset on every file in the repo. It is what it is.

PR #4773, opened 2021-06-04
Update to phpdocumentor v3.0

### LINKED ISSUE #1988: Bug: API docs generation is incomplete
The [phpDocumentor2](http://www.phpdoc.org/) tool used to generate our API docs does not properly recognize all PHP7 syntax. The source scanning flags the source code as having a syntax error, and the class is not included in the API docs. This is expected to be addressed in phpDocumentor3, which is in alpha.3. I will monitor the tool, and close this once we have resolution, but wanted to document the problem in the meantime.

The syntax errors, and affected API doc classes are shown below.

PHP7 allows [previously reserved words to be used as method names](https://www.php.net/manual/en/reserved.keywords.php).
- CLI/CLI (unexpected T_PRINT, expecting T_STRING on line 307)
- View/Filters (unexpected T_DEFAULT, expecting T_STRING on line 113)
- View/View (unexpected T_INCLUDE, expecting T_STRING on line 499)

PHP7 provides type hinting, which messes up parsing of some of our source.
- Files/File (unexpected '?' on line 121)
- Helpers/number_helper (unexpected '?' on line 270)
- HTTP/FIles/UploadedFile (unexpected '?' on line 359)
- I18n/Time (unexpected '?', expecting T_VARIABLE on line 1039)
- Test/CUUnitTestCase (unexpected '?' on line 253)

PHP7 provides for array declarations on the left side of an assignment, sorta like the [list construct](https://www.php.net/manual/en/function.list.php), which has caused some parsing hiccups.
- Router/Router (unexpected '=' on line 472)
- Language/Language (unexpected '=' on line 149)

**EDIT: It sounds like version 3.0.0 of the tool will solve many (all?) of the issues**

paulbalandan:
**Description**
Fixes #1988 
Snippet of successful run: https://github.com/paulbalandan/CodeIgniter4/actions/runs/903865690

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/deploy-apidocs.yml
+++ b/.github/workflows/deploy-apidocs.yml
@@ -1,6 +1,6 @@
 # When changes are pushed to the develop branch,
 # build the current version of the API documentation
-# with phpDocumentor and deploy it to the api branch.
+# with phpDocumentor and deploy it to codeigniter4/api.
 name: Deploy API Documentation
 
 on:
@@ -9,13 +9,13 @@ on:
       - 'develop'
     paths:
       - 'system/**'
+      - '.github/workflows/deploy-apidocs.yml'
 
 jobs:
-
   build:
     name: Deploy to api
-    if: (github.repository == 'codeigniter4/CodeIgniter4')
-    runs-on: ubuntu-latest
+    if: github.repository == 'codeigniter4/CodeIgniter4'
+    runs-on: ubuntu-20.04
     steps:
       - name: Setup credentials
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-apidocs.yml
```diff
@@ -37,37 +37,48 @@ jobs:
...
 
-      - name: Deploy
+      - name: Deploy to API repo
+        working-directory: api
         run: |
-          cd ./api
           git add .
-          git commit -m "API sync for ${GITHUB_SHA}"
-          git push
+          git commit -S -m "Updated API for commit ${{ github.event.after }}"
```
MGatner, line 72: Are you sure this is the correct reference?

https://github.community/t/accessing-commit-message-in-pull-request-event/17158/8

paulbalandan: I have checked this but if that gotcha exists then it's better to use the sha directly. Reverted.

MGatner (approved):

MGatner:
🥳

PR #4781, opened 2021-06-06
Don't include `utils/` in the `.gitattributes`

paulbalandan:
**Description**
This allows modules/libraries depending on `codeigniter4/codeigniter4` to use its "utilities" for their own use.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,7 +21,6 @@ Vagrantfile.dist export-ignore
 
 # They don't want our test files
 tests/system/ export-ignore
-utils/ export-ignore
 rector.php export-ignore
 phpunit.xml.dist export-ignore
 phpstan.neon.dist export-ignore


paulbalandan:
Apparently this won't work as `Utils\` is not a registered namespace in autoload.

sfadschm:
Any other idea on how this could be achieved?
This would be really helpful instead of copying over all files by hand whenever they are updated here 😃 

Only thing that came to my mind was using `composer` scripts, but that does not work too well either.

paulbalandan:
This can be achieved (I think) by (1) removing it from `.gitattributes` and (2) registering the `Utils\` namespace in the main `autoload` section (instead of `autoload-dev`). However, I tried mocking this behavior in a library of mine but can't get it to work. It seems composer ignores manual changes in composer.json of a project's dependencies, so running `composer dump` makes no sense.

sfadschm:
Alright, this did not work for me, too for some reason.
Guess I will try copying them with composer scripts.

PR #4778, opened 2021-06-06
[CS Migration]Update indentation to spaces

paulbalandan:
**Description**
Continuation for #4771 

**Checklist:**
- [x] Securely signed commits

PR #4782, opened 2021-06-07
Update EditorConfig to match PSR-12

jeromegamez:
PSR-12 is already ensured through PHP CS Fixer, but editors with an EditorConfig plugin should do it correctly to begin with 💪 

:octocat:


jeromegamez:
Ooops, duplicates #4778, sorry!

PR #4784, opened 2021-06-07
Use static lambda if a binding to `$this` is not required.

jeromegamez:
When not declared `static`, anonymous functions are bound to/hold a reference to `$this`, which can prevent garbage collection. It can also bring additional performance improvements, e.g. as [here](https://github.com/Ocramius/GeneratedHydrator/releases/tag/3.0.0)

> Also, by using `static function () {}` closures, we squeezed out another 15% of hydration performance when dealing with private properties.

:octocat:

### DIFF
--- a/app/Config/Events.php
+++ b/app/Config/Events.php
@@ -22,7 +22,7 @@ use CodeIgniter\Exceptions\FrameworkException;
  *      Events::on('create', [$myInstance, 'myMethod']);
  */
 
-Events::on('pre_system', function () {
+Events::on('pre_system', static function () {
     if (ENVIRONMENT !== 'testing')
     {
         if (ini_get('zlib.output_compression'))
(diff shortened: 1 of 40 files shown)


lonnieezell (approved):
I don't know exactly how this change would affect existing developers, if at all. I'm guessing not at all, and the changes seem simple enough. I'll approve but I'd love to hear from @MGatner or @paulbalandan

paulbalandan (approved):
This is definitely one of the rules I want to enforce because of the optimizations it can give. 👍

jeromegamez:
Rebased branch to resolve merge conflicts 💪

paulbalandan:
@jeromegamez  can you rebase again? Thanks!

jeromegamez:
Done!

MGatner:
Wow, I had no idea. We should update our `Events` docs since those will always be in this category and should take advantage. Thanks @jeromegamez! 🥳

PR #4786, opened 2021-06-07
Define yoda_style rule and initially apply no_yoda style

jeromegamez:
I noticed @paulbalandan's https://github.com/codeigniter4/CodeIgniter4/pull/4442#discussion_r596571689 

> We don't do yoda conditions. :)

and thought: "While I'm at it, why rulified not shall it be" 😬 

:octocat:

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -240,7 +240,7 @@ class CodeIgniter
             spl_autoload_register(function ($class) {
                 $class = explode('\\', $class);
 
-                if ('Kint' !== array_shift($class))
+                if (array_shift($class) !== 'Kint')
                 {
                     return;
                 }
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1746,7 +1746,7 @@ class BaseBuilder
      */
     protected function _limit(string $sql, bool $offsetIgnore = false): string
     {
-        return $sql . ' LIMIT ' . (false === $offsetIgnore && $this->QBOffset ? $this->QBOffset . ', ' : '') . $this->QBLimit;
+        return $sql . ' LIMIT ' . ($offsetIgnore === false && $this->QBOffset ? $this->QBOffset . ', ' : '') . $this->QBLimit;
     }
 
     //--------------------------------------------------------------------
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1495,7 +1495,7 @@ abstract class BaseConnection implements ConnectionInterface
         $driver = strtolower($this->DBDriver);
         $driver = ($driver === 'postgre' ? 'pg' : $driver) . '_';
 
-        if (false === strpos($driver, $functionName))
+        if (strpos($driver, $functionName) === false)
         {
             $functionName = $driver . $functionName;
         }
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1171,7 +1171,7 @@ class RouteCollection implements RouteCollectionInterface
      */
     public function environment(string $env, Closure $callback): RouteCollectionInterface
     {
-        if (ENVIRONMENT === $env)
+        if ($env === ENVIRONMENT)
         {
             $callback($this);
         }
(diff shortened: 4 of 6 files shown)


paulbalandan (approved):

paulbalandan:
Kindly rebase again.

jeromegamez:
Done!

MGatner:
I dislike yoda conditions as well (though now I'm pumped that I know what they are called!). However the one case where I do think they are a) cleaner and b) safer is with assignment tests:
```
if (42 !== $userId = $this->request->getPost('user_id'))
{
    model(UserModel::class)->update($userId, $this->request->getPost());
}
```

Will these rules still allow for those cases?

paulbalandan:
We can set `identical` to `true` to allow for this condition, however, all cases of `===` and `!==` will be in yoda even if not involving assignments.
```diff
-$a === null;
+null === $a;

-($userId = $this->request->getPost('user_id')) !== 42
+42 !== $userId = $this->request->getPost('user_id')
```

MGatner:
I'm not totally sure what our options are but I want to be able to use the green version without fighting our styler:
```diff
-($userId = $this->request->getPost('user_id')) !== 42
+42 !== $userId = $this->request->getPost('user_id')
```
If that precludes using an otherwise-useful rule then I'm not sure what to do.

jeromegamez:
Yoda conditions were invented to avoid accidental variable assignments so that a `if ($userId = 42)` cannot happen, but I (personally) don't think they help readability/comprehensibility. "If this is 42" sounds more natural than "If 42 is this".

In that particular case, I would rather changing the code in the first place because of the mixture of comparison and assignment (`42 !== $userId = ...`), either to

```php
if (($userId = $this->request->getPost('user_id')) !== 42) {
    // ...
}
```

or 

```php
$userId = $this->request->getPost('user_id');

if ($userId !== 42) {
    // ...
}
```

I've seen both variants in the CI codebase already, and I don't think this would be worth softening the rule - especially since accidental assignments will surely be prevented by code reviews and PHPStan. But again, that's just my opinion! I'll gladly change the PHP CS Fixer rule if you tell me to! 💪

MGatner:
It looks to me like we can set `identical` to `null` and still use the other rules: https://php-cs-fixer-playground.herokuapp.com/run/081bf10f-090b-43bc-bd9f-0d2ee6e11ec4

Would that be an acceptable compromise? We can always enforce non-assignment identical clauses manually, and do an initial fix for them.

paulbalandan:
I forgot about the `null`. Setting `null` will force the fixer to ignore those conditions and do not make a fix. That's fine with me.

jeromegamez:
Done! I also reset the changes before re-applying the rule, which resulted in... no more changes 😂

I'll rename the PR title accordingly!

paulbalandan:
I think the previous changes are needed. No yoda in general except for mixed assignment and comparison in a single line. Then do manual reviews moving forward. @MGatner what do you think?

MGatner:
> no more changes

🤦‍♂️ Haha a lot of bluster for nothing! I like knowing this remains an option. I do use these, probably mostly in modules (which will follow this style guide). I also see at least one (mostly unnecessary) in the upcoming email rewrite: https://github.com/MGatner/CodeIgniter4/blob/4de1ddf1b417bb4e1292f0b458733d370357113e/system/Mailer/Address.php#L112

Thanks for making that change!

MGatner:
> No yoda in general except for mixed assignment and comparison in a single line. 

Yes, precisely. Now I see what you mean by "no changes"! let's re-apply with `identical = false` to catch those changes but then remove the rule prior to merging. Since @jeromegamez doesn't have anything better to do 🙃

jeromegamez:
@MGatner Could we meet in the middle and use

```diff
- if ('' === $name = trim($name, SELF::TRIM))
+ if ('' === ($name = trim($name, SELF::TRIM)))
```

🙈

jeromegamez:
> Since @jeromegamez doesn't have anything better to do 🙃

I actually do have a day off today 😅

Changes re-applied, commit message and PR title re-changed 💪

MGatner (approved):

PR #4787, opened 2021-06-07
Don't override `$path` parameter

jeromegamez:
`$path` is given as parameter and overridden in case `$namespace` is an array. While it doesn't make a difference when executing the method, it could introduce side effects in the future in case the method is changed and the override is overlooked 🤞 

:octocat:


paulbalandan (approved):

MGatner (approved):
Good catch! That was definitely a mistake.

PR #4790, opened 2021-06-07
Enable `blank_line_before_statement`

paulbalandan:
**Description**
This adds an empty line feed to precede the configured statements. This improves readability.

Available (and used) statements are:
```php
      //  'break',
        'case',
        'continue',
        'declare',
        'default',
        'do',
        'exit',
        'for',
        'foreach',
        'goto',
      // 'if',
      // 'include',
      // 'include_once',
      // 'require',
      // 'require_once',
        'return',
        'switch',
        'throw',
        'try',
        'while',
        'yield',
        'yield_from',
```

**Checklist:**
- [x] Securely signed commits


paulbalandan:
I updated and re-run to exclude `if` and `break` also.

MGatner (approved):

PR #4791, opened 2021-06-07
Add brackets to clarify intent and avoid unwanted side-effects

jeromegamez:
[Operator precedence](https://www.php.net/manual/en/language.operators.precedence.php) always is a headscratcher for me 😅 

:octocat:


MGatner (approved):
Most of these would probably merit rewriting more clarify anyway, but this is an improvement.

jeromegamez:
Agreed! I had already started rewriting some of them, but then I thought to myself "on thing at a time" (hence the many many small PRs) 🙈

paulbalandan (approved):

PR #4793, opened 2021-06-07
Use constants instead of functions

jeromegamez:
Using constants instead of methods saves some OPCode steps 💪 

[`PHP_OS_FAMILY` has a smaller, predefined set of possible values](https://www.php.net/manual/en/reserved.constants.php#constant.php-os-family)

:octocat:


paulbalandan:
Can you also enable the `function_to_constant` rule here?

jeromegamez:
Good call! Done!

paulbalandan (approved):

MGatner (approved):

PR #4794, opened 2021-06-07
Replace alias functions with origins

jeromegamez:
In order to avoid surprises (e.g. `mysqli_*` aliases were dropped in PHP 5.4 or `magic_quotes_runtime` was dropped in PHP 7.0), we should not rely on alias functions at all.

:octocat:


paulbalandan:
We can use the `no_alias_functions` here. With `['sets' => '@all']` to cover all aliased versions.

jeromegamez:
Done! 🚀

paulbalandan (approved):

MGatner (approved):
Wow PhpCsFixer is awesome, a rule for everything! Also, who knew `join()` existed? Not I.

PR #4795, opened 2021-06-07
Remove removed `safe_mode` ini Option

jeromegamez:
`safe_mode` was removed in PHP 5.4 🙈 

:octocat:

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -793,8 +793,8 @@ if (! function_exists('is_really_writable'))
      */
     function is_really_writable(string $file): bool
     {
-        // If we're on a Unix server with safe_mode off we call is_writable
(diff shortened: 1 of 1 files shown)


system/Common.php
```diff
@@ -794,7 +794,7 @@ function is_cli(): bool
     function is_really_writable(string $file): bool
     {
         // If we're on a Unix server with safe_mode off we call is_writable
```
MGatner, line 796:
```suggestion
        // If we're on a Unix server we call is_writable
```

MGatner (changes requested):
Drop the comment reference too

system/Common.php
jeromegamez: Eagle eye MGatner! Fixed, thanks!

paulbalandan (approved):

MGatner (approved):

PR #4796, opened 2021-06-07
Use null coalescing operator

jeromegamez:
The null coalescing operator saves code and reduces cognitive overload 💆 

:octocat:

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1707,12 +1707,7 @@ abstract class BaseModel
             return $this->$name;
         }
 
-        if (isset($this->db->$name))
-        {
-            return $this->db->$name;
-        }
-
-        return null;
+        return $this->db->$name ?? null;
     }
 
     /**
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -2363,7 +2363,12 @@ class Email
             return $_SERVER['SERVER_NAME'];
         }
 
-        return isset($_SERVER['SERVER_ADDR']) ? '[' . $_SERVER['SERVER_ADDR'] . ']' : '[127.0.0.1]';
+        if (isset($_SERVER['SERVER_ADDR']))
+        {
+            return '[' . $_SERVER['SERVER_ADDR'] . ']';
+        }
+
+        return '[127.0.0.1]';
     }
 
     /**
--- a/system/Test/Mock/MockCache.php
+++ b/system/Test/Mock/MockCache.php
@@ -53,9 +53,7 @@ class MockCache extends BaseHandler implements CacheInterface
     {
         $key = static::validateKey($key, $this->prefix);
 
-        return array_key_exists($key, $this->cache)
-            ? $this->cache[$key]
-            : null;
+        return $this->cache[$key] ?? null;
     }
 
     //--------------------------------------------------------------------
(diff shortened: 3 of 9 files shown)


paulbalandan (approved):

system/Email/Email.php
```diff
@@ -2358,12 +2358,16 @@ protected function getSMTPData()
      */
     protected function getHostname()
     {
-        if (isset($_SERVER['SERVER_NAME']))
+        if ($serverName = $_SERVER['SERVER_NAME'] ?? null)
         {
-            return $_SERVER['SERVER_NAME'];
+            return $serverName;
         }
```
MGatner, line 2364: This one also is a logic change from anything in `SERVER_NAME` that was falsy. For example:
```
$_SERVER['SERVER_NAME'] = '';

// Previously
echo $this->getHostname(); // ''

// Now
echo $this->getHostname(); // '[127.0.0.1]'
```

system/Test/Mock/MockCache.php
```diff
@@ -53,9 +53,7 @@ public function get(string $key)
     {
         $key = static::validateKey($key, $this->prefix);
 
-        return array_key_exists($key, $this->cache)
-            ? $this->cache[$key]
-            : null;
+        return $this->cache[$key] ?? null;
```
MGatner, line 56: Took me a minute, but this one is kosher :)

MGatner (changes requested):

system/BaseModel.php
```diff
@@ -1702,17 +1702,7 @@ protected function transformDataToArray($data, string $type): array
      */
     public function __get(string $name)
     {
-        if (property_exists($this, $name))
-        {
-            return $this->$name;
-        }
```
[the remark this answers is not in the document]
jeromegamez: Good catch!

system/Email/Email.php
jeromegamez: Is this good or bad? I would argue that yes, it's a logic change, but an empty host name would definitely be invalid and I assume unwanted 🤔 - what should I do here?

jeromegamez:
I ❤️ code reviews!

system/Email/Email.php
MGatner: Let's keep the logic for now, since we're really focused on styles for this release. I will open an Issue so we come back and take a look at what the expected and desired output form this is. FWIW I totally agree that the current logic is not ideal.

MGatner: Actually, when I went to make the note I noticed this is in `Email`, which is being replaced soon anyway, so I don't really care either way :P

jeromegamez: I changed it back to `isset()` but rewrote it slightly to avoid the ternary operator for readability (it does more than just `$a ? $a : $b`),

MGatner (approved):

samsonasik:
Please update `rector.php` and add the following:

```php
use Rector\CodeQuality\Rector\Ternary\ArrayKeyExistsTernaryThenValueToCoalescingRector;

$services->set(ArrayKeyExistsTernaryThenValueToCoalescingRector::class);
```

and run:

```bash
vendor/bin/rector process --clear-cache
```

to ensure all places applied for array_key_exists ternary then value coalescing transformation and rector will take care of this check in the future.

samsonasik:
Oh, just see there is `ternary_to_null_coalescing` config for that, so the rector rule for it seems unneeded. 

Just need rebase then.

jeromegamez:
Done! 💪

samsonasik (approved):

PR #4797, opened 2021-06-07
Ensure single space between cast and variable

paulbalandan:
**Description**
This enables the `cast_spaces` fixer and set to ensure there's a single space between the cast and the casted variable.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Session/Handlers/DatabaseHandler.php
+++ b/system/Session/Handlers/DatabaseHandler.php
@@ -295,9 +295,10 @@ class DatabaseHandler extends BaseHandler
      */
     public function gc($maxlifetime): bool
     {
-        $interval = implode(" '"[(int)($this->platform === 'postgre')], ['', "{$maxlifetime} second", '']);
+        $separator = $this->platform === 'postgre' ? '\'' : ' ';
+        $interval  = implode($separator, ['', "{$maxlifetime} second", '']);
 
-        return ($this->db->table($this->table)->delete("timestamp < now() - INTERVAL {$interval}")) ? true : $this->fail();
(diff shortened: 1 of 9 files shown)


system/Session/Handlers/DatabaseHandler.php
```diff
@@ -317,7 +317,7 @@ public function destroy($sessionID): bool
      */
     public function gc($maxlifetime): bool
     {
-        $interval = implode(" '"[(int)($this->platform === 'postgre')], ['', "{$maxlifetime} second", '']);
+        $interval = implode(" '"[(int) ($this->platform === 'postgre')], ['', "{$maxlifetime} second", '']);
```
MGatner, line 320: The cast is approved, but... what is this mess? Holy cow that's hard to read.

MGatner (approved):

system/Session/Handlers/DatabaseHandler.php
paulbalandan: When I'm manually checking the changes I also did not understand this witchcraft. 😂

paulbalandan: Apparently, this is perfectly legal in all PHP versions. https://3v4l.org/Bkf7k
That's news for me.

paulbalandan: Ah I get it. It is array access on a string. 😅

paulbalandan: I reworked this to get explicitly the `$glue` first.

lonnieezell (approved):

PR #4802, opened 2021-06-08
Phpdoc param return types

paulbalandan:
**Description**
Changes to PHPDocs:
- use short forms (`bool` instead of `boolean`, etc.)
- Align phpdocs
- Separate and group similar phpdocs (removes Laravel-like docblocks)

**Checklist:**
- [x] Securely signed commits


MGatner (approved):
So much yes!

samsonasik (approved):

PR #4820, opened 2021-06-10
Initial Phpdoc-related rules

paulbalandan:
**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Config/Logger.php
+++ b/app/Config/Logger.php
@@ -125,7 +125,7 @@ class Logger extends BaseConfig
             'path' => '',
         ],
 
-        /**
+        /*
          * The ChromeLoggerHandler requires the use of the Chrome web browser
          * and the ChromeLogger extension. Uncomment this block to use it.
          */
@@ -137,7 +137,7 @@ class Logger extends BaseConfig
         //                   'error', 'info', 'notice', 'warning'],
         // ],
 
-        /**
+        /*
          * The ErrorlogHandler writes the logs to PHP's native `error_log()` function.
          * Uncomment this block to use it.
          */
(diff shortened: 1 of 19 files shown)


app/Config/Logger.php
```diff
@@ -125,7 +125,7 @@ class Logger extends BaseConfig
             'path' => '',
         ],
 
-        /**
```
MGatner, line 128: Why are these happening?

system/API/ResponseTrait.php
```diff
@@ -158,8 +158,8 @@ public function fail($messages, int $status = 400, string $code = null, string $
     /**
      * Used after successfully creating a new resource.
      *
-     * @param mixed  $data    Data.
-     * @param string $message Message.
+     * @param mixed  $data    data
+     * @param string $message message
```
MGatner, line 162: These seem silly. Is there perhaps a rule for "delete useless comments", like when `comment == property name`?

MGatner (changes requested):
Generally these are good. This biggest stylistic opinion seems to be whether we treat phpdoc comments as sentences (i.e. capitalization and punctuation). I generally do write them with full proper sentence syntax - not a strong opinion here, but I'm curious if the choice to go the opposite direction is based on some other standards?

Also I commented inline, but it would be great to remove comments that basically just duplicate their property or method name.

app/Config/Logger.php
paulbalandan: This is a hack actually on my part. Since the underlying code is commented out, php-cs-fixer treats the documentable structural element as the parent element.

system/API/ResponseTrait.php
paulbalandan: I don't think there is a single fixer for this. But I think we can achieve this by removing then re-adding the phpdocs (which php-cs-fixer has).

paulbalandan:
I'll break the for-resolution fixers in a separate PR.

MGatner (approved):

PR #4823, opened 2021-06-11
Phpdoc types

paulbalandan:
**Description**
- Use correct case for phpdoc types
- Sorts phpdoc types by alpha with `null` always last

**Checklist:**
- [x] Securely signed commits


MGatner (changes requested):
Is there a way to make `mixed` come last (or penultimate if there is a `null`) when it is listed with others?

paulbalandan:
> Is there a way to make `mixed` come last (or penultimate if there is a `null`) when it is listed with others?

I'm afraid none. Sort algorithm options are only `alpha` and `none` (no sorting).

MGatner:
No worries, was but a high-brow request 😂

MGatner (approved):

PR #4830, opened 2021-06-13
FileCollection

MGatner:
**Description**
This PR extracts a lot of the file operations from the recent `Publisher` PR into a dedicated class: `FileCollection`. `Publisher` then becomes an extension of this class, while allowing the core functionality to stand on its own for gather files and `File` objects from across the filesystem.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Files/FileCollection.php
+++ b/system/Files/FileCollection.php
@@ -45,7 +45,7 @@ class FileCollection implements Countable, IteratorAggregate
 	 *
 	 * @throws FileException
 	 */
-	protected static function resolveDirectory(string $directory): string
+	final protected static function resolveDirectory(string $directory): string
 	{
 		if (! is_dir($directory = set_realpath($directory)))
 		{
@@ -65,7 +65,7 @@ class FileCollection implements Countable, IteratorAggregate
 	 *
 	 * @throws FileException
 	 */
-	protected static function resolveFile(string $file): string
+	final protected static function resolveFile(string $file): string
 	{
 		if (! is_file($file = set_realpath($file)))
 		{
@@ -84,7 +84,7 @@ class FileCollection implements Countable, IteratorAggregate
 	 *
 	 * @return string[]
 	 */
-	protected static function filterFiles(array $files, string $directory): array
+	final protected static function filterFiles(array $files, string $directory): array
 	{
 		$directory = self::resolveDirectory($directory);
 
@@ -101,7 +101,7 @@ class FileCollection implements Countable, IteratorAggregate
 	 *
 	 * @return string[]
 	 */
-	protected static function matchFiles(array $files, string $pattern): array
+	final protected static function matchFiles(array $files, string $pattern): array
 	{
 		// Convert pseudo-regex into their true form
 		if (@preg_match($pattern, null) === false) // @phpstan-ignore-line
--- a/system/Files/FileCollection.php
+++ b/system/Files/FileCollection.php
@@ -185,10 +185,7 @@ class FileCollection implements Countable, IteratorAggregate
 	 */
 	public function add($paths, bool $recursive = true)
 	{
-		if (! is_array($paths))
-		{
-			$paths = [$paths];
-		}
+		$paths = (array) $paths;
 
 		foreach ($paths as $path)
 		{
--- a/system/Files/FileCollection.php
+++ b/system/Files/FileCollection.php
@@ -35,6 +35,8 @@ class FileCollection implements Countable, IteratorAggregate
 	protected $files = [];
 
 	//--------------------------------------------------------------------
+	// Support Methods
+	//--------------------------------------------------------------------
 
 	/**
 	 * Resolves a full path and verifies it is an actual directory.
@@ -104,7 +106,7 @@ class FileCollection implements Countable, IteratorAggregate
 	final protected static function matchFiles(array $files, string $pattern): array
 	{
 		// Convert pseudo-regex into their true form
-		if (@preg_match($pattern, null) === false) // @phpstan-ignore-line
+		if (@preg_match($pattern, '') === false)
 		{
 			$pattern = str_replace(
 				['#', '.', '*', '?'],
@@ -119,6 +121,8 @@ class FileCollection implements Countable, IteratorAggregate
 		});
 	}
 
+	//--------------------------------------------------------------------
+	// Class Core
 	//--------------------------------------------------------------------
 
 	/**
@@ -210,6 +214,8 @@ class FileCollection implements Countable, IteratorAggregate
 	}
 
 	//--------------------------------------------------------------------
+	// File Handling
+	//--------------------------------------------------------------------
 
 	/**
 	 * Verifies and adds files to the list.
@@ -269,6 +275,8 @@ class FileCollection implements Countable, IteratorAggregate
 	}
 
 	//--------------------------------------------------------------------
+	// Directory Handling
+	//--------------------------------------------------------------------
 
 	/**
 	 * Verifies and adds files from each
@@ -318,6 +326,8 @@ class FileCollection implements Countable, IteratorAggregate
 	}
 
 	//--------------------------------------------------------------------
+	// Filtering
+	//--------------------------------------------------------------------
 
 	/**
 	 * Removes any files from the list that match the supplied pattern
@@ -365,6 +375,8 @@ class FileCollection implements Countable, IteratorAggregate
 		return $this->removeFiles(array_diff($files, self::matchFiles($files, $pattern)));
 	}
 
+	//--------------------------------------------------------------------
+	// Interface Methods
 	//--------------------------------------------------------------------
 
     /**
--- a/system/Files/FileCollection.php
+++ b/system/Files/FileCollection.php
@@ -122,7 +122,7 @@ class FileCollection implements Countable, IteratorAggregate
 	//--------------------------------------------------------------------
 
 	/**
-	 * Loads the Filesystem helper and stores initial files.
+	 * Loads the Filesystem helper and adds any initial files.
 	 *
 	 * @param string[] $files
 	 */
@@ -130,7 +130,17 @@ class FileCollection implements Countable, IteratorAggregate
 	{
 		helper(['filesystem']);
 
-		$this->set($files);
+		$this->add($files)->define();
+	}
+
+	/**
+	 * Applies any initial inputs after the constructor.
+	 * This method is a stub to be implemented by child classes.
+	 *
+	 * @return void
+	 */
+	protected function define(): void
+	{
 	}
 
 	/**
--- a/system/Files/FileCollection.php
+++ b/system/Files/FileCollection.php
@@ -88,7 +88,7 @@ class FileCollection implements Countable, IteratorAggregate
 	{
 		$directory = self::resolveDirectory($directory);
 
-		return array_filter($files, function ($value) use ($directory) {
+		return array_filter($files, static function (string $value) use ($directory): bool {
 			return strpos($value, $directory) === 0;
 		});
 	}
@@ -114,7 +114,7 @@ class FileCollection implements Countable, IteratorAggregate
 			$pattern = "#{$pattern}#";
 		}
 
-		return array_filter($files, function ($value) use ($pattern) {
+		return array_filter($files, static function ($value) use ($pattern) {
 			return (bool) preg_match($pattern, basename($value));
 		});
 	}
@@ -150,7 +150,7 @@ class FileCollection implements Countable, IteratorAggregate
 	 */
 	public function get(): array
 	{
-		$this->files = array_unique($this->files, SORT_STRING);
+		$this->files = array_unique($this->files);
 		sort($this->files, SORT_STRING);
 
 		return $this->files;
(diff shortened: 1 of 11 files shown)


system/Files/FileCollection.php
```diff
@@ -0,0 +1,386 @@
...
+	 * Removes files that are not part of the given directory (recursive).
+	 *
+	 * @param string[] $files
+	 * @param string   $directory
+	 *
+	 * @return string[]
+	 */
+	protected static function filterFiles(array $files, string $directory): array
+	{
+		$directory = self::resolveDirectory($directory);
```
samsonasik, line 89: since the class is not final, if the `resolveDirectory()` is `protected`, should it be called via `static::` ?

MGatner: Hmm good point. This moved from Publisher where it was private, but since Publisher extends FileCollection and still needs that method I loosed it to protected. Maybe these static support functions should all be final? Or use the static call like you said?

If anyone has thoughts on the scoping in this class I'm open to suggestions.

lonnieezell: To be honest, I've never really been a fan of final classes/methods in a framework since that limits the end-users ability to extend the files. Or maybe I've never been confident enough that my ability to see the future was good enough lol. Always seemed like a better feature for applications than a framework. 

For a class like this, I'm not against it, though, since it functions as a security layer of sorts. Though people could maliciously insert a mock version of the class with new features? NVM that's only available during testing I believe. Either way, we can't stop everything. 

tl;dr; I honestly have no preference for whether this is a final class or not, but am not opposed if that's what others prefer.

MGatner: > maybe I've never been confident enough that my ability to see the future was good enough lol

This is actually why I'm a fan of `final` in many cases! We can be certain that nobody has extended the class/method so it is safe to make more changes than usual, and there is no penalty for removing `final` down the road.

In this case I decided to make the static helper methods `final` since they are essentially the equivalent of a new helper file - should this become an issue we can remove the `final` and call them with `static::` later.

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+	}
+
+	/**
+	 * Optimizes and returns the current file list.
+	 *
+	 * @return string[]
+	 */
+	public function get(): array
+	{
+		$this->files = array_unique($this->files, SORT_STRING);
```
paulbalandan, line 153:
```suggestion
		$this->files = array_unique($this->files);
```
`SORT_STRING` is the default flag.

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+	 *
+	 * @param string[] $files
+	 * @param string   $pattern Regex or pseudo-regex string
+	 *
+	 * @return string[]
+	 */
+	final protected static function matchFiles(array $files, string $pattern): array
+	{
+		// Convert pseudo-regex into their true form
+		if (@preg_match($pattern, null) === false) // @phpstan-ignore-line
```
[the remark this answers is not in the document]
MGatner: From a bit of research this is the best way to validate whether a string is an acceptable regex.

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+	 * @param string $file
+	 *
+	 * @return $this
+	 */
+	public function removeFile(string $file)
+	{
+		return $this->removeFiles([$file]);
+	}
+
+	//--------------------------------------------------------------------
```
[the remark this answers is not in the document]
MGatner: I love the dividers! But if you want them gone I can strip them. I like the way they allow for grouping methods/properties across large classes visually.

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+	 * @param string[] $files
+	 * @param string   $directory
+	 *
+	 * @return string[]
+	 */
+	final protected static function filterFiles(array $files, string $directory): array
+	{
+		$directory = self::resolveDirectory($directory);
+
+		return array_filter($files, function ($value) use ($directory) {
```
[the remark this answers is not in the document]
MGatner: I forgot about those! This will still all be run through our new rules once they are ready.

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+		{
+			$pattern = str_replace(
+				['#', '.', '*', '?'],
+				['\#', '\.', '.*', '.'],
+				$pattern
+			);
+			$pattern = "#{$pattern}#";
+		}
+
+		return array_filter($files, function ($value) use ($pattern) {
```
MGatner, line 117:
```suggestion
		return array_filter($files, static function ($value) use ($pattern) {
```

system/Files/FileCollection.php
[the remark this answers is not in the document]
lonnieezell: I still like "headers" with those where the sections are titled, just not a fan of them separating every method anymore. 

```
// Scope Functions
```

And maybe that's overkill but sounds like it serves the same purpose to what you were doing here. I am open to discussion on this, of course. Especially for the use case you were doing. I didn't realize it was a section divider since it looked the same as the ones I originally carried over from CI3 code, that most IDE's now have the option to insert for you I think.

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+	/**
+	 * Resolves a full path and verifies it is an actual directory.
+	 *
+	 * @param string $directory
+	 *
+	 * @return string
+	 *
+	 * @throws FileException
+	 */
+	final protected static function resolveDirectory(string $directory): string
```
samsonasik, line 48: `final` method doesn't need `protected`:

```suggestion
	final static function resolveDirectory(string $directory): string
```

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+	/**
+	 * Resolves a full path and verifies it is an actual file.
+	 *
+	 * @param string $file
+	 *
+	 * @return string
+	 *
+	 * @throws FileException
+	 */
+	final protected static function resolveFile(string $file): string
```
samsonasik, line 68: `final` method doesn't need `protected` as cannot be overriden, I think it can be private:

```suggestion
	final private static function resolveFile(string $file): string
```

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+
+	/**
+	 * Removes files that are not part of the given directory (recursive).
+	 *
+	 * @param string[] $files
+	 * @param string   $directory
+	 *
+	 * @return string[]
+	 */
+	final protected static function filterFiles(array $files, string $directory): array
```
samsonasik, line 87: `final` method doesn't need `protected` as cannot be overriden, I think it can be private:

```suggestion
	final private static function filterFiles(array $files, string $directory): array
```

system/Files/FileCollection.php
```diff
@@ -0,0 +1,396 @@
...
+
+	/**
+	 * Returns any files whose `basename` matches the given pattern.
+	 *
+	 * @param string[] $files
+	 * @param string   $pattern Regex or pseudo-regex string
+	 *
+	 * @return string[]
+	 */
+	final protected static function matchFiles(array $files, string $pattern): array
```
samsonasik, line 104: `final` method doesn't need `protected` as cannot be overriden, I think it can be private:

```suggestion
	final private static function matchFiles(array $files, string $pattern): array
```

system/Files/FileCollection.php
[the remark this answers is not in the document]
paulbalandan:
```suggestion
		if (@preg_match($pattern, '') === false)
```
Can we use empty string instead of `null` in subject?

system/Files/FileCollection.php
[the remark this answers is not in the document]
paulbalandan: I prefer dividers for "header" like purposes, but I use `=` for "bolder" appeal.
```
//====================================
// The header name
//====================================
```

system/Files/FileCollection.php
paulbalandan: Why not just change to `private` so it can't be accessed by child classes and no need for `final`?

samsonasik: @paulbalandan I just realized that 👍

system/Files/FileCollection.php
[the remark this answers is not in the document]
MGatner: I'm not sure if there was some trick to the original article I said that had `null`. I can't find it again now. If this seems to work as well that would certainly be better.

system/Files/FileCollection.php
MGatner: It still needs to be accessible from extends classes, even if it is not extended itself.

system/Files/FileCollection.php
MGatner: `Publisher` needs this method (which is where it came from) so when it moved to `FileCollection` I changed it from private to protected. We could make it private and give `Publisher` its own version, but I think it likely that if one child needs it then others might. 

An alternate approach would be to make these support methods into functions and add them to Filesystem Helper. I generally shy away from helper bloat but at least some of these seem generic enough.

system/Files/FileCollection.php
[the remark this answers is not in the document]
MGatner: So in this case do you believe pushing forward to headers is in order, or pull the dividers altogether?

lonnieezell:
I think making it a header is fine

MGatner:
I just realized we have a `FileCollection` in the HTTP implementation: **system/HTTP/Files/FileCollection.php**. Is that going to be too confusing?

lonnieezell:
> I just realized we have a FileCollection in the HTTP implementation: system/HTTP/Files/FileCollection.php. Is that going to be too confusing?

If they both have similar functions within them we should definitely try to combine, otherwise I think it's fine. The HTTP one deals with uploaded files so most likely completely different functionality.

MGatner:
> most likely completely different functionality

Precisely this, they have no overlap. I will proceed.

system/Files/FileCollection.php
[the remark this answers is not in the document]
MGatner: I did a little more research and testing and I believe your proposal works fine and is better in all ways.

system/Files/FileCollection.php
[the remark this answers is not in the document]
MGatner: I removed some dividers and switched the remaining to headers. I stuck with the single dash for now since that is what we have elsewhere in **system** but I wouldn't be opposed to a sweeping redo around dividers in general.

MGatner:
I believe this is ready again. In my opinion the scopes are all correct, but I'm still open to other suggestions.

system/Language/en/Files.php
```diff
@@ -11,6 +11,8 @@
 
 // Files language settings
 return [
-    'fileNotFound' => 'File not found: {0}',
-    'cannotMove'   => 'Could not move file {0} to {1} ({2}).',
+    'fileNotFound'      => 'File not found: {0}',
+    'cannotMove'        => 'Could not move file {0} to {1} ({2}).',
+	'expectedDirectory' => '{0} expects a valid directory.',
+	'expectedFile'      => '{0} expects a valid file.',
```
samsonasik, line 17: spacing need an enhancement?

MGatner: I'll run this whole thing (actually all of 4.2 branch) through CS Fixer after we release it. You can ignore code standard issues for now.

system/Files/FileCollection.php
```diff
@@ -0,0 +1,408 @@
...
+	 * @param boolean         $recursive
+	 *
+	 * @return $this
+	 */
+	public function add($paths, bool $recursive = true)
+	{
+		if (! is_array($paths))
+		{
+			$paths = [$paths];
+		}
```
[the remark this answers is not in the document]
MGatner: 😳 I never knew you could do this! We have this a bunch of places in the framework. Any chance there is a CS Fixer or Rector rule for it?

tests/system/Files/FileCollectionTest.php
```diff
@@ -0,0 +1,568 @@
...
+	{
+		$collection = new class([$this->file]) extends FileCollection {
+
+			protected function define(): void
+			{
+				$this->add(SUPPORTPATH . 'Files/baker/banana.php');
+			}
+		};
+
+		$this->assertSame([$this->file], $collection->get());
```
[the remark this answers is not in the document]
MGatner: This was a copy-paste error - the constructor should have no arguments so it is apparent that the file comes from `define()`. I will fix that.

tests/system/Files/FileCollectionTest.php
```diff
@@ -0,0 +1,568 @@
...
+				SUPPORTPATH . 'Files/able/apple.php',
+			];
+		};
+
+		$this->assertSame($expected, $collection->get());
+	}
+
+	public function testConstructorCallsDefine()
+	{
+		$collection = new class([$this->file]) extends FileCollection {
```
MGatner, line 121:
```suggestion
		$collection = new class() extends FileCollection {
```

paulbalandan (approved):

PR #4839, opened 2021-06-15
Define initial "m" rules

paulbalandan:
**Description**
- Ensures all PHP magic constants (`__FILE__`, `__DIR__`, etc.) and methods (`__sleep()`, `__get()`, etc.) are written in the correct case.
- Sets the `mb_str_functions` fixer as `false` (off) by default

**Checklist:**
- [x] Securely signed commits


MGatner (approved):
I didn't even know you could use lowercase magic constants. Sometimes PHP makes me 🙄

PR #4840, opened 2021-06-15
PSR12: Enable `method_argument_space` fixer

paulbalandan:
**Description**
When calling functions or class methods and the arguments are written multiline, ensure that all arguments are on their own line.

I have manually changed those I believe would be more visually appealing if single line written.

**Checklist:**
- [x] Securely signed commits


MGatner (approved):
This one has a few I'm not thrilled about, but I think it is mostly the way the functions are written and not the rule's fault. Let's proceed and then we can do some more manual cleanup as we come across things. Unlike some other frameworks, I will very happily break a statement up into two separate steps to improve readability.

PR #4844, opened 2021-06-17
Make CLI detection as interface independent as possible

### LINKED ISSUE #4843: Bug: Rules for is_cli() are too restrictive
**Describe the bug**
I came across this because some benchmarks stopped working: https://the-benchmarker.semaphoreci.com/jobs/21790820-d57f-4457-834e-2f93bdddef91. No matter what URL we would enter, the default route would be fired instead.

I started looking at it closely and figured out that new rules in the `is_cli()` function are the reason - exactly these:
```php
if (stristr(PHP_SAPI, 'cgi') && getenv('TERM')) {
    return true;
}
```
Values for these are:
```
PHP_SAPI => fpm-fcgi
getenv('TERM') => xterm
```
Here is a Dockerfile: https://github.com/the-benchmarker/web-frameworks/blob/master/php/Dockerfile it can be tested by anyone.

**CodeIgniter 4 version**
4.1.3 and above

**Affected module(s)**
`is_cli()` function

Sorry if this is a little chaotic - I can provide more information later if needed.

### LINKED ISSUE #4848: Undefined constant "CodeIgniter\CLI\STDOUT"
CodeIgniter Version: 4.1.3
PHP Version: 8.0.6
PHP SAPI: fpm-fcgi

Getting `Undefined constant "CodeIgniter\CLI\STDOUT"` error from vendor/codeigniter4/framework/system/CLI/CLI.php Line Number: 162

This was not an error before upgrading to 4.1.3; downgrading back to 4.1.2 removed this error from my project.

paulbalandan:
**Description**
CLI detection is a problematic case as it involves a lot of factors to consider, which may not work at all times at all platforms. This attempts to make detection independent of the interface (i.e. no longer depending on the value of `PHP_SAPI`) but instead rely on other variables.

Fixes #4843 
Fixes #4848 

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -683,19 +683,11 @@ if (! function_exists('is_cli')) {
      *
      * @return bool
      *
-     * @codeCoverageIgnore Cannot be tested fully as PHPUnit always run in CLI
+     * @codeCoverageIgnore Cannot be tested fully as PHPUnit always run in php-cli
      */
     function is_cli(): bool
     {
-        if (PHP_SAPI === 'cli') {
-            return true;
-        }
-
         if (defined('STDIN')) {
-            return true;
-        }
-
-        if (stristr(PHP_SAPI, 'cgi') && getenv('TERM')) {
             return true;
         }
 


paulbalandan:
@michalsn this is my understanding of the current issue. Please correct me if I missed anything.
@pixobit since you initially brought the CLI detection issue in your cron jobs, can you check if this will still work on your case?

system/Common.php
```diff
@@ -683,22 +683,14 @@ function helper($filenames)
...
      *
-     * @codeCoverageIgnore Cannot be tested fully as PHPUnit always run in CLI
+     * @codeCoverageIgnore Cannot be tested fully as PHPUnit always run in php-cli
      */
     function is_cli(): bool
     {
-        if (PHP_SAPI === 'cli') {
-            return true;
-        }
-
```
michalsn, line 693: This is probably the most common setting. So I'm just wondering why we're removing this... were there any issues with it?

michalsn: Oh... okay sorry... I missed your description. I will test it tomorrow but still... I have mixed feelings about removing this check as I don't think it was returning any false results and it's the most common setting we can rely on.

paulbalandan: From a bit of research from php.net, it seems the `STDIN` is defined by the CLI SAPI. So we can at least be sure that if `STDIN` is defined then we're running in php-cli.

MGatner:
I don't have any particular knowledge in this area but I appreciate the work and research.

michalsn (approved):
I tested it on 3 different environments and it works fine.

michalsn:
Thanks @paulbalandan :)

pixobit:
Can confirm that it's working now!

PR #4846, opened 2021-06-17
PSR12: All instances created with `new` keyword must be followed by braces.

paulbalandan:
This is regardless whether the instantiated class has constructor arguments or not.

**Checklist:**
- [x] Securely signed commits


MGatner (approved):
I'm a fan of the parenthesis-free instantiation but I realize this is best for the framework.

PR #4847, opened 2021-06-17
Master language constructs shall be used instead of aliases.

paulbalandan:
**Checklist:**
- [x] Securely signed commits


MGatner (approved):
Wow I never knew `die()` was an alias to exit!

samsonasik (approved):

PR #4851, opened 2021-06-19
CURL (CURLRequest) do not send int/float headers

nowackipawel:
All the headers that were ints/floats in config['headers'] were lost after passing to setHeader.
$config['headers'] = ['X' => 1323, 'Y' => '3566']; 

Of course there is no reason to support complex types, but it would be good to cast ints/floats/booleans to string.
Tested/working.

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -309,6 +309,8 @@ class CURLRequest extends Request
 
         if (array_key_exists('headers', $options) && is_array($options['headers'])) {
             foreach ($options['headers'] as $name => $value) {
+            	if(is_int($value) || is_float($value) || is_bool($value))
+            		$value = (string)$value;
                 $this->setHeader($name, $value);
             }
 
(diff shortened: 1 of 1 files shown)


system/HTTP/CURLRequest.php
```diff
@@ -309,6 +309,8 @@ protected function parseOptions(array $options)
 
         if (array_key_exists('headers', $options) && is_array($options['headers'])) {
             foreach ($options['headers'] as $name => $value) {
+            	if(is_int($value) || is_float($value) || is_bool($value))
+            		$value = (string)$value;
```
samsonasik, line 313: you need to fix the cs, you can run:

```bash
vendor/bin/php-cs-fixer fix system/HTTP/CURLRequest.php
```

paulbalandan (commented):
Please target the `4.2` branch and run the cs on this.

system/HTTP/CURLRequest.php
```diff
@@ -309,6 +309,9 @@ protected function parseOptions(array $options)
 
         if (array_key_exists('headers', $options) && is_array($options['headers'])) {
             foreach ($options['headers'] as $name => $value) {
+                if(is_scalar($value)) {
```
samsonasik, line 312: space after if, you can fix with php-cs-fixer

MGatner:
Headers require values to be a string or array of strings, but this isn't enforced on the Header or Message classes anywhere. I think this change is good, but should we be casting and/or throwing `InvalidArgumentException`s on those classes as well?

PR #4858, opened 2021-06-22
Add dynamic schema support to SQLSRV driver

wradstok:
**Description**
By changing these access modifiers, the functions can be overwritten or extended by end-users.

One example of why this is usefull to us: we utilize different database schema's within our single database, for the same website. However, there is no support for this in the `getFullName` function, as this always adds the default schema. To this end, we would like to override this function with our own implemenation which takes these schema's into account.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


michalsn (commented):
I think that #4860 should solve all the issues with different schemas.

You can change schema simply by: `$this->db->schema = 'myschema'` or via config file.

wradstok:
This is not really helpful, as it would require us to set the schema before
every database access.


Rather, we would do `$this->db->table("schema.table");`

Op wo 23 jun. 2021 18:25 schreef Michal Sniatala ***@***.***>:

paulbalandan:
If your current implementation would be better than the other PR and it would be fitting for framework use, I would rather like to see that implementation here in the PR instead of just changing the methods' visibility. This way it would be useful for the larger community.

wradstok:
I agree on your last point. I can create a pull request (or enhance this pull request) with a new implementation of `getFullName` that checks whether a schema setting is present. However, I have only looked at the SQLSRV driver and not at how this operation is performed in other database driver. I think having seperate behaviours for each driver is not wanted, which is why I opted only to change the methods visibility.

michalsn:
Well... no driver is designed to use schemes in such a dynamic way as you present. If you wanna try to make it happen - then go for it. The only drivers you have to worry about are SQLSRV and Postgre.

wradstok:
I modified the SQLSRV driver with dynamic schema support as we are currently using it. However, we are still in the progress of migrating our CI3 website to CI4, so we may not have caught all scenarios yet.

Regarding the Postgre driver, I'm not sure where to start here. It doesn't seem to do any type of schema management, the table names are just input directly into the query, without any type of escaping or namespacing.

wradstok:
Thanks for hitting the test button 😄 

I see I made the mistake of using str_contains which is a PHP 8.0 function.
I'll check the other test errors later.

qury:
> Rather, we would do `$this->db->table("schema.table");`

@wradstok
I think you would still have issues with listing tables in schemas.
Take a look at the below table where you can't really use a "schema"."table" approach unless you update _listTables() in Connections and split them out again.
`SELECT * FROM [INFORMATION_SCHEMA].[TABLES]`

wradstok:
Changed to be based on 4.2 branch, as this contains the sqlsrv schema updates from #4860

paulbalandan:
Please rebase your PR against `4.2` so only changes from your PR will appear here.

wradstok:
Apologies, since I had already merged the 4.2 changes into my branch it was a weird mashup between CI4 dev and CI4.2. I created a new pull-request based on the 4.2 which contains a single commit that applies all the changes in #4955

Issue #4853, opened 2021-06-21
Bug: SQLSRV driver does not work correctly with schemas and migration


qury:
**Describe the bug**
Migrations do not work with SQLSRV driver if the schema is anything but the default 'dbo'.

CodeIgniter 4 version: 4.1.3

**Affected module(s)**
CodeIgniter\Database\BaseConnection -> when table() is initialized it does not take into consideration the schema
CodeIgniter\Database\SQLSRV ->  $createTableIfStr  does not take into consideration schemas

**Expected behavior, and steps to reproduce if appropriate**

1. create a database for test and create a new schema called 'test'
2. configure CI4 to use 'dbo' as schema
3. create a migration file for any kind of table and run it against 'dbo' schema
4. change the schema to 'test' and run the migration again

Expectation is to have the same table created in both dbo and test schema.
Actual behaviour: table is not created in 'test' schema.

**Context**
 - OS: Linux
 - Web server : n/a -> CLI
 - PHP version 7.4.20

wradstok:
The SQLSRV driver in itself does not account for multiple schemas, entirely, as the method`getFullName` always append the default schema. If none is set (i.e., `$this-db->schema == ""`  , the output looks like: `DatabaseName..TableName`, which is of course not a valid name!

I have opened a pull request that allows you to override the `getFullName` method with your own by changing the access modifier #4858. This is what we are doing now to allow for multiple schema's. However, seeing as probably more people are affected by this issue, perhaps the method itself should be modified?

I would suggest something checking for something like:

```
if (empty($this->db->schema)) {
	if (!str_contains($table, '.')) {
		throw new UnexpectedValueException("Database schema for table ${table} was not specified.");
	}
	return '"' . $this->db->getDatabase() . '"."' . str_replace('"', '', $table) . '"' . $alias;
}
```

qury:
Unfortunately there is a host of other places where the schema will need to be added.
Check my work in progress fix here: https://github.com/qury/CodeIgniter4/tree/fix/slqsrvSchema-4853

I've started working through any errors that comes up from  "tests/system/Database/Live/" scripts:
Current progress: _Tests: 225, Assertions: 451, Errors: 2, Failures: 6._

_The above is with using 'test' for schema instead of the default dbo._

michalsn:
Resolved via #4903

PR #4860, opened 2021-06-22
Implement proper schema support for SQLSRV driver

qury:
**Description**
Update SQLSRV driver and added schema to all places where it was required. See defect #4853 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


samsonasik:
You can fix rector notice with run:

```
vendor/bin/rector
```

michalsn (changes requested):
This looks good, thanks!

I believe this should target branch `4.2`. There also seems to be some code style issues, so please run:
```
vendor/bin/php-cs-fixer fix system/Database/SQLSRV
```

qury:
Hi Guys, 
I've run rector and php-cs-fixer and pushed the updates

> I believe this should target branch `4.2`. There also seems to be some code style issues, so please run:
How do i do this?

samsonasik:
You can change branch target via edit PR button:

Screen Shot 2021-06-23 at 17 59 22

Please ensure rebase before it.

system/Database/SQLSRV/Builder.php
```diff
@@ -199,6 +201,26 @@ protected function _insert(string $table, array $keys, array $unescapedKeys): st
...
+     *
+     * @param string $table  Table name
+     * @param array  $keys   INSERT keys
+     * @param array  $values INSERT values
+     *
+     * @return string
+     */
+    protected function _insertBatch(string $table, array $keys, array $values): string
+    {
+        return 'INSERT ' . $this->compileIgnore('insert') . 'INTO ' . $this->db->escapeIdentifiers($this->db->schema) . '.' . $table . ' (' . implode(', ', $keys) . ') VALUES ' . implode(', ', $values);
```
michalsn, line 219: Any reason why we're using `$this->db->schema` instead of `getFullName()` method? I mean not only here but in general - in the `Builder` class.

michalsn (commented):
Final thoughts.

system/Database/SQLSRV/Builder.php
qury: In the Builder class we can probably switch back to using `getFullName()` in the other classes there are different ways how the schema is used in raw sql. Sometimes as string sometimes as object identifier.
Once i was in the flow with the other classes i just kept my changes aligned to use `$this->db->schema` or `$db->schema`.

system/Database/SQLSRV/Builder.php
```diff
@@ -220,11 +242,54 @@ protected function _update(string $table, array $values): string
         $fullTableName = $this->getFullName($table);
 
         $statement = 'UPDATE ' . (empty($this->QBLimit) ? '' : 'TOP(' . $this->QBLimit . ') ') . $fullTableName . ' SET '
```
paulbalandan, line 244: Can we use `sprintf` for creating the SQL statements? With the current way, it is very error prone and quite hard to understand the concatenations happening.

system/Database/SQLSRV/Builder.php
michalsn: I get it :)
Since we already have this method, I think it would be a good idea to use it - only in `Builder` class of course.

system/Database/SQLSRV/Connection.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+namespace CodeIgniter\Database\SQLSRV;
```
paulbalandan, line 3: namespace declarations should be after the file-level docblock. This place is reserved (if ever) for `declare(strict_types=1);`

system/Database/SQLSRV/Forge.php
```diff
@@ -80,7 +84,7 @@ class Forge extends BaseForge
      *
      * @var string
      */
-    protected $createTableIfStr = "IF NOT EXISTS (SELECT * FROM sysobjects WHERE ID = object_id(N'%s') AND OBJECTPROPERTY(id, N'IsUserTable') = 1)\nCREATE TABLE";
+    protected $createTableIfStr = "IF NOT EXISTS(SELECT t.name, s.name as schema_name, t.type_desc FROM sys.tables t INNER JOIN sys.schemas s on s.schema_id = t.schema_id WHERE s.name=N'%s' AND t.name=REPLACE(N'%s', '\"', '') AND t.type_desc='USER_TABLE')\nCREATE TABLE ";
```
paulbalandan, line 87: This is duplicated in lines 109 to 115 below. I'd prefer keeping the one in the constructor though.

paulbalandan (changes requested):

system/Database/SQLSRV/Builder.php
qury: I will clean this up and add it back, will probably be in the tomorrow

system/Database/SQLSRV/Builder.php
qury: Something like this?
```php
if(empty($this->QBLimit)){
  $statement = sprintf('UPDATE %s SET',$fullTableName);
}else{
  $statement = sprintf('UPDATE TOP(%S)  %s SET',$this->QBLimit, $fullTableName);
}
```

system/Database/SQLSRV/Forge.php
qury: I agree, it would make sense to just declare the variables without any default value and set them in the constructor.

system/Database/SQLSRV/Builder.php
paulbalandan: You can have it like this:
```php
$statement = sprintf(
    'UPDATE %s%s SET',
    empty($this->QBLimit) ? '' : 'TOP('.$this->QBLimit.') ',
    $fullTableName,
);
```

michalsn:
See #4903

PR #4866, opened 2021-06-24
Allow `Utils\\` namespace to be available in external CI4 libraries

paulbalandan:
**Description**
Caution: Developers using the utilities should know that it is "use-at-your-own-risk" and is not covered by backward compatibility promise.

I'm targeting `develop` as most libs use `dev-develop` as their dependency branch so this can be tested easily.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,7 +21,6 @@ Vagrantfile.dist export-ignore
 
 # They don't want our test files
 tests/system/ export-ignore
-utils/ export-ignore
 rector.php export-ignore
 phpunit.xml.dist export-ignore
 phpstan.neon.dist export-ignore
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,8 @@
     },
     "autoload": {
         "psr-4": {
-            "CodeIgniter\\": "system/"
+            "CodeIgniter\\": "system/",
+            "Utils\\": "utils/"
         },
         "exclude-from-classmap": [
             "**/Database/Migrations/**"
@@ -49,8 +50,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "CodeIgniter\\": "tests/system/",
-            "Utils\\": "utils/"
+            "CodeIgniter\\": "tests/system/"
         }
     },
     "scripts": {


MGatner:
I'm fine with the **.gitattributes** change. If we're going to expose the namespace universally we might want to consider a different name, either something more "CodeIgniter" or just less likely to conflict. See if @lonnieezell has any preferences.

sfadschm:
Will this also be the case for upcoming releases (i.e., after the coding standard is released)?
I think most projects (not libs) will likely stay on stable versions.

MGatner:
@sfadschm I think Paul just meant that he was applying the PR to `develop` even though all content changes are headed to `4.2` right now. Maybe I'm mistaken, but it should work with stable releases just fine.

sfadschm:
@MGatner I think so, too.
I was just wondering, whether the `Utils` folder will then be served with releases, too.
If I'm thinking right, it would need to be added here in that case:
https://github.com/codeigniter4/CodeIgniter4/blob/625be30564412f265c77685e6496d8dc70247059/.github/scripts/deploy-framework#L23

MGatner:
Ahh I see what you mean! Whether it will propagate to `codeignter4/framework`. I think that would need intervention, you're right. I wonder if this would make sense as a separate package? Then we could put any of our dev toolkit requirements separate and any other projects or modules could use it.

sfadschm:
👍 Lets see what Paul has planned for this 😄

paulbalandan:
I intend this change for the benefit of module creators (with the disclaimer, ofc) whose dependency is `codeigniter4/codeigniter4: dev-develop`. So I don't think it needs to be packaged to `framework` as I see users will not see any use for that. The utils only serves as an addon to the dev version of CI4 module creators are using.

Since the CS is going to have a separate repo soon, I think it's ok to package the other utils on their own repos as well.

sfadschm:
Alright, just thought it might be helpful for people creating projects (like me) to be able to follow the same standards.
If they will be pulled to their own repos, that will do as well for my use case, I think.

lonnieezell:
For the record, I'm a fan of separating this to it's own repo. I definitely see that as being a simple way for module developers get access to it.

paulbalandan:
I'm in favor too of moving our other toolsets in their own repos. Closing then.

PR #4868, opened 2021-06-24
Normalize tests to use dedicated PHPUnit assertions

paulbalandan:
**Checklist:**
- [x] Securely signed commits


MGatner (approved):
This is great! I've been wanting to do a pass like this. I'll let you merge because I'm sure this could conflict with the other tests pass.

PR #4870, opened 2021-06-24
Test classes should be marked as final and internal

paulbalandan:
**Description**
`CodeIgniter\Test\FeatureTestCase` and `CodeIgniter\Test\TestResponse` were inadvertently marked as `@internal` because they are extending `PHPUnit\Framework\TestCase`. In order for these two not to be marked as final, I added the pseudo-PHPDoc `@no-final`. A permanent fix for this is to make these two as abstract classes as they are intended primarily to be extended by test classes, but that's beyond this PR.

**Checklist:**
- [x] Securely signed commits


MGatner:
I read that wrong. `TestResponse` should *not* be abstract - `CIUnitTestCase` should be.

paulbalandan:
`CIUnitTestCase` is already abstract so it is not affected. `TestResponse` is extended by the deprecated FeatureResponse and probably other test classes too. Either we set TestResponse as abstract and adjust tests directly instantiating it or make it as final and not extensible.

MGatner:
We can't really do either, but it shouldn't be detected as a test case anyways. Developers need to be able to access it directly for testing their responses, and our existing extensions rely on it even though they are deprecated. 

I guess if there were some compelling reason why it needed to be `@final` we could copy all the code to `FeatureResponse` and `ControllerResponse` - but I think some classes are already checking for `TestResponse` so that would probably still be a breaking change.

paulbalandan:
> I guess if there were some compelling reason why it needed to be `@final` we could copy all the code to `FeatureResponse` and `ControllerResponse` - but I think some classes are already checking for `TestResponse` so that would probably still be a breaking change.

Hmm, I guess I'll leave the temp fix of using the pseudo-type `@no-final` for now so it won't be touched by php-cs-fixer, and until such time we can revisit this.

kenjis:
@paulbalandan Why should test classes be marked as final and internal?

`ExampleSessionTest` extends `SessionTestCase` 
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/admin/starter/tests/session/ExampleSessionTest.php
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/admin/starter/tests/_support/SessionTestCase.php

paulbalandan:
@kenjis Test classes are only for use in testing a library's code so it is not appropriate to be used in userland, thus `@internal`. Also, since their use is limited, ,then you don't need to have that flexibility of extending test classes, thus `@final`.

I see in your linked classes and issue that the test cases are final. This is a bug. The ones who created `SessionTestCase` and `DatabaseTestCase` failed to make these as abstract classes. Best practices would be that test cases (test classes containing a pool of common methods for use by child test classes) be made as abstract ones.

kenjis:
@paulbalandan Okay. I will fix the bug.

PR #4871, opened 2021-06-24
Other PHPUnit-related fixers

paulbalandan:
**Description**
Configures other PHPUnit-related fixers.

The last one, `php_unit_strict`, will be done separately as it requires others to be merged and it would require a lot of fixing test failures.

**Checklist:**
- [x] Securely signed commits


samsonasik (approved):

MGatner (approved):
I went through and added all those `void`s when we updated to PHPUnit 9, only to find out later that I should have changed the scope as well. Always meant to go back for it - glad to see this happen!

Issue #4827, opened 2021-06-12
Bug: php spark make: migration--session, "timestamp" column type


tomomo:
Hello.

I noticed while reading #4807.

The documentation(https://codeigniter4.github.io/userguide/libraries/sessions.html?highlight=ci_sessions) specifies `timestamp`.

```php
CREATE TABLE IF NOT EXISTS `ci_sessions` (
        `id` varchar(128) NOT NULL,
        `ip_address` varchar(45) NOT NULL,
        `timestamp` timestamp DEFAULT CURRENT_TIMESTAMP NOT NULL,
        `data` blob NOT NULL,
        KEY `ci_sessions_timestamp` (`timestamp`)
);
```

But, the file created by `php spark make: migration--session` is an `int`.

```php
class CreateCiSessionsTable extends Migration
{
	protected $DBGroup = 'default';

	public function up()
	{
		$this->forge->addField([
			'id'         => ['type' => 'VARCHAR', 'constraint' => 128, 'null' => false],
			'ip_address' => ['type' => 'VARCHAR', 'constraint' => 45, 'null' => false],
			'timestamp'  => ['type' => 'INT', 'unsigned' => true, 'null' => false, 'default' => 0],
			'data'       => ['type' => 'TEXT', 'null' => false, 'default' => ''],
		]);
		$this->forge->addKey('id', true);
		$this->forge->addKey('timestamp');
		$this->forge->createTable('ci_sessions', true);
	}
  :
```

---
**Context**
 - PHP version 7.4.20
 - Codeigniter 4.1.3

michalsn:
Resolved via #4877

Issue #4807, opened 2021-06-09
Bug: Session Library


gmeister2:
**Describe the bug**
Timestamp of session data, saved in database, is always set to 0000-00-00 00:00:00

**CodeIgniter 4 version**
CI 4.1.3

**Affected module(s)**
system/Session/Handlers/DatabaseHandler.php
- write() (line 201 & line 222, 'timestamp'  => 'now()')

**Expected behavior, and steps to reproduce if appropriate**
When inserting values into the session table (default 'ci_sessions'), the timestamp is always set to 0000-00-00 00:00:00.
The session table is set up as described in: https://codeigniter.com/user_guide/libraries/sessions.html#databasehandler-driver

The frameworks internal query builder composed this query (gives a timestamp of 0000-00-00 00:00:00):
INSERT INTO `ci_sessions` (`id`, `ip_address`, `timestamp`, `data`) VALUES ('123456789012345678901234567890', '123.12.0.0', 'now()', '<blob_string>');

If I do a manual insert into the database table, I get a correct result (e.g. timestamp of 2021-06-09 09:56:12):
INSERT INTO `ci_sessions` (`id`, `ip_address`, `timestamp`, `data`) VALUES ('123456789012345678901234567890', '123.12.0.0', now(), '<blob_string>');

The difference being:
'now()' -> now()

My main concern is that the  garbage collector (gc() on line 318) will delete all session data.

**Additional question:**
Am I overlooking something? If not, how would I be able to do a temporary fix for this.

**Context**
 - PHP version 7.4.20
 - MySQL 5.7.34

kenjis:
> Am I overlooking something? If not, how would I be able to do a temporary fix for this.

How about removing the `timestamp` line?
https://github.com/codeigniter4/CodeIgniter4/blob/995c51f383844bc44a607026ea6ab85b06c7e87e/system/Session/Handlers/DatabaseHandler.php#L201

gmeister2:
Thanks for your suggestion.

These might also work (on line 201 & 222):

**change**
`'timestamp'  => 'now()',`

**to**
`'timestamp'  => date('Y-m-d H:i:s'),`
**or** (keeping the 'now()' with postgre)
`'timestamp'  => $this->platform === 'postgre' ? 'now()' : date('Y-m-d H:i:s'),`

Both work for me (my db system: MySQL 5.7.34).

[...]

PR #4877, opened 2021-06-26
Fix for generating session table structure via migrations

michalsn:
**Description**
This PR fixes the invalid table structure that is generated by migrations: `php spark make:migration --session`

Ref: #4827

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Generators/MigrationGenerator.php
+++ b/system/Commands/Generators/MigrationGenerator.php
@@ -107,10 +107,11 @@ class MigrationGenerator extends BaseCommand
             $table   = $this->getOption('table');
             $DBGroup = $this->getOption('dbgroup');
 
-            $data['session'] = true;
-            $data['table']   = is_string($table) ? $table : 'ci_sessions';
-            $data['DBGroup'] = is_string($DBGroup) ? $DBGroup : 'default';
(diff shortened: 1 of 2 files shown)


paulbalandan (approved):

MGatner (approved):
This looks good! I didn't have in my mind how `set()` would circumvent that. @michalsn Do you think this would help with #4807?

michalsn:
@MGatner It definitely will help since now migrations will produce the correct table for every database. But #4876 is dedicated to fix the general problem described in #4807

heriniaina:
Is this fixed?

I changed 
`public $sessionSavePath = 'ci4_session';`

But when running make migration
`> php spark make:migration --session`

I still get ci_sessions as table

michalsn:
The migration command doesn't rely on `$sessionSavePath`.

To achieve what you want, use the command:
```
php spark make:migration --session --table ci4_session
```

heriniaina:
Ok this note should be removed in documentation.

> This command will take the sessionSavePath and sessionMatchIP settings into account when it generates the code.

cf: [here](https://www.codeigniter.com/user_guide/libraries/sessions.html?highlight=make:migration#databasehandler-driver)

PR #4880, opened 2021-06-26
Enable `escape_implicit_backslashes` fixer

paulbalandan:
**Description**
Escape implicit backslashes in strings and heredocs to ease the understanding of which are special chars interpreted by PHP and which not.

**Checklist:**
- [x] Securely signed commits


samsonasik (approved):

system/Commands/Encryption/GenerateKey.php
```diff
@@ -212,6 +212,6 @@ protected function keyPattern(string $oldKey): string
             $escaped = "[$escaped]*";
         }
 
-        return "/^[#\s]*encryption.key[=\s]*{$escaped}$/m";
+        return "/^[#\\s]*encryption.key[=\\s]*{$escaped}$/m";
```
MGatner, line 215: Is that correct? Without reading the context this looks to me like it should be `\s` for "white space character".

system/Database/BaseBuilder.php
```diff
@@ -671,7 +671,7 @@ public function join(string $table, string $cond, string $type = '', bool $escap
                 $operator = $this->getOperator($condition);
 
                 $cond .= $joints[$i];
-                $cond .= preg_match("/(\(*)?([\[\]\w\.'-]+)" . preg_quote($operator, '/') . '(.*)/i', $condition, $match) ? $match[1] . $this->db->protectIdentifiers($match[2]) . $operator . $this->db->protectIdentifiers($match[3]) : $condition;
+                $cond .= preg_match('/(\(*)?([\[\]\w\.\'-]+)' . preg_quote($operator, '/') . '(.*)/i', $condition, $match) ? $match[1] . $this->db->protectIdentifiers($match[2]) . $operator . $this->db->protectIdentifiers($match[3]) : $condition;
```
MGatner, line 674: I'm not sure if this is a separate rule but IMO this was better off with the double quotes and no escape.

system/Typography/Typography.php
```diff
@@ -267,7 +267,7 @@ public function formatCharacters(string $str): string
                 '/"(\W)/'       => '&#8221;$1',
                 '/(\W)"/'       => '$1&#8220;',
                 // apostrophes
-                "/(\w)'(\w)/" => '$1&#8217;$2',
+                '/(\w)\'(\w)/' => '$1&#8217;$2',
```
MGatner, line 270: Same as above, and a few other places.

tests/system/Database/Live/EscapeTest.php
```diff
@@ -72,7 +72,7 @@ public function testEscapeLikeString()
     public function testEscapeLikeStringDirect()
     {
         if ($this->db->DBDriver === 'MySQLi') {
-            $expected = "SHOW COLUMNS FROM brands WHERE column LIKE 'wild\_chars%'";
+            $expected = "SHOW COLUMNS FROM brands WHERE column LIKE 'wild\\_chars%'";
```
MGatner, line 75: Is this right? I'm not sure this should ever have been "escaped".

tests/system/Helpers/TextHelperTest.php
```diff
@@ -27,8 +27,8 @@ public function testStripSlashes()
             "No, my name is O'connor.",
         ];
         $str = [
-            "Is your name O\'reilly?",
-            "No, my name is O\'connor.",
+            "Is your name O\\'reilly?",
+            "No, my name is O\\'connor.",
```
MGatner, line 31: Does this actually change the test case? I think not, but want a second opinion.

MGatner (changes requested):
A few inline comments. Generally, I would prefer this leave alone "single quotes inside double quotes", if possible. I find that to be the cleanest way to deal with quotes within strings (and the reverse) and the conversions to single quotes with slashes makes it look less clear.

system/Commands/Encryption/GenerateKey.php
paulbalandan: I've checked this too. This is correct. Since this is not inside the context of PCRE functions, `\s` does not mean anything even inside double quotes. https://3v4l.org/YVrdo

system/Typography/Typography.php
paulbalandan: Same comment as above.

tests/system/Database/Live/EscapeTest.php
paulbalandan: This is still the same, though uncomfortable at first.

https://3v4l.org/lk32I

tests/system/Helpers/TextHelperTest.php
paulbalandan: See https://3v4l.org/lk32I

system/Database/BaseBuilder.php
MGatner: No yours looks much better 😅

MGatner:
So basically I was mistaken, and the change to single quotes were put in manually by you to prevent some really weird results?

paulbalandan:
> So basically I was mistaken, and the change to single quotes were put in manually by you to prevent some really weird results?

Yes, I changed them because those look weird. But it seems either way the value didn't change. It's just how it looks. I see options here:

1. There's a rule that forces single quotes out there that runs before this fixer. We can use that to remove the weird looks. 😂
2. No manual changes. Have the current fixer fix things as-is.
3. Go with the current manual changes.

MGatner:
Number 3! I'm relieved to hear these changes were manual - I have no problem with the specific ones, I just didn't want it to be a rule than double-quoted single quotes are converted.

MGatner (approved):

PR #4881, opened 2021-06-26
Enable `explicit_indirect_variable`

paulbalandan:
**Description**
Add curly braces to indirect variables to make them clear to understand.

**Checklist:**
- [x] Securely signed commits


MGatner (approved):
I probably wouldn't have chosen this but it does seem to be for the best. Variable methods and variables are often overlooked when scanning code, this makes that very certain it won't happen. 😵‍💫

PR #4882, opened 2021-06-26
Convert to explicit those implicit string variables

paulbalandan:
**Description**
_(excerpted from the fixers' summary and description)_

- `explicit_string_variable`
Converts implicit variables into explicit ones in double-quoted strings or heredoc syntax. The reasoning behind this rule is the following: - When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow - PHP manual marks `"$var"` syntax as implicit and `"${var}"` syntax as explicit: explicit code should always be preferred - Explicit syntax allows word concatenation inside strings, e.g. `"${var}IsAVar"`, implicit doesn't - Explicit syntax is easier to detect for IDE/editors and therefore has colors/highlight with higher contrast, which is easier to read Backtick operator is skipped because it is harder to handle; you can use `backtick_to_shell_exec` fixer to normalize backticks to strings

- `simple_to_complex_string_variable`
Converts explicit variables in double-quoted strings and heredoc syntax from simple to complex format (`${` to `{$`). Doesn't touch implicit variables. Works together nicely with `explicit_string_variable`.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Views/errors/html/error_exception.php
+++ b/app/Views/errors/html/error_exception.php
@@ -92,7 +92,7 @@
 
                                         foreach ($row['args'] as $key => $value) : ?>
 											<tr>
-												<td><code><?= esc(isset($params[$key]) ? '$' . $params[$key]->name : "#$key") ?></code></td>
+												<td><code><?= esc(isset($params[$key]) ? '$' . $params[$key]->name : "#{$key}") ?></code></td>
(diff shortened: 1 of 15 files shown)


MGatner (approved):
This is one I already always do myself, glad to see it.

PR #4883, opened 2021-06-26
Add support for PHP 8.1

paulbalandan:
**Description:**
PHP 8.1 will be released in a few months and we should ensure the framework is compatible by then.

**Detected issues:**
- [x] Declaration of CodeIgniter\Session\Handlers\ArrayHandler::gc($maxlifetime): bool should be compatible with SessionHandlerInterface::gc(int $max_lifetime): int|false (Fixed in #5012)
- [x] `FILTER_SANITIZE_STRING` is deprecated #4938 (Fixed in #5005 and #5263)
- [x] During inheritance of ArrayAccess: Uncaught Return type of CodeIgniter\Cookie\Cookie::offsetExists($offset) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice (Fixed in #5004)
- [x] Fatal error: During inheritance of IteratorAggregate: Uncaught Return type of CodeIgniter\Cookie\CookieStore::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice (Fixed in #5010)
- [x] During inheritance of php_user_filter: Uncaught ErrorException: Return type of CodeIgniter\Test\Filters\CITestStreamFilter::filter($in, $out, &$consumed, $closing) should either be compatible with php_user_filter::filter($in, $out, &$consumed, bool $closing): int, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice (Fixed in #5014)
- [x] `predis/predis` related incompatible signatures (Fixed in https://github.com/predis/predis/pull/708)
- [x] During inheritance of DateTime: Uncaught ErrorException: Return type of CodeIgniter\I18n\Time::__wakeup() should either be compatible with DateTime::__wakeup(): void, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice (Fixed in #5022)
- [x] `mikey179/vfsstream` related incompatible signatures (Fixed in https://github.com/bovigo/vfsStream/pull/261)
- [x] During inheritance of JsonSerializable: Uncaught ErrorException: Return type of CodeIgniter\Entity\Entity::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice (Fixed in #5028)
- [x] During inheritance of SplFileInfo: Uncaught ErrorException: Return type of CodeIgniter\Files\File::getSize() should either be compatible with SplFileInfo::getSize(): int|false (Fixed in #5040)
- [x] Usage of static variables in inherited methods (Fixed in #5262)
- [x] Passing `null` to parameters historically allowing to pass null but now explicitly disallowing so. (Fixed in #5371, #5388)
- [x] Precision loss on conversion of float to int. (Fixed in #5370)
- [x] Compatibility with `PgSql\Result` (Fixed in #5278, #5279)

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -4,12 +4,7 @@ on:
   push:
     branches:
       - develop
-      - '4.*'
     paths:
-      - 'app/**'
-      - 'public/**'
-      - 'system/**'
-      - 'tests/**'
       - composer.json
       - spark
       - phpunit.xml.dist
@@ -18,12 +13,7 @@ on:
   pull_request:
     branches:
       - develop
-      - '4.*'
     paths:
-      - 'app/**'
-      - 'public/**'
-      - 'system/**'
-      - 'tests/**'
       - composer.json
       - spark
       - phpunit.xml.dist
@@ -31,22 +21,24 @@ on:
       - .github/workflows/test-phpunit.yml
 
 jobs:
-
   tests:
+    name: PHP ${{ matrix.php-versions }} - ${{ matrix.db-platforms }}
     runs-on: ubuntu-20.04
     if: "!contains(github.event.head_commit.message, '[ci skip]')"
-    name: PHP ${{ matrix.php-versions }} - ${{ matrix.db-platforms }}
 
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.3', '7.4', '8.0']
+        php-versions: ['7.3', '7.4', '8.0', '8.1']
         db-platforms: ['MySQLi', 'Postgre', 'SQLite3', 'SQLSRV']
         mysql-versions: ['5.7']
         include:
           - php-versions: '7.4'
             db-platforms: MySQLi
             mysql-versions: '8.0'
+          # @todo remove once 8.1 is stable enough
+          - php-versions: '8.1'
+            composer-flag: '--ignore-platform-req=php'
 
     services:
       mysql:
@@ -57,6 +49,7 @@ jobs:
         ports:
           - 3306:3306
         options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
       postgres:
         image: postgres
         env:
@@ -66,6 +59,7 @@ jobs:
         ports:
           - 5432:5432
         options: --health-cmd=pg_isready --health-interval=10s --health-timeout=5s --health-retries=3
+
       mssql:
         image: mcr.microsoft.com/mssql/server:2019-CU10-ubuntu-20.04
         env:
@@ -75,11 +69,13 @@ jobs:
         ports:
           - 1433:1433
         options: --health-cmd="/opt/mssql-tools/bin/sqlcmd -S 127.0.0.1 -U sa -P 1Secure*Password1 -Q 'SELECT @@VERSION'" --health-interval=10s --health-timeout=5s --health-retries=3
+
       redis:
         image: redis
         ports:
           - 6379:6379
         options: --health-cmd "redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
       memcached:
         image: memcached:1.6-alpine
         ports:
@@ -105,6 +101,7 @@ jobs:
 
       - name: Install latest ImageMagick
         run: |
+          sudo apt-get update
           sudo apt-get install --reinstall libgs9-common fonts-noto-mono libgs9:amd64 libijs-0.35:amd64 fonts-urw-base35 ghostscript poppler-data libjbig2dec0:amd64 gsfonts libopenjp2-7:amd64 fonts-droid-fallback ttf-dejavu-core
           sudo apt-get install -y imagemagick
           sudo apt-get install --fix-broken
@@ -122,13 +119,13 @@ jobs:
 
       - name: Install dependencies
         run: |
-          composer update --ansi --no-interaction
-          composer remove --ansi --dev --unused -W rector/rector phpstan/phpstan friendsofphp/php-cs-fixer nexusphp/cs-config codeigniter/coding-standard
+          composer update --ansi --no-interaction ${{ matrix.composer-flag }}
+          composer remove --ansi --dev --unused -W ${{ matrix.composer-flag }} -- rector/rector phpstan/phpstan friendsofphp/php-cs-fixer nexusphp/cs-config codeigniter/coding-standard
         env:
           COMPOSER_AUTH: ${{ secrets.COMPOSER_AUTH }}
 
-      - name: Profile slow tests in PHP 7.4 MySQLi for now
-        if: matrix.php-versions == '7.4' && matrix.db-platforms == 'MySQLi'
+      - name: Profile slow tests in PHP 8.0
+        if: matrix.php-versions == '8.0'
         run: echo "TACHYCARDIA_MONITOR_GA=enabled" >> $GITHUB_ENV
 
       - name: Test with PHPUnit
@@ -137,8 +134,8 @@ jobs:
           DB: ${{ matrix.db-platforms }}
           TERM: xterm-256color
 
-      - if: github.repository_owner == 'codeigniter4' && matrix.php-versions == '7.4'
-        name: Run Coveralls
+      - name: Run Coveralls
+        if: github.repository_owner == 'codeigniter4' && matrix.php-versions == '8.0'
         run: |
           composer global require --ansi php-coveralls/php-coveralls:^2.4
           php-coveralls --coverage_clover=build/logs/clover.xml -v
@@ -151,6 +148,7 @@ jobs:
     if: github.repository_owner == 'codeigniter4'
     needs: [tests]
     runs-on: ubuntu-20.04
+
     steps:
       - name: Coveralls Finished
         uses: coverallsapp/github-action@master


.github/workflows/test-phpunit.yml
```diff
@@ -121,17 +123,18 @@ jobs:
...
+          composer update --ansi --no-interaction ${{ matrix.composer-flag }}
+          composer remove --ansi --dev --unused -W ${{ matrix.composer-flag }} -- rector/rector phpstan/phpstan friendsofphp/php-cs-fixer nexusphp/cs-config
           php -r 'file_put_contents("vendor/laminas/laminas-zendframework-bridge/src/autoload.php", "");'
         env:
           COMPOSER_AUTH: ${{ secrets.COMPOSER_AUTH }}
 
-      - name: Profile slow tests in PHP 7.4 MySQLi for now
-        if: matrix.php-versions == '7.4' && matrix.db-platforms == 'MySQLi'
+      - name: Profile slow tests in PHP 8.0 MySQLi for now
+        if: matrix.php-versions == '8.0' && matrix.db-platforms == 'MySQLi'
```
MGatner, line 133: This is a change I would like to see now. If the rest of the 8.1 workflow takes a while please consider lifting this out.

.github/workflows/test-phpunit.yml
```diff
@@ -40,13 +40,15 @@ jobs:
...
         db-platforms: ['MySQLi', 'Postgre', 'SQLite3', 'SQLSRV']
         mysql-versions: ['5.7']
         include:
-          - php-versions: 7.4
+          - php-versions: '7.4'
             db-platforms: MySQLi
-            mysql-versions: 8.0
+            mysql-versions: '8.0'
+          - php-versions: '8.1'
+            composer-flag: '--ignore-platform-req=php'
```
MGatner, line 51: Is this necessary for some of our dependents? I think the framework should work fine without it, no?

paulbalandan: Some of our dev-deps expressly forbids 8.1 yet. This can be lifted once all our deps allows 8.1

.github/workflows/test-phpunit.yml
paulbalandan: You mean you want to see profiling annotations for each matrix of 7.3 to 8.0 and MySqli to Sqlsrv? GHA allows a limit of 10 warnings per matrix so that would be 160 annotations run.

paulbalandan:
Initial incompat we have is the session handlers extending `SessionHandlerInterface`.
```
During inheritance of SessionHandlerInterface: Uncaught ErrorException: Declaration of
CodeIgniter\Session\Handlers\ArrayHandler::gc($maxlifetime): bool should be compatible
with SessionHandlerInterface::gc(int $max_lifetime): int|false
```

lonnieezell:
Where are we at on this? It's a couple months old now, and I'd love to get some of these cleaned up and either merged or closed.

.github/workflows/test-phpunit.yml
MGatner: No I meant that all single-version tests be run on the latest stable PHP. I think it is moot now that we're close to the next release.

MGatner (approved):

MGatner:
Glad this is going smoothly! Should make for an easy transition this fall so we can focus on the 7.4 minimum.

kenjis:
https://github.com/codeigniter4/CodeIgniter4/pull/4883/checks?check_run_id=3580456786
```
102) CodeIgniter\Database\Builder\SelectTest::testSimpleSelectWithSQLSRV
--- Expected
+++ Actual
@@ @@
-'SELECT * FROM "test"."dbo"."users"'
+'SELECT * FROM "test"."dbo"."`users`"'

/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Database/Builder/SelectTest.php:247
```

kenjis:
```sh-session
$ vendor/bin/phpunit tests/system/Database/Builder/SelectTest.php 
PHPUnit 9.5.9 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.0RC3-dev
Configuration: /Users/kenji/work/codeigniter/CodeIgniter4/phpunit.xml.dist
Warning:       No code coverage driver available

....................                                              20 / 20 (100%)

Time: 00:00.198, Memory: 12.00 MB

OK (20 tests, 22 assertions)
```

kenjis:
When I run all tests,
```sh-session
$ ./phpunit --exclude-group DatabaseLive,CacheLive -d memory_limit=1024m 
```
The Builder tests fail like this:
```
4) CodeIgniter\Database\Builder\AliasTest::testAlias
--- Expected
+++ Actual
@@ @@
-'SELECT * FROM "jobs" "j"'
+'SELECT * FROM `jobs` `j`'
```
But I run only `tests/system/Database/Builder/`, all tests pass:
```sh-session
$ ./phpunit --exclude-group DatabaseLive,CacheLive -d memory_limit=1024m tests/system/Database/Builder/
PHPUnit 9.5.10 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.0RC6-dev
Configuration: /Users/kenji/work/codeigniter/CodeIgniter4/phpunit.xml

...............................................................  63 / 154 ( 40%)
............................................................... 126 / 154 ( 81%)
............................                                    154 / 154 (100%)

Time: 00:00.996, Memory: 16.00 MB

OK (154 tests, 219 assertions)
```

system/HTTP/CLIRequest.php
```diff
@@ -172,17 +170,17 @@ protected function parseCommand()
                 if ($optionValue) {
                     $optionValue = false;
                 } else {
-                    $this->segments[] = filter_var($arg, FILTER_SANITIZE_STRING);
+                    $this->segments[] = esc(strip_tags($arg));
```
kenjis, line 173: This has nothing to do with 8.1 support.
Isn't the existing implementation that does this kind of conversion a bug?

> This is a very dubious filter that has almost no use. It removes NUL bytes, encodes single and double quotes in HTML, and removes anything between < and optional >. It does not function the same as strip_tags() as pointed out in this comment https://www.php.net/manual/en/filter.filters.sanitize.php#118186
https://wiki.php.net/rfc/deprecations_php_8_1

paulbalandan:
> When I run all tests,

Can you run with the `--debug` flag to see what tests are done before the Builder tests.?

system/HTTP/CLIRequest.php
paulbalandan: The point of this change is to move away from the use of FILTER_SANITIZE_STRING, which use will trigger a deprecation in 8.1, while still retaining the behavior this flag is supposed to bring.

kenjis: Sorry, my English was bad.

I wanted to say, the behavior this flag is supposed to bring is a bug?
`$arg` is a param from CLI, isn't it? It is not HTML.

kenjis:
> Can you run with the --debug flag to see what tests are done before the Builder tests.?

See https://gist.github.com/kenjis/6140df495728569e3ce3c2be7499b3de

`AliasTest::testAlias` is the first failed Builder test.

kenjis:
When you rebase this, add this, please.

```diff
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -275,7 +275,7 @@ class Security implements SecurityInterface
             throw SecurityException::forDisallowedAction();
         }
 
-        $json = json_decode($request->getBody());
+        $json = json_decode((string) $request->getBody());
 
         if (isset($_POST[$this->tokenName])) {
             // We kill this since we're done and we don't want to pollute the POST array.
@@ -310,7 +310,7 @@ class Security implements SecurityInterface
         if ($request->hasHeader($this->headerName) && ! empty($request->header($this->headerName)->getValue())) {
             $tokenName = $request->header($this->headerName)->getValue();
         } else {
-            $json = json_decode($request->getBody());
+            $json = json_decode((string) $request->getBody());
 
             if (! empty($request->getBody()) && ! empty($json) && json_last_error() === JSON_ERROR_NONE) {
                 $tokenName = $json->{$this->tokenName} ?? null;

```

system/HTTP/CLIRequest.php
paulbalandan: That was my confusion too when I stumbled here.  Maybe it was assumed you can pass HTML tags in the command line. 🧐 Well I just kinda went along with it.

kenjis: Okay, it is another issue. I will think it again later. 

This PR is to support 8.1. This change preserves the existing (buggy) behavior as much as possible.

kenjis:
@paulbalandan I've found the cause of escape chars difference in QB.

All QB tests passed:
```diff
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1119,7 +1119,7 @@ abstract class BaseConnection implements ConnectionInterface
             return $item;
         }
 
-        static $pregEc = [];
+        $pregEc = [];
 
         if (empty($pregEc)) {
             if (is_array($this->escapeChar)) {
```

kenjis:
I sent a PR: #5262

.github/workflows/test-phpunit.yml
```diff
@@ -4,12 +4,7 @@ on:
   push:
     branches:
       - develop
-      - '4.*'
     paths:
-      - 'app/**'
-      - 'public/**'
-      - 'system/**'
-      - 'tests/**'
```
kenjis, line 12: Why do you remove these paths?

paulbalandan:
I'll be splitting unrelated changes in separate PRs.

.github/workflows/test-phpunit.yml
paulbalandan: Because of the `'**.php'` below. It essentially covers them

paulbalandan:
All green 🎉

I just would have to comment. Tests in PHP 8.1, excluding MySQLi, are pretty fast. Slow tests are now between 20-30 tests.

kenjis (approved):

MGatner:
That's awesome! I'm excited to see what dropping 7.3 does to our overall Actions runtime. If 8.1 is noticeably faster would should be looking at a sizable improvement. 

8.1 should land today! Merge when you're ready.

MGatner (approved):

MGatner:
Just in time! 8.1 is now stable on `setup-php`: https://www.reddit.com/r/PHP/comments/r260o2/github_action_for_php_shivammathursetupphp_v2160/

PR #4890, opened 2021-06-28
Import global symbols

paulbalandan:
**Description**
This is the result of the internal discussion whether to import global symbols (classes, functions, constants) or to fully prepending them with `T_NS_SEPARATOR`.

This PR imports all internal global symbols to speed up scope resolution. Take note that this imports only the native/internal symbols. This means that this does not import user-defined constants and functions exposed to the global namespace.

Linebreaks separating imports of classes, constants and functions are manually added in accordance with PSR-12.

**Checklist:**
- [x] Securely signed commits


paulbalandan:
Not sure what happened here. 🤔

MGatner:
I know this was what we agreed on, but seeing it in action makes me realize: this is a lot of extra lines. I wonder if at some point there is a diminishing return between the scoping performance gains versus the overall line and file size of the framework? We should calculate the difference and maybe look into running some benchmarks against both versions.

paulbalandan:
Closing this for now. The cost of resolving merge conflicts and maintaining PR up to date with latest develop is way too much than its perceived benefits. I can just recreate this when the time comes.

PR #4891, opened 2021-06-29
Request_Is Global Method

ajmeireles:
**Description**
 Introduction of the `request_is` method, an easy way to check if the request method is the same as expected.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -1264,3 +1264,24 @@ if (! function_exists('trait_uses_recursive')) {
         return $traits;
     }
 }
+
+if (! function_exists('request_is')) {
+    /**
+     * Check if a request meets the desired type
+     *
+     * @param string $type
+     *
+     * @return boolean
+     */
+    function request_is(string $type): bool
+    {
+        $request = Services::request(null, true);
+
+        if ($type === 'ajax')
+        {
+            return (bool) $request->isAJAX();
+        }
+
+        return (bool) $request->getMethod() === $type;
+    }
+}
--- a/system/Common.php
+++ b/system/Common.php
@@ -1275,7 +1275,7 @@ if (! function_exists('request_is')) {
      */
     function request_is(string $type): bool
     {
-        $request = Services::request(null, true);
+        $request = service('request');
 
         if ($type === 'ajax')
         {
--- a/system/Common.php
+++ b/system/Common.php
@@ -1282,6 +1282,6 @@ if (! function_exists('request_is')) {
             return (bool) $request->isAJAX();
         }
 
-        return (bool) $request->getMethod() === $type;
+        return strtolower($request->getMethod()) === strtolower($type);
     }
 }
--- a/user_guide_src/source/general/common_functions.rst
+++ b/user_guide_src/source/general/common_functions.rst
@@ -405,7 +405,7 @@ Miscellaneous Functions
     :param   string    $type: The type needed to be compared
     :rtype: bool
 
-    Helper function used to easily compare if the HTTP method matches the need.
+    Helper function to determine if the HTTP method matches the expected value.
 
 ================
 Global Constants
--- a/user_guide_src/source/general/common_functions.rst
+++ b/user_guide_src/source/general/common_functions.rst
@@ -400,6 +400,13 @@ Miscellaneous Functions
 
     Helper function used to convert a string, array, or object of attributes to a string.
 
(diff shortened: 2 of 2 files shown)


paulbalandan:
- While I think this can be useful, we'll be needing some tests to cover this function. You can add in CommonFunctionsTest.
- Please sort your new function. You can put it after the request function.
- Please target the 4.2 branch instead of develop.

iRedds:
I think this is unacceptable.
The function is ambiguous. Both the HTTP method and the technology of the request to the server are checked at the same time.

system/Common.php
```diff
@@ -1264,3 +1264,24 @@ function trait_uses_recursive($trait)
...
+    function request_is(string $type): bool
+    {
+        $request = Services::request(null, true);
+
+        if ($type === 'ajax')
+        {
+            return (bool) $request->isAJAX();
+        }
+
+        return (bool) $request->getMethod() === $type;
```
iRedds, line 1285: 1. Always returns false. Because the result of the getMethod() method will be converted to a boolean value.
2. In the getMethod() method, the $upper parameter is marked as deprecated, which means that the method can return a case-insensitive value.

iRedds (changes requested):

system/Common.php
```diff
@@ -1264,3 +1264,24 @@ function trait_uses_recursive($trait)
...
+    /**
+     * Check if a request meets the desired type
+     *
+     * @param string $type
+     *
+     * @return boolean
+     */
+    function request_is(string $type): bool
+    {
+        $request = Services::request(null, true);
```
MGatner, line 1278: Please use the helper for acquiring services:
```suggestion
        $request = service('request');
```

system/Common.php
```diff
@@ -1264,3 +1264,24 @@ function trait_uses_recursive($trait)
...
+
+        if ($type === 'ajax')
```
MGatner, line 1280: I think that this "muddies" the function a bit. "AJAX" is not a valid HTTP method, and indeed any method may be AJAX so this does not allow checking for POST AJAX versus GET AJAX, etc. I would be in favor of dropping this since the following one-liner will work anywhere in the framework to cover this use:
```php
service('request')->isAJAX();
```

system/Common.php
MGatner, line 1285: As @iRedds noted this should look more like this:
```suggestion
        return strtolower($request->getMethod()) === strtolower($type);
```

user_guide_src/source/general/common_functions.rst
```diff
@@ -400,6 +400,13 @@ Miscellaneous Functions
 
     Helper function used to convert a string, array, or object of attributes to a string.
 
+.. php:function:: request_is ( $type )
```
MGatner, line 403:
```suggestion
.. php:function:: request_is($type)
```

user_guide_src/source/general/common_functions.rst
```diff
@@ -400,6 +400,13 @@ Miscellaneous Functions
...
+
+    :param   string    $type: The type needed to be compared
+    :rtype: bool
+
+    Helper function used to easily compare if the HTTP method matches the need.
```
MGatner, line 408:
```suggestion
    Helper function to determine if the HTTP method matches the expected value.
```

MGatner (changes requested):
A few inline comments. This should be pointed at the `4.2` branch since content is locked on `develop` right now - you can do that with the "Edit" button on GitHub. 

We will also need some tests - there should be some existing ones you could copy for pretty similar behavior.

lonnieezell:
@ajmeireles I think this is a good addition. Are you able to make the changes requested?

lonnieezell:
@MGatner can you re-review this one?

MGatner (changes requested):
@lonnieezell Still conflates AJAX and HTTP method, still no tests.

ajmeireles:
@lonnieezell @MGatner 

I'll push the requested updates soon.

paulbalandan:
@ajmeireles would you be able to add the requests soon? If not, we'll promptly close this.

kenjis:
@ajmeireles This PR is based on too old `develop`.
Please rebase it.

paulbalandan:
No feedback received. Please open a new PR with requested changes applied. Closing.

PR #4892, opened 2021-06-30
Updated Query Build custom string option for where to remove make it clear the values do not get escaped.

lonnieezell:
The current version of the user guide presents some confusion making it appear that the custom string option of the `where` function in the Query Builder would escape the data. That was incorrect. It appears that NO escaping is done in that case. 

The user guide has been updated to make it very clear that user-supplied data MUST be escaped manually when using that option.

### DIFF
--- a/user_guide_src/source/database/queries.rst
+++ b/user_guide_src/source/database/queries.rst
@@ -15,6 +15,7 @@ Regular Queries
 
 To submit a query, use the **query** function::
 
+    $db = db_connect();
     $db->query('YOUR QUERY HERE');
 
 The ``query()`` function returns a database result **object** when "read"
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -245,7 +245,10 @@ This function enables you to set **WHERE** clauses using one of four
 methods:
 
 .. note:: All values passed to this function are escaped automatically,
-    producing safer queries.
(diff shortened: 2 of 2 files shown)


MGatner (approved):
I haven't confirmed your assessment, but I don't use direct query strings so I'll take your word for it. Updates look good!

michalsn (approved):

mmrtonmoybd:
Thanks to all @lonnieezell  @michalsn @MGatner  
I was discovered this sql injection issue, when I was pentesting in our client project. This project is build it with CodeIgniter 4. 

After founding the sql injection issue I was directly contacted to our developer. But he said that all value are passed to the codeigniter are escape automatically because he read it in user guide. 

How stupid developer he is ?

He did not belive it that I found sql injection in where method and he also challenge me to hack this project. Then I deface it within 20 minutes. 

After founding this issue I am directly email to codeigniter security team. But the email was not delivery due to misconfigure DNS in codeigniter server. Then I directly email to @lonnieezell  with full details of the bug and request to fix the bug.

MGatner:
@mmrtonmoybd If you would like to repost your comment without deriding other developers you may. To be clear, this was not an issue with the code, but rather making sure developers knew: **if you use the direct SQL bypass, you are responsible for handling SQL injections**.

Either way, thank you for bringing it to the team's attention.

PR #4902, opened 2021-07-02
Remove `?` from nullable typehints

paulbalandan:
**Description**
Basing from https://github.com/codeigniter4/CodeIgniter4/pull/3033#discussion_r429997442

**Checklist:**
- [x] Securely signed commits


sfadschm:
I would still like the `?` to stay, but thats just personal opinion. Implementing it with just `= null` seems a bit hacky to me (although it is valid).
As we strive for nicely formatted and readable code, I personally would leave the explicit nullable hint.

MGatner:
I assume this does not count as "changing the signature"? Like, extended version with the `?` will still work fine?

paulbalandan:
This does not break LSP. https://3v4l.org/CiQth

sfadschm:
Note that reverting these changes for any reason later will break BC though:
Screenshot_20210703-153032~2

MGatner:
@sfadschm I don't think that applies here since both are actually nullable and amount to the same *actual* types. I just wanted to make sure the parser didn't complain about the *signature* which it seems it will not.

paulbalandan:
@sfadschm I don't think that's the same here. This PR changes the signature for nullables with default null value. It turns explicitly nullable to implicitly nullable.

Your example, on the other hand, requires explicit nullability since the parameter is required. One cannot deduce a required parameter as nullable unless explicitly marked with ?.

sfadschm:
Alright, agreed. I was misleaded here.

MGatner:
@sfadschm does that mean your original opinion (preferring the `?`) is changed?

sfadschm:
@MGatner not really. I will try to elaborate.

While the functionality is the same with or without the `?` operator, I would vote for handling this consistently all over the framework. For me, the typehint is meant to tell the developer what values an argument can take, the default value is for assigning values in case the developer does not provide any. There are two different cases to think about:

**Case 1:**
https://github.com/codeigniter4/CodeIgniter4/blob/c42e79d95b9b842275fed90d0d8b0ada22ed551e/system/BaseModel.php#L1176
Here, the parameter is clearly intended to be optional, as in `BaseModel` itself, it is called various times without passing any parameter.

**Case 2:**
https://github.com/codeigniter4/CodeIgniter4/blob/c42e79d95b9b842275fed90d0d8b0ada22ed551e/system/Database/BaseBuilder.php#L1620
Here, `limit` will never be called without its first parameter. The original intend to use `int $value = null,` in this case hence was to allow `null` to be passed to this argument before `nullable` types where introduced to PHP. The parameter is not optional, so the most precise statement here would be `?int $value,` in my opinion. This is however a breaking change I think.

Espacially in _Case 2_ I think the syntax proposed in this PR misleads the developer to believe that `$value` is intended to be an optional parameter.

While I understand the intention to use the  superset functionality, I think that the gain from these changes (as far as I see, thats only sparing some single characters) does not compensate for the loss of readablity and precision.

Thats my two cents 😄

michalsn:
Honestly, I'm used to scanning variables for the `?` sign just to know if the `null` value is acceptable, even without looking at the default value. That being said, I would vote to leave the `?`.

MGatner:
I don't have a strong opinion either way, but these are all good points toward keeping them. I suppose one more would be for once we are `>= 7.4` and we have many more nullable types throughout the framework.

paulbalandan:
Personally, I would like to have things be explicitly marked to avoid surprises in the future. However, this PR is based on @lonnieezell 's preference. If majority wants to keep the `?` and Lonnie is okay with it, I can gladly turn this PR around to do the opposite. 😁

paulbalandan:
Closed in favor of #4928

PR #4905, opened 2021-07-04
Deptrac

MGatner:
**Description**
[Deptrac](https://github.com/qossmic/deptrac) is a new (to me) tool for defining and inspecting a project's architecture. It works by identifying "layers" (think components, like "Cache" or "Database") and then defining which layers may depend on each other. Any dependency that is not explicitly permitted will cause a "violation". Defining these structures accomplishes a few different things...
1. The **depfile** becomes a mini dependency chart, giving an easy visualization of framework components. Running `deptrac` can also create an actual dependency chart in numerous formats (see https://github.com/qossmic/deptrac/blob/main/docs/formatters.md)
2. Defining layers helps identify classes that may be "misplaced". For example during this process I noticed `URI` is so heavily used apart from the rest of `HTTP` that I made it into its own layer, which might suggest it should be given its own namespace in a future version.
3. Creating the rulesets makes it clear when there are components interconnected that probably shouldn't be. Making these changes now would likely cause breaking changes, but Deptrac allows for "skipping" some violations to keep them in mind for the future.
4. Adding Deptrac to CI/CD ensures that Pull Requests are not creating new dependencies unmindfully. Should new relationships be desired that benefit the framework they can always be added to **depfile**.
5. A **depfile** can import another **depfile**, so if we offer a framework dependency structure then modules and projects which use the framework can "opt in" to our same architecture.

What this PR includes:
* **depfile**: This is my best attempt at a "first pass" for defining layers. The rulesets are based off our current actual dependencies, with the exception of some "skips" that I noted.
* **test-deptrac.yml**: Based off my module workflow, this will install and run Deptrac for new PRs and pushes as described above.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/.github/workflows/test-deptrac.yml
+++ b/.github/workflows/test-deptrac.yml
@@ -1,0 +1,82 @@
+# When a PR is opened or a push is made, perform an
+# architectural inspection on the code using Deptrac.
+name: Deptrac
+
+on:
+  pull_request:
+    branches:
+      - 'develop'
+      - '4.*'
+    paths:
+      - 'app/**'
+      - 'system/**'
+      - 'composer.json'
+      - 'depfile.yaml'
+      - '.github/workflows/test-deptrac.yml'
+  push:
+    branches:
+      - 'develop'
+      - '4.*'
+    paths:
+      - 'app/**'
+      - 'system/**'
+      - 'composer.json'
+      - 'depfile.yaml'
+      - '.github/workflows/test-deptrac.yml'
+
+jobs:
+  build:
+    name: PHP ${{ matrix.php-versions }} Architectural Inspection
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.4', '8.0']
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+          tools: composer, pecl, phive, phpunit
+          extensions: intl, json, mbstring, gd, mysqlnd, xdebug, xml, sqlite3
+
+      - name: Use latest Composer
+        run: composer self-update
+
+      - name: Validate composer.json
+        run: composer validate --strict
+
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+      - name: Create composer cache directory
+        run: mkdir -p ${{ steps.composer-cache.outputs.dir }}
+
+      - name: Cache composer dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: ${{ runner.os }}-composer-
+
+      - name: Create Deptrac cache directory
+        run: mkdir -p build/
+
+      - name: Cache Deptrac results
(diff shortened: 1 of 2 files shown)


sfadschm:
Sounds interesting.
Could you give an example, e.g., what needs to be done with `CodeIgniter\HTTP\ResponseTrait` to fix the violation? I cannot really image that yet.

MGatner:
Sure! That one was easy because it was a bizarre dependency (why would HTTP need Pager?) and then there was already a note on `ResponseTrait::setLink()`: 

> @todo Recommend moving to Pager

Of course we can't do that now because it would be a breaking change, but now it is "documented" for version 5 when we have the opportunity to change things around and make components less interdependent (or at least more intentionally dependent).

sfadschm:
Thanks, understood.
Seems like a good addition!

michalsn (approved):
This may be helpful. We should give it a try.

samsonasik (approved):

.github/workflows/test-deptrac.yml
```diff
@@ -0,0 +1,82 @@
...
+      - '.github/workflows/test-deptrac.yml'
+
+jobs:
+  build:
+    name: PHP ${{ matrix.php-versions }} Architectural Inspection
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.4', '8.0']
```
paulbalandan, line 34: Is there merit for testing it on different PHP versions? I mean we can test only on highest version allowed.

.github/workflows/test-deptrac.yml
```diff
@@ -0,0 +1,82 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+          tools: composer, pecl, phive, phpunit
```
paulbalandan, line 43:
```suggestion
          tools: composer, phive
```

.github/workflows/test-deptrac.yml
```diff
@@ -0,0 +1,82 @@
...
+          extensions: intl, json, mbstring, gd, mysqlnd, xdebug, xml, sqlite3
+
+      - name: Use latest Composer
```
paulbalandan, line 46: No need for this as setup-php installs the latest available.

.github/workflows/test-deptrac.yml
MGatner: Hmm that's a good question. I'm not sure? Probably we don't need both, I can't imagine they would amount to meaningful differences. Most of this workflow file was cobbled together from existing templates, so I won't say every part of it was intentional :D

.github/workflows/test-deptrac.yml
MGatner: Ironically, I took this from our PHPStan workflow! Never used it myself but I figured somebody knew something I didn't :D

https://github.com/codeigniter4/CodeIgniter4/blob/1928006ccc9fa86be11bb8ee5708f4d7bec95fda/.github/workflows/test-phpstan.yml#L43-L44

paulbalandan: I think this was because composer v2 wasn't released then and anyone can force update with `composer self-update --2`

.github/workflows/test-deptrac.yml
paulbalandan: I would go for one version only since this only tests architectural relations which I believe won't change in between PHP versions.

paulbalandan (approved):

PR #4908, opened 2021-07-04
Update docs for Throttler

### LINKED ISSUE #4904: Bug: Cache keys erros using throttle
**Describe the bug**
When using throttle, the cache keys error occurs. Just instantiate and put in middeware as it is in the documentation. 
If used the host localhost causes the error.

Error:
image

Print key: 
image

File error:
image

**CodeIgniter 4 version**
4.1.3

**Context**
 - OS: Windows 10
 - Web server PHP - Server
 - PHP version 7.4

michalsn:
Fixes #4904


paulbalandan (approved):

MGatner (approved):
Thanks for that quick turnaround!

PR #4886, opened 2021-06-28
[Debug] [4.2] Structure debug toolbar timeline with collapsible elements.

sfadschm:
This is an extension of #4865.

This PR intrduces structuring of the debugbar timeline elements.
They are now sorted in a parent-children structure like this:

    [
        'name'     => 'parentTimer',
        'start'    => '0',
        'duration' => '10',
        'children' => [
            [
                'name'     => 'childTimer',
                'start'    => 2,
                'duration' => 4
            ]
        ]
    ]

These elements are then echoed recursively and embedded in JS collapsible table rows.

For now, the `Controller` parent element is unfolded by default, as this will likely alway be the biggest and last part of the timeline.

One thing I am not yet fond of is the flexible indentation of the `Timer` name in the current state. That is, each deeper level is indented by an additional 10px, which I think might pose some issue on very narrow screens.

**Demonstration**

https://user-images.githubusercontent.com/40514119/123594641-41250480-d7f0-11eb-8382-9eae38802671.mp4

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #4910, opened 2021-07-05
[4.2] Add toolbar timer styles to sass files.

sfadschm:
In #4886, the toolbar styles were updated, but I failed to notice that they are defined in `admin` in `scss` files.
This PR pulls the changes over to the preprocessor files.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/admin/css/debug-toolbar/_theme-dark.scss
+++ b/admin/css/debug-toolbar/_theme-dark.scss
@@ -12,11 +12,14 @@
 // ========================================================================== */
 
 #debug-icon {
-	background-color: $t-dark;
-	@include box-shadow(0, 0, 4px, $m-gray);
-	a:active, a:link, a:visited {
-		color: $g-orange;
-	}
+    background-color: $t-dark;
+    @include box-shadow(0, 0, 4px, $m-gray);
+
+    a:active,
+    a:link,
+    a:visited {
+        color: $g-orange;
+    }
 }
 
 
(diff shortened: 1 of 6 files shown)


MGatner (approved):
Thanks!

MGatner:
@LittleJ anything to add?

system/Debug/Toolbar/Views/toolbar.css
```diff
@@ -186,22 +186,6 @@
...
-        height: 15px;
-        width: 15px;
-        margin-right: 3px;
-        vertical-align: middle; }
-      #debug-bar .timeline tr.timeline-parent.timeline-parent-open td:first-child nav {
-        background-position: 0 75%; }
-    #debug-bar .timeline tr.timeline-parent.timeline-parent-open {
-      background-color: #DFDFDF; }
-    #debug-bar .timeline tr.child-row:hover {
-      background: transparent; }
```
paulbalandan, line 204: Is the removal of these lines the result of compiling the modified scss files?

sfadschm: yes, those were the lines that I manually inserted to the css file in the referenced PR.
they are now in another (hierarchically more suitable) place due to the compiling

paulbalandan: Thanks!

admin/css/debug-toolbar/toolbar.scss
```diff
@@ -326,6 +326,23 @@
 
 			&:first-child {
 				border-left: 0;
+				max-width: none;
```
paulbalandan, line 329: little request. Since we're moving to spaces over tabs, can you convert this to spaces also? You can use a separate commit in case others won't like it. 😅

admin/css/debug-toolbar/_theme-dark.scss
```diff
@@ -12,141 +12,156 @@
...
- 		background-color: $g-gray;
- 		@include box-shadow(0, 0, 4px, $g-gray);
-		img {
-			filter: brightness(0) invert(1);
-		}
- 	}
-
-	// Fixed top
-	&.fixed-top {
-		& .toolbar {
```
sfadschm, line 85: Not 100% sure about removing the `&` here, but the css turn out same with and without as it is originally written as `& .toolbar` which should  be the same.

sfadschm:
Happily switching to spaces.
Also "fixed" some other, not unified styling.

system/Debug/Toolbar/Views/toolbar.css
```diff
@@ -1,8 +1,8 @@
 /*! CodeIgniter 4 - Debug bar
```
sfadschm, line 1: Should the header be copied from the php source files?

sfadschm: And in addition, should all scss files have one?

paulbalandan: I think yes, for consistency.

paulbalandan (approved):

paulbalandan:
Just a tip. `[ci skip]` doesn't work on PR titles. It should be on the commit message to actually work.

sfadschm:
> Just a tip. `[ci skip]` doesn't work on PR titles. It should be on the commit message to actually work.

Ah this explains a lot 😆 
However, I think it worked when creating the PR. But I will try to remember next time ^.^

LittleJ:
Sorry for the late answer. I forgot I had turned off all the notifications... I will have more time to dedicate to CI the next days and weeks. I will check all the great new updates :-) and those pull requests. Thanks @MGatner for asking me ❤️😄

paulbalandan:
I'll proceed first with merging this. Thanks, @sfadschm .

PR #4911, opened 2021-07-05
[4.2] Scalar values convertion for setHeader / Cookie settings improved

nowackipawel:
PR contains:
1. changes related to PR #4851  (I closed it)
2. ability to define different files for COOKIEJAR and COOKIEFILE (there was no such an option; and those are different type of files). Backward compatibility is provided
3. ability to define CURLOPT_COOKIESESSION

File has been fixed with php-cs-fixer.

I'm calling all the folks who wrote under #4851 @samsonasik , @MGatner , @paulbalandan

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -311,7 +311,9 @@ class CURLRequest extends Request
             foreach ($options['headers'] as $name => $value) {
                 if (is_scalar($value)) {
                     $this->setHeader($name, $value);
-                }
+                } else {
+					throw HTTPException::forNonScalarHeaderValue();
+				}
             }
 
             unset($options['headers']);
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -309,7 +309,9 @@ class CURLRequest extends Request
 
         if (array_key_exists('headers', $options) && is_array($options['headers'])) {
             foreach ($options['headers'] as $name => $value) {
-                $this->setHeader($name, $value);
+                if (is_scalar($value)) {
+                    $this->setHeader($name, $value);
+                }
             }
 
             unset($options['headers']);
@@ -705,9 +707,16 @@ class CURLRequest extends Request
         }
 
         // Cookie
-        if (isset($config['cookie'])) {
-            $curlOptions[CURLOPT_COOKIEJAR]  = $config['cookie'];
-            $curlOptions[CURLOPT_COOKIEFILE] = $config['cookie'];
+        if ($cookie = ($config['cookie_jar'] ?? $config['cookie'] ?? false)) {
+            $curlOptions[CURLOPT_COOKIEJAR] = $cookie;
+        }
+
+        if ($cookie = ($config['cookie_file'] ?? $config['cookie'] ?? false)) {
+            $curlOptions[CURLOPT_COOKIEFILE] = $cookie;
+        }
+
+        if ($cookie = ($config['cookie_session'] ?? false)) {
+            $curlOptions[CURLOPT_COOKIESESSION] = $cookie;
         }
 
         // User Agent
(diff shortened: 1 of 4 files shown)


system/HTTP/CURLRequest.php
```diff
@@ -309,7 +309,9 @@ protected function parseOptions(array $options)
 
         if (array_key_exists('headers', $options) && is_array($options['headers'])) {
             foreach ($options['headers'] as $name => $value) {
-                $this->setHeader($name, $value);
+                if (is_scalar($value)) {
```
MGatner, line 312: In my opinion if an unacceptable value is passed it should not be quietly skipped but rather should throw an exception. Otherwise we are creating a scenario where developers have to dig pretty deep to find out why their Entity header did not get set.

system/HTTP/CURLRequest.php
```diff
@@ -705,9 +707,16 @@ protected function setCURLOptions(array $curlOptions = [], array $config = [])
...
-        if (isset($config['cookie'])) {
-            $curlOptions[CURLOPT_COOKIEJAR]  = $config['cookie'];
-            $curlOptions[CURLOPT_COOKIEFILE] = $config['cookie'];
+        if ($cookie = ($config['cookie_jar'] ?? $config['cookie'] ?? false)) {
+            $curlOptions[CURLOPT_COOKIEJAR] = $cookie;
+        }
+
+        if ($cookie = ($config['cookie_file'] ?? $config['cookie'] ?? false)) {
+            $curlOptions[CURLOPT_COOKIEFILE] = $cookie;
+        }
```
MGatner, line 716: I think this would be cleaner if we didn't reuse variables:
```suggestion
        if ($cookieJar = ($config['cookie_jar'] ?? $config['cookie'] ?? false)) {
            $curlOptions[CURLOPT_COOKIEJAR] = $cookieJar;
        }

        if ($cookieFile = ($config['cookie_file'] ?? $config['cookie'] ?? false)) {
            $curlOptions[CURLOPT_COOKIEFILE] = $cookieFile;
        }
```

system/HTTP/CURLRequest.php
```diff
@@ -705,9 +707,16 @@ protected function setCURLOptions(array $curlOptions = [], array $config = [])
...
+
+        if ($cookie = ($config['cookie_session'] ?? false)) {
+            $curlOptions[CURLOPT_COOKIESESSION] = $cookie;
```
MGatner, line 719: I understand the unusual structure above but this one can use the typical format:
```suggestion
        if (isset($config['cookie_session'])) {
            $curlOptions[CURLOPT_COOKIESESSION] = $config['cookie_session'];
```

MGatner (changes requested):
Generally this looks like a good start. I made a few inline comments. This will still need to target the `4.2` branch, which might cause conflicts if these files have undergone other style changes. You can try repointing it on GitHub by clicking "Edit" from the Pull Request. 

We need tests and User Guide updates. I know you have been hesitant in the past but these amount to addition keys for things that already have existing content so it should be a pretty easy copy-paste. We cannot proceed without those.

system/HTTP/CURLRequest.php
nowackipawel: As for now developers have to dig^2 cuz there is no explanation why their scalar values that are not stings are not passing to CURL request but they are ignoring instead. I would say scalar values should be converted here.

system/HTTP/CURLRequest.php
nowackipawel: In this case there should be only boolean value passed. I will update it according to your coding style advice.

system/HTTP/CURLRequest.php
MGatner: No no, sorry, I was unclear. This part is great - I want to add something like this:
```php
                if (! is_scalar($value)) {
    throw new DomainException('Only scalar values are allowed');
```

MGatner (changes requested):
Still needs to be pointed at the `4.2` branch (*not* `develop`). Still needs tests.

PR #4914, opened 2021-07-06
[4.2][Debug] Do not overwrite startTime.

sfadschm:
This PR gives a little change to the execution time collection, by only overwriting the `startTime` property of the `CodeIgniter4` object, if it is not yet set.

This property gets first set here in the constructor:
https://github.com/codeigniter4/CodeIgniter4/blob/9ceb5e3c023afd083ebeb9dfb8b47e5bcabf6770/system/CodeIgniter.php#L160

After that, it gets overwritten from `bootstrap.php` during `$app->run()` via:
https://github.com/codeigniter4/CodeIgniter4/blob/9ceb5e3c023afd083ebeb9dfb8b47e5bcabf6770/system/CodeIgniter.php#L514

I am not sure what was the goal here. In the current state, we basically restart the "total execution" timer and possibly miss some ms (i.e., the time that `$app->initialize();` takes).  This is not really helpful I think.

As far as I can tell, the part now included in the `if` could be removed completely, as `startTime` will likely never be `null` at this point.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -511,7 +511,9 @@ class CodeIgniter
      */
     protected function startBenchmark()
     {
-        $this->startTime = microtime(true);
+        if($this->startTime === null) {
+            $this->startTime = microtime(true);
+        }
 
         $this->benchmark = Services::timer();
         $this->benchmark->start('total_execution', $this->startTime);


sfadschm:
Judging from @lonnieezell (very old) commits, this has been some debate in the past:

https://github.com/codeigniter4/CodeIgniter4/commits/9521545bdce6a33f56e0f48767f4bf9855110e7d

However, in the current state, `startTime` will always be overwritten before it is used unless someone edits `index.php` and puts some content between the `CodeIgniter4` app construction and `$app->run()`.

MGatner (approved):
I think this is fine. We can see if Lonnie weighs in but we have come a long ways and I don't think comparing internal benchmarking between CI3 and CI4 does anything useful. 

FWIW I have plans for a benchmarking workflow. I will probably dig into this a bit more when I get to that.

paulbalandan (approved):

sfadschm:
Should I change to just remove the line?
I do not think the conditional will ever be true as the method cannot be run without the constructor being run beforehand.
Only way would be if someone intentionally resets `startTime` to `null` after the constructor.

sfadschm:
> FWIW I have plans for a benchmarking workflow. I will probably dig into this a bit more when I get to that.

Sounds great 😄

PR #4919, opened 2021-07-07
Define "i" rules

paulbalandan:
**Description**
- implode should be using the documented order of parameters
- include/require should not be bracketed
- is_null($var) converted to null === $var (and fixed logic errors suppressed by is_null)

**Checklist:**
- [x] Securely signed commits


app/Views/errors/cli/error_exception.php
```diff
@@ -49,7 +49,7 @@
                 case is_array($value):
                     return count($value) ? '[...]' : '[]';
 
-                case is_null($value):
+                case $value === null:
```
sfadschm, line 52: Not directly related to this PR, but what is this `switch` statement? 😆 
While I like the appearance, is this really better than consecutive `if`?

paulbalandan: This serializes the args by checking it against their types and returns a string version.

While switch is not strictly checking the types in its cases arms, I think it is safe to assume so since the cases all evaluate to bool. But you're right, if guards work well here too. I don't have preference either way. 😂

MGatner (approved):
I looked through a lot of them but not all.

PR #4923, opened 2021-07-08
Define initial "s" rules

paulbalandan:
**Description**
- Inside class or interface element `self` should be preferred to the class name itself.
- Inside a final class or anonymous class `self` should be preferred to `static`.
- ~Instructions must be terminated with a semicolon.~
- Simplify if control structures that return the boolean result of their condition.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Pager/Pager.php
+++ b/system/Pager/Pager.php
@@ -481,7 +481,7 @@ class Pager implements PagerInterface
      *
      * @return Pager
      */
-    public function only(array $queries): Pager
+    public function only(array $queries): self
     {
         $this->only = $queries;
 
--- a/system/Validation/ValidationInterface.php
+++ b/system/Validation/ValidationInterface.php
@@ -53,7 +53,7 @@ interface ValidationInterface
      *
      * @return ValidationInterface
      */
-    public function withRequest(RequestInterface $request): ValidationInterface;
+    public function withRequest(RequestInterface $request): self;
 
     //--------------------------------------------------------------------
     // Rules
@@ -67,7 +67,7 @@ interface ValidationInterface
      *
      * @return ValidationInterface
      */
-    public function setRules(array $rules, array $messages = []): ValidationInterface;
+    public function setRules(array $rules, array $messages = []): self;
 
     //--------------------------------------------------------------------
 
@@ -119,7 +119,7 @@ interface ValidationInterface
      *
      * @return ValidationInterface
      */
-    public function setError(string $alias, string $error): ValidationInterface;
+    public function setError(string $alias, string $error): self;
 
     //--------------------------------------------------------------------
     // Misc
@@ -131,7 +131,7 @@ interface ValidationInterface
      *
      * @return ValidationInterface
      */
-    public function reset(): ValidationInterface;
+    public function reset(): self;
 
     //--------------------------------------------------------------------
 }
--- a/system/Validation/ValidationInterface.php
+++ b/system/Validation/ValidationInterface.php
@@ -53,7 +53,7 @@ interface ValidationInterface
      *
      * @return ValidationInterface
      */
-    public function withRequest(RequestInterface $request): self;
+    public function withRequest(RequestInterface $request): ValidationInterface;
 
     //--------------------------------------------------------------------
     // Rules
@@ -67,7 +67,7 @@ interface ValidationInterface
      *
      * @return ValidationInterface
      */
-    public function setRules(array $rules, array $messages = []): self;
+    public function setRules(array $rules, array $messages = []): ValidationInterface;
 
     //--------------------------------------------------------------------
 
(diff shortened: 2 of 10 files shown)


mostafakhudair:
About 3rd point, Instructions shouldn't terminated with semicolon, as it will be include automatically when closing tag on the same line of the code

 PHP Manual: https://www.php.net/manual/en/language.basic-syntax.instruction-separation.php

app/Views/errors/html/error_404.php
```diff
@@ -74,10 +74,10 @@
 
 		<p>
 			<?php if (! empty($message) && $message !== '(null)') : ?>
-				<?= nl2br(esc($message)) ?>
+				<?= nl2br(esc($message)); ?>
 			<?php else : ?>
```
MGatner, line 78: I think this should be `else:` - do we have a run yet coming for that.

app/Views/errors/html/error_exception.php
```diff
@@ -5,32 +5,32 @@
 	<meta charset="UTF-8">
 	<meta name="robots" content="noindex">
 
-	<title><?= esc($title) ?></title>
+	<title><?= esc($title); ?></title>
```
MGatner, line 8: I usually don't use a semicolon on echoes, it doesn't make sense to me to have them and it makes scanning PHP in HTML easier to distinguish. Is this configurable or is it a bundle deal? Is this a PSR-12 requirement or an addition?

system/Pager/Pager.php
```diff
@@ -481,7 +481,7 @@ public function getDetails(string $group = 'default'): array
      *
      * @return Pager
      */
-    public function only(array $queries):Pager
+    public function only(array $queries):self
```
MGatner, line 484: Probably covered in some other rule but:
```suggestion
    public function only(array $queries): self
```

system/Validation/ValidationInterface.php
```diff
@@ -53,7 +53,7 @@ public function check($value, string $rule, array $errors = []): bool;
      *
      * @return ValidationInterface
      */
-    public function withRequest(RequestInterface $request): ValidationInterface;
+    public function withRequest(RequestInterface $request): self;
```
MGatner, line 56: I'm not sure about this change for interfaces. I assume PHP handles this correctly (i.e. the implementing class must return a `ValidationInterface` instead of `self`) but it makes it confusing for developers. For example I like to copy-paste an interface when starting a new implementing class and fill in the methods but this will be incorrect if I failed to update the signatures.

MGatner (changes requested):

app/Views/errors/html/error_404.php
paulbalandan: I cannot recall if there is.

app/Views/errors/html/error_exception.php
paulbalandan: This isn't in PSR12. And is not configurable. I'll just disable this as consensus wants otherwise.

system/Pager/Pager.php
paulbalandan: This is covered by return type declaration fixer.

system/Validation/ValidationInterface.php
paulbalandan: I'm not sure what you mean, but something like this?
```php
interface ValidationInterface
{
    public function validate(array $data): self;
}

class Validator implements ValidationInterface
{
    private $data;
    public function validate(array $data): ValidationInterface
    {
        $this->data = $data;
        return $this;
    }
}

class Validation implements ValidationInterface
{
    private $data;
    public function validate(array $data): self
    {
        $this->data = $data;
        return $this;
    }
}
```

The second class will fail for 7.3 and below because full return covariance is supported in 7.4+
See https://3v4l.org/bpubF

MGatner: Right. My question is: since the implementing class (Validator, in your example) *must* return `ValidationInterface` instead of `self` would it be cleaner & clearer to keep the class name on the interface itself?

paulbalandan: It could be done but we'll have to disable the rule entirely as there's no option to skip a rule for a file.

OR

I can defer enabling this rule to when we're requiring 7.4+ so as to support full return covariance.

MGatner: This is always a mess with 7.3 and under. I would be in favor of applying this rule's fixes then disabling the rule and changing these special cases back manually. I'm open to other opinions though.

paulbalandan: For now, I did what you want. Enabling the effects on classes but not on interfaces, then disabling the fixer. We can revisit this later when we bump PHP requirements.

MGatner: Thanks! I imagine we'll have a ton of changes around typehints when we drop 7.3 so I think that's a great plan for when to revisit.

MGatner (approved):

PR #4924, opened 2021-07-08
Define `single_**` rules

paulbalandan:
**Description**
- A PHP file without end tag must always end with a single empty line feed.
- There MUST NOT be more than one property or constant declared per statement.
- There MUST be one use keyword per declaration.
- Each namespace use MUST go on its own line and there MUST be one blank line after the use statements block.
- Single-line comments and multi-line comments with only one line of actual content should use the `//` syntax.
- Convert double quotes to single quotes for simple strings.
- Ensures a single space after language constructs.
- Each trait `use` must be done as single statement.

**Checklist:**
- [x] Securely signed commits


system/Database/SQLSRV/PreparedQuery.php
```diff
@@ -50,12 +50,12 @@ class PreparedQuery extends BasePreparedQuery
...
     public function _prepare(string $sql, array $options = [])
     {
-        /* Prepare parameters for the query */
+        // Prepare parameters for the query
         $queryString = $this->getQueryString();
 
         $parameters = $this->parameterize($queryString);
 
-        /* Prepare  the query */
+        // Prepare  the query
```
MGatner, line 58:
```suggestion
        // Prepare the query
```

MGatner (approved):
Looks good! One spare space.

MGatner (approved):

PR #4926, opened 2021-07-08
Define "r" rules

paulbalandan:
**Description**
- Replaces `rand`, `srand`, `getrandmax` functions calls with their `mt_*` analogs.
- Callables must be called without using `call_user_func*` when possible.
- Local, dynamic and directly referenced variables should not be assigned and directly returned by a function or method.
- There should be no space before colon, and one space after it in return type declarations.

**Checklist:**
- [x] Securely signed commits


MGatner:
> Many random number generators of older libcs have dubious or unknown characteristics and are slow. The mt_rand() function is a drop-in replacement for the older rand(). It uses a random number generator with known characteristics using the » Mersenne Twister, which will produce random numbers four times faster than what the average libc rand() provides.

I had no idea! Thanks for introducing me.

MGatner (approved):

PR #4933, opened 2021-07-13
Define remaining phpdoc-related rules

paulbalandan:
**Description**
- PHPDoc should contain @param for all params.
- The type of @return annotations of methods returning a reference to itself must the configured one.
- Single line @var PHPDoc should have proper spacing.
- Fixes casing of PHPDoc tags.
- Forces PHPDoc tags to be either regular annotations or inline.
- Docblocks should only be used on structural elements. (Enabled only on a few files then disabled. Current behavior is not desired)
- Remaining rules are turned off as these breaks signatures. These can be used on migration rules instead.

**Checklist:**
- [x] Securely signed commits


system/Helpers/inflector_helper.php
```diff
@@ -11,9 +11,7 @@
 
 // --------------------------------------------------------------------
 
-/**
- * CodeIgniter Inflector Helpers
- */
+// CodeIgniter Inflector Helpers
```
sfadschm, line 14: Sometimes there seems to be an empty line after the comment, sometimes not.
I think there should not be one?

system/Email/Email.php
```diff
@@ -1317,6 +1317,8 @@ protected function buildMessage()
     }
 
     /**
+     * @param string $type
```
sfadschm, line 1320: Why is thos typehinted? The argument does not seem to be.

tests/_support/Log/Handlers/TestHandler.php
```diff
@@ -19,6 +19,8 @@ class TestHandler extends \CodeIgniter\Log\Handlers\FileHandler
 
     /**
      * Where would the log be written?
+     *
+     * @param array $config
      */
     //--------------------------------------------------------------------
```
sfadschm, line 25: Not directly related to this PR, but I recall there was a discussion about these dividers.
Not sure how it concluded, but looks weird here.

MGatner:
> PHPDoc should contain `@param` for all params.

I'd like to see some more discussion on this one. According to PHPDocs there is no need to annotate existing typehints, and it adds nothing to IDEs etc. So something like this:

```php
     * @param string $locale
     * @param string $file
     * @param string $parsedLine
     *
    protected function getTranslationOutput(string $locale, string $file, string $parsedLine)
```
... adds three lines of docblocks for nothing, and sometimes will even add block comments where they would not be needed at all. 

This might be a discussion for after 7.3 when we will have much better typing available, but I'm not opposed to slimming down our docblocks generally. 

I do think in cases where some parameters are specified they should all be included, and this PR catches a few of those mistakes.

system/Helpers/inflector_helper.php
paulbalandan: We can manually clean these later on.

paulbalandan:
> > PHPDoc should contain `@param` for all params.
> 

Actually, there's an opposite rule for this. `no_superfluous_phpdoc_tags`. This rule removes redundant phpdoc tags. However it retains them if (1) there's a description for the param, or (2) for arrays, the specific array shape is specified. So if we'll enable that, something like this would happen:

```diff
 /**
  * @param string   $bar This is bar
- * @param int      $baz
- * @param bool     $qux
- * @param array    $opts
  * @param string[] $defaults
- *
- * @return void
public function foo(string $bar, int $baz, bool $qux, array $opts, array $defaults = []): void;
```

My concern, though, is if PHPDoc v3 can still infer the type from the typehint (currently v2.9 relies on the phpdocs entirely). But this concern is not big as we have alternating rules for our use cases.

system/Helpers/inflector_helper.php
sfadschm: Just a follow up question: If we do manual cleanups, e.g. like in the other comment where you added the 'string' typehint.
Will those not be overwritten by the fixer as soon as someone tries to push changes to this file?

system/Email/Email.php
sfadschm: No, I think it is correct. I was just wondering how the fixer determined it. Turns out it wasn't the fixer 😆

system/Helpers/inflector_helper.php
paulbalandan: Can you clarify further?

sfadschm: I was wondering whether running the php-cs-fixer with `phpdoc_add_missing_param_annotation` again would replace your manual

     * @param string $type

with 

     * @param $type

in `Email.php`. I just tested, and it does not, so all good.

MGatner:
I'm not a fan of the "partial" parameter list but it seems like those are our two options. I would still like to hear from others but my hunch is to apply what you have now and then revisit this when we drop 7.3.

MGatner:
@paulbalandan was applying this your preference?

paulbalandan:
> @paulbalandan was applying this your preference?

Actually, no. I was just considering our API generation which uses PHPDoc v2.9 currently.

sfadschm:
> 
> 
> I'm not a fan of the "partial" parameter list but it seems like those are our two options. I would still like to hear from others but my hunch is to apply what you have now and then revisit this when we drop 7.3.

Personally, I prefer complete parameter lists. But I also prefer all parameters to have a description, so I am possibly more on the exaggerating side.
Would additional comments slow down loading of the framework due to larger file sizes?

paulbalandan:
Personally, I would prefer removing those redundant phpdocs unless they give additional value (like providing a description, or providing the internals for iterable types). With decent IDEs nowadays (like VS Code which I'm using), a properly typehinted element will have a beautifully generated phpdoc when hovered. Also, it may become a maintenance  burden to sync the manual phpdocs with the typehiints, which for example what we'd encountered in phpstorm where the mismatch breaks the code scanning.

MGatner:
Maybe for now we should split that rule out and address the rest of these rules. That would make looking at each of these options by themselves easier. More work for you though, sorry 😬

tests/_support/Log/Handlers/TestHandler.php
paulbalandan: Lonnie wants to remove those also. Exception is when the dividers are used as a form of header
```
//---------------------------
// Some comment
//---------------------------
```

I have a custom fixer that removes these dividers. Just need some refining to skip header usages.

paulbalandan:
Closing for now. I'll send later the commits as separate PRs.

PR #4899, opened 2021-07-02
Import global classes but fully qualify functions and constants

paulbalandan:
**Description**
An alternative approach to #4890, which I am using in my libraries (except for the import part).

Strategy:
- All native classes are imported with `use` statements.
- A select subset of constants are fully qualified `DIRECTORY_SEPARATOR, PHP_VERSION_ID, PHP_SAPI, PHP_INT_SIZE` as frequently used. (This one is debatable)
- Native functions which are optimized by the Zend engine are fully qualified. See https://github.com/php/php-src/blob/php-7.2.6/Zend/zend_compile.c and https://github.com/php/php-src/blob/php-7.2.6/ext/opcache/Optimizer/pass1_5.c

**Checklist:**
- [x] Securely signed commits

PR #4934, opened 2021-07-13
Define remaining rules

paulbalandan:
**Description**
Aside from #4933 , this defines the last remaining built-in fixers from php-cs-fixer into our ruleset. However, these are turned off by default.

- global_namespace_import, native_function_invocation, native_constant_invocation are depending on #4890 / #4899 
- header_comment is off by default. this is turned on by using `Factory::forLibrary()`
- use_arrow_functions is available only for PHP7.4+ code
- void_return adds the void return typehint so this breaks signature

**Checklist:**
- [x] Securely signed commits


MGatner:
Woohoo! So close! Thank you for all your work on these Paul. I'm excited to launch the new face of CI4

MGatner (approved):

PR #4940, opened 2021-07-17
Add param to all annotations

paulbalandan:
**Description**
First alternative: This PR ensures all params are annotated by @param tags.

**Checklist:**
- [x] Securely signed commits


sfadschm:
I still prefer this option, but I'm also fine with the other if there is a benefit to file size or such.

samsonasik (approved):

PR #4941, opened 2021-07-17
Remove superfluous phpdoc tags

paulbalandan:
**Description**
Alternative to #4940 : Adds param tags if only untyped. However, if deemed superfluous, it will be removed along with other superfluous tags.

**Checklist:**
- [x] Securely signed commits


sfadschm:
I just checked and the difference in file size between this and #4940 is `3 %` . As long as there is no down side to this, might be worth it, so I tend to change my mind 😄

paulbalandan:
By consensus, this is the winner. Merging.

PR #4945, opened 2021-07-17
Change some phpdocs to simple comments

paulbalandan:
**Description**
- Docblocks should only be used on structural elements. (Enabled only on a few files then disabled. Current behavior is not desired)
- Remaining rules are turned off as these breaks signatures. These can be used on migration rules instead.

**Checklist:**
- [x] Securely signed commits


MGatner (approved):
So basically, `/**` indicates a "docblock" so should only be used thus - block comments can use the single asterisk?

paulbalandan:
> So basically, `/**` indicates a "docblock" so should only be used thus - block comments can use the single asterisk?

Yes. `/**` is the marker for `T_DOC_COMMENT`. `/* */` is for multi-line comments

PR #4946, opened 2021-07-18
Remove useless code separator comments

paulbalandan:
**Description**
...unless used as headers (except the header use is 🙄 or 🤷‍♂️)

**Checklist:**
- [x] Securely signed commits


sfadschm:
Never thought there are ao many of them ...

system/CodeIgniter.php
```diff
@@ -1065,10 +1009,6 @@ protected function sendResponse()
      */
     protected function callExit($code)
     {
-        // @codeCoverageIgnoreStart
-        exit($code);
-        // @codeCoverageIgnoreEnd
+        exit($code); // @codeCoverageIgnore
```
MGatner, line 1012: Any chance there is a CS Fixer or Rector rule for converting these start-end for single lines? @samsonasik

MGatner (approved):
Wow the number of these caught me off guard! I can't promise I'm done with dividers but I'm glad to see most of these go.

system/CodeIgniter.php
samsonasik: removing before and after comment (before and next different line) is doable with rector, but I am not sure about placing comment on single line directly right after statement. Probably need to extends the `pComments()` method:

https://github.com/nikic/PHP-Parser/blob/c758510a37218d631fd10f67bca5bccbfef864fa/lib/PhpParser/PrettyPrinterAbstract.php#L449-L457

with some kind of custom attribute to flag single line.

PR #4951, opened 2021-07-19
CURLRequest raise exception only when `http_errors` is true

abydahana:
Add the statement to returning the exception only when "http_errors" configuration isn't set to "false".


paulbalandan:
As this is a change in behavior, please add tests demonstrating what will happen with and without the curl option supplied.

lonnieezell:
@abydahana  Are you able to submit tests for this?

abydahana:
Hello @lonnieezell, @paulbalandan. I'm trying to contribute but there are terms on the label that I don't fully understand. Like, where can i do a test?

paulbalandan:

PR #4966, opened 2021-07-26
[Rector] Apply MakeInheritedMethodVisibilitySameAsParentRector

samsonasik:
Make method visibility same as parent one with `MakeInheritedMethodVisibilitySameAsParentRector` rector rule.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -24,6 +24,7 @@ use Rector\CodeQuality\Rector\Return_\SimplifyUselessVariableRector;
 use Rector\CodeQuality\Rector\Ternary\UnnecessaryTernaryExpressionRector;
 use Rector\CodeQualityStrict\Rector\Variable\MoveVariableDeclarationNearReferenceRector;
 use Rector\CodingStyle\Rector\ClassMethod\FuncGetArgsToVariadicParamRector;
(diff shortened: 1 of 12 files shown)


paulbalandan (approved):

MGatner:
These are all changes from `public` to `protected`, which technically is a breaking change. All the database classes have magic callers so should be fine; the Image handlers are "underscore protected" so should probably be protected and `@internal` anyway; the rest are all part of the Test namespace and seem appropriate for how those classes are intended to be used. 

If we want to proceed with these I would still like these documented appropriately.

samsonasik:
@MGatner do you think it needs to be in the changelog https://github.com/codeigniter4/CodeIgniter4/blob/df419297c5f85c28e1cb27ac0dac66c408e6d0ea/user_guide_src/source/changelogs/v4.1.4.rst ?

MGatner:
@samsonasik Yes, sorry to be unclear - that is precisely what I want. I think it is fine jus to list them out. 

```
Breaking Changes

The following methods were changed from "public" to "protected" to match their parent class methods and better align with their uses:
* Alpha\Beta::foo()
* Gamma\Delta::bar()

samsonasik:
@MGatner changelog updated.

MGatner (approved):
Thanks!

PR #4969, opened 2021-07-27
Fix API repo preparation step

paulbalandan:
**Description**
The current apidocs workflow is largely based on the `apidoc` script. Up until now, there's a typo in the script preventing the removal of the contents of the `api/docs` folder:
> rm -rf api/docs*

As you can see, instead of deleting the contents this tries to delete files starting with `api/docs` which doesn't exist. This PR fixes that and makes the output more verbose.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/.github/workflows/deploy-apidocs.yml
+++ b/.github/workflows/deploy-apidocs.yml
@@ -55,9 +55,9 @@ jobs:
       - name: Prepare API repo
         working-directory: api
         run: |
-          rm -rf api/docs*
-          mkdir -p api/docs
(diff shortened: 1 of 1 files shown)


MGatner (approved):
`api/docs*` will still match `api/docs`. I'm all for this change but are you sure this was actually the issue?

paulbalandan:
Another issue is the misplaced git reset. If the current rm -rf works, which TIL, the changes are reset by the git command.

PR #4970, opened 2021-07-27
Fix properly the phpstan error in 0.12.93

paulbalandan:
**Description**
- Proper fix for phpstan error in #4967 (explode does not accept an empty string delimiter which the code does not take into account)
- Add tests for the `File::getDestination` method where this error arises

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -9,7 +9,6 @@ services:
 			- phpstan.rules.rule
 
 parameters:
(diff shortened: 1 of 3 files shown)


samsonasik (approved):

samsonasik:
Thank you @paulbalandan

PR #4972, opened 2021-07-27
Required rule error fix

### LINKED ISSUE #4971: Bug: false value is blocked in required rule
**Direction**
This is a bug question. I had found the error cause.
Is the empty function in required rule.
A disclaimer: You can clean more the code to a best development experience.

**Describe the bug**
When I use the required rule, the value false is blocked.

**CodeIgniter 4 version**
The composer version presents this error.

**Affected module(s)**
The module Validation/Rules is the only affected.

To reproduce this error, you can create a database table with a boolean value or number in a field. 
Before, you try add false  value in the field.

**Context**
 - OS: [Manjaro Linux, latest]
 - Web server [spark server, latest]
 - PHP version [8.0.8]

OzRib:
# Fixed #4971 

**Description**
I modified the required rule in Vallidation/Rules. 
I did it because the required rule blocks the false value when the user try to send a false value to the database.
This problem affected one of my applications when I update the  available value in a product field.
So I decided to solve this problem and helps the community.
The code got bigger but I guarantee it is scalable and fast. The needed function is executed only once.
The code got more readable too.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -245,7 +245,7 @@ class Rules
      */
     public function required($str = null): bool
     {
-        if (is_object($str)) {
+        if (is_object($str) || is_bool($str)) {
             return true;
         }
 
--- a/tests/system/Validation/RulesTest.php
+++ b/tests/system/Validation/RulesTest.php
@@ -270,11 +270,6 @@ final class RulesTest extends CIUnitTestCase
                 ['foo' => null],
                 false,
             ],
-            [
-                ['foo' => 'permit_empty|required'],
-                ['foo' => false],
-                false,
-            ],
             // This tests will return true because the input data is trimmed
             [
                 ['foo' => 'permit_empty|required'],


MGatner (changes requested):
I think you are right that there is a bug or inconsistency. User Guide on `required`:

> Fails if the field is an empty array, empty
string, null or false.

We don't seem to handle null or false properly. However there are some issues with your PR in order to get this mergeable. 

1. `develop` is currently locked to code style changes. Please open a new PR against branch `4.2`. 
2. Please mind the code styles - you can use PHP CS Fixer to apply the changes automatically. 
3. In my opinion an array of anonymous functions is much less clear. I would prefer to see this as a series of `if` statements for clarity. 
4. Why did you remove the test? To help ensure we aren't making breaking changes all tests should still pass. `required|permit_empty` is an edge case that we need explicit handling for. 
5. Please make sure any new code added is fully covered by the test cases, or add additional as needed.

paulbalandan (changes requested):
In addition to @MGatner 's comments, I agree that we can swap the array of lambdas with series of `if` guards.

Another, you have a handful of checks basically returning true. I would be in favor of those as the default response, i.e., `return true;` as the last line. All other with logic switch would be handled beforehand.

OzRib:
Okay guys.
@MGatner I remove a test closure because it's claims false is empty.
In addition. Okay. I'll create a new pull request to branch 4.2.
I had modified the coding style. No worries. 
I'm sorry. I'm still a begginer.

PR #4974, opened 2021-07-28
IncomingRequest - Trim trailing slash

najdanovicivan:
**Description**
If trailing slash is forced via apache or nginx the router will be unable to find the appropriate route. The router will try to match path from request uri with trailing slash but it is impossible to define such route in the route collection since slash is used as separator. 

This changes the way request uri is handled so that trailing slash is removed at the end which enables the router to handle the url in both cases with or without trailing slash 

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -328,7 +328,9 @@ class Router implements RouterInterface
             return false;
         }
 
-        $uri = trim($uri, '/ ') ?: '/';
+        $uri = $uri === '/'
+            ? $uri
+            : trim($uri, '/ ');
 
         // Loop through the route array looking for wildcards
         foreach ($routes as $key => $val) {
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -328,9 +328,7 @@ class Router implements RouterInterface
             return false;
         }
 
-        $uri = $uri === '/'
-            ? $uri
-            : ltrim($uri, '/ ');
+        $uri = trim($uri, '/ ') ?: '/';
 
         // Loop through the route array looking for wildcards
         foreach ($routes as $key => $val) {
--- a/tests/system/Router/RouterTest.php
+++ b/tests/system/Router/RouterTest.php
@@ -103,6 +103,16 @@ final class RouterTest extends CIUnitTestCase
         $this->assertSame('index', $router->methodName());
     }
 
+    public function testURIWithTrailingSlashMapsToController()
+    {
+        $router = new Router($this->collection, $this->request);
+
+        $router->handle('users/');
+
+        $this->assertSame('\Users', $router->controllerName());
+        $this->assertSame('index', $router->methodName());
+    }
+
     public function testURIMapsToControllerAltMethod()
     {
         $router = new Router($this->collection, $this->request);
@@ -161,6 +171,16 @@ final class RouterTest extends CIUnitTestCase
         $router = new Router($this->collection, $this->request);
 
         $router->handle('objects/123/sort/abc/FOO');
+
+        $this->assertSame('objectsSortCreate', $router->methodName());
+        $this->assertSame(['123', 'abc', 'FOO'], $router->params());
+    }
+
+    public function testURIWithTrailingSlashMapsParamsWithMany()
+    {
+        $router = new Router($this->collection, $this->request);
+
+        $router->handle('objects/123/sort/abc/FOO/');
 
         $this->assertSame('objectsSortCreate', $router->methodName());
         $this->assertSame(['123', 'abc', 'FOO'], $router->params());


paulbalandan (commented):
Can you add a test for this?

najdanovicivan:
@paulbalandan Since this is only being handled inside IncomingRequest instead of Router I've added test for the Incoming request. 

With this implementation URI path will always be returned without trailing slash. I'm not sure it this is the best way to do it since with this implementation it is impossible to detect if url have trailing slash or not. The better solution might be to handle the stripping of slash inside Router class instead of stripping it from the URI path in incoming request. 

With current setup $request->getPath() will return the same value of 'fruits/banana' for both  $_SERVER['REQUEST_URI'] = '/index.php/fruits/banana/'; and  $_SERVER['REQUEST_URI'] = '/sub/folder/index.php/fruits/banana';

MGatner:
I'm hesitant about this one with all the URI issues we've had in the past. My hunch is that we need a way to distinguish between `fruits/banana` and `fruits/banana/`, but I'm not sure if that actually matters? Anyone know what the RFC says about trailing slashes?

najdanovicivan:
@MGatner looking at PSR-7 underlying RCF document I found this https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.4

But I also found this middleware https://github.com/middlewares/trailing-slash which can do a redirection to force urls either with or without trailing slash

But none of this defines if there should be slash there or no

MGatner:
After reading those and looking around a little more I think it is more clear that trailing slashes *may differentiate a URI*. For framework purposes I think this means we need to leave the URI intact for the developer. However, I also think it is logical for the Router to treat them as equivalent, since we don't have a meaningful way to differentiate the trailing slash in Route Collection anyways. 

Instead of changing IncomingRequest, check out `Router::checkRoutes()`. There are two `ltrim()` calls in there that could be changed to `trim()` to ignore trailing slashes during route comparison. If you decide to implement this, we would need some updated test cases and maybe a side note in the User Guide.

najdanovicivan:
Sure thing. I also found another issue that will occur if this is done in such way. Web Page Caching uses url to generate the md5 filename which means urls with slash and without will have different caches which in the end will result in cache using double amount of space if urls work with and without slash.

So maybe generating the cache filename should also strip the trailing slash

lonnieezell:
I don't see a problem stripping trailing slash for the cache filename.

najdanovicivan:
I've updated the code trim the path in Router class. I also fixed the trailing slash in page cache and while I was there I added method to be used for manually purging single cached page

system/CodeIgniter.php
```diff
@@ -630,6 +630,25 @@ public function cachePage(Cache $config)
...
+     * Purges the cached page
+     *
+     * @param string     $uri    Uri
+     * @param Cache|null $config Config
+     */
+    public function purgeCachedPage(string $uri, ?Cache $config = null): bool
+    {
+        $uri = new URI($uri);
+
+        if (($config ?? new Cache())->cacheQueryString) {
```
MGatner, line 643: This should fall back on the shared instance. 
```suggestion
        if (($config ?? config('Cache'))->cacheQueryString) {
```

MGatner (changes requested):
Much better! I like the new method but it will need test coverage. Please either add the tests or pull the method for a separate PR.

najdanovicivan:
I've split the page cache in separate PR #5073 so that this one can be merged

MGatner (approved):

system/Router/Router.php
```diff
@@ -330,7 +330,7 @@ protected function checkRoutes(string $uri): bool
 
         $uri = $uri === '/'
```
paulbalandan, line 331: Can we make this a one-liner?

najdanovicivan: How do you mean one-liner it is already one line ?
 
```
        $uri = $uri === '/'
            ? $uri
            : trim($uri, '/ ');
```

Maybe

```
$url = trim($url) ?: '/';
```

paulbalandan: Yes.
```php
$uri = trim($uri, '/ ') ?: '/';
```

system/Router/Router.php
```diff
@@ -328,9 +328,7 @@ protected function checkRoutes(string $uri): bool
             return false;
         }
 
-        $uri = $uri === '/'
-            ? $uri
-            : ltrim($uri, '/ ');
+        $uri = trim($uri, '/ ') ?: '/';
```
paulbalandan, line 331: Minor typo I think why tests are failing.
```suggestion
        $uri = ltrim($uri, '/ ') ?: '/';
```

kenjis: @paulbalandan 
The following two code is not the same.
What if `$uri` is `"0"`.

```php
$uri = $uri === '/'
            ? $uri
            : trim($uri, '/ ');
```
```php
$uri = trim($uri, '/ ') ?: '/';
```

kenjis: @najdanovicivan This would make all tests pass.

```suggestion
        $uri = $uri === '/'
            ? $uri
            : trim($uri, '/ ');
```

najdanovicivan: That was my initial implementation. But we thought one liner will do the job. I completely missed out the empty string or '0' string cases

@paulbalandan the whole idea with this is to change from ltrim to trim so that urls with trailing slash are properly routed

kenjis (approved):

paulbalandan (approved):

PR #4975, opened 2021-07-28
Model - Fix zero value primary key

najdanovicivan:
**Description**
Primary key in SQL can have value of 0 or empty string. This modifies the model to use isset to handle such values instead of doing an empty check so that rows which such keys can be inserted/updated in the DB

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -225,7 +225,7 @@ class Model extends BaseModel
 
         // Require non empty primaryKey when
         // not using auto-increment feature
-        if (! $this->useAutoIncrement && empty($data[$this->primaryKey])) {
+        if (! $this->useAutoIncrement && ! isset($data[$this->primaryKey])) {
             throw DataException::forEmptyPrimaryKey('insert');
         }
 
@@ -263,7 +263,7 @@ class Model extends BaseModel
             foreach ($set as $row) {
                 // Require non empty primaryKey when
                 // not using auto-increment feature
-                if (! $this->useAutoIncrement && empty($row[$this->primaryKey])) {
+                if (! $this->useAutoIncrement && ! isset($row[$this->primaryKey])) {
                     throw DataException::forEmptyPrimaryKey('insertBatch');
                 }
             }
@@ -332,7 +332,7 @@ class Model extends BaseModel
     {
         $builder = $this->builder();
 
-        if ($id) {
+        if (isset($id)) {
             $builder = $builder->whereIn($this->primaryKey, $id);
         }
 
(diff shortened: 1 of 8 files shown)


system/Model.php
```diff
@@ -662,10 +662,8 @@ protected function objectToRawArray($data, bool $onlyChanged = true, bool $recur
         $properties = parent::objectToRawArray($data, $onlyChanged);
 
         // Always grab the primary key otherwise updates will fail.
-        if (
-            method_exists($data, 'toRawArray') && (! empty($properties) && ! empty($this->primaryKey) && ! in_array($this->primaryKey, $properties, true)
-            && ! empty($data->{$this->primaryKey}))
-        ) {
+        if (method_exists($data, 'toRawArray') && (! empty($properties) && ! empty($this->primaryKey) && ! in_array($this->primaryKey, $properties, true)
+                && isset($data->{$this->primaryKey}))) {
```
paulbalandan, line 666:
```suggestion
        if (
            method_exists($data, 'toRawArray')
            && (! empty($properties) && ! empty($this->primaryKey)
            && ! in_array($this->primaryKey, $properties, true)
            && isset($data->{$this->primaryKey}))
        ) {
```

MGatner (changes requested):
Please add some tests for primary keys of '', 0, '0', false, and 'false'. 

Also while you're here, can we replace those two awful composite `if` statements with a static function? Something like `getObjectPrimaryKey()`

lonnieezell:
@najdanovicivan are you able to make these changes? If not, I'm closing.

najdanovicivan:
I’m on it.

najdanovicivan:
@MGatner we already have getIdValue() method so I've used it in place of those `if` statements

Regarding the other place where we have those nasty `if` statements it is a deprecated static method and because it's static we cannot call getIdValue() there so I left it unchanged is should be removed in future releases anyway

I also added couple of tests.  I haven't worked with DataBase test before on CI so it took me a while to figure it out how those work.

najdanovicivan:
I'm not sure what's going on with test on 4.2 branch some of them are failing.  

What I also see now that MSSQL, SQLite and Postgres behave differently with empty values

MGatner:
It looks like the failing tests are related to the way soft deletes are handled with primary keys. I'm on mobile so I can't dig very far right now, but some other tests might need changing for this to work. 

In general we have a few open issues/PRs around boolean/empty values in the database layer. I think these will all need to be pulled together and a final decision made on how to handle things before `4.2` will be predictably consistent.

MGatner:
`4.2` branch has been rebased and merged into `develop`! Those outstanding changes should all be in place - @najdanovicivan could you rebase this and point it back to `develop`? That should give us a good idea of what remains to be fixed.

MGatner:
Tests are still failing.

paulbalandan:
I am seeing no progress in this since September. Please open a new PR with updated code where tests are passing.

PR #4977, opened 2021-07-28
required rule error fix

### LINKED ISSUE #4971: Bug: false value is blocked in required rule

OzRib:
Fixes #4971

**Description**
I modified the required rule in Validation/Rules. 
I did it because the required rule blocks the false value when the user try to send a false value to the database.
This problem affected one of my applications when I update the  available value in a product field.
So I decided to solve this problem and helps the community.
The modification is short. Nothing too big.
I updated one of the tests because my objective is modify this behavior.
This a new pull request. More clean, more readable. I think more acceptable.
A little modification but I needed modify this.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -324,7 +324,7 @@ class Rules
      */
     public function required($str = null): bool
     {
-        if (is_object($str)) {
+        if (is_object($str) || is_bool($str)) {
             return true;
         }
 
--- a/tests/system/Validation/RulesTest.php
+++ b/tests/system/Validation/RulesTest.php
@@ -144,6 +144,23 @@ class RulesTest extends CIUnitTestCase
 
     //--------------------------------------------------------------------
 
+    public function testRequiredBoolean()
+    {
+        $data = [
+            'foo' => true,
+            'bar' => false
+        ];
+
+        $this->validation->setRules([
+            'foo' => 'required',
+            'bar' => 'required'
+        ]);
+
+        $this->assertTrue($this->validation->run($data));
+    }
+
+    //--------------------------------------------------------------------
+
     /**
      * @dataProvider ifExistProvider
      *
@@ -280,10 +297,11 @@ class RulesTest extends CIUnitTestCase
                 ['foo' => null],
                 false,
             ],
+            // This test will return true because false value is boolean
             [
                 ['foo' => 'permit_empty|required'],
                 ['foo' => false],
-                false,
+                true,
             ],
             // This tests will return true because the input data is trimmed
             [


MGatner:
Something still isn't right. I think the goal should be to make the code match the User Guide for `required`:

> Fails if the field is an empty array, empty
string, null or false.

Passing when `is_bool()` means that `false` will pass. I was shocked the tests didn't catch this, only to find that we don't have a test case for this!

My recommendation: update the test files *first* to have explicit cases for each of those inputs mentioned in the User Guide, *then* fix the code to pass the tests. 

One more thing to pay attention to: it seems that `permit_empty` should be the logical opposite of `required`, *except* that when both are present `required` should take precedence.

MGatner (changes requested):
Please update tests for `false`, etc, then make sure the code aligns.

OzRib:
@MGatner okay. Doing

OzRib:
@MGatner I've finished my modifications. 

- [x] Added explicit tests
- [x] Updated tests cases
- [x] Fixed code to pass the tests

But I have a question. Where I can modify the documentation? 
Or only you guys can update the documentation?
My suggestion is modify to accept false value because when the PHP calls a database request, boolean values are turned in integer values. 
Without conflicts.
At least in the MySQL database.

MGatner:
You can modify the User Guide files in this same repo - look in the **user_guide** directory in the root. 

But before you go further: we cannot change what this function accepts arbitrarily. The User Guide is very clear that `false` is not a passing input. I ran a few edge case tests myself:

https://3v4l.org/rTPMd

Surprisingly, the current framework code works as expected. I think it is because `trim()` only accepts a `string` so when `null` or `false` are passed in they get cast to `''`. I would still like to see this method rewritten to make that a little more clear, as well as actual tests added for these cases, but I don't think there is a bug.

OzRib:
Okay. I'll close the PR. 
If I send a new PR will be to turn the code more clean.

PR #4978, opened 2021-07-29
[Rector] Apply Rector: RemoveDuplicatedArrayKeyRector

samsonasik:
Remove duplicated key in defined arrays with `RemoveDuplicatedArrayKeyRector` rector rule.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/tests/system/Helpers/InflectorHelperTest.php
+++ b/tests/system/Helpers/InflectorHelperTest.php
@@ -244,12 +244,6 @@ final class InflectorHelperTest extends CIUnitTestCase
     public function testOrdinal()
     {
         $suffixes = [
-            'st' => 1,
-            'nd' => 2,
-            'rd' => 3,
-            'th' => 4,
-            'th' => 11,
-            'th' => 20,
             'st' => 21,
             'nd' => 22,
             'rd' => 23,
--- a/tests/system/Helpers/InflectorHelperTest.php
+++ b/tests/system/Helpers/InflectorHelperTest.php
@@ -241,19 +241,29 @@ final class InflectorHelperTest extends CIUnitTestCase
         }
     }
 
-    public function testOrdinal()
+    public function provideOrdinal()
     {
-        $suffixes = [
-            'st' => 21,
-            'nd' => 22,
-            'rd' => 23,
-            'th' => 24,
+        return [
+            ['st', 1],
+            ['nd', 2],
+            ['rd', 3],
+            ['th', 4],
+            ['th', 11],
+            ['th', 20],
+            ['st', 21],
+            ['nd', 22],
+            ['rd', 23],
+            ['th', 24],
         ];
+    }
 
-        foreach ($suffixes as $suffix => $number) {
-            $ordinal = ordinal($number);
-            $this->assertSame($suffix, $ordinal);
-        }
+    /**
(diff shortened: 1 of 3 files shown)


tests/system/Validation/CreditCardRulesTest.php
```diff
@@ -148,46 +148,6 @@ public function creditCardProvider()
                 '6011 0009 9013 9424',
                 true,
             ],
-            'jcb1' => [
-                'jcb',
-                '3530 1113 3330 0000',
-                true,
-            ],
```
samsonasik, line 155: this possibly need a different key if the value is different?

paulbalandan: Since these are from the data provider, we can just remove the string keys to use numeric keys instead, or change these duplicate keys to use different names

tests/system/Helpers/InflectorHelperTest.php
```diff
@@ -244,12 +244,6 @@ public function testDasherize()
     public function testOrdinal()
     {
         $suffixes = [
-            'st' => 1,
```
paulbalandan, line 247: This test should be revised to use data provider instead of deleting the test values

tests/system/Validation/CreditCardRulesTest.php
samsonasik: implemented.

tests/system/Helpers/InflectorHelperTest.php
samsonasik: implemented.

tests/system/Helpers/InflectorHelperTest.php
```diff
@@ -241,25 +241,29 @@ public function testDasherize()
...
-            $ordinal = ordinal($number);
-            $this->assertSame($suffix, $ordinal);
-        }
+    /**
+     * @dataProvider provideOrdinal
+     */
+    public function testOrdinal(string $suffix, int $number)
+    {
+        $ordinal = ordinal($number);
+        $this->assertSame($suffix, $ordinal);
```
paulbalandan, line 266:
```suggestion
        $this->assertSame($suffix, ordinal($number));
```

paulbalandan (approved):

MGatner (changes requested):
The array keys on the provider in **CreditCardRulesTest.php** should stay. These are used to name data sets during test failure. (e.g. "testValidCCNumber failed with data set 'not_numierc'") and make it easier to track down the actual failure - otherwise it enumerates all the data from the provider which can make for a messy output.

samsonasik:
@MGatner when the key are same, only last that used, do you think that should be named indexed, eg: Mastercard1, etc

tests/system/Validation/CreditCardRulesTest.php
MGatner: Sorry, I missed this conversation originally!

There is some merit to keeping the keys (see my top-level comment) but they will need to be deduplicated. The names don't really matter very much, I'm fine with @samsonasik 's proposal of just tacking a digit on the end.

MGatner:
Yes

samsonasik:
@MGatner implemented

samsonasik:
@MGatner I am merging it

PR #4979, opened 2021-07-30
Update of BaseModel.php to fix INSERT issue

katie1348:
Applied the fix that enables the INSERT to work correctly when inserting a row that already contains data.

Description
Applied the fix suggested by @arbitmcdonald in Bug: onlyChanged is currently true for INSERT codeigniter4#4247

Checklist:

[x]  Securely signed commits
[x]  Component(s) with PHPdocs
[x]  Unit testing, with >80% coverage
[x]  User guide updated
[x]  Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1574,7 +1574,7 @@ abstract class BaseModel
         // properties representing the collection elements, we need to grab
         // them as an array.
         if (is_object($data) && ! $data instanceof stdClass) {
-            $data = $this->objectToArray($data, true, true);
+            $data = $this->objectToArray($data, ($type == 'update'), true);
         }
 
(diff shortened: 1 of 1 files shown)


paulbalandan:
We would need tests to verify that the fix is actually working.

paulbalandan:
You can take at a look at `tests/system/Models/` dir to check the tests for the Model.

katie1348:
OK, so it looks like the error is a coding standards error of == rather than ===.

Does his mean I need to start the process from the beginning to make this change?

paulbalandan:
No need to open new PR. You can do the edit directly here.

paulbalandan:
Replaced by #4980

PR #4980, opened 2021-07-30
fix: BaseModel::insert() may not pass all the values from Entity

### LINKED ISSUE #4247: Bug: onlyChanged is currently true for INSERT
**Describe the bug**

CI4 doesn't respect the default values of entity attributes on INSERT. Currently only the values that have changed are passed to the INSERT query. This means that if the attributes default value is 500, and I try to insert an entity after setting the attributes value to 500, no value is written to the database at all (the default value from the table would be used, as CI would not include the column in the INSERT). 

So [the following statement](https://codeigniter4.github.io/userguide/models/entities.html#working-with-the-entity-class) is not currently accurate:
image

**CodeIgniter 4 version**

v4.1.1

INSERT of entity should include all values, changed and unchanged. INSERT currently only includes changed values. 

**To reproduce:**

- Create an entity class
- Define an attribute with a default value:
```php
namespace App\Entities;

use CodeIgniter\Entity;

class BugExample extends Entity
{
    protected $attributes = [
        'the_attribute' => 500,
    ];

    public function setTheAttribute(int $value)
    {
        $this->attributes['the_attribute'] = 0;
        return $this;
    }

    public function getTheAttribute(): int
    {
        return (int) $this->attributes['the_attribute'];
    }
}
```
- Create an entity, set the attrib value to match the default, insert:
```php
$bugExample = new BugExample();
$bugExample->setTheAttribute(500);
$this->insert($bugExample);
```
- The insert query will not include `the_attribute`

**The root cause:**

- Model `insert()` calls `model->transformDataToArray($data, 'insert');`
- `transformDataToArray()` converts the entity object to an array using `baseModel->objectToArray($data, true, true);`. The 2nd parameter `true` refers to `bool $onlyChanged` - which means object to array conversion will only include changed properties.

**The solution:**

- Change `$data = $this->objectToArray($data, true, true);` to `$data = $this->objectToArray($data, ($type == 'update'), true);` in BaseModel.php:1690. This will mean that all attributes will be included for INSERT and only changed attributes will be included for UPDATE.
- I have tested this, INSERT and UPDATE, and received the expected outcome.

```php
protected function transformDataToArray($data, string $type): array
{
    if (! in_array($type, ['insert', 'update'], true))
    {
        throw new InvalidArgumentException(sprintf('Invalid type "%s" used upon transforming data to array.', $type));
    }

    if (empty($data))
    {
        throw DataException::forEmptyDataset($type);
    }

    // If $data is using a custom class with public or protected
    // properties representing the collection elements, we need to grab
    // them as an array.
    if (is_object($data) && ! $data instanceof stdClass)
    {
        ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
        ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
        // THE LINE BELOW.... 
            $data = $this->objectToArray($data, true, true);
        // NEEDS TO BE CHANGED TO: 
            $data = $this->objectToArray($data, ($type == 'update'), true);
        ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
        ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
    }

    // If it's still a stdClass, go ahead and convert to
    // an array so doProtectFields and other model methods
    // don't have to do special checks.
    if (is_object($data))
    {
        $data = (array) $data;
    }

    // If it's still empty here, means $data is no change or is empty object
    if (empty($data))
    {
        throw DataException::forEmptyDataset($type);
    }

    return $data;
}
```

**Workaround options for those affected:**
1. Set all attribute default values to `null` and set the default values in the database table, or:
2. Set all attribute default values to `null` and create a constructor in the entity class to set your default values:
```php
    public function __construct()
    {
        parent::__construct();
        $this->setTheAttribute(500);
    }
```

katie1348:
Fixes #4247 

Updated the parameter passed, so the function behaves differently for Insert and Update, aligned with what would be expected. 

**Description**
Added in a test to use all the fields when Inserting and to do a changed fields when Updating. The old version always only inserted changed fields, which is not what is expected.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


lonnieezell:
@katie1348 Your description says you added in a test but I don't see any in the PR. Are you able to add that?

MichalPB1:
What is the status of? It also resolves my problem

paulbalandan:
@MichalPB1 if you can help add a test for this change then it is mergeable.

katie1348:
I would write a test, but I don't know how.

kenjis:
See 
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#unit-testing
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Models/InsertModelTest.php

katie1348:
Hi Kenjis,

Thanks for the pointers and I have the system installed and the repo updated and cloned to my Mac.

When I run:

phpunit

I get the following error (stack trace omitted)

PHP Fatal error:  Uncaught Error: Class "CodeIgniter\Models\LiveModelTestCase" not found in /Users/user/Sites/CodeIgniter4/tests/system/Models/AffectedRowsTest.php:19

I have no idea why a clean unchanged clone of the repo would throw up this error, any thoughts?

Thanks in advance.

kenjis:
@katie1348 You maybe miss something.

This is how to:
```sh-session
$ git clone git@github.com:<your-name>/CodeIgniter4.git
$ cd CodeIgniter4/
$ composer install
$ vendor/bin/phpunit
```

katie1348:
@kenjis  Thank you for your help, much appreciated.

That now has it working, mostly. It fails some tests as I don't have Redis etc installed and will not. But it looks like it is up and running enough to write the necessary test.

Hopefully next week.

kenjis:
@katie1348 

You can run one test file:
```sh-session
$ vendor/bin/phpunit tests/system/HTTP/RequestTest.php
```

Or exclude live tests:

```sh-session
$ vendor/bin/phpunit --exclude-group DatabaseLive,CacheLive
```

katie1348:
@kenjis 

Once again, thank you for the assistance. 

With your help, I have been able to get the test working and it is now ready for additional checking etc.

I am now stuck on the cs-fixer. I cannot seem to get it to run correctly. I have installed it with homebrew and it looks like it is working sometimes, but it does not seem to change the files.

When run from the command line, it get an error:

Fatal error: Uncaught Error: Class "Nexus\CsConfig\FixerGenerator" not found in /Users/user/Sites/CodeIgniter4/.php-cs-fixer.dist.php:52

I am once again stuck,  your assistance would be appreciated.

kenjis:
@katie1348 See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#php-style

> I have installed it with homebrew and it looks like it is working sometimes, but it does not seem to change the files.

This is a common mistake. We use php-cs-fixer for this project, and you have already installed it
via Composer.

And you can run with `composer`.

```
$ composer cs-fix
```

katie1348:
@kenjis Thank you for that assistance. It got me over the hump.

OK, I have uploaded the changes to my fork, in the correct branch.

How do I go about getting the 6 Workflows awaiting approval to be run?

kenjis:
> How do I go about getting the 6 Workflows awaiting approval to be run?

You can't do anything. I approved.

katie1348:
Updated to fix Reactor failures.

katie1348:
@kenjis 

Thanks for all your help in getting this done. I don't think I could have done it without you.

Is there anything else I need to do?

tests/system/Models/InsertModelTest.php
```diff
@@ -286,4 +288,29 @@ public function testInsertWithSetAndEscape(): void
...
+
+        $resultEntity->name = 'Katherine';
+
+        $this->model->insert($resultEntity);
+        $this->assertGreaterThan(0, $this->model->getInsertID());
+
+        $resultEntity2 = $this->model->where('name', 'Katherine')->first();
+        $this->assertGreaterThan(0, $this->model->getInsertID());
+        $this->assertSame('mark@example.com', $resultEntity2->email);
+    }
```
kenjis, line 315: I'm sorry, but I don't understand this test's intention.

Test code needs to be very easily to understand.

And the code works without errors, but it seem it is a wrong usage of Model/Entity.
You create an Entity with name Mark.
And you changed the name to Katherine.
I don't imagine that Mark becomes Katherine.

kenjis: This PR is to fix #4247, and it has sample code like:
```php
$bugExample = new BugExample();
$bugExample->setTheAttribute(500);
$this->insert($bugExample);
// The insert query will not include the_attribute
```
Why don't you write test like it?

tests/_support/Entity/User.php
```diff
@@ -0,0 +1,41 @@
...
+     *  $datamap = [
+     *      'db_name' => 'class_name'
+     *  ];
+     */
+    protected $datamap = [];
+
+    /**
+     * Define properties that are automatically converted to Time instances.
+     */
+    // protected $dates = ['created_at', 'updated_at', 'deleted_at', 'expiry'];
```
kenjis, line 33: We don't need the commented code like this.

kenjis (changes requested):
@katie1348 I added review comments.
Please fix your test code.

If you have any questions, feel free to ask.

tests/system/Models/InsertModelTest.php
katie1348: @kenjis, I'm sorry if I have done this incorrectly.

The only way I have seen the bug appear is when doing a write (insert) read back the value and the write (insert) the original row with a change. In this case, only the changed item is saved back. This is what the test checks for. I have run the test with the old code, and tested the old path and it fails.

So the test, while not perhaps obvious is the code that provokes the error and tests that it is fixed.

Now, there might be other ways of testing for it, but I don't know what they are. This is what I was doing when it showed up.

As I understand it, the code must be run with a non-Standard Class object, which means I need to use a Model and Entity. Once again, this might be my lack of understanding, but I don't know how else to do it.

tests/_support/Entity/User.php
katie1348: I thought those comments were standard for an Entity. I will be happy to remove all the comments.

tests/_support/Models/UserObjModel.php
```diff
@@ -0,0 +1,31 @@
...
+        'email',
+        'country',
+        'deleted_at',
+    ];
+    protected $returnType     = \Tests\Support\Entity\User::class;
+    protected $useSoftDeletes = true;
+    protected $dateFormat     = 'datetime';
+    public $name              = '';
+    public $email             = '';
+    public $country           = '';
```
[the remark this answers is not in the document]
katie1348: I simply copied the UserModel.php and made the change that I needed.

I thought it best to make them similar. I can remove the public if you want?

tests/_support/Entity/User.php
katie1348: Comments from the code all deleted.

tests/_support/Models/UserObjModel.php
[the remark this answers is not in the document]
katie1348: Public properties deleted.

katie1348:
Requested changes made.

tests/_support/Entity/User.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Tests\Support\Entity;
+
+use CodeIgniter\Entity;
+
+class User extends Entity
+{
+    protected $datamap = [];
+    protected $dates   = ['created_at', 'expiry'];
+    protected $casts   = [];
+}
```
kenjis, line 21: `CodeIgniter\Entity` is deprecated.

```suggestion
use CodeIgniter\Entity\Entity;

class User extends Entity
{
    protected $attributes = [
        'country' => 'India',
    ];
}
```

tests/system/Models/InsertModelTest.php
kenjis, line 315:
```suggestion
    /**
     * @see https://github.com/codeigniter4/CodeIgniter4/issues/4247
     */
    public function testInsertWithDefaultValue(): void
    {
        $this->createModel(UserObjModel::class);

        $entity             = new User();
        $entity->name       = 'Mark';
        $entity->email      = 'mark@example.com';
        $entity->country    = 'India';  // same as the default
        $entity->deleted    = 0;
        $entity->created_at = new Time('now');

        $this->model->insert($entity);

        $id = $this->model->getInsertID();
        $this->assertSame($entity->country, $this->model->find($id)->country);
    }
```

kenjis: > So the test, while not perhaps obvious is the code that provokes the error and tests that it is fixed.

Yes, you got the bug and fixed it. And your test failed before the fix.
It is very good practice that writing a test that detects a bug, and fix the bug.

> As I understand it, the code must be run with a non-Standard Class object, which means I need to use a Model and Entity.

Yes, you're correct. This is a bug related with `Entity::toRawArray()`.

I suggested the test code referring to #4247.

katie1348:
@kenjis I have accepted your changes to the tests.

kenjis (approved):
LGTM

katie1348:
@kenjis Thanks for all your help and patience.

system/BaseModel.php
```diff
@@ -1578,7 +1578,7 @@ protected function transformDataToArray($data, string $type): array
         // properties representing the collection elements, we need to grab
         // them as an array.
         if (is_object($data) && ! $data instanceof stdClass) {
-            $data = $this->objectToArray($data, true, true);
+            $data = $this->objectToArray($data, ($type === 'update'), true);
```
paulbalandan, line 1581: The parenthesis is not needed.
```suggestion
            $data = $this->objectToArray($data, $type === 'update', true);
```

paulbalandan:
Also, please squash before merging to remove the merge commit.

system/BaseModel.php
kenjis: It is difficult to read.

paulbalandan: Hmm. On second look I guess you're right.

paulbalandan (approved):

kenjis:
@katie1348 Congratulations! You are now a contributor to this project. 🎉

MGatner:
Thanks all for the work! Sorry this one sat so long, looks like a quality PR - just slipped through the cracks.

katie1348:
> @katie1348 Congratulations! You are now a contributor to this project. 🎉

Only because you did all the hand holding I needed. Thank you once again.

PR #4987, opened 2021-08-05
Update Validation.php

maniaba:
Replacing values for custom user message


lonnieezell:
This matches current code base. Closing. 

https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Validation/Validation.php#L296

PR #4955, opened 2021-07-21
SQLSRV Driver: Update getFullName function to add dynamic schema support

wradstok:
This commit allows one to set the schema for the table in the query builder. For example we can do:

`$this-db->table("myschema.users")->select("userid")`

to create

`SELECT "userid" FROM "myschema"."users"`
 rather than the default
`SELECT "userid" FROM "dbo"."users"`

Each pull request should address a single issue and have a meaningful title.

**Description**
Explain what you have changed, and why.

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

---------Remove from here down in your description----------

**Notes**
- Pull requests must be in English
- If the PR solves an issue, reference it with a suitable verb and the issue number
(e.g. fixes <hash>12345)
- Unsolicited pull requests will be considered, but there is no guarantee of acceptance
- Pull requests should be from a feature branch in the contributor's fork of the repository
  to the develop branch of the project repository

PR #4990, opened 2021-08-05
This is to allow changing SQLSRV schema from dbo to any other.

hdrodriguez:

**Description**

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

This change in getFullName function allows defining different schemas in the table parameter of the models, example "adm.tablename" or sales.users etc.

### DIFF
--- a/system/Database/SQLSRV/Builder.php
+++ b/system/Database/SQLSRV/Builder.php
@@ -230,9 +230,33 @@ class Builder extends BaseBuilder
 
     /**
      * Get full name of the table
+     *
+     * @param string $tableName
+     *
+     * @return string
+     */
+
+    private function validateSchema(string $tableName): string
+    {
+        $tableExploded = explode('.', $tableName);
+        if (count($tableExploded) === 2)
+        {
+            $this->db->schema = str_replace('"', '', $tableExploded[0]);
+            $tableName        = $tableExploded[1];
+        }
+
+        return $tableName;
+    }
+
+
+
+
+    /**
+     * Get full name of the table
      */
     private function getFullName(string $table): string
     {
+        $table = $this->validateSchema($table);
         $alias = '';
 
         if (strpos($table, ' ') !== false) {
--- a/system/Database/SQLSRV/Builder.php
+++ b/system/Database/SQLSRV/Builder.php
@@ -231,7 +231,6 @@ class Builder extends BaseBuilder
     /**
      * Get full name of the table
      */
-
     private function validateSchema(string $tableName): string
     {
         $tableExploded = explode('.', $tableName);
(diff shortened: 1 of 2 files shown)


paulbalandan:
Please add some tests to this.

hdrodriguez:
Added test

MGatner:
This also needs to be pointed at `4.2` branch.

lonnieezell:
> @lonnieezell it looks like you pointed this to 4.2 then back?

@MGatner yes, was hoping I could just change the branch and merge, but it needs a rebase as dozens of other commits showed up when pointed to that branch.

michalsn (changes requested):
Sorry, but I personally can't give the green light. We've already had at least one attempt on this topic (#4955) and it was "blocked" simply because this type of functionality should be implemented in all database drivers that can support it.

MGatner:
Is there some special reason SQLSRV needs this and other drivers do not? I defer to @michalsn on these matters, so unless there is some counter-justification we will close this.

qury:
I guess this is just for convenience to be able to create cross-schema queries using the query builder instead of using raw SQL.
Other than this i do not see any reason for this.

michalsn:
@MGatner This is just a feature that makes changing schema more convenient (dynamic). We always can change the schema, but we need to use more code. 

This can probably be useful in some cases, but again - this must be implemented for all drivers that support schemas: SQLSRV and Postgre. The code in the database layer is not perfect and I don't want to create more mess. At some point, we will have to think about rewriting the code and that is one more reason why we should keep the features we offer in some "shape".

hdrodriguez:
I honestly do not understand why you do not accept the request, I am sure that this can help thousands of people to solve their problems when working with the SQLSRV driver, I do not know if you have seen that the getFullName method only exists in the SQLSRV builder it does not exist an interface that commits the BaseBuilder to implement the getFullName method, I understand that there must be an order to do things, but many of us take the time to contribute and help improve the framework, as it was proposed in #4955

michalsn:
> I honestly do not understand why you do not accept the request, I am sure that this can help thousands of people to solve their problems when working with the SQLSRV driver, 

I don't know if there are thousands or just ten people waiting for this. It doesn't really matter. The point is that this new functionality should be implemented for all drivers that may support it. Nothing less, nothing more.

> I do not know if you have seen that the getFullName method only exists in the SQLSRV builder it does not exist an interface that commits the BaseBuilder to implement the getFullName method

Yes, I am aware of that. However, this is not a reason to not implement this functionality for other drivers.

> I understand that there must be an order to do things, but many of us take the time to contribute and help improve the framework

We always appreciate every developer who takes his time to contribute to the framework. However it doesn't mean we will accept every pull request. I say it from the beginning, in other PR too - this functionality has to be supported for all database drivers that can handle it.

paulbalandan:
Thanks for the PR! Since there has been no activity since the requested changes, I believe there's no more interest here. Closing.

PR #4998, opened 2021-08-12
Fix lang() function is overriding locale

pjsde:
After using the `lang()` function with the `$locale` parameter filled in, it overrides the `Language` locale, which shouldn't happen, because when using the `lang()` function with the `$locale` parameter it should be just for that call.

With this change, this override is corrected and the locale that was previously selected is restored.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -729,8 +729,21 @@ if (! function_exists('lang')) {
      */
     function lang(string $line, array $args = [], ?string $locale = null)
     {
-        return Services::language($locale)
-            ->getLine($line, $args);
+        if($locale)
+        {
+            //Save previous locale
+            $oldLocale = Services::language()->getLocale();
+        }
+
+        $line = Services::language($locale)
+                        ->getLine($line, $args);
+
+        if($locale && $oldLocale != Services::language()->getLocale())
+        {
+            //Reset to previous locale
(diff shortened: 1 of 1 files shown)


MGatner (changes requested):
If Language really should not be affected by the locale from each call then the helper function should stop using the shared service:
```php
Services::language($locale, false)
```
That would ensure a new instance of the Language class every time in the correct locale. 

If you want to proceed with this change this PR will need to be pointed to the `4.2` branch and add a test that demonstrates this is working as intended.

pjsde:
It could be an solution, but in terms off performance it will be very bad because we will always creating a new instance of the Language class, loading again the same language file. 
Imagine if your view use 100 langs, it will be 100 new Language class doing tha same thing only because one of them must be in a different locale.
What do you think?

MGatner:
I'm not sure, I have not inspected the whole class so it sounds like you know better than me! If you are going with the shared instance I would prefer you do a single service call and then use `setLocale()`, just so the instance interactions are more clear.

pjsde:
Ok. I agree with your comments. I will close this PR and make changes in the `4.2` branch

PR #5001, opened 2021-08-14
Fix lang() function overriding default locale in Language class

pjsde:
After using the `lang()` function with the `$locale` parameter filled in, it overrides the default locale in the Language class, which shouldn't happen, because when using the `lang()` function with the `$locale` parameter it should be just for that call.

With this change, this override is corrected and the locale that was previously selected is restored.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -763,8 +763,25 @@ if (! function_exists('lang')) {
      */
     function lang(string $line, array $args = [], string $locale = null)
     {
-        return Services::language($locale)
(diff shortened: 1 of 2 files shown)


MGatner (approved):
Looks good! Thanks for changing that. There are a few style issues but this whole branch needs to be run through CS Fixer once it is ready so not a big deal - just mind it for the future.

PR #5004, opened 2021-08-17
Make Cookie compatible with ArrayAccess

paulbalandan:
**Description**
Related to #4883 
> During inheritance of ArrayAccess: Uncaught Return type of CodeIgniter\Cookie\Cookie::offsetExists($offset) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Cookie/Cookie.php
+++ b/system/Cookie/Cookie.php
@@ -17,6 +17,7 @@ use Config\Cookie as CookieConfig;
 use DateTimeInterface;
 use InvalidArgumentException;
 use LogicException;
+use ReturnTypeWillChange;
 
 /**
  * A `Cookie` class represents an immutable HTTP cookie value object.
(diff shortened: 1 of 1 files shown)


system/Cookie/Cookie.php
```diff
@@ -17,6 +17,7 @@
 use DateTimeInterface;
 use InvalidArgumentException;
 use LogicException;
+use ReturnTypeWillChange;
```
MGatner, line 20: Can we really `use` a non-existent entity (on some PHP versions)?

MGatner (approved):
I assume these are all LSP-kosher.

system/Cookie/Cookie.php
paulbalandan: Yes. The class resolution will occur on the actual usage, like `new X`

michalsn (approved):

MGatner:
My hunch is that this should go to `4.2` but that's not a strong opinion.

paulbalandan:
My take is this is not really a bug so it's fine to be in `develop` and the parent PR is in develop too.

PR #5005, opened 2021-08-17
Replace deprecated FILTER_SANITIZE_STRING

paulbalandan:
Related to #4883 

**Checklist:**
- [x] Securely signed commits


MGatner:
Can we get a Rector rule for this @samsonasik ?

MGatner (approved):

samsonasik:
@MGatner I may can create custom rector rule for that.

MGatner:
@samsonasik don't worry about it if there isn't something. Looks like this was very isolated, probably unlikely anyone uses this accidentally in the future.

PR #5007, opened 2021-08-17
[Rector] Apply Rector dead code set list

samsonasik:
Rector dead code set list `SetList::DEAD_CODE` is collection of rector rules to remove dead code at 

https://github.com/rectorphp/rector-src/blob/c3c1411d107534e9d5c103982c9e5442213cc800/config/set/dead-code.php

This PR apply rector dead code set list with use latest rector 0.11.48, with skip some edge cases on purpose or need wait for next release for the false positive check functionality.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -125,7 +125,9 @@ class Security implements SecurityInterface
      */
     public function __construct(App $config)
     {
-        /** @var SecurityConfig */
+        /**
+         * @var SecurityConfig $security
+         */
         $security = config('Security');
 
         // Store CSRF-related configurations
@@ -134,7 +136,9 @@ class Security implements SecurityInterface
         $this->regenerate = $security->regenerate ?? $config->CSRFRegenerate ?? $this->regenerate;
         $rawCookieName    = $security->cookieName ?? $config->CSRFCookieName ?? $this->cookieName;
 
-        /** @var CookieConfig */
+        /**
+         * @var CookieConfig $cookie
+         */
         $cookie = config('Cookie');
 
         $cookiePrefix     = $cookie->prefix ?? $config->cookiePrefix;
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -125,9 +125,7 @@ class Security implements SecurityInterface
      */
     public function __construct(App $config)
     {
-        /**
-         * @var SecurityConfig $security
-         */
+        /** @var SecurityConfig $security */
         $security = config('Security');
 
         // Store CSRF-related configurations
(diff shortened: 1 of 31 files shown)


system/Security/Security.php
```diff
@@ -125,7 +125,9 @@ class Security implements SecurityInterface
      */
     public function __construct(App $config)
     {
-        /** @var SecurityConfig */
+        /**
+         * @var SecurityConfig $security
+         */
```
MGatner, line 130: How come this was forced to multi-line when the same thing in **Events.php** reverted to a single line?

system/Security/Security.php
```diff
@@ -134,7 +136,9 @@ public function __construct(App $config)
         $this->regenerate = $security->regenerate ?? $config->CSRFRegenerate ?? $this->regenerate;
         $rawCookieName    = $security->cookieName ?? $config->CSRFCookieName ?? $this->cookieName;
 
-        /** @var CookieConfig */
+        /**
+         * @var CookieConfig $cookie
+         */
         $cookie = config('Cookie');
```
MGatner, line 138: Ditto.

system/Session/Session.php
```diff
@@ -182,7 +182,9 @@ public function __construct(SessionHandlerInterface $driver, App $config)
         $this->cookieSecure   = $config->cookieSecure ?? $this->cookieSecure;
         $this->cookieSameSite = $config->cookieSameSite ?? $this->cookieSameSite;
 
-        /** @var CookieConfig */
+        /**
+         * @var CookieConfig $cookie
+         */
```
MGatner, line 187: Ditto.

tests/system/Session/SessionTest.php
```diff
@@ -180,9 +180,9 @@ public function testGetReturnsItemValueisZero()
         $session = $this->getInstance();
         $session->start();
 
-        $session->set('foo', (int) 0);
```
MGatner, line 183: That's just funny. 🤗

MGatner (approved):
Looks good! Mostly useless casts and incomplete or unnecessary docblocks. I'm not sure why some inline docblocks are multi-line and others are single but I would like to see that standardized.

MGatner (changes requested):
I swear I hit "Request Changes". GitHub Mobile must be trolling me.

system/Security/Security.php
samsonasik: this change is my manual change as it remove `@var` due to no variable definition after `@var` type. Updated to single line.

system/Security/Security.php
samsonasik: fixed.

system/Session/Session.php
samsonasik: fixed.

MGatner (approved):

samsonasik:
@MGatner thank you for the review.

PR #5008, opened 2021-08-17
Cache Config

MGatner:
**Description**
Adds a new config value for reserved characters so developers may set their own restrictions.

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/Cache.php
+++ b/app/Config/Cache.php
@@ -2,6 +2,7 @@
 
 namespace Config;
 
+use CodeIgniter\Cache\Handlers\BaseHandler;
 use CodeIgniter\Cache\Handlers\DummyHandler;
 use CodeIgniter\Cache\Handlers\FileHandler;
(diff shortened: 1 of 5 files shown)


MGatner:
Effectively supersedes #4765

michalsn (approved):

app/Config/Cache.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Config;
 
+use CodeIgniter\Cache\Handlers\BaseHandler;
```
paulbalandan, line 5: Why is this added? I don't remember it being used.

MGatner: Oops! I was using it to reference the original value but then I used the string directly after I deprecated the constant. Good catch.

MGatner, line 5:
```suggestion
```

paulbalandan (approved):

PR #5012, opened 2021-08-19
Make the session handlers all compatible with SessionHandlerInterface

paulbalandan:
**Description**
Related to #4883 

Reference for the updated signatures: https://www.php.net/manual/en/class.sessionhandlerinterface.php
I've refactored and cleaned the code a bit as I see fit.

My only concern is for the `gc` method. Our current implementation is using the pre-7.1 behavior where it returns `true` on success. Starting PHP 7.1, on success it should return the number of garbage collected sessions. My initial solution is to turn `true` to `1` for handlers where the number can't easily be determined while returning the true number on others (like the file handler).

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Session/Handlers/ArrayHandler.php
+++ b/system/Session/Handlers/ArrayHandler.php
@@ -11,7 +11,7 @@
 
 namespace CodeIgniter\Session\Handlers;
 
-use Exception;
+use ReturnTypeWillChange;
 
 /**
  * Session handler using static array for storage.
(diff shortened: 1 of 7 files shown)


system/Session/Handlers/ArrayHandler.php
```diff
@@ -22,78 +22,70 @@ class ArrayHandler extends BaseHandler
...
-     * Deletes expired sessions
+     * @param int $max_lifetime Sessions that have not updated
+     *                          for the last max_lifetime seconds will be removed.
      *
-     * @param int $maxlifetime Maximum lifetime of sessions
+     * @return false|int Returns the number of deleted sessions on success, or false on failure.
      */
-    public function gc($maxlifetime): bool
+    #[ReturnTypeWillChange]
+    public function gc($max_lifetime)
```
michalsn, line 87: Any reason why not to use camelCase?

paulbalandan: I based it here https://www.php.net/manual/en/sessionhandlerinterface.gc.php

I'm not sure though if this will have a bearing on 8.1.

MGatner: > And finally, named arguments will deal in a pragmatic way with name changes during inheritance.

From https://stitcher.io/blog/php-8-named-arguments. There is a good example there but basically: you can change the names in child classes, but names arguments will always have to match the actual class's argument signature. 

I think matching this to its parent signature is a good idea.

MGatner (approved):

system/Session/Handlers/ArrayHandler.php
michalsn: Ok, thank you both. 
Sorry, apparently I'm not up to date with the all-new features of the language 😓

michalsn (approved):

system/Session/Handlers/ArrayHandler.php
MGatner: All good! It can be a side-effect of maintaining all supported versions: you're always working with the oldest feature set. I had a moment looking at the PHP version chart (https://www.php.net/supported-versions.php) where I was like "we should be on active support, not security support!" But in reality I think dropping 7.3 this year will already be a big enough change. Next year when we are `>= 8.0` the real fun will begin 🤗

PR #5038, opened 2021-08-29
Switch to official coding standard

paulbalandan:
**Checklist:**
- [x] Securely signed commits


sfadschm:
Awesome to finally see this, thanks for the crazy work!

Will the `SpaceAfterCommentStartFixer` be transferred, too, or will it stay with the framework?

paulbalandan:
It might stay for a while, while I'm working on a generic way to easily support custom fixers.

MGatner (approved):

MGatner:
So exciting!

paulbalandan:
I'm going to defer merging this, to catch up with the new fixers and options introduced in php cs fixer v3.1

sfadschm:
That will mean fixing the docs again, doesn't it?
I think I read sometime ago that `rst` allows code blocks to be `include`d from php files. Could this be an option to pull out the examples to a subfolder of php files so the cs-fixer can simply be run on them?

samsonasik (approved):

paulbalandan:
> That will mean fixing the docs again, doesn't it?
> I think I read sometime ago that `rst` allows code blocks to be `include`d from php files. Could this be an option to pull out the examples to a subfolder of php files so the cs-fixer can simply be run on them?

I don't think the docs will be changed as new fixers don't affect them. I'm intrigued by this `rst` capability. Can you share some links to docs for that?

sfadschm:
I think this is either the `include` directive or `raw` with `file`:
https://docutils.sourceforge.io/docs/ref/rst/directives.html#include
https://docutils.sourceforge.io/docs/ref/rst/directives.html#raw-data-pass-through

My best guess would be somthing like:

    .. include:: example.php
       :code: php

But I would have to try that 😄

paulbalandan:
Requesting for re-review

- Bump to php-cs-fixer v3.1, coding-standard v1.1, cs-config v3.3
- Used more robust SpaceAfterCommentStartFixer
- Used FixerGenerator for easy registration of custom fixers
- Removed code smell from commented out code
- Removed unneeded comments

MGatner (approved):

.no-header.php-cs-fixer.dist.php
```diff
@@ -11,29 +11,29 @@
...
 
 $finder = Finder::create()
     ->files()
     ->in([
         __DIR__ . '/admin',
         __DIR__ . '/app',
         __DIR__ . '/public',
-    ]);
+    ])
+    ->notName('#Logger.php$#');
```
samsonasik, line 27: is `.` need to be escaped? 

```suggestion
    ->notName('#Logger\.php$#');
```

PR #5045, opened 2021-08-31
Bump to php-cs-fixer v3.1

paulbalandan:
**Description**
New fixers applied:
- `declare_parentheses`
- `empty_loop_body`
- removed deprecated `class_keyword_remove`
- `types_spaces`
- new option for `class_definition`

Only fixer with effect is the new option of `class_definition` adding a space before the parenthesis of anonymous class instantiation to line it with PSR12 requirement that anonymous classes should follow and behave like closures.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/tests/system/API/ResponseTraitTest.php
+++ b/tests/system/API/ResponseTraitTest.php
@@ -82,7 +82,7 @@ final class ResponseTraitTest extends CIUnitTestCase
         }
 
         // Create the controller class finally.
-        $controller = new class($this->request, $this->response, $this->formatter) {
+        $controller = new class ($this->request, $this->response, $this->formatter) {
             use ResponseTrait;
 
             protected $request;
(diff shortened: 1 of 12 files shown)


sfadschm:
I may be blind, but should the `composer.json` not be updated with `3.1`?

MGatner (approved):
I would not have guessed `class_definition`, rather the opposite since I always thought of the "class" as a stand-in for the actual class name, so the parentheses were adjoining for construction. I like the consistency with Closure you pointed out though.

MGatner:
@sfadschm You're not wrong, it's a bit confusing. This will ultimate land at https://github.com/CodeIgniter/coding-standard which will be applied here, but the easiest way to see the effects of each rule (especially while we're getting set up) is to apply them to the framework directly. I'm sure Paul will bump the Composer version at the source repo when it is ready.

sfadschm:
Oh right. Just thought the fixer might complain in the checks, when its on 3.0 but some rules are from 3.1.
This does not seem to be the case, so all good :-)

PR #5055, opened 2021-09-04
Fix entity name generation when bundled in model

### LINKED ISSUE #5050: make:model generate wrong entity
situation 1:
```
php spark make:model MyTable --suffix Model --return entity
```

result:
Models/MyTableModel.php generated. (correct)
the returnType of MyTableModel.php is 'App\Entities\MyTableEntity' (correct)
Entities/MytableEntity.php generated. ( Entities/MyTableEntity.php expected )

situation 2:
```
php spark make:model SubFolder/MyTable --suffix Model --return entity
```

result:
Models/SubFolder/MyTableModel.php generated. (correct)
the returnType of MyTableModel.php is 'App\Entities\SubFolder\MyTableEntity' (correct)
Entities/MytableEntity.php generated. ( Entities/SubFolder/MyTableEntity.php expected )

**CodeIgniter 4 version**
4.1.3

paulbalandan:
**Description**
Fixes #5050 

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Generators/Views/model.tpl.php
+++ b/system/Commands/Generators/Views/model.tpl.php
@@ -6,22 +6,22 @@ use CodeIgniter\Model;
 
 class {class} extends Model
 {
-    protected $DBGroup              = '{DBGroup}';
-    protected $table                = '{table}';
-    protected $primaryKey           = 'id';
-    protected $useAutoIncrement     = true;
-    protected $insertID             = 0;
-    protected $returnType           = '{return}';
-    protected $useSoftDeletes       = false;
-    protected $protectFields        = true;
-    protected $allowedFields        = [];
+    protected $DBGroup          = '{dbGroup}';
+    protected $table            = '{table}';
+    protected $primaryKey       = 'id';
+    protected $useAutoIncrement = true;
+    protected $insertID         = 0;
+    protected $returnType       = {return};
+    protected $useSoftDeletes   = false;
+    protected $protectFields    = true;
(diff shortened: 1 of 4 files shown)


system/Commands/Generators/Views/model.tpl.php
```diff
@@ -6,22 +6,22 @@
...
-    protected $DBGroup              = '{DBGroup}';
-    protected $table                = '{table}';
-    protected $primaryKey           = 'id';
-    protected $useAutoIncrement     = true;
-    protected $insertID             = 0;
-    protected $returnType           = '{return}';
-    protected $useSoftDeletes       = false;
-    protected $protectFields        = true;
-    protected $allowedFields        = [];
+    protected $DBGroup          = '{dbGroup}';
```
MGatner, line 9: This could theoretically be a breaking change. Maybe we should add a note somewhere about template files that they are "internal"?

MGatner (changes requested):

system/Commands/Generators/Views/model.tpl.php
paulbalandan: I have a doc comment under `app/Config/Generators.php` although on reading it is only implicit that the templates are internal. Can that be enough or should I add there a more explicit note?

MGatner: No that sounds good! Sorry I missed it.

MGatner (approved):

PR #5060, opened 2021-09-07
Merge branch '4.2'

paulbalandan:
**Description**
- This was rebased against latest `develop`
- For every merge conflict, incoming change is used. If there is doubt, the original PR is consulted. Then run php-cs-fixer to fix CS
- Run a final CS linting to clean violations

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/.github/workflows/test-deptrac.yml
+++ b/.github/workflows/test-deptrac.yml
@@ -1,0 +1,82 @@
+# When a PR is opened or a push is made, perform an
+# architectural inspection on the code using Deptrac.
+name: Deptrac
+
+on:
+  pull_request:
+    branches:
+      - 'develop'
+      - '4.*'
+    paths:
+      - 'app/**'
+      - 'system/**'
+      - 'composer.json'
+      - 'depfile.yaml'
+      - '.github/workflows/test-deptrac.yml'
+  push:
+    branches:
+      - 'develop'
+      - '4.*'
+    paths:
+      - 'app/**'
+      - 'system/**'
+      - 'composer.json'
+      - 'depfile.yaml'
+      - '.github/workflows/test-deptrac.yml'
+
+jobs:
+  build:
+    name: PHP ${{ matrix.php-versions }} Architectural Inspection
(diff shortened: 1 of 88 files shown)


MGatner:
Sorry Paul, I had done most of this last night and should have pushed it up - I didn't know you were going to work on it! We should compare results with this and #5061 because the rebase was messy and I probably made a few mistakes.

paulbalandan:
This one is fairly ready. I just don't know how to fix PHPCPD.

I don't know why the total commits in yours differ from mine. Maybe because empty commits during rebase got discarded? This PR is just missing 2 commits from yours (which I believe are just empty commits)

MGatner:
Yes I think that's the case - there were a couple of commits that I manually adjusted to the proper styling only to find that a layer commit was already going to do that, so those probably got removed. I will check out PHP CPD when I'm on desktop but let's plan to go with this one.

MGatner:
Not sure what was happening there, those lines weren't duplicates at all! Since we were already excluding part of SQLSRV I added Forge as well, not a big deal.

MGatner (approved):

sfadschm:
I think the marked passages are quite similar, except for the use of `schema` for SQLSRV. Does CPD have a similarity threshold so that this maybe just happened to be an edge case?

paulbalandan:
I also noticed that in a way that schema is the only difference so I was looking on how to extract the common functionality. Im not yet familiar with the options but I think I saw it can.

MGatner:
Yes you can adjust the threshold with CLI parameters. If someone wants to dig into this please do but it also isn't a big deal to exclude some files.

sfadschm:
> Yes you can adjust the threshold with CLI parameters. If someone wants to dig into this please do but it also isn't a big deal to exclude some files.

Agreed. However, as Paul said, the duplicate parts are extractable in principle. The general question would be, how much do we want to modularize for the sake of not duplicating code.

PR #5073, opened 2021-09-11
Ignore trailing slash in webpage cache, Add purge method

najdanovicivan:
**Description**
This one makes the webpage cache ignore the trailing slash. It also added a method for purging single page from cache 

splitted from #4974 

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -639,6 +639,25 @@ class CodeIgniter
     }
 
     /**
+     * Purges the cached page
+     *
+     * @param string     $uri    Uri
+     * @param Cache|null $config Config
+     */
+    public function purgeCachedPage(string $uri, ?Cache $config = null): bool
+    {
+        $uri = new URI($uri);
+
+        if (($config ?? config('Cache'))->cacheQueryString) {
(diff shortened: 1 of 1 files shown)


najdanovicivan:
@MGatner you said we need tests for this one but I'm not sure how to start writing those. I haven't seen a single existing test for the webpage cache so some hints will be a lot of help

paulbalandan:
Looking at the new method, I would suggest the ff for the test:

1. Simulate creating a cached dummy page identified by a dummy URI. The cached page will be one with a query string and one without.
2. There'll be 4 tests: 
- cacheQueryString is true against the cached page with query string
- cacheQueryString is true against cached page without query string
- false against with
- false against without
3. In each of those tests you will just assert thru assertTrue or assertFalse the result of the cache delete when the new method is called

paulbalandan:

MGatner:
This was actually a (potential) bugfix and a new feature rolled into one. @najdanovicivan Any chance you could resubmit the bugfix by itself (if you still believe it is a bug)?

PR #5074, opened 2021-09-11
[Pager] Add setTotal, getRawCurrentPage methods

najdanovicivan:
**Description**

Added function setTotal(int $total, string $group = 'default'): self
This one is intended to be used for NoSQL like databases where the total item count can be returned within the data query so that total can be set after query execution.

Added public function getRawCurrentPage(string $group = 'default'): int which is basically taken out from calculateCurrentPage function.  This one makes it possible to get the actual page requested from the URL so that 404 page can be rendered if the requested page is higher that total available pages.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Pager/Pager.php
+++ b/system/Pager/Pager.php
@@ -181,6 +181,18 @@ class Pager implements PagerInterface
     }
 
     /**
+     * Set the total number of items in data store.
+     */
+    public function setTotal(int $total, string $group = 'default'): self
+    {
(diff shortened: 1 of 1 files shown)


lonnieezell:
@najdanovicivan are you going to be able to add some tests for this? We can only accept PR's with tests.

paulbalandan:

PR #5021, opened 2021-08-24
Session Config

mostafakhudair:
Split session configuration from `Config::App` to a new config class `Config::Session`

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPdocs
- [x] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #5088, opened 2021-09-14
Session config

mostafakhudair:
Split session configuration from `Config\App` to a new config class `Config\Session`

Alternative #5021 

Checklist:
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Session.php
+++ b/app/Config/Session.php
@@ -81,14 +81,14 @@ class Session extends BaseConfig
 
     /**
      * --------------------------------------------------------------------------
-     * Session Time to Update
+     * Session Time to Live
      * --------------------------------------------------------------------------
      *
      * How many seconds between CI regenerating the session ID.
      *
      * @var int
      */
-    public $cooldown = 300;
+    public $ttl = 300;
 
     /**
      * --------------------------------------------------------------------------
--- a/app/Config/Session.php
+++ b/app/Config/Session.php
@@ -1,0 +1,105 @@
+<?php
+
+namespace Config;
+
+use CodeIgniter\Config\BaseConfig;
+use CodeIgniter\Session\Handlers\FileHandler;
+
+class Session extends BaseConfig
+{
+    /**
+     * --------------------------------------------------------------------------
+     * Session Handler
+     * --------------------------------------------------------------------------
+     *
+     * The session storage handler to use:
+     * - `CodeIgniter\Session\Handlers\DatabaseHandler`
+     * - `CodeIgniter\Session\Handlers\FileHandler`
+     * - `CodeIgniter\Session\Handlers\MemcachedHandler`
+     * - `CodeIgniter\Session\Handlers\RedisHandler`
+     *
+     * @var string
+     */
+    public $handler = FileHandler::class;
+
+    /**
+     * --------------------------------------------------------------------------
+     * Session Name
+     * --------------------------------------------------------------------------
+     *
+     * The name of the session which is used as cookie name.
+     * It should only contain alphanumeric characters.
+     *
+     * @var string
+     */
+    public $name = 'ci_session';
+
+    /**
+     * --------------------------------------------------------------------------
+     * Session Life Time
+     * --------------------------------------------------------------------------
+     *
+     * The number of SECONDS you want the session to last.
+     * Setting to 0 (zero) means expire when the browser is closed.
+     *
+     * @var int
+     */
+    public $lifetime = 7200;
+
+    /**
+     * --------------------------------------------------------------------------
+     * Session Save Path
+     * --------------------------------------------------------------------------
+     *
+     * The location to save sessions to and is driver dependent.
+     *
+     * For the 'files' driver, it's a path to a writable directory.
+     * WARNING: Only absolute paths are supported!
+     *
+     * For the 'database' driver, it's a table name.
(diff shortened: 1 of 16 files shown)


MGatner (changes requested):
A few changes dealing with compatibility. If there were any lingering issues from the other PR please make sure the reviewers have had a chance to check them here as well. Also, thanks for this work! Excited for this change.

system/Session/Handlers/BaseHandler.php
```diff
@@ -102,13 +114,18 @@ abstract class BaseHandler implements SessionHandlerInterface
 
     public function __construct(AppConfig $config, string $ipAddress)
     {
+        /** @var SessionConfig $session */
+        $session = config('Session');
```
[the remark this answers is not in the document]
kenjis: @MGatner Why is it a breaking change?
Without `app/Config/Session.php` or `system/Config/Session.php`, `$session` will be `null`.
And `$this->cookieName` will be `$config->sessionCookieName` or `$this->cookieName`.
Am I something wrong?

system/Session/Session.php
```diff
@@ -36,59 +38,59 @@ class Session implements SessionInterface
...
-    protected $sessionMatchIP = false;
+    protected $matchIP = false;
 
     /**
      * How many seconds between CI regenerating the session ID.
      *
      * @var int
      */
-    protected $sessionTimeToUpdate = 300;
+    protected $cooldown = 300;
```
kenjis, line 93: I'm not sure `cooldown` is proper naming. 
At least I don't understand cooldown means session regenerating time.

app/Config/Session.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * --------------------------------------------------------------------------
+     * Session Time to Update
+     * --------------------------------------------------------------------------
+     *
+     * How many seconds between CI regenerating the session ID.
+     *
+     * @var int
+     */
+    public $cooldown = 300;
```
kenjis, line 91: I'm not sure `cooldown` is proper naming. 
At least I don't understand cooldown means session id regenerating time.

app/Config/Session.php
```diff
@@ -0,0 +1,105 @@
...
+     * Session Regenerate
+     * --------------------------------------------------------------------------
+     *
+     * Whether to destroy session data associated with the old session ID
+     * when auto-regenerating the session ID. When set to FALSE, the data
+     * will be later deleted by the garbage collector.
+     *
+     * @var bool
+     */
+    public $regenerate = false;
```
kenjis, line 104: `regenerate` is not proper naming.
`regenerate` implies whether to regenerate or not.
But it actually means whether to delete old session.

`$deleteOldSession` is better?

system/Session/Session.php
```diff
@@ -36,59 +38,59 @@ class Session implements SessionInterface
     protected $driver;
 
     /**
-     * The storage driver to use: files, database, redis, memcached
+     * The session storage driver to use.
      *
      * @var string
      */
-    protected $sessionDriverName;
+    protected $handler = FileHandler::class;
```
kenjis, line 45: @MGatner Isn't renaming protected property is a breaking change?

system/Session/Handlers/BaseHandler.php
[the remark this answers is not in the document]
MGatner: No, you're correct! I didn't realize bull coalescence would allow property access on a non-object, but apparently it does: https://3v4l.org/pmAoE

I still think this would be cleaner the other way, with presence of the Session config as the determinant instead of checking each property on a potentially-nonexistent object. But it isn't a breaking change, which was the important part.

app/Config/Session.php
MGatner: If we want to be consistent with Cache this would be "ttl" (time to live).

system/Session/Session.php
MGatner: I believe the he kept the original, just moved it lower and deprecated it. I'm on mobile so it's hard to see but I think I saw that during my review.

system/Session/Handlers/BaseHandler.php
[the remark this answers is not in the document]
kenjis: Okay. I agree with @MGatner .
`if ($session = config('Session')) { ...` is cleaner.

paulbalandan:
This has been at draft status for couple of months now. Where are we on this? Or should this be closed?

mostafakhudair:
I'll try to finish it tomorrow

MGatner:
I'm still keen on having this.

system/Session/Handlers/BaseHandler.php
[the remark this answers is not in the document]
mostafakhudair: The same approach was used to create the `Config\Cookie` and `Config\Security` classes

app/Config/Session.php
mostafakhudair: `deleteOldSession` is more accurate, but I am trying to find a short property name, if there is no more suggestion, yours will be.

app/Config/Session.php
mostafakhudair: i'll go with `$ttl`

PR #5094, opened 2021-09-15
Make log_message() compatible with PSR-3

kenjis:
**Description**

> Now, any call that is done through the log_message() function will use your library instead.
https://codeigniter4.github.io/CodeIgniter4/general/logging.html#using-third-party-loggers

But the method `log()` does not exist in PSR-3, so `log_message()` does not work with PSR-3 logger.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -767,7 +767,7 @@ if (! function_exists('log_message')) {
 
         // @codeCoverageIgnoreStart
         return Services::logger(true)
-            ->log($level, $message, $context);
+            ->{$level}($message, $context);
         // @codeCoverageIgnoreEnd
     }
 }


paulbalandan (commented):
I'm not sure what you mean, but `log` is present in PSR-3 `LoggerInterface`.
```php
    /**
     * Logs with an arbitrary level.
     *
     * @param mixed $level
     * @param string $message
     * @param array $context
     * @return void
     */
    public function log($level, $message, array $context = array());
```

kenjis:
@paulbalandan 
You are correct.
I overlooked that method.

PR #5100, opened 2021-09-16
Fix debugbar styles printing

### LINKED ISSUE #5097: Possible debug toolbar css problem
**Direction**

**Describe the bug**
When the debug toolbar is enabled, the style element at the end of the toolbarContainer div is missing the style declaration.  I think this may be from bad data getting in the $styles object.  

````html
<style type="text/css">
    .debu
</style>
````
I think this is around like 270 in the toolbar.tpl.php where it's looping through the names in the $styles, but I'm not sure.

**CodeIgniter 4 version**
4.0.5

**Affected module(s)**
The debug toolbar

Expect a complete style element. with a reference to an existing class.  There may be an element with a debu class, but I didn't see it immediately.  Something like this:
````html
<style type="text/css">
    .debug {
        font-size: 18pt;
        color: blue;
    }
</style>
````
Or whatever style is appropriate.

### PR DESCRIPTION
**Pull request by paulbalandan**
**Description**
Fixes #5097 

The truncated styles would actually not pose a problem because in `toolbarloader.js.php` the styles and scripts from `toolbarContainer` is dynamically copied to the style and script in the `head`. After dynamically copying, those should be supposedly removed from `toolbarContainer` thru `responseText.substr`. This works well until the last copying of the last style where the removal was "misdirected".

To check that those styles would be redundant if we let it stay is to see that the timeline bars still got styled (in orange bars of varying lengths). Right-clicking on 'Inspect element' you would be that those bars have the class `.debug-bar-timeline-**` styles with the definition at the bottom.

If any of @rramsey or @kenjis can test this ?

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -406,6 +406,8 @@ class Toolbar
 
     /**
      * Inject debug toolbar into the response.
+     *
+     * @codeCoverageIgnore
      */
     public function respond()
     {
@@ -413,20 +415,20 @@ class Toolbar
             return;
         }
 
-        // @codeCoverageIgnoreStart
         $request = Services::request();
 
         // If the request contains '?debugbar then we're
         // simply returning the loading script
         if ($request->getGet('debugbar') !== null) {
-            // Let the browser know that we are sending javascript
             header('Content-Type: application/javascript');
 
             ob_start();
             include $this->config->viewsPath . 'toolbarloader.js.php';
             $output = ob_get_clean();
+            $output = substr($output, 8, -10); // trim the script tags
+            echo $output;
 
-            exit($output);
+            exit;
         }
 
         // Otherwise, if it includes ?debugbar_time, then
@@ -435,29 +437,24 @@ class Toolbar
             helper('security');
 
             // Negotiate the content-type to format the output
-            $format = $request->negotiate('media', [
-                'text/html',
-                'application/json',
-                'application/xml',
-            ]);
+            $format = $request->negotiate('media', ['text/html', 'application/json', 'application/xml']);
             $format = explode('/', $format)[1];
 
-            $file     = sanitize_filename('debugbar_' . $request->getGet('debugbar_time'));
-            $filename = WRITEPATH . 'debugbar/' . $file . '.json';
+            $filename = sanitize_filename('debugbar_' . $request->getGet('debugbar_time'));
+            $filename = WRITEPATH . 'debugbar/' . $filename . '.json';
 
-            // Show the toolbar
             if (is_file($filename)) {
-                $contents = $this->format(file_get_contents($filename), $format);
+                // Show the toolbar if it exists
+                echo $this->format(file_get_contents($filename), $format);
 
-                exit($contents);
+                exit;
             }
 
-            // File was not written or do not exists
+            // Filename not found
             http_response_code(404);
 
-            exit; // Exit here is needed to avoid load the index page
+            exit; // Exit here is needed to avoid loading the index page
         }
-        // @codeCoverageIgnoreEnd
     }
 
     /**
--- a/system/Debug/Toolbar/Views/toolbar.tpl.php
+++ b/system/Debug/Toolbar/Views/toolbar.tpl.php
@@ -265,10 +265,7 @@
 	</div>
 </div>
 <style type="text/css">
-	<?php foreach ($styles as $name => $style) : ?>
-	.<?= $name ?> {
-		<?= $style ?>
-	}
-
-	<?php endforeach ?>
+<?php foreach ($styles as $name => $style): ?>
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: system/Debug/Toolbar/Views/toolbarloader.js.php [hunk 1]
```diff
@@ -9,45 +12,43 @@ function loadDoc(time) {
...
+			start = responseText.indexOf('>', responseText.indexOf('<script')) + 1;
+			end = responseText.indexOf('\<\/script>', start);
+			dynamicScript.innerHTML = responseText.substr(start, end - start);
+			responseText = responseText.substr(end + 9);
+
+			// check for last style block, append contents to dynamic_style, then remove here
+			start = responseText.indexOf('>', responseText.indexOf('<style')) + 1;
+			end = responseText.indexOf('</style>', start);
+			dynamicStyle.innerHTML += responseText.substr(start, end - start);
+			responseText = responseText.substr(0, start - 8);
```
kenjis, line 48: before: `responseText = responseText.substr( 0, PosBeg + 8 );`
after: `responseText = responseText.substr(0, start - 8);`

Why did the `+` change to `-`?

kenjis:
`<style type="text/css">.debu</style>` is gone. It is okay.

Screenshot 2021-09-17 17 43 30

system/Debug/Toolbar/Views/toolbarloader.js.php
paulbalandan: In the `before` the substr will include the `<style>` tag. This change ensures it doesn't

kenjis: I got it. It is where the bug existed.

kenjis (approved):

kenjis:
@paulbalandan 
By the way, your one commit includes more than one change (intent).
In this commit, 
- documentation (comment) removal
- refactoring (coding style, variable name change, etc)
- bug fix

In fact, I couldn't figure out what was a real bug fix at first.
If the commit is separated, we could get the intention of code changes,
and our review would be more easily.

system/Debug/Toolbar.php
```diff
@@ -406,58 +406,53 @@ public function prepare(?RequestInterface $request = null, ?ResponseInterface $r
...
     {
         if (ENVIRONMENT === 'testing') {
             return;
         }
 
-        // @codeCoverageIgnoreStart
         $request = Services::request();
 
-        // If the request contains '?debugbar then we're
-        // simply returning the loading script
```
MGatner, line 420: As someone largely unfamiliar with the Debug Bar I find the inline comments in here helpful. Did they seem too obvious to you? Or other issue?

system/Debug/Toolbar/Views/toolbar.tpl.php
```diff
@@ -265,10 +265,7 @@
 	</div>
 </div>
 <style type="text/css">
-	<?php foreach ($styles as $name => $style) : ?>
-	.<?= $name ?> {
-		<?= $style ?>
-	}
-
-	<?php endforeach ?>
+<?php foreach ($styles as $name => $style) : ?>
```
MGatner, line 268:
```suggestion
<?php foreach ($styles as $name => $style): ?>
```

MGatner (changes requested):

system/Debug/Toolbar.php
paulbalandan: If you will look at the code, the inline comments tell you _what_ it is doing instead of _why_ it is doing it. So, stating the obvious is too much noise here.

kenjis: At first, I thought they were obvious. 

But after I read carefiully, I found that I didn't get the concept `the loading script` and `the entire debugbar` without the comments.

system/Debug/Toolbar.php
```diff
@@ -406,58 +406,53 @@ public function prepare(?RequestInterface $request = null, ?ResponseInterface $r
...
 
-                exit($contents);
+                exit;
             }
 
-            // File was not written or do not exists
+            // Filename not found
             http_response_code(404);
 
-            exit; // Exit here is needed to avoid load the index page
```
kenjis, line 458: The comment is *why*.

system/Debug/Toolbar.php
MGatner: @paulbalandan I appreciate the rubric, I will adopt that for my own comment analysis - but I would also suggest that the *what* can be helpful when the code is a bit obscure. If both kenjis and I found the comments helpful to decipher the content they might be worth keeping.

kenjis:
Okay. No `<style type="text/css">.debu</style>`.
Screenshot 2021-10-01 17 18 07

kenjis (approved):
LGTM

MGatner (approved):

PR #5102, opened 2021-09-17
docs: replace tab with space

kenjis:
**Description**
- replace tab with space

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


paulbalandan (approved):
Can we somehow come up with a bash script that checks for usage of tabs then replace them with spaces?

kenjis:
Simple replacement might break the RST format.

Eventually, we won't need to use tabs in user guide.
After that, it is better to add a script to check there is no tabs.

MGatner (approved):
Thanks!!

kenjis:
@paulbalandan I made a PR to check RST files with pre-commit: #5106

PR #5103, opened 2021-09-17
[Rector] Demo update to PHP 7.4 Syntax

samsonasik:
Demo of use Rector to Php 7.4 Syntax, <strike>skip Typed Attribute due will be BC break for non-private property</strike>. Use TypedPropertyRector with configured on apply on private property only to avoid BC break.

**Checklist:**
- [x] Securely signed commits


app/Config/Constants.php
```diff
@@ -38,9 +38,9 @@
 defined('HOUR')   || define('HOUR', 3600);
 defined('DAY')    || define('DAY', 86400);
 defined('WEEK')   || define('WEEK', 604800);
-defined('MONTH')  || define('MONTH', 2592000);
-defined('YEAR')   || define('YEAR', 31536000);
-defined('DECADE') || define('DECADE', 315360000);
+defined('MONTH')  || define('MONTH', 2_592_000);
+defined('YEAR')   || define('YEAR', 31_536_000);
+defined('DECADE') || define('DECADE', 315_360_000);
```
MGatner, line 43: What is this black magic?

system/Config/Services.php
```diff
@@ -234,9 +234,9 @@ public static function exceptions(
             return static::getSharedInstance('exceptions', $config, $request, $response);
         }
 
-        $config   = $config ?? config('Exceptions');
-        $request  = $request ?? AppServices::request();
-        $response = $response ?? AppServices::response();
+        $config ??= config('Exceptions');
+        $request ??= AppServices::request();
+        $response ??= AppServices::response();
```
MGatner, line 239: Probably a style question for @paulbalandan - are these intentionally exempt from alignment?

MGatner (commented):
This looks great! A couple questions. 

I think we should start a new branch for these changes and anything that will build on them.

app/Config/Constants.php
samsonasik: It generated by 

https://github.com/rectorphp/rector-src/blob/1e3e5dff59b05d497580f52e9a7491a2afa6bd29/rules/Php74/Rector/LNumber/AddLiteralSeparatorToNumberRector.php#L30

We can skip the rule if we want to exclude it

paulbalandan: I'm not aware this was possible with PHP 7.4. Nice 👍

system/Config/Services.php
paulbalandan: By default, our binary operators alignment is set to `single_space` to avoid surprises. We just add some exceptions. This could be added to that list.

app/Config/Constants.php
MGatner: Yes, I like it as well! I didn't know that it was possible was all. Keep!

system/Config/Services.php
MGatner: I think this would make sense to include. I guess the question is how these operators interact with each other, e.g.:
```php
$fruit = 'banana';
$vegetable ??= 'celery';

// versus
$fruit       = 'banana';
$vegetable ??= 'celery';
```

samsonasik:
@MGatner I've updated to use rector/rector:dev-main to apply typed property on private property https://github.com/codeigniter4/CodeIgniter4/pull/5103/commits/49f5dc3a41412ae36f408cb38b3ef53391355948

samsonasik:
It seems there is issue on publish coveralls on php 7.4 github action, even on latest develep:

```bash
Dump submitting json file: /home/runner/work/CodeIgniter4/CodeIgniter4/build/logs/coveralls-upload.json
434
File size: 2,476.05 kB
435
Submitting to https://coveralls.io/api/v1/jobs
436
Client error occurred. status: 405 Not Allowed
437
<html>
438
<head><title>405 Not Allowed</title></head>
439
<body>
440
<center><h1>405 Not Allowed</h1></center>
441
<hr><center>nginx</center>
442
</body>
```

see https://github.com/codeigniter4/CodeIgniter4/runs/3644065245#step:12:434 . 

It seems we can disable that //cc @paulbalandan

system/Config/Services.php
paulbalandan: PHP-CS-Fixer is smart enough to distinguish the two, as it relies on tokens. `=` is `=` while `??=` is `T_COALESCE_EQUAL`

system/Database/SQLSRV/Connection.php
```diff
@@ -229,9 +229,7 @@ protected function _indexData(string $table): array
             $obj->name = $row->index_name;
 
             $_fields     = explode(',', trim($row->index_keys));
-            $obj->fields = array_map(static function ($v) {
-                return trim($v);
-            }, $_fields);
+            $obj->fields = array_map(static fn ($v) => trim($v), $_fields);
```
paulbalandan, line 232: Not related to PHP 7.4 migration but can be simplified further.
```suggestion
            $obj->fields = array_map('trim', $_fields);
```

tests/system/Debug/ExceptionsTest.php
```diff
@@ -25,10 +25,7 @@ final class ExceptionsTest extends CIUnitTestCase
 {
     use ReflectionHelper;
 
-    /**
-     * @var Exceptions
-     */
-    private $exception;
+    private \CodeIgniter\Debug\Exceptions $exception;
```
paulbalandan, line 28:
```suggestion
    private Exceptions $exception;
```

system/Config/AutoloadConfig.php
```diff
@@ -130,6 +130,6 @@ public function __construct()
 
         $this->psr4     = array_merge($this->corePsr4, $this->psr4);
         $this->classmap = array_merge($this->coreClassmap, $this->classmap);
```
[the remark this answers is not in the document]
samsonasik: the rule seems possibly currently can't detect non-numbered key generic type (`array<string, string>`)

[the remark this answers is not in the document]
samsonasik: @paulbalandan array with string key seems cannot be unpack see https://3v4l.org/vabHB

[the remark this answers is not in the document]
paulbalandan: Ok. I was not aware of that.

.github/workflows/test-phpunit.yml
```diff
@@ -139,6 +140,7 @@ jobs:
 
       - if: github.repository_owner == 'codeigniter4' && matrix.php-versions == '7.4'
         name: Run Coveralls
+        continue-on-error: true
```
paulbalandan, line 143: I think this can be reverted now.

rector.php
```diff
@@ -52,6 +53,7 @@
 return static function (ContainerConfigurator $containerConfigurator): void {
     $containerConfigurator->import(SetList::DEAD_CODE);
     $containerConfigurator->import(SetList::PHP_73);
+    $containerConfigurator->import(SetList::PHP_74);
```
paulbalandan, line 56: Do we still need to include SetList::PHP_73 here?

samsonasik: Yes, that means php 7.3 syntax will be included as well.

.github/workflows/test-phpunit.yml
samsonasik: Reverted

rector.php
samsonasik: See example of multiple php setlist so all features from php-x to php-y applied https://github.com/samsonasik/ErrorHeroModule/blob/master/rector.php#L21-L26

paulbalandan (approved):

samsonasik:
I will rework it.

PR #5108, opened 2021-09-20
Multiple route filters

iRedds:
**Description**
https://forum.codeigniter.com/thread-80117-post-390128.html

Several filters can be specified for a route.
Instead of filter aliases, you can use the name of the filter class.

```php
$route->add('foo', 'Home::index', ['filter' => ['filterAlias', \App\Filters\SomeFilter::class]);
```

Also fixed typehint in dockblock in the RouteCollection methods. Missing ``Closure``.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -512,7 +512,7 @@ class RouteCollection implements RouteCollectionInterface
      * Example:
      *      $routes->add('news', 'Posts::index');
      *
-     * @param array|string $to
+     * @param array|string|Closure $to
      */
     public function add(string $from, $to, ?array $options = null): RouteCollectionInterface
     {
@@ -821,7 +821,7 @@ class RouteCollection implements RouteCollectionInterface
      * Example:
      *  $route->match( ['get', 'post'], 'users/(:num)', 'users/$1);
      *
-     * @param array|string $to
+     * @param array|string|Closure $to
      */
     public function match(array $verbs = [], string $from = '', $to = '', ?array $options = null): RouteCollectionInterface
     {
@@ -841,7 +841,7 @@ class RouteCollection implements RouteCollectionInterface
     /**
      * Specifies a route that is only available to GET requests.
      *
-     * @param array|string $to
+     * @param array|string|Closure $to
      */
     public function get(string $from, $to, ?array $options = null): RouteCollectionInterface
     {
@@ -853,7 +853,7 @@ class RouteCollection implements RouteCollectionInterface
     /**
      * Specifies a route that is only available to POST requests.
      *
-     * @param array|string $to
+     * @param array|string|Closure $to
      */
     public function post(string $from, $to, ?array $options = null): RouteCollectionInterface
     {
@@ -865,7 +865,7 @@ class RouteCollection implements RouteCollectionInterface
     /**
      * Specifies a route that is only available to PUT requests.
      *
-     * @param array|string $to
+     * @param array|string|Closure $to
      */
     public function put(string $from, $to, ?array $options = null): RouteCollectionInterface
     {
@@ -877,7 +877,7 @@ class RouteCollection implements RouteCollectionInterface
     /**
      * Specifies a route that is only available to DELETE requests.
      *
-     * @param array|string $to
+     * @param array|string|Closure $to
      */
     public function delete(string $from, $to, ?array $options = null): RouteCollectionInterface
     {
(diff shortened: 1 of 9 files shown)


kenjis:
@iRedds Coding standards checking was failed.
https://github.com/codeigniter4/CodeIgniter4/pull/5108/checks?check_run_id=3648422172

Run `composer cs-fix`, please.

system/Router/RouteCollection.php
```diff
@@ -1040,12 +1040,14 @@ public function isFiltered(string $search, ?string $verb = null): bool
      *    'role:admin,manager'
      *
      * has a filter of "role", with parameters of ['admin', 'manager'].
+     *
+     * @retun string|string[]
      */
-    public function getFilterForRoute(string $search, ?string $verb = null): string
+    public function getFilterForRoute(string $search, ?string $verb = null)
```
kenjis, line 1046: This is a breaking API change. But it seems we must change the return type for multiple filters.

iRedds: What are your suggestions?

kenjis: @MGatner This PR contains public API changes. 
Can we include this on the next release or we need to wait for the next major version (v5)?

If we can change the return type, `array`(`string[]`) is better.

MGatner: We have to stick to semantic versioning, which means we need to leave the signature or save this for version 5. Workarounds could be adding a new method and deprecating this one, or returning an encoded string (like a CSV of classes & aliases).

kenjis: If we want to include this in the next version (not v5), 
it seems returning an encoded string (like a CSV of classes & aliases) is the only choice.

@iRedds What do you think?

iRedds: I think this is counterproductive.
The filter alias already contains ":" and "," to separate the alias and parameters. And adding one more separator will complicate read/write (by human). 

An array of strings is the best solution in my opinion.

kenjis: @iRedds I didn't say about `RouteCollection::add()` usage.
```php
$collection->add('foo', 'TestController::foo', ['filter' => ['filter1', 'filter2:param']]);
```
This is not a breaking change. Because the 3rd param type is `?array`.

But if we change public method signatures, it is breaking change.
In this PR, `getFilterForRoute()` and `enableFilter()` are changed.
So this PR will not be able to be merged until the next major version 5.0.

iRedds: I think maintainers are too conservative about breaking changes.
If the backward compatibility of commonly used functions is broken, that's BC. It's my opinion.

The current policy negatively affects the improvement and popularization of the framework.
And version 5.0 looks like a meme "ten thousand years later" (c) SBSP

I have nothing to say. ggwp

kenjis: I also feel strongly constrained by the inability to change the method signatures.

I also understand that the policy of following semantic versioning has its value.

MGatner:
`getFilterForRoute()` is actually not a part of the `RouteCollectionInterface`. While this has its own issues, it makes it easier for us to work with. The only places this method is used is a negligible `Test` support and in the `Router`:
https://github.com/codeigniter4/CodeIgniter4/blob/984373748ad2514812356c307d70acce6bac4aaf/system/Router/Router.php#L147

Here's my recommendation:
* Add a new method `getFiltersForRoute()` (notice the plural) with the desired signature and behavior
* Deprecated the current `getFilterForRoute()` with a doc block and in the User Guide files
* Change `Router` to use the new method when possible (e.g. `if (method_exists('getFiltersForRoute', $this->collection))`

kenjis:
@MGatner I sent another PR  #5128 with the direction you recommend.

* Change `Router` to use the new method when possible (e.g. `if (method_exists('getFiltersForRoute', $this->collection))`

`method_exists()` can't be used, because if user extends CI4's class, when upgrading CI4, the user class automatically inherits the new method.

I couldn't think of a way to automatically determine when to use a new method.
So I added a new Config class and users define it.

MGatner:
If a user class extends `RouteCollection` then we should be fine. Basically we are finding a way to add a new "optional" feature, and need to make the rest of our code work with or without that feature. 99% of cases developers will be using our native routing classes, no problem; in the 1% (more like .001%) case we need to be sure the framework still works with any `RouteCollectionInterface`.

MGatner:
@kenjis I understand better after looking at your PR: this is a larger issue than just this one method. I will wait and look these over when I'm on desktop and can compare files more easily.

paulbalandan:
Closed in favor of #5128

PR #5117, opened 2021-09-23
Fix query binding with two colons in query

### LINKED ISSUE #5114: Bug: Query binding fails when date_format is set with '%H:%i:%s'
**Describe the bug**
Query builder is binding incorrectly the statement with date_format (eg. %d/%m/%Y %H:%i:%s) - 22/09/2021 10:45:00 and crash.

```php
<?php

$id = 1;
//Works
$sql = " SELECT mytable.id, DATE_FORMAT(mytable.created_at,'%d/%m/%Y %H:%i') AS created_at_uk FROM mytable WHERE mytable.id = ?";
$query = $this->db->query($sql, [$id]);

//DOES NOT work
$sql = " SELECT mytable.id, DATE_FORMAT(mytable.created_at,'%d/%m/%Y %H:%i:%s') AS created_at_uk FROM mytable WHERE mytable.id = ?";
$query = $this->db->query($sql, [$id]);

```
As you can see, the only difference between both SQL is the `:%s`. This should not be binding as the (%s) means seconds on date_format.

**CodeIgniter 4 version**
CI 4.1.3

**Affected module(s)**
Not sure but looks like is Query.php

Should return the query with day/month/year hour:minute:second

**Context**
 - OS: Mac
 - Web server Apache
 - PHP version 8.0

kenjis:
**Description**
Fixes #5114

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -275,12 +275,13 @@ class Query implements QueryInterface
     {
         $sql = $this->finalQueryString;
 
-        $hasNamedBinds = strpos($sql, '?') === false
+        $hasBinds      = strpos($sql, $this->bindMarker) !== false;
+        $hasNamedBinds = ! $hasBinds
             && preg_match('/:((?!=).+):/', $sql) === 1;
 
         if (empty($this->binds)
             || empty($this->bindMarker)
-            || (! $hasNamedBinds && strpos($sql, $this->bindMarker) === false)
+            || (! $hasNamedBinds && ! $hasBinds)
         ) {
             return;
         }
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -277,7 +277,7 @@ class Query implements QueryInterface
 
         $hasBinds      = strpos($sql, $this->bindMarker) !== false;
         $hasNamedBinds = ! $hasBinds
-            && preg_match('/:((?!=).+):/', $sql) === 1;
+            && preg_match('/:(?!=).+:/', $sql) === 1;
 
         if (empty($this->binds)
             || empty($this->bindMarker)
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -269,7 +269,7 @@ class Query implements QueryInterface
     /**
      * Escapes and inserts any binds into the finalQueryString object.
      *
-     * @see https://regex101.com/r/EUEhay/4
+     * @see https://regex101.com/r/EUEhay/5
      */
     protected function compileBinds()
     {
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -275,7 +275,8 @@ class Query implements QueryInterface
     {
         $sql = $this->finalQueryString;
 
-        $hasNamedBinds = preg_match('/:((?!=).+):/', $sql) === 1;
+        $hasNamedBinds = strpos($sql, '?') === false
+            && preg_match('/:((?!=).+):/', $sql) === 1;
 
         if (empty($this->binds)
             || empty($this->bindMarker)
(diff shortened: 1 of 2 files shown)


MGatner (approved):
This looks good to me but Database is my "blind spot" so I'd like @paulbalandan or @samsonasik to take a look.

system/Database/Query.php
```diff
@@ -275,11 +275,13 @@ protected function compileBinds()
     {
         $sql = $this->finalQueryString;
 
-        $hasNamedBinds = preg_match('/:((?!=).+):/', $sql) === 1;
+        $hasBinds      = strpos($sql, $this->bindMarker) !== false;
+        $hasNamedBinds = ! $hasBinds
+            && preg_match('/:(?!=).+:/', $sql) === 1;
```
samsonasik, line 280: Could you add `@see https://regex101.com` link for the regex update, it  can be reference in the future

system/Database/Query.php
```diff
@@ -275,11 +275,13 @@ protected function compileBinds()
...
 
         if (empty($this->binds)
             || empty($this->bindMarker)
-            || (! $hasNamedBinds && strpos($sql, $this->bindMarker) === false)
+            || (! $hasNamedBinds && ! $hasBinds)
```
paulbalandan, line 284: $hasBinds is essentially duplicated here. Instead, we can either:
1. Interchange the operands, with `strpos` being called first
2. Retain $hasNamedBinds only, as it contains already the $hasBinds check

paulbalandan (changes requested):

system/Database/Query.php
kenjis: @paulbalandan 
1
I don't understand your intention. I removed duplicated `strpos()` call.

2
If I remove `&& ! $hasBinds`, it breaks.
https://3v4l.org/UWTvF

system/Database/Query.php
kenjis: @samsonasik I updated the link.

system/Database/Query.php
paulbalandan: Okay. I stand corrected with my second suggestion. Basically, I want something like this.
https://3v4l.org/V4pJv

paulbalandan (approved):
I overlooked a detail in the later part. Sorry for the noise.

paulbalandan:
Thank you, @kenjis

vlakoff:
I have checked the differences between the two codes, and I don't see where is the functional difference.

The change in code is that a `! $hasBinds` condition has been added when determining `$hasNamedBinds`,
but nevertheless the early return is triggered when both `! $hasNamedBinds && ! $hasBinds`.

So, if `$hasBinds` is true, there can't be early return, and if `$hasBinds` is false, `$hasNamedBinds` gets determined as previously.

kenjis:
What are the two codes?

`$hasBinds` means the query contains bindings, that is `?`.
`$hasNamedBinds` means the query contains named bindings, that is something like`:name:`.
They are exclusive. We cannot use both at a time.
So I added `! $hasBinds` condition when determining `$hasNamedBinds`.

`! $hasNamedBinds && ! $hasBinds` means the query does not have `:name:` nor `?`.
That is no bindings at all. So we have nothing to process bindings.
So we go early return.

vlakoff:
I understood, what your change influences is actually a line that is a bit below in the code:

```php
$sql = $hasNamedBinds ? $this->matchNamedBinds($sql, $binds) : $this->matchSimpleBinds($sql, $binds, $bindCount, $ml);
```

vlakoff:
Though, the code would fail with named binds, simply if there is a "?" somewhere in the supplied SQL string.

As a quick example:
```sql
SELECT * FROM songs WHERE band = :band: AND title = 'Do You Love Me?'
```

`hasBinds` would be erroneously true, which in turn would make `hasNamedBinds` erroneously false.

vlakoff:
Refs #5138.

PR #5127, opened 2021-09-25
Allow calling getQuery() multiple times, and other improvements

vlakoff:
Main fix:
- `getQuery()` can now be called multiple times without trouble, `compileBinds()` will be called only on original SQL string. Previously, calling `getQuery()` additional times was executing `compileBinds()` again and again, and on the already compiled SQL string…

Additional improvements and notes:
- [Performances] If `getQuery()` is called multiple times, `compileBinds()` is executed only the first time.
- Calling `setQuery()` or `setBinds()` discard `finalQueryString`, so they can be called multiple times with new data, and the SQL will be generated with this new data.
- `swapPrefix()` will always be executed on the original SQL string (with the bind placeholders), which is a bit safer than on the final SQL string (as the search/replace just operates on the whole SQL string).
- [BC] As previously, it is possible to call `swapPrefix()` multiple times, to execute multiple prefix swaps.
- `swapPrefix()` can be called before or after `setBinds()` (but not before `setQuery()`, as it operates on `originalQueryString`).
- `debugToolbarDisplay()` would fail if called before `finalQueryString` is assigned, as `compileBinds()` was relying on `getQuery()` to assign this property beforehand. (refs #4534)

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -24,9 +24,16 @@ class Query implements QueryInterface
     protected $originalQueryString;
 
     /**
+     * The query string if table prefix has been swapped.
+     *
+     * @var string|null
+     */
+    protected $swappedQueryString;
+
+    /**
      * The final query string after binding, etc.
      *
-     * @var string
+     * @var string|null
      */
     protected $finalQueryString;
 
@@ -99,6 +106,7 @@ class Query implements QueryInterface
     public function setQuery(string $sql, $binds = null, bool $setEscape = true)
     {
         $this->originalQueryString = $sql;
+        unset($this->swappedQueryString);
 
         if ($binds !== null) {
             if (! is_array($binds)) {
@@ -115,6 +123,8 @@ class Query implements QueryInterface
             }
             $this->binds = $binds;
         }
+
+        unset($this->finalQueryString);
 
         return $this;
     }
@@ -134,6 +144,8 @@ class Query implements QueryInterface
 
         $this->binds = $binds;
 
+        unset($this->finalQueryString);
+
         return $this;
     }
 
@@ -144,10 +156,8 @@ class Query implements QueryInterface
     public function getQuery(): string
     {
         if (empty($this->finalQueryString)) {
-            $this->finalQueryString = $this->originalQueryString;
+            $this->compileBinds();
         }
-
-        $this->compileBinds();
 
         return $this->finalQueryString;
     }
@@ -251,9 +261,14 @@ class Query implements QueryInterface
      */
     public function swapPrefix(string $orig, string $swap)
     {
-        $sql = empty($this->finalQueryString) ? $this->originalQueryString : $this->finalQueryString;
+        $sql = $this->swappedQueryString ?? $this->originalQueryString;
 
-        $this->finalQueryString = preg_replace('/(\W)' . $orig . '(\S+?)/', '\\1' . $swap . '\\2', $sql);
+        $from = '/(\W)' . $orig . '(\S+?)/';
+        $to   = '\\1' . $swap . '\\2';
+
+        $this->swappedQueryString = preg_replace($from, $to, $sql);
+
+        unset($this->finalQueryString);
 
         return $this;
     }
@@ -267,16 +282,18 @@ class Query implements QueryInterface
     }
 
     /**
-     * Escapes and inserts any binds into the finalQueryString object.
+     * Escapes and inserts any binds into the finalQueryString property.
      *
      * @see https://regex101.com/r/EUEhay/5
      */
     protected function compileBinds()
     {
-        $sql   = $this->finalQueryString;
+        $sql   = $this->swappedQueryString ?? $this->originalQueryString;
         $binds = $this->binds;
 
         if (empty($binds)) {
+            $this->finalQueryString = $sql;
+
             return;
         }
 
@@ -391,11 +408,7 @@ class Query implements QueryInterface
             'WHERE',
         ];
 
-        if (empty($this->finalQueryString)) {
-            $this->compileBinds(); // @codeCoverageIgnore
-        }
-
-        $sql = esc($this->finalQueryString);
+        $sql = esc($this->getQuery());
 
         /**
          * @see https://stackoverflow.com/a/20767160
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -263,7 +263,7 @@ class Query implements QueryInterface
     {
         $sql = $this->swappedQueryString ?? $this->originalQueryString;
 
-        $from = '/(\W)' . $orig . '(\S+?)/';
+        $from = '/(\W)' . $orig . '(\S)/';
         $to   = '\\1' . $swap . '\\2';
 
         $this->swappedQueryString = preg_replace($from, $to, $sql);
--- a/tests/system/Database/BaseQueryTest.php
+++ b/tests/system/Database/BaseQueryTest.php
@@ -420,6 +420,163 @@ final class BaseQueryTest extends CIUnitTestCase
         $this->assertSame($expected, $query->getQuery());
     }
 
+    public function testGetQueryMultipleTimes()
+    {
+        $query = new Query($this->db);
+
+        $query->setQuery('SELECT * FROM users WHERE name = :name:', ['name' => 'placeholder :name: in value']);
+
+        $expected = "SELECT * FROM users WHERE name = 'placeholder :name: in value'";
+
+        // Triggers compileBinds()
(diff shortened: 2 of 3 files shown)


vlakoff:
I added a handful of tests, as a first draft. Suggestions are welcome.

Maybe it could be polished. Maybe some tests are redundant, maybe some cases are still not covered (combinations)…

Note I couldn't (yet?) write a test where simple binds markers could be erroneously replaced twice, as the method `Query::matchSimpleBinds()` ensures avoiding bind markers in strings.

vlakoff:
To summarize the approach of this PR:

I'm adding an optional, intermediary state "swappedQueryString". Starting from the original query string, two changes can be applied to this original string: swapping the db prefix, and compiling the binds. We have to always keep record of the original string, and we also need a state to know if the binds have been compiled (here "finalQueryString" fills this purpose, by being discarded when compiling has to be done again). We can't apply the swaps to "finalQueryString" as the property would no longer let us know when bind compiling is needed. And of course we should not apply the swaps to "originalQueryString", as its purpose is to keep the user-supplied string untouched. Thus, we need to record an additional state, which is "swappedQueryString" here.

An alternative would have been to store an array of prefix swaps to make (a list of "from → to" pairs), and apply them to the original string, before the binds, when compiling. But nevertheless this would make us record an additional state as well, and the "swappedQueryString" approach is simpler.

vlakoff:
The code itself should be ready. I'm just a bit unsure that the tests are perfect (for instance: remaining uncovered edge cases, and where I split hairs: redundant tests!). I should have a fresh look at these, and I'm open to reviews.

paulbalandan:
This has been at draft status for couple of months now. Where are we on this? Or should this be closed?

kenjis:
@vlakoff Do you mean this PR is completed and ready for review?
Or do you still have something to do?

vlakoff:
I have squashed the commits.
Information provided in this PR should be quite exhaustive (and exhausting).
Test coverage seems to be pretty complete, all the cases I'm thinking of are covered.

Therefore, marking the PR for review.

system/Database/Query.php
```diff
@@ -23,10 +23,17 @@ class Query implements QueryInterface
      */
     protected $originalQueryString;
 
+    /**
+     * The query string if table prefix has been swapped.
```
kenjis, line 27:
```suggestion
     * The query string after table prefix has been swapped and compiled the binds.
```

tests/system/Database/BaseQueryTest.php
```diff
@@ -420,6 +420,163 @@ public function testSetQueryBinds()
...
+    {
+        $query = new Query($this->db);
+
+        $origSQL = 'SELECT * FROM db_users WHERE db_users.name = ?';
+        $newSQL  = "SELECT * FROM ci_users WHERE ci_users.name = 'John db_foobar John'";
+
+        $query->setQuery($origSQL);
+
+        $query->setBinds(['John db_foobar John']);
+        $query->swapPrefix('db_', 'ci_');
```
kenjis, line 559:
```suggestion
        $query->swapPrefix('db_', 'ci_');
        $query->setBinds(['John db_foobar John']);
```

vlakoff: No, the purpose of this test is precisely to ensure that if we call, in this order, `setBinds()` then `swapPrefix()`, the binds are still applied *after* the prefix swap.

The reason is that [`swapPrefix()`](https://github.com/codeigniter4/CodeIgniter4/blob/4e3792d1382518c2f6f89d03aab857356c45a040/system/Database/Query.php#L252-L259) is quite dumb, and just operates on the whole query string. If it were applied after the binds, it would also replace in the binded values, as demonstrated in this test.

As I listed in the PR description:
> - `swapPrefix()` will always be executed on the original SQL string (with the bind placeholders), which is a bit safer than on the final SQL string (as the search/replace just operates on the whole SQL string).

kenjis: Okay, I got it.

tests/system/Database/BaseQueryTest.php
```diff
@@ -448,7 +605,6 @@ public function testHighlightQueryKeywords($expected, $sql)
     {
         $query = new Query($this->db);
         $query->setQuery($sql);
-        $query->getQuery();
```
kenjis, line 451: Why do you remove it?

tests/system/Database/BaseQueryTest.php
```diff
@@ -420,6 +420,163 @@ public function testSetQueryBinds()
...
+        $origPrefix = 'db_';
+        $newPrefix1 = 'ci_';
+        $newPrefix2 = 'zz_';
+
+        $origSQL = 'SELECT * FROM db_users WHERE db_users.id = 1';
+        $newSQL  = 'SELECT * FROM zz_users WHERE zz_users.id = 1';
+
+        $query->setQuery($origSQL);
+        $query->swapPrefix($origPrefix, $newPrefix1);
+        $query->swapPrefix($newPrefix1, $newPrefix2);
```
kenjis, line 514: I think there is no use case that calls swapPrefix() more than once.

kenjis:
I think the user guide should be updated. Just remove the "final" ?

> swapPrefix()
Replaces one table prefix with another value **in the final SQL**. The first parameter is the original prefix that you want replaced, and the second parameter is the value you want it replaced with:
https://codeigniter4.github.io/CodeIgniter4/database/queries.html?highlight=swapprefix#working-with-query-objects

kenjis:
@vlakoff You missed GPG sign.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/signing.md

kenjis:
> [BC] As previously, it is possible to call swapPrefix() multiple times, to execute multiple prefix swaps.

Why do you need to make it possible?
I think there is no use case.

kenjis:
The branch coverage is 84.75%, good.

Screenshot 2022-01-24 at 13-51-49 Code Coverage for Users kenji work codeigniter CodeIgniter4 system Database Query php

cijagani:
@kenjis how to do this setup? Can you please guide me to do this? Which editor you are using?

kenjis:
@cijagani The coverage report? It is a HTML file generated by PHPUnit.

```
$ php -d memory_limit=-1 vendor/bin/phpunit --path-coverage tests/system/Database/BaseQueryTest.php --coverage-html build/coverage
```

cijagani:
@kenjis oh wow. thank you.

tests/system/Database/BaseQueryTest.php
vlakoff: This had been noticed previously, see [above](https://github.com/codeigniter4/CodeIgniter4/pull/5127#pullrequestreview-781371440).

It's because `getQuery()` can now be called multiple times without error, I could remove the ugly hack [here](https://github.com/codeigniter4/CodeIgniter4/pull/5127/files#diff-33f862bc94ae2386d78473bf248d6bf3f04c342f3044ce52870dd7ddf7b0c37fR411). That hack was in place so that `getQuery()` isn't called an additional time.

But… it required that `getQuery()` had been be called beforehand, so that `finalQueryString` is set, as `compileBinds()` needed it to be set beforehand… hence the "artificial" `getQuery()` here, just so that `finalQueryString` is set.

tests/system/Database/BaseQueryTest.php
vlakoff: It is possible to swap prefix several times with the current code, so I made sure it was still possible with my new code, just in order to not introduce any regression.

Still, it should be worth covering this with a test, to ensure the behaviour is consistent.

vlakoff:
It has been a bit of a hassle to configure, but these commits are now signed.

tests/system/Database/BaseQueryTest.php
kenjis: Okay.

tests/system/Database/BaseQueryTest.php
kenjis: I see.

vlakoff:
> I think the user guide should be updated. Just remove the "final" ?

Done: https://github.com/codeigniter4/CodeIgniter4/pull/5127/commits/0839468402eaa50f5f7548f4bb8c5636669079db

kenjis (approved):
Thank you!
LGTM

samsonasik (approved):

kenjis:
@vlakoff @samsonasik Thank you!

PR #5129, opened 2021-09-25
Fix highlighting in database debug toolbar

vlakoff:
Highlighting wasn't working for multiple-word terms, as they were searched in the SQL string with nbsp entities instead of regular spaces.

This code could be optimized, though it's not crucial as it's a debug function.

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -372,23 +372,23 @@ class Query implements QueryInterface
             'FROM',
             'WHERE',
             'AND',
-            'LEFT&nbsp;JOIN',
-            'RIGHT&nbsp;JOIN',
+            'LEFT JOIN',
+            'RIGHT JOIN',
             'JOIN',
-            'ORDER&nbsp;BY',
-            'GROUP&nbsp;BY',
+            'ORDER BY',
+            'GROUP BY',
             'LIMIT',
             'INSERT',
             'INTO',
             'VALUES',
             'UPDATE',
-            'OR&nbsp;',
+            'OR ',
             'HAVING',
             'OFFSET',
-            'NOT&nbsp;IN',
+            'NOT IN',
             'IN',
             'LIKE',
-            'NOT&nbsp;LIKE',
+            'NOT LIKE',
             'COUNT',
             'MAX',
             'MIN',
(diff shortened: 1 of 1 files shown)


vlakoff:
Added a commit implementing replacement using regexp with word boundaries, to avoid overlaps / false positives with strings such as "OR", "IN"…

sfadschm:
Can you add before/after schreenshots?

system/Database/Query.php
```diff
@@ -406,10 +404,15 @@ public function debugToolbarDisplay(): string
...
-        }
+        $search = '/\b(?:' . implode('|', $highlightWords) . ')\b/';
+
+        $sql = preg_replace_callback($search, static function ($matches) {
+            return '<strong>' . str_replace(' ', '&nbsp;', $matches[0]) . '</strong>';
+        }, $sql);
+
+        $sql = str_replace(['(', ')'], ['<strong>(</strong>', '<strong>)</strong>'], $sql);
 
+        /** @noRector \Rector\CodeQuality\Rector\Return_\SimplifyUselessVariableRector */
```
paulbalandan, line 415: I would agree with Rector here to simplify the return. Besides, our coding standard also includes this rule to simplify the return when possible (although it is not taking effect here).

vlakoff:
I confirmed the fix, but if ever you want screenshots of this toolbar content, you may have a look at #5136.

sfadschm:
> I confirmed the fix, but if ever you want screenshots of this toolbar content, you may have a look at #5136.

There is unfortunately none of the key words changed here visible in the other PR. Would be great to see just one example what has been fixed in this PR, too.

paulbalandan (changes requested):
Since this PR addresses an error in highlighting multiple-word identifiers, please add before/after screenshots, as @sfadschm suggested, that show there are indeed problems with the current and how this PR fixes them.

vlakoff:
Before:
5129-before

After:
5129-after

vlakoff:
Maybe we should add the ASC and DESC keywords?

vlakoff:
Also, I'd suggest either highlighting commas in addition to parentheses, or not highlighting parentheses. Your pick?

sfadschm:
I think adding the keywords would be good, but would rather remove the parentheses. But lets wait for other calls.

kenjis:
I would like removing parentheses, adding ASC and DESC.

paulbalandan:
Let's highlight only keywords, leaving out parentheses and other punctuation marks.

vlakoff:
* Removed the highlighting of parentheses
* Added ASC and DESC keywords
* Reduced the number of commits by removing the commits related to the Rector warning

paulbalandan (approved):

paulbalandan:
Thank you, @vlakoff

PR #5132, opened 2021-09-25
Always escape identifiers in the set(), setUpdateBatch(), and insertBatch()

ytetsuro:
Fixed a problem where identifiers were not escaped in the set, setUpdateBatch, and insertBatch methods. 

**Description**

I decided that there was no case where the column name in the set clause did not need to be escaped, so I turned off the unescaping process.
The same goes for the table name when inserting.

see: https://github.com/codeigniter4/CodeIgniter4/pull/2487#discussion_r716012806

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Database/Builder/UpdateTest.php
+++ b/tests/system/Database/Builder/UpdateTest.php
@@ -235,6 +235,59 @@ final class UpdateTest extends CIUnitTestCase
         $this->assertSame($expected, $query->getQuery());
     }
 
+    public function testSetUpdateBatchWithoutEscape()
+    {
+        $builder = new BaseBuilder('jobs', $this->db);
+        $escape  = false;
+
+        $builder->setUpdateBatch([
+            [
+                'id'          => 2,
+                'name'        => 'SUBSTRING(name, 1)',
+                'description' => 'SUBSTRING(description, 3)',
+            ],
+            [
+                'id'          => 3,
+                'name'        => 'SUBSTRING(name, 2)',
+                'description' => 'SUBSTRING(description, 4)',
+            ],
+        ], 'id', $escape);
+
+        $this->db->shouldReturn('execute', 1)->shouldReturn('affectedRows', 1);
+        $builder->updateBatch(null, 'id');
+
+        $query = $this->db->getLastQuery();
+        $this->assertInstanceOf(MockQuery::class, $query);
+
+        $space = ' ';
+
+        $expected = <<<EOF
+            UPDATE "jobs" SET "name" = CASE{$space}
+            WHEN "id" = :id: THEN :name:
+            WHEN "id" = :id.1: THEN :name.1:
+            ELSE "name" END, "description" = CASE{$space}
+            WHEN "id" = :id: THEN :description:
+            WHEN "id" = :id.1: THEN :description.1:
+            ELSE "description" END
+            WHERE "id" IN(:id:,:id.1:)
+            EOF;
+
+        $this->assertSame($expected, $query->getOriginalQuery());
+
+        $expected = <<<EOF
+            UPDATE "jobs" SET "name" = CASE{$space}
+            WHEN "id" = 2 THEN SUBSTRING(name, 1)
+            WHEN "id" = 3 THEN SUBSTRING(name, 2)
+            ELSE "name" END, "description" = CASE{$space}
+            WHEN "id" = 2 THEN SUBSTRING(description, 3)
+            WHEN "id" = 3 THEN SUBSTRING(description, 4)
+            ELSE "description" END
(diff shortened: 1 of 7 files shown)


kenjis:
You need to update the PHPdoc.
https://github.com/codeigniter4/CodeIgniter4/blob/ef08b6f4298678da304e5663f5862043338d3451/system/Database/BaseBuilder.php#L1343

And user guide, too.
https://codeigniter4.github.io/CodeIgniter4/database/query_builder.html#set

ytetsuro:
> You need to update the PHPdoc.
> And user guide, too.

Thanks, response.
Fixed it.

Thanks to your pointers, I realized that I have a problem similar to this case with the setUpdateBatch and insertBatch identifiers.
Thank you very much.
I've fixed this as well and added a test.

kenjis:
@ytetsuro 
Your commit message like the following means the opposite what you did.

> fix: removed unnecessary escaping process in setUpdateBatch as well. 

It should be like `fix: always escape identifiers in setUpdateBatch()`

ytetsuro:
@kenjis 

> Your commit message like the following means the opposite what you did.

Thanks, response.

You are right.

Fixed it.💪

kenjis:
@ytetsuro 

The first commit:
> fix: Because we have made escaping unnecessary in places where it is not necessary.

Sorry, I don't know what you are saying.

DeepL translation:
> 逃げることを必要としない場所では、逃げることを不要としたからです。

Google translation:
> 必要のない場所での脱出を不要にしたからです。

ytetsuro:
> I don't know what you are saying.

Sorry.
Fixed it.💪

kenjis:
I didn't understand it before, but this change is necessary for the normal use case in Oracle.

The normal use case is when an identifier is quoted and lowercased, like `"name"`.

In this case, if a user specify `$escape` to false, the identifiers will not be quoted, and Oracle will not find the identifier.
`name` is not equal to `"name"`.

See https://seeq.atlassian.net/wiki/spaces/KB/pages/443088907/SQL+Column+Names+and+Case+Sensitivity
for Oracle identifier case sensitivity.

kenjis (approved):

tests/system/Database/Builder/UpdateTest.php
```diff
@@ -350,7 +403,7 @@ public function testSetWithoutEscape()
             ->where('id', 2)
             ->update(null, null, null);
 
-        $expectedSQL   = 'UPDATE "mytable" SET field = field+1 WHERE "id" = 2';
+        $expectedSQL   = 'UPDATE "mytable" SET "field" = field+1 WHERE "id" = 2';
```
kenjis, line 406: @ytetsuro Does this SQL statement run without errors on Oracle?

kenjis (changes requested):
Cancel the approval once.

tests/system/Database/Builder/UpdateTest.php
kenjis: We must write code for Oracle like this:
```php
$builder
    ->set('field', '"field"+1', false)
    ->where('id', 2)
    ->update(null, null, null);
```
But we still need this PR to quote the `field` immediately after `SET`.

kenjis (approved):

tests/system/Database/Builder/UpdateTest.php
ytetsuro: > Does this SQL statement run without errors on Oracle?

Thanks, response.

No.
As you can imagine, the `field ` is not escaped, so an error occurs.

ytetsuro:
@kenjis 

I'm sorry to have bothered you.
I've aligned the commit messages.

paulbalandan (commented):
Let us include this change in the user guide changelog.

ytetsuro:
@paulbalandan 

I'm sorry.
I have no experience in writing changelog.

Please tell me.
Is the changelog to be updated using the following procedure?
Is it my CI4 repository to push the release branch?

https://github.com/codeigniter4/CodeIgniter4/blob/0f0c1d6514c0234f4d16bdca0e5e308a39168986/admin/RELEASE.md#codeigniter4

paulbalandan:
No, not there. I meant here. 
https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/changelogs/v4.1.5.rst

ytetsuro:
@paulbalandan 

Thanks, response.

Fixed it.

paulbalandan (approved):

paulbalandan:
Thank you, @ytetsuro

PR #5133, opened 2021-09-26
Fixed the problem that _createTable does not take into account that it returns true.

ytetsuro:
Fixed the problem that `_createTable` does not take into account that it returns true.

**Description**

If `_createTable` returns true, then `createTable` will return true.
This occurs only when the `createTableIfStr` property is false in the Forge class.
I had trouble with this behavior when I was creating the OCI8 driver.

see: https://github.com/codeigniter4/CodeIgniter4/pull/2487#discussion_r716022511

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -466,6 +466,8 @@ class Forge
 
                 return false;
             }
+
+            return true;
         }
 
         if (($result = $this->db->query($sql)) !== false) {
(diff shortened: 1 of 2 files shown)


kenjis (approved):
LGTM

PR #5138, opened 2021-09-26
Determine if binds are simple or named by looking at the $binds array

vlakoff:
To determine if binds are simple or named, rather than trying to deduce it from the SQL string, which is the source of many errors (for example see #5114, and all the issues it links to), simply look at the `$binds` array. If it's an indexed array, the binds are simple, if it's an associative array, the binds are named.

Note that if #5127 is merged before, this PR will need a slight rebasing before merging.

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -288,7 +288,7 @@ class Query implements QueryInterface
         }
 
         if (is_numeric(key(array_slice($binds, 0, 1)))) {
-            if (empty($this->bindMarker) || strpos($sql, $this->bindMarker) === false) {
+            if (empty($this->bindMarker)) {
                 return;
             }
 
@@ -296,10 +296,6 @@ class Query implements QueryInterface
 
             $this->finalQueryString = $this->matchSimpleBinds($sql, $binds, $bindCount, $ml);
         } else {
-            if (! preg_match('/:(?!=).+:/', $sql)) {
-                return;
-            }
-
             // Reverse the binds so that duplicate named binds
             // will be processed prior to the original binds.
             $binds = array_reverse($binds);
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -288,10 +288,6 @@ class Query implements QueryInterface
         }
 
         if (is_numeric(key(array_slice($binds, 0, 1)))) {
-            if (empty($this->bindMarker)) {
-                return;
-            }
-
             $ml = strlen($this->bindMarker);
 
             $this->finalQueryString = $this->matchSimpleBinds($sql, $binds, $bindCount, $ml);
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -281,7 +281,7 @@ class Query implements QueryInterface
 
         $binds = (array) $this->binds;
 
-        if (is_int(key(array_slice($binds, 0, 1)))) {
+        if (is_int(array_key_first($binds))) {
             $bindCount = count($binds);
             $ml        = strlen($this->bindMarker);
 
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -281,7 +281,7 @@ class Query implements QueryInterface
 
         $binds = (array) $this->binds;
 
-        if (is_numeric(key(array_slice($binds, 0, 1)))) {
+        if (is_int(key(array_slice($binds, 0, 1)))) {
             $bindCount = count($binds);
             $ml        = strlen($this->bindMarker);
 
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -279,16 +279,11 @@ class Query implements QueryInterface
             return;
         }
 
-        if (! is_array($this->binds)) {
-            $binds     = [$this->binds];
-            $bindCount = 1;
-        } else {
-            $binds     = $this->binds;
-            $bindCount = count($binds);
-        }
+        $binds = (array) $this->binds;
 
         if (is_numeric(key(array_slice($binds, 0, 1)))) {
-            $ml = strlen($this->bindMarker);
+            $bindCount = count($binds);
+            $ml        = strlen($this->bindMarker);
 
             $this->finalQueryString = $this->matchSimpleBinds($sql, $binds, $bindCount, $ml);
         } else {
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -275,14 +275,7 @@ class Query implements QueryInterface
     {
         $sql = $this->finalQueryString;
 
-        $hasBinds      = strpos($sql, $this->bindMarker) !== false;
-        $hasNamedBinds = ! $hasBinds
-            && preg_match('/:(?!=).+:/', $sql) === 1;
-
-        if (empty($this->binds)
-            || empty($this->bindMarker)
-            || (! $hasNamedBinds && ! $hasBinds)
-        ) {
+        if (empty($this->binds)) {
             return;
         }
 
@@ -294,16 +287,25 @@ class Query implements QueryInterface
             $bindCount = count($binds);
         }
 
-        // Reverse the binds so that duplicate named binds
-        // will be processed prior to the original binds.
-        if (! is_numeric(key(array_slice($binds, 0, 1)))) {
+        if (is_numeric(key(array_slice($binds, 0, 1)))) {
+            if (empty($this->bindMarker) || strpos($sql, $this->bindMarker) === false) {
+                return;
+            }
+
+            $ml = strlen($this->bindMarker);
+
+            $this->finalQueryString = $this->matchSimpleBinds($sql, $binds, $bindCount, $ml);
+        } else {
+            if (! preg_match('/:(?!=).+:/', $sql)) {
+                return;
+            }
+
+            // Reverse the binds so that duplicate named binds
+            // will be processed prior to the original binds.
             $binds = array_reverse($binds);
+
+            $this->finalQueryString = $this->matchNamedBinds($sql, $binds);
         }
-
-        $ml  = strlen($this->bindMarker);
-        $sql = $hasNamedBinds ? $this->matchNamedBinds($sql, $binds) : $this->matchSimpleBinds($sql, $binds, $bindCount, $ml);
-
-        $this->finalQueryString = $sql;
     }
 
     /**
--- a/tests/system/Database/BaseQueryTest.php
+++ b/tests/system/Database/BaseQueryTest.php
@@ -269,12 +269,36 @@ final class BaseQueryTest extends CIUnitTestCase
     /**
      * @see https://github.com/codeigniter4/CodeIgniter4/issues/3566
      */
-    public function testNamedBindsWithColonElseWhere()
+    public function testNamedBindsWithColonElsewhere()
     {
         $query = new Query($this->db);
         $query->setQuery('SELECT `email`, @total:=(total+1) FROM `users` WHERE `id` = :id:', ['id' => 10]);
 
         $sql = 'SELECT `email`, @total:=(total+1) FROM `users` WHERE `id` = 10';
+        $this->assertSame($sql, $query->getQuery());
+    }
+
+    /**
+     * @see https://github.com/codeigniter4/CodeIgniter4/pull/5138
+     */
+    public function testNamedBindsWithBindMarkerElsewhere()
+    {
+        $query = new Query($this->db);
+        $query->setQuery('SELECT * FROM posts WHERE id = :id: AND title = \'The default bind marker is "?"\'', ['id' => 10]);
+
+        $sql = 'SELECT * FROM posts WHERE id = 10 AND title = \'The default bind marker is "?"\'';
+        $this->assertSame($sql, $query->getQuery());
+    }
+
+    /**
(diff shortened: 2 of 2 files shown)


kenjis:
Coding Standard checking was failed.
Run `composer cs-fix`, please.

vlakoff:
On a related note, is the test `empty($this->bindMarker)` really useful?

To trigger it, the user should extend the Query class and assign the bindMarker property an empty value…

vlakoff:
The PR currently makes an unit test fail:

> 1) [CodeIgniter\Database\Builder\WhereTest](https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/Builder/WhereTest.php)::testWhereValueClosure
> Error: Object of class Closure could not be converted to string
> 
> /home/runner/work/CodeIgniter4/CodeIgniter4[/system/Database/Query.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/Query.php):325
> /home/runner/work/CodeIgniter4/CodeIgniter4[/system/Database/Query.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/Query.php):298
> /home/runner/work/CodeIgniter4/CodeIgniter4[/system/Database/Query.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/Query.php):150
> /home/runner/work/CodeIgniter4/CodeIgniter4[/system/Database/BaseBuilder.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseBuilder.php):1411
> /home/runner/work/CodeIgniter4/CodeIgniter4[/system/Database/BaseBuilder.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseBuilder.php):1395
> /home/runner/work/CodeIgniter4/CodeIgniter4[/system/Database/BaseBuilder.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseBuilder.php):676
> /home/runner/work/CodeIgniter4/CodeIgniter4[/system/Database/BaseBuilder.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Database/BaseBuilder.php):606
> /home/runner/work/CodeIgniter4/CodeIgniter4[/tests/system/Database/Builder/WhereTest.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/Builder/WhereTest.php):134

See for instance at: https://github.com/codeigniter4/CodeIgniter4/pull/5138/checks?check_run_id=3712623836

I've already spend a lot of time investigating this. It seems to be due to some stale binds data that should be cleaned between query compilations.

kenjis:
`$replacersTo[1]` is a Closure. It is from `$binds['advance_amount']`

Screenshot 2021-09-27 18 18 30

system/Database/Query.php
```diff
@@ -324,10 +318,11 @@ protected function matchNamedBinds(string $sql, array $binds): string
                 $escapedValue = '(' . implode(',', $escapedValue) . ')';
             }
 
-            $replacers[":{$placeholder}:"] = $escapedValue;
+            $replacersFrom[] = ":{$placeholder}:";
+            $replacersTo[]   = $escapedValue;
         }
 
-        return strtr($sql, $replacers);
+        return str_replace($replacersFrom, $replacersTo, $sql);
```
kenjis, line 325: Why do you replace `strtr()` with `str_replace()`?

kenjis:
This passed that test.

```diff
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -11,6 +11,8 @@
 
 namespace CodeIgniter\Database;
 
+use Closure;
+
 /**
  * Query builder
  */
@@ -301,6 +303,8 @@ class Query implements QueryInterface
 
     /**
      * Match bindings
+     *
+     * @param array<string, array{0: int|string|Closure, 1: bool}> $binds
      */
     protected function matchNamedBinds(string $sql, array $binds): string
     {
@@ -308,6 +312,11 @@ class Query implements QueryInterface
         $replacersTo   = [];
 
         foreach ($binds as $placeholder => $value) {
+            // If $value[0] is Closure, it is a sub query. We don't need to replace.
+            if ($value[0] instanceof Closure) {
+                continue;
+            }
+
             // $value[1] contains the boolean whether should be escaped or not
             $escapedValue = $value[1] ? $this->db->escape($value[0]) : $value[0];
 
```

vlakoff:
> `$replacersTo[1]` is a Closure. It is from `$binds['advance_amount']`

Indeed. I have pinpointed the deep cause of this, and have to work on a fix.

Basically, we really shouldn't have a closure here. The current code works because it doesn't look at the binds array but at the SQL string, which doesn't have :placeholders: at that point, so it returns early and doesn't go to the closure.

vlakoff:
> This passed that test.
> 
> (...)

Thanks. Along the fix for the root cause, maybe such guard should also be added to prevent errors coming from other places.

system/Database/Query.php
vlakoff: Per my commit diff:

> strtr() is purposed for single-character replacements, and less performant when used otherwise.

vlakoff:
On a related note, in this call:

```php
$this->matchSimpleBinds($sql, $binds, $bindCount, $ml);

protected function matchSimpleBinds(string $sql, array $binds, int $bindCount, int $ml): string
```

It's redundant to pass `$bindCount` and `$ml`, as their values can easily be deduced from `$binds` and `$this->bindMarker` respectively.

However, it would be a BC break to change the method signature…

vlakoff:
Now that the prerequisite #5247 has been merged, I think this PR should be ready to be merged.

kenjis:
@vlakoff This branch is too old.
Could you rebase it?

https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#pushing-your-branch

vlakoff:
Rebased, although there were no merge conflicts.

kenjis (approved):
LGTM

This is cleaner.

vlakoff:
This conditional to determine if the array is a list or an associative map may be improved:

```php
if (is_numeric(key(array_slice($binds, 0, 1)))) { ... }
```

First, rather than [`is_numeric()`](https://www.php.net/manual/en/function.is-numeric.php) we should use the stricter [`is_int()`](https://www.php.net/manual/en/function.is-int.php). As a reminder, PHP automatically casts array keys like `'1'` (string representing an integer) to `1` (integer).

```php
if (is_int(key(array_slice($binds, 0, 1)))) { ... }
```

Then, a possible optimization could be to use a `foreach` instead for the `array_slice()`:

```php
foreach ($binds as $key => $value) {
    $array_is_list = is_int($key);
    break;
}
```

Finally, we could check `$key === 0`. But the user might have omitted some values, who knows, so it could be a BC break.
Yet, we could ensure `$key >= 0` (i.e. is not a negative integer). But I don't think adding such test would be necessary in practice.

vlakoff:
Mmmm…

* https://github.com/vlakoff/CodeIgniter4/pull/5138/commits/05bab11660c58711102c8926d75e5e1f247f5a33 isn't really clearer after all…
* https://github.com/vlakoff/CodeIgniter4/pull/5138/commits/42f0021e39bafb7af3db2a5066053aa76f0a6ea4 makes the [global diff](https://github.com/codeigniter4/CodeIgniter4/pull/5138/files) a mess…

system/Database/Query.php
```diff
@@ -275,35 +275,24 @@ protected function compileBinds()
...
         }
 
-        if (! is_array($this->binds)) {
-            $binds     = [$this->binds];
-            $bindCount = 1;
-        } else {
-            $binds     = $this->binds;
+        $binds = is_array($this->binds) ? $this->binds : [$this->binds];
+
+        if (is_int(key(array_slice($binds, 0, 1)))) {
```
lonnieezell, line 284: Is `array_slice` necessary here? Since the array has just been created, the current index should be the first element, so I think just doing `is_int(key($binds))` is enough.

lonnieezell (approved):
Looks like a good set of changes that simplifies and should have a slight performance boost getting rid of the regex. Nicely done. 

I just have one question about `array_slice`. I realize that was in the original code, but looking at it in context now, I don't think it's necessary, but would love a second opinion on that.

system/Database/Query.php
paulbalandan: Or using `array_key_first`?

system/Database/Query.php
```diff
@@ -275,35 +275,24 @@ protected function compileBinds()
...
+        if (empty($this->binds)) {
             return;
         }
 
-        if (! is_array($this->binds)) {
-            $binds     = [$this->binds];
-            $bindCount = 1;
-        } else {
-            $binds     = $this->binds;
+        $binds = is_array($this->binds) ? $this->binds : [$this->binds];
```
paulbalandan, line 282:
```suggestion
        $binds = (array) $this->binds;
```

paulbalandan (commented):
Can we add some tests for this?

system/Database/Query.php
vlakoff: `$binds` may be a reference to `$this->binds`, and as PHP doesn't copy arrays until they are modified, the pointer position of `$binds` will be the same as of `$this->binds`, which may have been modifier beforehand, we can't know.

A benefit of `key(array_slice($binds, 0, 1))` over `reset($binds)` is that with the latter PHP would internally make a copy of the whole array, while the former just creates a tiny array of one item.

Good idea about [`array_key_first()`](https://www.php.net/manual/en/function.array-key-first.php). As we are targetting PHP 7.3+, that's probably what we want to use.

system/Database/Query.php
vlakoff: I can't think of any regressions with that change, so why not.

vlakoff:
> Can we add some tests for this?

Unit test aren't really my strong point…
(disclaimer: I have an aversion to writing tests, as I always fear there would be redundancies or missing spots)

If you can provide me tests that fail before and pass after this PR, I would happily add them.

vlakoff:
I added tests: https://github.com/vlakoff/CodeIgniter4/pull/5138/commits/d7ad8ae78e69022cf93ce3158e17a70d9c502f01.
I have yet to squash this commit (and apply the other suggestions), but you can already have a glance at these.

* `testNamedBindsWithBindMarkerElsewhere()`:
This PR makes this test pass.
* `testSimpleBindsWithNamedBindPlaceholderElsewhere()`:
The `! $hasBinds` condition in `$hasNamedBinds = ! $hasBinds && preg_match('/:(?!=).+:/', $sql) === 1` already made this test pass, but it had the side effect of making the other test fail…
(and in turn, removing that condition would make both `$hasBinds` and `$hasNamedBinds` true, which is wrong, and make the other test pass, but only because of the order in the ternary `$hasNamedBinds ? $this->matchNamedBinds(...) : $this->matchSimpleBinds(...)`)

vlakoff:
Force-pushed, squashing the unit tests and the code suggestions.

I don't think we could improve the PR even further :)

kenjis:
Coverage is perfect! 👍 

```
$ ./phpunit tests/system/Database/BaseQueryTest.php --coverage-html=tests/coverage/ --path-coverage -d memory_limit=1024m
```

Screenshot 2021-11-23 11 08 53

kenjis (approved):

paulbalandan (approved):
I think this is good for me. Thanks for the tests!

vlakoff:
I just thought of a possible regression with this change:
```php
- $binds = is_array($this->binds) ? $this->binds : [$this->binds];
+ $binds = (array) $this->binds;
```
If the provided binds are not an array or a scalar, but a "stringable object" (for example [Laravel's Str](https://laravel.com/docs/8.x/helpers#fluent-strings)). Here is how it is casted:

```
$value = Str::of('foobar');

// var_dump( $value );
object(Illuminate\Support\Stringable) {
  ["value":protected] => string "foobar"
}

// var_dump( (string) $value );
string "foobar"

// var_dump( (array) $value );
array {
  ["\0*\0value"] => string "foobar"
}
```

Therefore, the following code would now fail:

```php
$value = Str::of('foobar');

$sql = 'SELECT * FROM posts WHERE title = ?';
$db->query($sql, $value);
```

kenjis:
@vlakoff It is for all objects.
https://www.php.net/manual/en/language.types.array.php#language.types.array.casting

By the way, is there any possiblity that `$this->binds` is not an array?

vlakoff:
> It is for all objects.

I know it is for all objects, but passing a "stringable object" instead of a string seems to be a valid and possible use case.

> By the way, is there any possiblity that `$this->binds` is not an array?

* In method `setQuery()` we also have `if (! is_array($binds)) { $binds = [$binds]; }`.
* In method `setBinds()` the parameter `$binds` is array typed.

So `$this->binds` should be an array indeed. But I don't know, maybe the user would extend the class and assign to `$this->binds` in a new method.

Anyway, `$this->binds` is not expected to be assigned anything else than an array. Doing otherwise should be considered as an error.

kenjis:
`setQuery()`'s `$binds` seems to be `int|string|array`.

And I think the code `$binds = (array) $this->binds;` is not needed.

vlakoff:
Opened a PR for this: #5379.

PR #5139, opened 2021-09-26
Fix `Model::__call` throwing `BadMethodCallException` on empty results

### LINKED ISSUE #5137: Bug: affctedRows - BadMethodCallException
**Describe the bug**
Calling the affectedRows method through a model instance may throw a BadMethodCallException if the value is zero.

```php
$model = new Model()
$model->set('key', 'value')
            ->where('id', 'nonExistentID')
            ->update();
$model->affectedRows(); // BadMethodCallException 
$model->db->affectedRows(); // 0
```

Since the result of executing the method is zero, and ``empty(0)`` === true, the code below throws an exception.

https://github.com/codeigniter4/CodeIgniter4/blob/4649204306276af6c06bfbd229fb1895fb5b0e4e/system/Model.php#L723-L728

ytetsuro:
Fixes #5137.

**Description**

An error occurs when a method of the Database\Connection class returns a falsy value.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -719,7 +719,7 @@ class Model extends BaseModel
 
         if (method_exists($this->db, $name)) {
             $result = $this->db->{$name}(...$params);
-        } else if (method_exists($builder, $name)) {
+        } elseif (method_exists($builder, $name)) {
             $result = $builder->{$name}(...$params);
         } else {
             $className = static::class;
--- a/system/Model.php
+++ b/system/Model.php
@@ -722,9 +722,7 @@ class Model extends BaseModel
         } elseif (method_exists($builder, $name)) {
             $result = $builder->{$name}(...$params);
         } else {
-            $className = static::class;
-
-            throw new BadMethodCallException('Call to undefined method ' . $className . '::' . $name);
+            throw new BadMethodCallException('Call to undefined method ' . static::class . '::' . $name);
         }
 
         if ($result instanceof BaseBuilder) {
--- a/system/Model.php
+++ b/system/Model.php
@@ -710,7 +710,7 @@ class Model extends BaseModel
      * Provides direct access to method in the builder (if available)
      * and the database connection.
      *
-     * @return $this|null
+     * @return mixed
      */
     public function __call(string $name, array $params)
     {
--- a/system/Model.php
+++ b/system/Model.php
@@ -728,7 +728,6 @@ class Model extends BaseModel
             $result = $builder->{$name}(...$params);
         }
 
-
         if ($result instanceof BaseBuilder) {
             return $this;
         }
--- a/system/Model.php
+++ b/system/Model.php
@@ -714,18 +714,17 @@ class Model extends BaseModel
      */
     public function __call(string $name, array $params)
     {
-        try {
-            $result = parent::__call($name, $params);
-        } catch (BadMethodCallException $e) {
-            $builder = $this->builder();
+        $builder = $this->builder();
+        $result  = null;
 
-            if (! method_exists($builder, $name)) {
-                $className = static::class;
-
-                throw new BadMethodCallException('Call to undefined method ' . $className . '::' . $name);
-            }
-
+        if (method_exists($this->db, $name)) {
+            $result = $this->db->{$name}(...$params);
+        } else if (method_exists($builder, $name)) {
             $result = $builder->{$name}(...$params);
+        } else {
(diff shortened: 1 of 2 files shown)


system/BaseModel.php
```diff
@@ -1636,11 +1637,13 @@ public function __isset(string $name): bool
      */
     public function __call(string $name, array $params)
     {
-        if (method_exists($this->db, $name)) {
-            return $this->db->{$name}(...$params);
+        if (! method_exists($this->db, $name)) {
+            $className = static::class;
+
+            throw new BadMethodCallException('Call to undefined method ' . $className . '::' . $name);
```
kenjis, line 1643: It seems this is a breaking change.

paulbalandan: Yes, this is breaking for users using the `BaseModel` for their NoSQL models.

ytetsuro: @kenjis @paulbalandan 

Yes, It's a breaking change, that's for sure. However, I don't understand what the problem is.
I don't know what use case it would be better to call a non-existent method and return null when using NoSQL.
What do NoSQL users call non-existent methods for?

kenjis: We just can't merge breaking changes until the next major version (v5.0).
Because if we change the public API, user's applications may break.

paulbalandan: The problem is we do not know how they implemented this in their code. One I can think of is like this:
```php
class NoSqlModel extends BaseModel
{
    ...
}

// then somewhere in their code
$nosql = new NoSqlModel();

if ($nosql->doSomething() === null) {
    // do something else
}
```

In code above, the user expects a call to non-existent method to return `null`. However, if this be merged, their code would break because it will now throw `BadMethodCallException` instead.

ytetsuro: @kenjis @paulbalandan 

> We just can't merge breaking changes until the next major version (v5.0).
> The problem is we do not know how they implemented this in their code.

I understand.
Thank you.
FIxed it.

system/Model.php
```diff
@@ -714,20 +714,17 @@ public function __isset(string $name): bool
...
     public function __call(string $name, array $params)
     {
-        $result = parent::__call($name, $params);
+        $builder = $this->builder();
+        $result  = null;
 
-        if ($result === null && method_exists($builder = $this->builder(), $name)) {
+        if (method_exists($this->db, $name)) {
+            $result = $this->db->{$name}(...$params);
+        } else if (method_exists($builder, $name)) {
```
kenjis, line 722: @ytetsuro Fix coding style, please.
Run `composer cs-fix`.

system/Model.php
```diff
@@ -714,20 +714,17 @@ public function __isset(string $name): bool
...
-        }
-
-        if (empty($result)) {
-            if (! method_exists($this->builder(), $name)) {
-                $className = static::class;
-
-                throw new BadMethodCallException('Call to undefined method ' . $className . '::' . $name);
-            }
+        } else {
+            $className = static::class;
```
paulbalandan, line 726:
```suggestion
```

system/Model.php
```diff
@@ -714,20 +714,17 @@ public function __isset(string $name): bool
...
 
-            return $result;
+            throw new BadMethodCallException('Call to undefined method ' . $className . '::' . $name);
```
paulbalandan, line 727:
```suggestion
            throw new BadMethodCallException('Call to undefined method ' . static::class . '::' . $name);
```

system/Model.php
ytetsuro: @kenjis 

Thanks, response.
Fixed it.

system/Model.php
ytetsuro: @paulbalandan 

Thanks, review.

Fixed it.

paulbalandan (commented):
Last thing on my part, I guess. Can you check the return type in L713 of Model. I think it is not accurate and should be `mixed` instead.

ytetsuro:
@paulbalandan 

> Can you check the return type in L713 of Model. I think it is not accurate and should be mixed instead.

Thanks, review.
Fixed it.

kenjis (approved):

paulbalandan (approved):

paulbalandan:
Thank you, @ytetsuro

PR #5142, opened 2021-09-27
Error when value is an object - validating api data

daycry:
For example:
```json
{
	"configuration" :
        {
	        "consumerKey1" : "aa",
	        "consumerSecret" : "bb",
	        "oauthToken" : "cc",
	        "oauthTokenSecret" : "dd"
        }
}
```

**Description**

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -291,6 +291,8 @@ class Validation implements ValidationInterface
                 // if the $value is an array, convert it to as string representation
                 if (is_array($value)) {
                     $value = '[' . implode(', ', $value) . ']';
+                }elseif (\is_object($value)) {
+                    $value = \json_encode( $value );
                 }
 
                 $this->errors[$field] = $error ?? $this->getErrorMessage($rule, $field, $label, $param, $value);
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -291,8 +291,8 @@ class Validation implements ValidationInterface
                 // if the $value is an array, convert it to as string representation
                 if (is_array($value)) {
                     $value = '[' . implode(', ', $value) . ']';
-                }elseif (\is_object($value)) {
-                    $value = \json_encode( $value );
+                } elseif (\is_object($value)) {
+                    $value = \json_encode($value);
                 }
 
                 $this->errors[$field] = $error ?? $this->getErrorMessage($rule, $field, $label, $param, $value);
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -291,8 +291,8 @@ class Validation implements ValidationInterface
                 // if the $value is an array, convert it to as string representation
                 if (is_array($value)) {
                     $value = '[' . implode(', ', $value) . ']';
-                } elseif (\is_object($value)) {
-                    $value = \json_encode($value);
+                } elseif (is_object($value)) {
+                    $value = json_encode($value);
                 }
 
                 $this->errors[$field] = $error ?? $this->getErrorMessage($rule, $field, $label, $param, $value);
--- a/tests/system/Validation/ValidationTest.php
+++ b/tests/system/Validation/ValidationTest.php
@@ -634,26 +634,25 @@ final class ValidationTest extends CIUnitTestCase
         $config->baseURL = 'http://example.com/';
 
         $this->validation->setRules([
-            'configuration'   => 'required|check_object_rule'
+            'configuration' => 'required|check_object_rule',
         ]);
 
-        $data = (Object) array( 'configuration' => (Object)array( 'first' => 1, 'second' => 2 ) );
-        $this->validation->run( (array)$data );
+        $data = (object) ['configuration' => (object) ['first' => 1, 'second' => 2]];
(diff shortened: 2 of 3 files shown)


samsonasik:
Please provide unit test and run cs fix:

```
vendor/bin/php-cs-fixer fix system/Validation/Validation.php
```

daycry:
Done.

samsonasik:
please run cs fix

daycry:
Done, sorry.

samsonasik:
cs still error, see https://github.com/codeigniter4/CodeIgniter4/pull/5142/checks?check_run_id=3733816746#step:8:20

daycry:
ok,
I have launched the processes for fix it

- vendor/bin/php-cs-fixer fix .\tests\system\Validation\ValidationTest.php
- vendor/bin/php-cs-fixer fix .\tests\_support\Validation\TestRules.php

tests/system/Validation/ValidationTest.php
```diff
@@ -628,6 +628,33 @@ public function testTagReplacement()
         $this->assertSame($expected, $errors['Username']);
     }
 
+    public function testRulesForObjectField()
+    {
+        $config          = new App();
+        $config->baseURL = 'http://example.com/';
```
kenjis, line 634: Why do you need `$config`?
It seems it is not used.

system/Validation/Validation.php
```diff
@@ -291,6 +291,8 @@ protected function processRules(string $field, ?string $label, $value, $rules =
                 // if the $value is an array, convert it to as string representation
                 if (is_array($value)) {
                     $value = '[' . implode(', ', $value) . ']';
+                } elseif (\is_object($value)) {
+                    $value = \json_encode($value);
```
paulbalandan, line 295:
```suggestion
                } elseif (is_object($value)) {
                    $value = json_encode($value);
```

tests/system/Validation/ValidationTest.php
```diff
@@ -628,6 +628,33 @@ public function testTagReplacement()
...
+
+        $this->validation->setRules([
+            'configuration' => 'required|check_object_rule',
+        ]);
+
+        $data = (object) ['configuration' => (object) ['first' => 1, 'second' => 2]];
+        $this->validation->run((array) $data);
```
paulbalandan, line 641: Why do you need to cast this first to object then recast to array? Casting is not recursive. https://3v4l.org/gMad5
```suggestion
        $data = ['configuration' => (object) ['first' => 1, 'second' => 2]];
        $this->validation->run($data);
```

tests/system/Validation/ValidationTest.php
```diff
@@ -628,6 +628,33 @@ public function testTagReplacement()
...
+        $this->assertSame([], $this->validation->getErrors());
+
+        $this->validation->reset();
+
+        $this->validation->setRules([
+            'configuration' => 'required|check_object_rule',
+        ]);
+
+        $data = (object) ['configuration' => (object) ['first1' => 1, 'second' => 2]];
+        $this->validation->run((array) $data);
```
paulbalandan, line 651: Same comment as above

tests/system/Validation/ValidationTest.php
daycry: If you look at the method, it only accepts arrays.
I Know that isn't recursive. Configuration attribute is a Object.

daycry: The change that I propose can be used to validate that what an object has are the mandatory fields, for example:

_**public function check_token_configuration( Object $value, string $socialNetwork, array $data = [] )
    {
        $doctrine = \Config\Services::doctrine();

        if( !isset( $data[ $socialNetwork ] ) ) { return false; }

        $configSocialConnection = config( \ucfirst( $data[ $socialNetwork ] ) . '\\Connection' );
        
        foreach( \get_object_vars( $configSocialConnection ) as $key => $val )
        {
            if( !isset( $value->{ $key } ) )
            {
                return false;
            }

            if( empty( $value->{ $key } ) ){ return false; }

            $configSocialConnection->{ $key } = $value->{ $key };
        }

        return true;
    }**_


Without the change that I propose it does not work

tests/system/Validation/ValidationTest.php
daycry: image

daycry: I convert to array because the first param must be an array or null

daycry: I try to simulate an api rest petition.

image

tests/system/Validation/ValidationTest.php
paulbalandan: Hmm. Fair enough. Thanks for the explanation.

daycry:
If you try to validate an object and it fails, the application fails
because it only accepts strings and arrays, it does not do a
json_encode if it is an object, because when you work with rest, you
may have to validate objects.

paulbalandan (approved):

Issue #4026, opened 2020-12-24
Bug: can't get environment variable with dot


kenjis:
**Describe the bug**
We can't get environment variable that name has dot on some environment.
So we can't change Config values.

docker-compose.yml:
```
services:
  web:
    ...
    environment:
      - database.default.port=${DB_PORT:-3306}
      - database_default_port=${DB_PORT:-3306}
```

httpd.conf:
```
    PassEnv database.default.port
    PassEnv database_default_port
```

controller code:
```
dd(getenv('database.default.port')); // => false 
```

```
dd(getenv('database_default_port')); // => string 3306
```

**CodeIgniter 4 version**
4.0.4

**Affected module(s)**
Out of CodeIgniter?

**Expected behavior, and steps to reproduce if appropriate**
Environment variable name with dot is difficult to handle. It is better to support variable name without dot.

**Context**
 - OS: [Debian 10.7] Docker php:7.4-apache-buster
 - Web server [Apache 2.4.38-3+deb10u4]
 - PHP version [7.4.13]

michalsn:
It's a Docker thing (it can't handle the dots) and there is nothing we can do about it. You can use `__construct()` in your config class to override the variables you want.

Changing dots to underscores on a framework level won't happen at this stage - besides this is an "issue" only for Docker and there is a workaround for it anyway.

I'm closing this, but feel free to continue the conversation - especially if you have ideas to solve it better. Thanks.

kenjis:
@michalsn It is not a Docker thing. Because I can get the variable in the bash and Perl.

```
$ docker-compose exec web bash
# perl -e 'print $ENV{"database.default.port"}."\n"'
3306
```

But I can't get the variable with PHP. Is it a PHP thing?

```
# php -r 'echo $_ENV["database.default.port"];'
[02-Jan-2021 11:36:05 UTC] PHP Notice:  Undefined index: database.default.port in Command line code on line 1

Notice: Undefined index: database.default.port in Command line code on line 1
```

And it seems heroku also can't handle environment variables with dot. 
Is there any environment that can handle environment variables with dot?

kenjis:
It is not a PHP thing.

```
# php -r 'echo getenv("database.default.port").PHP_EOL;'
3306
```

Is it an Apache thing?

michalsn:
[...]

PR #5156, opened 2021-10-01
feat: `_` can be used as separators in environment variable names

kenjis:
**Description**
You can use both as environment variable name:
- `database.default.hostname` (now) - looking for it first
- `database_default_hostname` (new) - if not found, looking for it

See #4026

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Config/fixtures/.env
+++ b/tests/system/Config/fixtures/.env
@@ -6,15 +6,18 @@ NULL=
 
 SimpleConfig.onedeep=baz
 SimpleConfig.default.name=ci4
-SimpleConfig.simple.name=complex
+# Use underscore as separator
+SimpleConfig_simple_name=complex
 
-# for environment override testing
+## for environment override testing
 SimpleConfig.alpha=pow
 SimpleConfig.charliewrong=null
-SimpleConfig.notthere=missing
+# Use underscore as separator
+SimpleConfig_notthere=missing
 simpleconfig.delta=hubbahubba
 simpleconfig.foxtrot="true"
-simpleconfig.dessert="banana"
+# Use underscore as separator
+simpleconfig_dessert="banana"
 
 bravo=kazaam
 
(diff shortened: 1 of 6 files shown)


MGatner:
The code looks good. Could you share your motivation behind this change? Are there other frameworks that use underscores? Or perhaps deployment solutions? Or do you have an underscore fetish? 😉

samsonasik:
never use `_` usage, what if user has existing `_` config as a part of value? 

```php
SimpleConfig.onedeep_value=baz
```

as `onedeep_value` as the key itself. is that previously working or after this change, that will still works?

kenjis:
@MGatner Some environments can't use environment variables with dot.
See #4026

And I saw this tweet:
https://twitter.com/eclairtom/status/1443603184597962752
English Translation:
> Versioned up to #Docker desctop 4.1.0. docker-compose is no longer usable.
The .env used by CodeIgniter4 is now affected. Because it uses constants that contain dots. (4.0 was ignored.)
I have to change the folder structure for docker at worst.

kenjis:
@samsonasik Good question!
I think property name like `onedeep_value` is not used by CI4,
but users might use it.

I added a test case, and it passed.

tests/system/Config/fixtures/.env
```diff
@@ -6,15 +6,20 @@ NULL=
 
 SimpleConfig.onedeep=baz
 SimpleConfig.default.name=ci4
-SimpleConfig.simple.name=complex
+# Use underscore as separator
+SimpleConfig_simple_name=complex
+# Properity name with undersocre
+SimpleConfig.onedeep_value=bar
```
samsonasik, line 12: I think another case is needed, eg

```
SimpleConfig.one_deep.under_deep
```

for user use "." but use underscored keys

kenjis: I added a test.

michalsn (approved):
I think this is probably the right direction to solve the problem with dot variables for some environments.

The only thing missing is documentation. We don't want this to become a secret feature, so it needs to be documented in the user guide.

kenjis:
@michalsn Thanks! I added it to the documentation.

MGatner (approved):
Nice teamwork! This looks ready to me.

PR #5170, opened 2021-10-03
Fixed a problem with not run escape for identities in like when `insensitiveSearch` is true.

ytetsuro:
Fixed a problem with not run escape for identities in like when `insensitiveSearch` is true.

**Description**

Items that can be escaped are not escaped.
Even if the escape flag is true or null, there is no escaping process in place.
This has been fixed so that escaping will be done.

link: #2487

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1003,7 +1003,7 @@ class BaseBuilder
                 $bind = $this->setBind($k, "%{$v}%", $escape);
             }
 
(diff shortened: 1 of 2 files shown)


kenjis (approved):
LGTM

paulbalandan (approved):

kenjis:
Sorry, this PR introduced the bug #5360.

PR #5172, opened 2021-10-04
Fix debug toolbar db connection count

### LINKED ISSUE #5160: Dev: Debug Toolbar should show correct count of connections
Screen Shot 2021-09-30 at 10 39 18 PM

Currently, the Debug Toolbar displays `0 Connections`, even though I know it just used a connection. This should show the proper connection count.

danielTiringer:
Fixes #5160 

**Description**
The debug toolbar's database tab showed X queries 0 connections - the count of used connections was not reported.
The change counts the unique (name) connections while iterating through the queries, and saves them in a new static variable. Then this variable is used to display the count.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar/Collectors/Database.php
+++ b/system/Debug/Toolbar/Collectors/Database.php
@@ -62,19 +62,11 @@ class Database extends BaseCollector
     protected static $queries = [];
 
     /**
-     * Array of connections used in a collected set
-     * of queries.
-     *
-     * @var array
-     */
-    protected static $activeConnections = [];
-
-    /**
      * Constructor
      */
     public function __construct()
     {
-        $this->connections = \Config\Database::getConnections();
+        $this->getConnections();
     }
 
     /**
@@ -91,12 +83,6 @@ class Database extends BaseCollector
         $max = $config->maxQueries ?: 100;
 
         if (count(static::$queries) < $max) {
-            $connection = $query->db->getDatabase();
-
-            if (! in_array($connection, self::$activeConnections, true)) {
-                self::$activeConnections[] = $connection;
-            }
-
             static::$queries[] = $query;
         }
     }
@@ -162,8 +148,10 @@ class Database extends BaseCollector
      */
     public function getTitleDetails(): string
     {
+        $this->getConnections();
+
         $queryCount      = count(static::$queries);
-        $connectionCount = count(static::$activeConnections);
+        $connectionCount = count($this->connections);
 
         return sprintf(
             '(%d Quer%s across %d Connection%s)',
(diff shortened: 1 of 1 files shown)


system/Debug/Toolbar/Collectors/Database.php
```diff
@@ -148,7 +162,8 @@ public function getBadgeValue(): int
      */
     public function getTitleDetails(): string
     {
-        return '(' . count(static::$queries) . ' Queries across ' . ($countConnection = count($this->connections)) . ' Connection' .
+        return '(' . count(static::$queries) . ' Queries across ' .
+                ($countConnection = count(self::$activeConnections)) . ' Connection' .
                 ($countConnection > 1 ? 's' : '') . ')';
```
paulbalandan, line 167:
```suggestion
        $connectionCount = count(static::$activeConnections);

        return sprintf(
            '(%d Queries across %d Connection%s)',
            count(static::$queries),
            $connectionCount,
            $connectionCount > 1 ? 's' : ''
        );
```

system/Debug/Toolbar/Collectors/Database.php
```diff
@@ -148,8 +162,16 @@ public function getBadgeValue(): int
      */
     public function getTitleDetails(): string
     {
-        return '(' . count(static::$queries) . ' Queries across ' . ($countConnection = count($this->connections)) . ' Connection' .
-                ($countConnection > 1 ? 's' : '') . ')';
+        $queryCount      = count(static::$queries);
+        $connectionCount = count(static::$activeConnections);
+
+        return sprintf(
+            '(%d Quer%s across %d Connection%s)',
```
paulbalandan, line 169:
```suggestion
            '(%d Quer% across %d Connection%s)',
```

lonnieezell (commented):
It looks like it's already trying to count the connections that it finds in the constructor. Have you verified what is going wrong there?

danielTiringer:
> It looks like it's already trying to count the connections that it finds in the constructor. Have you verified what is going wrong there?

It seems to me that the toolbar's Database collector is created before a database connection is made. Its constructor gets the connections from the config, but at that time, nothing's connected and an empty array is returned.
If we do the `getConnections()` call at the start of the `getTitleDetails()` function, the connection is returned as expected.

That can be an alternative too, refreshing the connection array.

lonnieezell:
Ah - good point. I think doing a `getConnections()` at the start of `getTitleDetails()` as you've mentioned works well. That doesn't have us keeping duplicate info around.

danielTiringer:
> Ah - good point. I think doing a `getConnections()` at the start of `getTitleDetails()` as you've mentioned works well. That doesn't have us keeping duplicate info around.

Alright, changed the PR accordingly.

paulbalandan (approved):

lonnieezell:
Looks good. Thanks!

PR #5176, opened 2021-10-05
Change behavior of `class_attributes_separation` rule

paulbalandan:
**Description**
This PR proposes to change how we use the `class_attributes_separation` rule. Currently, we only activate changing the separation between class `method`s to always have a single blank line. This now includes class constants, properties, and trait uses in the picture by using this logic:

1. If the const, property, or trait use has metadata (PHPDoc or Attribute) attached to it, it gets one line.
2. If none, then no blank lines are enforced.
3. If preceding class attribute has metadata while the current has none, there will still be a blank line.
```diff
class Foo
{
    /** @var Foo */
    public $bar;
+
    public $baz;
-
    public $qux;
}
```
4. There are blank lines in between attributes:
```diff
class Foo
{
    use FooTrait;
    use BarTrait;
+
    public $bar;
    public $baz;
}
```

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/.no-header.php-cs-fixer.dist.php
+++ b/.no-header.php-cs-fixer.dist.php
@@ -30,14 +30,22 @@ $finder = Finder::create()
 $overrides = [
     // @TODO Remove once these are live in coding-standard
     'assign_null_coalescing_to_coalesce_equal' => false, // requires 7.4+
-    'control_structure_continuation_position'  => ['position' => 'same_line'],
-    'empty_loop_condition'                     => ['style' => 'while'],
-    'integer_literal_case'                     => true,
-    'modernize_strpos'                         => false, // requires 8.0+
-    'no_alternative_syntax'                    => ['fix_non_monolithic_code' => false],
-    'no_space_around_double_colon'             => true,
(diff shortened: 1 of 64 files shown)


paulbalandan:
I tend to agree it may lose the intention. However, some of the spacings are arbitrary, if not useless. I would love to keep the spacing if the intention for doing so is clearly evident in the code (by way of regular comments, perhaps) -- for example, the groupings made in https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Commands/Generators/Views/model.tpl.php is clear to the intention.
```php
    // Dates
    protected $useTimestamps = false;
    protected $dateFormat    = 'datetime';
    protected $createdField  = 'created_at';
    protected $updatedField  = 'updated_at';
    protected $deletedField  = 'deleted_at';

    // Validation
    protected $validationRules      = [];
    protected $validationMessages   = [];
    protected $skipValidation       = false;
    protected $cleanValidationRules = true;
```

Otherwise, I'd argue those blank lines adds no value.

PR #5178, opened 2021-10-05
Extract method to get prefix for DB access function

ytetsuro:
The process of getting the prefix of the db access function was extracted to the method.

**Description**

It is incorrect that this is in the parent class, BaseConnection.
This is because the parent class will behave as if it knows the implementation of the child class.
With this code, there will be a lack of coverage when trying to add a new driver as shown below.

https://github.com/codeigniter4/CodeIgniter4/pull/2487#discussion_r720942310

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1269,8 +1269,7 @@ abstract class BaseConnection implements ConnectionInterface
      */
     public function callFunction(string $functionName, ...$params): bool
     {
-        $driver = strtolower($this->DBDriver);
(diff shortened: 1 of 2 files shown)


paulbalandan (approved):
I agree. BaseConnection and other Base** classes should be agnostic as possible of each driver's implementation.

ytetsuro:
@paulbalandan 

Thanks, review.

PR #5181, opened 2021-10-06
Enable `no_alternative_syntax` rule

paulbalandan:
**Description**
```console
$ vendor/bin/php-cs-fixer describe no_alternative_syntax
Description of no_alternative_syntax rule.
Replace control structure alternative syntax to use braces.

Fixer is configurable using following option:
* fix_non_monolithic_code (bool): whether to also fix code with inline HTML; defaults to true

Fixing examples:
   --- Original
   +++ New
   @@ -1,2 +1,2 @@
    <?php
   -if(true):echo 't';else:echo 'f';endif;
   +if(true) { echo 't';} else { echo 'f';}

 * Example #2. Fixing with configuration: ['fix_non_monolithic_code' => true].
   --- Original
   +++ New
   @@ -1,3 +1,3 @@
   -<?php if ($condition): ?>
   +<?php if ($condition) { ?>
    Lorem ipsum.
   -<?php endif; ?>
   +<?php } ?>

```

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/.no-header.php-cs-fixer.dist.php
+++ b/.no-header.php-cs-fixer.dist.php
@@ -27,7 +27,10 @@ $finder = Finder::create()
     ])
     ->notName('#Logger\.php$#');
 
-$overrides = [];
+$overrides = [
+    // @TODO Remove once live in coding-standard
+    'no_alternative_syntax' => ['fix_non_monolithic_code' => false],
+];
 
 $options = [
     'cacheFile'    => 'build/.no-header.php-cs-fixer.cache',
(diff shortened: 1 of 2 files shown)


lonnieezell (approved):

paulbalandan:
Yes, I'm aware of that. That's why `fix_non_monolithic_code` option is set to `false`.

kenjis:
@paulbalandan What is `fix_non_monolithic_code`?

paulbalandan:
Heredoc/nowdoc are already supported throughout. It is just that there is no rule yet that autoconverts strings to heredoc/nowdoc, if that's what you mean.

paulbalandan:
In essence, this rule as configured will convert alternative syntax uses in pure PHP code only.

Issue #5161, opened 2021-10-01
Dev: Debug Toolbar should list the number of duplicate/unique queries


lonnieezell:
At the top of the `Database` section of the Debug Toolbar, it should list the number of unique and duplicate queries. 

It would be nice to change the background of the query rows that are duplicates, or use a small icon, so it can be easily noticed.

danielTiringer:
I'd like to take a look at this issue.

lonnieezell:
Thanks! I've assigned it to you for now.

PR #5185, opened 2021-10-07
Feature - Mark duplicate queries

danielTiringer:
The intent of this pull request is to address #5161 - providing a number of unique queries in the summary, and mark the duplicates with an icon or a different background.

The background part is WIP.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide


lonnieezell:
@danielTiringer Originally, I was thinking the background for the query would be slightly different, but I think this is a nice simple solution. If we find it doesn't stand out enough down the road, we can always modify how it's showing.

PR #5192, opened 2021-10-10
Feature - Add query strings to debug timeline

### LINKED ISSUE #5159: Dev: Debug Toolbar should display queries on the timeline
Currently, the Debug Toolbar will list Query on the timeline and display the time it took. However, there's no way to tell how what query it was. 

I would like us to show, on the row just below where `Query` is listed, the query itself. This like should be hidden by default, but should be expandable by clicking on the row where `Query` is listed. 

We should use the same type of highlighting that is already happening on the Queries page. 

Screen Shot 2021-09-30 at 10 35 34 PM

danielTiringer:
Fixes #5159 

**Description**
- [X] Pass the formatted query string to the timeline collector
- [X] Display the query string on the timeline view
- [X] Make the query string appear / disappear upon clicking on the word `Query`

**View**

| Closed |  |
|--------------|-----------|
|  One open |  |  
|  Two open |  |  

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -196,8 +196,7 @@ class Toolbar
 
             $output .= '<td class="' . ($isChild ? 'debug-bar-width30' : '') . '" style="--level: ' . $level . ';">' . ($hasChildren ? '<nav></nav>' : '') . $row['name'];
             if (isset($row['query'])) {
-                $output .= '<br>';
-                $output .= '<inline class="debug-bar-ndisplay">' . $row['query'] . '</inline>';
+                $output .= '<span class="debug-bar-ndisplay">' . $row['query'] . '</span>';
             }
             $output .= '</td>';
             $output .= '<td class="' . ($isChild ? 'debug-bar-width10' : '') . '">' . $row['component'] . '</td>';


sfadschm:
Can you maybe add screenshots for the default display and the display after a query was clicked? This proved to be a good tool to discuss changes to the toolbar in the past 😃

sfadschm:
Thanks for the screenshots! Looks good.
However, we talked past each other with the `inline`.
I'm perfectly fine with using `span`. Yet, the default CSS `display` value for a `span` is `inline`. The JS you added then toggles between `block` and `none`, which would be inconsistent, as the default style (`display: inline`) would not be restored after toggling the display two times (`display: block`).

Actually, I think we could go with a `span style="display: block;"` and even spare the `<br>` this way. I will clone the branch tomorrow and check.

sfadschm:
I will need to test a little more tomorrow morning, but the current behavior interferes with the collapsible functionality of the timeline. I think it might be the cleanest solution to just add the query string as a 'child' of the query row, such that we can just use the current funtionality with it. Gonna try if that works.

PR #5200, opened 2021-10-12
[Debug] Improve keyword highlighting and escaping of query strings. 

sfadschm:
**Description**
~This PR adds a few more composite query keywords to the highlight list for the debug toolbar.~
~I also decided to sort the keywords alphabetically, as it was a pain checking which ones are missing. (This is debatable of course.)~

- Deconstruct composite keywords.
- Sort kewords alphabetically.
- Modify regex pattern to ignore kewords within escaped string values (`''`).
- [WIP] Add HTML escaping.

Depends on #5196.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

**Before:**

**After:**

### DIFF
--- a/system/Database/Query.php
+++ b/system/Database/Query.php
@@ -407,13 +407,13 @@ class Query implements QueryInterface
             $this->compileBinds(); // @codeCoverageIgnore
         }
 
-        $sql = htmlspecialchars($this->finalQueryString, ENT_COMPAT);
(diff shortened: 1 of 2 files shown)


sfadschm:
> 
> 
> The escaping way looks good to me.

The `esc()` or `htmlspecialchars()` way?

PR #5202, opened 2021-10-13
Reduce memory usage of insertBatch(), updateBatch()

### LINKED ISSUE #4099: Bug: Using insertBatch with a large amount of data will result in an error.
Using insertBatch with a large amount of data will result in an error.

```log
CRITICAL - 2021-01-13 21:23:37 --> Allowed memory size of 1073741824 bytes exhausted (tried to allocate 20480 bytes)
```

The data is about 150,000.
I also increased the memory limit, but it doesn't change.

```php
set_time_limit(120);
// ini_set('memory_limit', '256M');
// ini_set('memory_limit', '512M');
ini_set('memory_limit', '1024M');

$db->table('address')->truncate();
$db->table('address')->insertBatch($data);
```

It does not change even if "Size" is specified.
```php
$db->table('address')->insertBatch($chunk, null, 50);
```

It has a solid track record in CI3 as it is just a rewrite of what was originally used in CI3 into CI4.
I think it's a bug.

It seems that it can be avoided by using `array_chunk`.
```php
foreach (array_chunk($rows, 50) as $chunk)
{
  $db->table('address')->insertBatch($chunk);
}
```

**Context**
 - CodeIgniter version 4.0.4, develop
 - PHP version 7.4

kenjis:
**Description**
Fixes #4099

- fix return type of `insertBatch()` to be the same as `updateBatch()`
- reduce memory usage of `insertBatch()` and `updateBatch()`
  - do not use binds
  - If `$set` is not null, `setInsertBatch()`/`setUpdateBatch()` will be called during each batch process
 
See [Benchmark Result](https://github.com/codeigniter4/CodeIgniter4/pull/5202#issuecomment-944037647).

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


kenjis:
Should we clear `$binds` and `$bindsKeyCount` in `resetWrite()` and `resetSelect()`?

kenjis:
The sample app that uploads zipped CSV file, unzips it, and imports about 150,000 records 20MB of data to MySQL:
https://github.com/tomomo/ci4-data-import-example
I ran the app on macOS, not in Docker container.

**Before**
It seems the current BaseBuilder needs **2.4GB** for PHP (I was watching the memory in Activity Monitor on macOS).
And probably the import would not finished in **13 hours**.

Here is log of progress:
```
DEBUG - 2021-10-15 01:09:01 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 80
DEBUG - 2021-10-15 01:09:28 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 160
DEBUG - 2021-10-15 01:09:56 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 240
DEBUG - 2021-10-15 01:10:22 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 320
DEBUG - 2021-10-15 01:10:48 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 400
DEBUG - 2021-10-15 01:11:13 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 480
DEBUG - 2021-10-15 01:11:39 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 560
DEBUG - 2021-10-15 01:12:05 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 640
DEBUG - 2021-10-15 01:12:30 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 720
DEBUG - 2021-10-15 01:12:56 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 800
DEBUG - 2021-10-15 01:13:21 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 880
DEBUG - 2021-10-15 01:13:46 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 960
DEBUG - 2021-10-15 01:14:11 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 1040
DEBUG - 2021-10-15 01:14:37 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 1120
DEBUG - 2021-10-15 01:15:01 --> CodeIgniter\Database\BaseBuilder::insertBatch $affectedRows: 1200
```

(1200 - 80) / 6 mins = 187 records/min
150,000 / 187 = 802 mins = 13.4 hours.

**After** (This PR)
Peak memory: **321 MB**
Elapsed time: **66.3 seconds**

system/Database/BaseBuilder.php
```diff
@@ -1588,38 +1588,49 @@ public function insertBatch(?array $set = null, ?bool $escape = null, int $batch
...
+                $QBSet = array_slice($this->QBSet, $i, $batchSize);
+            } else {
+                $this->setInsertBatch(array_slice($set, $i, $batchSize), '', $escape);
+                $QBSet = $this->QBSet;
+            }
+            $sql = $this->_insertBatch($this->db->protectIdentifiers($table, true, null, false), $this->QBKeys, $QBSet);
 
             if ($this->testMode) {
-                $affectedRows++;
+                $savedSQL[] = $sql;
```
michalsn, line 1620: Some tests may start failing because of that. It will be an easy fix for any developer, but we have to mention this in the changelog and upgrade notes.

tests/system/Database/Builder/InsertTest.php
```diff
@@ -91,9 +91,6 @@ public function testInsertBatch()
         $query = $this->db->getLastQuery();
         $this->assertInstanceOf(Query::class, $query);
 
-        $raw = 'INSERT INTO "jobs" ("description", "id", "name") VALUES (:description:,:id:,:name:), (:description.1:,:id.1:,:name.1:)';
-        $this->assertSame($raw, str_replace("\n", ' ', $query->getOriginalQuery()));
```
michalsn, line 95: I guess now `getOriginalQuery` has no data or the result has changed? And that's the reason we had to remove this part of the test? If so, we have to consider this as a BC break change. I understand the goal here - reduce memory consumption and add speed bust - and that's why I'm not against it. But we have to make it clear in the changelog and upgrade notes.

michalsn (commented):
I don't have time for a full review today, so I'll leave a comment on the things that first caught my eye

system/Database/BaseBuilder.php
```diff
@@ -1950,28 +1961,37 @@ public function updateBatch(?array $set = null, ?string $index = null, int $batc
...
+        } elseif (empty($set)) {
+            if (CI_DEBUG) {
+                throw new DatabaseException('updateBatch() called with no data');
             }
 
-            $this->setUpdateBatch($set, $index);
+            return false; // @codeCoverageIgnore
         }
 
+        $hasQBSet = ($set === null);
```
samsonasik, line 1972: `()` i snot needed:

```suggestion
        $hasQBSet = $set === null;
```

system/Database/BaseBuilder.php
```diff
@@ -1588,38 +1588,49 @@ public function insertBatch(?array $set = null, ?bool $escape = null, int $batch
...
+        } elseif (empty($set)) {
+            if (CI_DEBUG) {
+                throw new DatabaseException('insertBatch() called with no data');
             }
 
-            $this->setInsertBatch($set, '', $escape);
+            return false; // @codeCoverageIgnore
         }
 
+        $hasQBSet = ($set === null);
```
samsonasik, line 1599: `()` is not needed as well:

```suggestion
        $hasQBSet = $set === null;
```

tests/system/Database/Builder/InsertTest.php
kenjis: Yes, `getOriginalQuery()` returns `INSERT INTO "jobs" ("description", "id", "name") VALUES ('There''s something in your teeth',2,'Commedian'), ('I am yellow',3,'Cab Driver')`. Because we do not use binds.

system/Database/BaseBuilder.php
kenjis: Added.

tests/system/Database/Builder/InsertTest.php
kenjis: Added docs.

user_guide_src/source/changelogs/v4.1.5.rst
```diff
@@ -14,6 +14,8 @@ BREAKING
 
 - Fixed `a bug <https://github.com/codeigniter4/CodeIgniter4/issues/2913>`_ on CSRF protection. Now CSRF protection works on PUT/PATCH/DELETE requests when CSRF filter is applied. If you use such requests, you need to send CSRF token.
 - In the previous version, if you didn't provide your own headers, ``CURLRequest`` would send the request-headers from the browser, due to a bug. As of this version, it does not send them.
+- Fixed ``BaseBuilder::insertBatch()`` return value. Now it returns SQL string array instead of wrong affected row number when testMode.
+- Because of the optimization, when ``BaseBuilder::insertBatch()`` and ``BaseBuilder::updateBatch()`` are used, the return value of ``$query->getOriginalQuery()`` changed.
```
michalsn, line 18:
```suggestion
- Major optimizations have been made to the way data is processed in ``BaseBuilder::insertBatch()`` and ``BaseBuilder::updateBatch()`` methods. This resulted in reduced memory usage and faster query processing. As a trade-off, the result generated by the ``$query->getOriginalQuery()`` method was changed. It no longer returns the query with the binded parameters, but the actual query that was run.
```
I think we should be a little more specific in terms of explaining the changes that we've made and that could potentially affect developers.

michalsn (commented):
I have one question about the return value for `testMode` and some suggestions for the docs.

system/Database/BaseBuilder.php
kenjis: As I wrote the commit message https://github.com/codeigniter4/CodeIgniter4/pull/5202/commits/0f3088c2786c5bc66f39b1857659ca99c47bcdd0 to make the return type of insertBatch() the same as updateBatch() .

> Now, the result for this method will be quite different from the result we get without the testMode.

I did not intend to change the behavior without the testMode. Am I something missing?

user_guide_src/source/changelogs/v4.1.5.rst
kenjis: Thanks! My English skill limits the explanation, and I'm not sure how detailed we write to changelog.

And I believe there are few developers to use `$query->getOriginalQuery()` for insertBatch/updateBatch.
So I didn't think it was necessary to go into too much detail.

michalsn (approved):
I have last two suggestions for the docs but everything looks good.

Regarding the accuracy of changelog descriptions - we're usually not super detailed, but when it comes to changing the result of some methods, we should always explain a little better what was changed and why. This will save time for developers, who won't have to search through the code to find out what exactly was changed.

kenjis:
@michalsn Thanks for your review and fixing the documentation!

This PR are making major design changes regarding **Batch** methods and I would like others to review this.

PR #5214, opened 2021-10-18
Fix FeatureTestCase does not work

### LINKED ISSUE #4722: Bug: Feature Tests not routing
**Describe the bug**
After updating to 4.1.2 from 4.1.1, my feature tests that utilize the FeatureTestCase are no longer routing correctly. Everything is routing to the default controller.

**CodeIgniter 4 version**
4.1.2

**Affected module(s)**
Testing and Routing

I have this test:
```php
    /** @test */
    public function invalidClientShouldReturn404()
    {
        $this->get('/client/10')
             ->assertStatus(404);
    }
```
With this in the routes:
```php
$routes->get('/client/(:num)', 'Clients::view/$1');
```

But the assertion is failing because it gets back a 200. Upon investigation I discovered that what I'm getting back is actually the view from the Home controller. This is what pretty much all of my feature tests are getting back. This test is extending the FeatureTestCase class which I know is deprecated but shouldn't it still work?

**Context**
 - OS: CentOS 8
 - Web server  Apache 2.4.37
 - PHP version 8.0.2

kenjis:
**Description**
Fixes #4722

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


kenjis:
Okay. I take down this.

PR #5218, opened 2021-10-19
Fix CURLRequest extra headers

### LINKED ISSUE #4826: Bug: Extra Headers on CURLRequest
**Describe the bug**
When you make a POST call to any controller on ServerA (RequestA) and then you use \Config\Services::curlrequest to make a "GET" request to a ServerB (RequestB), "MessageTrait.php" attach all the  RequestA Headers from the POST to the RequestB.  So the header "Content-Length" is sent to the ServerB.

**CodeIgniter 4 version**
CI 4.1.3

**Affected module(s)**
system/HTTP/CURLRequest.php
system/HTTP/MessageTrait.php

The header "Content-Length" from the RequestA is sent to the ServerB as a Header.
Then ServerB wait for the body of the RequestB but since RequesB is "GET" the body is <EMPTY> that means the server make a GatewayTimeout.

**Context**
 - PHP version 7.4.20

kenjis:
**Description**
Fixes #4826
- reset unneeded headers
- reset unneeded configs

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -129,6 +129,9 @@ class CURLRequest extends Request
 
         $this->send($method, $url);
 
+        // Reset unshared configs
+        unset($this->config['multipart'], $this->config['form_params']);
+
         return $this->response;
     }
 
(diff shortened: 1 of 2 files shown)


MGatner:
The code looks good but I would like it if someone else with more CURL experience took a look.

PR #5220, opened 2021-10-19
Fix CURLRequest keeping the last value provided by headers with the same name

### LINKED ISSUE #5041: Bug: CURLRequest keeps the last value provided by headers with the same name
It is not possible to get and forward all headers having multiple values of a response when performing a CURLRequest.
My experience was with 'set-cookie' header, but after investigating it was pretty obvious that the CURLRequest (and probably other involved or related classes) will only retain the last value of headers with the same name and in case of redirection trigger ErrorException: Header may not contain more than a single header, new line detected (SYSTEMPATH/HTTP/ResponseTrait.php at line 471).

The following fixed the issue for me:
----------------------------------------------------------------------------------------
System\HTTP\CURLRequest
```php
    protected function parseOptions(array $options)
    {
        if (array_key_exists('baseURI', $options)) {
            $this->baseURI = $this->baseURI->setURI($options['baseURI']);
            unset($options['baseURI']);
        }

                // my adjustment
                if($this->hasHeader($name)) {
                    $this->appendHeader($name, $value);
                } else {
                    $this->setHeader($name, $value);
                }
                // end my adjustment
            }

            unset($options['headers']);
        }
...
```

----------------------------------------------------------------------------------------
System\HTTP\ResponseTrait
```php
    public function sendHeaders()
    {
        // Have the headers already been sent?
        if ($this->pretend || headers_sent()) {
            return $this;
        }

        // Per spec, MUST be sent with each request, if possible.
        // http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html
        if (! isset($this->headers['Date']) && PHP_SAPI !== 'cli-server') {
            $this->setDate(DateTime::createFromFormat('U', (string) time()));
        }

        // HTTP Status
        header(sprintf('HTTP/%s %s %s', $this->getProtocolVersion(), $this->getStatusCode(), $this->getReason()), true, $this->getStatusCode());

        // Send all of our headers
        foreach (array_keys($this->getHeaders()) as $name) {
            // my adjustment
            $headerValue = $this->getHeader($name)->getValue();
            if(is_array($headerValue)) {
                foreach($headerValue as $h_value) {
                    header($name . ': ' . $h_value, false, $this->getStatusCode());
                }
            } else {
                header($name . ': ' . $this->getHeaderLine($name), false, $this->getStatusCode());
            }
        // end my adjustment
        }

        return $this;
    }
```

**CodeIgniter 4 version**
4.1.3

**Affected module(s)**
System\HTTP\CURLRequest
System\HTTP\ResponseTrait
probably some of the other classes as well 

1. Perform a CURLRequest
2. Obtain response with multiple values for i.e. 'set-cookie' header: expected - all values, obtained - the last value 
3. Redirect to endpoint while passing all 'set-cookie' headers: expected - redirection OK, obtained - error

**Context**
All

NicolaeIotu:
Fixes #5041 
CURLRequest can now get and forward all headers including multiple headers with the same name

**Description**
Basic minimal modifications to ensure multiple headers with the same name are correctly received and  forwarded.

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -236,7 +236,12 @@ class CURLRequest extends Request
 
         if (array_key_exists('headers', $options) && is_array($options['headers'])) {
             foreach ($options['headers'] as $name => $value) {
-                $this->setHeader($name, $value);
+                // fix issue #5041 by taking into account multiple headers with the same name
+                if($this->hasHeader($name)) {
+                    $this->appendHeader($name, $value);
+                } else {
+                    $this->setHeader($name, $value);
+                }
             }
 
             unset($options['headers']);
--- a/system/HTTP/ResponseTrait.php
+++ b/system/HTTP/ResponseTrait.php
@@ -468,7 +468,15 @@ trait ResponseTrait
 
         // Send all of our headers
         foreach (array_keys($this->getHeaders()) as $name) {
-            header($name . ': ' . $this->getHeaderLine($name), false, $this->getStatusCode());
+            // fix issue #5041 by taking into account multiple headers with the same name
+            $headerValue = $this->getHeader($name)->getValue();
+            if(is_array($headerValue)) {
+                foreach($headerValue as $h_value) {
+                    header($name . ': ' . $h_value, false, $this->getStatusCode());
+                }
+            } else {
+                header($name . ': ' . $this->getHeaderLine($name), false, $this->getStatusCode());
+            }
         }
 
         return $this;


MGatner:
Please address kenjis comments. I will also add that `setHeader()` already appends values for existing names, with the caveat that it is an array. **MessageTrait.php**:
```php
        if (isset($this->headers[$origName]) && is_array($this->headers[$origName]->getValue())) {
            if (! is_array($value)) {
                $value = [$value];
            }

            foreach ($value as $v) {
                $this->appendHeader($origName, $v);
            }
        } else {
            $this->headers[$origName]               = new Header($origName, $value);
            $this->headerMap[strtolower($origName)] = $origName;
        }
```
Can you give an example of where this is not working?

kenjis:
@MGatner A header having multiple values, multiple headers with the same name. They are different.
I think the following your comment is correct.

>  the framework does not support multiple headers with the same name, but should.

See https://github.com/codeigniter4/CodeIgniter4/issues/5041#issuecomment-1806610292
At least we cannot get all the cookie values with CURLRequest.

kenjis:
I sent PR #8194

NicolaeIotu:
Awesome! Much appreciated

PR #5258, opened 2021-10-29
Speed up `CommonSingleServiceTest`

paulbalandan:
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/CommonSingleServiceTest.php
+++ b/tests/system/CommonSingleServiceTest.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter;
 
+use CodeIgniter\Autoloader\FileLocator;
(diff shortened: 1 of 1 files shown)


MGatner (approved):
I assume these static variables don't fall under the newly-deprecated kind that we are removing for 8.1, right? Because the method is static?

kenjis:
@MGatner These static variables are not problem. The problem may occurs only when inherited.

> When a method using static variables is inherited (but not overridden), the inherited method will now share static variables with the parent method. 
https://www.php.net/manual/en/migration81.incompatible.php

kenjis (approved):

PR #5263, opened 2021-10-31
Replace usage of `FILTER_SANITIZE_STRING`

paulbalandan:
**Description**
Related to #4883 and follow-up to #5005

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CLIRequest.php
+++ b/system/HTTP/CLIRequest.php
@@ -15,8 +15,6 @@ use Config\App;
 use RuntimeException;
 
 /**
- * Class CLIRequest
- *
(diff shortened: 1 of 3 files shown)


kenjis (approved):
Strictly speaking, this is a breaking change.
See https://www.php.net/manual/en/filter.filters.sanitize.php#118186

And `FILTER_SANITIZE_STRING` is deprecated because it should not be used.
So I think this escaping process is originally a bug.

But I approve this to support 8.1 for the time being.

PR #5268, opened 2021-11-01
feat: add valid_url_strict rule

### LINKED ISSUE #3156: valid_url does not work correctly

kenjis:
**Description**
Supersedes #4658 
Fixes #3156

The current `valid_url` is too loose.

I referred to https://github.com/codeigniter4/CodeIgniter4/pull/4658#issuecomment-838524347

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -311,9 +311,8 @@ class FormatRules
             strtolower($validSchemes ?? 'http,https,mailto,tel,sms')
         );
 
-        return ! in_array($scheme, $validSchemes, true)
-            ? false
-            : filter_var($str, FILTER_VALIDATE_URL) !== false;
+        return in_array($scheme, $validSchemes, true)
+            && filter_var($str, FILTER_VALIDATE_URL) !== false;
     }
 
     /**
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -305,19 +305,15 @@ class FormatRules
             return false;
         }
 
-        if ($validSchemes === null) {
-            $validSchemes = 'http,https,mailto,tel,sms';
-        }
+        $scheme       = strtolower(parse_url($str, PHP_URL_SCHEME));
+        $validSchemes = explode(
+            ',',
+            strtolower($validSchemes ?? 'http,https,mailto,tel,sms')
+        );
 
-        $scheme           = strtolower(parse_url($str, PHP_URL_SCHEME));
-        $validSchemes     = strtolower($validSchemes);
-        $validSchemeArray = explode(',', $validSchemes);
-
-        if (! in_array(($scheme), $validSchemeArray, true)) {
-            return false;
-        }
-
-        return filter_var($str, FILTER_VALIDATE_URL) !== false;
+        return ! in_array($scheme, $validSchemes, true)
+            ? false
+            : filter_var($str, FILTER_VALIDATE_URL) !== false;
     }
 
     /**
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -308,7 +308,7 @@ class FormatRules
         $scheme       = strtolower(parse_url($str, PHP_URL_SCHEME));
         $validSchemes = explode(
             ',',
-            strtolower($validSchemes ?? 'http,https,mailto,tel,sms')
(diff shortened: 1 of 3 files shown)


system/Validation/FormatRules.php
```diff
@@ -291,6 +294,32 @@ public function valid_url(?string $str = null): bool
...
+
+        $scheme           = strtolower(parse_url($str, PHP_URL_SCHEME));
+        $validSchemes     = strtolower($validSchemes);
+        $validSchemeArray = explode(',', $validSchemes);
+
+        if (! in_array(($scheme), $validSchemeArray, true)) {
+            return false;
+        }
+
+        return filter_var($str, FILTER_VALIDATE_URL) !== false;
```
MGatner, line 320: Maybe something like this?
```suggestion
        $scheme       = strtolower(parse_url($str, PHP_URL_SCHEME));
        $validSchemes = explode(',', strtolower($validSchemes ?? 'http,https,mailto,tel,sms'));

        return in_array($scheme, $validSchemes, true)
                ?: filter_var($str, FILTER_VALIDATE_URL) !== false;
        }
```

MGatner (changes requested):
Thanks for implementing this! An old one. I made a code consolidation suggestion.

system/Validation/FormatRules.php
kenjis: A bit different. I pushed the refactoring.

paulbalandan: The return can be simplified further to use &&.

return in_array() && filter_var();

kenjis: Refactored.

michalsn (commented):
Oh... apparently I was the one who suggested this type of solution, but I'm not sure I thought it through properly.

I have concerns about the `mailto,tel,sms` options. I don't think they will be really helpful to most people and will give users the misconception that we are validating email or phone number here. I know this can be helpful in some cases, but in my opinion only in conjunction with other validation rules.

I would only use `https` as the default `validScheme`. I don't even know if browsers accept anything with `http` without making a big deal out of it these days.

We should indicate that the validation process follows `RFC2396`.

Overall, I'm just a little worried that if we leave it the way it is, users will start complaining that it doesn't work the way it should "again"... Am I overthinking this?

kenjis:
How about these?

- remove `mailto,tel,sms` from the default valid schemes
  - I can't imagine use cases for validating mailto, tel, and sms URLs.
- add using `FILTER_VALIDATE_URL` to the user guide

michalsn:
Yes, I agree that we should get rid of `mailto,tel,sms`. I was searching why we added these and it seems like we took it because laravel helper had it. Maybe there is some use case, but I don't see a reason to have these as allowed protocols by default.

I was thinking about it a little more and maybe we should go a step further and assume that the majority of users would like to validate a "normal" URL address - with a valid TLD name in it? I believe that would be the most common use case (we have a separate rule for validating IP addresses).

We could validate TLD based on a list e.g: https://github.com/umpirsky/tld-list. It would be an additional step in url validation. Thoughts?

kenjis:
> I was thinking about it a little more and maybe we should go a step further and assume that the majority of users would like to validate a "normal" URL address - with a valid TLD name in it? I believe that would be the most common use case (we have a separate rule for validating IP addresses).

I also think the majority of users would like to validate web addresses.

But if users really need to check the URL, they need to check not only TLD, but also domain name.
Is there much value in checking only the TLDs? I'm not sure.

If we check the hostname in a URL, there are a lot of tasks to do.
See https://docs.laminas.dev/laminas-validator/validators/hostname/

And even if we use laminas-validator, we can't check if  the URL really exists.

MGatner:
I think sticking with validating the "format" is the way to go. I'm not opposed to something larger, like verifying domains and whatnot, but I don't think that belongs in the Format Rules but perhaps an extension of URI class or a separate Validation rule in its own library.

MGatner (approved):
I'm good with this current version.

michalsn (approved):
Okay.

kenjis:
Thanks, guys!

PR #5269, opened 2021-11-01
[PHPStan] Upgrade to PHPStan 1.0

samsonasik:
Upgrading to PHPStan 1.0, Apply rector `dev-main` before next release.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -39,6 +39,19 @@ parameters:
 		- '#Method CodeIgniter\\Validation\\ValidationInterface::run\(\) invoked with 3 parameters, 0-2 required#'
 		- '#Negated boolean expression is always (true|false)#'
 		- '#Return type \(bool\) of method CodeIgniter\\HTTP\\Files\\UploadedFile::move\(\) should be compatible with return type \(CodeIgniter\\Files\\File\) of method CodeIgniter\\Files\\File::move\(\)#'
+		- '#.*\son left side of \?\? is not nullable#'
+		- '#While loop condition is always true#'
+		- '#Right side of && is always true#'
+		- '#.*in isset\(\) is not nullable#'
+		- '#.*in empty\(\) is not falsy#'
+		- '#.*on left side of \?\? always exists and is not nullable#'
+		- '#Variable \$error on left side of \?\? always exists and is always null#'
+		- '#Variable \$config in empty\(\) always exists and is not falsy#'
+		- '#If condition is always true#'
+		- '#Dead catch - CodeIgniter\\Exceptions\\PageNotFoundException is never thrown in the try block#'
+		- '#.* in isset\(\) does not exist#'
+		- '#Variable \$mockService in empty\(\) always exists and is always falsy#'
+		- '#PHPDoc type string\|null of property CodeIgniter\\Images\\Handlers\\ImageMagickHandler::\$resource is not covariant with PHPDoc type resource\|null of overridden property CodeIgniter\\Images\\Handlers\\BaseHandler::\$resource#'
 	parallel:
 		processTimeout: 300.0
 	scanDirectories:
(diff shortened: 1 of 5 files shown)


samsonasik:
ignoring errors for now so we can verify next one by one.

phpstan.neon.dist
```diff
@@ -39,6 +39,19 @@ parameters:
...
+		- '#.*in isset\(\) is not nullable#'
+		- '#.*in empty\(\) is not falsy#'
+		- '#.*on left side of \?\? always exists and is not nullable#'
+		- '#Variable \$error on left side of \?\? always exists and is always null#'
+		- '#Variable \$config in empty\(\) always exists and is not falsy#'
+		- '#If condition is always true#'
+		- '#Dead catch - CodeIgniter\\Exceptions\\PageNotFoundException is never thrown in the try block#'
+		- '#.* in isset\(\) does not exist#'
+		- '#Variable \$mockService in empty\(\) always exists and is always falsy#'
+		- '#PHPDoc type string\|null of property CodeIgniter\\Images\\Handlers\\ImageMagickHandler::\$resource is not covariant with PHPDoc type resource\|null of overridden property CodeIgniter\\Images\\Handlers\\BaseHandler::\$resource#'
```
MGatner, line 54: Most of these are very sweeping exclusions that I would think we want to catch in general. Are there so many violations in the code that we have to ignore them globally? That makes me uneasy.

MGatner (changes requested):
I expect all the Rector changes are due to dependency incompatibilities, but is there a way we can split this out into separate PRs?

phpstan.neon.dist
samsonasik: Rector is currently still uses PHPStan as dependency:

https://github.com/rectorphp/rector/blob/607db8843ec560d549836e9e88d40ba1faed4a2a/composer.json#L10

so updating PHPStan will require updating rector as well. The other solution is:

- when running PHPStan CI workflow, remove rector/rector dep first
- when running Rector CI workflow, remove PHPStan dep first

next, we verify one by one above errors, eg:

```php
class A
{
    /** @var array */
    public $config;
}

class B
{
    public function __construct(A $a)
    {
        $a->config ?? [];
    }
}
```

will show error "Property A::$config (array) on left side of ?? is not nullable.", It previously working OK on PHPStan 0.12.99.

We can change to:

```diff
-$a->config ?? [];
+$a->config ?: [];
```

but that may have different behaviour as it still rely on docblock.

MGatner: Thanks for explaining. I've become a fan of analysis as external tools, since they aren't really "dependencies" of a project in most cases. PHPStan is available via Phive (https://phar.io/#Tools) which would be an option, and Rector can be installed as `composer global` - the only catch would be deciding if/how to incorporate them into our admin git hooks.
***
For your example I would consider that a violation. if `A::$config` is not assigned during construction then it needs to be a nullable type or fail static analysis. I see your point though, that a lot of these violations may just be increased precision in PHPStan that we need to address individually.

samsonasik: ref https://phpstan.org/r/972a7adc-10ce-49c4-a304-63605655b96f for example

paulbalandan: I agree we should throw the new errors and errors in the future under the rug. I would be in favor of using the baseline feature so that the current errors will only be ignored, while new errors of the same type will be reported.

samsonasik: @paulbalandan I think update PHPStan level should be in separate PR as the error list are different with current 0.12.99.

@MGatner I updated to use latest rector 0.12.0 that support PHPStan 1.0

samsonasik:
@MGatner @paulbalandan All green 🎉

kenjis (approved):

MGatner (approved):

PR #5272, opened 2021-11-01
unsafe-inline

nowackipawel:
An allow-list for specific inline scripts using a cryptographic nonce (number used once). The server must generate a unique nonce value each time it transmits a policy. It is critical to provide an unguessable nonce, as bypassing a resource's policy is otherwise trivial. See unsafe inline script for an example. Specifying nonce makes a modern browser ignore 'unsafe-inline' which could still be set for older browsers without nonce support.

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/HTTP/ContentSecurityPolicy.php
+++ b/system/HTTP/ContentSecurityPolicy.php
@@ -165,6 +165,8 @@ class ContentSecurityPolicy
     protected $validSources = [
         'self',
         'none',
+        'unsafe-hashes',
+        'strict-dynamic',
         'unsafe-inline',
         'unsafe-eval',
     ];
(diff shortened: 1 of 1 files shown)


kenjis:
@nowackipawel Thank you for your PR.

Could you update to a meaningful title and add the description to explain what you have changed, and why?

And You have two failed checks for coding style. 
https://github.com/codeigniter4/CodeIgniter4/runs/4073452774?check_suite_focus=true
https://github.com/codeigniter4/CodeIgniter4/runs/4073452630?check_suite_focus=true
See 

We expects unit testing.

paulbalandan:

Issue #606, opened 2017-07-13
Too much escaping negative int  (int < 0)


nowackipawel:
When I'm trying to save entity with a property which is a negative number then there is too much escaping with final query:

![sth](https://user-images.githubusercontent.com/8414802/28169578-29f9c9be-6826-11e7-8d20-cccff13acc2c.png)

is:   \\-10   should be -10

lonnieezell:
@nowackipawel Please try it now. I believe it should be fixed.

nowackipawel:
Yeah it works now. I thought you won't be escaping integer number but anyway it work as should.

PR #5277, opened 2021-11-02
Fix db escape negative integers

### LINKED ISSUE #4973: Bug: Query builder escapes negative integers
**Describe the bug**
In #606, a bit of code was added that escapes negative values. However, I think negative values should not be escaped at all. For example, I would like to look back at negative dates using something like

```sql
WHERE date_pickup < DateAdd(month, ?, Convert(date, GetDate())
```

Due the conversion, this is translated to

```sql
WHERE date_pickup < DateAdd(month, '-6', Convert(date, GetDate())
```
Which results in an error, since we now pass a varchar to the function instead of an integer.
My question is: why are negative values escaped at all? The offending bit of code checks that the integer is negative before escaping it, and I'm not sure what the original commit intends to solve.

**CodeIgniter 4 version**
Tested in 4.1.3, should be present in all versions.

**Affected module(s)**
`Database/BaseConnection.php` lines 1312 - 1314 in the escape function. Specifically when using SQLSRV driver. Though it should be independent of driver used.

**Context**
 - OS: Windows 10
 - Web server IIS
 - PHP version 8.0

kenjis:
**Description**
Fixes #4973

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -565,7 +565,7 @@ abstract class BaseConnection implements ConnectionInterface
      *
      * @param mixed ...$binds
      *
-     * @return BaseResult|bool|Query
+     * @return BaseResult|bool|Query BaseResult when “read” type query, bool when “write” type query, Query when prepared query
      *
      * @todo BC set $queryClass default as null in 4.1
      */
@@ -955,6 +955,8 @@ abstract class BaseConnection implements ConnectionInterface
      * the correct identifiers.
      *
      * @param array|string $item
+     * @param bool         $prefixSingle Prefix an item with no segments?
+     * @param bool         $fieldExists  Supplied $item contains a field name?
      *
      * @return array|string
      */
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1202,10 +1202,6 @@ abstract class BaseConnection implements ConnectionInterface
             return ($str === false) ? 0 : 1;
         }
 
-        if (is_numeric($str) && $str < 0) {
-            return "{$str}";
-        }
-
         return $str ?? 'NULL';
     }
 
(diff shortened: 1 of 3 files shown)


system/Database/BaseConnection.php
```diff
@@ -1201,7 +1203,7 @@ public function escape($str)
         }
 
         if (is_numeric($str) && $str < 0) {
```
paulbalandan, line 1205: Why is there a need to add this for negative numbers? I think we can remove this if block entirely as the last return will do the job.

paulbalandan: And it seems inconsistent, with negative numbers we get a string. For positive, we can numbers.

kenjis: Good point!

For now, I remove it.

kenjis: However, I am not sure why this commit https://github.com/codeigniter4/CodeIgniter4/commit/19c173a49f35388d90736b6ee596f33a5f29a870 was fixed #606.

paulbalandan: let's have @lonnieezell shed the light on this. 😂

lonnieezell: I don't remember exactly. It's been 4 years :) . Looking back through the issue, though, I'm guessing that wrapping it in quotes would get it past the escaping and adding of the slash, which was then converted by MySQL into an INT or whatever type when it was saved to the table. 

I would say as long as we have tests in place that ensure his problem is still fixed, and it doesn't escape the negative sign, this is fine.

tests/system/Database/Live/EscapeTest.php
```diff
@@ -40,7 +40,7 @@ protected function setUp(): void
      */
     public function testEscapeProtectsNegativeNumbers()
```
paulbalandan, line 41: This should be renamed to say that this does not escape negative numbers now.

kenjis: Fixed.

system/Database/BaseConnection.php
kenjis: @lonnieezell Thanks!

paulbalandan (approved):

PR #5280, opened 2021-11-03
fix: module filters are not discovered when using route filters

### LINKED ISSUE #4720: Bug: Module route filters not loading
**Describe the bug**
If the module route has a filter, then the filter is checked in App\Config\Filters.
If the filter is not found, then an error is thrown. 
If a module filter must be registered in App\Config\Filters, then trying to load the module filters extension is meaningless. 

Example
```php
//Module/Namespace/Config/Routes
$routes->add('route', 'Controller::method', ['filter' => 'module_filter']);
//Module/Namespace/Config/Filters
$filters->aliases['module_filter'] = Module\Namespace\Filters\SomeFilter::class;
```

**CodeIgniter 4 version**
4.1.2

**Affected module(s)**
CodeIgniter, Filters

kenjis:
**Description**
Fixes #4720

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -363,9 +363,6 @@ class CodeIgniter
         // Start up the filters
         $filters = Services::filters();
 
-        // Discover filters
-        $filters->initialize(strtolower($uri));
-
         // If any filters were specified within the routes file,
         // we need to ensure it's active for the current request
         if ($routeFilter !== null) {
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -358,8 +358,13 @@ class CodeIgniter
     {
         $routeFilter = $this->tryToRouteIt($routes);
 
+        $uri = $this->determinePath();
+
         // Start up the filters
         $filters = Services::filters();
+
+        // Discover filters
+        $filters->initialize(strtolower($uri));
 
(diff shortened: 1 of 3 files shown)


iRedds:
Yes, this solution works.
But what if we move the search for filters to the class constructor?

kenjis:
@iRedds I moved it. Does it work?

iRedds:
@kenjis You asks me? How about tests? =)

kenjis:
@iRedds Yes, of course. You are the only one user for this functionality in this planet.
Most important. :-)

kenjis:
@michalsn https://github.com/codeigniter4/CodeIgniter4/pull/3459#discussion_r474690853

Is there something wrong with moving`$this->discoverFilters()` back to the constructor?

I think the Filters files in the namespace are determined at the time of the instance creation, 
so it seems better to execute it once in the constructor.

michalsn:
I think this is a good fix. 

Can we add a test for this to make sure it won't be moved around anymore?

kenjis:
@michalsn Thanks! I added a test.

michalsn (approved):
Looks good to me.

Issue #4572, opened 2021-04-15
Bug: Filter auto discover for subproject doesn't behave as expected


chez14:
**Describe the bug**
`Filter.php` from subproject is not loaded as expected. That causes the CI throws `
pnd_auth filter must have a matching alias defined.` exception while the Filter is correctly defined in `Filters.php` subproject.

By exploration, I found that when the exception triggered, the `Filter.php` is not loaded. But when the routing doesnt have `filter`, the `Filters.php` is correctly loaded.

Loaded Filters when exception thrown:
![image](https://user-images.githubusercontent.com/5555026/114801547-1963fa00-9dc6-11eb-95ec-f93daa2291bb.png)

Loaded Filters when no exception thrown:
![image](https://user-images.githubusercontent.com/5555026/114801629-47493e80-9dc6-11eb-8230-3adf5f71823a.png)

**CodeIgniter 4 version**
Based on `composer.lock` file on both ([the main app](https://github.com/chez14/ci4-routing-filter-bug/blob/master/main-app/composer.lock#L63-L64) and [the admin panel](https://github.com/chez14/ci4-routing-filter-bug/blob/master/extendables/composer.lock#L10-L11)) uses `v4.1.1`.

**Affected module(s)**
I think it is the `discoverFilters()` function from [/system/Filters/Filters.php](https://github.com/codeigniter4/CodeIgniter4/blob/v4.1.1/system/Filters/Filters.php#L123-L144).

**Expected behavior, and steps to reproduce if appropriate**
`Filters.php` on sub project is loaded before the routing starts working, so when accessing `/admin/panel` route will redirects to `/admin/auth/login`. 

Step to reproduce:
- Create sub-project with `composer create-project codeigniter4/appstarter extendables`
- Remove all unneccessary files like configuration and etc.
- Create main app project with `composer create-project codeigniter4/appstarter main-app`
- Add local repository to main app project by [modifying the composer](https://github.com/chez14/ci4-routing-filter-bug/blob/master/main-app/composer.json#L42-L47) and install it on the main app by executing
  ```
  composer install chez14/extendables:dev-master
  ```
- [Add filter "Auth.php"](https://github.com/chez14/ci4-routing-filter-bug/blob/master/extendables/app/Filters/Auth.php) to the subproject.
- Register the filter to **subproject**'s [`Filters.php`](https://github.com/chez14/ci4-routing-filter-bug/blob/master/extendables/app/Config/Filters.php#L23)
[...]

PR #5281, opened 2021-11-03
docs: improve modules.rst

kenjis:
**Description**
- add sample code for `Acme\Blog namespace` definition
- add about Filter file, related #4572

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide


MGatner (approved):
This is good and accurate. I do think this discovery should be reworked as it has some issues but for now updating the docs to match the code is the best we can do.

PR #5289, opened 2021-11-06
Add PhpStorm metadata file for service()

kenjis:
**Description**
- to autocomplete `service()`

Screenshot 2021-11-06 11 06 22

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide


MGatner:
How important is it for this to be in the same repo? We've talked about creating a "tools" or "devkit" library for supplemental content like this.

kenjis:
Good point. The file does not need to be in this repo.

MGatner:
I will launch a toolkit repo soon and have you direct this over there.

MGatner:
@kenjis This is safe to move to https://github.com/codeigniter4/devkit - even though we don't have a release someone could use a VCS entry to include the file.

BTW do you know of a way to accomplish this same information for PHPStan? I read up on [stub files](https://phpstan.org/user-guide/stub-files) but they don't seem to support "this parameter => this outcome" sort of notation that we would need. I am constantly adding inline comments for helper functions like `model()` and `service()` and it would be nice to be able to define those centrally.

kenjis:
@MGatner 
`Factories::model()` uses `__call()`, so probably PHPStan extensions are needed.
https://phpstan.org/developing-extensions/class-reflection-extensions#methods-class-reflection-extensions

But the documentation is too little. I don't get how to write them.

MGatner:
@kenjis I actually asked about this (https://github.com/phpstan/phpstan/discussions/6019) and think I have a path to a solution.

kenjis:
@MGatner How about this #5358 ?

PR #5295, opened 2021-11-08
Toolbar - Make it possible to turn off var data collection

najdanovicivan:
**Description**
When there are a lot of vars being passed to the view for example array of entities with lots of parameters in each entity, var data can get very big and will consume a lot of memory.  A new parameter in toolbar config is introduced here to that collection of var data can be disabled.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -310,7 +310,7 @@ class Toolbar
      */
     protected function collectVarData(): array
     {
-        if (! $this->config->collectVarData) {
+        if (! ($this->config->collectVarData ?? true)) {
             return [];
         }
 
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -310,6 +310,10 @@ class Toolbar
      */
     protected function collectVarData(): array
     {
+        if (! $this->config->collectVarData) {
+            return [];
+        }
+
         $data = [];
 
         foreach ($this->collectors as $collector) {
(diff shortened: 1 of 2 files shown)


sfadschm:
I like this.
Maybe a step to far, but could this also be connected to a session setting so we can overwrite the value for the current session on-the-fly, lets say, from a toggle button in the toolbar, maybe?

najdanovicivan:
Cookie might better for this one. But confif should still exist to be able to fully disable it especially in situations where memory is being used over the limit as it wont be possible to load the site to triger the switch

sfadschm:
> Cookie might better for this one. But confif should still exist to be able to fully disable it especially in situations where memory is being used over the limit as it wont be possible to load the site to triger the switch

Completely agree!

system/Debug/Toolbar.php
```diff
@@ -310,6 +310,10 @@ protected function structureTimelineData(array $elements): array
      */
     protected function collectVarData(): array
     {
+        if (! $this->config->collectVarData) {
```
lonnieezell, line 313: So that it's not a BC break, we have to assume the value doesn't exist in the config file and treat it as `true` in that case. 

```suggestion
        if (! ($this->config->collectVarData ?? true)) {
```

lonnieezell (changes requested):

lonnieezell (approved):

PR #5298, opened 2021-11-09
chore: add .github/release.yml

kenjis:
**Description**
I propose that labeling PRs and use GitHub auto-generated release note for generating `CHANGELOG.md`.

See https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/admin/RELEASE.md
+++ b/admin/RELEASE.md
@@ -12,7 +12,7 @@ To auto-generate changelog, each PR to be listed in changelog must have one of t
 - **new feature** ... PRs for new features
 - **refactor** ... PRs to refactor
 
-And a PRs that has the breaking changes must have the following label:
+And PRs that have the breaking changes must have the following label:
 - **breaking change** ... PRs that may break existing functionalities
 
 ## Preparation
--- a/admin/RELEASE.md
+++ b/admin/RELEASE.md
@@ -4,6 +4,17 @@
 > Updated for `4.1.2` on May 17, 2021.
 > -MGatner
 
+## Labeling PRs
+
+To auto-generate changelog, each PR to be listed in changelog must have one of the following [labels](https://github.com/codeigniter4/CodeIgniter4/labels):
+- **bug** ... PRs that fix bugs
+- **enhancement** ... PRs to improve existing functionalities
+- **new feature** ... PRs for new features
+- **refactor** ... PRs to refactor
+
+And a PRs that has the breaking changes must have the following label:
+- **breaking change** ... PRs that may break existing functionalities
+
 ## Preparation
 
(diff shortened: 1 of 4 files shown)


admin/RELEASE.md
```diff
@@ -4,6 +4,17 @@
...
+
+To auto-generate changelog, each PR to be listed in changelog must have one of the following [labels](https://github.com/codeigniter4/CodeIgniter4/labels):
+- **bug** ... PRs that fix bugs
+- **enhancement** ... PRs to improve existing functionalities
+- **new feature** ... PRs for new features
+- **refactor** ... PRs to refactor
+
+And a PRs that has the breaking changes must have the following label:
+- **breaking change** ... PRs that may break existing functionalities
+
```
MGatner, line 17: I like having this here for reference, but it should also be in https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md so submitters (hopefully) apply their own labels.

MGatner (changes requested):

admin/RELEASE.md
kenjis: Okay, I will add.

"Apply/dismiss labels" needs `Triage` role.
https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-roles-for-an-organization#permissions-for-each-role

sfadschm:
Never knew we could assign labels ourselves.

admin/RELEASE.md
kenjis: Added.

MGatner (approved):
Okay that was my mistake, I thought people could add labels when they created a PR. @paulbalandan with your recent work on the PR forms, do you know if there is a way to template this so certain PRs get created with tags?

Looking at the example gist, I don't like how it is all one big block - but I assume that is because we didn't have PRs tagged this last round, right?

Regardless we don't have another solution right now so we will try this one. Thanks for the suggestion!

kenjis:
> Looking at the example gist, I don't like how it is all one big block - but I assume that is because we didn't have PRs tagged this last round, right?

Right.

So I added the template `release.yml`.
See https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configuration

kenjis:
@MGatner How about giving `triage` role to some contributors?

> Triage: Recommended for contributors who need to proactively manage issues and pull requests without write access
https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-roles-for-an-organization

PR #5294, opened 2021-11-08
4.1.5 Ready code

MGatner:

PR #5301, opened 2021-11-10
chore: update .github/workflows/deploy-framework.yml

kenjis:
**Description**
- replace `github.repos.` with `github.rest.repos.`

See https://github.com/codeigniter4/CodeIgniter4/pull/5294#issuecomment-963919466

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide


MGatner (approved):
I tested the original by changing the repo names to my own account space and making releases there. Since this is only the final release step though I think we just try it on next release, because it only takes about 30 seconds to fix if t workflow fails.

PR #5302, opened 2021-11-10
fix: UploadedFile::move() may return incorrect value

kenjis:
Supersedes and closes #5290

**Description**
- if `move_uploaded_file()` returns `false` without warning, `move()` and `hasMoved()` returns `true`

> https://www.php.net/manual/en/function.move-uploaded-file.php#refsect1-function.move-uploaded-file-returnvalues
> Return Values ¶
>
> Returns true on success.
>
> **If from is not a valid upload file, then no action will occur, and move_uploaded_file() will return false.**
>
> If from is a valid upload file, but cannot be moved for some reason, no action will occur, and move_uploaded_file() will return false. **Additionally, a warning will be issued.**

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/Files/UploadedFile.php
+++ b/system/HTTP/Files/UploadedFile.php
@@ -141,7 +141,7 @@ class UploadedFile extends File implements UploadedFileInterface
         $destination = $overwrite ? $targetPath . $name : $this->getDestination($targetPath . $name);
 
         try {
-            move_uploaded_file($this->path, $destination);
+            $this->hasMoved = move_uploaded_file($this->path, $destination);
         } catch (Exception $e) {
             $error   = error_get_last();
             $message = isset($error['message']) ? strip_tags($error['message']) : '';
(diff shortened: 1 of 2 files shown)


system/HTTP/Files/UploadedFile.php
```diff
@@ -141,20 +141,25 @@ public function move(string $targetPath, ?string $name = null, bool $overwrite =
         $destination = $overwrite ? $targetPath . $name : $this->getDestination($targetPath . $name);
 
         try {
-            move_uploaded_file($this->path, $destination);
+            $this->hasMoved = move_uploaded_file($this->path, $destination);
         } catch (Exception $e) {
```
paulbalandan, line 145: How does `move_uploaded_file()` throws an Exception? From the PHP docs, the function throws a warning when a valid file cannot be moved for some reason.

kenjis: Warnings turn into Exceptions by the framework.
https://www.php.net/manual/en/function.set-error-handler.php

paulbalandan: Yeah, I missed that. 👍

system/HTTP/Files/UploadedFile.php
```diff
@@ -141,20 +141,25 @@ public function move(string $targetPath, ?string $name = null, bool $overwrite =
...
             $error   = error_get_last();
             $message = isset($error['message']) ? strip_tags($error['message']) : '';
```
paulbalandan, line 147:
```suggestion
            $message = strip_tags($error['message'] ?? '');
```

tests/system/HTTP/Files/FileMovingTest.php
```diff
@@ -287,6 +290,41 @@ public function testFailedMove()
...
+        if (! is_dir($destination)) {
+            mkdir($destination, 0777, true);
+        }
+
+        $file = $collection->getFile('userfile1');
+
+        // Set the mock's return value to false
+        move_uploaded_file('', '', false);
+
+        $this->expectException(HTTPException::class);
```
paulbalandan, line 324: Add the expected exception message.

kenjis: Added.

paulbalandan (approved):

PR #5311, opened 2021-11-11
Update DatabaseHandler.php (SQLite3 related)

abydahana:
Add the support for timestamp field type for SQLite3 that unsupported "now()" function.

**Description**
Add the statement to using "strftime" (replacement of now() function) when using SQLite3 as database handler for session


michalsn (commented):
Sorry, but no. You would need a lot more than that to add session support for SQLite3.

AFAIK advisory locking on SQLite3 is buggy and even the official documentation mentions it: https://www.sqlite.org/lockingv3.html.

abydahana:
@michalsn okay then. I will use FileHandler instead for the session handler when using SQLite3 as the database.

PR #5320, opened 2021-11-14
[Rector] add PHPUnitSetList::PHPUNIT_SPECIFIC_METHOD

kenjis:
**Description**
- to fix assertion method like [AssertCompareToSpecificMethodRector](https://github.com/rectorphp/rector-phpunit/blob/main/docs/rector_rules_overview.md#assertcomparetospecificmethodrector)

See https://github.com/rectorphp/rector-phpunit#rector-rules-for-phpunit

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -42,6 +42,7 @@ use Rector\Php70\Rector\FuncCall\RandomFunctionRector;
 use Rector\Php71\Rector\FuncCall\CountOnNullRector;
 use Rector\Php73\Rector\FuncCall\JsonThrowOnErrorRector;
 use Rector\Php73\Rector\FuncCall\StringifyStrNeedlesRector;
+use Rector\PHPUnit\Rector\MethodCall\AssertIssetToSpecificMethodRector;
 use Rector\PHPUnit\Set\PHPUnitSetList;
 use Rector\Set\ValueObject\LevelSetList;
 use Rector\Set\ValueObject\SetList;
@@ -111,6 +112,15 @@ return static function (ContainerConfigurator $containerConfigurator): void {
 
         // use mt_rand instead of random_int on purpose on non-cryptographically random
         RandomFunctionRector::class,
+
+        // $this->assertTrue(isset($bar['foo']))
+        // and $this->assertArrayHasKey('foo', $bar)
+        // or $this->assertObjectHasAttribute('foo', $bar);
+        // are not the same
+        AssertIssetToSpecificMethodRector::class => [
+            __DIR__ . '/tests/system/Entity/EntityTest.php',
+            __DIR__ . '/tests/system/Session/SessionTest.php',
+        ],
     ]);
 
     // auto import fully qualified class names
--- a/rector.php
+++ b/rector.php
@@ -42,6 +42,7 @@ use Rector\Php70\Rector\FuncCall\RandomFunctionRector;
 use Rector\Php71\Rector\FuncCall\CountOnNullRector;
 use Rector\Php73\Rector\FuncCall\JsonThrowOnErrorRector;
 use Rector\Php73\Rector\FuncCall\StringifyStrNeedlesRector;
+use Rector\PHPUnit\Rector\MethodCall\AssertFalseStrposToContainsRector;
 use Rector\PHPUnit\Rector\MethodCall\AssertIssetToSpecificMethodRector;
 use Rector\PHPUnit\Set\PHPUnitSetList;
 use Rector\Set\ValueObject\LevelSetList;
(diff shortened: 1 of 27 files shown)


samsonasik:
@kenjis could you run:

```
vendor/bin/rector
```

and apply the change?

kenjis:
@samsonasik I committed.

I also added `PHPUnitSetList::PHPUNIT_80`,
`PHPUnitSetList::PHPUNIT_90`,
`PHPUnitSetList::PHPUNIT_91`.

samsonasik:
@kenjis I think we can add setlist one by one, as it cause error : 

```bash
Error: ] Could not process "system/Cookie/CloneableCookieInterface.php" file,   
         due to:                                                                
         "Argument 1 passed to Rector\Core\Rector\AbstractRector::isObjectType()
         must implement interface PhpParser\Node, null given, called in         
         vendor/rector/rector/rules/TypeDeclaration/Rector/ClassMethod/AddParamT
         ypeDeclarationRector.php:83". On line: 315        
```

first, ensure `PHPUnitSetList::PHPUNIT_SPECIFIC_METHOD` working first, and we can add more set in separate PR.

If we have specific rule that failed, eg: `AddParamTypeDeclarationRector` above, we can skip first under Option::SKIP on specific file.

kenjis:
@samsonasik Okay, I will do one by one. Thanks!

tests/system/Entity/EntityTest.php
```diff
@@ -69,11 +69,11 @@ public function testIssetWorksLikeTraditionalIsset()
     {
         $entity = $this->getEntity();
 
-        $this->assertFalse(isset($entity->foo));
+        $this->assertObjectNotHasAttribute('foo', $entity);
 
         $attributes = $this->getPrivateProperty($entity, 'attributes');
-        $this->assertFalse(isset($attributes['foo']));
-        $this->assertTrue(isset($attributes['default']));
+        $this->assertArrayNotHasKey('foo', $attributes);
```
kenjis, line 75: This assertion fails. Because `$attributes` is
```
array (4) [
    'foo' => null
    'bar' => null
    'default' => string (6) "sumfin"
    'created_at' => null
]
```
What do we do?

samsonasik: You can register `Rector\PHPUnit\Rector\MethodCall\AssertIssetToSpecificMethodRector` into Option::SKIP

```php
use Rector\PHPUnit\Rector\MethodCall\AssertIssetToSpecificMethodRector;

    $parameters->set(Option::SKIP, [
             // ....

            AssertIssetToSpecificMethodRector::class => [
                     __DIR__ . '/tests/system/Entity/EntityTest.php',
           ],
    ]);
```

tests/system/Entity/EntityTest.php
```diff
@@ -908,10 +908,10 @@ public function testIssetKeyMap()
         $entity = $this->getEntity();
 
         $entity->created_at = '12345678';
-        $this->assertTrue(isset($entity->createdAt));
+        $this->assertObjectHasAttribute('createdAt', $entity);
```
kenjis, line 911: This assertion fails.

tests/system/Session/SessionTest.php
```diff
@@ -248,7 +248,7 @@ public function testIssetReturnsTrueOnSuccess()
 
         $_SESSION['foo'] = 'bar';
 
-        $this->assertTrue(isset($session->foo));
+        $this->assertObjectHasAttribute('foo', $session);
```
kenjis, line 251: This assertion fails.

samsonasik: You can register `Rector\PHPUnit\Rector\MethodCall\AssertIssetToSpecificMethodRector` into Option::SKIP

```php
use Rector\PHPUnit\Rector\MethodCall\AssertIssetToSpecificMethodRector;

    $parameters->set(Option::SKIP, [
             // ....

            AssertIssetToSpecificMethodRector::class => [
                     __DIR__ . '/tests/system/Entity/EntityTest.php',
                    __DIR__ . '/tests/system/Session/SessionTest.php',
           ],
    ]);
```

and then rollback change for `tests/system/Entity/EntityTest.php`

samsonasik: If `AssertIssetToSpecificMethodRector` cause invalid result in many places, we can just skip it al together:

```php
use Rector\PHPUnit\Rector\MethodCall\AssertIssetToSpecificMethodRector;

    $parameters->set(Option::SKIP, [
             // ....

            AssertIssetToSpecificMethodRector::class ;
    ]);
```

samsonasik (approved):

tests/system/Session/SessionTest.php
kenjis: I would rather change the test code to remove the skip `AssertIssetToSpecificMethodRector`.

Thank you for telling me about rector!

PR #5332, opened 2021-11-15
Exclude field of type identity from insert/update statement

on-dev:
SqlServer doesn't support insert/update for identity fields. I've defined a public function "get_identities()" in my model class and this function return an array containing te name of the identity fields so i can exclude from MDL statement,  

**Description**
When it's necessary you can define an array of not updatebale fields like identities: Microsoft SQL Server return error when you try to run statement like "update my_table set id=5, some_text='hello' where id=5" or "insert into my_table (id, some_text) values (null, 'hello')" id filed "id" is an identity.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2493,12 +2493,18 @@ class BaseBuilder
         if (! is_object($object)) {
             return $object;
         }
-
+       
+        try {
+			$identities=$object->get_identities();
+		} catch (\Throwable $th) {
+			$identities=array();
+		}
+        
         $array = [];
 
         foreach (get_object_vars($object) as $key => $val) {
             // There are some built in keys we need to ignore for this conversion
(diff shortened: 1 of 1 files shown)


kenjis:
Thank you for sending us your PR!
But unfortunately, this PR is not going to be able to merge.

- containing PHP error
- coding style violations
- no unit tests
- static analysis errors

Would you please fix the errors?
See 

kenjis:
And my question.

> I've defined a public function "get_identities()" in my model class and this function return an array containing te name of the identity fields

The `CodeIgniter\Model` has the `$primaryKey` property that has the name of the ID column.
https://codeigniter4.github.io/CodeIgniter4/models/model.html#configuring-your-model

Why do you need such a function?

paulbalandan:

PR #5334, opened 2021-11-16
Display file:line and trace information to database queries in debug toolbar

lonnieezell:
Added backtrace information and the file:line of each query into the Database pane of the debug toolbar. The backtrace is hidden by default but opens by clicking on the row. 

Screen Shot 2021-11-15 at 12 51 16 AM


sfadschm:
looks nice!

MGatner:
Looks good! I think the Rector change is relevant but after that everything looks shipshape.

PR #5361, opened 2021-11-20
fix: table alias is prefixed when LIKE clause

### LINKED ISSUE #5360: Bug: table alias is prefixed when LIKE clause
### PHP Version

7.3

### CodeIgniter4 Version

4.1.5, develop (03f5195dceacdf05b574fde05f74dc1742ef7a12)

Git

macOS

### Which server did you use?

cli

### Database

_No response_

### What happened?

See https://forum.codeigniter.com/thread-80559.html

### Steps to Reproduce

```php
$builder = $this->db->table('jobs j');
$builder->like('j.name', 'veloper');
$actual = $builder->getCompiledSelect()
```

```
Expected :'SELECT * FROM "db_jobs" "j" WHERE "j"."name" LIKE '%veloper%' ESCAPE '!''
Actual   :'SELECT * FROM "db_jobs" "j" WHERE "db_""j"."name" LIKE '%veloper%' ESCAPE '!''
```

### Expected Output

`SELECT * FROM "db_jobs" "j" WHERE "j"."name" LIKE '%veloper%' ESCAPE '!'`

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5360

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1012,7 +1012,8 @@ abstract class BaseConnection implements ConnectionInterface
             //
             // NOTE: The ! empty() condition prevents this method
             // from breaking when QB isn't enabled.
-            if (! empty($this->aliasedTables) && in_array($parts[0], $this->aliasedTables, true)) {
+            $firstSegment = trim($parts[0], $this->escapeChar);
+            if (! empty($this->aliasedTables) && in_array($firstSegment, $this->aliasedTables, true)) {
                 if ($protectIdentifiers === true) {
                     foreach ($parts as $key => $val) {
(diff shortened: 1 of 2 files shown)


MGatner:
Be had a strange bug with LIKE table names. Do you think this would apply here? https://github.com/duke-bluesmith/bluesmith/blob/develop/app/Libraries/Notices.php:107

MGatner (approved):

kenjis:
@MGatner If you got an invalid identifier like `"db_""j"."name"`, it is because of this bug.

PR #5370, opened 2021-11-21
Fix precision loss deprecation on float to int conversion

paulbalandan:
**Description**
Related to #4883

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Images/Handlers/GDHandler.php
+++ b/system/Images/Handlers/GDHandler.php
@@ -188,7 +188,7 @@ class GDHandler extends BaseHandler
             imagesavealpha($dest, true);
         }
 
-        $copy($dest, $src, 0, 0, $this->xAxis, $this->yAxis, $this->width, $this->height, $origWidth, $origHeight);
+        $copy($dest, $src, 0, 0, (int) $this->xAxis, (int) $this->yAxis, $this->width, $this->height, $origWidth, $origHeight);
 
         imagedestroy($src);
         $this->resource = $dest;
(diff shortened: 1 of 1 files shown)


MGatner (approved):
So basically we are embracing the loss and circumventing the warning?

system/Images/Handlers/GDHandler.php
```diff
@@ -472,9 +472,9 @@ protected function textOverlay(string $text, array $options = [], bool $isShadow
         // Add the shadow to the source image
         if (! empty($options['fontPath'])) {
             // We have to add fontheight because imagettftext locates the bottom left corner, not top-left corner.
-            imagettftext($src, $options['fontSize'], 0, $xAxis, $yAxis + $options['fontheight'], $color, $options['fontPath'], $text);
+            imagettftext($src, $options['fontSize'], 0, (int) $xAxis, $yAxis + $options['fontheight'], $color, $options['fontPath'], $text);
```
sfadschm, line 475: Is the yAxis automatically cast due to the '+'?

kenjis (approved):

system/Images/Handlers/GDHandler.php
paulbalandan: Good catch! Fixed

paulbalandan:
> So basically we are embracing the loss and circumventing the warning?

Basically, yes. And we're just making it explicit now. Previously, because of loose typing, float got coerced to int implicitly but because this behavior is now deprecated, we're making it explicit to match with existing bahavior. If ever this would be buggy in the future, then we can adjust later.

PR #5375, opened 2021-11-22
fix: Validation for invalid type data

### LINKED ISSUE #5368: Bug: Validation rule can't handle array data
PHP Version

7.4

CodeIgniter4 Version

4,1,5

Composer (new project)

Linux

Which server did you use?

apache

Database

MariaDB 10.1.31

What happened?

Steps to Reproduce
New rule created as below
> is_array($value) {
 return is_array($value);
}

> $data = [
 'arr' => [1,2,3,4,5],
];
$validation->setRules([
 'arr' => 'is_array',
]);

Expected Output

Anything else?

### LINKED ISSUE #5374: Bug: Validation rules `integer` and `numeric` accepts boolean
### PHP Version

7.4

### CodeIgniter4 Version

4.1.5

macOS

### Which server did you use?

apache

### Database

MariaDB 10.2

### What happened?

The validation rules "integer" and "numeric" accepts boolean true&false and return true

### Steps to Reproduce

```php
$data['numeric'] = false;
$data['integer'] = true;

$validation = service('validation');
$validation->setRules([
    'numeric' => 'numeric',
    'integer' => 'integer'
]);

d($validation->run($data));
```

### Expected Output

The validation should fail and return false?

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5374
Fixes #5368

Background:
- validation rule should accept non string data, because it may be int, float, bool, null, array or object when JSON data come
- the type of the data to validate can't be define, because a user (or an attacker) can send unexpected types of data
- the current implementation results in implicit type conversions, and e.g, `true` is converted `'1'`, which may change the result of the validation

This PR does:
- add `declare(strict_types=1)` for Validation and Rule classes
- remove parameter types for data to validate in validation rules

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/CreditCardRules.php
+++ b/system/Validation/CreditCardRules.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * This file is part of CodeIgniter 4 framework.
  *
@@ -169,9 +171,15 @@ class CreditCardRules
      *  $rules = [
      *      'cc_num' => 'valid_cc_number[visa]'
      *  ];
+     *
+     * @param mixed $ccNumber
      */
-    public function valid_cc_number(?string $ccNumber, string $type): bool
+    public function valid_cc_number($ccNumber, string $type): bool
     {
+        if (! is_string($ccNumber)) {
+            return false;
+        }
+
         $type = strtolower($type);
         $info = null;
 
--- a/system/Validation/FileRules.php
+++ b/system/Validation/FileRules.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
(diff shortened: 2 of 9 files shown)


MGatner:
We obviously cannot implement this in version 4 without huge breaking changes. What about moving it to a new namespace (like `CodeIgniter\StrictValidation`) or publishing it as a separate package? That way people could opt-in to it, since the string-based validation is working for most basic cases. 

Also please add a note in the Version 5 project.

kenjis:
@MGatner Thanks for the comment.
I'm going to go a new namespace, because this is a security matter.
Every user should opt-in easily.

kenjis:
Go to #5402.

PR #5376, opened 2021-11-23
fix: invalid INSERT/DELETE query when Query Builder uses table alias

### LINKED ISSUE #5365: Bug: [QueryBuilder] Invalid INSERT query when builder use table alias
### PHP Version

8.0

### CodeIgniter4 Version

4,1,5

Linux

### Which server did you use?

apache

### Database

MariaDB  10.1.31

### What happened?

I get builder in my Model (I do not extend CodeIgniter Model) assigned to property like that:
`$this->builder = $this->dbConn->table($this->getTable() . ' as t');`
`getTable()` return set table name.
When I produce queries for in eg. select, they are produced nicely.
When I want to produce insert query, builder use table alias I provided and produce query like this:

```INSERT INTO `users` as `t` (fields) VALUES (values)```

### Steps to Reproduce

Create builder object with passed table alias and try to produce insert query.

### Expected Output

```INSERT INTO `users` (fields) VALUES (values)```

### Anything else?

I do not extend CodeIgniter Model.

kenjis:
**Description**
Fixes #5365
- remove alias from *FROM string* when INSERT/DELETE

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1863,7 +1863,12 @@ class BaseBuilder
         }
 
         $sql = $this->_insert(
-            $this->db->protectIdentifiers($this->removeAlias($this->QBFrom[0]), true, null, false),
+            $this->db->protectIdentifiers(
+                $this->removeAlias($this->QBFrom[0]),
+                true,
+                null,
+                false
+            ),
             array_keys($this->QBSet),
             array_values($this->QBSet)
         );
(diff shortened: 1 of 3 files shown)


iRedds:
I have nothing against the motive of this PR.
But it starts to smell like crutches.

lonnieezell:
I agree. This feels like we're putting a bandaid on  an error instead of fixing a root cause.

MGatner:
Checking back in on this... I'm all for deeper fixes but we have an open bug with a working solution which seems like a good merge to me. I assume the "do it right" approach would involve a fair amount of class refactoring around conditional aliasing, or something like that - and the trend has been to stay away from large changes to the database layer.

kenjis:
I think if a PR makes CI4 better, we should merge it.
Of course, if the quality is too low or the technical debt is too great, it should be rejected.

If you have a better solution, you should send another PR.
I think it is not better to leave a bug unfixed for months.

MGatner (approved):
I agree. This doesn't lock us out from a refactor fix so even if it is a bandaid it will stop some of the bleeding 😅 Good by me.

michalsn (approved):

PR #5378, opened 2021-11-24
fix: Incorrect type `BaseBuilder::$tableName`

### LINKED ISSUE #4457: Bug: Incorrect type BaseBuilder::$tableName
**Direction**
Property: string 
https://github.com/codeigniter4/CodeIgniter4/blob/8322f3aa396ce91b7e56185cc7eb7c8a0dcf8b56/system/Database/BaseBuilder.php#L163-L165

But in the constructor, you can assign an array to the property. (See docBlock) 
https://github.com/codeigniter4/CodeIgniter4/blob/8322f3aa396ce91b7e56185cc7eb7c8a0dcf8b56/system/Database/BaseBuilder.php#L254-L269

The method returns a string. ( See docBlock and the return typehint)
https://github.com/codeigniter4/CodeIgniter4/blob/8322f3aa396ce91b7e56185cc7eb7c8a0dcf8b56/system/Database/BaseBuilder.php#L317-L321

**Affected module(s)**
BaseBuilder аnd maybe somewhere else

kenjis:
**Description**
Fixes #4457
- make sure `$this->tableName` is string

`$this->tableName` is used only in `BaseBuilder::getTable()`.
`BaseBuilder::getTable()` is used only in `Model::builder()`:
https://github.com/codeigniter4/CodeIgniter4/blob/b8fe16d2ff708971693a5e3325b74447ffb93fdd/system/Model.php#L524

`$table` or `$this->table` in the `Model` is a single table name without an alias.
So the above code is just to make sure that `$table` and the Builder's table name are the same or not.
If these are the same table, it returns existing `$this->builder`.

In other words, if the Builder's table name is not a single table, these are always not the same.

Therefore `BaseBuilder::getTable()` does not have to return array or multiple table string like `'tableA, tableB'`.
Returning empty string is enough.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -158,6 +158,9 @@ class BaseBuilder
      * Tracked separately because $QBFrom gets escaped
      * and prefixed.
      *
+     * When $tableName to the constructor has multiple tables,
+     * the value is empty string.
+     *
      * @var string
      */
     protected $tableName;
@@ -276,7 +279,13 @@ class BaseBuilder
          */
         $this->db = $db;
 
-        $this->tableName = $tableName;
+        // If it contains `,`, it has multiple tables
+        if (is_string($tableName) && strpos($tableName, ',') === false) {
+            $this->tableName = $tableName;  // @TODO remove alias if exists
+        } else {
+            $this->tableName = '';
+        }
+
         $this->from($tableName);
 
         if (! empty($options)) {
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -264,7 +264,9 @@ class BaseBuilder
     /**
      * Constructor
      *
-     * @param array|string $tableName
+     * @param array|string $tableName tablename or tablenames with or without aliases
+     *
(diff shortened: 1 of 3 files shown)


MGatner (changes requested):
I think this is a fine fix for a messy situation. Definitely need to revisit this in version 5. Please add a note to the changelogs on the off chance someone was using this, and link this PR or issue to our version 5 project column.

kenjis:
@MGatner Is there any chance to include this PR in the next version (4.1.6)?

I created a more aggressive PR #5387 for v5.0.
It's kind of a memo to remember my current thoughts.

Added a note in the project Version 5.

MGatner:
@kenjis I think that is a good approach but I would still like a note in the changelogs, something like this:

> Multiple table names will no longer be stored in `BaseBuilder::$tableName` - an empty string will be used instead.

kenjis:
@MGatner I added in the changelog.

MGatner (approved):

MGatner:
Thank you!

PR #5383, opened 2021-11-24
docs: replace `PSR*` with `PSR-*`

kenjis:
**Description**
- unify notation

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/general/modules.rst
+++ b/user_guide_src/source/general/modules.rst
@@ -15,8 +15,8 @@ language files, etc. Modules may contain as few, or as many, of these as you lik
 Namespaces
 ==========
 
-The core element of the modules functionality comes from the :doc:`PSR4-compatible autoloading </concepts/autoloader>`
-that CodeIgniter uses. While any code can use the PSR4 autoloader and namespaces, the primary way to take full advantage of
+The core element of the modules functionality comes from the :doc:`PSR-4 compatible autoloading </concepts/autoloader>`
+that CodeIgniter uses. While any code can use the PSR-4 autoloader and namespaces, the primary way to take full advantage of
 modules is to namespace your code and add it to **app/Config/Autoload.php**, in the ``psr4`` section.
 
 For example, let's say we want to keep a simple blog module that we can re-use between applications. We might create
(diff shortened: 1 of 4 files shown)


user_guide_src/source/general/modules.rst
```diff
@@ -129,7 +129,7 @@ Discovery and Composer
 ======================
 
 Packages that were installed via Composer will also be discovered by default. This only requires that the namespace
-that Composer knows about is a PSR4 namespace. PSR0 namespaces will not be detected.
+that Composer knows about is a PSR-4 namespace. PSR-0 namespaces will not be detected.
```
paulbalandan, line 132: off-topic to this PR..

This notion on `PSR-0` is quite misleading. Composer can also detect PSR-0 namespaces via the `autoload.psr-0` key in composer.json. In fact, `mikey179/vfsstream` is using `PSR-0`. I think this should be removed in a separate PR.

paulbalandan (approved):

user_guide_src/source/general/modules.rst
kenjis: Okay, I will do.

kenjis: Wait a second. This is not about Composer, but about `$discoverInComposer` in CI4.
We need to take a look.

kenjis: It calls `$composer->getPrefixesPsr4()`, so PSR-0 namespaces will not be detected, right?
https://github.com/codeigniter4/CodeIgniter4/blob/7c08819c7e39770fb68cb3bc3aebb35a0aa32c44/system/Autoloader/Autoloader.php#L308

paulbalandan: Hmm. Then we need a little bit of rewording.

`Packages installed via Composer using PSR-4 namespaces will also be discovered by default. Discovery of PSR-0 namespaced packages are not currently supported.`

kenjis: PSR-0 is deprecated. We will not support it. I created a PR #5389.

PR #5392, opened 2021-11-25
fix: `IncomingRequest::getJsonVar()` may cause TypeError

### LINKED ISSUE #5391: Bug: public function getJsonVar error in the code
### PHP Version

8.0

### CodeIgniter4 Version

4.1.2

Manual (zip or tar.gz)

Linux

### Which server did you use?

apache

### Database

_No response_

### What happened?

If "null" is sent to the server and when trying to get data through "Services :: request () -> getVar('myKey')", the error "Argument 2 passed to dot_array_search () must be of the type array..."

### Steps to Reproduce

I am sending a POST request without data

### Expected Output

Services :: request () -> getVar ('myKey') function should return "null"

### Anything else?

File: 
system/HTTP/IncomingRequest.php

`public function getJsonVar()`

 Line 540: `$data = dot_array_search($index, $this->getJSON(true));`

if `$this->getJSON(true)` returns null then an error appears

Solution:

```
$getJson = $this->getJSON(true) ?: [];
$data = dot_array_search($index, $getJson);
```

kenjis:
**Description**
Fixes #5391

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -537,7 +537,11 @@ class IncomingRequest extends Request
     {
         helper('array');
 
-        $data = dot_array_search($index, $this->getJSON(true));
+        $json = $this->getJSON(true);
+        if (! is_array($json)) {
+            return null;
+        }
+        $data = dot_array_search($index, $json);
 
         if ($data === null) {
             return null;
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -520,7 +520,7 @@ class IncomingRequest extends Request
      */
     public function getJSON(bool $assoc = false, int $depth = 512, int $options = 0)
     {
-        return json_decode($this->body, $assoc, $depth, $options);
+        return json_decode($this->body ?? '', $assoc, $depth, $options);
     }
 
     /**
--- a/tests/system/HTTP/IncomingRequestTest.php
+++ b/tests/system/HTTP/IncomingRequestTest.php
@@ -394,6 +394,29 @@ final class IncomingRequestTest extends CIUnitTestCase
         $this->assertSame('buzz', $all['fizz']);
     }
 
+    /**
+     * @see https://github.com/codeigniter4/CodeIgniter4/issues/5391
+     */
+    public function testGetJsonVarReturnsNullFromNullBody()
+    {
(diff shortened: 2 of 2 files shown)


iRedds:
JSON can contain null as a value. Therefore, I do not know if this way of solving the problem is correct.

kenjis:
If the user input is `'null'`, `json_decode()` returns `null`.
`null` is not an array, so there is no value for a key.

iRedds:
i mean
```php
// body: {"var" : null}
$request->getVar('var'); // null

// body: {}
$request->getVar('var'); // null

// body: null
$request->getVar('var'); // null
```
If we assume that the value "var" can be null, then the last two examples are incorrect behavior.
And it seems to me you need to throw an exception.

If we compare it with working with an array, then:

```php
$a = [];
$a['test'];
// Warning:  Undefined array key "test"

$a = null;
$a['test'];
// Warning:  Trying to access array offset on value of type null
```

kenjis:
It depends on `getJsonVar()`'s specification.
It throws an exception or not.

If it should throw an exception, #5391 is not a bug.

MGatner (approved):

samsonasik:
@kenjis could you verify PHPUnit error on develop branch?

kenjis:
@samsonasik Do you mean this?
> PHP Fatal error:  Declaration of PhpCoveralls\Bundle\CoverallsBundle\Console\Application::getDefinition() must be compatible with Symfony\Component\Console\Application::getDefinition(): Symfony\Component\Console\Input\InputDefinition in /home/runner/.composer/vendor/php-coveralls/php-coveralls/src/Bundle/CoverallsBundle/Console/Application.php on line 44
https://github.com/codeigniter4/CodeIgniter4/runs/4376649089?check_suite_focus=true

It is an error of PhpCoveralls: https://github.com/php-coveralls/php-coveralls/issues/327

samsonasik:
Yes, is there a way to pin to specific coveralls version that is working?

kenjis:
Adding this to composer.json?
```
"symfony/console": "< 6.0",
```

samsonasik:
Yes, that's probably the temporary fix 👍

kenjis:
@samsonasik Okay, I will do it.

PR #4739, opened 2021-05-25
Additional check for `$argv` variable when detecting CLI

paulbalandan:
**Description**
After checking existence of `REMOTE_ADDR` and `HTTP_USER_AGENT` keys in `$_SERVER`, additionally check if the `$argv` is set. Please note this will fail if the ini directive `register_argv_argc` is turned on for non-CLI requests.

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

Issue #4699, opened 2021-05-17
Bug: cli not working with cron


pixobit:
**Direction**
We use github issues to track bugs, not for support.  
If you have a support question, or a feature request, raise these as threads on our
[forum](https://forum.codeigniter.com/index.php).

**Describe the bug**
I was trying to run CI4 through cli, and it didn't seem to be working... So after a long debugging session, I've found out that CI4 wasn't recognizing it as a cli execution, and the routing was failing because of this. The reason was, that if CI4 is called by crontab, the PHP_SAPI constant will return 'cgi-fcgi' instead of 'cli' which is checked against in Common.php fn is_cli()

**CodeIgniter 4 version**
4.1.1

**Affected module(s)**
Which package or class is the bug in, if known.

**Expected behavior, and steps to reproduce if appropriate**
Expected to work the same way, as when I run it from cli

**Context**
 - OS: CentOS 7
 - Web server Apache 2
 - PHP version 7.4

paulbalandan:
When running inside the cron, can you check the value of the `TERM` env variable?
```php
var_dump(getenv('TERM'));
```

pixobit:
Will check later today when I get a bit of free time

pixobit:
it returns "bool(false)"

paulbalandan:
Last checks to see if CGI can be safely regarded as CLI.
```php
var_dump(! isset($_SERVER['REMOTE_ADDR']));
var_dump(! isset($_SERVER['HTTP_USER_AGENT']));
var_dump(count($_SERVER['argv']) > 0);
var_dump(! isset($_SERVER['REQUEST_METHOD']));
```

PR #5393, opened 2021-11-26
fix: `is_cli()` returns `true` when `$_SERVER['HTTP_USER_AGENT']` is missing

### LINKED ISSUE #5336: Bug: `is_cli()` returns `true` on Apache
### PHP Version

7.4

### CodeIgniter4 Version

4.1.5

Linux

### Which server did you use?

apache

### Database

_No response_

### What happened?

`is_cli()` returns `true` after upgrade to 4.1.5 from 4.0.

### Steps to Reproduce

Navigate to the site.

### Expected Output

`is_cli()` returns `false`.

### Anything else?

From https://forum.codeigniter.com/thread-80530-post-391559.html

The environment is:
- Apache 2.0 Handler
- no `$_SERVER['HTTP_USER_AGENT']`
- has `$_SERVER['argv']`
- has `$_SERVER['REMOTE_ADDR']`

kenjis:
**Description**
Fixes #5336

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -643,7 +643,13 @@ if (! function_exists('is_cli')) {
      */
     function is_cli(): bool
     {
-        return in_array(PHP_SAPI, ['cli', 'phpdbg'], true);
+        if (in_array(PHP_SAPI, ['cli', 'phpdbg'], true)) {
+            return true;
+        }
+
+        // PHP_SAPI could be 'cgi-fcgi', 'fpm-fcgi'.
+        // See https://github.com/codeigniter4/CodeIgniter4/pull/5393
+        return ! isset($_SERVER['REMOTE_ADDR']) && ! isset($_SERVER['REQUEST_METHOD']);
     }
 }
 
--- a/system/Common.php
+++ b/system/Common.php
@@ -643,16 +643,7 @@ if (! function_exists('is_cli')) {
      */
     function is_cli(): bool
     {
-        if (defined('STDIN')) {
-            return true;
-        }
-
-        if (! isset($_SERVER['REMOTE_ADDR'], $_SERVER['HTTP_USER_AGENT']) && isset($_SERVER['argv']) && count($_SERVER['argv']) > 0) {
-            return true;
-        }
-
-        // if source of request is from CLI, the `$_SERVER` array will not populate this key
-        return ! isset($_SERVER['REQUEST_METHOD']);
+        return in_array(PHP_SAPI, ['cli', 'phpdbg'], true);
     }
 }
 


MGatner:
These were changes pretty recently by @paulbalandan to address other issues - he should take a look.

paulbalandan (commented):
I'm not sure how to approach this. The recent changes in the `is_cli()` method were to account for cron jobs running in `php-cgi` (#4699 fixed by #4725), edge case in #4739, as well as #4843 and #4848. Obviously, accepting this fix would fail cronjobs in cgi fixed previously.

I still believe the proper fix for this is in my https://github.com/codeigniter4/CodeIgniter4/issues/5336#issuecomment-973801389 by expanding the `isset` calls. Also, seen in StackOverflow https://stackoverflow.com/a/25967493

kenjis:
Thanks.

#4699 is `cgi-fcgi`.
#4739 does not show why. What's the issue to solve?
#4843 is `fpm-fcgi`.
#4848 is `fpm-fcgi`.

> Obviously, accepting this fix would fail cronjobs in cgi fixed previously.

I agree with it.
It seems there are users who do not use `php-cli` with cron.

kenjis:
@paulbalandan I added conditions.

I removed `HTTP_USER_AGENT` and `argv` conditions, because #5336 has `argv` and no `HTTP_USER_AGENT` with Apache environment.

paulbalandan:
#4739 was raised by @najdanovicivan in the Slack channel. I cannot remember the issue then, but adding the `argv` checks fixes his issue.

kenjis:
Searched slack, and I found Ivan said the following causes the issue.
```php
if (! isset($_SERVER['REMOTE_ADDR'], $_SERVER['HTTP_USER_AGENT']))
{
   return true;
}
```
And said `$_SERVER[‘argv’]` is not set.

The following is the fixed code at that time:
```php
		if (PHP_SAPI === 'cli')
		{
			return true;
		}

		if (defined('STDIN'))
		{
			return true;
		}

		if (stristr(PHP_SAPI, 'cgi') && getenv('TERM'))
		{
			return true;
		}

		if (! isset($_SERVER['REMOTE_ADDR'], $_SERVER['HTTP_USER_AGENT']) && isset($_SERVER['argv']) && count($_SERVER['argv']) > 0)
		{
			return true;
		}

		// if source of request is from CLI, the `$_SERVER` array will not populate this key
		return ! isset($_SERVER['REQUEST_METHOD']);
```

kenjis:
@najdanovicivan Do you have any problem with this PR?

najdanovicivan:
> @najdanovicivan Do you have any problem with this PR?

All good here. I had issues with some web requests being treated as CLI due to missing HTTP_USER_AGENT

paulbalandan (approved):
Since all involved parties are okay with this, then it is good to go.

MGatner:
💪

PR #5394, opened 2021-11-26
chore: set locale and timezone in test bootstrap

kenjis:
**Description**
In app execution, the default locale and timezone are set from `Config\App` values.
But in test execution, these are not set now.

https://github.com/codeigniter4/CodeIgniter4/blob/0a5508202b1c08928c7bcb863837a8a119951b32/system/CodeIgniter.php#L180-L184

The app and system code may depend on the default locale/timezone.
It is better to set these in testing.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -12,6 +12,7 @@
 use CodeIgniter\Config\DotEnv;
 use CodeIgniter\Router\RouteCollection;
 use CodeIgniter\Services;
+use Config\App;
 use Config\Autoload;
 use Config\Modules;
 use Config\Paths;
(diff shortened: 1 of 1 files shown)


paulbalandan (approved):

MGatner (changes requested):
This should already be handled when the App instance is created and initialized during testing: https://github.com/codeigniter4/CodeIgniter4/blob/0a5508202b1c08928c7bcb863837a8a119951b32/system/Test/CIUnitTestCase.php#L511-L512

If you are seeing that it is not set somewhere then it might be a different bug. Please do not merge this change, additionally because you should not call `new App()` directly in case the service has been overridden.

kenjis:
@MGatner You're correct! It is set in `CIUnitTestCase::setUp()`.

If I recall it correctly, I saw the timezone UTC during testing, but I can't reproduce it now.

PR #5398, opened 2021-11-26
fix: prefix primary key with table name

Xscratch:
This fix prevents ambiguity in case of complex queries.

**Description**
When injecting joins in the Model's `beforeFind` event it could happen that two tables have the same `$this->primaryKey` (ex: id) so the final query could result in the following mysql error: `Column 'id' in where clause is ambiguous`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -589,7 +589,7 @@ class Model extends BaseModel
         return parent::shouldUpdate($data)
             && $this->useAutoIncrement
                 ? true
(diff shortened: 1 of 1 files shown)


kenjis:
Thank you for sending a PR.

Could you add *why* in `Description`?

And we expect all contributions unit tested (in the test folder).

Xscratch:
Thanks @kenjis.
Description added, I'll try writing the tests ASAP.

paulbalandan:

PR #5402, opened 2021-11-28
feat: Add Strict Validation

### LINKED ISSUE #5368: Bug: Validation rule can't handle array data

### LINKED ISSUE #5374: Bug: Validation rules `integer` and `numeric` accepts boolean

kenjis:
**Description**
Supersedes #5375
Fixes #5374
Fixes #5368

- add `CodeIgniter\ValidationStrict`
  - add `declare(strict_types=1)` for Validation and Rule classes
  - remove parameter types for data to validate in validation rules

See #5375, if you want to see the change to the current Validation class.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.github/workflows/test-phpcpd.yml
+++ b/.github/workflows/test-phpcpd.yml
@@ -38,4 +38,4 @@ jobs:
           extensions: dom, mbstring
 
       - name: Detect code duplication
-        run: phpcpd --exclude system/Test --exclude system/ThirdParty --exclude system/Database/SQLSRV/Builder.php --exclude system/Database/SQLSRV/Forge.php -- app/ public/ system/
+        run: phpcpd --exclude system/Test --exclude system/ThirdParty --exclude system/Database/SQLSRV/Builder.php --exclude system/Database/SQLSRV/Forge.php --exclude system/ValidationStrict -- app/ public/ system/
--- a/app/Config/Feature.php
+++ b/app/Config/Feature.php
@@ -24,4 +24,15 @@ class Feature extends BaseConfig
      * @var bool
      */
     public $multipleFilters = false;
+
+    /**
+     * Enable strict validation or not
+     *
+     * If you enable this:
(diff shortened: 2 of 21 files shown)


MGatner:
I don't recall seeing any breaking changes in the original PR for `CodeIgniter\Validation\Calidation` - is there a compelling reason to make a new validation engine instead of just updated rules with tweaks to the engine? That would allow us to roll out the new rules to **app/** without requiring Feature toggle.

kenjis:
@MGatner Nice opinion!
The biggest breaking change is adding `declare(strict_types=1)` to `Validation.php`, but it seems to be not a must.

I will create a new PR for just fixing a bug of Validatiion. If you accept it, we don't need the Feature toggle.

kenjis:
@MGatner I sent a PR #5405.

kenjis:
I sent another PR  #5445.
So this PR is no longer needed.

PR #5405, opened 2021-11-29
fix: Validation cannot handle array item

### LINKED ISSUE #5368: Bug: Validation rule can't handle array data

kenjis:
**Description**
Supersedes #5402
Fixes #5368

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -143,6 +143,8 @@ class Validation implements ValidationInterface
             if ($values === []) {
                 // We'll process the values right away if an empty array
                 $this->processRules($field, $setup['label'] ?? $field, $values, $rules, $data);
+
+                continue;
             }
 
             if (strpos($field, '*') !== false && is_array($values)) {
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -295,7 +295,8 @@ class Validation implements ValidationInterface
                     $value = json_encode($value);
                 }
 
-                $this->errors[$field] = $error ?? $this->getErrorMessage($rule, $field, $label, $param, $value);
+                $param                = ($param === false) ? '' : $param;
+                $this->errors[$field] = $error ?? $this->getErrorMessage($rule, $field, $label, $param, (string) $value);
 
                 return false;
             }
@@ -646,6 +647,9 @@ class Validation implements ValidationInterface
      */
     protected function getErrorMessage(string $rule, string $field, ?string $label = null, ?string $param = null, ?string $value = null): string
     {
+        $param = $param ?? '';
+
+        // Check if custom message has been defined by user
         if (isset($this->customErrors[$field][$rule])) {
             $message = lang($this->customErrors[$field][$rule]);
         } else {
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -215,7 +215,7 @@ class Validation implements ValidationInterface
         }
 
         if (in_array('permit_empty', $rules, true)) {
-            if (! in_array('required', $rules, true) && (is_array($value) ? $value === [] : trim($value ?? '') === '')) {
+            if (! in_array('required', $rules, true) && (is_array($value) ? $value === [] : trim((string) $value) === '')) {
                 $passed = true;
 
                 foreach ($rules as $rule) {
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -308,7 +308,8 @@ class Validation implements ValidationInterface
                     $value = json_encode($value);
                 }
 
-                $param                = ($param === false) ? '' : $param;
+                $param = ($param === false) ? '' : $param;
+
                 $this->errors[$field] = $error ?? $this->getErrorMessage(
                     $rule,
                     $field,
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -285,7 +285,8 @@ class Validation implements ValidationInterface
                     }
 
                     $found  = true;
-                    $passed = $param === false ? $set->{$rule}($value, $error)
+                    $passed = $param === false
+                        ? $set->{$rule}($value, $error)
                         : $set->{$rule}($value, $param, $data, $error);
 
                     break;
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -261,7 +261,7 @@ class Validation implements ValidationInterface
             // Placeholder for custom errors from the rules.
             $error = null;
 
-            // If it's a callable, call and and get out of here.
+            // If it's a callable, call and get out of here.
             if ($isCallable) {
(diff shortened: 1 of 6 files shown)


MGatner (changes requested):
The code is a bit hard to review on mobile but what I see all looks good, and the tests are great. 

Your User Guide write-up is very thorough, but I would rather see that in the changelogs (i.e. ** user_guide_src/source/changelogs/v4.1.6.rst**) and keep the Upgrading section as actual *tasks* that need to be done when upgrading. A simple sentence with reference to the longer explanation in the changelog would be good, like "Due to a bug fix Validation now might ____ (see ____.rst), so make sure you ____ to any code calling it."

MGatner:
Will this conflict with the Strict Validation changes?

kenjis:
@MGatner I added the last commit to fix the docs. I rebased because the commits are not verified.

> Will this conflict with the Strict Validation changes?

No.
#5402 will be split into two PRs. This is one.

MGatner (approved):
Doc changes look good! Since this is a breaking change I'd like it if you could get another reviewer. I'll ping the team on Slack.

system/Validation/Validation.php
```diff
@@ -139,16 +139,20 @@ public function run(?array $data = null, ?string $group = null, ?string $dbGroup
...
-                // Otherwise, we'll let the loop do the job
-                $this->processRules($field, $setup['label'] ?? $field, $value, $rules, $data);
+            if (strpos($field, '*') !== false && is_array($values)) {
+                // Process multiple fields
+                foreach ($values as $value) {
+                    $this->processRules($field, $setup['label'] ?? $field, $value, $rules, $data);
+                }
+            } else {
+                // Process single field
+                $this->processRules($field, $setup['label'] ?? $field, $values, $rules, $data);
```
paulbalandan, line 155: Since this one is the same with L145, I think we can combine this with that?

system/Validation/Validation.php
```diff
@@ -274,7 +285,8 @@ protected function processRules(string $field, ?string $label, $value, $rules =
                     }
 
                     $found  = true;
-                    $passed = $param === false ? $set->{$rule}($value, $error) : $set->{$rule}($value, $param, $data, $error);
+                    $passed = $param === false ? $set->{$rule}($value, $error)
+                        : $set->{$rule}($value, $param, $data, $error);
```
paulbalandan, line 289: The `operator_linebreak` should break this one.
```suggestion
                    $passed = $param === false
                        ? $set->{$rule}($value, $error)
                        : $set->{$rule}($value, $param, $data, $error);
```

system/Validation/Validation.php
```diff
@@ -295,7 +307,14 @@ protected function processRules(string $field, ?string $label, $value, $rules =
                     $value = json_encode($value);
                 }
 
-                $this->errors[$field] = $error ?? $this->getErrorMessage($rule, $field, $label, $param, $value);
+                $param                = ($param === false) ? '' : $param;
+                $this->errors[$field] = $error ?? $this->getErrorMessage(
```
paulbalandan, line 311: Can we add a line in between? The empty space between `$param` and `=` is quite much.

user_guide_src/source/changelogs/v4.1.6.rst
```diff
@@ -7,12 +7,26 @@ Release Date: Not released
...
 ========
 - Multiple table names will no longer be stored in ``BaseBuilder::$tableName`` - an empty string will be used instead.
 
+.. _changelog-v416-validation-changes:
+
+Validation changes
+------------------
+- The previous version of the Validation can't handle an array item.
+  Because of the bug fix, the validation results may be different,
+  or raise an TypeError.
```
paulbalandan, line 22:
```suggestion
  or raise a ``TypeError``.
```

system/Validation/Validation.php
kenjis: If I remove the `if` statement from L143, tests in `tests/system/Validaton/` passes.
But I'm not sure removing it causes no problem.

What input makes `$values` empty array?

system/Validation/Validation.php
kenjis: Done.

system/Validation/Validation.php
kenjis: I added `continue` in the `if` to fix the bug that `$this->processRules()` run twice.

system/Validation/Validation.php
```diff
@@ -646,6 +669,9 @@ public function setError(string $field, string $error): ValidationInterface
      */
     protected function getErrorMessage(string $rule, string $field, ?string $label = null, ?string $param = null, ?string $value = null): string
     {
+        $param = $param ?? '';
```
samsonasik, line 672: how about `(string) $param` ? ref https://3v4l.org/NijdY#v7.3.33

lonnieezell (approved):

system/Validation/Validation.php
kenjis: @samsonasik What is better?

samsonasik: probably just cast it? 

```suggestion
        $param = (string) $param;
```

kenjis: `$param` is null or string. So `$param = $param ?? ''` is no problem.

I'm not sure, but it seems @paulbalandan likes `$param = $param ?? ''`.
Is there any advantage of `(string)`?

samsonasik: oh, ok it seems other file have it as well, ok then 👍

paulbalandan: I don't know of optimization gains but typing `?? ''` is much faster than `(string)` 😂

system/Validation/Validation.php
paulbalandan: > What input makes `$values` empty array?

This is raised in an issue report. I'll check.

paulbalandan: https://github.com/codeigniter4/CodeIgniter4/issues/4510#issuecomment-813532899

samsonasik (approved):

system/Validation/Validation.php
kenjis: I read the comment and the fix PR.
But I'm not sure we can remove the `if ($values === []) { ... }`.
I feel it is safer not to remove it.

kenjis: @paulbalandan I want to update `upgrade_416.rst`, so I merge this for now.

PR #5415, opened 2021-11-30
refactor: Time::getDst()

kenjis:
**Description**
- failed test on local PHP 8.1
  - https://bugs.php.net/bug.php?id=81660

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -479,7 +479,7 @@ class Time extends DateTime
      */
     public function getDst(): bool
     {
-        return (bool) $this->format('I');
+        return $this->format('I') === '1'; // 1 if Daylight Saving Time, 0 otherwise.
     }
 
     /**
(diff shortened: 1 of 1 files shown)


system/I18n/Time.php
```diff
@@ -479,21 +479,7 @@ public function getQuarter(): string
...
-
-        foreach ($transitions as $transition) {
-            if ($transition['time'] > $this->format('U')) {
-                $daylightSaving = (bool) ($transition['isdst'] ?? $daylightSaving);
-                break;
-            }
-        }
-
-        return $daylightSaving;
+        return (bool) $this->format('I');
```
paulbalandan, line 482: Very nice with the `I` flag. But I would like this to compare against it return values for the benefit of myself and others not aware of the flag.
```suggestion
        return $this->format('I') === '1'; // 1 if Daylight Saving Time, 0 otherwise.
```

paulbalandan (approved):

PR #5419, opened 2021-12-01
feat: add SecureHeaders filter

kenjis:
**Description**
- add filter that add common security related filters to response object

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Filters/SecureHeaders.php
+++ b/system/Filters/SecureHeaders.php
@@ -1,0 +1,71 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Filters;
+
+use CodeIgniter\HTTP\RequestInterface;
+use CodeIgniter\HTTP\ResponseInterface;
+
+/**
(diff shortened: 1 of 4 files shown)


system/Filters/SecureHeaders.php
```diff
@@ -0,0 +1,71 @@
...
+        // https://owasp.org/www-project-secure-headers/#x-permitted-cross-domain-policies
+        'X-Permitted-Cross-Domain-Policies' => 'none',
+
+        // https://owasp.org/www-project-secure-headers/#referrer-policy
+        'Referrer-Policy' => 'same-origin',
+
+        // https://owasp.org/www-project-secure-headers/#x-xss-protection
+        // If you do not need to support legacy browsers, it is recommended that you use
+        // Content-Security-Policy without allowing unsafe-inline scripts instead.
+        // 'X-XSS-Protection' => '1; mode=block',
```
paulbalandan, line 44: If I want to activate this without editing this system file, how will I do it?

kenjis: Extend and override it.

paulbalandan: Hmm ok. Maybe we should add these options in the docs and tell which are activated by default.

kenjis: Added the explanation in the docs.

paulbalandan (approved):

PR #5429, opened 2021-12-02
fix: Security class sends cookies immediately

### LINKED ISSUE #5406: Bug: Security class sends cookies immediately, instead of coordinating with the Response class.
### PHP Version

7.4

### CodeIgniter4 Version

4.1.5

Git

macOS

### Which server did you use?

### Database

n/a

### What happened?

Originally encountered during Feature testing. Code that passed previously is now giving errors that `Cannot modify header information - headers already sent by`. This appears to be happening while the CSRF token is being set while calling a page using the `FeatureTestTrait`. 

### Steps to Reproduce

Create a test using `FeatureTestTrait`. Use that to call a page that is protected with a CSRF hash. Run test. 

### Expected Output

Instead of sending the cookie immediately, it should add the cookie to the Response class, which then sends all cookies out at once to protect against the possibility of sending cookies early and giving the `Output already started` errors. 

### Anything else?

This appears to require letting the Response class (or ResponseTrait?) know how to handle raw cookies, also, in order to completely remove that functionality from the Security class. 

Relevant trace: 
```
ErrorException: Cannot modify header information - headers already sent by (output started at /Users/kilishan/WebSites/personal/Bonfire2/vendor/phpunit/phpunit/src/Util/Printer.php:104)

/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Cookie/CookieStore.php:248
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Cookie/CookieStore.php:172
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Security/Security.php:521
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Security/Security.php:507
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Security/Security.php:493
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Security/Security.php:467
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Security/Security.php:188
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Config/Services.php:552
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Config/BaseService.php:248
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Config/BaseService.php:189
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Config/Services.php:547
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Config/BaseService.php:248
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Common.php:243
/Users/kilishan/WebSites/personal/Bonfire2/vendor/codeigniter4/framework/system/Common.php:277
```

kenjis:
**Description**
Fixes #5406
- move secure cookie check from `Security` to `CookieStore`
- `Security` class does not send cookies 
- add Response::setCookieStore()

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


MGatner:
This adds a new dependency (Cookie on Security), which is fine but it must be defined in [](depfile.yaml).

lonnieezell:
Does this need to handle the `setRawCookie` (?) method from the security class also?

paulbalandan (commented):
Looking at the previous dependency addition warning, I think the `SecurityException` should not be inside the `CookieStore` class. IMO, `CookieStore` should be clean from outside dependencies and just dispatch the cookies in its collection. No outside interference.

The addition of `setCookieStore()` method in `Response` looks good to me. Reading on the related issue, my suggestion would be:
1) On sending the CSRF cookie, `Security` gets the saved instance of the `CookieStore` from the `Response` then save to it the CSRF cookie.
2) `Security` still retains the secure checks on the cookie.
3) `Security` puts the new `CookieStore` instance to the `Response` using `$response->setCookieStore($cookieStore);`
4) `Response`, after saving the new cookie store instance, invokes `sendCookies()` via the `send()` method. This calls `$cookieStore->dispatch()`
5) `CookieStore` will send all the cookies in its store, including the CSRF cookie.

kenjis:
@lonnieezell 
> Does this need to handle the setRawCookie (?) method from the security class also?

What do you mean?

kenjis:
@paulbalandan 

I think it is weird that `CookieStore` dispatches cookies.
`Response` should dispatch cookies.
But if do so, it seems breaking  changes.

Your opinion is just moving the secure checks on the cookie back to `Security`, right?

kenjis:
> Response should dispatch cookies.
> But if do so, it seems breaking changes.

I thought that. But it seems I could go.
I moved dispatching cookie from `CookieStore` to `Respose`.

MGatner:
It looks like @paulbalandan has a better grasp on this so I will not review. Thanks for the work though, I assume this fixes the test case issues.

paulbalandan (approved):
I've looked at how other frameworks send cookies. Particularly, for CakePHP, they send cookies using the `ResponseEmitter` class so I think this is good. In the future, or future versions, I think we can also modularize class responsibilities.

kenjis:
I rebased for just signing commits.

kenjis:
I will add the documentation.

lonnieezell (approved):

system/Security/Security.php
```diff
@@ -528,13 +529,18 @@ private function saveHashInCookie(): void
                 'expires' => $this->expires === 0 ? 0 : time() + $this->expires,
             ]
         );
-        $this->sendCookie($this->request);
+
+        /** @var Response $response */
+        $response = Services::response();
+        $response->setCookie($this->cookie);
```
mostafakhudair, line 535: Any point of storing response service in a variable?

```suggestion
        Services::response()ٍ->setCookie($this->cookie);
```

kenjis: To explicitly indicate dependency.
This is also to avoid disabling the deptrac check.

You may like the short code, but it hides the dependency Response.
deptrac can't detect it.
```php
Services::response()->setCookie($this->cookie);
```

kenjis:
I added the documentation.
Can someone review?

paulbalandan (approved):

MGatner:
💪

PR #5439, opened 2021-12-06
fix: Model::save() may call unneeded countAllResults()

kenjis:
**Description**
When `useAutoIncrement` is disabled, `countAllResults()` is called, even if `$data` does not have id value.
You'll see the query like the following:
```sql
SELECT COUNT(*) AS `numrows` FROM `news` WHERE `id` IS NULL
```

```php
<?php

function f1() {
  return false && false ? true : 'a';
}
function f2() {
  return false && (false ? true : 'a');
}

var_dump(f1()); // "a"
var_dump(f2()); // false
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -584,12 +584,12 @@ class Model extends BaseModel
      */
     protected function shouldUpdate($data): bool
     {
-        // When useAutoIncrement feature is disabled check
+        // When useAutoIncrement feature is disabled, check
         // in the database if given record already exists
         return parent::shouldUpdate($data)
-            && $this->useAutoIncrement
+            && ($this->useAutoIncrement
                 ? true
-                : $this->where($this->primaryKey, $this->getIdValue($data))->countAllResults() === 1;
+                : $this->where($this->primaryKey, $this->getIdValue($data))->countAllResults() === 1);
     }
 
     /**
--- a/system/Model.php
+++ b/system/Model.php
@@ -584,12 +584,15 @@ class Model extends BaseModel
      */
     protected function shouldUpdate($data): bool
     {
+        if (parent::shouldUpdate($data) === false) {
+            return false;
+        }
+
         // When useAutoIncrement feature is disabled, check
         // in the database if given record already exists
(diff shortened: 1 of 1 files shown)


kenjis:
@paulbalandan 
This bug was introduced in v4.1.4:
https://github.com/codeigniter4/CodeIgniter4/pull/4840/files#diff-1a690370bdbdc5997068f9ff210bacef314eeb65d1df73af206783ce055d8d1dL613

Is this a human error or php-cs-fixer error?

paulbalandan (approved):
I am not sure what happened, but I think it was an error. 😂

kenjis:
Don't worry. Nobody's perfect.

I thought if php-cs-fixer causes this error, we have to check the rule.
But now I run php-cs-fixer, `(` `)` are not removed.

I don't like long conditions. It is difficult to read and do step debugging.
I refactored.

paulbalandan (approved):

system/Model.php
```diff
@@ -584,10 +584,13 @@ public function set($key, $value = '', ?bool $escape = null)
...
+            return false;
+        }
+
+        // When useAutoIncrement feature is disabled, check
         // in the database if given record already exists
-        return parent::shouldUpdate($data)
-            && $this->useAutoIncrement
+        return $this->useAutoIncrement
                 ? true
                 : $this->where($this->primaryKey, $this->getIdValue($data))->countAllResults() === 1;
```
paulbalandan, line 595: This could use `||`
```suggestion
        return $this->useAutoIncrement
                || $this->where($this->primaryKey, $this->getIdValue($data))->countAllResults() === 1;
```

kenjis: It is true, but it is difficult for me to get the intent.
I added a `if` statement.

Then suddenly I thought...
Why only when `$this->useAutoIncrement` is true, does it return `true`?
This code comes from #3930.

paulbalandan (approved):
The new code looks much better and understandable. 👍

PR #5450, opened 2021-12-10
update docs about `esc()`

totoprayogo1916:
**Description**
`null` is not `$context`

see:
https://github.com/codeigniter4/CodeIgniter4/blob/b5478b352bc6dc4b993651c41c7e157509d797cd/system/Common.php#L406-L410

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -380,7 +380,7 @@ if (! function_exists('esc')) {
      * If $data is an array, then it loops over it, escaping each
      * 'value' of the key/value pairs.
      *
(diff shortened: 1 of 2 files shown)


kenjis:
@totoprayogo1916 Could you also fix this line?

https://github.com/codeigniter4/CodeIgniter4/blob/b5478b352bc6dc4b993651c41c7e157509d797cd/system/Common.php#L383

kenjis (approved):

PR #5453, opened 2021-12-11
fix: VENDORPATH definition

kenjis:
**Description**
- When using multiple apps, `vendor/` is not `ROOTPATH/vendor/`.

See https://forum.codeigniter.com/thread-80732.html

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -84,20 +84,6 @@ if (! class_exists('CodeIgniter\Services', false)) {
 // Initialize and register the loader with the SPL autoloader stack.
 Services::autoloader()->initialize(new Autoload(), new Modules())->register();
 
-// Now load Composer's if it's available
-if (is_file(COMPOSER_PATH)) {
-    /*
-     * The path to the vendor directory.
-     *
-     * We do not want to enforce this, so set the constant if Composer was used.
-     */
-    if (! defined('VENDORPATH')) {
-        define('VENDORPATH', realpath(ROOTPATH . 'vendor') . DIRECTORY_SEPARATOR);
-    }
-
-    require_once COMPOSER_PATH;
-}
-
 // Load environment settings from .env files into $_SERVER and $_ENV
 require_once SYSTEMPATH . 'Config/DotEnv.php';
 
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -84,6 +84,11 @@ if (! class_exists('CodeIgniter\Services', false)) {
 // Initialize and register the loader with the SPL autoloader stack.
 Services::autoloader()->initialize(new Autoload(), new Modules())->register();
 
+// Now load Composer's if it's available
+if (is_file(COMPOSER_PATH)) {
+    require_once COMPOSER_PATH;
+}
+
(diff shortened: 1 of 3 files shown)


totoprayogo1916:
I've tried it
by changing `COMPOSER_PATH` to `'../../vendor/autoload.php'`

but it still shows HTTP ERROR 500

totoprayogo1916:
I checked, `vendor/autoload.php` is executed now.
But when serving, still HTTP ERROR 500.

kenjis:
@totoprayogo1916 I recommend you don't use relative path like `'../../vendor/autoload.php'`.
It depends on the location of the current directory. Use `__DIR__`.

kenjis:
> But when serving, still HTTP ERROR 500.

Set environment to `development`.

totoprayogo1916:
no change & still showing HTTP ERROR 500

with `defined('COMPOSER_PATH') || define('COMPOSER_PATH', __DIR__ . '../../../../vendor/autoload.php');`

kenjis:
```php
__DIR__ . '/../../../../vendor/autoload.php'
```

And check `COMPOSER_PATH` value if it is correct path.

kenjis:
@totoprayogo1916 Did you run `composer update`?

totoprayogo1916:
> ```
> __DIR__ . '/../../../../vendor/autoload.php'
> ```
> 
> And check `COMPOSER_PATH` value if it is correct path.
my test: https://github.com/totoprayogo1916/ci4-multi
with (this) your branch 
I have test with:
`COMPOSER_PATH` 
- `__DIR__ . '/vendor/autoload.php'`
-  `__DIR__ . '/../vendor/autoload.php'`
-  `__DIR__ . '/../../vendor/autoload.php'`
-  `__DIR__ . '/../../../../vendor/autoload.php'`

and the result is `Class 'Mpdf\Mpdf' not found`

with `__DIR__ . '/../../../vendor/autoload.php'`
the result spark cannot run and the message is
image

kenjis:
@totoprayogo1916 You added unneeded `autoload` in `composer.json`.
That's why you got the error from `spark`.

I sent a PR to your repos. Check it.

totoprayogo1916:
Thanks @kenjis 👍🏻

totoprayogo1916 (approved):

paulbalandan (approved):

system/Test/bootstrap.php
```diff
@@ -84,20 +84,6 @@ class_alias('Config\Services', 'CodeIgniter\Services');
...
-    /*
-     * The path to the vendor directory.
-     *
-     * We do not want to enforce this, so set the constant if Composer was used.
-     */
-    if (! defined('VENDORPATH')) {
-        define('VENDORPATH', realpath(ROOTPATH . 'vendor') . DIRECTORY_SEPARATOR);
-    }
-
-    require_once COMPOSER_PATH;
```
kenjis, line 98: Removing this changes the behavior?

kenjis: @paulbalandan I overlooked the deletion of `require_once COMPOSER_PATH;`.
I restored.

paulbalandan (approved):

PR #5454, opened 2021-12-11
fix: `number_to_currency()` error on PHP 8.1

### LINKED ISSUE #5452: Bug: number_to_currency error on NumberFormatter::setAttribute() using default parameter values
### PHP Version

8.1

### CodeIgniter4 Version

4.1.5

Git

Linux

### Which server did you use?

fpm-fcgi

### Database

_No response_

### What happened?

When use `number_to_currency()` in PHP 8.1, it will throw  `NumberFormatter::setAttribute(): Passing null to parameter #2 ($value) of type int|float is deprecated` exceptions. It doesn't occur on PHP 8.0.13 and older, much like problem discussed in #5371 because of PHP 8.1 disallow null type on non-nullable. It makes problem since `$fraction` is null by default and `NumberFormatter::setAttribute()` won't accept null variables.

### Steps to Reproduce

Simply call to `number_to_currency()` with default value

```
<?php

namespace App\Modules\Home\Controllers;

{
    public function index()
    {
        echo number_to_currency(1000000.00, 'IDR');
    }
}
```

### Expected Output

It should handle currency format as expected.

### Anything else?

Here's partial log trace of problematic functions
```
{
    trace: [
        {
            "function": "errorHandler",
            "class": "CodeIgniter\\Debug\\Exceptions",
            "type": "->"
        },
        {
            "file": "/home/thiekus/web/redacted/system/Helpers/number_helper.php",
            "line": 146,
            "function": "setAttribute",
            "class": "NumberFormatter",
            "type": "->"
        },
        {
            "file": "/home/thiekus/web/redacted/system/Helpers/number_helper.php",
            "line": 124,
            "function": "format_number"
        }
    ]
}
```

kenjis:
**Description**
Fixes #5452
Follow-up #4883

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/number_helper.php
+++ b/system/Helpers/number_helper.php
@@ -112,10 +112,7 @@ if (! function_exists('number_to_amount')) {
 }
 
 if (! function_exists('number_to_currency')) {
-    /**
-     * @param string $locale
-     * @param int    $fraction
-     */
+
     function number_to_currency(float $num, string $currency, ?string $locale = null, int $fraction = 0): string
     {
         return format_number($num, 1, $locale, [
--- a/system/Helpers/number_helper.php
+++ b/system/Helpers/number_helper.php
@@ -112,7 +112,6 @@ if (! function_exists('number_to_amount')) {
 }
 
 if (! function_exists('number_to_currency')) {
-
     function number_to_currency(float $num, string $currency, ?string $locale = null, int $fraction = 0): string
     {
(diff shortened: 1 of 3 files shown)


lonnieezell (approved):

samsonasik:
Thank you @kenjis

PR #5463, opened 2021-12-14
refactor: fix `ThrottleTest::testFlooding`

kenjis:
**Description**
Fixes #5444
- remove time dependency from the test

**Checklist:**
- [x] Securely signed commits
- [] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [] User guide updated
- [x] Conforms to style guide

PR #5470, opened 2021-12-15
fix: Throttler does not show correct token time

### LINKED ISSUE #5458: Bug: Throttler does not show correct token time until 1 token is available
### PHP Version

8.0

### CodeIgniter4 Version

4.1.5

Windows

### Which server did you use?

### Database

_No response_

### What happened?

After some experiments with Throttler where I execute an action and get tokenTime returned after each action, I tackled some unexpected behaviour where I executed a successful action too early.

**Example**:

1. Action
2. token time: 50 seconds
3. wait 5 seconds
4. Action
5. token time: 45 seconds
6. wait 5 seconds
7. Action
8. Successful (even though there should be 40 seconds)

After some more experiments I realized that tokenTime is calculated using timestamp of last successful check instead of calculating when 1 token is available again.

### Steps to Reproduce

Here is a unit test which uses a capacity of 2 and a refill timer of 200. This means

- 1 token every 100 seconds
- 0.01 tokens per second

It measures the token time after doing following steps:

1. check
2. sleep for 3 seconds
3. check
4. check (here it should expect false)
5. Expect token time 100 - 3 seconds = 97 seconds

```php
<?php

use Config\Services;

class ThrottlerTest extends CIUnitTestCase
{
    public function testThrottlerTokenTime()
    {
        $throttler = Services::throttler();

        $seconds = 200;
        $capacity = 2;
        
        // refresh = 200 / 2 = 100 seconds
        // refresh rate = 2 / 200 = 0.01 token per second

        // token should be 2
        $this->assertTrue($throttler->check('test', $capacity, $seconds));
        // token should be 2 - 1 = 1

        // do nothing for 3 seconds
        sleep(3);
        // token should be 1 + 3 * 0.01 = 1.03
        $this->assertTrue($throttler->check('test', $capacity, $seconds));
        // token should be 1.03 - 1 = 0.03
        $this->assertFalse($throttler->check('test', $capacity, $seconds));
        // token should still be 0.03 because check failed
        
        // expect remaining time: (1 - 0.03) * 100 = 97
        $this->assertEquals(97, $throttler->getTokenTime(), 'Wrong token time');
        
    }
}

?>
```

Here is the test output

```
PHPUnit 9.5.10 by Sebastian Bergmann and contributors.

Warning:       XDEBUG_MODE=coverage or xdebug.mode=coverage has to be set

F                                                                   1 / 1 (100%)

Time: 00:03.125, Memory: 10.00 MB

There was 1 failure:

1) ThrottlerTest::testThrottlerTokenTime      
Wrong token time
Failed asserting that 100 matches expected 97.

C:\Users\chris\Documents\projects\php\spotify-web-party\tests\unit\ThrottlerTest.php:31

FAILURES!
Tests: 1, Assertions: 4, Failures: 1.
```

### Expected Output

`Throttler::tokenTime` should return the remaining time until 1 token is available in case check failed.

### Anything else?

Looking at your source code, I saw that you don't use consistent time. Sometimes you use `time()` (php function) and sometimes you use `Throttler::time()` (which uses either test time or `time()` from php). I tried above tests using `$throttler-setTestTime(...)`, but got assertion errors not at `tokenTime`, but at the 2nd `check`.

kenjis:
Need to rebase after merging #5463.

**Description**
Fixes #5458

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Throttle/Throttler.php
+++ b/system/Throttle/Throttler.php
@@ -95,12 +95,20 @@ class Throttler implements ThrottlerInterface
     {
         $tokenName = $this->prefix . $key;
 
+        // Number of tokens to add back per second
+        $rate = $capacity / $seconds;
+        // Number of seconds to get one token
+        $refresh = 1 / $rate;
+
         // Check to see if the bucket has even been created yet.
         if (($tokens = $this->cache->get($tokenName)) === null) {
             // If it hasn't been created, then we'll set it to the maximum
             // capacity - 1, and save it to the cache.
-            $this->cache->save($tokenName, $capacity - $cost, $seconds);
+            $tokens = $capacity - $cost;
+            $this->cache->save($tokenName, $tokens, $seconds);
             $this->cache->save($tokenName . 'Time', $this->time(), $seconds);
+
+            $this->tokenTime = max(1, (int) $refresh);
 
             return true;
         }
@@ -109,15 +117,6 @@ class Throttler implements ThrottlerInterface
         // based on how long it's been since the last update.
         $throttleTime = $this->cache->get($tokenName . 'Time');
         $elapsed      = $this->time() - $throttleTime;
-
-        // Number of tokens to add back per second
-        $rate = $capacity / $seconds;
-
-        // How many seconds till a new token is available.
-        // We must have a minimum wait of 1 second for a new token.
-        // Primarily stored to allow devs to report back to users.
-        $newTokenAvailable = (1 / $rate) - $elapsed;
-        $this->tokenTime   = max(1, $newTokenAvailable);
 
         // Add tokens based up on number per second that
         // should be refilled, then checked against capacity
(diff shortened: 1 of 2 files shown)


kenjis:
I don't get why the tokenTime should be 0.
Why not 60? 

https://github.com/codeigniter4/CodeIgniter4/blob/6cd0839a0eab262c1515d56eba57cab386ae2f4c/tests/system/Throttle/ThrottleTest.php#L39

kenjis:
> After check() has been run and returned false, this method can be used to determine the time until a new token should be available and the action can be tried again.
https://codeigniter4.github.io/CodeIgniter4/libraries/throttler.html#getTokentime

It means when check() returned false, getTokentime() can't be used.
It returns 0.

system/Throttle/Throttler.php
```diff
@@ -128,12 +126,21 @@ public function check(string $key, int $capacity, int $seconds, int $cost = 1):
...
+            $this->cache->save($tokenName, $tokens, $seconds);
+            $this->cache->save($tokenName . 'Time', $this->time(), $seconds);
+
+            $this->tokenTime = 0;
 
             return true;
         }
 
+        // How many seconds till a new token is available.
+        // We must have a minimum wait of 1 second for a new token.
```
MGatner, line 139: Is this a new minimum? If so this should be noted in the user guide and changelog. Is there any reason we would want to support fractional seconds, make this whole thing millisecond based?

MGatner:
Generally looks good, a couple questions inline.

system/Throttle/Throttler.php
kenjis: No, it is not new. And it is already documented:
> In this case, the minimum enforced wait time is one second.
https://codeigniter4.github.io/CodeIgniter4/libraries/throttler.html#getTokentime

MGatner (approved):

PR #5474, opened 2021-12-16
refactor: update deprecated method in DatetimeCast

kenjis:
**Description**
- update deprecated method

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


MGatner (approved):
There is a PHPStan extension to prohibit use of deprecations we should probably add to catch these.

PR #5476, opened 2021-12-17
docs: fix environments.rst

kenjis:
**Description**
- remove environment-specific configuration files 

It seems there is no such a thing in CI4.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/general/environments.rst
+++ b/user_guide_src/source/general/environments.rst
@@ -99,7 +99,7 @@ Error Reporting
 
 Setting the ENVIRONMENT constant to a value of ``development`` will cause
 all PHP errors to be rendered to the browser when they occur.
-Conversely, setting the constant to 'production' will disable all error
+Conversely, setting the constant to ``production`` will disable all error
 output. Disabling error reporting in production is a
 :doc:`good security practice </concepts/security>`.
 
--- a/user_guide_src/source/general/environments.rst
+++ b/user_guide_src/source/general/environments.rst
@@ -103,11 +103,3 @@ Conversely, setting the constant to 'production' will disable all error
 output. Disabling error reporting in production is a
 :doc:`good security practice </concepts/security>`.
 
-Configuration Files
--------------------
-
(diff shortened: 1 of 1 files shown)


user_guide_src/source/general/environments.rst
```diff
@@ -97,17 +97,9 @@ is affected.
 Error Reporting
 ---------------
 
-Setting the ENVIRONMENT constant to a value of 'development' will cause
+Setting the ENVIRONMENT constant to a value of ``development`` will cause
 all PHP errors to be rendered to the browser when they occur.
 Conversely, setting the constant to 'production' will disable all error
```
MGatner, line 102:
```suggestion
Conversely, setting the constant to ``production`` will disable all error
```

user_guide_src/source/general/environments.rst
```diff
@@ -97,17 +97,9 @@ is affected.
...
 :doc:`good security practice </concepts/security>`.
 
-Configuration Files
--------------------
-
-Optionally, you can have CodeIgniter load environment-specific
-configuration files. This may be useful for managing things like
-differing API keys across multiple environments. This is described in
-more detail in the Handling Different Environments section of the
-:doc:`Working with Configuration Files </general/configuration>` documentation.
```
MGatner, line 113: Why was this removed? Does that section not exist?

MGatner (changes requested):

MGatner:
I see now, we don't have such a thing as "Environment-specific configuration files". Boot files are the closest thing, and those are already covered.

I agree with this removal.

kenjis:
Yes, CI3 has environment-specific configuration files, but CI4 does not have.

paulbalandan (approved):

user_guide_src/source/general/environments.rst
kenjis: Fixed.

PR #5482, opened 2021-12-18
fix: KINT d()/trace() visual error when activating CSP

### LINKED ISSUE #5475: Bug:  KINT `d()` visual error in debugging when activating CSP
### PHP Version

7.4

### CodeIgniter4 Version

4.1.5

Manual (zip or tar.gz)

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

kint-rich-script and kint-rich-style in RichRenderer.php inside thirdparty KIN need to put {csp-style-nonce} and {csp-script-nonce} because when "CONTENT SECURITY POLICY" is activated it disables it for security for the same. What causes that being in debug mode with CSP activated disables its visual style.

### Steps to Reproduce

Activate developer mode and CSP.

Use `d('example');` in some controller to debug

### Expected Output

Being able to see the theme used by kint in the web browser for debugging

### Anything else?

into folder _system/ThirdParty/Kint/Renderer/_ in file **RichRenderer.php** into line 383 a 392

`                switch ($type) {
                    case 'script':
                        $output .= '<script class="kint-rich-script">'.$contents.'</script>';
                        break;
                    case 'style':
                        $output .= '<style class="kint-rich-style">'.$contents.'</style>';
                        break;
                    default:
                        $output .= $contents;
                }`

replace by:

`switch ($type) {
                    case 'script':
                        $output .= '<script {csp-script-nonce} class="kint-rich-script">'.$contents.'</script>';
                        break;
                    case 'style':
                        $output .= '<style {csp-style-nonce} class="kint-rich-style">'.$contents.'</style>';
                        break;
                    default:
                        $output .= $contents;
                }`

To return the visual style of KINT when CSP is activated

kenjis:
**Description**
Fixes #5475
- fix `d()`
- fix `trace()`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,9 @@
     "extra": {
         "branch-alias": {
             "dev-develop": "4.x-dev"
+        },
+        "kint": {
+            "disable-helpers": true
         }
     },
     "autoload": {
--- a/system/Common.php
+++ b/system/Common.php
@@ -332,6 +332,41 @@ if (! function_exists('dd')) {
     }
 }
 
+if (! function_exists('d')) {
+    /**
+     * Prints a Kint debug report with CSP support
+     *
+     * @param array ...$vars
+     */
+    function d(...$vars)
+    {
+        Kint::$aliases[] = 'd';
+
+        ob_start();
+        Kint::dump(...$vars);
+        $output = ob_get_clean();
+
+        /** @var App $config */
+        $config = config(App::class);
+
+        if ($config->CSPEnabled === true) {
+            $output = str_replace(
+                [
+                    '<script class="kint-rich-script">',
+                    '<style class="kint-rich-style">',
+                ],
+                [
+                    '<script {csp-script-nonce} class="kint-rich-script">',
+                    '<style {csp-style-nonce} class="kint-rich-style">',
+                ],
+                $output
+            );
+        }
+
+        echo $output;
+    }
+}
+
 if (! function_exists('env')) {
     /**
      * Allows user to retrieve values from the environment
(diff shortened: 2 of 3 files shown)


sfadschm:
Can we still use other helpers like `dd()` this way?

kenjis:
> Can we still use other helpers like dd() this way?

Yes. I confirmed `dd()` and `trace()`.

But `trace()` has the same problem with CSP.
`dd()` works fine, because it exits and CSP header is not sent.

kenjis:
I added fix for `trace()`.

MGatner (commented):
How will this effect CLI output? Will Kint [function modifiers](https://kint-php.github.io/kint/advanced/ ) like `@` and `+` still work?

kenjis:
@MGatner 
Without this PR, you will see the output like this:
screenshott 2021-12-23 20 12 23

But with this PR, `@` does not work.
How Kint changes the behavior just adding `@`?

MGatner:
Have no idea. 🤔 Kint has always seemed like magic to me. I assume it is based on some Reflection trace capabilities. 

I don't fully understand what the issue is to begin with so I'm not sure if this trade-off is worth it. I use the Kint modifiers in some projects so this would be a breaking change for me, no clue if others are even aware of them.

sfadschm:
They actually parse back the source file calling the kint method:
https://stackoverflow.com/a/69890023
https://stackoverflow.com/a/69890287

kenjis:
@sfadschm Thanks!

It is too magic. 
I take down this PR. This is Kint's with CSP issue.
We shouldn't take care of this.

kenjis:
@MGatner I sent another PR #5501
It works with Kint function modifiers.

PR #5497, opened 2021-12-24
fix: Entity's isset() and unset()

### LINKED ISSUE #5495: Bug: Can't unset mapped property in Entitiy
### PHP Version

8.0

### CodeIgniter4 Version

`develop` (60f1367c8ad16c4dea998a9d9a4bb1a7fd48c75c)

Git

macOS

### Which server did you use?

cli

### Database

_No response_

### What happened?

> In this example, when the model sets the full_name field on the User class, it actually assigns that value to the class’ $name property, so it can be set and retrieved through $user->name. The value will still be accessible through the original $user->full_name, also, as this is needed for the model to get the data back out and save it to the database. However, **unset and isset only work on the mapped property, $name, not on the original name, full_name.**
https://codeigniter.com/user_guide/models/entities.html#data-mapping

But unsetting mapped property does not work.

```php
unset($user->name);
```

From https://forum.codeigniter.com/thread-80835.html

**Note**: the user guide says "unset and isset only work on the mapped property", but `isset()` works on the original name.

### Steps to Reproduce

```php
<?php
namespace App\Entities;
use CodeIgniter\Entity\Entity;
class User extends Entity
{
    protected $attributes = [
        'id' => null,
        'name' => null, // Represents a username
        'email' => null,
        'password' => null,
        'created_at' => null,
        'updated_at' => null,
    ];

    protected $datamap = [
        'name' => 'full_name',
    ];
}
```
```php
<?php
namespace App\Controllers;
use App\Entities\User;
{
    public function index()
    {
        $user = new User();
        echo $user->name . PHP_EOL;

        $user->name = 'John Smith';
        echo $user->name . PHP_EOL;

        unset($user->name);
        echo $user->name . PHP_EOL;
    }
}
```
```sh-session
$ php public/index.php 

John Smith
John Smith
```

### Expected Output

```sh-session
$ php public/index.php 

John Smith

```

### Anything else?

_No response_

kenjis:
**Description**
- fixes #5495
- fixes `isset()`may work on db column name in some cases

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity/Entity.php
+++ b/system/Entity/Entity.php
@@ -284,8 +284,8 @@ class Entity implements JsonSerializable
     }
 
     /**
-     * Checks the datamap to see if this column name is being mapped,
-     * and returns the mapped name, if any, or the original name.
+     * Checks the datamap to see if this property name is being mapped,
+     * and returns the db column name, if any, or the original property name.
      *
      * @return mixed|string
      */
@@ -510,6 +510,10 @@ class Entity implements JsonSerializable
      */
     public function __isset(string $key): bool
     {
+        if ($this->isMappedDbColumn($key)) {
+            return false;
+        }
+
         $key = $this->mapProperty($key);
 
         $method = 'get' . str_replace(' ', '', ucwords(str_replace(['-', '_'], ' ', $key)));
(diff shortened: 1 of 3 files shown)


system/Entity/Entity.php
```diff
@@ -530,6 +534,37 @@ public function __isset(string $key): bool
...
+
+    /**
+     * Whether this key is mapped db column name?
+     */
+    protected function isMappedDbColumn(string $key): bool
+    {
+        $maybeColumnName = $this->mapProperty($key);
+
+        // Property name which has mapped column name
+        if (($key !== $maybeColumnName)) {
```
samsonasik, line 554: multiple parentheses can be removed:

```suggestion
        if ($key !== $maybeColumnName) {
```

kenjis: Thanks. 
But this should be removed by cs-fixer or rector.

MGatner (approved):
I think this system is a bit too confusing, but this is a step in the right direction. We might consider a different way to handle this in version 5.

kenjis:
Yes, this is very confusing. At first I didn't understand which is correct the documentation or the implementation.
It is not good design.

PR #5501, opened 2021-12-25
fix: KINT visual error when activating CSP

### LINKED ISSUE #5475: Bug:  KINT `d()` visual error in debugging when activating CSP

kenjis:
**Description**
Supersede #5482
Fixes #5475
- overrides `RichRenderer::preRender()`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -71,6 +71,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $parameters->set(Option::SKIP, [
         __DIR__ . '/app/Views',
         __DIR__ . '/system/Debug/Toolbar/Views/toolbar.tpl.php',
+        __DIR__ . '/system/Debug/Kint/RichRenderer.php',
         __DIR__ . '/system/ThirdParty',
         __DIR__ . '/tests/system/Config/fixtures',
         __DIR__ . '/tests/_support',
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -12,6 +12,7 @@
 namespace CodeIgniter;
 
 use Closure;
+use CodeIgniter\Debug\Kint\RichRenderer;
 use CodeIgniter\Debug\Timer;
 use CodeIgniter\Events\Events;
 use CodeIgniter\Exceptions\FrameworkException;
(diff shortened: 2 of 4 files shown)


lonnieezell (approved):
I like this solution from what I understand of Kint's internals.

MGatner (commented):
This looks much better! Does this require the Kint files published to **system** or will it work with a regular Composer installation too?

sfadschm:
> This looks much better! Does this require the Kint files published to **system** or will it work with a regular Composer installation too?

The autoloader adds Composer namespaces last, so this should work for this case. The Kint files in the ThirdParty folder should also be loaded after the regular system folder.

MGatner (approved):

PR #5507, opened 2021-12-27
Presenter named route feature added

sajibAdhi:

**Description**
I have changed presenter function. I added "Named Routes" feature to the presenter function. 

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -781,7 +781,7 @@ class RouteCollection implements RouteCollectionInterface
             }
         }
 
-        $as = (isset($options['as'])) ? $options['as'] : null;
+        $as = $options['as'] ?? null;
 
         if (in_array('index', $methods, true)) {
             $options['as'] = (isset($options['as'])) ? $as . '.index' : null;
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -229,7 +229,7 @@ class RouteCollection implements RouteCollectionInterface
      */
     public function addPlaceholder($placeholder, ?string $pattern = null): RouteCollectionInterface
     {
-        if (! is_array($placeholder)) {
+        if (!is_array($placeholder)) {
             $placeholder = [$placeholder => $pattern];
         }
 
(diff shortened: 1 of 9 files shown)


kenjis:
Thank you for sending PR.

We expect all code changes or bug-fixes to be accompanied by one or more tests added to our test suite to prove the code works. Please add test code.

It seems there are coding style violations. 

And I don't know Presenter named route. What is it?
How to use?

system/Session/Handlers/ArrayHandler.php
```diff
@@ -76,15 +76,15 @@ public function destroy($id): bool
...
-     */
+    * Cleans up expired sessions.
+    *
+    * @param int $maxLifetime Sessions that have not updated
+                            for the last max_lifetime seconds will be removed.
+    *
+    * @return false|int Returns the number of deleted sessions on success, or false on failure.
+    */
     #[ReturnTypeWillChange]
-    public function gc($max_lifetime)
```
samsonasik, line 87: this is on purpose to have underscored variable, see

https://github.com/codeigniter4/CodeIgniter4/blob/efc1bca5f144a279b3aa52eb70158b5790494347/rector.php#L98-L101

please revert these changes on session handlers files.

tests/system/Cache/Handlers/FileHandlerTest.php
```diff
@@ -48,7 +48,6 @@ protected function setUp(): void
         }
 
         $this->handler = new FileHandler($this->config);
-        $this->handler->initialize();
```
samsonasik, line 51: the empty method call is called on purpose for nothing happen check, see 

https://github.com/codeigniter4/CodeIgniter4/blob/efc1bca5f144a279b3aa52eb70158b5790494347/rector.php#L88-L91

please revert these changes. You can run rector check via:

```bash
vendor/bin/rector --dry-run
```

system/Session/Handlers/ArrayHandler.php
MGatner: It isn't apparent to me why this PR should be touching Session at all - maybe working off an old branch, or somehow some files got crossed?

MGatner (changes requested):
I think something went wrong on your PR. I believe the only changes relevant to your feature should be in `RouteCollection` - please restore the other files. 

You will also need to fix some coding styles to match our Coding Standard - see the docs for automating this. 

Finally we will need some tests and User Guide updates. I glanced through the code and I think I get the gist but having tests and docs not only ensures it is working as intended but also demonstrates how to use this feature. 

EDIT: Sorry for the duplicate response, I missed that @kenjis had already said these things 😅

system/Session/Handlers/ArrayHandler.php
samsonasik: I am not sure, probably direct run rector to specific file/dir on some os (windows ?) which probably somehow make skip config skipped. The run rector should be on root to ensure all applied with the files dependencies if any.

```bash
vendor/bin/rector --dry-run
```

system/Router/RouteCollection.php
```diff
@@ -781,30 +781,41 @@ public function presenter(string $name, ?array $options = null): RouteCollection
             }
         }
 
+        $as = (isset($options['as'])) ? $options['as'] : null;
+
         if (in_array('index', $methods, true)) {
+            $options['as'] = (isset($options['as'])) ? $as . '.index' : null;
```
iRedds, line 787: Too much duplicate code.
I suggest replacing it with an anonymous function.
For example like this
```php
$alias = function ($alias) use ($options) {
    if (isset($options['as']) {
        $options['as'] = "{$options['as']}.{$alias}";
    }
    
    return $options;
};

// use 
$this->get($name . '/show/' . $id, $newName . '::show/$1', $alias('show'));
```
Using an anonymous function:
- code is being optimized
- the original ``$options`` array will not be changed.

kenjis:
It seems you forgot GPG signing. 

system/Session/Handlers/ArrayHandler.php
paulbalandan: @samsonasik I recently tried running Rector latest version to try the parallel feature on my Windows using `vendor/bin/rector` and it seems these changes (changing underscore to camelCase, and removing method calls) occur. Not sure but it seems the root `rector.php` is not used?

samsonasik: @paulbalandan I am not sure, could you try latest `"rector/rector": "dev-main"`? The latest dev-main solve the --config usage if used. If the issue persist in parallel only, probably in rector.php, it needs if os check to not run parallel on windows, 

It would be better to have issue open at rector repo with step to reproduce or github action repo sample that show the issue

paulbalandan:
I see no sign of progress here. Closing for now.

PR #5510, opened 2021-12-28
Feature: Subqueries in the FROM section

iRedds:
**Description**
This PR makes it possible to use subqueries created using the QueryBuilder instead of the table in the FROM section.

This is where we add a subquery to an existing table.::
```php
    $subquery = $db->table('users');
    $builder  = $db->table('jobs')->fromSubquery($subquery, 'alias');
    $query = $builder->get();

    // Produces: SELECT * FROM `jobs`, (SELECT * FROM `users`) AS alias
```

Use the ``$db->newQuery()`` method to make a subquery the main table.::
```php
    $subquery = $db->table('users')->select('id, name');
    $builder  = $db->newQuery()->fromSubquery($subquery, 't');
    $query = $builder->get();

    // Produces: SELECT * FROM (SELECT `id`, `name` FROM users) AS t
```

---
**The following is the first interface. It has been modified along the way.**

A subquery can be passed either directly as a variable or as an array ``[BaseBuilder, 'alias']`` to specify an alias.

```php
$subquery = $db->table('users');
$db->table($subquery)->get();
// SELECT * FROM (SELECT * FROM `users`)

$subquery = $db->table('jobs')
$db->table('users)->from([$subquery, 't])->get();
// SELECT * FROM `users`, (SELECT * FROM `jobs`) AS t
```
Only one subquery can be passed at a time.
```php
$db->table([$subquery1, $subquery2])
// An exception will be thrown
```
This solution will add flexibility to the QueryBuilder and prepare the ground for other features.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2761,6 +2761,7 @@ class BaseBuilder
     /**
      * @param BaseBuilder|Closure $builder
      * @param bool                $wrapped Wrap the subquery in brackets
+     * @param string              $alias   Subquery alias
      */
     protected function buildSubquery($builder, bool $wrapped = false, string $alias = ''): string
     {
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -264,7 +264,7 @@ class BaseBuilder
     /**
      * Constructor
      *
-     * @param array|BaseBuilder|Closure|string $tableName tablename or tablenames with or without aliases
+     * @param array|string $tableName tablename or tablenames with or without aliases
      *
      * Examples of $tableName: `mytable`, `jobs j`, `jobs j, users u`, `['jobs j','users u']`
      *
@@ -515,53 +515,42 @@ class BaseBuilder
     /**
      * Generates the FROM portion of the query
      *
-     * @param array|BaseBuilder|Closure|string $from
+     * @param array|string $from
      *
      * @return $this
      */
-    public function from($from, bool $overwrite = false)
+    public function from($from, bool $overwrite = false): self
     {
         if ($overwrite === true) {
             $this->QBFrom = [];
             $this->db->setAliasedTables([]);
         }
 
-        $tableProcessing = function (string $table, bool $protectIdentifiers = true): void {
-            $table = trim($table);
-            $this->trackAliases($table);
-
-            $this->QBFrom[] = $protectIdentifiers
-                ? $this->db->protectIdentifiers($table, true, null, false)
-                : $table;
-        };
-
-        if (is_array($from) && count($from) === 2 || $this->isSubquery($from)) {
-            [$table, $alias] = is_array($from) ? $from : [$from, ''];
-            if ($this->isSubquery($table)) {
-                if ($this->QBFrom !== [] && $alias === '') {
-                    throw new DatabaseException('Subquery must have an alias');
-                }
-
-                $table = $this->buildSubquery($table, true, $alias);
-                $tableProcessing($table, false);
-
-                return $this;
-            }
-        }
-
-        foreach ((array) $from as $val) {
-            if (! is_string($val)) {
-                throw new DatabaseException('Table name must be string or instance of the BaseBuilder');
-            }
-
-            if (strpos($val, ',') !== false) {
-                foreach (explode(',', $val) as $v) {
-                    $tableProcessing($v);
-                }
+        foreach ((array) $from as $table) {
+            if (strpos($table, ',') !== false) {
+                $this->from(explode(',', $table));
             } else {
-                $tableProcessing($val);
+                $table = trim($table);
+                $this->trackAliases($table);
+                $this->QBFrom[] = $this->db->protectIdentifiers($table, true, null, false);
             }
         }
+
+        return $this;
+    }
+
+    /**
+     * @param BaseBuilder $from  Expected subquery
+     * @param string      $alias Subquery alias
+     *
+     * @return $this
+     */
+    public function fromSubquery(BaseBuilder $from, string $alias): self
+    {
+        $table = $this->buildSubquery($from, true, $alias);
+
+        $this->trackAliases($table);
+        $this->QBFrom[] = $table;
 
         return $this;
     }
@@ -2766,8 +2755,7 @@ class BaseBuilder
     protected function buildSubquery($builder, bool $wrapped = false, string $alias = ''): string
     {
         if ($builder instanceof Closure) {
-            $instance = (clone $this)->from([], true)->resetQuery();
-            $builder  = $builder($instance);
+            $builder($builder = $this->db->newQuery());
         }
 
         $subquery = strtr($builder->getCompiledSelect(), "\n", ' ');
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -264,7 +264,7 @@ class BaseBuilder
     /**
      * Constructor
      *
-     * @param array|string $tableName tablename or tablenames with or without aliases
+     * @param array|BaseBuilder|Closure|string $tableName tablename or tablenames with or without aliases
      *
      * Examples of $tableName: `mytable`, `jobs j`, `jobs j, users u`, `['jobs j','users u']`
      *
@@ -515,7 +515,7 @@ class BaseBuilder
     /**
      * Generates the FROM portion of the query
      *
-     * @param array|string $from
+     * @param array|BaseBuilder|Closure|string $from
      *
      * @return $this
      */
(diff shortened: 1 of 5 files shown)


MGatner:
@iRedds This looks really good - I try to stay away from the database layer so I'm calling for backup. 😅

system/Database/BaseBuilder.php
```diff
@@ -2744,7 +2762,7 @@ protected function isSubquery($value): bool
      * @param BaseBuilder|Closure $builder
      * @param bool                $wrapped Wrap the subquery in brackets
      */
-    protected function buildSubquery($builder, bool $wrapped = false): string
+    protected function buildSubquery($builder, bool $wrapped = false, string $alias = ''): string
```
michalsn, line 2765: Please update docblock for this method.

user_guide_src/source/database/query_builder.rst
```diff
@@ -210,6 +210,26 @@ Permits you to write the FROM portion of your query::
     in the ``$db->table()`` function. Additional calls to ``from()`` will add more tables
     to the FROM portion of your query.
 
+#. **Subqueries:**
+
+In special cases, you can compose a database query using subqueries::
+
+    $subquery = $db->table('users);
```
michalsn, line 217:
```suggestion
    $subquery = $db->table('users');
```

user_guide_src/source/database/query_builder.rst
```diff
@@ -210,6 +210,26 @@ Permits you to write the FROM portion of your query::
...
+    $builder  = $db->table($subquery);
+    $query = $builder->get();
+
+    // Produces: SELECT * FROM (SELECT * FROM users)
+
+You can specify an alias for a subquery by passing an array the first value of which will be the subquery, and the second an alias.::
+
+    $subquery = $db->table('users)->select('id, name');
```
michalsn, line 225:
```suggestion
    $subquery = $db->table('users')->select('id, name');
```

michalsn (approved):
Good addition.

system/Database/BaseConnection.php
```diff
@@ -837,7 +837,7 @@ abstract protected function _transRollback(): bool;
     /**
      * Returns a non-shared new instance of the query builder for this connection.
      *
-     * @param array|string $tableName
+     * @param array|BaseBuilder|Closure|string $tableName
```
kenjis, line 840: This PR seems to be good, but the method name `table()` and the variable name `$tableName` don't make sense to me.

iRedds: What do you mean?

system/Database/BaseBuilder.php
iRedds: The rector removes ``@param $alias`` and ``@return`` attributes from the docblock.
If the behavior of the rector with the ``@param $alias`` attribute can be corrected by adding a description, then with the ``@return`` attribute I do not know what to do.

michalsn: Oh, I see. Let's try calling @samsonasik to the rescue on this one.

samsonasik: @iRedds when there is no description, it is ok to remove the docblock, so current change is ok for me if it is already obvious 👍

system/Database/BaseConnection.php
samsonasik: @kenjis changing parameter name is bc break per php 8.0, ref https://3v4l.org/hpqdm#v8.0.14 vs https://3v4l.org/Gtdk2#v8.0.14

The current change is ok for me

samsonasik (approved):

michalsn:
I'm not up to date on the release planning process, so I'm not clicking the merge button. But this PR looks ready.

system/Database/BaseConnection.php
kenjis: Originally, I was uncomfortable with the name of this table() method. However, the table() method barely made sense since it specified a table name.
I wondered if we could change the name  (add a new method) to a better name if it could specify a subquery.

If no one is uncomfortable with `table()`, I won't insist on it.

iRedds: Can you suggest your own version of the method name?
I will say right away that the names of ``subquery`` and ``fromSubquery`` are not correct. In the first case, we get not a subquery, but the main query with the inclusion of a subquery. In the second case, it is associated with the QueryBuilder.

kenjis: `createBuilder()` or `newBuilder()`.

iRedds: I would agree with such method names, but I associate them with creating a new/clean instance, and now this is impossible due to the fact that the class constructor requires a table name.
For example, in Laravel, the DB::table() method takes a string, a closure, a builder in the $ table parameter. That is, in this case it does not differ from my version. Except for the array. One has to scatter a rake.
But there is also an undocumented DB::query() method that returns a clean instance of the QB.
Unfortunately, our implementation already uses the query() method for a raw request. And the inability to create a clean instance of QB is a headache.

kenjis: Laravel has such `table()` method, then it might be not so bad to use `table()`. Because CI4 has many similar methods with Laravel.
And how about that `$db->table(null)` returns a new/clean QB instance?
If QB contructor's `$tableName` accepts null, we can do it.

iRedds: Alternatively, it's interesting.
But in fact, the ``table()`` method is an alias for the ``from()`` method.
And then the construction ``$db->table(null)->from()`` looks a little strange )

iRedds:
I want to apologize to the team, but I've reworked the implementation and request a review.
It seems to me that using an array to pass the subquery + alias is bullshit. 

* The from() method no longer expects a subquery. Subqueries are now processed by the fromSubquery() method. 
* The table() method no longer expects subqueries.*
* Added newQuery() method that creates a "clean" instance of the BaseBuilder class.

*I thought to add the $ alias parameter to the table() method in order to correctly pass the subquery with an alias, 
```php
$db->table($subquery, 'alias');
```
but for this I would have to change the ConnectionInterface.

MGatner:
You don't need to apologize for improving on your work! I won't attempt a review right now but your description sounds great. Thanks for taking the extra time.

system/Database/BaseConnection.php
```diff
@@ -854,6 +854,14 @@ public function table($tableName)
         return new $className($tableName, $this);
     }
 
+    /**
+     * Returns a new instance of the BaseBuilder class without a table.
+     */
+    public function newQuery(): BaseBuilder
+    {
+        return $this->table('crunch')->from([], true);
```
lonnieezell, line 862: What's with the `crunch` table name? This seems to be getting a build instance using `crunch` as the table name, not "a new instance of BaseBuilder without a table" as the docblock reads. 

I'm assuming this is just to fill a fake table name need? If that's the case let's use some type of random string here instead to ensure we don't accidentally conflict with a table someone might have.

lonnieezell (commented):
This looks like a great addition! I like this way of handling it, also. 

Can you add a test making sure that a `fromSubQuery` and `from` calls play nicely together?

system/Database/BaseConnection.php
iRedds: All attempts to change the BaseBuilder code to create a really clean instance are just marked as "breaking changes".
I hope that someday we will remove the table name from the constructor. 
But so far, as I wrote above, we have no other way to create an instance other than passing a temporary fake table name to the constructor.
In fact, the fake table name will not conflict with the one already existing in the application, but if you think so, what can you say about using the ``uniqid()`` function instead of a fixed table name.

kenjis:
What I want to say is that this is a big API decision that cannot be changed later and should be carefully considered by many maintainers/contributors.

And my opinion is it seems the changed `fromSubquery()` style is better as CI's QB than the first Laravel-like API.

tests/system/Database/Builder/FromTest.php
```diff
@@ -101,6 +101,28 @@ public function testFromReset()
...
+    public function testFromSubquery()
+    {
+        $expectedSQL = 'SELECT * FROM (SELECT * FROM "users") AS alias';
+        $subquery    = new BaseBuilder('users', $this->db);
+        $builder     = $this->db->newQuery()->fromSubquery($subquery, 'alias');
+
+        $this->assertSame($expectedSQL, str_replace("\n", ' ', $builder->getCompiledSelect()));
+
+        $expectedSQL = 'SELECT * FROM (SELECT "id", "name" FROM "users") AS users_1';
+
```
kenjis, line 113:
```suggestion
```

michalsn (approved):

kenjis (approved):

system/Database/BaseConnection.php
```diff
@@ -854,6 +854,14 @@ public function table($tableName)
         return new $className($tableName, $this);
     }
 
+    /**
+     * Returns a new instance of the BaseBuilder class with a cleared FROM clause.
+     */
+    public function newQuery(): BaseBuilder
+    {
+        return $this->table(uniqid())->from([], true);
```
kenjis, line 862: I think the cost of calling `uniqid()` is wasteful.
The table name will be removed by the next `from()` method.
It seems there is no need for uniqueness.

iRedds: Yes and no.
The table name specified in the constructor of the BaseBuilder class is stored in the ``$tableName`` property, which is then used in the Model. Yes, the dependence is weak, but who knows what else will come up with.

Suggest a temporary table name right here.
I will accept it immediately.

kenjis: This is a bit tricky, `','`. It can set empty string $tableName property.

iRedds: I like it

kenjis (approved):

MGatner (approved):
I'm keeping my hands off the heavy DB stuff, but the code looks good and this PR is high quality. @kenjis to you

user_guide_src/source/database/query_builder.rst
```diff
@@ -210,6 +210,28 @@ Permits you to write the FROM portion of your query::
...
+
+Use the ``$db->newQuery()`` method to make a subquery the main table.::
+
+    $subquery = $db->table('users')->select('id, name');
+    $builder  = $db->newQuery()->fromSubquery($subquery, 't');
+    $query = $builder->get();
+
+    // Produces: SELECT * FROM (SELECT `id`, `name` FROM users) AS t
+
+.. note:: Only one subquery can be passed to a method.
```
kenjis, line 233: @iRedds I've already merged this, but I noticed when seeing the user guide.
Is the note necessary?

iRedds: Oops, I really missed that.

MGatner:
🥳

PR #5512, opened 2021-12-28
Logger - Can use custom log levels

najdanovicivan:
**Description**

Logger currently prohibits logging levels which are not included in $logLevels
This one removes that limitation so that custom levels can be implemented using custom handlers.

One possible use case is when you want to log API errors in separate files for each API identified by slug. 

So a custom handler can be made like this:

```php
class FetcherFileHandler extends FileHandler
{
    public function canHandle(string $level): bool
    {
        return substr($level, 0, 8) === 'fetcher-';
    }

    public function handle($level, $message): bool
    {
        $slug   = substr($level, 8);

        $this->path = WRITEPATH . 'logs'. DIRECTORY_SEPARATOR . 'fetcher' . DIRECTORY_SEPARATOR . $slug; 

        if (! file_exists($this->path)) {
            mkdir($this->path, 0755, true);
        }

        return parent::handle($level, $message);
    }
}
```

and then logging can simply be done with
```php
log('fetcher-'.$slug, $message);
```

This will not affect behavior of other handlers in any way as they still have handles specified in config so they won't even be triggered for custom levels.

The only thing that cannot be handled with this is threshold from config but those can be handled in the custom handler implementation.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Log/Logger.php
+++ b/system/Log/Logger.php
@@ -81,7 +81,7 @@ class Logger implements LoggerInterface
     /**
      * Caches instances of the handlers.
      *
-     * @var array
+     * @var HandlerInterface[]
      */
     protected $handlers = [];
 
@@ -252,13 +252,8 @@ class Logger implements LoggerInterface
             $level = array_search((int) $level, $this->logLevels, true);
         }
 
-        // Is the level a valid level?
-        if (! array_key_exists($level, $this->logLevels)) {
-            throw LogException::forInvalidLogLevel($level);
-        }
-
-        // Does the app want to log this right now?
-        if (! in_array($level, $this->loggableLevels, true)) {
+        // Is the level a valid level and Does the app want to log this right now?
+        if (array_key_exists($level, $this->logLevels) && ! in_array($level, $this->loggableLevels, true)) {
             return false;
         }
 
@@ -276,20 +271,25 @@ class Logger implements LoggerInterface
             ];
         }
 
+        $handlers = [];
+
         foreach ($this->handlerConfig as $className => $config) {
             if (! array_key_exists($className, $this->handlers)) {
                 $this->handlers[$className] = new $className($config);
             }
 
-            /**
-             * @var HandlerInterface $handler
-             */
             $handler = $this->handlers[$className];
 
-            if (! $handler->canHandle($level)) {
-                continue;
+            if ($handler->canHandle($level)) {
+                $handlers[] = $handler;
             }
+        }
 
+        if (! in_array($level, $this->loggableLevels, true) && empty($handlers)) {
+            throw LogException::forInvalidLogLevel($level);
+        }
+
+        foreach ($handlers as $handler) {
             // If the handler returns false, then we
             // don't execute any other handlers.
             if (! $handler->setDateFormat($this->dateFormat)->handle($level, $message)) {
--- a/tests/system/Log/LoggerTest.php
+++ b/tests/system/Log/LoggerTest.php
@@ -345,6 +345,21 @@ final class LoggerTest extends CIUnitTestCase
         $this->assertSame($expected, $logs[0]);
     }
 
+    public function testCustomLogsCorrectly()
+    {
+        $config = new LoggerConfig();
+        $logger = new Logger($config);
+
+        $expected = 'CUSTOM - ' . date('Y-m-d') . ' --> Test message';
+
+        $logger->log('custom', 'Test message');
+
+        $logs = TestHandler::getLogs();
+
+        $this->assertCount(1, $logs);
+        $this->assertSame($expected, $logs[0]);
+    }
+
     public function testLogLevels()
     {
         $config = new LoggerConfig();
--- a/system/Test/Mock/MockLogger.php
+++ b/system/Test/Mock/MockLogger.php
@@ -92,6 +92,7 @@ class MockLogger
                 'info',
                 'notice',
                 'warning',
+                'custom',
             ],
 
             // Logging Directory Path


system/Log/Logger.php
```diff
@@ -252,13 +252,8 @@ public function log($level, $message, array $context = []): bool
             $level = array_search((int) $level, $this->logLevels, true);
         }
 
-        // Is the level a valid level?
-        if (! array_key_exists($level, $this->logLevels)) {
-            throw LogException::forInvalidLogLevel($level);
-        }
```
kenjis, line 258: Wouldn't removing this check be a violation of the PSR Log?

> A ninth method, log, accepts a log level as the first argument. Calling this method with one of the log level constants MUST have the same result as calling the level-specific method. Calling this method with a level not defined by this specification MUST throw a Psr\Log\InvalidArgumentException if the implementation does not know about the level.
https://www.php-fig.org/psr/psr-3/

najdanovicivan: It states 

if the implementation does not know about the level. Users SHOULD NOT use a custom level without knowing for sure the current implementation supports it.

I'll update the code to include this but it should be handled a bit differently

kenjis:
@najdanovicivan I think `'fetcher-'.$slug` is not a log level.
So this PR does not seems to be good direction.

najdanovicivan:
I've updated the code to still keep the Exception if there are no handlers which are able to handle the custom level

'fetcher-'.$slug might not be the the best solution but I needed a way to log errors from each API in separate directory and it was easier to add it via $level 

I could have prepended the message with $slug and get it from there or Even use $context but I wanted to avoid using so that I do not get `'{' . $slug . '}'` replaced in the message

But in any case support for custom level is needed so that those log messages do not end up in the regular CI logs

tests/system/Log/LoggerTest.php
```diff
@@ -345,6 +345,21 @@ public function testDebugLogsCorrectly()
         $this->assertSame($expected, $logs[0]);
     }
 
+    public function testCustomLogsCorrectly()
+    {
+        $config = new LoggerConfig();
+        $logger = new Logger($config);
+
+        $expected = 'CUSTOM - ' . date('Y-m-d') . ' --> Test message';
```
kenjis, line 353: This test depends on time. But is there no way to inject time?

najdanovicivan: I've copied that piece of code from existing tests. There are no mocks for time there. The date is generated in Log file handler so it will have to be mocked to use injected time as well.

kenjis: I see. It depends on `date()`. We need to fix the `MockLogger` someday.

kenjis (approved):

kenjis:
@MGatner 

PSR Log accepts custom log levels. So at least custom log levels should be acceptable.

> A ninth method, log, accepts a log level as the first argument. Calling this method with one of the log level constants MUST have the same result as calling the level-specific method. Calling this method with a level not defined by this specification MUST throw a Psr\Log\InvalidArgumentException **if the implementation does not know about the level.**
https://www.php-fig.org/psr/psr-3/

MGatner:
@kenjis your quote actually makes me think we should not add a custom log level. 

> MUST throw a Psr\Log\InvalidArgumentException if the implementation does not know about the level.

Our implementation won't know about this log level, it would merely be making it available to other log handlers that *might* implement `custom`. Which leads me back to the same conclusion: this scenario requires a custom logger. 

@najdanovicivan How do you interpret that PSR quote?

paulbalandan:
If I may to add, the PSR levels are based on RFC 5424 (the Syslog Protocol). IMO, adding a custom level is a violation of that RFC.

I believe a better solution would be by implementing "channels". If I'm not mistaken, monolog/monolog or cakephp/log has that concept. In this we have one channel using an instance of FileLogger saving the logs to one place. Then, we have another channel using another instance of FileLogger saving the logs to another place. By these channels we have many loggers in place that respond differently to a log message and/or log level (which is what this PR generally wants to achieve).

lonnieezell:
> I believe a better solution would be by implementing "channels". If I'm not mistaken, monolog/monolog or cakephp/log has that concept. In this we have one channel using an instance of FileLogger saving the logs to one place. Then, we have another channel using another instance of FileLogger saving the logs to another place. By these channels we have many loggers in place that respond differently to a log message and/or log level (which is what this PR generally wants to achieve).

100% this. That's something that has been on my personal wish list for a while. I believe this fulfills the OP's needs and is a more flexible solution. 

My personal opinion is that this PR should NOT be accepted and we should pursue channels instead.

kenjis:
@MGatner 
> Our implementation won't know about this log level, it would merely be making it available to other log handlers that might implement custom. Which leads me back to the same conclusion: this scenario requires a custom logger.

I see. I hadn't thought of it that way.
I'm not sure which is correct. Your interpretation may be correct.
I don't insist on my interpretation.

najdanovicivan:
PSR states that if level cannot be handled an exception should be thrown which is exactly what the code in the PR does. It goes trough all the handlers to find if there is a handler available which can process the custom level.  If there is no such handler exception will be thrown.

So with this there are no changes to the existing functionality.

So if custom level is passed to the logger and there is no handler for such level the exception will be thrown.  If there is a matched handler using the response form canHandle method handle will be executed on the handler object.

The only thing that is left unchecked is numeric value for the level and threshold. As custom levels cannot be tied to numeric value the handler can decided in handle method if message should be logged or not based on the threshold config.

It is all bit confusing but it gives possibility to just add single handler class and do custom logging. Having to add custom log handler will introduce a couple of other problems.  global log() method uses logger service so in order to be able to use log() everywhere in the code. The custom logger should be used for all the logging in the application. 

This will require maintaining whole custom logger classes which is not a thing I want to do. There is also an option to use custom logger only where I need it but then log() method won't be usable as it will require calling the custom logger directly.

MGatner:
Extending the Logger and creating your own service should be a fairly easy task. That aside, I do see that you handled the custom log level throws appropriately now (thank you) but I think the bigger issues still remain: the perceptual input and the "channels" concept.

Even if nobody cared about the former I think that merging this feature now would make implanting "log channels" much harder: they would have overlapping scenarios and possible multiplicative complexity (e.g. "does each channel need to handle custom log levels?"). 

I'm open to more conversation but rereading this thread my impression is that there is definitely Logger work to be done but this may not be the right way to take it.

paulbalandan:
I believe the consensus here is to not accept this PR. Despite the time and effort contributed in this PR, it's unfortunate we need to close this PR.

PR #5513, opened 2021-12-29
Remove semicolons from SQL statements.

ytetsuro:
**Description**

Removed unnecessary semicolons.
The reason is the following.

* In [oci_parse](https://www.php.net/manual/en/function.oci-parse.php#refsect1-function.oci-parse-parameters), if there is a semicolon, it is treated as PL/SQL.
* There is no semicolon at the end of the SQL except for the one fixed here.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -135,7 +135,7 @@ class Forge
      *
      * @var false|string
      */
-    protected $renameTableStr = 'ALTER TABLE %s RENAME TO %s;';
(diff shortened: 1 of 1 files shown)


michalsn (approved):

kenjis (approved):

ytetsuro:
Thansk merged!

PR #5526, opened 2022-01-02
Add app.supportedLocales feature in .env

FrancoisCapon:
**Description**
* Add `app.supportedLocales` feature in `.env` because indexed arrays are not supported in the `.env` file
* @paulbalandan finded after the pull requests : https://forum.codeigniter.com/thread-77309.html
```
.env
app.supportedLocales = 'fr, en, de'
```

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide (I hope!)

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -461,4 +461,15 @@ class App extends BaseConfig
      * @var bool
      */
     public $CSPEnabled = false;
+    
+    public function __construct()
+    {
+        parent::__construct();
+        // app.supportedLocales = 'fr,  en ,de' => $this->supportedLocales = array('fr,'en,'de');
+        $envSupportedLocales = getenv('app.supportedLocales');
+        if ($envSupportedLocales)
+        {
+            $this->supportedLocales = array_map('trim', explode(',', $envSupportedLocales));
+        }
+    }
 }


kenjis:
>  indexed arrays are not supported in the .env file

No. it can override only existing values.

`.env`:
```
app.supportedLocales.0 = fr
app.supportedLocales.1 = en
app.supportedLocales.2 = de
```

```diff
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -96,7 +96,7 @@ class App extends BaseConfig
      *
      * @var string[]
      */
-    public $supportedLocales = ['en'];
+    public $supportedLocales = ['en', null, null];
 
     /**
      * --------------------------------------------------------------------------
```

Controller:
```
        $config = config('App');
        dd($config->supportedLocales);
```

Result:
```
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ $config->supportedLocales                                                                                  │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
array (3) [
    0 => string (2) "fr"
    1 => string (2) "en"
    2 => string (2) "de"
]
══════════════════════════════════════════════════════════════════════════════════════════════════════════════
Called from .../app/Controllers/Home.php:10 [dd()]
```

app/Config/App.php
```diff
@@ -461,4 +461,15 @@ class App extends BaseConfig
...
+    
+    public function __construct()
+    {
+        parent::__construct();
+        // app.supportedLocales = 'fr,  en ,de' => $this->supportedLocales = array('fr,'en,'de');
+        $envSupportedLocales = getenv('app.supportedLocales');
+        if ($envSupportedLocales)
+        {
+            $this->supportedLocales = array_map('trim', explode(',', $envSupportedLocales));
+        }
```
kenjis, line 473: To be honest, I don't like the solution. Because it is too specific.
The item (`app.supportedLocales`) is hard coded.

kenjis:
@FrancoisCapon Thank you for sending a PR.
But it seems some violations. Please read the [Contributing to CodeIgniter](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/README.md), especially:
 
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#php-style
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#unit-testing

FrancoisCapon:
1. I tested ` app.supportedLocales.n` and I understand why it doesn't work: _No. it can override only existing values._
2. As you wich, it just a proposal. I use it on a produciton site, the goal is to can manage all the locales settings in the `.env` without touching the "code". 
```
app.negotiateLocale = true
app.supportedLocales = 'en, fr' # defaultLocale needed
app.defaultLocale = 'en'
```
3. I will install "pull request requirement tools" and check this pull request for training.

Do I close the pull request ?

MGatner:
You can keep the pull request and add new commits or reset and force-push over your old ones - that way we keep the conversation in one place. 

Up front, I am not okay with overriding the constructor to apply special formatting to a single property. If our `DotEnv` parser needs to be able to support arrays then the change needs to be made there. Maybe we could have modifiers? something like:
```php
app.supportedLocales = 'fr, en, de'.csv
```

paulbalandan:
Sorry, but I am also not in favor of tackling only one property. The proper fix (or enhancement) would boil down to the `DotEnv` class when it parses the `.env` file. It should learn how to handle array values.

@FrancoisCapon thank you for this contribution but I'm afraid we'll have to close this. Feel free to open a new PR in case you have an implementation in the DotEnv class.

PR #5540, opened 2022-01-04
Replace deprecated filter `FILTER_SANITIZE_STRING`

TimZ99:
**Description**
As of PHP8.1 `FILTER_SANITIZE_STRING` is deprecated. Replaced it with `FILTER_SANITIZE_FULL_SPECIAL_CHARS`. This is equivalent to calling htmlspecialchars() with ENT_QUOTES set. ([PHP.net](https://www.php.net/manual/en/filter.filters.sanitize.php))

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/cookie_helper.php
+++ b/system/Helpers/cookie_helper.php
@@ -65,7 +65,7 @@ if (! function_exists('get_cookie')) {
     {
         $prefix  = isset($_COOKIE[$index]) ? '' : config(App::class)->cookiePrefix;
         $request = Services::request();
-        $filter  = $xssClean ? FILTER_SANITIZE_STRING : FILTER_DEFAULT;
+        $filter  = $xssClean ? FILTER_SANITIZE_FULL_SPECIAL_CHARS : FILTER_DEFAULT;
 
         return $request->getCookie($prefix . $index, $filter);
     }
--- a/user_guide_src/source/incoming/incomingrequest.rst
+++ b/user_guide_src/source/incoming/incomingrequest.rst
@@ -399,7 +399,7 @@ The methods provided by the parent classes that are available are:
         The second optional parameter lets you run the data through the PHP's
         filters. Pass in the desired filter type as the second parameter::
 
-            $request->getVar('some_data', FILTER_SANITIZE_STRING);
+            $request->getVar('some_data', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
 
         To return an array of all POST items call without any parameters.
 
@@ -407,7 +407,7 @@ The methods provided by the parent classes that are available are:
         first parameter to null while setting the second parameter to the filter
         you want to use::
 
-            $request->getVar(null, FILTER_SANITIZE_STRING);
+            $request->getVar(null, FILTER_SANITIZE_FULL_SPECIAL_CHARS);
             // returns all POST items with string sanitation
 
         To return an array of multiple POST parameters, pass all the required keys as an array::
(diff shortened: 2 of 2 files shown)


lonnieezell:
Good catch. Thanks!

lonnieezell (approved):

kenjis:
See #5005,  #5263
`esc(strip_tags($arg))` is used in other places.

kenjis (changes requested):
The changes in this PR are not inconsistent.
Please make the same changes in other places.

TimZ99:
@kenjis changing to striptags() requires the underlying functions to be changed and removes flexibility for developers.

kenjis:
Sorry, what do you mean by *removes flexibility* ?

First of all, input filtering like FILTER_SANITIZE_STRING is a incorrect practice.
`htmlspecialchars()` should be used just before outputting HTML.

TimZ99:
Developers now have the option to use a different filter. If we force 'htmlspecialchars()' the dev doesn't have that flexibility.

kenjis:
@TimZ99 Thank you for the explanation. I don't mean to fix `htmlspecialchars()`.
I mean if `$xssClean` is true, the return value should be `esc(strip_tags($value))`.
Because in other places such a value is returned.

How about `get_cookie()` returns `esc(strip_tags($request->getCookie($prefix . $index)))`?

@paulbalandan What do you think?

paulbalandan:
`esc(strip_tags($value))` was used to replace use of `FILTER_SANITIZE_STRING` to maintain the current behavior, although this behavior is incorrect (that's why the filter was deprecated by PHP). Thus, the replacement was just for the time being in preparation for PHP 8.1 compatibility.

If this PR's solution is way better and deviates from the current incorrect behavior, I'd approve this. So, I don't mind if the previously replaced places by `esc(strip_tags($value))` would revert to using `filter_var` and this filter value. However, since this is an improvement, this needs to be documented and mentioned in the changelog.

kenjis:
I searched the source code and found only one `$xssClean` in `get_cookie()`.
It probably comes from CI3. 

I would like to deprecate the param. Because it is a bad practice.
So I don't think FILTER_SANITIZE_FULL_SPECIAL_CHARS is better than FILTER_SANITIZE_STRING or `esc(strip_tags($value))`.

Input filtering like `FILTER_SANITIZE_FULL_SPECIAL_CHARS` (or `FILTER_SANITIZE_STRING`)  is a bad practice. 
It barely makes sense only when the value is used only for HTML output.
But generally, most data is not coupled with the output system.
If you use the filtered data to other system, you must revert it before outputting.
Eg, if you have the value `M&M`, you get the filtered value `M&amp;M`. But if you save it in the database,
the filtering does not make sense. So you should revert it to `M&M` and save it. 
It is double effort. Why don't you convert when you output?

And when you output the filtered value as HTML, you don't escape (`esc()`) the value. 
If you escape it, the value will be double-escaped.
We encourage to use `esc()` in the view. So no escape is against the recommendation.

kenjis:
> I don't mind if the previously replaced places by esc(strip_tags($value)) would revert to using filter_var and this filter value.

I think there are three types of `FILTER_SANITIZE_STRING` usage.

1. Inside Routing/URI processing
2. Request::get*() 2nd param filter
3. get_cookie() 2nd param xssClean

2 or 3 is a bad practice. I don't recommend to use.

2 is to specify filter itself. If a user uses FILTER_SANITIZE_STRING, it is not framework responsibility.
But we have to remove FILTER_SANITIZE_STRING in 3.

1 is a security measure. I'm not sure the way is good practice or not, but some measures must be needed.
If we change it, we must consider it very well.

kenjis:
@TimZ99 Thank you for your contribution.
I will prepare for the next release, I create a new PR based on this, and will add the documentation.
So I close this.

PR #5543, opened 2022-01-05
fix: Undefined index: file: SYSTEMPATH\Debug\Toolbar\Collectors\Database.php at line 142

### LINKED ISSUE #5539: Bug: ErrorException: Undefined index: file: SYSTEMPATH\Debug\Toolbar\Collectors\Database.php at line 142
### PHP Version

7.3

### CodeIgniter4 Version

v4.1.6

Windows

### Which server did you use?

apache

### Database

MariaDB 10.2

### What happened?

The issue is related to the latest build where there seems to be an issue with the trace line of the debugger (Toolbar) when the file attribute is not available.

output log of the trace line causing this issue

```
Array
(
    [function] => read
    [class] => CodeIgniter\Session\Handlers\DatabaseHandler
    [object] => CodeIgniter\Session\Handlers\DatabaseHandler Object
        (
            [DBGroup:protected] => default
            [table:protected] => ...
```

### Steps to Reproduce

Update from 4.1.5 to 4.1.6.

### Expected Output

Maybe check if the file attribute is set?
https://github.com/codeigniter4/CodeIgniter4/commit/e97b95b2ffc8b11a4a97202cab12d6a839ddca93#diff-48dee7e39770b002fe3f48f70db03aaa01fd79d5830d5ee91ee9f8ccda44b4ea

`
if (!isset($traceLine['file'])) {
...then do the clean-up?
}
`

### Anything else?

_No response_

iprixi:
- Fix for using the undefined key of `file`

**Description**
Fixes #5539

### DIFF
--- a/system/Debug/Toolbar/Collectors/Database.php
+++ b/system/Debug/Toolbar/Collectors/Database.php
@@ -139,17 +139,20 @@ class Database extends BaseCollector
 
             foreach ($query['trace'] as &$traceLine) {
                 // Clean up the file paths
-                $traceLine['file'] = str_ireplace(APPPATH, 'APPPATH/', $traceLine['file']);
-                $traceLine['file'] = str_ireplace(SYSTEMPATH, 'SYSTEMPATH/', $traceLine['file']);
-                if (defined('VENDORPATH')) {
-                    // VENDORPATH is not defined unless `vendor/autoload.php` exists
-                    $traceLine['file'] = str_ireplace(VENDORPATH, 'VENDORPATH/', $traceLine['file']);
-                }
-                $traceLine['file'] = str_ireplace(ROOTPATH, 'ROOTPATH/', $traceLine['file']);
+                if (isset($traceLine['file'])) {
+                    $traceLine['file'] = str_ireplace(APPPATH, 'APPPATH/', $traceLine['file']);
+                    $traceLine['file'] = str_ireplace(SYSTEMPATH, 'SYSTEMPATH/', $traceLine['file']);
+                    if (defined('VENDORPATH')) {
+                        // VENDORPATH is not defined unless `vendor/autoload.php` exists
+                        $traceLine['file'] = str_ireplace(VENDORPATH, 'VENDORPATH/', $traceLine['file']);
+                    }
+                    $traceLine['file'] = str_ireplace(ROOTPATH, 'ROOTPATH/', $traceLine['file']);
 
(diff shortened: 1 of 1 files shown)


system/Debug/Toolbar/Collectors/Database.php
```diff
@@ -139,17 +139,20 @@ public function display(): array
...
-                    // VENDORPATH is not defined unless `vendor/autoload.php` exists
-                    $traceLine['file'] = str_ireplace(VENDORPATH, 'VENDORPATH/', $traceLine['file']);
+                if (isset($traceLine['file'])) {
+                    $traceLine['file'] = str_ireplace(APPPATH, 'APPPATH/', $traceLine['file']);
+                    $traceLine['file'] = str_ireplace(SYSTEMPATH, 'SYSTEMPATH/', $traceLine['file']);
+                    if (defined('VENDORPATH')) {
+                        // VENDORPATH is not defined unless `vendor/autoload.php` exists
+                        $traceLine['file'] = str_ireplace(VENDORPATH, 'VENDORPATH/', $traceLine['file']);
+                    }
+                    $traceLine['file'] = str_ireplace(ROOTPATH, 'ROOTPATH/', $traceLine['file']);
```
paulbalandan, line 149: This could use `clean_path`
```suggestion
                    $traceLine['file'] = clean_path($traceLine['file']);
```

kenjis:
@iprixi Thank you for sending this PR!

Could you add test code for this?
Set the debug trace like this: https://github.com/codeigniter4/CodeIgniter4/issues/5539#issuecomment-1005328392
and run `display()` and check the results.

mkilinskidev:
Wouldn't be better if you use `array_key_exist` instead of `isset` ?

paulbalandan:
But a null file does not make sense, is it?

iprixi:
@mkilinskidev why would we want to do a string clean up on a null value? 
+ What @paulbalandan has mentioned.
Am I missing something here?

system/Debug/Toolbar/Collectors/Database.php
iprixi: @paulbalandan I have not made any other changes and this update is specifically to address the breaking changed that was released earlier. This can be addressed later separately.

kenjis:
Screenshot 2022-01-07 14 44 40
Screenshot 2022-01-07 14 45 00

Issue #5541, opened 2022-01-04
Bug: Debug bar javascript issue with external script toggle


mihail-minkov:
### PHP Version

7.4

### CodeIgniter4 Version

4.1.6

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

apache

### What happened?

I am using Foundation 6.7 for front-end and CodeIgniter 4.1.6 for the back-end.

I thought it was an issue with a library inside Foundation till I revised what the code did. 

I basically get this error when toggling my element:

```
Uncaught TypeError: Cannot read properties of null (reading 'getAttribute')
    at HTMLAnchorElement.toggleRows (<anonymous>:145:45)
```

That is inside the ciDebugBar element in the toggleRows function:

![image](https://user-images.githubusercontent.com/4673722/148130706-4727fc84-fbe1-4456-a9cc-e752f58cd894.png)

### Steps to Reproduce

I use the off-canvas element of Foundation and a toggle button which uses the `data-toggle` button.

You can check it here https://get.foundation/sites/docs/off-canvas.html 

The smallest reproducible code would be:

```html
<!doctype html>
<html>
<head>
 <meta charset="utf-8">
 <script src="https://cdnjs.cloudflare.com/ajax/libs/foundation/6.7.4/js/foundation.min.js" integrity="sha512-+CnEgueYRDzx2s3IIAOpZd8HeOfrSbK3loeHw0jXJODAe3eFROHpBxEsun2kU+sZCaoAA0eFJrhfVV3RGG50GA==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/foundation/6.7.4/css/foundation.min.css" integrity="sha512-TgmH0v8FUwmsr3yDgd5PTCgR6lRZ2Q5c7KsUNTHcoxZpOExCX16MYECIL4xdRQOhQlz7pCnZlmA4zda58QWxBw==" crossorigin="anonymous" referrerpolicy="no-referrer" />
</head>
<body>
<div class="off-canvas position-left" id="menu-principal" data-off-canvas></div>
<div class="off-canvas-content" data-off-canvas-content>
    <main id="main-content">        
        <a data-toggle="menu-principal">Ver menu</a>
    </main>
</div>
</body>
</html>
```

### Expected Output

[...]

PR #5544, opened 2022-01-05
Fix: Debug toolbar selectors

iRedds:
**Description**
https://github.com/codeigniter4/CodeIgniter4/issues/5541

Selectors are not unique, and elements are searched throughout the document, so there might be a conflict.

``Document`` object replaced with ``toobar`` fragment

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar/Views/toolbar.js
+++ b/system/Debug/Toolbar/Views/toolbar.js
@@ -24,10 +24,10 @@ var ciDebugBar = {
 		document.getElementById('debug-icon-link').addEventListener('click', ciDebugBar.toggleToolbar, true);
 
 		// Allows to highlight the row of the current history request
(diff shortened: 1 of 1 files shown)


lonnieezell (approved):

kenjis:
Probably because of this PR, debug toolbar does not show up, if I click the icon.
Screenshot 2022-01-07 14 06 57
Screenshot 2022-01-07 14 07 18

paulbalandan:
I think the fix would be to call `ciDebugbar.init()` in the js. But I don't know where to place it.

iRedds:
@paulbalandan Calling the showTab method changes the context. (this = <a> instead of this = ciDebugBar)
Don't understand how i missed this.
Now my question is how to fix this correctly? New PR?

kenjis:
@iRedds There is no other way than new PR.

PR #3910, opened 2020-11-19
Validation: support placeholders for anything

element-code:
This fixes #3774 

**Checklist:**
- [ ] Securely signed commits
- [x] Component(s) with PHPdocs
- [x] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #5545, opened 2022-01-05
Validation: support placeholders for anything

### LINKED ISSUE #3774: Bug: Validation: Placeholder replacement not working for complex validation rules
When using the following Syntax for the rules, `Validation::fillPlaceholders()` cant find the rules and fails to replace any placeholders
```php
$rules = [
	'email' => [
		'rules' => ['required', 'trim', 'valid_email', 'max_length[191]', 'is_unique[users.email, id, {id}]'],
	]
]
```

`Validation::run()` handles this for the validation itself:
https://github.com/codeigniter4/CodeIgniter4/blob/b8ad7782ab91be361a3ec3501b1ecf0c20b99023/system/Validation/Validation.php#L172-L173

PR Incomming

paulbalandan:
**Description**
Supersedes and closes #3910 
Fixes #3774

I tried cherry-picking @element-code 's commit but ended up with merge conflicts instead, so the original authorship was lost.
The breaking change with this is the signature change of `setRule`.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/libraries/validation.rst
+++ b/user_guide_src/source/libraries/validation.rst
@@ -261,20 +261,28 @@ methods.
 setRule()
 =========
 
-This method sets a single rule. It takes the name of the field as
-the first parameter, an optional label and a string with a pipe-delimited list of rules
-that should be applied::
+This method sets a single rule. It has the method signature
 
-    $validation->setRule('username', 'Username', 'required');
+    setRule(string $field, ?string $label, array|string $rules[, array $errors = []])
 
-The **field name** must match the key of any data array that is sent in. If
+The ``$rules`` either takes in a pipe-delimited list of rules or an array collection of rules.
+
+    $validation->setRule('username', 'Username', 'required|min_length[3]');
+    $validation->setRule('password', 'Password', ['required', 'min_length[8]', 'alpha_numeric_punct']);
+
+The value you pass to ``$field`` must match the key of any data array that is sent in. If
 the data is taken directly from ``$_POST``, then it must be an exact match for
 the form input name.
+
+.. warning:: Prior to v4.2.0, this method's third parameter, ``$rules``, was typehinted to accept
+    ``string``. In v4.2.0 and after, the typehint was removed to allow arrays, too. To avoid LSP being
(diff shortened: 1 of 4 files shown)


MGatner:
This method change should be safe, right? Since all extended classes will still have `string $rules` which is LSP-compatible.

paulbalandan:
No, the parameter will no longer be contravariant since extending classes will now be typehinting a specific type (string) from the parent's  type (this PR proposing as mixed).

https://3v4l.org/h6Ehe

MGatner:
In that case we will we need additions to the upgrade guide noting the BC. Since we have a `ValidationInterface` and this method is **not** included I think this is an acceptable break. Definitely one of the things we need in version 5 is a massive pass at all classes and methods, making everything possible `private`, `final`, or `@internal` so we don't keep running into these issue.

kenjis:
LGTM except documentation.

user_guide_src/source/libraries/validation.rst
```diff
@@ -261,20 +261,28 @@ methods.
...
+The ``$rules`` either takes in a pipe-delimited list of rules or an array collection of rules.
+
+    $validation->setRule('username', 'Username', 'required|min_length[3]');
+    $validation->setRule('password', 'Password', ['required', 'min_length[8]', 'alpha_numeric_punct']);
+
+The value you pass to ``$field`` must match the key of any data array that is sent in. If
 the data is taken directly from ``$_POST``, then it must be an exact match for
 the form input name.
 
+.. warning:: Prior to v4.2.0, this method's third parameter, ``$rules``, was strictly typehinted to
```
kenjis, line 277: I think it is better to remove "strictly".
It implies [strict typing](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict) or strict mode, but we don't use it yet.

paulbalandan: Fixed

kenjis (approved):

MGatner:
🥳

PR #5554, opened 2022-01-07
Fix: Toolbar. ciDebugBar.showTab() context.

iRedds:
**Description**
Follow-up #5544
Calling the showTab method changes the context if the method is not called directly, but as an "internal call".
Rollback of previous changes.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


kenjis (approved):
It worked. Thanks!

paulbalandan (approved):

PR #5561, opened 2022-01-11
fix: Model's getCompiledUpdate() & getCompiledInsert() throws DatabaseException

### LINKED ISSUE #5549: Bug: getCompiledUpdate() & getCompiledInsert() throws DatabaseException
### PHP Version

7.4

### CodeIgniter4 Version

4.1.5

Windows, Linux

### Which server did you use?

apache

### Database

MySQL 8.0.27-0ubuntu0.20.04.1

### What happened?

Calling methods  getCompiledInsert() & getCompiledUpdate() while using MySql adapter does not work. 

### Steps to Reproduce

The below is the code that tries to retreive compiled update and compiled insert, in both cases it produces the same error:
**[CodeIgniter\Database\Exceptions\DatabaseException 8]** 
**You must use the "set" method to update an entry.**
```php
$model = new Model();
$compiledUpdate =  $model->set("test", "1")->where("id", 1)->getCompiledUpdate();
$compiledInsert =  $model->set("test", "1")->getCompiledInsert();
```

### Expected Output

In case of getCompiledUpdate() the expected string is: 
`UPDATE testTable SET id= 1 WHERE id = 1;`

In case of getCompiledInsert() the expected string is: 
`INSERT INTO testTable (test) VALUES (1);`

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5549

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -802,4 +802,60 @@ class Model extends BaseModel
 
         return $properties;
     }
+
+    /**
+     * Compiles an insert query and returns the sql
+     *
+     * @throws DatabaseException
+     *
+     * @return bool|string
+     */
+    public function getCompiledInsert(bool $reset = true)
+    {
+        $builder = $this->builder();
+
+        if (empty($this->tempData['data'])) {
+            return $builder->getCompiledInsert($reset);
+        }
+
+        $data   = $this->tempData['data'];
+        $escape = $this->tempData['escape'];
+
+        if ($reset) {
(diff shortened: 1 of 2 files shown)


lonnieezell:
These methods don't belong in the model. I understand what you're trying to fix but this isn't the way to do it. This should be fixed at the builder level.

PR #5562, opened 2022-01-11
Throws Exception when getCompiled*() is called in the Model

kenjis:
**Description**
Alternative to #5561
Related to #5549
- throws Exception when `getCompiled*()` is called in the Model

Reasons:
- `CodeIgniter\Model` and Query Builder should not be expected to return the same data. They are separate classes with different purposes.
- If you need to get the compiled Insert or Update you should do so directly on the Query Builder instance.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Language/en/Database.php
+++ b/system/Language/en/Database.php
@@ -27,5 +27,5 @@ return [
     'fieldNotExists'                   => 'Field `{0}` not found.',
     'forEmptyInputGiven'               => 'Empty statement is given for the field `{0}`',
     'forFindColumnHaveMultipleColumns' => 'Only single column allowed in Column name.',
-    'methodNotAvailable'               => 'You cannot use `{1}` in `{0}`. The CodeIgniter\Model is not the Query Builder.',
+    'methodNotAvailable'               => 'You cannot use `{1}` in `{0}`. This is a method of the `Query Builder` class.',
 ];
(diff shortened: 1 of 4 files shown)


system/Model.php
```diff
@@ -737,6 +750,13 @@ public function __call(string $name, array $params)
         return $result;
     }
 
+    private function checkBuilderMethod(string $name): void
```
lonnieezell, line 753: Not 100% sure of the current docblock guidelines, other than I know they were updated recently, but doesn't this need a docblock?

lonnieezell (approved):

system/Model.php
kenjis: Added docblock.

system/Language/en/Database.php
```diff
@@ -27,4 +27,5 @@
     'fieldNotExists'                   => 'Field `{0}` not found.',
     'forEmptyInputGiven'               => 'Empty statement is given for the field `{0}`',
     'forFindColumnHaveMultipleColumns' => 'Only single column allowed in Column name.',
+    'methodNotAvailable'               => 'You cannot use `{1}` in `{0}`. The CodeIgniter\Model is not the Query Builder.',
```
paulbalandan, line 30:
```suggestion
    'methodNotAvailable'               => 'You cannot use `{1}` in `{0}`. This is a method of the `Query Builder` class.',
```

paulbalandan (approved):

PR #5567, opened 2022-01-13
Added new View Decorators.

lonnieezell:
**Description**
Created View Decorators that allow you to modify HTML generated from the View or Parser classes just prior to it being cached. 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


system/View/Interfaces/ViewDecorator.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+namespace CodeIgniter\View\Interfaces;
+
+/**
+ * View Decorators are simple classes that are given the
+ * chance to modify the output from the view() calls
+ * prior to it being cached.
+ */
+interface ViewDecorator
```
iRedds, line 19: I think that the name of the interface should contain the suffix ``Interface``.

system/View/Traits/DecoratesViews.php
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        foreach ($decorators as $className) {
+            $decorator = new $className();
+
+            if (! $decorator instanceof ViewDecorator) {
+                throw ViewException::forInvalidDecorator($className);
+            }
+
+            $html = $decorator->decorate($html);
```
iRedds, line 35: This code can be optimized.
```suggestion
        foreach ($decorators as $decorator)
            if (! is_subclass_of($decorator, ViewDecorator::class) {
                throw ViewException::forInvalidDecorator($decorator);
            }
            
            $html = (new $decorator())->decorate($html);
```

iRedds (commented):
This proposal defines global decorators, which makes it impossible to generate a view without decorators or selectively apply them.

iRedds:
Can you give a real life example of using view decorators?

system/View/Interfaces/ViewDecorator.php
MGatner: I think the current name makes sense in the subdirectory **Interfaces**, but none of our other libraries have a dedicated `Interfaces` namespace so I would be in favor of moving this file up a directory and renaming it as @iRedds suggests.

system/View/Interfaces/ViewDecorator.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Takes $html and has a chance to alter it.
+     * MUST return the modified HTML.
+     *
+     * @param string $html
+     *
+     * @return string
+     */
+    public function decorate(string $html): string;
```
MGatner, line 29: Since the execution of this method always mounts to `(new $class())->decorate($html)` is there any reason not to make this static and just call it without ever instantiating the class?
```suggestion
    public static function decorate(string $html): string;
```

app/Config/View.php
```diff
@@ -41,4 +41,15 @@ class View extends BaseView
...
     public $plugins = [];
+
+    /**
+     * View Decorators are class methods that will be run in sequence to
+     * have a chance to alter the generated output just prior to caching
+     * the results.
+     *
+     * All classes must implement CodeIgniter\View\ViewDecorator
+     *
+     * @var array
```
MGatner, line 52: The current class is actually `CodeIgniter\View\Interfaces\ViewDecorator` but my suggestion will assume we go with the rename below. Regardless this should use the `class-string` prototype.
```suggestion
     * @var class-string<CodeIgniter\View\ViewDecoratorInterface>[]
```

system/Language/en/View.php
```diff
@@ -17,4 +17,5 @@
     'noCellClass'           => 'No view cell class provided.',
     'invalidCellClass'      => 'Unable to locate view cell class: {0}.',
     'tagSyntaxError'        => 'You have a syntax error in your Parser tags: {0}',
+    'invalidDecoratorClass' => '{0} is not a valid ViewDecorator.',
```
MGatner, line 20: If we go with the rename I think this would read better without the `Interface` suffix so I suggest making it natural:
```suggestion
    'invalidDecoratorClass' => '{0} is not a valid View Decorator.',
```

system/View/Parser.php
```diff
@@ -124,6 +127,9 @@ public function render(string $view, ?array $options = null, ?bool $saveData = n
         if ($saveData) {
             $this->data = $this->tempData;
         }
+
+        $output = $this->decorateOutput($output);
```
MGatner, line 131: Is there any reason to apply decorators to `renderString()` as well? Probably not, but worth asking.

system/View/Traits/DecoratesViews.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace CodeIgniter\View\Traits;
+
+use CodeIgniter\View\Exceptions\ViewException;
+use CodeIgniter\View\Interfaces\ViewDecorator;
+
+use function config;
+
+trait DecoratesViews
```
MGatner, line 10: I really like the Symfony-esque trait naming scheme... but this would be our first. All our other traits have boring names, à la `ViewDecoratingTrait`. I'm not opposed, just pointing it out. 

Regardless, let's make a version 5 card entry for "Trait Naming Schemes"!

system/View/Traits/DecoratesViews.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * Runs the generated output through and declared
+     * view decorators.
+     *
+     * @param string $html
+     *
+     * @return string
+     */
+    private function decorateOutput(string $html): string
```
MGatner, line 20: Still asking questions... If anyone has their own `RendererInterface` service wouldn't it likely be an extension of `View`? If so it might make sense to have this method accessible. We could accomplish the same protection with `final protected function decorateOutput()`.

system/View/Traits/DecoratesViews.php
MGatner: If we go with the static method suggestion above this can be simplified even more:
```suggestion
        foreach ($decorators as $className) {
            if (! is_subclass_of($decorator, ViewDecorator::class) {
                throw ViewException::forInvalidDecorator($className);
            }

            $html = $decorator::decorate($html);
```

system/View/View.php
```diff
@@ -134,6 +137,17 @@ class View implements RendererInterface
...
+     * View Decorators are class methods that will be run in sequence to
+     * have a chance to alter the generated output just prior to caching
+     * the results.
+     *
+     * All classes must implement CodeIgniter\View\ViewDecorator
+     *
+     * @var array
+     */
+    public $decorators = [];
+
```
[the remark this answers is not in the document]
MGatner: Ah, I think you actually meant this for `CodeIgniter\Config\View`. If you move this over to the correct class it will satisfy my comments about the Config files above.

MGatner (changes requested):
I'm super excited for this! Lots of comments, and also some CI fixes that CS Fixer/Rector ought tot take care of.

MGatner:
> Can you give a real life example of using view decorators?

@iRedds I'm excited for this because it takes some pressure off of `Filters` while still leveraging View Caching. Currently the only "middleware" solution we have for modifying the Response is to use a Filter with the `after()` method. IMO this puts too much pressure on Filters plus, as I mentioned, it means that your Filters run on the cached View content every time unnecessarily. 

Most of my "HTML injection" libraries use the Filters method, here is an example that would be replaced by a single Decorator: https://github.com/tattersoftware/codeigniter4-alerts/blob/develop/src/Filters/AlertsFilter.php

***

 One other "plus" of Decorators is they are a lot easier to configure and apply! For a third-party library to activate its Filter it needs to use one of these hack-y FIlter configs: https://github.com/tattersoftware/codeigniter4-alerts/blob/develop/src/Config/Filters.php
... *and then* the implementing project still needs to add it to the desired routes in **app/Config/Filters.php**:
```php
    public $globals = [
        'after'  => [
            'alerts' => '*',
        ],
    ];
```

MGatner:
@lonnieezell what about some built-in tools or core handling for non-HTML output? Is there any scenario where we need to be "header-aware" (like disabling Decorators for JSON content), or is that up to the developer?

For example, I add this to all my HTML-injection Filters:

https://github.com/tattersoftware/codeigniter4-alerts/blob/develop/src/Filters/AlertsFilter.php#L55-L68

MGatner:
Thinking about this more my hunch is that View doesn't care at all about formatting or headers and it would be on the developer to know that passing content through will subject it to Decorators. I suspect most APIs dealing with JSON/XML aren't using View for the content anyway.

Probably worth mentioning in the User Guide either way.

system/View/Traits/DecoratesViews.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $decorators = config('View')->decorators;
+
+        if (empty($decorators)) {
+            return $html;
+        }
```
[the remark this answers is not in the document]
lonnieezell: I did add it the system config file. Well I've corrected that since I put it in the wrong file last night. Silly me :) Will remove.

system/View/Interfaces/ViewDecorator.php
lonnieezell: Agree - will make it match the rest of them. I was even the one who initially argued the point that they should be named that way.

system/View/Traits/DecoratesViews.php
lonnieezell: I like the static method suggestion here. Will change and simplify.

lonnieezell:
> Can you give a real life example of using view decorators?

Two of them: 

I've implemented View Components in Bonfire, but they require on overriding the View class to work. Essentially they look for custom html tags and replace them with other content. 

It could also be used for a "snippet" functionality that kept small bits of text/html and will replace placeholders with it globally. I've seen this on some CMS' before.

lonnieezell:
> what about some built-in tools or core handling for non-HTML output? Is there any scenario where we need to be "header-aware" (like disabling Decorators for JSON content), or is that up to the developer?

I agree with your assessment that View doesn't care here.

system/Config/View.php
```diff
@@ -88,6 +88,17 @@ class View extends BaseConfig
...
     ];
 
+    /**
+     * View Decorators are class methods that will be run in sequence to
+     * have a chance to alter the generated output just prior to caching
+     * the results.
+     *
+     * All classes must implement CodeIgniter\View\ViewDecorator
+     *
+     * @var array
```
MGatner, line 98: This should use `class-string` to match the suggestion in **app/Config/View.php**.

MGatner (changes requested):
I assume you are still looking at some of the others. One typing change here.

Also, is there a case to be made for `View::enableDecorators()`/`View::disableDecorators()`?

system/View/Parser.php
lonnieezell: That's a good question that I don't have a good answer to. It's not used in the framework anywhere I don't believe, so I'm not sure what it actually gets used for. I'm thinking probably no for now, and we'll see if anyone needs it.

app/Config/View.php
lonnieezell: I'll change it but what's the purpose of labeling it like this?

system/View/Interfaces/ViewDecorator.php
lonnieezell: Done

lonnieezell:
Oh - @MGatner this should target 4.2, also, shouldn't it?

system/View/Traits/DecoratesViews.php
lonnieezell: It doesn't like that when it's in a trait, during testing at least. Complained about not being able to override the function. Looks like either private or protected. MOving to protected for now.

lonnieezell:
I think everything is accounted for now, though I had to skip the pre-commit checks since there are two PHPStan errors for me locally with the latest develop in `text_helper` and the `Autoloader`.

lonnieezell:
> Also, is there a case to be made for View::enableDecorators()/View::disableDecorators()?

My initial instinct is not really, but both you and IReds brought something similar up so maybe. :)

kenjis:
php-cs-fixer check failed, but it seems php-cs-fixer's `phpdoc_types_order` has a bug.

```diff
   1) CodeIgniter4/tests/system/Cookie/CookieStoreTest.php (phpdoc_types_order)
      ---------- begin diff ----------
--- /home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Cookie/CookieStoreTest.php
+++ /home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Cookie/CookieStoreTest.php
@@ -107,7 +107,7 @@
         $prod = $cookies['prod']->getOptions();
 
         /**
-         * @var MockObject&CookieStore $store
+         * @var CookieStore|MockObject $store
          */
         $store = $this->getMockBuilder(CookieStore::class)
             ->setConstructorArgs([$cookies])
```
https://github.com/codeigniter4/CodeIgniter4/runs/4812733176?check_suite_focus=true

MGatner:
> Oh - @MGatner this should target 4.2, also, shouldn't it? 

We merge 4.2 into develop! We'll see how that goes 😅

system/View/Parser.php
MGatner: I think it is for handling "snippets" like `<div class="{class"">{foo}</div>`. I've never used it myself but my hunch is that decorators don't belong and would only make that confusing. I like your "wait and see" approach.

MGatner (changes requested):
A few wording changes. And since we are `>= 7.4` now: typed properties!

app/Config/View.php
MGatner: Yep! It tells static analysis "this can only contain strings that are actually class names of this type". Pretty handy actually.

MGatner (approved):
Looks good to me! See if there are any workflow issues remaining.

MGatner:
🥳

PR #5571, opened 2022-01-13
Include CI-Upgrader in docs

FlorianNelles:
The CI-Upgrader is a tool, which will help you to upgrade your CodeIgniter 3 project to CodeIgniter 4.
It is able to run the CI4 installation, copy the most important files and directories from your old CI3 project to the new one and also to transfer most of the Code into CI4 syntax.
So I think it would be useful to include a reference to this tool on the 'Upgrading from 3.x to 4.x' -page of the documentation.

### DIFF
--- a/user_guide_src/source/installation/upgrade_4xx.rst
+++ b/user_guide_src/source/installation/upgrade_4xx.rst
@@ -155,3 +155,7 @@ Upgrading Libraries
 
 .. note::
     More upgrade guides coming soon
+    
+.. note::
+    To get a much faster and easier way to upgrade your CI3 project to CI4, you can check out the `CI-Upgrader <https://github.com/FlorianNelles/CI-Upgrader>`_.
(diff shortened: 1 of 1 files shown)


lonnieezell:
While this might be a great tool I have a couple of concerns including it in the official user guide. 

The first being that it's not an official project. We currently don't list any third-party tools in the guide, I don't believe. If we list this then it gives the appearance we can vouch for this tool. Unfortunately I've never used it and thanks to the second issue with it I never will be able to. Additionally, I don't have time currently to get to an understanding of the code base to be able to vouch for it and that it's a safe tool to use. 

The second issue is that it only works on Windows and doesn't support either MacOS or Linux. We strive to maintain feature parity across operating systems, or even different handlers for the same task. 

I know you've already shared this on the forums, which is where this is appropriate. I think the tool sounds great, and looks like it can be a huge help for developers, but I don't believe it can be added to the user guide. 

Thanks for creating it, though, and I wish you the best of luck with it.

FlorianNelles:
I can totally understand this and you have some good points there. 
Thanks for your feedback.

Issue #5576, opened 2022-01-15
Bug: MyModel->delete command returns true in any case.


datamweb:
### PHP Version

8.0

### CodeIgniter4 Version

4.1.7 

### CodeIgniter4 Installation Method

Manual (zip or tar.gz)

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

apache

### Database

MySQL 5.6

### What happened?

`protected $primaryKey       = 'id';`

In my table there is no id =1.
But by executing the following command.
`$myModel->delete(1);`
**true** is returned!!.
My God, why do you return **true** for something that does not exist?
This is if the **false** should be returned.

**_Another problem._**
When a record has already been delete by `$myModel->delete(1);`
If you run again `$myModel->delete(1);` , the deletion operation will be performed again and the deletion deleted_at will be updated and the **true**  value will be returned.
This does not make sense because it has already been deleted.
 Therefore the **false** must be returned.
Help me, is not my speech logical?

### Steps to Reproduce

```
tbl : 
id | deleted_at
 2 | 2022-01-14 20:48:58
 3 | 2022-01-14 20:48:58
 4 | null
```

`$myModel->delete(1);`
return **true** but  must be **false** returned because **id=1** does not exist in the table.
**_and_**
`$myModel->delete(2);`
Update Field **deleted_at** . Value **true** is returned. But it should not update **deleted_at**  because it has already been deleted(2022-01-14 20:48:58). **False** must be returned.

### Expected Output

If the **id=1** is not in the table and the `$myModel->delete(1);` is executed, the value of **false** is returned.
_And_
If a row has already been `$myModel->delete(1);` and if `$myModel->delete(1);` run again, it will not be updated **deleted_at** and returned **false** .

iRedds:
There is a difference between a successfully executed query and a successfully executed query that resulted in a change to the data.
Insert/update/delete queries return the execution status of the query.
To check if the rows have been changed, you can use the BaseConnection::affectedRows() method.
Since the behavior is not described in the documentation for the model, any change can be critical.

kenjis:
> My God, why do you return true for something that does not exist?

Because the query was successful.

> Returns false on failure. For successful queries which produce a result set, such as SELECT, SHOW, DESCRIBE or EXPLAIN, mysqli_query() will return a mysqli_result object. For other successful queries, mysqli_query() will return true. 
[...]

PR #5578, opened 2022-01-15
Fix: Deletion timestamp of the Model is updated when a record that has been soft-deleted is deleted again

iRedds:
**Description**
Related #5576
If the model provides for soft deletion of records and an attempt is made to re-delete, then timestamps are changed in the record marked as deleted.

This PR adds a condition under which the update will only occur if the deletion timestamp has not yet been set.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -358,7 +358,7 @@ class Model extends BaseModel
                 return false; // @codeCoverageIgnore
             }
 
-            $builder->where($this->deletedField, null);
+            $builder->where($this->deletedField);
 
             $set[$this->deletedField] = $this->setDate();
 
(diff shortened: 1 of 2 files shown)


kenjis:
Can you add test case to prove the code works?

> Note, we expect all code changes or bug-fixes to be accompanied by one or more tests added to our test suite to prove the code works. 

iRedds:
@kenjis Of all the model tests, I found only life tests.
I don't know if this will be enough.

kenjis (approved):

kenjis:
@iRedds This is a bug fix. So the workflow is like this:
1. write a test which reproduces the bug, and confirm it fails.
2. fix the bug, and confirm the test passes.

I think your test is enough.

paulbalandan (approved):

PR #5584, opened 2022-01-16
Fixing phpstan and style errors

lonnieezell:
Just cleaning up phpstan and style errors I was getting locally.

### DIFF
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -1,12 +1,12 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Parameter \\#1 \\$autoload_function of function spl_autoload_register expects callable\\(string\\)\\: void, array\\{\\$this\\(CodeIgniter\\\\Autoloader\\\\Autoloader\\), 'loadClass'\\} given\\.$#"
+			message: "#^Parameter \\#1 \\$callback of function spl_autoload_register expects \\(callable\\(string\\)\\: void\\)\\|null, array\\{\\$this\\(CodeIgniter\\\\Autoloader\\\\Autoloader\\), 'loadClass'\\} given\\.$#"
 			count: 1
 			path: system/Autoloader/Autoloader.php
 
 		-
-			message: "#^Parameter \\#1 \\$autoload_function of function spl_autoload_register expects callable\\(string\\)\\: void, array\\{\\$this\\(CodeIgniter\\\\Autoloader\\\\Autoloader\\), 'loadClassmap'\\} given\\.$#"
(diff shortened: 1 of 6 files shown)


phpstan-baseline.neon.dist
```diff
@@ -1,12 +1,12 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Parameter \\#1 \\$callback of function spl_autoload_register expects \\(callable\\(string\\)\\: void\\)\\|null, array\\{\\$this\\(CodeIgniter\\\\Autoloader\\\\Autoloader\\), 'loadClass'\\} given\\.$#"
+			message: "#^Parameter \\#1 \\$autoload_function of function spl_autoload_register expects callable\\(string\\)\\: void, array\\{\\$this\\(CodeIgniter\\\\Autoloader\\\\Autoloader\\), 'loadClass'\\} given\\.$#"
```
paulbalandan, line 4: This change should be reverted. PHPStan has different parameter names for the function based on 7.4 and 8.0, and PHPStan does not want to normalize them either. So I opted to run PHPStan on PHP 8.0+ so that the updated parameter name will be used.

paulbalandan (changes requested):

phpstan-baseline.neon.dist
kenjis: I didn't know it. It should be added in https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#static-analysis-on-php-code

lonnieezell: Ok, that makes sense. Was wondering why it was failing now that it got pushed up here. Then should we set the ignore unused error to false? The main reason I was addressing these is because using 7.4 locally it was blocking me from committing any changes.

paulbalandan: I think phpstan has a parameter option to specify the php version it should "run".

paulbalandan: phpVersion

default: null (current PHP version is used)

If you want to analyse a codebase as if it was written for a different PHP version than you’re currently running, change the phpVersion parameter:

parameters:
    phpVersion: 70400 # PHP 7.4

paulbalandan: We should set it to 80000 to match our CICD

kenjis: I reverted the ignoreErrors and added phpVersion.

kenjis: Okay, all green!

kenjis (approved):

PR #5591, opened 2022-01-19
Fix ResponseTrait methods visibility

kenjis:
**Description**
- fix ResponseTrait methods visibility. Public methods can be accessible via HTTP.
- fix `@return` types

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/API/ResponseTrait.php
+++ b/system/API/ResponseTrait.php
@@ -87,7 +87,7 @@ trait ResponseTrait
      *
      * @return mixed
      */
-    public function respond($data = null, ?int $status = null, string $message = '')
+    protected function respond($data = null, ?int $status = null, string $message = '')
     {
         if ($data === null && $status === null) {
             $status = 404;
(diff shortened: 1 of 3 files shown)


MGatner (approved):
The property type changes make me a little nervous. I think it was `mixed` because the actual Controller property will be `service('response')` which is `ResponseInterface`, however the interface is missing some methods we require `Response` to support. 

Given that the `@property` tag already has `Response` I think this is acceptable, but more as a "workaround" than an actual correction. Some more work for version 5...

kenjis:
@MGatner 
`service('response')` returns `Response`:
https://github.com/codeigniter4/CodeIgniter4/blob/904404c5fdc9cb73802f87aaa13de665318e058e/system/Config/Services.php#L492-L494

If we will make it `ResponseInterface`, do we add missing methods to ResponseInterface?
Or remove methods from Response that are not in PSR?

kenjis:
This PR is included in v4.1.8 hotfix, and merged into develop.
So I close.

PR #5596, opened 2022-01-19
BaseService - Use lowercase key in resetSingle

### LINKED ISSUE #5595: Bug: BaseService injectMock resetSingle case sensitivity
### PHP Version

7.4

### CodeIgniter4 Version

git

Git

macOS, Linux

### Which server did you use?

fpm-fcgi

### Database

None

### What happened?

In Base Service injectMock will convert key to lowercase while resetSingle will try unset the key without converting to lowercase.

https://github.com/codeigniter4/CodeIgniter4/blob/833c67f10a35ad12843d66150af915a6beb5b1e8/system/Config/BaseService.php#L286-L302

### Steps to Reproduce

```
Services::injectMock('someService', $someObject);
Services::resetSingle('someService');
```

### Expected Output

Service should be reset but instead it is still kept due to key mismatch 

### Anything else?

_No response_

najdanovicivan:
Since injectMock converts key to lowercase we need to use lowercase key as well in resetSingle method 

**Description**
Fixes #5595

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -288,6 +288,7 @@ class BaseService
      */
     public static function resetSingle(string $name)
     {
+        $name = strtolower($name);
         unset(static::$mocks[$name], static::$instances[$name]);
     }
 
--- a/tests/system/Config/ServicesTest.php
+++ b/tests/system/Config/ServicesTest.php
@@ -343,6 +343,17 @@ final class ServicesTest extends CIUnitTestCase
         $this->assertSame($security, $security2);
     }
 
+    public function testResetSingleCaseInsensitive()
+    {
+        Services::injectMock('response', new MockResponse(new App()));
+        $someService = service('response');
+        $this->assertInstanceOf(MockResponse::class, $someService);
+
+        Services::resetSingle('Response');
+        $someService = service('response');
+        $this->assertNotInstanceOf(MockResponse::class, $someService);
+    }
+
     public function testFilters()
     {
         $result = Services::filters();


kenjis:
Can you add test case to prove the code works?

> Note, we expect all code changes or bug-fixes to be accompanied by one or more tests added to our test suite to prove the code works. 

najdanovicivan:
@kenjis test is added

tests/_support/Config/Services.php
```diff
@@ -43,4 +44,20 @@ public static function uri(?string $uri = null, bool $getShared = true)
 
         return new URI($uri);
     }
+
+    /**
+     * The URI class provides a way to model and manipulate URIs.
```
kenjis, line 49: This is not updated.

tests/_support/Config/Services.php
```diff
@@ -43,4 +44,20 @@ public static function uri(?string $uri = null, bool $getShared = true)
...
+     *
+     * @param mixed $getShared
```
kenjis, line 51: This is nott `mixed`.

tests/_support/Config/Services.php
```diff
@@ -43,4 +44,20 @@ public static function uri(?string $uri = null, bool $getShared = true)
...
+     *
+     * @return stdClass
+     */
+    public static function someService($getShared = true)
```
kenjis, line 55: I think our convention is that a service name is all lower case.
This should be `someservice`.

tests/_support/Config/Services.php
```diff
@@ -43,4 +44,20 @@ public static function uri(?string $uri = null, bool $getShared = true)
...
+    {
+        if ($getShared) {
+            return static::getSharedInstance('someService');
+        }
+
+        return new stdClass();
+    }
```
kenjis, line 62: By the way, why do you need to add a service?
We can use the existing service for testing.

system/Config/BaseService.php
```diff
@@ -288,7 +288,7 @@ public static function reset(bool $initAutoloader = false)
      */
     public static function resetSingle(string $name)
     {
-        unset(static::$mocks[$name], static::$instances[$name]);
+        unset(static::$mocks[strtolower($name)], static::$instances[strtolower($name)]);
```
kenjis, line 291: Adding `$name = strtolower($name);` above is better?

tests/_support/Config/Services.php
najdanovicivan: I did not know about that convention but there is nothing limiting the user not to use any letter casing.

Function names in PHP are case insensitive so it will work with any case.  And it make sense to me to use camel case for service names

najdanovicivan:
Ok I made this very simple now. I though we should have some real case example with non lowercase name and that is the reason why I added `someService()`

But since there is a convention for lowercase in service method there is no need for that. We just need to make the method do to lowercase conversion to match how other methods in Service classes behave.

tests/_support/Config/Services.php
kenjis: This must be lowercase.
https://github.com/codeigniter4/CodeIgniter4/blob/812af1e6285c352ae9b4da044fd7c1a0e8815d90/system/Config/Services.php#L180

kenjis (approved):
Thank you!

As you say, a user may use camelCase, 
but I think this test case is enough.

paulbalandan (approved):

MGatner (approved):
Since `serviceExists()` uses lowercase only for method lookup I think all services must be lower unless the user is calling them directly from `Config\Services`. 

This is a good catch and fix.

PR #5604, opened 2022-01-21
chore: generate coverage report by PHPUnit only on PHP 8.0

kenjis:
**Description**
- To reduce test time
  - add `--no-coverage` option to PHPUnit other than on PHP 8.0

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -126,7 +126,15 @@ jobs:
         if: matrix.php-versions == '8.0'
         run: echo "TACHYCARDIA_MONITOR_GA=enabled" >> $GITHUB_ENV
 
-      - name: Test with PHPUnit
+      - name: Test with PHPUnit w/o Coverage
+        if: matrix.php-versions != '8.0'
+        run: script -e -c "vendor/bin/phpunit --color=always --exclude-group=auto-review --no-coverage"
+        env:
+          DB: ${{ matrix.db-platforms }}
+          TERM: xterm-256color
+
+      - name: Test with PHPUnit with Coverage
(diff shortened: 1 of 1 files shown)


.github/workflows/test-phpunit.yml
```diff
@@ -126,7 +126,15 @@ jobs:
...
+      - name: Test with PHPUnit w/o Coverage
+        if: matrix.php-versions != '8.0'
+        run: script -e -c "vendor/bin/phpunit --color=always --exclude-group=auto-review --no-coverage"
+        env:
+          DB: ${{ matrix.db-platforms }}
+          TERM: xterm-256color
+
+      - name: Test with PHPUnit with Coverage
+        if: matrix.php-versions == '8.0'
         run: script -e -c "vendor/bin/phpunit --color=always --exclude-group=auto-review"
```
paulbalandan, line 138: I think it may be better to use GHA flags instead of duplicating. Something like:
```suggestion
      - name: Compute coverage option
        uses: actions/github-script@v3
        id: phpunit-coverage-option
        with:
          script: 'return "${{ matrix.php-versions }}" == "8.0" ? "--no-coverage" : ""'
          result-encoding: string

      - name: Test with PHPUnit
        run: script -e -c "vendor/bin/phpunit --color=always --exclude-group=auto-review ${{ steps.phpunit-coverage-option.outputs.result }}"
```

kenjis:
`PHPUnit / PHP 7.4 & 8.1` testing finished in 3m to 6m!

paulbalandan (approved):

PR #5609, opened 2022-01-23
Fix: Validation. Error key for field with asterisk

iRedds:
**Description**
Now, when validating a multidimensional array using a wildcard, only the last error is saved. Therefore, it is impossible to determine which field the error belongs to.

```php
// data
[
    'users' => [
        'friends' => [
            ['name' => 'Jim'],
            ['name' => 'James'],
            ['name' => ''],
        ]
    ]
]

// rules
['users.*.name' => 'required|min_length[4]']

// errors
['users.*.name' => 'The users.*.name field is required.']
```

This PR changes behavior. Errors have a corresponding key.
```php
[
    'users.friends.0.name' => 'The users.*.name field must be at least 4 characters in length.',
    'users.friends.2.name' => 'The users.*.name field is required.',
]
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/libraries/validation.rst
+++ b/user_guide_src/source/libraries/validation.rst
@@ -634,7 +634,7 @@ When using a wildcard, the error will point to a specific field, replacing the a
         ]
     ]
 
-    //rule
+    // rule
     contacts.*.name => 'required'
 
     // error will be
@@ -658,6 +658,16 @@ You can check to see if an error exists with the ``hasError()`` method. The only
     if ($validation->hasError('username')) {
         echo $validation->getError('username');
     }
+
+When specifying a field with a wildcard, all errors matching the mask will be checked.::
+
+    // for errors
+    [
+        'foo.0.bar'   => 'Error',
+        'foo.baz.bar' => 'Error',
+    ]
+
+    $validation->hasError('foo.*.bar'); // return true
 
 Customizing Error Display
 *************************
--- a/user_guide_src/source/libraries/validation.rst
+++ b/user_guide_src/source/libraries/validation.rst
@@ -603,6 +603,8 @@ We can simply use the language lines defined in this file, like this::
         ]
     );
 
+.. _validation-getting-all-errors:
+
 Getting All Errors
 ==================
 
--- a/user_guide_src/source/libraries/validation.rst
+++ b/user_guide_src/source/libraries/validation.rst
@@ -650,6 +650,8 @@ name::
 
 If no error exists, an empty string will be returned.
 
+.. note:: When using a wildcard, all found errors that match the mask will be combined into one line separated by the EOL character.
+
 Check If Error Exists
 =====================
 
--- a/user_guide_src/source/libraries/validation.rst
+++ b/user_guide_src/source/libraries/validation.rst
@@ -618,6 +618,26 @@ If you need to retrieve all error messages for failed fields, you can use the ``
 
 If no errors exist, an empty array will be returned.
 
+When using a wildcard, the error will point to a specific field, replacing the asterisk with the appropriate key/keys.::
+
+    // for data
+    'contacts' => [
+        'friends' => [
+            [
+                'name' => 'Fred Flinstone',
+            ],
+            [
+                'name' => '',
+            ],
+        ]
+    ]
+
+    //rule
+    contacts.*.name => 'required'
+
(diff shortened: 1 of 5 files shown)


paulbalandan (approved):
This looks good to me.

kenjis:
The error keys (and values) will be changed. It should be documented in changelog.

iRedds:
@kenjis added

user_guide_src/source/libraries/validation.rst
```diff
@@ -618,6 +620,26 @@ If you need to retrieve all error messages for failed fields, you can use the ``
...
+            [
+                'name' => 'Fred Flinstone',
+            ],
+            [
+                'name' => '',
+            ],
+        ]
+    ]
+
+    //rule
```
kenjis, line 637:
```suggestion
    // rule
```

kenjis:
Isn't this PR a breaking change?
At least, the output of `getError()` and `hasError()` will change, because the key will change.
If a user has code like `if ($this->validation->hasError('id_user.*')) { ... }`, it breaks.

The following test will fail.

```diff
--- a/tests/system/Validation/ValidationTest.php
+++ b/tests/system/Validation/ValidationTest.php
@@ -851,6 +851,10 @@ final class ValidationTest extends CIUnitTestCase
             'name_user.2'             => 'The name_user.* field may only contain alphabetical characters.',
             'contacts.friends.0.name' => 'The contacts.*.name field is required.',
         ], $this->validation->getErrors());
+        $this->assertSame(
+            'The id_user.* field must contain only numbers.',
+            $this->validation->getError('id_user.*')
+        );
     }
 
     public function testRulesForSingleRuleWithSingleValue(): void
```

kenjis:
How about these?
- `hasError('users.*.name')` returns `true` as before.
- `getError('users.*.name')` returns something like `"The users.*.name field must be at least 4 characters in length.\nThe users.*.name field is required."`.

iRedds:
I'm not sure about the behavior of ``getError('users.*.name')``
@paulbalandan Can you express your opinion?

paulbalandan:
Technically, this is a _potentially_ breaking change. However, I believe the chances this change will break apps is low. So, this must be disclosed in the changelog with a caveat for those using `getError('users.*.name'')` to check their code.

kenjis:
If I want to check the result, do I write code like this?
```php
if (
    $this->hasError(''users.friends.0.name')
    || $this->hasError(''users.friends.1.name')
    || $this->hasError(''users.friends.2.name')
) {
    ...
}
```

iRedds:
@kenjis Wait, I was only talking about getError().

kenjis (approved):

paulbalandan (approved):

kenjis:
@iRedds Can you rebase and resolve conflicts?

iRedds:
@kenjis Yes i can

kenjis:
@iRedds Thanks!

kenjis:
@iRedds @paulbalandan 

Now the error array is like this:
```php
[
    'users.friends.0.name' => 'The users.*.name field must be at least 4 characters in length.',
    'users.friends.2.name' => 'The users.*.name field is required.',
]
```
But would it be easier to use?
```php
[
    'users.*.name' => [
        0 => 'The users.*.name field must be at least 4 characters in length.',
        2 => 'The users.*.name field is required.',
    ]
]
```
What do you think?

iRedds:
@kenjis Here we lose the ``friends`` key.
What will happen for such a data structure?
```php
'users' => [
    'friends' => [
        0 => [
            'name' => 'correct',
        ],
    ],
    'enemies' => [
        0 => [
            'name' => null,
        ],
    ],
]
```
Although my example looks like an attempt to pull an owl on a globe.
But it seems to me that such an option should be envisaged.

kenjis:
I don't know we need to support such a use case, but then:

```php
[
    'users.*.name' => [
        'friends' => [ 
            0 => 'The users.*.name field must be at least 4 characters in length.',
            2 => 'The users.*.name field is required.',
        ],
    ]
]
```

iRedds:
Since I don't use CI. It's hard for me to imagine how it would be convenient for a real project to provide errors.
Wouldn't that make it difficult to get errors?

PR #5610, opened 2022-01-24
deprecate Database\SQLSRV\Connection::getError()

kenjis:
**Description**
- The method is a duplicate of `error()`.
- `BaseConnection` does not have `getError()`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1526,7 +1526,8 @@ abstract class BaseConnection implements ConnectionInterface
      *
      * Must return an array with keys 'code' and 'message':
      *
-     * @return array{code: int|string|null, message: string|null}
+     * @return array<string, int|string|null>
+     * @phpstan-return array{code: int|string|null, message: string|null}
      */
     abstract public function error(): array;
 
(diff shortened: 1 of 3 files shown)


paulbalandan (approved):

system/Database/BaseConnection.php
```diff
@@ -1526,7 +1526,7 @@ public function isWriteType($sql): bool
      *
      * Must return an array with keys 'code' and 'message':
      *
-     *  return ['code' => null, 'message' => null);
+     * @return array{code: int|string|null, message: string|null}
```
MGatner, line 1529: Is this legit doc code with VS Studio, etc? @paulbalandan

paulbalandan: Good catch. VS Code does not understand array shapes yet.
```suggestion
     * @return array<string, int|string|null>
     * @phpstan-return array{code: int|string|null, message: string|null}
```

paulbalandan (approved):

PR #5630, opened 2022-01-29
Rename docblocks in Routes utilities classes

paulbalandan:
**Description**
Extensions in Visual Studio Code do not have access yet to advanced doc blocks like `class-string`, `list`, `T`, causing type error hints when files are opened. I suggest changing them into the conventional formats.

However, for SA, phpstan uderstands them so I just swapped them to the compatible tags (like `@phpstan-return`).

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Commands/Utilities/Routes/AutoRouteCollector.php
+++ b/system/Commands/Utilities/Routes/AutoRouteCollector.php
@@ -36,6 +36,7 @@ final class AutoRouteCollector
 
     /**
      * @return array<int, array<int, string>>
+     * @phpstan-return list<list<string>>
      */
     public function get(): array
     {
--- a/system/Commands/Utilities/Routes/AutoRouteCollector.php
+++ b/system/Commands/Utilities/Routes/AutoRouteCollector.php
@@ -35,7 +35,7 @@ final class AutoRouteCollector
     }
 
     /**
-     * @return list<list<string>>
+     * @return array<int, array<int, string>>
      */
     public function get(): array
     {
--- a/system/Commands/Utilities/Routes/ControllerMethodReader.php
+++ b/system/Commands/Utilities/Routes/ControllerMethodReader.php
@@ -36,6 +36,7 @@ final class ControllerMethodReader
      * @phpstan-param class-string $class
      *
      * @return array<int, array{route: string, handler: string}>
+     * @phpstan-return list<array{route: string, handler: string}>
      */
     public function read(string $class, string $defaultController = 'Home', string $defaultMethod = 'index'): array
     {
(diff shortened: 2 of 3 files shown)


system/Commands/Utilities/Routes/AutoRouteCollector.php
```diff
@@ -35,7 +35,7 @@ public function __construct(string $namespace, string $defaultController, string
     }
 
     /**
-     * @return list<list<string>>
+     * @return array<int, array<int, string>>
```
kenjis, line 38: PHPStan does not fully support `list`. However, there should be no error.
https://github.com/phpstan/phpstan/issues/6358#issuecomment-1012168140

MGatner (approved):
Elegant solution. How does this affect CS Fixer?

system/Commands/Utilities/Routes/AutoRouteCollector.php
paulbalandan: Hmm. I did not know that for `list`.

paulbalandan:
> Elegant solution. How does this affect CS Fixer?

php-cs-fixer can still process these, with the different `phpdoc_**` rules.

sfadschm:
Just a feeling, but could we maybe provide a list of "fully" compatible IDEs for development? It seems to me that some adjustments for editors in recent time 
cause more trouble than they resolve (see the phpstorm auto-completion issue) or make workarounds necessary, just because some IDE does not (yet) fully support all docblock Features.

This is not meant against the changes here, they look reasonable, just a general idea 😄

system/Commands/Utilities/Routes/AutoRouteCollector.php
kenjis, line 38:
```suggestion
     * @return array<int, array<int, string>>
     * @phpstan-return list<list<string>>
```

system/Commands/Utilities/Routes/ControllerMethodReader.php
```diff
@@ -33,9 +33,9 @@ public function __construct(string $namespace)
     }
 
     /**
-     * @param class-string $class
+     * @phpstan-param class-string $class
      *
-     * @return list<array{route: string, handler: string}>
+     * @return array<int, array{route: string, handler: string}>
```
kenjis, line 38:
```suggestion
     * @return array<int, array{route: string, handler: string}>
     * @phpstan-return list<array{route: string, handler: string}>
```

MGatner (approved):

kenjis (approved):

PR #5631, opened 2022-01-31
docs: fix dbmgmt/migration.rst

### LINKED ISSUE #5611: Doc: Same migration class name
### PHP Version

7.3

### CodeIgniter4 Version

latest

macOS

### Which server did you use?

apache

### Database

_No response_

### What happened?

It is not possible to apply a migration if a migration with the same class name was already called at the time the migrations were applied.

Example. You create migrations sequentially.
create_users_table, ``change_users_table``, create_posts_table, ``change_users_table``.
Now you want to apply them to the new database.
And you get an error.

### Steps to Reproduce

> php spark make:migration test
> php spark make:migration test
> php spark migrate

Cannot declare class App\Database\Migrations\Test, because the name is already in use

### Expected Output

It is not possible to fix this in the current way.
It is only possible to clarify in the documentation that the name of the migration must be unique.

### Anything else?

It seems to me a practical solution would be to use anonymous classes.

```php
<?php

return new class extends Migration { /*...*/}
```
This solution is not affected by the class name error and also removes the redundant code for parsing the migration file.

I understand that this is currently impossible to implement, so I propose to add a task to the v5 roadmap.

kenjis:
**Description**
Fixes #5611
- update Migration filename and classname
- add note on Migration classname uniqueness
- fix tablename

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/user_guide_src/source/dbmgmt/migration.rst
+++ b/user_guide_src/source/dbmgmt/migration.rst
@@ -83,8 +83,8 @@ as **2022-01-31-013057_AddBlog.php**.
 The database connection and the database Forge class are both available to you through
 ``$this->db`` and ``$this->forge``, respectively.
 
-Alternatively, you can use a command-line call to generate a skeleton migration file. See
-below for more details.
+Alternatively, you can use a command-line call to generate a skeleton migration file.
+See **make:migration** in :ref:`command-line-tools` for more details.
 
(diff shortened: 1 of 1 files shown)


MGatner (approved):

samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #5636, opened 2022-02-01
Feature: Request::isMethod() to test the HTTP method.

iRedds:
**Description**
This PR suggests introducing a helper method to check the current HTTP request. This simplifies the code for checking the request method.

```php
$request->isMethod('post'); 
// instead of 
$request->getMethod() === 'post'; 
```
Part of the core has also been reworked. 
The ``Request::getMethod()`` method now pre-sets the value of the ``Request::$method`` property if it has not been previously defined, considering method spoofing. The request method definition has been removed from the Request class constructor, and the ``CodeIgniter::spoofRequestMethod()`` method has been marked deprecated and is no longer called in ``CodeIgniter::run()``. Because it doesn't make sense anymore.

Removed the pointless check for ``Request::getMethod()`` in the constructor of the Router class when setting the request method on a collection of routes.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -297,8 +297,6 @@ class CodeIgniter
 
         $this->forceSecureAccess();
 
-        $this->spoofRequestMethod();
-
         Events::trigger('pre_system');
 
         // Check for a cached page. Execution will stop
@@ -972,6 +970,8 @@ class CodeIgniter
     /**
      * Modifies the Request Object to use a different method if a POST
      * variable called _method is found.
+     *
+     * @deprecated 4.2
      */
     public function spoofRequestMethod()
     {
--- a/system/HTTP/CLIRequest.php
+++ b/system/HTTP/CLIRequest.php
@@ -43,8 +43,10 @@ class CLIRequest extends Request
 
     /**
      * Set the expected HTTP verb
+     *
+     * @var string
      */
-    protected string $method = 'cli';
+    protected $method = 'cli';
 
     /**
      * Constructor
--- a/system/HTTP/CLIRequest.php
+++ b/system/HTTP/CLIRequest.php
@@ -43,10 +43,8 @@ class CLIRequest extends Request
 
     /**
      * Set the expected HTTP verb
-     *
-     * @var string
      */
-    protected $method = 'cli';
+    protected string $method = 'cli';
 
     /**
      * Constructor
--- a/system/HTTP/Request.php
+++ b/system/HTTP/Request.php
@@ -31,10 +31,8 @@ class Request extends Message implements MessageInterface, RequestInterface
 
     /**
      * Request method.
-     *
-     * @var string
      */
-    protected $method;
+    protected string $method;
 
     /**
      * A URI instance.
@@ -56,10 +54,6 @@ class Request extends Message implements MessageInterface, RequestInterface
          * @deprecated $this->proxyIps property will be removed in the future
          */
         $this->proxyIPs = $config->proxyIPs;
-
-        if (empty($this->method)) {
-            $this->method = $this->getServer('REQUEST_METHOD') ?? 'GET';
-        }
 
         if (empty($this->uri)) {
             $this->uri = new URI();
@@ -92,7 +86,17 @@ class Request extends Message implements MessageInterface, RequestInterface
      */
     public function getMethod(bool $upper = false): string
     {
-        return ($upper) ? strtoupper($this->method) : strtolower($this->method);
+        if (empty($this->method)) {
+            $this->method = $this->getServer('REQUEST_METHOD') ?? 'GET';
+
+            if ($this->method === 'POST'
+                && in_array(strtoupper($_POST['_method'] ?? ''), ['PUT', 'PATCH', 'DELETE'], true)
+            ) {
+                $this->method = strtoupper($_POST['_method']);
+            }
+        }
+
+        return $upper ? strtoupper($this->method) : strtolower($this->method);
     }
 
     /**
--- a/system/HTTP/Request.php
+++ b/system/HTTP/Request.php
@@ -31,8 +31,10 @@ class Request extends Message implements MessageInterface, RequestInterface
 
     /**
      * Request method.
+     *
+     * @var string
      */
-    protected string $method;
+    protected $method;
 
     /**
      * A URI instance.
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -124,7 +124,7 @@ class Router implements RouterInterface
         $this->controller = $this->collection->getDefaultController();
         $this->method     = $this->collection->getDefaultMethod();
 
-        $this->collection->setHTTPVerb($request->getMethod() ?? strtolower($_SERVER['REQUEST_METHOD']));
+        $this->collection->setHTTPVerb($request->getMethod());
     }
 
     /**
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -731,11 +731,6 @@ parameters:
 			path: system/Router/Router.php
 
 		-
-			message: "#^Expression on left side of \\?\\? is not nullable\\.$#"
-			count: 1
-			path: system/Router/Router.php
-
-		-
 			message: "#^Method CodeIgniter\\\\Router\\\\RouteCollectionInterface\\:\\:getRoutes\\(\\) invoked with 1 parameter, 0 required\\.$#"
 			count: 2
 			path: system/Router/Router.php
--- a/system/HTTP/RequestTrait.php
+++ b/system/HTTP/RequestTrait.php
@@ -332,4 +332,9 @@ trait RequestTrait
                 break;
         }
     }
+
+    public function isMethod(string $method): bool
+    {
+        return $this->getMethod(true) === strtoupper($method);
+    }
 }
--- a/tests/system/HTTP/RequestTest.php
+++ b/tests/system/HTTP/RequestTest.php
@@ -679,4 +679,22 @@ final class RequestTest extends CIUnitTestCase
         $this->assertSame('get', $this->request->getMethod());
         $this->assertSame('GET', $this->request->getMethod(true));
     }
+
+    public function testIsMethod()
+    {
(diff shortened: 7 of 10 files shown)


iRedds:
I'm not sure about the analysis.
Should I change phpstan-baseline.neon.dist?

kenjis:
> Should I change phpstan-baseline.neon.dist?

Yes to removing to suppress errors.
No to adding to suppress errors.

kenjis:
In my local, removing one iIgnored error pattern makes PHPStan pass.
```
(feature/request-is-method *$)$ git diff
diff --git a/phpstan-baseline.neon.dist b/phpstan-baseline.neon.dist
index 46aa7e5d1..a51b77b72 100644
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -740,11 +740,6 @@ parameters:
            count: 1
            path: system/Router/Router.php
 
-       -
-           message: "#^Expression on left side of \\?\\? is not nullable\\.$#"
-           count: 1
-           path: system/Router/Router.php
-
        -
            message: "#^Method CodeIgniter\\\\Router\\\\RouteCollectionInterface\\:\\:getRoutes\\(\\) invoked with 1 parameter, 0 required\\.$#"
            count: 2
```
```
(feature/request-is-method *$)$ composer analyze
> phpstan analyse
Note: Using configuration file /Users/kenji/work/codeigniter/CodeIgniter4/phpstan.neon.dist.
 423/423 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                                                  
 [OK] No errors                                                     
```

system/HTTP/Request.php
```diff
@@ -92,7 +86,17 @@ public function isValidIP(?string $ip = null, ?string $which = null): bool
      */
     public function getMethod(bool $upper = false): string
     {
-        return ($upper) ? strtoupper($this->method) : strtolower($this->method);
+        if (empty($this->method)) {
+            $this->method = $this->getServer('REQUEST_METHOD') ?? 'GET';
+
+            if ($this->method === 'POST'
+                && in_array(strtoupper($_POST['_method'] ?? ''), ['PUT', 'PATCH', 'DELETE'], true)
```
lonnieezell, line 93: I'm not sure we need to limit the method here. From what I can find custom methods are allowed via the HTTP 1.1 specification, though not really encouraged. WebDav, for instance, has their own custom methods, I believe. 

If we are to limit, we need to include all of those that are in the spec, including HEAD, OPTIONS, TRACE, and CONNECT.  I'm interested to hear what others on the team have to say.

lonnieezell (changes requested):
In general, I like the addition. Just had the one concern about method limiting that I'd like to get other's responses on.

system/HTTP/Request.php
iRedds: AFAIK the HEAD, OPTIONS, TRACE, and CONNECT methods don't send any data in the message body. That's why I didn't include them in the test.

samsonasik:
@iRedds please rebase

kenjis:
First of all, can we add a method that does not defined in PSR-7?
https://www.php-fig.org/psr/psr-7/

iRedds:
What method are you talking about?

lonnieezell:
> First of all, can we add a method that does not defined in PSR-7?

We don't currently adhere to PSR-7, though I know that's been a discussion topic for future releases.

system/HTTP/Request.php
lonnieezell: Correct - but POST requests can have empty bodies so I see no reason not to support all of the standard methods. Otherwise we're needlessly limiting the available options.

kenjis:
The `Controller` depends on `RequestInterface`:
https://github.com/codeigniter4/CodeIgniter4/blob/6706388a214c31f9751d160cf0820a1df4d9f244/system/Controller.php#L74

If our RequestInterface will be PSR-7 ServerRequestInterface,
adding this method does not make sense, or this method should be removed in the next major version.

iRedds:
ServerRequestInterface extends RequestInterface.
An interface defines methods that SHOULD exist, not ONLY methods defined in an interface.

kenjis:
@iRedds What do you mean?
I searched `isMethod` in PSR-7, and it was not found.

iRedds:
Interfaces describe mandatory methods. They do not prohibit adding any other methods.
None of the RequestTrait methods are specified in PSR-7. The Request::isValidIP() method is not specified in PSR-7.

system/HTTP/Request.php
MGatner: Right, you can still make a "POST OPTIONS" submission without content, and we need to be able to support that. I think leaving this open is the way to go. Nobody is going to submit a `_method` field without having some idea of what it does.

system/Router/Router.php
```diff
@@ -124,7 +124,7 @@ public function __construct(RouteCollectionInterface $routes, ?Request $request
         $this->controller = $this->collection->getDefaultController();
         $this->method     = $this->collection->getDefaultMethod();
 
-        $this->collection->setHTTPVerb($request->getMethod() ?? strtolower($_SERVER['REQUEST_METHOD']));
```
MGatner, line 127: I don't see how this was ever correct 🤦‍♂️ Thanks for fixing!

MGatner (changes requested):
A few comments... Also please add the deprecation to the User Guide changelogs.

system/HTTP/Request.php
iRedds: The check I suggested is based on the documentation for the spoofing method, where PUT and DELETE methods are given as an example.
In any case, checking the spoofing method should be.

system/HTTP/CLIRequest.php
```diff
@@ -43,10 +43,8 @@ class CLIRequest extends Request
 
     /**
      * Set the expected HTTP verb
-     *
-     * @var string
      */
-    protected $method = 'cli';
+    protected string $method = 'cli';
```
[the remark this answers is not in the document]
iRedds: The property in docBlock is marked as a string. And the framework treats it like a string. How can it be BC? Can I have an example?

system/CodeIgniter.php
```diff
@@ -967,6 +965,8 @@ public function storePreviousURL($uri)
     /**
      * Modifies the Request Object to use a different method if a POST
      * variable called _method is found.
+     *
+     * @deprecated 4.2
```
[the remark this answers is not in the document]
iRedds: Yes, because of the handling of the spoofing method inside Request::getMethod() , there is no point in this method. But I left it, otherwise it will suddenly be BC.

system/HTTP/CLIRequest.php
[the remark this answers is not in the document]
samsonasik: that's bc break if re-defined in child https://3v4l.org/iZv2i

system/HTTP/Request.php
MGatner: I'm not an HTTP expert but at the end of the day if we weren't enforcing a method filter on spoofing previously then we shouldn't add it now. If there are legitimate uses of custom methods out there then this would break them. If you have compelling evidence for this to be a bug or a security issue then please submit it; otherwise please remove this for now, optionally submitting an issue to revisit for version 5.

kenjis: All user (may be an attacker) input should be validated.

MGatner: Validated or contained. In this case we may want to add a string regex filter `[A-Za-z]` or something more, but as long as the only place this is going is into the Request method I don't think there is a problem with leaving it up to user input. If someone wants to use POSTMAN to probe alternate route methods that's no different than some developer wanting to write an API based on methods named after fruits or Malazan characters.

iRedds: The framework supports routes based on HTTP methods (most). Until the main class for registering routes is extended, there can be no talk of any fruit methods.
POSTMAN supports custom HTTP methods.
The spoofing method was invented in order to send data at the POST request level for those HTTP methods that HTML form and PHP do not support.
You will not be able to submit form data via a PUT/PATCH/DELETE request. And PHP can't get that kind of data natively, only $_GET or $_POST.
This, and this alone, is the point of the spoofing method.

Can you give an example of why you need to send HEAD, OPTIONS, (TRACE, CONNECT - not supported by the framework) via an HTML form.

By the way, CSRF check is also tied to HTTP methods.

kenjis: I think there is no use case that sends requests other than methods `PUT/PATCH/DELETE` via HTML Form (method spoofing). The current behavior is a bug, and should be fixed.

kenjis: > If the form's method is not GET or POST, but PUT, PATCH or DELETE, Symfony will insert a hidden field with the name _method that stores this method. The form will be submitted in a normal POST request, but [Symfony's routing](https://symfony.com/doc/6.1/routing.html) is capable of detecting the _method parameter and will interpret it as a PUT, PATCH or DELETE request.
https://symfony.com/doc/6.1/forms.html#changing-the-action-and-http-method

> HTML forms do not support PUT, PATCH, or DELETE actions. So, when defining PUT, PATCH, or DELETE routes that are called from an HTML form, you will need to add a hidden _method field to the form. The value sent with the _method field will be used as the HTTP request method:
https://laravel.com/docs/8.x/routing#form-method-spoofing

> 'put', 'delete', 'patch' - Will override the HTTP method with PUT, DELETE or PATCH respectively, when the form is submitted.
https://book.cakephp.org/4/en/views/helpers/form.html#options-for-form-creation

kenjis: > as long as the only place this is going is into the Request method I don't think there is a problem with leaving it up to user input. 

It is used outside of the Request object.

> If someone wants to use POSTMAN to probe alternate route methods that's no different than some developer wanting to write an API based on methods named after fruits or Malazan characters.

CI4 Router does not support custom HTTP methods.
Auto-routing just ignores or does not consider HTTP methods.

iRedds: I'll wait

PR #5639, opened 2022-02-01
feat: add Controller::validateData()

kenjis:
**Description**
- `$this->validate()` can't validate arbitrary data
  - e.g. method parameters, trimmed data
- This method can validate any user supplied data

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Controller.php
+++ b/system/Controller.php
@@ -146,12 +146,13 @@ class Controller
      * @param array        $data     The data to validate
      * @param array|string $rules
      * @param array        $messages An array of custom error messages
+     * @param string|null  $dbGroup  The database group to use
      */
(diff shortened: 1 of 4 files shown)


iRedds:
With all due respect, I think it's overkill.
Let me explain my position.
In general, I think that validation should be moved to a service / model (I'm not talking about validation before changing it in the database, although I generally think this is pointless).

MGatner (approved):
This seems a bit unnecessary to me as it just saves a step over using the Validation Service, but if you think it would be beneficial it does synergize nicely with `validate()` so I'm not opposed.

kenjis:
Added method param, documentation.

MGatner (approved):

PR #5643, opened 2022-02-02
debugbar logging requests using microtime

maniaba:
Feature request from the forum: https://forum.codeigniter.com/thread-81113.html

**Description**

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


kenjis:
@maniaba Thank you for sending the PR.
You forgot GPG signing.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/signing.md

It ssems there are coding style violations and Static Analysis error. 
See
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#php-style
- https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#static-analysis-on-php-code

maniaba:
> @maniaba Thank you for sending the PR. You forgot GPG signing. See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/signing.md
> 
> It ssems there are coding style violations and Static Analysis error. See
> 
> * https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#php-style
> * https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#static-analysis-on-php-code

I just made changes.

kenjis:
Your sign is unverified.
> The email in this signature doesn’t match the committer email.

system/Debug/Toolbar/Collectors/History.php
```diff
@@ -56,10 +58,10 @@ class History extends BaseCollector
     /**
      * Specify time limit & file count for debug history.
      *
-     * @param int $current Current history time
-     * @param int $limit   Max history files
+     * @param float $current Current history time
+     * @param int   $limit   Max history files
      */
-    public function setFiles(int $current, int $limit = 20)
+    public function setFiles(float $current, int $limit = 20)
```
sfadschm, line 64: Do we need to break the signature here? I think we could just pass around the  `microtime`  as an `int` and then do `number_format` right before outputting.

system/Debug/Toolbar.php
```diff
@@ -374,7 +374,7 @@ public function prepare(?RequestInterface $request = null, ?ResponseInterface $r
             helper('filesystem');
 
             // Updated to time() so we can get history
```
paulbalandan, line 376: This comment is no longer relevant

system/Debug/Toolbar.php
```diff
@@ -374,7 +374,7 @@ public function prepare(?RequestInterface $request = null, ?ResponseInterface $r
...
-            $time = time();
+            $time = number_format(microtime(true), 4, '.', '');
```
paulbalandan, line 377: Do you really need the 4 decimals?

system/Debug/Toolbar/Collectors/History.php
```diff
@@ -81,14 +83,21 @@ public function setFiles(int $current, int $limit = 20)
 
             $contents = @json_decode($contents);
             if (json_last_error() === JSON_ERROR_NONE) {
-                preg_match_all('/\d+/', $filename, $time);
-                $time = (int) end($time[0]);
+                preg_match_all('/debugbar_(.*)\.json/s', $filename, $time, PREG_SET_ORDER);
+                $time = number_format((float) end($time[0]), 4, '.', '');
```
paulbalandan, line 87: This can be simplified to use `preg_match`

system/Debug/Toolbar/Collectors/History.php
paulbalandan: This is BC break.

system/Debug/Toolbar/Collectors/History.php
```diff
@@ -81,14 +83,21 @@ public function setFiles(int $current, int $limit = 20)
...
+
+                // Prevent throw exception is old debugbar time format
+                if ($dateTime = DateTime::createFromFormat('U.u', $time)) {
```
paulbalandan, line 90: Remove the assignment from the if block.

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
```diff
@@ -0,0 +1,90 @@
...
+                    'statusCode'  => 200,
+                    'contentType' => 'text/html; charset=UTF-8',
+                ],
+            ],
+            'method' => 'get',
+            'url'    => 'localhost',
+            'isAJAX' => false,
+        ];
+        // create 10 dummy debugbar json files
+        for ($i = 0; $i < 20; $i++) {
```
paulbalandan, line 49: Wrong comment or wrong loop count

system/Debug/Toolbar.php
maniaba: It is important to note that microtime(TRUE) does NOT always return a float, with the same number of decimals (php.ini can also affect microtime output).  
If it happens to land on an exact second, it returns an integer instead.

Here is some simple code to illustrate this:
```php
while(true) {
	usleep(100);
	$mt = microtime(TRUE);
	if(!str_contains($mt, ".")) {   /// str_contains only for php > 8.0
		echo $mt;
		break;
	}
}
```


When you run this at a command line, you'll see output like this (and you'll know EXACTLY when I ran this):

1643869315.6995 
1643869315.6996 
1643869315.6998 
1643869315          <-- 
1643869315.7001 
1643869315.7003 
1643869315.7005

Why does this matter?  Because if you want to format microtime into a very friendly but exact date/time, this popular code will fail on exact seconds (but works great otherwise):

```php
   $raw_time = DateTime::createFromFormat('U.u', microtime(TRUE)); // This line will not create an object when microtime returns an integer because it depends on the number being a float, including a decimal point!
   $today = $raw_time->format('Y-m-d H:i:s.u'); // This will die when format is attempted on a non-object
```

The fix is to format the microtime output to force it to always be a float:
```php
   $utime = sprintf('%.4f', microtime(TRUE)); /// or use number_format
   $raw_time = DateTime::createFromFormat('U.u', $utime);
   $today = $raw_time->format('Y-m-d H:i:s.u');
```

sfadschm: I think the easiest solution here is `$time = microtime(true) * 10000;`.
This should give us the microtime as an integer value represnting microseconds in any of the cases you mentioned. It will also avoid the BC breaks in the method signatures.

system/Debug/Toolbar/Collectors/History.php
maniaba: Done

system/Debug/Toolbar/Collectors/History.php
maniaba: I have deleted this condition.

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
maniaba: Comment updated.

system/Debug/Toolbar.php
sfadschm: On second thought this might not work as it exceeds the `int` limit.

system/Debug/Toolbar.php
maniaba: Comment updated.

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+    protected float $time;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->time = (float) sprintf('%.4f', microtime(true));
+    }
+
+    protected function createDummyDebugbarJson()
```
paulbalandan, line 32: This can be set as private

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+use DateTime;
+
+/**
+ * @internal
+ */
+final class HistoryTest extends CIUnitTestCase
+{
+    private const STEP = 0.0001;
+
+    protected float $time;
```
paulbalandan, line 24: private float $time;

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+            ],
+            'method' => 'get',
+            'url'    => 'localhost',
+            'isAJAX' => false,
+        ];
+        // create 20 dummy debugbar json files
+        for ($i = 0; $i < 20; $i++) {
+            $path = str_replace($time, number_format($time - self::STEP, 4, '.', ''), $path);
+            file_put_contents($path, json_encode($dummyData));
+            $time = number_format($time - self::STEP, 4, '.', '');
```
paulbalandan, line 52: For consistency, use `sprintf` over `number_format`

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+        // test dir is now populated with json
+        $this->createDummyDebugbarJson();
+
+        $activeRowTime = $time = (float) sprintf('%.4f', $time - self::STEP);
+
+        $history = new History();
+        $history->setFiles($time, 20);
+        $this->assertIsArray($history->display());
+        $this->assertArrayHasKey('files', $history->display());
+        $this->assertNotEmpty($history->display(), 'Dummy Debugbar data is empty');
```
paulbalandan, line 73: I think you mean `$history->display()['files']`?
```suggestion
        $this->assertNotEmpty($history->display()['files'], 'Creating dummy debugbar data has failed.');
```

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+        foreach ($history->display()['files'] as $request) {
+            $this->assertSame((float) $request['time'], (float) $time, 'Microtime fail');
+            $this->assertSame($request['datetime'], DateTime::createFromFormat('U.u', $time)->format('Y-m-d H:i:s.u'), 'DateTime fail');
+            $this->assertSame($request['active'], ($time === $activeRowTime), 'Active row fail');
```
paulbalandan, line 78: Interchange the order of the arguments. The `$request` variables should be  the actual values being tested upon (in the 2nd param).

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+            $time = number_format($time - self::STEP, 4, '.', '');
```
paulbalandan, line 80: Use `sprintf` for consistency

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
paulbalandan, line 78: No need for the extra failure messages.

tests/system/Debug/Toolbar/Collectors/HistoryTest.php
maniaba: Why is that important because, this is final class, also this is test class.

paulbalandan: Exactly. This is a final class. You use `protected` because you intend child classes to be able to use the property. However, this is already final so there will be no inheritance. Protected visibility is pointless.

system/Debug/Toolbar/Collectors/History.php
paulbalandan: Given the limitations, I think it is acceptable to change the signature albeit a breaking change. Let's wait for other's opinions. @kenjis @MGatner ?

kenjis: I don't mind (necessary) BC in minor version up if it is documented.
See https://forum.codeigniter.com/thread-80892-post-392846.html#pid392846

By the way, is float enough?
It seems only 4 digits like 1643882644.5667 in my PHP.

maniaba: Maybe change this to a string, and make the file name look like "debugbar_{**microtime**} _ {**session_value**} .json", then we will solve the request **when multiple developers work on different modules I suggest the option that developers can filter requests by session so that everyone can see it's own requests made by their browser.**
I suggest adding a checkbox to the history tab, which would filter sessions by the browser.

paulbalandan: That is a good refinement to the Debug Toolbar, but let us limit this PR to adding the microtime only. We can add the session and filtering in another PR.

paulbalandan:
@maniaba Please add to the user guide changelog (1) debug toolbar is now using microtime instead of time, and (2) there is a possible BC break for those users extending the History Collector and they should probably update setFiles() method

system/Debug/Toolbar/Collectors/History.php
kenjis: I think type `string` is better. It is actually (a part of) filename, not timestamp.

kenjis:
This is not `(int)`:
https://github.com/codeigniter4/CodeIgniter4/blob/4ca2159818ca96d26bb9713b00941461509ad0b1/system/Debug/Toolbar.php#L485-L488

kenjis:
@maniaba 
Rector check was failed. Can you fix?
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#static-analysis-on-php-code

And when you sync your repository, can you use `git rebase`?

kenjis:
@maniaba
It seems you used `git merge`. Is it difficult to use `git rebase`?

system/Debug/Toolbar.php
```diff
@@ -485,10 +485,10 @@ protected function format(string $data, string $format = 'html'): string
     {
         $data = json_decode($data, true);
 
-        if ($this->config->maxHistory !== 0) {
+        if ($this->config->maxHistory !== 0 && preg_match('/\d+\.\d{4}/s', (string) Services::request()->getGet('debugbar_time'), $debugbarTime)) {
```
kenjis, line 488:
```suggestion
        if ($this->config->maxHistory !== 0 && preg_match('/\d+\.\d{6}/s', (string) Services::request()->getGet('debugbar_time'), $debugbarTime)) {
```

system/Debug/Toolbar/Collectors/History.php
```diff
@@ -81,13 +83,13 @@ public function setFiles(int $current, int $limit = 20)
 
             $contents = @json_decode($contents);
             if (json_last_error() === JSON_ERROR_NONE) {
-                preg_match_all('/\d+/', $filename, $time);
-                $time = (int) end($time[0]);
+                preg_match('/debugbar_(.*)\.json$/s', $filename, $time);
+                $time = sprintf('%.4f', $time[1] ?? 0);
```
kenjis, line 87:
```suggestion
                $time = sprintf('%.6f', $time[1] ?? 0);
```

kenjis (changes requested):
The current implementation always shows `00` at the end of Datetime.
Microtime has 6 digits, so it is better to show 6 digits.

screenshot 2022-02-24 11 09 30

kenjis:
```diff
Your changes to the SCSS files did not match the expected CSS output.
diff --git a/system/Debug/Toolbar/Views/toolbar.css b/system/Debug/Toolbar/Views/toolbar.css
index 680ef52..11a3301 100644
--- a/system/Debug/Toolbar/Views/toolbar.css
+++ b/system/Debug/Toolbar/Views/toolbar.css
@@ -774,7 +774,8 @@
 }
 
 .debug-bar-width190p {
-  width: 190px;}
+  width: 190px;
+}
 
 .debug-bar-width20e {
   width: 20em;
```
https://github.com/codeigniter4/CodeIgniter4/runs/5314886769?check_suite_focus=true

kenjis:
```
There was 1 failure:

1) CodeIgniter\Debug\Toolbar\Collectors\HistoryTest::testSetFiles
--- Expected
+++ Actual
@@ @@
-'1645684473.268600'
+'1645684473.2686'

/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Debug/Toolbar/Collectors/HistoryTest.php:75
```

kenjis:
@maniaba 
Fixing the failed tests seems to complete this PR.
Can you?

kenjis:
Closed via #5958

PR #5649, opened 2022-02-03
refactor: SPARKED

kenjis:
**Description**
The current `CLIRequest` does not know whether it is *Spark request* or *PHP CLI request*.

- add `CLIRequest::isSparked()`
- `spark` command uses sparked `CLIRequest`
- and we will not need the constant `SPARKED` that makes testing difficult

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [] User guide updated
- [x] Conforms to style guide

PR #5650, opened 2022-02-03
refactor: CodeIgniter has context

kenjis:
**Description**
Supersede: #5649
- add `CodeIgniter::$context`: `web`, `php-cli`, `spark`
- make the constant `SPARKED` deprecated
- remove `is_cli()` in `CodeIgniter`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/spark
+++ b/spark
@@ -21,6 +21,11 @@
  * this class mainly acts as a passthru to the framework itself.
  */
 
+/**
+ * @var bool
+ *
+ * @deprecated No longer in use. `CodeIgniter` has `$context` property.
+ */
 define('SPARKED', true);
 
 /*
@@ -51,7 +56,9 @@ $paths = new Config\Paths();
 chdir(FCPATH);
 
 $bootstrap = rtrim($paths->systemDirectory, '\\/ ') . DIRECTORY_SEPARATOR . 'bootstrap.php';
-$app       = require realpath($bootstrap) ?: $bootstrap;
+/** @var CodeIgniter\CodeIgniter $app */
+$app = require realpath($bootstrap) ?: $bootstrap;
+$app->setContext('spark');
 
 // Grab our Console
 $console = new CodeIgniter\CLI\Console($app);
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -142,7 +142,12 @@ class CodeIgniter
     protected $useSafeOutput = false;
 
     /**
-     * Context: 'web', 'php-cli', 'spark'
+     * Context
+     *  web:     Invoked by HTTP request
+     *  php-cli: Invoked by CLI via `php public/index.php`
+     *  spark:   Invoked by CLI via the `spark` command
+     *
+     * @phpstan-var 'php-cli'|'spark'|'web'
      */
     protected string $context;
 
@@ -1049,6 +1054,8 @@ class CodeIgniter
 
     /**
      * Sets the app context.
+     *
+     * @phpstan-param 'php-cli'|'spark'|'web' $context
      *
      * @return $this
      */
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -919,7 +919,7 @@ class CodeIgniter
         }
 
         throw PageNotFoundException::forPageNotFound(
-            ENVIRONMENT !== 'production' || ! $this->isWeb() ? $e->getMessage() : ''
+            (ENVIRONMENT !== 'production' || ! $this->isWeb()) ? $e->getMessage() : ''
         );
     }
 
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -142,6 +142,11 @@ class CodeIgniter
     protected $useSafeOutput = false;
 
     /**
+     * Context: 'web', 'php-cli', 'spark'
+     */
+    protected string $context;
+
+    /**
      * Constructor.
      */
     public function __construct(App $config)
@@ -294,6 +299,8 @@ class CodeIgniter
      */
     public function run(?RouteCollectionInterface $routes = null, bool $returnResponse = false)
     {
+        assert($this->context !== null, 'Context must be set before run() is called');
+
         $this->startBenchmark();
 
         $this->getRequestObject();
@@ -321,7 +328,7 @@ class CodeIgniter
         }
 
         // spark command has nothing to do with HTTP redirect and 404
-        if (defined('SPARKED')) {
+        if ($this->isSparked()) {
             return $this->handleRequest($routes, $cacheConfig, $returnResponse);
         }
 
@@ -359,6 +366,30 @@ class CodeIgniter
     }
 
     /**
+     * Invoked via spark command?
+     */
+    private function isSparked(): bool
+    {
+        return $this->context === 'spark';
+    }
+
+    /**
+     * Invoked via php-cli command?
+     */
+    private function isPhpCli(): bool
+    {
+        return $this->context === 'php-cli';
+    }
+
+    /**
+     * Web access?
+     */
+    private function isWeb(): bool
+    {
+        return $this->context === 'web';
+    }
+
+    /**
      * Handles the main request logic and fires the controller.
      *
      * @throws RedirectException
@@ -389,7 +420,7 @@ class CodeIgniter
         }
 
         // Never run filters when running through Spark cli
-        if (! defined('SPARKED')) {
+        if (! $this->isSparked()) {
             // Run "before" filters
             $this->benchmark->start('before_filters');
             $possibleResponse = $filters->run($uri, 'before');
@@ -430,7 +461,7 @@ class CodeIgniter
         $this->gatherOutput($cacheConfig, $returned);
 
         // Never run filters when running through Spark cli
-        if (! defined('SPARKED')) {
+        if (! $this->isSparked()) {
             $filters->setResponse($this->response);
 
             // Run "after" filters
@@ -548,10 +579,8 @@ class CodeIgniter
             return;
         }
 
-        if (is_cli() && ENVIRONMENT !== 'testing') {
-            // @codeCoverageIgnoreStart
+        if ($this->isSparked() || $this->isPhpCli()) {
             $this->request = Services::clirequest($this->config);
-        // @codeCoverageIgnoreEnd
         } else {
             $this->request = Services::request($this->config);
             // guess at protocol if needed
@@ -567,7 +596,7 @@ class CodeIgniter
     {
         $this->response = Services::response($this->config);
 
-        if (! is_cli() || ENVIRONMENT === 'testing') {
+        if ($this->isWeb()) {
             $this->response->setProtocolVersion($this->request->getProtocolVersion());
         }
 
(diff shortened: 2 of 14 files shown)


kenjis:
@samsonasik Is this a bug of rector?
```diff
1 file with changes
===================

1) system/CodeIgniter.php:369

    ---------- begin diff ----------
@@ @@
     }

     /**
-     * Invoked via php-cli command?
-     */
-    private function isPhpCli(): bool
-    {
-        return $this->context === 'php-cli';
-    }
-
-    /**
      * Web access?
      */
     private function isWeb(): bool
    ----------- end diff -----------

Applied rules:
 * RemoveUnusedPrivateMethodRector
```

kenjis:
@samsonasik 
> could you create failing test case?

No, I tried on getrector.org/demo, but `RemoveUnusedPrivateMethodRector` was not applied.

spark
```diff
@@ -12,6 +12,11 @@
  * this class mainly acts as a passthru to the framework itself.
  */
 
+/**
+ * @var bool
```
MGatner, line 16: Is this a valid docblock for a `define()` statement?

system/CodeIgniter.php
```diff
@@ -141,6 +141,11 @@ class CodeIgniter
      */
     protected $useSafeOutput = false;
 
+    /**
+     * Context: 'web', 'php-cli', 'spark'
+     */
```
MGatner, line 146:
```suggestion
     * @phpstan-property 'cli'|'spark'|'web'
     */
```

system/CodeIgniter.php
```diff
@@ -290,6 +295,8 @@ protected function initializeKint()
      */
     public function run(?RouteCollectionInterface $routes = null, bool $returnResponse = false)
     {
+        assert($this->context !== null, 'Context must be set before run() is called');
```
MGatner, line 298: I'm not sure `assert()` is available outside of a dev context. Confirm?

system/CodeIgniter.php
```diff
@@ -354,6 +361,30 @@ public function useSafeOutput(bool $safe = true)
...
+    }
+
+    /**
+     * Web access?
+     */
+    private function isWeb(): bool
+    {
+        return $this->context === 'web';
+    }
+
```
MGatner, line 387: Unless there is a compelling reason for these methods I prefer the simple property check inline.

MGatner (changes requested):
I like this much better than your first attempt. Some comments. Also, I think we should set a default context on `MockCodeIgniter` (probably "web"?) so people's tests don't start failing unexpectedly.

spark
kenjis: Yes. 
https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/var.html
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#518-var

system/CodeIgniter.php
kenjis: Changed to `@phpstan-var`.

`@property` is for Magic properties.
See https://phpstan.org/writing-php-code/phpdocs-basics#magic-properties

system/CodeIgniter.php
kenjis: This check is not needed in production. Because it is not a runtime error,
but a logic error.

And php.ini decided assert() is available or not. 
See https://www.php.net/manual/en/ini.core.php#ini.zend.assertions

system/CodeIgniter.php
kenjis: I removed the private methods.

kenjis:
>  Also, I think we should set a default context on MockCodeIgniter (probably "web"?) so people's tests don't start failing unexpectedly.

Added.

system/CodeIgniter.php
```diff
@@ -946,8 +958,8 @@ protected function gatherOutput(?Cache $cacheConfig = null, $returned = null)
     public function storePreviousURL($uri)
     {
         // Ignore CLI requests
-        if (is_cli() && ENVIRONMENT !== 'testing') {
-            return; // @codeCoverageIgnore
+        if (! ($this->context === 'web')) {
```
samsonasik, line 961:
```suggestion
        if ($this->context !== 'web') {
```

system/CodeIgniter.php
```diff
@@ -880,7 +890,9 @@ protected function display404errors(PageNotFoundException $e)
             ob_end_flush(); // @codeCoverageIgnore
         }
 
-        throw PageNotFoundException::forPageNotFound(ENVIRONMENT !== 'production' || is_cli() ? $e->getMessage() : '');
+        throw PageNotFoundException::forPageNotFound(
+            ENVIRONMENT !== 'production' || ! ($this->context === 'web') ? $e->getMessage() : ''
```
samsonasik, line 894: It seems needs parentheses to avoid confusion when read:

```suggestion
            ENVIRONMENT !== 'production' || ($this->context !== 'web' ? $e->getMessage() : '')
```

and possibly cause different result of expected output, ref https://3v4l.org/jNQEV vs https://3v4l.org/EqcGv

kenjis:
@MGatner 

> Unless there is a compelling reason for these methods I prefer the simple property check inline.

The simple property check inline is too complex to me, as @samsonasik pointed out.
Can I use private methods?

system/CodeIgniter.php
kenjis: It should be:
```php
            (ENVIRONMENT !== 'production' || ! ($this->context === 'web')) ? $e->getMessage() : ''
```
```php
            (ENVIRONMENT !== 'production' || $this->context !== 'web') ? $e->getMessage() : ''
```

spark
MGatner: Neither of those links provides any information on using `@var` on a define that I can see. Possibly having trouble locating it on mobile, but could you paste the relevant quote here?

system/CodeIgniter.php
MGatner: Ah thanks!

system/CodeIgniter.php
MGatner: Wow just read a lot on this, I was completely unaware of this whole area of PHP.

system/CodeIgniter.php
MGatner: This one is painful to read on mobile so I'll let you two figure it out 😅

system/Test/Mock/MockCodeIgniter.php
```diff
@@ -15,6 +15,16 @@
...
 class MockCodeIgniter extends CodeIgniter
 {
+    /**
+     * Context
+     *  web:     Invoked by HTTP request
+     *  php-cli: Invoked by CLI via `php public/index.php`
+     *  spark:   Invoked by CLI via the `spark` command
+     *
+     * @phpstan-var 'php-cli'|'spark'|'web'
+     */
```
MGatner, line 25: IDEs will use the parent docblock so it's fine to drop this.

tests/system/ControllerTest.php
```diff
@@ -71,6 +71,7 @@ protected function setUp(): void
         $this->response    = new Response($this->config);
         $this->logger      = \Config\Services::logger();
         $this->codeigniter = new MockCodeIgniter($this->config);
+        $this->codeigniter->setContext('web');
```
MGatner, line 74: I think these are no longer needed with the default property?

tests/system/RESTful/ResourceControllerTest.php
```diff
@@ -65,6 +65,7 @@ protected function setUp(): void
 
         $config            = new App();
         $this->codeigniter = new MockCodeIgniter($config);
+        $this->codeigniter->setContext('web');
```
MGatner, line 68: I think these are no longer needed with the default property?

tests/system/RESTful/ResourcePresenterTest.php
```diff
@@ -59,6 +59,7 @@ protected function setUp(): void
 
         $config            = new App();
         $this->codeigniter = new MockCodeIgniter($config);
+        $this->codeigniter->setContext('web');
```
MGatner, line 62: I think these are no longer needed with the default property?

MGatner (changes requested):
Looking good, still a few conversations going.

MGatner:
> The simple property check inline is too complex to me, as @samsonasik pointed out.
> 
> Can I use private methods?

I'm fine with the private methods. It still isn't my preference but I don't think either is "wrong" and I trust you as the author, unless there are other stronger opinions.

system/Test/Mock/MockCodeIgniter.php
kenjis: Dropped.

tests/system/ControllerTest.php
kenjis: Removed.

tests/system/RESTful/ResourceControllerTest.php
kenjis: Removed.

tests/system/RESTful/ResourcePresenterTest.php
kenjis: Removed.

spark
kenjis: > You may use the @var tag to document the "Type" of the following "Structural Elements":
>    Constants, both class and global scope

system/CodeIgniter.php
kenjis: Reverted to use the private method.

system/CodeIgniter.php
kenjis: Added parentheses,

samsonasik (approved):

system/CodeIgniter.php
```diff
@@ -290,6 +300,8 @@ protected function initializeKint()
      */
     public function run(?RouteCollectionInterface $routes = null, bool $returnResponse = false)
     {
+        assert($this->context !== null, 'Context must be set before run() is called');
```
samsonasik, line 303: is this will cause warning when user run `composer update` without update public/index.php?

kenjis: Yes, I added upgrade_420.rst.

samsonasik: Probably update the message? like:

```suggestion
assert(
    $this->context !== null, 
    'Context must be set before run() is called, if you are upgrading from 4.1.x, you can merge public/index.php and spark file from vendor/codeigniter4/framework'
);
```

kenjis: Updated.

samsonasik (approved):

kenjis:
@MGatner Can you review?

MGatner (approved):
Looks good. Should **spark** and **public/index.php** be included in the "All Changes" section?

kenjis:
> Should spark and public/index.php be included in the "All Changes" section?

Strictly speaking no, I think. Because:

> All changes noted here are **optiona**l until the next major version,
    and any mandatory changes will be covered in the sections above.

kenjis:
@samsonasik 
Can't reproduce.

```
$ php spark serve

CodeIgniter v4.1.8 Command Line Tool - Server Time: 2022-02-18 17:51:33 UTC-06:00

CodeIgniter development server started on http://localhost:8080
Press Control-C to stop.
[Sat Feb 19 08:51:33 2022] PHP 8.0.15 Development Server (http://localhost:8080) started
```

PR #5653, opened 2022-02-04
refactor: isCLI() in CLIRequest and IncomingRequest

kenjis:
**Description**
From  #5649

- no need to depend on `is_cli()`

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


samsonasik:
Thank you @kenjis

PR #5656, opened 2022-02-05
refactor: Session does not use cookies()

kenjis:
**Description**
- The Response should send all cookies.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/depfile.yaml
+++ b/depfile.yaml
@@ -198,6 +198,7 @@ ruleset:
     - HTTP
   Session:
     - Cookie
+    - HTTP
     - Database
   Throttle:
     - Cache
--- a/system/Session/Session.php
+++ b/system/Session/Session.php
@@ -12,8 +12,10 @@
 namespace CodeIgniter\Session;
 
 use CodeIgniter\Cookie\Cookie;
+use CodeIgniter\HTTP\Response;
 use Config\App;
 use Config\Cookie as CookieConfig;
+use Config\Services;
 use Psr\Log\LoggerAwareTrait;
 use Psr\Log\LoggerInterface;
 use SessionHandlerInterface;
@@ -904,6 +906,8 @@ class Session implements SessionInterface
         $expiration   = $this->sessionExpiration === 0 ? 0 : time() + $this->sessionExpiration;
         $this->cookie = $this->cookie->withValue(session_id())->withExpires($expiration);
 
-        cookies([$this->cookie], false)->dispatch();
+        /** @var Response $response */
+        $response = Services::response();
+        $response->setCookie($this->cookie);
     }
 }
--- a/user_guide_src/source/changelogs/v4.2.0.rst
+++ b/user_guide_src/source/changelogs/v4.2.0.rst
@@ -34,6 +34,7 @@ Changes
 
 - Update minimal PHP requirement to 7.4.
 - The current version of Content Security Policy (CSP) outputs one nonce for script and one for style tags. The previous version outputted one nonce for each tag.
+- The process of sending cookies has been moved to the ``Response`` class. Now the ``Session`` class doesn't send cookies, set them to the Response.
 
 Deprecations
 ************


paulbalandan (approved):

skuadron45:
hy kenji, i have problem after this commit. I am using myth auth and login with ajax. i see response for Set Cookie order was different before this commit. please you can see my repo this use case

https://github.com/skuadron45/ajaxmythauth

skuadron45:
focus on line 
/**
* if you comment this line, cookie will using old session and you are not logged in although successful with login process.
*/
Services::response()->getCookieStore()->clear();

skuadron45:
this problem i found when using version 4.2.0 and earlier, but there is no problem for version 4.1.9 and older

kenjis:
> i see response for Set Cookie order was different before this commit. 

Yes, it may change the order.
But why is it a problem? Did you send more than two cookies with the same name?

skuadron45:
so it's confirm it will change the order of Set Cookie,

The problem is when i am using myth auth for authentication with ajax, the Set Cookie order was changed after version 4.2.0

In AuthenticationBase in myauth, after we successfully loggin, they regenerate the session, i don't know how it work, but when we regenerate session, there is will Set Cookie for new session, but in V 4.2.0, the problem was Set Cookie order was changed.
last Set Cookie was old session.

maybe you can see my repo if you can help me with this commit issue.
https://github.com/skuadron45/ajaxmythauth

I already compare Set Cookie order for this project only success until version 4.1.9 before this commit

Thanks

kenjis:
You mean, after successful login, CI sends two session cookies?

skuadron45:
yes if i use ajax method @kenjis

skuadron45:
should i send a screenshoot ?

skuadron45:
gambar

blue was old session, and red was new session after session regenerate was calling. on 4.1.9 
Set Cookie order is blue after that red, but in 4.2.0, Set Cookie order is like the picture above.

If i call this method before send the response, it will fix the problem because Set Cookie is only one for new session (red)
gambar

I hope you can help me to troubleshoot about this problem, i don't know if this issue for CI or myth auth, because in myth auth, after successfull login, they will regenerate the session

gambar

Thanks

skuadron45:
i have a question, what is Set Cookie was send if we regenerate session by default ?
because i didn't see this on debug the code, so i can't found why there is Set Cookie if there was a regenerate session.

kenjis:
It seems this bug is related to Apache.
https://bugs.php.net/bug.php?id=75554

kenjis:
> this problem didn't appear if we use redirrect response and not using ajax

Did you redirect with `withCookies()`?
```php
return redirect()->back()->withCookies();
```
https://codeigniter4.github.io/CodeIgniter4/general/common_functions.html#redirect

kenjis:
It seems this is a bug in the framework Session.

As a workaround, remove the old session cookie with `CookieStore::remove()`.
If there is the old session cookie only, `CookieStore::clear()` as you did is fine.

skuadron45:
> > this problem didn't appear if we use redirrect response and not using ajax
> 
> Did you redirect with `withCookies()`?
> 
> ```
> return redirect()->back()->withCookies();
> ```
> 
> https://codeigniter4.github.io/CodeIgniter4/general/common_functions.html#redirect

I don't have problem with page request, the problem only appear if i use ajax,

skuadron45:
> It seems this is a bug in the framework Session.
> 
> As a workaround, remove the old session cookie with `CookieStore::remove()`. If there is the old session cookie only, `CookieStore::clear()` as you did is fine.

Okay thanks, so that was i done for resolve this case, thanks kenjis

kenjis:
I don't know why this issue appears only on Ajax.

skuadron45:
> It seems this bug is related to Apache. https://bugs.php.net/bug.php?id=75554
> 
> But in this case, it may be a bit different from the PHP bug.

I think so, because at v 4.1.9, there are 2 Set Cookie, the different just only in ordering as you can see on picture above

skuadron45:
But this was very clear for me

"Servers SHOULD NOT include more than one Set-Cookie header field in the same response with the same cookie-name.  (See Section 5.2 for how user agents handle this case.)"

kenjis:
I created an issue for this: #6166

skuadron45:
> I created an issue for this: #6166

Thanks @kenjis 

I hope you can reproduce this issue.

PR #5657, opened 2022-02-05
refactor: update Kint CSP nonce

kenjis:
**Description**
- use Kint 4.1 CSP nonce functionality

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


system/CodeIgniter.php
```diff
@@ -257,7 +258,12 @@ protected function initializeKint()
             Kint::$plugins = $config->plugins;
         }
 
-        Kint::$renderers[Kint::MODE_RICH] = RichRenderer::class;
+        /** @var ContentSecurityPolicy $csp */
+        $csp = Services::csp();
```
paulbalandan, line 262: I think the inline PHPDoc is unnecessary since it's already typehinted in `Config\Services`

kenjis: Yes, I changed my mind. We should write `@var` for local variables as little  as possible. It is a last resort.

Removed.

paulbalandan (approved):

PR #5659, opened 2022-02-05
[Demo] Loading code examples in docs from external files to ease CS.

sfadschm:
**Description**
A lot of recent PRs deal with fixing the CS of examples in the docs.

This PR demonstrates loading the examples as php files into the docs.

The output is identical:

Unbenannt

This way, e.g., `php-cs-fixer` can easily be applied to all source files of the docs.

Note that this requires discussion of the folder structure to store the snippet files and requires some initial work to extract all samples.

**Checklist:**
Only for demonstration.

### DIFF
--- a/.github/workflows/test-coding-standards.yml
+++ b/.github/workflows/test-coding-standards.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Install dependencies
         run: composer update --ansi --no-interaction
 
-      - name: Run lint on `app/`, `admin/`, `public/`
+      - name: Run lint on `app/`, `admin/`, `public/`, 'user_guide_src/'
         run: vendor/bin/php-cs-fixer fix --verbose --ansi --dry-run --config=.no-header.php-cs-fixer.dist.php --using-cache=no --diff
 
       - name: Run lint on `system/`, `tests`, `utils/`, and root PHP files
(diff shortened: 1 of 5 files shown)


sfadschm:
Why does the fixer not recognize the extra space in the `if` condition?

paulbalandan:
Not sure, but maybe because it is missing the leading `<?php` ?

sfadschm:
> Not sure, but maybe because it is missing the leading `<?php` ?

Possibly, however, other rules were applied. I will play around a litte.

sfadschm:
The demo is working now.

Not so happy about the `:lines: 2-` solution.
Possible workarounds:

- tell sphinx to ignore the first line or opening tags in general
- tell php-cs-fixer to also handle files without php opening tag

Yet, none of these seems to be possible as far as I could find.

Just curious whether the people think this would be helpful or just increase the number of files to maintain.
Personally, I feel this approach is worth the work it requires to be initialized, as I like includes in general and also storing code in the format it should be  😃

kenjis:
I think this way is very good, but the initial work is a lot of work.

sfadschm:
I can do the conversion, might just take some time.

sfadschm:
I will make a fresh PR with the changes in the next days.

PR #5668, opened 2022-02-09
Permit use of lates php-cs-fixer v3.6

paulbalandan:
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "require-dev": {
         "codeigniter/coding-standard": "^1.1",
         "fakerphp/faker": "^1.9",
-        "friendsofphp/php-cs-fixer": "3.4.*",
+        "friendsofphp/php-cs-fixer": "^3.6",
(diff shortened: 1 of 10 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
     "require-dev": {
         "codeigniter/coding-standard": "^1.1",
         "fakerphp/faker": "^1.9",
-        "friendsofphp/php-cs-fixer": "3.4.*",
+        "friendsofphp/php-cs-fixer": "^3.6",
```
kenjis, line 20: Do we also update `admin/framework/composer.json` ?

paulbalandan: Good catch! I'll update in a while.

paulbalandan: Added.

kenjis (approved):

PR #5672, opened 2022-02-10
fix: $builder->ignore()->insertBatch() only ignores on first iteration

### LINKED ISSUE #5671: Bug: $builder->ignore()->insertBatch() only ignores on first iteration of batch
### PHP Version

8.0

### CodeIgniter4 Version

4.1.8

Windows

### Which server did you use?

apache

### Database

MariaDB

### What happened?

When using ignore() with insertBatch() only the first iteration of the batch uses the proper INSERT IGNORE INTO... statement. 

### Steps to Reproduce

In model call: $this->ignore()->insertBatch($data);

Be sure that batch is larger than one.

### Expected Output

Expected that each iteration of a batch insert will use INSERT INGNORE INTO.. in SQL statement.

### Anything else?

It looks like the code has recently changed quite a bit for /codeigniter4/system/Database/BaseBuilder.php.

 line 349 the property $this->QBIgnore is set
```
    public function ignore(bool $ignore = true)
    {
        $this->QBIgnore = $ignore;

        return $this;
    }
```
 line 1636 starts a loop for batches in the method insertBatch()

```
        for ($i = 0, $total = count($set); $i < $total; $i += $batchSize) {
            if ($hasQBSet) {
                $QBSet = array_slice($this->QBSet, $i, $batchSize);
            } else {
                $this->setInsertBatch(array_slice($set, $i, $batchSize), '', $escape);
                $QBSet = $this->QBSet;
            }
            $sql = $this->_insertBatch($this->db->protectIdentifiers($table, true, null, false), $this->QBKeys, $QBSet);

            if ($this->testMode) {
                $savedSQL[] = $sql;
            } else {
                $this->db->query($sql, null, false);
                $affectedRows += $this->db->affectedRows();
            }

            if (! $hasQBSet) {
                $this->resetWrite();
            }
        }
```
 the problem occurs on line 1652 where the method resetWrite() is called. This method sets the property QBIgnore to false

```
            if (! $hasQBSet) {
                $this->resetWrite();
            }
```

 as a patch I have modified the resetWrite() method to accept the QBIgnore property and allow this field to be persisted each iteration of the loop. 

 the resetWrite() method as I have modified it on line 2632

```
    protected function resetWrite($qbIgnore=false)
    {
        $this->resetRun([
            'QBSet'     => [],
            'QBJoin'    => [],
            'QBWhere'   => [],
            'QBOrderBy' => [],
            'QBKeys'    => [],
            'QBLimit'   => false,
            'QBIgnore'  => $qbIgnore,
        ]);
    }
```

 then on line 1652 I have modified the call to resetWrite() method to include the current state of QBIgnore property and allow it to be persisted through loop.

```
            if (! $hasQBSet) {
                $this->resetWrite($this->QBIgnore);
            }
```

 then on line 1657 after batch loop ends the method is called without the QBIgnore  parameter to reset. This was not modified.
        
`$this->resetWrite();`

I'm not sure if this is the best way of fixing the problem as I'm not sure exactly of everything going on here. The fix does provide the desired behavior in the basic scenario mentioned above.

kenjis:
**Description**
Fixes #5671

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1662,7 +1662,10 @@ class BaseBuilder
             }
 
             if (! $hasQBSet) {
-                $this->resetWrite();
+                $this->resetRun([
+                    'QBSet'  => [],
+                    'QBKeys' => [],
+                ]);
             }
         }
 
--- a/system/Test/Mock/MockBuilder.php
+++ b/system/Test/Mock/MockBuilder.php
@@ -15,4 +15,9 @@ use CodeIgniter\Database\BaseBuilder;
 
 class MockBuilder extends BaseBuilder
 {
+    protected $supportedIgnoreStatements = [
+        'update' => 'IGNORE',
+        'insert' => 'IGNORE',
+        'delete' => 'IGNORE',
+    ];
 }
--- a/tests/system/Database/Builder/InsertTest.php
+++ b/tests/system/Database/Builder/InsertTest.php
@@ -100,6 +100,43 @@ final class InsertTest extends CIUnitTestCase
         $this->assertSame($expected, str_replace("\n", ' ', $query->getQuery()));
     }
 
+    /**
+     * @see https://github.com/codeigniter4/CodeIgniter4/issues/5671
+     */
+    public function testInsertBatchIgnore()
+    {
+        $builder = $this->db->table('jobs');
+
+        $insertData = [
+            [
+                'id'          => 2,
+                'name'        => 'Commedian',
+                'description' => 'There\'s something in your teeth',
+            ],
+            [
+                'id'          => 3,
+                'name'        => 'Cab Driver',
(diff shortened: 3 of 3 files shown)


system/Database/BaseBuilder.php
```diff
@@ -1662,7 +1662,9 @@ public function insertBatch(?array $set = null, ?bool $escape = null, int $batch
             }
 
             if (! $hasQBSet) {
+                $ignore = $this->QBIgnore;
                 $this->resetWrite();
+                $this->QBIgnore = $ignore;
```
iRedds, line 1667: Maybe just reset $QBSet and $QBKeys?

samsonasik (approved):

kenjis:
@samsonasik Thank you!

PR #5675, opened 2022-02-10
Allow for custom exception handlers.

lonnieezell:
Allows developers to easily create and integrate their own Exception handler classes, and limiting their usage
to only the exception or HTTP status code desired. 

Supercedes #5355 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/ExceptionHandler.php
+++ b/system/Debug/ExceptionHandler.php
@@ -1,0 +1,92 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Debug;
+
+use CodeIgniter\Exceptions\PageNotFoundException;
+use Config\Paths;
+use Throwable;
+
+class ExceptionHandler extends BaseExceptionHandler
+{
+    /**
+     * Determines the correct way to display the error.
+     *
+     * @return mixed|void
+     */
+    public function handle()
+    {
+        if (! is_cli()) {
+            $this->response->setStatusCode($this->statusCode);
+            header(sprintf('HTTP/%s %s %s', $this->request->getProtocolVersion(), $this->response->getStatusCode(), $this->response->getReasonPhrase()), true, $this->statusCode);
+
+            // Display a JSON/XML response if we weren't asked for HTML
+            if (strpos($this->request->getHeaderLine('accept'), 'text/html') === false) {
+                $this->respond(ENVIRONMENT === 'development' ? $this->collectVars($this->exception, $this->statusCode) : '', $this->statusCode)->send();
+
+                exit($this->exitCode);
+            }
+        }
+
+        // Determine possible directories of error views
+        $path    = $this->viewPath;
+        $altPath = rtrim((new Paths())->viewDirectory, '\\/ ') . DIRECTORY_SEPARATOR . 'errors' . DIRECTORY_SEPARATOR;
+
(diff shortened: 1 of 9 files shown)


MGatner:
This is more then I can handle on mobile, but looks really good! On my "to do" list.

iRedds (commented):
It seems to me that this PR is BC.
Services::exceptions() returns an instance of the CodeIgniter\Debug\Exceptions class, which can be extended.
Changing this class will cause breaking changes for the custom implementation.

system/Debug/ExceptionHandler.php
```diff
@@ -0,0 +1,92 @@
...
+                exit($this->exitCode);
+            }
+        }
+
+        // Determine possible directories of error views
+        $path    = $this->viewPath;
+        $altPath = rtrim((new Paths())->viewDirectory, '\\/ ') . DIRECTORY_SEPARATOR . 'errors' . DIRECTORY_SEPARATOR;
+
+        $path    .= (is_cli() ? 'cli' : 'html') . DIRECTORY_SEPARATOR;
+        $altPath .= (is_cli() ? 'cli' : 'html') . DIRECTORY_SEPARATOR;
```
samsonasik, line 44: repetitive same assign expr can be moved to a variable before:

```suggestion
        $addPath  = (is_cli() ? 'cli' : 'html') . DIRECTORY_SEPARATOR;
        $path    .= $addPath;
        $altPath .= $addPath;
```

system/Debug/Exceptions.php
```diff
@@ -107,20 +85,25 @@ public function exceptionHandler(Throwable $exception)
...
-                $this->respond(ENVIRONMENT === 'development' ? $this->collectVars($exception, $statusCode) : '', $statusCode)->send();
-
-                exit($exitCode);
-            }
+        if (! method_exists($config, 'handler')) {
+            exit('Config\Exception must have a handler() method.');
         }
 
-        $this->render($exception, $statusCode);
+        $handler = config('Exceptions')->handler($statusCode, $exception);
```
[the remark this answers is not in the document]
lonnieezell: Good catch. Thanks. I forgot to go back and clean that up. Will do.

lonnieezell:
> It seems to me that this PR is BC. Services::exceptions() returns an instance of the CodeIgniter\Debug\Exceptions class, which can be extended. Changing this class will cause breaking changes for the custom implementation.

Yeah, you're correct. I have been sick the last couple of days but needed to do something other than sleep for a few hours and wasn't up to tackling the refactor of Tasks needed to get it integrated into the framework so I figured I could knock this out. Obviously wasn't thinking 100%. 

We can shelve it or start a 5.0 branch so it doesn't just clog up the PR lists, or something.

system/Debug/ExceptionHandler.php
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Determines the correct way to display the error.
+     *
+     * @return mixed|void
+     */
+    public function handle()
+    {
+        if (! is_cli()) {
+            $this->response->setStatusCode($this->statusCode);
+            header(sprintf('HTTP/%s %s %s', $this->request->getProtocolVersion(), $this->response->getStatusCode(), $this->response->getReasonPhrase()), true, $this->statusCode);
```
sclubricants, line 29: If headers have been sent, calling header() causes error.

https://github.com/codeigniter4/CodeIgniter4/pull/5680

system/Debug/Exceptions.php
```diff
@@ -107,20 +85,25 @@ public function exceptionHandler(Throwable $exception)
             ]);
```
sclubricants, line 85: https://github.com/codeigniter4/CodeIgniter4/pull/5684

kenjis:
I have updated ths PR for 4.4, and it is waiting for reviews.
#7087

kenjis:
Closed by #7087

PR #5682, opened 2022-02-12
fix: $routes->setDefaultMethod() does not work

kenjis:
**Description**
- `$routes->setDefaultMethod()` does not work when you go to `/`

See https://forum.codeigniter.com/thread-81269.html

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -638,7 +638,7 @@ class Router implements RouterInterface
      */
     protected function setRequest(array $segments = [])
     {
-        // If we don't have any segments - try the default controller;
+        // If we don't have any segments - use the default controller;
         if (empty($segments)) {
             return;
         }
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -459,15 +459,12 @@ class Router implements RouterInterface
     {
         $segments = explode('/', $uri);
 
+        // WARNING: Directories get shifted out of the segments array.
         $segments = $this->scanControllers($segments);
 
-        // If we don't have any segments left - try the default controller;
-        // WARNING: Directories get shifted out of the segments array.
-        if (empty($segments)) {
-            $this->setDefaultController();
-        }
+        // If we don't have any segments left - use the default controller;
         // If not empty, then the first segment should be the controller
-        else {
+        if (! empty($segments)) {
             $this->controller = ucfirst(array_shift($segments));
         }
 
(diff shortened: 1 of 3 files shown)


iRedds:
Wrong approach.
The constructor of the Router class already defines a controller and a default method.
And the Router::setDefaultController() method doesn't make sense.

kenjis:
Removing a protected method is a breaking change.

I reverted ``Router::setDefaultController()`` and make it deprecated.

system/Router/Router.php
```diff
@@ -646,8 +640,6 @@ protected function setRequest(array $segments = [])
     {
         // If we don't have any segments - try the default controller;
```
paulbalandan, line 641: This comment is no longer true

system/Router/Router.php
```diff
@@ -462,15 +459,12 @@ public function autoRoute(string $uri)
...
 
+        // WARNING: Directories get shifted out of the segments array.
         $segments = $this->scanControllers($segments);
 
-        // If we don't have any segments left - try the default controller;
-        // WARNING: Directories get shifted out of the segments array.
-        if (empty($segments)) {
-            $this->setDefaultController();
-        }
+        // If we don't have any segments left - use the default controller;
```
paulbalandan, line 465: I think this comment should also be removed?

user_guide_src/source/changelogs/v4.2.0.rst
```diff
@@ -42,6 +42,7 @@ Deprecations
 - ``CodeIgniter\Database\SQLSRV\Connection::getError()`` is deprecated. Use ``CodeIgniter\Database\SQLSRV\Connection::error()`` instead.
 - ``CodeIgniter\Debug\Exceptions::cleanPath()`` and ``CodeIgniter\Debug\Toolbar\Collectors\BaseCollector::cleanPath()`` are deprecated. Use the ``clean_path()`` function instead.
 - ``CodeIgniter\Log\Logger::cleanFilenames()`` and ``CodeIgniter\Test\TestLogger::cleanup()`` are both deprecated. Use the ``clean_path()`` function instead.
+- ``CodeIgniter\Router\setDefaultController()`` is deprecated.
```
paulbalandan, line 45:
```suggestion
- ``CodeIgniter\Router\Router::setDefaultController()`` is deprecated.
```

system/Router/Router.php
kenjis: It is true. I don't think to be removed.

system/Router/Router.php
kenjis: Indeed. "try" is strange.
Updated.

paulbalandan (approved):

PR #5685, opened 2022-02-12
Do not call non-statically the static `getPrivateMethodInvoker()` method

paulbalandan:
**Description**
Although PHP does not complain with it (only complains with the reverse scenario), the usage is still bad code.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -38,6 +38,7 @@ use Rector\EarlyReturn\Rector\Return_\PreparedValueToEarlyReturnRector;
 use Rector\Php55\Rector\String_\StringClassNameToClassConstantRector;
 use Rector\Php56\Rector\FunctionLike\AddDefaultValueForUndefinedVariableRector;
 use Rector\Php70\Rector\FuncCall\RandomFunctionRector;
+use Rector\Php70\Rector\MethodCall\ThisCallOnStaticMethodToStaticCallRector;
 use Rector\Php71\Rector\FuncCall\CountOnNullRector;
 use Rector\Php73\Rector\FuncCall\JsonThrowOnErrorRector;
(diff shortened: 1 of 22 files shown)


sfadschm:
This looks like something that could be automated:
`Rector\Php\Rector\MethodCall\ThisCallOnStaticMethodToStaticCallRector`

kenjis:
> the usage is still bad code.

What's bad?

paulbalandan:
> > the usage is still bad code.
> 
> What's bad?

IMO since the method is defined as static, it should be invoked as static also. The fact that PHP allows non-static calls on static methods but not static calls on non-static methods is one of the reason why it is inconsistent. At least in the framework we enforce consistency where PHP failed to do so.

sfadschm:
> What's bad?

IMO not really bad code, but poor style. 
The methods are static by design and there is no drawback in calling them statically. Therefore, this should be the intended way to go.

kenjis:
>  At least in the framework we enforce consistency where PHP failed to do so.

PHP does not fail. It is the specification of PHP.

rector.php
```diff
@@ -145,4 +146,5 @@
     $services->set(MakeInheritedMethodVisibilitySameAsParentRector::class);
     $services->set(SimplifyEmptyArrayCheckRector::class);
     $services->set(NormalizeNamespaceByPSR4ComposerAutoloadRector::class);
+    $services->set(ThisCallOnStaticMethodToStaticCallRector::class);
```
samsonasik, line 149: this service is already in php70 config set, which included already when use `LevelSetList::UP_TO_PHP_74`, so it possibly a bug

PR #5691, opened 2022-02-13
Convert variable type to int if it was an integer fixes #5688

### LINKED ISSUE #5688: Bug: Ajax request makes the session expire when app.sessionExpiration is set to 0
### PHP Version

7.4

### CodeIgniter4 Version

4.1.8

Linux

### Which server did you use?

apache

### Database

_No response_

### What happened?

```
<script> 
        var checkSession;
        function CheckForSession() {
          jQuery.ajax({
            type: "POST",
            cache: false,
            success: function(result) {
              }
            }
          });
        }
      });
</script>
```
```
    {
            $data = [
                'sessionStatus' => 0
            ];
        } else {
            $data = [
                'sessionStatus' => 1
            ];
        }
    }
```

### Steps to Reproduce

### Expected Output

### Anything else?

_No response_

rfikree92:
Fixes #5688 

**Description**
convert any variable that is a number to int type, so errors like #5688 doesn't happen.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Config/BaseConfig.php
+++ b/system/Config/BaseConfig.php
@@ -102,9 +102,17 @@ class BaseConfig
             } elseif ($value === 'true') {
                 $value = true;
             }
-            $property = is_bool($value) ? $value : trim($value, '\'"');
-            if (preg_match("/^[0-9]+$/", $property)) {
-                $property = (int) $property;
+            if (is_bool($value)) {
+                $property = $value;
+            } elseif(is_numeric($value)) {
+                // assume it's a float number if passed is_numeric and has ".", "e" or "E" in it
+                if(preg_match('/\.|[eE]/', $value)) {
+                    $property = (float) $value;
+                } else {
+                    $property = (int) $value;
+                }
+            } else {
+                $property = trim($value, '\'"');
             }
         }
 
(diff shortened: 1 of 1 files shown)


system/Config/BaseConfig.php
```diff
@@ -103,6 +103,9 @@ protected function initEnvValue(&$property, string $name, string $prefix, string
                 $value = true;
             }
             $property = is_bool($value) ? $value : trim($value, '\'"');
+            if (preg_match("/^[0-9]+$/", $property)) {
+                $property = (int) $property;
+            }
```
MGatner, line 108: Placing this after the trim removes the possibility of passing string numerals. Maybe switch the order of these? Also, any sense to expanding this to `is_numeric()` and casting to float more generally?

MGatner (changes requested):
Inline comment. Also we would need a test to cover this - should be a straightforward adjustment of current tests.

kenjis:
@rfikree92 Thank you for sending this PR.

Some GitHub Action checks failed.
First, please fix coding style. See

And fix and add unit tests.

kenjis:
I think the following test cases are needed:
```
'02471'
'1337e0'
" 42"
"42 "
```
https://www.php.net/manual/en/function.is-numeric.php#refsect1-function.is-numeric-examples

PR #5696, opened 2022-02-15
refactor: DB Session Handler

kenjis:
**Description**
- make database specific handlers

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -47,6 +47,9 @@ use CodeIgniter\Router\RouteCollection;
 use CodeIgniter\Router\RouteCollectionInterface;
 use CodeIgniter\Router\Router;
 use CodeIgniter\Security\Security;
+use CodeIgniter\Session\Handlers\Database\MySQLiHandler;
+use CodeIgniter\Session\Handlers\Database\PostgreHandler;
+use CodeIgniter\Session\Handlers\DatabaseHandler;
 use CodeIgniter\Session\Session;
 use CodeIgniter\Throttle\Throttler;
 use CodeIgniter\Typography\Typography;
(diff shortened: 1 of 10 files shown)


system/Config/Services.php
```diff
@@ -585,7 +589,21 @@ public static function session(?App $config = null, bool $getShared = true)
         $logger = AppServices::logger();
 
         $driverName = $config->sessionDriver;
-        $driver     = new $driverName($config, AppServices::request()->getIPAddress());
+
+        if ($driverName === DatabaseHandler::class) {
```
samsonasik, line 593: how about using `is_a()` check?

```suggestion
        if (is_a($driverName, DatabaseHandler::class, true)) {
```

for handle extended DatabaseHandler ?

kenjis: Thanks for suggestion, but in this case, I think string comparison is enough.
Because it really compares just strings. It has nothing to do with the parent class.

samsonasik: ok 👍

system/Session/Handlers/DatabaseHandler.php
```diff
@@ -11,14 +11,17 @@
...
 use CodeIgniter\Session\Exceptions\SessionException;
 use Config\App as AppConfig;
 use Config\Database;
 use ReturnTypeWillChange;
 
 /**
- * Session handler using current Database for storage
+ * Base database session handler
+ *
+ * Do not use this class. Use database specific handler class.
```
MGatner, line 24: What are SQLite and MSSQL supposed to use?

kenjis: We now support only MySQL and PostgreSQL.

> Only MySQL and PostgreSQL databases are officially supported, due to lack of advisory locking mechanisms on other platforms. Using sessions without locks can cause all sorts of problems, especially with heavy usage of AJAX, and we will not support such cases. Use session_write_close() after you’ve done processing session data if you’re having performance issues.
https://codeigniter4.github.io/CodeIgniter4/libraries/sessions.html#databasehandler-driver

I think MSSQL would be supported if someone write a handler.

kenjis:
Rebased to resolve conflicts.

PR #5698, opened 2022-02-15
docs: fix out of dated description in autoloader.rst

kenjis:
**Description**
- Legacy Support was removed.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/concepts/autoloader.rst
+++ b/user_guide_src/source/concepts/autoloader.rst
@@ -10,10 +10,9 @@ hard-coding that location into your files in a series of ``requires()`` is a mas
 headache and very error-prone. That's where autoloaders come in.
 
 CodeIgniter provides a very flexible autoloader that can be used with very little configuration.
-It can locate individual non-namespaced classes, namespaced classes that adhere to
+It can locate individual namespaced classes that adhere to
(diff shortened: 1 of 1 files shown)


user_guide_src/source/concepts/autoloader.rst
```diff
@@ -10,7 +10,7 @@ hard-coding that location into your files in a series of ``requires()`` is a mas
 headache and very error-prone. That's where autoloaders come in.
 
 CodeIgniter provides a very flexible autoloader that can be used with very little configuration.
-It can locate individual non-namespaced classes, namespaced classes that adhere to
+It can locate individual namespaced classes that adhere to
 `PSR-4 <https://www.php-fig.org/psr/psr-4/>`_ autoloading
 directory structures, and will even attempt to locate classes in common directories (like Controllers,
 Models, etc).
```
paulbalandan, line 16: `and will even attempt to locate classes in common directories (like Controllers, Models, etc)`
Does this make sense? I think this is already covered by PSR4 as we enforce controllers/models match their paths?

kenjis: No, I removed.

paulbalandan (approved):

PR #5704, opened 2022-02-17
Expand `.env` values conversion ability of `BaseConfig`

### LINKED ISSUE #5688: Bug: Ajax request makes the session expire when app.sessionExpiration is set to 0

paulbalandan:
**Description**
Initial (current):
```php
'true' => true
'false' => false
all other strings => trimmed
```

Additional (proposed):
```php
'12' => 12
'1.25' => 1.25
'null' => null
'empty' => ''
```

Supersedes and closes #5691 
Fixes #5688 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/BaseConfig.php
+++ b/system/Config/BaseConfig.php
@@ -96,13 +96,39 @@ class BaseConfig
             foreach (array_keys($property) as $key) {
                 $this->initEnvValue($property[$key], "{$name}.{$key}", $prefix, $shortPrefix);
             }
-        } elseif (($value = $this->getEnvValue($name, $prefix, $shortPrefix)) !== false && $value !== null) {
-            if ($value === 'false') {
-                $value = false;
-            } elseif ($value === 'true') {
-                $value = true;
+        } else {
+            $value = $this->getEnvValue($name, $prefix, $shortPrefix);
+
+            if ($value !== null) {
+                $lowered = strtolower($value);
+
+                switch (true) {
+                    case $lowered === 'false':
+                        $value = false;
+                        break;
+
+                    case $lowered === 'true':
+                        $value = true;
+                        break;
+
+                    case $lowered === 'empty':
+                        $value = '';
+                        break;
+
+                    case $lowered === 'null':
+                        $value = null;
+                        break;
+
+                    default:
+                        $value = trim($value, '\'"');
+                }
+
+                if (is_numeric($value)) {
+                    $value = strpos($value, '.') !== false ? (float) $value : (int) $value;
+                }
+
+                $property = $value;
             }
-            $property = is_bool($value) ? $value : trim($value, '\'"');
         }
 
         return $property;
@@ -111,7 +137,7 @@ class BaseConfig
     /**
      * Retrieve an environment-specific configuration setting
      *
-     * @return mixed
+     * @return string|null
      */
     protected function getEnvValue(string $property, string $prefix, string $shortPrefix)
     {
--- a/tests/system/Config/fixtures/.env
+++ b/tests/system/Config/fixtures/.env
@@ -32,3 +32,8 @@ SimpleConfig.crew.captain = Malcolm
 SimpleConfig.crew.pilot = Wash
 SimpleConfig.crew.comms = true
 SimpleConfig.crew.doctor = false
+SimpleConfig.crew.ranger = null
+SimpleConfig.crew.poacher = empty
+
+SimpleConfig.shortie = '12.45'
+SimpleConfig.longie = '250'
--- a/tests/system/Config/BaseConfigTest.php
+++ b/tests/system/Config/BaseConfigTest.php
@@ -62,15 +62,15 @@ final class BaseConfigTest extends CIUnitTestCase
     public function testServerValues()
     {
         $_SERVER = [
-            'simpleconfig.shortie' => 123,
-            'SimpleConfig.longie'  => 456,
+            'simpleconfig.shortie' => '123',
+            'simpleconfig.longie'  => '456',
         ];
         $dotenv = new DotEnv($this->fixturesFolder, '.env');
         $dotenv->load();
         $config = new SimpleConfig();
 
-        $this->assertSame('123', $config->shortie);
-        $this->assertSame('456', $config->longie);
+        $this->assertSame(123, $config->shortie);
+        $this->assertSame(456, $config->longie);
     }
 
     public function testEnvironmentOverrides()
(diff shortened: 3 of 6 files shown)


iRedds:
I vote for php.
Type of variables out of the box.
No extra parsing needed.
=)

tests/system/Config/fixtures/.env
```diff
@@ -32,3 +32,8 @@ SimpleConfig.crew.captain = Malcolm
 SimpleConfig.crew.pilot = Wash
 SimpleConfig.crew.comms = true
 SimpleConfig.crew.doctor = false
+SimpleConfig.crew.ranger = null
+SimpleConfig.crew.poacher = empty
+
+SimpleConfig.shortie = '12.45'
+SimpleConfig.longie = '250'
```
kenjis, line 39: I think `'12.45'` or `"12.45"` is string, `12.45` is float.

kenjis:
@iRedds See https://github.com/codeigniter4/CodeIgniter4/issues/5688#issuecomment-1037975842

tests/system/Config/fixtures/.env
paulbalandan: So you mean drop the trimming of quotes? And regard encapsed numbers as intended to be strings?

kenjis:
> The type cast must be inside.

What do you mean?

At least, I disagree with the current Paul's implementation, which both `SimpleConfig.shortie = '12.45'` and `SimpleConfig.shortie = 12.45` are converted to float.

tests/system/Config/fixtures/.env
kenjis: As a PHP user, I think `'12.45'` is a string, not an integer.
It is quoted intentionally.

I think adding the following two cases are enough.
`SimpleConfig.shortie = 12.45` as float, and `SimpleConfig.longie = 250` as integer.

iRedds:
> > The type cast must be inside.
> 
> What do you mean?
> 

$this->propertyExpectsInt = (int) env('property');

tests/system/Config/fixtures/.env
paulbalandan: However, inside a `.env` file where everything is string, quotes really serve no purpose other than capturing in-between spaces. That's why in `DotEnv::sanitizeValue()` we capture what's inside those quotes.
https://github.com/codeigniter4/CodeIgniter4/blob/968c0278a0ba4138cf3e64a2bbab5f3a800532d5/system/Config/DotEnv.php#L163

So, in the absence of a whitespace in between, `12.45 === '12.45'`
And in the default `env`, you can write strings with or without quotes. It doesn't matter as long there is no whitespace.
> # database.tests.password = root
> # contentsecuritypolicy.styleSrc = 'self'

MGatner: I think this is an opportunity to standardize, and I like the ability to retain numerical strings. I can't think off the top of my head where that might be important (maybe with strongly typed Config properties?) but my hunch is we need to support it and checking for the presence of quotes is a decent approach.

system/Config/BaseConfig.php
```diff
@@ -96,13 +96,39 @@ protected function initEnvValue(&$property, string $name, string $prefix, string
...
+                    case $lowered === 'null':
+                        $value = null;
+                        break;
+
+                    default:
+                        $value = trim($value, '\'"');
+                }
+
+                if (is_numeric($value)) {
+                    $value = strpos($value, '.') !== false ? (float) $value : (int) $value;
```
MGatner, line 127: I think this is accurate but can't test on mobile... `is_numeric()` accounts for multiple periods, right?
```php
is_numeric('4.1.8') // false
```

MGatner (changes requested):
One other thing you may have accounted for but just to be sure, from https://www.php.net/manual/en/function.is-numeric :
> 8.0.0	Numeric strings ending with whitespace ("42 ") will now return true. Previously, false was return instead

system/Config/BaseConfig.php
paulbalandan: The check returns false.
https://3v4l.org/nTQpm

paulbalandan:
> One other thing you may have accounted for but just to be sure, from https://www.php.net/manual/en/function.is-numeric :
> 
> > 8.0.0	Numeric strings ending with whitespace ("42 ") will now return true. Previously, false was return instead

The current and proposed behavior do not remove edge whitespaces, only edge quotes.
https://3v4l.org/p1At6

So, your example will have different values on php 8.

tests/system/Config/fixtures/.env
kenjis: The current code does not matter.
`12.45 === '12.45'` is wired.
And there is no way to set numerical strings.

kenjis:
> $this->propertyExpectsInt = (int) env('property');

The `BaseConfig` provides the feature to set properties of the object.
https://github.com/codeigniter4/CodeIgniter4/blob/968c0278a0ba4138cf3e64a2bbab5f3a800532d5/system/Config/BaseConfig.php#L72
https://github.com/codeigniter4/CodeIgniter4/blob/968c0278a0ba4138cf3e64a2bbab5f3a800532d5/system/Config/BaseConfig.php#L93

kenjis:
Please add test cases like these:
```
Foo.bar = '02471'
Foo.bar = 02471
Foo.bar = '1337e0'
Foo.bar = 1337e0
Foo.bar = ' 42'
Foo.bar = '42 '
```

tests/system/Config/fixtures/.env
kenjis: First of all, all environment variable's values are strings in PHP.
And we don't have to use DotEnv, we can set environment variables to web server.
So the DotEnv code does not matter.

Now we are going to change the interpretation of the values. We already changed when `true`/`false`.
We can decide the behavior for us.

kenjis: Yes, integer 12 and string 12 are different.
Not being able to use numeric strings for properties in the Config files is too big a restriction.

I just wanted to say, the following two are different things. They are related and must work with, but not one.
- DotEnv reads `.env` and sets environment variables in PHP
- BaseConfig fetches environment variables and sets properties

samsonasik:
@paulbalandan for database use case, is this will still works for password db: 123456 without quote?

```bash
database.default.password = 123456
```

kenjis:
#5779 was merged.

PR #5713, opened 2022-02-20
feat: can add route handler as callable

kenjis:
**Description**
- You can jump to the controller method with IDE.

E.g.:
```php
$routes->get('home', [Hello::class, 'index']);
$routes->get('product/(:num)/(:num)', [[Hello::class, 'index'], '$1/$2']);
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1137,6 +1137,10 @@ class RouteCollection implements RouteCollectionInterface
             $from = trim($from, '/');
         }
 
+        if (is_array($to)) {
+            $to = $this->processArrayCallableSyntax($from, $to);
+        }
+
         $options = array_merge($this->currentOptions ?? [], $options ?? []);
 
         // Route priority detect
@@ -1225,6 +1229,52 @@ class RouteCollection implements RouteCollectionInterface
         if (isset($options['redirect']) && is_numeric($options['redirect'])) {
             $this->routes['*'][$name]['redirect'] = $options['redirect'];
         }
+    }
+
+    /**
+     * @return string
+     */
+    private function processArrayCallableSyntax(string $from, array $to)
+    {
+        // [classname, method]
+        // eg, [Home::class, 'index']
+        if (is_callable($to, true, $callableName)) {
+            // If the route has placeholders, add params automatically.
+            $params = $this->getMethodParams($from);
+
+            return '\\' . $callableName . $params;
+        }
+
+        // [[classname, method], params]
+        // eg, [[Home::class, 'index'], '$1/$2']
+        if (
+            isset($to[0], $to[1])
+            && is_callable($to[0], true, $callableName)
+            && is_string($to[1])
+        ) {
+            $to = '\\' . $callableName . '/' . $to[1];
+        }
+
+        return $to;
+    }
+
+    /**
+     * Returns the method param string like `/$1/$2` for placeholders
+     */
+    private function getMethodParams(string $from): string
+    {
+        preg_match_all('/\(.+?\)/', $from, $matches);
+        $count = is_countable($matches[0]) ? count($matches[0]) : 0;
+
+        $params = '';
+
+        if ($count > 0) {
+            for ($i = 1; $i <= $count; $i++) {
+                $params .= '/$' . $i;
+            }
+        }
+
+        return $params;
     }
 
     /**
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1137,7 +1137,8 @@ class RouteCollection implements RouteCollectionInterface
             $from = trim($from, '/');
         }
 
-        if (is_array($to)) {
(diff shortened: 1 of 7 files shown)


mostafakhudair:
Please include additions in the user guide

iRedds:
``[[Home::class, 'index'], '$1/$2']`` doesn't look nice

What about?
```php
//[Home::class, 'index', '$1/$2'] 

[$class, $method, $params] = $to; 
is_callable([$class, $method], true, $callableName)

//... 
```php

**Comment by kenjis** [comment]
`[Home::class, 'index', '$1/$2']` seems to be nice.
But we cannot jump to `Home::index()` source code when we click `'index'`

Screenshot 2022-02-21 9 18 29
Screenshot 2022-02-21 9 18 44

**Comment by iRedds** [comment]
Yes, you are right.

**Review by paulbalandan — COMMENTED** [review]
This needs writeup in the docs.

**Comment by iRedds** [comment]
In theory. Can we drop ``$1/$2``?

**Comment by kenjis** [comment]
I will write the docs later.

**Comment by kenjis** [comment]
Added User Guide.

**Comment by iRedds** [comment]
@kenjis can you reply to the comment above?
I want to clarify. I meant that ``$1/$2`` is useless in my opinion.
```php
$routes->get('product/(:num)/(:num)', [Hello::class, 'index']);
```
Here we already have two arguments ``(:num)/(:num)``. Which can be passed to a controller method without the ``$1/$2`` notation.
The same principle as the closure works.
```php
$routes->get('product/(:num)/(:num)', fn ($param1, $param2) => 'do something');
```
i.e. we don't specify ``$1/$2`` and parameters are passed as is.

lonnieezell:
This looks like a great addition.  I would be in favor of removing the $1/$2 params as being specifically needed. While I've had times in the past where I needed to reorder or leave out some parameters, it's never been a big enough deal that it couldn't be worked around, and would clean this up a bit. Maybe make it optional? If it's not provided it defaults to the order given?

iRedds:
To skip the parameter, the developer can use non-capturing grouping.
For example, for the url ``product/1/2`` and the route ``product/(?::num)/(:num)``, the parameter will only be ``2``.

kenjis:
@iRedds @lonnieezell how about this?

iRedds:
It will not work correctly with a non-capturing group.
For ``product/(?::num)/(:num)`` the replacement string will be ``controller::method/$1/$2`` and we get the string ``controller::method/1/`` which gives parameter [1, ''].
Perhaps with nested groups there will be a similar situation.

It seems to me that if you use your approach, then you need to trim the slash in ``Router::checkRoutes()``.
Or you can add parameters directly to ``Router::checkRoutes()``
```php
} elseif (strpos($val, '$') === false && count($matches) > 1) {
    array_shift($matches);
    $val.= '/' . implode('/', $matches);
}

$this->setRequest(explode('/', $val));
```
This solution will also allow parameters to be applied to more than just the $to array. 
For example, ``$routes->get('x/(:num)', 'Home::index');``
But this is a quick example and I could miss the nuances.

kenjis:
I think regex is originally complex, and we do not need to support all the pattern.

- do not support non-capturing groups
- do not support nested groups

Or I don't mind to change the implementation if there is simple better one.

But if `$routes->get('x/(:num)', 'Home::index');` will add params automatically,
it is a breaking change.

kenjis:
Updated the documentation: https://github.com/codeigniter4/CodeIgniter4/pull/5713/files#diff-2a34c841c28af79b5550af42e3e5c24b337fa9ce9e90f8164000ac6c7d5ea7fe

lonnieezell (approved):
Looks great. Thanks!

system/Router/RouteCollection.php
```diff
@@ -1227,6 +1231,52 @@ protected function create(string $verb, string $from, $to, ?array $options = nul
         }
     }
 
+    /**
+     * @return string
+     */
+    private function processArrayCallableSyntax(string $from, array $to)
```
paulbalandan, line 1237:
```suggestion
    private function processArrayCallableSyntax(string $from, array $to): string
```

system/Router/RouteCollection.php
```diff
@@ -1227,6 +1231,52 @@ protected function create(string $verb, string $from, $to, ?array $options = nul
...
+        preg_match_all('/\(.+?\)/', $from, $matches);
+        $count = is_countable($matches[0]) ? count($matches[0]) : 0;
+
+        $params = '';
+
+        if ($count > 0) {
+            for ($i = 1; $i <= $count; $i++) {
+                $params .= '/$' . $i;
+            }
+        }
```
paulbalandan, line 1275: No need for this if. If count is 0, the for will not execute.

paulbalandan (approved):

kenjis:
@paulbalandan The commit you approved failed tests.
Fixed.

paulbalandan (approved):

PR #5715, opened 2022-02-21
Fix alignment on few buttons in debug toobar and change color

Nudasoft:
I don't know if this was a design choice or not. But it's looks bit broken with few alignment issues. So I have fixed that and made little cosmetic color changes to make those elements go align with CI4 color theme.

Before...

After...

Thanks.

### DIFF
--- a/admin/css/debug-toolbar/_theme-dark.scss
+++ b/admin/css/debug-toolbar/_theme-dark.scss
@@ -129,8 +129,8 @@
         }
 
         .badge {
-            background-color: $g-blue;
-            color: $m-gray;
+            background-color: $g-red;
+            color: $t-light;
         }
     }
 
--- a/admin/css/debug-toolbar/_theme-light.scss
+++ b/admin/css/debug-toolbar/_theme-light.scss
@@ -125,7 +125,7 @@
         }
 
         .badge {
-            background-color: $g-blue;
+            background-color: $g-red;
             color: $t-light;
         }
     }
--- a/admin/css/debug-toolbar/toolbar.scss
+++ b/admin/css/debug-toolbar/toolbar.scss
@@ -77,16 +77,9 @@
 
     // General elements
     h1 {
-        bottom: 0;
-        display: inline-block;
-        font-size: $base-size - 2;
+        display: flex;
         font-weight: normal;
-        margin: 0 16px 0 0;
-        padding: 0;
-        position: absolute;
-        right: 30px;
-        text-align: left;
-        top: 0;
+        margin: 0 0 0 auto;
 
         svg {
             width: 16px;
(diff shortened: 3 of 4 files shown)


kenjis:
When viewed in Firefox 97.0.1, the entire page shifts slightly.

Without Debug Bar:
screenshot 2022-02-21 17 55 51

Before:
screenshot 2022-02-21 17 55 54

After:
screenshot 2022-02-21 17 55 57

Nudasoft:
This layout shifting thing is not related to debug bar. its how firefox render things. no matter what webpage you compare between firefox and chorme you will see that shifting. It's completely unrelated.

kenjis:
I compared in Firefox *Before* and *After*, and the layout shits slightly.
I compared in Chrome *Before* and *After*, and the layout are exactly the same.

Nudasoft:
I cant really reproduce what you are saying. It's pretty much the same before and after.

Firefox 97.01 on Windows10
toolbar

And these are the only changes I've done that related to layout/typography to the toolbar.css file.

#debug-bar h1 {
    display: flex;
    font-weight: normal;
    margin: 0 0 0 auto;
}

#debug-bar #debug-bar-link {
    padding: 6px;
    display: flex;
}

As you can see I'm not targeting anything globally. Only two very specific elements. So it's pretty much impossible to have any unexpected layout issues.

Can you assign someone else to check this pr? 

Thanks.

kenjis:
Thank you for confirmation.
It seems the output of Firefox are a bit different between macOS and Windows.

paulbalandan (changes requested):
Please update also the CSS in **admin/css/debug-toolbar/**

kenjis:
Ah! We use SASS.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/css.md

paulbalandan:
> Ah! We use SASS. https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/css.md

Yes, but sadly we don't have a workflow that updates the CSS from the SASS/SCSS.

sfadschm:
> > Ah! We use SASS. https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/css.md
> 
> Yes, but sadly we don't have a workflow that updates the CSS from the SASS/SCSS.

Right, but changes should still be made to the SASS files and the compiled locally to produce the CSS.
Otherwise we will run into issues if the next contributer follows this way.

paulbalandan (approved):
I've compiled the changed SCSS files locally using Dart SASS and found some minor formatting issues. I think I'll tackle that in a separate PR together with a workflow. Anyway, this looks good. Checking also the final output seems fine to me.

kenjis:
I've checked the layout in Firefox. 
I don't know why but the slight shift was gone. Both are exact same.

kenjis (approved):

paulbalandan:
Thank you, @Nudasoft

PR #5720, opened 2022-02-22
docs: fix @param of view cell

### LINKED ISSUE #5719: Bug: error in view_cell() doc comment
### PHP Version

7.4

### CodeIgniter4 Version

4.1.5

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

i have error in phpstan analyse

_Parameter #2 $params of function view_cell expects null, string given._

on line:
`return view_cell('\App\Models\CalendrierModel::vueModal', 'id=' . $id);`

i think the view_cell doc comment is not good
```
/**
     * View cells are used within views to insert HTML chunks that are managed
     * by other classes.
     *
     * @param null $params
     *
     * @throws ReflectionException
     */
    function view_cell(string $library, $params = null, int $ttl = 0, ?string $cacheName = null): string
    {
        return Services::viewcell()
            ->render($library, $params, $ttl, $cacheName);
    }
```

### Steps to Reproduce

`return view_cell('\App\Models\CalendrierModel::vueModal', 'id=' . $id);`
and phpstan analyse

### Expected Output

Parameter 2 $params of function view_cell will accept string or array

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5719

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -1125,7 +1125,7 @@ if (! function_exists('view_cell')) {
      * View cells are used within views to insert HTML chunks that are managed
      * by other classes.
      *
-     * @param null $params
+     * @param array|string|null $params
      *
      * @throws ReflectionException
      */
--- a/system/View/Cell.php
+++ b/system/View/Cell.php
@@ -62,7 +62,7 @@ class Cell
     /**
      * Render a cell, returning its body as a string.
      *
-     * @param null $params
+     * @param array|string|null $params
      *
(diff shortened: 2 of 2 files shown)


paulbalandan (approved):

paul45:
yes this is fine for #5719

PR #5724, opened 2022-02-22
[Docs] Extract new examples from validation.

sfadschm:
This is now cherry-picked and standalone.

[x] Extract new validation examples.


sfadschm:
@kenjis 
- I think the other new example is not executable to I left it in the rst.
- Renumeration will then be done later for all modules, when the small changes are through.

kenjis (approved):

PR #5725, opened 2022-02-22
[Docs] Replace utf-8 characters.

sfadschm:
This is now standalone cherry-picked.

[x] Remove occurences of . :.
[x] Replace utf-8 characters (except the euro and pound signs).


sfadschm:
Rebased.

kenjis (approved):

kenjis:
By the way, how did you find the utf-8 characters?

sfadschm:
> By the way, how did you find the utf-8 characters?

I have a Linux-subsystem on Windows (WSL). You can use the file-command. It will tell you if a text file is ASCII or UTF-8.
Then I just changed the file encoding and replaced the broken characters.

Unfortunately, this does not work for € and the pound sterling sign, as they are not in 1252.

PR #5726, opened 2022-02-22
refactor: remove `&` before $db

kenjis:
**Description**
- `$db` is an object. We don't need to use reference. 
  - see https://www.php.net/manual/en/language.oop5.references.php

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -270,7 +270,7 @@ class BaseBuilder
      *
      * @throws DatabaseException
      */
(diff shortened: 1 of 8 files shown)


paulbalandan (approved):
My thoughts, exactly.

kenjis:
In my understanding, we only need to use the reference when we have to change the property outside.
https://3v4l.org/QN7F9

However, the usage destroys the encapsulation. Generally speaking, it is not good design.

In this `$db` case, I don't see any reason to use the reference.

PR #5729, opened 2022-02-23
Forge use statement

mostafakhudair:


lonnieezell (approved):

sfadschm:
I think there used to be a static analysis rule in either rector or cs-fixer which should handle this.

kenjis (approved):

kenjis:
Is it possible for rector or cs-fixer to resolve classname conflict?

PR #5730, opened 2022-02-24
fix: forceGlobalSecureRequests break URI schemes other than HTTP

### LINKED ISSUE #5728: Bug: enabled forceGlobalSecureRequests prevents uri schemes besides http/https
[HTTP/URI.php#L599](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/URI.php#L599) assumes the only URI schemes are `http` and `https` and therefore ignores other URI schemes when using `forceGlobalSecureRequests`.

This is allread tackled in [Helpers/url_helper.php#L65](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Helpers/url_helper.php#L65) and  [HTTP/IncomingRequest.php#L411](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/HTTP/IncomingRequest.php#L411).
Since this is required in multiple places and therfore error-prone, maybe consider outsourcing this to one tested static method.
Even introduced this differing behaviors in the same commit https://github.com/codeigniter4/CodeIgniter4/commit/89d588fab6dadc904c76859b4a8a86a3e8d11189

```php
d( (string) new URI('ftp://localhost') )
// Wrong Output -> "https://localhost"
// Expected Output -> "ftp://localhost"
```

kenjis:
**Description**
Fixes #5728

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -581,11 +581,30 @@ class URI
         $path   = $this->getPath();
         $scheme = $this->getScheme();
 
+        // If the hosts matches then assume this should be relative to baseURL
+        [$scheme, $path] = $this->changeSchemeAndPath($scheme, $path);
+
+        return static::createURIString(
+            $scheme,
+            $this->getAuthority(),
+            $path, // Absolute URIs should use a "/" for an empty path
+            $this->getQuery(),
+            $this->getFragment()
+        );
+    }
+
+    /**
+     * Change the path (and scheme) assuming URIs with the same host as baseURL
+     * should be relative to the project's configuration.
+     *
+     * @deprecated This method will be deleted.
+     */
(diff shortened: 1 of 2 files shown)


paulbalandan (approved):

kenjis:
@MGatner 
> URI should not be in the business of dealing with baseURL at all, but since this is currently the case we can't remove it entirely. #4646

What's happened if it is removed?

MGatner:
There's a lot of functionality based on `URI` and people have been told to use it for working with their project URLs. I even made a PR at one point to split these into two classes but ultimately it was too complicated. For now we need to leave this and plan to change it in version 5.

MGatner (approved):

PR #5736, opened 2022-02-24
Feature: Subquery for SELECT

iRedds:
**Description**
Added ``BaseBuilder::selectSubquery(BaseBuilder $subquery, string $as)`` method

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


user_guide_src/source/changelogs/v4.2.0.rst
```diff
@@ -24,6 +24,7 @@ Enhancements
     - See :ref:`content-security-policy` for details.
 - New View Decorators allow modifying the generated HTML prior to caching.
 - Added Subqueries in the FROM section. See :ref:`query-builder-from-subquery`.
+- Added Subqueries in the SELECT section.
```
lonnieezell, line 27: Please add link to the docs  like the surrounding changes have with theirs.

lonnieezell (changes requested):
Man, I just struggled with this on a different framework for the day job and wish this would have been available there. Nice addition, just one small ask.

user_guide_src/source/database/query_builder.rst
```diff
@@ -167,6 +167,8 @@ the resulting field.
 .. literalinclude:: query_builder/014.php
    :lines: 2-
 
+.. _query-builder-select-subquery:
```
kenjis, line 170: @iRedds The label must  precede a section header.

`**$builder->selectSubquery()**` is not a section.

system/Database/BaseBuilder.php
```diff
@@ -445,6 +445,22 @@ public function selectCount(string $select = '', string $alias = '')
...
 
+    /**
+     * Adds a subquery to the selection
+     *
+     * @return static
+     */
+    public function selectSubquery(BaseBuilder $subquery, string $as)
+    {
+        if (! $this->isSubquery($subquery)) {
+            throw new DatabaseException('The BaseBuilder::selectSubquery method expects a BaseBuilder instance');
```
kenjis, line 456:
```suggestion
            throw new DatabaseException('The BaseBuilder::selectSubquery() method expects a BaseBuilder instance');
```

user_guide_src/source/database/query_builder/098.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+$subquery = $db->table('countries')->select('name')->where('id', 1);
+$builder  = $db->table('users')->select('name')->selectSubquery($subquery, 'country');
+$query = $builder->get();
+
+// Produces: SELECT `name`, (SELECT `name` FROM `countries` WHERE `id` = 1) AS `country` FROM `users`
```
kenjis, line 7: Please add a newline.

kenjis (dismissed):

system/Database/BaseBuilder.php
```diff
@@ -445,6 +445,24 @@ public function selectCount(string $select = '', string $alias = '')
         return $this->maxMinAvgSum($select, $alias, 'COUNT');
     }
 
+    /**
+     * Adds a subquery to the selection
+     *
+     * @return static
+     */
+    public function selectSubquery(BaseBuilder $subquery, string $as)
```
kenjis, line 453:
```suggestion
    public function selectSubquery(BaseBuilder $subquery, string $as): self
```

system/Database/BaseBuilder.php
```diff
@@ -445,6 +445,24 @@ public function selectCount(string $select = '', string $alias = '')
         return $this->maxMinAvgSum($select, $alias, 'COUNT');
     }
 
+    /**
+     * Adds a subquery to the selection
+     *
+     * @return static
```
kenjis, line 451:
```suggestion
     * @return $this
```

iRedds: If you specify a return type after the method, then the rector removes the @return from the docblock.
```
     /**
      * Adds a subquery to the selection
      *
-     * @return static
      */
     public function selectSubquery(BaseBuilder $subquery, string $as): self
     {

Applied rules:
 * RemoveUselessReturnTagRector

```

kenjis: Okay.

kenjis (approved):
LGTM

paulbalandan (approved):

system/Database/BaseBuilder.php
```diff
@@ -445,6 +445,22 @@ public function selectCount(string $select = '', string $alias = '')
         return $this->maxMinAvgSum($select, $alias, 'COUNT');
     }
 
+    /**
+     * Adds a subquery to the selection
+     */
+    public function selectSubquery(BaseBuilder $subquery, string $as): self
+    {
+        if (! $this->isSubquery($subquery)) {
```
samsonasik, line 453: looking at mobile, it seems `$subquery` already type hinted, passed non `BaseBuilder` will already cause instance error so isSubquery() check no longer needed?

iRedds: You're right, this check was unnecessary

samsonasik (approved):

samsonasik:
@iRedds looking good, how about support both `BaseBuilder` and `Closure` with `@param BaseBuilder|Closure $subquery` with remove param type hint so the `->isSubquery()` check will make sense, as buildSubquery() support both types?

iRedds:
Honestly, I do not know. It seems to me that using a closure as a non-last argument worsens the readability of the code.

samsonasik:
Ok then 👍

houjl04:
The query statement “select... for update”, how do I do that? Is there a method?

PR #5737, opened 2022-02-24
Delete stale.yml

paulbalandan:
**Description**
We have a `stale.yml` created over 4 years ago which I don't see being used anywhere. Even a workflow using `actions/stale` is not present. So, I assume this is unnecessary.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


kenjis (approved):

samsonasik:
Thank you @paulbalandan

PR #5763, opened 2022-03-03
fix: rename Entity::setAttributes() to avoid name collision

### LINKED ISSUE #5762: Bug: Entity can't handle column named `attributes`
### CodeIgniter4 Version

4.1.9

### What happened?

Names collision.

### Steps to Reproduce

```php
$e->a = 1;
dd($e->toRawArray());

//$e->toRawArray() array (3)
//   0 => integer 1
//   1 => integer 2
//   2 => integer 3
```

### Expected Output

```
//$e->toRawArray() array (2)
//    a => integer 1
//    attributes => array (3)
//        0 => integer 1
//        1 => integer 2
//        2 => integer 3
```

### Anything else?

```php
{
}
```

kenjis:
**Description**
Fixes #5762

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity/Entity.php
+++ b/system/Entity/Entity.php
@@ -274,7 +274,7 @@ class Entity implements JsonSerializable
      *
      * @return $this
      */
-    public function setAttributes(array $data)
+    public function injectRawData(array $data)
     {
         $this->attributes = $data;
 
--- a/system/Database/MySQLi/Result.php
+++ b/system/Database/MySQLi/Result.php
@@ -142,7 +142,7 @@ class Result extends BaseResult
     protected function fetchObject(string $className = 'stdClass')
     {
         if (is_subclass_of($className, Entity::class)) {
-            return empty($data = $this->fetchAssoc()) ? false : (new $className())->setAttributes($data);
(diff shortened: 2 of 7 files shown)


system/Entity/Entity.php
```diff
@@ -274,7 +274,7 @@ public function hasChanged(?string $key = null): bool
      *
      * @return $this
      */
-    public function setAttributes(array $data)
+    public function injectRawData(array $data)
```
mostafakhudair, line 277: removing public function is BC, isn't it?

kenjis: Yes. I added `beraking change` label.

MGatner:
@kenjis Is this still relevant after #5509?

kenjis:
At least, this PR won't be merged into 4.x.
But I want to *fix* this bad design in the future.

najdanovicivan:
I like the injectRawData naming. It can be merged into 4.x if setAttributes is still kept as deprecated so that we can upgrade before the next major version.

kenjis:
I sent PR #7208

PR #5764, opened 2022-03-03
refactor: cast string to int in OpenSSLHandler

kenjis:
**Description**
- to suppress PHPStan 1.4.7 error

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Encryption/Handlers/OpenSSLHandler.php
+++ b/system/Encryption/Handlers/OpenSSLHandler.php
@@ -89,7 +89,7 @@ class OpenSSLHandler extends BaseHandler
         // derive a secret key
         $secret = \hash_hkdf($this->digest, $this->key);
 
-        $hmacLength = self::substr($this->digest, 3) / 8;
+        $hmacLength = (int) self::substr($this->digest, 3) / 8;
         $hmacKey    = self::substr($data, 0, $hmacLength);
         $data       = self::substr($data, $hmacLength);
         $hmacCalc   = \hash_hmac($this->digest, $data, $secret, true);


system/Encryption/Handlers/OpenSSLHandler.php
```diff
@@ -89,7 +89,7 @@ public function decrypt($data, $params = null)
         // derive a secret key
         $secret = \hash_hkdf($this->digest, $this->key);
 
-        $hmacLength = self::substr($this->digest, 3) / 8;
+        $hmacLength = (int) self::substr($this->digest, 3) / 8;
         $hmacKey    = self::substr($data, 0, $hmacLength);
         $data       = self::substr($data, $hmacLength);
         $hmacCalc   = \hash_hmac($this->digest, $data, $secret, true);
```
paulbalandan, line 95: I think this should be refactored completely (and the equivalent encrypt logic). `$this->digest` is also used by `hash_hmac()` as its algo, so it follows that `$this->digest` should accept any available digest (try `hash_hmac_algos()`). Now, `hash_hmac_algos()`'s list of digests is not confined to `SHA512` and the likes. See https://www.php.net/manual/en/function.hash-hmac-algos.php. So, if a user chooses `ripemd160` as his digest, `self::substr($this->digest, 3)` will give `emd160`, which is why phpstan is giving the error. Casting to `int` will give `0` which is not correct, see https://3v4l.org/8XP68.

kenjis: I don't know why the current implementation is like that, and you say correct thing.
But this PR just refactored the current implementation, and without this all GitHub Action checks will fail.
I hope to merge this for now if this PR does not introduce any new bug.

kenjis: Ci3 supports only from sha224 to sha512:
https://github.com/bcit-ci/CodeIgniter/blob/3378180483140fa9251e50843bda15a508e85e89/system/libraries/Encryption.php#L132-L135
This is why?

paulbalandan: Hmm.. Maybe a port from CI3. I say we just include the error in `phpstan-baseline.neon.dist`. At least, if a user will use `ripemd160` or others in their apps then PHP will error on `Binary operation between string and 8 ...` and then report to GitHub. By then we can refactor the logic completely.

MGatner: I'm okay with this as a "quick fix for static analysis", maybe with a new Issue to look into a refactor. But I will let you two work this out, I've never touched the encryption stuff.

kenjis:
Ignored phpstan 1.4.7 notice by #5767

kenjis:
I created an issue: #5769

PR #5765, opened 2022-03-03
chore: add .user-guide.php-cs-fixer.dist.php for sample code in docs

kenjis:
**Description**
- add `.user-guide.php-cs-fixer.dist.php`
- add user guide cs check in git pre-commit hook
- add user guide cs check in GitHub Action
- update `composer cs-fix`

I propose the following before fixing sample code:
- move CI3 sample code in Upgrading into the folder `ci3sample/`
  - we need exclude the CI3 sample code
  - #5771
- use `/* ... */` style comment for explanation of the code result
  - The comments are not normal comments, so it is better not to use `//`
  - #5770

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/.github/workflows/test-coding-standards.yml
+++ b/.github/workflows/test-coding-standards.yml
@@ -56,3 +56,6 @@ jobs:
 
       - name: Run lint on `system/`, `tests`, `utils/`, and root PHP files
         run: vendor/bin/php-cs-fixer fix --verbose --ansi --dry-run --using-cache=no --diff
+
+      - name: Run lint on `user_guide_src/source/`
+        run: vendor/bin/php-cs-fixer fix --verbose --ansi --dry-run --config=.user-guide.php-cs-fixer.dist.php --using-cache=no --diff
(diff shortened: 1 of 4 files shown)


paulbalandan:
I agree with the proposals. Side comment, latest master of php-cs-fixer has its own implementation of `space_after_comment_start` so on next release the use on `//` will be strictly for single-lines only. So, as early as now, multiline comments should start using the `/* */` way and reserve `//` for single-line comments.

kenjis:
> This would then be applied and integrated into the pre-commit hook, right?

I added the check in the pre-commit.

kenjis:
I sent PRs #5770, #5771

paulbalandan (approved):

sfadschm:
Applying the fixer will then be another PR?

kenjis:
> Applying the fixer will then be another PR?

Yes, because the difference is huge.
But as you say, the code fixes need to be committed soon right after this PR's merging.

kenjis:
I created #5773 to apply the fixer.

kenjis:
Added GitHub Action update.

paulbalandan (approved):

PR #5766, opened 2022-03-03
Wrap $this->request->getMethod() with strtolower()

borachcha:
Just to be safe.


kenjis:
@borachcha It seems the commit includes other than the commit message says.
Please remove unneeded changes.

kenjis:
Close via #5963

PR #5767, opened 2022-03-03
[Rector] Update to Rector 0.12.17 and Re-run it

samsonasik:
- Update Rector to 0.12.17
- Re-run it to apply more removal on unneded variable in assign
- Apply typed property on final class protected property.
- Ignore phpstan 1.4.7 notice on `Binary operation "/" between string and 8 results in an error.` on system/Encryption/Handlers/OpenSSLHandler.php

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "phpstan/phpstan": "^1.0",
         "phpunit/phpunit": "^9.1",
         "predis/predis": "^1.1",
-        "rector/rector": "0.12.16"
+        "rector/rector": "0.12.17"
     },
     "suggest": {
         "ext-fileinfo": "Improves mime type detection for files"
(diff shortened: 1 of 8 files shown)


samsonasik:
All green 🎉

paulbalandan:
I think this should supersede #5764

paulbalandan (approved):
@samsonasik how about your comment in my closed PR that non-static call on static methods is a Rector bug, is that resolved?

samsonasik:
@paulbalandan probably not yet, feel free to create failing test case to rector-src so we can try to solve https://github.com/rectorphp/rector-src, the tests mostly under rules-tests.

samsonasik:
@paulbalandan we can update `OpenSSLHandler` when PR https://github.com/codeigniter4/CodeIgniter4/pull/5764 updated, I ignored PHPStan notice for now to avoid regression in functionality.

kenjis (approved):

kenjis:
@samsonasik Thank you!

PR #5778, opened 2022-03-06
fix: table name is double prefixed when LIKE clause

### LINKED ISSUE #5775: Bug: table has double prefix when LIKE clause
### PHP Version

7.3

### CodeIgniter4 Version

4.1.9 + dev

Git

Windows

### Which server did you use?

cli

### Database

_No response_

### What happened?

see https://forum.codeigniter.com/thread-81454.html

### Steps to Reproduce

```php
// database.default.DBPrefix = db_

db_connect()->table('test')->like('test.field', 'string')->getCompiledSelect();

// SELECT * FROM `db_test` WHERE `db_``db_test`.`field` LIKE '%string%' ESCAPE '!' 

```

### Expected Output

```
SELECT * FROM `db_test` WHERE `db_test`.`field` LIKE '%string%' ESCAPE '!' 
```

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5775

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1051,7 +1051,7 @@ class BaseBuilder
                 $bind = $this->setBind($k, "%{$v}%", $escape);
             }
 
-            $likeStatement = $this->_like_statement($prefix, $this->db->protectIdentifiers($k, false, $escape), $not, $bind, $insensitiveSearch);
+            $likeStatement = $this->_like_statement($prefix, $k, $not, $bind, $insensitiveSearch);
 
             // some platforms require an escape sequence definition for LIKE wildcards
             if ($escape === true && $this->db->likeEscapeStr !== '') {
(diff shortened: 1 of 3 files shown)


iRedds:
Double escaping when using the LIKE clause.
https://github.com/codeigniter4/CodeIgniter4/blob/f969513ddedf14409bede320a1ff2f34434b9dc0/system/Database/BaseBuilder.php#L1054

https://github.com/codeigniter4/CodeIgniter4/blob/f969513ddedf14409bede320a1ff2f34434b9dc0/system/Database/BaseBuilder.php#L2452

Does it make sense?

kenjis:
It is from #5170.
We might have made a mistake.
We need to reconsider.

kenjis:
Removed double escape.

samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #5779, opened 2022-03-06
fix: [BaseConfig] string value is set from environment variable even if it should be int/float

### LINKED ISSUE #5688: Bug: Ajax request makes the session expire when app.sessionExpiration is set to 0

kenjis:
**Description**
Fixes #5688
Supersede #5704,  #5691 and #5690

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Config/fixtures/.env
+++ b/tests/system/Config/fixtures/.env
@@ -33,9 +33,12 @@ SimpleConfig.crew.pilot = Wash
 SimpleConfig.crew.comms = true
 SimpleConfig.crew.doctor = false
 
-SimpleConfig.float = '0.0'
-SimpleConfig.int = '999'
+# The default value's type in the Config class is float, so it will be converted to float
+SimpleConfig.float = 0.0
+# The default value's type in the Config class is int, so it will be converted to int
+SimpleConfig.int = 999
 
 SimpleConfig.password = 123456
 
+# The property type in the Config class is ?int, so it will be converted to int by PHP
 SimpleConfig.size=20
--- a/tests/system/Config/fixtures/.env
+++ b/tests/system/Config/fixtures/.env
@@ -32,3 +32,8 @@ SimpleConfig.crew.captain = Malcolm
 SimpleConfig.crew.pilot = Wash
 SimpleConfig.crew.comms = true
 SimpleConfig.crew.doctor = false
+
+SimpleConfig.float = '0.0'
+SimpleConfig.int = '999'
+
+SimpleConfig.password = 123456
--- a/tests/system/Config/fixtures/.env
+++ b/tests/system/Config/fixtures/.env
@@ -37,3 +37,5 @@ SimpleConfig.float = '0.0'
 SimpleConfig.int = '999'
 
 SimpleConfig.password = 123456
+
+SimpleConfig.size=20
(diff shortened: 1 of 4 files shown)


MGatner (commented):
This is much cleaner, but what about null handling? I think it could be quite common to have a Config with `public property $foo;` as `int|null` and since no initial value was set we would end up with a `string`.

kenjis:
@MGatner 
> what about null handling?

This does nothing.

> I think it could be quite common to have a Config with public property $foo; as int|null

Not so common. I think it might be bad design.
Now `null` is converted to a string. This PR changes nothing about null handling.
It is out of scope.

Do you really need `null` handling?

kenjis:
@MGatner If you type the property like your sample, it is no problem.
See [the test](https://github.com/codeigniter4/CodeIgniter4/pull/5779/files#diff-6eb759fa5bcf6f5d404cd43003f3417974fe0c82222818f4cbbd4c7e651ca1baR122) passes.

If you don't type it like the following, string `20` is set in the `$size`.
```php
class Widget extends BaseConfig
{
    property $size = null;
}
```

kenjis:
It seems typed properties would solve most of the type problems.
So I think we don't need to add extra logic to convert types.

This PR is enough to the current problems.

MGatner (approved):

tests/system/Config/fixtures/.env
```diff
@@ -32,3 +32,10 @@ SimpleConfig.crew.captain = Malcolm
 SimpleConfig.crew.pilot = Wash
 SimpleConfig.crew.comms = true
 SimpleConfig.crew.doctor = false
+
+SimpleConfig.float = '0.0'
+SimpleConfig.int = '999'
```
paulbalandan, line 37: Per your https://github.com/codeigniter4/CodeIgniter4/pull/5704#discussion_r808983149, this should result in strings since encapsed in quotes. However, in lines 64-65 of the tests, those are tested as float and int respectively.

Isn't it contradictory? The reason why my PR was stalled is because of trying to find a solution to that distinction (i.e., string 12 and int 12) and this PR is not addressing that.

kenjis: Yes, I've changed my mind.

After all, the correct type is the property type in the Config class.

But I don't recommend to use `'` for int/float value, I change the test fixture.

kenjis:
@paulbalandan Can I merge this?
Or you still have something wrong?

MGatner:
@paulbalandan still keen on your final opinion. @kenjis if he is unavailable over the next 48 hours let's proceed.

PR #5781, opened 2022-03-07
fix: Entity can't handle column named attributes

### LINKED ISSUE #5762: Bug: Entity can't handle column named `attributes`

kenjis:
**Description**
Fixes #5762
Supersede #5763
- fix Entity can't handle column named attributes
- add new setter method "`_set` + key" for a attribute

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity/Entity.php
+++ b/system/Entity/Entity.php
@@ -439,12 +439,21 @@ class Entity implements JsonSerializable
 
         $value = $this->castAs($value, $key, 'set');
 
-        // if a set* method exists for this key, use that method to
+        // if a setter method exists for this key, use that method to
         // insert this value. should be outside $isNullable check,
         // so maybe wants to do sth with null value automatically
         $method = 'set' . str_replace(' ', '', ucwords(str_replace(['-', '_'], ' ', $key)));
 
-        if (method_exists($this, $method)) {
+        // If a "`_set` + $key" method exists, it is a setter.
+        if (method_exists($this, '_' . $method)) {
+            $this->{$method}($value);
+
+            return $this;
+        }
+
+        // If a "`set` + $key" method exists, it is also a setter.
+        // But setAttributes() is not a setter for this.
+        if ($key !== 'attributes' && method_exists($this, $method)) {
             $this->{$method}($value);
 
             return $this;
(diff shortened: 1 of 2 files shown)


MGatner:
I like this solution better, but let's wrap up the Issue conversation first. I think it is quite possible that someone is using `$entity->attributes = $_POST` not realizing the current behavior is "unintended".

system/Entity/Entity.php
```diff
@@ -439,12 +439,21 @@ public function __set(string $key, $value = null)
...
+        // If a "`_set` + $key" method exists, it is a setter.
+        if (method_exists($this, '_' . $method)) {
+            $this->{$method}($value);
+
+            return $this;
+        }
+
+        // If a "`set` + $key" method exists, it is also a setter.
+        // But setAttributes() is not a setter for this.
+        if ($key !== 'attributes' && method_exists($this, $method)) {
```
mostafakhudair, line 456:
```suggestion
        // If a "`_set` + $key" method exists, it is a setter,  or
        // if a "`set` + $key" method exists, it is also a setter.
        // But setAttributes() is not a setter for this.
        if (method_exists($this, '_' . $method) || ($key !== 'attributes' && method_exists($this, $method))) {
```

mostafakhudair: Let's make code shorter,

kenjis: Thanks, I found a bug in my code.

Why do you like shorter code?

MGatner:
Maybe mark this one to revisit for version 5?

kenjis:
Added a note: https://github.com/codeigniter4/CodeIgniter4/projects/6#card-79354180

PR #5793, opened 2022-03-11
fix: Publisher $restrictions regex to FCPATH

kenjis:
**Description**
- fix regex

Related: #5792

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Publisher.php
+++ b/app/Config/Publisher.php
@@ -23,6 +23,6 @@ class Publisher extends BasePublisher
      */
     public $restrictions = [
         ROOTPATH => '*',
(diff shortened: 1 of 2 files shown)


haugli92:
Add `|woff2?|`. Im not the best on regex, can someone validate it?

kenjis:
@haugli92 Thanks!
Added.

kenjis:
@haugli92 Thanks!
Added.

samsonasik (approved):

MGatner:
I missed this one but I think it is now unduly restricting `.htm` files (that was the original intent of `htm?`. Do we need `htm??` for some reason?

MGatner:
Also `woff2` should've been covered by `woff?`, right?

MGatner:
Ah, nope - I'm a dummy, never mind me. 🤦‍♂️

PR #5796, opened 2022-03-11
[Docs] Renumerate examples.

sfadschm:
**Description**
This renumerates the current php examples in the docs sources.

I decided to remove the `renumerate` script as it cannot handle, e.g., the "ci3sample" subfolders and is generaally prone to errors.

I suggest we use the current state of the docs as a starting point and revisit enumeration if it gets messy with time. 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


kenjis:
> CodeIgniter4/user_guide_src/source/database/utilities.rst:19: WARNING: Could not lex literal_block as "html+php". Highlighting skipped.

sfadschm:
> > CodeIgniter4/user_guide_src/source/database/utilities.rst:19: WARNING: Could not lex literal_block as "html+php". Highlighting skipped.
> 

Oh yeah, forgot to pin that.
This however also happens before this PR for me...
I will look into it tomorrow.

kenjis:
Okay, the warning is not because of this PR.

kenjis (approved):
I've confirmed the output html is exactly the same.

sfadschm:
Php and rst syntax looks fine.
Maybe it cannot handle the direct call to the closure return? This is quite uncommon syntax I think:

`...) (...)`

sfadschm:
> Okay, the warning is not because of this PR.
> 

👍
Why does the example not use a "normal" model but the Closure syntax? Is there any benefit?

kenjis:
> Why does the example not use a "normal" model but the Closure syntax? Is there any benefit?

It seems no.
https://codeigniter4.github.io/CodeIgniter4/database/utilities.html#get-xml-from-result
The sample code is very tricky. It is better to rewrite.

PR #5798, opened 2022-03-12
fix: Timer::getElapsedTime() returns incorrect value

### LINKED ISSUE #5797: Bug: Timer::getElapsedTime() returns incorrect value
### PHP Version

8.1

### CodeIgniter4 Version

latest

Git

Linux

### Which server did you use?

apache

### Database

_No response_

### What happened?

There is a bug in `system/Debug/Timer.php`'s `getElapsedTime` function when elapsed times are `>= 1,000` seconds.

(This bug has been in CI since the 3.0 days, in Benchmark.php's elapsed_time function.)

The code is currently:
```php
return (float) number_format($timer['end'] - $timer['start'], $decimals);
```

note that:
```php
$number = (float) number_format(1000 - 1, 4); // returns 999
$number = (float) number_format(1001 - 1, 4); // returns ... 1
```

This is because of the default thousands separator of number_format. To fix, use no thousands separator:
```php
$number = (float) number_format(1001 - 1, 4, '.', '');
```

which returns the expected 1000.0000

However, even without the cast to (float), if the return value is compared to an integer:

```php
if($elapsed_time <= 300) ...
```

this code will trigger, as 1,000.0000 <= 300 . No matter of casting that I have tried will fix this.

### Steps to Reproduce

Call timer start() , then 1,0001 seconds later, call timer stop(). Compare the return value of getElapsedTime to say 5. It will be < 5

### Expected Output

getElapsedTime should not return numbers decorated with a thousands separator, as it messes with comparisons against other integers.

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5797

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Timer.php
+++ b/system/Debug/Timer.php
@@ -78,7 +78,7 @@ class Timer
      * @param string $name     The name of the timer.
      * @param int    $decimals Number of decimal places.
      *
-     * @return float|null Returns null if timer exists by that name.
+     * @return float|null Returns null if timer does not exist by that name.
      *                    Returns a float representing the number of
      *                    seconds elapsed while that timer was running.
      */
--- a/system/Debug/Timer.php
+++ b/system/Debug/Timer.php
@@ -96,7 +96,7 @@ class Timer
             $timer['end'] = microtime(true);
         }
 
-        return (float) number_format($timer['end'] - $timer['start'], $decimals);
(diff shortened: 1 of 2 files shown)


samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #5803, opened 2022-03-16
fix: [Pager] can't get correct current page from segment

kenjis:
**Description**
From https://forum.codeigniter.com/thread-81531.html
- `setSegment()` recalculates the current page
- `calculateCurrentPage()` can be called more than once

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #5804, opened 2022-03-16
fix: [Validation] valid_date ErrorException when the field is not sent

kenjis:
**Description**
From https://forum.codeigniter.com/thread-81505-post-394652.html
```
DateTime::createFromFormat(): Passing null to parameter #2 ($datetime) of type string is deprecated
```

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -333,7 +333,7 @@ class FormatRules
      */
     public function valid_date(?string $str = null, ?string $format = null): bool
     {
-        if (($str === null)) {
+        if ($str === null) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


system/Validation/FormatRules.php
```diff
@@ -333,6 +333,10 @@ public function valid_url_strict(?string $str = null, ?string $validSchemes = nu
      */
     public function valid_date(?string $str = null, ?string $format = null): bool
     {
+        if (($str === null)) {
```
paulbalandan, line 336: Why the need for additional parentheses?

kenjis: Oh, just a mistake.

kenjis: Removed.

paulbalandan (approved):

PR #5808, opened 2022-03-18
docs: clarify that current_url() returns URL based on Config\App

### LINKED ISSUE #5807: Bug: current_url() returning wrong value when different host
### PHP Version

8.0

### CodeIgniter4 Version

4.1.9

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

The function `current_url()` returns the `$baseURL` in `app/Config/App.php` instead of the real current url in the browser.
This leads to confusion, know the documentation says:

> You can restrict groups of routes to function only in certain domain or sub-domains of your application by passing the “hostname” option along with the desired domain to allow it on as part of the options array:
> 
> https://codeigniter.com/user_guide/incoming/routing.html?highlight=router#limit-to-hostname

Which means an app could have multiple domain pointing at it, but `current_url()` would only return the one set in config.

Use case: CMS, multi-tenant app, multi-vendor app, etc.

### Steps to Reproduce

Call `current_url()` when loading the app with a host like **sub.example.com** while in the config, you would set `$baseURL` to **www.example.com** for example. 

### Expected Output

`current_url()` is the name implies should retour the **current url** and not the **current url with $baseURL as host**.

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5807

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -109,7 +109,7 @@ if (! function_exists('base_url')) {
 
 if (! function_exists('current_url')) {
     /**
-     * Returns the current full URL based on the IncomingRequest.
+     * Returns the current full URL based on the Config\App settings and IncomingRequest.
      * String returns ignore query and fragment parts.
      *
      * @param bool                 $returnObject True to return an object instead of a string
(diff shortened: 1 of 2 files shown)


MGatner (approved):
I can't wait to overhaul URIs in version 5. What a mess 😓

Good clarification.

Issue #2787, opened 2020-04-02
Bug: Routing placeholder in "controller"part of route doesn't work


pbquet:
Hello,
This route work
```php
$routes->add('{locale}/admin/parameter/base/config/(:num)', 'Admin\Base::config/$1');
```
but in the following one, the first placeholder doesn't work because it is in the "controller" part of the route instead of "method" part :
```php
$routes->add('{locale}/admin/parameter/(:segment)/config/(:num)', 'Admin\$1::config/$2');
```
and that's because the '\\' before $1 escape the '$'

It's possible to resolve this issue in "system/Router/Router.php::checkRoutes" line 470 by changing :

```php
$val = preg_replace('#^' . $key . '$#', $val, $uri);
```
to
```php
$val = preg_replace('#^' . $key . '$#',  str_replace('\\', '\\\\', $val), $uri);
```
What do you think ?

**CodeIgniter 4 version**
4.02

**Affected module(s)**
Router/Router.php

**Context**
 - OS: Windows 10
 - Web server nginx 1.14
 - PHP 7.3.15

lonnieezell:
We've had this brought up before but I have to turn down that feature. I view that as too big of a security risk because that allows anyone to try to load up common controller/method combos during an attempted attack. Not a good feature.

PR #5814, opened 2022-03-20
fix: bug that allows dynamic controllers to be used

kenjis:
**Description**
Dynamic controllers are not allowed for security reasons. See https://github.com/codeigniter4/CodeIgniter4/issues/2787#issuecomment-608198225
But there is a bug that allows dynamic controllers.

- add check for dynamic controllers
- refactor

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Language/en/Router.php
+++ b/system/Language/en/Router.php
@@ -11,6 +11,7 @@
 
 // Router language settings
 return [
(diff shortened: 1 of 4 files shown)


MGatner (approved):
Thanks for splitting out the commits, made it much easier to find the actual fix. Looks good!

Issue #5810, opened 2022-03-18
Bug: Subquery in query builder


SupremeSimon:
### PHP Version

8.0

### CodeIgniter4 Version

4.1.9

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

apache

### Database

MySQL 8.0.27

### What happened?

The CI query breaks with a written subquery that contains a WHERE AND in it.

I have a query that needs to build from a base query and data from a datatable.
When the users search on the datatable, the search value gets passed into a function that adds the search value into a orLike query builder function.

In debugging I've just written this query manually and the same issue occurs.

I've only noticed it break when you add an AND to it.  A single WHERE it doesn't have a problem with.

I have also found out that if you switch off the escape function, the query generates correctly.  (Although in test this morning with the most basic query example I'm going to use in the reproduce step, it's actually not fixing it)

### Steps to Reproduce

The full query I'm using is:

            $builder->select("`c`.`contact_id`, `c`.`file_as`,");
            $builder->select("(SELECT COALESCE(GROUP_CONCAT(CONCAT(corr.description, ' ', corr.note) ORDER BY corr.order_field ASC SEPARATOR ' '), '') FROM correspondence corr WHERE corr.contact_fk = c.contact_id AND 1 = 1)");
            $builder->join("correspondence corr", "corr.contact_fk = c.contact_id", 'LEFT');
            $builder->join("contact_person cp", "c.contact_id = cp.contact_id", 'LEFT');
            $builder->join("address_contact_link acl", "c.contact_id = acl.contact_fk", 'LEFT');
            $builder->join("address ad", "ad.address_id = acl.address_fk", 'LEFT');
            $builder->join("contact_type_link ctl", "c.contact_id = ctl.contact_fk", 'LEFT');
            $builder->where('c.agent_fk', session("agent_id"));
            $builder->where('c.contact_class', 'person');
            $builder->orLike("(SELECT COALESCE(GROUP_CONCAT(CONCAT(corr.description, ' ', corr.note) ORDER BY corr.order_field ASC SEPARATOR ' '), '') FROM correspondence corr WHERE corr.contact_fk = c.contact_id AND 1 = 1)", "", "both", false);
            $builder->groupBy("c.contact_id");

            echo $builder->getCompiledSelect();

Without the and in the orLike function or with the escape set to false I get a working query:

[...]

Issue #3970, opened 2020-12-07
Bug: [QueryBuilder] Like function of Database library fails when used with Concat


mmfarhan:
Hi,

it looks like there is some issue with the like function in db library:
`("concat(users.name, ' ', IF(users.surname IS NULL or users.surname = '', '', users.surname)) full_name")` works in the select statement but fails in the like function.
It produces this:
`OR  concat(users.name, ' ', IF(users.surname IS NULL or `users`.`surname` = '', '', users.surname))  LIKE :concat(users.name, ' ', IF(users.surname IS NULL or `users`.`surname` = '', '', users.surname)): ESCAPE '!'`

This simple code will result into error:
```php
function sqlTest(){
		$search['value'] = 'perfect';
		$usersModel = new \App\Models\Users_model();
		$usersModel
				->select("concat(users.name, ' ', IF(users.surname IS NULL or users.surname = '', '', users.surname)) full_name")
				->groupStart()
				->orLike("users.email", $search['value'], 'both')
				->orLike("users.phone", $search['value'], 'both')
				->orLike("concat(users.name, ' ', IF(users.surname IS NULL or users.surname = '', '', users.surname))", $search['value'], 'both')
				->groupEnd()
				->findAll();
		
		echo $usersModel->getLastQuery();
	}
```

najdanovicivan:
Try using havingLike() for concatenate value

kenjis:
https://codeigniter4.github.io/CodeIgniter4/database/query_builder.html#query-builder-like-rawsql
Fixed by #5817

PR #5817, opened 2022-03-21
feat: QueryBuilder raw SQL string support

kenjis:
Related: #4355, #3970, #5810

**Description**
See https://forum.codeigniter.com/showthread.php?tid=80720

- add `RawSql` class
- `select()` supports `RawSql`
- `where()` supports `RawSql`
- `like()` supports `RawSql`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1083,11 +1083,7 @@ class BaseBuilder
             return $this;
         }
 
-        if (! is_array($field)) {
-            $keyValue = [$field => $match];
-        } else {
-            $keyValue = $field;
-        }
+        $keyValue = ! is_array($field) ? [$field => $match] : $field;
 
         foreach ($keyValue as $k => $v) {
             if ($insensitiveSearch === true) {
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -371,6 +371,12 @@ class BaseBuilder
             $escape = $this->db->protectIdentifiers;
         }
 
+        if ($select instanceof RawSql) {
+            $this->QBSelect[] = $select;
+
+            return $this;
+        }
+
         foreach ($select as $val) {
             $val = trim($val);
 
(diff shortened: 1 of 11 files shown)


iRedds:
Is it possible to add a second argument to the constructor to pass the binding?
As for BaseConnection::query();
```php 
new RawSql('a = ? AND b = ?', [1, 2]);
new RawSql('a = :a AND b = :b', ['a'=>1, 'b' => 2]);
```

kenjis:
@iRedds I'm not sure.

Can you tell me what's good if it is possible?
Use case?

iRedds:
Usage example above.
It seems to me that it will be convenient with a large SQL statement and a large number of variables.

For example, for such a query (from https://github.com/codeigniter4/CodeIgniter4/issues/5810 )
```php
$builder->orLike("(SELECT COALESCE(GROUP_CONCAT(CONCAT(corr.description, ' ', corr.note) ORDER BY corr.order_field ASC SEPARATOR ' '), '') FROM correspondence corr WHERE corr.contact_fk = c. contact_id AND 1 = 1)", "", "both", false);
```
Although this query does not use value substitution, this is a possible situation.

kenjis:
@iRedds What's about escaping?
If we need to escape the values, RawSql must have DB Connection.

iRedds:
It seems to me that there are two options without using DBC
1. The developer himself will escape the values.
2. The binds will be transferred to the $binds property when processing the RawSql instance.

kenjis:
I do not see much need for binding, so I will not include it in this PR.

lonnieezell (changes requested):
I think this solution is fine, and I don't see an issue with leaving binds out. Being raw SQL I don't think that's unexpected.   Thanks @kenjis 

But it does need docs added.

kenjis:
I added docs.

michalsn (approved):
I thought that the `$escape` parameter would do the trick for this type of scenario but I guess I was wrong. I think it is a good addition.

kenjis:
@michalsn Thank you!

PR #5824, opened 2022-03-23
config: remove App\ and Config\ in autoload.psr-4 in app starter composer.json

### LINKED ISSUE #5818: Bug: defining new namespace under `app/` causes `Cannot declare class ..., because the name is already in use`
### PHP Version

8.0

### CodeIgniter4 Version

v4.1.9

Windows

### Which server did you use?

apache

### Database

--

### What happened?

Problem in command `php spark make:command --namespace MyNameSpace`

### Steps to Reproduce

> `composer create-project codeigniter4/appstarter project-root`
```
    public $psr4 = [
        'MyNameSpace' 	        => APPPATH . 'ThirdParty\MyFolder',
    ];
```
`php spark make:command --namespace MyNameSpace`

Command class name : TeseCommand 
_File created: APPPATH\ThirdParty\MyFolder\Commands\TeseCommand.php_

Everything is done well.
The corresponding file is created.
But now typing command `php spark` displays the following error!
```
PHP Fatal error:  Cannot declare class MyNameSpace\Commands\TeseCommand, because the name is already in use in P:\Program\xampp8010\htdocs\cimc\app\ThirdParty\MyFolder\Commands\TeseCommand.php on line 8

[ErrorException]

Cannot declare class MyNameSpace\Commands\TeseCommand, because the name is already in use

at APPPATH\ThirdParty\MyFolder\Commands\TeseCommand.php:8
```

### Expected Output

@paulbalandan , @kenjis  need to check to fix the problem. Please check.

### Anything else?

This problem only exists in the installation method through **composer** .
This problem does **not** exist in the **manual installation** method.

kenjis:
**Description**
Revert #3423
Fixes #5818

Why
- The `autoload.psr-4` causes problems:
  - Cannot change `app` folder name. Because the composer's path overwrite the config in `Config\Autoload`. See https://forum.codeigniter.com/thread-81153.html
  - Causes error when defining new namespace under `app/`. See https://github.com/codeigniter4/CodeIgniter4/issues/5818
  - These are very difficult to debug.
- Only app starter has the configs. So it is not consistent.

If there is a user who wants to use Composer autoloader optimization (`dump-autoload -o`) for `App` namespace class files, they could add it in `autoload.psr-4` by themself.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/admin/starter/composer.json
+++ b/admin/starter/composer.json
@@ -17,10 +17,6 @@
         "ext-fileinfo": "Improves mime type detection for files"
     },
     "autoload": {
-        "psr-4": {
-            "App\\": "app",
-            "Config\\": "app/Config"
-        },
         "exclude-from-classmap": [
             "**/Database/Migrations/**"
         ]
--- a/system/CLI/Commands.php
+++ b/system/CLI/Commands.php
@@ -76,6 +76,10 @@ class Commands
     /**
      * Discovers all commands in the framework and within user code,
      * and collects instances of them to work with.
+     *
+     * @TODO this approach (find qualified classname from path) causes error,
+     *      when using Composer autoloader.
+     *      See https://github.com/codeigniter4/CodeIgniter4/issues/5818
      */
     public function discoverCommands()
     {


samsonasik (approved):

kenjis:
@samsonasik Thank you!

PR #5834, opened 2022-03-26
Unset Composer's autoloader from the autoload stack

paulbalandan:
**Description**
Fixes https://github.com/codeigniter4/CodeIgniter4/issues/5818#issuecomment-1077407039

Since we only need Composer to register its PSR4 namespaces and classmaps into our internal autoloader, it should not be in the `spl_autoload` stack after as this messes with `class_exists` checks. When our autoloader fails to find the file, PHP passes the search to the next autoloader, which is Composer. Our internal autoloader uses `include_once` while Composer uses `include` causing the redeclaration error.

Cannot write a unit test for this. To check manually using this repo:
1. Add `"App\\": "app"` in the `autoload.psr4` section in the root composer.json
2. Run `composer dump -o` to regenerate the class map.
3. Add `'MyNameSpace' => APPPATH . 'ThirdParty/MyFolder'` in `Config\Autoload::$psr4` array.
4. Run `php spark make:command TeseCommand --namespace MyNameSpace`
5. Run `php spark`. This will now error.
6. Apply this fix ef7dec6efa9d5ce06d83ab5f4c5f211c64dbbc75
7. Run `php spark`. No errors should appear.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -312,6 +312,7 @@ class Autoloader
         $paths    = $composer->getPrefixesPsr4();
         $classes  = $composer->getClassMap();
 
+        $composer->unregister();
         unset($composer);
 
         // Get rid of CodeIgniter so we don't have duplicates
--- a/system/CLI/Commands.php
+++ b/system/CLI/Commands.php
@@ -102,7 +102,7 @@ class Commands
         foreach ($files as $file) {
             $className = $locator->findQualifiedNameFromPath($file);
 
-            if (empty($className) || ! class_exists($className)) {
+            if ($className === false) {
                 continue;
             }
 
--- a/system/Test/bootstrap.php
+++ b/system/Test/bootstrap.php
@@ -84,11 +84,6 @@ if (! class_exists('CodeIgniter\Services', false)) {
 // Initialize and register the loader with the SPL autoloader stack.
 Services::autoloader()->initialize(new Autoload(), new Modules())->register();
 
-// Now load Composer's if it's available
-if (is_file(COMPOSER_PATH)) {
-    require_once COMPOSER_PATH;
-}
-
 // Load environment settings from .env files into $_SERVER and $_ENV
 require_once SYSTEMPATH . 'Config/DotEnv.php';
 
--- a/system/bootstrap.php
+++ b/system/bootstrap.php
@@ -109,18 +109,13 @@ if (! class_exists('CodeIgniter\Services', false)) {
 // Initialize and register the loader with the SPL autoloader stack.
 Services::autoloader()->initialize(new Autoload(), new Modules())->register();
 
-// Now load Composer's if it's available
-if (is_file(COMPOSER_PATH)) {
-    /*
-     * The path to the vendor directory.
-     *
-     * We do not want to enforce this, so set the constant if Composer was used.
-     */
-    if (! defined('VENDORPATH')) {
-        define('VENDORPATH', dirname(COMPOSER_PATH) . DIRECTORY_SEPARATOR);
-    }
-
-    require_once COMPOSER_PATH;
+/*
+ * The path to the vendor directory.
+ *
+ * We do not want to enforce this, so set the constant if Composer was used.
+ */
+if (is_file(COMPOSER_PATH) && ! defined('VENDORPATH')) {
+    define('VENDORPATH', dirname(COMPOSER_PATH) . DIRECTORY_SEPARATOR);
 }
 
 // Load environment settings from .env files into $_SERVER and $_ENV


samsonasik (approved):

system/Autoloader/Autoloader.php
```diff
@@ -312,6 +312,7 @@ protected function discoverComposerNamespaces()
         $paths    = $composer->getPrefixesPsr4();
         $classes  = $composer->getClassMap();
 
+        $composer->unregister();
```
kenjis, line 315: This is in the `discoverComposerNamespaces()` method.
So when `$modules->discoverInComposer` is false, it does not run.
When `$modules->discoverInComposer` is false, Composer autoloader is still enabled.
Is it intentional?

kenjis:
> Since we only need Composer to register its PSR4 namespaces and classmaps into our internal autoloader,

When a user does not have Composer classmap (don't use `-o`) and sets `discoverInComposer` false,
we need Composer autoloader to find classes in Composer packages.

So this PR is okay unless there is no bugs in the CI4 autoloader of PSR-4 class loading?

kenjis:
@paulbalandan This is not directly related this PR, but I think the following logic is the cause of the error,
and should be changed.

`$locator->listFiles()` uses psr-4 namespaces to find files, but returns file paths only, no namespaces.
And from the file path list, `findQualifiedNameFromPath()` guesses full classname. So the namespace of the class might be incorrect.

https://github.com/codeigniter4/CodeIgniter4/blob/7107e782f86b3394c22cf980c85bc385b42b9207/system/CLI/Commands.php#L90-L103

MGatner:
Looks like good work and review underway already, let me know if you need me.

system/Autoloader/Autoloader.php
paulbalandan: Composer's autoloader is only enabled when `include`d in `$this->discoverComposerNamespaces()`. In the code, once we include `COMPOSER_PATH`, Composer will register its autoloader and prepend it in the autoload stack. If `$modules->discoverInComposer` is false, then `COMPOSER_PATH` is not included and won't be registered.

kenjis:
```php
<?php

namespace App\Controllers;

use Closure;

{
    public function index()
    {
        foreach (spl_autoload_functions() as $autoloader) {
            switch ($autoloader) {
                case $autoloader instanceof Closure:
                    echo get_class($autoloader) . PHP_EOL;
                    break;

                case is_string($autoloader[0]):
                    echo $autoloader[0] . '::' . $autoloader[1] . PHP_EOL;
                    break;

                case is_object($autoloader[0]):
                    echo get_class($autoloader[0]) . '::' . $autoloader[1] . PHP_EOL;
                    break;
            }
        }
    }
}
```

With this PR, there is no Composer autoloader.
```
$ php public/index.php 
CodeIgniter\Autoloader\Autoloader::loadClassmap
CodeIgniter\Autoloader\Autoloader::loadClass
PHPStan\PharAutoloader::loadClass
Closure
```

Next, make `$discoverInComposer` false.

```diff
--- a/app/Config/Modules.php
+++ b/app/Config/Modules.php
@@ -29,7 +29,7 @@ class Modules extends BaseModules
      *
      * @var bool
      */
-    public $discoverInComposer = true;
+    public $discoverInComposer = false;
 
     /**
      * --------------------------------------------------------------------------
```

```
$ php public/index.php 
Composer\Autoload\ClassLoader::loadClass
CodeIgniter\Autoloader\Autoloader::loadClassmap
CodeIgniter\Autoloader\Autoloader::loadClass
PHPStan\PharAutoloader::loadClass
Closure
```
Composer autoloader is registered.

paulbalandan: Okay. So there's a place in the application lifecycle where Composer is included regardless of the `$discoverInComposer` flag. I'll take a look into it but most probably it shouldn't be included automatically.

paulbalandan: Found it!

In `system/bootstrap.php` and `system/Test/bootstrap.php`, we `require_once COMPOSER_PATH` regardless of the `$discoverInComposer` flag, which shouldn't be the case as it is handled already in the `Services::autoloader()->initialize()` call.

paulbalandan:
```php
if (is_file(COMPOSER_PATH)) {
    require_once COMPOSER_PATH;
}
```

paulbalandan:
> `$locator->listFiles()` uses psr-4 namespaces to find files, but returns file paths only, no namespaces. And from the file path list, `findQualifiedNameFromPath()` guesses full classname. So the namespace of the class might be incorrect.
> 

`findQualifiedNameFromPath()` uses `class_exists()` check on the guessed class name to see if it is a real class.

kenjis:
@paulbalandan If you disable `$discoverInComposer`, you can't autoload Composer package files.
Is this what you want?

[Auto-Discovery](https://codeigniter4.github.io/CodeIgniter4/general/modules.html#id3) is the functionality to find automatically different file types like Routes.php. [$discoverInComposer](https://codeigniter4.github.io/CodeIgniter4/general/modules.html#discovery-and-composer) means whether to find them in Composer PSR-4 namespaces.
I don't think it disables Composer autoloader.

```diff
--- a/app/Config/Modules.php
+++ b/app/Config/Modules.php
@@ -29,7 +29,7 @@ class Modules extends BaseModules
      *
      * @var bool
      */
-    public $discoverInComposer = true;
+    public $discoverInComposer = false;
 
     /**
      * --------------------------------------------------------------------------
```

```diff
--- a/app/Controllers/Home.php
+++ b/app/Controllers/Home.php
@@ -6,6 +6,6 @@ class Home extends BaseController
 {
     public function index()
     {
-        return view('welcome_message');
+        $faker = \Faker\Factory::create();
     }
 }
```
```
$ php public/index.php 


[Error]

Class "Faker\Factory" not found

at APPPATH/Controllers/Home.php:9

Backtrace:
  1    SYSTEMPATH/CodeIgniter.php:894
       App\Controllers\Home()->index()

  2    SYSTEMPATH/CodeIgniter.php:467
       CodeIgniter\CodeIgniter()->runController(Object(App\Controllers\Home))

  3    SYSTEMPATH/CodeIgniter.php:350
       CodeIgniter\CodeIgniter()->handleRequest(null, Object(Config\Cache), false)

  4    FCPATH/index.php:40
       CodeIgniter\CodeIgniter()->run()
```

kenjis:
@paulbalandan Okay, you think the error in https://github.com/codeigniter4/CodeIgniter4/pull/5834#issuecomment-1081198386 is not a bug, and no problem.

And I don't agree with it.

@MGatner Can you give me your opinion?

MGatner:
I think I'm not understanding something in these threads. Tell me what I have wrong here:
1. Composer's Autoloader should be registered once, always (if it is available) regardless of Config\Modules
2. Class discovery should work across all namespaces using our Autoloader
3. Module discovery should only control whether a component "goes looking" for more of its kind, e.g. if Events will check MyNamespace\Config\Events for a trigger, *not* whether I can access that class directly via autoloading

MGatner:
Maybe Events was a bad example because they aren't actually classes. But take Services - I should be able to do this regardless of settings in Config\Modules:
```
\MyNamespace\Config\Services::foo()
```
But if discovery is disabled in Config\Modules then this should fail:
```
Config\Services::foo()
```

MGatner:
Let me also add: from Paul's original description that it seems certain there are some things currently wrong with the implementation.

kenjis:
I agree with 1 and 3.

> 1. Composer's Autoloader should be registered once, always (if it is available) regardless of Config\Modules
> 2. Class discovery should work across all namespaces using our Autoloader
> 3. Module discovery should only control whether a component "goes looking" for more of its kind, e.g. if Events will check MyNamespace\Config\Events for a trigger, *not* whether I can access that class directly via autoloading. 

About 2. 
Class discovery should work across all namespaces using our Autoloader or Composer autoloader.
Basically Composer autoloader loads Composer package files, our Autoloader loads `Autoload::$psr4` namespaces.

kenjis:
I agree with 1 and 3.

> 1. Composer's Autoloader should be registered once, always (if it is available) regardless of Config\Modules
> 2. Class discovery should work across all namespaces using our Autoloader
> 3. Module discovery should only control whether a component "goes looking" for more of its kind, e.g. if Events will check MyNamespace\Config\Events for a trigger, *not* whether I can access that class directly via autoloading. 

About 2. 
Class discovery should work across all namespaces using our Autoloader or Composer autoloader.
Basically Composer autoloader loads Composer package files, our Autoloader loads `Autoload::$psr4` namespaces.

kenjis:
I sent another PR #5849.

kenjis:
I have sent a PR to fix what appears to be a bug in Autoloader: #5850

MGatner:
Discussion continued on the bugfix PR. @paulbalandan Can you weigh in there? I think deciding on that (whether it is indeed a bug) will clarify the rest of these.

kenjis:
#5850 was merged. So close this.

PR #5841, opened 2022-03-30
feat: add Config\View::$enableConditionals to disable Parser conditionals

### LINKED ISSUE #5831: Bug: View parser throws view exception with nested 'if' within javascript code
### PHP Version

8.0

### CodeIgniter4 Version

4.1.8

Windows

### Which server did you use?

apache

### Database

mysql

### What happened?

when parsing a view with inline javascript (within a script tag, specifically an if statement within outer braces {})
the view parser mis-interpretes this code as a view conditional and improperly transforms the code to a php conditional.

### Steps to Reproduce

controller code:
```php
public function test(){
		$parser = service('parser');

        $somedata = [
            'message' => 'massagee'
        ];
        $view = <<<'EOL'
                 <script type="text/javascript">
                     var f = function(){
                         if(true){
                             alert('{message}');
                         }
                     }
                 </script>
                 EOL;

        return $parser->setData($somedata)->renderString($view);
    }
```

error message:
```
You have a syntax error in your Parser tags: <script type="text/javascript">
var f = function()if ((true){
alert('{message): ');
}
}
</script>
```

problem:
the nested javascript if statement has been parsed
and php tags have been placed around it 
(observed by printing the output directly pre-evaluation in the parser class)

### Expected Output

Expected output would be the non-modified javascript,
with the substituted tag within.

```js
<script type="text/javascript">
                     var f = function(){
                         if(true){
                             alert('massagee');
                         }
                     }
                 </script>
```

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5831

CI3 Parser does not have conditionals. See https://codeigniter.com/userguide3/libraries/parser.html
CI4 Parser has conditionals, and it is difficult to distinguish if there is JavaScript code in templates.

- add `Config\View::$enalbeConditionals` to disable Parser conditionals
  - The setting is for compatibility with templates that works on CodeIgniter3.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/View.php
+++ b/app/Config/View.php
@@ -44,6 +44,16 @@ class View extends BaseView
     public $plugins = [];
 
     /**
+     * Whether to use Parser conditionals (if, else, and elseif).
+     *
+     * This setting is for compatibility with templates that works on CodeIgniter3.
+     * If you have JavaScript code in templates, Parser may raise error
+     * when there are strings that can be interpreted as conditionals.
+     * In that case, set this false.
+     */
+    public bool $enableConditionals = true;
+
+    /**
      * View Decorators are class methods that will be run in sequence to
      * have a chance to alter the generated output just prior to caching
      * the results.
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -60,6 +60,16 @@ class Parser extends View
     protected $dataContexts = [];
 
     /**
+     * Whether to use Parser conditionals (if, else, and elseif).
+     *
+     * This setting is for compatibility with templates that works on CodeIgniter3.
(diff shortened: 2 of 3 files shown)


totoprayogo1916 (commented):
Is this a typo?
enable or enalbe?

kenjis:
@totoprayogo1916 Thanks! These are typos. Fixed.

kenjis:
I found another solution: #5842
This PR might not to be needed.

kenjis:
#5842 was merged, so I think this is not needed.

PR #5843, opened 2022-03-31
docs: improve tutorial

kenjis:
**Description**
- use `return` in a controller method, instead of `echo`
  - `echo` has side effects, so `return` is better.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


MGatner (approved):
I strongly favor `return` as well. I would never string views together like this without a layout, but I guess in the interest of keeping the tutorial simpler that makes sense?

paulbalandan (approved):

PR #5844, opened 2022-03-31
fix: Validation returns incorrect errors after Redirect with Input

### LINKED ISSUE #5839: Bug: Validation returns incorrect errors after Redirect with Input
### PHP Version

7.4

### CodeIgniter4 Version

4.1.9

Linux

### Which server did you use?

apache

### Database

MySQL

### What happened?

Validation fails incorrectly following a redirect that uses `_ci_validation_errors`.

### Steps to Reproduce

1. Submit a form to a Controller endpoint that runs validation
2. Have the validation fail and `return redirect()->...->withInput()`
3. On the destination page run a passing validation and see it fail

This is due to `getErrors()` filling in the flashdata errors from `_ci_validation_errors` because `$this->errors` is empty:
https://github.com/codeigniter4/CodeIgniter4/blob/94cc0fecde66a8b5ec0fb1f5e65944b36dafa7eb/system/Validation/Validation.php#L672-L677 

Probably need to differentiate `null` but I can't dig into the implications of this right now.

### Expected Output

The new Validation should pass regardless of the redirected errors.

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5839
- remove session _ci_validation_errors before running validation

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -108,6 +108,12 @@ class Validation implements ValidationInterface
      */
     public function run(?array $data = null, ?string $group = null, ?string $dbGroup = null): bool
     {
+        // If there are still validation errors for redirect_with_input request, remove them.
+        // See `getErrors()` method.
+        if (isset($_SESSION, $_SESSION['_ci_validation_errors'])) {
+            unset($_SESSION['_ci_validation_errors']);
+        }
+
         $data ??= $this->data;
 
         // i.e. is_unique


system/Validation/Validation.php
```diff
@@ -108,6 +108,12 @@ public function __construct($config, RendererInterface $view)
      */
     public function run(?array $data = null, ?string $group = null, ?string $dbGroup = null): bool
     {
+        // If there are still validation errors for redirect_with_input request, remove them.
+        // See `getErrors()` method.
+        if (isset($_SESSION, $_SESSION['_ci_validation_errors'])) {
+            unset($_SESSION['_ci_validation_errors']);
+        }
```
MGatner, line 115: `unset()` does not care about existence, why not just call it?
```suggestion
        unset($_SESSION['_ci_validation_errors']);
```

MGatner (changes requested):

system/Validation/Validation.php
```diff
@@ -108,6 +108,10 @@ public function __construct($config, RendererInterface $view)
      */
     public function run(?array $data = null, ?string $group = null, ?string $dbGroup = null): bool
     {
+        // If there are still validation errors for redirect_with_input request, remove them.
+        // See `getErrors()` method.
+        unset($_SESSION['_ci_validation_errors']);
```
samsonasik, line 113: `isset($_SESSION, $_SESSION['_ci_validation_errors']` check may be needed to avoid race condition.

```suggestion
        if (isset($_SESSION, $_SESSION['_ci_validation_errors'])) {
            unset($_SESSION['_ci_validation_errors']);
        }
```

kenjis: @samsonasik Can you explain the race condition and why the `if` avoid it?

samsonasik: That's just in case, dealing with session is sometime like dealing with cache, which possibly already removed on concurrent request.

kenjis: Even if it was already removed by a concurrent request, there is no error when unsetting a non-existing variable.

samsonasik: it seems show the error if $_SESSION not exists ( probably happen with unset($_SESSION)), see

- https://3v4l.org/p5Okf ( $_SESSION never defined)
- https://3v4l.org/E9mWR ($_SESSION defined but unset next for whatever reason)

kenjis: I've confirmed the code in a controller. Yes, you're correct. `unset($_SESSION['_ci_validation_errors']);` shows Undefined variable $_SESSION.

system/Validation/Validation.php
kenjis: @MGatner No, `unset()` care about existence.
See https://github.com/codeigniter4/CodeIgniter4/pull/5844#discussion_r840335862

samsonasik (approved):

system/Validation/Validation.php
MGatner: Ah, thank you - sorry I missed that originally. Mobile doesn't always show inline comments for resolved threads.

MGatner (approved):

samsonasik:
Thank you @kenjis

PR #5849, opened 2022-04-02
fix: Commands::discoverCommands() loads incorrect classname

kenjis:
**Description**
Fixes #5818
Supersedes #5834

- fix: FileLocator::listFiles() returns directories
- fix: discoverCommands() loads incorrect classname

Cannot write a unit test for `Cannot declare class ..., because the name is already in use`. 
To check manually using this repo:
1. Add `"App\\": "app/"` in the `autoload.psr4` section in the root composer.json
2. Run `composer dump -o` to regenerate the class map.
3. Add `'MyNameSpace' => APPPATH . 'ThirdParty/MyFolder'` in `Config\Autoload::$psr4` array.
4. Run `php spark make:command TeseCommand --namespace MyNameSpace`
5. Run `php spark`. This will now error.
6. Check out this PR branch (`fix-discoverCommands`).
7. Run `php spark`. No errors should appear.

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #5850, opened 2022-04-02
fix: [Autoloader] Composer classmap usage

kenjis:
**Description**
The current implementation uses Composer classmap  only when `$modules->discoverInComposer` is true.
`discoverInComposer` is the setting for Auto-Discovery within Composer Packages.
It has nothing to do with classmap usage.

This PR makes it use always if Composer is available.

Related: #5834

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -111,7 +111,7 @@ class Autoloader
         return $this;
     }
 
-    protected function loadComposerInfo(Modules $modules): void
+    private function loadComposerInfo(Modules $modules): void
     {
         /**
          * @var ClassLoader $composer
@@ -312,7 +312,7 @@ class Autoloader
         return trim($filename, '.-_');
     }
 
-    protected function loadComposerNamespaces(ClassLoader $composer): void
+    private function loadComposerNamespaces(ClassLoader $composer): void
     {
         $paths = $composer->getPrefixesPsr4();
 
@@ -331,7 +331,7 @@ class Autoloader
         $this->prefixes = array_merge($this->prefixes, $newPaths);
     }
 
-    protected function loadComposerClassmap(ClassLoader $composer): void
+    private function loadComposerClassmap(ClassLoader $composer): void
     {
         $classes = $composer->getClassMap();
 
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -337,4 +337,40 @@ class Autoloader
 
         $this->classmap = array_merge($this->classmap, $classes);
     }
+
+    /**
+     * Locates autoload information from Composer, if available.
+     *
+     * @deprecated No longer used.
+     */
+    protected function discoverComposerNamespaces()
+    {
+        if (! is_file(COMPOSER_PATH)) {
+            return;
+        }
+
+        /**
+         * @var ClassLoader $composer
+         */
+        $composer = include COMPOSER_PATH;
+        $paths    = $composer->getPrefixesPsr4();
+        $classes  = $composer->getClassMap();
+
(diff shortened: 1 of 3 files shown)


MGatner (approved):
This is more along the lines of what I expected of the interaction between Autoloader, Composer, and Modules. I'm approving but let's be sure to hear @paulbalandan on this bc I am not entirely sure my understanding was correct. I also have never turned off Composer discovery so I'm not sure what people are expecting from that. Maybe @lonnieezell will have some insight from earlier development of these features?

system/Autoloader/Autoloader.php
```diff
@@ -296,23 +312,9 @@ public function sanitizeFilename(string $filename): string
         return trim($filename, '.-_');
     }
 
-    /**
-     * Locates autoload information from Composer, if available.
-     */
-    protected function discoverComposerNamespaces()
+    protected function loadComposerNamespaces(ClassLoader $composer): void
```
mostafakhudair, line 315:
```suggestion
    private function loadComposerNamespaces(ClassLoader $composer): void
```

system/Autoloader/Autoloader.php
```diff
@@ -327,6 +329,12 @@ protected function discoverComposerNamespaces()
         }
 
         $this->prefixes = array_merge($this->prefixes, $newPaths);
+    }
+
+    protected function loadComposerClassmap(ClassLoader $composer): void
```
mostafakhudair, line 334:
```suggestion
    private function loadComposerClassmap(ClassLoader $composer): void
```

system/Autoloader/Autoloader.php
```diff
@@ -296,23 +312,9 @@ public function sanitizeFilename(string $filename): string
         return trim($filename, '.-_');
     }
 
-    /**
-     * Locates autoload information from Composer, if available.
-     */
-    protected function discoverComposerNamespaces()
```
mostafakhudair, line 302: removing this function is BC

paulbalandan (commented):
I took the time to re-read the thread from my PR up to here and somehow it cleared the misconception I had from reading only the code. I think this approach, coupled with #5849, is better than mine. This PR now syncs the code with the documentation. 👍

Some technical notes:
1. I would like the new methods to be internal only (i.e. private) to reduce future BC concerns. And I believe these methods would never be needed to be overridden by users.
2. Annoying as it can be, let's just deprecate `discoverComposerNamespaces()` instead of removing. We can cleanup in v5.

system/Autoloader/Autoloader.php
kenjis: Restored.

system/Autoloader/Autoloader.php
kenjis: Done.

system/Autoloader/Autoloader.php
kenjis: Done.

kenjis:
@MGatner 
> I also have never turned off Composer discovery so I'm not sure what people are expecting from that.

If we have a lot of Composer packages, and not CodeIgniter packages, Composer discovery needs to look for files in these packages and it found nothing.

In that case, if we turn off Composer discovery and set the CodeIgniter Composer package namespaces in the Autoloader Config file, we can avoid the useless Composer discovery.

I thought that, but I have never tried yet.

kenjis:
@MGatner @mostafakhudair @paulbalandan 
Thank you for the reviews.

I updated all the things, and added changelog.

paulbalandan (approved):

MGatner (approved):
Thanks all! Great collaboration.

PR #5851, opened 2022-04-02
wrong function name `or_where()`

tcgumus:
Needs to be orWhere(). There is no or_where in CI 4.

**Description**
given function name was wrong and cost me some time. I wanted to change it to the correct one so othere won't lose time.


kenjis (approved):
Thank you!

PR #5861, opened 2022-04-05
fix: validation errors in Model are not cleared when running validation again

### LINKED ISSUE #5859: Bug: Validation Errors not cleared after database insert with Model
### PHP Version

8.1

### CodeIgniter4 Version

4.1.9

Linux

### Which server did you use?

apache

### Database

PostgreSQL 12.9

### What happened?

I tried to insert values into the default database one by one with a simple validation (`'required|alpha'`). However, after the first value where validation failed, all other following inserts, that should have passed validation without errors failed as well. I looked at the validation errors and the error messages did not make much sense to me. Then I wrote some custom validation errror messages that also showed me the value. I discovered that the error of an earlier entry was shown when the validation for the valid current entry failed.

### Steps to Reproduce

Short: Insert multiple entries with validation via model->insert(), where the validation fails for at least one entry that is not the last one. 

Long: 

1. Create test table in the PostgreSQL database:

` CREATE table test( id SERIAL PRIMARY KEY, name VARCHAR);`

2. Setup test model in CodeIgniter4: see attachement TestModel.php.txt

3. Write test controller for inserting data: see attachment TestController.php.txt

4. Run App and open servername/testcontroller/test_insert
Output: see attachment bug_output.txt

[TestModel.php.txt](https://github.com/codeigniter4/CodeIgniter4/files/8409847/TestModel.php.txt)
[Testcontroller.php.txt](https://github.com/codeigniter4/CodeIgniter4/files/8409879/Testcontroller.php.txt)
[bug_output.txt](https://github.com/codeigniter4/CodeIgniter4/files/8409831/bug_output.txt)

```php
<?php

namespace App\Controllers;

use CodeIgniter\CodeIgniter;
use Shuchkin\SimpleXLSX;


class Testcontroller extends BaseController
{
    public function test_insert()
    {
        $testModel = model('\App\Models\TestModel2', false);

        $test_names = ['Adam', 'Eve', null, 'Maria', '123', 'Marc', 'Aron' ];

        foreach($test_names as $name)
        {
            $test_data = array('name' => $name);
            $testModel->insert($test_data);
            echo $name .' ';
            echo print_r($testModel->validation->getErrors(), true);
            echo '<br>';
        }  
    }
}
```

```php
<?php

namespace app\Models;

use CodeIgniter\Model;

class TestModel2 extends Model
{
    protected $table = 'test2';
    protected $primaryKey = 'id';

    protected $useAutoIncrement = true;
    protected $returnType = 'array';
    protected $skipValidation = false;

    protected $allowedFields = ['name'];

    protected $validationRules = [
        'name'    => 'required|alpha'
    ];

    protected $validationMessages = [
        'name' => ['required' => 'Is required ({value})',
                   'alpha' => 'Must only contain alpha chars ({value})'                
                  ],
    ];

}
```

```
Adam Array ( )
Eve Array ( )
Array ( [name] => Is required () )
Maria Array ( [name] => Is required () )
123 Array ( [name] => Must only contain alpha chars (123) )
Marc Array ( [name] => Must only contain alpha chars (123) )
Aron Array ( [name] => Must only contain alpha chars (123) )
```

### Expected Output

I expect that the validation fails only for the inserts that do not fulfill the validation rules.

### Anything else?

I looked into the framework and could 'quickfix' the problem for me by resetting the error array at the very beginning of the `Validation->run()` method in codeigniter4/system/Validation/Validation.php (line 110). I do not know if this is an intelligent or a very dumb solution. I could not find another place where the error array is cleared.

Please be gentle with me. This is my first github issue. Please let me know if I did something wrong or what I can do to improve my next one.

kenjis:
**Description**
Fixes #5859

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1344,7 +1344,9 @@ abstract class BaseModel
             return true;
         }
 
-        return $this->validation->setRules($rules, $this->validationMessages)->run($data, null, $this->DBGroup);
+        $this->validation->reset()->setRules($rules, $this->validationMessages);
+
+        return $this->validation->run($data, null, $this->DBGroup);
     }
 
     /**
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -26,11 +26,6 @@ parameters:
 			path: system/Autoloader/Autoloader.php
 
 		-
-			message: "#^Method CodeIgniter\\\\Validation\\\\ValidationInterface\\:\\:run\\(\\) invoked with 3 parameters, 0\\-2 required\\.$#"
-			count: 1
-			path: system/BaseModel.php
-
-		-
 			message: "#^Property Config\\\\Cache\\:\\:\\$backupHandler \\(string\\) in isset\\(\\) is not nullable\\.$#"
 			count: 1
 			path: system/Cache/CacheFactory.php
--- a/tests/system/Models/ValidationModelTest.php
+++ b/tests/system/Models/ValidationModelTest.php
@@ -55,6 +55,29 @@ final class ValidationModelTest extends LiveModelTestCase
         $this->assertSame('You forgot to name the baby.', $errors['name']);
     }
 
+    /**
+     * @see https://github.com/codeigniter4/CodeIgniter4/issues/5859
+     */
+    public function testValidationTwice(): void
+    {
+        $data = [
+            'name'        => null,
+            'description' => 'some great marketing stuff',
+        ];
+
+        $this->assertFalse($this->model->insert($data));
+
+        $errors = $this->model->errors();
+        $this->assertSame('You forgot to name the baby.', $errors['name']);
+
+        $data = [
+            'name'        => 'some name',
+            'description' => 'some great marketing stuff',
+        ];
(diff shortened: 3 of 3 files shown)


samsonasik (approved):

system/BaseModel.php
```diff
@@ -1344,7 +1344,9 @@ public function validate($data): bool
             return true;
         }
 
-        return $this->validation->setRules($rules, $this->validationMessages)->run($data, null, $this->DBGroup);
+        $this->validation->reset()->setRules($rules, $this->validationMessages);
+
+        return $this->validation->run($data, null, $this->DBGroup);
```
paulbalandan, line 1349: Aside from adding the `->reset()` method, may I know why the line got broken into two? Or was this because of line length?

paulbalandan (approved):

system/BaseModel.php
kenjis: Yes, it is too long.
And one line is a bit difficult to debug, step debugging.

PR #5867, opened 2022-04-05
chore: update composer scripts

kenjis:
**Description**
Follow-up #5866
- update composer scripts

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,14 +62,14 @@
         "analyze": "phpstan analyse",
         "test": "phpunit",
         "cs": [
-            "php-cs-fixer fix --verbose --dry-run --diff --config=.user-guide.php-cs-fixer.dist.php",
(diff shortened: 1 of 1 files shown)


paulbalandan (commented):
This is more of a request than a review :) Can we add the `--ansi` flag to the composer scripts so that output will be colored rather than bland black and white.

kenjis:
Added `--ansi`.
On macOS the output is colored, so I didn't know the need.

paulbalandan (approved):
Not sure why, but Composer running these scripts in Windows gives black and white output.

sfadschm:
I am on mobile, but I only see one changed file (composer. json). Are the cs-fixer files renamed?

sfadschm:
Nevermind, they already were before 😄

PR #5876, opened 2022-04-07
[PHPStan] Prepare for PHPStan 1.6.x-dev

samsonasik:
PHPStan 1.6.0 will remove `parent`, `next`, `previous` attribute from node on `bleedingEdge` feature toggles, while it only disable when `bleedingEdge` enabled, it may possibly be removed in the future, so I think it is better to prepare early. 

We used the `previous` attribute check on `CheckUseStatementsAfterLicenseRule` for it:

https://github.com/codeigniter4/CodeIgniter4/blob/622d50aa2976b85f7cb1479faec7558cd9ec2341/utils/PHPStan/CheckUseStatementsAfterLicenseRule.php#L60

This PR is for draft and can be updated to use ^1.6 when PHPStan 1.6 released.

References:

- https://github.com/rectorphp/rector/issues/7088
- https://github.com/rectorphp/rector-src/pull/2014

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "mikey179/vfsstream": "^1.6",
         "nexusphp/cs-config": "^3.3",
         "nexusphp/tachycardia": "^1.0",
-        "phpstan/phpstan": "1.6.x-dev",
+        "phpstan/phpstan": "^1.5.6",
         "phpunit/phpunit": "^9.1",
         "predis/predis": "^1.1",
         "rector/rector": "0.12.20"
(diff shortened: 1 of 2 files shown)


MGatner:
Thanks for being ahead of this!

samsonasik:
PHPStan 1.5.6 released and backward compatible already conditionalTags on NodeConnectingVisitor already added.

samsonasik:
PHPStan github action green 🎉 ref https://phpstan.org/blog/preprocessing-ast-for-custom-rules for config reference on PHPStan 1.6.0 that the compat can be configured via conditionalTags

samsonasik:
I am merging it ;). Refs for same example applies:

- https://github.com/symplify/symplify/pull/4029
- https://github.com/rectorphp/rector-src/pull/2014

PR #5880, opened 2022-04-08
docs: add `literalinclude` for curlrequest -> debug

totoprayogo1916:
before:
image

after:
image

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/user_guide_src/source/libraries/curlrequest/034.php
+++ b/user_guide_src/source/libraries/curlrequest/034.php
@@ -1,0 +1,3 @@
+<?php
+
+$response->request('GET', 'http://example.com', ['debug' => true]);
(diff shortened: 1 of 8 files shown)


user_guide_src/source/libraries/curlrequest/034.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+$response->request('GET', 'http://example.com', ['debug' => true]);
```
kenjis, line 3: This is not your mistake, but isn't it `$client->request(`?

kenjis: Also all the other sample code in [Request Options](https://codeigniter4.github.io/CodeIgniter4/libraries/curlrequest.html#request-options).

totoprayogo1916: updated

kenjis (approved):
Thank you!

PR #5882, opened 2022-04-08
doc: fix `$pathsConfig` for spark

totoprayogo1916:
ref:
https://github.com/codeigniter4/CodeIgniter4/blob/957656faf0875e8ecf0771e977c1f2d35a4d8961/spark#L49

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide


samsonasik:
Thank you @totoprayogo1916

PR #5885, opened 2022-04-09
feat: throws exception when controller name in routes contains `/`

kenjis:
**Description**
- Throws Exception when a routes has invalid controller name like `'Admin/AdminGlavni::cam_edit_show/$1'`

It is better users know it should be `\` than replacing `/` with `\` without notice.
Because `Admin/AdminGlavni` is a wrong classname.

See https://forum.codeigniter.com/showthread.php?tid=81681&pid=395315#pid395315

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -426,6 +426,11 @@ class Router implements RouterInterface
                         throw RouterException::forDynamicController($handler);
                     }
 
+                    // Checks `/` in controller name
+                    if (strpos($controller, '/') !== false) {
+                        throw RouterException::forInvalidControllerName($handler);
+                    }
+
                     if (strpos($routeKey, '/') !== false) {
                         $replacekey = str_replace('/(.*)', '', $routeKey);
                         $handler    = preg_replace('#^' . $routeKey . '$#u', $handler, $uri);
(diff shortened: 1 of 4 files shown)


system/Router/Router.php
```diff
@@ -426,6 +426,11 @@ protected function checkRoutes(string $uri): bool
                         throw RouterException::forDynamicController($handler);
                     }
 
+                    // Checks `/` in controller name
+                    if (strpos($controller, '/') !== false) {
```
MGatner, line 430: I wasn't able to find parameters for "valid PHP class names", but would it make sense to expand this out to a regex and fail anything that didn't fit?

MGatner (changes requested):

system/Router/Router.php
kenjis: `Admin/AdminGlavni` is not a valid PHP class name.
Because `/` is wrong.

Many users write like `'Admin/AdminGlavni::cam_edit_show/$1'` and CI4 causes errors.

MGatner: Right, I get that. My question was: is there any reason why not to make this check for all class names? `Admin\Admin-Glavni` is also not valid. This could be implemented as a filter and the error as "Your controller contains the following invalid characters: -"

MGatner (approved):

PR #5893, opened 2022-04-13
fix: using multiple CLI::color() in CLI::write() outputs strings with wrong color

### LINKED ISSUE #5892: Bug: Multiple CLI::color inside CLI::write change color of strings that shouldn't be affected
### PHP Version

7.4, 8.0

### CodeIgniter4 Version

4.1.9

Windows, Linux

### Which server did you use?

apache

### Database

_No response_

### What happened?

When used `CLI::color` at the beginning of the string inside `CLI::write` and then again later inside string, all characters **between the two** instances of `CLI::color` became **affected by the first** instance if `CLI::color`.
If there is only one `CLI::color` inside `CLI::write`, output is just as expected.
Doesn't matter if it is run via CLI inside controller or as a spark command.

### Steps to Reproduce

```php
<?php
namespace App\Controllers;

use CodeIgniter\CLI\CLI;

{
    public function index(): void
    {
        CLI::write('white white white white', 'white');
        CLI::write('white white ' . CLI::color('red', 'red') . ' white', 'white');
        CLI::write('white white white ' . CLI::color('red', 'red'), 'white');
        CLI::write(CLI::color('red', 'red') . ' white white white', 'white');
        CLI::write(CLI::color('red', 'red') . ' white white ' . CLI::color('red', 'red'), 'white');
        CLI::write(CLI::color('red', 'red') . ' white ' . CLI::color('red', 'red') . ' white', 'white');
    }
}

```
Run via CLI.

### Expected Output

All **white** words should be in white color, all **red** words should be in red.

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5892

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -473,6 +473,10 @@ class CLI
             return $text;
         }
 
+        if ($text === '') {
+            return $text;
+        }
+
         if (! array_key_exists($foreground, static::$foreground_colors)) {
             throw CLIException::forInvalidColor('foreground', $foreground);
         }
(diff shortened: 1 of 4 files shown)


system/CLI/CLI.php
```diff
@@ -473,6 +473,10 @@ public static function color(string $text, string $foreground, ?string $backgrou
             return $text;
         }
 
+        if ($text === '') {
+            return $text;
+        }
+
```
paulbalandan, line 479: Can combine this with the first `if`?

system/CLI/CLI.php
```diff
@@ -491,31 +538,7 @@ public static function color(string $text, string $foreground, ?string $backgrou
...
-                    // Add color reset before chunk and clear end of the string
-                    $text .= rtrim("\033[0m" . $chunk, "\033[0m");
-                } else {
-                    $text .= $chunk;
-                }
-            }
-        }
-
-        return $string . $text . "\033[0m";
+        return $string . ($text . "\033[0m");
```
paulbalandan, line 541: Why is the need for parenthesis?

paulbalandan (commented):
How is this a breaking change?

system/CLI/CLI.php
kenjis: Done.

system/CLI/CLI.php
kenjis: It seems rector added.
Removed.

kenjis:
I added docs.

paulbalandan (approved):

kenjis:
@paulbalandan Thank you!

PR #5899, opened 2022-04-15
docs: replace `/app/` with `app/`

kenjis:
**Description**
- Because there are more places written as `app/`.
- `/` at first implies root directory of the file system.
 
**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


samsonasik:
Thank you @kenjis

PR #5909, opened 2022-04-18
fix: incorrect type declarations

kenjis:
**Description**
I propose to fix incorrect type declarations as a bug fix in 4.x.

Related  #4356

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -82,7 +82,7 @@ class CodeIgniter
     /**
      * Current request.
      *
-     * @var CLIRequest|IncomingRequest|Request|null
+     * @var CLIRequest|IncomingRequest|null
      */
     protected $request;
 
@@ -432,7 +432,7 @@ class CodeIgniter
             return $returnResponse ? $possibleResponse : $possibleResponse->pretend($this->useSafeOutput)->send();
         }
 
-        if ($possibleResponse instanceof Request) {
+        if ($possibleResponse instanceof IncomingRequest) {
             $this->request = $possibleResponse;
         }
 
@@ -549,9 +549,11 @@ class CodeIgniter
      * Sets a Request object to be used for this request.
      * Used when running certain tests.
      *
+     * @param CLIRequest|IncomingRequest $request
+     *
      * @return $this
      */
-    public function setRequest(Request $request)
+    public function setRequest($request)
     {
         $this->request = $request;
 
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -432,7 +432,10 @@ class CodeIgniter
             return $returnResponse ? $possibleResponse : $possibleResponse->pretend($this->useSafeOutput)->send();
         }
 
-        if ($possibleResponse instanceof IncomingRequest) {
+        if (
+            $possibleResponse instanceof IncomingRequest
+            || $possibleResponse instanceof CLIRequest
+        ) {
             $this->request = $possibleResponse;
         }
 
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -44,7 +44,6 @@ use CodeIgniter\Language\Language;
 use CodeIgniter\Log\Logger;
 use CodeIgniter\Pager\Pager;
 use CodeIgniter\Router\RouteCollection;
-use CodeIgniter\Router\RouteCollectionInterface;
 use CodeIgniter\Router\Router;
 use CodeIgniter\Security\Security;
 use CodeIgniter\Session\Handlers\Database\MySQLiHandler;
@@ -597,7 +596,7 @@ class Services extends BaseService
      *
      * @return Router
      */
-    public static function router(?RouteCollectionInterface $routes = null, ?Request $request = null, bool $getShared = true)
+    public static function router(?RouteCollection $routes = null, ?Request $request = null, bool $getShared = true)
     {
         if ($getShared) {
             return static::getSharedInstance('router', $routes, $request);
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -914,7 +914,7 @@ abstract class BaseConnection implements ConnectionInterface
 
         $this->pretend(false);
 
-        if ($sql instanceof QueryInterface) {
+        if ($sql instanceof Query) {
             $sql = $sql->getOriginalQuery();
         }
 
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1510,6 +1510,13 @@ abstract class BaseConnection implements ConnectionInterface
     }
 
     /**
+     * Returns platform-specific SQL to disable foreign key checks.
+     *
+     * @return string
+     */
+    abstract protected function _disableForeignKeyChecks();
+
+    /**
      * Enables foreign key checks temporarily.
      */
     public function enableForeignKeyChecks()
@@ -1518,6 +1525,13 @@ abstract class BaseConnection implements ConnectionInterface
 
         return $this->query($sql);
     }
+
+    /**
+     * Returns platform-specific SQL to enable foreign key checks.
+     *
+     * @return string
+     */
+    abstract protected function _enableForeignKeyChecks();
 
     /**
      * Allows the engine to be set into a mode where queries are not
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -71,34 +71,9 @@ parameters:
 			path: system/Cache/Handlers/RedisHandler.php
 
 		-
-			message: "#^Call to an undefined method CodeIgniter\\\\HTTP\\\\Request\\:\\:getPost\\(\\)\\.$#"
-			count: 1
-			path: system/CodeIgniter.php
-
-		-
-			message: "#^Call to an undefined method CodeIgniter\\\\HTTP\\\\Request\\:\\:setLocale\\(\\)\\.$#"
-			count: 1
-			path: system/CodeIgniter.php
-
-		-
 			message: "#^Property CodeIgniter\\\\Database\\\\BaseBuilder\\:\\:\\$db \\(CodeIgniter\\\\Database\\\\BaseConnection\\) in empty\\(\\) is not falsy\\.$#"
 			count: 1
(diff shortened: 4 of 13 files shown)


system/CodeIgniter.php
```diff
@@ -446,7 +446,7 @@ protected function handleRequest(?RouteCollectionInterface $routes, Cache $cache
                 return $returnResponse ? $possibleResponse : $possibleResponse->pretend($this->useSafeOutput)->send();
             }
 
-            if ($possibleResponse instanceof Request) {
+            if ($possibleResponse instanceof IncomingRequest) {
```
lonnieezell, line 449: I believe it's possible for this to be CLIRequest also, isn't it?

system/Config/Services.php
```diff
@@ -545,7 +544,7 @@ public static function routes(bool $getShared = true)
      *
      * @return Router
      */
-    public static function router(?RouteCollectionInterface $routes = null, ?Request $request = null, bool $getShared = true)
+    public static function router(?RouteCollection $routes = null, ?Request $request = null, bool $getShared = true)
```
lonnieezell, line 547: This makes it more difficult for people to pass their own classes that implement the `RouteCollectionInterface`, which they currently can, so this would be BC.

system/Database/BaseConnection.php
```diff
@@ -897,7 +897,7 @@ public function prepare(Closure $func, array $options = [])
 
         $this->pretend(false);
 
-        if ($sql instanceof QueryInterface) {
+        if ($sql instanceof Query) {
```
lonnieezell, line 900: Again - BC break since they can currently pass custom implementations.

system/Config/Services.php
kenjis: The current Router does not depend on `RouteCollectionInterface`, it actually depends on `RouteCollection`.

If a user creates their own `RouteCollection` which implements `RouteCollectionInterface`,
it is not compatible with the `Router`, because some methods are missing in the `RouteCollectionInterface`.

system/Database/BaseConnection.php
kenjis: `QueryInterface` does not have `getOriginalQuery()`.
So it can't be used.

system/CodeIgniter.php
kenjis: Oh, thanks! I was confused with spark.

kenjis: Fixed.

MGatner:
It's a mess, which we've known. I'm still torn on what to do about it. We use methods that don't exist in our interfaces; we have classes that don't use our own interfaces. The current state is not healthy, but would require breaking changes in either direction to fix.

Are we "breaking" things that wouldn't work anyways? Maybe? Hard to know. This is why we were leaving this for version 5. 

Maybe an easier approach would be to split these apart into smaller PRs, by category or class components? Then we could deliberate on them individually?

kenjis:
@MGatner 
This PR drops the dependency on interfaces, but methods can be added to interfaces.

However, it may not be healthy to add more methods to the interface. Fat interfaces are a sign of bad design.

Are you of the opinion that mechanically deciding one way or the other is not a good idea? That may indeed be true.

MGatner:
This is mostly version 5 talk but I think if interfaces need more methods then they probably should be (or should have been) abstract classes. An abstract class gives us as developers the flexibility to add and adjust class needs while still supporting any extensions from other developers.

Interfaces are a great way to keep the framework extensible, especially with anything that is offered as a Service (and this easily replaceable) but as you said these should be kept "bare minimum" which is necessarily restrictive on framework use of them.

MGatner:
For this PR at hand, here's my opinion: split it out by independent components and then identifying the missing interface method/methods of each component.

If those methods appear to be highly specialized/advanced/niche (i.e. unlikely to be re-implemented) then I would consider that interface "safe for advanced replacement", that is, anyone who is supplying their own interfaced class is either extending our class or competent enough to handle a breaking change due to the replacement. 

If the missing methods appear to be core to the use of the class, or likely for extension/reimplementation then we should leave the interface, because someone supplying their own interfaces class has likely already gone through the trouble of figuring out the "method _____ doesn't exist" error and adjusted.

MGatner:
Looking back on this after three months of framework cleanup: I am in favor of these changes. The current state is disgraceful and I think leaving it for version 5 (which we still don't have planned) is perpetuating this sad state. Rip the bandaid off, do this the right way!

system/Config/Services.php
MGatner: @lonnieezell can we pick this discussion back up?

system/Database/BaseConnection.php
MGatner: @lonnieezell

MGatner:
Why 4.3? Just because there are so many potentially-breaking changes?

samsonasik:
I think changing signature should be major, but... if it really needed, should be a next minor 4.3

MGatner:
> I think changing signature should be major, but... if it really needed, should be a next minor 4.3

For sure it _should_ be major, but version 5 is a long way off and this is currently a mess. We define interfaces and then don't even stick to our own interfaces 😅 I think the 27 PHPStan errors that we can stop ignoring is very telling.

kenjis:
We defined interfaces, but our classes does not depend on the interfaces correctly.
That is if a dev creates a class that implements the interface, there is no guarantee that the class will work with CI4.
Because CI4 classes use methods or params that the interface does not have. E.g. https://github.com/codeigniter4/CodeIgniter4/issues/6224

As a result, no one could trust the interfaces, and VS Code always shows errors!

This is not a normal bug fix. So it should not go to `4.2.x`.

system/Database/BaseConnection.php
lonnieezell: Would it be less breaking to add `getOriginalQuery` to the interface? I realize they can just extend the original Query class to do custom implementations but that defeats the whole purpose of having an interface to begin with. 

Without looking at the history of things I want to say that method came after the initial release, but even if it didn't I think the bug would be that it wasn't included.

system/CodeIgniter.php
```diff
@@ -432,7 +432,10 @@ protected function handleRequest(?RouteCollectionInterface $routes, Cache $cache
             return $returnResponse ? $possibleResponse : $possibleResponse->pretend($this->useSafeOutput)->send();
         }
 
-        if ($possibleResponse instanceof Request) {
+        if (
+            $possibleResponse instanceof IncomingRequest
```
lonnieezell, line 436: While I think it unlikely, what if someone had created their own Request class? LIke perhaps a more PSR-friendly version. What is the benefit of this change other than limiting those options for devs?

lonnieezell (commented):
I started leaving comments but they all became the same thing - what is the benefit to developers of these changes? It seems that every single one of them reduces the ability of devs to extend and customize the framework. As far as I can tell none of them really has a benefit....

lonnieezell:
I know I typically stay out of these, and the arguments have already been made, but I agree that if we do it we should make it at least 4.3. Personally, I'm more for adding the methods to the traits, if they're actually needed by all implementing classes, and keeping the flexibility and extendability in the framework. That's kind of the whole point behind Services, which become a bit irrelevant if if you can't actually replace framework classes with your own implementations. I know you can extend the class and still get around it but that feels messier to me. 

The other option is take a nuke to it all and get rid of interfaces that don't currently have more than one implementation because if we're hard-coding dependencies, they don't matter anymore. But that seems like a worse solution.

MGatner:
I'm fine with this alternate approach of expanding interfaces to cover the actual method dependencies. This prevents a breaking change for extensions of the class depending on the interface. Also if someone _has_ actually provided their own class fulfilling the interface they have likely already added these methods due to "method not exists" errors from our disjoint.

kenjis:
I created this PR, because it seems we have a rule that we change our Interfaces only in major version up.

kenjis:
I sent a PR to fix ValidationInterface with the alternate approach of expanding interfaces to cover the actual method dependencies. See #6253

MGatner:
> we have a rule that we change our Interfaces only in major version

I don't think this rule is any more important than general breaking changes. In these instances where signatures use interfaces but rely on extra methods I suspect that altering the interfaces will have less impact since most developer classes that fulfill the interfaces will be extensions of our classes already. Even if they aren't a developer who tried using their own implementation would quickly find that they had to add the extra methods already.

MGatner:
The rest of this is waiting to see how the ValidationInterface changes are received in 4.3

martinemily3:
https://matemaskk-login.knorish.com/
https://robinhudlogin.jigsy.com/
https://gemini-log-official.knorish.com/
https://kucoinloginu.bravesites.com/
https://robinn-hood-loggin.godaddysites.com/
https://robinnhudd_logg_official.godaddysites.com/
https://metmask-log-official.knorish.com/
https://kucoin-log-official.knorish.com/
https://blockfilogiofficial.knorish.com
https://blockfiloginn1.knorish.com
https://geminii_logines.godaddysites.com/
https://gemini_ww_login1.godaddysites.com/
https://gemini_x_login93.godaddysites.com/
https://metamaskio-login.knorish.com/
https://metausmaskulogin.knorish.com/
https://metamaskk-io.knorish.com/
https://metumaskk-log-in.knorish.com/
https://metemas-klogin.knorish.com/
https://genimi-logg-official.godaddysites.com/
https://metam-musk-log.knorish.com/
https://gemini-us-logi9.godaddysites.com/
https://metaska-sklogin.knorish.com/
https://lunologinn.knorish.com
https://phantomloginn.knorish.com
https://luno_login2.godaddysites.com/
https://luno_walletlogin.godaddysites.com/
https://krakenwallet.knorish.com/
https://exodus_login.godaddysites.com/
https://exoduslog.knorish.com/
https://lunowalletlogin94.godaddysites.com
https://phantom_login.godaddysites.com/
https://geminlog.knorish.com/

MGatner:
@kenjis Is there any part of this that should still be considered for 4.3?

kenjis:
@MGatner Ignore this PR; there are no further fixes I wish to include in 4.3.
 In fact, I believe quite a bit has already been fixed in 4.3.

PR #5910, opened 2022-04-19
Comments Update in routes

lyimolucasl:
Missing asterisk in block comment opening


paulbalandan:
Sorry, but those blocks function as multiline comments. What you're proposing is for doc blocks, which in this case are not.

PR #5913, opened 2022-04-20
Fix onlyDeleted and purgeDeleted not working for non-nullable timestamp

t-kouyama:
**Description**
In MySQL, the default value for a non-nullable datetime is '0000-00-00 00:00:00', which would be included in 'IS NOT NULL'.
The default value for all timestamp formats can also be treated as 0, so using '> 0' as the where condition will give the expected behavior.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -382,7 +382,7 @@ class Model extends BaseModel
     protected function doPurgeDeleted()
     {
         return $this->builder()
-            ->where($this->table . '.' . $this->deletedField . ' IS NOT NULL')
+            ->where($this->table . '.' . $this->deletedField . ' > 0')
             ->delete();
     }
 
@@ -393,7 +393,7 @@ class Model extends BaseModel
      */
     protected function doOnlyDeleted()
     {
-        $this->builder()->where($this->table . '.' . $this->deletedField . ' IS NOT NULL');
+        $this->builder()->where($this->table . '.' . $this->deletedField . ' > 0');
     }
 
     /**


datamweb:
> 
Did you define time fields in the **model** as follows?
```
    public function up()
    {
        //up run
        $this->forge->addField([
            'id'  => [
                    'type'           => 'INT',
                    'constraint'     => 11,
                    'unsigned'       => true,
                    'auto_increment' => true,
            ],
               
            'created_at datetime default NULL',
            "updated_at datetime default NULL", 
            "deleted_at datetime default NULL",

        ]);
```

iRedds:
> In MySQL, the default value for a non-nullable datetime is '0000-00-00 00:00:00'
This is true only for non-strict mode.

kenjis:
Thank you for taking time.

See https://github.com/codeigniter4/CodeIgniter4/actions/runs/2196734762
Tests for Postgre and OCI8 failed.

kenjis:
> If `$useSoftDeletes` is true, then the find*() methods will not return any rows where ‘**deleted_at IS NOT NULL**’. 
https://codeigniter4.github.io/CodeIgniter4/models/model.html#withdeleted

This is documented, but It might be better to properly describe the column definition in the $useSoftDeletes section.
https://codeigniter4.github.io/CodeIgniter4/models/model.html#usesoftdeletes

t-kouyama:
Thank you for your comments.

The find*() method does not exclude 'deleted_at IS NOT NULL', it only includes 'IS NULL'.
Oddly enough, these have different results for non-nullable datetime in MySQL.
I have seen unofficial documents that define deleted_at as non-nullable and have had problems using onlyDeleted.

The official document does not require deleted_at to be nullable, which is different from the actual behavior, so I think either one needs to be fixed.
Since my fix has so many problems, I think it would be better to state in the document that deleted_at needs to be nullable.

kenjis:
I sent a PR to fix the docs: #5989

t-kouyama:
kenjis has fixed the document, so this has been closed.

PR #5915, opened 2022-04-21
docs: improve migration namespace descriptions

kenjis:
**Description**
Setting `null` in `$namespace` means all namespaces. It is not intuitive.
So we better explain it in all relevant places.
- add explanation for `null` in `$namespace`
- improve testing/database.rst format
- improve dbmgmt/migration.rst format

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/system/Database/MigrationRunner.php
+++ b/system/Database/MigrationRunner.php
@@ -40,7 +40,8 @@ class MigrationRunner
     protected $table;
 
     /**
-     * The Namespace  where migrations can be found.
+     * The Namespace where migrations can be found.
+     * `null` is all namespaces.
      *
      * @var string|null
      */
--- a/system/Test/CIUnitTestCase.php
+++ b/system/Test/CIUnitTestCase.php
@@ -116,7 +116,7 @@ abstract class CIUnitTestCase extends TestCase
 
     /**
      * The namespace(s) to help us find the migration classes.
-     * Empty is equivalent to running `spark migrate -all`.
+     * Empty is equivalent to running `spark migrate --all`.
      * Note that running "all" runs migrations in date order,
      * but specifying namespaces runs them in namespace order (then date)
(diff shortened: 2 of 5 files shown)


lonnieezell:
I think this looks good but would it be easy enough to make it do all namespaces when either `null` or `--all` were present? I think that might make it more intuitive while not harming BC.

kenjis:
@lonnieezell Sorry, I can't get what you mean.
`null` is a property value, but `--all` is a command option.

lonnieezell:
Sorry. I meant it would be nice to set $namespace to either `null` or the string `--all` within a test case to have it run all migrations. That way the property value and the command option are the same. Or maybe `all` would be enough? 

```php

class MyTestCase extends DatabaseTestCase
{
    protected $namespace = '--all';
    // or
    protected $namespace = 'all';
}
```

MGatner (approved):
Very nice explanations.

kenjis:
@MGatner How about adding new special strings for all namespaces as @lonnieezell says?

```php
protected $namespace = 'all';
// or
protected $namespace = '*';
```
Personally `$namespace = '--all'` seems odd to me.

MGatner:
I'm not opposed. Out of all those I prefer `*` since there is no possible confusion that it is a namespace. 

Honestly I don't know how to feel about this feature. IMO running tests against a "partially-migrated" database seems like setup for a logical failure. I frequently have migrations modifying the same table across namespaces (like Myth:Auth creates `users` then my app adds `users.firstname`) and these would either fail or be a false representation of the table if I only ran one namespace or the other. 

What selective namespace migrating does add is efficiency, but I think this is a problem better solved with other tools (see my other comment about generating a "one-and-done" schema, like Rails).

PR #5919, opened 2022-04-23
[PHPStan] Allow null type for request property in CodeIgniter class

samsonasik:
to allow remove ignored unreachable phpstan notice as the property is not filled on __construct.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -116,11 +116,6 @@ parameters:
 			path: system/CodeIgniter.php
 
 		-
-			message: "#^Unreachable statement \\- code above always terminates\\.$#"
-			count: 1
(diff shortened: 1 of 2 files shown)


samsonasik:
Ref on rector dev-main using check over stmts with unreachable phpstan attribute, which `@var` make certain even with `treatPhpDocTypesAsCertain: false`

  https://github.com/rectorphp/rector-src/pull/2132#issuecomment-1107249764

samsonasik:
All green 🎉 , it is also prepare for nullable union when we have php 8 requirement in the future to avoid error accessed before initialization

kenjis (approved):

paulbalandan (approved):

Issue #5922, opened 2022-04-25
Bug: [Validation] field with asterisk (nested array) throws exception


mostafakhudair:
### PHP Version

7.4

### CodeIgniter4 Version

4.1.9

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

apache

### Database

MySQL 4.9

### What happened?

```php
    protected array $validationRules = [
        'customer_account_number'                 => [
            'label' => 'ACCOUNT NUMBER',
            'rules' => 'required|exact_length[5]',
        ],
        'debit_amount'                            => [
            'label' => 'DEBIT AMOUNT',
            'rules' => 'required|decimal|is_natural_no_zero',
        ],
        'beneficiaries_accounts.*.account_number' => [
            'label' => 'BENEFICIARY ACCOUNT NUMBER',
            'rules' => 'exact_length[5]',
        ],
        'beneficiaries_accounts.*.credit_amount'  => [
            'label' => 'CREDIT AMOUNT',
            'rules' => 'required|decimal|is_natural_no_zero',
        ],
        'beneficiaries_accounts.*.purpose'        => [
            'label' => 'PURPOSE',
            'rules' => 'required_without[beneficiaries_accounts.*.account_number]|min_length[3]|max_length[255]',
        ],
    ];
```

```php
    $this->validation->setRules($model->getValidationRules());
```

**Case 1:**
```php
    $data = [
        "customer_account_number" => "A_490",
        "debit_amount"            => "1500",
        "beneficiaries_accounts"  => [],
    ];
```

**Case 2:**
```php
    $data = [
        "customer_account_number" => "A_490",
        "debit_amount"            => "1500",
        "beneficiaries_accounts"  => [
            "account_1" => [
                "account_number" => "A_103",
                "credit_amount"  => 1000,
                "purpose"        => "",
            ],
        ],
    ];
```

**Case 3:**
```php
    $data = [
        "customer_account_number" => "A_490",
        "debit_amount"            => "1500",
        "beneficiaries_accounts"  => [
            "account_1" => [
                "account_number" => "A_103",
                "credit_amount"  => 1000,
                "purpose"        => "Personal",
            ],
            "account_2" => [
                "account_number" => "A_258",
                "credit_amount"  => null,
                "purpose"        => "A",
            ],
        ],
    ];
```

```php
    if (! $this->validation->run($data))) {
[...]

PR #5925, opened 2022-04-26
Fix: Validation. Fields with an asterisk.

iRedds:
Fixes https://github.com/codeigniter4/CodeIgniter4/issues/5922

**Description**
If the field contains a asterisk and is not found in the validation dataset, then `array_filter()` returns an empty array. This case is not handled, which may cause errors.
So I suggest setting ``$values = [$field => null]`` if the field is not found.

It also prevents a TypeError exception from being thrown if an array is passed to the function (rule) instead of a string. But this only applies to fields not found.

### DIFF
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -145,12 +145,16 @@ class Validation implements ValidationInterface
                 $rules = $this->splitRules($rules);
             }
 
-            $values = strpos($field, '*') !== false
-                ? array_filter(array_flatten_with_dots($data), static fn ($key) => preg_match(
+            if (strpos($field, '*') !== false) {
+                $values = array_filter(array_flatten_with_dots($data), static fn ($key) => preg_match(
                     '/^' . str_replace('\.\*', '\..+', preg_quote($field, '/')) . '$/',
                     $key
-                ), ARRAY_FILTER_USE_KEY)
-                : dot_array_search($field, $data);
+                ), ARRAY_FILTER_USE_KEY);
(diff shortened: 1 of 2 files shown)


kenjis:
Thank you!
Can you add test code?

iRedds:
I don't know how to correctly test this case. If you have suggestions or desires to take this PR from me, I will not mind)

kenjis:
This PR is a bug fix.
So first we write tests to reproduce the bug, then we fix the product code. 
Otherwise, we will not know if the bug has been fixed.

How did you write the fix?

iRedds:
I'm not sure about the test, but so far I can't think of anything else to check for a null value.

kenjis:
@iRedds Thank you.
But your test code is difficult to understand.
I will take over this PR.

iRedds:
@kenjis The test expects null to be passed to the processRules method if the key is not found.

iRedds:
@kenjis 
The exception is thrown because the passed value does not match the expected value, not because the key was not found.
```php
// This code will throw a TypeError exception.
$this->validation->setRules(['a' => 'alpha'])->run(['a' => []]);
```
Due to the [strange behavior](https://forum.codeigniter.com/showthread.php?tid=81764) of the dot_array_search helper, the code below will also throw an exception. Because instead of null for a key not found, an array will be passed.

```php
// This code will throw a TypeError exception.
$this->validation->setRules(['a.b.c' => 'alpha'])->run(['a' => ['b' => []]]);
```

PR #5927, opened 2022-04-27
docs: fix BaseCommand $params @param

### LINKED ISSUE #5926: Bug: BaseCommand $params Type
### PHP Version

7.4

### CodeIgniter4 Version

develop

Git

Linux

### Which server did you use?

cli

### Database

_No response_

### What happened?

I believe that the typing on `BaseCommand::run()` was supposed to be `@param array<int, string> $params` (a "list" of strings), not `@param array<string, mixed> $params` (string keys, mixed values).

https://github.com/codeigniter4/CodeIgniter4/blob/bca263c5fef57cf3152b31916c26a63496341be6/system/CLI/BaseCommand.php#L96-L101

I started tracing this back and it was taking too long but should be quick follow-up at some point.

### Steps to Reproduce

See code

### Expected Output

See code

### Anything else?

_No response_

kenjis:
**Description**
Fixes #5926
- fix types in doc comments
- update phpstan-baseline.neon.dist 

I got the following errors, so I also fixed them.
```
Run vendor/bin/phpstan analyse
Note: Using configuration file /home/runner/work/CodeIgniter4/CodeIgniter4/phpstan.neon.dist.
   0/428 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0%
 360/428 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░]  84%
 428/428 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ------------------------------------------------------- 
  Line   system/Autoloader/FileLocator.php                      
 ------ ------------------------------------------------------- 
  269    Offset 'prefix' does not exist on array{path: mixed}.  
 ------ ------------------------------------------------------- 

  Line   system/Database/MySQLi/Connection.php                                  
         Ignored error pattern #^Cannot access property \$affected_rows on      
         bool\|object\|resource\.$# in path                                     
         /home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/MySQLi/Co  
         nnection.php was not matched in reported errors.                       

  Line   system/Session/Handlers/DatabaseHandler.php                            
         Ignored error pattern #^Access to an undefined property                
         Config\\App\:\:\$sessionDBGroup\.$# in path                            
         /home/runner/work/CodeIgniter4/CodeIgniter4/system/Session/Handlers/D  
         atabaseHandler.php was not matched in reported errors.                 
```
https://github.com/codeigniter4/CodeIgniter4/runs/6201648310?check_suite_focus=true

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/system/CLI/BaseCommand.php
+++ b/system/CLI/BaseCommand.php
@@ -96,7 +96,7 @@ abstract class BaseCommand
     /**
      * Actually execute a command.
      *
-     * @param array<int|string, string> $params
+     * @param array<int|string, string|null> $params
      */
     abstract public function run(array $params);
 
--- a/system/CLI/BaseCommand.php
+++ b/system/CLI/BaseCommand.php
@@ -96,7 +96,7 @@ abstract class BaseCommand
     /**
      * Actually execute a command.
      *
-     * @param array<string, mixed> $params
+     * @param array<int|string, string> $params
      */
     abstract public function run(array $params);
 
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -186,11 +186,6 @@ parameters:
 			path: system/Database/MigrationRunner.php
 
 		-
-			message: "#^Cannot access property \\$affected_rows on bool\\|object\\|resource\\.$#"
-			count: 1
-			path: system/Database/MySQLi/Connection.php
-
-		-
 			message: "#^Cannot access property \\$errno on bool\\|object\\|resource\\.$#"
 			count: 1
 			path: system/Database/MySQLi/Connection.php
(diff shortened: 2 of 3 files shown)


system/CLI/BaseCommand.php
```diff
@@ -96,7 +96,7 @@ public function __construct(LoggerInterface $logger, Commands $commands)
     /**
      * Actually execute a command.
      *
-     * @param array<string, mixed> $params
+     * @param array<int|string, string> $params
```
paulbalandan, line 99: How about if the option is used as a flag (i.e. without values)? `php spark foo --sort`
I think the value would be `true` or `null`? Not sure which one.

kenjis: `null`.

kenjis: Fixed.

paulbalandan: Ah yes. The `true` is when you use `CLI::getOption()` to get the value.

paulbalandan (approved):

MGatner (approved):
I noticed that PHPStan released 1.6.2 and it had a significant different set of detection where I was working on Myth:Auth - probably the same source here. 

Thanks for the quick fix in the command types!

PR #5940, opened 2022-04-30
fix: `dot_array_search()` unexpected behavior

### LINKED ISSUE #5939: Bug: `dot_array_search()` unexpected behavior
From: https://forum.codeigniter.com/showthread.php?tid=81764

Why is the result not `null`?
https://github.com/codeigniter4/CodeIgniter4/blob/aee95af56f9d036aca55ee254740fb7e50291a1d/tests/system/Helpers/ArrayHelperTest.php#L40-L49

The production code came from https://github.com/codeigniter4/CodeIgniter4/commit/6eefb80d6f01b03f2c5d8e7fc1b828264156f41a
Ref #3909
Related https://github.com/codeigniter4/CodeIgniter4/pull/5925#issuecomment-1113899854

kenjis:
**Description**
Fixes #5939

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -89,8 +89,8 @@ if (! function_exists('_array_search_dot')) {
             return _array_search_dot($indexes, $array[$currentIndex]);
         }
 
-        // Otherwise we've found our match!
-        return $array[$currentIndex];
+        // Otherwise, not found.
+        return null;
     }
 }
 
--- a/tests/system/Helpers/ArrayHelperTest.php
+++ b/tests/system/Helpers/ArrayHelperTest.php
@@ -77,15 +77,24 @@ final class ArrayHelperTest extends CIUnitTestCase
 
     public function testArraySearchDotMultiLevels()
     {
-        $data1 = ['bar' => [['foo' => 'baz']]];
-        $data2 = ['bar' => [
-            ['foo' => 'bizz'],
-            ['foo' => 'buzz'],
-        ]];
-        $data3 = ['baz' => 'none'];
-
+        $data1 = [
+            'bar' => [
+                ['foo' => 'baz'],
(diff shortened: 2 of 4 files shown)


sfadschm:
I think I would prefer an exception to be thrown ("Array key does not exist." or so).
That is because I think `null` might be a valid return of a successful search.

I think there is two cases where the statement can be reached:
- An array of post data which could be manipulated by the user/browser. In that case the search should fail rather then return (say as an additional validation).
- The array to search was built by ourselves. In that case we should know which keys are accessible and it would make sense to me to let the dev know that the error occured instead of failing silently.

kenjis:
> I think null might be a valid return of a successful search.

Yes.

But this returns null. Do you want to change to throw an exception?
https://github.com/codeigniter4/CodeIgniter4/blob/aee95af56f9d036aca55ee254740fb7e50291a1d/tests/system/Helpers/ArrayHelperTest.php#L99-L108

iRedds:
@sfadschm What if, instead of an exception, add a third argument to the function, which will contain a default value for the case when the key is not found?

sfadschm:
> But this returns null. Do you want to change to throw an exception?

Oh, right. I missed that we already use `null` as a failure indicator. I'm fine with this solution then! 

@iRedds: Sounds like a nice addition, but might be out of scope for this PR. I also think changing the signature would be another BC break and while `null` might be a valid return value on success, it is still likely to be to most rarely used one.

sfadschm:
Definitely needs a changelog and/or user guide entry though ☺

kenjis:
Added the documentation.

kenjis:
Rebased.
Please review.

paulbalandan (approved):

Issue #5905, opened 2022-04-17
Bug: \CodeIgniter\Entity::hasChanged() is unreliable for non-string casts


rmilecki:
### PHP Version

8.0

### CodeIgniter4 Version

4.1.9

### CodeIgniter4 Installation Method

Manual (zip or tar.gz)

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

cgi-fcgi

### What happened?

The problem is that `\CodeIgniter\Entity::hasChanged()` may return `true` even if there isn't any data change.

This problem is caused by two facts:

1. Casting is performed only when *reading* properties
2. `hasChanged()` ignores all casts (it compares `$original` with `$attributes` directly, bypassing `castAs()`)

### Steps to Reproduce

```php
/* This matches what \CodeIgniter\Database\BaseResult does */
$data = [
    'id' => '1',
    'name' => 'John',
    'age' => '35',
];
$entity = new class ($data) extends \CodeIgniter\Entity {
    protected $casts = [
        'id'   => 'integer',
        'name' => 'string',
        'age'  => 'integer',
    ];
};
$entity->syncOriginal();

var_dump($entity->age);
var_dump($entity->hasChanged());
$entity->age = 35;
var_dump($entity->age);
var_dump($entity->hasChanged());
```

### Expected Output

```
int(35)
bool(false)
int(35)
bool(false)
```

### Anything else?

Current output:

```
int(35)
bool(false)
int(35)
bool(true)
```

kenjis:
@rmilecki Thank you for reporting.

#5678 indicates that if you use `fill()`, you need to pass the same typed value from the database.

Do you think this bug can be fixed?

kenjis:
Ref https://github.com/codeigniter4/CodeIgniter4/issues/2441#issuecomment-562604358

kenjis:
`\CodeIgniter\Entity` is deprecated. Please use `CodeIgniter\Entity\Entity` class instead.

kenjis:
Workaround:
```php
<?php

namespace App\Controllers;

class Home extends BaseController
{
    public function index()
    {
        $data = [
            'id' => '1',
            'name' => 'John',
            'age' => '35',
        ];
        $entity = new class ($data) extends \CodeIgniter\Entity\Entity {
            protected $casts = [
                'id'   => 'integer',
                'name' => 'string',
                'age'  => 'integer',
            ];

            public static function setAge($value)
            {
                return (string) $value;
[...]

PR #5944, opened 2022-04-30
Entity: cast scalar types when setting

rmilecki:
**Description**
```
This way Entity stores its attributes using types as specified in the
$casts. This results in:

1. Cleaner logic: data is always stored as expected. Not only read.
2. Fixing reliability of hasChanged()

Without this change what seemed like setting already set (identical to
read) value could actually change type and result in hasChanged()
unexpectedly returning true.

Ref: https://github.com/codeigniter4/CodeIgniter4/issues/5905
Signed-off-by: Rafał Miłecki <rafal@milecki.pl>
```

### DIFF
--- a/system/Entity/Cast/BooleanCast.php
+++ b/system/Entity/Cast/BooleanCast.php
@@ -23,4 +23,12 @@ class BooleanCast extends BaseCast
     {
         return (bool) $value;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function set($value, array $params = []): bool
+    {
+        return (bool) $value;
+    }
 }
--- a/system/Entity/Cast/FloatCast.php
+++ b/system/Entity/Cast/FloatCast.php
@@ -23,4 +23,12 @@ class FloatCast extends BaseCast
     {
         return (float) $value;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function set($value, array $params = []): float
+    {
+        return (float) $value;
+    }
 }
--- a/system/Entity/Cast/IntegerCast.php
+++ b/system/Entity/Cast/IntegerCast.php
@@ -23,4 +23,12 @@ class IntegerCast extends BaseCast
     {
         return (int) $value;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function set($value, array $params = []): int
+    {
+        return (int) $value;
+    }
 }
--- a/system/Entity/Cast/StringCast.php
+++ b/system/Entity/Cast/StringCast.php
@@ -23,4 +23,12 @@ class StringCast extends BaseCast
     {
         return (string) $value;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function set($value, array $params = []): string
+    {
+        return (string) $value;
+    }
 }


kenjis:
Thank you for sending a PR.

This PR is not unit tested, and an existing test failed.

Your PRs need to be GPG-signed before they will be considered.

kenjis:
Aside from the lack of testing, this appears to be a major design change and requires careful consideration.

MGatner:
Entities have gotten very confusing, but: this is a mistake. The stored entity value is supposed to represent its "database state", so when Model converts an entity in `objectToArray()` it can use `Entity::toRawArray` deliberately to bypass the casts. If we transform the "raw" attributes during `set()` to match their cast state we are essentially removing the raw state. This might work fine for some types (like boolean) but won't hold consistent, especially for database strict mode. 

On a personal note, I've found `Entity` very frustrating to work with in a NoSQL context specifically because of these assumptions around their "raw" value states. This might be a PR to re-inspect for version 5 when we think about our database layer refactor and if/how NoSQL support is included.

MGatner:
Follow-up example to my personal note: https://github.com/tattersoftware/codeigniter4-firebase/blob/develop/src/Firestore/Collection.php#L188

Notice how for my Firebase (NoSQL) driver is explicitly use `toArray()` instead of `toRawArray()`.

rmilecki:
Thanks for your comments.

If I can share my opinion, the whole casting in `Entity` is just totally confusing. If:
1. We don't cast database values
2. We cast user-set values

that is inconsistent and results in hard to understand behaviour.

Maybe there are good reasons for that I simply can't understand (I'm not advanced programmer). Please just note that end users can get confused by that. Also it seems to break `Entity::hasChanged()` as described in the https://github.com/codeigniter4/CodeIgniter4/issues/5905 . So maybe you can think of some simpler & more reliable alternative solution for further releases.

MGatner:
I'm sorry for the confusion. I would like to help you work through this so we can get your suggestions on how this could have been more clear. I think you still aren't understanding how Entity casting works, as this is (mostly) untrue:

> We cast user-set values

The idea is that the Entity class gives the developer developer-friendly types and the database database-friendly types. With a few exceptions, Entity does *not* cast user-set values (this was what your PR was adding). From the User Guide:

> Casting only affects when values are read. No conversions happen that affect the permanent value in either the entity or the database.

https://codeigniter4.github.io/CodeIgniter4/models/entities.html#property-casting

If there's some confusion it probably comes from some of the Cast names. For example CSV casting is "database: csv_string, developer: array", whereas scalar casts are names for the developer types (Boolean Cast is "developer: boolean"). I'm not sure what can be done to improve that but I think the User Guide describes each cast type pretty well.

rmilecki:
Thanks @MGatner, I just read everything carefully again and I think I understand it now.

AFAIU: casting is meant to affect reads only making accessing (reading) data easier.

Now I think about it it sounds sane and useful. It should be enough for most (or all) cases. It does what it's supposed to do.

My real problem is unreliable `hasChanged()` (https://github.com/codeigniter4/CodeIgniter4/issues/5905). While redesigning `Entity` (e.g. as in this pull request) could be a workaround for it, it's probably not the best choice. I guess this pull request was correctly `Closed` and we should look for another `hasChanged()` issue solution instead.

kenjis:
I think a radical solution requires a redesigned new Entity. But that would be a major undertaking and would destroy compatibility.

rmilecki:
@kenjis: we could introduce another class though, right?
`\CodeIgniter\Entity\StrictEntity`
`\CodeIgniter\Entity\TypedEntity`
`\CodeIgniter\Entity\Entity2`
or whatever we call it ;)

kenjis:
@rmilecki Yes.

lonnieezell:
I'm honestly not fond of having multiple entity classes personally, as it just makes things more confusing when trying to decide which to use. I think we'll need to brainstorm on a good solution for this.

MGatner:
My opinion... the current Entity is trying to do too many things. It is pretty useful as a "Jack of all trades" but starts to fail when you need depth or complexity. Working with NoSQL it is almost useless, since data types are much more consistent between database and code. Trying to use them as dynamic objects not connected to databases sometimes runs into weird behavior. 

I mostly agree with Lonnie that we don't need seven different Entities in the framework - mostly those should be an extension class in personal projects or libraries - but I do think we could get creative for version 5.

PR #5949, opened 2022-05-02
[docs] add DB nested transaction

michaelrk02:
There's an explanation that a nested transaction is possible in CodeIgniter 4 by keeping track of the transaction depth (see https://github.com/codeigniter4/CodeIgniter4/blob/febc03a555b222a7a5617f04d7ba2058700cfdb1/system/Database/BaseConnection.php#L766).

I want it to be clearly documented so developers can save their time instead of figuring out by examining the system/ folder.

This implies that every start (transStart/transBegin) command should behave the same as others (pushing the transaction depth) and complete (transComplete/transCommit/transRollback) command should also behave the same as the others (popping the transaction depth) as a requirement. I've examined the code and they actually behave like that so I think it's safe to document this feature.

### DIFF
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -110,5 +110,4 @@ track of the transaction "depth" and only take action at the outermost layer
 
 .. literalinclude:: transactions/007.php
 
-.. note:: Use this technique ONLY IF you actually know what you are doing.
-
+.. note:: Use this technique carefully in order to keep the commands balanced.
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -97,3 +97,18 @@ If you would like to run transactions manually you can do so as follows:
 
 .. note:: Make sure to use ``$this->db->transBegin()`` when running manual
     transactions, **NOT** ``$this->db->transStart()``.
+
+Nested Transactions
+===================
+
+In CodeIgniter, transactions can be nested in a way such that only the
+outmost or top-level transaction commands are executed. You can include as
+many pairs of ``transStart``/``transComplete`` or ``transBegin``/``transCommit``/``transRollback``
+as you want inside a transaction block and so on. CodeIgniter will keep
+track of the transaction "depth" and only take action at the outermost layer
+(zero depth).
+
+.. literalinclude:: transactions/007.php
+
+.. note:: Use this technique ONLY IF you actually know what you are doing.
+
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -103,7 +103,7 @@ Nested Transactions
 
 In CodeIgniter, transactions can be nested in a way such that only the
 outmost or top-level transaction commands are executed. You can include as
-many pairs of ``transStart``/``transComplete`` or ``transBegin``/``transCommit``/``transRollback``
+many pairs of ``transStart()``/``transComplete()`` or ``transBegin()``/``transCommit()``/``transRollback()``
 as you want inside a transaction block and so on. CodeIgniter will keep
 track of the transaction "depth" and only take action at the outermost layer
 (zero depth).
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -110,4 +110,7 @@ track of the transaction "depth" and only take action at the outermost layer
 
 .. literalinclude:: transactions/007.php
 
-.. note:: Use this technique carefully in order to keep the commands balanced.
+.. note:: In case the structure is far more complex, it's your responsibility
+    to ensure that the inner transactions can reach the outermost layer again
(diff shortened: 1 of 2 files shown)


user_guide_src/source/database/transactions.rst
```diff
@@ -97,3 +97,17 @@ If you would like to run transactions manually you can do so as follows:
 
 .. note:: Make sure to use ``$this->db->transBegin()`` when running manual
     transactions, **NOT** ``$this->db->transStart()``.
+
+Nested Transactions
+===================
+
+In CodeIgniter, transactions can be nested in a way such that only the
+outmost or top-level transaction commands are executed. You can include as
+many pairs of ``transStart``/``transComplete`` or ``transBegin``/``transCommit``/``transRollback``
```
kenjis, line 106:
```suggestion
many pairs of ``transStart()``/``transComplete()`` or ``transBegin()``/``transCommit()``/``transRollback()``
```

user_guide_src/source/database/transactions.rst
```diff
@@ -97,3 +97,17 @@ If you would like to run transactions manually you can do so as follows:
...
+as you want inside a transaction block and so on. CodeIgniter will keep
+track of the transaction "depth" and only take action at the outermost layer
+(zero depth).
+
+.. literalinclude:: transactions/007.php
+
+.. note:: Use this technique carefully in order to keep the commands balanced.
```
kenjis, line 113: Sorry, I don't understand "keep the commands balanced".

michaelrk02: From my observation this concept is a bit similar to a stack. The goal is to make the stack elements (here transaction depth) to be zero by pusing and popping with the same amount. There's also an interesting [balanced parentheses problem](https://avikdas.com/2020/01/28/the-balanced-parentheses-problem.html) that will relate to this issue (suppose start commands are left parentheses and complete commands are right parentheses). We should keep them as balanced as possible to prevent unintended commits/rollbacks (due to unbalanced commands). At the worst case scenario, the transaction may never be completed (transaction depth still above zero because the start commands are executed more times than the complete commands), or completed early (transaction depth reaches zero before it actually ends because the complete commands are executed more times than the start commands). So (I think?) we also need to warn devs of that too, or is it not really necessary? (i.e. common sense of a dev)

michaelrk02: In order to verify the "balancedness" of a transaction, especially when the structure is too complex such that it involves many functions that intiate another transactions under the hood, I use a bit of "mathematical induction" here.

In simple words, if we initiate a transaction and inside that transaction block there exist functions that issue another transactions, and ensuring if those functions **are guaranteed** to be balanced, we can safely say that the overall transaction block **is** balanced (transaction depth can reach zero at the intended time).

Furthermore, the depth before calling those functions must be at the same number with the depth after calling those functions (and those functions are responsible for pushing and restoring the depth to it's initial value). So we can pop out the transaction level at the correct time to match the pushed start command at the beginning of the block.

michaelrk02: I pay attention to this issue because sometimes there's an *if* condition inside the code that will determine the action of the transaction (whether to commit/rollback it or not) and we should ensure that a transaction ends regardless of the condition (and commit/rollback should be issued **only** once to match the begin command).

Some devs with an experience dealing with similar issues like [reference counting](https://en.m.wikipedia.org/wiki/Reference_counting), [mutual exclusion with recursive locks](https://en.m.wikipedia.org/wiki/Reentrant_mutex), and etc would probably care about this too

kenjis: Thank you for explanation. I got it now. I just didn't understand the meaning of the English phrase "keep the commands balanced".

If you want to add a note, the following is fine to me. Or make it a bit shorter.
> You must keep commands balanced to prevent unintended commits/rollbacks due to unbalanced commands. 

This seems like common sense, but I think it would be kinder to have a note.

michaelrk02: Thanks, how about:

> In case the structure is far more complex, it's your responsibility to ensure that the inner transactions can reach the outermost layer again in order to be fully executed by the database, thus prevents unintended commits/rollbacks

I think that explains more straightforward (and still in the context of the paragraph) and thus prevents confusion with word "balanced". Experienced devs should be able to infer that they should maintain the balance of the transaction commands.

I also added a comment on the source code example: "... required to finish the inner transaction" implying that the commands must be balanced

What do you think?

kenjis: Okay, it is good to me. Thanks!

michaelrk02: Okay, done applying all suggestions & changes. It's good to go :+1:

kenjis (approved):

user_guide_src/source/database/transactions.rst
kenjis: @michaelrk02 Thank you!

PR #5950, opened 2022-05-02
PHPUnitTest: Extend the default PHPUnit result printer with a modern, pretty printer

datamweb:
Hello friends.
PR is about beautifying the **output** of the **unit test in the terminal**.
I am currently learning unit test. What I saw at the beginning of running command `vendor/bin/phpunit` caused phrases like `...FF.FE.....F..` To see in the terminal.
I'm a beginner, I did not know what **pass**, **fail**, **error**, **skipped**, **incomplete**, **risky** means. I now know.
In my opinion, if this package was used. In executing Command `vendor/bin/phpunit`, I was not scared at first glance.
Please consider executing command `vendor/bin/phpunit` by a novice user unfamiliar with command **PHPUnit** as well as a user whose native language is not english. Check the subject.
Using this package made me less afraid of **phpunit**.

**Description**
In addition, the package does not require much user-specific configuration, which can be confusing, although it can be customized if desired.
sample

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "nexusphp/tachycardia": "^1.0",
         "phpstan/phpstan": "^1.5.6",
         "phpunit/phpunit": "^9.1",
+        "codedungeon/phpunit-result-printer": "^0.32.0",
         "predis/predis": "^1.1",
         "rector/rector": "0.12.22"
     },
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,7 +9,8 @@
          colors="true"
          failOnRisky="true"
          failOnWarning="true"
-         verbose="true">
+         verbose="true"
+         printerClass="Codedungeon\PHPUnitPrettyResultPrinter\Printer">
 
     <coverage processUncoveredFiles="true" ignoreDeprecatedCodeUnits="true">
         <include>


kenjis:
This output might be easier to see and understand a bit.
I don't oppose this.
If devs don't want this output, they can remove easily with `phpunit.xml`.

```
bash-3.2$ ./phpunit --exclude-group DatabaseLive,CacheLive -c phpunit.xml.dist

PHPUnit Pretty Result Printer 0.32.0 by Codedungeon and contributors.
==> Configuration: ~/work/codeigniter/CodeIgniter4/vendor/codedungeon/phpunit-result-printer/src/phpunit-printer.yml

PHPUnit 9.5.20 #StandWithUkraine

Runtime:       PHP 8.1.5 with Xdebug 3.1.1
Configuration: phpunit.xml.dist


 ==> ComposerJsonTest           ✔  ✔  
 ==> ResponseTraitTest          ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  
 ==> AutoloaderTest             ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  
 ==> FileLocatorTest            ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  
 ==> CLITest                    ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  ✔  
                                ✔  ✔  ✔  
```

paulbalandan:
PHPUnit 10 won't be supporting custom printer classes.
```
- Removed the --printer option
- Removed the printerClass and printerFile attributes on the <phpunit> XML configuration element
```
https://github.com/sebastianbergmann/phpunit/blob/main/ChangeLog-10.0.md

kenjis:
The new Event System will be introduced.
https://github.com/sebastianbergmann/phpunit/issues/4676
I don't know this package will be upgraded, though.

datamweb:
> PHPUnit 10 won't be supporting custom printer classes.
> 
> ```
> - Removed the --printer option
> - Removed the printerClass and printerFile attributes on the <phpunit> XML configuration element
> ```
> 
> https://github.com/sebastianbergmann/phpunit/blob/main/ChangeLog-10.0.md

In these ChangeLog is **PHP 7.3, PHP 7.4, and PHP 8.0 are no longer supported**, is this true for **codeigniter4** in the future?

kenjis:
CI4 supports the Currently Supported PHP Versions.
https://www.php.net/supported-versions.php
So at least PHP 8.0 would be supported until 26 Nov 2023.

datamweb:
@kenjis Thanks , a new day, a new tip.

MGatner (commented):
@datamweb makes a good case for this, especially in an era where we are trying to be more "contributor friendly". I don't like the additional dependency which (as the discussion points out) makes the future of this integration uncertain. 

Alternative ideas: could we add notes to the docs on how to get this set up locally? Or include it in the CodeIgniter DevKit?

MGatner:
Just to call it out, here are the additional dev dependencies this would add:
```yml
    "hassankhan/config": "^0.11.2|^1.0|^2.0|^3.0",
    "symfony/yaml": "^2.7|^3.0|^4.0|^5.0",
    "codedungeon/php-cli-colors": "^1.10.2",
    "2bj/phanybar": "^1.0"
```

Also it looks like the library has some PHP 8.1 compatibility work (4 months ago) but the last release was 2 years ago; if we want to pursue this we should probably ask the maintainers to push through another release.

kenjis:
> Also it looks like the library has some PHP 8.1 compatibility work (4 months ago) but the last release was 2 years ago; if we want to pursue this we should probably ask the maintainers to push through another release.

Good point. Unfortunately, it doesn't seem to be maintained very properly.

datamweb:
@MGatner Thank you for your importance to PR. I also disagree with the ambiguous uncertain.
Adding `--testdox` to the documents seems to be useful. This is the least that can be done now.
It seems that having `vendor\bin\phpunit --testdox` can be helpful for beginners.

kenjis:
@datamweb I agree with you. Why don't you send a PR?

PR #5961, opened 2022-05-05
fix: route limit to subdomains does not work

### LINKED ISSUE #5959: Bug: Limit to subdomains, not work
### PHP Version

8.0

### CodeIgniter4 Version

4.1.9 develop

Manual (zip or tar.gz)

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

> https://codeigniter4.github.io/CodeIgniter4/incoming/routing.html#limit-to-subdomains
> You can restrict it to any subdomain by setting the value to an asterisk, (*). If you are viewing from a URL that does not have any subdomain present, this will not be matched:

// Limit to any sub-domain
`$routes->add('from', 'to', ['subdomain' => '*']);` <-not work

### Steps to Reproduce

Change in file: [https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Router/RouteCollection.php#L1354](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Router/RouteCollection.php#L1354)

change old: `unset($host[count($host)]);`
to new: `unset($host[count($host)-1]);`

### Expected Output

bug fix

### Anything else?

_No response_

kenjis:
**Description**
Fixes  #5959

```php
// Limit to any sub-domain
$routes->add('from', 'to', ['subdomain' => '*']); // not work
```

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1297,7 +1297,7 @@ class RouteCollection implements RouteCollectionInterface
      * Compares the subdomain(s) passed in against the current subdomain
      * on this page request.
      *
-     * @param mixed $subdomains
+     * @param string|string[] $subdomains
      */
     private function checkSubdomains($subdomains): bool
     {
@@ -1330,7 +1330,7 @@ class RouteCollection implements RouteCollectionInterface
      * It's especially not perfect since it's possible to register a domain
      * with a period (.) as part of the domain name.
      *
-     * @return mixed
+     * @return false|string the subdomain
      */
     private function determineCurrentSubdomain()
     {
@@ -1351,7 +1351,7 @@ class RouteCollection implements RouteCollectionInterface
         }
 
         // Get rid of any domains, which will be the last
-        unset($host[count($host)]);
+        unset($host[count($host) - 1]);
 
         // Account for .co.uk, .co.nz, etc. domains
         if (end($host) === 'co') {
--- a/tests/system/Router/RouteCollectionTest.php
+++ b/tests/system/Router/RouteCollectionTest.php
@@ -1157,6 +1157,20 @@ final class RouteCollectionTest extends CIUnitTestCase
     }
 
     /**
+     * @see https://github.com/codeigniter4/CodeIgniter4/issues/5959
+     */
+    public function testWithNoSubdomainAndDot()
+    {
+        $_SERVER['HTTP_HOST'] = 'example.com';
+
+        $routes = $this->getCollector();
+
+        $routes->add('/objects/(:alphanum)', 'App::objectsList/$1', ['subdomain' => '*']);
+
+        $this->assertSame([], $routes->getRoutes());
+    }
+
+    /**
      * @see https://github.com/codeigniter4/CodeIgniter4/issues/1692
      */
     public function testWithSubdomainOrdered()
--- a/tests/system/Router/RouteCollectionTest.php
+++ b/tests/system/Router/RouteCollectionTest.php
@@ -1128,7 +1128,7 @@ final class RouteCollectionTest extends CIUnitTestCase
     {
         $routes = $this->getCollector();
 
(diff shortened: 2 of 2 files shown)


iRedds:
Not sure if this will work correctly if the domain name is a first level domain. 
The first-level domain can be used in internal (corporate) networks.

kenjis:
@iRedds What do you mean by "first level domain"?
Example?

iRedds:
A local resource can have a ``mycompany`` domain.
Then the behavior will be incorrect for the ``api.mycompany`` subdomain.

kenjis:
It seems Brand top-level domains.
https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#Brand_top-level_domains

This method is not perfect from the beginning.
https://github.com/codeigniter4/CodeIgniter4/blob/38d5c109981bb51dc3ba2309ca6152c49fc07f28/system/Router/RouteCollection.php#L1327-L1328

iRedds:
Not perfect and not correct are two different cases.
I think subdomain parsing should be based on app.baseURL

szajens:
Tomorrow will show my solution, I have been using for several years. Because I'm leaving today

szajens:
`        /**
         * domainFilter()
         *
         *
         * @param mixed $allowed_domain - allowed domain format ['example.net','subdomain.example.com'] , don't use http://
         * @param mixed $domain_to_check - input domain to check e.g. www.www.www.www.a1.g.h.r.r.e.e.e.e.e.e.d.subdomain.example.com
         *
         * @return array associates
         *
         *               if not found:
         *    ['matches'] type (int) == 0
         *    [0] == empty array()
         *
         *               if checked:
         *
         *    ['matches'] type (int) == 1
         *    [0][0] => www. - prefix
         *    [0][1] => a1.g.h.r.r.e.e.e.e.e.e.d - subdomain or null if not check
         *    [0][2] => subdomain.example.com - domain   <br>
         *
         *
         *
         */

        function domainFilter(array $allowed_domain, string $domain_to_check):array {

            $result = [];//only for IDE (warning undefined variable)

            $allowed_domain = implode('|', $allowed_domain);

            $result['matches'] = preg_match('/^(?:(w{3})\.)*(?:((?:[a-z0-9\-\_]+\.)*[a-z0-9\-\_]+)\.)*('.$allowed_domain.')$/', $domain_to_check, $result[0]);

            //remove first element from array, but first array return input param $allowed_domain
            array_shift($result[0]);

            return $result;
        }



        d(
            domainFilter(['crazy.sub.domain.com', 'ci.loc'], 'www.usersubdomain.crazy.sub.domain.com')
        );

        d(
            domainFilter(['crazy.sub.domain.com', 'ci.loc'], 'aaa.bbb.usersubdomain.crazy.sub.domain.com')
        );`

domain-screenshottt

szajens:
[link to source domain filter function](https://github.com/szajens/myFunction/blob/main/domainFilter.php)

could be added Placeholders and Regular Expression to domain and subdomain

kenjis:
I believe this PR fixed the bug #5959. So this PR is done.

@iRedds The Bug with Brand top-level domains is another bug that exists already.

@szajens Thank you for sample code. Adding like that code is an enhancement. It is better to create another PR.
From the beginning, using `$_SERVER['HTTP_HOST']` without validation is kind of vulnerability.

I think it is better to add hostname validation and logic that is more reliable to determine subdomains.
Contributions are welcome.

szajens:
That's right, because someone can dynamically replace the host. 

Powyższy kod eliminuje błąd związany z domenami najwyższego poziomu marki

szajens:
@kenjis maybe you will create new PR,because my English no perfect 😁

MGatner (approved):

MGatner:
Open to other enhancements and bugfixes, but I'm with kenjis that this addresses the issue at hand.

PR #5962, opened 2022-05-05
refactor: remove `$_SERVER['HTTP_HOST']` in RouteCollection

kenjis:
**Description**
Depending on superglobals are bad practice.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner (commented):
I like this change but that is a *lot* of adjustments to tests to accommodate it. Is this going to break people's tests if they are using anything HTTP-specific?

kenjis:
> Is this going to break people's tests if they are using anything HTTP-specific?

Maybe yes. This PR changes the point that fetches the `$_SERVER['HTTP_HOST']` value.
The RouteCollection's constructor fixes the value for the class.

Tests that sets `$_SERVER['HTTP_HOST']` value after instantiation of RouteCollection
would break.

The current CI4 code has many property or variable values that can be changed later. 
Setters are in many classes, and some classes depend on superglobals like this.
It makes harder to read and use the code because I don't know when the correct value will be set,
when the object state is not invalid.

I want more classes that have complete constructor, and want more immutable classes.
But changes for them will break existing tests...

MGatner (approved):
I went and looked through my packages that I thought might set this directly, but none of them so. I think this is highly unlikely to break anything, and it brings good benefits and actually better testability.

kenjis:
Good to hear!
Probably there are few people that test `RouteCollection` and `$_SERVER['HTTP_HOST']` directly.
And changing supergloabls after instantiating objects to test is wrong wary to write test code.

PR #5963, opened 2022-05-05
refactor: add strtolower() to Request::getMethod()

kenjis:
**Description**
Supersedes #5766

The `$upper` functionality will be removed.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Views/errors/html/error_exception.php
+++ b/app/Views/errors/html/error_exception.php
@@ -199,7 +199,7 @@
 						</tr>
 						<tr>
 							<td>HTTP Method</td>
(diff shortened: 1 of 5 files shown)


paulbalandan (approved):

iRedds:
Why did you choose the strtolower function instead of the strtoupper function?
As far as I know, 'REQUEST_METHOD' returns a value in upper case.
The strtoupper function can be used in fewer cases, only where we do not control the value of the method.

kenjis:
I chose strtolower() or strtoupper() to fit the current implementation.

PR #5964, opened 2022-05-05
chore: update deptrac config file

kenjis:
**Description**
- update deptrac config file

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


MGatner:
That's odd; looks like maybe they changed the config file name in a minor release? I will ask them about it, because we will need to update DevKit and website and other paces as well. This looks good though.

MGatner (approved):

PR #5965, opened 2022-05-05
fix: Images\Handlers\GDHandler Implicit conversion from float to int loses precision

### LINKED ISSUE #5941: Bug: PHP 8.1 Images\Handlers\GDHandler Implicit conversion from float to int loses precision
See https://forum.codeigniter.com/showthread.php?tid=81803

```
ErrorException

Implicit conversion from float 63.5 to int loses precision [search →](https://www.duckduckgo.com/?q=ErrorException+Implicit+conversion+from+float+63.5+to+int+loses+precision)

SYSTEMPATH/Images/Handlers/GDHandler.php at line 467
```

```php
<?php
namespace App\Controllers;
{
    public function index()
    {
        \Config\Services::image('gd')
            ->withFile(WRITEPATH . 'img.png')
            ->text('Copyright 2017 My Photo Co', [
                'color'      => '#fff',
                'opacity'    => 0.5,
                'withShadow' => true,
                'hAlign'    => 'center',
                'vAlign'    => 'bottom',
                'fontSize'  => 20
            ])
            ->save(WRITEPATH . 'img2.png');
    }
}
```

kenjis:
**Description**
Fixes #5941

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Images/Handlers/GDHandler.php
+++ b/system/Images/Handlers/GDHandler.php
@@ -451,7 +451,7 @@ class GDHandler extends BaseHandler
          * Get the rest of the string and split it into 2-length
          * hex values:
          */
-        $opacity = ($options['opacity'] * 127);
+        $opacity = (int) ($options['opacity'] * 127);
 
         // Allow opacity to be applied to the text
         imagealphablending($src, true);
--- a/tests/system/Images/GDHandlerTest.php
+++ b/tests/system/Images/GDHandlerTest.php
@@ -289,7 +289,10 @@ final class GDHandlerTest extends CIUnitTestCase
     public function testText()
     {
         $this->handler->withFile($this->path);
-        $this->handler->text('vertical', ['hAlign' => 'right', 'vAlign' => 'bottom']);
+        $this->handler->text(
+            'vertical',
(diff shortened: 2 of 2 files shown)


system/Images/Handlers/GDHandler.php
```diff
@@ -451,7 +451,7 @@ protected function textOverlay(string $text, array $options = [], bool $isShadow
          * Get the rest of the string and split it into 2-length
          * hex values:
          */
-        $opacity = ($options['opacity'] * 127);
+        $opacity = (int) ($options['opacity'] * 127);
```
paulbalandan, line 454: Casting by `(int)` removes the decimals part. https://3v4l.org/8lXKD
Is that the intended behavior or should we use `round()`?

paulbalandan: Looking at my comment in https://github.com/codeigniter4/CodeIgniter4/pull/5370#issuecomment-976165819, it seems this is to match the current behavior.

kenjis: Frankly, I don't think there is much difference either way.
And yes, this is to match the current behavior.

paulbalandan (approved):

PR #5970, opened 2022-05-06
fix: Model::_call() static analysis

### LINKED ISSUE #4509: Bug: Model::_call() static analysis
**Describe the bug**
With the removal of `@mixin Model` from `BaseBuilder`, magic calls to Model that match a `BaseBuilder` method will now "return" that method's type, causing subsequent calls too create static analysis issues. For example:
```php
$users = model(UserModel::class)->where('name', 'Bob');
```
`$users` is now a `Model` with its `$builder` property primed for the criterium, but because the `@mixin BaseBuilder` matched `BaseBuild::where(): self` PHPStan thinks `$users` is `BaseBuilder`. So calling a model method after this:
```php
$user = $users->first();
```
... will cause static analysis to complain:

> Call to an undefined method CodeIgniter\Database\BaseBuilder::first()

I have tried a few things, like adding `@method` to Model, but thus far I don't see a good solution.

**CodeIgniter 4 version**
`develop` latest

**Affected module(s)**
`Model`, `BaseBuilder`

**Context**
 - OS: Ubuntu
 - Web server: Apache
 - PHP version: 7.4

kenjis:
**Description**
Fixes #4509

- remove `@mixin BaseBuilder`, add `@method`

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -666,11 +666,6 @@ parameters:
 			path: system/Log/Logger.php
 
 		-
-			message: "#^Call to an undefined method CodeIgniter\\\\Database\\\\BaseBuilder\\:\\:asArray\\(\\)\\.$#"
-			count: 1
-			path: system/Model.php
-
-		-
 			message: "#^Property CodeIgniter\\\\RESTful\\\\ResourceController\\:\\:\\$formatter \\(CodeIgniter\\\\Format\\\\FormatterInterface\\) in isset\\(\\) is not nullable\\.$#"
 			count: 1
 			path: system/RESTful/ResourceController.php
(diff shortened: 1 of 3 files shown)


MGatner (approved):
I'm sad that dynamic extensions didn't work out (but grateful you tried it). This looks like an improvement, and it already allowed SA to detect a mistake!

kenjis:
It might be my misunderstanding that the dynamic extensions didn't work out.
But the true problem is not sure what methods really can be used.
I mean, it was later discovered that `getCompiled*()` could not be used after all.
The current implementation can call all Builder methods, but not sure if that will actually work well.

It would be better to add really usable methods to `@method`.

PR #5972, opened 2022-05-07
refactor: `system/bootstrap.php` only loads files and registers autoloader

kenjis:
**Description**
When we want to use [Preloading](https://www.php.net/manual/en/opcache.preloading.php), we need CI4 Autoloader.
But it is not easy to make the Autoloader work fine.
If you use this `system/bootstrap.php`, you can make it a bit easily.

- move `CodeIgniter` instantiation and `DotEnv` loading from `system/bootstrap.php`
- tweak `index.php` and `spark.php`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/changelogs/v4.2.0.rst
+++ b/user_guide_src/source/changelogs/v4.2.0.rst
@@ -16,6 +16,7 @@ BREAKING
 - The ``CodeIgniter\CodeIgniter`` class has a new property ``$context`` and it must have the correct context at runtime. So the following files have been changed:
     - ``public/index.php``
     - ``spark``
+- The ``system/bootstrap.php`` no longer returns a ``CodeIgniter`` instance, and does not load ``.env`` file.
 - The method signature of ``CodeIgniter\CLI\CommandRunner::_remap()`` has been changed to fix a bug.
 - The ``CodeIgniter\Autoloader\Autoloader::initialize()`` has changed the behavior to fix a bug. It used to use Composer classmap only when ``$modules->discoverInComposer`` is true. Now it always uses the Composer classmap if Composer is available.
 - The color code output by :ref:`CLI::color() <cli-library-color>` has been changed to fix a bug.
--- a/user_guide_src/source/changelogs/v4.2.0.rst
+++ b/user_guide_src/source/changelogs/v4.2.0.rst
@@ -16,7 +16,7 @@ BREAKING
 - The ``CodeIgniter\CodeIgniter`` class has a new property ``$context`` and it must have the correct context at runtime. So the following files have been changed:
     - ``public/index.php``
     - ``spark``
-- The ``system/bootstrap.php`` no longer returns a ``CodeIgniter`` instance, and does not load ``.env`` file.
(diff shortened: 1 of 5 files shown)


MGatner (commented):
I will have to look at this o desktop, it's a little hard on mobile to tell what moved where. 

I'm not sure the answer to this but... is changing the return of a file (**bootstrap.php**) considered a breaking change?

kenjis:
> I'm not sure the answer to this but... is changing the return of a file (bootstrap.php) considered a breaking change?

Yes. I added the changelog and upgrade guide.

user_guide_src/source/installation/upgrade_420.rst
```diff
@@ -24,7 +24,7 @@ The following files received significant changes and
 Breaking Changes
 ****************
 
-
+- The ``system/bootstrap.php`` file no longer returns a ``CodeIgniter`` instance, and does not load ``.env`` file. The ``index.php`` and ``spark`` now do them. If you have code that expects these behaviors, it will no longer work and must be modified.
```
MGatner, line 27:
```suggestion
- The ``system/bootstrap.php`` file no longer returns a ``CodeIgniter`` instance, and does not load the ``.env`` file (now handled in ``index.php`` and ``spark``). If you have code that expects these behaviors it will no longer work and must be modified.
```

MGatner (approved):
I understand all the moves now. I don't think the **bootstrap.php** return is an issue - I'm actually glad we won't be relying on that anymore. 

One phrasing suggestion, and let's make sure @lonnieezell has seen this one.

user_guide_src/source/installation/upgrade_420.rst
```diff
@@ -24,7 +24,7 @@ The following files received significant changes and
 Breaking Changes
 ****************
 
-
+- The ``system/bootstrap.php`` file no longer returns a ``CodeIgniter`` instance, and does not load the ``.env`` file (now handled in ``index.php`` and ``spark``). If you have code that expects these behaviors it will no longer work and must be modified.
```
lonnieezell, line 27: Please mention _why_ this change was made as well.

lonnieezell (approved):
Since it already had a couple of other approvals I didn't scan every line. I have no problems with the changes, but I do think we should let people know _why_ the changes are happening. My first thought on  reading the title of the issue was, "Why are we changing? Is there any benefit to possibly breaking people's apps for this or is it just to be "proper"?" Turns out there is a very good reason, but let's help people get to that conclusion before they get upset lol.

user_guide_src/source/changelogs/v4.2.0.rst
```diff
@@ -16,6 +16,7 @@ BREAKING
 - The ``CodeIgniter\CodeIgniter`` class has a new property ``$context`` and it must have the correct context at runtime. So the following files have been changed:
     - ``public/index.php``
     - ``spark``
+- The ``system/bootstrap.php`` no longer returns a ``CodeIgniter`` instance, and does not load ``.env`` file.
```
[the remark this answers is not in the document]
kenjis: Done.

user_guide_src/source/installation/upgrade_420.rst
kenjis: Done.

paulbalandan (approved):

PR #5975, opened 2022-05-08
Docs: add part result of the PHP Unit Test

datamweb:
PR is related to https://github.com/codeigniter4/CodeIgniter4/pull/5950#issuecomment-1120319983 .
Provides a brief explanation of the result **after run the test**.
I hope it is useful for beginners.

**Description**
To continue **clear documentation**.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -57,6 +57,25 @@ You can run the tests without running the live database and the live cache tests
 
 	> ./phpunit --exclude-group DatabaseLive,CacheLive
 
+## Result Of The Unit Test
+
+After running the test. For each test run, the PHPUnit command-line tool prints one character to indicate progress:
+
+1. `.` Printed when the test succeeds.
+2. `F` Printed when an assertion fails while running the test method.
+3. `E` Printed when an error occurs while running the test method.
+4. `R` Printed when the test has been marked as risky (see [Risky Tests](https://phpunit.readthedocs.io/en/9.5/risky-tests.html#risky-tests)).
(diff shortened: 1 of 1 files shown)


paulbalandan (commented):
This is just me thinking aloud.

I know we're in the stage of trying to pull in more contributions from the community and that's great. But what this PR is trying to add is quite too much to be honest. Why is it our responsibility to discuss in the docs what our tools do? Shouldn't it be the tools' responsibilities? What I mean is that when I made my first PR here I didn't know about PHPUnit etc. then **and I did not expect** CI4's docs to be discussing those. How I learned is I read those tools' docs, not CI4's docs.

Honestly, I do not expect a framework to be documenting things other than framework stuff. It is like doing things outside your job description. I'm -1 to this.

kenjis:
@paulbalandan I'm wondering why you do react like that on this.
Can you tell us what's wrong with this for you?

This PR just adds one section that explains the PHPUnit output.
It is not so long.

We already explains how to run phpunit command in the README:
https://github.com/codeigniter4/CodeIgniter4/tree/develop/tests#running-the-tests
It is also common knowledge to PHPUnit users.

> Shouldn't it be the tools' responsibilities?

Yes, and they have good documentation.

> What I mean is that when I made my first PR here I didn't know about PHPUnit etc. then and I did not expect CI4's docs to be discussing those. How I learned is I read those tools' docs, not CI4's docs.

It is very good way. The issue here is how to encourage it.
Do you think too much explanations in CI4's docs will prevent it?

PR #5987, opened 2022-05-10
change dies() to php's function name die()

sba:
dies() may be linguistically more correct, but I would write die() here


kenjis (approved):

kenjis:
It indicates PHP code (the language construct), so `die()` is correct.
Thank you.

PR #5999, opened 2022-05-17
Changing the subquery builder for the Oracle

iRedds:
**Description**
Because Oracle does not support using the ``AS`` keyword for table aliases. 
This PR removes the ``AS`` keyword when building subqueries.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2846,7 +2846,7 @@ class BaseBuilder
             $alias    = trim($alias);
 
             if ($alias !== '') {
-                $subquery .= ' AS ' . ($this->db->protectIdentifiers ? $this->db->escapeIdentifiers($alias) : $alias);
+                $subquery .= ' ' . ($this->db->protectIdentifiers ? $this->db->escapeIdentifiers($alias) : $alias);
             }
         }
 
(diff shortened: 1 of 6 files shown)


paulbalandan (commented):
I think this should be in the changelog as people might be expecting in their tests the `AS` in the SQL.

kenjis:
Subquery has not been released. Do we need it to be written in the changelog?
https://codeigniter4.github.io/CodeIgniter4/changelogs/v4.2.0.html#others

iRedds:
@paulbalandan An improvement to the BaseBuilder::buildSubquery() method, namely adding an alias, has not yet been released.
But I added a description of this improvement to the changelog. https://github.com/codeigniter4/CodeIgniter4/pull/6006

If this PR and the fix in the changelog are merged before the release, then it seems to me that it makes no sense to further clarify the changes from this PR.

paulbalandan:
Ohh. I thought this was documented already. Thanks for the clarification.

samsonasik (approved):

MGatner (approved):
I wasn't familiar with this syntax but tests pass and everyone else likes it so it's good by me. Thanks again @iRedds

PR #6004, opened 2022-05-18
Router class optimization.

iRedds:
**Description**
Duplicate code moved to a separate method.
Removed code that had no effect on the work.
Optimization.

**Checklist:**
- [x] Securely signed commits
- [x] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -457,54 +457,31 @@ class Router implements RouterInterface
 
                     $this->params = $matches;
 
-                    $this->matchedRoute = [
-                        $matchedKey,
-                        $handler,
-                    ];
-
-                    $this->matchedRouteOptions = $this->collection->getRoutesOptions($matchedKey);
+                    $this->setMatchedRoute($matchedKey, $handler);
 
                     return true;
                 }
 
-                if (strpos($handler, '$') !== false && strpos($routeKey, '(') !== false) {
-                    // Using back-references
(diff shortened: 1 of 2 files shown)


kenjis (approved):

system/Router/Router.php
```diff
@@ -672,4 +649,18 @@ protected function setDefaultController()
...
+     *
+     * @return static
+     */
+    protected function setMatchedRoute(string $route, $handler)
+    {
+        $this->matchedRoute = [$route, $handler];
+
+        $this->matchedRouteOptions = $this->collection->getRoutesOptions($route);
+
+        return $this;
```
samsonasik, line 664: I think return $this is not needed if we are not going to use it as fluent

system/Router/Router.php
```diff
@@ -672,4 +649,16 @@ protected function setDefaultController()
 
         log_message('info', 'Used the default controller.');
     }
+
+    /**
+     * @param callable|string $handler
+     *
+     * @return void
+     */
+    protected function setMatchedRoute(string $route, $handler)
```
samsonasik, line 658: `: void` return can be added since it is new method.

samsonasik (approved):

system/Router/Router.php
iRedds: Thanks

samsonasik:
Thank you @iRedds

PR #6010, opened 2022-05-19
chore: fix indent in views

kenjis:
**Description**
Although not explicitly stated, it appears that view file indentation is basically done with tabs.
This PR adds `.editorconfig` for Views and fixes indentation in some files.

It seems two space size indentation is common for HTML/CSS, so I set `indent_size = 2`.

The only one exception is `system/Debug/Toolbar/Views/toolbar.css`.
It is generated by scss and it uses space 2 indentation.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Views/.editorconfig
+++ b/app/Views/.editorconfig
@@ -1,0 +1,6 @@
+; This file is for unifying the coding style for different editors and IDEs.
+; More information at http://editorconfig.org
+
+[*]
+indent_style = tab
+indent_size = 2
--- a/app/Views/errors/cli/.editorconfig
+++ b/app/Views/errors/cli/.editorconfig
@@ -1,13 +1,6 @@
 ; This file is for unifying the coding style for different editors and IDEs.
 ; More information at http://editorconfig.org
 
-; top-most EditorConfig file
-root = true
-
 [*]
-charset = utf-8
 indent_style = space
 indent_size = 4
-end_of_line = lf
-insert_final_newline = true
-trim_trailing_whitespace = true
--- a/system/Debug/Toolbar/Views/.editorconfig
+++ b/system/Debug/Toolbar/Views/.editorconfig
@@ -1,0 +1,6 @@
+; This file is for unifying the coding style for different editors and IDEs.
+; More information at http://editorconfig.org
+
+[*]
+indent_style = tab
+indent_size = 2
--- a/system/Debug/Toolbar/Views/_database.tpl
+++ b/system/Debug/Toolbar/Views/_database.tpl
@@ -1,26 +1,26 @@
 <table>
-    <thead>
-        <tr>
-            <th class="debug-bar-width6r">Time</th>
-            <th>Query String</th>
-        </tr>
-    </thead>
-    <tbody>
-    {queries}
-        <tr class="{class}" title="{hover}" data-toggle="{qid}-trace">
-            <td class="narrow">{duration}</td>
-            <td>{! sql !}</td>
-            <td style="text-align: right"><strong>{trace-file}</strong></td>
-        </tr>
-        <tr class="muted" id="{qid}-trace" style="display:none">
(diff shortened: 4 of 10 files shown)


MGatner:
I really don't have a preference here. Maybe see what @paulbalandan thinks from a CS Fixer perspective?

lonnieezell:
I don't have a strong opinion and will go with the general consensus, but my personal thoughts are why make it more complex than needed? We chose to use PSR12, so maybe just keep everything 4 spaces for consistency?

paulbalandan (approved):
I have always been the spaces guy for all file types as it looks good on local and on Github. But, if `tabs` is the convention used for some files here then let's go with it.

lonnieezell:
I guess - where do you see that this is a common practice? I personally haven't run across that and this feels like opening another "tabs vs spaces" war where there isn't a real winner, just personal choice. I've always been a tabs guy in the past, but find I really don't care anymore in that war, but why make things more difficult for ourselves and potential contributors?  Unless everyone's IDEs default to this?

sfadschm:
I am kind of confused.
Does this PR propose using tabs in the view files or not (can not really distinguish on mobile)?

While this comes down to personal preference of course, I think I would like to have the same convention for all files that belong to the framework, regardless of their extension.

kenjis:
> Does this PR propose using tabs in the view files or not (can not really distinguish on mobile)?

Yes, and tabs are mainly used now.
(No problem. If I see on PC, I can not distinguish!)

kenjis:
> where do you see that this is a common practice?

Probably 2 space size indentation came from Google guideline.
In fact, it is by spaces, not by tabs.

> Indent by 2 spaces at a time.
https://google.github.io/styleguide/htmlcssguide.html#Indentation

> For readability, add two spaces of indentation. Do not use the tab key.
https://www.w3schools.com/html/html5_syntax.asp

kenjis:
Facts:
- now we use tabs and spaces for indentation of view (HTML/CSS/JS) files.
- we don't have proper `.editorconfig` for view files.

> feels like opening another "tabs vs spaces" war where there isn't a real winner, just personal choice.

I don't like the word war, but unfortunately already someone is using tabs and another is using space for view files.
I think it would be better to at least be consistent.

If we agree with 4 space, I am okay.

lonnieezell:
Obviously, I'm a fan of the consistent approach so I vote for 4 space in views, also.

MGatner:
Four spaces works for me. @paulbalandan can we just lift the CS Fixer exemption for **Views** and have it apply the changes? Or is that likely to have more side effects?

sfadschm:
This will have side effects and would need a different config file, like we did for the examples in the docs.

kenjis:
@paulbalandan Your opinion?

paulbalandan:
I think I skipped **Views** before because of some other rules affecting the files, not by the `indentation` rule.

Looks like majority wants 4 spaces for the view files, and that's okay with me. But I think it should be only for the PHP files. For the other files like JS or HTML, I think we should use what's the convention for those files.

paulbalandan:
It may be overkill but maybe we can setup GHA to run formatters on those other files, like ESLint or Prettier.

kenjis:
No one strongly objected to the 4 spaces, I created a new PR: #6039 

> It may be overkill but maybe we can setup GHA to run formatters on those other files, like ESLint or Prettier.

View files are changed rarely, so I want to see the new editorconfig works fine or not.
We are heavy user of GHA, so I don't want to add loads by adding checks that do not help a lot.

paulbalandan:
> View files are changed rarely, so I want to see the new editorconfig works fine or not. We are heavy user of GHA, so I don't want to add loads by adding checks that do not help a lot.

Fair point. Yeah, that was overkill. 😂

PR #6012, opened 2022-05-21
fix: Ignore non-HTML responses in storePreviousURL

tearoom6:
**Description**

In this PR, I added an exclusion condition for `storePreviousURL`: "Response Content-Type is not text/html".

**Background**

In our application, some images are composited in the CodeIgnitor controller and returned.

The request URL for such case is saved as `_ci_previous_url` because it does not meet any of the exclusion conditions of the existing `storePreviousURL` implementation, and when redirecting back, etc., there is a case where the request is redirected to the URL of the image.

The URL to store as `previous_url` is usually considered to be an HTML (`text/html`) case, so I have added that as a condition.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/changelogs/v4.2.0.rst
+++ b/user_guide_src/source/changelogs/v4.2.0.rst
@@ -32,6 +32,7 @@ Behavior Changes
 - To prevent unexpected access from the web browser, if a controller is added to a cli route (``$routes->cli()``), all methods of that controller are no longer accessible via auto-routing.
 - There is a possible backward compatibility break for those users extending the History Collector and they should probably update ``History::setFiles()`` method.
 - The :php:func:`dot_array_search`'s unexpected behavior has been fixed. Now ``dot_array_search('foo.bar.baz', ['foo' => ['bar' => 23]])`` returns ``null``. The previous versions returned ``23``.
+- The ``CodeIgniter::storePreviousURL()`` has been changed to store only the URLs whose Content-Type was ``text/html``. It also affects the behavior of ``previous_url()`` and ``redirect()->back()``.
 
 Enhancements
 ************
--- a/user_guide_src/source/changelogs/v4.2.0.rst
+++ b/user_guide_src/source/changelogs/v4.2.0.rst
@@ -32,7 +32,7 @@ Behavior Changes
 - To prevent unexpected access from the web browser, if a controller is added to a cli route (``$routes->cli()``), all methods of that controller are no longer accessible via auto-routing.
 - There is a possible backward compatibility break for those users extending the History Collector and they should probably update ``History::setFiles()`` method.
 - The :php:func:`dot_array_search`'s unexpected behavior has been fixed. Now ``dot_array_search('foo.bar.baz', ['foo' => ['bar' => 23]])`` returns ``null``. The previous versions returned ``23``.
-- The ``CodeIgniter::storePreviousURL()`` has been changed to store only the URLs whose Content-Type was ``text/html``. It also affects the behavior of ``previous_url()`` and ``redirect()->back()``.
+- The ``CodeIgniter::storePreviousURL()`` has been changed to store only the URLs whose Content-Type was ``text/html``. It also affects the behavior of :php:func:`previous_url` and :php:func:`redirect()->back() <redirect>`.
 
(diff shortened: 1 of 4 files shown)


MGatner (changes requested):
I was worried this would be too restrictive because there might be some served content of a variant type. A little research suggests that the content type *should* actually be this restrictive:
> `text/html`: All HTML content should be served with this type.

I would like to see a tests added to confirm & exempt this - it should be fairly easy, copying what we already have.

kenjis:
This fix seems to be reasonable. The image URL is not a page.

> previous_url([$returnObject = false])
Returns the full URL (including segments) of the page the user was previously on.
https://codeigniter4.github.io/CodeIgniter4/helpers/url_helper.html

But this changes the behavior of `storePreviousURL()` and `previous_url()`, so 
it is better to add at least the changelog.

https://codeigniter4.github.io/CodeIgniter4/changelogs/v4.2.0.html#breaking

kenjis:
@tearoom6 
Thank you for sending this PR.

The user guide and unit testing are missing. 
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#unit-testing for unit testing.
Could you add them?

tearoom6:
@MGatner @kenjis
Thank you for your feedbacks! I have added the unit test case and updated the documentation.
In addition, the original commit was not signed, so I also added signature to the commit. (forced push was needed to do so.)

I would appreciate it if you could review it again.

user_guide_src/source/changelogs/v4.2.0.rst
```diff
@@ -23,6 +23,7 @@ BREAKING
 - To prevent unexpected access from the web browser, if a controller is added to a cli route (``$routes->cli()``), all methods of that controller are no longer accessible via auto-routing.
 - There is a possible backward compatibility break for those users extending the History Collector and they should probably update ``History::setFiles()`` method.
 - The :php:func:`dot_array_search`'s unexpected behavior has been fixed. Now ``dot_array_search('foo.bar.baz', ['foo' => ['bar' => 23]])`` returns ``null``. The previous versions returned ``23``.
+- The ``CodeIgniter::storePreviousURL()`` has been changed to store only the URLs whose Content-Type was ``text/html``. It also affects the behavior of ``previous_url()`` and ``redirect()->back()``.
```
kenjis, line 26: Adding links is helpful.

```suggestion
- The ``CodeIgniter::storePreviousURL()`` has been changed to store only the URLs whose Content-Type was ``text/html``. It also affects the behavior of :php:func:`previous_url` and :php:func:`redirect()->back() <redirect>`.
```

user_guide_src/source/installation/upgrade_420.rst
```diff
@@ -33,6 +33,7 @@ Breaking Changes
 ****************
 
 - The ``system/bootstrap.php`` file no longer returns a ``CodeIgniter`` instance, and does not load the ``.env`` file (now handled in ``index.php`` and ``spark``). If you have code that expects these behaviors it will no longer work and must be modified. This has been changed to make `Preloading <https://www.php.net/manual/en/opcache.preloading.php>`_ easier to implement.
+- ``previous_url()`` has been changed to return only the URLs whose Content-Type was ``text/html``. Accordingly, the behavior of ``redirect()->back()`` has been changed.
```
kenjis, line 36: In the upgrade note, we write *what to do* when upgrading.
Since it seems unlikely that any developers were using non-HTML URLs, is there no need to list it?

tearoom6: I see, I understood. I have removed it from the update notes 👍

kenjis:
@tearoom6 Thank you for updating.

The PHPStan errors are not related to this PR. 

But these errors was fixed in the latest `develop`.
So if you rebase this branch, all GitHub Action checks would pass.

There is no conflict, so we can merge this PR if you don't rebase.

kenjis (approved):
LGTM!

tearoom6:
Thank you for reviewing and the advices!

MGatner (approved):
Very nice work! Thank you for your contribution.

PR #6015, opened 2022-05-22
Feature. QueryBuilder. Query union.

iRedds:
**Description**
This is the second attempt to add a query union to QueryBuilder.
Ref #4291

The implementation provides two methods:
-  `BaseBuilder::union(BaseBuilder|Closure $union)`
-  `BaseBuilder::unionAll(BaseBuilder|Closure $union)`

The union() method can be called in any order relative to the main query, but the union query will always be appended to the end. That is, the ``LIMIT`` or ``ORDER BY`` clauses will be relative to the main query.
```php
$builder->union($union)->limit(1)->orderBy('id');
// SELECT * FROM table LIMIT 1 ORDER BY id UNION query
```

Since DBMSs (MSSQL and Oracle) are demanding on queries using ``LIMIT`` and ``ORDER BY``, when generating SQL, all queries are wrapped in a ``SELECT * FROM(....) alias`` query.
The alias uwrp0 is used for the main query. Each subsequent query will have an alias with index +1.
```php
$builder->union($union)->limit(1)->orderBy('id')->unionAll($union2);
// SELECT * (SELECT * FROM table LIMIT 1 ORDER BY id) uwrp0 
// UNION SELECT * FROM (query) uwrp1 
// UNION ALL SELECT * FROM (query2) uwrp2
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -110,6 +110,13 @@ class BaseBuilder
     public $QBOrderBy = [];
 
     /**
+     * QB UNION data
+     *
+     * @var array<string>
+     */
+    protected array $QBUnion = [];
+
+    /**
      * QB NO ESCAPE data
      *
      * @var array
@@ -1136,6 +1143,48 @@ class BaseBuilder
         }
 
         return "{$prefix} {$column} {$not} LIKE :{$bind}:";
+    }
+
+    /**
+     * Add UNION statement
+     *
+     * @param BaseBuilder|Closure $union
+     *
+     * @return $this
+     */
+    public function union($union)
+    {
+        return $this->addUnionStatement($union);
+    }
+
+    /**
+     * Add UNION ALL statement
+     *
+     * @param BaseBuilder|Closure $union
+     *
+     * @return $this
+     */
+    public function unionAll($union)
+    {
+        return $this->addUnionStatement($union, true);
+    }
+
+    /**
+     * @used-by union()
+     * @used-by unionAll()
+     *
+     * @param BaseBuilder|Closure $union
+     *
+     * @return $this
+     */
+    protected function addUnionStatement($union, bool $all = false)
+    {
+        $this->QBUnion[] = "\n" . 'UNION '
+            . ($all ? 'ALL ' : '')
+            . 'SELECT * FROM '
+            . $this->buildSubquery($union, true, 'uwrp' . (count($this->QBUnion) + 1));
+
+        return $this;
     }
 
     /**
@@ -2427,10 +2476,10 @@ class BaseBuilder
             . $this->compileOrderBy();
 
         if ($this->QBLimit) {
-            return $this->_limit($sql . "\n");
+            $sql = $this->_limit($sql . "\n");
         }
 
-        return $sql;
+        return $this->unionInjection($sql);
     }
 
     /**
@@ -2585,6 +2634,17 @@ class BaseBuilder
         return '';
     }
 
+    protected function unionInjection(string $sql): string
+    {
+        if ($this->QBUnion === []) {
+            return $sql;
+        }
+
+        return 'SELECT * FROM (' . $sql . ') '
+            . ($this->db->protectIdentifiers ? $this->db->escapeIdentifiers('uwrp0') : 'uwrp0')
+            . implode("\n", $this->QBUnion);
+    }
+
     /**
      * Takes an object as input and converts the class variables to array key/vals
      *
@@ -2704,6 +2764,7 @@ class BaseBuilder
             'QBDistinct' => false,
             'QBLimit'    => false,
             'QBOffset'   => false,
+            'QBUnion'    => [],
         ]);
 
         if (! empty($this->db)) {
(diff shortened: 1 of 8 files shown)


iRedds:
I don't think replacing assertEquals() with assertSame() is the correct behavior for checking code style.

kenjis:
With this PR, I understand we can't write a query to use an `ORDER BY` or `LIMIT` clause to sort or limit the entire `UNION` result like the following. Am I correct?
```sql
(SELECT a FROM t1 WHERE a=10 AND B=1)
UNION
(SELECT a FROM t2 WHERE a=11 AND B=2)
ORDER BY a LIMIT 10;
```

iRedds:
Yes, such code is not possible by default. But it can be wrapped in another query using ``newQuery()->fromSubquery()``.
Like in a live test.
```php
$union   = $this->db->table('user')->limit(1)->orderBy('id', 'ASC');

$builder = $this->db->table('user')->union($union)->limit(1)->orderBy('id', 'DESC');

$result   = $this->db->newQuery()->fromSubquery($builder, 'q')->orderBy('id', 'DESC')->get();
```
```sql
SELECT * FROM (
    SELECT * FROM (SELECT * FROM user ORDER BY id DESC LIMIT 1)
    UNION
    SELECT * FROM (SELECT * FROM user ORDER BY id ASC LIMIT 1)
) q ORDER BY id
```

system/Database/BaseBuilder.php
```diff
@@ -1130,6 +1137,48 @@ protected function _like_statement(?string $prefix, string $column, ?string $not
...
+     * @param BaseBuilder|Closure $union
+     *
+     * @return $this
+     */
+    protected function addUnionStatement($union, bool $all = false)
+    {
+        $this->QBUnion[] = "\n" . 'UNION '
+            . ($all ? 'ALL ' : '')
+            . 'SELECT * FROM '
+            . $this->buildSubquery($union, true, 'uwrp' . (count($this->QBUnion) + 1));
```
kenjis, line 1177: What is `uwrp`?

iRedds: alias

kenjis: Why is it the name `uwrp`? 
What's the meaning?

iRedds: union wrapper =) 

If you can suggest another name for the alias, I'd be grateful.

kenjis: Aha! 
It is difficult to name it. I'm okay with uwrp.

kenjis:
I would like the test code to follow the Arrange Act Assert pattern as much as possible.
http://wiki.c2.com/?ArrangeActAssert

There are many test codes that do not follow it, though.

user_guide_src/source/changelogs/v4.2.0.rst
```diff
@@ -72,6 +72,7 @@ Database
 - QueryBuilder raw SQL string support
     - Added the class ``CodeIgniter\Database\RawSql`` which expresses raw SQL strings.
     - :ref:`select() <query-builder-select-rawsql>`, :ref:`where() <query-builder-where-rawsql>`, :ref:`like() <query-builder-like-rawsql>`, :ref:`join() <query-builder-join-rawsql>` accept the ``CodeIgniter\Database\RawSql`` instance.
+- QueryBuilder. Union queries.
```
kenjis, line 75: If there is a link to the detailed page, readers can go and see it easily.

kenjis:
@iRedds I thinks the query example to use an ORDER BY or LIMIT clause to sort or limit the entire UNION result should be documented.
And there are conflicts in the documentation.
Can you rebase this PR?

user_guide_src/source/changelogs/v4.2.0.rst
iRedds: added

iRedds:
@kenjis I've added an example to the documentation, but now @lonnieezell has nothing to add to the new version of the book )))

kenjis (approved):
Thank you!

iRedds:
@kenjis yes, i do

samsonasik (approved):

user_guide_src/source/database/query_builder.rst
```diff
@@ -679,6 +679,41 @@ As is in ``countAllResult()`` method, this method resets any field values that y
...
+Union queries
+*************
+
+Union
+=====
+
+$builder->union()
+-----------------
+
+Is used to combine the result-set of two or more SELECT statements.
```
michalsn, line 694:
```suggestion
Is used to combine the result-set of two or more SELECT statements. It will return only the unique results.
```

user_guide_src/source/database/query_builder.rst
```diff
@@ -679,6 +679,41 @@ As is in ``countAllResult()`` method, this method resets any field values that y
...
+In some cases, it may be necessary, for example, to sort or limit the number of records of the query result.
+The solution is to use the wrapper created via ``$db->newQuery()``.
+In the example below, we get the first 5 users + the last 5 users and sort the result by id:
+
+.. literalinclude:: query_builder/104.php
+
+$builder->unionAll()
+--------------------
+
+The behavior is the same as the ``union()`` method.
```
michalsn, line 715:
```suggestion
The behavior is the same as the ``union()`` method. However, all results will be returned, not just the unique ones.
```

Since there is a difference between `UNION` and `UNION ALL`. I feel like we should either explain a difference in one sentence or change the description since it might be misleading to users who are not entirely familiar with these methods.

michalsn (approved):
Excellent addition.

There are small additions to the docs I would like to see, though.

iRedds:
Thanks to everyone who helped along this long journey.

PR #6020, opened 2022-05-23
fix: change BaseService::reset() $initAutoloader to true by default

kenjis:
**Description**
`BaseService::reset(false)` makes the autoloader's namespaces empty.
Such use cases are rare and will prevent `lang()` from working. See https://github.com/codeigniter4/shield/issues/88#issuecomment-1134594714

- change `BaseService::reset()` default parameter value to `true`
- change `CIUnitTestCase::resetServices()` default parameter value to `true`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -273,7 +273,7 @@ class BaseService
     /**
      * Reset shared instances and mocks for testing.
      */
-    public static function reset(bool $initAutoloader = false)
+    public static function reset(bool $initAutoloader = true)
     {
         static::$mocks     = [];
         static::$instances = [];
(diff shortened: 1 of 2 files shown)


MGatner (commented):
Changing the parameter default and behavior in a minor release worries me a little bit. I do think this is the desired behavior and is a little bit "hidden". Let's see what others  think, but a compromise would be to change the default on CIUnitTestCase but not Services yet.

kenjis:
Basically `Services::reset()` is also only for testing.
It is not used in normal apps.

Apps definitely need namespaces, so if it is used, it would surely be `Services::reset(true)`.
I know only one example:
https://github.com/SDPM-lab/Codeigniter4-Roadrunner/blob/ef6752648ee90e5228ae9ab04ace41fc50f7b5d0/src/Commands/file/psr-worker.php#L136

So I think this change will not affect existing apps.

MGatner (approved):
All good points, I'm with you now.

samsonasik (approved):

paulbalandan (approved):

PR #6021, opened 2022-05-24
docs: add note to $allowedFields

### LINKED ISSUE #5904: Bug: DataException "There is no data to update." doesn't fire if $primaryKey is present in $allowedFields
### PHP Version

8.0

### CodeIgniter4 Version

4.1.7

Manual (zip or tar.gz)

Linux

### Which server did you use?

cgi-fcgi

### Database

_No response_

### What happened?

Calling `BaseModel::update()` should throw `DataException` "There is no data to update." if nothing in database entry (row) has changed. It doesn't happen if model's `$allowedFields` contains `$primaryKey` (for whatever reason - developer mistake or some rare case).

Example:
```php
class FooModel extends Model
{
    protected $table = 'foo';
    protected $primaryKey = 'id';
    protected $returnType = '\App\Entities\Foo';
    protected $allowedFields = [ 'id', 'name' ];
}
```

### Steps to Reproduce

```php
$fooModel = new FooModel();

$foo = $fooModel->find(1);
$foo->fill(['unused' => 'foo' ]);
$fooModel->update($foo);
```

### Expected Output

DataException "There is no data to update."

### Anything else?

Pseudo-code explaining what actually happens:
```php
BaseModel::update()
    BaseModel::transformDataToArray()
        BaseModel::objectToArray()
            Model::objectToRawArray()
                BaseModel::objectToRawArray()
                    Entity::toRawArray()                                       ← This gets an array of changed only values
                $properties[$this->primaryKey] = $data->{$this->primaryKey}    ← This sets $primaryKey even if it wasn't set in $data passed to the BaseModel::update()
    BaseModel::doProtectFields()                                               ← This filters to-be-updated columns using $allowedFields
    if (empty($data))
        throw DataException::forEmptyDataset('update');
```

kenjis:
**Description**
Fixes #5904

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/models/model.rst
+++ b/user_guide_src/source/models/model.rst
@@ -141,7 +141,7 @@ This array should be updated with the field names that can be set during ``save(
 against just taking input from a form and throwing it all at the model, resulting in
 potential mass assignment vulnerabilities.
 
-.. note:: The ``$primaryKey`` should never be an allowed field.
+.. note:: The ``$primaryKey`` field should never be an allowed field.
 
 $useTimestamps
 --------------
--- a/user_guide_src/source/models/model.rst
+++ b/user_guide_src/source/models/model.rst
@@ -141,6 +141,8 @@ This array should be updated with the field names that can be set during ``save(
 against just taking input from a form and throwing it all at the model, resulting in
 potential mass assignment vulnerabilities.
 
+.. note:: The ``$primaryKey`` should never be an allowed field.
+
 $useTimestamps
 --------------
 


user_guide_src/source/models/model.rst
```diff
@@ -141,6 +141,8 @@ This array should be updated with the field names that can be set during ``save(
 against just taking input from a form and throwing it all at the model, resulting in
 potential mass assignment vulnerabilities.
 
+.. note:: The ``$primaryKey`` should never be an allowed field.
```
samsonasik, line 144: I think it should be primary key field as term, not variable with $ prefix

kenjis: I mean `$primaryKey` value.
`$primaryKey` is probably primary key field, but it can be other unique key.

samsonasik: how about change to:

```suggestion
.. note:: The ``primary key field`` should never be an allowed field.
```

kenjis: The text decoration ` `` ` is for source code (variables, method names, classnames).

How about this?

```suggestion
.. note:: The ``$primaryKey`` field should never be an allowed field.
```

samsonasik: Yes, that seems better 👍

samsonasik (approved):

paulbalandan (approved):

PR #6023, opened 2022-05-24
Fix cookie prefixing

### LINKED ISSUE #6009: Bug: `get_cookie()` cookie prefix behavior
### PHP Version

7.3

### CodeIgniter4 Version

4.1.9

macOS, Linux

### Which server did you use?

apache

### Database

MySQL 8

### What happened?

The cookie prefixing to prevent collisions doesn't exactly work as expected. There is an issue with at least the get_cookie() helper function. The cookie prefix is specifically toted as a way to prevent collisions, however this function doesn't solve for collisions.

```php
    function get_cookie($index, bool $xssClean = false)
    {
        $prefix  = isset($_COOKIE[$index]) ? '' : config(App::class)->cookiePrefix;
        $filter  = $xssClean ? FILTER_SANITIZE_FULL_SPECIAL_CHARS : FILTER_DEFAULT;

        return $request->getCookie($prefix . $index, $filter);
    }
```
This function looks to see if the base cookie without the prefix exists, and if it does not exist then it looks for the prefixed cookie. It should look for the prefixed cookie and reject the un-prefixed cookie entirely, otherwise it creates collisions.

### Steps to Reproduce

If you take the following example:

Config/App.php
`
public $cookiePrefix   = 'prefix_';
`

Theoretical array of cookies:
```php
$_COOKIES = [
    'prefix_test' => 'Right Value',
    'text'        => 'Wrong Value',
];
```

`
get_cookie('test');
`

returns 'Wrong Value' by default, ideally it should return 'Right Value' and even if 'prefix_text' is not set, it should return null and never 'Wrong Value'

### Expected Output

returns 'Wrong Value' by default, ideally it should return 'Right Value' and even if 'prefix_text' is not set, it should return null and never 'Wrong Value'

### Anything else?

This could be fairly easily fixed by just enforcing the prefix:
`
$prefix  = config(App::class)->cookiePrefix ?? '';
`

However, I'm not sure if this should be enforced at a deeper stage of getting the cookie, and/or the function should be extended to allow for a prefix to be set independently if desired like:

```php
    function get_cookie($index, bool $xssClean = false, string $prefix = null)
    {
        $prefix  = $prefix !== null ? $prefix : config(App::class)->cookiePrefix;
        $filter  = $xssClean ? FILTER_SANITIZE_FULL_SPECIAL_CHARS : FILTER_DEFAULT;

        return $request->getCookie($prefix . $index, $filter);
    }
```

colethorsen:
Fixes #6009

This allows the cookie_helper to enforce the default prefix if one exists, and also allows the user to set their own prefix including none at all to overwrite the prefix if necessary. This way there is no longer a collision issue if there are cookies that are prefixed and un-prefixed.

There is also an issue with the session setup where it would inconsistently set the name of the session cookie with and without the cookie prefix leading to multiple session cookies and also effectively creating the same prefixed vs un-prefixed issue.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/cookie_helper.php
+++ b/system/Helpers/cookie_helper.php
@@ -60,6 +60,7 @@ if (! function_exists('get_cookie'))
 	 *
 	 * @param string  $index
 	 * @param boolean $xssClean
+	 * @param string  $prefix
 	 *
 	 * @return mixed
 	 *
--- a/system/Helpers/cookie_helper.php
+++ b/system/Helpers/cookie_helper.php
@@ -65,9 +65,9 @@ if (! function_exists('get_cookie'))
 	 *
 	 * @see \CodeIgniter\HTTP\IncomingRequest::getCookie()
 	 */
-	function get_cookie($index, bool $xssClean = false)
+	function get_cookie($index, bool $xssClean = false, string $prefix = null)
 	{
-		$prefix  = isset($_COOKIE[$index]) ? '' : config(App::class)->cookiePrefix;
+		$prefix  = $prefix !== null ? $prefix : config(App::class)->cookiePrefix;
 		$request = Services::request();
 		$filter  = $xssClean ? FILTER_SANITIZE_STRING : FILTER_DEFAULT;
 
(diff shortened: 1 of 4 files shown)


colethorsen:
Based the update on the wrong branch

PR #6024, opened 2022-05-24
Fix Cookie Prefix

### LINKED ISSUE #6009: Bug: `get_cookie()` cookie prefix behavior

colethorsen:
Fixes #6009

This allows the cookie_helper to enforce the default prefix if one exists, and also allows the user to set their own prefix including none at all to overwrite the prefix if necessary. This way there is no longer a collision issue if there are cookies that are prefixed and un-prefixed.

There is also an issue with the session setup where it would inconsistently set the name of the session cookie with and without the cookie prefix leading to multiple session cookies and also effectively creating the same prefixed vs un-prefixed issue.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/cookie_helper.php
+++ b/system/Helpers/cookie_helper.php
@@ -61,9 +61,9 @@ if (! function_exists('get_cookie')) {
      *
      * @see \CodeIgniter\HTTP\IncomingRequest::getCookie()
      */
-    function get_cookie($index, bool $xssClean = false)
+    function get_cookie($index, bool $xssClean = false, string $prefix = null)
     {
-        $prefix  = isset($_COOKIE[$index]) ? '' : config(App::class)->cookiePrefix;
+        $prefix  = $prefix !== null ? $prefix : config(App::class)->cookiePrefix;
         $request = Services::request();
         $filter  = $xssClean ? FILTER_SANITIZE_FULL_SPECIAL_CHARS : FILTER_DEFAULT;
 
--- a/system/Helpers/cookie_helper.php
+++ b/system/Helpers/cookie_helper.php
@@ -63,7 +63,7 @@ if (! function_exists('get_cookie')) {
      */
     function get_cookie($index, bool $xssClean = false, ?string $prefix = null)
     {
-        $prefix  = $prefix !== null ? $prefix : config(App::class)->cookiePrefix;
+        $prefix ??= config(App::class)->cookiePrefix;
         $request = Services::request();
         $filter  = $xssClean ? FILTER_SANITIZE_FULL_SPECIAL_CHARS : FILTER_DEFAULT;
 
--- a/system/Helpers/cookie_helper.php
+++ b/system/Helpers/cookie_helper.php
@@ -61,7 +61,7 @@ if (! function_exists('get_cookie')) {
      *
      * @see \CodeIgniter\HTTP\IncomingRequest::getCookie()
      */
-    function get_cookie($index, bool $xssClean = false, string $prefix = null)
+    function get_cookie($index, bool $xssClean = false, ?string $prefix = null)
     {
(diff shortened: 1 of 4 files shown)


kenjis:
Thank you for sending a PR.

> There is also an issue with the session setup where it would inconsistently set the name of the session cookie with and without the cookie prefix leading to multiple session cookies and also effectively creating the same prefixed vs un-prefixed issue.

I don't know this issue. Could you make it  another PR?
We don't change the Session cookie name.
See https://github.com/codeigniter4/CodeIgniter4/pull/5951#issuecomment-1124407187

kenjis:
This is my simple question.

> We expect all contributions to conform to our [style guide](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/styleguide.md), be commented (inside the PHP source files), be documented (in the [user guide](https://codeigniter4.github.io/userguide/)), and unit tested (in the [test folder](https://github.com/codeigniter4/CodeIgniter4/tree/develop/tests)).

It is clearly stated that we need unit tests.
But almost (especially new) contributors don't write test code.
Do you know why?

datamweb:
> Do you know why?

I understand why! 🥇

colethorsen:
Well if it’s not supposed to set the prefix on the session cookie with the header then it’s broken occasionally in the opposite way.

kenjis:
@colethorsen I don't know why the session is broken occasionally, and why this PR fixes it.

Anyway, the session cookie name is another issue.
Would you please remove it from this PR?
Otherwise this PR won't go ahead.

PR #6035, opened 2022-05-26
fix: IP address validation with type error prevent

valmorflores:
Ip validation update for default option using FILTER_DEFAULT

**Description**
In php8, $which = 0; results error like:
Fatal error: Uncaught TypeError: filter_var(): Argument #3 ($options) must be of type array

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -274,7 +274,7 @@ class FormatRules
                 break;
 
             default:
-                $which = 0;
+                $which = FILTER_DEFAULT;
         }
 
         return filter_var($ip, FILTER_VALIDATE_IP, $which) !== false


kenjis:
Thank you for sending the PR.

But I can't reproduce the error.
The existing tests pass on PHP 8.
https://github.com/codeigniter4/CodeIgniter4/blob/5395498038abccc381fc739d26098e17fc94d4e5/tests/system/Validation/FormatRulesTest.php#L305

And, we expect all contributions to conform to our [style guide](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/styleguide.md), be commented (inside the PHP source files), be documented (in the [user guide](https://codeigniter4.github.io/userguide/)), and unit tested (in the [test folder](https://github.com/codeigniter4/CodeIgniter4/tree/develop/tests)).

Please write test code before fixing the product code if you can.

paulbalandan:
`filter_var()`'s third parameter accepts either an array or disjunction of integers.
https://www.php.net/manual/en/function.filter-var.php

kenjis:
And `0` is the default value.

> filter_var(mixed $value, int $filter = FILTER_DEFAULT, array|int $options = 0): mixed

valmorflores:
hi! thanks @kenjis!
I'm using an old version and I can't reproduce error in the latest version.
For parsing: the field declaration is "string" and the data is changed to 0.
Wouldn't it be better to use another var in this case? just for type consistency

Regards

paulbalandan:
> Wouldn't it be better to use another var in this case? just for type consistency
@valmorflores, what do you mean?

paulbalandan:
> Wouldn't it be better to use another var in this case? just for type consistency
@valmorflores, what do you mean?

valmorflores:
Hi @paulbalandan 


The param $which  is a string type (user opt for 'ipv4', 'ipv6' etc)
`filter_var($ip, FILTER_VALIDATE_IP, $options) //<--- options are another type`

Just for clear flow, I suggest we use another variable name: $options and code like this:

```
public function valid_ip(?string $ip = null, ?string $which = null): bool {

        if (empty($ip)) {
            return false;
        }

        switch (strtolower($which ?? '')) {
            case 'ipv4':
                $options = FILTER_FLAG_IPV4;
                break;

            case 'ipv6':
                $options = FILTER_FLAG_IPV6;
                break;

            default:
                $options  = 0;
        }

        return filter_var($ip, FILTER_VALIDATE_IP, $options) !== false
            || (! ctype_print($ip) && filter_var(inet_ntop($ip), FILTER_VALIDATE_IP, $options) !== false);
    }
```

kenjis:
Reassignments may cause bugs. It is good practice not to use reassignments.

paulbalandan:
Agreed. @valmorflores can you open a new PR with your proposal?

valmorflores:
Yes! Thanks @kenjis @paulbalandan

Issue #6032, opened 2022-05-26
Bug: SQLSRV with specified port over network


alfisahr:
### PHP Version

7.4

### CodeIgniter4 Version

4.0

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

cli-server (PHP built-in webserver)

### Database

SQL Server 2019

### What happened?

Database doesn't connect using non default SQL Server port, which is 1433 is default port. Especially when we connect to other machine (over network) instead of our local. Configuration `database.default.port = `  (in .env) or `$config['default']['port'] = `  (in config) didn't give any effect, not worked.

If we use local db, whatever port that use in the server db, the connection still work. But the problem comes if we connect other machine in the network.

In the `Connection.php` class under `/vendor/codeigniter4/framework/system/Database/SQLSRV` there are
```php
...
$this->connID = sqlsrv_connect($this->hostname, $connection);
...
```

then I modified like this
```php
...
$this->connID = sqlsrv_connect($this->hostname . ', ' . $this->port, $connection);
...
```

it solved the problem

### Steps to Reproduce

* Use db from other machine that have custom port, instead of default port.
* You can specify port db in the configuration following that db server custom port

### Expected Output

Connection to db works as same like using default port

### Anything else?

I tried to modify this class under `/vendor/codeigniter4/framework/system/Database/SQLSRV/Connection.php`

from this
```php
...
$this->connID = sqlsrv_connect($this->hostname, $connection);
...
```

to this
```php
...
$this->connID = sqlsrv_connect($this->hostname . ', ' . $this->port, $connection);
...
```

How to override this class in our library, without directly modify like that.

Thank you.

kenjis:
This bug was fixed in the `develop` branch,
and it will be included in v4.2.0.

Thank you for reporting.

iRedds:
@alfisahr Can you tell me if this fix solved the problem?

PR #6036, opened 2022-05-27
Bug: The SQLSRV driver ignores the port value from the config.

iRedds:
**Description**
Fixes https://github.com/codeigniter4/CodeIgniter4/issues/6032

The SQLSRV driver ignores the port value from the config.

This PR fixes a bug.
Also, if a port is already specified in the hostname, the port from the configuration will be ignored.

It seems to me that the default tests are enough, because the database connection settings for the tests use the default port, which was ignored when configuring the connection.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/Database/SQLSRV/Connection.php
+++ b/system/Database/SQLSRV/Connection.php
@@ -120,6 +120,10 @@ class Connection extends BaseConnection
             unset($connection['UID'], $connection['PWD']);
         }
 
+        if (strpos($this->hostname, ',') === false) {
+            $this->hostname = $this->port !== '' ? "{$this->hostname}, {$this->port}" : $this->hostname;
+        }
+
         sqlsrv_configure('WarningsReturnAsErrors', 0);
         $this->connID = sqlsrv_connect($this->hostname, $connection);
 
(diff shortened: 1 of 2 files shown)


system/Database/SQLSRV/Connection.php
```diff
@@ -120,6 +120,10 @@ public function connect(bool $persistent = false)
             unset($connection['UID'], $connection['PWD']);
         }
 
+        if (strpos($this->hostname, ',') === false) {
```
samsonasik, line 123: if can be combined so no need to re-assign hostname when no port:

```php
if (strpos($this->hostname, ',') === false && $this->port !== '')

iRedds: I agree. Thanks

samsonasik (approved):

kenjis:
Ref https://www.php.net/sqlsrv_connect

kenjis (approved):

kenjis:
> It seems to me that the default tests are enough, because the database connection settings for the tests use the default port, which was ignored when configuring the connection.

This bug was not detected by the existing tests, so tests are not enough.
But to test this, we build another SQL Server with the different port.
It costs a lot.

I've confirmed Port 1444 server can be connected with this PR.

iRedds:
@kenjis I don't see the point in an additional server.
We can use non-standard ports for each DBMS for github actions and in test configs. Or at least change the ports in the docker containers to listen.

kenjis:
@iRedds Okay, you are correct.
Changing the port of the docker containers to listen is enough.

PR #6043, opened 2022-05-28
Added `type` param for `get_file_info()`

datamweb:
PR allows the `get_file_info()` function in file `filesystem_helper.php` to return the file type.

**Description**
Need rebase and cs-fix.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/filesystem_helper.php
+++ b/system/Helpers/filesystem_helper.php
@@ -289,7 +289,7 @@ if (! function_exists('get_file_info')) {
      *
      * Given a file and path, returns the name, path, size, date modified
      * Second parameter allows you to explicitly declare what information you want returned
-     * Options are: name, server_path, size, date, readable, writable, executable, fileperms
+     * Options are: name, server_path, size, date, type, readable, writable, executable, fileperms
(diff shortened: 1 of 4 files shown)


kenjis:
It is not *type*, it is *extention*. So the name should be `extention` or `ext`.

https://www.php.net/manual/en/splfileinfo.gettype.php
https://www.php.net/manual/en/splfileinfo.getextension.php

kenjis:
@datamweb Run `composer cs-fix`.

kenjis:
I'm not sure that this PR should be merged.

We can get it with `File` class.
https://codeigniter4.github.io/CodeIgniter4/libraries/files.html#taking-advantage-of-spl

```php
$file = new \CodeIgniter\Files\File($path);
echo $file->getExtension();
```

datamweb:
> We can get it with `File` class.

@kenjis , I did not know of the existence of `File` class.
It does not make sense to merge with these conditions.
So I close **PR**.
Thanks for the explanation.

PR #5973, opened 2022-05-07
fix: slash_item() errors when asked for a configuration item which doesn't exist

aletoropov:
**Description**
Check object property which doesn't exist

Fixes #5967

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #6058, opened 2022-06-01
Fix `slash_item()` erroring when property fetched does not exist on `Config\App`

### LINKED ISSUE #5967: Bug: slash_item() errors when asked for a configuration item which doesn't exist
### PHP Version

8.1

### CodeIgniter4 Version

4.1.9

Manual (zip or tar.gz)

Linux

### Which server did you use?

### Database

_No response_

### What happened?

`slash_item` in `Sys\Common` errors when the configuration item doesn’t exist, which is contrary to expected outcome as per the documention

### Steps to Reproduce

```
<?php

namespace App\Controllers;

{
    public function poc()
    {
        # returns the baseURL as expected
        echo slash_item('baseURL');

        # this generates an error - detailed below
        echo slash_item('item_does_not_exist');
    }
}
```

The error that is displayed is this:-
```
ErrorException
Undefined property: Config\App::$item_does_not_exist
SYSTEMPATH/Common.php at line 987

980      *
981      * @return string|null The configuration item or NULL if
982      *                     the item doesn't exist
983      */
984     function slash_item(string $item): ?string
985     {
986         $config     = config(App::class);
987         $configItem = $config->{$item};
988 
989         if (! isset($configItem) || empty(trim($configItem))) {
990             return $configItem;
991         }
992 
993         return rtrim($configItem, '/') . '/';
994     }

```

### Expected Output

As [per the documentation](https://codeigniter.com/user_guide/general/common_functions.html#slash_item), the return should be `null` when a configuration doesn't exist

### Anything else?

I've never done a pull request before, but can attempt to do my first one if required

However I have identified what I believe to be the cause of this and here is the updated `slash_item` code for `Common.php`

```
if (! function_exists('slash_item')) {
    // Unlike CI3, this function is placed here because
    // it's not a config, or part of a config.
    /**
     * Fetch a config file item with slash appended (if not empty)
     *
     * @param string $item Config item name
     *
     * @return string|null The configuration item or NULL if
     *                     the item doesn't exist
     */
    function slash_item(string $item): ?string
    {
        $config     = config(App::class);
        
        if (property_exists($config, $item))
        {
            $configItem = $config->{$item};
        }
        
        if (! isset($configItem) || empty(trim($configItem))) {
            return null;
        }

        return rtrim($configItem, '/') . '/';
    }
}
```

paulbalandan:
**Description**
Supersedes and closes #5973 
Fixes #5967 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/CommonFunctionsTest.php
+++ b/tests/system/CommonFunctionsTest.php
@@ -391,7 +391,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
     public function testSlashItem()
     {
         $this->assertSame('/', slash_item('cookiePath')); // slash already there
-        $this->assertSame('', slash_item('cookieDomain')); // empty, so untouched
+        $this->assertNull(null, slash_item('cookieDomain')); // empty, so untouched
         $this->assertSame('en/', slash_item('defaultLocale')); // slash appended
     }
 
--- a/tests/system/CommonFunctionsTest.php
+++ b/tests/system/CommonFunctionsTest.php
@@ -30,6 +30,7 @@ use Config\App;
 use Config\Logger;
 use Config\Modules;
 use Kint;
+use RuntimeException;
 use stdClass;
 use Tests\Support\Models\JobModel;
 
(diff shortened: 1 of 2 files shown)


tests/system/CommonFunctionsTest.php
```diff
@@ -390,9 +390,9 @@ public function testReallyWritable()
 
     public function testSlashItem()
     {
-        $this->assertSame('/', slash_item('cookiePath')); // slash already there
-        $this->assertSame('', slash_item('cookieDomain')); // empty, so untouched
-        $this->assertSame('en/', slash_item('defaultLocale')); // slash appended
+        $this->assertSame('/', slash_item('cookiePath'));
+        $this->assertNull(slash_item('cookieDomain'));
```
kenjis, line 394: `$cookieDomain` is empty string, not null.
https://github.com/codeigniter4/CodeIgniter4/blob/451ebb0e4c9fab01312cf1091bd24cf741708825/app/Config/App.php#L261

```suggestion
        $this->assertSame('', slash_item('cookieDomain'));
```

kenjis: And a test for `slash_item('item_does_not_exist')` is needed.

tests/system/CommonFunctionsTest.php
```diff
@@ -390,9 +391,28 @@ public function testReallyWritable()
...
     {
-        $this->assertSame('/', slash_item('cookiePath')); // slash already there
-        $this->assertSame('', slash_item('cookieDomain')); // empty, so untouched
-        $this->assertSame('en/', slash_item('defaultLocale')); // slash appended
+        $this->assertSame('/', slash_item('cookiePath')); // /
+        $this->assertSame('', slash_item('cookieDomain')); // ''
+        $this->assertSame('en/', slash_item('defaultLocale')); // en
+        $this->assertSame('7200/', slash_item('sessionExpiration')); // int 7200
+        $this->assertSame('', slash_item('negotiateLocale')); // false
+        $this->assertSame('1/', slash_item('cookieHTTPOnly')); // true
```
kenjis, line 399: I can't think use cases like this, but this behavior is not changed.

kenjis, line 399: I can't think use cases like this, but this behavior is not changed.

paulbalandan: Yeah. Just to show these cases are covered by tests.

PR #6059, opened 2022-06-01
Fix MIME guessing of extension from type

### LINKED ISSUE #6046: Bug: Misdiagnosis `extension` for file `zip`
### PHP Version

8.1

### CodeIgniter4 Version

4.1.9

Windows

### Which server did you use?

cli

### Database

_No response_

### What happened?

Incorrect detection of zip file extensions.

```
$file = new \CodeIgniter\Files\File($myZipFilePath);
$ext = $file->guessExtension();
```
The expected output is **zip**. But the value of **pptx** is returned.

### Steps to Reproduce

Create a zip file in path `SUPPORTPATH . 'Files/myZipFile.zip';`.
Run the following test.

```
    public function testHaveExtensionZip()
    {
       $fileZip = SUPPORTPATH . 'Files/myZipFile.zip';
        $file = new File($fileZip);
        $this->assertSame('zip', $file->guessExtension());
    }
```
Test failure. Because **pptx** is returned.

### Expected Output

The value of `zip ` is returned.

### Anything else?

Temporary solution:
Transfer [L143](https://github.com/codeigniter4/CodeIgniter4/blob/d464fbf859afad076f69d4e2677dbf55955ac26e/app/Config/Mimes.php#L143) ~ [L149]https://github.com/codeigniter4/CodeIgniter4/blob/d464fbf859afad076f69d4e2677dbf55955ac26e/app/Config/Mimes.php#L149
 to [L103](https://github.com/codeigniter4/CodeIgniter4/blob/d464fbf859afad076f69d4e2677dbf55955ac26e/app/Config/Mimes.php#L103) lines in file `app/Config/Mimes.php`.

paulbalandan:
**Description**
Fixes #6046 by relying on the path extension as the `$proposedExtension` argument to `Config\Mimes::guessExtensionFromType()`. The extension is derived using `pathinfo()`

However, this fix also changes the behavior of the guessing if a proposed extension is given. Previously, if the proposal is invalid, the guessing ends and returns `null`. Now, this early termination is removed and the guessing is continued using the array mapping of extension to MIME types.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Mimes.php
+++ b/app/Config/Mimes.php
@@ -102,8 +102,6 @@ class Mimes
         ],
         'pptx' => [
             'application/vnd.openxmlformats-officedocument.presentationml.presentation',
-            'application/x-zip',
-            'application/zip',
         ],
         'wbxml' => 'application/wbxml',
         'wmlc'  => 'application/wmlc',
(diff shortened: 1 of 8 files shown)


paulbalandan:
Some edge cases I didn't see.

app/Config/Mimes.php
```diff
@@ -102,8 +102,6 @@ class Mimes
         ],
         'pptx' => [
             'application/vnd.openxmlformats-officedocument.presentationml.presentation',
-            'application/x-zip',
-            'application/zip',
```
paulbalandan, line 106: According to https://cdn.jsdelivr.net/gh/jshttp/mime-db@v1.49.0/db.json and https://gitlab.freedesktop.org/xdg/shared-mime-info/-/raw/master/data/freedesktop.org.xml.in which are used by Symfony in updating their MIME types, the first one is the IANA recognized MIME type.

paulbalandan:
Rebased to add the relevant docs.

kenjis (approved):

lonnieezell (approved):
I had to read through this a couple of times, but I believe the changes still keep the intended security features of those methods in place.

user_guide_src/source/installation/upgrade_421.rst
```diff
@@ -0,0 +1,53 @@
...
+
+.. contents::
+    :local:
+    :depth: 2
+
+Mandatory File Changes
+**********************
+
+app/Config/Mimes.php
+===================
```
paulbalandan, line 19:
```suggestion
app/Config/Mimes.php
====================
```

MGatner (approved):

kenjis:
@paulbalandan Thank you!

PR #6065, opened 2022-06-02
docs: update repositories.rst

kenjis:
**Description**
- update repositories.rst

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/installation/repositories.rst
+++ b/user_guide_src/source/installation/repositories.rst
@@ -1,6 +1,13 @@
 CodeIgniter Repositories
 ########################
 
+.. contents::
+    :local:
+    :depth: 2
+
+codeIgniter4 organization
+=========================
+
 The CodeIgniter 4 open source project has its own
 `GitHub organization <https://github.com/codeigniter4>`_.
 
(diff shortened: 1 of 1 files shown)


user_guide_src/source/installation/repositories.rst
```diff
@@ -1,6 +1,13 @@
 CodeIgniter Repositories
 ########################
 
+.. contents::
+    :local:
+    :depth: 2
+
+codeIgniter4 organization
```
paulbalandan, line 8: Is the lowercased `c` intended?

paulbalandan: Then I suppose the `i` should be, too.

MGatner (approved):
I agree with Paul, either all PascalCase or all lower. Rest looks good!

user_guide_src/source/installation/repositories.rst
kenjis: Oops, I overlooked that.

MGatner (approved):

PR #6071, opened 2022-06-03
Adds information that ignore() is not unique to insert

eSkiSo:
Adds information that ignore() can be used in insertBatch, update and delete (when database supports it).
Not really clear on the docs at the moment.

**Description**
State that ignore() can be used in insertBatch(), update() and delete(), not only insert().

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


MGatner (approved):
Thanks!

samsonasik:
Thank you @eSkiSo

PR #6073, opened 2022-06-04
feat: add routes useSupportedLocalesOnly property

pjsde:
Adds a control flag so that the ``Router`` only validates the locales ​​that are defined in ``App::$supportedLocales``, returning an ``CodeIgniter\Exceptions\PageNotFoundException::forLocaleNotSupported()`` if the locale introduced is not supported.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/outgoing/localization.rst
+++ b/user_guide_src/source/outgoing/localization.rst
@@ -81,7 +81,7 @@ segment will be your locale:
 In this example, if the user tried to visit ``http://example.com/fr/books``, then the locale would be
 set to ``fr``, assuming it was configured as a valid locale.
 
-If the value doesn't match a valid locale as defined in the App configuration file, the default
+If the value doesn't match a valid locale as defined in ``$supportedLocales`` in **app/Config/App.php**, the default
 locale will be used in it's place, unless you set to use only the supported locales defined in the App configuration
 file:
 
--- a/user_guide_src/source/outgoing/localization.rst
+++ b/user_guide_src/source/outgoing/localization.rst
@@ -81,8 +81,12 @@ segment will be your locale:
 In this example, if the user tried to visit ``http://example.com/fr/books``, then the locale would be
 set to ``fr``, assuming it was configured as a valid locale.
 
-.. note:: If the value doesn't match a valid locale as defined in the App configuration file, the default
-    locale will be used in it's place.
+If the value doesn't match a valid locale as defined in the App configuration file, the default
+locale will be used in it's place, unless you set to use only the supported locales defined in the App configuration
+file:
+
+.. literalinclude:: localization/018.php
+
 
 Retrieving the Current Locale
 =============================
--- a/system/Exceptions/PageNotFoundException.php
+++ b/system/Exceptions/PageNotFoundException.php
@@ -45,6 +45,11 @@ class PageNotFoundException extends OutOfBoundsException implements ExceptionInt
         return new static(self::lang('HTTP.methodNotFound', [$method]));
     }
 
+    public static function forLocaleNotSupported(string $locale)
+    {
+        return new static(self::lang('HTTP.localeNotSupported', [$locale]));
+    }
+
     /**
      * Get translated system message
      *
--- a/system/Language/en/HTTP.php
+++ b/system/Language/en/HTTP.php
@@ -52,6 +52,7 @@ return [
     'emptyController'    => 'No Controller specified.',
     'controllerNotFound' => 'Controller or its method is not found: {0}::{1}',
     'methodNotFound'     => 'Controller method is not found: {0}',
+    'localeNotSupported' => 'Locale is not supported: {0}',
 
     // CSRF
     // @deprecated use `Security.disallowedAction`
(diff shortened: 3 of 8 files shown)


iRedds:
Or maybe it would be easier to bind to App::$supportedLocales?

pjsde:
@iRedds With this that can be done in the Routes config with ``$routes->setLocalePattern('('.implode(config('App')->supportedLocales).')')``, but also you can do this ``$routes->setLocalePattern('[a-z]{2}[_-][A-Z]{2}')``, that is, you are not limited to the languages ​​​​that are in ``$App::supportedLocales`` and we are free to use it in the way that is most useful at any time.

iRedds:
However, the application locale depends on App::$supportedLocales and App::$defaultLocale.
Yes, you can specify any pattern and specify unsupported languages, but if the languages are not supported, then the application will be assigned a default locale.
Therefore, defining a pattern manually is pointless.
Although maybe I misunderstood something.

kenjis:
> However, the application locale depends on App::$supportedLocales and App::$defaultLocale.
Yes, you can specify any pattern and specify unsupported languages, but if the languages are not supported, then the application will be assigned a default locale.
Therefore, defining a pattern manually is pointless.

I agree. I don't know why the locale pattern needs to override `App::$supportedLocales`.

iRedds:
@pjsde 
1. If the pattern ``(x|y)`` is used, then this is a capture group. For the ``{locale}/test/(:any)`` route, the ``Controller::method/$1`` handler will be passed the locale instead of the expected value. That is, for the URL ``x/test/1``, the value of ``$1`` will be ``x``, not ``1``.

2. If the pattern ``x|y`` is used, then the locale is not captured, but the route ``{locale}/test/(:any)`` will match the URLs ``y/test/1`` and ` `x``. Which is wrong, and it will also cause locale detection to fail because no match will be found.

These cases are not covered by tests.

pjsde:
You are right. What I intended to do was to limit the routes to the languages ​​defined in ``$App::supportedLocales`` so that it would give an error if the introduced language was not defined, but this can be done through the filters.

iRedds:
I got your idea. But for now I don't know the solution.
Maybe the team will have some thoughts about it.

kenjis:
See https://forum.codeigniter.com/showthread.php?tid=82071

kenjis:
> If the value doesn’t match a valid locale as defined in the App configuration file, the default locale will be used in it’s place.
https://codeigniter4.github.io/CodeIgniter4/outgoing/localization.html#in-routes

If you don't want this, add option to change the behavior?

E.g.
```php
$routes->useDefinedLocalesOnly(true);
```

pjsde:
@kenjis I like that solution. It wil solve the problem.

Do you think there should be a config like ``$App::$useDefinedLocalesOnly = false;`` to initially define the state of ``RouteCollection::$useDefinedLocalesOnly`` or should it be initialized directly in the class?

kenjis:
1. We are separating the configuration for each class from the `Config\App`, like `Config\Cookie`, `Config\Security`.
2. `Routes.php` is not a class.

So, it seems we need to add a property and a setter method for it.

The config in App is `$supportedLocales`, so the name is `$useSupportedLocalesOnly` or something is better?

InsiteFX:
The problem that I ran into is when designing a DropDown for lanuages with the country flag the lang8uage does not match with the country code.

Example: In the CI Translations the code for Japan is ja but using the Country+Region it is ja-JP or en-JPN there is also
a three character code.

If you look at the flags-icons which are svg images they are named different Japan is jp.

I think we need to change the translation files to the correct 2, 4 or 5 codes to match the language or language+region codes.

Just my opinion.

pjsde:
@InsiteFX

> The problem that I ran into is when designing a DropDown for lanuages with the country flag the language does not match with the country code.

I think that is a diferent subject from this PR

> I think we need to change the translation files to the correct 2, 4 or 5 codes to match the language or language+region codes.

The `` Language::getLine()`` already implements what you are proposing

user_guide_src/source/outgoing/localization.rst
```diff
@@ -81,8 +81,12 @@ segment will be your locale:
 In this example, if the user tried to visit ``http://example.com/fr/books``, then the locale would be
 set to ``fr``, assuming it was configured as a valid locale.
 
-.. note:: If the value doesn't match a valid locale as defined in the App configuration file, the default
-    locale will be used in it's place.
+If the value doesn't match a valid locale as defined in the App configuration file, the default
```
kenjis, line 84:
```suggestion
If the value doesn't match a valid locale as defined in ``$supportedLocales`` in **app/Config/App.php**, the default
```

kenjis:
@pjsde Can you update the explanation in https://github.com/codeigniter4/CodeIgniter4/pull/6073#issue-1260856474 ?

pjsde:
@kenjis done.

kenjis (approved):
Thank you!

paulbalandan (approved):

PR #6077, opened 2022-06-05
feat: get codeigniter context access

pjsde:
With CI 4.2.0 constant ``SPARKED`` is deprecated and we must use the ``CodeIgniter::$context``, but there is no method to obtain it.

Also the ``CodeIgniter::is parked()``, ``CodeIgniter::Php Cli`` and ``CodeIgniter::isWeb()`` methods must be public since ``CodeIgniter::$context`` already has an accessor.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -383,7 +383,7 @@ class CodeIgniter
     /**
      * Invoked via spark command?
      */
-    public function isSparked(): bool
+    private function isSparked(): bool
     {
         return $this->context === 'spark';
     }
@@ -391,7 +391,7 @@ class CodeIgniter
     /**
      * Invoked via php-cli command?
      */
-    public function isPhpCli(): bool
+    private function isPhpCli(): bool
     {
         return $this->context === 'php-cli';
     }
@@ -399,7 +399,7 @@ class CodeIgniter
     /**
      * Web access?
      */
-    public function isWeb(): bool
+    private function isWeb(): bool
     {
         return $this->context === 'web';
     }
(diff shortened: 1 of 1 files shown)


kenjis:
Why do you need to access `CodeIgniter::$context`?

pjsde:
I have classes that are shared between spark and web, and I need to know what context I'm in. Until now I was able to know through the ``SPARKED`` constant, now I will not know since the constant is deprecated

kenjis:
> I have classes that are shared between spark and web, and I need to know what context I'm in. 

Why do you need to know the context?

Probably most models/libraries are shared between spark and web, but they probably don't need to know the context.
At least all libraries that CI4 provides do not know the context.

If you want to know CLI or not, `is_cli()` can be used.

MGatner:
CLI/web is not the same as context. Since this is a feature we used to have but deprecated I am okay with opening access to the context. A compromise would be to keep the `isXxxx()` methods private and just add the getter.

kenjis:
@pjsde Okay.
This is my personal interest, why do you need to know the context?
In my opinion, all classes except `CodeIgniter` does not need to know it.

pjsde:
@kenjis after more digging in code i realize that you are right. I can resolve with ``is_cli()``.

But despite that, I think @MGatner  is right in the sense that it was a feature that existed and will no longer be available once it has been defined as deprecated.

Tell me if I should close PR or not. Thanks.

kenjis:
@pjsde Thank you for the feedback.

If you really don't need this, I hope not to add the method.

If a dev uses the method, the class is coupled with `CodeIgniter`, that is not good generally.
`CodeIgniter` is the kernel of the framework.

PR #6080, opened 2022-06-06
fix: get_cookie() does not take Config\Cookie::$prefix

kenjis:
**Description**
Fixes #6079

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #6081, opened 2022-06-06
fix: Session Handlers do not take Config\Cookie

kenjis:
**Description**
- fix Session Handlers do not take `Config\Cookie`

Related:  #6080

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Session/Handlers/BaseHandler.php
+++ b/system/Session/Handlers/BaseHandler.php
@@ -12,6 +12,7 @@
 namespace CodeIgniter\Session\Handlers;
 
 use Config\App as AppConfig;
+use Config\Cookie as CookieConfig;
 use Psr\Log\LoggerAwareTrait;
 use SessionHandlerInterface;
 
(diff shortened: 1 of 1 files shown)


system/Session/Handlers/BaseHandler.php
```diff
@@ -102,14 +103,27 @@ abstract class BaseHandler implements SessionHandlerInterface
...
-        $this->cookieSecure = $config->cookieSecure;
-        $this->cookieName   = $config->sessionCookieName;
-        $this->matchIP      = $config->sessionMatchIP;
-        $this->savePath     = $config->sessionSavePath;
-        $this->ipAddress    = $ipAddress;
+        /** @var CookieConfig|null $cookie */
+        $cookie = config('Cookie');
+
+        if ($cookie instanceof CookieConfig) {
+            $this->cookiePrefix = $cookie->prefix;
```
pjsde, line 110: This line should be removed because ``BaseHandler::$cookiePrefix`` is not used anywhere since session cookies have no prefix.

system/Session/Handlers/BaseHandler.php
```diff
@@ -102,14 +103,27 @@ abstract class BaseHandler implements SessionHandlerInterface
...
+            $this->cookieDomain = $cookie->domain;
+            $this->cookiePath   = $cookie->path;
+            $this->cookieSecure = $cookie->secure;
+        } else {
+            // @TODO Remove this fallback when deprecated `App` members are removed.
+            // `Config/Cookie.php` is absence
+            $this->cookiePrefix = $config->cookiePrefix;
```
pjsde, line 117: This line should be removed because ``BaseHandler::$cookiePrefix`` is not used anywhere since session cookies have no prefix.

system/Session/Handlers/BaseHandler.php
kenjis: Thanks!

Done.

system/Session/Handlers/BaseHandler.php
kenjis: Done.

paulbalandan (approved):

PR #6083, opened 2022-06-06
docs: add about setCookie() prefix details

kenjis:
**Description**
- update `set_cookie()`
- update `Response::setCookie()`

Ref https://github.com/codeigniter4/CodeIgniter4/issues/6009#issuecomment-1136003776

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/system/HTTP/ResponseTrait.php
+++ b/system/HTTP/ResponseTrait.php
@@ -543,7 +543,7 @@ trait ResponseTrait
      * @param string              $expire   Cookie expiration time in seconds
      * @param string              $domain   Cookie domain (e.g.: '.yourdomain.com')
      * @param string              $path     Cookie path (default: '/')
-     * @param string              $prefix   Cookie name prefix
+     * @param string              $prefix   Cookie name prefix ('': the default prefix)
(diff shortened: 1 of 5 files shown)


kenjis:
The current behavior of cookie prefix:
> If set to `''`, the default value from **app/Config/Cookie.php** will be used.

Is this a bug?

kenjis:
On CI3,
```php
$config['cookie_prefix']	= 'ci3_';
```
```php
	$cookie = [
		'name'          => 'The Cookie Name',
		'value'         => 'The Value',
		'prefix'        => null,
	];
	$this->input->set_cookie($cookie);
```

`'prefix' => null` --> `"ci3_The Cookie Name"`
`'prefix' => ''` --> `"ci3_The Cookie Name"`
`'prefix' => 'prefix'` --> `"prefixThe Cookie Name"`

Ref, https://codeigniter.com/userguide3/libraries/input.html#CI_Input::set_cookie

MGatner (approved):
This surprised me but does seem to be intentional. I'm going to label this to revisit for version 5, because I don't think this is ideal.

Issue #6086, opened 2022-06-06
Bug: CI 4.2.0 - running CLI command with csrf enabled causes error


sclubricants:
### PHP Version

8.0

### CodeIgniter4 Version

4.2

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

cli

### What happened?

After updating to version 4.2 I'm having errors running cli. If I disable csrf filter in `App/Config/Filters` then I don't get the error.

### Steps to Reproduce

Create a controller App\cron\test\run().

Ensure csrf filter enabled.

```diff
diff --git a/app/Config/Cookie.php b/app/Config/Cookie.php
index 8ee01c71b..37a24299c 100644
--- a/app/Config/Cookie.php
+++ b/app/Config/Cookie.php
@@ -62,7 +62,7 @@ class Cookie extends BaseConfig
      *
      * @var bool
      */
-    public $secure = false;
+    public $secure = true;
 
     /**
      * --------------------------------------------------------------------------
diff --git a/app/Config/Filters.php b/app/Config/Filters.php
index d0a97238b..960ae2358 100644
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -34,7 +34,7 @@ class Filters extends BaseConfig
     public $globals = [
         'before' => [
             // 'honeypot',
-            // 'csrf',
+            'csrf',
             // 'invalidchars',
         ],
         'after' => [
```

Run command: `php8.0 /var/www/portal/public/index.php cron/test/run`

Response:
[CodeIgniter\Security\Exceptions\SecurityException]

The action you requested is not allowed.

at SYSTEMPATH/HTTP/ResponseTrait.php:711

And in the logs:
CRITICAL - 2022-06-06 09:54:01 --> The action you requested is not allowed.
in SYSTEMPATH/HTTP/ResponseTrait.php on line 711
 1 SYSTEMPATH/HTTP/ResponseTrait.php(711): CodeIgniter\Security\Exceptions\SecurityException::forDisallowedAction()
 2 SYSTEMPATH/HTTP/ResponseTrait.php(701): CodeIgniter\HTTP\Response->dispatchCookies()
 3 SYSTEMPATH/HTTP/ResponseTrait.php(445): CodeIgniter\HTTP\Response->sendCookies()
[...]

PR #6089, opened 2022-06-07
fix: when running on CLI, two Request objects were used in the system

kenjis:
**Description**
Fixes https://github.com/codeigniter4/CodeIgniter4/issues/6086

The current implementation has two Request objects (`IncommingRequest` and `CLIRequest`) when running on CLI.
Because some classes call `Services::request()` inside and it always returns `IncommingRequest`.

When `CLIRequest` is used in `CodeIgniter`, developers assume that the `CLIRequest` object is used system-wide.
This PR makes that so.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide
- [x] Apply reverted #5653 again

### DIFF
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -487,7 +487,7 @@ class Services extends BaseService
         }
 
         // @TODO remove the following code for backward compatibility
-        return static::incommingrequest($config, $getShared);
+        return static::incomingrequest($config, $getShared);
     }
 
     /**
@@ -502,7 +502,7 @@ class Services extends BaseService
         if ($isCli) {
             $request = AppServices::clirequest($config);
         } else {
-            $request = AppServices::incommingrequest($config);
+            $request = AppServices::incomingrequest($config);
 
             // guess at protocol if needed
             $request->setProtocolVersion($_SERVER['SERVER_PROTOCOL'] ?? 'HTTP/1.1');
@@ -519,7 +519,7 @@ class Services extends BaseService
      *
      * @internal
      */
-    public static function incommingrequest(?App $config = null, bool $getShared = true)
+    public static function incomingrequest(?App $config = null, bool $getShared = true)
     {
         if ($getShared) {
             return static::getSharedInstance('request', $config);
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -472,7 +472,9 @@ class Services extends BaseService
     /**
      * The Request class models an HTTP request.
      *
-     * @return IncomingRequest
+     * CodeIgniter::getRequestObject() injects CLIRequest if $this->request is CLIRequest.
+     *
+     * @return CLIRequest|IncomingRequest
      */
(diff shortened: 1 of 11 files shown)


samsonasik (approved):

MGatner (commented):
I want to think about this one a bit. On the one hand it is convenient (from a developer standpoint) to have the appropriate Request class anytime the service is called; on the other hand it complicates (from a coding standpoint) our PSR-15 layer even more.

kenjis:
Yes, we need to consider this.

kenjis:
Changed the implementation a bit.

MGatner:
I'm coming around to this. It is a bit hacky, but our current Request implementation is a mess so that might just need to be the case.

system/Config/Services.php
```diff
@@ -470,16 +472,59 @@ public static function renderer(?string $viewPath = null, ?ViewConfig $config =
...
+     * This method is called from CodeIgniter::getRequestObject().
+     *
+     * @internal
+     */
+    public static function createRequest(App $config, bool $isCli = false): void
+    {
+        if ($isCli) {
+            $request = AppServices::clirequest($config);
+        } else {
+            $request = AppServices::incommingrequest($config);
```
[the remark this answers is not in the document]
kenjis: Fixed.

kenjis:
I've been thinking about it for a while and can't think of any other way to fix it except this way in the current situation.
Any idea is welcome.

MGatner:
If you've given this plenty of thought I don't feel the need to. At this point pretty much anything would be an improvement and what you've done looks reasonable to me.

MGatner (approved):

samsonasik (approved):

kenjis:
Added changelog.

user_guide_src/source/changelogs/v4.2.2.rst
```diff
@@ -23,11 +24,13 @@ Changes
 *******
 
 - Fixed: ``BaseBuilder::increment()`` and ``BaseBuilder::decrement()`` do not reset the ``BaseBuilder`` state after a query.
+- Now ``CLIRequest::isCLI()`` always return true.
```
MGatner, line 27:
```suggestion
- Now ``CLIRequest::isCLI()`` always returns true.
```

user_guide_src/source/changelogs/v4.2.2.rst
```diff
@@ -23,11 +24,13 @@ Changes
...
+- Now ``IncommingRequest::isCLI()`` always return false.
```
MGatner, line 28:
```suggestion
- Now ``IncommingRequest::isCLI()`` always returns false.
```

MGatner (changes requested):
Tiny syntax changes then I'm good with this.

MGatner (approved):

PR #6090, opened 2022-06-07
Fix decorators

lonnieezell:
**Description**
Fixed a typo for View Decorators documentation. Also, the current code, while passing tests, wasn't working
for me in a real-word test. Modified to better check the class implements the ViewDecoratorInterface.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/ViewDecoratorTrait.php
+++ b/system/View/ViewDecoratorTrait.php
@@ -24,7 +24,7 @@ trait ViewDecoratorTrait
         $decorators = \config('View')->decorators;
 
         foreach ($decorators as $decorator) {
-            if (! in_array(ViewDecoratorInterface::class, class_implements($decorator))) {
+            if (! in_array(ViewDecoratorInterface::class, class_implements($decorator), true)) {
                 throw ViewException::forInvalidDecorator($decorator);
             }
 
--- a/system/View/ViewDecoratorTrait.php
+++ b/system/View/ViewDecoratorTrait.php
@@ -24,7 +24,7 @@ trait ViewDecoratorTrait
         $decorators = \config('View')->decorators;
 
         foreach ($decorators as $decorator) {
-            if (! in_array(ViewDecoratorInterface::class, class_implements($decorator), true)) {
+            if (! is_subclass_of($decorator, ViewDecoratorInterface::class)) {
                 throw ViewException::forInvalidDecorator($decorator);
             }
 
--- a/system/View/ViewDecoratorTrait.php
+++ b/system/View/ViewDecoratorTrait.php
@@ -24,7 +24,7 @@ trait ViewDecoratorTrait
         $decorators = \config('View')->decorators;
 
         foreach ($decorators as $decorator) {
-            if (! is_subclass_of($decorator, ViewDecoratorInterface::class)) {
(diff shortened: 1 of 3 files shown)


system/View/ViewDecoratorTrait.php
```diff
@@ -24,7 +24,7 @@ protected function decorateOutput(string $html): string
         $decorators = \config('View')->decorators;
 
         foreach ($decorators as $decorator) {
-            if (! is_subclass_of($decorator, ViewDecoratorInterface::class)) {
```
kenjis, line 27: The code seems fine, but didi you get the ViewException thrown?

system/View/ViewDecoratorTrait.php
```diff
@@ -24,7 +24,7 @@ protected function decorateOutput(string $html): string
...
+            if (! in_array(ViewDecoratorInterface::class, class_implements($decorator), true)) {
```
MGatner, line 27: I think what we actually want here is the very underrated `is_a()`:
```suggestion
            if (! is_a($decorator, ViewDecoratorInterface::class, true)) {
```

MGatner (changes requested):

system/View/ViewDecoratorTrait.php
iRedds: In this context, is_a() and is_subclass_of() do the same.
class_implements() only returns interfaces, but an additional array lookup is added.

I don't really see the point of these changes.

lonnieezell: Honestly, I forget about `is_a`, but gave it a whirl - and it doesn't work either. If you feel like giving my "MadLibs" example I put together for a new feature video last night a whirl, you try it yourself: https://github.com/lonnieezell/ci-tutorials/tree/4.2-decorators

iRedds: @lonnieezell In this repository, all three functions work the same way. No exceptions are thrown. windows, php 7.4.27, Spark 4.2

MGatner: I also have no problems: https://madlibs.gatner.be/madlib

lonnieezell: Super strange - is_a and is_subclass_of both throw the Invalid Decorator exception for me on an M1 Mac. 

Sounds like the safest thing is to go ahead with the change, though, so that we know it operates correctly everywhere.

kenjis: I also have no problems with PHP 8.0 and 7.4 on Intel Mac.

lonnieezell: No idea what was happening last night. Trashed the vendor folder on that project and reinstalled and it worked. Reverted this change.

kenjis (approved):

paulbalandan (approved):

system/View/ViewDecoratorTrait.php
MGatner: 🤷‍♂️ When computers be like this, I always just say: keeps me employed!

MGatner (approved):

PR #6095, opened 2022-06-07
Remove useless catch

paulbalandan:
**Description**
It is useless to catch the exception then rethrow it without doing anything on it.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Cache/Handlers/MemcachedHandler.php
+++ b/system/Cache/Handlers/MemcachedHandler.php
@@ -115,8 +115,6 @@ class MemcachedHandler extends BaseHandler
             } else {
                 throw new CriticalError('Cache: Not support Memcache(d) extension.');
             }
-        } catch (CriticalError $e) {
-            throw $e;
         } catch (Exception $e) {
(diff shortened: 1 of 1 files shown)


system/Cache/Handlers/MemcachedHandler.php
```diff
@@ -115,8 +115,6 @@ public function initialize()
             } else {
                 throw new CriticalError('Cache: Not support Memcache(d) extension.');
             }
-        } catch (CriticalError $e) {
-            throw $e;
```
kenjis, line 119: This change will change the Exception message.
I don't think this catch is useless.

kenjis: Oh, I thought the last `catch (Exception $e)` catches all `CriticalError`.
You're correct. `class CriticalError extends Error`.

kenjis (approved):

MGatner (approved):

Issue #6097, opened 2022-06-08
Bug: improved autorouting doesn't work if class has one routed method


SteeveDroz:
### PHP Version

8.1

### CodeIgniter4 Version

4.2.0

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

Nginx

### What happened?

The example in the tutorial breaks if I add a defined route for a method in the same controller.

### Steps to Reproduce

Following the tutorial on [the official website](https://codeigniter.com/user_guide/incoming/controllers.html#let-s-try-it-hello-world), create a controller called `Helloworld` with the methods `getIndex` and `getComment`.

Create a third method called `test` (not `getTest`).

Define a route in `app/Config/Routes.php` like this:

```php
$routes->add('/test', 'Helloworld::test');
```

Everythink still works.

Change the route to:

```php
$routes->get('/test', 'Helloworld::test');
```

The URI `/test` works, but `/helloworld` and `/helloworld/comment` don't (*404 Page Not Found*).

### Expected Output

Adding a defined route to a method should not change the behavior of the auto routing of the other methods of the same controller.

### Anything else?

By adding the method `Home::getTest`, the command `php spark routes` gives the result:

```
CodeIgniter v4.2.0 Command Line Tool - Server Time: 2022-06-08 01:20:31 UTC-05:00

ERROR: 404
Cannot access the default controller "Home" with the controller name URI path.
```

kenjis:
Thank you for taking time.

This is not a bug. It is the specification for security reason.

> Can’t access controllers in Defined Routes.
> It completely separates controllers accessible via Auto Routes from those accessible via Defined Routes.
https://codeigniter.com/user_guide/changelogs/v4.2.0.html#new-improved-auto-routing

> Adding a defined route to a method should not change the behavior of the auto routing of the other methods of the same controller.

Why do you think so?

kenjis:
> Create a third method called `test` (not `getTest`).
>
> Define a route in `app/Config/Routes.php` like this:
>
```php
$routes->add('/test', 'Helloworld::test');
```
>
> Everythink still works.

If you add the route definition, http://localhost:8080/helloworld  and http://localhost:8080/helloworld/comment will be 404.

SteeveDroz:
> Can’t access controllers in Defined Routes.
[...]

PR #6098, opened 2022-06-08
fix: `spark routes` shows "ERROR: 404"

kenjis:
**Description**
Whe you enable Auto Route Improved,
By adding the method `Home::getTest`, the command `php spark routes` gives the result:
```
ERROR: 404
Cannot access the default controller "Home" with the controller name URI path.
```
Reported in #6097

**How to Test**
```diff
diff --git a/app/Config/Feature.php b/app/Config/Feature.php
index 4c5ec90cd..480697b61 100644
--- a/app/Config/Feature.php
+++ b/app/Config/Feature.php
@@ -28,5 +28,5 @@ class Feature extends BaseConfig
     /**
      * Use improved new auto routing instead of the default legacy version.
      */
-    public bool $autoRoutesImproved = false;
+    public bool $autoRoutesImproved = true;
 }
```
```diff
diff --git a/app/Config/Routes.php b/app/Config/Routes.php
index ff2ac645c..78a76bc5c 100644
--- a/app/Config/Routes.php
+++ b/app/Config/Routes.php
@@ -25,7 +25,7 @@ $routes->set404Override();
 // where controller filters or CSRF protection are bypassed.
 // If you don't want to define all routes, please use the Auto Routing (Improved).
 // Set `$autoRoutesImproved` to true in `app/Config/Feature.php` and set the following to true.
-//$routes->setAutoRoute(false);
+$routes->setAutoRoute(true);
 
 /*
  * --------------------------------------------------------------------
```
```diff
diff --git a/app/Controllers/Home.php b/app/Controllers/Home.php
index 7f867e95f..3d34582e1 100644
--- a/app/Controllers/Home.php
+++ b/app/Controllers/Home.php
@@ -8,4 +8,8 @@ class Home extends BaseController
     {
         return view('welcome_message');
     }
+
+    public function getTest()
+    {
+    }
 }
```
and run `php spark routes`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Utilities/Routes/AutoRouterImproved/AutoRouteCollector.php
+++ b/system/Commands/Utilities/Routes/AutoRouterImproved/AutoRouteCollector.php
@@ -65,7 +65,7 @@ final class AutoRouteCollector
 
         foreach ($finder->find() as $class) {
             // Exclude controllers in Defined Routes.
-            if (in_array($class, $this->protectedControllers, true)) {
+            if (in_array('\\' . $class, $this->protectedControllers, true)) {
                 continue;
             }
 


system/Commands/Utilities/Routes/AutoRouterImproved/AutoRouteCollector.php
```diff
@@ -65,7 +65,7 @@ public function get(): array
 
         foreach ($finder->find() as $class) {
             // Exclude controllers in Defined Routes.
-            if (in_array($class, $this->protectedControllers, true)) {
+            if (in_array('\\' . $class, $this->protectedControllers, true)) {
```
MGatner, line 68: Would it make more sense to compare these as actual classes? Then we could rely more on autoloading and static analysis.

kenjis: What do you mean *to compare these as actual classes*?

MGatner: Something like `if (is_a($class, $protectedController, true)) ...`.

kenjis: > is_a — Checks if the object is of this class or has this class as one of its parents
https://www.php.net/manual/en/function.is-a.php

`or has this class as one of its parents` is not what we need.
It seems a good function for this does not exists.

MGatner: I assumed that would be a perk - do we ever want to allow extensions of a controller that is protected? If we only care about the class itself then `is_a()` isn't appropriate.

kenjis: When you have a controller A, and use it in a route definition.
You create `controller B extend A`, then do you expect to prohibit access to B via auto routing?

At least I did not expect that. 

It seems there is no good way to compare these as actual classes. 
So the current implementation in this PR seems to be okay.

MGatner (approved):

PR #6102, opened 2022-06-09
feat: Adding Later in Date Helper

ddevsr:
**Description**
Adding feature Later date in Date Helper

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/helpers/date_helper.rst
+++ b/user_guide_src/source/helpers/date_helper.rst
@@ -38,6 +38,27 @@ The following functions are available:
 
     If a timezone is not provided, it will return ``time()`` based on the
     **time_reference** setting.
+ 
+.. php:function:: later($dateSet = 'now', $amountLater = 1, $typeLater = 'D', $timezone = null)
+ 
+    :param string   $dateSet     j-n-Y or now
+    :param int      $amountLater Number
+    :param string   $typeLater   D/W/M/Y
+    :param string   $timezone    Timezone
+    :returns:    UNIX timestamp
+    :rtype:    int
+
+    Returns the current time as a UNIX timestamp, referenced either to your server's
+    local time or any PHP supported timezone, based on the "time reference" setting
+    in your config file. If you do not intend to set your master time reference to
+    any other PHP supported timezone (which you'll typically do if you run a site
+    that lets each user set their own timezone settings) there is no benefit to using
+    this function over PHP's ``time()`` function.
+
+    .. literalinclude:: date_helper/004.php
+
+    If a timezone is not provided, it will return ``time()`` based on the
+    **time_reference** setting.
 
 .. php:function:: timezone_select([$class = '', $default = '', $what = \DateTimeZone::ALL, $country = null])
 
--- a/user_guide_src/source/helpers/date_helper.rst
+++ b/user_guide_src/source/helpers/date_helper.rst
@@ -48,14 +48,13 @@ The following functions are available:
     :returns:    UNIX timestamp
     :rtype:    int
 
-    Returns the current time as a UNIX timestamp, referenced either to your server's
+    Returns the additional datetime as a UNIX timestamp, referenced either to your server's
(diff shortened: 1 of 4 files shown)


MGatner:
This is a very nice PR, but the provided feature is largely a functional duplicate of `CodeIgniter\I18n\Time` with more rigid parameters. I'm going to need to hear a good case as to what this provides that isn't in conflict with the single responsibility principle.

ddevsr:
So sorry, i make this function with more rigid parameter. Suggest me better parameter. In this case, company multi regional will be problem with exact date. So UNIX timestamp will be helpful. With one function `later` can getting later date (UNIX format) just one step.

In [`Codeigniter\I18n\Time`](https://codeigniter.com/user_guide/libraries/time.html#getters) :
1. `Time::parse('May 10, 2017', 'America/Chicago');` or `Time::now('America/Chicago','en_US');`
2. Modifier date in step 1 with [`$time->addYears(5);`](https://codeigniter.com/user_guide/libraries/time.html#modifying-the-value) or other addition
3. `$time->getTimestamp();` to get UNIX format

With `later` :
1. `later('now', 5, 'Y', 'America/Chicago')` or `later('May 10, 2017', 5, 'Y', 'America/Chicago')`

user_guide_src/source/helpers/date_helper.rst
```diff
@@ -38,6 +38,27 @@ The following functions are available:
...
+    local time or any PHP supported timezone, based on the "time reference" setting
+    in your config file. If you do not intend to set your master time reference to
+    any other PHP supported timezone (which you'll typically do if you run a site
+    that lets each user set their own timezone settings) there is no benefit to using
+    this function over PHP's ``time()`` function.
+
+    .. literalinclude:: date_helper/004.php
+
+    If a timezone is not provided, it will return ``time()`` based on the
+    **time_reference** setting.
```
kenjis, line 61: What is the `time_reference` setting?

kenjis:
I think we should carefully consider whether to add helper functions.

Once added, they are difficult to remove, and it is undesirable to bloat CodeIgniter 
by adding features that are rarely used.

user_guide_src/source/helpers/date_helper.rst
```diff
@@ -38,6 +38,23 @@ The following functions are available:
...
+    :param string   $dateSet     j-n-Y or now
+    :param int      $amountLater Number
+    :param string   $typeLater   D/W/M/Y
+    :param string   $timezone    Timezone
+    :returns:    UNIX timestamp
+    :rtype:    int
+
+    Returns the additional datetime as a UNIX timestamp, referenced either to your server's
+    local time or any PHP supported timezone, based on the "time reference" setting
+    in your config file.
```
kenjis, line 53: What is the "time reference" setting?

ddevsr: so sorry unremoved, check it again

MGatner:
@ddevsr I understand the convenience of this function but that does not by itself justify duplicating a feature. Developers can add their own helper files to their projects, so helper functions in the framework core should either 1) be used by the framework itself and thus necessary, or 2) provide a unique feature or mode of access to underlying features. 

Your examples when written side-by-side accomplish the same thing, but the class-based approach is much easier to understand without looking up function references:
```php
Time::parse('May 10, 2017', 'America/Chicago')->addYears(5)->getTimestamp();

later('now', 5, 'Y', 'America/Chicago') or later('May 10, 2017', 5, 'Y', 'America/Chicago');
```

Next steps for you... if `later()` is something you need across many projects you could consider releasing it as its own library, which would make adding it to projects a single Composer command. If you strongly believe this brings a much-needed feature then start a forum thread and get community feedback to demonstrate. 

Whatever comes of this, I want to thank you again for working up a very nice PR, and for your commitment to the framework.

MGatner:
@ddevsr has the next steps for pursuing this. Closing this PR.

PR #6105, opened 2022-06-10
feat: Adding validation `valid_jwt`

ddevsr:
**Description**
Adding new validator for everyone want validation token JWT.

Example JWT Token :
`eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -373,8 +373,6 @@ class FormatRules
 
                 return false;
             }
-            
-            return true;
         }
 
         return mb_strlen($segments[$index], '8bit') >= 32 && mb_strlen($segments[$index], '8bit') <= 64;
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -359,7 +359,7 @@ class FormatRules
         if (count($newString) === 3) {
             foreach ($newString as $value) {
                 $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
-                $isValid =  $encoded === $value;
+                $isValid = $encoded === $value;
                 
                 if (! $isValid) {
                     return false;
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -373,6 +373,6 @@ class FormatRules
             }
         }
 
-        return mb_strlen($segments[$index], '8bit') >= 32 && mb_strlen($segments[$index], '8bit') <= 64;
+        return (mb_strlen($segments[$index], '8bit') >= 32 && mb_strlen($segments[$index], '8bit') <= 64) || (mb_strlen($segments[$index], '8bit') >= 86 && mb_strlen($segments[$index], '8bit') <= 176) || mb_strlen($segments[$index], '8bit') === 342;
     }
 }
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -376,7 +376,7 @@ class FormatRules
         $strLength = mb_strlen($segments[$index], '8bit');
 
         return ($strLength >= 32 && $strLength <= 64)
-            || $strLength >= 86 && $strLength <= 176)
+            || $strLength >= 86 && $strLength <= 176
             || $strLength === 342;
     }
 }
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -373,6 +373,6 @@ class FormatRules
             }
         }
 
-        return true;
+        return mb_strlen($segments[$index], '8bit') >= 32 && mb_strlen($segments[$index], '8bit') <= 64;
     }
 }
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -370,8 +370,6 @@ class FormatRules
                 if ($encoded !== $value) {
                     return false;
                 }
-
-                return false;
             }
         }
 
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -362,11 +362,7 @@ class FormatRules
                     $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
                     $isValid = $encoded === $value;
 
-                    if (! $isValid) {
-                        return false;
-                    }
-                    
-                    return true;
+                    return ! (! $isValid);
                 }
 
                 return false;
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -370,6 +370,8 @@ class FormatRules
                 if ($encoded !== $value) {
                     return false;
                 }
+                
+                return false;
             }
             
             return true;
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -373,6 +373,10 @@ class FormatRules
             }
         }
 
-        return (mb_strlen($segments[$index], '8bit') >= 32 && mb_strlen($segments[$index], '8bit') <= 64) || (mb_strlen($segments[$index], '8bit') >= 86 && mb_strlen($segments[$index], '8bit') <= 176) || mb_strlen($segments[$index], '8bit') === 342;
+        $strLength = mb_strlen($segments[$index], '8bit');
+
+        return ($strLength >= 32 && $strLength <= 64)
+            || $strLength >= 86 && $strLength <= 176)
+            || $strLength === 342;
     }
 }
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -357,9 +357,15 @@ class FormatRules
         $newString = explode('.', $str);
 
         if (count($newString) === 3) {
-            foreach ($newString as $key => $value) {
-                return base64_encode(base64_decode($value, true)) === $value;
+            foreach ($newString as $value) {
+                $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
+                $isValid =  $encoded === $value;
+                if (! $isValid) {
+                    return false;
+                }
             }
+            
+            return true;
         }
 
         return false;
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -371,6 +371,8 @@ class FormatRules
                     return false;
                 }
             }
+            
+            return true;
         }
 
         return mb_strlen($segments[$index], '8bit') >= 32 && mb_strlen($segments[$index], '8bit') <= 64;
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -370,7 +370,7 @@ class FormatRules
                 if ($encoded !== $value) {
                     return false;
                 }
-                
+
                 return false;
             }
             
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -354,21 +354,21 @@ class FormatRules
      */
     public function valid_jwt(string $str = ''): bool
     {
-        $newString = explode('.', $str);
+        $segments = explode('.', $str);
 
-        if (count($newString) === 3) {
-            foreach ($newString as $index => $value) {
-                if ($index < 2) {
-                    $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
-                    $isValid = $encoded === $value;
+        if (count($segments) !== 3) {
+            return false;
+        }
 
-                    return ! (! $isValid);
+        foreach ($segments as $index => $value) {
+            if ($index < 2) {
+                $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
+                if ($encoded !== $value) {
+                    return false;
                 }
-
-                return false;
             }
         }
 
-        return false;
+        return true;
     }
 }
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -346,4 +346,22 @@ class FormatRules
 
         return $date !== false && $errors !== false && $errors['warning_count'] === 0 && $errors['error_count'] === 0;
     }
+
+    /**
+     * Token JWT Validator
+     * 
+     * @param string $str String
+     */
+    public function valid_token(string $str = ''): bool
+    {
+        $newString = explode('.', $str);
+
+        if (count($newString) === 3) {
+            foreach ($newString as $key => $value) {
+                return base64_encode(base64_decode($value, true)) === $value;
+            }
+        }
+
+        return false;
+    }
 }
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -349,7 +349,7 @@ class FormatRules
 
     /**
      * Token JWT Validator
-     * 
+     *
      * @param string $str String
      */
     public function valid_token(string $str = ''): bool
@@ -360,6 +360,7 @@ class FormatRules
             foreach ($newString as $value) {
                 $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
                 $isValid =  $encoded === $value;
+                
                 if (! $isValid) {
                     return false;
                 }
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -357,13 +357,17 @@ class FormatRules
         $newString = explode('.', $str);
 
         if (count($newString) === 3) {
-            foreach ($newString as $value) {
-                $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
-                $isValid = $encoded === $value;
-                
-                if (! $isValid) {
-                    return false;
+            foreach ($newString as $index => $value) {
+                if ($index < 2) {
+                    $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
+                    $isValid = $encoded === $value;
+
+                    if (! $isValid) {
+                        return false;
+                    }
                 }
+                
+                return false;
             }
             
             return true;
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -352,11 +352,15 @@ class FormatRules
      *
      * @param string $str String
      */
-    public function valid_jwt(string $str = ''): bool
+    public function valid_jwt(?string $str = null): bool
     {
+        if (empty($str)) {
+            return false;
+        }
+
         $segments = explode('.', $str);
 
-        if (count($segments) !== 3) {
+        if (count($segments) !== 3 || in_array('', $segments, true)) {
             return false;
         }
 
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -352,7 +352,7 @@ class FormatRules
      *
      * @param string $str String
      */
-    public function valid_token(string $str = ''): bool
+    public function valid_jwt(string $str = ''): bool
     {
         $newString = explode('.', $str);
 
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -358,15 +358,21 @@ class FormatRules
 
         if (count($newString) === 3) {
             foreach ($newString as $index => $value) {
-                if ($index < 2) {
-                    $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
-                    $isValid = $encoded === $value;
+                if (count($newString) === 3) {
+                    foreach ($newString as $index => $value) {
+                        if ($index < 2) {
+                            $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
+                            $isValid = $encoded === $value;
 
-                    if (! $isValid) {
-                        return false;
+                            if (! $isValid) {
+                                return false;
+                            }
+                        }
                     }
+
+                    return true;
                 }
-                
+
                 return false;
             }
             
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -358,25 +358,19 @@ class FormatRules
 
         if (count($newString) === 3) {
             foreach ($newString as $index => $value) {
-                if (count($newString) === 3) {
-                    foreach ($newString as $index => $value) {
-                        if ($index < 2) {
-                            $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
-                            $isValid = $encoded === $value;
+                if ($index < 2) {
+                    $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
+                    $isValid = $encoded === $value;
 
-                            if (! $isValid) {
-                                return false;
-                            }
-                        }
+                    if (! $isValid) {
+                        return false;
                     }
-
+                    
                     return true;
                 }
 
                 return false;
             }
-            
-            return true;
         }
 
         return false;
--- a/tests/system/Validation/FormatRulesTest.php
+++ b/tests/system/Validation/FormatRulesTest.php
@@ -80,43 +80,50 @@ final class FormatRulesTest extends CIUnitTestCase
         $this->assertFalse($this->validation->run($data));
     }
 
-    public function testValidJWTValid(): void
+    /**
+     * @dataProvider tokenJWTProvider
+     */
+    public function testValidJWT(?string $token, bool $expected): void
     {
         $data = [
-            'token' => 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
+            'token' => $token,
         ];
 
         $this->validation->setRules([
             'token' => 'valid_jwt',
         ]);
 
-        $this->assertTrue($this->validation->run($data));
+        $this->assertSame($expected, $this->validation->run($data));
     }
 
-    public function testValidJWTUnformatted(): void
+    public function tokenJWTProvider(): Generator
     {
-        $data = [
-            'token' => 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
+        yield from [
+            [
+                'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
+                true,
+            ],
+            [
+                'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.INVALID.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
+                false,
+            ],
+            [
+                'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
+                false,
+            ],
+            [
+                null,
+                false,
+            ],
+            [
+                '..',
+                false,
+            ],
+            [
+                'RGVubnk',
+                false,
+            ],
         ];
-
-        $this->validation->setRules([
-            'token' => 'valid_jwt',
-        ]);
-
-        $this->assertFalse($this->validation->run($data));
-    }
-
-    public function testValidJWTInvalidBase64(): void
-    {
-        $data = [
-            'token' => 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.INVALID.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
-        ];
-
-        $this->validation->setRules([
-            'token' => 'valid_jwt',
-        ]);
-
-        $this->assertFalse($this->validation->run($data));
     }
 
     /**
--- a/tests/system/Validation/FormatRulesTest.php
+++ b/tests/system/Validation/FormatRulesTest.php
@@ -131,56 +131,6 @@ final class FormatRulesTest extends CIUnitTestCase
     }
 
     /**
-     * @dataProvider tokenJWTProvider
-     */
-    public function testValidJWT(?string $token, bool $expected): void
-    {
-        $data = [
-            'token' => $token,
-        ];
-
-        $this->validation->setRules([
-            'token' => 'valid_jwt',
-        ]);
-
-        $this->assertSame($expected, $this->validation->run($data));
-    }
-
-    public function tokenJWTProvider(): Generator
-    {
-        yield from [
-            [
-                'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
-                true,
-            ],
-            [
-                'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.INVALID.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
-                false,
-            ],
-            [
-                'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.INVALID',
-                false,
-            ],
-            [
-                'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
-                false,
-            ],
-            [
-                null,
-                false,
-            ],
-            [
-                '..',
-                false,
-            ],
-            [
-                'RGVubnk',
-                false,
-            ],
-        ];
-    }
-
-    /**
      * @dataProvider urlProvider
      */
     public function testValidURL(?string $url, bool $isLoose, bool $isStrict)
--- a/tests/system/Validation/FormatRulesTest.php
+++ b/tests/system/Validation/FormatRulesTest.php
@@ -128,11 +128,11 @@ final class FormatRulesTest extends CIUnitTestCase
                 false,
             ],
             [
-                '..',
+                '..', // Manipulate format delimeter comma
                 false,
             ],
             [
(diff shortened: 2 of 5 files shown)


system/Validation/FormatRules.php
```diff
@@ -346,4 +346,22 @@ public function valid_date(?string $str = null, ?string $format = null): bool
...
+     * @param string $str String
+     */
+    public function valid_token(string $str = ''): bool
+    {
+        $newString = explode('.', $str);
+
+        if (count($newString) === 3) {
+            foreach ($newString as $key => $value) {
+                return base64_encode(base64_decode($value, true)) === $value;
+            }
```
samsonasik, line 362: I think it should check all data to be valid, not only one that found valid early:

```suggestion
            foreach ($newString as $value) {
                $isValid = base64_encode(base64_decode($value, true)) === $value;
                if (! $isValid) {
                    return false;
                }
            }
            
            return true;
```

samsonasik: also, `==` suffix seems need to be verified:

```bash
php > echo base64_encode(base64_decode('eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ', true));
eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ==
```

samsonasik: should be something like this:

```suggestion
            foreach ($newString as $value) {
                $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
                $isValid =  $encoded === $value;
                if (! $isValid) {
                    return false;
                }
            }
            
            return true;
```

tests/system/Validation/FormatRulesTest.php
```diff
@@ -80,6 +80,32 @@ public function testRegexMatchFalse(): void
...
+        $data = [
+            'token' => 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM',
+        ];
+
+        $this->validation->setRules([
+            'token' => 'valid_token',
+        ]);
+
+        $this->assertFalse($this->validation->run($data));
+    }
```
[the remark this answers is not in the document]
ddevsr: Okay added

system/Validation/FormatRules.php
```diff
@@ -346,4 +346,29 @@ public function valid_date(?string $str = null, ?string $format = null): bool
 
         return $date !== false && $errors !== false && $errors['warning_count'] === 0 && $errors['error_count'] === 0;
     }
+
+    /**
+     * Token JWT Validator
+     *
+     * @param string $str String
+     */
+    public function valid_token(string $str = ''): bool
```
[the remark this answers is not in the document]
ddevsr: Okay, will be renamed to `valid_jwt`

kenjis:
Does anyone understand JWT format correctly?

> Base64url Encoding
      Base64 encoding using the URL- and filename-safe character set
      defined in [Section 5 of RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648#section-5) [[RFC4648](https://datatracker.ietf.org/doc/html/rfc4648)], with all trailing '='
      characters omitted (as permitted by [Section 3.2](https://datatracker.ietf.org/doc/html/rfc7515#section-3.2)) and without the
      inclusion of any line breaks, whitespace, or other additional
      characters.  Note that the base64url encoding of the empty octet
      sequence is the empty string.  (See [Appendix C](https://datatracker.ietf.org/doc/html/rfc7515#appendix-C) for notes on
      implementing base64url encoding without padding.)
https://datatracker.ietf.org/doc/html/rfc7515#section-2

ddevsr:
> Does anyone understand JWT format correctly?
> 
> > Base64url Encoding
> > Base64 encoding using the URL- and filename-safe character set
> > defined in [Section 5 of RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648#section-5) [[RFC4648](https://datatracker.ietf.org/doc/html/rfc4648)], with all trailing '='
> > characters omitted (as permitted by [Section 3.2](https://datatracker.ietf.org/doc/html/rfc7515#section-3.2)) and without the
> > inclusion of any line breaks, whitespace, or other additional
> > characters.  Note that the base64url encoding of the empty octet
> > sequence is the empty string.  (See [Appendix C](https://datatracker.ietf.org/doc/html/rfc7515#appendix-C) for notes on
> > implementing base64url encoding without padding.)
> > https://datatracker.ietf.org/doc/html/rfc7515#section-2

In [JWT](https://jwt.io/), formated with 3 parameter

Example : `eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ.BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM`

`eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9` is base 64 as `header`
`eyJzbV91bmlxdWUiOjE2NTQ4NDc0MzcsInNtX2lwYWRkciI6IjEwMy44My4xNzMuMTc4Iiwic21fZW1haWwiOiJpdC4yZGV2QGFiYi5jb20ifQ` is base 64 as `payload`
`BEx-B52BzVaAt2E30X0Oiq97s53riONrqYxWiNp-ArM` is `HMACSHA256` as `signature`

Signature cant be validation because is unique with `secret key`. I originally made this to validate the format of the JWT token.

samsonasik:
PR title need to be updated to valid_jwt

system/Validation/FormatRules.php
```diff
@@ -346,4 +346,39 @@ public function valid_date(?string $str = null, ?string $format = null): bool
...
+     *
+     * @param string $str String
+     */
+    public function valid_jwt(string $str = ''): bool
+    {
+        $newString = explode('.', $str);
+
+        if (count($newString) === 3) {
+            foreach ($newString as $index => $value) {
+                if (count($newString) === 3) {
```
samsonasik, line 361: duplication check and foreach is not needed

iRedds:
Thanks for trying to make CI better.
But I have one question. For what?
For example, I want to use JWT for an application. The simplest solution is to take a ready-made library, in which both generation and validation are already implemented.
OK. I want to write my own library because I don't trust. To validate the token, I will need to decode base64 anyway.

About the code. In fact, it checks if the encoded message contains characters that are not in the base64 alphabet.
Therefore, the token containing 1.1.1 will successfully pass the verification. Because it checks that ``base64_encode(1) === base64_encode(1)``; But 1.1.1 is not a valid JWT token.

system/Validation/FormatRules.php
```diff
@@ -346,4 +346,29 @@ public function valid_date(?string $str = null, ?string $format = null): bool
...
+    {
+        $newString = explode('.', $str);
+
+        if (count($newString) === 3) {
+            foreach ($newString as $index => $value) {
+                if ($index < 2) {
+                    $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
+                    $isValid = $encoded === $value;
+
+                    return ! (! $isValid);
```
samsonasik, line 365: this direct return will make next check by passed, better to check return early on non-valid.

system/Validation/FormatRules.php
```diff
@@ -346,4 +346,29 @@ public function valid_date(?string $str = null, ?string $format = null): bool
...
+                }
+
+                return false;
+            }
+        }
+
+        return false;
```
MGatner, line 372: Here's what I think @samsonasik is suggesting:
```suggestion
        $segments = explode('.', $str);

        if (count($segments) !== 3) {
            return false;
        }

        foreach ($segments as $index => $value) {
            if ($index < 2) {
                $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
                if ($encoded !== $value) {
                    return false;
                }
            }
        }

        return true;
```

system/Validation/FormatRules.php
```diff
@@ -346,4 +346,29 @@ public function valid_date(?string $str = null, ?string $format = null): bool
...
+     *
+     * @param string $str String
+     */
+    public function valid_jwt(string $str = ''): bool
+    {
+        $newString = explode('.', $str);
+
+        if (count($newString) === 3) {
+            foreach ($newString as $index => $value) {
+                if ($index < 2) {
```
MGatner, line 361: Verifying the signature is of course beyond the scope of this rule but I think we should at least check that it is non-empty.

tests/system/Validation/FormatRulesTest.php
```diff
@@ -80,6 +80,45 @@ public function testRegexMatchFalse(): void
         $this->assertFalse($this->validation->run($data));
     }
 
+    public function testValidJWTValid(): void
```
MGatner, line 83: Thank you for including tests! I think there are a lot more "edge cases" to check. I would like to see these tests refactored with a data provider. We have lots of examples you could mimic in here already, but basically it would be an array of `[string $tokenToTest, bool $expectedOutcome]`. This way you can test lots of input in one pass and it is easy to see what cases are covered.

user_guide_src/source/libraries/validation.rst
```diff
@@ -686,6 +686,8 @@ valid_url_strict        Yes        Fails if field does not contain a valid URL.
 valid_date              No         Fails if field does not contain a valid date. valid_date[d/m/Y]
                                    Accepts an optional parameter to matches
                                    a date format.
+valid_jwt               No         Fails if field contains anything other than
+                                   JWT Format.
```
MGatner, line 690: I know it seems silly, but I think we should add a note like: "Note: this does not verify the token itself, just the string format". Because people don't look at things sometimes. Or think.

MGatner (changes requested):
I think this is a good addition but needs to be handled carefully. Please see individual comments.

MGatner:
This should be pointed to `4.3` as well.

ddevsr:
@MGatner already added new list test token JWT

system/Validation/FormatRules.php
```diff
@@ -346,4 +346,33 @@ public function valid_date(?string $str = null, ?string $format = null): bool
...
+        foreach ($segments as $index => $value) {
+            if ($index < 2) {
+                $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
+                if ($encoded !== $value) {
+                    return false;
+                }
+            }
+        }
+
+        return true;
```
MGatner, line 376: Is there anything else we can check about the third segment? Minimum signature length, or a regex of characters?

MGatner (changes requested):

paulbalandan:
I agree with @iRedds here.

Validating JWT is a security concern and there's even an [RFC](https://datatracker.ietf.org/doc/html/rfc7519) for that. The code claims to be checking for "valid JWT" although it is only "touching the surface" (i.e., validating only the string format).

> ## Manually implement checks
> We discourage doing manual JWT validation since it might be easy to improperly implement and miss some important details that will lead to serious security vulnerabilities. Most JWT libraries take care of JWT validation for you. Visit JWT.io to find a JWT library for your platform and programming language.

>
> For instructions on how to manually validate a JWT, see [RFC 7519](https://tools.ietf.org/html/rfc7519#section-7.2). All Auth0-issued JWTs have a JSON Web Signature (JWS), meaning they are signed rather than encrypted.

https://auth0.com/docs/secure/tokens/json-web-tokens/validate-json-web-tokens

The use case here is to have a preliminary validation of the token before passing it to the third-party library. My question is: Why? Can't the TP library do the validation itself? At least with `Firebase\JWT\JWT::decode()`, it throws an `UnexpectedValueException` if it encounters an invalid JWT token which I believe is more secure.

kenjis:
@paulbalandan What's the TP library? What's TP?

system/Validation/FormatRules.php
ddevsr: In algo SHA256-SHA512 & Sodium, get min `32` and max `64` byte.

Ref : [firebase/php-jwt](https://github.com/firebase/php-jwt/blob/main/src/JWT.php#L55) & [JWT.io](https://jwt.io/libraries?language=PHP)

paulbalandan:
> @paulbalandan What's the TP library? What's TP?

Sorry, that's for 'third-party'. Sort of acronym

kenjis:
Oh, I got it. Thanks.

kenjis:
I think generally multiple validations are useful and good practice.
When there is invalid or evil input, it is better to reject as soon as possible. 
Because computer power is not infinite, and processing invalid data takes more security risks.

Order of input validation is:
1. Origin
2. Size
3. Lexical (characters) 
4. Syntactical (words, tokens) 
5. Semantical (makes sense, logic)

@ddevsr 
> In my case real project. I create new Rules for validate JWT format before validate token jwt is token from encode JWT with Library. 

Can you tell us why do you need this rule to validate JWT format?
Why don't you use `Firebase\JWT\JWT::decode()` directly?

ddevsr:
> @ddevsr
> 
> Can you tell us why do you need this rule to validate JWT format? Why don't you use `Firebase\JWT\JWT::decode()` directly?

The most people use library `Firebase\JWT\JWT`, reason to make this PR is want represent validation behaviorally format all library PHP JWT.

kenjis:
This PR's base branch is `4.3`, not `develop`.

```console
git fetch upstream
git switch token-validator
git rebase upstream/4.3
```

kenjis:
We have changed the branch model recently.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#branching

kenjis:
Can't you rebase this branch?
We can't review this PR.

system/Validation/FormatRules.php
```diff
@@ -346,4 +346,33 @@ public function valid_date(?string $str = null, ?string $format = null): bool
...
+        foreach ($segments as $index => $value) {
+            if ($index < 2) {
+                $encoded = rtrim(base64_encode(base64_decode($value, true)), '==');
+                if ($encoded !== $value) {
+                    return false;
+                }
+            }
+        }
+
+        return mb_strlen($segments[$index], '8bit') >= 32 && mb_strlen($segments[$index], '8bit') <= 64;
```
MGatner, line 376: Thanks for adding this! Please include test cases that fail both the minimum and maximum lengths.

tests/system/Validation/FormatRulesTest.php
```diff
@@ -80,6 +80,64 @@ public function testRegexMatchFalse(): void
...
+                false,
+            ],
+            [
+                '..',
+                false,
+            ],
+            [
+                'RGVubnk',
+                false,
+            ],
```
kenjis, line 137: Can you add indexes to explain what's testing?

system/Validation/FormatRules.php
kenjis, line 376: Why is the length 32 to 64?

ddevsr: In algo SHA256-SHA512 & Sodium, get min 32 and max 64 byte.

Ref : [firebase/php-jwt](https://github.com/firebase/php-jwt/blob/main/src/JWT.php#L55) & [JWT.io](https://jwt.io/libraries?language=PHP)

MGatner (approved):

system/Validation/FormatRules.php
kenjis: Is this not valid?
```
eyJhbGciOiJSUzUxMiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlhdCI6MTUxNjIzOTAyMn0.jYW04zLDHfR1v7xdrW3lCGZrMIsVe0vWCfVkN2DRns2c3MN-mcp_-RE6TN9umSBYoNV-mnb31wFf8iun3fB6aDS6m_OXAiURVEKrPFNGlR38JSHUtsFzqTOj-wFrJZN4RwvZnNGSMvK3wzzUriZqmiNLsG8lktlEn6KA4kYVaM61_NpmPHWAjGExWv7cjHYupcjMSmR8uMTwN5UuAwgW6FRstCJEfoxwb0WKiyoaSlDuIiHZJ0cyGhhEmmAPiCwtPAwGeaL1yZMcp0p82cpTQ5Qb-7CtRov3N4DcOHgWYk6LomPR5j5cCkePAz87duqyzSMpCB0mCOuE3CU2VMtGeQ
```

tests/system/Validation/FormatRulesTest.php
```diff
@@ -80,6 +80,64 @@ public function testRegexMatchFalse(): void
...
+            [
+                null,
+                false,
+            ],
+            [
+                '..', // Manipulate format delimeter comma
+                false,
+            ],
```
kenjis, line 133: I mean like this:
```suggestion
            'Manipulate format delimeter comma' => [
                '..',
                false,
            ],
```

system/Validation/FormatRules.php
ddevsr: Ohh sorry, already added length for RS, ES and PS

Sample :
`
$str = array(
    'jYW04zLDHfR1v7xdrW3lCGZrMIsVe0vWCfVkN2DRns2c3MN-mcp_-RE6TN9umSBYoNV-mnb31wFf8iun3fB6aDS6m_OXAiURVEKrPFNGlR38JSHUtsFzqTOj-wFrJZN4RwvZnNGSMvK3wzzUriZqmiNLsG8lktlEn6KA4kYVaM61_NpmPHWAjGExWv7cjHYupcjMSmR8uMTwN5UuAwgW6FRstCJEfoxwb0WKiyoaSlDuIiHZJ0cyGhhEmmAPiCwtPAwGeaL1yZMcp0p82cpTQ5Qb-7CtRov3N4DcOHgWYk6LomPR5j5cCkePAz87duqyzSMpCB0mCOuE3CU2VMtGeQ',
    'SefW4LRE8rlUTjvGXEuC5oqqW0jsjaUR9pBnTt9Jarp3EjDo3v_djd8nN9Cwa_eFOvxkF61PERv08DiIb7DU1ceGOVhh49btyjAe5hxqo_fKS0c95PNV4ANgv9ca3ZZ1Df2i3dSRRmo-9yDg54QIgWFrOGu8t2bgbZfVnOv_qSJdagMa9QxhgAqRJd9Jrz0qEMLYPuLdIn-5XHYk07pruFhHgjcycGXYjwYSfH0kOke2K6NsX3VAjsHgU5GLVU9J9XCnyP5lx2cgk3IWF9y03fykigh820hqHWs28dwv-BHcMscumOvlAGW6YrnQqjreNtaLIwqs7LdjlHnqc5ybIQ',
    'Sv4y1O38_Pj9JMpWz_dn7Ms1LJBVGw-TId6a0lwTn0coTXo_JAbjIAogiMya8mvhF0nIkt7lF0WHSgL5kBuXtODrwPHbQPOtLeumkJ2N2sRDsN0LP0m_0QN7gCEkFSZ8FXukaPM8NqJsoNXEpeNLPzDoACNNtRZj5JiNwnxtpQt1CMunzUSeo_Cul7wHojlMy-Y5duxXnWTt_4rVVTspUjWcxBQxdRM1QQ_mP3eHN2ABVcBdkrRDK-szZn5RoHA6nPfKTZ51QX3UPT5eGJb1MsQGSKC2vlSi-TPP8snXCU7SqdyXC6rchLXtH_pKWwe4vDFwOcxhIgP3XzKTEO7nAw',
    'QrOYZgxP7_DfAuMnppKVPPtCMrxVIwJFPUe47-Cnw_of50pEoZ5M57utxS78FAq48UnYyu76nMe2CKfjhj0LvI3WIbhTnrxdrs6tQiJZHs-j2qwcHRCfSFlqKN1HuOWDAXnsQd1CzKa597urll3MxdeMIN3PKFaWCUVozlrodr_qqKDOxeCpl64mPfCsKb22U_SYKn5kQLg9qPjPGayXnCF6iZiOAahN9PqzjWIjEe0lkJTySS9KJUPpSKhVjRuwabELsW4DDtp6Q9KCVZQUw5VwJVWm9vqhRUTfbPoC9WViOIYvBnjOf_eYMFm1F2y7Ee2dnj4XBI7gg8_tTX-SJQ',
    'ufxA9G1842f7K5tmDrbGh2gljMI9wNcspQjei0U7xipJwGBopUlc7HC-83IMEcnPLluOfB11GnJIdXUa-JC9VQ1klHvjMJ4t0rQKyPqd_BgJ5beuyjBmpCXkE5uR-4KolLZKlwCzPll8TzpRvMSQiamx1c6MqQmP2Vy-gOGfi-LTNNZhcr4lmyEVc73okFZAAB0nbnZ-GLN7qRc0ZtBLjOdR-R60yhihCw_nEKP3auvKnzi7hC_DwdT8bxkbcHAEE4AeyvERUm6dleoMqax55qLO2sHQ8ZnnsP8eUFDGvDn--rLyNIa0jklNW5LWLVIF3gmmpTOyQ2wldElJqmllhQ',
    'W7Sb3oA56Cp_sdRLQ7UzUiPlGqOZr24UVzY7ZB84MuZHM-TnDgi9QJTE5fX87JM6Q-_TyDjI5DY0l1CCoHIohbjPJTZAhmfyJfbeMsQH7nZv0hGlzRIUA_5bpLjHQPQk_X5J-_9RvCe5wFUq9d1h3Uvz_UTNLoF2j26nENFdDJ2JTsJ0OH3ihsQ5LYPAcsNtTET0gFSJWfBClqfrFbPcGe1Y54sMvBuf8fNMhhemPDIP3HGTpbSY-OmQ-jcvLp30ld38LQEklBReI5cS5J9vpST0fgkZ02v-Ex-s1Yt6TErldBOpjEAhVCMTciFLepp1kzyNO0AhlfWXl7QoEsSfWQ',
    'CUdWRfjVbbDnZNLyuJjgiEr85ngGZ4DXjKDhGCKX5pH2NVkNwIwM1y9zw9FN2MiSFG8tzWdjCBxkVSmGzqhFCkz65Gx5OffnD57iHy0cc56bgGfNq-WHlqvAUiAJX3HzJ5MRn2HDZ4eGm0bmBQDhB91qvDpNIyIaLHgxWl8Brmn-tLuxIt4B1kpKKFy-crCeAsdHkZABFvGmtNeelH_XpJWc-5DKTd6JtpIhpPrsUKd8iFvWYJkc9Vp1k3OXTZARzuJ7qkt2xCBC38wKU4U_KXYmsJ5fAUuXWwk4AoirYlGalLLXIppg3T_Lg8AGKrJpM1tXD-GbZyK4tL1M2O2lSg',
    'jYW04zLDHfR1v7xdrW3lCGZrMIsVe0vWCfVkN2DRns2c3MN-mcp_-RE6TN9umSBYoNV-mnb31wFf8iun3fB6aDS6m_OXAiURVEKrPFNGlR38JSHUtsFzqTOj-wFrJZN4RwvZnNGSMvK3wzzUriZqmiNLsG8lktlEn6KA4kYVaM61_NpmPHWAjGExWv7cjHYupcjMSmR8uMTwN5UuAwgW6FRstCJEfoxwb0WKiyoaSlDuIiHZJ0cyGhhEmmAPiCwtPAwGeaL1yZMcp0p82cpTQ5Qb-7CtRov3N4DcOHgWYk6LomPR5j5cCkePAz87duqyzSMpCB0mCOuE3CU2VMtGeQ', // Sample kenjis
    'tyh-VfuzIxCyGYDlkBA7DfyjrqmSHu6pQ2hoZuFqUSLPNY2N0mpHb3nk5K17HWP_3cYHBw7AhHale5wky6-sVA', // ES256
    'VUPWQZuClnkFbaEKCsPy7CZVMh5wxbCSpaAWFLpnTe9J0--PzHNeTFNXCrVHysAa3eFbuzD8_bLSsgTKC8SzHxRVSj5eN86vBPo_1fNfE7SHTYhWowjY4E_wuiC13yoj', // ES384
    'AbVUinMiT3J_03je8WTOIl-VdggzvoFgnOsdouAs-DLOtQzau9valrq-S6pETyi9Q18HH-EuwX49Q7m3KC0GuNBJAc9Tksulgsdq8GqwIqZqDKmG7hNmDzaQG1Dpdezn2qzv-otf3ZZe-qNOXUMRImGekfQFIuH_MjD2e8RZyww6lbZk', // ES512
    'iOeNU4dAFFeBwNj6qdhdvm-IvDQrTa6R22lQVJVuWJxorJfeQww5Nwsra0PjaOYhAMj9jNMO5YLmud8U7iQ5gJK2zYyepeSuXhfSi8yjFZfRiSkelqSkU19I-Ja8aQBDbqXf2SAWA8mHF8VS3F08rgEaLCyv98fLLH4vSvsJGf6ueZSLKDVXz24rZRXGWtYYk_OYYTVgR1cg0BLCsuCvqZvHleImJKiWmtS0-CymMO4MMjCy_FIl6I56NqLE9C87tUVpo1mT-kbg5cHDD8I7MjCW5Iii5dethB4Vid3mZ6emKjVYgXrtkOQ-JyGMh6fnQxEFN1ft33GX2eRHluK9eg', // PS256
);

foreach ($str as $key => $val){
    echo mb_strlen($val).PHP_EOL;
}
`

kenjis: Please use ` ```php ` and ` ``` ` for code block.

ddevsr:
```
$str = array(
    'jYW04zLDHfR1v7xdrW3lCGZrMIsVe0vWCfVkN2DRns2c3MN-mcp_-RE6TN9umSBYoNV-mnb31wFf8iun3fB6aDS6m_OXAiURVEKrPFNGlR38JSHUtsFzqTOj-wFrJZN4RwvZnNGSMvK3wzzUriZqmiNLsG8lktlEn6KA4kYVaM61_NpmPHWAjGExWv7cjHYupcjMSmR8uMTwN5UuAwgW6FRstCJEfoxwb0WKiyoaSlDuIiHZJ0cyGhhEmmAPiCwtPAwGeaL1yZMcp0p82cpTQ5Qb-7CtRov3N4DcOHgWYk6LomPR5j5cCkePAz87duqyzSMpCB0mCOuE3CU2VMtGeQ',
    'SefW4LRE8rlUTjvGXEuC5oqqW0jsjaUR9pBnTt9Jarp3EjDo3v_djd8nN9Cwa_eFOvxkF61PERv08DiIb7DU1ceGOVhh49btyjAe5hxqo_fKS0c95PNV4ANgv9ca3ZZ1Df2i3dSRRmo-9yDg54QIgWFrOGu8t2bgbZfVnOv_qSJdagMa9QxhgAqRJd9Jrz0qEMLYPuLdIn-5XHYk07pruFhHgjcycGXYjwYSfH0kOke2K6NsX3VAjsHgU5GLVU9J9XCnyP5lx2cgk3IWF9y03fykigh820hqHWs28dwv-BHcMscumOvlAGW6YrnQqjreNtaLIwqs7LdjlHnqc5ybIQ',
    'Sv4y1O38_Pj9JMpWz_dn7Ms1LJBVGw-TId6a0lwTn0coTXo_JAbjIAogiMya8mvhF0nIkt7lF0WHSgL5kBuXtODrwPHbQPOtLeumkJ2N2sRDsN0LP0m_0QN7gCEkFSZ8FXukaPM8NqJsoNXEpeNLPzDoACNNtRZj5JiNwnxtpQt1CMunzUSeo_Cul7wHojlMy-Y5duxXnWTt_4rVVTspUjWcxBQxdRM1QQ_mP3eHN2ABVcBdkrRDK-szZn5RoHA6nPfKTZ51QX3UPT5eGJb1MsQGSKC2vlSi-TPP8snXCU7SqdyXC6rchLXtH_pKWwe4vDFwOcxhIgP3XzKTEO7nAw',
    'QrOYZgxP7_DfAuMnppKVPPtCMrxVIwJFPUe47-Cnw_of50pEoZ5M57utxS78FAq48UnYyu76nMe2CKfjhj0LvI3WIbhTnrxdrs6tQiJZHs-j2qwcHRCfSFlqKN1HuOWDAXnsQd1CzKa597urll3MxdeMIN3PKFaWCUVozlrodr_qqKDOxeCpl64mPfCsKb22U_SYKn5kQLg9qPjPGayXnCF6iZiOAahN9PqzjWIjEe0lkJTySS9KJUPpSKhVjRuwabELsW4DDtp6Q9KCVZQUw5VwJVWm9vqhRUTfbPoC9WViOIYvBnjOf_eYMFm1F2y7Ee2dnj4XBI7gg8_tTX-SJQ',
    'ufxA9G1842f7K5tmDrbGh2gljMI9wNcspQjei0U7xipJwGBopUlc7HC-83IMEcnPLluOfB11GnJIdXUa-JC9VQ1klHvjMJ4t0rQKyPqd_BgJ5beuyjBmpCXkE5uR-4KolLZKlwCzPll8TzpRvMSQiamx1c6MqQmP2Vy-gOGfi-LTNNZhcr4lmyEVc73okFZAAB0nbnZ-GLN7qRc0ZtBLjOdR-R60yhihCw_nEKP3auvKnzi7hC_DwdT8bxkbcHAEE4AeyvERUm6dleoMqax55qLO2sHQ8ZnnsP8eUFDGvDn--rLyNIa0jklNW5LWLVIF3gmmpTOyQ2wldElJqmllhQ',
    'W7Sb3oA56Cp_sdRLQ7UzUiPlGqOZr24UVzY7ZB84MuZHM-TnDgi9QJTE5fX87JM6Q-_TyDjI5DY0l1CCoHIohbjPJTZAhmfyJfbeMsQH7nZv0hGlzRIUA_5bpLjHQPQk_X5J-_9RvCe5wFUq9d1h3Uvz_UTNLoF2j26nENFdDJ2JTsJ0OH3ihsQ5LYPAcsNtTET0gFSJWfBClqfrFbPcGe1Y54sMvBuf8fNMhhemPDIP3HGTpbSY-OmQ-jcvLp30ld38LQEklBReI5cS5J9vpST0fgkZ02v-Ex-s1Yt6TErldBOpjEAhVCMTciFLepp1kzyNO0AhlfWXl7QoEsSfWQ',
    'CUdWRfjVbbDnZNLyuJjgiEr85ngGZ4DXjKDhGCKX5pH2NVkNwIwM1y9zw9FN2MiSFG8tzWdjCBxkVSmGzqhFCkz65Gx5OffnD57iHy0cc56bgGfNq-WHlqvAUiAJX3HzJ5MRn2HDZ4eGm0bmBQDhB91qvDpNIyIaLHgxWl8Brmn-tLuxIt4B1kpKKFy-crCeAsdHkZABFvGmtNeelH_XpJWc-5DKTd6JtpIhpPrsUKd8iFvWYJkc9Vp1k3OXTZARzuJ7qkt2xCBC38wKU4U_KXYmsJ5fAUuXWwk4AoirYlGalLLXIppg3T_Lg8AGKrJpM1tXD-GbZyK4tL1M2O2lSg',
    'jYW04zLDHfR1v7xdrW3lCGZrMIsVe0vWCfVkN2DRns2c3MN-mcp_-RE6TN9umSBYoNV-mnb31wFf8iun3fB6aDS6m_OXAiURVEKrPFNGlR38JSHUtsFzqTOj-wFrJZN4RwvZnNGSMvK3wzzUriZqmiNLsG8lktlEn6KA4kYVaM61_NpmPHWAjGExWv7cjHYupcjMSmR8uMTwN5UuAwgW6FRstCJEfoxwb0WKiyoaSlDuIiHZJ0cyGhhEmmAPiCwtPAwGeaL1yZMcp0p82cpTQ5Qb-7CtRov3N4DcOHgWYk6LomPR5j5cCkePAz87duqyzSMpCB0mCOuE3CU2VMtGeQ', // Sample kenjis
    'tyh-VfuzIxCyGYDlkBA7DfyjrqmSHu6pQ2hoZuFqUSLPNY2N0mpHb3nk5K17HWP_3cYHBw7AhHale5wky6-sVA', // ES256
    'VUPWQZuClnkFbaEKCsPy7CZVMh5wxbCSpaAWFLpnTe9J0--PzHNeTFNXCrVHysAa3eFbuzD8_bLSsgTKC8SzHxRVSj5eN86vBPo_1fNfE7SHTYhWowjY4E_wuiC13yoj', // ES384
    'AbVUinMiT3J_03je8WTOIl-VdggzvoFgnOsdouAs-DLOtQzau9valrq-S6pETyi9Q18HH-EuwX49Q7m3KC0GuNBJAc9Tksulgsdq8GqwIqZqDKmG7hNmDzaQG1Dpdezn2qzv-otf3ZZe-qNOXUMRImGekfQFIuH_MjD2e8RZyww6lbZk', // ES512
    'iOeNU4dAFFeBwNj6qdhdvm-IvDQrTa6R22lQVJVuWJxorJfeQww5Nwsra0PjaOYhAMj9jNMO5YLmud8U7iQ5gJK2zYyepeSuXhfSi8yjFZfRiSkelqSkU19I-Ja8aQBDbqXf2SAWA8mHF8VS3F08rgEaLCyv98fLLH4vSvsJGf6ueZSLKDVXz24rZRXGWtYYk_OYYTVgR1cg0BLCsuCvqZvHleImJKiWmtS0-CymMO4MMjCy_FIl6I56NqLE9C87tUVpo1mT-kbg5cHDD8I7MjCW5Iii5dethB4Vid3mZ6emKjVYgXrtkOQ-JyGMh6fnQxEFN1ft33GX2eRHluK9eg', // PS256
    'J5W09-rNx0pt5_HBiydR-vOluS6oD-RpYNa8PVWwMcBDQSXiw6-EPW8iSsalXPspGj3ouQjAnOP_4-zrlUUlvUIt2T79XyNeiKuooyIFvka3Y5NnGiOUBHWvWcWp4RcQFMBrZkHtJM23sB5D7Wxjx0-HFeNk-Y3UJgeJVhg5NaWXypLkC4y0ADrUBfGAxhvGdRdULZivfvzuVtv6AzW6NRuEE6DM9xpoWX_4here-yvLS2YPiBTZ8xbB3axdM99LhES-n52lVkiX5AWg2JJkEROZzLMpaacA_xlbUz_zbIaOaoqk8gB5oO7kI6sZej3QAdGigQy-hXiRnW_L98d4GQ', // PS512
);

foreach ($str as $key => $val){
    echo mb_strlen($val).PHP_EOL;
}
```

tests/system/Validation/FormatRulesTest.php
ddevsr: Already added index of rule test

lonnieezell:
I realize I'm very late coming to the party here, but this seems like a very strange addition to CI core. I'm having an issue trying to determine when anyone is going to use this? Anyone using JWT is either using an off the shelf solution, or they're rolling their own library. If they're using an off the shelf solution then it would be best for them to create a think wrapper around the library's functionality. 

To offer something that validates it's in the proper format but doesn't actually validate the token, seems like a part of a bigger solution, and, as a developer, I'd be looking around CI docs wondering where the rest of it is. 

Now, if this were to be included as part of a more complete JWT solution in Shield, then that at least seems like the proper place and usage for this.

sfadschm:
Have Pauls comments been discussed? I feel they are valid concerns (although I am not working with jwt myself).

Also I am not sure why we need to "bloat" the CI core here when this functionality seems more like a task for a module/extension or third-party solution (and apparently, such already exist).

kenjis:
> Also I am not sure why we need to "bloat" the CI core here when this functionality seems more like a task for a module/extension or third-party solution (and apparently, such already exist).

I don't think this validation rule is worthless, but I certainly question whether it should be included in **CI Core**.

@ddevsr 
> The most people use library Firebase\JWT\JWT, reason to make this PR is want represent validation behaviorally format all library PHP JWT.

I don't understand your use case well. Can you elaborate?

ddevsr:
In case, create
- Backend (API)
- Frontend (Web)

Web send request to API and response result `$payload` only for web to be changed the `token` by `payload` & `ip_web` and send `token` to API. Its JWT Token become a legitimate tool on web activity. And every activity on web must be check to endpoint API e.g `checkToken` for validation `token` and `ip_web` matching in database (API).

I think if `token` must validation with `Firebase::JWT` or other, on Web & API can addition time right? and must be same `SECRET_KEY`. API Services do not want know `SECRET_KEY`, every website that uses this API will use its own secret key.

kenjis:
@ddevsr Thank you, but I don't get what you say.

In my uderstanding:

Client --(1)--> Web --(2)--> API

(1) is JWT, and Web checks the JWT,
and (2) is not JWT?

ddevsr:
> In my uderstanding:
> 
> Client --(1)--> Web --(2)--> API
> 
> (1) is JWT, and Web checks the JWT, and (2) is not JWT?

Sorry for taking so long to reply.

In my case,
1. Users send cookies to Web
2. Web encrypt payload of cookies to JWT Token and send token to API
3. API check token and save token to database

Web & API Services is B2B connection to selling services API to Users directly.  Token/SecretKey used by Users to API Services and Web get commision by activity users.  If this function will be very rarely used, I close first and if needed will be opened again.

kenjis:
Then the situation is like this?

User --(1)--> Web --(2)--> API

(2) is JWT.

I don't know why you don't validate JWT with Firebase::JWT (or other library) on API.
I think you should verify the signature of the JWT after validating it with this `valid_jwt`.

PR #6107, opened 2022-06-10
Add upsert() and upsertBatch() to Builder

sclubricants:
**Description**
As a new feature I've created a new method for insert/update in batch mode. I used the replace naming convention however when studying the REPLACE INTO vs INSERT.. ON DUPLICATE KEY UPDATE sql statements I found that because REPLACE INTO causes a DELETE this can affect foreign key constraints and cause issues. If there is a cascade delete this would remove related records when calling REPLACE. It might  be worth changing the replace method as well? Alternatively I thought of naming the method insertUpdateBatch(). Else, if you guys would rather stick with the REPLACE statement I can change to use this.

Let me know what you guys think!

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


sclubricants:
Here is one example of the issues with REPLACE

https://stackoverflow.com/questions/44886066/mysql-replace-into-behavior-with-unique-constraint

sclubricants:
Another option is to handle things like ignore.

Then:

$builder->onDuplicateKeyUpdate(true)->insert($data);

$builder->onDuplicateKeyUpdate(true)->insertBatch($data);

And we could also have replaceBatch using the REPLACE INTO statement.

kenjis:
Thank you for trying to make CI better.

Unfortunately, `ON DUPLICATE KEY UPDATE` seems to be MySQL specific.
Query Builder is not only for MySQL, but also for other supported databases.

This PR needs to support all supported databases.

> https://codeigniter4.github.io/CodeIgniter4/intro/requirements.html
        MySQL via the MySQLi driver (version 5.1 and above only)
        PostgreSQL via the Postgre driver
        SQLite3 via the SQLite3 driver
        MSSQL via the SQLSRV driver (version 2005 and above only)
        Oracle via the OCI8 driver (version 12.1 and above only)

tests/system/Database/Builder/ReplaceTest.php
```diff
@@ -53,4 +54,36 @@ public function testReplaceThrowsExceptionWithNoData()
 
         $builder->replace();
     }
+
+    public function testReplaceBatch()
```
kenjis, line 58: I think a live test is also needed to test all supported databases.

kenjis (changes requested):
There is also coding style violation. Please fix.

tests/system/Database/Builder/ReplaceTest.php
sclubricants: This would rule out using ON DUPLICATE KEY UPDATE with INSERT but it seems REPLACE is a MySQL extention.

This shouldn't affect the replace() method if it only works with MySQL to begin with.

https://dev.mysql.com/doc/refman/8.0/en/replace.html

PostgreSQL uses a similar ON CONFLICT DO UPDATE 
https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-upsert/

I guess the proper term here would be UPSERT

kenjis: > I guess the proper term here would be UPSERT

Good point. REPLACE and UPSERT are different.

REPLACE = INSERT or (DELETE + INSERT)
UPSERT = INSERT or UPDATE

sclubricants:
I'll have to see what it takes to implement in the other drivers. I forgot about those..

tests/system/Database/Builder/ReplaceTest.php
sclubricants: > > I guess the proper term here would be UPSERT
> 
> Good point. REPLACE and UPSERT are different.
> 
> REPLACE = INSERT or (DELETE + INSERT) UPSERT = INSERT or UPDATE

Looking at the Postgre implementation of replace() they use a second query to see if the record exists and if it does then it updates it, otherwise inserts it. This would be an upsert. For that matter they should have used ON CONFLICT DO UPDATE.
https://www.prisma.io/dataguide/postgresql/inserting-and-modifying-data/insert-on-conflict

Personally I'd prefer the UPSERT operation to the REPLACE. The current implementation has muddied the difference.  

I currently have a need to upsert 1700 records at a time pretty regularly and was looking to give CI this functionality. I never use MySQL's REPLACE so I got looking into the difference and decided it might cause too many problems. 

I've used Soql's upsert function a lot as well. With Soql you run an primary or external key. The nice thing though with mysql is that it runs on any unamed constraint.

If we are going to get picky on UPSERT vs REPLACE than maybe I should pursue developing upsert and upsertBatch. I don't know that we need both though and having both might just bloat CI.

Also, could use some help with the other databases. Would it be improper to throw an error not yet implemented in some of them? Then as we get around to it we add that functionality. That or maybe an insert ignore query followed by an update query.

kenjis: > Personally I'd prefer the UPSERT operation to the REPLACE. The current implementation has muddied the difference.

I agree with it. 
Oracle implementation uses `MERGE` that is UPSERT.

> Also, could use some help with the other databases. Would it be improper to throw an error not yet implemented in some of them? Then as we get around to it we add that functionality.

Implementing a new feature for all database drivers by one person is too hard.
Personally I agree with using an error not yet implemented.
But the new feature can be implemented in all supported database, and needs to be approved.

sclubricants: I've done some research on all the current databases and think I could put them all together. 

The Oracle driver is almost done already with the current replace method. It just needs to pass multiple values by select * from Dual Union All.

Postgre and SQLite both use a very similar ON CONFLICT DO.. 

The standard SQL specification:
"Database management systems Oracle Database, IBM Db2, Teradata, EXASOL, Firebird, CUBRID, H2, HSQLDB, MS SQL, Vectorwise and Apache Derby support the standard syntax."
https://en.wikipedia.org/wiki/Merge_(SQL)

So MSSQL should work the same of very similar to Oracle. The MERGE should be the BaseBuilder method. 

Mysql is mostly done already.

So question is, should I create new methods upsert() and upsertBatch() and leave alone the replace() implementation? If I were to change to upsert, should we scrap the PR and create a new one?

sclubricants:
Rather than expand the replace method I changed and went the route of adding upsert and upsertBatch methods.

The default SQL uses MERGE. This is standard. https://en.wikipedia.org/wiki/Merge_(SQL)

The default MERGE statement has been tested and working with Oracle. It should work for MSSQL but I need to test.

Mysql driver is using INSERT INTO.. ON DUPLICATE KEY UPDATE

I know it looks rough right now. But I would appreciate some input. I tried to refactor some things to avoid duplicating a bunch of code. Let me know if you approve. There probably is some more that could be done.

I made the upsert method to just convert to an array and then process with upsertBatch. I don't see any reason to have a separate method. 

Thoughts?

kenjis:
Personally I agree with adding `upsertBatch()` method.

sclubricants:
@kenjis I'm still working on the live tests. I'm unable to see the tests run on here though. Something about conflicting files that need resolved. Help?

sclubricants:
> @sclubricants You have conflicts. Please resolve.
> 
> Conflicting files system/BaseModel.php

I"m not understanding where the issue is or how to fix. Sorry, I've not used git very much. Can you explain how I can resolve?

kenjis:
The conflicts happened because the `4.3` branch was being developed.

You need to update your branch. Rebase it.
But your base branch is `4.3`, not `develop`.

```
git switch 4.3
git fetch upstream
git merge upstream/4.3
git push origin 4.3
```
```
git switch DatabaseBaseBuilderreplaceBatch
git rebase upstream/4.3
```

sclubricants:
@kenjis What edition of Oracle do the live tests use?

I'm using version 19 in my test environment and am not having any problem running the queries.

This one is hanging up in my test:

        CREATE OR REPLACE NONEDITIONABLE TRIGGER "REBATE_TRG"
        BEFORE INSERT ON REBATE
        FOR EACH ROW
        BEGIN
        <<COLUMN_SEQUENCES>>
        BEGIN
        IF INSERTING AND :NEW.REBATEREC IS NULL THEN
        SELECT REBATE_SEQ.NEXTVAL INTO :NEW.REBATEREC FROM DUAL;
        END IF;
        END COLUMN_SEQUENCES;
        END;

kenjis:
```
There was 1 error:
1) CodeIgniter\Database\Live\OCI8\UpsertTest::testSimpleUpsertTest
ErrorException: oci_execute(): OCI_SUCCESS_WITH_INFO: ORA-24344: success with compilation error
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/OCI8/Connection.php:198
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:691
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:618
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Database/Live/OCI8/UpsertTest.php:84
```
https://github.com/codeigniter4/CodeIgniter4/runs/6994971789?check_suite_focus=true

kenjis:
> What edition of Oracle do the live tests use?

Oracle Database 18c Express Edition (XE)

https://github.com/codeigniter4/CodeIgniter4/blob/099b6f4af8249fe26036291cb55fe4a65814eb3b/.github/workflows/test-phpunit.yml#L78
https://github.com/deusaquilus/docker-oracle-xe

sclubricants:
@kenjis @MGatner 

What is up with the coveralls tests? Sometimes they are all fine and others not. I can't seem to find out why. 

Other than that things are working fairly well. I haven't added any documentation yet. There are some things to discuss.

As it is it is in "auto" mode. It automatically selects which data to update and automatically assigns the first constraint where all fields are present. This could be coded to manually assign constraints for the databases that support it.

We could have something like: 

```php
$builder->updateFields(array)->onConstraint(array)->upsertBatch($data);
```
Also using a query as source data:

```php
$builder->fromSubquery($subquery, 't')->upsertBatch();
```

But before I go any farther maybe you guys could take a look at the code and let me know what you think?

kenjis:
> What is up with the coveralls tests? Sometimes they are all fine and others not. I can't seem to find out why.

coveralls sometimes gives us unexpected results.

MGatner:
I've noticed the Coveralls discrepancy, I'm not sure why it happens. At first I thought it was because we were running against different branches, but it kept happening even when 4.2.2 was released and the differences between develop and 4,3 were negligible.

My only other guess is that somehow new code from 4.3 is getting cached and being counted against PRs from develop. It might be worth updating our GitHub Actions cache to be branch-specific.

MGatner:
As for the database question, I had to read up on the distinctions. I haven't looked into other databases but for MySQL these behaviors are well defined, so REPLACE and INSERT UPDATE ON DUPLICATE are two distinct features and I don't think we should conflate them. We also don't need to try to "protect" the developer who should be aware of these nuances if they are using either of these commands with CASCADEs. We just choose names that match the underlying functionality and make it clear in the docs what it does.

kenjis:
Thank you for contributing.

Frankly, this PR is too big to review easily.
I would like to request the following.

1. split it
   - this PR is to add `upsertBatch()` in Builder, but `insertBatch()`, Model and Forge are changed.
   - e.g. 
     - make changes to the Model into a separate PR. 
     - make refactoring a separate PR.
2. put commits together in meaningful units (too many commits now)
   - e.g. 
      - put hacking commits together. 
      - put coding style fixing commits together with real changing commits.

And this branch has conflicts that must be resolved.

kenjis:
Thank you for your hard work to implement upsert().

My point is that no one can review PRs that are too big and will not be merged.

Anyway, I would like it to be a small PR. That way it has a better chance of being merged soon.
If changes are needed to prepare for upsert, then make those changes to other PRs.
Since upsert itself is a major feature addition, all other changes should be in separate PRs and make this upsert PR as small as possible.

> Also, without refactoring things it doesn't pass the tests because of a bunch of duplicate code. This PR replicates much of the code used in insertBatch(). Some of this code was moved around in order to reuse.

Do you mean PHPCPD checking? 
If so, I don't mind if it fails. We can fix it later.
However, the current PR does not show the duplicate code state before refactoring by looking at the commits.
So it is difficult to check if the refactoring is OK.

MGatner:
I agree with @kenjis: @sclubricants you have done some awesome work here and we would really like to capture that! Breaking up this PR into as many logical units as possible will help our already-taxed maintainers with the review and give you a chance to explain why the changes are necessary in each PR description. I know it is a bit more work for you (and you've already done lots) but you understand your code better than anyone else and it is the best way you can share that with us. 

I would say first focus on anything that you consider a "big fix" and send that to the `develop` branch; even if it is groundwork for upserts, a bug is a bug and it would be worth getting in sooner. 

We have very few contributors around who are comfortable across all the databases, so I really value your presence here. Let us know how we can help with the repo logistics.

PR #6109, opened 2022-06-11
feat: add method to insert empty data in Model

kenjis:
**Description**
- add `Model::allowEmptyInserts()` to insert empty data

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -315,9 +315,9 @@ abstract class BaseModel
     protected $afterDelete = [];
 
     /**
-     * Whether to prohibit inserting empty data.
+     * Whether to allow inserting empty data.
      */
-    protected bool $prohibitInsertEmpty = true;
+    protected bool $allowEmptyInserts = false;
 
     public function __construct(?ValidationInterface $validation = null)
     {
@@ -747,7 +747,7 @@ abstract class BaseModel
 
         // doProtectFields() can further remove elements from
         // $data so we need to check for empty dataset again
-        if ($this->prohibitInsertEmpty && empty($data)) {
+        if (! $this->allowEmptyInserts && empty($data)) {
             throw DataException::forEmptyDataset('insert');
         }
 
@@ -769,9 +769,6 @@ abstract class BaseModel
         }
 
         $result = $this->doInsert($eventData['data']);
-
-        // Reset $prohibitInsertEmpty value.
-        $this->prohibitInsertEmpty = true;
 
         $eventData = [
             'id'     => $this->insertID,
@@ -1648,7 +1645,7 @@ abstract class BaseModel
             throw new InvalidArgumentException(sprintf('Invalid type "%s" used upon transforming data to array.', $type));
         }
 
-        if ($this->prohibitInsertEmpty && empty($data)) {
+        if (! $this->allowEmptyInserts && empty($data)) {
             throw DataException::forEmptyDataset($type);
         }
 
@@ -1667,7 +1664,7 @@ abstract class BaseModel
         }
 
         // If it's still empty here, means $data is no change or is empty object
-        if ($this->prohibitInsertEmpty && empty($data)) {
+        if (! $this->allowEmptyInserts && empty($data)) {
             throw DataException::forEmptyDataset($type);
         }
 
@@ -1775,11 +1772,11 @@ abstract class BaseModel
     }
 
     /**
-     * Permits inserting empty date in the next insertion.
+     * Sets $allowEmptyInserts.
      */
-    public function permitInsertEmpty(): self
+    public function allowEmptyInserts(bool $value = true): self
     {
-        $this->prohibitInsertEmpty = false;
+        $this->allowEmptyInserts = $value;
 
         return $this;
     }
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -314,6 +314,11 @@ abstract class BaseModel
      */
     protected $afterDelete = [];
 
+    /**
+     * Whether to prohibit inserting empty data.
+     */
+    protected bool $prohibitInsertEmpty = true;
+
     public function __construct(?ValidationInterface $validation = null)
     {
         $this->tempReturnType     = $this->returnType;
@@ -742,7 +747,7 @@ abstract class BaseModel
 
         // doProtectFields() can further remove elements from
         // $data so we need to check for empty dataset again
-        if (empty($data)) {
+        if ($this->prohibitInsertEmpty && empty($data)) {
             throw DataException::forEmptyDataset('insert');
         }
 
@@ -764,6 +769,9 @@ abstract class BaseModel
         }
 
         $result = $this->doInsert($eventData['data']);
+
+        // Reset $prohibitInsertEmpty value.
+        $this->prohibitInsertEmpty = true;
 
         $eventData = [
             'id'     => $this->insertID,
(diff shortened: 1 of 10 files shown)


kenjis:
It seems OCI8 driver can't get the insert id.

```
1) CodeIgniter\Models\InsertModelTest::testInsertPermitInsertNoData
Failed asserting that a row in the table [insert_no_data] matches the attributes 
{
    "id": 0
}

Found: [].

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Test/DatabaseTestTrait.php:280
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Models/InsertModelTest.php:220
```
https://github.com/codeigniter4/CodeIgniter4/runs/6840259577?check_suite_focus=true

iRedds:
I saw on a forum or a github issue that someone came across the inability to use an empty request.
In my opinion, this is due to the problem of database design, since in my opinion the table for the test in PR does not make logical sense without links to other tables and containing only auto-increment and timestamps.

The solution is to use a raw query or add a tinyint(1) column with arbitrary data.
It seems to me that adding just such functionality is not rational.

kenjis:
> In my opinion, this is due to the problem of database design, since in my opinion the table for the test in PR does not make logical sense without links to other tables and containing only auto-increment and timestamps.

It makes sense if it has links to other tables. 
For example, imagine that is `users` table, and all other user's attributes are stored in other tables.

iRedds:
I wrote about links with other tables.
But what's the point of creating an empty record in the dependent table.

You either add data or you don't.
I can't imagine a case where it makes sense to add a record without data.
Maybe I don't understand something?

kenjis:
It is not to add a record without data. The record will have ID.
But just insert data is empty.
If you add timestamps to columns by the field default value, the record will have timestamps, but the insert data is empty.

The point is sometimes we need this functionality to avoid "no data to insert" exception of `Model`
It may be because of bad design for tables, but we don't say it is always  because of bad design.

iRedds:
The table consisting of (id, created_at, updated_at) seems strange to me.

system/BaseModel.php
```diff
@@ -286,6 +286,11 @@ abstract class BaseModel
      */
     protected $afterDelete = [];
 
+    /**
+     * Whether to prohibit inserting empty data.
+     */
+    protected bool $prohibitInsertEmpty = true;
```
MGatner, line 292: Consider naming after the existing properties. 
```suggestion
    protected bool $allowEmptyInserts = false;
```

system/BaseModel.php
```diff
@@ -737,6 +742,9 @@ public function insert($data = null, bool $returnID = true)
 
         $result = $this->doInsert($eventData['data']);
 
+        // Reset $prohibitInsertEmpty value.
+        $this->prohibitInsertEmpty = true;
```
MGatner, line 746: I think this is too forceful. If a developer sets the property and then it gets reset here that would feel very buggy. I think the workaround is to implement a temporary value, like how `$tempAllowCallbacks`, etc, work.

system/BaseModel.php
```diff
@@ -1697,4 +1705,14 @@ protected function fillPlaceholders(array $rules, array $data): array
 
         return $rules;
     }
+
+    /**
+     * Permits inserting empty date in the next insertion.
```
MGatner, line 1710:
```suggestion
     * Permits inserting empty data in the next insertion.
```

system/Model.php
```diff
@@ -275,7 +275,18 @@ protected function doInsert(array $data)
             $builder->set($key, $val, $escape[$key] ?? null);
         }
 
-        $result = $builder->insert();
+        if (! $this->prohibitInsertEmpty && empty($data)) {
+            $table = $this->db->protectIdentifiers($this->table, true, null, false);
+            if ($this->db->getPlatform() === 'MySQLi') {
+                $sql = 'INSERT INTO ' . $table . ' VALUES ()';
+            } else {
+                $sql = 'INSERT INTO ' . $table . ' DEFAULT VALUES';
```
MGatner, line 283: I'm not familiar with this but tests seem to vet it.

MGatner (changes requested):
I like this feature. Some implementation suggestions and requests.

MGatner:
The table could have any amount of data from default values. If a developer wants to let the database handle those, as it stands currently you need to submit some "dummy" data. I think our protections in general are good but I am a fan of the option to override those protections when needed. 

Example based on kenjis:
```
TABLE users
PRIMARY KEY id
TIMESTAMPS created_at, updated_at
FIELD role DEFAULT 'user'
FIELD active DEFAULT 0

TABLE emails
PRIMARY KEY email
TIMESTAMPS created_at
FOREIGN KEY user_id
```

This fails:
```
$id = model(UserModel::class)->insertReturnId([]);
model(EmailModel::class)->insert([
    'user_id' => $id,
    'email' => 'iredds@example.com',
]);
```

system/BaseModel.php
kenjis: Done.

system/BaseModel.php
kenjis: I removed it and now `$allowEmptyInserts` is a normal property.

system/BaseModel.php
kenjis: Fixed.

kenjis:
@ytetsuro Can you debug this? It seems only Oracle can't get the insert id.
https://github.com/codeigniter4/CodeIgniter4/pull/6109#issuecomment-1152840138

ytetsuro:
@kenjis 

Ok!
I understood!
I will check it!

ytetsuro (commented):
@kenjis 

This test is lose the database error message.
You can’t understand why fail.😭

I fixed it.👍

#6113

system/Model.php
kenjis: Oracle does not work with it. 😓

kenjis:
@ytetsuro Thank you so much!
I was wondering if there was some bug in getting the insert id.

MGatner:
I would like to hear if @iRedds has any other thoughts before we move on from the discussion.

kenjis:
I think if we support insertion of empty data, Query Builder should support it.
The query is not unified.
See https://github.com/codeigniter4/CodeIgniter4/pull/6109/files#diff-1a690370bdbdc5997068f9ff210bacef314eeb65d1df73af206783ce055d8d1dR280

MGatner (approved):
Looks good!

kenjis:
Rebased and added the documentation.

@iRedds Any comments?

MGatner (approved):

paulbalandan (approved):

kenjis:
Resolved conflict.

kenjis:
Sorry, I made a mistake when I rebased.

I mistakenly rebased with `develop` first and then rebased with `4.3`.
So this PR contains commits from `develop` that have not yet been merged into `4.3`.

MGatner:
Shouldn't be a problem, `develop => 4.3` should handle that after next patch release.

kenjis:
Okay, thanks.
Since `revert` does not make much sense, leave it as it is.

> develop => 4.3 should handle that after next patch release.

No, it is daily job.
I generally do it once a day.

If we do not merge, bugs will remain unfixed in `4.3`, and conflicts are likely to occur as the difference between `develop` and `4.3` grows. So it is better to merge continuously.

MGatner:
If that becomes tiresome we could set up a GitHub Action to merge to 4.3 anytime develop is updated. I'd be willing to look into that.

PR #6110, opened 2022-06-11
Extracting the call handler for Spark commands from kernel.

iRedds:
**Description**
This is the concept of extracting the call to Spark commands from processing via routing. Commands are called directly, as when using the ``command()`` function. This will reduce the load on the console application in the form of Request, Response, Filters and Routes services and more.

In the beginning, the idea of creating an Application class was considered, from which the ``CodeIniter`` and ``Console`` classes would be inherited. This was necessary to initialize the application. But due to the fact that, at the request of one person, the core of the application turned into a service, there were difficulties with the implementation.

Therefore, the initialization of the application through the ``CodeIgniter`` class is saved.

The ``Console`` class now uses the ``Services::commands()`` service directly instead of using the ``CommandRunner`` class.
The ``CodeIgniter`` class has been cleared of processing Spark commands, and the ``Console`` class no longer processes ``Response``.

~~Although this is not within the purview of the PR, I added the ``StreamFilterTrait`` trait, which correctly adds and removes filters for capturing stream output, since the current implementation in the tests is not correct enough in my opinion.~~
Depends on StreamFilterTrait https://github.com/codeigniter4/CodeIgniter4/pull/6112

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


kenjis:
I have not looked into this PR yet, but this is a big change.
It would take some time to review.

So could you please split `StreamFilterTrait` into a separate PR?
It is small and will be merged sooner than this.

system/Test/StreamFilterTrait.php
```diff
@@ -0,0 +1,92 @@
...
+{
+    /**
+     * @var resource|null
+     */
+    protected $streamFilterOutResource;
+
+    /**
+     * @var resource|null
+     */
+    protected $streamFilterErrResource;
```
MGatner, line 26: Any reason not to make these private?

MGatner (commented):
I think this is the right direction. We have some big conversations going around CLI and the app so it is helpful to have an actual implementation to look at. Let's make this our general discussion place at well, just for the sake of having a common thread.

system/Test/StreamFilterTrait.php
iRedds: Thanks for the suggestion.
I took it into account and moved the trait to a separate PR https://github.com/codeigniter4/CodeIgniter4/pull/6112

iRedds:
@kenjis StreamFilterTrait https://github.com/codeigniter4/CodeIgniter4/pull/6112

system/Config/Routes.php
```diff
@@ -20,4 +19,4 @@
  */
 
 // CLI Catchall - uses a _remap to call Commands
-$routes->cli('ci(:any)', '\CodeIgniter\CLI\CommandRunner::index/$1');
+// $routes->cli('ci(:any)', '\CodeIgniter\CLI\CommandRunner::index/$1');
```
kenjis, line 22: I think these lines are not necessary.

iRedds: Maybe then it makes sense to delete this file altogether and remove its connection from the application configuration?
https://github.com/codeigniter4/CodeIgniter4/blob/5879a3ade59dd25dc299e351e85ee51d3633cd16/app/Config/Routes.php#L10-L12

MGatner:
I'm still in favor of this. The rest of the team is oddly quiet 😅 Let's get the User Guide and changelog updates done and then I will give them one last chance.

system/CLI/Console.php
```diff
@@ -12,40 +12,28 @@
...
-        $path = CLI::getURI() ?: 'list';
-
-        // Set the path for the application to route to.
-        $this->app->setPath("ci{$path}");
+        $runner  = Services::commands();
+        $params  = array_merge(CLI::getSegments(), CLI::getOptions());
+        $command = array_shift($params) ?? 'list';
 
-        return $this->app->useSafeOutput($useSafeOutput)->run();
+        return $runner->run($command, $params);
```
paulbalandan, line 36: `BaseCommand::run()` called by `Commands::run()` returns `void` although not strictly because not typehinted. However, most of our implementations of commands I think don't return anything either. So the `$exit` in L85 in spark above will be `null` so always exit on `EXIT_SUCCESS`?

system/CodeIgniter.php
```diff
@@ -871,9 +847,7 @@ protected function createController()
      * CI4 supports three types of requests:
      *  1. Web: URI segments become parameters, sent to Controllers via Routes,
      *      output controlled by Headers to browser
-     *  2. Spark: accessed by CLI via the spark command, arguments are Command arguments,
-     *      sent to Commands by CommandRunner, output controlled by CLI class
```
paulbalandan, line 875: Since this is removed, does this mean that the `CommandRunner` will be removed? Or deprecated, at least?

tests/system/CLI/ConsoleTest.php
```diff
@@ -38,26 +37,19 @@ protected function setUp(): void
...
-        $this->app->setContext('spark');
+        $this->app->initialize();
     }
 
     public function testNew()
     {
-        $console = new Console($this->app);
+        $console = new Console();
         $this->assertInstanceOf(Console::class, $console);
     }
```
paulbalandan, line 48: This test is funny.

system/CLI/Console.php
iRedds: It turns out that exit (0), unless of course the developer forcibly returns the desired value of the exit code.

system/CodeIgniter.php
iRedds: The ``CommandRunner`` class will be removed as there is no point in it. I'm already preparing changes.

tests/system/CLI/ConsoleTest.php
iRedds: I agree. I just now noticed it.

user_guide_src/source/incoming/routing.rst
```diff
@@ -327,6 +327,9 @@ available from the command line:
 .. warning:: If you enable auto-routing and place the command file in **app/Controllers**,
     anyone could access the command with the help of auto-routing via HTTP.
 
+.. note:: It is recommended to use Spark Commands instead of CLI routes.
+    See the :doc:`spark_commands` page for detailed information.
```
kenjis, line 331:
```suggestion
    See the :doc:`../cli/spark_commands` page for detailed information.
```

user_guide_src/source/cli/cli_commands.rst
```diff
@@ -82,6 +82,12 @@ Our demo command might have a ``run()`` method something like:
 
 See the :doc:`CLI Library </cli/cli_library>` page for detailed information.
 
+Command termination
+===================
+
+By default, the command terminates with a success code of ``0``. To change the exit code, for example on error,
+the ``run()`` method must return an integer in the range 0 - 254. `PHP exit <https://www.php.net/manual/en/function.exit.php>`_.
```
kenjis, line 89:
```suggestion
the ``run()`` method must return an integer in the range 0 - 254. See `PHP exit <https://www.php.net/manual/en/function.exit.php>`_.
```

kenjis:
@iRedds Thank you for updating the docs.

One last thing, Upgrading Guide also needed.
Like this: https://codeigniter4.github.io/CodeIgniter4/installation/upgrade_420.html#mandatory-file-changes

See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#upgrading-guide

user_guide_src/source/cli/cli_commands.rst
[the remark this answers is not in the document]
kenjis: Good suggestion! I forgot the constants.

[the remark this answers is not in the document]
iRedds: Made changes. I hope this is similar to what you were talking about.

iRedds:
@kenjis I have doubts about the quality of the upgrade manual.

user_guide_src/source/cli/cli_commands.rst
```diff
@@ -82,6 +82,18 @@ Our demo command might have a ``run()`` method something like:
 
 See the :doc:`CLI Library </cli/cli_library>` page for detailed information.
 
+Command termination
```
kenjis, line 85:
```suggestion
Command Termination
```

kenjis:
Please add.

```diff
--- a/user_guide_src/source/installation/upgrading.rst
+++ b/user_guide_src/source/installation/upgrading.rst
@@ -12,6 +12,7 @@ upgrading from.
 .. toctree::
     :titlesonly:
 
+    upgrade_430
     upgrade_421
     upgrade_420
     upgrade_418
```

I think the upgrading is okay. Thanks.

MGatner (approved):

paulbalandan (approved):

kenjis (approved):
LGTM!

PR #6112, opened 2022-06-11
Feature: Adding StreamFilterTrait

iRedds:
**Description**
StreamFilterTrait trait, which correctly adds and removes filters for capturing stream output, since the current implementation in the tests is not correct enough in my opinion.

**Overview of methods**

- ``StreamFilterTrait::registerStreamFilterClass()`` Registering a filter to capture streams.
- ``StreamFilterTrait::appendStreamOutputFilter()`` Adding a filter to the output stream.
- ``StreamFilterTrait::appendStreamErrorFilter()`` Adding a filter to the error stream.
- ``StreamFilterTrait::removeStreamOutputFilter()`` Removing the filter from the output stream.
- ``StreamFilterTrait::removeStreamErrorFilter()`` Removing the filter from the error stream.
- ``StreamFilterTrait::getStreamFilterBuffer()`` Get the captured data from the buffer.
- ``StreamFilterTrait::resetStreamFilterBuffer()`` Reset captured data.

```diff
-        CITestStreamFilter::$buffer = '';
-
-        $this->streamFilter = stream_filter_append(STDOUT, 'CITestStreamFilter');
-        $this->streamFilter = stream_filter_append(STDERR, 'CITestStreamFilter');
+        $this->registerStreamFilterClass()
+            ->appendStreamOutputFilter()
+            ->appendStreamErrorFilter();
```
**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -1,0 +1,35 @@
+Version 4.3.0
+#############
+
+Release Date: Unreleased
+
+**4.3.0 release of CodeIgniter4**
+
+.. contents::
+    :local:
+    :depth: 2
+
+BREAKING
+********
+
+none.
+
+Enhancements
+************
+
+- Added the ``StreamFilterTrait`` to make it easier to work with capturing data from STDOUT and STDERR streams. See :ref:`testing-overview-stream-filters`.
+
+Changes
+*******
+
+none.
+
+Deprecations
+************
+
+none.
+
+Bugs Fixed
+**********
+
+none.
--- a/user_guide_src/source/testing/overview.rst
+++ b/user_guide_src/source/testing/overview.rst
@@ -262,13 +262,13 @@ might be helpful. The ``StreamFilterTrait`` helps you capture the output from th
 
 ``StreamFilterTrait::registerStreamFilterClass()`` Registering a filter to capture streams.
 
-``StreamFilterTrait::appendStreamOutputFilter()`` Adding a filter to the output stream.
+``StreamFilterTrait::appendOutputStreamFilter()`` Adding a filter to the output stream.
 
-``StreamFilterTrait::appendStreamErrorFilter()`` Adding a filter to the error stream.
+``StreamFilterTrait::appendErrorStreamFilter()`` Adding a filter to the error stream.
 
-``StreamFilterTrait::removeStreamOutputFilter()`` Removing the filter from the output stream.
+``StreamFilterTrait::removeOutputStreamFilter()`` Removing the filter from the output stream.
 
-``StreamFilterTrait::removeStreamErrorFilter()`` Removing the filter from the error stream.
+``StreamFilterTrait::removeErrorStreamFilter()`` Removing the filter from the error stream.
 
 ``StreamFilterTrait::getStreamFilterBuffer()`` Get the captured data from the buffer.
 
--- a/user_guide_src/source/testing/overview.rst
+++ b/user_guide_src/source/testing/overview.rst
@@ -253,10 +253,26 @@ component name:
 Stream Filters
 ==============
 
-**CITestStreamFilter** provides an alternate to these helper methods.
+**StreamFilterTrait** provides an alternate to these helper methods.
 
 You may need to test things that are difficult to test. Sometimes, capturing a stream, like PHP's own STDOUT, or STDERR,
-might be helpful. The ``CITestStreamFilter`` helps you capture the output from the stream of your choice.
+might be helpful. The ``StreamFilterTrait`` helps you capture the output from the stream of your choice.
+
+**Overview of methods**
+
+``StreamFilterTrait::registerStreamFilterClass()`` Registering a filter to capture streams.
+
+``StreamFilterTrait::appendStreamOutputFilter()`` Adding a filter to the output stream.
+
+``StreamFilterTrait::appendStreamErrorFilter()`` Adding a filter to the error stream.
+
+``StreamFilterTrait::removeStreamOutputFilter()`` Removing the filter from the output stream.
+
+``StreamFilterTrait::removeStreamErrorFilter()`` Removing the filter from the error stream.
+
+``StreamFilterTrait::getStreamFilterBuffer()`` Get the captured data from the buffer.
+
+``StreamFilterTrait::resetStreamFilterBuffer()`` Reset captured data.
 
 An example demonstrating this inside one of your test cases:
 
--- a/user_guide_src/source/testing/overview.rst
+++ b/user_guide_src/source/testing/overview.rst
@@ -103,6 +103,8 @@ parent as well so extended test cases do not interfere with staging:
 
 .. literalinclude:: overview/003.php
 
+.. _testing-overview-traits:
+
 Traits
 ------
 
@@ -250,6 +252,8 @@ component name:
 
 .. note:: All component Factories are reset by default between each test. Modify your test case's ``$setUpMethods`` if you need instances to persist.
 
+.. _testing-overview-stream-filters:
+
 Stream Filters
 ==============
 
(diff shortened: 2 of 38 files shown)


kenjis:
Can you rebase to remove unrelated commits?

iRedds:
I may be doing something wrong, but the PR has already been rebased.

kenjis:
Did you do this?
```
git fetch upstream
git rebase upstream/4.3
```

kenjis:
Okay, fixed.

kenjis:
Probably you created this PR branch from `develop` and it was ahead than `4.3` at that time.
So unrelated commits were included in this PR.

iRedds:
I've rebased onto upstream/develop.
As always. As in the manual.

Now the new PR should be based on 4.3 and rebased on upstream / 4.3?

kenjis:
Sorry, the manual is not updated yet.

Now we have two development branches, `develop` and `4.3`.
- `develop` is for 4.2.x, that is bug fixe only.
- `4.3` is for 4.3.x.

This is so that bug fix versions can be released quickly.

> Now the new PR should be based on 4.3 and rebased on upstream/4.3?

The bug fix only PRs are based on `develop` and rebased on `upstream/develop`.
PRs with any enhancements are based on `4.3`and rebased on `upstream/4.3`.

iRedds:
You came up with something complicated )

paulbalandan (commented):
This seems handy!

My suggestion: Can we make `StreamOutput` to `OutputStream` and `StreamError` to `ErrorStream`, i.e. `appendStreamOutputFilter` vs `appendOutputStreamFilter`? In my opinion, that will make the methods sound more "natural" and would make the methods descriptive of their names (e.g., `appendOutputStreamFilter` is "append filter to the output stream").

iRedds:
@paulbalandan I agree. It really does sound better.

MGatner (changes requested):
This is great! I always end up copying the framework test methods into projects when I need buffer testing. 

My only thought: should automate stream registration when using the trait? Or stream removal? Or both? 

https://codeigniter4.github.io/userguide/testing/overview.html#traits

paulbalandan (commented):
Please also update this: https://codeigniter4.github.io/userguide/testing/overview.html#stream-filters

iRedds:
@paulbalandan already added. https://github.com/codeigniter4/CodeIgniter4/pull/6112/commits/f67d745b7f3466690f96b953091e6586bcba7826

I'm not good at github actions, but if I'm right, the 4.3 branch won't show up here.

iRedds:
@MGatner If it's only about registering a filter class, then I think it's a good idea.
But regarding gluing the filter to the stream. I have doubts. So I'd like to hear more opinions on this.

kenjis:
> And most tests override the setUp() and tearDown() methods without calling the parent.

I think it is a bug that does not call parent.

Are there use cases that we don't want to append/remove the stream filters automatically?

iRedds:
@kenjis I can't think of a situation yet that would cause problems. That's why I asked for an opinion.

MGatner:
I guess the docs aren't very clear, probably my fault. But traits can provide their own methods that will be called automatically: `setUpStreamFilterTrait()`, `tearDownStreamFilterTrait()`.

I think since most cases if a developer is using the trait they want this baseline we should consider making that happen without intervention.

iRedds:
@MGatner Thanks for the clarification, but I initially understood your idea of trait preconfiguration.
I will try to implement the changes you suggested.

user_guide_src/source/testing/overview.rst
```diff
@@ -256,14 +258,31 @@ component name:
...
 .. literalinclude:: overview/018.php
+
+The ``StreamFilterTrait`` has a configurator that is called automatically.
+See :ref:`testing-overview-traits`.
+
+If you override the ``setUp()`` or ``tearDown()`` methods in your test, then you must call the ``parent::setUp()`` and
+``parent::tearDown()`` methods respectively to configure the ``StreamFilterTrait``.
+
+If your test does not inherit from the ``CIUnitTestCase`` class, then you need to call the ``setUpStreamFilterTrait()``
+and ``tearDownStreamFilterTrait()`` methods.
```
kenjis, line 286: Do you need this?
I don't see the use cases.

user_guide_src/source/testing/overview.rst
```diff
@@ -256,14 +258,31 @@ component name:
...
+
+.. literalinclude:: overview/019.php
```
kenjis, line 288: This sample code is difficult to understand, because it is mixture of overriding setUp and not extending CIUnitTestCase.

kenjis:
Please add:
```diff
--- a/user_guide_src/source/changelogs/index.rst
+++ b/user_guide_src/source/changelogs/index.rst
@@ -12,6 +12,7 @@ See all the changes.
 .. toctree::
     :titlesonly:
 
+    v4.3.0
     v4.2.1
     v4.2.0
     v4.1.9
```

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -0,0 +1,36 @@
...
+BREAKING
+********
+
+none.
+
+Enhancements
+************
+
+- Added the ``StreamFilterTrait`` to make it easier to work with capturing data from STDOUT and STDERR streams.
+See :ref:`testing-overview-stream-filters`.
```
kenjis, line 21: You cannot add a line break here. It is syntax error.

```suggestion
- Added the ``StreamFilterTrait`` to make it easier to work with capturing data from STDOUT and STDERR streams. See :ref:`testing-overview-stream-filters`.
```

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -0,0 +1,36 @@
+Version 4.3.0
+#############
+
+Release Date: MONTH DAY, 2022
```
kenjis, line 4:
```suggestion
Release Date: Unreleased
```

user_guide_src/source/testing/overview.rst
```diff
@@ -256,14 +258,31 @@ component name:
...
+**Overview of methods**
+``StreamFilterTrait::getStreamFilterBuffer()`` Get the captured data from the buffer.
+``StreamFilterTrait::resetStreamFilterBuffer()`` Reset captured data.
 
 An example demonstrating this inside one of your test cases:
 
 .. literalinclude:: overview/018.php
+
+The ``StreamFilterTrait`` has a configurator that is called automatically.
+See :ref:`testing-overview-traits`.
```
kenjis, line 280:
```suggestion
See :ref:`Testing Traits <testing-overview-traits>`.
```

user_guide_src/source/testing/overview.rst
iRedds: I decided to remove this explanation with an example.

kenjis:
Sorry for a comment after a comment.
Probably the previous one is the last.

user_guide_src/source/testing/overview/018.php
```diff
@@ -2,26 +2,19 @@
...
-
-    protected function tearDown(): void
-    {
-        stream_filter_remove($this->stream_filter);
-    }
+    use StreamFilterTrait;
 
     public function testSomeOutput(): void
     {
+        $this->resetStreamFilterBuffer();
```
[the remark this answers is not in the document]
iRedds: I don't remember why I added a buffer reset.
Maybe I should add another assertion so that the buffer reset makes sense?

kenjis, line 13: How about this?
```suggestion
        // Setup method of StreamFilterTrait clears the buffer.
        $this->assertSame('', $this->getStreamFilterBuffer());
```

user_guide_src/source/testing/overview/018.php
```diff
@@ -2,26 +2,19 @@
...
+
         CLI::write('first.');
 
         $expected = "first.\n";
```
kenjis, line 17: The assertion failed.

```suggestion
        $expected = "\nfirst.\n";
```

MGatner (approved):
Kenjis making you work for it 😂
This is looking great and I'm excited about the feature. I'm good with it whenever you all think it is ready.

paulbalandan (approved):

user_guide_src/source/testing/overview/018.php
kenjis: What do you mean?
Can you update the sample code to what you consider good?

iRedds: I think that the example should show not only getting data from the buffer, but also resetting it.

```php
public function testSomeOutput(): void
{
    CLI::write('first.');

    $expected = "\nfirst.\n";
    $this->assertSame($expected, $this->getStreamFilterBuffer());
    
    $this->resetStreamFilterBuffer();

    CLI::write('second.');
    
    $expected = "second.\n";
    $this->assertSame($expected, $this->getStreamFilterBuffer());
}

```

kenjis: Okay, I got it.

Can you update the sample code?
And let's finish this PR.

kenjis (approved):
LGTM!

kenjis:
@iRedds Thank you!
This is a great improvement for CLI testing.

PR #6113, opened 2022-06-12
Fix lost error message by test when after testInsertResultFail.

ytetsuro:
testInsertResultFail is set DBdebug in false.
so that means a lost error message when after a testInsertResultFail run.

**Description**

The test developer can not understand the failure reason.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Models/InsertModelTest.php
+++ b/tests/system/Models/InsertModelTest.php
@@ -26,6 +26,12 @@ use Tests\Support\Models\WithoutAutoIncrementModel;
  */
 final class InsertModelTest extends LiveModelTestCase
 {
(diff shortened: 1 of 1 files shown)


ytetsuro:
This tearDown may be implemented in the parent class.
What do you think?☺️

MGatner (approved):

MGatner:
I'm fine with it here or in the parent. Except in specific tests where it might be changed we should always be in debug mode.

kenjis:
The test that changed the DBDebug value should restore the value.

kenjis (approved):

PR #6115, opened 2022-06-13
test: fix forgetting to restore DBDebug value

kenjis:
**Description**
- add `disableDBDebug()` and `enableDBDebug()` in `DatabaseTestTrait`
- add missing `enableDBDebug()`
- refactor

Related #6113

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/DatabaseTestTrait.php
+++ b/system/Test/DatabaseTestTrait.php
@@ -323,4 +323,22 @@ trait DatabaseTestTrait
 
         $this->assertEquals($expected, $count, 'Wrong number of matching rows in database.');
     }
+
+    /**
+     * Sets $DBDebug to false.
+     *
+     * WARNING: this value will persist! take care to roll it back.
+     */
+    protected function disableDBDebug(): void
+    {
+        $this->setPrivateProperty($this->db, 'DBDebug', false);
+    }
+
+    /**
+     * Sets $DBDebug to true.
+     */
+    protected function enableDBDebug(): void
+    {
+        $this->setPrivateProperty($this->db, 'DBDebug', true);
+    }
 }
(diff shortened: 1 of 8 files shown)


kenjis:
All tests failed except MySQL. 😅

MGatner (approved):
Wow I had no idea we did this in so many cases. Very good consolidation. 

Something I would like to see in a future version is a way to "schedule" tear-down actions better - so something that knows it will need to be undone can go ahead and put that in place. It would something akin to this:
```php
    protected function enableDBDebug(): void
    {
        $this->setPrivateProperty($this->db, 'DBDebug', true);

        $this->tearDownMethods[] = 'disableDBDebug`;
    }
```

MGatner:
> the development environment CI throws Exceptions when DB errors occur,
but in the production environment, CI returns `false`.

This is pretty bad behavior. Does DBDebug control anything else? Or is it just error handling?

kenjis:
> This is pretty bad behavior. Does DBDebug control anything else? Or is it just error handling?

Probably major behavior change is throwing a Exception or returning false.
But we need to investigate.

MGatner:
For sure. I favor exceptions more than most of the team, so take it with a grain of salt but it seems to me anytime you "return false" to indicate failure and have a hidden error log that should have been an exception.

kenjis:
I think this should be changed to always throw an exception on this matter.

Since it is a BC, I think it needs to be configurable for exception mode and false mode.

MGatner:
What about something like `Config\Feature::$databaseExceptions` that we recommend turning on in 4.3?

kenjis:
Yes, something like that.

Issue #4708, opened 2021-05-19
Bug: Time::humanize() does not work with ar locale


perspolise:
I set arabic language in app config file for defualtLocale like this:

`public $defaultLocale = 'ar';`

Now, I work with `humanize` time function like this:

```
$time = CodeIgniter\I18n\Time::parse($date, 'America/Chicago');
return $time->humanize(); 
```

Now in action I get this error:

```

DateTime::__construct(): Failed to parse time string (٢٠٢١-٠٥-١٩ ٠١:٤٧:٢٨) at position 0 (�): Unexpected character

SYSTEMPATH\I18n\Time.php at line 1150

public function humanize()
1149     {
1150         $now  = IntlCalendar::fromDateTime(Time::now($this->timezone)->toDateTimeString());

```

I think the `DateTime` function have problem with arabic number. because when I set `en` for default locale, `humanize function` work truly. 

@lonnieezell @MGatner

MGatner:
No need to ping, I read every issue and respond according to my ability and time. 

I don't have much experience with Unicode string handling. Hopefully someone familiar will step in.

paulbalandan:
I cannot reproduce that this works using the default locale `en`. Even this simple script fails.
```php
$date = '٠١:٤٧:٢٨';
$time = (new DateTime())->modify($date);
var_dump($time);
```

Unfortunately, this is not a CI4 bug. This is an inherent limitation of the `DateTime` parser used by PHP. It can only understand Latin input as described [here](https://www.php.net/manual/en/datetime.formats.compound.php).

perspolise:
@paulbalandan Sure, You right. But if we need to a multilingual system, humanize not work truly(for `en` it's ok, for others is not ok). I think CI4 doesn't need to convert numbers.

paulbalandan:
Yes, I understand what you want. But this issue is not limited to `Time`'s humanize() function. The issue lies internally in the PHP core. Since PHP's own `DateTime` doesn't understand Arabic dates, naturally it follows that CI4's `Time` (which is an extension of `DateTime`) will also fail with Arabic dates. As you can see, even `DateTime::modify()` fails parsing the date so this is outside CI4.

What I can see as a potential solution is to have first a converter to/from Arabic dates. Something like `strtr` replacements. After conversion, pass it to DateTime/Time. However, I don't see that converter fit into the framework as we would need to also cater other languages with non-Latin datetimes.

kenjis:
[...]

PR #6116, opened 2022-06-13
fix: Time::setTestNow() does not work with fa Locale

kenjis:
**Description**
```
ErrorException : DateTime::modify(): Failed to parse time string (۲۰۱۷-۰۳-۱۰ ۰۰:۰۰:۰۰) at position 0 (�): Unexpected character
 .../CodeIgniter4/system/I18n/Time.php:96
```

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -82,7 +82,7 @@ class Time extends DateTime
         // If a test instance has been provided, use it instead.
         if ($time === '' && static::$testNow instanceof self) {
             $timezone = $timezone ?: static::$testNow->getTimezone();
-            $time     = (string) static::$testNow->toDateTimeString();
+            $time     = static::$testNow->format('Y-m-d H:i:s');
         }
 
         $timezone       = $timezone ?: date_default_timezone_get();
--- a/tests/system/I18n/TimeTest.php
+++ b/tests/system/I18n/TimeTest.php
@@ -1058,4 +1058,18 @@ final class TimeTest extends CIUnitTestCase
         $this->assertTrue($time2->equals($time1));
         $this->assertNotSame($time1, $time2);
     }
+
+    public function testSetTestNowWithFaLocale()
+    {
+        $currentLocale = Locale::getDefault();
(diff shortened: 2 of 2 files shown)


MGatner (approved):

MGatner:
I'm not sure what is happening with Coveralls. Maybe it is getting confused about the different branches?

kenjis:
It is difficult to read but I think `toDateTimeString()` is localized version of `Y-m-d H:i:s`.

> This is the first of three helper methods **to work with the IntlDateFormatter** without having to remember their values. This will return a string formatted as you would commonly use for datetime columns in a database (Y-m-d H:i:s):
https://codeigniter4.github.io/CodeIgniter4/libraries/time.html?highlight=time#todatetimestring

It is clear if you see the implementation.

MGatner:
You're right! But the UG is still confusing: I would have used `toDateTimeString()` here as well. I wonder if we need a new method specifically for non-localized database values? Is there ever a case where your timestamp values need to be localized for storage input?

MGatner:
After some reading it looks like DATETIME/TIMESTAMP inputs must always be decimal. I think we need to add `toDatabase()` or something like it and update the User Guide so developers aren't accidentally using localized strings for database input.

datamweb:
Friends just for information, the output of command `toLocalizedString()` is also `UTF-8 string (33) "۲۰۲۲-۰۶-۱۳ ۱۷:۳۰:۱۳"`.

paulbalandan:
Is this related to here?
https://github.com/codeigniter4/CodeIgniter4/issues/4708

kenjis:
@paulbalandan Yes. The issue is because of this bug.

samsonasik (approved):

kenjis:
> But the UG is still confusing

I sent a PR #6121

MGatner (approved):
Okay, I am on board - this is definitely e right direction. 4.3 feature request: `toDatabase()`

PR #6118, opened 2022-06-13
docs: remove explanation about `$setUpMethods` and `tearDownMethods`

kenjis:
**Description**
These properties are not so convenient, and users do not need to use in most cases.

If users override `$setUpMethods` and remove the default items, some error may occur.
E.g. `ErrorException: ini_set(): Session ini settings cannot be changed after headers have already been sent`

- remove about `$setUpMethods` and `tearDownMethods`
- improve explanation about PHPUnit setUp/tearDown methods

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


kenjis:
@MGatner When we override `$setUpMethods`, we need to write:
```php
    protected $setUpMethods = [
        'someMyOwnMethod',
        'resetFactories',
        'mockCache',
        'mockEmail',
        'mockSession',
    ];
```
It seems the property is not so convenient.

How do you use this property?

MGatner:
@kenjis I don't use it. It was a good idea in theory but it has ends up being difficult and awkward to use, as you noted. I would definitely like to do away with this in version 5 and find a better way to manage test staging.

kenjis:
Then can I remove these from the user guide?

MGatner:
Sure. I'm hesitant to deprecate the property because I'm worried static analysis might complain, but maybe add a note that it will go away in the future?

MGatner:
Not sure if this is related to the conversation on #6112, but the code I was referring to there is actually this one: https://github.com/codeigniter4/CodeIgniter4/blob/8e8973634ce98558fd230a663c00475dcd8f2dca/system/Test/CIUnitTestCase.php#L266-L285

kenjis:
We are using like this:
```php
    {
        $this->setUpMethods[] = 'setUpAddNamespace';

        parent::setUp();
    }
```
And if `$setUpMethods` is removed, we can't write such tests.

Because it modifies the Autoloader, and it is reset in `CIUnitTestCase::setUp()`.

https://github.com/codeigniter4/CodeIgniter4/blob/8e8973634ce98558fd230a663c00475dcd8f2dca/system/Test/CIUnitTestCase.php#L228-L234

https://github.com/codeigniter4/CodeIgniter4/blob/8e8973634ce98558fd230a663c00475dcd8f2dca/system/Test/CIUnitTestCase.php#L504-L507

MGatner (changes requested):
I think at least mentioning the format of the name would be good.

MGatner (approved):

paulbalandan (approved):

PR #6120, opened 2022-06-14
fix: `Time::humanize()` causes error with ar locale

### LINKED ISSUE #4708: Bug: Time::humanize() does not work with ar locale

kenjis:
Need to rebase after merging #6116

**Description**
Fixes #4708
Related #6116

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -82,7 +82,7 @@ class Time extends DateTime
         // If a test instance has been provided, use it instead.
         if ($time === '' && static::$testNow instanceof self) {
             $timezone = $timezone ?: static::$testNow->getTimezone();
-            $time     = (string) static::$testNow->toDateTimeString();
+            $time     = static::$testNow->format('Y-m-d H:i:s');
         }
 
         $timezone       = $timezone ?: date_default_timezone_get();
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -1005,7 +1005,7 @@ class Time extends DateTime
      */
     public function humanize()
     {
-        $now  = IntlCalendar::fromDateTime(self::now($this->timezone)->toDateTimeString());
+        $now  = IntlCalendar::fromDateTime(self::now($this->timezone));
         $time = $this->getCalendar()->getTime();
 
         $years   = $now->fieldDifference($time, IntlCalendar::FIELD_YEAR);
@@ -1106,7 +1106,7 @@ class Time extends DateTime
      */
     public function getCalendar()
     {
-        return IntlCalendar::fromDateTime($this->toDateTimeString());
+        return IntlCalendar::fromDateTime($this);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/system/I18n/TimeTest.php
```diff
@@ -1025,6 +1027,31 @@ public function testHumanizeNow()
...
+        $config->supportedLocales = ['ar'];
+        $config->defaultLocale    = 'ar';
+        Factories::injectMock('config', 'App', $config);
+
+        Time::setTestNow('2022-06-14 12:00', 'America/Chicago');
+
+        $date = '2022-06-07 12:00';
+        $time = Time::parse($date, 'America/Chicago');
+
+        $this->assertSame('١ week ago', $time->humanize());
```
kenjis, line 1050: I don't know `١` is correct or not.

datamweb: Numbers are similar in Arabic and Persian **(١ farsi = ١ arabic)**.

kenjis: https://www.saitak.com/Number/
Okay, I got `1` is `۱`.

datamweb:
@kenjis also had this problem for the Persian language(fa).
The problem is **now solved** (after [15cd4ca](https://github.com/codeigniter4/CodeIgniter4/pull/6120/commits/15cd4ca05020b1aa25090ff4186aa6fa97ccc0a1),)
The output is as expected.
```php
$time = Time::parse('March 9, 2016 12:00:00', 'Asia/Tehran');
dd ($time->humanize());
```
`$time->humanize() UTF-8 string (16) "۶ سال پیش"`

**Thanks**

tests/system/I18n/TimeTest.php
MGatner: It's the same in Hebrew, we use characters for numbers: `א` = 1

It would be like `a` = 1

system/I18n/Time.php
```diff
@@ -1005,7 +1005,7 @@ public function isAfter($testTime, ?string $timezone = null): bool
      */
     public function humanize()
     {
-        $now  = IntlCalendar::fromDateTime(self::now($this->timezone)->toDateTimeString());
+        $now  = IntlCalendar::fromDateTime(self::now($this->timezone));
```
MGatner, line 1008: That's kind of funny that we were using an unnecessary intermediate string to begin with 😅

MGatner (approved):
Looks good, don't need to re-review after rebase.

PR #6133, opened 2022-06-16
feat: add methods to modify files in Publisher

kenjis:
**Description**
To modify config files during setup.

- add `Publisher::replace()`
- add `Publisher::addLineAfter()`
- add `Publisher::addLineBefore()`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Publisher/ContentReplacer.php
+++ b/system/Publisher/ContentReplacer.php
@@ -32,7 +32,7 @@ class ContentReplacer
      *
      * @return bool|string true: already updated, false: regexp error.
      */
-    public function add(string $content, string $text, string $pattern, string $replace)
+    private function add(string $content, string $text, string $pattern, string $replace)
     {
         $return = preg_match('/' . preg_quote($text, '/') . '/u', $content);
 
--- a/system/Publisher/ContentReplacer.php
+++ b/system/Publisher/ContentReplacer.php
@@ -12,7 +12,6 @@
 namespace CodeIgniter\Publisher;
 
 use RuntimeException;
-use stdClass;
 
 class ContentReplacer
 {
@@ -33,12 +32,10 @@ class ContentReplacer
      * @param string $pattern Regexp search pattern.
      * @param string $replace Regexp replacement including text to add.
      *
-     * @return stdClass {updated: bool, content: string}
+     * @return string|null Updated content, or null if not updated.
      */
-    private function add(string $content, string $text, string $pattern, string $replace): stdClass
+    private function add(string $content, string $text, string $pattern, string $replace): ?string
     {
-        $result = new stdClass();
-
         $return = preg_match('/' . preg_quote($text, '/') . '/u', $content);
 
         if ($return === false) {
@@ -48,10 +45,7 @@ class ContentReplacer
 
         if ($return === 1) {
             // It has already been updated.
-            $result->updated = false;
-            $result->content = $content;
-
-            return $result;
+            return null;
         }
 
         $return = preg_replace($pattern, $replace, $content);
@@ -61,10 +55,7 @@ class ContentReplacer
             throw new RuntimeException('Regex error. PCRE error code: ' . preg_last_error());
         }
 
-        $result->updated = true;
-        $result->content = $return;
-
-        return $result;
+        return $return;
     }
 
     /**
@@ -74,9 +65,9 @@ class ContentReplacer
      * @param string $line    Line to add.
      * @param string $after   String to search.
      *
-     * @return stdClass {updated: bool, content: string}
+     * @return string|null Updated content, or null if not updated.
      */
-    public function addAfter(string $content, string $line, string $after): stdClass
+    public function addAfter(string $content, string $line, string $after): ?string
     {
         $pattern = '/(.*)(\n[^\n]*?' . preg_quote($after, '/') . '[^\n]*?\n)/su';
         $replace = '$1$2' . $line . "\n";
@@ -91,9 +82,9 @@ class ContentReplacer
      * @param string $line    Line to add.
      * @param string $before  String to search.
      *
-     * @return stdClass {updated: bool, content: string}
+     * @return string|null Updated content, or null if not updated.
      */
-    public function addBefore(string $content, string $line, string $before): stdClass
+    public function addBefore(string $content, string $line, string $before): ?string
     {
         $pattern = '/(\n)([^\n]*?' . preg_quote($before, '/') . ')(.*)/su';
         $replace = '$1' . $line . "\n" . '$2$3';
--- a/system/Publisher/ContentReplacer.php
+++ b/system/Publisher/ContentReplacer.php
@@ -13,6 +13,9 @@ namespace CodeIgniter\Publisher;
 
 use RuntimeException;
 
+/**
+ * Replace Text Content
+ */
 class ContentReplacer
 {
     /**
--- a/system/Publisher/ContentReplacer.php
+++ b/system/Publisher/ContentReplacer.php
@@ -1,0 +1,87 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Publisher;
+
+class ContentReplacer
+{
+    /**
+     * Replace content
+     *
+     * @param array $replaces [search => replace]
+     */
+    public function replace(string $content, array $replaces): string
+    {
+        return strtr($content, $replaces);
+    }
+
+    /**
+     * Add text
+     *
+     * @param string $text    Text to add.
+     * @param string $pattern Regexp search pattern.
+     * @param string $replace Regexp replacement including text to add.
+     *
+     * @return bool|string true: already updated, false: regexp error.
+     */
+    public function add(string $content, string $text, string $pattern, string $replace)
+    {
+        $return = preg_match('/' . preg_quote($text, '/') . '/u', $content);
+
+        if ($return === 1) {
+            // It has already been updated.
+
+            return true;
+        }
+
+        if ($return === false) {
+            // Regexp error.
+
+            return false;
+        }
+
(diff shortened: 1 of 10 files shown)


datamweb:
@kenjis Just out of curiosity, wouldn't it be better if it was `trait` ?

system/Publisher/ContentReplacer.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Add text
+     *
+     * @param string $text    Text to add.
+     * @param string $pattern Regexp search pattern.
+     * @param string $replace Regexp replacement including text to add.
+     *
+     * @return bool|string true: already updated, false: regexp error.
```
MGatner, line 33: I find these return types pretty confusing. What about returning `?string` (`null` being "already added") and then throwing an exception for regex failure?

system/Publisher/ContentReplacer.php
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public function add(string $content, string $text, string $pattern, string $replace)
```
MGatner, line 35: Any reason this needs to be public?
```suggestion
    private function add(string $content, string $text, string $pattern, string $replace)
```

MGatner (changes requested):
Wow @kenjis this is awesome! This was on my "long list" to get to down the road. So many places we would use Publisher we end up doing the same work manually because we need to replace namespaces. 

A few thoughts on the code. One more thing to mention: I regret bundling Publisher with the framework: it should be a dev tool that doesn't exist on production instances (helps security), and releasing it as a separate package will keep it on a healthier release cycle. Nothing to be done about that now, but let's keep in mind that this will probably move out in v5.

MGatner:
@datamweb the trait idea isn't a bad suggestion. The Publisher class is already too big so I like finding ways to split it out. I don't ever see the replacer being a swappable class.

MGatner:
I'll avoid #2 for now because I have no experience with that myself and I'm not sure what it offers to maintain. 

For #1, this has been one of my favorite features of CodeIgniter. We accomplish a *lot* by allowing dynamic configuration, whereas other array-based frameworks require "Config middleware" which gets pretty strange. We do have some issues, like inheritance collision and source obscurity, but I still think it is a strong asset of CodeIgniter.

iRedds:
Config classes contain logic, which turns them from files with settings into full-fledged scripts.
The database config also creates a connection to the database. This is really strange. =)

kenjis:
> I think there are deeper problems here.
> 1. Using classes for configs instead of arrays (namespace dependency)
> 2. Ability to change the namespace and the real path to the application.

I don't know what the exact problems here. Can you explain details?

What's wrong with classes?
This PR can change any namespaces. Of cause, it may not be so easy.

MGatner:
I agree that the database Config creating the connection is strange. I'm not saying there aren't oddities in there, because there are aplenty, but the general concept of Config as a class I really like. For example I've always thought it strange that other frameworks couldn't type their Config parameters for dependent classes.

iRedds:
@kenjis In my opinion, CI adds too much dynamics. If I know that app controllers are always stored in root + app/Controllers and namespace is App\Controllers. It will not be necessary to prompt the user of the library to manually copy and change the namespace. Just tell the publisher what to copy from here to there.

The class, as config, also depends on the namespace. You cannot use variables in properties or include a file for an extension:
```php
return [
    'a' => require 'config.extention.php'  // return ['a', 'b'];
];
```
Of course, such functionality can be done in the class, but at what cost =)
Just like the config from the database, it is very easy to export to an array than to a class.
And much more that can be done with an array easier than with a class.

iRedds:
> For example I've always thought it strange that other frameworks couldn't type their Config parameters for dependent classes.

Can you clarify what you mean. I do not understand )

MGatner:
@iRedds for example if I have a `WidgetFactory` that has configurable properties in CodeIgniter we can declare the Config as a dependency:
```php
use Config\WidgetConfig;

class WidgetFactory
{
    protected WidgetConfig $config;

    public __construct(?WidgetConfig $config = null)
    {
        $this->config = $config ?? config('WidgetConfig');
    }

    ...
```

This helps strongly type our classes while still defining the dependency external to the class itself. If your configurations are arrays then passing `array $config` into the constructor tells you nothing about what you're actually getting.

iRedds:
@MGatner ``WidgetConfig $config`` is the same as ``array $config``. In both cases, you can pass an empty class/empty array.
This could be somehow justified by the presence of autowiring, but there is no service container or autowiring in the framework. 

Maybe other frameworks just don't see the point in this overhead.

kenjis:
@iRedds It seems your point is costs to manipulate the config values.
It is true that some cases are easier in arrays, but throwing classes and going back to arrays is not realistic.
Mankind invented classes after arrays.

>  If I know that app controllers are always stored in root + app/Controllers and namespace is App\Controllers.

How do you implement HMVC or Module?

> The class, as config, also depends on the namespace. You cannot use variables in properties or include a file for an extension:

Something like this?
```php
public function __construct()
{
    parent::__construct();

    $this->a = require 'config.extention.php';
}
```

kenjis:
@iRedds MGatner wants to say, `WidgetConfig` can type all its property values, an array can't at all.

system/Publisher/ContentReplacer.php
kenjis: I've changed return type completely.
Because `preg_replace()` returns `null` when error occurs.

>If matches are found, the new subject will be returned, otherwise subject will be returned unchanged or null if an error occurred. 
https://www.php.net/manual/en/function.preg-replace.php#refsect1-function.preg-replace-returnvalues

MGatner:
Yet another thread hijacked for Config debates 😂 We need to start a major issue or discussion or project card for this. 
***
Back to the PR at hand, I am very interested in this feature and think that a number of components and official libraries could be refactored to take advantage of it. What remains in order to get this merged?

system/Publisher/ContentReplacer.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Add text
+     *
+     * @param string $text    Text to add.
+     * @param string $pattern Regexp search pattern.
+     * @param string $replace Regexp replacement including text to add.
+     *
+     * @return stdClass {updated: bool, content: string}
```
MGatner, line 36: I'm not opposed to the monad return, but it doesn't seem to be necessary since `content` is irrelevant when `updated === false`. This ends up being functionally the same as `?string`, where `null` means "not updated".

system/Publisher/ContentReplacer.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    private function add(string $content, string $text, string $pattern, string $replace): stdClass
+    {
+        $result = new stdClass();
+
+        $return = preg_match('/' . preg_quote($text, '/') . '/u', $content);
+
+        if ($return === false) {
+            // Regexp error.
+            throw new RuntimeException(preg_last_error_msg());
```
kenjis, line 46: preg_last_error_msg() requires PHP8.

MGatner: 😕 You're right. Looks like we can use https://www.php.net/manual/en/function.preg-last-error.php with the constants, but it's definitely more handling. Maybe just make the error message something like "Regex failed with error #__"?

kenjis: Yes. Done.

system/Publisher/ContentReplacer.php
kenjis: `$result->updated` always shows it is updated or not.
`$result->content` always has the content.

Yes, we can use `?string`, but we need to check if it is null, and if not we update the content.

Personally, I think return values with changing meanings, such as `?string`, are a bit difficult to understand.

MGatner: I like the concept of monads (see recent Slack article and thread) but we don't use them anywhere else in the entire framework. In this particular case it increases clarity but doesn't alter efficiency - you are either checking `$result->updated === false` or `$result === null`. 

If this is a development direction you think the framework should take I would rather lean into it and create a framework Monad class that we could use everywhere. If you want to use this here I would prefer to keep these returns internal, either with a docblock or changing the scope on these methods.

MGatner: I should add more personally: I'm pretty new to the concept of monads. I do think they can add a lot of clarity to returns but they definitely come with some functional overhead and they obscure things for new developers. `?string` is very clear that about what the return choices are here, even if there is not a clear mapping of type to "intent"; using `stdObject` or `Monad` requires "unpacking" the object to understand what type is inside it and (for someone who might be seeing it for the first time) how to determine success/failure. 

For these reasons I tend to think monads are more successful on large projects with either private code or a consistent team, where their strengths can be used without any of the weaknesses. For smaller, quicker, or public projects I prefer the simpler return types and either using `null` for "failure"/"no data" or using a straight type and throwing an exception that is meant to be caught. Both of these approaches accomplish the same thing functionally and are easier to understand if you are seeing them the first time in your IDE or docblock.

system/Publisher/ContentReplacer.php
MGatner: Looks good!

kenjis:
@iRedds I don't understand the shortcomings of the Config classes.
I just understand you don't like Config classes, and you like Config array,
and in some cases we can handle arrays easier than objects.
I recommend you post in Forum or open an issue.

I will not mention Config here anymore.

system/Publisher/ContentReplacer.php
kenjis: I did not intend to recommend using Monad in the framework.
It was just a Result object, but it seems to use stdClass is not good practice.
Because the structure of stdClass object is not clear.

I've changed the all return types to `?string`, because there is less code to be changed.

MGatner:
@iRedds no need to apologize, I always appreciate your input even if I don't agree. You're an important voice around here.

I want to have these discussions, and it is good to be reminded of them whenever a PR touches on the subject, but since changing Config behavior is way beyond the scope of this PR it is a conversation that belongs elsewhere.

kenjis:
Added the docs.

MGatner (approved):
🥳

PR #6134, opened 2022-06-16
Updated to use ::class in $returnType example

tangix:
**Description**
Updated Model example with `$returnType` to use `::class` instead of string literal.
https://forum.codeigniter.com/showthread.php?tid=82127&pid=397352#pid397352

**Checklist:**
- [X] Securely signed commits
- [X] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/user_guide_src/source/models/model.rst
+++ b/user_guide_src/source/models/model.rst
@@ -116,7 +116,7 @@ The Model's CRUD methods will take a step of work away from you and automaticall
 the resulting data, instead of the Result object. This setting allows you to define
 the type of data that is returned. Valid values are '**array**' (the default), '**object**', or the **fully
 qualified name of a class** that can be used with the Result object's ``getCustomResultObject()``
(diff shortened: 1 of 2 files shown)


user_guide_src/source/models/model.rst
```diff
@@ -116,7 +116,8 @@ The Model's CRUD methods will take a step of work away from you and automaticall
 the resulting data, instead of the Result object. This setting allows you to define
 the type of data that is returned. Valid values are '**array**' (the default), '**object**', or the **fully
 qualified name of a class** that can be used with the Result object's ``getCustomResultObject()``
-method.
+method. Using the ``::class`` suffix of the class will allow most IDEs to
```
paulbalandan, line 119: The `::class` is not a suffix but a [special constant](https://www.php.net/manual/en/language.oop5.constants.php) of the class.
```suggestion
method. Using the special ``::class`` constant of the class will allow most IDEs to 
```

kenjis (approved):

Issue #6132, opened 2022-06-16
Dev: Database config that always throws Exceptions


kenjis:
Now in the `development` environment CI throws Exceptions when DB errors occur,
but in the `production` environment, CI returns false.

So most devs don't check query return values, and DB errors in the production environment
will be hidden by the configuration.

See https://github.com/codeigniter4/CodeIgniter4/pull/6115#issuecomment-1156333145

Ref
-  #6127

kenjis:
@MGatner I've checked all `$DBDebug`, and confirmed the behavior changes are only whether throwing an exception or returning false (or just ignored).

And I found that if I always want Exceptions thrown, all I need to do is this:
```diff
--- a/app/Config/Database.php
+++ b/app/Config/Database.php
@@ -39,7 +39,7 @@ class Database extends Config
         'DBDriver' => 'MySQLi',
         'DBPrefix' => '',
         'pConnect' => false,
-        'DBDebug'  => (ENVIRONMENT !== 'production'),
+        'DBDebug'  => true,
         'charset'  => 'utf8',
         'DBCollat' => 'utf8_general_ci',
         'swapPre'  => '',
```

The only issue is the name of the property `DBDebug` is not good.

```diff
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -124,7 +124,7 @@ abstract class BaseConnection implements ConnectionInterface
     /**
      * Debug flag
      *
-     * Whether to display error messages.
+     * Whether to throw Exception or not when an error occurs.
      *
      * @var bool
      */
```

MGatner:
Oh nice! That's a lot easier than I expected. Yes I will switch all my projects to `true`, much better to have the actual database error then some strange code errors farther down the stack. 

I agree the name is unfortunate but probably not worth fiddling with if the functionality is all there.

kenjis:
Yes, `DBDebug` is not a good name, but changing the name would have too great an impact,
and gain almost nothing.

PR #6140, opened 2022-06-17
config: DB Error always throws Exception

kenjis:
**Description**
See  #6132
- change the default config to always throw an exception when DB error occurs

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Database.php
+++ b/app/Config/Database.php
@@ -39,7 +39,7 @@ class Database extends Config
         'DBDriver' => 'MySQLi',
         'DBPrefix' => '',
         'pConnect' => false,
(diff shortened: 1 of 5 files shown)


MGatner (changes requested):
Please add a note to the changelog. Developers would need to apply this change for it to affect them but lots of us use upgrade tools and I don't want someone to miss that accidentally. 

Edit: I just noticed this was a draft, you were probably planning to do that already 😅

kenjis:
Added changelog.

MGatner (approved):
Small suggestion, looks good!

Issue #6145, opened 2022-06-17
Bug: BaseBuilder increment/decrement do not reset state after a query.


iRedds:
### PHP Version

7.4

### CodeIgniter4 Version

4.2.1

### CodeIgniter4 Installation Method

Git

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

cli-server (PHP built-in webserver)

### What happened?

If you call several increment/decrement in a row, then the state is not reset.

### Steps to Reproduce

```php
$db = db_connect()->table('table');

$db->where('id', 1)->increment('value'); // UPDATE `table` SET `value` = `value` + 1 WHERE `id` = 1;
$db->where('id', 2)->increment('value'); // UPDATE `table` SET `value` = `value` + 1 WHERE `id` = 1 AND `id` = 2;
```

kenjis:
Fixed by #6146

PR #6146, opened 2022-06-17
Fixed: BaseBuilder increment/decrement do not reset state after a query

iRedds:
**Description**
https://github.com/codeigniter4/CodeIgniter4/issues/6145

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2369,7 +2369,13 @@ class BaseBuilder
 
         $sql = $this->_update($this->QBFrom[0], [$column => "{$column} + {$value}"]);
 
-        return $this->db->query($sql, $this->binds, false);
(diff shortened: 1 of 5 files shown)


MGatner (approved):
I'm not familiar with `resetWrite()` but the tests look good and the PR is quality. Hopefully someone from the database-savvy crew takes a look.

iRedds:
@MGatner After each query, the state of the object is reset so that the next query does not include data from the previous one. There was no such state reset for increment/decrement methods.
The resetWrite() and resetSelect() methods are responsible for resetting the state for queries that change data and selects, respectively.

MGatner:
Thanks for the explanation!

kenjis (approved):

PR #6147, opened 2022-06-17
Update Exceptions.php - determineCodes crash

BrolyTheVVF:
**Description**
In `System/Debug/Exceptions.php::determineCodes`
I've been having error with my PDO (custom port of C3) Cause the errors codes are not numeric, and the function 'determineCodes' uses the abs function directly on the result of `$exception->getCode()`
I've just added a check for numeric value before using abs, and non numeric codes right now are treated as 500

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -300,18 +300,23 @@ class Exceptions
      */
     protected function determineCodes(Throwable $exception): array
     {
-        $statusCode = abs($exception->getCode());
+        $statusCode = $exception->getCode();
+        if(is_numeric($statusCode)){
+            $statusCode = abs($statusCode);
+            if ($statusCode < 100 || $statusCode > 599) {
+                $exitStatus = $statusCode + EXIT__AUTO_MIN;
 
-        if ($statusCode < 100 || $statusCode > 599) {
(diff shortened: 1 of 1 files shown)


kenjis:
> Returns the exception code as int in [Exception](https://www.php.net/manual/en/class.exception.php) but possibly as other type in [Exception](https://www.php.net/manual/en/class.exception.php) descendants (for example as string in [PDOException](https://www.php.net/manual/en/class.pdoexception.php)).
https://www.php.net/manual/en/exception.getcode.php#refsect1-exception.getcode-returnvalues

kenjis:
Thank you for contributing!

First, you forgot to sign your commits.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#signing

Second, there are coding standard violations. See GitHub Action checks.
Please fix the coding style.

BrolyTheVVF:
I can't get cs-fix to work properly on my end, i'll make another PR when i found how that works, if it's not fixed on your end already

kenjis:
What's error message?

kenjis:
And we need unit testing.

PR #5118, opened 2021-09-23
fix: QueryBuilder breaks select when escape is false

kenjis:
**Description**
fixes #4311
fixes #4355

I'm not sure correct behavior of `protectIdentifiers()`. See [tests](https://github.com/codeigniter4/CodeIgniter4/pull/5118/files#diff-65d39fd097198820866d41c227f1afbb8db3498670ad049d13f87ad6df2cc97e).

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPdocs
- [x] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #6153, opened 2022-06-18
Add missing escape parameter for updateBatch

pixobit:

- All bug fixes should be sent to the __"develop"__ branch, this is where the next bug fix version will be developed.
- PRs with any enhancement should be sent to the next minor version branch, e.g. __"4.3"__

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


kenjis:
Thank you for sending a PR.

Can you explain what you have changed, and why?

kenjis:
This is a breaking change.
https://3v4l.org/3Dn3D

pixobit:
Sorry, I thought the title was self explanatory. Basically most of the query builder methods have a boolean escape parameter, which can be set to false if you pass something that you don't want to escape, since it would break the query. This happens for rare edge cases, when you want to inject a sql function or something, and you do the escaping yourself.

I noticed the updateBatch was missing this parameter, even though the "setUpdateBatch" supports it, so I'm guessing it was overlooked or something. This pull request basically is about adding that missing escape parameter

kenjis:
I can't imagine use cases like that. Do you have any specific use case?

Only `updateBatch()` does not have `$escape` parameter, which is not consistent.
But this PR makes:
```php
public function updateBatch(?array $set = null, ?string $index = null, int $batchSize = 100, ?bool $escape = null)
```
and
```php
public function insertBatch(?array $set = null, ?bool $escape = null, int $batchSize = 100)
```
They are still not consistent.

pixobit:
hmm, you're right, they aren't consistent. I guess it is a breaking change then.
There are plenty of use cases, but one that I actually needed was something like this:

`$this->qry->updateBatch(array_map(function($quantityData) {
    $quantity = $this->db->escape($quantityData['quantity']);
    return [
        'id' => $quantityData['id'],
        'quantity' => '(CASE WHEN quantity > ' . $quantity . ' THEN quantity - ' . $quantity . ' ELSE 0 END)'
    ];
}, $data), 'id', 100, false);`

Basically it's a batch update for decreasing the product quantities.

kenjis:
@pixobit This PR includes a breaking change, and this is an enhancement.
We can't merge this into `develop` (which means v4.2.2).
Please change the base branch to `4.3`.

pixobit:
> @pixobit This PR includes a breaking change, and this is an enhancement. We can't merge this into `develop` (which means v4.2.2). Please change the base branch to `4.3`.

How do I change the base branch of the pull request?

pixobit:
Figured out, and changed it!

kenjis:
Please remove unrelated commits. See https://github.com/codeigniter4/CodeIgniter4/pull/6153/commits

1. Create new branch from `upstream/4.3`.
2. Cherry-pick  this PR commit.
3. Delete branch  `patch-2`.
4. Rename new branch `patch-2`
5. Force push

kenjis:
If users need raw SQL string, it seems better to make `RawSql` usable.
https://codeigniter4.github.io/CodeIgniter4/dbmgmt/forge.html#forge-addfield-default-value-rawsql

pixobit:
> If users need raw SQL string, it seems better to make `RawSql` usable. https://codeigniter4.github.io/CodeIgniter4/dbmgmt/forge.html#forge-addfield-default-value-rawsql

Why do you think so? If I use raw sql, then I have to write the whole updateBatch query by myself, while this does the job pretty well. It basically feels like reinventing the wheel, plus this is easier on the eye, since it's consistent with how the updateBatch is being used. So why is the RawSql a better approach?

kenjis:
> So why is the RawSql a better approach?

Good question. Because nobody can review #5118.
The  behavior of `$escape` is too complex and nobody can maintain.

> If I use raw sql, then I have to write the whole updateBatch query by myself

Why? If we use your way, the code would be like this.

```php
$this->qry->updateBatch(array_map(function($quantityData) {
    $quantity = $quantityData['quantity'];
    return [
        'id' => $quantityData['id'],
        'quantity' => new RawSql('(CASE WHEN quantity > ' . $quantity . ' THEN quantity - ' . $quantity . ' ELSE 0 END)'),
    ];
}, $data), 'id', 100);
```

pixobit:
Oh, I misunderstood your RawSql suggestion, sorry.
I like this approach better than the escape parameter. Guess this pull request can be removed then.

MGatner:
@pixobit since you seem confident in your database usage would you give a preliminary review of the PR that kenjis linked? We are always short on strong database maintainers and I would love to see more community reviews, especially for that layer.

pixobit:
@MGatner I took a peek, but honestly I rarely dig into the core, it's usually just get to it when I find something that's not working, which is pretty rare :)
But I see michalsn already reviewed it however.

I might try to contribute more in the future though.

PR #6154, opened 2022-06-18
docs: fix how to update next minor upgrade branch

kenjis:
**Description**
- the step to update local 4.x branch was missing

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide


MGatner (commented):
This is my fault for not looking closely. But I actually recommend working off the repo branches directly for release, as it makes the process faster and reduces mistakes:

> Work off direct clones of the repos so the release branches persist for a time

So the upstream remote is irrelevant.

MGatner (changes requested):
It won't let me suggest a change for untouched lines, but please rework for direct branch access.

kenjis:
Fixed the git commands.

PR #6157, opened 2022-06-19
fix: `is_image` causes PHP 8.1 deprecated error

### LINKED ISSUE #6155: Bug: Validation/FileRules/is_image: mb_strpos deprecation bug
### PHP Version

8.1

### CodeIgniter4 Version

4.2.1

Linux

### Which server did you use?

### Database

_No response_

### What happened?

CI4_deprecation_bug

mb_strpos does not allow $haystack to be null in php version 8.1.5

'Mimes::guessTypeFromExtension($file->getExtension());' returns null for files of type 'xcf' and thereby 'mb_strpos()' throws an error.

### Steps to Reproduce

Use 'is_image' as part of a form validation and input a 'xcf' file.

### Expected Output

'mb_strpos()' should not throw an error, so the function returns true or false.

### Anything else?

_No response_

### LINKED ISSUE #6156: Bug: broken test for 'is_image' validation rule.
### PHP Version

8.1

### CodeIgniter4 Version

4.2.1

### What happened?

https://github.com/codeigniter4/CodeIgniter4/blob/44cbd8918569d08db20c6e1906df6f9ee62a11b9/tests/system/Validation/FileRulesTest.php#L223-L234

1. ``$_FILES['stuff']`` though gets into the global array, but is not processed, because the IncomingRequest (service) saves its state, created during the first test.
2. If you reset the services, then an error occurs when you try to get the file extension. 
Explanation: It all starts with this line ``$type = Mimes::guessTypeFromExtension($file->getExtension());`` in the ``is_image`` rule.
``UploadedFile::getExtension()`` -> ``UploadedFile::guessExtension()`` -> ``File::getMimeType()`` -> ``finfo_file - throws an exception that the file is not found.`` Because the file 'tmp_name' => TESTPATH . '_support/Validation/uploads/abc77tz' does not exist, and when an instance of the ``UploadedFile`` class is created, the existence of the file is not checked in the parent constructor (``File``).
3. For php 8.1, the test will fail because ``Mimes::guessTypeFromExtension()`` in the ``is_image`` rule will return null. Example here https://github.com/codeigniter4/CodeIgniter4/issues/6155

Conclusion: the test is successful not because the file is not an image, but because it does not reach the check for an image.

kenjis:
**Description**
Fixes #6155
Fixes #6156

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FileRules.php
+++ b/system/Validation/FileRules.php
@@ -133,7 +133,7 @@ class FileRules
 
             // We know that our mimes list always has the first mime
             // start with `image` even when then are multiple accepted types.
-            $type = Mimes::guessTypeFromExtension($file->getExtension());
+            $type = Mimes::guessTypeFromExtension($file->getExtension()) ?? '';
 
             if (mb_strpos($type, 'image') !== 0) {
                 return false;
--- a/tests/system/Validation/FileRulesTest.php
+++ b/tests/system/Validation/FileRulesTest.php
@@ -41,7 +41,9 @@ final class FileRulesTest extends CIUnitTestCase
 
     protected function setUp(): void
     {
+        $this->resetServices();
         parent::setUp();
+
         $this->validation = new Validation((object) $this->config, Services::renderer());
         $this->validation->reset();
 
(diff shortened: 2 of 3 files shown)


MGatner (approved):

samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #6159, opened 2022-06-19
fix: Time::getAge() calculation

kenjis:
**Description**
- fix `Time::getAge()` calculation
- fix test case

> Returns the age, in years, of between the Time’s instance and the current time. **Perfect for checking the age of someone based on their birthday**:
https://codeigniter4.github.io/CodeIgniter4/libraries/time.html#getage

There are a few ways to calculate human age.
However, the current implementation does not seem to be common.

E.g.:
- birthdate: December 31, 2020
- now: January 1, 2022
- age: 2

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/I18n/Time.php
+++ b/system/I18n/Time.php
@@ -460,12 +460,8 @@ class Time extends DateTime
      */
     public function getAge()
     {
-        $now = self::now();
-
-        $age = (int) (((int) $now->format('Ymd') - (int) $this->format('Ymd')) / 10000);
-
         // future dates have no age
-        return max(0, $age);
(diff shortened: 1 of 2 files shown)


MGatner (commented):
Why not use `Time::difference()`? Or it's underlying method:
```php
   return $time->fieldDifference($this->testTime, IntlCalendar::FIELD_YEAR);
```

kenjis:
Refactored.

samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #6164, opened 2022-06-20
fix: prepared query is executed when using QueryBuilder

kenjis:
**Description**
- prepared query is executed when using QueryBuilder

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -136,11 +136,6 @@ parameters:
 			path: system/Database/BaseConnection.php
 
 		-
-			message: "#^Negated boolean expression is always true\\.$#"
-			count: 2
-			path: system/Database/BaseConnection.php
-
-		-
 			message: "#^Negated boolean expression is always false\\.$#"
 			count: 3
 			path: system/Database/BaseResult.php
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -603,6 +603,15 @@ abstract class BaseConnection implements ConnectionInterface
         // the getLastQuery() method.
         $this->lastQuery = $query;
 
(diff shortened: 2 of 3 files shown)


MGatner (approved):
I'm not very familiar with query pretending but the code logic looks good. Was this bug actually "hiding" because of the PHPStan exemption? Or was that just a coincidence?

lonnieezell:
The changes seem fine but I guess I'm not understanding how this stops the prepared query from being executed with a query builder call.

kenjis:
@lonnieezell See the current `develop` code. Here is the line to execute queries.
https://github.com/codeigniter4/CodeIgniter4/blob/e8f20804ba1e5d685d0f2802d507fee0b11f3324/system/Database/BaseConnection.php#L609
But the line to return when `$pretend` is true is here.
https://github.com/codeigniter4/CodeIgniter4/blob/e8f20804ba1e5d685d0f2802d507fee0b11f3324/system/Database/BaseConnection.php#L664-L669

I have moved this before the `simpleQuery()`.
In the original code, I don't see why the return is so late.

I believe this PR behaves the same as before in Prepared Queries without the bug, 
but please point out later if I have missed something.

kenjis:
> Was this bug actually "hiding" because of the PHPStan exemption?

Probably no.

If we remove the the PHPStan exemption, we get the errors.

```
 ------ -------------------------------------------- 
  Line   system/Database/BaseConnection.php          
 ------ -------------------------------------------- 
  637    Negated boolean expression is always true.  
  649    Negated boolean expression is always true.  
 ------ -------------------------------------------- 
```

It is true that the if conditions of the lines with errors are not necessary, but probably we don't notice this bug.

PR #6168, opened 2022-06-21
Change: PHPDoc return type in ControllerTestTrait methods

iRedds:
**Description**
Replacing the return type in the ControllerTestTrait methods in the docBlock section from mixed to $this.
This allows the IDE to use hints in the chain call.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


samsonasik:
Thank you @iRedds

PR #6169, opened 2022-06-21
Doc: Incorrect example of creating an IncomingRequest instance.

iRedds:
**Description**
There is a mistake in the example code in the documentation for testing controllers using an instance of the IncomingRequest class. The $userAgent argument was not specified.

Although the $userAgent parameter in the constructor of the IncomingRequest class can be null, this will cause an exception to be thrown.
https://github.com/codeigniter4/CodeIgniter4/blob/01804793734610951b4eb4a3e8b1cbfc7826cfc6/system/HTTP/IncomingRequest.php#L150-L154

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


samsonasik:
Thank you @iRedds

PR #6183, opened 2022-06-24
config: DB Error always throws Exception CI_DBUG

kenjis:
**Description**
Follow-up #6140
- use `$DBDebug` instead of `CI_DEBUG`
- change the `BaseConnection::$DBDebug` default value to `true`
- `InvalidArgumentException` in `BaseBuilder::_whereIn()` is never suppressed by config

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -895,19 +895,11 @@ class BaseBuilder
     protected function _whereIn(?string $key = null, $values = null, bool $not = false, string $type = 'AND ', ?bool $escape = null, string $clause = 'QBWhere')
     {
         if (empty($key) || ! is_string($key)) {
-            if ($this->db->DBDebug) {
-                throw new InvalidArgumentException(sprintf('%s() expects $key to be a non-empty string', debug_backtrace(0, 2)[1]['function']));
-            }
-
-            return $this; // @codeCoverageIgnore
+            throw new InvalidArgumentException(sprintf('%s() expects $key to be a non-empty string', debug_backtrace(0, 2)[1]['function']));
         }
 
         if ($values === null || (! is_array($values) && ! $this->isSubquery($values))) {
-            if ($this->db->DBDebug) {
-                throw new InvalidArgumentException(sprintf('%s() expects $values to be of type array or closure', debug_backtrace(0, 2)[1]['function']));
-            }
-
-            return $this; // @codeCoverageIgnore
+            throw new InvalidArgumentException(sprintf('%s() expects $values to be of type array or closure', debug_backtrace(0, 2)[1]['function']));
         }
 
         if (! is_bool($escape)) {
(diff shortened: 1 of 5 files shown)


user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -29,6 +29,7 @@ Changes
 *******
 
 - To be consistent in behavior regardless of environments, ``Config\Database::$default['DBDebug']`` and ``Config\Database::$tests['DBDebug']`` has been changed to ``true`` by default. With these settings, an exception is always thrown when a database error occurs.
+- The default value of ``BaseConnection::$DBDebu`` has been changed to ``true``.
```
samsonasik, line 32: Typo: DBDebu

kenjis: Thanks! Fixed.

system/Database/BaseBuilder.php
```diff
@@ -895,15 +895,15 @@ public function orHavingNotIn(?string $key = null, $values = null, ?bool $escape
     protected function _whereIn(?string $key = null, $values = null, bool $not = false, string $type = 'AND ', ?bool $escape = null, string $clause = 'QBWhere')
     {
         if (empty($key) || ! is_string($key)) {
-            if (CI_DEBUG) {
+            if ($this->db->DBDebug) {
```
MGatner, line 898: I am not certain about these two. This isn't actually a database exception at all, it is a developer error. I'm actually surprised it was guarded behind `CI_DEBUG` to begin with - I think this should be thrown regardless. And invalid argument is a mistake no matter what the environment or context.

MGatner (changes requested):
All the rest looks good. I'm open to discussion on the invalid arguments, but IMO those should always throw regardless.

system/Database/BaseBuilder.php
kenjis: Yes, this throws `InvalidArgumentException`, and in production do nothing.
Removing the `if` is a breaking change. Are you okay?

MGatner: If this exception is typed correctly then, yes: we should allow it. InvalidArgumentException is a type of LogicException which should never be suppressed. By definition (PHP.net):

> Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code.

If there is some conceivable execution path that could cause these two instances then they should be changed to RuntimeExceptions as allow suppression, but my impression looking at them is that we should just let them throw.

kenjis: Agreed. LogicException should not be suppressed.

Removed `if`.

kenjis:
Added the documentation.

MGatner (approved):

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -27,6 +27,7 @@ Others
 ------
 
 - The ``spark`` file has been changed due to a change in the processing of Spark commands.
+- ``InvalidArgumentException`` that is a kind of ``LogicException`` in ``BaseBuilder::_whereIn()`` is not suppressed by the configuration. Previously if ``CI_DEBUG`` was true, the exception was suppressed.
```
paulbalandan, line 30: I think you mean `CI_DEBUG` is false, the exception was suppressed.

kenjis: Yes! Thanks.

paulbalandan (approved):

PR #6184, opened 2022-06-24
fix: Email SMTP may throw Uncaught ErrorException

### LINKED ISSUE #3390: Error: Email SMTP may throw Uncaught ErrorException
Hello guys i found an Uncaught ErrorException: fwrite():

```
Fatal error: Uncaught ErrorException: fwrite(): send of 6 bytes failed with errno=32 Broken pipe in /path/system/Email/Email.php:2019 
Stack trace: #0 [internal function]: 
CodeIgniter\Debug\Exceptions->errorHandler(8, 'fwrite(): send ...', '/path...', 2019, Array) #1 /path/system/Email/Email.php(2019): 
fwrite(Resource id #10, 'QUIT\r\n') #2 /path/system/Email/Email.php(1941): 
CodeIgniter\Email\Email->sendData('QUIT\r\n') #3 /path/system/Email/Email.php(2139): 
CodeIgniter\Email\Email->sendCommand('quit') #4 [internal function]: 
CodeIgniter\Email\Email->__destruct() #5 {main} thrown in /path/system/Email/Email.php on line 2019
```

When the configuration of the email is wrong this error appears almost in every case

kenjis:
**Description**
Fixes #3390

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Email/EmailTest.php
+++ b/tests/system/Email/EmailTest.php
@@ -44,9 +44,8 @@ final class EmailTest extends CIUnitTestCase
      */
     public function testEmailSendWithClearance($autoClear)
     {
-        $config           = config('Email');
-        $config->validate = true;
-        $email            = new MockEmail($config);
+        $email = $this->createMockEmail();
+
         $email->setTo('foo@foo.com');
 
         $this->assertTrue($email->send($autoClear));
@@ -58,9 +57,8 @@ final class EmailTest extends CIUnitTestCase
 
     public function testEmailSendStoresArchive()
     {
-        $config           = config('Email');
-        $config->validate = true;
-        $email            = new MockEmail($config);
+        $email = $this->createMockEmail();
+
         $email->setTo('foo@foo.com');
         $email->setFrom('bar@foo.com');
         $email->setSubject('Archive Test');
@@ -75,9 +73,8 @@ final class EmailTest extends CIUnitTestCase
 
     public function testAutoClearLeavesArchive()
     {
-        $config           = config('Email');
-        $config->validate = true;
-        $email            = new MockEmail($config);
+        $email = $this->createMockEmail();
+
         $email->setTo('foo@foo.com');
 
         $this->assertTrue($email->send(true));
@@ -89,6 +86,7 @@ final class EmailTest extends CIUnitTestCase
     {
         $config = config('Email');
         $email  = new MockEmail($config);
+
         $email->setTo('foo@foo.com');
         $email->setFrom('bar@foo.com');
 
@@ -104,9 +102,8 @@ final class EmailTest extends CIUnitTestCase
 
     public function testSuccessDoesTriggerEvent()
     {
-        $config           = config('Email');
-        $config->validate = true;
-        $email            = new MockEmail($config);
+        $email = $this->createMockEmail();
+
         $email->setTo('foo@foo.com');
 
         $result = null;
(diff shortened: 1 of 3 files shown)


kenjis:
@samsonasik The following refactoring breaks test code.

> Method name is not configured

```diff
1) tests/system/Email/EmailTest.php:138

    ---------- begin diff ----------
@@ @@

     public function testDestructDoesNotThrowException()
     {
-        $email = $this->getMockBuilder(Email::class)
-            ->disableOriginalConstructor()
-            ->onlyMethods(['sendCommand'])
-            ->getMock();
+        $email = $this->createMock(Email::class);
         $email->method('sendCommand')
-            ->will($this->throwException(new ErrorException('SMTP Error.')));
+            ->willThrowException(new ErrorException('SMTP Error.'));

         // Force resource to be injected into the property
         $SMTPConnect = fopen(__FILE__, 'rb');
    ----------- end diff -----------

Applied rules:
 * GetMockBuilderGetMockToCreateMockRector
 * UseSpecificWillMethodRector
```

samsonasik:
@kenjis could you create issue on rector-phpunit repo? Should be bug on `GetMockBuilderGetMockToCreateMockRector`

kenjis:
@samsonasik Created: https://github.com/rectorphp/rector-phpunit/issues/86

MGatner (approved):
Looks good pending the Rector PR.

paulbalandan:
We can use `$this->createPartialMock(Email::class, ['sendCommand']);` for brevity. I think this is what Rector should have done.

tests/system/Email/EmailTest.php
```diff
@@ -139,4 +136,30 @@ public function testFailureDoesNotTriggerEvent()
...
+        $email->method('sendCommand')
+            ->willThrowException(new ErrorException('SMTP Error.'));
+
+        // Force resource to be injected into the property
+        $SMTPConnect = fopen(__FILE__, 'rb');
+        $this->setPrivateProperty($email, 'SMTPConnect', $SMTPConnect);
+
+        $email->__destruct();
+
+        $this->assertTrue(true);
```
paulbalandan, line 155: Is this necessary? If I remember, the invocation call using `will*()` (combined with `expects()`)  is effectively an assertion.

kenjis:
> We can use $this->createPartialMock(Email::class, ['sendCommand']); for brevity. I think this is what Rector should have done.

`createPartialMock()` uses `setMethods()` that is deprecated.
So using `createPartialMock()` changes the behavior.

tests/system/Email/EmailTest.php
kenjis: `expects($this->any())` causes still the warning, but `$this->once()` works fine.

paulbalandan:
> > We can use $this->createPartialMock(Email::class, ['sendCommand']); for brevity. I think this is what Rector should have done.
> 
> `createPartialMock()` uses `setMethods()` that is deprecated. So using `createPartialMock()` changes the behavior.

`createPartialMock` is not deprecated. Only the implementation detail is deprecated. To this, it is not our job to check if the hidden implementation detail uses deprecated API if the API consuming this is not deprecated.

`setMethods` is effectively an `onlyMethods`, only in that `onlyMethods` check method existence first.

In PHPUnit 10, `createPartialMock` will use `onlyMethods`.
https://github.com/sebastianbergmann/phpunit/blob/52357158d80913d448a2ed89b54240f9d6fc07c3/src/Framework/TestCase.php#L1380-L1396

kenjis:
@paulbalandan We and rector cannot use `createPartialMock()` in this case.
Because I'm using `onlyMethods()`, not `setMethods()`.
In PHPUnit9 `createPartialMock()` has a bug due to historical reasons, and will be fixed in PHPUnit10.

kenjis:
Rebased.

MGatner:
I appreciate all the conversation and effort to figure out these complicated tests. I also want to acknowledge that Email should be deprecated in 4.3 and is already a bit of a shabby class. If we need to hack the class or skip testing altogether I'm fine with it.

PR #6189, opened 2022-06-27
Config Property Types

MGatner:
**Description**
Applies non-breaking property types to Config files in `App`. Applied by Rector.

**Checklist:**
- [X] Securely signed commits
- [X] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -21,10 +21,8 @@ class App extends BaseConfig
      * and path to your installation. However, you should always configure this
      * explicitly and never rely on auto-guessing, especially in production
      * environments.
-     *
-     * @var string
      */
-    public $baseURL = 'http://localhost:8080/';
+    public string $baseURL = 'http://localhost:8080/';
 
     /**
      * --------------------------------------------------------------------------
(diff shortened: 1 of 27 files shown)


MGatner:
This should all be safe to change with the possible exception of `Email` which has some odd initial definitions.

MGatner:
Actually this may not work because of the class extensions in `CodeIgniter\Test\Mock` config classes. I have to drop this for now but will think on that later.

kenjis:
> Fatal error: Type of CodeIgniter\Test\Mock\MockCLIConfig::$baseURL must be string (as in class Config\App) in /home/runner/work/CodeIgniter4/CodeIgniter4/system/Test/Mock/MockCLIConfig.php on line 40

MGatner:
I saw that, there are a few other Configs in there that break as well. If we want to move forward with this I think the approach would be to remove the properties from the Mock Config files and use their constructors to change the values that need changing. That way they would still work with whatever is in a user's **app/Config** folder, typed or not.

kenjis:
It would be impossible to ensure that an existing app or tests would never break 
because a Type Error would occur if the configuration class were inherited.

MGatner:
But that is only if the developer manually updates their Config folder with the new files. This should be safe with the caveat of these Mock classes, which we can fix internally but can't guarantee that extensions don't exist.

kenjis:
> But that is only if the developer manually updates their Config folder with the new files. 

Yes, you are correct!

kenjis:
#6214 was merged.

PR #6192, opened 2022-06-28
docs: add note about special characters in paths

kenjis:
**Description**
Fixes  #6190

- add note about special characters in paths

**Checklist:**
- [x] Securely signed commits
- [] Component(s) with PHPDoc blocks, only if necessary or adds value
- [] Unit testing, with >80% coverage
- [x] User guide updated
- [] Conforms to style guide

PR #6193, opened 2022-06-28
feat: Autoloader::sanitizeFilename() throws Exception

kenjis:
**Description**
Generally Sanitization is bad practice.

before:
```console
$ php spark
```
> PHP Fatal error:  Uncaught Error: Class "Config\App" not found in /Users/kenji/tmp/software(locale)/project-root/vendor/codeigniter4/framework/system/Config/Factories.php:126

after:
```console
$ php spark 
```
> PHP Fatal error:  Uncaught InvalidArgumentException: The file path contains special character that is not allowed: "/Users/kenji/tmp/software(locale)/project-root/vendor/codeigniter4/framework/system/Autoloader/FileLocator.php" in /Users/kenji/tmp/software(locale)/project-root/vendor/codeigniter4/framework/system/Autoloader/Autoloader.php:315

Related #6192

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -290,9 +290,9 @@ class Autoloader
     }
 
     /**
-     * Sanitizes a filename, replacing spaces with dashes.
+     * Check file path.
      *
-     * Removes special characters that are illegal in filenames on certain
+     * Checks special characters that are illegal in filenames on certain
      * operating systems and special characters requiring special escaping
      * to manipulate at the command line. Replaces spaces and consecutive
      * dashes with a single dash. Trim period, dash and underscore from beginning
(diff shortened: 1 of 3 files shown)


system/Autoloader/Autoloader.php
```diff
@@ -306,10 +306,16 @@ public function sanitizeFilename(string $filename): string
...
         // Modified to allow backslash and colons for on Windows machines.
-        $filename = preg_replace('/[^0-9\p{L}\s\/\-\_\.\:\\\\]/u', '', $filename);
+        $tmp = preg_replace('/[^0-9\p{L}\s\/\-\_\.\:\\\\]/u', '', $filename);
 
         // Clean up our filename edges.
-        return trim($filename, '.-_');
+        $cleanFilename = trim($tmp, '.-_');
+
+        if ($filename !== $cleanFilename) {
+            throw new InvalidArgumentException('The file path contains special character that is not allowed: "' . $filename . '"');
```
paulbalandan, line 315: I think for better developer experience, the exception message would contain which disallowed characters were included in the filename. Like `The file path contains special characters "()" that are not allowed.`. Maybe use `preg_match_all` and implode the `$matches` into a single string, something like that?

kenjis: Improved exception message.

paulbalandan (approved):

system/Autoloader/Autoloader.php
```diff
@@ -306,10 +307,28 @@ public function sanitizeFilename(string $filename): string
...
+        if ($result > 0) {
+            $chars = implode('', $matches[0]);
+
+            throw new InvalidArgumentException(
+                'The file path contains special characters "' . $chars
+                . '" that are not allowed: "' . $filename . '"'
+            );
+        }
+        if ($result === false) {
+            throw new RuntimeException(preg_last_error_msg() . ' filename: "' . $filename . '"');
```
MGatner, line 321: When would this occur? `preg_match_all()`:

> Returns the number of full pattern matches (which might be zero), or false on failure.

kenjis: When regex error occurs.

MGatner: But *how* can it occur? The regex pattern is fixed so we know it won't cause an error, and this function should work with any string input which is enforced by the parameter `string $filename`. I'm fine with extra sanity checks but if we can't actually create the scenario it will lower our code coverage.

kenjis: Added test.

MGatner (approved):

Issue #6162, opened 2022-06-20
Adding dynamic properties by Registrars will be deprecated on PHP 8.2


kenjis:
Registrars can add dynamic properties now.

But the sample code in the user guide will cause the error on PHP 8.2.
https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#explicit-registrars

```console
bash-3.2$ php public/index.php 


[ErrorException]

Creation of dynamic property Config\MySalesConfig::$actual is deprecated

at SYSTEMPATH/Config/BaseConfig.php:211

Backtrace:
  1    SYSTEMPATH/Config/BaseConfig.php:211
       CodeIgniter\Debug\Exceptions()->errorHandler(8192, 'Creation of dynamic property Config\\MySalesConfig::$actual is deprecated', '.../CodeIgniter4/system/Config/BaseConfig.php', 211)

  2    SYSTEMPATH/Config/BaseConfig.php:64
       CodeIgniter\Config\BaseConfig()->registerProperties()

  3    APPPATH/Controllers/Home.php:9
       CodeIgniter\Config\BaseConfig()->__construct()

  4    SYSTEMPATH/CodeIgniter.php:896
       App\Controllers\Home()->index()

  5    SYSTEMPATH/CodeIgniter.php:466
       CodeIgniter\CodeIgniter()->runController(Object(App\Controllers\Home))

  6    SYSTEMPATH/CodeIgniter.php:349
       CodeIgniter\CodeIgniter()->handleRequest(null, Object(Config\Cache), false)

  7    FCPATH/index.php:55
       CodeIgniter\CodeIgniter()->run()
```

```console
bash-3.2$ php -v
PHP 8.2.0-dev (cli) (built: Jun 20 2022 00:32:41) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.0-dev, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.0-dev, Copyright (c), by Zend Technologies
```

References:
- https://wiki.php.net/rfc/deprecate_dynamic_properties
- #6160

iRedds:
#[AllowDynamicProperties]

kenjis:
The issue is we will continue to use Dynamic Properties that PHP will denies by default or not.

iRedds:
Why not use the Entity::$attributes case?

kenjis:
> Why not use the Entity::$attributes case?

What do you mean?
[...]

PR #6198, opened 2022-06-29
fix: BaseConnection may create dynamic property

kenjis:
**Description**
See https://github.com/codeigniter4/CodeIgniter4/issues/6162#issuecomment-1168005433

- do not create dynamic property in BaseConnection

The current database `tests` group is for SQLite3 that has `foreignKeys` property.
https://github.com/codeigniter4/CodeIgniter4/blob/24f67ea95aa463771360eb38f627675af2fda3d6/app/Config/Database.php#L77
If a user overrides `DBDriver`, the dynamic property `foreignKeys` will be created.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -336,7 +336,9 @@ abstract class BaseConnection implements ConnectionInterface
     public function __construct(array $params)
     {
         foreach ($params as $key => $value) {
-            $this->{$key} = $value;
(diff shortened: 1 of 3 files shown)


MGatner (approved):
It seems like there would be more properties that needed declaring but the tests look good so I guess not.

Issue #6205, opened 2022-06-29
Bug: SQLite DB Duplicate Indexes


MGatner:
### PHP Version

7.4

### CodeIgniter4 Version

4.2.1

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

cli

### Database

SQLite3

### What happened?

I have the following migrations:
1. Create table
2. Modify that table add column
3. Modify that table drop column

The `down()` methods do the reverse, so for example 3) looks like:
```php

    public function up(): void
    {
        $this->forge->dropColumn('actions', [
            'description',
            'summary',
            'category',
            'icon',
            'role',
            'class',
        ]);
    }

    public function down(): void
    {
        $this->forge->addColumn('actions', [
            'class'       => ['type' => 'varchar', 'constraint' => 63, 'null' => true],
            'role'        => ['type' => 'varchar', 'constraint' => 63, 'default' => ''],
            'icon'        => ['type' => 'varchar', 'constraint' => 63, 'default' => ''],
            'category'    => ['type' => 'varchar', 'constraint' => 63, 'default' => ''],
            'summary'     => ['type' => 'varchar', 'constraint' => 255],
            'description' => ['type' => 'text', 'constraint' => 255],
        ]);
    }
```

When running test where the database is migrated up and down multiple times the `dropColumn()` command in the last migration is duplicating indexes (`name` in this case):
```console
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ $sqls                                                                                                                                                                                                │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
array (5) [
    0 => string (56) "CREATE INDEX `db_actions_name` ON `db_actions` (`name`);"
    1 => string (54) "CREATE INDEX `db_actions_uid` ON `db_actions` (`uid`);"
    2 => string (56) "CREATE INDEX `db_actions_name` ON `db_actions` (`name`);"
    3 => string (77) "CREATE INDEX `db_actions_deleted_at_id` ON `db_actions` (`deleted_at`, `id`);"
    4 => string (68) "CREATE INDEX `db_actions_created_at` ON `db_actions` (`created_at`);"
]
[...]

PR #6207, opened 2022-06-29
SQLite3 Connection getIndexData()

sclubricants:
This is a rewrite of `system/Database/SQLite3/Connection::_indexData()`.

SQLite doesn't show indexes on tables created with `INTEGER PRIMARY KEY AUTO_INCREMENT`. This PR looks for primary keys that don't appear in the normal index lookup and adds it to the return of the method. 

```php
        $this->forge->addField([
```

```php
        $this->forge->addField([
```

The need is to get a comprehensive list of database constraints. This will be useful in adding `upsert()` to the database.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "phpstan/phpstan": "^1.7.1",
         "phpunit/phpunit": "^9.1",
         "predis/predis": "^1.1 || ^2.0",
-        "rector/rector": "0.13.6"
+        "rector/rector": "0.13.7"
     },
     "suggest": {
         "ext-fileinfo": "Improves mime type detection for files"
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -912,7 +912,7 @@ class CodeIgniter
         }
 
         throw PageNotFoundException::forPageNotFound(
-            (ENVIRONMENT !== 'production' || ! $this->isWeb()) ? $e->getMessage() : ''
+            (ENVIRONMENT !== 'production' || ! $this->isWeb()) ? $e->getMessage() : null
         );
     }
 
--- a/user_guide_src/source/general/configuration/010.php
+++ b/user_guide_src/source/general/configuration/010.php
@@ -8,7 +8,6 @@ class RegionalSales
     {
         return [
             'target' => 45,
-            'actual' => 72,
         ];
     }
 }
--- a/user_guide_src/source/general/configuration/011.php
+++ b/user_guide_src/source/general/configuration/011.php
@@ -2,4 +2,3 @@
 
 $target   = 45;
 $campaign = 'Winter Wonderland';
-$actual   = 72;


sclubricants:
Also, this adds the field index type - which all the other databases have. This tells whether the index is unique, primary, foreign, etc.


And something happened when rebasing my project.. There should  only be two files changed. Not sure what happend.

sclubricants:
Any idea how all these other commits got here? Any idea how to fix?

sclubricants:
Under the current implementation this:

```php
        $this->forge->addField([
        ])->addKey(['id', 'email'], true)->createTable('user', true); 
```
will list an index for email but not for id. I know its an odd combination but interesting behavior.

kenjis:
> Any idea how all these other commits got here? Any idea how to fix?

Rebase and force push.

```console
git switch SQLiteConnectiongetIndexData
git fetch upstream
git rebase upstream/4.3
git push --force-with-lease origin HEAD
```

system/Database/SQLite3/Connection.php
```diff
@@ -265,9 +265,19 @@ protected function _fieldData(string $table): array
      */
     protected function _indexData(string $table): array
     {
-        // Get indexes
-        // Don't use PRAGMA index_list, so we can preserve index order
-        $sql = "SELECT name FROM sqlite_master WHERE type='index' AND tbl_name=" . $this->escape(strtolower($table));
+        $sql = "SELECT 'PRIMARY' as indexname, l.name as fieldname, 'PRIMARY' as indextype FROM pragma_table_info('" . trim($table, '`') . "') as l WHERE l.pk <> 0 UNION ALL
```
kenjis, line 268: Too long. Please break the line.

kenjis, line 268: Why do we need to ```trim($table, '`')```?

kenjis, line 268: When we protect table name, we should use `protectIdentifiers()` method.

tests/system/Database/Live/SQLite/AlterTableTest.php
```diff
@@ -277,4 +278,66 @@ protected function createTable(string $tableName = 'foo')
...
+        $this->forge->addField([
+            'id'         => ['type' => 'INTEGER', 'constraint' => 3, 'auto_increment' => true],
+            'userid'     => ['type' => 'INTEGER', 'constraint' => 3],
+            'name'       => ['type' => 'VARCHAR', 'constraint' => 80],
+            'email'      => ['type' => 'VARCHAR', 'constraint' => 100],
+            'country'    => ['type' => 'VARCHAR', 'constraint' => 40],
+            'created_at' => ['type' => 'DATETIME', 'null' => true],
+            'updated_at' => ['type' => 'DATETIME', 'null' => true],
+            'deleted_at' => ['type' => 'DATETIME', 'null' => true],
+        ])->addKey(['id'], true)->addUniqueKey('email')->addKey('country')->addForeignKey('userid', 'userforeign', 'id')->addForeignKey('email', 'userforeign', 'email')->createTable('testuser', true);
```
kenjis, line 307: Too long. Please break the line.

kenjis:
`ForgeTest::testCompositeKey()` should be updated.

tests/system/Database/Live/SQLite/AlterTableTest.php
```diff
@@ -277,4 +278,66 @@ protected function createTable(string $tableName = 'foo')
         $this->forge->addForeignKey('key_id', $tableName . '_fk', 'id');
         $this->forge->createTable($tableName);
     }
+
+    public function testGetIndexData()
```
kenjis, line 282: This test method has nothing to do with ALTER TABLE.
Please create a new TestCase file.

system/Database/SQLite3/Connection.php
sclubricants: We need 'table' not `table`

system/Database/SQLite3/Connection.php
```diff
@@ -130,8 +130,9 @@ protected function execute(string $sql)
                 : $this->connID->query($sql);
         } catch (ErrorException $e) {
             log_message('error', (string) $e);
+
             if ($this->DBDebug) {
-                throw $e;
+                throw new DatabaseException($e->getMessage(), $e->getCode(), $e);
```
kenjis, line 135: This change has nothing to do with this PR.
Please drop this change.

kenjis:
There are a few commits that has nothing to do with this PR like https://github.com/codeigniter4/CodeIgniter4/pull/6207/commits/be0e5812dc8e82bfc917ed85430983ab57a14ee0, https://github.com/codeigniter4/CodeIgniter4/pull/6207/commits/2c731b72888d3423d821269976925a6ecea352be.
Please drop these commits.

system/Database/SQLite3/Connection.php
kenjis: Try `$this->protectIdentifiers($table)`.

I don't know we need `strtolower($table)` or not.
Can you check?

MGatner: Aren't the back-ticks desired here? I just did a bunch of work with tables in SQLite3 and used them everywhere... maybe this function is some special exception?

MGatner: Ah or maybe this is getting re-escaped below so the back-ticks were doubling?

MGatner:
@sclubricants Any chance this will also fix https://github.com/codeigniter4/CodeIgniter4/issues/6205 by handling the indexes differently?

sclubricants:
> There are a few commits that has nothing to do with this PR like [be0e581](https://github.com/codeigniter4/CodeIgniter4/commit/be0e5812dc8e82bfc917ed85430983ab57a14ee0), [2c731b7](https://github.com/codeigniter4/CodeIgniter4/commit/2c731b72888d3423d821269976925a6ecea352be). Please drop these commits.

How do I drop commits?

sclubricants:
> @sclubricants Any chance this will also fix #6205 by handling the indexes differently?

I don't think this would affect anything with that. This affects only INTEGER AUTO_INCREMENT PRIMARY KEY. ..and also adds in the type UNIQUE, PRIMARY

sclubricants:

```php
        
        foreach ($keys as $key) {
            // if($key->name != 'PRIMARY')
        }
```

In the unlikely case such as:

```php
        $this->forge->addField([
        ])->addKey(['id', 'email'], true)->createTable('user', true); 
```

kenjis:
> How do I drop commits?

`git rebase -i`
https://opensource.com/article/22/4/manage-git-commits-rebase-i-command

sclubricants:
I tried
git fetch upstream
git checkout master
fatal: 'master' matched multiple (2) remote tracking branches

sclubricants:
$ git remote -v
origin  https://github.com/sclubricants/CodeIgniter4.git (fetch)
origin  https://github.com/sclubricants/CodeIgniter4.git (push)
upstream        https://github.com/codeigniter4/CodeIgniter4.git (fetch)
upstream        https://github.com/codeigniter4/CodeIgniter4.git (push)

kenjis:
It seems you deleted `SQLiteConnectiongetIndexData` branch in your repository.

How to reset your `develop`:
```
git fetch upstream
git checkout 4.3
git branch -D develop
git checkout --track origin/develop
```

kenjis:
It seems you deleted `SQLiteConnectiongetIndexData` branch in your repository.

How to reset your `develop`:
```
git fetch upstream
git checkout 4.3
git branch -D develop
git checkout --track upstream/develop
```

PR #6200, opened 2022-06-29
fix: add missing CLIRequest::getLocale()

kenjis:
**Description**
- add missing `CLIRequest::getLocale()`

```console
$ php spark make:controller Test

CodeIgniter v4.2.1 Command Line Tool - Server Time: 2022-06-28 21:48:32 UTC-05:00


[Error]

Call to undefined method CodeIgniter\HTTP\CLIRequest::getLocale()

at SYSTEMPATH/Config/Services.php:367

Backtrace:
  1    SYSTEMPATH/Config/BaseService.php:253
       CodeIgniter\Config\Services::language(null, false)

  2    SYSTEMPATH/Config/BaseService.php:194
       CodeIgniter\Config\BaseService::__callStatic('language', [...])

  3    SYSTEMPATH/Config/Services.php:363
       CodeIgniter\Config\BaseService::getSharedInstance('language', null)

  4    SYSTEMPATH/Config/BaseService.php:253
       CodeIgniter\Config\Services::language()

  5    SYSTEMPATH/Common.php:745
       CodeIgniter\Config\BaseService::__callStatic('language', [])

  6    SYSTEMPATH/CLI/GeneratorTrait.php:160
       lang('CLI.generator.fileCreate', [...])

  7    SYSTEMPATH/Commands/Generators/ControllerGenerator.php:88
       CodeIgniter\Commands\Generators\ControllerGenerator()->execute([...])

  8    SYSTEMPATH/CLI/Commands.php:63
       CodeIgniter\Commands\Generators\ControllerGenerator()->run([...])

  9    SYSTEMPATH/CLI/CommandRunner.php:65
       CodeIgniter\CLI\Commands()->run('make:controller', [...])

 10    SYSTEMPATH/CLI/CommandRunner.php:51
       CodeIgniter\CLI\CommandRunner()->index([...])

 11    SYSTEMPATH/CodeIgniter.php:887
       CodeIgniter\CLI\CommandRunner()->_remap('index', [...])

 12    SYSTEMPATH/CodeIgniter.php:466
       CodeIgniter\CodeIgniter()->runController(Object(CodeIgniter\CLI\CommandRunner))

 13    SYSTEMPATH/CodeIgniter.php:345
       CodeIgniter\CodeIgniter()->handleRequest(null, Object(Config\Cache), false)

 14    SYSTEMPATH/CLI/Console.php:48
       CodeIgniter\CodeIgniter()->run()

 15    ROOTPATH/spark:86
[...]

PR #6209, opened 2022-06-30
fix: lang() causes Error on CLI

kenjis:
**Description**
Supersedes #6200

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner (approved):
Much better inversion of dependencies.

MGatner:
Action failures appear to be a Coveralls issue. It also is apparent that we are submitting coverage for every database driver, which I'm not sure how Coveralls handles.

PR #6211, opened 2022-06-30
fix: multiple pagers with models do not work

### LINKED ISSUE #6204: Bug: Can't use multiple pagers on the same view
### PHP Version

8.0

### CodeIgniter4 Version

4.1.2

Linux

### Which server did you use?

apache

### Database

MariaDB 10.6.5

### What happened?

Can't use multiple pagers on the same view. I followed docs from [here](https://codeigniter.com/user_guide/libraries/pagination.html?highlight=pagination#id3).

### Steps to Reproduce

Here is my simplified Controller function:

```php
public function view(int $clientId)
{
      $contacts = $this->getContactProvider();
      $orders   = $this->getOrderProvider();
      
      return view($this->viewPrefix . 'view', [
                  'title'      => lang('Client.client'),
                  'contacts'   => $contacts->whereClient($clientId)->paginate(preference('Pager.perPage'), 'contacts'),
                  'orders'     => $orders->whereClient($clientId)->paginate(preference('Pager.perPage'), 'orders'),
                  'pager'      => $contacts->whereClient($clientId)->pager,
              ]);
}
```

### Expected Output

Get two groups in Pager instance and show pagers in view with 

```php
<?= $pager->simpleLinks('contacts') ?>
<?= $pager->links('orders') ?>
```

but there is only one group
CleanShot 2022-06-29 at 15 34 27@2x

### Anything else?

Within view I can show two paginations, but can access only that one where I'm calling `pager` method.
CleanShot 2022-06-29 at 15 41 25@2x

But when I manually specify `page_contacts` like:

```
?page_contacts=2&page_orders=1
```

It will show previous page in pagination

CleanShot 2022-06-29 at 15 44 38@2x

I'm expecting both paginations the same, because there is two pages for both of results.

kenjis:
**Description**
Fixes  #6204

I am not sure if there is something wrong with sharing Pager.
But without sharing the Pager, multiple pagers never work.

See https://codeigniter4.github.io/CodeIgniter4/libraries/pagination.html#paginating-multiple-results

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Models/PaginateModelTest.php
+++ b/tests/system/Models/PaginateModelTest.php
@@ -79,4 +79,26 @@ final class PaginateModelTest extends LiveModelTestCase
         $this->model->paginate(1, 'default', 500);
         $this->assertSame($this->model->pager->getPageCount(), $this->model->pager->getCurrentPage());
     }
+
+    public function testMultiplePager(): void
+    {
+        $validModel = $this->createModel(ValidModel::class);
+        $userModel  = $this->createModel(UserModel::class);
+
+        $validModel->paginate(1, 'valid');
+        $userModel->paginate(1, 'user');
+        $pager = $this->model->pager;
+
+        $this->assertSame(4, $validModel->countAllResults());
+        $this->assertSame(4, $userModel->countAllResults());
+
+        $this->assertStringContainsString('?page_valid=1"', $pager->links('valid'));
+        $this->assertStringContainsString('?page_valid=2"', $pager->links('valid'));
+        $this->assertStringContainsString('?page_valid=3"', $pager->links('valid'));
+        $this->assertStringContainsString('?page_valid=4"', $pager->links('valid'));
+        $this->assertStringContainsString('?page_user=1"', $pager->links('user'));
+        $this->assertStringContainsString('?page_user=2"', $pager->links('user'));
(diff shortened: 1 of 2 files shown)


iRedds:
Can you add tests for this case?

kenjis:
Yes, I will add test later.

tests/system/Models/PaginateModelTest.php
```diff
@@ -79,4 +79,26 @@ public function testPaginatePageOutOfRange(): void
         $this->model->paginate(1, 'default', 500);
         $this->assertSame($this->model->pager->getPageCount(), $this->model->pager->getCurrentPage());
     }
+
+    public function testMultiplePager(): void
+    {
+        $validModel = $this->createModel(ValidModel::class);
+        $userModel  = $this->createModel(UserModel::class);
+
```
MGatner, line 87:
```suggestion
    $this->assertSame($userModel->pager, $validModel->pager);

```

MGatner (approved):
I don't really understand how or why one would "multiple paginate" but it does seem that with grouping we would need the shared Pagination instance. Test verified this. 

One suggestion to make the comparison explicit.

MGatner (approved):

PR #6212, opened 2022-06-30
Remove unused CSS on 404 View

ThibautPV:


MGatner (approved):
I'm good with this. 404 pages are the easiest way to identify the framework and version a site is running. While this will give more information on which version is running it should make it harder to identify as CodeIgniter in the long run. 
***
I am fine counting this as a bug (dead code) and sticking with `develop` branch unless anyone objects.

PR #6214, opened 2022-07-01
Config Property Types

kenjis:
**Description**
Supersedes #6189

- add types in app/Config files
- update MockConfig
- update test Config vaules
- remove tests that are no longer needed
- update `admin/starter/app/Config/Paths.php`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Routes.php
+++ b/app/Config/Routes.php
@@ -5,12 +5,6 @@ namespace Config;
 // Create a new instance of our RouteCollection class.
 $routes = Services::routes();
 
-// Load the system's routing file first, so that the app and ENVIRONMENT
-// can override as needed.
-if (is_file(SYSTEMPATH . 'Config/Routes.php')) {
-    require SYSTEMPATH . 'Config/Routes.php';
-}
-
 /*
  * --------------------------------------------------------------------
  * Router Setup
--- a/app/Config/Routes.php
+++ b/app/Config/Routes.php
@@ -5,6 +5,12 @@ namespace Config;
 // Create a new instance of our RouteCollection class.
 $routes = Services::routes();
 
+// Load the system's routing file first, so that the app and ENVIRONMENT
+// can override as needed.
+if (is_file(SYSTEMPATH . 'Config/Routes.php')) {
+    require SYSTEMPATH . 'Config/Routes.php';
+}
+
 /*
  * --------------------------------------------------------------------
  * Router Setup
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -52,6 +52,7 @@ Changes
     - The ``CodeIgniter\CLI\CommandRunner`` class has been removed due to a change in Spark commands processing.
     - The system route configuration file ``system/Config/Routes.php`` has been removed.
     - The route configuration file ``app/Config/Routes.php`` has been changed. Removed include of system routes configuration file.
+- All union type properties in ``Config`` classes have been typed.
 
 Deprecations
 ************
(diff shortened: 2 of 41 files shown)


samsonasik:
admin appstarter seems needs update too

https://github.com/codeigniter4/CodeIgniter4/blob/fca8878b742030aad8bbb6484a78c957b707a39b/admin/starter/app/Config/Paths.php#L28

kenjis:
@samsonasik Good catch! Thanks.

MGatner (approved):
Wow thanks @kenjis this looks great! So many PHPStan ignores cleaned up, that feels good. 

Since the breaking change is entirely confined to mock classes I am fine with documenting that and proceeding. I think the likelihood of developers extending our mock classes is very slim, and it would only be in a test context. These classes probably should have been `final` to begin with. 

Approval conditional on updated documentation.

kenjis:
@MGatner The changes of the Mock Config classes break existing tests.
The property types do not match unless developers update app/Config files manually.
https://3v4l.org/tnpIv

But as you say,
> it would only be in a test context. 

So I would like to proceed to add the docs.

app/Config/Routes.php
```diff
@@ -5,6 +5,12 @@
 // Create a new instance of our RouteCollection class.
 $routes = Services::routes();
 
+// Load the system's routing file first, so that the app and ENVIRONMENT
+// can override as needed.
+if (is_file(SYSTEMPATH . 'Config/Routes.php')) {
+    require SYSTEMPATH . 'Config/Routes.php';
+}
+
```
kenjis, line 13: These lines should be removed.

kenjis: Removed.

kenjis:
Added User Guide.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -52,6 +52,7 @@ Changes
     - The ``CodeIgniter\CLI\CommandRunner`` class has been removed due to a change in Spark commands processing.
     - The system route configuration file ``system/Config/Routes.php`` has been removed.
     - The route configuration file ``app/Config/Routes.php`` has been changed. Removed include of system routes configuration file.
+- All union type properties in ``Config`` classes have been typed.
```
samsonasik, line 55: Union is ambiguous here as we use PHP 7.4, better to use nullable and/or strict scalar types.

kenjis: Oh, thanks. I was completely confused. 
I meant that we cannot use union type, so I want to say *atomic* type.

MGatner: I like "scalar" because it has precedence with PHPStan and docblocks. It's been a long time since I heard "atomic types", we used that term back in C/C++ 😊

MGatner: But I see you sent atomic already, that is fine!

kenjis: I think `array` is not `scalar`.

MGatner: You're right: int, float, string, bool. https://www.php.net/manual/en/function.is-scalar.php

But are arrays atomic? Would "simple" be more clear? Or "non-class"?

kenjis: *atomic* is not the official word in PHP.
But Psalm uses it.
https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/

I just mean it is not union nor intersection.

MGatner: Thanks for the link! That's exactly the kind of thing I was looking for. "Atomic" in C/C++ means something different, having to do with thread-safeness IIRC, but ends up being a similar set of "simpler" types. 

I like what we have here, let's stick with it.

MGatner (approved):

PR #6216, opened 2022-07-01
fix: Builder insert()/update() does not accept an object

### LINKED ISSUE #6210: Bug: Builder insert()/update() does not accept an object
### PHP Version

7.4

### CodeIgniter4 Version

4.0.4 

Manual (zip or tar.gz)

Linux

### Which server did you use?

apache

### Database

MySQL 5.56

### What happened?

Hi, I have a problem with DB Builder 
Log:
`CRITICAL - 2022-06-30 09:24:34 --> Argument 1 passed to CodeIgniter\Database\BaseBuilder::update() must be of the type array or null, object given, called in ....php on line 41 `

My mine 40 & line 41: 
```php
$builder = $this->dbmail->table(self::TABLE); 
$builder->update($undelivered, ['undelivered_id' => $undelivered->undelivered_id]); 
```

According to documentation: 
```php
$object = new Myclass(); 
$builder->where('id', $id); 
$builder->update($object); 
```
https://codeigniter4.github.io/CodeIgniter4/database/query_builder.html#builder-update

Builder insert() have a same problem.

### Steps to Reproduce

Just do Builder update with object.

### Expected Output

No critical or any other error

### Anything else?

 Looks like the error is in all versions from 4.0.4 until current

kenjis:
**Description**
Fixes #6210

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2014,11 +2014,12 @@ class BaseBuilder
     /**
      * Compiles an update string and runs the query.
      *
-     * @param mixed $where
+     * @param array|object|null        $set
+     * @param array|RawSql|string|null $where
      *
      * @throws DatabaseException
      */
-    public function update(?array $set = null, $where = null, ?int $limit = null): bool
+    public function update($set = null, $where = null, ?int $limit = null): bool
     {
         if ($set !== null) {
             $this->set($set);
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1878,11 +1878,13 @@ class BaseBuilder
     /**
      * Compiles an insert string and runs the query
      *
+     * @param array|object|null $set
+     *
      * @throws DatabaseException
      *
      * @return bool|Query
      */
-    public function insert(?array $set = null, ?bool $escape = null)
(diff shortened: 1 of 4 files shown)


MGatner (approved):

kenjis:
Added changelog.

MGatner (commented):
We've gotten very lenient with breaking changes. This is out of sync with the User Guide but is actual risk of breaking any method extensions. I would like more of the team to sign off on this. 

My opinion: extensions of the `BaseBuilder` should be implementing `_insert()` so `insert()` is unlikely to be extended (and could probably be `final` at some point). Developers extending this should be doing rather advanced work and able to cope with this change. Given these points I am in favor.

michalsn (approved):
This is more like a bug fix that will accidentally introduce a breaking change for some rare cases, but I have no problem with that. It's not like we woke up one day and said

> Hey, this method should also support objects.

The behavior was documented before and was there even in v3, so from my perspective everything is fine.

samsonasik (approved):

MGatner:
Thanks for the feedback all! @kenjis I think we're good unless you have more thoughts.

PR #6219, opened 2022-07-01
Redesign SQLite Connection::_getIndexData() to include auto_increment PK

sclubricants:
Supersedes #6207

This is a rewrite of system/Database/SQLite3/Connection::_indexData().

SQLite doesn't show indexes on tables created with INTEGER PRIMARY KEY AUTO_INCREMENT. This PR looks for primary keys that don't appear in the normal index lookup and adds it to the return of the method.

```php
        $this->forge->addField([
```

```php
        $this->forge->addField([
```

The need is to get a comprehensive list of database constraints. This will be useful in adding upsert() to the database.

```php
        
        foreach ($keys as $key) {
            // if($key->name != 'PRIMARY')
        }
```

In the unlikely case such as:

```php
        $this->forge->addField([
        ])->addKey(['id', 'email'], true)->createTable('user', true); 
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


kenjis:
This is an enhancement.
The base branch should be not `develop` but `4.3`.
Please change.

sclubricants:
just forget it

PR #6221, opened 2022-07-01
SQLite3 Connection getIndexData()

sclubricants:
Supersedes https://github.com/codeigniter4/CodeIgniter4/pull/6207

This is a rewrite of `system/Database/SQLite3/Connection::_indexData()`.

SQLite doesn't show indexes on tables created with `INTEGER PRIMARY KEY AUTO_INCREMENT`. This PR looks for primary keys that don't appear in the normal index lookup and adds it to the return of the method. It also adds the key type: Primary, Unique, or Index.

```php
        $this->forge->addField([
```

```php
        $this->forge->addField([
```

The need is to get a comprehensive list of database constraints. This will be useful in adding `upsert()` to the database.

```php
        
        foreach ($keys as $key) {
            // if ($key->name !== 'PRIMARY')
        }
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -37,6 +37,7 @@ Enhancements
 - Added ``Model::allowEmptyInserts()`` method to insert empty data. See :ref:`Using CodeIgniter's Model <model-allow-empty-inserts>`
 - Added ``$routes->useSupportedLocalesOnly(true)`` so that the Router returns 404 Not Found if the locale in the URL is not supported in ``Config\App::$supportedLocales``. See :ref:`Localization <localization-in-routes>`
 - The call handler for Spark commands from the ``CodeIgniter\CodeIgniter`` class has been extracted. This will reduce the cost of console calls.
+- Added a psuedo index named 'PRIMARY' for SQLite `AUTOINCREMENT` column. SQLite treats these in a special way which was excluding them from being retrieved by calling ``BaseConnection::getIndexData()``.
 
 Changes
 *******
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -37,7 +37,7 @@ Enhancements
 - Added ``Model::allowEmptyInserts()`` method to insert empty data. See :ref:`Using CodeIgniter's Model <model-allow-empty-inserts>`
 - Added ``$routes->useSupportedLocalesOnly(true)`` so that the Router returns 404 Not Found if the locale in the URL is not supported in ``Config\App::$supportedLocales``. See :ref:`Localization <localization-in-routes>`
 - The call handler for Spark commands from the ``CodeIgniter\CodeIgniter`` class has been extracted. This will reduce the cost of console calls.
-- Added a psuedo index named 'PRIMARY' for SQLite `AUTOINCREMENT` column. SQLite treats these in a special way which was excluding them from being retrieved by calling ``BaseConnection::getIndexData()``.
+- SQLite ``BaseConnection::getIndexData()`` now can return pseudo index named ``PRIMARY`` for `AUTOINCREMENT` column, and each returned index data has ``type`` property.
 
 Changes
 *******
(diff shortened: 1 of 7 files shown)


sclubricants:
Got some work to do in CodeIgniter4\system\Database\SQLite3\Table

Line 111

```php
 $this->keys = array_merge($this->keys, $this->formatKeys($this->db->getIndexData($table)));
```

sclubricants:
Adding 

```sql
SELECT `table` || '_' || `from` || '_foreign' as indexname, `from` as fieldname, 'FOREIGN' as indextype FROM pragma_foreign_key_list('" . trim($table, '`') . "') UNION ALL
```
to _indexData query will retrieve foreign keys as well. This may be a step to far though it seems all the other databases return these with indexes.  There is a separate method that will retrieve these.

MGatner (approved):
The PR looks good, very nice tests. I really am not qualified to decide if this is a better approach to SQLite, but it seems like you are mostly expanding on what we already have so I'll give a tentative "approve". Need somebody else to chime in as well.

kenjis:
@sclubricants Thank you for the PR.
This is an enhancement, so please add something in the changelog.
https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/changelogs/v4.3.0.rst#enhancements

kenjis:
It may be kind to add the note in https://codeigniter4.github.io/CodeIgniter4/database/metadata.html#db-getindexdata

> Its important to note that something like the following would likely cause an error:

kenjis:
> In the unlikely case such as:
...

What do you mean? I got the array when specifying `addKey(['id', 'email'], true)`:

```
array (1) [
    0 => stdClass (3) (
        public 'indexname' -> string (7) "PRIMARY"
        public 'fieldname' -> string (2) "id"
        public 'indextype' -> string (7) "PRIMARY"
    )
]
```

sclubricants:
@kenjis User Guide and Change Log updated.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -37,6 +37,7 @@ Enhancements
 - Added ``Model::allowEmptyInserts()`` method to insert empty data. See :ref:`Using CodeIgniter's Model <model-allow-empty-inserts>`
 - Added ``$routes->useSupportedLocalesOnly(true)`` so that the Router returns 404 Not Found if the locale in the URL is not supported in ``Config\App::$supportedLocales``. See :ref:`Localization <localization-in-routes>`
 - The call handler for Spark commands from the ``CodeIgniter\CodeIgniter`` class has been extracted. This will reduce the cost of console calls.
+- Added a psuedo index named 'PRIMARY' for SQLite `AUTOINCREMENT` column. SQLite treats these in a special way which was excluding them from being retrieved by calling ``BaseConnection::getIndexData()``.
```
kenjis, line 40: How about this?

```suggestion
- SQLite ``BaseConnection::getIndexData()`` now can return pseudo index named ``PRIMARY`` for `AUTOINCREMENT` column, and each returned index data has ``type`` property.
```

kenjis:
Before:
```
array (2) [
    'testuser_email' => stdClass (2) (
        public 'name' -> string (14) "testuser_email"
        public 'fields' -> array (1) [
            0 => string (5) "email"
        ]
    )
    'testuser_country' => stdClass (2) (
        public 'name' -> string (16) "testuser_country"
        public 'fields' -> array (1) [
            0 => string (7) "country"
        ]
    )
]
```

After;
```
array (3) [
    'PRIMARY' => stdClass (3) (
        public 'name' -> string (7) "PRIMARY"
        public 'fields' -> array (1) [
            0 => string (2) "id"
        ]
        public 'type' -> string (7) "PRIMARY"
    )
    'testuser_email' => stdClass (3) (
        public 'name' -> string (14) "testuser_email"
        public 'fields' -> array (1) [
            0 => string (5) "email"
        ]
        public 'type' -> string (6) "UNIQUE"
    )
    'testuser_country' => stdClass (3) (
        public 'name' -> string (16) "testuser_country"
        public 'fields' -> array (1) [
            0 => string (7) "country"
        ]
        public 'type' -> string (5) "INDEX"
    )
]
```

kenjis (approved):

PR #6225, opened 2022-07-03
[Rector] Apply Rector run on utils/ directory

samsonasik:
Apply Rector run on `utils/` directory, not only `/utils/Rector` so other files (PHPStan rules, check_tabs_in_rst.php), will be checked as well.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/utils/PHPStan/CheckFrameworkExceptionInstantiationViaNamedConstructorRule.php
+++ b/utils/PHPStan/CheckFrameworkExceptionInstantiationViaNamedConstructorRule.php
@@ -22,9 +22,6 @@ use PHPStan\Rules\Rule;
 
 final class CheckFrameworkExceptionInstantiationViaNamedConstructorRule implements Rule
 {
-    /**
-     * @var string
-     */
     private const ERROR_MESSAGE = 'FrameworkException instance creation via new expression is not allowed, use its named constructor instead';
 
     public function getNodeType(): string
--- a/rector.php
+++ b/rector.php
@@ -64,7 +64,7 @@ return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->parallel();
 
     // paths to refactor; solid alternative to CLI arguments
-    $rectorConfig->paths([__DIR__ . '/app', __DIR__ . '/system', __DIR__ . '/tests', __DIR__ . '/utils/Rector']);
+    $rectorConfig->paths([__DIR__ . '/app', __DIR__ . '/system', __DIR__ . '/tests', __DIR__ . '/utils']);
 
     // do you need to include constants, class aliases or custom autoloader? files listed will be executed
     $rectorConfig->bootstrapFiles([
--- a/utils/PHPStan/CheckUseStatementsAfterLicenseRule.php
+++ b/utils/PHPStan/CheckUseStatementsAfterLicenseRule.php
@@ -22,14 +22,7 @@ use PHPStan\Rules\Rule;
 
 final class CheckUseStatementsAfterLicenseRule implements Rule
 {
-    /**
-     * @var string
-     */
-    private const ERROR_MESSAGE = 'Use statement must be located after license docblock';
-
-    /**
-     * @var string
-     */
+    private const ERROR_MESSAGE   = 'Use statement must be located after license docblock';
     private const COPYRIGHT_REGEX = '/\* \(c\) CodeIgniter Foundation <admin@codeigniter\.com>/m';
 
     public function getNodeType(): string
--- a/utils/check_tabs_in_rst.php
+++ b/utils/check_tabs_in_rst.php
@@ -11,6 +11,8 @@ declare(strict_types=1);
  * the LICENSE file that was distributed with this source code.
  */
 
+namespace Utils;
+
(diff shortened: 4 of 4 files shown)


paulbalandan (approved):

utils/PHPStan/CheckFrameworkExceptionInstantiationViaNamedConstructorRule.php
```diff
@@ -22,9 +22,6 @@
 
 final class CheckFrameworkExceptionInstantiationViaNamedConstructorRule implements Rule
 {
-    /**
-     * @var string
-     */
```
kenjis, line 27: @samsonasik Why does Rector remove this PHPDoc?

samsonasik: It is @MGatner request to create custom rile to remove var from constant https://github.com/codeigniter4/CodeIgniter4/blob/develop/utils/Rector/RemoveVarTagFromClassConstantRector.php

MGatner: A constant is a constant 🤷‍♂️

I do remember that conversation but I'm also fine if we want to revisit this. I think it is unnecessary to type constants but people do it.

samsonasik: ref https://github.com/codeigniter4/CodeIgniter4/pull/4766

kenjis: It is a common practice to put `@var` for constants.
I have never seen removing `@var` for constants.

But at least, PhpStorm does not need `@var` for constants.
So it does not seem to be inconvenient if it is removed.

samsonasik: In Rector, there is `VarTagRemover` service that can remove `@var` if useless, eg: only '@var string` without description. It will disallow on complex `@var array<string, int[]>` for example:

https://github.com/rectorphp/rector-src/blob/a70d0e5518260479d055a0d5a90e28e43cad58f7/rules/DeadCode/PhpDoc/TagRemover/VarTagRemover.php#L29

we can extends that if needed,eg:

```php
   public function removeVarTagIfUseless(PhpDocInfo $phpDocInfo, Property|ClassConst $node): void
    {
        $varTagValueNode = $phpDocInfo->getVarTagValueNode();
        if (! $varTagValueNode instanceof VarTagValueNode) {
            return;
        }

        $isVarTagValueDead = $this->deadVarTagValueNodeAnalyzer->isDead($varTagValueNode, $node);
        if (! $isVarTagValueDead) {
            return;
        }

        if ($this->phpDocTypeChanger->isAllowed($varTagValueNode->type)) {
            return;
        }

        $phpDocInfo->removeByType(VarTagValueNode::class);
    }
```

MGatner: Not to sound like a broken record, but the reason we removed these was because `@var` tags are *always* useless on a constant because the value is the type: it never changes. I am trying to think what value they possibly add, and the only thing I can think of is if you have a class constant that is expected to be provided by a child class:
```php
abstract class ExtendMe
{
    /**
     * Be sure to set the names
     * on your extended classes.
     *
     * @var string[]
     */
    public const NAMES = [];
}
```

kenjis: A constant is no longer a constant.

```php
<?php

class A {}

const CONSTANT = new A;

class B {
    public const ClassConstant1 = CONSTANT;
}
```

https://3v4l.org/K9ZCKU

MGatner: > A constant is no longer a constant.

The underlying values may be mutable but it is still a fixed type at runtime. If for some awful reason somebody used your example I would expect a docblock explanation, not a type. Providing `@var` typing for static analysis is unnecessary because constants are always strongly-typed to whatever they are.

kenjis: Yeah, the type is fixed.
But the value is dynamic determined. Very interesting. 
https://3v4l.org/dWAaW
I don't know how and when to use it, though.

MGatner: Object constants are strange to me. I can't even find the RFC that allowed them for a good example of _why_ someone would want one 🤷‍♂️ To me the power of a constant is it's fixed, simple value.

MGatner: https://php.watch/rfcs/new_in_initializers

MGatner: Typed class constants is under discussion for 8.2: https://wiki.php.net/rfc/typed_class_constants

PR #6233, opened 2022-07-05
docs: improve route_to() and url_to()

### LINKED ISSUE #6223: Docs: route_to() logic
The User Guide has examples of using `route_to()` to return a URI path, which will fail in certain circumstances (e.g. subfolders). Docs should be updated to be clear that this returns a *route*, and if a URI is needed then it should be passed through `site_url()` (or just use `url_to()`). 

See also: https://github.com/codeigniter4/shield/issues/267#issuecomment-1172875796

kenjis:
**Description**
Fixes #6223

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/incoming/routing/029.php
+++ b/user_guide_src/source/incoming/routing/029.php
@@ -5,6 +5,6 @@ $routes->get('users/(:num)/gallery(:any)', 'Galleries::showUserGallery/$1/$2');
 
 ?>
 
-<!-- Generate the relative URL to link to user ID 15, gallery 12: -->
+<!-- Generate the URI path (route) to link to user ID 15, gallery 12: -->
 <a href="<?= route_to('Galleries::showUserGallery', 15, 12) ?>">View Gallery</a>
 <!-- Result: '/users/15/gallery/12' -->
--- a/user_guide_src/source/incoming/routing/030.php
+++ b/user_guide_src/source/incoming/routing/030.php
@@ -5,6 +5,6 @@ $routes->get('users/(:num)/gallery(:any)', 'Galleries::showUserGallery/$1/$2', [
 
 ?>
 
-<!-- Generate the relative URL to link to user ID 15, gallery 12: -->
+<!-- Generate the URI path (route) to link to user ID 15, gallery 12: -->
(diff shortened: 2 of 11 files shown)


user_guide_src/source/incoming/routing/029.php
```diff
@@ -5,6 +5,6 @@
 
 ?>
 
-<!-- Generate the relative URL to link to user ID 15, gallery 12: -->
+<!-- Generate the URI path (route) to link to user ID 15, gallery 12: -->
 <a href="<?= route_to('Galleries::showUserGallery', 15, 12) ?>">View Gallery</a>
```
MGatner, line 9: We should never use a route I place of a URL, bad practice.

user_guide_src/source/incoming/routing/030.php
```diff
@@ -5,6 +5,6 @@
 
 ?>
 
-<!-- Generate the relative URL to link to user ID 15, gallery 12: -->
+<!-- Generate the URI path (route) to link to user ID 15, gallery 12: -->
 <a href="<?= route_to('user_gallery', 15, 12) ?>">View Gallery</a>
```
MGatner, line 9: Same.

MGatner (changes requested):
I know the examples are trying to demonstrate how to use `route_to()` but I would prefer we remain strict about routes vs. URIs. I can't think of a scenario where `route_to()` would ever be output to HTML - it is an internal tool.

kenjis:
> I can't think of a scenario where route_to() would ever be output to HTML - it is an internal tool.

You are correct!
We have less use cases for `route_to()`.

user_guide_src/source/incoming/routing/029.php
kenjis: Replaced route_to() with url_to().

user_guide_src/source/incoming/routing/030.php
kenjis: Replaced route_to() with url_to().

MGatner (approved):
The URL Structure section is a very nice addition.

PR #6238, opened 2022-07-07
docs: add note about the status code for set404Override()

### LINKED ISSUE #6237: Docs : set404Override() return status code 200
### PHP Version

7.4

### CodeIgniter4 Version

4.2.1

Manual (zip or tar.gz)

macOS

### Which server did you use?

apache

### Database

_No response_

### What happened?

If we use the method set404Override() as indicated in the documentation ([see here](https://codeigniter.com/user_guide/incoming/routing.html#override)) , it is not specified that the server code returned will be a 200

### Steps to Reproduce

Use the code present in the documentation : 

```
$routes->set404Override('App\Errors::show404');
```

OR

```
$routes->set404Override(static function () {
    echo view('my_errors/not_found.html');
});
```

### Expected Output

It would be nice to specify that you have to use the method "$response->setStatusCode(404);" ([see here](https://codeigniter.com/user_guide/outgoing/response.html#setStatusCode)) to return the correct error code

An example would certainly be nice for users

### Anything else?

_No response_

kenjis:
**Description**
Fixes #6237

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/incoming/routing.rst
+++ b/user_guide_src/source/incoming/routing.rst
@@ -519,6 +519,11 @@ a valid class/method pair, just like you would show in any route, or a Closure:
 
 .. literalinclude:: routing/051.php
 
+.. note:: The ``set404Override()`` method does not change the Response status code to ``404``.
+    If you don't set the status code in the controller you set, the default status code ``200``
+    will be returned. See :php:func:`Response::setStatusCode() <setStatusCode>` for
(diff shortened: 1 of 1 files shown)


MGatner (approved):

samsonasik:
Thank you @kenjis

PR #6243, opened 2022-07-08
Fix: Validation of fields with a leading asterisk.

iRedds:
**Description**
https://forum.codeigniter.com/showthread.php?tid=82391&pid=398147#pid398147

Fixed: Validation of fields with a leading asterisk (wildcard).
Also fixed getting an error by field name with a leading asterisk.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Validation/ValidationTest.php
+++ b/tests/system/Validation/ValidationTest.php
@@ -1098,6 +1098,12 @@ final class ValidationTest extends CIUnitTestCase
                 ['foo' => ['boz']],
             ]],
         ];
+
+        yield 'leading-asterisk' => [
+            true,
+            ['*.foo' => 'required'],
+            [['foo' => 'bar']],
+        ];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ddevsr (approved):

MGatner (approved):
I'm not great with regex patterns but your test case looks good and nothing else failed so it's good by me. Thank you!

MGatner:
Original reporter verifies this works 💪

tests/system/Validation/ValidationTest.php
```diff
@@ -1324,4 +1330,17 @@ public function testNestedArrayThrowsException(): void
...
     }
+
+    public function testRuleWithLeadingAsterisk(): void
+    {
+        $data = [
+            ['foo' => 1],
+            ['foo' => null],
+        ];
+
+        $this->validation->setRules(['*.foo' => 'required'], ['1.foo' => ['required' => 'Required {field}']]);
```
kenjis, line 1341: This is not related with this bug, but `['1.foo' => ['required' => 'Required {field}']]` seems tricky .

kenjis (approved):

kenjis:
Thank you! @iRedds

tests/system/Validation/ValidationTest.php
iRedds: You know, it turned out that the option you suggested does not work. The default error text is returned.

iRedds: The error text will be defined for the ``*.foo`` field, and the ``1.foo`` field is passed.

PR #6244, opened 2022-07-08
Add indexes to test user table

sclubricants:
Added a unique index 'email' and a non unique index 'country' to the user table that is created for tests.

This will allow for use with tests involving indexes and constraints. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Database/Live/UpdateTest.php
+++ b/tests/system/Database/Live/UpdateTest.php
@@ -51,8 +51,15 @@ final class UpdateTest extends CIUnitTestCase
                 ->get()
                 ->getResult();
 
-            $this->assertSame('Bobby', $result[0]->name);
-            $this->assertSame('Ahmadinejad', $result[1]->name);
+            // this is really a bad test - indexes and other things can affect sort order
+            if ($this->db->DBDriver === 'SQLSRV') {
+                $this->assertSame('Derek Jones', $result[0]->name);
+                $this->assertSame('Bobby', $result[1]->name);
+            } else {
+                $this->assertSame('Bobby', $result[0]->name);
+                $this->assertSame('Ahmadinejad', $result[1]->name);
+            }
+
             $this->assertSame('Richard A Causey', $result[2]->name);
             $this->assertSame('Chris Martin', $result[3]->name);
         } catch (DatabaseException $e) {
--- a/tests/_support/Database/Migrations/20160428212500_Create_test_tables.php
+++ b/tests/_support/Database/Migrations/20160428212500_Create_test_tables.php
@@ -26,7 +26,7 @@ class Migration_Create_test_tables extends Migration
             'created_at' => ['type' => 'DATETIME', 'null' => true],
             'updated_at' => ['type' => 'DATETIME', 'null' => true],
             'deleted_at' => ['type' => 'DATETIME', 'null' => true],
-        ])->addKey('id', true)->createTable('user', true);
+        ])->addKey('id', true)->addUniqueKey('email')->addKey('country')->createTable('user', true);
 
         // Job Table
         $this->forge->addField([
--- a/tests/system/Models/InsertModelTest.php
+++ b/tests/system/Models/InsertModelTest.php
@@ -185,16 +185,24 @@ final class InsertModelTest extends LiveModelTestCase
             ];
         };
 
+        $entityTwo = clone $entity;
+
         $this->createModel(UserModel::class);
 
-        $entity->name       = 'Mark';
-        $entity->email      = 'mark@example.com';
+        $entity->name       = 'Mark One';
+        $entity->email      = 'markone@example.com';
         $entity->country    = 'India';
         $entity->deleted    = 0;
         $entity->created_at = new Time('now');
 
+        $entityTwo->name       = 'Mark Two';
+        $entityTwo->email      = 'marktwo@example.com';
+        $entityTwo->country    = 'India';
+        $entityTwo->deleted    = 0;
+        $entityTwo->created_at = $entity->created_at;
+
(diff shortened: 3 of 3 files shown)


sclubricants:
Adding the indexes exposed some issues where create table is called but the table already exists.

The create table call uses the CREATE TABLE IF NOT EXISTS method so it doesn't fail if the table already exists.

However, because indexes are often added after the table creation using ALTER TABLE the index creation fails where the index already exists. So CREATE INDEX IF NOT EXISTS was added.

MGatner:
Creating tables has "if not exists" as an optional parameter. If we are going to check for indexes on creation I think it should be the same. I am generally not in favor of code that covers over database mistakes - for example if my code is trying to create a table that already exists then I want an exception thrown because I probably did something else wrong. 

Databases aren't my strongest area, but I imagine indexes are the same way: if you're trying to create an index that already exists then something else is probably wrong, and we should assume the developer is okay with hiding that.

sclubricants:
> Creating tables has "if not exists" as an optional parameter. 

@MGatner 
I didn't realize this. I thought it was kind of odd to just blindly use such a statement. When I added the indexes to the test table it caused all the other tests to fail. I was just trying a quick fix at a serious problem I guess.

So the problem is then, when a table create command comes with the extra parameter then follow likewise with the alter table statements.

Perhaps it might be better to query the database to find out if the table exists and if it does then do not execute the create table statement or the following create index statement.

As it stands, for the databases that require a second statement to create indexes they will fail if they already exists.

Personally, I'd rather have the option to drop the table if exists and then recreate. This way you know the schema is going to be what you are expecting. Also this resets the indexes and auto increments. However you might lose data if there was some already in the table.

sclubricants:
@MGatner 
```php
    protected function _createTable(string $table, bool $ifNotExists, array $attributes)
    {
        // For any platforms that don't support Create If Not Exists...
        if ($ifNotExists === true && $this->createTableIfStr === false) {
            if ($this->db->tableExists($table)) {
                return true;
            }

            $ifNotExists = false;
        }
```

Why not use this check for all platforms?

```php
    protected function _createTable(string $table, bool $ifNotExists, array $attributes)
    {
        if ($ifNotExists === true && $this->db->tableExists($table)) {
            return true;

        }

        $ifNotExists = false;
```

MGatner:
@sclubricants I'm very grateful for your work here. I'm afraid I'm a poor conversation partner when it comes to the database layer. I try to stay away from this code when I can because I am fairly ignorant of the underlying database details. Let me see if I can get someone else from the team in who might be better versed in these details.

sclubricants:
@MGatner  This is a bug. Should I create an issue for this and submit a separate PR under develop?

kenjis:
@sclubricants If this PR is still work in progress, please rebase and resolve the conflict.

sclubricants:
@kenjis Updated!

tests/system/Database/Live/UpdateTest.php
```diff
@@ -51,8 +51,15 @@ public function testUpdateSetsAllWithoutWhereAndLimit()
...
-            $this->assertSame('Ahmadinejad', $result[1]->name);
+            // this is really a bad test - indexes and other things can affect sort order
+            if ($this->db->DBDriver === 'SQLSRV') {
+                $this->assertSame('Derek Jones', $result[0]->name);
+                $this->assertSame('Bobby', $result[1]->name);
+            } else {
+                $this->assertSame('Bobby', $result[0]->name);
+                $this->assertSame('Ahmadinejad', $result[1]->name);
+            }
+
```
MGatner, line 62: If order is not consistent but we aren't testing for order then this should be refactored to something like:
```suggestion
            $this->assertEqualsCanonacalizing(['Bobby', 'Derek Jones'], array_column($result, 'name');
```
I'm on mobile so can't finish that but hopefully you get the idea.

MGatner (changes requested):

tests/system/Database/Live/UpdateTest.php
sclubricants: @MGatner 
A canonical test will fail here as well. We start with four names in the database.. none of which are Bobby. We the run a query:
UPDATE user SET name = 'Bobby' LIMIT 1
Prior to adding the indexes on the user table the database would select rows for update in a somewhat relyable fashion. It would order based on the Primary key. But after adding indexes at least SQLSRV updates on a different sort order. Depending on which name gets overwritten with Bobby you end up with differet combinations of names.

About the only relyable assertions that can be made here is that there should be one and only one user with the name Bobby. 

I could change to use jobs table which only has primary key, leave as is ..as i have worked out the order in which they get updated, or simplify the assertion that only one Bobby does exist. 

This sort of query is rather odd and would rarely be used. Depending on how indexing is setup it might be more useful in some circumstances.

sclubricants: https://www.brentozar.com/archive/2019/05/why-order-isnt-guaranteed-without-an-order-by/

kenjis: It seems there is no way to improve the test.
The test is unrealistic, but it tests the case of `testUpdateSetsAllWithoutWhereAndLimit`.

kenjis (approved):

MGatner:
Thanks @sclubricants! Good to have some polish happening on the database layer.

tests/system/Database/Live/UpdateTest.php
MGatner: Got it! I understand now, thanks for the explanation.

MGatner (approved):

PR #6247, opened 2022-07-10
fix: do not use ValidationInterface

### LINKED ISSUE #6224: Bug: The Validation::run() method does not match the ValidationInterface.
### CodeIgniter4 Version

4.2.1

### What happened?

Missing $dbGroup argument.

https://github.com/codeigniter4/CodeIgniter4/blob/7439b6ec5134492bbbf466d99bc206aa883a4b54/system/Validation/Validation.php#L109

https://github.com/codeigniter4/CodeIgniter4/blob/7439b6ec5134492bbbf466d99bc206aa883a4b54/system/Validation/ValidationInterface.php#L28

kenjis:
**Description**
Fixes #6224

`BaseModel` uses the 3rd param of `$this->validation->run($data, null, $this->DBGroup)`.
So it depends on `Validation`, not `ValidationInterface`.

Ref: https://github.com/codeigniter4/CodeIgniter4/pull/5909#issuecomment-1179518692

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -20,7 +20,6 @@ use CodeIgniter\Exceptions\ModelException;
 use CodeIgniter\I18n\Time;
 use CodeIgniter\Pager\Pager;
 use CodeIgniter\Validation\Validation;
-use CodeIgniter\Validation\ValidationInterface;
 use Config\Services;
 use InvalidArgumentException;
 use ReflectionClass;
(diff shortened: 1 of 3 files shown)


MGatner:
Let's see where the big PR takes this conversation. I'm fine with this one following from that - I do think that this is pretty likely to break some projects and libraries since Model constructors are likely to be extended.

kenjis:
If we can change the ValidationInterface, we have an option to change it instead.

MGatner:
You know that's not a bad idea. If I had to guess I would say there are probably more developers extending `Model::__construct()` than making their own Validation driver (since the driver can be "extended" via rules already).

kenjis:
> there are probably more developers extending Model::__construct() than making their own Validation driver

I agree. I guess so.
It seems changing ValidationInterface makes less apps break.

kenjis:
I sent another PR #6253

kenjis:
#6253 was merged.

Issue #6248, opened 2022-07-10
Bug: `$forge->createTable('table', true)` fails when table has indexes


sclubricants:
### PHP Version

8.0

### CodeIgniter4 Version

4.2.1

### CodeIgniter4 Installation Method

Git

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

apache

### Database

SQLite and others

### What happened?

When creating a table with the optional parameter IF NOT EXISTS the execution fails if the table already exists and there are indexes to be created as well. If fails for the databases that create the indexes after the table creation command.

### Steps to Reproduce

Run the following on SQLite, Postgre, SQLSRV and maybe others:

```php
$forge = \Config\Database::forge('tests');  

$forge->addField([
    'id'         => ['type' => 'INTEGER', 'constraint' => 3, 'auto_increment' => true],
    'name'       => ['type' => 'VARCHAR', 'constraint' => 80],
    'email'      => ['type' => 'VARCHAR', 'constraint' => 100],
    'country'    => ['type' => 'VARCHAR', 'constraint' => 40],
    'created_at' => ['type' => 'DATETIME', 'null' => true],
    'updated_at' => ['type' => 'DATETIME', 'null' => true],
    'deleted_at' => ['type' => 'DATETIME', 'null' => true],
])->addKey('id', true)->addUniqueKey('email')->addKey('country')->createTable('user', true);

$forge->addField([
    'id'         => ['type' => 'INTEGER', 'constraint' => 3, 'auto_increment' => true],
    'name'       => ['type' => 'VARCHAR', 'constraint' => 80],
    'email'      => ['type' => 'VARCHAR', 'constraint' => 100],
    'country'    => ['type' => 'VARCHAR', 'constraint' => 40],
    'created_at' => ['type' => 'DATETIME', 'null' => true],
    'updated_at' => ['type' => 'DATETIME', 'null' => true],
    'deleted_at' => ['type' => 'DATETIME', 'null' => true],
])->addKey('id', true)->addUniqueKey('email')->addKey('country')->createTable('user', true);
```

An error is thrown: SQLite3::exec(): index db_user_email already exists

### Expected Output

Expected that table will be created if it doesn't exist and if it does exists nothing will change and no errors will be thrown.

kenjis:
Indeed, the error occurs because `CREATE UNIQUE INDEX` is executed after `CREATE TABLE`.

```console
[ErrorException]

[...]

PR #6249, opened 2022-07-10
Fix create table if not exists when indexes already exist

sclubricants:
This fixes https://github.com/codeigniter4/CodeIgniter4/issues/6248

This removes all code for CREATE TABLE table `IF NOT EXISTS`. Because all platforms do not support this statement the current implementation queries  the database first to determine if the table exists. If it does then it doesn't execute the create table statement. The problem occurs when the create indexes statements are executed after the create table statement and the indexes already exists. Rather then try and apply CREATE INDEX index IF NOT EXISTS and carry through the logic, this PR removes completely the IF NOT EXISTS regime and instead relies on the query table lookup. If the table exists then there is need to execute anything. This way the code is simpler and all the problems of index creation is averted. 

All the magic happens in (base) Forge here:

```php
    public function createTable(string $table, bool $ifNotExists = false, array $attributes = [])
    {
        if ($table === '') {
            throw new InvalidArgumentException('A table name is required for that operation.');
        }

        $table = $this->db->DBPrefix . $table;

        if ($this->fields === []) {
            throw new RuntimeException('Field information is required.');
        }

        if ($ifNotExists === true && $this->db->tableExists($table)) {
            $this->reset();

            return true;
        }

        $sql = $this->_createTable($table, $attributes);
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -540,8 +540,6 @@ class Forge
      */
     protected function _createTable(string $table, bool $ifNotExists, array $attributes)
     {
-        $sql = 'CREATE TABLE';
-
         $columns = $this->_processFields(true);
 
         for ($i = 0, $c = count($columns); $i < $c; $i++) {
@@ -563,7 +561,7 @@ class Forge
 
         return sprintf(
             $this->createTableStr . '%s',
-            $sql,
+            'CREATE TABLE',
             $this->db->escapeIdentifiers($table),
             $columns,
             $this->_createTableAttributes($attributes)
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -107,9 +107,7 @@ class Forge
     protected $createTableStr = "%s %s (%s\n)";
 
     /**
-     * CREATE TABLE IF statement
-     *
-     * @var bool|string
+     * @deprecated This is no longer used.
      */
     protected $createTableIfStr = 'CREATE TABLE IF NOT EXISTS';
 
@@ -495,7 +493,14 @@ class Forge
             throw new RuntimeException('Field information is required.');
         }
 
-        $sql = $this->_createTable($table, $ifNotExists, $attributes);
+        // If table exists lets stop here
+        if ($ifNotExists === true && $this->db->tableExists($table)) {
+            $this->reset();
+
+            return true;
+        }
+
+        $sql = $this->_createTable($table, false, $attributes);
 
         if (is_bool($sql)) {
             $this->reset();
@@ -530,20 +535,12 @@ class Forge
 
     /**
      * @return bool|string
+     *
+     * @deprecated $ifNotExists is no longer used, and will be removed.
      */
     protected function _createTable(string $table, bool $ifNotExists, array $attributes)
     {
-        // For any platforms that don't support Create If Not Exists...
-        if ($ifNotExists === true && $this->createTableIfStr === false) {
-            if ($this->db->tableExists($table)) {
-                return true;
-            }
-
-            $ifNotExists = false;
-        }
-
-        $sql = ($ifNotExists) ? sprintf($this->createTableIfStr, $this->db->escapeIdentifiers($table))
-            : 'CREATE TABLE';
+        $sql = 'CREATE TABLE';
 
         $columns = $this->_processFields(true);
 
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -107,6 +107,10 @@ class Forge
     protected $createTableStr = "%s %s (%s\n)";
 
     /**
+     * CREATE TABLE IF statement
+     *
+     * @var bool|string
+     *
      * @deprecated This is no longer used.
      */
     protected $createTableIfStr = 'CREATE TABLE IF NOT EXISTS';
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -506,19 +506,6 @@ class Forge
 
         $sql = $this->_createTable($table, false, $attributes);
 
-        if (is_bool($sql)) {
-            $this->reset();
-            if ($sql === false) {
-                if ($this->db->DBDebug) {
-                    throw new DatabaseException('This feature is not available for the database you are using.');
-                }
-
-                return false;
-            }
-
-            return true;
-        }
-
         if (($result = $this->db->query($sql)) !== false) {
             if (isset($this->db->dataCache['table_names']) && ! in_array($table, $this->db->dataCache['table_names'], true)) {
                 $this->db->dataCache['table_names'][] = $table;
@@ -538,7 +525,7 @@ class Forge
     }
 
     /**
-     * @return bool|string
+     * @return string
      *
      * @deprecated $ifNotExists is no longer used, and will be removed.
      */
--- a/system/Database/OCI8/Forge.php
+++ b/system/Database/OCI8/Forge.php
@@ -33,14 +33,14 @@ class Forge extends BaseForge
     protected $createDatabaseStr = false;
 
     /**
-     * CREATE TABLE IF statement
-     *
-     * @var false
+     * @deprecated This is no longer used.
      */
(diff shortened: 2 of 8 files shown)


system/Database/Forge.php
```diff
@@ -531,19 +531,9 @@ public function createTable(string $table, bool $ifNotExists = false, array $att
     /**
      * @return bool|string
      */
-    protected function _createTable(string $table, bool $ifNotExists, array $attributes)
+    protected function _createTable(string $table, array $attributes)
```
MGatner, line 534: We could leave the parameter and mark it deprecated to prevent this from being a breaking change. Since this method is prefixed with an underscore I think we have slight plausible immunity. @kenjis What do you think?

user_guide_src/source/dbmgmt/forge.rst
```diff
@@ -191,8 +191,8 @@ with
 
 .. literalinclude:: forge/014.php
 
-An optional second parameter set to true adds an ``IF NOT EXISTS`` clause
-into the definition
+An optional second parameter set to true will only execute create table statement if
+table does not exist
```
MGatner, line 195: Let's make this clear that it is not relying on the SQL:
```suggestion
An optional second parameter set to true will only execute create table statement if
the table name is not in the array of know tables
```

MGatner (changes requested):
A rewording suggestion. Please also add a note to the upgrade guide. That could wait until we decide on the parameter removal.

MGatner:
@sclubricants Thanks for splitting this out! Your issue and this very specific PR makes it much easier to understand what is happening. I like this approach, I think the only drawback is a possible race condition but since forge operations should not ever be under load in production I think the benefits outweigh that slim possibility.

system/Database/Forge.php
sclubricants: @MGatner Also its a protected function, shouldn't that have limit its use?

sclubricants: @MGatner I added not to upgrade guide. I'm not sure exactly what you are looking for though.

system/Database/Forge.php
```diff
@@ -106,13 +106,6 @@ class Forge
      */
     protected $createTableStr = "%s %s (%s\n)";
 
-    /**
-     * CREATE TABLE IF statement
-     *
-     * @var bool|string
-     */
-    protected $createTableIfStr = 'CREATE TABLE IF NOT EXISTS';
```
kenjis, line 114: Don't remove non private properties. Please mark it deprecated.
```
@deprecated This is no longer used.
```

system/Database/Forge.php
kenjis: Don't change the method signature. Please add `@deprecated $ifNotExists is no longer used`.

kenjis:
I don't agree that changing any existing method signature in `develop` (4.2.x).
These change should go at least `4.3`.

I don't see anything particularly wrong with this PR, but it changes the behavior (SQL statements will change).
So I would like to hear a little more from others.

MGatner:
@michalsn any chance you have time for a review?

michalsn (approved):
The proposed changes are reasonable and will solve the initial problem.

I can't see any real downsides maybe except for some performance decrease. But since the `Forge` class is mainly used during testing, we should be fine.

We just need some small tweaks.

system/Database/OCI8/Forge.php
```diff
@@ -40,9 +40,7 @@ class Forge extends BaseForge
     protected $createTableIfStr = false;
 
     /**
-     * DROP TABLE IF EXISTS statement
-     *
-     * @var false
+     * @deprecated This is no longer used.
      */
     protected $dropTableIfStr = false;
```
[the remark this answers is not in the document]
sclubricants: Woops, accidently marked the wrong one.

system/Database/SQLSRV/Forge.php
```diff
@@ -91,24 +91,14 @@ class Forge extends BaseForge
     protected $createTableIfStr;
 
     /**
-     * CREATE TABLE statement
-     *
-     * @var string
+     * @deprecated This is no longer used.
      */
     protected $createTableStr;
```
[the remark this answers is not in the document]
sclubricants: I don't know what happened. I'll fix these.

system/Database/Forge.php
```diff
@@ -530,20 +535,12 @@ public function createTable(string $table, bool $ifNotExists = false, array $att
...
-            if ($this->db->tableExists($table)) {
-                return true;
-            }
-
-            $ifNotExists = false;
-        }
-
-        $sql = ($ifNotExists) ? sprintf($this->createTableIfStr, $this->db->escapeIdentifiers($table))
-            : 'CREATE TABLE';
+        $sql = 'CREATE TABLE';
```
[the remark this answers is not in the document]
sclubricants: fixed

system/Database/Forge.php
sclubricants: @kenjis I reverted the method signature as well as the protected properties.

system/Database/Forge.php
```diff
@@ -107,9 +107,7 @@ class Forge
     protected $createTableStr = "%s %s (%s\n)";
 
     /**
-     * CREATE TABLE IF statement
-     *
-     * @var bool|string
```
kenjis, line 112: Don't remove these lines.

system/Database/OCI8/Forge.php
```diff
@@ -33,9 +33,7 @@ class Forge extends BaseForge
     protected $createDatabaseStr = false;
 
     /**
-     * CREATE TABLE IF statement
-     *
-     * @var false
```
kenjis, line 38: Don't remove these lines.

system/Database/SQLSRV/Forge.php
```diff
@@ -84,9 +84,7 @@ class Forge extends BaseForge
     ];
 
     /**
-     * CREATE TABLE IF statement
-     *
-     * @var string
```
kenjis, line 89: Don't remove these lines.

user_guide_src/source/installation/upgrade_422.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Mandatory File Changes
+**********************
+
+
+Breaking Changes
+****************
+
+- The method ``Forge::createTable()`` no longer executes a ``CREATE TABLE IF NOT EXISTS``. If table is not found in ``$db->tableExists($table)`` then ``CREATE TABLE`` is executed.
+- The method signature of ``Forge::_createTable()`` is deprecated. The ``bool``  ``$ifNotExists`` is no longer used and will be removed in a future release.
```
kenjis, line 23:
```suggestion
- The second parameter ``$ifNotExists`` of ``Forge::_createTable()`` is deprecated. It is no longer used and will be removed in a future release.
```

system/Database/SQLSRV/Forge.php
sclubricants: @kenjis Fixed

kenjis (approved):
Thank you!
LGTM!

MGatner (approved):

sclubricants:
@kenjis I just realized a couple more changes.

_createTable() no longer will return a bool.

```php
    /**
     * @return string
     *
     * @deprecated $ifNotExists is no longer used, and will be removed.
     */
    protected function _createTable(string $table, bool $ifNotExists, array $attributes)
    {
```

And remove from createTable() logic to handle bool from _createTable().

```php
        if (is_bool($sql)) {
            $this->reset();
            if ($sql === false) {
                if ($this->db->DBDebug) {
                    throw new DatabaseException('This feature is not available for the database you are using.');
                }

                return false;
            }

            return true;
        }
```

kenjis:
@sclubricants You are correct. Good catch!

kenjis (approved):

PR #6250, opened 2022-07-11
Add db port entry into env file.

nalakapws:
I think it's good to have db port entry in the env file so we don't have to touch the Database.config file to change the port number.

### DIFF
--- a/env
+++ b/env
@@ -45,6 +45,7 @@
 # database.default.password = root
 # database.default.DBDriver = MySQLi
 # database.default.DBPrefix =
+# database.default.port = 3306
 
 # database.tests.hostname = localhost
 # database.tests.database = ci4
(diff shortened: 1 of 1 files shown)


kenjis:
Thank you for contributing.

You can add `port` or any propertiy in Config classes to your own `.env`.
Why do we need to add it `env`?

nalakapws:
I think the port setting entry is equelly important as other database-related entries in the env file like hostname, database, etc... Because any database connection requires hostname, database, username, password and port number. And even DSN records also include the port number.

Of course we can add it manully into the .env file. But it's convenient if that setting in the env file by default. So we don't have to do it manually.

MGatner:
I don't have a preference. Our example **env** file is intentionally not comprehensive but this seems to be the only missing database configuration field missing for a typical connection.

nalakapws:
> I don't have a preference. Our example **env** file is intentionally not comprehensive but this seems to be the only missing database configuration field missing for a typical connection. I'd actually argue `port` is more likely to be changed that `DBPrefix`.

For me having DBPrefix also very useful in the env file.

kenjis (approved):

paulbalandan (approved):

PR #6253, opened 2022-07-11
fix: ValidationInterface

### LINKED ISSUE #6224: Bug: The Validation::run() method does not match the ValidationInterface.

kenjis:
**Description**
Fixes #6224
Alternative #6247

- fix `ValidationInterface::run()` param
- add missing
  - `ValidationInterface::setRule()` 
  - `ValidationInterface::getRules()`
  - `ValidationInterface::getRuleGroup()`  
  - `ValidationInterface::setRuleGroup()`
  - `ValidationInterface::loadRuleGroup()`
  - `ValidationInterface::hasError()`
  - `ValidationInterface::listErrors()`
  - `ValidationInterface::showError()`
- fix `Validation::loadRuleGroup()`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -19,7 +19,6 @@ use CodeIgniter\Database\Exceptions\DataException;
 use CodeIgniter\Exceptions\ModelException;
 use CodeIgniter\I18n\Time;
 use CodeIgniter\Pager\Pager;
-use CodeIgniter\Validation\Validation;
 use CodeIgniter\Validation\ValidationInterface;
 use Config\Services;
 use InvalidArgumentException;
@@ -198,7 +197,7 @@ abstract class BaseModel
     /**
      * Our validator instance.
      *
-     * @var Validation
+     * @var ValidationInterface
      */
     protected $validation;
 
@@ -326,7 +325,7 @@ abstract class BaseModel
         $this->tempAllowCallbacks = $this->allowCallbacks;
 
         /**
-         * @var Validation|null $validation
+         * @var ValidationInterface|null $validation
          */
         $validation ??= Services::validation(null, false);
         $this->validation = $validation;
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -499,7 +499,7 @@ class Validation implements ValidationInterface
      *
      * @param string $group Group.
      *
-     * @throws InvalidArgumentException If group not found.
+     * @throws ValidationException If group not found.
      *
      * @return string[] Rule group.
      */
@@ -521,7 +521,7 @@ class Validation implements ValidationInterface
      *
      * @param string $group Group.
      *
-     * @throws InvalidArgumentException If group not found.
+     * @throws ValidationException If group not found.
      */
     public function setRuleGroup(string $group)
     {
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -405,7 +405,7 @@ class Validation implements ValidationInterface
      *
      *    [
      *        'rule' => 'message',
-     *        'rule' => 'message'
+     *        'rule' => 'message',
      *    ]
      *
      * @param array|string $rules
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -331,6 +331,7 @@ class Validation implements ValidationInterface
 
                 $param = ($param === false) ? '' : $param;
 
+                // @phpstan-ignore-next-line $error may be set by rule methods.
                 $this->errors[$field] = $error ?? $this->getErrorMessage(
                     $rule,
                     $field,
--- a/system/Validation/Validation.php
+++ b/system/Validation/Validation.php
@@ -593,12 +593,14 @@ class Validation implements ValidationInterface
      * same format used with setRules(). Additionally, check
      * for {group}_errors for an array of custom error messages.
      *
-     * @return array|ValidationException|null
+     * @throws ValidationException
+     *
+     * @return array
      */
     public function loadRuleGroup(?string $group = null)
     {
         if (empty($group)) {
-            return null;
+            return [];
         }
 
         if (! isset($this->config->{$group})) {
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -636,11 +636,6 @@ parameters:
 			path: system/Throttle/Throttler.php
 
 		-
-			message: "#^Variable \\$error on left side of \\?\\? always exists and is always null\\.$#"
-			count: 1
-			path: system/Validation/Validation.php
-
-		-
 			message: "#^Property CodeIgniter\\\\View\\\\Cell\\:\\:\\$cache \\(CodeIgniter\\\\Cache\\\\CacheInterface\\) in empty\\(\\) is not falsy\\.$#"
 			count: 2
 			path: system/View/Cell.php
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -49,7 +49,7 @@ use CodeIgniter\Security\Security;
 use CodeIgniter\Session\Session;
 use CodeIgniter\Throttle\Throttler;
 use CodeIgniter\Typography\Typography;
-use CodeIgniter\Validation\Validation;
+use CodeIgniter\Validation\ValidationInterface;
 use CodeIgniter\View\Cell;
 use CodeIgniter\View\Parser;
 use CodeIgniter\View\RendererInterface;
@@ -127,7 +127,7 @@ use Config\View as ConfigView;
  * @method static Toolbar                    toolbar(ConfigToolbar $config = null, $getShared = true)
  * @method static Typography                 typography($getShared = true)
  * @method static URI                        uri($uri = null, $getShared = true)
- * @method static Validation                 validation(ConfigValidation $config = null, $getShared = true)
+ * @method static ValidationInterface        validation(ConfigValidation $config = null, $getShared = true)
  * @method static Cell                       viewcell($getShared = true)
  */
 class BaseService
--- a/system/Config/Services.php
+++ b/system/Config/Services.php
@@ -54,6 +54,7 @@ use CodeIgniter\Session\Session;
 use CodeIgniter\Throttle\Throttler;
 use CodeIgniter\Typography\Typography;
 use CodeIgniter\Validation\Validation;
+use CodeIgniter\Validation\ValidationInterface;
 use CodeIgniter\View\Cell;
 use CodeIgniter\View\Parser;
 use CodeIgniter\View\RendererInterface;
@@ -733,7 +734,7 @@ class Services extends BaseService
     /**
      * The Validation class provides tools for validating input data.
      *
-     * @return Validation
+     * @return ValidationInterface
      */
     public static function validation(?ValidationConfig $config = null, bool $getShared = true)
     {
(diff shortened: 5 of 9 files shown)


system/Validation/Validation.php
```diff
@@ -559,12 +559,14 @@ protected function loadRuleSets()
...
      *
-     * @return array|ValidationException|null
+     * @throws ValidationException
+     *
+     * @return array
      */
     public function loadRuleGroup(?string $group = null)
     {
         if (empty($group)) {
-            return null;
```
MGatner, line 567: What did "null" mean? Would anyone have been using this? Why do we even allow `null` as a parameter if we're just going to return immediately...

MGatner (commented):
I like this much better. Let's still see what happens on the main thread but this is a good example of modifying the interface. I expect this to have way lower impact than changing Model constructor signature.

system/Validation/Validation.php
kenjis: It came from https://github.com/codeigniter4/CodeIgniter4/pull/1769/files#diff-fa78800b3fc00024660b9f8090bd6bee512cd4fbedca371233eb5faa9c57cb06

I'm not sure. It seems just a mistake.

kenjis: The return type was void at first, and changed to void and array.
https://github.com/codeigniter4/CodeIgniter4/commit/37dbc1a8#diff-fa78800b3fc00024660b9f8090bd6bee512cd4fbedca371233eb5faa9c57cb06

But `$this->rules` is an array, so it should be return `[]` when there are no rules.

kenjis: > Why do we even allow null as a parameter if we're just going to return immediately...

At least, now there is no place that expects null in the system code.
`getValidationRules()` uses this method and must return an array.
So if this returns null, it causes an error.

MGatner: I agree with your change. Probably should leave the parameter, makes me a bit sad though. Some days I just want to break everything and redo it the right way and release version 5 this month...

MGatner:
@kenjis Can you rebase this? I would like to propose that we merge this for `4.3` and see what kind of uproar it causes (if any), and if it seems generally well-received then we proceed with the other interfaces identified in #5909

kenjis:
Rebased.

MGatner (approved):
Good by me! Offering a final chance to #develop

MGatner:
@iRedds have you seen this?

iRedds:
@MGatner In fact, I think that interface hell is going on in the framework. In particular, this applies to the HTTP layer. When instead of describing the public API in interfaces, methods are moved to traits. Instead of inheritance, interfaces are enumerated in a class.
The fact that the interfaces are being fixed is good.

But for some reason these methods are missing.
- setRule
- getRules
- getRuleGroup
- setRuleGroup

system/BaseModel.php
```diff
@@ -198,7 +197,7 @@ abstract class BaseModel
     /**
      * Our validator instance.
      *
-     * @var Validation
+     * @var ValidationInterface
      */
     protected $validation;
```
iRedds, line 202: It seems to me that it makes sense to specify the property type directly instead of docBlock in BaseModel and Controller.

kenjis: The existing user's Model will break.
https://3v4l.org/NkQVr

paulbalandan (approved):

system/BaseModel.php
iRedds: Why override the property?

kenjis: Good question. I don't know, but we can override.
I should have said, the existing user's Model might break.

MGatner:
> interface hell is going on in the framework

Oh very much so! Please see https://github.com/codeigniter4/CodeIgniter4/pull/5909 if you haven't already. Basically after much discussion: we've decided to update the interfaces with the missing methods that are already used internally. This is a big "no no" but so is the current state, and this has the least likely path to ruin people's existing code. 

> But for some reason these methods are missing.
> 
> * setRule
> * getRules
> * getRuleGroup
> * setRuleGroup

I'm on mobile so can't really search the framework right now. @kenjis have you determined these aren't necessary for the interface?

MGatner:
Each Service should return an Interface or Abstract Class (when relevant). Remember that Services is our "alternative to a Container. So yes: we should add these methods if they are needed for our service locator to support the interface.

iRedds:
> Oh very much so! Please see #5909 if you haven't already.

Maybe it's easier to announce the development of the 5th version and already start writing correctly?
Because it's not just about interfaces.
With the current activity, as well as the number of strange and incomprehensible implementations, the development of the 5th version will take a year or 1.5. Look at Yii3, it has not been released for 3 years now.
And given the timing, it already needs to be done on php8.1

MGatner:
It's a bigger decision than "how to develop this"; OSS is largely software-driven but there are still business concerns. In this case the CodeIgniter Foundation is essentially our "stakeholder" and though we have a lot of discretion the major release cycle is an area where we have to be attentive.

I did push to start version 5 last April with a goal of a December release (to correspond with the dropping of 7.4 from security support) but the reality was we don't have enough contributors to accomplish everything that needs to happen in order to make a major version "count". I still think v5 is our next likely big release after 4.3 (maybe 4.4?) but ultimately that will be more of empirical decision based on workload versus contributions. 

The good news? I'm committed to [a healthy refactoring process](https://matthiasnoback.nl/2021/02/refactoring-prepare-to-stop/) which means as soon as we have a `v5` branch _it will always represent a usable state_ (non-production), and there's a **LOT** of low-hanging fruit. The jump from 7.4 minimum to 8.0 minimum unlocks an entirely new toolset and will give us a rich opportunity to tackle our typing-problems in a Grade A manner. 

TL;DR we're not there yet but there's a lot of hope for the journey ahead.

kenjis:
Do we need to add the following methods (except constructor)?

```console
$ grep 'public function' system/Validation/ValidationInterface.php  | cut -f 7 -d ' ' | cut -f 1 -d '(' > ValidationInterface.methods
$ grep 'public function' system/Validation/Validation.php | cut -f 7 -d ' ' | cut -f 1 -d '(' > Validation.methods
$ diff -u Validation.methods ValidationInterface.methods 
--- Validation.methods	2022-09-06 06:50:03.000000000 +0900
+++ ValidationInterface.methods	2022-09-06 06:49:59.000000000 +0900
@@ -1,17 +1,8 @@
-__construct
 run
 check
 withRequest
-setRule
 setRules
-getRules
 hasRule
-getRuleGroup
-setRuleGroup
-listErrors
-showError
-loadRuleGroup
-hasError
 getError
 getErrors
 setError
```

iRedds:
If these methods are not documented, they are not in the interface and they are not used by the framework, then maybe they are not needed?

kenjis:
The above result is on `4.3` branch.
On this PR branch:

```console
$ grep 'public function' system/Validation/Validation.php | cut -f 7 -d ' ' | cut -f 1 -d '(' | sort > Validation.methods
$ grep 'public function' system/Validation/ValidationInterface.php  | cut -f 7 -d ' ' | cut -f 1 -d '(' | sort > ValidationInterface.methods
$ diff -u Validation.methods ValidationInterface.methods 
--- Validation.methods	2022-09-06 08:22:47.000000000 +0900
+++ ValidationInterface.methods	2022-09-06 08:22:59.000000000 +0900
@@ -1,9 +1,6 @@
-__construct
 check
 getError
 getErrors
-getRuleGroup
-getRules
 hasError
 hasRule
 listErrors
@@ -11,8 +8,6 @@
 reset
 run
 setError
-setRule
-setRuleGroup
 setRules
 showError
 withRequest
```

kenjis:
`getRuleGroup()` has been documented in the sample code:
https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html?highlight=getrules#getting-setting-rule-groups

kenjis:
`getRules()` is used in tests.

MGatner:
Ugh. So much for picking an easy one to start with 😕 What matters at the end of the day:
1. If a component defines an interface and is locatable via Services, the service return type should be the interface
2. If another component relies on the service as a dependency (injected or located) and currently uses methods not in that dependency: they must be added

It is acceptable to have extra public methods on a component class, even if they are documented, so long as we aren't building dependencies around them. 

So for the method list provided... my guess is that we will need them all but I don't think there is a way to know without tracing back through the uses. One way to do this (I think this is how kenjis started) is to switch all the component class references to the interface (e.g. `Validation` to `ValidationInterface`) and let static analysis tell us the problems.

kenjis:
> But for some reason these methods are missing.
> - setRule
> - getRules
> - getRuleGroup
> - setRuleGroup

Added these methods in the interface.

kenjis:
Now `ValidationInterface` has all public methods in `Validation`.

```console
$ grep 'public function' system/Validation/Validation.php | cut -f 7 -d ' ' | cut -f 1 -d '(' | sort > Validation.methods
$ grep 'public function' system/Validation/ValidationInterface.php | cut -f 7 -d ' ' | cut -f 1 -d '(' | sort > ValidationInterface.methods
$ diff -u Validation.methods ValidationInterface.methods 
--- Validation.methods	2022-09-06 11:12:11.000000000 +0900
+++ ValidationInterface.methods	2022-09-06 11:12:19.000000000 +0900
@@ -1,4 +1,3 @@
-__construct
 check
 getError
 getErrors
```

kenjis:
I think the implementation is done.
If this is okay, I will add the documentation.

system/Validation/Validation.php
```diff
@@ -499,7 +500,7 @@ public function hasRule(string $field): bool
      *
      * @param string $group Group.
      *
-     * @throws InvalidArgumentException If group not found.
+     * @throws ValidationException If group not found.
```
MGatner, line 503: `InvalidArgumentException` seems like the correct throw here. It doesn't look like you touched the actual method - was this some previous change to consolidate exceptions or something?

MGatner (commented):
Yes the implementation looks good to me.

system/Validation/Validation.php
kenjis: It seems it was InvalidArgumentException but changed to ValidationException.
I just update the doc comment.
See the code: https://github.com/codeigniter4/CodeIgniter4/blob/a127d7190dcb709488d53104c8c84993b8233bd0/system/Validation/Validation.php#L507-L518

MGatner: Beyond the scope of this PR but I think that is a mistake. Catching a "validation exception" should not cover logic errors that were unrelated to the validation process itself.

MGatner: Never mind, I have it backwards: exceptions are never thrown when input fails validation rules. These are correct.

kenjis:
Added the docs.

iRedds (approved):

MGatner (approved):
Here goes! 😳

PR #6257, opened 2022-07-12
fix: use getenv() instead of $_SERVER in detectEnvironment()

fcosrno:
**Description**
Update CodeIgniter.php to use getenv() instead of $_SERVER in detectEnvironment. The outcome of environment variables loading in either  $_SERVER or $_ENV depends on the operating system and web server. For instance, in a Docker environment the `.env` file loads variables in $_ENV, resulting in detectEnvironment() not seeing CI_ENVIRONMENT when defined. Using getenv() instead of directly reading from $_SERVER solves this issue.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -528,7 +528,7 @@ class CodeIgniter
     {
         // Make sure ENVIRONMENT isn't already set by other means.
         if (! defined('ENVIRONMENT')) {
-            define('ENVIRONMENT', getenv('CI_ENVIRONMENT') ?? 'production');
+            define('ENVIRONMENT', env('CI_ENVIRONMENT', 'production'));
         }
     }
 
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -528,7 +528,7 @@ class CodeIgniter
     {
         // Make sure ENVIRONMENT isn't already set by other means.
         if (! defined('ENVIRONMENT')) {
-            define('ENVIRONMENT', $_SERVER['CI_ENVIRONMENT'] ?? 'production');
+            define('ENVIRONMENT', getenv('CI_ENVIRONMENT') ?? 'production');
         }
     }
 


system/CodeIgniter.php
```diff
@@ -528,7 +528,7 @@ protected function detectEnvironment()
     {
         // Make sure ENVIRONMENT isn't already set by other means.
         if (! defined('ENVIRONMENT')) {
-            define('ENVIRONMENT', $_SERVER['CI_ENVIRONMENT'] ?? 'production');
+            define('ENVIRONMENT', getenv('CI_ENVIRONMENT') ?? 'production');
```
samsonasik, line 531:
```suggestion
            define('ENVIRONMENT', getenv('CI_ENVIRONMENT') ?: 'production');
```

ref:

-  https://phpstan.org/r/f3cfec8d-8e15-4812-8a62-6036fe2d362c
- https://3v4l.org/EIKDT
- https://3v4l.org/4p0B8

kenjis:
> in a Docker environment the .env file loads variables in $_ENV, resulting in detectEnvironment() not seeing CI_ENVIRONMENT when defined. Using getenv() instead of directly reading from $_SERVER solves this issue.

Can't reproduce.

I checked `phpinfo()` output in my Docker environment.
- docker desktop 4.10.1
   - Engine: 20.10.17
   - Compose: v2.6.1

```
Environment
CI_ENVIRONMENT | development

PHP Variables
$_SERVER['CI_ENVIRONMENT'] | development
$_ENV['CI_ENVIRONMENT'] | development
```

kenjis:
If you have `.env` file, `setVariable()` sets `$_SERVER[$name]` if it is `empty()`.
I don't know why your `$_SERVER['CI_ENVIRONMENT']` is not set.

https://github.com/codeigniter4/CodeIgniter4/blob/d75e426a56e629a93a2fc8c9dc13520eb08b786f/system/Config/DotEnv.php#L89-L102

fcosrno:
Yes @kenjis, that should be the case in the presence of a .env file, like when you COPY the file in the Dockerfile. But it's not the case when loading environment variables in other scenarios, like when using `env_file` in Docker Compose or in deployments without persistent date (no copying of .env files into containers). In my original comment I should have written "in a Docker environment _the environment variables could load_ in $_ENV" and never in $_SERVER instead of explicitly calling out the .env file.

kenjis:
Okay, I got your situation. You don't use `.env` file and set environment variable only.
And your PHP does not set `$_SERVER['CI_ENVIRONMENT']`. Maybe not FastCGI API.

By the way, why do we use `$_SERVER['CI_ENVIRONMENT']`?
It comes from environment variable, so why don't we use `$_ENV['CI_ENVIRONMENT']`?

system/CodeIgniter.php
paulbalandan, line 531: Why not use `env()` instead, to cover all uses?

kenjis:
env() is slower than getenv().

`$iterator->run(100000)`
<table>
			<thead>
				<tr>
					<td>Test</td>
					<td>Time</td>
					<td>Memory</td>
				</tr>
			</thead>
			<tbody>
				<tr>
				<td>env()</td>
				<td>0.3829</td>
				<td>0 Bytes</td>
			</tr><tr>
				<td>getenv</td>
				<td>0.3150</td>
				<td>0 Bytes</td>
			</tr>
			</tbody>
		</table>

paulbalandan:
It may be slower because it checks all available places (`$_SERVER`, `$_ENV`, `getenv()`). But in cases where `putenv` is disabled for security reason, wouldn't `getenv` return `false` even if `CI_ENVIRONMENT` is really set?

kenjis:
1. Disable `putenv()`.

`php.ini`:
```
disable_functions = putenv
```

2. Remove `.env` file.

3. Make a controller.

```php
{
    public function index()
    {
        return getenv('CI_ENVIRONMENT');
    }
}
```

4. Run local server with Environment Variable.

```console
$ CI_ENVIRONMENT=development php spark serve
```

5. Navigate to http://localhost:8080/.
 
I see "development".

kenjis:
It seems there are three options:
1. use $_ENV
2. use $_ENV or getenv() (If $_ENV is not set, call getenv())
3. use env()

system/CodeIgniter.php
MGatner: I like that, it essentially covers all the scenarios:
```php

    function env(string $key, $default = null)
    {
        $value = $_ENV[$key] ?? $_SERVER[$key] ?? getenv($key);

        // Not found? Return the default value
        if ($value === false) {
            return $default;
        }
...
```
And it will always be available by now:
1. Entry point (public/index.php or spark)
2. bootstrap.php
3. DotEnv.ohp
4. Common.php
5. CodeIgniter.php

MGatner, line 531: See Paul's suggestion ~~below~~ above:
```suggestion
            define('ENVIRONMENT', env('CI_ENVIRONMENT', 'production'));

```
It might seem cyclical but I checked and everything needed is loaded by this point.

MGatner (changes requested):

MGatner:
Oops, I reviewed before I saw this discussion. I'm fine with .07s performance hit if `env()` covers more possible scenarios, but if it adds nothing then going with the more barebones approach seems good. One other advantage of `env()` is that a developer could theoretically override it if they wanted to do some fancy loading of their own.

kenjis:
`variables_order` in `php.ini-production` is `GPCS`.
So if not in CGI or FastCGI, `$_ENV` is not populated by default.
Using only $_ENV seems not good.

> ; This directive determines which super global arrays are registered when PHP
; starts up. G,P,C,E & S are abbreviations for the following respective super
; globals: GET, POST, COOKIE, ENV and SERVER. There is a performance penalty
; paid for the registration of these arrays and because ENV is not as commonly
; used as the others, ENV is not recommended on productions servers. You
; can still get access to the environment variables through getenv() should you
; need to.
; Default Value: "EGPCS"
; Development Value: "GPCS"
; Production Value: "GPCS";
; https://php.net/variables-order
variables_order = "GPCS"
https://github.com/php/php-src/blob/32d55f74229e7913db0d59ef874a401744479b6a/php.ini-production#L647

> Warning
In both the CGI and FastCGI SAPIs, [$_SERVER](https://www.php.net/manual/en/reserved.variables.server.php) is also populated by values from the environment; S is always equivalent to ES regardless of the placement of E elsewhere in this directive.
https://www.php.net/manual/en/ini.core.php#ini.variables-order

MGatner:
@fcosrno Are you available to confirm that `env()` fixes your issue? If so please update this PR. 

If we don't hear back today I will merge https://github.com/codeigniter4/CodeIgniter4/pull/6257#discussion_r919969434 and we can proceed with final review.

fcosrno:
Yes, I can confirm env() fixes the issue.

kenjis (approved):

kenjis (approved):
Okay, all green!
@fcosrno Thank you for confirmation.

MGatner:
💪

PR #6263, opened 2022-07-13
fix: format_number() can't be used on CLI

kenjis:
**Description**
`format_number()` depends on `IncommingRequest::getLocale()`, and causes Error `Call to undefined method CodeIgniter\HTTP\CLIRequest::getLocale()`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -35,6 +35,7 @@ use Exception;
 use Kint;
 use Kint\Renderer\CliRenderer;
 use Kint\Renderer\RichRenderer;
+use Locale;
 use LogicException;
 
 /**
@@ -190,7 +191,7 @@ class CodeIgniter
         }
 
         // Set default locale on the server
-        locale_set_default($this->config->defaultLocale ?? 'en');
+        Locale::setDefault($this->config->defaultLocale ?? 'en');
 
         // Set default timezone on the server
         date_default_timezone_set($this->config->appTimezone ?? 'UTC');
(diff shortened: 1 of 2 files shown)


system/Helpers/number_helper.php
```diff
@@ -129,8 +127,9 @@ function number_to_currency(float $num, string $currency, ?string $locale = null
      */
     function format_number(float $num, int $precision = 1, ?string $locale = null, array $options = []): string
     {
-        // Locale is either passed in here, negotiated with client, or grabbed from our config file.
-        $locale ??= Services::request()->getLocale();
+        // If locate is not passed, get from the default locale that is set from our config file
```
MGatner, line 130:
```suggestion
        // If locale is not passed, get from the default locale that is set from our config file
```

MGatner (approved):
One small typo but otherwise looks good 👍

system/CodeIgniter.php
```diff
@@ -190,7 +191,7 @@ public function initialize()
         }
 
         // Set default locale on the server
-        locale_set_default($this->config->defaultLocale ?? 'en');
+        Locale::setDefault($this->config->defaultLocale ?? 'en');
```
paulbalandan, line 194: I'm curious why you changed this. Isn't they the same?

kenjis: Now we are using `Locale::getDefault()` and I searched `Locale::setDefault()` but not found.
Using `locale_set_default()` and `Locale::getDefault()` is not consistent.

samsonasik:
Thank you @kenjis

PR #6264, opened 2022-07-13
use SVG logo on Welcome page

totoprayogo1916:
Use SVG like on Official Site.
The SVG's logo grabbed from codeigniter.com and optimized on https://jakearchibald.github.io/svgomg/

before:
image

after:
image

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


MGatner (approved):
Looks good. Can you speak to why you lived the image `alt` to a link `title`? I'm no expert but I think those accomplish different things.

paulbalandan:
I think the title to the `<a>` is to serve as tooltip when hovered.

totoprayogo1916:
Instead of the `alt` of the unused image, I move it as `a title`.
The reference of `a title`,  https://html.com/attributes/a-title/

kenjis:
> Instead of the alt of the unused image,

What do you mean by unused? It seems the image (logo) is used.

`alt` and `title` are different things. Why do you remove `alt`?

totoprayogo1916:
when using the `img` tag there is an `alt` attribute, while `svg` not.
so, the `alt` of the `img` tag is not used, then I move it as `title` in `a href`.

kenjis:
Oh, thanks. I got it.

I recommend you use `aria-label`:
```
<svg role="img" aria-label="Visit CodeIgniter.com official website!" ...>
```
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role#svg_and_roleimg

totoprayogo1916:
> I recommend you use `aria-label`

Updated

kenjis (approved):
Thank you!

PR #6275, opened 2022-07-18
config: update phpunit.xml.dist

kenjis:
**Description**
- update indentation
- remove unnecessary env XDEBUG_MODE

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -70,7 +70,6 @@
     </extensions>
 
     <php>
-        <env name="XDEBUG_MODE" value="coverage"/>
         <server name="app.baseURL" value="http://example.com/"/>
         <!-- Directory containing phpunit.xml -->
         <const name="HOMEPATH" value="./"/>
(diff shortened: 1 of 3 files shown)


phpunit.xml.dist
```diff
@@ -70,7 +70,6 @@
     </extensions>
 
     <php>
-        <env name="XDEBUG_MODE" value="coverage"/>
```
MGatner, line 73: Care to explain? I think this is required for code coverage to work. Typically a developer would set this in their environment but I think setting it here covers Actions and other CI pipelines as well.

kenjis: It does nothing. It is too late to set the environment variable.
See https://github.com/sebastianbergmann/php-code-coverage/issues/834#issuecomment-736324897

And you can see Coveralls got the coverage on this PR.

PR #6279, opened 2022-07-19
fix: `spark routes` shows 404 error when using regex

### LINKED ISSUE #6276: Bug: `spark routes` show error when using custom placeholder (regex)
### PHP Version

7.4

### CodeIgniter4 Version

4.2.1

Manual (zip or tar.gz)

Windows

### Which server did you use?

### Database

MariaDB

### What happened?

when run `php spark routes`
show 404 error like:
image

### Steps to Reproduce

add route
```php
// app/Config/Routes.php
$routes->get('detail/([a-z0-9]{40})', 'User::detail/$1');
// or add using custom placeholder

$routes->addPlaceholder('uid', '[a-z0-9]{40}');
$routes->get('detail/(:uid)', 'User::detail/$1');
```
then run 
```bash
php spark routes
```

### Expected Output

show route lists as expected

### Anything else?

_No response_

kenjis:
**Description**
Fixes #6276

**How to Test**
```php
$routes->get('detail/([a-z0-9]{40})', 'User::detail/$1');
```

Before:
```console
ERROR: 404
Can't find a route for 'get: detail/([a-z0-9]{40})'.
```

After:
```console
+--------+-----------------------+------------------------------------------+----------------+---------------+
| Method | Route                 | Handler                                  | Before Filters | After Filters |
+--------+-----------------------+------------------------------------------+----------------+---------------+
| GET    | /                     | \App\Controllers\Home::index             |                | toolbar       |
| GET    | detail/([a-z0-9]{40}) | \App\Controllers\User::detail/$1         | <unknown>      | <unknown>     |
| CLI    | ci(.*)                | \CodeIgniter\CLI\CommandRunner::index/$1 |                |               |
+--------+-----------------------+------------------------------------------+----------------+---------------+
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Utilities/Routes.php
+++ b/system/Commands/Utilities/Routes.php
@@ -18,6 +18,7 @@ use CodeIgniter\Commands\Utilities\Routes\AutoRouteCollector;
 use CodeIgniter\Commands\Utilities\Routes\AutoRouterImproved\AutoRouteCollector as AutoRouteCollectorImproved;
 use CodeIgniter\Commands\Utilities\Routes\FilterCollector;
 use CodeIgniter\Commands\Utilities\Routes\SampleURIGenerator;
+use CodeIgniter\Exceptions\PageNotFoundException;
 use Config\Services;
 
 /**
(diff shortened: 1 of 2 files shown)


iRedds:
I think it would make more sense to catch the exception in FilterFilder::find()

kenjis:
@iRedds Thanks. I was confusing `FilterFinder` with `Filters`.

totoprayogo1916 (approved):

MGatner (approved):
I'm not sure how this fixes the issue but if it does then I'm good with it.

kenjis:
This error is because of the current limitation of the functionality.
We need a real route (without regex) to find its filters, but we can't get a real route from the regex route.
So it tries to find filters for the route `detail/([a-z0-9]{40})`, but there is no such a route.

PR #6282, opened 2022-07-20
fix:  page cache saves Response data before running after filters

### LINKED ISSUE #6281: Bug: secureheaders does not work with cache
### PHP Version

8.1

### CodeIgniter4 Version

4.2.1

Manual (zip or tar.gz)

macOS, Linux

### Which server did you use?

apache

### Database

_No response_

### What happened?

Http headers of "secureheaders" are not present when cache is active

### Steps to Reproduce

1. Active "secureheaders" in "app/Filters.php"

2. Add "$this->cachePage(200);" in your Controller

### Expected Output

The following lines should appear in the http header : 

x-content-type-options: nosniff
x-download-options: noopen
x-frame-options: SAMEORIGIN
x-permitted-cross-domain-policies: none 

### Anything else?

_No response_

kenjis:
**Description**
Fixes #6281

The page cache saves Response data before running after filters.
So response headers added by after filters and response body changes by after filters will be lost.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -21,7 +21,6 @@ use CodeIgniter\HTTP\DownloadResponse;
 use CodeIgniter\HTTP\IncomingRequest;
 use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\HTTP\Request;
-use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
 use CodeIgniter\HTTP\URI;
 use CodeIgniter\Router\Exceptions\RedirectException;
@@ -298,10 +297,9 @@ class CodeIgniter
      * tries to route the response, loads the controller and generally
      * makes all of the pieces work together.
      *
-     * @throws Exception
      * @throws RedirectException
      *
-     * @return bool|mixed|RequestInterface|ResponseInterface|void
+     * @return ResponseInterface|void
      */
     public function run(?RouteCollectionInterface $routes = null, bool $returnResponse = false)
     {
@@ -409,7 +407,7 @@ class CodeIgniter
      * @throws PageNotFoundException
      * @throws RedirectException
      *
-     * @return mixed|RequestInterface|ResponseInterface
+     * @return ResponseInterface
      */
     protected function handleRequest(?RouteCollectionInterface $routes, Cache $cacheConfig, bool $returnResponse = false)
     {
@@ -641,7 +639,7 @@ class CodeIgniter
      *
      * @throws Exception
      *
-     * @return bool|ResponseInterface
+     * @return false|ResponseInterface
      */
     public function displayCache(Cache $config)
     {
@@ -1069,6 +1067,8 @@ class CodeIgniter
     /**
      * Sends the output of this request back to the client.
      * This is what they've been waiting for!
+     *
+     * @return void
      */
     protected function sendResponse()
     {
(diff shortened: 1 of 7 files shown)


iRedds:
Since the issue affects the operation of the SecureHeaders filter, it seems to me that it would also be correct to return the Response in the SecureHeaders::after() method.

kenjis:
What do you mean?
What's the difference between to return the Response in the SecureHeaders::after() and not to return?

iRedds:
@kenjis If we are targeting the implementation of PSR-15 in version 5, then it would be logical to prepare the codebase for the correct behavior. So that it doesn’t work out, as with legacy 3.x, when, with the transfer of the code base from 3.x to 4.x, the code started to work incorrectly or not as intended.

But yes, within the framework of the current concept of the framework, my proposal does not make sense.

kenjis:
@iRedds I got your point. But I think it is related not only SecureHeaders but also all after filters.
The `after()` method should always return `ResponseInterface`.
And if we change the API, at least we must go to 4.3 branch.

We can't do anything in this PR.

kenjis:
Added the upgrade guide.

user_guide_src/source/installation/upgrade_422.rst
```diff
@@ -19,6 +19,17 @@ Mandatory File Changes
 Breaking Changes
 ****************
 
+Web Page Caching Bug Fix
+========================
+
+- :doc:`../general/caching` now caches the Response data after :ref:`after-filters` are executed.
+- For example, if you enables :ref:`secureheaders`, the Response headers are now sent when the page comes from the cache.
```
MGatner, line 26:
```suggestion
- For example, if you enable :ref:`secureheaders`, the Response headers are now sent when the page comes from the cache.
```

user_guide_src/source/installation/upgrade_422.rst
```diff
@@ -19,6 +19,17 @@ Mandatory File Changes
...
+
+.. important:: If you have written **code based on the bug** that changes to the Response in the after filters are not cached, **sensitive information could be cached and compromised**. If this is the case, change your code to disable caching of the page.
```
MGatner, line 28:
```suggestion
.. important:: If you have written **code based on this bug** that assumes changes to the Response in "after" filters are not cached then **sensitive information could be cached and compromised**. If this is the case, change your code to disable caching of the page.
```

MGatner (changes requested):
A few typos then looks good! Thanks for adding the explicit docs.

MGatner (changes requested):
A few typos then looks good! Thanks for adding the explicit docs.

MGatner (approved):

system/CodeIgniter.php
```diff
@@ -298,17 +297,18 @@ protected function initializeKint()
...
-     * @return bool|mixed|RequestInterface|ResponseInterface|void
+     * @return ResponseInterface|void
      */
     public function run(?RouteCollectionInterface $routes = null, bool $returnResponse = false)
     {
         if ($this->context === null) {
             throw new LogicException('Context must be set before run() is called. If you are upgrading from 4.1.x, you need to merge `public/index.php` and `spark` file from `vendor/codeigniter4/framework`.');
         }
 
+        static::$cacheTTL = 0;
```
samsonasik, line 310: is it means that `cacheTTL` value is reset in every run?

kenjis: Yes. 

$cacheTTL will be set by a Controller.

samsonasik (approved):

PR #6283, opened 2022-07-21
feat: `spark filter:check` command

kenjis:
**Description**
- add ``spark filter:check`` command to check filters for a route.

screenshot 2022-07-22 9 56 39

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Utilities/FilterCheck.php
+++ b/system/Commands/Utilities/FilterCheck.php
@@ -1,0 +1,127 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Commands\Utilities;
+
+use CodeIgniter\CLI\BaseCommand;
+use CodeIgniter\CLI\CLI;
+use CodeIgniter\Commands\Utilities\Routes\FilterCollector;
+use Config\Services;
+
+/**
+ * Check filters for a route.
+ */
+class FilterCheck extends BaseCommand
+{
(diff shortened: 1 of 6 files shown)


MGatner (approved):
Splendid!

system/Commands/Utilities/FilterCheck.php
```diff
@@ -0,0 +1,127 @@
...
+     * @var array
+     */
+    protected $options = [];
+
+    /**
+     * @return int exit code
+     */
+    public function run(array $params)
+    {
+        if (count($params) !== 2) {
```
paulbalandan, line 75: I think it would be better to use `! isset($params[1])` to check if all arguments are present. If, for example, a user supplied only the first argument then, out of the blue, supplies an option (e.g. `--force`) then would `count($params)` be 2?

system/Commands/Utilities/FilterCheck.php
```diff
@@ -0,0 +1,127 @@
...
+
+        $filters = $filterCollector->get($method, $route);
+
+        // PageNotFoundException
+        if ($filters['before'] === ['<unknown>']) {
+            CLI::error(
+                "Can't find a route: " .
+                CLI::color(
+                    '"' . strtoupper($method) . ' ' . $route . '"',
+                    'black'
```
paulbalandan, line 102: How would this look on black consoles?

system/Commands/Utilities/FilterCheck.php
kenjis: Good suggestion!
Fixed.

system/Commands/Utilities/FilterCheck.php
kenjis: Fixed.

paulbalandan (approved):

PR #6284, opened 2022-07-21
fix: Entity::hasChanged() is unreliable for casts

### LINKED ISSUE #5905: Bug: \CodeIgniter\Entity::hasChanged() is unreliable for non-string casts

kenjis:
~~Need to rebase after merging #6285~~

**Description**
Fixes #5905

- fix `hasChanged()` is unreliable for casts
- ~~fix `hasChanged()` returns wrong result to mapped property~~ --> #6285

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity/Entity.php
+++ b/system/Entity/Entity.php
@@ -287,7 +287,7 @@ class Entity implements JsonSerializable
     {
         // If no parameter was given then check all attributes
         if ($key === null) {
-            foreach ($this->attributes as $key => $value) {
+            foreach (array_keys($this->attributes) as $key) {
                 if ($this->isChanged($key)) {
                     return true;
                 }
@@ -574,13 +574,11 @@ class Entity implements JsonSerializable
      */
     private function _getCastData(string $key, array $data)
     {
-        $result = null;
-
         if (array_key_exists($key, $data)) {
-            $result = $this->castAs($data[$key], $key);
+            return $this->castAs($data[$key], $key);
         }
 
-        return $result;
+        return null;
     }
 
     /**
--- a/system/Entity/Entity.php
+++ b/system/Entity/Entity.php
@@ -570,7 +570,7 @@ class Entity implements JsonSerializable
     /**
      * Get cast value from the data array.
      *
-     * @return mixed|null
+     * @return array|bool|float|int|object|string|null
      */
     private function _getCastData(string $key, array $data)
     {
--- a/system/Entity/Entity.php
+++ b/system/Entity/Entity.php
@@ -227,7 +227,7 @@ class Entity implements JsonSerializable
         }
 
         foreach ($this->attributes as $key => $value) {
-            if (! $this->hasChanged($key)) {
+            if (! $this->hasChangedAttributes($key)) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


system/Entity/Cast/BaseCast.php
```diff
@@ -19,10 +19,10 @@ abstract class BaseCast implements CastInterface
     /**
      * Get
      *
-     * @param mixed $value  Data
-     * @param array $params Additional param
+     * @param array|bool|float|int|object|string|null $value  Data
+     * @param array                                   $params Additional param
      *
-     * @return mixed
+     * @return array|bool|float|int|object|string|null
```
MGatner, line 25: What do these preclude? `resource`?

MGatner (commented):
This makes me very uncomfortable. I think we've already established that `Entity` is rather confusing, and my opinion on how to improve that is to separate "data source  logic" from "developed logic" and keep Entity methods as a way of traversing those realms.

`hasChanged()` is a _data source logic_ method that I interpret as: "does this instance still correctly represent its originating data?" Casting is a _developer logic_ method: "no matter the underlying data type, represent it like this." This works great as a read-only Database Transfer Object (DTO) but the problem comes when we allow modifying attributes (technically breaks the DTO design pattern). Since sweeping changes are off the table, the way to keep this clean is to keep set methods as _data source logic_ methods which means they need to remain ignorant of the developer representation of the data. Altering `hasChanged()` to compare developer representations of the underlying data source is mixing the two sides of logic in an unhelpful way. 

A silly example... I want to anonymize my user data so I define a new `JoeCast`:
```php
    public function get(): string
    {
        return 'Joe';
    }
```

Then I apply it to my user's names:
```php
class User extends Entity
{
    protected $casts = [
        'firstname' => 'joe',
    ];
}
```

Now I can use my _developer logic_ safely anywhere I want to display a user: `<?= $user->firstname ?>`. However, this sequence of methods that are **entirely data source logic** now fails:
```php
$user->firstname = 'Jill';
model(UserModel::class)->save($user);
```

MGatner:
Addendum, since most of the discussion so far has been about typing (e.g. `35` versus `'35'`)... Entity is mildly tied to `Model` (**not** `BaseModel`) in that the `hasChanged()` methods affect database interactions. Since `Model` is inherently SQL-based there is an impulse to assume `Entity` is a representation of SQL data but I would argue that this is a big violation. I think this assumption is what is mostly behind the "35 age" example we've been using: since the SQL column data type cannot change we are safe to manipulate types in _developer logic_. However as soon as the data source becomes collection, array store, NoSQL database, JSON file, etc... that assumption will wreak havoc.

system/Entity/Cast/BaseCast.php
kenjis: Yes.

system/Entity/Entity.php
```diff
@@ -254,11 +254,34 @@ public function hasChanged(?string $key = null): bool
...
+    }
+
+    /**
+     * Checks a attribute to see if it has changed since the entity
+     * was created. Or, without a parameter, checks if any
+     * attributes have changed.
+     *
+     * @param string|null $key key of $this->attributes
+     */
+    private function hasChangedAttributes(?string $key = null): bool
```
najdanovicivan, line 272: Why add new method here ? Why not just add parameter to hasChanged(?string $key = null, $cast = true)

kenjis: boolean parameters are not good practice.

> What is wrong with boolean parameters? | Understand Legacy Code
https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

MGatner:
@kenjis What's the update here?

kenjis:
Just rebased to resolve conflicts.

Since Entity has "raw data" (values retrieved from a database), it is difficult to determine if the value as PHP has changed. Raw data can change depending on the database driver and/or configuration.

kenjis:
@MGatner I sent a brand new PR #7995

PR #6286, opened 2022-07-22
refactor: Exception exit code

kenjis:
**Description**
- replace magic numbers with constants
- add test cases

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


MGatner (approved):
It looks like part of the problem is that the old CastException was a `CriticalError` while the new one is a `FrameworkException`. Not sure how that slipped by but it makes extending the old one a problem - unless we want to revert that change and have it be an error again.

PR #6288, opened 2022-07-23
error_404.php translation and design

sba:
language strings for error_404.php added and style updated to the same as production.php

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Views/errors/html/error_404.php
+++ b/app/Views/errors/html/error_404.php
@@ -5,74 +5,15 @@
     <title><?= lang('Errors.404pageNotFound') ?></title>
 
     <style>
-        div.logo {
-            height: 200px;
-            width: 155px;
-            display: inline-block;
-            opacity: 0.08;
-            position: absolute;
-            top: 2rem;
-            left: 50%;
-            margin-left: -73px;
-        }
-        body {
-            height: 100%;
-            background: #fafafa;
-            font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
-            color: #777;
-            font-weight: 300;
-        }
-        h1 {
-            font-weight: lighter;
-            letter-spacing: normal;
-            font-size: 3rem;
-            margin-top: 0;
-            margin-bottom: 0;
-            color: #222;
-        }
-        .wrap {
-            max-width: 1024px;
-            margin: 5rem auto;
-            padding: 2rem;
-            background: #fff;
-            text-align: center;
-            border: 1px solid #efefef;
-            border-radius: 0.5rem;
-            position: relative;
-        }
-        pre {
-            white-space: normal;
-            margin-top: 1.5rem;
-        }
-        code {
-            background: #fafafa;
-            border: 1px solid #efefef;
-            padding: 0.5rem 1rem;
-            border-radius: 5px;
-            display: block;
-        }
-        p {
-            margin-top: 1.5rem;
-        }
-        .footer {
-            margin-top: 2rem;
-            border-top: 1px solid #efefef;
-            padding: 1em 2em 0 2em;
-            font-size: 85%;
-            color: #999;
-        }
-        a:active,
-        a:link,
-        a:visited {
-            color: #dd4814;
-        }
+        <?= preg_replace('#[\r\n\t ]+#', ' ', file_get_contents(__DIR__ . DIRECTORY_SEPARATOR . 'debug.css')) ?>
     </style>
 </head>
 <body>
-    <div class="wrap">
-        <h1><?= lang('Errors.404pageNotFound') ?></h1>
+    <div class="container text-center">
 
-        <p>
+        <h1 class="headline"><?= lang('Errors.404pageNotFound') ?></h1>
+
+        <p class="lead">
             <?php if (! empty($message) && $message !== '(null)') : ?>
                 <?= nl2br(esc($message)) ?>
             <?php else : ?>
--- a/app/Views/errors/html/error_404.php
+++ b/app/Views/errors/html/error_404.php
@@ -5,15 +5,74 @@
     <title><?= lang('Errors.404pageNotFound') ?></title>
 
     <style>
-        <?= preg_replace('#[\r\n\t ]+#', ' ', file_get_contents(__DIR__ . DIRECTORY_SEPARATOR . 'debug.css')) ?>
+        div.logo {
+            height: 200px;
+            width: 155px;
+            display: inline-block;
+            opacity: 0.08;
+            position: absolute;
+            top: 2rem;
+            left: 50%;
+            margin-left: -73px;
+        }
+        body {
+            height: 100%;
+            background: #fafafa;
+            font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+            color: #777;
+            font-weight: 300;
+        }
+        h1 {
+            font-weight: lighter;
+            letter-spacing: normal;
+            font-size: 3rem;
+            margin-top: 0;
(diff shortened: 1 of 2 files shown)


kenjis:
Cannot checkout this PR with `gh` command, because the branch name is `4.3`.
It would be nice if you could create a feature branch from the next PR.

```console
$ gh pr checkout 6288
remote: Enumerating objects: 37, done.
remote: Counting objects: 100% (37/37), done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 27 (delta 16), reused 0 (delta 0), pack-reused 0
Unpacking objects: 100% (27/27), 4.09 KiB | 61.00 KiB/s, done.
From github.com:codeigniter4/CodeIgniter4
 ! [rejected]            refs/pull/6288/head -> 4.3  (non-fast-forward)
exit status 1
```

MGatner (changes requested):
Thanks for implementing this! I would like to see this split into two PRs. The translation change is great and I would approve that right now, but I'm on mobile and can't look at the style changes. Please either split it out or post some before/after screenshots.

sba:
> Thanks for implementing this! I would like to see this split into two PRs. The translation change is great and I would approve that right now, but I'm on mobile and can't look at the style changes. Please either split it out or post some before/after screenshots.

Screenshot before style changes:
grafik

After Style changes:
grafik

And after removing duplicate message:
grafik


production.php error message (for layout reference):
grafik

MGatner:
(And thanks for the screenshots!)

demyr17:
> I'm fine with this but maybe we should get some of the design team and other maintainers to sign off as well. @demyr17 @lonnieezell

Thanks, I will be working on it and hope to turn back very soon.

demyr17:
I have used our official font: _Raleway_ and our soft grey background color from the site as well. Plus, I have also tried them with our primary color. Additionally, I am thinking of adding a small note how to see real errors when faced with Whoops. We can place it in small letters under the message. What do you think ?
Ci-404-note
Ci-whoops-note
Ci-404-note-colored
Ci-whoops-note-colored

kenjis:
@demyr17 I prefer gray's.

kenjis:
@sba It seems it is better this PR contains translation only.

You used `git merge`, but it is better to use `git rebase`.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#updating-your-branch

MGatner:
Yes please split apart the translations and styling. 

@demyr17 I would prefer these pages be as generic as possible. There's a certain "first tier" security concern in how easy it is to identify the framework that a website is running, and the 404 page is typically the first place to check. The more we "brand" it the more obvious that is.

The "Whoops" page is a final resort: it should never happen in a healthy app. It's intent is to hide anything incriminating when an exception "escapes", so leaving it free of additional information is best and consistent with its intent.

demyr17:
> 

@MGatner , I think I have been misunderstood about the Whoops page. I didn't mean returning error or any solution. The Whoops message appears when our project is in production mode. So, I thought, we can add a line suggesting turning it into **_development_** mode. Or we can place a link to our documentation, for example:   [what does this message mean?](https://codeigniter.com/user_guide/installation/troubleshooting.html?highlight=whoops#what-s-with-an-unhelpful-whoops-page)

kenjis:
@demyr17 The Whoops should never inform additional information. It should show a server error occurs to end users.
If a dev does not customize the page, it will appear when an unexpected error occurs.

It does not show it is in production mode.

The explanation about the Whoops page might be not good.
https://codeigniter.com/user_guide/installation/troubleshooting.html?highlight=whoops#what-s-with-an-unhelpful-whoops-page
It seems that it suggests to change to development mode in the production server.

app/Views/errors/html/error_404.php
```diff
@@ -70,13 +70,13 @@
...
     <div class="wrap">
-        <h1>404 - File Not Found</h1>
+        <h1>404</h1>
 
         <p>
             <?php if (! empty($message) && $message !== '(null)') : ?>
                 <?= nl2br(esc($message)) ?>
             <?php else : ?>
-                Sorry! Cannot seem to find the page you were looking for.
+                <?= lang('Errors.404sorryCannotFind') ?>
```
kenjis, line 79: Is there any way to display this message?

kenjis (approved):

kenjis:
It seems `lang('Errors.404sorryCannotFind')` is never used.

system/Language/en/Errors.php
```diff
@@ -11,6 +11,8 @@
 
 // Errors language settings
 return [
-    'whoops'     => 'Whoops!',
-    'weHitASnag' => 'We seem to have hit a snag. Please try again later...',
+    '404pageNotFound'    => '404 - Page Not Found',
+    '404sorryCannotFind' => 'Sorry! Cannot seem to find the page you were looking for.',
```
MGatner, line 15: These should be camel case. I know the caps next to the number looks weird but that's the protocol. My recommendation would be to drop the number, or if you really want to keep it then move it to the end. 
```suggestion
    'pageNotFound'    => '404 - Page Not Found',
    'sorryCannotFind' => 'Sorry! Cannot seem to find the page you were looking for.',
```

MGatner (changes requested):

app/Views/errors/html/error_404.php
MGatner: If any falsey string is passed into the PageNotFoundException it would hit this (only explicit `null` uses the default lang message). Also if someone threw the exception directly it would have an empty string for the message: https://3v4l.org/FDkpg

What I want to know is when it would ever be `'(null)'`; that seems oddly specific and also never should happen.

kenjis: `'(null)'` is there:
https://github.com/codeigniter4/CodeIgniter4/blob/0e70ccd62e6e512e454b63aaf15ab7488e508f47/system/Debug/Exceptions.php#L272
But `$exception->getMessage()` never returns null.

MGatner: Yeah that's a mistake. I'm surprised that hasn't been caught by static analysis - very clear not-nullable logic error. I don't think `'(null)'` is ever helpful or needs accounting for, we should remove it.

kenjis: It is suppressed.
```
		-
			message: "#^Expression on left side of \\?\\? is not nullable\\.$#"
			count: 1
			path: system/Debug/Exceptions.php
```

kenjis (approved):

kenjis (approved):

MGatner (approved):

sba:
@MGatner any decicion on this, regarding the layout?

if it should generic as possible I suggest to show the standard 404 from apache http server:
grafik

`<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL was not found on this server.</p>
</body></html>`

MGatner:
@sba maybe "as generic as possible" was too strongly worded. I'm in favor of styling it, I would just rather not have things like the CodeIgniter logo or links to the User Guide. Security by obscurity is very weak and shouldn't be relied on. 

That said I would still like to see this PR split into two and I think @demyr17 has some very good style suggestions for improvements.

MGatner:
I should also add: I always encourage developers to use their own error pages! CodeIgniter makes this easy with the **Views/errors** folder.

PR #6291, opened 2022-07-23
fix: unable to add more than one file to FileCollection constructor

### LINKED ISSUE #6290: Bug: unable to add more than one file to FileCollection constructor
### PHP Version

7.3, 7.4, 8.0, 8.1

### CodeIgniter4 Version

4.2.1

macOS, Linux

### Which server did you use?

fpm-fcgi

### Database

MariaDB 10.8

### What happened?

When I'm trying to create `FileCollection` object from 2 or more files paths (not directories), passed in the `__constructor`, only first file is added. Because in `FileCollection->add()` method uses `return $this->addFile($path);` instead of `$this->addFile($path);` that ends the execution of the `foreach` loop

### Steps to Reproduce

`$object = new FileCollection(['/path/to/file', '/path/to/other-file']);`

### Expected Output

An object containing all passed files

### Anything else?

_No response_

kenjis:
**Description**
Fixes #6290

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Files/FileCollection.php
+++ b/system/Files/FileCollection.php
@@ -178,7 +178,9 @@ class FileCollection implements Countable, IteratorAggregate
                 // Test for a directory
                 self::resolveDirectory($path);
             } catch (FileException $e) {
-                return $this->addFile($path);
+                $this->addFile($path);
+
+                continue;
             }
 
             $this->addDirectory($path, $recursive);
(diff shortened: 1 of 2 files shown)


samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #6292, opened 2022-07-23
fix: Security::derandomize() may cause hex2bin() error

kenjis:
**Description**
From https://forum.codeigniter.com/showthread.php?tid=82544

`Security::derandomize()` may throw `ErrorException` "hex2bin(): Hexadecimal input string must have an even length".

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


MGatner (approved):
Looks great! Thanks for the quick fix @kenjis

PR #6293, opened 2022-07-24
refactor: loading app/Config/routes.php

kenjis:
**Description**
The current way is difficult to use.

- add `RouteCollection::loadRoutes()`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -240,6 +240,62 @@ class RouteCollection implements RouteCollectionInterface
     }
 
     /**
+     * Loads main routes file and discover routes.
+     *
+     * Loads only once unless reset.
+     *
+     * @return $this
+     */
+    public function loadRoutes(string $routesFile = APPPATH . 'Config/Routes.php')
+    {
+        if ($this->didDiscover) {
+            return $this;
+        }
+
+        $routes = $this;
+        require $routesFile;
+
+        $this->discoverRoutes();
+
+        return $this;
+    }
+
+    /**
+     * Will attempt to discover any additional routes, either through
+     * the local PSR4 namespaces, or through selected Composer packages.
+     */
+    protected function discoverRoutes()
+    {
+        if ($this->didDiscover) {
+            return;
+        }
+
+        // We need this var in local scope
+        // so route files can access it.
+        $routes = $this;
+
+        if ($this->moduleConfig->shouldDiscover('routes')) {
+            $files = $this->fileLocator->search('Config/Routes.php');
+
+            $excludes = [
+                APPPATH . 'Config' . DIRECTORY_SEPARATOR . 'Routes.php',
+                SYSTEMPATH . 'Config' . DIRECTORY_SEPARATOR . 'Routes.php',
+            ];
+
+            foreach ($files as $file) {
+                // Don't include our main file again...
+                if (in_array($file, $excludes, true)) {
+                    continue;
+                }
+
+                include $file;
+            }
+        }
+
+        $this->didDiscover = true;
+    }
+
+    /**
      * Registers a new constraint with the system. Constraints are used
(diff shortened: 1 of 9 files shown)


system/Router/RouteCollection.php
```diff
@@ -1393,6 +1393,7 @@ public function resetRoutes()
         }
 
         $this->prioritizeDetected = false;
+        $this->didDiscover        = false;
```
MGatner, line 1396: This seems like it could be a breaking change, though I'm not sure why this wasn't the case before.

system/Router/RouteCollection.php
```diff
@@ -1490,4 +1491,21 @@ public function shouldUseSupportedLocalesOnly(): bool
     {
         return $this->useSupportedLocalesOnly;
     }
+
+    /**
+     * Loads main routes file and discover routes.
+     *
+     * Loads only once unless reset.
+     */
+    public function loadRoutes(string $routesFile = APPPATH . 'Config/Routes.php'): void
```
MGatner, line 1500: All of the calls you replaced look like this:
```php
    $routes = Services::routes();
    $routes->loadRoutes();
```
Which makes me think this would be better off returning `$this` so those can become: `$routes = Services::routes()->loadRoutes()`.

system/Router/RouteCollection.php
```diff
@@ -1490,4 +1491,21 @@ public function shouldUseSupportedLocalesOnly(): bool
     {
         return $this->useSupportedLocalesOnly;
     }
+
+    /**
```
MGatner, line 1495: Unless you have a compelling reason for sticking this at the bottom I think it would make more organizational sense to be up above `discoverRoutes()`.

MGatner (changes requested):
This is great! I've always found this clunky, and it makes static analysis on external packages very ornery to pull a magic `$routes` variable out of nowhere. Good improvement, a few suggestions in review.

system/Router/RouteCollection.php
kenjis: Since `loadRoutes()` is essentially an initialization process, it is better to group it with `Services::routes()->loadRoutes()`.

Actually, I would like to call `loadRoutes()` in the constructor, but we can't because `Routes.php` has `$routes = Services::routes();`.

system/Router/RouteCollection.php
kenjis: Done.

MGatner (commented):
Given that **bootstrap.php** does the initial load already, does it make sense to move `loadRoutes()` into the Service so the returned version is always "loaded"?

kenjis:
Infinite loop occurs unless we remove: https://github.com/codeigniter4/CodeIgniter4/blob/c1bac7e30e46bf84dde9d0d68f468be124d2eac7/app/Config/Routes.php#L6

MGatner:
Gotcha. I think how you had it is good. I don't know why anyone would want the collection without loading the routes but it's there as an option - maybe would help with testing or something.

MGatner (approved):

kenjis:
If we remove `$routes = Services::routes();` in Routes.php, we can load routes in the constructor.
All existing users must remove it, though.

MGatner:
What are the consequences if they don't? Endless loop? If that's the case it seems too risky a change. This is already a huge improvement - I am content taking this from a 2 to a 9 without having to reach 10, which can happen with version 5.

kenjis:
Yes, endless loop. Services::routes() calls Services::routes() and ...
Okay, let's wait for v5.

Since no one objects, I will add changelog to complete this PR.

kenjis:
Added changelog.

MGatner (approved):
Splendid!

MGatner:
🥳 Very excited to get this one in

Issue #5674, opened 2022-02-10
Bug: [QueryBuilder] setInsertBatch() + insertBatch()


iRedds:
### PHP Version

7.4

### CodeIgniter4 Version

dev

### CodeIgniter4 Installation Method

Git

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

apache

### What happened?

When passing data to the setInsertBatch() and insertBatch() methods.
1. Ignoring batchSize.
2. Duplicate field(s) in the query string.

### Steps to Reproduce

```php
        $db = db_connect();

        $set = [
            ['value' => 1,],
            ['value' => 2,],
            ['value' => 3,],
        ];

        $data = [
            ['value' => 4,],
            ['value' => 5,],
            ['value' => 6,],
        ];

        dd($db->table('test')->testMode()->setInsertBatch($set)->insertBatch($data, null, 1));
```
```
array (3) [
    0 => string (66) "INSERT INTO `db_test` (`value`, `value`) VALUES (1), (2), (3), (4)"
    1 => string (42) "INSERT INTO `db_test` (`value`) VALUES (5)"
    2 => string (42) "INSERT INTO `db_test` (`value`) VALUES (6)"
]
```

### Expected Output

```
array (3) [
    0 => string (42) "INSERT INTO `db_test` (`value`) VALUES (1)"
    1 => string (42) "INSERT INTO `db_test` (`value`) VALUES (2)"
    2 => string (42) "INSERT INTO `db_test` (`value`) VALUES (3)"
    3 => string (42) "INSERT INTO `db_test` (`value`) VALUES (4)"
    4 => string (42) "INSERT INTO `db_test` (`value`) VALUES (5)"
    5 => string (42) "INSERT INTO `db_test` (`value`) VALUES (6)"
]
```

kenjis:
What about updateBatch()?

kenjis:
Personally, I think it's enough if the following works:

```php
$db->table('test')->testMode()->setInsertBatch($data)->insertBatch(null, null, 1)
```
```php
$db->table('test')->testMode()->insertBatch($data, null, 1)
```

iRedds:
> What about updateBatch()?

```php
$db->table('test')->testMode()->setUpdateBatch($set)->updateBatch($data, 'value', 1);
$db->table('test')->testMode()->setUpdateBatch($set)->updateBatch(null, 'value', 1);

//throws
[...]

Issue #5134, opened 2021-09-26
[QueryBuilder] The combination of the `setUpdateBatch()` and the `updateBatch()` is difficult to use.


ytetsuro:
The combination of the `setUpdateBatch` method and the `updateBatch` method is difficult to use.

If you try to use it, it will look like the following.

```php
$builder->setUpdateBatch([
    [
        'id'          => 2,
        'name'        => 'SUBSTRING(name, 1)',
        'description' => 'SUBSTRING(description, 3)',
    ],
    [
        'id'          => 3,
        'name'        => 'SUBSTRING(name, 2)',
        'description' => 'SUBSTRING(description, 4)',
    ],
], 'id', false);

$builder->updateBatch(null, 'id', 100);
```

1. updateBatch method hasn't `escape` argument.
2. We have specified index as the second argument of the setUpdateBatch method, but we need to specify the same value for the second argument of `updateBatch`.
3. In combination with 2, when using setUpdateBatch, you need to specify null as the first argument of the updateBatch method. It is difficult for a client to write such a code without knowing the implementation.

In my opinion, it would be easier for developers to use if they can write the following.

```php
$builder->updateBatch([
    [
        'id'          => 2,
        'name'        => 'SUBSTRING(name, 1)',
        'description' => 'SUBSTRING(description, 3)',
    ],
    [
        'id'          => 3,
        'name'        => 'SUBSTRING(name, 2)',
        'description' => 'SUBSTRING(description, 4)',
    ],
], 'id', false, 100);
```

or 

```php
 $builder->setUpdateBatch([
    [
        'id'          => 2,
        'name'        => 'SUBSTRING(name, 1)',
        'description' => 'SUBSTRING(description, 3)',
    ],
    [
        'id'          => 3,
        'name'        => 'SUBSTRING(name, 2)',
        'description' => 'SUBSTRING(description, 4)',
    ],
], 'id', false);

$builder->updateBatch(100);
```

What do you think?

kenjis:
In both cases, it is a breaking change in the public API of updateBatch().

ytetsuro:
You are right.

ytetsuro:
@kenjis 

Thanks, response.

> If you change it to accept int, it is breaking change.

Sure!
I'm sorry.😞

> You could write code like this on PHP8:

Oh! You are great!
This is very readable.

> Or how about this?

Thanks.
I think very very well.😍
I will make this PR.

PR #6294, opened 2022-07-24
Add Upsert/UpsertBatch to builder

sclubricants:
This PR adds Upsert and UpsertBatch methods to builder.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


sclubricants:
The duplicate code can be resolved by refactoring.

The insertBatch() method can point to the new executeBatch() method.

The setInsertBatch() method can point to the new setBatch() method.

Also a small adjustment to _insertBatch() method will be to use the new getValues() method.

I didn't want to change them yet so you guys could see what is going on here before the change.

system/Database/BaseBuilder.php
```diff
@@ -1719,6 +1726,238 @@ public function getWhere($where = null, ?int $limit = null, ?int $offset = 0, bo
...
     }
 
+    /**
+     * Compiles batch insert/upsert strings and runs the queries
+     *
+     * @throws DatabaseException
+     *
+     * @return false|int|string[] Number of rows inserted or FALSE on failure, SQL array when testMode
+     */
+    public function batchExecute(string $renderMethod, ?array $set = null, ?bool $escape = null, int $batchSize = 100)
```
iRedds, line 1736: I don't think this method should be public.

iRedds (changes requested):
1. CI4 follows PSR-12 so you need to limit lines to 80 (120) characters (https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/styleguide.md#lines).
2. I recommend wrapping SQL code in tests in Nowdoc (or Heredoc). This will make it easier to read. It also seems to me that it is worth separating the SQL generation tests into separate methods for each driver.
3. It is better to separate life tests and SQL generation tests, according to the same principle as it is now in the repository. 
3. Need tests for exceptions thrown in ``batchExecute()``.
4. I think the _upsertBatch() method of the SQLite and Postgre drivers should throw an exception if the ``$constraints`` variable is empty. Otherwise, the method becomes a duplicate of insertBatch()

MGatner:
Very nice work on this feature, much easier to review after splitting apart. @iRedds gave a great review so I will refrain, but I'm here watching.

sclubricants:
> 1. CI4 follows PSR-12 so you need to limit lines to 80 (120) characters
> 2. I recommend wrapping SQL code in tests in Nowdoc (or Heredoc). This will make it easier to read. It also seems to me that it is worth separating the SQL generation tests into separate methods for each driver.
> 3. It is better to separate life tests and SQL generation tests, according to the same principle as it is now in the repository.
> 4. Need tests for exceptions thrown in `batchExecute()`.
> 5. I think the _upsertBatch() method of the SQLite and Postgre drivers should throw an exception if the `$constraints` variable is empty. Otherwise, the method becomes a duplicate of insertBatch()

@iRedds 
1) Done
2) Nowdoc Done - I don't see separating all that beneficial. I feel this would add unnecessary duplication. You know which one fails by which test fails. 
3) I'm not sure what your talking about here.  There are some tests at tests\system\Database\Builder but these tests use mock connection and do not test all the databases. Also, I'm testing the method `getCompiledUpsert()` and not explicitly running `testMode()`. 
4) I have now refactored the `insertBatch()` method to use `executeBatch()`. The existing tests should provide code coverage now. 
5) Done - I was back and forth as to this matter. You settled it. The same could be said for Mysqli but then you would force an extra query on all the calls to check if a constraint exists for the table and the included fields.

iRedds (commented):
In the context of individual SQL code generation tests, see how the class is implemented https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/Live/OCI8/LastInsertIDTest.php
SQL for Postgre and SQLite is the same. SQL for Oracle and MSSQL are similar for the most part. Therefore, it is possible to confuse what exactly caused the error.
I think we need to hear the opinion of the maintainers on this issue.

MySQL does not depend on indexes (constraint) in the same way as other DBMSs for upsert queries. Honestly, I don’t even know if it makes sense in other DBMSs to perform search operations for intexes with an empty $constraints variable.

You need to add a description to the documentation and changelog.

sclubricants:
If the test "**PHPUnit / PHP 8.1 - Postgre (pull_request)**" fails the test then you know that the issue is with Postgre. Or if you are running locally whichever database assigned as "tests". 

The live/driver directory is for tests which have no counterpart in the other drivers. In this case the same call returns a value for all of them.

>I don’t even know if it makes sense in other DBMSs to perform search operations for intexes with an empty $constraints variable.

My original conception was to have all the DBMSs work in the same way as MySQL. However some of them don't allow using multiple constraints at one time. So I changed to pick up the first constraint in order.. primary keys first and then unique indexes. Originally I didn't have the option to manually set constraints but felt it was a must so I added this feature. For most purposes the auto method of assigning a constraint works quite well though and simplifies writing code.

sclubricants:
A question for you guys:

I created the method `setBatch()` which is a copy of `setInsertBatch()`. The reasoning was to disassociate the method from insert to be used with upsert or both insert and upsert. The way the codes sits now `setInsertBatch()` is an alias of `setBatch()`.

So currently you could do:

```php
$this->db->table('user')->setBatch($userData)->upsertBatch();
```
The question is, we could forgo the `setBatch()` method and just use the `setInsertBatch() ` method with upserts.

```php
$this->db->table('user')->setInsertBatch($userData)->upsertBatch();
```
What do you guys think?

sclubricants:
Here is a sneak peak at my next installment.. should you guys accept this PR.

```php
        // upsert from query rather than data set
        $fromQuery = $db->table('user_loggin')
            ->select('users.id, MAX(user_loggin.login_date) AS last_login')
            ->join('users','users.id = user_loggin.userid','inner')
            ->groupBy('users.id');
            
        $builder = $db->table('users')
            ->fromQuery($fromQuery)
            ->onConstraint('id')
            ->updateFields('last_login')
            ->upsert();

        // same can work with inserts
        $builder = $db->table('table_name')
            ->fromQuery($fromQuery)
            ->ignore(true)
            ->insert();
            
        // can work with raw SQL too
        $fromQuery = <<<'SQL'
                SELECT users.id, MAX(user_loggin.login_date) AS last_login
                FROM user_login
                INNER JOIN users ON users.id = user_login.userid
                GROUP BY users.id
            SQL;
        
        $builder = $db->table('users')
            ->fromQuery($fromQuery)
            ->onConstraint('id')
            ->updateFields('last_login')
            ->upsert();
```

iRedds:
> What do you guys think?

To be honest, the ``setBatch()`` method has a very general name structure. If you look at the surface and do not refer to the documentation, then you can assume that this method can be used for any batch changes to the database.
That is, a developer may natively think that the following code will work:

```php
$builder->setBatch($data)->updateBatch();
```
I think it would be more logical to see the method name as ``setUpsertBatch()``.

At the same time, the ``setInsertBatch()`` method explicitly defines the behavior of batch insertion. 
Which can also lead to misunderstanding.

And if I had to choose between the options you suggested, I would choose the third one.
```php
$builder->set(data); // ->insert()/update()/upsert() Both for single insert/update and batch.
```

sclubricants:
The set method already works. It doesnt work for multiple records.
I agree with setUpsertBatch(). I guess I was thinking to consolidate them but that would break things.

sclubricants:
@lonnieezell @paulbalandan @MGatner @michalsn @samsonasik @kenjis 
Any comments on this PR?

system/Database/BaseBuilder.php
```diff
@@ -1719,6 +1726,252 @@ public function getWhere($where = null, ?int $limit = null, ?int $offset = 0, bo
...
+
+        $this->resetWrite();
+
+        return $this->testMode ? $savedSQL : $affectedRows;
+    }
+
+    /**
+     * Allows key/value pairs to be set for batch inserts/upserts
+     *
+     * @param mixed $key
```
kenjis, line 1797: Please don't use `mixed`.
See #6310

system/Database/BaseBuilder.php
```diff
@@ -1719,6 +1726,252 @@ public function getWhere($where = null, ?int $limit = null, ?int $offset = 0, bo
...
+        // this doesn't work with current implimentation - is cleared in upsert method
+        if ($reset === true) {
+            $this->resetWrite();
+        }
+
+        return $this->compileFinalQuery($sql);
+    }
+
+    /**
+     * Converts call to batchUpsert
```
kenjis, line 1864: What do you mean?

system/Database/BaseBuilder.php
```diff
@@ -1841,6 +1840,166 @@ public function setInsertBatch($key, string $value = '', ?bool $escape = null)
         return $this;
     }
 
+    /**
+     * Compiles an upsert query and returns the sql
+     *
+     * @throws DatabaseException
+     *
+     * @return bool|string
```
kenjis, line 1848: When does it return bool?

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -41,6 +41,7 @@ Enhancements
 - SQLite ``BaseConnection::getIndexData()`` now can return pseudo index named ``PRIMARY`` for `AUTOINCREMENT` column, and each returned index data has ``type`` property.
 - Added ``spark filter:check`` command to check the filters for a route. See :ref:`Controller Filters <spark-filter-check>` for the details.
 - Now **Encryption** can decrypt data encrypted with CI3's Encryption. See :ref:`encryption-compatible-with-ci3`.
+- Added ``upsert()`` and ``upsertBatch()`` methods to builder. Also included for use with upserts are ``onConstraint()`` and ``updateFields()`` methods.
```
kenjis, line 44:
```suggestion
- Added ``upsert()`` and ``upsertBatch()`` methods to QueryBuilder. Also included for use with upserts are ``onConstraint()`` and ``updateFields()`` methods.
```

user_guide_src/source/database/query_builder.rst
```diff
@@ -851,6 +851,82 @@ The first parameter is an associative array of values.
...
+Upsert
+======
+
+$builder->upsert()
+------------------
+
+Generates an upsert string based on the data you supply, and runs the
+query. You can either pass an **array** or an **object** to the
+method. By default a constraint will be defined in order. A primary
+key will be selected first and then unique keys. Mysql will use any
```
kenjis, line 867:
```suggestion
key will be selected first and then unique keys. MySQL will use any
```

user_guide_src/source/database/query_builder/107.php
```diff
@@ -0,0 +1,18 @@
...
+
+$data = [
+    'email'   => 'ahmadinejad@world.com',
+    'name'    => 'Ahmadinejad',
+    'country' => 'Iran',
+];
+
+$sql = $builder->set($data)->getCompiledUpsert();
+echo $sql;
+/* MySQLi  produces:
```
kenjis, line 11:
```suggestion
/* MySQLi produces:
```

user_guide_src/source/database/query_builder/108.php
```diff
@@ -0,0 +1,23 @@
...
+    ],
+    [
+        'id'      => null,
+        'email'   => 'pedro@world.com',
+        'name'    => 'Pedro',
+        'country' => 'El Salvador',
+    ],
+];
+
+$builder->insertBatch($data);
```
kenjis, line 18:
```suggestion
$builder->upsertBatch($data);
```

user_guide_src/source/database/query_builder/109.php
```diff
@@ -0,0 +1,19 @@
...
+
+$data = [
+    'id'      => 2,
+    'email'   => 'ahmadinejad@world.com',
+    'name'    => 'Ahmadinejad',
+    'country' => 'Iran',
+];
+
+$builder->onConstraint('email')->upsert($data);
+/* Postgre  produces:
```
kenjis, line 11:
```suggestion
/* Postgre produces:
```

user_guide_src/source/database/query_builder/110.php
```diff
@@ -0,0 +1,22 @@
...
+
+$data = [
+    'id'      => 2,
+    'email'   => 'ahmadinejad@world.com',
+    'name'    => 'Ahmadinejad Zaghari',
+    'country' => 'Afghanistan',
+];
+
+$builder->updateFields('name, country')->upsert($data);
+/* SQLSRV  produces:
```
kenjis, line 11:
```suggestion
/* SQLSRV produces:
```

tests/system/Database/Live/UpsertTest.php
```diff
@@ -0,0 +1,393 @@
...
+
+            $new = $this->db->table('user')
+                ->getwhere(['email' => 'ahmadinejad@world.com'])
+                ->getRow();
+
+            $this->assertSame(5, (int) $new->id);
+        }
+    }
+
+    public function testGetCompiledUpsert()
```
kenjis, line 98: This test method is very long, and it will be longer if another DB driver is added.

I prefer to split into DB driver specific test case classes like
https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/Live/OCI8/LastInsertIDTest.php

system/Database/BaseBuilder.php
```diff
@@ -1719,6 +1726,252 @@ public function getWhere($where = null, ?int $limit = null, ?int $offset = 0, bo
...
+
+        $escape = is_bool($escape) ? $escape : $this->db->protectIdentifiers;
+
+        $keys = array_keys($this->objectToArray(current($key)));
+        sort($keys);
+
+        foreach ($key as $row) {
+            $row = $this->objectToArray($row);
+            if (array_diff($keys, array_keys($row)) !== [] || array_diff(array_keys($row), $keys) !== []) {
+                // batch function above returns an error on an empty array
```
kenjis, line 1817: What is the batch function above?

system/Database/BaseBuilder.php
```diff
@@ -153,6 +153,13 @@ class BaseBuilder
      */
     protected $QBIgnore = false;
 
+    /**
+     * QB Options data
```
kenjis, line 157: What is the QB Options data?
More explanation is needed.

system/Database/BaseBuilder.php
```diff
@@ -153,6 +153,13 @@ class BaseBuilder
...
+     *
+     * @var array
```
kenjis, line 159: It seems something like this. Can you add type for PHPStan like this?
```
@phpstan-var array{updateFields?: array, constraints?: array, tableIdentity?: string}

File: system/Database/BaseBuilder.php [hunk 15]
```diff
@@ -153,6 +153,13 @@ class BaseBuilder
...
+     */
+    protected $QBOptions;
```
kenjis, line 161: Why is the property named `QBOptions`, not `QBUpsert`?

kenjis:
Ref, #5134 and #5674

system/Database/BaseBuilder.php
sclubricants: Meaning it changes the data set from array to an array of arrays. Then it handles things as a batchUpsert of one row. 
$set = [$set};

system/Database/BaseBuilder.php
sclubricants: It doesn't, changed this.

tests/system/Database/Live/UpsertTest.php
sclubricants: From the developer perspective its nicer to have all the upsert tests in one place. This way when working on the code you only have to run one test file. 

This pattern is used many places:
tests/system/Database/Live/ForgeTest.php line 174
tests/system/Database/Live/ForgeTest.php line 434
tests/system/Database/Live/ForgeTest.php line 497
tests/system/Database/Live/ForgeTest.php line 825
tests/system/Database/Live/ForgeTest.php line 908

I have reorganized the code for this method so its very easy to read.

It is also nice to see them all together so you can see the differences. This makes refactoring easier.

I really find this method the best, but if you still find this unacceptable I will put them in separate files.

system/Database/BaseBuilder.php
sclubricants: I copied all of this from setInsertBatch(). I changed: // batchExecute() function returns an error on an empty array

system/Database/BaseBuilder.php
sclubricants: I had contemplated adding a parameter to the $renderMethod functions in executeBatch(). This would add $options parameter to _insertBatch() and _upsertBatch(). However I found it easier just to access the property from those methods rather than pass it to them. 

Additionally I want to add the method fromQuery() as I demonstrated in the comments here. I would use the $QBOptions to hold this query would could be used for both inserts and upserts. So I didn't want to associate it with upsert. 

I could of created unique properties for all these things but I felt it would clutter the code unnecessarily.

tests/system/Database/Live/UpsertTest.php
```diff
@@ -0,0 +1,391 @@
...
+                    ON DUPLICATE KEY UPDATE
+                    `country` = VALUES(`country`),
+                    `email` = VALUES(`email`),
+                    `name` = VALUES(`name`)
+                    SQL;
+                break;
+
+            case 'Postgre':
+                $expected = <<<'SQL'
+                    INSERT INTO "db_user"("country", "email", "name")
```
kenjis, line 115: There should be a space.

```suggestion
                    INSERT INTO "db_user" ("country", "email", "name")
```

tests/system/Database/Live/UpsertTest.php
```diff
@@ -0,0 +1,391 @@
...
+                    DO UPDATE SET
+                    "country" = "excluded"."country",
+                    "email" = "excluded"."email",
+                    "name" = "excluded"."name"
+                    SQL;
+                break;
+
+            case 'SQLite3':
+                $expected = <<<'SQL'
+                    INSERT INTO `db_user`(`country`, `email`, `name`)
```
kenjis, line 127:
```suggestion
                    INSERT INTO `db_user` (`country`, `email`, `name`)
```

tests/system/Database/Live/UpsertTest.php
kenjis: Okay, you refactored this method, and it is easier to read than before.
Using `switch` is better than a lot of `if` like in ForgeTest.

Personally the ForgeTest is really difficult to read, so I would like to refactor.

system/Database/BaseBuilder.php
sclubricants: insert() returns Query Object
insertBatch() returns Number of rows inserted or FALSE on failure, SQL array when testMode
upsert() and upsertBatch() return Number of rows inserted or FALSE on failure, SQL array when testMode

kenjis: > insert() returns Query Object

No, `insert()` returns `bool`.

kenjis: > upsert() and upsertBatch() return Number of rows inserted or FALSE on failure, SQL array when testMode

Number of rows inserted? Not include updated rows?

sclubricants: This gets messy with upserts. The data returned at least with myself is almost useless.

Per mysql:
>With ON DUPLICATE KEY UPDATE, the affected-rows value per row is 1 if the row is inserted as a new row and 2 if an existing row is updated.

And if not inserted or updated then 0.

kenjis: That is the return value of `upsertBatch()` with MySQL is not *Number of rows inserted*.

> With ON DUPLICATE KEY UPDATE, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values. If you specify the CLIENT_FOUND_ROWS flag to the [mysql_real_connect()](https://dev.mysql.com/doc/c-api/8.0/en/mysql-real-connect.html) C API function when connecting to [mysqld](https://dev.mysql.com/doc/refman/8.0/en/mysqld.html), the affected-rows value is 1 (not 0) if an existing row is set to its current values.
https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html

system/Database/BaseBuilder.php
```diff
@@ -1841,6 +1843,166 @@ public function setInsertBatch($key, string $value = '', ?bool $escape = null)
...
+    }
+
+    /**
+     * Converts call to batchUpsert
+     *
+     * @param array|object|null $set
+     *
+     * @throws DatabaseException
+     *
+     * @return false|int|string[] Number of rows replaced or FALSE on failure, SQL array when testMode
```
kenjis, line 1873:
```suggestion
     * @return false|int|string[] Number of affected rows or FALSE on failure, SQL array when testMode
```

system/Database/BaseBuilder.php
```diff
@@ -1841,6 +1843,166 @@ public function setInsertBatch($key, string $value = '', ?bool $escape = null)
...
+
+        return $this->batchExecute('_upsertBatch', $set, $escape, 1);
+    }
+
+    /**
+     * Compiles batch upsert strings and runs the queries
+     *
+     * @throws DatabaseException
+     *
+     * @return false|int|string[] Number of rows replaced or FALSE on failure, SQL array when testMode
```
kenjis, line 1898:
```suggestion
     * @return false|int|string[] Number of affected rows or FALSE on failure, SQL array when testMode
```

kenjis:
`upsert()` is a write type query. So should it return `bool` for consistancy?

system/Database/BaseBuilder.php
```diff
@@ -1785,21 +1795,13 @@ public function insertBatch(?array $set = null, ?bool $escape = null, int $batch
...
-     * Allows key/value pairs to be set for batch inserts
+     * Allows key/value pairs to be set for batch inserts/upserts
      *
-     * @param mixed $key
+     * @param array|object $key
      *
      * @return $this|null
      */
-    public function setInsertBatch($key, string $value = '', ?bool $escape = null)
+    public function setBatch($key, string $value = '', ?bool $escape = null)
```
kenjis, line 1804: Do you really need `$value`? For what?

kenjis:
There will be `setBatch()`, `setInsertBatch()` and `setUpdateBatch()`.
These are not consistent.

If there is a `setBatch()` method, I think it can be used for all batch methods.
But it is only used for inserts/upserts.

system/Database/BaseBuilder.php
sclubricants: The setInsertBatch method accepts key value pairs if $key is a string.
```php
if (! is_array($key)) {
            $key = [$key => $value];
        }
```
 Im really not sure why being a batch operation though. Im looking at a way to consolidate all the setBatch methods and this presents some issues. The setUpdateBatch() method accepts $index as a second parameter but it doesnt do much of anything with it.

kenjis: See #6315
The current batch APIs are broken or not consistent.
So I don't think it is necessary to match what is broken.

I don't know why one key/value string pair is needed for batch operation, and even if you need to set one record,
you can set as `[[key => value]]`.

> The setUpdateBatch() method accepts $index as a second parameter but it doesnt do much of anything with it.

It may indicate to consolidate all the setBatch methods is not reasonable.

I think the following is enough:
```php
setUpsertBatch($set, $escape)
```
And make the `$value` of other batch methods deprecated, and will remove in v5.
What do you think?

sclubricants: @kenjis 
I spent a lot of time looking at the code this past weekend and thinking how this should all work out. There are definitely some inconsistencies. I'd like to refactor everything.. perhaps after this PR is committed. Basically I think we should start from the ground up with the `setBatch()` method and make it work for `insertBatch()`, `upsertBatch()`, `updateBatch()`, and a possible new one `deleteBatch()`. Then the old `setUpdateBatch()`, `setInsertBatch()`, etc we can keep functioning as is and make them depreciated. 

One thing I noticed is that `setUpdateBatch()` and `updateBatch()` seems to only allow for one field to be used as $index. So if you have a dataset with composite primary keys then you couldn't update them in a batch. The update() method allows using a where() clause and so It would work for multiple keys. 

The `updateBatch()` SQL for MySQL is pretty ugly. I have a different method for writing SQL for batch updates. I'll run some performance tests to see how it performs against the existing methods.

```SQL
UPDATE table_name
INNER JOIN (
SELECT 1 primary1, 1 primary2, 10 column3, 20 column4, 30 column5 UNION ALL
SELECT 1 primary1, 2 primary2, 40 column3, 50 column4, 60 column5
) updates
  ON table_name.primary1= updates.primary1 # these are your $constraints
  AND table_name.primary2= updates.primary2
SET
  table_name.column3 = updates.column3,
  table_name.column4= updates.column4,
  table_name.column5= updates.column5; 
```
This same query design could be used for `deleteBatch()` where you have multiple keys you are working with.

And then to be consistent with upserts I think that the `onConstraint()` method may be used to set the fields to update on. 

for the above query a statement might look like:

```php
$builder->setBatch($data)->onConstraint('primary1, primary2')->updateBatch();
// or 
$builder->setBatch($data, 'primary1, primary2', 50)->updateBatch();
// or 
$builder->updateBatch($data, 'primary1, primary2', 50);
// or 
$builder->setBatch($data)->onConstraint('primary1, primary2')->updateBatch(null,null,50);
```
Basically I was thinking that the `setBatch()` method signature should should align with the `xBatch()` methods. The set method basically is just used when you want to set the data but not execute the query yet. Perhaps you use the `setBatch()` method in a loop collecting rows.

So: 
```php
$builder->setBatch($set, $constraint = null, $batchsize = 100, $escape = true);
// and
$builder->updateBatch($set = null, $constraint = null, $batchsize = 100, $escape);
// and
$builder->insertBatch($set = null, $batchsize = 100, $escape);
// and
$builder->upsertBatch($set = null, $constraint = null, $batchsize = 100, $escape);
// and
$builder->deleteBatch($set = null, $constraint = null, $batchsize = 100);
// OR reorder them so $constraint is last being its not used for inserts
$builder->setBatch($set, $batchsize = 100, $escape = true, $constraint = null);
// OR remove paramater and force to use other methods to set them.
$builder->setBatch($set, $escape = true)->onConstraint($constraint)->setBatchSize($batchSize)->updateBatch();
```
This would work for insert,update,upsert,delete Batch. When using with `insertBatch()` the constraint parameter would be unused.

Alternatively we could drop the $constraint parameter for force you to use `onConstraint()` method.

Similarly with $batchsize. The reason to include the `setBatch()` is keep the last method clean: `->updateBatch()`;

Then I would like to commit all the $set data to $QBSet. This can be done in methods such as `insertBatch()`, etc. This way the `executeBatch()` method can be called with out a $set parameter and there doesn't need to be dual logic for $set and $QBSet. The way it currently is if $set data is used instead of $QBSet then the data isn't cleaned, escaped, etc. Also I like to see if $binds could be used to the batch data as well... currently it is not. 

As it stands, the `setBatch()` method is pretty much just a direct copy of what the `setInsertBatch()` method used to be.

So in conclusion, for now we can design this PR around the `setBatch()` method, keeping in mind for future use with the other DML statements. If we commit this now, we still expect to revise and refactor before final release. To say exactly what parameters we should use now may not be the best time. I think doing some refactoring of the code and then having a look at things then when they are all aligned together might make for a better conversation on what exactly the public API should look like.

system/Database/BaseBuilder.php
sclubricants: I added to the tests to assert return values for the different DBMS. I included a default value which would show the ideal results - though none of them actually return this. I highlighted the unique behaviors of each DBMS and how they differ from the ideal.

sclubricants: I fixed _upsertBatch() where it throws an error if no constraints are found. I added  if ($this->db->DBDebug)  throw error else return false. This way if this error is thrown the executeBatch() can return false on error.

sclubricants: added testNoConstraintFound(). Can't test for return false without a setter.
$this->db->DBDebug = false; (protected property) 
Used return ''; // @codeCoverageIgnore

system/Database/BaseBuilder.php
sclubricants: I changed to `setBatch($key, ?bool $escape = null)`. We still can come back and take another look later.

kenjis: > Also I like to see if $binds could be used to the batch data as well... currently it is not.

We do not use $binds with batch operations intentionally. It significantly degrades performance.
See https://github.com/codeigniter4/CodeIgniter4/pull/5202#issuecomment-944037647

system/Database/BaseBuilder.php
```diff
@@ -1785,28 +1799,16 @@ public function insertBatch(?array $set = null, ?bool $escape = null, int $batch
...
-    {
-        return 'INSERT ' . $this->compileIgnore('insert') . 'INTO ' . $table . ' (' . implode(', ', $keys) . ') VALUES ' . implode(', ', $values);
-    }
-
-    /**
-     * Allows key/value pairs to be set for batch inserts
+     * Allows key/value pairs to be set for batch inserts/upserts
      *
-     * @param mixed $key
+     * @param array|object $key
```
kenjis, line 1804: The name `$key` is not good. It is not a key.

tests/system/Database/Live/UpsertTest.php
```diff
@@ -0,0 +1,484 @@
...
+        $userData = [
+            'id'      => 5,
+            'email'   => 'ahmadinejad@world.com',
+            'name'    => 'Ahmadinejad',
+            'country' => 'Iran',
+        ];
+
+        // some databases don't allow update of auto incrememnt
+        if ($this->db->DBDriver === 'SQLSRV') {
+            $this->assertTrue(true);
```
kenjis, line 130: Use `markTestSkipped()`.

E.g.
```php
$this->markTestSkipped('OCI8 does not support create database.');
```

sclubricants:
@kenjis I screwed this up trying to fix a conflict. Any idea of what to do?

kenjis:
What command did you run?

sclubricants:
I was using github desktop. I ran rebase UpsertUpsertBatch on 4.3

sclubricants:
It was suppose to add like 11 commits on top of 4.3

kenjis:
I've pushed my backup of your branch in my repository:
https://github.com/kenjis/CodeIgniter4/commits/UpsertUpsertBatch

sclubricants:
So how do I use your backup?

kenjis:
I recommend you create a new branch as a backup before running destructive commands:
```
$ git branch UpsertUpsertBatch.bk
```

Fetch my repository:
```
$ git remote add kenjis git@github.com:kenjis/CodeIgniter4.git
$ git fetch kenjis
```

Remove the existing branch:
```
$ git switch develop
$ git branch -D UpsertUpsertBatch
```

Checkout:
```
$ git branch UpsertUpsertBatch kenjis/UpsertUpsertBatch
$ git switch UpsertUpsertBatch
```

sclubricants:
```
$ git fetch kenjis
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

sclubricants:
I have the files. I'll just create a new branch and new PR.

kenjis:
If you don't use SSH, use HTTPS link.

```
$ git remote add kenjis https://github.com/kenjis/CodeIgniter4.git
```

Issue #6310, opened 2022-07-29
Dev: replace type mixed with more specific types


kenjis:
Type `mixed` means `object|resource|array|string|int|float|bool|null`.
https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.mixed

The current code base has a lot of type `mixed` in PHPDocs,
But I think all of them are not correct. 

And `mixed` prevents PHPStan level from increasing. 
> PHPStan’s [rule level 6](https://phpstan.org/user-guide/rule-levels) isn’t satisfied with implicit mixed
https://phpstan.org/writing-php-code/phpdoc-types#mixed

We need to replace type `mixed` with more specific types.

paulbalandan:
I think we should, in part, allow explicit `mixed` types if the parameter/return can be of *any* type and we have no way of telling what can be the possible types at runtime.

However, for those annotated as `mixed` but the actual types can be deduced to be a subset only of `object|resource|array|string|int|float|bool|null` then that is what we should flag as erroneous. The annotation should indicate specifically the actual types, not `mixed`.

kenjis:
@paulbalandan What's you say is correct, but I think it is difficult to do.

Currently, there are too many `mixed`, so I think it would be easier to understand the approach of banning and eliminating `mixed`, and then changing to `mixed` if `mixed` is appropriate.

kenjis:
We still have a lot of `@return mixed` and `@param mixed`.
Contributions are welcome!

ping-yee:
I already sent few PR, review please.

neznaika0:
```
    /**
     * Outputs a string to the cli on it's own line.
     *
     * @return void
     */
    public static function write(string $text = '', ?string $foreground = null, ?string $background = null)
```
Is it really important to use Docbook or do I need to add a clarification `: void` ?
Because now the methods have the only meaning `: void` looks right. But you accept comments where DocBlock

kenjis:
@neznaika0 This issue is about `@mixed` doc comment.

Basically, we cannot change the method signature if the method can be overridden (by a user).
Because it is a breaking change, and we permit breaking changes only in major version up.

kenjis:
```
AbstractLogger.php
     * @param mixed[] $context
     * @param mixed[] $context
     * @param mixed[] $context
     * @param mixed[] $context
     * @param mixed[] $context
     * @param mixed[] $context
     * @param mixed[] $context
     * @param mixed[] $context
[...]

PR #6311, opened 2022-07-29
docs: fix types in PHPDocs in RESTful controllers

kenjis:
**Description**
See #6310

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/RESTful/ResourceController.php
+++ b/system/RESTful/ResourceController.php
@@ -34,7 +34,7 @@ class ResourceController extends BaseResource
     /**
      * Return the properties of a resource object
      *
-     * @param string|null $id
+     * @param int|string|null $id
      *
      * @return Response|string|void
      */
@@ -66,7 +66,7 @@ class ResourceController extends BaseResource
     /**
      * Return the editable properties of a resource object
      *
-     * @param string|null $id
+     * @param int|string|null $id
      *
      * @return Response|string|void
      */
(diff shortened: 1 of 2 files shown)


system/RESTful/ResourceController.php
```diff
@@ -23,7 +24,7 @@ class ResourceController extends BaseResource
     /**
      * Return an array of resource objects, themselves in array format
      *
-     * @return mixed
+     * @return Response
```
MGatner, line 27: All of the RESTful methods are placeholders to be extended (optionally) so even though their current return value is `Response` they could also return `string` or possibly even nothing. If we type these here without including those then developers will need to retype the docblocks or risk breaking static analysis. 

My suggestion is to make all these `Response|string|null`.

system/RESTful/ResourceController.php
```diff
@@ -33,9 +34,9 @@ public function index()
     /**
      * Return the properties of a resource object
      *
-     * @param mixed $id
+     * @param string|null $id
```
MGatner, line 37: I'm always torn on Controller parameters. The way you have it is probably the technically-correct solution, but lots of times resource IDs end up coming in as `int` during testing. This change would require casting test values.

MGatner (changes requested):
Good changes, some points for discussion.

system/RESTful/ResourceController.php
kenjis: Changed to `Response|string|void`.

system/RESTful/ResourceController.php
kenjis: Changed to `int|string|null`.
We should support int type.

MGatner (approved):
🚀

PR #6321, opened 2022-07-31
Feature for Timer to measure callable performance

rumpfc:
**Description**
This is a convinience feature enhancement for `Timer` to measure short code blocks without adding its `start`/`stop` methods before and after the block. The new method `record()` accepts a string and a callable as parameters.

```php
// Example
$benchmark->record('slow_function', function () { slow_function('...'); });
```

Which is the equivalent to 

```php
$benchmark->start('slow_function');
slow_function('...');
$benchmark->stop('slow_function');
```

It can also return the callables return value

```php
$length = $benchmark->record('string length', fn () => strlen('CI4'));
// $length = 3
```

Alternatively the common function `timer()` also accepts an optional callable for the same mechanics. Leaving out the 2nd parameter behaves like before (returning a `Timer` instance).

```php
$length = timer('string length', fn () => strlen('CI4'));
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


system/Common.php
```diff
@@ -1076,19 +1076,25 @@ function stringify_attributes($attributes, bool $js = false): string
...
-    function timer(?string $name = null)
+    function timer(?string $name = null, ?callable $callable = null)
     {
         $timer = Services::timer();
 
         if (empty($name)) {
             return $timer;
         }
 
+        if (null !== $callable) {
```
MGatner, line 1094: I thought we had a CS Fixer rule for this but maybe not. 
```suggestion
        if ($callable !== null) {
```

system/Debug/Timer.php
```diff
@@ -126,4 +126,22 @@ public function has(string $name): bool
...
     }
+
+    /**
+     * Executes callable and measures its time.
+     * Returns its return value if any.
+     *
+     * @param string   $name     The name of the timer
+     * @param callable $callable callable to be executed
+     *
+     * @return mixed|null
```
MGatner, line 137: Since this is a callable return it's fine to leave off the implied `null`. 
```suggestion
     * @return mixed
```

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -41,6 +41,7 @@ Enhancements
 - SQLite ``BaseConnection::getIndexData()`` now can return pseudo index named ``PRIMARY`` for `AUTOINCREMENT` column, and each returned index data has ``type`` property.
 - Added ``spark filter:check`` command to check the filters for a route. See :ref:`Controller Filters <spark-filter-check>` for the details.
 - Now **Encryption** can decrypt data encrypted with CI3's Encryption. See :ref:`encryption-compatible-with-ci3`.
+- Added method ``Timer::record`` to measure performance in a callable. Also enhanced common function ``timer()`` to accept optional callable.
```
MGatner, line 44:
```suggestion
- Added method ``Timer::record()`` to measure performance in a callable. Also enhanced common function ``timer()`` to accept optional callable.
```

MGatner (changes requested):
Very nice PR! I like this new feature, especially the inclusion of `callable` which (in my opinion) allows the very valuable array notation benchmarking, not just closures. Our `timer()` function is starting to get a bit abstruse, but I'm okay with this expansion unless someone else objects. 

Thanks for the contribution @rumpfc! A few notes inline, feel free to discuss.

system/Debug/Timer.php
kenjis: See #6310
We should not add `mixed`.

kenjis:
There are coding style violation. Please fix.

system/Common.php
rumpfc: It used to be `! Is_null($callable)`, but fixer turned it into `null !== $callable`, which looked strange for me, but it was the tools suggestion.

system/Debug/Timer.php
rumpfc: > See #6310 We should not add `mixed`.

The method returns the callable return value, which can be any type. Making this type specific might not be possible.

@kenjis What do you suggest?

kenjis: `object|array|string|int|float|bool|null`

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +123,107 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
         $this->assertNull($timer->getElapsedTime('test1'));
     }
+
+    /**
+     * @timeLimit 1.5
+     */
+    public function testRecordFunctionNoReturn()
+    {
+        $timer       = new Timer();
+        $returnValue = $timer->record('longjohn', static function () { sleep(1); });
```
kenjis, line 133: There are several tests to wait for 1 second. If you really need to test 1 second duration,
write one test method for it, and remove other `sleep(1)` in other methods.

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +123,107 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
+    {
+        $this->expectException(RuntimeException::class);
+
+        $timer = new Timer();
+        $timer->record('ex', static function () { throw new RuntimeException(); });
+    }
+
+    public function testRecordThrowsErrorOnCallableWithParams()
+    {
+        $this->expectException(Throwable::class);
```
kenjis, line 174: The exception should be `ArgumentCountError`, not `Throwable`.

user_guide_src/source/testing/benchmark.rst
```diff
@@ -38,6 +38,20 @@ and stop timers:
 
 .. literalinclude:: benchmark/003.php
 
+If you use very small code blocks to benchmark, you can also use the ``record()`` method. It accepts
```
kenjis, line 41: Including the added version would reduce misunderstandings by users of the older version.

```suggestion
Since v4.3.0, if you use very small code blocks to benchmark, you can also use the ``record()`` method. It accepts
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +123,107 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
+        $this->assertLessThanOrEqual(1.0, $timer->getElapsedTime('longjohn'));
+        $this->assertSame(3, $returnValue);
+    }
+
+    public function testRecordThrowsException()
+    {
+        $this->expectException(RuntimeException::class);
+
+        $timer = new Timer();
+        $timer->record('ex', static function () { throw new RuntimeException(); });
```
paulbalandan, line 165: I have a question. If by mistake I made the callable throw an exception, then how can I get the elapsed time since `$this->stop($name)` will not be called? Should the invocation of the callable be wrapped in a `try-finally` ?

rumpfc: If an exception is thrown and you don't catch it, then I think you might have other problems than worrying about the time stoppage.

But you also have a point. It does not handle the case of purposely catching exceptions for error handling.

```php
try {
    $benchmark->record('test', 'func_that_throws');
} catch (Throwable $e) {
    // stop() isn't called yet
}
```

@kenjis shall we cover this case as well?

kenjis: What if an exception is thrown after `$benchmark->start()` or in `$iterator->run()`?

rumpfc: For `$iterator->run()`, nothing is caught internally. It would not record anything in the report and would go out of the method, ignoring all follow up Closures.

For `$benchmark->start()`, you would need to call `stop()` inside the `catch` block.

kenjis: I don't think we need `try-finally`. 

An exception should be thrown in an exceptional case, and not recoverable.
Normally, a developer would modify the code to prevent the exception from occurring.

If a developer really wants to know the elapsed time when an exception is thrown, they can catch and measure it by themselves.

tests/system/Debug/TimerTest.php
```diff
@@ -11,7 +11,10 @@
 
 namespace CodeIgniter\Debug;
 
+use ArgumentCountError;
 use CodeIgniter\Test\CIUnitTestCase;
+use RuntimeException;
+use Throwable;
```
kenjis, line 17:
```suggestion
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +124,95 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
+        $this->assertGreaterThanOrEqual(1.0, timer()->getElapsedTime('common'));
+    }
+
+    public function testCommonCallableExpectWithReturn()
+    {
+        $returnValue = timer('common', static fn () => strlen('CI4'));
+
+        $this->assertNotInstanceOf(Timer::class, $returnValue);
+        $this->assertSame(3, $returnValue);
+        $this->assertLessThanOrEqual(1.0, timer()->getElapsedTime('common'));
```
kenjis, line 216:
```suggestion
        $this->assertLessThan(0.1, timer()->getElapsedTime('common'));
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +124,95 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
+    /**
+     * @timeLimit 1.5
+     */
+    public function testCommonCallableExpectNoReturn()
+    {
+        $returnValue = timer('common', static function () { sleep(1); });
+
+        $this->assertNotInstanceOf(Timer::class, $returnValue);
+        $this->assertNull($returnValue);
+        $this->assertGreaterThanOrEqual(1.0, timer()->getElapsedTime('common'));
```
kenjis, line 207:
```suggestion
        $returnValue = timer('common', static function () { usleep(100000); });

        $this->assertNotInstanceOf(Timer::class, $returnValue);
        $this->assertNull($returnValue);
        $this->assertGreaterThanOrEqual(0.1, timer()->getElapsedTime('common'));
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +124,95 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
+        $this->assertLessThanOrEqual(1.0, $timer->getElapsedTime('longjohn'));
+        $this->assertSame('test', $returnValue);
+    }
+
+    public function testRecordArrowFunction()
+    {
+        $timer       = new Timer();
+        $returnValue = $timer->record('longjohn', static fn () => strlen('CI4'));
+
+        $this->assertLessThanOrEqual(1.0, $timer->getElapsedTime('longjohn'));
```
kenjis, line 156:
```suggestion
        $this->assertLessThan(0.1, $timer->getElapsedTime('longjohn'));
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +124,95 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
+    }
+
+    public function testRecordFunctionWithReturn()
+    {
+        $timer       = new Timer();
+        $returnValue = $timer->record('longjohn', static function () {
+            return 'test';
+        });
+
+        $this->assertLessThanOrEqual(1.0, $timer->getElapsedTime('longjohn'));
```
kenjis, line 147:
```suggestion
        $this->assertLessThan(0.1, $timer->getElapsedTime('longjohn'));
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +124,95 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
         $this->assertNull($timer->getElapsedTime('test1'));
     }
+
+    /**
+     * @timeLimit 1.5
+     */
+    public function testRecordFunctionNoReturn()
+    {
+        $timer       = new Timer();
+        $returnValue = $timer->record('longjohn', static function () { sleep(1); });
```
kenjis, line 134:
```suggestion
        $returnValue = $timer->record(
            'longjohn',
            static function () { usleep(100000); }
        );
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +124,95 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
+
+        $this->assertGreaterThanOrEqual(1.0, $timer->getElapsedTime('longjohn'));
```
kenjis, line 136:
```suggestion
        $this->assertGreaterThanOrEqual(0.1, $timer->getElapsedTime('longjohn'));
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +124,95 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
 
         $this->assertNull($timer->getElapsedTime('test1'));
     }
+
+    /**
+     * @timeLimit 1.5
+     */
```
kenjis, line 130:
```suggestion
```

tests/system/Debug/TimerTest.php
```diff
@@ -121,4 +124,95 @@ public function testReturnsNullGettingElapsedTimeOfNonTimer()
...
+    public function testCommonNoNameCallableExpectTimer()
+    {
+        $returnValue = timer(null, static fn () => strlen('CI4'));
+
+        $this->assertInstanceOf(Timer::class, $returnValue);
+    }
+
+    /**
+     * @timeLimit 1.5
+     */
```
kenjis, line 200:
```suggestion
```

kenjis:
Please fix coding style and failed tests.

tests/system/Debug/TimerTest.php
kenjis: This is not updated. Can you?

rumpfc:
Calling `strlen` without any parameters has 2 different behaviours between PHP7 and PHP8. That's why the unit test keep failing for 7.4, but not for 8.0 and 8.1:

```php
// PHP 7.4 and before
strlen();   // throws ErrorException (extends Exception)

// PHP 8.0 and after
strlen();   // throws ArgumentCountError (extends TypeError)
```

@kenjis I can offer 2 solutions:
1. Unit test expects a `Throwable` to be thrown
2. I catch `ErrorException` inside `record()` and throw an `ArgumentCountError`

kenjis:
@rumpfc Not specifically catch Exception in the method. So it's weird to catch only ErrorException.
I don't know all ErrorExceptions in PHP7 will be ArgumentCountError in PHP8.

So it's okay to expect Throwable in the test.
Also, please comment why we need to use Throwable.

kenjis (approved):
LGTM!

MGatner (approved):

MGatner:
Thanks all!

PR #6322, opened 2022-08-01
chore: add extensions for `suggest` in composer.json

kenjis:
**Description**
- add extentions for `suggest`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,18 @@
         "rector/rector": "0.13.10"
     },
     "suggest": {
-        "ext-fileinfo": "Improves mime type detection for files"
+        "ext-imagick": "If you use Image class ImageMagickHandler",
+        "ext-simplexml": "If you format XML",
+        "ext-mysqli": "If you use MySQL",
+        "ext-oci8": "If you use Oracle Database",
+        "ext-pgsql": "If you use PostgreSQL",
+        "ext-sqlsrv": "If you use SQL Server",
+        "ext-sqlite3": "If you use SQLite3",
+        "ext-memcache": "If you use Cache class MemcachedHandler with Memcache",
+        "ext-memcached": "If you use Cache class MemcachedHandler with Memcached",
+        "ext-redis": "If you use Cache class RedisHandler",
+        "ext-fileinfo": "Improves mime type detection for files",
+        "ext-readline": "Improves CLI::input() usability"
     },
     "config": {
         "optimize-autoloader": true,


kenjis:
`ext-xml` is not listed in `composer.json`, but it is checked in `CodeIgniter.php`
I'm not sure `xml` is really needed.

CodeIgniter4/CodeIgniter.php at develop · codeigniter4/CodeIgniter4
https://github.com/codeigniter4/CodeIgniter4/blob/a1a635814eb5c254b5a31cbfde054ff54f0b30a3/system/CodeIgniter.php#L214-L220

samsonasik (approved):

MGatner (changes requested):
Could you also update **user_guide/intro/requirements.html**? We have some inconsistencies between here and there about what is "required" versus "suggested".

MGatner:
> I'm not sure xml is really needed.

Sorry, missed this comment before review.

kenjis:
> what is "required" versus "suggested".

I think `curl` is suggested. It is needed only if you use CURLRequest.

composer.json
```diff
@@ -27,7 +27,15 @@
...
-        "ext-fileinfo": "Improves mime type detection for files"
+        "ext-imagick": "If you use ImageMagickHandler",
+        "ext-simplexml": "If you format XML",
+        "ext-mysqli": "If you use MySQL",
+        "ext-oci8": "If you use Oracle Database",
+        "ext-pgsql": "If you use PostgreSQL",
+        "ext-sqlsrv": "If you use SQL Server",
+        "ext-sqlite3": "If you use SQLite3",
+        "ext-fileinfo": "Improves mime type detection for files",
+        "ext-readline": "Improves CLI::input() usability"
```
MGatner, line 38: I think Composer Normalize will alphabetize these:
```suggestion
        "ext-fileinfo": "Improves mime type detection for files",
        "ext-imagick": "If you use ImageMagickHandler",
        "ext-mysqli": "If you use MySQL",
        "ext-oci8": "If you use Oracle Database",
        "ext-pgsql": "If you use PostgreSQL",
        "ext-readline": "Improves CLI::input() usability",
        "ext-simplexml": "If you format XML",
        "ext-sqlite3": "If you use SQLite3",
        "ext-sqlsrv": "If you use SQL Server"
```

MGatner (changes requested):

composer.json
kenjis: What is *Composer Normalize* ?

paulbalandan: It's a Composer plugin. `ergebnis/composer-normalize`

kenjis: It seems we don't use it.

```console
$ composer normalize
                                       
  Command "normalize" is not defined.  
```

MGatner: Oops! My mistake. I use it locally and it is part of DevKit so I assumed we had it here, but maybe not. We can always come back and normalize later - merging this now so it makes it into `4.2.2`.

PR #6332, opened 2022-08-02
Add RawSql to BaseConnection->escape()

sclubricants:
This PR adds ``RawSql`` type to ``BaseConnection->escape()``. It does not escape the ``RawSql``.

This allows passing SQL functions in columns used for DML operations.

For instance you could pass `CURRENT_TIMESTAMP()` to a date time field instead of a literal value.

You could call your own SQL function as well.

Another is to pass SQL constants like `DEFAULT` in order to fill a column with default values.

```php
<?php

    $data = [
        'id'          => new RawSql('DEFAULT'),
        'title'       => 'My title',
        'name'        => 'My Name',
        'date'        => '2022-01-01',
        'last_update' => new RawSql('CURRENT_TIMESTAMP()'),
    ];

    $builder->insert($data);
    /* Produces:
        INSERT INTO mytable (id, title, name, date, last_update)
        VALUES (DEFAULT, 'My title', 'My name', '2022-01-01', CURRENT_TIMESTAMP())
    */
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


MGatner:
Looks like a cool feature! The changes look good to me but I'd rather someone with a better idea of the larger database implications handle reviews. @iRedds or @paulbalandan ?

system/Database/BaseConnection.php
```diff
@@ -1238,6 +1238,10 @@ public function escape($str)
         }
 
         if (is_string($str) || (is_object($str) && method_exists($str, '__toString'))) {
```
paulbalandan, line 1240: Do we have other database related classes which have the `__toString()` that can be passed here as query string.?

user_guide_src/source/database/call_function.rst
```diff
@@ -37,3 +37,18 @@ database result ID. The connection ID can be accessed using:
 The result ID can be accessed from within your result object, like this:
 
 .. literalinclude:: call_function/004.php
+
+$db->escape();
+============================
```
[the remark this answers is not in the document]
sclubricants: I looked but couldn't find the escape section..

It does allow for calling custom functions.. but in a different way. Maybe that section I did on the function page should instead have a blurb on use of RawSql class to call imbedded functions.

system/Database/BaseConnection.php
sclubricants: I'm not sure, I just thought it was there because such objects would be compatible.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -43,6 +43,7 @@ Enhancements
 - Added ``spark filter:check`` command to check the filters for a route. See :ref:`Controller Filters <spark-filter-check>` for the details.
 - Now **Encryption** can decrypt data encrypted with CI3's Encryption. See :ref:`encryption-compatible-with-ci3`.
 - Now ``spark routes`` command shows route names. See :ref:`URI Routing <routing-spark-routes>`.
+- ``BaseConnection::escape()`` now excludes the ``RawSql`` data type. This allows passing SQL strings into data. See :ref:`database-queries-db-escape`.
```
sclubricants, line 47: undefined label: database-queries-db-escape

How do I link this? 

See :ref:\`database-queries-db-escape\`.

kenjis: See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/documentation.rst#references

user_guide_src/source/database/queries/030.php
```diff
@@ -0,0 +1,15 @@
...
+        'name'        => 'My Name',
+        'date'        => '2022-01-01',
+        'last_update' => new RawSql('CURRENT_TIMESTAMP()'),
+    ];
+
+    $builder->insert($data);
+    /* Produces:
+        INSERT INTO mytable (id, title, name, date, last_update)
+        VALUES (DEFAULT, 'My title', 'My name', '2022-01-01', CURRENT_TIMESTAMP())
+    */
```
kenjis, line 15: Remove this sample.
This sample code is weird. Because it is the explanation for `$db->escape()` but there is no `$db->escape()`.

user_guide_src/source/database/query_builder/076.php
```diff
@@ -1,10 +1,15 @@
...
-    'name'  => 'My Name',
-    'date'  => 'My date',
-];
+    $data = [
+        'id'          => new RawSql('DEFAULT'),
+        'title'       => 'My title',
+        'name'        => 'My Name',
+        'date'        => '2022-01-01',
+        'last_update' => new RawSql('CURRENT_TIMESTAMP()'),
+    ];
```
kenjis, line 9: Remove indentation.

user_guide_src/source/database/query_builder/076.php
```diff
@@ -1,10 +1,15 @@
...
 
-$builder->insert($data);
-// Produces: INSERT INTO mytable (title, name, date) VALUES ('My title', 'My name', 'My date')
+    $builder->insert($data);
+    /* Produces:
+        INSERT INTO mytable (id, title, name, date, last_update)
+        VALUES (DEFAULT, 'My title', 'My name', '2022-01-01', CURRENT_TIMESTAMP())
+    */
```
kenjis, line 15: Remove indentation.

user_guide_src/source/database/queries.rst
```diff
@@ -102,12 +102,17 @@ this:
...
-This function determines the data type so
-that it can escape only string data. It also automatically adds
-single quotes around the data so you don't have to:
+This function determines the data type so that it can escape only string
+data. It also automatically adds single quotes around the data so you
+don't have to:
 
 .. literalinclude:: queries/009.php
 
+When ``RawSql`` is used the string is not escaped. This allows you to call
```
kenjis, line 111:
```suggestion
Since v4.3.0, when ``RawSql`` is used the string is not escaped. This allows you to call
```

kenjis:
Please update the all notes in query_builder.rst:
> .. note:: All values are escaped automatically producing safer queries.

MGatner:
Looking good! Thanks for all the reviews.

sclubricants:
> .. note:: All values are escaped automatically producing safer queries.

@kenjis What do you propose we do here?

How about:

.. note:: All values except \`\`RawSql\`\` are escaped automatically producing safer queries.

user_guide_src/source/changelogs/v4.3.0.rst
sclubricants: I think I have it fixed..

user_guide_src/source/database/queries/030.php
sclubricants: I added escape() test sample instead.

kenjis:
I think the following section is weird. Because it explains `escape()`, but it explains about **not** escaping a lot.

Screenshot 2022-08-05 7 56 17

I think we should remove the sample and description for `RawSql` here.
Because it is not needed here.

This section explains how to use `escape()` and its behavior.
Devs never writes or should not write code like `$db->escape(new RawSql('...'))`.
It does not make sense.

See the first sample code. The assumption here is that SQL statements are generated by concatenating strings.
Devs can write SQL functions and constants as they like in a string.
But the sentence "*This allows you to call SQL functions and constants*" is an explanation when we use QueryBuilder.

kenjis:
> .. note:: All values except ``RawSql`` are escaped automatically producing safer queries.

Good. And can you add the following warning, too?

Screenshot 2022-08-05 8 16 12

sclubricants:
So you want to remove everything? Then none will know about its availability for use. 

I think the example is good because it explains what the method does. The method does different things to different types of data. It could be possible to add other data types such as Datetime.

Its important to know how these datatypes are treated.

sclubricants:
Developers might should understand that the escape method is what they are relying on when they employ functions like insert. They might not write the actual code for db->escape() but they are none the less calling it via ignoring the escape parameter on functions like insert().

kenjis:
> So you want to remove everything?

I thought again, and yes. Because the page explains how to use `$db->query()` (and `simpleQuery()`).
So `RawSql` is not needed for them.

Your explanation about `RawSql` and `escape()` is good, but it should not in the place.

> Developers might should understand that the escape method is what they are relying on when they employ functions like insert. They might not write the actual code for db->escape() but they are none the less calling it via ignoring the escape parameter on functions like insert().

Yes, and the explanation should be in the `insert()` page. It is a QueryBuilder method.

sclubricants:
Maybe RawSql needs its own page

sclubricants:
I wonder if this will conflict with binds..

MGatner:
> I wonder if this will conflict with binds.. 

Sounds like a test case is in order 😊

sclubricants:
No conflict with binds but did have to fix objectToArray() in BaseBuilder

kenjis:
> No conflict with binds but did have to fix objectToArray() in BaseBuilder

Why doesn't the existing test fail?
Is a test case missing?

sclubricants:
The existing test just tests escape(). Ill write some tests to test things like insert().

sclubricants:
Added RawSql test.. to test custom SQL function: `setDateTime('2022-06-01')` 

The function just adds 01:01:11 as time to the date.

This way the RawSql has quotes in it to test and make sure quotes aren't being escaped.

system/Database/BaseConnection.php
sclubricants: From the looks of it you could create a wrapper class for DateTime with __toString() and use it here. I thought it would be nice to build some sort of class that could work with each DB's date format.. and then also have a frontend format.

MGatner:
I think a dedicated documentation page is a good idea.

I don't like the `rawSql()` method - it functions as a named constructor which should belong on the class itself. In actuality though the constructor itself is sufficient, so this simply ends up creating a class dependency that wasn't there before. 

Open to other ideas or push-back if I have kissed something.

kenjis:
I don't see the benefit of `$this->db->rawSql()`. What's better?

MGatner:
If there's impetus to wrap this in a method then it should be something like `RawSql::of($sql)`, with an optional `protected` constructor. But I still don't know that it adds anything.

sclubricants:
I was just looking for a solution that made easy access to class without having to include the class path. I thought a method like would make development easier.

kenjis:
> I was just looking for a solution that made easy access to class without having to include the class path.

Why do you want not having the class path? Don't you use IDE?
If you really want it, it would be good to have a helper (global function) for it.

tests/system/Database/Live/MySQLi/RawSqlTest.php
```diff
@@ -0,0 +1,203 @@
...
+
+        $this->db->query($sql);
+    }
+
+    public function testRawSqlUpdateObject()
+    {
+        $data = [];
+
+        $row             = new stdclass();
+        $row->email      = 'derek@world.com';
```
kenjis, line 63: Don't use someone's domain name.

```suggestion
        $row->email      = 'derek@example.com';
```

kenjis (changes requested):
Please change all of the test email addresses to `example.com`. Otherwise it looks good.

tests/system/Database/Live/MySQLi/RawSqlTest.php
sclubricants: These email addresses are in seeder and used for all the tests.

https://github.com/codeigniter4/CodeIgniter4/blob/4.3/tests/_support/Database/Seeds/CITestSeeder.php

Perhaps changing these should be a separate PR?

kenjis: Oh, yes.

kenjis (approved):

MGatner:
I won't look back through all of it, so unless anyone else has more comments this is good by me. Thanks for this feature @sclubricants - very nice addition!

sclubricants:
I made a couple of fixes. One is that I didn't get the logic quite right for the `objectToArray()` method.

The other is that I was using `is_a($val, RawSql::class) `and changed it to `$val instanceof RawSql`.

The php function `is_a()` requires the first parameter be an object or class, whereas `instanceof` does not.

https://www.php.net/manual/en/function.is-a.php

https://www.php.net/manual/en/language.operators.type.php

MGatner:
`is_a()` defaults to objects only unless you use the third parameter, so what you had was fine but I think `instanceof` is the preferred choice here anyway.

MGatner (approved):
@kenjis last call?

kenjis:
Sorry, can you rebase to resolve the conflict?

system/Database/BaseBuilder.php
```diff
@@ -2698,7 +2698,7 @@ protected function objectToArray($object)
 
         foreach (get_object_vars($object) as $key => $val) {
             // There are some built in keys we need to ignore for this conversion
-            if ((! is_object($val) && ! is_array($val) && $key !== '_parent_name') || is_a($val, RawSql::class)) {
+            if (((! is_object($val) || $val instanceof RawSql) && ! is_array($val) && $key !== '_parent_name')) {
```
kenjis, line 2701: What's the difference?
Can you add a test case?

sclubricants: I was just keeping intact the logic that was there already. They had it before that $key !== '_parent_name' under any circumstances. I'm really not sure what its there for. 

The original logic was: if (! is_object($val) && ! is_array($val) && $key !== '_parent_name')

I added: (is not and object unless its RawSql)

sclubricants: I was just making sure we ignored:

```php
// There are some built in keys we need to ignore for this conversion
```

Apparently some objects must have a property _parent_name that needs to be ignored.

kenjis: Do you mean?

```php
        foreach (get_object_vars($object) as $key => $val) {
            // There are some built in keys we need to ignore for this conversion
            if (! is_object($val) && ! is_array($val) && $key !== '_parent_name') {
                if ((! $val instanceof RawSql)) {
                    $array[$key] = $val;
                }
            }
        }
```

kenjis: I got it.
Can you add the following test case?
```diff
--- a/tests/system/Database/Builder/InsertTest.php
+++ b/tests/system/Database/Builder/InsertTest.php
@@ -13,6 +13,7 @@ namespace CodeIgniter\Database\Builder;
 
 use CodeIgniter\Database\Exceptions\DatabaseException;
 use CodeIgniter\Database\Query;
+use CodeIgniter\Database\RawSql;
 use CodeIgniter\Test\CIUnitTestCase;
 use CodeIgniter\Test\Mock\MockConnection;
 
@@ -85,6 +86,21 @@ final class InsertTest extends CIUnitTestCase
         $this->assertSame($expectedBinds, $builder->getBinds());
     }
 
+    public function testInsertObjectWithRawSql()
+    {
+        $builder = $this->db->table('jobs');
+
+        $insertData = (object) [
+            'id'   => 1,
+            'name' => new RawSql('CONCAT("id", \'Grocery Sales\')'),
+        ];
+        $builder->testMode()->insert($insertData, true);
+
+        $expectedSQL = 'INSERT INTO "jobs" ("id", "name") VALUES (1, CONCAT("id", \'Grocery Sales\'))';
+
+        $this->assertSame($expectedSQL, str_replace("\n", ' ', $builder->getCompiledInsert()));
+    }
+
     /**
      * @see https://github.com/codeigniter4/CodeIgniter4/issues/5365
      */
```

system/Database/BaseBuilder.php
```diff
@@ -2698,7 +2698,7 @@ protected function objectToArray($object)
 
         foreach (get_object_vars($object) as $key => $val) {
             // There are some built in keys we need to ignore for this conversion
-            if (! is_object($val) && ! is_array($val) && $key !== '_parent_name') {
+            if (((! is_object($val) || $val instanceof RawSql) && ! is_array($val) && $key !== '_parent_name')) {
```
iRedds, line 2701: It seems to me that there are too many parentheses.

iRedds:
The usage example is rather ambiguous.
It can be replaced with.
```php
$data = [
    'title'       => 'My title',
    'name'        => 'My Name',
    'date'        => '2022-01-01',
];

$builder->set([
    'id'          => 'DEFAULT',
    'last_update' => 'CURRENT_TIMESTAMP()',
], null, false)->insert($data);
```

system/Database/BaseBuilder.php
sclubricants: Done

system/Database/BaseBuilder.php
sclubricants: Your right, thanks

kenjis (approved):

kenjis:
@iRedds  To be honest, I would like to throw an Exception to such code. It is difficult to read.

By the way, what do you mean *ambiguous* ?
It seems to me the sample code is easy to read and understand.

MGatner:
I think @iRedds was pointing out that we can already accomplish the same thing with other tools. `insertBatch()` is a good counter-example. I also assume this is the only way to use a Model to insert non-escaped data?

kenjis:
I agree that we can already accomplish the same thing with other tools.
But I think the sample code is okay even if there is another way to do the same thing.

iRedds (approved):

kenjis (approved):

MGatner:
Thanks all!

PR #6334, opened 2022-08-03
fix: random_string('crypto') may return string less than $len or ErrorException

### LINKED ISSUE #6330: Bug: random_string() Implicit conversion from float 4.5 to int loses precision
### PHP Version

8.1

### CodeIgniter4 Version

4.2.1

Linux Debian

### Which server did you use?

fpm-fcgi

### Database

MariaDB

### What happened?

I'm using the codeigniter build in helper `helper('text');` and want to create a random string called the function `random_string("crypto", 9);` With an odd number I always get the error: *Implicit conversion from float 4.5 to int loses precision*. So the numbers 3, 5, 7, 9, 11 (...) are not usable. With even numbers I always get a random string.

**What function is called?**

* system/Helpers/text_helper.php

```php
if (!function_exists('random_string')) {
  /**
   * Create a Random String
   *
   * Useful for generating passwords or hashes.
   *
   * @param string $type Type of random string.  basic, alpha, alnum, numeric, nozero, md5, sha1, and crypto
   * @param int $len Number of characters
   */
  function random_string(string $type = 'alnum', int $len = 8) : string {}
}
```

This case I call with `random_string("crypto", 9)`:

```php
return bin2hex(random_bytes($len / 2));
```

### Steps to Reproduce

Didn't work:
```php
helper('text');
$pw = random_string("crypto", 9);
```

Works fine:
```php
helper('text');
$pw = random_string("crypto", 8);
```

### Expected Output

A random string like *b3da1087d* or something else.

### Anything else?

Last two Backtrace rows:

1. {PHP internal code}   —  CodeIgniter\Debug\Exceptions->errorHandler ()
2. SYSTEMPATH/Helpers/text_helper.php : 576   —   random_bytes() 

grafik

**Solution maybe**

Cast the result to an integer:

```
return bin2hex(random_bytes((int)($len / 2)));
```

kenjis:
**Description**
Fixes #6330

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/text_helper.php
+++ b/system/Helpers/text_helper.php
@@ -573,6 +573,12 @@ if (! function_exists('random_string')) {
                 return sha1(uniqid((string) mt_rand(), true));
 
             case 'crypto':
+                if ($len % 2 !== 0) {
+                    throw new InvalidArgumentException(
+                        'You must set an even number to the second parameter when you use `crypto`.'
+                    );
+                }
+
                 return bin2hex(random_bytes($len / 2));
         }
         // 'basic' type treated as default
--- a/tests/system/Helpers/TextHelperTest.php
+++ b/tests/system/Helpers/TextHelperTest.php
@@ -12,6 +12,7 @@
 namespace CodeIgniter\Helpers;
 
 use CodeIgniter\Test\CIUnitTestCase;
+use InvalidArgumentException;
 
 /**
  * @internal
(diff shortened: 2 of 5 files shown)


MGatner (approved):

kenjis:
Added the docs.

paulbalandan (approved):

MGatner (approved):

PR #6342, opened 2022-08-05
chore: update Kint to 4.1.4 in ThirdParty

kenjis:
**Description**
- update Kint

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide


samsonasik (approved):

samsonasik:
composer.json require kint need to be updated to kint 4.1.4

MGatner (approved):

MGatner:
@samsonasik since `ThirdParty` is only for non-Composer installations I don't think we have to keep the versions consistent. We can discuss this more but I am merging this for now so it can make it into `4.2.2`

PR #6340, opened 2022-08-04
fix: validation custom error with asterisk field

ping-yee:
**Description**
Fixes #6245 

- increase judgment if user set up custom validation error with asterisk field.
- increase **testRunWithCustomErrorsAndAsteriskField** for testing

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [x] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

PR #6352, opened 2022-08-06
fix: validation custom error with asterisk field

### LINKED ISSUE #6245: Bug: Custom validation error for a field with an asterisk.
### CodeIgniter4 Version

dev

### What happened?

If you specify a custom error message for a field with an asterisk, the default error text will be used.
This is due to the fact that a field without an asterisk is passed to generate an error (it contains the full path of the key)

### Steps to Reproduce

```php
$data = [
    'user' => [
        ['name' => null],
        ['name' => null],
    ]
];

$validator = Services::validation();

$validator->setRules(['user.*.name' => 'required'], ['user.*.name' => ['required' => 'Required']]);

$validator->run($data);

dd($validator->getErrors());
// array (2) [
//     'user.0.name' => 'The user.*.name field is required.'
//     'user.1.name' => 'The user.*.name field is required.'
// ]
```

### Expected Output

```php
array (2) [
    'user.0.name' => 'Required'
    'user.1.name' => 'Required'
]
```

ping-yee:
**Description**
Fixes #6245 
Supersedes #6340 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


iRedds:
I would like to ask you to change the declaration of methods according to PSR-12 (there is an example here: https://www.php-fig.org/psr/psr-12/#45-method-and-function-arguments). Because the string is longer than 120 characters.

iRedds:
@kenjis @MGatner What do you think if the three arguments ($field, $label, $originalField) are replaced with an array or a DTO?

tests/system/Validation/StrictRules/ValidationTest.php
```diff
@@ -359,6 +359,28 @@ public function testRunGroupWithCustomErrorMessage(): void
...
+            ],
+        ];
+        $this->validation->setRules(
+            ['foo.*.bar' => ['label' => 'foo bar', 'rules' => 'required']],
+            ['foo.*.bar' => ['required' => 'Required']]
+        );
+        $this->validation->run($data);
+        $this->assertSame([
+            'foo.0.bar' => 'Required',
+            'foo.1.bar' => 'Required',
```
kenjis, line 380: Are you really okay with keys like `'foo.0.bar'`?
It seems a bit difficult to use.

system/Validation/Validation.php
```diff
@@ -203,7 +203,7 @@ public function check($value, string $rule, array $errors = []): bool
      * @param array|null   $rules
      * @param array        $data
```
kenjis, line 204:
```suggestion
     * @param array        $data          The array of data to validate, with `DBGroup`.
     * @param string|null  $originalField The original asterisk field name like "foo.*.bar".
```

kenjis:
> What do you think if the three arguments ($field, $label, $originalField) are replaced with an array or a DTO?

It seems this fix needs BC break, so I'm fine with the change.

MGatner:
These methods have way too many parameters. I don't know that adding a new one is worse than a breaking change though. Definitely we should break this up somehow in version 5. If anyone can think of a way to handle this now as backwards-compatible I'm open to it.

tests/system/Validation/StrictRules/ValidationTest.php
ping-yee: But as I known, validation service change the asterisk field to like `'foo.0.bar'`.
As Issue #6245 show that.

ping-yee:
@iRedds @kenjis  I have been changed declaration of methods to follow PSR-12 and added the comment out of `originalField` parameter.

iRedds:
@ping-yee Add a description to the changelog.

kenjis:
v4.2.3 has been released. See https://github.com/codeigniter4/CodeIgniter4/releases
This change will be included in v4.2.4.

Please rebase and add changelog v4.2.4.

ping-yee:
@kenjis Is this normal? Or I do the wrong thing when I rebase my branch...
If I do the wrong thing, please tell me how can I fix it :(

kenjis:
@ping-yee This is not normal. You did something wrong.
What commands did you run?
And please don't use `git merge`.

ping-yee:
@kenjis I use sourcetree to manage my git flow.
I syncing your repository first, and run rebase onto devlop branch.
After I resolve the conflicts, run rebase onto devlop branch again(Maybe there is a problem with this step).
Then I push the commit `fa442b4` and get the results.
Should I re-open the new PR or do something to fix this?
Sorry, this is my first time to contribute :(

kenjis:
> I syncing your repository first, and run rebase onto devlop branch.

My repository? Do you mean https://github.com/kenjis/Codeigniter4 ?
If so, my repository has nothing to do with you.

> After I resolve the conflicts, run rebase onto devlop branch again(Maybe there is a problem with this step).

You did rebase onto develop twice.
I don't know why your branch is like this.

> Should I re-open the new PR or do something to fix this?

If you drop the commits other than you did for this PR (the commits marked below) with `git rebase -i`,
you can revert the branch state. Then rebase again correctly.

Screenshot 2022-08-13 16 56 26

But you don't use `git` command, I can't support you.

If you can re-open the new PR, that is also fine.

ping-yee:
> I syncing your repository first, and run rebase onto devlop branch.
@kenjis I say the wrong word. Not `your repository`, `my repository` is right word.
Thanks for your help and I'll use command to rebase my git flow!

kenjis:
Okay, if syncing your repository means this https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#syncing-your-repository , you just did rebase twice.

In my understanding, if you did rebase twice, the result is the same. The second rebase does nothing.
So I still have no clue for the status of your mess branch.

ping-yee:
Would it be to I don't syncing my repo before I commit my change?
I try to fix the git flow later.

kenjis:
> Would it be that I don't syncing my repo before I commit my change?

Sorry, I don't get your situation.
I thought you synced your repository, and did rebase twice.
But you say you did not sync?

After all, the current your branch is abnormal state.
There are two commits that are the same content.
`git rebase` never creates a new commit. It just change the existing commits.

Either you did something I couldn't anticipate, or Sourcetree ran strange commands that I couldn't anticipate.
In any case, I do not know of any way to get the branch into this state.

ping-yee:
I think my git flow is too mess to fixed, so I decide to re-open the PR.
Thanks for your help and teach again! @kenjis

PR #6361, opened 2022-08-09
fix: Response::download() causes TypeError

### LINKED ISSUE #6358: Bug: Force File Download error in the latest version (4.2.3)
### PHP Version

7.4

### CodeIgniter4 Version

4.2.3

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

when I executed `return $this->response->download('some.txt', 'some text', true)`
it return an error below.
```
TypeError

Argument 1 passed to CodeIgniter\CodeIgniter::displayPerformanceMetrics() must be of the type string, null given, 
called in D:\laragon\www\iprisinsi\vendor\codeigniter4\framework\system\CodeIgniter.php on line 500 
```

**SYSTEMPATH\CodeIgniter.php at line 740**
```
733 
734         return md5($name);
735     }
736 
737     /**
738      * Replaces the elapsed_time tag.
739      */
740     public function displayPerformanceMetrics(string $output): string
741     {
742         return str_replace('{elapsed_time}', (string) $this->totalTime, $output);
743     }
744 
745     /**
746      * Try to Route It - As it sounds like, works with the router to
747      * match a route against the current URI. If the route is a
```

actually I found this when 4.2.2 released 

### Steps to Reproduce

just run this code on home Controller
`return $this->response->download('some.txt', 'some text', true)`

### Expected Output

file `some.txt` downloaded

### Anything else?

since I m new to this I couldn't point what file to fix

kenjis:
**Description**
Fixes  #6358

- never cache DownloadResponse/RedirectResponse
- add null check for $this->response->getBody()

**How to Test**
```php
<?php

namespace App\Controllers;

{
    public function index()
    {
        return $this->response->download('some.txt', 'some text', true);
    }
}
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -489,20 +489,26 @@ class CodeIgniter
             $this->response = $response;
         }
 
-        // Cache it without the performance metrics replaced
-        // so that we can have live speed updates along the way.
-        // Must be run after filters to preserve the Response headers.
-        if (static::$cacheTTL > 0) {
-            $this->cachePage($cacheConfig);
+        // Skip unnecessary processing for special Responses.
+        if (! $response instanceof DownloadResponse && ! $response instanceof RedirectResponse) {
+            // Cache it without the performance metrics replaced
+            // so that we can have live speed updates along the way.
+            // Must be run after filters to preserve the Response headers.
+            if (static::$cacheTTL > 0) {
+                $this->cachePage($cacheConfig);
+            }
+
+            // Update the performance metrics
+            $body = $this->response->getBody();
+            if ($body !== null) {
+                $output = $this->displayPerformanceMetrics($body);
+                $this->response->setBody($output);
+            }
+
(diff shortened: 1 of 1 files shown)


MGatner (approved):

MGatner:
Should `RedirectResponse` be checked? Also, I think `null` is an acceptable body for some API responses - we might need to check or cast that.

MGatner (changes requested):
_See previous comments._

nicojmb:
Hi, same problem here, just tested[kenjis](https://github.com/kenjis) fix and works fine.

kenjis:
Changed the implementation.

MGatner (approved):
Thank you!

MGatner:
Any ideas how this made it through tests to release?

Issue #6351, opened 2022-08-06
Bug: createTable() method in Forge class checks for table existence using cache and causes problems


AnakinA1:
### PHP Version

8.1

### CodeIgniter4 Version

4.2.2

### CodeIgniter4 Installation Method

Composer (as dependency to an existing project)

### Which operating systems have you tested for this bug?

macOS, Linux

### Which server did you use?

fpm-fcgi

### Database

MariaDB 10.8

### What happened?

In the new minor version of CI Database/Forge class has now:

```
// If table exists lets stop here
if ($ifNotExists === true && $this->db->tableExists($table)) {
    $this->reset();

    return true;
}
```

Let's focus on `$this->db->tableExists($table)` if flag `$ifNotExists ` is true

Will give a real example of issue:

For tests I have a migration which fully drop current database and restore it from the provided SQL file. I'm doing it with direct db connection (`$this->db->mysqli`) using `multi_query()`. Then I need to create a migration table in order for the migration system to work. I'm doing it with Forge class, because it's a framework technical table and should not be included into project SQL scheme.

### Steps to Reproduce

In the migration file:

- drop db
- add db
- then try to create migration table if it doesn't exist (to prevent possible error on failing)
- table is not created, because the check `$this->db->tableExists($table)` uses cache

This example only applies to my case, but let's assume the following:

Just drop any table with `query` method and then try to restore it via forge class

### Expected Output

Forge class shouldn't rely on cache, or should be highlighted in the documentation that Forge class relies on the cache in its work and should not be used (or it is necessary to clear the cache) when working with direct querying, not through the builder

### Anything else?

Ref #6249

sclubricants:
Perhaps the method should have a parameter that allows for specifying whether to get cached results or not.

```php
$this->db->tableExists($table, $cached = true);
```
Then could could call it with false before running table create.

Cache works fine though if you stay within the framework. 

Same goes cache for indexes.

kenjis:
Yes, I agree with the following.
```php
$this->db->tableExists($table, $cached = true);
```

sclubricants:
Problem is that `tableExists()` gets its data from `listTables()`.. so its not as easy as just looking up one table. 

This could load a list of hundreds of tables. Not somethings I'd want to do before every `createTable()`

We could add a fair bit of code to handle just one table.. OR

My suggestion is that if you are manipulating the database outside of the framework then you should call:

```php
$this->db->resetDataCache();
```
[...]

PR #6364, opened 2022-08-09
Add $cached param to BaseConnection::tableExists()

sclubricants:
Follow-up #6249
This PR fixes https://github.com/codeigniter4/CodeIgniter4/issues/6351

In `Forge::createTable()`

```php
        if ($ifNotExists === true && $this->db->tableExists($table, false)) {
            $this->reset();

            return true;
        }
```
This PR forces to check if table exists rather than relying on cache.

We add $tableName param to `Connection::_listTables()`. This allows us to lookup only one table rather than return a full list of tables.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1431,7 +1431,8 @@ abstract class BaseConnection implements ConnectionInterface
             );
 
             // table doesn't exist but still in cache - lets reset cache, it can be rebuilt later
-            if ($key !== false && ! $tableExists) {
+            // OR if table does exist but is not found in cache
+            if (($key !== false && ! $tableExists) || ($key === false && $tableExists)) {
                 $this->resetDataCache();
             }
         }
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1406,9 +1406,42 @@ abstract class BaseConnection implements ConnectionInterface
     /**
      * Determine if a particular table exists
      */
-    public function tableExists(string $tableName): bool
+    public function tableExists(string $tableName, bool $cached = true): bool
     {
-        return in_array($this->protectIdentifiers($tableName, true, false, false), $this->listTables(), true);
+        if ($cached === true) {
+            return in_array($this->protectIdentifiers($tableName, true, false, false), $this->listTables(), true);
+        }
+
+        if (false === ($sql = $this->_listTables(false, $tableName))) {
+            if ($this->DBDebug) {
+                throw new DatabaseException('This feature is not available for the database you are using.');
+            }
+
+            return false;
+        }
+
+        $result = $this->query($sql)->getResultArray() !== [];
+
+        // if cache has been built already
+        if (! empty($this->dataCache['table_names'])) {
+            $key = array_search(
+                strtolower($tableName),
+                array_map('strtolower', $this->dataCache['table_names']),
+                true
+            );
+
+            // remove from cache
+            if ($key !== false) {
+                unset($this->dataCache['table_names'][$key]);
+            }
+
+            // if exists add back to cache (if cache has been built already)
+            if ($result) {
+                $this->dataCache['table_names'][] = strtolower($tableName);
+            }
+        }
+
+        return $result;
     }
 
     /**
@@ -1575,9 +1608,11 @@ abstract class BaseConnection implements ConnectionInterface
     /**
      * Generates the SQL for listing tables in a platform-dependent manner.
      *
+     * @param string $tableName If $tableName is provided will return only this table if exists.
+     *
      * @return false|string
      */
-    abstract protected function _listTables(bool $constrainByPrefix = false);
+    abstract protected function _listTables(bool $constrainByPrefix = false, string $tableName = '');
 
     /**
      * Generates a platform-specific query string so that the column names can be fetched.
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1405,6 +1405,8 @@ abstract class BaseConnection implements ConnectionInterface
 
     /**
      * Determine if a particular table exists
+     *
+     * @param bool $cached Whether to use data cache
      */
     public function tableExists(string $tableName, bool $cached = true): bool
     {
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1604,11 +1604,11 @@ abstract class BaseConnection implements ConnectionInterface
     /**
      * Generates the SQL for listing tables in a platform-dependent manner.
      *
-     * @param string $tableName If $tableName is provided will return only this table if exists.
+     * @param null|string $tableName If $tableName is provided will return only this table if exists.
      *
      * @return false|string
      */
-    abstract protected function _listTables(bool $constrainByPrefix = false, string $tableName = '');
+    abstract protected function _listTables(bool $constrainByPrefix = false, ?string $tableName = null);
 
     /**
      * Generates a platform-specific query string so that the column names can be fetched.
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1420,7 +1420,7 @@ abstract class BaseConnection implements ConnectionInterface
             return false;
         }
 
-        $result = $this->query($sql)->getResultArray() !== [];
+        $tableExists = $this->query($sql)->getResultArray() !== [];
 
         // if cache has been built already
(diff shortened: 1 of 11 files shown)


system/Database/BaseConnection.php
```diff
@@ -1575,9 +1589,11 @@ abstract public function insertID();
     /**
      * Generates the SQL for listing tables in a platform-dependent manner.
      *
+     * @param false|string $tableName
```
kenjis, line 1592: Please add explanation. What do you mean by table name? and its value false?

system/Database/MySQLi/Connection.php
```diff
@@ -368,11 +368,17 @@ public function escapeLikeStringDirect($str)
...
+     *
+     * @param false|string $tableName
      */
-    protected function _listTables(bool $prefixLimit = false): string
+    protected function _listTables(bool $prefixLimit = false, $tableName = false): string
     {
         $sql = 'SHOW TABLES FROM ' . $this->escapeIdentifiers($this->database);
 
+        if ($tableName !== false) {
+            return $sql . " LIKE '" . $tableName . "'";
```
kenjis, line 379: SQL injection can be done. Use the escape function for it.

system/Database/BaseConnection.php
sclubricants: Means default value is false, else define a string.

kenjis: I know these from the function signature.

`_listTables()` generates the SQL for listing tables.
It is not clear why tableName is needed to list tables.

Please update the doc comment.

sclubricants:
Any idea what is wrong with Mock?

```
1) CodeIgniter\Database\Forge\CreateTableTest::testCreateTableWithExists
Error: Class "Mock_MockQuery_c8b84d32" not found

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Test/Mock/MockConnection.php:58
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:1423
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/Forge.php:501
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Database/Forge/CreateTableTest.php:39
```

```php
    public function query(string $sql, $binds = null, bool $setEscapeFlags = true, string $queryClass = '')
    {
        $queryClass = str_replace('Connection', 'Query', static::class);

        $query = new $queryClass($this);
```

system/Database/MySQLi/Connection.php
kenjis: By the way, why LIKE?

system/Database/MySQLi/Connection.php
```diff
@@ -368,11 +368,17 @@ public function escapeLikeStringDirect($str)
...
+     *
+     * @param false|string $tableName
      */
-    protected function _listTables(bool $prefixLimit = false): string
+    protected function _listTables(bool $prefixLimit = false, $tableName = false): string
     {
         $sql = 'SHOW TABLES FROM ' . $this->escapeIdentifiers($this->database);
 
+        if ($tableName !== false) {
+            return $sql . " LIKE " . $this->escape($tableName);
```
kenjis, line 379: If you use LIKE, escape the value with `escapeLikeString()`.

sclubricants: I'm not using LIKE in the traditional manner with a wildcard.

sclubricants: Not sure what the difference is.

kenjis:
> 1) CodeIgniter\Database\Forge\CreateTableTest::testCreateTableWithExists
Error: Class "Mock_MockQuery_c8b84d32" not found

Because it uses PHPUnit mock.
The `Mock_MockQuery_c8b84d32` does not exists.

system/Database/MySQLi/Connection.php
sclubricants: Well not sure what to do about it.  Do I need to create a mock class?

Honestly, I think scrapping the mock test and doing a live test instead would be better.

sclubricants:
I'll remove it and create a live test

kenjis:
> Well not sure what to do about it. Do I need to create a mock class?

No, you can't. The part `c8b84d32` of the classname `Mock_MockQuery_c8b84d32` will be changed randomly.

> I'll remove it and create a live test

Yes, `createTable()` now depends on the SQL, so the test with mock does not make sense much.

system/Database/BaseConnection.php
MGatner: Please also don't mix types. If you really need a "no input" case it should be `null` instead of `false`name the parameter should be strongly typed as `?string = null`.

MGatner (changes requested):
This is good work! Please address the comments.

system/Database/MySQLi/Connection.php
sclubricants: Its built in to SHOW TABLES. You can use WHERE but the field name is dynamic and changes. LIKE without wildcard should suffice just fine.

These two queries are equivalent.

```SQL
SHOW TABLES FROM `newdb` LIKE 'db_user'

SHOW TABLES FROM `newdb` WHERE Tables_in_newdb = 'db_user'
```

system/Database/BaseConnection.php
sclubricants: Updated comment, change default to empty string to avoid type mismatch.

sclubricants:
I added code that detects when cache is out of sync. If out of sync then we call `resetDataCache()`.

system/Database/MySQLi/Connection.php
```diff
@@ -368,11 +368,17 @@ public function escapeLikeStringDirect($str)
...
      * Uses escapeLikeStringDirect().
+     *
+     * @param string $tableName If $tableName is provided will return only this table if exists.
      */
-    protected function _listTables(bool $prefixLimit = false): string
+    protected function _listTables(bool $prefixLimit = false, string $tableName = ''): string
     {
         $sql = 'SHOW TABLES FROM ' . $this->escapeIdentifiers($this->database);
 
+        if (! empty($tableName)) {
```
MGatner, line 378: I still think the "unused" case for `$tableName` parameter should be null, but either way please make this a proper type check. For string-only type this should be:
```suggestion
        if ($tableName !== '') {
```

MGatner (changes requested):

system/Database/MySQLi/Connection.php
kenjis: I think we should not use `empty()`.
It is not strict typed.

sclubricants: But let your ‘Yes’ be ‘Yes’ and your ‘No’ be ‘No.’
Whatever is more than these is of the evil one

https://en.wikipedia.org/wiki/Matthew_5:37

MGatner (approved):
Looks good to me! Thanks for the null changes. @kenjis anything else from you?

system/Database/BaseConnection.php
```diff
@@ -1406,9 +1406,38 @@ public function listTables(bool $constrainByPrefix = false)
     /**
      * Determine if a particular table exists
```
kenjis, line 1407:
```suggestion
     * Determine if a particular table exists
     *
     * @param bool $cached Whether to use data cache
```

system/Database/BaseConnection.php
```diff
@@ -1406,9 +1406,38 @@ public function listTables(bool $constrainByPrefix = false)
...
      */
-    public function tableExists(string $tableName): bool
+    public function tableExists(string $tableName, bool $cached = true): bool
```
kenjis, line 1409: You have changed the public method. So please add the changelog
like https://codeigniter4.github.io/CodeIgniter4/changelogs/v4.2.2.html#breaking

system/Database/BaseConnection.php
```diff
@@ -1575,9 +1604,11 @@ abstract public function insertID();
     /**
      * Generates the SQL for listing tables in a platform-dependent manner.
      *
+     * @param string|null $tableName If $tableName is provided will return only this table if exists.
+     *
      * @return false|string
      */
-    abstract protected function _listTables(bool $constrainByPrefix = false);
+    abstract protected function _listTables(bool $constrainByPrefix = false, ?string $tableName = null);
```
kenjis, line 1611: You have changed the non final protect method. So please add the changelog.

kenjis (approved):
Thank you.

PR #6367, opened 2022-08-11
Add Upsert/UpsertBatch to builder

sclubricants:
This PR adds Upsert and UpsertBatch methods to builder.

This supersedes https://github.com/codeigniter4/CodeIgniter4/pull/6294 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


sclubricants:
@kenjis 

I'm working on a branch that combines this PR with https://github.com/codeigniter4/CodeIgniter4/pull/6373 and refactors everything.

Putting it all together helps to better envision and combine things. Upsert shares some things with Insert and some with Update.

When I get everything working combined I'll share so you guys can see this big picture.

There are a few things with this PR and https://github.com/codeigniter4/CodeIgniter4/pull/6373 I think I'd like to rework a bit. One thing is how protectIdentifiers is applied and how that affects handling of arrays of constraints and update keys. I'd like them to be consistent in Insert/Upsert/Update batch methods. Then will use some common functions that will apply the protectIdentifiers.

sclubricants:
Can we agree that the primary purpose of methods like setInsertBatch, setUpdateBatch, or my proposed methods setBatch, is to enable collecting of data without executing. This might be in a loop such as reading data from a csv file and setting the data for batch insert. Other than that it might be used to get the sql statements without executing them.

If the reason is this then we only need parameters needed setting the data.. the data itself and whether to escape or not. Things like batch size is not necessary information for setting the data.

kenjis (changes requested):
This PR has only one commit.
The previous PR had more commits.
Please revert this PR to look like the commits in the previous PR.

tomek-artneo:
Good job @sclubricants 👍 This feature will help me a lot 😃

PR #6327, opened 2022-08-01
fix: Parse error occurs before PHP version check

kenjis:
**Description**
Fixes #6306

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #6386, opened 2022-08-17
docs: fix `$pathsConfig` line number

totoprayogo1916:
`$pathsConfig` line number changed after #6327

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated

### DIFF
--- a/user_guide_src/source/general/managing_apps/002.php
+++ b/user_guide_src/source/general/managing_apps/002.php
@@ -1,4 +1,6 @@
 <?php
 
-$pathsConfig = 'app/Config/Paths.php';
+// Load our paths config file
+// This is the line that might need to be changed, depending on your folder structure.
+require FCPATH . '../app/Config/Paths.php';
 // ^^^ Change this line if you move your application folder
--- a/user_guide_src/source/general/managing_apps.rst
+++ b/user_guide_src/source/general/managing_apps.rst
@@ -42,12 +42,11 @@ your main **app/Config/Paths.php** and set a *full server path* in the
 You will need to modify two additional files in your project root, so that
 they can find the **Paths** configuration file:
 
-- **/spark** runs command line apps; the path is specified on or about line 69:
+- **/spark** runs command line apps.
 
   .. literalinclude:: managing_apps/002.php
 
-- **/public/index.php** is the front controller for your webapp; the config
-  path is specified on or about line 32:
+- **/public/index.php** is the front controller for your webapp.
 
   .. literalinclude:: managing_apps/003.php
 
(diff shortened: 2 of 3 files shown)


paulbalandan (approved):

paulbalandan:
Just wondering if we really need to specify in the docs the actual line numbers? IMO this is unnecessary maintenance of tracking the line numbers every time the code there changes.

totoprayogo1916:
Maybe it's better to just remove the line number.

kenjis:
This work is ridiculous. It is better to remove the line numbers.

And now the code is like this:
```php
// Load our paths config file
// This is the line that might need to be changed, depending on your folder structure.
require FCPATH . '../app/Config/Paths.php';
// ^^^ Change this line if you move your application folder
```

```php
// Load our paths config file
// This is the line that might need to be changed, depending on your folder structure.
require FCPATH . '../app/Config/Paths.php';
// ^^^ Change this line if you move your application folder
```

datamweb (approved):

kenjis (approved):
Thank you!

totoprayogo1916:
changes, for $systemDirectory I aborted because it's related to multiple apps

kenjis (approved):

user_guide_src/source/general/managing_apps/002.php
```diff
@@ -1,4 +1,6 @@
 <?php
 
-$pathsConfig = 'app/Config/Paths.php';
+// Load our paths config file
+// This is the line that might need to be changed, depending on your folder structure.
```
paulbalandan, line 4: Sorry, missed this. But should this line be the same with line 6 below?

totoprayogo1916: https://github.com/codeigniter4/CodeIgniter4/blob/0f16cae244d8dd73f3505ec544055c5160362264/spark#L67-L70

https://github.com/codeigniter4/CodeIgniter4/blob/0f16cae244d8dd73f3505ec544055c5160362264/public/index.php#L30-L33

No, it's appropriate because it uses FCPATH.

paulbalandan: Sorry to be not clear. I mean this comment `// This is the line that might need to be changed, depending on your folder structure.` and the other one `// ^^^ Change this line if you move your application folder` both talk about changing the path, "sandwiching" the `require` line in between. Should only one be enough?

MGatner: I think only one comment is sufficient.

totoprayogo1916: PR sent #6389

PR #6390, opened 2022-08-18
Fix: The subquery adds a prefix for the table alias.

iRedds:
**Description**
https://forum.codeigniter.com/showthread.php?tid=82741&pid=399285#pid399285
When creating a query, table aliases are registered. This prevents the prefix being added to the alias.

Due to the fact that aliases are stored in the property of the BaseConnection class, after compiling the subquery using the ``BaseBuilder::geCompiledSelect()`` method, the list of aliases is cleared, which leads to the fact that when compiling the main query, the alias is recognized as a table name and to the alias prefix is added.

This PR prevents resetting alias values.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -577,7 +577,7 @@ class BaseBuilder
     {
         $table = $this->buildSubquery($from, true, $alias);
 
-        $this->trackAliases($table);
+        $this->db->addTableAlias($alias);
         $this->QBFrom[] = $table;
 
         return $this;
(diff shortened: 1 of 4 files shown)


iRedds:
Added fixes for BaseBuilder::fromSubquery().
Some of the changes could be optimized, but this is not possible due to the current design of the classes for working with the database.

kenjis (approved):

kenjis:
Thank you.

PR #6391, opened 2022-08-19
docs: add about route priority

kenjis:
**Description**
Add route priority as it was not clearly stated.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/incoming/routing.rst
+++ b/user_guide_src/source/incoming/routing.rst
@@ -445,7 +445,7 @@ be used when the first parameter is a language string:
 
 .. _routing-priority:
 
-Route processing queue
+Route Processing Queue
 ----------------------
 
 When working with modules, it can be a problem if the routes in the application contain wildcards.
@@ -528,7 +528,7 @@ a valid class/method pair, just like you would show in any route, or a Closure:
     will be returned. See :php:func:`Response::setStatusCode() <setStatusCode>` for
     information on how to set the status code.
 
-Route processing by priority
+Route Processing by Priority
 ============================
 
 Enables or disables processing of the routes queue by priority. Lowering the priority is defined in the route option.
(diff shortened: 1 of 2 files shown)


user_guide_src/source/incoming/routing.rst
```diff
@@ -445,8 +445,17 @@ be used when the first parameter is a language string:
...
-Route processing queue
-----------------------
+Route Priority
+**************
+
+Routes are registered in the routing table in the order in which they are defined, and the first matching route is executed.
+
+.. note:: If exactly the same route (the URI path) is defined more than once, only the route defined first is registered.
+
+You can check the routing table by the :ref:`spark routes <spark-routes>` command.
```
paulbalandan, line 455: I'm not sure if `handlers` is the correct term for the controller actions. Kindly revise if needed.

```suggestion
Routes are registered in the routing table in the order in which they are defined. This means that when a URI is accessed, the first matching route will be executed.

.. note:: If a route (the URI path) is defined more than once with different handlers, only the first defined route is registered.

You can check registered routes in the routing table by running the :ref:`spark routes <spark-routes>` command.
```

kenjis: `handlers` is the correct term. We use it in the `spark routes`.

```
+--------+--------+------------------------------------------+----------------+---------------+
| Method | Route  | Handler                                  | Before Filters | After Filters |
+--------+--------+------------------------------------------+----------------+---------------+
| GET    | /      | \App\Controllers\Home::index             |                | toolbar       |
| CLI    | ci(.*) | \CodeIgniter\CLI\CommandRunner::index/$1 |                |               |
+--------+--------+------------------------------------------+----------------+---------------+
```

paulbalandan (approved):

Issue #4814, opened 2021-06-10
[Forge] Missing: Means of adding indexes on existing tables


MGatner:
The `Forge` class currently may only add indexes during table creation. We need a way to allow Forge (and thus migrations) to add indexes.

neilsf:
+1 for this. I came here to report the same thing.

sclubricants:
https://github.com/codeigniter4/CodeIgniter4/pull/6676 closes this

MGatner:
Beautiful! Thanks so much @sclubricants 🏆

PR #6396, opened 2022-08-20
Fix Sqlite Table::createTable()

sclubricants:
This PR fixes https://github.com/codeigniter4/CodeIgniter4/issues/6205

The indexes need to be checked for missing fields on composite keys.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/SQLite3/Table.php
+++ b/system/Database/SQLite3/Table.php
@@ -240,6 +240,14 @@ class Table
 
         $this->forge->addField($fields);
 
+        $fieldNames = array_keys($fields);
+
+        $this->keys = array_filter(
+            $this->keys,
+            static fn ($index) => count(array_intersect($index['fields'], $fieldNames))
+            === count($index['fields'])
+        );
+
         // Unique/Index keys
         if (is_array($this->keys)) {
             foreach ($this->keys as $key) {
--- a/system/Database/SQLite3/Table.php
+++ b/system/Database/SQLite3/Table.php
@@ -244,8 +244,7 @@ class Table
 
         $this->keys = array_filter(
             $this->keys,
-            static fn ($index) => count(array_intersect($index['fields'], $fieldNames))
-            === count($index['fields'])
+            static fn ($index) => count(array_intersect($index['fields'], $fieldNames)) === count($index['fields'])
         );
 
         // Unique/Index keys
--- a/tests/system/Database/Live/SQLite/AlterTableTest.php
+++ b/tests/system/Database/Live/SQLite/AlterTableTest.php
@@ -189,6 +189,10 @@ final class AlterTableTest extends CIUnitTestCase
         // check that composite index was dropped.
         $this->assertFalse(isset($indexes['actions_category_name']));
 
+        // check that that other keys are present
+        $this->assertTrue(isset($indexes['actions_name']));
+        $this->assertTrue(isset($indexes['actions_created_at']));
+
         $this->forge->dropTable('actions');
     }
 
--- a/tests/system/Database/Live/SQLite/AlterTableTest.php
+++ b/tests/system/Database/Live/SQLite/AlterTableTest.php
@@ -187,11 +187,11 @@ final class AlterTableTest extends CIUnitTestCase
         $indexes = $this->db->getIndexData('actions');
 
         // check that composite index was dropped.
-        $this->assertFalse(isset($indexes['actions_category_name']));
+        $this->assertArrayNotHasKey('actions_category_name', $indexes);
 
         // check that that other keys are present
-        $this->assertTrue(isset($indexes['actions_name']));
-        $this->assertTrue(isset($indexes['actions_created_at']));
+        $this->assertArrayHasKey('actions_name', $indexes);
+        $this->assertArrayHasKey('actions_created_at', $indexes);
 
         $this->forge->dropTable('actions');
     }
--- a/tests/system/Database/Live/SQLite/AlterTableTest.php
+++ b/tests/system/Database/Live/SQLite/AlterTableTest.php
@@ -156,6 +156,42 @@ final class AlterTableTest extends CIUnitTestCase
         $this->assertTrue($result);
     }
 
+    public function testDropColumnDropCompositeKey()
+    {
+        $this->forge->dropTable('actions', true);
+
+        $fields = [
+            'category'   => ['type' => 'varchar', 'constraint' => 63],
+            'name'       => ['type' => 'varchar', 'constraint' => 63],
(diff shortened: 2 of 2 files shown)


sclubricants:
@MGatner check and see if this resolves https://github.com/codeigniter4/CodeIgniter4/issues/6205 for you.

MGatner:
I will test when I'm on desktop. I actually ended up refactoring that whole table out of the repo that had the issue but I can pull an older version of the code that should still have the bug.

sclubricants:
Looking at your code It looks like if you want to restore it to your original table then you will have to add the composite index back.

```php
    private function createActions(): void
    {
        $this->forge = \Config\Database::forge('tests');
        
        $this->forge->dropTable('actions', true);
        // Actions
        $fields = [
            'category'    => ['type' => 'varchar', 'constraint' => 63],
            'name'        => ['type' => 'varchar', 'constraint' => 63],
            'uid'         => ['type' => 'varchar', 'constraint' => 63],
            'class'       => ['type' => 'varchar', 'constraint' => 63, 'null' => true],
            'input'       => ['type' => 'varchar', 'constraint' => 63, 'null' => true],
            'role'        => ['type' => 'varchar', 'constraint' => 63, 'default' => ''],
            'icon'        => ['type' => 'varchar', 'constraint' => 63, 'default' => ''],
            'summary'     => ['type' => 'varchar', 'constraint' => 255, 'default' => ''],
            'description' => ['type' => 'text', 'null' => true],
            'created_at'  => ['type' => 'datetime', 'null' => true],
            'updated_at'  => ['type' => 'datetime', 'null' => true],
            'deleted_at'  => ['type' => 'datetime', 'null' => true],
        ];

        $this->forge->addField('id');
        $this->forge->addField($fields);

        $this->forge->addKey('name');
        $this->forge->addKey('uid');
        $this->forge->addKey(['category', 'name']);
        $this->forge->addKey(['deleted_at', 'id']);
        $this->forge->addKey('created_at');

        $this->forge->createTable('actions');
    }

    public function up(): void
    {
        $this->forge->dropColumn('actions', [
            'description',
            'summary',
            'category',
            'icon',
            'role',
            'class',
        ]);
    }

    public function down(): void
    {
        $this->forge->addColumn('actions', [
            'class'       => ['type' => 'varchar', 'constraint' => 63, 'null' => true],
            'role'        => ['type' => 'varchar', 'constraint' => 63, 'default' => ''],
            'icon'        => ['type' => 'varchar', 'constraint' => 63, 'default' => ''],
            'category'    => ['type' => 'varchar', 'constraint' => 63, 'default' => ''],
            'summary'     => ['type' => 'varchar', 'constraint' => 255],
            'description' => ['type' => 'text', 'constraint' => 255],
        ]);

        // NEED TO ADD BACK COMPOSITE INDEX
        $this->forge->addKey(['category', 'name'])->processIndexes('actions');
    }
```

This would require a new method to `Forge`:

```php
    /**
     * Executes Sql to add or remove indexes
     */
    public function processIndexes(string $table): bool
    {
        if (! empty($this->keys)) {
            $this->fields = array_flip(array_map(
                static fn ($columnName) => $columnName->name,
                $this->db->getFieldData($this->db->DBPrefix . $table)
            ));

            for ($i = 0, $sqls = $this->_processIndexes($this->db->DBPrefix . $table), $c = count($sqls); $i < $c; $i++) {
                if ($this->db->query($sqls[$i]) === false) {
                    return false;
                }
            }
        }

        $this->reset();

        return true;
    }
```

Then of course we would need to test it for all DBMS.

I'm not sure why this method doesn't exist. There is a `dropKey()` method that executes but the `addKey()` method does not. I know its because sometimes the keys are created with tables while others do not. This really ought to be fixed.

MGatner:
I agree in the key methods. I think needing to re-add the key is fine, the important part is that the alteration doesn't cause a database error. If you've tests that with this PR then I will proceed with the review.

sclubricants:
I've found there are a lot of issues related to this. In MySql, if you have a composite index named category_name and then you drop the column category, the index drops the field category and retains name.. and still assumes the index name category_name. So then if you try and add the index category_name back in then it throws an error. Things really get messy when dealing with auto increment primary keys.

I think we need to just leave it up to the developer to properly manage the indexes. In your example you should have dropped any related indexes before dropping the columns.. then add back those indexes after recreating the columns.

One thing I realized though is that there is no method to drop primary keys and no method to create indexes on existing tables. I'm working on a separate PR to add these methods. At least in MySql dropKey() doesn't work with primary key and requires a different SQL statement.

The code in this PR should stop the error you were reporting, however it still doesn't provide a way to add the index back.

system/Database/SQLite3/Table.php
```diff
@@ -240,6 +240,14 @@ protected function createTable()
 
         $this->forge->addField($fields);
 
+        $fieldNames = array_keys($fields);
+
+        $this->keys = array_filter(
+            $this->keys,
+            static fn ($index) => count(array_intersect($index['fields'], $fieldNames))
+            === count($index['fields'])
```
kenjis, line 248: This is too long, but it seems better. 

```suggestion
            static fn ($index) => count(array_intersect($index['fields'], $fieldNames)) === count($index['fields'])
```

kenjis:
> The code in this PR should stop the error you were reporting, however it still doesn't provide a way to add the index back.

Should this PR be closed because this seems jiust a workaround to prevent the error?

sclubricants:
No this PR is still necessary.  It fixes a bug. SQLite uses a Table class unlike the rest of the DBMS. It basically copies the table but on recreate doesn't create fields that were dropped. The problem was that it was trying to create indexes for fields that were dropped.

I'm working on another PR that adds the features to create indexes after table creation. I can add that in here if you would rather. I was going to submit that one to 4.3

kenjis:
> The problem was that it was trying to create indexes for fields that were dropped.

I got it. Can you add test code?

> I'm working on another PR that adds the features to create indexes after table creation. I can add that in here if you would rather. I was going to submit that one to 4.3

#4814 
It is an enhancement. Please send to `4.3`.

MGatner:
All this tracks with what we've experienced around Forge and the different drivers. I'm happy to have this in as a bugfix, and look forward to what other changes you have for these other issues. Hopefully we can get some DB people to review.

MGatner (approved):

sclubricants:
Added `testDropColumnDropCompositeKey()`

MGatner (approved):
That test looks strangely familiar 😂
Thanks @sclubricants this looks good.

kenjis (approved):
Thank you! LGTM!

PR #6397, opened 2022-08-20
feat: autoload helpers

kenjis:
**Description**
- add functionality to autoload helpers

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Autoload.php
+++ b/app/Config/Autoload.php
@@ -85,4 +85,20 @@ class Autoload extends AutoloadConfig
      * @phpstan-var list<string>
      */
     public $files = [];
+
+    /**
+     * -------------------------------------------------------------------
+     * Helpers
+     * -------------------------------------------------------------------
+     * Prototype:
+     * ```
+     *   $helpers = [
+     *       'form',
+     *   ];
+     * ```
+     *
+     * @var string[]
+     * @phpstan-var list<string>
+     */
+    public $helpers = [];
 }
--- a/app/Config/Autoload.php
+++ b/app/Config/Autoload.php
@@ -81,7 +81,8 @@ class Autoload extends AutoloadConfig
      *  ];
      * ```
      *
-     * @var array<int, string>
+     * @var string[]
+     * @phpstan-var list<string>
      */
     public $files = [];
 }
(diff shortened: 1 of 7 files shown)


app/Config/Autoload.php
```diff
@@ -31,12 +31,12 @@ class Autoload extends AutoloadConfig
      * else you will need to modify all of those classes for this to work.
      *
      * Prototype:
-     *```
+     * ```
```
MGatner, line 34: Let's just drop all of these. No need for markdown back-ticks in block comments in a PHP file 🤔
```suggestion
```

system/Autoloader/Autoloader.php
```diff
@@ -71,10 +71,19 @@ class Autoloader
...
      */
     protected $files = [];
 
+    /**
+     * Stores helper list.
+     *
+     * @var string[]
+     * @phpstan-var list<string>
+     */
+    protected $helpers = [];
```
MGatner, line 85: Let's consume our own feature:
```suggestion
    protected $helpers = ['url'];
```
Then use `array_merge()` with the Config and drop the helper call from bootstrap.php

MGatner (changes requested):
Impressive moves on the tests! I have a flow suggestion, and obviously docs, but the rest is looking good.

app/Config/Autoload.php
kenjis: Done.

system/Autoloader/Autoloader.php
kenjis: Done.

kenjis:
Added the docs.

tests/system/Autoloader/AutoloaderTest.php
```diff
@@ -284,29 +294,45 @@ public function testFindsComposerRoutesWithComposerPathNotFound()
...
+    }
+
+    public function testLoadHelpers()
+    {
+        $config            = new Autoload();
+        $config->helpers[] = 'form';
+
+        $loader = new Autoloader();
+        $loader->initialize($config, new Modules());
+
```
MGatner, line 331: Does this need to be `@runInSeparateProcess` to be definitive? One way to check:
```suggestion
        $this->assertFalse(function_exists('form_open'));

```

kenjis: Yes! Done.

MGatner (approved):
Thanks!

PR #6407, opened 2022-08-22
Upsert + Refactor BaseBuilder *Batch() Methods

sclubricants:
This PR is a combination of https://github.com/codeigniter4/CodeIgniter4/pull/6367 and https://github.com/codeigniter4/CodeIgniter4/pull/6373 as well as refactoring all the Batch methods.

This is the bigger picture of where those PR's are headed. You can consider this PR or not. Its hard to break things down much smaller because everything is interdependent. 

I would suggest creating a special branch for upsert and refactor of Batch methods. Perhaps we could merge this PR there instead. This way we all may work together on it and contribute to it. I welcome any help!

This fixes https://github.com/codeigniter4/CodeIgniter4/issues/5674

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


sclubricants:
We must be running an older version of SQLite somewhere.

Capture

kenjis:
This PR is too big (`+2,441 −267`), and the commit is also too big.
It is not acceptable, because it has multiple changes.

> One thing at a time: A pull request should only contain one change.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#branching

The change in one commit is also so large that it is difficult to understand the code. 
I believe that merging this would be an obstacle to future maintenance.
I hope a PR like this size have more than 200 commits.

So I think this PR is just a reference.

kenjis:
Our PR guideline does not mention about commit messages well now.
I put some references for writing good commit messages:

- [Git Best Practices — AFTER Technique - DZone DevOps](https://dzone.com/articles/git-best-practices-after-technique-1)
- [Semantic Commit Messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716)
- [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

sclubricants:
>>I would suggest creating a special branch for upsert and refactor of Batch methods. Perhaps we could merge this PR there instead.

kenjis:
If we create special branch, can you send small PRs to it?

sclubricants:
I can try. A lot of the changes take place simultaneously.

The only way I could properly plan and put it all together was to put it all together at once. There are things shared between upsert and update and there are things shared between upsert and insert. I had to put them all together to normalize these common things. Then refactor the methods that set and execute them.

It would be a lot of piece work to try and do it one by one. I could summarize the major changes though.

I had all the commits at one time but you complained there were too many so I merged them all together.

sclubricants:
There are a few things Id like to discuss with what is here. Once those things are figured out then I can rebuild it piece by piece if necessary.

sclubricants:
The old code was really screwy. Things were being set in all different places. Now there is one method to set data.. setBatch(). Any data that comes in through the other methods is referred to setBatch()

sclubricants:
Create a new branch for this and I will submit it in pieces and explain each piece.

kenjis:
> I had all the commits at one time but you complained there were too many so I merged them all together.

I don't recall complaining about too many PR commits, but maybe there was a problem with my English.

I recommend atomic commits. Also, I would like the overall size of a PR to be small.

PR #6415, opened 2022-08-23
Group list route based on handler

agungsugiarto:

- All bug fixes should be sent to the __"develop"__ branch, this is where the next bug fix version will be developed.
- PRs with any enhancement should be sent to the next minor version branch, e.g. __"4.3"__

**Description**
Group list route based on handler controller, output look like this:
image

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Utilities/Routes.php
+++ b/system/Commands/Utilities/Routes.php
@@ -147,7 +147,7 @@ class Routes extends BaseCommand
         }
 
         // Sort based on controller.
-        usort($tbody, static fn ($handler1, $handler2) => strcmp($handler1[2], $handler2[2]));
+        usort($tbody, static fn ($handler1, $handler2) => strcasecmp($handler1[2], $handler2[2]));
 
         $thead = [
             'Method',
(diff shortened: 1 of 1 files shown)


system/Commands/Utilities/Routes.php
```diff
@@ -146,6 +146,11 @@ public function run(array $params)
             $tbody = [...$tbody, ...$autoRoutes];
         }
 
+        // Sort based on controller.
+        usort($tbody, function ($handler1, $handler2) {
+            return strcmp($handler1[2], $handler2[2]);
```
paulbalandan, line 151: Do you really need case-sensitive sorting? Why not use `strcasecmp`?

agungsugiarto: Already fixed, thank u @paulbalandan

kenjis:
Thank you for sending a PR.

1. Why do you need this?
2. If this is not a bug fix, you should sent a PR to `4.3` branch

kenjis:
I think this should be an option for the `spark routes` command.

kenjis:
Closed by #7015

PR #6414, opened 2022-08-23
Fixed consistency in language strings

tangix:
**Description**
Referencing #6341 
Updated strings in Language/en to improve consistency.

**Checklist:**
- [X] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [X] Conforms to style guide

PR #6418, opened 2022-08-24
docs: add Backward Compatibility Notes

kenjis:
**Description**
- add Backward Compatibility Notes. See https://github.com/codeigniter4/CodeIgniter4/pull/6414#issuecomment-1223897237

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/installation/backward_compatibility_notes.rst
+++ b/user_guide_src/source/installation/backward_compatibility_notes.rst
@@ -1,0 +1,16 @@
+############################
+Backward Compatibility Notes
+############################
+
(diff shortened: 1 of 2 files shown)


paulbalandan:
As a follow up PR perhaps, we should also enumerate what are not covered by the BC promise (as I believe the size is much smaller that what are covered by BC promise).

MGatner (approved):
😍

kenjis:
@paulbalandan Yes. This is a beginning.

PR #6421, opened 2022-08-24
fix: CLIRequest Erros in CLI

### LINKED ISSUE #6419: Bug: Call to undefined method CodeIgniter\HTTP\CLIRequest::getPost() while running the CLI routes
### PHP Version

8.0

### CodeIgniter4 Version

4.2.4

Windows

### Which server did you use?

apache

### Database

MySQL 8.0.27

### What happened?

Running CLI routes is getting an error
image

### Steps to Reproduce

1. Declare any CLI routes 
2. Try to execute it using CMD/Terminal

### Expected Output

It should execute the CLI request instead of the error

### Anything else?

_No response_

kenjis:
**Description**

Ref #6089

- remove `@mixin` in `RequestInterface`
- add missing `get*()` methods in `CLIRequest`
  - getGet()
  - getPost()
  - getPostGet()
  - getGetPost()
- fix filters
  - CSRF
  - Honeypot: fixes #6419
  - InvalidChars

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CLIRequest.php
+++ b/system/HTTP/CLIRequest.php
@@ -226,7 +226,7 @@ class CLIRequest extends Request
      */
     public function getGet($index = null, $filter = null, $flags = null)
     {
-        return null;
+        return $this->returnNullOrEmptyArray($index);
     }
 
     /**
@@ -240,7 +240,7 @@ class CLIRequest extends Request
      */
     public function getPost($index = null, $filter = null, $flags = null)
     {
-        return null;
+        return $this->returnNullOrEmptyArray($index);
     }
 
     /**
@@ -254,7 +254,7 @@ class CLIRequest extends Request
      */
     public function getPostGet($index = null, $filter = null, $flags = null)
     {
-        return null;
+        return $this->returnNullOrEmptyArray($index);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


system/HTTP/CLIRequest.php
```diff
@@ -214,4 +214,60 @@ public function isCLI(): bool
...
+     *
+     * @param array|string|null $index  Index for item to fetch from $_GET.
+     * @param int|null          $filter A filter name to apply.
+     * @param mixed|null        $flags
+     *
+     * @return null
+     */
+    public function getGet($index = null, $filter = null, $flags = null)
+    {
+        return null;
```
MGatner, line 229: I'm not sure about this. I worry that it will hide mistakes. Have you considered throwing an exception? `BadMethodCallException` maybe?

MGatner (commented):
This might need to be split up. Let's get at least one more review in here first. 

I think this is a good change, follow up to the `Services::createRequest()` work.

system/HTTP/CLIRequest.php
kenjis: Now I think `CLIRequest` should have all the methods that `IncomingRequest` has.

The following code does not work in CLI in 4.2.4. But I'm not sure the code should raise an error.
At least in CI3 or CI4.1 it runs without an error.

```php
<?php
namespace App\Controllers;
{
    public function index()
    {
        $this->request->getPost();
    }
}
```

It is true that there is no POST data in CLI, so you should not use `getPost()`.

But we do not know what legacy codes users have.
And It doesn't seem good to have code that used to work without error stop working.

I'm afraid that users will continue to report bugs that the code no longer works when it used to work.

MGatner: If the goal is code compatibility shouldn't it be something like this?
```php
return $index === null ? [] : null;
```

kenjis: Fixed.

MGatner (approved):

MGatner:
Looks good!

PR #6424, opened 2022-08-26
docs: replace mixed in Database doc comments

kenjis:
**Description**
Ref #6310

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


MGatner (approved):
I'm not going to dig through the code on mobile to verify each of these. If you are confident in the types then proceed; if you'd like to wait for a proper review I can do it later on desktop, just re-request.

PR #6426, opened 2022-08-26
docs: update date helper

kenjis:
**Description**
- fix explanation of `now()`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/helpers/date_helper.rst
+++ b/user_guide_src/source/helpers/date_helper.rst
@@ -50,7 +50,7 @@ The following functions are available:
     :returns:    Preformatted HTML select field
     :rtype:    string
 
-    Generates a `select` form field of available timezones (optionally filtered by `$what` and `$country`).
+    Generates a `select` form field of available timezones (optionally filtered by ``$what`` and ``$country``).
     You can supply an option class to apply to the field to make formatting easier, as well as a default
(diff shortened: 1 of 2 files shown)


user_guide_src/source/helpers/date_helper.rst
```diff
@@ -27,17 +29,18 @@ The following functions are available:
...
-    this function over PHP's ``time()`` function.
+    .. note:: It is recommended to use the :doc:`Time <../libraries/time>` class instead. Use ``Time::now()->getTimestamp()`` to get the current UNIX timestamp.
+
+    If a timezone is not provided, it will return the current UNIX timestamp by ``time()``.
 
     .. literalinclude:: date_helper/002.php
 
-    If a timezone is not provided, it will return ``time()`` based on the
-    **time_reference** setting.
+    If any PHP supported timezone is provided, it will return a timestamp that is off by the time difference. It is not the same as the current UNIX timestamp.
```
MGatner, line 38:
```suggestion
    If any PHP supported timezone is provided, it will return a timestamp that is offset by the time difference. It is not the same as the current UNIX timestamp.
```

MGatner (approved):
Small suggestion, looks good.

Issue #4653, opened 2021-05-07
[Forge] missing drop primary key


bczottegem:
- There seems to be missing a method to drop a primary key (which exists for foreign keys) in the Forge class.
- ~~It is also not possible to assign a new primary key to an existing table. One can use addPrimaryKey() for create table operations, but not for alter table.~~ See #4814

kenjis:
Closed by #6488

PR #6430, opened 2022-08-27
Add processIndexes() and dropPrimaryKey() to Forge

sclubricants:
Ref #4814 #4653

This PR adds functionality to create indexes, primary keys, and foreign keys on a table already created. It does this by calling the new function `processIndexes()` instead of `createTable()`.

```php
$forge->addKey(['category', 'name'])
      ->addPrimaryKey('id')
      ->addForeignKey('userid', 'user', 'id')
      ->processIndexes('actions');
```
Also a new method `dropPrimaryKey()` allows to drop the primary key.

There were some differences between DBMS of the naming of indexes. These have been consolidated so each platform follows the same convention.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/ConnectionInterface.php
+++ b/system/Database/ConnectionInterface.php
@@ -153,4 +153,39 @@ interface ConnectionInterface
      * @param string $sql
      */
     public function isWriteType($sql): bool;
+
+    /**
+     * Determines if table exists in database.
+     */
+    public function tableExists(string $tableName, bool $cached = true): bool;
+
+    /**
+     * Protect Identifiers
+     *
+     * This function is used extensively by the Query Builder class, and by
+     * a couple functions in this class.
+     * It takes a column or table name (optionally with an alias) and inserts
+     * the table prefix onto it. Some logic is necessary in order to deal with
+     * column names that include the path. Consider a query like this:
+     *
+     * SELECT hostname.database.table.column AS c FROM hostname.database.table
+     *
+     * Or a query with aliasing:
+     *
+     * SELECT m.member_id, m.member_name FROM members AS m
+     *
+     * Since the column name can include up to four segments (host, DB, table, column)
+     * or also have an alias prefix, we need to do a bit of work to figure this out and
+     * insert the table prefix (if it exists) in the proper position, and escape only
+     * the correct identifiers.
+     *
+     * @param array|string $item
+     * @param bool         $prefixSingle       Prefix a table name with no segments?
+     * @param bool         $protectIdentifiers Protect table or column names?
+     * @param bool         $fieldExists        Supplied $item contains a column name?
+     *
+     * @return array|string
+     * @phpstan-return ($item is array ? array : string)
+     */
+    public function protectIdentifiers($item, bool $prefixSingle = false, ?bool $protectIdentifiers = null, bool $fieldExists = true);
 }
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -406,12 +406,16 @@ class Forge
         $fieldName  = (array) $fieldName;
         $tableField = (array) $tableField;
 
+        // same structure as Connection::_foreignKeyData()
         $this->foreignKeys[] = [
+            'name'           => null,
+            'table'          => null,
             'field'          => $fieldName,
-            'referenceTable' => $tableName,
+            'referenceTable' => $this->db->DBPrefix . $tableName,
             'referenceField' => $tableField,
             'onDelete'       => strtoupper($onDelete),
             'onUpdate'       => strtoupper($onUpdate),
+            'match'          => null,
         ];
 
         return $this;
@@ -1135,7 +1139,7 @@ class Forge
             $nameIndex            = $table . '_' . implode('_', $fkey['field']) . '_foreign';
             $nameIndexFilled      = $this->db->escapeIdentifiers($nameIndex);
             $foreignKeyFilled     = implode(', ', $this->db->escapeIdentifiers($fkey['field']));
-            $referenceTableFilled = $this->db->escapeIdentifiers($this->db->DBPrefix . $fkey['referenceTable']);
+            $referenceTableFilled = $this->db->escapeIdentifiers($fkey['referenceTable']);
             $referenceFieldFilled = implode(', ', $this->db->escapeIdentifiers($fkey['referenceField']));
 
             $sqls[$i] = '';
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -408,14 +408,11 @@ class Forge
 
         // same structure as Connection::_foreignKeyData()
         $this->foreignKeys[] = [
-            'name'           => null,
-            'table'          => null,
             'field'          => $fieldName,
             'referenceTable' => $this->db->DBPrefix . $tableName,
             'referenceField' => $tableField,
             'onDelete'       => strtoupper($onDelete),
             'onUpdate'       => strtoupper($onUpdate),
-            'match'          => null,
         ];
 
         return $this;
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -405,19 +405,6 @@ class Forge
     {
         $fieldName  = (array) $fieldName;
         $tableField = (array) $tableField;
-        $errorNames = [];
-
-        foreach ($fieldName as $name) {
-            if (! isset($this->fields[$name])) {
-                $errorNames[] = $name;
-            }
-        }
-
-        if ($errorNames !== []) {
-            $errorNames[0] = implode(', ', $errorNames);
-
-            throw new DatabaseException(lang('Database.fieldNotExists', $errorNames));
-        }
 
         $this->foreignKeys[] = [
             'field'          => $fieldName,
@@ -452,6 +439,22 @@ class Forge
 
             return false;
         }
+
+        return $this->db->query($sql);
+    }
+
+    /**
+     * Drop Primary Key
+     *
+     * @return bool
+     */
+    public function dropPrimaryKey(string $table)
+    {
+        $sql = sprintf(
+            'ALTER TABLE %s DROP CONSTRAINT %s',
+            $this->db->escapeIdentifiers($this->db->DBPrefix . $table),
+            $this->db->escapeIdentifiers('pk_' . $this->db->DBPrefix . $table),
+        );
 
         return $this->db->query($sql);
     }
@@ -513,7 +516,7 @@ class Forge
 
             // Most databases don't support creating indexes from within the CREATE TABLE statement
             if (! empty($this->keys)) {
-                for ($i = 0, $sqls = $this->_processIndexes($table), $c = count($sqls); $i < $c; $i++) {
+                for ($i = 0, $sqls = $this->_processIndexes($table, true), $c = count($sqls); $i < $c; $i++) {
                     $this->db->query($sqls[$i]);
                 }
             }
@@ -541,10 +544,10 @@ class Forge
         $columns = implode(',', $columns);
 
         $columns .= $this->_processPrimaryKeys($table);
-        $columns .= $this->_processForeignKeys($table);
+        $columns .= current($this->_processForeignKeys($table));
 
         if ($this->createTableKeys === true) {
-            $indexes = $this->_processIndexes($table);
+            $indexes = current($this->_processIndexes($table));
             if (is_string($indexes)) {
                 $columns .= $indexes;
             }
@@ -731,7 +734,8 @@ class Forge
             return false;
         }
 
-        return $this->db->query($sql);
+        // sqlite returns ''
+        return ($sql === '') ? true : $this->db->query($sql);
     }
 
     /**
@@ -990,7 +994,7 @@ class Forge
         }
     }
 
-    protected function _processPrimaryKeys(string $table): string
+    protected function _processPrimaryKeys(string $table, bool $asQuery = false): string
     {
         $sql = '';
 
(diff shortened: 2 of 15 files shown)


kenjis:
The commits are too big. Could you make small, atomic commits?
And please write good commit messages, not just `Fix`.
See https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention

system/Database/ConnectionInterface.php
```diff
@@ -153,4 +153,22 @@ public function callFunction(string $functionName, ...$params);
...
      */
     public function isWriteType($sql): bool;
+
+    /**
+     * Determines if table exists in database.
+     */
+    public function tableExists(string $tableName, bool $cached = true): bool;
+
+    /**
+     * Determines if table exists in database.
```
MGatner, line 163: I think this was a copy-paste error.

MGatner (commented):
Wow @sclubricants this is huge! Very grateful for your attention to the database layer. Please address @kenjis' concerns so we can work with you to review; I will wait for that to do my review too.

sclubricants:
This PR fixes https://github.com/codeigniter4/CodeIgniter4/issues/6453

system/Database/ConnectionInterface.php
```diff
@@ -153,4 +153,39 @@ public function callFunction(string $functionName, ...$params);
      * @param string $sql
      */
     public function isWriteType($sql): bool;
+
```
sclubricants, line 156: I added these because Rector was complaining they were missing.

system/Database/Forge.php
```diff
@@ -405,23 +405,11 @@ public function addForeignKey($fieldName = '', string $tableName = '', $tableFie
     {
         $fieldName  = (array) $fieldName;
         $tableField = (array) $tableField;
```
sclubricants, line 407: I moved this logic to _processForeignKeys(). I can't remember why exactly but it had to do with the new method processIndexes() which the calls _processForeignKeys().

system/Database/Forge.php
```diff
@@ -456,6 +444,22 @@ public function dropKey(string $table, string $keyName)
         return $this->db->query($sql);
     }
```
sclubricants, line 446: Added new method dropPrimaryKey(). The existing methods didn't allow for dropping primary.

system/Database/Forge.php
```diff
@@ -513,7 +517,7 @@ public function createTable(string $table, bool $ifNotExists = false, array $att
 
             // Most databases don't support creating indexes from within the CREATE TABLE statement
             if (! empty($this->keys)) {
```
sclubricants, line 519: Added second parameter for _processIndexes(). with true the method will return stand alone sql statements. More on this later.

system/Database/Forge.php
```diff
@@ -541,10 +545,10 @@ protected function _createTable(string $table, bool $ifNotExists, array $attribu
         $columns = implode(',', $columns);
 
         $columns .= $this->_processPrimaryKeys($table);
```
sclubricants, line 547: I changed _processIndexes() to return an array rather than a string. current() is used to get the first (and only) element of array.

system/Database/Forge.php
```diff
@@ -990,7 +995,7 @@ protected function _attributeAutoIncrement(array &$attributes, array &$field)
         }
     }
```
sclubricants, line 997: Added second parameter $asQuery. This tells the function to return a stand alone query rather than the partial sql used when creating a table. This is used throughout this PR.

system/Database/Forge.php
```diff
@@ -1001,14 +1006,64 @@ protected function _processPrimaryKeys(string $table): string
...
+            }
+            $sql .= 'CONSTRAINT ' . $this->db->escapeIdentifiers('pk_' . $table)
                     . ' PRIMARY KEY(' . implode(', ', $this->db->escapeIdentifiers($this->primaryKeys)) . ')';
         }
 
         return $sql;
     }
 
-    protected function _processIndexes(string $table)
+    /**
```
sclubricants, line 1021: This is the magical method that runs all the indexes just like createTable() does. First we have to load the table fields so the methods work smoothly. Then we process Indexes, then primary keys, and then foreign keys. Some of these methods reset Forge so we have to carry their values and reset them. 
_processIndexes(), _processPrimaryKeys(), and _processForeignKeys() all have the second parameter $asQuery. This retrieves a stand alone query that can be executed outside of the CREATE TABLE statement.

system/Database/Forge.php
```diff
@@ -1041,9 +1096,28 @@ protected function _processIndexes(string $table)
         return $sqls;
     }
```
sclubricants, line 1098: Added second parameter $asQuery. This tells the function to return a stand alone query rather than the partial sql used when creating a table.

sclubricants: Note that instead of returning a string the method now returns an array

system/Database/MySQLi/Connection.php
```diff
@@ -487,7 +487,10 @@ protected function _foreignKeyData(string $table): array
                         tc.TABLE_NAME,
                         kcu.COLUMN_NAME,
                         rc.REFERENCED_TABLE_NAME,
```
sclubricants, line 489: Added Delete and Update Rules and Match option to data returned. This was done to all DBMS

system/Database/MySQLi/Connection.php
```diff
@@ -505,17 +508,33 @@ protected function _foreignKeyData(string $table): array
...
-        $retVal = [];
+        $ind = [];
 
         foreach ($query as $row) {
-            $obj                      = new stdClass();
-            $obj->constraint_name     = $row->CONSTRAINT_NAME;
-            $obj->table_name          = $row->TABLE_NAME;
-            $obj->column_name         = $row->COLUMN_NAME;
-            $obj->foreign_table_name  = $row->REFERENCED_TABLE_NAME;
-            $obj->foreign_column_name = $row->REFERENCED_COLUMN_NAME;
```
sclubricants, line 516: Changed the data structure a bit. There was some inconsistency between DBMS. They have all been standardized.

Instead of returning multiple rows in the case of composite keys only a single row is returned. The column names will be an array of columns used in key/composite key.

The first loop creates the data structure. The second converts it into an object.

system/Database/MySQLi/Forge.php
```diff
@@ -187,11 +187,18 @@ protected function _processColumn(array $field): string
      *
      * @param string $table (ignored)
      */
```
sclubricants, line 189: Added second parameter $asQuery. This tells the function to return a stand alone query rather than the partial sql used when creating a table.

system/Database/MySQLi/Forge.php
```diff
@@ -230,10 +244,25 @@ public function dropKey(string $table, string $keyName)
     {
         $sql = sprintf(
             $this->dropIndexStr,
-            $this->db->escapeIdentifiers($keyName),
+            $this->db->escapeIdentifiers($this->db->DBPrefix . $keyName),
             $this->db->escapeIdentifiers($this->db->DBPrefix . $table),
         );
 
         return $this->db->query($sql);
     }
```
sclubricants, line 252: MySql implementation of new dropPrimaryKey() method.

system/Database/OCI8/Connection.php
```diff
@@ -384,40 +384,59 @@ protected function _indexData(string $table): array
     protected function _foreignKeyData(string $table): array
     {
         $sql = 'SELECT
```
sclubricants, line 386: This query had lots of problems. First I wanted to get the delete rule so I changed some things there. Then the other problem is that the query was improperly joined and was returning duplicate rows. The additional join condition: accu.POSITION = acc.POSITION fixed this problem.

system/Database/OCI8/Forge.php
```diff
@@ -273,9 +274,28 @@ protected function _dropTable(string $table, bool $ifExists, bool $cascade)
         return $sql;
     }
```
sclubricants, line 276: Added second parameter $asQuery. This tells the function to return a stand alone query rather than the partial sql used when creating a table. Also returns array instead of string.

system/Database/Postgre/Connection.php
```diff
@@ -325,35 +325,55 @@ protected function _indexData(string $table): array
      */
     protected function _foreignKeyData(string $table): array
     {
```
sclubricants, line 327: added delete and update rules and match option. Fixed improper join using ordinal_position.

system/Database/Postgre/Forge.php
```diff
@@ -130,7 +130,7 @@ protected function _alterTable(string $alterType, string $table, $field)
     protected function _processColumn(array $field): string
     {
         return $this->db->escapeIdentifiers($field['name'])
```
sclubricants, line 132: This is from another PR: https://github.com/codeigniter4/CodeIgniter4/pull/6430

This was needed to get all the tests working.

system/Database/SQLSRV/Connection.php
```diff
@@ -278,7 +278,10 @@ protected function _foreignKeyData(string $table): array
             . 'OBJECT_NAME (f.parent_object_id) as table_name, '
             . 'COL_NAME(fc.parent_object_id,fc.parent_column_id) column_name, '
             . 'OBJECT_NAME(f.referenced_object_id) foreign_table_name, '
```
sclubricants, line 280: Joined another table to get delete and update rules and match option.

system/Database/SQLSRV/Forge.php
```diff
@@ -144,7 +145,25 @@ protected function _alterTable(string $alterType, string $table, $field)
                 }
             }
 
-            $sql = 'ALTER TABLE ' . $this->db->escapeIdentifiers($this->db->schema) . '.' . $this->db->escapeIdentifiers($table) . ' DROP ';
```
sclubricants, line 147: This is from my other PR: https://github.com/codeigniter4/CodeIgniter4/pull/6430
This allows all my tests to run.

system/Database/SQLSRV/Forge.php
```diff
@@ -218,10 +237,8 @@ protected function _dropIndex(string $table, object $indexData)
 
     /**
      * Process indexes
-     *
-     * @return array|string
      */
```
sclubricants, line 240: Added second parameter $asQuery. This tells the function to return a stand alone query rather than the partial sql used when creating a table. Also returns array instead of string.

system/Database/SQLSRV/Forge.php
```diff
@@ -276,51 +293,66 @@ protected function _processColumn(array $field): string
      *
      * @param string $table Table name
      */
```
sclubricants, line 295: Added second parameter $asQuery. This tells the function to return a stand alone query rather than the partial sql used when creating a table. Also returns array instead of string.

system/Database/SQLite3/Connection.php
```diff
@@ -332,26 +332,53 @@ protected function _foreignKeyData(string $table): array
             return [];
         }
```
sclubricants, line 334: Sqlite for some reason doesn't have an easy way to return the constraint name for foreign keys. The only thing you can do is look at the create table statement and see the name that was used.

The previous implementation returned some odd descriptions as constraint_name. The changes here build the index name as it would have been created with Forge. 

Throughout Forge and with these methods the constraint names have all been normalized. Each DBMS will use the same foreign key constraint names. This was not so before this PR. This made the tests quite a mess.

system/Database/SQLite3/Forge.php
```diff
@@ -159,7 +159,7 @@ protected function _processColumn(array $field): string
     /**
      * Process indexes
      */
```
sclubricants, line 161: Added second parameter $asQuery. This tells the function to return a stand alone query rather than the partial sql used when creating a table. Also returns array instead of string.

system/Database/SQLite3/Table.php
```diff
@@ -200,25 +201,78 @@ public function modifyColumn(array $field)
      *
      * @return Table
      */
-    public function dropForeignKey(string $column)
+    public function dropForeignKey(string $foreignName)
     {
         if (empty($this->foreignKeys)) {
             return $this;
         }
```
sclubricants, line 208: Changed the logic here to work with the new data structure of _foreignKeyData(). This was really ugly.

Interesting that this method was even here because there is no way the current implementation can even create foreign keys.

system/Database/SQLite3/Table.php
```diff
@@ -200,25 +201,78 @@ public function modifyColumn(array $field)
...
-            if (strpos($this->foreignKeys[$i]->constraint_name, $column) !== 0) {
-                continue;
-            }
+        return $this;
+    }
+
+    /**
+     * Add a foreign key
+     *
+     * @return Table
```
sclubricants, line 220: Added this method so foreign keys can be created after table creation and used with new processIndexes() method.

system/Database/SQLite3/Table.php
```diff
@@ -200,25 +201,78 @@ public function modifyColumn(array $field)
...
+
+            $fk[] = $obj;
+        }
+
+        $this->foreignKeys = array_merge($this->foreignKeys, $fk);
+
+        return $this;
+    }
 
-            unset($this->foreignKeys[$i]);
```
sclubricants, line 219: Added dropPrimaryKey() to work with Forge dropPrimaryKey()

system/Database/SQLite3/Table.php
```diff
@@ -273,6 +327,14 @@ protected function createTable()
             }
         }
```
sclubricants, line 329: This part fixes the Table class and actually allows the creation or recreation of foreign keys

tests/system/Database/Live/SQLite/AlterTableTest.php
```diff
@@ -224,11 +224,11 @@ public function testDropForeignKeySuccess()
         $this->createTable('aliens');
 
         $keys = $this->db->getForeignKeyData('aliens');
```
sclubricants, line 226: Fixed to work with new data structure getForeignKeyData() returns.

tests/system/Database/Live/ForgeTest.php
```diff
@@ -34,8 +34,16 @@ final class ForgeTest extends CIUnitTestCase
 
     protected function setUp(): void
     {
-        parent::setUp();
         $this->forge = Database::forge($this->DBGroup);
```
sclubricants, line 37: this helps when testing locally if something goes wrong and the test doesn't finish.

tests/system/Database/Live/ForgeTest.php
```diff
@@ -472,28 +481,16 @@ public function testForeignKey()
         $this->forge->addForeignKey('users_id', 'forge_test_users', 'id', 'CASCADE', 'CASCADE');
 
         $tableName = 'forge_test_invoices';
-        if ($this->db->DBDriver === 'OCI8') {
-            $tableName = 'forge_test_inv';
-        }
 
         $this->forge->createTable($tableName, true, $attributes);
 
         $foreignKeyData = $this->db->getForeignKeyData($tableName);
```
sclubricants, line 488: Now that everything is standardized with constraint names we don't need all the different implementations.

tests/system/Database/Live/ForgeTest.php
```diff
@@ -576,58 +573,32 @@ public function testCompositeForeignKey()
...
+                'constraint' => 11,
+            ],
+            'users_second_id' => [
+                'type'       => 'VARCHAR',
+                'constraint' => 50,
+            ],
         ];
 
         $this->forge->addField($fields);
         $this->forge->addPrimaryKey('id');
```
sclubricants, line 587: Don't need all the different implementations anymore

tests/system/Database/Live/ForgeTest.php
```diff
@@ -1237,4 +1204,168 @@ public function testDropKey()
...
+        ]);
+
+        $this->assertTrue($result);
+
+        // SQLSRV requires dropping default constraint before dropping column
+        $result = $this->forge->dropColumn('user', 'text_with_constraint');
+
+        $this->assertTrue($result);
+    }
+
```
sclubricants, line 1222: This is a really messy set of changes to a table. Add/drop columns and indexes. Lets test all our work and see if we can pass!

system/Database/Postgre/Forge.php
kenjis: If this change from another PR, you can't include it this PR.
One PR for one thing.

kenjis:
@sclubricants I appreciate your efforts.

I don't seem to have communicated well what the problem is.

### Small PR
The most important thing is to make a PR small.
 *Small* I mean is the count of diff lines `+825, −248` is small.
In this case, abs(825) + abs(-248) = 1,073.
I hope it is less than 500.

To make a PR small, we can split a big thing into slices.
It seems this PR includes:
1. processIndexes()
2. dropPrimaryKey()
3. the naming of indexes consolidation

### Atomic Commit
Next, atomic commit. This will help in understanding big PRs.
You put many comments on GitHub, but when you explain one thing, there should be only one commit.
But the first commit has too many your comments. It shows the commit is too big.

See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#committing

sclubricants:
>1. processIndexes()
>2. dropPrimaryKey()
>3. the naming of indexes consolidation

I work on this this weekend and break apart into seperate PRs.

PR #6436, opened 2022-08-27
chore: update Kint to 4.2.0

### LINKED ISSUE #6429: Bug: Long time to debug (with dd) for enum
### PHP Version

8.1

### CodeIgniter4 Version

4.2.4

Windows

### Which server did you use?

apache

### Database

MySQL

### What happened?

If we debugging with dd for enum, especially enum cases or enum class itself, its taking long time . but if we debug enum's name or value is not long.

### Steps to Reproduce

We create an enum
Screenshot (26)
```
namespace MrFrost\HtmlObject\Enums\Attributes;

enum GlobalAttributes: string
{
    case accesskey = "Specifies a shortcut key to activate/focus an element";
    case contenteditable = "Specifies whether the content of an element is editable or not";
    case data = "Used to store custom data private to the page or application";
    case dir = "Specifies the text direction for the content in an element";
    case draggable = "Specifies whether an element is draggable or not";
    case hidden = "Specifies that an element is not yet, or is no longer, relevant";
    case id = "Specifies a unique id for an element";
    case lang = "Specifies the language of the element's content";
    case spellcheck = "Specifies whether the element is to have its spelling and grammar checked or not";
    case style = "Specifies an inline CSS style for an element";
    case tabindex = "Specifies the tabbing order of an element";
    case title = "Specifies extra information about an element";
    case translate = "Specifies whether the content of an element should be translated or not";

    public function getType(): bool
    {
        return match ($this) {
            self::contenteditable, self::draggable, self::hidden, self::spellcheck, self::translate => "bool",
            default => "string"
        };
    }
}
```
And debug the enum cases
Screenshot (27)
`dd(GlobalAttributes::cases());`
is taking long time then we get "Maximum execution time" error
Screenshot (25)

### Expected Output

 Fatal error: Maximum execution time of 120 seconds exceeded in E:\Nova\Latihan\HtmlObject\vendor\kint-php\kint\src\Renderer\RichRenderer.php on line 518

### Anything else?

_No response_

kenjis:
**Description**
Fixes #6429
See https://github.com/kint-php/kint/releases/tag/4.2.0

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


paulbalandan (approved):

paulbalandan:
Wait. Should we pin kint version to `^4.2`?

kenjis:
> Should we pin kint version to ^4.2?

Done.

paulbalandan (approved):

datamweb (approved):

MGatner (commented):
I don't know why I haven't thought of this before but... shouldn't Kint be `require-dev`?

kenjis:
> shouldn't Kint be require-dev?

I don't know why but Kint is tightly coupled with CodeIgniter.
We can't disable it now.

paulbalandan:
> I don't know why I haven't thought of this before but... shouldn't Kint be `require-dev`?

If we are to agree that `dd()` is a testing-only function then we can move it to `require-dev`. But no, `dd()` is used by end users also, not just for the framework's testing use.

PR #6444, opened 2022-08-29
Psalm

MGatner:
**Description**
Adds Psalm with a basic config adapted from DevKit.

I have not included the GitHub Action yet because there are a number of errors (mostly false positives) that need to be addressed first. This either needs to be run in a good container with all the services (e.g. Redis, Memcache) or we need to ignore the missing service constant.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <psalm
-    errorLevel="7"
+    errorLevel="1"
     resolveFromConfigFile="true"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="7"
+    resolveFromConfigFile="true"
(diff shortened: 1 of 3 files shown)


psalm.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="7"
```
kenjis, line 3:
```suggestion
    errorLevel="1"
```

The reported error count is the same.
```
------------------------------
143 errors found
------------------------------
9521 other issues found.
You can display them with --show-info=true
------------------------------
Psalm can automatically fix 3982 of these issues.
Run Psalm again with 
--alter --issues=InvalidReturnType,MissingReturnType,MissingClosureReturnType,InvalidNullableReturnType,InvalidFalsableReturnType,MissingParamType --dry-run
to see what it can fix.
------------------------------
```

MGatner: If the count is the same then why not go with the higher level? That way when we integrate with the pipeline we will catch more issues.

kenjis: > Level 1 is the most strict, level 8 is the most lenient.
https://psalm.dev/docs/running_psalm/error_levels/

kenjis: Ah, 7 is too loose.

kenjis (approved):

PR #6448, opened 2022-08-29
feat: add locale param to `route_to()`

kenjis:
**Description**
Supersedes #6379

- Now you can set locale as the last parameter.

  ```php
  // for route 
  $routes->add('{locale}/path/(:any)/to/(:num)', 'myController::goto/$1/$2', ['as' => 'path-to']);
  
  // the helper should be called like this.
  route_to('path-to', 'string', 13, 'en'); // /en/path/string/to/13
  ```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -911,7 +911,8 @@ if (! function_exists('route_to')) {
      * have a route defined in the routes Config file.
      *
      * @param string     $method    Named route or Controller::method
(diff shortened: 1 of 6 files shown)


MGatner (approved):
Looks good!

PR #6449, opened 2022-08-29
chore: move Kint to `require-dev`

kenjis:
**Description**
Kint is not needed on production environment.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/kint_helper.php
+++ b/system/Helpers/kint_helper.php
@@ -38,13 +38,11 @@ if (! function_exists('dd')) {
     }
 }
 
-if (! function_exists('d')) {
-    if (! class_exists(Kint::class)) {
-        // In case that Kint is not loaded.
-        function d(...$vars)
-        {
-            return 0;
-        }
+if (! function_exists('d') && ! class_exists(Kint::class)) {
+    // In case that Kint is not loaded.
+    function d(...$vars)
+    {
+        return 0;
     }
 }
 
--- a/system/Helpers/kint_helper.php
+++ b/system/Helpers/kint_helper.php
@@ -1,0 +1,68 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+// This helper is autoloaded by CodeIgniter.
+
+if (! function_exists('dd')) {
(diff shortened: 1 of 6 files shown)


system/Helpers/kint_helper.php
```diff
@@ -0,0 +1,66 @@
...
+            Kint::dump(...$vars);
+
+            exit;
+            // @codeCoverageIgnoreEnd
+        }
+    } else {
+        // In case that Kint is not loaded.
+        function dd(...$vars)
+        {
+            return 0;
```
MGatner, line 36: This is potentially dangerous. If someone has been using `dd()` in production then they are likely relying on it terminating. I would propose a closer fallback:
```suggestion
            var_dump(...$vars);
            exit;
```

system/Helpers/kint_helper.php
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+}
+
+if (! function_exists('d') && ! class_exists(Kint::class)) {
+    // In case that Kint is not loaded.
+    function d(...$vars)
+    {
+        return 0;
+    }
```
MGatner, line 46: Why not approximate?
```suggestion
            var_dump(...$vars);
```

system/Helpers/kint_helper.php
```diff
@@ -0,0 +1,66 @@
...
+            Kint::$aliases[] = 'trace';
+            Kint::trace();
+        }
+    } else {
+        // In case that Kint is not loaded.
+        function trace()
+        {
+            return 0;
+        }
+    }
```
MGatner, line 65:
```suggestion
            var_dump(debug_backtrace());
```

MGatner (changes requested):
The `dd` comment is important; others are up for discussion.

system/Helpers/kint_helper.php
kenjis: Kint is disabled in production environment:
https://github.com/codeigniter4/CodeIgniter4/blob/2788a492ea9b2ae8c42873ed01ecf435df8d269c/system/CodeIgniter.php#L188-L190

So `dd()` returns `0`:
https://github.com/codeigniter4/CodeIgniter4/blob/2788a492ea9b2ae8c42873ed01ecf435df8d269c/system/ThirdParty/Kint/Kint.php#L558-L560

MGatner: Ah that makes sense! Thanks

MGatner (approved):

PR #6460, opened 2022-08-31
test: restore changed locale in Time tests

kenjis:
**Description**
A test class should not change the global state.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/I18n/TimeDifferenceTest.php
+++ b/tests/system/I18n/TimeDifferenceTest.php
@@ -19,12 +19,23 @@ use Locale;
  */
 final class TimeDifferenceTest extends CIUnitTestCase
 {
+    private string $currentLocale;
+
     protected function setUp(): void
     {
         parent::setUp();
 
         helper('date');
+
+        $this->currentLocale = Locale::getDefault();
         Locale::setDefault('America/Chicago');
+    }
+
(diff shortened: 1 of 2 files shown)


paulbalandan (approved):

tests/system/I18n/TimeDifferenceTest.php
```diff
@@ -19,14 +19,25 @@
...
+    private string $currentLocale;
+
     protected function setUp(): void
     {
         parent::setUp();
 
         helper('date');
+
+        $this->currentLocale = Locale::getDefault();
         Locale::setDefault('America/Chicago');
```
datamweb, line 31: Parameter locale Is a BCP 47 compliant language tag. Is a **BCP 47** compliant language tag.
I thought the input of the function should be `en-US`, not `America/Chicago`!
https://www.php.net/manual/en/locale.setdefault.php

kenjis: Exactly.

kenjis: Fixed.

datamweb (approved):
Thanks!🥀🥀

PR #6461, opened 2022-08-31
fix: make Time::__toString() database-compatible on any locale

### LINKED ISSUE #6439: make Time object usable with QueryBuilder in all locales

kenjis:
**Description**
Closes #6439
Supersedes #6456

In this PR, `Time::__toString()` no longer depends on locale.
It always returns database-compatible string like `'2016-03-09 12:00:00'`.

In the current implementation, the `__toString()` output will be changed by the locale.
(But it seems it is not documented in the User Guide.)
It causes database query error in a few locale.

Time is localized class, but `__toString()` does not need to return localized value.
If a dev needs localized string, they can call [toDateTimeString()](https://codeigniter4.github.io/CodeIgniter4/libraries/time.html#todatetimestring).

Implicitly changing behavior depending on locale is a bad practice because it confuses developers.
`__toString()` is a standard PHP method, not a localized helper method. 
The  return value should be database-compatible.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/I18n/TimeTest.php
+++ b/tests/system/I18n/TimeTest.php
@@ -17,6 +17,7 @@ use CodeIgniter\Test\CIUnitTestCase;
 use Config\App;
 use DateTime;
 use DateTimeZone;
+use Generator;
 use IntlDateFormatter;
 use Locale;
 
@@ -1137,5 +1138,26 @@ final class TimeTest extends CIUnitTestCase
         $this->assertSame('2017-03-10T12:00:00+09:00', $now);
 
         Locale::setDefault($currentLocale);
+    }
+
+    /**
+     * @dataProvider provideLocales
+     */
+    public function testToStringDoesNotDependOnLocale(string $locale)
+    {
+        Locale::setDefault($locale);
+
+        $time = new Time('2017/03/10 12:00');
+
+        $this->assertSame('2017-03-10 12:00:00', (string) $time);
+    }
+
+    public function provideLocales(): Generator
+    {
+        yield from [
+            ['en'],
+            ['de'],
+            ['fa'],
+        ];
     }
 }
--- a/tests/system/I18n/TimeTest.php
+++ b/tests/system/I18n/TimeTest.php
@@ -1154,6 +1154,7 @@ final class TimeTest extends CIUnitTestCase
         yield from [
             ['en'],
             ['de'],
+            ['ar'],
             ['fa'],
         ];
     }
--- a/tests/system/I18n/TimeTest.php
+++ b/tests/system/I18n/TimeTest.php
@@ -56,7 +56,7 @@ final class TimeTest extends CIUnitTestCase
         );
 
         $time = new Time('', 'America/Chicago');
-        $this->assertSame($formatter->format($time), (string) $time);
+        $this->assertSame($formatter->format($time), $time->toDateTimeString());
     }
 
     public function testTimeWithTimezone()
@@ -72,7 +72,7 @@ final class TimeTest extends CIUnitTestCase
 
         $time = new Time('now', 'Europe/London');
 
-        $this->assertSame($formatter->format($time), (string) $time);
+        $this->assertSame($formatter->format($time), $time->toDateTimeString());
     }
 
     public function testTimeWithTimezoneAndLocale()
@@ -88,7 +88,7 @@ final class TimeTest extends CIUnitTestCase
 
         $time = new Time('now', 'Europe/London', 'fr_FR');
 
-        $this->assertSame($formatter->format($time), (string) $time);
+        $this->assertSame($formatter->format($time), $time->toDateTimeString());
     }
 
     public function testTimeWithDateTimeZone()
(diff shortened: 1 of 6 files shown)


MGatner:
I think this is the way to go. Since that format is very clearly "database DateTime" I expect the intent was always to have the Stringable be database-compatible.

MGatner (approved):

tests/system/I18n/TimeTest.php
```diff
@@ -1139,4 +1139,25 @@ public function testSetTestNowWithFaLocale()
 
         Locale::setDefault($currentLocale);
     }
+
+    /**
+     * @dataProvider provideLocales
+     */
+    public function testToStringDoesNotDependOnLocale(string $locale)
```
datamweb, line 1146:
```suggestion
    public function testToStringDoesNotDependOnLocale(string $locale): void
```

tests/system/I18n/TimeTest.php
```diff
@@ -1139,4 +1139,25 @@ public function testSetTestNowWithFaLocale()
...
+        $time = new Time('2017/03/10 12:00');
+
+        $this->assertSame('2017-03-10 12:00:00', (string) $time);
+    }
+
+    public function provideLocales(): Generator
+    {
+        yield from [
+            ['en'],
+            ['de'],
```
datamweb, line 1159:
```suggestion
            ['de'],
            ['ar'],
```

datamweb (changes requested):

tests/system/I18n/TimeTest.php
kenjis: I did not add `void` because the code in this repository often does not have it.
No other test methods of this class have void.

Probably in the future, when we do a major version upgrade, we will add it with php-cs-fixer.

kenjis:
Added the docs.

tests/system/I18n/TimeTest.php
datamweb: Okay, Thanks👍

user_guide_src/source/changelogs/v4.2.7.rst
```diff
@@ -12,7 +12,7 @@ Release Date: Unreleased
 BREAKING
 ********
 
-none.
+-  ``Time::__toStrong()`` is now locale-independent. It returns database-compatible strings like '2022-09-07 12:00:00' in any locale.
```
paulbalandan, line 15:
```suggestion
-  ``Time::__toString()`` is now locale-independent. It returns database-compatible strings like '2022-09-07 12:00:00' in any locale.
```

user_guide_src/source/installation/upgrade_427.rst
```diff
@@ -0,0 +1,33 @@
...
+- :ref:`Manual Installation Upgrading <installing-manual-upgrading>`
+
+.. contents::
+    :local:
+    :depth: 2
+
+Breaking Changes
+****************
+
+-  ``Time::__toStrong()`` is now locale-independent. It returns database-compatible strings like '2022-09-07 12:00:00' in any locale. Most locales are not affected by this change. But in a few locales like `ar`, `fa`, ``Time::__toStrong()`` (or ``(string) $time`` or implicit casting to a string) no longer returns a localized datetime string. if you want to get a localized datetime string, use :ref:`Time::toDateTimeString() <time-todatetimestring>` instead.
```
paulbalandan, line 18:
```suggestion
-  ``Time::__toString()`` is now locale-independent. It returns database-compatible strings like '2022-09-07 12:00:00' in any locale. Most locales are not affected by this change. But in a few locales like `ar`, `fa`, ``Time::__toString()`` (or ``(string) $time`` or implicit casting to a string) no longer returns a localized datetime string. if you want to get a localized datetime string, use :ref:`Time::toDateTimeString() <time-todatetimestring>` instead.
```

paulbalandan (approved):

tests/system/I18n/TimeTest.php
kenjis: I mean in test classes.
If we add a new public method, we should add parameter/return type.

MGatner (approved):

datamweb (approved):
**Kenjis**, I know you have sent several PRs to solve this problem and finally here we are.
 It works fine now. 
Thanks.

kenjis:
Thank you all!

kenjis:
🎉 
I think this will prevent us from producing broken code that causes a database error only if the locale is fa or ar.

PR #6413, opened 2022-08-23
Set cookie max Expires to 400 days

paulbalandan:
**Description**
Based on [Replacement to RFC 6265](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#section-4.1.2.1), the max Expires SHOULD NOT be greater than 400 days.

> Expires attributes that are greater than the limit MUST be reduced to the limit.

Thus, browsers crop the expires limit to this amount.

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #6463, opened 2022-08-31
docs: deprecate Cookie::withNeverExpiring()

kenjis:
**Description**
See [#6413](https://github.com/codeigniter4/CodeIgniter4/pull/6413#issuecomment-1227060599)
Supersedes #6413

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Cookie/Cookie.php
+++ b/system/Cookie/Cookie.php
@@ -460,8 +460,6 @@ class Cookie implements ArrayAccess, CloneableCookieInterface
     }
 
     /**
-     * {@inheritDoc}
-     *
      * @deprecated See https://github.com/codeigniter4/CodeIgniter4/pull/6413
      */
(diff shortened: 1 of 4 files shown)


system/Cookie/Cookie.php
```diff
@@ -461,6 +461,8 @@ public function withExpired()
 
     /**
      * {@inheritDoc}
```
MGatner, line 463: Please drop this since you're here.

MGatner (commented):
Does this count as `docs:`?

system/Cookie/Cookie.php
kenjis: Done.

MGatner (approved):

PR #6370, opened 2022-08-11
feat: [Forge] set foreign key name manually

stausif:
**Description**
Ref #5075

custom name for foreign key constraint can be added.

```php
$this->forge->addForeignKey("column_name", "table_name", "id", "onDelete", "onUpdate", ["constraint"=>"foreignkey_name"])
```
auto create name for foreign key is still there, a check is added to choose between two.

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

Issue #5075, opened 2021-09-11
[Forge] Setting key name manually for keys (unique, foreign, etc.)


najdanovicivan:
**Describe the bug**
Currently there is no way to manually set the key name. This can lead to issue when there are a lot of columns for which key  should be added as auto generated name will be longer than the database will allow. 
Additionally when tables are being renamed it may be needed to manually set the key name as key with the same name might already exist

**CodeIgniter 4 version**
develop

**Affected module(s)**
Forge

**Expected behavior, and steps to reproduce if appropriate**
To reproduce with mysql just try adding key with a lot of columns
$forge->addKey(['column1','column2','column3','column4','column5','column6',);

We should be able to pass the key name string in the `addKey`, `addPrimaryKey`, `addUniqueKey`, `addForeignKey` methods

**TODO**
- [x] addKey #6552
- [x] addPrimaryKey #6552
- [x] addUniqueKey #6552
- [x] addForeignKey #6468

paulbalandan:
I don't see this as a bug but a current limitation of the Forge which can be improved further. I can try tackling this but that might take some time as I'm currently swamped but if anyone, including you, can open an introductory PR to implement this then that will be a great starter.

Issue #6453, opened 2022-08-29
Bug: Sqlite drops foreign keys when altering table


sclubricants:
### PHP Version

8.0

### CodeIgniter4 Version

4.3

### CodeIgniter4 Installation Method

Git

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

apache

### Database

SQLite3

### What happened?

When performing certain table modifications foreign keys are dropped. This happens because Sqlite uses a Table class to recreate the table when modified. This Table class is missing logic to recreate foreign keys.

### Steps to Reproduce

Test Sqlite with the following:

```php
    public function testDropColumnForeignKey()
    {
        $this->forge->dropTable('forge_test_invoices', true);
        $this->forge->dropTable('forge_test_users', true);

        $attributes = [];

        $this->forge->addField([
            'id' => [
                'type'       => 'INTEGER',
                'constraint' => 11,
            ],
            'second_id' => [
                'type'       => 'VARCHAR',
                'constraint' => 50,
            ],
            'name' => [
                'type'       => 'VARCHAR',
                'constraint' => 255,
            ],
        ]);
        $this->forge->addPrimaryKey(['id', 'second_id']);
        $this->forge->createTable('forge_test_users', true, $attributes);

        $fields = [
            'id' => [
                'type'       => 'INTEGER',
                'constraint' => 11,
            ],
            'name' => [
                'type'       => 'VARCHAR',
                'constraint' => 255,
            ],
            'users_id' => [
                'type'       => 'INTEGER',
                'constraint' => 11,
            ],
            'users_second_id' => [
                'type'       => 'VARCHAR',
                'constraint' => 50,
            ],
            'preferences' => ['type' => 'TEXT'],
        ];

        $this->forge->addField($fields);
        $this->forge->addPrimaryKey('id');
        $this->forge->addForeignKey(['users_id', 'users_second_id'], 'forge_test_users', ['id', 'second_id'], 'CASCADE', 'CASCADE');

        $this->forge->createTable('forge_test_invoices', true, $attributes);

        $foreignKeyData = $this->db->getForeignKeyData('forge_test_invoices');

        $this->assertCount(1, $foreignKeyData);

        $this->forge->dropColumn('forge_test_invoices', 'preferences');

        $foreignKeyData = $this->db->getForeignKeyData('forge_test_invoices');

        $this->assertCount(1, $foreignKeyData);

        $this->forge->dropTable('forge_test_invoices', true);
[...]

PR #6468, opened 2022-09-01
Improve Connection::_foreignKeyData()

sclubricants:
1. `BaseConnection::getForeignKeyData()`
   - Improves the method for retrieving foreign keys.
   - ~~It aligns all DBMS to use the same naming convention.~~
   - It adds some more information such as delete, update, and match rules. 
   - The returned dataset uses a new structure which uses one row per foreign key with an array of fields - useful for composite foreign keys.
2. `Forge::addForeignKey()`
   - Adds a name parameter to manual set foreign key names.
   - See #5075
   - Supersedes #6370 

This is the first part of https://github.com/codeigniter4/CodeIgniter4/pull/6430

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


system/Database/OCI8/Forge.php
```diff
@@ -284,7 +284,7 @@ protected function _processForeignKeys(string $table): string
         ];
 
         foreach ($this->foreignKeys as $fkey) {
-            $nameIndex            = $table . '_' . implode('_', $fkey['field']) . '_fk';
+            $nameIndex            = $table . '_' . implode('_', $fkey['field']) . '_foreign';
```
kenjis, line 287: `_foreign` is too long for OCI8.
Oracle Database 12.1 can have only 30 byte object name.

system/Database/MySQLi/Connection.php
```diff
@@ -482,40 +482,59 @@ protected function _indexData(string $table): array
...
+                    tc.table_schema = ' . $this->escape($this->database) . ' AND
+                    tc.table_name = ' . $this->escape($table);
 
         if (($query = $this->query($sql)) === false) {
             throw new DatabaseException(lang('Database.failGetForeignKeyData'));
         }
         $query = $query->getResultObject();
 
-        $retVal = [];
+        $ind = [];
```
kenjis, line 511: What is `ind`?

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -76,6 +76,7 @@ Others
 - SQLSRV now automatically drops ``DEFAULT`` constraint when using ``Forge::dropColumn()``.
 - Added :ref:`Time::toDatabase() <time-todatabase>` to get a datetime string that can be used with databases regardless of locale.
 - You can set the locale to :php:func:`route_to()` if you pass a locale value as the last parameter.
+- Improved data returned by ``BaseConnection::getForeignKeyData()``. All DBMS return the same structure and use the same naming conventions.
```
kenjis, line 79: This is a breaking change. It changed the public method return value.
We need to add it in `BREAKING` in this changelog, and need to add what to do when upgrading in upgrading guide for v4.3.0.

system/Database/OCI8/Forge.php
kenjis: This change also changed the behavior of `createTable()`. It is also a breaking change.

We need to discuss.

kenjis:
> The returned dataset uses a new structure which uses one row per foreign key with an array of fields

Can you add the structure in the user guide?

kenjis:
@sclubricants Thank you for this PR. This is easy to understand!

The new PHPStan reports new errors. We've fixed. Please rebase to get rid of the errors.

There are a few GitHub Action check failures. Also please check.

system/Database/OCI8/Forge.php
sclubricants: How about:

```php
            $nameIndex = 'fk_' . $table . '_' . implode('_', $fkey['field']);

            if (version_compare($this->db->getVersion(), "18.0.0", "<") && strlen($nameIndex) > 30) {
                if (strlen($nameIndex) > 30) {
                    $nameIndex = substr($nameIndex, 0, 28) . rand(10, 99);
                }
            }

            $nameIndexFilled      = $this->db->escapeIdentifiers($nameIndex);
```

We change them all to fk_

system/Database/MySQLi/Connection.php
sclubricants: It was short for index when index was already taken.. changed the code since then. I'll change them.

system/Database/OCI8/Forge.php
sclubricants: Can't do a prefix.

sclubricants:
There are two ``PHPUnit / PHP 7.4 - MySQLi (pull_request)`` tests. Not sure why. Not sure why one of them is failing. 

**ErrorException: Undefined property: stdClass::$constraint_name**

``Connection::_foreignKeyData()``

```php
        $sql = '
                SELECT
                    tc.constraint_name,
                    ....
                    tc.table_name = ' . $this->escape($table);

        if (($query = $this->query($sql)) === false) {
            throw new DatabaseException(lang('Database.failGetForeignKeyData'));
        }

        $query   = $query->getResultObject();
        $indexes = [];

        foreach ($query as $row) {
            $indexes[$row->constraint_name]['constraint_name']       = $row->constraint_name;
```
The only way I can see this failing is if MySql isn't returning the fields in lower case. I had this problem with Oracle in upper case.

kenjis:
> There are two PHPUnit / PHP 7.4 - MySQLi (pull_request) tests. Not sure why. Not sure why one of them is failing.

The last one is MySQL 8.0.

system/Database/MySQLi/Connection.php
```diff
@@ -483,14 +483,14 @@ protected function _foreignKeyData(string $table): array
     {
```
kenjis, line 483: > Fix MySql _foreignKeyData() SQL upper case 

Please write **why** you changed to uppercase in the commit message.

We can easily know that this commit changed the column names to uppercase,
but it is difficult to know why this change was needed from the commit.

system/Database/BaseConnection.php
```diff
@@ -1565,7 +1565,7 @@ protected function foreignKeyDataToObjects(array $data)
```
kenjis, line 1565: > Using a prefix causes problems when using dropForeignKey().

What's the problems?

kenjis:
https://github.com/codeigniter4/CodeIgniter4/pull/6468/commits/6bfd0bec2c1e71e3c18f8fcd9ad8fc7dad85a577

> Fix MySql _foreignKeyData() SQL upper case 

Please write **why** you changed to uppercase in the commit message.

We can easily know that this commit changed the column names to uppercase,
but it is difficult to know why this change was needed from the commit.

kenjis:
https://github.com/codeigniter4/CodeIgniter4/pull/6468/commits/1b03ee7a7fcdf8c21ea8d9e425cb71dfc5c58a88

> Using a prefix causes problems when using dropForeignKey().

What's the problems?

system/Database/Forge.php
```diff
@@ -1041,20 +1041,27 @@ protected function _processIndexes(string $table)
         return $sqls;
     }
 
-    protected function _processForeignKeys(string $table): string
+    protected function _processForeignKeys(string $table, array $allowActions = [], bool $onUpdate = true, bool $shortName = false): string
```
kenjis, line 1044: Please add doc comments about parameters.

system/Database/OCI8/Forge.php
```diff
@@ -273,31 +273,14 @@ protected function _dropTable(string $table, bool $ifExists, bool $cascade)
...
-
-            $formatSql = ",\n\tCONSTRAINT %s FOREIGN KEY (%s) REFERENCES %s(%s)";
-            $sql .= sprintf($formatSql, $nameIndexFilled, $foreignKeyFilled, $referenceTableFilled, $referenceFieldFilled);
-
-            if ($fkey['onDelete'] !== false && in_array($fkey['onDelete'], $allowActions, true)) {
-                $sql .= ' ON DELETE ' . $fkey['onDelete'];
-            }
+        $allowActions = ['CASCADE', 'SET NULL', 'NO ACTION'];
+
+        if (version_compare($this->db->getVersion(), '18.0.0', '<')) {
```
kenjis, line 280: Why `18.0.0 <`?

system/Database/Forge.php
```diff
@@ -1041,20 +1041,27 @@ protected function _processIndexes(string $table)
...
+            ];
+        }
 
         foreach ($this->foreignKeys as $fkey) {
-            $nameIndex            = $table . '_' . implode('_', $fkey['field']) . '_foreign';
+            $nameIndex = $table . '_' . implode('_', $fkey['field']) . '_foreign';
+
+            if ($shortName === true && strlen($nameIndex) > 30) {
+                $nameIndex = substr($nameIndex, 0, 28) . mt_rand(10, 99);
+            }
```
kenjis, line 1063: Why did you need to add this check? OCI8 does not show error if $nameIndex is too long?

This logic is only for OCI8. It should not be here.

system/Database/OCI8/Forge.php
kenjis: Why can't do prefix?

sclubricants:
> What's the problems?

When you call dropForeignKey() the table parameter gets prefixed by table prefix. I mentioned it elsewhere here..

sclubricants:
>Please write why you changed to uppercase in the commit message.

Good question.. basically cause it worked. Its something I've never seen with MySql. MySql has always returned the same letter case as the sql is typed in my experience but for some reason the server you are running that test on returns upper case no matter what.

kenjis:
> basically cause it worked.

It seems MySQL 8 changed the behavior.

what are the changes in mysql 8 result rowset case? - Stack Overflow
https://stackoverflow.com/questions/54538448/what-are-the-changes-in-mysql-8-result-rowset-case

kenjis:
> When you call dropForeignKey() the table parameter gets prefixed by table prefix.

Oh, we can't drop any key that does not begin with the DB prefix!

https://github.com/codeigniter4/CodeIgniter4/blob/a8072a51bb8c1ea05998d9a045ce61d2c78e32bc/system/Database/Forge.php#L466-L470

system/Database/OCI8/Forge.php
sclubricants: In version 18 and on there isn't a 30byte limit on key name. It seems the version before 18 is 12 but i'm not totally sure. I just verified that in 18 the limit was gone.

system/Database/Forge.php
sclubricants: I moved it here and added $shortName parameter. This got rid of duplicate code. In OCI8 it calls parent:: with the $shortName if version less than 18

sclubricants: In OCI8:

```php
    protected function _processForeignKeys(string $table, array $allowActions = [], bool $onUpdate = false, bool $shortName = false): string
    {
        $allowActions = ['CASCADE', 'SET NULL', 'NO ACTION'];

        if (version_compare($this->db->getVersion(), '18.0.0', '<')) {
            $shortName = true;
        }

        return parent::_processForeignKeys($table, $allowActions, $onUpdate, $shortName);
    }
```

Note: OCI8 does not have any actions on update either.

system/Database/OCI8/Forge.php
sclubricants: It looks like maybe even back to 12.2. But it depends on the compatibility setting.

https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/Database-Object-Names-and-Qualifiers.html#GUID-75337742-67FD-4EC0-985F-741C93D918DA

>The maximum length of identifier names depends on the value of the COMPATIBLE initialization parameter.
>If COMPATIBLE is set to a value of 12.2 or higher, then names must be from 1 to 128 bytes
>If COMPATIBLE is set to a value lower than 12.2, then names must be from 1 to 30 bytes long

sclubricants: https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/Database-Object-Names-and-Qualifiers.html#GUID-75337742-67FD-4EC0-985F-741C93D918DA

sclubricants: So use 12.2?

sclubricants:
>Oh, we can't drop any key that does not begin with the DB prefix!

We definitely want to have the prefix. Someone might have a different connection on same database with duplicates of tables but with different prefixes. This way we avoid name collision.

kenjis:
> We definitely want to have the prefix. Someone might have a different connection on same database with duplicates of tables but with different prefixes. This way we avoid name collision.

Do we need the prefix for foreign key name?

system/Database/Forge.php
```diff
@@ -1041,20 +1041,34 @@ protected function _processIndexes(string $table)
         return $sqls;
     }
 
-    protected function _processForeignKeys(string $table): string
+    /**
+     * Generates SQL to process foreign keys
+     *
+     * @param array $allowActions Allows child to override default value
+     * @param bool  $onUpdate     Allows child to avoid setting onUpdate rule (Oracle)
+     * @param bool  $shortName    Allows child to limit key name to 30 byte (Oracle)
```
kenjis, line 1049: If the two parameters are used only for Oracle, it seems we don't need to have them in this parent class.

system/Database/Forge.php
kenjis: Sorry, I don't get what you say.

The logic is obviously only for Oracle, so I think it should be in the OCI8 class only.

What if we want to add logic for MySQL that checks less than 65 characters?

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -76,6 +78,7 @@ Others
...
```
kenjis, line 81: Can you add a link to the detail page?

system/Database/Forge.php
sclubricants: It was a means of reusing code. The duplicate Code Detection was complaining. I suppose you could create another method instead but it would look just about like this one.

system/Database/Forge.php
sclubricants: I suppose we can pass a parameter of max characters and apply it to all of them.

kenjis: Personally I hope an error occurs if the length is more than the database limit.

system/Database/Forge.php
kenjis: > The duplicate Code Detection was complaining.

If it is an only issue, we can suppress the PHPCPD error.

There is no need to force premature generalizations, as they will yield poor results.
Adding unnecessary parameters to a public method is undesirable. If it is necessary for all DBs, it is fine, though.

sclubricants: reverted

system/Database/OCI8/Forge.php
```diff
@@ -273,18 +273,22 @@ protected function _dropTable(string $table, bool $ifExists, bool $cascade)
...
-        $allowActions = [
-            'CASCADE',
-            'SET NULL',
-            'NO ACTION',
-        ];
+        $sql = '';
 
         foreach ($this->foreignKeys as $fkey) {
-            $nameIndex            = $table . '_' . implode('_', $fkey['field']) . '_fk';
+            $nameIndex = $table . '_' . implode('_', $fkey['field']) . '_foreign';
```
kenjis, line 286: This is a breaking change. Please add this in the changelog.

system/Database/OCI8/Forge.php
```diff
@@ -273,18 +273,22 @@ protected function _dropTable(string $table, bool $ifExists, bool $cascade)
...
+
+            if (version_compare($this->db->getVersion(), '12.2.0', '<') && strlen($nameIndex) > 30) {
+                $nameIndex = substr($nameIndex, 0, 28) . mt_rand(10, 99);
+            }
```
kenjis, line 290: Do you really need this?
I guess it is better not to do like this.
If you have a long table name, and mt_rand() returns the same value,
$nameIndex would be the same name.

sclubricants: >Personally I hope an error occurs if the length is more than the database limit.

This isn't helpful because then Forge could never create the index.

mt_rand() is just an extra insurance policy. Its possible to return same number but highly unlikely that this happens and there is also a collision of names.

Ideally we add the ability to specify the names manually. I know there are some issues related to this. I planned to take a look at doing this in the future. Hopefully before the release of 4.3 we can add this and then get rid of this piece of code. But for now this should keep things working.

kenjis: We cannot create an index now with Forge if the index name length exceeds the database limit.
So it is not a problem.  It is the same as the current behavior.

This check is a breaking change. And if we remove this check, it will be another breaking change.
4.3 will  be released soon.

I don't like naming with random number. It is very bad naming. Humans can't get it.
If you plan to get rid of the code later, I hope you don't add it.

If you have the need of the code, please customize your own Forge.
I do not see why this code must be added.

sclubricants: >I do not see why this code must be added.

Because under certain circumstances a user of Oracle won't be able to create indexes without it. It would likely rarely get invoked but in that special case it wouldn't be a road block. Even for older versions the code isn't invoked unless its certain to fail. You can remove it.. then the developer will just have to manually create the index without using Forge.

sclubricants:
```php
            if (version_compare($this->db->getVersion(), '12.2.0', '<') && strlen($nameIndex) > 30) {
                $nameIndex = uniqid();
            }
```

sclubricants: >This check is a breaking change. And if we remove this check, it will be another breaking change.

This was broken before anything was done here.

sclubricants: It doesn't appear to me that it has ever been documented what the names of keys will be, Honestly If I were working with keys I would look them up and identify them by the tables and columns used.

kenjis: > This was broken before anything was done here.

What do you mean? Before this PR, too long index name causes an error, but this PR does not cause an error and silently changes the naming convention. I think this is a breaking change.
If my understanding is incorrect, please point it out.

kenjis: Devs can create any indexes with `$db->query()`.
So it is not crucial that Forge cannot create indexes in some cases.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -12,6 +12,9 @@ Release Date: Unreleased
 BREAKING
 ********
 
+- The data structure returned by ``BaseConnection::getForeignKeyData()`` has been changed.
+- The naming of foreign keys in OCI8 has changed. The previous suffix of ``_fk`` has changed to ``_foreign``.
```
kenjis, line 16: `_foreign` is a bit long. 
Unique key would be `_u` (See #6516).

We will make some breaking changes around the database code in 4.3. 
Why don't we change it to `_fk` instead?
@MGatner @michalsn @sclubricants

system/Database/OCI8/Forge.php
sclubricants: Ok, if you would rather it create an error then I'll remove the code. I feel that the index name is rather inconsequential to the developer. It does help out the framework though to have consistent names for testing. I feel that having forge working and not throwing an error is more important then what name is given. The documentation makes no express or implied warranties as to the name that will be used but it is implied that it will create a foreign key. By allowing the error you are breeching the implied warranty.

This was broken before because a naming scheme was chosen and put into use that could inevitably cause errors in certain circumstances. This is true not for just OCI8 but the rest as well. One of the challenges of having the framework name the constraints is ensuring unique names are generated. A method was chosen that insured this outcome but unfortunately it isn't compatible all the time. What my code does is try to stick to the naming scheme but ultimately insure a unique name is applied. 

I would have changed them all to a suffix of _fk instead of _foreign but this would have created more changes then the alternative. Oracle is probably the least used database of them all and foreign keys are probably the least used indexes so hopefully the change doesn't cause too many issues.

SQLite doesn't even assign a name to the foreign keys. Probably because you can't do anything with them other than drop and create a new table. This PR however assigns them a name based on the naming scheme which allows us to handle them as if they are named. This makes it easier to drop and add them.

Let me know if you just want to drop this bit of code and allow the error to occur.

user_guide_src/source/changelogs/v4.3.0.rst
sclubricants: I would favor _fk. However I would prefer not to make any guarantees.

system/Database/OCI8/Forge.php
kenjis: The random name is not consistent.
In any cases, naming is very important. It is a core of abstraction.
Bad naming should be always avoided.

I oppose the code in this PR. It is too adhoc. I cannot accept it. Please remove it.
If you want to try to include such code to avoid errors into CI4, send another PR only for it.
It is a big issue. We need more discussion.

user_guide_src/source/changelogs/v4.3.0.rst
sclubricants: > What do you mean by guarantees? 
I mean provide documentation that assures this is the name they are going to get. All DBMS have limits on the length of names so it may not always be possible to provide such name. 

Its best practice for the developer to lookup the constraint names on a table and use the names provided by lookup.. unless they have set the name themselves.

>If the CONSTRAINT symbol clause is not defined, or a symbol is not included following the CONSTRAINT keyword:
>For InnoDB tables, a constraint name is generated automatically.
>For NDB tables, the FOREIGN KEY index_name value is used, if defined. Otherwise, a constraint name is generated automatically.

https://dev.mysql.com/doc/refman/5.6/en/create-table-foreign-keys.html#foreign-key-identifiers

A name isn't even required. The generated names are just a unique random string.

I think your too hung up on the names.

https://stackoverflow.com/questions/2240929/what-is-a-proper-naming-convention-for-mysql-fks

system/Database/OCI8/Forge.php
sclubricants: >In any cases, naming is very important. It is a core of abstraction.

Well this PR tries to give them a consistent name. And in the event it cannot provide exactly it comes as close to it at possible. 

Do the math. You have 30 characters to create a unique name. the sum of all the tables, columns and database prefix is more than 30 characters. 

And if the developer is allowed to set their own names then that doesn't further your cause.

user_guide_src/source/changelogs/v4.3.0.rst
sclubricants: >So, to specify a named constraint, use the key word CONSTRAINT followed by an identifier followed by the constraint definition. (If you don't specify a constraint name in this way, the system chooses a name for you.)

https://www.postgresql.org/docs/current/ddl-constraints.html

system/Database/OCI8/Forge.php
kenjis: You have a point. It seems changing to `_foreign` is not good for Oracle 12.1.
It is better to keep `_fk` for  Oracle 12.1.

sclubricants: Well to be consistent Id change them all to _fk. But that still doesn't solve the issue where the name is too long. In some cases a different scheme is required one way or another. 

Its a well documented issue. Oracle makes having a table column based naming scheme impossible sometimes.

https://stackoverflow.com/questions/199498/foreign-key-naming-scheme

sclubricants: Looking through my databases it looks like my longest table name is 37 characters.

user_guide_src/source/changelogs/v4.3.0.rst
MGatner: Shorter seems bette to me. The changelog reads like it was already `_fk` though? See if @codeigniter4/database-team has any opinions.

visserrobin: I prefer _fk. I use '_fk' to define foreign keys in my databases, never '_foreign'.

kenjis: In the current implementaion, only OCI8 uses `_fk`, others use `_foreign`.

sclubricants: I run this on my db:

```SQL
SELECT
COUNT(DISTINCT t.TABLE_NAME) AS TABLE_COUNT,
ROUND(AVG(LENGTH(t.TABLE_NAME))) AS AVG_LENTH_TABLE,
MAX(LENGTH(t.TABLE_NAME)) AS MAX_LENGTH_TABLE,
c.COLUMN_COUNT,
c.AVG_LENTH_COLUMN,
c.MAX_LENGTH_COLUMN
FROM
information_schema.`TABLES` AS t
INNER JOIN (SELECT
            COUNT(DISTINCT CONCAT(t.TABLE_NAME,c.COLUMN_NAME)) AS COLUMN_COUNT,
            ROUND(AVG(LENGTH(c.COLUMN_NAME))) AS AVG_LENTH_COLUMN,
            MAX(LENGTH(c.COLUMN_NAME)) AS MAX_LENGTH_COLUMN
            FROM
            information_schema.`TABLES` AS t
            INNER JOIN information_schema.`COLUMNS` AS c ON c.TABLE_CATALOG = t.TABLE_CATALOG AND c.TABLE_SCHEMA = t.TABLE_SCHEMA AND c.TABLE_NAME = t.TABLE_NAME
            WHERE
            t.TABLE_TYPE NOT IN ('SYSTEM VIEW') AND
            t.TABLE_SCHEMA NOT IN ('mysql', 'performance_schema')) c
WHERE
t.TABLE_TYPE NOT IN ('SYSTEM VIEW') AND
t.TABLE_SCHEMA NOT IN ('mysql', 'performance_schema')
```
Results:

```
TABLE_COUNT       140
AVG_LENTH_TABLE   15
MAX_LENGTH_TABLE  37

COLUMN_COUNT      3132
AVG_LENTH_COLUMN  12
MAX_LENGTH_COLUMN 43
```

So:

```
prefix: 2
table:  15
column: 12
suffix: 2
------------
total:  31
```

So based on my averages I would fail the naming scheme even with the shorter `_fk` even with only a single column.

Using `_fk` would be better then `_foreign` but we still need to have an alternative when the name exceeds the maximum allowed.

system/Database/OCI8/Forge.php
```diff
@@ -273,18 +273,22 @@ protected function _dropTable(string $table, bool $ifExists, bool $cascade)
...
-        ];
+        $sql = '';
 
         foreach ($this->foreignKeys as $fkey) {
-            $nameIndex            = $table . '_' . implode('_', $fkey['field']) . '_fk';
+            $nameIndex = $table . '_' . implode('_', $fkey['field']) . '_fk';
+
+            if (version_compare($this->db->getVersion(), '12.2.0', '<') && strlen($nameIndex) > 30) {
+                $nameIndex = substr($nameIndex, 0, 28) . mt_rand(10, 99);
+            }
```
kenjis, line 290: This naming scheme makes it impossible to share exactly the same table schema with the DB migration.
It is not a good practice and the method of naming using random strings is not appropriate for CodeIgniter.

user_guide_src/source/changelogs/v4.3.0.rst
michalsn: Changing `_foreign` to `_fk` is more tricky, since it may be quite a big problem for some use cases.

I would rather follow our current convention and change everything to `_foreign` - simply because it will generate fewer problems for developers.

But I won't be a blocker and if the majority of you guys think we should change it then I'm okay with it too. We just have to put it under the BC break list.

sclubricants: Perhaps if we do change to _fk it would be better to allow manually setting the constraint name. This would make the transition easier because you could define the name as the old convention of _foreign.

michalsn: Yes, I would be in favor of enabling the possibility to set the constraint name manually. I believe it would simplify a lot of things.

kenjis: > Changing _foreign to _fk is more tricky, since it may be quite a big problem for some use cases.

Why tricky? What use cases will get quite a big problem?

sclubricants:
FYI, if we allow setting the name manually this won't apply to SQLite. SQLite does not really support naming the constraint. You can name it when you create table but you cannot retrieve the name with the foreign keys. My guess this is because you can't drop or alter the foreign keys so it doesn't need a name. Allowing people to specify a name should through an error. They will need to use the generated name that we assign to it.

user_guide_src/source/changelogs/v4.3.0.rst
michalsn: > Why tricky? What use cases will get quite a big problem?

With this change, we're forcing every existing project to change the constraint names. If I don't make a change in my current project, there are at least 2 scenarios where it might cause problems:

I have an existing application in production, and in my development environment, I fire the command:
```
php spark migrate:refresh
```
Now I'm screwed. Different constraints in production and development. Good luck to me with making changes now.

Second scenario. A new developer comes to the team, configures his environment from scratch, and now he has completely different constraint names in his dev and production environment. 

If this change would be necessary because of a bug or were introduced in v5, I wouldn't be opposed. But making such a change in a minor version is not responsible (from my perspective). It seems like we are making this change because we feel like it.

michalsn:
> FYI, if we allow setting the name manually this won't apply to SQLite. SQLite does not really support naming the constraint.

Yes, I'm aware of this. Since we had never any controls over constraint names in SQLite I don't recognize it as a problem. Adding a friendly note in the user guide about SQLite should be enough.

I was wondering if throwing an error when we define a name in SQLite would be a good choice... simply from a practical perspective. SQLite is frequently used during tests in smaller projects and throwing such an error may be frustrating for some use cases (tests on SQLite3 and production on MySQL). But probably throwing an error is the right thing to do here.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -46,6 +46,8 @@ Others
 - ``CITestStreamFilter::$buffer = ''`` no longer causes the filter to be registered to listen for streams. Now there
   is a ``CITestStreamFilter::registration()`` method for this. See :ref:`upgrade-430-stream-filter` for details.
 - ``Forge::addUniqueKey()`` now applies ``_u`` suffix to the constraint name. This was done to ensure a unique index name.
+- The data structure returned by ``BaseConnection::getForeignKeyData()`` has been changed.
+- The naming of foreign keys in OCI8 has changed. The previous suffix of ``_fk`` has changed to ``_foreign``.
```
michalsn, line 50: In the current code in this PR I see quite different behavior. This description should be updated.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -110,6 +112,8 @@ Database
 - ``BaseConnection::escape()`` now excludes the ``RawSql`` data type. This allows passing SQL strings into data.
 - The new method ``Forge::dropPrimaryKey()`` allows dropping the primary key on a table. See :ref:`dropping-a-primary-key`.
 - Improved the SQL structure for ``Builder::updateBatch()``. See :ref:`update-batch` for the details.
+- Improved data returned by ``BaseConnection::getForeignKeyData()``. All DBMS return the same structure and use the same naming conventions.
+- ``Forge::addForeignKey()`` now includes a name parameter to manuall set foreign key names.
```
michalsn, line 116:
```suggestion
- ``Forge::addForeignKey()`` now includes a name parameter to manual set foreign key names. Not supported in SQLite3.
```

user_guide_src/source/dbmgmt/forge.rst
```diff
@@ -323,13 +323,14 @@ Class Reference
...
 
-    .. php:method:: addForeignKey($fieldName, $tableName, $tableField[, $onUpdate = '', $onDelete = ''])
+    .. php:method:: addForeignKey($fieldName, $tableName, $tableField[, $onUpdate = '', $onDelete = '', $fkName = ''])
 
         :param    string|string[]    $fieldName: Name of a key field or an array of fields
         :param    string    $tableName: Name of a parent table
         :param    string|string[]    $tableField: Name of a parent table field or an array of fields
         :param    string    $onUpdate: Desired action for the "on update"
         :param    string    $onDelete: Desired action for the "on delete"
+        :param    string    $fkName: Name of foreign key
```
michalsn, line 333: It would be nice to have a note near this method description about one not supported driver.

user_guide_src/source/changelogs/v4.3.0.rst
sclubricants: Ya, thought I had changed that.. fixed it now.

user_guide_src/source/dbmgmt/forge.rst
sclubricants: Done

user_guide_src/source/changelogs/v4.3.0.rst
sclubricants: If you didn't set the constraint name you probably ought not depend on a name. I would hope they would look up the constraint names in order to drop them.

The default behavior of most databases is to give them a short description with a unique code. This is if you don't define the name yourself.

kenjis: I agreed with michalsn. There is no compelling reason why the foreign key suffix change has to be done in the minor version.

sclubricants: Well I have `_processForeignKeys()` all factored into a single method now.

Let me make the point though that as Forge is right now there is no way to add a foreign key to an existing table. So it would be less likely that anyone is adding/removing in migrations. If we add this ability which I intend to do after this PR then perhaps these sorts of migrations will become more common and therefore become a bigger headache later.

Another thing is use of foreign keys is fairly rare. I don't think I use any of them in all of my databases. I used them years ago but found them more of a headache than useful. This is probably one of the reasons we haven't heard of much complaint about the lack of features using foreign keys. Adding and removing foreign keys is tricky business so its not done very often.

Also, being that there are major changes in this area already I think this is the best time to make a change if you are going to.

sclubricants: We changed the suffix for unique indexes in 4.3 already.

michalsn: You know, it's not even about the Forge. The point here is not to disrupt the consistency we have now. We may like the current constraint naming convention or not. It doesn't matter. The question is whether this change will do more harm than good. And IMO, it will not do any good for existing projects.

Saying that people are not using foreign keys much is arbitrary. But it may depend on the project we're working on.
Also, support in Forge is not an indicator of whether something is used. We can always use a simple query to run the SQL.

Sadly, not every change we want can be delivered immediately. Hundreds of developers use this framework, and their projects rely on the code we all deliver. Making changes is always tempting, but we need to remember the consequences.

michalsn: > We changed the suffix for unique indexes in 4.3 already.

Well... IMO it should not happen. I will search for the issue related to this.

sclubricants: >You know, it's not even about the Forge.

I get your point but I'm thinking more of continuity between DBMS. If someone builds an app and want it to be compatible with different DBMS then things need to be consistent. Here the Oracle driver was added to the rest and the developer came across an issue specific to Oracle and decided to break the consistency.

michalsn: And I get your point, and it's valid. But... how often do we change the database engine during the project lifetime? Pretty close to never (yes, I took this data right from my ass 😅  ...sorry it was a long day).

I'm more focused on changes that potentially will make life harder for many developers than those that will cause problems for a few. 

Don't get me wrong. I want to see the changes you're proposing but not in a minor release.

sclubricants: Not to keep arguing.. but imagine someone who develops an application to check and manage email. They might want to have their application play with a number of databases for people with different environments.

I'll revert the names back to the way they were.. but we keep the option to set our own names. This will at least make the tests simpler.

michalsn: > Not to keep arguing.

I don't feel like we argue. Seems like we are all on the same page. The discussion is about whether or not it's the right time to introduce some of these changes.

> but imagine someone who develops an application to check and manage email. They might want to have their application play with a number of databases for people with different environments.

Yes, that's a valid point. But to be fair, it's not that Oracle introduced this. We had an SQLite3 driver that behaved differently almost from the beginning.

The question is, what is the better approach for the existing projects? Having an additional line of code in the migration file to handle this situation or to force everyone to make changes in their databases?

For the minor release, I will choose the first option.

> I'll revert the names back to the way they were.. but we keep the option to set our own names. This will at least make the tests simpler.

Yes, the option to set our own names should stay.

Also... I know it might sound different, but I appreciate your contribution to the project and looking forward to more.

sclubricants: Ok, got it reverted. All DBMS Except oracle use _foreign. Oracle uses _fk.

user_guide_src/source/dbmgmt/forge/012.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 $forge->addForeignKey('users_id', 'users', 'id');
-// gives CONSTRAINT `TABLENAME_users_foreign` FOREIGN KEY(`users_id`) REFERENCES `users`(`id`)
+// gives CONSTRAINT `TABLENAME_users_id_fk` FOREIGN KEY(`users_id`) REFERENCES `users`(`id`)
```
michalsn, line 4: Seems not correct.

user_guide_src/source/dbmgmt/forge/012.php
```diff
@@ -1,7 +1,7 @@
...
 
 $forge->addForeignKey(['users_id', 'users_name'], 'users', ['id', 'name']);
-// gives CONSTRAINT `TABLENAME_users_foreign` FOREIGN KEY(`users_id`, `users_name`) REFERENCES `users`(`id`, `name`)
+// gives CONSTRAINT `TABLENAME_users_id_fk` FOREIGN KEY(`users_id`, `users_name`) REFERENCES `users`(`id`, `name`)
```
michalsn, line 7: Seems not correct.

user_guide_src/source/installation/upgrade_430.rst
```diff
@@ -117,6 +117,34 @@ need to use::
 
 Or use the trait ``CodeIgniter\Test\StreamFilterTrait``. See :ref:`testing-cli-output`.
 
+Foreign Key Data
+=====================================================
+
+- The data structure returned by ``BaseConnection::getForeignKeyData()`` has been changed.
+  You will need to adjust any code depending on this method to use the new structure.
+- The Foreign key naming convention has been changed for all DBMS except Oracle. All DBMS
+  use the same convention now. Oracle version 12 is truncated to 30 bytes.
```
michalsn, line 126:
```suggestion
```

michalsn (changes requested):

user_guide_src/source/database/metadata.rst
```diff
@@ -136,6 +138,4 @@ Usage example:
 
 .. literalinclude:: metadata/009.php
 
-The object fields may be unique to the database you are using. For instance, SQLite3 does
-not return data on column names, but has the additional *sequence* field for compound
-foreign key definitions.
+Foreign keys use the naming convention ``tableprefix_table_column1_column2_foreign``.
```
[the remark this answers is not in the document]
sclubricants: Made a note

user_guide_src/source/dbmgmt/forge/012.php
sclubricants: Thought I had all these fixed.

michalsn (approved):
Seems ready... nice improvements. Thank you for your work!

system/Database/SQLite3/Table.php
```diff
@@ -214,23 +214,14 @@ public function dropPrimaryKey(): Table
      *
      * @return Table
      */
-    public function dropForeignKey(string $column)
+    public function dropForeignKey(string $foreignName)
```
kenjis, line 217: The parameter name has been changed.
Is this a breaking change or not?

sclubricants: I suppose it is.. especially for anyone who might of extended the class. I'll make mention in the change log.

The functionality of this in the Table class was very marginal. In a lot of cases it wasn't even working at all. See https://github.com/codeigniter4/CodeIgniter4/issues/6453

With the new structure of _foreignKeyData() things are starting to function much better.. especially SQLite.

kenjis: Okay, thank you.

We are discussing parameter name changing is a breaking change or not in our project.
But at least, we should add the change in the changelog.

Extended class can change the parameter name.
https://3v4l.org/l8Xf5

kenjis:
@sclubricants Please rebase to resolve conflicts.
It seems ready! Thank you for your cooperation.

kenjis (approved):

kenjis:
@sclubricants @michalsn Thank you!
This is a great improvement.

MGatner:
@kenjis said it best! Thanks to everyone for the work on this 🥳

PR #6472, opened 2022-09-02
Fix fatal error gets turned to `0` severity on shutdown handler

### LINKED ISSUE #6454: Bug: Shutdown handler diminishes fatal error's severity to a measly value of `0`
### PHP Version

8.1

### CodeIgniter4 Version

latest develop (4.2.x)

Git

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

It seems when the registered shutdown handler is called when there is a fatal error, it reduces the severity of the error to `0` (zero).

### Steps to Reproduce

```diff
diff --git a/app/Controllers/Home.php b/app/Controllers/Home.php
index 7f867e95f..7568a9619 100644
--- a/app/Controllers/Home.php
+++ b/app/Controllers/Home.php
@@ -6,6 +6,10 @@ class Home extends BaseController
 {
     public function index()
     {
+        foreach (range(1, 1e9) as $_) {
+            new \ReflectionClass($this);
+        }
+
         return view('welcome_message');
     }
 }
diff --git a/system/Debug/Exceptions.php b/system/Debug/Exceptions.php
index 9994aa7a0..2ec4468c7 100644
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -135,6 +135,7 @@ class Exceptions
             }
         }

+        var_dump($exception);
         $this->render($exception, $statusCode);

         exit($exitCode);
@@ -174,6 +175,8 @@ class Exceptions
             return;
         }

+        var_dump($error);
+
         ['type' => $type, 'message' => $message, 'file' => $file, 'line' => $line] = $error;

         if (in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE], true)) {
```

The error here is `E_ERROR` with value of `1`. However, based on the dumps:
image

The `type` key from the `$error` (which is `1`) gets turned into `$severity` of `0`.

### Expected Output

The type (or severity) of the error is preserved when converted into an `\ErrorException`.

### Anything else?

It seems the easy fix is to swap the `0` and `$type` when instantiating `ErrorException`.

```diff
diff --git a/system/Debug/Exceptions.php b/system/Debug/Exceptions.php
index 9994aa7a0..4a9220223 100644
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -135,6 +135,7 @@ class Exceptions
             }
         }

+        var_dump($exception);
         $this->render($exception, $statusCode);

         exit($exitCode);
@@ -174,10 +175,12 @@ class Exceptions
             return;
         }

         ['type' => $type, 'message' => $message, 'file' => $file, 'line' => $line] = $error;

         if (in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE], true)) {
-            $this->exceptionHandler(new ErrorException($message, $type, 0, $file, $line));
+            $this->exceptionHandler(new ErrorException($message, 0, $type, $file, $line));
         }
     }

```

but I'm not sure of BC concerns.

paulbalandan:
**Description**
Fixes #6454 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -177,7 +177,7 @@ class Exceptions
         ['type' => $type, 'message' => $message, 'file' => $file, 'line' => $line] = $error;
 
         if (in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE], true)) {
-            $this->exceptionHandler(new ErrorException($message, $type, 0, $file, $line));
+            $this->exceptionHandler(new ErrorException($message, 0, $type, $file, $line));
         }
     }
 


system/Debug/Exceptions.php
```diff
@@ -177,7 +177,7 @@ public function shutdownHandler()
         ['type' => $type, 'message' => $message, 'file' => $file, 'line' => $line] = $error;
 
         if (in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE], true)) {
-            $this->exceptionHandler(new ErrorException($message, $type, 0, $file, $line));
+            $this->exceptionHandler(new ErrorException($message, 0, $type, $file, $line));
```
kenjis, line 180: Looks good.

```php
public ErrorException::__construct(
    string $message = "",
    int $code = 0,
    int $severity = E_ERROR,
    ?string $filename = null,
    ?int $line = null,
    ?Throwable $previous = null
)
```

https://www.php.net/manual/en/errorexception.construct.php#refsect1-errorexception.construct-parameters

> severity
    The severity level of the exception.
        Note:
        While the severity can be any int value, it is intended that the [error constants](https://www.php.net/manual/en/errorfunc.constants.php) be used.

kenjis (approved):

MGatner (approved):

PR #6476, opened 2022-09-02
docs: add composer.json in Mandatory File Changes in upgrade_420.rst

kenjis:
**Description**
- add about composer.json autoload.psr-4

The config makes `spark migrate:status` output duplicated.
The same explanation is already here: https://codeigniter4.github.io/CodeIgniter4/general/managing_apps.html
But it is better to have it in Upgrading Guide.

The following is the result of installing v4.1.5 and updated to v4.2.5:
```console
$ php spark migrate:status

CodeIgniter v4.2.5 Command Line Tool - Server Time: 2022-09-02 18:25:16 UTC-05:00

+-----------+-------------------+----------+-------+-------------+-------+
| Namespace | Version           | Filename | Group | Migrated On | Batch |
+-----------+-------------------+----------+-------+-------------+-------+
| App       | 2022-09-02-231945 | Test     | ---   | ---         | ---   |
| App       | 2022-09-02-231945 | Test     | ---   | ---         | ---   |
+-----------+-------------------+----------+-------+-------------+-------+
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


MGatner (approved):
Ah that's the source of the duplicate migrations! Very interesting.

kenjis:
Yes, `App` namespace is registered twice, and it shows migrations twice.

PR #6480, opened 2022-09-03
Disable `date_time_create_from_format_call`

paulbalandan:
**Description**
This might need discussion.

```console
$ vendor/bin/php-cs-fixer describe date_time_create_from_format_call
Description of date_time_create_from_format_call rule.
The first argument of `DateTime::createFromFormat` method must start with `!`.
Consider this code:
    `DateTime::createFromFormat('Y-m-d', '2022-02-11')`.
    What value will be returned? '2022-01-11 00:00:00.0'? No, actual return value has 'H:i:s' section like '2022-02-11 16:55:37.0'.        
    Change 'Y-m-d' to '!Y-m-d', return value will be '2022-01-11 00:00:00.0'.
    So, adding `!` to format string will make return value more intuitive.

Fixer applying this rule is risky.
Risky when depending on the actual timings being used even when not explicit set in format.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,1 +1,1 @@
   -<?php \DateTime::createFromFormat('Y-m-d', '2022-02-11');
   +<?php \DateTime::createFromFormat('!Y-m-d', '2022-02-11');

```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -58,7 +58,8 @@ $overrides = [
         'allow_single_line_empty_anonymous_classes' => true,
         'allow_single_line_anonymous_functions'     => true,
     ],
-    'function_declaration' => [
+    'date_time_create_from_format_call' => false,
+    'function_declaration'              => [
         'closure_function_spacing'   => 'one',
         'trailing_comma_single_line' => false,
     ],
--- a/.php-cs-fixer.no-header.php
+++ b/.php-cs-fixer.no-header.php
@@ -50,7 +50,8 @@ $overrides = [
         'allow_single_line_empty_anonymous_classes' => true,
         'allow_single_line_anonymous_functions'     => true,
     ],
-    'function_declaration' => [
+    'date_time_create_from_format_call' => false,
+    'function_declaration'              => [
         'closure_function_spacing'   => 'one',
         'trailing_comma_single_line' => false,
     ],
--- a/.php-cs-fixer.user-guide.php
+++ b/.php-cs-fixer.user-guide.php
@@ -52,7 +52,8 @@ $overrides = [
         'allow_single_line_empty_anonymous_classes' => true,
         'allow_single_line_anonymous_functions'     => true,
     ],
-    'function_declaration' => [
+    'date_time_create_from_format_call' => false,
+    'function_declaration'              => [
         'closure_function_spacing'   => 'one',
         'trailing_comma_single_line' => false,
     ],


user_guide_src/source/outgoing/response/022.php
```diff
@@ -1,4 +1,4 @@
 <?php
 
 $response->setLastModified(date('D, d M Y H:i:s'));
-$response->setLastModified(DateTime::createFromFormat('u', $time));
+$response->setLastModified(DateTime::createFromFormat('!u', $time));
```
kenjis, line 4: This change is incorrect.

```
php > var_dump(DateTime::createFromFormat('u', 654321));
php shell code:1:
class DateTime#1 (3) {
  public $date =>
  string(26) "2022-09-04 00:00:00.654321"
  public $timezone_type =>
  int(3)
  public $timezone =>
  string(3) "UTC"
}
```
```
php > var_dump(DateTime::createFromFormat('!u', 654321));
php shell code:1:
class DateTime#1 (3) {
  public $date =>
  string(26) "1970-01-01 00:00:00.654321"
  public $timezone_type =>
  int(3)
  public $timezone =>
  string(3) "UTC"
}
```

user_guide_src/source/outgoing/response/018.php
```diff
@@ -1,4 +1,4 @@
 <?php
 
-$date = DateTime::createFromFormat('j-M-Y', '15-Feb-2016');
+$date = DateTime::createFromFormat('!j-M-Y', '15-Feb-2016');
 $response->setDate($date);
```
kenjis, line 4: I do not know in which case such a sample code would be used.
https://codeigniter4.github.io/CodeIgniter4/outgoing/response.html#CodeIgniter\HTTP\Response::setDate
So I also don't know the change is correct or not.

user_guide_src/source/outgoing/response/022.php
kenjis: The sample code doesn't make sense to begin with.

MGatner:
Skipping this one for now, looks like it may need some more work.

paulbalandan:
I disabled it for now.

kenjis (approved):
I think that enabling date_time_create_from_format_call and fixing existing sample code that I can't imagine use cases might be better.

MGatner:
I can't find documentation on `!` as a format modifier. Anyone have a link? It's not mentioned on https://www.php.net/manual/en/class.datetimeinterface.php#datetime.constants.types

paulbalandan:
See https://www.php.net/manual/en/datetimeimmutable.createfromformat.php

MGatner:
Ah thanks! It's specific to this method, makes sense. Anyone else bothered by this in that link?

> ALl fields are initialised

Yes I think this is a good change to apply but it will require taking them case by case. I will start going through as I have time.

MGatner (approved):
Disabling for now is fine but please set this to `true` in the Coding Standard.

user_guide_src/source/outgoing/response/022.php
paulbalandan: Based on the docs, this is expected.

kenjis: What do you mean?

Neither of the following seems to have a use case. What do we specify by $time?
```php
$response->setLastModified(DateTime::createFromFormat('u', $time));
$response->setLastModified(DateTime::createFromFormat('!u', $time));
```

paulbalandan: > Resets all fields (year, month, day, hour, minute, second, fraction and timezone information) to zero-like values ( 0 for hour, minute, second and fraction, 1 for month and day, 1970 for year and UTC for timezone information)

This means that since the format lacks any of "Y, m, d, H, i, s", the ! resets them.


I think the format should have been the capital U.

MGatner: I think you're correct. `u` generates _only_ the microseconds segment, so: `2022-09-10 07:37:24.5238` yields `5238`. This was supposed to be `U` for an actual full timestamp.

kenjis: There is no microseconds in the Last-Modified header:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

The sample code should be like the following:
```php
$response->setLastModified(DateTime::createFromFormat('U', $timestamp));
```
And it is the same as:
```php
$response->setLastModified(DateTime::createFromFormat('!U', $timestamp));
```

kenjis: I sent a PR #6526

MGatner: I think by definition `!U` is redundant.

PR #6482, opened 2022-09-03
Enable `no_trailing_comma_in_singleline`

paulbalandan:
**Description**
```console
$ vendor/bin/php-cs-fixer describe no_trailing_comma_in_singleline
Description of no_trailing_comma_in_singleline rule.
If a list of values separated by a comma is contained on a single line, then the last item MUST NOT have a trailing comma.

Fixer is configurable using following option:
* elements (a subset of ['arguments', 'array_destructuring', 'array', 'group_import']): which elements to fix; defaults to ['arguments', 'array_destructuring', 'array', 'group_import']

Fixing examples:
   --- Original
   +++ New
   @@ -1,5 +1,5 @@
    <?php
   -foo($a,);
   -$foo = array(1,);
   -[$foo, $bar,] = $array;
   -use a\{ClassA, ClassB,};
   +foo($a);
   +$foo = array(1);
   +[$foo, $bar] = $array;
   +use a\{ClassA, ClassB};

 * Example #2. Fixing with configuration: ['elements' => ['array_destructuring']].
   --- Original
   +++ New
   @@ -1,3 +1,3 @@
    <?php
    foo($a,);
   -[$foo, $bar,] = $array;
   +[$foo, $bar] = $array;

```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/database/call_function/002.php
+++ b/user_guide_src/source/database/call_function/002.php
@@ -1,3 +1,3 @@
 <?php
 
-$db->callFunction('some_function', $param1, $param2, /* ... */);
+$db->callFunction('some_function', $param1, $param2 /* , ... */);
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -50,8 +50,16 @@ $overrides = [
     ],
     'control_structure_braces'        => true,
     'no_multiple_statements_per_line' => true,
-    'no_useless_nullsafe_operator'    => true,
-    'phpdoc_separation'               => [
+    'no_trailing_comma_in_singleline' => [
+        'elements' => [
+            'arguments',
+            'array_destructuring',
+            'array',
+            'group_import',
+        ],
+    ],
+    'no_useless_nullsafe_operator' => true,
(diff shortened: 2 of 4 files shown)


user_guide_src/source/database/call_function/002.php
```diff
@@ -1,3 +1,3 @@
 <?php
 
-$db->callFunction('some_function', $param1, $param2, /* ... */);
+$db->callFunction('some_function', $param1, $param2/* ... */);
```
kenjis, line 3:
```suggestion
$db->callFunction('some_function', $param1, $param2 /* , ... */);
```

MGatner (approved):

user_guide_src/source/database/call_function/002.php
paulbalandan: Thanks! Done.

kenjis (approved):

MGatner:
I'm going to get some of these in so they can make it in the release.

PR #6483, opened 2022-09-03
Enable `no_useless_nullsafe_operator`

paulbalandan:
**Description**
```console
$ vendor/bin/php-cs-fixer describe no_useless_nullsafe_operator   
Description of no_useless_nullsafe_operator rule.
There should not be useless `null-safe-operators` `?->` used.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,7 +1,7 @@
    <?php
    class Foo extends Bar
    {
        public function test() {
   -        echo $this?->parentMethod();
   +        echo $this->parentMethod();
        }
    }

```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -43,6 +43,7 @@ $overrides = [
     'blank_line_between_import_groups' => true,
     'control_structure_braces'         => true,
     'no_multiple_statements_per_line'  => true,
+    'no_useless_nullsafe_operator'     => true,
     'phpdoc_separation'                => [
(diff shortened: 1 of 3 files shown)


kenjis (approved):

MGatner (approved):
Happy to see this one didn't trigger any changes! Null-safe operator is the new `empty()`: suspicious at best. Well, it would be if `empty()` weren't still alive and well 😤

Issue #6489, opened 2022-09-04
Bug: "integer" validation rule 500 error


shishamo:
### PHP Version

8.1

### CodeIgniter4 Version

4.2.5

### CodeIgniter4 Installation Method

Manual (zip or tar.gz)

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

apache

### What happened?

The validation rule "integer" fail as 500 error as below
```
{
    "title": "TypeError",
    "type": "TypeError",
    "code": 500,
    "message": "CodeIgniter\\Validation\\FormatRules::integer(): Argument #1 ($str) must be of type ?string, array given, called in /var/www/src/system/Validation/Validation.php on line 314",
    "file": "/var/www/src/system/Validation/FormatRules.php",
    "line": 132,
    "trace": [
        {
            "file": "/var/www/src/system/Validation/Validation.php",
            "line": 314,
            "function": "integer",
            "class": "CodeIgniter\\Validation\\FormatRules",
            "type": "->",
            "args": [
                [],
                null
            ]
        },
        {
            "file": "/var/www/src/system/Validation/Validation.php",
            "line": 163,
            "function": "processRules",
            "class": "CodeIgniter\\Validation\\Validation",
            "type": "->",
...
```

That validation allows string numeric (not integer) as well

### Steps to Reproduce

Validation rule used
```
protected $validationRules = [
        'integerAcceptsStringNumeric' => 'integer',
        'integerArrayError' => 'integer',
];
```

Data to validate
```
$data = [
        'integerAcceptsStringNumeric' => '1',
        'integerArrayError' => [],
];
```

### Expected Output

Validation fail and return 400 if the value is not an integer

ddevsr:
![image](https://user-images.githubusercontent.com/97607754/188298467-35c92c47-c8e7-4d57-81ac-738f1741b83c.png)


No error in my environment, i think this bug of behavior

iRedds:
@ddevsr you missed the rule ``'integerArrayError' => 'integer',``

ddevsr:
@iRedds Okay i updated

![image](https://user-images.githubusercontent.com/97607754/188298765-037fa3d5-bd47-4f90-9ed5-5695419b5055.png)

kenjis:
@shishamo If you validate non string data, I recommend you use Strict Rules.
See https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html#traditional-and-strict-rules
[...]

PR #6492, opened 2022-09-04
fix: Strict Validation Rules greater_than/less_than

kenjis:
**Description**
From https://github.com/codeigniter4/CodeIgniter4/issues/6489#issuecomment-1236288297

- fix type error in `greater_than` `greater_than_equal_to` `less_than` `less_than_equal_to`
  - `CodeIgniter\Validation\Rules::greater_than_equal_to(): Argument https://github.com/codeigniter4/CodeIgniter4/pull/1 ($str) must be of type ?string, int given, called in /var/www/src/system/Validation/StrictRules/Rules.php on line 88`
- refactor StrictRules/RulesTest

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


system/Validation/StrictRules/Rules.php
```diff
@@ -71,20 +71,28 @@ public function exact_length($str, string $val): bool
...
      * Greater than
      *
-     * @param mixed $str
+     * @param mixed $str expects int|string
      */
     public function greater_than($str, string $min): bool
     {
+        if (is_int($str)) {
+            $str = (string) $str;
+        }
```
MGatner, line 80: Is there a problem with using `@param int|string $str` and `$str = (string) $str;`?

kenjis: `$str` is user input, and we don't know what type it is.

But I found another bug. I'll fix.

kenjis: `$str = (string) $str;` makes `true` `'1'`.
Then `true` will be greater_than `0`. So it is not good.

I pushed a commit to fix another TypeError.

MGatner: The cast is dependent on the docblock type. At this point in dev in 2022 I'm fine requiring users to adhere to docblock types; if we say the parameter is `int|string` and someone passes `bool` then that's on them for not implementing any of the static analysis tools we and the larger PHP community provide.

This works though and is more conservative so it's fine to leave it how you have it.

MGatner (approved):

system/Validation/StrictRules/Rules.php
kenjis: A site user may send unexpected typed values.
In that case, if it is not an attack, returning false is better than TypeError.

Validation rules are special cases. Devs cannot force the input type.
So we cannot set `int|string` in this case.

PR #6499, opened 2022-09-05
Testing: fix failure of `testStreamFilter` function.

### LINKED ISSUE #6203: TestCaseTest::testStreamFilter() fails
```console
$ vendor/bin/phpunit tests/system/Test/
PHPUnit 9.5.21 #StandWithUkraine

Runtime:       PHP 8.0.20
Configuration: /Users/kenji/work/codeigniter/CodeIgniter4/phpunit.xml

...............................................................  63 / 206 ( 30%)
............................................................... 126 / 206 ( 61%)
.........................F..................................... 189 / 206 ( 91%)
.................                                               206 / 206 (100%)

Time: 00:04.397, Memory: 26.00 MB

There was 1 failure:

1) CodeIgniter\Test\TestCaseTest::testStreamFilter
--- Expected
+++ Actual
@@ @@
-'first.
+'
+first.
 '

.../CodeIgniter4/tests/system/Test/TestCaseTest.php:60

FAILURES!
Tests: 206, Assertions: 306, Failures: 1.
```

ping-yee:
**Description**
Fixed #6203 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Test/TestCaseTest.php
+++ b/tests/system/Test/TestCaseTest.php
@@ -56,7 +56,7 @@ final class TestCaseTest extends CIUnitTestCase
         CITestStreamFilter::$buffer = '';
         $this->stream_filter        = stream_filter_append(STDOUT, 'CITestStreamFilter');
         CLI::write('first.');
-        $expected = "first.\n";
+        $expected = "\nfirst.\n";
         $this->assertSame($expected, CITestStreamFilter::$buffer);
         stream_filter_remove($this->stream_filter);
     }
--- a/tests/system/Test/TestCaseTest.php
+++ b/tests/system/Test/TestCaseTest.php
@@ -27,6 +27,14 @@ final class TestCaseTest extends CIUnitTestCase
      */
     private $stream_filter;
 
+    protected function setUp(): void
+    {
+        parent::setUp();
+    
+        // Reset CLI::$lastWrite
+        CLI::print();
+    }
+
     public function testGetPrivatePropertyWithObject()
     {
         $obj    = new __TestForReflectionHelper();
--- a/tests/system/Test/TestCaseTest.php
+++ b/tests/system/Test/TestCaseTest.php
@@ -30,7 +30,7 @@ final class TestCaseTest extends CIUnitTestCase
     protected function setUp(): void
     {
         parent::setUp();
-    
+
         // Reset CLI::$lastWrite
         CLI::print();
     }


kenjis:
```diff
There was 1 failure:

1) CodeIgniter\Test\TestCaseTest::testStreamFilter
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'
-first.
+'first.
 '

/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Test/TestCaseTest.php:60
phpvfscomposer:///home/runner/work/CodeIgniter4/CodeIgniter4/vendor/phpunit/phpunit/phpunit:97
```

ping-yee:
@kenjis I don't have this failure in my side.

> ./vendor/bin/phpunit tests/system/Test/TestCaseTest.php
Do I order the wrong command to run the unit test or fix the wrong file?

I comment out the other function in this testing file, and this is the result I get.
image

iRedds:
The first CLI::write() call adds a newline before the line.
https://github.com/codeigniter4/CodeIgniter4/blob/ddee0c6e2dc8e4c3b1c06b09a375468435450138/system/CLI/CLI.php#L359-L364
It's a strange decision.

ping-yee:
@iRedds I also notice that design so that I make that change of testing.

Probably the design can make the first line of command print look more comfortable?!

kenjis:
> Do I order the wrong command to run unit test or fix the wrong file?

No. 

But both of the following commands should result in OK.

```
phpunit tests/system/Test/TestCaseTest.php
phpunit tests/system/Test/
```

This shows static variables are very difficult to handle.
In short, using static variables are not good design.

Speaking of the test cases, these are depending on the global state and we do not reset the global state after one test,
so the test result changes.

> And I don't think this is a good idea to change the CLI::write design cause there are too many other classes using this class.

Yes, it is difficult to change the behavior because it is a breaking change.

MGatner:
How about `@runInSeparateProcess` for that method?

kenjis:
`CLI::print()` resets `static::$lastWrite`.
I think it is better to use it and reset the value.

`@runInSeparateProcess` makes test slower and step debugging impossible (I don't know how to do step debugging).

ping-yee:
Whether `CLI::print()` more suitable for this testing function?
It may not be effected by `static::$lastWrite`.

kenjis:
How about this?
```diff
--- a/tests/system/Test/TestCaseTest.php
+++ b/tests/system/Test/TestCaseTest.php
@@ -27,6 +27,14 @@ final class TestCaseTest extends CIUnitTestCase
      */
     private $stream_filter;
 
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        // Reset CLI::$lastWrite
+        CLI::print();
+    }
+
     public function testGetPrivatePropertyWithObject()
     {
         $obj    = new __TestForReflectionHelper();
```

MGatner:
> How about this?

Love it! Hacky but efficient 🤗

ping-yee:
@kenjis I have been add your suggestion code and pushed, review pls.

kenjis:
@ping-yee See GitHub Actions results. PHPUnit failed.
It seems the code is not changed.
Did you fail the push?

kenjis:
Please run `composer cs-fix`.

kenjis (approved):

MGatner (approved):

PR #6502, opened 2022-09-06
docs: replace mixed in Database doc comments

kenjis:
**Description**
See #6310

- fix doc comments

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -1235,6 +1235,7 @@ abstract class BaseConnection implements ConnectionInterface
      * @param array|bool|float|int|object|string|null $str
      *
      * @return array|float|int|string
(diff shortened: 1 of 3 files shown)


paulbalandan (approved):

MGatner (commented):
How about `escapeIdentifiers()`? Looking at `ctype_digit()` we already need it to be a string:

> As of PHP 8.1.0, passing a non-string argument is deprecated. In the future, the argument will be interpreted as a string instead of an ASCII codepoint. Depending on the intended behavior, the argument should either be cast to string or an explicit call to chr() should be made.

kenjis:
Updated the doc comment for `escapeIdentifiers()` , and a bit more.

MGatner (approved):
🙏

PR #6503, opened 2022-09-06
feat: add manual config for Composer package auto-discovery

kenjis:
**Description**
See https://forum.codeigniter.com/showthread.php?tid=82772

Problems:
- Currently, auto-discovery searches the namespaces of all Composer packages. However, some packages may accidentally have files in the same directory as CI4, and loading the file may cause an error.
- It is useless to search files in Composer packages that are not CI4 modules.

This PR allows you to specify the packages to be searched by configuration.

- add optional config in `Config\Modules`
  - `$composerPackages['only']`: If you set this, only listed packages are searched for auto-discovery
  - `$composerPackages['exclude']`: If you set this, these packages are excluded for auto-discovery

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Modules.php
+++ b/app/Config/Modules.php
@@ -39,7 +39,7 @@ class Modules extends BaseModules
      *   [
      *       'only' => [
      *           // List up all packages to auto-discover
-     *           'codeigniter4/sheild',
+     *           'codeigniter4/shield',
      *       ],
      *   ]
      *   or
--- a/app/Config/Modules.php
+++ b/app/Config/Modules.php
@@ -32,6 +32,29 @@ class Modules extends BaseModules
     public $discoverInComposer = true;
 
     /**
+     * The Composer package list for Auto-Discovery
+     * This setting is optional.
+     *
+     * E.g.:
+     *   [
+     *       'only' => [
+     *           // List up all packages to auto-discover
+     *           'codeigniter4/sheild',
+     *       ],
+     *   ]
+     *   or
+     *   [
+     *       'exclude' => [
+     *           // List up packages to exclude.
+     *           'pestphp/pest',
+     *       ],
+     *   ]
+     *
+     * @var array
+     */
+    public $composerPackages = [];
+
+    /**
      * --------------------------------------------------------------------------
      * Auto-Discovery Rules
      * --------------------------------------------------------------------------
--- a/tests/system/Autoloader/AutoloaderTest.php
+++ b/tests/system/Autoloader/AutoloaderTest.php
@@ -286,6 +286,45 @@ final class AutoloaderTest extends CIUnitTestCase
         $this->assertStringContainsString(VENDORPATH, $namespaces['Psr\Log'][1]);
     }
 
+    public function testComposerPackagesOnly()
+    {
+        $config                      = new Autoload();
+        $config->psr4                = [];
+        $modules                     = new Modules();
+        $modules->discoverInComposer = true;
+        $modules->composerPackages   = ['only' => ['laminas/laminas-escaper']];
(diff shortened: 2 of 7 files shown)


MGatner:
👀

app/Config/Modules.php
```diff
@@ -31,6 +31,29 @@ class Modules extends BaseModules
...
+     *   ]
+     *   or
+     *   [
+     *       'exclude' => [
+     *           // List up packages to exclude.
+     *           'pestphp/pest',
+     *       ],
+     *   ]
+     *
+     * @var array
```
MGatner, line 53: An alternate suggestion to consider:
```suggestion
     * @var array<string, bool> Array of "package/name" => "whether to load/exclude"
```
It simplifies the Config but would take more explanation in documentation.

app/Config/Modules.php
```diff
@@ -31,6 +31,29 @@ class Modules extends BaseModules
...
 
+    /**
+     * The Composer package list for Auto-Discovery
+     * This setting is optional.
+     *
+     * E.g.:
+     *   [
+     *       'only' => [
+     *           // List up all packages to auto-discover
+     *           'codeigniter4/sheild',
```
MGatner, line 42:
```suggestion
     *           'codeigniter4/shield',
```

tests/system/Autoloader/AutoloaderTest.php
```diff
@@ -286,6 +286,45 @@ public function testComposerNamespaceDoesNotOverwriteConfigAutoloadPsr4()
...
+        $loader = new Autoloader();
+        $loader->initialize($config, $modules);
+
+        $namespaces = $loader->getNamespace();
+
+        $this->assertCount(1, $namespaces);
+        $this->assertStringContainsString(VENDORPATH, $namespaces['Laminas\Escaper'][0]);
+    }
+
+    public function testComposerPackagesExlcude()
```
MGatner, line 306: If you stick with the "only/exclude" Config style then please add another test for both:
```php
    public function testComposerPackagesOnlyAndExlcude()
```

MGatner (changes requested):

app/Config/Modules.php
kenjis: How do we handle unlisted packages?

tests/system/Autoloader/AutoloaderTest.php
kenjis: Added.

app/Config/Modules.php
MGatner: However they are handled now. Basically this would be the equivalent of:
```php
$only    = [];
$exclude = [];
foreach ($composerPackages as $package => $flag) {
    if ($flag) {
        $only[] = $package;
    } else {
        $exclude[] = $package;
    }
}
```
Same logic with a simpler Config. I'm not set on this though, just a suggestion for how to simplify the Config structure (but it has its own pitfalls).

kenjis: It is simpler as the array structure, but tihe config seems more difficult to understand for humans.

MGatner: Leave it; proceed.

tests/system/Autoloader/AutoloaderTest.php
MGatner: Thanks! Good exception.

MGatner (approved):
Looks good with docs

kenjis:
Added the docs.

MGatner (approved):
Looks good! Handy feature

PR #6510, opened 2022-09-08
feat: add request() and response()

kenjis:
**Description**
- add common functions

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/CommonFunctionsTest.php
+++ b/tests/system/CommonFunctionsTest.php
@@ -12,6 +12,7 @@
 namespace CodeIgniter;
 
 use CodeIgniter\Config\BaseService;
+use CodeIgniter\HTTP\CLIRequest;
 use CodeIgniter\HTTP\IncomingRequest;
 use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\HTTP\Response;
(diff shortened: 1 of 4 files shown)


lonnieezell (approved):
Ha! Now my docs on Shield work lol. 

Thanks for adding these!

paulbalandan (approved):

iRedds:
Is it possible to know the reason for implementing these features?
In one of our discussions, your point was that the IncomingRequest class object should only be accessed in the controller.
But this implementation is contrary to your own beliefs.

The controller also has access to the Response class.

tests/system/CommonFunctionsTest.php
```diff
@@ -137,6 +138,31 @@ public function testRedirectDefault()
         $this->assertInstanceOf(RedirectResponse::class, redirect());
     }
 
+    public function testRequestIncomingRequest()
+    {
+        Services::createRequest(new App());
+
+        $request = request();
+
+        $this->assertInstanceOf(IncomingRequest::class, $request);
```
MGatner, line 147: You should be able to assert that they reference the same object. 
```suggestion
        $this->assertInstanceOf(IncomingRequest::class, $request);
        $this->assertSame(Services::request(), $request);
```

MGatner (approved):
These are terribly convenient. I do want to point out that we now have a global helper function wrapping a service locator wrapping a mediator (for request type). No more layers than this, please!

MGatner (approved):

kenjis:
Added the docs.

kenjis:
@iRedds See https://github.com/codeigniter4/shield/pull/417#pullrequestreview-1098743114
Yes, I don't think these are good parts.
Because basically we don't need to use these in our apps.

But we have already `Services::request()` and `service('request')`,
and it seems many devs like global functions.
`request()` is a bit better than `service('request')`, because it can tell return type.

iRedds:
I hope this is a balanced decision.
Because too much flexibility gives too many headaches to maintain a framework.

PR #6514, opened 2022-09-08
Add options to `new_with_braces`

paulbalandan:
**Description**
```console
$ vendor/bin/php-cs-fixer describe new_with_braces     
Description of new_with_braces rule.
All instances created with `new` keyword must (not) be followed by braces.

Fixer is configurable using following options:
* named_class (bool): whether named classes should be followed by parentheses; defaults to true
* anonymous_class (bool): whether anonymous classes should be followed by parentheses; defaults to true

Fixing examples:
   --- Original
   +++ New
   @@ -1,4 +1,4 @@
    <?php

   -$x = new X;
   -$y = new class {};
   +$x = new X();
   +$y = new class() {};

 * Example #2. Fixing with configuration: ['anonymous_class' => false].
   --- Original
   +++ New
   @@ -1,3 +1,3 @@
    <?php

   -$y = new class() {};
   +$y = new class {};

 * Example #3. Fixing with configuration: ['named_class' => false].
   --- Original
   +++ New
   @@ -1,3 +1,3 @@
    <?php

   -$x = new X();
   +$x = new X;

```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -62,6 +62,10 @@ $overrides = [
         'closure_function_spacing'   => 'one',
         'trailing_comma_single_line' => false,
     ],
+    'new_with_braces' => [
+        'named_class'     => true,
+        'anonymous_class' => true,
+    ],
     'no_multiple_statements_per_line' => true,
     'no_trailing_comma_in_singleline' => [
         'elements' => [
--- a/.php-cs-fixer.no-header.php
+++ b/.php-cs-fixer.no-header.php
@@ -54,6 +54,10 @@ $overrides = [
         'closure_function_spacing'   => 'one',
         'trailing_comma_single_line' => false,
     ],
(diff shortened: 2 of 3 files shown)


kenjis (approved):

MGatner (approved):
Don't ask me where I picked this up, but I'm accustomed to spaces in between the declaration for anonymous classes:
```php
$banana = new class () extends Config {}
```
Do we have rules to enforce that one way or another?

paulbalandan:
It's the `space_before_parenthesis` option of `class_definition` rule.

```console
* Example #5. Fixing with configuration: ['space_before_parenthesis' => true].
   --- Original
   +++ New
   @@ -1,2 +1,2 @@
    <?php
   -$foo = new class(){};
   +$foo = new class () {};

```

Issue #4494, opened 2021-03-26
Bug: Forge create the same names for CREATE INDEX and UNIQUE


okatse:
This add Index
`$this->forge->addKey('username');`

This add Unique Key 
`$this->forge->addUniqueKey('username');`

> CREATE INDEX "users_username" ON "public"."users" ("username");
> ALTER TABLE "public"."users" ADD CONSTRAINT "users_username" UNIQUE ("username");

In both cases, the `_processIndexes` method generates the same names 
Perhaps CREATE INDEX is missing a prefix for example:  `CREATE INDEX "users_username_IDX"`

```php
	protected function _processIndexes(string $table)
	{
		$sqls = [];

		for ($i = 0, $c = count($this->keys); $i < $c; $i++)
		{
			$this->keys[$i] = (array) $this->keys[$i];

			for ($i2 = 0, $c2 = count($this->keys[$i]); $i2 < $c2; $i2++)
			{
				if (! isset($this->fields[$this->keys[$i][$i2]]))
				{
					unset($this->keys[$i][$i2]);
				}
			}
			if (count($this->keys[$i]) <= 0)
			{
				continue;
			}

			if (in_array($i, $this->uniqueKeys, true))
			{
				$sqls[] = 'ALTER TABLE ' . $this->db->escapeIdentifiers($table)
						  . ' ADD CONSTRAINT ' . $this->db->escapeIdentifiers($table . '_' . implode('_', $this->keys[$i]))
						  . ' UNIQUE (' . implode(', ', $this->db->escapeIdentifiers($this->keys[$i])) . ');';
				continue;
			}

			$sqls[] = 'CREATE INDEX ' . $this->db->escapeIdentifiers($table . '_' . implode('_', $this->keys[$i]))
					  . ' ON ' . $this->db->escapeIdentifiers($table)
					  . ' (' . implode(', ', $this->db->escapeIdentifiers($this->keys[$i])) . ');';
		}

		return $sqls;
	}
```

najdanovicivan:
Related to #5075

kenjis:
I'm not sure prefix or suffix should be added or not.
But longer index name may cause the limit of index name length.

Setting custom index name #5075 seems to be better solution.

PR #6516, opened 2022-09-08
Make Unique Index Name Unique

sclubricants:
This PR fixes https://github.com/codeigniter4/CodeIgniter4/issues/4494

Ultimately the better option is to allow setting the name manually but so long as we are automatically setting the name it needs to be unique.

I would like to get this change committed before working on some other changes.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Forge.php
+++ b/system/Database/Forge.php
@@ -1041,7 +1041,8 @@ class Forge
 
             if (in_array($i, $this->uniqueKeys, true)) {
                 $sqls[] = 'ALTER TABLE ' . $this->db->escapeIdentifiers($table)
-                    . ' ADD CONSTRAINT ' . $this->db->escapeIdentifiers($table . '_' . implode('_', $this->keys[$i]))
+                    . ' ADD CONSTRAINT ' . $this->db->escapeIdentifiers($table . '_'
+                    . implode('_', $this->keys[$i]) . '_u')
                     . ' UNIQUE (' . implode(', ', $this->db->escapeIdentifiers($this->keys[$i])) . ')';
 
                 continue;
--- a/system/Database/MySQLi/Forge.php
+++ b/system/Database/MySQLi/Forge.php
@@ -212,7 +212,8 @@ class Forge extends BaseForge
 
             $unique = in_array($i, $this->uniqueKeys, true) ? 'UNIQUE ' : '';
 
-            $sql .= ",\n\t{$unique}KEY " . $this->db->escapeIdentifiers(implode('_', $this->keys[$i]))
+            $sql .= ",\n\t{$unique}KEY " . $this->db->escapeIdentifiers(implode('_', $this->keys[$i])
+                . (in_array($i, $this->uniqueKeys, true) ? '_u' : ''))
                 . ' (' . implode(', ', $this->db->escapeIdentifiers($this->keys[$i])) . ')';
         }
 
--- a/system/Database/SQLSRV/Forge.php
+++ b/system/Database/SQLSRV/Forge.php
@@ -258,7 +258,8 @@ class Forge extends BaseForge
             if (in_array($i, $this->uniqueKeys, true)) {
                 $sqls[] = 'ALTER TABLE '
                     . $this->db->escapeIdentifiers($this->db->schema) . '.' . $this->db->escapeIdentifiers($table)
-                    . ' ADD CONSTRAINT ' . $this->db->escapeIdentifiers($table . '_' . implode('_', $this->keys[$i]))
+                    . ' ADD CONSTRAINT ' . $this->db->escapeIdentifiers($table . '_'
+                    . implode('_', $this->keys[$i]) . '_u')
                     . ' UNIQUE (' . implode(', ', $this->db->escapeIdentifiers($this->keys[$i])) . ');';
 
                 continue;
--- a/system/Database/SQLite3/Forge.php
+++ b/system/Database/SQLite3/Forge.php
@@ -176,7 +176,8 @@ class Forge extends BaseForge
             }
 
             if (in_array($i, $this->uniqueKeys, true)) {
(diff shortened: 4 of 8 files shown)


kenjis:
I'm fine with this PR.
Please add the changelog.

MGatner:
Looking good! @codeigniter4/database-team can also help with review or feature testing.

kenjis (approved):

michalsn:
I have a problem with the consequences of this PR. 

TLDR: Changing the convention in the names of unique indexes may cause errors in the production database and existing migration files.

Let's start from the beginning. Technically speaking, this (https://github.com/codeigniter4/CodeIgniter4/issues/4494) is a bug. But practically... it will never happen simply because it makes no sense to add a key and then a unique key for the same column.

I think we should resolve it differently. By allowing the possibility of adding a custom name for the index. Just like we are planning to do with foreign keys (https://github.com/codeigniter4/CodeIgniter4/pull/6468). It's the only way I see to handle this dubious case and be consistent with the convention we have now.

If we leave it like this, it will produce the same problems as described here https://github.com/codeigniter4/CodeIgniter4/pull/6468#discussion_r972334407 but with unique indexes.

This will affect every user who has an existing project and uses migrations (with unique indexes). I'm not sure if the consequences of this PR were properly recognized, and since it was already merged, I will tag everyone: @lonnieezell @MGatner @kenjis @paulbalandan @samsonasik

sclubricants:
>But practically... it will never happen simply because it makes no sense to add a key and then a unique key for the same column.

This could happen if they are trying to change between unique and non unique and they create one before dropping the other,

I really wouldn't get too hung up on index names. There is nothing documenting how names will be applied. It would have been nice if since the beginning developers could set their own names but it hasn't been that way. If I'm creating a migration and I didn't set the names then I'm querying the table to find the name and then dropping that. Most all DBMS don't even require a name and will generate a random one. This is basically how I see CI. I do believe it would be nice to be able to set the name but if you didn't set the name than you can't rely on the name.

michalsn:
> This could happen if they are trying to change between unique and non unique and they create one before dropping the other,

Well, that's quite easy to avoid.

> I really wouldn't get too hung up on index names. There is nothing documenting how names will be applied. It would have been nice if since the beginning developers could set their own names but it hasn't been that way. If I'm creating a migration and I didn't set the names then I'm querying the table to find the name and then dropping that.

The thing is, that I already know the index name that was created in my database, and this PR changes that name.

It becomes a problem when you have an existing project. Let me tell you what will happen if you have a project that has any operations that involve unique indexes (adding and dropping), and you will update CI to the dev version:
* you run the tests, and all your "dropKey" operations are failing (index name has changed)
* you run `php spark migrate:refresh` - all your "dropKey" operations are failing (index name has changed)
* then you probably realize you should fix your migrations and adjust them to the new unique index convention - ok, done
* so everything is fixed, right? Nope, you have to fix all your unique indexes in the production database manually - otherwise, it will get you into trouble at some point (different indexes in the development environment and different in production)

These are the real consequences of these changes. And I don't even mention the third-party packages you may be using - they may have their own migration files.

To be clear... if no one else thinks it's a problem (for the minor version release) then I don't have a problem with it either. No pressure.

sclubricants:
Well we certainly need to add the ability to set the name manually. I'll create a PR for this.

lonnieezell:
Sorry I didn't really look at this one previously. If it's changing the name of the key then, yes, we should avoid this. Especially on a point release. @michalsn already pointed out the challenges. 

I know it sucks having to worry about how badly we break people's apps, and we might become a little too loose with BC changes in general. But thousands of developers are counting on the framework for their livelihood. The more frustrations we throw in people's ways the more they're likely to leave the community, and bad mouth the framework in general. 

I'm of the opinion this should be reverted. 

I like the manual naming option, though, and glad to see @sclubricants is willing to work on that. Thanks to all involved in this. 

@MGatner  @kenjis what's your opinions?

kenjis:
> we might become a little too loose with BC changes in general.

Yes, maybe.

I agree to revert.

The change in index naming conventions seems to do more harm than good to existing users.
It seems to me that there should be a setting not to change it, even in the major version.

sclubricants:
I can revert the change in the PR I'm working on to add naming. Will this work for everyone? I'll submit the PR tomorrow.

kenjis:
Please send a PR only to revert this PR if you plan to send.

sclubricants:
This will make a mess of the PR I'm already working on. Its literally two or three lines of code to revert this. The rests would need to be changed but I wanted to work the naming into the tests so it would kill two birds with one stone.

kenjis:
@sclubricants Sorry for inconvenience.
I sent a PR to revert: https://github.com/codeigniter4/CodeIgniter4/pull/6548
After merging this, I think it is OK to rebase your working branch and resolve the conflicts.

MGatner:
I'm watching this but leaving it up to you all. It sounds like everyone wants deliberate naming _and_ that addresses the problems: win-win. Thanks for calling this out @michalsn and thanks for working on the name feature @sclubricants! Glad to see some good movement on the database layer.

michalsn:
Thanks for the feedback, everyone.

michalsn:
@kenjis mentioned something important

> It seems to me that there should be a setting not to change it, even in the major version.

This makes me think... we have a `Feature` config file... maybe it would be good, to make this change configurable? Then, it could work for unique keys as well as for foreign keys. 

I guess this would make everyone happy. And existing projects would be safe. Thoughts?

MGatner:
Sounds like a reasonable suggestion! But I formed the database team mostly so I wouldn't need to deliberate on these matters where I am less knowledgeable 😅 So I'm good with however you all want to take this. Just grateful it's getting attention...

PR #6520, opened 2022-09-09
New: Scaffold should generate entity too.

DhPandya:
The scaffold command should also create an entity.

**Description**
Entities are another most usable feature of CodeIgniter. So when we're using the scaffold command it helps by generating Controller, Model, Migrations, and Seeder. An entity also can be generated so no one has to create the same manually.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


kenjis:
Thank you for sending a PR.

It seems this PR is an enhancement. So this PR should go to `4.3` branch, not `develop`.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#branching

Please fix the base branch.

paulbalandan:
Have you done this?
> To include an Entity class in the scaffolded files, just include the `--return entity` to the command and it will be passed to the model generator.

https://codeigniter4.github.io/userguide/cli/cli_generators.html#scaffolding-a-complete-set-of-stock-code

DhPandya:
> Have you done this?
> 
> > To include an Entity class in the scaffolded files, just include the `--return entity` to the command and it will be passed to the model generator.
> 
> https://codeigniter4.github.io/userguide/cli/cli_generators.html#scaffolding-a-complete-set-of-stock-code

Yes, I've tried this and it also works for me.

kenjis:
This feature is already implemented as an option.

Issue #6523, opened 2022-09-09
refactor: Change `global $app` to `Config\Services::codeigniter()` in `CodeIgniter\Debug\Toolbar` file


ping-yee:
From https://forum.codeigniter.com/showthread.php?tid=82889&pid=399578#pid399578

kenjis:
Fixed by https://github.com/codeigniter4/CodeIgniter4/pull/6524

PR #6524, opened 2022-09-09
refactor: replace `global $app` with Services

ping-yee:
**Description**
Modified #6523.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -351,7 +351,6 @@ class Toolbar
      * @param RequestInterface  $request
      * @param ResponseInterface $response
      *
-     * @global \CodeIgniter\CodeIgniter $app
      */
     public function prepare(?RequestInterface $request = null, ?ResponseInterface $response = null)
     {
--- a/system/Debug/Toolbar.php
+++ b/system/Debug/Toolbar.php
@@ -350,7 +350,6 @@ class Toolbar
      *
      * @param RequestInterface  $request
      * @param ResponseInterface $response
-     *
      */
     public function prepare(?RequestInterface $request = null, ?ResponseInterface $response = null)
     {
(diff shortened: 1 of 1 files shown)


kenjis (approved):

iRedds:
I understand the reason for this PR, but I don't think it's the right solution.

To begin with, the ``Services::codeigniter()`` service was actually introduced at the suggestion of a developer who wanted to change the request handler.
The reason for this was the unsuccessful design of the framework and the rejection of autowiring.
Therefore, I do not consider it necessary to implement this service, which is a crutch.

I also think that the design of the toolbar filter is bad. Instead of simply injecting a toolbar header or script into the response, data collection is additionally triggered.

I think the way the toolbar works should be changed instead of using ``Services::codeigniter()``.

- Codeigniter 
--- filter toolbar::after( Turn on the toolbar. Inject a header or script into the response. )
-- sending the response
- if the toolbar is enabled, we collect data, including timer data.

ping-yee:
@iRedds Thanks for your advices, I understand what you mean and feel grateful.
Is there any alternative change can use?
If it isn't, I will close this PR.

kenjis:
@iRedds It seems true what you say, but this PR fixes the bug in Long-living PHP like RoadRunner.
I think even if a PR is not perfect, if it improves CI4, it is better to merge.

iRedds:
I just expressed my opinion.
So which way to choose is up to you, gentlemen.

kenjis:
@ping-yee Can you remove this?
https://github.com/codeigniter4/CodeIgniter4/blob/e1dbce0b79e6de31a2e6eb72c86163f01e06f959/system/Debug/Toolbar.php#L354

kenjis:
@iRedds Thank you for your opinion. You can send a PR to fix the bad design of the debugbar later.

ping-yee:
@kenjis I have been taken out the comment out, Appreciated that!

kenjis (approved):

PR #6525, opened 2022-09-10
Added view() method to route collections.

lonnieezell:
**Description**
Added a new method to simplify rendering out views directly through routes. While you could do this with a closure, this is tidier and easier to read when using many of these. 

This supports static info pages in sites, like About, Privacy, Terms, etc. It also supports a more view-first workflow, where the view is primary object, and View Cells are used to display the dynamic content.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -993,6 +993,21 @@ class RouteCollection implements RouteCollectionInterface
     }
 
     /**
+     * Specifies a route that will only display a view.
+     * Only works for GET requests.
+     */
+    public function view(string $from, string $view, ?array $options = null): RouteCollectionInterface
+    {
+        $to = static function (...$data) use ($view) {
+            return view($view, ['_page' => $data]);
+        };
+
+        $this->create('get', $from, $to, $options);
+
+        return $this;
+    }
+
+    /**
      * Limits the routes to a specified ENVIRONMENT or they won't run.
      */
     public function environment(string $env, Closure $callback): RouteCollectionInterface
--- a/tests/system/Router/RouteCollectionTest.php
+++ b/tests/system/Router/RouteCollectionTest.php
@@ -12,9 +12,11 @@
 namespace CodeIgniter\Router;
 
 use App\Controllers\Home;
+use CodeIgniter\Config\Factory;
 use CodeIgniter\Config\Services;
 use CodeIgniter\Exceptions\PageNotFoundException;
 use CodeIgniter\Test\CIUnitTestCase;
+use CodeIgniter\View\View;
 use Config\Modules;
 use Tests\Support\Controllers\Hello;
 
@@ -817,6 +819,16 @@ final class RouteCollectionTest extends CIUnitTestCase
 
         $routes->cli('here', 'there');
         $this->assertSame($expected, $routes->getRoutes('cli'));
+    }
+
+    public function testView()
+    {
+        $routes = $this->getCollector();
+
+        $routes->view('here', 'hello');
+
+        $route = $routes->getRoutes('get')['here'];
+        $this->assertIsCallable($route);
     }
 
     public function testEnvironmentRestricts()
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -127,6 +127,7 @@ Others
 ======
 
 - Added ``$routes->useSupportedLocalesOnly(true)`` so that the Router returns 404 Not Found if the locale in the URL is not supported in ``Config\App::$supportedLocales``. See :ref:`Localization <localization-in-routes>`
+- Added new ``$routes->view()`` method to echo the view directly. See :ref:`View Routes <view-routes>`.
 
 Changes
 *******
--- a/user_guide_src/source/incoming/routing.rst
+++ b/user_guide_src/source/incoming/routing.rst
@@ -206,9 +206,24 @@ Closures
 
 You can use an anonymous function, or Closure, as the destination that a route maps to. This function will be
 executed when the user visits that URI. This is handy for quickly executing small tasks, or even just showing
-a simple view:
+a simple view.
 
-.. literalinclude:: routing/020.php
+.. literalinclude:: routing/065.php
+
+If you use placeholders within your route, you can access them within the view in a special variable, ``$_page``.
+They are available as an array, indexed in the order they appear in the route.
+
+.. literalinclude:: routing/066.php
+
+.. _view_routes:
+
+Views
+=====
+
+If you just want to render a view out that has no logic associated with it, you can use the ``view()`` method.
+This method accepts the name of the view to load as the ``$to`` parameter.
+
+.. literalinclude:: routing/021.php
 
 Mapping Multiple Routes
 =======================
--- a/user_guide_src/source/incoming/routing/065.php
+++ b/user_guide_src/source/incoming/routing/065.php
@@ -1,0 +1,4 @@
+<?php
+
+// Displays the view in /app/Views/pages/about.php
+$routes->view('about', 'pages/about');
--- a/user_guide_src/source/incoming/routing/066.php
+++ b/user_guide_src/source/incoming/routing/066.php
@@ -1,0 +1,7 @@
+<?php
+
+// Displays the view in /app/Views/map.php
+$routes->view('map/(:segment)/(:segment)', 'map');
+
+// Within the view, you can access the segments with
+// $_page[0] and $_page[1] respectively.


system/Router/RouteCollection.php
```diff
@@ -992,6 +992,19 @@ public function cli(string $from, $to, ?array $options = null): RouteCollectionI
         return $this;
     }
 
+    /**
+     * Specifies a route that will only display a view.
+     * Only works for GET requests.
+     */
+    public function view(string $from, string $view, ?array $options = null): RouteCollectionInterface
+    {
+        $to = static fn (...$data) => view($view, ['_page' => $data]);
```
iRedds, line 1001: I don't see the point in assigning an anonymous function to a variable. We can pass it right away as an argument.
Also I'm not sure about the need to pass captured URL segments.

iRedds (commented):
I think it's worth specifying a new method in RouteCollectionInterface.

user_guide_src/source/incoming/routing.rst
```diff
@@ -206,9 +206,24 @@ Closures
...
 
-.. literalinclude:: routing/020.php
+.. literalinclude:: routing/065.php
+
+If you use placeholders within your route, you can access them within the view in a special variable, ``$_page``.
+They are available as an array, indexed in the order they appear in the route.
+
+.. literalinclude:: routing/066.php
+
+.. _view_routes:
```
kenjis, line 218:
```suggestion
.. _view-routes:
```

system/Router/RouteCollection.php
kenjis: The variable seems to make the code a bit easier to read.

But why the passed variable name is `_page`?

lonnieezell: @iRedds Good point about the variable, but I think readability trumps efficiency here. 

I think it's a must to pass URL segments as they can be used within View Cells for more flexible pages. 

@kenjis Thinking back it's probably not necessary. My original thought was I wanted to make sure it wouldn't conflict with anything the user might provide - but they wouldn't provide anything in this scenario. So removing the underscore makes sense.

system/Router/RouteCollection.php
```diff
@@ -992,6 +992,19 @@ public function cli(string $from, $to, ?array $options = null): RouteCollectionI
         return $this;
     }
 
+    /**
+     * Specifies a route that will only display a view.
+     * Only works for GET requests.
+     */
+    public function view(string $from, string $view, ?array $options = null): RouteCollectionInterface
```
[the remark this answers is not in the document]
lonnieezell: Perhaps it isn't the best way - but it's the way all of the other methods in the RouteCollection file are written. If we want to change that bad enough we can change them all in a separate PR.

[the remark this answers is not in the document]
MGatner: I agree with both. Null should not be used unless it is needed to convey a specific sentinel value. But given that this abuse is throughout the class already the consistency is the lesser evil.

system/Router/RouteCollection.php
```diff
@@ -992,6 +992,19 @@ public function cli(string $from, $to, ?array $options = null): RouteCollectionI
...
+    {
+        $to = static fn (...$data) => view($view, ['page' => $data]);
```
MGatner, line 1001: This adds a new dependency on the View class, which I think is acceptable but I don't like that it is hidden by the global function. I recommend changing this to use the actual class and adjust Deptrac (as necessary).

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -127,6 +127,7 @@ Others
 ======
 
 - Added ``$routes->useSupportedLocalesOnly(true)`` so that the Router returns 404 Not Found if the locale in the URL is not supported in ``Config\App::$supportedLocales``. See :ref:`Localization <localization-in-routes>`
+- Added new ``$routes->view()`` method to echo the view directly. See :ref:`View Routes <view-routes>`.
```
MGatner, line 130:
```suggestion
- Added new ``$routes->view()`` method to return a view directly. See :ref:`View Routes <view-routes>`.
```

tests/system/Router/RouteCollectionTest.php
```diff
@@ -819,6 +819,16 @@ public function testCLI()
...
+    public function testView()
+    {
+        $routes = $this->getCollector();
+
+        $routes->view('here', 'hello');
+
+        $route = $routes->getRoutes('get')['here'];
+        $this->assertIsCallable($route);
+    }
+
```
MGatner, line 831: Can we get something like `testViewRespondsToGetOnly()` that shows a POST resulting in 404?

user_guide_src/source/incoming/routing.rst
```diff
@@ -206,9 +206,24 @@ Closures
...
+
+Views
+=====
+
+If you just want to render a view out that has no logic associated with it, you can use the ``view()`` method.
+This method accepts the name of the view to load as the ``$to`` parameter.
+
+.. literalinclude:: routing/021.php
+
+If you use placeholders within your route, you can access them within the view in a special variable, ``$page``.
```
MGatner, line 223: I'm curious about the choice of `$page` over something like "data" or "parameters/params" or the like. Was this some precedence from Symfony UX?

MGatner (changes requested):

user_guide_src/source/incoming/routing.rst
lonnieezell: No, this portion has nothing to do with Symfony UX. I've been using more view-first pages lately and wanted something nicer, that's all. :) I think I got the name from either NextJS or SvelteKit, but I don't recall. I can change the name to `$params` or whatever we think is best, so there's no worry of conflicting with user data vars this way.

kenjis: I can't imagine what `$page[0]` means from the variable name.

lonnieezell: Agreed. Maybe it should be called `$segments` or something?

MGatner: I'm good with "segments". It will always be some portion of a URI parsed from the Router right? So that makes sense.

MGatner:
> I think it's worth specifying a new method in RouteCollectionInterface.

I agree; we shouldn't create a new public method in our implementing class without either updating the interface or creating a new interface.

system/Router/RouteCollection.php
kenjis: I agree, but it is difficult to notice this because global functions hide dependencies.

user_guide_src/source/incoming/routing.rst
kenjis: I'm not sure if it is the best, but it is a value of a URI segment, so `segments` make sense.

lonnieezell:
> I think it's worth specifying a new method in RouteCollectionInterface.

It's been a long time since I've been in that file. Just looked, and the only method in the interface relating to adding routes is the is the `add()` method. All other ones look to be thought of as convenience features that don't have to be present if someone creates their own implementation. We can address that in a different PR if desired.

tests/system/Router/RouteCollectionTest.php
lonnieezell: I've updated that test to verify it's only available in the GET routes, not in any of the others (like POST, HEAD, etc). Assuming all of the other tests in our system are correct, this should prove the same thing. :)

system/Router/RouteCollection.php
lonnieezell: @MGatner I assume you mean use `Services::renderer()` here? If we call the class directly it won't take any custom View classes into account when rendering the view.

tests/system/Router/RouteCollectionTest.php
MGatner: Even better!

MGatner (approved):
I didn't realize there were so many methods missing from the interface already. We are working on this in another project, which will be post-4.3, so fine to leave this one out as well. 

Looks good!

system/Router/RouteCollection.php
kenjis: @MGatner If we want to detect a new dependency, we need to code like the following:

```diff
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -14,6 +14,7 @@ namespace CodeIgniter\Router;
 use Closure;
 use CodeIgniter\Autoloader\FileLocator;
 use CodeIgniter\Router\Exceptions\RouterException;
+use CodeIgniter\View\View;
 use Config\App;
 use Config\Modules;
 use Config\Services;
@@ -998,7 +999,9 @@ class RouteCollection implements RouteCollectionInterface
      */
     public function view(string $from, string $view, ?array $options = null): RouteCollectionInterface
     {
-        $to = static fn (...$data) => Services::renderer()
+        /** @var View $renderer */
+        $renderer = Services::renderer();
+        $to       = static fn (...$data) => $renderer
             ->setData(['segments' => $data], 'raw')
             ->render($view, $options);
 
```

Then we can see the errors.

```console
$ deptrac 
 507/507 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ----------- -------------------------------------------------------------------------------------------- 
  Reason      Router                                                                                      
 ----------- -------------------------------------------------------------------------------------------- 
  Violation   CodeIgniter\Router\RouteCollection must not depend on CodeIgniter\View\View (View)          
              /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Router/RouteCollection.php:1002  
  Violation   CodeIgniter\Router\RouteCollection must not depend on CodeIgniter\View\View (View)          
              /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Router/RouteCollection.php:17    
 ----------- -------------------------------------------------------------------------------------------- 


 -------------------- ----- 
  Report                    
 -------------------- ----- 
  Violations           2    
  Skipped violations   32   
  Uncovered            813  
  Allowed              613  
  Warnings             0    
  Errors               0    
 -------------------- -----
```

kenjis:
rebase to fix coding style violations and conflicts, please.

system/Router/RouteCollection.php
MGatner: Yeah that was my mistake, I forgot how much work `view()` was doing. This is fine, or even reverting to the helper.

kenjis:
This PR with 121 commits was merged? Why?

kenjis:
I sent a PR to revert: #6558

lonnieezell:
Well damn. I'm sorry I didn't notice all of those. This was unintentional. I pulled down the latest 4.3, rebased that into this branch and pushed the changes. I could have sworn this branch was built off of 4.3 originally, so I'm not sure why it did this. 

This PR has 67 files changed now, and the one to revert only has 6? That doesn't seem complete.

kenjis:
I don't know how we can get this branch status.

> This PR has 67 files changed now, and the one to revert only has 6? That doesn't seem complete.

I created the revert PR by the `Revert` button on the GitHub.
It seems not complete, but I think the revert PR is okay.
I have `4.3` before merging this PR on local. I fetched upstream and check the diff:
```
$ git diff 4.3...upstream/4.3 --stat
 system/Router/RouteCollection.php              | 15 +++++++++++++++
 tests/system/Router/RouteCollectionTest.php    | 21 +++++++++++++++++++++
 user_guide_src/source/changelogs/v4.3.0.rst    |  1 +
 user_guide_src/source/incoming/routing.rst     | 19 +++++++++++++++++--
 user_guide_src/source/incoming/routing/065.php |  4 ++++
 user_guide_src/source/incoming/routing/066.php |  7 +++++++
 6 files changed, 65 insertions(+), 2 deletions(-)
```

It seems this PR has commits that are already merged into 4.3.

MGatner:
Let's force-push back instead of the revert, then cherry-pick to a new branch and open a second PR. I can do this on desktop in a bit if nobody has yet.

MGatner:
I would never force-push master or develop, but 4.3 is not protected nor planned to be a persistent branch - it's basically a long-term feature branch. I am fine forcing it, but I'm also not worried about having the messy history if you'd rather proceed with reverting & fixing.

kenjis:
`4.3` might not be a persistent branch, but it persists until 4.3 is released.
And many contributors uses it as a base branch. Falsifying history affects them.
I think `4.3` should be protected as long as it lives, just like `develop`.

As for having clean history, I don't think it is possible unless all contributors can use `git rebase` properly.

kenjis:
I reverted this PR, and cherry-picked and created a new PR #6568

PR #6527, opened 2022-09-11
fix: types in database classes

kenjis:
**Description**
- fix types

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Database.php
+++ b/system/Database/Database.php
@@ -35,7 +35,7 @@ class Database
      * Parses the connection binds and returns an instance of the driver
      * ready to go.
      *
-     * @return mixed
+     * @return BaseConnection
      *
      * @throws InvalidArgumentException
      */
@@ -61,7 +61,7 @@ class Database
     /**
      * Creates a Forge instance for the current database type.
      */
-    public function loadForge(ConnectionInterface $db): object
+    public function loadForge(ConnectionInterface $db): Forge
     {
         if (! $db->connID) {
             $db->initialize();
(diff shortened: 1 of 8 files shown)


system/Database/Database.php
```diff
@@ -71,9 +71,9 @@ public function loadForge(ConnectionInterface $db): object
     }
 
     /**
-     * Creates a Utils instance for the current database type.
+     * Creates an Utils instance for the current database type.
```
MGatner, line 74: We would say "a Utilities instance" (like "I went to a university") but the shortened version makes both seem weird. What about:
```suggestion
     * Creates an instance of Utils for the current database type.
```

system/Database/Database.php
```diff
@@ -61,7 +61,7 @@ public function load(array $params = [], string $alias = '')
     /**
      * Creates a Forge instance for the current database type.
      */
-    public function loadForge(ConnectionInterface $db): object
+    public function loadForge(ConnectionInterface $db): Forge
```
MGatner, line 64: IMO this should have been `final` anyways. I'm fine with breaking it.

system/Database/Database.php
```diff
@@ -124,8 +124,10 @@ protected function parseDSN(array $params): array
      * Initialize database driver.
      *
      * @param array|object $argument
+     *
+     * @return BaseConnection|BaseUtils|Forge
      */
-    protected function initDriver(string $driver, string $class, $argument): object
+    protected function initDriver(string $driver, string $class, $argument)
```
MGatner, line 130: Why remove this? It's consistent with the docblock and natively PHP will process as the much-broader `mixed` otherwise.

MGatner (changes requested):

system/Database/Database.php
kenjis: Yes, we don't need to remove it.
Reverted.

kenjis:
At first, this PR was directed to `develop`, but since the fixes are not urgent, I changed it to `4.3` because it has breaking changes.

MGatner (approved):
Looks good! Thanks for fixing the changelogs too

PR #6511, opened 2022-09-08
fix: table view style cannot applied to element `td` on header and footer

ddevsr:
**Description**
Fixed #6507, header and footer cant applied style if element with `td`

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [x] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #6531, opened 2022-09-11
fix: table view attribute cannot applied to element `td` on header and footer

### LINKED ISSUE #6507: Bug: Table footer will not take tag attributes
### PHP Version

8.1

### CodeIgniter4 Version

4.1.5

Manual (zip or tar.gz)

Linux

### Which server did you use?

apache

### Database

MySQL

### What happened?

**\CodeIgniter\View\Table**
Table footer will not incorporate extra attributes. It assumes the footer cell is `th` rather than `td` (as specified in the default template). This means you can not style the table footer by using element attributes (e.g. class, style).

### Steps to Reproduce

```php
// style wanted for cells containing numbers 
$style = 'background:#F99;text-align:right;width:5em;';
// table data
$data = [
	['item', ['data'=>'count', 'style'=>$style]],
	['shirt', ['data'=>2, 'style'=>$style]],
	['hat', ['data'=>1, 'style'=>$style]]
];
// add footing - summary of table data
$footing = ['total', ['data'=>3, 'style'=>$style]];

echo '<p>Use default template</p>';
$table = new \CodeIgniter\View\Table();
$table->setFooting($footing);
echo $table->generate($data);

echo '<p>Update template footer to use {th} rather then {td}.</p>';
$template = [
	'footing_cell_start' => '<th>',
	'footing_cell_end' => '</th>'
];
$table = new \CodeIgniter\View\Table($template);
$table->setFooting($footing);
echo $table->generate($data);
```

### Expected Output

I expect cells in the footer to be accept attributes in the same manner as cells in the table body or table header. 

### Anything else?

line 365 of System\View\Table
`$temp = str_replace('<th', '<th ' . $key . '="' . $val . '"', $temp);`

{th} should be {td}, if the template is default

---

But, of course, what if the template has been altered by the App? This replacement method is only going to work if the template 'table_footer_start' starts with '<th'.

For now... make it work for the _default_ template. If people change the template, they probably will not use the extra attributes.

ddevsr:
**Description**
Fixed #6507
Supersedes #6511

**Checklist:**
- [ ] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/Table.php
+++ b/system/View/Table.php
@@ -1,0 +1,495 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\View;
+
+use CodeIgniter\Database\BaseResult;
+
+/**
+ * HTML Table Generating Class
+ *
+ * Lets you create tables manually or from database result objects, or arrays.
+ */
+class Table
+{
+    /**
+     * Data for table rows
+     *
+     * @var array
+     */
+    public $rows = [];
+
+    /**
+     * Data for table heading
+     *
+     * @var array
+     */
+    public $heading = [];
+
+    /**
+     * Data for table footing
+     *
+     * @var array
+     */
+    public $footing = [];
+
+    /**
+     * Whether or not to automatically create the table header
+     *
+     * @var bool
+     */
+    public $autoHeading = true;
+
+    /**
+     * Table caption
+     *
+     * @var string|null
+     */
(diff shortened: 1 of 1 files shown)


ddevsr:
image

@kenjis On PC office cant running all sign commit (rebase error), and now i am using export file key to home PC cannot sign gpg. How do I fix this? so that my key id on both pc can be used.

kenjis:
@ddevsr Export private/public keys and import the  keys.

How to export and import keys with GPG
https://linuxhint.com/export-import-keys-with-gpg/

ddevsr:
image
@kenjis still same error

ddevsr:
Okay, i am already import private yesterday. Public key soon

Pada tanggal Sel, 13 Sep 2022 05.13, kenjis ***@***.***>
menulis:

PR #6532, opened 2022-09-12
docs: split CHANGELOG.md

kenjis:
**Description**
- split CHANGELOG.md

But GitHub still shows "This blob took too long to generate."

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide


MGatner (changes requested):
I think this is the right way to do it but I'd rather not have these filling up the root. Can we keep the latest in `/` then create a new `changelogs/` directory for the rest?

kenjis:
Indeed it is. Moved to `changelogs/`.

MGatner (approved):
👌

PR #6536, opened 2022-09-12
Refactor BaseBuilder *Batch() Methods

sclubricants:
This PR aligns all the *Batch() methods allowing them to share the same code.

This solves https://github.com/codeigniter4/CodeIgniter4/issues/5674

This is part of https://github.com/codeigniter4/CodeIgniter4/pull/6407/

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


sclubricants:
@codeigniter4/database-team Thoughts?

sclubricants:
The new setData() method allows setting data in multiple rows or just one row at a time. This way it can be used as a collector in a loop.

```php
        $db = db_connect();

        $data = [
            ['id' => 1, 'value' => 11,],
            ['id' => 2, 'value' => 12,],
            ['id' => 3, 'value' => 13,],
        ];

        $oneRow =  ['id' => 4, 'value' => 14,];

        $moreData = [
            ['value' => 15, 'id' => 5,],
            ['value' => 16, 'id' => 6,],
            ['value' => 17, 'id' => 7,],
        ];

        $EvenMoreData = [
            ['value' => 18, 'id' => 8,],
            ['value' => 19, 'id' => 9,],
            ['value' => 20, 'id' => 10,],
        ];
        
        dd($db->table('test')->testMode()
            ->setData($data)
            ->setData($oneRow)
            ->setData($moreData)
            ->insertBatch($EvenMoreData, null, 1));
```

```
0 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (1,11)"
1 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (2,12)"
2 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (3,13)"
3 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (4,14)"
4 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (5,15)"
5 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (6,16)"
6 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (7,17)"
7 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (8,18)"
8 => string (51) "INSERT INTO `db_test` (`id`, `value`) VALUES (9,19)"
9 => string (52) "INSERT INTO `db_test` (`id`, `value`) VALUES (10,20)"
```

Also, if in any of the *batch() methods the batch size is set to 0 then it will process everything in one batch.

system/Database/BaseBuilder.php
```diff
@@ -2171,94 +2276,31 @@ protected function validateUpdate(): bool
...
+     *
+     * @param array|object|string|null $set         a dataset or select query
+     * @param array|string|null        $constraints
      *
      * @return false|int|string[] Number of rows affected or FALSE on failure, SQL array when testMode
      *
      * @throws DatabaseException
      */
-    public function updateBatch(?array $set = null, ?string $index = null, int $batchSize = 100)
+    public function updateBatch($set = null, $constraints = null, int $batchSize = 100)
```
michalsn, line 2288: This is probably a BC break.

michalsn (commented):
Looks good to me. Nice improvements.

Two things: 

Changing public method variable names should be considered as a BC break. Not sure what is our current policy on this, but I would say "no" - at least to the moment, it is not listed under the breaking changes in the changelog.

I’m also not a fan of changing the exception messages to `setData() has no data.`. Previously messages were way more informative. If the new user will write the code:
```
$builder->updateBatch([], 'id');
```

he will have no idea what is going on. At least, this is my impression.

Of course, we need a changelog and some updates to the user guide - with code examples. 

Nice work. Let’s see what others things about it.

system/Database/BaseBuilder.php
sclubricants: I don't see how this breaks anything. It accepts a string of a single column just like it did before.

system/Database/BaseBuilder.php
```diff
@@ -1770,76 +1899,85 @@ protected function updateFields($set, bool $addToDefault = false, ?array $ignore
...
-            if ($this->testMode) {
-                $savedSQL[] = $sql;
-            } else {
-                $this->db->query($sql, null, false);
-                $affectedRows += $this->db->affectedRows();
-            }
+    /**
+     * Converts value array of array to array of strings
+     */
+    protected function getValues(array $values): array
```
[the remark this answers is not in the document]
sclubricants: How about `formatValues()`

sclubricants:
>I’m also not a fan of changing the exception messages to setData() has no data.. Previously messages were way more informative.

This crossed my mind at one point. How about:

```php
    public function updateBatch($set = null, $constraints = null, int $batchSize = 100)
    {
        $this->onConstraint($constraints);

        if ($set !== null && $set !== []) {
            $this->setData($set, true);
        } elseif (empty($this->QBSet)) {
            if ($this->db->DBDebug) {
                throw new DatabaseException('updateBatch() has no data.');
            }

            return false; // @codeCoverageIgnore
        }

        return $this->batchExecute('_updateBatch', $batchSize);
    }
```

system/Database/BaseBuilder.php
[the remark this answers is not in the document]
michalsn: It's good too.

system/Database/BaseBuilder.php
michalsn: I'm worried about the scenario when someone would use named parameters. It's a public method.

michalsn:
Yes, that would work better.

system/Database/BaseBuilder.php
sclubricants: Well we can keep the name index.. its just seemed out of place. Do you think that would be better? Or should we add this as a breaking change?

kenjis: If you change a type definition of a public/protected method, it is a BC. Extended classes will break.

All BCs must be documented in the changelogs.
See https://codeigniter4.github.io/CodeIgniter4/installation/backward_compatibility_notes.html

kenjis: I prefer a proper name. In this case, we need to change the type definitions, they are already BCs.
So I don't mind changing the variable names along with them.

sclubricants:
Added documentation, change log, etc.

tests/system/Database/Live/UpdateTest.php
```diff
@@ -278,7 +278,7 @@ public function testUpdateBatchConstraintsRawSqlandAlias()
             [
                 'id'      => 2,
                 'name'    => 'Ahmadinejad Changes',
-                'country' => 'Iraq', // same length but different
+                'country' => 'Uruguay',
```
kenjis, line 289: Why did you remove the existing test case?

system/Database/BaseBuilder.php
```diff
@@ -2385,7 +2385,7 @@ protected function _updateBatch(string $table, array $keys, array $values): stri
                 ",\n",
                 array_map(
                     static fn ($key, $value) => $key . ($value instanceof RawSql ?
-                        ' = ' . $value :
+                        ' = ' . str_replace('%', '%%', $value) :
```
kenjis, line 2337: Why do we need to change the RawSql string?
I think if we change the string, it is not *raw*.

kenjis:
> [Changed RawSql test](https://github.com/codeigniter4/CodeIgniter4/pull/6536/commits/47fb0e378a41dd7378281fe398354680b6de3708)
> Changed RawSql test to use LIKE 'U%' This also required escaping sprintf() in _updateBatch()

This is almost what you did. We can know easily what you did from the commit,
but don't know why you changed and why it required escaping. 
Please write why in a commit messages as possible.

michalsn (changes requested):
I will have some requests involving the user guide side.

Not sure if you would be up for it, but adding some context to the example queries in the user guide - what we want to achieve in every example would be very informative. 

You unlocked new and powerful features to the Query Builder, and it would be a shame not to advertise it correctly. The more detailed information we provide about these queries, the more users will benefit - they need to understand how the new methods work and what opportunities they bring.

user_guide_src/source/database/query_builder/092.php
```diff
@@ -2,28 +2,53 @@
...
+    ->setAlias('u')
+    ->onConstraint(new RawSql('`mytable`.`title` = `u`.`title` AND `mytable`.`author` = `u`.`author`'))
+    ->updateFields(['last_update' => new RawSql('CURRENT_TIMESTAMP()')], true)
+    ->updateBatch();
+
+// OR
+foreach ($data as $row) {
+    $builder->setData($row);
+}
+$builder->onConstraint('title, author')->updateBatch();
```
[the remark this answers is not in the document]
sclubricants: All of the examples all generate the same SQL (except one additional item in one). The idea here was to demonstrate all the different ways one might construct a statement. There is still come features I want to add to this.. the use of join() and where(). When I add those I thought I might make some more examples. I though we might want to make a separate page though at some point because it could get quite lengthy. Until then I thought it would suffice to demonstrate the basic methods available. Its not suppose to be an advanced lesson in SQL but just show what is available. A second page might could show some practical examples of using advanced methods.

tests/system/Database/Live/UpdateTest.php
sclubricants: I'm not where I removed any tests. I looked back through the commits and didn't see any. Sometimes I'll rename them if I add to them. In general though I try to build on existing tests where appropriate to try and not create redundant tests. I try and get the maximum code coverage for each test.

system/Database/BaseBuilder.php
sclubricants:
```php
return sprintf($sql, $data);
```

Have to escape % which might show up in RawSql.

user_guide_src/source/database/query_builder/092.php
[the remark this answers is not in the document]
michalsn: Okay, thank you for pointing this out. I misled myself by looking at the `last_update` part in the QB, which is present in the produced query. Now I even see there is a comment in the produced SQL... 😓 

Can we sort these examples so the comment becomes accurate? Now, the last scenario doesn't have the `last_update` part in the QB.

Yes, I think more advanced examples that use these new methods (including `RawSql`) would be desirable. A separate page would be great.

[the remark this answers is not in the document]
sclubricants: Ya I added another example and forgot to reorder them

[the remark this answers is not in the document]
sclubricants: I'd like to submit some more PRs though before adding a special page of examples. I'd like to add upsert as well as some more features to insert and update.. such as joining tables, insert/update from another table or a query. I'd just like to have a full feature set and then decide how best to demonstrate everything. I definitely will add more later.. these changes will generate endless possibilities.

sclubricants:
Some things I'm pondering for later include:

As in the update query you can select data into a psuedo table. A method that compiles this and gives an alias could be created.

```SQL
(select 1 id, 'a name' name UNION ALL
select 2 id, 'another name' name) AS psuedo_table
```

We could wrap this in a RawSql object. Then this could be used anywhere instead of an actual table name. For instance with joins. You could also find a way to use it with the table() method. Then you could query the table.

The same can be done with a query:

```SQL
(Select id, name FROM mytable Where id > 5) AS psuedo_table
```

I use queries on joins all the time as well as query queries. 

These are just some things in the back of my mind. I'm trying to keep them in mind so that the things I'm doing now lend themselves to this in the future. I'd like to add upsert in first though and get all the batch methods working well together.

user_guide_src/source/database/query_builder.rst
```diff
@@ -1608,19 +1608,45 @@ Class Reference
...
+        :param int $batchSize: Count of conditions to group in a single query
         :returns:   Number of rows updated or ``false`` on failure
         :rtype:     int|false
 
         Compiles and executes batch ``UPDATE`` statements.
 
-        .. note:: When more than ``$batch_size`` field/value pairs are provided,
-            multiple queries will be executed, each handling up to
-            ``$batch_size`` field/value pairs.
+        .. note:: Set ``$batchSize`` to 0 for unlimited. ``$constraints`` takes a comma delimited string of columns, and array, associative array, or RawSql.
```
michalsn, line 1621:
```suggestion
        .. note:: When more than ``$batchSize`` field/value pairs are provided,
             multiple queries will be executed, each handling up to ``$batchSize`` 
             field/value pairs. If we set ``$batchSize`` to 0, then all field/value 
             pairs will be executed in one query.
```

We should have a full explanation of the rules for `batchSize`.

user_guide_src/source/database/query_builder.rst
```diff
@@ -1608,19 +1608,45 @@ Class Reference
...
 
         :param array $set: Field name, or an associative array of field/value pairs
-        :param string $value: Field value, if $set is a single field
-        :param int $batch_size: Count of conditions to group in a single query
+        :param mixed $constraints: The field or fields used as keys to update on.
+        :param int $batchSize: Count of conditions to group in a single query
         :returns:   Number of rows updated or ``false`` on failure
         :rtype:     int|false
 
         Compiles and executes batch ``UPDATE`` statements.
```
michalsn, line 1620:
```suggestion
The ``$constraints`` variable takes a comma delimited string of columns, an array, an associative array, or RawSql.
```

I would move this description a little higher. Not sure if it should be put into "note".

michalsn (changes requested):
Some small updates to the user guide.

system/Database/BaseBuilder.php
kenjis: If you use `sprintf()`, you must escape all `%` (except what you intended to replace), not only in RawSql strings.

Isn't it dangerous to use `sprintf()` to generate SQL strings?

tests/system/Database/Live/UpdateTest.php
```diff
@@ -238,4 +239,89 @@ public function testSetWithBoolean()
...
+            'name'    => 'Derek Jones Changes',
+            'country' => 'US',
+        ]);
+        $this->seeInDatabase('user', [
+            'name'    => 'Ahmadinejad',
+            'country' => 'Iran',
+        ]);
+    }
+
+    public function testUpdateBatchConstraintsRawSqlandAlias()
```
kenjis, line 274:
```suggestion
    public function testUpdateBatchConstraintsRawSqlAndAlias()
```

tests/system/Database/Live/UpdateTest.php
kenjis: I searched `// same length but different` in UpdateTest.php, but not found.
You had a test case `Iraq` for *same length but different*, but it was removed
because `Uruguay` is not the same length.
Is it okay?

sclubricants: I changed this test around a few times.. I was having trouble useing RawSql with the % character. Ultimately I fixed the issue and changed the test to do what I was originally wanting to do.

system/Database/BaseBuilder.php
sclubricants: I wouldn't think if you did it properly. It is used quite a bit already. Most of the code where % is likely to show up is in the data to be placed. I just had to escape the RawSql.

user_guide_src/source/database/query_builder.rst
sclubricants: Done

tests/system/Database/Live/UpdateTest.php
sclubricants: I changed to Uruguay and used LIKE 'U%'

system/Database/BaseBuilder.php
kenjis: I sent a bug report: #6555

sclubricants: I changed to str_replace. Its not often you use a % in table or column names but I suppose it could happen. I had added query caching to insertBatch as well so I went ahead and changed those as well.

michalsn (changes requested):
Probably the last round from me.

There is a missing description for the setAlias() method in the user guide. It is a public method, so we should have it in the method reference list.

I don't really like the `{:_table_:}` part, and I would prefer to see something like `{!QUERY-PART!}` or similar, but that might just be me, so.... don't feel obligated to change it.

sclubricants:
The reason I chose `table` is because this is in effect what it is. The data goes in a JOIN statement which supports joining tables. 

As I mentioned above:

```php
(select 1 id, 'a name' name UNION ALL
select 2 id, 'another name' name) AS psuedo_table
```

So the data that goes there could just as well be substituted for an actual table name.

Would you like if I changed to: `{!TABLE-PART!}` or `{!TABLE-DATA!}` or better yet `{!DATA~TABLE!}` ?

michalsn:
You know... I am not convinced of any version (even my own). Maybe kenjis will have an opinion on this - we can always stay with what we have, too.

michalsn (approved):
This PR looks good to me. Nice work.

system/Database/BaseBuilder.php
```diff
@@ -1736,97 +1737,48 @@ public function getWhere($where = null, ?int $limit = null, ?int $offset = 0, bo
...
     /**
-     * Sets update fields for updateBatch
-     *
-     * @param string|string[] $set
-     * @param bool            $addToDefault future use
-     * @param array|null      $ignore       ignores items in set
-     *
-     * @return $this
-     */
-    protected function updateFields($set, bool $addToDefault = false, ?array $ignore = null)
```
kenjis, line 1747: (memo) This method will be `public`. It is not a BC. Okay.

system/Database/BaseBuilder.php
```diff
@@ -1881,16 +1837,172 @@ public function setInsertBatch($key, string $value = '', ?bool $escape = null)
...
+    }
+
+    /**
+     * Compiles batch insert strings and runs the queries
+     *
+     * @param array|string|null $set a dataset or select query
+     *
+     * @return false|int|string[] Number of rows inserted or FALSE on failure, SQL array when testMode
+     */
+    public function insertBatch($set = null, ?bool $escape = null, int $batchSize = 100)
```
kenjis, line 1957: `public function insertBatch(?array $set = null, ?bool $escape = null, int $batchSize = 100)`

The type for `$set` has changed. This is a BC.
Should be added to the changelog.

system/Database/BaseBuilder.php
```diff
@@ -2171,94 +2283,29 @@ protected function validateUpdate(): bool
...
      *
-     * @return false|int|string[] Number of rows affected or FALSE on failure, SQL array when testMode
+     * @param array|object|string|null $set         a dataset or select query
+     * @param array|RawSql|string|null $constraints
      *
-     * @throws DatabaseException
+     * @return false|int|string[] Number of rows affected or FALSE on failure, SQL array when testMode
      */
-    public function updateBatch(?array $set = null, ?string $index = null, int $batchSize = 100)
+    public function updateBatch($set = null, $constraints = null, int $batchSize = 100)
```
kenjis, line 2293: Changed:
1. type of `$set`
2. type of `$index`
3. name of `$index`

Please write all of them in the changelog.

system/Database/BaseBuilder.php
```diff
@@ -2171,94 +2283,29 @@ protected function validateUpdate(): bool
...
-
-        return $this->testMode ? $savedSQL : $affectedRows;
+        return $this->batchExecute('_updateBatch', $batchSize);
     }
 
     /**
      * Generates a platform-specific batch update string from the supplied data
      */
-    protected function _updateBatch(string $table, array $values, string $index): string
+    protected function _updateBatch(string $table, array $keys, array $values): string
```
kenjis, line 2307: The parameter names and a type has changed.
Please add to changelog.

kenjis:
> You know... I am not convinced of any version (even my own). 

I don't like the  replacement. I would like to remove if possible.

system/Database/BaseBuilder.php
sclubricants: Yes, This method will be especially useful for upserts. It was meant to be public from the beginning.

system/Database/BaseBuilder.php
sclubricants: I added this.. it now accepts an object of a single row of data. I also had here that it accepts a string for SQL but I haven't added all that functionality yet so I removed string from it.

system/Database/BaseBuilder.php
sclubricants: This has already been written in change log. I did add the change of type of $set though.

kenjis:
Please run `composer cs-fix`.

system/Database/BaseBuilder.php
kenjis: Something like this:
```diff
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -92,6 +92,9 @@ Others
 - The return type of ``CodeIgniter\Database\Database::loadForge()`` has been changed to ``Forge``.
 - The return type of ``CodeIgniter\Database\Database::loadUtils()`` has been changed to ``BaseUtils``.
 - Parameter ``$column`` has changed in ``Table::dropForeignKey()`` to ``$foreignName``.
+- ``BaseBuilder::_updateBatch()``
+    - The second parameter ``$values`` has changed to ``$keys``.
+    - The third parameter ``$index`` has changed to ``$values``. The parameter type also has changed to ``array``.
 
 Enhancements
 ************
```

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -46,6 +46,8 @@ Others
 - ``CITestStreamFilter::$buffer = ''`` no longer causes the filter to be registered to listen for streams. Now there
   is a ``CITestStreamFilter::registration()`` method for this. See :ref:`upgrade-430-stream-filter` for details.
 - The data structure returned by ``BaseConnection::getForeignKeyData()`` has been changed.
+- The second parameter ``$index`` of ``BaseBuilder::updateBatch()`` has changed to ``$constraints``. It now accepts an array, string, or ``RawSql``.
+- The ``$set`` parameter of ``BaseBuilder::insertBatch()`` and ``BaseBuilder::updateBatch()`` now accepts an object of a single row of data.
```
kenjis, line 50: Please move these to `Method Signature Changes` section.

user_guide_src/source/installation/upgrade_430.rst
```diff
@@ -159,6 +159,7 @@ Breaking Enhancements
 - Since the launch of Spark Commands was extracted from ``CodeIgniter\CodeIgniter``, there may be problems running these commands if the ``Services::codeigniter()`` service has been overridden.
 - The return type of ``CodeIgniter\Database\Database::loadForge()`` has been changed to ``Forge``. Extending classes should likewise change the type.
 - The return type of ``CodeIgniter\Database\Database::loadUtils()`` has been changed to ``BaseUtils``. Extending classes should likewise change the type.
+- The second parameter ``$index`` of ``BaseBuilder::updateBatch()`` has changed to ``$constraints``. It now accepts types array, string, or ``RawSql``. Extending classes should likewise change types.
```
kenjis, line 162: Please add about `insertBatch()` and `_updateBatch()`.

kenjis (changes requested):
Please fix coding style, and add a few documentation items.
See inline comments.
That's all. We are very close to completion.

sclubricants:
@kenjis I also marked `setUpdateBatch()` and `setInsertBatch()` as depreciated and removed them from documentation. These of course are replaced by `setData()` and I made note in change log.

user_guide_src/source/database/query_builder.rst
```diff
@@ -1588,16 +1588,6 @@ Class Reference
...
-    .. php:method:: setInsertBatch($key[, $value = ''[, $escape = null]])
-
-        :param mixed $key: Field name or an array of field/value pairs
-        :param string $value: Field value, if $key is a single field
-        :param bool $escape: Whether to escape values
-        :returns:   ``BaseBuilder`` instance (method chaining)
-        :rtype:     ``BaseBuilder``
-
-        Adds field/value pairs to be inserted in a table later via ``insertBatch()``.
-
```
kenjis, line 1600: Revert this, and add a note.
```
.. important:: This method is deprecated. It will be removed in future releases.
```

user_guide_src/source/database/query_builder.rst
```diff
@@ -1660,16 +1650,6 @@ Class Reference
...
-    .. php:method:: setUpdateBatch($key[, $value = ''[, $escape = null]])
-
-        :param mixed $key: Field name or an array of field/value pairs
-        :param string $value: Field value, if $key is a single field
-        :param bool    $escape: Whether to escape values
-        :returns:   ``BaseBuilder`` instance (method chaining)
-        :rtype:     ``BaseBuilder``
-
-        Adds field/value pairs to be updated in a table later via ``updateBatch()``.
-
```
kenjis, line 1634: Revert this, and add a note.
```
.. important:: This method is deprecated. It will be removed in future releases.
```

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -174,6 +179,7 @@ Deprecations
 
 - ``RouteCollection::localizeRoute()`` is deprecated.
 - ``RouteCollection::fillRouteParams()`` is deprecated. Use ``RouteCollection::buildReverseRoute()`` instead.
+- ``BaseBuilder::setUpdateBatch()`` and ``BaseBuilder::setInsertBatch()`` are depreciated. Use ``BaseBuilder::setData()`` instead.
```
kenjis, line 182:
```suggestion
- ``BaseBuilder::setUpdateBatch()`` and ``BaseBuilder::setInsertBatch()`` are deprecated. Use ``BaseBuilder::setData()`` instead.
```

user_guide_src/source/installation/upgrade_430.rst
kenjis: Please add:
```
- The ``$set`` parameter of ``BaseBuilder::insertBatch()`` and ``BaseBuilder::updateBatch()`` now accepts an object of a single row of data. Extending classes should likewise change the type.
- The third parameter ``$index`` of ``BaseBuilder::_updateBatch()`` has changed to ``$values``, and the parameter type has changed to ``array``. Extending classes should likewise change the type.
```

user_guide_src/source/database/query_builder.rst
sclubricants: Done

user_guide_src/source/installation/upgrade_430.rst
kenjis: @sclubricants This is the last. Please add above.

sclubricants: Done

kenjis (approved):
Thank you!

MGatner:
🥳🎉🎊

PR #6537, opened 2022-09-13
Update coding-standards version

paulbalandan:
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -38,71 +38,7 @@ $finder = Finder::create()
         __DIR__ . '/spark',
     ]);
 
-$overrides = [
-    // <<<<<<<<<<<<<<<<<<<<<<<< @TODO TO BE REMOVED ONCE LIVE IN CODING-STANDARD
-    'blank_line_between_import_groups' => true,
-    'class_definition'                 => [
-        'multi_line_extends_each_single_line' => true,
-        'single_item_single_line'             => true,
-        'single_line'                         => true,
-        'space_before_parenthesis'            => true,
-        'inline_constructor_arguments'        => true,
-    ],
-    'control_structure_braces' => true,
-    'curly_braces_position'    => [
-        'control_structures_opening_brace'          => 'same_line',
-        'functions_opening_brace'                   => 'next_line_unless_newline_at_signature_end',
-        'anonymous_functions_opening_brace'         => 'same_line',
-        'classes_opening_brace'                     => 'next_line_unless_newline_at_signature_end',
-        'anonymous_classes_opening_brace'           => 'same_line',
-        'allow_single_line_empty_anonymous_classes' => true,
-        'allow_single_line_anonymous_functions'     => true,
-    ],
-    'date_time_create_from_format_call' => false,
-    'function_declaration'              => [
-        'closure_function_spacing'   => 'one',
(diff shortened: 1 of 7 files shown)


kenjis:
It seem better to change the sample code.

```diff
    1) CodeIgniter4/user_guide_src/source/outgoing/response/018.php (date_time_create_from_format_call)
      ---------- begin diff ----------
--- /home/runner/work/CodeIgniter4/CodeIgniter4/user_guide_src/source/outgoing/response/018.php
+++ /home/runner/work/CodeIgniter4/CodeIgniter4/user_guide_src/source/outgoing/response/018.php
@@ -1,4 +1,4 @@
 <?php
 
-$date = DateTime::createFromFormat('j-M-Y', '15-Feb-2016');
+$date = DateTime::createFromFormat('!j-M-Y', '15-Feb-2016');
 $response->setDate($date);

      ----------- end diff -----------
```

MGatner:
Yes I am in favor of updating the examples so they always match our coding standard; that was the primary reason behind splitting them out.

I wish there were a way to exempt `'U'` from the `!` rule; I think it is better to have it than dropping the rule, but it is a logical redundancy.

MGatner:
@paulbalandan Could you make the parallel changes over at **codeigniter4/devkit** as well please?

paulbalandan:
> @paulbalandan Could you make the parallel changes over at **codeigniter4/devkit** as well please?

Done. https://github.com/codeigniter4/devkit/pull/47

user_guide_src/source/outgoing/response/018.php
```diff
@@ -1,4 +1,4 @@
 <?php
 
-$date = DateTime::createFromFormat('j-M-Y', '15-Feb-2016');
+$date = DateTime::createFromFormat('!j-M-Y', '15-Feb-2016');
 $response->setDate($date);
```
kenjis, line 4: What is the use case of this `Response::setDate()` method?
I feel that there is no need for it to be `public`.

MGatner (approved):

user_guide_src/source/outgoing/response/018.php
MGatner: I could see it being helpful for functional or E2E testing where you want to "spoof" a past/future date. That's a bit of a reach though, bc I'm unclear how that would ever affect an outcome. 🤷‍♂️

kenjis: If you want to do E2E testing on a date in the past, you need a system clock (which provides the current time) and you must use it in all of your application (and framework).

kenjis: @paulbalandan @MGatner Why don't we just remove this sample code?
This method is not likely to be used (and it is easy to use if you want to use).

MGatner: Yeah, that sounds best.

paulbalandan: If this sample code would be removed, should the next samples be renumbered?

kenjis: We don't change the sample file names. 
If we want to renumber, we should do it to all files at some point.

kenjis (approved):

PR #6544, opened 2022-09-15
fix: set_cookie() does not use Config\Cookie values

### LINKED ISSUE #6540: Bug: set_cookie not using default value of $secure from config
### PHP Version

7.4

### CodeIgniter4 Version

4.2.6

Manual (zip or tar.gz)

macOS

### Which server did you use?

apache

### Database

_No response_

### What happened?

set_cookie function in Helpers/cookie_helper.php is not using the default value $secure i specified in app/Config/Cookie.php. 

This used to work in CI3 because I think the default value of $secure param of set_cookie function used to be null which later on fallback to default.****

### Steps to Reproduce

1. Set public $secure = true in /app/Config/Cookie.php'
2. Call the helper function set_cookie
   ```php
        helper('cookie');
        set_cookie([
            'name' => 'pmw',
            'value' => 2,
            'samesite' => 'None',
        ]);
   ```
4. CookieException::forInvalidSameSiteNone() is thrown although $secure was suppose to be true

### Expected Output

No exception

### Anything else?

_No response_

kenjis:
**Description**
Fixes #6540

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/ResponseTrait.php
+++ b/system/HTTP/ResponseTrait.php
@@ -571,7 +571,7 @@ trait ResponseTrait
         /** @var CookieConfig|null $cookieConfig */
         $cookieConfig = config('Cookie');
 
-        if ($cookieConfig) {
+        if ($cookieConfig instanceof CookieConfig) {
             $secure ??= $cookieConfig->secure;
             $httponly ??= $cookieConfig->httponly;
             $samesite ??= $cookieConfig->samesite;
--- a/system/HTTP/ResponseTrait.php
+++ b/system/HTTP/ResponseTrait.php
@@ -17,6 +17,7 @@ use CodeIgniter\Cookie\Exceptions\CookieException;
 use CodeIgniter\HTTP\Exceptions\HTTPException;
 use CodeIgniter\Pager\PagerInterface;
 use CodeIgniter\Security\Exceptions\SecurityException;
+use Config\Cookie as CookieConfig;
 use Config\Services;
 use DateTime;
 use DateTimeZone;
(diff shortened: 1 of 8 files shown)


MGatner (approved):

system/HTTP/ResponseTrait.php
```diff
@@ -567,8 +568,17 @@ public function setCookie(
             return $this;
         }
 
+        /** @var CookieConfig|null $cookieConfig */
+        $cookieConfig = config('Cookie');
+
+        if ($cookieConfig) {
```
samsonasik, line 574: using instanceof is better  I think, that's we know that type is exactly a CookieConfig:

```suggestion
        if ($cookieConfig instanceof CookieConfig) {
```

tests/system/HTTP/ResponseCookieTest.php
```diff
@@ -255,6 +256,7 @@ public function testCookieStrictSetSameSite()
 
     public function testCookieBlankSetSameSite()
     {
+        /** @var \Config\Cookie $config */
```
samsonasik, line 259: there is alias for it, so use `CookieConfig` for shorter:

```suggestion
        /** @var CookieConfig $config */
```

tests/system/HTTP/ResponseCookieTest.php
```diff
@@ -314,6 +316,30 @@ public function testCookieInvalidSameSite()
         ]);
     }
 
+    public function testSetCookieConfigCookieIsUsed()
+    {
+        /** @var \Config\Cookie $config */
```
samsonasik, line 321: there is alias for it, so use `CookieConfig` for shorter:

```suggestion
        /** @var CookieConfig $config */
```

tests/system/HTTP/ResponseCookieTest.php
kenjis: Done.

tests/system/HTTP/ResponseCookieTest.php
kenjis: Done.

samsonasik (approved):

PR #6550, opened 2022-09-17
docs: replace types in `View\Parser` and `View\Filters`

ddevsr:
**Description**
See #6310

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/Filters.php
+++ b/system/View/Filters.php
@@ -208,7 +208,7 @@ class Filters
      *  - ceil      always rounds up
      *  - floor     always rounds down
      *
-     * @param float|int|string $precision
+     * @param int|string $precision
      *
      * @return float|string
      */
--- a/system/View/Filters.php
+++ b/system/View/Filters.php
@@ -62,7 +62,7 @@ class Filters
     /**
      * Returns the given default value if $value is empty or undefined.
      *
-     * @param int|string|null $value
+     * @param array|bool|float|int|object|resource|string|null $value
      */
     public static function default($value, string $default): string
     {
--- a/system/View/Filters.php
+++ b/system/View/Filters.php
@@ -214,7 +214,6 @@ class Filters
      */
     public static function round(string $value, $precision = 2, string $type = 'common')
     {
-        $testGatau = 1337.0;
         if (! is_numeric($precision)) {
             $type      = $precision;
             $precision = 2;
--- a/system/View/Filters.php
+++ b/system/View/Filters.php
@@ -30,7 +30,7 @@ class Filters
     /**
      * Formats a date into the given $format.
      *
-     * @param mixed $value
+     * @param int|string|null $value
      */
     public static function date($value, string $format): string
     {
@@ -48,7 +48,7 @@ class Filters
      * Example:
      *      my_date|date_modify(+1 day)
      *
-     * @param string $value
+     * @param int|string|null $value
      *
      * @return false|int
      */
@@ -62,7 +62,7 @@ class Filters
     /**
      * Returns the given default value if $value is empty or undefined.
      *
-     * @param mixed $value
+     * @param int|string|null $value
      */
     public static function default($value, string $default): string
     {
@@ -208,12 +208,13 @@ class Filters
      *  - ceil      always rounds up
      *  - floor     always rounds down
      *
-     * @param mixed $precision
+     * @param float|int|string $precision
      *
      * @return float|string
      */
     public static function round(string $value, $precision = 2, string $type = 'common')
     {
+        $testGatau = 1337.0;
         if (! is_numeric($precision)) {
             $type      = $precision;
             $precision = 2;
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -72,8 +72,10 @@ class Parser extends View
 
     /**
      * Constructor
+     *
+     * @param FileLocator|null $loader
      */
-    public function __construct(ViewConfig $config, ?string $viewPath = null, ?FileLocator $loader = null, ?bool $debug = null, ?LoggerInterface $logger = null)
+    public function __construct(ViewConfig $config, ?string $viewPath = null, $loader = null, ?bool $debug = null, ?LoggerInterface $logger = null)
     {
         // Ensure user plugins override core plugins.
         $this->plugins = $config->plugins ?? [];
--- a/system/View/Parser.php
+++ b/system/View/Parser.php
@@ -72,11 +72,8 @@ class Parser extends View
 
     /**
      * Constructor
-     *
-     * @param string           $viewPath
-     * @param FileLocator|null $loader
      */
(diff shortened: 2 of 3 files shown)


system/View/Filters.php
```diff
@@ -208,12 +208,13 @@ public static function prose(string $value): string
...
      *  - floor     always rounds down
      *
-     * @param mixed $precision
+     * @param float|int|string $precision
      *
      * @return float|string
      */
     public static function round(string $value, $precision = 2, string $type = 'common')
     {
+        $testGatau = 1337.0;
```
paulbalandan, line 217: What is this?

system/View/Parser.php
```diff
@@ -155,7 +148,6 @@ public function render(string $view, ?array $options = null, ?bool $saveData = n
      * replacing them with any data that has already been set.
      *
      * @param array $options
```
paulbalandan, line 150: This should likewise be removed.

system/View/Filters.php
ddevsr: Sorry, forgot delete

system/View/Parser.php
```diff
@@ -71,13 +72,8 @@ class Parser extends View
...
     /**
      * Constructor
-     *
-     * @param string          $viewPath
-     * @param mixed           $loader
-     * @param bool            $debug
-     * @param LoggerInterface $logger
      */
-    public function __construct(ViewConfig $config, ?string $viewPath = null, $loader = null, ?bool $debug = null, ?LoggerInterface $logger = null)
+    public function __construct(ViewConfig $config, ?string $viewPath = null, ?FileLocator $loader = null, ?bool $debug = null, ?LoggerInterface $logger = null)
```
kenjis, line 76: Changing type declaration is a breaking change. It is not acceptable in the patch release.
Please revert.

ddevsr: Done

system/View/Filters.php
```diff
@@ -62,7 +62,7 @@ public static function date_modify($value, string $adjustment)
     /**
      * Returns the given default value if $value is empty or undefined.
      *
-     * @param mixed $value
+     * @param int|string|null $value
```
MGatner, line 65: This should support a lot more; `mixed` might have even been correct here.

system/View/Filters.php
```diff
@@ -208,7 +208,7 @@ public static function prose(string $value): string
      *  - ceil      always rounds up
      *  - floor     always rounds down
      *
-     * @param mixed $precision
+     * @param float|int|string $precision
```
MGatner, line 211: Precision should never be a float.
```suggestion
     * @param int|string $precision
```

system/View/Parser.php
```diff
@@ -698,9 +690,9 @@ public function removePlugin(string $alias)
      * Converts an object to an array, respecting any
      * toArray() methods on an object.
      *
-     * @param mixed $value
+     * @param object $value
      *
-     * @return mixed
+     * @return array
```
MGatner, line 695: You would think these should be the types but this method appears to be "change objects to arrays, leave everything else". This fall-through effect might be intentional; I would like to see more research before changing the types based solely on this.

MGatner (changes requested):
Good changes! A few issues that need attention. The last one (`objectToArray()`) will be hard for me to look into on mobile, so please let me know if you have already examined the uses or I will dig later on desktop.

system/View/Filters.php
kenjis: If it is really `mixed`, please use `object|resource|array|string|int|float|bool|null` for now.
Because there are too many `mixed` in the code base, and it is difficult to know it is really `mixed` or just not yet fixed.

system/View/Filters.php
ddevsr: Done

system/View/Parser.php
ddevsr: Better `array|float|int|object|string $value`?

- When `$value` is `object` and have method `toArray` return array
- When `$value` is `object` and not have method `toArray` return array
- Or accept else $value

I already push sample test addition

system/View/Filters.php
ddevsr:
```php
    public function testDefaultManyValue()
    {
        $parser = new Parser($this->config, $this->viewsDir, $this->loader);

        $obj = (object) '';

        $data = [
            'value1' => 0.0,
            'value2' => [],
            'value3' => $obj->scalar,
        ];

        $template = '{ value1|default(foo) } { value2|default(bar) } { value3|default(baz) }';

        $parser->setData($data);
        $this->assertSame('foo bar baz', $parser->renderString($template));
    }
```
I already tests with above, and result error in array empty.

```diff
There was 1 failure:

1) CodeIgniter\View\ParserFilterTest::testDefaultManyValue
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'foo bar baz'
+'foo { value2|default(bar) } baz'

D:\Project\laragon\www\ci4\tests\system\View\ParserFilterTest.php:136
phpvfscomposer://D:\Project\laragon\www\ci4\vendor\phpunit\phpunit\phpunit:97

FAILURES!
Tests: 26, Assertions: 26, Failures: 1.
```

MGatner: That's odd! `empty([])` is definitely true - I wonder if this is somehow getting interpolated as `empty[''])` (which is false)? I think that we still allow `array` types here, and address this as a bug in another PR.

ddevsr: Okay, already push doctype `default()`

system/View/Parser.php
MGatner: The problem is that this is called by `setData()` where the input array `$data` is not typed, and it pass _every_ value into this method. So this method needs to be able to accept `mixed` and return everything **except** objects. Something like this:
```php
 * @param object|resource|array|string|int|float|bool|null $value
 *
 * @return resource|array|string|int|float|bool|null
 */
```
Some of those wouldn't make sense to pass to `setData()` though, so maybe we can drop `resource` and `array`? @kenjis?

system/View/Parser.php
```diff
@@ -447,7 +447,7 @@ protected function parseConditionals(string $template): string
         ob_start();
 
         if ($this->tempData === null) {
-            $this->tempData = $this->data;
+            $this->tempData = $this->data; // @codeCoverageIgnore
```
kenjis, line 450: Why `@codeCoverageIgnore` ?

system/View/Parser.php
kenjis: At least `resource` must not be needed.

MGatner: Yes that looks good to me!

Yes, I think this could have been named much better, but we're stuck with it for now.

kenjis (approved):
LGTM!

MGatner (approved):
Me too! Thanks for this attention, and the work to get it ready 👌

PR #6556, opened 2022-09-20
fix: ResponseInterface (1)

### LINKED ISSUE #5351: Bug: Undefined method 'setBody'.
### PHP Version

7.4

### CodeIgniter4 Version

4.1.5

Windows

### Which server did you use?

### Database

MariaDB

### What happened?

image

### Steps to Reproduce

using "intelephense"
and this my code

```php
return $this->response
            ->setStatusCode(200)
            ->setContentType($file->getMimeType())
            ->setBody(file_get_contents($img))
            ->send();
```

### Expected Output

The notice should not appear.

### Anything else?

_No response_

kenjis:
**Description**
Ref #4356
Fixes #5351

- fix `ResponseInterface`
- fix `MessageInterface`
- deprecate
   - `CodeIgniter::$useSafeOutput`
   - `CodeIgniter::useSafeOutput()`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -138,6 +138,8 @@ class CodeIgniter
      * to keep from setting headers/cookies/etc
      *
      * @var bool
+     *
+     * @deprecated No longer used.
      */
     protected $useSafeOutput = false;
 
@@ -340,7 +342,7 @@ class CodeIgniter
                 return $response;
             }
 
-            $this->response->pretend($this->useSafeOutput)->send();
+            $this->response->send();
             $this->callExit(EXIT_SUCCESS);
 
             return;
@@ -371,6 +373,8 @@ class CodeIgniter
      * not complain when ini_set() function is used.
      *
      * @return $this
+     *
+     * @deprecated No longer used.
      */
     public function useSafeOutput(bool $safe = true)
     {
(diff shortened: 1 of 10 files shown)


system/CodeIgniter.php
```diff
@@ -1057,7 +1061,7 @@ public function spoofRequestMethod()
      */
     protected function sendResponse()
```
MGatner, line 1062: Can we deprecate this method since now it is literally just "respond send"?

system/HTTP/Response.php
```diff
@@ -188,6 +188,7 @@ public function __construct($config)
 
     /**
      * Turns "pretend" mode on or off to aid in testing.
+     *
      * Note that this is not a part of the interface so
      * should not be relied on outside of internal testing.
```
MGatner, line 193: Let's start annotating these `@testTag` to indicate the class/method is only for testing (following this convention: https://github.com/DaveLiddament/php-language-extensions#testtag). Once we have Attributes we can decide if we want to elevate these to an actual attribute - but at least we can start accumulating them now.

I could probably craft a custom PHPStan rule to work with the annotation in the meantime as well.

MGatner (changes requested):
Much better separation of concerns! A few suggestions.

system/HTTP/Response.php
kenjis: Added `@testTa`

system/CodeIgniter.php
kenjis: I think it is better the CodeIgniter class has the method, this level of abstraction.

MGatner (approved):

kenjis:
Added the docs.

I've found this PR is not enough to fix ResponseInterface.
But it seems a lot of changes will be needed, so I'm going to another PR.

MGatner (approved):
Docs look good

PR #6559, opened 2022-09-21
test: add tests for required_without

kenjis:
**Description**
To confirm the current behavior.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Validation/RulesTest.php
+++ b/tests/system/Validation/RulesTest.php
@@ -639,4 +639,97 @@ class RulesTest extends CIUnitTestCase
             ],
         ];
     }
+
+    /**
+     * @dataProvider requiredWithoutMultipleProvider
+     */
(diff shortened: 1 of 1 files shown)


MGatner (approved):
I'm not saying yet whether I agree or not with the other PR but it is good to have tests for existing behavior.

kenjis:
> required_without
The field is required when all of the other fields are present in the data but not required.
https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html#available-rules

It seems the explanation in the user guide is not correct, or I can't get it.
Is it correct?

How about this?
> The field is required when any of the other fields do not pass `required` checks.

michalsn (approved):
> The field is required when any of the other fields do not pass required checks.

Sounds correct. The user guide description is very unfortunate.

MGatner:
Yes yours is much better.

PR #6566, opened 2022-09-22
refactor: make $precision int in View Filter round

kenjis:
**Description**
The second parameter of PHP function `round()` should be int.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner (approved):
I just made a very similar fix at my day job! A lot of people also don't realize that `round()` returns a `float`.

PR #6568, opened 2022-09-22
Added view() method to route collections

kenjis:
**Description**
#6525 again.

Added a new method to simplify rendering out views directly through routes. While you could do this with a closure, this is tidier and easier to read when using many of these.

This supports static info pages in sites, like About, Privacy, Terms, etc. It also supports a more view-first workflow, where the view is primary object, and View Cells are used to display the dynamic content.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/incoming/routing.rst
+++ b/user_guide_src/source/incoming/routing.rst
@@ -217,7 +217,7 @@ Views
 
 If you just want to render a view out that has no logic associated with it, you can use the ``view()`` method.
 This is always treated as GET request.
-This method accepts the name of the view to load as the ``$to`` parameter.
+This method accepts the name of the view to load as the second parameter.
 
 .. literalinclude:: routing/065.php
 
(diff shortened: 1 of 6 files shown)


user_guide_src/source/incoming/routing.rst
```diff
@@ -206,9 +206,24 @@ Closures
 
 You can use an anonymous function, or Closure, as the destination that a route maps to. This function will be
 executed when the user visits that URI. This is handy for quickly executing small tasks, or even just showing
-a simple view:
+a simple view. These are always treated as GET requests.
 
-.. literalinclude:: routing/020.php
+.. literalinclude:: routing/065.php
```
kenjis, line 211: Here is the description for Closures.

lonnieezell (approved):
This is fine to me. No matter how much I use that types of functions in Javascript, for some reason in PHP I find it not quite as easy to read. It works fine, though.

kenjis:
It does not seem to be communicated well. So I committed the fix.
Maybe the commits were unintentional and broken. 
The sample code had been changed to something apparently irrelevant. See the screenshot.

And there is no `$to` parameter, so I also changed it to *the second parameter*.

**Before:**
Screenshot 2022-09-23 14 15 52

**After:**
Screenshot 2022-09-23 14 25 44

MGatner (approved):
Glanced through the changes, all looks good.

PR #6329, opened 2022-08-02
Adds a when() method to the query builder.

lonnieezell:
**Description**

This adds a method, `when()` to the Query Builder that can conditionally modify queries based on the truthiness of a given condition. 

**Checklist:**
- [X] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [x] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #6574, opened 2022-09-23
When/WhenNot methods for db in a trait

lonnieezell:
This replaces #6329. It adds a new `ConditionalTrait` to CodeIgniter that can be used to add `when()` and `whenNot()` methods in various places in the framework. This PR adds it to the database layer for inline conditions to queries. 

Instead of: 

```php
$query = $this->where('foo', $foo);

if ($bar) {
    $query = $query->where('bar', $bar);
}

return $query->findAll();
```

you can now do: 

```php
return $query->where('foo', $foo)
    ->when($bar, static function($query) {
        return $query->where('bar', $bar);
    })
    ->findAll();
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Traits/ConditionalTrait.php
+++ b/system/Traits/ConditionalTrait.php
@@ -1,0 +1,47 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Traits;
+
+trait ConditionalTrait
+{
+    /**
+     * Only runs the query when $condition evaluates to true
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
+     */
+    public function when($condition, callable $callback, ?callable $defaultCallback = null): self
+    {
+        if ($condition) {
+            $callback($this, $condition);
+        } elseif ($defaultCallback) {
+            $defaultCallback($this);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Only runs the query when $condition evaluates to false
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
+     */
+    public function whenNot($condition, callable $callback, ?callable $defaultCallback = null): self
+    {
+        if (! $condition) {
+            $callback($this, $condition);
+        } elseif ($defaultCallback) {
+            $defaultCallback($this);
+        }
+
+        return $this;
+    }
+}
--- a/system/Traits/ConditionalTrait.php
+++ b/system/Traits/ConditionalTrait.php
@@ -17,6 +17,8 @@ trait ConditionalTrait
      * Only runs the query when $condition evaluates to true
      *
      * @param array|bool|float|int|object|resource|string|null $condition
+     *
+     * @return $this
      */
     public function when($condition, callable $callback, ?callable $defaultCallback = null): self
     {
@@ -33,6 +35,8 @@ trait ConditionalTrait
      * Only runs the query when $condition evaluates to false
      *
      * @param array|bool|float|int|object|resource|string|null $condition
+     *
+     * @return $this
      */
     public function whenNot($condition, callable $callback, ?callable $defaultCallback = null): self
     {
--- a/system/Traits/ConditionalTrait.php
+++ b/system/Traits/ConditionalTrait.php
@@ -16,6 +16,12 @@ trait ConditionalTrait
     /**
      * Only runs the query when $condition evaluates to true
      *
+     * @template TWhen of mixed
+     *
+     * @phpstan-param TWhen                         $condition
+     * @phpstan-param callable(self, TWhen): mixed  $callback
+     * @phpstan-param (callable(self): mixed)|null  $defaultCallback
+     *
      * @param array|bool|float|int|object|resource|string|null $condition
      *
      * @return $this
@@ -39,6 +45,8 @@ trait ConditionalTrait
      * @phpstan-param TWhenNot                        $condition
      * @phpstan-param callable(self, TWhenNot): mixed $callback
      * @phpstan-param (callable(self): mixed)|null    $defaultCallback
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
      *
      * @return $this
      */
--- a/system/Traits/ConditionalTrait.php
+++ b/system/Traits/ConditionalTrait.php
@@ -21,7 +21,6 @@ trait ConditionalTrait
      * @phpstan-param TWhen                         $condition
      * @phpstan-param callable(self, TWhen): mixed  $callback
      * @phpstan-param (callable(self): mixed)|null  $defaultCallback
-     *
      * @param array|bool|float|int|object|resource|string|null $condition
      *
      * @return $this
@@ -45,7 +44,6 @@ trait ConditionalTrait
      * @phpstan-param TWhenNot                        $condition
      * @phpstan-param callable(self, TWhenNot): mixed $callback
      * @phpstan-param (callable(self): mixed)|null    $defaultCallback
-     *
      * @param array|bool|float|int|object|resource|string|null $condition
      *
      * @return $this
--- a/system/Traits/ConditionalTrait.php
+++ b/system/Traits/ConditionalTrait.php
@@ -34,7 +34,11 @@ trait ConditionalTrait
     /**
      * Only runs the query when $condition evaluates to false
      *
-     * @param array|bool|float|int|object|resource|string|null $condition
+     * @template TWhenNot of mixed
+     *
+     * @phpstan-param TWhenNot                        $condition
+     * @phpstan-param callable(self, TWhenNot): mixed $callback
+     * @phpstan-param (callable(self): mixed)|null    $defaultCallback
      *
      * @return $this
      */
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -126,6 +126,7 @@ Database
 - Improved the SQL structure for ``Builder::updateBatch()``. See :ref:`update-batch` for the details.
 - Improved data returned by ``BaseConnection::getForeignKeyData()``. All DBMS returns the same structure.
 - ``Forge::addForeignKey()`` now includes a name parameter to manual set foreign key names. Not supported in SQLite3.
+- Added ``when()`` and ``whenNot()` methods to conditionally add clauses to the query. See :ref:`BaseBuilder::when() <db-builder-when>` for details.
 
 Model
 =====
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -126,7 +126,7 @@ Database
 - Improved the SQL structure for ``Builder::updateBatch()``. See :ref:`update-batch` for the details.
 - Improved data returned by ``BaseConnection::getForeignKeyData()``. All DBMS returns the same structure.
 - ``Forge::addForeignKey()`` now includes a name parameter to manual set foreign key names. Not supported in SQLite3.
-- Added ``when()`` and ``whenNot()` methods to conditionally add clauses to the query. See :ref:`BaseBuilder::when() <db-builder-when>` for details.
+- Added ``when()`` and ``whenNot()`` methods to conditionally add clauses to the query. See :ref:`BaseBuilder::when() <db-builder-when>` for details.
 
 Model
 =====
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -1022,6 +1022,42 @@ that it produces a **DELETE** SQL string instead of an **INSERT** SQL string.
 
 For more information view documentation for ``$builder->getCompiledInsert()``.
 
+**********************
+Conditional Statements
+**********************
+
+.. _db-builder-when:
+
+$builder->when()
+------------------
+
+This allows modifying the query based on a condition without breaking out of the
+query builder chain. The first parameter is the condition, and it should evaluate
+to a boolean. The second parameter is a callable that will be ran
+when the condition is true.
+
+For example, you might only want to apply a given WHERE statement based on the
+value sent within an HTTP request:
+
+.. literalinclude:: query_builder/105.php
+
+Since the condition is evaluated as ``true``, the callable will be called. The value
+set in the condition will be passed as the second parameter to the callable so it
+can be used in the query.
+
+Sometimes you might want to apply a different statement if the condition evaluates to false.
+This can be accomplished by providing a second closure:
+
+.. literalinclude:: query_builder/106.php
+
+$builder->whenNot()
+-------------------
+
+This works exactly the same way as ``$builder->when()`` except that it will
+only run the callable when the condition evaluates to ``false``, instead of ``true`` like ``when()``.
+
+.. literalinclude:: query_builder/107.php
+
 ***************
 Method Chaining
 ***************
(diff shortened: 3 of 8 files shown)


user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -109,6 +109,7 @@ Database
 - ``BaseConnection::escape()`` now excludes the ``RawSql`` data type. This allows passing SQL strings into data.
 - The new method ``Forge::dropPrimaryKey()`` allows dropping the primary key on a table. See :ref:`dropping-a-primary-key`.
 - Improved the SQL structure for ``Builder::updateBatch()``. See :ref:`update-batch` for the details.
+- Added ``when()`` and ``whenNot()` methods to conditionally add clauses to the query. See :ref:`BaseBuilder::when() <db-builder-when>` for details.
```
kenjis, line 112:
```suggestion
- Added ``when()`` and ``whenNot()`` methods to conditionally add clauses to the query. See :ref:`BaseBuilder::when() <db-builder-when>` for details.
```

user_guide_src/source/database/query_builder/107.php
```diff
@@ -0,0 +1,7 @@
+$status = service('request')->getPost('status');
```
kenjis, line 1:
```suggestion
<?php

$status = service('request')->getPost('status');
```

iRedds:
What do you think of this kind of code?

```php
function when ($when, callable $true, ?callable $false = null)
{
	($when ? $true : $false ?? function(){})($this, $when);
}

function whenNot ($when, callable $true, ?callable $false = null)
{
	when($when, $false ?? function(){}, $true);
}
```

system/Traits/ConditionalTrait.php
```diff
@@ -0,0 +1,47 @@
...
+namespace CodeIgniter\Traits;
+
+trait ConditionalTrait
+{
+    /**
+     * Only runs the query when $condition evaluates to true
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
+     */
+    public function when($condition, callable $callback, ?callable $defaultCallback = null): self
```
kenjis, line 21: It seems `static` is better, and please add `@return $this`.

system/Traits/ConditionalTrait.php
```diff
@@ -0,0 +1,47 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Only runs the query when $condition evaluates to false
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
+     */
+    public function whenNot($condition, callable $callback, ?callable $defaultCallback = null): self
```
kenjis, line 37: The same as above.

system/Traits/ConditionalTrait.php
MGatner, line 21:
```suggestion
     * @param array|bool|float|int|object|resource|string|null $condition
     *
     * @return $this
     */
    public function when($condition, callable $callback, ?callable $defaultCallback = null): self
```

system/Traits/ConditionalTrait.php
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        return $this;
+    }
+
+    /**
+     * Only runs the query when $condition evaluates to false
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
+     */
```
MGatner, line 36:
```suggestion
     * @param array|bool|float|int|object|resource|string|null $condition
     *
     * @return $this
     */
```

MGatner (commented):
I like this change! Good suggestion from @iRedds. A agree with @kenjis' tweaks but otherwise approve.

MGatner:
> What do you think of this kind of code?

@iRedds I like the efficiency but I had to read it like five times to figure it out; I think @lonnieezell's version is much easier to understand without galaxy brain 😉

I do like `whenNot()` simply being an inversion of `when()` instead of a repeat of the code - less to break in future changes. That would be a good change to apply here.

kenjis:
@iRedds To be honest, the code is difficult to understand.

> Code should be written to minimize the time it would take for someone else to understand it.

system/Traits/ConditionalTrait.php
lonnieezell: I'm confused. `static` function? That wouldn't work in a chain, though, which is the purpose.

system/Traits/ConditionalTrait.php
lonnieezell: I'll commit, but I thought we weren't adding parameter or return types to docblocks when we could add them to the method itself now?

system/Traits/ConditionalTrait.php
kenjis: No, the return type.

system/Traits/ConditionalTrait.php
kenjis: This is a special case. Because we cannot declare return type `$this`.

MGatner: I was just providing the suggestion for what @kenjis commented above. We don't get `static` as a return type until PHP 8 so the best we can do is use native `self` with a docblock of `$this`.

system/Traits/ConditionalTrait.php
kenjis: Sorry, it cannot be used in PHP 7.4.
https://php.watch/versions/8.0/static-return-type

user_guide_src/source/database/query_builder.rst
```diff
@@ -1022,6 +1022,42 @@ that it produces a **DELETE** SQL string instead of an **INSERT** SQL string.
...
 For more information view documentation for ``$builder->getCompiledInsert()``.
 
+**********************
+Conditional Statements
+**********************
+
+.. _db-builder-when:
+
+$builder->when()
+------------------
```
kenjis, line 1032:
```suggestion
When
====

$builder->when()
----------------
```

user_guide_src/source/database/query_builder.rst
```diff
@@ -1022,6 +1022,42 @@ that it produces a **DELETE** SQL string instead of an **INSERT** SQL string.
...
+set in the condition will be passed as the second parameter to the callable so it
+can be used in the query.
+
+Sometimes you might want to apply a different statement if the condition evaluates to false.
+This can be accomplished by providing a second closure:
+
+.. literalinclude:: query_builder/106.php
+
+$builder->whenNot()
+-------------------
```
kenjis, line 1054:
```suggestion
WhenNot
=======

$builder->whenNot()
-------------------
```

system/Traits/ConditionalTrait.php
```diff
@@ -0,0 +1,51 @@
...
+    public function whenNot($condition, callable $callback, ?callable $defaultCallback = null): self
+    {
+        if (! $condition) {
+            $callback($this, $condition);
+        } elseif ($defaultCallback) {
+            $defaultCallback($this);
+        }
+
+        return $this;
```
MGatner, line 49:
```suggestion
        return $this->whenNot(! $condition, $callback, $defaultCallback);
```

MGatner (changes requested):
I follow @iRedds suggestion to make this one method with its inversion.

system/Traits/ConditionalTrait.php
kenjis: `$this->when(! ...)`

lonnieezell: Unfortunately it's not that simple. All of the tests passed except the last one, `testWhenNotPassesParemeters`. This changes the value of `$condition`, which is passed as a parameter into the value. In this test's case it changed the `0` into `true` when passed into the `when` command. 

While this example is a trivial example of it the issue, it could be a much bigger issue with different data passed into it and is being used within the callback. 

For example: 

```php
$user = $users->where('id', $this->request->getPost('user_id'))->first();

$activeUsers = $users
    ->where('active', 1)
    ->whenNot($user, function($query, $user) {
        return $query->where('id', '!=', $user->id);
    })
    ->findAll();
```

Ok, this example stinks, also, but I think you see the issue. When passed into `when()`, `$user` becomes a boolean, which cannot be used within the callback.

kenjis: Yes, it is not so simple!
`@param array|bool|float|int|object|resource|string|null $condition`

system/Traits/ConditionalTrait.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+namespace CodeIgniter\Traits;
+
+trait ConditionalTrait
+{
+    /**
+     * Only runs the query when $condition evaluates to true
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
```
paulbalandan, line 19: How about using `@template` types for the `$condition` so that when the trait is actually used, we can annotate the actual type on it and static analyzers can check it.

paulbalandan, line 19: How about using `@template` types for the `$condition` so that when the trait is actually used, we can annotate the actual type on it and static analyzers can check it.

system/Traits/ConditionalTrait.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Only runs the query when $condition evaluates to false
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
+     *
+     * @return $this
+     */
+    public function whenNot($condition, callable $callback, ?callable $defaultCallback = null): self
```
paulbalandan, line 41: `whenNot` as the inverse of `when` seems quite not natural for me. It's a bit off. How about other antonyms of `when`:

- except
- unless
- if not
- lest

system/Traits/ConditionalTrait.php
MGatner: Ah, that makes sense! @iRedds' suggestion makes more sense now where you invert the callables instead of the condition:
```suggestion
        return $this->when($condition, $defaultCallback, $callback);
```

lonnieezell: That doesn't work either if you haven't specified a default callback, which I predict would be fairly often. 

I think we're stuck with 3 extra lines this time around.

system/Traits/ConditionalTrait.php
lonnieezell: The method expects the condition to evaluate to `false`, which makes `except` and `unless` not make semantic sense. `ifNot` as an opposite of `when` feels wrong also. `lest` feels like the condition should evaluate to `true` to me, also. 

`whenNot` does at least match the pattern of other methods like `where`/`whereNot`. I'm open to suggestions but I'm not sure any of these fit, unless we change the whole method to suit the name.

paulbalandan: `except`/`unless` makes sense to me if the statement is read in a specific way. In the context of the DB builder, since I know the usage of `when` then when I read its inverse (`except`/`unless`):
```php
$builder = $this->db->table('users');
$user = model('Users')->find(1);

return $builder->except($user, fn ($query) { /* some code */ })->get();
```
In this example, I would read the code as "Except when `$user` is truthy (or a found user record), let the `$builder` execute the callable then get the DB result."

kenjis: `whenNot` is easy to understand for non native English speakers.
It is logical. The `when` `whenNot` pair is easy to remember, because we already have where/whereNot.

`except` is difficult to imagine as the opposite of `when`.

system/Traits/ConditionalTrait.php
kenjis: I support the current code. It is no problem at all.
It works, and everybody can read and understand.

Adding *null callback* is a bit tricky.

kenjis (approved):

system/Traits/ConditionalTrait.php
MGatner: Could you give an example? I don't see where a template would help because the value is passed through to the callable, which should have its own typed parameters.

system/Traits/ConditionalTrait.php
MGatner: 👍

system/Traits/ConditionalTrait.php
paulbalandan: Actually not annotate, but let static analyzers like PHPStan analyse the condition's type.
This is a contrived example where template type may be useful.

https://phpstan.org/r/52e48ac1-6f8e-4b8d-9c51-f53918f4f64e

paulbalandan: Using the current typing (for the sake of discussion I used `mixed` instead of the long form):
https://phpstan.org/r/c6ed6719-28f3-4eed-acf9-3f50bb19065b

lonnieezell:
@MGatner are we ok to merge this one?

system/Traits/ConditionalTrait.php
MGatner: Ah I see what you mean! Yes I think that would be a good idea. I'm always for more specific types 💪

MGatner:
@lonnieezell It needs to be switched to 4.3. You can try this:
```shell
got rebase --onto develop 4.3 when-trait
```
I think the content is all good, though I do think @paulbalandan 's suggestion is good. Maybe he would add it as an actual GitHub suggested change?

system/Traits/ConditionalTrait.php
paulbalandan, line 19:
```suggestion
     * @template TWhen of mixed
     *
     * @phpstan-param TWhen                        $condition
     * @phpstan-param callable(self, TWhen): mixed $callback
     * @phpstan-param (callable(self): mixed)|null $defaultCallback
```

system/Traits/ConditionalTrait.php
```diff
@@ -0,0 +1,51 @@
...
+            $defaultCallback($this);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Only runs the query when $condition evaluates to false
+     *
+     * @param array|bool|float|int|object|resource|string|null $condition
```
paulbalandan, line 37:
```suggestion
     * @template TWhenNot of mixed
     *
     * @phpstan-param TWhenNot                        $condition
     * @phpstan-param callable(self, TWhenNot): mixed $callback
     * @phpstan-param (callable(self): mixed)|null    $defaultCallback
```

lonnieezell:
@MGatner It's already on 4.3 so we're good there. Just committed @paulbalandan suggestion.

MGatner (approved):
Looks good to me! Last comment is the negative name; I don't have a strong feeling either way.

kenjis:
Can you remove the merge commit 49f1e7096b52254c908c92debab50238872341b3 ?
We don't need merge commits in PR branches.

kenjis:
There is a coding style error.
https://github.com/codeigniter4/CodeIgniter4/actions/runs/3144200239/jobs/5109891442

system/Traits/ConditionalTrait.php
paulbalandan: It seems this is not yet committed.

lonnieezell:
> Can you remove the merge commit https://github.com/codeigniter4/CodeIgniter4/commit/49f1e7096b52254c908c92debab50238872341b3 ?

I think the likelihood of me messing something up trying to do that is pretty high, honestly. Seems we just squash and merge and the history is nice and clean that way, won't it be?

kenjis:
Okay, I rebased and force-pushed.

kenjis (approved):

kenjis (approved):

system/Traits/ConditionalTrait.php
MGatner: He applied it separately:
https://github.com/codeigniter4/CodeIgniter4/commit/b0a1bea2d805d8aec332f9a1e9c29a991264188b

MGatner (approved):
Hurrah!

datamweb (approved):
Finally, good result, thank you all.😍

MGatner:
🥳

PR #6581, opened 2022-09-24
docs: fix past changelogs format

kenjis:
**Description**
- fix past changelogs format

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


MGatner (approved):
I only looked through about half of these but it's clearly an improvement. Thanks!

Issue #6170, opened 2022-06-21
Dev: PHP 8.2 support


kenjis:
https://wiki.php.net/todo/php82

- [x] Add PHP 8.2 GitHub Actions
  - #6172 
- [x] [PHP: rfc:deprecate_dynamic_properties](https://wiki.php.net/rfc/deprecate_dynamic_properties)
  - #6162
     - #6586
  - #6187
  - #6198
  - #6587
  - #6666
  - #6667
  - #6669
  - #6735
  - #6757
- [x] [PHP: rfc:deprecate_partially_supported_callables](https://wiki.php.net/rfc/deprecate_partially_supported_callables)
  - Probably none. At least no error is reported in PHPUnit testing
- [] [PHP: rfc:strtolower-ascii](https://wiki.php.net/rfc/strtolower-ascii)
- [x] [mb_convert_encoding(): Handling HTML entities via mbstring is deprecated](https://php.watch/versions/8.2/mbstring-qprint-base64-uuencode-html-entities-deprecated)
  - #6672 
- [x] [DateTime::getLastErrors() returns false](https://github.com/codeigniter4/CodeIgniter4/pull/6172#issuecomment-1275496541)
  - #6671
- [x] [PHP: rfc:remove_utf8_decode_and_utf8_encode](https://wiki.php.net/rfc/remove_utf8_decode_and_utf8_encode)
   - #6564 
- [x] Dependent packages
  - #6758

References:
- https://php.watch/versions/8.2
- https://stitcher.io/blog/new-in-php-82
- https://www.php.net/manual/en/migration82.incompatible.php
- https://www.php.net/manual/en/migration82.deprecated.php

paulbalandan:
@kenjis can this be closed? or are we still waiting for the last two tickets to clear?

kenjis:
I'm not sure.
Probably we do not use deprecate_partially_supported_callables.
I did not check the impact of strtolower-ascii.

kenjis:
About [PHP: rfc:strtolower-ascii](https://wiki.php.net/rfc/strtolower-ascii).

I've checked `strtolower()` in `develop`.

Many parts of the code look fine.

> PHP 8.0 stopped respecting the locale environment variables. So the locale is always “C” unless the user explicitly calls setlocale().
https://wiki.php.net/rfc/strtolower-ascii

[...]

PR #6587, opened 2022-09-25
test: remove dynamic properties

kenjis:
**Description**
See #6170

- remove dynamic properties

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner (commented):
Would anonymous class extension injection be better than fixtures? These don't seem to have much reuse value.

kenjis:
> Would anonymous class extension injection be better than fixtures? These don't seem to have much reuse value.

Yes! Fixed.

MGatner (approved):
Thanks! Looks great.

paulbalandan (approved):

PR #6590, opened 2022-09-26
docs: add about named arguments in Backward Compatibility Notes

kenjis:
**Description**
[Named arguments](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments) are not covered by backwards compatibility (BC) promise.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


paulbalandan (approved):

MGatner (approved):
Looks good. Maybe this should be a goal for v5: a pass on all argument names so we are happy enough with them that was can commit them to BC promise

PR #6595, opened 2022-09-27
Allow HTTP/3 to work and not be blocked.

lonnieezell:
Currently, CodeIgniter blocks HTTP/3 and doesn't consider it a valid version. Chrome, Edge, Firefox, and Safari all support this, though Safari has support turned off by default. 

While this doesn't add any specific feature support, it does it from being ignored as a valid protocol.

### DIFF
--- a/system/HTTP/Message.php
+++ b/system/HTTP/Message.php
@@ -34,6 +34,7 @@ class Message implements MessageInterface
         '1.0',
         '1.1',
         '2.0',
+        '3.0',
     ];
 
     /**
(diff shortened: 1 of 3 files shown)


kenjis:
Does this really work with HTTP/3 server?
Can anyone test?

paulbalandan:
It seems this was reported as early as last March 2022.
https://forum.codeigniter.com/showthread.php?tid=81584

MGatner (approved):
If the framework won't run on the new protocol then we have some serious boundary bleeding! I'm all for securing the basics but your framework should rely on semantics and not have to adapt to the web protocol.

lonnieezell:
Maybe the better solution is to just remove that check altogether. I can't think of a great need for it. I imagine it was just to ensure someone didn't accidentally set an invalid version on a Response.

kenjis (approved):
But it seems better to remove the version check.

PR #6596, opened 2022-09-27
docs: replace type in `Pager`

ddevsr:
**Description**
See #6310

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Pager/PagerTest.php
+++ b/tests/system/Pager/PagerTest.php
@@ -182,9 +182,6 @@ final class PagerTest extends CIUnitTestCase
 
     public function testStoreWithURIReturnObject()
     {
-        $_GET['page'] = 5;
-        $_GET['id']   = 1;
-
         $this->pager->store('bar', 5, 25, 100, 1);
 
         $uri     = $this->pager->getPageURI(7, 'bar', true);
--- a/tests/system/Pager/PagerTest.php
+++ b/tests/system/Pager/PagerTest.php
@@ -180,7 +180,7 @@ final class PagerTest extends CIUnitTestCase
         );
     }
 
-    public function testStoreWithURIReturnObject()
+    public function testgetPageURIWithURIReturnObject()
     {
         $this->pager->store('bar', 5, 25, 100, 1);
 
--- a/tests/system/Pager/PagerTest.php
+++ b/tests/system/Pager/PagerTest.php
@@ -184,16 +184,9 @@ final class PagerTest extends CIUnitTestCase
     {
         $this->pager->store('bar', 5, 25, 100, 1);
 
-        $uri     = $this->pager->getPageURI(7, 'bar', true);
-        $onlyURI = $this->pager->only(['bar'])->getPageURI(9, 'bar', true);
+        $uri = $this->pager->getPageURI(7, 'bar', true);
 
-        // Total Segment
-        $this->assertSame(1, $uri->getTotalSegments());
-        $this->assertSame(1, $onlyURI->getTotalSegments());
-
-        // Get Segment
-        $this->assertSame('', $uri->getSegment(2));
-        $this->assertSame('9', $onlyURI->getSegment(1));
+        $this->assertInstanceOf(URI::class, $uri);
     }
 
     public function testHasMoreDefaultsToFalse()
--- a/tests/system/Pager/PagerTest.php
+++ b/tests/system/Pager/PagerTest.php
@@ -180,7 +180,7 @@ final class PagerTest extends CIUnitTestCase
         );
     }
 
-    public function testgetPageURIWithURIReturnObject()
+    public function testGetPageURIWithURIReturnObject()
     {
         $this->pager->store('bar', 5, 25, 100, 1);
 
--- a/tests/system/Pager/PagerTest.php
+++ b/tests/system/Pager/PagerTest.php
@@ -18,7 +18,7 @@ use CodeIgniter\HTTP\UserAgent;
 use CodeIgniter\Pager\Exceptions\PagerException;
 use CodeIgniter\Test\CIUnitTestCase;
 use Config\App;
-use Config\Pager;
+use Config\Pager as PagerConfig;
 use Config\Services;
 
 /**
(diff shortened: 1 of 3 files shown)


tests/system/Pager/PagerTest.php
```diff
@@ -182,6 +182,22 @@ public function testStoreWithSegments()
         );
     }
 
+    public function testStoreWithURIReturnObject()
```
kenjis, line 185: What is this test testing?

ddevsr: For covered testing if `$returnObject = true`

```php
$this->pager->store('bar', 5, 25, 100, 1);
$uri = $this->pager->getPageURI(7, 'bar', true);
```

kenjis: Why don't you assert that `$uri` is instance of `URI`?

tests/system/Pager/PagerTest.php
```diff
@@ -182,6 +182,22 @@ public function testStoreWithSegments()
...
+    {
+        $this->pager->store('bar', 5, 25, 100, 1);
+
+        $uri     = $this->pager->getPageURI(7, 'bar', true);
+        $onlyURI = $this->pager->only(['bar'])->getPageURI(9, 'bar', true);
```
kenjis, line 190: The URI does not contain query string for pager.
Why do you call `only()`?

tests/system/Pager/PagerTest.php
```diff
@@ -182,6 +182,22 @@ public function testStoreWithSegments()
...
+
+        // Total Segment
+        $this->assertSame(1, $uri->getTotalSegments());
+        $this->assertSame(1, $onlyURI->getTotalSegments());
```
kenjis, line 194: Why do you check total segments?

tests/system/Pager/PagerTest.php
```diff
@@ -182,6 +182,22 @@ public function testStoreWithSegments()
...
+
+        // Get Segment
+        $this->assertSame('', $uri->getSegment(2));
```
kenjis, line 197: What do you want to check?

tests/system/Pager/PagerTest.php
ddevsr: I call `getTotalSegments` and `getSegment` because part of the function URI class. Its okay if only assert instance of `URI`

kenjis: Please try to write a test that is as clear as possible what you are trying to test.

And if you want to test two aspects, you can add two test methods.

tests/system/Pager/PagerTest.php
```diff
@@ -182,6 +182,15 @@ public function testStoreWithSegments()
         );
     }
 
+    public function testStoreWithURIReturnObject()
```
kenjis, line 185:
```suggestion
    public function testgetPageURIWithURIReturnObject()
```

tests/system/Pager/PagerTest.php
```diff
@@ -182,6 +182,15 @@ public function testStoreWithSegments()
         );
     }
 
+    public function testgetPageURIWithURIReturnObject()
```
kenjis, line 185: Sorry.

```suggestion
    public function testGetPageURIWithURIReturnObject()
```

tests/system/Pager/PagerTest.php
ddevsr: What about this?

```diff
--- a/tests/system/Pager/PagerTest.php
+++ b/tests/system/Pager/PagerTest.php
@@ -184,11 +184,28 @@ final class PagerTest extends CIUnitTestCase

     public function testGetPageURIWithURIReturnObject()
     {
-        $this->pager->store('bar', 5, 25, 100, 1);
+        $_GET['page'] = 3;
+        $_GET['foo']  = 'bar';
+
+        $this->pager->store('default', 3, 25, 100, 1);

-        $uri = $this->pager->getPageURI(7, 'bar', true);
+        $uri = $this->pager->getPageURI(5, 'default', true);

         $this->assertInstanceOf(URI::class, $uri);
+        $this->assertSame(
+            'page=3&foo=bar',
+            $uri->getQuery()
+        );
+
+        $this->pager->store('new', 3, 25, 100, 1);
+
+        $uriOnly = $this->pager->only(['foo'])->getPageURI(7, 'new', true);
+
+        $this->assertInstanceOf(URI::class, $uriOnly);
+        $this->assertSame(
+            'foo=bar',
+            $uriOnly->getQuery()
+        );
     }
```

kenjis: It is better than the first test.
But there are many test aspects:
- return URI object
- query parameters in URI object
- only() method

It is better to write one test for one aspect.

If the test method is `testGetPageURIWithURIReturnObject`, the current test is enough.
If you want to write a test for only() or query parameters, please add another test method.
Or is there any test already?

kenjis:
@ddevsr It seems you did `git merge`, but if you can use `git rebase`, please use it in stead of `git merge`
when you update your PR branch.

kenjis (approved):

tests/system/Pager/PagerTest.php
ddevsr: `getPageURI` only 2 aspect, `with only` or `without only`. And this topic coverage `getPageURI` return `URI` object

```php
    public function testGetPageURIWithURIReturnObject()
    {
        $_GET['page'] = 3;
        $_GET['foo']  = 'bar';

        $this->pager->store('default', 5, 25, 100, 1);

        $uri = $this->pager->getPageURI(5, 'default', true);

        $this->assertInstanceOf(URI::class, $uri);
        $this->assertSame(
            'page=3&foo=bar',
            $uri->getQuery()
        );
    }

    public function testOnlyGetPageURIWithURIReturnObject()
    {
        $_GET['page'] = 3;
        $_GET['foo']  = 'bar';

        $this->pager->store('default', 5, 25, 100, 1);

        $uriOnly = $this->pager->only(['foo'])->getPageURI(5, 'default', true);

        $this->assertInstanceOf(URI::class, $uriOnly);
        $this->assertSame(
            'foo=bar',
            $uriOnly->getQuery()
        );
    }
```

MGatner (approved):
@kenjis has gone through this thoroughly so I just checked out the end result. Very nice QA, thank you!

PR #6600, opened 2022-09-27
Add BaseBuilder::upsert() and BaseBuilder::upsertBatch()

sclubricants:
This PR adds `BaseBuilder::upsert()` and `BaseBuilder::upsertBatch()` methods.

Related: https://github.com/codeigniter4/CodeIgniter4/pull/6407/

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


sclubricants:
@codeigniter4/database-team Have at it!

kenjis:
The code coverage is down. Can you add tests for uncovered code?

Screenshot 2022-09-28 12 05 03

sclubricants:
There is a few lines of code uncovered because its there for a feature not yet fully implemented. 

I never really get the code coverage tests.. they don't make a lot of sense at least on the database. From my experience it seems anytime you change anything in the database these coverage tests complain. I think its because what changes in one driver affects the other coverage tests. Its odd because you look at the mysql test and it shows coverage for say sqlsrv.

Ill try and run some coverage tests locally and make sure we are covered.

kenjis:
Yeah, something wrong with coverall results.
So if the results are wrong, I don't mind the coverage down.

sclubricants:
The coverage was really good. Just a couple of lines missed. I added a test and changed one to cover the missing lines. All the upsert and related methods had > 95% coverage.

I think what happens in a case like this is that when the coverage test for MySql runs, all the code that was added to the other drivers is not visible. So basically you increase the denominator and the coverage drops. One thing that would help would be to exclude the other drivers for each coverage test. This isn't perfect either though because sometimes a driver's code is implemented in `BaseBuilder`.

system/Database/SQLite3/Builder.php
```diff
@@ -78,12 +79,20 @@ protected function _truncate(string $table): string
     protected function _updateBatch(string $table, array $keys, array $values): string
     {
         if (version_compare($this->db->getVersion(), '3.33.0') >= 0) {
-            return parent::_updateBatch($table, $keys, $values);
+            // return parent::_updateBatch($table, $keys, $values);
```
kenjis, line 82: Why just commented out?
If this line is no longer needed, entire `if` is not needed.

kenjis:
The docs have conflicts now. Please resolve.

system/Database/SQLSRV/Builder.php
```diff
@@ -618,4 +618,137 @@ public function get(?int $limit = null, int $offset = 0, bool $reset = true)
...
+        // if this is the first iteration of batch then we need to build skeleton sql
+        if ($sql === '') {
+            $fullTableName = $this->getFullName($table);
+
+            $constraints = $this->QBOptions['constraints'] ?? [];
+
+            $tableIdentity = $this->QBOptions['tableIdentity'] ?? '';
+            $sql           = "SELECT name from syscolumns where id = Object_ID('" . $table . "') and colstat = 1";
+            if (($query = $this->db->query($sql)) === false) {
+                throw new DatabaseException('Failed to get table identity'); // @codeCoverageIgnore
```
kenjis, line 640: Why is `@codeCoverageIgnore` added?

system/Database/BaseBuilder.php
```diff
@@ -1851,6 +1852,113 @@ public function setData($set, ?bool $escape = null, string $alias = '')
...
+            $sql .= implode(
+                ",\n",
+                array_map(
+                    static fn ($key, $value) => $table . '.' . $key . ($value instanceof RawSql ?
+                        ' = ' . $value :
+                        ' = ' . 'VALUES(' . $value . ')'),
+                    array_keys($updateFields),
+                    $updateFields
+                )
+            );
```
kenjis, line 1948: How about this?
```suggestion
            $sql = 'INSERT INTO ' . $table . ' (' . implode(', ', $keys) . ')' . "\n"
                . '{:_table_:}'
                . 'ON DUPLICATE KEY UPDATE' . "\n"
                . implode(
                 ",\n",
                 array_map(
                     static fn ($key, $value) => $table . '.' . $key . ($value instanceof RawSql ?
                         ' = ' . $value :
                         ' = ' . 'VALUES(' . $value . ')'),
                     array_keys($updateFields),
                     $updateFields
                 )
                );
```

kenjis:
The following tests fails on my local.

```console
$ vendor/bin/phpunit tests/system/Database/Live/UpdateTest.php

Runtime:       PHP 8.1.10
Configuration: /Users/kenji/work/codeigniter/official/CodeIgniter4/phpunit.xml

...........EEFEE.                                                 17 / 17 (100%)

Time: 00:00.532, Memory: 16.00 MB

There were 4 errors:

1) CodeIgniter\Database\Live\UpdateTest::testUpdateBatchTwoConstraints
CodeIgniter\Database\Exceptions\DatabaseException: You are trying to use a feature which requires SQLite version 3.33 or higher.

/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/SQLite3/Builder.php:96
/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:1771
/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:2411
/Users/kenji/work/codeigniter/official/CodeIgniter4/tests/system/Database/Live/UpdateTest.php:262

2) CodeIgniter\Database\Live\UpdateTest::testUpdateBatchConstraintsRawSqlAndAlias
CodeIgniter\Database\Exceptions\DatabaseException: You are trying to use a feature which requires SQLite version 3.33 or higher.

/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/SQLite3/Builder.php:96
/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:1771
/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:2411
/Users/kenji/work/codeigniter/official/CodeIgniter4/tests/system/Database/Live/UpdateTest.php:307

3) CodeIgniter\Database\Live\UpdateTest::testUpdateBatchWithoutOnConstraint
CodeIgniter\Database\Exceptions\DatabaseException: You are trying to use a feature which requires SQLite version 3.33 or higher.

/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/SQLite3/Builder.php:96
/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:1771
/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:2411
/Users/kenji/work/codeigniter/official/CodeIgniter4/tests/system/Database/Live/UpdateTest.php:449

4) CodeIgniter\Database\Live\UpdateTest::testRawSqlConstraint
CodeIgniter\Database\Exceptions\DatabaseException: You are trying to use a feature which requires SQLite version 3.33 or higher.

/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/SQLite3/Builder.php:96
/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:1771
/Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:2411
/Users/kenji/work/codeigniter/official/CodeIgniter4/tests/system/Database/Live/UpdateTest.php:481

--

There was 1 failure:

1) CodeIgniter\Database\Live\UpdateTest::testUpdateBatchUpdateFieldsAndAlias
Failed asserting that null is not null.

/Users/kenji/work/codeigniter/official/CodeIgniter4/tests/system/Database/Live/UpdateTest.php:361

ERRORS!
Tests: 17, Assertions: 19, Errors: 4, Failures: 1.
```

system/Database/SQLite3/Builder.php
sclubricants: I had this commented out to test it. It should not be commented out.

system/Database/SQLSRV/Builder.php
sclubricants: because the query shouldn't really ever fail.. only if for some reason the database is not available.

sclubricants:
>The following tests fails on my local.

This failed because I forgot to remove the comment in SQLite Builder to return parent. The github actions runs an old version of SQLite so it didn't cause an error when uploaded here. This should be fixed now.

system/Database/SQLSRV/Builder.php
kenjis: This code is not unreachable because databases could be unavailable and that is what the code is for. 
It is simply not tested by PHPUnit and the tag `@codeCoverageIgnore` should be removed.

Then the code with the tag added after throwing exception is simply not tested either.

I don't think it is good practice to add the tag to make it look like coverage has increased 
just because it is difficult to write test code.

sclubricants: I removed it

michalsn (approved):
I don't see any problems... Really nice feature. Even the user guide part is polished. No complains from me this time 😅 

Thanks, @sclubricants!

kenjis:
Please rebase to fix PHPStan errors.

system/Database/BaseBuilder.php
```diff
@@ -1854,6 +1855,108 @@ public function setData($set, ?bool $escape = null, string $alias = '')
...
+     * @param array|object|null $set
+     *
+     * @return false|int|string[] Number of affected rows or FALSE on failure, SQL array when testMode
+     *
+     * @throws DatabaseException
+     */
+    public function upsert($set = null, ?bool $escape = null)
+    {
+        if ($set === null) {
+            $set = empty($this->binds) ? null : [array_map(static fn ($columnName) => $columnName[0], $this->binds)];
```
kenjis, line 1888: This retrieve only values from the `$this->binds`.
And calling `$this->setData($set, $escape)` below sets the escape values with the parameter `$escape`.

The first escape values set are overridden with the parameter value.

kenjis:
The following test causes an error. Is this misuse?

```diff
--- a/tests/system/Database/Live/UpsertTest.php
+++ b/tests/system/Database/Live/UpsertTest.php
@@ -234,11 +234,11 @@ final class UpsertTest extends CIUnitTestCase
         $userData = [
             'email'   => 'ahmadinejad@world.com',
             'name'    => 'Ahmadinejad',
-            'country' => 'Iran',
+            'country' => 'now()',
         ];
 
         $this->assertSame($expected, $this->db->table('user')
-            ->set($userData)
+            ->set($userData, '', false)
             ->getCompiledUpsert());
     }
 
```
```
CodeIgniter\Database\Exceptions\DatabaseException : setData() has no data.
 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:1805
 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:1900
 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Database/BaseBuilder.php:1869
 /Users/kenji/work/codeigniter/official/CodeIgniter4/tests/system/Database/Live/UpsertTest.php:242
```

sclubricants:
>The following test causes an error. Is this misuse?

Im on mobile at the minute so I'm not able to see everything but thinking about it, it might be a case that it doesn't support this use. Probably a quick copy and paste a few lines of code from upsert() should get this working.

sclubricants:
Ok, had a look at the code. Yes there is a bit of mismatch here. This should work if you use setData() rather than set(). setData() accepts a single row or an array of rows. It converts a single row to an array of one row. set() only allows a single row and does not convert it to an array of arrays. upsert() and getCompiledUpsert() is really just an upsertBatch() of one row. Honestly you really don't need both upsert() and upsertBatch() but I added it to be consistent... though you have exposed an inconsistency. upsert() does not use binds either.. which the set() automatically applies. 

It might be nice to combine set() and setData() methods at some point. It would require a lot more refactoring though. setData() evolved out of setInsertBatch(). We would just need to rewrite the single row methods to use an array of arrays.

sclubricants:
On a second look, I'm not sure everything I said above is correct. It looks like it should work with set(). It extracts the data from binds. Ill have to tests some things tomorrow and see what is going on here.

sclubricants:
I have fixed your issue. It appears that when calling set() without escape binds is not used. I made some adjustments that allow for this.

I also added the ability to use setData() with upsert(). I'd like to eventually (maybe version 5) move all the calls to use setData(). set() might be more useful of a command for the SQL command SET. But by adding setData() this allows calling getCompiledUpsert() on a batch of data. I added a test for this too.

tests/system/Database/Live/UpsertTest.php
```diff
@@ -0,0 +1,613 @@
...
+                'email'   => 'pedro@example.com',
+                'name'    => 'Pedro',
+                'country' => 'El Salvador',
+            ],
+        ];
+
+        $sql = $this->db->table('user')
+            ->setData($userData)
+            ->onConstraint('email')
+            ->getCompiledUpsert();
```
kenjis, line 263: It seems something wrong. The `$sql` value  is like this:
```
string (254) "INSERT INTO `db_user` (0, 1, 2)
VALUES ('\'Iran\'','\'ahmadinejad@example.com\'','\'Ahmadinejad\''), ('\'El Salvador\'','\'pedro@example.com\'','\'Pedro\'')
ON DUPLICATE KEY UPDATE
`db_user`.0 = VALUES(0),
`db_user`.1 = VALUES(1),
`db_user`.2 = VALUES(2)"
```

```
string (244) "INSERT INTO `db_user` (0, 1, 2)
VALUES ('''Iran''','''ahmadinejad@example.com''','''Ahmadinejad'''), ('''El Salvador''','''pedro@example.com''','''Pedro''')
ON CONFLICT(`email`)
DO UPDATE SET
0 = `excluded`.0,
1 = `excluded`.1,
2 = `excluded`.2"
```

kenjis:
`set()` works fine:
```php
        $data = [
            'email'   => 'ahmadinejad@example.com',
            'name'    => 'Ahmadinejad',
            'country' => 'Iran',
        ];
        $sql = $this->db->table('user')->set($data)->getCompiledUpsert();
        echo $sql;
```
```
INSERT INTO `db_user` (`country`, `email`, `name`)
VALUES ('Iran','ahmadinejad@example.com','Ahmadinejad')
ON DUPLICATE KEY UPDATE
`db_user`.`country` = VALUES(`country`),
`db_user`.`email` = VALUES(`email`),
`db_user`.`name` = VALUES(`name`)
```

`setData()` does not work:
```php
        $data = [
            'email'   => 'ahmadinejad@example.com',
            'name'    => 'Ahmadinejad',
            'country' => 'Iran',
        ];
        $sql = $this->db->table('user')->setData($data)->getCompiledUpsert();
        echo $sql;
```
```
INSERT INTO `db_user` (0, 1, 2)
VALUES ('\'Iran\'','\'ahmadinejad@example.com\'','\'Ahmadinejad\'')
ON DUPLICATE KEY UPDATE
`db_user`.0 = VALUES(0),
`db_user`.1 = VALUES(1),
`db_user`.2 = VALUES(2)
```

user_guide_src/source/database/query_builder/114.php
```diff
@@ -0,0 +1,18 @@
...
+$sql = $builder->setData($data)->getCompiledUpsert();
+echo $sql;
+/* MySQLi produces:
+    INSERT INTO `db_user` (`country`, `email`, `name`)
+    VALUES ('Iran','ahmadinejad@example.com','Ahmadinejad')
+    ON DUPLICATE KEY UPDATE
+    `country` = VALUES(`country`),
+    `email` = VALUES(`email`),
+    `name` = VALUES(`name`)
+*/
```
kenjis, line 18:
```php
        $data = [
            'email'   => 'ahmadinejad@example.com',
            'name'    => 'Ahmadinejad',
            'country' => 'Iran',
        ];
        $sql = $this->db->table('user')->setData($data)->getCompiledUpsert();
        echo $sql;
```
```
INSERT INTO `db_user` (0, 1, 2)
VALUES ('\'Iran\'','\'ahmadinejad@example.com\'','\'Ahmadinejad\'')
ON DUPLICATE KEY UPDATE
`db_user`.0 = VALUES(0),
`db_user`.1 = VALUES(1),
`db_user`.2 = VALUES(2)
```

kenjis:
```php
        $this->db->table('user')->truncate();
        $builder = $this->db->table('user');

        $builder->set('email', 'ahmadinejad@example.com');
        $builder->set('name', 'Ahmadinejad');
        $builder->set('country', 'now()', false);
        $builder->upsert();

        echo $this->db->getLastQuery();
```
```
INSERT INTO `db_user` (`email`, `name`)
VALUES ('ahmadinejad@example.com','Ahmadinejad')
ON DUPLICATE KEY UPDATE
`db_user`.`email` = VALUES(`email`),
`db_user`.`name` = VALUES(`name`)
```
The `country` is gone.

tests/system/Database/Live/UpsertTest.php
sclubricants: Fixed this. In my last fix I had not had the proper conditions evaluate before running setData().

These issues stem from allowing the use of set() and setData() to both be used with upsert(). For some reason set() uses binds except when not escaping. So if QBSet has been created with set() I have to get the data and set it with setData(). This way it is an array of arrays/objects. The error was caused because setData() removes the associated array and separates the columns to QBKeys.

user_guide_src/source/database/query_builder/114.php
sclubricants: fixed

sclubricants:
>The country is gone.

This helped me figure out what was going on. set() without escape saves data to QBSet. set() with escape saves data to binds. So I changed the code to merge the two sources. setData() doesn't allow using escape with one item and then not with the next. So I merged the two from set() and used RawSql for the items to remain unescaped and then call setData().

michalsn:
Can we add the examples that kenjis provided to our test cases? I think that would be very helpful.

sclubricants:
>Can we add the examples that kenjis provided to our test cases? I think that would be very helpful.

I previously adjusted a test to cover one scenario and also added a test for the use of multiple set() calls.

tests/system/Database/Live/UpsertTest.php
```diff
@@ -0,0 +1,647 @@
...
+
+        $pos = strpos($sql, 'ahmadinejad@example.com');
+        $this->assertNotFalse($pos);
```
kenjis, line 266: Please use `assertStringContainsString()` instead of `strpos()`.
```suggestion
        $this->assertStringContainsString('ahmadinejad@example.com', $sql);
```

tests/system/Database/Live/UpsertTest.php
```diff
@@ -0,0 +1,647 @@
...
+
+        $pos = strpos($sql, 'pedro@example.com');
+        $this->assertNotFalse($pos);
```
kenjis, line 269: Use assertStringContainsString().

tests/system/Database/Live/UpsertTest.php
```diff
@@ -0,0 +1,647 @@
...
+            . ' (' . $this->db->protectIdentifiers('country')
+            . ', ' . $this->db->protectIdentifiers('email')
+            . ', ' . $this->db->protectIdentifiers('name') . ')';
+
+        if ($this->db->DBDriver === 'SQLSRV' || $this->db->DBDriver === 'OCI8') {
+            $insertString = 'INSERT ("country", "email", "name")';
+        }
+
+        $pos = strpos($sql, $insertString);
+        $this->assertNotFalse($pos);
```
kenjis, line 282: Use assertStringContainsString().

kenjis:
> set() without escape saves data to QBSet. set() with escape saves data to binds. So I changed the code to merge the two sources. setData() doesn't allow using escape with one item and then not with the next. So I merged the two from set() and used RawSql for the items to remain unescaped and then call setData().

I am not a fan of this process of retrieving and changing data once it has been saved, 
but it seems to work.

sclubricants:
>I am not a fan of this process of retrieving and changing data once it has been set, but it seems to work.

Id like to maybe refactor the set() method sometime and the methods that use it and make the data structure the same as with setData(). Then maybe rework binds a bit perhaps having it as an optional setting. It would just be nice to have all setting of data funnel though the same method. But this is for another day.

kenjis (approved):

kenjis:
Thank you for this big contribution!

kenjis:
After I've merged, I created two issues.

kenjis:
`UpsertTest::testUpsertWithMultipleSet` failed.
https://github.com/codeigniter4/CodeIgniter4/pull/6690#issuecomment-1279611246

MGatner:
🥳🕺👍

PR #6601, opened 2022-09-28
Allow Cells to be auto-located within */Cells directories

lonnieezell:
Currently, when calling a View Cell you must enter the fully qualified class name. This change uses the ~~`FileLocator`~~ `Factories` class to search for the class within `*/Cells` directories, allowing a simpler naming in views and a more standardized location for them. 

So now we could do: 

```php
<?= view_cell('Blog::recent', 'limit=5') ?>
```

instead of 

```php
<?= view_cell('App\Modules\Blog\Cells\Blog::recent', 'limit=5') ?>
```

and it could be found in `app/Cells`, `App/Modules/Blog/Cells`, or any other namespace's `Cells` folder.

### DIFF
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -154,6 +154,7 @@ Others
 - Added ``$routes->useSupportedLocalesOnly(true)`` so that the Router returns 404 Not Found if the locale in the URL is not supported in ``Config\App::$supportedLocales``. See :ref:`Localization <localization-in-routes>`
 - Now you can specify Composer packages to auto-discover manually. See :ref:`Code Modules <modules-specify-composer-packages>`.
 - Added new ``$routes->view()`` method to return a the view directly. See :ref:`View Routes <view-routes>`.
+- View Cells are now first-class citizens and can located in the ``app/Cells`` directory. See :ref:`View Cells <view-cells>`.
 
 Changes
 *******
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -154,7 +154,7 @@ Others
 - Added ``$routes->useSupportedLocalesOnly(true)`` so that the Router returns 404 Not Found if the locale in the URL is not supported in ``Config\App::$supportedLocales``. See :ref:`Localization <localization-in-routes>`
 - Now you can specify Composer packages to auto-discover manually. See :ref:`Code Modules <modules-specify-composer-packages>`.
 - Added new ``$routes->view()`` method to return a the view directly. See :ref:`View Routes <view-routes>`.
-- View Cells are now first-class citizens and can located in the ``app/Cells`` directory. See :ref:`View Cells <view-cells>`.
+- View Cells are now first-class citizens and can located in the **app/Cells** directory. See :ref:`View Cells <view-cells-app-cells>`.
 
 Changes
 *******
--- a/user_guide_src/source/outgoing/view_cells.rst
+++ b/user_guide_src/source/outgoing/view_cells.rst
@@ -2,22 +2,19 @@
 View Cells
 ##########
 
-View Cells allow you to insert HTML that is generated outside of your controller. It simply calls the specified
-class and method, which must return a string of valid HTML. This method could be in any callable method, found in any class
-that the autoloader can locate. The only restriction is that the class can not have any constructor parameters.
-This is intended to be used within views, and is a great aid to modularizing your code.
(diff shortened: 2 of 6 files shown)


MGatner (commented):
An alternative which would offer a slightly broader feature (maybe desirable?) would be to use `Factories`; this would allow View Cells to support the "modular unless overwritten by App" features characteristic of Models and Configs.

For example, if Shield published a `LoginCell` then a developer would only need to supply **app/Cells/LoginCell.php** that extended it in order to use the replacement. 

Just a thought; I like the improved locator skills generally.

lonnieezell:
I had considered using Factories for this, but was thinking more about the caching uses and thinking it wouldn't be worth it for this. So I could go either way on this. I'm not sure the memory usage would be worth it here. Although the "modular unless overwritten by App" aspect makes it tempting. 

I'll go ahead and try this with factories.

MGatner:
Looks good with Factories! I'm fine with either way, was just an idea. 

Picked up a User Guide syntax warning:

> Warning: undefined label: view-cells

user_guide_src/source/outgoing/view_cells.rst
```diff
@@ -2,16 +2,19 @@
...
-In this example, the class ``App\Libraries\Blog`` is loaded, and the method ``recentPosts()`` is run. The method
-must return the generated HTML as a string. The method can be either a static method or not. Either way works.
+In this example, the class ``App\Libraries\Blog`` is loaded, and the method ``recentPosts()`` is run. The method must return the generated HTML as a string. The method can be either a static method or not. Either way works.
+
+If the Cell is a class, you can locate it in the ``app/Cells`` directory and it can be discovered automatically, allowing the use of just the class name and method.
+
+::
+        <?= view_cell('Blog::recentPosts') ?>
+
+This respects the ``Config\Factories->preferApp`` setting, so if you have a class in both ``app/Cells`` and ``MyNamespace/Cells``, the one in ``app/Cells`` will be used when ``preferApp`` is ``true``.
```
kenjis, line 17:
```suggestion
This respects the Factories ``preferApp`` option, so if you have a class in both ``app/Cells`` and ``MyNamespace/Cells``, the one in ``app/Cells`` will be used when ``preferApp`` is ``true``.
```

`preferApp` is not a property.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -154,6 +154,7 @@ Others
 - Added ``$routes->useSupportedLocalesOnly(true)`` so that the Router returns 404 Not Found if the locale in the URL is not supported in ``Config\App::$supportedLocales``. See :ref:`Localization <localization-in-routes>`
 - Now you can specify Composer packages to auto-discover manually. See :ref:`Code Modules <modules-specify-composer-packages>`.
 - Added new ``$routes->view()`` method to return a the view directly. See :ref:`View Routes <view-routes>`.
+- View Cells are now first-class citizens and can located in the ``app/Cells`` directory. See :ref:`View Cells <view-cells>`.
```
kenjis, line 157:
```suggestion
- View Cells are now first-class citizens and can located in the **app/Cells** directory. See :ref:`View Cells <view-cells-app-cells>`.
```

user_guide_src/source/outgoing/view_cells.rst
```diff
@@ -2,16 +2,19 @@
...
 
     <?= view_cell('\App\Libraries\Blog::recentPosts') ?>
 
-In this example, the class ``App\Libraries\Blog`` is loaded, and the method ``recentPosts()`` is run. The method
-must return the generated HTML as a string. The method can be either a static method or not. Either way works.
+In this example, the class ``App\Libraries\Blog`` is loaded, and the method ``recentPosts()`` is run. The method must return the generated HTML as a string. The method can be either a static method or not. Either way works.
+
+If the Cell is a class, you can locate it in the ``app/Cells`` directory and it can be discovered automatically, allowing the use of just the class name and method.
+
+::
```
kenjis, line 14:
```suggestion
.. _view-cells-app-cells:

app/Cells
---------

Since v4.3.0, if the Cell is a class, you can locate it in the **app/Cells** directory and it can be discovered automatically, allowing the use of just the class name and method::
```

user_guide_src/source/outgoing/view_cells.rst
kenjis, line 17:
```suggestion
This respects the :ref:`Factories preferApp option <factories-options>`, so if you have a class in both **app/Cells** and **MyNamespace/Cells**, the one in **app/Cells** will be used when ``preferApp`` is ``true``.
```

`preferApp` is not a property.

kenjis:
I suggested the fixes for the docs.
The implementation seems good!

lonnieezell:
@kenjis thanks for the docs updates. I've merged those. So I think we've covered everything.

kenjis (approved):

MGatner (approved):
🥳

PR #6604, opened 2022-09-29
fix: safe_mailto() does not work with CSP

### LINKED ISSUE #6602: Bug: url_helper / safe_mailto() not CSP useable (needs nonce)
### PHP Version

7.3, 7.4, 8.0, 8.1

### CodeIgniter4 Version

4.2.6

Git

macOS, Windows, Linux

### Which server did you use?

fpm-fcgi

### Database

_No response_

### What happened?

Browser refused to load/run safe_mailto() function from url_helper, no mail address is shown. The mail address is javascript encoded, but the inline script does not add a nonce when app.CSPEnabled = true

### Steps to Reproduce

enable CSP (e.g. `app.CSPEnabled = true`)

insert `<?= safe_mailto('john.doe@inter.net','John Doe'); ?>` into your view, check output for missing address and debug console for csp errors.

### Expected Output

Visible mail link which loads mail client, no CSP error in debug console.

### Anything else?

I think this issue may concern all helpers/functions that insert inline javascript. I actually did not search for them. The workaround is pretty easy. Just add {csp-script-nonce} to script tag. That's the way I fixed it, but breaks updates via git, since my chances get overwritten.

kenjis:
**Description**
Fixes  #6602

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -369,7 +369,9 @@ if (! function_exists('safe_mailto')) {
         $x = array_reverse($x);
 
         // improve obfuscation by eliminating newlines & whitespace
-        $output = '<script type="text/javascript">'
+        $cspNonce = csp_script_nonce();
+        $cspNonce = $cspNonce ? ' ' . $cspNonce : $cspNonce;
+        $output   = '<script type="text/javascript"' . $cspNonce . '>'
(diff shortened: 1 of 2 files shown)


MGatner (approved):
👍

PR #6589, opened 2022-09-26
fix: `required_without` rule logic in `Validation` class.

ping-yee:
**Description**
Fixed the logic to support #5922 and #5942.

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [x] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #6607, opened 2022-09-29
docs: replace types `mixed` at all classes in `Validation`.

ping-yee:
**Description**
See #6310 
I don't modify the `mixed` at `reuquired_without[]` comment out in `Validation\StrictRules\Rules.php` because I already changed it at PR #6589 .

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/StrictRules/CreditCardRules.php
+++ b/system/Validation/StrictRules/CreditCardRules.php
@@ -41,7 +41,7 @@ class CreditCardRules
      *      'cc_num' => 'valid_cc_number[visa]'
      *  ];
      *
-     * @param mixed $ccNumber
+     * @param string|null $ccNumber
      */
     public function valid_cc_number($ccNumber, string $type): bool
     {
--- a/system/Validation/StrictRules/FormatRules.php
+++ b/system/Validation/StrictRules/FormatRules.php
@@ -30,7 +30,7 @@ class FormatRules
     /**
      * Alpha
      *
-     * @param mixed $str
+     * @param string|null $str
      */
     public function alpha($str = null): bool
     {
@@ -44,7 +44,7 @@ class FormatRules
     /**
      * Alpha with spaces.
      *
-     * @param mixed $value Value.
+     * @param string|null $value Value.
      *
      * @return bool True if alpha with spaces, else false.
      */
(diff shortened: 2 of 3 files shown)


kenjis:
Probably the input value type is `array|bool|float|int|object|string|null` for all methods.
Because an attacker can send any value that can be used in JSON.

ping-yee:
Honestly, I am not sure about the different between normal rule and strict rule.

Is it just different between having strict mode and not having strict mode?
If it is, there is the fixed type `?string $str` paramter of these function at not having strict mode.

Like: 

> In not having strict mode.

https://github.com/codeigniter4/CodeIgniter4/blob/dbb34262d3d7c98922a5be555b9a141b7967e18a/system/Validation/Rules.php#L27

https://github.com/codeigniter4/CodeIgniter4/blob/dbb34262d3d7c98922a5be555b9a141b7967e18a/system/Validation/Rules.php#L39

> In having strict mode.

https://github.com/codeigniter4/CodeIgniter4/blob/dbb34262d3d7c98922a5be555b9a141b7967e18a/system/Validation/StrictRules/Rules.php#L37

https://github.com/codeigniter4/CodeIgniter4/blob/dbb34262d3d7c98922a5be555b9a141b7967e18a/system/Validation/StrictRules/Rules.php#L51

If all of the values need to be changed to `array|bool|float|int|object|string|null`, probably all of paramters need to be changed from `?string $str` to `$str` at not having strict mode file.

kenjis:
See https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html#traditional-and-strict-rules

Traditional Rules are legacy. These are not designed to validate non-string values.

Strictly speaking, probably traditional rule parameter types should be removed, too.
But it is a breaking changes.

MGatner (changes requested):
I'm afraid it will need to be more nuanced and painstaking than this. For example, the first line of `int`:
```php
if (is_int($str)) {
```
... clearly means to support `int` input. 

I think it is a problem with our Validation class in a larger scale, that it is trying to be both Form Validation and more General/Persistence Validation. But until a rework is agreed upon and executed we will need to try to hold these together.

PR #6610, opened 2022-09-30
docs: add `.. versionadded:: 4.3.0`, etc.

kenjis:
**Description**
So that users of older versions will not be confused when they see the latest docs.

- add that new feature can be used from v4.3.0
- add links to detail pages
- small fixes

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


lonnieezell (approved):
Nice addition. Thanks!

PR #6611, opened 2022-09-30
fix: script_tag() does not work with CSP

kenjis:
**Description**
Related #6604
- add nonce when CSP is enabled

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner (approved):
🤦‍♂️

datamweb (approved):
Thanks🥀

PR #6612, opened 2022-09-30
test: fix failed MigrationRunnerTest

kenjis:
**Description**
I don't know why GA check does not fail, but on my local environment the following test always fails.

```console
There was 1 failure:

1) CodeIgniter\Database\Migrations\MigrationRunnerTest::testLoadsDefaultDatabaseWhenNoneSpecified Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'.../CodeIgniter4/writable/:memory:'
+':memory:'

.../CodeIgniter4/tests/system/Database/Migrations/MigrationRunnerTest.php:84
```

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Database/Migrations/MigrationRunnerTest.php
+++ b/tests/system/Database/Migrations/MigrationRunnerTest.php
@@ -79,8 +79,13 @@ final class MigrationRunnerTest extends CIUnitTestCase
         $db = $this->getPrivateProperty($runner, 'db');
 
         $this->assertInstanceOf(BaseConnection::class, $db);
+        $database = (
+            $dbConfig->tests['DBDriver'] === 'SQLite3'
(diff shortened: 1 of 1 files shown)


MGatner (approved):
Looks like this didn't affect GA still 🤷‍♂️

kenjis:
I found the reason. Testing on GA uses a database file. Why not use in-memory database?
https://github.com/codeigniter4/CodeIgniter4/blob/4a6fecb97bbfd4e4389075c4ce805e3ffd33f584/tests/_support/Config/Registrar.php#L70

MGatner:
Not sure why we aren't using in-memory! It might speed things up a bit, though SQLites3 tests will never be our bottleneck.

PR #6613, opened 2022-10-01
docs: add Warning to `$routes->map()`

kenjis:
**Description**
- The `map()` method is not recommended as well as `add()` because it calls `add()` internally.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


datamweb (approved):

MGatner (approved):
Maybe `map()` needs an additional parameter for `$method`?

kenjis:
HTTP method and optional route name seem to be needed.

PR #6614, opened 2022-10-01
fix: default values for Session Redis Handler

### LINKED ISSUE #6512: Bug: Redis::connect(): Passing null to parameter #3 ($timeout) of type float is deprecated
### PHP Version

8.1

### CodeIgniter4 Version

4.2.6

macOS

### Which server did you use?

fpm-fcgi

### Database

_No response_

### What happened?

将session 设置redis

php7.4正常

php8.1报错：Redis::connect(): Passing null to parameter #3 ($timeout) of type float is deprecated

image

### Steps to Reproduce

将session 设置redis

php8.1 下

### Expected Output

将session 设置redis

php8.1 下

### Anything else?

_No response_

kenjis:
**Description**
Fixes  #6512

Make the same as Cache Redis config.
https://github.com/codeigniter4/CodeIgniter4/blob/ed0cdd00eaf20f7d6efb373af991111ead4fece8/app/Config/Cache.php#L155-L161

Ref, https://github.com/phpredis/phpredis#php-session-handler

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Session/Handlers/RedisHandler.php
+++ b/system/Session/Handlers/RedisHandler.php
@@ -58,6 +58,8 @@ class RedisHandler extends BaseHandler
     protected $sessionExpiration = 7200;
 
     /**
+     * @param string $ipAddress User's IP address
+     *
      * @throws SessionException
      */
     public function __construct(AppConfig $config, string $ipAddress)
--- a/system/Session/Handlers/RedisHandler.php
+++ b/system/Session/Handlers/RedisHandler.php
@@ -85,7 +85,7 @@ class RedisHandler extends BaseHandler
                 'port'     => empty($matches[2]) ? self::REDIS_DEFAULT_PORT : $matches[2],
                 'password' => preg_match('#auth=([^\s&]+)#', $matches[3], $match) ? $match[1] : null,
                 'database' => preg_match('#database=(\d+)#', $matches[3], $match) ? (int) $match[1] : 0,
-                'timeout'  => preg_match('#timeout=(\d+\.\d+)#', $matches[3], $match) ? (float) $match[1] : 0,
+                'timeout'  => preg_match('#timeout=(\d+\.\d+|\d+)#', $matches[3], $match) ? (float) $match[1] : 0,
             ];
 
             preg_match('#prefix=([^\s&]+)#', $matches[3], $match) && $this->keyPrefix = $match[1];
--- a/system/Session/Handlers/RedisHandler.php
+++ b/system/Session/Handlers/RedisHandler.php
@@ -266,14 +266,15 @@ class RedisHandler extends BaseHandler
      */
     protected function lockSession(string $sessionID): bool
     {
+        $lockKey = $this->keyPrefix . $sessionID . ':lock';
+
         // PHP 7 reuses the SessionHandler object on regeneration,
         // so we need to check here if the lock key is for the
         // correct session ID.
-        if ($this->lockKey === $this->keyPrefix . $sessionID . ':lock') {
+        if ($this->lockKey === $lockKey) {
             return $this->redis->expire($this->lockKey, 300);
         }
 
-        $lockKey = $this->keyPrefix . $sessionID . ':lock';
         $attempt = 0;
 
         do {
--- a/system/Session/Handlers/RedisHandler.php
+++ b/system/Session/Handlers/RedisHandler.php
@@ -23,6 +23,11 @@ use ReturnTypeWillChange;
 class RedisHandler extends BaseHandler
 {
     /**
+     * @var int
+     */
+    private const REDIS_DEFAULT_PORT = 6379;
+
+    /**
      * phpRedis instance
      *
      * @var Redis|null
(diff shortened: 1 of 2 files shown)


system/Session/Handlers/RedisHandler.php
```diff
@@ -75,10 +77,10 @@ public function __construct(AppConfig $config, string $ipAddress)
 
             $this->savePath = [
                 'host'     => $matches[1],
-                'port'     => empty($matches[2]) ? null : $matches[2],
+                'port'     => empty($matches[2]) ? 6379 : $matches[2],
```
MGatner, line 80: I recommend a class constant instead of the number.

MGatner (commented):
Really great test case! One change to consider.

system/Session/Handlers/RedisHandler.php
kenjis: I just set the default port number for Redis, but I think again,
It seems odd that `tcp://127.0.0.1` means `tcp://127.0.0.1:6379`?
Is it better to raise an error?

MGatner: I don't see a problem with that. It looks like it is a common setup for local Redis dev: https://duckduckgo.com/?q=+tcp%3A%2F%2F127.0.0.1%3A6379&t=iphone&ia=qa

kenjis: `tcp://127.0.0.1:6379` is no problem.
But does `tcp://127.0.0.1` equals `tcp://127.0.0.1:6379`?
I doubt it.

MGatner: That, my friend, is beyond me. It looks like the port is pre-specified in the Config file (https://raw.githubusercontent.com/redis/redis/7.0/redis.conf) so my hunch is "yes", but without testing I don't know.

system/Session/Handlers/RedisHandler.php
```diff
@@ -75,10 +77,10 @@ public function __construct(AppConfig $config, string $ipAddress)
...
                 'password' => preg_match('#auth=([^\s&]+)#', $matches[3], $match) ? $match[1] : null,
-                'database' => preg_match('#database=(\d+)#', $matches[3], $match) ? (int) $match[1] : null,
-                'timeout'  => preg_match('#timeout=(\d+\.\d+)#', $matches[3], $match) ? (float) $match[1] : null,
+                'database' => preg_match('#database=(\d+)#', $matches[3], $match) ? (int) $match[1] : 0,
+                'timeout'  => preg_match('#timeout=(\d+\.\d+)#', $matches[3], $match) ? (float) $match[1] : 0,
```
paulbalandan, line 83: Should this be `0.0`?

paulbalandan: I think the regex should also be updated to check for int timeouts.

tests/system/Session/Handlers/Database/RedisHandlerTest.php
```diff
@@ -0,0 +1,114 @@
...
+{
+    private string $sessionName     = 'ci_session';
+    private string $sessionSavePath = 'tcp://127.0.0.1:6379';
+    private string $userIpAddress   = '127.0.0.1';
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        if (! class_exists(Redis::class)) {
```
paulbalandan, line 32: How about moving this to setUpBeforeClass? Or, using a class docblock instead, like `@requires extension redis`?

system/Session/Handlers/RedisHandler.php
kenjis: I checked the test in this PR with `$sessionSavePath = 'tcp://127.0.0.1'`on PHP 8.0,
and the tests passed.
So to keep compatibility we need to accept `tcp://127.0.0.1` as `tcp://127.0.0.1:6379` in this case.

kenjis: Refactored to use class constant.

system/Session/Handlers/RedisHandler.php
kenjis: Done.

tests/system/Session/Handlers/Database/RedisHandlerTest.php
kenjis: Oh, thanks. It is better.
Done.

MGatner (approved):
👌

system/Session/Handlers/RedisHandler.php
```diff
@@ -22,6 +22,8 @@
  */
 class RedisHandler extends BaseHandler
 {
+    private const REDIS_DEFAULT_PORT = 6379;
```
mostafakhudair, line 25:
```suggestion
    private const DEFAULT_PORT = 6379;
```

What about `RedisHandler::DEFAULT_PORT`?

kenjis: Yeah, `REDIS_` is not needed.
Done.

datamweb (approved):

MGatner (approved):

PR #6615, opened 2022-10-01
Decamelize function

lonnieezell:
New inflector helper function, `decamelize`. 

Converts strings form camelCase or PascalCase into snake_case. 

```php
echo decamelize('MyDogSpot');    // echoes my_dog_spot
```

### DIFF
--- a/tests/system/Helpers/InflectorHelperTest.php
+++ b/tests/system/Helpers/InflectorHelperTest.php
@@ -309,6 +309,12 @@ final class InflectorHelperTest extends CIUnitTestCase
             'aBaBaB'          => 'a_ba_ba_b',
             'BaBaBa'          => 'ba_ba_ba',
             'libC'            => 'lib_c',
+            'a'               => 'a',
+            'A'               => 'a',
+            'aB'              => 'a_b',
+            'AB'              => 'ab',
+            'ab'              => 'ab',
+            'Ab'              => 'ab',
         ];
 
         foreach ($strings as $camelized => $expects) {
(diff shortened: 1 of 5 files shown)


kenjis (approved):

datamweb (approved):
LGTM!

MGatner (changes requested):
Nice addition! Makes me wonder what you are working on 🤔

Please add tests for smaller strings: a, A, Ab, aB, AB, ab.

lonnieezell:
Additional test cases added.

kenjis (approved):

tests/system/Helpers/InflectorHelperTest.php
```diff
@@ -285,4 +285,41 @@ public function testOrdinalize()
...
+            'hello_World'     => 'hello_world',
+            'hello-world'     => 'hello-world',
+            'myHTMLFiLe'      => 'my_html_fi_le',
+            'aBaBaB'          => 'a_ba_ba_b',
+            'BaBaBa'          => 'ba_ba_ba',
+            'libC'            => 'lib_c',
+            'a'               => 'a',
+            'A'               => 'a',
+            'aB'              => 'a_b',
+            'AB'              => 'ab',
```
MGatner, line 315: This variance bothers me a little bit, but ultimately I don't know that two characters are enough to be meaningful anyways, so if you're okay with these then I won't make a fuss.

MGatner (approved):

tests/system/Helpers/InflectorHelperTest.php
lonnieezell: I'm not concerned with this, personally. It follows the same logic as for longer strings, and I would expect the consistency, not special treatment for shorter strings. I'll merge now and we can always revisit if we need to later.

PR #6625, opened 2022-10-04
Fix DotEnv class turning `export` to empty string

### LINKED ISSUE #6621: Bug: Unable to connect to the database when name contains word `export`
### PHP Version

8.1

### CodeIgniter4 Version

4.2.1

Windows

### Which server did you use?

cli

### Database

PostgreSQL

### What happened?

Unable to connect to the database when database group name contains word _export_, even is it's only a partial name, ie. _exports_data_ will also fail.

### Steps to Reproduce

Define db group with word `export` in .env file:
```
database.exports_test.hostname = localhost
database.exports_test.database = dev_db
database.exports_test.username = postgres
database.exports_test.password = postgres
database.exports_test.DBDriver = Postgre
database.exports_test.DBDebug = FALSE
database.exports_test.schema = exports_test
database.exports_test.port = 5432
```

define this group in app/Config/Database.php:
```
public $exports_test= [
    'hostname' => '',
    'schema' => '',
    'username' => '',
    'password' => '',
    'database' => '',
    'DBDriver' => '',
    'DBPrefix' => '',
    'pConnect' => false,
    'cacheOn'  => false,
    'cacheDir' => '',
    'charset'  => 'utf8',
    'swapPre'  => '',
    'encrypt'  => false,
    'compress' => false,
    'strictOn' => false,
    'failover' => [],
];
```

try to run a query against the db. You will receive `You have not selected a database type to connect to` error message

### Expected Output

Query results

### Anything else?

_No response_

paulbalandan:
**Description**
Fixes #6621 

I think the bug is on the incorrect sanitization of exported env variables with the syntax `export varname=value`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/DotEnv.php
+++ b/system/Config/DotEnv.php
@@ -116,7 +116,8 @@ class DotEnv
         $value = trim($value);
 
         // Sanitize the name
-        $name = str_replace(['export', '\'', '"'], '', $name);
+        $name = preg_replace('/^export[ \t]++(\S+)/', '$1', $name);
+        $name = str_replace(['\'', '"'], '', $name);
 
         // Sanitize the value
         $value = $this->sanitizeValue($value);
--- a/tests/system/Config/fixtures/.env
+++ b/tests/system/Config/fixtures/.env
@@ -4,6 +4,11 @@ SPACED="with spaces"
 
 NULL=
 
+char.expo.foo="exported foo"
+character.export.var=variable
+export char.var=character
+export char.exports=imports
+
 SimpleConfig.onedeep=baz
 SimpleConfig.default.name=ci4
 # Use underscore as separator
--- a/tests/system/Config/fixtures/.env
+++ b/tests/system/Config/fixtures/.env
@@ -8,6 +8,7 @@ char.expo.foo="exported foo"
 character.export.var=variable
 export char.var=character
 export char.exports=imports
+fruit.export = "banana"
 
 SimpleConfig.onedeep=baz
 SimpleConfig.default.name=ci4
--- a/tests/system/Config/DotEnvTest.php
+++ b/tests/system/Config/DotEnvTest.php
@@ -55,14 +55,29 @@ final class DotEnvTest extends CIUnitTestCase
         $this->assertFalse($dotenv->load());
     }
 
-    public function testLoadsVars()
+    /**
+     * @dataProvider provideLoadVars
+     */
+    public function testLoadsVars(string $expected, string $varname)
     {
         $dotenv = new DotEnv($this->fixturesFolder);
         $dotenv->load();
-        $this->assertSame('bar', getenv('FOO'));
-        $this->assertSame('baz', getenv('BAR'));
-        $this->assertSame('with spaces', getenv('SPACED'));
-        $this->assertSame('', getenv('NULL'));
+
+        $this->assertSame($expected, getenv($varname));
+    }
+
+    public function provideLoadVars(): iterable
+    {
+        yield from [
+            ['bar', 'FOO'],
+            ['baz', 'BAR'],
+            ['with spaces', 'SPACED'],
+            ['', 'NULL'],
+            ['exported foo', 'char.expo.foo'],
+            ['variable', 'character.export.var'],
+            ['character', 'char.var'],
+            ['imports', 'char.exports'],
+        ];
     }
 
     /**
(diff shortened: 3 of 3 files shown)


system/Config/DotEnv.php
```diff
@@ -116,7 +116,8 @@ public function normaliseVariable(string $name, string $value = ''): array
         $value = trim($value);
 
         // Sanitize the name
-        $name = str_replace(['export', '\'', '"'], '', $name);
```
kenjis, line 119: What is the purpose of deleting these strings?

kenjis:
Is `export varname=value` valid syntax for `.env`?
It seems no such explanation in the user guide.

system/Config/DotEnv.php
paulbalandan: I'm not entirely sure, but I think this was meant to make `'foo'` or `"foo"` to just `foo`

paulbalandan:
> Is `export varname=value` valid syntax for `.env`? It seems no such explanation in the user guide.

I think that was the intention.

paulbalandan:
Okay. Since there was mention that the code comes mostly from `vlucas/phpdotenv`, I checked the repo and found this code in v2.0.0
https://github.com/vlucas/phpdotenv/blob/058b9b03b945b6c3eced241935c45822fd2bf4e8/src/Loader.php#L249-L254

It seems our DotEnv class is based on an old code.

tests/system/Config/fixtures/.env
```diff
@@ -4,6 +4,11 @@ SPACED="with spaces"
 
 NULL=
 
+char.expo.foo="exported foo"
+character.export.var=variable
+export char.var=character
+export char.exports=imports
+
```
MGatner, line 11: Can you add a test for this spacing?
```suggestion
fruit.export = "banana"
```

MGatner (changes requested):
Looks good, an edge case I would like to see covered.

paulbalandan:
> Just saw your comment - should we use an actual package for this, either via Composer or unlined like we do with Kint? Seems like a high-risk security area to be based off old third-party code.

Well I'm in favor. Less maintenance burden on us.

kenjis:
What security concern is in DotEnv?
It seems a few methods were borrowed from vlucas/phpdotenv when we see the code comment.

Config is a part that has slowed down considerably with CI4, and I don't want to see it get any heavier.

kenjis:
It seems vlucas/phpdotenv support `export varname=value`:
https://github.com/vlucas/phpdotenv/blob/master/tests/fixtures/env/exported.env

But for what?

MGatner (approved):
I'm good with this fix for now. I do agree that Config has gotten rather bloated, but I think DotEnv is distinct enough (and necessary) that I would be in favor of some research at least about outsourcing it.

kenjis:
This is not related to this PR, but I've found this PHP extention:
https://github.com/sjinks/php-ext-dotenv
> The parser is compatible with that of [Node.js dotenv](https://www.npmjs.com/package/dotenv) version 10.0 (test cases use the same environment file to test parsing).

tests/system/Config/fixtures/.env
```diff
@@ -4,6 +4,12 @@ SPACED="with spaces"
 
 NULL=
 
+char.expo.foo="exported foo"
+character.export.var=variable
+export char.var=character
+export char.exports=imports
+fruit.export = "banana"
```
kenjis, line 11: Before: $_ENV
```
    'char.expo.foo' => string (12) "exported foo"
    'character..var' => string (8) "variable"
    ' char.var' => string (9) "character"
    ' char.s' => string (7) "imports"
    'fruit.' => string (6) "banana"
```
After: $_ENV
```
    'char.expo.foo' => string (12) "exported foo"
    'character.export.var' => string (8) "variable"
    'char.var' => string (9) "character"
    'char.exports' => string (7) "imports"
    'fruit.export' => string (6) "banana"
```

kenjis: Node.js dotenv@16.0.3:
```
  'char.expo.foo': 'exported foo',
  'character.export.var': 'variable',
  'char.var': 'character',
  'char.exports': 'imports',
  'fruit.export': 'banana'
```
It seems that `export` is handled the same way in Node dotenv as in this PR.

kenjis (approved):
As the bug fix, this has no problem.

paulbalandan:
Thanks for the reviews

PR #6637, opened 2022-10-06
fix: phpstan error in v1.8.8

ddevsr:
**Description**
See [https://github.com/codeigniter4/CodeIgniter4/actions/runs/3197626681/jobs/5221091878](https://github.com/codeigniter4/CodeIgniter4/actions/runs/3197626681/jobs/5221091878)

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/array_helper.php
+++ b/system/Helpers/array_helper.php
@@ -48,7 +48,7 @@ if (! function_exists('_array_search_dot')) {
         // Grab the current index
         $currentIndex = $indexes ? array_shift($indexes) : null;
 
-        if ((empty($currentIndex) && (int) $currentIndex !== 0) || (! isset($array[$currentIndex]) && $currentIndex !== '*')) {
+        if (empty($currentIndex) || (! isset($array[$currentIndex]) && $currentIndex !== '*')) {
             return null;
         }
 


ddevsr:
Now passed

[https://github.com/codeigniter4/CodeIgniter4/actions/runs/3198075094](https://github.com/codeigniter4/CodeIgniter4/actions/runs/3198075094)

datamweb (approved):
A poorly timed release!

MGatner (approved):

MGatner:
Thanks for the quick response @ddevsr! And the mention + review @datamweb. Waiting on that last test.

MGatner:
The PHPUnit failure looks unrelated (Predis connection failure m). Merging!

system/Helpers/array_helper.php
```diff
@@ -46,9 +46,10 @@ function dot_array_search(string $index, array $array)
     function _array_search_dot(array $indexes, array $array)
     {
         // Grab the current index
-        $currentIndex = $indexes ? array_shift($indexes) : null;
+        $currentIndex    = $indexes ? array_shift($indexes) : null;
+        $intCurrentIndex = (int) $currentIndex;
 
-        if ((empty($currentIndex) && (int) $currentIndex !== 0) || (! isset($array[$currentIndex]) && $currentIndex !== '*')) {
+        if (empty($currentIndex) && $intCurrentIndex !== 0 || (! isset($array[$currentIndex]) && $currentIndex !== '*')) {
```
kenjis, line 52: If `$currentIndex` is 0, `empty($currentIndex)` is true and `$intCurrentIndex !== 0` is false.
If `$currentIndex` is not 0, `empty($currentIndex)` is false and `$intCurrentIndex !== 0` is true.
So the left side is still always false.

kenjis: Does this seem better?
```php
        // Grab the current index
        $currentIndex = $indexes ? array_shift($indexes) : null;

        if ($currentIndex === null || (! isset($array[$currentIndex]) && $currentIndex !== '*')) {
            return null;
        }
```

MGatner: I'm sorry, I didn't realize it changed between my review and the merge. This was originally just removing the extraneous part of the conditional. 

Yours is better but this still has a lot of cognitive complexity and anyone working on it in the future will need to spend some time figuring it out. Maybe we should break it up into separate return statements? Or extract the second half of the conditional as a variable with a helpful name?

kenjis: Okay, I will fix this.

kenjis: I sent a PR #6645

PR #6638, opened 2022-10-06
docs: replace types in `Logger`

ddevsr:
**Description**
- See #6310
- Supersedes #6633

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -782,7 +782,7 @@ if (! function_exists('log_message')) {
      *  - info
      *  - debug
      *
-     * @return mixed
+     * @return bool
      */
     function log_message(string $level, string $message, array $context = [])
     {
--- a/system/Log/Handlers/ChromeLoggerHandler.php
+++ b/system/Log/Handlers/ChromeLoggerHandler.php
@@ -128,7 +128,7 @@ class ChromeLoggerHandler extends BaseHandler
     /**
      * Converts the object to display nicely in the Chrome Logger UI.
      *
-     * @param mixed $object
+     * @param array|int|object|string $object
      *
      * @return array
      */
--- a/system/Log/Logger.php
+++ b/system/Log/Logger.php
@@ -243,7 +243,7 @@ class Logger implements LoggerInterface
     /**
      * Logs with an arbitrary level.
      *
-     * @param mixed  $level
+     * @param string $level
      * @param string $message
      */
     public function log($level, $message, array $context = []): bool
@@ -312,9 +312,9 @@ class Logger implements LoggerInterface
      * {file}
      * {line}
      *
-     * @param mixed $message
+     * @param string $message
      *
-     * @return mixed
+     * @return string
      */
     protected function interpolate($message, array $context = [])
     {


tests/system/Log/Handlers/FileHandlerTest.php
```diff
@@ -77,6 +78,100 @@ public function testHandleCreateFile()
...
+    {
+        $config                                                = new LoggerConfig();
+        $config->handlers[TestHandler::class]['path']          = $this->start;
+        $config->handlers[TestHandler::class]['fileExtension'] = 'php';
+        $logger                                                = new MockFileLogger($config->handlers[TestHandler::class]);
+
+        $logger->setDateFormat('Y-m-d');
+        $date     = date('Y-m-d');
+        $expected = 'log-' . $date . '.php';
+        $logger->handle('info', 'This is a test log');
```
kenjis, line 91: These codes are time-dependent.
If `date('Y-m-d')` is called on today 23:59:59.9999 and `$logger->handle()` is called on tomorrow 0:00:00,
This test fails.

system/Log/Handlers/ChromeLoggerHandler.php
```diff
@@ -13,6 +13,7 @@
 
 use CodeIgniter\HTTP\ResponseInterface;
 use Config\Services;
+use Stringable;
```
MGatner, line 16: Stringable comes with PHP 8, we can't use it yet. 😢

MGatner (changes requested):

system/Log/Handlers/ChromeLoggerHandler.php
paulbalandan: Yes, that's true. But a `use` statement won't error on an unknown class.

tests/system/Log/Handlers/FileHandlerTest.php
ddevsr: Yes, rare moment to get difference `ms`. If want real same time must give parameter to set name file manual.

system/Log/Handlers/ChromeLoggerHandler.php
ddevsr: PHP-CS-Fixer automate `use` statement if set `string|\Stringable`

tests/system/Log/Handlers/FileHandlerTest.php
```diff
@@ -77,6 +78,100 @@ public function testHandleCreateFile()
...
+        $result = stream_get_contents($fp);
+        fclose($fp);
+
+        // did the log file get created?
+        $expectedResult = 'INFO - ' . $date . " --> stdClass Object\n(
+    [firstName] => John
+    [lastName] => Doe\n)\n\n";
+        $this->assertSame($expectedResult, $result);
+    }
+
```
paulbalandan, line 174: This formatting of `$message` was from commit https://github.com/codeigniter4/CodeIgniter4/commit/4fcb7607b6a082cf51cdd053f95a940abbc17ced. However, I think this is in contradiction to this line:
https://github.com/codeigniter4/CodeIgniter4/blob/53f17e8ff080b1490cd7df3c57ec2924c858e21b/system/Log/Logger.php#L44

Also, this is not compliant with PSR-3 considering that the `Logger` is implementing `Psr\Log\LoggerInterface`.

All in all, I believe this behavior is a bug to begin with.

ddevsr: What about now?

tests/system/Log/Handlers/FileHandlerTest.php
kenjis: Now `FileHandler::handle()` is using `date()`. We should use `Time` instead of `date()`.
Then we can stop time during test execution.

tests/system/Log/Handlers/FileHandlerTest.php
kenjis: I did not know that.

```php
log_message('error', $this); // TypeError
$this->logger->error($this); // Okay. but PHPStan error
```

```
 ------ --------------------------------------------------------------------------------------------------------------------- 
  Line   app/Controllers/Home.php                                                                                             
 ------ --------------------------------------------------------------------------------------------------------------------- 
  9      Parameter #2 $message of function log_message expects string, $this(App\Controllers\Home) given.                     
  10     Parameter #1 $message of method Psr\Log\LoggerInterface::error() expects string, $this(App\Controllers\Home) given.  
 ------ --------------------------------------------------------------------------------------------------------------------- 
```

> Also, this is not compliant with PSR-3 considering that the Logger is implementing Psr\Log\LoggerInterface.
> All in all, I believe this behavior is a bug to begin with.

Yes, but fixing this may break existing apps.

ddevsr: @kenjis Before changes in this PR appear PHP 2 errors. With condition now PR, what about ignoring this 2 error phpstan?

kenjis: Ignoring PHPStan errors is very bad practice. 
We should remove phpstan-baseline.neon.dist entries.

system/Log/Handlers/ChromeLoggerHandler.php
kenjis: As long as supporting PHP 7.4, we should not use `Stringable`.
It means a normal class `Stringable` and does not make sense.
See https://3v4l.org/TF3Ct

system/Common.php
```diff
@@ -782,9 +782,11 @@ function lang(string $line, array $args = [], ?string $locale = null)
      *  - info
      *  - debug
      *
-     * @return mixed
+     * @param string $message
+     *
+     * @return bool
      */
-    function log_message(string $level, string $message, array $context = [])
+    function log_message(string $level, $message, array $context = [])
```
kenjis, line 789: Please revert.

system/Log/Handlers/ChromeLoggerHandler.php
```diff
@@ -90,8 +91,8 @@ public function __construct(array $config = [])
      * will stop. Any handlers that have not run, yet, will not
      * be run.
      *
-     * @param string $level
-     * @param string $message
+     * @param string            $level
+     * @param string|Stringable $message
```
kenjis, line 95: Please revert.

system/Log/Handlers/ErrorlogHandler.php
```diff
@@ -62,8 +63,8 @@ public function __construct(array $config = [])
      * will stop. Any handlers that have not run, yet, will not
      * be run.
      *
-     * @param string $level
-     * @param string $message
+     * @param string            $level
+     * @param string|Stringable $message
```
kenjis, line 67: Please revert.

system/Log/Handlers/FileHandler.php
```diff
@@ -61,8 +62,8 @@ public function __construct(array $config = [])
      * will stop. Any handlers that have not run, yet, will not
      * be run.
      *
-     * @param string $level
-     * @param string $message
+     * @param string            $level
+     * @param string|Stringable $message
```
kenjis, line 66: Please revert.

system/Log/Handlers/HandlerInterface.php
```diff
@@ -22,8 +24,8 @@ interface HandlerInterface
      * will stop. Any handlers that have not run, yet, will not
      * be run.
      *
-     * @param string $level
-     * @param string $message
+     * @param string            $level
+     * @param string|Stringable $message
```
kenjis, line 28: Please revert.

system/Log/Logger.php
```diff
@@ -152,7 +160,7 @@ public function __construct($config, bool $debug = CI_DEBUG)
     /**
      * System is unusable.
      *
-     * @param string $message
+     * @param string|Stringable $message
```
kenjis, line 163: Please revert.

system/Log/Logger.php
```diff
@@ -165,7 +173,7 @@ public function emergency($message, array $context = []): bool
      * Example: Entire website down, database unavailable, etc. This should
      * trigger the SMS alerts and wake you up.
      *
-     * @param string $message
+     * @param string|Stringable $message
```
kenjis, line 176: Please revert.

system/Log/Logger.php
```diff
@@ -177,7 +185,7 @@ public function alert($message, array $context = []): bool
      *
      * Example: Application component unavailable, unexpected exception.
      *
-     * @param string $message
+     * @param string|Stringable $message
```
kenjis, line 188: Please revert.

system/Log/Logger.php
```diff
@@ -188,7 +196,7 @@ public function critical($message, array $context = []): bool
      * Runtime errors that do not require immediate action but should typically
      * be logged and monitored.
      *
-     * @param string $message
+     * @param string|Stringable $message
```
kenjis, line 199: Please revert.

system/Log/Logger.php
```diff
@@ -201,7 +209,7 @@ public function error($message, array $context = []): bool
      * Example: Use of deprecated APIs, poor use of an API, undesirable things
      * that are not necessarily wrong.
      *
-     * @param string $message
+     * @param string|Stringable $message
```
kenjis, line 212: Please revert.

system/Log/Logger.php
```diff
@@ -211,7 +219,7 @@ public function warning($message, array $context = []): bool
     /**
      * Normal but significant events.
      *
-     * @param string $message
+     * @param string|Stringable $message
```
kenjis, line 222: Please revert.

system/Log/Logger.php
```diff
@@ -223,7 +231,7 @@ public function notice($message, array $context = []): bool
      *
      * Example: User logs in, SQL logs.
      *
-     * @param string $message
+     * @param string|Stringable $message
```
kenjis, line 234: Please revert.

system/Log/Logger.php
```diff
@@ -233,7 +241,7 @@ public function info($message, array $context = []): bool
     /**
      * Detailed debug information.
      *
-     * @param string $message
+     * @param string|Stringable $message
```
kenjis, line 244: Please revert.

system/Log/Logger.php
```diff
@@ -243,8 +251,8 @@ public function debug($message, array $context = []): bool
     /**
      * Logs with an arbitrary level.
      *
-     * @param mixed  $level
-     * @param string $message
+     * @param string            $level
+     * @param string|Stringable $message
```
kenjis, line 255: Please revert.

system/Log/Logger.php
```diff
@@ -312,14 +316,16 @@ public function log($level, $message, array $context = []): bool
      * {file}
      * {line}
      *
-     * @param mixed $message
+     * @param string|Stringable $message
```
kenjis, line 319:
```suggestion
     * @param string $message
```

system/Log/Logger.php
```diff
@@ -312,14 +316,16 @@ public function log($level, $message, array $context = []): bool
...
      *
-     * @return mixed
+     * @return string|Stringable
```
kenjis, line 321:
```suggestion
     * @return string
```

system/Log/Handlers/ChromeLoggerHandler.php
```diff
@@ -128,7 +129,7 @@ public function handle($level, $message): bool
     /**
      * Converts the object to display nicely in the Chrome Logger UI.
      *
-     * @param mixed $object
+     * @param array|int|object|string|Stringable $object
```
kenjis, line 132:
```suggestion
     * @param string $object
```

system/Log/Logger.php
```diff
@@ -414,6 +420,14 @@ public function determineFile(): array
         ];
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function __toString()
```
kenjis, line 426: We should not use ` __toString()`, because a Logger object is not stringable.

system/Log/Logger.php
```diff
@@ -102,6 +103,13 @@ class Logger implements LoggerInterface
      */
     public $logCache;
 
+    /**
+     * Temporary messages.
+     *
+     * @var Stringable
+     */
+    protected $tempMessages;
```
kenjis, line 111: Please remove.
Logger should not hold the message.

kenjis (changes requested):
It is recommended to only change the PHPDoc types. Because this is title is *docs: replace types in Logger*.
If you want to change any implementation, it is better to send another PR.

system/Log/Logger.php
ddevsr: Only message or both?

system/Log/Handlers/ChromeLoggerHandler.php
ddevsr: Only string appear PHPStan error

system/Log/Logger.php
kenjis: Only message.

kenjis (approved):

system/Log/Logger.php
```diff
@@ -265,10 +265,6 @@ public function log($level, $message, array $context = []): bool
         // Parse our placeholders
         $message = $this->interpolate($message, $context);
 
-        if (! is_string($message)) {
```
paulbalandan, line 268: Please revert. This should be for a separate PR.

system/Log/Logger.php
```diff
@@ -329,7 +325,7 @@ protected function interpolate($message, array $context = [])
             // Verify that the 'exception' key is actually an exception
             // or error, both of which implement the 'Throwable' interface.
             if ($key === 'exception' && $val instanceof Throwable) {
-                $val = $val->getMessage() . ' ' . $this->cleanFileNames($val->getFile()) . ':' . $val->getLine();
+                $val = $val->getMessage() . ' ' . clean_path($val->getFile()) . ':' . $val->getLine();
```
paulbalandan, line 328: This is a refactor. Please revert.

system/Test/Mock/MockLogger.php
```diff
@@ -95,6 +95,22 @@ class MockLogger
                 'notice',
                 'warning',
             ],
+            /*
```
paulbalandan, line 98: Revert the changes here and in the added tests.

tests/system/Log/Handlers/FileHandlerTest.php
paulbalandan: Please revert.

paulbalandan (changes requested):
Let's minimize for now the changes in this PR to the docblock changes, otherwise those "other" changes would be a blocker for the merge.

Also, please squash the unrelated "trial-and-error" commits into the others. We don't want to show an unordered history of the file.

kenjis (changes requested):
Please squash commits.

kenjis (approved):

MGatner (approved):
Thanks @ddevrs

paulbalandan (approved):

PR #6651, opened 2022-10-09
refactor: remove unnecessary "&" in Filters

kenjis:
**Description**
`$response` is an object, so it is passed as an object id.
https://www.php.net/manual/en/language.oop5.references.php

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #6652, opened 2022-10-09
feat: add method to disable controller filters

kenjis:
**Description**
- add method to disable controller filters

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -153,6 +153,11 @@ class CodeIgniter
     protected ?string $context = null;
 
     /**
+     * Whether to enable Control Filters.
+     */
+    protected bool $enableFilters = true;
+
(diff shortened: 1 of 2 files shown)


MGatner (approved):
An interesting feature. What do you recommend this for? Troubleshooting?

kenjis:
I would like to make an internal sub request.
That is, creating another request in a controller like this:
https://github.com/ivanvoitovych/viewi-symfony-demo/blob/5c944f1c919efacc045c633d659e415bccd6e721/src/Adapters/ViewiSymfonyAdapter.php#L25-L27

Another use case I imagine is, when using CLI controllers and no filters are needed.

PR #6658, opened 2022-10-10
Update closing HTML tags in form_helper.php

### LINKED ISSUE #6649: Dev: remove self-closing HTML tags in helpers
> The W3C validator displays a warning for : Self-closing tag syntax in text/html documents is widely discouraged.

See
- https://github.com/codeigniter4/CodeIgniter4/issues/6602#issuecomment-1268432153
- https://forum.codeigniter.com/showthread.php?tid=83552

visserrobin:
**Description**
Fixes #6649 
Removed the '/' in closing tags for input fields for text, upload and checkbox. 

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -148,7 +148,7 @@ if (! function_exists('form_input')) {
             'value' => $value,
         ];
 
-        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " />\n";
+        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " >\n";
     }
 }
 
@@ -194,7 +194,7 @@ if (! function_exists('form_upload')) {
 
         $data['type'] = 'file';
 
-        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " />\n";
+        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " >\n";
     }
 }
 
@@ -366,7 +366,7 @@ if (! function_exists('form_checkbox')) {
             unset($defaults['checked']);
         }
 
-        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " />\n";
+        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " >\n";
     }
 }
 


ddevsr:
Dont forget changes unit tests for this PR

kenjis:
Sorry, this is not a bug fix, so please send it to `4.3` not `develop`.
See https://github.com/codeigniter4/CodeIgniter4/blob/caa355d78bc2eb0029842ee3765d6dda23f95f71/contributing/workflow.md#if-you-sent-to-the-wrong-branch

neznaika0:
Please remove the space at the end.
The output will be: 
<input type="text" > to <input type="text">

sclubricants:
@kenjis 

Im not sure that removing the closing tag is the best way to go here. The closing tag is needed for the more stricter XHTML. XHTML requires all elements to be closed including void elements. The closing tag on a plain HTML document is still valid even if some validation says its not. It is simply ignored. By omitting these you are causing problems for those who want to use XHTML markup. The presence of them has no affect on plain HTML markup.

If you really have to get rid of them then I would have a config definition to decide how to render the tags.

kenjis:
I think it is better to follow the standard.
https://html.spec.whatwg.org/

[XHTML 1.1 specification](https://www.w3.org/TR/2018/SPSD-xhtml11-20180327/) is Superseded Recommendation.

kenjis:
@sclubricants 
> If you really have to get rid of them then I would have a config definition to decide how to render the tags.

Do you need to support XHTML in year 2022?

sclubricants:
I don't use it with codeigniter but I do in Salesforce. Just saying some people might be using it and if you change it it could cause problems.

sclubricants:
I at least think this is a breaking change.

ddevsr:
No activity by author, i can send PR at home later

ddevsr:
One of criteria SEO is data structure schema must be `valid`,

Google said on [https://developers.google.com/search/docs/appearance/structured-data](https://developers.google.com/search/docs/appearance/structured-data)
```
We removed Google-specific validation from the Structured Data Testing Tool and migrated the tool to a new domain, Schema Markup Validator (https://validator.schema.org/).
```

And now based on HTML5, we must support standard structure.

ddevsr:
@kenjis I have problem in my env with PHPStan 1.8.10

```bash
------ ---------------------------------------------------------------------------------------------------------------------------- 
  Line   system\Helpers\form_helper.php
 ------ ----------------------------------------------------------------------------------------------------------------------------      
         Ignored error pattern #^Offset 'checked' on array\{type\: 'checkbox', name\: mixed, value\: string\} in isset\(\) does not       
         exist\.$# in path D:\Project\laragon\www\ci4form\system\Helpers\form_helper.php was not matched in reported errors.
  366    Offset 'checked' on array{type: 'checkbox', name: string, value: string} in isset() does not exist.
 ------ ----------------------------------------------------------------------------------------------------------------------------      


 [ERROR] Found 2 errors

Fix the phpstan error(s) before commit.
```

```php
    function form_checkbox($data = '', string $value = '', bool $checked = false, $extra = ''): string
    {
        $defaults = [
            'type'  => 'checkbox',
            'name'  => (! is_array($data) ? $data : ''),
            'value' => $value,
        ];

        if (is_array($data) && array_key_exists('checked', $data)) {
            $checked = $data['checked'];
            if ($checked === false) {
                unset($data['checked']);
            } else {
                $data['checked'] = 'checked';
            }
        }

        if ($checked === true) {
            $defaults['checked'] = 'checked';
        } elseif (isset($defaults['checked'])) {
            unset($defaults['checked']);
        }

        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . ">\n";
    }
```

kenjis:
@ddevsr I cannot reproduce.

ddevsr:
> @ddevsr I cannot reproduce. Can you send a PR?

Okay, i will send PR with PHPStan 1.8.8

kenjis:
See #6717

PR #6660, opened 2022-10-10
Modify the user guide of the insert() method to include the second parameter

domaingenerator:
**Description**
The insert method takes 2 paraemtres, but only 1 parametre has been detailed. Furthermore, the description currently details using the getInsertID() to get the rows insert id, but this is already the default return value, so it doesn't make sense to add this to the description.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


user_guide_src/source/models/model/015.php
```diff
@@ -5,4 +5,4 @@
     'email'    => 'd.vader@theempire.com',
 ];
 
-$userModel->insert($data);
+$userModel->insert($data, bool $returnID = true);
```
kenjis, line 8: Please revert. It is syntax error, and `$returnID = true`  is the default value, we don't need it.

user_guide_src/source/models/model.rst
```diff
@@ -311,13 +311,15 @@ Saving Data
...
-row of data in the database. The array's keys must match the name of the columns in a ``$table``, while
-the array's values are the values to save for that key:
+The first parametre is an associative array of data to create a new row of data in the database. 
+If an object is passed instead of an array, it will attempt to convert it to an array.
+
+The array's keys must match the name of the columns in the ```$table```, while the array's values are the values to save for that key:
 
 .. literalinclude:: model/015.php
 
-You can retrieve the last inserted row's primary key using the ``getInsertID()`` method.
```
kenjis, line 320: Do not delete the description for ``getInsertID()``. The method is only here.

user_guide_src/source/models/model.rst
```diff
@@ -311,13 +311,15 @@ Saving Data
...
+The second parametre is of type boolean which determines whether an insert ID should be returned. 
+If it is omitted, then the default return value is the insert id.
```
kenjis, line 322: What if it is `false`?

kenjis:
Thank you for sending the PR!

You forgot the signing to your commits.

user_guide_src/source/models/model.rst
domaingenerator: This results in an error: Call to a member function getInsertID() on bool.

`echo $this->users->insert($data, false)->getInsertID();`

The way to get the insert ID is to omit the second parameter or if its `false` use the `getInsertID()` on the model:

`$this->users->getInsertID();`

Which is not related to the `insert()` method. So do you think having details related to the `getInsertID()` method under the `insert()` to be out of context?

Should it just have its own section like `insert()`, `update()`, `getInsertID()` ...etc?

kenjis: `getInsertID()` is related to only insert. So it should be under `insert()` section.

domaingenerator: As far as I can tell it is used `$this->model->getInsertID();` not `$this->model->insert($data)->getInsertID();` so should a more detailed paragraph under the `insert()` be created then if its too small to be in its own section?

user_guide_src/source/models/model/015.php
```diff
@@ -5,4 +5,8 @@
     'email'    => 'd.vader@theempire.com',
 ];
 
+// Returns inserted row's primary key
 $userModel->insert($data);
+
+// Returns true on success and false on failure
+$userModel->insert($data, false);
```
michalsn, line 12:
```suggestion
// Inserts data and returns inserted row's primary key
$userModel->insert($data);

// Inserts data and returns true on success and false on failure
$userModel->insert($data, false);

// Returns inserted row's primary key
$userModel->getInsertID();
```

user_guide_src/source/models/model.rst
michalsn: I agree with kenjis. There is no need for a detailed paragraph because the `getInsertID()` method is only usable in the context of the `insert()` method.

The only thing that might be missing is a simple example + description, which should clear up any doubts.

domaingenerator:
The suggestions were perfect. Thank you.

kenjis (changes requested):
The contents seem good!

Please squash the commits, and sign.
Because there are some commits without signing.

domaingenerator:
Its possible my squash / commit went wrong!

domaingenerator:
Yep, it squashed into the wrong commit. I'll do it again.

user_guide_src/source/models/model/015.php
```diff
@@ -5,4 +5,12 @@
...
 
+// Inserts data and returns inserted row's primary key
 $userModel->insert($data);
+
+// Inserts data and returns true on success and false on failure
+$userModel->insert($data, false);
+
+// Returns inserted row's primary key
+$userModel->getInsertID();
+
```
kenjis, line 16: Please remove the empty line.

user_guide_src/source/models/model/015.php
```diff
@@ -5,4 +5,11 @@
...
 ];
 
+// Inserts data and returns inserted row's primary key
 $userModel->insert($data);
+
+// Inserts data and returns true on success and false on failure
+$userModel->insert($data, false);
+
+// Returns inserted row's primary key
+$userModel->getInsertID();
```
kenjis, line 15: You deleted too much!
We need LF at the end of a file.

domaingenerator: Thanks I'll check that out.

kenjis (approved):

kenjis:
Thank you!

Issue #5344, opened 2021-11-18
IncomingRequest property $uri is public and should be protected


pawelkg:
### PHP Version

8.0

### CodeIgniter4 Version

4.1.5

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

apache

### What happened?

It's not a bug, but an inconsistency. Class `Request` have property `$uri` which is protected, but `IncomingRequest` class change this access level to public. It's should be protected also, especially that there is `getUri()` method to get it. Unfortunetly I don't see this method in CodeIniter documentation. This is why I assume that most of people get this by property name, not method (as I done it first), so changing it can couse problems in projects.

### Steps to Reproduce

In eg. Controller usage of `$this->request->uri` Should not be available. Should use this: `$this->request->getUri()`

### Expected Output

`$this->request->uri`
Cannot access protected property CodeIgniter\HTTP\IncomingRequest::$uri

kenjis:
Thank  you!
I agree with you.

Probably it seems `$uri` is public because of historical reasons.
And we can't change the visibility in short term, because it is a  BC break.

I sent a PR #5345 for now.

kenjis:
Ref #3230

MGatner:
Yes as @kenjis pointed out we are stuck with this for now for backwards-compatibility but very much would like it changed. Version 5 already has plans on refactoring the HTTP layer to be PSR-compliant so that will address this. Safe to close this issue.

kenjis:
We deprecated $uri in v4.3.0. #6662
So we can remove it in `4.5`.

> The deprecated items are not covered by backwards compatibility (BC) promise. It may be removed in the next next minor version or later. For example, if an item has been deprecated since 4.3.x, it may be removed in 4.5.0.
https://codeigniter4.github.io/CodeIgniter4/installation/backward_compatibility_notes.html#what-are-not-breaking-changes

kenjis:
Closed by #8067

PR #6662, opened 2022-10-11
Deprecate public properties in IncomingRequest

kenjis:
**Description**
See #5344

Deprecate
- IncomingRequest::$uri
- IncomingRequest::$config

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


MGatner (approved):
We should have `readonly` properties available by the next major version. I don't necessarily think any of these are good candidates, but something to keep in mind.

PR #6665, opened 2022-10-11
Deallocate prepared statements

### LINKED ISSUE #1256: Deallocate prepared statements
Database/SQLite3/PreparedQuery::_execute(..) has todo note...
finalize()

fpoy:
Fixes #1256

**Description**

Implement a common method in each database connector to destroy the prepared query when we're done with it

Cf Issue codeigniter4/CodeIgniter4#1256 : TODO Deallocate prepared statements

To closes a prepared statement :

- MySQLi have mysqli_stmt_close()
  https://www.php.net/manual/en/mysqli-stmt.close.php
- OCI8 have oci_free_statement()
  https://www.php.net/manual/en/function.oci-free-statement.php
- Postgre does not have a specific PHP function for deleting a prepared statement,
  the SQL DEALLOCATE statement can be used for that purpose :
  https://www.php.net/manual/en/function.pg-prepare.php
  https://www.postgresql.org/docs/8.1/sql-deallocate.html
- SQLite3 have SQLite3Stmt::close()
  https://www.php.net/manual/en/sqlite3stmt.close.php
- SQLSRV have sqlsrv_free_stmt()
  https://www.php.net/manual/en/function.sqlsrv-free-stmt.php

**Checklist:**

- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -148,6 +148,8 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
 
     /**
      * Explicitly closes the prepared statement.
+     *
+     * @throws BadMethodCallException
      */
     public function close(): bool
     {
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -157,7 +157,7 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
     /**
      * The database dependant version of the close method.
      */
-    abstract public function _close(): bool;
+    abstract protected function _close(): bool;
 
     /**
      * Returns the SQL that has been prepared.
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -25,7 +25,7 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
     /**
      * The prepared statement itself.
      *
-     * @var object|resource
+     * @var object|resource|null
      */
     protected $statement;
 
@@ -158,7 +158,7 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
         try {
             return $this->_close();
         } finally {
-            unset($this->statement);
+            $this->statement = null;
         }
     }
 
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -163,7 +163,7 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
     }
 
     /**
-     * The database dependant version of the close method.
+     * The database-dependent version of the close method.
      */
     abstract protected function _close(): bool;
 
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -151,7 +151,15 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
      */
     public function close(): bool
     {
-        return $this->_close();
+        if (! isset($this->statement)) {
+            return true;
+        }
+
+        try {
+            return $this->_close();
+        } finally {
+            unset($this->statement);
+        }
     }
 
     /**
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -147,16 +147,17 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
     abstract public function _getResult();
 
     /**
-     * Explicitly closes the statement.
+     * Explicitly closes the prepared statement.
      */
-    public function close()
+    public function close(): bool
     {
-        if (! is_object($this->statement) || ! method_exists($this->statement, 'close')) {
-            return;
-        }
-
-        $this->statement->close();
+        return $this->_close();
     }
+
+    /**
+     * The database dependant version of the close method.
+     */
+    abstract public function _close(): bool;
 
     /**
      * Returns the SQL that has been prepared.
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -152,7 +152,7 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
     public function close(): bool
     {
         if (! isset($this->statement)) {
-            return true;
+            throw new BadMethodCallException('Cannot call close on a non-existing prepared statement.');
         }
 
         try {
--- a/system/Database/Postgre/PreparedQuery.php
+++ b/system/Database/Postgre/PreparedQuery.php
@@ -105,7 +105,7 @@ class PreparedQuery extends BasePreparedQuery
     {
         $error = true;
         if (isset($this->statement)) {
-            $error = (bool) pg_query($this->db->connID, 'DEALLOCATE ' . $this->name);
+            $error = (bool) pg_query($this->db->connID, 'DEALLOCATE "' . $this->db->escapeIdentifiers($this->name) . '"');
             unset($this->statement);
         }
 
--- a/system/Database/Postgre/PreparedQuery.php
+++ b/system/Database/Postgre/PreparedQuery.php
@@ -103,13 +103,7 @@ class PreparedQuery extends BasePreparedQuery
      */
     protected function _close(): bool
     {
-        $error = true;
-        if (isset($this->statement)) {
-            $error = (bool) pg_query($this->db->connID, 'DEALLOCATE "' . $this->db->escapeIdentifiers($this->name) . '"');
-            unset($this->statement);
-        }
-
-        return $error;
+        return (bool) pg_query($this->db->connID, 'DEALLOCATE "' . $this->db->escapeIdentifiers($this->name) . '"');
     }
 
     /**
--- a/system/Database/Postgre/PreparedQuery.php
+++ b/system/Database/Postgre/PreparedQuery.php
@@ -101,7 +101,7 @@ class PreparedQuery extends BasePreparedQuery
     /**
      * Deallocate prepared statements
      */
-    public function _close(): bool
+    protected function _close(): bool
     {
         $error = true;
         if (isset($this->statement)) {
--- a/system/Database/Postgre/PreparedQuery.php
+++ b/system/Database/Postgre/PreparedQuery.php
@@ -99,6 +99,20 @@ class PreparedQuery extends BasePreparedQuery
     }
 
     /**
+     * Deallocate prepared statements
+     */
+    public function _close(): bool
+    {
+        $error = true;
+        if (isset($this->statement)) {
+            $error = (bool) pg_query($this->db->connID, 'DEALLOCATE ' . $this->name);
+            unset($this->statement);
+        }
+
+        return $error;
+    }
+
+    /**
      * Replaces the ? placeholders with $1, $2, etc parameters for use
      * within the prepared query.
      */
--- a/system/Database/Postgre/PreparedQuery.php
+++ b/system/Database/Postgre/PreparedQuery.php
@@ -103,7 +103,7 @@ class PreparedQuery extends BasePreparedQuery
      */
     protected function _close(): bool
     {
-        return (bool) pg_query($this->db->connID, 'DEALLOCATE "' . $this->db->escapeIdentifiers($this->name) . '"');
+        return pg_query($this->db->connID, 'DEALLOCATE "' . $this->db->escapeIdentifiers($this->name) . '"') !== false;
     }
 
     /**
--- a/tests/system/Database/Live/PreparedQueryTest.php
+++ b/tests/system/Database/Live/PreparedQueryTest.php
@@ -42,12 +42,10 @@ final class PreparedQueryTest extends CIUnitTestCase
     {
         parent::tearDown();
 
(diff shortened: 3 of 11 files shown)


kenjis:
Thank you for contribution!

This is not a bug fix, so this should go to `4.3` branch.

fpoy:
@kenjis that's done, I've rebased the PR to 4.3 branch

kenjis:
@fpoy There are four commits made by Paul and me. Please drop them.
You can do it with `git rebase`.

fpoy:
> @fpoy There are four commits made by Paul and me. Please drop them. You can do it with `git rebase`. See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#updating-your-branch

@kenjis it's done, there are only my commits left on the branch...

kenjis:
Thank you. The commits are okay.

1. We need the documentation. Please add changelog.
2. If possible, please reduce rather than increase `phpstan-baseline.neon.dist`.

system/Database/BasePreparedQuery.php
```diff
@@ -147,17 +147,18 @@ abstract public function _execute(array $data): bool;
     abstract public function _getResult();
 
     /**
-     * Explicitly closes the statement.
+     * Explicitly closes the prepared statement.
      */
-    public function close()
+    public function close(): bool
```
kenjis, line 152: This is a breaking change. Please add it to the changelog:
https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/changelogs/v4.3.0.rst#others-1

fpoy: done

fpoy:
The documentation on this subject already exists
What should be added to it ?
image

fpoy:
in `user_guide_src/source/database/queries/020.php`
```
$error = $pQuery->close();
```
Would that be sufficient ?

kenjis:
Please add about this enhancement in the Enhancements section:
https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/changelogs/v4.3.0.rst#database

kenjis:
> The documentation on this subject already exists
> What should be added to it ?

https://codeigniter4.github.io/CodeIgniter4/database/queries.html#close

It seems the  `close()` method does nothing with some DBMS before this PR.
How about adding the note for it?
Something like this:
> Since v4.3.0, the close() method deallocates the prepared query in all DBMS. In previous versions, the prepared query was not deallocated in xxx and xxx.

sclubricants:
LGMT

The only thing is you might add to the test to check that the statement is actually cleared. For MySQL:

```SQL
SHOW GLOBAL STATUS LIKE '%prepared_stmt_count%';
# 0
```

SQLITE https://www.sqlite.org/c3ref/stmt_status.html

Postgres https://www.postgresql.org/docs/current/view-pg-prepared-statements.html

user_guide_src/source/database/queries.rst
```diff
@@ -240,6 +242,8 @@ close out the prepared statement when you're done with it:
 
 .. literalinclude:: queries/020.php
 
+.. note:: Since v4.3.0, the close() method deallocates the prepared statement in all DBMS. Previously, they were not deallocated in Postgre, SQLSRV and OCI8.
```
kenjis, line 245:
```suggestion
.. note:: Since v4.3.0, the ``close()`` method deallocates the prepared statement in all DBMS. Previously, they were not deallocated in Postgre, SQLSRV and OCI8.
```

system/Database/BasePreparedQuery.php
```diff
@@ -147,17 +147,18 @@ abstract public function _execute(array $data): bool;
...
-        }
-
-        $this->statement->close();
+        return $this->_close();
     }
 
+    /**
+     * The database dependant version of the close method.
+     */
+    abstract public function _close(): bool;
```
paulbalandan, line 160: Why is this `public`? Shouldn't this be `protected` since you are using the template pattern for the `close` method?

system/Database/BasePreparedQuery.php
```diff
@@ -147,17 +147,18 @@ abstract public function _execute(array $data): bool;
...
      */
-    public function close()
+    public function close(): bool
     {
-        if (! is_object($this->statement) || ! method_exists($this->statement, 'close')) {
-            return;
-        }
-
-        $this->statement->close();
+        return $this->_close();
```
paulbalandan, line 154: It seems to me PHPCPD didn't detect the amount of code duplication in the implementations of `_close()`. And the common pattern is (1) setting `$errors` to `true`, (2) checking existence of `$this->statement`, (3) doing the driver specific closing, (4) unsetting `$this->statement`, and (5) return the `$errors`.

Would it be better to do this?
```php
public function close(): bool
{
    if (! isset($this->statement)) {
        return true;
    }

    try {
        return $this->_close();
    } finally {
        unset($this->statement);
    }
}

abstract protected function _close(): bool;
```

michalsn: That's right. I would like to see these changes.

system/Database/BasePreparedQuery.php
fpoy: Indeed, it is `public`, just like the declarations of prepare/_prepare and execute/_execute methods

But if the database-dependent part should be protected, shouldn't it be for these 3 methods also ?

paulbalandan: Hmm... that should be wrong. That would be unnecessary visibility for those methods. However, we cannot change the visibilities of `_prepare` and `_execute` as it would be a BC break.

Let us start the correct way with this one and do the others in separate PRs.

user_guide_src/source/database/queries.rst
fpoy: done...

system/Database/BasePreparedQuery.php
fpoy: OK, i've changed the visibility to "protected"

system/Database/BasePreparedQuery.php
fpoy: OK, done.  I've moved duplication in the implementations of _close() to the base class as suggested

system/Database/Postgre/PreparedQuery.php
```diff
@@ -98,6 +98,14 @@ public function _getResult()
         return $this->result;
     }
 
+    /**
+     * Deallocate prepared statements
+     */
+    protected function _close(): bool
+    {
+        return (bool) pg_query($this->db->connID, 'DEALLOCATE "' . $this->db->escapeIdentifiers($this->name) . '"');
```
paulbalandan, line 106: I prefer to see this checked against its specific return type.
```suggestion
        return pg_query($this->db->connID, 'DEALLOCATE "' . $this->db->escapeIdentifiers($this->name) . '"') !== false;
```

system/Database/BasePreparedQuery.php
```diff
@@ -147,17 +147,26 @@ abstract public function _execute(array $data): bool;
...
-        $this->statement->close();
+        try {
+            return $this->_close();
+        } finally {
+            unset($this->statement);
+        }
     }
 
+    /**
+     * The database dependant version of the close method.
```
paulbalandan, line 166:
```suggestion
     * The database-dependent version of the close method.
```

user_guide_src/source/database/queries/020.php
```diff
@@ -1,3 +1,3 @@
 <?php
 
-$pQuery->close();
+$bool = $pQuery->close();
```
paulbalandan, line 3: The variable is unused.

system/Database/Postgre/PreparedQuery.php
fpoy: I've changed this
But other methods in Postgre connector are still using the cast to boolean (eg in Database\Postgre\Connection)

system/Database/BasePreparedQuery.php
fpoy: fixed...

user_guide_src/source/database/queries/020.php
fpoy: OK, I've completed the example query to better reflect the new mode of operation

system/Database/Postgre/PreparedQuery.php
kenjis: You can send another PR to refactor the code, if you like.

system/Database/BasePreparedQuery.php
```diff
@@ -147,17 +147,26 @@ abstract public function _execute(array $data): bool;
...
-     * Explicitly closes the statement.
+     * Explicitly closes the prepared statement.
      */
-    public function close()
+    public function close(): bool
     {
-        if (! is_object($this->statement) || ! method_exists($this->statement, 'close')) {
-            return;
+        if (! isset($this->statement)) {
+            return true;
```
kenjis, line 155: Why return `true`?
`BadMethodCallException`?

tests/system/Database/Live/PreparedQueryTest.php
```diff
@@ -162,12 +166,20 @@ public function testDeallocatePreparedQuery()
...
         } catch (BadMethodCallException $e) {
             $this->assertSame('You must call prepare before trying to execute a prepared statement.', $e->getMessage());
         }
 
+        // Try to close a non-existing prepared statement
+        try {
+            $this->query->close();
+        } catch (BadMethodCallException $e) {
+            $this->assertSame('Cannot call close on a non-existing prepared statement.', $e->getMessage());
+        }
```
kenjis, line 181: Please write tests for Exceptions using `$this->expectException()` and `$this->expectExceptionMessage()`.
Not try catch.

system/Database/BasePreparedQuery.php
```diff
@@ -147,17 +147,26 @@ abstract public function _execute(array $data): bool;
...
-            return;
+        if (! isset($this->statement)) {
+            throw new BadMethodCallException('Cannot call close on a non-existing prepared statement.');
         }
 
-        $this->statement->close();
+        try {
+            return $this->_close();
+        } finally {
+            unset($this->statement);
```
paulbalandan, line 161: I just found out that unsetting a property removes it from the object at runtime, which may not be desired or when accessed accidentally after will error "Undefined property ...". I think this should instead be set to `null`.
```suggestion
            $this->statement = null;
```

Then, to satisfy phpstan, please add `|null` in the docblock of `$statement` in `BasePreparedQuery`.

system/Database/BasePreparedQuery.php
fpoy: Done, a `BadMethodCallException` is now thrown when trying to close a non-existing prepared statement

tests/system/Database/Live/PreparedQueryTest.php
fpoy: Done, I've refactored the corresponding unit tests

system/Database/BasePreparedQuery.php
paulbalandan: That should be ok.

system/Database/BasePreparedQuery.php
paulbalandan: Please add `@throws` in the docblock of `close()`

fpoy: Done, docblock added

kenjis:
Please rebase to resolve the conflict.

tests/system/Database/Live/PreparedQueryTest.php
```diff
@@ -42,7 +43,11 @@ protected function tearDown(): void
         parent::tearDown();
 
         if ($this->query !== null) {
-            $this->query->close();
+            try {
+                $this->query->close();
+            } catch (BadMethodCallException $e) {
+                $this->query = null;
+            }
         }
```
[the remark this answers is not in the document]
paulbalandan: Ah you're right! I'm mixing things up. Sorry for that.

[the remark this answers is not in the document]
fpoy: I just removed the null-check because we need to keep the try/catch because during the `testDeallocatePreparedQueryThenTryToClose` method, there are two calls to `close`, and an exception is throwed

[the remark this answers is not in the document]
fpoy: Sorry I did not see your comments in time, and I pushed an update
The unit test are passing without the null check
Do I need to revert, or do I keep the code as it is now ?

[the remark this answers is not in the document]
kenjis: This is testing for BasePreparedQuery, and if there are no cases where $this->query is null, then it is fine.

kenjis (approved):

MGatner (commented):
Looks great! I can't tell if that last thread is resolved yet. 

Whoever merges: please squash this one.

paulbalandan (approved):

tests/system/Database/Live/PreparedQueryTest.php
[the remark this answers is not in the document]
fpoy: ok so I'm not touching anything anymore

PR #6172, opened 2022-06-22
Add PHP 8.2 to Unit Tests GHA

paulbalandan:
See #6170

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

PR #6667, opened 2022-10-11
test: remove dynamic properties in MigrationRunnerTest

kenjis:
**Description**
See #6170, #6172

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


MGatner:
Maybe? Refactoring Migration Runner was one of my first larger projects, but it was so long ago I remember next to nothing about it.

PR #6668, opened 2022-10-11
chore: update PHPStan PHP version to 8.1

kenjis:
**Description**
PHP 8.1 is the most used version with Composer.

Screenshot 2022-10-12 8 15 14

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/.github/workflows/test-phpstan.yml
+++ b/.github/workflows/test-phpstan.yml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.0', '8.1']
+        php-versions: ['8.1']
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/test-phpstan.yml
+++ b/.github/workflows/test-phpstan.yml
@@ -33,8 +33,6 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
-      matrix:
-        php-versions: ['8.1']
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -42,7 +40,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: ${{ matrix.php-versions }}
+          php-version: '8.1'
           extensions: intl
 
       - name: Use latest Composer
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -301,7 +301,7 @@ parameters:
 			path: system/Database/MySQLi/Result.php
 
 		-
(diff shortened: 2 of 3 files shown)


.github/workflows/test-phpstan.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.0', '8.1']
+        php-versions: ['8.1']
```
paulbalandan, line 37: Remove `matrix` then put php-version in L45.

MGatner (commented):
Shouldn't we be running PHPStan against each version of PHP we want to support? There doesn't appear to be a command line argument for it but we could have a config file for each that imports the main file.

kenjis:
> Shouldn't we be running PHPStan against each version of PHP we want to support?

Ideally yes, but IIRC we were not able to support PHP 7.4 and 8.0 at the same time.
And PHP 8.1 has new classes like `PgSql\Result`, so it seems we cannot support 8.0 and 8.1 after all.

Or if we prepare `phpstan.neon.dist` for each version, we can do it?
But it costs more.

paulbalandan:
It would be cumbersome to maintain, as shown before. We have code flagged by phpstan in a lower PHP version which we add `@phpstan-ignore-line`. However, that line is perfectly okay with phpstan when run on a higher PHP version, erroring now at "No error to ignore..."

kenjis:
It seems to me that rather than increasing the number of PHP versions to check, we should focus on reducing `phpstan-baseline.neon.dist` and `@phpstan-ignore-line` and increasing the level (now `level: 5`).

.github/workflows/test-phpstan.yml
kenjis: Done.

paulbalandan (approved):

kenjis:
I think if we have phpstan-x.x.neon, we can run PHPStan for all supported PHP versions.
My question is: is it worth  doing?

MGatner:
> Is it worth doing?

What other language versioning safety nets do we have? Obviously if a test case tries to use a deprecated or not-yet-available feature that will fail, but that requires test coverage.

I work regularly across 7.4, 8.0, and 8.1 these days and frequently have to look things up. Maybe other people have their heads around that better, but I don't necessarily trust myself to catch a mistake during a code review.

kenjis:
Okay, it is true that the current code coverage is a bit low.
I'm not sure PHPStan helps us so much, but if you say so, let's try to run in the different PHP versions.

MGatner (approved):
Let's proceed with this for now. The version of PHP running shouldn't affect the version of analysis.

PR #6671, opened 2022-10-12
refactor: valid_date does not work in PHP 8.2

kenjis:
**Description**
See #6170
See https://github.com/codeigniter4/CodeIgniter4/pull/6172#issuecomment-1275496541

```console
bash-3.2$ ./phpunit tests/system/Validation/FormatRulesTest.php

Runtime:       PHP 8.2.0-dev
Configuration: /Users/kenji/work/codeigniter/official/CodeIgniter4/phpunit.xml

...............................................................  63 / 193 ( 32%)
............................................................... 126 / 193 ( 65%)
............................................................... 189 / 193 ( 97%)
....                                                            193 / 193 (100%)

Time: 00:00.325, Memory: 14.00 MB

OK (193 tests, 193 assertions)
```

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/FormatRules.php
+++ b/system/Validation/FormatRules.php
@@ -344,6 +344,15 @@ class FormatRules
         $date   = DateTime::createFromFormat($format, $str);
         $errors = DateTime::getLastErrors();
 
-        return $date !== false && $errors !== false && $errors['warning_count'] === 0 && $errors['error_count'] === 0;
+        if ($date === false) {
+            return false;
+        }
+
+        // PHP 8.2 or later.
+        if ($errors === false) {
+            return true;
+        }
+
+        return $errors['warning_count'] === 0 && $errors['error_count'] === 0;
     }
 }


system/Validation/FormatRules.php
```diff
@@ -344,6 +344,11 @@ public function valid_date(?string $str = null, ?string $format = null): bool
         $date   = DateTime::createFromFormat($format, $str);
         $errors = DateTime::getLastErrors();
 
+        // PHP 8.2 or later.
+        if ($date !== false && $errors === false) {
```
samsonasik, line 348: `$date === false` or `! $date instanceof DateTime` should can be returned early

paulbalandan:
Is there a case where `$date` is a `DateTime` instance yet `$errors` contains actual errors? If no, then can we remove the `getLastErrors` check?

system/Validation/FormatRules.php
kenjis: Done!

kenjis:
> Is there a case where $date is a DateTime instance yet $errors contains actual errors?

Yes.
```php
            [
                '1991-02-29',
                'Y-m-d',
                false,
            ],
            [
                'Thu, 31 Jun 2013 13:31:00',
                'D, d M Y H:i:s',
                false,
            ],
            [
                '2, 17, 2001',
                'j, n, Y',
                false,
            ],
            [
                '09-42-25, 12-00-17',
                'h-i-s, j-m-y',
                false,
            ],
```

paulbalandan (approved):

samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #6672, opened 2022-10-12
refactor: DOMParser::withString() for PHP 8.2

kenjis:
**Description**
See #6170
See https://php.watch/versions/8.2/mbstring-qprint-base64-uuencode-html-entities-deprecated

- `mb_convert_encoding($str, 'HTML-Entities')` is deprecated in PHP 8.2.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/DOMParser.php
+++ b/system/Test/DOMParser.php
@@ -61,9 +61,8 @@ class DOMParser
      */
     public function withString(string $content)
     {
-        // converts all special characters to HTML-Entities
-        $content = htmlentities($content);
-        $content = htmlspecialchars_decode($content);
+        // encode character to HTML numeric string reference
+        $content = mb_encode_numericentity($content, [0x80, 0x10FFFF, 0, 0x1FFFFF], 'UTF-8');
 
         // turning off some errors
         libxml_use_internal_errors(true);
--- a/system/Test/DOMParser.php
+++ b/system/Test/DOMParser.php
@@ -61,7 +61,10 @@ class DOMParser
      */
     public function withString(string $content)
     {
-        // encode character to HTML numeric string reference
+        // DOMDocument::loadHTML() will treat your string as being in ISO-8859-1
+        // (the HTTP/1.1 default character set) unless you tell it otherwise.
+        // https://stackoverflow.com/a/8218649
+        // So encode characters to HTML numeric string references.
         $content = mb_encode_numericentity($content, [0x80, 0x10FFFF, 0, 0x1FFFFF], 'UTF-8');
 
         // turning off some errors
--- a/system/Test/DOMParser.php
+++ b/system/Test/DOMParser.php
@@ -89,7 +89,7 @@ class DOMParser
      * Loads the contents of a file as a string
      * so that we can work with it.
      *
-     * @return DOMParser
+     * @return $this
      */
     public function withFile(string $path)
     {
@@ -184,7 +184,7 @@ class DOMParser
     /**
      * Search the DOM using an XPath expression.
      *
-     * @return DOMNodeList
+     * @return DOMNodeList|false
      */
     protected function doXPath(?string $search, string $element, array $paths = [])
     {
--- a/system/Test/DOMParser.php
+++ b/system/Test/DOMParser.php
@@ -61,8 +61,9 @@ class DOMParser
      */
     public function withString(string $content)
     {
-        // converts all special characters to utf-8
-        $content = mb_convert_encoding($content, 'HTML-ENTITIES', 'UTF-8');
+        // converts all special characters to HTML-Entities
+        $content = htmlentities($content);
+        $content = htmlspecialchars_decode($content);
 
         // turning off some errors
         libxml_use_internal_errors(true);
--- a/tests/system/Test/DOMParserTest.php
+++ b/tests/system/Test/DOMParserTest.php
@@ -80,8 +80,9 @@ final class DOMParserTest extends CIUnitTestCase
     public function provideText()
     {
         return [
-            ['Hello World'],
-            ['Hellö Wörld'],
+            'en' => ['Hello World'],
+            'sv' => ['Hej, världen'],
+            'ja' => ['こんにちは、世界'],
         ];
     }
 


MGatner:
Compare: https://github.com/Kristories/symfony/commit/eea3b10327c1a9b1298a069ea8f844c70e73b6a7

kenjis:
The latest code in Symfony is different.

```php
$this->line = htmlspecialchars_decode(utf8_decode(htmlentities($this->line, ENT_COMPAT, 'UTF-8', false)));
```
https://github.com/Kristories/symfony/blob/eea3b10327c1a9b1298a069ea8f844c70e73b6a7/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L981

```php
$this->line = mb_encode_numericentity($this->line, [0x80, 0x10FFFF, 0, 0x1FFFFF], 'UTF-8');
```
https://github.com/symfony/symfony/blob/b527899f4fb4355fd356eef6f910ea67c5f86192/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L939

system/Test/DOMParser.php
```diff
@@ -61,8 +61,9 @@ public function getBody(): string
      */
     public function withString(string $content)
     {
-        // converts all special characters to utf-8
-        $content = mb_convert_encoding($content, 'HTML-ENTITIES', 'UTF-8');
```
kenjis, line 65: @MGatner To begin with, why do we need this convertion?
The comment is wrong. It never converts to utf-8.

```php
$str_utf8 = '<s>&lt;div&gt;あいう&lt;/div&gt;</s>';
$str_sjis = mb_convert_encoding($str_utf8, 'SJIS-win', 'utf8');

$encoded = mb_convert_encoding($str_utf8, 'HTML-ENTITIES', 'UTF-8');
echo $encoded . PHP_EOL;

$encoded = mb_convert_encoding($str_sjis, 'HTML-ENTITIES', 'UTF-8');
echo $encoded . PHP_EOL;
```
Results:
```
<s>&lt;div&gt;&#12354;&#12356;&#12358;&lt;/div&gt;</s>
<s>&lt;div&gt;&#4294967295;&#4294967295;&#4294967295;&#4294967295;&#4294967295;&#4294967295;&lt;/div&gt;</s>
```

kenjis:
> I'm having a hard time finding a clear answer: are you sure these are the same?

No. The following tests fail with this PR.

```diff
--- a/tests/system/Test/DOMParserTest.php
+++ b/tests/system/Test/DOMParserTest.php
@@ -80,8 +80,9 @@ final class DOMParserTest extends CIUnitTestCase
     public function provideText()
     {
         return [
-            ['Hello World'],
-            ['Hellö Wörld'],
+            'en' => ['Hello World'],
+            'sv' => ['Hej, världen'],
+            'ja' => ['こんにちは、世界'],
         ];
     }
 
```

system/Test/DOMParser.php
kenjis: If I remove the line, tests in DOMParserTest fail.

kenjis:
I changed the code, and test case.

system/Test/DOMParser.php
MGatner: I have no idea why this is here. Seems very strange to be part of our test DOM parser.

MGatner:
Have you `blame`d it for any clues? I can look into it later on desktop, but it is a rather baffling feature.

system/Test/DOMParser.php
kenjis: It seems DOMDocument does not process Japanese correctly.

```php
        $html = 'こんにちは、世界';

        $dom = new DOMDocument('1.0', 'utf-8');
        $dom->loadHTML($html);
        dd($dom->saveHTML($dom->documentElement));
```
Screenshot 2022-10-24 17 48 26

kenjis: This works.

```php
        $html = 'こんにちは、世界';

        $dom = new DOMDocument('1.0', 'utf-8');
        $html = mb_encode_numericentity($html, [0x80, 0x10FFFF, 0, 0x1FFFFF], 'UTF-8');
        $dom->loadHTML($html);
        var_dump($dom->saveHTML($dom->documentElement));
```
```
string(57) "<html><body><p>こんにちは、世界</p></body></html>"
```

kenjis: Maybe non-ASCII characters are not handled correctly, and we need to convert all of them to numeric character references.

MGatner: Have you tried setting the object encoding? https://stackoverflow.com/questions/6573258/domdocument-and-special-characters#6620546

kenjis: The `encoding` is specified in the second parameter of the constructor.

```php
$dom = new DOMDocument('1.0', 'utf-8');
```
And the stackoverflow sample code uses `utf8_decode()` that is also deprecated in 8.2.

kenjis: > DOMDocument::loadHTML will treat your string as being in ISO-8859-1 (the HTTP/1.1 default character set) unless you tell it otherwise. 
https://stackoverflow.com/questions/8218230/php-domdocument-loadhtml-not-encoding-utf-8-correctly/8218649#8218649

kenjis: I think `mb_encode_numericentity($html, [0x80, 0x10FFFF, 0, 0x1FFFFF], 'UTF-8')` is the best for this case.
Prepending `<?xml encoding="utf-8" ?>` changes the output.
```diff
--- Expected
+++ Actual
@@ @@
 '<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
-<html><body><div><h1>Hello</h1></div></body></html>
+<?xml encoding="utf-8" ?><html><body><div><h1>Hello</h1></div></body></html>
 '
```

kenjis:

system/Test/DOMParser.php
MGatner: We could always pretend the XML tag and then cut it off again, but let's go with your solution since it seems cleaner.

MGatner (approved):

paulbalandan (approved):

kenjis:
Thanks for the reviews!

PR #6681, opened 2022-10-13
fix: support for assigning extra data for the view() method in controlled cells

michalsn:
**Description**
Follow-up #6620

This is related to the Controlled Cells.

According to the user guide, we should be able to assign extra variables to the `view()` method. But this function doesn't seem to be implemented. This PR fixes it.

Ref: https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/outgoing/view_cells.rst#customize-the-rendering

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/Cells/Cell.php
+++ b/system/View/Cells/Cell.php
@@ -65,10 +65,11 @@ class Cell
      * current scope and captures the output buffer instead of
      * relying on the view service.
      */
-    final protected function view(?string $view): string
+    final protected function view(?string $view, array $data = []): string
     {
         $properties = $this->getPublicProperties();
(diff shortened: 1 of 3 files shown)


lonnieezell (approved):
Thanks for adding that!

MGatner (commented):
Love the Lost reference 😍
This looks good and logical to me. Let's see if @lonnieezell is around to confirm his intention.

MGatner:
🤦‍♂️ Always read the comments first. I'm so bad at it!

kenjis (approved):

PR #6682, opened 2022-10-14
test: remove $request->uri and fix incorrect assertions

kenjis:
**Description**
See #5344

- `IncomingRequest::$uri` will be protected.
- `$request->getUri()->getPath()` does not start with `/`. `IncomingRequest::$uri` is a bit different from a normal URI object.
https://github.com/codeigniter4/CodeIgniter4/blob/a3eedf4055c1a2895051b654c49311641dd2975f/system/HTTP/IncomingRequest.php#L60-L71

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/HTTP/URITest.php
+++ b/tests/system/HTTP/URITest.php
@@ -863,56 +863,56 @@ final class URITest extends CIUnitTestCase
 
     public function testBasedNoIndex()
     {
-        $this->resetServices();
-
         $_SERVER['HTTP_HOST']   = 'example.com';
         $_SERVER['REQUEST_URI'] = '/ci/v4/controller/method';
 
-        $config            = new App();
-        $config->baseURL   = 'http://example.com/ci/v4';
-        $config->indexPage = 'index.php';
-        $request           = Services::request($config);
-        $request->uri      = new URI('http://example.com/ci/v4/controller/method');
+        $this->resetServices();
 
+        $config            = new App();
+        $config->baseURL   = 'http://example.com/ci/v4/';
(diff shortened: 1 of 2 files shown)


tests/system/HTTP/URITest.php
```diff
@@ -863,56 +863,56 @@ public function testSetBadSegmentSilent()
...
+        $config->baseURL   = 'http://example.com/ci/v4/';
+        $config->indexPage = '';
+        Factories::injectMock('config', 'App', $config);
 
+        $request = Services::request($config);
         Services::injectMock('request', $request);
 
         // going through request
         $this->assertSame('http://example.com/ci/v4/controller/method', (string) $request->getUri());
-        $this->assertSame('/ci/v4/controller/method', $request->getUri()->getPath());
```
MGatner, line 881: What causes this? Moving the call to `resetServices()`?

kenjis: No.

`$request->getUri()->getPath()` does not start with `/`. `IncomingRequest::$uri` is a bit different from a normal URI object. The path is changed in the constructor.
https://github.com/codeigniter4/CodeIgniter4/blob/a3eedf4055c1a2895051b654c49311641dd2975f/system/HTTP/IncomingRequest.php#L60-L71

MGatner: Ha yes, that was my note. It's all coming back now.

MGatner (approved):

PR #6684, opened 2022-10-14
Bug: DateHelperTest::testNowSpecific depends on time

### LINKED ISSUE #5492: Bug: `DateHelperTest::testNowSpecific` depends on time
```
There was 1 failure:

1) CodeIgniter\Helpers\DateHelperTest::testNowSpecific
Failed asserting that 4 is equal to 1 or is less than 1.

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Test/CIUnitTestCase.php:461
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Helpers/DateHelperTest.php:35
```
https://github.com/codeigniter4/CodeIgniter4/runs/4593459157?check_suite_focus=true

fpoy:
Fixes #5492

**Description**

In the `testNowSpecific()` method, there are two different `now()` calls

There may be sometimes a latency between the two calls, and by default `assertCloseEnough` test if the difference is lower than 1 second.

To reproduce, just add a delay `sleep(n)` inside the `now()` function

A solution may be to increase the tolerance threshold to an acceptable level (eg 10 sec for 2 hours should be sufficient)

**Checklist:**

- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Helpers/DateHelperTest.php
+++ b/tests/system/Helpers/DateHelperTest.php
@@ -32,8 +32,8 @@ final class DateHelperTest extends CIUnitTestCase
 
     public function testNowSpecific()
     {
-        // Chicago should be two hours ahead of Vancouver
-        $this->assertCloseEnough(7200, now('America/Chicago') - now('America/Vancouver'));
+        // Chicago should be two hours ahead of Vancouver (+- 10 sec)
+        $this->assertCloseEnough(7200, now('America/Chicago') - now('America/Vancouver'), '', 10);
     }
 
     public function testTimezoneSelectDefault()


kenjis:
The true solution is to stop time during testing.
But for the time being this PR may be okay.

To be hoest, I would like to deprecate the `now()` function,
or at least to deprecate the parameter `$timezone`.
Because I don't understand Unix timestamp with timezone.

tests/system/Helpers/DateHelperTest.php
```diff
@@ -32,8 +32,8 @@ public function testNowDefault()
 
     public function testNowSpecific()
     {
-        // Chicago should be two hours ahead of Vancouver
-        $this->assertCloseEnough(7200, now('America/Chicago') - now('America/Vancouver'));
+        // Chicago should be two hours ahead of Vancouver (+- 10 sec)
+        $this->assertCloseEnough(7200, now('America/Chicago') - now('America/Vancouver'), '', 10);
```
kenjis, line 36: After all, this test is strange.
It is difficult to accept the *specification* that a 10-second error is not a problem.

kenjis (changes requested):
Sorry, I don't accept this.
Because test code is a specification.

MGatner:
Is the test actually failing? 1 second should be ample time to cover any single test case (especially that doesn't involve database or HTTP).

kenjis:
```
1) CodeIgniter\Helpers\DateHelperTest::testNowSpecific
Failed asserting that 4 is equal to 1 or is less than 1.

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Test/CIUnitTestCase.php:478
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Helpers/DateHelperTest.php:36
```
https://github.com/codeigniter4/CodeIgniter4/actions/runs/3317017866/jobs/5479434095

kenjis:
I sent another PR #6753

kenjis:
Closed by #6753

PR #6689, opened 2022-10-14
insertBatch updateBatch upsertBatch deleteBatch from query

sclubricants:
This PR adds the final piece of some changes to BaseBuilder. 

This adds the ability to insert, update, upsert from a query rather than a set of data.

Much of the required pieces of this PR were already in place.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


sclubricants:
@codeigniter4/database-team This is the last major piece for now. Please review.

sclubricants:
Perhaps it might be better to not pass a query through the $set parameter of th *Batch() methods and rely on using the fromQuery() method only. The $set parameter is really the same as in setData().

sclubricants:
It should be noted that this design works on the assumption that the columns in the query are aliased to match those of the target table. While SQL usually works based on position and doesn't require the column names to match it is usually good practice as queries are much more readable. It also makes it much simpler to handle an upsert scenario where the target and source share the same name.. otherwise the developer would have to map them separately.

sclubricants:
Example: 

This is valid SQL:

```SQL
INSERT INTO pageview (user_id, user_name, last_view)
SELECT id, name, created_date
FROM user
ON DUPLICATE KEY UPDATE
SET user_id = VALUES(id),
user_name = VALUES(name),
last_view = VALUES(created_date)
```
The problem with this is that the fields would have to be mapped: `user_id => id`

This PR will assume this is done like:

```SQL
INSERT INTO pageview (user_id, user_name, last_view)
SELECT id AS user_id, name AS user_name, created_date AS last_view
FROM user
ON DUPLICATE KEY UPDATE
SET user_id = VALUES(user_id ),
user_name = VALUES(user_name ),
last_view = VALUES(last_view )
```
This way the mapping is done in SQL and allows to easily generate the SQL statements. The code will parse the field names from the select statement and use them to generate the other parts of the statement.

kenjis:
Sorry, we reorganized the changelog v4.3.0.
Please rebase.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -177,6 +177,7 @@ Others
 - Improved data returned by :ref:`BaseConnection::getForeignKeyData() <metadata-getforeignkeydata>`. All DBMS returns the same structure.
 - SQLite :ref:`BaseConnection::getIndexData() <db-metadata-getindexdata>` now can return pseudo index named ``PRIMARY`` for `AUTOINCREMENT` column, and each returned index data has ``type`` property.
 - ``BasePreparedQuery::close()`` now deallocates the prepared statement in all DBMS. Previously, they were not deallocated in Postgre, SQLSRV and OCI8. See :ref:`database-queries-stmt-close`.
+- Added ``BaseBuilder::fromQuery()`` which allows insert, update, upsert from a query. See :ref:`insert-batch-data`.
```
kenjis, line 180: Please move this up to `Query Builder` section.

system/Database/BaseBuilder.php
```diff
@@ -2051,6 +2067,72 @@ public function onConstraint($set)
...
     }
 
+    /**
+     * Sets data source as a query for insert/update/upsert
+     *
+     * @param BaseBuilder|RawSql|string $query
+     */
+    public function fromQuery($query): BaseBuilder
+    {
+        if (! empty($query)) {
```
kenjis, line 2077: Why skip if `$query = ''`?

system/Database/BaseBuilder.php
```diff
@@ -2399,15 +2497,31 @@ protected function validateUpdate(): bool
     /**
      * Sets data and calls batchExecute to run queryies
      *
-     * @param array|object|null        $set         a dataset or select query
-     * @param array|RawSql|string|null $constraints
+     * @param array|BaseBuilder|object|RawSqwl|null $set         a dataset or select query
```
kenjis, line 2500: Not `RawSqwl` but `RawSql`.

system/Database/BaseBuilder.php
```diff
@@ -2399,15 +2497,31 @@ protected function validateUpdate(): bool
     /**
      * Sets data and calls batchExecute to run queryies
```
kenjis, line 2498:
```suggestion
     * Sets data and calls batchExecute to run queries
```

system/Database/BaseBuilder.php
```diff
@@ -2051,6 +2067,72 @@ public function onConstraint($set)
         return $this;
     }
 
+    /**
+     * Sets data source as a query for insert/update/upsert
+     *
+     * @param BaseBuilder|RawSql|string $query
```
kenjis, line 2073: QueryBuilder does not accept raw SQL query.
Why don't we drop `string`?

tests/system/Database/Live/InsertTest.php
```diff
@@ -24,6 +26,11 @@ final class InsertTest extends CIUnitTestCase
 {
     use DatabaseTestTrait;
 
+    /**
+     * @var Forge|mixed
```
kenjis, line 30: Why mixed?

tests/system/Database/Live/UpdateTest.php
```diff
@@ -27,6 +28,11 @@ final class UpdateTest extends CIUnitTestCase
 {
     use DatabaseTestTrait;
 
+    /**
+     * @var Forge|mixed
```
kenjis, line 32: Why mixed?

tests/system/Database/Live/UpsertTest.php
```diff
@@ -27,6 +29,11 @@ final class UpsertTest extends CIUnitTestCase
 {
     use DatabaseTestTrait;
 
+    /**
+     * @var Forge|mixed
```
kenjis, line 33: Why mixed?

user_guide_src/source/database/query_builder.rst
```diff
@@ -1358,6 +1378,14 @@ Class Reference
 
         Specifies the ``FROM`` clause of a query using a subquery.
 
+    .. php:method:: fromQuery($query)
+
+        :param mixed $query: Instance of the BaseBuilder or RawSql class or string
```
kenjis, line 1383: Do not use `mixed`. `mixed` means any type.

system/Database/BaseBuilder.php
```diff
@@ -2051,6 +2067,72 @@ public function onConstraint($set)
...
+     */
+    public function fromQuery($query): BaseBuilder
```
kenjis, line 2075: Do we really need this public method?

system/Database/BaseBuilder.php
```diff
@@ -2051,6 +2067,72 @@ public function onConstraint($set)
...
+        return $this;
+    }
+
+    /**
+     * Gets column names from a select query
+     */
+    protected function fieldsFromQuery(string $sql): array
+    {
+        $sql = preg_replace('/\\(([^()]*+|(?R))*\\)/', '', $sql); // remove everything in parenthesis - removes "FROM" and commas
+        $d   = ['`', "'", '"'];
```
kenjis, line 2100: What is d?

system/Database/BaseBuilder.php
```diff
@@ -2051,6 +2067,72 @@ public function onConstraint($set)
...
+
+        $o   = ' ';
```
kenjis, line 2102: What is o?

system/Database/BaseBuilder.php
```diff
@@ -2051,6 +2067,72 @@ public function onConstraint($set)
...
+        $r   = '$';
```
kenjis, line 2103: What is r?

system/Database/BaseBuilder.php
```diff
@@ -2051,6 +2067,72 @@ public function onConstraint($set)
...
+            }
+        }
+
+        return $this;
+    }
+
+    /**
+     * Gets column names from a select query
+     */
+    protected function fieldsFromQuery(string $sql): array
```
kenjis, line 2097: Parsing SQL with regular expressions is not a good way, but are you sure this way is not a problem?

Could you move this method to a separate class and write tests for it?

kenjis:
It seems we should not support the following:
```
$query = 'SELECT user2.name, user2.email, user2.country
          FROM user2
          LEFT JOIN user ON user.email = user2.email
          WHERE user.email IS NULL';
$sql = $builder->ignore(true)->fromQuery($query)->insertBatch();
```
Because QueryBuilder is what builds queries.
And it requires to parse SQL statement, but it is not good way.

sclubricants:
I have found that QueryBuilder doesn't produce reliable column names either if you use aliases and other complicated sql. A lot of the queries I use in this situation are very complicated and are difficult to replicate in QueryBuilder. 

I challenge you to break my query parser. You can use subqueries, aliases, even multi word column names with any escape character. The one thing it wont accept is the wildcard *.

kenjis:
Even if you need to pass a raw query to QueryBuilder, it is sufficient that we can pass `RawSql`.

QueryBuilder is database abstraction. We should not accept parameters that cannot run on other DBMS.
But the exceptions are `RawSql` (and `$escape = false`).

> I have found that QueryBuilder doesn't produce reliable column names either if you use aliases and other complicated sql. A lot of the queries I use in this situation are very complicated and are difficult to replicate in QueryBuilder.

Can you show a few example?

system/Database/BaseBuilder.php
sclubricants: I guess because it would set the variables and I preferred not to. It would make a call to protectIdentifiers() as well.

```php
if (is_string($query)) {
    $this->QBOptions['fromQuery'] = $query;
    $this->QBKeys                 = $this->db->protectIdentifiers($this->fieldsFromQuery($query));
    $this->QBSet                  = [];
}
```

system/Database/BaseBuilder.php
sclubricants: I added RawSql as a just in case scenario. 

Everything gets converted to a string so why not accept a string?

tests/system/Database/Live/InsertTest.php
sclubricants: I don't know how that got there.

system/Database/BaseBuilder.php
sclubricants: I prefer to call the method directly rather than pass it in the *Batch function. I think it is easier to read the code when you do it this way. 

```php
$this->db->table('user')->insertBatch($query);

$this->db->table('user')->fromQuery($query)->insertBatch();

$this->db->table('user')->setData($data)->insertBatch();
```

I also debated weather or not I should accept queries in the $set parameter of the *Batch() methods. $set was really intended as the parameter for setData(). But it works passing query in *Batch() methods.

At some point It might be useful for other methods as well. We could always change it to public later I suppose though.

system/Database/BaseBuilder.php
sclubricants: delimeters. These are the escape characters. This code is pulling out what is between escape characters and replacing things like spaces.

I used one letter variables to keep the code small.

system/Database/BaseBuilder.php
sclubricants: I can't remember what o was suppose to stand for..

system/Database/BaseBuilder.php
sclubricants: the Replace character.

sclubricants: This preserves things like: SELECT 2.99 AS \`total profit\`

system/Database/BaseBuilder.php
sclubricants: I'm not sure why it should be in another class. This method could potentially be useful for other things within this class.

I can write a test for it. I wish I still had the one I used when I built the method. It had every conceivable thing in it.

sclubricants: Or maybe you were meaning a separate test class?

kenjis: > I'm not sure why it should be in another class.

Because this BaseBuilder class is too big. It has more than 3,000 lines!
and the Class Complexity is 348 !!!

We should follow SOILD principles more.

`4.3` branch:
```console
$ phploc system/Database/BaseBuilder.php 
phploc 7.0.2 by Sebastian Bergmann.

Size
  Lines of Code (LOC)                             3212
  Comment Lines of Code (CLOC)                    1111 (34.59%)
  Non-Comment Lines of Code (NCLOC)               2101 (65.41%)
  Logical Lines of Code (LLOC)                     588 (18.31%)
    Classes                                        588 (100.00%)
      Average Class Length                         588
        Minimum Class Length                       588
        Maximum Class Length                       588
      Average Method Length                          4
        Minimum Method Length                        1
        Maximum Method Length                       28
      Average Methods Per Class                    129
        Minimum Methods Per Class                  129
        Maximum Methods Per Class                  129
    Functions                                        0 (0.00%)
      Average Function Length                        0
    Not in classes or functions                      0 (0.00%)

Cyclomatic Complexity
  Average Complexity per LLOC                     0.59
  Average Complexity per Class                  348.00
    Minimum Class Complexity                    348.00
    Maximum Class Complexity                    348.00
  Average Complexity per Method                   3.69
    Minimum Method Complexity                     1.00
    Maximum Method Complexity                    19.00
```

kenjis: If this continues, this class will only get longer. It is my hope that this class will not grow any larger. Because it is difficult for humans to read code that is too long.

Aside from that, if you are going to adopt the way of parsing SQL statements, please add tests for just this method.

sclubricants:
>QueryBuilder is database abstraction. We should not accept parameters that cannot run on other DBMS.

I don't follow what your trying to say. The method excepts a string of sql. this string can be for any DBMS. The RawSql gets converted to a string so why not just accept a string?

sclubricants:
>Can you show a few example?


```SQL
SELECT
				pds_promast.O_EDSCTPRDCTDSCO AS GROUPREB,
				pds_promast.VENDORCODE0NONE AS VEND,
				LEFT(pds_promast.REBATENUMBER,length(pds_promast.REBATENUMBER)-2) AS REBATE,
				pds_promast.DESCRIPTION AS REBATE_DESCRIPTION,
				pds_promast.REFERENCENUMBER AS AcctNo,
				pds_procprod.PRODUCTCODE AS PRODCODE,
				pds_procprod.PACKCODE AS PK,
				GROUP_CONCAT(DISTINCT IF(pds_procprod.SUBMITPROD='',' [NONE] ',CONCAT(pds_procprod.SUBMITPROD,'-',pds_procprod.SUBMITPACK)) SEPARATOR' | ') AS `SUBPRODPK`,
				pds_prodhead.PRODUCTDESCRIPTION AS PRODNAME,
				GROUP_CONCAT(DISTINCT pds_proprc.PRICETABLENUMBER ORDER BY pds_proprc.PRICETABLENUMBER SEPARATOR' , ') AS PT,
				pds_proprc.DESCRIPTION AS PT_DESC,
				IF(t2.AccountNumber IS NULL,0,1) AS PT_ACTIVE,
				GROUP_CONCAT(DISTINCT IF(pds_prodpack.VENDORPRODCODE='',NULL,pds_prodpack.VENDORPRODCODE) SEPARATOR' | ') AS SAP,
				IF(t1.AccountNumber IS NULL,0,1) AS SAP_ACTIVE,
				GROUP_CONCAT(DISTINCT FORMAT(pds_proprc.DISCOUNTRATELVL0,6) ORDER BY pds_proprc.DISCOUNTRATELVL0 SEPARATOR' | ') AS RATE,
				#GROUP_CONCAT(DISTINCT t1.RATE SEPARATOR' | ') AS CHV_RATE,
				GROUP_CONCAT(DISTINCT FORMAT(IF(t1.CONVFACTOR >= 0.56 AND t1.CONVFACTOR <= 0.939,(t1.CONVFACTOR*t1.RATE)/6,t1.RATE),6) SEPARATOR' | ') AS CHV_RATE,
				#GROUP_CONCAT(DISTINCT IF(pds_procprod.PRODUCTCODE LIKE 'TECH%',ROUND((t1.RATE*t1.CONVFACTOR)/6,6),t1.RATE) SEPARATOR' | ') AS CHV_RATE,
								#IF(GROUP_CONCAT(DISTINCT FORMAT(pds_proprc.DISCOUNTRATELVL0,4) ORDER BY pds_proprc.DISCOUNTRATELVL0)=GROUP_CONCAT(DISTINCT IF(pds_procprod.PRODUCTCODE LIKE 'TECH%',ROUND((t1.RATE*t1.CONVFACTOR)/6,4),t1.RATE) SEPARATOR' | '),'','{x}') AS "MATCH",
				#IF(IF(GROUP_CONCAT(DISTINCT FORMAT(pds_proprc.DISCOUNTRATELVL0,4) ORDER BY pds_proprc.DISCOUNTRATELVL0)=GROUP_CONCAT(DISTINCT IF(pds_procprod.PRODUCTCODE LIKE 'TECH%',ROUND((t1.RATE*t1.CONVFACTOR)/6,4),t1.RATE) SEPARATOR' | '),'','{x}') = '{x}' 
								#OR COUNT(DISTINCT IF(pds_procprod.SUBMITPROD='','',CONCAT(pds_procprod.SUBMITPROD,'-',pds_procprod.SUBMITPACK))) > 1
								#OR COUNT(DISTINCT pds_proprc.PRICETABLENUMBER) != 1
								#OR COUNT(DISTINCT FORMAT(pds_proprc.DISCOUNTRATELVL0,4)) != 1
								#OR COUNT(DISTINCT t1.RATE) != 1
								#OR COUNT(DISTINCT pds_prodpack.VENDORPRODCODE) != 1
								#OR MID(pds_proprc.DESCRIPTION,1,9) != GROUP_CONCAT(DISTINCT IF(pds_prodpack.VENDORPRODCODE='',NULL,pds_prodpack.VENDORPRODCODE) SEPARATOR' | '),1,0) AS ERROR,
								IF(GROUP_CONCAT(DISTINCT FORMAT(pds_proprc.DISCOUNTRATELVL0,6) ORDER BY pds_proprc.DISCOUNTRATELVL0)=GROUP_CONCAT(DISTINCT FORMAT(IF(t1.CONVFACTOR >= 0.56 AND t1.CONVFACTOR <= 0.939,(t1.CONVFACTOR*t1.RATE)/6,t1.RATE),6) SEPARATOR' | '),'','{x}') AS "MATCH",
								IF(IF(GROUP_CONCAT(DISTINCT FORMAT(pds_proprc.DISCOUNTRATELVL0,6) ORDER BY pds_proprc.DISCOUNTRATELVL0)=GROUP_CONCAT(DISTINCT FORMAT(IF(t1.CONVFACTOR >= 0.56 AND t1.CONVFACTOR <= 0.939,(t1.CONVFACTOR*t1.RATE)/6,t1.RATE),6) SEPARATOR' | '),'','{x}') = '{x}' 
								OR COUNT(DISTINCT IF(pds_procprod.SUBMITPROD='','',CONCAT(pds_procprod.SUBMITPROD,'-',pds_procprod.SUBMITPACK))) > 1
								OR COUNT(DISTINCT pds_proprc.PRICETABLENUMBER) != 1
								OR COUNT(DISTINCT FORMAT(pds_proprc.DISCOUNTRATELVL0,6)) != 1
								OR COUNT(DISTINCT t1.RATE) != 1
								OR COUNT(DISTINCT pds_prodpack.VENDORPRODCODE) != 1
								OR MID(pds_proprc.DESCRIPTION,1,9) != GROUP_CONCAT(DISTINCT IF(pds_prodpack.VENDORPRODCODE='',NULL,pds_prodpack.VENDORPRODCODE) SEPARATOR' | '),1,0) AS ERROR,
				CONCAT('',GROUP_CONCAT(DISTINCT pds_procprod.CUSTOMERCODE ORDER BY pds_procprod.CUSTOMERCODE SEPARATOR',')
				OR t2.AccountNumber IS NULL
				OR t1.AccountNumber IS NULL,'') AS CUSTs
				FROM
				pds.pds_promast
				INNER JOIN pds.pds_procprod ON pds_procprod.PROMONUMBER = pds_promast.REBATENUMBER
				INNER JOIN pds.pds_customer ON pds_customer.CUSTOMERCODE = pds_procprod.CUSTOMERCODE
				INNER JOIN pds.pds_prodhead ON pds_prodhead.PRODUCTCODE = pds_procprod.PRODUCTCODE
				INNER JOIN pds.pds_proprc ON pds_proprc.PROMONUMBER = pds_promast.REBATENUMBER AND pds_proprc.PRICETABLENUMBER = pds_procprod.REBATEPRICETABLE
				LEFT JOIN pds.pds_prodpack ON pds_prodpack.PRODUCTCODE = IF(pds_procprod.SUBMITPROD='',pds_procprod.PRODUCTCODE,pds_procprod.SUBMITPROD) AND pds_prodpack.PACKAGECODE = IF(pds_procprod.SUBMITPACK='',pds_procprod.PACKCODE,pds_procprod.SUBMITPACK)
				LEFT JOIN (SELECT t5.AccountNumber,t5.SAP,t5.RATE,t5.CONVFACTOR,t5.CONTRACT
										FROM(
										SELECT
										ca.AccountNumber,
										#IF(pc.ToSAP IS NULL,cp.PRODUCTCODE,pc.ToSAP) AS SAP,
										cp.PRODUCTCODE AS SAP,
										FORMAT(cli.RATE__C,6) AS RATE,
										cp.ISACTIVE,
										cli.CONVERSION_FACTOR__C AS CONVFACTOR,
										bc.`NAME` AS CONTRACT
										FROM mpower.mpower_account ca
										INNER JOIN mpower.mpower_business_contract bc ON bc.C2_NAME__C = ca.Id AND bc.STATUS__C = 'Active'
										INNER JOIN mpower.mpower_business_contract_line_item cli ON cli.BUSINESS_CONTRACT__C = bc.ID
										INNER JOIN mpower.mpower_product2 cp ON cp.ID = cli.PRODUCT_NAME__C
										#LEFT JOIN mpower.prod_crossover pc ON pc.FromSAP = cp.PRODUCTCODE
										WHERE ca.Active__c = 1 AND bc.STATUS__C = 'Active' AND ca.IsDeleted = 0 
										GROUP BY ca.Id,bc.ID,cli.ID
										ORDER BY #IF(pc.ToSAP IS NULL,0,1),
										SAP, cp.ISACTIVE DESC,cp.PRODUCTCODE) t5
										GROUP BY t5.AccountNumber, t5.SAP
										ORDER BY t5.AccountNumber, t5.SAP,t5.ISACTIVE ASC) t1 ON t1.AccountNumber = pds_promast.REFERENCENUMBER AND t1.SAP = pds_prodpack.VENDORPRODCODE
								
								LEFT JOIN (SELECT t5.AccountNumber,t5.SAP
										FROM(
										SELECT
										ca.AccountNumber,
										cp.PRODUCTCODE AS SAP,
										cp.ISACTIVE
										FROM mpower.mpower_account ca
										INNER JOIN mpower.mpower_business_contract bc ON bc.C2_NAME__C = ca.Id AND bc.STATUS__C = 'Active'
										INNER JOIN mpower.mpower_business_contract_line_item cli ON cli.BUSINESS_CONTRACT__C = bc.ID
										INNER JOIN mpower.mpower_product2 cp ON cp.ID = cli.PRODUCT_NAME__C
										#LEFT JOIN mpower.prod_crossover pc ON pc.FromSAP = cp.PRODUCTCODE
										WHERE ca.Active__c = 1 AND bc.STATUS__C = 'Active' AND ca.IsDeleted = 0 
										GROUP BY ca.Id,bc.ID,cli.ID
										ORDER BY #IF(pc.ToSAP IS NULL,0,1),
										SAP, cp.ISACTIVE DESC,cp.PRODUCTCODE) t5
										GROUP BY t5.AccountNumber, t5.SAP
										ORDER BY t5.AccountNumber, t5.SAP,t5.ISACTIVE ASC) t2 ON t2.AccountNumber = pds_promast.REFERENCENUMBER AND t2.SAP = MID(pds_proprc.DESCRIPTION,1,9)
				
				WHERE
				pds_promast.VENDORCODE0NONE IN (26,27)
				AND IF(IF(CURDATE() >= IFNULL(pds_procprod.STARTDATE,CURDATE()),1,0)+IF(CURDATE() <= IFNULL(pds_procprod.ENDDATE,CURDATE()),1,0)=2,1,0) = 1
				AND IF(IF(CURDATE() >= IFNULL(pds_promast.STARTDATEYYMMDD,CURDATE()),1,0)+IF(CURDATE() <= IFNULL(pds_promast.ENDDATEYYMMDD,CURDATE()),1,0)=2,1,0) = 1
				AND IF(IF(CURDATE() >= IFNULL(pds_proprc.STARTDATELEVEL0,CURDATE()),1,0)+IF(CURDATE() <= IFNULL(pds_proprc.ENDDATELEVEL0,CURDATE()),1,0)=2,1,0) = 1
				
				GROUP BY
				pds_promast.REBATENUMBER,
				pds_procprod.PRODUCTCODE,
				pds_procprod.PACKCODE
				HAVING 1=1  AND ERROR = 1
				ORDER BY pds_promast.VENDORCODE0NONE,pds_promast.REBATENUMBER,pds_procprod.PRODUCTCODE,pds_procprod.PACKCODE
```

system/Database/BaseBuilder.php
sclubricants: Well if we need to break this class down then we should do it in a methodical way and not just random methods. Maybe for 5.0 we can refactor things?

Test has been added for parsing SQL. Feel free to make suggestions of more ways to test.

system/Database/BaseBuilder.php
sclubricants: I remember.. o is for original character

kenjis:
Thank you for the example.
Surely you need to pass raw SQL statements.

system/Database/BaseBuilder.php
kenjis: > Well if we need to break this class down then we should do it in a methodical way and not just random methods. 

Yes.

> Maybe for 5.0 we can refactor things?

Maybe it is better. We need to discuss more with other members.

sclubricants:
I think your confusing QueryBuilder with the code the developer writes. Yes, sometimes you may want your code to be portable but not always. I mean what is the point of the query() method then? Sometimes you need to bend the rules. My method allows you both ways.

I can't remember how exactly but I one came across a special join that wasn't doable with QueryBuilder. I wish I knew where that code is..

When I have long complicated queries I build them in a program called Navicat. I do lots of testing and tweaking until I get what I need. Then when they get so complicated I don't have the time to try and replicate them in QB. My code sits behind a firewall and is mostly only executed by me. Sql injection is not an issue. Sometimes it just saves a lot of time to run raw queries. Other code I maintain is public facing and I use QB. The SQL is much simpler though.

michalsn (changes requested):

system/Database/BaseBuilder.php
sclubricants: `fromSubquery()` works a bit differently. It sets the query to QBFrom[0] . `fromQuery()` sets the query to `QBOptions['fromQuery']` and `QBKeys` to the columns.

One option might be to move the `fromQuery()` logic to `setData()`. This kind of makes sense programming wise.. the query is an alternate to a dataset. The only problem might be is that people might try and use it where it is not supported.

system/Database/BaseBuilder.php
sclubricants: I'll comment each variable. That should suffice.

system/Database/BaseBuilder.php
sclubricants: I removed the string type.

system/Database/BaseBuilder.php
kenjis: The following API seems better if it is possible:

```php
$this->db->table('user')->fromSubquery($query)->insertBatch();
```

> fromSubquery() works a bit differently. It sets the query to QBFrom[0] . fromQuery() sets the query to QBOptions['fromQuery'] and QBKeys to the columns.

It is just internal implementation. We can change the implementation.

system/Database/BaseBuilder.php
```diff
@@ -2061,6 +2079,76 @@ public function onConstraint($set)
...
+    {
+        $sql = preg_replace('/\\(([^()]*+|(?R))*\\)/', '', $sql); // remove everything in parenthesis - removes "FROM" and commas
+        $d   = ['`', "'", '"']; // delimeters - delimits identifiers
```
kenjis, line 2114: Please replace the variable names with meaningful ones.
We don't accept single letter variable name except for loop counter.
> Code should be written to minimize the time it would take for someone else to understand it.

```suggestion
        $identifierDelimeters = ['`', "'", '"'];
```

Comments for what it is are generally not good practice. 
Making the code more readable is better.

kenjis:
> I challenge you to break my query parser. You can use subqueries, aliases, even multi word column names with any escape character. The one thing it wont accept is the wildcard *.

QB may not support the column name:
```php
    public function testColumnWithQuote()
    {
        $builder = new BaseBuilder('test', $this->db);

        // `my"colum` is a column name.
        $sql = <<<'SQL'
                SELECT
                    "my""colum"
                FROM
                    test_table
            SQL;

        $query = $builder->testMode()->fromQuery(new RawSql($sql))->insertBatch();

        $expected = 'INSERT INTO "test" ("my""colum")';

        $this->assertStringContainsString($expected, $query);
    }
```
A bit more realistic:
```php
    public function testWithComment()
    {
        $builder = new BaseBuilder('test', $this->db);

        $sql = <<<'SQL'
                SELECT
                    id -- , test
                FROM
                    test_table
            SQL;

        $query = $builder->testMode()->fromQuery(new RawSql($sql))->insertBatch();

        $expected = 'INSERT INTO "test" ("id")';

        $this->assertStringContainsString($expected, $query);
    }
```

system/Database/BaseBuilder.php
michalsn: Can we also throw an exception when the variable is of type string? I'm honestly concerned that this could be abused. Not a lot of people are reading the docs when something is "just working".

sclubricants:
Ok, I changed fieldsFromQuery() to use the DBMS. I had forgot about comments and making exceptions for this would get much more complicated. I was trying to avoid running the query to get the column names because a lot of the time I do these sorts of operations my queries are very complicated and sometimes they take 20 or 30 seconds to run. Now they will have to run twice, else we could allow the developer to manually specify the columns. I like the ease of use not having to manually specify columns though. I suppose we could make it an option to specify names else get them automatically.

system/Database/BaseBuilder.php
sclubricants: I have added a parameter for manually setting columns now. This differentiates it even farther from fromSubQuery(). I admit it would be nice to have one method but i'm not seeing how everything could play nicely.

I really don't like the implementation of fromSubQuery() - at least as far as the examples in the documentation.

```php
$subquery = $db->table('users')->select('id, name');
$builder  = $db->newQuery()->fromSubquery($subquery, 't');
$query    = $builder->get();
// Produces: SELECT * FROM (SELECT `id`, `name` FROM users) `t`
```
I'd rather see something like:

```php
$queryTable= $db->table('users')->select('id, name');
$queryJoin= $db->table('users2')->select('id, email');

$builder  = $db->table($db->queryAsTable($queryTable, 'alias'))
                        ->select('alias.id, alias.name, alias2.email')
                        ->join($db->queryAsTable($queryJoin, 'alias2'), 'alias2.id = alias.id', 'inner');
$query    = $builder->get();

/*
SELECT alias.id, alias.name , alias2.email
FROM (SELECT id, name FROM users) alias
INNER JOIN (SELECT id, email FROM users2) alias2 ON alias2.id = alias.id
*/

// or perhaps even:
$queryAsTable= $db->table('users')->select('id, name')->queryAsTable('alias'); // returns RawSql ?
$db->table($queryAsTable)->select('alias.id, alias.name')
```

sclubricants:
How about:

```
$builder->setQuery($query)->insertBatch();
```

In this case it is the alternative to `setData()`.

system/Database/BaseBuilder.php
kenjis: https://github.com/codeigniter4/CodeIgniter4/blob/978059c4fac2dbf56e5019b84427de8a7fb01b80/system/Database/BaseBuilder.php#L594-L600
https://github.com/codeigniter4/CodeIgniter4/blob/978059c4fac2dbf56e5019b84427de8a7fb01b80/system/Database/BaseBuilder.php#L2082-L2089

system/Database/BaseBuilder.php
kenjis: @sclubricants Can you throw an exception when the variable is of type string?

system/Database/BaseBuilder.php
kenjis: Is the name `setQuery()` appropriate? Would a more specific name be better?

MGatner:
An immense amount of work! Checking in on the schedule: if we shoot for 4.3 release today or tomorrow does this have any hope of being ready?

kenjis:
I think not.
I am not sure if we can merge this yet. We need input from other members.

system/Database/BaseBuilder.php
sclubricants: Done

system/Database/BaseBuilder.php
sclubricants: setQuery() seems a logical alternative to setData(). Maybe you could do setQueryAsData() but that seems a bit long.

kenjis: If I see `setData()` without any knowledge of the implementation, I would guess it will set data for the query, that is the data to save.

But if I see `setQuery()`, I would wonder what the query is. Why do I set a query to QueryBuilder? QueryBuilder is not what to build a query?

In my opinion, `setQuery()` is not as clear about what it does as `setData()`.

I'm not sure which is better.:
```php
$this->db->table('user')->setQuery($query)->insertBatch();
```
```php
$this->db->table('user')->setQueryAsData($query)->insertBatch();
```
```php
$this->db->table('user')->setDataAsQuery($query)->insertBatch();
```
```php
$this->db->table('user')->fromQuery($query)->insertBatch();
```

sclubricants: >`setDataAsQuery()`

This is what `setData()` actually does. 

`fromQuery()` was ok but it might get confused with `fromSubQuery()`

In the future I'd like to have a method that is something like `queryAsTable()` This would be similar to what `fromSubQuery()` does in that it puts the query in parenthesis and adds an alias. Basically the results of this function could go anywhere a table name could go. This method could be used instead of `fromSubQuery()` by just setting `table()`. It could also go in `join()`.

`setQueryAsData()` is the only other option I see that might make sense.

One thing to keep in mind is that the API language doesn't have to necessarily follow the SQL. I mean for the most part yes it is nice but with some of these more advanced SQL statements they vary with each DBMS. I liked fromQuery() because it seemed clear that the data was coming from the query.

michalsn: Both `fromQuery()` and `setQueryAsData()` seem right to me because they describe well what will happen.

The only disadvantage of `fromQuery()` is that it has a very similar name to an already existing method, and people may be confused about why they're working differently.

For that reason, I would go with `setQueryAsData()`.

sclubricants:
I removed use of setAlias().

What else?

Any chance of getting this or https://github.com/codeigniter4/CodeIgniter4/pull/6734 in 4.3? What can I do to get this moving along?

MGatner:
I have t paid as much attention to these, but aren't they pretty much required for 4.3 because of the previous PRs? I don't want to create a scenario where this work becomes a breaking change because we released something too fast.

kenjis:
I would like to defer this PR to 4.4. I still think it is strange to setQuery to the QueryBuilder.
We need more discussion, more people eyes.

michalsn:
I agree. Looking forwards to more opinions on this.

@MGatner As I recall, the only problem with `setAlias()` was that it is a public method. Kenjis already took care of it in a separate PR: #6871. So there should be no problems.

MGatner:
Sounds good to me. Thanks for the team input. Once 4.3 is settled I can take a proper look as one of those set of 👀

system/Database/BaseBuilder.php
sclubricants: >For that reason, I would go with setQueryAsData().

I agree.  I think this clearly presents itself as an alternative to setData().

I changed everything to this.

@kenjis how do you feel about this?

sclubricants:
I changed `fromQuery()` to `setDataAsQuery()` as discussed.

I also added the use with `deleteBatch()`

Any chance on getting this moving along now?

system/Database/BaseBuilder.php
kenjis: `setQueryAsData()` is much better than `setData()` and `setQuery()`.

sclubricants: Normally with *Batch you use setData(). setQueryAsData() is a fitting alternative. It is self explanatory.

michalsn (approved):
Since `setAlias()` problem is resolved, and I think we all agreed on the name `setQueryAsData()` - I see no other issues here.

This PR doesn't change anything else.

sclubricants:
> Since `setAlias()` problem is resolved, and I think we all agreed on the name `setQueryAsData()` - I see no other issues here.
> 
> This PR doesn't change anything else.

@kenjis anything else? Do you think we could get this into 4.3?

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -211,6 +211,7 @@ Query Builder
 - Added ``deleteBatch()`` method to QueryBuilder. See :ref:`delete-batch`.
 - Added ``when()`` and ``whenNot()`` methods to conditionally add clauses to the query. See :ref:`BaseBuilder::when() <db-builder-when>` for details.
 - Improved the SQL structure for ``Builder::updateBatch()``. See :ref:`update-batch` for the details.
+- Added ``BaseBuilder::setQueryAsData()`` which allows insert, update, upsert, delete from a query. See :ref:`insert-batch-data`.
```
kenjis, line 214:
```suggestion
- Added ``BaseBuilder::setQueryAsData()`` which allows ``insertBatch()``, ``updateBatch()``, ``upsertBatch()``, ``deleteBatch()`` from a query. See :ref:`insert-batch-data`.
```

user_guide_src/source/database/query_builder/115.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+$query = $this->db->table('user2')
+    ->select('user2.name, user2.email, user2.country')
+    ->join('user', 'user.email = user2.email', 'left')
+    ->where('user.email IS NULL');
+
+$additionalUpdateField = ['updated_at' => new RawSql('CURRENT_TIMESTAMP')];
+
+$sql = $builder->onConstraint('email')->updateFields($additionalUpdateField, true)->upsertBatch($query);
```
kenjis, line 10: I'm not comfortable with this code `upsertBatch($query)`.
I feel that the following way is sufficient.

```php
$sql = $builder->setQueryAsData($query)->onConstraint('email')
        ->updateFields($additionalUpdateField, true)->upsertBatch();
```

Do we really need to accept a query with `*batch()` methods?

user_guide_src/source/database/query_builder.rst
```diff
@@ -878,6 +880,12 @@ The first parameter is an associative array of values.
 
 .. warning:: When you use ``RawSql``, you MUST escape the data manually. Failure to do so could result in SQL injections.
 
+You can also insert from a query:
+
+.. literalinclude:: query_builder/117.php
+
+.. note:: It is required to alias the columns of the select query to match those of the target table.
```
kenjis, line 887: Please add a note that `setQueryAsData()` can be used since v4.3.0.
```
.. note:: ``$setQueryAsData()`` can be used since v4.3.0.
```

This is necessary to ensure that users of any version can take advantage of the latest documentation.

Or if you add a section for a new item, you can use the tag:
```
.. versionadded:: 4.3.0
```

user_guide_src/source/database/query_builder.rst
```diff
@@ -938,6 +946,12 @@ The first parameter is an associative array of values.
 
 .. note:: All values are escaped automatically producing safer queries.
 
+You can also upsert from a query:
+
+.. literalinclude:: query_builder/115.php
+
```
kenjis, line 952: Please add a note that `setQueryAsData()` can be used since v4.3.0.

user_guide_src/source/database/query_builder.rst
```diff
@@ -1073,6 +1087,12 @@ The first parameter is an associative array of values, the second parameter is t
 .. note:: ``affectedRows()`` won't give you proper results with this method,
     due to the very nature of how it works. Instead, ``updateBatch()``
     returns the number of rows affected.
+    
+You can also update from a query:
+
+.. literalinclude:: query_builder/116.php
+
```
kenjis, line 1094: Please add a note that `setQueryAsData()` can be used since v4.3.0.

system/Database/BaseBuilder.php
```diff
@@ -2067,6 +2085,61 @@ public function onConstraint($set)
         return $this;
     }
 
+    /**
+     * Sets data source as a query for insert/update/upsert
```
kenjis, line 2089:
```suggestion
     * Sets data source as a query for insertBatch()/updateBatch()/upsertBatch()/deleteBatch()
```

system/Database/BaseBuilder.php
```diff
@@ -2067,6 +2085,61 @@ public function onConstraint($set)
...
+     *
+     * @param BaseBuilder|RawSql $query
+     * @param string             $alias
+     * @param array|string|null  $columns an array or comma delimited string of columns
+     */
+    public function setQueryAsData($query, $alias = null, $columns = null): BaseBuilder
```
kenjis, line 2095:
```suggestion
    public function setQueryAsData($query, ?string $alias = null, $columns = null): BaseBuilder
```

system/Database/BaseBuilder.php
```diff
@@ -2067,6 +2085,61 @@ public function onConstraint($set)
         return $this;
     }
 
+    /**
+     * Sets data source as a query for insert/update/upsert
+     *
+     * @param BaseBuilder|RawSql $query
+     * @param string             $alias
```
kenjis, line 2092:
```suggestion
```

kenjis:
If I specify two queries as follows, how should QueryBuilder behave?

```php
$builder->setQueryAsData($query1)->onConstraint('email')
->updateFields($additionalUpdateField, true)->upsertBatch($query2)
```

michalsn:
> If I specify two queries as follows, how should QueryBuilder behave?
> 
> ```
> $builder->setQueryAsData($query1)->onConstraint('email')
> ->updateFields($additionalUpdateField, true)->upsertBatch($query2)
> ```

I think this is pretty straightforward - this would be the equivalent of:
```
$builder->setQueryAsData($query1)->setQueryAsData($query2)->onConstraint('email')
->updateFields($additionalUpdateField, true)->upsertBatch();
```
So only `$query2` would be used.

The same story is with `onConstraint()`. It also can be set via `upsertBatch()` or `deleteBatch()`.

But... from the other side, ***Batch()** methods would be the only ones that allow this type of data, leading us to inconsistency. From that point of view, I think I agree with @kenjis. `BaseBuilder` and `RawSql` as a parameter should be treated as special cases and allowed to be set via a dedicated method only.

user_guide_src/source/database/query_builder/115.php
sclubricants: I was thinking we might should leave this out too. I went ahead and removed it.

user_guide_src/source/database/query_builder.rst
```diff
@@ -1120,6 +1140,10 @@ This method may be especially useful when deleting data in a table with a compos
 
 .. note:: SQLite does not support the use of ``where()``.
 
+You can also delete from a query:
+
+.. literalinclude:: query_builder/119.php
+
```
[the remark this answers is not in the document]
sclubricants: .. note:: ``$deleteBatch()`` can be used since v4.3.0.

sclubricants:
I think that the singularity of the method's name suggests that it can only be used once. Its not `setQueriesAsData()`.

I have removed the `BaseBuilder` and `RawSql` inputs from `$set` parameter in `*Batch()` methods. I agree that it is a special use case and should not be included. The parameter types should mirror that of `setData()` and not `setQueryAsData()`.

michalsn (approved):
LGTM

kenjis (approved):
LGTM!

MGatner:
🥳🎉🎊🕺

PR #6692, opened 2022-10-15
Fix UpsertTest::testUpsertWithMultipleSet

sclubricants:
SQLite was using GMT time and not timezone set on OS.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


sclubricants:
Now Postgre

kenjis:
It seems SQLite test still fails:
https://github.com/codeigniter4/CodeIgniter4/actions/runs/3253838286/jobs/5341483157

kenjis:
Your fix changed the SQLite time to real my localtime.
But my timezone is not the same with the CI4 default timezone,
so the test fails.

sclubricants:
Should be fixed for good. I statically set the date

sclubricants:
Merge!

kenjis (approved):

PR #6696, opened 2022-10-15
Coveralls Failures

MGatner:
**Description**
The pipeline failures for -7% coverage have been driving me crazy. I'm not sure if this will fit the issue, but I'm going to tinker with it.

**Checklist:**
- [X] Securely signed commits
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -184,15 +184,3 @@ jobs:
           COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           COVERALLS_PARALLEL: true
           COVERALLS_FLAG_NAME: PHP ${{ matrix.php-versions }} - ${{ matrix.db-platforms }}
-
-  coveralls-finish:
-    if: github.repository_owner == 'codeigniter4' && matrix.php-versions == '8.0'
-    needs: [tests]
-    runs-on: ubuntu-20.04
-
-    steps:
-      - name: Coveralls Finished
-        uses: coverallsapp/github-action@master
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          parallel-finished: true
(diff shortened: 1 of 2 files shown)


paulbalandan:
How about using the `phpcov` option?
https://github.com/php-coveralls/php-coveralls#phpcov

paulbalandan:
Although green, this fix is wrong, because the build is "PENDING COMPLETION".
See https://coveralls.io/jobs/107756800

MGatner:
I think it is a bigger issue. Coveralls still has us pegged at "85% coverage", so when the PR results get posted it always sees it as a decrease. I poked around on the site but I'm still unclear what the root problem is.
Screen Shot 2022-10-15 at 1 41 10 PM

MGatner:
> How about using the phpcov option?
I will take a look.

MGatner:
There is some discrepancy between "Builds" (pull requests, 388 total files) and the baselines (`develop`, 377 total files).

kenjis:
> There is some discrepancy between "Builds" (pull requests, 388 total files) and the baselines (develop, 377 total files).

It is the difference between 4.3 and develop.

kenjis:
It seems that *push* has around 85%, but *Pull Request* has around 77% recently.
https://coveralls.io/repos/237520/builds?page=1

But in the past, sometimes *Pull Request* has more than 80% coverage.
https://coveralls.io/repos/237520/builds?page=44

But, the lastest *Pull Request* build also has  84.91% system/ coverage:
https://coveralls.io/builds/53301874

kenjis:
I have no clue.

1. Let's proceed with this for now. The build will be "PENDING COMPLETION", but it is better than now.
2. Switch to another service like https://about.codecov.io/

sclubricants:
If the coveralls for one database driver is running is it possible to exclude the files for all the other drivers? This isn't perfect as some of the implementation is in the base classes but it should help some.

kenjis:
> If the coveralls for one database driver is running is it possible to exclude the files for all the other drivers?

It seems we need to prepare `phpunit.xml` for each DB driver, and use it when running phpunit.

Issue #6573, opened 2022-09-23
Dev: OutgoingRequestInterface is missing


kenjis:
The current CI4 does not have Request interface for an outgoing request.

Problem:
- `CURLRequest` must implement methods for incoming requests that do not have to have.

![](http://www.plantuml.com/plantuml/png/ZP9FIyD04CNl-oc6djMJq6kX5mCLGbflNfOu6uBx9-vCH91-TrVTDYMTb9rJillc_LuUsH3hoAEpQj3jk-uGLlM-kL-lu5j1cjupHgDRX9hY-lKPYPAeEfqgITJWnuZ4Pn7W5wD_emKBIEC38rWq3C789WhoQ0zXv2xqleFuBthukqo8NnyvqnxZ9yR5P09II2PQ2VbwtmQtdE-efhAtrUDrrM95qxM8RgqcEcr2xmQB3ZsVZoHrDiej56tVNTCl3AlyuNFZ_3uLoc60OEPwjTWWc13XwN2ePrLUiBUxgXIRf6N49Bm1V6bsnSH9gqf_uwFQfEVHsHy0)
http://www.plantuml.com/plantuml/uml/ZPBD3i8W48JlFCN6iteoyG5wiOacJTMR5rAtJMCB5T2OcBwxzB-nMuKRy34p31Gykl44OH4SmzLs0My6TgJ2e8fvX11ejRqSK6iB-Tqgew0Gxm_KfeS0NmR5LSz8L0f26eGCOmCofaLgh1fSW7oOHAOY0TL0wthJMjGwPAOZgYUgtyb8osaoMcgSBnMHpEVpbOmLNTcUlvbMJLOqR9FMKSQrxav2cXSPvYYCfiKQkXNgMt0ulPowZmJzzkBVmkryADJ8ekhpYsUp3K8i5UnEfu1wwiRRs_kkjWFwnz62FmpFuTxPSM2PwpSksGS0

```puml
@startuml

package "PSR-7" {
    interface Psr7\MessageInterface

    interface Psr7\RequestInterface extends Psr7\MessageInterface
    note left of Psr7\RequestInterface
        an outgoing request.
    end note

    interface Psr7\ServerRequestInterface extends Psr7\RequestInterface
    note left of Psr7\ServerRequestInterface
        an incoming request.
    end note
}

package "CI4" {
    interface MessageInterface
    interface RequestInterface
    class Message implements MessageInterface
[...]

PR #6698, opened 2022-10-16
feat: add OutgoingRequestInterface

kenjis:
**Description**
See #6573, #4356

- add `OutgoingRequest`
- add `OutgoingRequestInterface`
- `CURLRequest` extends `OutgoingRequest`
- `Request` extends `OutgoingRequest`
- `RequestInterface` extends `OutgoingRequestInterface`

http://www.plantuml.com/plantuml/uml/ZP1TQWCX58NVNSKX2-XJ5n38IuMKbf0kGEoTGUhFLAzGADrxnQOIbCpaRPXpplUf-yGgSdPMB4f_g9cmEyZ77Ru501ZF52Ub2S-KKadb_uykViay1-Fd4trcIjnge2yc_vwszhTsDy6Y0hHLgR5Xt6B9aUTHrze3iPb6oYWVQJsbcrpJrguWlGvkE5JRrAOFJE2m84nzl-Q0yZ2N0F52caB4q_dy-b0x1Levr-x3KTwhRTKjOock2DzEnurtwyrrCwZ3twYfE-MFs9VlxEml

```puml
@startuml
package "CI4" {
    interface MessageInterface
    interface OutgoingRequestInterface extends MessageInterface
    interface RequestInterface extends OutgoingRequestInterface

    class Message implements MessageInterface
    class OutgoingRequest extends Message implements OutgoingRequestInterface
    note top of OutgoingRequest
      an outgoing request.
    end note

    class Request extends OutgoingRequest implements RequestInterface

    class IncomingRequest extends Request
    note top of IncomingRequest
      an incoming request for HTTP.
    end note

    class CLIRequest extends Request
    note top of CLIRequest
      an incoming request for CLI.
    end note

    class CURLRequest extends OutgoingRequest
}
@enduml
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/OutgoingRequest.php
+++ b/system/HTTP/OutgoingRequest.php
@@ -59,9 +59,8 @@ class OutgoingRequest extends Message implements OutgoingRequestInterface
     private function getHostFromUri(URI $uri): string
     {
         $host = $uri->getHost();
-        $host .= $uri->getPort() ? ':' . $uri->getPort() : '';
 
-        return $host;
+        return $host . ($uri->getPort() ? ':' . $uri->getPort() : '');
     }
 
     /**
@@ -157,6 +156,6 @@ class OutgoingRequest extends Message implements OutgoingRequestInterface
             return true;
         }
 
-        return (bool) ($this->header('Host')->getValue() === '');
+        return $this->header('Host')->getValue() === '';
     }
 }
--- a/system/HTTP/OutgoingRequest.php
+++ b/system/HTTP/OutgoingRequest.php
@@ -1,0 +1,162 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\HTTP;
+
+/**
+ * Representation of an outgoing, client-side request.
+ */
+class OutgoingRequest extends Message implements OutgoingRequestInterface
+{
+    /**
+     * Request method.
+     *
+     * @var string
+     */
+    protected $method;
+
+    /**
+     * A URI instance.
+     *
+     * @var URI|null
+     */
+    protected $uri;
+
+    /**
+     * @param string      $method HTTP method
+     * @param string|null $body
+     */
+    public function __construct(
+        string $method,
+        ?URI $uri = null,
+        array $headers = [],
+        $body = null,
+        string $version = '1.1'
+    ) {
+        $this->method = $method;
+        $this->uri    = $uri;
+
+        foreach ($headers as $header => $value) {
+            $this->setHeader($header, $value);
+        }
+
+        $this->body            = $body;
+        $this->protocolVersion = $version;
+
+        if (! $this->hasHeader('Host') && $this->uri->getHost() !== '') {
+            $this->setHeader('Host', $this->getHostFromUri($this->uri));
+        }
+    }
+
+    private function getHostFromUri(URI $uri): string
+    {
+        $host = $uri->getHost();
+        $host .= $uri->getPort() ? ':' . $uri->getPort() : '';
+
+        return $host;
+    }
+
+    /**
+     * Get the request method.
+     *
+     * @param bool $upper Whether to return in upper or lower case.
+     *
+     * @deprecated The $upper functionality will be removed and this will revert to its PSR-7 equivalent
+     */
+    public function getMethod(bool $upper = false): string
+    {
+        return ($upper) ? strtoupper($this->method) : strtolower($this->method);
+    }
+
+    /**
+     * Sets the request method. Used when spoofing the request.
+     *
+     * @return $this
+     *
+     * @deprecated Use withMethod() instead for immutability
+     */
+    public function setMethod(string $method)
+    {
+        $this->method = $method;
+
+        return $this;
+    }
+
+    /**
+     * Returns an instance with the specified method.
+     *
+     * @param string $method
+     *
+     * @return static
+     */
+    public function withMethod($method)
+    {
+        $request         = clone $this;
+        $request->method = $method;
+
+        return $request;
+    }
+
+    /**
+     * Retrieves the URI instance.
+     *
+     * @return URI|null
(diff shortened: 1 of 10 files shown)


system/HTTP/OutgoingRequest.php
```diff
@@ -0,0 +1,161 @@
...
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\HTTP;
+
+/**
+ * Representation of an outgoing, client-side request.
+ */
+class OutgoingRequest extends Message implements OutgoingRequestInterface
```
MGatner, line 17: Why doesn't this extend Request?

system/HTTP/OutgoingRequest.php
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        if ($uri->getHost() !== '') {
+            $request->setHeader('Host', $this->getHostFromUri($uri));
+        }
+
+        return $request;
+    }
+
+    private function isHostHeaderMissingOrEmpty(): bool
```
MGatner, line 153: How about `hasHostHeader()`?

system/HTTP/OutgoingRequestInterface.php
```diff
@@ -0,0 +1,94 @@
...
+namespace CodeIgniter\HTTP;
+
+use InvalidArgumentException;
+
+/**
+ * Representation of an outgoing, client-side request.
+ *
+ * Corresponds to Psr7\RequestInterface.
+ */
+interface OutgoingRequestInterface extends MessageInterface
```
MGatner, line 21: Let's talk about why OutgoingRequest isn't a request first, but I'm also not sure if we need to interface this. It feels pretty bad to create a new interface with a deprecation in it 😢

system/HTTP/Request.php
```diff
@@ -12,11 +12,12 @@
...
 
 use CodeIgniter\Validation\FormatRules;
+use Config\App;
 
 /**
- * Representation of an HTTP request.
+ * Representation of an incoming, server-side HTTP request.
  */
-class Request extends Message implements RequestInterface
+class Request extends OutgoingRequest implements RequestInterface
```
MGatner, line 20: The comment and the extension seem in opposition. Is it incoming or outgoing?

system/HTTP/OutgoingRequest.php
MGatner: Okay reading the rest of the code I see I must have misunderstood your UML diagram: it goes Message > OutgoingRequest > Request.

This seems very unintuitive; why don't we have a generic Request with common code that Incoming and Outgoing both extend from?

MGatner (changes requested):

MGatner:
I'm not sure if you implemented it this way because of the restrictions in our current HTTP layer, but the inheritance path for requests is confusing. What is restricting us from making Request a common base for both Incoming and Outgoing, and those not being in the same path?

kenjis:
PSR-7: MessageInterface > RequestInterface (outgoing) > ServerRequestInterface (incomming)
This PR: MessageInterface > OutgoingRequestInterface (outgoing) > RequestInterface (incomming)

> What is restricting us from making Request a common base for both Incoming and Outgoing

CI4's Request is incomming request.

system/HTTP/OutgoingRequest.php
MGatner: Answered in main thread!

MGatner:
Wow you are totally right! TIL an incoming request is actually an outgoing request with some extra methods. 

So next question... why the new class? If we already have Message > Request > IncomingRequest, what is stopping us from making Request into the equivalent of OutgoingRequest? (sorry if this is obvious, it is very hard to flip between interfaces on mobile)

kenjis:
> what is stopping us from making Request into the equivalent of OutgoingRequest?

The reason for adding the new class is to make as few changes as possible.

`RequestInterface` is incoming request in CI4.
If we change `Request` to outgoing request, it seems better to change `RequestInterface` to outgoing request.
Then we need to add ` IncommingRequestInterface`.

Do you mean like this?
MessageInterface -> Message
OutgoingRequestInterface -> Request
RequestInterface -> IncommingRequest

MGatner:
I'm with you now, took me a little bit. The structure and names are confusing, but that's not your fault. I would like to look at this one more time on desktop, but generally I think it is a good direction and you should feel free to proceed with the docs.

system/HTTP/Request.php
kenjis: (incoming)Request extends OutgoingRequest.

system/HTTP/OutgoingRequest.php
kenjis: If we rename to it, the return value should be reversed.

This is from PSR-7 method explanation.
```
     * If the Host header is missing or empty, and the new URI contains
     *   a host component, this method MUST update the Host header in the returned
     *   request.
```
I think it is easy to get the implementation is really followed by PSR when we read the code.

If you can't bear it, you could change it later, because it is a private method.

kenjis:
Added changelog.

system/HTTP/OutgoingRequest.php
MGatner: Thanks for explaining, this is fine.

system/HTTP/Request.php
MGatner: Right! We cleared this up in the other threads, I never came back and closed this. Not an issue!

MGatner (approved):
I think we're there! I would like one other set of eyes here. @paulbalandan or @samsonasik or @iRedds ?

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -58,8 +58,19 @@ Others
 Interface Changes
 =================
 
+OutgoingRequestInterface
+------------------------
+
```
MGatner, line 63: Since 4.3 breaks so many other interfaces maybe add a note:
```suggestion
Note: these changes are all backwards-compatible and require no intervention.

```

MGatner: Great!

samsonasik (approved):

PR #6703, opened 2022-10-17
Exceptions - Log Deprecations

najdanovicivan:
**Description**
Currently the framework will always throw critical error on deprecation. This adds option to make deprecations only be logged as warnings.

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [X] Conforms to style guide

PR #6705, opened 2022-10-17
feat: Opt-in logging of deprecations

paulbalandan:
**Description**
~Depends on #6704~
Alternative to #6703 

Opt-in feature to log deprecations (both internal and user-supplied) instead of thrown by the error handler.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Exceptions.php
+++ b/app/Config/Exceptions.php
@@ -3,6 +3,7 @@
 namespace Config;
 
 use CodeIgniter\Config\BaseConfig;
+use Psr\Log\LogLevel;
 
 /**
  * Setup how the exception handler works.
@@ -49,4 +50,28 @@ class Exceptions extends BaseConfig
      * ex. ['server', 'setup/password', 'secret_token']
      */
     public array $sensitiveDataInTrace = [];
+
+    /**
+     * --------------------------------------------------------------------------
+     * LOG DEPRECATIONS INSTEAD OF THROWING?
+     * --------------------------------------------------------------------------
+     * By default, CodeIgniter converts deprecations into exceptions. Also,
+     * starting in PHP 8.1 will cause a lot of deprecated usage warnings.
+     * Use this option to temporarily cease the warnings and instead log those.
+     * This option also works for user deprecations.
+     */
+    public bool $logDeprecationsOnly = false;
+
+    /**
+     * --------------------------------------------------------------------------
+     * LOG LEVEL THRESHOLD FOR DEPRECATIONS
+     * --------------------------------------------------------------------------
+     * If `$logDeprecationsOnly` is set to `true`, this sets the log level
+     * to which the deprecation will be logged. This should be one of the log
+     * levels recognized by PSR-3.
+     *
+     * The related `Config\Logger::$threshold` should be adjusted, if needed,
+     * to capture logging the deprecations.
+     */
+    public string $deprecationLogLevel = LogLevel::WARNING;
 }
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -22,6 +22,7 @@ use CodeIgniter\HTTP\ResponseInterface;
 use Config\Exceptions as ExceptionsConfig;
 use Config\Paths;
 use ErrorException;
+use Psr\Log\LogLevel;
 use Throwable;
 
 /**
@@ -81,6 +82,10 @@ class Exceptions
         // workaround for upgraded users
         if (! isset($this->config->sensitiveDataInTrace)) {
             $this->config->sensitiveDataInTrace = [];
+        }
+        if (! isset($this->config->logDeprecationsOnly, $this->config->deprecationLogLevel)) {
+            $this->config->logDeprecationsOnly = false;
+            $this->config->deprecationLogLevel = LogLevel::WARNING;
         }
     }
 
@@ -155,6 +160,10 @@ class Exceptions
      */
     public function errorHandler(int $severity, string $message, ?string $file = null, ?int $line = null)
     {
+        if ($this->isDeprecationError($severity) && $this->config->logDeprecationsOnly) {
+            return $this->handleDeprecationError($message, $file, $line);
+        }
+
         if (! (error_reporting() & $severity)) {
             return;
         }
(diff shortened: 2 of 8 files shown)


app/Config/Exceptions.php
```diff
@@ -49,4 +50,28 @@ class Exceptions extends BaseConfig
...
      */
     public array $sensitiveDataInTrace = [];
+
+    /**
+     * --------------------------------------------------------------------------
+     * LOG DEPRECATIONS INSTEAD OF THROWING?
+     * --------------------------------------------------------------------------
+     * By default, CodeIgniter converts deprecations into exceptions. Also,
+     * starting in PHP 8.1 will cause a lot of deprecated usage warnings.
+     * Use this option to temporarily cease the warnings ang instead log those.
```
MGatner, line 60:
```suggestion
     * Use this option to temporarily cease the warnings and instead log those.
```

system/Debug/Exceptions.php
```diff
@@ -82,6 +82,10 @@ public function __construct(ExceptionsConfig $config, $request, ResponseInterfac
         if (! isset($this->config->sensitiveDataInTrace)) {
             $this->config->sensitiveDataInTrace = [];
         }
+        if (! isset($this->config->logDeprecationsOnly, $this->config->deprecationLogLevel)) {
+            $this->config->logDeprecationsOnly = false;
+            $this->config->deprecationLogLevel = 'warning';
```
MGatner, line 87: Use the constant here too?

system/Debug/Exceptions.php
```diff
@@ -328,6 +336,29 @@ protected function determineCodes(Throwable $exception): array
...
     }
 
+    private function isDeprecationError(int $error): bool
+    {
+        $deprecations = E_DEPRECATED | E_USER_DEPRECATED;
+
+        return ($error & $deprecations) !== 0;
+    }
+
+    private function deprecationErrorHandler(string $message, ?string $file = null, ?int $line = null): bool
```
MGatner, line 346: I was expecting this to return some handler object. Can we give this an action name? Like:
```suggestion
    private function handleDeprecationError(string $message, ?string $file = null, ?int $line = null): bool
```

system/Debug/Exceptions.php
```diff
@@ -328,6 +336,29 @@ protected function determineCodes(Throwable $exception): array
         return [$statusCode, $exitStatus];
     }
 
+    private function isDeprecationError(int $error): bool
+    {
+        $deprecations = E_DEPRECATED | E_USER_DEPRECATED;
+
+        return ($error & $deprecations) !== 0;
+    }
+
```
MGatner, line 345:
```suggestion
    /**
     * @return true
     */
```

MGatner (changes requested):
I'm haven't looked at test, waiting on rebase. Overall looks good! I think you're right: this will become a much more appealing feature for users once 7.4 drops and 8.1 and 8.2 surge.

system/Debug/Exceptions.php
paulbalandan: Rector removes the docblock as useless.

MGatner: Huh that's odd. We have `true` as a native type coming up soon so it clearly has value over just `bool`. @samsonasik any ideas?

tests/system/Debug/ExceptionsTest.php
```diff
@@ -32,9 +33,55 @@ final class ExceptionsTest extends CIUnitTestCase
 
     protected function setUp(): void
     {
+        parent::setUp();
+
         $this->exception = new Exceptions(new ExceptionsConfig(), Services::request(), Services::response());
     }
 
+    /**
+     * @requires PHP >= 8.1
```
MGatner, line 42: What about one of the 8.0 deprecations? This one would probably be easy to do with an anonymous function:

> Deprecate required parameters after optional parameters in function/method signatures

MGatner (commented):
Looks good! A couple of thoughts still there, but I'm good with this if we need to launch at any point.

system/Debug/Exceptions.php
samsonasik: @MGatner That's possibly a bug on `RemoveUselessReturnTagRector`.

Could you create failing test case PR to https://github.com/rectorphp/rector-src/tree/4cecebfd7e86cafd9e09f8e1fd493046aa1181ea/rules-tests/DeadCode/Rector/ClassMethod/RemoveUselessReturnTagRector/Fixture ? You can copy one of fixture with `skip_` prefix.

That's should can be skipped via `@noRector` tag:

```php
    /**
     * @noRector \Rector\DeadCode\Rector\ClassMethod\RemoveUselessReturnTagRector
     * @return true
     */
```

or register `RemoveUselessReturnTagRector` on rector.php into skip config:

```php
$rectorConfig->skip([
     // ...
     RemoveUselessReturnTagRector::class => [
           __DIR__ . '/system/Debug/Exceptions.php'
    ],
]);
```

MGatner: Thanks @samsonasik. Today I learned about the `@noRector` tag: I've always wanted that!!

I probably won't get to this today but I can follow up with Rector at some point. For now let's ignore this and keep the docblock.

tests/system/Debug/ExceptionsTest.php
paulbalandan: It seems this is not workable. PHP analyzes the function signature at compile time so our error handler is not registered yet.
```
Deprecated: Optional parameter $str declared before required parameter $num is implicitly treated as a required parameter in C:\Users\P\Desktop\Web Dev\CodeIgniter4\tests\system\Debug\ExceptionsTest.php on line 55

[Call Stack]
....

Runtime:       PHP 8.1.11
Configuration: C:\Users\P\Desktop\Web Dev\CodeIgniter4\phpunit.xml.dist

F......                                                             7 / 7 (100%)

Time: 00:00.497, Memory: 48.00 MB

There was 1 failure:

1) CodeIgniter\Debug\ExceptionsTest::testDeprecationsOnPhp80DoNotThrow
Failed asserting that logs have a record of message containing "Optional parameter $str declared before required parameter $num" with level "error".
Failed asserting that false is true.

C:\Users\P\Desktop\Web Dev\CodeIgniter4\system\Test\CIUnitTestCase.php:374
C:\Users\P\Desktop\Web Dev\CodeIgniter4\tests\system\Debug\ExceptionsTest.php:56
```

paulbalandan: It's happening at zend_compile.c
https://github.com/nikic/php-src/blob/9b4601a65c84d0e35ef70fc64473904cb5a5d77c/Zend/zend_compile.c#L5848-L5856

paulbalandan: https://github.com/php/php-src/issues/8529

MGatner: Nice sleuthing! Let's stick with what you have; the rest of the 8.0 deprecations look even more complicated.

MGatner (approved):
💪

tests/system/Debug/ExceptionsTest.php
```diff
@@ -32,9 +33,55 @@ final class ExceptionsTest extends CIUnitTestCase
...
+        $config = new ExceptionsConfig();
+
+        $config->logDeprecationsOnly = true;
+        $config->deprecationLogLevel = 'error';
+
+        $this->exception = new Exceptions($config, Services::request(), Services::response());
+        $this->exception->initialize();
+
+        @trigger_error('Hello! I am a deprecation!', E_USER_DEPRECATED);
+        $this->assertLogContains('error', '[DEPRECATED] Hello! I am a deprecation!', false);
```
paulbalandan, line 79:
```suggestion
        $this->assertLogContains('error', '[DEPRECATED] Hello! I am a deprecation!');
```

kenjis (approved):

PR #6709, opened 2022-10-18
Add template types to `Connection`

paulbalandan:
**Description**
Making `BaseConnection` a generic type, thus extensions need to specify the type of the connection.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -45,6 +45,10 @@ use Throwable;
  * @property int        $transDepth
  * @property bool       $transFailure
  * @property bool       $transStatus
+ *
+ * @template TConnection of false|object|resource
+ *
+ * @implements ConnectionInterface<TConnection>
  */
 abstract class BaseConnection implements ConnectionInterface
 {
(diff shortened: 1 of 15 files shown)


system/Database/BaseConnection.php
```diff
@@ -45,6 +45,10 @@
  * @property int        $transDepth
  * @property bool       $transFailure
  * @property bool       $transStatus
+ *
+ * @template TConnection of bool|object|resource
```
MGatner, line 49: You're going to need to explain the `bool` to me... Is this to support returning `false` on failure?

MGatner (commented):
Where did those abstract functions go?

paulbalandan:
> Where did those abstract functions go?

They are actually defined by `ConnectionInterface` and unnecessarily duplicated as `abstract public` in `BaseConnection`.

system/Database/BaseConnection.php
paulbalandan: On second look, I think this should have been `false` not `bool`.

paulbalandan:
I'll fix the conflicts once #6713 is merged.

MGatner:
Thanks for explaining (and extracting).

kenjis (approved):

MGatner (approved):
Amazing cleanup of the baseline!

PR #6710, opened 2022-10-18
feat: custom exception handler

kenjis:
**Description**
Supersedes #5675

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/BaseExceptionHandler.php
+++ b/system/Debug/BaseExceptionHandler.php
@@ -123,15 +123,9 @@ abstract class BaseExceptionHandler
      */
     protected static function describeMemory(int $bytes): string
     {
-        if ($bytes < 1024) {
-            return $bytes . 'B';
-        }
+        helper('number');
 
-        if ($bytes < 1_048_576) {
-            return round($bytes / 1024, 2) . 'KB';
-        }
-
-        return round($bytes / 1_048_576, 2) . 'MB';
+        return number_to_size($bytes, 2);
     }
 
     /**
--- a/system/Debug/BaseExceptionHandler.php
+++ b/system/Debug/BaseExceptionHandler.php
@@ -118,6 +118,8 @@ abstract class BaseExceptionHandler
     /**
      * Describes memory usage in real-world units. Intended for use
      * with memory_get_usage, etc.
+     *
+     * @used-by app/Views/errors/html/error_exception.php
      */
     protected static function describeMemory(int $bytes): string
     {
@@ -134,6 +136,8 @@ abstract class BaseExceptionHandler
 
     /**
      * Creates a syntax-highlighted version of a PHP file.
+     *
+     * @used-by app/Views/errors/html/error_exception.php
      *
      * @return bool|string
      */
--- a/system/Debug/BaseExceptionHandler.php
+++ b/system/Debug/BaseExceptionHandler.php
@@ -1,0 +1,246 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Debug;
+
+use CodeIgniter\API\ResponseTrait;
+use CodeIgniter\HTTP\RequestInterface;
+use CodeIgniter\HTTP\ResponseInterface;
+use Config\Exceptions as ExceptionsConfig;
+use Throwable;
+
+/**
+ * Provides common functions for exception handlers,
+ * especially around displaying the output.
+ */
+abstract class BaseExceptionHandler
+{
+    use ResponseTrait;
+
+    /**
+     * ResponseTrait needs this.
+     */
+    protected ?RequestInterface $request;
+
+    /**
+     * ResponseTrait needs this.
+     */
+    protected ?ResponseInterface $response;
+
+    /**
+     * Config for debug exceptions.
+     */
+    protected ExceptionsConfig $config;
+
+    /**
+     * Nesting level of the output buffering mechanism
+     */
+    protected int $obLevel;
+
+    /**
+     * The path to the directory containing the
+     * cli and html error view directories.
+     */
+    protected string $viewPath;
+
+    public function __construct(ExceptionsConfig $config)
+    {
+        $this->config = $config;
+
+        $this->obLevel  = ob_get_level();
+        $this->viewPath = rtrim($this->config->errorViewPath, '\\/ ') . DIRECTORY_SEPARATOR;
+    }
+
+    /**
+     * The main entry point into the handler.
+     *
+     * @return void
+     */
+    abstract public function handle(
+        Throwable $exception,
+        RequestInterface $request,
+        ResponseInterface $response,
+        int $statusCode,
+        int $exitCode
+    );
+
+    /**
+     * Gathers the variables that will be made available to the view.
+     */
+    protected function collectVars(Throwable $exception, int $statusCode): array
+    {
+        $trace = $exception->getTrace();
+
+        if ($this->config->sensitiveDataInTrace !== []) {
+            $this->maskSensitiveData($trace, $this->config->sensitiveDataInTrace);
+        }
+
+        return [
+            'title'   => get_class($exception),
+            'type'    => get_class($exception),
+            'code'    => $statusCode,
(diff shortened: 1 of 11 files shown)


system/Debug/BaseExceptionHandler.php
```diff
@@ -0,0 +1,248 @@
...
+use Config\Exceptions as ExceptionsConfig;
+use Throwable;
+
+/**
+ * Provides common functions for exception handlers,
+ * especially around displaying the output.
+ */
+abstract class BaseExceptionHandler
+{
+    use ResponseTrait;
```
MGatner, line 26: What is needed from this trait? I find it hard to believe our exception handler needs all those methods. If it is just 1-2 I'd rather see them duplicated.

MGatner (commented):
This is a big one, I will have to come back to it! Initially it looks really good. Please be sure to work with @lonnieezell for feature consistency and @paulbalandan if there's any collaboration to be had with the deprecations handler PR.

system/Debug/BaseExceptionHandler.php
kenjis: This PR is based on Lonnie's one.

lonnieezell:
I'll be honest - this week I'm completely wrecked for time so I won't be able to really review until next week. 

That said - the reason nothing happened with the original one was because it was a BC break and would have to wait until v5. Has the BC break portion been fixed with this PR?

kenjis:
No problem. The original PR was created many months ago.

I have tried to keep this PR free of BC breaks.

kenjis:
Rebased.

system/Debug/BaseExceptionHandler.php
kenjis: Moved to the child class.

system/Debug/BaseExceptionHandler.php
```diff
@@ -0,0 +1,239 @@
...
+        }
+    }
+
+    /**
+     * Describes memory usage in real-world units. Intended for use
+     * with memory_get_usage, etc.
+     *
+     * @used-by app/Views/errors/html/error_exception.php
+     */
+    protected static function describeMemory(int $bytes): string
```
MGatner, line 124: Why not use Number Helper's `number_to_size()`?

system/Debug/BaseExceptionHandler.php
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * Creates a syntax-highlighted version of a PHP file.
+     *
+     * @used-by app/Views/errors/html/error_exception.php
+     *
+     * @return bool|string
+     */
+    protected static function highlightFile(string $file, int $lineNumber, int $lines = 15)
```
MGatner, line 144: Since the existing handler doesn't actually extend this base, can we go ahead an type this?

system/Debug/Exceptions.php
```diff
@@ -122,6 +125,23 @@ public function exceptionHandler(Throwable $exception)
             ]);
         }
 
+        // For upgraded users.
+        if (! method_exists($this->config, 'handler')) {
+            $this->defaultExceptionHandler($exception, $statusCode, $exitCode);
```
MGatner, line 130: Why doesn't this use `ExceptionHandler`? Is it not backwards-compatible behavior?

system/Debug/ExceptionHandler.php
```diff
@@ -0,0 +1,149 @@
...
+use CodeIgniter\API\ResponseTrait;
+use CodeIgniter\Exceptions\PageNotFoundException;
+use CodeIgniter\HTTP\Exceptions\HTTPException;
+use CodeIgniter\HTTP\IncomingRequest;
+use CodeIgniter\HTTP\RequestInterface;
+use CodeIgniter\HTTP\ResponseInterface;
+use Config\Paths;
+use Throwable;
+
+class ExceptionHandler extends BaseExceptionHandler
```
MGatner, line 23: I think we should make this `final` and see how it goes. Someone wanting to make a slight adjustment can use a decorator/proxy.

system/Debug/ExceptionHandler.php
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * ResponseTrait needs this.
+     */
+    protected ?ResponseInterface $response = null;
+
+    /**
+     * Determines the correct way to display the error.
+     *
+     * @return void
```
MGatner, line 40: It feels like a big loss not having this in the abstraction. Any reason not to do that?

MGatner (changes requested):
I haven't looked at docs and tests yet, but this is looking great! I added some comments but two broad thoughts:
1. It seems like the new classes are independent of the existing ones, and you use the current handle as a mediator to call them when available. I like the freedom that gives us - but then you kept ExceptionHandler compatible with the current class. It don't understand the need for this backwards-compatibility, unless I am missing something?
2. I know we've been burned by class-interface-abstraction, but enforcing our BaseExceptionHandler for a single `handle()` method feels unduly restrictive. I would like to see an interface (just that method) used as the return from the Config file and the abstract class becomes optional.

system/Debug/BaseExceptionHandler.php
kenjis: Done.

system/Debug/ExceptionHandler.php
kenjis: Done.

system/Debug/Exceptions.php
kenjis: Good point.
It should use ExceptionHandler. 
When using ExceptionHandler, if the behavior changes it is a breaking change.

system/Debug/BaseExceptionHandler.php
MGatner: Thanks! Also, apologies: when I originally wrote this I didn't realize you had just moved this from its previous location.

system/Debug/BaseExceptionHandler.php
kenjis: What do you mean?

kenjis:
No. If a dev overrides the following methods, this PR breaks the extended Exceptions:
- Exceptions::render()
- Exceptions::determineView()
- Exceptions::collectVars()
- Exceptions::maskSensitiveData()

kenjis:
Reverting https://github.com/codeigniter4/CodeIgniter4/pull/6710/commits/9d5d44f336912e1d4e548cc966f3617ab86ce400 and a dev does not update the Config class,
then there is no BC break even if the dev extends `Exceptions`.

MGatner:
I'm confused; I think I missed something. `ExceptionHandler` and `BaseExceptionHandler` are new classes, and `Exceptions` is not being changed to extend either. So why do the new classes need to be backwards-compatible with `Exceptions`?

MGatner:
> No. If a dev overrides the following methods, this PR breaks the extended Exceptions

If we want to support extended classes just check `if (static::class === self::class)` before passing off to the handler.

kenjis:
> I'm confused; I think I missed something. ExceptionHandler and BaseExceptionHandler are new classes, and Exceptions is not being changed to extend either. So why do the new classes need to be backwards-compatible with Exceptions?

The new classes do not need to be backwards-compatible with Exceptions.

A dev can replace Exceptions with their custom extended class now, and if the dev overrides one of some methods,
this PR breaks the custom Exceptions.

> If we want to support extended classes just check if (static::class === self::class) before passing off to the handler.

It may be Okay with that check, as a dev would not need to use their own ExceptionHandler and replace Exceptions at the same time.
Alternatively, it can also be determined by setting it to opt-in in the Config.

system/Debug/ExceptionHandler.php
```diff
@@ -0,0 +1,149 @@
...
+
+        // Determine the views
+        $view    = $this->determineView($exception, $path);
+        $altView = $this->determineView($exception, $altPath);
+
+        // Check if the view exists
+        $viewFile = null;
+        if (is_file($path . $view)) {
+            $viewFile = $path . $view;
+        } elseif (is_file($altPath . $altView)) {
```
lonnieezell, line 106: Seems like we need to handle if neither file exists?

lonnieezell (commented):
Looks pretty good to me. I just found one edge case that we don't handle.

MGatner:
> The new classes do not need to be backwards-compatible with Exceptions. 

Since this is the case please add explicit types to all the new class methods. 

I am also still keen on adding an interface - nothing else in the base class is mandatory so we shouldn't restrict users to extending it.

PR #6712, opened 2022-10-18
fix: add missing methods to BaseConnection

kenjis:
**Description**
- add missing methods `_disableForeignKeyChecks()` and `_ebleForeignKeyChecks()`
- fix a bug that `Forge::dropTable()` causes an error if an extended connection class does not have these methods

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/Mock/MockConnection.php
+++ b/system/Test/Mock/MockConnection.php
@@ -240,4 +240,14 @@ class MockConnection extends BaseConnection
     {
         return true;
     }
+
+    protected function _disableForeignKeyChecks()
+    {
+        return '';
+    }
+
+    protected function _enableForeignKeyChecks()
+    {
+        return '';
+    }
 }
--- a/system/Test/Mock/MockConnection.php
+++ b/system/Test/Mock/MockConnection.php
@@ -240,14 +240,4 @@ class MockConnection extends BaseConnection
     {
         return true;
     }
-
-    protected function _disableForeignKeyChecks()
-    {
-        return '';
-    }
-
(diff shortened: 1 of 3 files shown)


MGatner (commented):
How necessary are these? Would adding `if (method_exists($this, ...` be an alternative? Surely there are DB handlers that don't even bother with FKs.

paulbalandan:
This can be breaking for custom driver classes extending `BaseConnection`. With this change, if I run my app using `MongoDbConnection` for example, I will get an error that my driver does not implement the two abstract methods which needs to be declared abstract. Maybe instead of abstract just create a default implementation of those two methods in BaseConnection which the child drivers can just override if they support it.

kenjis:
I've changed the implementation to avoid BC breaks.

MGatner:
Looks better!

system/Test/Mock/MockConnection.php
```diff
@@ -240,4 +240,14 @@ protected function _transRollback(): bool
     {
         return true;
     }
+
+    protected function _disableForeignKeyChecks()
```
paulbalandan, line 244: These can be removed.

kenjis: Removed.

kenjis:
I've changed my mind.
To begin with, BaseConnection is already too big. 
Methods related foreign keys should be moved to another interface in the future.

paulbalandan (approved):
I agree. Too many classes in the framework need to be modularized. They perform too many responsibilities.

PR #6713, opened 2022-10-19
refactor: remove unneeded abstract public methods

paulbalandan:
**Description**
These are already defined by `ConnectionInterface`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -427,13 +427,6 @@ abstract class BaseConnection implements ConnectionInterface
     }
 
     /**
-     * Connect to the database.
-     *
-     * @return mixed
-     */
-    abstract public function connect(bool $persistent = false);
-
-    /**
      * Close the database connection.
      */
     public function close()
@@ -462,14 +455,6 @@ abstract class BaseConnection implements ConnectionInterface
     }
 
     /**
-     * Keep or establish the connection if no queries have been sent for
-     * a length of time exceeding the server's idle timeout.
-     *
-     * @return mixed
-     */
-    abstract public function reconnect();
-
-    /**
      * Returns the actual connection object. If both a 'read' and 'write'
      * connection has been specified, you can pass either term in to
      * get that connection. If you pass either alias in and only a single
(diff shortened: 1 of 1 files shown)


kenjis (approved):

MGatner:
It feels a little weird to have `implements ConnectionInterface` without these methods. Normally I would look at an abstract class to see what I need to fulfill as far as the contract, which won't work in this case. I understand the deduplication, just noting that it feels a bit odd. 

Does anyone have examples of this somewhere else (even outside of CodeIgniter)? Where an abstract class implements the interface without the actual methods?

paulbalandan:
If you are not using an IDE then I believe it is normal to feel the weirdness.

PHP won't complain the absence of the interface's methods in the abstract class because it is _abstract_ (i.e. incomplete). In that case, PHP will assume the concrete classes will complete fulfilling the interface.

MGatner:
It's an interesting consideration that I haven't come across before. I did a little research and didn't find much but this rather mixed-quality, old conversation thread: https://thuvienphapluat.edu.vn/can-abstract-class-implement-interface-in-php

If you two concur on this approach then I'm good with it.

MGatner:
> If you are not using an IDE

I use Intelephense but I don't know that it really matters for this case unless your IDE is somehow generating the initial class. When I'm fulfilling an abstraction I literally copy the definitions (either from the interface or the abstract methods from the abstract class) and work on filling them in. I was just noting that in this case one would have to know to go to the interface in this case, not the abstract class.

paulbalandan:
This is an example from Symfony Console's InputInterface

InputInterface defines a getFirstArgument method:
https://github.com/symfony/symfony/blob/c2ad569b9f66f6ff0080a93598a4c89682da1cb6/src/Symfony/Component/Console/Input/InputInterface.php#L30

but abstract class Input does not define it nor list it as abstract:
https://github.com/symfony/symfony/blob/c2ad569b9f66f6ff0080a93598a4c89682da1cb6/src/Symfony/Component/Console/Input/Input.php#L28

because the child classes like ArgvInput will fulfill the contract:
https://github.com/symfony/symfony/blob/c2ad569b9f66f6ff0080a93598a4c89682da1cb6/src/Symfony/Component/Console/Input/ArgvInput.php#L263

MGatner:
Awesome! Thanks for that research. Yes let's proceed with this.

PR #6715, opened 2022-10-19
docs: fix PHPDoc type mixed

kenjis:
~~Needs #6722~~

**Description**
See #6310

- fix PHPDoc types
- ~~remove unneeded empty check~~

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -416,7 +416,7 @@ abstract class BaseModel
      * through soft deletes (deleted = 1)
      * This methods works only with dbCalls
      *
-     * @return bool|string
+     * @return bool|string Returns a string if in test mode.
      */
     abstract protected function doPurgeDeleted();
 
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -416,7 +416,7 @@ abstract class BaseModel
      * through soft deletes (deleted = 1)
      * This methods works only with dbCalls
      *
-     * @return bool|mixed
+     * @return bool|string
      */
     abstract protected function doPurgeDeleted();
 
--- a/system/Common.php
+++ b/system/Common.php
@@ -824,8 +824,9 @@ if (! function_exists('old')) {
      * Provides access to "old input" that was set in the session
      * during a redirect()->withInput().
      *
-     * @param string|null $default
-     * @param bool|string $escape
+     * @param string|null  $default
+     * @param false|string $escape
+     * @phpstan-param false|'attr'|'css'|'html'|'js'|'raw'|'url' $escape
      *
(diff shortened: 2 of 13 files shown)


system/BaseModel.php
```diff
@@ -416,7 +416,7 @@ abstract protected function doDelete($id = null, bool $purge = false);
      * through soft deletes (deleted = 1)
      * This methods works only with dbCalls
      *
-     * @return bool|mixed
+     * @return bool|string
```
MGatner, line 419: What the heck is a string return here??

system/Common.php
```diff
@@ -414,10 +414,11 @@ function env(string $key, $default = null)
      * If $data is an array, then it loops over it, escaping each
      * 'value' of the key/value pairs.
      *
-     * Valid context values: html, js, css, url, attr, raw
-     *
      * @param array|string $data
-     * @param string       $encoding
+     * @param string       $context  Valid context values: html, js, css, url, attr, raw
```
MGatner, line 418: Let's make these literals, at least for PHPStan:
```suggestion
     * @param 'html'|'js'|'css'|'url'|'attr'|'raw' $context
```

MGatner (changes requested):

system/BaseModel.php
kenjis: When in test mode.

system/Common.php
```diff
@@ -437,7 +438,7 @@ function esc($data, string $context = 'html', ?string $encoding = null)
             // Provide a way to NOT escape data since
             // this could be called automatically by
             // the View library.
-            if (empty($context) || $context === 'raw') {
+            if ($context === 'raw') {
```
samsonasik, line 441: please revert, '0' is empty, ref https://3v4l.org/mgSdd

this PR should only change PHPDoc, change code may cause regression.

samsonasik: If it needs to be applied, I think it should go to 4.3 as it changed behaviour.

kenjis: @samsonasik Technically you are correct.

But I cannot imagine users who write code like `esc($string, '0')`.
I see no problem with this change in develop.

samsonasik: Ok then 👍

kenjis: @samsonasik I think accepting `esc($string, '0')` is a bug.
I'll fix it in another PR.
Thank you for commenting!

kenjis: I sent #6722

MGatner (approved):

PR #6717, opened 2022-10-19
feat: void element tags in helpers are selectable between `>` and `/>` 

### LINKED ISSUE #6649: Dev: remove self-closing HTML tags in helpers

ddevsr:
**Description**
Fixes #6649
Supersedes and closes #6658

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/DocTypes.php
+++ b/app/Config/DocTypes.php
@@ -30,4 +30,19 @@ class DocTypes
         'xhtml-rdfa-1'      => '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML+RDFa 1.0//EN" "http://www.w3.org/MarkUp/DTD/xhtml-rdfa-1.dtd">',
         'xhtml-rdfa-2'      => '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML+RDFa 1.1//EN" "http://www.w3.org/MarkUp/DTD/xhtml-rdfa-2.dtd">',
     ];
+
+    /**
+     * Endline HTML Tag
+     *
+     * Generate endline HTML Tag by Document Type
+     *
+     * Currently :
+     * - html5
+     * - Older HTML like `xhtml`
+     *
+     * For :
+     * - `<input>`
+     * - `<img>`
+     */
+    public string $endlineTag = 'html5';
 }
--- a/app/Config/DocTypes.php
+++ b/app/Config/DocTypes.php
@@ -34,15 +34,11 @@ class DocTypes
     /**
      * Endline HTML Tag
      *
-     * Generate endline HTML Tag by Document Type
-     *
-     * Currently :
-     * - html5
-     * - Older HTML like `xhtml`
+     * This setting is for html and form helpers.
      *
      * For :
      * - `<input>`
      * - `<img>`
      */
-    public string $endlineTag = 'html5';
+    public bool $xhtml = false;
 }
--- a/system/Common.php
+++ b/system/Common.php
@@ -571,19 +571,14 @@ if (! function_exists('htmltag_endline')) {
      */
     function htmltag_endline(): string
     {
-        $config       = new Config\DocTypes();
-        $doctype      = $config->endlineTag;
-        $listDoctypes = $config->list;
+        $config  = config('DocTypes');
+        $doctype = $config->xhtml;
 
-        if (! array_key_exists($doctype, $listDoctypes)) {
-            return '';
+        if ($doctype) {
+            return '/>';
         }
 
-        if ($doctype === 'html5') {
-            return '>';
-        }
-
-        return '/>';
+        return '>';
     }
 }
 
--- a/system/Common.php
+++ b/system/Common.php
@@ -563,6 +563,30 @@ if (! function_exists('function_usable')) {
     }
 }
 
+if (! function_exists('htmltag_endline')) {
+    /**
+     * Endline HTML Tag
+     *
+     * Generates endline html tag by document type
+     */
+    function htmltag_endline(): string
+    {
+        $config       = new Config\DocTypes();
+        $doctype      = $config->endlineTag;
+        $listDoctypes = $config->list;
+
+        if (! array_key_exists($doctype, $listDoctypes)) {
+            return '';
+        }
+
+        if ($doctype === 'html5') {
+            return '>';
+        }
+
+        return '/>';
+    }
+}
+
 if (! function_exists('helper')) {
     /**
      * Loads a helper file into memory. Supports namespaced helpers,
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -148,7 +148,7 @@ if (! function_exists('form_input')) {
             'value' => $value,
         ];
 
-        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " />\n";
+        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . htmltag_endline() . "\n";
     }
 }
 
@@ -194,7 +194,7 @@ if (! function_exists('form_upload')) {
 
         $data['type'] = 'file';
 
-        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " />\n";
+        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . htmltag_endline() . "\n";
     }
 }
 
@@ -365,7 +365,7 @@ if (! function_exists('form_checkbox')) {
             $defaults['checked'] = 'checked';
         }
 
-        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " />\n";
+        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . htmltag_endline() . "\n";
     }
 }
 
--- a/tests/system/Helpers/FormHelperTest.php
+++ b/tests/system/Helpers/FormHelperTest.php
@@ -101,7 +101,7 @@ final class FormHelperTest extends CIUnitTestCase
             $Name     = csrf_token();
             $expected = <<<EOH
                 <form action="http://example.com/index.php" name="form" id="form" method="POST" accept-charset="utf-8">
-                <input type="hidden" name="{$Name}" value="{$Value}" style="display:none;" />
+                <input type="hidden" name="{$Name}" value="{$Value}" style="display:none;">
 
                 EOH;
         } else {
@@ -128,7 +128,7 @@ final class FormHelperTest extends CIUnitTestCase
             $Name     = csrf_token();
             $expected = <<<EOH
                 <form action="http://example.com/index.php/foo/bar" name="form" id="form" method="post" accept-charset="utf-8">
-                <input type="hidden" name="{$Name}" value="{$Value}" style="display:none;" />
+                <input type="hidden" name="{$Name}" value="{$Value}" style="display:none;">
 
                 EOH;
         } else {
@@ -155,15 +155,15 @@ final class FormHelperTest extends CIUnitTestCase
             $Name     = csrf_token();
             $expected = <<<EOH
                 <form action="http://example.com/index.php/foo/bar" name="form" id="form" method="POST" accept-charset="utf-8">
-                <input type="hidden" name="foo" value="bar" />
-                <input type="hidden" name="{$Name}" value="{$Value}" />
+                <input type="hidden" name="foo" value="bar">
+                <input type="hidden" name="{$Name}" value="{$Value}">
 
                 EOH;
         } else {
             $expected = <<<'EOH'
                 <form action="http://example.com/index.php/foo/bar" name="form" id="form" method="POST" accept-charset="utf-8">
 
-                <input type="hidden" name="foo" value="bar" />
+                <input type="hidden" name="foo" value="bar">
 
                 EOH;
         }
@@ -189,7 +189,7 @@ final class FormHelperTest extends CIUnitTestCase
             $Name     = csrf_token();
             $expected = <<<EOH
                 <form action="http://example.com/index.php/foo/bar" name="form" id="form" method="POST" enctype="multipart/form-data" accept-charset="utf-8">
-                <input type="hidden" name="{$Name}" value="{$Value}" style="display:none;" />
+                <input type="hidden" name="{$Name}" value="{$Value}" style="display:none;">
 
                 EOH;
         } else {
@@ -214,7 +214,7 @@ final class FormHelperTest extends CIUnitTestCase
     {
         $expected = <<<EOH
 
-            <input type="hidden" name="username" value="johndoe" />\n
+            <input type="hidden" name="username" value="johndoe">\n
             EOH;
         $this->assertSame($expected, form_hidden('username', 'johndoe'));
     }
@@ -226,7 +226,7 @@ final class FormHelperTest extends CIUnitTestCase
         ];
         $expected = <<<'EOH'
 
-            <input type="hidden" name="foo" value="bar" />
+            <input type="hidden" name="foo" value="bar">
 
             EOH;
         $this->assertSame($expected, form_hidden($data, null));
@@ -239,7 +239,7 @@ final class FormHelperTest extends CIUnitTestCase
         ];
         $expected = <<<'EOH'
 
-            <input type="hidden" name="name[foo]" value="bar" />
+            <input type="hidden" name="name[foo]" value="bar">
 
             EOH;
         $this->assertSame($expected, form_hidden('name', $data));
@@ -248,7 +248,7 @@ final class FormHelperTest extends CIUnitTestCase
     public function testFormInput()
     {
         $expected = <<<EOH
-            <input type="text" name="username" value="johndoe" id="username" maxlength="100" size="50" style="width:50%" />\n
+            <input type="text" name="username" value="johndoe" id="username" maxlength="100" size="50" style="width:50%">\n
             EOH;
         $data = [
             'name'      => 'username',
@@ -264,7 +264,7 @@ final class FormHelperTest extends CIUnitTestCase
     public function testFormInputWithExtra()
     {
         $expected = <<<EOH
-            <input type="email" name="identity" value="" id="identity" class="form-control form-control-lg" />\n
+            <input type="email" name="identity" value="" id="identity" class="form-control form-control-lg">\n
             EOH;
         $data = [
             'id'   => 'identity',
@@ -280,7 +280,7 @@ final class FormHelperTest extends CIUnitTestCase
     public function testFormPassword()
     {
         $expected = <<<EOH
-            <input type="password" name="password" value="" />\n
+            <input type="password" name="password" value="">\n
             EOH;
         $this->assertSame($expected, form_password('password'));
     }
@@ -288,7 +288,7 @@ final class FormHelperTest extends CIUnitTestCase
     public function testFormUpload()
     {
         $expected = <<<EOH
-            <input type="file" name="attachment" />\n
+            <input type="file" name="attachment">\n
             EOH;
         $this->assertSame($expected, form_upload('attachment'));
     }
@@ -612,7 +612,7 @@ final class FormHelperTest extends CIUnitTestCase
     public function testFormCheckbox()
     {
         $expected = <<<EOH
-            <input type="checkbox" name="newsletter" value="accept" checked="checked" />\n
+            <input type="checkbox" name="newsletter" value="accept" checked="checked">\n
             EOH;
         $this->assertSame($expected, form_checkbox('newsletter', 'accept', true));
     }
@@ -625,7 +625,7 @@ final class FormHelperTest extends CIUnitTestCase
             'checked' => true,
         ];
         $expected = <<<'EOH'
-            <input type="checkbox" name="foo" value="bar" checked="checked" />
+            <input type="checkbox" name="foo" value="bar" checked="checked">
 
             EOH;
         $this->assertSame($expected, form_checkbox($data));
@@ -639,7 +639,7 @@ final class FormHelperTest extends CIUnitTestCase
             'checked' => false,
         ];
         $expected = <<<'EOH'
-            <input type="checkbox" name="foo" value="bar" />
+            <input type="checkbox" name="foo" value="bar">
 
             EOH;
         $this->assertSame($expected, form_checkbox($data));
@@ -648,7 +648,7 @@ final class FormHelperTest extends CIUnitTestCase
     public function testFormRadio()
     {
         $expected = <<<EOH
-            <input type="radio" name="newsletter" value="accept" checked="checked" />\n
+            <input type="radio" name="newsletter" value="accept" checked="checked">\n
             EOH;
         $this->assertSame($expected, form_radio('newsletter', 'accept', true));
     }
@@ -656,7 +656,7 @@ final class FormHelperTest extends CIUnitTestCase
     public function testFormSubmit()
     {
         $expected = <<<EOH
-            <input type="submit" name="mysubmit" value="Submit Post!" />\n
+            <input type="submit" name="mysubmit" value="Submit Post!">\n
             EOH;
         $this->assertSame($expected, form_submit('mysubmit', 'Submit Post!'));
     }
@@ -683,7 +683,7 @@ final class FormHelperTest extends CIUnitTestCase
     public function testFormReset()
     {
         $expected = <<<EOH
-            <input type="reset" name="myreset" value="Reset" />\n
+            <input type="reset" name="myreset" value="Reset">\n
             EOH;
         $this->assertSame($expected, form_reset('myreset', 'Reset'));
     }
@@ -724,7 +724,7 @@ final class FormHelperTest extends CIUnitTestCase
             'bar1',
         ];
         $expected = <<<'EOH'
-            <input type="text" name="foo" value="bar" list="foo_list" />
+            <input type="text" name="foo" value="bar" list="foo_list">
 
             <datalist id='foo_list'><option value='foo1'>
             <option value='bar1'>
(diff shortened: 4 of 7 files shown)


ddevsr:
@kenjis PHPStan error will be appear if changes typedoc in `form_checkbox` to `array|string`. Maybe you can fix in PR #6715

kenjis:
@ddevsr Okay, I will take care of that.

tests/system/Helpers/FormHelperTest.php
```diff
@@ -943,7 +943,7 @@ public function testValidationShowError()
 
         $html = validation_show_error('id');
 
-        $this->assertSame('<span class="help-block">The ID field is required.</span>' . "\n", $html);
+        $this->assertSame('<span class="help-block">The ID field is required.</span>' . PHP_EOL, $html);
```
kenjis, line 946: Why `PHP_EOL`?

kenjis:
Is this a breaking change or not?
See https://github.com/codeigniter4/CodeIgniter4/pull/6658#issuecomment-1280241607

tests/system/Helpers/FormHelperTest.php
ddevsr: Passed for PHPCSFixer, i forgot template location render for this.

```html
<span class="help-block">The ID field is required.</span>

```

ddevsr:
> Is this a breaking change or not? See [#6658 (comment)](https://github.com/codeigniter4/CodeIgniter4/pull/6658#issuecomment-1280241607)

For XHTML is breaking change, but analytics say `94,8%` using HTML in 2022

See [https://w3techs.com/technologies/history_overview/markup_language/ms/y](https://w3techs.com/technologies/history_overview/markup_language/ms/y)

system/Helpers/form_helper.php
```diff
@@ -766,8 +766,8 @@ function validation_show_error(string $field, string $template = 'single'): stri
      *
      * @internal
      *
-     * @param array|string $attributes List of attributes
-     * @param array        $default    Default values
+     * @param mixed $attributes List of attributes
```
kenjis, line 769: I've fixed the PHPStan error. Please run `git rebase` and drop this change.

kenjis:
> doctype([$type = 'html5'])
> ...
> Helps you generate document type declarations, or DTD’s. HTML 5 is used by default, but many doctypes are available.
https://codeigniter4.github.io/CodeIgniter4/helpers/html_helper.html#doctype

kenjis:
To begin with, HTML5 is already obsolete.
This [list](https://github.com/codeigniter4/CodeIgniter4/blob/77e5dcf27d5dcba6d9c47964bc0805af6d569ccd/app/Config/DocTypes.php#L12) is only used with `doctype()`.
Why don't we make `doctype()` deprecate?

tests/system/Helpers/FormHelperTest.php
ddevsr: image

See in https://github.com/codeigniter4/CodeIgniter4/blob/4.3/system/Validation/Views/single.php#L1, have 2 line

kenjis: What's is it? I don't get what you want to say.

ddevsr:
> To begin with, HTML5 is already obsolete. This [list](https://github.com/codeigniter4/CodeIgniter4/blob/77e5dcf27d5dcba6d9c47964bc0805af6d569ccd/app/Config/DocTypes.php#L12) is only used with `doctype()`. Why don't we make `doctype()` deprecate?

For other HTML5, `doctype()` is useful and very simple

ddevsr:
I already added dynamic appear HTML Tag with setting property `endlineTag` document type by user.

app/Config/DocTypes.php
```diff
@@ -30,4 +30,19 @@ class DocTypes
...
+     *
+     * Currently :
+     * - html5
+     * - Older HTML like `xhtml`
+     *
+     * For :
+     * - `<input>`
+     * - `<img>`
+     */
+    public string $endlineTag = 'html5';
```
kenjis, line 47: This setting is for html and form helpers. I hope to add that in the comment.

I would like to change it to `$xhtml = false`. Because it just has two state.

system/Common.php
```diff
@@ -563,6 +564,26 @@ function function_usable(string $functionName): bool
...
 
+if (! function_exists('htmltag_endline')) {
+    /**
+     * Endline HTML Tag
+     *
+     * Generates endline html tag by document type
+     */
+    function htmltag_endline(): string
+    {
+        $config       = new DocTypes();
```
kenjis, line 575:
```suggestion
        $config       = config('DocTypes');
```

system/Common.php
```diff
@@ -563,6 +564,26 @@ function function_usable(string $functionName): bool
     }
 }
 
+if (! function_exists('htmltag_endline')) {
+    /**
+     * Endline HTML Tag
+     *
+     * Generates endline html tag by document type
+     */
+    function htmltag_endline(): string
```
kenjis, line 573: This function should be `@internal`. It is not a public API.

system/Common.php
```diff
@@ -563,6 +564,26 @@ function function_usable(string $functionName): bool
     }
 }
 
+if (! function_exists('htmltag_endline')) {
```
kenjis, line 567: `Common.php` is already quite large and I would like to add as few functions as possible.
This function is not common. Only form/html helper uses it. So it should not be added here.

What about adding a class and a static method?
Or is there some other appropriate place?
@codeigniter4/core-team

michalsn: I'm okay with adding this to the Common.php. It seems to be well-aligned with what we usually do.

I would definitely add a static variable inside this function.

Adding an extra class just to handle this case seems odd. Are we expecting to have more use cases like this? Maybe something like `config('DocTypes')->htmlTagEndline()`? But I'm unsure if I want this type of call in the helper.

app/Config/DocTypes.php
ddevsr: Makes sense, thank you for suggest

system/Common.php
```diff
@@ -563,6 +563,27 @@ function function_usable(string $functionName): bool
...
+    /**
+     * Endline HTML Tag
+     *
+     * Generates endline html tag by document type
+     *
+     * @internal
+     */
+    function htmltag_endline(): string
```
kenjis, line 574: I think this method name is a little different.
Native English speakers, please give us a more appropriate name.

ddevsr: What about `selfClosingHTMLTag` or `selfclosing_html`?

Ref : https://validator.w3.org/check

paulbalandan: `>` or `/>` are not the self-closing tags referred by the HTML standard. This is just the [`U+003E GREATER-THAN SIGN character (>)`](https://html.spec.whatwg.org/multipage/syntax.html#start-tags)

`<html></html>` => standard elements with start and end tags
`<input>` => void element with start tag only
`<path/>` => foreign element in SVG namespace that is self-closing

app/Config/DocTypes.php
```diff
@@ -30,4 +30,15 @@ class DocTypes
...
+    /**
+     * Endline HTML Tag
+     *
+     * This setting is for html and form helpers.
+     *
+     * For :
+     * - `<input>`
+     * - `<img>`
+     */
+    public bool $xhtml = false;
```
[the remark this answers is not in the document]
kenjis: > This config default as html?

Yes. it is the purpose of this PR.

> so its Breaking Changes?

Not sure, but it should be documented in the changelog.

system/Common.php
```diff
@@ -563,6 +563,27 @@ function function_usable(string $functionName): bool
...
+     * Endline HTML Tag
+     *
+     * Generates endline html tag by document type
+     *
+     * @internal
+     */
+    function selfclosing_html(): string
+    {
+        $config  = config('DocTypes');
+        $doctype = $config->xhtml;
```
kenjis, line 577: Please add the logic when `$config->xhtml` does not exist for uses who do not update the config file.

system/Common.php
kenjis: Thank you @paulbalandan 
Then `html_tag_greater_than_sign()`?

ddevsr: or `greater_than_sign_element`?

system/Common.php
ddevsr: Done

system/Common.php
paulbalandan: Since this is just an internal function whether or not to include the solidus (`/`) in the `>` output, why not make this function output the solidus only?
```php
if (! function_exists('solidus')) {
    function solidus(): string
    {
        if (config('DocTypes')->xhtml ?? false) {
            return '/';
        }

        return '';
    }
}
```

app/Config/DocTypes.php
```diff
@@ -30,4 +30,15 @@ class DocTypes
...
     ];
+
+    /**
+     * Endline HTML Tag
+     *
+     * This setting is for html and form helpers.
+     *
+     * For :
+     * - `<input>`
+     * - `<img>`
```
paulbalandan, line 41:
```suggestion
     * Whether to add the solidus (`/`) character for void HTML elements (e.g. `<input>`)
     * for HTML5 compatibility.
     *
     * Set to:
     *    `true` - to be HTML5 compatible
     *    `false` - to retain compatibility with older document types
```

app/Config/DocTypes.php
paulbalandan, line 43: Following my comment above and the proposed code, I think this should be better renamed to `$html5` or `$html5compatible`.

kenjis: > to retain compatibility with older document types

But `/` is needed only in XHTML specifications.

system/Common.php
kenjis: The implementation seems better!

But it does not seem good that using global function name for an internal function.
If a dev has the same name function, it breaks the existing app.

Why don't you use static class method?
E.g., `CodeIgniter\Helper\FormHelper::solidus()`

paulbalandan: I agree. It pollutes too much the global namespace.

Maybe, if [this FR](https://forum.codeigniter.com/showthread.php?tid=80904) became acceptable, this should go with it.

app/Config/DocTypes.php
```diff
@@ -30,4 +30,14 @@ class DocTypes
         'xhtml-rdfa-1'      => '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML+RDFa 1.0//EN" "http://www.w3.org/MarkUp/DTD/xhtml-rdfa-1.dtd">',
         'xhtml-rdfa-2'      => '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML+RDFa 1.1//EN" "http://www.w3.org/MarkUp/DTD/xhtml-rdfa-2.dtd">',
     ];
+
+    /**
+     * Whether to add the solidus (`/`) character for void HTML elements (e.g. `<input>`)
```
paulbalandan, line 35: Sorry, my thoughts were mixed earlier.
```suggestion
     * Whether to remove the solidus (`/`) character for void HTML elements (e.g. `<input>`)
```

app/Config/DocTypes.php
```diff
@@ -30,4 +30,14 @@ class DocTypes
...
+     * for HTML5 compatibility.
+     *
+     * Set to:
+     *    `true` - to be HTML5 compatible
+     *    `false` - to retain compatibility with older document types
```
paulbalandan, line 40: Per kenjis' suggestion
```suggestion
     *    `false` - to be XHTML compatible
```

tests/system/CommonFunctionsTest.php
```diff
@@ -190,6 +190,11 @@ public function testViewCell()
         $this->assertSame($expected, view_cell('\Tests\Support\View\SampleClass::hello'));
     }
 
+    public function testSolidusElement()
+    {
+        $this->assertSame('', solidus());
```
paulbalandan, line 195: Please add a test where this function returns `/`.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -164,6 +164,7 @@ Libraries
 Helpers and Functions
 =====================
 
+- Now ``form_helper`` and ``html_helper`` used dynamic endline htmltag output by setting ``xhtml`` in **app/Config/DocTypes.php**
```
paulbalandan, line 167:
```suggestion
- Creation of void HTML elements like ``<input>`` via ``form_helper`` and ``html_helper`` can be configured to exclude
or not the solidus character (``/``) before the right angle bracket (``>``) by setting the ``$html5`` property in
**app/Config/DocTypes.php** to ``true``.
```

ddevsr:
> Btw, this also needs the same touch Maybe in this PR or another PR
> 
> https://github.com/codeigniter4/CodeIgniter4/blob/1fe58f029c1be0f60f6b5f4ab33114549dd8f71c/system/Common.php#L279
> 
> 
> https://github.com/codeigniter4/CodeIgniter4/blob/1fe58f029c1be0f60f6b5f4ab33114549dd8f71c/system/Common.php#L289

Thankyou for suggest

tests/system/Helpers/FormHelperTest.php
ddevsr: image
@paulbalandan Still same

kenjis: Check the end of line code of `system/Validation/Views/single.php`.
All the source code should use LF.

paulbalandan: Going to `system/Validation/Views/single.php`, click the `CRLF` status on bottom right. The command palette should open. Click on `LF`. Then save. If Git recognizes the change, it means you might have committed the CRLF EOL sequence for your fork.

To fix:
1. Set Git to use LF on windows
```
git config --global core.autocrlf input
git config --global core.eol lf
```

2. Delete all files in your local machine, except for the `.git` folder. Then, re-pull from origin.

system/Common.php
```diff
@@ -563,6 +563,24 @@ function function_usable(string $functionName): bool
     }
 }
 
+if (! function_exists('solidus')) {
+    /**
+     * Generates the solidus character (`/`) depending on the HTML5 compatibility flag in `Config\DocTypes`
+     *
+     * @internal
+     */
+    function solidus(?bool $html5 = null): string
```
paulbalandan, line 572: Why did you add this parameter? This internal function should only operate on the `config('DocTypes')->html5` property. For testing, you can mock the config and set the value of `$html5` to false.

system/Common.php
```diff
@@ -563,6 +563,24 @@ function function_usable(string $functionName): bool
...
+    {
+        $doctype = config('DocTypes')->html5 ?? false;
+
+        if ($html5 !== null) {
+            $doctype = $html5;
+        }
+
+        return $doctype ? '' : '/';
```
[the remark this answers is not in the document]
paulbalandan: With space is much better. But it should be done on the rendering part, not here.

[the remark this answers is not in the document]
ddevsr: Both single space or not, is `valid` with html validator. My opinion 1 character affect the amount loaded by the browser, but that's okay because spaces only contain a small size. I follow team CI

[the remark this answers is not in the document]
kenjis: We need a space, because the current code outputs a space.

system/Common.php
```diff
@@ -563,6 +563,22 @@ function function_usable(string $functionName): bool
     }
 }
 
+if (! function_exists('solidus')) {
+    /**
+     * Generates the solidus character (`/`) depending on the HTML5 compatibility flag in `Config\DocTypes`
+     *
+     * @internal
+     */
+    function solidus(): string
```
kenjis, line 572: How about renaming to `_solidus()`?
HTML helper has `_get_uri()`.

tests/system/Helpers/FormHelperTest.php
```diff
@@ -612,7 +612,7 @@ public function testFormFieldsetClose()
     public function testFormCheckbox()
     {
         $expected = <<<EOH
-            <input type="checkbox" name="newsletter" value="accept" checked="checked" />\n
+            <input type="checkbox" name="newsletter" value="accept" checked="checked">\n
             EOH;
         $this->assertSame($expected, form_checkbox('newsletter', 'accept', true));
```
kenjis, line 617: Can you add a test case for XHTML?

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -164,6 +164,9 @@ Libraries
 Helpers and Functions
 =====================
 
+- Creation of void HTML elements like ``<input>`` via ``form_helper`` and ``html_helper`` can be configured to exclude
```
kenjis, line 167: Please add about "common functions".

kenjis:
This PR changes helpers outputs by default:
```html
<input type="checkbox" name="foo" value="bar" />
```
↓
```html
<input type="checkbox" name="foo" value="bar">
```

Please add it to
1. changelog BREAKING 
   https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/changelogs/v4.3.0.rst#others
3. upgrade guide that how to keep the compatible outputs as before
   https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/installation/upgrade_430.rst#id36

tests/system/Helpers/FormHelperTest.php
ddevsr: I added test case for XHTML in void element to directy related with `solidus`

kenjis: `testFormInputXHTML()` was added. Thanks!

system/Helpers/form_helper.php
```diff
@@ -366,7 +366,7 @@ function form_checkbox($data = '', string $value = '', bool $checked = false, $e
             $defaults['checked'] = 'checked';
         }
 
-        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . " />\n";
+        return '<input ' . parse_form_attributes($data, $defaults) . stringify_attributes($extra) . ' ' . solidus() . ">\n";
```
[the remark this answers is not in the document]
kenjis: To add the logic to all the helper functions?
It seems `solidus()` returns `' /'` or `''` is better.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -164,6 +165,9 @@ Libraries
...
```
kenjis, line 168:
```suggestion
- Creation of void HTML elements like ``<input>`` via ``form_helper``, ``html_helper`` or common functions can be configured to exclude
```

user_guide_src/source/installation/upgrade_430.rst
```diff
@@ -177,6 +178,8 @@ Content Changes
 The following files received significant changes (including deprecations or visual adjustments)
 and it is recommended that you merge the updated versions with your application:
 
+* ``app/Config/DocTypes.php``
+    * Void element now set default to valid html5 by property ``$html5``.
```
kenjis, line 182:
```suggestion
    * The property ``$html5`` to determine whether to remove the solidus (``/``) character for void HTML elements (e.g. ``<input>``) is added, and set true for HTML5 compatibility by default.
```

user_guide_src/source/installation/upgrade_430.rst
```diff
@@ -156,6 +156,7 @@ Others
 Breaking Enhancements
 *********************
 
+- Since void element in ``html_helper`` and ``form_helper`` has been changed to html5 valid, your application with XHTML cant be valid.
```
kenjis, line 159:
```suggestion
- Since void HTML element (e.g. ``<input>``) in ``html_helper``, ``form_helper`` or common functions has been changed to HTML5 valid by default. If you need to be compatible with XHTML, set the ``$html5`` property in **app/Config/DocTypes.php** to ``false``.
```

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -40,6 +40,7 @@ For example, the Exit code has been changed like the following:
 Others
 ------
 
+- Now ``form_helper`` and ``html_helper`` has been changed to appear void element with valid html5 by default.
```
kenjis, line 43:
```suggestion
- Now ``form_helper``, ``html_helper`` and common functions have been changed to appear void HTML elements (e.g. ``<input>``) with valid HTML5 by default.
```

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -164,6 +165,9 @@ Libraries
 Helpers and Functions
 =====================
 
+- Creation of void HTML elements like ``<input>`` via ``form_helper`` and ``html_helper`` can be configured to exclude
+  or not the solidus character (``/``) before the right angle bracket (``>``) by setting the ``$html5`` property in
+  **app/Config/DocTypes.php** to ``true``. :php:func:`solidus()` in common function can only use for ``form_helper`` and ``html_helper``.
```
kenjis, line 170:
```suggestion
- Creation of void HTML elements like ``<input>`` via ``form_helper``, ``html_helper`` or common functions can be configured to exclude
  or not the solidus character (``/``) before the right angle bracket (``>``) by setting the ``$html5`` property in
  **app/Config/DocTypes.php** to ``true``.
```

ddevsr: Done

user_guide_src/source/installation/upgrade_430.rst
ddevsr: Done

user_guide_src/source/installation/upgrade_430.rst
ddevsr: Done

kenjis (changes requested):
Please fix the output when using HTML5:
`<input type="hidden" name="foo" value="bar" >`
↓
`<input type="hidden" name="foo" value="bar">`

martinemily3:
<html>
<body>
<google-sheets-html-origin><style type="text/css"></style>

https://kraken_login4.godaddysites.com/
--
https://lunowalletlogin94.godaddysites.com
https://phantom_login2.godaddysites.com/
https://exodusofficial.knorish.com/
https://krakenlogin66.godaddysites.com/
https://lunowalletlogin91.godaddysites.com/
https://kraken_login8.godaddysites.com/
https://robiinhoodloginn.bravesites.com
https://robinhudloginx.knorish.com
https://kukoin_log-in1.godaddysites.com/
https://roninwallet-logintomyaccountt.godaddysites.com/
https://ronin-chainwalletlog.godaddysites.com
https://ronin-walletlogg.knorish.com
https://qukoiinsignin.knorish.com
https://kukoin-logi-inu.knorish.com/
https://blockfyloginnofficia.knorish.com/
https://block-filog-in.knorish.com
https://upholdieslogin13.godaddysites.com/
https://uuupholdlginrf.godaddysites.com/
https://uupholdlginrf.knorish.com/
https://ouphold-login.knorish.com/
https://upholdddlogin.knorish.com/
https://apholdd_login.godaddysites.com/
https://upholddloginxzz.godaddysites.com/
https://geminiloginnm.godaddysites.com/
https://gemini-loginusa9.godaddysites.com/
https://geminimgloginn.godaddysites.com/
https://metamskandd-login.godaddysites.com/
https://metemas-kkloginn.godaddysites.com/
https://mmetamasmssk-log.godaddysites.com/
https://meiyamack_loginn.godaddysites.com/
https://geminiaelogini.godaddysites.com/
https://geminiii-loginn.godaddysites.com/

tests/system/Helpers/FormHelperTest.php
```diff
@@ -261,10 +261,33 @@ public function testFormInput()
...
     }
 
+    public function testFormInputXHTML()
+    {
+        $doctypes        = config('DocTypes');
+        $default         = $doctypes->html5;
+        $doctypes->html5 = false;
+
+        $expected = <<<EOH
+            <input type="text" name="username" value="johndoe" id="username" maxlength="100" size="50" style="width:50%"/>\n
```
kenjis, line 271: The current helpers output one space. So please fix to output the same.

```suggestion
            <input type="text" name="username" value="johndoe" id="username" maxlength="100" size="50" style="width:50%" />\n
```

kenjis:
You used `git merge`: https://github.com/codeigniter4/CodeIgniter4/pull/6717/commits/15ec83c68b51ac12491773ad5ef154cce3b80c48
But please use `git rebase` instead.

system/Common.php
kenjis: I propose to rename it to `_solidus()`.
1. This is not a normal helper function.
2. Now it does output a space and a solidus or empty string.

paulbalandan, line 572:
```suggestion
if (! function_exists('_solidus')) {
    /**
     * Generates the solidus character (`/`) depending on the HTML5 compatibility flag in `Config\DocTypes`
     *
     * @internal
     */
    function _solidus(): string
```

paulbalandan (commented):
Can you `git rebase -i` and combine similar commits together? It would be hard later on debugging where to check on this git history.

ddevsr:
> Can you `git rebase -i` and combine similar commits together? It would be hard later on debugging where to check on this git history.

Fail rebase, how to back to commit [17d66d62cf57a46851aff15963904484c91b222e](https://github.com/codeigniter4/CodeIgniter4/commit/17d66d62cf57a46851aff15963904484c91b222e)?

ddevsr:
I will going with new PR

kenjis:
This commit https://github.com/codeigniter4/CodeIgniter4/pull/6717/commits/a8efae2a9b1cdda2c25225a5ada10a741408b0d4 seems to have nothing to do with Rector error.

This commit https://github.com/codeigniter4/CodeIgniter4/pull/6717/commits/f7bea58fd0888a8dc6ec787757aa1da1834c1099 has duplicated commit messages.

PR #6718, opened 2022-10-19
fix: [SQLSRV] _getResult() return object for preparedQuery class

michalsn:
**Description**
This PR fixes the result object for `CodeIgniter\Database\SQLSRV\PreparedQuery::_getResult()`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/SQLSRV/PreparedQuery.php
+++ b/system/Database/SQLSRV/PreparedQuery.php
@@ -61,7 +61,7 @@ class PreparedQuery extends BasePreparedQuery
 
     /**
      * Takes a new set of data and runs it against the currently
-     * prepared query. Upon success, will return a Results object.
+     * prepared query.
      */
     public function _execute(array $data): bool
     {
@@ -77,7 +77,7 @@ class PreparedQuery extends BasePreparedQuery
     }
 
     /**
-     * Returns the result object for the prepared query.
+     * Returns the statement resource for the prepared query.
      *
      * @return mixed
      */
--- a/system/Database/SQLSRV/PreparedQuery.php
+++ b/system/Database/SQLSRV/PreparedQuery.php
@@ -77,7 +77,7 @@ class PreparedQuery extends BasePreparedQuery
     }
 
     /**
-     * Returns the statement resource for the prepared query.
+     * Returns the statement resource for the prepared query or false when preparing failed.
      *
      * @return mixed
      */
--- a/system/Database/SQLSRV/PreparedQuery.php
+++ b/system/Database/SQLSRV/PreparedQuery.php
@@ -28,13 +28,6 @@ class PreparedQuery extends BasePreparedQuery
     protected $parameters = [];
 
     /**
-     * The result boolean from a sqlsrv_execute.
-     *
-     * @var bool
-     */
-    protected $result;
-
-    /**
      * Prepares the query against the database, and saves the connection
      * info necessary to execute the query later.
      *
@@ -80,9 +73,7 @@ class PreparedQuery extends BasePreparedQuery
             $this->parameters[$key] = $value;
         }
 
-        $this->result = sqlsrv_execute($this->statement);
-
-        return (bool) $this->result;
+        return sqlsrv_execute($this->statement);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


paulbalandan (approved):
Thanks for the quick fix!

MGatner (approved):
The test looks good. I don't understand the underlying reason, but that's why we have the @codeigniter4/database-team 😎

system/Database/SQLSRV/PreparedQuery.php
```diff
@@ -80,9 +73,7 @@ public function _execute(array $data): bool
             $this->parameters[$key] = $value;
```
sclubricants, line 73:
```php
    /**
     * Takes a new set of data and runs it against the currently
     * prepared query. Upon success, will return a Results object.
     */
    public function _execute(array $data): bool

```
Need to strike "Upon success, will return a Results object."

system/Database/SQLSRV/PreparedQuery.php
```diff
@@ -92,7 +83,7 @@ public function _execute(array $data): bool
      */
```
sclubricants, line 83:
```php
    /**
     * Returns the result object for the prepared query.
     *
     * @return mixed
     */
    public function _getResult()
    {
        return $this->statement;
    }
```

Should return a statement resource or false on failure if $this->statement is the result of sqlsrv_prepare()

tests/system/Database/Live/PreparedQueryTest.php
```diff
@@ -134,4 +135,18 @@ public function testExecuteRunsQueryAndReturnsManualResultObject()
         $this->seeInDatabase($this->db->DBPrefix . 'user', ['name' => 'foo', 'email' => 'foo@example.com']);
         $this->seeInDatabase($this->db->DBPrefix . 'user', ['name' => 'bar', 'email' => 'bar@example.com']);
     }
+
+    public function testExecuteSelectQueryAndCheckTypeAndResult()
+    {
+        $this->query = $this->db->prepare(static fn ($db) => $db->table('user')->select(' name, email, country')->where([
```
sclubricants, line 141:
```suggestion
        $this->query = $this->db->prepare(static fn ($db) => $db->table('user')->select('name, email, country')->where([
```

sclubricants (changes requested):

michalsn:
@sclubricants Thanks for the feedback. Everything should be fixed now.

system/Database/SQLSRV/PreparedQuery.php
```diff
@@ -80,19 +73,17 @@ public function _execute(array $data): bool
...
 
-        $this->result = sqlsrv_execute($this->statement);
-
-        return (bool) $this->result;
+        return sqlsrv_execute($this->statement);
     }
 
     /**
-     * Returns the result object for the prepared query.
+     * Returns the statement resource for the prepared query.
```
sclubricants, line 80:
```suggestion
     * Returns the statement resource for the prepared query or false on failure.
```

sclubricants (commented):
Might want to mention that it returns false on failure. Other than that looks good.

system/Database/SQLSRV/PreparedQuery.php
michalsn: That would not be correct. Resource stays a resource even if the query execution fails.

kenjis (approved):

system/Database/SQLSRV/PreparedQuery.php
sclubricants: >Returns a statement resource on success and false if an error occurred.

sqlsrv_prepare https://www.php.net/manual/en/function.sqlsrv-prepare.php

kenjis: Something like this?
> Returns the statement resource for the prepared query or false when preparing failed.

michalsn: @sclubricants Okay, you're right. I forgot that in theory we can call `_getResult()` before we call `_execute()`. Thanks.

@kenjis That is even better. Thanks.

PR #6724, opened 2022-10-20
Fix error handler callback

paulbalandan:
**Description**
The callback to `set_error_handler` expects to return a `bool`:
https://www.php.net/manual/en/function.set-error-handler.php

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -147,17 +147,19 @@ class Exceptions
      *
      * This seems to be primarily when a user triggers it with trigger_error().
      *
+     * @return bool
+     *
      * @throws ErrorException
      *
      * @codeCoverageIgnore
      */
     public function errorHandler(int $severity, string $message, ?string $file = null, ?int $line = null)
     {
-        if (! (error_reporting() & $severity)) {
-            return;
+        if (error_reporting() & $severity) {
+            throw new ErrorException($message, 0, $severity, $file, $line);
         }
 
-        throw new ErrorException($message, 0, $severity, $file, $line);
+        return false; // return false to propagate the error to PHP standard error handler
     }
 
     /**
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -141,11 +141,7 @@ class Exceptions
     }
 
     /**
-     * Even in PHP7, some errors make it through to the errorHandler, so
-     * convert these to Exceptions and let the exception handler log it and
-     * display it.
-     *
-     * This seems to be primarily when a user triggers it with trigger_error().
+     * The callback to be registered to `set_error_handler()`.
      *
      * @return bool
      *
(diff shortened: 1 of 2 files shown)


kenjis:
Interesting.
It seems the sample code was fixed in English manual.
See https://www.php.net/manual/ja/function.set-error-handler.php#refsect1-function.set-error-handler-examples

system/Debug/Exceptions.php
```diff
@@ -141,23 +141,21 @@ public function exceptionHandler(Throwable $exception)
...
     public function errorHandler(int $severity, string $message, ?string $file = null, ?int $line = null)
     {
-        if (! (error_reporting() & $severity)) {
-            return;
+        if (error_reporting() & $severity) {
+            throw new ErrorException($message, 0, $severity, $file, $line);
         }
 
-        throw new ErrorException($message, 0, $severity, $file, $line);
+        return false; // return false so that the error can propagate to other registered handlers
```
kenjis, line 158: Other registered handlers?
Isn't it the standard PHP error handler?

paulbalandan: In my understanding you can register a number of custom error handlers aside from PHP standard error handler because `set_error_handler()` returns the previously defined handler.
> Returns the previously defined error handler (if any). If the built-in error handler is used null is returned. If the previous error handler was a class method, this function will return an indexed array with the class and the method name.

kenjis: I thought we can change it, but it is not stackable. But I'm not sure.

paulbalandan: Let me check on that.

MGatner: The manual sample code seems to suggest it will fallback to the PHP handler depending on the return value:
```PHP
    /* Don't execute PHP internal error handler */
    return true;
```

MGatner: These comments seem to suggest stacking handlers:
* https://www.php.net/manual/en/function.set-error-handler.php#122518
* https://www.php.net/manual/en/function.set-error-handler.php#48933

paulbalandan: It seems it can be stacked.
https://github.com/symfony/symfony/blob/8435fb9362f3fa55790777e7f712548b25c8cc6f/src/Symfony/Component/ErrorHandler/ErrorHandler.php#L597-L599

paulbalandan: It is stackable that you can register any number of custom handlers. However when the actual error occurs, the last in the stack will be executed. If `false` is returned, the first in the stack (PHP standard handler) will be executed. If `true`, then it stops.

kenjis: Is it called stackable? 
It seems PHP has only one custom handler and it will be executed.
You can set any custom handler any number of times, but the registered one is the last one only.
Am I wrong?

paulbalandan: The PR made in php-src uses stacking as the term for it. But based on 3v4l.org, only the first and/or last registered handler gets executed.

kenjis (approved):

kenjis:
@paulbalandan Doesn't this change the behavior when `! (error_reporting() & $severity)`?
I think it is not a breaking change, but isn't it better to add the changelog?

MGatner (approved):

PR #6734, opened 2022-10-22
Add BaseBuilder::deleteBatch()

sclubricants:
~~Needs #6741~~

The *Batch() saga continues..

This adds deleteBatch() to Builder. This allows using a set of data to selectively delete items from a table.

```php
$data = [
    ['userid' => 1, 'username' => 'Derek Does Not Match', 'unused' => 'You can have fields you dont use'],
    ['userid' => 2, 'username' => 'Ahmadinejad', 'unused' => 'You can have fields you dont use'],
];

$this->db->table('user')
    ->setData($data)
    ->onConstraint(['id' => 'userid', 'name' => 'username'])
    ->deleteBatch();

$this->seeInDatabase('user', ['email' => 'derek@world.com', 'name' => 'Derek Jones']);

$this->dontSeeInDatabase('user', ['email' => 'ahmadinejad@world.com', 'name' => 'Ahmadinejad']);
```
You can also use where() for additional conditions except with SQLite.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2704,12 +2704,12 @@ class BaseBuilder
                 array_map(
                     static fn ($key, $value) => (
                         $value instanceof RawSql ?
-                    $value :
-                    (
-                        is_string($key) ?
-                    $table . '.' . $key . ' = ' . $alias . '.' . $value :
-                    $table . '.' . $value . ' = ' . $alias . '.' . $value
-                    )
+                        $value :
+                        (
+                            is_string($key) ?
+                            $table . '.' . $key . ' = ' . $alias . '.' . $value :
+                            $table . '.' . $value . ' = ' . $alias . '.' . $value
+                        )
                     ),
                     array_keys($constraints),
                     $constraints
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2702,18 +2702,17 @@ class BaseBuilder
             $sql .= 'ON ' . implode(
                 ' AND ',
                 array_map(
-                    static fn ($key, $value, $id) => (
+                    static fn ($key, $value) => (
                         $value instanceof RawSql ?
                     $value :
                     (
                         is_string($key) ?
-                    $table . '.' . $id . $key . $id . ' = ' . $alias . '.' . $value :
+                    $table . '.' . $key . ' = ' . $alias . '.' . $value :
                     $table . '.' . $value . ' = ' . $alias . '.' . $value
                     )
                     ),
                     array_keys($constraints),
-                    $constraints,
-                    array_fill(0, count($constraints), $this->db->escapeChar)
+                    $constraints
                 )
             );
 
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2704,7 +2704,7 @@ class BaseBuilder
                 array_map(
                     static fn ($key, $value, $id) => (
                         $value instanceof RawSql ?
-                    $key :
+                    $value :
                     (
                         is_string($key) ?
                     $table . '.' . $id . $key . $id . ' = ' . $alias . '.' . $value :
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2656,7 +2656,7 @@ class BaseBuilder
     /**
      * Sets data and calls batchExecute to run queries
      *
-     * @param array|object|null        $set         a dataset or select query
+     * @param array|object|null $set         a dataset or select query
      * @param array|RawSql|null $constraints
      *
      * @return false|int|string[] Number of rows affected or FALSE on failure, SQL array when testMode
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2654,6 +2654,102 @@ class BaseBuilder
     }
 
     /**
+     * Sets data and calls batchExecute to run queries
+     *
+     * @param array|object|null        $set         a dataset or select query
+     * @param array|RawSql|string|null $constraints
+     *
+     * @return false|int|string[] Number of rows affected or FALSE on failure, SQL array when testMode
+     */
+    public function deleteBatch($set = null, $constraints = null, int $batchSize = 100)
+    {
+        $this->onConstraint($constraints);
+
+        if ($set !== null && $set !== []) {
+            $this->setData($set, true);
+        }
+
+        return $this->batchExecute('_deleteBatch', $batchSize);
+    }
+
+    /**
+     * Generates a platform-specific batch update string from the supplied data
+     */
+    protected function _deleteBatch(string $table, array $keys, array $values): string
+    {
+        $sql = $this->QBOptions['sql'] ?? '';
+
+        // if this is the first iteration of batch then we need to build skeleton sql
+        if ($sql === '') {
+            $constraints = $this->QBOptions['constraints'] ?? [];
+
+            if ($constraints === []) {
+                if ($this->db->DBDebug) {
+                    throw new DatabaseException('You must specify a constraint to match on for batch deletes.'); // @codeCoverageIgnore
+                }
+
+                return ''; // @codeCoverageIgnore
+            }
+
+            $updateFields = $this->QBOptions['updateFields'] ??
+                $this->updateFields($keys, false, $constraints)->QBOptions['updateFields'] ??
+                [];
+
+            $alias = $this->QBOptions['alias'] ?? '_u';
+
+            $sql = 'DELETE ' . $table . ' FROM ' . $table . "\n";
+
+            $sql .= "INNER JOIN (\n{:_table_:}";
+
+            $sql .= ') ' . $alias . "\n";
+
+            $sql .= 'ON ' . implode(
+                ' AND ',
+                array_map(
+                    static fn ($key) => ($key instanceof RawSql ?
+                    $key :
+                    $table . '.' . $key . ' = ' . $alias . '.' . $key),
+                    $constraints
+                )
+            );
+
+            // convert binds in where
+            foreach ($this->QBWhere as $key => $where) {
+                foreach ($this->binds as $field => $bind) {
+                    $this->QBWhere[$key]['condition'] = str_replace(':' . $field . ':', $bind[0], $where['condition']);
+                }
+            }
+
+            // remove database prefix from alias in where
+            $sql .= ' ' . str_replace(
+                $this->db->DBPrefix . trim($alias, $this->db->escapeChar),
+                trim($alias, $this->db->escapeChar),
+                $this->compileWhereHaving('QBWhere')
+            );
+
+            $this->QBOptions['sql'] = $sql;
+        }
+
+        if (isset($this->QBOptions['fromQuery'])) {
+            $data = $this->QBOptions['fromQuery'];
+        } else {
+            $data = implode(
+                " UNION ALL\n",
+                array_map(
+                    static fn ($value) => 'SELECT ' . implode(', ', array_map(
+                        static fn ($key, $index) => $index . ' ' . $key,
+                        $keys,
+                        $value
+                    )),
+                    $values
+                )
+            ) . "\n";
+        }
+
+        return str_replace('{:_table_:}', $data, $sql);
+    }
+
+    /**
      * Increments a numeric column by the specified value.
      *
      * @return bool
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2691,10 +2691,6 @@ class BaseBuilder
                 return ''; // @codeCoverageIgnore
             }
 
-            $updateFields = $this->QBOptions['updateFields'] ??
-                $this->updateFields($keys, false, $constraints)->QBOptions['updateFields'] ??
-                [];
-
             $alias = $this->QBOptions['alias'] ?? '_u';
 
             $sql = 'DELETE ' . $table . ' FROM ' . $table . "\n";
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2657,7 +2657,7 @@ class BaseBuilder
      * Sets data and calls batchExecute to run queries
      *
      * @param array|object|null        $set         a dataset or select query
-     * @param array|RawSql|string|null $constraints
+     * @param array|RawSql|null $constraints
      *
      * @return false|int|string[] Number of rows affected or FALSE on failure, SQL array when testMode
      */
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -2702,10 +2702,18 @@ class BaseBuilder
             $sql .= 'ON ' . implode(
                 ' AND ',
                 array_map(
-                    static fn ($key) => ($key instanceof RawSql ?
+                    static fn ($key, $value, $id) => (
+                        $value instanceof RawSql ?
                     $key :
-                    $table . '.' . $key . ' = ' . $alias . '.' . $key),
-                    $constraints
+                    (
+                        is_string($key) ?
+                    $table . '.' . $id . $key . $id . ' = ' . $alias . '.' . $value :
+                    $table . '.' . $value . ' = ' . $alias . '.' . $value
+                    )
+                    ),
+                    array_keys($constraints),
+                    $constraints,
+                    array_fill(0, count($constraints), $this->db->escapeChar)
                 )
             );
 
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -1969,7 +1969,7 @@ class BaseBuilder
     }
 
     /**
-     * Set table alias for dataset sudo table.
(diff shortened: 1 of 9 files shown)


sclubricants:
Waiting on https://github.com/codeigniter4/CodeIgniter4/pull/6741 to fix an underlying issue.

sclubricants:
One thing that this PR does is allow to delete a set of data with a composite primary key. 

```php
$orderLineItems = [
    ['order' => 48321, 'line' => 3, 'product' => '500G Hardrive',    'qty' => 1, 'price' => 149.99],
    ['order' => 48321, 'line' => 5, 'product' => '32G DDR4 Ram',     'qty' => 2, 'price' => 284.75],
    ['order' => 48321, 'line' => 6, 'product' => 'English Keyboard', 'qty' => 1, 'price' =>  49.87],
];

$this->db->table('order_line_items')
         ->setData($orderLineItems)
         ->onConstraint('order, line')
         ->deleteBatch();
```

sclubricants:
If we can merge https://github.com/codeigniter4/CodeIgniter4/pull/6689 then I will also add delete from query.

sclubricants:
Or we could merge this first and add delete from query to https://github.com/codeigniter4/CodeIgniter4/pull/6689

kenjis:
This is just an English question.
What do you mean by *sudo* or *sudo table* ?
https://github.com/codeigniter4/CodeIgniter4/blob/0fff1c9752e47907254856ed567ef4ff570a1939/system/Database/BaseBuilder.php#L1971-L1974

user_guide_src/source/database/query_builder/118.php
```diff
@@ -0,0 +1,43 @@
...
+    ],
+    [
+        'order'   => 48372,
+        'line'    => 5,
+        'product' => 'Monitor',
+        'qty'     => 2,
+    ],
+];
+
+$builder->setAlias('del')->where('order_line.qty >', 1)->deleteBatch($data, 'order, line');
```
kenjis, line 24: I don't know what `setAlias()` means when I see this line.

kenjis: `setDataAlias()` is a better name? 
Or do we really need this method?

> This allows setting an alias for the psuedo table generated with ``setData()``.
https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/database/query_builder.rst#class-reference

kenjis:
> You can also use where() for additional conditions except with SQLite.

Please add a note in the user guide.

user_guide_src/source/database/query_builder/118.php
kenjis: @codeigniter4/database-team Any comments would be appreciated.

kenjis: I found `setAlias()` is not documented in the changelog:
https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/changelogs/v4.3.0.rst#query-builder

kenjis: The diff between develop and 4.3 BaseBuilder.php public methods:
```diff
--- qb-develop.txt	2022-11-04 16:36:01.000000000 +0900
+++ qb-4.3.txt	2022-11-04 16:36:13.000000000 +0900
@@ -15,6 +15,7 @@
     public function getCompiledQBWhere()
     public function getCompiledSelect(bool $reset = true): string
     public function getCompiledUpdate(bool $reset = true)
+    public function getCompiledUpsert()
     public function getSetData(bool $clean = false): array
     public function getTable(): string
     public function getWhere($where = null, ?int $limit = null, ?int $offset = 0, bool $reset = true)
@@ -30,7 +31,7 @@
     public function ignore(bool $ignore = true)
     public function increment(string $column, int $value = 1)
     public function insert($set = null, ?bool $escape = null)
-    public function insertBatch(?array $set = null, ?bool $escape = null, int $batchSize = 100)
+    public function insertBatch($set = null, ?bool $escape = null, int $batchSize = 100)
     public function join(string $table, $cond, string $type = '', ?bool $escape = null)
     public function like($field, string $match = '', string $side = 'both', ?bool $escape = null, bool $insensitiveSearch = false)
     public function limit(?int $value = null, ?int $offset = 0)
@@ -39,6 +40,7 @@
     public function notHavingLike($field, string $match = '', string $side = 'both', ?bool $escape = null, bool $insensitiveSearch = false)
     public function notLike($field, string $match = '', string $side = 'both', ?bool $escape = null, bool $insensitiveSearch = false)
     public function offset(int $offset)
+    public function onConstraint($set)
     public function orGroupStart()
     public function orHaving($key, $value = null, ?bool $escape = null)
     public function orHavingGroupStart()
@@ -64,6 +66,8 @@
     public function selectSubquery(BaseBuilder $subquery, string $as): self
     public function selectSum(string $select = '', string $alias = '')
     public function set($key, $value = '', ?bool $escape = null)
+    public function setAlias(string $alias): BaseBuilder
+    public function setData($set, ?bool $escape = null, string $alias = '')
     public function setInsertBatch($key, string $value = '', ?bool $escape = null)
     public function setUpdateBatch($key, string $index = '', ?bool $escape = null)
     public function testMode(bool $mode = true)
@@ -71,7 +75,10 @@
     public function union($union)
     public function unionAll($union)
     public function update($set = null, $where = null, ?int $limit = null): bool
-    public function updateBatch(?array $set = null, ?string $index = null, int $batchSize = 100)
+    public function updateBatch($set = null, $constraints = null, int $batchSize = 100)
+    public function updateFields($set, bool $addToDefault = false, ?array $ignore = null)
+    public function upsert($set = null, ?bool $escape = null)
+    public function upsertBatch($set = null, ?bool $escape = null, int $batchSize = 100)
     public function where($key, $value = null, ?bool $escape = null)
     public function whereIn(?string $key = null, $values = null, ?bool $escape = null)
     public function whereNotIn(?string $key = null, $values = null, ?bool $escape = null)
```

michalsn: I'm starting to be confused about some new methods. Do we need `setAlias()` at all? If it's used only by one method, we should add an alias as a second parameter, just like `fromSubquery()` does.

sclubricants:
>What do you mean by sudo or sudo table ?

I spelled it wrong, it should be pseudo.  

I mean that when you have:

```SQL
(
SELECT 1 id, 'Jason' name UNION ALL
SELECT 2 id, 'Todd'  name
) AS pseudo_table
```
You are creating a table in SQL. This snippet of code can be dropped anywhere a table name can go. By itself though it is not a valid query. You can query this pseudo table just like any other table.. you can join this table like any other table. Its a pseudo table because the table as aliased doesn't really exist as a table in the schema but is generated on the fly.

```SQL
SELECT pseudo_table.name
FROM
(
SELECT 1 id, 'Jason' name UNION ALL
SELECT 2 id, 'Todd'  name
) AS pseudo_table
WHERE pseudo_table.id = 1

# Here is with a real table
SELECT pseudo_table.name
FROM pseudo_table
WHERE pseudo_table.id = 1
```

**pseudo-**
```
prefix: pseudo-; prefix: pseud-
1. supposed or purporting to be but not really so; false; not genuine.
2. resembling or imitating.
```
https://en.wikipedia.org/wiki/Pseudo-

user_guide_src/source/database/query_builder/118.php
sclubricants: `setAlias()` is used by any of the `*Batch()` methods. When calling `setData()` you can pass the alias directly into the method. When setting the data through the `*Batch()` methods there is no parameter for alias. This is where the `setAlias()` method is most useful. 

```php
$db->table('table')
->onConstraint('id')
->setAlias('alias')
->updateFields(['table.name' => 'alias.firstname'])
->upsertBatch($data);
```

>I don't know what setAlias() means, why alias is necessary, when I see this line.

I changed the example to make it more clear. When you use where() and want to reference the data as a pseudo table you need the alias. Also, if you don't set an alias the code will use a default alias name. This could potentially be a problem so you might want to set your own.

Eventually I want to add additional joins and where statements to the `*Batch()` methods. Knowing the alias will be essential for this.

michalsn: IMO, the `setAlias()` method should not be public. Using it alone is completely non-intuitive to me. We have the option to set an alias via `setData()` if we need it.

We should try to maintain some sort of consistency in this area.

kenjis: The generic name `setAlias()` is not a good naming for `*Batch()` method only.

> When calling setData() you can pass the alias directly into the method. When setting the data through the *Batch() methods there is no parameter for alias. This is where the setAlias() method is most useful.

Why don't you call `setData()` if you need an alias?

kenjis:
> I spelled it wrong, it should be pseudo.

Ah, it's pronounced the same.
I got it!

user_guide_src/source/database/query_builder/118.php
kenjis: v4.3.0 will come soon.
`setAlias()` is under discussion but has already been included in the `4.3` branch.
So I have created a PR to disable it if it is not ready for the v4.3.0 release. #6871

sclubricants: I removed to use of setAlias()

system/Database/BaseBuilder.php
```diff
@@ -2653,6 +2653,105 @@ public function delete($where = '', ?int $limit = null, bool $resetData = true)
...
+                    (
+                        is_string($key) ?
+                    $table . '.' . $key . ' = ' . $alias . '.' . $value :
+                    $table . '.' . $value . ' = ' . $alias . '.' . $value
+                    )
+                    ),
+                    array_keys($constraints),
+                    $constraints
+                )
+            );
```
michalsn, line 2717: Does this code have the correct indentation? It is terrible to read - or it's just me.
Can you see if anything will scream at you if you try to make it more readable?
I hope it is not forced by CS Fixer...

system/Database/OCI8/Builder.php
```diff
@@ -434,4 +434,88 @@ protected function _upsertBatch(string $table, array $keys, array $values): stri
...
+                array_map(
+                    static fn ($key, $value) => (
+                        $value instanceof RawSql ?
+                    $value :
+                    (
+                        is_string($key) ?
+                    $table . '.' . $key . ' = ' . $alias . '.' . $value :
+                    $table . '.' . $value . ' = ' . $alias . '.' . $value
+                    )
+                    ),
```
michalsn, line 476: Same story here.

system/Database/Postgre/Builder.php
```diff
@@ -408,4 +408,88 @@ protected function _upsertBatch(string $table, array $keys, array $values): stri
...
+                    static fn ($key, $value) => (
+                        $value instanceof RawSql ?
+                    $value :
+                    (
+                        is_string($key) ?
+                    $table . '.' . $key . ' = ' . $alias . '.' . $value :
+                    $table . '.' . $value . ' = ' . $alias . '.' . $value
+                    )
+                    ),
+                    array_keys($constraints),
```
michalsn, line 451: And here.

michalsn (approved):
I would like to ask you to check one thing, but other than that, I think this is ready.

system/Database/BaseBuilder.php
sclubricants: I was able to fix these and CS Fixer didn't revert them.

sclubricants:
> I would like to ask you to check one thing, but other than that, I think this is ready.

This PR is pretty strait forward I think. This mostly falls in the footprint of the other *Batch() methods. I don't think there are any major controversies to discuss here are there? I know for me this would be very helpful as most of my database has composite primary keys. I think the order line item is a good example of where this is useful.

@kenjis what do you think? Any chance of getting this committed before 4.3?

kenjis:
I would like to include this in v4.3.0.
I will see this today or tomorrow.

MGatner:
Great news! Let me know how I can help.

user_guide_src/source/database/query_builder/118.php
```diff
@@ -0,0 +1,39 @@
...
+    ->where('del.qty >', 1)
+    ->deleteBatch();
+
+/*
+ * MySQL Produces:
+ * DELETE `order_line` FROM `order_line`
+ * INNER JOIN (
+ * SELECT 48372 `order`, 3 `line`, 'Keyboard' `product`, 1 `qty` UNION ALL
+ * SELECT 48372 `order`, 4 `line`, 'Mouse'    `product`, 1 `qty` UNION ALL
+ * SELECT 48372 `order`, 5 `line`, 'Monitor'  `product`, 2 `qty`
```
kenjis, line 35:
```suggestion
 * SELECT 3 `line`, 48372 `order`, 'Keyboard' `product`, 1 `qty` UNION ALL
 * SELECT 4 `line`, 48372 `order`, 'Mouse'    `product`, 1 `qty` UNION ALL
 * SELECT 5 `line`, 48372 `order`, 'Monitor'  `product`, 2 `qty`
```

system/Database/Postgre/Builder.php
```diff
@@ -408,4 +408,88 @@ protected function _upsertBatch(string $table, array $keys, array $values): stri
...
+            );
+
+            // convert binds in where
+            foreach ($this->QBWhere as $key => $where) {
+                foreach ($this->binds as $field => $bind) {
+                    $this->QBWhere[$key]['condition'] = str_replace(':' . $field . ':', $bind[0], $where['condition']);
+                }
+            }
+
+            // remove database prefix from alias in where
```
kenjis, line 463: This comment is for the next `str_replace($this->db->DBPrefix...)`.

system/Database/OCI8/Builder.php
```diff
@@ -434,4 +434,88 @@ protected function _upsertBatch(string $table, array $keys, array $values): stri
...
+            );
+
+            // convert binds in where
+            foreach ($this->QBWhere as $key => $where) {
+                foreach ($this->binds as $field => $bind) {
+                    $this->QBWhere[$key]['condition'] = str_replace(':' . $field . ':', $bind[0], $where['condition']);
+                }
+            }
+
+            // remove database prefix from alias in where
```
kenjis, line 489: This comment is for the next `str_replace($this->db->DBPrefix...)`.

system/Database/BaseBuilder.php
```diff
@@ -2653,6 +2653,105 @@ public function delete($where = '', ?int $limit = null, bool $resetData = true)
...
+                    $this->QBWhere[$key]['condition'] = str_replace(':' . $field . ':', $bind[0], $where['condition']);
+                }
+            }
+
+            // remove database prefix from alias in where
+            $sql .= ' ' . str_replace(
+                $this->db->DBPrefix . trim($alias, $this->db->escapeChar),
+                trim($alias, $this->db->escapeChar),
+                $this->compileWhereHaving('QBWhere')
+            );
```
kenjis, line 2731: Can you add a test case for this replacement?

kenjis:
I have made some minor comments, but overall it looks good. 

Also, the base is a bit old, so please rebase.

user_guide_src/source/database/query_builder/118.php
sclubricants: I'm not sure why you suggest this. the table name is `order_line`... so order + line. The pk would have order first and then line.

system/Database/OCI8/Builder.php
sclubricants: fixed

system/Database/BaseBuilder.php
sclubricants: This replacement is not longer necessary when alias tracking was added to setAlias(). I have removed the replacement in all these functions. I did add a where() clause to the test case to validate any sql generated.

```php
    public function setAlias(string $alias): BaseBuilder
    {
        if ($alias !== '') {
            $this->db->addTableAlias($alias); 
```

system/Database/OCI8/Builder.php
sclubricants: removed

system/Database/Postgre/Builder.php
sclubricants: removed

user_guide_src/source/database/query_builder/118.php
kenjis: That's the result in my local environment. MySQL 5.7.

sclubricants: Oh ok.. that is because setData() sorts the keys alphabetically.

sclubricants: So technically it will get reordered but for readability I like the way it reads.. can we leave it this way?

kenjis: Why does setData() sort the keys?
If the output SQL will be difficult to read, it seems better not to sort.

sclubricants: I don't know. That is the way things were previously. I left things as the were when designing setData().

sclubricants: Done

kenjis (approved):

PR #6737, opened 2022-10-22
fix: CodeIgniter::run() doesn't respect $returnResponse

### LINKED ISSUE #6650: Bug: sendResponse doesn't respect returnResponse parameter in codeigniter.php
### PHP Version

8.1

### CodeIgniter4 Version

4.2.7

Linux

### Which server did you use?

cli

### Database

MariaDB 10.2

### What happened?

File in the question: `system/CodeIgniter.php`

The run() method (line: 292) in the file has the second parameter to return the response instead of printing it.
However, in the case of  page not found exception, the script straight prints the result.
Which makes it harder to cache the exception response in the caller script.

Indeed there are other ways to deal with the issue but all of them require extending a few methods in the child class (extending `Codeigniter.php`) just to include a return statement.

### Steps to Reproduce

1. Use a non-traditional server like [ReactPHP](https://reactphp.org/) or [Swoole](https://github.com/swoole).
2. Initiate requests from the server to CodeIgniter.
```php
$app = Services::codeigniter(config('App'));
$app->initialize();
$app->setContext('php-cli');
```
3. Execute the `run()` method along with `$returnResponse` set to true.
`$app->run(null, true);`
4. Override default 404 page with a controller/method.
5. Request a resource that doesn't exist.
6. In the case of a response code 200, CI will return the HTTP\Response object.
7. In the case of a response code 404, CI will return null and the result (HTML formatted response) will be written to the server's log file.

I haven't tested but the behavior might be seen in the case of response code 500 as well.

### Expected Output

When there is a Page not found exception, the server should return the response object instead of printing (echoing) it. 
Same is true when a internal server error occurs.

### Anything else?

I have extended the core class to overcome this issue but I strongly believe that the framework should respect the type of return requested for either of the HTTP response codes.

Hence it, would be better to make `$returnResponse` as a class property and change the `sendResponse()` method's behavior.

Otherwise below are lines to update in the `system/CodeIgniter.php` file to let the the application return the response when it is asked rather than echoing when a page not found or server error exceptions occur.

Line 312 in the run() method
`return $this->sendResponse($returnResponse);`

Line 348 in the run() method
`return $this->sendResponse($returnResponse);`

Line 354 pass the parameter
`$this->display404errors($e, $returnResponse);`

Line 914 method signature
`protected function display404errors(PageNotFoundException $e, bool $returnResponse = false)`

Line 937 in the display404errors() method
`return $this->sendResponse($returnResponse);`

Line 1072 sendResponse() method
```php
    protected function sendResponse($returnResponse = false)
    {
        if (!$returnResponse) {
			$this->response->pretend($this->useSafeOutput)->send();
			return;
		}
		return $this->response;
    }
```

kenjis:
**Description**
Fixes #6650

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -152,6 +152,11 @@ class CodeIgniter
     protected ?string $context = null;
 
     /**
+     * Whether to return Response object or send response.
+     */
+    protected bool $returnResponse = false;
+
+    /**
      * Constructor.
      */
     public function __construct(App $config)
@@ -291,6 +296,8 @@ class CodeIgniter
      */
     public function run(?RouteCollectionInterface $routes = null, bool $returnResponse = false)
     {
+        $this->returnResponse = $returnResponse;
+
         if ($this->context === null) {
             throw new LogicException('Context must be set before run() is called. If you are upgrading from 4.1.x, you need to merge `public/index.php` and `spark` file from `vendor/codeigniter4/framework`.');
         }
@@ -309,7 +316,11 @@ class CodeIgniter
         if ($this->request instanceof IncomingRequest && strtolower($this->request->getMethod()) === 'cli') {
             $this->response->setStatusCode(405)->setBody('Method Not Allowed');
 
-            $this->sendResponse();
+            if (! $this->returnResponse) {
+                $this->sendResponse();
+            } else {
+                return $this->response;
+            }
 
             return;
         }
@@ -345,13 +356,22 @@ class CodeIgniter
             // If the route is a 'redirect' route, it throws
             // the exception with the $to as the message
             $this->response->redirect(base_url($e->getMessage()), 'auto', $e->getCode());
-            $this->sendResponse();
+
+            if (! $this->returnResponse) {
+                $this->sendResponse();
+            } else {
+                return $this->response;
+            }
 
             $this->callExit(EXIT_SUCCESS);
 
             return;
         } catch (PageNotFoundException $e) {
-            $this->display404errors($e);
+            $return = $this->display404errors($e);
+
+            if ($return instanceof ResponseInterface) {
+                return $return;
+            }
         }
     }
 
@@ -400,6 +420,8 @@ class CodeIgniter
      *
      * @throws PageNotFoundException
      * @throws RedirectException
+     *
+     * @deprecated $returnResponse is deprecated, and no longer used.
      */
     protected function handleRequest(?RouteCollectionInterface $routes, Cache $cacheConfig, bool $returnResponse = false)
     {
@@ -433,7 +455,8 @@ class CodeIgniter
 
             // If a ResponseInterface instance is returned then send it back to the client and stop
             if ($possibleResponse instanceof ResponseInterface) {
-                return $returnResponse ? $possibleResponse : $possibleResponse->pretend($this->useSafeOutput)->send();
+                return $this->returnResponse ? $possibleResponse
+                    : $possibleResponse->pretend($this->useSafeOutput)->send();
             }
 
             if ($possibleResponse instanceof Request) {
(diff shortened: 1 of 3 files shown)


system/CodeIgniter.php
```diff
@@ -309,7 +316,11 @@ public function run(?RouteCollectionInterface $routes = null, bool $returnRespon
...
             $this->response->setStatusCode(405)->setBody('Method Not Allowed');
 
-            $this->sendResponse();
+            if (! $this->returnResponse) {
+                $this->sendResponse();
+            } else {
+                return $this->response;
+            }
 
             return;
```
samsonasik, line 325: this can be changed to:

```suggestion
            if ($this->returnResponse) {
                return $this->response;
            }

            $this->sendResponse();
            return;
```

system/CodeIgniter.php
```diff
@@ -345,13 +356,22 @@ public function run(?RouteCollectionInterface $routes = null, bool $returnRespon
             // If the route is a 'redirect' route, it throws
             // the exception with the $to as the message
             $this->response->redirect(base_url($e->getMessage()), 'auto', $e->getCode());
-            $this->sendResponse();
+
+            if (! $this->returnResponse) {
+                $this->sendResponse();
+            } else {
+                return $this->response;
+            }
```
samsonasik, line 364: this can be changed to:

```suggestion
            if ($this->returnRespose) {
                return $this->response;
            }

            $this->sendResponse();
```

system/CodeIgniter.php
```diff
@@ -934,7 +959,11 @@ protected function display404errors(PageNotFoundException $e)
...
             $cacheConfig = new Cache();
             $this->gatherOutput($cacheConfig, $returned);
-            $this->sendResponse();
+            if (! $this->returnResponse) {
+                $this->sendResponse();
+            } else {
+                return $this->response;
+            }
 
             return;
```
samsonasik, line 968: this can be changed to:

```suggestion
            if ($this->returnResponse) {
                return $this->response;
            }

            $this->sendResponse();
            return;
```

system/CodeIgniter.php
```diff
@@ -345,13 +355,22 @@ public function run(?RouteCollectionInterface $routes = null, bool $returnRespon
...
             // the exception with the $to as the message
             $this->response->redirect(base_url($e->getMessage()), 'auto', $e->getCode());
+
+            if ($this->returnRespose) {
+                return $this->response;
+            }
+
             $this->sendResponse();
 
             $this->callExit(EXIT_SUCCESS);
```
kenjis, line 365: Not related to this PR, but is this `callExit()` necessary?

system/CodeIgniter.php
kenjis: Yes, that makes it a bit easier to read!

system/CodeIgniter.php
MGatner: I'm not sure - maybe just to enforce a "never return" condition?

system/CodeIgniter.php
```diff
@@ -400,6 +420,8 @@ private function isWeb(): bool
      *
      * @throws PageNotFoundException
      * @throws RedirectException
+     *
+     * @deprecated $returnResponse is deprecated, and no longer used.
```
MGatner, line 424: So what happens if a user called `run()` without returnResponse and `handleRequest()` with it? Is this a changed behavior?

MGatner (commented):
Let's see if @AmitSonkhiya can test this.

system/CodeIgniter.php
kenjis: I can't think of any such use case, but I have fixed it so that the behavior does not change.

MGatner (approved):
Code looks good, I would like to see another tester if anyone is up for it.

system/CodeIgniter.php
paulbalandan: If you will add `@phpstan-return never` to `protected function callExit`, then phpstan correctly flags that the `return;` below will be reached. So, this must be a bug?

PR #6741, opened 2022-10-22
Fix BaseBuilder setAlias() and RawSql use with key value pairs

sclubricants:
This fixes an issue when using an alias with setAlias() method. When using escape identifiers it was adding database prefix to alias. This was fixed.

Also, a small change to use of RawSql in constraints for updateBatch() and upsertBatch(). RawSql can be used for the entire expression or used as a key value pair. This is how I originally intended it but somehow it didn't get in correctly.

Can we expedite merging this? These fixes are needed for my other PRs.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Postgre/Builder.php
+++ b/system/Database/Postgre/Builder.php
@@ -364,6 +364,8 @@ class Builder extends BaseBuilder
                 }
             }
 
+            $alias = '"excluded"';
+
             $updateFields = $this->QBOptions['updateFields'] ?? $this->updateFields($keys, false, $constraints)->QBOptions['updateFields'] ?? [];
 
             $sql = 'INSERT INTO ' . $table . ' (';
@@ -382,8 +384,8 @@ class Builder extends BaseBuilder
                 ",\n",
                 array_map(
                     static fn ($key, $value) => $key . ($value instanceof RawSql ?
-                        ' = ' . $value :
-                        ' = "excluded".' . $value),
+                    " = {$value}" :
+                    " = {$alias}.{$value}"),
                     array_keys($updateFields),
                     $updateFields
                 )
--- a/system/Database/Postgre/Builder.php
+++ b/system/Database/Postgre/Builder.php
@@ -366,7 +366,7 @@ class Builder extends BaseBuilder
 
             $alias = $this->QBOptions['alias'] ?? '"excluded"';
 
-            if (strtolower($alias) !== '"excluded"' && $this->db->DBDebug) {
+            if (strtolower($alias) !== '"excluded"') {
                 throw new DatabaseException('Postgres alias is always named "excluded". A custom alias cannot be used.');
             }
 
--- a/system/Database/Postgre/Builder.php
+++ b/system/Database/Postgre/Builder.php
@@ -364,7 +364,11 @@ class Builder extends BaseBuilder
                 }
             }
 
-            $alias = '"excluded"';
+            $alias = $this->QBOptions['alias'] ?? '"excluded"';
+
+            if (strtolower($alias) !== '"excluded"' && $this->db->DBDebug) {
+                throw new DatabaseException('Postgres alias is always named "excluded". A custom alias cannot be used.');
+            }
 
             $updateFields = $this->QBOptions['updateFields'] ?? $this->updateFields($keys, false, $constraints)->QBOptions['updateFields'] ?? [];
 
--- a/system/Database/Postgre/Builder.php
+++ b/system/Database/Postgre/Builder.php
@@ -13,6 +13,7 @@ namespace CodeIgniter\Database\Postgre;
 
 use CodeIgniter\Database\BaseBuilder;
 use CodeIgniter\Database\Exceptions\DatabaseException;
+use CodeIgniter\Database\InvalidArgumentException;
 use CodeIgniter\Database\RawSql;
 
 /**
@@ -367,7 +368,7 @@ class Builder extends BaseBuilder
             $alias = $this->QBOptions['alias'] ?? '"excluded"';
 
             if (strtolower($alias) !== '"excluded"') {
-                throw new DatabaseException('Postgres alias is always named "excluded". A custom alias cannot be used.');
+                throw new InvalidArgumentException('Postgres alias is always named "excluded". A custom alias cannot be used.');
             }
 
             $updateFields = $this->QBOptions['updateFields'] ?? $this->updateFields($keys, false, $constraints)->QBOptions['updateFields'] ?? [];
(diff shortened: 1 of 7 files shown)


sclubricants:
The bug can be seen in UpdateTest `testRawSqlConstraint()`

Before I had to use the database prefix when defining the alias so that protectIdentifiers matched. This is no longer the case.

sclubricants:
@kenjis Can we expedite this fix?

kenjis:
>  Fix use of setAlias()
>
> Added $this->db->addTableAlias($alias) to setAlias(). This keeps track of the alias and prevents adding the database prefix when using $this->db->protectIdentifiers(). There was some issues with this.
>
> Also made some fixes to updateBatch() and use of RawSql. Just a small fix that allows key value pairs with RawSql.
36f2432da9e912661373f537674d52418e59682c

Make only one change per commit if possible. 
I would also like you to add test code to prove that the bug has been fixed.

tests/system/Database/Live/UpdateTest.php
```diff
@@ -475,14 +475,38 @@ public function testRawSqlConstraint()
 
         $builder = $this->db->table('user');
 
-        $builder->setData($data, true, 'db_myalias')
+        $builder->setData($data, true, 'myalias')
```
kenjis, line 478: Without this change, this test passes.
Is it okay?

kenjis: The alias does not matter?

When `db_myalias`:
```sql
UPDATE `db_user`
SET
`name` = `db_myalias`.`name`,
`country` = `db_myalias`.`country`
FROM (
SELECT 'Germany' `country`, 'derek@world.com' `email`, 'Derek Jones' `name`
) `db_myalias`
WHERE `db_user`.`email` = `db_myalias`.`email`
```

When `myalias`:
```sql
UPDATE `db_user`
SET
`name` = `myalias`.`name`,
`country` = `myalias`.`country`
FROM (
SELECT 'Germany' `country`, 'derek@world.com' `email`, 'Derek Jones' `name`
) `myalias`
WHERE `db_user`.`email` = `myalias`.`email`
```

kenjis: I got it.
`$this->db->protectIdentifiers('myalias.email')` converts `myalias` to `db_myalias` when `$builder->setData($data, true, 'db_myalias')`. Because `myalias` is not an alias.

This is okay. Passing `db_myalias` to `setData()` is just a misuse.

kenjis (dismissed):
I have not looked at the details of the specific DB drivers, but the test code is okay and the test passes.

LGTM.

sclubricants:
>When fixing a bug, it is recommended that you first write a test case that fails because of the bug.

This would be the difference of the test that I altered. I had hacked the bug in the test and now you see the test as it should be.

sclubricants:
I think one more thing I should add is to throw an exception if you use an alias for upsert with Postgres and SQLite. They are always Identified by "excluded". This way a developer doesn't get lost trying to figure out why their alias is not working.

system/Database/Postgre/Builder.php
```diff
@@ -364,6 +364,12 @@ protected function _upsertBatch(string $table, array $keys, array $values): stri
                 }
             }
 
+            $alias = $this->QBOptions['alias'] ?? '"excluded"';
+
+            if (strtolower($alias) !== '"excluded"' && $this->db->DBDebug) {
+                throw new DatabaseException('Postgres alias is always named "excluded". A custom alias cannot be used.');
+            }
```
kenjis, line 371: I guess it doesn't work if `$this->db->DBDebug` is false and using a custom alias?

sclubricants: Honestly I'm not quite sure what the reasons are for using this. I was just following the pattern that I saw being used in the rest of the code. Can you elaborate on why this should or should not be used?

sclubricants: I guess this is just an extra layer of protection to prevent sensitive database errors from being displayed. ?? In this case if the error is not thrown the script will go on to throw an error with the query which could then potentially display sensitive info so maybe its best to throw regardless of debug mode.

sclubricants: Removed DBDebug test.

kenjis: This is because of historical reasons.

1. The `DBDebug` option was originally a debugging option that decides whether or not database errors should be displayed on the screen. So if `DBDebug` is false, CI should return `false` without throwing an exception.
   > db_debug | TRUE/FALSE (boolean) - Whether database errors should be displayed. https://codeigniter.com/userguide3/database/configuration.html#explanation-of-values
2. Before v4.3.0, in production environment `DBDebug` is `false` by default. https://github.com/codeigniter4/CodeIgniter4/blob/020af6d2d2e61dca96797240308bb96bb080a77a/app/Config/Database.php#L42
3. But it is not good practice that changing the behavior between production and development environment. So we have changed `DBDebug` to `true` anytime by default since v4.3.0. See https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/changelogs/v4.3.0.rst#changes
4. As a result, now `DBDebug` means whether or not to throw an exception when an database error occurs.

kenjis: > I guess this is just an extra layer of protection to prevent sensitive database errors from being displayed. 

I also think the `DBDebug` option was originally created like that.
But displaying database errors has nothing to do with the option now.
Because CI4 never displays error messages on production environment.

However, if a database error occurs and `DBDebug` is `false`, methods must be implemented to return `false` without throwing an exception.

kenjis: I think `DatabaseException` is a runtime exception that is thrown when a database query fails.

In this case, it seems better to use `InvalidArgumentException` or `BadMethodCallException` that is a logic exception. Logic Exception represents error in the program logic. We don't need to check `DBDebug`.

kenjis: I would like to remove `DBDebug` in the future. Because it seems useless.
But it is a big breaking change. So it will take long time.
Until then, we need to implement database methods to return `false` when `DBDebug` is `false`.

sclubricants: Changed to use `CodeIgniter\Database\InvalidArgumentException`

kenjis (approved):

michalsn (approved):

kenjis:
@sclubricants Thank you!

PR #6742, opened 2022-10-23
docs: add PHPDoc to BaseBuilder

kenjis:
**Description**
- add PHPDoc comments

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -86,8 +86,10 @@ class BaseBuilder
 
     /**
      * QB keys
+     * list of column names.
      *
      * @var array
+     * @phpstan-var list<string>
      */
     protected $QBKeys = [];
 
@@ -130,6 +132,7 @@ class BaseBuilder
      * QB data sets
      *
      * @var array
+     * @phpstan-var array<string, string>|list<list<string|int>>
      */
     protected $QBSet = [];
 
@@ -1931,6 +1934,8 @@ class BaseBuilder
 
     /**
      * Generates a platform-specific upsertBatch string from the supplied data
+     *
+     * @used-by batchExecute
      */
     protected function _upsertBatch(string $table, array $keys, array $values): string
     {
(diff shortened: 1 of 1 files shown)


system/Database/BaseBuilder.php
```diff
@@ -86,8 +86,10 @@ class BaseBuilder
 
     /**
      * QB keys
+     * list of column names.
      *
      * @var array
+     * @phpstan-var list<string>
```
MGatner, line 92: Just for my own edification... what is the benefit of using the vendor-specific annotation? I know Psalm and some IDEs understand `@var string[]` - won't they miss out on this nuance?

system/Database/BaseBuilder.php
```diff
@@ -2417,6 +2424,11 @@ public function updateBatch($set = null, $constraints = null, int $batchSize = 1
 
     /**
      * Generates a platform-specific batch update string from the supplied data
+     *
+     * @used-by batchExecute
+     *
+     * @param array $keys   QBKeys
+     * @param array $values QBSet
```
MGatner, line 2431: Can these get any more specific? Let's probably are all strings, right?

sclubricants: $values can be RawSql

system/Database/BaseBuilder.php
kenjis: `string[]` means `array<string>`. It does not care about keys.
https://psalm.dev/docs/annotating_code/type_syntax/array_types/#phpdoc-syntax

`list` means true list that the key starts with 0, and goes 1,2,3,...

system/Database/BaseBuilder.php
kenjis: @sclubricants But it is `array` typed.
`protected function _updateBatch(string $table, array $keys, array $values): string`

kenjis: $QBKeys: `list<string>`
$QBSet: `array<string, string>|list<list<string|int>>`, but in this case `list<list<string|int>>` only.

kenjis: Updated.

MGatner (approved):

system/Database/BaseBuilder.php
paulbalandan: VS Code does not support it yet.

PR #6743, opened 2022-10-23
refactor: Database connect()

kenjis:
**Description**
Reduce the complexity of the method.

- refactor
- remove `&`
- fix PHPDoc

```console
$ phploc system/Database/Config.php
```
Before:
```
Cyclomatic Complexity
  Average Complexity per LLOC                     0.52
  Average Complexity per Class                   13.00
    Minimum Class Complexity                     13.00
    Maximum Class Complexity                     13.00
  Average Complexity per Method                   3.00
    Minimum Method Complexity                     1.00
    Maximum Method Complexity                    11.00
```
After:
```
Cyclomatic Complexity
  Average Complexity per LLOC                     0.38
  Average Complexity per Class                   10.00
    Minimum Class Complexity                     10.00
    Maximum Class Complexity                     10.00
  Average Complexity per Method                   2.50
    Minimum Method Complexity                     1.00
    Maximum Method Complexity                     8.00
```

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/Config.php
+++ b/system/Database/Config.php
@@ -65,18 +65,12 @@ class Config extends BaseConfig
             if (! isset($dbConfig->{$group})) {
                 throw new InvalidArgumentException($group . ' is not a valid database connection group.');
             }
+
+            $config = $dbConfig->{$group};
         }
 
         if ($getShared && isset(static::$instances[$group])) {
             return static::$instances[$group];
-        }
-
-        if (isset($dbConfig->{$group})) {
-            $config = $dbConfig->{$group};
-        }
-
-        if (! isset($config)) {
-            throw new InvalidArgumentException('There is no valid database config.');
         }
 
         static::ensureFactory();
--- a/system/Database/Config.php
+++ b/system/Database/Config.php
@@ -59,7 +59,7 @@ class Config extends BaseConfig
         /** @var DbConfig $dbConfig */
         $dbConfig = config('Database');
 
-        if (empty($group)) {
+        if ($group === null) {
             $group = ENVIRONMENT === 'testing' ? 'tests' : $dbConfig->defaultGroup;
         }
 
--- a/system/Database/Config.php
+++ b/system/Database/Config.php
@@ -60,7 +60,7 @@ class Config extends BaseConfig
         $dbConfig = config('Database');
 
         if ($group === null) {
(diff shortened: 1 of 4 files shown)


system/Database/Config.php
```diff
@@ -38,8 +39,9 @@ class Config extends BaseConfig
     /**
      * Creates the default
      *
-     * @param array|string $group     The name of the connection group to use, or an array of configuration settings.
-     * @param bool         $getShared Whether to return a shared instance of the connection.
+     * @param array|BaseConnection|string|null $group     The name of the connection group to use,
```
MGatner, line 42: Are these newly-supported types? Or a correction of what is actually already acceptable?

system/Database/Config.php
```diff
@@ -53,16 +55,21 @@ public static function connect($group = null, bool $getShared = true)
         if (is_array($group)) {
             $config = $group;
             $group  = 'custom-' . md5(json_encode($config));
-        }
+        } else {
+            /** @var DbConfig $dbConfig */
+            $dbConfig = config('Database');
 
-        $config ??= config('Database');
+            if ($group === null) {
```
MGatner, line 62: I'm in favor of getting rid `empty()` but we no longer handle `''` which I think is a possibility.

MGatner: Actually it looks like this is acceptable: https://3v4l.org/qSBFc

system/Database/Config.php
kenjis: A correction of what is actually already acceptable.

system/Database/Config.php
kenjis: The current `develop` code does not support empty string as `$group`.
https://github.com/codeigniter4/CodeIgniter4/blob/020af6d2d2e61dca96797240308bb96bb080a77a/system/Database/Config.php#L60-L62
`$group` will be the default group name.

And `Database::load()` does not accept empty string.
https://github.com/codeigniter4/CodeIgniter4/blob/020af6d2d2e61dca96797240308bb96bb080a77a/system/Database/Config.php#L78
https://github.com/codeigniter4/CodeIgniter4/blob/020af6d2d2e61dca96797240308bb96bb080a77a/system/Database/Database.php#L42-L46

MGatner (approved):

paulbalandan (approved):

PR #6749, opened 2022-10-24
Add template types to Result

paulbalandan:
**Description**
Make ResultInterface a generic type.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/MySQLi/PreparedQuery.php
+++ b/system/Database/MySQLi/PreparedQuery.php
@@ -15,13 +15,14 @@ use BadMethodCallException;
 use CodeIgniter\Database\BasePreparedQuery;
 use CodeIgniter\Database\Exceptions\DatabaseException;
 use mysqli;
+use mysqli_result;
 use mysqli_sql_exception;
 use mysqli_stmt;
 
 /**
  * Prepared query for MySQLi
  *
- * @extends BasePreparedQuery<false|mysqli, mysqli_stmt>
+ * @extends BasePreparedQuery<mysqli, mysqli_stmt, mysqli_result>
  */
 class PreparedQuery extends BasePreparedQuery
 {
@@ -94,7 +95,7 @@ class PreparedQuery extends BasePreparedQuery
     /**
      * Returns the result object for the prepared query or false on failure.
      *
-     * @return mixed
+     * @return false|mysqli_result
      */
     public function _getResult()
     {
--- a/system/Database/OCI8/PreparedQuery.php
+++ b/system/Database/OCI8/PreparedQuery.php
@@ -18,7 +18,7 @@ use CodeIgniter\Database\Exceptions\DatabaseException;
 /**
  * Prepared query for OCI8
  *
- * @extends BasePreparedQuery<false|resource, resource>
+ * @extends BasePreparedQuery<resource, resource, resource>
  */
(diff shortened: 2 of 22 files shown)


system/Database/MySQLi/PreparedQuery.php
```diff
@@ -34,7 +35,7 @@ class PreparedQuery extends BasePreparedQuery
      * @param array $options Passed to the connection's prepare statement.
      *                       Unused in the MySQLi driver.
      *
-     * @return mixed
+     * @return static
```
kenjis, line 38:
```suggestion
     * @return $this
```

system/Database/OCI8/PreparedQuery.php
```diff
@@ -45,7 +45,7 @@ class PreparedQuery extends BasePreparedQuery implements PreparedQueryInterface
      * @param array $options Passed to the connection's prepare statement.
      *                       Unused in the OCI8 driver.
      *
-     * @return mixed
+     * @return static
```
kenjis, line 48:
```suggestion
     * @return $this
```

system/Database/Postgre/PreparedQuery.php
```diff
@@ -51,7 +51,7 @@ class PreparedQuery extends BasePreparedQuery
      * @param array $options Passed to the connection's prepare statement.
      *                       Unused in the MySQLi driver.
      *
-     * @return mixed
+     * @return static
```
kenjis, line 54:
```suggestion
     * @return $this
```

system/Database/SQLSRV/PreparedQuery.php
```diff
@@ -51,7 +51,7 @@ public function __construct(Connection $db)
      *
      * @param array $options Options takes an associative array;
      *
-     * @return mixed
+     * @return static
```
kenjis, line 54:
```suggestion
     * @return $this
```

system/Database/SQLite3/PreparedQuery.php
```diff
@@ -41,7 +42,7 @@ class PreparedQuery extends BasePreparedQuery
      * @param array $options Passed to the connection's prepare statement.
      *                       Unused in the MySQLi driver.
      *
-     * @return $this
+     * @return static
```
kenjis, line 45:
```suggestion
     * @return $this
```

kenjis:
I think if the code is `return $this;`, the phpdoc should be `@return $this`.

> $this: the element to which this type applies is the same exact instance as the current class in the given context. As such, this type is a stricter version of static, because the returned instance must not only be of the same class but also the same instance.
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md

But if we can set return type, it should be `static` (or `self`).

system/Database/MySQLi/PreparedQuery.php
paulbalandan: I will revert this, following changes in #6750

MGatner:
This one can be added at any point as a bugfix/tech-debt change so no need to rush it for 4.3.

MGatner (approved):
Looks good! Lots more baseline cleanup, makes me happy 🤗

kenjis (approved):

PR #6163, opened 2022-06-20
fix: throws DatabaseException in DB connections

kenjis:
**Description**
Fixes #6007

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #6750, opened 2022-10-24
fix: BasePreparedQuery class to return boolean values for write-type queries

### LINKED ISSUE #4342: Bug: prepared statement execute() returns Result object for all queries
**Describe the bug**
In the CI4 spec, `query()` returns bool for write-type queries.
But the current implementation of `execute()` returns Result object for all queries.

Isn't it a bug?

See <https://github.com/codeigniter4/CodeIgniter4/pull/4176#issuecomment-784805837>.

**CodeIgniter 4 version**
develop branch.

**Affected module(s)**
Database

The same as `query()` return value.

michalsn:
**Description**
Fixes #4342

This PR fixes the behavior of Prepared Queries, which always return a `Result` object.

Some time ago, we fixed the regular query: #4176, but somehow Prepared Queries were forgotten.

Desired behavior is the same as for regular queries, return:
* `Result` object for "read" queries
* `bool` value for "write" queries

Potentially this is a breaking change, but we should treat it as a bug fix.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BasePreparedQuery.php
+++ b/system/Database/BasePreparedQuery.php
@@ -111,7 +111,9 @@ abstract class BasePreparedQuery implements PreparedQueryInterface
      * Takes a new set of data and runs it against the currently
      * prepared query. Upon success, will return a Results object.
      *
-     * @return ResultInterface
+     * @return bool|ResultInterface
+     *
+     * @throws DatabaseException
      */
     public function execute(...$data)
     {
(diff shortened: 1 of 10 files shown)


system/Database/BasePreparedQuery.php
```diff
@@ -101,24 +102,74 @@ abstract public function _prepare(string $sql, array $options = []);
      * Takes a new set of data and runs it against the currently
      * prepared query. Upon success, will return a Results object.
      *
-     * @return ResultInterface
+     * @return bool|ResultInterface
+     *
+     * @throws Exception
```
kenjis, line 107: Why throws `Exception`?
If the query execution fails, it seems it should throw `DatabaseException`.

user_guide_src/source/changelogs/v4.2.8.rst
```diff
@@ -12,7 +12,8 @@ Release Date: Unreleased
 BREAKING
 ********
 
-none.
+- ``CodeIgniter\Database\BasePreparedQuery`` class return now a bool value for write-type queries instead of the ``Result`` class object.
```
kenjis, line 15:
```suggestion
- ``CodeIgniter\Database\BasePreparedQuery`` class returns now a bool value for write-type queries instead of the ``Result`` class object.
```

system/Database/BasePreparedQuery.php
michalsn: Actually, it will throw different exceptions depending on the driver. For MySQL it will be `mysqli_sql_exception` for others `ErrorException`. There is no point in changing this and hiding the actual exception - at least I don't think it would be appropriate. I believe that the `DatabaseException` should be used for things that are coming from the framework side and are in our gesture to handle.

kenjis: We have changed the exception classes already in v4.3. 
See #6163 and https://github.com/codeigniter4/CodeIgniter4/blob/4.3/user_guide_src/source/changelogs/v4.3.0.rst#id1

When CI throws many kind of Exception, if we switch DBMS, our code will break when we want to catch the execption.
Or we must always catch `Exception`. But it is too wide. It is not good practice.

Anyway this PR is for `develop`. So it seems okay. But what about 4.3?

MGatner:
What's the status on this? Can we get it in soon for 4.3?

kenjis (approved):

paulbalandan (approved):

PR #6751, opened 2022-10-25
docs: fix PHPDoc types in Factories

kenjis:
**Description**

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/Factories.php
+++ b/system/Config/Factories.php
@@ -31,7 +31,8 @@ class Factories
      * Store of component-specific options, usually
      * from CodeIgniter\Config\Factory.
      *
-     * @var array<string, array>
+     * @var array[]
+     * @phpstan-var array<string, array<string, string|bool|null>>
      */
     protected static $options = [];
 
@@ -39,7 +40,7 @@ class Factories
      * Explicit options for the Config
      * component to prevent logic loops.
      *
-     * @var array<string, mixed>
+     * @phpstan-var array<string, string|bool|null>
      */
     private static array $configOptions = [
         'component'  => 'config',
@@ -53,7 +54,8 @@ class Factories
      * Mapping of class basenames (no namespace) to
      * their instances.
      *
-     * @var array<string, string[]>
+     * @var array[]
+     * @phpstan-var array<string, array<string, class-string>>
      */
     protected static $basenames = [];
 
@@ -63,7 +65,8 @@ class Factories
      * A multi-dimensional array with components as
      * keys to the array of name-indexed instances.
      *
-     * @var array<string, array>
+     * @var array[]
+     * @phpstan-var  array<string, array<class-string, object>>
      */
     protected static $instances = [];
 
@@ -74,9 +77,10 @@ class Factories
      *
      * @template T of Model
      *
-     * @param class-string<T> $name
+     * @phpstan-param class-string<T> $name
      *
-     * @return T
+     * @return object
+     * @phpstan-return T
      */
     public static function models(string $name, array $options = [], ?ConnectionInterface &$conn = null)
     {
@@ -228,7 +232,7 @@ class Factories
      *
      * @param string $component Lowercase, plural component name
      *
-     * @return array<string, mixed>
+     * @phpstan-return array<string, string|null|bool>
      */
     public static function getOptions(string $component): array
     {
@@ -253,7 +257,8 @@ class Factories
      *
      * @param string $component Lowercase, plural component name
      *
-     * @return array<string, mixed> The result after applying defaults and normalization
+     * @return array The result after applying defaults and normalization
+     * @phpstan-return array<string, string|null|bool>
      */
     public static function setOptions(string $component, array $values): array
     {
--- a/system/Config/Factories.php
+++ b/system/Config/Factories.php
@@ -31,8 +31,7 @@ class Factories
      * Store of component-specific options, usually
      * from CodeIgniter\Config\Factory.
      *
-     * @var array[]
-     * @phpstan-var array<string, array<string, string|bool|null>>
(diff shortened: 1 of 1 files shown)


system/Config/Factories.php
```diff
@@ -253,7 +257,8 @@ public static function getOptions(string $component): array
      *
      * @param string $component Lowercase, plural component name
      *
-     * @return array<string, mixed> The result after applying defaults and normalization
+     * @return array The result after applying defaults and normalization
+     * @phpstan-return array<string, string|null|bool>
```
paulbalandan, line 261: Should `@return` be the same with `@phpstan-return`?

system/Config/Factories.php
```diff
@@ -31,15 +31,16 @@ class Factories
      * Store of component-specific options, usually
      * from CodeIgniter\Config\Factory.
      *
-     * @var array<string, array>
+     * @var array[]
+     * @phpstan-var array<string, array<string, string|bool|null>>
```
paulbalandan, line 35: Should these two be the same? I mean the vendor-specific annotation does not use its vendor-specific tags (like `class-string` or `list`) only the regular annotations so `@var` and `@phpstan-var` should have the same values.

kenjis: If all the major IDEs understand `array<string, array<string, string|bool|null>>`, 
`@phpstan-var` is not need.

MGatner: Be always assumed they do but honestly I never looked into it. I'm not even sure what you would call this notation, since it isn't officially PHPDoc: https://docs.phpdoc.org/3.0/guide/guides/types.html#Arrays

system/Config/Factories.php
```diff
@@ -74,9 +77,10 @@ class Factories
      *
      * @template T of Model
      *
-     * @param class-string<T> $name
+     * @phpstan-param class-string<T> $name
      *
-     * @return T
+     * @return object
```
MGatner, line 82: Technically this doesn't need to return an instance of `CodeIgniter\Model` unless `instanceOf` is specified (which I think `model()` does) but for consistency maybe we should use that for the vanilla return type anyways?
```suggestion
     * @return Model
```

MGatner:
@kenjis it seems like you've made a move to keep PHPDoc tags strictly to the PHPDoc officially supported annotations. I would like us to step back and find consensus on our approach, because this is not something we have held to before.

I could see this going either way. If we start using vendor tags, do other vendors miss out on the enhanced types? Or, if we only use the vanilla tags: does PHPDoc or certain vendors fail to parse correctly?

If anyone has answers or best practices please share. I will plan to do some research on the subject as well. 

@codeigniter4/core-team

system/Config/Factories.php
samsonasik: `@var` only should be fine, the `@phpstan-var` only needed when it using very specific use case, eg: `non-empty-array` https://phpstan.org/r/554bcbfe-693f-47f0-8608-3cbacca36d85

paulbalandan:
In my repos, I always strive to use the generally available PHPDoc tags. However, if there's a chance to enhance the type further with specific tags, for example with PHPStan, I will **add** the phpstan-specific tags.

system/Config/Factories.php
kenjis: Is `@var array<string, array<string, string|bool|null>>` used with VS Code?

samsonasik: Yes 

Screen Shot 2022-10-27 at 19 34 35

kenjis: Thank you!

system/Config/Factories.php
kenjis: Done.

system/Config/Factories.php
kenjis: Removed `@phpstan-return`.

samsonasik (approved):

MGatner:
@paulbalandan Does that mean you never use `<>` notation on a vanilla annotation?

MGatner:
I'm seeing three levels here:
1. values explicitly defined by PHPDoc
2. values generally supported by major IDEs and static analyzers
3. value explicitly created by one vendor

The problem seems to be that 1. is clearly defined but 2. and 3. are not. Probably some more research (I haven't done it yet) could help clarify but it is probably a bit of a moving target. 

We should define explicitly what we want to support product-wise so once we have an idea of 2 versus 3 we can make a policy (unless of course we only want 1 for vanilla tags and everything else is vendored).

paulbalandan:
> @paulbalandan Does that mean you never use `<>` notation on a vanilla annotation?

On regular tags like `@var` I interchangeably use `array<int, string>` and `string[]` because intelephense understands both. But I use array shapes `array{0: string}` for phpstan specific tags.

paulbalandan (approved):

MGatner:
Intelephense and PHPStorm both support array shapes (I use them a lot at work), and I *think* VSCode does as well (for sure with Intelephense).

paulbalandan:
I'm using intelephense for vscode and it cannot recognize array shapes yet.

MGatner:
Now I'm going to have to go test... I'm using Intelephense in BBEdit and I swear it recognizes array shapes but maybe not??

FYI this has been on my list to get set up and try as an alternative to Intelephense: https://github.com/phpactor/phpactor It basically lets you use PHPStan as an LSP!

kenjis:
VS Code deep-assoc-completion seems to support array shapes.
https://plugins.jetbrains.com/plugin/9927-deep-assoc-completion

kenjis:
@MGatner Can I merge this?

MGatner (approved):
Yes! Please do.

PR #6752, opened 2022-10-25
fix: Time::now() does not respect timezone when using setTestNow()

kenjis:
**Description**
- fix bug 

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/I18n/TimeTrait.php
+++ b/system/I18n/TimeTrait.php
@@ -77,8 +77,13 @@ trait TimeTrait
 
         // If a test instance has been provided, use it instead.
         if ($time === '' && static::$testNow instanceof self) {
-            $timezone = $timezone ?: static::$testNow->getTimezone();
-            $time     = static::$testNow->format('Y-m-d H:i:s');
+            if ($timezone !== null) {
+                $testNow = static::$testNow->setTimezone($timezone);
+                $time    = $testNow->format('Y-m-d H:i:s');
+            } else {
+                $timezone = static::$testNow->getTimezone();
+                $time     = static::$testNow->format('Y-m-d H:i:s');
+            }
         }
 
         $timezone       = $timezone ?: date_default_timezone_get();
--- a/system/I18n/TimeTrait.php
+++ b/system/I18n/TimeTrait.php
@@ -315,7 +315,7 @@ trait TimeTrait
 
     /**
      * Creates an instance of Time that will be returned during testing
-     * when calling 'Time::now' instead of the current time.
+     * when calling 'Time::now()' instead of the current time.
      *
      * @param DateTimeInterface|self|string|null $datetime
      * @param DateTimeZone|string|null           $timezone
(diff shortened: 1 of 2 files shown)


system/I18n/Time.php
```diff
@@ -81,8 +81,13 @@ public function __construct(?string $time = null, $timezone = null, ?string $loc
 
         // If a test instance has been provided, use it instead.
         if ($time === '' && static::$testNow instanceof self) {
-            $timezone = $timezone ?: static::$testNow->getTimezone();
-            $time     = static::$testNow->format('Y-m-d H:i:s');
+            if ($timezone !== null) {
+                $testNow = static::$testNow->setTimezone($timezone);
```
MGatner, line 85: `Time` extends `DateTime`, not `DateTimeImmutable`, so it is mutable. In this case `setTimezone` is an internal method that returns a new instance, but I had to go look that up to be sure. I would still prefer to see this cloned for clarity, since the parent `DateTime::setTimezone()` would actually effect a change.

MGatner (changes requested):

system/I18n/Time.php
kenjis: > CodeIgniter provides a fully-localized, **immutable**, date/time class that is built on PHP’s DateTime object, but uses the Intl extension’s features to convert times across timezones and display the output correctly for different locales. https://codeigniter4.github.io/CodeIgniter4/libraries/time.html

Time is immutable. If not, it is a bug. We should fix it.

MGatner: It is not. Any of the following methods will mutate the current object:
* __construct()
* add()
* modify()
* setDate()
* setISODate()
* setTime()
* setTimestamp()
* sub()

kenjis: Oh, no! It is a big bug.

`setTimestamp()` is also overridden. But other methods are not.

MGatner: `Time` should have extended `DateTimeImmutable` if it was going to be immutable. Seems like a pretty big oversight there, but maybe immutability was added later to the User Guide? This all predates me.

For now I guess we change the User Guide? Changing the underlying class at this point seems to big a break. Alternatively we could override *all* the mutable methods but that could also be breaking and seems silly.

kenjis: It has been stated immutable since v4.0.0-alpha.1.
> CodeIgniter provides a fully-localized, immutable, date/time class that is built on PHP's DateTime object, but uses the Intl extension's features to convert times across timezones and display the output correctly for different locales. This class is the Time class and lives in the CodeIgniter\I18n namespace.
https://github.com/codeigniter4/CodeIgniter4/blob/v4.0.0-alpha.1/user_guide_src/source/libraries/time.rst

kenjis: Even though it is a breaking change, but it seems we must fix it.
The current Time has methods of immutable and methods of mutable.
It is extremely bad design and difficult to use.

kenjis: > For now I guess we change the User Guide?

It is not good. Because all methods in `Time` are immutable.
I sent a PR for true immutable Time: #6771

kenjis: I fix this in `4.3` that Time is fully immutable.

MGatner (approved):
That took quite a detour 😅

PR #6761, opened 2022-10-26
Leverage Tachycardia's Expeditable trait to obtain slim execution times

paulbalandan:
**Description**
To test this, use the `--testdox` option in `phpunit` and compare PHPUnit's reported time vs. the slow time reported. This should be lesser.
Ref: https://github.com/NexusPHP/tachycardia/blob/1.x/docs/limiting_test_times.md

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -64,6 +64,9 @@
                     <element key="tabulate">
                         <boolean>false</boolean>
                     </element>
+                    <element key="collectBare">
+                        <boolean>true</boolean>
+                    </element>
                 </array>
             </arguments>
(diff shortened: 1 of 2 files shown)


MGatner (approved):
This is already our recommendation on DevKit: https://github.com/codeigniter4/devkit/blob/develop/src/Template/phpunit.xml.dist#L58-L60

Let's make sure the @codeigniter4/core-team see this so they're aware.

michalsn (approved):

PR #6767, opened 2022-10-27
docs: fix `@param` types in html/url helper

kenjis:
**Description**
See #6310

- fix `@param` types

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


MGatner (approved):
I'm not going to confirm all the types. If PHPStan and Psalm are happy with these and you feel confident in them then they is good enough for me.

Issue #6764, opened 2022-10-26
Bug: errorHandler removes Faker's deprecated errors


kenjis:
From https://github.com/codeigniter4/CodeIgniter4/pull/6758#issuecomment-1291995601

### What happened?

```diff
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -151,7 +151,7 @@ class Exceptions
      */
     public function errorHandler(int $severity, string $message, ?string $file = null, ?int $line = null)
     {
-        if (error_reporting() & $severity) {
+        if (E_ALL & $severity) {
             throw new ErrorException($message, 0, $severity, $file, $line);
         }
 
```
```console
$ ./phpunit tests/system/Test/FabricatorTest.php
PHPUnit 9.5.25 #StandWithUkraine

Runtime:       PHP 8.1.11
Configuration: /Users/kenji/work/codeigniter/official/CodeIgniter4/phpunit.xml

.........E...........EEEEEEEEEEEEEEEE..........                   47 / 47 (100%)

Time: 00:00.625, Memory: 18.00 MB

There were 17 errors:

1) CodeIgniter\Test\FabricatorTest::testGetFakerReturnsUsableGenerator
ErrorException: Since fakerphp/faker 1.14: Accessing property "randomDigit" is deprecated, use "randomDigit()" instead.

/Users/kenji/work/codeigniter/official/CodeIgniter4/vendor/symfony/deprecation-contracts/function.php:25
/Users/kenji/work/codeigniter/official/CodeIgniter4/vendor/fakerphp/faker/src/Faker/Generator.php:950
/Users/kenji/work/codeigniter/official/CodeIgniter4/tests/system/Test/FabricatorTest.php:123

...

ERRORS!
Tests: 47, Assertions: 32, Errors: 17.
```

### Anything else?

The value of `error_reporting()` will change in error handler when using `@`.
See 
- https://3v4l.org/0DbSF
- https://3v4l.org/aftRX

> If a custom error handler function is set with [set_error_handler()](https://www.php.net/manual/en/function.set-error-handler.php), it will still be called even though the diagnostic has been suppressed.
>
> Warning
[...]

PR #6773, opened 2022-10-28
Turn on logging for deprecation notices by default

paulbalandan:
**Description**
Following up on #6705 , this turns on the feature by default.

Deprecation notices should not halt execution. These notices are here to _notify_ the users that something will not work in the _future_. They should not be forced to fix it now.

This PR changes the following:
- Turns on deprecation logging
- Raise logger reporting threshold to `5`
- Use the env variable `CODEIGNITER_SCREAM_DEPRECATIONS` to temporarily throw deprecations in tests (for convenience rather than changing the config).

Related: #6764

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Exceptions.php
+++ b/app/Config/Exceptions.php
@@ -60,7 +60,7 @@ class Exceptions extends BaseConfig
      * Use this option to temporarily cease the warnings and instead log those.
      * This option also works for user deprecations.
      */
-    public bool $logDeprecationsOnly = false;
+    public bool $logDeprecationsOnly = true;
 
     /**
      * --------------------------------------------------------------------------
--- a/app/Config/Exceptions.php
+++ b/app/Config/Exceptions.php
@@ -60,13 +60,13 @@ class Exceptions extends BaseConfig
      * Use this option to temporarily cease the warnings and instead log those.
      * This option also works for user deprecations.
      */
-    public bool $logDeprecationsOnly = true;
+    public bool $logDeprecations = true;
 
     /**
      * --------------------------------------------------------------------------
      * LOG LEVEL THRESHOLD FOR DEPRECATIONS
      * --------------------------------------------------------------------------
-     * If `$logDeprecationsOnly` is set to `true`, this sets the log level
+     * If `$logDeprecations` is set to `true`, this sets the log level
      * to which the deprecation will be logged. This should be one of the log
      * levels recognized by PSR-3.
      *
--- a/app/Config/Logger.php
+++ b/app/Config/Logger.php
@@ -38,7 +38,7 @@ class Logger extends BaseConfig
      *
      * @var array|int
      */
-    public $threshold = 4;
+    public $threshold = (ENVIRONMENT === 'production') ? 4 : 9;
 
     /**
      * --------------------------------------------------------------------------
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -83,8 +83,8 @@ class Exceptions
         if (! isset($this->config->sensitiveDataInTrace)) {
             $this->config->sensitiveDataInTrace = [];
         }
-        if (! isset($this->config->logDeprecationsOnly, $this->config->deprecationLogLevel)) {
-            $this->config->logDeprecationsOnly = false;
+        if (! isset($this->config->logDeprecations, $this->config->deprecationLogLevel)) {
+            $this->config->logDeprecations     = false;
             $this->config->deprecationLogLevel = LogLevel::WARNING;
         }
     }
@@ -164,7 +164,7 @@ class Exceptions
                 return true;
             }
 
-            if (! $this->config->logDeprecationsOnly || (bool) env('CODEIGNITER_SCREAM_DEPRECATIONS')) {
+            if (! $this->config->logDeprecations || (bool) env('CODEIGNITER_SCREAM_DEPRECATIONS')) {
                 throw new ErrorException($message, 0, $severity, $file, $line);
             }
 
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -158,7 +158,11 @@ class Exceptions
      */
     public function errorHandler(int $severity, string $message, ?string $file = null, ?int $line = null)
     {
-        if ($this->isDeprecationError($severity) && $this->config->logDeprecationsOnly) {
+        if ($this->isDeprecationError($severity)) {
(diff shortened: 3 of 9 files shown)


kenjis:
Why does this error occur?

```
1) CodeIgniter\Cookie\CookieStoreTest::testCookieDispatching
ErrorException: The CodeIgniter\Cookie\CookieStore::setCookie method is deprecated ().

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Cookie/CookieStore.php:174
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Cookie/CookieStoreTest.php:119
/home/runner/work/CodeIgniter4/CodeIgniter4/vendor/nexusphp/tachycardia/src/Expeditable.php:43
```
https://github.com/codeigniter4/CodeIgniter4/actions/runs/3343093343/jobs/5535960965

system/Debug/Exceptions.php
```diff
@@ -158,17 +158,22 @@ public function exceptionHandler(Throwable $exception)
...
+        if ($this->isDeprecationError($severity)) {
             // @TODO Remove if Faker is fixed.
-            if ($this->isFakerDeprecationError($severity, $message, $file, $line)) {
-                // Ignore the error.
+            if ($this->isFakerDeprecationError($message, $file, $line)) {
                 return true;
             }
 
+            if (! $this->config->logDeprecationsOnly
+                || filter_var(env('CODEIGNITER_SCREAM_DEPRECATIONS'), FILTER_VALIDATE_BOOLEAN)
```
MGatner, line 168: This seems unduly specific and conflicts with your statement of using any truthy value. How about simply:
```suggestion
                || env('CODEIGNITER_SCREAM_DEPRECATIONS')
```

app/Config/Exceptions.php
```diff
@@ -60,7 +60,7 @@ class Exceptions extends BaseConfig
      * Use this option to temporarily cease the warnings and instead log those.
      * This option also works for user deprecations.
      */
-    public bool $logDeprecationsOnly = false;
+    public bool $logDeprecationsOnly = true;
```
MGatner, line 63: I realize the name predates this PR but the phrasing seems to imply "only log things that are deprecations". How about simply "logDeprecations" or even "convertDeprecations"?

app/Config/Logger.php
```diff
@@ -38,7 +38,7 @@ class Logger extends BaseConfig
      *
      * @var array|int
      */
-    public $threshold = 4;
+    public $threshold = 5;
```
MGatner, line 41: I would like to see some more justification for a base log level. Logs are a type of persistence and connected to a lot of value in enterprise. Personally I'm for more logging (I always set 1,2,3,4,8 even in production).

MGatner:
> Why does this error occur?

I'm clueless! Maybe PHPUnit is trigger a deprecation warning? We know it parses docblock tags.

MGatner:
I didn't know this feature existed: https://phpunit.readthedocs.io/en/9.5/configuration.html#the-ignoredeprecatedcodeunits-attribute

Doesn't explicitly seem like it would actually trigger a deprecation though.

paulbalandan:
> Why does this error occur?

Apparently, PHPUnit parses `@deprecated` tags when mocking a method:
https://github.com/sebastianbergmann/phpunit/blob/851867efcbb6a1b992ec515c71cdcf20d895e9d2/src/Framework/MockObject/MockMethod.php#L123-L130

When it detects the tag, it generates the `@trigger_error()` in a template:
https://github.com/sebastianbergmann/phpunit/blob/851867efcbb6a1b992ec515c71cdcf20d895e9d2/src/Framework/MockObject/MockMethod.php#L205-L216

MGatner:
I think this one has to get into 4.3 because of #6705. Is that ground for pushing on #6710 as well? When remains to be done here?

paulbalandan:
> I think this one has to get into 4.3 because of #6705. Is that ground for pushing on #6710 as well? When remains to be done here?

I'm not sure what justification do you want for the increase in logged level. That's what remains here.

kenjis:
I don't think `$threshold` should be changed easily.

app/Config/Logger.php
paulbalandan: I've reverted for now the change in `$threshold`

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -220,7 +220,9 @@ Helpers and Functions
 Error Handling
 ==============
 
-- You can now log deprecation errors instead of throwing them. See :ref:`logging_deprecation_errors` for details.
+- You can now log deprecation warnings instead of throwing them. See :ref:`logging_deprecation_warnings` for details.
```
kenjis, line 223: Does *throwing deprecation warnings* make sense?
```suggestion
- You can now log deprecation warnings instead of throwing exceptions. See :ref:`logging_deprecation_warnings` for details.
```

MGatner:
How about some way of supporting different environment-specific levels? Like if `$threshold = null` then: `ENVIRONMENT === 'production' ? 4 : 9`?

kenjis:
It seems good! I always change it to 9 in development.

```diff
-    public $threshold = 4;
+    public $threshold = (ENVIRONMENT === 'production') ? 4 : 9;
```

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -221,7 +221,10 @@ Helpers and Functions
 Error Handling
 ==============
 
-- You can now log deprecation errors instead of throwing them. See :ref:`logging_deprecation_errors` for details.
+- You can now log deprecation warnings instead of throwing exceptions. See :ref:`logging_deprecation_warnings` for details.
+- Logging of deprecations is turned on by default.
+- To *temporarily* enable throwing of deprecations, set the environment variable ``CODEIGNITER_SCREAM_DEPRECATIONS`` to a truthy value.
+- ``Config\\Logger::$threshold`` is now, by default, environment-specific. For production environment, default threshold is still ``4`` but changed to ``9`` for other environments.
```
kenjis, line 227:
```suggestion
- ``Config\Logger::$threshold`` is now, by default, environment-specific. For production environment, default threshold is still ``4`` but changed to ``9`` for other environments.
```

paulbalandan: Thanks.

kenjis (approved):

MGatner (approved):
Thanks team! Very nice work @paulbalandan, glad to have this one in.

PR #6775, opened 2022-10-28
Delete admin/module directory

paulbalandan:
**Description**
If I'm not mistaken, this was used for the `devstarter` which was abandoned long ago so this needs not to be maintained any more.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


kenjis:
I thought it was a template for CI4 packages.

paulbalandan:
But was this ever used? I'm not sure either.

kenjis:
It seems `admin/module` is not documented anywhere, so maybe nobody knows it.

It would be better to have templates and instructions for those who make CI4 packages.

MGatner:
Let's remove it. It was indeed a template for libraries that I started but I quickly found it was easier to start from one of my existing libraries and stopped keeping up with it. Most of what Modules was trying to accomplish is now covered by DevKit, and the remaining piece (test scaffold) ends up getting in the way more than being helpful.

MGatner:
If we really want to offer this feature in the future it should be a GitHub Template Repository, which would take it out of this repo and make maintenance easier.

MGatner (approved):

kenjis (approved):

Issue #6746, opened 2022-10-24
Bug: Cache does not respect subdomain


c4user:
### PHP Version

7.4

### CodeIgniter4 Version

4.2.7

### CodeIgniter4 Installation Method

Manual (zip or tar.gz)

### Which operating systems have you tested for this bug?

Windows, Linux

### Which server did you use?

apache

### Database

MySQL 

### What happened?

Cache always creates a filename (key) using the hostname. For this reason, cache does not work when using a route with limit a subdomain. For example, when the cache is enabled, it doesn't matter whether it is stored for xxx.domain.com or yyy.domain.com. It shows the same content in both addresses. 

### Steps to Reproduce

-

### Expected Output

It is expected to produce content based on subdomains.

Cache must be compatible with subdomains.

kenjis:
I think #6785 fixes the bug.
Can you try `4.3` branch?

kenjis:
Closed by #6785

PR #6785, opened 2022-10-29
feat: add $allowedHostnames for multiple domain support

### LINKED ISSUE #5807: Bug: current_url() returning wrong value when different host

kenjis:
**Description**
Supersedes #6759
Fixes #5807
Related #6746

- add `Config\App::$allowedHostnames`
- add `URI::$baseURL`, `URI::setBaseURL()` and `URI::getBaseURL()`
- `IncomingRequest::__construct()` sets `URI::$baseURL`
  - use `HTTP_HOST` when it is an allowed hostname 
- refactor

**Changes:**
- `current_url()`, `base_url()`, `site_url()` will return the URL with allowed `HTTP_HOST`
- `previous_url()` will return the site URL with allowed `HTTP_HOST` when cannot get it from session nor `HTTP_REFERER`

The following functions/methods are affected (There may be others):
- Using `site_url()`
  - form_open()
  - audio()
  - embed()
  - img()
  - link_tag()
  - object()
  - script_tag()
  - source()
  - video()
  - RedirectResponse::route()
  - RedirectResponse::to()
  - Toolbar::prepare()
  - Toolbar::respond()
  - anchor()
  - anchor_popup()
  - base_url()
  - previous_url()
  - url_to()
- Using `base_url()`
  - RedirectException in CodeIgnter.php L377
  - UserAgent::isReferral()
- Using `current_url()`
  - ChromeLoggerHandler::__construct()
  - CodeIgniter::storePreviousURL()
  - Pager::ensureGroup()
  - Toolbar::run()

Ref #4651

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -25,6 +25,20 @@ class App extends BaseConfig
     public string $baseURL = 'http://localhost:8080/';
 
     /**
+     * Allowed Hostnames in the Site URL other than the hostname in the baseURL.
+     * If you want to accept multiple Hostnames, set this.
+     *
+     * E.g. When your site URL ($baseURL) is 'http://example.com/', and your site
+     *      also accepts 'http://media.example.com/' and
+     *      'http://accounts.example.com/':
+     *          ['media.example.com', 'accounts.example.com']
+     *
+     * @var string[]
+     * @phpstan-var list<string>
+     */
+    public array $allowedHostnames = [];
+
+    /**
      * --------------------------------------------------------------------------
      * Index File
      * --------------------------------------------------------------------------
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -400,8 +400,8 @@ class IncomingRequest extends Request
         // baseURL, so let's help them out.
         $baseURL = ($config->baseURL === '') ? $config->baseURL : rtrim($config->baseURL, '/ ') . '/';
 
-        // Based on our baseURL provided by the developer
-        // set our current domain name, scheme
+        // Based on our baseURL and allowedHostnames provided by the developer
+        // and HTTP_HOST, set our current domain name, scheme.
         if ($baseURL !== '') {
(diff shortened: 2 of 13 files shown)


MGatner (approved):
I think this is a better solution. Thank you for enumerating the impacted methods and components: that would be a helpful addition to the changelogs. 

One thing I would like to note: for a while now I have been a proponent of a separate class for "internal URIs". Using one class for internal logic and hostname resolution **and** normal URL parsing & passing... it seems to be multiple responsibilities with more room for issues. Past conversations have concluded that maintaining multiple URI classes is overkill and confusing. This PR mixes concerns even more so I would like to bring this conversation back up again for v5.

kenjis:
Added the docs.

datamweb (approved):
Excellent, this feature always felt empty.

MGatner (approved):

PR #6802, opened 2022-11-01
Add a check for updated version before deployment

paulbalandan:
**Description**
"Successful" run: https://github.com/paulbalandan/CodeIgniter4/actions/runs/3369749982/jobs/5589787478

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


.github/workflows/deploy-distributables.yml
```diff
@@ -1,16 +1,42 @@
...
+      - name: Get latest version
+        run: |
+          echo 'LATEST_VERSION<<EOF' >> $GITHUB_ENV
+          echo $(git describe --tags --abbrev=0) | sed "s/v//" >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+
+      - name: Search for updated version
+        if: ${{ env.LATEST_VERSION }}
+        run: |
+          if ! grep -rq system/CodeIgniter.php user_guide_src/source/conf.py -e "${{ env.LATEST_VERSION }}"; then
```
kenjis, line 29: 1. Why `-r`?
2. Shouldn't we check both `system/CodeIgniter.php` and`user_guide_src/source/conf.py` have the latest version? Now it checks `system/CodeIgniter.php` or`user_guide_src/source/conf.py`.

paulbalandan: > 1. Why `-r`?

That `-r` is not needed. I'll remove.

> 2. Shouldn't we check both `system/CodeIgniter.php` and`user_guide_src/source/conf.py` have the latest version? Now it checks `system/CodeIgniter.php` or`user_guide_src/source/conf.py`.

Oh, you're right. I thought it checks for both presence.

.github/workflows/deploy-distributables.yml
```diff
@@ -1,16 +1,50 @@
-# When a new Release is created, deploy relevant
+# When a new release is created, deploy relevant
 # files to each of the generated repos.
-name: Deploy Framework
+name: Deploy Distributable Repos
```
MGatner, line 3: Please update this name in admin/RELEASE.md as well.

MGatner (commented):
I brought this up on the release thread in response to a contributor comment, but we need to decide what a "hotfix" looks like. This seems to be a vote for "go back through the entire release process" so that our version numbers are always consistent. I think that is probably the best approach, since we're not actually pushing anything to "production" to begin with. Alternatives would be to release sub-versions like 4.2.8.1 or 4.2.8-hotfix, or to delete tags from all distributable repos and redo the release.

kenjis:
If we are going to follow [Semantic Versioning](https://semver.org/), I think we cannot use 4.2.8.1 or 4.2.8-hotfix.
Also, we cannot remove the tags.
Am I wrong?

paulbalandan:
I think redoing the release is the best approach, but I'm not sure if the new tag's SHA will be picked up by packagist.

MGatner:
Packagist will update if you remove a tag and re-release it. This is my approach on small libraries when I make a mistake. We could do that here as well, just have to delete the tags from the distributable repos too. 

Sticking with SemVer is best. It looks like we could append "build metadata" like `v4.0.8+hotfix.1` but that seems more of a hack than a supported approach. So let's go with new release or removing tags.

kenjis:
When we use this PR,
if we forget to update the version number, the framework deploy will stop.
But in this repository the release was done already?

paulbalandan:
> ... But in this repository the release was done already?

Yes, because the release payload from this repo will be used in creating the respective releases in the 3 repos through `actions/github-script`

MGatner:
I think that's fine. Deleting tags isn't as big a deal as it seems, and if the Action stops it from happening on the distributables then we are pretty much guaranteed nobody has been affected.

MGatner (approved):

kenjis:
Yes, this repository is not a Composer package.
Removing tags/releases is not a big deal if we do it quickly.

kenjis (approved):

PR #6805, opened 2022-11-01
Enable xdebug regardless of coverage option

paulbalandan:
**Description**
Apparently, `CIUnitTestCase::assertHeaderEmitted()` and `CIUnitTestCase::assertHeaderNotEmitted()` requires `xdebug_get_headers()` but we disabled `xdebug` on PHP_VERSION !== '8.1'.

Not sure if this would work. 🤞 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/.github/workflows/test-phpunit.yml
+++ b/.github/workflows/test-phpunit.yml
@@ -125,7 +125,7 @@ jobs:
         with:
           php-version: ${{ matrix.php-versions }}
           tools: composer, pecl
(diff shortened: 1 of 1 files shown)


paulbalandan:
It seems this approach is not yet supported by `setup-php`.
https://github.com/shivammathur/setup-php/blob/1a18b2267f80291a81ca1d33e7c851fe09e7dfc4/src/coverage.ts#L148-L149

PR #6817, opened 2022-11-05
4.2.10 Ready code

kenjis:


MGatner:
> Failed asserting that 4 is identical to 5.

Is this related to the `Time::now()` changes?

kenjis:
No. Time::now() change is only in 4.3.
It seems due to Daylight Saving Time.
See #6818

kenjis:
The failed test is because of DST. It is not a new bug or I don't know it is a bug.
We proceed this release.

PR #6820, opened 2022-11-06
fix:  Request::getIPAddress()

kenjis:
**Description**
- use `header()` instead of `getServer()` for header values
   - getIPAddress() does not work with OpenSwoole.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


tests/system/HTTP/IncomingRequestTest.php
```diff
@@ -709,66 +712,75 @@ public function testGetIPAddressNormal()
...
         $this->assertSame($expected, $this->request->getIPAddress());
     }
 
     public function testGetIPAddressThruProxyInvalid()
     {
-        $expected                        = '123.456.23.123';
-        $_SERVER['REMOTE_ADDR']          = '10.0.1.200';
+        $_SERVER['HTTP_X_FORWARDED_FOR'] = '123.456.23.123';
+        $expected                        = '10.0.1.200';
+        $_SERVER['REMOTE_ADDR']          = $expected;
```
MGatner, line 731: Double assign?
```suggestion
        $_SERVER['HTTP_X_FORWARDED_FOR'] = '123.456.23.123';
        $_SERVER['REMOTE_ADDR']          = $expected = '10.0.1.200';
```

tests/system/HTTP/MessageTest.php
```diff
@@ -30,12 +30,6 @@ protected function setUp(): void
         $this->message = new Message();
     }
 
-    protected function tearDown(): void
-    {
-        $this->message = null;
-        unset($this->message);
-    }
-
```
MGatner, line 38: What a strange idea this was! 🤔

MGatner (approved):

MGatner:
Meant to "leave a comment".

AmitSonkhiya:
I have copied this code snippet for the `getIPAddress()` method and it is now correctly determined the IP address.
Just I had to add `127.0.0.1` in the `$proxyIPs` property of the `App` class since the server resides in the internet space.
It would be great if `x-real-ip` could be added as well in the method.

tests/system/HTTP/IncomingRequestTest.php
kenjis: This code seems a bit odd to me, because of the spaces to the left of the first `=`.

kenjis:
@AmitSonkhiya Thank you for testing!

tests/system/HTTP/IncomingRequestTest.php
MGatner: Very not necessary.

PR #6827, opened 2022-11-07
fix: Model cannot insert when $useAutoIncrement is false

kenjis:
**Description**
From https://forum.codeigniter.com/showthread.php?tid=84733

- The `$primaryKey` field should never be an allowed field.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -710,8 +710,13 @@ class Model extends BaseModel
 
         // Always grab the primary key otherwise updates will fail.
         if (
-            method_exists($data, 'toRawArray') && (! empty($properties) && ! empty($this->primaryKey) && ! in_array($this->primaryKey, $properties, true)
-            && ! empty($data->{$this->primaryKey}))
+            method_exists($data, 'toRawArray')
+            && (
+                ! empty($properties)
+                && ! empty($this->primaryKey)
+                && ! in_array($this->primaryKey, $properties, true)
+                && ! empty($data->{$this->primaryKey})
+            )
         ) {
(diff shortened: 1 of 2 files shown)


system/Model.php
```diff
@@ -118,6 +118,13 @@ class Model extends BaseModel
      */
     protected $escape = [];
 
+    /**
+     * Primary Key data when inserting and useAutoIncrement is false.
+     *
+     * @var int|string|null
+     */
+    private $tempPrimaryKeyData;
```
MGatner, line 126: Why "data"? If differentiating from the name I think "value" would be more consistent with what we usually do for a single item.

MGatner (commented):
This is a good change. I've been using a workaround in a base Model in a project that uses UUIDs so I didn't even notice that this was an issue!

system/Model.php
kenjis: Yes, *value* is better.
Fixed.

MGatner (approved):

michalsn (approved):

MaulanaMalikS:
This also happens in the model's `insertBatch()` function.
Is there any progress on resolving this issue?

kenjis:
@MaulanaMalikS If you find another bug, please report it.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/README.md

MaulanaMalikS:
@kenjis I've made an issue for this #7744 .

PR #6833, opened 2022-11-08
fix: `composer update` might cause error "Failed to open directory"

### LINKED ISSUE #6832: `composer update` fails when trying to delete a non-existent folder
### PHP Version

8.1

### CodeIgniter4 Version

4.2.10

Manual (zip or tar.gz)

Windows, Linux

### Which server did you use?

fpm-fcgi

### Database

_No response_

### What happened?

During a `composer update`, the function "recursiveDelete()" fail because the folder doesn't exist and block the update process.

```
$ php composer.phar update
Loading composer repositories with package information
Updating dependencies
Nothing to modify in lock file
Installing dependencies from lock file (including require-dev)
Nothing to install, update or remove
Generating autoload files
50 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
> CodeIgniter\ComposerScripts::postUpdate
Cannot recursively delete "/home/myst/xxxxxxx/system/ThirdParty/" as it does not exist.Script CodeIgniter\ComposerScripts::postUpdate handling the post-update-cmd event terminated with an exception

In ComposerScripts.php line 95:

  RecursiveDirectoryIterator::__construct(/home/myst/xxxxxxx/system/ThirdParty): Failed to open directory: No such file or directory


update [--with WITH] [--prefer-source] [--prefer-dist] [--prefer-install PREFER-INSTALL] [--dry-run] [--dev] [--no-dev] [--lock] [--no-install] [--no-autoloader] [--no-suggest] [--no-progress] [-w|--with-dependencies] [-W|--with-all-dependencies] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--apcu-autoloader-prefix APCU-AUTOLOADER-PREFIX] [--ignore-platform-req IGNORE-PLATFORM-REQ] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [-i|--interactive] [--root-reqs] [--] [<packages>...]
```

### Steps to Reproduce

Delete the folder "system/ThirdParty" and run the command `composer update`

### Expected Output

Only the `Cannot recursively delete "/home/myst/xxxxxxx/system/ThirdParty/" as it does not exist.` message without the Exception.

### Anything else?

Can be solved by simply add a `return;` in the `if` of `recursiveDelete()`.

LeMyst:
Fixes #6832

**Description**
Exit the function recursiveDelete() if the folder to delete didn't exist.
Fixes #6832 

**Checklist:**
- [x] Securely signed commits
- [?] Component(s) with PHPDoc blocks, only if necessary or adds value
- [?] Unit testing, with >80% coverage
- [?] User guide updated
- [?] Conforms to style guide

### DIFF
--- a/system/ComposerScripts.php
+++ b/system/ComposerScripts.php
@@ -129,7 +129,7 @@ final class ComposerScripts
         }
 
         if (! @mkdir($targetDir, 0755, true)) {
-            echo sprintf('Cannot create the target directory: "%s"' . PHP_EOL, $targetDir);
+            echo sprintf('Cannot create the target directory: "%s"', $targetDir) . PHP_EOL;
 
             exit(1);
         }
--- a/system/ComposerScripts.php
+++ b/system/ComposerScripts.php
@@ -87,7 +87,7 @@ final class ComposerScripts
     private static function recursiveDelete(string $directory): void
     {
         if (! is_dir($directory)) {
-            echo sprintf('Cannot recursively delete "%s" as it does not exist.', $directory);
+            echo sprintf('Cannot recursively delete "%s" as it does not exist.', $directory) . PHP_EOL;
 
             return;
         }
--- a/system/ComposerScripts.php
+++ b/system/ComposerScripts.php
@@ -128,7 +128,11 @@ final class ComposerScripts
             exit(1);
         }
 
-        @mkdir($targetDir, 0755, true);
+        if (! @mkdir($targetDir, 0755, true)) {
+            echo sprintf('Cannot create the target directory: "%s"' . PHP_EOL, $targetDir);
+
+            exit(1);
+        }
 
         $dirLen = strlen($originDir);
 
--- a/system/ComposerScripts.php
+++ b/system/ComposerScripts.php
@@ -88,6 +88,8 @@ final class ComposerScripts
     {
         if (! is_dir($directory)) {
             echo sprintf('Cannot recursively delete "%s" as it does not exist.', $directory);
+
+            return;
         }
 
         /** @var SplFileInfo $file */


kenjis:
Wouldn't it be better to leave it as an exception since this is an exceptional situation?

MGatner:
It looks like `recursiveMirror()` will fail if the directory isn't there (not 100% sure on that). If that's the case I'd be in favor of exiting early here, like @kenjis suggests.

LeMyst:
> Wouldn't it be better to leave it as an exception since this is an exceptional situation?
> 
> Should we throw an appropriate exception?


> It looks like `recursiveMirror()` will fail if the directory isn't there (not 100% sure on that). If that's the case I'd be in favor of exiting early here, like @kenjis suggests.


Hello @kenjis and @MGatner ,

If the recursiveDelete fails with the exception, the recursiveMirror is not executed and the `composer update` is not finalized.

It's better to not throw an Exception, to let the recursiveMirror do the job and correctly finish the update.

sfadschm:
I would agree with @LeMyst, here. The proposed change does not alter the behaviour of the method, but rather provides an early exit to skip the unnecessary generation of Iterator objects if the folder does not exist.

kenjis:
@LeMyst Please don't use `git merge`.

LeMyst:
> @LeMyst Please don't use `git merge`.
> See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#updating-your-branch
> 

I saw that after my action. I will clean my branch tomorrow.

system/ComposerScripts.php
```diff
@@ -88,6 +88,8 @@ private static function recursiveDelete(string $directory): void
     {
         if (! is_dir($directory)) {
```
kenjis, line 89:
```suggestion
        if (! file_exists($directory)) {
```

kenjis:
Yes, if the script is terminated by some reasons right after `recursiveDelete()` execution,
`system/ThirdParty/` does not exist when the next execution of `composer update`.

system/ComposerScripts.php
paulbalandan: Why suggest `file_exists()`? Will the iterator not throw an error if value passed is a file?

LeMyst: On Unix, with the trailing '/', PHP search for a directory.
I tried to delete ThirdParty and `touch` a file named ThirdParty, and the function fails because of the trailing '/'.

LeMyst:
> @LeMyst Please don't use `git merge`. See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#updating-your-branch

It's cleaned

system/ComposerScripts.php
kenjis: Oh, I missed the trailing `/`.

system/ComposerScripts.php
```diff
@@ -88,6 +88,8 @@ private static function recursiveDelete(string $directory): void
...
             echo sprintf('Cannot recursively delete "%s" as it does not exist.', $directory);
```
kenjis, line 90:
```suggestion
            echo sprintf('Cannot recursively delete "%s" as it does not exist.', $directory) . PHP_EOL;
```

kenjis:
How about this?
```diff
@@ -126,7 +128,11 @@ final class ComposerScripts
             exit(1);
         }
 
-        @mkdir($targetDir, 0755, true);
+        if (! @mkdir($targetDir, 0755, true)) {
+            echo sprintf('Cannot create the target directory: "%s"' . PHP_EOL, $targetDir);
+
+            exit(1);
+        }
 
         $dirLen = strlen($originDir);
 
```

paulbalandan (approved):

LeMyst:
> How about this?
> 
> ```diff
> @@ -126,7 +128,11 @@ final class ComposerScripts
>              exit(1);
>          }
>  
> -        @mkdir($targetDir, 0755, true);
> +        if (! @mkdir($targetDir, 0755, true)) {
> +            echo sprintf('Cannot create the target directory: "%s"' . PHP_EOL, $targetDir);
> +
> +            exit(1);
> +        }
>  
>          $dirLen = strlen($originDir);
>  
> ```

Does I need to put this change (who's good for me) into this PR?

kenjis:
> I tried to delete ThirdParty and touch a file named ThirdParty, and the function fails because of the trailing '/'.

Without the above code, `composer update` still raises an error in that case.

LeMyst:
> 

It's added to the PR, but that raise a question about consistency for the PHP_EOL. Inside or outside the sprintf()?

kenjis:
Thank you. And good point.
Outside the sprintf() is easier to read?

paulbalandan:
I would add that to the format string, because it's supposed to be part of the message.

LeMyst:
Another question I've, it's the difference between `composer update` and `composer install`

If I do a `composer update`, the hook `post-update-cmd` is used and the ThirdParty dependencies are updated.
But if I do a `composer install`, if I use the `composer.lock` on my production line, the ThirdParty folder is not updated.

I think a hook `post-install-cmd` must be added.

kenjis:
> I think a hook post-install-cmd must be added.

No. The third party libraries are updated only when you run `composer update`.
`composer install` does not update them. So the hook is not needed.

system/ComposerScripts.php
```diff
@@ -126,7 +128,11 @@ private static function recursiveMirror(string $originDir, string $targetDir): v
             exit(1);
         }
 
-        @mkdir($targetDir, 0755, true);
+        if (! @mkdir($targetDir, 0755, true)) {
+            echo sprintf('Cannot create the target directory: "%s"' . PHP_EOL, $targetDir);
```
kenjis, line 132:
```suggestion
            echo sprintf('Cannot create the target directory: "%s"', $targetDir) . PHP_EOL;
```

LeMyst:
> > I think a hook post-install-cmd must be added.
> 
> No. The third party libraries are updated only when you run `composer update`. `composer install` does not update them. So the hook is not needed.

It seems there is different behavior:
- If there is no composer.lock, the `composer install` run the `ComposerScripts::postUpdate`
- if you use the `composer update`, the `ComposerScripts::postUpdate` is ran too
- but if you use `composer install` with a composer.lock, the `ComposerScripts::postUpdate` is not ran and the libraries are not copied/updated in the ThirdParty folder.

You have more informations about this here: https://getcomposer.org/doc/articles/scripts.md#command-events

```
post-install-cmd: occurs after the install command has been executed with a lock file present.
post-update-cmd: occurs after the update command has been executed, or after the install command has been executed without a lock file present.
```

So I think a hook on post-install-cmd is necessary to copy the ThirdParty libraries.

kenjis (approved):

kenjis:
> So I think a hook on post-install-cmd is necessary to copy the ThirdParty libraries.

Why?
There is already latest Third party source code in the ThirdParty folder.

LeMyst:
> > So I think a hook on post-install-cmd is necessary to copy the ThirdParty libraries.
> 
> Why? There is already latest Third party source code in the ThirdParty folder.

If you add dependencies to your composer.json/composer.local.json and you push the composer.lock to your version control. When you update the vendor folder with `composer install` on your deployment server (production, preproduction, ...), the folder `system/ThirdParty` will not be updated.

MGatner:
Correct me if I'm wrong but... if we're using Composer then we don't even _need_ to update **ThirdParty** because the Composer sources will be used.

paulbalandan:
To be honest, I think if you are using manual install then you don't need ComposerScripts and the post update hook altogether. It's kinda weird you download the zip manually then use Composer to update. I mean, why not use composer initially?

LeMyst:
> Correct me if I'm wrong but... if we're using Composer then we don't even _need_ to update **ThirdParty** because the Composer sources will be used.

I don't understand why this is in place, I did not test to remove the ThirdParty folder when you have an up-to-date vendor folder.

> To be honest, I think if you are using manual install then you don't need ComposerScripts and the post update hook altogether. It's kinda weird you download the zip manually then use Composer to update. I mean, why not use composer initially?

I am probably not the best example. I build my project around CodeIgniter. I download the zip file and start my project from this, adding (or not) my dependencies to the composer.local.json, etc.
I never really used the AppStarter, so I'm between the two worlds (zip and composer), I don't think I'm a rare case.

kenjis:
@LeMyst Sorry, if you downloaded the Zip file from this `codeigniter4/CodeIgniter4` repository, it was because the official site download link was wrong in the past. Correct zip file is one from `codeigniter4/framework` like https://api.github.com/repos/codeigniter4/framework/zipball/v4.2.10 

If you use Zip installation, you need to replace `system` folder manually. And you don't need to install the third party packages via Composer. See https://codeigniter4.github.io/CodeIgniter4/installation/installing_manual.html#installing-manual-upgrading

But If you installed the package with Composer, the Composer package will be autoloaded by CI4's autoloader. `ThirdParty/` source code is not used.

If you use Composer, we recommend you don't use Zip installation.

I recommend you move to [Adding CodeIgniter4 to an Existing Project](https://codeigniter4.github.io/CodeIgniter4/installation/installing_composer.html#adding-codeigniter4-to-an-existing-project). Then you can update `system` folder with `composer update`. Your `composer.json` file will be similar to appstar's: https://github.com/codeigniter4/appstarter/blob/master/composer.json that does not have `post-update-cmd`.

The `ComposerScripts.php` is for framework maintainers, not for end users.

LeMyst:
> @LeMyst Sorry, if you downloaded the Zip file from this `codeigniter4/CodeIgniter4` repository, it was because the official site download link was wrong in the past. Correct zip file is one from `codeigniter4/framework` like https://api.github.com/repos/codeigniter4/framework/zipball/v4.2.10
> 
> If you use Zip installation, you need to replace `system` folder manually. And you don't need to install the third party packages via Composer. See https://codeigniter4.github.io/CodeIgniter4/installation/installing_manual.html#installing-manual-upgrading
> 
> But If you installed the package with Composer, the Composer package will be autoloaded by CI4's autoloader. `ThirdParty/` source code is not used.
> 
> If you use Composer, we recommend you don't use Zip installation.
> 
> I recommend you move to [Adding CodeIgniter4 to an Existing Project](https://codeigniter4.github.io/CodeIgniter4/installation/installing_composer.html#adding-codeigniter4-to-an-existing-project). Then you can update `system` folder with `composer update`. Your `composer.json` file will be similar to appstar's: https://github.com/codeigniter4/appstarter/blob/master/composer.json that does not have `post-update-cmd`.
> 
> The `ComposerScripts.php` is for framework maintainers, not for end users.

OK, I will do this next time.

Maybe remove the `composer.json` from the framework release if it's not intended to be used.

Otherwise, I think we can close this PR if the issue #6832 is solved?

kenjis:
> Maybe remove the composer.json from the framework release if it's not intended to be used.

If devs want to write PHPUnit tests, they need `composer.json` to install PHPUnit.
After all many devs need  `composer.json`.

We actually recommend Composer installation.

@codeigniter4/core-team 
But it seems we do not need the following in `codeigniter4/framework` composer.json:
```
        "post-update-cmd": [
            "CodeIgniter\\ComposerScripts::postUpdate"
        ],
```

PR #6839, opened 2022-11-10
Handle key generation when key is not present in .env

### LINKED ISSUE #6838: Bug: `spark key:generate` doesn't generate `encryption.key` if no template
### PHP Version

7.4

### CodeIgniter4 Version

4.2.10

Windows

### Which server did you use?

### Database

_No response_

### What happened?

Try deleting all entries in `.env` so you can sort out which ones to use.
But when running `spark key:generate` no `encryption.key` is created even though the message on the shell is successful.

### Steps to Reproduce

remove all `.env` then run `php spark key:generate`

### Expected Output

In `.env` should create `encryption.key =` even without template

### Anything else?

_No response_

paulbalandan:
**Description**
Fixes #6838

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Encryption/GenerateKey.php
+++ b/system/Commands/Encryption/GenerateKey.php
@@ -67,6 +67,7 @@ class GenerateKey extends BaseCommand
     public function run(array $params)
     {
         $prefix = $params['prefix'] ?? CLI::getOption('prefix');
+
         if (in_array($prefix, [null, true], true)) {
             $prefix = 'hex2bin';
         } elseif (! in_array($prefix, ['hex2bin', 'base64'], true)) {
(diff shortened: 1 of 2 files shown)


paulbalandan:
`key:generate` relies on `env()` to get the replaceable key. If the key is commented out, `env()` will return `''`.
```php
$currentKey = env('encryption.key', '');
```

paulbalandan:
Okay. I got what you mean. This is because `preg_replace()` did not find the old key in the file because it thinks it is an empty string.

paulbalandan:
@totoprayogo1916 Kindly check new code

datamweb (approved):

samsonasik (approved):

samsonasik:
Thank you @paulbalandan

PR #6846, opened 2022-11-13
test: fix DeleteModelTest

kenjis:
**Description**
- fix broken test
- add test case

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Models/DeleteModelTest.php
+++ b/tests/system/Models/DeleteModelTest.php
@@ -186,12 +186,14 @@ final class DeleteModelTest extends LiveModelTestCase
      */
     public function testDontDeleteRowsWhenSoftDeleteParamIsEmpty($emptyValue): void
     {
-        $this->expectException(DatabaseException::class);
-        $this->expectExceptionMessage('Deletes are not allowed unless they contain a "where" or "like" clause.');
-
         $this->seeInDatabase('user', ['name' => 'Derek Jones', 'deleted_at IS NULL' => null]);
 
-        $this->createModel(UserModel::class)->delete($emptyValue);
+        try {
+            $this->createModel(UserModel::class)->delete($emptyValue);
+        } catch (DatabaseException $e) {
+            // Do nothing.
+        }
+
         $this->seeInDatabase('user', ['name' => 'Derek Jones', 'deleted_at IS NULL' => null]);
     }
 
--- a/tests/system/Models/DeleteModelTest.php
+++ b/tests/system/Models/DeleteModelTest.php
@@ -153,7 +153,7 @@ final class DeleteModelTest extends LiveModelTestCase
      *
      * @dataProvider emptyPkValues
      *
-     * @param mixed $emptyValue
+     * @param int|string|null $emptyValue
      */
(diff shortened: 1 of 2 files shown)


tests/system/Models/DeleteModelTest.php
```diff
@@ -148,8 +148,8 @@ public function testOnlyDeleted(): void
     }
 
     /**
-     * If where condition is set, beyond the value was empty (0,'', NULL, etc.),
-     * Exception should not be thrown because condition was explicity set
+     * If where condition is set, beyond the value was empty (0, '', NULL, etc.),
+     * Exception should not be thrown because condition was explicitly set
```
kenjis, line 152: What does "beyond the value was empty" mean?

MGatner: Not sure! It seems to be explicitly allowing falsy values?

MGatner (approved):

tests/system/Models/DeleteModelTest.php
kenjis: I added `''` as a test case,  because it is here in the comment.
But Postgres does not accept it.

```
1) CodeIgniter\Models\DeleteModelTest::testDontThrowExceptionWhenSoftDeleteConditionIsSetWithEmptyValue with data set #3 ('')
ErrorException: pg_query(): Query failed: ERROR:  invalid input syntax for type integer: ""
LINE 2: WHERE "id" = ''
                     ^

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/Postgre/Connection.php:140
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:666
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseConnection.php:593
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Database/BaseBuilder.php:2079
/home/runner/work/CodeIgniter4/CodeIgniter4/system/Model.php:412
/home/runner/work/CodeIgniter4/CodeIgniter4/system/BaseModel.php:985
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Models/DeleteModelTest.php:163
```

kenjis: Fixed the comment, and removed the failed test case (`''`).

tests/system/Models/DeleteModelTest.php
```diff
@@ -148,12 +148,12 @@ public function testOnlyDeleted(): void
     }
 
     /**
-     * If where condition is set, beyond the value was empty (0,'', NULL, etc.),
-     * Exception should not be thrown because condition was explicity set
+     * If where condition is set, even if the value is empty (0, NULL, etc.),
+     * Exception should not be thrown because condition is explicitly set.
```
MGatner, line 152: Maybe refactor this? Doesn't really matter, it is a comment on a test method, but it is difficult to read. Gherkin syntax:
```suggestion
     * Given an explicit empty value in the WHERE condition
     * When executing a soft delete
     * Then an exception should not be thrown
```

PR #6848, opened 2022-11-13
docs: add a note about setting the null value for the primary key in Model::update()

michalsn:
**Description**
This behavior can be confusing for many users. It should be clearly described in the user guide.

Ref, #4617

### DIFF
--- a/user_guide_src/source/models/model.rst
+++ b/user_guide_src/source/models/model.rst
@@ -332,7 +332,7 @@ of the columns in a ``$table``, while the array's values are the values to save 
 
 .. literalinclude:: model/016.php
 
-.. note:: If the ``$primaryKey`` field is set to ``null`` then the update will affect all records in the table.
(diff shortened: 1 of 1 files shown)


kenjis (approved):

user_guide_src/source/models/model.rst
```diff
@@ -332,6 +332,8 @@ of the columns in a ``$table``, while the array's values are the values to save
 
 .. literalinclude:: model/016.php
 
+.. note:: If the ``$primaryKey`` field is set to ``null`` then the update will affect all records in the table.
```
paulbalandan, line 335: Should this be of higher importance, like `important::` or `warning::`

MGatner (approved):
I like Paul's suggestion. But generally: good addition!

user_guide_src/source/models/model.rst
michalsn: Good point.

paulbalandan (approved):

Issue #6815, opened 2022-11-04
Dev: replace `time()` with Time


kenjis:
Replace `time()` with `Time` class.
`time()` is not testable.

```php
time()
// ↓
Time::now()->getTimestamp()
```

See https://github.com/codeigniter4/CodeIgniter4/pull/6753#pullrequestreview-1168235642

```console
$ grep -lr 'time()' system
system/Database/MigrationRunner.php
system/Cache/Handlers/FileHandler.php
system/Cache/Handlers/PredisHandler.php
system/Cache/Handlers/WincacheHandler.php
system/Cache/Handlers/MemcachedHandler.php
system/Cache/Handlers/RedisHandler.php
system/Throttle/Throttler.php
system/Test/Fabricator.php
system/Test/TestResponse.php
system/Test/Mock/MockSession.php
system/Test/Mock/MockCache.php
system/Security/Security.php
system/Images/Handlers/ImageMagickHandler.php
system/HTTP/ResponseTrait.php
system/Files/File.php
system/Cookie/Cookie.php
system/Helpers/date_helper.php
system/ThirdParty/Kint/Parser/MicrotimePlugin.php
system/Email/Email.php
system/Debug/Toolbar/Collectors/Database.php
system/Debug/Toolbar.php
system/BaseModel.php
system/Session/Session.php
system/Session/Handlers/FileHandler.php
system/Session/Handlers/MemcachedHandler.php
system/Session/Handlers/RedisHandler.php
```

ddevsr:
Updated :
```
$ grep -lr 'time()' system
system/Cache/Handlers/FileHandler.php
system/Database/MigrationRunner.php
system/Test/Fabricator.php
system/Test/Mock/MockCache.php
system/Test/Mock/MockSession.php
system/Test/TestResponse.php
system/ThirdParty/Kint/Parser/MicrotimePlugin.php
```

kenjis:
`system/ThirdParty/Kint/Parser/MicrotimePlugin.php` is not our code.
We don't edit it.

ping-yee:
Does anyone can approve the last PR #6880 ?
If this PR get merged, this issue will be solved.

kenjis:
```console
$ grep -lr 'time()' system
system/Cache/Handlers/PredisHandler.php
system/Throttle/Throttler.php
system/ThirdParty/Kint/Parser/MicrotimePlugin.php
system/Debug/Toolbar.php
```

[...]

PR #6850, opened 2022-11-13
refactor: make `time()` to class `Time` in `Cache`

ddevsr:
**Description**
See #6815

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -155,6 +155,8 @@ parameters:
     API:
       - Format
       - HTTP
+    Cache:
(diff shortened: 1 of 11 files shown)


MGatner (approved):
Looks good! Thanks for taking this on. I think the "assert less than" calls could be controlled now with a fixed test time, but probably not necessary for the scope.

kenjis (approved):

PR #6852, opened 2022-11-13
refactor: replace time() with Time into `date_helper` and `Email` class

ping-yee:
**Description**
See #6815 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -162,6 +162,7 @@ parameters:
       - Entity
       - Events
     Email:
+      - I18n
       - Events
     Entity:
       - I18n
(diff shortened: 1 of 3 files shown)


ping-yee:
This is off topic, I have a question that what function in `time` class replace the `microtime()` in `system/Debug/Toolbar/Collectors/Database.php` is more appropriate?

kenjis:
Since PHP 7.1, [DateTimeImmutable](https://www.php.net/manual/en/class.datetimeimmutable.php) has microseconds when constructed from the current time, either explicitly or with a relative string (e.g. "first day of next month").
See https://www.php.net/manual/en/migration71.incompatible.php#migration71.incompatible.datetime-microseconds

We can use `u` in format:
See https://www.php.net/manual/en/datetime.format.php

kenjis (approved):

MGatner (approved):
Looks good!

PR #6853, opened 2022-11-13
Fix ShowTableInfoTest to pass on Windows

paulbalandan:
**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Commands/Database/ShowTableInfoTest.php
+++ b/tests/system/Commands/Database/ShowTableInfoTest.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter\Commands\Database;
 
+use CodeIgniter\CLI\CLI;
 use CodeIgniter\Test\CIUnitTestCase;
 use CodeIgniter\Test\DatabaseTestTrait;
 use CodeIgniter\Test\Filters\CITestStreamFilter;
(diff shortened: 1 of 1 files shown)


tests/system/Commands/Database/ShowTableInfoTest.php
```diff
@@ -64,11 +64,11 @@ public function testDbTable(): void
...
-        $expected = <<<'EOL'
-            +----+----------------+--------------------+-------+---------------+------------+-------+
-            | id | version        | class              | group | namespace     | time       | batch |
-            +----+----------------+--------------------+-------+---------------+------------+-------+
-            EOL;
+        $expected = implode(PHP_EOL, [
+            '+----+----------------+--------------------+-------+---------------+------------+-------+',
+            '| id | version        | class              | group | namespace     | time       | batch |',
+            '+----+----------------+--------------------+-------+---------------+------------+-------+',
+        ]);
```
kenjis, line 71: Writing this `$expected` is a bit pain. 
The reason to use `<<<'EOL'` is that we want to just copy and paste the output.

The newline code in the source is always `\n`.
Why don't you add a method to convert newline code?

paulbalandan: Thanks! I myself is struggling with these `implode()` calls. 😂

kenjis (approved):
This is good!

datamweb (approved):

PR #6855, opened 2022-11-13
Fix MigrationRunnerTest for Windows

paulbalandan:
**Description**
The different directory separators mess up local testing

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner (approved):
Usually I don't like `realpath` calls without fallback, because they fail in virtual file systems, but I think this is safe.

paulbalandan:
Yes, but `realpath()` returning `false` will be beneficial in this case when the file is missing as the test would catch that.

kenjis (approved):

Issue #6842, opened 2022-11-12
Bug: CURLRequest::request() cannot set int type header value


gentsir:
### PHP Version

7.3

### CodeIgniter4 Version

4.1.1

### CodeIgniter4 Installation Method

Manual (zip or tar.gz)

### Which operating systems have you tested for this bug?

Windows, Linux

### Which server did you use?

fpm-fcgi

### Database

MySQL 5.6

### What happened?

when i set some headers options to call `CURLRequest::request` method，i found that the  int type value is filtered. the last version also has this issue.

### Steps to Reproduce

the headers option:
```php
$url = 'https://192.168.8.119/api/v1/tenants';

$headers = [
    "Accept" => "application/json",
    "Accept-Language" => "zh-Hans",
    "Authorization" => "Bearer 6k8KGSDIGEz8y8QEftRIcCzqmUcjgc1WM2QfBHAesx",
    "Operator" => "Panda",
    "Tenat-Id" => 10001012,
];

$options = [
    'verify' => false,
    'http_errors' => false,
    'headers' => $headers,
];
$response = single_service('curlrequest', $options)->post($url);
```
-----------------------------------------------
the debug  output:

> [DEBUG] [2022-11-12 16:48:12.323709] [server:192.168.8.119] [client:192.168.0.123] -> CodeIgniter\HTTP\CURLRequest::sendRequest -> 
{
    "-1": true,
    "10002": "http://127.0.0.1:8890/api/v1/tenants/",
    "10015": "receivers=tenant%40gmail.com",
    "10023":
    [
        "Accept: application/json",
        "Accept-Language: zh-Hans",
        "Authorization: Bearer 6k8KGSDIGEz8y8QEftRIcCzqmUcjgc1WM2QfBHAesx",
        "Operator: Panda",
        **"Tenat-Id: "**
    ],
    "10036": "POST",
    "155": 5000,
    "156": 150000,
    "19913": true,
    "42": true,
    "45": false,
    "64": false,
    "74": true
}

### Expected Output

the second parameter $value can set to int type in the CodeIgniter\HTTP\Header constructor.

kenjis:
Thank you for reporting.

> the second parameter $value can set to int type in the CodeIgniter\HTTP\Header constructor.

Why? HTTP header values are all string.

kenjis:
This is off topic, but v4.1.1 has serious security bugs.
We recommend upgrade.
[...]

PR #6862, opened 2022-11-14
fix: cast to string all values except arrays in Header class

michalsn:
**Description**
Fixes: #6842

I think that casting will be a better option than throwing an exception in this case.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/Header.php
+++ b/system/HTTP/Header.php
@@ -84,7 +84,7 @@ class Header
      */
     public function setValue($value = null)
     {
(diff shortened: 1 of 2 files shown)


samsonasik (approved):

MGatner (approved):
This is potentially a breaking change. Though our docblock types do not allow objects so I'm okay proceeding.

kenjis (approved):

PR #6863, opened 2022-11-15
fix: remove CI_DEBUG check in Model

kenjis:
**Description**
It is not good if the behavior changes only in the production environment.
It is impossible to write tests for the production environment (CI_DEBUG is false).

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


MGatner (approved):
Yes, very much agree.

michalsn (approved):

PR #6873, opened 2022-11-16
test: add `@var` in ExampleDatabaseTest

kenjis:
**Description**
To prevent PHPStan from reporting errors.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/admin/starter/tests/database/ExampleDatabaseTest.php
+++ b/admin/starter/tests/database/ExampleDatabaseTest.php
@@ -31,6 +31,7 @@ final class ExampleDatabaseTest extends CIUnitTestCase
         $this->setPrivateProperty($model, 'useSoftDeletes', true);
         $this->setPrivateProperty($model, 'tempUseSoftDeletes', true);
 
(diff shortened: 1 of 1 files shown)


samsonasik (approved):

admin/starter/tests/database/ExampleDatabaseTest.php
```diff
@@ -31,6 +31,7 @@ public function testSoftDeleteLeavesRow()
         $this->setPrivateProperty($model, 'useSoftDeletes', true);
         $this->setPrivateProperty($model, 'tempUseSoftDeletes', true);
 
+        /** @var stdClass $object */
```
samsonasik, line 34: `stdClass` should be imported or use `\stdClass` instead.

kenjis: The current sample test files has no namespace.

It seems better to have a namespace and can be autoloaded.

PR #6877, opened 2022-11-17
[Entity] Properly identify boolean when its a string

colethorsen:
**Description**
When passing a string to a value cast as boolean it would always set as true. In some cases, such as a value of false from the post array identify as strings and would be cast as true regardless.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Entity/Cast/BooleanCast.php
+++ b/system/Entity/Cast/BooleanCast.php
@@ -23,4 +23,12 @@ class BooleanCast extends BaseCast
     {
         return (bool) $value;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function set($value, array $params = []): bool
+    {
+        return filter_var($value, FILTER_VALIDATE_BOOLEAN);
+    }
 }
--- a/tests/system/Entity/EntityTest.php
+++ b/tests/system/Entity/EntityTest.php
@@ -348,6 +348,11 @@ final class EntityTest extends CIUnitTestCase
 
         $this->assertIsBool($entity->fifth);
         $this->assertFalse($entity->fifth);
+
+        $entity->fifth = 'false';
+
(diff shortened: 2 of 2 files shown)


kenjis:
> Returns true for "1", "true", "on" and "yes". Returns false otherwise. 
https://www.php.net/manual/en/filter.filters.validate.php

kenjis:
This is strictly BC, but it seems unlikely that this change will break existing apps?

michalsn (changes requested):
Make sense to me. Please add info to the changelog.

system/Entity/Cast/BooleanCast.php
```diff
@@ -23,4 +23,12 @@ public static function get($value, array $params = []): bool
     {
         return (bool) $value;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function set($value, array $params = []): bool
+    {
+        return filter_var($value, FILTER_VALIDATE_BOOLEAN);
```
najdanovicivan, line 32: Can you add FILTER_NULL_ON_FAILURE here as well. I've made similar PR while ago but it was not merged in order to keep casting work similarly as other frameworks do it.

https://github.com/codeigniter4/CodeIgniter4/pull/4591

najdanovicivan (changes requested):

kenjis:
See #7483 and #4591

system/Entity/Cast/BooleanCast.php
kenjis: @najdanovicivan Can you show how other frameworks do it?

lonnieezell:
Last comment was 6 months ago and I think we have other things about to be merged that provide a solution for this. Closing.

PR #6880, opened 2022-11-18
refactor: replace `time()` with Time into `Cache`, `Database` and `Test` class.

ping-yee:
**Description**
See #6815 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -165,6 +165,7 @@ parameters:
     Database:
       - Entity
       - Events
+      - I18n
     Email:
       - I18n
       - Events
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -41,7 +41,7 @@ parameters:
 			path: system/Cache/CacheFactory.php
 
 		-
-			message: "#^Comparison operation \"\\>\" between int\\<1, max\\> and \\(array\\|float\\|int\\) results in an error\\.$#"
+			message: "#^Comparison operation \"\\>\" between int and \\(array\\|float\\|int\\) results in an error\\.$#"
 			count: 1
 			path: system/Cache/Handlers/FileHandler.php
 
--- a/system/Cache/Handlers/FileHandler.php
+++ b/system/Cache/Handlers/FileHandler.php
@@ -242,7 +242,7 @@ class FileHandler extends BaseHandler
             return false;
         }
 
-        if ($data['ttl'] > 0 && time() > $data['time'] + $data['ttl']) {
+        if ($data['ttl'] > 0 && Time::now()->getTimestamp() > $data['time'] + $data['ttl']) {
             // If the file is still there then try to remove it
             if (is_file($this->path . $filename)) {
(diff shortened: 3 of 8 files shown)


kenjis:
Please remove *Ignored error pattern* in `phpstan-baseline.neon.dist`.
```
Run vendor/bin/phpstan analyse
Note: Using configuration file /home/runner/work/CodeIgniter4/CodeIgniter4/phpstan.neon.dist.
   0/434 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0%
 220/434 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░░░░░░░░░░]  50%
 434/434 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

Error: Ignored error pattern #^Comparison operation "\>" between int\<1, max\> and \(array\|float\|int\) results in an error\.$# in path /home/runner/work/CodeIgniter4/CodeIgniter4/system/Cache/Handlers/FileHandler.php was not matched in reported errors.
Error: Comparison operation ">" between int and (array|float|int) results in an error.
  Line   system/Cache/Handlers/FileHandler.php                                  
         Ignored error pattern #^Comparison operation ">" between int<1,        
         max> and \(array\|float\|int\) results in an error\.$# in path         
         /home/runner/work/CodeIgniter4/CodeIgniter4/system/Cache/Handlers/Fil  
         eHandler.php was not matched in reported errors.                       
  245    Comparison operation ">" between int and (array|float|int) results in  
         an error.                                                              
```

ping-yee:
I personally agree that the class provided by CI4 are also used in the underlying development of CI4, so that it will be more convenient for expansion and maintenance in the future.
What's more, `time::now` is a static method. Although it is not as simple and fast as the built-in method, it will not use too many resources.

kenjis:
The main purpose for this change is improving testability.
We cannot change the `time()` return value, so we cannot write proper test code.
Currently we have some fragile test codes that depend on real time.
These rarely fail due to test execution time.

> Creating a standard way of accessing the clock would allow interoperability during testing, when testing behavior that has timing-based side effects. Common ways to get the current time include calling \time() or new \DateTimeImmutable('now'). However, this makes mocking the current time impossible in some situations.
https://www.php-fig.org/psr/psr-20/

kenjis:
> CodeIgniter4's time class is based on PHP's built-in DateTimeZone, which provides many convenient methods. If we could also use CodeIgniter4's own libraries in CodeIgniter4's core classes, we might be able to achieve greater consistency and maintainability in CodeIgniter4 source code.

Yes, we should use `Time` in all the framework code.

kenjis (approved):

kenjis:
I'm going to merge this for consistency. If there is performance issues, we should take care of all the places.

kenjis:
Thank you! @ping-yee

PR #6881, opened 2022-11-18
Change welcome_message.php view - Docs referrer

mostafakhudair:
Github's user guide may contain futures that are under development, and have not been added in the current version, so it should be replaced with it's user guide version.


kenjis:
https://codeigniter4.github.io/userguide/ is the official version on GitHub pages.
It is the same as https://www.codeigniter.com/user_guide/.

The develop version is https://codeigniter4.github.io/CodeIgniter4/.

PR #6883, opened 2022-11-18
feat: disallow `Model::update()` without WHERE clause

kenjis:
**Description**
It is not necessary that the `update()` method be able to update all records without WHERE clause. Updating all records is a risky operation and should not be mixed with update method with id(s). 

The current API is not good design because it is easily to create a vulnerability that allows all records to be updated if you accidentally forget to validate the `$id`. See https://forum.codeigniter.com/showthread.php?tid=84833

- [BC] disallow `Model::update()` without WHERE clause (throws Exception).
- disallow `$id` as bool for `update()`/`delete()` (throws Exception).
- ~~add `Model::updateAll($data)` for UPDATE without WHERE clause.~~
- add tests for `Model::detete()`. DELETE without WHERE clause is already not allowed.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -899,6 +899,10 @@ abstract class BaseModel
      */
     public function update($id = null, $data = null): bool
     {
+        if (is_bool($id)) {
+            throw new InvalidArgumentException('$id should not be boolean.');
+        }
+
         if (is_numeric($id) || is_string($id)) {
             $id = [$id];
         }
@@ -1037,6 +1041,10 @@ abstract class BaseModel
      */
     public function delete($id = null, bool $purge = false)
     {
+        if (is_bool($id)) {
+            throw new InvalidArgumentException('$id should not be boolean.');
+        }
+
         if ($id && (is_numeric($id) || is_string($id))) {
             $id = [$id];
         }
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -910,12 +910,12 @@ abstract class BaseModel
             return false;
         }
 
-        // Must be called first so we don't
+        // Must be called first, so we don't
         // strip out updated_at values.
         $data = $this->doProtectFields($data);
 
         // doProtectFields() can further remove elements from
-        // $data so we need to check for empty dataset again
+        // $data, so we need to check for empty dataset again
         if (empty($data)) {
             throw DataException::forEmptyDataset('update');
         }
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -900,7 +900,7 @@ abstract class BaseModel
     public function update($id = null, $data = null): bool
     {
         if (is_bool($id)) {
-            throw new InvalidArgumentException('$id should not be boolean.');
+            throw new InvalidArgumentException('update(): argument #1 ($id) should not be boolean.');
         }
 
         if (is_numeric($id) || is_string($id)) {
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1042,7 +1042,7 @@ abstract class BaseModel
     public function delete($id = null, bool $purge = false)
     {
         if (is_bool($id)) {
-            throw new InvalidArgumentException('$id should not be boolean.');
+            throw new InvalidArgumentException('delete(): argument #1 ($id) should not be boolean.');
         }
 
         if ($id && (is_numeric($id) || is_string($id))) {
--- a/tests/system/Models/UpdateModelTest.php
+++ b/tests/system/Models/UpdateModelTest.php
@@ -409,7 +409,7 @@ final class UpdateModelTest extends LiveModelTestCase
             [
                 false,
                 InvalidArgumentException::class,
-                '$id should not be boolean.',
+                'update(): argument #1 ($id) should not be boolean.',
             ],
         ];
     }
--- a/tests/system/Models/UpdateModelTest.php
+++ b/tests/system/Models/UpdateModelTest.php
@@ -11,8 +11,10 @@
 
 namespace CodeIgniter\Models;
 
+use CodeIgniter\Database\Exceptions\DatabaseException;
 use CodeIgniter\Database\Exceptions\DataException;
 use CodeIgniter\Entity\Entity;
+use Generator;
 use stdClass;
 use Tests\Support\Models\EventModel;
 use Tests\Support\Models\JobModel;
@@ -377,5 +379,31 @@ final class UpdateModelTest extends LiveModelTestCase
             'country' => '4',
             'email'   => '1+1',
         ]);
+    }
+
+    /**
+     * @dataProvider provideInvalidIds
+     *
+     * @param false|null $id
+     */
+    public function testUpdateThrowDatabaseExceptionWithoutWhereClause($id): void
+    {
+        $this->expectException(DatabaseException::class);
+        $this->expectExceptionMessage(
+            'Updates are not allowed unless they contain a "where" or "like" clause.'
+        );
+
+        // $useSoftDeletes = false
+        $this->createModel(JobModel::class);
+
+        $this->model->update($id, ['name' => 'Foo Bar']);
+    }
+
+    public function provideInvalidIds(): Generator
+    {
+        yield from [
+            [null],
+            [false],
+        ];
     }
 }
--- a/tests/system/Models/UpdateModelTest.php
+++ b/tests/system/Models/UpdateModelTest.php
@@ -15,6 +15,7 @@ use CodeIgniter\Database\Exceptions\DatabaseException;
 use CodeIgniter\Database\Exceptions\DataException;
 use CodeIgniter\Entity\Entity;
 use Generator;
+use InvalidArgumentException;
 use stdClass;
 use Tests\Support\Models\EventModel;
 use Tests\Support\Models\JobModel;
@@ -386,12 +387,10 @@ final class UpdateModelTest extends LiveModelTestCase
      *
      * @param false|null $id
      */
-    public function testUpdateThrowDatabaseExceptionWithoutWhereClause($id): void
+    public function testUpdateThrowDatabaseExceptionWithoutWhereClause($id, string $exception, string $exceptionMessage): void
     {
-        $this->expectException(DatabaseException::class);
-        $this->expectExceptionMessage(
-            'Updates are not allowed unless they contain a "where" or "like" clause.'
-        );
+        $this->expectException($exception);
+        $this->expectExceptionMessage($exceptionMessage);
 
         // $useSoftDeletes = false
         $this->createModel(JobModel::class);
@@ -402,8 +401,16 @@ final class UpdateModelTest extends LiveModelTestCase
     public function provideInvalidIds(): Generator
     {
         yield from [
-            [null],
-            [false],
+            [
+                null,
+                DatabaseException::class,
+                'Updates are not allowed unless they contain a "where" or "like" clause.',
+            ],
+            [
+                false,
+                InvalidArgumentException::class,
+                '$id should not be boolean.',
+            ],
         ];
     }
 }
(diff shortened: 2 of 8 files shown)


tests/system/Models/UpdateModelTest.php
```diff
@@ -378,4 +381,52 @@ public function testUpdateWithSetAndEscape(): void
...
+        // Updates all records.
+        $result = $this->model->updateAll([
+            'name'        => 'name',
+            'description' => 'description',
+        ]);
+
+        $this->assertTrue($result);
+
+        $jobs = $this->model->where(['name' => 'name'])->findAll();
+        $this->assertSame($this->model->countAllResults(), count($jobs));
```
ddevsr, line 430:
```suggestion
        $this->assertCount($this->model->countAllResults(), $jobs);
```

michalsn:
I would prefer to have something less invasive. `updateAll()` method is probably used pretty commonly, and I don't want to push developers to rename methods in their code.

Maybe something like `allowEmptyWhere()` method, which sets the proper flag inside the model and resets it after the query is executed?

That could serve both methods: `update()` and `delete()` - when `useSoftDeletes` option is enabled.

```php
$model->set($data)->allowEmptyWhere()->update();
$model->allowEmptyWhere()->delete();
```

kenjis:
> I would prefer to have something less invasive. updateAll() method is probably used pretty commonly, and I don't want to push developers to rename methods in their code.

What do you mean?
Even if we add `allowEmptyWhere()`, developers need to add the method when they update all records.
After all they need to change.

I would like to deprecate the usage `update(null, $data)` if possible.
You can write `updateAll($data)`, and it never happens that null is passed accidentally.

michalsn:
> What do you mean?
Even if we add allowEmptyWhere(), developers need to add the method when they update all records.
After all they need to change.

Yes, that's true. But there is a big difference, at least from my point of view.

We all have our habits. If I already have `updateAll()` method implemented in my code, I will get used to using it on multiple occasions in many models. The proposed change will make this method "restricted" to use, and I will have to change a code in my project (hopefully only one).

If instead of this, we will use `allowEmptyWhere()` for allowing updates with `null` (which we both think are very rare), then the amount of code that has to be changed is minimal.

Plus, as I mentioned, this also can be used with soft deletes. I don't think there is any straightforward way right now to soft delete all the records.

> I would like to deprecate the usage update(null, $data) if possible.
You can write updateAll($data), and it never happens that null as id is passed accidentally.

I get your point, but again - with `allowEmptyWhere()`, the API could stay the same, and still, we would not have to worry about accidents.

I'm not pushing to do it my way. I got the idea after reading the code from this PR.

MGatner (commented):
This feels very heavy-handed. I agree this will probably prevent more mistakes than actual existing use cases, but it could really break current projects. 

Alternatives: we could add it to our Features Config and encourage people to turn it on in the release notes / upgrade guides. We could trigger & log deprecations or warnings with the intent to make the hard change in a future version (4.4? 5.0?). Other ideas?

sclubricants:
I don't quite follow what the problem is. Can you give a code example of a problematic situation?

michalsn:
You can follow the thread on the forum: https://forum.codeigniter.com/showthread.php?tid=84833 - Kenjis gives some examples there.

sclubricants:
I'm just looking into this problem now and still don't yet get everything that is going on here but I can say that using the new updateBatch() method would probably make more sense for model. It works in a completely different manner and would probably not have issues like this associated with  it. updateBatch() performs an update based on each row of data and not based on a WHERE.

I'll look some more at this but those are my initial thoughts.

kenjis:
@sclubricants `update(null, $data)` executes UPDATE without WHERE clause (if you don't set WHERE at all).
I don't think it is a good API because if a dev forgets to check the first parameter is not null, all records will be updated.

sclubricants:
If when calling update() on model ..if model then calls $builder->onContraint(id)->updateBatch($dat) then it would throw an error if id is not present. If id is present but null then the data would match no rows and the rows with null ids wouldn't update anything.

Using model assumes there is a constraint or pk. So mapping update to updateBatch() should be feasible.

If your updating records in any other manner you really shouldn't use model. Any models without an id/pk should only be able to insert.

sclubricants:
the method says it only updates a single record. Unfortunately it does not always deliver that promise.

As an alternative: 

```php
    protected function doUpdate($id = false, $data = null): bool
    {
        $escape       = $this->escape;
        $this->escape = [];

        $builder = $this->builder();

        if ($id !== true) {
            $builder = $builder->whereIn($this->table . '.' . $this->primaryKey, $id ?? '');
        }

        // Must use the set() method to ensure to set the correct escape flag
        foreach ($data as $key => $val) {
        }

        return $builder->update();
    }
```

sclubricants:
I'd rather see something like this:

```php
    public function update($data = null, $alternativeConstraint = null): bool
    {

    }

    protected function doUpdate($data = null, $alternativeConstraint = null): bool
    {
        $escape       = $this->escape;
        $this->escape = [];

        $builder = $this->builder()->setData($data, $escape)->onConstraint($alternativeConstraint ?? $this->primaryKey);

        return $builder->updateBatch();
    }
```

sclubricants:
Maybe for 4.4 I can go though model and refactor some things to be more inline with builder.

kenjis:
> The method says it only updates a single record. Unfortunately it does not always deliver that promise.

Yes, the comment say so:
https://github.com/codeigniter4/CodeIgniter4/blob/4.3/system/BaseModel.php#L892-L893
But it is not true, because `$id` accepts an array:
https://codeigniter4.github.io/CodeIgniter4/models/model.html#update

kenjis:
> This feels very heavy-handed. I agree this will probably prevent more mistakes than actual existing use cases, but it could really break current projects.

I proposed this break change because I think it is a rare use case where all records are updated. 
If there is actually code that accepts both specific ids and null, then the app will break. 
But I can't imagine such code.

Adding Feature Config is fine, but I prefer the default is to disallow updates all.

sclubricants:
I don't see why we need updateAll(). It seems you should use builder for thos types of operations. 

For that matter make model method that returns an instance of builder with table set.

$model->builder()->update()

kenjis:
`update($id, $data)` for updating a specific id or specific ids.
`updateAll($data)` for updating without ids.

`$model->builder()->set()->update()` works, but `$model->set()->builder()->update()` does not work.
The model is difficult to understand. When using a model, it seems better to be able to handle it with the model's methods.

However, if many people say that `updateAll()` is not necessary, I am fine without it.

MGatner:
Let's make sure @lonnieezell @paulbalandan and @samsonasik have seen this and had a chance to weigh in.

system/BaseModel.php
```diff
@@ -899,6 +899,10 @@ public function insertBatch(?array $set = null, ?bool $escape = null, int $batch
      */
     public function update($id = null, $data = null): bool
     {
+        if (is_bool($id)) {
+            throw new InvalidArgumentException('$id should not be boolean.');
```
samsonasik, line 903: probably more verbose message: `argument #1: $id should not be boolean`, so it immediatelly know which part that invalid.

samsonasik: Ref for better verbose message https://3v4l.org/8JSTu

kenjis:
Rebased.

system/BaseModel.php
kenjis: Updated the exception message.

samsonasik (approved):

system/BaseModel.php
```diff
@@ -1037,6 +1041,10 @@ public function updateBatch(?array $set = null, ?string $index = null, int $batc
      */
     public function delete($id = null, bool $purge = false)
     {
+        if (is_bool($id)) {
+            throw new InvalidArgumentException('$id should not be boolean.');
```
paulbalandan, line 1045: This should likewise be changed.

system/BaseModel.php
```diff
@@ -899,6 +899,10 @@ public function insertBatch(?array $set = null, ?bool $escape = null, int $batch
      */
     public function update($id = null, $data = null): bool
     {
+        if (is_bool($id)) {
```
paulbalandan, line 902: In what instances would this be a bool?

kenjis: When an attacker passes a bool and the dev forgets to validate the id value.

system/BaseModel.php
kenjis: Changed.

michalsn:
> * add Model::updateAll($data) for UPDATE without WHERE clause.

This also should be treated as a BC change - a new reserved method name.

kenjis:
At least, @sclubricants and @iRedds say that `updateAll()` is not necessary.

@michalsn If it is a breaking change, adding any new method will be a breaking change?

michalsn:
> At least, @sclubricants and @iRedds say that updateAll() is not necessary.

I also think having a separate method for this is not the best idea. However, I would like to keep the ability to update all records.

> @michalsn If it is a breaking change, adding any new method will be a breaking change?

IMO, yes. But there is still the probability of whether the method name had a chance to be used by the developer. In this case, that probability is high because it's a rather generic and natural name.

sclubricants:
> @codeigniter4/database-team Please comment.
> 

Im traveling today. Ill take a look tomorrow.

kenjis:
The name `updateAll()` is not the best, because it may not update all the records.
It can be used like `where(...)->updateAll($data)`.
So `updateMany()`, `updateMultiple()`, `updateWithoutIds()`, etc. are also possible.

MGatner:
What's the advantage to having this feature at all? If we force people to use Query Builder (`$model->builder()->where(...)->update()`) then these mistakes won't happen. I guess Model will account for soft deletes and event triggers, but I don't know that those are relevant to non-identity updates anyways.

michalsn:
I think we're in some kind of a deadlock here. If we all can agree that the risk of assigning a `null` or empty `array` is too dangerous here, then we probably should resign with `updateAll()`-like methods and just always throw an exception when `$builder->getCompiledQBWhere() === []` and that's it.

Cases, when we want to update everything, are so rare that developers may prepare their method for it or just use a hack, like `where('1 = 1', null, true)` in the code.

sclubricants:
I think that any DML operations in model should be performed on the basis of the primary key. If you want to do other operations like updating all models then you should just use the builder methods. In the future Id like to see us work toward the ability of composite primary keys. The new *Batch() methods will work well with composite keys.

 Why are we worried about having the ability to update all records in model when this can easily be accomplished in builder? For me this type of operation just does not belong in model.

kenjis:
Rebased.

Okay, I will remove `updateAll()` in this PR.

> In the future Id like to see us work toward the ability of composite primary keys. The new *Batch() methods will work well with composite keys.

Yes. I think it is the way to go.

kenjis:
Removed updateAll() and added docs.

kenjis:
> I think that any DML operations in model should be performed on the basis of the primary key. 

That may indeed be true.

michalsn (approved):

ddevsr (approved):

kenjis:
```
Run vendor/bin/psalm
...
ERROR: UndefinedClass - system/Config/Services.php:338:16 - Type null cannot be called as a class (see https://psalm.dev/019)
        return new $class($config);


ERROR: UndefinedClass - system/Config/Services.php:659:19 - Type null cannot be called as a class (see https://psalm.dev/019)
        $driver = new $driverName($config, AppServices::request()->getIPAddress());
```
https://github.com/codeigniter4/CodeIgniter4/actions/runs/3655630049/jobs/6177588634

MGatner:
Ah the old static analysis standoff! 😂

Yes I'm thinking in the same direction as @sclubricants... The intent of Model is to work with identifiable entities (the concept, not the type). I know we aren't officially a repository pattern, but I think the concept stands - and anonymously updating persisted sets of data is not conceptually sound.

Issue #6882, opened 2022-11-18
Bug: The unit test of `tests/system/Cache/Handlers/FileHandlerTest` cannot pass.


ping-yee:
### PHP Version

8.1

### CodeIgniter4 Version

4.2.10

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

cli-server (PHP built-in webserver)

### What happened?

```
$ ./vendor/bin/phpunit tests/system/Cache/Handlers/FileHandlerTest.php 
PHPUnit 9.5.25 #StandWithUkraine

Runtime:       PHP 8.1.12
Configuration: /mnt/d/CodeIgniter4/phpunit.xml.dist
Warning:       No code coverage driver available

.F...F..........FFF....                                                                                      23 / 23 (100%)

Nexus\PHPUnit\Extension\Tachycardia identified these 2 slow tests:
⚠  Took 0.72s from 0.50s limit to run CodeIgniter\\Cache\\Handlers\\FileHandlerTest::testDeleteMatchingSuffix
⚠  Took 0.71s from 0.50s limit to run CodeIgniter\\Cache\\Handlers\\FileHandlerTest::testDeleteMatchingPrefix


Time: 00:09.449, Memory: 10.00 MB

There were 5 failures:

1) CodeIgniter\Cache\Handlers\FileHandlerTest::testNewWithNonWritablePath
Failed asserting that exception of type "CodeIgniter\Cache\Exceptions\CacheException" is thrown.

2) CodeIgniter\Cache\Handlers\FileHandlerTest::testSave
Failed asserting that true is false.

/mnt/d/CodeIgniter4/tests/system/Cache/Handlers/FileHandlerTest.php:140

3) CodeIgniter\Cache\Handlers\FileHandlerTest::testSaveMode with data set #0 (416, '640')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'640'
+'777'

/mnt/d/CodeIgniter4/tests/system/Cache/Handlers/FileHandlerTest.php:285

4) CodeIgniter\Cache\Handlers\FileHandlerTest::testSaveMode with data set #1 (384, '600')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'600'
+'777'

/mnt/d/CodeIgniter4/tests/system/Cache/Handlers/FileHandlerTest.php:285

[...]

PR #6884, opened 2022-11-19
Add `is_windows()` global function

paulbalandan:
**Description**
- Adds `is_windows()` global function, for uniform detection of Windows platform
- Deprecate `CLI::isWindows()`, as platform detection is not limited to console
- Normalize platform detection

Related: #6882 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -725,8 +725,18 @@ if (! function_exists('is_windows')) {
     /**
      * Detect if platform is running in Windows.
      */
-    function is_windows(): bool
+    function is_windows(?bool $mock = null): bool
     {
+        static $mocked;
+
+        if (func_num_args() === 1) {
+            $mocked = $mock;
+        }
+
+        if (isset($mocked)) {
+            return $mocked;
+        }
+
         return DIRECTORY_SEPARATOR === '\\';
     }
 }
--- a/system/Common.php
+++ b/system/Common.php
@@ -721,6 +721,16 @@ if (! function_exists('is_really_writable')) {
     }
 }
 
+if (! function_exists('is_windows')) {
+    /**
+     * Detect if platform is running in Windows.
+     */
+    function is_windows(): bool
+    {
+        return DIRECTORY_SEPARATOR === '\\';
+    }
+}
+
 if (! function_exists('lang')) {
     /**
      * A convenience method to translate a string or array of them and format
--- a/system/Common.php
+++ b/system/Common.php
@@ -691,7 +691,7 @@ if (! function_exists('is_really_writable')) {
     function is_really_writable(string $file): bool
     {
         // If we're on a Unix server we call is_writable
-        if (DIRECTORY_SEPARATOR === '/') {
+        if (! is_windows()) {
             return is_writable($file);
         }
 
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -518,10 +518,12 @@ class CLI
 
     /**
      * if operating system === windows
+     *
+     * @deprecated v4.3 Use `is_windows()` instead
      */
     public static function isWindows(): bool
     {
-        return PHP_OS_FAMILY === 'Windows';
+        return is_windows();
     }
 
     /**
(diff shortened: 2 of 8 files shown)


MGatner (changes requested):
I read conflicting options on whether `uname()` or `if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN')` is better. If you feel confident this is the way then I trust you. 

We should find a way to make this mockable so people can test their Windows-specific code without being on Windows. My initial thought is to add a parameter to set/clear a local static value:
```php
public function testWindowsUsesBananas() {
    is_windows(true);

    $this->assertSame('bananas', $this->execute('getFruitByOs');

    is_windows(null);
}
```
It's a little bit hacky but thus is functional programming 🤷‍♂️ Only other alternative might be to stick the value on CodeIgniter along with $context.

paulbalandan:
On WSL, PHP_OS is Windows so the check would fail on that. DIRECTORY_SEPARATOR is also "/" on WSL.

kenjis:
De we really need to detect WSL as Windows?

It seems WSL supports Chmod/Chown.
https://devblogs.microsoft.com/commandline/chmod-chown-wsl-improvements/

paulbalandan:
> It seems WSL supports Chmod/Chown. https://devblogs.microsoft.com/commandline/chmod-chown-wsl-improvements/

It seems the article is referring to the native filesystem functions, not the PHP equivalents. As tested in WSL, PHP's `chmod` behaves as if running on native Windows.

kenjis:
Do you mount your drive with `-o metadata`?
See https://github.com/Microsoft/WSL/issues/3172

In my understanding, PHP's chmod() is kind of wrapper for the native filesystem functions.

paulbalandan:
Thanks for the link!

I was not able to make it work on the first try. I first need to uninstall all php related from WSL, then proceeding with unmounting and remounting:
```console
sudo apt remove -y php8.2-*
cd /
sudo umount /mnt/c
sudo mount -t drvfs C: /mnt/c -o metadata,uid=1000,gid=1000
```

Now, `chmod` works on WSL:
```console
$ vendor/bin/phpunit --filter GeneratorsTest --no-coverage
PHPUnit 9.5.26 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.0RC5
Configuration: /mnt/c/Users/P/Desktop/Web Dev/CodeIgniter4/phpunit.xml.dist

........                                                                                                                              8 / 8 (100%)

Nexus\PHPUnit\Extension\Tachycardia identified these 4 slow tests:
⚠  Took 5.50s from 0.50s limit to run CodeIgniter\\Commands\\GeneratorsTest::testGenerateFileCreated
⚠  Took 1.52s from 0.50s limit to run CodeIgniter\\Commands\\GeneratorsTest::testSuffixingHasNoEffect
⚠  Took 1.36s from 0.50s limit to run CodeIgniter\\AutoReview\\FrameworkCodeTest::testEachTestClassHasCorrectGroupAnnotation with data set \"CodeIgniter\\Commands\\GeneratorsTest\"
⚠  Took 0.54s from 0.50s limit to run CodeIgniter\\Commands\\GeneratorsTest::testGenerateFileFailsOnUnwritableDirectory


Time: 00:10.806, Memory: 44.00 MB

OK (8 tests, 14 assertions)

```

paulbalandan:
I'll update the PR to limit the function on native Windows.

kenjis:
Please update the docs. Otherwise looks good.

kenjis (approved):

system/Common.php
```diff
@@ -721,6 +721,22 @@ function is_really_writable(string $file): bool
...
+    /**
+     * Detect if platform is running in Windows.
+     *
+     * @param bool|null $mock If given and a boolean then it will be the return value.
+     */
+    function is_windows(?bool $mock = null): bool
+    {
+        if (is_bool($mock)) {
+            return $mock;
+        }
```
MGatner, line 734: I don't think this will work for mocking. It would require every call to this function to pass the correct parameter. My proposal was more like this:
```suggestion
     * @param bool|null $mock Sets /clears the mock value to return for testing.
     */
    function is_windows(?bool $mock = null): bool
    {
        static $mocked;

        if (func_num_args()) {
            $mocked = $mock;
        }

        if (is_bool($mocked)) {
            return $mocked;
        }
```

MGatner (changes requested):

system/Common.php
kenjis: Yes, you are correct!

kenjis (approved):

MGatner (approved):
Thanks @paulbalandan! Looks good.

juanma386:
Thank you **very much for the improvements**, from Argentina

PR #6885, opened 2022-11-19
Cache phpunit's static analysis coverage cache

paulbalandan:
**Description**
Attempt to fix `file_put_contents` warning when merging cov files.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


.github/workflows/test-phpunit.yml
```diff
@@ -169,6 +169,15 @@ jobs:
         env:
           COVERAGE_NAME: php-v${{ env.COVERAGE_PHP_VERSION }}-${{ matrix.db-platforms }}
 
+      - name: Cache PHPUnit coverage cache
+        if: matrix.php-versions == env.COVERAGE_PHP_VERSION
+        uses: actions/cache@v3
+        with:
+          path: build/.phpunit.cache/code-coverage
+          key: phpunit-code-coverage-${{ hashFiles('**/phpunit.*') }}
```
MGatner, line 177: Is the intent to hit the Config files and workflow definition itself? This seems like it could hit a lot more, like some of the output files themselves.

paulbalandan: Yeah, you're right! The intent was for the output but hashing them didn't cross my mind. 😅

paulbalandan:
I changed the implementation to just remove the "codeCoverageDirectory" entry in phpunit.xml.dist

kenjis (dismissed):

phpunit.xml.dist
```diff
@@ -12,8 +12,7 @@
          failOnWarning="true"
          verbose="true">
 
-    <coverage cacheDirectory="build/.phpunit.cache/code-coverage"
```
kenjis, line 15: This file is the PHPUnit configuration we use, so it is better it has cache directory by default.

.github/workflows/test-phpunit.yml
paulbalandan: I decided to go with the first implementation because it _just_ works.

- Hashing the output files as filename will not work since those output files are not yet available here, so resulting in an empty string
- Uploading the results won't work because of the way download-artifact unzips the download

MGatner: If it works, I'm good with it until it doesn't 🤷‍♂️

MGatner (approved):

samsonasik (approved):

samsonasik:
Thank you @paulbalandan

PR #6886, opened 2022-11-20
docs: update DB transactions

kenjis:
~~Needs #6919~~

**Description**
- update docs
- ~~fix out of dated description. See https://github.com/bcit-ci/CodeIgniter/issues/4032~~
- add tests

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -66,7 +66,7 @@ Strict Mode can be disabled as follows:
 
 .. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
     When ``DBDebug`` is true, if a query error occurs, all the queries
-    will be rolled backed, and an exception will be thrown.
+    will be rolled backed, and a ``DatabaseException`` will be thrown.
     In previous versions, ``DBDebug`` was false only in production environment,
     and different database drivers might throw different exception classes.
 
@@ -85,7 +85,7 @@ If the ``DBDebug`` is false, you can manage your own errors like this:
 
 .. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
     When ``DBDebug`` is true, if a query error occurs, all the queries
-    will be rolled backed, and an exception will be thrown.
+    will be rolled backed, and a ``DatabaseException`` will be thrown.
     In previous versions, ``DBDebug`` was false only in production environment,
     and different database drivers might throw different exception classes.
 
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -46,7 +46,7 @@ methods and they will all be committed or rolled back based on the success
 or failure of any given query.
 
 .. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
-    When ``DBDebug`` is true, if an query error occurs, all the queries
+    When ``DBDebug`` is true, if a query error occurs, all the queries
     will be rolled backed, and an exception will be thrown.
     In previous versions, ``DBDebug`` was false in production environment,
     and different database drivers might throw different exception classes
@@ -66,7 +66,7 @@ Strict Mode can be disabled as follows:
 .. literalinclude:: transactions/002.php
 
 .. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
-    When ``DBDebug`` is true, if an query error occurs, all the queries
+    When ``DBDebug`` is true, if a query error occurs, all the queries
     will be rolled backed, and an exception will be thrown.
     In previous versions, ``DBDebug`` was false in production environment,
     and different database drivers might throw different exception classes
@@ -84,7 +84,7 @@ If the ``DBDebug`` is false, you can manage your own errors like this:
 .. literalinclude:: transactions/003.php
 
 .. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
-    When ``DBDebug`` is true, if an query error occurs, all the queries
+    When ``DBDebug`` is true, if a query error occurs, all the queries
     will be rolled backed, and an exception will be thrown.
     In previous versions, ``DBDebug`` was false in production environment,
     and different database drivers might throw different exception classes
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -72,6 +72,8 @@ Strict Mode can be disabled as follows:
     and different database drivers might throw different exception classes
     or did not throw exceptions.
 
+.. _transactions-managing-errors:
+
 Managing Errors
 ===============
 
@@ -110,6 +112,8 @@ a valid result. To use test mode simply set the first parameter in the
 ``$this->db->transStart()`` method to true:
 
 .. literalinclude:: transactions/005.php
+
+.. _transactions-manual-transactions:
 
 Running Transactions Manually
 =============================
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -32,6 +32,12 @@ contrast, we've implemented a smart transaction system that does all
 this for you automatically (you can also manage your transactions
 manually if you choose to, but there's really no benefit).
 
+.. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
+    When ``DBDebug`` is true, if a query error occurs, all the queries
+    will be rolled backed, and a ``DatabaseException`` will be thrown.
+    In previous versions, ``DBDebug`` was false only in production environment,
+    and different database drivers might throw different exception classes.
+
 Running Transactions
 ====================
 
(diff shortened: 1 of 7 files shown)


sclubricants:
How about a test with `$this->db->transStrict(false);`  This would change these test outcomes right?

kenjis:
I don't know why Strict Mode is needed.

kenjis:
> How about a test with $this->db->transStrict(false); This would change these test outcomes right?

Added. It works as expected.

But it seems the strict mode makes sense only when DBDebug is false.
Since v4.3.0, DBDebug is true by default, so if one query fails, an exception will be thrown and that's all.

Is the strict mode a safeguard that prevents all subsequent transactions from being executed if one transaction fails when there were no exceptions?

sclubricants:
Adding the transaction status helps to understand the tests better.

>But it seems the strict mode makes sense only when DBDebug is false.

yes, then you can use `$this->db->transStatus()` to determine what to do.

I wonder if calling `$this->db->transCommit();` or `$this->db->transRollback()`  will then allow you to start new transactions again or if they will keep failing.

sclubricants:
There is no resetTrans() method. In case you wanted to try a series of transactions again while in strict mode.. 

It can be accomplished like this however:

```php
    public function resetTransAfterFailure()
    {
        $this->disableDBDebug();

        $builder = $this->db->table('job');

        $this->db->transStrict(true);

        // The first transaction group
        $this->db->transStart();

        $jobData = [
            'name'        => 'Grocery Sales',
            'description' => 'Discount!',
        ];
        $builder->insert($jobData);

        $this->assertTrue($this->db->transStatus());

        // Duplicate entry '1' for key 'PRIMARY'
        $jobData = [
            'id'          => 1,
            'name'        => 'Comedian',
            'description' => 'Theres something in your teeth',
        ];
        $builder->insert($jobData);

        $this->assertFalse($this->db->transStatus());

        $this->db->transComplete();

        $this->dontSeeInDatabase('job', ['name' => 'Grocery Sales']);


        $this->db->transStrict(false); // resets transaction status
        $this->db->transComplete(); // resets transaction status
        $this->db->transStrict(true); // return back to strick mode


        // The second transaction group
        $this->db->transStart();

        $jobData = [
            'name'        => 'Comedian',
            'description' => 'Theres something in your teeth',
        ];
        $builder->insert($jobData);

        $this->assertTrue($this->db->transStatus());

        $this->db->transComplete();

        $this->seeInDatabase('job', ['name' => 'Comedian']);

        $this->enableDBDebug();
    }
```

sclubricants:
>But it seems the strict mode makes sense only when DBDebug is false.

We should show how to turn debug off in the managing errors section. 
https://codeigniter4.github.io/CodeIgniter4/database/transactions.html#managing-errors

sclubricants (changes requested):

sclubricants:
Can you add this method and documentation?

```php
    /**
     * Reset transaction status - to restart transactions after strict mode failure
     */
    public function resetTransStatus(): BaseConnection
    {
        $this->transStatus = true;

        return $this;
    }
```

sclubricants:
We probably need a public method on BaseConnection to set debug mode.

kenjis:
> We probably need a public method on BaseConnection to set debug mode.

Do you need `DBDebug = false`?
Since v4.3.0, `DBDebug` is true by default in all environments,
and I would like to deprecate `DBDebug` in the future.

Since PHP 8.0, PDO throws exceptions by default:
https://www.php.net/manual/en/pdo.error-handling.php

kenjis:
> We should show how to turn debug off in the managing errors section.
https://codeigniter4.github.io/CodeIgniter4/database/transactions.html#managing-errors

I added a comment in the sample code.

user_guide_src/source/database/transactions.rst
```diff
@@ -45,28 +45,49 @@ You can run as many queries as you want between the start/complete
 functions and they will all be committed or rolled back based on the success
 or failure of any given query.
 
+.. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
+    When ``DBDebug`` is true, if an query error occurs, all the queries
```
MGatner, line 49:
```suggestion
    When ``DBDebug`` is true, if a query error occurs, all the queries
```

user_guide_src/source/database/transactions.rst
```diff
@@ -45,28 +45,49 @@ You can run as many queries as you want between the start/complete
...
+one group fails all subsequent groups will be rolled back. If strict mode is
 disabled, each group is treated independently, meaning a failure of one
 group will not affect any others.
 
 Strict Mode can be disabled as follows:
 
 .. literalinclude:: transactions/002.php
 
+.. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
+    When ``DBDebug`` is true, if an query error occurs, all the queries
```
MGatner, line 69:
```suggestion
    When ``DBDebug`` is true, if a query error occurs, all the queries
```

user_guide_src/source/database/transactions.rst
```diff
@@ -45,28 +45,49 @@ You can run as many queries as you want between the start/complete
...
-you'll see a standard error message if the commit was unsuccessful. If
-debugging is turned off, you can manage your own errors like this:
+If you have ``DBDebug`` true in your **app/Config/Database.php** file
+you'll see an exception thrown if the commit was unsuccessful. If
+``DBDebug`` is false, you can manage your own errors like this:
 
 .. literalinclude:: transactions/003.php
 
+.. note:: Since v4.3.0, ``DBDebug`` is true by default in all environments.
+    When ``DBDebug`` is true, if an query error occurs, all the queries
```
MGatner, line 85:
```suggestion
    When ``DBDebug`` is true, if a query error occurs, all the queries
```

MGatner (changes requested):
A few typos, otherwise looks good!

sclubricants:
>I added a comment in the sample code.

I really wish we had a method. ` $this->db->setDBDebug(false);` You can't depend on the database config for handling these. Maybe you have it turned off in one environments and on in another. It kind of defeats the purpose of having `transStatus()` if it always throws an error. 

Perhaps a better way to handle this would be to have a parameter on `transStart()` or a separate method `transError(true/false)`. 

I looked back in my own code where I used transactions and found that I had implemented a workflow using `transStatus()`  but I had a comment by ```$this->db->transComplete(); // throws error?```. 

We need an easy way to set the failure behavior other than the database config file.

kenjis:
@codeigniter4/database-team 

Before v4.3.0, `DBDebug` was `false` in production environment, but `true` in other environments.
It is very bad practice, because many devs don't know the behavior will change only in production environment.
In develop environment, CI throws an exception but not in production environment.

Since v4.3.0, `DBDebug` is always `true` by default, and it has become nothing to do with debugging (showing errors). 
It is just a flag to throw an exception or not. It is error mode.

> It kind of defeats the purpose of having `transStatus()` if it always throws an error.

Yes, now in `4.3`, `transStatus()` does not work at all unless you set `DBDebug = false`.
By default if a query error occurs, it will be automatically rolled back and an exception will be thrown.

`DBDebug` is error mode in 4.3, and there seems to be no need to change it at runtime. 
So I would like to remove the setting if possible. I don't want to add a new method to change it.

> Perhaps a better way to handle this would be to have a parameter on `transStart()` or a separate method `transError(true/false)`.

I feel that such a way is necessary unless we remove manual transactions functionality.
It seems removing manual transactions is difficult.

MGatner:
I was strongly in favor of the DBDebug changes and @kenjis' explanation seems like the next logical steps. I've never actually used our `trans____()` methods so don't have a suggestion on how what the best implementation might be there. But generally it sounds like we need a separate flag for transaction error handling that isn't tied to our regular error handling.

And yes: let's plan to do away with DBDebug altogether. Firing a `DatabaseErrorEvent` or something should be plenty for developers, along with the new exception handling feature.

kenjis:
I created an issue #6909

This PR is only for documentation update and adding test code.
Can I merge this? @codeigniter4/database-team

michalsn (changes requested):
Shouldn't we also update the user guide if the general behavior in 4.3 has changed? If we're about to throw an exception every time, the examples in the user guide seem useless now.

I would like to see Transactions to be mentioned in the changelog. This is a critical change for someone who uses them.

kenjis:
Rebased.

kenjis:
Added about transactions in the changelog.

michalsn (commented):
I think we still need some code examples in the user guide to show how to handle things when exceptions are thrown, since this will now be the default behavior.

kenjis:
I add a sample code and updated some descriptions.

I also found that SQLite3, Postgres and OCI8 drivers throw ErrorException when a query error occurs
because of Error Handler. See #6912
https://github.com/codeigniter4/CodeIgniter4/blob/ac1422c404a364f9493bbf1b22dacb18e3913982/system/Debug/Exceptions.php#L174-L176
So depending on the value of the error reporting setting, an exception may not be thrown.

user_guide_src/source/database/transactions/008.php
```diff
@@ -0,0 +1,15 @@
...
+
+try {
+    $this->db->transStart();
+    $this->db->query('AN SQL QUERY...');
+    $this->db->query('ANOTHER QUERY...');
+    $this->db->query('AND YET ANOTHER QUERY...');
+    $this->db->transComplete();
+} catch (DatabaseException $e) {
+    // Automatically rolled back already.
+}
```
kenjis, line 15: I still feel uncomfortable with this code.

kenjis:
> Perhaps a better way to handle this would be to have a parameter on `transStart()` or a separate method `transError(true/false)`.

See #6917

michalsn:
Well, I wasn't saying anything since everyone seemed to be on board regarding throwing exceptions, but... personally, I don't think that throwing exceptions for anything related to transactions has sense. This is why we have `transStatus()` - to handle errors.

So, for me #6917 is definitely a good way to solve this.

kenjis:
> So, for me https://github.com/codeigniter4/CodeIgniter4/pull/6917 is definitely a good way to solve this.

Yes, I also think it is necessary missing feature for v4.3.

kenjis:
@sclubricants I sent #8767

PR #6897, opened 2022-11-23
fix: required PHP extentions

kenjis:
**Description**
- update the docs
- remove `curl` and `xml` from required PHP extension check in `CodeIgniter`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,6 @@ Made with [contrib.rocks](https://contrib.rocks).
 PHP version 7.4 or higher is required, with the following extensions installed:
 
 - [intl](http://php.net/manual/en/intl.requirements.php)
-- [libcurl](http://php.net/manual/en/curl.requirements.php) if you plan to use the HTTP\CURLRequest library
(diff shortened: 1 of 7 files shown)


cijagani:
@kenjis  i think CURL PHP extention is required.
image

https://www.codeigniter.com/user_guide/libraries/curlrequest.html?highlight=curl#

kenjis:
@cijagani Yes, ext-[curl](https://www.php.net/manual/en/curl.requirements.php) is required if you use `CURLRequest`.

cijagani:
we should keep curl-extenstion as required in doc. cz of if someone use `CURLRequest` and curl extenstion is not installed then it will throw exeptions.

kenjis:
It is listed:
https://github.com/codeigniter4/CodeIgniter4/blob/fc0a497f3939b0a3138d36e82d0fe15ce8ee0cba/user_guide_src/source/intro/requirements.rst

cijagani (approved):

cijagani (approved):

PR #6910, opened 2022-11-25
fix: $requset type in BaseController and BaseResource

kenjis:
**Description**
To suppress PHPStan errors like:
> Call to an undefined method CodeIgniter\HTTP\RequestInterface::getPost()

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


kenjis:
I have a question.

What is the `RequestInterface` in CI4?
Should this eventually have all the public methods of `IncommingRequest`? 
Or should it be the same interface as PSR-7?

samsonasik:
Typo: requset -> request

kenjis:
@samsonasik Fixed.
Feel free to fix typos.

michalsn (commented):
I'm not a fan of the `assert()` part. In `BaseController` it's not a big problem, but in `BaseResource` it can be problematic since I can create my own version without extending the "base" class.

As for `RequestInterface` I believe we wanted it to be compatible with PSR-7. But @MGatner is the person to confirm or deny that.

MGatner:
Yes the intent is to have a fully-compatible PSR-7 set of classes. That said, I have no problem with offering our own framework interfaces for people to swap out. I think ideally these are extensions of the PSR-7 versions and classes are typed to what they actually need (i.e. minimum actual methods required)

MGatner (approved):

kenjis:
@MGatner Will `RequestInterface` have all the `IncommingRequest` public methods?

michalsn:
> The reason I haven't included the assert() in the Controller is because I don't know what the Controller should be.

@kenjis Sorry - what I meant was that I can have my own request class like this:
```
class MyRequest implements RequestInterface
```
and use it in my `services`, and it should still work.

In `BaseController` we can simply change the code if our class not inherit from `CLIRequest` or `IncomingRequest`.

But in the `BaseResource` we will be forced to copy the entire class and use it instead of the original one because `assert()` will cause an error.

kenjis:
Oh, I didn't expect `class MyRequest implements RequestInterface`.
Because RequestInterface does not have many methods in IncommingRequest.
But it is a correct way to use interfaces.

kenjis:
Removed `assert()` because it is not necessary to suppress PHPStan errors.

michalsn (approved):
Tests are failing, but it's unrelated. Thanks!

MGatner:
> @MGatner Will `RequestInterface` have all the `IncommingRequest` public methods?

I hope we won't need to have a `RequestInterface`, but rather can type everything that needs the framework class to the class itself. Other components that rely on HTTP classes should be typed to the PSR-7 interfaces (when possible), or smaller specific interfaces like `LocaleDetection` (just an example).

MGatner (approved):

PR #6870, opened 2022-11-16
docs: make PHPDoc types more specific in Header

kenjis:
**Description**
Follow-up #6862

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [] Unit testing, with >80% coverage
- [] User guide updated
- [x] Conforms to style guide

PR #6914, opened 2022-11-27
docs: fix PHPDoc in Header

kenjis:
**Description**
Follow-up: #6870

- fix PHPDoc in Header

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/Header.php
+++ b/system/HTTP/Header.php
@@ -101,7 +101,7 @@ class Header
      * Appends a value to the list of values for this header. If the
      * header is a single value string, it will be converted to an array.
      *
-     * @param array<array<string>|string>|string|null $value
+     * @param array<string, string>|string|null $value
      *
      * @return $this
      */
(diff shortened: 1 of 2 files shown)


mackxpall (approved):

mackxpall (approved):

mackxpall (approved):

system/HTTP/Header.php
```diff
@@ -28,15 +28,23 @@ class Header
...
+     *   [
+     *       'foo',
+     *       [
+     *           'bar' => 'fizz',
+     *       ],
+     *       'baz' => 'buzz',
+     *   ]
      *
-     * @var array<array<string>|string>|string
+     * @var array<int|string, array<string, string>|string>|string
```
MGatner, line 40: This is indicative to me that we should probably be using a value object or collection class.

MGatner (approved):

system/HTTP/Header.php
kenjis: I don't know why we need to have both `['bar' => 'fizz']` and `'bar' => 'fizz'`.

MGatner: Agreed! v5? 😂

Issue #6909, opened 2022-11-25
Dev: manual DB transaction after 4.3


kenjis:
- Allow manual DB transactions even when `DBDebug` is true.
- `DBDebug` will be deprecated in the future if possible.

See https://github.com/codeigniter4/CodeIgniter4/pull/6886#issuecomment-1325774111

PR #6917, opened 2022-11-27
feat: do not throw exceptions during transactions by default

kenjis:
~~Needs #6919~~
Supersedes #6886

**Description**
- [BC] do not throw exceptions during transactions by default
- add `$db->transException(true)` to throw exceptions
- update docs
- add tests

Ref: #6909

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseConnection.php
+++ b/system/Database/BaseConnection.php
@@ -29,7 +29,7 @@ use Throwable;
  * @property string     $DBDriver
  * @property string     $DBPrefix
  * @property string     $DSN
- * @property mixed      $encrypt
+ * @property array|bool $encrypt
  * @property array      $failover
  * @property string     $hostname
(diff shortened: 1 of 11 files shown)


michalsn (approved):
Handling transactions with exceptions is not easy. Our current approach with `transStatus()` seems to be more natural.

kenjis:
Rebased, and this PR is ready.

@codeigniter4/database-team Please review.

kenjis:
This breaking change is only when `DBDebug` is true.
If a dev uses the default value (`ENVIRONMENT !== 'production'`), this does not change the behavior change in production.

michalsn:
Okay, so the impact is minimal.

PR #6912, opened 2022-11-26
test: add DB transaction tests

kenjis:
**Description**
This is tests for v4.2.

Related #6919, #6886

See https://github.com/codeigniter4/CodeIgniter4/actions/runs/3555886311/jobs/5972871727 to see Exception classes.

The current behaviors:
- All the current DB drivers throw an exception when a query error occurs.
- MySQLi driver sets `mysqli_report(MYSQLI_REPORT_ALL & ~MYSQLI_REPORT_INDEX)` that throws `mysqli_sql_exception`.
- SQLite3, Postgres and OCI8 drivers throw `ErrorException` because of Error Handler. 
- SQLSRV driver has different implementaion. It throws `Exception` when the query return value is false.

**Checklist:**
- [x] Securely signed commits
- [] Component(s) with PHPDoc blocks, only if necessary or adds value
- [] Unit testing, with >80% coverage
- [] User guide updated
- [x] Conforms to style guide

PR #6919, opened 2022-11-27
docs: fix transactions.rst

### LINKED ISSUE #6916: DB Transactions description for Strict Mode
On looking further the issue is bigger than I thought. The documentation has the following description for "strict mode";
"When strict mode is enabled, if you are running multiple groups of transactions, if one group fails all groups will be rolled back."
That is not true as transaction will be committed when level 0 of the transaction is reached.  As the code stands now the a query  failure in or out of a transaction will cause all future transaction to fail. A query failure will have no effect on and transaction that has been committed. 

Query faulure only effect previous queries is the transactions have been nested. Take a look at the following;
$this->db->trans_start();
//first queries
$this->db->trans_complete();
$this->db->trans_start();
//second queries
$this->db->trans_complete();
$this->db->trans_start();
//third queries
$this->db->trans_complete();
If the first queries complete it will be committed at the first trans_complete(). If the second queries do not complete the first transaction can not be rolled back as it has already been committed. The third  queries will roll back only in "strict mode".

Look at a nested query;
$this->db->trans_start();
    //some transaction
   $this->db->trans_start();
    //inside transaction
    $this->db->trans_complete();
    //last queries
$this->db->trans_complete();
In this case the commit only happens on the outside trans_complete(). Therefore all previous queries, after the first trans_start(), are rolled back. The trans_strict flag is meaningless as it only effects resetting on the outside trans_complete().

The issue comes in choosing where to reset the _trans_status flag. Currently it is set in trans_complete(). The problem with it being there is that queries before the transaction can set that flag to false and the transaction will fail.

I think it should be done in trans begin and the following line should be changed;
$this->_trans_failure = ($test_mode === TRUE) ? TRUE : FALSE;
Should be ;
if ($this->trans_strict === FALSE)
{
   $this->_trans_status = ($test_mode === TRUE) ? FALSE : TRUE ;
}
This resets the _trans_status flag on the outermost transaction with strict mode off and makes test_mode work. Currently test mode does not work as _trans_failure in not accessed anywhere. In the event that "strict mode" is on subsequent transactions will fail if any query fails. Right now any transaction fails if any previous query fails reguardless of "strict mode".

_Originally posted by @jklovanc in https://github.com/bcit-ci/CodeIgniter/issues/4026#issuecomment-127797894_

kenjis:
**Description**
Fixes #6916

- fix old wrong description
- improve "Managing Errors" explanation 
- small improvements

Related: #6917, #6886, #6912

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -61,7 +61,7 @@ Strict Mode can be disabled as follows:
 Managing Errors
 ===============
 
-If you have error reporting enabled in your Config/Database.php file
+If you have error reporting enabled in your **app/Config/Database.php** file
 you'll see a standard error message if the commit was unsuccessful. If
 debugging is turned off, you can manage your own errors like this:
 
--- a/user_guide_src/source/database/transactions.rst
+++ b/user_guide_src/source/database/transactions.rst
@@ -36,13 +36,13 @@ Running Transactions
 ====================
 
 To run your queries using transactions you will use the
-``$this->db->transStart()`` and ``$this->db->transComplete()`` functions as
+``$this->db->transStart()`` and ``$this->db->transComplete()`` methods as
 follows:
 
 .. literalinclude:: transactions/001.php
 
-You can run as many queries as you want between the start/complete
-functions and they will all be committed or rolled back based on the success
+You can run as many queries as you want between the ``transStart()``/``transComplete()``
+methods and they will all be committed or rolled back based on the success
 or failure of any given query.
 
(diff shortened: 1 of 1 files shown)


kenjis:
@codeigniter4/database-team 
I have some transaction-related PRs,
but please review this fix for the current docs first.

michalsn (approved):

kenjis:
@michalsn Thank you for the review!

PR #6923, opened 2022-11-29
refactor: upgrade to PHP 8.0 with rector

kenjis:
Needs #6922

**Description**
See #6921

Applied rules:
* StrEndsWithRector (https://wiki.php.net/rfc/add_str_starts_with_and_ends_with_functions)
* StrStartsWithRector (https://wiki.php.net/rfc/add_str_starts_with_and_ends_with_functions)
* StrContainsRector (https://externals.io/message/108562 https://github.com/php/php-src/pull/5179)

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -44,12 +44,12 @@ class FileLocator
         $file = $this->ensureExt($file, $ext);
 
         // Clears the folder name if it is at the beginning of the filename
-        if (! empty($folder) && strpos($file, $folder) === 0) {
+        if (! empty($folder) && str_starts_with($file, $folder)) {
             $file = substr($file, strlen($folder . '/'));
         }
 
         // Is not namespaced? Try the application folder.
-        if (strpos($file, '\\') === false) {
+        if (! str_contains($file, '\\')) {
             return $this->legacyLocate($file, $folder);
         }
 
@@ -71,7 +71,7 @@ class FileLocator
         $namespaces = $this->autoloader->getNamespace();
 
         foreach (array_keys($namespaces) as $namespace) {
-            if (substr($file, 0, strlen($namespace)) === $namespace) {
+            if (str_starts_with($file, $namespace)) {
                 // There may be sub-namespaces of the same vendor,
                 // so overwrite them with namespaces found later.
                 $paths = $namespaces[$namespace];
@@ -94,7 +94,7 @@ class FileLocator
             // If we have a folder name, then the calling function
             // expects this file to be within that folder, like 'Views',
             // or 'libraries'.
-            if (! empty($folder) && strpos($path . $filename, '/' . $folder . '/') === false) {
+            if (! empty($folder) && ! str_contains($path . $filename, '/' . $folder . '/')) {
                 $path .= trim($folder, '/') . '/';
             }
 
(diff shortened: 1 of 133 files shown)


system/Autoloader/FileLocator.php
```diff
@@ -123,7 +125,7 @@ public function getClassname(string $file): string
                 continue;
             }
 
-            if ((isset($tokens[$i - 2][1]) && ($tokens[$i - 2][1] === 'phpnamespace' || $tokens[$i - 2][1] === 'namespace')) || ($dlm && $tokens[$i - 1][0] === T_NS_SEPARATOR && $token[0] === T_STRING)) {
+            if ((isset($tokens[$i - 2][1]) && ($tokens[$i - 2][1] === 'phpnamespace' || $tokens[$i - 2][1] === 'namespace')) || ($dlm && $tokens[$i - 1][0] === T_NS_SEPARATOR && $token->is(T_STRING))) {
```
kenjis, line 128: @samsonasik 
This code causes an error.

> Running PHPStan...
Note: Using configuration file .../CodeIgniter4/phpstan.neon.dist.
Error thrown in .../CodeIgniter4/system/Autoloader/FileLocator.php on line 128 while loading bootstrap file .../CodeIgniter4/phpstan-bootstrap.php: Cannot use object of type PhpToken as array

samsonasik: Possibly bug on TokenGetAllToObjectRector

samsonasik:
Cherry-picking rule by rule per-PR from config

https://github.com/rectorphp/rector-src/blob/2b35d2e75ea03ffe643c012485952ec6df3ea2ba/config/set/php80.php#L34


can be a way to make easier to catch breaking changes, so smallest PR will be easy to merge

system/Autoloader/FileLocator.php
kenjis: Yes, this is a bug on TokenGetAllToObjectRector. 
The `$tokens` in the original code is just an array of array, and used as `$tokens[$i - 2][1]`.
But `$tokens` in the refactored code is an array of PhpToken object.

system/Cookie/Cookie.php
```diff
@@ -630,7 +626,7 @@ public function toHeaderString(): string
     /**
      * {@inheritDoc}
      */
-    public function __toString()
+    public function __toString(): string
```
kenjis, line 629: Is this a breaking change?

samsonasik: No, `__toString` is special, it keep working even child doesn't has return type

kenjis:
It seems difficult to apply ClasPropertyAssignToConstructorPromotionRector.
Some classes will be broken due to the existing type mismatches.

samsonasik:
That can posibly due to autoload overlapped, the tweak may be needed https://github.com/rectorphp/rector/blob/main/docs/static_reflection_and_autoload.md

kenjis:
Probably due to property type mismatch between parent class and child class. 
Fixing the type declaration might result in BC, so I skipped them for now.
See #6924

system/Autoloader/FileLocator.php
kenjis: Skipped the rule.

MGatner:
I think the small approach is good. I will also note that I've found CS Fixer to be a great companion tool for Rector during the upgrades to make sure the results are still compliant with our style guide.

PR #6924, opened 2022-11-29
refactor: PHP 8.0 constructor promotion

kenjis:
Needs #6923

**Description**
See #6921

Use constructor promotion.
This may contain breaking changes.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -18,15 +18,11 @@ namespace CodeIgniter\Autoloader;
 class FileLocator
 {
     /**
-     * The Autoloader to use.
-     *
-     * @var Autoloader
+     * @param Autoloader $autoloader The Autoloader to use.
      */
-    protected $autoloader;
-
-    public function __construct(Autoloader $autoloader)
-    {
-        $this->autoloader = $autoloader;
+    public function __construct(
+        protected Autoloader $autoloader
+    ) {
     }
 
     /**
--- a/system/Autoloader/FileLocator.php
+++ b/system/Autoloader/FileLocator.php
@@ -44,12 +44,12 @@ class FileLocator
         $file = $this->ensureExt($file, $ext);
 
         // Clears the folder name if it is at the beginning of the filename
-        if (! empty($folder) && strpos($file, $folder) === 0) {
+        if (! empty($folder) && str_starts_with($file, $folder)) {
             $file = substr($file, strlen($folder . '/'));
         }
 
         // Is not namespaced? Try the application folder.
-        if (strpos($file, '\\') === false) {
+        if (! str_contains($file, '\\')) {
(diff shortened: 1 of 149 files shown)


kenjis:
This is a breaking chage:
https://3v4l.org/BrHEP

samsonasik:
Yes, that seems need to be skipped

system/Autoloader/FileLocator.php
```diff
@@ -18,15 +18,11 @@
...
+     * @param Autoloader $autoloader The Autoloader to use.
      */
-    protected $autoloader;
-
-    public function __construct(Autoloader $autoloader)
-    {
-        $this->autoloader = $autoloader;
+    public function __construct(
+        protected Autoloader $autoloader
+    ) {
```
kenjis, line 25: For example, if a dev extends `FileLocator` and it has `protected $autoloader;`,
this change is a breaking change.
https://3v4l.org/NhqDb

samsonasik: There are possible options:

- make a note about property re-definition possible breaking change without add type
- wait to push to 4.4 only

samsonasik: Another way;: make Rector extension to help user migrating to new definition.

kenjis: Removing type in the constructor makes no errors (I don't know why), 
https://3v4l.org/2A85L
but I don't want to remove already declared types in constructors.

samsonasik: Better way: make a patch to rector to:

- skip non-typed property for protected/public property, except define `INLINE_PUBLIC` config that allow bc break on non-final class, same like TypedPropertyRector way

That's just quick guessing, probably there is a better way.

samsonasik:
@kenjis some rules like `StrEndsWithRector`, `StrStartsWithRector`, `StrContainsRector` seems easy merge 👍 , define small PRs and iterate to next rules when possible.

system/Autoloader/FileLocator.php
samsonasik: @kenjis here possibly the solution, I create PR to make it configurable `INLINE_PUBLIC` default false:

- https://github.com/rectorphp/rector-src/pull/3126

kenjis:
If we drop PHP 7.4 in v4.4, I would like to type missing public/protected properties with constructor promotion.
These will be breaking changes, though.

samsonasik:
Yes, I agree 👍

system/Autoloader/FileLocator.php
MGatner: This seems the best approach to me. Glad your patch was already accepted!

PR #6931, opened 2022-12-02
chore: update psalm to 5.0

kenjis:
**Description**
- update psalm

https://github.com/vimeo/psalm/releases/tag/5.0.0

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Cookie/Cookie.php
+++ b/system/Cookie/Cookie.php
@@ -37,6 +37,8 @@ use ReturnTypeWillChange;
  * $cookie2 = $cookie->withName('prod_cookie');
  * $cookie2->getName(); // prod_cookie
  * ```
+ *
+ * @template-implements ArrayAccess<string, bool|int|string>
  */
 class Cookie implements ArrayAccess, CloneableCookieInterface
 {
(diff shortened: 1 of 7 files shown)


system/Cookie/Cookie.php
```diff
@@ -37,6 +37,8 @@
  * $cookie2 = $cookie->withName('prod_cookie');
  * $cookie2->getName(); // prod_cookie
  * ```
+ *
+ * @template-implements ArrayAccess<string, bool|int|string>
```
paulbalandan, line 41: This is generic typing. If we want to impose this too for phpstan, we may consider raising phpstan level to 6.

paulbalandan (approved):

system/Cookie/Cookie.php
MGatner: 5 to 6 is hard for existing code bases. I would of course love to see our level increase, so maybe we can baseline existing array issues and make the move.

MGatner (approved):
Please consider updating DevKit to match.

PR #6088, opened 2022-06-07
Move preload.php example to starter app

samsonasik:
@kenjis per our discussion on slack.

**Checklist:**
- [x] Securely signed commits

PR #6936, opened 2022-12-04
chore: move preload.php

kenjis:
**Description**
Follow-up #6088
[
`codeigniter4/framework`](https://github.com/codeigniter4/framework) also should have this file.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


samsonasik (approved):

samsonasik:
If `preload.php` is on `framework` part, which in `vendor/` location already, the documentation usage may need update:

https://github.com/codeigniter4/CodeIgniter4/blob/197e99633e8fde8b66eecb1f896dcea29481897a/admin/starter/preload.php#L51

https://github.com/codeigniter4/CodeIgniter4/blob/197e99633e8fde8b66eecb1f896dcea29481897a/admin/starter/preload.php#L19

Or probably better change to:

```php
'include' => __DIR__ . '/system',
```

on inside `framework`

kenjis:
Update the comment.

The `preload.php` should be located in the project root folder.

If it is only in appstarter,
1. devs who do not use appstarter can't get it easily.
2. when it is updated, devs can't update it easily.

samsonasik (approved):
Looks good 👍

samsonasik:
Thank you @kenjis

Issue #6930, opened 2022-12-01
Bug: writing to SQLite3 db fails due to concurrency


dgvirtual:
### PHP Version

7.4

### CodeIgniter4 Version

4.2.1

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

apache

### Database

SQLite3

### What happened?

During the website peak usage times (when people rush to book rooms after a new period opens) the writing to SQLite3 db sometimes fails due to, I presume, concurrent attempts at writing; this is the error as logged: 

```
ERROR - 2022-12-01 08:02:20 --> ErrorException: SQLite3::exec(): database is locked in /home/my/project/myApp/vendor/codeigniter4/framework/system/Database/SQLite3/Connection.php:129
Stack trace:
#0 [internal function]: CodeIgniter\Debug\Exceptions->errorHandler()
#1 /home/my/project/myApp/vendor/codeigniter4/framework/system/Database/SQLite3/Connection.php(129): SQLite3->exec()
#2 /home/my/project/myApp/vendor/codeigniter4/framework/system/Database/BaseConnection.php(695): CodeIgniter\Database\SQLite3\Connection->execute()
#3 /home/my/project/myApp/vendor/codeigniter4/framework/system/Database/BaseConnection.php(609): CodeIgniter\Database\BaseConnection->simpleQuery()
#4 /home/my/project/myApp/vendor/codeigniter4/framework/system/Database/BaseBuilder.php(1904): CodeIgniter\Database\BaseConnection->query()
#5 /home/my/project/myApp/vendor/codeigniter4/framework/system/Model.php(278): CodeIgniter\Database\BaseBuilder->insert()
#6 /home/my/project/myApp/vendor/codeigniter4/framework/system/BaseModel.php(738): CodeIgniter\Model->doInsert()
#7 /home/my/project/myApp/vendor/codeigniter4/framework/system/Model.php(666): CodeIgniter\BaseModel->insert()
#8 /home/my/project/myApp/app/Controllers/Booking.php(278): CodeIgniter\Model->insert()
#9 /home/my/project/myApp/vendor/codeigniter4/framework/system/CodeIgniter.php(896): App\Controllers\Booking->reserve_hour()
#10 /home/my/project/myApp/vendor/codeigniter4/framework/system/CodeIgniter.php(466): CodeIgniter\CodeIgniter->runController()
[...]

PR #6939, opened 2022-12-05
feat: add SQLite3 Config busyTimeout

kenjis:
**Description**
See #6930

- add a new Config `busyTimeout`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Database.php
+++ b/app/Config/Database.php
@@ -67,6 +67,7 @@ class Database extends Config
         'failover'    => [],
         'port'        => 3306,
         'foreignKeys' => true,
+        'busyTimeout' => 1000,
     ];
 
     public function __construct()
--- a/system/Database/SQLite3/Connection.php
+++ b/system/Database/SQLite3/Connection.php
@@ -63,7 +63,7 @@ class Connection extends BaseConnection
         }
 
         if (is_int($this->busyTimeout)) {
-            $this->simpleQuery('PRAGMA busy_timeout=' . $this->busyTimeout);
+            $this->connID->busyTimeout($this->busyTimeout);
         }
     }
 
--- a/system/Database/SQLite3/Connection.php
+++ b/system/Database/SQLite3/Connection.php
@@ -45,12 +45,25 @@ class Connection extends BaseConnection
      */
     protected $foreignKeys = false;
 
+    /**
+     * The milliseconds to sleep
+     *
+     * @var int|null milliseconds
+     *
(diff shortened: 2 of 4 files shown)


dgvirtual:
@kenjis, I have a question, is there any particular reason why uniqueKeys and busyTimeout config options are added only on the test database array in the config file? Is it so since the main (default) db array defaults to mysqli driver, and the array has to only contain what could used in default case? 

I am thinking how to implement the routine of installation on apps like Bonfire2, where the DB is chosen and config options are set in CLI by a script modifying .env file. To add  uniqueKeys and busyTimeout options one would need to also modify the Database.php config file using CLI, as entries only in .env would have no effect if the keys are not set in the Database.php file.

kenjis:
Yes, the default is for MySQLi, and the tests is for SQLite3.

The `Config/Database.php` setting should be updated for the database driver you use.

MGatner (approved):
Actually read and understood this one 😁 Thanks for the quick solution! Looks good to me.

michalsn (commented):
Although this will fix the reported "bug", I'm not sure if this is the right path to solve this type of issues.

The next "bug" I expect to pop up will probably be support for [WAL](https://sqlite.org/wal.html) in SQLite3. The question we should ask is whether or not we're planning to add this type of config position for every single item. Also, what about other databases - there are some init options in MySQL... in other databases as well.

I would be opting for providing some universal config feature for options that developers want to fire upon connection initialization.

dgvirtual:
My main interest is to have the issue resolved (which any solution in config does). But in support of the @kenjis solution, I think that a special-config-option-based way to deal with concurrency (like handling foreign keys, for that matter, which is one other sqlite specific option) could be considered preferable to a generic way (which should also be made available), because it helps to turn on features that are on by default in mysql/postgresql, and that are assumed by some popular addons (like support for foreign keys is assumed by Shield).

kenjis:
The WAL mode is persistent. We don't need the config for it.
https://sqlite.org/wal.html#persistence_of_wal_mode

> The question we should ask is whether or not we're planning to add this type of config position for every single item. Also, what about other databases - there are some init options in MySQL... in other databases as well.

We can add `options` for init options.

I think it is better to add an config item for this in this case.
It can tell there is the config to devs.

kenjis:
I've changed the implementation.

paulbalandan (approved):

michalsn (approved):

Issue #6306, opened 2022-07-28
Parse error: syntax error, unexpected '_536_000' (T_STRING), expecting ')'


fre2mansur:
### PHP Version

7.3

### CodeIgniter4 Version

4

### CodeIgniter4 Installation Method

Manual (zip or tar.gz)

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

apache

### What happened?

Parse error: syntax error, unexpected '_536_000' (T_STRING), expecting ')' in E:\MYPROJECTS\codignitor\system\Common.php on line 480

### Steps to Reproduce

php spark serve

### Expected Output

This will launch the server and you can now view your application in your browser at http://localhost:8080/.

### Anything else?

![image](https://user-images.githubusercontent.com/8459473/181398575-92f1d3a5-bd3d-46f9-8f18-69c0218e2131.png)

datamweb:
Hi, CodeIgniter Version(eg : 4.2.0)?

kenjis:
CodeIgnier 4.2 requires PHP 7.4+
https://codeigniter4.github.io/CodeIgniter4/changelogs/v4.2.0.html#highlights

paulbalandan:
Actually, this can be a bug in how we bootstrap the framework. The PHP version check happens in `index.php` and `spark` after the  `bootstrap.php` file is `require`d. In the bootstrap file, we `require` both `Common.php` and `Constants.php` which contains version-specific features (e.g. numeric literal separators).

I think if we can move the version check before requiring these files, then we can avoid these reports in the future.

kenjis:
> Actually, this can be a bug in how we bootstrap the framework.

Exactly. The version check should be done in the very beginning of index.php and spark.

fre2mansur:
I downloaded the latest version as zip.
It works fine when I  install from cmd composer so I don't think the php version is the issue.

kenjis:
@fre2mansur Composer installs the version that works with your php.
Probably you have CI v4.1.9 on PHP 7.3.

Check the CI version:
```
$ php spark
```

PR #6945, opened 2022-12-06
Unify platform checking

paulbalandan:
**Description**
- Adds a `platform_check.php` to where all platform checks are performed. (The format of the file is patterned from Composer)
- Remove checks in spark, index.php, CodeIgniter::resolvePlatformExtensions
- Move the check inside bootstrap

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


system/platform_check.php
```diff
@@ -0,0 +1,52 @@
...
+    if (! headers_sent()) {
+        header('HTTP/1.1 500 Internal Server Error');
+    }
+
+    $preface = 'CodeIgniter cannot be booted because of issues detected in your platform:';
+
+    if (! ini_get('display_errors')) {
+        if (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') {
+            fwrite(STDERR, $preface . PHP_EOL . PHP_EOL . implode(PHP_EOL, $issues) . PHP_EOL . PHP_EOL);
+        } elseif (! headers_sent()) {
```
kenjis, line 46: Why `! headers_sent()`?

paulbalandan: This was copied from composer's version.

MGatner (commented):
It feels like we just moved these. 😳. Can't find the thread now but I seem to recall there was a case for doing the check immediately at the entry point because some files (don't remember if this included **bootstrap.php**) would not parse in lower versions of PHP.

kenjis:
Yes. See #6306

system/Test/bootstrap.php
```diff
@@ -49,6 +49,14 @@
 defined('COMPOSER_PATH') || define('COMPOSER_PATH', realpath(HOMEPATH . 'vendor/autoload.php'));
 defined('VENDORPATH')    || define('VENDORPATH', realpath(HOMEPATH . 'vendor') . DIRECTORY_SEPARATOR);
 
+/*
+ * ---------------------------------------------------------------
+ * CHECK PLATFORM REQUIREMENTS
+ * ---------------------------------------------------------------
+ */
+
+require SYSTEMPATH . 'platform_check.php';
```
samsonasik, line 58: need to be before `new Paths()` to avoid class constant concat string error on php 5.5 ref https://3v4l.org/WvlTa#v5.5.30

Yes, I found a user ask in facebook why got error on upload to shared hostig

paulbalandan:
Sorry for the late turnaround.

I moved the check inside bootstrap.php and after the constants definition because I need to have `SYSTEMPATH` defined because the platform check is inside "system/". I can move the check early but I will need to use `__DIR__./../` to reach `system/`. If this would be the compromise taken, I'll revise the implementation.

samsonasik:
probably extract to 2 files: `php_requirement_check.php` (early load) and `extension_check.php` ?

samsonasik:
Oh, I see, `SYSTEMPATH` got from $path object, so overlapped check, as new Paths() will got e
Syntax error early

samsonasik:
I prefer the old implementation, duplication is fine on purpose to avoid syntax error happen again.

system/platform_check.php
```diff
@@ -0,0 +1,52 @@
...
+}
+
+foreach (['intl', 'json', 'mbstring'] as $extension) {
+    if (! extension_loaded($extension)) {
+        $issues[] = sprintf(
+            'The framework needs the following extension installed and loaded: %s.',
+            $extension
+        );
+    }
+}
```
kenjis, line 34: These checks are needed only for manual installations.

system/platform_check.php
kenjis: The header is sent in the line 38.

kenjis:
I don't think there is much benefit in this change.
However, it does not work on PHP 5.5 as @samsonasik says.
> Parse error: syntax error, unexpected 'string' (T_STRING), expecting variable (T_VARIABLE) in /usr/src/myapp/app/Config/Paths.php on line 26

And we need to include one more file.

PR #6948, opened 2022-12-07
fix: Route placeholder `(:any)` does not pass the whole value to controller

### LINKED ISSUE #6947: Bug: Route placeholder `(:any)` does not pass the whole value to controller
### PHP Version

8.1

### CodeIgniter4 Version

`develop` (ab55fe1994210b873d49246ed24a2141ea8a55fa)

Git

macOS

### Which server did you use?

### Database

n/a

### What happened?

`(:any)` does not pass the matched value, but passes the first segment only.

### Steps to Reproduce

```php
$routes->get('(:any)', 'Home::index/$1');
```

```php
<?php

namespace App\Controllers;

{
    public function index($any = '')
    {
        dd($any);
    }
}
```

Navigate to http://localhost:8080/abc/def ,
you will see `$any string (3) "abc"`.

### Expected Output

`$any string (3) "abc/def"`

### Anything else?

```php
<?php

namespace App\Controllers;

{
    public function index($any = '', $next = '')
    {
        dd($any, $next);
    }
}
```

Navigate to http://localhost:8080/abc/def ,
you will see:
```
$any string (3) "abc"
$next string (3) "def"
```

kenjis:
**Description**
Fixes #6947

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -419,8 +419,8 @@ class Router implements RouterInterface
 
             // Does the RegEx match?
             if (preg_match('#^' . $routeKey . '$#u', $uri, $matches)) {
-                $matchedPlaceholders = $matches;
-                unset($matchedPlaceholders[0]);
+                $placeholderValues = $matches;
+                unset($placeholderValues[0]);
 
                 // Is this route supposed to redirect to another?
                 if ($this->collection->isRedirect($routeKey)) {
@@ -462,7 +462,7 @@ class Router implements RouterInterface
                 if (! is_string($handler) && is_callable($handler)) {
                     $this->controller = $handler;
 
-                    $this->params = $matchedPlaceholders;
+                    $this->params = $placeholderValues;
 
                     $this->setMatchedRoute($matchedKey, $handler);
 
@@ -495,7 +495,7 @@ class Router implements RouterInterface
 
                     foreach ($params as $param) {
                         $i          = ltrim($param, '$');
-                        $segments[] = $matchedPlaceholders[$i];
+                        $segments[] = $placeholderValues[$i];
                     }
                 }
                 $this->setRequest($segments);
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -462,10 +462,7 @@ class Router implements RouterInterface
                 if (! is_string($handler) && is_callable($handler)) {
                     $this->controller = $handler;
 
-                    // Remove the original string from the matches array
-                    array_shift($matches);
-
-                    $this->params = $matches;
+                    $this->params = $matchedPlaceholders;
 
                     $this->setMatchedRoute($matchedKey, $handler);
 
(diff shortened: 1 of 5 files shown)


samsonasik (approved):

michalsn (approved):

MGatner (approved):
Curious, I wonder if it has always been like this? I hope this isn't a surprise break for anyone!

michalsn:
From what I see, the docs are just fine. Nothing has changed.

kenjis:
This is a bug, but a breaking change. So I would like to move this to `4.3`.
I don't think there is any need to urgently fix this in 4.2, since it has been this way all along.

kenjis:
Added the docs.

kenjis:
> At least this is what I am used to from the CI3 times.

CI3 does not have `(:any)` in CI4. Do you mean you did that in CI3 with auto-routing?

> I understand the reason for this change from the perspective of the example provided in the tests, but maybe we should leave the current behavior of (:any) as it is and add another one which will behave as described in this PR? Something like (:any-segment) or something? I just think that the current behavior of the (:any) placeholder may be useful in many cases and changing it may break a lot of things.

This may be useful, but it seems we cannot implement it easily.

But I found the documentation that the behavior is not a bug.
> Note that a single (:any) will match multiple segments in the URL if present. For example the route:

```php
<?php
$routes->get('product/(:any)', 'Catalog::productLookup/$1');
```

> will match product/123, product/123/456, product/123/456/789 and so on. The implementation in the Controller should take into account the maximum parameters:

https://codeigniter4.github.io/CodeIgniter4/incoming/routing.html#examples

michalsn:
> CI3 does not have (:any) in CI4. Do you mean you did that in CI3 with auto-routing?

Oh, there is `(:any)` in CI3 [https://codeigniter.com/userguide3/general/routing.html#examples](https://codeigniter.com/userguide3/general/routing.html#examples)

> But I found the documentation that the behavior is not a bug.

Okay, thanks for digging into it.

kenjis:
@michalsn Do you mean `(:any)` in CI3 works like CI4's `(:any)`?
I think `(:any)` in CI3 is `(:segment)` in CI4. See #6957
Am I wrong?

michalsn:
Yes, you might be right, and I confused it with auto-routing in CI3.

kenjis:
My conclusion for now is that this behavior is a specification and cannot be changed. 

However, you should not add any placeholder after `(:any)`, and `(:any)` should be used only once in a route.

mackxpall:
Telling me you are just - getting here.

michalsn:
> However, you should not add any placeholder after (:any), and (:any) should be used only once in a route.

Sounds about right to me.

kenjis:
Closed by #6960

PR #6954, opened 2022-12-07
fix: don't change the variable type and filter all values in JSON request

### LINKED ISSUE #6878: `IncomingRequest::getVar('key')` casts some JSON value to string
`int` and `bool` values are cast to `string`.
Is this intended behavior or a bug?

The following test passes:
```php
    public function testGetVarAndGetVarSpecificItem()
    {
        $data = [
            'str'   => 'str',
            'int'   => 123,
            'null'  => null,
            'true'  => true, 'false' => false,
            'array' => ['a', 1],
            'obj'   => (object) ['a' => 'a', 'b' => 123],
        ];
        $json = json_encode($data);

        $config          = new App();
        $config->baseURL = 'http://example.com/';

        $request = new IncomingRequest($config, new URI(), $json, new UserAgent());
        $request->setHeader('Content-Type', 'application/json');

        $all = $request->getVar();

        $expected = (object) $data;
        $this->assertEqualsCanonicalizing($expected, $all);

        $this->assertSame('str', $request->getVar('str'));
        // converted to string.
        $this->assertSame('123', $request->getVar('int'));
        $this->assertNull($request->getVar('null'));
        // converted to string.
        $this->assertSame('1', $request->getVar('true'));
        // converted to string.
        $this->assertSame('', $request->getVar('false'));
        $this->assertSame(['a', 1], $request->getVar('array'));
        $this->assertEqualsCanonicalizing(
            (object) ['a' => 'a', 'b' => 123],
            $request->getVar('obj')
        );
    }
```

michalsn:
**Description**
Fixes #6878

Current behavior is inconsistence.

1. `getVar(null)` will call `getJSON()` which will return all the data without filtering.
2. `getVar('index')` will be filtered, but... it seems like `getJsonVar()` doesn't perform filtering on nested values (arrays, objects), only on single variables.

This is very problematic:

* some variables may not be filtered at all (arrays etc.), even if we set a proper flag - it will be ignored
* if we call a single variable, it will be converted to a string due to the nature of the `filter_var()` function
* basically, we can get different variable types depending on how we ask for them

I targeted this PR to 4.3 because I'm changing the parameter signature in one method. This will make the code more consistent, and the functionality will be the same as we have with `getGet()`, etc. methods.

BC label is related to changing a signature for `$index` parameter in `IncomingRequest::getJsonVar()`.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -185,6 +185,8 @@ Others
     - ``Debug\Exceptions::__construct()``
     - ``Services::exceptions()``
 
+- **Request:** The ``$index`` parameter of ``IncomingRequest::getJsonVar()`` now accepts an ``array``, ``string`` or ``null`` value.
+
 Enhancements
 ************
 
@@ -342,3 +344,5 @@ Bugs Fixed
 **********
 
 - Fixed a bug when all types of ``Prepared Queries`` were returning a ``Result`` object instead of a bool value for write-type queries.
+- Fixed a bug when not always all variables in a JSON request were filtered if filters were applied with ``IncomingRequest::getVar()`` or ``IncomingRequest::getJsonVar()`` methods.
+- Fixed a bug when variable type may be changed when using a specified index with ``IncomingRequest::getVar()`` or ``IncomingRequest::getJsonVar()`` methods.
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -344,5 +344,5 @@ Bugs Fixed
 **********
 
 - Fixed a bug when all types of ``Prepared Queries`` were returning a ``Result`` object instead of a bool value for write-type queries.
-- Fixed a bug when not always all variables in a JSON request were filtered if filters were applied with ``IncomingRequest::getVar()`` or ``IncomingRequest::getJsonVar()`` methods.
+- Fixed a bug with variable filtering in JSON requests using with ``IncomingRequest::getVar()`` or ``IncomingRequest::getJsonVar()`` methods.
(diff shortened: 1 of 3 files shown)


kenjis (approved):
LGTM

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -331,3 +333,5 @@ Bugs Fixed
 **********
 
 - Fixed a bug when all types of ``Prepared Queries`` were returning a ``Result`` object instead of a bool value for write-type queries.
+- Fixed a bug when not always all variables in a JSON request were filtered if filters were applied with ``IncomingRequest::getVar()`` or ``IncomingRequest::getJsonVar()`` methods.
```
MGatner, line 336:
```suggestion
- Fixed a bug with variable filtering in JSON requests ysing with ``IncomingRequest::getVar()`` or ``IncomingRequest::getJsonVar()`` methods.
```

MGatner (approved):
One wording suggestion, but looks good!

user_guide_src/source/changelogs/v4.3.0.rst
kenjis: What's *ysing* ?

michalsn: Thanks. Fixed.

kenjis (approved):

michalsn:
Psalm error is handled in another PR already, so I'm not touching it here.

user_guide_src/source/changelogs/v4.3.0.rst
MGatner: Haha oops! Sorry for the typo, yes that was supposed to be "using" (and drop the "with"). Looks good.

MGatner (approved):

PR #6958, opened 2022-12-08
Fix: FeatureTestTrait does not use custom Request class

lonnieezell:
When using a custom extension of the `IncomingRequest` class, running Feature tests would fail with a TypeError. The problem is that FeatureTestTrait would directly create an instance of the framework's IncomingRequest instead of grabbing it from the Services class. 

This changes that to use it from the Service class instead.


kenjis:
Why `4.3` branch?

lonnieezell:
Isn't that our next release?

kenjis:
v4.2.11 is the next release.

MGatner:
Looks good! But yes, let's send this to develop

lonnieezell:
Fixed and replaced by #6966 

Closing

PR #6959, opened 2022-12-08
docs: fix phpdoc types

kenjis:
**Description**
See #6310

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/CommonSingleServiceTest.php
+++ b/tests/system/CommonSingleServiceTest.php
@@ -36,6 +36,9 @@ final class CommonSingleServiceTest extends CIUnitTestCase
         $service1 = single_service($service);
         $service2 = single_service($service);
 
+        assert($service1 !== null);
+        assert($service1 !== null);
+
         $this->assertInstanceOf(get_class($service1), $service2);
         $this->assertNotSame($service1, $service2);
     }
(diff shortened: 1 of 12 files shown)


tests/system/CommonSingleServiceTest.php
```diff
@@ -36,6 +36,9 @@ public function testSingleServiceWithNoParamsSupplied(string $service): void
         $service1 = single_service($service);
         $service2 = single_service($service);
 
+        assert($service1 !== null);
+        assert($service1 !== null);
```
samsonasik, line 40: should be service1 and service2 ? Above 2 lines are same (service1)

tests/system/CommonSingleServiceTest.php
```diff
@@ -64,6 +67,9 @@ public function testSingleServiceWithAtLeastOneParamSupplied(string $service): v
         $service1 = single_service($service, ...$params);
         $service2 = single_service($service, ...$params);
 
+        assert($service1 !== null);
+        assert($service1 !== null);
```
samsonasik, line 71: here too, service1 and 2

tests/system/CommonSingleServiceTest.php
kenjis: Yes. it is copy and paste mistake.
But the `$service2` is not needed. So I removed it.

tests/system/CommonSingleServiceTest.php
kenjis: Fixed.

michalsn (approved):

PR #6966, opened 2022-12-10
Use Services for the FeatureTestTrait request.

lonnieezell:
Replaces #6958 

When using a custom extension of the IncomingRequest class, running Feature tests would fail with a TypeError. The problem is that FeatureTestTrait would directly create an instance of the framework's IncomingRequest instead of grabbing it from the Services class.

This changes that to use it from the Service class instead.


samsonasik (approved):

samsonasik:
Thank you @lonnieezell

PR #6971, opened 2022-12-12
docs: update changelogs 4.3.0

kenjis:
**Description**
- add highlights
- add missing items

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -323,6 +323,7 @@ Changes
 
 - Config
     - All atomic type properties in ``Config`` classes have been typed.
+    - See :ref:`Upgrading <upgrade_430_config>` for information on changing the default values.
 - Changed the processing of Spark commands:
     - The ``CodeIgniter\CodeIgniter`` no longer handles Spark commands.
     - The ``CodeIgniter::isSparked()`` method has been removed.
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -21,7 +21,9 @@ Highlights
   See `Forge`_.
 - To make the default configuration more secure, the default Validation Rules
   have been changed to **Strict Rules**.
-- The behavior when a database error occurs has been changed. See :ref:`exceptions-when-database-errors-occur`.
+- When a database error occurs, the conditions that an exception is thrown and
+  the exception classes that can be thrown have been changed.
+  See :ref:`exceptions-when-database-errors-occur`.
 
 BREAKING
 ********
--- a/user_guide_src/source/changelogs/v4.3.0.rst
+++ b/user_guide_src/source/changelogs/v4.3.0.rst
@@ -12,7 +12,16 @@ Release Date: Unreleased
 Highlights
 **********
 
-- TBD
+- Query Builder supports **upsert()**, **upsertBatch()**, **deleteBatch()** and
+  now ***batch()** methods can set data from a query
+  (*contributed by* `sclubricants <https://github.com/sclubricants>`_).
+  See `Query Builder`_.
+- Database Forge supports to **add indexes in the existing tables** and
+  **name indexes** (*contributed by* `sclubricants <https://github.com/sclubricants>`_).
+  See `Forge`_.
+- To make the default configuration more secure, the default Validation Rules
+  have been changed to **Strict Rules**.
+- The behavior when a database error occurs has been changed. See :ref:`exceptions-when-database-errors-occur`.
 
 BREAKING
 ********
--- a/user_guide_src/source/installation/upgrade_430.rst
+++ b/user_guide_src/source/installation/upgrade_430.rst
@@ -260,6 +260,9 @@ Config
 - app/Config/Exceptions.php
     - Two additional public properties were added: ``$logDeprecations`` and ``$deprecationLogLevel``.
       See See :ref:`logging_deprecation_warnings` for details.
+- app/Config/Logger.php
+    - The property ``$threshold`` has been changed to ``9`` in other than ``production``
+      environment.
 - app/Config/Modules.php
     - The new property ``$composerPackages`` is added to limit Composer package Auto-Discovery for better
       performance.
--- a/user_guide_src/source/installation/upgrade_430.rst
+++ b/user_guide_src/source/installation/upgrade_430.rst
@@ -236,6 +236,8 @@ Content Changes
 The following files received significant changes (including deprecations or visual adjustments)
 and it is recommended that you merge the updated versions with your application:
 
+.. _upgrade_430_config:
+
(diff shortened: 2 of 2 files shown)


michalsn (approved):

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -12,7 +12,16 @@ Release Date: Unreleased
...
+- Query Builder supports **upsert()**, **upsertBatch()**, **deleteBatch()** and
+  now ***batch()** methods can set data from a query
+  (*contributed by* `sclubricants <https://github.com/sclubricants>`_).
+  See `Query Builder`_.
+- Database Forge supports to **add indexes in the existing tables** and
+  **name indexes** (*contributed by* `sclubricants <https://github.com/sclubricants>`_).
+  See `Forge`_.
+- To make the default configuration more secure, the default Validation Rules
+  have been changed to **Strict Rules**.
+- The behavior when a database error occurs has been changed. See :ref:`exceptions-when-database-errors-occur`.
```
paulbalandan, line 24: This feels hanging even though included in Highlights. Can you state briefly what is that behavior that changed here?

kenjis: It is a difficult question.
Like this?

> When a database error occurs, the conditions that an exception is thrown and the exception classes that can be thrown have been changed.

MGatner (approved):
I'm fine with either on the database errors.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -12,7 +12,16 @@ Release Date: Unreleased
...
 
 BREAKING
 ********
```
sclubricants, line 27: >- **Model:** ``Model::update()`` method now raises a ``DatabaseException`` if it **generates an SQL
  statement** without a WHERE clause; Model does not support operations that update all records.

Line 99

Should  be **generates a SQL statement**

sclubricants: Line 181 and 184 seem to have an unneeded empty line.

Line 183

>- In addition to the added ``$asQuery`` parameter above the following methods also now **return an array**.

Should be:

- In addition to the added ``$asQuery`` parameter above the following methods also now **returns an array**.

sclubricants (commented):
>- ``BaseConnection::escape()`` now excludes the ``RawSql`` data type. This allows passing SQL strings into data.

This is potentially a really helpful feature, not sure if it deserves mention in highlights.

user_guide_src/source/changelogs/v4.3.0.rst
kenjis: It seems both `a SQL` and `an SQL` can be used.
https://english.stackexchange.com/questions/5265/an-sql-server-database-schema-or-a-sql-server-database-schema

user_guide_src/source/changelogs/v4.3.0.rst
kenjis: Done.

kenjis:
>>- ``BaseConnection::escape()`` now excludes the ``RawSql`` data type. This allows passing SQL strings into data.
>
> This is potentially a really helpful feature, not sure if it deserves mention in highlights. 

Just reading that description, I can't see where it can be used.
So I don't think it will be understood very well if we describe it in the highlights.

user_guide_src/source/changelogs/v4.3.0.rst
paulbalandan: I use `an SQL` when referring to the language because it is an acronym of Structured Query Language and should be pronounced as S-Q-L. If referring to the driver Microsoft has, then it is `a SQL` because it's pronounced as sequel.

`return` is correct because it refers to the `following methods`.

kenjis: > SQL does not begin with a vowel sound.. even if S-Q-L is the way you're saying it.

What do you mean?
I assume to pronounce this way.
https://translate.google.co.jp/?sl=en&tl=ja&text=an%20S%20Q%20L&op=translate

kenjis: > the also suggests plurality so your talking about more than one thing returning something.

I am not a native English speaker and I honestly don't know the difference.
I hope that native speakers can discuss and decide.

paulbalandan: > SQL does not begin with a vowel sound.. even if S-Q-L is the way you're saying it.

No offense but English is my second language since we're taught with that since grade school. And the [IPA](https://en.wikipedia.org/wiki/International_Phonetic_Alphabet) pronunciation of SQL is either `[/ˈsiːkwəl/]` "sequel" or `[/ˌɛsˌkjuːˈɛl/]` "S-Q-L". If you would listen to the [pronunciation](https://upload.wikimedia.org/wikipedia/commons/5/5f/En-us-SQL.ogg) of the second, it is clearly like "es-q-el" so it starts with a vowel sound. Maybe you're confusing vowel sound with vowel letter? Continuing with your example, "university" starts with the vowel letter "u" but starts with the consonant sound "y", so it is "a university" not "an university".

> the also suggests plurality so your talking about more than one thing returning something.

That's true, and "following methods" is a plural subject so we use the plural form (which is also the base form) of the verb  "return" which is also "return". "returns" is the singular form, isn't it?

sclubricants: You could say "an S-Q-L statement" 

This is correct. 

"An S-Q-L" is ambiguous

sclubricants: Just out of curiosity for you guys is it:

A SQLite database

Or

An SQLite database

kenjis: It seems to me that English native speakers pronounce "sequel".

In my country, SQL is a foreign word, and it is pronounced like "ess-queue-el" 100%.
It may surprise you but Microsoft SQL Server is pronounced like "ess-queue-el server".

SQLite is like "ess-queue-light".

MGatner: I move between both of them so I'm fine with either solution. I don't think there's a right or wrong here.

sclubricants: >The official way to pronounce “MySQL” is “My Ess Que Ell” (not “my sequel”), but we do not mind if you pronounce it as “my sequel” or in some other localized way.

https://dev.mysql.com/doc/refman/8.0/en/what-is-mysql.html#idm139761506951328

sclubricants: SQLite creator says pronounce it however you like.

https://english.stackexchange.com/questions/431329/what-is-the-correct-pronunciation-of-sqlite

PR #6979, opened 2022-12-14
[Rector] Update to Rector 0.15.1 and re-enable TypedPropertyFromAssignsRector

samsonasik:
**Description**

@kenjis @paulbalandan Rector 0.15.1 just released and we can re-enable `TypedPropertyFromAssignsRector` that was buggy on `union mixed type, new class extends SomeClass marked as object in union, and false replaced with bool in Union` which currently resolved.

**Checklist:**
- [x] Securely signed commits


samsonasik:
All green 🎉

kenjis (approved):

PR #6980, opened 2022-12-15
[4.3] fix: merge mistake for rector.php

kenjis:
**Description**
~~I forgot to add it when merging.~~
Remove the line that was removed in `develop`.
It should be removed when merging, but maybe I did something wrong when resolving conflicts.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -48,6 +48,7 @@ use Rector\Privatization\Rector\Property\PrivatizeFinalClassPropertyRector;
 use Rector\PSR4\Rector\FileWithoutNamespace\NormalizeNamespaceByPSR4ComposerAutoloadRector;
 use Rector\Set\ValueObject\LevelSetList;
 use Rector\Set\ValueObject\SetList;
+use Rector\TypeDeclaration\Rector\Property\TypedPropertyFromAssignsRector;
 use Utils\Rector\PassStrictParameterToFunctionParameterRector;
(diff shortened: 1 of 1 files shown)


samsonasik (approved):

rector.php
```diff
@@ -48,6 +48,7 @@
 use Rector\PSR4\Rector\FileWithoutNamespace\NormalizeNamespaceByPSR4ComposerAutoloadRector;
 use Rector\Set\ValueObject\LevelSetList;
 use Rector\Set\ValueObject\SetList;
+use Rector\TypeDeclaration\Rector\Property\TypedPropertyFromAssignsRector;
```
samsonasik, line 51: Actually, no need, as no longer skipped in develop

samsonasik (changes requested):
TypedPropertyFromAssignsRector already enabled at develop

- https://github.com/codeigniter4/CodeIgniter4/pull/6979

so no need to skip it

samsonasik (approved):

kenjis:
Yes, `TypedPropertyFromAssignsRector` should not be skipped.

Frankly, I am not sure why the line in skip was not deleted when I merged `develop`.
But in fact, the line is still there in `4.3`, so I will remove it in this PR.

PR #6984, opened 2022-12-16
fix: large HTTP input crashes framework

### LINKED ISSUE #6246: Large HTTP Input Crashes
### PHP Version

7.4

### CodeIgniter4 Version

4.1

Linux

### Which server did you use?

apache

### Database

_No response_

### What happened?

When `php://input` exceeds the PHP memory limit it crashes the framework. 

>  `CONTENT_LENGTH` is *only* set on POST requests and refers specifically to the size of the POST content. The POST content is **always** available from `php://input`, whether it is valid or not. The problem is that we read the input stream regardless of POST validity - we should instead be doing something like checking 
>
> `if (isset($_SERVER['CONTENT_LENGTH']) && $_SERVER['CONTENT_LENGTH'] > get_post_max_value()) ...`
>
> This will prevent reading invalid POST content into the request body. We will also need to check if there are other locations where `php://input` is accessed, and if those need similar checks or if we should somehow invalidate the resource. 

https://github.com/codeigniter4/CodeIgniter4/pull/4748#issuecomment-859526582 for resolution

### Steps to Reproduce

https://github.com/codeigniter4/CodeIgniter4/pull/4748

### Expected Output

Need to be able to handle inputs that exceed memory limit so a developer can manipulate them (via streams probably). 

### Anything else?

_No response_

kenjis:
**Description**
Fixes #6246

- add the conditions not to get the body from `php://input`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -154,8 +154,16 @@ class IncomingRequest extends Request
             throw new InvalidArgumentException('You must supply the parameters: uri, userAgent.');
         }
 
-        // Get our body from php://input
-        if ($body === 'php://input') {
+        $this->populateHeaders();
+
+        if (
+            $body === 'php://input'
+            // php://input is not available with enctype="multipart/form-data".
+            // See https://www.php.net/manual/en/wrappers.php.php#wrappers.php.input
+            && strpos($this->getHeaderLine('Content-Type'), 'multipart/form-data') === false
+            && (int) $this->getHeaderLine('Content-Length') <= $this->getPostMaxSize()
+        ) {
+            // Get our body from php://input
             $body = file_get_contents('php://input');
         }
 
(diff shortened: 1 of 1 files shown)


kenjis:
I think `$body` should be a stream, not a string.

Now and also in this PR  `$body` is set from `php://input` when a POST request comes,
but normally we do not use `$body`, but use data from `$_POST` superglobal.

lonnieezell (approved):
I think this is fine for now, but I agree - switching it to use streams would be much more capable for large inputs.

michalsn (approved):

samsonasik (approved):

samsonasik:
Thank you @kenjis

MGatner (commented):
Thanks! Looks great.

PR #6989, opened 2022-12-17
refactor: add Config\Session

kenjis:
**Description**
- add Config\Session

Related: https://github.com/codeigniter4/CodeIgniter4/issues/4297#issuecomment-813110373

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Session.php
+++ b/app/Config/Session.php
@@ -4,6 +4,7 @@ namespace Config;
 
 use CodeIgniter\Config\BaseConfig;
 use CodeIgniter\Session\Handlers\FileHandler;
+use SessionHandlerInterface;
 
 class Session extends BaseConfig
 {
@@ -18,7 +19,7 @@ class Session extends BaseConfig
      * - `CodeIgniter\Session\Handlers\MemcachedHandler`
      * - `CodeIgniter\Session\Handlers\RedisHandler`
      *
-     * @phpstan-var class-string
+     * @phpstan-var class-string<SessionHandlerInterface>
      */
     public string $driver = FileHandler::class;
 
--- a/app/Config/Session.php
+++ b/app/Config/Session.php
@@ -97,5 +97,5 @@ class Session extends BaseConfig
      *
      * DB Group for the database session.
      */
-    public ?bool $DBGroup = null;
+    public ?string $DBGroup = null;
 }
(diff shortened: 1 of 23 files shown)


michalsn (approved):

app/Config/Session.php
```diff
@@ -0,0 +1,101 @@
...
+     * Session Driver
+     * --------------------------------------------------------------------------
+     *
+     * The session storage driver to use:
+     * - `CodeIgniter\Session\Handlers\FileHandler`
+     * - `CodeIgniter\Session\Handlers\DatabaseHandler`
+     * - `CodeIgniter\Session\Handlers\MemcachedHandler`
+     * - `CodeIgniter\Session\Handlers\RedisHandler`
+     *
+     * @phpstan-var class-string
```
MGatner, line 21:
```suggestion
     * @phpstan-var class-string<SessionHandlerInterface>
```

system/Session/Handlers/FileHandler.php
```diff
@@ -67,9 +67,9 @@ public function __construct(AppConfig $config, string $ipAddress)
     {
         parent::__construct($config, $ipAddress);
 
-        if (! empty($config->sessionSavePath)) {
-            $this->savePath = rtrim($config->sessionSavePath, '/\\');
-            ini_set('session.save_path', $config->sessionSavePath);
+        if (! empty($this->savePath)) {
```
MGatner, line 70: Why did this change?

MGatner (changes requested):
I haven't made it through all the changes to tests yet.

system/Session/Handlers/FileHandler.php
kenjis: Because `$this->savePath` is set in the parent class.
See https://github.com/codeigniter4/CodeIgniter4/pull/6989/files#diff-d0ff4c461b6990417b08b47e7967ea04338dd557b3a04155896940c117a3a37fR122

app/Config/Session.php
kenjis: Done.

MGatner (approved):
Thanks @kenjis! That was a big one, but something I always wanted to see happen. ❤️

PR #6990, opened 2022-12-18
style: run php-cs-fixer v3.13.1

kenjis:
**Description**
- run php-cs-fixer

See https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/6590

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


michalsn:
I'm not a big fan 
* it makes the code less readable in many cases 
* it's not very natural for me to write code this way (it will probably be popping up frequently)

However, I'm not sure what others think. If the majority is on board, then let it be.

Also not sure there is an easy alternative to avoid this.

kenjis:
To be honest, I'm also not a fan of this.

But the change is a bug fix of php-cs-fixer.
I don't know we can avoid this easily.

kenjis:
I created an issue: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/6716

MGatner:
I think we hold off on this until we see what comes of kenjis' request.

PR #6975, opened 2022-12-13
[PHPStan] Fix PHPStan notice on PHPStan 1.9.3

samsonasik:
**Description**

Fix PHPStan notice on PHPStan 1.9.3:

```bash
Run vendor/bin/phpstan analyse
Note: Using configuration file /home/runner/work/CodeIgniter4/CodeIgniter4/phpstan.neon.dist.
   0/434 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0%
 360/434 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░]  82%
 434/434 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

Error: Variable $cols in isset() always exists and is not nullable.
 ------ -------------------------------------------------------------- 
  Line   system/CLI/CLI.php                                            
 ------ -------------------------------------------------------------- 
  985    Variable $cols in isset() always exists and is not nullable.  
```

Ref https://github.com/codeigniter4/CodeIgniter4/actions/runs/3686745072/jobs/6239400007#step:11:15

**Checklist:**
- [x] Securely signed commits

PR #6991, opened 2022-12-19
fix: empty paths for `rewrite.php`

datamweb:
Screenshot 2022-12-19 063650

**Description**
This PR fixes an issue where an url has an empty path segment and a non-trailing slash ending. 
```
Deprecated: urldecode(): Passing null to parameter #1 ($string) of type string is deprecated in P:\Bug\project-root\vendor\codeigniter4\framework\system\Commands\Server\rewrite.php on line 28
```
Instead, what we'll do is pass an empty string if that happens. Since the url couldn't be parsed, we give an empty string, which is an url that will definitely not exist (root is /) and it will properly show the **404 page**.

**PHP: 8.1.5**
see: http://localhost:8080//any

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Commands/Server/rewrite.php
+++ b/system/Commands/Server/rewrite.php
@@ -25,7 +25,7 @@ if (PHP_SAPI === 'cli') {
 }
 
 $uri = urldecode(
-    parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
+    parse_url('http://ci4.dev' . $_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
 );
 
 // All request handle by index.php file.
--- a/system/Commands/Server/rewrite.php
+++ b/system/Commands/Server/rewrite.php
@@ -24,7 +24,9 @@ if (PHP_SAPI === 'cli') {
     return;
 }
 
-$uri = urldecode(parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH));
+$uri = urldecode(
+    parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
+);
 
 // All request handle by index.php file.
 $_SERVER['SCRIPT_NAME'] = '/index.php';
--- a/system/Commands/Server/rewrite.php
+++ b/system/Commands/Server/rewrite.php
@@ -25,7 +25,7 @@ if (PHP_SAPI === 'cli') {
 }
 
 $uri = urldecode(
-    parse_url('http://ci4.dev' . $_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
+    parse_url('https://codeigniter.com' . $_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
 );
 
 // All request handle by index.php file.


kenjis:
At least, the usage of `parse_url()` is not correct.
> Caution
This function may not give correct results for relative or invalid URLs, and the results may not even match common behavior of HTTP clients.
https://www.php.net/manual/en/function.parse-url.php#refsect1-function.parse-url-notes

How about this?

```diff
--- a/system/Commands/Server/rewrite.php
+++ b/system/Commands/Server/rewrite.php
@@ -24,7 +24,7 @@ if (PHP_SAPI === 'cli') {
     return;
 }
 
-$uri = urldecode(parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH));
+$uri = urldecode(parse_url('http://example.jp'.$_SERVER['REQUEST_URI'], PHP_URL_PATH));
 
 // All request handle by index.php file.
 $_SERVER['SCRIPT_NAME'] = '/index.php';
```

system/Commands/Server/rewrite.php
```diff
@@ -24,7 +24,9 @@
     return;
 }
 
-$uri = urldecode(parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH));
+$uri = urldecode(
+    parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
```
kenjis, line 28: Is this better?
1. we should not pass a relative URL to `parse_url()`.
2. `parse_url()` might return null.

```suggestion
    parse_url('http://ci4.dev' . $_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
```

kenjis (approved):

system/Commands/Server/rewrite.php
```diff
@@ -24,7 +24,9 @@
     return;
 }
 
-$uri = urldecode(parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH));
+$uri = urldecode(
+    parse_url('http://ci4.dev' . $_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
```
samsonasik, line 28: what ci4.dev?

samsonasik:
Rebase latest develop may be needed

ddevsr:
> @samsonasik it related to PR #6975?
> 
> ```
>    0/434 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0%
>  280/434 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░░░░░░]  64%
>  434/434 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
> 
> Error: Variable $cols might not be defined.
>  ------ -------------------------------------- 
>   Line   system/CLI/CLI.php                    
>  ------ -------------------------------------- 
>   986    Variable $cols might not be defined.  
>  ------ -------------------------------------- 
> ```

@datamweb @kenjis @samsonasik I already send PR #6994

system/Commands/Server/rewrite.php
MGatner, line 28: Why not take the free advertising?
```suggestion
    parse_url('https://codeigniter.com' . $_SERVER['REQUEST_URI'], PHP_URL_PATH) ?? ''
```

datamweb: Done. 👍

MGatner (approved):
🎊

PR #6995, opened 2022-12-20
feat: add IncomingRequest::is() method

kenjis:
**Description**
Writing code like `strtolower($this->request->getMethod()) !== 'post'` is a pain.

- add `IncomingRequest::is()`

```php
$request->is('get')
$request->is('post')
$request->is('put')
$request->is('delete')
$request->is('head')
$request->is('patch')
$request->is('options')
$request->is('json')
$request->is('ajax')
```

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -376,14 +376,14 @@ class IncomingRequest extends Request
     }
 
     /**
-     * Checks if this request is.
+     * Checks this request type.
      *
-     * @param string $value HTTP verb or 'json' or 'ajax'
-     * @phpstan-param string|'get'|'post'|'put'|'delete'|'head'|'patch'|'options'|'json'|'ajax' $value
+     * @param string $type HTTP verb or 'json' or 'ajax'
+     * @phpstan-param string|'get'|'post'|'put'|'delete'|'head'|'patch'|'options'|'json'|'ajax' $type
      */
-    public function is(string $value): bool
+    public function is(string $type): bool
     {
-        $valueUpper = strtoupper($value);
+        $valueUpper = strtoupper($type);
 
         $httpMethods = ['GET', 'POST', 'PUT', 'DELETE', 'HEAD', 'PATCH', 'OPTIONS'];
 
@@ -399,7 +399,7 @@ class IncomingRequest extends Request
             return $this->isAJAX();
         }
 
-        throw new InvalidArgumentException('Unknown value: ' . $value);
+        throw new InvalidArgumentException('Unknown type: ' . $type);
     }
 
     /**
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -383,19 +383,19 @@ class IncomingRequest extends Request
      */
     public function is(string $value): bool
     {
-        $value = strtolower($value);
+        $valueUpper = strtoupper($value);
 
-        $httpMethods = ['get', 'post', 'put', 'delete', 'head', 'patch', 'options'];
+        $httpMethods = ['GET', 'POST', 'PUT', 'DELETE', 'HEAD', 'PATCH', 'OPTIONS'];
 
-        if (in_array($value, $httpMethods, true)) {
-            return strtolower($this->getMethod()) === $value;
+        if (in_array($valueUpper, $httpMethods, true)) {
+            return strtoupper($this->getMethod()) === $valueUpper;
         }
 
-        if ($value === 'json') {
+        if ($valueUpper === 'JSON') {
             return strpos($this->getHeaderLine('Content-Type'), 'application/json') !== false;
         }
 
-        if ($value === 'ajax') {
+        if ($valueUpper === 'AJAX') {
             return $this->isAJAX();
         }
 
--- a/system/HTTP/IncomingRequest.php
+++ b/system/HTTP/IncomingRequest.php
@@ -376,6 +376,30 @@ class IncomingRequest extends Request
     }
 
     /**
+     * Checks if this request is.
+     *
+     * @param string $value HTTP verb or 'json'
+     * @phpstan-param 'get'|'post'|'put'|'delete'|'head'|'patch'|'options'|'json' $value
+     */
+    public function is(string $value): bool
+    {
+        $value = strtolower($value);
+
+        $httpMethods = ['get', 'post', 'put', 'delete', 'head', 'patch', 'options'];
+
+        if (in_array($value, $httpMethods, true)) {
+            return strtolower($this->getMethod()) === $value;
+        }
+
+        if ($value === 'json') {
+            return strpos($this->getHeaderLine('Content-Type'), 'application/json') !== false;
+        }
+
+        // @phpstan-ignore-next-line
+        throw new InvalidArgumentException('Unknown value: ' . $value);
+    }
+
+    /**
      * Determines if this request was made from the command line (CLI).
      */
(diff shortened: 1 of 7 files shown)


datamweb (commented):
To be beautiful code.😄

system/HTTP/IncomingRequest.php
```diff
@@ -375,6 +375,29 @@ public function negotiate(string $type, array $supported, bool $strictMatch = fa
...
+    {
+        $value = strtolower($value);
+
+        $httpMethods = ['get', 'post', 'put', 'delete', 'head', 'patch', 'options'];
+
+        if (in_array($value, $httpMethods, true)) {
+            return strtolower($this->getMethod()) === $value;
+        }
+
+        if ($value === 'json') {
```
michalsn, line 394: Not sure if `json` belong here. But if we will have it, why not also have `ajax`? Basically the same story.

iRedds:
1. 
```php
public function isMethod(string $method) : bool {}
public function isGet() : bool {}
public function isPost() : bool {}
// e.t.c.
```
2. the default method name is in uppercase.
The fact that in the framework the method name is in lowercase is the wrong way.

system/HTTP/IncomingRequest.php
kenjis: I'm fine to add `ajax`.

kenjis:
> The fact that in the framework the method name is in lowercase is the wrong way.

What do you mean? Why is lowercase wrong way?

system/HTTP/IncomingRequest.php
kenjis: Added.

iRedds: We already have the isAJAX() method.

MGatner:
I think @iRedds was pointing out that we only need this method because we are inconsistent with our case in the `getMethod()` return. The `$upper` parameter is already deprecated which means we are currently forcing lowercase method names. However if we want to go PSR-7 we cannot modify the case:
```
     * While HTTP method names are typically all uppercase characters, HTTP
     * method names are case-sensitive and thus implementations SHOULD NOT
     * modify the given string.
```

I'm not sure the best approach here. I am not opposed to this helper method, but I also believe it should be unnecessary.

michalsn:
The lowercase/uppercase "problem" was indeed created by PSR-7 compatibility. I don't see a way to resolve it without introducing quite a big BC break because, by default, everything is in lowercase now (and was the same way in CI3).

system/HTTP/IncomingRequest.php
michalsn: Yes, we have, and as you may notice, we use it. IMO proposed `is()` method would serve as a convenient shortcut, so I don't see a problem with another way of accessing the same method.

kenjis:
> Because "the default method name is in uppercase."

> The method token is case-sensitive because it might be used as a gateway to object-based systems with case-sensitive method names. By convention, standardized methods are defined in all-uppercase US-ASCII letters.
https://www.rfc-editor.org/rfc/rfc9110#name-overview

It seems method names should be case-sensitive.

The `getMethod()` returns lowercase now.
So `strtolower($this->getMethod()) === $value;` is redundant.
But the default `$upper = false` in `getMethod()` will be removed.

kenjis:
@iRedds Do you say that the standard HTTP method names are in uppercase?
I changed the implementation a bit.

```console
$ curl -X get https://www.w3.org/ -i
HTTP/2 400 
server: cloudflare
date: Wed, 21 Dec 2022 08:40:47 GMT
content-type: text/html
content-length: 155
cf-ray: -

<html>
<head><title>400 Bad Request</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<hr><center>cloudflare</center>
</body>
</html>
```

MGatner:
I think case-sensitivity is in our favor. We don't need to adjust the case on the method return because `Get` is not `GET`. Forcing these to lower does appear to be a violation of the spec; maybe we add a new method for now?

kenjis:
> The lowercase/uppercase "problem" was indeed created by PSR-7 compatibility. I don't see a way to resolve it without introducing quite a big BC break because, by default, everything is in lowercase now (and was the same way in CI3).

Yes, the current `getMethod()` returns values that are converted to lowercase, and it seems a violation of the HTTP specification. 

The behavior should be fixed but it is a big BC break. So it is better that we wait for v5.0.

@MGatner This `is()` method is the new method to check HTTP methods in case insensitive manner.

michalsn:
> The behavior should be fixed but it is a big BC break. So it is better that we wait for v5.0.

I agree with this.

`is()` may serve as a handful helper method, and since it handles `json` and `ajax` it's not 100% redundant with `getMethod()`.

michalsn (approved):

kenjis:
Rebased and resolved, and all green!

kenjis:
Added docs.

user_guide_src/source/changelogs/v4.3.0.rst
```diff
@@ -323,6 +323,8 @@ Others
 - **Config:** Added ``Config\Session`` class to handle session configuration.
 - **Debug:** Kint has been updated to 5.0.1.
 - **Request:** Added new ``$request->getRawInputVar()`` method to return a specified variable from raw stream. See :ref:`Retrieving Raw data <incomingrequest-retrieving-raw-data>`.
+- **Request:** Added new ``$request->is()`` method to to query the request type.
```
datamweb, line 326:
```suggestion
- **Request:** Added new ``$request->is()`` method to query the request type.
```

MGatner (approved):
I do t feel great about it but I agree with the logic that landed here. Definitely we should fix all this in v5. 

Also, we need to move away from our AJAX methods as they are less and less reliable; maybe in the future we can provide some different terms for the same concept.

michalsn:
> next major version?

@MGatner I'm not sure if this label was meant to be added to this PR, but if so, I would have second thoughts about this PR. v5 means that we already have the `getMethod()` fixed, so `is()` method wouldn't make as much sense to me. 

Then, I would rather add the `isJSON()` method.

MGatner:
@michalsn I've been using that label for PRs or threads I want to come back and review for v5 content. It doesn't necessarily mean "this PR belongs in v5" (though sometimes) but more like "read this thread for a relevant v5 idea".

kenjis:
Rebased and pushed two commits.

MGatner (approved):

kenjis:
> Then, I would rather add the `isJSON()` method.

I found `isJSON()` method. It was deprecated in 4.0.5.
See https://github.com/codeigniter4/CodeIgniter4/pull/7019/files#diff-2cab244a084ce08ac666dff9a88aee13003b4170f6c02a690263e3260dac26d8R35

And it was removed in 4.1.2.
See https://github.com/codeigniter4/CodeIgniter4/commit/55b43083

michalsn:
Well... it's strange to me why we deprecated `isJSON()`... and now we're, in fact, introducing it again. I tried to follow the related discussions, but I still have no point. Anyway, I don't mind it at all. This is a common functionality that will make the life of developers easier. At least, in my opinion.

MGatner:
I think `isJSON()` was removed when we were trying to establish PSR-7 compatibility, but before we split framework-specific methods out into traits. At least that's my best recollection. It's convenient, I have no problem with it, but it is one more framework-coupling method that doesn't save on lines of code or much readability.

PR #6996, opened 2022-12-20
Declared session in BaseController to avoid Creation of dynamic property in php 8.2

thenerdyshark:
Declared session property in the BaseController to avoid the depreciation of dynamic properties in PHP 8.2

### DIFF
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -38,7 +38,7 @@ abstract class BaseController extends Controller
     protected $helpers = [];
 
     /**
-     * Be sure to declare properties for anything you assign.
+     * Be sure to declare properties for any property fetch you initialized.
      * The creation of dynamic property is deprecated in PHP 8.2.
      */
     // protected $session;
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -42,7 +42,7 @@ abstract class BaseController extends Controller
      * initialization in the initController() method
      * See the Creation of dynamic property depreciation i php 8.2 
      */
-    public $session;
+    // public $session;
 
     /**
      * Constructor.
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -40,7 +40,7 @@ abstract class BaseController extends Controller
      /**
      * To declare the session propety to allow for global
      * initialization in the initController() method
-     * @var object
+     * See the Creation of dynamic property depreciation i php 8.2 
      */
     public $session;
 
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -37,12 +37,11 @@ abstract class BaseController extends Controller
      */
     protected $helpers = [];
 
-     /**
+    /**
      * Be sure to declare properties for anything you assign.
      * The creation of dynamic property is deprecated in PHP 8.2.
-     * 
-     * protected $session; 
      */
+    // protected $session;
 
     /**
      * Constructor.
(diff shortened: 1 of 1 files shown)


samsonasik:
I prefer to just write comment instead, as session may or may not be used, session init is also commented under init method.

app/Controllers/BaseController.php
```diff
@@ -37,6 +37,13 @@ abstract class BaseController extends Controller
      */
     protected $helpers = [];
 
+     /**
+     * To declare the session propety to allow for global
+     * initialization in the initController() method
+     * See the Creation of dynamic property depreciation i php 8.2
```
kenjis, line 43:
```suggestion
     * See the Creation of dynamic property depreciation in php 8.2 
```

app/Controllers/BaseController.php
```diff
@@ -37,6 +37,13 @@ abstract class BaseController extends Controller
      */
     protected $helpers = [];
 
+     /**
+     * To declare the session propety to allow for global
```
ddevsr, line 41:
```suggestion
     * To declare the session property to allow for global
```

app/Controllers/BaseController.php
```diff
@@ -37,6 +37,13 @@ abstract class BaseController extends Controller
...
+     * initialization in the initController() method
+     * See the Creation of dynamic property depreciation i php 8.2 
+     */
+    // public $session;
```
samsonasik, line 45: protected modifier is fine since the code is commented

kenjis: It should be `protected`, because we don't need to set it public.

app/Controllers/BaseController.php
ddevsr: By the way, `depreciation` hasn't been added to the list to change

kenjis: What do you mean?
Ah, not `depreciation` but `deprecation`.

MGatner:
I appreciate the thoroughness but I'm fine expecting developers to handle this themselves. A compromise might be expanding the comment below under initController to say something like "be sure to declare properties for anything you assign."

thenerdyshark (commented):
Cleaned up the comment regarding the depreciation of dynamic properties and set the session method to be protected in the BaseController

app/Controllers/BaseController.php
```diff
@@ -37,6 +37,13 @@ abstract class BaseController extends Controller
      */
     protected $helpers = [];
 
+     /**
+     * Be sure to declare properties for anything you assign.
+     * The creation of dynamic property is deprecated in PHP 8.2.
+     * 
+     * protected $session; 
+     */
```
kenjis, line 45: Thank you for clean up.

```suggestion
    /**
     * Be sure to declare properties for anything you assign.
     * The creation of dynamic property is deprecated in PHP 8.2.
     */
    // protected $session;
```

kenjis (approved):

ddevsr (approved):

kenjis:
@thenerdyshark We do not use `git merge`. Please use `git rebase` to update your branch.

app/Controllers/BaseController.php
```diff
@@ -37,6 +37,12 @@ abstract class BaseController extends Controller
      */
     protected $helpers = [];
 
+    /**
+     * Be sure to declare properties for anything you assign.
```
samsonasik, line 41:
```suggestion
     * Be sure to declare properties for any property fetch you initialized in initController method.
```

kenjis: It is not only in `initController()` method, but also in all methods.

samsonasik: I see, updated the suggestion 👍

datamweb (approved):
Thanks!

PR #7017, opened 2022-12-24
docs: update tutorial

kenjis:
**Description**
- refactor the sample code

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/tutorial/create_news_items.rst
+++ b/user_guide_src/source/tutorial/create_news_items.rst
@@ -90,7 +90,8 @@ Most helper functions require the helper to be loaded before use.
 Next, we check if we deal with the **POST** request with the
 :doc:`IncomingRequest <../incoming/incomingrequest>` object ``$this->request``.
 It is set in the controller by the framework.
-If the HTTP method is not POST, that is it is GET,
+The :ref:`IncomingRequest::is() <incomingrequest-is>` method checks the type of the request.
+If the request is not a POST request, that is it is a GET request,
 the form is loaded and returned to display.
 
 Then, we get the necessary items from the POST data by the user and set them in the ``$post`` variable.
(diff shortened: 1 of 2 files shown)


datamweb (approved):

user_guide_src/source/tutorial/create_news_items.rst
```diff
@@ -90,7 +90,8 @@ Most helper functions require the helper to be loaded before use.
 Next, we check if we deal with the **POST** request with the
 :doc:`IncomingRequest <../incoming/incomingrequest>` object ``$this->request``.
 It is set in the controller by the framework.
-If the HTTP method is not POST, that is it is GET,
+The :ref:`IncomingRequest::is() <incomingrequest-is>` method checks the type of the request.
+If the request is not a POST request, that is it is a GET request,
```
michalsn, line 94:
```suggestion
Since the route for **create()** endpoint handles both: **GET** and **POST** requests we can safely assume that if the request is not POST then it is a GET type.
```
Maybe we can add some context to this?

user_guide_src/source/tutorial/create_news_items/002.php
```diff
@@ -24,8 +24,8 @@ public function create()
 
         // Checks whether the submitted data passed the validation rules.
         if (! $this->validateData($post, [
-            'title' => 'required|min_length[3]|max_length[255]',
-            'body'  => 'required|min_length[10]|max_length[5000]',
+            'title' => 'required|max_length[255]|min_length[3]',
+            'body'  => 'required|max_length[5000]|min_length[10]',
```
michalsn, line 28: Out of curiosity. Why is the order changed here? Is this just your preference or something more?

kenjis: Data that is too large should not be processed and should be discarded as soon as possible. This is because it is riskier to process large data.
For example, passing large data to a regular expression engine could result in a DoS attack.

So it is better to check the max_length first.

However, if the data size is checked first, it may not make much difference if the min_length is first.
But, It is obviously not a good idea to check size after another heavy checking, such as a database search.

user_guide_src/source/tutorial/create_news_items.rst
kenjis: Thank you! It is better.

user_guide_src/source/tutorial/create_news_items/002.php
michalsn: Thanks. That makes sense from the perspective of good practice.

michalsn (approved):

PR #3247, opened 2020-07-08
Update html_helper.php

avegacms:
Added hreflang parameter for link_tag function for calling <link href="https://domain.com/en" hreflang="x-default" rel="alternate"> and etc. And added checks for tags with rel expressions for  'alternate','canonical'  in this case attribute type dont use

Each pull request should address a single issue and have a meaningful title.

**Description**
Explain what you have changed, and why.

**Checklist:**
- [ ] Securely signed commits
- [ ] Component(s) with PHPdocs
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [ ] Conforms to style guide

---------Remove from here down in your description----------

**Notes**
- Pull requests must be in English
- If the PR solves an issue, reference it with a suitable verb and the issue number
(e.g. fixes <hash>12345)
- Unsolicited pull requests will be considered, but there is no guarantee of acceptance
- Pull requests should be from a feature branch in the contributor's fork of the repository
  to the develop branch of the project repository

PR #7022, opened 2022-12-25
fix: `link_tag()` missing `type="application/rss+xml"`

kenjis:
**Description**
Follow-up #3247
From https://forum.codeigniter.com/showthread.php?tid=85954

```php
link_tag('feed', 'alternate', 'application/rss+xml', 'My RSS Feed');
```
```
Expected :'<link href="http://example.com/feed" rel="alternate" type="application/rss+xml" title="My RSS Feed" />'
Actual   :'<link href="http://example.com/feed" rel="alternate" title="My RSS Feed" />'
```

`alternate` may use `type` attribute.
See https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/html_helper.php
+++ b/system/Helpers/html_helper.php
@@ -222,13 +222,20 @@ if (! function_exists('link_tag')) {
     /**
      * Link
      *
-     * Generates link to a CSS file
+     * Generates link tag
      *
-     * @param array|string $href      Stylesheet href or an array
-     * @param bool         $indexPage should indexPage be added to the CSS path.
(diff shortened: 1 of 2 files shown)


datamweb (approved):
thanks,
One of the beauties of testing is understanding the subject.

michalsn (approved):

kenjis:
No, this is not easy.
There are so many patterns of combinations that I am not sure of the correct patterns of combinations.

PR #7023, opened 2022-12-25
docs: replace types `mixed` in `CacheInterface` and `BaseHandler` file.

ping-yee:
**Description**
See #6310 
It seems the return types of `increment` and `decrement` functions is int or bool, so I only leave the `bool|int` at that two functions.
And wish you all a merry Christmas!

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Cache/CacheInterface.php
+++ b/system/Cache/CacheInterface.php
@@ -83,7 +83,7 @@ interface CacheInterface
      * The information returned and the structure of the data
      * varies depending on the handler.
      *
-     * @return array|bool|object|null
+     * @return array|false|object|null
      */
     public function getCacheInfo();
 
--- a/system/Cache/CacheInterface.php
+++ b/system/Cache/CacheInterface.php
@@ -83,7 +83,7 @@ interface CacheInterface
      * The information returned and the structure of the data
      * varies depending on the handler.
      *
-     * @return array|bool|float|int|object|string|null
+     * @return array|bool|object|null
      */
     public function getCacheInfo();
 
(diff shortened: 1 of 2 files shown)


system/Cache/CacheInterface.php
```diff
@@ -83,7 +83,7 @@ public function clean();
      * The information returned and the structure of the data
      * varies depending on the handler.
      *
-     * @return mixed
+     * @return array|bool|float|int|object|string|null
```
kenjis, line 86: array or object (Redis) or false or null (DummyHandler)?

ping-yee: Done.

system/Cache/CacheInterface.php
```diff
@@ -83,7 +83,7 @@ public function clean();
      * The information returned and the structure of the data
      * varies depending on the handler.
      *
-     * @return mixed
+     * @return array|bool|object|null
```
kenjis, line 86: Do we need `true`? 
I mean not `bool` but `false`.

kenjis (approved):
LGTM!

system/Cache/CacheInterface.php
ping-yee: Ok, I got it and I have been change `bool` to `false`.
I appreciated it!

kenjis: Thank you for quick fix!

datamweb (approved):

ddevsr (approved):

PR #7029, opened 2022-12-27
fix: Honeypot field appears when CSP is enabled

kenjis:
**Description**
From https://forum.codeigniter.com/showthread.php?tid=85960

- fix bug
- add `Config\Honeypot::$containerId`

### How to Test

```diff
diff --git a/app/Config/App.php b/app/Config/App.php
index 0a23462b6..e8f314371 100644
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -446,5 +446,5 @@ class App extends BaseConfig
      * @see http://www.html5rocks.com/en/tutorials/security/content-security-policy/
      * @see http://www.w3.org/TR/CSP/
      */
-    public bool $CSPEnabled = false;
+    public bool $CSPEnabled = true;
 }
diff --git a/app/Config/Filters.php b/app/Config/Filters.php
index 7b70c4fb3..12bb9619c 100644
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -29,13 +29,13 @@ class Filters extends BaseConfig
      */
     public array $globals = [
         'before' => [
-            // 'honeypot',
+            'honeypot',
             // 'csrf',
             // 'invalidchars',
         ],
         'after' => [
             'toolbar',
-            // 'honeypot',
+            'honeypot',
             // 'secureheaders',
         ],
     ];
diff --git a/app/Controllers/Home.php b/app/Controllers/Home.php
index 7f867e95f..c451f5302 100644
--- a/app/Controllers/Home.php
+++ b/app/Controllers/Home.php
@@ -6,6 +6,6 @@ class Home extends BaseController
 {
     public function index()
     {
-        return view('welcome_message');
+        return '<head></head><form></form>';
     }
 }
```
You will see `Fill This Field` before this PR.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Honeypot.php
+++ b/app/Config/Honeypot.php
@@ -24,7 +24,7 @@ class Honeypot extends BaseConfig
     /**
      * Honeypot HTML Template
      */
-    public string $template = '<label>{label}</label><input type="text" name="{name}" value=""/>';
+    public string $template = '<label>{label}</label><input type="text" name="{name}" value="">';
 
     /**
      * Honeypot container
--- a/app/Config/Honeypot.php
+++ b/app/Config/Honeypot.php
@@ -29,7 +29,7 @@ class Honeypot extends BaseConfig
     /**
      * Honeypot container
      *
-     * If you enables CSP, you can remove `style="display:none"`.
+     * If you enabled CSP, you can remove `style="display:none"`.
      */
     public string $container = '<div style="display:none">{template}</div>';
 
--- a/app/Config/Honeypot.php
+++ b/app/Config/Honeypot.php
@@ -28,6 +28,15 @@ class Honeypot extends BaseConfig
 
     /**
      * Honeypot container
+     *
+     * If you enables CSP, you can remove `style="display:none"`.
      */
     public string $container = '<div style="display:none">{template}</div>';
+
+    /**
+     * The id attribute for Honeypot container tag
+     *
+     * Used when CSP is enabled.
+     */
+    public string $containerId = 'hpc';
 }
--- a/tests/system/Honeypot/HoneypotTest.php
+++ b/tests/system/Honeypot/HoneypotTest.php
@@ -67,13 +67,13 @@ final class HoneypotTest extends CIUnitTestCase
     {
         $this->response->setBody('<form></form>');
         $this->honeypot->attachHoneypot($this->response);
-        $expected = '<form><div style="display:none"><label>Fill This Field</label><input type="text" name="honeypot" value=""/></div></form>';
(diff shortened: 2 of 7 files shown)


tests/system/Honeypot/HoneypotTest.php
```diff
@@ -66,16 +71,35 @@ public function testAttachHoneypotAndContainer()
...
         $this->assertSame($expected, $this->response->getBody());
     }
 
+    public function testAttachHoneypotAndContainerWithCSP()
+    {
+        $this->resetServices();
+
+        $config             = new App();
+        $config->CSPEnabled = true;
+        Factories::injectMock('config', 'App', $config);
```
kenjis, line 90: Because `CodeIgniter\HTTP\ContentSecurityPolicy` depends on `config('App')`,
and `Response` pulls `Services::csp()`, we must inject `Config\App` into the Factories.

app/Config/Honeypot.php
```diff
@@ -24,10 +24,19 @@ class Honeypot extends BaseConfig
     /**
      * Honeypot HTML Template
      */
-    public string $template = '<label>{label}</label><input type="text" name="{name}" value=""/>';
+    public string $template = '<label>{label}</label><input type="text" name="{name}" value="">';
 
     /**
      * Honeypot container
+     *
+     * If you enables CSP, you can remove `style="display:none"`.
```
MGatner, line 32:
```suggestion
     * If you enabled CSP, you can remove `style="display:none"`.
```

tests/system/Honeypot/HoneypotTest.php
MGatner: Thanks for the explanation! We can also just use the factory version here:
```php
config('App')->CSPEnabled = true;
```

But nothing wrong with the explicit version.

MGatner (changes requested):
Looks good! One typo. 

You sent this to 4.3 - is that your vote that we release 4.3 next? I believe we're ready.

michalsn (approved):

kenjis:
> You sent this to 4.3 - is that your vote that we release 4.3 next? I believe we're ready.

I sent to 4.3 because this needs a new Config item.

We merged two bug fixes into develop, so I would like to release v4.2.12 before v4.3.0.

tests/system/Honeypot/HoneypotTest.php
kenjis: Factories are reset each time during the test, so there is probably no problem with that way.
However, I do not think it is healthy for most framework class to depends on Factories (`config()`).

It's not easy to get rid of this dependency now, but I would like to reduce it if possible.

MGatner (approved):

PR #7032, opened 2022-12-28
docs: improve helpers.rst

kenjis:
**Description**
- add note for case sensitive file system and namespace

See https://forum.codeigniter.com/showthread.php?tid=85977

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/general/helpers.rst
+++ b/user_guide_src/source/general/helpers.rst
@@ -147,6 +147,8 @@ functions:
 
 .. literalinclude:: helpers/006.php
 
+.. important:: Do not specify the namespace ``App\Helpers``.
+
 The :php:func:`helper()` function will scan through all PSR-4 namespaces defined in **app/Config/Autoload.php**
 and load in ALL matching helpers of the same name. This allows any module's helpers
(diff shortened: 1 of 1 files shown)


datamweb (approved):
A long time ago, I had used `namespace` to write the helper, which caused the helper not to load.
```php
<?php

declare(strict_types=1);

namespace App\Helpers;

if (! defined('myhelper')) {

    function myhelper() 
    {
        return 'XX';
    }
}
```
Maybe adding a note or warning would be nice. **But I'm not sure if anyone other than me will repeat this funny mistake**.

kenjis:
Oh, it was loaded!

```php
        helper('myhelper');
        \App\Helpers\myhelper(); // returns 'XX'
```

datamweb (approved):
> ```
> \App\Helpers\myhelper(); // returns 'XX'
> ```

I used it like this.
```php
helper('myhelper');
myhelper();
```
And I had this error:
```
Call to undefined function App\Controllers\myhelper() 
```

I understand how to use.
Thank you @kenjis .

michalsn (approved):

PR #7038, opened 2022-12-29
fix: `spark migrate:status` shows incorrect filename when format is `Y_m_d_His_`

kenjis:
**Description**
See https://github.com/codeigniter4/CodeIgniter4/pull/7033#discussion_r1058721974

```console
+-----------+-------------------+-------------------------+-------+-------------+-------+
| Namespace | Version           | Filename                | Group | Migrated On | Batch |
+-----------+-------------------+-------------------------+-------+-------------+-------+
| App       | 2022_12_29_050744 | 12_29_050744_KenjisTest | ---   | ---         | ---   |
| App       | 2022_12_29_050800 | 12_29_050800_PooyaTest  | ---   | ---         | ---   |
+-----------+-------------------+-------------------------+-------+-------------+-------+
```

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #7039, opened 2022-12-29
feat: allow duplicate migration classnames

kenjis:
**Description**
- replace migration classname at runtime

See https://github.com/codeigniter4/CodeIgniter4/pull/7038#issuecomment-1367178671

### How to Test

```console
$ php spark make:migration KenjisTest
$ php spark make:migration KenjisTest
$ php spark migrate
$ php spark migrate:refresh
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/MigrationRunner.php
+++ b/system/Database/MigrationRunner.php
@@ -817,9 +817,8 @@ class MigrationRunner
      */
     protected function migrate($direction, $migration): bool
     {
-        include_once $migration->path;
-
-        $class = $migration->class;
+        $class = $this->generateUniqClassname($migration);
+        $this->loadMigrationClass($migration, $class);
         $this->setName($migration->name);
 
         // Validate the migration file structure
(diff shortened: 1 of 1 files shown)


paulbalandan:
Why not just make an anonymous class migration? I think that would save the trouble of generating unique class names.

kenjis:
If you say something like Laravel's anonymous class migration, it is a new feature, and it does nothing to existing migration files.

We have a bug that we can use duplicate classname if we run migrations one by one.
```
$ php spark make:migrate Abc
$ php spark migrate
$ php spark make:migrate Abc
$ php spark migrate
```
Can anonymous class migration solve the problem?

paulbalandan:
If I recall correctly, request for anonymous class migrations came before Laravel got one.

I think yes it can solve the issue. https://3v4l.org/8gehV
Maybe we can migrate existing migrations to use anonymous classes?

lonnieezell:
While I think anonymous class migrations is a fine idea - wouldn't it make sense to check for a unique file or class name before creating it? We could then ask for a new name if one is found.

iRedds:
1. Implement migrations only through anonymous classes.
2. Add a console command that will force change all existing migration classes to anonymous classes and change the migration table in the database.

Almost a year ago I already voiced this issue. And the option was to use anonymous classes, but I was told that the best solution would be to give more unique class names.

kenjis:
> wouldn't it make sense to check for a unique file or class name before creating it?

It makes sense.
But it does nothing to existing duplicate migration classes.

kenjis:
@iRedds I think they would work fine, but they are really big breaking changes.

kenjis:
I created this PR because
1. I knew the bug that allows duplicate class names (migrate one by one)
2. it seemed relatively easy to solve the all problems without breaking changes

However, you all don't seem to like this very much.

hexageek1337:
By the way, `eval()` will be rejected in more server hosting.

system/Database/MigrationRunner.php
```diff
@@ -817,9 +817,8 @@ public function ensureTable()
      */
     protected function migrate($direction, $migration): bool
     {
-        include_once $migration->path;
-
-        $class = $migration->class;
+        $class = $this->generateUniqClassname($migration);
```
MGatner, line 820: Person preference: I'd rather see this spelled out than save two characters:
```suggestion
        $class = $this->generateUniqueClassname($migration);
```

MGatner (commented):
I want to think on this a bit, and see others' opinions. It seems like quite a hack when the real problem is developers creating duplicate class names (a problem they create for themselves). Anonymous classes would solve this moving forward, but not for existing duplicates.

kenjis:
> the real problem is developers creating duplicate class names (a problem they create for themselves)

Yes, but it might be difficult if there are so many migration files.
So if the generator checks duplicates, it makes sense.

PR #7046, opened 2023-01-03
fix: change `getFile()` function of \CodeIgniter\Events\Events to static.

### LINKED ISSUE #7045: Bug: The getFiles method of \CodeIgniter\Events class is declared as 'public', but it should probably be 'public static'.
### PHP Version

8.2

### CodeIgniter4 Version

4.2.11

Linux

### Which server did you use?

apache

### Database

MySQL

### What happened?

The getFiles method of \CodeIgniter\Events class is declared as 'public', but it should probably be 'public static'?

### Steps to Reproduce

Just try to make this call (which will throw an error since the getFiles is not static, and its an anti-pattern to instantiate the Events class just to be able to call this method.):
\CodeIgniter\Events\Events::getFiles();

### Expected Output

A list of Events.php config file locations.

### Anything else?

We ran into this issue while trying to implement a solution that requires us to get a list of those locations, our system is a web-based configuration for installations were you don't have access to modify the files on the drive.

Our current workaround was just to patch the class to add the static keyword to that method definition and update our installation. Just wondering if there is maybe a valid reason why the getFiles method is not public static and if maybe we should change it to include the static keyword.

ping-yee:
**Description**
Fixes #7045
 
But I have a question is does the `$this->manager` still be needed in the test case of `test\system\Events\EventsTest.php` file?

Because all the functions in Event class are static, does it mean that this class no longer needs to be constructed?

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Events/EventsTest.php
+++ b/tests/system/Events/EventsTest.php
@@ -55,21 +55,21 @@ final class EventsTest extends CIUnitTestCase
 
         // it should start out empty
         MockEvents::setFiles([]);
-        $this->assertEmpty($this->manager->getFiles());
+        $this->assertEmpty(Events::getFiles());
 
         // make sure we have a default events file
         $default = [APPPATH . 'Config' . DIRECTORY_SEPARATOR . 'Events.php'];
         $this->manager->unInitialize();
         MockEvents::initialize();
-        $this->assertSame($default, $this->manager->getFiles());
+        $this->assertSame($default, Events::getFiles());
 
         // but we should be able to change it through the backdoor
         MockEvents::setFiles(['/peanuts']);
-        $this->assertSame(['/peanuts'], $this->manager->getFiles());
+        $this->assertSame(['/peanuts'], Events::getFiles());
 
         // re-initializing should have no effect
         MockEvents::initialize();
-        $this->assertSame(['/peanuts'], $this->manager->getFiles());
+        $this->assertSame(['/peanuts'], Events::getFiles());
     }
 
     public function testPerformance()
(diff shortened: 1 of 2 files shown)


kenjis:
`$this->manager` is no problem.
Because we can instantiate static classes with no problem.

By the way, can you remove `$this->manager` in the tests?

ping-yee:
Ok, I'll handle this.

tests/system/Events/EventsTest.php
```diff
@@ -55,21 +55,21 @@ public function testInitialize()
 
         // it should start out empty
         MockEvents::setFiles([]);
-        $this->assertEmpty($this->manager->getFiles());
+        $this->assertEmpty(MockEvents::getFiles());
```
kenjis, line 58: Please revert.

Testing mocks is a bad practice. It is inherently wrong. 
Because mocks are not real product code.
Mocks are tools for testing. The test target should be the real product code, not the mock.

However, in this test case, there is no other way but to test the mock because the design of Event is not good.

Still, the code to check the mock directly is uncomfortable. 
The `$this->manager` represents the test target, which is better than checking MockEvents directly.

kenjis (changes requested):
See the inline comment.

tests/system/Events/EventsTest.php
ping-yee: Ok, I understand the logic of the testing mock and sorry for this terrible change.
I have been revert this test case change.

kenjis (approved):

tests/system/Events/EventsTest.php
kenjis: Thanks for the fix.

PR #7052, opened 2023-01-05
fix: define of STDOUT in CLI init() method

### LINKED ISSUE #7047: Bug: Constant STDOUT already defined
### PHP Version

8.1

### CodeIgniter4 Version

4.2.11

macOS

### Which server did you use?

cli

### Database

MariaDB 10.9.3

### What happened?

When I trying to call command from Controller with params I'm getting 

```php
ErrorException
Constant STDOUT already defined
```

Here is the screenshot:

CleanShot 2023-01-03 at 13 05 49@2x

### Steps to Reproduce

Controller method includes line:

```php
command('command:command')
```

Command includes line:

```php
$isId = (bool) CLI::getOption('i');
```

where it's failing. 

### Expected Output

Execute command.

### Anything else?

I tried to look at method in `vendor/codeigniter4/framework/system/CLI/CLI.php` and what I found is when I delete `define('STDOUT', 'php://output'); // @codeCoverageIgnore` then it's working as expected.

Do we need to define it in else statement ?  Or I'm calling command wrong. 

```diff
/**
     * Static "constructor".
     */
    public static function init()
    {
        if (is_cli()) {
            // Readline is an extension for PHP that makes interactivity with PHP
            // much more bash-like.
            // http://www.php.net/manual/en/readline.installation.php
            static::$readline_support = extension_loaded('readline');

            // clear segments & options to keep testing clean
            static::$segments = [];
            static::$options  = [];

            // Check our stream resource for color support
            static::$isColored = static::hasColorSupport(STDOUT);

            static::parseCommandLine();

            static::$initialized = true;
        } else {
            // If the command is being called from a controller
            // we need to define STDOUT ourselves
-          define('STDOUT', 'php://output'); // @codeCoverageIgnore
        }
    }
```

jozefrebjak:
**Description**

In development and also in production environment where composer dependencies are installed without `--no-dev` composer flag we can have an issue with running commands from Controller with `command()` helper. 

Issue is there because of **Psalm** required package `amphp/byte-stream` in 

[functions.php](https://github.com/amphp/byte-stream/blob/b5a09a3abd7a638ca4cadcf0876ce0ccd8d911d3/src/functions.php#L15-L17)

Psalm is often installed with [devkit](https://github.com/codeigniter4/devkit), so we need to check if `STDOUT` is already defined.

Fixes #7047

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -161,8 +161,10 @@ class CLI
             static::parseCommandLine();
 
             static::$initialized = true;
-        // See: https://github.com/codeigniter4/CodeIgniter4/issues/7047
         } elseif (! defined('STDOUT')) {
+            // If the command is being called from a controller
+            // we need to define STDOUT ourselves
+            // For "! defined('STDOUT')" see: https://github.com/codeigniter4/CodeIgniter4/issues/7047
             // @codeCoverageIgnoreStart
             define('STDOUT', 'php://output');
             // @codeCoverageIgnoreEnd
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -164,7 +164,9 @@ class CLI
         } else {
             // If the command is being called from a controller
             // we need to define STDOUT ourselves
-            define('STDOUT', 'php://output'); // @codeCoverageIgnore
+            if (! defined('STDOUT')) {
+                define('STDOUT', 'php://output'); // @codeCoverageIgnore
+            }
         }
     }
 
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -150,23 +150,16 @@ class CLI
             // much more bash-like.
             // http://www.php.net/manual/en/readline.installation.php
             static::$readline_support = extension_loaded('readline');
-
             // clear segments & options to keep testing clean
             static::$segments = [];
             static::$options  = [];
-
             // Check our stream resource for color support
             static::$isColored = static::hasColorSupport(STDOUT);
-
             static::parseCommandLine();
-
             static::$initialized = true;
-        } else {
-            // If the command is being called from a controller
-            // we need to define STDOUT ourselves
-            if (! defined('STDOUT')) {
-                define('STDOUT', 'php://output'); // @codeCoverageIgnore
-            }
+        } elseif (! defined('STDOUT')) {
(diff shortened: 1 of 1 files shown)


system/CLI/CLI.php
```diff
@@ -150,21 +150,16 @@ public static function init()
...
             static::parseCommandLine();
-
             static::$initialized = true;
-        } else {
-            // If the command is being called from a controller
-            // we need to define STDOUT ourselves
-            define('STDOUT', 'php://output'); // @codeCoverageIgnore
+        } elseif (! defined('STDOUT')) {
+            define('STDOUT', 'php://output');
+            // @codeCoverageIgnore
```
kenjis, line 162: This code is tricky. Please add a comment why this is needed.

system/CLI/CLI.php
```diff
@@ -150,21 +150,16 @@ public static function init()
             // much more bash-like.
             // http://www.php.net/manual/en/readline.installation.php
             static::$readline_support = extension_loaded('readline');
-
```
kenjis, line 153: Why do you remove this empty line?

jozefrebjak: The rector did it. I don't know why he doesn't like it the way it was.

MGatner (approved):
Thanks for the quick fix!

system/CLI/CLI.php
jozefrebjak: @kenjis reverted rector line changes.

system/CLI/CLI.php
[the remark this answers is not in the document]
jozefrebjak: I changed it to 

```php
// @codeCoverageIgnoreStart
define('STDOUT', 'php://output');
// @codeCoverageIgnoreEnd
```

because php-cs-fixer is trying to add after line comment to the new line. 

Or do we not need the codeCoverageIgnore annotation at all?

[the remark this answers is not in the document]
paulbalandan: put `@codeCoverageIgnore` beside define

define(...); @codeCoverageIgnore

samsonasik (approved):

kenjis (approved):

kenjis:
Okay. Thanks!

PR #7060, opened 2023-01-07
Fixing bug with legacy autoRoute when testing

baycik:
When legacy autoRoute is used in case of using controller that sits in sub directory, that directory is determined and stored in `$this->directory` and new AutoRouter class. But if we subsequently try to load controller without that is not in a directory, then codeigniter can't find it because the `$this->directory` var in AutoRoute instance is still holding reference to previous directory. This only happens when we load several controllers from console. In my case after upgrading to 4.2.11 my tests fail because of it.

I suggest to remove this return; statement to reset directory in AutoRoute instance also.

**Description**

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Router/Router.php
+++ b/system/Router/Router.php
@@ -574,8 +574,6 @@ class Router implements RouterInterface
     {
         if (empty($dir)) {
             $this->directory = null;
-
-            return;
         }
 
         if ($this->autoRouter instanceof AutoRouter) {


kenjis:
I don't understand your use case.
Can you show us minimum sample code?

And why after upgrading to 4.2.11 your tests fail?
It seems there is no change in Router in v4.2.11.

baycik:
Ok I have upgraded my app from CI 4.0.x to 4.2.11. I'm using codeigniters phpunit test system. In test I mimic subsequental requests to app using FeatureTestTrait. After upgrade when I make
1) $this->withSession($_SESSION)->post('/Admin/Tariff/storeTariffAdd',$params);
then 
2) $this->withSession($_SESSION)->post('/User/signIn',$params);
I get this error 
3) CodeIgniter\Exceptions\PageNotFoundException: Controller or its method is not found: \App\Controllers\Admin\User::signIn
I have spent a lot time to trace down this bug and as I think problem is in the return statement. When I delete it test passes normally

kenjis:
Thank you for the explanation. I got the situation.
I agree with removing the `return`.

kenjis (approved):

samsonasik (approved):

samsonasik:
Thank you @baycik

PR #7079, opened 2023-01-10
Fixing bug on stuck content-type header

baycik:
When I use FeatureTestTrait if I use withBodyFormat('json') function on one request all subsequent requests are sent with content-type: application/json even if their body format is not 'json'.

This bug started when I upgraded CI from 4.0.x to 4.2.11. After long investigation I think I found the source of the problem. On line 289 initialization of request has been changed to Services::request($config, true) so for each subsequent test request same instance is used. That's why if content-type on one request is set then all other requests are sent with this header. We need some reset mechanism. I dont know if this method is right but it worked for me.

**Description**

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -364,7 +364,8 @@ trait FeatureTestTrait
 
             return $request;
         }
-
+        
+        $request->removeHeader('Content-Type');
         if (isset($this->bodyFormat) && $this->bodyFormat !== '') {
             if (empty($params)) {
                 $params = $request->fetchGlobal('request');
@@ -381,7 +382,7 @@ trait FeatureTestTrait
                 $request->setHeader('Content-Type', $formatMime);
             }
         }
-
+        
         return $request;
     }
 }


system/Test/FeatureTestTrait.php
```diff
@@ -364,7 +364,8 @@ protected function setRequestBody(Request $request, ?array $params = null): Requ
 
             return $request;
         }
-
+
```
ddevsr, line 367: run ``composer cs-fix``

```suggestion

```

system/Test/FeatureTestTrait.php
```diff
@@ -381,7 +382,7 @@ protected function setRequestBody(Request $request, ?array $params = null): Requ
                 $request->setHeader('Content-Type', $formatMime);
             }
         }
-
+
```
ddevsr, line 385: same as above comment

```suggestion

```

lonnieezell:
I think I better solution is to change that Services to call to NOT use a shared instance:

Services::request($config, false)

That way it grabs a new instance of the request object every time.

baycik:
You know better than me

baycik:
It also works I've checked

PR #7084, opened 2023-01-11
fix: model change `date()` to `Time` class in `intToDate`

### LINKED ISSUE #7082: Bug: [Model] created_at takes timezone of server but updated_at takes timezone of app
### PHP Version

8.2

### CodeIgniter4 Version

4.2.10

Linux

### Which server did you use?

apache

### Database

MariaDB 10.4.26

### What happened?

A strange error is occurring when I save any record in the database using the SAVE() method.
the created_at is taking the timezone of the server and the uptaded_at is the timezone configured for the application.
the time difference is 3 hours (time zones of different countries USA(server) - BRAZIL(app)).
I couldn't identify the problem.

### Steps to Reproduce

create any table with created_at and updated_at
save any record with SAVE() method
server time zone need to be different from your app configuration

### Expected Output

same timezone recorded 

### Anything else?

_No response_

ddevsr:
**Description**
Fixed #7082.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1259,6 +1259,8 @@ abstract class BaseModel
      * @return int|string
      *
      * @throws ModelException
+     *
+     * @deprecated The parameter $userData are deprecated.
      */
     protected function setDate(?int $userData = null)
     {
@@ -1291,10 +1293,10 @@ abstract class BaseModel
                 return $value;
 
             case 'datetime':
-                return (string) Time::createFromTimestamp($value, app_timezone());
+                return date('Y-m-d H:i:s', $value);
 
             case 'date':
-                return (string) Time::createFromTimestamp($value, app_timezone());
+                return date('Y-m-d', $value);
 
             default:
                 throw ModelException::forNoDateFormat(static::class);
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1291,10 +1291,10 @@ abstract class BaseModel
                 return $value;
 
             case 'datetime':
-                return date('Y-m-d H:i:s', $value);
+                return Time::createFromTimestamp($value, app_timezone())->toDateTimeString();
 
             case 'date':
-                return date('Y-m-d', $value);
+                return Time::createFromTimestamp($value, app_timezone())->toDateString();
 
             default:
                 throw ModelException::forNoDateFormat(static::class);
--- a/system/BaseModel.php
+++ b/system/BaseModel.php
@@ -1291,7 +1291,7 @@ abstract class BaseModel
                 return $value;
 
             case 'datetime':
-                return Time::createFromTimestamp($value, app_timezone())->toDateTimeString();
+                return (string) Time::createFromTimestamp($value, app_timezone());
 
             case 'date':
                 return Time::createFromTimestamp($value, app_timezone())->toDateString();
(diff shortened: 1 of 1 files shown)


kenjis:
Can you add test code to reproduce the bug?

system/BaseModel.php
```diff
@@ -1291,10 +1291,10 @@ protected function intToDate(int $value)
                 return $value;
 
             case 'datetime':
-                return date('Y-m-d H:i:s', $value);
+                return Time::createFromTimestamp($value, app_timezone())->toDateTimeString();
```
kenjis, line 1294: `toDateTimeString()` returns localized value. It is not good as data to save in database.
```suggestion
                return (string) Time::createFromTimestamp($value, app_timezone());
```

system/BaseModel.php
```diff
@@ -1291,10 +1291,10 @@ protected function intToDate(int $value)
...
 
             case 'date':
-                return date('Y-m-d', $value);
+                return Time::createFromTimestamp($value, app_timezone())->toDateString();
```
kenjis, line 1297:
```suggestion
                return (string) Time::createFromTimestamp($value, app_timezone());
```

kenjis:
@ddevsr Do you still need this PR?
#7082 was simply a user mistake, not a bug.

system/BaseModel.php
```diff
@@ -1291,10 +1291,10 @@ protected function intToDate(int $value)
                 return $value;
 
             case 'datetime':
-                return date('Y-m-d H:i:s', $value);
+                return (string) Time::createFromTimestamp($value, app_timezone());
```
kenjis, line 1294: The `$value` is the current timestamp from `setDate()`.
It comes from `Time::now()`.
So I don't see a need to call `Time` again.

system/BaseModel.php
```diff
@@ -1291,10 +1291,10 @@ protected function intToDate(int $value)
...
 
             case 'date':
-                return date('Y-m-d', $value);
+                return (string) Time::createFromTimestamp($value, app_timezone());
```
kenjis, line 1297: Sorry, this is my mistake, but they are not the same. We must return as `'Y-m-d'`.

kenjis:
By the way, the parameter `$userData` is not passed at all?
https://github.com/codeigniter4/CodeIgniter4/blob/c6ad1de5d501796d8ea91f134685a445f6aa8c56/system/BaseModel.php#L1263

ddevsr:
> By the way, the parameter `$userData` is not passed at all?

Yes, i see `$userData` no needed in Model or BaseModel. What do you think to remove it and use `setDate()` directly `return Time`

kenjis:
> What do you think to remove it and use setDate() directly return Time

We cannot remove it. Because it is a breaking change.
But we can make it deprecated.

I think the parameter is not needed.

system/BaseModel.php
```diff
@@ -1259,6 +1259,8 @@ protected function doProtectFields(array $data): array
      * @return int|string
      *
      * @throws ModelException
+     *
+     * @deprecated The parameter $userData are deprecated.
```
kenjis, line 1263: Why deprecated?
If we remove it, we cannot set arbitrary date.

kenjis:
This PR contains no code change.
Close.

PR #7085, opened 2023-01-11
fix: TypeError in Validation is_unique/is_not_unique

kenjis:
**Description**
See https://forum.codeigniter.com/showthread.php?tid=86107

- fix TypeError in `Rules::is_unique()` and `is_not_unique()` in Strict Rules
- refactor tests
  - move the test logic to `StrictRules\DatabaseRelatedRulesTest` because Strict Rules are now default

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/Rules.php
+++ b/system/Validation/Rules.php
@@ -86,7 +86,11 @@ class Rules
     public function is_not_unique(?string $str, string $field, array $data): bool
     {
         // Grab any data for exclusion of a single row.
-        [$field, $whereField, $whereValue] = array_pad(explode(',', $field), 3, null);
+        [$field, $whereField, $whereValue] = array_pad(
+            explode(',', $field),
+            3,
+            null
+        );
 
         // Break the table and field apart
         sscanf($field, '%[^.].%[^.]', $table, $field);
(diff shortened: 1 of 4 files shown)


michalsn (approved):

ddevsr (approved):

jlopes90:
I already tested it and it works.

system/Validation/Rules.php
```diff
@@ -86,7 +86,11 @@ public function greater_than_equal_to(?string $str, string $min): bool
     public function is_not_unique(?string $str, string $field, array $data): bool
```
jlopes90, line 86: I was trying something else, I removed "?string" and it worked.

`public function is_not_unique($str, string $field, array $data): bool`

jlopes90:
I'm using "codeIgniter 4 app starter", how to get this to update or change this branch to vendor/codeigniter4/framework or something?

I tested was to copy the files and replace.

kenjis:
@jlopes90 Thank you for testing.

See https://github.com/codeigniter4/shield/pull/195#issuecomment-1374456938
This is an example in Shield, but all you need to do is
1. add the repository URL in `repositories`
2. change the version to the branch (`dev-fix-Validtaion-is_unique` in this case)

PR #7090, opened 2023-01-11
docs: improve route_to() and url_to()

kenjis:
**Description**
- update PHPDoc
- update User Guide
- fix sample code
- add test

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -542,14 +542,15 @@ if (! function_exists('mb_url_title')) {
 
 if (! function_exists('url_to')) {
     /**
-     * Get the full, absolute URL to a controller method
+     * Get the full, absolute URL to a route named or controller method
      * (with additional arguments)
      *
      * NOTE: This requires the controller/method to
      * have a route defined in the routes Config file.
      *
-     * @param string     $controller Named route or Controller::method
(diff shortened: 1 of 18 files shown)


system/Helpers/url_helper.php
```diff
@@ -544,14 +544,15 @@ function mb_url_title(string $str, string $separator = '-', bool $lowercase = fa
 
 if (! function_exists('url_to')) {
     /**
-     * Get the full, absolute URL to a controller method
+     * Get the full, absolute URL to a route named or controller method
```
datamweb, line 547:
```suggestion
     * Get the full, absolute URL to a route name or controller method
```

kenjis:
By the way, do we need `route_to()`?
I don't see there is a particular use case.

lonnieezell:
route_to works with named routes and controller/methods instead of urls so yes I think it still needs to stay.

kenjis:
Rebased and add a commit.

kenjis:
Rebased and pushed 4 commits.

kenjis:
Please review.

kenjis:
Can anyone review this?

samsonasik (approved):

kenjis:
Thanks! @samsonasik

PR #7093, opened 2023-01-11
fix: remove `All` from `Options All -Indexes` in .htaccess

sba:
Options All -Indexes causes a 500 Internal Server Error. (Apache on hostpoint.ch / interxion).

Without the -All it works fine.

I'm not sure about all the effects, but I think the -All is already handled in the Apache directive and has not to be in .htaccess

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -1,5 +1,5 @@
 # Disable directory browsing
-Options All -Indexes
+Options -Indexes
 
 # ----------------------------------------------------------------------
 # Rewrite engine


paulbalandan:
What specifically is the Error 500 message?

kenjis:
It seems `Options -Indexes` depends on the context.
```
For example, without any + and - symbols:

<Directory "/web/docs">
  Options Indexes FollowSymLinks
</Directory>

<Directory "/web/docs/spec">
  Options Includes
</Directory>

then only Includes will be set for the /web/docs/spec directory. However if the second Options directive uses the + and - symbols:

<Directory "/web/docs">
  Options Indexes FollowSymLinks
</Directory>

<Directory "/web/docs/spec">
  Options +Includes -Indexes
</Directory>

then the options FollowSymLinks and Includes are set for the /web/docs/spec directory.
```
https://httpd.apache.org/docs/2.4/en/mod/core.html#options

sba:
> What specifically is the Error 500 message?

[Wed Jan 11 17:44:25.050920 2023] [core:alert] [pid 90226:tid 34394338560] [client 167.94.138.47:55274] /home/lindehof/www/rasen.lindehof.ch/public/.htaccess: Option All not allowed here

paulbalandan:
Can you check if your apache configuration file allows override of Options in the htaccess file?
It should be either:
> AllowOverride All

or 
> AllowOverride Options

sba:
> Can you check if your apache configuration file allows override of Options in the htaccess file?

Both are not allowed:

`[Thu Jan 12 08:56:11.083171 2023] [core:alert] [pid 34223:tid 34394500352] [client 2a02:1210:3615:e400:c2e5:e96a:b60c:9f58:57092] /home/lindehof/www/rasen.lindehof.ch/public/.htaccess: AllowOverride not allowed here
[Thu Jan 12 08:56:42.052244 2023] [core:alert] [pid 34121:tid 34394657792] [client 2a02:1210:3615:e400:c2e5:e96a:b60c:9f58:57098] /home/lindehof/www/rasen.lindehof.ch/public/.htaccess: AllowOverride not allowed here`

paulbalandan:
No, I mean your actual configuration file `httpd
conf` contains the AllowOverride lines.

ddevsr:
I share my config `apache2.conf`

```
<Directory /home/http7_data>
        Options -Indexes
        AllowOverride All
        Order allow,deny
        allow from all
        Require all granted
</Directory>
```

With this my `.htaccess` running normal now, Can you sharing `apache2.conf` or log file?

sba:
i can't see my `apache2.conf` or other `http conf` files, it's on a shared hosting and unfortunately I don't have access to it...

kenjis:
The comment says "Disable directory browsing".
Then, this PR is no problem.
It seems we do not need to set `All` options.

paulbalandan (approved):

kenjis (approved):

paulbalandan:
> i can't see my `apache2.conf` or other `http conf` files, it's on a shared hosting and unfortunately I don't have access to it...

In that case, we can assume your hosting provider does not allow override of all options. Anyway, this change seems harmless.

kenjis:
@sba Thank you!

PR #7101, opened 2023-01-12
docs: add about Composer error in upgrade_430.rst

kenjis:
**Description**
- add about "Call to undefined method ComposerInstalledVersions::getAllRawData()"

See
- https://forum.codeigniter.com/showthread.php?tid=86096&pid=405146#pid405146
- https://forum.codeigniter.com/showthread.php?tid=86096&pid=405169#pid405169

**Checklist:**
- [x] Securely signed commits
- [] Component(s) with PHPDoc blocks, only if necessary or adds value
- [] Unit testing, with >80% coverage
- [x] User guide updated
- [] Conforms to style guide

PR #7102, opened 2023-01-12
docs: add missing backslashes

kenjis:
**Description**
Follow-up #7101

- we need to escape `\` withh `\`. See https://codeigniter4.github.io/CodeIgniter4/installation/upgrade_430.html#call-to-undefined-method-composerinstalledversions-getallrawdata

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


datamweb (approved):
More report from users.

PR #7108, opened 2023-01-12
docs: update errors_list template

sba:
check for empty $errors

now with the same content as 
https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/models/model/033.php

or is it better if we include the code-snippet here?
`.. literalinclude:: model/033.php`

### DIFF
--- a/user_guide_src/source/libraries/validation/030.php
+++ b/user_guide_src/source/libraries/validation/030.php
@@ -1,7 +1,7 @@
 <?php if (! empty($errors)): ?>
     <div class="alert alert-danger">
     <?php foreach ($errors as $field => $error): ?>
-        <p><?= $error ?></p>
+        <p><?= esc($error) ?></p>
     <?php endforeach ?>
     </div>
 <?php endif ?>
--- a/user_guide_src/source/libraries/validation/030.php
+++ b/user_guide_src/source/libraries/validation/030.php
@@ -1,7 +1,7 @@
-<div class="alert alert-danger" role="alert">
-    <ul>
-    <?php foreach ($errors as $error): ?>
-        <li><?= esc($error) ?></li>
+<?php if (! empty($errors)): ?>
+    <div class="alert alert-danger">
+    <?php foreach ($errors as $field => $error): ?>
+        <p><?= $error ?></p>
     <?php endforeach ?>
-    </ul>
-</div>
+    </div>
+<?php endif ?>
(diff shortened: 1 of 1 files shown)


user_guide_src/source/libraries/validation/030.php
```diff
@@ -1,7 +1,7 @@
-<div class="alert alert-danger" role="alert">
-    <ul>
-    <?php foreach ($errors as $error): ?>
-        <li><?= esc($error) ?></li>
+<?php if (! empty($errors)): ?>
+    <div class="alert alert-danger">
+    <?php foreach ($errors as $field => $error): ?>
+        <p><?= $error ?></p>
```
kenjis, line 4: Why do you remove `esc()`?
Before: `esc($error)`
After: `$error`

sba: have not paid attention to it. Should of course contain `esc()`

user_guide_src/source/libraries/validation/030.php
```diff
@@ -1,7 +1,7 @@
-<div class="alert alert-danger" role="alert">
-    <ul>
-    <?php foreach ($errors as $error): ?>
-        <li><?= esc($error) ?></li>
+<?php if (! empty($errors)): ?>
+    <div class="alert alert-danger">
+    <?php foreach ($errors as $field => $error): ?>
+        <p><?= esc($error) ?></p>
```
kenjis, line 4: By the way, why do you want to change to `<p>` instead of `<li>`?
Or why do you want it to be the same code as Model's?

user_guide_src/source/libraries/validation/030.php
```diff
@@ -1,7 +1,7 @@
-<div class="alert alert-danger" role="alert">
-    <ul>
-    <?php foreach ($errors as $error): ?>
-        <li><?= esc($error) ?></li>
+<?php if (! empty($errors)): ?>
```
kenjis, line 1: I think adding this check is better.

user_guide_src/source/libraries/validation/030.php
sba: > Or why do you want it to be the same code as Model's?

for consistency, it's actually the same twice.
But you can leave it as is, with ul li. Will just put `if (! empty($errors))` around it

kenjis: I don't see that we need consistency in this case.

This section is for *Customizing* Error Display.
https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html#customizing-error-display

The sample in Model is for In-Model Validation.
https://codeigniter4.github.io/CodeIgniter4/models/model.html#in-model-validation

kenjis:
> or is it better if we include the code-snippet here?
`.. literalinclude:: model/033.php`

No, that is not good. They are different knowledge.

kenjis (approved):

paulbalandan (approved):

PR #7112, opened 2023-01-13
fix: bug on stuck content-type header in Feature Testing

baycik:
Supersedes #7079

When I use FeatureTestTrait if I use `withBodyFormat('json')` function on one request all subsequent requests are sent with `content-type: application/json` even if their body format is not 'json'.

This bug started when I upgraded CI from 4.0.x to 4.2.11. After long investigation I think I found the source of the problem. On line 289 initialization of request has been changed to `Services::request($config, true)` so for each subsequent test request same instance is used. That's why if content-type on one request is set then all other requests are sent with this header. We need some reset mechanism.

**Description**

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -364,7 +364,7 @@ trait FeatureTestTrait
 
             return $request;
         }
-        
+
         $request->removeHeader('Content-Type');
         if (isset($this->bodyFormat) && $this->bodyFormat !== '') {
             if (empty($params)) {
@@ -382,7 +382,7 @@ trait FeatureTestTrait
                 $request->setHeader('Content-Type', $formatMime);
             }
         }
-        
+
         return $request;
     }
 }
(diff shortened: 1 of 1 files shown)


kenjis:
@baycik Thank you for sending PR!
This seems good.

Can you squash the commits? It is only one line change.

baycik:
Yes it contains unnecesarry changes becase initially I thought to comment return statement to reset content-type header before every request;
Unfortunately I can't find how to do the squishing on github. So if this PR is not suitable to you make the change your self to fix the bug.

kenjis:
> Unfortunately I can't find how to do the squishing on github.

Just squash the commits on your local PC, and force push to GitHub.
But if you can't do it, no problem.

kenjis (approved):

kenjis:
@baycik Thank you!

PR #7137, opened 2023-01-17
docs: use Request::is() instead of getMethod()

kenjis:
**Description**
Follow-up #6995

- use Request::is() instead of getMethod()

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/libraries/security.rst
+++ b/user_guide_src/source/libraries/security.rst
@@ -44,9 +44,13 @@ Do one of the following:
 
 E.g.::
 
-    if (strtolower($this->request->getMethod()) !== 'post') {
+    if (! $this->request->is('post')) {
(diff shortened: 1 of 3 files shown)


lonnieezell (approved):
I guess I missed when that method was created. That's an unfortunate name since I think it's checking the URI, not the method, but since that's what the method is called, these changes look fine.

kenjis:
`Request::is()` checks the type of a request.
https://codeigniter.com/user_guide/incoming/incomingrequest.html#is

It is already overridden by CodeIgniter HTMX.
https://michalsn.github.io/codeigniter-htmx/incoming_request/#is

lonnieezell:
I wasn't suggesting we change it now. Just bummed I missed that discussion. No worries. :)

samsonasik:
Thank you @kenjis

PR #7142, opened 2023-01-17
fix: [Parser] `!` does not work if delimiters are changed

kenjis:
**Description**
From https://forum.codeigniter.com/showthread.php?tid=86155

- fix `!` does not work if delimiters are changed

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


MGatner (approved):
I'm not gonna try to grok that regex, code & test looks good!

kenjis:
@MGatner Thank you for reviewing!
Few members seem interested in Parser.

PR #7144, opened 2023-01-17
add `application/vnd.microsoft.portable-executable` and `application/x-dosexec`

totoprayogo1916:
**Description**
When using File Libraries,
`$file->getMimeType();` returns `null` in files with extension `.exe`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/app/Config/Mimes.php
+++ b/app/Config/Mimes.php
@@ -53,6 +53,8 @@ class Mimes
         'lzh' => 'application/octet-stream',
         'exe' => [
             'application/octet-stream',
+            'application/vnd.microsoft.portable-executable',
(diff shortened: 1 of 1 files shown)


paulbalandan:
> https://www.iana.org/assignments/media-types/application/vnd.microsoft.portable-executable
 
The first MIME type is IANA recognized. However, the second one I cannot find. Is it acceptable?

MGatner (approved):
I'm fine with both. MIMEs are for developers, someone could always change it in their project.

samsonasik:
Thank you @totoprayogo1916

Issue #7117, opened 2023-01-14
Bug: [QueryBuilder] `where()` with `BETWEEN` creates bad SQL even with `$escape = false`


tangix:
### PHP Version

8.1

### CodeIgniter4 Version

4.3.0

### CodeIgniter4 Installation Method

Composer (using `codeigniter4/appstarter`)

### Which operating systems have you tested for this bug?

macOS

### Which server did you use?

apache

### Database

MySQL 8.0.31

### What happened?

Upgrading from 4.2.11 to 4.3.0 I notice changes in the QueryBuilder´s escape logic affecting `BETWEEN` statements.

### Steps to Reproduce

This code from 4.2.11 works as expected:
```php
            $db      = db_connect(Table::READONLY);
            $between = 'BETWEEN ' . $db->escape($from) . ' AND ' . $db->escape($to);

            $completed = $db
                ->table(Table::subsredeem)
                ->selectCount('*', 'cnt')
                ->where('created_on IS NOT NULL', '', false)
                ->where('created_on ' . $between, '', false)    // <--- This is the interesting line!
                ->where('completed', 1)
                ->get()
                ->getRow(0)?->cnt;
```
After upgrading to 4.3.0 the "interesting line" produces bad SQL causing an Exception to be thrown.
The QueryBuilder adds a `=`:

```sql
$this->getCompiledSelect(false)
SELECT COUNT(*) AS `cnt`
FROM `subsredeem`
WHERE created_on IS NOT NULL 
AND created_on BETWEEN '2022-07-01 00:00:00' AND '2022-12-31 23:59:59' = 
AND `completed` = 1
```

### Expected Output

Working SQL query.

### Anything else?

The work-around I'm currently using:
```php
 $completed = $db
                ->table(Table::subsredeem)
                ->selectCount('*', 'cnt')
                ->where('created_on IS NOT NULL', '', false)
                ->where(new RawSql('created_on ' . $between)) // <--- Use RawSql here!
                ->where('completed', 1)
                ->get()
                ->getRow(0)?->cnt;
```

sclubricants:
Your problem seems to be that you are using an empty string for the second parameter of `where()`. 

Should be `$builder->where('created_on ' . $between, null, false)`

The default value of the second parameter is null, not an empty string. This is because you may want `column = ''`

If this was working in 4.3.x it was probably by mistake.

kenjis:
Using `''`:
```php
[...]

PR #7151, opened 2023-01-18
docs: add about QueryBuilder::where() change in v4.3.0

kenjis:
**Description**
Add a breaking change reported in #7117

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -332,10 +332,6 @@ methods:
 
         .. literalinclude:: query_builder/027.php
 
-    When you set the third parameter ``$escape`` to ``false``, set ``null`` as the second parameter:
-
-    .. literalinclude:: query_builder/120.php
-
 .. _query-builder-where-rawsql:
 
 5. RawSql
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -275,6 +275,8 @@ Since v4.2.0, ``$builder->join()`` accepts a ``CodeIgniter\Database\RawSql`` ins
 Looking for Specific Data
 *************************
 
+.. _query-builder-where:
+
 Where
 =====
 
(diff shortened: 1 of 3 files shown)


tangix:
Thanks!

kenjis:
I removed the explanation in the Query Builder page.
Because I think the sample code is not good.

It had `=`, so it can be set as `where($key, $value)`.
And I think passing a custom string to `where()` is not good practice,
and should be avoided as possible.

And I prefer `RawSql()` instead of `$escape = false`.
Because `$escape = false` might break the string.
`RawSql()` is mutable and never be changed.
`RawSql()` is easier to find out with IDE.

In conclusion, I do not want to recommend the use of custom string (and escape false).

sclubricants:
I kind of see this explanation as unnecessary because there are uses for second parameter = ''

```php
$where = "REPLACE(column, 'somestring', '')";
$builder->where($where, '', false); 
```

Gives:

```SQL
WHERE REPLACE(column, 'somestring', '') = ''
```

At least this is how it should have been working even if it wasn't.  The documentation has always been clear that the default value of the second parameter is null. Naturally a '' should give a value of an empty string.

kenjis:
v4.2.12:
```php
$where = "REPLACE(column, 'somestring', '')";
$builder->where($where, '', false); 
// 'SELECT * FROM "jobs" WHERE REPLACE(column, 'somestring', '') = '
```
The generated SQL string is invalid. In this case ```` does not work in v4.2.12.

kenjis:
@sclubricants 
```php
$where = "REPLACE(column, 'somestring', '')";
$builder->where($where, '', false); 
// 'SELECT * FROM "jobs" WHERE REPLACE(column, 'somestring', '') = '
```
The code does not work in 4.3 and 4.2 (probably and before). 
Because **the value `''` will be just empty string when escape is false**.

kenjis:
@sclubricants I think if we use `$escape = false`, the code should be:
```php
        $where = "REPLACE(column, 'somestring', '')";
        $builder->where($where, "''", false);
````

**Comment by sclubricants** [comment]
Sorry it should be escape = true

```php
        $where = "REPLACE(column, 'somestring', '')";
        $builder->where($where, "''", true);
```

If false and $value = '' then we should probably throw an error.

kenjis:
> If $escape = false and $value = '' then we should probably throw an error.

It may be. But I don't think it is worth doing it.
Misused code that depends on the behavior where the value `''` disappears when `$escape` is false will not work.

At least the issue with BETWEEN will be *fixed* by #7155

MGatner (approved):

kenjis:
The sample code in this PR is wrong.
It does work in v4.3.

The code that does not work in v4.3 needs `BETWEEN` or `EXISTS` in where() string like #7117.
But it will be *fixed* by #7155
So I close this.

PR #7155, opened 2023-01-20
fix: [QueryBuilder] getOperatorFromWhereKey() misses EXISTS, BETWEEN

kenjis:
**Description**
Follow-up #6986
Ref #7151, #7117

- fix `BaseBuilder::getOperatorFromWhereKey()` misses `EXISTS` or `BETWEEN`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -3381,16 +3381,16 @@ class BaseBuilder
                 : '';
             $this->pregOperators = [
                 '\s*(?:<|>|!)?=\s*', // =, <=, >=, !=
-                '\s*<>?\s*', // <, <>
-                '\s*>\s*', // >
-                '\s+IS NULL', // IS NULL
-                '\s+IS NOT NULL', // IS NOT NULL
-                '\s+EXISTS\s*\(.*\)', // EXISTS(sql)
+                '\s*<>?\s*',         // <, <>
+                '\s*>\s*',           // >å
+                '\s+IS NULL',             // IS NULL
+                '\s+IS NOT NULL',         // IS NOT NULL
+                '\s+EXISTS\s*\(.*\)',     // EXISTS (sql)
                 '\s+NOT EXISTS\s*\(.*\)', // NOT EXISTS(sql)
-                '\s+BETWEEN\s+', // BETWEEN value AND value
-                '\s+IN\s*\(.*\)', // IN(list)
-                '\s+NOT IN\s*\(.*\)', // NOT IN (list)
-                '\s+LIKE\s+\S.*(' . $_les . ')?', // LIKE 'expr'[ ESCAPE '%s']
+                '\s+BETWEEN\s+',          // BETWEEN value AND value
+                '\s+IN\s*\(.*\)',         // IN (list)
+                '\s+NOT IN\s*\(.*\)',     // NOT IN (list)
+                '\s+LIKE\s+\S.*(' . $_les . ')?',     // LIKE 'expr'[ ESCAPE '%s']
                 '\s+NOT LIKE\s+\S.*(' . $_les . ')?', // NOT LIKE 'expr'[ ESCAPE '%s']
             ];
         }
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -3382,7 +3382,7 @@ class BaseBuilder
             $this->pregOperators = [
                 '\s*(?:<|>|!)?=\s*', // =, <=, >=, !=
                 '\s*<>?\s*',         // <, <>
-                '\s*>\s*',           // >å
+                '\s*>\s*',           // >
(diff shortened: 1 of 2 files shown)


kenjis:
With this PR, the following test passes:
```php
$builder = $this->db->table('jobs');

$where = "created_on BETWEEN '2022-07-01' AND '2022-12-31'";
$builder->where($where, '', false); // empty string

$expectedSQL = "SELECT * FROM \"jobs\" WHERE created_on BETWEEN '2022-07-01' AND '2022-12-31' ";
$this->assertSame($expectedSQL, str_replace("\n", ' ', $builder->getCompiledSelect()));

$expectedBinds = [
    "created_on BETWEEN '2022-07-01' AND '2022-12-31'" => [
        0 => '',
        1 => false,
    ],
];
$this->assertSame($expectedBinds, $builder->getBinds());
```
But I still think this is a misuse, because the binding is completely wrong, and the output SQL has an extra space in the end.

MGatner:
@codeigniter4/database-team please take a look. @kenjis Let me know if I need to do this one.

system/Database/BaseBuilder.php
```diff
@@ -3373,16 +3381,16 @@ protected function getOperator(string $str, bool $list = false)
                 : '';
             $this->pregOperators = [
                 '\s*(?:<|>|!)?=\s*', // =, <=, >=, !=
-                '\s*<>?\s*', // <, <>
-                '\s*>\s*', // >
-                '\s+IS NULL', // IS NULL
-                '\s+IS NOT NULL', // IS NOT NULL
-                '\s+EXISTS\s*\(.*\)', // EXISTS(sql)
+                '\s*<>?\s*',         // <, <>
+                '\s*>\s*',           // >å
```
kenjis, line 3385:
```suggestion
                '\s*>\s*',           // >
```

sclubricants:
I don't think we should support BETWEEN, EXISTS, IS NULL, IS NOT NULL, IS TRUE, IS FALSE, etc. I'm not sure that these work the same across DBMS.

If your using these you should just do it with RawSql or with escape. 

Another example:

```SQL
SELECT 'something' AS test
WHERE IF(RAND()>=0.5,TRUE,FALSE)
```
 
This WHERE statement is great for randomly selecting rows.

kenjis:
> I don't think we should support BETWEEN, EXISTS, IS NULL, IS NOT NULL, IS TRUE, IS FALSE, etc. I'm not sure that these work the same across DBMS.

`IS NULL`, `IS NOT NULL` are supported already.

If we will support BETWEEN, EXISTS, the change is only to add ` =` after the key value or not.

kenjis:
> Another example:

What do you mean?
The following test passes in both before and after this PR:
```php
        $builder = $this->db->table('jobs');

        $where = "IF(RAND()>=0.5,TRUE,FALSE)";
        $builder->where($where, null, false);

        $expectedSQL   = 'SELECT * FROM "jobs" WHERE IF(RAND()>=0.5,TRUE,FALSE)';
        $this->assertSame($expectedSQL, str_replace("\n", ' ', $builder->getCompiledSelect()));

        $expectedBinds = [];
        $this->assertSame($expectedBinds, $builder->getBinds());
```

sclubricants:
Yes this is correct:

```php
        $where = "IF(RAND()>=0.5,TRUE,FALSE)";
        $builder->where($where, null, false);
```

I was trying to say we shouldn't support things like:

```php
        $builder->where('column', 'IS NULL');
```

kenjis:
The following is not supported before or after this PR:
```php
$builder->where('column', 'IS NULL');
// 'SELECT * FROM "users" WHERE "column" = 'IS NULL''
```

The following is also not supported.

```php
$builder->where('column', 'IS NULL', false);
// 'SELECT * FROM "users" WHERE column = IS NULL'
```

kenjis:
@codeigniter4/database-team Can someone review?

michalsn (approved):
I'm not a fan of allowing:
```
$builder->where($where, '', false);
```
But since developers were using it this way, we shouldn't break their code.

sclubricants (approved):
I'm not sure why we need all of these spelled out.

There are a lot more of these.

SOUNDS LIKE
RLIKE
NOT RLIKE
IS TRUE
IS FALSE
FIND_IN_SET

kenjis:
Probably no user used these.

In my opinion it is not a good practice to write SQL strings in the key of `where()`. 
If a user want to write SQL strings, it is better to use `RawSql`.
So I don't think it is necessary to add anything other than what are currently defined.

kenjis:
@tangix With this PR, your issue #7117 will be *fixed* in v4.3.2.
See https://github.com/codeigniter4/CodeIgniter4/pull/7155#issuecomment-1399087871

PR #7159, opened 2023-01-21
feat: Filter Arguments with $filters in Config\Filters

kenjis:
**Description**
- add Filter Arguments with `$filters` in `Config\Filters`
  - the route filter's arguments take precedence.
  - If the arguments are already defined, throws an exception.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


kenjis:
If both the Route filter and Config filter are applied, should the Route filter's arguments take precedence?

Or, when multiple Config filters are applied, only one is actually executed, what is the arguments in that case?

system/Filters/Filters.php
```diff
@@ -463,20 +493,59 @@ protected function processFilters(?string $uri = null)
...
+     * @param string $name      filter alias
+     * @param array  $arguments filter arguments
+     * @param bool   $check     if true, check if already defined
+     */
+    private function registerArguments(string $name, array $arguments, bool $check = true): void
+    {
+        if ($arguments !== []) {
+            if ($check && array_key_exists($name, $this->arguments)) {
+                throw new ConfigException(
+                    '"' . $name . '" has already arguments: '
```
MGatner, line 534:
```suggestion
                    '"' . $name . '" already has arguments: '
```

system/Filters/Filters.php
```diff
@@ -463,20 +493,59 @@ protected function processFilters(?string $uri = null)
...
+                    . (($this->arguments[$name] === null) ? 'null' : implode(',', $this->arguments[$name]))
```
MGatner, line 535: How can this get a `null` argument?

MGatner (changes requested):
Awesome! Long-requested feature.

system/Filters/Filters.php
kenjis: 1. Set route filter with no argument.
2. Set config filter with arguments.

Added a test case: https://github.com/codeigniter4/CodeIgniter4/pull/7159/files#diff-b7622ae6db0403f01050df526b06419d30ca51b85ad2220da78fec16aaa9f3c8R285

datamweb (approved):
Very good.
I like this PR 💯 .

MGatner (approved):
LGTM!

PR #7160, opened 2023-01-21
Fix: Throttler token time calculation

rumpfc:
**Description**
In one of my applications where I use `Throttler` I detected the `getTokenTime()` is already at the minimum of 1 seconds even if the token refresh rate is far from reached:

Example: 1 capacity, 3600 seconds -> After 10 minutes I expected 50 minutes to be left, but I was already at "1 second". Waiting another minute and executing the check, I was still at "1 second left".

This made me clear that the calculation for `Throttler::getTokenTime()` is still off, but the token value is calculated correctly after all, return `false` if `$token < 1`. This PR should fix that.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Throttle/Throttler.php
+++ b/system/Throttle/Throttler.php
@@ -139,7 +139,7 @@ class Throttler implements ThrottlerInterface
         // How many seconds till a new token is available.
         // We must have a minimum wait of 1 second for a new token.
         // Primarily stored to allow devs to report back to users.
-        $newTokenAvailable = (int) ($refresh - $elapsed - $refresh * $tokens);
+        $newTokenAvailable = (int) round((1 - $tokens) * $refresh);
         $this->tokenTime   = max(1, $newTokenAvailable);
 
         return false;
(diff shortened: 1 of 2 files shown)


system/Throttle/Throttler.php
```diff
@@ -139,7 +139,7 @@ public function check(string $key, int $capacity, int $seconds, int $cost = 1):
         // How many seconds till a new token is available.
         // We must have a minimum wait of 1 second for a new token.
         // Primarily stored to allow devs to report back to users.
-        $newTokenAvailable = (int) ($refresh - $elapsed - $refresh * $tokens);
+        $newTokenAvailable = (int) round((1 - $tokens) * $refresh);
```
kenjis, line 142: Why is `$elapsed` gone?

rumpfc: Because `$tokens` at this moment is already `$old_token + $rate * $elapsed`, in other words, the new number of tokens after `$elapsed` seconds. The old formula indirectly substracted `$refresh` by elapsed twice, making it look like the countdown is faster than it should actually be.

The unit tests I commited proofed my assumption. Every "actual" tokenTime was much lower than the "expected" tokenTIme.

kenjis: Okay, thanks.

kenjis (approved):
LGTM!

MGatner (approved):
Good sleuthing! Very cool test. Thanks for the PR!

PR #7166, opened 2023-01-23
chore: update admin/*/composer.json

kenjis:
**Description**
Both appstarter and framework are for CodeIgniter projects, and the configuration itmes to be added in this PR should be the same.

- add `config`
- remove `suggest` in appstarter

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/admin/framework/composer.json
+++ b/admin/framework/composer.json
@@ -42,6 +42,11 @@
         "ext-fileinfo": "Improves mime type detection for files",
         "ext-readline": "Improves CLI::input() usability"
     },
+    "config": {
+        "optimize-autoloader": true,
+        "preferred-install": "dist",
+        "sort-packages": true
+    },
     "autoload": {
         "psr-4": {
             "CodeIgniter\\": "system/"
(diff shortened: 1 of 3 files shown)


kenjis:
@totoprayogo1916 I remove `suggest` in appstarter.

If we add it, it shows the same things as the framework's suggest.
```console
$ composer suggest
codeigniter4/appstarter suggests:
 - ext-oci8: If you use Oracle Database
 - ext-sqlsrv: If you use SQL Server
 - ext-memcache: If you use Cache class MemcachedHandler with Memcache

codeigniter4/framework suggests:
 - ext-memcache: If you use Cache class MemcachedHandler with Memcache
 - ext-oci8: If you use Oracle Database
 - ext-sqlsrv: If you use SQL Server

fakerphp/faker suggests:
 - doctrine/orm: Required to use Faker\ORM\Doctrine

2 additional suggestions by transitive dependencies can be shown with --all
```

kenjis:
It seems because you have installed fileinfo.

MGatner (approved):
💪

PR #7169, opened 2023-01-23
[Rector] Apply Rector to app/Views

samsonasik:
**Description**

Apply Rector run into `app/Views` as well.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/app/Views/errors/html/error_exception.php
+++ b/app/Views/errors/html/error_exception.php
@@ -1,4 +1,9 @@
-<?php $error_id = uniqid('error', true); ?>
+<?php
+use Config\Services;
+use CodeIgniter\CodeIgniter;
+
+$errorId = uniqid('error', true);
+?>
 <!doctype html>
 <html>
 <head>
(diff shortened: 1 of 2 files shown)


app/Views/errors/html/error_exception.php
```diff
@@ -289,9 +291,7 @@
                                 continue;
                             }
 
-                            if (! is_array($value)) {
-                                $value = [$value];
-                            } ?>
+                            $value = [$value]; ?>
```
samsonasik, line 294: this change due to `$headers` is from `headers()` method which got `array<string, Header>` docblock return, so it always a `Header` detected, so to loop, it wrapped with `[]`, but probably incorrect docblock, it needs to be verified.

@MGatner  @kenjis @paulbalandan could you verify ? Thank you.

kenjis: `$request->headers()` returns `array<string, Header>`.
https://github.com/codeigniter4/CodeIgniter4/blob/29c31173b6ecdd3be75ea69b3f5eb876e7b5dcbe/system/HTTP/MessageTrait.php#L116
https://github.com/codeigniter4/CodeIgniter4/blob/29c31173b6ecdd3be75ea69b3f5eb876e7b5dcbe/system/HTTP/MessageTrait.php#L154

So `$value` is never an array.
I sent a PR to refactor: #7171

app/Views/errors/html/error_exception.php
```diff
@@ -1,4 +1,6 @@
-<?php $error_id = uniqid('error', true); ?>
+<?php use Config\Services;
```
MGatner, line 1:
```suggestion
<?php
use Config\Services;
```

app/Views/errors/html/error_exception.php
```diff
@@ -1,4 +1,6 @@
...
+use CodeIgniter\CodeIgniter;
+$errorId = uniqid('error', true); ?>
```
MGatner, line 3:
```suggestion
$errorId = uniqid('error', true);
?>
```

MGatner (changes requested):
Looks good! Can we make the imports more like our newline formatting on PHP files? I don't think Rector will care.

app/Views/errors/html/error_exception.php
samsonasik: updated https://github.com/codeigniter4/CodeIgniter4/pull/7169/commits/ba96124a1ef53714f6512b6e0ffa2b207f57de85

app/Views/errors/html/error_exception.php
samsonasik: updated https://github.com/codeigniter4/CodeIgniter4/pull/7169/commits/ba96124a1ef53714f6512b6e0ffa2b207f57de85

samsonasik:
Thank you @kenjis @MGatner for the review.

PR #7172, opened 2023-01-24
fix: cannot create shared View instance when using debugbar

kenjis:
**Description**
- (1) fix: DebugBar initializes shared renderer instance
- (2) fix: Event collector does not collect Timeline data

(1) 
DebugBar initializes shared `renderer` instance.
The following code in a controller does not create a new instance with the view path.
```php
\Config\Services::renderer(__DIR__ . '/../Views');
```
(2)
After:
Screenshot 2023-01-24 16 18 54

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Toolbar/Collectors/Events.php
+++ b/system/Debug/Toolbar/Collectors/Events.php
@@ -11,11 +11,8 @@
 
 namespace CodeIgniter\Debug\Toolbar\Collectors;
 
-use CodeIgniter\View\RendererInterface;
-use Config\Services;
-
 /**
- * Views collector
+ * Events collector
  */
 class Events extends BaseCollector
 {
@@ -25,7 +22,7 @@ class Events extends BaseCollector
      *
      * @var bool
      */
-    protected $hasTimeline = false;
+    protected $hasTimeline = true;
 
     /**
      * Whether this collector needs to display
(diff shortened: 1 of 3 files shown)


system/Debug/Toolbar/Collectors/Events.php
```diff
@@ -63,7 +63,7 @@ class Events extends BaseCollector
      */
     public function __construct()
     {
-        $this->viewer = Services::renderer();
+        $this->viewer = Services::renderer(null, null, false);
```
lonnieezell, line 66: Would it be clearer to use `single_service('renderer')` here?

kenjis: Thanks. Using not shared service was wrong. Because the shared View has the performance data.
I will fix.

kenjis: Done.

system/Debug/Toolbar/Collectors/Views.php
```diff
@@ -73,12 +73,9 @@ class Views extends BaseCollector
      */
     protected $views = [];
 
-    /**
-     * Constructor.
-     */
-    public function __construct()
+    private function getViewer(): void
```
samsonasik, line 76: probably initViewer() is better naming since it doesn't return anything

kenjis: Exactly!
Fixed.

samsonasik (approved):

michalsn (approved):

PR #7173, opened 2023-01-24
docs: add about removal of deprecated Config items

kenjis:
**Description**
Supersedes #7021
See https://github.com/codeigniter4/CodeIgniter4/pull/7000#issuecomment-1364524729
- update `backward_compatibility_notes.rst`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/installation/backward_compatibility_notes.rst
+++ b/user_guide_src/source/installation/backward_compatibility_notes.rst
@@ -13,5 +13,7 @@ However, the code is not mature and bug fixes may break compatibility in minor r
 What are not Breaking Changes
 *****************************
 
+- The deprecated Config items are not covered by backwards compatibility (BC) promise. It may be removed in the next
+  **minor** version or later.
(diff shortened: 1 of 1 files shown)


samsonasik (approved):

samsonasik:
Thank you @kenjis

kenjis:
@lonnieezell @MGatner Are you okay with this?

MGatner:
I will defer to @lonnieezell - I'm not sure why we would ever need to remove a deprecated Config item, other than the very rare collision of things.

kenjis:
The reason is that there may be bugs (e.g., https://github.com/codeigniter4/CodeIgniter4/issues/4297#issuecomment-813110373) that cannot be fixed, and removing them makes the code easier to read and maintain.

lonnieezell (commented):
Already been merged, but I don't have problems with this, I suppose.

MGatner:
I don't think that "easier to read and maintain" is enough of a reason to remove a deprecation early. Definitely in the case of bugs - we have already discussed that allowance under our practice of "Semver".

PR #7176, opened 2023-01-25
fix: `spark routes` shows incorrect hostname routes

kenjis:
**Description**
From https://forum.codeigniter.com/showthread.php?tid=86264

HTTP_HOST do not exist in CLI, so routes for hostnames cannot currently be displayed by `spark routes`.

It is strange that `another.domain` is shown as preferred.

```php
$routes->get('/', 'Controller::default');
$routes->get('/', 'Controller::another', ['hostname' => 'another.domain']);
$routes->get('/', 'Controller::subdomain', ['subdomain' => 'subdomain']);
```
Before:
```
+--------+-------+------+--------------------------------------+----------------+---------------+
| Method | Route | Name | Handler                              | Before Filters | After Filters |
+--------+-------+------+--------------------------------------+----------------+---------------+
| GET    | /     | »    | \App\Controllers\Controller::another |                | toolbar       |
+--------+-------+------+--------------------------------------+----------------+---------------+
```
After:
```
+--------+-------+------+--------------------------------------+----------------+---------------+
| Method | Route | Name | Handler                              | Before Filters | After Filters |
+--------+-------+------+--------------------------------------+----------------+---------------+
| GET    | /     | »    | \App\Controllers\Controller::default |                | toolbar       |
+--------+-------+------+--------------------------------------+----------------+---------------+
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -1316,13 +1316,12 @@ class RouteCollection implements RouteCollectionInterface
         // Hostname limiting?
         if (! empty($options['hostname'])) {
             // @todo determine if there's a way to whitelist hosts?
-            if (isset($this->httpHost) && strtolower($this->httpHost) !== strtolower($options['hostname'])) {
+            if (! $this->checkHostname($options['hostname'])) {
                 return;
             }
 
             $overwrite = true;
         }
-
         // Limiting to subdomains?
         elseif (! empty($options['subdomain'])) {
             // If we don't match the current subdomain, then
@@ -1393,6 +1392,22 @@ class RouteCollection implements RouteCollectionInterface
         if (isset($options['redirect']) && is_numeric($options['redirect'])) {
             $this->routes['*'][$name]['redirect'] = $options['redirect'];
         }
+    }
+
+    /**
+     * Compares the hostname passed in against the current hostname
+     * on this page request.
+     *
+     * @param string $hostname Hostname in route options
+     */
+    private function checkHostname($hostname): bool
+    {
+        // CLI calls can't be on hostname.
+        if (! isset($this->httpHost)) {
+            return false;
+        }
+
+        return strtolower($this->httpHost) === strtolower($hostname);
     }
 
     private function processArrayCallableSyntax(string $from, array $to): string


iRedds:
I think it should be like this
```
| GET    | /                    | »    | \App\Controllers\Controller::default   |                | toolbar       |
| GET    | another.domain/      | »    | \App\Controllers\Controller::another   |               | toolbar       |
| GET    | subdomain.localhost/ | »    | \App\Controllers\Controller::subdomain |                | toolbar       |

```

kenjis:
@iRedds I too think all routes should be shown by the command.
But it is not possible now. We need the enhancement.

iRedds:
@kenjis It's not just the command.
Now routes that don't match HTTP_HOST are ignored. That is, for example, you cannot get the route path through the route_to() function if the domain does not match the current one.

kenjis:
@iRedds I don't know a use case that `route_to()` (or `url_to()`) with other hostname is needed.

To begin with, I don't really see the need for this limit to host/domain feature itself, which changes the route only for certain hostnames. 

In what case would it be used? Do you have real use cases?

iRedds:
The framework allows that one application can serve different domains. Correctly? So it makes sense that this application should be able to generate links to different domains.

For example, "our" application consists of several parts.
example.com - landing page
app.example.com is the private part of the application.

Or, for example, we need to generate a url like <user>.example.com or <locale>.example.com

kenjis:
@iRedds Do you have a real use case of multiple domains in a app?
Do you really want the functionality?
Why don't you have two apps if you serve two domains?

iRedds:
I don't have the case you describe.
Good question. Two domains = two applications.
But I have a question.
Why is such functionality as serving multiple domains with one application added to the framework?

michalsn:
I think we need an enhancement or alternative to the `route_to()` and `url_to()`, to support the domain part of the URL.

> Do you have a real use case of multiple domains with different routes in a app?

I used it in the SaaS app. Very handful.

> Why is such functionality as serving multiple domains with one application added to the framework?

There was a demand from developers for this. Personally, I think this is quite a common "problem" - a lot of services use multiple domains for certain areas of the application.

Since we support different hostnames/subdomains in the routes, being not able to generate this type of route was always a lacking feature for me.

michalsn (approved):

kenjis:
> a lot of services use multiple domains for certain areas of the application.

For what? Example?

michalsn:
I think iReeds already gave pretty valid examples.

If I have something like: `user.example.com` there is a big chance that the routes used here won't be valid for `example.com`, because this is a different part of the site/service.

kenjis:
iRedds example is just an imagination.
I want to know real use case to understand the needs.

Why don't you have two apps `user.example.com` and `example.com`?

> If I have something like: user.example.com there is a big chance that the routes used here won't be valid for example.com,

In the current implemention,
all the routes for `example.com` (w/o host/domain limitation) are valid to `user.example.com`.
Is that okay?

michalsn:
> Why don't you have two apps user.example.com and example.com?

We should not require specific architecture from our users. Besides, different domains do not necessarily mean we will deal with different applications.

Example. A "social network" app built around sports activities. People are engaging in conversation in the main domain `example.com` and can plan a bike route, but `my-username-goes-here.example.com` is my user's unique, public address where I can view my activities, others can send me a DM or so...

Another example would be different departments in a company, which are under other subdomains. Some parts of the application are common, but others have special modules. Of course, whether these should be separate applications or not is open, but again, that's probably not our role. Anyway, separate applications will not always be justified, and a different domain does not always mean a separate service.

> In the current implemention,
all the routes for example.com (w/o host/domain limitation) are valid to user.example.com.
Is that okay?

To be clear, something like that?
```php
$routes->group('', static function ($routes) {
    $routes->('something', 'Whatever::something');
});

$routes->group('', ['hostname' => 'user.example.com'], static function ($routes) {
    $routes->('else', 'Whatever::else');
});
```
Then accessing `https://user.example.com/something` should be valid.
If there is no host/domain limitation specified, this behavior is okay.

kenjis:
Yes, something like that.

In your example `my-username-goes-here.example.com`,
if there are routes:
- https://example.com/privacy
- https://example.com/info

the following URLs seem to be accessible:
- https://my-username-goes-here.example.com/privacy
- https://my-username-goes-here.example.com/info

It seems odd to me.

Also, if a different department runs a site on a different subdomain, each time a route with no host limitation is added on the main site, the URL for that department's site will also be added automatically.

michalsn:
Yes, that's true, but the solution for this is very simple. We have to declare the hostname for every route group definition. This way, everything will be separated correctly.

kenjis:
Okay, thanks.

Then the current structure of the $routes array does not look good. 
It seems better to add the hostname (regex) as a key.

PR #7179, opened 2023-01-25
BaseResult - make isValidResultId() protected

najdanovicivan:
**Description**
When isValidResultId() was introduced it was defined as private. I have custom DB driver which has custom Result object which will return array as result. The private modifier makes it impossible to override the validation check.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseResult.php
+++ b/system/Database/BaseResult.php
@@ -470,7 +470,7 @@ abstract class BaseResult implements ResultInterface
         return $this->numRows = count($this->getResultArray());
     }
 
-    private function isValidResultId(): bool
+    protected function isValidResultId(): bool
     {
         return is_resource($this->resultID) || is_object($this->resultID);
     }


kenjis:
I'm fine with this change.

Can we accept this in `develop` branch?
Is this a bug fix? enhancement?

najdanovicivan:
@kenjis

I consider this a bug fix.  

9bdce7523c01b1ff374041e90dc5023efeb5b68e introduced the change 
from:
```
is_bool($this->resultID) || ! $this->resultID
```
to : 
```
is_resource($this->resultID) || is_object($this->resultID);
```

Which is BC for custom Result classes which are using something other than resource or object to return database query results

kenjis:
Why? Isn't it a bug fix?

`$this->resultID` is a return value (that is not false) of PHP database functions.
What PHP functions return values of other types?
https://github.com/codeigniter4/CodeIgniter4/blob/ecef7f748df25a20f65bd2e2df26070fc3ebffdd/system/Database/BaseConnection.php#L605
https://github.com/codeigniter4/CodeIgniter4/blob/ecef7f748df25a20f65bd2e2df26070fc3ebffdd/system/Database/BaseConnection.php#L666-L674

najdanovicivan:
I have custom Database drivers for ElasticSearch/OpenSearch and MongoDB which were working fine before this change

The result type comes from the Connection execute() method.

I have custom Connection class which is overriding the execute() method in case of ElasticSearch/OpenSearch it will use the Client object from the SDK to execute the database query. This will return result as array

MongoDB is different story it uses the Collection object which can returns sorts of types bases on which query is executed so it can return BSONDocument|BulkWriteResult|Cursor|InsertOneResult|int|UpdateResult

I've been using such implementation since the alpha version of CI4 and it was all working because before 4.3.0 the result was considered valid if it was anything except null

Now after the update the implementation completely broke as it will only work with object and resources so any scalar or array value will no longer be considered valid. 

I'm completely fine with having the method check for resource or object for SQL but it should be protected instead of private so that override can be made in custom driver implementations. Now in order to get around this this limitation i need to override 3 methods in custom Result classes getCustomResultObject(), getResultArray() and getResultObject() which are only copied over from BaseResult so that custom isValidResultId() can be used.

kenjis:
It seems from v4.0.0 the `$resultID` type is intended as `object|resource` or `false`.
https://github.com/codeigniter4/CodeIgniter4/blob/47dbb16bb5e5784f45bf09a7380016f2d36f7c39/system/Database/BaseConnection.php#L213-L218
`array` is not intended. So I think the change is not a bug.

najdanovicivan:
I don't mean that resource or object check is bug. I consider bug having the method defined as private instead of protected.

kenjis:
But after all `$resultID` should not have an array.
https://github.com/codeigniter4/CodeIgniter4/blob/f79b1753b1ccbd1a1e8afe62726cdcf80f244562/system/Database/BaseResult.php#L33-L38

At first I thought it is fine that just changing `private` to `protected`. 
But is it correct to change the type of a property in a child class?

1. if the type is different, it should not be extended in the first place.
2. the type of `$resultID` in `BaseResult` (and `BaseConnection`) should be changed more widely if it should be extended.

najdanovicivan:
I needed to extend those in order to make driver/model which will work with NoSQL databases. I know this is not a good option but it works.  I've been using such implementation for long time . Currently we do not have any other means or standards to implement such drivers. Having custom Connection, Query, Result and Model was the only way I was able to do NoSQL implementation which works. The only things that have something to do with the $resultID are Connection and Builder and in case with Custom Driver you'll also need to have custom Connection and custom builder since BaseBuilder does not make sense to be used for NoSQL. So I don't see any limitation why $resultID type should not be changed in such case.

kenjis:
We cannot change property types in a child class.
See https://3v4l.org/iv0Sn

I don't know if the base database classes can support NoSQL, but if you want it to, we will need to change them to do so.

Currently, there is no driver that supports NoSQL, so I don't even know how these should be changed.

PR #7180, opened 2023-01-26
docs: define new term "Route path"

kenjis:
**Description**
The term "route" is ambiguous. It may mean an entry of the route collection.

- change "Route" to "Route path" 
- add description

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


datamweb (approved):

kenjis:
@datamweb Thank you for review!

PR #7181, opened 2023-01-26
Feature: Make Model::updateBatch() work with Builder methods

sclubricants:
This PR allows the use of `setData()`, `updateFields()`, `onConstraint()`, etc with `Model::updateBatch()`.

This also changes the `$index` parameter of `Model::updateBatch()` to `$constraints` bringing it inline with `Builder`.

Because the new `Builder::updateBatch()` uses a different SQL structure some changes were made in Model's version. Instead of filtering the allowed fields from the dataset they are now filtered from `$QBOptions['updateFields']`. These are ultimately the fields that will get updated. Currently if someone were to use `Builder::updateFields()`  with the current version of model they could buypass the filter. Also not filtering the dataset allows inclusion of other fields that can be used with `updateFields()`  to create a calculated field update.

Once we can iron out this I will update the other batch methods as well.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


system/BaseModel.php
```diff
@@ -343,6 +345,15 @@ protected function initialize()
...
     }
 
+    /**
+     * Provides a shared instance of the Query Builder.
+     *
+     * @return BaseBuilder
+     *
+     * @throws ModelException
+     */
+    abstract public function builder(?string $table = null);
```
kenjis, line 355: Why does BaseModel have `builder()`?
Any kind of database should have the Query Builder?

system/BaseModel.php
```diff
@@ -421,16 +432,16 @@ abstract protected function doUpdate($id = null, $data = null): bool;
      * Compiles an update and runs the query.
      * This method works only with dbCalls.
      *
-     * @param array|null  $set       An associative array of update values
-     * @param string|null $index     The where key
-     * @param int         $batchSize The size of the batch to run
-     * @param bool        $returnSQL True means SQL is returned, false will execute the query
+     * @param array|null               $set         An associative array of update values
+     * @param array|RawSql|string|null $constraints
```
kenjis, line 436: `RawSql` seems to be SQL database specific.
Any kind of database should use it?

system/Database/BaseBuilder.php
```diff
@@ -2000,7 +2000,7 @@ protected function _upsertBatch(string $table, array $keys, array $values): stri
     /**
      * Set table alias for dataset pseudo table.
      */
-    private function setAlias(string $alias): BaseBuilder
+    public function setAlias(string $alias): BaseBuilder
```
kenjis, line 2003: Please revert. This method is not a public method.
We can set the alias with `setData()` public method.

system/Database/BaseBuilder.php
```diff
@@ -2136,6 +2150,28 @@ public function setQueryAsData($query, ?string $alias = null, $columns = null):
         return $this;
     }
 
+    /**
+     * Gets $QBOptions for use in Model
+     */
+    public function getQBOptions(): array
```
kenjis, line 2156: This method destroys encapsulation. Is there any way to remove it?

system/Database/OCI8/Builder.php
```diff
@@ -255,6 +255,8 @@ protected function _updateBatch(string $table, array $keys, array $values): stri
                 return ''; // @codeCoverageIgnore
             }
 
+            // remove any constraints from updateFields
+            $this->updateFields($this->QBOptions['updateFields'] ?? [], false, $constraints);
```
kenjis, line 259: > I found a bug where a field that exists in constraints and also is in updateFields causes and error. This is only for Oracle.

This bug seems to be also in `develop`.
Can you send another PR to `develop`?

system/Database/BaseBuilder.php
sclubricants: I would have to implement all the QBOptions methods is Model. Otherwise I need some way of accessing and destroying these values from outside of Builder.

system/Database/OCI8/Builder.php
sclubricants: $builder->onConstraint('email')->updateFields(['email', 'name'])->updateBatch($data);

If only happens when you explicitly define with updateFields(). If you let the system define them based on the data then everything is fine. So it is kind of a bug but not really because you shouldn't be defining a constraint as an updateField.

sclubricants:
>And BaseModel is for all databases, Model is only for SQL databases.

I really don't quite get the difference between BaseModel and Model. I just tried to follow what was already there.. otherwise I probably would have done it differently.

What is a database other than a SQL database? Are these suppose to have the same methods available as SQL Databases? These databases are not compatible with Builder?

For example: with updateBatch() the dataset need not even have the updated_at field when using Builder. It can be set using updateFields instead. For other databases not compatible with Builder who knows what format they need the data. Who knows if they can even perform such operations as updateBatch(). 

If Model is the class for SQL databases and Builder only supports SQL databases then why was updateBatch() only implemented in BaseModel? 

It seems I should leave the method in BaseModel alone and just completely override it in Model. No need to execute the parent method. But then the question remains, what is the point of the BaseModel method? What are its requirements?

system/BaseModel.php
sclubricants: Its a Builder thing anyways. I probably should move this logic to Model.

system/BaseModel.php
sclubricants: Ya, trying to understand the difference between BaseModel and Model. It seems Model is based on Builder but BaseModel is not.

iRedds:
Don't you think that the model turns into some kind of monster? We add methods to the model for convenient use of QueryBuilder, thereby turning it into an adapter for BaseBuilder.

sclubricants:
>Don't you think that the model turns into some kind of monster?

Ya I was trying to avoid that.

I almost think it might work better to change Builder to be able to define the timestamp fields and to filter the fields. Then use model to set these options on builder.

sclubricants:
@kenjis Ok, I completely changed this around to deal with all the issues you pointed out. 

The thing now is that BaseModel::updateBatch() isn't being used anymore by the framework. Should we remove the method?

iRedds:
@sclubricants In fact, you have now written the behavior of Laravel.
There, the model (entity) is injected into the builder. And the behavior of the builder is based on the data of the model.

sclubricants:
@iRedds 

I'd really like model to return Entity or a new EntityCollection class and then have DML methods on those classes.

Then you could do:

```php 
$entity = $model->where('some condition')->getEntity();
$entity->name = 'change the name';
$entity->update();

// OR

$entityCollection = $model->where('some condition')->getEntityCollection();

foreach ($entityCollection as $entity) {
    $entity->name = 'change the name';
    $entity->update(); // could save individually
}

$entityCollection->update(); // performs batch method
```

Really entity should be model.

kenjis:
> What is a database other than a SQL database? 

NoSQL databases.

> The goal with this change is to have BaseModel which does not contain any database related code so that it can easily be extended to support other kinds of databases like NoSQL document based
https://github.com/codeigniter4/CodeIgniter4/pull/3930

> Are these suppose to have the same methods available as SQL Databases? 

BaseModel should provide only methods for all databases.

> These databases are not compatible with Builder?

Probably yes.

kenjis:
> If Model is the class for SQL databases and Builder only supports SQL databases then why was updateBatch() only implemented in BaseModel?

I'm not sure, but something like batch updates may exist in other databases.
E.g., https://www.mongodb.com/docs/manual/reference/method/db.collection.updateMany/

And the execution of the query is delegated to `doUpdateBatch()`.

kenjis:
```php
$entity = $model->where('some condition')->getEntity();
$entity->name = 'change the name';
$entity->update();
```
This is completely different design from the current CI4 implementation.
An entity in CI4 represents a record, and should not know how to update itself.
So `$entity->update()` should not work.

iRedds:
@sclubricants This will no longer be an entity, but an ActiveRecord.

sclubricants:
Well this PR isn't attempting to do anything with Entity.

PR #7191, opened 2023-01-27
Correcting documentation mistakes in upgrading from one version to another

objecttothis:
**Description**
Documentation problems
https://codeigniter.com/user_guide/installation/upgrade_405.html#message-getheader-s
In the line `Request[Interface]::getMethod(): The $upper parameter will be removed, use str_to_upper()`
The correct PHP function is `strtoupper()`

https://codeigniter.com/user_guide/installation/upgrade_412.html#baseconnection-query-return-values
In the line `Review any use of query() method and be assess whether the value might be boolean instead of Result object.`
The word "be" does not belong in the sentence.

https://codeigniter.com/user_guide/installation/upgrade_431.html#config
Based on the description below it `app/Config/Encryption.php` should be `app/Config/Email.php`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


kenjis (approved):
Thank you!

PR #7193, opened 2023-01-28
fix: Autoloader may not add Composer package's namespaces

kenjis:
**Description**
Fixes https://github.com/codeigniter4/shield/issues/610

PHPStan invokes phar (`require 'phar://phpstan.phar/bin/phpstan';`).
In that case, `InstalledVersions::getAllRawData()[0]['versions']` does not contain correct data.

When I see `getInstalledPackages()`:
https://github.com/composer/composer/blob/50cded331ced9acb4e926be3dda1f74b86af2a3b/src/Composer/InstalledVersions.php#L46-L64
it seems better to use all items in the returned array.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -381,7 +381,12 @@ class Autoloader
             );
         }
         // This method requires Composer 2.0.14 or later.
-        $packageList = InstalledVersions::getAllRawData()[0]['versions'];
+        $allData     = InstalledVersions::getAllRawData();
+        $packageList = [];
+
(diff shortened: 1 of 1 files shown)


michalsn (approved):
The explanation makes sense to me.

kenjis:
Processing all elements increases the number of loops. 
If I could reduce the number of loops for sure, I would.
But I don't understand the details of the Composer implementation, 
so I'll go with it for now.

PR #7200, opened 2023-01-30
fix: base_url() removes trailing slash in baseURL

### LINKED ISSUE #7184: Bug: base_url() removes trailing slash in baseURL
### PHP Version

8.1

### CodeIgniter4 Version

4.3.1

Git

macOS

### Which server did you use?

### Database

n/a

### What happened?

`base_url()` returns `http://localhost:8080`.

The baseURL should have a trailing slash.
https://github.com/codeigniter4/CodeIgniter4/blob/ecef7f748df25a20f65bd2e2df26070fc3ebffdd/app/Config/App.php#L15-L16

### Steps to Reproduce

```php
<?php

namespace App\Controllers;

{
    public function index()
    {
        return base_url();
    }
}
```

### Expected Output

`http://localhost:8080/`

### Anything else?

CI 3.1.13 returns `http://localhost:8008/`.

kenjis:
**Description**
Fixes #7184

`base_url()` without arguments returns `http://localhost:8080`, not `http://localhost:8080/`.
The baseURL should have trailing slash, and there is no reason to remove it.
CI3's `base_url()` returns the URL with trailing slash.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -124,7 +124,7 @@ if (! function_exists('base_url')) {
         $config            = clone config('App');
         $config->indexPage = '';
 
-        return rtrim(site_url($relativePath, $scheme, $config), '/');
+        return site_url($relativePath, $scheme, $config);
     }
 }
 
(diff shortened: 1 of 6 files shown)


samsonasik:
Upgrade note seems need to be updated in case user use base_url() with their own tweak to add / after

MGatner (changes requested):
I don't recall why this was as such but please do indicate it in the changelogs as this could cause a mess for people, especially in tests.

kenjis:
Added the docs.

MGatner (approved):

samsonasik (approved):

PR #7208, opened 2023-02-01
feat: add `Entity::injectRawData()` to avoid name collision

### LINKED ISSUE #5762: Bug: Entity can't handle column named `attributes`

kenjis:
**Description**
Supersedes #5763, #5781
Fixes #5762

- add `Entity::injectRawData()`
- deprecate `Entity::setAttributes()`
- ~~require setter/getter methods must be public. See https://github.com/codeigniter4/CodeIgniter4/issues/5762#issuecomment-1179519591~~

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/MySQLi/Result.php
+++ b/system/Database/MySQLi/Result.php
@@ -146,7 +146,7 @@ class Result extends BaseResult
     protected function fetchObject(string $className = 'stdClass')
     {
         if (is_subclass_of($className, Entity::class)) {
-            return empty($data = $this->fetchAssoc()) ? false : (new $className())->setAttributes($data);
+            return empty($data = $this->fetchAssoc()) ? false : (new $className())->injectRawData($data);
         }
 
         return $this->resultID->fetch_object($className);
--- a/system/Database/OCI8/Result.php
+++ b/system/Database/OCI8/Result.php
@@ -102,7 +102,7 @@ class Result extends BaseResult
             return $row;
         }
         if (is_subclass_of($className, Entity::class)) {
-            return (new $className())->setAttributes((array) $row);
+            return (new $className())->injectRawData((array) $row);
         }
 
         $instance = new $className();
--- a/system/Database/Postgre/Result.php
+++ b/system/Database/Postgre/Result.php
@@ -112,7 +112,7 @@ class Result extends BaseResult
     protected function fetchObject(string $className = 'stdClass')
     {
         if (is_subclass_of($className, Entity::class)) {
-            return empty($data = $this->fetchAssoc()) ? false : (new $className())->setAttributes($data);
+            return empty($data = $this->fetchAssoc()) ? false : (new $className())->injectRawData($data);
         }
 
         return pg_fetch_object($this->resultID, null, $className);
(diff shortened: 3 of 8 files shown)


iRedds:
To be honest, I do not agree with this PR.
I believe that the implementation should be elegant like in laravel.
And adding reflection is already too much.
In my opinion, setters and getters should not be public, as they are specially made for the hidden implementation of logic.

kenjis:
Thank you for your comment.

> I believe that the implementation should be elegant like in laravel.

Can you elaborate? *elegant like in laravel* explains nothing to me.

iRedds:
To serve an entity
setAttributes()
getAttributes() instead of toRawArray()

setter and getter for attributes
set<AttributeName>Attribute()
get<AttributeName>Attribute()

Otherwise, any new entity functionality (method names) will need to be adjusted for the fact that the developer may have an attribute with the same name.

MGatner (changes requested):
I don't see the public requirement being necessary here, and it is a very likely breaking change. We can discuss it here, or my preference would be to split it out into a separate PR and proceed with the injection method.

kenjis:
> I don't see the public requirement being necessary here, and it is a very likely breaking change. 

Indeed.

I dropped it.

kenjis:
@iRedds 
I don't think we need `getAttributes()`. The `attributes` property (or the word attributes) should not be actively disclosed. It is just an internal implementation.

iRedds:
@kenjis The ``getAttributes()`` method is the normal name for a class property getter. Just like ``setAttributes()``.

Sorry, but I don't agree with you.
The developer can create his own implementation of the model that works with the entity, which means that he must be able to receive data from the entity as is.
It is not clear to me what kind of **internal implementation** we are talking about.

kenjis:
In my opinion, the `attributes` property is not a normal property.
It should be hidden from users.

In fact, the attributes holds all the data for the record, but that is just the current implementation. 
There could be another implementation, and it does not matter if the name attributes is changed. 
It is an internal implementation and should be hidden.
Users do not need to be aware of it, and if they cannot use it without being aware of it, 
then it is poorly implemented.

MGatner (approved):
Looks good! Re: getAttributes()... how is this different than toRawArray()?

kenjis:
Added changelog.

kenjis:

@mostafakhudair @najdanovicivan Any comments?

najdanovicivan (approved):

kenjis:
Thank you for the review! @MGatner @najdanovicivan

PR #7212, opened 2023-02-02
fix: stack trace displayed when Exception handler runs out of memory is useless

kenjis:
**Description**
See https://forum.codeigniter.com/showthread.php?tid=86536&pid=406349#pid406349

If "Allowed memory size of xxx bytes exhausted" error occurs in the Exception handler,
the true Exception does not show in the debug stack trace.

This PR is not perfect, but DX would be improved.

Before:
Screenshot 2023-02-02 13 29 05

After:
Screenshot 2023-02-02 at 13-29-41 ErrorException

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -68,6 +68,8 @@ class Exceptions
      */
     protected $response;
 
+    private ?Throwable $exceptionCaughtByExceptionHandler = null;
+
     /**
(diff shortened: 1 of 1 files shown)


MGatner (approved):
It's an improvement.

PR #7215, opened 2023-02-03
fix: site_url() does not use alt Config and refactoring

kenjis:
**Description**
Supersedes #7211

I need to refactor the three function code because I cannot fix without adding an parameter.

- fix `site_url()` does not use alt Config
- refactor `base_url()`, `site_url()`, `current_url()`
- change `_get_uri()`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


system/Helpers/url_helper.php
```diff
@@ -71,16 +82,7 @@ function _get_uri(string $relativePath = '', ?App $config = null): URI
...
-
-        $uri = new URI($url);
-
-        // Check if the baseURL scheme needs to be coerced into its secure version
-        if ($config->forceGlobalSecureRequests && $uri->getScheme() === 'http') {
-            $uri->setScheme('https');
-        }
-
-        return $uri;
+        return new URI($url . $relativePath);
```
michalsn, line 85: What about the scenario when we use `base_url()` and we have set `forceGlobalSecureRequests` to `true`?

system/Helpers/url_helper.php
```diff
@@ -121,37 +132,59 @@ function site_url($relativePath = '', ?string $scheme = null, ?App $config = nul
...
+        $uri = _get_uri($relativePath);
 
-        return site_url($relativePath, $scheme, $config);
+        return URI::createURIString(
+            $scheme ?? $uri->getScheme(),
+            $uri->getAuthority(),
+            $uri->getPath(),
+            $uri->getQuery(),
+            $uri->getFragment()
+        );
```
michalsn, line 143: For me, the previous code was more readable.

michalsn (commented):
Fixing the custom config is fine. However, I'm not a fan of other parts that were refactored.

system/Helpers/url_helper.php
kenjis: I need to refactor the whole three function code because I cannot fix without adding an parameter.
Because the previous code passes the Config object to `site_url()`.
So the Config value is always used and `base_url()` cannot support multiple domain.

system/Helpers/url_helper.php
kenjis: Thanks. Added a commit to fix it.

I'm not sure this is correct usage, but:
```diff
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -130,7 +130,7 @@ class App extends BaseConfig
      * secure, the user will be redirected to a secure version of the page
      * and the HTTP Strict Transport Security header will be set.
      */
-    public bool $forceGlobalSecureRequests = false;
+    public bool $forceGlobalSecureRequests = true;
 
     /**
      * --------------------------------------------------------------------------
diff --git a/app/Controllers/Home.php b/app/Controllers/Home.php
index 7f867e95ff..97f1617d13 100644
--- a/app/Controllers/Home.php
+++ b/app/Controllers/Home.php
@@ -6,6 +6,6 @@ class Home extends BaseController
 {
     public function index()
     {
-        return view('welcome_message');
+        return base_url();
     }
 }
```
```console
$ php public/index.php 
https://localhost:8080/
```

system/Helpers/url_helper.php
michalsn: Ok, thanks - now I see the problem.

Since we're cloning the App config here... can't we use something like this?
```php
$config            = clone config('App');
$config->baseURL   = Services::request()->getUri()->getBaseURL();
$config->indexPage = '';

return site_url($relativePath, $scheme, $config);
```

kenjis: Nice!

I refactored and we are back to almost the same code as before.

system/Helpers/url_helper.php
```diff
@@ -26,17 +26,24 @@
...
     {
-        $config ??= config('App');
+        $appConfig = null;
+        if ($config === null) {
+            /** @var App $appConfig */
+            $appConfig = config('App');
+        }
 
-        if ($config->baseURL === '') {
-            throw new InvalidArgumentException('_get_uri() requires a valid baseURL.');
```
michalsn, line 39: Is there a reason why we're removing this check?

system/Helpers/url_helper.php
michalsn: Thanks!

system/Helpers/url_helper.php
kenjis: I missed it. I will check.

kenjis: Added above.

michalsn: Thanks!

michalsn (approved):

MGatner (commented):
I've only given this a first read. If you and @michalsn feel confident about it feel free to proceed, otherwise I will come back for a more thorough review. Framework URI handling is a sticky business.

kenjis:
In the end, the code was not that different from the original code. 
Also, michalsn has done a careful review.

PR #7011, opened 2022-12-23
refactor: remove Cookie config items in Config\App

kenjis:
**Description**
See https://github.com/codeigniter4/CodeIgniter4/pull/7000#discussion_r1054048398

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #7221, opened 2023-02-04
refactor: remove Cookie config items in Config\App

kenjis:
**Description**
Supersedes #7011

This PR is needed to fix Session Config issue: https://github.com/codeigniter4/CodeIgniter4/issues/4297#issuecomment-813110373

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


michalsn (approved):
As you know, I'm not a fan of removing things like this in a minor version, but since others think differently and there is an agreement about it (I as understand) then I only checked the code and it seems ok 👍

MGatner (approved):
Actually glad to see this one happen. Hopefully not too many repercussions for devs.

PR #7223, opened 2023-02-05
docs: improve query_builder.rst

### LINKED ISSUE #7216: False Information In Documentation With Actual Code
My CI4 Version  : `4.2.10`
Documentation Link : https://codeigniter.com/user_guide/database/query_builder.html#updatebatch

I don't know if I'm reading the documentation wrong, but it does explain that the second parameter can take Array data.

`$builder->updateBatch($data, ['title', 'author']);`

But when I open the function in the code, it says that the second parameter is a string.

Path : `vendor/codeigniter4/framework/system/Database/BaseBuilder.php/updateBatch`

```
/**
   * Compiles an update string and runs the query
   *
   * @throws DatabaseException
   *
   * @return false|int|string[] Number of rows affected or FALSE on failure, SQL array when testMode
   */
   public function updateBatch(?array $set = null, ?string $index = null, int $batchSize = 100)
   {
   ....
```

What I want to do is do a Batch update but use more than one index. Back again on the documentation there also says can use `$builder->setData($data)->onConstraint('title, author')->updateBatch();`. But the onConstraint function is not found. And I want to ask if there is a way to batch update with more than one index?

kenjis:
**Description**
Fixes #7216

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/database/query_builder.rst
+++ b/user_guide_src/source/database/query_builder.rst
@@ -908,6 +908,8 @@ Upsert
 $builder->upsert()
 ------------------
 
+.. versionadded:: 4.3.0
+
 Generates an upsert string based on the data you supply, and runs the
 query. You can either pass an **array** or an **object** to the
 method. By default a constraint will be defined in order. A primary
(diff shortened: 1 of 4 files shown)


michalsn (approved):

samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #7224, opened 2023-02-05
refactor: Update PHPDoc Common::config

maniaba:
**Description**
The `@template` annotation relates to a concept called Generics, which does not currently exist in PHP, but are a way of dynamically describing the contents of a parameter or return type that would be unknown until a class is instantiated or method called.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Common.php
+++ b/system/Common.php
@@ -204,11 +204,12 @@ if (! function_exists('config')) {
     /**
      * More simple way of getting config instances from Factories
      *
-     * @template T of BaseConfig
+     * @template ConfigTemplate of BaseConfig
      *
-     * @param class-string<T> $name
+     * @param class-string<ConfigTemplate>|string $name
      *
-     * @return T
+     * @return ConfigTemplate
+     * @phpstan-return ($name is class-string<ConfigTemplate> ? ConfigTemplate : object|null)
      */
     function config(string $name, bool $getShared = true)
     {
--- a/system/Common.php
+++ b/system/Common.php
@@ -978,9 +978,8 @@ if (! function_exists('session')) {
      *    session()->set('foo', 'bar');
      *    $foo = session('bar');
      *
-     * @param string|null $val
-     *
      * @return array|bool|float|int|object|Session|string|null
+     * @phpstan-return ($val is null ? Session : array|bool|float|int|object|string|null)
      */
     function session(?string $val = null)
     {
--- a/system/Common.php
+++ b/system/Common.php
@@ -10,6 +10,7 @@
  */
 
 use CodeIgniter\Cache\CacheInterface;
+use CodeIgniter\Config\BaseConfig;
 use CodeIgniter\Config\Factories;
 use CodeIgniter\Cookie\Cookie;
 use CodeIgniter\Cookie\CookieStore;
@@ -47,8 +48,7 @@ if (! function_exists('app_timezone')) {
      */
     function app_timezone(): string
     {
-        /** @var App $config */
-        $config = config('App');
+        $config = config(App::class);
 
         return $config->appTimezone;
     }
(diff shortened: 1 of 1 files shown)


system/Common.php
```diff
@@ -973,10 +978,9 @@ function route_to(string $method, ...$params)
      *    session()->set('foo', 'bar');
      *    $foo = session('bar');
      *
-     * @param string $val
+     * @param string|null $val
      *
      * @return array|bool|float|int|object|Session|string|null
-     * @phpstan-return ($val is null ? Session : array|bool|float|int|object|string|null)
```
kenjis, line 979: Why do you remove this?

kenjis:
Thank you for sending PR.

Please sign all your commits:
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#signing

Fix the coding style:

Fix PHPStan errors:

system/Common.php
```diff
@@ -203,7 +204,12 @@ function command(string $command)
     /**
      * More simple way of getting config instances from Factories
      *
-     * @return object|null
+     * @template ConfigTemplate of BaseConfig
+     *
+     * @param class-string<ConfigTemplate>|string $name
+     *
+     * @return ConfigTemplate
+     * @phpstan-return ($name is class-string<ConfigTemplate> ? ConfigTemplate : object|null)
```
paulbalandan, line 212: Does it really return a generic object? I thought if not found the function will return `null`.

maniaba: I add null as a possible return

system/Common.php
```diff
@@ -203,7 +204,12 @@ function command(string $command)
     /**
      * More simple way of getting config instances from Factories
      *
-     * @return object|null
+     * @template ConfigTemplate of BaseConfig
+     *
+     * @param class-string<ConfigTemplate>|string $name
```
kenjis, line 209: Is this better according to our current rule?
Probably the type `ConfigTemplate` is not understood by IDEs.

```suggestion
     * @phpstan-param class-string<ConfigTemplate>|string $name
```

system/Common.php
```diff
@@ -203,7 +204,12 @@ function command(string $command)
...
+     *
+     * @return ConfigTemplate|null
```
kenjis, line 211: Like this?
```suggestion
     * @return BaseConfig|null
```

kenjis, line 211: Probably this must be reverted.

```suggestion
     * @return object|null
```

system/Common.php
maniaba: @kenjis After this change phpstan throws the errors.

kenjis: What errors?
No errors on my local environment.
```diff
--- a/system/Common.php
+++ b/system/Common.php
@@ -206,9 +206,9 @@ if (! function_exists('config')) {
      *
      * @template ConfigTemplate of BaseConfig
      *
-     * @param class-string<ConfigTemplate>|string $name
+     * @phpstan-param class-string<ConfigTemplate>|string $name
      *
-     * @return ConfigTemplate|null
+     * @return object|null
      * @phpstan-return ($name is class-string<ConfigTemplate> ? ConfigTemplate : object|null)
      */
     function config(string $name, bool $getShared = true)
```

maniaba: When we use @return object|null and only @phpstan-param there is no auto suggestion (is not understood by IDEs).

kenjis: @maniaba Exactly, PhpStorm does not do auto suggestion.

kenjis (approved):
This is the same thing as `model()`, and works with PhpStorm.

system/Common.php
maniaba: let's leave it as it is right now?
It is good for phpstan and IDE.

paulbalandan (approved):

kenjis:
@maniaba Thank you.

We will accept the change from `config('Foo')` to `config(Foo::class)` in everywhere.

PR #7226, opened 2023-02-05
fix: support for display of error message using wildcard (*)

sammyskills:
**Description**
The `validation_show_error()` form helper does not support the use of wildcards (*) like the `getError()` validation method. 

For example, if I have a rule like this:
```php
$rules = [
    'users.*.name' => "required"
];
```
I expect that I should be able to display the error in the view using `validation_show_error('users.*.name')` just the same way I would if I use `$validation->getError('users.*.name')`.
This PR fixes that.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Helpers/FormHelperTest.php
+++ b/tests/system/Helpers/FormHelperTest.php
@@ -999,7 +999,16 @@ final class FormHelperTest extends CIUnitTestCase
     public function testValidationShowErrorForWildcards()
     {
         $validation = Services::validation();
-        $validation->setRule('user.0.name', 'Name', 'required')->run([]);
+        $validation->setRule('user.*.name', 'Name', 'required')
+            ->run([
+                'user' => [
+                    'friends' => [
+                        ['name' => 'Name1'],
+                        ['name' => ''],
+                        ['name' => 'Name2'],
+                    ]
+                ]
+            ]);
 
         $html = validation_show_error('user.*.name');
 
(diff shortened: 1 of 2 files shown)


tests/system/Helpers/FormHelperTest.php
```diff
@@ -996,6 +996,16 @@ public function testValidationShowError()
         $this->assertSame('<span class="help-block">The ID field is required.</span>' . "\n", $html);
     }
 
+    public function testValidationShowErrorForWildcards()
+    {
+        $validation = Services::validation();
+        $validation->setRule('user.0.name', 'Name', 'required')->run([]);
```
kenjis, line 1002:
```suggestion
        $validation->setRule('user.*.name', 'Name', 'required')
            ->run([
                'user' => [
                    'friends' => [
                        ['name' => 'Name1'],
                        ['name' => ''],
                        ['name' => 'Name2'],
                    ],
                ],
            ]);
```

sammyskills: Done. 
Thank you.

michalsn (commented):
Why can't we just call the `showError()` from the `Validation` class? Am I missing something here?

sammyskills:
If you use `redirect()->back()->withInput()` in the controller and `validation_show_error()` in the view file, error message for rules that contains wildcards (*) will not be displayed.

IMO, I don't think it is a good idea to call the validation object again using `$validation->get error()` since it is already done in the form helper function.

michalsn (approved):
Ok. I missed the part where we're checking the session info. This is fine. Thanks!

kenjis (approved):

PR #7230, opened 2023-02-07
feat: add new setter/getter for Entity

kenjis:
~~Needs  #7208~~

**Description**
Add special getter/setter to Entity to avoid method name conflicts.
- add new special setter method "`_set` + key" for a attribute
- add new special getter method "`_get` + key" for a attribute

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Entity/Entity.php
+++ b/system/Entity/Entity.php
@@ -456,11 +456,19 @@ class Entity implements JsonSerializable
 
         $value = $this->castAs($value, $key, 'set');
 
-        // if a set* method exists for this key, use that method to
+        // if a setter method exists for this key, use that method to
         // insert this value. should be outside $isNullable check,
         // so maybe wants to do sth with null value automatically
         $method = 'set' . str_replace(' ', '', ucwords(str_replace(['-', '_'], ' ', $key)));
 
+        // If a "`_set` + $key" method exists, it is a setter.
+        if (method_exists($this, '_' . $method)) {
+            $this->{'_' . $method}($value);
+
+            return $this;
+        }
+
+        // If a "`set` + $key" method exists, it is also a setter.
         if (method_exists($this, $method) && $method !== 'setAttributes') {
             $this->{$method}($value);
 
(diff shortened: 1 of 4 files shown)


MGatner:
I'm not sure what this is needed for? `bar:foo`

kenjis:
For method name collision. https://github.com/codeigniter4/CodeIgniter4/issues/5762#issuecomment-1179520069

MGatner:
I meant the colons in the value, but now I see those are an artifact of the test fixture, not part of the code! My mistake.

kenjis:
Added docs.

kenjis:
@mostafakhudair @najdanovicivan @iRedds @the-gt99 @lonnieezell 
Any comments?

najdanovicivan:
I think this adds a lot of unnecessary complexity to have multiple ways of defining setters. So far the only problem we have is when we need to override setAttributes method. Since now we have setAttributes as deprecated and we're going forward with injectRawData() method. I'd rather use https://www.php.net/manual/en/reflectionmethod.getdeclaringclass.php to check if setAttributes is not from CodeIgniter\Entity or  CodeIgniter\Entity\Entity class. So if method is not from base CI4 classes it can be handled as regular setter for $this->attributes['attributes'].  The only problem with this one is BC in Entity classes which override the setAttributes method.

kenjis:
@najdanovicivan Thank you for your opinion.

The main reason of this PR is to avoid setter/getter method name conflicts,
not only for `setAttributes()`.
See https://github.com/codeigniter4/CodeIgniter4/pull/7230/files#diff-b43fcc536c7ffa370ea5d043988d111223780b2dade0198d7b91818f9bd7ab62

MGatner (approved):
Docs look good.

iRedds:
As far as I remember, PSR does not recommend using the underscore character as a prefix for method names.
I have already expressed my opinion about the names of methods for accessing attributes more than once.
get\<name>Attribute

kenjis:
We don't remove `get<name>()` getter.
So `get<name>Attribute()` may conflict to another column.

If there are columns `parent` and `parent_attribute` and
`getParent()` and `getParentAttribute()`, 
`getParentAttribute()` is used when getting `parent`.
This is a breaking change.

iRedds:
I just expressed my opinion.
The fact that some genius came up with the idea of using class property getter/setter method names for attributes, which gave rise to this garbage, is not my problem.
It seems that they were trying to combine the Doctrine and Eloquent.

I see two options.
1. We admit the problem, apologize and change the entity class with normal method names.
2. We add an alternative Entity that will be maintained further, and mark the existing one as deprecated.

PR #7239, opened 2023-02-10
feat: add URIFactory

kenjis:
Needs #7232

**Description**
Add `URIFactory` to create the current URI object.

To create the current URI object before the Request object. Also, to remove all URI adjustment processing currently performed in the constructor of the Request object.

This PR does not change any existing features yet.

- add `URIFactory::createFromGlobals()`
- add `URI::setRoutePath()` and `URI::getRoutePath()`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/URIFactory.php
+++ b/system/HTTP/URIFactory.php
@@ -51,61 +51,68 @@ class URIFactory
     }
 
     /**
-     * Detects the relative path based on
-     * the URIProtocol Config setting.
+     * Detects the current URI path relative to baseURL based on the URIProtocol
+     * Config setting.
+     *
+     * @param string $protocol URIProtocol
+     *
+     * @return string The route path
      */
-    public function detectPath(string $protocol = ''): string
+    public function detectRoutePath(string $protocol = ''): string
     {
-        if (empty($protocol)) {
-            $protocol = 'REQUEST_URI';
+        if ($protocol === '') {
+            $protocol = $this->appConfig->uriProtocol;
         }
 
         switch ($protocol) {
             case 'REQUEST_URI':
-                $this->path = $this->parseRequestURI();
+                $routePath = $this->parseRequestURI();
                 break;
 
             case 'QUERY_STRING':
-                $this->path = $this->parseQueryString();
+                $routePath = $this->parseQueryString();
                 break;
 
             case 'PATH_INFO':
             default:
-                $this->path = $this->fetchGlobal('server', $protocol) ?? $this->parseRequestURI();
+                $routePath = $this->server[$protocol] ?? $this->parseRequestURI();
                 break;
         }
 
-        return $this->path;
+        return $routePath;
     }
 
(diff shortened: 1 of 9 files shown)


system/HTTP/URIFactory.php
```diff
@@ -0,0 +1,248 @@
...
+    private App $appConfig;
+
+    /**
+     * @param array $server Superglobal $_SERVER array
+     * @param array $get    Superglobal $_GET array
+     */
+    public function __construct(array &$server, array &$get, App $appConfig)
+    {
+        $this->server    = &$server;
+        $this->get       = &$get;
```
iRedds, line 38: I don't think passing by reference, which means modifying superglobal arrays, is the right approach.

iRedds:
In my opinion, two factories are needed
``\CodeIgniter\HTTP\URI::createFromRequest(Request $request)`` and ``\CodeIgniter\HTTP\URI::createFromString(string $url)``

To create a separate class that works directly with ``_SERVER`` and ``_GET``, and even created before the request object, this is not correct in my opinion.

kenjis:
How do you create Request object without URI object?

iRedds:
> How do you create Request object without URI object?

Say hello to the framework architects for the URI class in the Request constructor.
For the same reason, in other classes, _SERVER is also directly accessed.

kenjis:
> Say hello to the framework architects for the URI class in the Request constructor.
For the same reason, in other classes, _SERVER is also directly accessed.

Sorry, I don't get what you say at all. 
Who are the framework architects?
What do you mean by say hello?

iRedds:
The URI class depends on the Request class for incoming requests because the Request class must provide access to a copy of _SERVER.

Therefore, making the Request class dependent on the URI is a framework architecture error.

I don't know who designed these classes.
I meant that the question "How do you create Request object without URI object?" should be forwarded to those who designed.

kenjis:
> The URI class depends on the Request class for incoming requests because the Request class must provide access to a copy of _SERVER.

Indeed.

> I meant that the question "How do you create Request object without URI object?" should be forwarded to those who designed.

Who designed does not matter.
The only question is how do we redesign.

iRedds:
> Who designed does not matter. The only question is how do we redesign.

I don't know.
There are two abstract options.
1. Qualitatively, but this is BC at the level of major versions.
2. If only it worked. Remove the restriction on $uri = null in the IncomingRequest constructor. And when null, use a request-based factory.
https://github.com/codeigniter4/CodeIgniter4/blob/6526113e9eb0add6d7a7cd05823c07023071496f/system/HTTP/IncomingRequest.php#L153-L155

```php
if (empty($userAgent)) {
    throw new InvalidArgumentException('You must supply the parameters: uri, userAgent.');
}

$this->uri = $uri ?? URI::createFromRequest($this);
```

PR #7240, opened 2023-02-10
fix: convert object to array for insert and update when tempdata is empty

### LINKED ISSUE #7238: Bug: Cannot use object of type App\Entities\[Custom Entity] as array
### PHP Version

8.2

### CodeIgniter4 Version

4.2.11

Linux

### Which server did you use?

fpm-fcgi

### Database

MariaDB 10.10.2

### What happened?

I have a custom entity created for my Customer object.  My CustomerModel is returning this type.  When attempting to use the $customerModel->save($customerEntity)
Received an error of

Cannot use object of type App\Entities\Customer as array  Model.php at line 688

### Steps to Reproduce

1. Create an entity for a model
2. set the model to use a custom primary (my primary id is set to 'uuid')
3. populate the entity with the $customerEntity->fill($this->request->getPost())
4. save the model ($customerModel->save($customerEntity))

### Expected Output

Model should save the entity object by converting it to an array and then saving it with the ->insert() function.
```php
public function insert($data = null, bool $returnID = true)
    {
        if (!empty($this->tempData['data'])) { <- ***at this step my tempData array is empty***
            if (empty($data)) {
                $data = $this->tempData['data'];
            } else {
                $data = $this->transformDataToArray($data, 'insert'); <- ***this step never happens***
                $data = array_merge($this->tempData['data'], $data);
            }
        }

        if ($this->useAutoIncrement === false && isset($data[$this->primaryKey])) { <- ***checks for an array key that is still an entity object and fails ***
            $this->tempPrimaryKeyValue = $data[$this->primaryKey];
        }

        $this->escape   = $this->tempData['escape'] ?? [];
        $this->tempData = [];

        return parent::insert($data, $returnID);
    }
```

### Anything else?

Although I did not test an update, it appears to have the same issue.

This code should fix the issue with both the insert and the update function in the Model.php class

```php
    /**
     * Inserts data into the database. If an object is provided,
     * it will attempt to convert it to an array.
     *
     * @param array|object|null $data
     * @param bool              $returnID Whether insert ID should be returned or not.
     *
     * @return BaseResult|false|int|object|string
     *
     * @throws ReflectionException
     */
    public function insert($data = null, bool $returnID = true)
    {
        if (!empty($this->tempData['data'])) {
            if (empty($data)) {
                $data = $this->tempData['data'];
            } else {
                $data = $this->transformDataToArray($data, 'insert');
                $data = array_merge($this->tempData['data'], $data);
            }
        } else {
            $data = $this->transformDataToArray($data, 'insert');
        }

        if ($this->useAutoIncrement === false && isset($data[$this->primaryKey])) {
            $this->tempPrimaryKeyValue = $data[$this->primaryKey];
        }

        $this->escape   = $this->tempData['escape'] ?? [];
        $this->tempData = [];

        return parent::insert($data, $returnID);
    }

    /**
     * Updates a single record in the database. If an object is provided,
     * it will attempt to convert it into an array.
     *
     * @param array|int|string|null $id
     * @param array|object|null     $data
     *
     * @throws ReflectionException
     */
    public function update($id = null, $data = null): bool
    {
        if (!empty($this->tempData['data'])) {
            if (empty($data)) {
                $data = $this->tempData['data'];
            } else {
                $data = $this->transformDataToArray($data, 'update');
                $data = array_merge($this->tempData['data'], $data);
            }
        } else {
            $data = $this->transformDataToArray($data, 'update');
        }

        $this->escape   = $this->tempData['escape'] ?? [];
        $this->tempData = [];

        return parent::update($id, $data);
    }

```

url84t:
**Description**
Fixes #7238

entity object was not converting to an array when tempdata was empty.  Added the option to convert object to array if the tempdata['data'] was empty in both the insert and update functions in system/Model.php class file

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Model.php
+++ b/system/Model.php
@@ -713,6 +713,8 @@ class Model extends BaseModel
                 $data = $this->transformDataToArray($data, 'insert');
                 $data = array_merge($this->tempData['data'], $data);
             }
+        } else {
+            $data = $this->transformDataToArray($data, 'insert');
         }
 
         if ($this->useAutoIncrement === false) {
@@ -747,8 +749,9 @@ class Model extends BaseModel
                 $data = $this->transformDataToArray($data, 'update');
                 $data = array_merge($this->tempData['data'], $data);
             }
+        } else {
+            $data = $this->transformDataToArray($data, 'update');
         }
-
         $this->escape   = $this->tempData['escape'] ?? [];
         $this->tempData = [];
 


kenjis:
Thank you for sending PR!

You missed GPG sign on your commits.

We expect all code changes or bug-fixes to be accompanied by one or more tests added to our test suite to prove the code works.

system/Model.php
```diff
@@ -747,8 +749,9 @@ public function update($id = null, $data = null): bool
                 $data = $this->transformDataToArray($data, 'update');
                 $data = array_merge($this->tempData['data'], $data);
             }
+        } else {
+            $data = $this->transformDataToArray($data, 'update');
         }
-
```
kenjis, line 751: Please revert this empty line.

kenjis:
The bug was fixed in v4.2.12.

PR #7249, opened 2023-02-14
fix: URI::getSegment() returns non-existent segment

### LINKED ISSUE #7246: URI::getSegment() returns non-existent segment
CodeIgniter version: 4.3.1

In this sample, there are three URI segments: 1:`users`, 2:`15`, 3:`profile`
But you can get the 4th segment `$uri->getSegment(4, 'bar')` without Exception.
But `$uri->getSegment(5, 'baz')` will throw Exception.

```php
// URI = http://example.com/users/15/profile

// will print 'profile'
echo $uri->getSegment(3, 'foo');
// will print 'bar'
echo $uri->getSegment(4, 'bar');
// will throw an exception
echo $uri->getSegment(5, 'baz');
// will print 'baz'
echo $uri->setSilent()->getSegment(5, 'baz');
// will print '' (empty string)
echo $uri->setSilent()->getSegment(5);
```
https://codeigniter4.github.io/CodeIgniter4/libraries/uri.html#uri-segments
 
Related: #7233

kenjis:
**Description**
Fixes #7246 (and #2962)

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -537,17 +537,21 @@ class URI
      * @param string $default Default value
      *
      * @return string The value of the segment. If no segment is found,
-     *                throws InvalidArgumentError
+     *                throws HTTPException
      */
     public function getSegment(int $number, string $default = ''): string
     {
-        // The segment should treat the array as 1-based for the user
-        // but we still have to deal with a zero-based array.
-        $number--;
+        if ($number < 1) {
+            throw HTTPException::forURISegmentOutOfRange($number);
+        }
 
         if ($number > count($this->segments) && ! $this->silent) {
             throw HTTPException::forURISegmentOutOfRange($number);
         }
+
+        // The segment should treat the array as 1-based for the user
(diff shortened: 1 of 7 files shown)


michalsn:
Sorry, by accident I wrote the comment under the issue report: https://github.com/codeigniter4/CodeIgniter4/issues/7246#issuecomment-1430124894

MGatner (approved):

kenjis:
Since michalsn is opposed, this PR is put on hold.
https://github.com/codeigniter4/CodeIgniter4/issues/7246#issuecomment-1431660158

michalsn:
I was pointing out the potential problem that this change brings.

If everyone wants to treat it as a bug, I'm fine with that. In my opinion, it will be a bit controversial, but we have already made bigger BC changes in the past.

kenjis:
I am going to withdraw this BC as it does not seem to be worth much.

PR #7251, opened 2023-02-14
fix: URI::setSegment() accepts the last +2 segment without Exception

### LINKED ISSUE #7250: Bug: URI::setSegment() accepts the last +2 segment without Exception
CodeIgniter version: develop (4a4f55b1308cf9f9e22f19be84ce444d69263c2d) after 4.3.1

This is also reported in #2962

```php
<?php

namespace App\Controllers;

use CodeIgniter\HTTP\URI;

{
    public function index()
    {
        $uri = new URI('http://example.com/one/two');

        $uri->setSegment(4, 'four');
        var_dump($uri->getSegments());
        echo $uri;
    }
}
```
The result:
```
array (size=3)
  0 => string 'one' (length=3)
  1 => string 'two' (length=3)
  2 => string 'four' (length=4)

http://example.com/one/two/four 
```

kenjis:
**Description**
Fixes #7250

Related #7249

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/URI.php
+++ b/system/HTTP/URI.php
@@ -562,9 +562,9 @@ class URI
      */
     public function setSegment(int $number, $value)
     {
-        // The segment should treat the array as 1-based for the user
-        // but we still have to deal with a zero-based array.
-        $number--;
+        if ($number < 1) {
+            throw HTTPException::forURISegmentOutOfRange($number);
+        }
 
         if ($number > count($this->segments) + 1) {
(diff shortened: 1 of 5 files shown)


michalsn (approved):
It seems correct to me. We just need to add this to the changelog.

kenjis:
Added docs.

MGatner (approved):

Issue #7123, opened 2023-01-15
Bug: $this->request->getUri()->getPath() is not compatible with PSR-7


kenjis:
CodeIgniter4 Version: v4.3.1

`$this->request->getUri()->getPath()` returns the route (URI path relative to baseURL).

```
app.baseURL = 'http://localhost:8888/ci431/public/'
```
When navigating to `http://localhost:8888/ci431/public/test?a=b`,

```php
$uri = $this->request->getUri();
echo (string) $uri;   // "http://localhost:8888/ci431/public/test?a=b" → Okay
echo $uri->getPath(); // "test" → NG. It should be "/ci431/public/test" when following PSR-7
```

Ref https://www.php-fig.org/psr/psr-7/

Related #5930

kenjis:
>  I will push again for a separate class to handle internal URIs; I think we need it despite the complexity it adds.
https://github.com/codeigniter4/CodeIgniter4/issues/7124#issuecomment-1383940519

kenjis:
Navigate to `http://localhost:8888/ci431/public/test?a=b`:
```
    uri_string(): test
uri_string(true): test

               current_url(): http://localhost:8888/ci431/public/index.php/test
current_url(true)->getPath(): /ci431/public/index.php/test
  (string) current_url(true): http://localhost:8888/ci431/public/index.php/test?a=b

  (string) $this->request->getUri(): http://localhost:8888/ci431/public/test?a=b
$this->request->getUri()->getPath(): test
```

kenjis:
Navigate to `http://localhost:8888/ci431/public/index.php/test?a=b`:
```
    uri_string(): test
uri_string(true): test

               current_url(): http://localhost:8888/ci431/public/index.php/test
current_url(true)->getPath(): /ci431/public/index.php/test
  (string) current_url(true): http://localhost:8888/ci431/public/index.php/test?a=b

  (string) $this->request->getUri(): http://localhost:8888/ci431/public/test?a=b
$this->request->getUri()->getPath(): test
```

kenjis:
`current_url()` returns the URL with `index.php`.
But `$this->request->getUri()` does not have `index.php`.

I think when `current_url()` has `index.php`, the URI object also should have `index.php`.

kenjis:
From https://github.com/codeigniter4/CodeIgniter4/pull/7233#issuecomment-1426700790

The inconsistency of URI Segments between `$this->request->getUri()` and `current_url(true)`.

[...]

Issue #5930, opened 2022-04-28
Dev: Project URI Parser


MGatner:
We have a lot of places where values can be interpolated into a (project) URI. There are also some places that it would be good to add support for such URIs. 

I would like to add a new class that handles these centrally, probably a child of URI and maybe merged with the idea of a “project-specific” URI class. 

See also: https://github.com/codeigniter4/CodeIgniter4/issues/5923#issuecomment-1110962160

iRedds:
Why do we need a new class if the URI class is already designed to parse the url?
Just add a URI::is($uri) method that will check for a match.

lonnieezell:
I would agree that I'm not a fan of creating another URI class. I feel that could be confusing. Granted, I'm also a little confused on what you mean by "project specific" URI class.

kenjis:
I also don't understand what exactly "project specific" URI and "URI meta-segment transformation" are.

MGatner:
So clearly I haven't explained this well 😅 Maybe "parse" isn't the right word: translate? interpret?

So within Router we have the ability to define a URI like `{locale}/foo/bar` - this is understood as "a URI within this project using the current locale". What I would like is two parts:
1. Separate handling for "internal URIs" (this has been discussed lots, see project roadmap and closed PRs)
2. The ability to use these "segment variables" like `{locale}` anywhere in the framework. 

Does that help?

kenjis:
In CodeIgniter context, *URI* has two meaning.
1. generally speaking URI like `https://codeigniter.com/controller/method`
2. The URI path after `index.php` like `controller/method`

This class handles 2, doesn't it?

lonnieezell:
Why does this need a new class? Can't we add placeholder support to the existing URI class and use that? I would expect that info to be available on the one that the IncomingRequest class has.

kenjis:
I don't know if this is the same as Project URI Parser, but I have come to the view that it would be better to have `SiteURL` class that extends the `URI` class representing the site (current) URL.

MGatner:
Yes, that's precisely the point. Though the intent was to handle more than just the _current_ URL, but also routes, etc

kenjis:
Yes, `{locale}/foo/bar` is kind of URI path in the project, and it is corresponding to the real route path `en/foo/bar`.
However, I don't know how SiteURI relates to this route or Routing classes..

lonnieezell:
Very stale. Closing.

PR #7252, opened 2023-02-14
feat: add SiteURI class

kenjis:
**Description**
This PR is needed for #7123

We need two URI classes. One for general purpose URI class, and the other is for URI of the app site.

- add SiteURI class
  - The `$routePath` never starts with `/`.
  - If the path ends with `/`, the URI retains the trailing `/`.
  - `getPath()` returns the full URI path.
  - `getRoutePath()` returns the path relative to baseURL.

Related #5930, #7239, #7249, #7251

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


system/HTTP/SiteURI.php
```diff
@@ -0,0 +1,312 @@
...
+        }
+        $this->applyParts($parts);
+
+        $this->setPath('/');
+    }
+
+    /**
+     * Sets baseSegments.
+     */
+    private function setBaseSeegments(): void
```
samsonasik, line 108: Seegments -> Segments

```suggestion
    private function setBaseSegments(): void
```

michalsn:
Maybe this is a silly question, but... as I understand, this new class will bring full PSR-7 compatibility, right?

I don't see any related changes in the framework so we won't rely on it. The question is, who needs this class and for what?

Like... I saw your description:

> We need two URI classes. One for general purpose URI class, and the other is for URI of the app site.

But I'm afraid I don't fully understand the use cases.

system/HTTP/SiteURI.php
kenjis: Thanks! Fixed.

kenjis:
This PR only adds a new SiteURI class, and does not change the current framework behavior yet.

In the next PR(s), I will do:
- add a factory to create the SiteURI instance for the current URL
- set the SiteURI instance in the `uri` property of the Request
- use the SiteURI instance when getting the current URL like `current_url()`

About compatibility with PSR-7, the URI class will be cleaner because site-related functions like baseURL, will be separated into the SiteURI class. And the URI class will be easier to change.

Finally, I think `URI` and `SiteURL` can be made fully compatible (and having some extended functions) with PSR-7.

MGatner (commented):
I am obviously in favor of this but I want to be sure we get reviews from @lonnieezell and @iRedds and anyone else who was opposed to the idea originally.

lonnieezell:
I was purposefully not commenting since my position was already known and it seemed I was overruled :) 

I'm confused why we need a second class that does 99% of the same things as the URI class. The URI class is just for working with URIs in general, crafted to meet the RFC as best we could at the time. It seems like all other issues should be fixed where it's used, or the base class should be changed to support our needs. 

The implementation itself is solid enough. 

It seems strange to me to make this new class and we're not using it ourselves in `IncomingRequest` or the URL helpers. What is the point of this other then to throw confusion about which class people should use when? I figured it was to fix issues in how the framework was returning answers, but it doesn't seem like we're actually doing that.

michalsn:
I agree with @lonnieezell. I don't understand why we need a second class to handle URLs. Can somebody give me a clear example of where it will be useful?

AFAIK the only issue that we have now with the URI class is that it does not follow PSR-7 (it's something path related).

Also, if the goal for both classes is to follow the PSR-7, what will be the difference? Wouldn't it be better to see some additional methods added to the current class instead?

kenjis:
> I don't understand why we need a second class to handle URLs.

Because we have already two kinds of concept URI,
and if we have two classes, It is easier to understand for us.

1. URI: a URI
2. SiteURI: a URI for a CI4 app site.

SiteURI has its own properties like `baseURL`, `indexPage` and we cannot change the URI path
inside `baseURL` (subfolders).

Now we are already confusing 1. and 2. So we have two kind of URI segments for one SiteURL.
See https://github.com/codeigniter4/CodeIgniter4/issues/7123#issuecomment-1426701327

kenjis:
> It seems strange to me to make this new class and we're not using it ourselves in IncomingRequest or the URL helpers. 

This fix (and refactoring) will be long story. This PR is just the first one.
We will use the new class in all places after all. Please wait for incoming PRs.

MGatner:
I won't repeat my arguments, the original PR is here with discussion: https://github.com/codeigniter4/CodeIgniter4/pull/4647

I would like to point out that after doing a deep dive on URI handling and bug fixes both @kenjis and I came to the same conclusion, independently. I think the problem is larger than credited and may indeed merit the potential confusion/clutter of an additional class.

lonnieezell:
After giving this a bit more thought, I think I'm ok with this on 2 conditions: 

1) It's designed for internal use within the request class and URL helper functions. Not intended for public use or really even documented. This avoids the confusion to end users, while still providing a "URI in Request is the source of truth" that @MGatner was looking for in the previous PR. 
2) It's saved for 5.0 (which we should be working on now....) since it's too big of a potential BC break.

iRedds:
```php
$uri = $this->request->getUri();
echo (string) $uri;   // "http://localhost:8888/ci431/public/test?a=b" → Okay
echo $uri->getPath(); // "test" → NG. It should be "/ci431/public/test" when following PSR-7
```
And I think that such behavior corresponds to PSR.
```php
// for http://localhost:8888/ci431/public/test?a=b
URI::getPath(); // => /ci431/public/test

// for http://localhost:8888/{base_path}/test?a=b
URI::getPath(); // => test    (without leading slash, as an indicator that the base path is being used.)

// Accordingly, the leading slash will determine whether the path is absolute or rootless

// for http://localhost:8888/{base_path}/test?a=b
URI::withPath('/xxx'); http://localhost:8888/xxx?a=b

// for http://localhost:8888/{base_path}/test?a=b
URI::withPath('xxx'); http://localhost:8888/{base_path}/xxx?a=b
```
https://github.com/php-fig/fig-standards/pull/503

kenjis:
What is `{base_path}` in this case? How do you define it?
```php
// for http://localhost:8888/{base_path}/test?a=b
URI::getPath(); // => test    (without leading slash, as an indicator that the base path is being used.)
```

iRedds:
@kenjis app.baseURL = 'http://localhost:8888{/ci431/public/}' <- base path

kenjis:
I don't get it.
How do you set it in PSR-7 URI object?

iRedds:
There are no objects in PSR-7. Interfaces only. That is, mandatory public methods for PSR compliance.
A class implementing UriInterface should not be limited to UriInterface methods and may contain a method that sets base_path or base_url or whatever is needed for it to work properly.

kenjis:
> Many frameworks provide the ability to get the "base path," usually considered the path up to and including the front controller. As an example, if the application is served at http://example.com/b2b/index.php, and the current URI used to request it is http://example.com/b2b/index.php/customer/register, the functionality to retrieve the base path would return /b2b/index.php. This value can then be used by routers to strip that path segment prior to attempting a match.
https://www.php-fig.org/psr/psr-7/meta/#why-is-no-functionality-included-for-retrieving-the-base-path

>
     * If an HTTP path is intended to be host-relative rather than path-relative
     * then it must begin with a slash ("/"). HTTP paths not starting with a slash
     * are assumed to be relative to some base path known to the application or
     * consumer.
       ...
     */
    public function withPath($path);
https://www.php-fig.org/psr/psr-7/

This means like this?
- if we call `URI::withPath('foo/bar')`, the path `foo/bar` is relative to some base path known to the application.
- the new URI instance `getPath()` will return `foo/bar`?

kenjis:
```php
<?php

require __DIR__ . '/vendor/autoload.php';

use League\Uri\Http;

$uri = Http::createFromBaseUri('foo/bar', 'http://localhost/ci4/');
echo $uri . PHP_EOL;            // http://localhost/ci4/foo/bar
echo $uri->getPath() . PHP_EOL; // /ci4/foo/bar
```
https://uri.thephpleague.com/uri/6.0/psr7/

kenjis:
```php
$uri = Http::createFromBaseUri('foo/bar', 'http://localhost/ci4/');
$uri2 = $uri->withPath('controller/method');
```
> PHP Fatal error:  Uncaught League\Uri\Exceptions\SyntaxError: If an authority is present the path must be empty or start with a `/`. in /.../league-uri/vendor/league/uri/src/Uri.php:921

kenjis:
```php
use Slim\Psr7\Factory\UriFactory;

$factory = new UriFactory();

$uri = $factory->createUri('http://localhost/ci4/');
$uri2 = $uri->withPath('controller/method');
echo $uri2 . PHP_EOL;            // http://localhost/controller/method
echo $uri2->getPath() . PHP_EOL; // controller/method

$uri = $factory->createUri('http://localhost/ci4/');
$uri2 = $uri->withPath('/controller/method');
echo $uri2 . PHP_EOL;            // http://localhost/controller/method
echo $uri2->getPath() . PHP_EOL; // /controller/method
```

kenjis:
```php
use Nyholm\Psr7\Uri;

$uri = new Uri('http://localhost/ci4/');
$uri2 = $uri->withPath('controller/method');
echo $uri2 . PHP_EOL;            // http://localhost/controller/method
echo $uri2->getPath() . PHP_EOL; // controller/method

$uri2 = $uri->withPath('/controller/method');
echo $uri2 . PHP_EOL;            // http://localhost/controller/method
echo $uri2->getPath() . PHP_EOL; // /controller/method
```

kenjis:
@iRedds 
Do you mean like this?
```php
$uri = new Uri('http://example.com/');
$uri2 = $uri->withPath('foo/bar');
echo $uri2->getPath(); // empty string ''
```

iRedds:
@kenjis you were talking about a new instance

kenjis:
@iRedds I meant the new instance.

Is this okay?
```php
$uri = new Uri('http://example.com/');
$uri2 = $uri->withPath('foo/bar');
echo $uri2->getPath(); // 'foo/bar'
```

kenjis:
The Behavior has been modified slightly, and test are refactored.

kenjis:
This is kind of a big class, but for me, this PR is OK.

Note that the next PR is #7256 and the last PR is #7282.

kenjis:
Regarding making PSR-7 compatible in the future, I am not sure how I would implement this.

From reading the spec, I think it is safe for `getPath()` to return relative paths excluding the base path. 
However, I could not find an implementation that handles the base path.

kenjis:
The rework on the URI has been completed.
This is the first PR.
The next PR is #7256 and the last PR is #7282.

Reviews are welcome.

MGatner:
What's the status here?

kenjis:
Done, ready for reviews.

And the whole story is #7282. It includes this PR and  next #7256.

TimexPeachtree (approved):
Awesome 👍😎

system/HTTP/SiteURI.php
```diff
@@ -0,0 +1,366 @@
...
+    /**
+     * The Index File.
+     */
+    private string $indexPage;
+
+    /**
+     * List of URI segments in baseURL and indexPage.
+     *
+     * If the URI is "http://localhost:8888/ci431/public/index.php/test?a=b",
+     * and the baseUR is "http://localhost:8888/ci431/public/", then:
```
MGatner, line 46:
```suggestion
     * and the baseURL is "http://localhost:8888/ci431/public/", then:
```

system/HTTP/SiteURI.php
```diff
@@ -0,0 +1,366 @@
...
+    private array $baseSegments;
+
+    /**
+     * List of URI segments after indexPage.
+     *
+     * The word "URI Segments" originally means only the URI path part relative
+     * to the baseURL.
+     *
+     * If the URI is "http://localhost:8888/ci431/public/index.php/test?a=b",
+     * and the baseUR is "http://localhost:8888/ci431/public/", then:
```
MGatner, line 62:
```suggestion
     * and the baseURL is "http://localhost:8888/ci431/public/", then:
```

system/HTTP/SiteURI.php
```diff
@@ -0,0 +1,366 @@
...
+     *   $segments = [
+     *       0 => 'test',
+     *   ];
+     *
+     * @var array
+     *
+     * @deprecated This property will be private.
```
MGatner, line 69: Why does it need to be protected to begin with?

MGatner: Ohhh because of parent! Got it.

system/HTTP/URI.php
```diff
@@ -259,6 +263,8 @@ public function __construct(?string $uri = null)
      * If $silent == true, then will not throw exceptions and will
      * attempt to continue gracefully.
      *
+     * Note: Method not in PSR-7
```
MGatner, line 266: Let's deprecate this. Conditional exceptions is a really terrible design. 
```suggestion
     * @deprecated Method not in PSR-7
```

system/HTTP/URI.php
```diff
@@ -690,6 +708,8 @@ public function setAuthority(string $str)
      * @see https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
      *
      * @return $this
+     *
+     * @TODO PSR-7: Should be `withScheme($scheme)`.
```
MGatner, line 712: Why not go ahead and add the new methods as aliases and deprecate these?

MGatner (changes requested):
Looks great! I feel validated in the need for this given how much code is in SiteURI on top of URI. I appreciate your verbosity in naming variables and methods - it is easy to overlook pieces which has led to many bugs in the past. I think this will be a good step forward. 

I assume the URL Helper methods get changed in the next PR? I haven't looked at that one yet...

system/HTTP/SiteURI.php
kenjis: Fixed.

system/HTTP/SiteURI.php
kenjis: Fixed.

system/HTTP/URI.php
kenjis: Done.

system/HTTP/URI.php
kenjis: Added.

MGatner (approved):

kenjis:
@TimexPeachtree @MGatner Thank you for approvals.

TimexPeachtree (approved):

PR #7247, opened 2023-02-13
feat: cors filter

yasinpalizban:
**Description**
  CI4  needs to pass cors for developing rest full api application with  , so  by using middleware we can handle  problem easily. i already  test cors filter and it's   works in development and production environment.

PR #7253, opened 2023-02-14
feat: cors filter

yasinpalizban:
Supersedes #7247

**Description**
i add  cors  middleware  to  default filter CI4 and there  is a config file which can easily modify settings. 
all we have to do  go to filter.php  and un comment  cors  that's all needs.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -3,6 +3,7 @@
 namespace Config;
 
 use CodeIgniter\Config\BaseConfig;
+use CodeIgniter\Filters\Cors;
 use CodeIgniter\Filters\CSRF;
 use CodeIgniter\Filters\DebugToolbar;
 use CodeIgniter\Filters\Honeypot;
@@ -21,6 +22,7 @@ class Filters extends BaseConfig
         'honeypot'      => Honeypot::class,
         'invalidchars'  => InvalidChars::class,
         'secureheaders' => SecureHeaders::class,
+        'cors'          =>Cors::class
     ];
 
     /**
@@ -32,6 +34,7 @@ class Filters extends BaseConfig
             // 'honeypot',
             // 'csrf',
             // 'invalidchars',
+            // 'cors' check
         ],
         'after' => [
             'toolbar',
--- a/system/Filters/Cors.php
+++ b/system/Filters/Cors.php
@@ -1,0 +1,62 @@
+<?php namespace CodeIgniter\Filters;
+
+
+use CodeIgniter\HTTP\RequestInterface;
+use CodeIgniter\HTTP\ResponseInterface;
+
+
+class Cors implements FilterInterface
+{
+    public function after(RequestInterface $request, ResponseInterface $response, $arguments = null)
+    {
+
+
+    }
+
+    public function before(RequestInterface $request, $arguments = null)
+    {
+
+
+            // get origin
+        if (array_key_exists('HTTP_ORIGIN', $_SERVER)) {
+            //get origin when it is using  front end framework angular or react
+            $origin = $_SERVER['HTTP_ORIGIN'];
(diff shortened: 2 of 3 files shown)


kenjis:
Thank you for the update.

You must GPG-sign your commits.

We need unit testing.

app/Config/Filters.php
```diff
@@ -32,6 +34,7 @@ class Filters extends BaseConfig
             // 'honeypot',
             // 'csrf',
             // 'invalidchars',
+            // 'cors' check
```
kenjis, line 37:
```suggestion
            // 'cors'
```

system/Filters/Cors.php
```diff
@@ -0,0 +1,62 @@
...
+            // get origin when it is  using Postman
+            $origin = $_SERVER['REMOTE_ADDR'];
+        }
+
+            // get  cors configuration
+        $corsConfig = config('Cors');
+
+        //  if origin is not in list , origin header no back in response object  and front end  framework get cors error
+        if (in_array($origin, $corsConfig->validDomains)) {
+            header('Access-Control-Allow-Origin: ' . $origin);
```
kenjis, line 37: Fix it so that header() is not used.

system/Filters/Cors.php
```diff
@@ -0,0 +1,62 @@
...
+        header("Access-Control-Allow-Credentials: {$corsConfig->credentials}");
+        // append set  max age
+        header("Access-Control-Max-Age: {$corsConfig->maxAge}");
+        // append content type
+        header('content-type: ' . implode("; ", $corsConfig->contentType));
+
+
+        // if request is option call back to browser 200 ok
+        if ($request->getMethod() == "options") {
+            header("HTTP/1.1 200 OK CORS");
```
kenjis, line 55: Fix it so that header() is not used.

system/Filters/Cors.php
```diff
@@ -0,0 +1,62 @@
...
+
```
kenjis, line 56: Please return a Response object.
See https://codeigniter4.github.io/CodeIgniter4/incoming/filters.html#before-filters

system/Filters/Cors.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+
+
+            // get origin
+        if (array_key_exists('HTTP_ORIGIN', $_SERVER)) {
+            //get origin when it is using  front end framework angular or react
+            $origin = $_SERVER['HTTP_ORIGIN'];
+        } else if (array_key_exists('HTTP_REFERER', $_SERVER)) {
+            //  get origin when it is not using  front end  only browser
+            $origin = $_SERVER['HTTP_REFERER'];
```
kenjis, line 26: Can you explain why do you use `HTTP_REFERER` as origin?

system/Filters/Cors.php
```diff
@@ -0,0 +1,62 @@
...
+        } else {
+            // get origin when it is  using Postman
+            $origin = $_SERVER['REMOTE_ADDR'];
```
kenjis, line 29: Can you explain why do you use `REMOTE_ADDR` as origin?

kenjis:
This library would be a good reference for configuration.
https://github.com/fruitcake/php-cors
I am not sure if we need all the features.

cijagani:
Here is one more example
https://github.com/agungsugiarto/codeigniter4-cors

@agungsugiarto may help on this

kenjis:
@cijagani The package are using another vendor's namespace `Fluent`.
https://packagist.org/packages/fluent/
So I don't recommend it.

cijagani:
Agree. We should consider as reference or we should compare features.

PR #7255, opened 2023-02-15
Remove Config\App Session items

kenjis:
**Description**
Supersedes #7000

This PR is needed for #7308
Session should not depends on `Config\App`. See https://github.com/codeigniter4/CodeIgniter4/issues/4297#issuecomment-813110373

- remove `Config\App` session items
- add property for `Config\Session` in `Session`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -3,7 +3,6 @@
 namespace Config;
 
 use CodeIgniter\Config\BaseConfig;
-use CodeIgniter\Session\Handlers\FileHandler;
 
 class App extends BaseConfig
 {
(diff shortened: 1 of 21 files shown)


mostafakhudair:
Shall we add default config properties in `system/Session/Session.php` to avoid breaking app?, what if the developer removed some properties from `app/Config/Session.php`?

kenjis:
> what if the developer removed some properties from app/Config/Session.php?

It breaks. 
If a dev removes `Config\App::$baseURL` now, the framework does not work.
It is not designed to work by removing necessary Config properties.

I don't see the significance of adding properties that just copies the properties of the Config class.

kenjis:
Rebased to resolve conflicts.

Can anyone review?

kenjis:
Rebased.

iRedds:
This PR tries to get rid of the dependency on Config\App in Services::session() to prevent an infinite loop if the Config\App constructor is accessing a session.
But Services::session() depends on Services::request() which depends on Config\App.

iRedds:
Oh, I must have missed that comment. My bad.

kenjis:
Rebased to resolve conflicts.

kenjis:
I would like to merge this. Review, please.

paulbalandan (approved):

PR #7256, opened 2023-02-15
feat: add SiteURIFactory

kenjis:
~~Needs #7252~~

**Description**
Supersedes #7239

Add `SiteURIFactory` to create the current URI object.

To create the current URI object before the Request object creation. 
Also, to remove all URI adjustment processing currently performed in the constructor of the Request object.

This PR does not change any existing features yet. The next #7282 changes the behavior.

- add `SiteURIFactory::createFromGlobals()`
- add `Superglobals`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/SiteURIFactory.php
+++ b/system/HTTP/SiteURIFactory.php
@@ -1,0 +1,251 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\HTTP;
+
+use CodeIgniter\HTTP\Exceptions\HTTPException;
+use Config\App;
+
+class SiteURIFactory
+{
+    /**
+     * @var array Superglobal SERVER array
+     */
+    private array $server;
+
+    private App $appConfig;
+
+    /**
+     * @param array $server Superglobal $_SERVER array
+     */
+    public function __construct(array $server, App $appConfig)
+    {
+        $this->server    = $server;
+        $this->appConfig = $appConfig;
+    }
+
+    /**
+     * Create the current URI object from superglobals.
+     *
+     * This method updates superglobal $_SERVER and $_GET.
+     */
+    public function createFromGlobals(): SiteURI
+    {
+        $routePath = $this->detectRoutePath();
+
+        return $this->createURIFromRoutePath($routePath);
+    }
+
+    /**
+     * Create the SiteURI object from URI string.
+     *
+     * @internal Used for testing purposes only.
+     */
+    public function createFromString(string $uri): SiteURI
+    {
+        // Validate URI
+        if (filter_var($uri, FILTER_VALIDATE_URL) === false) {
+            throw HTTPException::forUnableToParseURI($uri);
+        }
+
+        $parts = parse_url($uri);
+
+        if ($parts === false) {
+            throw HTTPException::forUnableToParseURI($uri);
+        }
+
+        $query = $fragment = '';
+        if (isset($parts['query'])) {
+            $query = '?' . $parts['query'];
+        }
+        if (isset($parts['fragment'])) {
+            $fragment = '#' . $parts['fragment'];
+        }
+
+        $relativePath = $parts['path'] . $query . $fragment;
+
+        return new SiteURI($this->appConfig, $relativePath, $parts['host'], $parts['scheme']);
+    }
+
+    /**
+     * Detects the current URI path relative to baseURL based on the URIProtocol
+     * Config setting.
+     *
+     * @param string $protocol URIProtocol
+     *
+     * @return string The route path
+     *
+     * @internal Used for testing purposes only.
+     */
+    public function detectRoutePath(string $protocol = ''): string
+    {
+        if ($protocol === '') {
+            $protocol = $this->appConfig->uriProtocol;
+        }
+
+        switch ($protocol) {
+            case 'REQUEST_URI':
+                $routePath = $this->parseRequestURI();
+                break;
+
+            case 'QUERY_STRING':
+                $routePath = $this->parseQueryString();
+                break;
+
+            case 'PATH_INFO':
+            default:
+                $routePath = $this->server[$protocol] ?? $this->parseRequestURI();
+                break;
+        }
+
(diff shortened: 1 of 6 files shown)


system/HTTP/SiteURIFactory.php
```diff
@@ -0,0 +1,224 @@
...
+    private App $appConfig;
+
+    /**
+     * @param array $server Superglobal $_SERVER array
+     * @param array $get    Superglobal $_GET array
+     */
+    public function __construct(array &$server, array &$get, App $appConfig)
+    {
+        $this->server    = &$server;
+        $this->get       = &$get;
```
kenjis, line 37: I'm using references. but since we have not used pass-by-reference for superglobals before, 
this may be breaking changes. I'll revert back to passing by values.

kenjis: Done.

system/HTTP/SiteURIFactory.php
```diff
@@ -0,0 +1,261 @@
...
+
+    private function updateServer(string $key, string $value): void
+    {
+        $_SERVER[$key] = $value;
+    }
+
+    private function updateGetArray(array $array): void
+    {
+        $_GET = $array;
+    }
```
MGatner, line 185: These make me pretty uneasy. What about an external service dependency that handles Globals, so it can be mocked or replaced? This behavior within a Factory is very counterintuitive.

kenjis: What do you mean?
Adding another class that handles globals?
```php
    private $superglobals;

    private function updateServer(string $key, string $value): void
    {
        $this->superglobals->setServer($key, $value);
    }
```

MGatner: Yes, precisely. It could be adapted to anywhere else in the framework we touch globals.

kenjis: Added class `Superglobals`.

system/HTTP/SiteURIFactory.php
```diff
@@ -0,0 +1,245 @@
...
+        // string contains a colon followed by a number. So we attach a dummy
+        // host since REQUEST_URI does not include the host. This allows us to
+        // parse out the query string and path.
+        $parts = parse_url('http://dummy' . $this->superglobals->server('REQUEST_URI'));
+        $query = $parts['query'] ?? '';
+        $path  = $parts['path'] ?? '';
+
+        // Strip the SCRIPT_NAME path from the URI
+        if (
+            $path !== '' && isset($this->superglobals->server('SCRIPT_NAME')[0])
```
kenjis, line 134: `isset($this->superglobals->server('SCRIPT_NAME')[0])`
What is this condition checked?
`$_SERVER['SCRIPT_NAME']` is a string. 

The condition comes from:
https://github.com/codeigniter4/CodeIgniter4/blob/6cedbe97a8863ccc6a3b4402fabe402ed24677e6/system/HTTP/IncomingRequest.php#L282-L286

MGatner: Does that end up being translated as this, because of string array access?
```php
$_SERVER['SCRIPT_NAME'] !== ''
```

MGatner: Yes. I assume this is unintended but that's the equivalence. 

https://3v4l.org/XC3NO
https://3v4l.org/hqebU

system/HTTP/SiteURIFactory.php
```diff
@@ -0,0 +1,245 @@
...
+    /**
+     * Will parse QUERY_STRING and automatically detect the URI from it.
+     *
+     * This method updates superglobal $_SERVER and $_GET.
+     *
+     * @return string The route path (before normalization).
+     */
+    private function parseQueryString(): string
+    {
+        $query = $this->superglobals->server('QUERY_STRING') ?? @getenv('QUERY_STRING');
```
MGatner, line 181: What are the potential errors suppressed here? 

> Returns the value of the environment variable name, or false if the environment variable name does not exist.

I think we need to cast to string if we are going to trim this below.

system/Superglobals.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @return array|string|null
+     */
+    public function get(string $key)
+    {
+        return $this->get[$key] ?? null;
+    }
+
+    public function setGetArray(array $array): void
```
MGatner, line 49: Did you only add setters for a single SERVER key and the entire GET array because that is all the framework used? I could see wanting at least `setGet()` as well.

system/HTTP/SiteURIFactory.php
```diff
@@ -0,0 +1,245 @@
...
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\HTTP;
+
+use CodeIgniter\HTTP\Exceptions\HTTPException;
+use CodeIgniter\Superglobals;
+use Config\App;
+
+class SiteURIFactory
```
MGatner, line 18: Please explain your scope choices - not making the class `final` but making most of its methods `private` instead of `protected` or `final protected`.

MGatner (changes requested):

system/HTTP/SiteURIFactory.php
kenjis: I don't know. It came from here:
https://github.com/codeigniter4/CodeIgniter4/commit/a8c5ac5df07#diff-2f873856503e4e4cdeca5087aa04d9944d5d732a92a1f1d756b5288a2b4b582dR402

system/HTTP/SiteURIFactory.php
kenjis: I made it final.

system/HTTP/SiteURIFactory.php
kenjis: Refactored. fde3216efdb84392aeb7c53df1f2df2c0a6bd74f

system/HTTP/SiteURIFactory.php
kenjis: Refactored. 78b0ac6ef9d1fed2a48d16d74b1389a550146d7e

system/Superglobals.php
kenjis: Added.

system/HTTP/SiteURIFactory.php
MGatner: Thanks! I think that is better.

MGatner (approved):

TimexPeachtree (approved):

Issue #7259, opened 2023-02-16
Bug: The selected data type from Entity or QueryBuilder is defferent from DB field type.


kai890707:
### PHP Version

8.1

### CodeIgniter4 Version

4.1.5

### CodeIgniter4 Installation Method

Git

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

cli

### Database

MariaDB 10.1

### What happened?

The data type selected is different from the DB field type.
For example, if you use a migrate file to define a field type as INTEGER or FLOAT, and then execute `php spark migrate` on the database to create a table, the data type returned using Query Builder will be different from the type initially set.

### Steps to Reproduce

1. Use Migration make a migrate file.
2. addField into mugrate file.
```php
          $this->forge->addField([
            'is_string'    => [
                'type'           => 'VARCHAR',
                'constraint'     => 30,
                'comment'        => "is_string"
            ],
            'is_integer'   => [
                'type'           => 'INT',
                'constraint'     => 30,
                'comment'        => "is_integer"
            ],
            'is_float'   => [
                'type'           => 'FLOAT',
                'comment'        => "is_float"
            ],
        ]);
        $this->forge->addKey('is_string', true);
        $this->forge->createTable('CastTable');
```
3. Use db query in Controller to get data
```php
$db = \Config\Database::connect();
$queryData = $db->table('CastTable')->get()->getResult();
var_dump($queryData);
```
4.Get data of the wrong type
```
array(1) {
  [0]=>
  object(stdClass)#57 (3) {
    ["is_string"]=>
    string(17) "Hi CodeIgniter4 !"
    ["is_integer"]=>
    string(3) "999"
    ["is_float"]=>
    string(5) "9.999"
  }
}
```

### Expected Output

```
array(1) {
  [0]=>
  object(stdClass)#57 (3) {
    ["is_string"]=>
    string(17) "Hi CodeIgniter4 !"
    ["is_integer"]=>
    int(999)
    ["is_float"]=>
    float(9.999)
  }
}
```

kai890707:
I finished fixing the errors in this section and I can send a PR.

kenjis:
This is not a bug in CodeIgniter.
It is the specification of PHP database module like `MySQLi`.

kai890707:
So does this need to be adjusted by the user? Or can CodeIgniter 4 add this part of the fix?

tangix:
[...]

PR #7260, opened 2023-02-16
Enhance: [MySQLi] use MYSQLI_OPT_INT_AND_FLOAT_NATIVE

kai890707:
**Description**
See #7259 

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/MySQLi/Connection.php
+++ b/system/Database/MySQLi/Connection.php
@@ -99,6 +99,8 @@ class Connection extends BaseConnection
 
         $this->mysqli->options(MYSQLI_OPT_CONNECT_TIMEOUT, 10);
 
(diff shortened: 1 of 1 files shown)


kenjis:
Thank you for sending PR!

This is a breaking change, so we don't accept this in `develop`.

I think this should be configurable in MySQLi Database Config.,
and this is an enhancement so this PR should go to `4.4` branch.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#if-you-sent-to-the-wrong-branch

@codeigniter4/database-team 
Opinions?

kenjis:
@kai890707 You forgot your GPG signing.

PR #7265, opened 2023-02-16
Enhance: [MySQLi] use MYSQLI_OPT_INT_AND_FLOAT_NATIVE.

kai890707:
Supersedes #7260

**Description**
See #7259 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


michalsn:
As @kenjis mentioned in #7260 - this should be configurable. Otherwise, we will break other people's code.

kai890707:
Are the current changes in this area acceptable?
Currently only enhancements are made for the `MYSQLI_OPT_INT_AND_FLOAT_NATIVE` section.

app/Config/Database.php
```diff
@@ -25,23 +25,24 @@ class Database extends Config
...
+        'DBDebug'      => true,
+        'charset'      => 'utf8',
+        'DBCollat'     => 'utf8_general_ci',
+        'swapPre'      => '',
+        'encrypt'      => false,
+        'compress'     => false,
+        'strictOn'     => false,
+        'failover'     => [],
+        'port'         => 3306,
+        'DBTypeNative' => false,
```
michalsn, line 45: I don't like the name you have chosen. This would be my suggestion, but I would like to hear from others about it... maybe `nativeType`... @kenjis?

```suggestion
        'strictType' => false,
```

system/Database/BaseConnection.php
```diff
@@ -339,6 +340,13 @@ abstract class BaseConnection implements ConnectionInterface
      */
     protected $queryClass = Query::class;
 
+    /**
+     * Use MYSQLI_OPT_INT_AND_FLOAT_NATIVE
+     *
+     * @var bool
+     */
+    protected $DBTypeNative = false;
```
michalsn, line 348: This should be placed in the MySQLi driver, since it is related to this driver exclusively.

```suggestion
    protected bool $strictType = false;
```

system/Database/MySQLi/Connection.php
```diff
@@ -98,6 +98,10 @@ public function connect(bool $persistent = false)
         mysqli_report(MYSQLI_REPORT_ALL & ~MYSQLI_REPORT_INDEX);
 
         $this->mysqli->options(MYSQLI_OPT_CONNECT_TIMEOUT, 10);
+        
+        if(isset($this->DBTypeNative) && $this->DBTypeNative){
+            $this->mysqli->options(MYSQLI_OPT_INT_AND_FLOAT_NATIVE,1);
```
michalsn, line 103: Please run `composer cs-fix` to fix the coding style issues.

michalsn (changes requested):
In general, it looks good, thanks.

We would need some tests for this feature to ensure it works. You can see how you can change the database config when you write your tests here: https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/Live/ConnectTest.php#L33 We should have some sample data in the database already that you can use for the sample query - you don't need to create a new table etc.

Finally, we need the user guide to be updated - other users have to know there is this new feature that they can use. Please add the description for this config option here: https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/database/configuration.rst (a table with all option is at the end).

app/Config/Database.php
kenjis: I don't like `DBTypeNative`. `strictType` seems a bit different.

`intAndFloatNative`. `numberNative`. Too specific?
`nativeType` may be good if we prefer generic name.

system/Database/BaseConnection.php
kenjis: Yes, please move the property to MySQLi driver.

app/Config/Database.php
michalsn: +1 for `numberNative` - this is only for MySQL so we should be good.

kai890707: Thank you both for your replies. 
I'll change the variable name to numberNative.

system/Database/BaseConnection.php
kai890707: Okay!

system/Database/MySQLi/Connection.php
kai890707: Wow ~ I overlooked this, thanks for reminding.

kenjis:
About the user guide. Please add this in the v4.4.0 changelog, too.
https://github.com/codeigniter4/CodeIgniter4/blob/4.4/user_guide_src/source/changelogs/v4.4.0.rst#database

app/Config/Database.php
kai890707: Sorry to interrupt your time, @michalsn .
I would like to ask about the test, if I move `numberNative `to Connect.php in MySQLi, can I do the data type test directly in RawSqlTest.php? Or do I need to do the test for this property in ConnectTest.php separately?

michalsn: Personally, I think the best place for these tests will be in `Database/Live/MySQLi/NumberNativeTest.php` (new file).

You can look at https://github.com/codeigniter4/CodeIgniter4/blob/develop/tests/system/Database/Live/TransactionDBDebugTrueTest.php - as a reference of how you can build these tests - it might be a lot better example than I suggested before.

Make just one test for selecting number values when `numberNative` is enabled and one more when this option is disabled. That's all we need.

michalsn: Crap... sorry, that won't be a better example. We need to set the correct config before we connect to the database.

kai890707: I have added a test file, the content of some simple tests, if there is a need to add I will fill in.

tests/system/Database/Live/MySQLi/NumberNativeTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $config = config('Database');
+
+        $this->tests = $config->tests;
+
+        $this->tests['DBDriver'] = 'MySQLi';
```
michalsn, line 40: This can be removed. DBDriver is set automatically by the environment.

tests/system/Database/Live/MySQLi/NumberNativeTest.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public function testEnableNumberNative()
+    {
+        $this->tests['numberNative'] = true;
+
+        $db1 = Database::connect($this->tests);
+
```
michalsn, line 48: We have to add this check to every test.

```suggestion
if ($db1->DBDriver !== 'MySQLi') {
    $this->markTestSkipped('Only MySQLi can complete this test.');
}
```

michalsn (commented):
Adding `markTestSkipped()` should cover all failed database tests.

Please check out the other failed tests - the descriptions are quite self-explanatory:
* https://github.com/codeigniter4/CodeIgniter4/actions/runs/4211879196/jobs/7311399553#step:8:28
* https://github.com/codeigniter4/CodeIgniter4/actions/runs/4211879194/jobs/7311399426#step:11:18

user_guide_src/source/database/configuration.rst
```diff
@@ -108,49 +108,50 @@ default group's configuration settings. The values should be name following this
...
+                 while developing an application (``MySQLi`` only).
+**port**         The database port number.
+**foreignKeys**  true/false (boolean) - Whether or not to enable Foreign Key constraint (``SQLite3`` only).
+
+                 .. important:: SQLite3 Foreign Key constraint is disabled by default.
+                     See `SQLite documentation <https://www.sqlite.org/pragma.html#pragma_foreign_keys>`_.
+                     To enforce Foreign Key constraint, set this config item to true.
+**busyTimeout**  milliseconds (int) - Sleeps for a specified amount of time when a table is locked (``SQLite3`` only).
+**numberNative** true/false (boolean) - Whether or not to enable MYSQLI_OPT_INT_AND_FLOAT_NATIVE (``MySQLi`` only).
+===============  ===========================================================================================================
```
kenjis, line 154: This table is broken. `**numberNative**` (the longest item) and `===...` must be the same length.
The `===...` line in the header must also be exactly the same length.

```suggestion
**numberNative** true/false (boolean) - Whether or not to enable MYSQLI_OPT_INT_AND_FLOAT_NATIVE (``MySQLi`` only).
================ ===========================================================================================================
```

kai890707: Thankfully, I have made changes to this issue.

tests/system/Database/Live/MySQLi/NumberNativeTest.php
kai890707: Thanks! I understand and have made the changes in the next commit.

tests/system/Database/Live/MySQLi/NumberNativeTest.php
kai890707: Thanks! I understand and have made the changes in the next commit.

kai890707:
> * https://github.com/codeigniter4/CodeIgniter4/actions/runs/4211879196/jobs/7311399553#step:8:28

Sorry, @michalsn , I don't quite understand how to fix this error.
Do you have any suggestions to help me clear this error?
https://github.com/codeigniter4/CodeIgniter4/actions/runs/4211879196/jobs/7311399553#step:8:28

kenjis:
@kai890707 Please do not use `4.4` as your PR branch.
It breaks my local repository's `4.4` branch when I checkout your PR branch.

kenjis:
@kai890707 Just run `composer cs-fix` in your project root folder.

kai890707:
> @kai890707 Please do not use `4.4` as your PR branch in the next PR. It breaks my local repository's `4.4` branch when I checkout your PR branch.

Oh, I'm so sorry.
This is my first contribution to an open source project and I will improve it if there is anything missing.



> @kai890707 Just run `composer cs-fix` in your project root folder.

Okay, thanks for the answer.

user_guide_src/source/changelogs/v4.4.0.rst
```diff
@@ -59,6 +59,9 @@ Forge
 Others
 ------
 
+- **numberNative:** Added the ``numberNative`` attribute to the Database setting to keep the variable type obtained after SQL Query consistent with the type set in the DataBase.
```
kenjis, line 62:
```suggestion
- **Database:** Added the ``numberNative`` attribute to the Database Config for MySQLi to keep the variable type obtained after SQL Query consistent with the type set in the datadase.
```

kai890707: Thank you, I have revised this description.

user_guide_src/source/changelogs/v4.4.0.rst
```diff
@@ -59,6 +59,10 @@ Forge
 Others
 ------
 
+- **Database:** Added the ``numberNative`` attribute to the Database Config for MySQLi to keep the variable type obtained after SQL Query consistent with the type set in the database.
+  See `configuration.rst <https://github.com/codeigniter4/CodeIgniter4/blob/4.4/user_guide_src/source/database/configuration.rst>`_
```
kenjis, line 63: This is a link to the GitHub repository.
We need to link to the User Guide itself, and it is not only by HTML but also EPUB.
So we need to use links of Sphinx.

```suggestion
  See :ref:`Database Configuration <database-configuration-explanation-of-values>`.
```

and add  the following:
```diff
--- a/user_guide_src/source/database/configuration.rst
+++ b/user_guide_src/source/database/configuration.rst
@@ -104,6 +104,8 @@ default group's configuration settings. The values should be name following this
     database.default.password = '';
     database.default.database = 'ci4';
 
+.. _database-configuration-explanation-of-values:
+
 **********************
 Explanation of Values:
 **********************
```

kai890707: Before push commit, I would like to ask, is this the desired modification?
image

kenjis: No.
```

.. _database-configuration-explanation-of-values:

**********************
Explanation of Values:
**********************
```

kai890707: Does it look like this?
image

kenjis: Yes!

kai890707: Thank you, I have corrected these descriptions.

app/Config/Database.php
```diff
@@ -25,49 +25,51 @@ class Database extends Config
...
+        'DBCollat'     => 'utf8_general_ci',
+        'swapPre'      => '',
+        'encrypt'      => false,
+        'compress'     => false,
+        'strictOn'     => false,
+        'failover'     => [],
+        'port'         => 3306,
+        'foreignKeys'  => true,
+        'busyTimeout'  => 1000,
+        'numberNative' => false,
```
kenjis, line 72: I don't think `numberNative` is needed here. Because this is a sample config for SQLite3.

kai890707: Thank you, I will change this part, anything else needs to be changed?

user_guide_src/source/changelogs/v4.4.0.rst
```diff
@@ -59,6 +59,9 @@ Forge
 Others
 ------
 
+- **Database:** Added the ``numberNative`` attribute to the Database Config for MySQLi to keep the variable type obtained after SQL Query consistent with the type set in the database.
```
kenjis, line 62: Sorry, here is "Enhancements > Database > Others".

```suggestion
- **MySQLi:** Added the ``numberNative`` attribute to the Database Config to keep the variable type obtained after SQL Query consistent with the type set in the database.
```

app/Config/Database.php
kenjis: Added one inline comment.
That's all.
Thank you.

user_guide_src/source/changelogs/v4.4.0.rst
kai890707: Thanks, I've made a correction in the next commit.

kai890707:
@kenjis  @michalsn 
All changes have been modified, please review and I will fill in any omissions.
Thank you both for your help!  😄

michalsn (approved):
Good job. Thanks!

kenjis (approved):
LGTM!

kenjis:
@kai890707 Thank you!

PR #7272, opened 2023-02-18
docs: Typographic correction in forge.rst

grimpirate:
The documentation for forge states that the parameters for cascading foreign keys produce an "on update" and "on update" SQL statement, wherein the produce an "on delete" and "on update" SQL statement when specified.

### DIFF
--- a/user_guide_src/source/dbmgmt/forge.rst
+++ b/user_guide_src/source/dbmgmt/forge.rst
@@ -198,7 +198,7 @@ you may add them directly in forge:
 
 .. literalinclude:: forge/012.php
 
-You can specify the desired action for the "on delete" and "on update" properties of the constraint as well as the name:
+You can specify the desired action for the "on update" and "on delete" properties of the constraint as well as the name:
 
 .. literalinclude:: forge/013.php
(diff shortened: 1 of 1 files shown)


user_guide_src/source/dbmgmt/forge.rst
```diff
@@ -198,7 +198,7 @@ you may add them directly in forge:
 
 .. literalinclude:: forge/012.php
 
-You can specify the desired action for the "on update" and "on update" properties of the constraint as well as the name:
+You can specify the desired action for the "on delete" and "on update" properties of the constraint as well as the name:
```
kenjis, line 201: The parameter order is $onUpdate and $onDelete.
https://github.com/codeigniter4/CodeIgniter4/blob/ba75eed0751dd48cb84be6116438632f4b3c9a93/system/Database/Forge.php#L407

```suggestion
You can specify the desired action for the "on update" and "on delete" properties of the constraint as well as the name:
```

It seems better to fix the order in the comments in the sample code.

MGatner (approved):
Good catch!

PR #7277, opened 2023-02-19
fix: [QueryBuilder] RawSql causes error when using like() and countAllResults()

### LINKED ISSUE #7268: Bug: RawSql() triggers an error when used in combination with like() and countAllResults()
### PHP Version

8.1

### CodeIgniter4 Version

4.3.1

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

```php
$builder = $this->db->table('users');
$builder->select('id');
$builder->like(new RawSql("CONCAT(username, email)"), 'sky');
$count = $builder->countAllResults(FALSE);
$tableList = $builder->get()->getResultArray();
```

In Query Builder Class, if you use new RawSql() in $builder ->like(), and then use $builder ->countAllResults (FALSE); An error will be triggered when the total number of query results is obtained and then the query results are obtained.

Error prompt:
```
code: 500
file:"G:\\Code\\website\\ci4_cms\\vendor\\codeigniter4\\framework\\system\\Database\\BaseBuilder.php"
line: 3071
message: "Cannot use object of type CodeIgniter\\Database\\RawSql as array"
```

After debugging, it is found that if the countAllResults parameter is not set to FALSE, no error will be reported, but the subsequent get() cannot obtain the correct query results;

In addition, the third parameter of get() can also reset the query object. After setting, $builder ->countAllResults(); The same error is reported!

Problem conclusion: New RawSql() cannot be used in the link() before resetting values for SELECT; Or if you want to use new RawSql(), you cannot reset values for SELECTs??

### Steps to Reproduce

```php
$builder = $this->db->table('users');
$builder->select('id');
$builder->like(new RawSql("CONCAT(username, email)"), 'sky');
$count = $builder->countAllResults(FALSE);
$tableList = $builder->get(25, 0)->getResultArray();
```
or
```php
$builder = $this->db->table('users');
$builder->select('id');
$builder->like(new RawSql("CONCAT(username, email)"), 'sky');
$tableList = $builder->get(25, 0, FALSE)->getResultArray();
$count = $builder->countAllResults();
```

### Expected Output

Get the total number of query results and query result array normally

### Anything else?

_No response_

kenjis:
**Description**
Fixes #7268

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/BaseBuilder.php
+++ b/system/Database/BaseBuilder.php
@@ -3082,6 +3082,10 @@ class BaseBuilder
                     continue;
                 }
 
+                if ($qbkey instanceof RawSql) {
+                    continue;
+                }
+
                 if ($qbkey['condition'] instanceof RawSql) {
                     $qbkey = $qbkey['condition'];
 
--- a/tests/system/Database/Live/LikeTest.php
+++ b/tests/system/Database/Live/LikeTest.php
@@ -11,6 +11,7 @@
 
 namespace CodeIgniter\Database\Live;
 
+use CodeIgniter\Database\RawSql;
 use CodeIgniter\Test\CIUnitTestCase;
 use CodeIgniter\Test\DatabaseTestTrait;
 use Tests\Support\Database\Seeds\CITestSeeder;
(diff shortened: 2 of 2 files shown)


samsonasik (approved):

kenjis:
@codeigniter4/database-team Please review.

michalsn (approved):

PR #7278, opened 2023-02-19
fix: missing cipher field for ci3 encryption compatibility

objecttothis:
**Description**
@kenjis did the work of getting backward compatibility with CI3 encryption. However since CI3 defaulted to the aes-128-cbc encryption cipher, we need a way to tell CI4 to use that algorithm when decoding CI3 encrypted data. This change adds the field to \App\Config\Encryption.php

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/app/Config/Encryption.php
+++ b/app/Config/Encryption.php
@@ -63,7 +63,7 @@ class Encryption extends BaseConfig
      *
      * Set to aes-128-cbc for CI3 Encryption compatibility.
      */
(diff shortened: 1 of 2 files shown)


objecttothis:
Looks like Deptrac failed due to rate limiting, not an actual problem in the code.

kenjis:
Thank you for trying to improve CI4!

But I have already sent #7273.

PR #7282, opened 2023-02-20
rework: URI creation and URL helper

### LINKED ISSUE #7123: Bug: $this->request->getUri()->getPath() is not compatible with PSR-7

kenjis:
Needs ~~#7252~~, ~~#7256~~, ~~#7297~~

**Description**
Fixes #7123, #7296, #7643

- Fixes
  - `$this->request->getUri()` returns a **normalized** current URI (`SiteURI`)
    - if `$indexPage` is `index.php`, the URI contains `index.php`
  - [BC] `$this->request->getUri()->getPath()` returns full URI path
  - `current_url(true)` returns the same instance with `$this->request->getUri()`
  - `base_url()` and `site_url()` now support protocol-relative links
  - `ControllerTestTrait::withUri()` updates the Request instance, too. Because the Request instance has the URI instance. See https://forum.codeigniter.com/showthread.php?tid=88213
- use `SiteURIFactory`, `SiteURI`
- add `Services::siteurifactory()`
- add `Services::superglobals()`
- URL helper
  - add `SiteURI::baseUrl()` and `SiteURI::siteUrl()`
  - remove `_get_uri()`
  - `site_url()`, `base_url()`, `current_url()` use  `SiteURI` (`Services::request()->getUri()`)

Before:
before
After:
after

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


kenjis:
All checks passed! 

However, to be honest I am a little concerned about whether the existing Controller/HTTP tests will be broken.

kenjis:
Rebased.

kenjis:
Rebased and updated and added docs.

kenjis:
Rebased.

kenjis:
I would like to include this in 4.4.0 if possible, can anyone review this?

system/HTTP/SiteURI.php
```diff
@@ -145,7 +145,7 @@ private function determineBaseURL(
         $uri = new URI($baseURL);
 
         // Update scheme
-        if ($scheme !== null) {
+        if ($scheme !== null && $scheme !== '') {
```
MGatner, line 148: How about:
```suggestion
        if ($scheme) {
```

system/HTTP/SiteURI.php
```diff
@@ -363,4 +363,67 @@ protected function applyParts(array $parts): void
             $this->password = $parts['pass'];
         }
     }
+
+    /**
+     * For baser_url() helper.
```
MGatner, line 368:
```suggestion
     * For base_url() helper.
```

system/HTTP/SiteURI.php
```diff
@@ -363,4 +363,67 @@ protected function applyParts(array $parts): void
...
+
+        $config            = clone config(App::class);
+        $config->indexPage = '';
+
+        $host = $this->getHost();
+
+        $uri = new self($config, $relativePath, $host, $scheme);
+
+        // Support protocol-relative links
+        if ($scheme === '') {
```
MGatner, line 385: Should this handle `null` as well, like above?

system/HTTP/SiteURIFactory.php
```diff
@@ -42,6 +47,7 @@ public function createFromGlobals(): SiteURI
      * Create the SiteURI object from URI string.
      *
      * @internal Used for testing purposes only.
+     * @testTag
```
MGatner, line 50: 🤗

system/Test/FeatureTestTrait.php
```diff
@@ -265,15 +268,24 @@ public function options(string $path, ?array $params = null)
...
 
         // $path may have a query in it
-        $parts                   = explode('?', $path);
-        $_SERVER['QUERY_STRING'] = $parts[1] ?? '';
+        $path  = URI::removeDotSegments($path);
+        $parts = explode('?', $path);
+        $path  = $parts[0];
+        $query = $parts[1] ?? '';
+
+        $_SERVER['QUERY_STRING'] = $query;
```
MGatner, line 280: Any reason to handle this via `Superglobals`?

MGatner (changes requested):
I didn't look too closely at the test updates but the overall PR looks very good. I'm so glad to be moving this out of those helpers. A few comments.

system/HTTP/SiteURI.php
samsonasik: @MGatner truthy check should only for bool value, using exact compare null or empty string is better to know the exact usage

MGatner:
Checked in.

system/HTTP/SiteURI.php
kenjis: No. `null` is the default value for `$scheme`, and `base_url()` must return the baseURL (e.g., `http://example.com/`).
But `''` is for protocol-relative links (`//example.com/img/test.jpg`). It must omit the protocol part.

system/Test/FeatureTestTrait.php
kenjis: No. Refactored.

MGatner (approved):
This is very good! I'm a little nervous how many test changes there are. It looks like the only real assertion changes are leading `/` which hopefully are just internal and won't affect projects. 

Very nice work; I've wanted this framework-wide site URI handling for a long time.

MGatner:
I see a few outstanding lint issues.

kenjis:
Needs #7826

kenjis:
> I'm a little nervous how many test changes there are. It looks like the only real assertion changes are leading / which hopefully are just internal and won't affect projects.

I expect few tests to fail unless you test the URI by setting the state of dependent objects and/or global variables, as CI4 does.

I don't think this is such an unrealistic expectation, since in a normal web app, site URIs are rarely tested.

samsonasik (approved):

kenjis:
Thank you for reviews!

MGatner:
🥳

PR #7283, opened 2023-02-20
Setting Rector Cache in Github Action

samsonasik:
**Description**

To make faster build.

**Checklist:**
- [x] Securely signed commits


samsonasik:
It seems not really has different, closing for now.

PR #7285, opened 2023-02-20
docs: [Model] Add missing @method

maniaba:
**Description**

I added the methods in the PHPDoc that are missing in the model.
New methods for feat `upsert` and `deleteBatch`, `distinct`, `emptyTable`, `fromSubquery`, `ignore`, `union`, `unionAll`, `updateBatch`

- deleteBatch
- distinct
- emptyTable
- fromSubquery
- ignore
- union
- unionAll
- updateBatch
- upsert

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


kenjis:
Does the methods really work with the Model?
Should work with the Model?

kenjis:
The model does not support the `upsert()`. See #7286

system/Model.php
```diff
@@ -40,47 +41,60 @@
...
+ * @method $this                              selectMax(string $select = '', string $alias = '')
+ * @method $this                              selectMin(string $select = '', string $alias = '')
+ * @method $this                              selectSum(string $select = '', string $alias = '')
+ * @method $this                              setQueryAsData(BaseBuilder|RawSql $query, ?string $alias = null, array|string|null $columns = null)
+ * @method bool                               truncate()
+ * @method $this                              union(BaseBuilder|Closure $union)
+ * @method $this                              unionAll(BaseBuilder|Closure $union)
+ * @method BaseResult|false|int|object|string updateBatch(array|object|null $set = null, array|RawSql|string|null $constraints = null, int $batchSize = 100)
+ * @method $this                              updateFields(string|string[] $set, bool $addToDefault = false, ?array $ignore = null)
+ * @method BaseResult|false|int|object|string upsert(array|object|null $set = null, ?bool $escape = null)
```
kenjis, line 94: The model does not support the `upsert()`. See #7286

system/Model.php
```diff
@@ -40,47 +41,60 @@
...
- * @method $this select($select = '*', ?bool $escape = null)
- * @method $this selectAvg(string $select = '', string $alias = '')
- * @method $this selectCount(string $select = '', string $alias = '')
- * @method $this selectMax(string $select = '', string $alias = '')
- * @method $this selectMin(string $select = '', string $alias = '')
- * @method $this selectSum(string $select = '', string $alias = '')
- * @method $this where($key, $value = null, ?bool $escape = null)
- * @method $this whereIn(?string $key = null, $values = null, ?bool $escape = null)
- * @method $this whereNotIn(?string $key = null, $values = null, ?bool $escape = null)
+ * @method false|int|string[]                 deleteBatch(array|object|null $set = null, array|RawSql|null $constraints = null, int $batchSize = 100)
```
kenjis, line 44: This returns `false|int|string[]`. So the model does not work correctly.

system/Model.php
```diff
@@ -40,47 +41,60 @@
...
+ * @method $this                              distinct(bool $val = true)
+ * @method bool                               emptyTable()
```
kenjis, line 46: This returns `bool`. So the model does not work correctly.
I don't know this method should remove soft deleted records or not.

system/Model.php
```diff
@@ -40,47 +41,60 @@
...
+ * @method $this                              selectAvg(string $select = '', string $alias = '')
+ * @method $this                              selectCount(string $select = '', string $alias = '')
+ * @method $this                              selectMax(string $select = '', string $alias = '')
+ * @method $this                              selectMin(string $select = '', string $alias = '')
+ * @method $this                              selectSum(string $select = '', string $alias = '')
+ * @method $this                              setQueryAsData(BaseBuilder|RawSql $query, ?string $alias = null, array|string|null $columns = null)
+ * @method bool                               truncate()
+ * @method $this                              union(BaseBuilder|Closure $union)
+ * @method $this                              unionAll(BaseBuilder|Closure $union)
+ * @method BaseResult|false|int|object|string updateBatch(array|object|null $set = null, array|RawSql|string|null $constraints = null, int $batchSize = 100)
```
kenjis, line 92: This returns `BaseResult|false|int|object|string`. So the model does not work correctly.

PR #7287, opened 2023-02-21
docs: add note for case-sensitive file systems

kenjis:
**Description**
- add note for case-sensitive file systems

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/concepts/autoloader.rst
+++ b/user_guide_src/source/concepts/autoloader.rst
@@ -30,6 +30,12 @@ they work in sequence and don't get in each other's way.
 The autoloader is always active, being registered with ``spl_autoload_register()`` at the
 beginning of the framework's execution.
 
+.. important:: You should always be careful about the case of filenames. Many
+    developers develop on case-insensitive file systems on Windows or macOS.
+    However, most server environments use case-sensitive file systems. If the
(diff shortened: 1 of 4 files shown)


ddevsr:
How about note this for installation & troubleshooting?

kenjis:
Thanks! Adding in troubleshooting seems good.

installation? in where?

ddevsr:
In here https://codeigniter4.github.io/CodeIgniter4/installation/running.html?

cijagani:
This is actually depending on OS. ex. windows file system is case insensitive, and Linux file system is fully case sensitive.

MGatner (approved):

kenjis:
Added in installation/troubleshooting.rst and installation/running.rst.

kenjis:
> This is actually depending on OS. ex. windows file system is case insensitive, and Linux file system is fully case sensitive.

It depends on Filesystem.
On macOS, HFS+ and APFS can be configured as case sensitive when you format a filesystem.
On Windows, there are Case sensitivity options for mounting a drive in WSL.
See https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity

samsonasik (approved):

samsonasik:
Thank you @kenjis

PR #7288, opened 2023-02-21
test: fix out-of-dated test code in CodeIgniterTest

kenjis:
**Description**
If I run the test method only, it causes Error.

- fix out-of-dated test code

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


MGatner (approved):

samsonasik:
Thank you @kenjis

PR #7305, opened 2023-02-25
feat: add timestamp to `base_url()`

heyfranks:
**Description**
when you create a website, the local cache is a problem, if you add a timestamp to a url in develop, this get more confortable.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/Helpers/url_helper.php
+++ b/system/Helpers/url_helper.php
@@ -147,8 +147,19 @@ if (! function_exists('base_url')) {
             : $request->getUri()->getBaseURL();
 
         $config->indexPage = '';
+        
+        
+        $addTimestamp = '';
 
-        return site_url($relativePath, $scheme, $config);
+        if(env('CI_ENVIRONMENT')!='production') {
+            $array = explode('.', $relativePath);
+            $extension = array_pop($array);
+            if(in_array($extension, ['png', 'jpg', 'css', 'js', 'ico'])) {
+                $addTimestamp = '?t='.time();
+            }
+        }
+
+        return site_url($relativePath, $scheme, $config).$addTimestamp;
     }
 }
 


kenjis:
Thank you for sending the PR!

We don't think this is a bug. This is an enhancement, so you need to send to `4.4` branch,
instead of `develop`.

I'm not sure this feature should be merged.
1. The `base_url()` should have this feature?
2. What if on production environment?
3. There is no way to disable this feature.

kenjis:
```
There was 1 error:

1) CodeIgniter\Helpers\URLHelper\SiteUrlTest::testUrls with data set #19 ('http://example.com/', 'index.php', null, false, array('news', 'local', '123'), 'http://example.com/index.php/...al/123', 'http://example.com/news/local/123')
ErrorException: explode() expects parameter 2 to be string, array given

/home/runner/work/CodeIgniter4/CodeIgniter4/system/Helpers/url_helper.php:155
/home/runner/work/CodeIgniter4/CodeIgniter4/tests/system/Helpers/URLHelper/SiteUrlTest.php:81
```
See https://github.com/codeigniter4/CodeIgniter4/actions/runs/4267638325/jobs/7430504033

michalsn:
I understand what you want to do, but IMO this problem should be addressed in the addon, not in the framework - at least not in the way it's proposed now.

Personally, I would suggest disabling caching in the developer tools for local development.

lonnieezell:
I agree. This does not belong at framework level for the `base_url` helper. This enforces a cache-busting scheme at one place, but provides not way to utilize it elsewhere. It also doesn't allow any other forms of cache-busting to be used. And to have it work on dev but not production is wrong - cache busting is required more in production than development, IMO. 

As @michalsn said, disable caching in your browser's developer tools while you're working on your code.

PR #7307, opened 2023-02-26
fix: incorrect metadata querying of Redis cache

### LINKED ISSUE #7304: Bug: Cache RedisHandler Always Returns null MetaData if a prefix is Set
### PHP Version

8.2

### CodeIgniter4 Version

4.3.2

Linux

### Which server did you use?

### Database

_No response_

### What happened?

When a prefix is provided for the Redis cache handler, the `getMetaData()` method always returns null.

### Steps to Reproduce

1. Set up Redis handler in the file `App\Config\Cache.php`.
2. Set the handler `redis`.
3. Set a key prefix.
4. Provide the appropriate Redis configuration.
5. Ensure that Redis is valid handler.

Now try to store any value in the Redis caching using the service.
Try to access the metadata using the method `getMetaData()`.
It will be received null.

### Expected Output

Proper Metadata array about the existing cache key.

### Anything else?

In the file `system\Cache\Handlers\RedisHandler.php` [Line #236](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Cache/Handlers/RedisHandler.php#L236),
the `key` is validated and the `prefix` is prepended.
In the next line the `get()` method is called which again prepend the `prefix`.

Hence, prepending the `prefix` twice turns the `key` into an always non-existing key.
The resolution is exchanging the order of execution for both lines.
```php
public function getMetaData(string $key)
    {
        $value = $this->get($key);
        $key   = static::validateKey($key, $this->prefix);

        if ($value !== null) {
            $time = Time::now()->getTimestamp();
            $ttl  = $this->redis->ttl($key);

            return [
                'expire' => $ttl > 0 ? $time + $ttl : null,
                'mtime'  => $time,
                'data'   => $value,
            ];
        }

        return null;
    }
```

paulbalandan:
**Description**
Fixes #7304 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Cache/Handlers/RedisHandlerTest.php
+++ b/tests/system/Cache/Handlers/RedisHandlerTest.php
@@ -198,6 +198,16 @@ final class RedisHandlerTest extends AbstractHandlerTest
         $this->assertIsArray($this->handler->getCacheInfo());
     }
 
+    public function testGetMetadataNotNull(): void
+    {
+        $this->handler->save(self::$key1, 'value');
+
+        $metadata = $this->handler->getMetaData(self::$key1);
+
+        $this->assertNotNull($metadata);
+        $this->assertIsArray($metadata);
+    }
+
     public function testIsSupported()
     {
         $this->assertTrue($this->handler->isSupported());
--- a/system/Cache/Handlers/RedisHandler.php
+++ b/system/Cache/Handlers/RedisHandler.php
@@ -233,12 +233,11 @@ class RedisHandler extends BaseHandler
      */
     public function getMetaData(string $key)
     {
-        $key   = static::validateKey($key, $this->prefix);
         $value = $this->get($key);
 
         if ($value !== null) {
             $time = Time::now()->getTimestamp();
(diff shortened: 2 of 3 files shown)


tests/system/Cache/Handlers/RedisHandlerTest.php
```diff
@@ -198,6 +198,15 @@ public function testGetCacheInfo()
         $this->assertIsArray($this->handler->getCacheInfo());
     }
 
+    public function testGetMetadataNotNull(): void
+    {
+        $this->handler->save(self::$key1, 'value');
+
+        $metadata = $this->handler->getMetaData(self::$key1);
```
kenjis, line 205: Please add an empty line.
Arrange-Act-Assert.
See https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

kenjis (approved):

kenjis:
@paulbalandan Thank you!

PR #7310, opened 2023-02-27
Remove Autoloader::sanitizeFilename() checking

### LINKED ISSUE #7055: Bug: a tilde (~) in the path, such as within iCloud local storage, will cause CodeIgniter to fail
### PHP Version

8.2

### CodeIgniter4 Version

4.2.11

macOS

### Which server did you use?

### Database

none

### What happened?

When a fresh composer-created CodeIgniter 4 application has a tilde(~) in its path, it crashes when staring with "php spark serve".

This is a critical failure, because on MacOS in all filesystem areas synced to iCloud, tildes are present, like so:
`/Users/moo/Library/Mobile Documents/com~apple~CloudDocs/Documents/development/swapper/vendor/codeigniter4`

I moved the fresh new app out of the com~apple~CloudDocs directory tree to /Users/shared/dev/swapper, and it immediately starts up fine with "php spark serve".

After renaming the directory from "/Users/shared/dev/swapper" to "/Users/shared/dev/~swapper", the application immediately fails when started with "php spark serve".

[CodeIgniter_4.2.11_tilde_bug.pdf](https://github.com/codeigniter4/CodeIgniter4/files/10357688/CodeIgniter_4.2.11_tilde_bug.pdf)

### Steps to Reproduce

1. cd to directory that does not include a tilde(~), such as /Users/shared/ on MacOS.
2. composer create-project codeigniter4/appstarter newApp
3. cd ./newApp
4. php spark serve
5. newApp works as expected
6. cd ../
7. mv ./newApp ./\~newApp
8. cd ./\~newApp
9. php spark serve
10. newApp fails due to the tilde-intolerance bug
11. cd ../
12. mv ./\~newApp ./newApp
13. cd newApp
14. php spark serve now works as expected

### Expected Output

php spark serve should start the application, no matter what characters are present in the path name.
And ESPECIALLY tildes should be accounted for, because all MacOS users should be using iCloud to sync their files.

### Anything else?

The test log on my machine:
```
{23-01-05 23:51}MoosacreM1ProMax:/Users/Shared/dev moo% cd swapper 
{23-01-05 23:51}MoosacreM1ProMax:/Users/Shared/dev/swapper moo% php spark serve

CodeIgniter v4.2.11 Command Line Tool - Server Time: 2023-01-05 23:51:41 UTC-06:00

CodeIgniter development server started on http://localhost:8080
Press Control-C to stop.
[Thu Jan  5 23:51:41 2023] PHP 8.2.1 Development Server (http://localhost:8080) started
^C
{23-01-05 23:51}MoosacreM1ProMax:/Users/Shared/dev/swapper moo% cd ..
{23-01-05 23:51}MoosacreM1ProMax:/Users/Shared/dev moo% mv swapper \~swapper
{23-01-05 23:52}MoosacreM1ProMax:/Users/Shared/dev moo% cd \~swapper 
{23-01-05 23:52}MoosacreM1ProMax:/Users/Shared/dev/~swapper moo% php spark serve
PHP Fatal error:  Uncaught Error: Class "Config\App" not found in /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/Factories.php:129
Stack trace:
#0 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Common.php(204): CodeIgniter\Config\Factories::__callStatic('config', Array)
#1 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/Services.php(146): config('App')
#2 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/BaseService.php(253): CodeIgniter\Config\Services::codeigniter(NULL, false)
#3 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/BaseService.php(194): CodeIgniter\Config\BaseService::__callStatic('codeigniter', Array)
#4 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/Services.php(143): CodeIgniter\Config\BaseService::getSharedInstance('codeigniter', NULL)
#5 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/BaseService.php(253): CodeIgniter\Config\Services::codeigniter()
#6 /Users/Shared/dev/~swapper/spark(82): CodeIgniter\Config\BaseService::__callStatic('codeigniter', Array)
#7 {main}
  thrown in /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/Factories.php on line 129

Fatal error: Uncaught Error: Class "Config\App" not found in /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/Factories.php:129
Stack trace:
#0 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Common.php(204): CodeIgniter\Config\Factories::__callStatic('config', Array)
#1 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/Services.php(146): config('App')
#2 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/BaseService.php(253): CodeIgniter\Config\Services::codeigniter(NULL, false)
#3 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/BaseService.php(194): CodeIgniter\Config\BaseService::__callStatic('codeigniter', Array)
#4 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/Services.php(143): CodeIgniter\Config\BaseService::getSharedInstance('codeigniter', NULL)
#5 /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/BaseService.php(253): CodeIgniter\Config\Services::codeigniter()
#6 /Users/Shared/dev/~swapper/spark(82): CodeIgniter\Config\BaseService::__callStatic('codeigniter', Array)
#7 {main}
  thrown in /Users/Shared/dev/~swapper/vendor/codeigniter4/framework/system/Config/Factories.php on line 129
{23-01-05 23:52}MoosacreM1ProMax:/Users/Shared/dev/~swapper moo% composer create-project codeigniter4/appstarter swapper
```

kenjis:
**Description**
Closes #7055

- remove Autoloader::sanitizeFilename() checking
- deprecate Autoloader::sanitizeFilename()

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Autoloader/Autoloader.php
+++ b/system/Autoloader/Autoloader.php
@@ -306,8 +306,6 @@ class Autoloader
      */
     protected function includeFile(string $file)
     {
-        $file = $this->sanitizeFilename($file);
-
         if (is_file($file)) {
             include_once $file;
 
@@ -327,6 +325,8 @@ class Autoloader
      * and end of filename.
      *
      * @return string The sanitized filename
+     *
+     * @deprecated No longer used. See https://github.com/codeigniter4/CodeIgniter4/issues/7055
      */
     public function sanitizeFilename(string $filename): string
     {
--- a/user_guide_src/source/changelogs/v4.4.0.rst
+++ b/user_guide_src/source/changelogs/v4.4.0.rst
@@ -94,11 +94,18 @@ Changes
 
 - **Config:** The deprecated Cookie items in **app/Config/App.php** has been removed.
 - **DownloadResponse:** When generating response headers, does not replace the ``Content-Disposition`` header if it was previously specified.
+- **Autoloader:** Before v4.4.0, CodeIgniter autoloader did not allow special
+  characters that are illegal in filenames on certain operating systems.
+  The symbols that can be used are ``/``, ``_``, ``.``, ``:``, ``\`` and space.
+  So if you installed CodeIgniter under the folder that contains the special
+  characters like ``(``, ``)``, etc., CodeIgniter didn't work. Since v4.4.0,
+  this restriction has been removed.
 
 Deprecations
 ************
 
 - **Entity:** ``Entity::setAttributes()`` is deprecated. Use ``Entity::injectRawData()`` instead.
+- **Autoloader:** ``Autoloader::sanitizeFilename()`` is deprecated.
 
 Bugs Fixed
 **********
--- a/user_guide_src/source/installation/installing_composer.rst
+++ b/user_guide_src/source/installation/installing_composer.rst
@@ -44,9 +44,12 @@ The command above will create a **project-root** folder.
 If you omit the "project-root" argument, the command will create an
 "appstarter" folder, which can be renamed as appropriate.
 
-.. note:: CodeIgniter autoloader does not allow special characters that are illegal in filenames on certain operating systems.
+.. note:: Before v4.4.0, CodeIgniter autoloader did not allow special
+    characters that are illegal in filenames on certain operating systems.
     The symbols that can be used are ``/``, ``_``, ``.``, ``:``, ``\`` and space.
-    So if you install CodeIgniter under the folder that contains the special characters like ``(``, ``)``, etc., CodeIgniter won't work.
+    So if you installed CodeIgniter under the folder that contains the special
+    characters like ``(``, ``)``, etc., CodeIgniter didn't work. Since v4.4.0,
+    this restriction has been removed.
 
 .. important:: When you deploy to your production server, don't forget to run the
     following command::
--- a/user_guide_src/source/installation/installing_manual.rst
+++ b/user_guide_src/source/installation/installing_manual.rst
@@ -22,9 +22,12 @@ Installation
 Download the `latest version <https://github.com/CodeIgniter4/framework/releases/latest>`_,
 and extract it to become your project root.
 
-.. note:: CodeIgniter autoloader does not allow special characters that are illegal in filenames on certain operating systems.
+.. note:: Before v4.4.0, CodeIgniter autoloader did not allow special
+    characters that are illegal in filenames on certain operating systems.
     The symbols that can be used are ``/``, ``_``, ``.``, ``:``, ``\`` and space.
-    So if you install CodeIgniter under the folder that contains the special characters like ``(``, ``)``, etc., CodeIgniter won't work.
+    So if you installed CodeIgniter under the folder that contains the special
+    characters like ``(``, ``)``, etc., CodeIgniter didn't work. Since v4.4.0,
+    this restriction has been removed.
 
 Initial Configuration
 =====================


paulbalandan (approved):
I agree to this solution.

samsonasik (approved):

lonnieezell:
I can't think of a specific scenario of the top of my head, but doesn't this open a security risk that any filename could then be passed in, including ones that have special meanings in the OS? Seems like it would be safer to allow the character that caused an issue in iCloud and not just nuke the whole safety feature....

It would be one thing to open this up just for the dev server but this is removing it from all autoloading. Have you done any research into if this poses an issue?

kenjis:
Good question!
Just because we can't find a way to attack doesn't mean it is truly unattackable. 
So I think we should consider this carefully.

First, this concerns autoloading of classes. Specifically, it takes a class name as input,
and looks for the class file to load and load it if found.

The class name is often provided by the developer in the source code. In that case, there is no risk.

Even if the attacker can specify the class name, it can only load files matching the class name that
exist under the (PSR-4) source directory configured in the autoloader.

Can the class name be crafted to allow directory traversal? No, PHP checking protects it.
See https://3v4l.org/SUUZe

What if there is a vulnerability that allows files to be uploaded to the source directory? 
In that case, an attacker would be able to execute arbitrary code, 
but this filename sanitization check would not prevent that.

In my opinion, this filename checking does not particularly mitigate security risks.

kenjis:
> doesn't this open a security risk that any filename could then be passed in, including ones that have special meanings in the OS?

What is the special meaning?
In this case, the file path is generated from the passed class name and loaded. 
This means that it is a special meaning as a file path.
E.g., `.` that means the current directory, or `~` that means home directory in Unix.
They are a part of the file path anyway.

Originally, the paths to be searched are restricted to directories in the defined namespaces.
E.g., `APPPATH . 'classname'`.
And PHP checks the classname now.

> in newer versions, PHP organization patched the class_exists function to only accept strings that contain valid characters for a class name. Testing all the available versions, the only vulnerable one is PHP 5.3 and older.
https://www.nc-lp.com/blog/breaking-php-internals

Even if you do not trust the classname checks made by PHP, then it is safe to simply make checks 
regarding class names.

kenjis:
One case where this filename sanitization check is useful is to prevent developers
 from using characters in file paths that are not available on other OS. 
Portability is guaranteed.

In such cases, however, it is sufficient to check only the paths in the project directory. 
The current check is excessive because it checks the full path.

However, I don't think there are many cases where characters that cannot be used 
on other OS are used in class file paths. So I don't see much need to do checking.

lonnieezell:
Did a little research and the only time I could find that anything might happen is if another package that was included via Composer got corrupted, as has happened in npm and pear in the past. However, if we are restricting to within known namespaces already then we should be safe in those instances. 

I think we're good.

MGatner (approved):

kenjis:
I believe this change is safe. But if anyone finds a security issue,
feel free to report: https://github.com/codeigniter4/CodeIgniter4/blob/develop/SECURITY.md

PR #7301, opened 2023-02-24
feat: [SQLSRV] getFieldData() supports nullable

kenjis:
**Description**
- `getFieldData()` supports nullable

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

Issue #7235, opened 2023-02-09
Bug: [Forge] addColumn() does not set NOT NULL by default


kenjis:
### PHP Version

8.1

### CodeIgniter4 Version

`develop` (5ffe7321776045630c2309b9f6fd0b75a64eb179) after 4.3.1

### CodeIgniter4 Installation Method

Git

### Which operating systems have you tested for this bug?

macOS

### Which server did you use?

cli-server (PHP built-in webserver)

### Database

MySQL 5.7

### What happened?

`addColumn()` does not set NOT NULL by default.

### Steps to Reproduce

```php
<?php

namespace App\Database\Migrations;

use CodeIgniter\Database\Migration;

class CreateTableTestTable extends Migration
{
    public function up()
    {
        $this->forge->addField(
            [
                'col1' => ['type' => 'VARCHAR', 'constraint' => 255],
                'col2' => ['type' => 'VARCHAR', 'constraint' => 255, 'null' => true],
                'col3' => ['type' => 'VARCHAR', 'constraint' => 255, 'null' => false],
            ]
        );
        $this->forge->createTable('test_table');
    }

    public function down()
    {
    }
}
```
```php
<?php

namespace App\Database\Migrations;

use CodeIgniter\Database\Migration;

class AddColumnToTestTable extends Migration
{
    public function up()
    {
        $this->forge->addColumn('test_table', [
            'col4' => ['type' => 'VARCHAR', 'constraint' => 255],
            'col5' => ['type' => 'VARCHAR', 'constraint' => 255, 'null' => true],
            'col6' => ['type' => 'VARCHAR', 'constraint' => 255, 'null' => false],
        ]);
    }

    public function down()
    {
    }
}
```
```
$ php spark migrate
```
```sql
create table test_table
(
    col1 varchar(255) not null, -- NOT NULL by default
    col2 varchar(255) null,
    col3 varchar(255) not null,
    col4 varchar(255) null, -- NULL by default
    col5 varchar(255) null,
    col6 varchar(255) not null
)
    charset = utf8;
```

### Expected Output

```
    col4 varchar(255) not null,
```

### Anything else?

> The addColumn() method is used to modify an existing table. It accepts the same field array as above, and can be used for an unlimited number of additional fields.
https://codeigniter4.github.io/userguide/dbmgmt/forge.html#forge-addcolumn

And in the *above*:
> null/true : to generate “null” in the field definition. Without this, the field will default to “NOT null”.
[...]

PR #7319, opened 2023-03-01
test: Forge::addColumn() and null

kenjis:
~~Needs #7301~~

**Description**
See #7235

- add test for `Forge::addColumn()` and null

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/tests/system/Database/Live/ForgeTest.php
+++ b/tests/system/Database/Live/ForgeTest.php
@@ -827,6 +827,41 @@ final class ForgeTest extends CIUnitTestCase
         $this->assertSame('username', $fieldNames[1]);
     }
 
+    public function testAddColumnNull()
+    {
+        $this->forge->dropTable('forge_test_table', true);
+
+        $this->forge->addField([
+            'col1' => ['type' => 'VARCHAR', 'constraint' => 255],
+            'col2' => ['type' => 'VARCHAR', 'constraint' => 255, 'null' => true],
+            'col3' => ['type' => 'VARCHAR', 'constraint' => 255, 'null' => false],
+        ]);
+        $this->forge->createTable('forge_test_table');
+
+        $this->forge->addColumn('forge_test_table', [
+            'col4' => ['type' => 'VARCHAR', 'constraint' => 255],
+            'col5' => ['type' => 'VARCHAR', 'constraint' => 255, 'null' => true],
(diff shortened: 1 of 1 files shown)


tests/system/Database/Live/ForgeTest.php
```diff
@@ -825,6 +826,41 @@ public function testAddColumn()
...
+        $this->db->resetDataCache();
+
+        $col1 = $this->getMetaData('col1', 'forge_test_table');
+        $this->assertFalse($col1->nullable);
+        $col2 = $this->getMetaData('col2', 'forge_test_table');
+        $this->assertTrue($col2->nullable);
+        $col3 = $this->getMetaData('col3', 'forge_test_table');
+        $this->assertFalse($col3->nullable);
+        $col4 = $this->getMetaData('col4', 'forge_test_table');
+        $this->assertTrue($col4->nullable);
```
michalsn, line 855: Since `$col1` is `false` from a logical point of view, I think `$col4` should be `false` too. Thoughts?

kenjis: See #7235
I thought that, but if we change it, it is a breaking change.

michalsn: Yes, fair point but in this case, I would treat it as a bug. 
API should be predictable and having a different behavior for the "same thing" is something we should avoid.

michalsn: Okay, I'll be fine with documenting this and leaving it to be fixed in the next major version.

@codeigniter4/database-team any thoughts on this one?

kenjis: It is inconsistent to have different behavior when creating and adding, but I am a little doubtful that it is worth changing in v5.

However, if we decide to change it, I would like to include that in the documentation.

@codeigniter4/database-team Opinions?

MGatner: I am ambivalent.

kenjis:
Rebased to resolve conflicts.

tests/system/Database/Live/ForgeTest.php
kenjis: It seems nobody has strong opinion.

kenjis:
This PR just adds test for the current implementation.
Merge.

PR #7322, opened 2023-03-01
Fix @param null on RequestInterface::getServer() $filter argument

samsonasik:
**Description**

Looking at the usage, the `$filter` can be null or integer, eg: 

```php
$referer = $_SESSION['_ci_previous_url'] ?? Services::request()->getServer('HTTP_REFERER', FILTER_SANITIZE_URL);
```

or

```php
$request->getServer(['SERVER_PROTOCOL', 'REQUEST_URI']);
```

so it seems can use `@param string|int|null` for it.

**Checklist:**
- [x] Securely signed commits

### DIFF
--- a/system/HTTP/RequestInterface.php
+++ b/system/HTTP/RequestInterface.php
@@ -40,8 +40,8 @@ interface RequestInterface extends OutgoingRequestInterface
      * Fetch an item from the $_SERVER array.
      * Supplied by RequestTrait.
      *
-     * @param string   $index  Index for item to be fetched from $_SERVER
-     * @param int|null $filter A filter name to be applied
+     * @param string          $index  Index for item to be fetched from $_SERVER
+     * @param int|string|null $filter A filter name to be applied
      *
      * @return mixed
      */
--- a/system/HTTP/RequestInterface.php
+++ b/system/HTTP/RequestInterface.php
@@ -40,8 +40,8 @@ interface RequestInterface extends OutgoingRequestInterface
      * Fetch an item from the $_SERVER array.
      * Supplied by RequestTrait.
      *
-     * @param string $index  Index for item to be fetched from $_SERVER
-     * @param null   $filter A filter name to be applied
+     * @param string   $index  Index for item to be fetched from $_SERVER
+     * @param int|null $filter A filter name to be applied
      *
      * @return mixed
      */
--- a/system/HTTP/RequestInterface.php
+++ b/system/HTTP/RequestInterface.php
@@ -40,8 +40,8 @@ interface RequestInterface extends OutgoingRequestInterface
      * Fetch an item from the $_SERVER array.
      * Supplied by RequestTrait.
      *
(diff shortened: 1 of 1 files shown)


system/HTTP/RequestInterface.php
```diff
@@ -40,8 +40,8 @@ public function isValidIP(string $ip, ?string $which = null): bool;
      * Fetch an item from the $_SERVER array.
      * Supplied by RequestTrait.
      *
-     * @param string $index  Index for item to be fetched from $_SERVER
-     * @param null   $filter A filter name to be applied
+     * @param string          $index  Index for item to be fetched from $_SERVER
+     * @param int|string|null $filter A filter name to be applied
```
kenjis, line 44: Why `string`?

kenjis:
The `$filter` will be passed to `filter_var()`:
https://github.com/codeigniter4/CodeIgniter4/blob/2e1761337668b7c1c5887a619859fa9864ff4161/system/HTTP/RequestTrait.php#L268
and the parameter is `int`: https://www.php.net/filter_var

system/HTTP/RequestInterface.php
samsonasik: I found the usage:

```
$request->getServer(['SERVER_PROTOCOL', 'REQUEST_URI']);
```

On find in code

kenjis: It it the first parameter as an array. The `$filter` is null.

system/HTTP/RequestInterface.php
```diff
@@ -40,8 +40,8 @@ public function isValidIP(string $ip, ?string $which = null): bool;
      * Fetch an item from the $_SERVER array.
      * Supplied by RequestTrait.
      *
-     * @param string $index  Index for item to be fetched from $_SERVER
-     * @param null   $filter A filter name to be applied
+     * @param string          $index  Index for item to be fetched from $_SERVER
```
kenjis, line 43: `array|string|null`

system/HTTP/RequestInterface.php
samsonasik: Ok, updated https://github.com/codeigniter4/CodeIgniter4/pull/7322/commits/d6c8e5a5eb96ae42c978ec5a9ab14b225c966742

system/HTTP/RequestInterface.php
samsonasik: Ok, updated https://github.com/codeigniter4/CodeIgniter4/pull/7322/commits/d6c8e5a5eb96ae42c978ec5a9ab14b225c966742

kenjis (approved):
LGTM! Thank you.

paulbalandan (approved):
Need to rebase on top of latest develop to remove Kint changes.

samsonasik:
@paulbalandan rebased :+1:

paulbalandan (approved):

samsonasik:
@kenjis @paulbalandan thank you for the review

PR #7337, opened 2023-03-06
Fixed: [Kint] Bool value exception while file validation

DhPandya:
**Description**
When a file validation run fails, $this->mtime is a bool and not null. So the condition got true and the date() function throws the exception. This is the case when you use the validation like below.

```php
 'userfile' => [
    'label' => 'Image File',
    'rules' => [
          'uploaded[userfile]',
          'is_image[userfile]',
          'mime_in[userfile,image/jpg,image/jpeg,image/gif,image/png,image/webp]',
          'max_size[userfile,100]',
          'max_dims[userfile,1024,768]',
    ],
]
```

Screenshot (52)

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/system/ThirdParty/Kint/Zval/Representation/SplFileInfoRepresentation.php
+++ b/system/ThirdParty/Kint/Zval/Representation/SplFileInfoRepresentation.php
@@ -181,7 +181,7 @@ class SplFileInfoRepresentation extends Representation
 
     public function getMTime(): ?string
     {
-        if (null !== $this->mtime) {
+        if ($this->mtime) {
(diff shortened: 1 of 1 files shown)


kenjis:
Thank you for sending PR!

But unfortunately the code is not ours.
We do not accept this PR.

kenjis:
I reported to Kint. https://github.com/kint-php/kint/issues/407

kenjis:
Fixed by #7358

PR #7341, opened 2023-03-07
fix: use first exception in exceptionHandler()

### LINKED ISSUE #7339: Bug: cli request is confusing
### PHP Version

8.2

### CodeIgniter4 Version

4.1.3

Linux

### Which server did you use?

apache

### Database

Mongodb

### What happened?

calling a webpage via cli is a pain in ci4. we faced a lot of issues when tried  running a phpcode via terminal. 

1.  i had a fatal error in my initController function of parent class.but this was not shown to me instead it stopped the execution at my __destruct function complaining different issue.it took me a lot of time to identify the real problem
2. $this->request->uri is public in normal request(via browser) . but when running via a terminal it throws error Cannot access protected property CodeIgniter\HTTP\CLIRequest::$uri

 

### Steps to Reproduce

[ParentController.php.txt](https://github.com/codeigniter4/CodeIgniter4/files/10899777/ParentController.php.txt)
[Test.php.txt](https://github.com/codeigniter4/CodeIgniter4/files/10899782/Test.php.txt)

(please rename the extension to .php as it wont allow me to upload this files) 
run test.php

### Expected Output

1. run via terminal/browser getting error at __destruct . it should have stoped at initcontroller right?
2. comment the line
`$this->dd(); //no such function fatal error`
run via terminal it should be working like running via a browser. now getting error: Cannot access protected property CodeIgniter\HTTP\CLIRequest::$uri

### Anything else?

please suggest whether this was an error

kenjis:
**Description**
Fixes #7339

Now the last Exception is shown, but showing the first Exception helps for debugging.
This PR shows the first Exception.

```php
<?php

namespace App\Controllers;

use Psr\Log\LoggerInterface;

class Tests extends BaseController
{
    public function initController(RequestInterface $request, ResponseInterface $response, LoggerInterface $logger)
    {
        // Do Not Edit This Line
        parent::initController($request, $response, $logger);

        $this->xx(); // Call to undefined method App\Controllers\Tests::xx()
    }

    public function __destruct()
    {
        dd($this->info); // Undefined property: App\Controllers\Tests::$info
    }

    public function index()
    {
    }
}
```
Before:
```
$ php public/index.php tests


[ErrorException]

Undefined property: App\Controllers\Tests::$info

at APPPATH/Controllers/Tests.php:21

Backtrace:
  1    APPPATH/Controllers/Tests.php:21
       CodeIgniter\Debug\Exceptions()->errorHandler(2, 'Undefined property: App\\Controllers\\Tests::$info', '/Users/kenji/work/codeigniter/official/CodeIgniter4/app/Controllers/Tests.php', 21)

  2    SYSTEMPATH/CodeIgniter.php:490
       App\Controllers\Tests()->__destruct()

  3    SYSTEMPATH/CodeIgniter.php:368
       CodeIgniter\CodeIgniter()->handleRequest(null, Object(Config\Cache), false)

  4    FCPATH/index.php:67
       CodeIgniter\CodeIgniter()->run()
```
After:
```
$ php public/index.php tests


[Error]

Call to undefined method App\Controllers\Tests::xx()

at APPPATH/Controllers/Tests.php:16

Backtrace:
  1    SYSTEMPATH/CodeIgniter.php:907
       App\Controllers\Tests()->initController(Object(CodeIgniter\HTTP\CLIRequest), Object(CodeIgniter\HTTP\Response), Object(CodeIgniter\Log\Logger))

  2    SYSTEMPATH/CodeIgniter.php:490
       CodeIgniter\CodeIgniter()->createController()

  3    SYSTEMPATH/CodeIgniter.php:368
       CodeIgniter\CodeIgniter()->handleRequest(null, Object(Config\Cache), false)

  4    FCPATH/index.php:67
       CodeIgniter\CodeIgniter()->run()
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Debug/Exceptions.php
+++ b/system/Debug/Exceptions.php
@@ -119,6 +119,11 @@ class Exceptions
 
         [$statusCode, $exitCode] = $this->determineCodes($exception);
 
+        // Get the first exception.
+        while ($prevException = $exception->getPrevious()) {
+            $exception = $prevException;
+        }
+
         if ($this->config->log === true && ! in_array($statusCode, $this->config->ignoreCodes, true)) {
             log_message('critical', "{message}\nin {exFile} on line {exLine}.\n{trace}", [
                 'message' => $exception->getMessage(),


paulbalandan:
Would it better to display all exceptions in the stack? Something like:

[Exception 1]
Text
Backtrace:
...

Caused by:

[Exception 2]
Text
Backtrace:
...

Caused by:
...
...

kenjis:
I thought about that, but I think one would be simpler.

By the way, I don't know why the first exception is now hidden in the sample code.
Why is it so?

paulbalandan:
Not sure about that. Maybe because PHP still called `__destruct` before calling the exception handler?

kenjis:
> Maybe because PHP still called __destruct before calling the exception handler?

It seems like that.
I guess that if a class fails to instantiate, the destructor is called.

MGatner (approved):

PR #7342, opened 2023-03-07
docs: add note for addColumn() and NULL

### LINKED ISSUE #7235: Bug: [Forge] addColumn() does not set NOT NULL by default

kenjis:
**Description**
Fixes #7235

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/dbmgmt/forge.rst
+++ b/user_guide_src/source/dbmgmt/forge.rst
@@ -254,6 +254,9 @@ The ``addColumn()`` method is used to modify an existing table. It
 accepts the same field array as :ref:`Creating Tables <adding-fields>`, and can
 be used to add additional fields.
 
+.. note:: Unlike when creating a table, if ``null`` is not specified, the column
+    will be ``NULL``, not ``NOT NULL``.
+
 .. literalinclude:: forge/022.php
 
 If you are using MySQL or CUBIRD, then you can take advantage of their


paulbalandan (approved):
Seems counterintuitive, but since it's the behavior since CI3 then it's better to be documented.

PR #7372, opened 2023-03-23
composer.lock removed from .gitignore

shahghasiadil:
> once composer creates lock file for the first time it locks the dependencies ! 
once you push it to git composer.lock is important becuase without composer.lock file composer updates all dependencies!

I have faced with that issue my project is created with CIv4.1 and  someone else cloned my project 
and he has faced with issue of no composer.lock file in git repo and composer updates all dependecies

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,7 +85,6 @@ phpunit
 # Composer
 #-------------------------
 vendor/
-composer.lock
 
 #-------------------------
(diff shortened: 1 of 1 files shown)


kenjis:
Don't use this repository for your app.
Use appstarter or something.
See https://codeigniter4.github.io/CodeIgniter4/installation/installing_composer.html

PR #7377, opened 2023-03-24
refactor: Security::getPostedToken()

### LINKED ISSUE #7374: Bug: CSRF Token Check Order is Mismatched in the Documentaion and the Implementation
### PHP Version

8.1

### CodeIgniter4 Version

4.3.2

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

As per the [documentation](https://codeigniter4.github.io/userguide/libraries/security.html?highlight=security#the-order-of-token-sent-by-users)

> The order of checking the availability of the CSRF token is as follows:
> 
> 1. $_POST array
> 2. HTTP header
> 3. php://input (JSON request) 

However, when you notice the [implementation](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Security/Security.php#L340), the order of execution is:

1. If HTTP header is present than assign it to the `tokenName` variable.
2. Else treat it as a JSON request and perform `json_decode` and assign token to variable if it exists.
3. Now check the same in the POST array and return it otherwise otherwise return from either 1, or 2.

So even the POST array contains the token, the header is still checked.
Further, if it is absent then a slower `json_decode` is performed even through the body is empty.
So while the POST array contains the token, if will be still fetched at the last.

Hence, it feels like this implementation doesn't confirm the behavior mentioned in the documentation.

### Steps to Reproduce

Kindly review the documentation and the implementation as given in links.

### Expected Output

Either documentation should be updated or code execution should be corrected.

### Anything else?

_No response_

kenjis:
**Description**
Closes #7374

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Security/Security.php
+++ b/system/Security/Security.php
@@ -342,20 +342,23 @@ class Security implements SecurityInterface
         assert($request instanceof IncomingRequest);
 
         // Does the token exist in POST, HEADER or optionally php:://input - json data.
-        if ($request->hasHeader($this->headerName) && ! empty($request->header($this->headerName)->getValue())) {
-            $tokenName = $request->header($this->headerName)->getValue();
-        } else {
-            $body = (string) $request->getBody();
-            $json = json_decode($body);
 
-            if ($body !== '' && ! empty($json) && json_last_error() === JSON_ERROR_NONE) {
-                $tokenName = $json->{$this->tokenName} ?? null;
-            } else {
-                $tokenName = null;
-            }
+        if ($tokenName = $request->getPost($this->tokenName)) {
+            return $tokenName;
         }
 
-        return $request->getPost($this->tokenName) ?? $tokenName;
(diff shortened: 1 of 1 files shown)


samsonasik (approved):

iRedds:
Why not use Request::getJsonVar()?

michalsn (approved):
We want to save the particular order of the checks (POST, Header, JSON) for performance reasons, so this is good.

samsonasik:
Thank you @kenjis @iRedds @michalsn  👍

kenjis:
@iRedds If you think using getJsonVar() does not change anything, you can send another PR to refactor.
If the behavior is exactly the same, the code would be cleaner to use getJsonVar().
I have not seen a detailed implementation of getJsonVar().

Issue #7308, opened 2023-02-26
Bug: cannot use Session in Config\App


kenjis:
CodeIgniter version: 4.3.2

See https://github.com/codeigniter4/CodeIgniter4/issues/4297#issuecomment-813110373

```diff
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -442,4 +442,11 @@ class App extends BaseConfig
      * @see http://www.w3.org/TR/CSP/
      */
     public bool $CSPEnabled = false;
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        session();
+    }
 }
```

```console
$ php spark
PHP Fatal error:  Uncaught Error: Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '256' frames in /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Config/Factories.php:98
Stack trace:
#0 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Config/Factories.php(98): array_shift(Array)
#1 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Common.php(216): CodeIgniter\Config\Factories::__callStatic('config', Array)
#2 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Config/BaseConfig.php(62): config('Modules')
#3 /Users/kenji/work/codeigniter/official/CodeIgniter4/app/Config/App.php(448): CodeIgniter\Config\BaseConfig->__construct()
#4 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Config/Factories.php(129): Config\App->__construct()
#5 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Common.php(216): CodeIgniter\Config\Factories::__callStatic('config', Array)
#6 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Config/Services.php(644): config('App')
#7 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Config/BaseService.php(252): CodeIgniter\Config\Services::session(NULL, false)
#8 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Config/BaseService.php(193): CodeIgniter\Config\BaseService::__callStatic('session', Array)
#9 /Users/kenji/work/codeigniter/official/CodeIgniter4/system/Config/Services.php(641): CodeIgniter\Config\BaseService::getSharedInstance('session', NULL)
[...]

PR #7380, opened 2023-03-25
Clean router config

lonnieezell:
The router file had way too much information in it, from walls of comments with descriptions of settings, to settings, and including additional files. This PR cleans this up so that the Routes.php file is only for adding routes. It does this by: 

1. Creating a new config file, `app/Config/Routing.php`. All of the settings that were set by default within the old `Routes` file has been moved to this config file.
2. <strike>The `Routes` file has been moved to `app/Routes.php` to make it more prominent, and reduce confusion by having two closely-named files. </strike>
3. ~~`Routes.php` files within modules are now expected to be at the root of the module directory, though this is easily changed.~~
4. A new setting within the `Routing` config file allows existing applications to simply point to the old `Routes` file if they don't want to move it. This also allows the user to specify multiple routes files to include if they want to separate routes logically, like into an `api` and  `web` files, etc.
5. ~~A new setting within the `Routing` config file allows setting the path to module config files.~~ ~~This now also defaults to `{module}/Routes.php`.~~

UPGRADING EXISTING PROJECTS: 
 
The process for upgrading an existing project should be simple: 
1. copy the `Routing.php` config file into the project folder.
2. ~~edit the `$routeFiles` to point to the existing file (or move the existing file into the `app` folder)~~
3. ~~edit the `$modulePath` to be `Config/Routes.php` (or move any existing files up to the module root folder)~~

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


kenjis:
This is not a trivial change. It should go to `4.4` branch.

@lonnieezell Don't use `wip` as a commit message, please.
It does not tell anything.

iRedds:
I don't think this PR is a good decision.

**About configs.**
In CodeIgniter, configs are not cached, which means that with each request, each loaded config inherited from CodeIgniter\Config will try to find the class properties in .env.
+1 to overhead

Almost all configs used in class dependencies are used as DTO, overriding class properties.
The meaning of using the config-class is lost. The same can be obtained using an array.

**About routes.**
All settings from the routes file can be removed, since they duplicate the state of the properties of the RouteCollection class. Except for the namespace, but it can be defined in the class by default.
The framework has documentation, so extra comments can be removed from the routes file.

Also, the presence of two files associated with routing can cause confusion.
Especially in modules, the position of the file with routes remains the same.

kenjis:
> Almost all configs used in class dependencies are used as DTO, overriding class properties.
The meaning of using the config-class is lost. The same can be obtained using an array.

What is the meaning of using the config-class?

And I know you loves array config, but we have been using config-class already.
Changing to an array would be a breaking change. Is it worth the change?

kenjis:
> Especially in modules, the position of the file with routes remains the same.

This is a good point.
If it's in `Acme/Config/Routes.php` in a module, but in `app/Routes.php` in the app, that's odd.

lonnieezell:
>  It should go to 4.4 branch.

Good point! I'll fix that up. 

> All settings from the routes file can be removed...

That's true, I could just remove those settings all together and document how to call the methods within the Routes file. I have two issues with that, though. 

First - the current Routes file isn't what I would consider a config file, honestly. It was left in that folder for purely historical reasons during the rewrite. 

Second - this is the first of a two-step process to make the Router more flexible, converting it to use Handlers for the 2 distinct routing systems we have, but expandable with dev-provided routers. I've heard devs searching for a way to get routes dynamically from the database for years, and this would allow them to easily handle that if they needed, as well as allow others to create other routing if they needed. Knowing this, we need a place to configure the handlers that are in use. This could be handled in the Services class, I guess, though that doesn't feel right when you look at other portions that use handlers, like the Cache, Session, or Database.

> Especially in modules, the position of the file with routes remains the same.

Good point. I missed that. I see three options: 

1. Move `Routes` file back to the Config folder
2. Look for module `Routes` file in the main module
3. Add a separate config item that specifies where they should exist in Modules, allowing a move to the module root, but an easy setting during an upgrade for existing projects. 

Thoughts?

iRedds:
@kenjis You're exaggerating when you say I love arrays. I believe that arrays are the most optimal format for configuration. A class has only one advantage - it can be specified as a dependency.

Of course, in this version of the framework, changing classes to arrays is unacceptable, but I still hope that a discussion about switching to arrays is possible in the next version.

If we talk about the use of classes, then in my opinion a very strange approach is used.
[php]
// instead of
$this->a = $config->a;
$this->b = $config->b;

// better
$this->config = $config;
[/php]
Since the config is in dependencies, without it we will not be able to create an instance of the class. Then it makes no sense to duplicate class properties in the config and we can use the config directly.

iRedds:
> That's true, I could just remove those settings all together and document how to call the methods within the Routes file. I have two issues with that, though.
> 
> First - the current Routes file isn't what I would consider a config file, honestly. It was left in that folder for purely historical reasons during the rewrite.

Precisely because they tried to make CI4 more like CI3 and used its code base, we are now in a lot of trouble.
Let's take the config out of the application directory, and the routes into a separate directory.

> Second - this is the first of a two-step process to make the Router more flexible, converting it to use Handlers for the 2 

If you are talking about flexibility, then it already exists. Services! Though I still think ServiceContainer and Dependency Autowiring would be the best solution.

If the developer wants to change the behavior of the router, then he can replace the service with his own implementation.
When there is a lot of flexibility where it is not needed, this is a problem.

Take for example the Model. It serves arrays, stdClass and Entity. Everything seems to be fine, but when adding features, you have to service all return types. Which increases the code base and complicates maintenance.

> > Especially in modules, the position of the file with routes remains the same.
> 
> Good point. I missed that. I see three options:
 
I think a single configuration point should be used for modules. Something similar to Registrars and Laravel ServiceProvider.

```php
namespace Module;

class Config
{
    public function boot()
    {
        // This is where we define routes, filters, and other available settings.
        // Instead of separate files that repeat the names of the application configs.
    }
}
```

kenjis:
@iRedds 
The following code is needed because the `$config` has items that the class does not need.
```php
$this->a = $config->a
``` 
A Config class should be a dependency for one class. If it is used in more than one class, it is not good.
If one Config class is for one class, we can use:
```php
$this->config = $config;
```

I did it in #7255 Why don't you approve it?

kenjis:
> Look for module Routes file in the main module

What is the main module?

I feel `app/Routes.php` is strange. Why no directory?
What if we have multiple routes files?

iRedds:
@kenjis To be honest, I don't quite understand what happened. That is, in CI3 the session settings were in the main config and everything worked, but in CI4 it does not work.

kenjis:
@iRedds That is because CI4 uses Config classes.

 In the current implementation, the Session class constructor requires Config\App object. If you want to use Session object in Config\App to set a property, It will be an infinite loop.
See this error: https://github.com/codeigniter4/CodeIgniter4/issues/7308#issue-1600233210

This issue does not occur in array config because there is no instantiation.
It may be one advantage of array config.

iRedds:
@kenjis Forbid using the constructor in configs.
Otherwise, any class called in Config\App::__construct() that references Config\App in its constructor will cause an infinite loop.

kenjis:
@iRedds I created #7388

lonnieezell:
> A class has only one advantage - it can be specified as a dependency.

The original reason Narf and I discussed using classes as config files years ago was that it could be loaded instantly from anywhere. Obviously, the framework has expanded and the Factory class now handles that to ensure it's only ever loaded once for performance, so it might be valid to rethink for v5. But that's not relevant to this PR. 

> I think a single configuration point should be used for modules. 

Interesting idea, but it feels more relevant to a plug-in system for a CMS or other app. This would warrant other discussion, but is a huge breaking change and not something we can handle within this PR. 

> I feel app/Routes.php is strange. Why no directory? 

I was hesitant to create a new directory for a single file. If we had logic around the handling of different specific route file types, like Laravel has bulit-in logic around `web.php` and `api.php` route files, than it would make sense. As it stands, it is only a single file. 

> What if we have multiple routes files?

Their locations can all be specified within the config file. And the current one can be moved to join them just by changing its path in the config file. 

> If the developer wants to change the behavior of the router, then he can replace the service with his own implementation. When there is a lot of flexibility where it is not needed, this is a problem.

To some extent I agree with this direction. The three problems I'm struggling with are:

1. It's not consistent with other portions of the framework. Lack of consistency is also a problem. And I'm sure you'll now point out areas that aren't quite consistent, which I'm sure we'd all love to fix at some point, but I'd rather not create MORE inconsistencies. 
2. Like I mentioned, a large part of the reason for doing this (though not the only one) was to split the main routing and the auto-routing into different handlers. I would hate to replace the current functionality of "setAutoRoute(true)" to become one of "copy the Service over to the App space, and modify it to add in these two handlers". That's too large of a BC break for a point release. 
3. I would really love for the Routes file itself to just be route definitions and not have config settings in it also. I know you mentioned removing the docs would clean the file up, and it would to some extent, but I think it is mixing concerns right now that it shouldn't be. 

@MGatner @michalsn what are your thoughts on these changes?

kenjis:
> It turns out that these settings in the config will be redundant for different handlers.

Each route handler has its own settings.

> On the other hand, based on the experience of CI3, even with autorouting, it is necessary to be able to define routes manually. Is it possible to separate handlers in this case?

In CI3 auto-routing is primary routing system, and there is no way to turn it off.
And it is not flexible as much as defined routes, so defined routes are necessary.

But if you define all routes, you don't need to use auto-routing at all.
CI4 can disable auto-routing.
If your URI design is very simple, you may not need to use defined routes at all.
If a site does not need defined routes, I would like to disable it.

lonnieezell:
Ok, so without re-inventing the core of the framework here, what are people leaning toward? Please think solutions that are consistent within the framework as it exists now. Or should we just close this PR? I'm fine with that if it's too controversial, though it would have been nice to get input from other core team members.

samsonasik:
I prefer go to 5.0 for this change, except, deprecate existing functionality to keep working while suggesting new feature.

michalsn:
I'm a bit worried about the BC break part this PR brings when upgrading the framework, but overall I like the idea.

Routes inside the modules should be auto-discovered as before, though I'm unsure how we should specify the default location after these changes.

If we make a clear path of how to upgrade, I'm okay with including this in the v4.4 release.

kenjis:
I would like to get this into 4.4 without breaking changes as much as possible.

lonnieezell:
I've rebased to 4.4 branch. I also made the path within any "modules" configurable. This migration should be pretty painless - see the instructions I added to the PR description for details).

kenjis:
Defaulting to `{module}/Routes.php` will break all existing Composer packages with routes.

Wouldn't it be better to leave the default settings under Config?

- `app/Config/Routes.php`
- `{module}/Config/Routes.php`

It doesn't seem worth breaking all the existing packages and changing to `{module}/Routes.php`.

michalsn:
TBH, I think I would be in favor of staying with the current `Routes.php` file location.

In addition to what Kenjis wrote. The change is cosmetic, and it can really create a lot of confusion. We don't have a lot of tutorials for CI4, and such a change will make them all obsolete. For newcomers, this might be a hitting-a-wall experience.

I will have no problems with moving this file in v5 though.

kenjis:
And `Routes.php` is the config file for the Defined Route Handler. 
It is not necessary if we use only auto routing.

lonnieezell:
> TBH, I think I would be in favor of staying with the current Routes.php file location.

> In addition to what Kenjis wrote. The change is cosmetic, and it can really create a lot of confusion. We don't have a lot of tutorials for CI4, and such a change will make them all obsolete. For newcomers, this might be a hitting-a-wall experience.

> I will have no problems with moving this file in v5 though.

Fair enough. I've reverted those. And dealing with composer-loaded packages does create an issue. Leaving that change for v5 definitely makes sense. 

Thanks everyone.

user_guide_src/source/changelogs/v4.3.3.rst
```diff
@@ -16,6 +16,18 @@ SECURITY
...
+
+Message Changes
+***************
+
+Changes
+*******
+
+Deprecations
+************
+
```
kenjis, line 30: Revert, please.

user_guide_src/source/incoming/routing.rst
```diff
@@ -601,6 +600,8 @@ a valid class/method pair, just like you would show in any route, or a Closure:
 
 .. literalinclude:: routing/051.php
 
+Using the ``set404Override`` method within the routes file, you can use closures. Defining the override in the Routing file is restricted to class/method pairs.
```
kenjis, line 603:
```suggestion
Using the ``$override404`` property within the routes file, you can use closures. Defining the override in the Routing file is restricted to class/method pairs.
```

user_guide_src/source/incoming/routing.rst
```diff
@@ -585,7 +584,7 @@ When no defined route is found that matches the URI, the system will attempt to
 controllers and methods when Auto Routing is enabled.
 
 You can disable this automatic matching, and restrict routes
-to only those defined by you, by setting the ``setAutoRoute()`` option to false:
+to only those defined by you, by setting the ``$autoRoute`` option to false:
```
kenjis, line 587:
```suggestion
to only those defined by you, by setting the ``$autoRoute`` property to false:
```

user_guide_src/source/tutorial/static_pages.rst
```diff
@@ -136,10 +136,9 @@ We have made the controller. The next thing is to set routing rules.
 Routing associates a URI with a controller's method.
 
 Let's do that. Open the routing file located at
-**app/Config/Routes.php** and look for the "Route Definitions"
-section of the configuration file.
+**app/Routes.php**.
```
kenjis, line 139:
```suggestion
**app/Config/Routes.php**.
```

user_guide_src/source/changelogs/v4.3.3.rst
lonnieezell, line 30:
```suggestion
```

user_guide_src/source/incoming/routing/049.php
```diff
@@ -1,3 +1,7 @@
 <?php
 
+// In app/Config/Routing.php
+public boolean $translateURIDashes = true;
```
kenjis, line 4: This is a syntax error. If you use `*.php` file, write correct PHP code.
The same applies to the following php files.

kenjis:
```diff
diff --git a/tests/system/CodeIgniterTest.php b/tests/system/CodeIgniterTest.php
index a7f5b9bd19..f01986553e 100644
--- a/tests/system/CodeIgniterTest.php
+++ b/tests/system/CodeIgniterTest.php
@@ -22,6 +22,7 @@ use Config\App;
 use Config\Cache;
 use Config\Filters as FiltersConfig;
 use Config\Modules;
+use Config\Routing;
 use Tests\Support\Filters\Customfilter;
 
 /**
@@ -165,7 +166,7 @@ final class CodeIgniterTest extends CIUnitTestCase
         $_SERVER['argc'] = 2;
 
         // Inject mock router.
-        $routes = new RouteCollection(Services::locator(), new Modules());
+        $routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         $routes->setAutoRoute(false);
         $routes->set404Override(static function () {
             echo '404 Override by Closure.';
diff --git a/tests/system/CommonFunctionsTest.php b/tests/system/CommonFunctionsTest.php
index cf0555fba3..92c853f4cf 100644
--- a/tests/system/CommonFunctionsTest.php
+++ b/tests/system/CommonFunctionsTest.php
@@ -32,6 +32,7 @@ use Config\App;
 use Config\Cookie;
 use Config\Logger;
 use Config\Modules;
+use Config\Routing;
 use Config\Services;
 use Kint;
 use RuntimeException;
@@ -124,7 +125,8 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $response = $this->createMock(Response::class);
         $routes   = new RouteCollection(
             Services::locator(),
-            new Modules()
+            new Modules(),
+            new Routing()
         );
         Services::injectMock('response', $response);
         Services::injectMock('routes', $routes);
@@ -389,7 +391,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
@@ -424,7 +426,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
@@ -459,7 +461,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
```

system/Router/RouteCollection.php
```diff
@@ -257,14 +273,32 @@ public function __construct(FileLocator $locator, Modules $moduleConfig)
      *
      * @return $this
      */
-    public function loadRoutes(string $routesFile = APPPATH . 'Config/Routes.php')
+    public function loadRoutes(string $routesFile = APPPATH . 'Routes.php')
```
kenjis, line 276: Revert, please.

system/Router/RouteCollection.php
```diff
@@ -242,12 +248,22 @@ class RouteCollection implements RouteCollectionInterface
     /**
      * Constructor
      */
-    public function __construct(FileLocator $locator, Modules $moduleConfig)
+    public function __construct(FileLocator $locator, Modules $moduleConfig, Routing $routing)
```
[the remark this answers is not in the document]
lonnieezell: All of the tests are already passing related to this. 

In the same way the module config is sent it, I sent the Routing config in, so - seems like that's the proper way to do it....

user_guide_src/source/incoming/routing/049.php
lonnieezell: Why do we enforce this linting in the user guide folder? Seems a bit pedantic....

```suggestion
$translateURIDashes = true;
```

kenjis: To check and fix sample code style, we are using `*.php` files and check them.

kenjis:
If you change the method signature:
```diff
diff --git a/tests/system/CommonFunctionsTest.php b/tests/system/CommonFunctionsTest.php
index cf0555fba3..92c853f4cf 100644
--- a/tests/system/CommonFunctionsTest.php
+++ b/tests/system/CommonFunctionsTest.php
@@ -32,6 +32,7 @@ use Config\App;
 use Config\Cookie;
 use Config\Logger;
 use Config\Modules;
+use Config\Routing;
 use Config\Services;
 use Kint;
 use RuntimeException;
@@ -124,7 +125,8 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $response = $this->createMock(Response::class);
         $routes   = new RouteCollection(
             Services::locator(),
-            new Modules()
+            new Modules(),
+            new Routing()
         );
         Services::injectMock('response', $response);
         Services::injectMock('routes', $routes);
@@ -389,7 +391,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
@@ -424,7 +426,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
@@ -459,7 +461,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
diff --git a/tests/system/HTTP/RedirectResponseTest.php b/tests/system/HTTP/RedirectResponseTest.php
index 65fc0e62d0..22c3c719b7 100644
--- a/tests/system/HTTP/RedirectResponseTest.php
+++ b/tests/system/HTTP/RedirectResponseTest.php
@@ -19,6 +19,7 @@ use CodeIgniter\Test\Mock\MockIncomingRequest;
 use CodeIgniter\Validation\Validation;
 use Config\App;
 use Config\Modules;
+use Config\Routing;
 use Config\Services;
 
 /**
@@ -47,7 +48,7 @@ final class RedirectResponseTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest(
kenji:~/work/codeigniter/official/CodeIgniter4 (clean-router-config *$=)$ git diff
diff --git a/tests/system/CommonFunctionsTest.php b/tests/system/CommonFunctionsTest.php
index cf0555fba3..92c853f4cf 100644
--- a/tests/system/CommonFunctionsTest.php
+++ b/tests/system/CommonFunctionsTest.php
@@ -32,6 +32,7 @@ use Config\App;
 use Config\Cookie;
 use Config\Logger;
 use Config\Modules;
+use Config\Routing;
 use Config\Services;
 use Kint;
 use RuntimeException;
@@ -124,7 +125,8 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $response = $this->createMock(Response::class);
         $routes   = new RouteCollection(
             Services::locator(),
-            new Modules()
+            new Modules(),
+            new Routing()
         );
         Services::injectMock('response', $response);
         Services::injectMock('routes', $routes);
@@ -389,7 +391,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
@@ -424,7 +426,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
@@ -459,7 +461,7 @@ final class CommonFunctionsTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest($this->config, new URI('http://example.com'), null, new UserAgent());
diff --git a/tests/system/HTTP/RedirectResponseTest.php b/tests/system/HTTP/RedirectResponseTest.php
index 65fc0e62d0..22c3c719b7 100644
--- a/tests/system/HTTP/RedirectResponseTest.php
+++ b/tests/system/HTTP/RedirectResponseTest.php
@@ -19,6 +19,7 @@ use CodeIgniter\Test\Mock\MockIncomingRequest;
 use CodeIgniter\Validation\Validation;
 use Config\App;
 use Config\Modules;
+use Config\Routing;
 use Config\Services;
 
 /**
@@ -47,7 +48,7 @@ final class RedirectResponseTest extends CIUnitTestCase
         $this->config          = new App();
         $this->config->baseURL = 'http://example.com/';
 
-        $this->routes = new RouteCollection(Services::locator(), new Modules());
+        $this->routes = new RouteCollection(Services::locator(), new Modules(), new Routing());
         Services::injectMock('routes', $this->routes);
 
         $this->request = new MockIncomingRequest(
```

kenjis:
This test code seemed wrong from the beginning. Because it passes only when the default namespace is `\`.

```diff
--- a/tests/system/RESTful/ResourceControllerTest.php
+++ b/tests/system/RESTful/ResourceControllerTest.php
@@ -66,7 +66,7 @@ final class ResourceControllerTest extends CIUnitTestCase
 
         // Inject mock router.
         $this->routes = Services::routes();
-        $this->routes->resource('work', ['controller' => Worker::class]);
+        $this->routes->resource('work', ['controller' => '\\' . Worker::class]);
         Services::injectMock('routes', $this->routes);
 
         $config            = new App();
```

system/Router/RouteCollection.php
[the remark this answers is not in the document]
lonnieezell: It would only break code for those that have extended the RouteCollection class, which I can't imagine is that many. If you think this particular BC is too much, we can scrap this and save it for 5. Seemed like a smaller BC possibility than a number of our other breaks, unless I'm completely missing something here.

lonnieezell:
Starting to think this one might not be worth the effort for what is gained. Going to close it, but if people really want it we can always re-open.

kenjis:
Rebased and updated failed tests, and added documentation a bit.

kenjis:
I think this is ready to merge. @codeigniter4/core-team

system/Router/RouteCollection.php
```diff
@@ -263,8 +279,26 @@ public function loadRoutes(string $routesFile = APPPATH . 'Config/Routes.php')
...
+        }
+
+        // We need this var in local scope
+        // so route files can access it.
         $routes = $this;
-        require $routesFile;
+
+        foreach ($routeFiles as $routesFile) {
+            if (! is_file($routesFile)) {
+                log_message('warning', 'Routes file not found: ' . $routesFile . '.');
```
samsonasik, line 295: probably sprintf is cleaner:

```suggestion
                log_message('warning', sprintf('Routes file not found: %s.', $routesFile));
```

kenjis: Thank you for reveiw!

By the way, is the last dot necessary?

samsonasik: last dot is same with the previous usage ` . $routesFile . '.'` to keep same value.

kenjis: I mean, isn't `log_message('warning', 'Routes file not found: ' . $routesFile);` sufficient?

samsonasik: I have no opinion on that ,probably @MGatner can give opinion on it

michalsn (approved):
It seems ready to me.

system/Router/RouteCollection.php
kenjis:
```
Cast.php
    'invalidCastMethod'      => 'The "{0}" is invalid cast method, valid methods are: ["get", "set"].',
CLI.php
        'fileCreate'       => 'File created: {0}',
        'fileError'        => 'Error while creating file: "{0}"',
        'fileExist'        => 'File exists: "{0}"',
        'fileOverwrite'    => 'File overwritten: "{0}"',
        'usingCINamespace' => 'Warning: Using the "CodeIgniter" namespace will generate the file in the system directory.',
    'invalidColor'        => 'Invalid "{0}" color: "{1}".',
Cookie.php
    'invalidExpiresTime'    => 'Invalid "{0}" type for "Expires" attribute. Expected: string, integer, DateTimeInterface object.',
Core.php
    'invalidFile'                  => 'Invalid file: "{0}"',
    'invalidPhpVersion'            => 'Your PHP version must be {0} or higher to run CodeIgniter. Current version: {1}',
    'missingExtension'             => 'The framework needs the following extension(s) installed and loaded: "{0}".',
Database.php
    'invalidAllowedFields'             => 'Allowed fields must be specified for model: "{0}"',
Email.php
    'invalidAddress'       => 'Invalid email address: "{0}"',
    'attachmentMissing'    => 'Unable to locate the following email attachment: "{0}"',
    'attachmentUnreadable' => 'Unable to open this attachment: "{0}"',
    'noRecipients'         => 'You must include recipients: To, Cc, or Bcc',
    'sent'                 => 'Your message has been successfully sent using the following protocol: {0}',
    'SMTPError'            => 'The following SMTP error was encountered: {0}',
    'noSMTPAuth'           => 'Error: You must assign an SMTP username and password.',
    'failedSMTPLogin'      => 'Failed to send AUTH LOGIN command. Error: {0}',
    'SMTPAuthUsername'     => 'Failed to authenticate username. Error: {0}',
    'SMTPAuthPassword'     => 'Failed to authenticate password. Error: {0}',
    'SMTPDataFailure'      => 'Unable to send data: {0}',
    'exitStatus'           => 'Exit status code: {0}',
Encryption.php
    'authenticationFailed' => 'Decrypting: authentication failed.',
Fabricator.php
    'createFailed'      => 'Fabricator failed to insert on table "{0}": {1}',
Files.php
    'fileNotFound'      => 'File not found: "{0}"',
    'cannotMove'        => 'Could not move file "{0}" to "{1}". Reason: {2}',
Format.php
    'invalidJSON'      => 'Failed to parse JSON string. Error: {0}',
    'invalidMime'      => 'No Formatter defined for mime type: "{0}".',
HTTP.php
    'sslCertNotFound' => 'SSL certificate not found at: "{0}"',
    'curlError'       => '{0} : {1}',
    'invalidNegotiationType' => '"{0}" is not a valid negotiation type. Must be one of: media, charset, encoding, language.',
    'invalidHTTPProtocol' => 'Invalid HTTP Protocol Version. Must be one of: {0}',
    'cannotSetFilepath'      => 'When setting binary cannot set filepath: "{0}"',
    'cannotSetStatusCode'    => 'It does not support change status code for downloading. code: {0}, reason: {1}',
    'unknownStatusCode'     => 'Unknown HTTP status code provided with no message: {0}',
    'cannotParseURI'       => 'Unable to parse URI: "{0}"',
    'segmentOutOfRange'    => 'Request URI segment is out of range: "{0}"',
    'invalidPort'          => 'Ports must be between 0 and 65535. Given: {0}',
    'controllerNotFound' => 'Controller or its method is not found: {0}::{1}',
    'methodNotFound'     => 'Controller method is not found: "{0}"',
    'localeNotSupported' => 'Locale is not supported: {0}',
    'moveFailed'   => 'Could not move file "{0}" to "{1}". Reason: {2}',
    'uploadErrNoTmpDir'  => 'File could not be uploaded: missing temporary directory.',
    'invalidSameSiteSetting' => 'The SameSite setting must be None, Lax, Strict, or a blank string. Given: {0}',
Images.php
    'invalidDirection'       => 'Flip direction can be only "vertical" or "horizontal". Given: "{0}"',
Migrations.php
    'notFound'      => 'Migration file not found: ',
    'batchNotFound' => 'Target batch not found: ',
    'gap'           => 'There is a gap in the migration sequence near version number: ',
    'rollingBack'       => 'Rolling back migrations to batch: ',
    'removed'           => 'Rolling back: ',
    'added'             => 'Running: ',
    'on'        => 'Migrated On: ',
Publisher.php
    'destinationNotAllowed' => 'Destination is not on the allowed list of Publisher directories: "{0}"',
    'fileNotAllowed'        => '"{0}" fails the following restriction for "{1}": {2}',
Router.php
    'invalidDynamicController' => 'A dynamic controller is not allowed for security reasons. Route handler: "{0}"',
    'invalidControllerName'    => 'The namespace delimiter is a backslash (\), not a slash (/). Route handler: "{0}"',
Security.php
    'invalidSameSite' => 'The SameSite value must be None, Lax, Strict, or a blank string. Given: "{0}"',
Session.php
    'invalidSavePath'        => 'Session: Configured save path "{0}" is not a directory, does not exist or cannot be created.',
    'writeProtectedSavePath' => 'Session: Configured save path "{0}" is not writable by the PHP process.',
    'emptySavePath'          => 'Session: No save path configured.',
    'invalidSavePathFormat'  => 'Session: Invalid Redis save path format: "{0}"',
    'invalidSameSiteSetting' => 'Session: The SameSite setting must be None, Lax, Strict, or a blank string. Given: "{0}"',
Time.php
    'invalidMonth'   => 'Months must be between 1 and 12. Given: {0}',
    'invalidDay'     => 'Days must be between 1 and 31. Given: {0}',
    'invalidOverDay' => 'Days must be between 1 and {0}. Given: {1}',
    'invalidHours'   => 'Hours must be between 0 and 23. Given: {0}',
    'invalidMinutes' => 'Minutes must be between 0 and 59. Given: {0}',
    'invalidSeconds' => 'Seconds must be between 0 and 59. Given: {0}',
Validation.php
    'alpha_numeric_punct'   => 'The {field} field may contain only alphanumeric characters, spaces, and  ~ ! # $ % & * - _ + = | : . characters.',
    'equals'                => 'The {field} field must be exactly: {param}.',
    'in_list'               => 'The {field} field must be one of: {param}.',
    'not_equals'            => 'The {field} field cannot be: {param}.',
    'not_in_list'           => 'The {field} field must not be one of: {param}.',
View.php
    'invalidCellClass'      => 'Unable to locate view cell class: "{0}".',
    'tagSyntaxError'        => 'You have a syntax error in your Parser tags: "{0}"',
```

kenjis:
```
BaseConnection.php
            log_message('error', 'Error connecting to the database: ' . $e);
                        log_message('error', 'Error connecting to the database: ' . $e);
                        log_message('error', 'Database: Failure during an automated transaction commit/rollback!');
BasePreparedQuery.php
                        log_message('error', 'Database: Failure during an automated transaction commit/rollback!');
CacheFactory.php
            log_message('critical', $e . ' Resorting to using ' . $backup . ' handler.');
CLI.php
            log_message('error', (string) $e);
Common.php
    function log_message(string $level, string $message, array $context = [])
Connection.php
                    log_message('error', $message);
                    log_message('error', "Database: Unable to set the configured connection charset ('{$this->charset}').");
            log_message('error', (string) $e);
Connection.php
            log_message('error', (string) $e);
Connection.php
            log_message('error', (string) $e);
Connection.php
            log_message('error', $error['message']);
Connection.php
            log_message('error', (string) $e);
CookieStore.php
                log_message('error', (string) $e);
Email.php
            log_message('error', 'Email: ' . $method . ' throwed ' . $e);
            log_message('error', 'Email: ' . $message);
            log_message('error', $this->printDebuggerRaw());
                log_message('error', 'Email: ' . $method . ' throwed ' . $e);
Exceptions.php
            log_message('critical', "{message}\nin {exFile} on line {exLine}.\n{trace}", [
        log_message(
FileHandler.php
                log_message('debug', 'Failed to set mode on cache file: ' . $e);
GeneratorTrait.php
            log_message('error', (string) $e);
RedisHandler.php
                // Note:: I'm unsure if log_message() is necessary, however I'm not 100% comfortable removing it.
                log_message('error', 'Cache: Redis connection failed. Check your configuration.');
                log_message('error', 'Cache: Redis authentication failed.');
                log_message('error', 'Cache: Redis select database failed.');
RedisHandler.php
            log_message('error', 'Session: Unable to obtain lock for ' . $this->keyPrefix . $sessionID . ' after 30 attempts, aborting.');
            log_message('debug', 'Session: Lock for ' . $this->keyPrefix . $sessionID . ' had no TTL, overriding.');
ResponseTrait.php
            log_message('error', (string) $e);
Router.php
        log_message('info', 'Used the default controller.');
Security.php
        log_message('info', 'CSRF token verified.');
        log_message('info', 'CSRF cookie sent.');
Toolbar.php
                log_message(
```

kenjis: It seems we often use the pattern like `Invalid file: "{0}"`.

kenjis: Fixed. 64130a27b5bd6440f2dcfc30b44410f669651f51

kenjis (approved):

PR #7383, opened 2023-03-25
docs: update CSP description

kenjis:
**Description**
- use `getCSP()` instead of the deprecated `CSP` property

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


samsonasik:
Thank you @kenjis

PR #7397, opened 2023-03-30
docs: add link list of timezones php.net

datamweb:
**Description**

Add find the complete list of timezones supported by PHP.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -105,6 +105,8 @@ class App extends BaseConfig
      *
      * The default timezone that will be used in your application to display
      * dates with the date helper, and can be retrieved through app_timezone()
+     *
+     * @see https://www.php.net/manual/en/timezones.php for list of timezones supported by PHP.
      */
     public string $appTimezone = 'UTC';
 


app/Config/App.php
```diff
@@ -1,5 +1,14 @@
...
 
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
```
kenjis, line 11: Why do you add the license notice?

This file is a part of an app, not the framework.
There is no need to claim CodeIgniter Foundation copyright.

datamweb: The reason for adding the license notice is because of the run of command `cs-fix`.

If it is not going to be run on this file and other similar files of cs-fix, then it should be introduced in configuration file .php-cs-fixer, in section **exclude**.

Or at least I thought so.

paulbalandan: This should not be the case. `app/**/*.php` are linted by `.php-cs-fixer.no-header.php`. If the linter adds the header files, then the command run used the default `.php-cs-fixer.php` with file specified.

kenjis (approved):

samsonasik:
Thank you @datamweb

PR #7398, opened 2023-03-30
fix: CURLRequest - clear response headers between requests

### LINKED ISSUE #7394: Bug: [CURLRequest] not clear Response headers before next request
### PHP Version

8.1

### CodeIgniter4 Version

4.3.3

Manual (zip or tar.gz)

Linux

### Which server did you use?

fpm-fcgi

### Database

_No response_

### What happened?

When I send more than one requests by same instance of CURLRequest class, some headers of responses in front will exist in the later responses. For example:
At the first time,I requested 'http://domain.com/URI/A' successfully, then I can get some headers such as 'expires' or 'etag' by method 'getHeaderLine'.
At the second time, I requested 'http://domain.com/URI/B' by the same instance, it failed and the server doesn't response the header 'etag'. But when I call the 'hasHeader' method, I call still get it.

### Steps to Reproduce

Send more than one requests by same instance of CURLRequest class, and the server responses different header names.

### Expected Output

Headers of responses in front should be cleared before the later request.

### Anything else?

My way to solve this problem is just setting two properties of response instance (a property of CURLRequest instance) to empty array before parsing output of curl:
$break = strpos($output, $breakString);
$this->response->headers   = [];
$this->response->headerMap = [];
...
I‘m a green hand,maybe there are some better ways?

michalsn:
**Description**
Fixes #7394

- create a new Response object for a request

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -112,7 +112,7 @@ class CURLRequest extends OutgoingRequest
 
         parent::__construct('GET', $uri);
 
-        $this->responseOrig   = $response;
+        $this->responseOrig   = $response ?? new Response(config('App'));
         $this->baseURI        = $uri->useRawQueryString();
         $this->defaultOptions = $options;
 
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -29,6 +29,13 @@ class CURLRequest extends OutgoingRequest
     protected $response;
 
     /**
+     * The original response object associated with this request
+     *
+     * @var ResponseInterface|null
+     */
+    protected $responseOrig;
+
+    /**
      * The URI associated with this request
      *
      * @var URI
@@ -105,7 +112,7 @@ class CURLRequest extends OutgoingRequest
 
         parent::__construct('GET', $uri);
 
-        $this->response       = $response;
+        $this->responseOrig   = $response;
         $this->baseURI        = $uri->useRawQueryString();
         $this->defaultOptions = $options;
 
@@ -125,6 +132,8 @@ class CURLRequest extends OutgoingRequest
      */
     public function request($method, string $url, array $options = []): ResponseInterface
     {
+        $this->response = clone $this->responseOrig;
+
         $this->parseOptions($options);
 
         $url = $this->prepareURL($url);
@@ -469,10 +478,6 @@ class CURLRequest extends OutgoingRequest
      */
     protected function setResponseHeaders(array $headers = [])
     {
-        foreach ($this->response->headers() as $header) {
-            $this->response->removeHeader($header->getName());
-        }
-
         foreach ($headers as $header) {
             if (($pos = strpos($header, ':')) !== false) {
                 $title = substr($header, 0, $pos);
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -469,10 +469,8 @@ class CURLRequest extends OutgoingRequest
      */
     protected function setResponseHeaders(array $headers = [])
     {
-        if ($this->shareOptions === false) {
-            foreach ($this->response->headers() as $header) {
-                $this->response->removeHeader($header->getName());
-            }
+        foreach ($this->response->headers() as $header) {
+            $this->response->removeHeader($header->getName());
         }
 
         foreach ($headers as $header) {
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -469,6 +469,12 @@ class CURLRequest extends OutgoingRequest
      */
     protected function setResponseHeaders(array $headers = [])
     {
+        if ($this->shareOptions === false) {
+            foreach ($this->response->headers() as $header) {
+                $this->response->removeHeader($header->getName());
+            }
+        }
+
         foreach ($headers as $header) {
             if (($pos = strpos($header, ':')) !== false) {
                 $title = substr($header, 0, $pos);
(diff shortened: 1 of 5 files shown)


iRedds:
I believe that for each request, a new response instance should be created, and not passed as a dependency.

kenjis:
I can still understand sharing request options, but I don't believe there is a use case for sharing and overriding response objects.

michalsn:
> I believe that for each request, a new response instance should be created, and not passed as a dependency.

The idea was to allow the use of Response classes other than the default one.

> I can still understand sharing request options, but I don't believe there is a use case for sharing and overriding response objects.

Existing headers are now always removed before collecting the new response headers.

kenjis:
Tests were failed. Please rebase.

kenjis:
> I believe that for each request, a new response instance should be created, and not passed as a dependency.

I agree.

> The idea was to allow the use of Response classes other than the default one.

I don't understand why do we need to pass a Response object.

I think it is not correct to share the same Response object between requests.
Because they are the different responses.

This is not a bit clean, but I think better:

```diff
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -137,7 +137,12 @@ class CURLRequest extends OutgoingRequest
             $this->resetOptions();
         }
 
-        return $this->response;
+        $response = $this->response;
+
+        // Set new Response for the next request.
+        $this->response = new Response(config('App'));
+
+        return $response;
     }
 
     /**
```

kenjis:
If possible, I think this (and deprecate `$response` in the constructor) is better:

```diff
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -125,6 +125,8 @@ class CURLRequest extends OutgoingRequest
      */
     public function request($method, string $url, array $options = []): ResponseInterface
     {
+        $this->response = new Response(config('App'));
+
         $this->parseOptions($options);
 
         $url = $this->prepareURL($url);
```

michalsn:
Returning a new response class for every CURL request would mean we tightly couple the response with the CodeIgniter `Response` class.

For me, that would be a regression in extensibility. Of course, "resetting" headers would be a compromise, but there is something we get in return.

The question is, do we want to offer fewer options and introduce a BC break?

iRedds:
@michalsn I don't know the real situation, but it seems to me that 90% of CI developers will use the classes that are available.
The remaining 10% will not experience problems with class extensions.

But for example, I want to give the code. And I have not seen angry comments about the impossibility of replacing the class.

https://github.com/codeigniter4/CodeIgniter4/blob/354dc67f873429a543c8a1793910fcd883e275b1/system/HTTP/IncomingRequest.php#L904-L911
https://github.com/codeigniter4/CodeIgniter4/blob/354dc67f873429a543c8a1793910fcd883e275b1/system/HTTP/Files/FileCollection.php#L164-L187

kenjis:
We should always avoid BC breaks especially in patch versions.

Then how about something like this:
```diff
--- a/system/HTTP/CURLRequest.php
+++ b/system/HTTP/CURLRequest.php
@@ -125,6 +125,8 @@ class CURLRequest extends OutgoingRequest
      */
     public function request($method, string $url, array $options = []): ResponseInterface
     {
+        $this->response = clone $this->responseOrig;
+
         $this->parseOptions($options);
 
         $url = $this->prepareURL($url);
```

michalsn:
@iRedds I don't know if anybody is using other response classes than the default one, but ~5 years ago, I would take advantage of this approach in one project. 

The argument that framework already has places where things are tightly coupled doesn't speak to me. Just because we don't give everywhere a choice about what class we will use doesn't mean we should "level down" and take that choice away from the places where we have it now.

@kenjis I see the added value in replacing the default class in some cases. A long time ago I had a project where I would extend the Response class to simplify handling the response headers. I was consuming a really specific API.

I like your suggestion with `clone`, and this is a good option.

---

The bottom line is that I allow the thought that keeping the support for the custom Response class is just my fixation.

Since dropping the support would involve a BC break, I would love to hear from other members. @codeigniter4/core-team any thoughts on this one?

iRedds:
@michalsn I can offer two alternatives.
1. Instead of a Response object, pass a factory through an interface.
```php
public function __construct(protected CURLFactoryInterface $factory, protected array $options = []){}

new CURLRequest(new CURLFactory(), $options)

// Now we can get a new Response object.
$this->factory->response();
```

2. You can also simply add a method directly in the CURLRequest class, which will return the object we need.
```php
protected function buildResponse(): ResponseInterface
{
return new Response(config('App'));
}
```

kenjis:
@michalsn Okay, you had the real use cases. So removing injection of ResponseInterface is degrade and a breaking change. We should use `clone` now.

michalsn:
@iRedds These are good options. I especially like the second one, but both would involve BC changes. That's why I send a change with `clone` version.

@kenjis Yes, I agree. The new response class for every request will be much better.

Thank you both for the input. Much appreciate.

kenjis:
`$this->responseOrig` may be null.

```php
<?php

namespace App\Controllers;

use CodeIgniter\HTTP\CURLRequest;
use CodeIgniter\HTTP\URI;
use Config\App;

{
    public function index()
    {
        $client = new CURLRequest(new App(), new URI());
        $client->get('https://codeigniter.com/');
    }
}
```
```
Error
__clone method called on non-object

SYSTEMPATH/HTTP/CURLRequest.php at line 135
```

michalsn:
Good point. Fixed.

kenjis (approved):

PR #7404, opened 2023-04-03
Fixed filters exec order

c4user:
Exec order of the filters was rather irregular. This can cause performance problems. As defaults in config, the csrf filter is always run after the router filters. Let's say you are using a database connection in router filters and the data sent is not suitable for csrf but the filters are working. or you need to use errorhandler for php8 (flood warning), you can't use it in global filter. Because you cannot catch errors in router filters. 

before filters:
Previously: routers > global > method > filters
Now: global > method > filters > routers

after filters:
Previously: routers > global > filters
Now: routers > filters > global

### DIFF
--- a/system/Filters/Filters.php
+++ b/system/Filters/Filters.php
@@ -227,9 +227,9 @@ class Filters
             return $this;
         }
 
-        $this->processGlobals($uri);
-        $this->processMethods();
         $this->processFilters($uri);
+        $this->processMethods();
+        $this->processGlobals($uri);
 
         // Set the toolbar filter to the last position to be executed
         if (in_array('toolbar', $this->filters['after'], true)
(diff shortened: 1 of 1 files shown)


kenjis:
Thank you for trying to improve CI4.

But this PR has missing items.
1. GPG siginig
2. Test code

Please read https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md

And we don't accept breaking changes in `develop` as much as possible.
At least this PR should go to `4.4` branch.

Also please read the past discussion #6262

MGatner:
@c4user are you available to address kenjis' concerns above?

kenjis:
I sent a PR based on this. See #7955

kenjis:
Closed by #7955

PR #7162, opened 2023-01-22
feat: [Auto Routing Improved] fallback to default method

kenjis:
**Description**
If there is no method in the controller for the request URI, fallback to the default method.

E.g.:
`GET /product/15`
→ `Product::get15()` Method not found
→ `Product::getIndex($id = '')` Method found and run `getIndex('15')`

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #7406, opened 2023-04-04
feat: [Auto Routing Improved] fallback to default controller's default method

kenjis:
**Description**
Related #7162

You can pass arguments to the default method of the default controller.

When you have `News\Home` and `public function getIndex($id = null)` in it:
  - `GET /news/101`
    - → `News` not found
    - → `News\101` not found (invalid)
    - → `News\101\Home` not found (invalid)
    - → `News\Home` found
    - → `News\Home::getIndex()` found
    - → run `News\Home::getIndex('101')`

This PR makes it possible that one controller one URI.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Utilities/Routes/AutoRouterImproved/ControllerMethodReader.php
+++ b/system/Commands/Utilities/Routes/AutoRouterImproved/ControllerMethodReader.php
@@ -85,23 +85,7 @@ final class ControllerMethodReader
                             continue;
                         }
 
-                        $params      = [];
-                        $routeParams = '';
-                        $refParams   = $method->getParameters();
-
-                        foreach ($refParams as $param) {
-                            $required = true;
-                            if ($param->isOptional()) {
-                                $required = false;
-
-                                $routeParams .= '[/..]';
-                            } else {
-                                $routeParams .= '/..';
-                            }
-
-                            // [variable_name => required?]
-                            $params[$param->getName()] = $required;
-                        }
+                        [$params, $routeParams] = $this->getParameters($method);
 
                         // Route for the default method.
                         $output[] = [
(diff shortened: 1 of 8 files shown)


kenjis:
I have a feeling that 1 (can use the default controller's method) is not necessary.

kenjis:
I would like to merge this. Can someone review?

iRedds:
I think this autorouting is too auto.
For example, the ``News::getIndex()`` method returns a list of news.
If I didn't specify the ``/news/(:num)`` route, then I would expect a request to the URL ``/news/15`` to return an error.
Now I have to check the argument in the ``getIndex()`` method in order to get a 404 error for the URL ``/news/15``.

It seems to me that such a change will lead to incorrect operation of existing applications.

If I compared on a scale of breaking changes, then I would rate this PR as 10 out of 10, and for example, changing the interface in the core of the framework 6 out of 10.

kenjis:
If the `getIndex()` method has no parameter, `/news/15` is not routed to the controller.
`/news` is routed to. Because Auto Routing Improved checks method parameter count.
If there are more parameters in the URI than the method parameters, it results in 404.

There is no way to specify the type of a parameter like `/news/(:num)` in Auto Routing.

I do not think this PR (or #7162) to be a breaking change. 
Give use cases where existing apps would be broken. 
I will consider it.

iRedds:
I haven't looked at the Autorouter code because I think it's a waste of core team resources. Therefore, my comment was based only on the concept proposed in the PR. But after your answer, I saw that the code uses ReflectionClass.
While I still think the framework's support for autorouting is a waste of resources, I don't have any more questions for this PR.

kenjis:
Defining all routes is tedious. 
Many sites do not need flexible routing that defined routes provide.
Even if needed, most routes can be handled by auto routing.
So I don't want to remove auto routing feature.

PR #7414, opened 2023-04-07
docs: improve auto routing

kenjis:
**Description**
- improve description for Default Controller in Auto-Routing
- update for consistency

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/incoming/controllers.rst
+++ b/user_guide_src/source/incoming/controllers.rst
@@ -274,7 +274,7 @@ Your method will be passed URI segments 3 and 4 (``'sandals'`` and ``'123'``):
 
 .. literalinclude:: controllers/022.php
 
-.. note:: If there are more parameters in the URI than the method parameters,
+.. important:: If there are more parameters in the URI than the method parameters,
     Auto Routing (Improved) does not execute the method, and it results in 404
     Not Found.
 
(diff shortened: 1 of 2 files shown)


user_guide_src/source/incoming/controllers.rst
```diff
@@ -274,16 +274,21 @@ Your method will be passed URI segments 3 and 4 (``'sandals'`` and ``'123'``):
...
 
-.. note:: If there are more parameters in the URI than the method parameters,
+.. important:: If there are more parameters in the URI than the method parameters,
     Auto Routing (Improved) does not execute the method, and it results in 404
     Not Found.
 
+Default Controller
+==================
+
+The Default Controller is a special controller that is used when a URI end with
```
michalsn, line 284:
```suggestion
The Default Controller is a special controller that is used when a URI ends with
```

michalsn (commented):
This question is not very important, but why are we changing `folder` to `directory` everywhere? Do we stick to this convention on other pages? I have nothing against it, though.

kenjis:
This is an explanation of subdirectories in Controllers, and there is no need to use the two terms *folder* and *directory*. Suddenly changing what we call a directory to a folder may confuse beginners.

I do not intend to change all *folders* in the user guide to *directories*.

michalsn (approved):

kenjis:
Thank you! @michalsn

PR #7420, opened 2023-04-11
feat: add method to get the validated data

kenjis:
**Description**
- add `Validation::getValidated()`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Validation/DotArrayFilter.php
+++ b/system/Validation/DotArrayFilter.php
@@ -1,0 +1,109 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Validation;
+
+final class DotArrayFilter
+{
+    /**
+     * Creates a new array with only the elements specified in dot array syntax.
+     *
+     * This code comes from the dot_array_search() function.
+     *
+     * @param array $indexes The dot array syntax pattern to use for filtering.
+     * @param array $array   The array to filter.
+     *
+     * @return array The filtered array.
+     */
+    public static function run(array $indexes, array $array): array
+    {
+        $result = [];
+
+        foreach ($indexes as $index) {
+            // See https://regex101.com/r/44Ipql/1
+            $segments = preg_split(
+                '/(?<!\\\\)\./',
+                rtrim($index, '* '),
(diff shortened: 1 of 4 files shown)


iRedds:
I'd like to hear the reason for adding this method.
What is the real benefit of it?

kenjis:
Security enhancement. 
This method returns only validated data, so developers can be sure to use validated data.

iRedds:
Let's say the script receives data, some of which do not need to be validated.
It turns out that after validation, the getValidated() method will return only part of the data, which will add an extra operation to manipulate the data further.
For me, this feature looks dubious.

kenjis:
> Let's say the script receives data, some of which do not need to be validated.

This method is intended for input validation like validation for HTML form.
What data don't you need to validate?

kenjis:
Added the docs.

user_guide_src/source/libraries/validation/043.php
```diff
@@ -0,0 +1,21 @@
...
+]);
+
+$data = [
+    'username'   => 'john',
+    'password'   => 'BPi-$Swu7U5lm$dX',
+    'csrf_token' => '8b9218a55906f9dcc1dc263dce7f005a',
+];
+
+if ($validation->run($data)) {
+    $validatedData = $validation->getValidated();
```
michalsn, line 16: This example is fine, but I would like to ask you to add an example that will cover controller validation (with the [$this->validate()](https://github.com/codeigniter4/CodeIgniter4/blob/develop/user_guide_src/source/incoming/controllers.rst#this-validate) method). This is the place where this method is the most valuable.

kenjis: I have a plan to add a method to get validated data in `Controller`, and update all sample code with `$this->validate()`.

michalsn (approved):

user_guide_src/source/libraries/validation/043.php
michalsn: I don't think need a separate method in the controller for this, though. 
Since we can get the errors via `$this->validator->getErrors()`, I see no problem with getting validated data the same way, via `$this->validator->getValidated()`.

ddevsr (approved):

user_guide_src/source/libraries/validation/043.php
kenjis: Oh, I see. Thank you for opinion.

kenjis: See #7425

PR #7422, opened 2023-04-12
fix: [Auto Routing Improved] one controller method has more than one URI when $translateURIDashes is true

kenjis:
~~Needs #7406~~

**Description**
When `$translateURIDashes` is true, two URIs correspond to a single controller method,
one URI for dashes (`foo-bar`) and one URI for underscores (`foo_bar`). 
This is incorrect behavior, contrary to the design philosophy.

This PR prohibits the URI for underscores (`foo_bar`) when `$translateURIDashes` is true.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/AutoRouterImproved.php
+++ b/system/Router/AutoRouterImproved.php
@@ -71,6 +71,11 @@ final class AutoRouterImproved implements AutoRouterInterface
     private string $defaultMethod;
 
     /**
+     * The URI segments.
+     */
+    private array $segments = [];
+
+    /**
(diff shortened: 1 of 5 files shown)


kenjis:
Rebased to resolve conflicts.

MGatner (approved):
Very nice testing solution. Really glad you made this class `final` to begin with! I'd like to make most service classes `final` in v5

PR #7432, opened 2023-04-13
docs: replace type `mixed` at Config class.

ping-yee:
**Description**
See #6310 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Config/BaseService.php
+++ b/system/Config/BaseService.php
@@ -173,9 +173,9 @@ class BaseService
      *
      * $key must be a name matching a service.
      *
-     * @param mixed ...$params
+     * @param array|object|string|null ...$params
      *
-     * @return mixed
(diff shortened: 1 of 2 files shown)


system/Config/BaseService.php
```diff
@@ -173,9 +173,9 @@ class BaseService
      *
      * $key must be a name matching a service.
      *
-     * @param mixed ...$params
+     * @param array|object|string|null ...$params
```
kenjis, line 176: These are parameters for services. Maybe all except for resource?
Or resource might be passed by devs?

ping-yee: I have been added!

kenjis (approved):
Thank you!

PR #7435, opened 2023-04-14
docs: replace `mixed` type at CLI class.

ping-yee:
**Description**
See #6310 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CLI/CLI.php
+++ b/system/CLI/CLI.php
@@ -915,7 +915,7 @@ class CLI
      *
      * **IMPORTANT:** The index here is one-based instead of zero-based.
      *
-     * @return array|string|null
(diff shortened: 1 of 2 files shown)


system/CLI/CLI.php
```diff
@@ -915,7 +915,7 @@ public static function getURI(): string
      *
      * **IMPORTANT:** The index here is one-based instead of zero-based.
      *
-     * @return mixed
+     * @return array|string|null
```
kenjis, line 918: Can be an array?

```suggestion
     * @return string|null
```

kenjis (approved):

PR #7436, opened 2023-04-14
fix: PHPDoc @template and @return in database

kenjis:
**Description**
To fix this error: https://github.com/codeigniter4/shield/pull/195#issuecomment-1508037593

It seems Psalm does not allow to use `resource` in @template type.
See
- https://github.com/codeigniter4/settings/pull/60#issuecomment-1414894845
- https://github.com/codeigniter4/shield/pull/195#issuecomment-1508037593
    
**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Database/ResultInterface.php
+++ b/system/Database/ResultInterface.php
@@ -13,6 +13,10 @@ namespace CodeIgniter\Database;
 
 use stdClass;
 
+/**
+ * @template TConnection
+ * @template TResult
+ */
 interface ResultInterface
 {
     /**
--- a/system/Database/ResultInterface.php
+++ b/system/Database/ResultInterface.php
@@ -13,10 +13,6 @@ namespace CodeIgniter\Database;
 
 use stdClass;
 
-/**
- * @template TConnection of object|resource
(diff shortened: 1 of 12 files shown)


MGatner (approved):

system/Database/ResultInterface.php
```diff
@@ -13,10 +13,6 @@
 
 use stdClass;
 
-/**
- * @template TConnection of object|resource
- * @template TResult of object|resource
- */
```
kenjis, line 19: I removed `TConnection` and `TResult` because they are not used in this interface.
@paulbalandan Are you okay?

paulbalandan: I think I added those template types to help phpstan think that the interface is a generic and thus implementing classes should specify the specific types.

kenjis: Restored `@template` without types, because of https://github.com/vimeo/psalm/issues/2625

paulbalandan (approved):

MGatner (approved):

PR #7445, opened 2023-04-20
fix: redirect status code

kenjis:
**Description**
With this PR, if you specify status code for redirect, the code is surely used in the response.
`$routes->addRedirect()` specifies 302 by default. So 302 will be used by default.

- fix: specified code may not be used in redirect (Should use the specified code)
- fix: default value to `null` for `$code` in `RedirectResponse::route()` (For consistency)
- fix: HEAD/OPTION request changes to GET because of 303
- change the default code to 302 for GET request (No need to use 307)

Ref #3057

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


iRedds:
1. The default redirect code should be 302.
2. There should be no code substitution inside the handler.

301 and 302 for a redirect without data (GET -> GET).
308 and 307 for data redirect (POST -> POST).
303 for redirect after request processing (POST -> GET)

kenjis:
> There should be no code substitution inside the handler.

Why?
It is done since v3.
https://github.com/bcit-ci/CodeIgniter/blob/63d037565dd782795021dcbd3b1ca6f8c8a509e4/system/helpers/url_helper.php#L522

kenjis:
> I hope you fix it, because the POST -> POST redirect in the current implementation will replace 307 or 308 with 303.

I think it will be fixed by this PR.
See https://github.com/codeigniter4/CodeIgniter4/pull/7445/commits/4da5f0d8761a37d74699a2a39a3bae57fc21b922#diff-e3e8e9e1b70160a0f569050d2f248d3df024213a3bd66909a3d0308024b60638

system/HTTP/ResponseTrait.php
```diff
@@ -503,28 +503,27 @@ public function sendBody()
...
             $method = 'refresh';
+        } elseif ($method !== 'refresh' && $code === null) {
+            // override status code for HTTP/1.1 & higher
+            if (
+                isset($_SERVER['SERVER_PROTOCOL'], $_SERVER['REQUEST_METHOD'])
+                && $this->getProtocolVersion() >= 1.1
+            ) {
+                $code = ($_SERVER['REQUEST_METHOD'] !== 'GET')
+                    ? 303 // reference: https://en.wikipedia.org/wiki/Post/Redirect/Get
+                    : 307;
```
iRedds, line 521: For the HEAD and OPTIONS methods, there will be a 303 code, that is, a redirect with the GET method.

iRedds:
> Why? It is done since v3. https://github.com/bcit-ci/CodeIgniter/blob/63d037565dd782795021dcbd3b1ca6f8c8a509e4/system/helpers/url_helper.php#L522

There is a lot written in CI 3. But that doesn't mean it's right.
Don't get too attached to things. Learn to let go(с) Ivan Vanko.

kenjis:
> There is a lot written in CI 3. But that doesn't mean it's right.

Right or wrong is just a direction. If you want to go to the North Pole, it is right to go north.
It is not right to go south.

Yes, if there is a code in CI3, it does not means the code is right code.
But if we change the code, apps might break.
So more careful consideration is needed to make the change, and
we need to show the reason why.

I just want you to show why your opinion is right.

iRedds:
I will repeat what I said above.
1. The 303 response code implies that a POST request was sent, processed, and the processing result can be obtained at the specified URL via a GET request.
2. Response code 307 implies that the resource was found but temporarily moved to the specified URL, so you need to use the same method with which the request was sent and resend the request body.

Regarding code CI3.
The code assumes that if the response code is not passed as an argument, then any non-GET request is a POST request (given the link to the wiki). That is, it is processed and the result can be obtained from the specified link through the GET method.

The HEAD and OPTIONS methods are not GET, which means that they will have a 303 response code, but this makes no sense when redirecting using the GET method.

For a GET request, it makes no sense to use 307 (308) response codes, since a GET request does not imply a request body. It makes more sense to use 302 (301).

Conclusion.
We can set the default value of the code argument to 303. And remove the condition for defining the response code.

For non-GET methods, the behavior will not change. Since they are already redirected to 303.
For the GET method, the behavior will not change. Since the body of the GET request is not provided.

system/HTTP/ResponseTrait.php
kenjis: Fixed.

michalsn:
Maybe we could add https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections as a reference for `redirect()` method? I think this could be useful regarding the rules we're following.

IMO this is a good change.

kenjis:
Added docs.

kenjis:
Rebased to resolve conflicts, and tweaks the added note.

michalsn (approved):

kenjis:
@michalsn Thank you.

PR #7448, opened 2023-04-22
fix: button's type should not default to button

grimpirate:
An html button's default type is submit. The form helper form_button creates a button which exclusively allows a "button" type. However, there are certain scenarios where the value of the submit button should be different from its content. The current option is to use an form_submit with a form_hidden to reproduce the same result.

### DIFF
--- a/system/Helpers/form_helper.php
+++ b/system/Helpers/form_helper.php
@@ -420,11 +420,11 @@ if (! function_exists('form_button')) {
      * @param array|string        $data
      * @param array|object|string $extra string, array, object that can be cast to array
      */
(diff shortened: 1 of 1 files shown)


michalsn:
We already can change the type.
```php
echo form_button(['name' => 'name', 'type' => 'submit'], 'content', ['class' => 'btn'])
```

grimpirate:
Hmm... thank you for pointing that out michalsn.

kenjis:
> An html button's default type is submit. 

It is true, but if we change it, it is a breaking change.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#backwards-compatibility

PR #7454, opened 2023-04-26
docs: improve description on Env Vars and Config values

### LINKED ISSUE #7453: Bug: MySQLi Config "encrypt" values not being respected
### PHP Version

8.2

### CodeIgniter4 Version

4.3.2

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

While following the [instructions for configuring a database connection](https://codeigniter.com/user_guide/database/configuration.html#configuring-with-env-file), I have configured my env file to contain the following:

```
database.default.DBDriver = MySQLi
database.default.encrypt.ssl_ca = C:\xampp\perl\vendor\lib\Mozilla\CA\cacert.pem
database.default.encrypt.ssl_verify = true
```

However, it appears these are not being respected by the actual SSL connection, as seen in [Database.php](https://github.com/codeigniter4/CodeIgniter4/blob/develop/app/Config/Database.php#L27). Thus, checking in `vendor/codeigniter4/framework/system/Database/MySQLi/Connection.php`. it appears that the entire section for checking if `is_array($this->encrypt)` is skipped over due to not being an array.

I have "fixed" this by adding a simple line beforehand that manually sets the encrypt values:

```php
$this->encrypt = [
    'ssl_key'    => '',
    'ssl_cert'   => '',
    'ssl_ca'     => 'C:\xampp\perl\vendor\lib\Mozilla\CA\cacert.pem',
    'ssl_capath' => '',
    'ssl_cipher' => '',
    'ssl_verify' => 'true'
];
```

Thus continuing the checks and ensuring I can make an encrypted connection.

### Steps to Reproduce

1. Set any of the encryption subvalues in your .env file, or in Database.php
2. Go to vendor/codeigniter4/framework/system/Database/MySQLi/Connection.php
3. Within the connection function, add a simple `print_r($this->encrypt);` and observe that it appears to be being interpreted as a boolean rather than a array.

### Expected Output

The correct values to be passed to Database.php / Connection.php to allow for connecting to databases that enforce SSL connections.

### Anything else?

_No response_

kenjis:
**Description**
Closes #7453
- improve description on Env Vars and Config values
- add example to set database.default.encrypt with `.env`

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide

### DIFF
--- a/user_guide_src/source/database/configuration.rst
+++ b/user_guide_src/source/database/configuration.rst
@@ -104,6 +104,9 @@ default group's configuration settings. The values should be name following this
     database.default.password = '';
     database.default.database = 'ci4';
 
+.. important:: You cannot add a new property by setting environment variables,
+    nor change a scalar value to an array. See :ref:`env-var-replacements-for-data`.
+
 **********************
 Explanation of Values:
 **********************
--- a/user_guide_src/source/database/configuration.rst
+++ b/user_guide_src/source/database/configuration.rst
@@ -104,8 +104,19 @@ default group's configuration settings. The values should be name following this
     database.default.password = '';
     database.default.database = 'ci4';
 
-.. important:: You cannot add a new property by setting environment variables,
-    nor change a scalar value to an array. See :ref:`env-var-replacements-for-data`.
+But you cannot add a new property by setting environment variables, nor change a
+scalar value to an array. See :ref:`env-var-replacements-for-data` for details.
+
+So if you want to use SSL with MySQL, you need a hack. For example, set the array
+values as a JSON string in your **.env** file:
+
+::
+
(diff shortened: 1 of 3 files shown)


kenjis:
I have never had a case where the array value needed to be set in an environment variable, but it seems to me that the only way to set SSL settings in MySQL is to set the array in an environment variable.

Therefore, I have added an example of how to set it.

datamweb (commented):
Interesting description! This explanation will be useful for everyone.

Shouldn't these descriptions be added in the Environment Variables section or related `env` sections?

MGatner (approved):
Not thrilled about having a hack in our user guide but it is actually a handy tip.

PR #7392, opened 2023-03-30
fix: view cell cannot locate the auto-generated view file

sammyskills:
Fixes #7391 

- fix a bug that View Cells cannot load a view file with `_cell` when the classname ends with `Cell`.
- fix a bug that `spark make:cell` creates a view file with `_cell` when the classname does not end with `Cell`.

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #7479, opened 2023-05-03
fix: forbid single words as cell name input in `make:cell`

### LINKED ISSUE #7472: Bug: case-insensitive treatment of single-word view cells in Windows and macOS
### PHP Version

8.2

### CodeIgniter4 Version

develop

Git

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

This test (CellGeneratorTest) currently fails on Windows:

```php
    public function testGenerateCellSimpleName()
    {
        command('make:cell Another');

        // Check the class was generated
        $file = APPPATH . 'Cells/Another.php';
        $this->assertStringContainsString('File created: ' . clean_path($file), $this->getStreamFilterBuffer());
        $this->assertFileExists($file);
        $this->assertStringContainsString('class Another extends Cell', $this->getFileContents($file));

        // Check the view was generated
        $file = APPPATH . 'Cells/another.php';
        $this->assertStringContainsString('File created: ' . clean_path($file), $this->getStreamFilterBuffer());
        $this->assertFileExists($file);
        $this->assertSame("<div>\n    <!-- Your HTML here -->\n</div>\n", $this->getFileContents($file));
    }
```

This is because on Windows, `Cells/Another.php` is treated the same as `Cells/another.php`. Thus, only one file is generated, the class file. The view file is skipped.

### Steps to Reproduce

Run the test above in Windows. Linux and MacOS are not affected.

### Expected Output

I think we should add a check if the cell name is a single word and it is run through Windows. If that's the case, we forbid single word as cell name through a nice error message.

### Anything else?

_No response_

paulbalandan:
**Description**
Fixes #7472 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Commands/Generators/CellGenerator.php
+++ b/system/Commands/Generators/CellGenerator.php
@@ -12,6 +12,7 @@
 namespace CodeIgniter\Commands\Generators;
 
 use CodeIgniter\CLI\BaseCommand;
+use CodeIgniter\CLI\CLI;
 use CodeIgniter\CLI\GeneratorTrait;
 
 /**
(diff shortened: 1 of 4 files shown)


kenjis (approved):

lonnieezell:
This seems like the wrong fix, honestly. This is why originally the intent was to have either the class with `Cell` appended, or the view with `_cell` appended, so they were different. 

I know we had a cleanup recently that I didn't get a chance to look into, but if it removed the appending from either one then that should be fixed so that one or the other gets something appended.

kenjis:
The recent fix is #7392

@lonnieezell Do you mean the original intent is like this?

- Blog → blog_cell.php
- BlogCell → blog.php

It seems to be confusing.

The current naming rule is already complicated and I personally would like to see them simplified.
For example,
- Blog → Blog.html.php
- BlogCell → BlogCell.html.php

or
- Blog → blog.html.php
- BlogCell → blog_cell.html.php


or
- Blog → blog_cell.php
- BlogCell → blog_cell_cell.php

lonnieezell:
The original intent was, I believe: 

Controller: BlogCell.php
View: blog.php

It's nice and simple, matches what we have a tendency to do elsewhere (view is lowercased, class has function attached, like UserModel or UserController).

kenjis:
If we follow the following rule (it is valid in the current implementation),
the classname must be two or more words.

Controller: BlogCell.php
View: blog.php

@lonnieezell 
Do you want to append `Cell` automatically when we run `make:cell Blog`?

paulbalandan:
The appending of the `Cell` in the controller should be controlled by the `--suffix` option.

lonnieezell:
> Do you want to append Cell automatically when we run make:cell Blog?

That's the way I would handle it, yes. 

> The appending of the Cell in the controller should be controlled by the --suffix option.

While I understand what you're saying I think it's a better experience to automatically add it then to throw and error and make them try again.

paulbalandan:
@lonnieezell If I understand correctly, the specification would be:

```console
$ php spark make:cell BlogCell

File created: APPPATH . 'Cells/BlogCell.php'

File created: APPPATH . 'Cells/blog.php'

```

```console
$ php spark make:cell Another

File created: APPPATH . 'Cells/AnotherCell.php'

File created: APPPATH . 'Cells/another.php'

```

lonnieezell:
@paulbalandan Correct.

PR #7471, opened 2023-05-02
fix: make:cell using class as view file

paulbalandan:
**Description**
Fixes #7469 

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [x] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #7481, opened 2023-05-05
fix: standardize behavior of `make:cell` and `Cells`

### LINKED ISSUE #7469: Bug: spark make:cell doesn't create view file properly
deleted because of a mistake, sorry, I recap in the following reply

### LINKED ISSUE #7472: Bug: case-insensitive treatment of single-word view cells in Windows and macOS

paulbalandan:
**Description**
Fixes #7472 
Fixes #7469 

Supersedes and closes #7479 
Supersedes and closes #7471 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/View/Cells/Cell.php
+++ b/system/View/Cells/Cell.php
@@ -13,6 +13,7 @@ namespace CodeIgniter\View\Cells;
 
 use CodeIgniter\Traits\PropertiesTrait;
 use ReflectionClass;
+use RuntimeException;
 
 /**
  * Class Cell
@@ -64,6 +65,8 @@ class Cell
      * from within the view, this method extracts $data into the
      * current scope and captures the output buffer instead of
      * relying on the view service.
+     *
+     * @throws RuntimeException
      */
     final protected function view(?string $view, array $data = []): string
     {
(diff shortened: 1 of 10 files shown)


michalsn (approved):

kenjis:
This behavior does not match the current specification in the docs.
> So, for example, if you have a MyCell class, the view file should be my_cell.php.
https://codeigniter4.github.io/userguide/outgoing/view_cells.html#controlled-cells

In my understanding, the current specification is:
1. class: `MyCell` view: `my_cell.php` (default)
2. class: `MyCell` view: `my.php` (fallback; undocumented)

There is inconsistency between `make:cell` and the docs.
It will bring confusion or bug reports.

paulbalandan:
Based on @lonnieezell 's comment, the specification should be:
1. class: `MyCell`, view: `my.php` (default)
2. class: `MyCell`, view: `my_cell.php` (fallback for BC)

That part of the docs should be updated, too.

kenjis:
If we change the specification, add the following in this PR:
1. update the View Cells page's descriptions and sample code
2. add description that we accept "class: `MyCell` view: `my_cell.php`"

Also, it would be nice to have an option to generate `_cell.php` views, but this can be added later.

user_guide_src/source/outgoing/view_cells.rst
```diff
@@ -80,7 +80,15 @@ Controlled Cells
...
+Controlled cells have two primary goals: to make it as fast as possible to build the cell, and provide additional logic and
+flexibility to your views, if they need it. The class must extend ``CodeIgniter\View\Cells\Cell``. They should have a view file
+in the same folder. By convention, the class name should be in PascalCase suffixed with ``Cell`` and the view should be
+the snake_cased version of the class name, without the suffix. For example, if you have a ``MyCell`` class, the view file
+should be ``my.php``.
+
+.. note:: Prior to v4.3.5, the generated view file ends with ``_cell.php``. This was deemed incorrect and corrected accordingly.
+    For backward compatibility, existing view files ending with ``_cell.php`` will still be located and loaded. However, it is
+    recommended to update the filenames to remove the trailing ``_cell``.
```
kenjis, line 91: Is it needed to recommend to remove `_cell` in view files?
Do you have a plan to remove the logic to search `_cell` views in the future?

I think the current convention (`MyCell` → `my_cell.php`) should be also supported as ever.

paulbalandan: Let's have @lonnieezell have a say on this.

lonnieezell: Remember, people can also specify the view file to use within the cell class itself, so I don't think it's that big of a concern. 

Unless we have a plan to drop that support in the future, I don't know that there's much need to recommend them renaming the files. The performance gain would be unnoticeable.

lonnieezell (approved):
With the exception of th

user_guide_src/source/outgoing/view_cells.rst
kenjis: I hope CI4 supports both conventions, and does not recommend renaming the view files.

paulbalandan: I removed the recommendation. Please check if wording is now okay.

system/View/Cells/Cell.php
```diff
@@ -64,37 +65,48 @@ public function setView(string $view)
...
+            }
         }
 
-        // Locate our view, preferring the directory of the class.
-        if (! is_file($view)) {
-            // Get the local pathname of the Cell
-            $ref  = new ReflectionClass($this);
-            $view = dirname($ref->getFileName()) . DIRECTORY_SEPARATOR . $view . '.php';
+        if (! isset($foundView)) {
+            throw new RuntimeException('Cannot locate the view file for the cell.');
```
kenjis, line 101: It would be better to show the Cell classname.

user_guide_src/source/outgoing/view_cells.rst
kenjis: @paulbalandan I'm okay. Thank you.

system/View/Cells/Cell.php
kenjis, line 101: Is this a runtime exception?
It seems to me that it is a logic exception.

> LogicException
Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code. 
https://www.php.net/manual/en/class.logicexception.php

kenjis (approved):

kenjis:
@paulbalandan Thank you!

PR #7485, opened 2023-05-06
docs: fix syntax error in simple cells

datamweb:
**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/user_guide_src/source/outgoing/view_cells.rst
+++ b/user_guide_src/source/outgoing/view_cells.rst
@@ -46,7 +46,7 @@ Simple Cells are classes that return a string from the chosen method. An example
     {
         public function show(array $params): string
         {
-            return "<div class="alert alert-{$params['type']}">{$params['message']}</div>";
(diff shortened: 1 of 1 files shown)


user_guide_src/source/outgoing/view_cells.rst
```diff
@@ -46,7 +46,7 @@ Simple Cells are classes that return a string from the chosen method. An example
     {
         public function show(array $params): string
         {
-            return "<div class="alert alert-{$params['type']}">{$params['message']}</div>";
+            return "<div class='alert alert-{$params['type']}'>{$params['message']}</div>";
```
samsonasik, line 49: How about keep using double quote, escaped?

michalsn: That would work too.

kenjis: The use of single quotes is a bad practice and should not be used in documents.

PR #7487, opened 2023-05-08
fix: directory separator from routing file.

### LINKED ISSUE #7474: Bug: [4.4] Failed tests on Windows.
### PHP Version

7.4

### CodeIgniter4 Version

4.4

Git

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

After PR https://github.com/codeigniter4/CodeIgniter4/pull/7380 in Windows CodeIgniter\CodeIgniterTest::testRun404Override() and similar tests fail.

Instead of the expected response from the 404 error handler, a welcome_page is returned (from App\Controller\Home::index)

```
$ d:\1\CodeIgniter4\vendor\bin\phpunit.bat tests\system\CodeIgniterTest.php
PHPUnit 9.6.7 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.1
Configuration: D:\1\CodeIgniter4\phpunit.xml.dist
Warning:       No code coverage driver available

...FFFF............................                                                      35 / 35 (100%)

Time: 00:04.175, Memory: 28.00 MB

There were 4 failures:

1) CodeIgniter\CodeIgniterTest::testRun404Override
Failed asserting that '<!-- DEBUG-VIEW START 1 APPPATH\Views\welcome_message.php -->\r\n
[cut]
D:\1\CodeIgniter4\tests\system\CodeIgniterTest.php:124

2) CodeIgniter\CodeIgniterTest::testRun404OverrideControllerReturnsResponse
Failed asserting that '<!-- DEBUG-VIEW START 1 APPPATH\Views\welcome_message.php -->\r\n
[cut]
...etc
```
This is happening in this code snippet due to the difference in paths (directory separators) between Windows and UNIX.
https://github.com/codeigniter4/CodeIgniter4/blob/d9fd9d5e9dc34bcaa249c0df42c5d371ca45e582/system/Router/RouteCollection.php#L323-L332

```
$files
array(2) {
  [0]=>
  string(59) "D:\1\CodeIgniter4\app\Config\Routes.php"
  [1]=>
  string(70) "D:\1\CodeIgniter4\tests\_support\Config\Routes.php"
}

$this->routeFiles
array(1) {
  [0]=>
  string(59) "D:\1\CodeIgniter4\app\Config/Routes.php"  <--- DS
}
```
Due to the difference in paths, the comparison fails and the file with routes from the application is loaded, thereby adding the route that is accessed in the tests.

### Steps to Reproduce

1

### Expected Output

Since the framework can be used not only on UNIX, the tests must also pass on Windows. This issue is just a special case and fixing by changing the requested resource in the tests will not solve the problem. That is, the behavior should be like in UNIX, that is, the route from the application should not be loaded.

### Anything else?

In other tests related to file/directory paths, there are also errors on Windows due to the directory separator.

ping-yee:
**Description**
Fixes #7474 

Normalize the route path.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Router/RouteCollection.php
+++ b/system/Router/RouteCollection.php
@@ -264,6 +264,11 @@ class RouteCollection implements RouteCollectionInterface
         $this->autoRoute          = $routing->autoRoute;
         $this->routeFiles         = $routing->routeFiles;
         $this->prioritize         = $routing->prioritize;
+
+        // Normalize the path string in routeFiles array.
+        foreach ($this->routeFiles as $routeKey => $routesFile) {
+            $this->routeFiles[$routeKey] = realpath($routesFile) ?: $routesFile;
+        }
     }
 
     /**


app/Config/Routing.php
```diff
@@ -26,7 +26,7 @@ class Routing extends BaseRouting
      * Default: APPPATH . 'Config/Routes.php'
      */
     public array $routeFiles = [
-        APPPATH . 'Config/Routes.php',
+        APPPATH . 'Config' . DIRECTORY_SEPARATOR . 'Routes.php',
```
kenjis, line 29: If we use this way, we would need to use `DIRECTORY_SEPARATOR` in all file path strings.
To be honest, I don't want to write such code.

ping-yee: @kenjis I think the best way is handling normalize the `routeFile` array property in constructor.

lonnieezell:
Have you checked this on Windows and verified that `realpath()` will actually solve the problem?

ping-yee:
> Have you checked this on Windows and verified that realpath() will actually solve the problem?

Yep, I can pass the test cases on Windows (Linux also) with using `realpath()` to normalize the route path.

ping-yee:
In Windows, the route path will be:
```
D:\project\github\CodeIgniter4\app\Config/Routes.php <= before handle normalize work with realpath()

D:\project\github\CodeIgniter4\app\Config\Routes.php <= after handled with realpath()
```

In Linux, the route path will not be changed:
```
/mnt/d/project/github/CodeIgniter4/app/Config/Routes.php
```

@lonnieezell

kenjis (approved):
Thank you!

PR #7488, opened 2023-05-09
fix: PostgreSQL getVersion() logic

### LINKED ISSUE #7484: Bug: PosgreSQL server version
### PHP Version

8.0

### CodeIgniter4 Version

v4.3.4

Linux

### Which server did you use?

cli

### Database

PostgreSQL 13.10

### What happened?

`$db->getVersion();`  returns `false`

### Steps to Reproduce

```
$db = db_connect();
echo $db->getVersion();
```

### Expected Output

13.10 (Debian 13.10-0+deb11u1)

### Anything else?

I can make a pull request to fix it. Please specify if you want to store and return entire `$pgVersion['server']`  or just version accorsing to [version semantic](https://semver.org/)

marekmosna:
**Description**
Fixes #7484

Fixies logic of PostgreSQL getVersion() method.

### DIFF
--- a/system/Database/Postgre/Connection.php
+++ b/system/Database/Postgre/Connection.php
@@ -128,11 +128,14 @@ class Connection extends BaseConnection
             return $this->dataCache['version'];
         }
 
-        if (! $this->connID || ($pgVersion = pg_version($this->connID)) === false) {
+        if (! $this->connID) {
             $this->initialize();
         }
 
-        return isset($pgVersion['server']) ? $this->dataCache['version'] = $pgVersion['server'] : false;
+        $pgVersion                  = pg_version($this->connID);
+        $this->dataCache['version'] = $pgVersion['server'] ?? '';
+
+        return $this->dataCache['version'];
     }
 
     /**
--- a/phpstan-baseline.neon.dist
+++ b/phpstan-baseline.neon.dist
@@ -66,11 +66,6 @@ parameters:
 			path: system/Database/MySQLi/PreparedQuery.php
 
 		-
-			message: "#^Strict comparison using \\=\\=\\= between array<string, int|string|null> and false will always evaluate to false\\.$#"
-			count: 1
-			path: system/Database/Postgre/Connection.php
-
-		-
 			message: "#^Access to an undefined property CodeIgniter\\\\Database\\\\BaseConnection\\:\\:\\$schema\\.$#"
 			count: 2
 			path: system/Database/SQLSRV/Builder.php


ddevsr (changes requested):
Remove below code in `phpstan-baseline.neon.dist`

```
-
			message: "#^Strict comparison using \\=\\=\\= between array<string, int|string|null> and false will always evaluate to false\\.$#"
			count: 1
			path: system/Database/Postgre/Connection.php
```

ddevsr:
By the way,

Following step how to create PR in https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#php-style

Dont forget signing your commit
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#gpg-signing-old-commits

michalsn:
@marekmosna Almost there! Could you follow the suggestion from https://github.com/codeigniter4/CodeIgniter4/pull/7488#pullrequestreview-1419630508 to fix the PHPStan error?

system/Database/Postgre/Connection.php
```diff
@@ -128,10 +128,12 @@ public function getVersion(): string
...
         }
 
-        if (! $this->connID || ($pgVersion = pg_version($this->connID)) === false) {
+        if (! $this->connID) {
             $this->initialize();
         }
 
+        $pgVersion = pg_version($this->connID);
+
         return isset($pgVersion['server']) ? $this->dataCache['version'] = $pgVersion['server'] : false;
```
kenjis, line 137: The interface is `public function getVersion(): string;`. So `false` should not be returned.

system/Database/Postgre/Connection.php
```diff
@@ -128,11 +128,14 @@ public function getVersion(): string
...
         }
 
-        if (! $this->connID || ($pgVersion = pg_version($this->connID)) === false) {
+        if (! $this->connID) {
             $this->initialize();
         }
 
-        return isset($pgVersion['server']) ? $this->dataCache['version'] = $pgVersion['server'] : false;
+        $pgVersion                  = pg_version($this->connID);
+        $this->dataCache['version'] = $pgVersion['server'] ?? 'PostgreSQL 7.4 or lower';
```
kenjis, line 136: `getVersion()` returns the version number like `'12.1'`, and the number should be used with [version_compare()](https://www.php.net/manual/en/function.version-compare.php)

Is it impossible to get the version number 7.4 or lower?

kenjis: > pg_version() returns an array with the client, protocol and server version. Protocol and server versions are only available if PHP was compiled with PostgreSQL 7.4 or later. 
https://www.php.net/manual/en/function.pg-version.php#refsect1-function.pg-version-description

There is no version requirement in the docs:
https://codeigniter4.github.io/CodeIgniter4/intro/requirements.html#supported-databases

But PostgreSQL 11 will stop receiving fixes on November 9, 2023.
https://www.postgresql.org/about/news/postgresql-153-148-1311-1215-and-1120-released-2637/

@codeigniter4/core-team @codeigniter4/database-team 
Do we need to support versions prior to 7.4?
Why don't we add `(version 7.4 and above only)` in the docs?

samsonasik: PostgreSQL 7.4 seems too old, it seems it need to be documented in docs

marekmosna: > getVersion() returns the version number like '12.1', and the number should be used with [version_compare()](https://www.php.net/manual/en/function.version-compare.php)

that was the question in the related issue #7488

michalsn: If there is no `$pgVersion['server']`, we should return an empty string.

Adding a note `(version 7.4 and above only)` is a good idea.

kenjis, line 136:
```suggestion
        $this->dataCache['version'] = $pgVersion['server'] ?? '';
```

kenjis:
@marekmosna It seems there is no test code for `getVersion()`.
If you can write test code, add test code.
We would appreciate it.

marekmosna:
> @marekmosna It seems there is no test code for `getVersion()`. If you can write test code, add test code. We would appreciate it.

I'm sorry @kenjis but I'm not a php developer so I can't

kenjis:
@marekmosna Okay, I got it.

michalsn (approved):
LGTM!

kenjis (approved):
LGTM!

kenjis:
@codeigniter4/database-team 
It seems PostgreSQL may return the version like `'15.3 (Debian 15.3-1.pgdg110+1)'`.
https://github.com/codeigniter4/CodeIgniter4/actions/runs/5035338925/jobs/9030854318

Is this okay?

michalsn:
Good point. I think we should filter out the version.

system/Database/Postgre/Connection.php
```diff
@@ -128,11 +128,14 @@ public function getVersion(): string
...
         }
 
-        if (! $this->connID || ($pgVersion = pg_version($this->connID)) === false) {
+        if (! $this->connID) {
             $this->initialize();
         }
 
-        return isset($pgVersion['server']) ? $this->dataCache['version'] = $pgVersion['server'] : false;
+        $pgVersion                  = pg_version($this->connID);
+        $this->dataCache['version'] = $pgVersion['server'] ?? '';
```
michalsn, line 136:
```suggestion
        $this->dataCache['version'] = isset($pgVersion['server']) ? 
                preg_match('/^(\d+\.\d+)/', $pgVersion['server'], $matches) ? $matches[1] : '' :
                '';

```
This should work if we deal with something like `15.3 (Debian 15.3-1.pgdg110+1)` - as Kenjis pointed out earlier.

kenjis: Okay, I will take care of it in #7509

PR #5623, opened 2022-01-27
fix: DownloadResponse memory leak

kenjis:
**Description**
Fixes #5620

**Checklist:**
- [x] Securely signed commits
- [] Component(s) with PHPDoc blocks, only if necessary or adds value
- [] Unit testing, with >80% coverage
- [] User guide updated
- [x] Conforms to style guide

PR #7500, opened 2023-05-15
[4.4] Fix output buffering

### LINKED ISSUE #7493: Bug: output buffer
### PHP Version

7.4

### CodeIgniter4 Version

all

Git

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

The application starts monitoring the buffer in the ``CodeIgniter::tryToRouteIt()`` method.
https://github.com/codeigniter4/CodeIgniter4/blob/42b174fc75d01098936514606f6147dffbaccb8b/system/CodeIgniter.php#L813

For a correct response, and in case of a 404 error, if a custom handler is defined, the ``CodeIgniter::gatherOutput()`` method is called. In which we are trying to complete the monitoring of the buffer.

https://github.com/codeigniter4/CodeIgniter4/blob/42b174fc75d01098936514606f6147dffbaccb8b/system/CodeIgniter.php#L1009-L1014

But this code only works if something was passed to the buffer, such as using ``echo``.
If nothing has been written to the buffer, then buffer monitoring is not disabled.
Also, if the ``ob_start()`` function has been called multiple times while the application is running, then ``ob_end_clean()`` will only terminate the last one.

### Steps to Reproduce

```php
namespace CodeIgniter;

use CodeIgniter\Test\Mock\MockCodeIgniter;
use Config\App;

/**
 * @internal
 */
final class StepsToReproduceTest extends CIUnitTestCase
{
    public function testOB()
    {
        $actual['phpunit'] = ob_get_level();

        ob_start();

        $actual['beforeApp'] = ob_get_level();

        (new MockCodeIgniter(new App()))->run();

        $actual['afterApp'] = ob_get_level();

        ob_get_clean();

        $this->assertSame(
            [
                'phpunit' => 1,
                'beforeApp' => 2,
                'afterApp' => 2,
            ],
            $actual
        );
    }
}
```
```
1) CodeIgniter\StepsToReproduceTest::testOB
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
     'phpunit' => 1
     'beforeApp' => 2
-    'afterApp' => 2
+    'afterApp' => 3
 )

```

### Expected Output

Normalize work on the buffer.

### Anything else?

In the \CodeIgniter\Debug\Exceptions class, buffer monitoring also ends incorrectly.
https://github.com/codeigniter4/CodeIgniter4/blob/42b174fc75d01098936514606f6147dffbaccb8b/system/Debug/Exceptions.php#L273-L275

```php
class OBuffer
{
    protected ?int $level;

    public function start(): void
    {
        $this->level = ob_get_level();
        ob_start();
    }

    public function getClean(): string
    {
        $content = '';

        while (ob_get_level() > $this->level) {
            $content .= ob_get_contents();
            ob_end_clean();
        }

        return $content;
    }
}
```

This is an example of code that, in my opinion, will solve the buffer monitoring problem if you replace the ``ob_`` function calls in the ``CodeIgniter::tryToRouteIt()`` and ``CodeIgniter::gatherOutput()`` methods with the appropriate methods class.
It will also remove redundant code.

https://github.com/codeigniter4/CodeIgniter4/blob/42b174fc75d01098936514606f6147dffbaccb8b/system/CodeIgniter.php#L1018-L1022

iRedds:
**Description**
Fixes #7493

Work with output buffering was carried out incorrectly.

To bypass an existing bug in tests, a hack was used
https://github.com/codeigniter4/CodeIgniter4/blob/3b3cfb57ee564c84487729b3fd5459efa6733920/tests/system/CodeIgniterTest.php#L57-L59

In my opinion, this PR correctly completes output buffering, including for exceptions.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -1154,7 +1154,7 @@ class CodeIgniter
         $buffer = '';
 
         while (ob_get_level() > $this->bufferLevel) {
-            $buffer = ob_get_contents();
+            $buffer .= ob_get_contents();
             ob_end_clean();
         }
 
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -36,6 +36,7 @@ use Kint\Renderer\CliRenderer;
 use Kint\Renderer\RichRenderer;
 use Locale;
 use LogicException;
+use Throwable;
 
 /**
  * This class is the core of the framework, and will analyse the
@@ -163,6 +164,11 @@ class CodeIgniter
      * Whether to return Response object or send response.
      */
     protected bool $returnResponse = false;
+
+    /**
+     * Application output buffering level
+     */
+    protected int $bufferLevel;
 
     /**
      * Constructor.
@@ -367,6 +373,7 @@ class CodeIgniter
         try {
             return $this->handleRequest($routes, $cacheConfig, $returnResponse);
         } catch (RedirectException $e) {
+            $this->outputBufferingEnd();
             $logger = Services::logger();
             $logger->info('REDIRECTED ROUTE at ' . $e->getMessage());
 
@@ -389,6 +396,10 @@ class CodeIgniter
             if ($return instanceof ResponseInterface) {
                 return $return;
             }
+        } catch (Throwable $e) {
+            $this->outputBufferingEnd();
+
+            throw $e;
         }
     }
 
@@ -810,7 +821,7 @@ class CodeIgniter
         $this->benchmark->stop('bootstrap');
         $this->benchmark->start('routing');
 
-        ob_start();
+        $this->outputBufferingStart();
 
         $this->controller = $this->router->handle($path);
         $this->method     = $this->router->methodName();
@@ -979,15 +990,8 @@ class CodeIgniter
         // Display 404 Errors
         $this->response->setStatusCode($e->getCode());
 
-        if (ENVIRONMENT !== 'testing') {
-            if (ob_get_level() > 0) {
-                ob_end_flush(); // @codeCoverageIgnore
-            }
-        }
-        // When testing, one is for phpunit, another is for test case.
-        elseif (ob_get_level() > 2) {
-            ob_end_flush(); // @codeCoverageIgnore
-        }
+        echo $this->outputBufferingEnd();
+        flush();
 
         // Throws new PageNotFoundException and remove exception message on production.
         throw PageNotFoundException::forPageNotFound(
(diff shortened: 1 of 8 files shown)


iRedds:
Error installing Imagic. Tests are not available.

michalsn (approved):
That is a really good find and improvement. Thanks!

I have tested it a bit and see no issues.

system/CodeIgniter.php
```diff
@@ -1006,21 +1010,9 @@ protected function display404errors(PageNotFoundException $e)
...
+        $this->output = $this->outputBufferingEnd();
 
         if ($returned instanceof DownloadResponse) {
-            // Turn off output buffering completely, even if php.ini output_buffering is not off
-            if (ENVIRONMENT !== 'testing') {
-                while (ob_get_level() > 0) {
-                    ob_end_clean();
-                }
-            }
-
```
kenjis, line 1023: Revert these lines. It is intentional.
See #5623

iRedds: Before checking the response type, the ``outputBufferingEnd()`` method is called which contains the same code.
```php
        while (ob_get_level() > $this->bufferLevel) {
            $buffer .= ob_get_contents();
            ob_end_clean();
        }
```
In a default application, ``$this->bufferLevel`` will always be ``0``.

kenjis:
```diff
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -1146,6 +1146,7 @@ class CodeIgniter
     protected function outputBufferingStart(): void
     {
         $this->bufferLevel = ob_get_level();
+        dd($this->bufferLevel);
         ob_start();
     }
 
```
```
$this->bufferLevel integer 1
⧉ Called from .../system/CodeIgniter.php:1149 [dd()]
```

iRedds: I was wrong.
But can we move this code into the DownloadResponse class and call it directly in the ``send`` method?
Like this
```php
// DownloadResponse::class
    public function send()
    {
        if (ENVIRONMENT !== 'testing') {
            while (ob_get_level() > 0) {
                ob_end_clean();
            }
        }
        $this->buildHeaders();
        $this->sendHeaders();
        $this->sendBody();

        return null;
    }
```

kenjis: Okay, it might be better.
Please keep the comment `// Turn off output buffering completely, even if php.ini output_buffering is not off`.

kenjis (approved):

PR #7505, opened 2023-05-17
docs: improve Session::destroy()

kenjis:
**Description**
Not recomment to use PHP function.
It is a bad practice.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/libraries/sessions.rst
+++ b/user_guide_src/source/libraries/sessions.rst
@@ -360,10 +360,8 @@ simply use the library's ``destroy()`` method:
 This method will work in exactly the same way as PHP's
 `session_destroy() <https://www.php.net/session_destroy>`_ function.
 
-.. note:: This must be the last session-related operation that you do
-    during the same request. All session data (including flashdata and
-    tempdata) will be destroyed permanently and functions will be
-    unusable during the same request after you destroy the session.
+This must be the last session-related operation that you do during the same request.
+All session data (including flashdata and tempdata) will be destroyed permanently.
 
 .. _session-stop:
 
--- a/user_guide_src/source/libraries/sessions.rst
+++ b/user_guide_src/source/libraries/sessions.rst
@@ -353,11 +353,12 @@ destroy()
 ---------
 
 To clear the current session (for example, during a logout), you may
-simply use either PHP's `session_destroy() <https://www.php.net/session_destroy>`_
-function, or the library's ``destroy()`` method. Both will work in exactly the
-same way:
+simply use the library's ``destroy()`` method:
 
 .. literalinclude:: sessions/037.php
+
+This method will work in exactly the same way as PHP's
+`session_destroy() <https://www.php.net/session_destroy>`_ function.
 
 .. note:: This must be the last session-related operation that you do
     during the same request. All session data (including flashdata and
--- a/user_guide_src/source/libraries/sessions.rst
+++ b/user_guide_src/source/libraries/sessions.rst
@@ -363,6 +363,9 @@ This method will work in exactly the same way as PHP's
 This must be the last session-related operation that you do during the same request.
 All session data (including flashdata and tempdata) will be destroyed permanently.
 
+.. note:: You do not have to call this method from usual code. Cleanup session
+    data rather than destroying the session.
+
 .. _session-stop:
 
 stop()
--- a/user_guide_src/source/libraries/sessions/037.php
+++ b/user_guide_src/source/libraries/sessions/037.php
@@ -1,5 +1,3 @@
 <?php
 
-session_destroy();
-// or
 $session->destroy();


user_guide_src/source/libraries/sessions.rst
```diff
@@ -346,12 +346,13 @@ Destroying a Session
...
 To clear the current session (for example, during a logout), you may
-simply use either PHP's `session_destroy() <https://www.php.net/session_destroy>`_
-function, or the library's ``destroy()`` method. Both will work in exactly the
-same way:
+simply use the library's ``destroy()`` method:
 
 .. literalinclude:: sessions/037.php
 
+This method will work in exactly the same way as PHP's
+`session_destroy() <https://www.php.net/session_destroy>`_ function.
```
kenjis, line 354: The method just calls `session_destroy()`, so this is correct.

But when using the Session library, a cookie that removes the session cookie will be issued
if you call `Session::destory()` or `session_destroy()`.
This behavior is different from PHP's `session_start()` and `session_destroy()`.

I don't know why the cookie is issued.

user_guide_src/source/libraries/sessions.rst
```diff
@@ -346,12 +346,13 @@ Destroying a Session
...
+
 .. note:: This must be the last session-related operation that you do
     during the same request. All session data (including flashdata and
     tempdata) will be destroyed permanently and functions will be
```
kenjis, line 358: `session_destroy()` does not unset any of the global variables associated with the session.
https://www.php.net/manual/en/function.session-destroy.php#refsect1-function.session-destroy-description

So we can still get the session data after calling `Session::destory()`.

```php
    public function destroy()
    {
        $session = session();
        $session->destroy();

        d($session->get(), $_SESSION);
        $session->set('foo', 'bar');
        d($session->get(), $_SESSION);
    }
```
```
/Users/kenji/work/codeigniter/official/CodeIgniter4/app/Controllers/Home.php:20:
array (size=1)
  '__ci_last_regenerate' => int 1684292786

/Users/kenji/work/codeigniter/official/CodeIgniter4/app/Controllers/Home.php:20:
array (size=1)
  '__ci_last_regenerate' => int 1684292786

/Users/kenji/work/codeigniter/official/CodeIgniter4/app/Controllers/Home.php:22:
array (size=2)
  '__ci_last_regenerate' => int 1684292786
  'foo' => string 'bar' (length=3)

/Users/kenji/work/codeigniter/official/CodeIgniter4/app/Controllers/Home.php:22:
array (size=2)
  '__ci_last_regenerate' => int 1684292786
  'foo' => string 'bar' (length=3)
```

Is this intended behavior or a bug?

kenjis: @codeigniter4/core-team Any opinion?

michalsn: Hmm... since we set the `session.use_strict_mode`, the new cookie shouldn't be needed, but I'm wondering about the cases where the hosting provider disabled `ini_set()` for security reasons.

There are definitely differences between the native `session_destroy()` and our implementation.

MGatner: I don't think this is a "bug", but it might be unexpected behavior for some. Since this PR is just an update to the docs I don't think we worry about it, except maybe to add a note of caution.

kenjis: @michalsn We use `ini_set()` outside of the Session.
```
BaseHandler.php
        ini_set('session.save_path', $this->savePath);
bootstrap.php
ini_set('display_errors', '1');
ini_set('display_startup_errors', '1');
CodeIgniter.php
     * not complain when ini_set() function is used.
Exceptions.php
            ini_set('highlight.comment', '#767a7e; font-style: italic');
            ini_set('highlight.default', '#c7c7c7');
            ini_set('highlight.html', '#06B');
            ini_set('highlight.keyword', '#f1ce61;');
            ini_set('highlight.string', '#869d6a');
FileHandler.php
            ini_set('session.save_path', $this->savePath);
            ini_set('session.sid_length', (string) $SIDLength);
MemcachedHandler.php
            ini_set('memcached.sess_prefix', $this->keyPrefix);
Session.php
            ini_set('session.name', $this->sessionCookieName);
        ini_set('session.cookie_samesite', $sameSite);
            ini_set('session.gc_maxlifetime', (string) $this->sessionExpiration);
            ini_set('session.save_path', $this->sessionSavePath);
        ini_set('session.use_trans_sid', '0');
        ini_set('session.use_strict_mode', '1');
        ini_set('session.use_cookies', '1');
        ini_set('session.use_only_cookies', '1');
            ini_set('session.sid_length', (string) $sidLength);
```

kenjis: @MGatner Okay, this is not a bug.
I updated the description.

kenjis:
I added a note: https://github.com/codeigniter4/CodeIgniter4/pull/7505/commits/cba77132b8ff00f2a2a434301d80686ed5ee45f6

It comes from the note in https://www.php.net/manual/en/function.session-destroy.php#refsect1-function.session-destroy-description
and Shield also does not destroy the session when logging out.
https://github.com/codeigniter4/shield/blob/7f4503b3181a2eeee1cdb4a41b9611bfdcb7826f/src/Authentication/Authenticators/Session.php#L829-L838

michalsn (approved):

MGatner (approved):

PR #7506, opened 2023-05-17
feat: `GDHandler` make `WebP` with option quality

ddevsr:
**Description**
With condition of SEO Strategic, we need compress all image web with `WebP` for speed performance.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Images/Handlers/GDHandler.php
+++ b/system/Images/Handlers/GDHandler.php
@@ -267,7 +267,7 @@ class GDHandler extends BaseHandler
                     throw ImageException::forInvalidImageCreate(lang('Images.webpNotSupported'));
                 }
 
-                if (! @imagewebp($this->resource, $target)) {
+                if (! @imagewebp($this->resource, $target, $quality)) {
                     throw ImageException::forSaveFailed();
                 }
                 break;
--- a/user_guide_src/source/changelogs/v4.4.0.rst
+++ b/user_guide_src/source/changelogs/v4.4.0.rst
@@ -120,7 +120,7 @@ Message Changes
 Changes
 *******
 
-- **Images:** Since ``v4.4`` quality WebP not using default PHP.
+- **Images:** The default quality for WebP in ``GDHandler`` has been changed from 80 to 90.
 - **Config:** The deprecated Cookie items in **app/Config/App.php** has been removed.
 - **Config:** Routing settings have been moved to **app/Config/Routing.php** config file.
   See :ref:`Upgrading Guide <upgrade-440-config-routing>`.
--- a/user_guide_src/source/changelogs/v4.4.0.rst
+++ b/user_guide_src/source/changelogs/v4.4.0.rst
@@ -120,6 +120,7 @@ Message Changes
 Changes
 *******
 
+- **Images:** Since ``v4.4`` quality WebP not using default PHP.
 - **Config:** The deprecated Cookie items in **app/Config/App.php** has been removed.
 - **Config:** Routing settings have been moved to **app/Config/Routing.php** config file.
   See :ref:`Upgrading Guide <upgrade-440-config-routing>`.
--- a/user_guide_src/source/changelogs/v4.4.0.rst
+++ b/user_guide_src/source/changelogs/v4.4.0.rst
@@ -87,7 +87,7 @@ Libraries
 
 - **Validation:** Added ``Validation::getValidated()`` method that gets
   the actual validated data. See :ref:`validation-getting-validated-data` for details.
-- **Images:** Now WebP images can be used option ``$quality`` for compression.
+- **Images:** The option ``$quality`` can now be used to compress WebP images.
 
 Helpers and Functions
 =====================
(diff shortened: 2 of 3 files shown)


kenjis:
Isn't this an enhancement?

ddevsr:
> Isn't this an enhancement?

Thanks for correct

michalsn (approved):

user_guide_src/source/libraries/images.rst
```diff
@@ -85,7 +85,7 @@ Image Quality
 
 ``save()`` can take an additional parameter ``$quality`` to alter the resulting image
 quality. Values range from 0 to 100 with 90 being the framework default. This parameter
-only applies to JPEG images and will be ignored otherwise:
+only applies to JPEG and WEBP images, will be ignored otherwise:
```
kenjis, line 88: Add a note (`.. note::`) something like: $quality for WEBP images can be used since v4.4.0.

kenjis:
Add this enhancement in the changelog 4.4.0 to let devs know.
https://github.com/codeigniter4/CodeIgniter4/blob/4.4/user_guide_src/source/changelogs/v4.4.0.rst#libraries

user_guide_src/source/changelogs/v4.4.0.rst
```diff
@@ -87,6 +87,7 @@ Libraries
 
 - **Validation:** Added ``Validation::getValidated()`` method that gets
   the actual validated data. See :ref:`validation-getting-validated-data` for details.
+- **Images:** Now WebP images can be used option ``$quality`` for compression.
```
kenjis, line 90:
```suggestion
- **Images:** The option ``$quality`` can now be used to compress WebP images.
```

user_guide_src/source/libraries/images.rst
```diff
@@ -85,7 +85,9 @@ Image Quality
...
+
+.. note:: Parameter ``$quality`` for WebP can be used since ``v4.4.0``
```
kenjis, line 90: We use ` `` ` for code.

```suggestion
.. note:: The parameter ``$quality`` for WebP can be used since v4.4.0.
```

kenjis:
Thank you for updates. We are close to merge.

system/Images/Handlers/GDHandler.php
```diff
@@ -267,7 +267,7 @@ public function save(?string $target = null, int $quality = 90): bool
                     throw ImageException::forInvalidImageCreate(lang('Images.webpNotSupported'));
                 }
 
-                if (! @imagewebp($this->resource, $target)) {
+                if (! @imagewebp($this->resource, $target, $quality)) {
```
kenjis, line 270: This PR changes the default quality from -1 to 90.
https://www.php.net/manual/en/function.imagewebp.php#refsect1-function.imagewebp-description
I don't know what -1 means.
Does this degrade the image quality?

michalsn: It's probably the same as in `imagejpeg()` - https://www.php.net/manual/en/function.imagejpeg.php

> The default (-1) uses the default IJG quality value (about 75).

kenjis: 80?
https://github.com/php/php-src/blob/8d0345d94ef98bfe29df264b338aeef16ddefda9/ext/gd/libgd/gd_webp.c#L121

michalsn: Yes, it appears to be a fixed value.

kenjis: Since the behavior changes (80 -> 90), should we add this to changelog?
This doesn't seem to destroy the application, though.

michalsn: Well... it's a change in behavior, so we can mention it.

ddevsr: I think we must change default $quality to 80, And i can set parameter $quality to PNG with 0 - 9.

kenjis: Do you mean that the default quality for webp is 80, for jpeg is 90?

michalsn: I think we should stay with 90 as the default value for quality. Having a separate value for WEBP makes not much sense to me.

Regarding PNG, we can't speak about the quality since these files are lossless. The image quality, as we know it from JPEG or WEBP, will not be changed. Only the compression - file size will be changed.

ddevsr: Its okay, i agree to `$quality` to `WebP` using default 90. And with this changes make developer have preferences to changes quality compression for WebP

kenjis: @ddevsr This changes the default quality for webp from 80 to 90.
Can you add it in https://github.com/codeigniter4/CodeIgniter4/blob/4.4/user_guide_src/source/changelogs/v4.4.0.rst#changes ?

user_guide_src/source/changelogs/v4.4.0.rst
```diff
@@ -119,6 +120,7 @@ Message Changes
 Changes
 *******
 
+- **Images:** Since ``v4.4`` quality WebP not using default PHP.
```
kenjis, line 123:
```suggestion
- **Images:** The default quality for WebP in ``GDHandler`` has been changed from 80 to 90.
```

kenjis (approved):

kenjis:
@ddevsr Thank you!

PR #7509, opened 2023-05-21
fix: PostgreSQL getVersion() output

kenjis:
**Description**
Follow-up #7488

- add test for getVersion()
- fix output version string

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide


michalsn (approved):
Thank you!

PR #7519, opened 2023-05-24
[4.4] refactor: a single point of sending the Response.

iRedds:
**Description**
This PR refactors the CodeIgniter class code so that all successful responses, except those caused by exceptions, are sent from one place.

- The ``CodeIgniter::handleRequest()`` method now always returns a Response.
- Moved HTTP cli check and page cache check to ``CodeIgniter::handleRequest()`` method. Because these methods generate a Response and are actually part of the request processing.
- Event call ``Events::trigger('pre_system')`` moved to method ``CodeIgniter::handleRequest()`` and is now on the same "level" as ``Events::trigger('post_system')``
- The ``CodeIgniter::$returnResponse`` property has been deprecated as it is no longer used.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/user_guide_src/source/changelogs/v4.4.0.rst
+++ b/user_guide_src/source/changelogs/v4.4.0.rst
@@ -138,6 +138,7 @@ Deprecations
   are deprecated. Because these methods have been moved to ``BaseExceptionHandler`` or
   ``ExceptionHandler``.
 - **Autoloader:** ``Autoloader::sanitizeFilename()`` is deprecated.
+- **CodeIgniter:** ``CodeIgniter::$returnResponse`` property is deprecated. no longer used.
 
 Bugs Fixed
 **********
--- a/user_guide_src/source/changelogs/v4.4.0.rst
+++ b/user_guide_src/source/changelogs/v4.4.0.rst
@@ -138,7 +138,7 @@ Deprecations
   are deprecated. Because these methods have been moved to ``BaseExceptionHandler`` or
   ``ExceptionHandler``.
 - **Autoloader:** ``Autoloader::sanitizeFilename()`` is deprecated.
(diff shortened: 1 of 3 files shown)


kenjis:
I don't know why but sometimes Psalm reports errors that cannot reproduce.
I cannot reproduce the following error:

```
Run vendor/bin/psalm
Target PHP version: 7.4 (inferred from composer.json).
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░░░░░░░

ERROR: UndefinedClass - tests/system/CodeIgniterTest.php:141:33 - Class, interface or enum named Tests\Support\Controllers\Popcorn does not exist (see https://psalm.dev/019)
        $routes->set404Override('Tests\Support\Controllers\Popcorn::pop');
```
https://github.com/codeigniter4/CodeIgniter4/actions/runs/5067800693/jobs/9099299107?pr=7519

And I reran the check, it passed.

user_guide_src/source/changelogs/v4.4.0.rst
```diff
@@ -138,6 +138,7 @@ Deprecations
   are deprecated. Because these methods have been moved to ``BaseExceptionHandler`` or
   ``ExceptionHandler``.
 - **Autoloader:** ``Autoloader::sanitizeFilename()`` is deprecated.
+- **CodeIgniter:** ``CodeIgniter::$returnResponse`` property is deprecated. no longer used.
```
kenjis, line 141:
```suggestion
- **CodeIgniter:** ``CodeIgniter::$returnResponse`` property is deprecated. No longer used.
```

kenjis (approved):

kenjis:
@iRedds Thank you!

PR #7522, opened 2023-05-25
Email refactor

### LINKED ISSUE #7520: Bug: The error occurred when the built-in Email library was forced to switch to TLS while setting port 465
### PHP Version

8.1

### CodeIgniter4 Version

4.3.4

Windows

### Which server did you use?

apache

### Database

_No response_

### What happened?

The built-in Email library, when specifying the SMTP port as 465, disregards the SMTPCrypto settings, and directly sets the encryption method as tls://, which is incomprehensible to me. 

Many email services provide SMTP SSL ports that are designated as 465, but the protocols they offer may only be one of ssl:// and tls://, or one may be stable while the other is not. This choice should ideally be decided by the user. 

My current situation is that in Codeigniter3, the smtp_crypto was set to ssl on port 465 and has been running stably. But after upgrading the system to Codeigniter4, I found that the framework forcibly sets tls:// when setting port 465; causing the email failure rate to be as high as 30%.

For now, there's no solution, so I had to temporarily set it to port 25 without specifying SMTPCrypto, and no errors have occurred so far. In the old system, the same SMTP parameters of 465+SSL combination also never caused errors. 

The SMTP service I use is from 126.com's email service, and the SMTP server address is smtp.126.com.

### Steps to Reproduce

```php
$config['protocol'] = 'smtp';
$config['SMTPHost'] = 'smtp.126.com';
$config['SMTPPort'] = 465;
$config['SMTPCrypto'] = 'ssl';
$config['SMTPUser'] = '***';
$config['SMTPPass'] = '***';

$email->initialize($config);

$email->setFrom('your@example.com', 'Your Name');
$email->setTo('someone@example.com');
$email->setSubject('Email Test');
$email->setMessage('Testing the email class.');

$email->send();
```

### Expected Output

To send emails steadily and normally.

### Anything else?

_No response_

ddevsr:
**Description**
Fixed #7520.

Reference:
1. https://sendgrid.com/blog/whats-the-difference-between-ports-465-and-587/
2. https://mailtrap.io/blog/smtp-ports-25-465-587-used-for/

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -1854,10 +1854,8 @@ class Email
 
         $ssl = '';
 
-        if ($this->SMTPPort === 465) {
-            $ssl = 'tls://';
-        } elseif ($this->SMTPCrypto === 'ssl') {
-            $ssl = 'ssl://';
+        if (in_array($this->SMTPCrypto, ['tls','ssl'], true)) {
+            $ssl = $this->SMTPCrypto . '://';
         }
 
         $this->SMTPConnect = fsockopen(
--- a/system/Email/Email.php
+++ b/system/Email/Email.php
@@ -1854,7 +1854,7 @@ class Email
 
         $ssl = '';
 
-        if (in_array($this->SMTPCrypto, ['tls','ssl'], true)) {
+        if (in_array($this->SMTPCrypto, ['tls', 'ssl'], true)) {
             $ssl = $this->SMTPCrypto . '://';
         }
 


kenjis:
Why do we need this change?

ddevsr:
I based on https://github.com/bcit-ci/CodeIgniter/blob/develop/system/libraries/Email.php#L2022, No strict port to protocol

kenjis:
CI3 implementation is not a reference here.
It is an old implementation, and we have improved it.
See #3430

If you change the current implementation, please show the reasoning by RFC or something.
In the [reference](https://sendgrid.com/blog/whats-the-difference-between-ports-465-and-587/) you wrote:
- Tl;dr Port 465 is used for implicit TLS, however, port 587 and startTLS are preferred.

iRedds:
If the port for SSL and TLS schemes is not specified, then we use 465.
If port 578 (stratTLS) is specified then TLS scheme.

MGatner:
I don't know enough about these protocol specifics to weigh in on this. @ddevsr did you check out iRedds' suggestion?

PR #7540, opened 2023-06-04
docs: remove unneeded @throws in CodeIgniter

### LINKED ISSUE #7434: Bug: PhpStorm warnings in public/index.php
### PHP Version

8.2

### CodeIgniter4 Version

4.3.3 updated today

Manual (zip or tar.gz)

Windows

### Which server did you use?

apache

### Database

MariaDB 10.4.28

### What happened?

phpStorm throwing Warnings on these 2 lines now.

Line #16 - Can be replaced with 'const' syntax

// Path to the front controller (this file)
define('FCPATH', )__DIR__ . DIRECTORY_SEPARATOR;

Line #67 - Unhandled \CodeIgniter\Router\Exceptions\RedirectException
$app->run();

### Steps to Reproduce

Open in phpStorm
Top right Warning sign.

### Expected Output

No Warnings.

### Anything else?

_No response_

kenjis:
**Description**
Closes #7434
See https://github.com/codeigniter4/CodeIgniter4/issues/7434#issuecomment-1507972531

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/system/CodeIgniter.php
+++ b/system/CodeIgniter.php
@@ -310,8 +310,6 @@ class CodeIgniter
      * makes all of the pieces work together.
      *
      * @return ResponseInterface|void
-     *
-     * @throws RedirectException
      */
     public function run(?RouteCollectionInterface $routes = null, bool $returnResponse = false)
     {
--- a/system/Test/FeatureTestCase.php
+++ b/system/Test/FeatureTestCase.php
@@ -148,9 +148,6 @@ class FeatureTestCase extends CIUnitTestCase
      * instance that can be used to run many assertions against.
      *
      * @return FeatureResponse
-     *
-     * @throws Exception
-     * @throws RedirectException
      */
     public function call(string $method, string $path, ?array $params = null)
     {
--- a/system/Test/FeatureTestTrait.php
+++ b/system/Test/FeatureTestTrait.php
@@ -138,9 +138,6 @@ trait FeatureTestTrait
      * instance that can be used to run many assertions against.
      *
      * @return TestResponse
-     *
-     * @throws RedirectException
-     * @throws Exception
      */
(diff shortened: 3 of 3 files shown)


system/CodeIgniter.php
```diff
@@ -310,8 +310,6 @@ private function configureKint(): void
      * makes all of the pieces work together.
      *
      * @return ResponseInterface|void
-     *
-     * @throws RedirectException
```
samsonasik, line 314: since the method has `LogicException` throw, how about change to `@throws LogicException` ?

kenjis: I think we should write Exceptions in `@throws` which should be caught in the client code.
LogicException should not be caught.

paulbalandan: I think adding `@throws` doesn't mean it have to be caught in client code. It's a way of explicitly telling that calling the function may throw this exception.

kenjis: @paulbalandan Do you think we should write all Exceptions that may be thrown?

paulbalandan: Not really *all*. Only those that are meant to be checked by the framework and/or user. Please read this [PHPStan article](https://phpstan.org/blog/bring-your-exceptions-under-control) for details.

kenjis: I read the article.
I think checked exceptions should be caught. So we should write only checked exceptions in `@thrown`.
Something wrong with my understanding?

iRedds:
You forgot to fix \CodeIgniter\Test\FeatureTestTrait and \CodeIgniter\Test\FeatureTestCase.

kenjis:
@iRedds Thanks. Removed.

paulbalandan (approved):

